PR #3618, opened 2017-12-24
Hide block duration flags.

brian-brazil:
Users are starting to use these mistakenly thinking they'll help
with issues, and thus causing some confusion.
Thus hide them and make it clear that they're only there for testing
reasons.

@fabxc @grobie


fabxc:
👍

redbaron:
is there a way to cap or at least estimate memory usage of a given prometheus instance?

Issue #3351, opened 2017-10-25
Support lables/{label}/values through the remote read API


tomwilkie:
From IRC:
> Main reason I noticed was I was using the metric() param in Grafana which reads from label/%/values. Wasn't sure if it was intentional or not that the end point didn't hit remote read. I could kinda imagine arguments for and against :)

We'll probably want to move to gRPC before we do this, to make the API handling simpler.

AlekSi:
I think `/api/v1/series` can be added to this issue too.

lgogolin:
Any update on this ? 

Getting labels from remote_read would be really helpful

kkotamar:
This feature will be very  useful, if prometheus data is read from remote completely.

lgogolin:
@kkotamar don't know what You mean, You can do remote_read from prometheus

Anyway, there was already a PR adding this feature:
https://github.com/prometheus/prometheus/pull/3620

But was closed, because someone said "it would be nice to have gRPC", ok nice but this is opened for almost a year a nothing happened

aeciopires:
I would also like this feature implemented

bwplotka:
Added `/labels` to the title as if we would do values, we need label names as well.

The current proposal is to just existing JSON APIs: 

* https://prometheus.io/docs/prometheus/latest/querying/api/#getting-label-names
* https://prometheus.io/docs/prometheus/latest/querying/api/#querying-label-values

Commented on https://github.com/prometheus/prometheus/pull/7076 as well.

Harkishen-Singh:
I was thinking to take this, but it seems like #7078 does the work. A gentle reminder to @bwplotka for review.

thislyc:
Mogul，how's the issue going

startan:
Is this problem very difficult to solve? Why hasn't there been any progress after so many years?

beorn7:
https://github.com/prometheus/prometheus/pull/11877 is a PR addressing this issue. It's still under review. There might be many reasons why the review is proceeding slowly. Currently, the ball is in @sgrzemski's court (since late September). Maybe they are not available anymore to continue the work. If you want to pick up the work, you are welcome to do so (but you should say so on the [PR](https://github.com/prometheus/prometheus/pull/11877) to avoid redundant work).

startan:
@sgrzemski Is there any progress on this MR?

sgrzemski:
The code is more less ready, will be testing it tomorrow with some real life examples :) 
[...]

PR #3620, opened 2017-12-24
feat(remote read): read labelvalues

caitong93:
@tomwilkie @brian-brazil 
This is for issue https://github.com/prometheus/prometheus/issues/3351
The pr add additional fields in `prompb.ReadRequest` and `prompb.ReadResponse` to allow reading label values from adapters
```
message ReadRequest {
  repeated Query queries = 1;
  string labelName = 2;
}

message ReadResponse {
  // In same order as the request's queries.
  repeated QueryResult results = 1;
  repeated string labelValues = 2;
}

```


tomwilkie:
Thanks @caitong93! 

I’m not sure we want to add this to the same read endpoint though, it really deserves its own pair of protos for request and response and a fresh URL for it. 

I’d been waiting until I had a chance to move this interface to gRPC to implement it, but if you’re keen could you make the changes to this PR?

caitong93:
@tomwilkie Thanks for review!
I agreed that label values should have its own protos and url. Then I think it will look like

```
message LabelValuesRequest {
  string labelName = 2;
}

message LabelValuesResponse {
  repeated string labelValues = 2;
}
```

the config 

```
remote_read:
 - url: http://remote4/read
   label_values_path: `/labelvalues`
```

tomwilkie:
@caitong93 looking good! Couple of comments:
-  Field names should use underscores as per the style guide (https://developers.google.com/protocol-buffers/docs/style). 
- I would specify a separate URL for the labels values call, not just a extra path - it will be more flexible.
Thanks!

caitong93:
@tomwilkie updated

AlekSi:
Gentle poke for @tomwilkie

tomwilkie:
Thanks @AlekSi.  Code looks good, a bunch of minor nits but nothing major.

Before I address those, on #3351 I said "We'll probably want to move to gRPC before we do this, to make the API handling simpler." - which this PR doesn't do.  I inclined to accept this PR anyway, but I think it would be good to get a second opinion from @juliusv first.

config/config.go
```diff
@@ -703,7 +703,8 @@ type RemoteReadConfig struct {
 
 	// RequiredMatchers is an optional list of equality matchers which have to
 	// be present in a selector to query the remote read endpoint.
-	RequiredMatchers model.LabelSet `yaml:"required_matchers,omitempty"`
+	RequiredMatchers model.LabelSet   `yaml:"required_matchers,omitempty"`
+	LabelValuesURL   *config_util.URL `yaml:"label_values_url,omitempty"`
```
brian-brazil: If you're adding a new field please update the documentation.

caitong93:
@tomwilkie 
I didn't understand `move to gRPC` at first. But now I know it is `moving remote read and write from http to gRPC`. I would like to give a try.  I can make changes to this pr, or sending another pr after this one. WDYT ?

Issue #2580, opened 2017-04-05
Add query hints to remote read


brian-brazil:
As discussed back at promcon, we should add a map[string]string with hints like the function wrapping the selector, the range and the step. This is unstructured and experimental as we don't know yet what if any of this will prove useful for things like downsampling optimisations on the other end.

At some point (likely after the rest of remote read is considered non-experimental), this should be changed to proper structured data when we have a good idea from multiple implementations what information is needed.

fabxc:
I don't see an immediate benefit to making it unstructured initially. If you change something, the other end won't work anymore either way. Just that you won't notice it at compile time with `map[string]string`.

None is more or less work than the other really.

fabxc:
This would be another argument to `Select` I assume, like `Select(*Params, ...labels.Matcher)`?

What I can see immediately would be:

```go
type Aggregation int

const (
    AggrNone Aggregation = iota
    AggrAvg
    AggrMin
    AggrMax
    AggrCounter
    AggrCount
)

type Param struct {
    Step int64
    Aggr Aggregation
}
```

The aggregation would make some assumptions already about what a given PromQL function needs. We could just make it a function name of course. But I'd believe that every implementation making use of these would just end up mapping our dozens of functions theirselves, which introduces lots of room for error.

brian-brazil:
I don't think we should make any assumptions about how downsampling works and what information it finds useful for a first version. Thus why I propose passing it down and only finalising this API after there's more experience.

fabxc:
Then just the raw info

```
type Params struct {
    Step int64
    Func string
}
```

Strictly speaking `sum/avg/...` are not functions. Should this go into an extra `Aggr string` field?

brian-brazil:
There's no overlap between function names and operator names, so for a first pass I'd suggest putting them all in one field.

henridf:
I'd like to work on this - I've started taking a look and I'll report back here if I run into anything, otherwise will hopefully have a PR soon.

PR #3627, opened 2017-12-26
Change StalenessDelta to be per-statement

jacksontj:
With the current package-global StalenessDelta you have to pick a single
StalenessDelta for all queries that will be ever done in your process.
This is difficult, when the queries you are processing vary wildly in
duration (1h vs 30d). In the case of a 1h query, a 5m StalenessDelta is
okay (maybe too long) but for a 30d range that might be too low (as the
step interval may be significnatly longer).

This change allows the user of the engine to define a function to
determine a StalenessDelta based on the query that you are processing.
This maintains compatibility by providing a default function to return a
5m duration (the old hard-code).

### DIFF
--- a/promql/ast.go
+++ b/promql/ast.go
@@ -72,6 +72,10 @@ type EvalStmt struct {
 	Start, End model.Time
 	// Time between two evaluated instants for the range [Start:End].
 	Interval time.Duration
+
+	// StalenessDelta determines the time since the last sample after which a time
+	// series is considered stale.
+	stalenessDelta time.Duration
 }
 
 // RecordStmt represents an added recording rule.
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -315,6 +315,14 @@ type Queryable interface {
 	Querier() (local.Querier, error)
 }
 
+// Function to determine the StalenessDelta of an Eval Stmt
+type StalenessDeltaFunc func(*EvalStmt) time.Duration
+
+// Default function to return 5m
+func DefaultStalenessDelta(*EvalStmt) time.Duration {
+	return time.Minute * 5
+}
+
 // NewEngine returns a new engine.
 func NewEngine(queryable Queryable, o *EngineOptions) *Engine {
 	if o == nil {
@@ -332,12 +340,14 @@ func NewEngine(queryable Queryable, o *EngineOptions) *Engine {
 type EngineOptions struct {
 	MaxConcurrentQueries int
 	Timeout              time.Duration
+	StalenessDeltaFunc   StalenessDeltaFunc
 }
 
 // DefaultEngineOptions are the default engine options.
 var DefaultEngineOptions = &EngineOptions{
 	MaxConcurrentQueries: 20,
 	Timeout:              2 * time.Minute,
+	StalenessDeltaFunc:   DefaultStalenessDelta,
 }
 
 // NewInstantQuery returns an evaluation query for the given expression at the given time.
@@ -370,11 +380,17 @@ func (ng *Engine) NewRangeQuery(qs string, start, end model.Time, interval time.
 
 func (ng *Engine) newQuery(expr Expr, start, end model.Time, interval time.Duration) *query {
 	es := &EvalStmt{
-		Expr:     expr,
-		Start:    start,
-		End:      end,
-		Interval: interval,
+		Expr:     expr,     // promql
+		Start:    start,    // beginning of range
+		End:      end,      // end of range (same as start for instant)
+		Interval: interval, // Step
 	}
+	es.stalenessDelta = ng.options.StalenessDeltaFunc(es)
+	// Ensure that the staleness delta is non-negative
+	if es.stalenessDelta < 0 {
+		panic("must have non-negative staleness delta")
+	}
+
 	qry := &query{
 		stmt:  es,
 		ng:    ng,
@@ -466,6 +482,7 @@ func (ng *Engine) execEvalStmt(ctx context.Context, query *query, s *EvalStmt) (
 		evaluator := &evaluator{
 			Timestamp: s.Start,
 			ctx:       ctx,
+			es:        s,
 		}
 		val, err := evaluator.Eval(s.Expr)
 		if err != nil {
@@ -499,6 +516,7 @@ func (ng *Engine) execEvalStmt(ctx context.Context, query *query, s *EvalStmt) (
 		evaluator := &evaluator{
 			Timestamp: ts,
 			ctx:       ctx,
+			es:        s,
 		}
 		val, err := evaluator.Eval(s.Expr)
 		if err != nil {
@@ -576,13 +594,13 @@ func (ng *Engine) populateIterators(ctx context.Context, querier local.Querier, 
 				n.iterators, queryErr = querier.QueryInstant(
 					ctx,
 					s.Start.Add(-n.Offset),
-					StalenessDelta,
+					s.stalenessDelta,
 					n.LabelMatchers...,
 				)
 			} else {
 				n.iterators, queryErr = querier.QueryRange(
 					ctx,
-					s.Start.Add(-n.Offset-StalenessDelta),
+					s.Start.Add(-n.Offset-s.stalenessDelta),
 					s.End.Add(-n.Offset),
 					n.LabelMatchers...,
 				)
@@ -627,6 +645,7 @@ func (ng *Engine) closeIterators(s *EvalStmt) {
 // cancellation of its context it terminates.
 type evaluator struct {
 	ctx context.Context
+	es  *EvalStmt
 
 	Timestamp model.Time
 }
@@ -813,7 +832,7 @@ func (ev *evaluator) vectorSelector(node *VectorSelector) vector {
 	for _, it := range node.iterators {
 		refTime := ev.Timestamp.Add(-node.Offset)
 		samplePair := it.ValueAtOrBeforeTime(refTime)
-		if samplePair.Timestamp.Before(refTime.Add(-StalenessDelta)) {
+		if samplePair.Timestamp.Before(refTime.Add(-ev.es.stalenessDelta)) {
 			continue // Sample outside of staleness policy window.
 		}
 		vec = append(vec, &sample{
@@ -1385,10 +1404,6 @@ func shouldDropMetricName(op itemType) bool {
 		return false
 	}
 }
-
-// StalenessDelta determines the time since the last sample after which a time
-// series is considered stale.
-var StalenessDelta = 5 * time.Minute
 
 // A queryGate controls the maximum number of concurrently running and waiting queries.
 type queryGate struct {


jacksontj:
This is necessary for metrics proxies (such as [promxy](https://github.com/jacksontj/promxy) to function.

brian-brazil:
What are you trying to do that a single value doesn't cut it? It is recommended to have a single interval per Prometheus to keep things sane, https://www.robustperception.io/keep-it-simple-scrape_interval-id/

jacksontj:
@brian-brazil I'm not sure I follow the concern here. The original issue states that the issue is with StalenessDelta, but the link you pasted is for scrape interval.

To restate the issue *again* the issue comes when querying ranges of data with a `step` that outstrips the StalenessDelta. For example, if I where to get data from prometheus at a step of `600s` (6m) then there would be *many* holes in the outputted graph, because the datapoints aren't within the StalenessDelta (currently hardcoded at 5m).

This PR gives the users of the promql.Engine a mechanism to override these values depending on their use case. In this example the client could set the StalenessDelta as a multiple of the step (or to another hard-coded value).

brian-brazil:
> To restate the issue again the issue comes when querying ranges of data with a step that outstrips the StalenessDelta.

Changing the staleness delta doesn't change that. You need to change the query in that case, not the staleness delta. In particular you need to increase the range on your `rate` function.

jacksontj:
> Changing the staleness delta doesn't change that. You need to change the query in that case, not the staleness delta. In particular you need to increase the range on your rate function.

Actually the staleness delta is *exactly* the thing to change. If we look at the code we can see that the [vectorSelector](https://github.com/prometheus/prometheus/blob/release-1.7/promql/engine.go#L816) (as an example) uses the StalenessDelta to determine which values belong to the vector. If the iterators are populated with data that has a long step (greater than the StalenessDelta) then datapoints will be missing in the resulting vector.

brian-brazil:
> If the iterators are populated with data that has a long step (greater than the StalenessDelta) then datapoints will be missing in the resulting vector.

That is the intended behaviour, intervals greater than 2 minutes are not advised. Can you expand more on the type of data you're working on where this is coming up?

jacksontj:
> That is the intended behaviour, intervals greater than 2 minutes are not advised. Can you expand more on the type of data you're working on where this is coming up?

This is not an unreasonable default for scrape interval, but retrieval of data may not be able to adhere.

The specific case I'm working with is with [promxy](https://github.com/jacksontj/promxy) -- an aggregating promql proxy to give you a unified access point for all your various prometheus clusters. Promxy is implemented as a `Querier` which in turn fetches its data from prometheus over the HTTP api. With a hard-coded 5m StalenessDelta that means that promxy *must* fetch data from the downstream prometheus hosts with a `step` of no greater than 5m. This is extremely limiting especially for fetching data over *large* time ranges. For example, if I where to fetch the data for 30d of data that would come out to `172800` datapoints (at 15s scrape interval). With the current hard-coded 5m I could reduce the number of points fetched to `8640`, but thats still significantly more datapoints that something like grafana would show. Meaning I'm stuck putting additional load on the downstream asking for *more* data than necessary just to appease promql's hard-coded `StalenessDelta`.

With the patch in this PR I could provide a function that could calculate a `StalenessDelta` for the query I'm returning. In this case where I'm getting 30d of data, I may ask for data every 10m (which is `4320` points) with this patch I could set the `StalenessDelta` for this Statement execution to 20m -- which would allow for missing datapoints etc.

The goal of `StalenessDelta` is to determine "how old is too old" for a datapoint, or in other words if there was a "gap" in data because there wasn't a scrape. To do this the `StalenessDelta` needs to be relative to the Step. In the case of local-to-prometheus this is done because the `ScrapeInterval` is lower than the `StalenessDelta`, but for other people trying to implement the `Querier` interface such queries maybe be excessively expensive (meaning this issue isn't *unique* to promxy).

brian-brazil:
I can kinda see what you are trying to do, but this is not a good approach as you will often be returning almost random data to your users due to ignoring lots of data points.

When you're doing this sort of thing you need the raw data, without involving PromQL. I'd suggest looking at the `query` endpoint with a range selectior rather than `query_range`, and possibly also the remote read endpoint.

jacksontj:
First things first, down sampling isn't random. If the user wants to
downsample and they get a downsampled response they are happy. This is done
in prom sometimes with aggregation clusters etc.

As for fetching raw data through querry, this is exactly what I used to do,
but in the case of a user asking for a large time range (or large number of
timeseries) this can easily overwhelm Prometheus, and more importantly the
original query from the client (e.g. grafana) only wanted a specific step,
so we don't need 100% since they defined what step they want.

As it stands today the only way to get fork/reimplement promql is to fetch
every data point which isn't scalable. Similarly I'd expect Prometheus to
do similar things under the hood for a query_range with large steps.

On Dec 26, 2017 12:33 PM, "Brian Brazil" <notifications@github.com> wrote:

I can kinda see what you are trying to do, but this is not a good approach
as you will often be returning almost random data to your users due to
ignoring lots of data points.

When you're doing this sort of thing you need the raw data, without
involving PromQL. I'd suggest looking at the query endpoint with a range
selectior rather than query_range, and possibly also the remote read
endpoint.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/prometheus/prometheus/pull/3627#issuecomment-354010322>,
or mute the thread

brian-brazil:
> First things first, down sampling isn't random

That it's deterministic doesn't mean that throwing away 90% of your data is going to produce a good result.

> If the user wants to downsample and they get a downsampled response they are happy.

That depends on whether the response is meaningful and the meaning is what the user expects.

> As for fetching raw data through querry, this is exactly what I used to do, but in the case of a user asking for a large time range (or large number of timeseries) this can easily overwhelm Prometheus

This is one of the big reasons we recommend against trying to build something like this. Doing it correctly usually requires moving a lot of data about, and it's easy to take out all of your monitoring with a single query.

> Similarly I'd expect Prometheus to do similar things under the hood for a query_range with large steps.

It doesn't, that's up to the user and we for example have been working with Grafana to help make this easier.

The staleness delta is not the solution to the problem you're having, it's merely a symptom. If you can share with me an example query that you're having issues with I can show you how to approach this.

jacksontj:
> That it's deterministic doesn't mean that throwing away 90% of your data is going to produce a good result.

My reply there was to point out that downsampling != random (as you had implied previously 

> often be returning almost random data

"good result" is relative to the desired result. If the user requests a downsampled result, then returning a downsampled result is completely fine. The issue of "goodness" is when the desired and actual outcomes differ.

> This is one of the big reasons we recommend against trying to build something like this.

The goal here of this proxy is less about aggregation and more about metric discovery. With any moderate install base you end up with *many* prometheus instances. These all end up as a variety of sources in grafana which confuses users to no end. And then if you want to increase the reliability of prometheus by having redundant hosts, then you introduce problems there as well (no backfilling, etc.). For this sort of a system where the server product wants to ignore as much as it can about the surrounding infrastructure, a proxy that can route metrics sounds exactly like what it wants.

> The staleness delta is not the solution to the problem you're having, it's merely a symptom.

If that is the case, what is the problem I am having? From the data I have I see missing datapoints coming back from the query_range endpoints, and I have to deal with them in some mechanism. Right now the source prometheus node has a fixed 5m staleness delta, meaning if there isn't a datapoint within 5m of the `step` then it returns no datapoint. For me to then infer up a level that this means there was *no* data since the last step is incorrect as well.

To make this a concrete example, lets say there is a metric `foo` and it is being scraped every 1m. If I where to send a query_range for foo for 2d of data with a step of 10m starting at time `t`. In this case we'd expect a response of `288` datapoints. Lets say there was a gap in scrape data, specifically from `t+1m -> t+11m`. In this case prometheus' response at time `t+10m` will be null, as there was no value in the StalenessDelta. This leaves the caller (a remote instance of promql) to show that there was no data from `t -> t+20` because there was no datapoint there. In this case of a downsampled response, its not unreasonable to expect/want that to be smoothed over. If I'm looking at a report of the last 3m of data I don't want holes for a day because there was a gap in data for 10m. Conversely if I had a tight scrape interval of 1s then I wouldn't want data thats 5m to count as relevant, because that means that 300 scrapes failed (which is a lot)!

In the case of prometheus the server promql works alright because the scrape interval is consistent, and it has all of the data local to the box. In the case of *many* datasources, the `promql ` package isn't at all equipped to handle this. With a single package-level global there is no way for me to change it based on datasource, so I have to (at best) create different processes in different binaries to handle multiple sources that want different StalnessDeltas.

I could see an argument for moving `StalenessDelta` to be an engine-wide metric, but this path of "we can't change the StalenessDelta" doesn't make sense to me. If promql is supposed to be used for a variety of sources (which the `Querier` interface and various remote read plugins seem to imply) then this should be configurable at a more granular level than it currently is. IMO doing it on a per-request basis would be ideal, as it gives the caller the most flexibility in setting the value.

brian-brazil:
I've had a look at your code. Instead of `retCount` you should use a `WaitGroup`, which fix the race you have there.

From what I can see for the case your support there's a much simpler way to do what you want using remote read and the remote read endpoint. Configure all the sub-Prometheus servers as remote read endpoints of the main Prometheus. This isn't quite the way to do this, but it'll do similar to what you have. The caveats about taking out all your monitoring at once remain.

jacksontj:
>  which fixs the race you have there.

What race condition? That code uses channels and goroutines, there is a single goroutine waiting for the responses, and because of it there is no need for locks etc. Doing it this way means I can start processing responses as soon as the first one shows up, instead of waiting until the last one responds. If you see a race, could you expand on it?

> there's a much simpler way to do what you want using remote read and the remote read endpoint.

How is this *much simpler*? I can see that this would hit a different (unstable) endpoint, but that doesn't solve any of the issues we're talking about in this PR, it simply moves my implementation to an unstable API-- unless remote read/write somehow give you access to controlling promql internals?

brian-brazil:
> And then if you want to increase the reliability of prometheus by having redundant hosts, then you introduce problems there as well (no backfilling, etc.).

This is a very hard problem, Prometheus avoids this type of clustering as it's quite difficult to get right, let alone right and reliable. Your code appears to be ignoring this problem. For simplicity I suggest talking to only one of a HA pair.

> These all end up as a variety of sources in grafana which confuses users to no end.

Ah, you're trying to build one of those. You'll need to add in the external_labels of the Prometheus you're talking to in the query result then. Otherwise metrics between different Prometheus servers will clash.

In practice I find it fairly obvious which Prometheus contains which data. For example the Cassandra Prometheus contains Cassandra metrics, the infrastructure Prometheus contains node exporter metrics etc. matching your general architecture.

> From the data I have I see missing datapoints coming back from the query_range endpoints, and I have to deal with them in some mechanism.

If the data isn't being returned due to staleness, then that is as the data is stale and thus you don't want it returned. This is a problem with naive downsampling, you're getting the same result as if the scrape interval was 10m - which is correct. And why a scape interval of no more than 2m is recommended.

> In this case of a downsampled response, its not unreasonable to expect/want that to be smoothed over.

What you want here for gauges is `avg_over_time`, so that all values get smoothed equally and all data is used.

> Conversely if I had a tight scrape interval of 1s then I wouldn't want data thats 5m to count as relevant, because that means that 300 scrapes failed (which is a lot)!

The new staleness handling in 2.0 should cover you there. Where this setting matters there is around crashes, which are expected to be rare.

> IMO doing it on a per-request basis would be ideal, as it gives the caller the most flexibility in setting the value.

Having seen the confusion the global setting already causes, I'd be wary of making it possible to make the semantics more convoluted. You could bump the staleness delta to a high value and leave it to new staleness to handle the rest, it's not the sort of thing you should need to tweak per query.


At a high level, this is not a great way to try and downsample. As you've found for example missing data is a bit tricky to deal with, and you'll also run into issues with counters such as missing resets. The `/query_range` function is intended for use directly in graphs, not as a way to downsample raw data.

It's expected that users would implement a long term storage system that featured downsampling and were made to work with the remote read feature. Downsampling is not a feature that belongs in Prometheus itself, and attempting to shoehorn it in is not likely to end well. If you want to downsample properly you need to work with the raw data directly so that you can handle all the edge cases

When presenting data back to PromQL you also need to make it look right, as in not downsampled. This is what remote read endpoints would need to do, and considering that you're basically implementing an in-process remote read endpoint also what you should do.

> What race condition?

Ah, I misread. I thought the goroutine started earlier than it did.

> it simply moves my implementation to an unstable API-- unless remote read/write somehow give you access to controlling promql internals

It's not that unstable, it's pretty safe to build on and many companies are doing so. It does give you something that no other API does in that it completely bypasses PromQL, and thus gives direct access to the actual raw data including the stale markers.

jacksontj:
> Your code appears to be ignoring this problem. For simplicity I suggest talking to only one of a HA pair.

I'd like some more specifics about what all problems are being ignored, but let that whole thread 

> Ah, you're trying to build one of those. You'll need to add in the external_labels of the Prometheus you're talking to in the query result then. Otherwise metrics between different Prometheus servers will clash.

Already done and in the example config :)

>  This is a problem with naive downsampling, you're getting the same result as if the scrape interval was 10m - which is correct.

Well, this is actually an alignment problem. There is already a concept in getting data thats "close" to the requested time (which is why `StalenessDelta` even exists). The problem here is that you can't control what your promql thinks is stale-- which seems especially crazy since you (as the binary using promql) have more insight into the data you are feeding it.

> What you want here for gauges is avg_over_time, so that all values get smoothed equally and all data is used.

How do you do this from the `Querier` interface? From that level all I know is a labelset and a range of time, there is no mechanism for me to know whether it is a counter or a gauge.

> The new staleness handling in 2.0 should cover you there. 

From looking at master it seems that 2.0 still has a package-level variable (renamed, but still the same thing), which means it doesn't address my concern of more-granular control over `StalenessDelta`.

> Having seen the confusion the global setting already causes, I'd be wary of making it possible to make the semantics more convoluted. 

I'm not sure I follow how this makes things more convoluted? My suggestion is to leave the setting exactly the same for `prometheus` but make it controllable for clients of the `promql` package -- which ash been my intent from the beginning of this PR.

> You could bump the staleness delta to a high value and leave it to new staleness to handle the rest, it's not the sort of thing you should need to tweak per query.

Not sure what you mean by this? In the case of someone implementing the `Querier` interface, you may control that setting based on which datasource you are pulling from. It seems to me that setting the value to the largest you think you could ever want would be *significantly* worse than allowing it to be configured on a per-query (or per engine) basis.

> If you want to downsample properly you need to work with the raw data directly so that you can handle all the edge cases

This is a somewhat flawed statement, especially since the scaling of prometheus is almost entirely based around rescraping and aggregation (which all "downsample" in their own way without access to the original data).

> When presenting data back to PromQL you also need to make it look right, as in not downsampled. This is what remote read endpoints would need to do, and considering that you're basically implementing an in-process remote read endpoint also what you should do.

I'm not sure what you mean by this? You mean to say that from downsampled data I need to present to promQL data that is *not* downsampled? That is by definition not possible (I could make up data, but thats not generally a good plan). The problem I'm running into is that promql has some assumptions about what the data looks like, and I can't configure them (`StalenessDelta` in this case). So if I'm understanding your suggestion it is to return data *as if* I had datapoints every <whatever_the_scrape_interval_is> -- meaning I need to just fill in lots of datapoints that match for the time between the steps? If so, that causes all the rate numbers etc. to be calculate incorrectly, as `promql` assumes that the datapoints you have are *real*.

> It's not that unstable, it's pretty safe to build on and many companies are doing so. 

From the documentation about [remote_read](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#%3Cremote_read%3E): `Remote read is experimental: breaking changes to configuration are likely in future releases.` I'm not sure how to read that as a stable supported API.

brian-brazil:
> I'd like some more specifics about what all problems are being ignored, but let that whole thread

The code doesn't consider that the Prometheus servers in a pair will have slightly different data, let alone the possibility that they have wildly different data due to events such as a network partition.

> There is already a concept in getting data thats "close" to the requested time (which is why StalenessDelta even exists). The problem here is that you can't control what your promql thinks is stale-- which seems especially crazy since you (as the binary using promql) have more insight into the data you are feeding it.

StalenessDelta is primarily a performance optimisation to avoid pull in too much data. If you're doing things in a normal fashion you shouldn't care what its value is as it'll have no effect semantically.

> How do you do this from the Querier interface? 

You'd adjust the query you are making to Prometheus.

> Well, this is actually an alignment problem. 

I don't follow. A different alignment will have the same issue, depending on the underlying data.

> From that level all I know is a labelset and a range of time, there is no mechanism for me to know whether it is a counter or a gauge.

There isn't, this is one of the big challenges of downsampling. The approach I propose is looking at the function around the selector, which remote read will pass on at some point in future.

> I'm not sure I follow how this makes things more convoluted?

For example you think this solves your problem, in reality it would lead to you implementing incorrect semantics causing further confusion and contortions as the new incorrect semantics were attempted to be worked around.

The gap you are seeing here is correct and desirable. Pondering it overnight you actually need to do a step further to ensure the user sees the gap and have consistent semantics given the downsampling algorithm that is in use. In particular when a time series disappears from one step to the next you should insert a stale marker.

> This is a somewhat flawed statement, especially since the scaling of prometheus is almost entirely based around rescraping and aggregation (which all "downsample" in their own way without access to the original data).

Not really, the recommended ways of doing things includes downsampling so as not to skip data when a higher scrape interval is used. So for example if you were scraping once an hour, you'd be pulling a `rate[1h]`.

> You mean to say that from downsampled data I need to present to promQL data that is not downsampled? That is by definition not possible (I could make up data, but thats not generally a good plan).

Yes, that's what needs to be done.

> So if I'm understanding your suggestion it is to return data as if I had datapoints every <whatever_the_scrape_interval_is> -- meaning I need to just fill in lots of datapoints that match for the time between the steps? If so, that causes all the rate numbers etc. to be calculate incorrectly, as promql assumes that the datapoints you have are real.

That's the basic approach I propose, if you look at other monitoring systems it's pretty normal to fill in gaps between downsampled data. For example for counters you'd want to make it increase at the right rate. This is only something that will be possible via remote read though, see #2580.

> I'm not sure how to read that as a stable supported API.

We're planning on changing the transport to GRPC, but the core seems solid. Note that no stability guarantees are offered on the various Prometheus internals you are pulling in, I'd worry more about those.

jacksontj:
> The code doesn't consider that the Prometheus servers in a pair will have slightly different data, let alone the possibility that they have wildly different data due to events such as a network partition.

This is actually covered in the merge logic in the proxyquerier, I'd be more than happy to explain that some other time-- but this is completely off-topic of this PR

> StalenessDelta is primarily a performance optimisation to avoid pull in too much data.

How is this a performance optimization to not *pull* so much data? `StalenessDelta` is applied after fetching the data from the datastore, it is used (according to the code and the [comments](https://github.com/prometheus/prometheus/blob/release-1.7/promql/engine.go#L1391)) to `[determine] the time since the last sample after which a time series is considered stale.` This is not a performance optimization, this is a mechanism to determine when a dataseries is "stale" meaning it has no data that would be valid for a calculation at that particular point in time.

> If you're doing things in a normal fashion you shouldn't care what its value is as it'll have no effect semantically.

This actually has *huge* effects semantically even in normal usage. In the previous example I gave, if you have a scrape interval of 1s with a `StalenessDelta` of 5m any datapoint can be used from a vector that is within 5m of the time that you are fetching. This means that the data result will vary based on the value, which means that there is a semantic implication of the `StalnessDelta` configuration.

> You'd adjust the query you are making to Prometheus.

I'll repeat the question `How do you do this from the Querier interface?`, if you are working at the `Querier` interface you don't have access to the query to rewrite it in any mechanism.

> The approach I propose is looking at the function around the selector, which remote read will pass on at some point in future.

So to be clear the suggestion you are giving me is to use non-existant functionality?

>  in reality it would lead to you implementing incorrect semantics causing further confusion and contortions as the new incorrect semantics were attempted to be worked around.

You keep mentioning that the proxy will cause more problems, but I'm not asking you to make or implement it. We are *really* off in the weeds talking about that proxy when this PR is *only* about making `StalenessDelta` more configurable-- can we get back on topic?

> Pondering it overnight you actually need to do a step further to ensure the user sees the gap and have consistent semantics given the downsampling algorithm that is in use. In particular when a time series disappears from one step to the next you should insert a stale marker.

Users do (and will) see gaps, the problem is right now I can't control *when* they'll see a gap with how much I downsample because of the hard-coded `StalenessDelta` -- if that was configurable I could scale it with the sample rate.

> Yes, that's what needs to be done.

So, what is the suggestion here? You have asked for an impossibility or made-up data.

> That's the basic approach I propose, if you look at other monitoring systems it's pretty normal to fill in gaps between downsampled data. For example for counters you'd want to make it increase at the right rate. This is only something that will be possible via remote read though, see #2580.

Well, this actually varies dramatically across monitoring systems in *how* they do this. If you look at RRD/Whisper files they can do this on the storage layer because they know if the metric is a counter or a gauge and as such can make a decision. This doesn't end up with "correct" data, but it ends up with less "close enough" data -- which is sufficient for their purposes. In prometheus there is no mechansim for me to determine from the promql statement whether a metric is a counter or a gauge, so I'm left having to ask all of the data. Ideally I'd ask for 100% of data, but if the time range is large, downsampling is required to not break downstreams. As to "fill in the gaps" this was a previous approach in proxmy, but you cannot "make up" datapoints to pass back up from the `Querier` interface, as promql assumes they are actual scrape instances. Meaning that if I return 2 matching datapoints it assumes that it didn't change during that duration. If I only have downsampled data I cannot make that assertion, so it'd be better for me to only pass the data I have instead of guessing what the data might be.

> We're planning on changing the transport to GRPC, but the core seems solid. Note that no stability guarantees are offered on the various Prometheus internals you are pulling in, I'd worry more about those.

I'm not sure what the objective of that point is? For the golang dependencies I have them pinned, so I'm not immediately concerned with them changing on me -- as they won't be a suprise. The APIs that I need to "rely" on are the HTTP v1 APIs from prometheus, since those are outside the project and how I actually fetch data.


Overall this entire PR conversation has drifted entirely from the original proposal to make `StalenessDelta` more configurable. I realize that this naive downsampling method can have some problems, but for a large section of use-cases it is sufficient. It is also significantly less complicated than a "more correct" approach-- which will come later (as that will probably require either (1) large changes to promql engine to add in hooks etc to rewrite or (2) a re-implementation of promql to handle creating eqivalent "cheaper" queries to downstream prometheus servers.) The question I have is does this PR break or make promql engine worse? Since its 100% compatible I don't see this as a problem.

brian-brazil:
> How is this a performance optimization to not pull so much data? StalenessDelta is applied after fetching the data from the datastore, it is used (according to the code and the comments) to [determine] the time since the last sample after which a time series is considered stale.

It's used to control how much data instant vector evaluation pulls and processes from local storage by the engine. It's not used after the data is fetched.

> So to be clear the suggestion you are giving me is to use non-existant functionality?

Yes, discussions around downsampling are at a relatively early stage so some of the required functionality hasn't been implemented yet. Even though I've been designing this for some time now, as it happens the discussions only really started yesterday.

> In the previous example I gave, if you have a scrape interval of 1s with a StalenessDelta of 5m any datapoint can be used from a vector that is within 5m of the time that you are fetching. This means that the data result will vary based on the value, which means that there is a semantic implication of the StalnessDelta configuration.

That's not how staleness works in Prometheus, see https://www.youtube.com/watch?v=GcTzd2CLH7I for more detail (though that's slightly outdated).

The way to handle this is by inserting staleness markers when a timeseries disappears from one step to the next. No need to change StalenessDelta.

> I'll repeat the question How do you do this from the Querier interface?, if you are working at the Querier interface you don't have access to the query to rewrite it in any mechanism.

I meant the queries being send out to /query_range.

> Users do (and will) see gaps, the problem is right now I can't control when they'll see a gap with how much I downsample because of the hard-coded StalenessDelta -- if that was configurable I could scale it with the sample rate.

With the current downsampling algorithm the gaps you are seeing are correct, in fact there are fewer gaps than there should be.

> We are really off in the weeds talking about that proxy when this PR is only about making StalenessDelta more configurable-- can we get back on topic?

The topic is implementation of downsampling, StalenessDelta doesn't really come into it. You should consider the StalenessDelta as a universal constant, and design things to work with rather than against it.

> So, what is the suggestion here? You have asked for an impossibility or made-up data.

Yes, make up data. If you request data from query_range that is more than 2m apart, then you need to do some interpolation.

> In prometheus there is no mechansim for me to determine from the promql statement whether a metric is a counter or a gauge, so I'm left having to ask all of the data.

If it's being used with `rate`, `irate`, `increase` or `resets` it's a counter. Otherwise it's a gauge. Making that and other useful information accessible to downsampling is what #2580 is about.

> As to "fill in the gaps" this was a previous approach in proxmy, but you cannot "make up" datapoints to pass back up from the Querier interface, as promql assumes they are actual scrape instances. 

I'm not following you here.

> If I only have downsampled data I cannot make that assertion, so it'd be better for me to only pass the data I have instead of guessing what the data might be.

Downsampling is the art of making it seem as though you have full data when you don't. At it stands the code is already futzing with the data, as using query_range loses the timestamp of the original data.

> I realize that this naive downsampling method can have some problems, but for a large section of use-cases it is sufficient. 

Yes it's the naive way of doing things using an interface that isn't suited for the task. I understand what you are trying to do, and I'm trying to help you implement it in the best way possible.

> The question I have is does this PR break or make promql engine worse? Since its 100% compatible I don't see this as a problem.

In my opinion yes it makes the engine worse, as it'll lead to more support requests by users who mistakenly think that changing staleness delta helps them with whatever problem they're having. We've had quite a problem with this for this particular setting historically. Every knob and tunable we have results in increased support load and user confusion. For example we recently had to hide some settings intended for our own loadtesting/development as users were starting to get confused by them.

If someone were to request this for something that made sense semantically, then I'd consider it.

juliusv:
@jacksontj If I understand correclty (I haven't read all the code), promxy executes PromQL via `/query_range` against backend Prometheus servers, and then runs PromQL once again on the returned data?

Unfortunately I would have to agree with @brian-brazil that this is not a use case we would want to support, since it is fraught with all the problems mentioned above, when not computing PromQL on raw data (configurable staleness delta or not).

Maybe there will be full distributed query evaluation at some point in Prometheus that will do all this correctly, but I believe that's quite far off.

jacksontj:
@juliusv promxy optionally switches between query and query_range. I agree that ideally we'd all be able to use raw data always, but sometimes thats not feasible (slow serialization, limited memory, etc).

Regardless of people's opinions on promql against query_range, this PR isn't about promql over query_range, its about being able to control the `StalenessDelta` with *some* flexibility when using promql as a library. As I had mentioned earlier this is just as useful for the downsampling case as it is for cases with very tight scrape intervals (as mentioned previously, 5m staleness for a 1s scrape is excessive).

brian-brazil:
> That still doesn't address the lack of granular configuration for the StalenessDelta but hopefully we can completely separate out these two issues.

If someone presents us with a good use case for it we can look at it then, but that has yet to occur.

jacksontj:
> If someone presents us with a good use case for it we can look at it then, but that has yet to occur.

Wouldn't wanting to change it based on the scrape interval of the remote storage node be a "good use case"? If not, what constitutes a "good use case"?

brian-brazil:
> Wouldn't wanting to change the StalenessDelta based on the scrape interval of the remote storage node be a "good use case"? 

No, that's a bad setup. Data in a remote storage node should have the same interval as the the Prometheus, and even then that's something that does not require changing it on a per-query basis.

> If not, what constitutes a "good use case"?

One that comes from following the recommended ways to use Prometheus and makes semantic sense.

jacksontj:
>  Data in a remote storage node should have the same interval as the the Prometheus

I think the point here is being missed. If I make something that uses promql and fetches data from *multiple* prometheus hosts based on some criteria, depending on *which* prometheus server I send the request to I would want to modify the `StalenessDelta`.

> recommended ways to use Prometheus

Meaning people using the promql library directly are by definition excluded?

brian-brazil:
> Meaning people using the promql library directly are by definition excluded?

Only users using promql directly have the potential to have a valid use for this. Fishing for a valid use case to justify an invalid use case is not a winning strategy.

It has been explained how to handle this properly in your code, I'd suggest taking this advice rather than continuing to follow red herrings.

jacksontj:
> Only users using promql directly have the potential to have a valid use for this.

This is *literally* my use-case, I am using promql directly -- not inside prometheus.

juliusv:
> > Only users using promql directly have the potential to have a valid use for this.
>
> This is *literally* my use-case, I am using promql directly -- not inside prometheus.

I think what @brian-brazil meant is that there is no use case beyond that, and the use case laid out here using the library directly runs into all kinds of problems (and is very arcane besides that) that make the use case not valid enough.

I was briefly considering that if we at some point add general support for per-query options (that would be exposed to the end-user via the web API), it would be somewhat more palatable to coincidentally also have the staleness delta as one of several options in there. But I do tend to agree with @brian-brazil that this should normally be a server-wide setting and we have not yet seen a valid use case that justifies supporting per-query changes of this.

jacksontj:
> using the library directly runs into all kinds of problems

What sorts of problems? Currently using promql directly and so far its been working quite well.

> and is very arcane besides that

Not sure what you mean by arcane? Its not really old, nor is it magical.

The ask here is not necessarily per-query, its per datasource (since this is a proxy of many prometheus hosts). Because promql is just a library IMO it doesn't make sense for it to know too much about what data came from where, so changing it on a per-query basis seems to make more sense. If we'd rather make promql's package keep track of the scrape interval of the data we could do that-- but that seems like the wrong approach.

jacksontj:
Another way of stating it (which is hopefully clearer)-- the `StalenessDelta` should be tied to the data that you are processing (this is because you want to set it based on granularity of data etc.). In the current state (package-level variable) any program that wants to use the `promql` package can't use data from sources with different granularity. To solve this I believe there are 2 options (1) per-statement or (2) per-engine. This PR is an implenentation of 1 (since that allows greater flexibility) but if we want to take the second approach thats also fine with me (as I've stated before). Having a single package-wide variable severely limits the usefulness of the `promql` package

brian-brazil:
I understand what you are saying, however this is not a use case we feel makes sense to support. There are better ways to approach what you're trying to do, and I following up on the suggestions above.

This discussion is going around in circles, so I'm close it off now.

PR #3630, opened 2017-12-27
Add Healthcheck to Dockerfile

kulinacs:
Adds an internal healthcheck using the `/-/healthy` input to see
if the container is healthy. 

I have tested this with a healthy container and it correctly reports the container as healthy. However, I am not sure how to force Prometheus to report itself as unhealthy and as such, that still needs to be tested.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,7 @@ RUN ln -s /usr/share/prometheus/console_libraries /usr/share/prometheus/consoles
     ln -s /prometheus /etc/prometheus/data
 
 USER       nobody
+HEALTHCHECK CMD wget -q --spider http://localhost:9090/-/healthy || exit 1
 EXPOSE     9090
 VOLUME     [ "/prometheus" ]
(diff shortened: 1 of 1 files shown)


kulinacs:
Looks like CircleCI's version of Docker is too old to support the Healthcheck directive.

brian-brazil:
> I am not sure how to force Prometheus to report itself as unhealthy and as such, that still needs to be tested.

`SIGSTOP` is useful for this sort of thing.

kulinacs:
After testing with `SIGSTOP`, the container does show as unhealthy.

codesome:
@kulinacs could you refactor this to pass all the tests?

kulinacs:
@codesome done

gouthamve:
Hi thanks for submitting the PR. We've taken a look at this and this looks like something that is docker compose specific. We don't think this should be part of the dockerfile but rather the docker compose specification.

PR #3631, opened 2017-12-27
Add engine option for NodeReplacer

jacksontj:
NodeReplacer is a mechanism for clients of the promql library to inject
themselves into the processing of the node tree. This is especially
helpful for (1) introspecting queries and (2) mutating queries.

This is required for non-naive downsampling using promql. For my use-case in [promxy](https://github.com/jacksontj/promxy) this allows for me to be smarter about downsampling.

**current state** 
In the current state the only interface to implement is the `Querier` interface which is supposed to fetch raw data. This means you end up pulling significantly more data off a box than necessary, but the results are always correct. The only way to "downsample" at this API layer is to change from a `query` to `query_range` which has a variety of naive downsampling problems -- but solves the scale problem.

**new state, with this PR**
With this new NodeReplacer API the creator of the engine can define a function which will be called with each node in the promql tree. This allows us to mutate the query however we like. To better describe this, lets use a concrete example.

Promql: `sum(irate(http_request_duration_microseconds[1m]))`

This (effectively) get parsed into an `AggregateExpr` -> `Call` -> `VectorSelector`. If we where going soely through the `Querier` API we'd have to return all datapoints for timeseries that match `http_request_duration_microseconds`. With the `NodeReplacer` API we can instead ask the datastore for `irate(http_request_duration_microseconds[1m])` and sum those together. This is done by replacing the `AggregateExpr.Expr` with the fetched data results in a `VectorSelector`.

Ideally we'd add an implementation of `Node` which was "here's some data" (maybe we can call it a `StaticSelector`) which would make the debug printing work.

### DIFF
--- a/promql/ast.go
+++ b/promql/ast.go
@@ -139,6 +139,10 @@ type MatrixSelector struct {
 	iterators []local.SeriesIterator
 }
 
+func (m *MatrixSelector) SetIterators(its []local.SeriesIterator) {
+	m.iterators = its
+}
+
 // NumberLiteral represents a number.
 type NumberLiteral struct {
 	Val model.SampleValue
@@ -170,6 +174,10 @@ type VectorSelector struct {
 
 	// The series iterators are populated at query preparation time.
 	iterators []local.SeriesIterator
+}
+
+func (v *VectorSelector) SetIterators(its []local.SeriesIterator) {
+	v.iterators = its
 }
 
 func (e *AggregateExpr) Type() model.ValueType  { return model.ValVector }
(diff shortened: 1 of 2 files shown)


brian-brazil:
> This is required for non-naive downsampling using promql.

As I've previously indicated, #2580 is what you're looking for. 

When is comes to downsampling the thing evaluating the selectors (LTS) and the thing evaluating PromQL (Prometheus) are separate concerns, and the thing evaluating PromQL should have no knowledge that downsampling is going on.

jacksontj:
I think the point here is being missed, the point here is this gives you a mechanism to rewrite the promql before execution. The primary driver for downsampling in a remote promql parser is that the marshaling is so slow (and the data on the wire can easily get large). With this change instead of asking for all the raw data we can ask the remote prometheus endpoint to calculate the data on their side, to be aggregated remotely. In the defined example this would mean that instead of every downstream having to send all datapoints in the window for all timeseries matching `http_request_duration_microseconds` they can simply return `irate(http_request_duration_microseconds[1m])` which is itself a summarized value -- meaning that we don't need to downsample, we just allow the remote end to return summarized data that gets an equivalent result.

brian-brazil:
> With this change instead of asking for all the raw data we can ask the remote prometheus endpoint to calculate the data on their side, to be aggregated remotely. 

That's not how we plan to do for downsampling, and query push down is not on our current roadmap.

The standing consensus within the project is that #2580 is what we want to do, so I'd suggest directing your energies there.

jacksontj:
This "query hints" approach means that you can *only* downsample, this approach allows us to *not* downsample, but reduce the bytes on the wire between the metrics store and the promql evaluator. From my example with this PR we could send something like `irate(http_request_duration_microseconds[1m])` to the datastores and sum them together. This is **not** downsampling, this is just reducing the number of bytes on the wire. 

Actual downsampling would still need these hints, so they would know to mutate their query downstream from `http_request_duration_microseconds` to some other query based on your proposed query hints.

I don't see these 2 as at odds with eachother I actually see them as complementary. With *only* query hints you can *only* downsample and with this PR you can decide to *not* downsample, and instead reduce bytes on the wire between storage and evaluator hosts.

brian-brazil:
I understand what you are proposing, however as has already been start query pushdown is not a feature that we plan on offering anytime soon (if ever). I'm afraid that I do not have the time to discuss the intricacies of downsampling, pushdown and other distributed query strategies further as such discussions do not help us implement what is planned.

PR #3533, opened 2017-12-01
Update Graphite remote example for Graphite v1.1.x

jtlisi:
This pull request updates the graphite remote write example to take advantage of Graphite's tagging support. It includes an extension to allow for remote-reads.

If tag support is not enabled there is no change in functionality.

If tag support is enabled, series will be written to graphite using the 1.1.x tagging support, and if a graphite web address is specified then remote read functionality will be enabled

https://graphite.readthedocs.io/en/latest/tags.html

PR #3635, opened 2017-12-29
remote_storage_adapter - reuse connection to graphite

catac:
Hello,

Currently, the graphite remote_storage_adapter creates a new connection each time it sends data to graphite. This can be an issue for logstash, that is not able to cope with so many connections and crashes.

This patch tries to reuse the connection, if exists already. Otherwise, it creates it. Since the send can be initiated by a remote call to the adapter (that can be in parallel with the current one), the connection object is protected with a lock.

Another small addition is a flag to allow configuring logging verbosity, as I've seen lots of metrics being logged that they are not sent when they have a NaN value.

If you want to integrate PR #3533 first, it's no problem - I can redo the changes here - just let me know.

Thanks!

### DIFF
--- a/documentation/examples/remote_storage/remote_storage_adapter/graphite/client.go
+++ b/documentation/examples/remote_storage/remote_storage_adapter/graphite/client.go
@@ -19,6 +19,7 @@ import (
 	"math"
 	"net"
 	"sort"
+	"sync"
 	"time"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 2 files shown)


gouthamve:
Thanks for this PR! We've included the storage adapters as examples and references to implement your own adapters but not something production worthy. We recommend using https://github.com/criteo/graphite-remote-adapter/ in production.

PR #3638, opened 2017-12-30
reimplement the Notifier discovery

### LINKED ISSUE #3637: alertmanagers static_configs no longer working on HEAD
**What did you do?**
Start Prometheus (587dec9e) with the example config (with the example alertmanager target enabled).

Status page (http://localhost:9090/status) should show one configured Alertmanager.

Status page does not show any Alertmanagers.

**Environment**

* System information: Linux 4.14.8-1-ARCH x86_64
* Prometheus version:
```
prometheus, version 2.0.0 (branch: HEAD, revision: 587dec9eb970531cddc7f1803d258e72129b5aa0)
  build user:       christian@wuechoo
  build date:       20171229-22:55:19
  go version:       go1.9.2
```

```yaml
# my global config
global:

# Alertmanager configuration
alerting:
  alertmanagers:
  - static_configs:
    - targets:
       - alertmanager:9093

rule_files:
  # - "first_rules.yml"
  # - "second_rules.yml"

scrape_configs:
  - job_name: 'prometheus'

    # scheme defaults to 'http'.

    static_configs:
      - targets: ['localhost:9090']
```

file_sd_configs seems to yield the same result. This may indicate a general issue with alertmanager target discovery in master. I have not tried other discovery methods though. Maybe I am misunderstand something or I am supposed to use different configs now, but for now, this looks like a bug to me as it unexpectedly changes previous behaviour.

I tried bisect'ing this issue but could not identify a single commit due to build issues (`retrieval/manager.go:95:11: undefined: TargetManager`).

The last working and building commit is a8cce41882e5d94f876d93c1b7c29d37a32f58eb.
The first non-working and building commit is 587dec9eb970531cddc7f1803d258e72129b5aa0

```bash
$ git bisect log
# only skipped commits left to test
# possible first bad commit: [587dec9eb970531cddc7f1803d258e72129b5aa0] rebased and resolved conflicts with the new Discovery GUI page
# possible first bad commit: [60ef2016d508545f0e8b575a954899cae37a96e2] add a cancel func to the scrape pool as it is needed in the scrape loop select block
# possible first bad commit: [80182a5d82d0f29fe8ec483f2ec1ae6d10f9f1de] use poolKey as the pool map key to avoid multi dimensional maps
# possible first bad commit: [1ec76d1950862c1cd5263065616212a97670449d] rearange the contexts variables and logic split the groupsMerge function to set and get other small nits
# possible first bad commit: [6ff1d5c51e3cb0e0df58aa90a473317e25c6d4ac] add the scrape manager config reloader handle errors with invalid scrape config
# possible first bad commit: [b0d4f6ee08632e09d0b0cc8ec0b2a0cbf61b5861] resolved merge confilc in main.go
# possible first bad commit: [f2df712166ef4e16e69d8aace76202619483c0f8] updated README
# possible first bad commit: [aca8f85699211c2453f0121a4015975c605fc2dd] fixed the tests
# possible first bad commit: [fe6c544532360a09c3cbdc6d235d6798b366263d] some renaming and comments fixes. remove some select state that is most likely obsoleete and hoepfully doesn't braje anything :) merge targets will sort by Discoverer name so we can have consistent tests for the maps.
# possible first bad commit: [f5c2c5ff8fc8e21f9bfaa23d0f81c8245ba8632e] brake the start provider func so that can run unit tests against it.
# possible first bad commit: [c5cb0d2910692cfd77b36d7d270677f58f55d3f9] simplify naming and API.
# possible first bad commit: [9c61f0e8a0c1d4f83d06e003af610a7d671287b2] scrape pool doesn't rely on context as Stop() needs to be blocking to prevent Scrape loops trying to write to a closed TSDB storage.
# possible first bad commit: [e405e2f1ea2156f76a7e70f5471b88a590a8e24a] refactored discovery

```

According to the commits, this may be related to pull request #3362 (@krasi-georgiev).

krasi-georgiev:
fixes #3637

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -16,6 +16,8 @@ package main
 
 import (
 	"context"
+	"crypto/md5"
+	"encoding/json"
 	"fmt"
 	"net"
 	"net/http"
@@ -295,9 +297,12 @@ func main() {
 		func(cfg *config.Config) error {
 			c := make(map[string]sd_config.ServiceDiscoveryConfig)
 			for _, v := range cfg.AlertingConfig.AlertmanagerConfigs {
-				// AlertmanagerConfigs doesn't hold an unique identifier so we use the config pointer as the identifier.
-				// TODO Krasi - Maybe use the slice index as the reference.
-				c[fmt.Sprintf("%p", v)] = v.ServiceDiscoveryConfig
+				// AlertmanagerConfigs doesn't hold an unique identifier so we use the config hash as the identifier.
+				b, err := json.Marshal(v)
+				if err != nil {
+					return err
+				}
+				c[fmt.Sprintf("%x", md5.Sum(b))] = v.ServiceDiscoveryConfig
 			}
 			return discoveryManagerNotify.ApplyConfig(c)
 		},
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -296,6 +296,7 @@ func main() {
 			c := make(map[string]sd_config.ServiceDiscoveryConfig)
 			for _, v := range cfg.AlertingConfig.AlertmanagerConfigs {
 				// AlertmanagerConfigs doesn't hold an unique identifier so we use the config pointer as the identifier.
+				// TODO Krasi - Maybe use the slice index as the reference.
 				c[fmt.Sprintf("%p", v)] = v.ServiceDiscoveryConfig
 			}
 			return discoveryManagerNotify.ApplyConfig(c)
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -44,6 +44,7 @@ import (
 	promlogflag "github.com/prometheus/common/promlog/flag"
 	"github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/discovery"
+	sd_config "github.com/prometheus/prometheus/discovery/config"
 	"github.com/prometheus/prometheus/notifier"
 	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/retrieval"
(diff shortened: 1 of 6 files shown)


cmd/prometheus/main.go
```diff
@@ -283,7 +285,21 @@ func main() {
...
+			for _, v := range cfg.ScrapeConfigs {
+				c[v.JobName] = v.ServiceDiscoveryConfig
+			}
+			return discoveryManagerScrape.ApplyConfig(c)
+		},
+		func(cfg *config.Config) error {
+			c := make(map[string]sd_config.ServiceDiscoveryConfig)
+			for _, v := range cfg.AlertingConfig.AlertmanagerConfigs {
+				// AlertmanagerConfigs doesn't hold an unique identifier so we use the config pointer as the identifier.
+				c[fmt.Sprintf("%p", v)] = v.ServiceDiscoveryConfig
```
Gouthamve: Can this be the index of the element in the array?

notifier/notifier.go
```diff
@@ -257,7 +256,8 @@ func (n *Notifier) ApplyConfig(conf *config.Config) error {
 
 		ams.metrics = n.metrics
 
-		amSets = append(amSets, ams)
+		// The config pointer is used for the map lookup identifier.
+		amSets[fmt.Sprintf("%p", cfg)] = ams
```
Gouthamve: See here too.

retrieval/manager.go
```diff
@@ -127,11 +125,12 @@ func (m *ScrapeManager) Targets() []*Target {
 	return <-targets
 }
 
-func (m *ScrapeManager) reload(t map[string][]*targetgroup.Group) error {
+func (m *ScrapeManager) reload(t map[string][]*targetgroup.Group) {
 	for tsetName, tgroup := range t {
 		scrapeConfig, ok := m.scrapeConfigs[tsetName]
 		if !ok {
-			return fmt.Errorf("target set '%v' doesn't have valid config", tsetName)
+			level.Error(m.logger).Log("msg", "error reloading target set", "err", fmt.Sprintf("invalid config id:%v", tsetName))
```
Gouthamve: Why is this check here? I am not able to understand why this is triggered.

cmd/prometheus/main.go
krasi-georgiev: I looked it over and don't think it would be unique enough. There is a chance that when receiving an update from the Discovery to read the wrong config for the target set.

I know it is ugly, but I think less likely to get into a bad sync bug between the configs and the target sets sent from the Discovery manager.

retrieval/manager.go
krasi-georgiev: a fail safe if the target set doesn't hold a valid config for some reason.

I will probably refactor this in one of the next PRs.

cmd/prometheus/main.go
```diff
@@ -283,7 +285,22 @@ func main() {
...
+				c[v.JobName] = v.ServiceDiscoveryConfig
+			}
+			return discoveryManagerScrape.ApplyConfig(c)
+		},
+		func(cfg *config.Config) error {
+			c := make(map[string]sd_config.ServiceDiscoveryConfig)
+			for _, v := range cfg.AlertingConfig.AlertmanagerConfigs {
+				// AlertmanagerConfigs doesn't hold an unique identifier so we use the config pointer as the identifier.
+				// TODO Krasi - Maybe use the slice index as the reference.
+				c[fmt.Sprintf("%p", v)] = v.ServiceDiscoveryConfig
```
brancz: Maybe a slight overkill, but how about "<index>-<hash>"? The index would really only be there for debug-ability. The hash could be built by marshaling the `v.ServiceDiscoveryConfig` and md5 hashing it.

PR #3588, opened 2017-12-14
Added dropped targets to targets API endpoint

Conorbro:
Added the targets dropped during relabelling to the targets API endpoint (v1) following #3544.

PR #3646, opened 2018-01-02
Added dropped targets API

Conorbro:
A continuation from #3588  following resolving of merge conflicts.


krasi-georgiev:
@Conorbro , @brian-brazil  do we have a use case for this?

Conorbro:
I guess it could be handy for debugging service discovery.
Also probably best to be consistent with the rest of the API which will soon allow for dropped alertmanagers.

krasi-georgiev:
I don't mind, just wondering if we should add something that is not in immediate need.

PR #3660, opened 2018-01-05
Fix Kubernetes endpoints SD for empty subsets

simonpasquier:


discovery/kubernetes/endpoints.go
```diff
@@ -185,13 +185,14 @@ const (
...
 
 func (e *Endpoints) buildEndpoints(eps *apiv1.Endpoints) *targetgroup.Group {
-	if len(eps.Subsets) == 0 {
-		return nil
-	}
-
 	tg := &targetgroup.Group{
 		Source: endpointsSource(eps),
 	}
+	if len(eps.Subsets) == 0 {
```
brian-brazil: Do you even need this if?

simonpasquier: Good point, probably not...

simonpasquier:
@brian-brazil PTAL

brian-brazil:
Thanks!

PR #3661, opened 2018-01-05
documentation: Adapt Kubernetes RBAC to use metrics roles

brancz:
Recommending proxying metrics requests to the Kubelet through the apiserver seems odd and cetainly doesn't scale. Instead we should be recommending to use RBAC on the kubelet as well.

@fabxc @grobie @tomwilkie

### DIFF
--- a/documentation/examples/prometheus-kubernetes.yml
+++ b/documentation/examples/prometheus-kubernetes.yml
@@ -70,6 +70,13 @@ scrape_configs:
   # <kubernetes_sd_config>.
   tls_config:
     ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+    # If your node certificates are self-signed or use a different CA to the
+    # master CA, then disable certificate verification below. Note that
+    # certificate verification is an integral part of a secure infrastructure
+    # so this should only be disabled in a controlled environment. You can
+    # disable certificate verification by uncommenting the line below.
+    #
+    # insecure_skip_verify: true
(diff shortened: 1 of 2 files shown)


brancz:
Note that the previously used `nodes/proxy` role allows performing *any* action on all APIs of the kubelets, effectively giving Prometheus root on all nodes available.

tomwilkie:
Sorry for delay @brancz, just trying this out on our GKE cluster now.

grobie:
LGTM, but our k8s version is a bit older *cough*. @matthiasr any comments from your side?

tomwilkie:
Results from trying it on GKE: their kubelets used self-signed certs, so this doesn't work.

I think we should make the standard configuration work for most people (ie proxying through the master), and have the commented out / advanced configuration for direct, secure access.  Its suboptimal, for sure, but working for more people out the box is more desirable IMO.

juliusv:
One issue I have with this on a `kubeadm`-generated 1-node test cluster is that scraping `/metrics` or `/metrics/cadvisor` on the nodes fails because the Kubelet's cert does not contain any IP SANs (it works with `insecure_skip_verify: true` of course). Maybe we should put the hostname (from a meta label) into the `__address__` instead? (Not that that would work in my test setup, since that hostname isn't resolvable from the Prometheus container)

beorn7:
Is this still relevant?

brancz:
Yes the changes here are still accurate.

gouthamve:
@simonpasquier / @brancz Could you tag someone who can review this PR and get it merged?

PR #3663, opened 2018-01-08
Add query to retrieve number of samples per second

dominikschulz:
acording to https://groups.google.com/d/msg/prometheus-users/u-dBte5pafs/TfeAuR5BAwAJ

Do you think it would make sense to add this to the docs? Would be helpful IMHO.


brian-brazil:
Without the rest of the equation, this is kinda out of place. A section on the metrics the tsdb provides might work.

dominikschulz:
Ok, I don't think I can do that right now. If I get some time I'll just raise another PR.
Thank you!

PR #3669, opened 2018-01-09
Fix the incrementing of prometheus_target_scrapes_exceeded_sample_limit_total

### LINKED ISSUE #3668: `prometheus_target_scrapes_exceeded_sample_limit_total` not incremented
**What did you do?**
When a target goes over the configured `sample_limit`, I looked at the `prometheus_target_scrapes_exceeded_sample_limit_total` metric.

I expected this to increase for each scrape which hit the sample limit, as per the help on the metric:

> Total number of scrapes that hit the sample limit and were rejected.

(from [scrape.go#L83](https://github.com/prometheus/prometheus/blob/97464236c76301dab933ab8407f7ebe0ee1ec292/retrieval/scrape.go#L83))

This metric was not incremented.

**Environment**

* System information:

	Linux 4.4.0-103-generic x86_64

* Prometheus version:

Seen on the `prom/prometheus:v2.0.0` Docker image, reproduced with local build from master branch:

    prometheus, version 2.0.0 (branch: master, revision: 97464236c76301dab933ab8407f7ebe0ee1ec292)

* Alertmanager version:
NA

```
global:
  scrape_interval:     15s
  evaluation_interval: 15s

scrape_configs:
  - job_name: 'some_job'

    static_configs:
    - targets: ['localhost:9091']

    sample_limit: 100
```

NA

* Logs:
```
level=warn ts=2018-01-09T11:36:18.344766728Z caller=scrape.go:683 component="scrape manager" scrape_pool=some_job target=http://localhost:9091/metrics msg="append failed" err="sample limit exceeded"
```

**Notes**
It appears to be the case that at [scrape.go#L907](https://github.com/prometheus/prometheus/blob/97464236c76301dab933ab8407f7ebe0ee1ec292/retrieval/scrape.go#L907) the metric will be incremented if `err == nil && sampleLimitErr != nil`.
However, if I am understanding correctly, `err` will always be the same as `sampleLimitErr` as once the limit has been hit all further samples will also result in that error. Though there is the possibility that a different error is raised at the final stage of the loop.

My guess would be that [scrape.go#L907](https://github.com/prometheus/prometheus/blob/97464236c76301dab933ab8407f7ebe0ee1ec292/retrieval/scrape.go#L907) should be changed to just `sampleLimitErr != nil` , but I'm not certain how the line `err = sampleLimitErr` should be changed. If I am understanding things correctly, `err` would always already be  `sampleLimitErr`, so that line is unneeded.

ant1441:
Once a scrape target has reached the samle_limit, any further samples
from that target will continue to result in errSampleLimit.

This means that after the completion of the append loop `err` must still
be errSampleLimit, and so the metric would not have been incremented.

Fixes #3668 (@brian-brazil)

### DIFF
--- a/retrieval/scrape.go
+++ b/retrieval/scrape.go
@@ -905,10 +905,9 @@ loop:
 		err = p.Err()
 	}
 	if sampleLimitErr != nil {
-		// We only want to increment this once per scrape, so this is Inc'd outside the loop
+		// We only want to increment this once per scrape, so this is Inc'd outside the loop.
 		targetScrapeSampleLimit.Inc()
 	}
-
 	if numOutOfOrder > 0 {
 		level.Warn(sl.l).Log("msg", "Error on ingesting out-of-order samples", "num_dropped", numOutOfOrder)
 	}
--- a/retrieval/scrape.go
+++ b/retrieval/scrape.go
@@ -904,10 +904,11 @@ loop:
 	if err == nil {
 		err = p.Err()
 	}
-	if err == nil && sampleLimitErr != nil {
+	if sampleLimitErr != nil {
+		// We only want to increment this once per scrape, so this is Inc'd outside the loop
 		targetScrapeSampleLimit.Inc()
-		err = sampleLimitErr
 	}
+
 	if numOutOfOrder > 0 {
 		level.Warn(sl.l).Log("msg", "Error on ingesting out-of-order samples", "num_dropped", numOutOfOrder)
 	}
--- a/retrieval/scrape_test.go
+++ b/retrieval/scrape_test.go
@@ -32,6 +32,8 @@ import (
 	"github.com/prometheus/common/model"
 	"github.com/stretchr/testify/require"
 
+	dto "github.com/prometheus/client_model/go"
+
 	"github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/pkg/timestamp"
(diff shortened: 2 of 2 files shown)


retrieval/scrape.go
```diff
@@ -904,10 +904,11 @@ loop:
 	if err == nil {
 		err = p.Err()
 	}
-	if err == nil && sampleLimitErr != nil {
+	if sampleLimitErr != nil {
+		// We only want to increment this once per scrape, so this is Inc'd outside the loop
```
brian-brazil: Full stop at the end of a comment.

retrieval/scrape.go
```diff
@@ -904,10 +904,11 @@ loop:
...
 		targetScrapeSampleLimit.Inc()
-		err = sampleLimitErr
 	}
+
```
brian-brazil: Why the newline?

retrieval/scrape_test.go
```diff
@@ -627,6 +629,49 @@ func TestScrapeLoopAppend(t *testing.T) {
...
+	// Check that the Counter has been incremented a simgle time for the scrape,
+	// not multiple times for each sample
+	metric := dto.Metric{}
+	err = targetScrapeSampleLimit.Write(&metric)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	value := metric.GetCounter().GetValue()
+	if value != 1 {
```
brian-brazil: This tests that its value is 1, depending on test order that may not be the case. Test that is increased by 1.

ant1441:
Thanks for the comments, good point on the testing the Counter, they should all be resolved now.

brian-brazil:
Thanks!

PR #3671, opened 2018-01-09
*: implement query params

fabxc:
This adds a parameter to the storage selection interface which allows
query engine(s) to pass information about the operations surrounding a
data selection.
This can for example be used by remote storage backends to infer the
correct downsampling aggregates that need to be provided.

@brian-brazil  @Bplotka

### DIFF
--- a/promql/ast.go
+++ b/promql/ast.go
@@ -238,56 +238,58 @@ type VectorMatching struct {
 }
 
 // Visitor allows visiting a Node and its child nodes. The Visit method is
-// invoked for each node encountered by Walk. If the result visitor w is not
-// nil, Walk visits each of the children of node with the visitor w, followed
-// by a call of w.Visit(nil).
+// invoked for each node with the path leading to the node provided additionally.
+// If the result visitor w is not nil, Walk visits each of the children
+// of node with the visitor w, followed by a call of w.Visit(nil, nil).
 type Visitor interface {
-	Visit(node Node) (w Visitor)
+	Visit(node Node, path []Node) (w Visitor)
 }
 
 // Walk traverses an AST in depth-first order: It starts by calling
-// v.Visit(node); node must not be nil. If the visitor w returned by
-// v.Visit(node) is not nil, Walk is invoked recursively with visitor
+// v.Visit(node, path); node must not be nil. If the visitor w returned by
+// v.Visit(node, path) is not nil, Walk is invoked recursively with visitor
 // w for each of the non-nil children of node, followed by a call of
 // w.Visit(nil).
-func Walk(v Visitor, node Node) {
-	if v = v.Visit(node); v == nil {
+// As the tree is descended the path of previous nodes is provided.
+func Walk(v Visitor, node Node, path []Node) {
+	if v = v.Visit(node, path); v == nil {
 		return
 	}
+	path = append(path, node)
 
 	switch n := node.(type) {
 	case Statements:
 		for _, s := range n {
-			Walk(v, s)
+			Walk(v, s, path)
 		}
 	case *AlertStmt:
-		Walk(v, n.Expr)
+		Walk(v, n.Expr, path)
 
 	case *EvalStmt:
-		Walk(v, n.Expr)
+		Walk(v, n.Expr, path)
 
 	case *RecordStmt:
-		Walk(v, n.Expr)
+		Walk(v, n.Expr, path)
 
 	case Expressions:
 		for _, e := range n {
-			Walk(v, e)
+			Walk(v, e, path)
 		}
 	case *AggregateExpr:
-		Walk(v, n.Expr)
+		Walk(v, n.Expr, path)
 
 	case *BinaryExpr:
-		Walk(v, n.LHS)
-		Walk(v, n.RHS)
+		Walk(v, n.LHS, path)
+		Walk(v, n.RHS, path)
 
 	case *Call:
-		Walk(v, n.Args)
+		Walk(v, n.Args, path)
 
 	case *ParenExpr:
-		Walk(v, n.Expr)
+		Walk(v, n.Expr, path)
 
 	case *UnaryExpr:
-		Walk(v, n.Expr)
+		Walk(v, n.Expr, path)
 
 	case *MatrixSelector, *NumberLiteral, *StringLiteral, *VectorSelector:
 		// nothing to do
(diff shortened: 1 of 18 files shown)


promql/ast.go
```diff
@@ -242,52 +242,53 @@ type VectorMatching struct {
 // nil, Walk visits each of the children of node with the visitor w, followed
 // by a call of w.Visit(nil).
 type Visitor interface {
-	Visit(node Node) (w Visitor)
+	Visit(node Node, path []Node) (w Visitor)
```
Bplotka: Can you update comment here and for `Walk` method?

promql/ast.go
```diff
@@ -311,6 +312,6 @@ func (f inspector) Visit(node Node) Visitor {
 // Inspect traverses an AST in depth-first order: It starts by calling
 // f(node); node must not be nil. If f returns true, Inspect invokes f
 // for all the non-nil children of node, recursively.
-func Inspect(node Node, f func(Node) bool) {
-	Walk(inspector(f), node)
+func Inspect(node Node, f func(Node, []Node) bool) {
```
Bplotka: ditto (comment to be updated)

promql/engine.go
```diff
@@ -151,25 +153,19 @@ func contextDone(ctx context.Context, env string) error {
 // It is connected to a querier.
 type Engine struct {
 	// A Querier constructor against an underlying storage.
```
Bplotka: Move that comment to the same place where you moved queryable (:

promql/engine.go
```diff
@@ -514,15 +493,21 @@ func (ng *Engine) populateIterators(ctx context.Context, s *EvalStmt) (storage.Q
...
 	}
 
-	Inspect(s.Expr, func(node Node) bool {
+	Inspect(s.Expr, func(node Node, path []Node) bool {
+		params := &storage.SelectParams{
+			Step: int64(s.Interval / time.Millisecond),
+		}
+
 		switch n := node.(type) {
 		case *VectorSelector:
```
Bplotka: Can we reuse some code here? The code for Vector and Matrix is totally the same except:
`durationMilliseconds(LookbackDelta)` vs `durationMilliseconds(n.Range)`

storage/interface.go
```diff
@@ -61,6 +61,12 @@ type Querier interface {
 	Close() error
 }
 
+// SelectParams specifies parameters passed to data selections.
+type SelectParams struct {
+	Step int64
```
brian-brazil: Is this seconds or milliseconds? Comments would be good.

promql/engine.go
```diff
@@ -559,6 +546,21 @@ func (ng *Engine) populateIterators(ctx context.Context, s *EvalStmt) (storage.Q
 	return querier, err
 }
 
+// extractFuncFromPath walks up the path and searches for the first instance of
```
brian-brazil: I think we should only look back one node. For example `max(1-metric)` might in fact be a `min` in practice, so a hint of `max` would incorrect.

storage/remote/read.go
```diff
@@ -43,7 +43,7 @@ type querier struct {
 
 // Select implements storage.Querier and uses the given matchers to read series
 // sets from the Client.
-func (q *querier) Select(matchers ...*labels.Matcher) (storage.SeriesSet, error) {
+func (q *querier) Select(_ *storage.SelectParams, matchers ...*labels.Matcher) (storage.SeriesSet, error) {
```
brian-brazil: I think the params should be passed down to the proto we send for a read

promql/engine.go
fabxc: Mh, maybe early exit on binary expressions? Because just looking back one node won't work for `max_over_time(a[5m])` or `count((a))` as the range selector or parens are their own nodes.

brian-brazil: SGTM

storage/remote/read.go
fabxc: Now – or rather once we actually have someone who needs it?

brian-brazil: Noone has explicitly asked for it yet, but it is the plan of record from #2580

fabxc: Let's do it when we need it then.

promql/ast.go
```diff
@@ -242,52 +242,53 @@ type VectorMatching struct {
...
 // v.Visit(node) is not nil, Walk is invoked recursively with visitor
 // w for each of the non-nil children of node, followed by a call of
 // w.Visit(nil).
-func Walk(v Visitor, node Node) {
-	if v = v.Visit(node); v == nil {
+func Walk(v Visitor, node Node, path []Node) {
+	if v = v.Visit(node, path); v == nil {
 		return
 	}
+	path = append(path, node)
```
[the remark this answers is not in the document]
fabxc: No, as we return from recursion our length decrease and we'll overwrite nodes in the underlying array again.

fabxc:
@brian-brazil rebased and added the early-exit for binary expressions.

promql/ast.go
```diff
@@ -238,56 +238,58 @@ type VectorMatching struct {
 }
 
 // Visitor allows visiting a Node and its child nodes. The Visit method is
-// invoked for each node encountered by Walk. If the result visitor w is not
-// nil, Walk visits each of the children of node with the visitor w, followed
-// by a call of w.Visit(nil).
+// invoked for each node and path leading to it encountered by Walk.
```
brian-brazil: The path is passed as an additional parameter, it's the nodes that are getting walked - not the paths.

promql/engine.go
```diff
@@ -559,6 +544,25 @@ func (ng *Engine) populateIterators(ctx context.Context, s *EvalStmt) (storage.Q
...
+	case *AggregateExpr:
+		return n.Op.String()
+	case *Call:
+		return n.Func.Name
+	case *BinaryExpr:
+		// If we hit a binary expression we terminate since we only care about functions
+		// or aggregations over a single metric.
+		return ""
+	}
+	return extractFuncFromPath(p[:len(p)-1])
```
brian-brazil: This tail recursion can be converted to an iterated version.

storage/interface.go
```diff
@@ -61,6 +61,12 @@ type Querier interface {
 	Close() error
 }
 
+// SelectParams specifies parameters passed to data selections.
+type SelectParams struct {
+	Step int64  // query step size in milliseconds
```
brian-brazil: Comments should be full sentences.

promql/engine.go
brian-brazil: Go doesn't do tail call elimination, so this is less efficient and uses unnecessary stack memory.

juliusv: Since I read in chat that there was disagreement on this, adding one more opinion: I think this is fine either as a recursive or iterative variant. Performance or stack space shouldn't matter here. I also don't find the recursive variant confusing, so I think it's fine.

juliusv:
:+1: in general. Would be nice to add a description to the PR / commit that details what this is for and about.

promql/engine.go
fabxc: I recall sitting over this quite a bit back than and all re-usages turned out complicated and/or ugly.
Mainly because we get the type assurance through the type-switch statement and extracting something requires us to create another helper interface just for this and a function somewhere else.

The reuslting cognitive overhead just seemed worse.

fabxc:
Added description

juliusv:
:+1: I allowed myself to copy the description from the commit into the PR description.

PR #3681, opened 2018-01-13
some comments to explain why we use panic instead of t.Fatalf 

krasi-georgiev:


brian-brazil:
We don't have these comments in all the other places this happens, I don't think it's worth polluting the code base with them. Staticcheck will catch it.

krasi-georgiev:
@brian-brazil 
ok
do you know why staticcheck is not in  travis CI ?

brian-brazil:
I thought @SuperQ had added it

Issue #1766, opened 2016-06-26
Docker engine swarm api service discovery


F21:
In docker 1.12, the docker engine will ship with swarm mode built in. This means that it is now possible to stand up a swarm cluster using a bunch of nodes with just docker installed. In addition, swarm mode will come with dns and health checks built-in, negating the need to run consul or some other service discovery mechanism. More info here: https://docs.docker.com/engine/swarm/

It would be nice if prometheus can directly use the new services API to discover services running in a swarm cluster: https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/#3-8-services

Perhaps the config option could be called `docker_swarm_sd`.

brian-brazil:
> In addition, swarm mode will come with dns and health checks built-in, 

This will need to be bypassed for Prometheus service discovery.

We may want to wait for a release or two for this to stabilise before adding it, and ensuring there's sufficient interest to justify the maintenance effort of another SD.

bvis:
This feature would be amazing. It would allow us to simplify some dependencies now we need to manage to maintain a dynamic Prometheus environment

michaelharrer:
You could use dns_sd_configs.
Im running a global cadvisor service and a global prometheus/node-exporter service and can scrape all nodes with following config, using the
`tasks.<servicename>` feature of swarm mode discovery.

```
  - job_name: 'cadvisor'
    dns_sd_configs:
    - names:
      - 'tasks.cadvisor'
      type: 'A'
      port: 8080

  - job_name: 'node-exporter'
    dns_sd_configs:
    - names:
      - 'tasks.node-exporter'
      type: 'A'
      port: 9100
```

Its a workaround but functional.

Cas-pian:
Any progress on this? I'm really looking forward to use this feature.
Thanks very much!

genki:
@michaelharrer Unfortunately, there is no way to determine the node_exporter is running on which node. Only node_exporter itself knows it, but there's no option to provide it into its metrics (https://github.com/prometheus/node_exporter/issues/319)

joonas-fi:
I just hacked in a proof-of-concept that syncs tasks from Swarm manager to Prometheus: https://github.com/function61/prometheus-docker-swarm

Current limitation is that Prometheus has to be running on a Swarm manager node.

bvis:
[...]

PR #3687, opened 2018-01-16
Add Swarm discovery

cuigh:
Add Swarm discovery, see #1766 
@fabxc @brian-brazil


dsw9742:
@fabxc @brian-brazil hi guys, any chance of getting this valuable feature merged in the near future?

krasi-georgiev:
We are working on a new tool that writes simple json files and Prometheus picks up the targets via a file  SD  job. 
The tool would use the same interface as the built in once and if it gets widely adopted it might be added to the official ones supported by Prometheus.
So anyone interested in new Discoverers won't be blocked but upstream and would be able to implement a new Discoverer with a very little effort.

The Docker swarm discovery might be a good candidate to do a test run.

Here is the current progress https://github.com/prometheus/prometheus/pull/3720

cuigh:
@krasi-georgiev Yes, a common file SD is good, but I don't think we should use it to completely replace Swarm discovery. Swarm is a popular container orchestration system(not as popular as k8s, though), it's worth to be treated as a first-class citizen to be supported by Prometheus.

genki:
Why still no actions have taken for this PR?
Having dedicated SD for Docker swarm has enough value.

discovery/swarm/swarm.go
```diff
@@ -0,0 +1,504 @@
...
+
+	// serviceLabel is used for the name of the service in Swarm.
+	serviceLabel model.LabelName = metaLabelPrefix + "service"
+	// imageLabel is the label that is used for the docker image running the service.
+	imageLabel model.LabelName = metaLabelPrefix + "image"
+	// taskLabel contains the task name of the app instance.
+	taskLabel model.LabelName = metaLabelPrefix + "task"
+	// nodeIPLabel contains the node ip.
+	nodeIPLabel model.LabelName = metaLabelPrefix + "node_ip"
+	// nodeNameLabel contains the node ip.
```
ElfoLiNk: typo ?  should be `// nodeNameLabel contains the node name.`

krasi-georgiev:
You are completely right on all accounts , but the overhead of bug-fixes and support for a new SD is quite big. 

I think the current agreement is that any new SD would be implemented as a standalone binary that spit out json files which will be consumed by the file SD and eventually once it matures and has enough people to support  it might be added to Prometheus itself.

We are designing this [new tool](https://github.com/prometheus/prometheus/pull/3720) to implement the same SD interface so eventual integration would require almost no code changes.

All that said the docker swarm SD is a great addition and very usefull to many users and  it would be ideal candidate to see if this idea from https://github.com/prometheus/prometheus/pull/3720 will work well.

krasi-georgiev:
btw I have also started a discussion about a HTTP vs File SD so any  feedback backed with real use case would be really helpful
https://groups.google.com/forum/#!topic/prometheus-developers/JcE5nSbCe4k

discovery/swarm/swarm.go
cuigh: Already fixed, thx.

cuigh:
@krasi-georgiev Thanks for the reply. After reading all the posts relevant to this topic, it seems that Prometheus is unlikely to support Swarm SD directly within a short time. In this case, making file SD compatible with HTTP looks like a really good general solution, it also does satisfy our current usage scenarios-though it will complicate things a little bit. 

BTW, if HTTP is supported, perhaps I will consider adding a feature to [swirl](https://github.com/cuigh/swirl) as Prometheus Discovery adapter.:)

brian-brazil:
There is already a docker swarm file SD out there: https://github.com/ContainerSolutions/prometheus-swarm-discovery

cuigh:
@brian-brazil Yes, I know this unofficial project a long time ago, but for us, using the File SD is a little too tedious. We have dozens of Swarm clusters with at least two Prometheus instances deployed per cluster, this means that hundreds of additional tools need to be deployed at the same time to generate the configuration of the File SD, and if HTTP is supported, only one tool is needed. 

Of course, the built-in Swarm SD is always the best.

brian-brazil:
> this means that hundreds of additional tools need to be deployed at the same time to generate the configuration of the File SD, and if HTTP is supported, only one tool is needed.

At that scale you must have configuration management, and there's no real difference then. It'd still be expected that you run one per Prometheus, as that aligns the failure domains.

cuigh:
HTTP is more flexible relative to File, you can design your API like this:

```
{adapter-domain}/swarm?cluster={swarm-cluster} // Swarm
{adapter-domain}/xxx?addr={address}            // Other applications...
```

PR #3695, opened 2018-01-17
notifier: use testutil for testing

smonv:
- Use `util/testutil` in `notifier` testsuit. Related #3242

### DIFF
--- a/notifier/notifier_test.go
+++ b/notifier/notifier_test.go
@@ -22,7 +22,6 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
-	"reflect"
 	"testing"
 	"time"
 
@@ -64,9 +63,7 @@ func TestPostPath(t *testing.T) {
 		},
 	}
 	for _, c := range cases {
-		if res := postPath(c.in); res != c.out {
-			t.Errorf("Expected post path %q for %q but got %q", c.out, c.in, res)
-		}
+		testutil.Equals(t, c.out, postPath(c.in))
 	}
 }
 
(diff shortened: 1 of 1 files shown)


notifier/notifier_test.go
```diff
@@ -80,34 +79,23 @@ func TestHandlerNextBatch(t *testing.T) {
...
 
-	if len(b) != maxBatchSize {
-		t.Fatalf("Expected first batch of length %d, but got %d", maxBatchSize, len(b))
-	}
-	if !alertsEqual(expected[0:maxBatchSize], b) {
-		t.Fatalf("First batch did not match")
-	}
+	testutil.Equals(t, maxBatchSize, len(b))
+
+	testutil.Assert(t, alertsEqual(expected[0:maxBatchSize], b), "First batch did not match")
```
cstyan: maybe `alertsEqual` is something we want in `testutil?

krasi-georgiev: Yep , that sound reasonable.

tthanh:
thanks for feedback, I'll update this PR

notifier/notifier_test.go
tthanh: I think we can move the check len of 2 alert array outsite of `alertEqual` so `alertEqual` only contain the for loop.

tthanh:
@cstyan @krasi-georgiev I won't move `alertEqual` to `testutil` package because it specific to the alerts. Can you help provide more idea which part of `alertEqual` should I improve.

Updated resolve conflicts of `notifier` package with master

krasi-georgiev:
Did you check if you can refactor these to make it table based.
I find it more readable and easy to understand.

something like:
https://github.com/prometheus/prometheus/blob/6b0e9ef183fddf5b27b4ea4d2052153a6bdf0660/notifier/notifier_test.go#L41-L71

tthanh:
@krasi-georgiev thanks for feedback. I'll check again and update.

smonv:
@krasi-georgiev After check I think I can't not make more test like the table you mentioned. The test was step by step check, identical test flow, not multiple case.

notifier/notifier_test.go
```diff
@@ -443,13 +411,10 @@ func (a alertmanagerMock) url() *url.URL {
...
-	if err != nil {
-		t.Fatal(err)
-	}
 
-	if !reflect.DeepEqual(tg, makeInputTargetGroup()) {
-		t.Fatal("Target modified during alertmanager extraction")
-	}
+	testutil.Ok(t, err)
+
+	testutil.Assert(t, reflect.DeepEqual(tg, makeInputTargetGroup()), "Target modified during alertmanager extraction")
```
brian-brazil: Can you not use testutil.Equals here?

smonv: I didn't use `testutil.Equals` because it can not print custom message and I want respect original fatal message.

brian-brazil: That doesn't give us the diff then. I'd suggest making the message a comment.

smonv: Ok, I'll update it to `testutil.Equals`. Thank for feedback

brian-brazil:
Thanks!

Issue #2756, opened 2017-05-22
Prom2: SIGHUP for config reload causes artifacts in TSDB storage


mwitkow:
**What did you do?**
Run P2 from dev-2.0 branch side by side our P1 stack with the same config reloading and scraped targets.

**What did you expect to see?**
A stable retention similarly to P1.

**What did you see instead? Under which circumstances?**

A periodic, every 1h weird whole in our data, which massively crops up in rates.

![image](https://cloud.githubusercontent.com/assets/9655971/26313517/0af0c474-3f03-11e7-910e-5de778707b93.png)

This coincides with our config syncing job that sends a SIGHUP to prometheus at least every hour. The data loss corelates *exactly* with the SIHGUP log statement and with the prometheus_config_last_reload_success_timestamp

![image](https://cloud.githubusercontent.com/assets/9655971/26313578/35d3c33a-3f03-11e7-9f6b-6d9267a85f56.png)

@fabxc tracked down the issue

fabxc:
Thanks for figuring out the correlation. However, this puzzles me a fair bit as the storage is purely configured through flags and a reload does not touch it in any way. That only really leaves either the retrieval our the rule evaluation layer doing something different than 1.x.

Given that this is not a recording rule metric, it can only be retrieval. But still cannot think of anything that changed that could cause this.

mwitkow:
Maybe (wild ass guess) there's a subtle interplay between the timestamp of
it being stored and the delay in retrieval confusing rate?

Interestingly a side by side P1 is happy during these so it's definitely a
change in dev 2.

fabxc:
Can you provide a dump of the JSON response. Once of the raw metric and
once with rate() applied?
Ideally at the most coarse interval that still shows the problem and
annotation of the time that show it in the graph.

mwitkow:
![simleaf_bad_timing](https://cloud.githubusercontent.com/assets/9655971/26353248/5e28bb38-3fb7-11e7-9949-306dbca84cf6.png)

The JSON responses for both the `rate` and unrated data (as seen in the graph) is included below.
[dropped_hup_bad_data.tar.gz](https://github.com/prometheus/prometheus/files/1022029/dropped_hup_bad_data.tar.gz)

fabxc:
Having a bit of trouble here. The unaggregated window in the image is a bit to narrow.
Does the unaggregated data also have artifacts or not? Which window does the raw data cover? Around the artifact or just right until it occurs?

[...]

PR #2830, opened 2017-06-10
Reload only when config changes.

Gouthamve:
Pending tests and fixes #2756 

So what happens when we reload config is that we stop all scrape loops (which might have inflight scrapes) and start new ones. This causes some series to "miss" a scrape and causes the artifacts shown in #2756. This could be reproduced in 1.x also.

I propose that we compare the configs and reload only if it changes. But before I write the tests, I want to know if its sane to do this.

cc @brian-brazil @fabxc @beorn7

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/prometheus/2830)

PR #3698, opened 2018-01-17
Scraper config reload bugfix

krasi-georgiev:
When the discovery manager send the target list to the scrape manager , the scrape manager needs to read the configuration for the given target set(job name). 
If the discovery manager send targets for a job which configuration is still not loaded in the scrape manager it will fail to create scrape pools for this target set(job name).

This happens at boot time or when reloading all configs.

Another fix would be to send the job configs when sending the  targets list from the Discovery manager.

the config handling is due for some refactoring anyway so I think best to have this fix for the 2.1 release and think about the refactoring after that.

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -28,6 +28,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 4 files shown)


Gouthamve:
This also needs to handle the config reload when the HTTPClientConfig changes.

Gouthamve:
LGTM. Tests remain to be ported from https://github.com/prometheus/prometheus/pull/3673/

Gouthamve:
merge on green.

krasi-georgiev:
@Gouthamve didn't know this new command :)

krasi-georgiev:
@Gouthamve closes  #2756, #2830

Issue #3702, opened 2018-01-18
Default consul server to localhost:8500


brian-brazil:
You should always be using the local consul server and it usually runs on a well known port, so it'd be reasonable to default it accordingly.

zemek:
I'd like to try and make this change if nobody else has started on it

PR #3703, opened 2018-01-18
Set consul server default to localhost:8500

zemek:
Set the consul server default to `localhost:8500` as that is the default port consul listens on.

see: https://github.com/prometheus/prometheus/issues/3702
@brian-brazil

### DIFF
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -305,10 +305,10 @@ The following meta labels are available on targets during [relabeling](#relabel_
 ```yaml
 # The information to access the Consul API. It is to be defined
 # as the Consul documentation requires.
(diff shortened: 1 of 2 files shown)


docs/configuration/configuration.md
```diff
@@ -305,7 +305,7 @@ The following meta labels are available on targets during [relabeling](#relabel_
 ```yaml
 # The information to access the Consul API. It is to be defined
 # as the Consul documentation requires.
-server: <host>
+server: <host> | default = "localhost:8500"
```
brian-brazil: We put [] around items with defaults

zemek:
@brian-brazil how does it look?

brian-brazil:
Thanks!

PR #3709, opened 2018-01-19
read bearer token on every request + some http and scrape tests

krasi-georgiev:
read bearer token on every request
removed unuseful  scrape manager startup log
new tests -TestScrapeManagerReloadNoChange( scrape pool is not reloaded
when the config hasn't changed), TestMissingBearerAuthFile ,
TestBearerAuthFileRoundTripper

### DIFF
--- a/retrieval/manager.go
+++ b/retrieval/manager.go
@@ -56,8 +56,6 @@ type ScrapeManager struct {
 
 // Run starts background processing to handle target updates and reload the scraping loops.
 func (m *ScrapeManager) Run(tsets <-chan map[string][]*targetgroup.Group) error {
-	level.Info(m.logger).Log("msg", "Starting scrape manager...")
-
 	for {
 		select {
 		case ts := <-tsets:
--- a/util/httputil/client_test.go
+++ b/util/httputil/client_test.go
@@ -289,13 +289,13 @@ func TestBearerAuthRoundTripper(t *testing.T) {
 		}
 	}, nil, nil)
 
-	//Normal flow
+	// Normal flow.
 	bearerAuthRoundTripper := NewBearerAuthRoundTripper(BearerToken, fakeRoundTripper)
 	request, _ := http.NewRequest("GET", "/hitchhiker", nil)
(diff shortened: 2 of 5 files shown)


krasi-georgiev:
@Gouthamve it is ready. 
most of this is your code anyway :)

util/httputil/client_test.go
```diff
@@ -272,6 +302,32 @@ func TestBearerAuthRoundTripper(t *testing.T) {
...
+
+	fakeRoundTripper := testutil.NewRoundTripCheckRequest(func(req *http.Request) {
+		bearer := req.Header.Get("Authorization")
+		if bearer != ExpectedBearer {
+			t.Errorf("The expected Bearer Authorization (%s) differs from the obtained Bearer Authorization (%s)",
+				ExpectedBearer, bearer)
+		}
+	}, nil, nil)
+
+	//Normal flow
```
Gouthamve: space after `//` and should end with `.`.

Same below.

krasi-georgiev: fixed. 
btw these were some comment from a someone elses commit, so no red points for me :)

krasi-georgiev: @Gouthamve ^

retrieval/manager.go
```diff
@@ -56,8 +56,6 @@ type ScrapeManager struct {
 
 // Run starts background processing to handle target updates and reload the scraping loops.
 func (m *ScrapeManager) Run(tsets <-chan map[string][]*targetgroup.Group) error {
-	level.Info(m.logger).Log("msg", "Starting scrape manager...")
```
brancz: why was this removed?

util/httputil/client_test.go
brancz: @krasi-georgiev it was only fixed in the comments below, not these

retrieval/manager.go
krasi-georgiev: it is not useful at all and no other Managers has this. 

It is useful only for the tsdb  - starting - started

PR #3721, opened 2018-01-21
promql: Make printer formatting less vintage

juliusv:
- lower-case modifiers
- reverse order of aggregation modifiers and aggregated expression
- remove spacing before modifier parentheses


juliusv:
@grobie I kinda like the current briefer style, but wouldn't be opposed to changing it. Probably a question for a separate PR though if you feel strongly :)

PR #3722, opened 2018-01-22
rulefmt: Allow parsing from bytes instead of file

juliusv:
This makes the rulefmt package more reusable for users whose rulesets are not
stored in files on disk. Concrete example: Cortex.


juliusv:
@Gouthamve

Gouthamve:
👍

PR #3723, opened 2018-01-22
discovery: Rename file SD mtime metric

juliusv:
- "timestamp" -> "mtime" to be in line with node exporter and clearer.
- add unit suffix


brian-brazil:
:+1:

PR #3736, opened 2018-01-25
Don't allocate a mergeSeries if there is only one series to merge.

### LINKED ISSUE #3715: CPU Usage triples after upgrade to v2.1.0
**What did you do?**

Upgraded Prometheus from v2.0.0 to v2.1.0 (using the official Docker image)

Similar or better performance figures.

CPU Usage almost tripled and rule evaluation times increased, we observed the same on two different deployment (two separate K8s clusters), after failing back to v2.0.0 figures are back to normal.

Cluster 1 [CPU]:
image

Cluster 2 [CPU]:
image

Cluster 1 [Rule evluation duration]:
image

Cluster 2 [Rule evluation duration]: graph looks about the same, but this cluster is smaller:
image

Except changing the docker image version, everything remained unchanged.

**Environment**

* System information:

`Linux 4.12.10-coreos x86_64` (Runs in K8s 1.5.3 Cluster)

* Prometheus version:

```
  go version:       go1.9.2
```

* Alertmanager version:

	N/A

```
Can provide if needed
```

```
N/A
```

* Logs:
```
Can provide if needed
```

tomwilkie:
~Might help #3715~

Doesn't actually seem to help, but are good to have none-the-less.

### DIFF
--- a/storage/fanout_test.go
+++ b/storage/fanout_test.go
@@ -255,22 +255,17 @@ func BenchmarkNoMergeSeriesSet_100_100(b *testing.B) {
 	benchmarkDrain(seriesSet, b)
 }
 
-func BenchmarkMergeSeriesSet_1_100_100(b *testing.B) {
-	seriesSet := makeMergeSeriesSet(1, 100, 100)
-	benchmarkDrain(seriesSet, b)
-}
-
-func BenchmarkMergeSeriesSet_10_100_100(b *testing.B) {
-	seriesSet := makeMergeSeriesSet(10, 100, 100)
-	benchmarkDrain(seriesSet, b)
-}
-
-func BenchmarkMergeSeriesSet_100_100_100(b *testing.B) {
-	seriesSet := makeMergeSeriesSet(100, 100, 100)
-	benchmarkDrain(seriesSet, b)
-}
-
-func BenchmarkMergeSeriesSet_1000_100_100(b *testing.B) {
-	seriesSet := makeMergeSeriesSet(100, 100, 100)
-	benchmarkDrain(seriesSet, b)
+func BenchmarkMergeSeriesSet(b *testing.B) {
+	for _, bm := range []struct {
+		numSeriesSets, numSeries, numSamples int
+	}{
+		{1, 100, 100},
+		{10, 100, 100},
+		{100, 100, 100},
+	} {
+		seriesSet := makeMergeSeriesSet(bm.numSeriesSets, bm.numSeries, bm.numSamples)
+		b.Run(fmt.Sprintf("%d_%d_%d", bm.numSeriesSets, bm.numSeries, bm.numSamples), func(b *testing.B) {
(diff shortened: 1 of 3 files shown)


tomwilkie:
Circle CI failures are unrelated (not sure whats going on, but theres nothing wrong with this PR)

grobie:
Do we have a benchmark showing the impact in terms of allocations? If not, could we add one, so that we can catch regressions in future releases?

tomwilkie:
> Do we have a benchmark showing the impact in terms of allocations?

Not thats I know of :-(

> If not, could we add one, so that we can catch regressions in future releases?

Sure, will try and get it done today.

tomwilkie:
Before these changes:
```
$ go test -bench=MergeSeriesSet ./storage/
BenchmarkNoMergeSeriesSet_100_100-4      	500000000	         2.78 ns/op
BenchmarkMergeSeriesSet_1_100_100-4      	200000000	         6.88 ns/op
```

After:
```
$ go test -bench=MergeSeriesSet ./storage/
BenchmarkNoMergeSeriesSet_100_100-4      	1000000000	         2.51 ns/op
BenchmarkMergeSeriesSet_1_100_100-4      	1000000000	         2.89 ns/op
```

Basically, calling MergeSeriesSet on a single SeriesSet is now no extra cost.  I added some extra benchmarks but they need more reasoning first.

storage/fanout_test.go
```diff
@@ -218,13 +219,58 @@ func (m *mockSeriesSet) Err() error {
...
+	benchmarkDrain(seriesSet, b)
+}
+
+func BenchmarkMergeSeriesSet_100_100_100(b *testing.B) {
+	seriesSet := makeMergeSeriesSet(100, 100, 100)
+	benchmarkDrain(seriesSet, b)
+}
+
+func BenchmarkMergeSeriesSet_1000_100_100(b *testing.B) {
+	seriesSet := makeMergeSeriesSet(100, 100, 100)
```
grobie: This is identical to the one above.

storage/fanout_test.go
```diff
@@ -218,13 +219,58 @@ func (m *mockSeriesSet) Err() error {
...
+		}
 	}
 }
+
+func BenchmarkNoMergeSeriesSet_100_100(b *testing.B) {
+	seriesSet := makeSeriesSet(100, 100)
+	benchmarkDrain(seriesSet, b)
+}
+
+func BenchmarkMergeSeriesSet_1_100_100(b *testing.B) {
```
grobie: Why not using sub-benchmarks?

tomwilkie: Didn't know about those!  Much nicer.

storage/fanout_test.go
tomwilkie: Roger; I removed it.

tomwilkie:
@grobie can I take the 'Looks great' as an LGTM :-) ?

Issue #3735, opened 2018-01-25
fatal error: concurrent map iteration and map write


tonobo:
**What did you do?**

Just a few queries and scraping ~1,5k targets.

**Environment**

* System information:

```
level=info ts=2018-01-25T09:28:37.474952519Z caller=main.go:225 msg="Starting Prometheus" version="(version=2.1.0, branch=HEAD, revision=85f23d82a045d103ea7f3c89a91fba4a93e6367a)"
level=info ts=2018-01-25T09:28:37.474989196Z caller=main.go:226 build_context="(go=go1.9.2, user=root@6e784304d3ff, date=20180119-12:01:23)"
level=info ts=2018-01-25T09:28:37.475001574Z caller=main.go:227 host_details="(Linux 4.10.0-37-generic #41~16.04.1-Ubuntu SMP Fri Oct 6 22:42:59 UTC 2017 x86_64 internal1 (none))"
```
* Logs:

The whole go panic: [prom.log.gz](https://github.com/prometheus/prometheus/files/1663463/prom.log.gz)

krasi-georgiev:
@tonobo it might help if you can provide a minimal config to replicate the issue.
btw does it happen on every start ?


@cstyan any idea why this might happen? the locks seem properly used.
```
fatal error: concurrent map iteration and map write

goroutine 2820346 [running]:
runtime.throw(0x1c1d139, 0x26)
	/usr/local/go/src/runtime/panic.go:605 +0x95 fp=0xc5c29d5c50 sp=0xc5c29d5c30 pc=0x42bca5
runtime.mapiternext(0xc5c29d5e58)
	/usr/local/go/src/runtime/hashmap.go:778 +0x6f1 fp=0xc5c29d5ce8 sp=0xc5c29d5c50 pc=0x40a031
github.com/prometheus/prometheus/discovery/file.(*TimestampCollector).Collect(0xc4201a3b60, 0xc613b36720)
	/go/src/github.com/prometheus/prometheus/discovery/file/file.go:99 +0x17b fp=0xc5c29d5f98 sp=0xc5c29d5ce8 pc=0xaacd7b
github.com/prometheus/prometheus/vendor/github.com/prometheus/client_golang/prometheus.(*Registry).Gather.func2(0xc66aa367c0, 0xc613b36720, 0x28f9b00, 0xc4201a3b60)
[...]

PR #3738, opened 2018-01-25
Fixed race condition in map iteration and map write in Discovery (#3735)

codesome:
Added lock for this loop 
[https://github.com/prometheus/prometheus/blob/master/discovery/file/file.go#L99-L101](https://github.com/prometheus/prometheus/blob/master/discovery/file/file.go#L99-L101) 
using `fileSD.lock`

(Issue [#3735](https://github.com/prometheus/prometheus/issues/3735))

### DIFF
--- a/discovery/file/file.go
+++ b/discovery/file/file.go
@@ -96,9 +96,11 @@ func (t *TimestampCollector) Collect(ch chan<- prometheus.Metric) {
 	uniqueFiles := make(map[string]float64)
 	t.lock.RLock()
 	for fileSD := range t.discoverers {
+		fileSD.lock.Lock()
 		for filename, timestamp := range fileSD.timestamps {
 			uniqueFiles[filename] = timestamp
 		}
+		fileSD.lock.Unlock()
 	}
 	t.lock.RUnlock()
 	for filename, timestamp := range uniqueFiles {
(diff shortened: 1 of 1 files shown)


discovery/file/file.go
```diff
@@ -96,9 +96,11 @@ func (t *TimestampCollector) Collect(ch chan<- prometheus.Metric) {
 	uniqueFiles := make(map[string]float64)
 	t.lock.RLock()
 	for fileSD := range t.discoverers {
+		fileSD.lock.Lock()
```
Gouthamve: This would cause rapid locking and unlocking. Can you put it outside the loop?

Gouthamve: Sorry, not sure what I was thinking. Looks good :)

cstyan:
hmmm, read/write locks should have worked :|

Gouthamve:
@cstyan https://github.com/prometheus/prometheus/issues/3735#issuecomment-360455165

Yes, but the locks are different :P

cstyan:
ooooooooohhh, oh boy

yeah, good catch! @Gouthamve

codesome:
@cstyan yes, did not notice that it was `RWMutex`. Changed it to `RLock` now.

cstyan:
this looks good to me

PR #3740, opened 2018-01-25
discovery: Throttle updating discovery per provider

### LINKED ISSUE #3715: CPU Usage triples after upgrade to v2.1.0

brancz:
This PR is still in progress as context handling needs to be added. The initial round of profiling suggested that syncs are happening very often so targets and their labels are built/allocated often. This PR throttles updates to targets to once every 5 seconds. This is the same behavior that we had in `v2.0.0`, for `v2.1.0` this throttle was removed in a series of refactorings.

This is what the memory allocations look like with the patch applied

query: `rate(go_memstats_alloc_bytes_total{job="prometheus"}[1m])`

screenshot from 2018-01-25 22-35-53

As you can see this is still higher than `v2.0.0`, so I profiled again and the new result is very different.

alloc

It instead shows that the majority of allocations are coming from reads from disk, this is a separate problem though.

I believe this closes #3715.

### DIFF
--- a/notifier/notifier.go
+++ b/notifier/notifier.go
@@ -298,8 +298,14 @@ func (n *Notifier) nextBatch() []*Alert {
 
 // Run dispatches notifications continuously.
 func (n *Notifier) Run(tsets <-chan map[string][]*targetgroup.Group) {
-
 	for {
+		// Throttle syncing to once per five seconds.
+		select {
+		case <-n.ctx.Done():
+			return
+		case <-time.After(5 * time.Second):
+		}
+
 		select {
 		case <-n.ctx.Done():
 			return
--- a/notifier/notifier_test.go
+++ b/notifier/notifier_test.go
@@ -396,7 +396,7 @@ func TestHandlerQueueing(t *testing.T) {
 		case <-called:
 			expected = alerts[i*maxBatchSize : (i+1)*maxBatchSize]
 			unblock <- struct{}{}
-		case <-time.After(5 * time.Second):
+		case <-time.After(60 * time.Second):
 			t.Fatalf("Alerts were not pushed")
 		}
 	}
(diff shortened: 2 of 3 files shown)


krasi-georgiev:
with this change aren't we blocking the sender?

for example if the provider tries to send 20 update cycles the receiver will receive the last (most updated one) after 1 minute.

I was thinking something like - drain the sender , but process the update only after 5 seconds.
https://github.com/prometheus/prometheus/compare/master...krasi-georgiev:throtle-discovery-updates?expand=1

FUSAKLA:
sadly i have to confirm that the number of reloads from disk were not influenced by this fix

figure showing this fix influence on consul SD but not influencing the reloaded blocks from disk

screenshot from 2018-01-26 01-10-59

brancz:
While this "fixes" the alloc problem it doesn't actually work as throttling is too aggressive and SD is updated only very slowly.

Issue #3718, opened 2018-01-21
Autocomplete on expression inputs is overzealous


juliusv:
See:

![autocomplete](https://user-images.githubusercontent.com/538008/35195514-0ea4532e-fec5-11e7-9158-d8252939364b.png)

When I type that expression and just press "Return", it will now select a completely unrelated historical expression and execute that instead. It's now quite hard to just query what you want to query when having "Search query history" enabled.

This should be smarter / more restrictive in the matching of history items, and/or it should not auto-select the first completion item, but rather whatever the user has typed (which would require the user to press DownArrow to select the first completion item instead).

lovisas:
Not auto selecting the first completion item is something that definitively should be fixed. I'll take a go at making the matching against historical expressions smarter as well.

juliusv:
@lovisas Awesome, thanks!

lovisas:
Let me know if this is blocking anyone/anything and I'll split up the work in two smaller PRs to get it through quicker!

beorn7:
We could consider the current behavior a bug (even a regression as we had the same problem before and then reverted the advanced autocompletion) and then release a fix in 2.1.1.
Or we could just release 2.2 quite soon. As discussed earlier, we are not shy releasing frequently if it delivers some value.

juliusv:
Yeah, I did mark this issue as a bug and consider it one (that really impacts Prometheus use), so doing a 2.1.1 soon would be a good idea.

lovisas:
Did some investigation this morning - the autocomplete behavior is there by default. The query history checkbox is just changing what metrics you have available in the dropdown and otherwise it works the same as if you were to have the history unchecked. 

1) I could modify the dropdown for the expression browser so that it does not select the top metric when clicking enter. This could just be for when the checkbox is checked, but I'd argue the behavior should be the same regardless of mode. Fastest fix would be to disable the submission of the query when you hit `return`, but that might just make the situation worse.
2) The problem you bring up could potentially be solved by just making the matching algorithm smarter.

paddynewman:
+1 for a 2.1.1 release soon.

PR #3742, opened 2018-01-26
Fix: Improve typeahead on /graph page

lovisas:
This PR is an attempt to improve the autocomplete behavior in the expression browser to support the query history feature. 

The issues described in #3718 can be avoided by not automatically selecting the first item in the dropdown. With only one query in the dropdown, and a different query in the expression field, when you hit `return` it would take the first item in the dropdown and query for it since it was selected. With this change, you have to press the down arrow in order to have the query populate in the expression browser.


juliusv:
Thanks! Am I understanding it correctly that you are modifying the existing typeahead library in place to make it support that? Is there no way to configure this behavior from the outside? Or is there some file missing from this PR.

lovisas:
Yes I am! There is no file missing from the PR. If this change is accepted, I was going to replace the minified bootstrap-typeahead file with this one. Changing the library directly was the fastest/easiest solution I found. 

I'm usually not a fan of modifying third party code. However, given that this dependency hadn't been updated since it was added three years ago and that a potential rewrite would probably remove this dependency altogether, I thought it was ok this time.

juliusv:
So if we really *need* to modify it, we should unvendor/fork (just put it where own JS files live) and give it a slightly different name that doesn't make people think it's an unmodified version of the original. Otherwise it's too easy for someone to accidentally overwrite the vendored file and lose any modifications. We should still have a comment at the top of the forked file of course that explains what it's based on (with the original copyright comments) and what the modifications were.

Normally I would really be against forking something like this for a tiny modification (makes it hard to update the dependency), but your argument about the new UI getting rid of this altogether is a good point that will hopefully hold true :)

lovisas:
Unvendored the dependency and added a comment at the top to explain the changes. 

I completely agree about forking dependencies. I see this as a temporary fix that I wanted to get out quickly given that it impacts users negatively today. If the rewrite doesn't happen for whatever reason, I promise I'll revisit this and make it better! :)

lovisas:
ping @juliusv - let me know what you think of this. I've spent more time in the Prometheus UI the past few days and this has definitively bit me several times...

web/ui/static/js/graph/typeahead.js
```diff
@@ -0,0 +1,724 @@
...
+      // Add item
+      if (this.options.addItem){
+        items.push(this.options.addItem);
+      }
+
+      return this.render(items).show();
+    },
+
+    matcher: function (item) {
+      console.log("using built in macther")
```
juliusv: I guess these `console.log`s should no longer be in here?

juliusv:
Thanks! I still stumbled over some bugs while giving this a test run:

- When selecting a dropdown item, input field focus is lost. The cursor should remain in the input field and should be placed after the inserted item.

- Clicking "Search query history" before executing any valid query (that is not just a metric query) causes a JS error and the button stays deactivated.

- New expressions aren't immediately found in the search. In a fresh browser, do: a) Execute `rate(a[1m])`, b) Enable "Search query history", c) Validate that `rate(a[1m])` is found in the history and select it, d) modify the query to `rate(a[1m]) + rate(b[1m])` and execute it, e) enter `rate(` and see that only the original `rate(a[1m])` query appears in the dropdown. After re
loading the page, however, the second query also shows up.

paddynewman:
I'm just going to put this out there in case there's any interest. How about simply reverting to the 2.0 behaviour?

Is this worth forking a vendor library for? I don't think so but that's only my opinion.

juliusv:
@paddynewman I think it's worth it as a temporary solution before we get a UI rewrite. However, if there's pressure to release 2.1.1 and we can't get this fixed up before that (hopefully we can), then just rolling back for that release could be an option.

lovisas:
Yeah, that's definitively an option. I don't like forking vendor libraries either, but those dependencies hadn't been touched in 3 years, since they were added. [The proposal to rewrite the UI ](https://groups.google.com/forum/#!topic/prometheus-developers/CAUUFwH-8Fg)has gained some momentum so hopefully that ball starts rolling in the next few weeks.

I am working on fixing all the things Julius pointed out above, this should be done in a few hours.

juliusv:
@lovisas Awesome, thanks. I'm currently flying to the other end of the globe, so it might be a day or two before I am able to get back to it.

lovisas:
No rush, safe travels! Thanks for all your feedback on this so far @juliusv.

lovisas:
Sooooo it turns out that we didn't need to fork the bootstrap-typeahead at all. Instead of modifying the version of the library that we were using, I took the latest version which introduced all of the bugs. 

The current state of the PR: 
- the dependency isn't forked and we pass additional options to modify the behavior
- items in the dropdown are not autoselected 
- historical queries are only shown when the dropdown is toggled on
- small refactor to move shared behavior to a function

Apologize for all the back and forth that could have been avoided, next time I'll pay better attention to the versions... s/o to @csmarchbanks for getting me on the right track.

web/ui/static/js/graph/index.js
```diff
@@ -260,10 +260,11 @@ Prometheus.Graph.prototype.populateInsertableMetrics = function() {
 };
 
 Prometheus.Graph.prototype.initTypeahead = function(self) {
-  const historyIsChecked = $("div.query-history").hasClass("is-checked");
-  const source = historyIsChecked ? pageConfig.allMetrics.concat(JSON.parse(localStorage.getItem("history"))) : pageConfig.allMetrics;
+  const source = queryHistory.isEnabled() ? pageConfig.queryHistMetrics.concat(pageConfig.allMetrics) : pageConfig.allMetrics;
 
   self.expr.typeahead({
+    afterSelect: this.submitQuery.bind(this),
```
csmarchbanks: Just a thought, in master prometheus you have to click "enter" twice right now to do a query from the typeahead (once to select an entry in the typeahead and once to submit the query). Now the query would be submitted on the first enter if something in the typeahead is selected, but would still need two enters if nothing is selected in the typeahead. 

The difference in behavior could feel a bit buggy, but I am not sure on the correct answer since one enter queries are nice.

juliusv:
@lovisas That's excellent news! Taking another look...

web/ui/static/js/graph/index.js
juliusv: Good catch. At least when selecting just a normal metric from the dropdown, I normally wouldn't want it to evaluate the expression immediately, since usually you'll want to add other expression bits around it first. For the case of selecting an entire historical expression, I think either is fine, but maybe it's better to be consistent and not immediately evaluate in either case.

juliusv:
@lovisas Works great now and obviously it's much nicer now :)

juliusv:
:+1: Awesome, with the last change to not auto submit, it looks good. Merging.

PR #3743, opened 2018-01-26
Add debounce when changing range

### LINKED ISSUE #406: Debounce timerange setting changes
After pressing any of the buttons that cause timerange changes, we should wait a short amount of time before updating the graph. That way, if the user presses the button multiple times, we don't load every intermediate result immediately.

biglovisa:
Fixes #406

Wraps all functions that change range in a debounce function.


lovisas:
@Conorbro would you be able to review this? I took a different approach than suggested in the related issue so that it wouldn't slow down the page.

simonpasquier:
@biglovisa can you rebase on master?

biglovisa:
I will open up a new PR for this instead, I don't have access to this fork any longer.

biglovisa:
@simonpasquier new PR is up on #4495

PR #3745, opened 2018-01-26
Update api.md

krysennn:
Update doc with correct example for endpoint "POST /api/v1/admin/tsdb/delete_series"


brian-brazil:
Thanks!

PR #3747, opened 2018-01-26
Update throttle & tsdb update

brancz:
@krasi-georgiev @Gouthamve and I have been working tirelessly to find a way to reasonably throttle SD updates, and for now settled on this one. After a couple rounds of benchmarking this seems to to the trick.

Testing + feedback would be highly appreciated! :slightly_smiling_face: 

/update

`rate(prometheus_target_scrape_pool_sync_total{prometheus!="meta"}[1m])`
screenshot from 2018-01-26 18-27-41

`container_memory_usage_bytes{container_name="prometheus"}`
screenshot from 2018-01-26 18-26-17

`sum(rate(container_cpu_usage_seconds_total{container_name="prometheus"}[1m])) by(name)`
screenshot from 2018-01-26 18-25-52

`rate(go_memstats_alloc_bytes_total{job="prometheus"}[1m])`
screenshot from 2018-01-26 18-25-34

### DIFF
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
@@ -81,6 +82,10 @@ type Manager struct {
 	targets map[poolKey]map[string]*targetgroup.Group
 	// The sync channels sends the updates in map[targetSetName] where targetSetName is the job value from the scrape config.
 	syncCh chan map[string][]*targetgroup.Group
+	// True if updates were received in the last 5 seconds.
+	recentlyUpdated bool
+	// Protects recentlyUpdated.
+	recentlyUpdatedMtx sync.Mutex
 }
 
 // Run starts the background processing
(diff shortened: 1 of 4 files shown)


discovery/manager.go
```diff
@@ -137,7 +143,26 @@ func (m *Manager) runProvider(ctx context.Context, poolKey poolKey, updates chan
 				return
 			}
 			m.updateGroup(poolKey, tgs)
-			m.syncCh <- m.allGroups()
+			m.recentlyUpdatedMtx.Lock()
+			m.recentlyUpdated = true
+			m.recentlyUpdatedMtx.Unlock()
```
fabxc: Why not just put the following after `m.syncCh <- ...` here:

```go
select {
case <-time.After(5 * time.Second):
case <-ctx.Done():
    return
}
```

fabxc: The solution here will send updates even if there were none.

krasi-georgiev: We tried this initially, but this just blocks the sender and the updates just pile up and get delayed.

What we want is to drain the sender , but send updates to the Scraper only once at a given interval.

Gouthamve: @fabxc Not sure why this will send updates even if none exist though.

The issue with doing a `time.After(5 * time.Second)` is that if the SD mechanism launches a goroutine per update, we might end up with a lot of waiting goroutines because we are not draining the channel.

Gouthamve:
👍 from me

brancz:
:+1: on @krasi-georgiev's change

krasi-georgiev:
:+1:  :smile_cat:  from me  as well

brancz:
Ok then let's move forward.

PR #3748, opened 2018-01-26
Discovery/Kubernetes: If target is Node, attach labels of it.

cofyc:
## What this PR does

For all targets backed by a node, all labels of the `role: node`
discovery are attached.

## Why we need it

We are currently using [prometheus-operator](https://github.com/coreos/prometheus-operator) to scrape metrics from kubelet. `prometheus-operator` maintains a endpoint object (`kubelet` by default), e.g:

```
apiVersion: v1
kind: Endpoints
metadata:
  labels:
    k8s-app: kubelet
  name: kubelet
  namespace: kube-system
subsets:
- addresses:
  - ip: 192.168.128.116
    targetRef:
      kind: Node
      name: nodea
      uid: 1c75a366-1e85-11e7-bacc-f23c9184ec92
  - ip: 192.168.130.217
    targetRef:
      kind: Node
      name: nodeb
      uid: 5165df10-888b-11e7-83d7-f23c9184ec92
  ports:
  - name: http-metrics
    port: 10255
    protocol: TCP
  - name: cadvisor
    port: 4194
    protocol: TCP
  - name: https-metrics
    port: 10250
    protocol: TC
```

In current prometheus, if kubernetes discovery role is `endpoint`, it does not attach labels of `role: node`. Metrics scraped from `kubelet` on each node does not contain node labels. In many cases, we need to do [vector matching](https://prometheus.io/docs/prometheus/latest/querying/operators/#vector-matching) with others metrics (e.g. `kube_node_labels` from `kube-state-metrics`). But without node name label, it's hard to do.

IMHO, it will be convenient, if targets backed by a node, all labels of the `role: node`
discovery are attached, like `Pod` target.

### DIFF
--- a/discovery/kubernetes/endpoints.go
+++ b/discovery/kubernetes/endpoints.go
@@ -23,6 +23,7 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/discovery/targetgroup"
+	"github.com/prometheus/prometheus/util/strutil"
 	apiv1 "k8s.io/client-go/pkg/api/v1"
 	"k8s.io/client-go/tools/cache"
 )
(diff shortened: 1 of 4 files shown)


cofyc:
cc @brian-brazil @fabxc @brancz

brancz:
I don't think the entire node labels are necessary. A generic target `kind` and `name` would be the correct mapping to Prometheus meta labels I think.

brian-brazil:
I'm a big confused by this, can you share the prometheus.yml you are using?

cofyc:
@brian-brazil 
Here: https://pastebin.com/u6kj0x1t, it is generated and managed by prometheus-operator. 

In short, we want to vector matching on metrics from kubelet and kube-state-metrics. But with prometheus-operator kubelet target are discoveried through `endpoints` role, no node info are attached now.  So I try to attach node labels if endpoint target is backed by node, like `Pod` target.

@brancz That would be much simpler! I created a separate PR to address this: https://github.com/prometheus/prometheus/pull/3770

brancz:
Got it, that makes sense. As an explanation why the nodes are in the `Endpoints` object, is because it basically simulates self-hosted kubelets, until we actually have them self-hosted.

PR #3750, opened 2018-01-27
last config success time gauge: use SetToCurrentTime()

roidelapluie:


roidelapluie:
I think we should use Gauge's built in SetToCurrent time for this.

The side effect is that the gauge now will have nanosecond precision.

brian-brazil:
Thanks!

PR #3751, opened 2018-01-27
update the discover tests

krasi-georgiev:
the discovery test is now only testing update and get groups.
It doesn't do an e2e test but just a unit test of setting and receiving
target groups


krasi-georgiev:
will open directly against the  prometheus/sched-update-throttle branch

Issue #3746, opened 2018-01-26
rate()/increase() extrapolation considered harmful


free:
**The problem**

*[Feel free to skip ahead to the proposed solution, as I'm guessing the problem is widely understood. Or just read the last couple of paragraphs for a real-life example of the problem and my current workaround.]*

I've been trying to deal with the errors introduced by the `rate()` and `increase()` functions for so long that I keep going back every few months and spend a couple of hours tweaking Grafana dashboards before I remember that there is literally **no way of getting an accurate result** out of them short of using recorded rules for everything and, even in that case, working around and against Prometheus.

To clarify, let's assume I have a monitoring configuration with **scrapes every 5 seconds** and **rule evaluation every 10 seconds**. It collects among other things a `request_count` counter, that increases by 1 every 20 seconds. Let's also assume that I would really like a Grafana graph that displays an accurate number of requests over time, on a dashboard with an adjustable time range. (I.e. I want to be able to look at 15 minutes, 1 hour and 7 days on the same dashboard, with Grafana automatically adjusting the resolution.) Not a tall order, right?

For the purpose of clarity, here is how the `request_count` samples would look like (at a 5 second resolution):

    [0, 1, 1, 1, 1, 2, 2, 2, 2, 3, 3, 3, 3, 4, 4, 4, 4, ...]

The naive approach is you add a graph and put in an expression along the lines of:

    increase(request_count[$__interval])

and set *Min step* to `10s` (I would prefer a *Min step* of `5s`, since the resolution of the underlying data is 5 seconds, but let's not even go there yet). Ideally the graphed values (with a 10 second resolution) would look something like:

    [0, 1, 0, 1, 0, 1, 0, 1, 0, ...]

What I'm getting instead, depending on the exact time I reload the page is one of:

    [0, 2, 0, 2, 0, 2, 0, 2, 0, ...]

or

    [0, 0, 0, 0, 0, 0, 0, 0, 0, ...]

[...]

PR #3760, opened 2018-01-30
Implement xrate/xincrease/xdelta functions, as per #3746

free:
Took a while to figure out that I also needed to tweak `(*Engine).populateIterators()` in order to get the extra point(s) before the range start, but it's finally done. And, if I might say so, it's relatively clean (minus the range adjustment at the top of `exactRate()`.

I ended up making one slight change from my original proposal, which is not to adjust the increase by `range / sampledInterval`, as it would only add noise. If the range is not an "exact" multiple of the sampling interval, I believe it's OK to get a sawtooth pattern for smoothly increasing counters for the price of an exact value (which is quite useful for not-so-smoothly increasing counters). Rates (`xrate` and `xdelta`) are not affected by this, as they are always divided by the actual sampled interval, so they will be perfectly smooth for a perfectly smooth counter.

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -493,21 +493,20 @@ func (ng *Engine) populateIterators(ctx context.Context, s *EvalStmt) (storage.Q
 	var maxOffset time.Duration
 
 	Inspect(s.Expr, func(node Node) bool {
+		nodeOffset := LookbackDelta
 		switch n := node.(type) {
 		case *VectorSelector:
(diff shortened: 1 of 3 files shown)


brian-brazil:
As explained in https://github.com/prometheus/prometheus/issues/3746#issuecomment-361572859 I'm afraid I cannot accept this PR as-is. If you have a proposal that fits in with the rest of PromQL I'd be happy to consider it.

Issue #3707, opened 2018-01-19
Prometheus 2.0 reload configuration issues


Chevron94:
**What did you do?**
Prometheus sends samples to our remote storage. One time when remote storage is going to down, we see in logs:
level=warn ts=2018-01-19T08:29:08.516631616Z caller=queue_manager.go:485 component=remote msg="Error sending samples to remote storage" count=1 err="Post <storage_url>: getsockopt: no route to host"
If we try to reload prometheus configuration by sending post on <prometheus_url>/-/reload  when remote storage is down, this request is never finished. Looks like prometheus trying to send data to storage again and again, before reload configuration.
**What did you expect to see?**
Prometheus should reload configuration
**What did you see instead? Under which circumstances?**
Configuration wasn't reloaded, no response on POST <prometheus_url>/-/reload request 
**Environment**

* System information:

	Linux 3.10.0-514.26.2.el7.x86_64 x86_64

* Prometheus version:

	2.0.0

* Prometheus configuration file:
```
global:
  scrape_interval: 15s
  scrape_timeout: 10s
  evaluation_interval: 15s
  external_labels:
    monitor: default
...
remote_write:
- url: <remote_storage_url>
  remote_timeout: 30s
  write_relabel_configs:
  - source_labels: [indicatorName]
    separator: ;
    regex: (.+)
    replacement: $1
    action: keep
  queue_config:
    capacity: 100000
    max_shards: 1000
    max_samples_per_send: 100
    batch_send_deadline: 5s
    max_retries: 10
    min_backoff: 30ms
    max_backoff: 100ms
```

* Logs:
```
2018/01/19 08:25:22 Redirected: /-/reload
level=info ts=2018-01-19T08:25:22.313752379Z caller=main.go:490 msg="Loading configuration file" filename=/config/prometheus/default.yml
level=info ts=2018-01-19T08:25:22.319062591Z caller=queue_manager.go:253 component=remote msg="Stopping remote storage..."
level=warn ts=2018-01-19T08:25:23.406480826Z caller=queue_manager.go:485 component=remote msg="Error sending samples to remote storage" count=8 err="Post <remote_storage_url>: getsockopt: no route to host"
level=warn ts=2018-01-19T08:25:26.412528632Z caller=queue_manager.go:485 component=remote msg="Error sending samples to remote storage" count=8 err="Post <remote_storage_url>: getsockopt: no route to host"
[...]

PR #3762, opened 2018-01-30
Support force reloading configuration (return response without waiting of restartiong of all scrape loops)

Chevron94:
There is a case, when you have some loops which can be executed for large amount of time. If at this time prometheus receive reload configuration request, response will be send only when all loops are restarted. 
Adding /-/reload/force endpoint will solve this problem. It will return response after validation rules and prometheus config. Restarting scraping loops will be done in background.

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -325,6 +325,26 @@ func main() {
 		},
 	}
 
+	forceReloaders := []func(cfg *config.Config) error{
+		remoteStorage.ApplyConfig,
+		scrapeManager.ApplyConfigForce,
+		webHandler.ApplyConfig,
+		notifier.ApplyConfig,
+		func(cfg *config.Config) error {
+			// Get all rule files matching the configuration oaths.
+			var files []string
+			for _, pat := range cfg.RuleFiles {
+				fs, err := filepath.Glob(pat)
+				if err != nil {
+					// The only error can be a bad pattern.
+					return fmt.Errorf("error retrieving rule files for %s: %s", pat, err)
+				}
+				files = append(files, fs...)
+			}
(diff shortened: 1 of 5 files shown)


brian-brazil:
I don't think we should have two reload options as that'll cause confusion. Rather that we should have one that works well

Chevron94:
When we trying to reload prometheus configuration, prometheus 2.x returns response too late, because it waits untill all loops will be stopped. In case when some targers are not available it waits timeout for not available targets and only after it returns response (delay could be from several seconds to minutes). The same problem with loops which execution requires large amount of time.
Prometheus 1.x returns response immediatly in any of this cases. 
In #3707 I wrote that in similar situation reload request on 1.x version executes 0,1 second, but on 2.x about 10 seconds. 
Is it correct behaviour?

krasi-georgiev:
@Chevron94 did you try to refactor so that ApplyConfig does the reloading in the background?
Not sure if this will play well with the tsdb, but it might be worth the try.

btw thanks for reporting  #3707  I need to finish few other issues and will try to find the culprit on that one.

Chevron94:
@krasi-georgiev I just tried to move reloading configuration for target loops in separated goroutine and it looks like works well. Fast response and reloading targets in background

krasi-georgiev:
great , should run some tests at scale and hitting the reload many times to make sure this doesn't  brake something in the tsdb.

Chevron94:
Any updates?

krasi-georgiev:
@Chevron94  We are teaming up with [CNCF CI](https://github.com/crosscloudci/crosscloudci/issues/7) to try and build an automated performance tests at scale so I decided to wait and see what happens on that front and test this change there first.

brian-brazil:
This introduces a race I suspect, what if there's two reloads going on concurrently?

krasi-georgiev:
@Chevron94 yep it might be worth checking this with --race enabled.

PR #3770, opened 2018-01-31
Map target infos of endpoints to prometheus meta labels.

cofyc:
See https://github.com/prometheus/prometheus/pull/3748#issuecomment-361817651.


simonpasquier:
:+1:

brian-brazil:
Thanks!

Issue #3791, opened 2018-02-03
provide median() operator or note how to get it with quantile(0.5)


tzz:
Currently, the `quantile` operator is provided, but there's no `median`. The `median` according to https://prometheus.io/docs/practices/histograms/ is trivially `quantile(0.5, ...)`. Can you please consider either adding a new `median` operator to do this, or adding a note to `quantile` in the docs https://prometheus.io/docs/prometheus/latest/querying/operators/ for users that don't know this?

Thank you.

matthiasr:
Since it's so easy to do with the existing quantile function, I don't think
we need a new one.

For mentioning this, can you please make a PR in <
https://github.com/prometheus/docs>?

tzz:
The operators.md reference was not in the docs repo so I created https://github.com/prometheus/prometheus/pull/3792 which should resolve it trivially.

Thanks.

beorn7:
With #3792 closed, I'll try to craft a docs PR that might find more acceptance.

PR #3792, opened 2018-02-04
Mention that quantile(0.5, ...) is the median

tzz:
Resolves https://github.com/prometheus/prometheus/issues/3791#issuecomment-362853500

### DIFF
--- a/docs/querying/operators.md
+++ b/docs/querying/operators.md
@@ -192,7 +192,7 @@ vector of fewer elements with aggregated values:
 * `count_values` (count number of elements with the same value)
 * `bottomk` (smallest k elements by sample value)
 * `topk` (largest k elements by sample value)
-* `quantile` (calculate φ-quantile (0 ≤ φ ≤ 1) over dimensions)
+* `quantile` (calculate φ-quantile (0 ≤ φ ≤ 1) over dimensions; use 0.5 to get the median)
 
 These operators can either be used to aggregate over **all** label dimensions
 or preserve distinct dimensions by including a `without` or `by` clause.


docs/querying/operators.md
```diff
@@ -192,7 +192,7 @@ vector of fewer elements with aggregated values:
 * `count_values` (count number of elements with the same value)
 * `bottomk` (smallest k elements by sample value)
 * `topk` (largest k elements by sample value)
-* `quantile` (calculate φ-quantile (0 ≤ φ ≤ 1) over dimensions)
+* `quantile` (calculate φ-quantile (0 ≤ φ ≤ 1) over dimensions; use 0.5 to get the median)
```
brian-brazil: This should be be no more than an aside, it shouldn't clutter up the main table.

tzz: OK, but you're not telling me where to put it. Could you tell me where you want it?

As an alternative, considering it's 3-4 words, you can make the edit directly and close this.

jamtur01:
Is the aside needed at all?

matthiasr:
Yes, I think we could do better in explaining what the heck a φ-quantile is. We don't give any explanation at all, and no link; googling "φ-quantile" drops me straight into Math Wikipedia but even if [this article](https://en.wikipedia.org/wiki/Quantile_function) were readable to non-mathematicians it does not even mention it. None of the other results that come before this very page offer a concise explanation either.

φ-quantile Google Search

If you know a good explanation, we can link it instead, but if there isn't one I think we should create it.

matthiasr:
… looking again, it's not this page, it's [this one](https://prometheus.io/docs/practices/histograms/#quantiles) and we already have an explanation, which even mentions the Median, we're just not linking it. Should we just turn φ-quantile into a link there?

brian-brazil:
I think a brief mention that 0.95 quantile = 95th percentile on this page would make sense, that other page is quite technical.

matthiasr:
@brian-brazil what do you think about just adding some examples? "To calculate the median request latency …" "To calculate the 95th percentile …". That would also give a place to mention the "first rate then quantile".

matthiasr:
and then we could link to the other page "for more details see".

tzz:
Please go ahead and make the changes directly, I'm happy to close this and #3791 in favor of a practical example as suggested by @matthiasr

jamtur01:
How about adding a quantile example to the end of the examples list? I'd be happy to do it.

jamtur01:
Although a cross-pollination question - to continue with the `http_requests_total` metric being used in those examples seems odd with `quantile`. Can I other metric or does that disrupt the flow?

brian-brazil:
> @brian-brazil what do you think about just adding some examples? "

The reason the count_values and topk examples are there as I felt my prose was a bit hard to grok, so added one numeric and one string example. In reality I'd expect the use of the quantile aggregator to be relatively uncommon.

> "To calculate the median request latency …"

The histogram_quantile docs would be the place to do that, they already have percentile examples - an aside there mentioning median maybe?

tzz:
ping... this is a trivial fix, just needs a placement decision.

TylerHendrickson:
Considering that a google search for `prometheus median` leads to this PR, it seems worth pointing out that there is some objective proof that adding text explaining that "median = .5 quantile" here would be quite useful for a number of users. 

Since a placement decision is the only blocker for merging this PR, is it possible to either receive guidance on proper placement/wording or have the PR merged as-is? 

If this can simply be merged, perhaps the concern over placement can be raised separately in its own [enhancement  issue](https://github.com/prometheus/prometheus/issues?q=is%3Aopen+is%3Aissue+label%3Akind%2Fenhancement), rather than continuing to hold up a trivial but useful change.

brian-brazil:
I don't think we should be merging PRs that we know are suboptimal. Would you like to propose a new PR?

gouthamve:
Hi thanks for this PR, but we decided to close this in the bug scrub as we already mention the median here https://prometheus.io/docs/practices/histograms/#quantiles

beorn7:
I'll think about a good way to link from the function to an explanation. PR imminent…

PR #3798, opened 2018-02-05
initilise `ss` var only once - reduce mem usage a bit.

krasi-georgiev:


brian-brazil:
I'd expect `ss` to be on the stack, have you benchmarks to indicate this makes a difference?

krasi-georgiev:
yes , I did some mem pprof a while back when I was trying to improve promql , but can run again if you have suspicions.

brian-brazil:
If you don't mind, as I'd expect this to be something that Go would handle automatically. Would be worth a feature request to Go if it doesn't put this on the stack.

krasi-georgiev:
I am not able to replicate and your suggestion seems right so will close for now and will reopen if I see it again next time I pprof promql.

Issue #3649, opened 2018-01-03
iincrease


roidelapluie:
I would like to have a iincrease function; that would be like idelta, but for counters.

brian-brazil:
Can you explain your use case for this? `idelta` exists to allow implementation of custom rate&friends via recording rules, and the features to make that possible don't all exist yet. It's not meant to be used for anything currently.

roidelapluie:
I want to see the Volume of data going on a SNMP interface ; All I can do is irate ; which is only the rate ; not the total volume. And idelta by purpose will not handle counters resets correctly.

brian-brazil:
> not the total volume

Nothing in Prometheus can calculate the total volume, as history is limited. Are you sure that's what you're looking for?

roidelapluie:
Yes idelta will do. Thanks.

PR #3802, opened 2018-02-06
Implement a iincrease function

roidelapluie:
Implements a iincrease function, similar to idelta but breaks in
monotonicity are automatically adjusted for.


brian-brazil:
This function makes no sense to me, you want `idelta`. See #3649

PR #3805, opened 2018-02-06
Meta labels sd 3693

### LINKED ISSUE #3693: No refresh of file service discovery targets if only __meta labels changed
**What did you do?**

Using file_sd_config as discovery mechanism for a target group generated by a custom docker swarm service discovery (similar to https://github.com/ContainerSolutions/prometheus-swarm-discovery). When updating the targets file, I couldn't see any update in the targets list of the Prometheus web interface. My first assumption was that there are issues with fsnotify inside of container volumes, but this was not the case. The issue I faced was that there is no updated configuration visible in the targets view when only __meta labels change.

After some investigation, I found the reason for the omitted updates in the code where [the list of scrape tasks is synced](https://github.com/prometheus/prometheus/blob/b20a1b1b1b6e2818991cec45fd0d11646449b250/retrieval/scrape.go#L290) using a [hash function of the target](https://github.com/prometheus/prometheus/blob/b20a1b1b1b6e2818991cec45fd0d11646449b250/retrieval/target.go#L76). This hash is calculated using the `url` and the `labels` but not the `discoveredLabels` where these __meta labels are part of.

```
// hash returns an identifying hash for the target.
func (t *Target) hash() uint64 {
	h := fnv.New64a()
	h.Write([]byte(fmt.Sprintf("%016d", t.labels.Hash())))
	h.Write([]byte(t.URL().String()))

	return h.Sum64()
}
```

Is it an intended behavior that an update of __meta labels does not trigger an update of the scrape targets - probably to prevent too frequent updates? My assumption was that I can use these __meta labels to add all information I get when talking to the Docker API and so I would be able to make use of them during the relabeling phase. In my opinion, it can be a regular use case that service labels are updated without any further changes of the target urls or regular labels. If one should not use these, which would be a better "namespace" for these custom labels?

Even if the targets and regular labels remain the same, the configuration should reflect the change/addition/removal of __meta labels.

The configuration update was not applied. The targets view still lists the old combination of __meta labels.

**Environment**

Prometheus 2.0.0 deployed using Docker Swarm - and also reproduced with a local setup on a pure Ubuntu Server 16.04 LTS VM.

* System information:

Host: Ubuntu 16.04.3 LTS
Docker 17.12.0-ce
Linux 4.4.0-108-generic x86_64

* Prometheus version:

2.0.0

```
scrape_configs:
  - job_name: 'services'
    file_sd_configs:
      - files:
        - "/xxx/targets.json"
```

krasi-georgiev:
fixes #3693
added an additional function to update the `DiscoveredLabels` even when the new labels don't affect the target hash.
The `hash` is calculated  form the target labels after the relabelling takes place so if the new labels don't affect the hash than `DiscoveredLabels` were never updated.

*The  `DiscoveredLabels` are used in the Service discovery page.

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -301,6 +301,10 @@ func (sp *scrapePool) sync(targets []*Target) {
 			sp.loops[hash] = l
 
 			go l.run(interval, timeout, nil)
+		} else {
+			// Need to keep the most updated labels information
+			// for displaying it in the Service Discovery web page.
+			sp.targets[hash].SetDiscoveredLabels(t.DiscoveredLabels())
 		}
 	}
 
--- a/scrape/scrape_test.go
+++ b/scrape/scrape_test.go
@@ -96,6 +96,40 @@ func TestDroppedTargetsList(t *testing.T) {
 	}
 }
 
+// TestDiscoveredLabelsUpdate checks that DiscoveredLabels are updated
+// even when new labels don't affect the target `hash`.
+func TestDiscoveredLabelsUpdate(t *testing.T) {
+
+	sp := &scrapePool{}
+	// These are used when syncing so need this to avoid a panic.
+	sp.config = &config.ScrapeConfig{
+		ScrapeInterval: model.Duration(1),
+		ScrapeTimeout:  model.Duration(1),
+	}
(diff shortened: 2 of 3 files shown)


brian-brazil:
Thanks!

PR #3811, opened 2018-02-07
time() return milliseconds

### LINKED ISSUE #3808: time() loses fractional part of the time
The `time()` function only returns integers, it should be returning decimals with up to millisecond resolution.

krasi-georgiev:
fixes #3808
@brian-brazil is this what you meant?

TODO: 
~- [ ] update website docs.~

### DIFF
--- a/promql/functions.go
+++ b/promql/functions.go
@@ -38,7 +38,7 @@ type Function struct {
 // === time() float64 ===
 func funcTime(ev *evaluator, args Expressions) Value {
 	return Scalar{
-		V: float64(ev.Timestamp / 1000),
+		V: float64(ev.Timestamp) / 1000,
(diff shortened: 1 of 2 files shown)


brian-brazil:
No the time must still be in seconds, but we're losing the milliseconds from the time. This is due to doing the float conversion too late.

krasi-georgiev:
Thanks PR updates

brian-brazil:
Can you add a test?

brian-brazil:
You should be able to test this with standard promql tests rather than needing to do something bespoke.

krasi-georgiev:
That is nice , didn't notice it. 
test amended.

brian-brazil:
Thanks!

PR #3813, opened 2018-02-07
[Fix 3789]: storing/loading multiple graphs from URL

### LINKED ISSUE #3789: Storing/loading multiple graphs from URL doesn't work anymore
When adding more than one graph on the `/graph` page, only the first one is stored in the URL. Also, when trying to open a URL like [http://localhost:9090/graph?g0.range_input=1h&g0.expr=rate(node_cpu%5B1m%5D)&g0.tab=0&g1.range_input=1h&g1.expr=rate(http_request_duration_microseconds_count%5B1m%5D)&g1.tab=0](http://localhost:9090/graph?g0.range_input=1h&g0.expr=rate(node_cpu%5B1m%5D)&g0.tab=0&g1.range_input=1h&g1.expr=rate(http_request_duration_microseconds_count%5B1m%5D)&g1.tab=0), the second entry is halfway rendered, but not executed (and disappears from the URL).

According to `git bisect` this regression was introduced by https://github.com/prometheus/prometheus/commit/a8ff643464ee5f3cf97f1c55c0aab44796f09167. @lovisas would you be able to take a look?

lovisas:
Fixes #3789 
This PR has two commits. 

1) The first one fixes a JS error which occurred when the page was refreshed and the Query History checkbox was checked. It attempted to update the typeahead metric source without the typeahead having been initialized. 

2) The second one adds a step to update the URL when queries were submitted by pressing enter. 

I've tested this in various ways and as far as I can tell this issue is fixed. Please try it out and let me know what kinds of things you run into!


lovisas:
When testing this what I found was that when queries were submitted by pressing Enter, the url was not updated. When submitting the query by pressing the `Execute` button, the url did update. I've tested this with the query history box toggled off/on and with different combinations of queries. Let me know if this solves the issue on your end. 

There is some room for improvement in the `Prometheus.Graph.prototype.initialize` function, but given the upcoming rewrite and the high likelihood of introducing regressions, I left it as is and just added a function call. 

cc @juliusv

lovisas:
I saw that the 2.2 pre-release was committed recently, this bug fix should probably go in with the next release if possible since it's a very visible bug. Does anyone have capacity to test/review this? 

ping @juliusv @brancz @brian-brazil

juliusv:
@lovisas Oh sorry, this slipped through. I'll give it a good testing tomorrow!

lovisas:
No problem! Thanks :)

juliusv:
Also restarted the Travis build, it was failing due to an unrelated flaky test :-/

juliusv:
:+1: I wasn't able to break it anymore :)

Thanks for the fix!

brancz:
We'll make sure this lands in the 2.2 release, thanks for taking care of this @lovisas!

Issue #3711, opened 2018-01-19
consul: more efficient tag and service filtering


iksaif:
This is related to:
* https://github.com/hashicorp/consul/pull/2549
* https://github.com/prometheus/prometheus/pull/2028
* https://github.com/prometheus/prometheus/issues/2285
* https://github.com/prometheus/prometheus/pull/2066
* https://github.com/prometheus/prometheus/issues/3452
* https://github.com/prometheus/prometheus/issues/771
* And some others

The official answer so far as been "use relabeling", but doing that ends up opening one TCP connection per service (it looks like all the services are being watched). This happens because if you don't specify a service name, and drop using relabeling `shouldWatch()`  will always return True and all services will be watched (which is currently very inefficient with consul, and clearly non-necessary).

According to https://www.consul.io/api/catalog.html#list-services, we still can only filter on `dc` and `node-service`. As discussed in https://github.com/hashicorp/consul/pull/2549 they would like to also add `service-meta` here, but that hasn't been done (and that was almost a year ago).

Currently the discovery/consul/consul.go simply watches the Catalog without any restrictions, *but* `Services()` returns both the service names and tags, so we can start to do some filtering here based on the tags (example in https://github.com/hashicorp/consul/blob/b3292d13fb8bbc8b14b2a1e2bbae29c6e105b8f4/command/catalog/list/services/catalog_list_services.go#L85)

`catalog.Service()` can take an also optional `tag` argument that could be used to do some filtering server-side, so we can use that too to get notified only when necessary.

I (or somebody else from my team) will try to get a PR ready next week as it is quite annoying to have thousands of connections for nothing, and relabeling isn't that obvious to do this.

iksaif:
cc: @vguerci, @Thib17, @karthimohan, @mchataigner

brian-brazil:
I understand where you are coming from however as noted on several of those issues, Prometheus uses service discovery that exists - we don't invent new one or variants thereof. If there are filtering options that Consul supports natively then they can be considered for addition.

Service-meta as is discussed on https://github.com/hashicorp/consul/pull/2549 sounds like a better way to handle this (as would more efficient watching), and I am against adding short-term hacks to Prometheus which will ultimately lead to impedance mismatches. The place to tackle this problem is at its source.

Put another way, we already have one custom way of doing filtering in Prometheus (relabelling) I don't think we should be adding more.

iksaif:
[...]

Issue #2191, opened 2016-11-15
Consider coalescing identical SD definitions


brian-brazil:
There may be cases where users have multiple different scrape configs that share a SD configuration, but where other settings such as relabelling are different.

We have been slowly adding the ability to filter down at the SD level in an ad-hoc way for each SD, but given the number of things that a Prometheus is likely monitoring I think there's another potential approach to keeping load on the SD sane.

If there's two identical SD configs, let's only instantiate it once and then feed it into each required scrape config. This means a naieve config will have constant load on SD relative to the number of scrape configs, and avoid a good chunk of users having to worry about SD performance.

fabxc:
:+1: that generalisation makes a lot of sense. I guess that's good enough
for k8s as well for single roles.

brian-brazil:
We've seen already one user concerned about this for EC2, and given Azure's very low rate limits I imagine it'd be really useful there too.

grandbora:
Hi all,

I want to take a stab at this issue. I dug into the code a bit and formalized a vague idea of how to do this. Below are the details of what I am thinking and couple of questions related to that.

Current flow:
* prometheus starts
  * main instantiates and runs a [targetManager](https://github.com/prometheus/prometheus/blob/master/retrieval/targetmanager.go#L31), and applies the config
  * targetManager `reload` [loops over each scrapeConfig](https://github.com/prometheus/prometheus/blob/master/retrieval/targetmanager.go#L101)
    * a [targetSet is created](https://github.com/prometheus/prometheus/blob/master/retrieval/targetmanager.go#L107) for each scrapeConfig
    * [UpdateProviders](https://github.com/prometheus/prometheus/blob/master/retrieval/targetmanager.go#L128) is called for each targetSet with the ProvidersFromConfig

in UpdateProviders two things are done:
* [clean up](https://github.com/prometheus/prometheus/blob/master/discovery/discovery.go#L232-L244) (cancel existing providers, remove all the target groups)
* [start listening new providers](https://github.com/prometheus/prometheus/blob/master/discovery/discovery.go#L249-L288) one by one for each targetSet

Proposed flow
First I plan to break apart the steps of UpdateProviders I mentioned above. There will be two methods;
* CleanUp(Reset): does the clean up step for each targetSet of the targetManager
* ListenChanges: starts listening a targetProvider and apply changes to all the targetSets that rely on it.

With the changes above the flow I would like to implement is:

* prometheus starts
[...]

PR #3814, opened 2018-02-08
consul: improve consul service discovery

iksaif:
Related to #3711

- Add the ability to filter by tag and node-meta in an efficient way (`/catalog/services`
  allow filtering by node-meta, and returns a `map[string]string` or `service`->`tags`).
  Tags and nore-meta are also used in `/catalog/service` requests.
- Do not require a call to the catalog if services are specified by name. This is important
  because on large cluster `/catalog/services` changes all the time.
- Add `allow_stale` configuration option to do stale reads. Non-stale
  reads can be costly, even more when you are doing them to a remote
  datacenter with 10k+ targets over WAN (which is common for federation).
- Add `refresh_interval` to minimize the strain on the catalog and on the
  service endpoint. This is needed because of that kind of behavior from
  consul: https://github.com/hashicorp/consul/issues/3712 and because a catalog
  on a large cluster would basically change *all* the time. No need to discover
  targets in 1sec if we scrape them every minute.
- Added plenty of unit tests.

Benchmarks
----------

```yaml
scrape_configs:

- job_name: prometheus
  scrape_interval: 60s
  static_configs:
    - targets: ["127.0.0.1:9090"]

- job_name: "observability-by-tag"
  scrape_interval: "60s"
  metrics_path: "/metrics"
  consul_sd_configs:
    - server: consul.service.par.consul.prod.crto.in:8500
      tag: marathon-user-observability  # Used in After
      refresh_interval: 30s             # Used in After+delay
  relabel_configs:
    - source_labels: [__meta_consul_tags]
      regex: ^(.*,)?marathon-user-observability(,.*)?$
      action: keep

- job_name: "observability-by-name"
  scrape_interval: "60s"
  metrics_path: "/metrics"
  consul_sd_configs:
    - server: consul.service.par.consul.prod.crto.in:8500
      services:
        - observability-cerebro
        - observability-portal-web

- job_name: "fake-fake-fake"
  scrape_interval: "15s"
  metrics_path: "/metrics"
  consul_sd_configs:
    - server: consul.service.par.consul.prod.crto.in:8500
      services:
        - fake-fake-fake
```

Note: tested with ~1200 services, ~5000 nodes.

| Resource | Empty | Before | After | After + delay |
| -------- |:-----:|:------:|:-----:|:-------------:|
|/service-discovery size|5K|85MiB|27k|27k|27k|
|`go_memstats_heap_objects`|100k|**1M**|120k|110k|
|`go_memstats_heap_alloc_bytes`|24MB|**150MB**|28MB|27MB|
|`rate(go_memstats_alloc_bytes_total[5m])`|0.2MB/s|**28MB/s**|2MB/s|0.3MB/s|
|`rate(process_cpu_seconds_total[5m])`|0.1%|**15%**|2%|0.01%|
|`process_open_fds`|16|*1236*|22|22|
|`rate(prometheus_sd_consul_rpc_duration_seconds_count{call="services"}[5m])`|~0|1|1|**0.03**|
|`rate(prometheus_sd_consul_rpc_duration_seconds_count{call="service"}[5m])`|0.1|**80**|0.5|0.5|
|`prometheus_target_sync_length_seconds{quantile="0.9",scrape_job="observability-by-tag"}`|N/A|200ms|0.2ms|0.2ms|
|Network bandwidth|~10kbps|**~2.8Mbps**|**~1.6Mbps**|~10kbps|

Filtering by tag using relabel_configs uses **100kiB and 23kiB/s per service per job** and quite a lot of CPU. Also sends and additional *1Mbps* of traffic to consul.
Being a little bit smarter about this reduces the overhead quite a lot.
Limiting the number of `/catalog/services` queries per second almost removes the overhead of service discovery.

### DIFF
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -84,7 +84,7 @@ var (
 		Scheme:          "http",
 		Server:          "localhost:8500",
 		AllowStale:      true,
-		RefreshInterval: model.Duration(15 * time.Second),
+		RefreshInterval: model.Duration(watchTimeout),
 	}
 )
 
@@ -329,7 +329,8 @@ func (d *Discovery) watchServices(ctx context.Context, ch chan<- []*targetgroup.
 		AllowStale: d.allowStale,
 		NodeMeta:   d.watchedNodeMeta,
 	})
-	rpcDuration.WithLabelValues("catalog", "services").Observe(time.Since(t0).Seconds())
+	elapsed := time.Since(t0)
+	rpcDuration.WithLabelValues("catalog", "services").Observe(elapsed.Seconds())
 
 	if err != nil {
 		level.Error(d.logger).Log("msg", "Error refreshing service list", "err", err)
@@ -376,8 +377,9 @@ func (d *Discovery) watchServices(ctx context.Context, ch chan<- []*targetgroup.
 			}
 		}
 	}
-
-	time.Sleep(d.refreshInterval)
+	if elapsed < d.refreshInterval {
+		time.Sleep(d.refreshInterval - elapsed)
+	}
 	return nil
 }
 
@@ -425,7 +427,8 @@ func (srv *consulService) watch(ctx context.Context, ch chan<- []*targetgroup.Gr
 			AllowStale: srv.discovery.allowStale,
 			NodeMeta:   srv.discovery.watchedNodeMeta,
 		})
-		rpcDuration.WithLabelValues("catalog", "service").Observe(time.Since(t0).Seconds())
+		elapsed := time.Since(t0)
+		rpcDuration.WithLabelValues("catalog", "service").Observe(elapsed.Seconds())
 
 		// Check the context before potentially falling in a continue-loop.
 		select {
@@ -497,6 +500,8 @@ func (srv *consulService) watch(ctx context.Context, ch chan<- []*targetgroup.Gr
 			return
 		case ch <- []*targetgroup.Group{&tgroup}:
 		}
-		time.Sleep(srv.discovery.refreshInterval)
+		if elapsed < srv.discovery.refreshInterval {
+			time.Sleep(srv.discovery.refreshInterval - elapsed)
+		}
 	}
 }
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -100,9 +100,9 @@ type SDConfig struct {
 
 	// See https://www.consul.io/docs/internals/consensus.html#consistency-modes,
 	// stale reads are a lot cheaper and are a necessity if you have >5k targets.
-	AllowStale bool `yaml:"allow_stale,omitempty"`
+	AllowStale bool `yaml:"allow_stale"`
 	// By default use blocking queries (https://www.consul.io/api/index.html#blocking-queries)
-	// but allow users to delay updates if necessary. This can be useful because of "bugs" like
+	// but allow users to throttle updates if necessary. This can be useful because of "bugs" like
 	// https://github.com/hashicorp/consul/issues/3712 which cause an un-necessary
 	// amount of requests on consul.
 	RefreshInterval model.Duration `yaml:"refresh_interval,omitempty"`
@@ -267,7 +267,14 @@ func (d *Discovery) getDatacenter() error {
 		return err
 	}
 
-	d.clientDatacenter = info["Config"]["Datacenter"].(string)
+	dc, ok := info["Config"]["Datacenter"].(string)
+	if !ok {
+		err := fmt.Errorf("Invalid value '%v' for Config.Datacenter", info["Config"]["Datacenter"])
+		level.Error(d.logger).Log("msg", "Error retrieving datacenter name", "err", err)
+		return err
+	}
+
+	d.clientDatacenter = dc
 	return nil
 }
 
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -299,7 +299,7 @@ func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
 		} else {
 			// We only have fully defined services.
 			for _, name := range d.watchedServices {
-				d.watchService(name, ctx, ch)
+				d.watchService(ctx, name, ch)
 			}
 			// Wait for cancellation.
 			<-ctx.Done()
@@ -350,7 +350,7 @@ func (d *Discovery) watchServices(ctx context.Context, ch chan<- []*targetgroup.
 		}
 
 		wctx, cancel := context.WithCancel(ctx)
-		d.watchService(name, wctx, ch)
+		d.watchService(wctx, name, ch)
 		services[name] = cancel
 	}
 
@@ -386,7 +386,7 @@ type consulService struct {
 }
 
 // Start watching a service.
-func (d *Discovery) watchService(name string, ctx context.Context, ch chan<- []*targetgroup.Group) {
+func (d *Discovery) watchService(ctx context.Context, name string, ch chan<- []*targetgroup.Group) {
 	srv := &consulService{
 		discovery: d,
 		client:    d.client,
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -84,7 +84,7 @@ var (
 		Scheme:          "http",
 		Server:          "localhost:8500",
 		AllowStale:      true,
-		RefreshInterval: model.Duration(0 * time.Second),
+		RefreshInterval: model.Duration(15 * time.Second),
 	}
 )
 
@@ -97,11 +97,12 @@ type SDConfig struct {
 	Scheme       string             `yaml:"scheme,omitempty"`
 	Username     string             `yaml:"username,omitempty"`
 	Password     config_util.Secret `yaml:"password,omitempty"`
+
 	// See https://www.consul.io/docs/internals/consensus.html#consistency-modes,
 	// stale reads are a lot cheaper and are a necessity if you have >5k targets.
 	AllowStale bool `yaml:"allow_stale,omitempty"`
-	// By default use blocking queries () but allow users to delay
-	// updates if necessary. This can be useful because of "bugs" like
+	// By default use blocking queries (https://www.consul.io/api/index.html#blocking-queries)
+	// but allow users to delay updates if necessary. This can be useful because of "bugs" like
 	// https://github.com/hashicorp/consul/issues/3712 which cause an un-necessary
 	// amount of requests on consul.
 	RefreshInterval model.Duration `yaml:"refresh_interval,omitempty"`
@@ -110,8 +111,9 @@ type SDConfig struct {
 	// The list of services for which targets are discovered.
 	// Defaults to all services if empty.
 	Services []string `yaml:"services"`
-	// An optional tag used to filter instances inside a service.
-	Tag string `yaml:"tag,omitempty"`
+	// An optional tag used to filter instances inside a service. A single tag is supported
+	// here to match the Consul API.
+	ServiceTag string `yaml:"tag,omitempty"`
 	// Desired node metadata.
 	NodeMeta map[string]string `yaml:"node_meta,omitempty"`
 
@@ -203,7 +205,7 @@ func NewDiscovery(conf *SDConfig, logger log.Logger) (*Discovery, error) {
 		clientConf:       clientConf,
 		tagSeparator:     conf.TagSeparator,
 		watchedServices:  conf.Services,
-		watchedTag:       conf.Tag,
+		watchedTag:       conf.ServiceTag,
 		watchedNodeMeta:  conf.NodeMeta,
 		allowStale:       conf.AllowStale,
 		refreshInterval:  time.Duration(conf.RefreshInterval),
@@ -234,6 +236,8 @@ func (d *Discovery) shouldWatchFromName(name string) bool {
 }
 
 // shouldWatch returns whether the service of the given name should be watched based on its tags.
+// This gets called when the user doesn't specify a list of services in order to avoid watching
+// *all* services. Details in https://github.com/prometheus/prometheus/pull/3814
 func (d *Discovery) shouldWatchFromTags(tags []string) bool {
 	// If there's no fixed set of watched tags, we watch everything.
 	if d.watchedTag == "" {
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -299,7 +299,6 @@ func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
 		} else {
 			// We only have fully defined services.
 			for _, name := range d.watchedServices {
-				ctx, _ := context.WithCancel(ctx)
 				d.watchService(name, ctx, ch)
 			}
 			// Wait for cancellation.
@@ -309,9 +308,9 @@ func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
 	}
 }
 
-// Watch the catalog for services we would like to watch. This is only necessary
-// for services filtered by tag. If we already have the name it is more efficient to
-// watch nodes the service directly.
+// Watch the catalog for new services we would like to watch. This is called only
+// when we don't know yet the names of the services and need to ask Consul the
+// entire list of services.
 func (d *Discovery) watchServices(ctx context.Context, ch chan<- []*targetgroup.Group, lastIndex *uint64, services map[string]func()) error {
 	catalog := d.client.Catalog()
 	level.Debug(d.logger).Log("msg", "Watching services", "tag", d.watchedTag)
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -263,6 +263,7 @@ func (d *Discovery) getDatacenter() error {
 	info, err := d.client.Agent().Self()
 	if err != nil {
 		level.Error(d.logger).Log("msg", "Error retrieving datacenter name", "err", err)
+		rpcFailuresCount.Inc()
 		return err
 	}
 
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -80,9 +80,11 @@ var (
 
 	// DefaultSDConfig is the default Consul SD configuration.
 	DefaultSDConfig = SDConfig{
-		TagSeparator: ",",
-		Scheme:       "http",
-		Server:       "localhost:8500",
+		TagSeparator:    ",",
+		Scheme:          "http",
+		Server:          "localhost:8500",
+		AllowStale:      true,
+		RefreshInterval: model.Duration(0 * time.Second),
 	}
 )
 
@@ -95,9 +97,23 @@ type SDConfig struct {
 	Scheme       string             `yaml:"scheme,omitempty"`
 	Username     string             `yaml:"username,omitempty"`
 	Password     config_util.Secret `yaml:"password,omitempty"`
+	// See https://www.consul.io/docs/internals/consensus.html#consistency-modes,
+	// stale reads are a lot cheaper and are a necessity if you have >5k targets.
+	AllowStale bool `yaml:"allow_stale,omitempty"`
+	// By default use blocking queries () but allow users to delay
+	// updates if necessary. This can be useful because of "bugs" like
+	// https://github.com/hashicorp/consul/issues/3712 which cause an un-necessary
+	// amount of requests on consul.
+	RefreshInterval model.Duration `yaml:"refresh_interval,omitempty"`
+
+	// See https://www.consul.io/api/catalog.html#list-services
 	// The list of services for which targets are discovered.
 	// Defaults to all services if empty.
 	Services []string `yaml:"services"`
+	// An optional tag used to filter instances inside a service.
+	Tag string `yaml:"tag,omitempty"`
+	// Desired node metadata.
+	NodeMeta map[string]string `yaml:"node_meta,omitempty"`
 
 	TLSConfig config_util.TLSConfig `yaml:"tls_config,omitempty"`
 	// Catches all undefined fields and must be empty after parsing.
@@ -138,6 +154,10 @@ type Discovery struct {
 	clientDatacenter string
 	tagSeparator     string
 	watchedServices  []string // Set of services which will be discovered.
+	watchedTag       string   // A tag used to filter instances of a service.
+	watchedNodeMeta  map[string]string
+	allowStale       bool
+	refreshInterval  time.Duration
 	logger           log.Logger
 }
 
@@ -183,6 +203,10 @@ func NewDiscovery(conf *SDConfig, logger log.Logger) (*Discovery, error) {
 		clientConf:       clientConf,
 		tagSeparator:     conf.TagSeparator,
 		watchedServices:  conf.Services,
+		watchedTag:       conf.Tag,
+		watchedNodeMeta:  conf.NodeMeta,
+		allowStale:       conf.AllowStale,
+		refreshInterval:  time.Duration(conf.RefreshInterval),
 		clientDatacenter: clientConf.Datacenter,
 		logger:           logger,
 	}
@@ -190,11 +214,17 @@ func NewDiscovery(conf *SDConfig, logger log.Logger) (*Discovery, error) {
 }
 
 // shouldWatch returns whether the service of the given name should be watched.
-func (d *Discovery) shouldWatch(name string) bool {
+func (d *Discovery) shouldWatch(name string, tags []string) bool {
+	return d.shouldWatchFromName(name) && d.shouldWatchFromTags(tags)
+}
+
+// shouldWatch returns whether the service of the given name should be watched based on its name.
+func (d *Discovery) shouldWatchFromName(name string) bool {
 	// If there's no fixed set of watched services, we watch everything.
 	if len(d.watchedServices) == 0 {
 		return true
 	}
+
 	for _, sn := range d.watchedServices {
 		if sn == name {
 			return true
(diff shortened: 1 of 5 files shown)


pierresouchay:
+1

discovery/consul/consul.go
```diff
@@ -95,9 +97,23 @@ type SDConfig struct {
 	Scheme       string             `yaml:"scheme,omitempty"`
 	Username     string             `yaml:"username,omitempty"`
 	Password     config_util.Secret `yaml:"password,omitempty"`
+	// See https://www.consul.io/docs/internals/consensus.html#consistency-modes,
+	// stale reads are a lot cheaper and are a necessity if you have >5k targets.
+	AllowStale bool `yaml:"allow_stale,omitempty"`
+	// By default use blocking queries () but allow users to delay
```
brian-brazil: Might it make sense to always use this with a value like 15s? Internally we already throttle all SD to 5s.

What is the () for?

discovery/consul/consul.go
```diff
@@ -203,21 +233,46 @@ func (d *Discovery) shouldWatch(name string) bool {
 	return false
 }
 
+// shouldWatch returns whether the service of the given name should be watched based on its tags.
+func (d *Discovery) shouldWatchFromTags(tags []string) bool {
```
brian-brazil: This looks like filtering on our side at a first glance, which wouldn't be kosher.

It should be made clear in the comments that this is an optimisation on top of the option passed to the Service call, not a duplication of what should be in relabelling. We're going to get questions for example as to why only one tag is supported.

discovery/consul/consul.go
```diff
@@ -226,92 +281,138 @@ func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
 		default:
 		}
 
+		// Get the local datacenter first, if necessary.
+		err := d.getDatacenter()
 		if err != nil {
-			level.Error(d.logger).Log("msg", "Error refreshing service list", "err", err)
-			rpcFailuresCount.Inc()
 			time.Sleep(retryInterval)
```
brian-brazil: We should be logging and incrementing here

discovery/consul/consul.go
```diff
@@ -95,9 +97,23 @@ type SDConfig struct {
...
+	// https://github.com/hashicorp/consul/issues/3712 which cause an un-necessary
+	// amount of requests on consul.
+	RefreshInterval model.Duration `yaml:"refresh_interval,omitempty"`
+
+	// See https://www.consul.io/api/catalog.html#list-services
 	// The list of services for which targets are discovered.
 	// Defaults to all services if empty.
 	Services []string `yaml:"services"`
+	// An optional tag used to filter instances inside a service.
+	Tag string `yaml:"tag,omitempty"`
```
brian-brazil: ServiceTag would be clearer.

brian-brazil:
I'm wondering if we could get away with just RefreshInterval, the other settings will be less useful/counter-productive once #2191 is in.

discovery/consul/consul.go
iksaif: Done

discovery/consul/consul.go
iksaif: * Set default to 15s
* The `()` was to put a link to the doc

discovery/consul/consul.go
iksaif: I updated the documentation and comments. Hope it's more clear now.

discovery/consul/consul.go
iksaif: These got moved in getDatacenter(), I added a `rpcFailuresCount.Inc()`

iksaif:
Update the patchs

> The config documentation needs updating.

This should be done now. Please tell me if this is still unclear.

> Mixing in the core change plus several optimisations on top of it makes this a bit challenging to review.

Sorry for that :/

> I'm wondering if we could get away with just RefreshInterval, the other settings will be less useful/counter-productive once #2191 is in.

I guess that would still be an improvement. But I believe exposing the consul filtering capabilities has a real value because there is a non-significant overhead when using relabeling to do the filtering (as seen on the table).

discovery/consul/consul.go
```diff
@@ -226,92 +286,138 @@ func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
...
-				level.Error(d.logger).Log("msg", "Error retrieving datacenter name", "err", err)
-				time.Sleep(retryInterval)
-				continue
+		if len(d.watchedServices) == 0 || d.watchedTag != "" {
+			// We need to watch the catalog.
+			d.watchServices(ctx, ch, &lastIndex, services)
+		} else {
+			// We only have fully defined services.
+			for _, name := range d.watchedServices {
+				ctx, _ := context.WithCancel(ctx)
```
cstyan: do we need the cancel func here? if not we can just remove this line, seems like we're creating a new cancel func for each service in watchServices

discovery/consul/consul_test.go
```diff
@@ -25,22 +36,172 @@ func TestConfiguredService(t *testing.T) {
...
+		case "/v1/catalog/services?wait=30000ms":
+			response = ServicesTestAnswer
+		case "/v1/catalog/services?index=1&node-meta=rack_name%3A2304&stale=&wait=30000ms":
+			time.Sleep(5 * time.Second)
+			response = ServicesTestAnswer
+		case "/v1/catalog/services?index=1&wait=30000ms":
+			time.Sleep(5 * time.Second)
+			response = ServicesTestAnswer
+		default:
+			t.Error("Unhandeld consul call: %s", r.URL)
```
cstyan: `t.Errorf`?

discovery/consul/consul.go
```diff
@@ -226,92 +286,138 @@ func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
...
-				},
-				tagSeparator: d.tagSeparator,
-				logger:       d.logger,
-			}
-
-			wctx, cancel := context.WithCancel(ctx)
-			go srv.watch(wctx, ch)
+// Watch the catalog for services we would like to watch. This is only necessary
+// for services filtered by tag. If we already have the name it is more efficient to
+// watch nodes the service directly.
```
cstyan: this comment is a bit unclear imo (unless I'm misunderstanding what this function is doing), can we change it to something like `Watch the catalog for changes in instances of the services we would like to watch. Adds/removes instances of the service found in the catalog to Discovery's targets.`?

iksaif: `service instance` (node?) is unclear too. I changed it to:

```
// Watch the catalog for new services we would like to watch. This is called only                                                                                                                                 
// when we don't know yet the names of the services and need to ask Consul the                                                                                                                                    
// entire list of services.   
```

discovery/consul/consul_test.go
iksaif: Good catch

discovery/consul/consul.go
iksaif: you're right, this isn't needed, the same context can be reused. Removing this.

iksaif:
We deployed this patch on our pre-production cluster (~1k services), here are the results:

consul
consul-2

Considering that Prometheus is far from being the only user of Consul, this is pretty impressive :)

discovery/consul/consul.go
cstyan: Hmmm. Okay, now I think I just misunderstood what's supposed to happen here. This catches the case where we want to watch a service that consul doesn't yet know about, which it then finds later. 👍

cstyan:
lgtm, @brian-brazil ? I think we can merge before #2191 since I don't know of anyone actively working on the fix for this

brian-brazil:
> I guess that would still be an improvement. But I believe exposing the consul filtering capabilities (currently tag + node-meta, soon service-meta) has a real value because there is a non-significant overhead when using relabeling to do the filtering (as seen on the table).

It'd still be interested in knowing what just the delay would give us. The other settings are going to cause support hassle, and reduce the effectiveness of #2191.

> I think we can merge before #2191 since I don't know of anyone actively working on the fix for this

That was never blocking this.

discovery/consul/consul.go
```diff
@@ -95,9 +97,25 @@ type SDConfig struct {
...
 	Password     config_util.Secret `yaml:"password,omitempty"`
+
+	// See https://www.consul.io/docs/internals/consensus.html#consistency-modes,
+	// stale reads are a lot cheaper and are a necessity if you have >5k targets.
+	AllowStale bool `yaml:"allow_stale,omitempty"`
+	// By default use blocking queries (https://www.consul.io/api/index.html#blocking-queries)
+	// but allow users to delay updates if necessary. This can be useful because of "bugs" like
+	// https://github.com/hashicorp/consul/issues/3712 which cause an un-necessary
+	// amount of requests on consul.
+	RefreshInterval model.Duration `yaml:"refresh_interval,omitempty"`
```
brian-brazil: This is really more of a throttle than the refresh intervals that other SDs have.

discovery/consul/consul.go
```diff
@@ -203,21 +235,49 @@ func (d *Discovery) shouldWatch(name string) bool {
...
+	}
+
+	info, err := d.client.Agent().Self()
+	if err != nil {
+		level.Error(d.logger).Log("msg", "Error retrieving datacenter name", "err", err)
+		rpcFailuresCount.Inc()
+		return err
+	}
+
+	d.clientDatacenter = info["Config"]["Datacenter"].(string)
```
brian-brazil: This will panic if this is not a string, you should check the type here.

discovery/consul/consul.go
```diff
@@ -95,9 +97,25 @@ type SDConfig struct {
 	Scheme       string             `yaml:"scheme,omitempty"`
 	Username     string             `yaml:"username,omitempty"`
 	Password     config_util.Secret `yaml:"password,omitempty"`
+
+	// See https://www.consul.io/docs/internals/consensus.html#consistency-modes,
+	// stale reads are a lot cheaper and are a necessity if you have >5k targets.
+	AllowStale bool `yaml:"allow_stale,omitempty"`
```
brian-brazil: For clarity this should probably not be omitempty.

docs/configuration/configuration.md
```diff
@@ -317,11 +317,27 @@ tls_config:
 
 # A list of services for which targets are retrieved. If omitted, all services
 # are scraped.
+# See https://www.consul.io/api/catalog.html#list-nodes-for-service to know more
```
brian-brazil: This goes with node_meta

docs/configuration/configuration.md
```diff
@@ -317,11 +317,27 @@ tls_config:
...
+[ tag: <string> ]
+
+# Node metadata used to filter nodes for a given service.
+[ node_meta:
+  [ <name>: <value> ... ] ]
+
 # The string by which Consul tags are joined into the tag label.
 [ tag_separator: <string> | default = , ]
+
+# Allow stale consul results (see https://www.consul.io/api/index.html#consistency-modes). Will reduce load on consul.
```
brian-brazil: Consul

docs/configuration/configuration.md
```diff
@@ -330,6 +346,10 @@ Consul setups, the relevant address is in `__meta_consul_service_address`.
 In those cases, you can use the [relabel](#relabel_config)
 feature to replace the special `__address__` label.
 
+The Consul API only has basic support for filtering nodes (currently by node
```
brian-brazil: This should be worded so that relabelling is the primary way to do things, and the other options only exist for performance reasons. I want to avoid users trying to add features just because they don't like relabelling.

iksaif: should be a bit better now, I'm open to suggestion if not clear enough

docs/configuration/configuration.md
iksaif: done

docs/configuration/configuration.md
iksaif: This goes to both `tag` and `node_meta`, but moved after `sevices`

discovery/consul/consul.go
iksaif: Done

discovery/consul/consul.go
iksaif: Done

discovery/consul/consul.go
iksaif: done "to /delay/throttle/ updates"

iksaif:
> It'd still be interested in knowing what just the delay and allowstale would give us. 

Here is the table with the new requested column (`Delay`) with `refresh_interval: 10s`

| Resource | Empty | Before | After | After + delay | Delay |
| -------- |:-----:|:------:|:-----:|:-------------:|
|/service-discovery size|5K|85MiB|27k|27k|27k|**92MiB**|
|`go_memstats_heap_objects`|100k|**1M**|120k|110k|**1M**|
|`go_memstats_heap_alloc_bytes`|24MB|**150MB**|28MB|27MB|**140MiB**|
|`rate(go_memstats_alloc_bytes_total[5m])`|0.2MB/s|**28MB/s**|2MB/s|0.3MB/s|**23MB/s**|
|`rate(process_cpu_seconds_total[5m])`|0.1%|**15%**|2%|0.01%|**15%**|
|`process_open_fds`|16|*1236*|22|22|18|
|`rate(prometheus_sd_consul_rpc_duration_seconds_count{call="services"}[5m])`|~0|1|1|**0.03**|0.1|
|`rate(prometheus_sd_consul_rpc_duration_seconds_count{call="service"}[5m])`|0.1|**80**|0.5|0.5|**90**|
|`prometheus_target_sync_length_seconds{quantile="0.9",scrape_job="observability-by-tag"}`|N/A|200ms|0.2ms|0.2ms|180ms|
|Network bandwidth|~10kbps|**~2.8Mbps**|**~1.6Mbps**|~10kbps|**~2Mbps**|

Something weird, process_open_fds is pretty low, but I *did* have the `socket: too many open files` before I played with ulimit. This is probably because this happens only on startup and after the the `time.Sleep()` is preventing all the things to happen at the exact same time.. maybe (or it's a gauge, so we can't know).

The other results are what would be expected: filtering only with relabel_config means that all the information needs to be fetched by the client (prometheus) *before* filtering, so this is less efficient than doing it server side (even if with the patch tag filtering is *partly* client-side, but still more efficient).

docs/configuration/configuration.md
```diff
@@ -330,6 +347,12 @@ Consul setups, the relevant address is in `__meta_consul_service_address`.
 In those cases, you can use the [relabel](#relabel_config)
 feature to replace the special `__address__` label.
 
+The [relabeling phase](#relabel_config) is the preferred and more powerful
+way to filter services or nodes for a service based on arbitrary labels. For
+users with thousands of services it can be more efficient to use the Consul API
+directly which has basic support for filtering nodes (currently by node
+metadata and tag).
```
brian-brazil: I'd say "a single tag" so users don't think it's a mistake that only one works.

discovery/consul/consul.go
brian-brazil: I meant more the name of the setting. This is not like the refresh_interval setting of other SDs.

brian-brazil:
That result doesn't seem quite consistent with the rest, for example service calls shouldn't have gone up. Can you double check you performed the test correctly?

@brancz This PR contains major changes to Consul SD. Do you think we should hold it for after 2.2 for the sake of stabilising things?

iksaif:
You are right, this seems counter intuitive, but let's look at the numbers

I have 1235 services. Using relabeling I'll watch each of them.
With a refresh_interval or 10s I expect to have 123 calls per second at maximum.
With a static consul cluster, with a `WaitTime` of 30s, this should be at minimum 41 calls per second.

This means that the expected numbers are:
* **A** `refresh_interval: 0s`: [41, +inf[ (bounded by latency)
* **B** `refresh_interval: 10s`: [41, 123[

But you would expect A >> B. Which doesn't seem to be the case here. In facts it looks like there are *more* calls when refresh_interval is non-zero, which is confirmed by the following trace:
```
level=debug ts=2018-02-14T14:18:12.817694053Z caller=consul.go:419 component="discovery manager scrape" discovery=consul msg="Watching service" service=observability-portal-web tag=
level=debug ts=2018-02-14T14:18:12.825106702Z caller=consul.go:428 component="discovery manager scrape" discovery=consul msg="Watching service !!done!!" service=observability-portal-web tag= meta.LastIndex=849509831 LastIndex=849509470
level=debug ts=2018-02-14T14:18:22.825509212Z caller=consul.go:419 component="discovery manager scrape" discovery=consul msg="Watching service" service=observability-portal-web tag=
level=debug ts=2018-02-14T14:18:22.826759492Z caller=consul.go:428 component="discovery manager scrape" discovery=consul msg="Watching service !!done!!" service=observability-portal-web tag= meta.LastIndex=849510183 LastIndex=849509831
level=debug ts=2018-02-14T14:18:32.827078498Z caller=consul.go:419 component="discovery manager scrape" discovery=consul msg="Watching service" service=observability-portal-web tag=
level=debug ts=2018-02-14T14:18:32.828510762Z caller=consul.go:428 component="discovery manager scrape" discovery=consul msg="Watching service !!done!!" service=observability-portal-web tag= meta.LastIndex=849510481 LastIndex=849510183
level=debug ts=2018-02-14T14:18:42.828818192Z caller=consul.go:419 component="discovery manager scrape" discovery=consul msg="Watching service" service=observability-portal-web tag=
level=debug ts=2018-02-14T14:18:42.829889865Z caller=consul.go:428 component="discovery manager scrape" discovery=consul msg="Watching service !!done!!" service=observability-portal-web tag= meta.LastIndex=849510787 LastIndex=849510481
level=debug ts=2018-02-14T14:18:52.830123327Z caller=consul.go:419 component="discovery manager scrape" discovery=consul msg="Watching service" service=observability-portal-web tag=
level=debug ts=2018-02-14T14:18:52.831354479Z caller=consul.go:428 component="discovery manager scrape" discovery=consul msg="Watching service !!done!!" service=observability-portal-web tag= meta.LastIndex=849511144 LastIndex=849510787
level=debug ts=2018-02-14T14:19:02.831627833Z caller=consul.go:419 component="discovery manager scrape" discovery=consul msg="Watching service" service=observability-portal-web tag=
level=debug ts=2018-02-14T14:19:02.832902286Z caller=consul.go:428 component="discovery manager scrape" discovery=consul msg="Watching service !!done!!" service=observability-portal-web tag= meta.LastIndex=849511497 LastIndex=849511144
```

This happens as soon as the `refresh_interval` is bigger than 2sec, meaning that it effectively becomes a regular refresh interval. I'm trying to dig into consul's code to understand what's happening.

docs/configuration/configuration.md
iksaif: done

iksaif:
I can confirm this behavior using the following script:

```
#!/bin/bash
if test ${1:-none} = "none"
then
  echo "USAGE: $0 Consul_URL"
  echo "       Example: localhost:8500/v1/health/service/MY_SUPER_SERVICE"
  exit 1
fi
url_to_check=$1

headers=$(mktemp)
content=$(mktemp)
index=0
while true;
do
  url="${url_to_check}?wait=30s&index=${index}&pretty=true&stale"
  curl -fs --dump-header "$headers" -o "${content}.new" "${url}" || { "echo Failed to query ${url}"; exit 1; }
  sleep 5
  if test $index -ne 0
  then
    diff -u "$content" "$content.new" && echo " diff: No Differences found in service"
  fi
  index=$(grep "X-Consul-Index" "$headers" | sed 's/[^0-9]*\([0-9][0-9]*\)[^0-9]*/\1/g')
  mv "$content.new" "$content"
  printf "X-Consul-Index: $index at $(date) \n"
done
```
the output is
```
$ ./consul.sh consul05-par.central.criteo.prod:8500/v1/catalog/service/graphite-relay
X-Consul-Index: 849580974 at Wed Feb 14 15:55:16 CET 2018 
 diff: No Differences found in service
X-Consul-Index: 849581128 at Wed Feb 14 15:55:21 CET 2018 
 diff: No Differences found in service
X-Consul-Index: 849581289 at Wed Feb 14 15:55:26 CET 2018 
 diff: No Differences found in service
X-Consul-Index: 849581439 at Wed Feb 14 15:55:31 CET 2018 
 diff: No Differences found in service
X-Consul-Index: 849581603 at Wed Feb 14 15:55:36 CET 2018 
 diff: No Differences found in service
X-Consul-Index: 849581775 at Wed Feb 14 15:55:41 CET 2018 
 diff: No Differences found in service
X-Consul-Index: 849581926 at Wed Feb 14 15:55:46 CET 2018
```
In other words, waiting when reusing an index makes it ineffective.

I'm not sure what to do here, if we don't change the current patch: `refresh_interval` is effectively a refresh interval as with current versions of consul this disables blocking queries.

It's all fine for users, except if that a low `refresh_interval` will be less efficient than none. Maybe this is the solution:
```
diff --git a/discovery/consul/consul.go b/discovery/consul/consul.go
index 80f756c6..d2a1855a 100644
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -84,7 +84,7 @@ var (
                Scheme:          "http",
                Server:          "localhost:8500",
                AllowStale:      true,
-               RefreshInterval: model.Duration(15 * time.Second),
+               RefreshInterval: model.Duration(watchTimeout),
        }
 )
 
@@ -136,6 +136,10 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
        if strings.TrimSpace(c.Server) == "" {
                return fmt.Errorf("Consul SD configuration requires a server address")
        }
+       interval := time.Duration(c.RefreshInterval)
+       if interval > 0 && interval < watchTimeout {
+               return fmt.Errorf("refresh_interval (%v) can't be lower than %v", interval, watchTimeout)
+       }
        return nil
 }
```

pierresouchay:
The behaviour of indexes updated while no change happens seems to be linked to this behaviour of Consul: https://github.com/hashicorp/consul/issues/3712

The fact that being notified of change even when no real change occurs seems to be in the contract.

Having a sleep min delay is thus a good idea

iksaif:
Ok, pushed https://github.com/prometheus/prometheus/pull/3814/commits/3e9ef40136a48b5ba493c1352c0a5a6e32358a19 which I think does the correct thing:

`refresh_interval` now does what is advertised in the documentation, there won't be more that one update per `refresh_interval`. It now defaults to 30s (which was also the current waitTime in the consul query).

This also make sure we don't wait another 30s if we already waited 29s in the blocking call by substracting the number of elapsed seconds.

Hopefully this will do what people expect it does and will be safer for existing consul infrastructures.

I also updated the table with the results of setting `refresh_interval: 60s` (instead of 10s, which was too short). As expected memory stays more or less the same and the rest is divided by 6.

iksaif:
Rebased on master now that 2.2.1 is out

brian-brazil:
I'm waiting to make sure 2.2.1 is stable before merging.

discovery/consul/consul.go
```diff
@@ -226,94 +293,142 @@ func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
...
+		NodeMeta:   d.watchedNodeMeta,
+	})
+	elapsed := time.Since(t0)
+	rpcDuration.WithLabelValues("catalog", "services").Observe(elapsed.Seconds())
 
-			services[name] = cancel
+	if err != nil {
+		level.Error(d.logger).Log("msg", "Error refreshing service list", "err", err)
+		rpcFailuresCount.Inc()
+		time.Sleep(retryInterval)
```
brian-brazil: This could cause RPCs faster than the refresh interval, which could make load issues worse

discovery/consul/consul.go
```diff
@@ -226,94 +293,142 @@ func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
...
+
+			// Send clearing target group.
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case ch <- []*targetgroup.Group{{Source: name}}:
 			}
 		}
 	}
+	if elapsed < d.refreshInterval {
```
brian-brazil: This would be easier with a ticker up in Run

discovery/consul/consul.go
iksaif: Clearly. Would a max(retryInterval, d.refreshInterval) be ok ?

discovery/consul/consul.go
iksaif: Yes. But I'm not sure how to do the same thing for `watch()` which does the same thing (since a ticker is for a single "reader"). Maybe https://github.com/VividCortex/multitick but not sure we want a dependency on that.

brian-brazil: You'd have a ticker per watch()

discovery/consul/consul.go
brian-brazil: Yes, or leave it to the overall ticker

iksaif:
rebased, using tickers.

circleci seems to fail on unrelated tests.

discovery/consul/consul.go
```diff
@@ -298,34 +296,42 @@ func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
 			time.Sleep(retryInterval)
 			continue
 		}
+		// We are  good to go.
```
brian-brazil: double space

discovery/consul/consul.go
```diff
@@ -298,34 +296,42 @@ func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
...
 		// We need to watch the catalog.
+		ticker := time.NewTicker(d.refreshInterval)
 		go func() {
+			// Watched services and their cancellation functions.
+			services := make(map[string]func())
 			var lastIndex uint64
-			for range ticker.C {
-				d.watchServices(ctx, ch, ticker, &lastIndex, services)
+
+			for ; true; <-ticker.C {
```
brian-brazil: You should have a select that also checks if the context is done

discovery/consul/consul.go
```diff
@@ -414,103 +420,100 @@ func (d *Discovery) watchService(ctx context.Context, ch chan<- []*targetgroup.G
 	}
 
 	go func() {
+		ticker := time.NewTicker(d.refreshInterval)
 		var lastIndex uint64
-		for range ticker.C {
-			srv.watch(ctx, ch, &lastIndex)
+		catalog := srv.client.Catalog()
+		for ; true; <-ticker.C {
```
brian-brazil: You should be checking the context here

discovery/consul/consul.go
```diff
@@ -414,103 +420,100 @@ func (d *Discovery) watchService(ctx context.Context, ch chan<- []*targetgroup.G
...
+
+		// Add all key/value pairs from the node's metadata as their own labels
+		for k, v := range node.NodeMeta {
+			name := strutil.SanitizeLabelName(k)
+			labels[metaDataLabel+model.LabelName(name)] = model.LabelValue(v)
 		}
+
+		tgroup.Targets = append(tgroup.Targets, labels)
 	}
+	// Check context twice to ensure we always catch cancellation.
```
brian-brazil: I don't think you need to do this

discovery/consul/consul.go
```diff
@@ -414,103 +420,100 @@ func (d *Discovery) watchService(ctx context.Context, ch chan<- []*targetgroup.G
...
+	nodes, meta, err := catalog.Service(srv.name, srv.tag, &consul.QueryOptions{
+		WaitIndex:  *lastIndex,
+		WaitTime:   watchTimeout,
+		AllowStale: srv.discovery.allowStale,
+		NodeMeta:   srv.discovery.watchedNodeMeta,
+	})
+	elapsed := time.Since(t0)
+	rpcDuration.WithLabelValues("catalog", "service").Observe(elapsed.Seconds())
+
+	// Check the context before in order to exit early.
```
brian-brazil: It's more usual to do this up in the for

iksaif: I think this was originally meant to exit before handling the response or the error.
I'm happy to remove it though.

discovery/consul/consul.go
iksaif: removed

discovery/consul/consul.go
iksaif: isn't it sufficient to check ticker.C ? (which is going to be closed when the context is done)

discovery/consul/consul.go
iksaif: fixed

discovery/consul/consul.go
```diff
@@ -226,164 +290,225 @@ func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
...
+
+	if len(d.watchedServices) == 0 || d.watchedTag != "" {
+		// We need to watch the catalog.
+		ticker := time.NewTicker(d.refreshInterval)
+		go func() {
+			// Watched services and their cancellation functions.
+			services := make(map[string]func())
+			var lastIndex uint64
+
+			for ; true; <-ticker.C {
```
brian-brazil: `for range ticker.C` is the usual way to spell this

iksaif: `for ; true; <-ticker.C`  is similar, but execute the loop once immediately to speed up initial target discover

brian-brazil:
Thanks!

Issue #3807, opened 2018-02-07
Improve health detection of /-/healthy endpoint


jeremyeder:
We're trying to find the best way to monitor prometheus itself.  We were evaluating using the /healthy and /ready endpoints.  But we have a case where prometheus was failed, yet the endpoints still reported back as healthy.

**What did you do?**
Used Prometheus 2.0, ran out of disk space.

**What did you expect to see?**
I expected to see prometheus marked unhealthy when querying the /-/healthy endpoint.

**What did you see instead? Under which circumstances?**
Prometheus responded "Prometheus is Healthy."

**Environment**
- RHEL/OpenShift

* System information:
```Linux 3.10.0-693.11.6.el7.x86_64 x86_64```

* Prometheus version:
prometheus:v2.0.0

* Logs:
```
level=error ts=2018-02-05T21:42:18.480373214Z caller=wal.go:713 component=tsdb msg="sync failed" err="flush buffer: write data/wal/000001: file already closed"
level=warn ts=2018-02-05T21:42:21.495046779Z caller=scrape.go:673 component="target manager" scrape_pool=kubernetes-service-endpoints target=https://10.131.12.9:8443/metrics msg="append failed" err="WAL log samples: log series: write data/wal/000001: file already closed"
```

krasi-georgiev:
I think `tsdb` is out of scope here so removed the local storage label.

gertvdijk:
I can confirm this is still an issue with latest stable (tested 2.15.1). Encountering a disk-full event with Prometheus is really bad from an ops situation on two more aspects imo:

- VERY spammy logs, multiple lines per second. Why keep trying for every single write over and over?
- No exit, keeps running. Process manager thinks everything is OK. Why not exit status > 0?
- Liveness / health check says OK. (That's what this issue is about.)

brian-brazil:
If your disk has filled up, you're already in a really bad place. You should aim to avoid getting into that situation, rather than trying to rearrange the deckchairs when it happen. For example on restart Prometheus will still have a full disk.

gertvdijk:
> If your disk has filled up, you're already in a really bad place. You should aim to avoid getting into that situation, rather than trying to rearrange the deckchairs when it happen. For example on restart Prometheus will still have a full disk.

Of course, I agree, shit hit the fan already. But that doesn't mean the process has to continue running and pretend everything is OK.

krasi-georgiev:
[...]

Issue #3650, opened 2018-01-04
Improve `/-/healthy` endpoint to check more than the webHandler


Thib17:
Prometheus already provides an `/-/healthy` endpoint (https://github.com/prometheus/prometheus/blob/master/web/web.go#L262). But this endpoint keep responding 200 for as long as the webHandler is up. In some situation the webHandler is up, but the Prometheus doesn't work as expected (e.g. it can't write on disk, some important mutex are deadlocked, ...).

Therefore the `/-/healthy` should be improved to better check Prometheus health in order to detect when it needs to get killed. In my opinion testing that we can at least read samples that are fresh enough could be a good start.

I would be happy to help on that, but I have no idea how should I start.

discordianfish:
We have both /-/healthy which does what you describe and /-/ready which is doing what I think you want, right? See this for context: https://github.com/prometheus/prometheus/pull/2997

Thib17:
`/-/ready` seems to be only used at start-up. AFAIK, once the webHandler is set ready, it never change. 
So if the Prometheus properly starts and then stop working as expected after a while, both current `/-/healthy` and `/-/ready` will keep respond 200. In such situation I believe that `/-/ready` is right to respond 200, but `/-/healthy` should respond 5xx.

discordianfish:
If I read this correctly, it should always check if it's ready before returning 200 on the ready endpoint. The handler is wrapped by https://github.com/prometheus/prometheus/pull/2997/files#diff-73cbd2009b0e4f156e8ca0f47e95b016R273

Thib17:
True, but as I read it `h.ready` is never set to 0. So once it's set to 1, isReady() always return true, and `/-/ready` always responds 200. And this looks normal to me, as it checks that start-up is done.

What doesn't look normal to me, is that `/-/healthy` always return 200, no matter if the Prometheus works as expected or not.

discordianfish:
/-/ready is suppose to respond with 200 no matter what. It can be used as liveness probe in kubernetes: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/

Now regarding /-/healthy, it looks like you're right. Nothing sets that to non-ready. So I suppose it's a reasonable request to have this return non-2xx if something breaks after startup. @beorn7 / @grobie Thoughts?

grobie:
Sounds good. What and how do you want to check?

Thib17:
I would check that reads and writes to the local storage are working.
[...]

PR #3816, opened 2018-02-08
Improve the healthy handler

simonpasquier:
Prometheus will report itself as healthy if it can write to the local
metrics storage directory.

It is a naive and simplistic implementation but should at least address the issue described in #3807. There's also a discussion on #3650 about other things to check.

closes https://github.com/prometheus/prometheus/issues/3807

### DIFF
--- a/util/health/disk_default.go
+++ b/util/health/disk_default.go
@@ -1,0 +1,35 @@
+// Copyright 2018 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build !netbsd
+// +build !openbsd
+// +build !windows
+
+package health
+
+import (
+	"errors"
+	"syscall"
+)
+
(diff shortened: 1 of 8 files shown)


krasi-georgiev:
I think that would be a very good improvement. 

@Gouthamve is there any way to use tsdb itself to report if it is healthy and not run out of space?

Another idea that comes to mind is to add a middleware to the global Logger and if any of the components logs an error of some sort to set the Healthy handler to false with the actual error for the `/healthy` endpoint to display.

util/health/disk_default.go
```diff
@@ -0,0 +1,35 @@
...
+	"syscall"
+)
+
+func (d *dsReporter) Report() error {
+	fs := syscall.Statfs_t{}
+	err := syscall.Statfs(d.path, &fs)
+	if err != nil {
+		return err
+	}
+	if fs.Bavail == 0 {
```
iksaif: does that work with quotas ?

simonpasquier: You mean [Linux disk quotas](https://www.centos.org/docs/5/html/Deployment_Guide-en-US/ch-disk-quotas.html)? I checked quickly with an ext4 fs and apparently quotas aren't taken into account (but the same is true for `df`).

simonpasquier:
> What if local storage is disabled (using memory only ?)

I don't think it is possible with 2.x.

 > Proposal: also implement isHealthy() in a few subsystems, and for each of them implement a lightweight check

Agreed although I'd be wary to implement health checks which are not based on concrete and reproducible errors.

krasi-georgiev:
@fabxc also mentioned an idea to write to the tsdb every time the health endpoint is hitand use that as an indication.

anandsinghkunwar:
Is there any update on this?

simonpasquier:
This is too old. Closing for now.

PR #3817, opened 2018-02-09
Improve wording of remote write documentation.

SuperQ:
Reduce the use of the term `long-term`, when what we're really talking
about is remote clustered storage for increased capacity and durability.

### DIFF
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -64,7 +64,7 @@ If your local storage becomes corrupted for whatever reason, your best bet is to
 
 ## Remote storage integrations
 
-As outlined above, Prometheus's local storage is limited in its scalability and durability. Instead of trying to solve long-term storage in Prometheus itself, Prometheus has a set of interfaces that allow integrating with remote long-term storage systems.
+Prometheus's local storage is limited by single nodes in its scalability and durability. Instead of trying to solve clustered storage in Prometheus itself, Prometheus has a set of interfaces that allow integrating with remote storage systems.
 
 ### Overview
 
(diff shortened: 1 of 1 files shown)


docs/storage.md
```diff
@@ -64,7 +64,7 @@ If your local storage becomes corrupted for whatever reason, your best bet is to
 
 ## Remote storage integrations
 
-As outlined above, Prometheus's local storage is limited in its scalability and durability. Instead of trying to solve long-term storage in Prometheus itself, Prometheus has a set of interfaces that allow integrating with remote long-term storage systems.
+Prometheus's local storage is limited by single nodes in its scalability and durability. Instead of trying to solve clustered storage in Prometheus itself, Prometheus has a set of interfaces that allow integrating with remote clustered storage systems.
```
jamtur01: Is it always clustered though? One of the targets is Graphite for example...?

SuperQ: It's not always clustered, but the intended use case of remote storage is for larger / clustered storage systems.  The main point of this PR is to change the message about remote write being used for "long-term" storage, as the 2.0 TSDB is perfectly capable of handling long retention times.

Users are too often trying to implement remote write to TSDBs, like Graphite, thinking that they need to because Prometheus is unreliable.  All they really need to do is adjust the storage retention flag.

Single node Graphite on whisper is, IMO a worse solution.  However there are several clustered Graphite options that would be more suitable.

docs/storage.md
```diff
@@ -75,7 +75,7 @@ Prometheus integrates with remote storage systems in two ways:
 
 ![Remote read and write architecture](images/remote_integrations.png)
 
-The read and write protocols both use a snappy-compressed protocol buffer encoding over HTTP. The protocols are not considered as stable APIs yet and may change to use gRPC over HTTP/2 in the future, when all hops between Prometheus and the long-term storage can safely be assumed to support HTTP/2.
+The read and write protocols both use a snappy-compressed protocol buffer encoding over HTTP. The protocols are not considered as stable APIs yet and may change to use gRPC over HTTP/2 in the future, when all hops between Prometheus and the external storage can safely be assumed to support HTTP/2.
```
brian-brazil: "remote" would be more consistent, we don't use the word external for this anywhere else

PR #3819, opened 2018-02-10
merge pkg/relabel and relabel

### LINKED ISSUE #3647: Why do we need 2 packages relabel packages? /pkg/relabel, /relabel
Wondering what is the reason to have 2 almost identical relabel packages and if we can combine it into one.
/pkg/relabel
/relabel

or rename to avoid the confusion if they serve different purpose.

krasi-georgiev:
fixes #3647
both packages have the same purpose, but the Process func takes
different parameters so need to keep both.

another PR could be to refactor and remove the duplicated functionality.


pkg/relabel/relabel.go
```diff
@@ -102,6 +102,87 @@ func relabel(lset labels.Labels, cfg *config.RelabelConfig) labels.Labels {
 	return lb.Labels()
 }
 
+// ProcessSet returns a relabeled copy of the given label set. The relabel configurations
+// are applied in order of input.
+// If a label set is dropped, nil is returned.
+// May return the input labelSet modified.
+func ProcessSet(labels model.LabelSet, cfgs ...*config.RelabelConfig) model.LabelSet {
```
brian-brazil: We should need only one of these two functions, we should be moving away from model.

krasi-georgiev: refactoring should need quite a bit of code changes so I thought to leave it for another PR.

krasi-georgiev: in general I think I will have time to do the refactoring , but I prefer to do it in few smaller PRs for easy review and merging without the need of constant rebase with master.

krasi-georgiev: @brian-brazil is it ok to do this in few stages to keep the PRs small?

brian-brazil: It's fine to do it in stages, but this PR is not one of those stages.

brian-brazil: Yes, and then you can delete the older relabel package.

brian-brazil:
This needs a different approach.

PR #3821, opened 2018-02-11
Drop metric name for bool modifier

### LINKED ISSUE #3820: Bool modifier should remove metric name
The `bool` modifier changes the value of the metric, and thus in line with everywhere else in PromQL should be removing the metric name.

krasi-georgiev:
fixes #3820

a single line , but took 3h of my life!
Overall very good exercise  for the promql execution. :dog:

### DIFF
--- a/promql/testdata/operators.test
+++ b/promql/testdata/operators.test
@@ -236,6 +236,9 @@ eval instant at 50m 0 == bool 1
 eval instant at 50m 1 == bool 1
 	1
 
+eval instant at 50m http_requests{job="api-server", instance="0", group="production"} == bool 100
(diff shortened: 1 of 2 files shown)


promql/testdata/operators.test
```diff
@@ -236,6 +236,10 @@ eval instant at 50m 0 == bool 1
 eval instant at 50m 1 == bool 1
 	1
 
+# bool modifier.
```
brian-brazil: This is still part of the comparison tests, so you don't need this comment.

krasi-georgiev: :+1:  I thought you might say that : updated

brian-brazil:
Thanks!

PR #3829, opened 2018-02-12
docs: Fix remote_read/remote_timeout default

AndreaGiardini:
As we can see here: https://github.com/prometheus/prometheus/blob/master/config/config.go#L130

The default timeout for `read_remote` is 1m and not 30s as specified in the documentation


brian-brazil:
Thanks!

PR #3833, opened 2018-02-12
Add USERS.md

RichiH:


brian-brazil:
Do we really need one of these, and in mixed in with our code? A wiki page would be better I think.

RichiH:
I thought that's what we said in the call; this seemed to be the obvious place, but I can see the argument of why a different place might be better suited.

brian-brazil:
Was there was no real outcome, for the purposes of graduation what we have is already sufficient.

caniszczyk:
+1 to code as people will issue a PR which grants you permission (it's also somewhat of a CNCF graduation requirement)

jamtur01:
These have always ended in a sea of PRs adding people/co's comingled with actual code PRs. PR management is already hard enough for a big project. Better suited to a webpage IMHO. A webpage can also mean styling, logos, etc, etc.

caniszczyk:
If you make the webpage easy to update than sure, but right now it's really hard to add yourself to the user list. It's significantly easier to make an ADOPTERS.md or USERS.md in the repo

brian-brazil:
The `prometheus` repository is our primary repository, but it is meant only for things directly related to the Prometheus server. We have split things across the 30ish repositories we currently use, the `prometheus` repo is not a place where things go by default.

What is the problem that trying to list some of the several thousand organisations using Prometheus will solve?

RichiH:
https://github.com/prometheus/docs/pull/969

PR #3834, opened 2018-02-13
Added "No rules defined" to Rules page in Web UI

jamtur01:
When you have no alerting rules defined you get a screen sharing this
information in the WebUI. If no rules are defined then you instead see
an empty white screen. This adds a "No rules" defined `else` clause and
a `Rules` header to the page.


jamtur01:
I think this is @juliusv's area according to the MAINTAINERS list.

juliusv:
@jamtur01 Thanks :+1:

PR #3835, opened 2018-02-13
Pool package generalize

krasi-georgiev:
makes the pool package general to be used for any type by specifying its own constructor when initialising  the pool
this is in order to use the pool in promql to optimise the memory pool reuse.

@fabxc as we discussed it when talking about the promql optimisation. The text PR would be to refactor promql use the Pool.

### DIFF
--- a/pkg/pool/pool.go
+++ b/pkg/pool/pool.go
@@ -71,16 +71,16 @@ func (p *Pool) Get(sz int) interface{} {
 }
 
 // Put adds a slice to the right bucket in the pool.
-func (p *Pool) Put(s interface{}) error {
+func (p *Pool) Put(s interface{}) {
 	slice := reflect.ValueOf(s)
-	if slice.Kind() == reflect.Slice {
-		for i, size := range p.sizes {
-			if slice.Cap() > size {
-				continue
-			}
-			p.buckets[i].Put(slice.Slice(0, 0).Interface())
-			return nil
-		}
+
+	if slice.Kind() != reflect.Slice {
+		panic(fmt.Sprintf("%+v is not a slice", slice))
 	}
-	return fmt.Errorf("%+v is not a slice", slice)
+	for i, size := range p.sizes {
+		if slice.Cap() > size {
+			continue
+		}
+		p.buckets[i].Put(slice.Slice(0, 0).Interface())
+	}
 }
--- a/pkg/pool/pool.go
+++ b/pkg/pool/pool.go
@@ -23,13 +23,13 @@ import (
 type Pool struct {
 	buckets []sync.Pool
 	sizes   []int
-	// initialize is the function used to create an empty slice when none exist yet.
-	initialize func(int) interface{}
+	// make is the function used to create an empty slice when none exist yet.
+	make func(int) interface{}
 }
 
 // New returns a new Pool with size buckets for minSize to maxSize
 // increasing by the given factor.
-func New(minSize, maxSize int, factor float64, newFunc func(int) interface{}) *Pool {
+func New(minSize, maxSize int, factor float64, makeFunc func(int) interface{}) *Pool {
 	if minSize < 1 {
 		panic("invalid minimum pool size")
 	}
@@ -47,9 +47,9 @@ func New(minSize, maxSize int, factor float64, newFunc func(int) interface{}) *P
 	}
 
 	p := &Pool{
-		buckets:    make([]sync.Pool, len(sizes)),
-		sizes:      sizes,
-		initialize: newFunc,
+		buckets: make([]sync.Pool, len(sizes)),
+		sizes:   sizes,
+		make:    makeFunc,
 	}
 
 	return p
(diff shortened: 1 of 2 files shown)


scrape/scrape.go
```diff
@@ -662,19 +667,26 @@ mainLoop:
 				time.Since(last).Seconds(),
 			)
 		}
-		b := sl.buffers.Get(sl.lastScrapeSize)
-		buf := bytes.NewBuffer(b)
+		temp, err := sl.buffers.Get(sl.buffers, sl.lastScrapeSize)
```
fabxc: I'm not sure I quite understand this. Why would we pass the object itself as the first argument to a method on itself?

pkg/pool/pool.go
```diff
@@ -13,17 +13,35 @@
...
+// New initializes a new pool.
+func New() *Pool {
+	p := &Pool{
+		buckets: make(map[interface{}]bucket),
+	}
+	return p
+}
+
+// Add adds a new bucket with a given id to the Pool with size for minSize to maxSize increasing by the given factor.
+func (p *Pool) Add(id interface{}, minSize, maxSize int, factor float64, newFunc func(int) interface{}) {
```
fabxc: A pool with similar semantics for arbitrary slice types is certainly useful. But I'm not sure we also need to wrap this once again to have an object holding multiple of those pools.

The naming is also rather confusing as the former Pool is now called a bucket – but a bucket for a certain type/id (?) now rather than a certain slice size range.

I would make it a bit simpler again and really just move the existing bucket from `[]byte` to []interface{}`.

krasi-georgiev: classic over-engineering  :)

please check the updated version and let me know if you see anyhow to simplify it..

krasi-georgiev:
:clapper:  @fabxc

pkg/pool/pool.go
```diff
@@ -40,36 +46,41 @@ func NewBytesPool(minSize, maxSize int, factor float64) *BytesPool {
...
+			if slice.Cap() > size {
+				continue
+			}
+			p.buckets[i].Put(slice.Slice(0, 0).Interface())
+			return nil
 		}
-		p.buckets[i].Put(b[:0])
-		return
 	}
+	return fmt.Errorf("%+v is not a slice", slice)
```
fabxc: I think an error on `Put` is quite annoying to deal with which is in every case a programming bug. Panicking here seems fine and in line with similar things in the standard library, e.g. https://godoc.org/sort#Slice

Also, the shorter conditions should generally be checked first so we can exit early and keep indentations low, e.g.:

```go
slice := reflect.ValueOf(s)

if slice.Kind() != reflect.Slice {
  panic("exit early")
}

for i, size := range p.sizes {
  if slice.Cap() > size {
    continue
  }
  p.buckets[i].Put(slice.Slice(0, 0).Interface())
  return
}
```

fabxc: Probably the check is not necessary as `slice.Cap` will panick on non-slices anyway. But not sure.

pkg/pool/pool.go
```diff
@@ -13,17 +13,23 @@
...
+)
 
-// BytesPool is a bucketed pool for variably sized byte slices.
-type BytesPool struct {
+// Pool is a bucketed pool for variably sized byte slices.
+type Pool struct {
 	buckets []sync.Pool
 	sizes   []int
+	// initialize is the function used to create an empty slice when none exist yet.
+	initialize func(int) interface{}
```
fabxc: I find the name confusing – "initialization" != "creation"

fabxc:
Looks good aside from some comments.

pkg/pool/pool.go
krasi-georgiev: I thought about panic-ing as well, but wasn't completely sure.
All updated.
Thanks

krasi-georgiev:
:clapper: @fabxc

pkg/pool/pool.go
```diff
@@ -40,36 +46,41 @@ func NewBytesPool(minSize, maxSize int, factor float64) *BytesPool {
...
+	if slice.Kind() != reflect.Slice {
+		panic(fmt.Sprintf("%+v is not a slice", slice))
+	}
+	for i, size := range p.sizes {
+		if slice.Cap() > size {
 			continue
 		}
-		p.buckets[i].Put(b[:0])
-		return
+		p.buckets[i].Put(slice.Slice(0, 0).Interface())
```
fabxc: You dropped the early `return` – this won't end well ;)

scrape/scrape.go
```diff
@@ -662,7 +662,11 @@ mainLoop:
 				time.Since(last).Seconds(),
 			)
 		}
-		b := sl.buffers.Get(sl.lastScrapeSize)
+		b, ok := sl.buffers.Get(sl.lastScrapeSize).([]byte)
+		if !ok {
+			b = make([]byte, 0, sl.lastScrapeSize)
+			level.Error(sl.l).Log("msg", "buffer pool type assertion error")
+		}
```
fabxc: Mh, I think a failing assertion here is a hard programming bug just as much as passing a non-slice to `Put`. Maybe just cast it without the `ok` check and let it fail otherwise?

pkg/pool/pool.go
krasi-georgiev: oops :shamrock:

fabxc:
👍  Thanks

krasi-georgiev:
ping @brian-brazil

krasi-georgiev:
ping @brian-brazil

PR #3836, opened 2018-02-13
Removed alert_header from "No alert rules defined" table row in WebUI

jamtur01:
This is a minor UX change. The current "No Alert rules" table row has the `alert_header` class attached. This changes the cursor and some other stuff and makes sense with the populated table but less sense with the unpopulated table. So removing it in the latter case.

/cc @juliusv


juliusv:
:+1: Thanks!

PR #3839, opened 2018-02-14
promql: Remove old and unused alerting/reconding syntax

### LINKED ISSUE #3818: Remove old PromQL artifacts
I was reading through PromQL parser/lexer code and realized there are old and unused artifacts in the lexer/parser from the old (pre 2.0) recording and alerting rule format.

For example: https://github.com/prometheus/prometheus/blob/dfdd2cb5f421a7d71dd50025ac1dbd07f019ec8b/promql/lex.go#L221-L225

This should also simplify the lexer/parser, as only expressions will be left to parse.

brancz:
I had the changes done so I figured I'd open the PR. I would suggest that we hold off on this until we get the stable v2.2.0 release out, so we make sure that at least for 2.0, 2.1 and 2.2 promtool has the ability to convert from old to new syntax.

Closes #3818 

@fabxc @Gouthamve @brian-brazil


fabxc:
Nice! 👍 I've been looking forward to ditch that code.

brian-brazil:
We might want to wait a little longer on that, my gut says at least 6 months after 2.0.

fabxc:
We probably won't get to 2.3 before April so what Frederic said should line up with the 6 months quite well.

juliusv:
FWIW, I'm also making Cortex support both v1 and v2 rule formats in the same binary (flag-configurable), which might stick around for a while until everyone is on v2. So it'd be handy from that side to not have to vendor two different versions of the relevant Prometheus packages, but I understand if that gets too annoying for Prometheus.

juliusv:
Just checked with @bboreham, and apparently Weaveworks is still completely on 1.x rules, so Cortex there would at least still need 1.x rule support for a while. There might be other Cortex users still on 1.x too. There might also still be normal Prometheus users still on 1.x who still need a conversion tool later :) For Cortex this would mean that it gets stuck on an old vendoring, and duplicating the necessary code doesn't seem workable from a glance at this PR.

Of course we never guaranteed external stability of these parts of Prometheus, so I'll leave it up to you to decide.

fabxc:
Conversion tool can always be retrieved by just downloading `promtool` from an old release though. For a one-off operation that seems fine.

I'm not super in a rush – but a timeline to get rid of essentially dead code would be nice.

juliusv:
> Conversion tool can always be retrieved by just downloading promtool from an old release though. For a one-off operation that seems fine.

Yeah, I think so too.

> I'm not super in a rush – but a timeline to get rid of essentially dead code would be nice.

@bboreham Any timeline guarantees you could give from Weaveworks' side at least? Would be unfortunate for WW if Cortex was stuck on an old PromQL engine, and I think it's not easy to just duplicate the relevant parts, as it's all intertwined in the same (`promql`) package.

brian-brazil:
I'd like us to have a solid 2.x out before removing this.

brancz:
@juliusv if I remember correctly I spoke to @tomwilkie about this and he mentioned he has a go-yacc based parser for rules and PromQL, but not sure that has landed in Cortex. Maybe he can comment on it.

tomwilkie:
I wouldn't worry about Cortex: we already copied the old 1.x chunk into the code base, I'm sure we can do the same with the old rules parser code.

brancz:
How does everyone feel about merging this now that v2.3.0 is out?

brian-brazil:
We don't yet know how stable 2.3.0 is.

brancz:
Makes sense. Let's wait a couple of weeks to see how it goes.

brancz:
How does everyone feel about this for v2.4.0? If you think we should have it for v2.4.0 I'm happy to rebase this.

brancz:
It has been one year since the 2.0 release, and I have just confirmed on IRC, that others agree this is fine to merge now.

PR #3847, opened 2018-02-15
Make it easier to build docker images for other architectures

dmathieu:


brancz:
@grobie and @sdurrheimer probably have the best idea on this, but I would say that building container images for multiple platforms probably belongs mostly in promu, and we should build the busybox images for the different architectures like in [prometheus/busybox](https://github.com/prometheus/busybox).

brian-brazil:
This is covered under https://github.com/prometheus/promu/issues/89, we shouldn't be doing this ad-hoc in other repo before that is resolved.

PR #3849, opened 2018-02-15
Tests empty labels and honor

### LINKED ISSUE #3841: Empty labels being returned from PromQL
https://github.com/prometheus/alertmanager/issues/1240 shows empty labels somehow coming out of storage and into alerts. I've seen this previously myself also, but haven't narrowed it down. I suspect an issue on the ingestion path somewhere.

krasi-georgiev:
Fixes #3841 #3845

### DIFF
--- a/scrape/scrape_test.go
+++ b/scrape/scrape_test.go
@@ -660,42 +660,97 @@ func TestScrapeLoopRunCreatesStaleMarkersOnParseFailure(t *testing.T) {
 }
 
 func TestScrapeLoopAppend(t *testing.T) {
-	app := &collectResultAppender{}
 
-	sl := newScrapeLoop(context.Background(),
-		nil, nil, nil,
-		nopMutator,
-		nopMutator,
-		func() storage.Appender { return app },
-	)
-
-	now := time.Now()
-	_, _, err := sl.append([]byte("metric_a 1\nmetric_b NaN\n"), now)
-	if err != nil {
-		t.Fatalf("Unexpected append error: %s", err)
-	}
-
-	ingestedNaN := math.Float64bits(app.result[1].v)
-	if ingestedNaN != value.NormalNaN {
-		t.Fatalf("Appended NaN samples wasn't as expected. Wanted: %x Got: %x", value.NormalNaN, ingestedNaN)
-	}
-
-	// DeepEqual will report NaNs as being different, so replace with a different value.
-	app.result[1].v = 42
-	want := []sample{
+	tests := []struct {
+		title           string
+		honorLabels     bool
+		scrapeLabels    string
+		discoveryLabels []string
+		expLset         labels.Labels
+		expValue        float64
+	}{
 		{
-			metric: labels.FromStrings(model.MetricNameLabel, "metric_a"),
-			t:      timestamp.FromTime(now),
-			v:      1,
-		},
-		{
-			metric: labels.FromStrings(model.MetricNameLabel, "metric_b"),
-			t:      timestamp.FromTime(now),
-			v:      42,
+			// When "honor_labels" is not set
+			// label name collision is handler by adding a prefix.
+			title:           "Label name collision",
+			honorLabels:     false,
+			scrapeLabels:    `metric{n="1"} 0`,
+			discoveryLabels: []string{"n", "2"},
+			expLset:         labels.FromStrings("__name__", "metric", "exported_n", "1", "n", "2"),
+			expValue:        0,
+		}, {
+			// Labels with no value need to be removed as these should not be ingested.
+			title:           "Delete Empty labels",
+			honorLabels:     false,
+			scrapeLabels:    `metric{n=""} 0`,
+			discoveryLabels: nil,
+			expLset:         labels.FromStrings("__name__", "metric"),
+			expValue:        0,
+		}, {
+			// Honor Labels should ignore labels with the same name.
+			title:           "Honor Labels",
+			honorLabels:     true,
+			scrapeLabels:    `metric{n1="1" n2="2"} 0`,
+			discoveryLabels: []string{"n1", "0"},
+			expLset:         labels.FromStrings("__name__", "metric", "n1", "1", "n2", "2"),
+			expValue:        0,
+		}, {
+			title:           "Stale - NaN",
+			honorLabels:     false,
+			scrapeLabels:    `metric NaN`,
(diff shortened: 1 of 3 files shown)


scrape/scrape_test.go
```diff
@@ -1138,6 +1138,117 @@ func TestTargetScrapeScrapeNotFound(t *testing.T) {
...
+				},
+			},
+			Expected: []labels.Label{
+				{
+					Name:  "n",
+					Value: "v",
+				},
+			},
+		}, {
+			Title: "Delete When label name is empty",
```
brian-brazil: You also need to test that an empty scraped label overrides a target label when honor labels is set

krasi-georgiev: Just added this check as part of the `Honour Label` test.

krasi-georgiev:
If I make it broader isn't this more of an integration test? 

At the moment is seems that we have a mix of unit/integration tests , but might want to start thinking if we can separate these a bit. Especially if we would decide to start working with the CNCF CI.

krasi-georgiev:
My logic here is that we want to test the functionality of this `unit` - `mutateSampleLabels`

Also I find in the past that too broader test require refactoring on every code change which I am hoping it can be avoided by smaller unit tests.

I can add one more broader test , but would prefer if we keep this one as well.
Would also be also curious to see which one turns out to be more useful in the long run.

fabxc:
I tend to agree that this is probably fine to test as a unit test.
If the parser gets a unit test itself for the empty label stuff it should be fine.

scrape/scrape_test.go
```diff
@@ -1138,6 +1138,125 @@ func TestTargetScrapeScrapeNotFound(t *testing.T) {
...
+func TestMutateSampleLabels(t *testing.T) {
+
+	tests := []struct {
+		Title      string
+		ScrapePool *scrapePool
+		Input      struct {
+			Lset   labels.Labels
+			Target *Target
+		}
+		Expected labels.Labels
```
fabxc: No realt need to export all those.

scrape/scrape_test.go
```diff
@@ -1138,6 +1138,125 @@ func TestTargetScrapeScrapeNotFound(t *testing.T) {
...
+				Target: &Target{
+					labels: []labels.Label{
+						{
+							Name:  "n",
+							Value: "v0",
+						},
+						{
+							Name:  "n1",
+							Value: "v1",
+						},
```
fabxc: I'd inline those like `{Name: "n1", Value: "v1"}` so the lines don't bloat up as bad. I find it hard to read a single test unit otherwise.

brian-brazil:
> If I make it broader isn't this more of an integration test?

Integration tests involve running other binaries, which we're not doing here. I want a test for the feature that if a user specifies an empty scrape label with honor labels enabled that it will result in that label not being ingested, and this test does not fully do that.

What the parser does with empty labels is basically unspecified behaviour, so a unittest over there doesn't prevent someone from changing that behaviour, changing the unittest and breaking the actual feature we care about.

krasi-georgiev:
@brian-brazil yes that makes total sense we need to cover that. 
I added an additional test to make sure the parser doesn't strip empty labels with an additional comment to explain why we need this behaviour.
If you still think this is insufficient  can you give me an example of a possible scenario how this bug can be reintroduced.

brian-brazil:
> My main goal is to prevent reintroducing the bug , but keep the tests as simple as possible to avoid unnecessary refactoring.

This style of testing actually makes refactoring harder, as you need to rewrite the tests if the structure around mutateSamples changes. Just like other tests in this file I would like to test from scrape data all the way to appending, as that is what we really care about.

krasi-georgiev:
yep , you are right here. I refactored it and really happy how it turned out.

krasi-georgiev:
@fabxc I have refactored the TestScrapeLoopAppend to include e2e ingestion path.
This would make it more difficult to maintain and most likely would require more frequent refactoring, but if @brian-brazil thinks it is important I can back off on this one.

I am thinking that better names for the  tests case labels could be `scrapeLabels` , `discoveryLabels.` 
it shows the origin of the labels which makes the test a bit easy to understand.

scrape/scrape_test.go
```diff
@@ -660,42 +661,109 @@ func TestScrapeLoopRunCreatesStaleMarkersOnParseFailure(t *testing.T) {
...
+		expected        []string
+	}{
+		{
+			// When "honor_labels" is not set
+			// label name collision is handler by adding a prefix.
+			title:           "Label name collision",
+			honorLabels:     false,
+			scrapeLabels:    `metric{n="1"} 0`,
+			discoveryLabels: []string{"n", "2"},
+			expected:        []string{"metric", "0", "exported_n", "1", "n", "2"},
```
fabxc: Mh, so this encodes the metric name in the first entry, then a value in the 2nd and then label pairs in the subsequent one? Seems rather convoluted and I had to read the test code to check what the test case actually expresses.

Maybe split it up into dedicated fields?

fabxc:
Seems generally fine to me aside from that one comment.

scrape/scrape_test.go
krasi-georgiev: now when I see it changed it looks so logical and so simple , thanks for the :1st_place_medal:  comments

fabxc:
LGTM. @brian-brazil  for final approval

scrape/scrape_test.go
```diff
@@ -660,42 +660,100 @@ func TestScrapeLoopRunCreatesStaleMarkersOnParseFailure(t *testing.T) {
...
+		expected := []sample{
+			{
+				metric: test.expLset,
+				t:      timestamp.FromTime(now),
+				v:      test.expValue,
+			},
+		}
+
+		// The special case with the NaN is that we only check if the result reports that the value is NaN.
+		if math.Float64bits(app.result[0].v) > 0 {
```
brian-brazil: This seems more complicated than it needs to be. Why not check that the value for all cases matches what's expected, using math.Float64bits?

brian-brazil:
Can you squash down your changes?

scrape/scrape_test.go
krasi-georgiev: the updated version is much cleaner.

krasi-georgiev:
@brian-brazil ping

brian-brazil:
Thanks!

PR #3853, opened 2018-02-16
Remove unused flock package

simonpasquier:
This package was used in the pre-2.x code base.


krasi-georgiev:
how did you even catch that?
LGTM on green - always happy when reducing code count :smile:

brian-brazil:
Thanks!

krasi-georgiev:
shouldn't we also have some check for unused vendored packages?

a quick check with `vendorcheck -u` shows few unused (some false positive)
```
go get github.com/FiloSottile/vendorcheck
vendorcheck -u
```

simonpasquier:
> how did you even catch that?

I was just browsing `util/` and wondered why there was  a `flock` directory there. It didn't took me long to realize that it was unused.

krasi-georgiev:
govendor support this as well `govendor list +unused` and has less false positive , only need to check for each supported OS which might be a blocker.

peterbourgon:
Oh no, I used this dependency. Does it exist somewhere else or shall I mirror it?

brian-brazil:
I think this is the only one, so I'd say mirror it.

Issue #835, opened 2015-06-23
Check upon config load whether alerting rule template fields can be parsed


juliusv:
A user got `error parsing template __alert_GlimpseDNSHighLatency: template: __alert_GlimpseDNSHighLatency:1: unexpected "*" in operand; missing space?` in the Alertmanager email message because they had `{{$labels.quantile * 100}}` in their alerting message template, which isn't valid in Go templates. Can we detect this sort of thing at configuration load time by test-parsing the SUMMARY and DESCRIPTION fields?

fabxc:
We can add a parse+execute cycle to the `promtool`. Parsing alone doesn't guarantee that much for templates.
I'll look into that later today.

codesome:
I have done some similar checks in https://github.com/prometheus/prometheus/pull/4277. Only catch is that it wont detect `$labels.something_wrong`, but would detect `$something_wrong`, and the one mentioned in the issue description.

I will open a PR for this.

PR #3854, opened 2018-02-16
bugfix: template: nil pointer dereference, e.g. if non-ASCII space

Dominik-K:
If you have a non-ASCII space, e.g. the [non-breaking space](https://en.wikipedia.org/wiki/Non-breaking_space) ` `, in a text template action, Prometheus returns as `label`/`annotation` value:

    <error expanding template: runtime error: invalid memory address or nil pointer dereference>

The correct parsing error should be

    error parsing template test: template: test:1: unexpected unrecognized character in action: U+00A0 in command

as the [Go `text/template` lexer only accepts ASCII spaces ` \t\r\n`](https://github.com/golang/go/blob/master/src/text/template/parse/lex.go#L98).

A failure on Prometheus start resp. `promtool check config` for this would be good, as well.

@brian-brazil

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -265,8 +265,7 @@ func (te Expander) Expand() (result string, resultErr error) {
 		}
 	}()
 
-	tmpl, err := text_template.New(te.name).Funcs(te.funcMap).Parse(te.text)
-	tmpl.Option("missingkey=zero")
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thanks!

#835 covers catching this sort of thing earlier.

Issue #3112, opened 2017-08-25
Reloading Prometheus 2.0 has significant memory spikes


fabxc:
Since every scraper has a rather large cache now for all staleness handling and usage of `AddFast` a reload causes severe memory spikes of up to 2x.

On reload we terminate all scrapers as the configuration might have changed. In 2.0, each scraper carries a rather large cache around for staleness handling and usage of `AddFast`. Those caches are dropped and get re-created for all the "new" targets. The same happens with the known buffer sizes for each scraper.

We could investigate not purging all scrapers and instead try to hot-reload them as well where necessary. This needs some very careful consideration though.

@brian-brazil

siavashs:
I noticed that the number of open files are doubled during reload (kept open ~5 minutes)
<img width="606" alt="start_-_reload" src="https://user-images.githubusercontent.com/113189/32500310-4b2d57a8-c3d5-11e7-8ebd-ed2c7613a007.png">
^ markers: start, reload

roidelapluie:
It looks like 910c22418c9a94e6ccfbc46dd760f5f392a25775 fixed the initial issue.

roidelapluie:
I have taken this approach over at #6670.

There are only a few fields changes that won't trigger a new cache:
- honor_timestamp
- scrape_interval
- scrape_timeout
- sample_limit
- http_config

Issue #3858, opened 2018-02-17
config reloads cause huge increase in memory


jhorwit2:
**What did you do?**

I updated the config in a way that should not noticeably impact cpu/memory (updated the scrape timeout from 5s to 10s). 

**What did you expect to see?**

I expected the scrape timeout to increase, which it did and I expected CPU/memory usage to remain the same.

**What did you see instead? Under which circumstances?**

Memory spikes at reload time

![image](https://user-images.githubusercontent.com/3876739/36342894-f2fc3c18-13d2-11e8-97ab-1d64ebdab8fa.png)

I've noticed this happen with any config change as did the memory usage. The series created/removed and appended samples per second had no fluctuation.

**Environment**

* System information:

Running the prometheus 2.2.0-RC0 docker container from dockerhub
Host is `Linux 4.1.12-112.14.13.el7uek.x86_64 x86_64`

* Prometheus version:

```
prometheus, version 2.2.0-rc.0 (branch: HEAD, revision: 1fe05d40e4b2f4f7479048b1cc3c42865eb73bab)
  build user:       root@f7abb25edc70
  build date:       20180213-11:40:47
  go version:       go1.9.2
```

jhorwit2:
I should add I've noticed this on 2.1.0 as well. It causes OOMs once prometheus has enough data.

brian-brazil:
This is expected when you change your target labels, we need to create new chunks for all the new metrics.

jhorwit2:
@brian-brazil That's understandable, but I'm not changing any target labels in this scenario. The only thing i updated was the global scrape timeout (i had nothing timing out before i just wanted to prove this is why it was increasing).

brian-brazil:
In that case this was probably coincidental, 15M new chunks is going to take ~15GB of RAM. That's what the graphs you originally had in this bug indicates to me.

jhorwit2:
@brian-brazil the reason I removed that graph is I've since run it a couple more times and don't see a correlation between memory chunks increasing and the memory increase.

jhorwit2:
![image](https://user-images.githubusercontent.com/3876739/36343546-6efa9d98-13db-11e8-9d34-4ccf803fb3c3.png)

![image](https://user-images.githubusercontent.com/3876739/36343551-7ee37cb6-13db-11e8-84a5-168eee68ea04.png)

[...]

PR #3861, opened 2018-02-19
scrape: pass scrape cache to replacement loops

fabxc:
On reload we restart all loops to adapt to changed configurations. This
discarded all scrape caches which caused memory spikes due to rebuilding
them all after reload.
We only start a loop after its predecessor has terminated successfully.
Thus we can safely pass down the scrape cache.

@brian-brazil I _think_ this is safe. Targets between loops, and thereby the target labels which are part of the cached label sets, do not change and thus all cache entries should remain to be correct.
Put another very careful walk through the cases would be good.

#3112

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"math"
 	"net/http"
+	"reflect"
 	"sync"
 	"time"
 	"unsafe"
@@ -130,7 +131,7 @@ type scrapePool struct {
 	cancel         context.CancelFunc
 
 	// Constructor for new scrape loops. This is settable for testing convenience.
-	newLoop func(*Target, scraper) loop
+	newLoop func(*Target, scraper, *scrapeCache) loop
 }
 
 const maxAheadTime = 10 * time.Minute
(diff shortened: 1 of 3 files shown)


krasi-georgiev:
should also close https://github.com/prometheus/prometheus/issues/3858

fabxc:
Oops, that was the one I actually wanted to mention :) Thanks

trnl:
@fabxc is it gonna be included in 2.2.0?

fabxc:
Depends on the review state – but we want to get 2.2 out rather sooner than later as it contains important fixes. So this being yet another non-trivial change, it might go out as a 2.3-beta instead.

scrape/scrape.go
```diff
@@ -197,6 +199,38 @@ func (sp *scrapePool) stop() {
 	wg.Wait()
 }
 
+func equalRelabelConfigs(as, bs []*config.RelabelConfig) bool {
```
brian-brazil: This seems a little fragile if we ever add more fields to actions. Could we use deep equal or move this into the relabel package?

krasi-georgiev:
should also close: https://github.com/prometheus/prometheus/issues/3112

roidelapluie:
I have taken this approach over at #6670

PR #3862, opened 2018-02-19
web: replace deprecated InstrumentHandler()

### LINKED ISSUE #3844: Switch from InstrumentHandler to InstrumentHandlerDuration
The current InstrumentHandler we are using in web.go is deprecated, we should replace it with InstrumentRoundTripperDuration.

simonpasquier:
This change replaces the deprecated InstrumentHandler function by the
equivalent functions from the promhttp package.

The following metrics are removed:

* http_request_duration_microseconds (Summary).
* http_request_size_bytes (Summary).
* http_requests_total (Counter).

And the following metrics are added instead:

* http_request_duration_seconds (Histogram).
* http_response_size_bytes (Histogram).
* promhttp_metric_handler_requests_in_flight (Gauge).
* promhttp_metric_handler_requests_total (Counter).

Fixes #3844.

In the initial version, I've chosen arbitrary bucket sizes for http_request_duration_seconds (0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10) and http_response_size_bytes (200, 600, 1800, 5400, 16200, 48600, 145800, 437400). Not sure if they are optimal...

### DIFF
--- a/web/web.go
+++ b/web/web.go
@@ -170,9 +170,19 @@ type Options struct {
 	EnableAdminAPI       bool
 }
 
+func instrumentHandler(handlerName string, handler http.HandlerFunc) http.HandlerFunc {
+	return promhttp.InstrumentHandlerDuration(
+		requestDuration.MustCurryWith(prometheus.Labels{"handler": handlerName}),
+		promhttp.InstrumentHandlerResponseSize(
+			responseSize.MustCurryWith(prometheus.Labels{"handler": handlerName}),
+			handler,
+		),
+	)
+}
+
 // New initializes a new web Handler.
 func New(logger log.Logger, o *Options) *Handler {
-	router := route.New()
+	router := route.New().WithInstrumentation(instrumentHandler)
 	cwd, err := os.Getwd()
 
 	if err != nil {
@@ -206,16 +216,6 @@ func New(logger log.Logger, o *Options) *Handler {
 		ready: 0,
 	}
 
-	instrf := func(handlerName string, handler http.HandlerFunc) http.HandlerFunc {
-		return promhttp.InstrumentHandlerDuration(
-			requestDuration.MustCurryWith(prometheus.Labels{"handler": handlerName}),
-			promhttp.InstrumentHandlerResponseSize(
-				responseSize.MustCurryWith(prometheus.Labels{"handler": handlerName}),
-				handler,
-			),
-		)
-	}
-
 	h.apiV1 = api_v1.NewAPI(h.queryEngine, h.storage, h.scrapeManager, h.notifier,
 		func() config.Config {
 			h.mtx.RLock()
@@ -224,7 +224,6 @@ func New(logger log.Logger, o *Options) *Handler {
 		},
 		o.Flags,
 		h.testReady,
-		instrf,
 		h.options.TSDB,
 		h.options.EnableAdminAPI,
 	)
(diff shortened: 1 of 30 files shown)


beorn7:
Sorry for the delay. If no maintainer of this repo picks this up or assigns it, I guess I can review it, but my review queue is pretty long right now.

web/web.go
```diff
@@ -67,6 +68,29 @@ import (
 
 var localhostRepresentations = []string{"127.0.0.1", "localhost"}
 
+var (
+	requestDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "http_request_duration_seconds",
+			Help:    "Histogram of latencies for HTTP requests.",
+			Buckets: []float64{.05, .1, .25, .5, 1, 2.5, 5, 10},
```
brian-brazil: Our timeout is 2m, we should probably go up at least that far

web/web.go
```diff
@@ -67,6 +68,29 @@ import (
...
+		},
+		[]string{"handler"},
+	)
+	responseSize = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "http_response_size_bytes",
+			Help:    "Histogram of response size for HTTP requests.",
+			Buckets: prometheus.ExponentialBuckets(200, 3.0, 8),
```
brian-brazil: A 100MB response is perfectly possible, I'd go for bigger jumps here and aim to top out around 1GB

web/web.go
```diff
@@ -67,6 +68,29 @@ import (
 
 var localhostRepresentations = []string{"127.0.0.1", "localhost"}
 
+var (
+	requestDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "http_request_duration_seconds",
```
brian-brazil: prometheus_http_request_duration_seconds

web/web.go
```diff
@@ -189,6 +223,7 @@ func New(logger log.Logger, o *Options) *Handler {
 		},
 		o.Flags,
 		h.testReady,
+		instrf,
```
brian-brazil: Passing a metric down to a dependency is a bit of a smell. Could we do this in the router rather than on each of the routes individually?

simonpasquier: Agree that passing the instrumentation handler to build the API isn't required. I'm not sure what you mean by "could we do this in the router".
What I have in mind is to pass the instrumentation handler to the `Register()` method since it is the only one needing the handler.

https://github.com/prometheus/prometheus/blob/7b7e3506acb5e0e5f19de26d89bf9029d6e86ddf/web/api/v1/api.go#L155-L157

to

```
func (api *API) Register(r *route.Router, instrh func(handlerName string, handler http.HandlerFunc) http.HandlerFunc) {
...
```

brian-brazil: I'd go for the route that is passed in to already have everything needed

simonpasquier: I opened https://github.com/prometheus/common/pull/120. Is that what you had in mind?

web/web.go
```diff
@@ -202,38 +236,36 @@ func New(logger log.Logger, o *Options) *Handler {
...
-	router.Get("/alerts", readyf(instrf("alerts", h.alerts)))
-	router.Get("/graph", readyf(instrf("graph", h.graph)))
-	router.Get("/status", readyf(instrf("status", h.status)))
-	router.Get("/flags", readyf(instrf("flags", h.flags)))
-	router.Get("/config", readyf(instrf("config", h.serveConfig)))
-	router.Get("/rules", readyf(instrf("rules", h.rules)))
-	router.Get("/targets", readyf(instrf("targets", h.targets)))
-	router.Get("/version", readyf(instrf("version", h.version)))
-	router.Get("/service-discovery", readyf(instrf("servicediscovery", h.serviceDiscovery)))
+	router.Get("/alerts", readyf(instrumentHandler("alerts", h.alerts)))
```
brian-brazil: Why do you have the instrumentHandler here rather than using WithInstrumentation?

simonpasquier: Because of the other handlers like `/-/*` and `/debug/*` which aren't instrumented ([link](https://github.com/simonpasquier/prometheus/blob/2db058636409e32d7ab2cbb4dd00ea7095d9aa94/web/web.go#L271-L303)).

brian-brazil: I don't see any harm in instrumenting them.

simonpasquier:
@brian-brazil please take a look.

brian-brazil:
This looks perfect. I'll merge tomorrow.

web/web.go
Dominik-K: @brian-brazil What's the reason for prefixing `http_request_duration` with  `prometheus` now? I consider it as a non-Prometheus specific metric and it is a [standardized best practice](https://github.com/prometheus/docs/blob/master/content/docs/practices/naming.md#metric-names).

brian-brazil: These metrics are coming from the "prometheus library" and thus should be prefixed accordingly, so they don't clash with http metrics from other libraries.

DominikGM: The meaning of a `http_request_duration` shouldn't differ no matter which library creates it, shall it? I think a label like `metric_producer="client_golang"` is sufficient to bug-hunt in case a library offers values with different meaning. That way, looking at Prometheus metrics could still serve as best practice 👍
Prometheus (normally) adds a `job="prometheus"` label to the metric anyway.

Dominik-K: The meaning of a `http_request_duration` shouldn't differ no matter which library creates it, shall it? I think a label like `metric_producer="client_golang"` is sufficient to bug-hunt in case a library offers values with different meaning. That way, looking at Prometheus metrics could still serve as best practice 👍
Prometheus (normally) adds a `job="prometheus"` label to the metric anyway.

brian-brazil: I would be extremely surprised if it didn't differ. Each library has different semantics, different labels, and different middleware. The same request might hit multiple such libraries. Accordingly this does not make sense as one metric separated by labels.
In addition a metric should almost always live inside one file, largely for the above reasons.

PR #3863, opened 2018-02-19
Add OS type meta label to Azure SD

pharaujo:
There is currently no way to differentiate Windows instances from Linux ones in Azure SD. This is needed when you have a mix of node_exporters / wmi_exporters for OS-level metrics and you want to have them in separate scrape jobs.

This change allows you to do just that. Example:

```
  - job_name: 'node'
    azure_sd_configs:
      - <azure_sd_config>
    relabel_configs:
      - source_labels: [__meta_azure_machine_os_type]
        regex: Linux
        action: keep
```

The way the vendor'd AzureSDK provides to get the OsType is a bit awkward - as far as I can tell, this information can only be gotten from the startup disk. Newer versions of the SDK appear to improve this a
bit (by having OS information in the InstanceView), but the current way still works in them.

As per MAINTAINERS.md, can either @brian-brazil or @fabxc review this please?

### DIFF
--- a/discovery/azure/azure.go
+++ b/discovery/azure/azure.go
@@ -40,6 +40,7 @@ const (
 	azureLabelMachineID            = azureLabel + "machine_id"
 	azureLabelMachineResourceGroup = azureLabel + "machine_resource_group"
 	azureLabelMachineName          = azureLabel + "machine_name"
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thanks!

PR #3864, opened 2018-02-19
api: Added v1/status/flags endpoint.

Bplotka:
Endpoint URL: `/api/v1/status/flags`
Example Output:
```json
{
  "status": "success",
  "data": {
    "alertmanager.notification-queue-capacity": "10000",
    "alertmanager.timeout": "10s",
    "completion-bash": "false",
    "completion-script-bash": "false",
    "completion-script-zsh": "false",
    "config.file": "my_cool_prometheus.yaml",
    "help": "false",
    "help-long": "false",
    "help-man": "false",
    "log.level": "info",
    "query.lookback-delta": "5m",
    "query.max-concurrency": "20",
    "query.timeout": "2m",
    "storage.tsdb.max-block-duration": "36h",
    "storage.tsdb.min-block-duration": "2h",
    "storage.tsdb.no-lockfile": "false",
    "storage.tsdb.path": "data/",
    "storage.tsdb.retention": "15d",
    "version": "false",
    "web.console.libraries": "console_libraries",
    "web.console.templates": "consoles",
    "web.enable-admin-api": "false",
    "web.enable-lifecycle": "false",
    "web.external-url": "",
    "web.listen-address": "0.0.0.0:9090",
    "web.max-connections": "512",
    "web.read-timeout": "5m",
    "web.route-prefix": "/",
    "web.user-assets": ""
  }
}
```

### DIFF
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -382,6 +382,79 @@ $ curl http://localhost:9090/api/v1/alertmanagers
 }
 ```
 
+## Status
+
+Following status endpoints expose current Prometheus configuration.
+
+### Config
+
+The following endpoint returns currently loaded configuration file:
+
+```
+GET /api/v1/status/config
+```
+
+Currently, config is returned as dumped YAML file.
+
+```json
+$ curl http://localhost:9090/api/v1/status/config
+{
+  "status": "success",
+  "data": {
+    "yaml": "<content of the loaded config file in YAML>",
+  }
+}
+```
+
+### Flags
+
+The following endpoint returns flag values that Prometheus was configured with:
+
+```
+GET /api/v1/status/flags
+```
+
+All values are in a form of "string".
+
+```json
+$ curl http://localhost:9090/api/v1/status/flags
+{
+  "status": "success",
+  "data": {
+    "alertmanager.notification-queue-capacity": "10000",
+    "alertmanager.timeout": "10s",
+    "completion-bash": "false",
+    "completion-script-bash": "false",
+    "completion-script-zsh": "false",
+    "config.file": "/etc/prometheus/prometheus.yaml",
+    "help": "false",
+    "help-long": "false",
+    "help-man": "false",
+    "log.level": "info",
+    "query.lookback-delta": "5m",
+    "query.max-concurrency": "20",
+    "query.timeout": "2m",
+    "storage.tsdb.max-block-duration": "36h",
+    "storage.tsdb.min-block-duration": "2h",
+    "storage.tsdb.no-lockfile": "false",
+    "storage.tsdb.path": "data/",
+    "storage.tsdb.retention": "15d",
+    "version": "false",
+    "web.console.libraries": "console_libraries",
+    "web.console.templates": "consoles",
+    "web.enable-admin-api": "false",
+    "web.enable-lifecycle": "false",
+    "web.external-url": "",
+    "web.listen-address": "0.0.0.0:9090",
+    "web.max-connections": "512",
+    "web.read-timeout": "5m",
+    "web.route-prefix": "/",
+    "web.user-assets": ""
+  }
+}
+```
+
+*New in v2.2*
 
 ## TSDB Admin APIs
 These are APIs that expose database functionalities for the advanced user. These APIs are not enabled unless the `--web.enable-admin-api` is set.
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -394,7 +394,8 @@ The following endpoint returns currently loaded configuration file:
 GET /api/v1/status/config
 ```
 
-Currently, config is returned as dumped YAML file.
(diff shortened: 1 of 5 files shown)


Bplotka:
What maybe is worth to consider as well is to preserve some basic types from flags itself instead of treating everything as `string`. That was just easier, since everything is ready in `web.go`. 

I think this is fine, but let me know, I can try to treat `int` and `booleans` flags properly here.

fabxc:
Fine with me. Though this is now basically adding stuff to what we consider the "Promtheus compatibility API". Might get tricky later if we convolute those APIs.

Maybe move the Prometheus query API under a dedicated prefix?
I'd also probably vote for having all our own APIs in gRPC and _optionally_ provide an unstable JSON mapping for those that we consider relevant to query via `curl` etc. during debugging (_not_ through grpc-rest-gateway though).

Bplotka:
@brian-brazil What exactly documentation you have in mind? Even `api/v1/status/config` seems to be undocumented. Can add some info, but please let me know where (: There are some bits around API here: https://github.com/prometheus/prometheus/blob/master/docs/querying/api.md but only related to querying itself.

brian-brazil:
Yes, it should be on that page.

Bplotka:
Done.

docs/querying/api.md
```diff
@@ -382,6 +382,79 @@ $ curl http://localhost:9090/api/v1/alertmanagers
...
+
+### Config
+
+The following endpoint returns currently loaded configuration file:
+
+```
+GET /api/v1/status/config
+```
+
+Currently, config is returned as dumped YAML file.
```
brian-brazil: The "currently" doesn't add anything here, we won't be changing that as it'd be a breaking change. You might mention though that comments are not present, due to limitations of the yaml library we use.

docs/querying/api.md
```diff
@@ -382,6 +382,79 @@ $ curl http://localhost:9090/api/v1/alertmanagers
...
+```
+
+All values are in a form of "string".
+
+```json
+$ curl http://localhost:9090/api/v1/status/flags
+{
+  "status": "success",
+  "data": {
+    "alertmanager.notification-queue-capacity": "10000",
```
brian-brazil: There's no need to list all of the flags here, just a small sampling will do.

docs/querying/api.md
```diff
@@ -382,6 +382,79 @@ $ curl http://localhost:9090/api/v1/alertmanagers
...
+    "alertmanager.timeout": "10s",
+    "completion-bash": "false",
```
brian-brazil: This is not a flag of Prometheus, why is it here?

fabxc: That and things like `help-man` are probably coming in through the flag library.
Maybe that can be disabled – otherwise we can just blacklist those.

docs/querying/api.md
fabxc: Not sure I understand. Why would we "sample" the configured flags we are showing.

brian-brazil: I meant only in the docs, otherwise we get to update this every time we get new flags.

docs/querying/api.md
Bplotka: Well.. This and same others (e.g `help`) are all kingpin flags, so they are included when we are dumping them into `flagsMap`.

UI `/flags` page shows them too. 
image

To be consistent I think we should leave them, or if we want we can filter some out on both API and flag page. What do you think?

docs/querying/api.md
Bplotka: I can agree with @brian-brazil here, it is just an example.

docs/querying/api.md
Bplotka: @brian-brazil comment still valid

Bplotka:
OK, Excluded kingpin flags from flags API and flags UI page. Found (I think) easy way to do it.
Now it looks like this:
```json
  "alertmanager.notification-queue-capacity": "10000",
  "alertmanager.timeout": "10s",
  "config.file": "p.yaml",
  "log.level": "info",
  "query.lookback-delta": "5m",
  "query.max-concurrency": "20",
  "query.timeout": "2m",
  "storage.tsdb.max-block-duration": "36h",
  "storage.tsdb.min-block-duration": "2h",
  "storage.tsdb.no-lockfile": "false",
  "storage.tsdb.path": "data/",
  "storage.tsdb.retention": "15d",
  "web.console.libraries": "console_libraries",
  "web.console.templates": "consoles",
  "web.enable-admin-api": "false",
  "web.enable-lifecycle": "false",
  "web.external-url": "",
  "web.listen-address": "0.0.0.0:9090",
  "web.max-connections": "512",
  "web.read-timeout": "5m",
  "web.route-prefix": "/",
  "web.user-assets": ""
```

brian-brazil:
Thanks!

Issue #3809, opened 2018-02-07
Remote Storage Queue won't flush


fmoessbauer:
**What did you do?**

Update prometheus to commit 404b306fb9d5c0976cc83f0dff39254c0918493e

**What did you expect to see?**

Remote storage working properly.

**What did you see instead? Under which circumstances?**

Writes to the remote storage are not flushed. The remote_storage_queue fills up until no more elements fit in and then samples are dropped.

This is probably related to PR #3731 

![remote-storage-queue-annotate](https://user-images.githubusercontent.com/9655545/35919684-298513b4-0c16-11e8-8868-f5f8821ee666.png)

**Environment**

* System information:

`Linux 4.13.0-25-generic x86_64`

* Prometheus version:

```
prometheus, version 2.1.0 (branch: master, revision: 404b306fb9d5c0976cc83f0dff39254c0918493e)                                            
  build user:       root@3dc6e252e6c2                                
  build date:       20180207-10:31:30                                
  go version:       go1.9.2
```

* Prometheus configuration file:
```
global:
  scrape_interval:     15s # By default, scrape targets every 15 seconds.
  external_labels:
    monitor: 'codelab-monitor'

scrape_configs:
  - job_name: 'prometheus'

    scrape_interval: 15s

    static_configs:
      - targets: ['localhost:9090', 'collectd-exporter:9103']

  - job_name: 'linux-server'
    scrape_interval: 60s
    static_configs:
      - targets:
        - '192.168.2.xx:xxxx'

# [...]

rule_files:
  - alert-rules/*.rule

alerting:
  alertmanagers:
    - static_configs:
      - targets:
        - alertmanager:9093

# Remote write configuration (for Graphite, OpenTSDB, or InfluxDB).
remote_write:
  - url: "http://prom-io:9201/write"

# Remote read configuration (for InfluxDB only at the moment).
remote_read:
  - url: "http://prom-io:9201/read"
```

* Logs:
```
level=warn ts=2018-02-07T13:55:17.944112632Z caller=queue_manager.go:225 component=remote msg="Remote storage queue full, discarding sample. Multiple subsequent messages of this kind may be suppressed."
```

krasi-georgiev:
@tomwilkie might give some ideas here

juliusv:
Both I and @tomwilkie can reproduce this from master, so it's also in 2.2.

AndreaGiardini:
Same thing here. V2.2 is not able to flush remote_storage queue

[...]

PR #3731, opened 2018-01-24
Re-use timer in remote storage queue

bboreham:
Current code is allocating a new `Timer` object and `chan Time` on every sample enqueued for remote storage. The [docs](https://golang.org/pkg/time/#After) for `time.After()` note that:

> The underlying `Timer` is not recovered by the garbage collector until the timer fires. If efficiency is a concern, use `NewTimer` instead and call `Timer.Stop` if the timer is no longer needed.

To illustrate the impact:
```
.../prometheus/storage/remote$ go test -run=TestSampleDelivery -memprofile=mem.out
PASS
ok  	github.com/prometheus/prometheus/storage/remote	1.416s
.../prometheus/storage/remote$ go tool pprof -inuse_space -top mem.out
File: remote.test
Type: inuse_space
Time: Jan 24, 2018 at 12:40pm (UTC)
Showing nodes accounting for 19070.72kB, 100% of 19070.72kB total
      flat  flat%   sum%        cum   cum%
15361.12kB 80.55% 80.55% 16482.56kB 86.43%  time.NewTimer /usr/local/go/src/time/sleep.go
 1121.44kB  5.88% 86.43%  1121.44kB  5.88%  time.startTimer /usr/local/go/src/runtime/time.go
 1024.56kB  5.37% 91.80%  1024.56kB  5.37%  reflect.mapassign /usr/local/go/src/runtime/hashmap.go
[...]
```

Same test with this change:
```
$ go tool pprof -inuse_space -top mem.out
File: remote.test
Type: inuse_space
Time: Jan 24, 2018 at 12:45pm (UTC)
Showing nodes accounting for 3602.65kB, 100% of 3602.65kB total
      flat  flat%   sum%        cum   cum%
 1554.07kB 43.14% 43.14%  1554.07kB 43.14%  github.com/prometheus/prometheus/vendor/github.com/beorn7/perks/quantile.NewTargeted /home/vagrant/src/github.com/prometheus/prometheus/vendor/github.com/beorn7/perks/quantile/stream.go
 1024.38kB 28.43% 71.57%  1024.38kB 28.43%  reflect.mapassign /usr/local/go/src/runtime/hashmap.go
```

PR #3867, opened 2018-02-20
fix reset of timer in remote storage queue

fmoessbauer:
This PR fixes a bug introduced in #3731 and reported in #3809.

The timer has to be resetted in both cases, as otherwise the for loop does not continue if the queue is flushed due to a timeout. The proposed solution is race-free, as in both cases the timer has already expired before it is resetted.

@tomwilkie

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -454,13 +454,13 @@ func (s *shards) runShard(i int) {
 			if !timer.Stop() {
 				<-timer.C
 			}
-			timer.Reset(s.qm.cfg.BatchSendDeadline)
 		case <-timer.C:
 			if len(pendingSamples) > 0 {
 				s.sendSamples(pendingSamples)
 				pendingSamples = pendingSamples[:0]
 			}
 		}
+		timer.Reset(s.qm.cfg.BatchSendDeadline)
 	}
 }
 


tomwilkie:
I’m not sure why this is necessary, as the timer will be set when there are samples to send, and not sent when there isn’t. 

But I could be missing something. @bboreham?

jkohen:
@tomwilkie Before this change, if the <-timer.C case is entered, the timer isn't reset, but the channel is drained, so if timer.Stop() returns false and L454 tries to read from timer.C, it gets stuck.

bboreham:
Oops, my bad.  This change LGTM.

bmoyles0117:
A comment on the resetting of the timer. It would seem that the timer should only be stopped, cleared, and reset, only if a sample was sent. 

The code as it stands here would reset the timer after every sample, potentially delaying the delivery of samples by `sample delivery rate * MaxSamplesPerSend`. I don't think this is a desirable outcome, and leads to a lot of churn on the timer unnecessarily.

I've sent a PR as an example of what I'm trying to say: https://github.com/prometheus/prometheus/pull/3907

bboreham:
As I read it, it will send every `MaxSamplesPerSend` or, if no more samples arrive on the chan, at the timeout.

fmoessbauer:
@bmoyles0117 From a first glance at your code, I guess it does only cover some rare corner cases. In the example, `sentSamples` will almost always be true and hence the timer is reset. Did you profile your code to measure the benefits of the additional LOC?

For me, both solutions are ok, however if there is not a great difference in performance, I prefer to merge the simpler version, considering the maintainablility.

bboreham:
I think it makes a difference when you have a steady trickle of samples. Eg one new sample every second - reset the timer every second - it will only send after `MaxSamplesPerSend` seconds. 

It was like that before I broke it, but still a valid point.

fmoessbauer:
@bboreham Ah, thanks for this clarification. That's true. If #3907 passes CI, we could close my PR in favor of #3907 which already covers the fix.

PR #3868, opened 2018-02-20
Prevent invalid label names with labelmap

### LINKED ISSUE #3850: Invalid target labelnames?
Per https://groups.google.com/forum/#!topic/prometheus-users/4JIBSa1zyDA it appears that a user on 2.0.0 has managed to have target labels containing hyphens in the label name. This indicates a lack of validation somewhere.

simonpasquier:
This change ensures that the relabeling configurations using labelmap
can't generate invalid label names. ~~It also checks that the label names
generated by labelmap actions are valid.~~

Fixes #3850.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -636,6 +636,9 @@ var expectedErrors = []struct {
 		filename: "labeldrop5.bad.yml",
 		errMsg:   "labeldrop action requires only 'regex', and no other fields",
 	}, {
+		filename: "labelmap.bad.yml",
+		errMsg:   "\"l-$1\" is invalid 'replacement' for labelmap action",
+	}, {
 		filename: "rules.bad.yml",
 		errMsg:   "invalid rule file path",
 	}, {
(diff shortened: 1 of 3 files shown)


pkg/relabel/relabel.go
```diff
@@ -79,8 +79,10 @@ func relabel(lset labels.Labels, cfg *config.RelabelConfig) labels.Labels {
 	case config.RelabelLabelMap:
 		for _, l := range lset {
 			if cfg.Regex.MatchString(l.Name) {
-				res := cfg.Regex.ReplaceAllString(l.Name, cfg.Replacement)
-				lb.Set(res, l.Value)
+				res := model.LabelName(cfg.Regex.ReplaceAllString(l.Name, cfg.Replacement))
+				if res.IsValid() {
+					lb.Set(string(res), l.Value)
```
brian-brazil: To be consistent, we should really be sanitizing this.

config/config_test.go
```diff
@@ -635,6 +635,9 @@ var expectedErrors = []struct {
 	}, {
 		filename: "labeldrop5.bad.yml",
 		errMsg:   "labeldrop action requires only 'regex', and no other fields",
+	}, {
+		filename: "labelmap.bad.yml",
+		errMsg:   "\"l-$1\" is invalid 'replacement' for labelmap action",
```
brian-brazil: This indicates that this won't even parse (which would be fine too), so the new relabel code wouldn't apply. I don't see code changes implementing this though.

brian-brazil: Yes, I'd just keep the config.

simonpasquier: Done.

brian-brazil:
Thanks!

PR #3872, opened 2018-02-21
Place Service Discovery labels inline

### LINKED ISSUE #3739: Long label values hard to read on service discovery status page
Some k8 SD labels like `__meta_kubernetes_pod_annotation_kubernetes_io_created_by` can be very long, making the service discovery status page scroll horizontally.

We should really wrap these, such as by removing the `whitespace: nowrap` and adding `display: inline`.

callumj:
screenshot 2018-02-20 20 39 59

Fixes #3739


web/ui/templates/service-discovery.html
```diff
@@ -11,6 +11,10 @@
     height: 65px;
     visibility: hidden;
   }
+
+  .label {
```
callumj: Happy to move this out into the dedicated CSS pages if required, though I suspect then we'll want to scope this down to some service discovery class?

brian-brazil: It should be in the main CSS file, and as labels are displayed in many places we probably want this global.

callumj: @brian-brazil Moved this into the `prometheus.css` file, I checked across the UI and it seems to work nicely.

brian-brazil:
Can you confirm this looks okay if one label is really really long?

callumj:
Sorry about that @brian-brazil, here it is in Targets and SD:

screenshot 2018-02-24 11 41 41
screenshot 2018-02-24 11 41 35

callumj:
Sorry about that @brian-brazil, here it is in Targets and SD. I think it doesn't look as good on Targets but given the constraints of that page (and the use case) it is still able to do most of the job.

screenshot 2018-02-24 11 41 41
screenshot 2018-02-24 11 41 35

brian-brazil:
Thanks!

PR #3878, opened 2018-02-22
Copy the regex object before using it

jkohen:
This avoids lock contention when many goroutines are relabeling metrics based on the same configuration.

I've seen this contention with several thousand targets and a pretty simple configuration based on the Kubernetes examples in the Prometheus documentation. As part of my load test I didn't notice any meaningful increase in CPU usage.

### DIFF
--- a/pkg/relabel/relabel.go
+++ b/pkg/relabel/relabel.go
@@ -48,9 +48,9 @@ func relabel(lset labels.Labels, cfg *config.RelabelConfig) labels.Labels {
 
 	lb := labels.NewBuilder(lset)
 	// Copying the regex avoids contention when this runs from multiple goroutines.
-	var regex regexp.Regexp
+	var regex *regexp.Regexp
 	if cfg.Regex.Regexp != nil {
-		regex := cfg.Regex.Copy()
+		regex = cfg.Regex.Copy()
 	}
 
 	switch cfg.Action {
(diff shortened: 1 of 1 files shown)


brian-brazil:
Why would there be lock contention for a regex? This code is on a hot path, so extra allocations are not desirable.

jkohen:
Brian, thanks for the quick response. I'm seeing contention in [Regexp.get](https://golang.org/src/regexp/regexp.go#L210), in the internal cache for machine objects. As per Regexp.Copy's documentation, the Go team expected contention here, but I also understand the concern about allocations in the hot path.

In other platforms I would stick the Regexp object in a CPU-local variable, but it seems Go doesn't have such facility. Prometheus conveniently wraps Regexp objects in config.Regexp. What do you think about changing config.Regexp.Regexp to a sync.Pool of Regexp instead? It should be quick work.

In the meantime, I'll reach out to the Go team and propose changing the internal machine cache to use sync.Pool.

jkohen:
Brian, I wanted to give you a heads-up that I have been unable to reproduce this. I had failed to notice that the process was being CPU throttled, and that may have triggered the issue. Without a reliable reproduction, I'm happy to drop this, unless you think it's worth digging deeper.

brian-brazil:
We can wait to see if anyone else hits it.

jkohen:
For future reference, I fixed the lock contention upstream, and it's expected to be part of Go 1.11: https://github.com/golang/go/issues/24411

PR #3879, opened 2018-02-22
alerting: Set StartedAt to the time the alerti is active, not firing

roidelapluie:
It makes more sense to send to the Alertmanager the time the alert has
really started, not the time it fires.

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -672,7 +672,7 @@ func sendAlerts(n *notifier.Manager, externalURL string) rules.NotifyFunc {
 				continue
 			}
 			a := &notifier.Alert{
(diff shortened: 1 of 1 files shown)


brian-brazil:
This makes sense, but given the issues we've had with 2.x and alerting I'd say we hold off until after 2.2 to merge this.

brian-brazil:
Hmm, this should really have a unittest.

roidelapluie:
I honestly do not know how I can write a nice test for it. This is a function that returns a function, that takes a complex input as parameters.

simonpasquier:
After more feedback, this isn't needed. See https://github.com/prometheus/prometheus/pull/4910#issuecomment-443221848

PR #3885, opened 2018-02-24
Race in sd triton test

krasi-georgiev:
I can't reproduce locally , but should fix the fail  when test -race is enabled
https://travis-ci.org/prometheus/prometheus/jobs/345641319

### DIFF
--- a/discovery/triton/triton_test.go
+++ b/discovery/triton/triton_test.go
@@ -86,7 +86,11 @@ func TestTritonSDRun(t *testing.T) {
 	assert.Nil(t, err)
 	assert.NotNil(t, td)
 
-	go td.Run(ctx, ch)
+	wait := make(chan struct{})
+	go func() {
+		td.Run(ctx, ch)
+		close(wait)
+	}()
 
 	select {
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
@brian-brazil

brian-brazil:
I don't see how this fixes a race. Where's the concurrent access that this is fixing?

krasi-georgiev:
Read
https://github.com/prometheus/prometheus/blob/ffa673f7d813d7c8f4f1677aca9859340d4d4629/discovery/triton/triton_test.go#L89-L90

Write
https://github.com/prometheus/prometheus/blob/ffa673f7d813d7c8f4f1677aca9859340d4d4629/discovery/triton/triton_test.go#L168

Although I don't see why this happens since the tests don't run in parallel , I think it happens because these share the same config via the pointer.

@brian-brazil can you think of any other explanation why is this happening?

brian-brazil:
The context is cancelled, but the code isn't waiting for `Run` to actually terminate.

krasi-georgiev:
Thanks, added an extra blocking channel for this.

PR #3892, opened 2018-02-27
Check for unused vendored packages

krasi-georgiev:
usnig govendor to check for unused packages


krasi-georgiev:
@brian-brazil , @SuperQ all green

brian-brazil:
Is this our canonical Makefile or is it the one in promu we should be changing first?

krasi-georgiev:
I will let you decide. this is simply checking the code base doesn't have any unused vendored packages. Similar to the checkstatic and checkstyle

krasi-georgiev:
@brian-brazil  promu seems to do only builds so I think this is not related and should stay here.

brian-brazil:
My point is more that we shouldn't be hand-editing all the Makefiles and letting them drift across repos. We should have one canonical Makefile, and copy it everywhere else. My question is whether the prometheus or the promu Makefile is meant to be the canonical one.

krasi-georgiev:
I makes sense to make this one the source , but how do you think we can auto sync it with the other ones?

krasi-georgiev:
^ @brian-brazil

brian-brazil:
It'd be copied by hand every now and then. I'm more worried by different features/behaviours in different repos than some repos having an old Makefile.

krasi-georgiev:
so maybe put a comment somewhere in the Makefile as a reminder that any reusable parts should be copied from https://github.com/prometheus/prometheus/blob/master/Makefile ?

brian-brazil:
My point is that the entire file should be reusable and always copied from one place.

PR #3893, opened 2018-02-27
Cleanup licence check script

SuperQ:
* `echo -e` is not supported by POSIX shell.
* Add a license to the check script.
* Split out license check loop into function for clarity.
* Exit 1.

### DIFF
--- a/scripts/check_license.sh
+++ b/scripts/check_license.sh
@@ -1,10 +1,29 @@
 #!/bin/sh
+#
+# Copyright 2018 The Prometheus Authors
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


scripts/check_license.sh
```diff
@@ -1,10 +1,29 @@
...
-licRes=$(
-for file in $(find . -type f -iname '*.go' ! -path './vendor/*'); do
-	head -n3 "${file}" | grep -Eq "(Copyright|generated|GENERATED)" || echo -e "  ${file}"
-done;)
 if [ -n "${licRes}" ]; then
-	echo -e "license header checking failed:\n${licRes}"
-	exit 255
+  echo "license header checking failed:"
+  echo "${licRes}"
+  exit 255
```
brian-brazil: An exit code of 1 would make more sense I think

SuperQ: Yea, I have no idea why it's 255.

sdurrheimer:
There is `promu check licenses` command which was meant to replace the shellscript, I guess we should begin to work on using it.

PR #3896, opened 2018-02-28
Reset the timer after reading from its channel

jkohen:
Before this change, if the <-timer.C case is entered, the timer isn't reset, but the channel is drained, so if timer.Stop() returns false and L454 tries to read from timer.C, it gets stuck.


brian-brazil:
@tomwilkie

brian-brazil:
Dupe of #3867

PR #3899, opened 2018-03-01
revert to using the gogofast plugin and regenerate grpc server

krasi-georgiev:
@fabxc


fabxc:
Thanks 👍  Just to complete this, what was the original intention regarding switching it?

PR #3900, opened 2018-03-01
Display annotations on Alerts page

### LINKED ISSUE #1219: Alerts page should display expanded ANNOTATIONS
Even though the expansion happens in Prometheus, it is only _visible_ in Alertmanager. It should be available on `/alerts`.

roidelapluie:
Fixes #1219

### DIFF
--- a/web/ui/templates/alerts.html
+++ b/web/ui/templates/alerts.html
@@ -5,7 +5,12 @@
 
 {{define "content"}}
 <div class="container-fluid">
-  <h2>Alerts</h2>
+  <h1>Alerts</h1>
+  <div class="filters">
+     <button type="button" class="targets btn btn-primary show-annotations">
+       <i class="glyphicon glyphicon-unchecked"></i> Show annotations
+     </button>
+  </div>
   <table class="table table-bordered table-collapsed">
     <tbody>
     {{$alertStateToRowClass := .AlertStateToRowClass}}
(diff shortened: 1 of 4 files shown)


roidelapluie:
anno

brian-brazil:
Thanks, these clutter up the page quite a bit. Can you hide them by default or something?

roidelapluie:
They are hidden by default, with a checkbox on the top.

brian-brazil:
It's not really "Labels and Annotations" if the annotations aren't there by default. Maybe have them as another column that's hidden by default?

roidelapluie:
That will not fit a screen width. What about changing the title of the column of it is enabled?

roidelapluie:
Updated

roidelapluie:
anno
It looks better now

web/ui/templates/alerts.html
```diff
@@ -38,6 +43,21 @@ <h2>Alerts</h2>
               <td>{{.ActiveAt.UTC}}</td>
               <td>{{.Value}}</td>
             </tr>
+                {{ if .Annotations.Map}}
+                  <tr style="display:none" class="alert_annotations">
```
brian-brazil: The <tr should be aligned with the others

roidelapluie: I have aligned the code

simonpasquier:
:+1:

brian-brazil:
Thanks!

PR #3901, opened 2018-03-01
xrate, take two

free:
This is the xrate() V2 implementation, [as described](https://groups.google.com/forum/#!topic/prometheus-developers/B_CMEp40PHE) on the Prometheus Developers list and as illustrated by screenshots at https://free.github.io/xrate/index.html

It is not intended to be accepted as is, but as a base for discussion and so that anyone interested could try it out.


brian-brazil:
As discussed elsewhere, this is not going to be accepted.

PR #3905, opened 2018-03-02
Properly clean up timer on flush

jkohen:
This PR fixes a bug introduced in #3731

I believe this code at HEAD is leaking resources when the runShard() exits, e.g. on reshard. This change works for me locally and I see less CPU usage than before after a few reshardings. Both my low QPS and high QPS deployments seem to be working correctly with this change.

@tomwilkie

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -430,6 +430,11 @@ func (s *shards) runShard(i int) {
 	pendingSamples := model.Samples{}
 
 	timer := time.NewTimer(s.qm.cfg.BatchSendDeadline)
+	defer func() {
+		if !timer.Stop() {
+			<-timer.C
+		}
+	}()
 
 	for {
 		select {


jkohen:
Retrying Travis.

tomwilkie:
Thanks @jkohen.  Would it be easier (and also correct) to do a non-blocking read from the channel after the stop?  There is also some discussion going on here: https://github.com/prometheus/prometheus/pull/3907

jkohen:
@tomwilkie I'm not sure what you have in mind, can you show in a PR? The `defer` deals with the return buried inside the loop. I'm coding defensively in case someone adds another return point later in this function; the timer API is clearly tricky to get right.

I'm aware of #3907, thanks. @bmoyles identified that other issue when I showed him this fix.

tomwilkie:
Sorry @jkohen I confused this with a different issue.  I'll try and get to this soon!

tomwilkie:
@jkohen I've put what I mean in #3941

tomwilkie:
Superseded by #3941

PR #3907, opened 2018-03-02
Only reset the timer when samples are sent or the timer expires.

bmoyles0117:


storage/remote/queue_manager.go
```diff
@@ -446,19 +446,25 @@ func (s *shards) runShard(i int) {
 			queueLength.WithLabelValues(s.qm.queueName).Dec()
 			pendingSamples = append(pendingSamples, sample)
 
+			sentSamples := len(pendingSamples) >= s.qm.cfg.MaxSamplesPerSend
+
 			for len(pendingSamples) >= s.qm.cfg.MaxSamplesPerSend {
 				s.sendSamples(pendingSamples[:s.qm.cfg.MaxSamplesPerSend])
 				pendingSamples = pendingSamples[s.qm.cfg.MaxSamplesPerSend:]
```
jkohen: Just  for readability I'd make sentSamples = true here, and avoid the complex condition above the loop. Otherwise, this makes sense, thanks for pointing this out.

bmoyles0117: Done, thanks for the feedback!

bboreham:
Ok, you want the timer to run from the time the first pending sample arrived, rather than the last. That does seem better.

tomwilkie:
I think another way of saying that is the invariant we want is that no sample can be delayed for more than `BatchSendDeadline`?

If so, wouldn't it be easier to reset the timer after the call to `sendSamples`?  There shouldn't be any problem resetting the timer multiple times if multiple batches are sent immediately, nor should that actually happen.  Then we can keep the `Stop`/`Reset` logic as is (with a no blocking read from the channel) and we won't need the extra `Reset` on the timer firing.

tomwilkie:
@bmoyles0117  I've put my ideas here: #3941

tomwilkie:
Superseded by #3941

Issue #3067, opened 2017-08-14
Validation for all SD configs


Gouthamve:
There is no validation for some SD configurations which might lead to issues like: #3066 The KubernetesSD config is a good example where validation happens: https://github.com/prometheus/prometheus/blob/master/config/config.go#L1049-L1065

This should be done for every SD mechanism.

krasi-georgiev:
DIBS 

There is no pending PR for this right?

serathius:
@krasi-georgiev Are you still working on it? if not can I take it over?

krasi-georgiev:
@serathius , nope you can go ahead, I got lost in other PR's

serathius:
https://github.com/prometheus/prometheus/pull/3911

serathius:
@Gouthamve I added validation based on prometheus documentation, it should be enough to eliminate case in https://github.com/prometheus/prometheus/issues/3066 
Should we consider more aggressive version? Writing full rules for openstack optional flags looks impossible without breaking someone.

simonpasquier:
@gouthamve what do you think of closing this issue? The scope being unclear, it is impossible to declare that it is done.

palash25:
ping

PR #3911, opened 2018-03-05
Validate SD configuration

serathius:
This PR adds validation to SD configuration so it matches required fields described in SD documentation.
I avoided validating all the fields, because some clients extract that information from environment variables or have own defaults. This should avoid breaking users in case of errors, if they configured SD based on documentation they should have those fields already filled. 

fixes https://github.com/prometheus/prometheus/pull/3911

### DIFF
--- a/discovery/azure/azure.go
+++ b/discovery/azure/azure.go
@@ -86,6 +86,9 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err != nil {
 		return err
 	}
+	if c.SubscriptionID == "" {
+		return fmt.Errorf("Azure SD configuration requires a subscription_id")
+	}
 
 	return yaml_util.CheckOverflow(c.XXX, "azure_sd_config")
 }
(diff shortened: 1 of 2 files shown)


docs/configuration/configuration.md
```diff
@@ -387,8 +387,8 @@ See below for the configuration options for EC2 discovery:
 ```yaml
 # The information to access the EC2 API.
 
-# The AWS Region.
-region: <string>
+# The AWS Region. If blank, will try to fetch from metadata service.
```
brian-brazil: This is not a guaranteed behaviour of Prometheus, merely something the library we are using does itself.

serathius: I disagree https://github.com/serathius/prometheus/blob/master/discovery/ec2/ec2.go#L99

brian-brazil: It's still not something we guarantee, configuration is meant to come entirely from the config file. That code really shouldn't be there.

krasi-georgiev:
@serathius  wrong fixes ref , should be #3067

brian-brazil:
CircleCI failed, let's try a rebuild.

krasi-georgiev:
this one is  failing randomly and makes no sense what is happening there.

brian-brazil:
Thanks!

PR #3912, opened 2018-03-05
discovery: coalesce identical SD configurations

### LINKED ISSUE #2191: Consider coalescing identical SD definitions

simonpasquier:
Closes #2191.

For now, it is still early stages but it looks like it is working in my environment. I'm going to test with lots of identical k8s configs and check if we see noticeable changes on Prometheus/k8s sides.
cc @brian-brazil @krasi-georgiev @fabxc.

### DIFF
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -249,7 +249,8 @@ func (m *Manager) allGroups() map[string][]*targetgroup.Group {
 }
 
 func (m *Manager) registerProviders(cfg sd_config.ServiceDiscoveryConfig, setName string) {
-	add := func(t string, cfg interface{}, df func() (Discoverer, error)) {
+	add := func(cfg interface{}, newDiscoverer func() (Discoverer, error)) {
+		t := reflect.TypeOf(cfg).String()
 		for _, p := range m.providers {
 			if reflect.DeepEqual(cfg, p.config) {
 				p.subs = append(p.subs, setName)
@@ -257,7 +258,7 @@ func (m *Manager) registerProviders(cfg sd_config.ServiceDiscoveryConfig, setNam
 			}
 		}
 
-		d, err := df()
+		d, err := newDiscoverer()
 		if err != nil {
 			level.Error(m.logger).Log("msg", "Cannot create service discovery", "err", err, "type", t)
 			failedConfigs.Inc()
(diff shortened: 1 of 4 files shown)


krasi-georgiev:
I will let @Gouthamve have a look at this one if he has time  as a present for his return :)

I am tackling some races in tsdb.

discovery/manager.go
```diff
@@ -59,8 +60,18 @@ type poolKey struct {
 	provider string
 }
 
+type provider struct {
+	name   string
+	d      Discoverer
+	sets   []string
```
cstyan: it's not entirely clear to me what sets are supposed to be

discovery/manager.go
```diff
@@ -203,85 +222,134 @@ func (m *Manager) allGroups() map[string][]*targetgroup.Group {
...
-	providers := map[string]Discoverer{}
-
-	app := func(mech string, i int, tp Discoverer) {
-		providers[fmt.Sprintf("%s/%d", mech, i)] = tp
+func (m *Manager) registerProviders(cfg sd_config.ServiceDiscoveryConfig, setName string) {
+	register := func(cfg interface{}) bool {
+		for _, p := range m.providers {
+			if reflect.DeepEqual(cfg, p.config) {
+				p.sets = append(p.sets, setName)
+				//level.Debug(m.logger).Log("msg", "Registering new set for provider", "provider", p.name, "sets", fmt.Sprintf("%v", p.sets))
```
cstyan: did you want to keep these log lines?

simonpasquier: Thanks for spotting!

discovery/manager.go
simonpasquier: I used `sets` because `poolKey` has `setName` but I agree that the naming can be improved.

cstyan: okay so then `setName` is either scrape config `jobName` or the md5 checksum of the AlertmanagerConfig as JSON

simonpasquier: Correct

brancz:
lgtm

solidspark:
+1, we could use this.

grobie:
Lgtm, but I think we should fix #4124 first.

discovery/manager.go
```diff
@@ -210,85 +234,132 @@ func (m *Manager) allGroups() map[string][]*targetgroup.Group {
...
+			config: cfg,
+			subs:   []string{setName},
+		}
+		m.providers = append(m.providers, &provider)
 	}
 
-	for i, c := range cfg.DNSSDConfigs {
-		app("dns", i, dns.NewDiscovery(*c, log.With(m.logger, "discovery", "dns")))
+	for _, c := range cfg.DNSSDConfigs {
+		if register(c) {
```
brian-brazil: I think it might be clearer to put all this into one function, rather than splitting across two

simonpasquier:
@brian-brazil I've added `prometheus_sd_failed_configs`, a gauge metric with a `name` label that is either `scrape` or `notify`.

brian-brazil:
That's not a great metric, I'd suggest an unlabelled counter.

simonpasquier:
I thought about having a counter first but I didn't see how it could be useful for alerting as the value wouldn't change after the configuration has been fixed.

brian-brazil:
The problem with what's here is that once there's a failure it'll stick. In general with SD we expose failure count metrics, so I think we should stick with that.

brian-brazil:
:+1:

discovery/manager.go
```diff
@@ -210,85 +248,96 @@ func (m *Manager) allGroups() map[string][]*targetgroup.Group {
...
+		for _, p := range m.providers {
+			if reflect.DeepEqual(cfg, p.config) {
+				p.subs = append(p.subs, setName)
+				return
+			}
+		}
+
+		d, err := df()
+		if err != nil {
+			level.Error(m.logger).Log("msg", "Cannot create service discovery", "err", err, "type", t)
```
krasi-georgiev: instead of passing `t string` can't we use  something like `reflect.TypeOf(cfg).Elem()`

discovery/manager.go
```diff
@@ -210,85 +248,96 @@ func (m *Manager) allGroups() map[string][]*targetgroup.Group {
...
+func (m *Manager) registerProviders(cfg sd_config.ServiceDiscoveryConfig, setName string) {
+	add := func(t string, cfg interface{}, df func() (Discoverer, error)) {
+		for _, p := range m.providers {
+			if reflect.DeepEqual(cfg, p.config) {
+				p.subs = append(p.subs, setName)
+				return
+			}
+		}
+
+		d, err := df()
```
krasi-georgiev: wouldn't be better to call this something more descriptive like: `NewDiscoverer`

discovery/manager.go
```diff
@@ -146,7 +181,9 @@ func (m *Manager) runProvider(ctx context.Context, poolKey poolKey, updates chan
 			if !ok {
 				return
 			}
-			m.updateGroup(poolKey, tgs)
+			for _, s := range p.subs {
+				m.updateGroup(poolKey{setName: s, provider: p.name}, tgs)
```
krasi-georgiev: why not use the `p *provider` as a pool key directly?  I used a seperate struct just to make it unique enough, but now with the `type provider struct` maybe we can use it directly and remove the `poolKey type` entirely.

simonpasquier: Hmm we still need to dispatch the target groups from 1 provider to N jobs at some point.

simonpasquier:
I didn't test specifically with `-race` but as you said, this code tends to be racy so will do.

brian-brazil:
You've got a conflict

brian-brazil:
Anyone have any final comments before I merge?

brian-brazil:
Those issues are all from code introduced after this PR, and I don't think
they should be blocking it. Rather we should be thinking about reverting
them as they're causing issues.

krasi-georgiev:
> Those issues are all from code introduced after this PR, and I don't think
they should be blocking it. Rather we should be thinking about reverting
them as they're causing issues.

@brian-brazil I am talking about fixing  #4124 first. Which is the one you are referring to?

@grobie  there are 2 issues
1. exhaustion of open HTTP2 streams
   -  this PR mitigates the problem, but doesn't guarantee that you won't hit the exhaustion. A proper fix would be to revert the  http2  package, before https://github.com/golang/net/commit/1c05540f6879653db88113bc4a2b70aec4bd491f and wait for a fix upstream.
2. Processing the scrape reloading in serial which takes long to process so on systems with constant changes the reloading can never keep up.
   - fixed in #4526

krasi-georgiev:
@brian-brazil  any objections if  I open a PR to revert the http package and possibly add a test to catch exhaustion of open HTTP2 streams regression.

govendor fetch golang.org/x/net/...@167db312be3be3d239e24dbe0ab099ab5b9aa783

brian-brazil:
> Which is the one you are referring to?

I'd have to dig it up, but I have gotten the impression that there have been a series of changes each of which introduced new issues. If we can't get that under control, maybe we should be considering a revert.

krasi-georgiev:
the only change that I am aware of is for the  throttling in https://github.com/prometheus/prometheus/pull/4523 (merged) and a small fix in  https://github.com/prometheus/prometheus/pull/4556 (waiting for a review)

I would like to move forward with the fix and can do a RC right after that, but if you prefer to revert that is fine by me.

brian-brazil:
Those changes broke static_configs per #4551, if we're missing breakage to such basic functionality then what else are we missing?

krasi-georgiev:
LGTM for this PR , thanks @simonpasquier

brian-brazil:
> #4556 confirms to fix the behaviour, but as I said if you prefer to revert that is fine.

I'm not saying we have to revert, but we should be taking due care with such key and relatively intricate code. It seems every time someone touches it that it breaks.

PR #3913, opened 2018-03-06
Refactor/redesign Unhealthy checkbox on Targets page

jamtur01:
I've simplified the selection box into a toggle for All/Unhealthy and reduced the code complexity. More in keeping with the Show More/Show less buttons and looks prettier too IMHO. :)

screenshot 2018-03-06 02 12 37


brian-brazil:
I'm not sure this is worth the clutter, if you have very few healthy targets and need filtering you have bigger problems.

jamtur01:
@brian-brazil Sure. Fair enough. I've insetad simplified the selection box into a toggle for All/Unhealthy and significantly reduced the code complexity.

screenshot 2018-03-06 02 12 37

fabxc:
What do you mean by clutter @brian-brazil it's actually less code and looks better.

brian-brazil:
The first version of the PR wasn't as nice as this one.

fabxc:
Are we okay with the updated version then?

juliusv:
Just giving it a spin as well now.

juliusv:
Works fine for me except for the fact that our buttons normally have no rounded corners. Should we style this one equally (even though it's a slightly different input element than a normal button)?

jamtur01:
@brian-brazil Yes apologies. I should have raised a new PR. :(

brian-brazil:
There's no problem evolving a PR like this, the edit to the top comment just caused a little bit of confusion.

jamtur01:
@juliusv I was actually curious about the square/round distinction - I note that the content in tables in targets have:

screenshot 2018-03-08 10 46 31

Rounded corners. But the buttons do not. I personally like rounded (and it avoids an override of bootstrap) and I'd be happy to make it consistent across the UI if folks agreed. Otherwise happy to make them square too.

juliusv:
@jamtur01 Yep, the labels and health indicators aren't input elements, so I guess they have some more justification for looking different. I'd recommend just making it square for now (basically extend https://github.com/prometheus/prometheus/blob/master/web/ui/static/css/graph.css#L146-L148 to match your element too?). If we generally want to change things to a round look, it's probably better done in a separate PR. Just looking at this, I'm not 100% sold yet (though I haven't tried also rounding the output display box yet):

round

jamtur01:
@juliusv It'd definitely be a new PR.

I like the clean feel of rounded elements but it's definitely a personal taste.

I'll square them for now.

juliusv:
:+1: Thanks!

Issue #1481, opened 2016-03-09
Show big and fat warning on web status page upon certain conditions.


beorn7:
Like dirty storage, quarantining of series, too high persist pressure, too many memory chunks...

Ideally with a little explanation what's happening and what can be done.

beorn7:
@fabxc fallout from our discussion

fabxc:
Sounds good.

RichiH:
Arguably, certain machine-level metrics like available disk space, system load, heavy swap/full memory should, high IO wait, and possibly others fall into the same category of "the user should really handle this themselves, but this is really bad and we need to yell about it".

beorn7:
@RichiH Those are machine level metrics you want to monitor anyway, no matter if Prometheus runs on your machine or not. Also, those metrics are not even accessible for the Prometheus server binary.

The conditions this issue refers to are Prometheus specific. You should still monitor it in you meta-monitoring, too. But putting on the status page if something is fundamentally wrong within Prometheus sounds like a good idea.

There is no intent, though, to turn the status page into a system-health monitor.

dominikschulz:
Maybe a little off-topic, but it fits this discussion: Is that status page suitable as a health check URL (e.g. for K8s Liveness Probes)?

juliusv:
@dominikschulz This would just be part of the normal web UI. It wouldn't return a non-200 if it visually reports some errors there. So not really.

Generally, you can use the availability of the web UI in a limited way as a startup health probe, as the web UI is started as the last thing on Prometheus startup. However, it won't tell you much about whether Prometheus is generally healthy, only that it has finished starting up.

brian-brazil:
See also #1468

Gouthamve:
Is this still relevant given 2.0?

brian-brazil:
The storage less so, but there's still things like rule groups taking longer than their interval.

RichiH:
It probably changes a bit. For example isolation might want to let the users know if there's a _really_ old write hanging somewhere or, as Brian beat me to writing, recording rules taking longer than they should.

RichiH:
Though @Gouthamve 's implicit question is valid; there should be a new list of things to handle.

roidelapluie:
Easy ones:

- Prometheus failed to reload its configuration
- WAL corruptions

beorn7:
Some thoughts about it:
- Let's only consider this for the new UI at this point.
[...]

PR #3921, opened 2018-03-07
Add prometheus runtime information on status page

### LINKED ISSUE #1468: Show gomaxprocs on status page
It'd be helpful to have access to golang runtime information somewhere.

serathius:
This PR adds information about tsdb, wal and config reload to status page

fixes #1468
See also #1481
cc @beorn7 @brian-brazil


cmd/prometheus/main.go
```diff
@@ -61,14 +61,12 @@ import (
 
 var (
 	configSuccess = prometheus.NewGauge(prometheus.GaugeOpts{
-		Namespace: "prometheus",
-		Name:      "config_last_reload_successful",
-		Help:      "Whether the last configuration reload attempt was successful.",
+		Name: "prometheus_config_last_reload_successful",
```
beorn7: Note that this is equivalent to the old version. Not sure what the motivation behind the change here is. I leave it to @brian-brazil to decide if it makes sense to change it at this time.

cmd/prometheus/main.go
```diff
@@ -61,14 +61,12 @@ import (
...
+		Help: "Whether the last configuration reload attempt was successful.",
 	})
 	configSuccessTime = prometheus.NewGauge(prometheus.GaugeOpts{
-		Namespace: "prometheus",
-		Name:      "config_last_reload_success_timestamp_seconds",
-		Help:      "Timestamp of the last successful configuration reload.",
+		Name: "prometheus_config_last_reload_success_timestamp_seconds",
```
beorn7: See above.

web/web.go
```diff
@@ -52,6 +53,7 @@ import (
 	"github.com/prometheus/tsdb"
 	"golang.org/x/net/netutil"
 
+	"github.com/prometheus/client_model/go"
```
beorn7: Nit: This should be in the previous block. (We only put packages from the same Git repo in the last block.)

cmd/prometheus/main.go
serathius: As you said it's equivalent in for metrics, but it will allow to search code by string and find match. I thinks it's also only case where namespace is used.

brian-brazil: I personally prefer having the full name for greppability.

web/ui/templates/status.html
```diff
@@ -13,6 +13,26 @@ <h2 id="runtime">Runtime Information</h2>
...
+        <tr>
+          <th>Configuration reload</th>
+          <td>{{if .ReloadConfigSuccess}}Succesfull{{else}}Failed{{end}}</td>
+        </tr>
+        <tr>
+          <th>Last successfull configuration reload</th>
+          <td>{{.LastConfigTime.UTC}}</td>
+        </tr>
+        <tr>
+          <th>Number of chunks</th>
```
brian-brazil: These are head chunks and time series

web/ui/templates/status.html
```diff
@@ -13,6 +13,26 @@ <h2 id="runtime">Runtime Information</h2>
           <th>Working Directory</th>
           <td>{{.CWD}}</td>
         </tr>
+        <tr>
+          <th>Configuration reload</th>
+          <td>{{if .ReloadConfigSuccess}}Succesfull{{else}}Failed{{end}}</td>
+        </tr>
+        <tr>
+          <th>Last successfull configuration reload</th>
```
brian-brazil: successful

web/ui/templates/status.html
```diff
@@ -13,6 +13,26 @@ <h2 id="runtime">Runtime Information</h2>
           <th>Working Directory</th>
           <td>{{.CWD}}</td>
         </tr>
+        <tr>
+          <th>Configuration reload</th>
+          <td>{{if .ReloadConfigSuccess}}Succesfull{{else}}Failed{{end}}</td>
```
brian-brazil: Can we do some colouring on this to make it stand out?

web/ui/templates/status.html
```diff
@@ -13,6 +13,26 @@ <h2 id="runtime">Runtime Information</h2>
...
+        <tr>
+          <th>Number of time series</th>
+          <td>{{.TimeSeriesCount}}</td>
+        </tr>
+        <tr>
+          <th>Number of WAL corruptions</th>
+          <td>{{.CorruptionCount}}</td>
+        </tr>
         <tr>
           <th>Number of goroutines</th>
```
brian-brazil: You can just call this Goroutines

web/ui/templates/status.html
serathius: I added changing row color to red
https://www.w3schools.com/bootstrap/bootstrap_tables.asp

cmd/prometheus/main.go
beorn7: I don't feel strongly either way. I was just confused about the change here as it is unrelated to the purpose of this PR.
If @brian-brazil is fine with it, there is no need to take it out, I guess.

web/ui/templates/status.html
```diff
@@ -13,8 +13,28 @@ <h2 id="runtime">Runtime Information</h2>
           <th>Working Directory</th>
           <td>{{.CWD}}</td>
         </tr>
+        <tr{{if not .ReloadConfigSuccess}} class="danger"{{end}}>
+          <th>Configuration reload</th>
+          <td>{{if .ReloadConfigSuccess}}Successful{{else}}Failed{{end}}</td>
+        </tr>
+        <tr>
+          <th>Last successfull configuration reload</th>
```
beorn7: There is still a typo here: successfull → successful

web/ui/templates/status.html
```diff
@@ -13,8 +13,28 @@ <h2 id="runtime">Runtime Information</h2>
...
+        <tr{{if not .ReloadConfigSuccess}} class="danger"{{end}}>
+          <th>Configuration reload</th>
+          <td>{{if .ReloadConfigSuccess}}Successful{{else}}Failed{{end}}</td>
+        </tr>
+        <tr>
+          <th>Last successful configuration reload</th>
+          <td>{{.LastConfigTime.UTC}}</td>
+        </tr>
+        <tr>
+          <th>Number of head chunks</th>
```
brian-brazil: Number of can be removed from these too

web/web.go
```diff
@@ -556,23 +558,63 @@ func (h *Handler) graph(w http.ResponseWriter, r *http.Request) {
...
+		GOMAXPROCS          int
+		GOGC                string
+		CorruptionCount     int64
+		ChunkCount          int64
+		TimeSeriesCount     int64
+		LastConfigTime      time.Time
+		ReloadConfigSuccess bool
 	}{
-		Birth:          h.birth,
+		// Match resolution of configuration reload time to avoid case where birth is after reload
```
brian-brazil: Why are you doing this?

serathius: metric for reloading configuration is using `SetToCurrentTime` which is rounded to whole seconds. 
To avoid confusion I Truncate birth same way, so there is never case when reload time < birth time

brian-brazil: That doesn't sound right, the code shows it has ns resolution.

serathius: https://github.com/prometheus/client_golang/blob/master/prometheus/gauge.go#L51
https://github.com/prometheus/client_golang/blob/master/prometheus/gauge.go#L99

brian-brazil: That has ns resolution. It's converted to a float before the division, and if it wasn't we'd fix it to be that way.

serathius:
@brian-brazil can you confirm if everything is ok?

brian-brazil:
That looks okay. We're holding off on non-critical changes at the moment while we're getting things stable, so I'll merge this after that's done.

brian-brazil:
Can you resolve conflicts please?

brian-brazil:
Thanks!

PR #3923, opened 2018-03-07
WIP: Change docker build to multi-stage

kavehmz:
Current docker image build depends on binaries built outside of container.
It depends on compile enviroment and result might not be completely preditable.

Using multi-stage build will make the result more predicatle.

https://docs.docker.com/develop/develop-images/multistage-build/

This change will require Docker 17.05 or higher and refactoring circleci setup to 2.0.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,19 @@
+FROM golang:1 AS build
+
+RUN apt-get update && apt-get install -y build-essential
+
+COPY    . /go/src/github.com/prometheus/prometheus/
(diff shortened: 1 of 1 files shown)


sdurrheimer:
Hi there,

Our current build process ensure the binary in the docker image is the same as the binary in the github release.

While building the project manually might depend on the compile enviroment, our compile environnement use stable docker images to build our releases.

IMO we don't need multi-stage docker builds right now.

kavehmz:
Hi, 
That makes sense.

PR #3928, opened 2018-03-08
Add option to skip head on snapshots

fabxc:
@brian-brazil 

Since we (for reasons still unclear to me) back-ported the snapshot stuff to /v1 I actually had to invert the flag. Feels a bit inconsistent but seems like the best way for now.


fabxc:
@SuperQ Codacy back again from your account it seems. What kind of cancer is this =/

SuperQ:
@fabxc Yea, it's pretty awful.  Are we trying to keep this Codacy thing, should I try and disconnect it?

brian-brazil:
We've been trying to get rid of it for a while.

brian-brazil:
Can you update the docs?

fabxc:
Done.

docs/querying/api.md
```diff
@@ -452,9 +452,10 @@ We also expose a gRPC API whose definition can be found [here](https://github.co
 
 ### Snapshot
 Snapshot creates a snapshot of all current data into `snapshots/<datetime>-<rand>` under the TSDB's data directory and returns the directory as response.
+It will optionally skip snapshotting data that is only present in memory and the write ahead log.
```
brian-brazil: This is hard to understand. Maybe: skip snapshotting data that is only present in the head block, and which has not yet been written out as a block.

fabxc:
done

brian-brazil:
:+1:

PR #3929, opened 2018-03-08
Add target_label comparison to keep/drop relabel actions.

tcolgate:
This allows for comparison of the concatenated source_labels against
the value of the label specified by target_label in the keep/drop
actions.


tcolgate:
This would help to resolve issues around duplicate scrapes with the example kubernetes  configuration (and I believe any possible existing configuration that didn't require micromanagement of scrape configs)

The keep / drop action will fire if either the regexp match, or the  named target_label has a matching value.

tcolgate:
Currently the implementation contradicts the documentation in that regex replacement is not supported in the target_label, it potentially could be. I thing things can still work.

brian-brazil:
Per discussion on the list, this is to workaround some form of mis-configuration and doesn't make sense to add.

PR #3941, opened 2018-03-09
Correctly stop the timer used in the remote write path.

### LINKED ISSUE #3809: Remote Storage Queue won't flush

tomwilkie:
Fixes #3809, supersedes #3905 and #3907 

Also, instead of resetting the timer every time we scrape a sample (potentially delaying them indefinitely), only reset when we send a batch.

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -430,14 +430,15 @@ func (s *shards) runShard(i int) {
 	pendingSamples := model.Samples{}
 
 	timer := time.NewTimer(s.qm.cfg.BatchSendDeadline)
-	defer func() {
+	stop := func() {
 		if !timer.Stop() {
 			select {
 			case <-timer.C:
 			default:
 			}
 		}
-	}()
+	}
+	defer stop()
 
 	for {
 		select {
@@ -454,16 +455,11 @@ func (s *shards) runShard(i int) {
 			queueLength.WithLabelValues(s.qm.queueName).Dec()
 			pendingSamples = append(pendingSamples, sample)
 
-			for len(pendingSamples) >= s.qm.cfg.MaxSamplesPerSend {
+			if len(pendingSamples) >= s.qm.cfg.MaxSamplesPerSend {
 				s.sendSamples(pendingSamples[:s.qm.cfg.MaxSamplesPerSend])
 				pendingSamples = pendingSamples[s.qm.cfg.MaxSamplesPerSend:]
 
-				if !timer.Stop() {
-					select {
-					case <-timer.C:
-					default:
-					}
-				}
+				stop()
 				timer.Reset(s.qm.cfg.BatchSendDeadline)
 			}
 
@@ -472,6 +468,7 @@ func (s *shards) runShard(i int) {
 				s.sendSamples(pendingSamples)
 				pendingSamples = pendingSamples[:0]
 			}
+			timer.Reset(s.qm.cfg.BatchSendDeadline)
 		}
 	}
 }
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -430,6 +430,14 @@ func (s *shards) runShard(i int) {
 	pendingSamples := model.Samples{}
 
 	timer := time.NewTimer(s.qm.cfg.BatchSendDeadline)
+	defer func() {
+		if !timer.Stop() {
+			select {
+			case <-timer.C:
+			default:
+			}
+		}
+	}()
 
 	for {
 		select {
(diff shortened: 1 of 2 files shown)


storage/remote/queue_manager.go
```diff
@@ -449,11 +457,16 @@ func (s *shards) runShard(i int) {
 			for len(pendingSamples) >= s.qm.cfg.MaxSamplesPerSend {
 				s.sendSamples(pendingSamples[:s.qm.cfg.MaxSamplesPerSend])
 				pendingSamples = pendingSamples[s.qm.cfg.MaxSamplesPerSend:]
+
+				if !timer.Stop() {
```
bboreham: DRY - these six lines are the same as the `func()` you have defined at the top - just name it and call it.

bboreham: It occurred to me we don't really want to do this in a loop, but then I realised the loop can never execute more than once since we add samples one at a time.  Maybe it shouldn't be a loop, or should be commented as a special kind of loop.

storage/remote/queue_manager.go
```diff
@@ -430,6 +430,14 @@ func (s *shards) runShard(i int) {
 	pendingSamples := model.Samples{}
 
 	timer := time.NewTimer(s.qm.cfg.BatchSendDeadline)
+	defer func() {
+		if !timer.Stop() {
+			select {
+			case <-timer.C:
+			default:
+			}
```
jkohen: Nit, the Golang docs suggest a direct read from the timer, any reason to use this more verbose version?
if !t.Stop() {
	<-t.C
}
https://golang.org/pkg/time/#Timer.Stop

bboreham: This is to fix the bug that #3907 fixes - the `default:` option stops it stalling on an empty queue.

jkohen: It's a minor question, but this code is already rather complex. In what case is it unknown whether the timer channel is empty?

storage/remote/queue_manager.go
tomwilkie: I made it a special kind of loop - an `if` loop.

bboreham: I was talking about the `for len(pendingSamples) >= s.qm.cfg.MaxSamplesPerSend` loop

bboreham: oops, see what you mean

tomwilkie:
Thanks @bboreham, have factored that out and made the `for` an `if`.

Also thanks @jkohen, thats the best explanation of what #3907 is trying to fix, I didn't realise this.  

> In what case is it unknown whether the timer channel is empty?

When we run the sendSamples for the first case of the select, we don't know if the second case has been triggered or not.

> PS: is it too weird to suggest a unit test? :-)

Not at all.  I'll see if I can write one, and why the existing ones didn't catch this.

storage/remote/queue_manager_test.go
```diff
@@ -122,6 +125,44 @@ func TestSampleDelivery(t *testing.T) {
 	c.waitForExpectedSamples(t)
 }
 
+func TestSampleDeliveryTimeout(t *testing.T) {
+	// Let's send on less sample than batch size, and wait the timeout duration
```
simonpasquier: (nit) s/on/one/

jkohen:
LGTM, thanks for the test!

PR #3942, opened 2018-03-09
Correctly handle pruning wraparound after ring expansion

### LINKED ISSUE #3939: Range selector over 16s is broken
Apologies for the weirdly specific and yet vague issue title. Essentially what I see happening is that when a function that takes a range selector (e.g. `rate()`) is used in a `query_range` request with a range selector of 16s over the start of a series with 1s resolution, it appears as if it will be fed the first 16 points in the series, then (as if those first 16 points suddenly disappeared), the following 1 point, 2 points, 3 points and so on. So essentially the rate graph looks like this:

rate foo 16s

The 17s rate (and most other range values except 15s, 31s, 32s, 61s...) look correct:

rate foo 17s

**What did you do?**

Did a `/query_range?query=rate(gen_counter[16s])&step=1` over the start of a counter with 1s resolution, incrementing by 1 every second.

A smooth increase from 0 to 1 over the first 16 seconds, followed by a flat rate of 1.

A smooth increase from 0 to 1, followed by a missing value, followed by a smooth increase from 0 to 1, followed by a flat rate of 1.

**Environment**

* System information:

	Linux 4.13.0-36-generic x86_64

* Prometheus version:

	prometheus, version 2.2.0 (branch: HEAD, revision: f63e7db4cbdb616337ca877b306b9b96f7f4e381)
  build user:       root@52af9f66ce71
  build date:       20180308-16:40:42
  go version:       go1.10

```
# Global configuration
global:
  scrape_interval:     1s
  evaluation_interval: 1s

  external_labels:
      monitor: 'prometheus'

rule_files:
  - 'rules/*.rules.yml'
```

* rules/generator.rules.yml:
```
groups:
- name: generator.rules
  rules:

  - record: gen_helper
    expr: 0
    labels:
      job: gen_test

  - record: gen_counter
    expr: (gen_counter + 1) or gen_helper
```

To reproduce, dump the contents of the config directory in the attached [config.tar.gz](https://github.com/prometheus/prometheus/files/1796586/config.tar.gz) into a fresh Prometheus 2.2.0 installation, fire up Prometheus, then go to http://localhost:9090/graph?g0.range_input=1m&g0.expr=rate(gen_counter%5B16s%5D)&g0.tab=0&g1.range_input=1h&g1.expr=gen_counter%5B100s%5D&g1.tab=1 and keep refreshing until you get the first half minute of the newly generated series covered by the `rate(gen_counter[16s])` graph. I suspect the problem is somewhere in `promql/engine.go` but I couldn't figure out where exactly.

brian-brazil:
Fixes #3939

### DIFF
--- a/storage/buffer.go
+++ b/storage/buffer.go
@@ -178,6 +178,7 @@ func (r *sampleRing) add(t int64, v float64) {
 		r.buf = buf
 		r.i = r.f
 		r.f += l
+		l = 2 * l
 	} else {
 		r.i++
 		if r.i >= l {
--- a/storage/buffer_test.go
+++ b/storage/buffer_test.go
@@ -48,6 +48,11 @@ func TestSampleRing(t *testing.T) {
 			delta: 7,
 			size:  1,
 		},
+		{
+			input: []int64{1, 2, 3, 4, 6},
+			delta: 4,
+			size:  4,
+		},
 	}
 	for _, c := range cases {
 		r := newSampleRing(c.delta, c.size)


simonpasquier:
:+1: works in my setup.

PR #3949, opened 2018-03-12
rulefmt: supply filename of malformed content if possible

adamdecaf:
Otherwise errors like the following are printed, which don't give much
to help debug.

```
level=error ts=2018-03-12T18:43:18.392113733Z caller=manager.go:485 component="rule manager" msg="loading groups failed" err="yaml: line 17: found unexpected end of stream"
```

### DIFF
--- a/pkg/rulefmt/rulefmt.go
+++ b/pkg/rulefmt/rulefmt.go
@@ -176,9 +176,17 @@ func checkOverflow(m map[string]interface{}, ctx string) error {
 
 // Parse parses and validates a set of rules.
 func Parse(content []byte) (*RuleGroups, []error) {
+	return parse(content, "")
+}
+
+func parse(content []byte, filename string) (*RuleGroups, []error) {
 	var groups RuleGroups
 	if err := yaml.Unmarshal(content, &groups); err != nil {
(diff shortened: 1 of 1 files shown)


adamdecaf:
cc @fabxc

pkg/rulefmt/rulefmt.go
```diff
@@ -176,9 +176,17 @@ func checkOverflow(m map[string]interface{}, ctx string) error {
...
+}
+
+func parse(content []byte, filename string) (*RuleGroups, []error) {
 	var groups RuleGroups
 	if err := yaml.Unmarshal(content, &groups); err != nil {
-		return nil, []error{err}
+		if filename == "" {
+			return nil, []error{err}
+		} else {
+			return nil, []error{fmt.Errorf("filename=%s, err=%v", filename, err)}
```
adamdecaf: Will this make an error like: `err="filename="foo" err="yaml: line 17: found unexpected end of stream""`?

gouthamve:
Hi thanks for submitting the PR! Sorry for dropping the ball on this, but it is now already fixed in master.

PR #3950, opened 2018-03-13
Cherrypick #3941 "Correctly stop the timer used in the remote write path."

tomwilkie:
See #3941


fabxc:
Squash it this time.

brian-brazil:
We should keep the commits the same across the two branches, otherwise it'll be annoying to work with.

PR #3952, opened 2018-03-13
*: cut 2.2.1

fabxc:
@brian-brazil


fabxc:
@brian-brazil updated notes

brian-brazil:
We'll want https://github.com/prometheus/tsdb/pull/301 and https://github.com/prometheus/tsdb/pull/300 too I think

Issue #2573, opened 2017-04-04
Remote read should handle down or misbehaving backends gracefully


brian-brazil:
I'm playing a bit with remote read, and it looks like any failure on the part of the remote read endpoint (e.g. down, malformed response, wrong number of result series) causes the query as a whole to fail.

When any remote read backend fails, the query should continue to work on the rest of the data available. 

We may also in future want to report this out to the user via the query api, metrics etc.

FYI @tomwilkie as you mentioned this being relevant for Cortex.

tomwilkie:
Yes we have the same problem in Cortex, in that we want to return partial results in the even of partial failures (for instance, if S3 is down we still have recent data in the ingesters).  Cortex issue: https://github.com/weaveworks/cortex/issues/309

There doesn't seem to be a particularly appropriate HTTP status code for this - some references imply 207 (http://www.restpatterns.org/HTTP_Status_Codes/207_-_Multi-Status) is appropriate, others that it should be 500 with a body.  WDYT?

brian-brazil:
I'd have a standard 200, and the extra information in an warning field in the response.

krasi-georgiev:
DIBS

krasi-georgiev:
@brian-brazil just to make sure I understand the issue here is what I did
set to write and read from influxdb , than I stop the influxdb server and when running any query I am getting
```
{"status":"error","errorType":"internal","error":"error sending request: Post http://localhost:9201/read: dial tcp [::1]:9201: getsockopt: connection refused"}
```

this shouldn't happen as the local tsdb should be able to provide some samples if available and the the http header should be:
`HTTP/1.1 206 Partial Content`

brian-brazil:
The http response code should be 200. A 206 is something different.

Gouthamve:
This should also handle broken checksums in TSDB.

brian-brazil:
That's a bit of a different problem, but may share the warning plumbing.

sipian:
@brian-brazil @krasi-georgiev 

Hi,
I read about this on GSOC'18 ideas page and would like to contribute to it as part of GSOC.

After going through all references, I came to the following conclusions. Please do correct me if my interpretation is wrong.
1. From [this](https://github.com/prometheus/prometheus/pull/3421), @brian-brazil proposed to handle this in fanout storage, using the primary and secondary storage .

After going through the code, I came to the following conclusions about it's implementation
[...]

PR #3963, opened 2018-03-14
promql: propagate storage errors

atereshchenkov:
`populateIterators()` does not propagate storage errors. Looks like it was introduced in 83cd270e after updating Querier interface.

@fabxc Please review.


fabxc:
Thanks 👍 ! I think that fixes an oddity I've been observing a few times now :)

krasi-georgiev:
what about the case when the remote errors, but the local tsdb can provide the result? 
I had  a quick look and it seems that an error in either Querer errors the entire result.

brian-brazil:
That's #2573

krasi-georgiev:
yes this is the one I was referring to.

PR #3966, opened 2018-03-14
Optimise PromQL

brian-brazil:
I started this last week, and while I'm not finished I want to share it as it's already a large change.

This reworks PromQL to be 5x faster for `rate(x[5m])` with hundred of series and thousands of steps and generate less than 1% of the garbage.
The main approach is to evaluate entire nodes at once rather than step by step, which is better for locality, puts better bounds on memory usage, allows reusing more point slices, and allows for caching of label processing across steps within a node.

Instant evaluations on small numbers of series might be a few % slower due to new overheads, however these were fast to start with and everything else is faster. At this stage almost all of the allocations are in the storage layer, which will be a bit trickier to tackle.

Unittests were kept passing throughout these changes, and I've also added some new ones, so I'm reasonably confident this doesn't break anything. I also fixed two PromQL bugs in passing.

Full benchmark results are attached.
[cmp.txt](https://github.com/prometheus/prometheus/files/1812478/cmp.txt)


krasi-georgiev:
Wow , great , work , don't have to review, but Well done!

brian-brazil:
I've done all the main stuff I want to do in PromQL itself, so this is ready for review.

promql/engine.go
```diff
@@ -606,88 +579,73 @@ func (ev *evaluator) recover(errp *error) {
...
-func (ev *evaluator) evalOneOf(e Expr, t1, t2 ValueType) Value {
-	val := ev.eval(e)
-	if val.Type() != t1 && val.Type() != t2 {
-		ev.errorf("expected %s or %s but got %s", documentedType(t1), documentedType(t2), documentedType(val.Type()))
+// signatureFunc is a cached version of signatureFunc.
+func (enh *evalNodeHelper) signatureFunc(on bool, names ...string) func(labels.Labels) uint64 {
+	if enh.sigf == nil {
+		enh.sigf = make(map[uint64]uint64, len(enh.out))
+	}
+	f := signatureFunc(on, names...)
```
fabxc: Does it generally still make sense to have the methods attached to the helper _and_ the normal functions. Seems a bit confusing and I could see people using the plain one if they modify code somewhere.

promql/engine.go
```diff
@@ -606,88 +579,73 @@ func (ev *evaluator) recover(errp *error) {
...
-func (ev *evaluator) evalVector(e Expr) Vector {
-	val := ev.eval(e)
-	vec, ok := val.(Vector)
-	if !ok {
-		ev.errorf("expected instant Vector but got %s", documentedType(val.Type()))
-	}
-	return vec
-}
+// Extra information and caches for evaluating a single node across steps.
+type evalNodeHelper struct {
```
fabxc: Generally seems a bit odd that we have a struct that we pass around all the time. Would it make sense to just directly embed the functionality into the evaluator? Ultimately they are always used together, no?

promql/engine.go
```diff
@@ -697,117 +655,345 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+			return 0, 0, false
+		}
+	}
+	if value.IsStaleNaN(v) {
+		return 0, 0, false
+	}
+	return t, v, true
 }
 
 var pointPool = sync.Pool{}
```
fabxc: We have a size-based pool package now. Probably makes sense to use that.

https://github.com/prometheus/prometheus/tree/master/pkg/pool

Can be another PR of course.

promql/engine.go
```diff
@@ -896,13 +1039,41 @@ func (ev *evaluator) matrixSelector(node *MatrixSelector) Matrix {
 	return matrix
 }
 
-func (ev *evaluator) VectorAnd(lhs, rhs Vector, matching *VectorMatching) Vector {
+func (ev *evaluator) matrixIterSlice(it *storage.BufferedSeriesIterator, maxt, mint int64, out []Point) []Point {
```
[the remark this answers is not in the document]
brian-brazil: This was broken out of matrixSelector (as vectorSelectorSingle was broken out of vectorSelector).

promql/engine.go
brian-brazil: Given how the pools are used I'm not sure that'd actually be a win, it'd take a fair bit of operational experience to determine that.

promql/engine.go
brian-brazil: We'd hopefully catch it with benchmarks. This might be something that can be inlined later on with future optimisations I have in mind.

promql/engine.go
brian-brazil: The helper is tied to a node rather than the evaluator, you'd have to wipe it at each node. I've some future optimisations in mind for recording rules where it would end up as part of the node to be re-used across queries.

promql/engine.go
```diff
@@ -606,88 +579,73 @@ func (ev *evaluator) recover(errp *error) {
...
-	if !convertibleToInt64(sc.V) {
-		ev.errorf("Scalar value %v overflows int64", sc.V)
-	}
-	return int64(sc.V)
-}
+	// Caches.
+	dmn                          map[uint64]labels.Labels      // dropMetricName and label_*.
+	sigf                         map[uint64]uint64             // signatureFunc.
+	signatureToMetricWithBuckets map[uint64]*metricWithBuckets // funcHistogramQuantile.
+	regex                        *regexp.Regexp                // label_replace.
```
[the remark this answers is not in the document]
brian-brazil: I'm not sure it makes sense to document that here, the functions they're used in are mentioned so you can see what they're for.

promql/engine.go
```diff
@@ -697,117 +655,345 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+		mat := ev.eval(e.Expr).(Matrix)
+		if e.Op == itemSUB {
+			for i := range mat {
+				mat[i].Metric = dropMetricName(mat[i].Metric)
+				for j := range mat[i].Points {
+					mat[i].Points[j].V = -mat[i].Points[j].V
+				}
+			}
+		}
+		return mat
```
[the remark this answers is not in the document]
brian-brazil: I've added comments.

promql/engine.go
```diff
@@ -697,117 +655,355 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+						// output labels is dropping the metric name so just do
+						// it once here.
+						Metric: dropMetricName(sel.series[i].Labels()),
+						Points: getPointSlice(numSteps),
+					}
+					inMatrix[0].Metric = sel.series[i].Labels()
+					for ts, step := ev.Timestamp, -1; ts <= ev.EndTimestamp; ts += ev.Interval {
+						step++
+						// Set the non-matrix arguments.
+						// They are scalar, so it is safe to use the step number
```
free: If they're all scalar, why bother using the step number at all? They are all going to be equal anyway. You might as well set the value only once, outside the for loops (line 792 above).

promql/engine.go
```diff
@@ -697,117 +655,355 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+						// when looking up the argument, as there will be no gaps.
+						for j := range e.Args {
+							if j != matrixArgIndex {
+								otherArgsIn[j][0].V = otherArgs[j][0].Points[step].V
+							}
+						}
+						maxt := ts - offset
+						mint := maxt - durationMilliseconds(sel.Range)
+						// Evaluate the matrix selector for this series for this step.
+						points = ev.matrixIterSlice(it, maxt, mint, points[:0])
```
free: Would be worth checking if there is any overlap between previous and desired points (as would be the case with e.g. `rate[5m]` with an interval of `1m`) copying over the overlap and requesting only the difference.

Might even integrate that with the iterator and only `Seek` once.

Probably better done in a separate PR though, this one is already huge.

promql/engine.go
```diff
@@ -697,117 +655,355 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+						if len(points) == 0 {
+							continue
+						}
+						inMatrix[0].Points = points
+						enh.ts = ts
+						// Make the function call.
+						outVec := e.Func.Call(inArgs, e.Args, enh)
+						enh.out = outVec[:0]
```
free: Same as above, would be more readable as

    enh.out = enh.out[:0]

and done before the function call.

promql/engine.go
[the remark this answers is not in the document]
free: I don't mean to pile up, but I have to agree this code is really hard to follow. Needed one hour just to wrap my head around it (having seen the code it replaces before).

Splitting it up into smaller functions would be really helpful. Some variable names are not ideal either (e.g. `inArgs`, `otherArgs`, `otherArgsIn`).

promql/engine.go
```diff
@@ -697,117 +655,355 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+				// Reuse objects across steps to save memory allocations.
+				points := getPointSlice(16)
+				inMatrix := make(Matrix, 1)
+				inArgs[matrixArgIndex] = inMatrix
+				enh := &evalNodeHelper{out: make(Vector, 0, 1)}
+				// Process all the calls for one time series at a time.
+				var it *storage.BufferedSeriesIterator
+				for i, s := range sel.series {
+					if it == nil {
+						it = storage.NewBuffer(s.Iterator(), durationMilliseconds(sel.Range))
```
free: At this point it looks like it may be useful to have a

    storage.NewBuffer(durationMilliseconds(sel.Range))

factory function that initializes the underlying iterator with a dummy. Then you wouldn't need to repeat this if-else all over the place, you could just call `Reset()` directly.

promql/functions.go
```diff
@@ -32,29 +33,42 @@ type Function struct {
...
+	//    Vector, scalars a Vector with one series whose value is the scalar
+	//    value,and nil for strings.
+	// args are the original arguments to the function, where you can access
+	//    matrixSelectors, vectorSelectors, and StringLiterals.
+	// ts is the evaluation timestamp.
+	// enh.out is a pre-allocated empty vector that you may use to accumulate
+	//    output before returning it. The vectors in vals should not be returned.a
+	// Range vector functions need only return a vector with the right value,
+	//     the metric and timestamp are not neded.
+	// Range vector functions need only return a vector with the right value and
```
free: There seems to be some duplication here. Or incomplete copy-pasta.

`s/Range vector/Vector/` ?

promql/engine.go
free: "Hope is not a strategy." *-- Google SRE*

promql/engine.go
```diff
@@ -697,117 +655,355 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+			result := f(args, enh)
+			enh.out = result[:0] // Reuse result vector.
+			// If this could be an instant query, shortcut so as not to change sort order.
+			if ev.EndTimestamp == ev.Timestamp {
+				mat := make(Matrix, len(result))
+				for i, s := range result {
+					s.Point.T = ts
+					mat[i] = Series{Metric: s.Metric, Points: []Point{s.Point}}
+				}
+				return mat
```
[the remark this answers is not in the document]
brian-brazil: That's a good point. They're only going to have length 1 though, so aren't the best to reuse.

promql/engine.go
```diff
@@ -697,117 +655,355 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+							break
+						}
+					}
+				}
+				args[i] = vectors[i]
+			}
+			// Make the function call.
+			enh.ts = ts
+			result := f(args, enh)
+			enh.out = result[:0] // Reuse result vector.
```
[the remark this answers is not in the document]
brian-brazil: The output vector might not be the same as the input vector, and the output is (likely) big enough.

promql/engine.go
brian-brazil: That's presuming there's only one iterator at a time. This is something that needs to be figured out generally, as I want to do this in several other places to reduce allocations.

promql/engine.go
free: Oops, my bad. I had briefly considered the possibility, but dismissed it before I actually thought it through, on the grounds that you mention.

promql/engine.go
free: Well, I have no idea how Go handles memory allocations, but I would guess it's more efficient to `copy()` the pointers to overlapping `Point` instances than to create new ones.

Particularly with very long ranges. I always kick myself when I change the range in a query but not the resolution and everything just hangs for a while.

brian-brazil:
Thanks for the nitpick, I've addressed most of your comments.

promql/engine.go
brian-brazil: It's a bit annoying code wise, but those are the documented semantics. I don't think it's something we can realistically remove, as someone might want to do `log(2)` as a value or something which is reasonable.

promql/engine.go
free: As the code is now you are actually iterating over each series in turn, so you do have one iterator at a time.

But I don't think it makes any difference whether you're scanning one series or more, as long as you're using the same range for all. Am I missing something?

It's just a way of allocating the buffer ahead of time (which will be unnecessary when you have zero series) so you don't always have to check for `nil`.

promql/engine.go
brian-brazil: Most of this PR is about reducing memory allocations. Given how things work you want to copy Points structs around, otherwise all those Point objects would need to be GCed.

> I always kick myself when I change the range in a query but not the resolution and everything just hangs for a while.

That'll be a lot better after this PR :)

free: I realize now I was thinking in Java (it's a difficult context switch to Go) and was assuming these were actually `Point` pointer arrays rather than `Point` struct arrays.

Even with struct arrays, it's still likely that a `copy()` will be more efficient than repopulating the array one by one, even from memory, but one could argue it is not worth the added complexity.

promql/engine.go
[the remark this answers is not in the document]
brian-brazil: If they're different either it doesn't matter as this function is one of the few not taking advantage of it, or it's going to be bigger due to reallocation in which case we want to use the bigger vector we have again next time.

promql/engine.go
```diff
@@ -368,13 +380,15 @@ func (ng *Engine) execEvalStmt(ctx context.Context, query *query, s *EvalStmt) (
...
+	// Instant evaluation. This is executed as a range evaluation with one step.
 	if s.Start == s.End && s.Interval == 0 {
 		start := timeMilliseconds(s.Start)
 		evaluator := &evaluator{
-			Timestamp: start,
-			ctx:       ctx,
-			logger:    ng.logger,
+			Timestamp:    start,
+			EndTimestamp: start,
+			Interval:     1,
```
juliusv: What's the purpose of exporting (capitalizing) all these fields of an unexported struct type?

promql/engine.go
```diff
@@ -98,6 +101,8 @@ type query struct {
 	stmt Statement
 	// Timer stats for the query execution.
 	stats *stats.TimerGroup
+	// Result matrix for reuse.
+	mat Matrix
```
juliusv: As part of a struct, I would tend to spell this out.

promql/engine.go
```diff
@@ -383,84 +397,55 @@ func (ng *Engine) execEvalStmt(ctx context.Context, query *query, s *EvalStmt) (
...
+			return mat, nil
 		default:
-			panic(fmt.Errorf("promql.Engine.exec: invalid expression type %q", val.Type()))
+			panic(fmt.Errorf("promql.Engine.exec: unexpected expression type %q", s.Expr.Type()))
 		}
+
+	}
+	// Range evaluation.
+	if err := contextDone(ctx, "range evaluation"); err != nil {
+		return nil, err
```
juliusv: Not sure if it matters much to the usual caller, but should we stop any started timers before returning? A timer doesn't actually "run", so it's not a goroutine leak, but it won't have any duration recorded to it either before calling `Stop()`.

It seems we already didn't do this in every case before (although I'm pretty sure initially before many changes to the engine, we did), but should we? Or do we just say that the timers aren't useful after a query has errored / timeouted?

promql/engine.go
```diff
@@ -615,88 +587,73 @@ func (ev *evaluator) recover(errp *error) {
...
-// evalInt attempts to evaluate e into an integer and errors otherwise.
-func (ev *evaluator) evalInt(e Expr) int64 {
-	sc := ev.evalScalar(e)
-	if !convertibleToInt64(sc.V) {
-		ev.errorf("Scalar value %v overflows int64", sc.V)
-	}
-	return int64(sc.V)
-}
+	// Caches.
+	dmn                          map[uint64]labels.Labels      // dropMetricName and label_*.
```
juliusv: I'd at least spell this one out.

promql/engine.go
```diff
@@ -706,117 +663,354 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+				matrixes[i] = ev.eval(e).(Matrix)
+
+				// Keep a copy of the original point slices so that they
+				// can be returned to the pool.
+				origMatrixes[i] = make(Matrix, len(matrixes[i]))
+				copy(origMatrixes[i], matrixes[i])
+			}
+		}
+
+		Seriess := map[uint64]Series{}        // Output series by series hash.
```
juliusv: Never capitalize local variables?

promql/engine.go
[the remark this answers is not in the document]
juliusv: I agree, this really needs to be split up or rearranged in some way to make this part more accessible. If we add huge understanding and contribution barriers to our code and basically prevent anyone from ever touching it again, I think that's not even worth a 5x rate speedup. The PromQL engine is crazy complex as it is...

promql/engine.go
brian-brazil: Fixed

promql/engine.go
brian-brazil: Fixed

promql/engine.go
brian-brazil: Seems pretty harmless to me currently

promql/engine.go
brian-brazil: It's doing more than one thing now, so I'm not sure what a good name would be. Do you have a suggestion?

promql/engine.go
brian-brazil: Fixed

promql/engine.go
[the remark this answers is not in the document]
brian-brazil: The previous code wasn't the easiest to understand either, the logic was spread across 50+ functions making it challenging to work with. This concentrates the core evaluation logic into one function, so the complexity is now more obvious.

Do ye have specific suggestions?

promql/engine.go
```diff
@@ -383,84 +397,55 @@ func (ng *Engine) execEvalStmt(ctx context.Context, query *query, s *EvalStmt) (
...
+		logger:       ng.logger,
+	}
+	val, err := evaluator.Eval(s.Expr)
+	if err != nil {
 		return nil, err
 	}
+	evalTimer.Stop()
+	ng.metrics.queryInnerEval.Observe(evalTimer.ElapsedTime().Seconds())
 
-	appendTimer := query.stats.GetTimer(stats.ResultAppendTime).Start()
```
henridf: is `ResultAppendTime`  still needed in package stats, and `queryResultAppend` still needed at the top of this file?

brian-brazil: Right you are, removed.

henridf: I think you can also remove `ResultAppendTime` from `util/stats/query_stats.go`

brian-brazil: Done

promql/engine.go
```diff
@@ -383,84 +388,55 @@ func (ng *Engine) execEvalStmt(ctx context.Context, query *query, s *EvalStmt) (
...
+		case ValueTypeMatrix:
+			return mat, nil
 		default:
-			panic(fmt.Errorf("promql.Engine.exec: invalid expression type %q", val.Type()))
+			panic(fmt.Errorf("promql.Engine.exec: unexpected expression type %q", s.Expr.Type()))
 		}
+
+	}
+	// Range evaluation.
+	if err := contextDone(ctx, "range evaluation"); err != nil {
```
simonpasquier: is it worth checking for the context here since it will be checked in `evaluator.Eval` right after?

promql/engine.go
```diff
@@ -580,25 +548,20 @@ func expandSeriesSet(it storage.SeriesSet) (res []storage.Series, err error) {
...
 	logger log.Logger
 }
 
-func (ev *evaluator) close() {
-	for _, f := range ev.finalizers {
-		f()
-	}
-}
-
 // fatalf causes a panic with the input formatted into an error.
```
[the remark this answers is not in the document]
brian-brazil: Fixed.

They don't fatal the entire program, so error is less confusing.

promql/engine.go
brian-brazil: Done.

There's more work to be done on checking for a timeout, but I left it out of this PR.

promql/bench_test.go
```diff
@@ -13,36 +13,183 @@
...
+			tmp = append(tmp, c)
+		} else {
+			tmp = append(tmp, benchCase{expr: c.expr, steps: 1})
+			tmp = append(tmp, benchCase{expr: c.expr, steps: 10})
+			tmp = append(tmp, benchCase{expr: c.expr, steps: 100})
+			tmp = append(tmp, benchCase{expr: c.expr, steps: 1000})
+		}
+	}
+	cases = tmp
+	for _, c := range cases {
```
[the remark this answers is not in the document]
brian-brazil: This is using subbenchmarks, so nothing in the main function body is timed.

promql/engine.go
```diff
@@ -706,117 +654,354 @@ func (ev *evaluator) eval(expr Expr) Value {
 	if err := contextDone(ev.ctx, "expression evaluation"); err != nil {
 		ev.error(err)
 	}
+	numSteps := int((ev.endTimestamp-ev.timestamp)/ev.interval) + 1
+
+	// rangeWrapper evaluates the given expressions as matrixes, and then for
+	// each step calls the given function with the expressions for that step.
+	// The return value is the combination of all the function call results.
+	rangeWrapper := func(f func([]Value, *evalNodeHelper) Vector, exprs ...Expr) Matrix {
```
free: I find the documentation somewhat difficult to parse. I would:

1. drop the "as matrixes" (it's obvious from the return type);
2. use a less generic name than `f` for the function (e.g. `instantEval`) so you don't have to refer to it as "the provided function";
3. replace " the expressions for that step" with "the values computed for each expression at that step";
4. be more specific about how the results are combined (e.g. "is the combination into time series of ...".

promql/engine.go
```diff
@@ -706,117 +654,354 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+			if ok {
+				return rangeWrapper(func(v []Value, enh *evalNodeHelper) Vector {
+					return e.Func.Call([]Value{ev.vectorSelector(vs, enh.ts)}, e.Args, enh)
+				})
+			}
+		}
+		// Check if the function has a matrix argument.
+		for matrixArgIndex, a := range e.Args {
+			sel, ok := a.(*MatrixSelector)
+			if ok {
```
free: This seems to go against Go style (return early to reduce nesting). The `else` case is only a few lines long, so it may make sense to handle that first. Then you won't need the 2 extra levels of nesting for the big block of code below.

promql/engine.go
free: Like I said, I don't see how this has anything to do with one vs. more iterators. (Although I might be missing something.)

All I was trying to say was that it would be helpful to have a [zero value](https://golang.org/ref/spec#The_zero_value) for `storage.BufferedSeriesIterator`, so you don't always have to check for `nil` (which happens repeatedly throughout the code).

promql/engine.go
```diff
@@ -526,10 +502,6 @@ func (ng *Engine) populateIterators(ctx context.Context, q storage.Queryable, s
 				level.Error(ng.logger).Log("msg", "error expanding series set", "err", err)
 				return false
 			}
-			for _, s := range n.series {
-				it := storage.NewBuffer(s.Iterator(), durationMilliseconds(LookbackDelta))
-				n.iterators = append(n.iterators, it)
```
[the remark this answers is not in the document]
brian-brazil: Good point

promql/engine.go
```diff
@@ -706,117 +654,354 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+				matrixes[i] = ev.eval(e).(Matrix)
+
+				// Keep a copy of the original point slices so that they
+				// can be returned to the pool.
+				origMatrixes[i] = make(Matrix, len(matrixes[i]))
+				copy(origMatrixes[i], matrixes[i])
+			}
+		}
+
+		seriess := map[uint64]Series{}        // Output series by series hash.
```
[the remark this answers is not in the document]
brian-brazil: I'd actually been holding off on minor optimisations like this as they're only once per node and the PR is complex enough as-is. Done now.

promql/engine.go
brian-brazil: It's a bit trickier than that due to the for. I've reworked it.

promql/engine.go
brian-brazil: Changed, though I think `f` is fine as a function name argument.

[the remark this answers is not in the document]
brian-brazil: I'm not sure what you're getting at around the separate struct, but I've moved it to a separate function.

promql/engine.go
brian-brazil: I'll keep it in mind for future work (which will be doing more of this), but I don't want to expand the scope of this PR further.

promql/engine.go
free: The reason I was suggesting a name change was purely for documentation purposes (i.e. the same reason why you don't necessarily call most integer arguments `i`).

But I have no strong feelings about it, code already looks a lot more readable. Thank you for doing the work, BTW.

brian-brazil: Thanks for your comments, they have been useful.

promql/engine.go
```diff
@@ -706,117 +746,270 @@ func (ev *evaluator) eval(expr Expr) Value {
...
-				T: ev.Timestamp,
+		inArgs := make([]Value, len(e.Args))
+		// Evaluate any non-matrix arguments.
+		otherArgs := make([]Matrix, len(e.Args))
+		otherArgsIn := make([]Vector, len(e.Args))
+		for i, e := range e.Args {
+			if i != matrixArgIndex {
+				otherArgs[i] = ev.eval(e).(Matrix)
+				otherArgsIn[i] = Vector{Sample{}}
+				inArgs[i] = otherArgsIn[i]
```
free: I would re-evaluate the naming here. It's `inArgs` in some cases and `argsIn` in others.

promql/engine.go
```diff
@@ -706,117 +746,270 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+				// Set the non-matrix arguments.
+				// They are scalar, so it is safe to use the step number
+				// when looking up the argument, as there will be no gaps.
+				for j := range e.Args {
+					if j != matrixArgIndex {
+						otherArgsIn[j][0].V = otherArgs[j][0].Points[step].V
+					}
+				}
+				maxt := ts - offset
+				mint := maxt - durationMilliseconds(sel.Range)
```
free: You are precomputing `offset` (as `durationMilliseconds(sel.Offset)`) and using it above but then using `durationMilliseconds(sel.Range)` directly here.

promql/functions.go
```diff
@@ -32,29 +33,42 @@ type Function struct {
...
 	ReturnType ValueType
-	Call       func(ev *evaluator, args Expressions) Value
+
+	// vals is a list of the evaluated arguments for the function call.
+	//    For range vectors it will be a Matrix with one series, instant vectors a
+	//    Vector, scalars a Vector with one series whose value is the scalar
+	//    value,and nil for strings.
+	// args are the original arguments to the function, where you can access
+	//    matrixSelectors, vectorSelectors, and StringLiterals.
+	// ts is the evaluation timestamp.
```
free: There is no `ts` argument. Should this be `enh.ts` instead?

promql/engine.go
```diff
@@ -615,88 +575,168 @@ func (ev *evaluator) recover(errp *error) {
...
+	for ts := ev.timestamp; ts <= ev.endTimestamp; ts += ev.interval {
+		// Gather input vectors for this timestamp.
+		for i := range exprs {
+			vectors[i] = vectors[i][:0]
+			for si, series := range matrixes[i] {
+				for pos, point := range series.Points {
+					if point.T == ts {
+						vectors[i] = append(vectors[i], Sample{Metric: series.Metric, Point: point})
+					}
+					if point.T >= ts {
```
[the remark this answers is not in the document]
brian-brazil: We can simplify a little here, but still need the 2nd if to deal with a time series that only appears half way through efficiently.

You can't have just `series.Points` on the LHS, as that's only a local variable so doesn't actually reduce the size of the slice for the next round.

promql/engine.go
```diff
@@ -615,88 +575,168 @@ func (ev *evaluator) recover(errp *error) {
...
+			if !ok {
+				ss = Series{
+					Metric: sample.Metric,
+					Points: getPointSlice(numSteps),
+				}
+				seriess[h] = ss
+			}
+			sample.Point.T = ts
+			ss.Points = append(ss.Points, sample.Point)
+			seriess[h] = ss
```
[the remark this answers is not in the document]
brian-brazil: Removed

promql/functions.go
brian-brazil: That's actually from a previous iteration of the PR, before ts was moved to enh.

promql/engine.go
brian-brazil: Makes sense.

promql/engine.go
brian-brazil: Done, though a little messy as range is a keyword.

promql/engine.go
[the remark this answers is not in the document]
brian-brazil: Actually it is required, but the one above can be removed.

promql/engine.go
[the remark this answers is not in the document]
free: > We can simplify a little here, but still need the 2nd if to deal with a time series that only appears half way through efficiently.

Good catch, but that's only because there's a `for` loop. Now that I look at it, it seems to me you could drop the for loop altogether: the point at timestamp `ts` is either there or it's missing, in which case `point.T` is already past `ts`.

> You can't have just `series.Points` on the LHS, as that's only a local variable so doesn't actually reduce the size of the slice for the next round.

I'm still struggling with' Go's value vs. reference thing (I think they should have kept the `*` for clarity), so still only wondering: is `series` a reference to or a copy of the series in the matrix? Because if it's the former using `series.Points` as the LHS should work just fine. Else I really need to re-read the Go documentation.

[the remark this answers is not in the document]
free: Figured out the value vs. reference and you are of course right. Apologies for the confusion.

The `for` loop though, still appears to be unnecessary.

promql/engine.go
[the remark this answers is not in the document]
free: Might be useful to move the comments onto separate rows preceding the field (possibly prefixed with "Used by") so that it works as documentation (e.g. to be shown as a tooltip by an IDE).

promql/engine.go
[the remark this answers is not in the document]
brian-brazil: We can remove the 2nd if and unconditionally break, the for handles the case then where the series disappears half way through.

promql/engine.go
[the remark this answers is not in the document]
brian-brazil: Done

promql/engine.go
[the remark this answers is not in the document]
free: I see, you would need an `if len(series.Points) > 0` there instead of the `for`. I guess it's arguable whether that would be easier to read and/or more efficient so I guess I don't have a strong opinion.

promql/functions.go
```diff
@@ -32,29 +33,41 @@ type Function struct {
...
+	// args are the original arguments to the function, where you can access
+	//    matrixSelectors, vectorSelectors, and StringLiterals.
+	// enh.out is a pre-allocated empty vector that you may use to accumulate
+	//    output before returning it. The vectors in vals should not be returned.a
+	// Range vector functions need only return a vector with the right value,
+	//     the metric and timestamp are not neded.
+	// Instant vector functions need only return a vector with the right values and
+	//     metrics, the timestamp are not needed.
+	// Scalar results should be returned as the value of a sample in a Vector.
+	Call func(vals []Value, args Expressions, enh *evalNodeHelper) Vector
```
fabxc: Via `evalNodeHelper` we now have an exported type referencing an unexported type, i.e. it cannot actually be used outside of the package.

promql/engine.go
```diff
@@ -905,13 +1057,42 @@ func (ev *evaluator) matrixSelector(node *MatrixSelector) Matrix {
...
+func (ev *evaluator) VectorAnd(lhs, rhs Vector, matching *VectorMatching, enh *EvalNodeHelper) Vector {
 	if matching.Card != CardManyToMany {
 		panic("set operations must only use many-to-many matching")
 	}
-	sigf := signatureFunc(matching.On, matching.MatchingLabels...)
+	sigf := enh.signatureFunc(matching.On, matching.MatchingLabels...)
 
-	var result Vector
 	// The set of signatures for the right-hand side Vector.
 	rightSigs := map[uint64]struct{}{}
```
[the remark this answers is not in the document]
brian-brazil: The performance improvement was insignificant iirc, clearing maps isn't cheap.

promql/engine.go
```diff
@@ -580,31 +545,26 @@ func expandSeriesSet(it storage.SeriesSet) (res []storage.Series, err error) {
...
-// engine through which it connects to a querier and reports errors. On timeout or
-// cancellation of its context it terminates.
+// An evaluator evaluates given expressions over given fixed timestamps. It
+// is attached to an engine through which it connects to a querier and reports
+// errors. On timeout or cancellation of its context it terminates.
 type evaluator struct {
 	ctx context.Context
 
-	Timestamp int64 // time in milliseconds
+	timestamp int64 // Start time in milliseconds.
```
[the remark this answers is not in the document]
brian-brazil: Renamed

promql/engine.go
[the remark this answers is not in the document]
brian-brazil: If memory serves it was only a 0.5% performance improvement, so I didn't bother with keeping it in the PR for the sake of simplicity.

brian-brazil:
> It seems like lots of abstractions get weakened and edge case handling is spread across different places.

I see is as the evaluation logic being centralised (it was previously in 50+ functions, it's now in ~5). The function interface is now a lot tighter.

> I don't feel like I can follow an execution path and thereby understand why the correct result was computed.

The old code wasn't exactly the easiest to grok either, it took me a good while to understand exactly what was going on as the control flow bounces around quite a bit. I suspect this is just an artifact of any sufficiently complex evaluator.

fabxc:
In the absence of more input, I'd suggest we merge this. Yes, things are getting more complex, which is to some degree a necessity when starting to optimize heavily.
OTOH, I think it's becoming extremely unlikely that PromQL will experience any major changes at this point. If so, it'd probably rather be a new QL from scratch.
Small changes, like adding functions and such, should not be harder to make than before really.

So given the benefits, I'd propose we merge this within the next few days.

brian-brazil:
Yeah, major changes seem unlikely. There's a few things we can do now with this design that are long awaited (e.g. memory limits for queries), but they're relatively minor changes.

fabxc:
That's a very fair point, even a rough estimate of query memory consumption will be extremely beneficial. It's still trivial to knock out any Prometheus server with more than a couple of series.

PR #3967, opened 2018-03-14
Update ISSUE_TEMPLATE.md - add FAQ link to the wiki

krasi-georgiev:
these FAQ will be updated frequently and don't quite fit in the official FAQ page on the website.
I am thinking to keep these to a bare minimum - 20-30 at max and delete questions that haven't been asked in the last 2-3 months.

I would like to try and see if it will help reduce the issues and if it becomes useless will just remove the link.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,6 +1,8 @@
 <!--
 
     Please do *NOT* ask usage questions in Github issues.
+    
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
@brian-brazil do you want to give it a try to see if it will help? 
Since you are most active in answering issues I will close if you don't think it will do any good.

brian-brazil:
I looked through the last 2 weeks of issues, and none are covered by the FAQ. It would seem that user support requests are so varied that maintaining such a list isn't very practical.

krasi-georgiev:
ok

PR #3977, opened 2018-03-17
Parser test cleanup

wfernandes:
- Test against the exported package functions instead of the private functions./
   This helped to remove some redundant error checking.
- Moves package function closer to parser function for easier readability.

I did notice that `parseSeriesDesc` is unexported which makes sense since it is only used as part of the tests.


brian-brazil:
Thanks!

PR #3978, opened 2018-03-18
use the Makefile.common

### LINKED ISSUE #3904: Reuse the same Makefile across projects
Most projects will benefit from reusing  a big part of the current prometheus Makefile.
Many are already sharing 90% , but slowly starting to drift.

The idea is to keep the prometheus Makefile as the main source so any PR would be opened against that and all other projects will simply copy the prometheus Makefile. 

Any differences should be able to  tackle using the makefile include directive to include any custom targets and env variable for each repo.

krasi-georgiev:
fixes #3904 
depends on https://github.com/prometheus/common/pull/121

cc @brian-brazil


brian-brazil:
This is a good start. I think that the common Makefile should live in either the promu or prometheus repos, not in common which is library code. Downloading it on the fly will be fragile, I'd prefer that it get updated via tooling+PRs.

krasi-georgiev:
the idea is to keep a local cached copy like I did in the prometheus PR , the `-f` flag means if the pull fails don't override the existing file and just run the cached copy.

What is your idea to keep it up to date?

brian-brazil:
We send a batch of PRs every time it changes. That way we know that each repo is always buildable.

krasi-georgiev:
a PR on every change of Makefile.common? 
wasn't this how it was few years ago and didn't work well?
https://docs.google.com/document/d/1Ql-f_aThl-2eB5v3QdKV_zgBdetLLbdxxChpy-TnWSE

what about that -f flag do you see any problem with that?

brian-brazil:
> wasn't this how it was few years ago and didn't work well?

Makefiles never went away, and each Makefile was still altered on an as-needed basis and things still drifted.

> what about that -f flag do you see any problem with that?

Changing behaviour depending on whether we get a HTTP response is fragile.

krasi-georgiev:
so your idea is to keep Makefile.common in lets say prometheus/prometheus and each time it needs a change to open a PR against each other prometheus repo ?

> Changing behaviour depending on whether we get a HTTP response is fragile.
don't quite understand what you mean here. 
my suggestion is to keep a local cached copy of the Makefile.common at all times.

brian-brazil:
Yes.

> my suggestion is to keep a local cached copy of the Makefile.common at all times.

Only using it sometimes is fragile and will lead to confusion.

krasi-georgiev:
maybe adding `make update` to explicitly update the cached version.
but anyway what kind of confusion do you think it might lead to?

one problem I am trying to avoid is merging PR's by mistakes on repos different than the main source of the Makefile.common  which might happen quite often if Makefile.common gets addopted by many repos.

brian-brazil:
> but anyway what kind of confusion do you think it might lead to?

People on flaky internet connections get different behaviour, and there's now N potential makefiles that could be in use depending on when it last worked.

brian-brazil:
:+1:

krasi-georgiev:
@brian-brazil thanks
@SuperQ do you see any problems with this or I can merge?

 next I will open issues for the other golang repos to start using the common targets.

Makefile.common
```diff
@@ -0,0 +1,99 @@
...
+	@echo ">> fetching promu"
+	@GOOS= GOARCH= $(GO) get -u github.com/prometheus/promu
+
+$(FIRST_GOPATH)/bin/staticcheck:
+	@GOOS= GOARCH= $(GO) get -u honnef.co/go/tools/cmd/staticcheck
+
+$(FIRST_GOPATH)/bin/govendor:
+	@GOOS= GOARCH= $(GO) get -u github.com/kardianos/govendor
+
+.PHONY: all style check_license format build test vet assets tarball docker promu staticcheck $(FIRST_GOPATH)/bin/staticcheck govendor $(FIRST_GOPATH)/bin/govendor
```
grobie: End the file with a newline.

Makefile.common
```diff
@@ -0,0 +1,99 @@
...
+	@echo ">> building release tarball"
+	@$(PROMU) tarball --prefix $(PREFIX) $(BIN_DIR)
+
+docker:
+	@echo ">> building docker image"
+	@docker build -t "$(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)" .
+
+promu:
+	@echo ">> fetching promu"
+	@GOOS= GOARCH= $(GO) get -u github.com/prometheus/promu
```
grobie: Not showing what is being executed and especially downloaded in a Makefile is something I consider an anti-pattern and has led to several questions and support overhead when these downloads hang/fail.

I don't understand why a line `>> fetching promu` is preferred over showing the user what we install directly (e.g. `go get -u github.com/prometheus/promu`). Not show the commands makes debugging very hard.

brian-brazil:
> I dislike making a fast internet connection a development requirement (by redownloading promu on every single make build execution).

I'm not a fan either, but it's what the existing Makefile does so I don't think this is the PR to propose changing it.

krasi-georgiev:
I agree that we should show the commands that are running so updated and will merge on green.

the promu fetch will be left for another PR.

grobie:
> it's what the existing Makefile does so I don't think this is the PR to propose changing it.

I understood this PR to introduce the common Makefile which will be introduced in all projects, as it aims to resolve #3904. If this is not the final PR for that, feel free to merge.

krasi-georgiev:
@grobie yes it is
any suggestions or a link to another makefile that handles the promu fetch better?

krasi-georgiev:
anyway it doesn't have to be the final PR , once we start adding to other repos will adjust as needed.

krasi-georgiev:
as of golang 1.9 the vendored dir is excluded so I updated it to just use `./...` with any command.

brian-brazil:
Though on the other hand client_golang doesn't have any vendoring.

krasi-georgiev:
I thougght of that , but client_golang doesn't  have a `Makefile` , do we even need one there?

brian-brazil:
Most of it won't be useful, but the things like tests apply.

krasi-georgiev:
what it is the promise on the golang version for the client_golang?
At the moment it doesn't have a makefile and if you think that by the time we would need a makefile the minimum golang version would be 1.9 , than best to leave the common makefile as is.

krasi-georgiev:
ping @brian-brazil

Makefile.common
```diff
@@ -0,0 +1,97 @@
...
+all: format staticcheck unused build test
+
+style:
+	@echo ">> checking code style"
+	! $(GOFMT) -d $$(find . -path ./vendor -prune -o -name '*.go' -print) | grep '^'
+
+check_license:
+	@echo ">> checking license header"
+	mkdir -p scripts
+	curl https://raw.githubusercontent.com/prometheus/common/master/scripts/check_license.sh -s -f -o scripts/check_license.sh || true
```
brian-brazil: This doesn't work, and avoid downloading scripts on the fly.

Makefile.common
```diff
@@ -0,0 +1,97 @@
...
+PROMU        := $(FIRST_GOPATH)/bin/promu
+STATICCHECK  := $(FIRST_GOPATH)/bin/staticcheck
+GOVENDOR     := $(FIRST_GOPATH)/bin/govendor
+pkgs          = ./...
+
+PREFIX                  ?= $(shell pwd)
+BIN_DIR                 ?= $(shell pwd)
+DOCKER_IMAGE_TAG        ?= $(subst /,-,$(shell git rev-parse --abbrev-ref HEAD))
+
+all: format staticcheck unused build test
```
grobie: The `all` target should abort incorrectly formatted code, instead of silently formatting it. Replace `format` with `style`.

brian-brazil: I think this should stay as-is as this is what a human will use. Maybe a separate CI target?

This is a matter for another PR in any case, we shouldn't be making changes here just refactoring.

grobie: All other targets abort the command and require the user to fix it, format is the only one behaving differently. Introducing another target is unnecessary and complicates all CI integrations which commonly just execute `make`.

This PR is explicitely labeled to introduce the new common Makefile for all projects, so either that needs to be changed or this is the PR where I have to request such changes.

Makefile.common
krasi-georgiev: tried it for about 1h and can't get it right. This is a lost fight  with quotes and new lines.

sdurrheimer: Maybe the time to use the [`promu check licenses` command](https://github.com/prometheus/promu/pull/84) ?

krasi-georgiev: finally!

SuperQ:
Sorry for being idle on this conversation.

I like the idea of having `make update_common_makefile`.

One idea, we could include this in our cicleci test pipeline.

Makefile.common
krasi-georgiev: @sdurrheimer I am not a big fan of promu for to be honest :) It obscures simple tasks and need to look into a different repo and files to troubleshooting issues.
I understand why it was created, but I am trying to avoid using it where possible - KISS principle.

Makefile.common
krasi-georgiev: updated!

krasi-georgiev:
> One  idea, we could include this in our cicleci test pipeline.

@SuperQ can you explain a bit more as I didn't quite understand the idea

Makefile.common
sdurrheimer: `promu` was meant to simplify Makefiles and avoid duplicated scripts in every prometheus repository.

krasi-georgiev: I red the google doc for the proposal and it is a good idea just had some bad experience the last time I upgraded the golang version and it was a bit obscure why it was failing. I  think there was another project which was also failing and another  maintainer also had some trouble finding the source of the problem. I think it was when we replaced Json Unmarhsal with Unmarhsal Strict.
Promu auto updates on every run which is exactly what my first proposal was for the common Makefile and got rejected with some valid points. 
I looked at the code at the time and I am not seeing any benefits of using promu instead of a Makefile and the golang-builder directly.

I do like the golang-builder, it does a great job and with some changes we made it do the entyre build process directly.

here is a working POC pipeline config to build prometheus on drone CI
https://github.com/sipian/prometheus/blob/0d74ade16889807b4ecc943fec9f61feeb297249/.drone.yml

and here are the changes needed to make this work.
https://github.com/prometheus/golang-builder/compare/master...sipian:drone?expand=1

I am quite sure we can also add support for multi arch images.

krasi-georgiev:
would need to merge this one first to pass the tests.
https://github.com/prometheus/prometheus/pull/4073

SuperQ:
@krasi-georgiev We run a number of tests automatically.  We could add a test failure into the build setup to fail if the common makefile changes.  This might be slightly annoying for unrelated updates, but it helps address @grobie's desire for offline development.  (I agree with the desire for offline development)

krasi-georgiev:
@SuperQ the original PR allowed offline mode, by using a cached version when no internet connection.
The main concern was that a bad  update of the Makefile.common  would propagate in all projects that are using it causing many open PRs to fail.
If I understand your suggestion properly it will lead to the same problem, no ?

Makefile.common
```diff
@@ -0,0 +1,96 @@
...
+PROMU        := $(FIRST_GOPATH)/bin/promu
+STATICCHECK  := $(FIRST_GOPATH)/bin/staticcheck
+GOVENDOR     := $(FIRST_GOPATH)/bin/govendor
+pkgs          = ./...
+
+PREFIX                  ?= $(shell pwd)
+BIN_DIR                 ?= $(shell pwd)
+DOCKER_IMAGE_TAG        ?= $(subst /,-,$(shell git rev-parse --abbrev-ref HEAD))
+
+all: style staticcheck unused build test
```
brian-brazil: This is changed from the original, don't change semantics in a refactoring PR

Makefile.common
```diff
@@ -0,0 +1,96 @@
...
+# Ensure GOBIN is not set during build so that promu is installed to the correct path
+unexport GOBIN
+
+GO           ?= go
+GOFMT        ?= $(GO)fmt
+FIRST_GOPATH := $(firstword $(subst :, ,$(shell $(GO) env GOPATH)))
+PROMU        := $(FIRST_GOPATH)/bin/promu
+STATICCHECK  := $(FIRST_GOPATH)/bin/staticcheck
+GOVENDOR     := $(FIRST_GOPATH)/bin/govendor
+pkgs          = ./...
```
brian-brazil: This is changed from the original

Makefile.common
krasi-georgiev: completely agree in general.
These changes are small enough and I hope you can make an exception.

brian-brazil: All the small changes in this PR are rewriting most of the file. It'd be easier to review if things were kept simpler. At least split it into sane commits.

krasi-georgiev: I think I had them split logically, but I squashed few because of the DCO.

Now rebased with master and all tests look good.

Makefile.common
```diff
@@ -0,0 +1,96 @@
...
+	@echo ">> vetting code"
+	$(GO) vet $(pkgs)
+
+staticcheck: $(STATICCHECK)
+	@echo ">> running staticcheck"
+	$(STATICCHECK) -ignore "$(STATICCHECK_IGNORE)" $(pkgs)
+
+unused: $(GOVENDOR)
+	@echo ">> running check for unused packages"
+	@$(GOVENDOR) list +unused | grep . && exit 1 || echo 'No unused packages'
```
brian-brazil: This has changed

krasi-georgiev:
Reverted all mentioned changes.

brian-brazil:
:+1:

krasi-georgiev:
if no one else has anymore suggestions merging tomorrow.

Makefile.common
```diff
@@ -0,0 +1,96 @@
...
+
+all: format staticcheck unused build test
+
+style:
+	@echo ">> checking code style"
+	! $(GOFMT) -d $$(find . -path ./vendor -prune -o -name '*.go' -print) | grep '^'
+
+check_license:
+	@echo ">> checking license header"
+	for file in $$(find . -type f -iname '*.go' ! -path './vendor/*') ; do \
```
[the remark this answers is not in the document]
krasi-georgiev: yep , thanks , updated.

PR #3979, opened 2018-03-19
un-generalize the pool package

twmb:
PR #3835 generalized the pool package by allowing users to pass a make function that returned an interface. Presumably, this is so that the pool can be used for any type of slice.

The only usage of the pool at the moment is for byte slices, but allowing it to be generic introduces a bunch of type safety issues.

This respecializes the pool to be for byte slices only, which buys higher type safety, makes the usage more obvious, and removes the reflect package.

Lastly, we now put `*[]byte` into the internal sync.Pool's as opposed to `[]byte`, see: https://staticcheck.io/docs/staticcheck#SA6002.

(I'm find with closing & deleting this if the desire to use `pool` for other than byte slices exists).

### DIFF
--- a/pkg/pool/pool.go
+++ b/pkg/pool/pool.go
@@ -14,8 +14,6 @@
 package pool
 
 import (
-	"fmt"
-	"reflect"
 	"sync"
 )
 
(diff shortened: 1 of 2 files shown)


krasi-georgiev:
I added this not long ago and the idea is to be used in the promql as it can also benefit from a buffer pool

IIRC I think here was to be used.
https://github.com/prometheus/prometheus/blob/master/promql/engine.go#L842

krasi-georgiev:
@twmb I think for now we should close this and if in few months it happens that we don't use the pool for anything else can reopen.

krasi-georgiev:
Thanks , that would be much appreciated.

Issue #3545, opened 2017-12-05
Prometheus 2.0 prioritizes EC2 instance role over credentials in environment variables


moshebs:
I was using Prometheus 1.8 with EC2 discovery, and I gave Prometheus the AWS access key using environment variables (AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY), as documented [here](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#<ec2_sd_config>) - everything worked fine.

When I moved to Prometheus 2.0, this stopped working and I found out Prometheus ignores the environment variables and tries to use the EC2 instance role. since I my EC2 instance role did not have the required permissions, it failed. 

Is this the expected behavior? If it is, I think it should be documented in the migration guide.

brian-brazil:
AWS is a mish-mash of I believe 5 auth methods currently. For normal developers they know what they want, whereas we have to use heuristics. We also have the general principle that all config should be in the config file, not coming from elsewhere.

Here I'd suggest putting the secrets directly in the config, as I don't see a way for us to correctly choose between two different ways of automatically acquiring credentials from outside the config.

moshebs:
While I understand why operationally-wise it is simple to have the credentials in the config file, security-wise it is not the best choice and AWS recommend using instance profile. From that perspective I am kind of happy I ran through that because it forces me to use Instance profile and drop the env variables.
You are probably using an AWS SDK, right? so the approach I would choose (maybe this is what you do): take credentials from Prometheus config file, and if it does not exist simply call SDK APIs, and let them take it from there. If this is what you do, maybe AWS SDK changed its behavior?

brian-brazil:
> if it does not exist simply call SDK APIs, and let them take it from there. If this is what you do, maybe AWS SDK changed its behavior?

Two of the auth mechanisms require extra code, when to apply those is the issue here.

fajpunk:
Does this mean that EC2 target discovery is no longer expected to work by specifying the `profile` key in the config?  For example, here's my config that worked with 1.8, is this no longer expected to work?

```- job_name: 'aws_primary'
    ec2_sd_configs:
      - region: 'us-east-2'
        profile: '<my profile name, defined in ~/.aws/credentials of the user running prometheus>'
        refresh_interval: 60s
        port: 9100
```
To be clear, I'm not complaining, I'm just looking for clarification on how ec2 target discovery works.  I'm glad to submit a PR with doc updates once I understand it :)

[...]

PR #3343, opened 2017-10-24
Feature: Allow getting credentials via EC2 role

diurnalist:
This is subtly different than the existing `role_arn` solution, which
allows Prometheus to assume an IAM role given some set of credentials
already in-scope. With EC2 roles, one specifies the role at instance
launch time (via an instance profile.) The instance then exposes
temporary credentials via its metadata. The AWS Go SDK exposes a
credential provider that polls the [instance metadata endpoint][1]
already, so we can simply use that and it will take care of renewing the
credentials when they expire.

Without this, if this is being used inside EC2, it is difficult to
cleanly allow the use of STS credentials. One has to set up a proxy role
that can assume the role you really want, and launch the EC2 instance
with the proxy role. This isn't very clean, and also doesn't seem to be
[supported very well][2].

[1]:
http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html
[2]: https://github.com/aws/aws-cli/issues/1390

PR #3985, opened 2018-03-20
Revert "Feature: Allow getting credentials via EC2 role (#3343)"

jeeyoungk:
This reverts commit 808f79f00aab29cd6464c71b120105765533c4c7.

This ensures that without any custom configuration, EC2 instance profile will be used while running in an EC2 instance / ECS container.

See the discussion https://github.com/prometheus/prometheus/pull/3343 for more info.

This should also resolve https://github.com/prometheus/prometheus/issues/3545


brian-brazil:
Thanks!

PR #3990, opened 2018-03-20
Minor edits to help text

jamtur01:
Both very minor. Tthe first just reads awkwardly and the second alert manager to Alertmanager for correctness and consistency.


brian-brazil:
Thanks!

PR #3992, opened 2018-03-21
all: remove unnecessary type conversions

ferhatelmas:
except promql due to not to create conflict with #3966


brian-brazil:
This is going to cause conflicts with #3966, can you revert the promql changes for now?

ferhatelmas:
@brian-brazil Updated.

brian-brazil:
Thanks!

PR #3995, opened 2018-03-21
Fix moved fsnotify dependency

aaron7:
The `fsnotify` package was moved from https://github.com/go-fsnotify/fsnotify to https://github.com/fsnotify/fsnotify


aaron7:
One second, I will set to same revision

brian-brazil:
Thanks!

PR #3238, opened 2017-10-04
Support basic_auth in marathon_sd_config.  Closes #3090

mciura:

PR #4009, opened 2018-03-25
Use common HTTPClientConfig for marathon_sd configuration

### LINKED ISSUE #3090: basic_auth support in marathon_sd_config
**What did you do?**
Well, trying to configure Marathon monitoring in Prometheus. Marathon does not support built in authentication, so obviously someone has to use basic_auth to protect Marathon/Mesos. We did the same. 

There is already a way to support basic auth for other modules, then why shouldn't we implement that for marathon_sd_config? Copy/Past stuff!! :)

* Prometheus version: 

	prometheus, version 1.7.1 (branch: master, revision: 3afb3fffa3a29c3de865e1172fb740442e9d0133)
  build user:       root@0aa1b7fc430d
  build date:       20170612-11:44:05
  go version:       go1.8.3

plaflamme:
This adds support for basic authentication which closes #3090

The support for specifying the client timeout was removed as discussed in https://github.com/prometheus/common/pull/123. Marathon was the only sd mechanism doing this and configuring the timeout is done through `Context`.

@brian-brazil @fabxc this is another take on adding support for basic auth to marathon SD through the use of the standard `HTTPClientConfig` as mentioned in https://github.com/prometheus/prometheus/pull/3238


discovery/marathon/marathon.go
```diff
@@ -140,37 +135,28 @@ func NewDiscovery(conf SDConfig, logger log.Logger) (*Discovery, error) {
...
-	tls, err := httputil.NewTLSConfig(conf.TLSConfig)
+	client, err := httputil.NewHTTPClientFromConfig(conf.HTTPClientConfig, "marathon_sd")
 	if err != nil {
 		return nil, err
 	}
 
-	token := string(conf.BearerToken)
-	if conf.BearerTokenFile != "" {
-		bf, err := ioutil.ReadFile(conf.BearerTokenFile)
+	// Unfortunately, we have to do this part ourselves because Marathon uses a
```
brian-brazil: I don't think it's a good idea to override the behaviour of the common HTTP client, bearer token should mean bearer token. If there's a bespoke auth method that some SD uses I think it should have different configuration field names to avoid confusion.

plaflamme: Sure, that's reasonable. The [DC/OS documentation](https://docs.mesosphere.com/1.11/security/oss/iam-api/) uses the term "authentication token". So Perhaps `auth_token` / `auth_token_file`?

How should backwards compatibility be handled (because this is currently supported through the standard property names)?

* no backwards compatibility, users will have to rename the configuration properties
* support both the old and new property names and log a warning
* ?

plaflamme:
@brian-brazil I've added a commit to address your concern. Marathon SD now uses custom configuration properties to configure the token-based authentication mechanism.

I also adjusted the original commit which was broken and failing tests.

Please let me know how backwards compatibility should be handled.

discovery/marathon/marathon.go
brian-brazil: > How should backwards compatibility be handled (because this is currently supported through the standard property names)?

Is basic auth supported by Marathon natively? I didn't see it in the docs, so the things to do might be to leave things as-is.

plaflamme: Marathon itself only supports basic authentication as per its [documentation here](https://mesosphere.github.io/marathon/docs/ssl-basic-access-authentication.html). Marathon deployed as part of DC/OS only supports token-based authentication (as you noticed in the other documentation).

So arguably, it is the token-based authentication that should be dropped from `marathon_sd`. I wouldn't suggest doing that, since that would break existing functionality and leave no workaround.

Another complication is that the DC/OS documentation says they will eventually support the `Bearer` authorization header. This would also be supported with this PR, but it also means that we cannot easily be backwards compatible and future compatible with a new DC/OS version.

Perhaps this PR is the best way to go forward?

* supports basic auth for marathon itself
* supports token-based authentication for DC/OS
* supports future `Bearer` authentication for DC/OS

Existing DC/OS users will have to rename their configuration properties from `bearer_token` to `auth_token` when deploying the new version. Not doing so will simply fail authenticating...

brian-brazil: That does seem like the best approach overall, even if it does involve a breaking change.

plaflamme: Sounds good.

Is there anything else required to move this forward? Should this log a warning of some sort when `bearer_token` is set? Should I also make the changes to the documentation?

brian-brazil: I don't see a need for a warning. Docs should be updated in the same PR.

plaflamme: I've added 2 commits:

* one that updates the documentation and updates configuration validation accordingly
* another that updates `CHANGELOG.md` to document the breaking change

I'm not 100% sure what the process is around `CHANGELOG.md`, I can drop that commit if you prefer.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## next release
+
+* [CHANGE] `marathon_sd`: (**breaking**) use `auth_token` and `auth_token_file` for token-based authentication instead of `bearer_token` and `bearer_token_file` respectively.
```
brian-brazil: CHANGES by definition are breaking, so no need to have the breaking.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
...
+* [ENHANCEMENT] `marathon_sd`: adds support for basic authentication.
```
brian-brazil: basic and bearer, plus all the other option the common http client provides

discovery/marathon/marathon.go
```diff
@@ -109,8 +106,17 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if len(c.Servers) == 0 {
 		return fmt.Errorf("Marathon SD config must contain at least one Marathon server")
 	}
-	if len(c.BearerToken) > 0 && len(c.BearerTokenFile) > 0 {
-		return fmt.Errorf("at most one of bearer_token & bearer_token_file must be configured")
+	if len(c.AuthToken) > 0 && len(c.AuthTokenFile) > 0 {
```
brian-brazil: We should have unittests for these, see where we test bad configs already.

discovery/marathon/marathon.go
```diff
@@ -140,37 +146,26 @@ func NewDiscovery(conf SDConfig, logger log.Logger) (*Discovery, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
-	token := string(conf.BearerToken)
-	if conf.BearerTokenFile != "" {
-		bf, err := ioutil.ReadFile(conf.BearerTokenFile)
+	authToken := conf.AuthToken
+	if len(authToken) == 0 && len(conf.AuthTokenFile) > 0 {
+		b, err := ioutil.ReadFile(conf.AuthTokenFile)
```
brian-brazil: This needs to be read on every request.

discovery/marathon/marathon.go
```diff
@@ -288,20 +283,20 @@ type AppList struct {
...
+func fetchApps(client *http.Client, url string, token config_util.Secret) (*AppList, error) {
 	request, err := http.NewRequest("GET", url, nil)
 	if err != nil {
 		return nil, err
 	}
 
-	// According to  https://dcos.io/docs/1.8/administration/id-and-access-mgt/managing-authentication
+	// According to https://docs.mesosphere.com/1.11/security/oss/managing-authentication/
 	// DC/OS wants with "token=" a different Authorization header than implemented in httputil/client.go
 	// so we set this implicitly here
```
brian-brazil: explicitly

docs/configuration/configuration.md
```diff
@@ -786,16 +786,42 @@ See below for the configuration options for Marathon discovery:
...
+[ auth_token: <secret> ]
+
+# Optional authentication information for token-based authentication
+# https://docs.mesosphere.com/1.11/security/ent/iam-api/#passing-an-authentication-token
+# It is mutually exclusive with `auth_token` and other authentication mechanisms.
+[ auth_token_file: <filename> ]
+
+# Sets the `Authorization` header on every request with the
+# configured username and password.
+# This is mutually exclusive from other authentication mechanisms.
```
brian-brazil: s/from/with/

docs/configuration/configuration.md
```diff
@@ -786,16 +786,42 @@ See below for the configuration options for Marathon discovery:
...
+basic_auth:
+  [ username: <string> ]
+  [ password: <string> ]
+
+# Sets the `Authorization` header on every request with
+# the configured bearer token. It is mutually exclusive with `bearer_token_file` and other authentication mechanisms.
+# NOTE: At this time, DC/OS marathon does not support standard Bearer token authentication. Use `auth_token` instead.
```
brian-brazil: Mention a version number so this doesn't go stale

discovery/marathon/marathon.go
plaflamme: For this part, I did the same as what the http client does for the `bearer_token_file` configuration property [here](https://github.com/prometheus/common/blob/93dfbf7/config/http_config.go#L135) which doesn't seem to read the file on every request.

Perhaps I am mis-reading that code somehow? Or perhaps that change is in the works? I can certainly give it a stab if you feel strongly about supporting this now.

brian-brazil: That's not right, https://github.com/prometheus/prometheus/pull/3709/files was meant to do that.

plaflamme: Oh, ok, I looked at the wrong http client. Alrighty, I'll take a stab at implementing this.

brian-brazil: You're looking at the right one, it needs fixing.

plaflamme: I may need a bit of guidance on this one. The approach I'm considering is to create a `AuthTokenRoundTripper` and then use that to wrap the `RoundTripper` that `httputil` creates. But to put that new `rt` on the client, I see 2 approaches:

* create client using `NewClientFromConfig` and then override its `RoundTripper` member
* create `NewHTTPRoundTripperFromConfig` in `httputil` (which is basically `NewHTTPClientFromConfig` minus the call to `newClient`)

Mutating the client seems like a bad idea to me, but it makes the change entirely local to `marathon.go` which perhaps is what we want in this case, I'm not sure.

Perhaps there's an alternative approach that you might suggest?

plaflamme: Ok, I took the second approach in 2a0e760 and 904902e. Please let me know if that's okay.

plaflamme:
The `travis-ci` build failure looks like a flake in the k8s tests. Is there anything that I should do about that?

discovery/marathon/marathon.go
```diff
@@ -146,30 +145,77 @@ func NewDiscovery(conf SDConfig, logger log.Logger) (*Discovery, error) {
...
 		logger:          logger,
 	}, nil
 }
 
+type authTokenRoundTripper struct {
+	authToken config_util.Secret
+	rt        http.RoundTripper
+}
+
+// newAuthTokenRoundTripper adds the provided auth token to a request
```
brian-brazil: Comments should end with a full stop.

brian-brazil:
You've got conflicts

plaflamme:
@brian-brazil resolved

brian-brazil:
Thanks!

PR #4010, opened 2018-03-25
Improve Kubernetes Discovery Performance

cofyc:
Hi, we are running a multi-tenant Kubernetes cluster. Users of kubernetes cluster create Kuberntes service discovery jobs to discovery targets in their namespaces. If there are two many jobs, latency to discover new or updated target is huge. I did some improvements  to mitigate this issue:

- Share kubernete1s informers in kubernetes discovery to improve performance.
  - In previous Kubernetes discovery, each job creates its own client and informers, if there are many jobs, prometheus will keep a lot of connections to Kubernetes apiserver. After sharing kubernetes client and informers to same apiserver, network traffic between prometheus and kube-apiserver goes down tremendously, and also number of go routines. 
- Refactor kubernetes controller
  - Do initial listing and syncing to scrape manager, then register event handlers may lost events happening in listing and syncing (if it lasted a long time). We should register event handlers at the very begining, before processing just wait until informers synced (sync in informer will list all objects and call `OnUpdate` event handler).
  - Use a queue then we don't block event callbacks and an object will be processed only once if added multiple times before it being processed.

What do you think?

### DIFF
--- a/discovery/kubernetes/endpoints.go
+++ b/discovery/kubernetes/endpoints.go
@@ -25,6 +25,7 @@ import (
 	"github.com/prometheus/prometheus/discovery/targetgroup"
 	apiv1 "k8s.io/client-go/pkg/api/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
 )
 
 // Endpoints discovers new endpoint targets.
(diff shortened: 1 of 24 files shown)


cofyc:
cc @brancz @brian-brazil

brian-brazil:
It sounds like you're duplicating the benefits of #2191, and I don't think it's wise to have multiple layers of sharing. I'd suggest limiting your PR to just the other fix.

cofyc:
@brian-brazil Didn't know there was already a tracking issue. I agree it's not good to have multiple layers of sharing.
I submitted a separate PR (https://github.com/prometheus/prometheus/pull/4013) for the second fix.

PR #4011, opened 2018-03-25
discovery/ec2: Support filtering instances in discovery

### LINKED ISSUE #3920: Optional Filter for EC2 Discovery
I would really like to see an optional filter for AWS EC2 Discovery.

Currently, prometheus describes *all* instances, and then filters the results locally. In an AWS account with hundreds or thousands of instances, many of which are not related to prometheus' monitoring configuration, the amount of data requested and returned from the API is unnecessarily huge and time-consuming.

I would like to able to provide a custom filter string to be used with the DescribeInstances API call, so that the API only returns relevant instances for further processing.

Here's a really really bad example, but an example nonetheless of something we have done to patch in the functionality. We should not have done it in this hardcoded, and environment-variable-dependent manner, but it works for now:

```
                        &ec2.Filter{
                                Name:   aws.String("tag:Environment"),
                                Values: []*string{aws.String(os.Getenv("ENVIRONMENT"))},
                        },
```

cnicolov:
Fixes #3920 

Wasn't sure how to do the documentation markup 🗒

### DIFF
--- a/discovery/ec2/ec2.go
+++ b/discovery/ec2/ec2.go
@@ -228,7 +228,6 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 	}
 
 	var filters []*ec2.Filter
-
 	for _, f := range d.filters {
 		filters = append(filters, &ec2.Filter{
 			Name:   aws.String(f.Name),
--- a/discovery/ec2/ec2.go
+++ b/discovery/ec2/ec2.go
@@ -114,6 +114,11 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		}
 		c.Region = region
 	}
+	for _, f := range c.Filters {
+		if len(f.Values) == 0 {
+			return fmt.Errorf("EC2 SD configuration filter values cannot be empty")
+		}
+	}
 	return nil
 }
 
@@ -222,19 +227,16 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 		Source: *d.aws.Region,
 	}
 
-	var input *ec2.DescribeInstancesInput
+	var filters []*ec2.Filter
 
-	if len(d.filters) > 0 {
-		var filters []*ec2.Filter
-		for _, f := range d.filters {
-			filters = append(filters, &ec2.Filter{
-				Name:   aws.String(f.Name),
-				Values: aws.StringSlice(f.Values),
-			})
-		}
-
-		input = &ec2.DescribeInstancesInput{Filters: filters}
+	for _, f := range d.filters {
+		filters = append(filters, &ec2.Filter{
+			Name:   aws.String(f.Name),
+			Values: aws.StringSlice(f.Values),
+		})
 	}
+
+	input := &ec2.DescribeInstancesInput{Filters: filters}
 
 	if err = ec2s.DescribeInstancesPages(input, func(p *ec2.DescribeInstancesOutput, lastPage bool) bool {
 		for _, r := range p.Reservations {
--- a/discovery/ec2/ec2.go
+++ b/discovery/ec2/ec2.go
@@ -70,6 +70,12 @@ var (
 	}
 )
 
+// Filter is the configuration for filtering EC2 instances.
+type Filter struct {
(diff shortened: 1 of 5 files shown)


discovery/ec2/ec2.go
```diff
@@ -212,7 +221,22 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 	tg = &targetgroup.Group{
 		Source: *d.aws.Region,
 	}
-	if err = ec2s.DescribeInstancesPages(nil, func(p *ec2.DescribeInstancesOutput, lastPage bool) bool {
+
+	var input *ec2.DescribeInstancesInput
+
+	if len(d.filters) > 0 {
```
brian-brazil: Do you need to special case this?

discovery/ec2/ec2.go
```diff
@@ -212,7 +221,22 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
...
+		var filters []*ec2.Filter
+		for _, f := range d.filters {
+			filters = append(filters, &ec2.Filter{
+				Name:   aws.String(f.Name),
+				Values: aws.StringSlice(f.Values),
```
brian-brazil: EC2 docs say there must be at least one value, you should ensure that.

docs/configuration/configuration.md
```diff
@@ -429,6 +429,13 @@ region: <string>
 # The port to scrape metrics from. If using the public IP address, this must
 # instead be specified in the relabeling rule.
 [ port: <int> | default = 80 ]
+
+# Filters can be used optionally to filter the instance list by other criteria.
+# Available filter criteria can be found here:
+# https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html
+# Filter API documentation: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Filter.html
+filters:
+  [ - name: <string>, values: [ - <string> ] ... ]
```
brian-brazil: This isn't quite understandable, I'd expand it out a bit.

You should also have a warning here that this is only for performance, and that relabelling is still the main way of doing things. See the one added for Consul.

docs/configuration/configuration.md
```diff
@@ -429,7 +429,28 @@ region: <string>
...
+# For example if we need to filter instances with tags environment=production and team=backend,
+# we can use the following filters:
+# filters:
+# - name: tag:environment
+#   values:
+#   - production
+# - name: tag:team
+#   values:
+#   - backend
+filters:
```
brian-brazil: I meant to expand this out, not to put an example in the comments.

cnicolov: 👍 But still not sure how to expand it out, could you please give me a hint?

brian-brazil: Maybe something like:

```
[ - name: <string>
    values: <string>, [...] ]
```

discovery/ec2/ec2.go
```diff
@@ -107,6 +114,11 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		}
 		c.Region = region
 	}
+	for _, f := range c.Filters {
```
brian-brazil: You should add a config test for this, they're under the config/testdata directory.

discovery/ec2/ec2.go
```diff
@@ -212,7 +226,19 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 	tg = &targetgroup.Group{
 		Source: *d.aws.Region,
 	}
-	if err = ec2s.DescribeInstancesPages(nil, func(p *ec2.DescribeInstancesOutput, lastPage bool) bool {
+
+	var filters []*ec2.Filter
+
```
brian-brazil: You can remove this blank line

brian-brazil:
Thanks!

docs/configuration/configuration.md
paulcdejean: @brian-brazil It's really unfortunate that I had to go to the pull request to see an example of the syntax on this. It's not clear from the currently existing documentation and no examples are provided.

I wish this config example had made it into the documentation it was very helpful.

PR #3895, opened 2018-02-28
upgrade k8s client to v6.0.0

krasi-georgiev:
fixes: #3685

it was a big pain to update the client due to enormous dependency web

PR #4013, opened 2018-03-26
Refactor Kubernetes Discovery

cofyc:
I splited the PR into two parts:

- Vendor files: https://github.com/prometheus/prometheus/pull/4013/commits/ef1dd80be4119ea6269e340ef0cc10b9e0f4dc31
- Refactoring: https://github.com/prometheus/prometheus/pull/4013/commits/ed2c78350fbc61145460edea0a2121f0bde572e0 https://github.com/prometheus/prometheus/pull/4013/commits/5b1fad92a1f1cc79f86b64be380c355587d2cc49 https://github.com/prometheus/prometheus/pull/4013/commits/c880d5e477316eb58064f343f9d073779c2dfa82 https://github.com/prometheus/prometheus/commit/e2a3a3f6260a816a0b77fab6c191cd685db2a973

What refactoring does:

- Do initial listing and syncing to scrape manager, then register event
  handlers may lost events happening in listing and syncing (if it
  lasted a long time). We should register event handlers at the very
  begining, before processing just wait until informers synced (sync in
  informer will list all objects and call event handler).
- Use a queue then we don't block event callbacks and an object will be
  processed only once if added multiple times before it being processed.
- Fix bug in `serviceUpdate` in endpoints.go, we should build endpoints
  when `exists && err == nil`. Add `^TestEndpointsDiscoveryWithService`
  tests to test this feature.

Testing:

- Use `k8s.io/client-go` testing framework and fake implementations which are
  more robust and reliable for testing.
- `Test\w+DiscoveryBeforeRun` are used to test objects created before
  discoverer runs
- `Test\w+DiscoveryAdd\w+` are used to test adding objects
- `Test\w+DiscoveryDelete\w+` are used to test deleting objects
- `Test\w+DiscoveryUpdate\w+` are used to test updating objects
- `TestEndpointsDiscoveryWithService\w+` are used to test endpoints
  events triggered by services
- `cache.DeletedFinalStateUnknown` related stuffs are removed, because
  we don't care deleted objects in store, we only need its name to send
  a specical `targetgroup.Group` to scrape manager


grobie:
Thank you a lot for your contribution!

I'm extremely worried about the lack and even _removal_ of tests in this PR. We've experienced issues with all released Prometheus versions starting v2.0, and we still receive reports for broken K8s discovery in the 2.2.1 release (see #4019 among others).

I acknowledge that testing the interaction with Kubernetes is not trivial, but especially when continuing to refactor the code, we should ensure the correctness of the implementation and have a good test suite.

cofyc:
@grobie `Test\w+Initial` tests are removed because there are no initial code anymore, so I simply removed them. I'll add more tests to cover more code paths.

brancz:
The general approach seems right, but I'd prefer to convert the removed tests from testing the implementation specific behavior to tests that ensure that events are not missed if objects are added between the informer being created to `Run` being called, which is the actual issue we're experiencing if I understand correctly.

discovery/kubernetes/endpoints.go
```diff
@@ -152,6 +109,81 @@ func (e *Endpoints) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
...
+			return
+		}
+		level.Debug(e.logger).Log("msg", "endpoints update", "tg", fmt.Sprintf("%#v", tg))
+		select {
+		case <-ctx.Done():
+		case ch <- []*targetgroup.Group{tg}:
+		}
+	}
+
+	workFunc := func() bool {
```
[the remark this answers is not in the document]
cofyc: @brancz OK.

cofyc:
@brancz 

Hi, I am trying to integrate kubernetes client-go [testing framework]( https://godoc.org/k8s.io/client-go/testing ) and [fake implementations](https://godoc.org/k8s.io/client-go/kubernetes/fake) now. Tests will be robust and reliable.

cofyc:
@grobie @brancz done. PR are split into two parts, see latest changes: https://github.com/prometheus/prometheus/pull/4013#issue-177331376.

discovery/kubernetes/endpoints.go
```diff
@@ -152,31 +110,102 @@ func (e *Endpoints) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
...
+		}
+		level.Debug(e.logger).Log("msg", "endpoints update", "tg", fmt.Sprintf("%#v", tg))
+		select {
+		case <-ctx.Done():
+		case ch <- []*targetgroup.Group{tg}:
+		}
+	}
+
+	go func() {
+		for e.process(send) {
```
brancz: I think it makes more sense to pass down the channel and have a separate "send" function bound to the respective role struct (Endpoints, Node, Pod, Service, Ingress). That way the `Run` functions will just be "wait for caches to be synced then start processing".

discovery/kubernetes/endpoints.go
```diff
@@ -152,31 +110,102 @@ func (e *Endpoints) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
...
+// Run implements the Discoverer interface.
+func (e *Endpoints) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
+	defer e.queue.ShutDown()
+
+	cacheSyncs := []cache.InformerSynced{
+		e.endpointsInf.HasSynced,
+		e.serviceInf.HasSynced,
+		e.podInf.HasSynced,
+	}
+	if !cache.WaitForCacheSync(ctx.Done(), cacheSyncs...) {
```
brancz: as the method is written with a variadic parameter so let's use it that way, then there's no need for the []cache.InformerSynced slice before.

discovery/kubernetes/endpoints.go
cofyc: @brancz Done in https://github.com/prometheus/prometheus/pull/4013/commits/e2a3a3f6260a816a0b77fab6c191cd685db2a973.

Actually, `send` functions are almost same (except some do not log). I unified it, code is more concise and consistent.

Example debug msg:

```
level=debug ts=2018-04-10T08:51:53.442225749Z caller=kubernetes.go:380 msg="kubernetes discovery update" role=endpoints tg="&targetgroup.Group{Targets:[]model.LabelSet(nil), Labels:model.LabelSet(nil), Source:\"endpoints/default/testendpoints\"}"
```

discovery/kubernetes/endpoints.go
cofyc: Done in https://github.com/prometheus/prometheus/commit/e2a3a3f6260a816a0b77fab6c191cd685db2a973.

grobie:
Awesome, thank you so mcuh @cofyc! I found a memory leak in the k8s API #4095. There is this pull request which updates the Kubernetes API #3895 from @krasi-georgiev. Somehow we'll need to merge your two PRs.

brancz:
@grobie I say we merge this, and rebase the client-go upgrade on top of it (it already needs to be rebased) and get this through a release process with various pre-releases to give people chances to test this thoroughly.

grobie:
I guess it boils down to whether release v2.2.2 with just @krasi-georgiev's k8s package update, or to release v2.3.0 with all the current master changes and this refactoring. I'd be fine with both. I would make a guess what @beorn7 would prefer though. I can confirm that v2.2.1 has k8s memory leak.

brancz:
Sorry if I sounded very set in stone in my last comment, I'm happy with either way. I would understand the reasoning of cutting the bugfix release first and then 2.3 separately.

brian-brazil:
We've built up quite a few changes, so I'd like to cut 2.3.0 reasonably soon.

discovery/kubernetes/endpoints.go
```diff
@@ -152,31 +110,85 @@ func (e *Endpoints) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
...
-	}
-	endpoints, ok = deletedState.Obj.(*apiv1.Endpoints)
-	if !ok {
-		return nil, fmt.Errorf("DeletedFinalStateUnknown contained non-Endpoints object: %v", deletedState.Obj)
-	}
-	return endpoints, nil
+	return nil, fmt.Errorf("Received unexpected object: %v", o)
 }
 
 func endpointsSource(ep *apiv1.Endpoints) string {
```
simonpasquier: nit, it could be rewritten like this to avoid duplication:

```go
func endpointsSource(ep *apiv1.Endpoints) string {
    return endpointsSourceFromNamespaceAndName(s.Namespace, s.Name)
}
```

The same comment applies for the `<role>Source(...)` functions in the other files.

cofyc:
@simonpasquier All fixed in https://github.com/prometheus/prometheus/pull/4013/commits/f893c1cb01b77216fd492d016ca5cc36e3c71fcd.

discovery/kubernetes/kubernetes_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+	if d.expectedRes != nil {
+		res := <-resChan
+		requireTargetGroups(t, d.expectedRes, res)
+	}
+}
+
+// readResultWithoutTimeout reads all targegroups from channel with timeout.
+// It merges targegroups by source and sends the result to result channel.
+func readResultWithoutTimeout(t *testing.T, ch <-chan []*targetgroup.Group, max int, timeout time.Duration, resChan chan<- map[string]*targetgroup.Group) {
```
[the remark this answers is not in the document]
cofyc: Thanks! It's my typo.

beorn7:
I have cherrypicked this PR into release-2.2 and will run it on our canaries today (in hope it will fix #4095 ). If that's really the case, I think we should not merge this PR but merge my cherrypicked commits into release-2.2 and then merge release-2.2 into master (to avoid different commits with essentially the same code changes).

beorn7:
Good news: No data races, and the memory leak #4095 seems not to happen (but need to observe a bit longer for conclusive evidence).

Bad news: It's still not picking up new K8s pods and keeps trying to scrape pods that have gone away.

beorn7:
See #4117

beorn7:
All the changes of this PR should be in master, albeit as different commits. See #4117 . Closing this one.

brancz:
Thanks a lot for your work @cofyc!

beorn7:
https://github.com/prometheus/prometheus/issues/4124 for the target update issue.

PR #4020, opened 2018-03-28
promql: allow `.` in metric names

gouthamve:
Tests pending.

This would allow people implement PromQL on datasources that also support metrics with `.` in their name. We are the only system I know of that doesn't allow `.` and given how many systems want to integrate with PromQL, I think it only makes sense to give them a way to do so.

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/prometheus/4020)

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -270,6 +270,11 @@ func (ng *Engine) newQuery(q storage.Queryable, expr Expr, start, end time.Time,
 	return qry
 }
 
+// NewQuery returns a Query. Most users should use NewRangeQuery or NewInstantQuery.
+func (ng *Engine) NewQuery(q storage.Queryable, expr Expr, start, end time.Time, interval time.Duration) Query {
+	return ng.newQuery(q, expr, start, end, interval)
+}
+
 // testStmt is an internal helper statement that allows execution
 // of an arbitrary function during handling. It is used to test the Engine.
 type testStmt func(context.Context) error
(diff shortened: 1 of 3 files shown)


brian-brazil:
I'm against this change. It adds complexity to the code for something that is of no use to Prometheus and will cause us additional support load and also limits us for future changes.

RichiH:
As this could touch namespacing, also see here: https://docs.google.com/document/d/1q7HSZyRv4Ay4hTlva4uy8WPh5WlYv9S0Yy-xQS2vKo4/edit @ 2018-02-13

cstyan:
To me the fact that metric names can't include `.`, meaning that getting a very specific metric/timeseries requires the metric name and some set of labels, makes it much easier as a user to find out which metrics are available. Supporting the `.` naming format is a step backwards imo.

The naming format that Prometheus uses is one selling point over systems that allow `.` in the name and use that as a way to get to a specific service/series/error number/etc.

brian-brazil:
This PR is not about changing the Prometheus data model, it's about whether we should facilitate systems with other data models in our code.

cstyan:
@brian-brazil I know. I was just pointing out that the current data model is a plus that we have over other systems data models. The same way that we discourage users from using the push gateway for all of their metrics, because Prometheus is designed to use a pull based model, supporting the `.` style metric names would put us in a situation where a significant amount of users may never use the intended name + labels data model. At least, that's what I would suspect would happen.

brian-brazil:
I agree on those benefits, and there's others too but that's beside the point. This PR is not the direction we're going in.

PR #4027, opened 2018-03-29
Add an OpenTracing span for each rule

bboreham:
So that tags and child spans can be traced back to the rule that they refer to.

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -325,16 +325,16 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
 		}
 
 		func(i int, rule Rule) {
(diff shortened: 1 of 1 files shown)


rules/manager.go
```diff
@@ -331,6 +332,9 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
 
 			evalTotal.Inc()
 
+			sp, ctx := opentracing.StartSpanFromContext(ctx, "rule")
+			sp.SetTag("name", rule.Name())
+			defer sp.Finish()
```
brian-brazil: You can reuse the defer above

bboreham: ok, modified as suggested

brian-brazil:
Thanks!

PR #4033, opened 2018-03-31
Yaml UnmarshalStrict implementation.

### LINKED ISSUE #1275: Config parser does not complain about overriden parts of configuration
While I readily agree that autogenerated config files should not have this problem, and that YAML works that way, especially in the trial phase, this can confuse users. Warning or, preferably, erroring out would be appreciated.

If you run with a configuration like:

    scrape_configs:
      - job_name: 'prometheus_system'
        target_groups:
        - targets: ['localhost:9100']

    scrape_configs:
      - job_name: 'foo_system'
        target_groups:
        - targets: ['foot:9100']

the former `scrape_configs` is silently discarded. While that's the obvious case, constructs like

    scrape_configs:
      - job_name: 'prometheus_system'
        target_groups:
        - targets: ['localhost:9100']

    rule_files:
      - 'prometheus.rules'
    
    scrape_configs:
      - job_name: 'foo_system'
        target_groups:
        - targets: ['foo:9100']

are silently discarded as well.

So to summarize: Yes, I fat-fingered, but principle of least surprise would be to error out instead of silently carrying on.

### LINKED ISSUE #4017: Use UnmarshalStrict for yaml Unmarshalling
The yaml library we are using now support `UnmarshalStrict` which will return an error for for unknowns, or double definitions
https://godoc.org/gopkg.in/yaml.v2#UnmarshalStrict

After start using `UnmarshalStrict` `CheckOverflow` will be no longer needer
https://github.com/prometheus/prometheus/blob/dfd6709a44eb4a809c41b6676500b8b4ccfd9af1/util/yaml/yaml.go#L22-L31

manosf:
This resolves #4017.
As suggested, I updated configuration to call `UnmarshalStrict()` over simple 
unmarshalling, thus rendering `util/yaml/yaml.go` no longer needed.
The original vendor package `gopkg.in/yaml.v2` was updated and used (https://godoc.org/gopkg.in/yaml.v2#UnmarshalStrict).
Tests ran successfully against `config/` and using `make test` with go1.10.
Edit: As @krasi-georgiev pointed out, this also fixes #1275.


config/config.go
```diff
@@ -668,8 +624,7 @@ func (c *RemoteWriteConfig) UnmarshalYAML(unmarshal func(interface{}) error) err
 	if err := c.HTTPClientConfig.Validate(); err != nil {
 		return err
 	}
-
-	return yaml_util.CheckOverflow(c.XXX, "remote_write")
+	return nil
```
krasi-georgiev: `return c.HTTPClientConfig.Validate()`

config/config.go
```diff
@@ -722,13 +674,12 @@ func (c *RemoteReadConfig) UnmarshalYAML(unmarshal func(interface{}) error) erro
...
-
 	// The UnmarshalYAML method of HTTPClientConfig is not being called because it's not a pointer.
 	// We cannot make it a pointer as the parser panics for inlined pointer structs.
 	// Thus we just do its validation here.
 	if err := c.HTTPClientConfig.Validate(); err != nil {
 		return err
 	}
 
-	return yaml_util.CheckOverflow(c.XXX, "remote_read")
+	return nil
```
krasi-georgiev: `return c.HTTPClientConfig.Validate()`

krasi-georgiev:
noticed that `pro.../common/config` also uses `checkOverflow` so I opened a  follow up issue https://github.com/prometheus/common/issues/126 to remove to update that package as well.

krasi-georgiev:
LGTM

waiting for one more LGTM and will merge.

brian-brazil:
There seems to be quite a few unrelated cleanup changes in here. Considering the breadth of this change, I'd prefer this be limited to just the core change.

krasi-georgiev:
all changes are related to the UnmarshalStrict usage.

brian-brazil:
No, there's quite a few other changes in here.

krasi-georgiev:
can you put a comment next to the line you think is unrelated and will double check.

brian-brazil:
There are multiple unrelated changes to error handling and scoping.

manosf:
I reverted some of the changes, if there are any more to be reverted please let me know which ones when you have the time. Thank you.

discovery/file/file.go
```diff
@@ -385,7 +378,7 @@ func (d *Discovery) readFile(filename string) ([]*targetgroup.Group, error) {
 			return nil, err
 		}
 	case ".yml", ".yaml":
-		if err := yaml.Unmarshal(content, &targetGroups); err != nil {
+		if err := yaml.UnmarshalStrict(content, &targetGroups); err != nil {
```
brian-brazil: Looks like we don't have unittests for this, can you add one?

config/config_test.go
```diff
@@ -657,7 +657,7 @@ var expectedErrors = []struct {
 		errMsg:   "invalid rule file path",
 	}, {
 		filename: "unknown_attr.bad.yml",
-		errMsg:   "unknown fields in scrape_config: consult_sd_configs",
+		errMsg:   "field consult_sd_configs not found in type config.plain",
```
brian-brazil: consult -> consul

brian-brazil:
Thanks!

PR #4034, opened 2018-04-01
main: actor functionality comments

si74:


krasi-georgiev:
do we really need these?
Isn't it obvious from the variable names inside each block?

juliusv:
@krasi-georgiev I like it, both because it allows you to find things at a quick glance, and because some of the constructs aren't immediately obvious to the uninitiated, especially when you're not too familiar with the group/actor semantics yet.

Disclaimer: I proposed this to @si74 as a first small change to see where it goes :)

krasi-georgiev:
sure it doesn't hurt to have these.

PR #4044, opened 2018-04-04
fix of endless loop while doing Consul service discovery.

sev3ryn:
There is and endless loop while doing Consul service discovery, reloading Prometheus configs does not stop this loop. This makes a goroutine leak every time prometheus configs are reloaded.

To reproduce bug:
1. start Prometheus with log.level=debug pointing to one of Consul services. Wait until line  `Watching service` appears. 
2. change Prometheus config pointing to another service.
3. reload config sending SIGHUP. 
4. Previous `Watching service` message continues to appear along with a new one.

cc: @brian-brazil

### DIFF
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -312,8 +312,9 @@ func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
 			case <-ctx.Done():
 				ticker.Stop()
 				return
-			case <-ticker.C:
+			default:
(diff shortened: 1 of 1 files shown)


brian-brazil:
I'd rather avoid that delay

sev3ryn:
@iksaif - good point - fixed that. See cb6aff340510d4bbbb02f77bbf2c69cf0620dc63
@brian-brazil I'm running latest Prometheus version from master

brian-brazil:
Thanks!

PR #4050, opened 2018-04-05
add queue_time slice to query_duration_seconds

### LINKED ISSUE #4049: Queue time component of query duration is not exposed
The queue time for a query is calculated and included in the total query time in query_stats, but there is no `slice="queue_time"` label in the `query_duration_seconds` metric where everything else from query_stats is exposed.

I looked at `promql/engine.go` history but couldn't see anywhere it was explicitly removed, so I can only figure it was an oversight.

In any case it was trivial to add, so I'll have a PR for it shortly.

roganartu:
Fixes #4049


brian-brazil:
Thanks!

PR #4055, opened 2018-04-06
vendor: update golang client

si74:


brian-brazil:
Why are you updating only some of the client?

gouthamve:
I think it was a `govendor add github.com/prometheus/client_golang`. The top level has only md files and I don't think it's needed.

si74:
yep! that's correct. I did a govendor fetch github.com/prometheus/client_golang

brian-brazil:
Okay, nothing to do here then.

Issue #1605, opened 2016-04-29
Promtool: Add querying functionality


beorn7:
We want to replace prometheus_cli with promtool.

See also https://github.com/prometheus/prometheus_cli/issues/9

alileza:
Does anyone working on this?

beorn7:
I don't think so. @fabxc might know more about the current state about the API client library. See prometheus/prometheus_cli#9

fabxc:
No one is currently working on this.
I'm happy to provide guidance to anyone willing to tackle this.

alileza:
Nice, i guess i'll try to start working on this.

fabxc:
Sounds great!
Be aware of the API client package:
https://godoc.org/github.com/prometheus/client_golang/api/prometheus

I've recently been thinking about kicking out the dependency on the `model`
package. But with vendoring those potential changes are not blocking.

alileza:
Ok, so  basically we want to move `prometheus_cli` features into `promtool` using [client_golang api](https://godoc.org/github.com/prometheus/client_golang/api/prometheus)?

alileza:
is this changes is what we expect on promtool? if it is, i'll continue to work on query_range etc and submit at PR.
https://github.com/alileza/prometheus/commit/36da400c342eb7a2ebcf6f39c5f44bed39f4ab17

beorn7:
Sorry, dropped the ball here. Yes, this all sounds good to me. What @fabxc tried to say about the prometheus/common/model dependency is essentially "don't worry". @fabxc Please shout if there is anything else to say at this point.

xmudrii:
What's the status of this issue?
I saw Hacktoberfest label and I want to contribute to Prometheus project for the first time.

If this is still needed things I would love to try working on it. Also if there are some instructions/guidelines I would appreciate them.

brancz:
Not sure how far we want to go with this, but for a quickstart I threw this together a while ago, and just queried further with jq. It's not much, but it's a start https://gist.github.com/brancz/1c252d4c87a48c80df7c6b046823c466

Questions we may want to ask ourselves is, which format(s) do we want to output? Interactive? One-off?

Personally I don't think this needs any dependency on common/model, but that also somewhat depends on what we want to build here.

si74:
Hey! I'd be interested working on this.

krasi-georgiev:
based on the original issue this is now fixed

PR #4056, opened 2018-04-06
vendor: correctly update golang client

si74:
oops! had a bit of a mistake in vendoring this last time. this should do the trick now.

### DIFF
--- a/vendor/vendor.json
+++ b/vendor/vendor.json
@@ -742,16 +742,28 @@
 			"revisionTime": "2016-06-15T09:26:46Z"
 		},
 		{
-			"checksumSHA1": "I87tkF1e/hrl4d/XIKFfkPRq1ww=",
-			"path": "github.com/prometheus/client_golang/prometheus",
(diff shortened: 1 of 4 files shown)


vendor/vendor.json
```diff
@@ -741,17 +741,29 @@
 			"revision": "d77da356e56a7428ad25149ca77381849a6a5232",
 			"revisionTime": "2016-06-15T09:26:46Z"
 		},
+		{
+			"checksumSHA1": "4VppKBbzCSmoFfQxGNUm0TYiFCA=",
+			"path": "github.com/prometheus/client_golang/api",
```
brian-brazil: Why are you pulling these in? We don't use them.

si74: I'm working a particular open task (adding query to promtool) and was hoping to use these for them.

si74: This one! https://github.com/prometheus/prometheus/issues/1605

brian-brazil:
Usually we only add new dependencies in the PR that needs them.

si74:
oops! thanks for the heads up - will do so next time

PR #4060, opened 2018-04-08
Expose controller kind and name to labelling rules

dlespiau:
Relabelling rules can use this information to attach the name of the controller that has created a pod.

In turn, this can be used to slice metrics by workload at query time, ie. "Give me all metrics that have been created by the $name Deployment"

As an illustration when used with the following labelling rules:
```
            - source_labels:
                - __meta_kubernetes_pod_controller_kind
              target_label: kubernetes_pod_controller_kind
            - source_labels:
                - __meta_kubernetes_pod_controller_name
              target_label: kubernetes_pod_controller_name
```
we get:
```
node_disk_bytes_written{device="sr0",instance="192.168.99.106:9100",job="kubernetes-pods",kubernetes_pod_controller_kind="DaemonSet",kubernetes_pod_controller_name="prom-node-exporter",node="minikube"}
```
Of course, with a Deployment we only get access to the ReplicatSet, but that's enough to derive the Deployment name with an extra regex in the labelling rules:
```
kubedns_dnsmasq_errors{job="kubernetes-pods",kubernetes_pod_controller_kind="ReplicaSet",kubernetes_pod_controller_name="kube-dns-1301475494"}

```

### DIFF
--- a/discovery/kubernetes/pod.go
+++ b/discovery/kubernetes/pod.go
@@ -23,6 +23,7 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/common/model"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/pkg/api"
 	apiv1 "k8s.io/client-go/pkg/api/v1"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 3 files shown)


brancz:
Sorry, I missed this, thanks for assigning me. (@dlespiau for future Kubernetes SD PRs feel free to request a review from me directly)

Generally I'm skeptical whether this is the correct thing to do as there can at least theoretically be multiple owners that are controllers, in practice however I don't see how there can be multiple controllers for one Pod and given that the upstream apimachinery code uses the same method I'm ok with this.

Looking at this with fresh eyes again, I think we should probably do the same thing in kube-state-metrics, we used to have this when the `created-by` annotations still existed so I think it would be useful to have.

lgtm :+1:

Issue #422, opened 2014-12-10
Persist alert 'for' state across restarts


brian-brazil:
Currently if Prometheus restarts, we lose the 'for' state for firing alerts. While this isn't an issue for short for clauses, it presents a problem for clauses that are in the hours to days range.

It'd be good to persist this state in some way, so that alerts don't have to start again from scratch. We probably don't want to count the time the prometheus server is down against the 'for' clause.

Gouthamve:
How would this work? Do we evaluate from the point we terminated on startup again as a way of catching up?

For example take an alert A1 with `FOR 12h` that has started life at 0000. Now it is pending and we shut down the server at 0600 and we persist `A1 pending for 6h at 0600`. Suppose we restart at 01500, do we start evaluating from 0600-01500 and then say catched up, continue normal evaluation? Isn't this going to be costly?

Are we going to have a threshold? Something like `you lose alert state if Prometheus is down for more than 1h`? We can make this a flag to support use-cases where the server was down for a long time.

brian-brazil:
> Suppose we restart at 01500, do we start evaluating from 0600-01500 and then say catched up, continue normal evaluation?

No, and the alert probably wouldn't be firing during that period anyway as there'll be no data from the Prometheus going down.

What we want is that we remember that 6h of the FOR has already been satisfied, and when it's satisfied for another 6h then we alert. We need to be a little careful here, for safety if the remaining time is <10m we should treat that as min(remaining-10m, for).

To implement I was thinking an analouge to ALERTs that'd track when an alert started, and when Promethus restarts it'd check that metric (adjusting if needed for the 10m safety) and initialise internal state accordingly.


One other thing is that it can take a while for a Prometheus to warm up and have done enough scrapes for alerts to start firing properly. Thus we should delay rule&alert evaluation for 2 scrape_intervals at starttup to mitigate this.

Gouthamve:
Okay, biggest facepalm moment ever! How could I not see that the server is not running! Sorry about that!

Coming to when to start evaluation, will 2 `global scrape_interval` be okay? Because we might be alerting on series that scraped much less frequently. And what about alerts on range queries?

brian-brazil:
global scrape interval is all that makes sense.

Range queries will usually be rate(), and two scrapes+slack should be enough for that to work.

jacksontj:
[...]

PR #4061, opened 2018-04-08
Persist alert 'for' state across restarts

codesome:
Addresses [#422](https://github.com/prometheus/prometheus/issues/422)

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -248,7 +248,7 @@ func main() {
 		)
 
 		ruleManager = rules.NewManager(&rules.ManagerOptions{
-			Appendable:  fanoutStorage,
+			Storage:     fanoutStorage,
 			QueryFunc:   rules.EngineQueryFunc(queryEngine, fanoutStorage),
 			NotifyFunc:  sendAlerts(notifier, cfg.web.ExternalURL.String()),
 			Context:     ctxRule,
--- a/rules/alerting.go
+++ b/rules/alerting.go
@@ -286,7 +286,7 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc, 
 			if a.State != StateInactive {
 				a.State = StateInactive
 				a.ResolvedAt = ts
-				vec = append(vec, r.forStateSample(a, ts, 0))
+				vec = append(vec, r.forStateSample(a, ts, -1))
 			}
 			continue
 		}
--- a/rules/alerting.go
+++ b/rules/alerting.go
@@ -62,16 +62,6 @@ const (
 	StateFiring
 )
 
-// AlertForState denotes the 'for' state of an active alert.
-type AlertForState float64
-
-const (
-	// ForStateInactive is the state when alert expression is false.
-	ForStateInactive AlertForState = iota
-	// ForStateActive is the state when alert expression is true.
-	ForStateActive
-)
-
 func (s AlertState) String() string {
 	switch s {
 	case StateInactive:
@@ -170,13 +160,10 @@ func (r *AlertingRule) forStateSample(alert *Alert, ts time.Time, v float64) pro
 
 	for _, l := range alert.Labels {
 		lb.Set(l.Name, l.Value)
-		fmt.Println("-->", l.Name, l.Value)
 	}
 
 	lb.Set(labels.MetricName, alertForStateMetricName)
-	fmt.Println("-->", labels.MetricName, alertForStateMetricName)
 	lb.Set(labels.AlertName, r.name)
-	fmt.Println("-->", labels.AlertName, r.name)
 
 	s := promql.Sample{
 		Metric: lb.Labels(),
@@ -299,7 +286,6 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc, 
 			if a.State != StateInactive {
 				a.State = StateInactive
 				a.ResolvedAt = ts
-				fmt.Println("Not Active added")
 				vec = append(vec, r.forStateSample(a, ts, 0))
 			}
 			continue
@@ -311,7 +297,6 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc, 
 		}
 
 		vec = append(vec, r.sample(a, ts))
-		fmt.Printf("Active added %f\n", float64(a.ActiveAt.UnixNano()))
 		vec = append(vec, r.forStateSample(a, ts, float64(a.ActiveAt.UnixNano())))
 	}
 
@@ -357,6 +342,19 @@ func (r *AlertingRule) currentAlerts() []*Alert {
 		alerts = append(alerts, &anew)
 	}
 	return alerts
+}
+
+// ForEachActiveAlert runs the given function on each alert.
+// This should be used when you want to use the actual alerts from the AlertingRule
+// and not on its copy.
+// If you want to run on a copy of alerts then don't use this, get the alerts from 'ActiveAlerts()'.
+func (r *AlertingRule) ForEachActiveAlert(f func(*Alert)) {
+	r.mtx.Lock()
+	defer r.mtx.Unlock()
+
+	for _, a := range r.active {
+		f(a)
+	}
 }
 
 func (r *AlertingRule) String() string {
@@ -410,7 +408,7 @@ func (r *AlertingRule) HTMLSnippet(pathPrefix string) html_template.HTML {
 	return html_template.HTML(byt)
 }
 
-// SubtractHoldDuration  TODO: add proper comment
+// SubtractHoldDuration subtracts holdDuration from given ts.
 func (r *AlertingRule) SubtractHoldDuration(ts time.Time) time.Time {
 	return ts.Add(-r.holdDuration)
 }
--- a/rules/alerting.go
+++ b/rules/alerting.go
@@ -39,6 +39,8 @@ import (
 const (
 	// AlertMetricName is the metric name for synthetic alert timeseries.
 	alertMetricName = "ALERTS"
+	// AlertForStateMetricName is the metric name for 'for' state of alert.
+	alertForStateMetricName = "ALERTS_FOR_STATE"
 
 	// AlertNameLabel is the label name indicating the name of an alert.
 	alertNameLabel = "alertname"
@@ -58,6 +60,16 @@ const (
 	// StateFiring is the state of an alert that has been active for longer than
 	// the configured threshold duration.
 	StateFiring
+)
+
+// AlertForState denotes the 'for' state of an active alert.
+type AlertForState float64
+
+const (
+	// ForStateInactive is the state when alert expression is false.
+	ForStateInactive AlertForState = iota
+	// ForStateActive is the state when alert expression is true.
+	ForStateActive
 )
 
 func (s AlertState) String() string {
@@ -153,6 +165,26 @@ func (r *AlertingRule) sample(alert *Alert, ts time.Time) promql.Sample {
 	return s
 }
 
+func (r *AlertingRule) forStateSample(alert *Alert, ts time.Time, v float64) promql.Sample {
+	lb := labels.NewBuilder(r.labels)
+
+	for _, l := range alert.Labels {
+		lb.Set(l.Name, l.Value)
+		fmt.Println("-->", l.Name, l.Value)
+	}
+
+	lb.Set(labels.MetricName, alertForStateMetricName)
+	fmt.Println("-->", labels.MetricName, alertForStateMetricName)
+	lb.Set(labels.AlertName, r.name)
+	fmt.Println("-->", labels.AlertName, r.name)
+
+	s := promql.Sample{
+		Metric: lb.Labels(),
+		Point:  promql.Point{T: timestamp.FromTime(ts), V: v},
+	}
+	return s
+}
+
 // SetEvaluationTime updates evaluationTime to the duration it took to evaluate the rule on its last evaluation.
 func (r *AlertingRule) SetEvaluationTime(dur time.Duration) {
 	r.mtx.Lock()
@@ -186,6 +218,7 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc, 
 	// or update the expression value for existing elements.
 	resultFPs := map[uint64]struct{}{}
 
+	var vec promql.Vector
 	for _, smpl := range res {
 		// Provide the alert information to the template.
 		l := make(map[string]string, len(smpl.Metric))
@@ -239,6 +272,7 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc, 
 
 		// Check whether we already have alerting state for the identifying label set.
 		// Update the last value and annotations if so, create a new alert entry otherwise.
+
 		if alert, ok := r.active[h]; ok && alert.State != StateInactive {
 			alert.Value = smpl.V
 			alert.Annotations = annotations
@@ -254,7 +288,6 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc, 
 		}
 	}
 
-	var vec promql.Vector
 	// Check if any pending alerts should be removed or fire now. Write out alert timeseries.
 	for fp, a := range r.active {
 		if _, ok := resultFPs[fp]; !ok {
@@ -266,6 +299,8 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc, 
 			if a.State != StateInactive {
 				a.State = StateInactive
 				a.ResolvedAt = ts
+				fmt.Println("Not Active added")
+				vec = append(vec, r.forStateSample(a, ts, 0))
 			}
 			continue
 		}
@@ -276,6 +311,8 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc, 
 		}
 
 		vec = append(vec, r.sample(a, ts))
+		fmt.Printf("Active added %f\n", float64(a.ActiveAt.UnixNano()))
+		vec = append(vec, r.forStateSample(a, ts, float64(a.ActiveAt.UnixNano())))
 	}
 
 	return vec, nil
@@ -371,4 +408,9 @@ func (r *AlertingRule) HTMLSnippet(pathPrefix string) html_template.HTML {
 		return html_template.HTML(fmt.Sprintf("error marshalling alerting rule: %q", err.Error()))
 	}
 	return html_template.HTML(byt)
+}
+
+// SubtractHoldDuration  TODO: add proper comment
+func (r *AlertingRule) SubtractHoldDuration(ts time.Time) time.Time {
+	return ts.Add(-r.holdDuration)
 }
--- a/rules/alerting.go
+++ b/rules/alerting.go
@@ -155,6 +155,8 @@ func (r *AlertingRule) sample(alert *Alert, ts time.Time) promql.Sample {
 	return s
 }
 
+// forStateSample returns the sample for ALERTS_FOR_STATE.
+// v should be -1 for StateInactive, UnixNano for StatePending and StateFiring.
 func (r *AlertingRule) forStateSample(alert *Alert, ts time.Time, v float64) promql.Sample {
 	lb := labels.NewBuilder(r.labels)
 
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -537,53 +537,67 @@ func (m *Manager) restoreForState(ts time.Time) {
 			if alertRule, ok := rule.(*AlertingRule); ok {
 
 				_, err := alertRule.Eval(m.opts.Context, ts, newg.opts.QueryFunc, newg.opts.ExternalURL)
-				if err == nil {
-					mint := alertRule.SubtractHoldDuration(ts)
-					mintMS := int64(model.TimeFromUnixNano(mint.UnixNano()))
-
-					q, err := m.opts.Storage.Querier(m.opts.Context, mintMS, maxtMS)
-					if err != nil {
-						continue
-					}
-
-					alertFunc := func(a *Alert) {
-
-						var matchers []*labels.Matcher
-						for _, l := range alertRule.labels {
-							mt, _ := labels.NewMatcher(labels.MatchEqual, l.Name, l.Value)
-							matchers = append(matchers, mt)
-						}
-						for _, l := range a.Labels {
-							mt, _ := labels.NewMatcher(labels.MatchEqual, l.Name, l.Value)
-							matchers = append(matchers, mt)
-						}
-						mt, _ := labels.NewMatcher(labels.MatchEqual, labels.MetricName, alertForStateMetricName)
-						matchers = append(matchers, mt)
-						mt, _ = labels.NewMatcher(labels.MatchEqual, labels.AlertName, alertRule.name)
-						matchers = append(matchers, mt)
-
-						sset, err := q.Select(nil, matchers...)
-						if err == nil && sset.Next() {
-							// Series found for the 'for' state.
-							var v float64
-							s := sset.At()
-							it := s.Iterator()
-							for it.Next() {
-								_, v = it.At()
-							}
-
-							if v >= 0 {
-								restoredTime := model.TimeFromUnixNano(int64(v))
-								a.ActiveAt = restoredTime.Time()
-								level.Info(m.logger).Log("msg", "'for' state restored",
-									labels.AlertName, alertRule.name, "restored_time_ms", int64(restoredTime))
-							}
-						}
-					}
-
-					alertRule.ForEachActiveAlert(alertFunc)
-
+				if err != nil {
+					level.Error(m.logger).Log("msg", "Failed to restore 'for' state",
+						"stage", "Eval", "err", err)
+					continue
 				}
+
+				mint := alertRule.SubtractHoldDuration(ts)
+				mintMS := int64(model.TimeFromUnixNano(mint.UnixNano()))
+
+				q, err := m.opts.Storage.Querier(m.opts.Context, mintMS, maxtMS)
+				if err != nil {
+					level.Error(m.logger).Log("msg", "Failed to restore 'for' state",
+						"stage", "Querier", "err", err)
+					continue
+				}
+
+				alertFunc := func(a *Alert) {
+
+					var matchers []*labels.Matcher
+					for _, l := range alertRule.labels {
+						mt, _ := labels.NewMatcher(labels.MatchEqual, l.Name, l.Value)
+						matchers = append(matchers, mt)
+					}
+					for _, l := range a.Labels {
+						mt, _ := labels.NewMatcher(labels.MatchEqual, l.Name, l.Value)
+						matchers = append(matchers, mt)
+					}
+					mt, _ := labels.NewMatcher(labels.MatchEqual, labels.MetricName, alertForStateMetricName)
+					matchers = append(matchers, mt)
+					mt, _ = labels.NewMatcher(labels.MatchEqual, labels.AlertName, alertRule.name)
+					matchers = append(matchers, mt)
+
+					sset, err := q.Select(nil, matchers...)
+					if err != nil {
+						level.Error(m.logger).Log("msg", "Failed to restore 'for' state",
+							"stage", "Select", "err", err)
+						return
+					} else if !sset.Next() {
+						// No series found for the 'for' state,
+						// hence getting active for the first time.
+						return
+					}
+
+					// Series found for the 'for' state.
+					var v float64
+					s := sset.At()
+					it := s.Iterator()
+					for it.Next() {
+						_, v = it.At()
+					}
+					if v < 0 { // Alert was not active.
+						return
+					}
+
+					restoredTime := model.TimeFromUnixNano(int64(v))
+					a.ActiveAt = restoredTime.Time()
+					level.Info(m.logger).Log("msg", "'for' state restored",
+						labels.AlertName, alertRule.name, "restored_time_ms", int64(restoredTime))
+				}
+
+				alertRule.ForEachActiveAlert(alertFunc)
 
 			}
 		}
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -30,6 +30,7 @@ import (
 	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
 
+	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/pkg/rulefmt"
 	"github.com/prometheus/prometheus/pkg/timestamp"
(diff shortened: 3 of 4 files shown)


cmd/prometheus/main.go
```diff
@@ -248,7 +248,7 @@ func main() {
 		)
 
 		ruleManager = rules.NewManager(&rules.ManagerOptions{
-			Appendable:  fanoutStorage,
+			Storage:     fanoutStorage,
```
gouthamve: Not sure we want to use `fanoutStorage`

rules/alerting.go
```diff
@@ -276,6 +299,7 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc,
 		}
 
 		vec = append(vec, r.sample(a, ts))
+		vec = append(vec, r.forStateSample(a, ts, float64(a.ActiveAt.UnixNano())))
```
gouthamve: Curious if the floating point inaccuracies will mess something later.

rules/manager.go
```diff
@@ -520,9 +521,90 @@ func (m *Manager) Update(interval time.Duration, files []string) error {
...
+func (m *Manager) restoreForState(ts time.Time) {
+
+	maxtMS := int64(model.TimeFromUnixNano(ts.UnixNano()))
+	for _, newg := range m.groups {
+		for _, rule := range newg.Rules() {
+			if alertRule, ok := rule.(*AlertingRule); ok {
+
+				_, err := alertRule.Eval(m.opts.Context, ts, newg.opts.QueryFunc, newg.opts.ExternalURL)
+				if err != nil {
+					level.Error(m.logger).Log("msg", "Failed to restore 'for' state",
```
gouthamve: Add the alerting rule also to the logs.

rules/manager.go
```diff
@@ -520,9 +521,90 @@ func (m *Manager) Update(interval time.Duration, files []string) error {
...
+						matchers = append(matchers, mt)
+					}
+					mt, _ := labels.NewMatcher(labels.MatchEqual, labels.MetricName, alertForStateMetricName)
+					matchers = append(matchers, mt)
+					mt, _ = labels.NewMatcher(labels.MatchEqual, labels.AlertName, alertRule.name)
+					matchers = append(matchers, mt)
+
+					sset, err := q.Select(nil, matchers...)
+					if err != nil {
+						level.Error(m.logger).Log("msg", "Failed to restore 'for' state",
```
gouthamve: Here too.

rules/manager.go
```diff
@@ -520,9 +521,90 @@ func (m *Manager) Update(interval time.Duration, files []string) error {
...
 }
 
+// restoreForState restores the 'for' state of the alerts
+// by looking up last ActiveAt from storage.
+func (m *Manager) restoreForState(ts time.Time) {
+
+	maxtMS := int64(model.TimeFromUnixNano(ts.UnixNano()))
+	for _, newg := range m.groups {
+		for _, rule := range newg.Rules() {
+			if alertRule, ok := rule.(*AlertingRule); ok {
```
gouthamve: You can reduce nesting by doing:

```
alertRule, ok := rule.(*AlertingRule)
if !ok {
  continue
}
```

rules/alerting.go
codesome: Floating point actually doesn't cause much harm (maybe few nano seconds?). We actually lose some small info when we are converting nano to milli [here](https://github.com/prometheus/prometheus/pull/4061/files#diff-d23c7264c420b842612a4fe05e018217R594). Hence at max we may end up losing few milliseconds (1-2), which should be fine? 

In my tests when I did multiple restarts, I saw a divergence of max 2 milliseconds from the actual time.

rules/manager.go
codesome: Done.

rules/manager.go
codesome: Done.

rules/manager.go
codesome: Done.

rules/alerting.go
```diff
@@ -266,6 +288,7 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc,
 			if a.State != StateInactive {
 				a.State = StateInactive
 				a.ResolvedAt = ts
+				vec = append(vec, r.forStateSample(a, ts, -1))
```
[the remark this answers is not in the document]
codesome: Will it also handle when we query over a range (which is being done when restoring)?

If we consider this case

* for_duration: 10m
* Event timeline:
`for_active <1m elapsed> for_inactive <1m elapsed> prom_goes_down <6m elapsed> prom_is_up and for_is_active on startup`

And if by default if I check for samples within last `1h`, I will still get samples.

In such case, if we dont add a sample when `for` state goes inactive, we might end up restoring wrong `for` state and cause premature alerts.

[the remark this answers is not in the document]
brian-brazil: Let me put it another way. You're duplicating the functionality of staleness, don't do that.

rules/manager.go
```diff
@@ -520,9 +521,91 @@ func (m *Manager) Update(interval time.Duration, files []string) error {
...
+func (m *Manager) restoreForState(ts time.Time) {
+
+	maxtMS := int64(model.TimeFromUnixNano(ts.UnixNano()))
+	for _, newg := range m.groups {
+		for _, rule := range newg.Rules() {
+			alertRule, ok := rule.(*AlertingRule)
+			if !ok {
+				continue
+			}
+			_, err := alertRule.Eval(m.opts.Context, ts, newg.opts.QueryFunc, newg.opts.ExternalURL)
```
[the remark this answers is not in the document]
codesome: One simple solution that I can think of is first waiting [here](https://github.com/prometheus/prometheus/blob/4a4e8a7d3bab177aa446e8b0fd4de2c41bf5298d/cmd/prometheus/main.go#L560) till the db is open. And inside the `ruleManager.Run()`, delay for one scrape interval, and move the restoring part inside `run()` of every group. Does that sound right?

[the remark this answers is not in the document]
brian-brazil: That should do it offhand, just be careful that there's no restoration logic if a new group is added due to a config reload.

cmd/prometheus/main.go
```diff
@@ -166,6 +168,12 @@ func main() {
 	a.Flag("alertmanager.timeout", "Timeout for sending alerts to Alertmanager.").
 		Default("10s").SetValue(&cfg.notifierTimeout)
 
+	a.Flag("alertmanager.outage-tolerance", "Max time to tolerate prometheus outage for restoring 'for' state of alert.").
+		Default("1h").SetValue(&cfg.outageTolerance)
+
+	a.Flag("alertmanager.for-grace-period", "Minimum duration between alert and restored 'for' state. Applies only for alerts with 'for' time greater than grace period.").
```
brian-brazil: That's not quite right. If I have an alert with a for of 30m and it was 29m pending, then we should have to wait 10m before setting it to fire.

rules/manager.go
```diff
@@ -355,15 +355,31 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
 				numDuplicates = 0
 			)
 
-			app, err := g.opts.Storage.Appender()
+			app, err := g.opts.Appendable.Appender()
 			if err != nil {
 				level.Warn(g.logger).Log("msg", "creating appender failed", "err", err)
 				return
 			}
+			tsdbApp, err := g.opts.TSDB.Appender()
```
brian-brazil: You don't need a 2nd appender for this, use the usual one.

rules/manager.go
```diff
@@ -520,9 +521,91 @@ func (m *Manager) Update(interval time.Duration, files []string) error {
...
+			}
+
+			alertFunc := func(a *Alert) {
+
+				var matchers []*labels.Matcher
+				for _, l := range alertRule.labels {
+					mt, _ := labels.NewMatcher(labels.MatchEqual, l.Name, l.Value)
+					matchers = append(matchers, mt)
+				}
+				for _, l := range a.Labels {
```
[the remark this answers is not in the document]
brian-brazil: This is still not right, alert labels are not static.

rules/manager.go
```diff
@@ -534,72 +569,77 @@ func (m *Manager) restoreForState(ts time.Time) {
...
+				sset, err := q.Select(nil, matchers...)
+				if err != nil {
+					level.Error(m.logger).Log("msg", "Failed to restore 'for' state",
+						labels.AlertName, alertRule.Name(), "stage", "Select", "err", err)
+					return
+				} else if !sset.Next() {
+					// No series found for the 'for' state,
+					// hence getting active for the first time.
+
+					level.Error(m.logger).Log("msg", "No series found")
```
brian-brazil: This is not an error.

rules/manager.go
```diff
@@ -534,72 +569,77 @@ func (m *Manager) restoreForState(ts time.Time) {
...
-						_, v = it.At()
-					}
-					if v < 0 { // Alert was not active.
-						return
-					}
+				// Series found for the 'for' state.
+				var v float64
+				s := sset.At()
+				it := s.Iterator()
+				for it.Next() {
```
brian-brazil: You need to use the labels for these samples.

rules/manager.go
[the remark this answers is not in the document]
codesome: Labels in alert are added after expanding ([here](https://github.com/codesome/prometheus/blob/0d6ec4dc567f62ecce89cabfe0673bcfcecbbca9/rules/alerting.go#L229)), and given that it is from the last `Eval`, wont it be the expanded and new labels?

cmd/prometheus/main.go
codesome: Yes, it does that. Maybe the description was not clear. 

Are you referring to "Applies only for alerts with 'for' time greater than grace period"?
If yes, then I actually meant the `for` duration set in the config, and not the `for` duration remaining.

codesome:
@brian-brazil PTAL, I have fixed some comments since last time. These 2 comments were not very clear though.
https://github.com/prometheus/prometheus/pull/4061#discussion_r183417333
https://github.com/prometheus/prometheus/pull/4061#discussion_r183417614

rules/alerting.go
```diff
@@ -153,6 +155,25 @@ func (r *AlertingRule) sample(alert *Alert, ts time.Time) promql.Sample {
 	return s
 }
 
+// forStateSample returns the sample for ALERTS_FOR_STATE.
+// v should be -1 for StateInactive, UnixNano for StatePending and StateFiring.
```
brian-brazil: There should be no value for inactive, we don't know the labels of inactive alerts

rules/alerting.go
```diff
@@ -276,6 +298,7 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc,
 		}
 
 		vec = append(vec, r.sample(a, ts))
+		vec = append(vec, r.forStateSample(a, ts, float64(a.ActiveAt.Unix())))
```
brian-brazil: As we'll be bumping this forwards at restart, the intended firing time might be easier to understand. It's weird for ActiveAt to change

rules/manager.go
```diff
@@ -523,6 +527,91 @@ func (m *Manager) Update(interval time.Duration, files []string) error {
...
+				if value.IsStaleNaN(v) { // Alert was not active.
+					return
+				}
+
+				restoredTime := time.Unix(int64(v), 0)
+				if alertHoldDuration > m.opts.ForGracePeriod && restoredTime.Add(alertHoldDuration).Sub(ts) < m.opts.ForGracePeriod {
+					restoredTime = ts.Add(m.opts.ForGracePeriod).Add(-alertHoldDuration)
+				}
+				a.ActiveAt = restoredTime
+				level.Info(m.logger).Log("msg", "'for' state restored",
```
brian-brazil: Debug, and include the labels and timestamps

rules/manager.go
```diff
@@ -523,6 +527,91 @@ func (m *Manager) Update(interval time.Duration, files []string) error {
...
+				for it.Next() {
+					_, v = it.At()
+				}
+				if value.IsStaleNaN(v) { // Alert was not active.
+					return
+				}
+
+				restoredTime := time.Unix(int64(v), 0)
+				if alertHoldDuration > m.opts.ForGracePeriod && restoredTime.Add(alertHoldDuration).Sub(ts) < m.opts.ForGracePeriod {
+					restoredTime = ts.Add(m.opts.ForGracePeriod).Add(-alertHoldDuration)
```
brian-brazil: I don't think this is right, it always waits another 10m. If I've a 2h alert that I'm only 5m into and do an instant restart, I should have to wait the remaining 1h55m - not only 10m.

rules/manager.go
```diff
@@ -523,6 +527,91 @@ func (m *Manager) Update(interval time.Duration, files []string) error {
...
+				}
+				a.ActiveAt = restoredTime
```
brian-brazil: Should this be persisted? Will it clash with the value the Eval wrote out?

rules/alerting.go
codesome: Forgot to remove that comment. Currently staleness handling takes care of it.

rules/manager.go
codesome: `restoredTime.Add(alertHoldDuration).Sub(ts) < m.opts.ForGracePeriod` takes care of it.
`restoredTime.Add(alertHoldDuration).Sub(ts)` will be `1h55m`, which is > grace period. So it wont go inside the `if` block.

To explain the logic more clearly:
If alert became active at `0` and prometheus was down for `5m`, then:
1. `restoredTime` will be 0.
2. `alertHoldDuration > m.opts.ForGracePeriod`: we see that the hold duration is greater than grace period to apply the logic of grace period.
3. `restoredTime.Add(alertHoldDuration).Sub(ts)` can be viewed as

```
// ts = current time
  restoredTime                        ts                      firing wrt restoredTime
       |-------------------------------|-------------------------------|

       ^-----------------------alertHoldDuration-----------------------^

                                       ^-------------------------------^
                                     restoredTime.Add(alertHoldDuration).Sub(ts)

       ^--------------5m--------------^^-------------1h55m-------------^
```

4. `ts.Add(m.opts.ForGracePeriod).Add(-alertHoldDuration)`: it makes the alert to fire at `ts.Add(m.opts.ForGracePeriod)`

As `restoredTime.Add(alertHoldDuration).Sub(ts)` (`1h55m`) > `m.opts.ForGracePeriod` (`10m`), the `if` block wont run.

brian-brazil: It'll never go inside the if for an alert with more than a 10m for - which can't be right.

codesome: Isn't it the other way round?
`alertHoldDuration > m.opts.ForGracePeriod` specifies that it goes inside `if` only when the `for` is `>10m`. If `for` was `<10`, then pending the alert for another `10m` won't be right, hence doesn't go inside in that case.

rules/alerting.go
codesome: If we want to change the intended firing time, it would require changing `holdDuration`, which wont be right. I can't think of any other way to achieve this without adding another variable in the `AlertingRule` for an addition expected firing time.

rules/manager.go
codesome: > it's discarding the timestamp of the sample

The timestamp of the sample comes to use indirectly when querying. As we query for `ALERTS_FOR_STATE` in the outage tolerance period, the timestamp of the samples come into picture in the query processing.

If you talk about explicitly using the timestamp of the sample, its not needed, as the `ActiveAt` is the value of the sample.

I have updated the comments to make the logic more clear. Please let me know which part doesn't seem fine.

rules/manager.go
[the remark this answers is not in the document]
brian-brazil: Alerts could also depend on recording rules, so we really need to run those too. So we'll probably need a bit of a different control flow here

rules/manager.go
```diff
@@ -523,6 +527,124 @@ func (m *Manager) Update(interval time.Duration, files []string) error {
...
+
+				if !seriesFound {
+					return
+				}
+
+				// Series found for the 'for' state.
+				var v float64
+				it := s.Iterator()
+				for it.Next() {
+					_, v = it.At()
```
brian-brazil: I still don't see how this can be right. You're not using the timestamp at which the alert last fired.

rules/manager.go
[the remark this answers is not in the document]
codesome: Another way is:
1. Rule groups perform the first `Eval` as usual.
2. Then we perform restoration once all groups are done with the first `Eval` (a barrier).
3. After the restoration we continue the evaluations as usual.

Does this flow sound right? During restoration currently I am doing a dry `Eval` without persisting the samples, just to have the alerts updated in the rule. With this new flow the dry run can be skipped.

rules/manager.go
codesome: Before I answer this I have 1 doubt,
If the alert was firing when the prometheus went down, should it be firing again just after restoration or should it wait for the grace period?

brian-brazil: It should still be firing if it's still evaluating to true I'd say, though it if waits for the grace period that's not the end of the world.

rules/manager.go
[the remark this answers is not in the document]
brian-brazil: Something like that sounds sane as a first pass. Even if an alert isn't firing from the first eval we can restore it, so it can potentially work on the 2nd (the 1st eval may not have enough data to work with to produce all the alerts it should).

rules/manager.go
codesome: I had assumed that it should wait for grace period even in that case. If we have to wait for grace period, we actually don't need to know if the alert was firing. With the `ActiveAt`, we can make out when it should be firing, and then handle the grace period according. 

If we want it to fire immediately after restoration, then we might want to use the timestamp of the sample.

I guess we would like it to fire after restoration as you said. I will make the necessary changes.

codesome:
@brian-brazil I have updated the PR. Now the timestamp of the sample is being used to check for 'firing' state. Control flow has been modified to restore the state after the first `Eval`.

rules/alerting.go
```diff
@@ -276,6 +306,9 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc,
 		}
 
 		vec = append(vec, r.sample(a, ts))
+		if r.restored {
```
brian-brazil: We probably shouldn't write out the above line either

rules/manager.go
```diff
@@ -407,12 +422,133 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
...
+
+	for _, rule := range g.Rules() {
+		alertRule, ok := rule.(*AlertingRule)
+		if !ok {
+			continue
+		}
+
+		alertHoldDuration := alertRule.HoldDuration()
+
+		alertRule.ForEachActiveAlert(func(a *Alert) {
```
brian-brazil: This is still presuming that the first evaluation was firing, which it might not due to not having enough data yet.

rules/manager.go
```diff
@@ -407,12 +422,133 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
...
+
+			if !seriesFound {
+				return
+			}
+
+			// Series found for the 'for' state.
+			var v float64
+			var t int64
+			it := s.Iterator()
+			for it.Next() {
```
brian-brazil: You should check the iterator for an error too

rules/manager.go
```diff
@@ -407,12 +422,133 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
...
+			if value.IsStaleNaN(v) { // Alert was not active.
+				return
+			}
+
+			// v = the latest ActiveAt of the alert.
+			restoredActiveAt := time.Unix(int64(v), 0)
+			crashTime := time.Unix(t/1000, 0)
+			if crashTime.Sub(restoredActiveAt) >= alertHoldDuration {
+				// The alert was firing when it crashed. Hence we need not wait for grace
+				// period, we can immediatly make it fire.
```
brian-brazil: I think we should still apply the same logic here, and wait the 10m. Just because it was firing an hour ago doesn't mean the condition is still happening.

rules/manager.go
```diff
@@ -407,12 +422,133 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
...
+
+			// restoredActiveAt               ts (current)             firing time = restoredActiveAt.Add(alertHoldDuration)
+			//     |-------------------------------|-------------------------------|
+			//
+			//     ^-----------------------alertHoldDuration-----------------------^
+			//
+			//                                     ^-------------------------------^
+			// 			    					restoredActiveAt.Add(alertHoldDuration).Sub(ts)
+			//
+			// Hence `restoredActiveAt.Add(alertHoldDuration).Sub(ts)` is the time left
```
brian-brazil: I think I see where the disconnect is. If an alert has a for 6h, I think it must spend at least 6h firing. If the Prometheus is down for half an hour, it now needs to spend at least 6h+30m+10m firing. This will reduce false positives due to alerts resolving while Prometheus is down.

rules/manager.go
```diff
@@ -209,6 +216,14 @@ func (g *Group) run(ctx context.Context) {
 	}
 	lastTriggered := time.Now()
 	iter()
+	if g.shouldRestore {
+		// Waiting for all other groups to finish first evalution.
```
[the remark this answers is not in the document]
codesome: Yes it was not required, I have removed it now.

rules/manager.go
```diff
@@ -484,6 +623,7 @@ func (m *Manager) Update(interval time.Duration, files []string) error {
 		}
 		return errors.New("error loading rules, previous rule set restored")
 	}
+	m.restored = true
```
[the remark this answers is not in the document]
codesome: From what I understand, `Update` is called while loading the rule files for the first time too. So we have to set `m.restored` to be true after the first loading right? Or are the rule files loaded somewhere else for the first time?

rules/manager.go
codesome: Okay. In that case we have no use of timestamp of the sample, as we wait for 10 minutes in any case.

rules/manager.go
codesome: First evaluation happens after waiting for eval interval, wont we have enough samples by then? If not then we can restore after 2nd eval, does that sound fine?

rules/manager.go
codesome: With firing do you mean pending time? I did not quite follow what you meant.

brian-brazil: Yes, I meant pending.

rules/manager.go
brian-brazil: After the 2nd is probably best

codesome: Done

rules/manager.go
codesome: So 6h+30m+10m is a problem?

codesome: According to the implementation, it will be pending for another 3h30m before firing, as the total time after it got active would be as `2h + 30m (down) + 3h30m = 6h`.  Or do we want it to be pending for more 4h making total time from when it was active to be `6h30m`?

brian-brazil: The total time there should be 6h30m or 6h40m depending on how we deal with the 10m in that case.

codesome: Okay, now I get why you were behind the timestamp. I will make that change.

codesome: I have make that change. Now in the above case the total time will be 6h30m or 6h40m depending on grace period. The down time wont be considered in pending time.

rules/manager.go
```diff
@@ -407,12 +432,119 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
...
+// RestoreForState restores the 'for' state of the alerts
+// by looking up last ActiveAt from storage.
+func (g *Group) RestoreForState(ts time.Time) {
+	maxtMS := int64(model.TimeFromUnixNano(ts.UnixNano()))
+	// We allow restoration only if alerts were active before some certain time.
+	mint := ts.Add(-g.opts.OutageTolerance)
+	mintMS := int64(model.TimeFromUnixNano(mint.UnixNano()))
+	q, err := g.opts.TSDB.Querier(g.opts.Context, mintMS, maxtMS)
+	if err != nil {
+		level.Error(g.logger).Log("msg", "Failed to get Querier", "err", err)
```
brian-brazil: Should we return here?

rules/manager.go
```diff
@@ -407,12 +432,119 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
...
+			if value.IsStaleNaN(v) { // Alert was not active.
+				return
+			}
+
+			downAt := time.Unix(t/1000, 0)
+			restoredActiveAt := time.Unix(int64(v), 0)
+			timeSpentPending := downAt.Sub(restoredActiveAt)
+			timeRemainingPending := alertHoldDuration - timeSpentPending
+
+			if alertHoldDuration > g.opts.ForGracePeriod && timeRemainingPending < g.opts.ForGracePeriod {
```
brian-brazil: If the alert was already firing, we should keep it firing.

rules/manager.go
```diff
@@ -407,12 +432,119 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
...
+				//                        = (ts + m.opts.ForGracePeriod) - ts
+				//                        = m.opts.ForGracePeriod
+				restoredActiveAt = ts.Add(g.opts.ForGracePeriod).Add(-alertHoldDuration)
+			} else {
+				// By shifting ActiveAt to the future (ActiveAt + some_duration),
+				// the total pending time from the original ActiveAt
+				// would be `alertHoldDuration + some_duration`.
+				// Here, some_duration = downDuration.
+				downDuration := ts.Sub(downAt)
+				restoredActiveAt = restoredActiveAt.Add(downDuration)
```
brian-brazil: The previous predicate `alertHoldDuration > g.opts.ForGracePeriod` means that a for 5m which has 1m left to go would be restored when it shouldn't be at all.

codesome: So if any alert has `alertHoldDuration` less than grace period, we don't restore at all?

brian-brazil: Yes

rules/manager.go
brian-brazil: Yes

rules/manager.go
codesome: Done. Added that check before I call `alertRule.ForEachActiveAlert(...)`.

rules/manager.go
codesome: Done. `if timeRemainingPending <= 0` should take care of it.

rules/manager.go
```diff
@@ -407,12 +432,128 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
...
+
+			downAt := time.Unix(t/1000, 0)
+			restoredActiveAt := time.Unix(int64(v), 0)
+			timeSpentPending := downAt.Sub(restoredActiveAt)
+			timeRemainingPending := alertHoldDuration - timeSpentPending
+
+			if timeRemainingPending <= 0 {
+				// It means that alert was firing when prometheus went down.
+				// In the next Eval, the state of this alert will be set back to
+				// firing again.
```
brian-brazil: ...if it's still firing in that Eval.

brian-brazil:
I've given it a quick test. When an alert stops firing it seems that the two ALERTS series aren't getting marked stale.

The flag names also need changing, as these settings are not AM related.

codesome:
I think they should have been marked stale at least after few Eval if not active, as that part of code was not tinkered. I will give it a test and see what is the issue.

And about the flags:
`alertmanager.outage-tolerance` => `for-outage-tolerance`
`alertmanager.for-grace-period` => `for-grace-period`?

brian-brazil:
The flags still need a prefix, rule.alert... would be one option.

codesome:
This?
`rules.alert.for-outage-tolerance`
`rules.alert.for-grace-period`

brian-brazil:
It's as good as anything.

rules/alerting.go
```diff
@@ -275,7 +305,10 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc,
 			a.FiredAt = ts
 		}
 
-		vec = append(vec, r.sample(a, ts))
+		if r.restored {
```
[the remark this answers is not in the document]
codesome: We only lose the samples of first 2 `Eval` at startup (because it is restored after 2nd Eval), I guess that's not a big harm? 

In my view if the restoration changed the state of the alert, then adding the samples before restoration can be like noise in the graph.

rules/manager.go
```diff
@@ -407,12 +422,133 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
...
+			if err != nil {
+				level.Error(g.logger).Log("msg", "Failed to restore 'for' state",
+					labels.AlertName, alertRule.Name(), "stage", "Select", "err", err)
+				return
+			}
+
+			seriesFound := false
+			var s storage.Series
+			for sset.Next() {
+				if len(sset.At().Labels()) == len(smpl.Metric) {
```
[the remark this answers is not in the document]
codesome: From the query it is assured that labels in `smpl.Metric` are present in the `sset.At()`. So checking the length would indeed assure that we have the exact set of lable-values, nothing less nothing more.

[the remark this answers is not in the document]
gouthamve: `labels.Compare(sset.At().Labels(), smpl.Metric) == 0` is better no?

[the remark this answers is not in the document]
gouthamve: Ah, I see. A comment for the same would be good :) But the above exactness check wouldn't hurt.

rules/alerting.go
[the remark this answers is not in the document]
gouthamve: Ah, I see. One sec. But if Group is created with `shouldRestore=false`, then `restored` is never true. Which means we never insert samples?

If `NewGroup` is to be called with `shouldRestore=true` always, remove that parameter?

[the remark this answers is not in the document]
codesome: From [this](https://github.com/prometheus/prometheus/pull/4061/files#diff-d23c7264c420b842612a4fe05e018217R677), `m.restored=true` when `shouldRestore=false`.
And we pass `m.restored` to the rule [here](https://github.com/prometheus/prometheus/pull/4061/files#diff-d23c7264c420b842612a4fe05e018217R705).
So if `shouldRestore=false`, `rule.restored=true`, hence samples are added.


And no, when groups are re-loaded using API call, `shouldRestore=false`, hence that parameter would be needed.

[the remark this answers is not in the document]
gouthamve: Ah, I see. Got it.

codesome:
@brian-brazil I don't see why the series won't be marked stale when the alert becomes inactive, the original logic is still in place. Did you observe that even 1-2 Evals after it became inactive?

brian-brazil:
Yes, it seemed to be that way even after waiting a bit.

codesome:
@brian-brazil while testing for that I found a bug in another place where samples were not being added if `alertHoldDuration` was less than hold duration (as it was not set as restored). 

The series was being marked stale after that (I did not test it before that fix).

brian-brazil:
That seems to be working okay now. Can you ensure all your commits have a DCO?

codesome:
@brian-brazil There was an error while rebasing to change commit messages. Can this PR be merged without DCO for some commits ([very]old commits)?

brian-brazil:
We require DCO on all commits. You could try squashing.

codesome:
@brian-brazil Many issues while trying to change commit messages. With some methods there is a conflict, and with another method I get
```
The previous cherry-pick is now empty, possibly due to conflict resolution.
```

brian-brazil:
Try squashing everything into one commit. That should bypass all that.

codesome:
@brian-brazil It's done.

seanjfellows:
I haven't used Borgmon in 4 years or used Prometheus in 9 months ( 😭 ) but this is so great ❤️

PR #4064, opened 2018-04-09
Update vendoring of prometheus/common/route to include data race fix

beorn7:
See https://github.com/prometheus/common/pull/125

@krasi-georgiev @brian-brazil


brian-brazil:
:+1: 

You need to add the DCO to your commit.

PR #4071, opened 2018-04-10
support use config file path for kubernetes

ckeyer:
eq.
```
# k8s config file admin.conf 
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: [DATA]
    server: https://10.1.10.8:6443
  name: kubernetes
contexts:
- context:
    cluster: kubernetes
    user: kubernetes-admin
  name: kubernetes-admin@kubernetes
current-context: kubernetes-admin@kubernetes
kind: Config
preferences: {}
users:
- name: kubernetes-admin
  user:
    client-certificate-data: [DATA]
    client-key-data: [DATA]
```


brian-brazil:
We generally want the configuration in the Prometheus configuration file, as this makes it easier for us to debug issues. I'm also unsure of the wisdom of depending on the internals of a tool. I don't think this makes sense for us, we already have ways to configure all of this.

ckeyer:
@brian-brazil  `configfile` is more useful than others in versions later than 1.9 at kubernetes, Please think about it again.

PR #4075, opened 2018-04-10
request a use case for proposals

krasi-georgiev:
@brian-brazil Saw this in https://github.com/influxdata/telegraf/issues/new and thought is a good idea.
 Update to your liking or let me know if you are against and I will close it.


brian-brazil:
I guess we can try it and see. Missing a DCO.

PR #4083, opened 2018-04-13
merge relabel into pkg/relabel

sipian:
Signed-off-by: sipian <cs15btech11019@iith.ac.in>
#3647 
relabel.go is removed by interconverting between model.LabelSet and labels.Labels before sending to Process function in pkg/relabel

### DIFF
--- a/relabel/relabel.go
+++ b/relabel/relabel.go
@@ -28,7 +28,7 @@ import (
 // If a label set is dropped, nil is returned.
 // May return the input labelSet modified.
 // TODO(https://github.com/prometheus/prometheus/issues/3647): Get rid of this package in favor of pkg/relabel
-//  once usage of `model.LabelSet` is removed.
+// once usage of `model.LabelSet` is removed.
 func Process(labels model.LabelSet, cfgs ...*pkgrelabel.Config) model.LabelSet {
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for looking at this. The idea would be to remove usage of model.Labels, would you be willing to change the remaining code that's using it?

krasi-georgiev:
@sipian  and I will work together on this one :) so yes will update all code that it needs changes.

bwplotka:
Heyo, any progress on it? (:

sipian:
Since I'm relatively free now, I plan to return to this.

brian-brazil:
We should be working to remove model, not expand it.

sipian:
Since, `model.Labelset` is used as a type definition for [model.Metric](https://github.com/prometheus/common/blob/master/model/metric.go#L33) and also in the struct of [model.Alert](https://github.com/prometheus/common/blob/master/model/alert.go#L29),
should I move all the components of `common/model` to `pkg` folder.

brian-brazil:
The idea would be more to use pkg/labels rather than model.Metric.

tomwilkie:
I've done the required changes to remove one of the relabelling packages in #5316, so I'm going to close this.

Issue #4067, opened 2018-04-09
Race when reloading the config


krasi-georgiev:
**What did you do?**
 - started prometheus to scrape itself - 
`go run --race main.go  uname_default.go fdlimits_default.go --storage.tsdb.min-block-duration=5m --storage.tsdb.max-block-duration=15m --config.file=../../promBench.yml`
 - wait few seconds until prometheus registeres all targets and starts scraping.
 - in a separate terminal run `watch -n 1 killall -1 main` to keep reloading the config file

```
==================
WARNING: DATA RACE
Read at 0x00c420013838 by goroutine 14:
  github.com/prometheus/prometheus/scrape.(*scrapePool).appender()
      /home/krasi/src/github.com/prometheus/prometheus/scrape/scrape.go:404 +0x262
  github.com/prometheus/prometheus/scrape.(*scrapePool).(github.com/prometheus/prometheus/scrape.appender)-fm()
      /home/krasi/src/github.com/prometheus/prometheus/scrape/scrape.go:171 +0x41
  github.com/prometheus/prometheus/scrape.(*scrapeLoop).report()
      /home/krasi/src/github.com/prometheus/prometheus/scrape/scrape.go:991 +0x207
  github.com/prometheus/prometheus/scrape.(*scrapeLoop).run()
      /home/krasi/src/github.com/prometheus/prometheus/scrape/scrape.go:720 +0xbf7

Previous write at 0x00c420013838 by goroutine 86:
  github.com/prometheus/prometheus/scrape.(*scrapePool).reload()
      /home/krasi/src/github.com/prometheus/prometheus/scrape/scrape.go:214 +0x1b2
  github.com/prometheus/prometheus/scrape.(*Manager).ApplyConfig()
      /home/krasi/src/github.com/prometheus/prometheus/scrape/manager.go:104 +0x392
  github.com/prometheus/prometheus/scrape.(*Manager).ApplyConfig-fm()
      /home/krasi/src/github.com/prometheus/prometheus/cmd/prometheus/main.go:306 +0x4b
  main.reloadConfig()
      /home/krasi/src/github.com/prometheus/prometheus/cmd/prometheus/main.go:606 +0x2eb
  main.main.func14()
      /home/krasi/src/github.com/prometheus/prometheus/cmd/prometheus/main.go:450 +0x57e
  github.com/prometheus/prometheus/vendor/github.com/oklog/oklog/pkg/group.(*Group).Run.func1()
[...]

PR #4084, opened 2018-04-13
Fix scrape races

simonpasquier:
The change avoids passing the full scrape configuration down to the scrape loop which resolves the data races when the scrape configuration is being reloaded.

The PR adds a unit test that fails as described in the initial report from #4067. I've seen the [other data race](https://github.com/prometheus/prometheus/issues/4067#issuecomment-380108030) too but it happened far less often.

cc @krasi-georgiev  @beorn7 @brian-brazil

### DIFF
--- a/scrape/manager_test.go
+++ b/scrape/manager_test.go
@@ -249,7 +249,7 @@ func TestManagerReloadNoChange(t *testing.T) {
 	scrapeManager := NewManager(nil, nil)
 	scrapeManager.scrapeConfigs[tsetName] = reloadCfg.ScrapeConfigs[0]
 	// As reload never happens, new loop should never be called.
-	newLoop := func(_ *Target, s scraper) loop {
+	newLoop := func(_ *Target, s scraper, _ int, _ bool, _ []*config.RelabelConfig) loop {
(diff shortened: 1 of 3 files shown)


krasi-georgiev:
nice! 
can confirm that it fixes the race I found.

@simonpasquier any idea how to replicate the second one?

maybe the problem is that we are passing `[]*config.RelabelConfig` with a pointer , what if we pass a copy `[]config.RelabelConfig` ?

krasi-georgiev:
I see what you mean. I have run this for a while and no races were detected and I don't see any problems in the code changes.
+1 from me.

brian-brazil:
:+1:

krasi-georgiev:
@beorn7 do you want to test it?

beorn7:
I'm on vacations this week. I'll happily try it out next week.

Can this be rebased on top of the release-2.2 branch and then merged into that one first?

simonpasquier:
@beorn7 yep, i'll try this.

beorn7:
I'll close this in lieu of #4096 . (We just merge from the release-2.2 into master once #4096 is merged into release-2.2.)

PR #4085, opened 2018-04-13
promtool: add query instant and query range commands

si74:
Relating to this issue: https://github.com/prometheus/prometheus/issues/1605

Yup! Some context behind this PR: 
- for now, I'm printing the json result of the query request 
- in the future i may add further flags to print a particular format

### DIFF
--- a/cmd/promtool/main.go
+++ b/cmd/promtool/main.go
@@ -28,8 +28,8 @@ import (
 	"gopkg.in/alecthomas/kingpin.v2"
 	"gopkg.in/yaml.v2"
 
-	api "github.com/prometheus/client_golang/api"
-	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"github.com/prometheus/client_golang/api"
+	"github.com/prometheus/client_golang/api/prometheus/v1"
 	config_util "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/common/version"
@@ -350,7 +350,7 @@ func CheckMetrics() int {
 	return 0
 }
 
-// QueryInstant performs an instant query against prometheus server.
+// QueryInstant performs an instant query against a Prometheus server.
 func QueryInstant(url *url.URL, query string) int {
 	config := api.Config{
 		Address: url.String(),
@@ -379,7 +379,7 @@ func QueryInstant(url *url.URL, query string) int {
 	return 0
 }
 
-// QueryRange performs a range query against a prometheus server.
+// QueryRange performs a range query against a Prometheus server.
 func QueryRange(url *url.URL, query string, start string, end string) int {
 	config := api.Config{
 		Address: url.String(),
@@ -419,7 +419,7 @@ func QueryRange(url *url.URL, query string, start string, end string) int {
 
 	resolution := math.Max(math.Floor(etime.Sub(stime).Seconds()/250), 1)
 	// Convert seconds to nanoseconds such that time.Duration parses correctly.
-	step := time.Duration(resolution * math.Pow(10, 9))
+	step := time.Duration(resolution * 1e9)
 
 	// Run query against client.
 	api := v1.NewAPI(c)
--- a/cmd/promtool/main.go
+++ b/cmd/promtool/main.go
@@ -16,6 +16,7 @@ package main
 import (
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -23,6 +24,7 @@ import (
 	"gopkg.in/alecthomas/kingpin.v2"
 	"gopkg.in/yaml.v2"
 
+	api "github.com/prometheus/client_golang/api"
 	config_util "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/common/version"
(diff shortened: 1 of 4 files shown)


cmd/promtool/main.go
```diff
@@ -57,6 +64,17 @@ func main() {
 	updateRulesCmd := updateCmd.Command("rules", "Update rules from the 1.x to 2.x format.")
 	ruleFilesUp := updateRulesCmd.Arg("rule-files", "The rule files to update.").Required().ExistingFiles()
 
+	queryCmd := app.Command("query", "Run query against a Prometheus server.")
+	queryInstantCmd := queryCmd.Command("instant", "Run instant query.")
+	queryServer := queryInstantCmd.Arg("server", "Prometheus server to query.").Required().URL()
+	queryExpr := queryInstantCmd.Arg("expr", "PromQL query expression.").Required().String()
```
juliusv: `queryServer` -> `queryInstantServer`
`queryExpr` -> `queryInstantExpr`

Too bad identical subarguments are apparently not reusable between range and instant queries.

cmd/promtool/main.go
```diff
@@ -326,3 +349,101 @@ func CheckMetrics() int {
 
 	return 0
 }
+
+// QueryInstant performs an instant query against prometheus server.
```
juliusv: "against prometheus server." -> "against a Prometheus server."

cmd/promtool/main.go
```diff
@@ -14,15 +14,22 @@
...
 	"path/filepath"
+	"strconv"
 	"strings"
+	"time"
 
 	"gopkg.in/alecthomas/kingpin.v2"
 	"gopkg.in/yaml.v2"
 
+	api "github.com/prometheus/client_golang/api"
+	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
```
juliusv: Hmm, why are these aliased to the package name they already have anyway?

cmd/promtool/main.go
```diff
@@ -326,3 +349,101 @@ func CheckMetrics() int {
...
+		fmt.Fprintln(os.Stderr, "query error:", err)
+		return 1
+	}
+
+	fmt.Println(val.String())
+
+	return 0
+}
+
+// QueryRange performs a range query against a prometheus server.
```
juliusv: nit: "prometheus" -> "Prometheus" (hey, it's our brand! :P)

cmd/promtool/main.go
```diff
@@ -326,3 +349,101 @@ func CheckMetrics() int {
...
+		}
+	}
+
+	if !stime.Before(etime) {
+		fmt.Fprintln(os.Stderr, "start time is not before end time")
+	}
+
+	resolution := math.Max(math.Floor(etime.Sub(stime).Seconds()/250), 1)
+	// Convert seconds to nanoseconds such that time.Duration parses correctly.
+	step := time.Duration(resolution * math.Pow(10, 9))
```
juliusv: `math.Pow(10, 9)` can just be `1e9`.

Or better, do `time.Duration(resolution * time.Second)` because that keeps it independent of the integer representation of `time.Duration`.

juliusv:
:+1: besides nits.

One thing we might want to think about is whether we want to keep the current pure JSON dump the default output format or implement another default format now first (and do the raw JSON dump only when asked via flag). Because if promtool is versioned along with Prometheus, it might mean that we cannot change the default output format before Prometheus 3.x. For a nicer default output format, I'm thinking of something like our text-based metrics exposition format, but it'd have to support non-vector types as well, and multiple values per series for range queries.

brian-brazil:
Promtool is not listed as part of the stability guarantees.

brian-brazil:
Of course it wouldn't hunt to mention it explicitly on that page somewhere.

juliusv:
Ah wait, I'm stupid. I didn't even try it out. The current format isn't JSON (somehow had that in my mind from reading https://github.com/prometheus/prometheus/issues/1605#issuecomment-333480005), but a nicer stringified representation that the model provides. I guess that's also cool for now.

cmd/promtool/main.go
krasi-georgiev: I think  somehting like  this should  work.

```
queryServer := queryCmd.Arg("server", "Prometheus server to query.").Required().URL()
queryExpr := queryCmd.Arg("expr", "PromQL query expression.").Required().String()

queryInstantCmd := queryCmd.Command("instant", "Run instant query.")

queryRangeCmd := queryCmd.Command("range", "Run range query.")
```

juliusv: Yeah, we tried that earlier, but then kingpin complains with:

```
promtool: error: can't mix Arg()s with Command()s, try --help
```

krasi-georgiev: what about using it as a flag?
`queryServer := queryCmd.Flag...`

juliusv: That might work, but it'd be nice to be able to use it as a positional arg, since it's required anyway...

cmd/promtool/main.go
si74: addressed

cmd/promtool/main.go
si74: addressed

cmd/promtool/main.go
si74: addressed

si74:
@juliusv I addressed the major comments!

krasi-georgiev:
@si74 tests are failing , if you add new packages these need to be vendored.

juliusv:
@si74 :+1: besides missing vendoring.

cmd/promtool/main.go
simonpasquier: Thinking out loud but maybe we could have a single command for both instant and range queries?

```
usage: main query [<flags>] <server> <expr>

Run query against a Prometheus server.

Flags:
  -h, --help         Show context-sensitive help (also try --help-long and --help-man).
      --version      Show application version.
      --start=START  Query range start time (RFC3339 or Unix timestamp).
      --end=END      Query range end time (RFC3339 or Unix timestamp).

Args:
  <server>  Prometheus server to query.
  <expr>    PromQL query expression.
```

When `start` and `end` are given, the command returns a range query otherwise it's an instance query.

juliusv: @simonpasquier But for an instant query you can also provide a timestamp, for which `start` or `end` wouldn't be appropriate naming, so then we'd have yet another flag and there'd be more combinatorial confusion going on. IMO it's better to keep the range/instant queries explicit and separate, as they also hit different API endpoints.

PR #4086, opened 2018-04-15
Rewrote tests with testutil for several test files

elifkus:
Used util/testutil in functions_test, relabel_test, and consul_test. Related to #3242


brian-brazil:
Thanks!

PR #4097, opened 2018-04-18
Fix the k8s discovery flaky test

krasi-georgiev:
Testing if the discovery refactoring https://github.com/prometheus/prometheus/pull/4013 fixes the flaky test.


krasi-georgiev:
after about 50 runs and no failures it seems that the k8s refacotring fixes the flaky test.

PR #4100, opened 2018-04-19
use package shorthand selection that excludes vendored.

krasi-georgiev:


brian-brazil:
:+1:

PR #4106, opened 2018-04-21
Update CircleCI build

SuperQ:
Use CircleCI 2.0 build config.


SuperQ:
Running `make test` seems to fail in circleci due to OOM.  The container has 4G of ram, not sure what's going on here.

.circleci/config.yml
```diff
@@ -0,0 +1,134 @@
...
+  test:
+    docker:
+    - image: circleci/golang:1.10
+    working_directory: /go/src/github.com/prometheus/prometheus
+    resource_class: large
+
+    steps:
+    - checkout
+    - run: make promu
+    - run: make check_license style unused staticcheck build
```
krasi-georgiev: is it worth  contacting the circle support about the OOM when running `make test` ? It would be nice if we didn't need travis anymore.

.circleci/config.yml
```diff
@@ -0,0 +1,134 @@
...
+    - run: rm -v prometheus
```
krasi-georgiev: just curious why do we need this?

.circleci/config.yml
SuperQ: I did, I asked about upgrading to the larger containers, but they say we have to pay for it.  I'm working out if that's necessary for our project.  If we need to, we can get the CNCF to pay for it.

.circleci/config.yml
SuperQ: We don't need to, but it's somewhat a pattern to double check that we actually built a binary.  `rm` will error if there's no file.

.circleci/config.yml
krasi-georgiev: In that case  I think just leave it for now and lets see what happens with the new CI we are working on.

.circleci/config.yml
krasi-georgiev: ah , got it thank!

PR #4107, opened 2018-04-21
Only set CORS headers if they aren't set already

jacksontj:
This allow for middleware to set these to more specific values if desired

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -101,7 +101,9 @@ type response struct {
 // Enables cross-site script calls.
 func setCORS(w http.ResponseWriter) {
 	for h, v := range corsHeaders {
-		w.Header().Set(h, v)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Do we have code that needs this?

jacksontj:
I have a few use-cases that do this, and as of today I have to maintain a fork of the repo to make prometheus not set these. I'm also open to moving this to a configuration option, but this seemed like the simplest solution to the issue.

brian-brazil:
We don't generally keep code that we don't use, as our internal APIs are internal. If there are existing endpoints that don't have the right headers currently we can discuss changing them.

brian-brazil:
If we don't have code that needs this, then it's best not to have this. Less chance of a mixup that causes an unintentional security issue.

aloababa:
What if i don't want '*' for Access-Control-Allow-Origin ?

PR #4109, opened 2018-04-24
API documentation: mention step unit

jotak:
Mention in http api docs that step parameter is in seconds.

@brian-brazil I imagine this is trivial enough to not have to create an issue first, right?

Signed-off-by: Joel Takvorian <jtakvori@redhat.com>

### DIFF
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -132,7 +132,7 @@ URL query parameters:
 - `query=<string>`: Prometheus expression query string.
 - `start=<rfc3339 | unix_timestamp>`: Start timestamp.
 - `end=<rfc3339 | unix_timestamp>`: End timestamp.
-- `step=<duration>`: Query resolution step width.
+- `step=<duration>`: Query resolution step width in seconds.
 - `timeout=<duration>`: Evaluation timeout. Optional. Defaults to and
    is capped by the value of the `-query.timeout` flag.
 


docs/querying/api.md
```diff
@@ -132,7 +132,7 @@ URL query parameters:
 - `query=<string>`: Prometheus expression query string.
 - `start=<rfc3339 | unix_timestamp>`: Start timestamp.
 - `end=<rfc3339 | unix_timestamp>`: End timestamp.
-- `step=<duration>`: Query resolution step width.
+- `step=<duration>`: Query resolution step width in seconds.
```
brian-brazil: Not quite. This is a <duration> as defined over in the configuration docs, but it also accepts a floating point number of seconds. We should copy the definition of <duration> over, and add that this can be a float too.

jotak: Do you mean it's the same as this:
`<duration>: a duration matching the regular expression [0-9]+(ms|[smhdwy])`
?
I assumed it was seconds cause all example I've seen generally don't specify the unit, and if I understand correctly it defaults to seconds then.

jotak: ok - so this PR is not necessary, I'm closing it

brian-brazil: It's not the same as a <duration>, as it also allows a floating point value.

PR #4111, opened 2018-04-25
add ignore_error option for remote read

mtanda:
I expect Prometheus to keep working when remote read endpoint is stopped.
Prometheus is important for alerting, and incident response.
Most of remote read endpoint is used for Long term storage, not important for incident response.
So, I propose to add ignore_error option.

I should mention about another point.
`read_recent` could avoid this problem for long term storage, but I want to use this option with `read_recent = true` case.
I create a read adapter for internal use, it is not stable like Prometheus, and not important too.
This is the main use case of `ignore_error` option for me.

Signed-off-by: Mitsuhiro Tanda mitsuhiro.tanda@gmail.com


brian-brazil:
This should not be an option, see #2573.

mtanda:
@brian-brazil Keep working without option is better.
Thanks for information! 😄 
Close this PR.

PR #4117, opened 2018-04-26
K8s SD refactoring for release-2.2

beorn7:
This is #4013 cherrypicked so that it merges into release-2.2.

If this PR works, we should _not_ merge #4013 but simply merge release-2.2 into master (same code as result, but no duplicated commits).


beorn7:
However, it is also not picking up newly scheduled pods.

And on 2.0 and 1.8, everything works as expected.

krasi-georgiev:
spinning a new minikube locally and will let you know whether the problem is with the discovery manager or the k8s  provider.

krasi-georgiev:
using the configs below seems to work as expected with minikube. 
When scaling up and down the deployment pods the  targets list get updated properly in the prometheus GUI.

```
global:
  scrape_interval:     500ms

scrape_configs:
- job_name: 'kubernetes-pods'
  tls_config:
    insecure_skip_verify: true
  
  kubernetes_sd_configs:
  - api_server: https://192.168.99.101:8443
    role: pod
    namespaces:
      names:
      - default
    tls_config:
      ca_file: /home/krasi/.minikube/ca.crt
      cert_file: /home/krasi/.minikube/client.crt
      key_file: /home/krasi/.minikube/client.key
```

```
apiVersion: apps/v1 
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 2 
  template: 
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.7.9
        ports:
        - containerPort: 80
```

beorn7:
Since the refactoring definitely fixes the memory leak (and got reviewed in #4013), I'll merge this now (and then merge into master, which should supercede #4013). I'll then file a separate issue for the failing target updates (which seems unrelated to the refactoring as it happens with exactly the same symptoms before and after).

beorn7:
https://github.com/prometheus/prometheus/issues/4124 for the target update issue.

jurgenweber:
can a new release be made please, the memory leak is hurting our monitoring.

thanks

krasi-georgiev:
few days and it should be out , just polishing few more bits.

PR #4122, opened 2018-04-30
Pass query hints down into remote read query proto

henridf:
Completes  #2580

This currently doesn't include tests, but I'd gladly add some if there are suggestions on how/where to do this.


prompb/remote.proto
```diff
@@ -35,6 +35,7 @@ message Query {
   int64 start_timestamp_ms = 1;
   int64 end_timestamp_ms = 2;
   repeated prometheus.LabelMatcher matchers = 3;
+  prometheus.ReadParams params = 4;
```
brian-brazil: Hints would be clearer than params I think

prompb/types.proto
```diff
@@ -49,3 +49,8 @@ message LabelMatcher {
   string name  = 2;
   string value = 3;
 }
+
+message ReadParams {
+  int64 step = 1;
```
brian-brazil: You should mention the unit here

prompb/remote.proto
henridf: I modeled this on `SelectParams`, but I agree that Hints is clearer; changed. 
(we should maybe also rename `SelectParams` -> `SelectHints`)

prompb/types.proto
henridf: done

prompb/types.proto
```diff
@@ -49,3 +49,8 @@ message LabelMatcher {
   string name  = 2;
   string value = 3;
 }
+
+message ReadHints {
+  int64 step_ms = 1; // Query step size in milliseconds.
+  string func = 2;   // String representation of surrounding function or aggregation.
```
AlekSi: Could it be any function? Or we should limit it to some small set common for remote storages? Maybe enum will be better?

AlekSi:
I looked into Prometheus code, and it is not clear to me if those hints can be ignored by remote storage. Can they?

prompb/types.proto
brian-brazil: It's only a hint, and it can be any function.

henridf: @AlekSi string vs enum was discussed at #2580, this PR follows that

henridf:
@brian-brazil i addressed your comments, anything further needed to merge this PR?

brian-brazil:
Thanks!

jiacai2050:
@henridf @AlekSi I still have no idea if hints can be ignored by remote storage. Can they?

PR #4127, opened 2018-05-01
feat(writers): Warp10 support

miton18:
Support Warp10 as long term storage


brian-brazil:
This adapter is here mainly as an example and for transition users off the previous in-built support for these systems. We're not accepting any new additions here. I'd suggest maintaining this in your own repository.

PR #4130, opened 2018-05-01
web: add named anchors for each rule group

adamdecaf:
The `/rules` list isn't ordered and often gets shuffled around between page loads, which is annoying because you have to scroll to find the rule group.

This change adds named anchors (`<a href="#foo" name="foo">..</a>`) to the `/rules` list. That allows people to be anchored on that on page refreshes.

screen shot 2018-05-01 at 11 04 39 am

### DIFF
--- a/web/ui/bindata.go
+++ b/web/ui/bindata.go
@@ -111,7 +111,7 @@ func (fi bindataFileInfo) Sys() interface{} {
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


adamdecaf:
I didn't run `make assets` on this PR because I assume that's done at release?

jamtur01:
@adamdecaf You will need to run make assets.

adamdecaf:
@brian-brazil Done!

brian-brazil:
Thanks!

It's probably also worthwhile ensuring the rule groups have a consistent ordering.

PR #4131, opened 2018-05-02
example: Added notes to the K8s SD config examples explaining limitation of annotations relabeling.

Bplotka:
Problem: Annotations appear like the best, perfect solution to specify details about "how to" scrape pod (what port, what endpoint, should we even scrape?). There is one major drawback that is not obvious from the first glance. It does not support multi-container (monitored) pods, which are quite often seen in the wild. (well, that's why k8s schedules pods not containers, right?). The problem is that users take literally these examples that it is official/idiotmatic pattern for Prometheus, even though this is just example how to achieve it. (Example discussion around this: https://github.com/improbable-eng/thanos/pull/316)

What this PR fixes?:
- Adds important note that annotations are just an example and they do not work with multi-container pods

------ 
Not for this PR, but we are left with other question - what is the idiomatic way? I totally agree with @brian-brazil that "we shouldn't be dictating how you orchestrate your systems in examples", but we could at least enumerate what are the options... and it seems there is no perfect option here. Let's discuss that on Prometheus User Group though then.

### DIFF
--- a/documentation/examples/kubernetes-rabbitmq/svc.yml
+++ b/documentation/examples/kubernetes-rabbitmq/svc.yml
@@ -4,9 +4,6 @@ metadata:
   name: rabbitmq
   labels:
     name: rabbitmq
-  annotations:
-    prometheus.io/scrape: "true"
-    prometheus.io/port: "9090"
 spec:
   ports:
   - port: 9090
--- a/documentation/examples/prometheus-kubernetes.yml
+++ b/documentation/examples/prometheus-kubernetes.yml
@@ -128,39 +128,43 @@ scrape_configs:
     target_label: __metrics_path__
     replacement: /api/v1/nodes/${1}/proxy/metrics/cadvisor
 
-# Scrape config for service endpoints.
+# Example scrape config for service endpoints.
 #
 # The relabeling allows the actual service scrape endpoint to be configured
-# via the following annotations:
-#
-# * `prometheus.io/scrape`: Only scrape services that have a value of `true`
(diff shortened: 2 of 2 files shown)


brian-brazil:
I think we should remove these relabel actions, rather than making the existing comments longer and more likely to be ignored.

There is no idomatic way, as with all service discovery questions this is a question of how you as the user orchestrate things generally.

Bplotka:
So just scrape all pods/endpoints by default in this example, no filtering?

brian-brazil:
Yeah, it's simplest I think. Maybe a commented out example of some filtering of something?

Bplotka:
PTAL @brian-brazil, but I feel we can simplify it more. However, would like to show our users what is even possible via relabelling (maybe it's not the correct place for it)

brian-brazil:
I'm always a bit wary of going overboard with this, as it's easy to lose the message in endless examples. I think what's here is tolerable.

brancz:
Thanks for doing this, it makes a lot of sense in my opinion.

As a side note, the Prometheus Operator, basically enables you to do exactly what you suggested in the Thanos PR. You can effectively do configs on a per port basis. I don't think it makes sense to use the Prometheus Operator as a default in the repo as it's too large of a surface for a getting started. Fabian, Max and I have been working on some [sample Thanos manifests for the Prometheus Operator](https://github.com/coreos/prometheus-operator/tree/master/contrib/kube-prometheus/experimental/thanos). Expect to see more in this space :wink: .

This lgtm :+1: (sorry for taking so long to get to this)

Bplotka:
Thanks!

PR #4132, opened 2018-05-02
circleci: clean up build configuration

simonpasquier:
cc @SuperQ I've removed a couple of things that weren't strictly necessary given how jobs are split now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,12 +18,6 @@ jobs:
     machine: true
     working_directory: /home/circleci/.go_workspace/src/github.com/prometheus/prometheus
 
-    environment:
-      DOCKER_IMAGE_NAME: prom/prometheus
-      QUAY_IMAGE_NAME: quay.io/prometheus/prometheus
-      DOCKER_TEST_IMAGE_NAME: quay.io/prometheus/golang-builder:1.10-base
-      REPO_PATH: github.com/prometheus/prometheus
-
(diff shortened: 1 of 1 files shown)


SuperQ:
Yes, thanks for the cleanup.  There were some build steps that got duplicated from the `node_exporter` build where we use a test docker image build.

SuperQ:
@krasi-georgiev The only "test" is to cut a release, it worked just fine in node_exporter.  We can cut a release pretty easily by pushing a tag and setting up a draft release.  It never has to hit public visibility.

simonpasquier:
@SuperQ I'm wondering about this snippet:

https://github.com/prometheus/prometheus/blob/c207920500d1d55c21d7b7dbb27ba213394ce369/.circleci/config.yml#L76-L86

As I read, it only pushes the `$CIRCLE_TAG` tag to the registries, but not the `latest` tag. The previous version of the CircleCI configuration ran `docker push <IMAGE NAME>` which would push all local tags.

simonpasquier:
Link to the equivalent section in the previous CircleCI configuration:

https://github.com/prometheus/prometheus/blob/bc6058c81272a8d938c05e75607371284236aadc/circle.yml#L47-L63

SuperQ:
I think that may have just been a copy/paste error when I duplicated the `node_exporter`'s circle2 config.  Feel free to fix it.

PR #4136, opened 2018-05-03
Bubble up errors to promql from populating iterators

jacksontj:
Right now if a storage layer returns an error from `Select` promql continues on as if there was no error. Worse than the single series being dropped, Inspect will actually stop populating any iterators *after* meaning potentially a lot of data is missing.

Options for fixing this are (1) make Walk/Inspect return errors (so the `inspector` would return an error instead of a bool) or (2) have an error variable outside which the functons set

This PR is option 2, but I'd be more than happy to implement option 1 if that is preferred.

### DIFF
--- a/promql/ast.go
+++ b/promql/ast.go
@@ -237,57 +237,80 @@ type VectorMatching struct {
 
 // Visitor allows visiting a Node and its child nodes. The Visit method is
 // invoked for each node with the path leading to the node provided additionally.
-// If the result visitor w is not nil, Walk visits each of the children
+// If the result visitor w is not nil and no error, Walk visits each of the children
 // of node with the visitor w, followed by a call of w.Visit(nil, nil).
 type Visitor interface {
-	Visit(node Node, path []Node) (w Visitor)
+	Visit(node Node, path []Node) (w Visitor, err error)
 }
 
 // Walk traverses an AST in depth-first order: It starts by calling
 // v.Visit(node, path); node must not be nil. If the visitor w returned by
-// v.Visit(node, path) is not nil, Walk is invoked recursively with visitor
-// w for each of the non-nil children of node, followed by a call of
-// w.Visit(nil).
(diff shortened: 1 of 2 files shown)


brian-brazil:
Yes, this is a bug and we probably want 1. However I've got a #3966  PR reworking all of this code and may have already fixed that so let's wait until that's in first.

jacksontj:
This sounds like something worthy of a bugfix backport, if so should I submit this PR against a different branch (I assume we don't want to backport the entire promql cleanup/optimization/rewrite).

jacksontj:
I also did verify in that PR of yours that this issue it not resolved yet, feel free to pull this into that PR if you prefer (the change is small, so likely they'll rebase nicely together).

brian-brazil:
That's merged, so if you could rebase and DCO I can review.

jacksontj:
@brian-brazil PR updated to solution 1 as discussed above.

promql/ast.go
```diff
@@ -249,45 +249,68 @@ type Visitor interface {
 // w for each of the non-nil children of node, followed by a call of
 // w.Visit(nil).
 // As the tree is descended the path of previous nodes is provided.
-func Walk(v Visitor, node Node, path []Node) {
```
brian-brazil: The comments needs updating

promql/ast.go
```diff
@@ -296,21 +319,23 @@ func Walk(v Visitor, node Node, path []Node) {
...
+	} else {
+		return nil, err
 	}
-	return nil
 }
 
 // Inspect traverses an AST in depth-first order: It starts by calling
 // f(node, path); node must not be nil. If f returns true, Inspect invokes f
 // for all the non-nil children of node, recursively.
-func Inspect(node Node, f func(Node, []Node) bool) {
```
brian-brazil: Comments needs updating

jacksontj:
Comments updated

brian-brazil:
Thanks!

PR #4144, opened 2018-05-06
Url service discovery

iobestar:
@brian-brazil


krasi-georgiev:
there is a discussions with lots of background for file sd vs http  and so far most use cases can be solved with the  file sd.
If your case cannot be solved by file sd please read the discussion and post your use case.
https://groups.google.com/forum/#!searchin/prometheus-developers/good$20enough%7Csort:date/prometheus-developers/JcE5nSbCe4k/Z0ChZaR7AgAJ

There is also an open [PR](https://github.com/prometheus/prometheus/pull/3720) with an example how to implement any SD to a json file which than can be used by a file SD


I don't mind having both http and file sd , but I know @brian-brazil  is against this so closing for now, until there is enough info that the others maintainers and users  want to  support this.

Issue #4098, opened 2018-04-18
Prometheus 2.2.1 /graph page doesn't work with Internet Explorer 11


adamc99:
**What did you do?**
Load the main /graph page with Internet Explorer 11.

**What did you expect to see?**
I expected it to load successfully and allow normal functionality, like with Chrome.

**What did you see instead? Under which circumstances?**
The page partially loads but doesn't include elements like the query text box, execute button and data tables/graphs.  

With the IE developer tools open, the error 
"SCRIPT1003: Expected ':'
index.js, Line: 265, Column:11" 
appears in the console tab.

This seems to refer to the line 
```
    source,
```
in the context of:

```
Prometheus.Graph.prototype.initTypeahead = function(self) {
  const source = queryHistory.isEnabled() ? pageConfig.queryHistMetrics.concat(pageConfig.allMetrics) : pageConfig.allMetrics;
  self.expr.typeahead({
    autoSelect: false,
    source,
    items: "all",
```

This occurs every time, tested from many machines and against multiple server instances (2.2.1 and built from master as of 18th Apr 2018).

As an attempt to investigate the issue, I manually compiled a version (based off of master as of 18th Apr 2018) with the line 
```
    source,
```
 changed to 
```
    source: source,
```
 and this seems to correct the issue: the page loads and basic functionality like querying for data and drawing graphs appear to work in both IE 11 and Chrome.

**Prometheus version:**

prometheus, version 2.2.1 (branch: HEAD, revision: bc6058c81272a8d938c05e75607371284236aadc)
  build user:       root@149e5b3f0829
  build date:       20180314-14:15:45
  go version:       go1.10

PR #4148, opened 2018-05-07
Fix graph page for IE11

bege13mot:
Issue #4098 

Fix to load the main /graph page for Internet Explorer 11


brian-brazil:
You need to `make assets`

bege13mot:
Hi @brian-brazil ,

Thank you for your help. Could you please advise me about execution make assets? Why should we update bindata?

brian-brazil:
The bindata is what Prometheus actually loads, so it's required for any js/css changes.

bege13mot:
Got it. Thank you for your help

brian-brazil:
Thanks!

Issue #4137, opened 2018-05-03
Expose GCE labels during service discovery


woolite64:
## Proposal
Expose GCE [instance labels](https://cloud.google.com/compute/docs/labeling-resources) as \__meta_gce_label\_<name> so that values can be passed to Target Labels.

It's currently possible to filter by GCE labels using Prometheus, but the values aren't available once the host is found. Tags and metadata are exposed ([here](https://github.com/prometheus/prometheus/blob/c207920500d1d55c21d7b7dbb27ba213394ce369/discovery/gce/gce.go#L222-L240)) with __meta prefixes.

brian-brazil:
It sounds like you'll need to update the vendoring. We use govendor for this,

woolite64:
Opened a PR for this. First time using Go, any feedback is appreciated.

I was able to compile, run tests, and confirm that labels were working on my end.

PR #4150, opened 2018-05-08
Expose GCE labels during service discovery

woolite64:
This addresses #4137, cc @brian-brazil 

Updated packages under google.golang.org/api/* to support Instance Labels.


brian-brazil:
Can you update the docs too?

woolite64:
Updated the docs to include the label field under gce_sd_configs.

woolite64:
@brian-brazil Thanks for your help in getting this added.

Do you have plans to release a new version soon? I may run a custom build with this feature in the meantime, but would like to get back on an official release when available.

brian-brazil:
Thanks!

2.3.0 should be soon hopefully.

PR #4162, opened 2018-05-15
pkg/textparse: refactor and add metadata handling

fabxc:
Extends the parser to allow retrieving metadata.
The lexer now yields proper tokens that are fed into a hand-written
parser on top.

@jkohen @brian-brazil 

```
benchmark                                                    old ns/op     new ns/op     delta
BenchmarkParse/no-decode-metric/testdata.txt-8               266           270           +1.50%
BenchmarkParse/decode-metric/testdata.txt-8                  432           427           -1.16%
BenchmarkParse/decode-metric-reuse/testdata.txt-8            355           356           +0.28%
BenchmarkParse/no-decode-metric/testdata.nometa.txt-8        209           219           +4.78%
BenchmarkParse/decode-metric/testdata.nometa.txt-8           372           372           +0.00%
BenchmarkParse/decode-metric-reuse/testdata.nometa.txt-8     300           304           +1.33%
BenchmarkGzip/testdata.txt-8                                 441           447           +1.36%
BenchmarkGzip/testdata.nometa.txt-8                          287           289           +0.70%

benchmark                                                    old allocs     new allocs     delta
BenchmarkParse/no-decode-metric/testdata.txt-8               0              0              +0.00%
BenchmarkParse/decode-metric/testdata.txt-8                  1              1              +0.00%
BenchmarkParse/decode-metric-reuse/testdata.txt-8            1              1              +0.00%
BenchmarkParse/no-decode-metric/testdata.nometa.txt-8        0              0              +0.00%
BenchmarkParse/decode-metric/testdata.nometa.txt-8           1              1              +0.00%
BenchmarkParse/decode-metric-reuse/testdata.nometa.txt-8     1              1              +0.00%
BenchmarkGzip/testdata.txt-8                                 0              0              +0.00%
BenchmarkGzip/testdata.nometa.txt-8                          0              0              +0.00%

benchmark                                                    old bytes     new bytes     delta
BenchmarkParse/no-decode-metric/testdata.txt-8               0             0             +0.00%
BenchmarkParse/decode-metric/testdata.txt-8                  128           127           -0.78%
BenchmarkParse/decode-metric-reuse/testdata.txt-8            48            47            -2.08%
BenchmarkParse/no-decode-metric/testdata.nometa.txt-8        0             0             +0.00%
BenchmarkParse/decode-metric/testdata.nometa.txt-8           128           127           -0.78%
BenchmarkParse/decode-metric-reuse/testdata.nometa.txt-8     48            47            -2.08%
BenchmarkGzip/testdata.txt-8                                 416           416           +0.00%
BenchmarkGzip/testdata.nometa.txt-8                          256           256           +0.00%
```

Ultimately performance remains the same as before.


fabxc:
@brian-brazil can I have a review?

jkohen:
Looks great, thanks for doing this.

pkg/textparse/lex.l
```diff
@@ -65,65 +54,31 @@ M     [a-zA-Z_:]
...
-                                      if err != nil {
-                                          l.err = err
-                                          return -1
-                                      }
-                                      l.ts = &ts
-<lstateTimestamp>[\r\n]+              l.nextMstart = l.i
-                                      return 1
-<lstateTimestamp>\0                   return 1
+\0                                    return tEOF
+\r?\n                                 l.state = sInit; return tLinebreak
```
brian-brazil: \n is the only valid linebreak

pkg/textparse/lex.l
```diff
@@ -65,65 +54,31 @@ M     [a-zA-Z_:]
...
+<*>[ \t]+                             return tWhitespace
+
+#\ HELP\                              l.state = sMeta1; return tHelp
```
brian-brazil: Any whitespace is allowed here

pkg/textparse/lex.l
```diff
@@ -65,65 +54,31 @@ M     [a-zA-Z_:]
...
+#\ TYPE\                              l.state = sMeta1; return tType
+#.?.?.?.?.?.?                         // Hack to make comments work that start with a prefix
+                                      // of the above two rules. The generator does not
+                                      // seem to do the right thing here.
+                                      l.state = sComment
+<sComment>[^\r\n]*                    l.state = sInit; return tComment
+<sMeta1>{M}({M}|{D})*                 l.state = sMeta2; return tMName
+<sMeta2>[^\n\r]+                      l.state = sInit; return tText
```
brian-brazil: Ditto, only \n is a line break

pkg/textparse/lex.l
```diff
@@ -65,65 +54,31 @@ M     [a-zA-Z_:]
...
+{M}({M}|{D})*                         l.state = sValue; return tMName
+<sValue>\{                            l.state = sLabels; return tBraceOpen
+<sLabels>{L}({L}|{D})*                return tLName
+<sLabels>\}                           l.state = sValue; return tBraceClose
+<sLabels>=                            l.state = sLValue; return tEqual
+<sLabels>,                            return tComma
+<sLValue>\"(\\.|[^\\"])*\"            l.state = sLabels; return tLValue
+<sValue>[^{ \t\n\r]+                  l.state = sTimestamp; return tValue
+<sTimestamp>{D}+                      return tTimestamp
+<sTimestamp>\r?\n                     l.state = sInit; return tLinebreak
```
brian-brazil: and here

pkg/textparse/lex.l
brian-brazil: How are you handling escaping here?

pkg/textparse/parse.go
```diff
@@ -67,77 +133,255 @@ func (l *lexer) Error(es string) {
...
+			case "counter":
+				p.mtype = MetricTypeCounter
+			case "gauge":
+				p.mtype = MetricTypeGauge
+			case "histogram":
+				p.mtype = MetricTypeHistogram
+			case "summary":
+				p.mtype = MetricTypeSummary
+			default:
+				return EntryInvalid, fmt.Errorf("invalid metric type %q", s)
```
brian-brazil: untyped is a valid type

pkg/textparse/lex.l
fabxc: I can get rid of that but this is not something I added in this PR. So we might see people file issues after changing it. Though unlikely probably.

pkg/textparse/lex.l
fabxc: Yea, then none of this will work at all fwiw – see the contraption below to distinguish comments. I don't think anyone uses anything else in practice – and if it will just be treated as a regular comment. The consequences don't seem too bad.

On another note – why did we think allowing so much arbitrariness everywhere makes sense?

pkg/textparse/lex.l
fabxc: Escaping of what?

pkg/textparse/parse.go
fabxc: Yes, added in #4183.

pkg/textparse/lex.l
brian-brazil: It is a bug, and you may aswell fix it while you're chaning all of this code.

pkg/textparse/lex.l
brian-brazil: The allowance of arbitrary whitespace is based on @beorn7's argument that it makes it easier to generate the format.

pkg/textparse/lex.l
brian-brazil: The value of HELP can contain `\n` meaning a literal newline, and `\\` meaning a single backslash.

pkg/textparse/lex.l
beorn7: https://github.com/RichiH/OpenMetrics/blob/master/protobuf_vs_text.md#original-role-of-the-protobuf-and-text-format

brian-brazil: You could parse this token by token rather than trying to do it all in one go

fabxc: Yea, tokenzining every possible thing in a free-form help text and funneling it through the parser will probably have performance implications.

brian-brazil: I meant the # then the HELP/TYPE, then slurp in the line.

fabxc:
Addressed all those and added tests. Had to resort to another little work around to not make comment handling overly complex. But test coverage is pretty good so that should be fine.

pkg/textparse/lex.l
```diff
@@ -65,65 +55,46 @@ M     [a-zA-Z_:]
...
+    if l.state == sComment {
+        return l.consumeComment()
+    }
+    return tInvalid
+}
+
+func (l *lexer) consumeComment() token {
+    for c := l.cur(); ; c = l.next() {
+        switch c {
+        case 0:
```
brian-brazil: The zero byte is a valid UTF-8 character

pkg/textparse/parse.go
```diff
@@ -67,77 +137,259 @@ func (l *lexer) Error(es string) {
...
+		}
+		p.offsets = append(p.offsets, p.l.start, p.l.i)
+
+		if t := p.nextToken(); t != tEqual {
+			return parseError("expected equal", t)
+		}
+		if t := p.nextToken(); t != tLValue {
+			return parseError("expected label value", t)
+		}
+		if !utf8.Valid(p.l.buf()) {
```
brian-brazil: We should check this for HELP too to be on the safe side

pkg/textparse/lex.l
fabxc: Fine, will add another hack to make that work.
Is this madness also carried into OpenMetrics' text format?

brian-brazil: Yes, nuls have always been valid UTF-8.

fabxc: To clarify the problem. EOF matching is not supported: https://github.com/cznic/lex/blob/master/lex.go#L74

The zero byte is used for that. Thus any matching on it is practically not possible unless we pick another `uint8` as the EOF signal.

brian-brazil: Byte 255 is not possible in UTF-8.

fabxc: Yes, but that doesn't help since we then couldn't distinguish between an invalid string and a valid EOF :) We've to pick one and work around it.

fabxc:
What's the deal with this Travis flake? I've seen it many times the last ~2 weeks:

```
>> running staticcheck
/home/travis/gopath/bin/staticcheck -ignore "github.com/prometheus/prometheus/discovery/kubernetes/kubernetes.go:SA1019 github.com/prometheus/prometheus/discovery/kubernetes/node.go:SA1019 github.com/prometheus/prometheus/documentation/examples/remote_storage/remote_storage_adapter/main.go:SA1019 github.com/prometheus/prometheus/pkg/textparse/lex.l.go:SA4006 github.com/prometheus/prometheus/pkg/pool/pool.go:SA6002 github.com/prometheus/prometheus/promql/engine.go:SA6002" ./...
Killed
make: *** [staticcheck] Error 137
```

fabxc:
It seems to always happen in the staticcheck part. Since we are running everything sequentially, the race detector in the previous steps shouldn't matter. Maybe staticcheck just gets bloated with all our dependencies.

pkg/textparse/lex.l
brian-brazil: Best to parse all valid input then. We'll gain an EOF marker with OpenMetrics that'll cover that case.

pkg/textparse/parse.go
```diff
@@ -124,7 +123,7 @@ func (l *lexer) next() byte {
 	}
 	// Lex struggles with null bytes. If we are in a label value, where
 	// they are allowed, consume them here immediately.
```
brian-brazil: or help string

fabxc:
Merging this into master now, `release-2.3` branch was created.

Issue #4048, opened 2018-04-05
prometheus 2.2.1 alert doesn't work properly


hbceylan:
**What did you do?**
I have created an alert rule for InstanceDown name like this

```
alert: InstanceDown
expr: up == 0
for: 1m
annotations:
  summary: '{{ $labels.node }} is down for more than 1 minutes!!!'
```

**What did you expect to see?**

I want to see works fine normal instance and marathon apps

**What did you see instead? Under which circumstances?**

This rule works fine for normal instance but doesn't work for marathon app.

**Example:** The marathon app service-a working on service-a:192.168.1.1:12345 and when the app restart or down we have received InstanceDown alert, It's OK.

But when the app again up and running different IP address and port like service-a:192.168.2.1:6789 still alert is down.

[https://i.hizliresim.com/XPr7Ej.jpg](url)

**Environment**
Prometheus version: 2.2.1 
Alert Manager: 0.14.0

* System information:

	insert output of `uname -srm` here

* Alertmanager version:

0.14.0

* Prometheus version:

2.2.1

* Alertmanager configuration file:
```
insert configuration here
```

* Prometheus configuration file:
```
- job_name: marathon
  scrape_interval: 15s
  scrape_timeout: 10s
  metrics_path: /metrics
  scheme: http
  marathon_sd_configs:
  - servers:
    - http://marathon.url
    timeout: 30s
    refresh_interval: 30s
  relabel_configs:
  - source_labels: [__meta_marathon_app]
    separator: ;
    regex: ^/(.+)
    target_label: app
    replacement: $1
    action: replace
  - source_labels: [app]
    separator: ;
    regex: (service-a|service-b)
    replacement: $1
    action: keep
  - source_labels: [app]
    separator: ;
    regex: (service-a|service-b)
    target_label: __metrics_path__
    replacement: /prometheus
    action: replace```

* Logs:
```
insert Prometheus and Alertmanager logs relevant to the issue here
```

hbceylan:
@brian-brazil I have analyzed and updated the issue. I think this is a bug. Could you please re-open issue and could you please check this error?

Thanks,

brian-brazil:
I see no indication here yet of a bug in Prometheus. What evidence do you have that the issue is with Prometheus rather than your setup?

hbceylan:
I wrote my prometheus config for marathon_sd_config and prometheus error log in the issue. Did you see any syntax error or anything? If you see an error please let me, I'll appricate this. 

[...]

PR #4171, opened 2018-05-17
discovery/marathon: include url in fetchApps error

adamdecaf:
This was previously part of a larger PR, but that was closed.

See: https://github.com/prometheus/prometheus/issues/4048#issuecomment-389899997

This change could include auth information in the URL. That's been fixed in upstream go, but not until Go 1.11. See: https://github.com/golang/go/issues/24572

### DIFF
--- a/discovery/marathon/marathon.go
+++ b/discovery/marathon/marathon.go
@@ -344,7 +344,11 @@ func fetchApps(client *http.Client, url string) (*AppList, error) {
 		return nil, err
 	}
 
-	return parseAppJSON(body)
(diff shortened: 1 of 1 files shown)


brian-brazil:
I'm a bit confused, what does this have to do with that issue?

adamdecaf:
This change is more related to that comment. Not the basic auth support. It'd be helpful to know which marathon instance prometheus can't scrape in the error returned.

brian-brazil:
Thanks!

If someone has ignored the docs and put a password here it'll already be exposed on the configuration page.

Issue #4173, opened 2018-05-17
Regression: `basic_auth` should not require a `username`


juliusv:
After https://github.com/prometheus/prometheus/pull/4077, `basic_auth` in HTTP configs now requires setting a username, otherwise you get an error:

```
[...] err="Error loading config couldn't load configuration (--config.file=prometheus.yml): parsing YAML file prometheus.yml: basic_auth requires a username"
```

Using basic auth without a username has been standard practice, for example in the suggested `remote_write` configs for Cortex. So this is a breaking change and needs to be fixed.

adamdecaf:
Whoops, https://github.com/prometheus/common/pull/135 fixes that.

juliusv:
@adamdecaf Awesome, thanks :)

adamdecaf:
https://github.com/prometheus/prometheus/pull/4174 Should fix it in promtheus itself.

PR #4174, opened 2018-05-18
update vendor'd github.com/prometheus/common/config

adamdecaf:
Updating from https://github.com/prometheus/common/pull/135 

Via: 

```
$ govendor update github.com/prometheus/common/config
```

Fixes: https://github.com/prometheus/prometheus/issues/4173

### DIFF
--- a/vendor/github.com/prometheus/common/config/config.go
+++ b/vendor/github.com/prometheus/common/config/config.go
@@ -11,23 +11,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// This package no longer handles safe yaml parsing. In order to
+// ensure correct yaml unmarshalling, use "yaml.UnmarshalStrict()".
+
 package config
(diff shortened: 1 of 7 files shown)


krasi-georgiev:
shouldn't this be done using `fetch` instead of `update`  to get the latest from the repo rather than the local copy?
or maybe even include all subpackages 
```
govendor fetch github.com/prometheus/common/...
```

krasi-georgiev:
I always forget the correct commands so I added a simple wiki page with the examples. Feel free to update if you see anything wrong.
https://github.com/prometheus/prometheus/wiki/Updating-a-vendored-package

adamdecaf:
@krasi-georgiev Good call. I get the various tool commands confused. I ran your `fetch` command and pushed.

krasi-georgiev:
Thanks LGTM :+1:

PR #4182, opened 2018-05-22
Fix CircleCI config

SuperQ:
Re-add workspace attachment removed by #4132.


SuperQ:
/cc @simonpasquier

It looks like we dropped the `attach_workspace` in the release workflow.  We need the binaries produced by the `build` step for `promu crossbuild tarballs` to work.

simonpasquier:
:+1: thanks for fixing it!

SuperQ:
Ugh, our `make test` is getting too big for our CI providers.

PR #4183, opened 2018-05-22
scrape,api: provide per-target metric metadata

fabxc:
This adds a per-target cache of scraped metadata. The metadata is only
available for the lifecycle of the attached target. An API endpoint allows
to select metadata by metric name and a label selection of targets.

Per my (so far short-running test) I somewhat surprisingly cannot detect any difference in memory or CPU utilization.

I'll throw in an API-level test once we have agreement on it. It's kinda tedious due to some tech debt.

API example:

```bash
curl -G http://localhost:9091/api/v1/targets/metadata \
    --data-urlencode 'match=go_goroutines{job="prometheus"}' \
    --data-urlencode 'limit=3' | jq
{
  "status": "success",
  "data": [
    {
      "target": {
        "instance": "127.0.0.1:9090",
        "job": "prometheus"
      },
      "type": "gauge",
      "help": "Number of goroutines that currently exist."
    },
    {
      "target": {
        "instance": "127.0.0.1:9091",
        "job": "prometheus"
      },
      "type": "gauge",
      "help": "Number of goroutines that currently exist."
    },
    {
      "target": {
        "instance": "localhost:9090",
        "job": "prometheus"
      },
      "type": "gauge",
      "help": "Number of goroutines that currently exist."
    }
  ]
}
```

@brian-brazil @jkohen

### DIFF
--- a/pkg/textparse/parse.go
+++ b/pkg/textparse/parse.go
@@ -237,6 +237,7 @@ const (
 type MetricType string
 
 const (
+	MetricTypeUnknown   = ""
 	MetricTypeCounter   = "counter"
 	MetricTypeGauge     = "gauge"
 	MetricTypeHistogram = "histogram"
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -161,6 +161,10 @@ func newScrapePool(cfg *config.ScrapeConfig, app Appendable, logger log.Logger) 
 		logger:     logger,
 	}
 	sp.newLoop = func(t *Target, s scraper, limit int, honor bool, mrc []*config.RelabelConfig) loop {
+		// Update the targets retrieval function for metadata to a new scrape cache.
+		cache := newScrapeCache()
+		t.setMetadataFunc(cache.getMetadata)
+
 		return newScrapeLoop(
 			ctx,
 			s,
@@ -175,6 +179,7 @@ func newScrapePool(cfg *config.ScrapeConfig, app Appendable, logger log.Logger) 
 				}
 				return appender(app, limit)
 			},
+			cache,
 		)
 	}
 
@@ -523,43 +528,61 @@ type scrapeCache struct {
 
 	// Parsed string to an entry with information about the actual label set
 	// and its storage reference.
-	entries map[string]*cacheEntry
+	series map[string]*cacheEntry
 
 	// Cache of dropped metric strings and their iteration. The iteration must
 	// be a pointer so we can update it without setting a new entry with an unsafe
 	// string in addDropped().
-	dropped map[string]*uint64
+	droppedSeries map[string]*uint64
 
 	// seriesCur and seriesPrev store the labels of series that were seen
 	// in the current and previous scrape.
 	// We hold two maps and swap them out to save allocations.
 	seriesCur  map[uint64]labels.Labels
 	seriesPrev map[uint64]labels.Labels
+
+	metaMtx  sync.Mutex
+	metadata map[string]*metaEntry
+}
+
+// metaEntry holds meta information about a metric.
+type metaEntry struct {
+	lastIter uint64
+	typ      textparse.MetricType
+	help     string
 }
 
 func newScrapeCache() *scrapeCache {
 	return &scrapeCache{
-		entries:    map[string]*cacheEntry{},
-		dropped:    map[string]*uint64{},
-		seriesCur:  map[uint64]labels.Labels{},
-		seriesPrev: map[uint64]labels.Labels{},
+		series:        map[string]*cacheEntry{},
+		droppedSeries: map[string]*uint64{},
+		seriesCur:     map[uint64]labels.Labels{},
+		seriesPrev:    map[uint64]labels.Labels{},
+		metadata:      map[string]*metaEntry{},
 	}
 }
 
 func (c *scrapeCache) iterDone() {
-	// refCache and lsetCache may grow over time through series churn
+	// All caches may grow over time through series churn
 	// or multiple string representations of the same metric. Clean up entries
 	// that haven't appeared in the last scrape.
-	for s, e := range c.entries {
+	for s, e := range c.series {
 		if c.iter-e.lastIter > 2 {
-			delete(c.entries, s)
+			delete(c.series, s)
 		}
 	}
-	for s, iter := range c.dropped {
+	for s, iter := range c.droppedSeries {
 		if c.iter-*iter > 2 {
-			delete(c.dropped, s)
+			delete(c.droppedSeries, s)
 		}
 	}
+	c.metaMtx.Lock()
+	for m, e := range c.metadata {
+		if c.iter-e.lastIter > 10 {
+			delete(c.metadata, m)
+		}
+	}
+	c.metaMtx.Unlock()
 
 	// Swap current and previous series.
 	c.seriesPrev, c.seriesCur = c.seriesCur, c.seriesPrev
@@ -573,7 +596,7 @@ func (c *scrapeCache) iterDone() {
 }
 
 func (c *scrapeCache) get(met string) (*cacheEntry, bool) {
-	e, ok := c.entries[met]
+	e, ok := c.series[met]
 	if !ok {
 		return nil, false
 	}
@@ -585,16 +608,16 @@ func (c *scrapeCache) addRef(met string, ref uint64, lset labels.Labels, hash ui
 	if ref == 0 {
 		return
 	}
-	c.entries[met] = &cacheEntry{ref: ref, lastIter: c.iter, lset: lset, hash: hash}
+	c.series[met] = &cacheEntry{ref: ref, lastIter: c.iter, lset: lset, hash: hash}
 }
 
 func (c *scrapeCache) addDropped(met string) {
 	iter := c.iter
-	c.dropped[met] = &iter
+	c.droppedSeries[met] = &iter
 }
 
 func (c *scrapeCache) getDropped(met string) bool {
-	iterp, ok := c.dropped[met]
+	iterp, ok := c.droppedSeries[met]
 	if ok {
 		*iterp = c.iter
 	}
@@ -615,6 +638,47 @@ func (c *scrapeCache) forEachStale(f func(labels.Labels) bool) {
 	}
 }
 
+func (c *scrapeCache) setType(metric []byte, t textparse.MetricType) {
+	c.metaMtx.Lock()
+
+	e, ok := c.metadata[yoloString(metric)]
+	if !ok {
+		e = &metaEntry{typ: textparse.MetricTypeUnknown}
+		c.metadata[string(metric)] = e
+	}
+	e.typ = t
+	e.lastIter = c.iter
+
+	c.metaMtx.Unlock()
+}
+
+func (c *scrapeCache) setHelp(metric, help []byte) {
+	c.metaMtx.Lock()
+
+	e, ok := c.metadata[yoloString(metric)]
+	if !ok {
+		e = &metaEntry{typ: textparse.MetricTypeUnknown}
+		c.metadata[string(metric)] = e
+	}
+	if e.help != yoloString(help) {
+		e.help = string(help)
+	}
+	e.lastIter = c.iter
+
+	c.metaMtx.Unlock()
+}
+
+func (c *scrapeCache) getMetadata(metric string) (textparse.MetricType, string, bool) {
+	c.metaMtx.Lock()
+	defer c.metaMtx.Unlock()
+
+	m, ok := c.metadata[metric]
+	if !ok {
+		return textparse.MetricTypeUnknown, "", false
+	}
+	return m.typ, m.help, true
+}
+
 func newScrapeLoop(ctx context.Context,
 	sc scraper,
 	l log.Logger,
@@ -622,6 +686,7 @@ func newScrapeLoop(ctx context.Context,
 	sampleMutator labelsMutator,
 	reportSampleMutator labelsMutator,
 	appender func() storage.Appender,
+	cache *scrapeCache,
 ) *scrapeLoop {
 	if l == nil {
 		l = log.NewNopLogger()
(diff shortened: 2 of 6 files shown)


scrape/scrape.go
```diff
@@ -523,43 +528,61 @@ type scrapeCache struct {
...
-	dropped map[string]*uint64
+	droppedSeries map[string]*uint64
 
 	// seriesCur and seriesPrev store the labels of series that were seen
 	// in the current and previous scrape.
 	// We hold two maps and swap them out to save allocations.
 	seriesCur  map[uint64]labels.Labels
 	seriesPrev map[uint64]labels.Labels
+
+	metaMtx  sync.Mutex
```
jkohen: An RWMutex could prove beneficial here.

scrape/target.go
```diff
@@ -72,6 +74,23 @@ func (t *Target) String() string {
...
 
+// Metadata returns type and help metadata for the given metric.
+func (t *Target) Metadata(metric string) (textparse.MetricType, string, bool) {
+	t.mtx.Lock()
+	defer t.mtx.Unlock()
+
+	if t.metadataFunc == nil {
+		return textparse.MetricTypeUnknown, "", false
+	}
+	return t.metadataFunc(metric)
```
jkohen: You probably want to either unlock before calling this function or wrap with RLock(), or concurrency for this API will effectively be limited to 1.

If you unlock, consider switching the RWMutex to a regular Mutex or an atomic pointer, which are lighter options for this use.

jkohen: Never mind, the concurrency limit is for the target, not the API, so probably not important. I think it's still cleaner to either call RLock or switch to Mutex.

scrape/scrape.go
```diff
@@ -523,43 +528,61 @@ type scrapeCache struct {
...
-	for s, iter := range c.dropped {
+	for s, iter := range c.droppedSeries {
 		if c.iter-*iter > 2 {
-			delete(c.dropped, s)
+			delete(c.droppedSeries, s)
+		}
+	}
+	c.metaMtx.Lock()
+	for m, e := range c.metadata {
+		if c.iter-e.lastIter > 10 {
```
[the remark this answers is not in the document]
fabxc: I want a reasonable amount of buffer time where we can still catch metadata after a metric has officially disappeared. With normal scrape intervals, 10 will give us 3 to 10 minutes.
I'd do even more personally. But I think @brian-brazil advocated for being rather strict about it.

scrape/scrape.go
```diff
@@ -615,24 +638,69 @@ func (c *scrapeCache) forEachStale(f func(labels.Labels) bool) {
...
+	}
+	if e.help != yoloString(help) {
+		e.help = string(help)
+	}
+	e.lastIter = c.iter
+
+	c.metaMtx.Unlock()
+}
+
+func (c *scrapeCache) getMetadata(metric string) (textparse.MetricType, string, bool) {
```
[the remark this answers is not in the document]
fabxc: We'll probably need that anyway for a list API. So that'll likely happen.

scrape/scrape.go
fabxc: The hold time is very short everywhere. Based on past experience in those scenarios, the RWMutex performs notably worse in those cases. Especially since >99% of the time we will need the write lock.

scrape/target.go
fabxc: Yes, that makes sense.

web/api/v1/api.go
```diff
@@ -482,6 +484,63 @@ func (api *API) targets(r *http.Request) (interface{}, *apiError) {
 	return res, nil
 }
 
+func (api *API) targetMetadata(r *http.Request) (interface{}, *apiError) {
+	limit := -1
+	if s := r.FormValue("limit"); s != "" {
+		var err error
+		if limit, err = strconv.Atoi(s); err != nil {
+			return nil, &apiError{errorBadData, fmt.Errorf("limit must be a  number")}
```
brian-brazil: double space before number

pkg/textparse/parse.go
```diff
@@ -237,6 +237,7 @@ const (
 type MetricType string
 
 const (
+	MetricTypeUnknown   = ""
```
brian-brazil: This should have untyped as a value (or unknown if you want to future-proof for openmetrics).

scrape/scrape.go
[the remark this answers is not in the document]
brian-brazil: I was talking mostly about targets going away. Metrics should ~never go away once they are exposed on a target, so this isn't buying you anything.

web/api/v1/api.go
```diff
@@ -482,6 +484,63 @@ func (api *API) targets(r *http.Request) (interface{}, *apiError) {
...
+	var metric string
+	for i, m := range matchers {
+		if m.Name == labels.MetricName && m.Type == labels.MatchEqual {
+			metric = m.Value
+			matchers = append(matchers[:i], matchers[i+1:]...)
+			break
+		}
+	}
+	if metric == "" {
+		return nil, &apiError{errorBadData, errors.New("matcher must specify metric name")}
```
brian-brazil: I don't think this should be required, you should be able to dump all metadata for one target.

web/api/v1/api.go
```diff
@@ -482,6 +484,63 @@ func (api *API) targets(r *http.Request) (interface{}, *apiError) {
...
+		}
+	}
+
+	matchers, err := promql.ParseMetricSelector(r.FormValue("match"))
+	if err != nil {
+		return nil, &apiError{errorBadData, err}
+	}
+
+	var metric string
+	for i, m := range matchers {
```
brian-brazil: A comment that this is extracting/removing the metric name would be useful

web/api/v1/api.go
```diff
@@ -482,6 +484,63 @@ func (api *API) targets(r *http.Request) (interface{}, *apiError) {
...
+		return nil, &apiError{errorBadData, errors.New("matcher must specify metric name")}
+	}
+
+	var res []metricMetadata
+Outer:
+	for _, t := range api.targetRetriever.TargetsActive() {
+		if limit >= 0 && len(res) >= limit {
+			break
+		}
+		for _, m := range matchers {
```
brian-brazil: A comment that this is filtering targets that don't match would be useful

pkg/textparse/parse.go
fabxc: Yes, missed untyped intitially but have this added locally as yet another one. Untyped is known and unknown. I wanted `MetricTypeUnknown` for unknown unknown.

web/api/v1/api.go
fabxc: Yes, was just about to add that as a "list" implementation that's still per-target.
After some consideration, a global list will generate massive amounts of data and seems to dangerous.

brian-brazil: Yea, this requires some thought but it'd still be very useful to get an (arbitrarily) deduped list.

scrape/scrape.go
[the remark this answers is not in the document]
fabxc: I misunderstood you then. My initial version did not have any garbage collection but I added this based on your comment.
Now that we have it, it probably makes sense to just keep it. Metrics should never disappear, but this is probably good to have to be resilient against clients doing something sketchy.

pkg/textparse/parse.go
brian-brazil: Within the current format untyped is that, there is no "unknown".

scrape/scrape.go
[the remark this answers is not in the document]
brian-brazil: It's inconsistent with the the others, which is a bit weird. Needs a comment at the very least.

scrape/scrape.go
```diff
@@ -523,43 +528,61 @@ type scrapeCache struct {
...
+		droppedSeries: map[string]*uint64{},
+		seriesCur:     map[uint64]labels.Labels{},
+		seriesPrev:    map[uint64]labels.Labels{},
+		metadata:      map[string]*metaEntry{},
 	}
 }
 
 func (c *scrapeCache) iterDone() {
-	// refCache and lsetCache may grow over time through series churn
+	// All caches may grow over time through series churn
```
brian-brazil: I thought we'd fixed that?

pkg/textparse/parse.go
fabxc: Fair, fine with using a boolean in places where that doesn't quite cut it.

web/api/v1/api.go
fabxc: I'd opt for providing the metadata first and waiting for feature requests. Better than providing something now with guessed semantics and risky performance implications.

pkg/textparse/parse.go
brian-brazil: What cases are you thinking of?

scrape/scrape.go
fabxc: I merely changed the comment since it was a bit outdated. This just points out why we are doing the garbage collection below.

pkg/textparse/parse.go
fabxc: With the new changes it's no longer relevant I suppose.

krasi-georgiev:
just curious what is the use case for this metadata?

docs/querying/api.md
```diff
@@ -363,6 +363,85 @@ $ curl http://localhost:9090/api/v1/targets
...
+
+The following endpoint returns metadata about metrics currently scraped by targets.
+
+```
+GET /api/v1/targets/metadata
+```
+
+URL query parameters:
+
+- `match=<series_selector>`: A selector that matches targets by label matchers. If a metric name is provided, only metadata for that metric name is returned.
```
brian-brazil: You should be clear about what the label matchers are matching against

I'm wondering now if we should split the metric name from the target labels URL parameter wise, it's a bit unclean to have them as one

docs/querying/api.md
```diff
@@ -363,6 +363,85 @@ $ curl http://localhost:9090/api/v1/targets
 }
 ```
 
+## Target metadata
+
+The following endpoint returns metadata about metrics currently scraped by targets.
```
brian-brazil: This should be marked experimental for now

docs/querying/api.md
```diff
@@ -363,6 +363,85 @@ $ curl http://localhost:9090/api/v1/targets
...
+        "job": "prometheus"
+      },
+      "type": "gauge",
+      "help": "Number of goroutines that currently exist."
+    }
+  ]
+}
+```
+
+The following example returns metadata for all metrics for the target with
```
brian-brazil: for all the targets

scrape/scrape.go
```diff
@@ -615,24 +639,87 @@ func (c *scrapeCache) forEachStale(f func(labels.Labels) bool) {
...
+		Help:   m.help,
+	}, true
+}
+
+func (c *scrapeCache) listMetadata() (res []MetricMetadata) {
+	c.metaMtx.Lock()
+	defer c.metaMtx.Unlock()
+
+	for m, e := range c.metadata {
+		res = append(res, MetricMetadata{
```
brian-brazil: Can we pre-initilise this to be the right size?

scrape/scrape.go
```diff
@@ -838,8 +925,16 @@ loop:
 			}
 			break
 		}
-		if et != textparse.EntrySeries {
+		switch et {
+		case textparse.EntryType:
+			sl.cache.setType(p.Type())
+			continue
+		case textparse.EntryHelp:
+			sl.cache.setHelp(p.Help())
```
brian-brazil: Where is the Help getting unescaped? I didn't see it in the lexing code.

fabxc: This is problematic since we'd allocate a new string. If we store it unescaped, we'd need to unescape all help strings on each scrape and do the same. If we store escaped and unescaped version, our memory usage for this feature doubles, since it's virtually all in help strings.

We could only do it on demand when reading the help strings via the API – that will also come with heavy allocs, but is certainly better.
Only for everything to get escaped all over again when we return it as JSON... 

Any ideas on how to handle this all?

There's also no generic unescape function AFAICS. So we have to do replacements ourselves. Is there a full list of escaped symbols we've to cover?

brian-brazil: https://golang.org/pkg/strconv/#Unquote

We've already got code for this for label values, might be worth seeing how it works. The vast majority of strings won't need escaping.

For help strings it's only \\ and \n.

fabxc: Yes, I realized it shortly after I posted. The docs don't quite give it away, the docs for the symmetric `Quote` are clearer. It expects actual quotes at beginning and end though, which means we need to allocate yet another transient string to add those.

fabxc: Ah right, we did the whole thing for label values already. The notable difference is that it's called only on first sight a series there. We'd be doing it on every scrape – so if there's escaped stuff in there, the impact is notably bigger.

brian-brazil: Tabs can be represented by themselves, so no need to escape them.

fabxc: It seems rather unusual. I don't think I've encountered anything in the wild yet the opted to only escape some things but not others. And I'm relatively sure it's gonna throw of more common escape handling (Go's JSON encoder errors for example) – after all this parser is not the only consumer.

docs/querying/api.md
fabxc: Fine with me. `metric` and `match_target` ok? `target_matcher` would work too but be inconsistent with our other endpoints with match params.

brian-brazil: sgtm

scrape/scrape.go
```diff
@@ -523,43 +528,62 @@ type scrapeCache struct {
...
 	seriesCur  map[uint64]labels.Labels
 	seriesPrev map[uint64]labels.Labels
+
+	metaMtx  sync.Mutex
+	metadata map[string]*metaEntry
+}
+
+// metaEntry holds meta information about a metric.
+type metaEntry struct {
+	lastIter uint64 // last scrape iteration the entry was observed
```
brian-brazil: Comment format is off here

scrape/scrape.go
brian-brazil: If that's the case then those parsers are buggy. \t for example should end up as a backslash followed by a t.

jacobsy:
Method Not Allowed
b 0g cqayvd b3n_5278vv9

PR #4201, opened 2018-05-29
Run rule evaluation with timestamps precisely evaluation_interval apart

free:
Tests still to come.

I don't particularly like the fact that I'm relying on the ticker to trigger (1) soon after `interval` has passed and (2) never trigger early. I believe it's perfectly safe, I just don't like it.

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -226,12 +226,12 @@ func (g *Group) run(ctx context.Context) {
 			case <-g.done:
 				return
 			case <-tick.C:
-				missed := (time.Since(evalTimestamp).Nanoseconds() / g.interval.Nanoseconds()) - 1
+				missed := (time.Since(evalTimestamp) / g.interval) - 1
 				if missed > 0 {
 					iterationsMissed.Add(float64(missed))
 					iterationsScheduled.Add(float64(missed))
 				}
-				evalTimestamp = evalTimestamp.Add(time.Duration((missed + 1) * g.interval.Nanoseconds()))
+				evalTimestamp = evalTimestamp.Add((missed + 1) * g.interval)
 				iter()
 			}
 		}
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -205,9 +205,10 @@ func (g *Group) run(ctx context.Context) {
 
 		start := time.Now()
 		g.Eval(ctx, g.evaluationTimestamp)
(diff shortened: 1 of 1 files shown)


free:
Hmm... I can't seem to find any Prometheus code/tests that allow for the use of fake clocks. Are there any you could point me to?

And if there aren't, should I:
(a) go ahead and vendor a package like https://github.com/jonboulle/clockwork; or 
(b) roll my own implementation that would also provide something along the lines of `NewTicker(interval, offset)`; or
(c) run some manual testing and declare it safe?

free:
BTW, here's the output of a manual test I ran with `--config.file=documentation/examples/prometheus.yml`, plus `first_rules.yml` defining the following rules:

```yml
groups:
  - name: example
    rules:
      - record: uptime
        expr: time() - process_start_time_seconds

      - record: iterations
        expr: prometheus_rule_group_iterations_total
```
screenshot from 2018-05-29 14-34-49

You'll notice from the series' timestamps that I had Prometheus suspended twice: once from 1527596169+ to 1527596197; and once from 1527596223+ to 1527596285. Also notice how the recorded rule samples all have the exact same phase (timestamp modulo eval interval) as do the values of the `uptime` recorded rule, exactly as expected.

brian-brazil:
This seems more complicated than it needs to be. How about in `iter` taking the current time, rounding it down and then adding the offset?

free:
Assuming an interval of `1m`, a current time of 16:41:10 and an offset of 50 seconds, you'd end up with an eval timestamp of 16:41:50, which is in the future. You could first subtract the offset; then round; and finally add it back; but I'm not sure that's actually simpler.

That does give me some ideas on how to clean it up a bit though.

free:
With apologies for the number of commits, I've cleaned it up a little. I'm still using the same logic for advancing the timestamp, but I've dropped the type casts (`missing` is now a `time.Duration`, but no one should really care); and I've rewritten the eval time calculation so they don't use an `if` block anymore.

I could also use the eval time calculation directly within the `for` loop instead of the current increment based calculation, but it's going to be less efficient. Your call.

brian-brazil:
Thanks!

I wouldn't worry about the number of commits, it usually makes it easier to review for me and for a change like this I'm going to squash them anyway.

PR #4206, opened 2018-05-30
add nan function

r0fls:


r0fls:
For the record, I'm aware there is a workaround already with `x != x` returning true for `NaN` values of `x`. However I think having a function for this would be cleaner.

brian-brazil:
Why do you care about the presence of NaNs?

r0fls:
We're having to do `x != x` for certain alert criteria.

r0fls:
In general, considering they do exist sometimes, it seems like it would be nice to be able to detect them elegantly.

brian-brazil:
Can you explain more about your use case?

New functions shouldn't be added lightly, and there's already a way to do this. Having issues with NaNs usually indicate that someone is attempting to average an average, which is not something we should be encouraging.

r0fls:
In this case it’s the result of performing division where the denominator occasionally becomes zero or absent.

r0fls:
I don’t see the reasoning for not adding new functions lightly either: as a math guy I was hoping to find any missing holes and possibly fill them in. Are there no plans for adding new functions to promql?

brian-brazil:
NaN is the correct result in that case, which indicates that you're not using the best query.

Can you share your full expression?

> I don’t see the reasoning for not adding new functions lightly either

Every additional function is additional cognitive load and complexity that all users must deal with, so they're not free. In addition some functions (like this one) have a good change to be abused far more than used correctly.

r0fls:
>Can you share your full expression?

Mostly I'm just disappointed because as I said, I was hoping to find some maybe slightly uncommon but hopefully useful-to-some-people math functions and add them.

Anyway, back to the matter at hand: I know that NaN is the correct value, I just think it would be more elegant to alert on `x/y > .6 or isnan(x/y)` rather than `x/y > .6 or absent(y) or y == 0` (particularly since y is a large expression). I don't see why there would be much added cognitive load here, since `isnan` or some other variant is fairly common in SQL and programming. If someone doesn't find it useful, they don't have to use it.

brian-brazil:
I can't help you much if you're not willing to share your use actual case, we add features on the basis of real requirements not that they might be theoretically nice to have. This then has to be balanced against the chances that users will use it incorrectly (which is a certainty in this case, every previous person asking questions relating to this was incorrectly aggregating), implementation complexity, support cost, cognitive load, and that every feature requires ongoing maintenance.

Your expression can be written as `x/y > .6 or y == 0`, which is simpler and faster. I'm not sure why you're bringing `absent` into it.

r0fls:
If `y` is absent, will `y == 0` be true? 🤔

brian-brazil:
It'll return an empty vector.

In rules you generally want to avoid the result being potentially missing in the first place, very tricky to work with. `absent` is unlikely to be the right solution, as it only works with one time series. `or` is better as it works with labels.

r0fls:
The above, x/y etc... is our actual use case, except x and y are actual metrics with specified dimensions. In particular: `kube_deployment_status_replicas_available{namespace="tectonic-system", deployment=~"tectonic-.*", deployment!="tectonic-prometheus-operator", deployment!="tectonic-cluo-operator"} / kube_deployment_status_replicas{namespace="tectonic-system", deployment=~"tectonic-.*", deployment!="tectonic-prometheus-operator", deployment!="tectonic-cluo-operator"} `


is the `x/y` expression. Namely, 

x=`kube_deployment_status_replicas_available{namespace="tectonic-system", deployment=~"tectonic-.*", deployment!="tectonic-prometheus-operator", deployment!="tectonic-cluo-operator"} / kube_deployment_status_replicas{namespace="tectonic-system", deployment=~"` 

and 

y=`kube_deployment_status_replicas{namespace="tectonic-system", deployment=~"tectonic-.*", deployment!="tectonic-prometheus-operator", deployment!="tectonic-cluo-operator"}`. 

Otherwise the alert config is precisely as stated (just figured I'd spare myself the trouble of trying to format this). The current alert rule, that deals with `NaN` values is:

```
IF kube_deployment_status_replicas_available{namespace="tectonic-system", deployment=~"tectonic-.*"} / kube_deployment_status_replicas{namespace="tectonic-system", deployment=~"tectonic-.*"} * 100 < 60
or kube_deployment_status_replicas_available{namespace="tectonic-system", deployment=~"tectonic-.*"} / kube_deployment_status_replicas{namespace="tectonic-system", deployment=~"tectonic-.*"} != kube_deployment_status_replicas_available{namespace="tectonic-system", deployment=~"tectonic-.*"} / kube_deployment_status_replicas{namespace="tectonic-system", deployment=~"tectonic-.*"}
```

With this change I could switch to:

```
IF kube_deployment_status_replicas_available{namespace="tectonic-system", deployment=~"tectonic-.*"} / kube_deployment_status_replicas{namespace="tectonic-system", deployment=~"tectonic-.*"} * 100 < 60 
or isnan(kube_deployment_status_replicas_available{namespace="tectonic-system", deployment=~"tectonic-.*"} / kube_deployment_status_replicas{namespace="tectonic-system", deployment=~"tectonic-.*"})
```

r0fls:
I realize after posting this that `"tectonic-.*"` vs `"tectonic-prometheus-operator"` etc are also redundant, was also hoping to spare myself the embarrassment of any public QA on such a gnarly expression 😊 I'll fix that later... but ya, I was hoping to do anything i could to make that expression less gnarly.

r0fls:
>I was hoping to do anything i could to make that expression less gnarly

Beyond that, to me, it seems like a missing feature. People misusing it doesn't seem like a serious enough problem. For them, the answer would be RTFM. James Joyce didn't rewrite Ulysses because it was too complicated (I didn't read that, just saying) 😉 Nor does future maintenance seem that worrisome: this should be pretty straightforward to maintain. You can ping me personally for needed changes to this, cross my heart and hope to die.

brian-brazil:
You can simplify to:
```
      kube_deployment_status_replicas_available{namespace="tectonic-system", deployment=~"tectonic-.*"}
    /
      kube_deployment_status_replicas{namespace="tectonic-system", deployment=~"tectonic-.*"} * 100
  < 
   60
or 
   kube_deployment_status_replicas{namespace="tectonic-system", deployment=~"tectonic-.*"}  ==  0

```

What you really care about is there being no replicas, not that a derived calculation returns NaN.

> People misusing it doesn't seem like a serious enough problem. For them, the answer would be RTFM. 

I'd rather not make it easier for users to shoot themselves in the foot when calculating key metrics such as latency, which based on experience is what at least 95% of users of a function like this would do. We already have documentation on this which some users are not following, so there's no reason to believe they'd read the docs on this either. Given this, adding this function (which I have pondered in the past) would require an extremely strong use case - particularly as there's already a way to do it in PromQL.

As you can see here there's simpler ways to do what you want without producing a NaN that needs filtering, so I appreciate your contribution but it's not something that makes sense for the project at the moment.

r0fls:
thanks for the deliberation

PR #4207, opened 2018-05-31
Add prometheus_oldest_timestamp_seconds metric

### LINKED ISSUE #2988: New prometheus metric storing information about series age
Suggested scenario for Prometheus HA is to have multiple instances of Prometheus scraping same targets. But there is no way to distinguish which instance is oldest.

This can be problem for ex. graphing tools (grafana).

1) Spawn 2 instances of prometheus (A, B)
2) Instance A is down
3) Spawn new instance (C)
4) Grafana should use 'B' to serve graphs, till 'C' will catch up with series

Prometheus `process_start_time_seconds` can't be treated as reliable for this purpose.

We can use load-balancer to disable 'fresh' instances, but to make this more autonomous (no operator intervention needed) we need metric which gives us information which Prometheus stores oldest series.

We can expose a few additional metrics:
1) `oldest_metric` timestamp
2) local storage size
3) number of series in storage

Option 2) sounds ridiculous and not reliable
Option 3) based on @brian-brazil opinion will not be easy to implement in Prometheus 2.X

bobmshannon:
Fixes #2988

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -49,6 +49,7 @@ import (
 	"github.com/prometheus/prometheus/discovery"
 	sd_config "github.com/prometheus/prometheus/discovery/config"
 	"github.com/prometheus/prometheus/notifier"
+	"github.com/prometheus/prometheus/pkg/timestamp"
 	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/rules"
 	"github.com/prometheus/prometheus/scrape"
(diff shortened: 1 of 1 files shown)


cmd/prometheus/main.go
```diff
@@ -68,6 +69,10 @@ var (
 		Name: "prometheus_config_last_reload_success_timestamp_seconds",
 		Help: "Timestamp of the last successful configuration reload.",
 	})
+	oldestTimestamp = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "prometheus_oldest_timestamp_seconds",
```
grobie: I wonder if `prometheus_tsdb_start_time_seconds` would be a better name, and whether this metric should be exported by tsdb itself.

bobmshannon: Yeah good point, going to follow up on https://github.com/prometheus/tsdb/pull/363.

PR #4209, opened 2018-06-01
Allow override of PROMU to prevent rebuilding every make

claesjac:
This patch allows override of PROMU to prevent a go get of promu on every build attempt. Similar to commit 41271cd in node_exporter


gouthamve:
@SuperQ @simonpasquier Is this still relevant? Could you take a look at it ask for a DCO?

Also @claesjac sorry dropping the ball on this! We definitely do need this enhancement!

simonpasquier:
I think that the PR can be closed. `Makefile.common` has been modified some time ago to only install `promu` when it isn't found in `$(GOPATH)/bin`. It also installs from the release tarball instead of doing `go get`.

PR #4210, opened 2018-06-01
Add queue name to logger for remote writes

bboreham:
More than one remote_write destination can be configured, in which case it's essential to know which one each log message refers to.


fabxc:
👍 Thanks

PR #4212, opened 2018-06-04
web: limit the number of dropped targets

### LINKED ISSUE #4134: /service-discovery returns excessive amount of HTML for large K8s clusters
The handy `/service-discovery` endpoint also show all dropped labels. However, SD often retrieve essentially all the meta data and then drop most of it for each job. In the case of K8s, imagine a decently sized cluster of 10k pods, each with 10 labels or annotations, each needs ~1kiB to display in HTML with its operation. A Prometheus server with 10 configured job would display all 100k labels/annotations in the K8s cluster for each of the 10 job, i.e. 1M labels/annotations, resulting in about 1GiB of data. That's just a broad estimate. I have just done a practical measurement and retrieved 1.4GiB of HTML code from the `/service-discovery` endpoint. This is a drain on the Prometheus server and the network, and usually kills the browser tab. Most K8s clusters might be smaller, but the ~10k pod clusters at SoundCloud are not meant to be beyond the scalability limit of Prometheus. Selected few organizations even run much larger clusters, which Prometheus should still be able to monitor.

In short, the `/service-discovery` endpoint needs to be capped for such large clusters. Details need to be figured out, but it could be something like a message "displaying only 100 dropped labels. There are 999,900 more labels." with a pagination mechanism or a download link for same raw data of the labels (which is then explicitly requested by the user rather than accidentally triggered by visiting the Service Discovery status page).

IIRC @Conorbro implemented the endpoint and might have the best ideas how to proceed here.

@krasi-georgiev : as discussed.

simonpasquier:
Closes #4134.

cc @beorn7 @brian-brazil 

Here is the updated service-discovery page (with the warning message):

image

### DIFF
--- a/web/web.go
+++ b/web/web.go
@@ -600,10 +600,33 @@ func (h *Handler) serviceDiscovery(w http.ResponseWriter, r *http.Request) {
 	scrapeConfigData := struct {
 		Index   []string
 		Targets map[string][]*scrape.Target
+		Active  []int
+		Total   []int
 	}{
 		Index:   index,
-		Targets: targets,
+		Targets: make(map[string][]*scrape.Target),
+		Active:  make([]int, len(index)),
+		Total:   make([]int, len(index)),
 	}
+	for i, job := range scrapeConfigData.Index {
+		nbDropped := 0
+		scrapeConfigData.Targets[job] = make([]*scrape.Target, 0, len(targets[job]))
+		scrapeConfigData.Total[i] = len(targets[job])
+		for _, target := range targets[job] {
+			// Do not display more than 1k dropped targets per job to avoid
+			// returning too much data to the clients.
+			if target.Labels().Len() == 0 {
+				nbDropped++
+				if nbDropped > 1000 {
+					continue
+				}
+			} else {
+				scrapeConfigData.Active[i]++
+			}
+			scrapeConfigData.Targets[job] = append(scrapeConfigData.Targets[job], target)
+		}
+	}
+
 	h.executeTemplate(w, "service-discovery.html", scrapeConfigData)
 }
 
--- a/web/web.go
+++ b/web/web.go
@@ -601,23 +601,24 @@ func (h *Handler) serviceDiscovery(w http.ResponseWriter, r *http.Request) {
 		Index   []string
 		Targets map[string][]*scrape.Target
 		Active  []int
+		Dropped []int
 		Total   []int
 	}{
 		Index:   index,
 		Targets: make(map[string][]*scrape.Target),
 		Active:  make([]int, len(index)),
+		Dropped: make([]int, len(index)),
 		Total:   make([]int, len(index)),
 	}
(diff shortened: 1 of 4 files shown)


web/web.go
```diff
@@ -600,10 +600,33 @@ func (h *Handler) serviceDiscovery(w http.ResponseWriter, r *http.Request) {
...
+		Targets: make(map[string][]*scrape.Target),
+		Active:  make([]int, len(index)),
+		Total:   make([]int, len(index)),
+	}
+	for i, job := range scrapeConfigData.Index {
+		nbDropped := 0
+		scrapeConfigData.Targets[job] = make([]*scrape.Target, 0, len(targets[job]))
+		scrapeConfigData.Total[i] = len(targets[job])
+		for _, target := range targets[job] {
+			// Do not display more than 1k dropped targets per job to avoid
```
brian-brazil: We should display a warning to the user when this happens

beorn7:
Sweet. Thanks for tackling this.

I'll leave the detailed review to Brian, since he is already on it.

Just one high level remark: Is 1k targets sufficient? In the SoundCloud case, the cluster had around 10k pods, and the size of the HTML on the SD endpoint was 1.4GiB. The math might be more complicated, but I would naively assume the limit of 1k will cut the size by a factor of 10. But 140MiB still sounds like a lot for hitting a simple status information endpoint.

brian-brazil:
The limit is only on the dropped targets, the active ones are always shown.

beorn7:
Idea: Perhaps do something like displaying everything as long as at most 1k dropped targets will actually display everything. However, if we have to limit the displayed targets anyway, there is not much gain between showing 100 or 1000. In that case, we could limit to something low, e.g. 100 or even less, and display a warning along the lines: "12,345 targets dropped, displaying only the first 100 as examples."

beorn7:
> The limit is only on the dropped targets, the active ones are always shown.

To avoid misunderstandings: I'm only talking about dropped targets here.
By "Is 1k targets sufficient?" I mean "Is 1k targets sufficiently _low_?"

brian-brazil:
How many different kubernetes_sd_configs did ye have inside one Prometheus again?

beorn7:
That really depends. I think we need to cater for the use case of tens just by number of jobs configured, times a small factor if we have multi-port per pod setups.

brian-brazil:
I was trying to get a gauge of how many jobs that 1.4GB covered.

beorn7:
That particular config is gone by now. It would require some forensics to find out.

However, I don't think it matters so much as Prometheus is meant to scale nicely, and we should make absolutely sure that the scalability limit is not determined by the status page of the web UI.

Thus we should have very conservative limits.

brian-brazil:
Okay, let's go with 100.

simonpasquier:
I've updated the PR according to the comments.

brian-brazil:
Thanks!

PR #4222, opened 2018-06-05
Add prompb README

henridf:
Just a small doc on requirements to compile the proto defs.

### DIFF
--- a/prompb/README.md
+++ b/prompb/README.md
@@ -1,0 +1,14 @@
+The compiled protobufs are version controlled and you won't normally need to
+re-compile them when building Prometheus.
+
+If however you have modified the defs and do need to re-compile, run
+`./scripts/genproto.sh` from the parent dir.
+
+In order for the script to run, you'll need `protoc` (version 3.5) in your
+PATH, and the following Go packages installed:
+
+- github.com/gogo/protobuf
+- github.com/gogo/protobuf/protoc-gen-gogofast
+- github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway/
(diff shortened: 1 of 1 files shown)


prompb/README.md
```diff
@@ -0,0 +1,15 @@
+The compiled protobufs are version controlled and you won't normally need to
+re-compile them when building Prometheus.
+
+If however you have modified the defs and do need to re-compile, run
+`./scripts/genproto.sh` from the parent dir.
+
+In order for the script to run, you'll need `protoc` (version 3.5) in your
+PATH, and the following Go packages installed:
```
brian-brazil: It seems a bit odd to me that we'd need to list transitive dependencies like this. Is there one particular Go package that's required that can be go get-ed?

henridf: you're right; I simplified the list

brian-brazil: I was thinking more that there should only be one dependency we need, it feels odd that there's 4 (and I didn't need 4 the last time I regenerated this).

henridf: If you're thinking that we shouldn't need all of these packages, I'm pretty sure that's the case. It took a bit of fumbling about to figure out what needed to be imported and that's why I figured this README would be useful; 

But maybe I'm missing something altogether and there's a simpler way - happy to drop this PR if so.

brian-brazil: The PR is welcome, I'm just wondering if we can make it a bit simpler. Let me have a poke.

brian-brazil: Okay, looks like all those are required and github.com/gogo/protobuf too

brian-brazil:
Thanks!

PR #4226, opened 2018-06-05
Add offset to selectParams

### LINKED ISSUE #4224: storage.Querier doesn't handle offsets efficiently
## Bug Report
I have an implementation of the storage.Querier interface for my storage, with promql on top. In this setup if a user sends a query for a metric at a single point with an offset of 4w the querier created by promql spans 4w. This makes it difficult (to say the least) to create an efficient querier implementation. presumably similar issues exist inside prom itself (since it would potentially load significantly more data than necessary). I currently have 2 fixes for the issue, either of which I'd be glad to merge upstream-- I'm looking for some input on what fix you'd prefer.

### Option 1: multiple queriers
If we don't want to change the querier interface at all we can simply make promql/engine.go create a querier for each offset. This does potentially mean more connections etc. depending on what the querier does under the hood, but does avoid the gross over-scanning of data.
Code: https://github.com/jacksontj/prometheus/commit/4475e2a546953619d738f4a92923484fe021d783

### Option 2: add offset to selectParams
If we are okay adding a field to selectParams we can simply pass in the offset there on each actual `Select`. This mechanism allows for the querier to be re-used, and then makes it optional for the querier to actually look at the offset. This method would require Queriers to make code changes to pick it up, but it is backwards compatible (meaning if they don't look at the selectParam field, then they just fall back to current behavior).
**Given the interface etc. I think this is the one we should do.**
Code: https://github.com/jacksontj/prometheus/commit/cb722690f3aa9f38ac5d9a3b0e95ea1a417cfb78

**Environment**

* Prometheus version:

2.2 (same issue exists in master)

jacksontj:
This adds the Offset from the promql.EvalStmt to the selectParams which
is sent to the querier during Select()

Fixes #4224


storage/interface.go
```diff
@@ -63,8 +64,9 @@ type Querier interface {
 
 // SelectParams specifies parameters passed to data selections.
 type SelectParams struct {
-	Step int64  // Query step size in milliseconds.
-	Func string // String representation of surrounding function or aggregation.
+	Step   int64         // Query step size in milliseconds.
+	Func   string        // String representation of surrounding function or aggregation.
+	Offset time.Duration // Offset for this particular select statement
```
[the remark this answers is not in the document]
brian-brazil: This should also be going to the proto as this should also be sent with remote read, and a time.Duration could be confusing in that regard. Can this have sub-millisecond resolution values?

brian-brazil:
It should be.

jacksontj:
@henridf to clarify, the "remote read" changes you mean are to the Querier client for remote read to use the offset, correct?

jacksontj:
@brian-brazil anything else that I should include in this PR? Or are we all set (assuming the cross-platform build CI completes with no issue)?

henridf:
@jacksontj what I meant was passing down the new `Offset` hint as part of a read request, similar to what I did in #4122.

jacksontj:
@henridf the `Offset` shouldn't be required in the remote read proto files etc. as the Remote read is already a query for a specific time range (start -> end) unlike the Querier interface which is an iterator for a time range. The change I put in changes it so that prom won't ask for the large sections of data, meaning the remote read situation is already covered.

jacksontj:
I can add the offset value to the proto file, but they define different interfaces. The Querier interface is an iterator (meaning it needs to respond to data within a timerange) whereas the remote read is a  read request for a specified time range. If the remote read API used duplexed streaming I think it'd make sense (since it'd effectively be an iterator) but since its just request+response it seems unnecessary.

This is probably more obvious looking at https://github.com/prometheus/prometheus/pull/4226/commits/2e963e0706631b5dd215e4293ee74c130ea6c409 -- where you can see the offset being used already before hitting the remote read.

That being said, if we want to pass it for completeness I can add it-- I'll just add a comment there stating the same.

henridf:
@jacksontj i've taken a closer look and I'm not sure I understand the need for changing `SelectParams` at all - why can't we simply adjust `maxt` when building the Querier, similar to what we do for `mint`?

brian-brazil:
This should be a mirror of the existing APIs, we shouldn't be inventing a subtly different API that'll end up confusing people.

jacksontj:
@henridf let me explain with an example. Lets start with the generic (to make sure we're on the same page as far as flows go). 

Let say you need to query a single datapoint. 

With the querier interface it'd look something like:
```
q, _ :+ Querier(ctx, mint, maxt)
set, err := q.Select(params, labelMatchers...)
```
With the remote read interface it'd look something like:
```
result, err := remoteClient.Read*ctx, &prompb.Query{
    StartTimestampMs: start,
    EndTimestampMs: end,
    Matchers: <insert matchers>,
})
```

The main difference in usage is that the Querier interface is 2 steps (create a querier, then query) and the remote read is a single RPC call. The issue comes down to when we define time ranges that these should return data for. The short version is that the querier's time range is querier-wide whereas the remote-read times are per-query  (the old querier interface was like this too, FYI).

So to clarify with an example, lets say we have a query like so: `error_rate > (error_rate offset 1w)`. This promql is going to check if the current `errror_rate` (assume a recording rule or something, just simplifying for the example) is greater now than it was last week (effectively a week over week alert, fairly common).

This promql breaks intoa a binary clause of (1) `error_rate` (2) `error_rate offset 1w) with a comparator of `>`.
Going back to our two paths:

Querier: 
```
q, _ :+ Querier(ctx, <now-1w>, now)  // not real variables, but you get the idea
// Fetch data for clause 1
set1, err := q.Select(params, clause1Matchers...)
set2, err := q.Select(params, clause2Matchers...)
```

Remote read:
```
result1, err := remoteClient.Read*ctx, &prompb.Query{
    StartTimestampMs: <now-LookbackDelta>,
    EndTimestampMs: now,
    Matchers: clause1Matchers,
})
result2, err := remoteClient.Read*ctx, &prompb.Query{
    StartTimestampMs: now-1week-LookbackDelta,
    EndTimestampMs: now-1week,
    Matchers: clause2Matchers,
})
```

In this example we see that the `Select` happening on the querier only defines a single time range, which means that the two selects are being asked to fetch data for the entire time range. In contrast the remote read has separate RPC calls for both data fetches where it can be explicit about what timerange it wants data for.

So, this means that the fix in this PR is actually insufficient. This PR fixes the case where all clauses in the promql are in the same offset, but in the case where there are multiple offsets we're back to the same problem we had originally (so better, but not good enough). In the issue (#4224 ) I had referenced another fix (https://github.com/jacksontj/prometheus/commit/4475e2a546953619d738f4a92923484fe021d783) which makes a querier per-offset. This would effectively solve this issue with a promql-only change. From thinking about this more I think this is the best of the two options. Alternatively if we where to change the querier interface back to the 1.x interface-- where the time range was defined at select time, that would also solve the issue. Given the work required for that change, I believe the per-offset querier is the better way to go.

brian-brazil:
As I said, please don't make changes to the protocol that'll make it differ from the internal protocol. You can pass down the offset, but not any more than that.

jacksontj:
As I just described in the comment above, we can do that but do realize that it is insufficient to resolve the actual issue. The underlying issue is that the promql engine isn't passing enough time context to the Querier interface, leaving the underlying queriers to have to "lazily load" data in an attempt to not over-fetch data. This all seems excessive since the promql engine knows exactly what range of data it needs.

Also, to clarify the other change I propose makes no changes to the protocol its an internal promql fix that requires no changes outside of the promql package (and a small one to the promql package at that).

henridf:
@jacksontj thanks for the explanation. 

for anyone following from afar who just wants to get the gist, the key point is: 
> The short version is that the querier's time range is querier-wide whereas the remote-read times are per-query

brian-brazil:
> which makes a querier per-offset. This would effectively solve this issue with a promql-only change. 

That would break isolation, there needs to be exactly one querier per query. Passing context is okay, changing core internals is not.

jacksontj:
@brian-brazil I'm not sure how it breaks isolation (unless we are using variables from inside the context or something)? The only other thing I could think of is concerns about re-using series from the querier (cache hits) but that shouldn't be a concern since we'd be splitting on a time range which won't overlap (or at least largely won't).

Since promql allows for mutliple offsets (in other words multiple time ranges) in a given query we need to handle it.

If we want a single querier per query then we can pass times in the `Select` in addition to `Querier` creation (defining a set of times within the querier times). This way the `Querier` is bounded to know what the max range each query would be, and the `Select` is scoped to the exact data that we want to fetch. If we go with this approach we could put it in selectParams (startMs and endMs -- instead of offset).

If we are okay with multiple queriers we don't have to change anything outside of promql. The only problem with this so far is a worry about breaking isolation, since I'm unaware of any isolation thats being broken I'll have to wait for your response with some more detail.

jacksontj:
cc @juliusv as I imagine your cortex product is impacted by this same issue.

juliusv:
@jacksontj Thanks for flagging this! I think what Brian meant with isolation is described starting at slide 33 here: https://www.slideshare.net/brianbrazil/staleness-and-isolation-in-prometheus-20-promcon-2017

So basically, when you create a TSDB querier, each querier gets a consistent view of the TSDB (e.g. no half scrapes, as batched writes are atomic), and creating two queriers for different selectors in a single PromQL query would mean that those queriers could get different views of the data.

I haven't managed to take a deeper look at the solution to this problem here yet, but will try to find some time.

brian-brazil:
>  unless you want to stuff the context with all kinds of magical per-series hints.

That's been the approach we agreed on thus far, and what this PR was meant to do. We don't know all the information remote read endpoints might need yet, so should avoid redesigning things every time we find a new one.

juliusv:
@brian-brazil This is different though in that the remote protobuf protocol or the remote end don't need to know at all about this change because they already support per-selector offsets, that support just needs to be used by the remote read code. Selecting only the needed time ranges for each selector sounds like something pretty fundamental.

Could you point me at where it was agreed to stuff things into the `ctx` parameter of the `storage.Querier()` method? Or were you referring to putting this into the `storage.SelectParams`, where it would be too late, because that only applies in the `querier.Select()` phase, whereas we already need to know it earlier when we create the querier. At least if we don't want to lazy-load everything during `Select()`, which would cause unneeded roundtrips.

brian-brazil:
> This is different though in that the remote protobuf protocol or the remote end don't need to know at all about this change because they already support per-selector offsets

I wouldn't be 100% sure about that. The remote endpoint could be doing optimisations where the range of the original query matters, such as by bringing it into a different set of downsampling rules.

> Could you point me at where it was agreed to stuff things into the ctx parameter of the storage.Querier() method?

That's an implementation detail. The agreement we'd pass information that could be of use to remote read as a set of hints was at the dev meeting after the last Promcon. This is as we don't know yet what they actually need, and we can formalise it later when the dust settles.

juliusv:
> I wouldn't be 100% sure about that. The remote endpoint could be doing optimisations where the range of the original query matters, such as by bringing it into a different set of downsampling rules.

I think I would need an example to understand what you mean... from what I can see, passing each selector only with the actually needed time ranges would be by default better for the remote end, and I'm not even sure you would lose any information about the original range of the query, because that could be deduced from the different offsetted subqueries.

> That's an implementation detail. The agreement we'd pass information that could be of use to remote read as a set of hints was at the dev meeting after the last Promcon. This is as we don't know yet what they actually need, and we can formalise it later when the dust settles.

I don't think that such an agreement should include the passing of needed time ranges in that kind of context hint. IMO that's such a basic and explicit requirement that it should be part of the main interface.

juliusv:
Or to put it differently: the remote protocol already *has* the explicit fields for this per selector, we're just not using the protocol correctly / as intended.

juliusv:
@henridf

> If the remote end has only downsampled data for 1 week ago, it seems like it should provide the same downsampled resolution for the current-time data

Ah, thanks. But it could still do that, since no information is really lost, the overall start/end of all subqueries are still known. If at all, we could add an extra hint (if ever needed, which I doubt) about the total time range, but that part should be the hint then, not the other way around.

henridf:
>   But it could still do that, since no information is really lost, the overall start/end of all subqueries are still known.

I might be misunderstanding something, but the remote read endpoint just has no way to distinguish if the successive `ReadRequest` messages it receives are the result of a single PromQL `error_rate > (error_rate offset 1w)` or of two separate PromQL queries `error_rate` and `(error_rate offset 1w)`.

brian-brazil:
> Or to put it differently: the remote protocol already has the explicit fields for this per selector, we're just not using the protocol correctly / as intended.

I'd disagree there, the protocol is being used as intended. It's just not optimised for this use case.

> If at all, we could add an extra hint (if ever needed, which I doubt) about the total time range, but that part should be the hint then, not the other way around.

That counts as a breaking change to the protocol, and we should save those to when we're doing them all in one big bang.

> I might be misunderstanding something, but the remote read endpoint just has no way to distinguish if the successive ReadRequest messages it receives are the result of a single PromQL

Not today, but long term they'll all come in one request. This is also required for isolation.

juliusv:
>> Or to put it differently: the remote protocol already has the explicit fields for this per selector, we're just not using the protocol correctly / as intended.
>
> I'd disagree there, the protocol is being used as intended. It's just not optimised for this use case.

So to clarify here, are you saying as the reverse of this that if I pass in the correct time ranges instead of the wrong ones into the explicit fields that already exist for selecting the time ranges you need, it's a wrong use of the protocol? That would seem a bit... surprising.

brian-brazil:
> That would seem a bit... surprising.

The protocol was designed for use with the local storage, and then extended to other uses. It's the time range for the query as a whole, rather than each specific selector. So it looks odd in another light, but it is how it was meant to be used.

juliusv:
> The protocol was designed for use with the local storage

@brian-brazil I was now talking about the remote storage protocol, not the `storage.Querier()` method... the remote storage protocol always allowed specifying a time range per selector by putting multiple `Query` messages into a single overall remote read request.

I still think it also makes sense to change the internal Go storage interfaces to work better with remote storages.

juliusv:
@brian-brazil Do you think sending only the required ranges rather than overly broad ones would break any existing implementations? I mean it's not like anything about the protocol actually changes, we just request less.

brian-brazil:
Of the ones that I know exist, only Thanos is sophisticated enough that it could be relevant.

juliusv:
While Thanos does use remote read to read *from* Prometheus, I don't think they provide a remote read endpoint themselves (instead you query via their querier, which is v1 query API on the frontend and does its own gRPC protocol in the backend). What would you expect to break if specifying only required vs. broader time ranges?

brian-brazil:
They're dug into the internals enough that this may affect them and they are using these interfaces, and I know they have downsampling.

juliusv:
@brian-brazil But to repeat myself, Thanos does not even implement a Prometheus remote read endpoint.

jacksontj:
> So basically, when you create a TSDB querier, each querier gets a consistent view of the TSDB (e.g. no half scrapes, as batched writes are atomic), and creating two queriers for different selectors in a single PromQL query would mean that those queriers could get different views of the data.

This is good, but I don't see this requirement documented in the storage interface at all-- which means that I doubt this consistency guarantee exists for anything besides the TSDB. And in a distributed storage system thats not really feasible, but that being said I can get the want for a Querier to be one per query. The main driver I can see from a remote storage case would be a time-sharded store. If the querier was created with the max time range it could have leased connections to the shards necessary to cover that range.

If we want to keep a single querier per promql query (which it very much sounds like we want), then we need to pass start/end times to the `Select()` itself. It sounds like we aren't willing to "break" the API by adding that to the select itself, so I think we should just add a `startMs` and `endMS` to the `selectParams` and call it a day. With those parameters the storage implementation would know exactly what range *that* `Select` is for, and if they don't look at it they'd over-fetch (as opposed to changing the Querier interface and causing a compile-time error). 

IMO the time range for the `Select` is important enough that it should go direct as args to the `Select`, but I could live with that being delayed for some future release (I'm not sure what the API compatibility is at this layer, so I'll defer to you guys on that).

From the other side of the conversation, it doesn't sound like the same start/end times are needed in the remote read API since that RPC call already has the start/end times (instead of the range stuff).

brian-brazil:
> But to repeat myself, Thanos does not even implement a Prometheus remote read endpoint.

Sure, but this indicates that things in this space may be depending on the current behaviour. Just simple changing the API could leave them stuck.

> This is good, but I don't see this requirement documented in the storage interface at all

It's still to be implemented and thus documented in Prometheus, but we shouldn't be doing things that'll prevent it from ultimately working. In practical terms this will likely only really matter for Prometheus to Prometheus remote read, as for everything else the data is likely old and thus settled enough not to have isolation issues.

> I'm not sure what the API compatibility is at this layer, so I'll defer to you guys on that

There is officially no API compatibility on internals, only on certain external things.

juliusv:
>> But to repeat myself, Thanos does not even implement a Prometheus remote read endpoint.
>
> Sure, but this indicates that things in this space may be depending on the current behaviour. Just simple changing the API could leave them stuck.

While it would surprise me if it caused problems, I understand the reluctance to spread one PromQL query over multiple `remote.Query` (https://github.com/prometheus/prometheus/blob/master/prompb/remote.proto#L34-L39) messages, even if those end up being part of a single remote read request.

You're right, in the case of Prometheus->Prometheus remote read, the serving Prometheus would currently split up multiple incoming `remote.Query` messages from the same remote read request into multiple `storage.Querier`s (https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L602-L661) and thus the PromQL query from the client Prometheus server will not be guaranteed isolation / consistency.

It might be the only case, and even in this case, it probably matters little, but I see the general worry. Not sure what to do yet, will think further...

brian-brazil:
As @jacksontj says we can pass down one range in the hints, and the other as the primary range. The point of debate is more which range goes where.

jacksontj:
I think at this point it seems that we should have the quererise get the
larger range (defining a range that will encapsulate all ranges for all
selects) then add a start/end to the select params that defines the range
for the query itself (with a guarantee that all selectparam ranges fall
within the querier range).

This isn't an API break, but this bug is a huge perf issue. So I'd rather
make that time range part of the main API interface (so people have to fix
and rebuild, instead of continuing with the incorrect ranges). Given that
this is all internal APIs we might be able to do so, I'll defer to @bbrazil
for that.

Depending on which of those 2 are the way to go I can update this PR to be
either fix. It sounds then like we'd leave the remote read alone, as we'd
change the server itself to create q querier for all queries in the single
remote read request when we need to make guarantees for request isolation.

jacksontj:
To followup here, what did we want to do? It sounds like we have settled on either (1) adding start/stop to selectParams or (2) adding start/stop as args to Select() (unless there was a third option I've missed?)

brian-brazil:
I'm for (1).

jacksontj:
I can live with (1) although I'd prefer (2) since its really a fundamentally required field when querying. My reasoning is that it is such a HUGE performance liability that it would make sense for people to fix that with the API breakage, whereas if we leave it people's implementations that exist relying only on the querier range will continue to over-fetch data (prometheus is also impacted, just less so as it lazily fetches blocks, so it over-fetches *less* data). For this I'll defer to you on what the support/api-compatibility requirements are. It sounds like the API contract there is not guaranteed, so a "break" like this should be okay. 

Are there concerns with breaking the API there?

jacksontj:
cc @fabxc (as adding time selectors to the Select() method in the TSDB would be good as well -- to avoid over-fetching)

brian-brazil:
Correctness is more important that performance in my opinion, and (2) is a correctness risk.

> (as adding time selectors to the Select() method in the TSDB would be good as well -- to avoid over-fetching)

As I've already indicated, that's not the way the TSDB works.

fabxc:
> (as adding time selectors to the Select() method in the TSDB would be good as well -- to avoid over-fetching)

As you wrote further up, TSDB loads data lazily. So there is no over-fetching beyond the very high level (how many blocks to we include to begin with), which is unlikely to show up in any performance profile.  
Could you clarify what kind of problem adding time selectors would solve?

brian-brazil:
> Could you clarify what kind of problem adding time selectors would solve?

That works for the TSDB, but not remote read. So we want to pass down somehow the data that'll actually be fetched, as distinct from the entire time range the query is looking for.

fabxc:
Yes, I understand that remote read may have different requirements and the glue interface in Prometheus may have to deviate more from TSDB's – that's what its there for after all.
But here the specific proposal was to make changes to TSDB 's interfaces.

brian-brazil:
The option I propose does not change the TSDB/PromQL interface, it's just more hints.

jacksontj:
> As you wrote further up, TSDB loads data lazily. So there is no over-fetching beyond the very high level (how many blocks to we include to begin with), which is unlikely to show up in any performance profile. Could you clarify what kind of problem adding time selectors would solve?

The time selector passed into the Querier() is actually the maximum range of data that you want to fetch. In the example from the issue (#4224) -- this can be a HUGE over-estimation of data actually required. As a quick re-cap a point-in-time query of `foo > bar offset 4w` would have a querier range of 4w, meaning we'd potentially fetch 4w of blocks, when in reality we need 2 points (one in the most recent block, and one from 4w ago). More specitically the querier would be called twice, once with `foo` and once with `bar`. The suggestion in this PR is to pass time ranges to the Select() as well, so the select can pick the exact ranges that data is needed for (instead of relying soely on lazy loading of data, as that has some impact-- and isn't feasible for remote storage).

> The option I propose does not change the TSDB/PromQL interface, it's just more hints.

This would be fine (although I still prefer passing it as a first class argument, but as I said before I'm fine with it as a selectParam). The key would be if we are going to pass it as a selectParam TSDB might want to actually use the selectParams to avoid the over-fetching.

jacksontj:
I have updated this PR to include the start/end time in the [selectParams](https://github.com/prometheus/prometheus/pull/4226/commits/687825e63292f5e9ce64c6f58ca2464b14f5f9d4) (as agreed on above). I have also changed the remote-read API to properly pass in [start/end times](https://github.com/prometheus/prometheus/pull/4226/commits/d79f7a146d32629312cf5d8c015aab1b815493ac).

For posterity, I've also made a [test branch](https://github.com/jacksontj/prometheus/tree/issue_4224_example) to show how much of a difference this can make.

In this branch I print out the query that the remote read client creates for a given promql statement

Vector Statement: `http_requests_total + http_requests_total offset 30m`
Old queries: 
 - `start_timestamp_ms:1528996031853 end_timestamp_ms:1528998131853 matchers:<name:"__name__" value:"http_requests_total" > hints:<> ` (35 minute select)
 - `start_timestamp_ms:1528996031853 end_timestamp_ms:1528998131853 matchers:<name:"__name__" value:"http_requests_total" > hints:<> ` (35 minute select)

New queries:
- `start_timestamp_ms:1528997831853 end_timestamp_ms:1528998131853 matchers:<name:"__name__" value:"http_requests_total" > hints:<> ` (5 minute select)
- `start_timestamp_ms:1528996031853 end_timestamp_ms:1528996331853 matchers:<name:"__name__" value:"http_requests_total" > hints:<> ` (5 minute select).

So in addition to not over fetching, we aren't *duplicating* the fetch.

This same benefit can be seen on the matrix selectors as well:

Matrix Statement: `rate(http_requests_total[1m]) + rate(http_requests_total[1m] offset 30m)`
Old queries:
 - `start_timestamp_ms:1528996126989 end_timestamp_ms:1528997986989 matchers:<name:"__name__" value:"http_requests_total" > hints:<func:"rate" > ` (31 minute select)
 - `start_timestamp_ms:1528996126989 end_timestamp_ms:1528997986989 matchers:<name:"__name__" value:"http_requests_total" > hints:<func:"rate" > ` (31 minute select)

New queries:
 - `start_timestamp_ms:1528997626989 end_timestamp_ms:1528997986989 matchers:<name:"__name__" value:"http_requests_total" > hints:<func:"rate" > ` (6 minute select)
- `start_timestamp_ms:1528995826989 end_timestamp_ms:1528996186989 matchers:<name:"__name__" value:"http_requests_total" > hints:<func:"rate" > ` (6 minute select)

storage/interface.go
```diff
@@ -63,6 +63,9 @@ type Querier interface {
 
 // SelectParams specifies parameters passed to data selections.
 type SelectParams struct {
+	Start int64 // Start time in milliseconds for this select
```
brian-brazil: Full stop

storage/remote/read.go
```diff
@@ -44,7 +44,7 @@ type querier struct {
 // Select implements storage.Querier and uses the given matchers to read series
 // sets from the Client.
 func (q *querier) Select(p *storage.SelectParams, matchers ...*labels.Matcher) (storage.SeriesSet, error) {
-	query, err := ToQuery(q.mint, q.maxt, matchers, p)
+	query, err := ToQuery(p.Start, p.End, matchers, p)
```
brian-brazil: This needs to stay as is, so that the protocol doesn't change and the remote endpoint still knows the full range of the query.

jacksontj: This is actually a bug, as described in detail (with example code showing the problem) in https://github.com/prometheus/prometheus/pull/4226#issuecomment-397382955 With this the way it was we are *severely* over fetching data for promql queries that have offsets.

storage/interface.go
jacksontj: fixed!

brian-brazil:
> I have updated this PR to include the start/end time in the selectParams (as agreed on above). I have also changed the remote-read API to properly pass in start/end times.

As I've said several times above, you may add hints but you may not change the protocol. Please stop making changes beyond what was agreed.

storage/remote/read.go
brian-brazil: I understand what you are saying and have since your initial report. You can use the new hints to make your software deal better with this scenario, but you don't get to make breaking changes.

jacksontj:
I'm not making changes to the protocol at all, I'm simply passing in the *correct* time range. I'd really appreciate it if you'd take the minute to read over the example code etc. that I spent the time putting together to showcase the issue (which has been discussed repeatedly here) and clearly shows the problem with leaving that time range as is -- as it is *literally* the reason why this entire issue started.

juliusv:
Hey guys, I can see the heat going up a bit here.

I think on the changing-the-protocol front, @brian-brazil actually means what you are doing as changing the protocol because now the remote end does not get the overall query range anymore, so it's a change of semantics, even if no fields are new.

Personally I'm also totally for changing the remote protocol to support this as first-class fields, but maybe that should be done later (I think @brian-brazil would even be on board to add that properly to the protocol later, just not now?).

I'd also be for changing the glue interface to the TSDB within Prometheus to be better for remote storages (without changing the prometheus/tsdb interface itself).

jacksontj:
> I think on the changing-the-protocol front, @brian-brazil actually means what you are doing as changing the protocol because now the remote end does not get the overall query range anymore, so it's a change of semantics, even if no fields are new.

This would be fine to do, but this *would* require a protocol change as we'd need to pass both the overall range as well as the specific range for the select. I'm also okay to do that, but I was asked to *not* change that protocol, so we have to pick which one.

> Personally I'm also totally for changing the remote protocol to support this as first-class fields, but maybe that should be done later

My main question there is *when* is later, as discussed above it sounds like there are no API contracts that we need to maintain, so it seems like now would be the time to do it. I am also okay continuing with this path-- I don't like it as much since people can easily miss this whole thread and continue over-fetching, but at least it enables people who *do* find this thread to do the right thing.

> I'd also be for changing the glue interface to the TSDB within Prometheus to be better for remote storages (without changing the prometheus/tsdb interface itself).

The more I've dug into this the less I think we need a separate interface. A parsed Promql statement actually knows exactly what data (matchers over what ranges) it needs. In that case we don't need to split the interface-- we just need the interface to pass more context about the query (in this case the range of the matchers we are selecting, not the entire promql statement).

As discussed some above, I think maintaining the overall timerange is nice on the Querier level as that allows implementations (be it local or remote) to know what all Selects would potentially touch (this is important for sharded storage that is remote or even tsdb -- since then it knows what blocks it might touch, so it won't roll them off). The key is we need to be more explicit in the Select -- since promql knows exactly what it wants, there is no reason to leave the storage layer to guess based on lazy loading etc.

brian-brazil:
> (I think @brian-brazil would even be on board to add that properly to the protocol later, just not now?).

Yes, it's a logical thing to cleanup later. It's too bad that we didn't think of this earlier, but as we're still in an experimental phase we should avoid spurious breaking changes for the moment.

> I'd also be for changing the glue interface to the TSDB within Prometheus to be better for remote storages (without changing the prometheus/tsdb interface itself).

For now, as long as it doesn't change the remote read interface. No point in optimising further until we know what remote read is going to look like long term, otherwise it'd just be making work for remote read endpoint implementors while they're still largely starting out.

> this would require a protocol change as we'd need to pass both the overall range as well as the specific range for the select.

Adding hints is a change, but not a breaking change. It's expected that we'll add more hints over time as various remote read endpoint implementors want more information to help them optimise.

If you update codec.go and revert read.go, then that should be it.

juliusv:
@brian-brazil When do you think we'll have enough experience with remote storages using this particular interface to solidify some of this into proper fields, even breaking changes?

brian-brazil:
Hard to tell, we've only one system doing anything really fancy thus far. However there's also several vendors working on integrating, so we'll probably see more soon enough. If I had to guess I'd say we're talking 1-2 years. Providing a stable interface will help that process.

juliusv:
@jacksontj 

>> I'd also be for changing the glue interface to the TSDB within Prometheus to be better for remote storages (without changing the prometheus/tsdb interface itself).
>
> The more I've dug into this the less I think we need a separate interface. A parsed Promql statement actually knows exactly what data (matchers over what ranges) it needs. In that case we don't need to split the interface-- we just need the interface to pass more context about the query (in this case the range of the matchers we are selecting, not the entire promql statement).

I'm not proposing to create a new interface, there are already these two:

- The local TSDB implementation in `prometheus/tsdb` (ok, that's not an interface, it's a struct)
- The glue wrapper around it in `prometheus/prometheus`

I would only propose to change the latter to distinguish better between metadata and bulk data queries, as well as to have explicit fields for time ranges in selects (as I think you are proposing as well). It would then not put that extra information into its usage of the local TSDB, which can remain unchanged.

brian-brazil:
> I would only propose to change the latter to distinguish better between metadata and bulk data queries, as well as to have explicit fields for time ranges in selects (as I think you are proposing as well). It would then not put that extra information into its usage of the local TSDB, which can remain unchanged.

That sounds sane to me.

jacksontj:
> I would only propose to change the latter to distinguish better between metadata and bulk data queries, as well as to have explicit fields for time ranges in selects (as I think you are proposing as well). It would then not put that extra information into its usage of the local TSDB, which can remain unchanged.

I think this could be doable, but I would actually leave the interface the same and leave it up to the querier (TSDB or otherwise). A proposal querier interface (rough comments etc, but hopefully good enough to get the idea):

```
// Querier provides reading access to time series data.
type Querier interface {
        Select(start, end int64, *SelectParams, ...*labels.Matcher) (SeriesSet, error)

        // LabelValues returns all potential values for a label name.
        LabelValues(name string) ([]string, error)

        // Close releases the resources of the Querier.
        Close() error
}

type MetadataQuerier interface [
    Querier
    MetadataSelect(start, end int64, *SelectParams, ...*labels.Matcher) (SeriesSet, error)
}

```

The thought being that the promql implementation would check if the querier is a metadata one, if so it would call the `MetadataSelect` method instedad, otherwise it'd just fallback to current behavior. We could even make the `MetadataSelect` return a different interface (SeriesMetaSet or something) but thats probably unnecessary. Alternatively we could pass a flag in selectParams explicitly calling it out as a meta or data query.


To clarify a bit, none of this is a breaking change to the RPC api, this is simply a golang interface compatability change.

The discussion earlier about *which* time goes in the start/end of the RPC call is a bug-- not an API change. In the [rpc](https://github.com/prometheus/prometheus/blob/master/prompb/remote.pb.go#L99) a single start/end time pair is sent in the RPC request, which defines what timerange to return the data matching the associated matchers. Right now we are sending the range for the *entire* promql statement, not just the range required for that matcher -- hence it is a bug not a protocol change. As it stands today actually the remote read API has no concept of the "querier" times that we use locally.

juliusv:
> The thought being that the promql implementation would check if the querier is a metadata one, if so it would call the MetadataSelect method instedad, otherwise it'd just fallback to current behavior. We could even make the MetadataSelect return a different interface (SeriesMetaSet or something) but thats probably unnecessary. Alternatively we could pass a flag in selectParams explicitly calling it out as a meta or data query.

That'd work, although that'd require more type-checking work on the user side. If we changed the glue adapter https://github.com/prometheus/prometheus/blob/master/storage/tsdb/tsdb.go#L102-L106 to simply not pass on the optional extra information, then it'd all be contained in there and external users can always use the more explicit interfaces. Nicer IMO at least :)

> The discussion earlier about which time goes in the start/end of the RPC call is a bug-- not an API change. In the rpc a single start/end time pair is sent in the RPC request, which defines what timerange to return the data matching the associated matchers. Right now we are sending the range for the entire promql statement, not just the range required for that matcher -- hence it is a bug not a protocol change. As it stands today actually the remote read API has no concept of the "querier" times that we use locally.

I mostly agree with you. Currently any query with a long enough offset will basically not work, and most remote storage developers are probably not aware of this and might not become aware of a hint-based fix. But I also got the argument that if we spread multiple selectors of a single PromQL query over multiple `Query` messages in a single remote write request, that the other side might then not enable the same isolation / consistency guarantees as it might otherwise do for a single `Query` message. *However*, it seems that we are *already* doing this spreading in the current code? At https://github.com/prometheus/prometheus/blob/2952387ed1fcbb9d66d73eedba5a56838bf7cb51/storage/remote/read.go#L44-L58. So every vector/matrix selector in a PromQL query *already* becomes a separate remote read request. So either we fix that somehow, or it renders the argument of breakage on that front moot anyway?

brian-brazil:
> But I also got the argument that if we spread multiple selectors of a single PromQL query over multiple Query messages in a single remote write request, that the other side might then not enable the same isolation / consistency guarantees as it might otherwise do for a single Query message.

Consistency is not an issue no matter how the query is split currently. Isolation is a problem, but Prometheus doesn't support that yet so it's a moot point currently.

> So every vector/matrix selector in a PromQL query already becomes a separate remote read request. So either we fix that somehow, or it renders the argument of breakage on that front moot anyway?

That needs to be fixed at some point.

juliusv:
> Consistency is not an issue no matter how the query is split currently. Isolation is a problem, but Prometheus doesn't support that yet so it's a moot point currently.

Thanks, yeah. I'm a bit fuzzy on the DB terminology.

>> So every vector/matrix selector in a PromQL query already becomes a separate remote read request. So either we fix that somehow, or it renders the argument of breakage on that front moot anyway?
>
> That needs to be fixed at some point.

Ok, but it *does* mean that this PR doesn't actually do a breaking change (which was somewhat debatable before) because the remote end already doesn't get all PromQL selectors in one query?

brian-brazil:
> Ok, but it does mean that this PR doesn't actually do a breaking change (which was somewhat debatable before) because the remote end already doesn't get all PromQL selectors in one query?

It is a breaking change, as it changes the semantics of the time range which could then affect consistency.

brian-brazil:
Or more generally, any time you change the meaning of a field it is a almost always a breaking change.

juliusv:
> It is a breaking change, as it changes the semantics of the time range which could then affect consistency.

Ok, I get that, on a very nit-picky level - not that I think it'd currently affect any remote storage implementation. Since you seem to feel strongly about this, I'm ok with going hints-only for now, though not too happy with that. On the other hand, I would also not have been super happy with just changing the passed time ranges as the PR currently suggests because that ties us to having multiple `Query` message, which even if they were all part of the same read request (that code needs to be fixed at some point), would still not be great because the original (implied?) intention of the protocol was that a single `Query` message would handle all data for a single PromQL query.

brian-brazil:
> would still not be great because the original (implied?) intention of the protocol was that a single Query message would handle all data for a single PromQL query.

I don't believe that was ever the intention, a Query only has one set of matchers. Digging through the archives confirms this, it was a todo to use ReadRequest for an entire PromQL query.

jacksontj:
> That'd work, although that'd require more type-checking work on the user side. If we changed the glue adapter https://github.com/prometheus/prometheus/blob/master/storage/tsdb/tsdb.go#L102-L106 to simply not pass on the optional extra information, then it'd all be contained in there and external users can always use the more explicit interfaces. Nicer IMO at least :)

I don't think I follow how this makes more type-checking on the user side? The type check would be done in the promql package. Meaning you the user would still create a struct that implements the `Querier` interface, and if you *also* have a `MetadataSelect` method you'll get called there for metadata calls instead of the main one. Definitely possible I'm missing something though, but as I see it there would be no requirement for type-checking on the user-side. Alternatively to doing this separate interface we could add it to the selectParams-- which seems to be where all these things are ending up.

> I don't believe that was ever the intention, a Query only has one set of matchers. Digging through the archives confirms this, it was a todo to use ReadRequest for an entire PromQL query.

Given that I think the mapping should be that each `Query` has its start/end time (for the specific matcher) and the overarching `Query` struct can have a mint/maxt. That way the `ReadRequest` == `Querier()` and the `Query` == `Select()`

> It is a breaking change, as it changes the semantics of the time range which could then affect consistency.

It doesn't change the semantics of the time range, it simply passes in the correct time. The name of the field is "start_timestamp" (meaning the beginning of the query we are asking for) not "mintimems" (the min time of the higher-level promql statement). Right now it fetches *more* data than it needs or wants and simply ignores the excess data.

In reality this is 2 things, which is why I have it in 2 commits. (1) pass the start/end time through the selectParams and (2) correctly map the selectParams to the remote-read API

brian-brazil:
> Given that I think the mapping should be that each Query has its start/end time (for the specific matcher) and the overarching Query struct can have a mint/maxt. 

That's making an assumption, a single ReadRequest could in principle be handling multiple PromQL queries such as fetching all the data for a rule group at once.

> It doesn't change the semantics of the time range, it simply passes in the correct time.

I don't see how you can view a change from passing the promql query range to passing the selector range as not being a change in semantics. The current protocol may not be optimal, but I don't see anything about it that is incorrect. It will produce the correct result for the end user, and changing it could produce an incorrect result due to the change in semantics.

> The name of the field is "start_timestamp" (meaning the beginning of the query we are asking for) not "mintimems" (the min time of the higher-level promql statement).

That's supposition given the lack of documentation in either direction. The protocol has been out there for over a year now so whatever it currently does is now, one way or the other, the expected behaviour.

juliusv:
@brian-brazil 

>> would still not be great because the original (implied?) intention of the protocol was that a single Query message would handle all data for a single PromQL query.
>
> I don't believe that was ever the intention, a Query only has one set of matchers. Digging through the archives confirms this, it was a todo to use ReadRequest for an entire PromQL query.

Ah ok, then I misunderstood your comments about intentions at the beginning.

@jacksontj 

> I don't think I follow how this makes more type-checking on the user side? The type check would be done in the promql package.

Depending on who you see as the user of what package. I meant from the perspective of internal users in Prometheus, which is both PromQL and the web API. But I guess it's a question of transparently hiding an implementation detail or making it visible to the user of that package (`github.com/prometheus/prometheus/storage/`)... where I like the former more at the moment.

jacksontj:
@bbrazil

> That's making an assumption, a single ReadRequest could in principle be handling multiple PromQL queries such as fetching all the data for a rule group at once.

Sure, anything is possible in the future-- but right now we never do more than a single query in a read request-- so really all the conversations around remote read are seemingly useless (since we keep putting restrictions on it even though nothing yet needs any of those restrictions). If we want to continue passing 2 time ranges around **AND** we want to put arbitrary messages together in a remote-read request then we can add more fields to the query itself. Seeing as this stuff is basically undocumented, its effectively impossible to contribute -- as rules are springing up out of nowhere ;)

> The current protocol may not be optimal, but I don't see anything about it that is incorrect.

So to make sure I'm understanding your position correctly-- you are stating that the remote-read API is intended to all points for each matcher for the *entire* promql query range -- even if the promql statement doesn't need it? If so then you are correct -- we'd be breaking the API. However if the intention and usage of the remote-read API is to get data for matchers at a time range (which is what is *doing* and what things are using it for) then this isn't a breaking change, we're simply only asking for the data we need instead of all the data we need and somtimes lots more.

> That's supposition given the lack of documentation in either direction. The protocol has been out there for over a year now so whatever it currently does is now, one way or the other, the expected behaviour.

I'm not sure this makes sense. Right now as a remote-read storage layer, I get a query for a matcher and a time-range. Right now people just respond with all the points in that time range. I'd also like to point out here that *prometheus itself* does this, if you send a remote-read query for matcher and timerange it will return *all* points in that range (meaning the behavior of prometheus itself matches what I am describing).



@juliusv
> But I guess it's a question of transparently hiding an implementation detail or making it visible to the user of that package (github.com/prometheus/prometheus/storage/)... where I like the former more at the moment.

Yea, my expectation is that this would be hidden entirely in the promql package, so everyone else still calls `InstantQuery` (or whatever) and passes in their storage engine, and promql internally will pick which method to call.

juliusv:
>> The current protocol may not be optimal, but I don't see anything about it that is incorrect.
>
> So to make sure I'm understanding your position correctly-- you are stating that the remote-read API is intended to all points for each matcher for the entire promql query range -- even if the promql statement doesn't need it? If so then you are correct -- we'd be breaking the API. However if the intention and usage of the remote-read API is to get data for matchers at a time range (which is what is doing and what things are using it for) then this isn't a breaking change, we're simply only asking for the data we need instead of all the data we need and somtimes lots more.

If I understood @brian-brazil correctly, then this is explained by this comment and the following ones where I ask for clarification: https://github.com/prometheus/prometheus/pull/4226#issuecomment-395376397

That is, if the remote end knows the overall query range, then it can also choose different downsampling rules, and those should be the same among selectors of a single PromQL query.

I do not think that anyone is currently doing that, so I think in reality the effect of "breaking" this would be zero, but we can't be sure.

> Yea, my expectation is that this would be hidden entirely in the promql package, so everyone else still calls InstantQuery (or whatever) and passes in their storage engine, and promql internally will pick which method to call.

The web API will also need to deal with the extra interface methods for metadata queries. IMO it's still nicer to give people outside of the `storage` package a single interface and then leave it up to implementations to do things efficiently (TSDB via lazy loading, remote storage via including extra time info in sub-queries).

jacksontj:
> That is, if the remote end knows the overall query range, then it can also choose different downsampling rules, and those should be the same among selectors of a single PromQL query.

Well, this is only partially true. Right now it actually *only* sends the one time range which is the entire promql range-- not the overall promql range. Actually as I'm thinking about it I don't think the overall promql range matters when downsampling the specific read, as it doesn't change how you'd respond. In all cases I can think of you'll be getting points from the offsets in the query, but each selection will be for the same duration of time (just at different offset start points). In addition, we'd need something *else* to control downsampling, because if we allow remote read endpoints to just downsample as they wish we can't make any guarantees about correctness of the data? I think thats a much larger discussion (which I've attempted in the [past](https://github.com/prometheus/prometheus/pull/3631) -- but we should probably hold out of this conversation to keep the conversation focused).

> The web API will also need to deal with the extra interface methods for metadata queries. IMO it's still nicer to give people outside of the storage package a single interface and then leave it up to implementations to do things efficiently (TSDB via lazy loading, remote storage via including extra time info in sub-queries).

I think we're conflating a couple things here. The "extra interface" would only be for people that are using whatever implements the `Querier()` interface (presumably just promql, but others could use it as well). As for the lazy-loading vs exact matching stuff, I think thats separate from the metadata discussion, as thats all time-range questions. And as far as the API is concerned we have a specific thing we want, so we should (IMO) ask for exactly what we want instead of relyin on lazy loading hueristics to get *close* to what we want.

jacksontj:
Friendly bump (seems that this PR might have been lost in the last week) :)

brian-brazil:
My previous comments still stand. You need to remove the changes to read.go, and update codec.go.

jacksontj:
I have updated the PR to continue sending the wrong time range in the outer-layer read request, and send the correct one as a ReadHint. While I was in there I also made it so that the remote read API will properly unmarshal the `ReadHint` to a `SelectParam` (since it was being dropped entirely before).

prompb/types.proto
```diff
@@ -53,4 +53,6 @@ message LabelMatcher {
 message ReadHints {
   int64 step_ms = 1; // Query step size in milliseconds.
   string func = 2;   // String representation of surrounding function or aggregation.
+  int64 start = 3;   // Start time in milliseconds.
```
brian-brazil: Let's do start_ms for consistency and clarity.

vendor/vendor.json
```diff
@@ -383,12 +383,6 @@
 			"version": "=v1.5.4",
 			"versionExact": "v1.5.4"
 		},
-		{
-			"checksumSHA1": "FhLvgtYfuKY0ow9wtLJRoeg7d6w=",
-			"path": "github.com/gogo/protobuf/gogoproto",
```
brian-brazil: Why is this getting removed?

jacksontj: @brian-brazil details for that are in the commit (https://github.com/prometheus/prometheus/pull/4226/commits/a7528730306bff87d60cb538a5db54c5e37f4126) TLDR; the script that codegens out the proto files changed to not include that anymore, so I removed it. If this *should* be in use, then there is a bug in `genproto.sh` as it quite literally seds it out (https://github.com/prometheus/prometheus/blob/master/scripts/genproto.sh#L43)

jacksontj: @brian-brazil bump :)

brian-brazil: Usually I would have already merged your change as your explanation makes sense, however we're still in moratorium.

jacksontj: Oh, didn't realize there was a moratorium. So if we're all set here from a code perspective then I'll just wait for merge. I mostly wanted to make sure there wasn't something outstanding to do on this :)

brian-brazil:
Thanks!

promql/engine.go
```diff
@@ -481,12 +481,19 @@ func (ng *Engine) populateSeries(ctx context.Context, q storage.Queryable, s *Ev
 	Inspect(s.Expr, func(node Node, path []Node) error {
 		var set storage.SeriesSet
 		params := &storage.SelectParams{
-			Step: int64(s.Interval / time.Millisecond),
+			Start: timestamp.FromTime(s.Start.Add(-LookbackDelta)),
```
free: With apologies for being late to the party, I only noticed this change when I was fixing some merge conflicts: why are we now loading `LookbackDelta` worth of data before range start for matrix selectors?

I was doing the same in my (thus far) ill fated `xrate` attempt and there was a noticeable performance penalty to loading an extra 5 minutes of data for every single matrix query when only `xrate` needed it. IIRC, it was around 10%, which is non-negligible. AFAIK no existing function actually uses any samples outside of the range interval.

free: You're right, of course. My changes were to the `Inspect` call above, where `mint` is calculated, not this one. Probably worth fixing, though.

brian-brazil: Yes, thanks for catching this. Would you like to send a PR?

Issue #968, opened 2015-08-06
Option for immediate deletion of obsolete series on startup


aecolley:
Prometheus needs a startup flag or HTTP endpoint to trigger an immediate maintenance sweep to delete old files from local storage.

Background: when disk usage hits 100% (usually due to bad planning), a reasonable recovery strategy is to reduce the storage.local.retention value and restart Prometheus. There are two problems with this strategy. First, Prometheus can't make any progress if it can't write to new files. Second, local/storage.go waits for 10% of the retention duration before it begins the first maintenance sweep. This feature request is about the second problem, not the (more difficult) first.

Advising people to make free with rm under the prometheus storage dir makes me queasy, even if it's safe to remove the older series files. For operations work, we need a less-risky procedure for the sleep-deprived pager-carrier who is facing a full disk and a stuck Prometheus.

I want to say: stop Prometheus; delete the orphaned directory to clear up some space; then start it up with a shorter storage.local.retention; and POST to /force-maintenance-sweep; then watch the disk usage drop.

Possible alternatives that I can think of: using a new storage.local.maxbytes limit to stop the problem arising in the first

aecolley:
(oopsie, silly mobile interface, making the submit button so big and so close to the comment box)

place; or using a "minfree" parameter to tell Prometheus to keep an eye on the filesystem free space and not to get too close to full.

fabxc:
I could imagine extending our API deletion endpoint to delete certain ranges of time series or simpler older-than samples. This way one could manually and explicitly cut off old time series and then restart with a new retention time. Ideally this would be available through `promtool`, too.
Certainly cleaner than to adjust the retention time and then wait for an implementation detail to kick in.

This would still need added support from the storage layer, so @beorn7 probably has an opinion on this. He is on vacation though.

beorn7:
Thinking about this, I'm not sure if catering for the "disk completely full" case makes a lot of sense. Experience tells us that the storage is often in very sorry state if Prometheus runs into a disk-full scenario. You probably cannot shut down cleanly in that state, and the crash recovery will need its own share of disk, so an immediate deletion of time series will only help in the cases where you have just enough disk space for crash recovery but not enough to wait for normal purging...

I was thinking a couple of time about a "storage tool", i.e. a stand alone command line tool that can be used to manipulate and analyze the on-disk storage in cold state.

[...]

PR #4230, opened 2018-06-06
Added flags for size based retention

mknapphrt:
Added a new flags for setting maximum number of bytes that can be stored locally and necessary fields in tsdb for storing it. This PR is in conjunction with a PR with prometheus/tsdb. The idea is to allow a user to set a maximum number of bytes that the local storage can take up, in conjunction with the retention period, and whenever that limit is exceeded the oldest block will be deleted to regain space.

closes https://github.com/prometheus/prometheus/issues/968

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -163,6 +163,9 @@ func main() {
 	a.Flag("storage.tsdb.retention", "How long to retain samples in storage.").
 		Default("15d").SetValue(&cfg.tsdb.Retention)
 
+	a.Flag("storage.tsdb.maxbytes", "Maximum number of bytes that can be stored for blocks.").
+		Default("0").Int64Var(&cfg.tsdb.MaxBytes)
+
 	a.Flag("storage.tsdb.no-lockfile", "Do not create lockfile in data directory.").
 		Default("false").BoolVar(&cfg.tsdb.NoLockfile)
 
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -160,10 +160,13 @@ func main() {
 		"Maximum duration compacted blocks may span. For use in testing. (Defaults to 10% of the retention period).").
 		Hidden().PlaceHolder("<duration>").SetValue(&cfg.tsdb.MaxBlockDuration)
 
-	a.Flag("storage.tsdb.retention", "How long to retain samples in storage.").
+	a.Flag("storage.tsdb.retention", "How long to retain samples in storage. (This flag has been deprecated. Use storage.tsdb.retention.time as a replacement.)").
 		Default("15d").SetValue(&cfg.tsdb.Retention)
 
-	a.Flag("storage.tsdb.maxbytes", "Maximum number of bytes that can be stored for blocks.").
+	a.Flag("storage.tsdb.retention.time", "How long to retain samples in storage.").
+		Default("15d").SetValue(&cfg.tsdb.Retention)
+
+	a.Flag("storage.tsdb.retention.size", "Maximum number of bytes that can be stored for blocks.").
 		Default("0").Int64Var(&cfg.tsdb.MaxBytes)
 
 	a.Flag("storage.tsdb.no-lockfile", "Do not create lockfile in data directory.").
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -52,7 +52,8 @@ For further details on file format, see [TSDB format](https://github.com/prometh
 Prometheus has several flags that allow configuring the local storage. The most important ones are:
 
 * `--storage.tsdb.path`: This determines where Prometheus writes its database. Defaults to `data/`.
-* `--storage.tsdb.retention`: This determines when to remove old data. Defaults to `15d`.
+* `--storage.tsdb.retention.time`: This determines when to remove old data. Defaults to `15d`.
+* `--storage.tsdb.retention.size`: This determines the maximum number of bytes that storage blocks can use (note that this does not include the WAL size, which can be substantial). The oldest data will be removed first. Defaults to `0` or disabled.
 
 On average, Prometheus uses only around 1-2 bytes per sample. Thus, to plan the capacity of a Prometheus server, you can use the rough formula:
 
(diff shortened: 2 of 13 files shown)


mknapphrt:
https://github.com/prometheus/tsdb/pull/343

mknapphrt:
Uhh... I was trying to rebase and i'm not sure why everything that's been done is showing up as a change right now.

krasi-georgiev:
before I start the benchmarking test lets make the unit tests pass.

mknapphrt:
@krasi-georgiev So this passes right now. I just squashed and fixed a merge conflict on the tsdb pr, I'll start playing around with testing it

krasi-georgiev:
thanks , I will run a bench test and will update the configs so we can both monitor that it works as expected.

krasi-georgiev:
need to change few thins in the tests to be able to run it with custom retention policy.

btw I think the flag should be called `storage.tsdb.retention.size` and maybe at some point rename the time retention to `storage.tsdb.retention.time` or `storage.tsdb.retention.duration`

krasi-georgiev:
ok now the prometheus running at  prombench.prometheus.io/4230/prometheus-pr has a size retention set at 1gb an it is monitored by the meta isntance at  http://prombench.prometheus.io/prometheus-meta

krasi-georgiev:
one thing we overlooked is the size of the WAL which can be substantial. It grows very rapidly and haven't looked into the code yet if we can add this to the size retention policy, but should have it mind.

krasi-georgiev:
@mknapphrt also need to update the docs in https://github.com/prometheus/prometheus/blob/master/docs/storage.md

krasi-georgiev:
@juliusv, @brancz , @simonpasquier  	now that we would have time and size retention I was thinking to name the flags. `storage.tsdb.retention.size` , `storage.tsdb.retention.time` (renamed from `storage.tsdb.retention`) . Since this is a breaking change , what is the procedure here?

gouthamve:
A breaking change in flags would mean `prometheus 3.0`, if I'm not wrong. We can mark it deprecated now, and should only remove it later.

krasi-georgiev:
ooh yeah that is a good idea, lets mark it deprecated.

mknapphrt:
So I originally intended to include the WAL but but brian brazil suggested starting with just the blocks because the WAL calculation could be trickier to implement and would also include a lot more overhead. It's certainly something to be pursued though. I'll update the names and docs and ping you when i'm done.

mknapphrt:
@krasi-georgiev Let me know what you think

cmd/prometheus/main.go
```diff
@@ -160,9 +160,15 @@ func main() {
...
 
-	a.Flag("storage.tsdb.retention", "How long to retain samples in storage.").
+	a.Flag("storage.tsdb.retention", "How long to retain samples in storage. (This flag has been deprecated. Use storage.tsdb.retention.time as a replacement.)").
 		Default("15d").SetValue(&cfg.tsdb.Retention)
 
+	a.Flag("storage.tsdb.retention.time", "How long to retain samples in storage.").
+		Default("15d").SetValue(&cfg.tsdb.Retention)
+
+	a.Flag("storage.tsdb.retention.size", "Maximum number of bytes that can be stored for blocks.").
+		Default("0").Int64Var(&cfg.tsdb.MaxBytes)
```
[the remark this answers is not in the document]
krasi-georgiev: which makes me think that the options should be called something like.
cfg.tsdb.Retention.Time
cfg.tsdb.Retention.Size

cmd/prometheus/main.go
```diff
@@ -160,9 +160,15 @@ func main() {
 		"Maximum duration compacted blocks may span. For use in testing. (Defaults to 10% of the retention period).").
 		Hidden().PlaceHolder("<duration>").SetValue(&cfg.tsdb.MaxBlockDuration)
 
-	a.Flag("storage.tsdb.retention", "How long to retain samples in storage.").
+	a.Flag("storage.tsdb.retention", "How long to retain samples in storage. (This flag has been deprecated. Use storage.tsdb.retention.time as a replacement.)").
```
krasi-georgiev:
```suggestion
	// [DEPRECATED] use storage.tsdb.retention.time instead.
	a.Flag("storage.tsdb.retention", "How long to retain samples in storage. (This flag has been deprecated. Use storage.tsdb.retention.time as a replacement.)").
```

this will allows to do a global search and remove the obsolete code.

cmd/prometheus/main.go
```diff
@@ -160,9 +160,15 @@ func main() {
...
 		Default("15d").SetValue(&cfg.tsdb.Retention)
 
+	a.Flag("storage.tsdb.retention.time", "How long to retain samples in storage.").
+		Default("15d").SetValue(&cfg.tsdb.Retention)
+
+	a.Flag("storage.tsdb.retention.size", "Maximum number of bytes that can be stored for blocks.").
```
krasi-georgiev: as @gouthamve 	mentioned on irc lets mark this experimental to warn people and reduce bugfix urgency. again using some midleware pattern to show a warning when starting Prometheus with this flag and not just a comment in the code.

krasi-georgiev:
@simonpasquier 	do you know why is this test failing?

here is how we ended up here.

I opened a PR against mknapphrt:size_based_retention https://github.com/mknapphrt/prometheus/pull/2

after the merge this test started failing.

before the merge this PR was passing all tests and also my PR in https://github.com/mknapphrt/prometheus/pull/2 also passes all tests(when run locally). So two PRs with passing tests were merged which resulted in a code that fails the `make unused` test.

mknapphrt:
@krasi-georgiev Sorry I haven't been responsive, I've been busy while the semester is wrapping up. What needs to be done, right now there's merge conflicts on here and tsdb and I'm still trying to figure out what's going on with the unused test. I think the merge conflicts on the tsdb need to get resolved first because they're causing a lot of the ones here. Though you've been touching the 343-orig branch so I'm not sure if you want me to fix the conflicts or if you want to do it?

krasi-georgiev:
@mknapphrt no problem , I have fixed the conflicts and updated the tsdb PR and will ask Goutham to have a final look. We have done enough iterations and proved that the implementation in Prometheus works as expected so once this is merged in tsdb will continue the work on this PR.

krasi-georgiev:
will continue the work in https://github.com/prometheus/prometheus/pull/5109

PR #4232, opened 2018-06-07
Wrong result with rate/irate functions.

gywndi:
Hi~

Promethues painted a strange graph with rate/irate funciton. So I checked data, and find out it decreased at a certain point, as below.
```
512408305 @1525061864.282 
512410956 @1525061874.282 
512417456 @1525061884.282 <== greater? 
512416256 @1525061894.282 <== less? 
512418830 @1525061904.282 
512421378 @1525061914.282 
512423749 @1525061924.282 
```

I have confirmed that this is a problem caused by adding the data just prior to the decline to correct the data. The data correction method is implemented differently, and the graph is shown normally.

In the case of irate, we changed the loop to calculate the data until the correct result is shown.
I attached the graph before and after the change.

Thanks.
Chan.
## Before ##
before
## After ##
after

### DIFF
--- a/promql/functions.go
+++ b/promql/functions.go
@@ -77,16 +77,21 @@ func extrapolatedRate(vals []Value, args Expressions, enh *EvalNodeHelper, isCou
 			continue
 		}
 		var (
-			counterCorrection float64
-			lastValue         float64
+			lastValue       float64
+			resultValue     float64
+			startPointValue float64
 		)
+
+		// Correct count data which reset or reduced
+		startPointValue = samples.Points[0].V
 		for _, sample := range samples.Points {
 			if isCounter && sample.V < lastValue {
-				counterCorrection += lastValue
+				resultValue += (lastValue - startPointValue)
+				startPointValue = sample.V
 			}
 			lastValue = sample.V
 		}
-		resultValue := lastValue - samples.Points[0].V + counterCorrection
+		resultValue += lastValue - startPointValue
 
 		// Duration between first/last samples and boundary of range.
 		durationToStart := float64(samples.Points[0].T-rangeStart) / 1000
@@ -167,35 +172,41 @@ func instantValue(vals []Value, out Vector, isRate bool) Vector {
 	for _, samples := range vals[0].(Matrix) {
 		// No sense in trying to compute a rate without at least two points. Drop
 		// this Vector element.
-		if len(samples.Points) < 2 {
-			continue
-		}
-
-		lastSample := samples.Points[len(samples.Points)-1]
-		previousSample := samples.Points[len(samples.Points)-2]
 
 		var resultValue float64
-		if isRate && lastSample.V < previousSample.V {
-			// Counter reset.
-			resultValue = lastSample.V
-		} else {
+		var sampledInterval int64
+
+		for i, _ := range samples.Points {
+			if len(samples.Points) < i+2 {
+				break
+			}
+
+			lastSample := samples.Points[len(samples.Points)-(i+1)]
+			previousSample := samples.Points[len(samples.Points)-(i+2)]
+
+			// Correct count data which reset or reduced
+			if isRate && lastSample.V < previousSample.V {
+				continue
+			}
 			resultValue = lastSample.V - previousSample.V
-		}
+			sampledInterval = lastSample.T - previousSample.T
 
-		sampledInterval := lastSample.T - previousSample.T
-		if sampledInterval == 0 {
-			// Avoid dividing by 0.
-			continue
-		}
+			if sampledInterval == 0 {
+				// Avoid dividing by 0.
+				continue
+			}
 
-		if isRate {
-			// Convert to per-second.
-			resultValue /= float64(sampledInterval) / 1000
-		}
+			if isRate {
+				// Convert to per-second.
+				resultValue /= float64(sampledInterval) / 1000
+			}
 
-		out = append(out, Sample{
-			Point: Point{V: resultValue},
-		})
+			out = append(out, Sample{
+				Point: Point{V: resultValue},
+			})
+
+			break
+		}
 	}
 	return out
 }


gywndi:
Hi Brian

Thanks for your message.

I do agree with your mention "Counters only go up and reset". This must be.

However, mysql still has a bug that shows incorrect results. You can check the contents in the bottom bug report.
>> https://bugs.mysql.com/bug.php?id=90351

The Linux OS also shows reduced count. (Not often, but somtimes)
Sometimes, an alarm is triggered by incorrect OS utilization detection like network usage.

Do you have a good idea on how to properly refine your data in this situation?

If it is a broken count, it may be correct to drop it. However, the logic to correct the data in the RESET case seems to make more sense, so I have only slightly changed the way we do calculations. The calculation algorithm does not have a major effect on the current result and will guarantee the data even in the case of the broken counts mentioned above.

I think that prometheus becomes more stable.
Tell me what might be a problem.

Best Regards,
Chan.

brian-brazil:
> Do you have a good idea on how to properly refine your data in this situation?

You might use something like clamp_max.


> The calculation algorithm does not have a major effect on the current result and will guarantee the data even in the case of the broken counts mentioned above.

Doing so will break working counters though.

gywndi:
> You might use something like clamp_max.
It's the same story that I can not use rate or irate.

> Doing so will break working counters though.
The rate result has already spoiled the results. It is more reasonable to calculate the resultValue in a drop or stepwise fashion than a ridiculous error in rate as my code.

gywndi:
> You might use something like clamp_max.

It's the same story that I can not use rate or irate.

> Doing so will break working counters though.

The rate result has already spoiled the results. It is more reasonable to calculate the resultValue in a drop or stepwise fashion than a ridiculous error in rate as my code.

brian-brazil:
> The rate result has already spoiled the results. It is more reasonable to calculate the resultValue in a drop or stepwise fashion than a ridiculous error in rate as my code.

As far as Prometheus is concerned your counter reset and then increased by 512416256 in the space of 10 seconds. We have no way to tell if your counter really had such a reset and spike, or if your counter is broken. We have to presume that your counter is working correctly, otherwise everything falls apart and we'd lose data on frequently resetting counters.

I'm sorry that you have a broken counter, but we can't break other users in order to workaround your breakage.

If you wish to do custom counter logic, the idelta function and recording rules allow for that.

gywndi:
The proposed code does not affect RESET.
In the case of COUNT, the case where the data is decremented should never occur except reset. If spike, data will never decrease. I think that the data of other people seems to be broken.

Prometheus is used to monitor mysql, but before mysql bugs are fixed, I can not help but see a lot of false positives.

With this case
```
512408305 @1525061864.282 
512410956 @1525061874.282 
512417456 @1525061884.282 <== greater? 
512416256 @1525061894.282 <== less? 
512418830 @1525061904.282 
512421378 @1525061914.282 
512423749 @1525061924.282 
```

Original result is
```
512423749 - 512408305 + 512417456
```
Fixed result is
```
512408305 @1525061864.282 
512410956 @1525061874.282 
512417456 @1525061884.282
resultValue += (512417456 - 512408305)

512416256 @1525061894.282 <== less? 
512418830 @1525061904.282 
512421378 @1525061914.282 
512423749 @1525061924.282 
resultValue += (512423749 - 512416256)

=> (512417456 - 512408305) + (512423749 - 512416256)
```
Do you really this this is messing up the data? The original is rather strange.

gywndi:
With RESET case
```
512408305 @1525061864.282 
512410956 @1525061874.282 
512417456 @1525061884.282 <== greater? 
.. loss ..                <== reset
10 @1525061894.282 
40 @1525061904.282 
50 @1525061914.282 
```
Original result is
```
50 - 512408305 + 512417456 = 9201
```
Fixed result is
```
(512417456 - 512408305) + (50 - 10) = 9191
```
It has some influence because it can not know the exact time to reset.
However, original data is not correct because it is not known whether the value immediately before reset is the maximum value at the correct time.

brian-brazil:
> The proposed code does not affect RESET.

The whole point of your code is to change how counter resets are handled. In your final example there it's failing to count the 10 events immediately after the reset which would be a regression. 9201 is what that should be returning.

gywndi:
You think original result is correct one?
In fact, prometheus does not know the exact time to reset.
When a reset occurs, prometheus also estimates the data but does not output the correct result.

For positive processing, all cases where counts are decreasing should be dropped.

As I said above, this is currently a bug in mysql, and I can not guarantee that it never happens on linux metric, cpu/network.

gywndi:
Prometheus rate function computes with the data already lost from the last maximum value to the reset point

brian-brazil:
> As I said above, this is currently a bug in mysql, and I can not guarantee that it never happens on linux metric, cpu/network.

A bug in MySQL should be fixed there, and bugs in any other code should be fixed there too. Rate is always fundamentally an estimation, and your proposal makes our estimate worse so I'm afraid we cannot accept it.

gywndi:
> fundamentally an estimation

The code I present, and the code of the current prometheus, is also an estimation.
It is only an aim to find an accurate approximation. In the example given above, if you are concerned about the number 10, add the logic to add 10. (If so, let's add this part.)

It is not understood to be an intended result to be able to input data in an uncleaned form, but to show the calculation results in nonsense numbers.

Whether the count value can not be reduced or whether it is fixed the result, Prometheus also thinks it needs modification. In the current situation, if I use rate / irate, not only mysql, but linux is also suitable for monitoring. The delta is likely to show more missing data.

Think about what is better suited to your service from showing proper monitoring results.

brian-brazil:
> In the example given above, if you are concerned about the number 10, add the logic to add 10. (If so, let's add this part.)

We already have that logic, your PR proposes to remove it.

We're going around in circles here, I'm afraid your PR is not going to be accepted as it breaks existing semantics.

gywndi:
> We already have that logic, your PR proposes to remove it.

I've seen it included in the logic. However, there is another bug that shows the wrong data results in the wrong data entry. I do not think I completely broke the data correction. If necessary, you can think of a way to skip a slight data reduction.

I think that prometheus is not suitable for intermittent monitoring of mysql and linux, which is currently a logic, and which may be a fix at this time.

It is better to drop data instead of correcting it. I am talking about service improvement points, and I understand that prometheus is perfect, so just put in clean data by application level.

brian-brazil:
For the reasons I've already indicated, I cannot accept this PR.

PR #4233, opened 2018-06-07
Update make test

SuperQ:
* Make `test` without `-race` to avoid OOMing.
* Add separate `test-race` with `-race`.
* Add `test` to circleci build.


brian-brazil:
:+1:

SuperQ:
`test` still fails with OOM.

PR #4234, opened 2018-06-07
Makefile: update .PHONY target

simonpasquier:
As spotted by @mxinden in https://github.com/prometheus/alertmanager/pull/1396, the `govendor` target doesn't exist in Makefile.common (as well as `assets` which is declared in Makefile). On the other hand, `unused` and `test-short` were missing.

cc @krasi-georgiev

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -97,4 +97,4 @@ $(FIRST_GOPATH)/bin/staticcheck:
 $(FIRST_GOPATH)/bin/govendor:
 	GOOS= GOARCH= $(GO) get -u github.com/kardianos/govendor
 
-.PHONY: all style check_license format build test vet assets tarball docker promu staticcheck $(FIRST_GOPATH)/bin/staticcheck govendor $(FIRST_GOPATH)/bin/govendor
+.PHONY: all style check_license format build test-short test vet tarball docker promu unused staticcheck $(FIRST_GOPATH)/bin/staticcheck $(FIRST_GOPATH)/bin/govendor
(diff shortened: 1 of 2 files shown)


Makefile.common
```diff
@@ -97,4 +97,4 @@ $(FIRST_GOPATH)/bin/staticcheck:
 $(FIRST_GOPATH)/bin/govendor:
 	GOOS= GOARCH= $(GO) get -u github.com/kardianos/govendor
 
-.PHONY: all style check_license format build test vet assets tarball docker promu staticcheck $(FIRST_GOPATH)/bin/staticcheck govendor $(FIRST_GOPATH)/bin/govendor
\ No newline at end of file
+.PHONY: all style check_license format build test-short test vet tarball docker promu unused staticcheck $(FIRST_GOPATH)/bin/staticcheck $(FIRST_GOPATH)/bin/govendor
```
krasi-georgiev: why do we need both? `test-short test`

simonpasquier: because both targets exist

https://github.com/prometheus/prometheus/blob/0e2aa35771b23c9d65681f3b19b5a0421a206ec5/Makefile.common#L56-L63

krasi-georgiev: ooh yes , sorry I had a different understanding of what `.PHONY` targets are.

krasi-georgiev:
LGTM

krasi-georgiev:
@SuperQ PTAL

simonpasquier:
> Should we move all the PHONY entries to be listed with their targets?

That would definitely be clearer/cleaner from my POV.

PR #4236, opened 2018-06-07
Information on how to write case-insensitive regex in label matchers

avonwyss:


docs/querying/basics.md
```diff
@@ -96,7 +96,7 @@ For example, this selects all `http_requests_total` time series for `staging`,
     http_requests_total{environment=~"staging|testing|development",method!="GET"}
 
 Label matchers that match empty label values also select all time series that do
-not have the specific label set at all. Regex-matches are fully anchored.
+not have the specific label set at all. Regex-matches are fully anchored.
```
simonpasquier: (nit) the extra space isn't necessary.

docs/querying/basics.md
```diff
@@ -116,6 +116,10 @@ The following expression selects all metrics that have a name starting with `job
 
     {__name__=~"job:.*"}
 
+All label matchers are case-sensitive. To perform a case-insensitive match, use a 
+regex with the `(?i)` option at the beginning. Partial case-insensivity can be 
+achieved with the `(?i:pattern)` group.
```
simonpasquier: You could add examples as for the other paragraphs.

```
{method=~"(?i)get"} # case-insensitive match on "get"
{method=~"(?i:g)et"} # match on "get" or "Get"
```

brian-brazil: I'm not sure it's appropriate to add niche details on how regexes work, particularly as we don't even cover the basic syntax here. I think this is already covered by the link just below.

avonwyss:
@simonpasquier I had explicit samples but I found it to be too verbose for the importance it has. Basically I wanted to clarify that the comparison is by default always case-sensitive (which is not documented) and how case-insensivity can be achieved (which people usually do with a regex flag, the option groups are not a well-known regex construct - see also https://groups.google.com/forum/#!topic/prometheus-users/xOeIocdR_hI ).

Feel free to discard and close the PR if you think it is not valuable to users.

brian-brazil:
One thing we could do is the current link on RE2 details is to a reference doc, which is fine if you're already familiar with regexes but not really for everyone else. Maybe the thing to do it add a link to some tutorial or something to get beginners to regular expressions going?

juliusv:
> Maybe the thing to do is to add a link to some tutorial or something to get beginners to regular expressions going?

Yeah, I think that's the better way to go. Though I have to admit I didn't even know that RE2 was case-insensitive by default.

brian-brazil:
It's case sensitive by default.

juliusv:
Ah misread, thanks. Then it makes even more sense to not document this detail here IMO.

gouthamve:
Thanks for submitting this PR! We already link to the RE2 reference right below and we don't think we should have intricate regular expression details in this part of the documentation. This could be part of a guide but not part of the reference docs.

PR #4239, opened 2018-06-07
limit size of POST requests against remote read endpoint

### LINKED ISSUE #4238: Remote read endpoint should limit the size of the HTTP Post data it reads
## Bug Report
**What did you do?**
I'm able to kill a prometheus server. I've read the Prometheus [security model](https://prometheus.io/docs/operating/security/) and I can violate:
> It is more likely that a component will be accidentally taken out by a trusted user than by malicious action.

So I reached out on [twitter](https://twitter.com/_aead_/status/1004404417267986433) to ask how to report a security issue for Prometheus but got no response.

I did not disclose details since I don't know how the disclosure policy looks like - I can disclose the vulnerability in this issue if you want to.    

I should not be able to kill a prometheus server the way I can at the moment. 

I can kill (arbitrary) prometheus servers (AFAICT).

* Prometheus version:

`prometheus version 2.2.1` (locally verified)

Older versions are affected too and I think the latest version 2.3.0 as well (but I haven't verified this)

aead:
This commit fixes a denial-of-service issue of the remote
read endpoint. It limits the size of the POST request body
to 32 MB such that clients cannot write arbitrary amounts
of data to the server memory.

Fixes #4238

Signed-off-by: Andreas Auernhammer <aead@mail.de>

### DIFF
--- a/storage/remote/codec.go
+++ b/storage/remote/codec.go
@@ -15,6 +15,7 @@ package remote
 
 import (
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"sort"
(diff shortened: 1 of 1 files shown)


aead:
Travis-CI failure seems unrelated...

brian-brazil:
Thanks!

Issue #4215, opened 2018-06-05
Scrape delay when using remote write to influxdb with huge samples appended


jojohappy:
## Proposal
**Use case. Why is this important?**

## Bug Report
**What did you do?**
I want to export data to the influxdb for long-term storage, so I use the remote_write in prometheus.

**What did you expect to see?**
![image](https://user-images.githubusercontent.com/2532941/40954451-44b4a408-68b7-11e8-87d4-a0a42926756d.png)
![image](https://user-images.githubusercontent.com/2532941/40954468-51b54a22-68b7-11e8-9c42-aee85a6716af.png)

**What did you see instead? Under which circumstances?**
![image](https://user-images.githubusercontent.com/2532941/40954693-46bd1c70-68b8-11e8-89d5-c8b8526af14d.png)
![image](https://user-images.githubusercontent.com/2532941/40954661-2540180e-68b8-11e8-9f38-8b2873d2f789.png)

**Environment**

* System information:

```
/prometheus $ uname -srm
Linux 4.4.0-104-generic x86_64
```

* Prometheus version:

```
/prometheus $ prometheus --version
prometheus, version 2.2.1 (branch: HEAD, revision: bc6058c81272a8d938c05e75607371284236aadc)
  build user:       root@149e5b3f0829
  build date:       20180314-14:15:45
  go version:       go1.10
```

* Alertmanager version:

	insert output of `alertmanager --version` here (if relevant to the issue)

* Prometheus configuration file:
```
global:
  scrape_interval:     15s # Set the scrape interval to every 15 seconds. Default is every 1 minute.
  scrape_timeout: 10s
  evaluation_interval: 15s # Evaluate rules every 15 seconds. The default is every 1 minute.
  external_labels:
      monitor: 'kube-remote'

rule_files:
  - "/etc/prom/rules.d/kubernetes.rules"

scrape_configs:
- job_name: 'prometheus'
  static_configs:
    - targets: ['localhost:9090']

- job_name: 'kubernetes-apiservers'

  kubernetes_sd_configs:
  - role: endpoints
  scheme: https
  tls_config:
    ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
    insecure_skip_verify: true
  bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token

  relabel_configs:
  - source_labels: [__meta_kubernetes_namespace, __meta_kubernetes_service_name, __meta_kubernetes_endpoint_port_name]
[...]

PR #4242, opened 2018-06-08
Improve remote write performance

jojohappy:
This PR aims to decouple the `scrapeLoop` and `fanoutStorage's secondaries appender`. Before this PR, the `secondaries appender` will append samples into the queue of Remote Storage's QueueManager directly after convert data to `model.Sample`. It will be no performance problem when the rate of samples in is low. 
In my case #4215 , we have about 20w / s samples for kubernetes cluster using one Prometheus instance, and we want to take the data flow into our influxdb cluster. We test the Prometheus in version 2.2.1, but only 4w / s samples for `scrapeLoop`, and make `scrapeLoop` delay.
In this PR, I write a `storageAppender` to implement the `scrape.Appendable` to store samples in the queue of `storageAppender` temporarily and append them into `QueueManager` when the `fanoutStorage` commit. This solution come from tsdb `headAppender`.
In another commit, I try to refactor `QueueManager` to remove shards mutex and use single queue that receive samples for every `QueueManager` and reshard workers to consume samples in the queue.
Sorry for my poor English, I'm glad to see feedback for this PR.


tomwilkie:
Thanks @jojohappy! Just about to jump on a plane but will review tomorrow.

storage/remote/queue_manager.go
```diff
@@ -218,9 +218,7 @@ func (t *QueueManager) Append(s *model.Sample) error {
 		return nil
 	}
 
-	t.shardsMtx.Lock()
 	enqueued := t.shards.enqueue(&snew)
-	t.shardsMtx.Unlock()
```
tomwilkie: IIUC, if you remove this lock its possible for this goroutine to read a stale value and do on an enqueue on a set of shards that has been stopped.

If you've found this to be a bottle neck you might want to test using a RWLock instead.  

I may be wrong, but at the very least I'd want to understand why you think removing this lock is safe?

storage/remote/queue_manager.go
```diff
@@ -359,62 +357,62 @@ func (t *QueueManager) reshardLoop() {
...
+	// We start the newShards (worker) after we have stopped (the therefore completely
+	// flushed) the oldShards (worker), to guarantee we only every deliver samples in
 	// order.
 	newShards.start()
 }
 
 type shards struct {
 	qm      *QueueManager
-	queues  []chan *model.Sample
+	queue   chan *model.Sample
```
tomwilkie: By removing the per-shard queue, you also break the guarantee that samples for the same timeseries will be delivered in order.

storage/remote/write.go
```diff
@@ -21,35 +21,45 @@ import (
...
+	s.s.mtx.RLock()
+	defer s.s.mtx.RUnlock()
+	for _, a := range s.samples {
+		for _, q := range s.s.queues {
+			q.Append(&a)
+		}
+	}
+
 	return nil
 }
```
tomwilkie: I actually think you might be on to something here: samples should be batched up and posted to the remote storage in a batch - but I think it should be the batch of samples that is place on the queue and handled as a unit - you're currently unwinding any batching you've done in the commit, and I can't see how this make things any faster.

The problem then would be deciding which queue to place the batch on.  You could potentially just shard on the (job, instance) labels, which would also mean histograms become and atomic etc.

tomwilkie:
Hi @jojohappy; sorry for the delay.  I'm not sure your changes are safe - see my comments.  Let me know when you want to have another look, I think the batching up of samples in the appender is a good idea and could be plumbed through the whole remote pipeline for some nice performance improvements.

storage/remote/queue_manager.go
jojohappy: Yes, I also found this problem after testing. But I have no idea to solve it. Do you have some suggestion?

storage/remote/write.go
jojohappy: Did you mean, the samples should be appended into the queue of shard as a batch, and do something in the shard for each sample?

storage/remote/queue_manager.go
jojohappy: Sorry, I did not realize this problem and how to guarantee the samples. I will try to use a RWLock instead.

jojohappy:
Hi @tomwilkie ! Thank you for your review. In this PR, I try to refactor `QueueManager`, but it seems to be have some big problems. 

I have a question that should `QueueManager` use one queue with multi workers (shards) or multi queues with shards manager? My opinion is that it will be remove the `FastFingerprint`, only one queue will be receive samples. We can scale the worker (shard) to post samples into remote storage.

storage/remote/queue_manager.go
tomwilkie: > I also found this problem after testing

What problem?

storage/remote/write.go
tomwilkie: Indeed, thats an option.  In the worker, the batch can then be translated into a write request proto and sent to the server.

tomwilkie:
The sharding between multiple queues (with one worked per queue) is needed to ensure samples for a given timeseries are delivered in order, which is a guarantee some backends rely on.  Unless you can think of another way to ensure in-order delivery, they need to stay.

If you go down the route of putting batches on the queue (instead of samples), this should reduce the number of times you need to call FastFingerprint.  The question will be what key to use to decide where to shard when you have a batch of samples.

jojohappy:
> ensure samples for a given timeseries are delivered in order, which is a guarantee some backends rely on

@tomwilkie  Thank you for your help! I miss this scenarios.

Maybe it is not a good way to refactor `QueueManager` in this PR, but I will continue to implement that samples can be translated as batch in `QueueManager`.  

So should I close this PR util it should be reopened again?

tomwilkie:
Sounds good!  Tag me on the next PR you open and I'll give that a look.

Issue #4214, opened 2018-06-05
scraping should not be effected when reloading Prometheus without change any config


codwu:
## Bug Report
**What did you do?**

reload Prometheus without change any config
```
curl -X POST http://localhost:9090/-/reload
```

**What did you expect to see?**
there is no effect on target scraping

**What did you see instead? Under which circumstances?**
scrape interval become longer than I configured
![default](https://user-images.githubusercontent.com/12193868/40949508-f62e969c-68a0-11e8-9139-1346579bd3f2.PNG)

**Environment**

* System information:

```
Linux 3.10.0-229.el7.x86_64 x86_64
```

* Prometheus version:

```
prometheus, version 2.2.1 (branch: HEAD, revision: bc6058c81272a8d938c05e75607371284236aadc)
  build user:       root@149e5b3f0829
  build date:       20180314-14:15:45
  go version:       go1.10
```

* Prometheus configuration file:
```

global:
  scrape_interval:     30s 
  evaluation_interval: 30s

alerting:
  alertmanagers:
  - static_configs:
    - targets:
      # - alertmanager:9093

rule_files:

scrape_configs:
  - job_name: 'prometheus1'
    static_configs:
      - targets: ['localhost:9090']
  - job_name: 'prometheus2'
    static_configs:
      - targets: ['localhost:9090']
  - job_name: 'prometheus3'
    static_configs:
      - targets: ['localhost:9090']

```

* Alertmanager configuration file:
```
insert configuration here (if relevant to the issue)
```

* Logs:
```
level=info ts=2018-06-05T01:10:15.681143246Z caller=main.go:220 msg="Starting Prometheus" version="(version=2.2.1, branch=HEAD, revision=bc6058c81272a8d938c05e75607371284236aadc)"
level=info ts=2018-06-05T01:10:15.681218089Z caller=main.go:221 build_context="(go=go1.10, user=root@149e5b3f0829, date=20180314-14:15:45)"
level=info ts=2018-06-05T01:10:15.681243105Z caller=main.go:222 host_details="(Linux 3.10.0-229.el7.x86_64 #1 SMP Fri Mar 6 11:36:42 UTC 2015 x86_64 centos7b (none))"
level=info ts=2018-06-05T01:10:15.681261809Z caller=main.go:223 fd_limits="(soft=1024, hard=4096)"
level=info ts=2018-06-05T01:10:15.683833415Z caller=main.go:504 msg="Starting TSDB ..."
[...]

PR #4245, opened 2018-06-08
config: set target group source index during unmarshalling

pgier:
Fixes issue #4214 where the scrape pool is unnecessarily reloaded for a
config reload where the config hasn't changed.  Previously, the discovery
manager changed the static config after loading which caused the in-memory
config to differ from a freshly reloaded config.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -370,6 +370,13 @@ func (c *ScrapeConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 			}
 		}
 	}
+
+	// Add index to the static config target groups for unique identification
+	// within scrape pool.
+	for i, tg := range c.ServiceDiscoveryConfig.StaticConfigs {
+		tg.Source = fmt.Sprintf("%d", i)
+	}
+
 	return nil
 }
 
@@ -432,6 +439,13 @@ func (c *AlertmanagerConfig) UnmarshalYAML(unmarshal func(interface{}) error) er
 			}
 		}
 	}
+
+	// Add index to the static config target groups for unique identification
+	// within scrape pool.
+	for i, tg := range c.ServiceDiscoveryConfig.StaticConfigs {
+		tg.Source = fmt.Sprintf("%d", i)
+	}
+
 	return nil
 }
 
--- a/discovery/manager_test.go
+++ b/discovery/manager_test.go
@@ -774,6 +774,45 @@ scrape_configs:
 	verifyPresence(discoveryManager.targets, poolKey{setName: "prometheus", provider: "static/0"}, "{__address__=\"bar:9090\"}", false)
 }
 
+func TestApplyConfigDoesNotModifyStaticProviderTargets(t *testing.T) {
+	cfgText := `
+scrape_configs:
+ - job_name: 'prometheus'
+   static_configs:
+   - targets: ["foo:9090"]
+   - targets: ["bar:9090"]
+   - targets: ["baz:9090"]
+`
+	originalConfig := &config.Config{}
+	if err := yaml.UnmarshalStrict([]byte(cfgText), originalConfig); err != nil {
+		t.Fatalf("Unable to load YAML config cfgYaml: %s", err)
+	}
+	origScrpCfg := originalConfig.ScrapeConfigs[0]
+
+	processedConfig := &config.Config{}
+	if err := yaml.UnmarshalStrict([]byte(cfgText), processedConfig); err != nil {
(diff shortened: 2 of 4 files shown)


config/config.go
```diff
@@ -370,6 +370,13 @@ func (c *ScrapeConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 			}
 		}
 	}
+
+	// Add index to the static config target groups for unique identification
+	// within scrape pool
+	for i, tg := range c.ServiceDiscoveryConfig.StaticConfigs {
```
simonpasquier: this could also be done in [targetgroup.UnmarshalYAML()](https://github.com/prometheus/prometheus/blob/25e2d9f152b634d65123ef11342aceb23e66a1a2/discovery/targetgroup/targetgroup.go#L40). This would avoid duplication for handling AlertManager SD.

pgier: It doesn't work within the target group because the id/index to set for tg.Source is based on the array index which is not known in the targetgroup unmarshal function.

I also wasn't able to remove the duplication using the ServiceDiscoveryConfig unmarshall because it's inlined in ScrapeConfig and AlertmanagerConfig, so its unmarshal function never gets called.

I could move the loop to a function to reduce the duplication a bit, wdyt?

config/config.go
```diff
@@ -370,6 +370,13 @@ func (c *ScrapeConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 			}
 		}
 	}
+
+	// Add index to the static config target groups for unique identification
+	// within scrape pool
```
simonpasquier: (nit) a dot is needed at the end.

config/config.go
```diff
@@ -432,6 +439,13 @@ func (c *AlertmanagerConfig) UnmarshalYAML(unmarshal func(interface{}) error) er
...
```
simonpasquier: (nit) a dot is needed at the end.

config/config.go
simonpasquier: Ah right, I misread the code. I'm fine with it then.

discovery/manager_test.go
```diff
@@ -774,6 +774,47 @@ scrape_configs:
...
+		t.Fatalf("Unable to load YAML config cfgYaml: %s", err)
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	discoveryManager := NewManager(ctx, nil)
+	go discoveryManager.Run()
+
+	for i, scrpConfig := range processedConfig.ScrapeConfigs {
+		origScrpCfg := originalConfig.ScrapeConfigs[i]
+		for _, discoverer := range discoveryManager.providersFromConfig(scrpConfig.ServiceDiscoveryConfig) {
```
simonpasquier: I would find it easier to read if the test called `discoveryManager.ApplyConfig()` then waited on `<- discoveryManager.SyncCh()` as it is done in the other tests.

krasi-georgiev:
thanks Paul looks good.

I did a test a  while back to cache this,but apparently it  wasn't very well designed. Do you think it can be modified to catch this bug?
https://github.com/prometheus/prometheus/blob/0e5e7f75cd8607a0406ff38bc10dc4b084aac5f5/scrape/manager_test.go#L229

discovery/manager_test.go
pgier: Yeah, I agree that's a better way to do it.  Branch is now updated, thanks!

discovery/manager_test.go
```diff
@@ -774,6 +774,49 @@ scrape_configs:
...
+			t.Fatalf("problem getting target grouops from channel\n")
+		}
+		for _, tgs := range tgsmap {
+			if !reflect.DeepEqual(origScrpCfg.ServiceDiscoveryConfig.StaticConfigs, tgs) {
+				t.Fatalf("discovery manager modified static config \n  expected: %v\n  got: %v\n",
+					origScrpCfg.ServiceDiscoveryConfig.StaticConfigs, tgs)
+			}
+		}
+	}
+}
```
simonpasquier: The test fails in Travis because there's no guarantee on the order of targets. What about just waiting for the discovery manager to send the updates and check that it didn't touch the configuration?

```go
<-discoveryManager.SyncCh()
if !reflect.DeepEqual(origScrpCfg.ServiceDiscoveryConfig.StaticConfigs, processedConfig...) {
...
}
```

brian-brazil:
Thanks!

Issue #1551, opened 2016-04-12
Create tool to make it easy to share debug information


grobie:
A common request during debugging issues is to request various pprof information from users, for example in #1549.

In order to make that process easier for users and less time intensive for developers, it'd be great to have a `promtool debug share-profile` to automate the process of:
- getting a profile
- creating SVG/whatever on it
- optionally uploading and attaching it to a github issue.

juliusv:
I was always thinking of a web tool that we could offer where people could enter their Prometheus URL and it would pull interesting info out of it automatically via AJAX and store it as a permalink somehow. However, that would only work for API endpoints, because those have CORS enabled. Of course, we could add more info to those as well.

grobie:
That might be even easier. I think all information of the prometheus server should be exposed via the API.

chyeh:
Hello, I'm new here! I would like to know if anyone is working on this issue. Any update?

juliusv:
@chyeh Hi and welcome! I don't think anyone is currently working on this, but it would still be something that's nice to have.

chyeh:
@juliusv tks for your reply. After reading through this thread, I feel this issue needs some further discussion. My question is what exactly is debug information that should be exposed? Before implementation, I might need to confirm the spec of the API with you.

juliusv:
Here's an initial list of details that would be useful for debugging:

- Information that can be gained via Prometheus's query API:
  - All current values of metrics about Prometheus itself (only works if Prom is scraping itself, otherwise we have to scrape its `/metrics`)
  - Derived expressions on those metrics like `rate(prometheus_tsdb_head_samples_appended_total[1m])`. Also requires Prom to scrape itself.
  - Information relating to target scrapes like `scrape_duration`, etc.
- Information that can be retrieved via other API endpoints:
  - The Prometheus configuration (including rules), available via `/api/v1/status/config` and `/api/v1/status/flags`. Rules are not exposed yet via API, but should be.
  - Information about targets. Available via `/api/v1/targets`.
  - Information about connected Alertmanagers via `/api/v1/alertmanagers`.
- Information available via `/debug/pprof` (usually obtained via `go tool pprof`):
  - CPU profile
  - Heap profile
  - goroutine dump
  - etc.

[...]

PR #4247, opened 2018-06-09
Add 3 commands in `promtool` for getting debug information from prometheus server

chyeh:
This is for issue #1551. In this PR, the following commands are added:
1. `promtool debug metrics <prometheus url>`:
    It generates `debug.tar.gz` which contains `metrics.txt`
2. `promtool debug pprof <prometheus url>`:
    It generates `debug.tar.gz` which contains `block.pb`, `goroutine.pb`, `heap.pb`, `mutex.pb.gz` and `threadcreate.pb`
3. `promtool debug all <prometheus url>`:
    It generates `debug.tar.gz` which contains `block.pb`, `goroutine.pb`, `heap.pb`, `mutex.pb.gz`, `threadcreate.pb` and `metrics.txt`


krasi-georgiev:
this  would be super useful for debugging, thanks ,
I see that you are still pushing commits , but let us know when it is ready for review.
Thanks again.

chyeh:
It's ready for review now, thanks!

cmd/promtool/archive.go
```diff
@@ -0,0 +1,81 @@
...
+	"os"
+)
+
+const FILE_PERM = 0644
+
+type archiveWriterConfig struct {
+	archiveName string
+}
+
+type archiveWriter interface {
```
krasi-georgiev: not a deal breaker, but maybe `archiveWriter` could be just `archiver`

krasi-georgiev: or can we use the io.WriteCloser?

cmd/promtool/resource.go
```diff
@@ -0,0 +1,131 @@
...
+	m := make(map[string]*http.Request)
+	for path, filename := range cfg.pathToFileName {
+		req, err := http.NewRequest(http.MethodGet, client.urlJoin(path), nil)
+		if err != nil {
+			panic(err)
+		}
+		m[filename] = req
+	}
+	return &resourceSaver{
+		writer:            tw,
```
krasi-georgiev: embed this to remove the stutter? `c.writer.write` will become `c.write`

cmd/promtool/main.go
```diff
@@ -66,15 +65,23 @@ func main() {
 
 	queryCmd := app.Command("query", "Run query against a Prometheus server.")
 	queryInstantCmd := queryCmd.Command("instant", "Run instant query.")
-	queryServer := queryInstantCmd.Arg("server", "Prometheus server to query.").Required().URL()
+	queryServer := queryInstantCmd.Arg("server", "Prometheus server to query.").Required().String()
```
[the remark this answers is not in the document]
chyeh: The [constructor](https://github.com/prometheus/client_golang/blob/master/api/client.go#L66) takes the [argument](https://github.com/prometheus/client_golang/blob/master/api/client.go#L43) of string type and it returns error if it fails to parse the URL string. So I think it's redundant that it parses the url string by `kingpin.URL()` method in advance.

cmd/promtool/archive.go
```diff
@@ -0,0 +1,81 @@
+// Copyright 2015 The Prometheus Authors
```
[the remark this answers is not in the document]
chyeh: I copied and pasted and I didn't noticed. Thanks for reminding me!

cmd/promtool/archive.go
chyeh: I think `archiver` is better! I'm not sure what do you mean by `io.WriteCloser`?

cmd/promtool/main.go
```diff
@@ -66,15 +65,23 @@ func main() {
...
 	queryRangeBegin := queryRangeCmd.Flag("start", "Query range start time (RFC3339 or Unix timestamp).").String()
 	queryRangeEnd := queryRangeCmd.Flag("end", "Query range end time (RFC3339 or Unix timestamp).").String()
 
+	debugCmd := app.Command("debug", "debug cmd")
+	debugPprofCmd := debugCmd.Command("pprof", "pprof cmd")
+	debugPprofServer := debugPprofCmd.Arg("server", "Prometheus server to get pprof files.").Required().String()
+	debugMetricsCmd := debugCmd.Command("metrics", "metrics cmd")
+	debugMetricsServer := debugMetricsCmd.Arg("server", "Prometheus server to get metrics.").Required().String()
+	debugAllCmd := debugCmd.Command("all", "all cmd")
+	debugAllServer := debugAllCmd.Arg("server", "Prometheus server to get all debug infomation.").Required().String()
```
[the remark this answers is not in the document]
chyeh: I'm not familiar with `kingpin`. I will take a took if this way is possible!

cmd/promtool/resource.go
chyeh: Sure, good suggestion!

cmd/promtool/archive.go
chyeh: I was thinking about this. The `write` method needs the argument of `filename`. If we want to implement the `Write(p []byte) (n int, err error)` method of `io.WriteCloser` instead, we will need to turn the argument of `filename` into the field of the implementation of `io.WriteCloser`, and every time before `Write()` is called, we need to reset the value of the `filename` so it can write to the corresponding file in the archive. It makes us need another method `SetHeader(filename string, size int64)` to this interface (currently `archiveWriter `) so we can't still just use `io.WriteCloser`. In this case, it will become something like:

```
type archiveWriter interface {
	io.WriteCloser
	SetHeader(filename string, size int64)
}
```
Another choice is that we embed only the `io.Closer` interface and let the `write()` method still handle the header setting part:
```
type archiveWriter interface {
	io.Closer
	write(fileName string, buf *bytes.Buffer) error
}
```
So what do you think?

cmd/promtool/main.go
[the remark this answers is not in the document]
chyeh: According to the [discussion](https://github.com/alecthomas/kingpin/issues/36), we can't do this. To make the code cleaner. We can make `server` a flag. For being consistent I believe the `query` command should change too. What do you think?

chyeh:
I've modified accordingly except the `debugCmd` part! Let me know if there is anything else. Thanks!

cmd/promtool/resource.go
```diff
@@ -41,7 +41,7 @@ func newResourceSaver(cfg resourceSaverConfig) *resourceSaver {
 	if err != nil {
 		panic(err)
 	}
-	tw, err := newArchiveWriter(archiveWriterConfig{archiveName: cfg.tarballName})
+	a, err := newArchiver(archiverConfig{archiveName: cfg.tarballName})
```
krasi-georgiev: I personally prefer more descriptive names - `archiver`, this would also make more  sense when you initialise the struct below , but not a deal breaker.

cmd/promtool/resource.go
```diff
@@ -30,7 +30,7 @@ type resourceSaverConfig struct {
 }
 
 type resourceSaver struct {
-	writer            archiveWriter
+	archiver
 	httpClient        httpClient
 	fileNameToRequest map[string]*http.Request
```
krasi-georgiev: wouldn't this be better reversed?
`requestToFile map[*http.Request]string`
we are making a request and writing to a file right?

cmd/promtool/resource.go
```diff
@@ -54,37 +54,39 @@ func newResourceSaver(cfg resourceSaverConfig) *resourceSaver {
 		m[filename] = req
 	}
 	return &resourceSaver{
-		writer:            tw,
-		httpClient:        client,
-		fileNameToRequest: m,
-		postProcess:       cfg.postProcess,
+		a,
+		client,
+		m,
```
krasi-georgiev: reqs or requests  would be more descriptive.

cmd/promtool/resource.go
```diff
@@ -0,0 +1,133 @@
...
+		if err != nil {
+			panic(err)
+		}
+		m[filename] = req
+	}
+	return &resourceSaver{
+		a,
+		client,
+		m,
+		cfg.postProcess,
```
krasi-georgiev: same here I think we need more descriptive var names , reqs/requests , archiver are not so long.

cmd/promtool/resource.go
```diff
@@ -30,7 +30,7 @@ type resourceSaverConfig struct {
 }
 
 type resourceSaver struct {
-	writer            archiveWriter
+	archiver
 	httpClient        httpClient
```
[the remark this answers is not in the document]
chyeh: Yes! I should've thought about it!

cmd/promtool/resource.go
chyeh: This makes more sence!

cmd/promtool/resource.go
chyeh: Agree, since we removed the field name below.

cmd/promtool/resource.go
chyeh: It seems like map variables are named after the value more than the key. It's going to be `map[*http.Request]string`, but I feel using the key (`reqs` instead of `files`) makes more sense in this context!

cmd/promtool/archive.go
```diff
@@ -0,0 +1,86 @@
...
+const FILE_PERM = 0644
+
+type archiverConfig struct {
+	archiveName string
+}
+
+type archiver interface {
+	write(fileName string, buf *bytes.Buffer) error
+	close() error
+	archive() *os.File
```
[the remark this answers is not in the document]
chyeh: If we use `name()` here, in [resource.go#L89](https://github.com/chyeh/prometheus/blob/PROM-1551/cmd/promtool/resource.go#L89) it will become `fmt.Println("debug complete all files written in:", rs.name())` which I believe is less descriptive. What do you think?

cmd/promtool/main.go
```diff
@@ -93,6 +100,45 @@ func main() {
...
+			pathToFileName: map[string]string{
+				"/debug/pprof/block":        "block.pb",
+				"/debug/pprof/goroutine":    "goroutine.pb",
+				"/debug/pprof/heap":         "heap.pb",
+				"/debug/pprof/mutex":        "mutex.pb",
+				"/debug/pprof/threadcreate": "threadcreate.pb",
+				"/metrics":                  "metrics.txt",
+			},
+			postProcess: allPostProcess,
+		}).exec())
```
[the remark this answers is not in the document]
chyeh: It has the embedded `write()` method from `archiver` which conflicts, so I'll make it `Write()`.

cmd/promtool/archive.go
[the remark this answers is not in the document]
chyeh: Since `file` conflicts to the `tarGzFileWriter`'s `file` field, so we need to change the name of the field. Is `archive` ok? Something like:
```go
type tarGzFileWriter struct {
	tarWriter *tar.Writer
	gzWriter  *gzip.Writer
	archive   *os.File
}
```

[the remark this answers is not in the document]
krasi-georgiev: aah I didn't know about this, I added a PR to your branch with my suggestion also cleared few more nits from the golang linter.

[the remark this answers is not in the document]
chyeh: Merged! Thanks!!

krasi-georgiev:
@chyeh Can't find anymore places where we can improve this so just waiting if any of the other maintainers has any other suggestions.
Thanks a lot this is a great PR.
now need to start start suggesting it to get debug info in issues and see how it will work in practice. 
Once we make sure it works well I will also suggest to add a binary  to the release page so people can download and run directly.

LGTM

@brian-brazil , @simonpasquier  - this one is ready for one final look.

chyeh:
Thanks! I actually learned a lot from your suggestions!

cmd/promtool/resource.go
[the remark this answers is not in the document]
juliusv: Generally I would use struct field embedding with care. Explicit is better than implicit, reducing magic, etc.

cmd/promtool/archive.go
```diff
@@ -0,0 +1,83 @@
...
+const filePerm = 0644
+
+type archiverConfig struct {
+	archiveName string
+}
+
+type archiver interface {
+	Write(fileName string, buf *bytes.Buffer) error
+	Close() error
+	File() *os.File
```
juliusv: These aren't used outside of this package (and can't be, since the surrounding type is unexported), so unexport (lower-case) them.

cmd/promtool/archive.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func newArchiver(cfg archiverConfig) (archiver, error) {
+	file, err := os.Create(cfg.archiveName)
+	if err != nil {
+		return nil, fmt.Errorf("error of creating archive %s: %s", cfg.archiveName, err)
```
juliusv: "error of creating" -> "error creating"

Prefer `%q` over `%s` for file names.

cmd/promtool/archive.go
```diff
@@ -0,0 +1,83 @@
...
+	"bytes"
+	"compress/gzip"
+	"fmt"
+	"os"
+)
+
+const filePerm = 0644
+
+type archiverConfig struct {
+	archiveName string
```
juliusv: If it's just one arg, omit the config struct and pass the name as a constructor param directly.

cmd/promtool/debug.go
```diff
@@ -0,0 +1,133 @@
...
+	"bytes"
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/google/pprof/profile"
+)
+
+type debugWriterConfig struct {
+	server         string
```
juliusv: serverURL would be more clear

cmd/promtool/debug.go
```diff
@@ -0,0 +1,133 @@
...
+	archiver
+	httpClient
+	requestToFile map[*http.Request]string
+	postProcess   func(b []byte) (*bytes.Buffer, error)
+}
+
+func newDebugWriter(cfg debugWriterConfig) *debugWriter {
+	client, err := newHTTPClient(httpClientConfig{serverURL: cfg.server})
+	if err != nil {
+		panic(err)
```
juliusv: Panics should only be used to catch programming errors, otherwise we want to output a clean error message instead and return with the right exit code (`1`). In this case, `newHTTPClient` can fail when the user-provided server URL cannot be parsed.

cmd/promtool/debug.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	archiver, err := newArchiver(archiverConfig{archiveName: cfg.tarballName})
+	if err != nil {
+		panic(err)
```
juliusv: Same here and below.

cmd/promtool/debug.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	return &debugWriter{
+		archiver,
+		client,
+		reqs,
+		cfg.postProcess,
+	}
+}
+
+func (w *debugWriter) Write() int {
```
juliusv: "Write()" -> "write()", as we are only using things in this package.

cmd/promtool/http.go
```diff
@@ -0,0 +1,62 @@
...
+	do(req *http.Request) (*http.Response, []byte, error)
+	urlJoin(path string) string
+}
+
+func newHTTPClient(cfg httpClientConfig) (httpClient, error) {
+	hc, err := api.NewClient(api.Config{
+		Address: cfg.serverURL,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("error of creating http client: %s", err)
```
juliusv: "error of creating" -> "error creating"

"http" -> "HTTP"

cmd/promtool/archive.go
chyeh: @krasi-georgiev if we make the `File()` unexported, what new name for the `file` field in `tarGzFileWriter` should we use?

cmd/promtool/debug.go
```diff
@@ -0,0 +1,133 @@
...
+	for req, filename := range w.requestToFile {
+		_, body, err := w.do(req)
+		if err != nil {
+			fmt.Fprintln(os.Stderr, err)
```
juliusv: All our existing `fmt.Println(...)`s have some context string prepended to them, like:

    fmt.Fprintln(os.Stderr, "error executing HTTP request:", err)

Would be good to have here (and below) too.

cmd/promtool/main.go
```diff
@@ -66,15 +65,23 @@ func main() {
...
 	queryExpr := queryInstantCmd.Arg("expr", "PromQL query expression.").Required().String()
 
 	queryRangeCmd := queryCmd.Command("range", "Run range query.")
-	queryRangeServer := queryRangeCmd.Arg("server", "Prometheus server to query.").Required().URL()
+	queryRangeServer := queryRangeCmd.Arg("server", "Prometheus server to query.").Required().String()
 	queryRangeExpr := queryRangeCmd.Arg("expr", "PromQL query expression.").Required().String()
 	queryRangeBegin := queryRangeCmd.Flag("start", "Query range start time (RFC3339 or Unix timestamp).").String()
 	queryRangeEnd := queryRangeCmd.Flag("end", "Query range end time (RFC3339 or Unix timestamp).").String()
 
+	debugCmd := app.Command("debug", "debug cmd")
```
juliusv: "debug cmd" -> "Fetch debug information."

cmd/promtool/main.go
```diff
@@ -66,15 +65,23 @@ func main() {
...
+	debugPprofCmd := debugCmd.Command("pprof", "pprof cmd")
```
juliusv: "pprof cmd" -> "Fetch profiling debug information."

cmd/promtool/main.go
```diff
@@ -66,15 +65,23 @@ func main() {
...
+	debugPprofServer := debugPprofCmd.Arg("server", "Prometheus server to get pprof files.").Required().String()
```
juliusv: "files" -> "files from"

cmd/promtool/main.go
```diff
@@ -66,15 +65,23 @@ func main() {
...
+	debugMetricsCmd := debugCmd.Command("metrics", "metrics cmd")
+	debugMetricsServer := debugMetricsCmd.Arg("server", "Prometheus server to get metrics.").Required().String()
```
juliusv: "get metrics" -> "get metrics from"

cmd/promtool/main.go
```diff
@@ -66,15 +65,23 @@ func main() {
...
+	debugAllCmd := debugCmd.Command("all", "all cmd")
```
juliusv: "all cmd" -> "Fetch all debug information."

cmd/promtool/main.go
juliusv: "infomation" -> "information"

"to get all debug information" -> "to get all debug information from"

cmd/promtool/main.go
```diff
@@ -351,9 +397,9 @@ func CheckMetrics() int {
 }
 
 // QueryInstant performs an instant query against a Prometheus server.
-func QueryInstant(url *url.URL, query string) int {
+func QueryInstant(url string, query string) int {
 	config := api.Config{
-		Address: url.String(),
+		Address: url,
```
juliusv: Since you are introducing that new API HTTP client with built-in timeout in http.go, should it be used here and in the other function below as well?

cmd/promtool/main.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2015 The Prometheus Authors
+// Copyright 2018 The Prometheus Authors
```
juliusv: We usually just leave years in existing copyright notices as they are (the full situation is a bit more complicated if you wanted to do it completely right, see: https://opensource.stackexchange.com/questions/2171/refreshing-shared-copyright-dates-in-apache-2-0-licensed-code)

cmd/promtool/archive.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+type archiver interface {
```
juliusv: I'm not sure we need an interface here, since we are not expecting to replace the implementation. Just the struct would do.

But in case you do want to keep the interface, then the constructor method for the concrete type should a) be called after the concrete type, i.e. `tarGzFileWriter` vs. `newArchiver`, b) return the concrete type, not the interface.

cmd/promtool/main.go
krasi-georgiev: I think @brian-brazil asked me to update these on one of my PRs.

cmd/promtool/debug.go
chyeh: I see! Thanks for reminding!

cmd/promtool/archive.go
juliusv: Generally we tried out all of the following in Prometheus:

- long list of constructor arguments
- option structs (like here)
- functional arguments (https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis)

In the end, just simple and plain constructor arguments turned out best for most cases because they make it crystal clear what needs to be passed and explicitly let the compiler know if you need a new / different argument. If you add a new option to config struct, it's often not well-defined whether it's a required or optional option, and someone using the API will not even notice during compile time that they are missing an argument.

Still, we sometimes use option structs when the argument list gets too long, but even with long argument lists we often just use normal arguments now.

cmd/promtool/debug.go
```diff
@@ -0,0 +1,133 @@
...
+func validate(b []byte) (*profile.Profile, error) {
+	p, err := profile.Parse(bytes.NewReader(b))
+	if err != nil {
+		return nil, err
+	}
+	return p, nil
+}
+
+func buffer(p *profile.Profile) (*bytes.Buffer, error) {
+	buf := new(bytes.Buffer)
```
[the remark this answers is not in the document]
chyeh: Since the return type is `*bytes.Buffer` so I decided to use `new` for getting a pointer. These two approaches both seem to be recommended by the [official document](https://golang.org/pkg/bytes/#Buffer). Actually I was bothered by the case when it occurs error. What should I return as the empty value of `bytes.Buffer`? And I feel it's simpler to just use a pointer over than a value so `nil` can be the empty value. What do you think?

cmd/promtool/main.go
chyeh: Actually I was thinking so, but I feel I should open another issue and resolve with another PR. Just easier for tracking the issues this way. What do you think?

cmd/promtool/archive.go
chyeh: I was thinking that there might be some situations in the future that they require different ways on writing the information to the files so I want to keep it a `interface` and just change the `newArchiver ` to `newTarGzFileWriter` and it returns `tarGzFileWriter`. Thanks for the suggestion!

cmd/promtool/archive.go
chyeh: I see! Thanks for explanation!

cmd/promtool/debug.go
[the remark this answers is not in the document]
juliusv: If you expand the first example for `bytes.Buffer` (https://golang.org/pkg/bytes/#Buffer), it also just does `var b bytes.Buffer` and then takes the address of it when it's needed (`&b`). But yeah, it doesn't matter too much.

You don't seem to be actually needing any of the `bytes.Buffer` functionality (like writing more into it) after this function returns, so I would just return `buf.Bytes()` directly as a `[]byte` slice.

cmd/promtool/archive.go
juliusv: Fine with me, no biggie. I just wanted to point out that sometimes it's better to not prematurely introduce interfaces before one really needs them.

cmd/promtool/main.go
juliusv: Yeah, since this PR is getting a bit busy, maybe let's do it in a separate PR.

cmd/promtool/archive.go
krasi-georgiev: @juliusv the problem here was the we can't have a method `file`  and a field name `file` in  `tarGzFileWriter`

juliusv: That's not a great reason for changing the exposure of a method though... actually I don't think you even need the `File()` method on here, because outside of the type you don't even need the `*os.File`, you just need the filename, so it could be `filename()`? (don't expose more than you have to)

cmd/promtool/archive.go
krasi-georgiev: removing the interface might actually remove the naming collisions

cmd/promtool/debug.go
chyeh: Should [here](https://github.com/chyeh/prometheus/blob/PROM-1551/cmd/promtool/main.go#L356) be the same?

juliusv: A panic is fine there because the only way to get there is a programming error (not a user error, not a transient network or IO error or such).

chyeh: I see!! Thanks for the explantation!!

cmd/promtool/archive.go
chyeh: In this case, I'll change [httpClientConfig](https://github.com/chyeh/prometheus/blob/PROM-1551/cmd/promtool/http.go#L28) too.

cmd/promtool/archive.go
chyeh: Hey I know why I did this. I saw [the constructor for `httpClient` in client_golang](https://github.com/prometheus/client_golang/blob/master/api/client.go) and take it as a reference. In this case should it be changed too?

chyeh: Also I did the same thing [here](https://github.com/chyeh/prometheus/blob/PROM-1551/cmd/promtool/http.go#L36) so I'll change it too.

chyeh:
Thanks @juliusv for suggestions! I've push new commits. There are some followups:
1. Should I change the license back to year 2015 ?
2. Should I submit a PR at [client_golang](https://github.com/prometheus/client_golang) to make the change on the [constructor of `httpClient`](https://github.com/prometheus/client_golang/blob/master/api/client.go)?
3. I will open an issue and submit a relative PR to make `QueryInstant()` and `QueryRange()` switch to the local HTTP client.

cmd/promtool/main.go
```diff
@@ -66,15 +65,23 @@ func main() {
 
 	queryCmd := app.Command("query", "Run query against a Prometheus server.")
 	queryInstantCmd := queryCmd.Command("instant", "Run instant query.")
-	queryServer := queryInstantCmd.Arg("server", "Prometheus server to query.").Required().URL()
+	queryServer := queryInstantCmd.Arg("server", "Prometheus server to query.").Required().String()
 	queryExpr := queryInstantCmd.Arg("expr", "PromQL query expression.").Required().String()
 
 	queryRangeCmd := queryCmd.Command("range", "Run range query.")
-	queryRangeServer := queryRangeCmd.Arg("server", "Prometheus server to query.").Required().URL()
+	queryRangeServer := queryRangeCmd.Arg("server", "Prometheus server to query.").Required().String()
```
juliusv: Maybe this has been discussed in one of the many outdated comment threads, but what was the reason here for using `String()` instead of `URL()`?

juliusv:
@chyeh Thanks!

> 1. Should I change the license back to year 2015 ?

Yep, I would leave it at the date it had before for the existing license header. TBH, it doesn't matter much though.

> 2. Should I submit a PR at [client_golang](https://github.com/prometheus/client_golang) to make the change on the [constructor of `httpClient`](https://github.com/prometheus/client_golang/blob/master/api/client.go)?

That `Client` is a type that is exported and a user of the package has the ability to create a custom client implementation that they then pass into `NewAPI()` at https://github.com/prometheus/client_golang/blob/master/api/prometheus/v1/api.go#L205. The interface is actually used to swap out the HTTP client for a test client in the API tests: https://github.com/prometheus/client_golang/blob/master/api/prometheus/v1/api_test.go

So I would keep that as it is.

> 3. I will open an issue and submit a relative PR to make `QueryInstant()` and `QueryRange()` switch to the local HTTP client.

Cool, thanks!

Looks good now, besides my one new comment.

cmd/promtool/main.go
chyeh: Since I didn't make any change there so you can still see above. 
>The [constructor](https://github.com/prometheus/client_golang/blob/master/api/client.go#L66) takes the [argument](https://github.com/prometheus/client_golang/blob/master/api/client.go#L43) of string type and it returns error if it fails to parse the URL string. So I think it's redundant that it parses the url string by `kingpin.URL()` method in advance.

juliusv: Ah ok, hm yeah. Although IMO it's still nice to be as specific as possible in flag definitions, even if it might implicitly lead to checking the URL twice somewhere later in the program.

chyeh:
> The interface is actually used to swap out the HTTP client for a test client in the API tests: 

I'm still a little bit confused. Since the [`NewClient` constructor](https://github.com/prometheus/client_golang/blob/master/api/client.go#L66) returns the interface type variable instead of the type of the implementation, which is [`httpClient`](https://github.com/prometheus/client_golang/blob/master/api/client.go#L79), is it ok to use the [`NewClient` constructor](https://github.com/prometheus/client_golang/blob/master/api/client.go#L66) to create a implementation of the [`Client`](https://github.com/prometheus/client_golang/blob/master/api/client.go#L58) interface?

juliusv:
@chyeh Yeah, that's fine since `httpClient` implements `Client` (so it *is* a `Client`), although it might be nicer if the constructor returned the concrete type. Then if that concrete type has more features or fields than the interface it implements, they would still be (more easily) usable by the caller.

chyeh:
I just figured out the issues in this repository are mostly for feature requests or bugs so I will just submit another PR that refers to this one. Also I might do some refactoring by the way.

Let me know if there is any change that should be done in this PR. Thanks!

juliusv:
Cool, I think this looks good for now (I might still want the `.String()` -> `.URL()` back in the future ;)).

Now we just have to wait until the feature moratorium is over before merging this: https://groups.google.com/forum/#!topic/prometheus-developers/8PCVxpYhXI8

cmd/promtool/archive.go
krasi-georgiev: yes this is what I mean, but  @chyeh can try different ways and see what works/reads/looks best.

krasi-georgiev:
Thanks @chyeh

Issue #4004, opened 2018-03-23
--web.route-prefix doesn't work as expected


beorn7:
**What did you do?**

Ran prometheus with `--web.route-prefix=/prefix`

**What did you expect to see?**

All paths prefixed with `/prefix`

**What did you see instead? Under which circumstances?**

- `http://localhost:9090/prefix` redirects to `http://localhost:9090/graph`, which obviously doesn't work.
- Typing `http://localhost:9090/prefix/graph` explicitly still has little success because all the `/static/` resources are not prefixed.

* Prometheus version:

prometheus, version 2.2.0 (branch: master, revision: 60dafd425cdc96f7df3019cf756998b42209cf1d)
  build user:       bjoern@reasonableresolution
  build date:       20180323-15:46:34
  go version:       go1.10

brian-brazil:
#4025 was not the right fix for this. Can you indicate what the external URL was in this case?

abh:
It seems like the original problem here would have been fixed by using external-url instead of route-prefix.

juliusv:
Has this been fixed in https://github.com/prometheus/prometheus/pull/4256 now?

roidelapluie:
I think this can be closed @juliusv @beorn7

beorn7:
I'd assume so. But I got repeatedly confused by the behavior in the past, so it would be good if somebody else could confirm.

roidelapluie:
I confirm that it works as expected and that in most cases you want external.url.

This behavious was even revisited recently for the new UI: https://github.com/prometheus/prometheus/issues/6163

beorn7:
OK, closing then…

PR #4256, opened 2018-06-12
web: restore old path prefix behavior

fabxc:
Signed-off-by: Fabian Reinartz <freinartz@google.com>

@brian-brazil 

@beorn7 I believe your issue in #4004 was not a bug in that sense. `--web.route-prefix` sets a prefix for the internal HTTP router, while `--web.external-url` allows defining a prefix path that is exposed to the outside world. 
The intention being that different kinds of reverse prefix may or may not strip path prefixes away. Thus we need support to configure both things independently.

### DIFF
--- a/web/web.go
+++ b/web/web.go
@@ -770,13 +770,7 @@ func tmplFuncs(consolesPath string, opts *Options) template_text.FuncMap {
 			return time.Since(t) / time.Millisecond * time.Millisecond
 		},
 		"consolesPath": func() string { return consolesPath },
-		"pathPrefix": func() string {
-			if opts.RoutePrefix == "/" {
-				return ""
-			} else {
-				return opts.RoutePrefix
-			}
-		},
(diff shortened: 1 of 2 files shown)


brancz:
I think the confusion is that if route prefix is unset and the external url path is set, then route prefix defaults to that path of the external url - I think that’s kind of confusing. I think the next time we do a major release we should change this behavior and not make one flag (sometimes) depend on the value of the other.

brian-brazil:
:+1:

The behaviour of these flags is a result of long long discussions over years, I don't think we should change them lightly.

fabxc:
Yes, I think the intent was to do the least surprising thing for the user. But what is surprising probably also depends on what kind of reverse proxy they are dealing with, so that was possibly not the right choice.

beorn7:
Perhaps the meaning of the flags can be explained better in their docstring then? People seem very confused about them in general.

Issue #4246, opened 2018-06-09
2.3.0 returns wrong content-type for built-in javascript libraries


veksh:
## Bug Report

I've installed 2.3.0 in a VM with port 9090 forwarded to localhost, and
web interface does not work as expected: no javascript libraries are loaded (2.2.1 is OK).
On closer inspection, error console has 14 errors like

```
Refused to execute script from 'http://localhost:9090/static/vendor/js/jquery.min.js?v=290d71791a507a5057b9a099c9d48703d86dc941' because its MIME type ('text/x-js') is not executable, and strict MIME type checking is enabled.
```

so most probably `Content-Type` header for built-in javascript libraries is wrong.

Command-line is

```
/usr/local/bin/prometheus --config.file=/usr/local/etc/prometheus.yml --storage.tsdb.path=/var/lib/prometheus/data/ --web.console.templates=/usr/local/share/prometheus/console --web.console.libraries=/usr/local/share/prometheus/console_libraries --log.level=debug
```

**Environment**

* System information:

	Linux 3.0.101-108.38-default x86_64

* Prometheus version:

	prometheus, version 2.3.0 (branch: HEAD, revision: 290d71791a507a5057b9a099c9d48703d86dc941)
           build user:       root@d539e167976a
           build date:       20180607-08:46:54
           go version:       go1.10.2

* Prometheus configuration file:
```
global:
  scrape_interval:     15s
  evaluation_interval: 15s
  scrape_timeout:      10s
scrape_configs:
  - job_name: 'prometheus'
    static_configs:
    - targets: ['localhost:9090']
```

krasi-georgiev:
that is probably because of this commit
https://github.com/prometheus/prometheus/commit/c207920500d1d55c21d7b7dbb27ba213394ce369

fabxc:
@krasi-georgiev any further info how that commit causes this?

brian-brazil:
It's likely the nosniff.

fabxc:
Okay, what was the motivation for adding those security headers? Were we exposed to any security risk?

brian-brazil:
General security hardening was the motivation.

fabxc:
Yes, but what are the risks those headers are fixing? I personally haven't seen them before and don't know what they do and how they are causing this issue. I'm all for hardening, but would be good to understand what we are hardening against and how.

hasso:
[...]

PR #3583, opened 2017-12-13
Added HTTP security headers to web.go

eavalenzuela:
Added standard HTTP security headers for both the root/redirect requests and all template-based UI pages.

This configuration for the headers is very unrestrictive, while still meeting most compliance frameworks and industry best practices. They should not interfere with any functionality in Prometheus (and did not in my testing).

PR #4258, opened 2018-06-12
web: remove security headers

fabxc:
This reverts changes from #3583 due to issues it causes in #4246

From the original PR:

> This configuration for the headers is very unrestrictive, while still meeting most compliance frameworks and industry best practices. They should not interfere with any functionality in Prometheus (and did not in my testing).

This does not seem to hold true. Maybe we just need to set yet another set of headers when serving our JS files? I don't know and the original PR didn't go into any detail what those headers actually do and how they help.  
Given they were not addressing an imminent problem and are breaking things in the current state, reverting them seems to be the best solution for now.

@eavalenzuela if you can shed some lights on this or have a solution to fix this quickly, that'd be great.

### DIFF
--- a/web/web.go
+++ b/web/web.go
@@ -71,17 +71,6 @@ import (
 
 var localhostRepresentations = []string{"127.0.0.1", "localhost"}
 
-// secureHeadersMiddleware adds common HTTP security headers to responses.
-func secureHeadersMiddleware(h http.Handler) http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.Header().Add("X-XSS-Protection", "1; mode=block")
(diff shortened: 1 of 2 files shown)


krasi-georgiev:
yeah I think it was more like "nice to have" PR. 
I will try to be more strict on the use cases and request more details on how and why we need the changes.

LGTM on green

brian-brazil:
:+1:

roidelapluie:
I do think that security is out of prometheus scope and that those headers should be managed at the reverse proxy level.

krasi-georgiev:
@roidelapluie yep I don't think these are considered out of scope either, but just need to be more careful so we don't brake existing users.
Just will be more careful requiring more  details what  is the security issue and how it solves it before merging the PR.

Issue #4240, opened 2018-06-08
Read endpoint for grafana5 throws invalid memory address or nil pointer dereference


ximply:
Prometheus 2.3 (and 2.2.1 is ok) use as a read endpoint for grafana5 will throw a error like this:

level=error ts=2018-06-08T02:39:50.80339667Z caller=stdlib.go:89 component=web caller="http: panic serving 192.168.10.176:31525" msg="runtime error: invalid memory address or nil pointer dereference"

192.168.10.176 is the grafana server

brian-brazil:
Can you clarify what you mean by "read endpoint"? Which URL is Grafana hitting?

ximply:
Sorroy, **read enpoint** is **remote read endpoint**.

The setting of grafana datasource for prometheus:
the **_HTTP URL_** is **http://192.168.10.248:8080**
the **_HTTP Access_** is **proxy**

ximply:
Here are the settings:
![default](https://user-images.githubusercontent.com/3693445/41146891-72938696-6b37-11e8-86b1-eb8e1d562206.png)

brian-brazil:
Grafana doesn't use the remote read endpoint. Can you capture the HTTP request in question?

ximply:
But I use Prometheus 2.2.1, it is ok.

y2sarakawa:
Grafana 5 accesses the Prometheus /series endpoint when referring to variables.

If more than one entry is in remote_read, nil is passed to "p * storage.SelectParams" in ToQuery () of storage/remote/codec.go.

After merging #4122, nil pointer dereference will occur to refer to "p" when creating ReadHints in ToQuery ().
https://github.com/prometheus/prometheus/blob/v2.3.0/storage/remote/codec.go#L96

Please also see #4205.

brian-brazil:
Looks like a dupe of #4205 alright.

PR #4260, opened 2018-06-12
fix panic caused by series endpoint with remote_read

y2sarakawa:
This is for issue #4240.

PR #4213 was complicated, so this is its rewrite based on release-2.3.

### DIFF
--- a/web/api/v1/api_test.go
+++ b/web/api/v1/api_test.go
@@ -29,9 +29,12 @@ import (
 	"testing"
 	"time"
 
+	"github.com/go-kit/kit/log"
 	"github.com/gogo/protobuf/proto"
 	"github.com/golang/snappy"
+	config_util "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
+	"github.com/prometheus/common/promlog"
 	"github.com/prometheus/common/route"
 
 	"github.com/prometheus/prometheus/config"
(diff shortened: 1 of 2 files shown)


web/api/v1/api_test.go
```diff
@@ -646,6 +649,90 @@ func TestReadEndpoint(t *testing.T) {
...
+	}
+	seriesURL.RawQuery = q.Encode()
+	req, err := http.NewRequest("GET", seriesURL.String(), nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, apiErr, _ := api.series(req)
+	if apiErr != nil {
+		if apiErr.typ == errorExec {
```
henridf: I realize you had a request on another PR to test the series/ endpoint, but having to whitelist various connection errors this way seems fragile.... I'd say that just testing the changed function (`ToQuery`) in `codec_test.go` would be ok.

y2sarakawa: Thanks. I will cancel this PR.

y2sarakawa:
I will recreate another PR that excluded the unit test later.

henridf:
hi @y2sarakawa i'm happy to open a PR with your change and push it through if you'd like. just ping me here if so.

y2sarakawa:
#4263 was opened. I have nothing to do.

PR #4263, opened 2018-06-13
remote read: nil pointer deference when using remove read

iksaif:
```
level=error ts=2018-06-13T07:19:04.515149169Z caller=stdlib.go:89 component=web caller="http: panic serving [::1" msg="]:56202: runtime error: invalid memory address or nil pointer dereference"
level=error ts=2018-06-13T07:19:04.516199547Z caller=stdlib.go:89 component=web caller="http: panic serving [::1" msg="]:56204: runtime error: invalid memory address or nil pointer dereference"
level=error ts=2018-06-13T07:19:04.51717692Z caller=stdlib.go:89 component=web caller="http: panic serving [::1" msg="]:56206: runtime error: invalid memory address or nil pointer dereference"
level=error ts=2018-06-13T07:19:04.564952878Z caller=stdlib.go:89 component=web caller="http: panic serving [::1" msg="]:56208: runtime error: invalid memory address or nil pointer dereference"
level=error ts=2018-06-13T07:19:04.566575791Z caller=stdlib.go:89 component=web caller="http: panic serving [::1" msg="]:56210: runtime error: invalid memory address or nil pointer dereference"
level=error ts=2018-06-13T07:19:04.567106063Z caller=stdlib.go:89 component=web caller="http: panic serving [::1" msg="]:56212: runtime error: invalid memory address or nil pointer dereference"
```

When remove read is enabled, federation will call `q.Select(nil, mset...)`
which will break remote reads because it currently doesn't handle empty
SelectParams.

This does not contain unit test, but since it's a critical 2.3.0 bug it is probably still worth merging.

### DIFF
--- a/storage/remote/codec.go
+++ b/storage/remote/codec.go
@@ -95,8 +95,8 @@ func ToQuery(from, to int64, matchers []*labels.Matcher, p *storage.SelectParams
 	if err != nil {
 		return nil, err
 	}
-	var rp *prompb.ReadHints = nil
 
+	var rp *prompb.ReadHints
 	if p != nil {
 		rp = &prompb.ReadHints{
 			StepMs: p.Step,
--- a/storage/remote/codec.go
+++ b/storage/remote/codec.go
@@ -95,10 +95,13 @@ func ToQuery(from, to int64, matchers []*labels.Matcher, p *storage.SelectParams
 	if err != nil {
 		return nil, err
 	}
+	var rp *prompb.ReadHints = nil
 
-	rp := &prompb.ReadHints{
-		StepMs: p.Step,
-		Func:   p.Func,
+	if p != nil {
+		rp = &prompb.ReadHints{
+			StepMs: p.Step,
+			Func:   p.Func,
+		}
 	}
 
 	return &prompb.Query{
--- a/web/api/v1/api_test.go
+++ b/web/api/v1/api_test.go
@@ -31,7 +31,9 @@ import (
 
 	"github.com/gogo/protobuf/proto"
 	"github.com/golang/snappy"
+	config_util "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
+	"github.com/prometheus/common/promlog"
 	"github.com/prometheus/common/route"
 
 	"github.com/prometheus/prometheus/config"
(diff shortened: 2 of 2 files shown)


iksaif:
cc: @henridf, @tomwilkie who recently contributed to this code

henridf:
looks good to me, thanks @y2sarakawa and @iksaif  for the fix.

tomwilkie:
@iksaif I've extended all the API tests to also cover remote read; there is some duplication of the remote read API code that could potentially be reduced, but other wise this seems to tickle the bug (and potentially other) quite nicely.  WDYT?

@fabxc @y2sarakawa @brian-brazil could do with an extra pair of eyes on this, if you have time.  Thanks!

web/api/v1/api_test.go
```diff
@@ -128,30 +130,125 @@ func TestEndpoints(t *testing.T) {
...
+			}
+		}
+
+		if err := remote.EncodeReadResponse(&resp, w); err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+	})
+
+	return httptest.NewServer(handler)
```
tomwilkie: With a little refactoring this could be deduped against https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L592, happy to do so if anyone thinks it worth it?

storage/remote/codec.go
```diff
@@ -95,10 +95,13 @@ func ToQuery(from, to int64, matchers []*labels.Matcher, p *storage.SelectParams
 	if err != nil {
 		return nil, err
 	}
+	var rp *prompb.ReadHints = nil
```
brian-brazil: Isn't it nil already by default?

web/api/v1/api_test.go
brian-brazil: Mightn't hurt to have it as an "unofficial" helper for those writing remote read endpoints.

brian-brazil:
:+1: with one nit

storage/remote/codec.go
tomwilkie: Yup

web/api/v1/api_test.go
tomwilkie: only downside is our code deals with adding/removing external labels, so I'll see if we can factor that out to a wrapper.

tomwilkie: Meh no its pretty horrific factoring it out.

tomwilkie:
Just going to rebase this against release-2.3 and open a new PR.

tomwilkie:
Done, see #4282

PR #4265, opened 2018-06-13
Pass group-by names in ReadHints

henridf:
This PR adds a list of group-by fields to SelectParams and ReadHints. For example for the query `sum by (url, app)(my_gauge)` that list would be `[url, app]`.

This is needed in order for our (sysdig) remote read backend to allow grouping by fields that are derived metadata and that would not otherwise be returned. There may be other uses for it in other contexts.


brian-brazil:
> to allow grouping by fields that are derived metadata and that would not otherwise be returned.

This sounds a bit off semantically. Hints should never change the end results that a user sees, they should only enable optimisations.

henridf:
> This sounds a bit off semantically. Hints should never change the end results that a user sees, they should only enable optimisations.

There's already a precedent where SelectParams _do_ change the results (in Thanos). 

What's the difference between ReadHints and SelectParams in that respect? (I should probably have asked that question when the name came up 
 in https://github.com/prometheus/prometheus/pull/4122#pullrequestreview-116263721 ...)

brian-brazil:
> There's already a precedent where SelectParams do change the results (in Thanos).

Not as far as I'm aware, the end user result doesn't change.

> What's the difference between ReadHints and SelectParams in that respect?

SelectParams is the actual selector being sent down which you are responding to. ReadHints gives you some additional context to allow you to process it more efficiently. For example if someone is asking for data with a 1m step, there's no need to send second granularity data back. You can instead prune the samples that would have been discarded.

brian-brazil:
> Depending on the value of SelectParams.Func, you get different results back e.g here https://github.com/improbable-eng/thanos/blob/master/pkg/query/querier.go#L144

Prometheus gets back different results for remote read, the user does not. These are optimisations that don't affect the query result (beyond the fundamental changes due to downsampling).

> ReadHints is nothing more than the "forwarding on" of SelectParams to be sent over with a remote read query.

Ah wait, I got confused. You are right. In both cases the result the end-user gets should not change based on ReadHints/SelectParams though.

henridf:
> Prometheus gets back different results for remote read, the user does not. These are optimisations that don't affect the query result (beyond the fundamental changes due to downsampling).

With downsampling, the user absolutely will get different results back based on the value of `ReadHints.Func`. In this case it's not just an optimization hint: its presence and value is going to affect the results of the promql query. 

But regardless, if the goal is that additional SelectParams/ReadHints should not carry result-modifying information, then PR can probably be closed.

promql/engine.go
```diff
@@ -519,6 +521,18 @@ func (ng *Engine) populateSeries(ctx context.Context, q storage.Queryable, s *Ev
 	return querier, err
 }
 
+func extractGroupsFromPath(p []Node) []string {
```
brian-brazil: I don't think this is safe as-is, for example:

```
max by  (foo)(a - b)
```

Would fail to do proper label matching between a and b if a remote storage was using this. I think we need to limit this to aggregators either directly around the selector, or around a function which is around the selector.

henridf:
Closing as per discussion above.

DeeeFOX:
@henridf @brian-brazil 

In my use case, i do this by：

 sum(abc{__HINTS__="{'by':'url|app'}"}) by (url, app)

Issue #4257, opened 2018-06-12
Make web interface not indexable by search engines


pachico:
## Proposal
**Make Prometheus' web interface not indexable by search engines.**

It might be a silly request but I trust it would be a good addition to append a `<meta name=”robots” content=”noindex,nofollow”>` in the head section.

I know it's up users to ensure proper security policies but a quick search in Google helps to understand this is not always happening.

Thanks for your great work.

PR #4266, opened 2018-06-13
Forbid rule-abiding robots from indexing.

MartinLeedotOrg:
Resolves github issue #4257

@juliusv

### DIFF
--- a/web/ui/templates/_base.html
+++ b/web/ui/templates/_base.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
-    <meta name=”robots” content=”noindex,nofollow”>
+    <meta name="robots" content="noindex,nofollow">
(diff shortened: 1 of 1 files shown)


web/ui/templates/_base.html
```diff
@@ -2,6 +2,7 @@
 <html lang="en">
   <head>
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+    <meta name=”robots” content=”noindex,nofollow”>
```
juliusv: Thanks! Could you use the regular `"` quotes here instead of `”`?

MartinLeedotOrg: Amended. Not sure how you want to operate in this repo, should I rebase?

brian-brazil:
We can squash ourselves on merge. This is not a bugfix, so given the current moratorium on non-bugfix changes this will not be merged for 2 weeks.

MartinLeedotOrg:
Understood!

brian-brazil:
Thanks!

PR #4267, opened 2018-06-13
fix the TestManagerReloadNoChange test

krasi-georgiev:
after moving the config check in ApplyConfig , this test wasn't actually doing anything.

### DIFF
--- a/scrape/manager_test.go
+++ b/scrape/manager_test.go
@@ -16,13 +16,13 @@ package scrape
 import (
 	"fmt"
 	"testing"
-	"time"
 
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/config"
-	"github.com/prometheus/prometheus/discovery/targetgroup"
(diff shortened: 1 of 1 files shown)


brian-brazil:
Shouldn't this test be failing given #4245 isn't in?

krasi-georgiev:
nope , the bug is different and can't be catched in this test.

here is what happened in #4245 
 - Load the configs in ApplyConfig for Scrape
 - Load the configs in ApplyConfig for Discovery
 - The Discovery changes a value in the Scrape config and since this is a pointer it also affect the Config saved in Scrape.
 - On next  Scrape ApplyConfig  the saved config doesn't match the new config due to the change that happened in the Discovery ApplyConfig.

I have suggested a new test as part of the PR to that bug which  will catch such bugs but that is a separate issue. 
https://github.com/pgier/prometheus/pull/1

IMO this is another proof that the config handling and this global state needs some rethinking.
I remember few races caused by this as well.

PR #4268, opened 2018-06-13
test to ensure no change of the global config after unmarshling.

krasi-georgiev:
Came up with this idea for a more global check.
This is a better more global test to catch bugs like #4214 

In general the config handling needs some serious refactoring as this global state and pointers have caused few other issues and races and I think this will bite us again.

Signed-off-by: Krasi Georgiev kgeorgie@redhat.com

### DIFF
--- a/cmd/prometheus/main_test.go
+++ b/cmd/prometheus/main_test.go
@@ -14,6 +14,7 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"net/http"
@@ -24,7 +25,16 @@ import (
 	"testing"
 	"time"
 
+	"github.com/go-kit/kit/log"
+	"github.com/prometheus/prometheus/config"
+	"github.com/prometheus/prometheus/discovery"
+	sd_config "github.com/prometheus/prometheus/discovery/config"
+	"github.com/prometheus/prometheus/notifier"
+	"github.com/prometheus/prometheus/scrape"
+	"github.com/prometheus/prometheus/storage/remote"
 	"github.com/prometheus/prometheus/util/testutil"
+	"github.com/prometheus/prometheus/web"
+	k8s_runtime "k8s.io/apimachinery/pkg/util/runtime"
 )
 
 var promPath string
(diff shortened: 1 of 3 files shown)


cmd/prometheus/main_test.go
```diff
@@ -155,3 +165,54 @@ func TestComputeExternalURL(t *testing.T) {
 		}
 	}
 }
+
+// TestApplyConfig ensures that no component is modifing the global config.
+// The config includes many pointers so modifing it has global side effects.
```
brian-brazil: modifying

cmd/prometheus/main_test.go
```diff
@@ -155,3 +165,54 @@ func TestComputeExternalURL(t *testing.T) {
...
+		scrapeManager.ApplyConfig,
+		func(cfg *config.Config) error {
+			c := make(map[string]sd_config.ServiceDiscoveryConfig)
+			for _, v := range cfg.ScrapeConfigs {
+				c[v.JobName] = v.ServiceDiscoveryConfig
+			}
+			return discoveryManager.ApplyConfig(c)
+		},
+	}
+	testutil.Ok(t, reloadConfig("../../config/testdata/conf.good.yml", logger, reloaders...))
```
brian-brazil: There is not a test case in this file that covers the test you are removing.

cmd/prometheus/main_test.go
krasi-georgiev: thanks.

cmd/prometheus/main_test.go
krasi-georgiev: not sure what you mean here. 
If you revert the changes for this bug https://github.com/prometheus/prometheus/issues/4214 you will see that the tests will fail.

the bug is triggered by the following execution

config reload is triggered.
scrape.ApplyConfig - loads and saves the config.
discovery.ApplyConfig - changes the config and since the some fields are pointers the saved config in scrape.ApplyConfig is also changed.

another config reload is triggered without any config file changes.
scrape.ApplyConfig - checks the provided config to check if it matches with the saved one and since the config was changed by discovery.ApplyConfig the check fails and triggers scrape loop reload.

this test ensures that no ApplyConfig modifies the original file which will prevent such bugs.

krasi-georgiev:
@brian-brazil can't make sense about this race that popped up. any ideas ?

cmd/prometheus/main_test.go
brian-brazil: good.yml doesn't contain a static_configs with multiple entires.

krasi-georgiev: that is not important in this case.

krasi-georgiev: actually I just double checked and it does have a static config with 2 targets in it.

https://github.com/prometheus/prometheus/blob/91ce63a1402ff2e2447eb0286ba0e0c17b64624a/config/testdata/conf.good.yml#L50-L54

still trying to figure out the failing race

brian-brazil: That's only one static_config though

krasi-georgiev: Adding more to this file brakes `TestLoadConfig` and I really don't see why you think it is necessary. the bug wasn't related at all to having more than one static config entry.
if you really think this is necessary lets do it in a separate PR as it will require a non related changes to the `TestLoadConfig` if we do it here.

krasi-georgiev:
ping @pgier , @simonpasquier , @brian-brazil

cmd/prometheus/main_test.go
brian-brazil: We still need a test added with two static_configs to ensure this doesn't regress.

krasi-georgiev: I understand your point, but as I mentioned this PR solves a different problem so your concern needs to be addressed in a separate one.

brian-brazil: You are still removing test cases without replacing them.

krasi-georgiev: the test I am removing is completely covered in by the change I am suggesting.

The test I am removing is so specific that it wouldn't cover regressions anyway.

krasi-georgiev: the original bug was not about this, but it seems that we won't come to an agreement this way so I updated the test to use a different config file. Hopefully this should satisfy your concern.

brian-brazil: Thanks, there's no need to create a new file though. The existing good configuration is meant to be an exhaustive example of every possible configuration option, so you can add it in there.

krasi-georgiev: I already mentioned that I did this, but it brakes some other tests and I don't feel like fixing unrelated tests. this should be done in a separate PR.

PR #4271, opened 2018-06-14
Use first label pointer as Labels ID instead of the hash

free:
Use first label pointer as Labels ID instead of computing the hash. This provides a boost of ~10% to vector binary ops with one series on each side -- compare the `a_one_and_b_one,steps=1000` benchmark -- and ~25% to vector binary ops with 100 series on each side -- compare the `a_hundred_and_b_hundred,steps=1000` benchmark: [before](https://github.com/prometheus/prometheus/files/2101515/before.txt) and [after](https://github.com/prometheus/prometheus/files/2101517/after.txt). The gain to functions that deal with labels is slightly lower, because of the string/regexp operations overhead.

Within the context of an `EvalNodeHelper`, the `labels.Labels` instance associated with an eval vector never changes, meaning that a pointer to the instance could be used as a unique identifier instead of computing a hash value (which is relatively costly). Slices however, seem to actually be references to the underlying array, so the address of a slice (which is what `labels.Labels` is) actually seems to change as it gets passed around. The workaround is to use the address of the first label in the slice, or `nil` iff the slice is empty.

Note that this performance optimization doesn't affect the correctness of the eval code: it is only used for caching, so even if somehow the `labels.Labels` instance identifying a vector changes during evaluation, what happens is that no cached value will be found and the signature and/or label drop will have to be recomputed. The only case in which correctness might be affected is if two `labels.Labels` values get swapped (or one is deallocated and the other gets allocated in the same memory location).

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -601,9 +601,9 @@ type EvalNodeHelper struct {
 
 	// Caches.
 	// dropMetricName and label_*.
-	dmn map[uint64]labels.Labels
+	dmn map[*labels.Label]labels.Labels
 	// signatureFunc.
-	sigf map[uint64]uint64
+	sigf map[*labels.Label]uint64
 	// funcHistogramQuantile.
 	signatureToMetricWithBuckets map[uint64]*metricWithBuckets
 	// label_replace.
@@ -615,35 +615,43 @@ type EvalNodeHelper struct {
 	resultMetric map[uint64]labels.Labels
 }
 
+// firstLabel returns the address of the first label, as a unique identifier of a Labels instance.
+func firstLabel(l labels.Labels) *labels.Label {
+	if len(l) == 0 {
+		return nil
+	}
+	return &l[0]
+}
+
 // dropMetricName is a cached version of dropMetricName.
 func (enh *EvalNodeHelper) dropMetricName(l labels.Labels) labels.Labels {
 	if enh.dmn == nil {
-		enh.dmn = make(map[uint64]labels.Labels, len(enh.out))
+		enh.dmn = make(map[*labels.Label]labels.Labels, len(enh.out))
 	}
-	h := l.Hash()
-	ret, ok := enh.dmn[h]
+	k := firstLabel(l)
+	ret, ok := enh.dmn[k]
 	if ok {
 		return ret
 	}
 	ret = dropMetricName(l)
-	enh.dmn[h] = ret
+	enh.dmn[k] = ret
 	return ret
 }
 
 // signatureFunc is a cached version of signatureFunc.
 func (enh *EvalNodeHelper) signatureFunc(on bool, names ...string) func(labels.Labels) uint64 {
 	if enh.sigf == nil {
-		enh.sigf = make(map[uint64]uint64, len(enh.out))
+		enh.sigf = make(map[*labels.Label]uint64, len(enh.out))
 	}
 	f := signatureFunc(on, names...)
 	return func(l labels.Labels) uint64 {
-		h := l.Hash()
-		ret, ok := enh.sigf[h]
+		k := firstLabel(l)
+		ret, ok := enh.sigf[k]
 		if ok {
 			return ret
 		}
 		ret = f(l)
-		enh.sigf[h] = ret
+		enh.sigf[k] = ret
 		return ret
 	}
 }
--- a/promql/functions.go
+++ b/promql/functions.go
@@ -730,14 +730,14 @@ func funcLabelReplace(vals []Value, args Expressions, enh *EvalNodeHelper) Vecto
 		if !model.LabelNameRE.MatchString(dst) {
 			panic(fmt.Errorf("invalid destination label name in label_replace(): %s", dst))
 		}
-		enh.dmn = make(map[uint64]labels.Labels, len(enh.out))
+		enh.dmn = make(map[*labels.Label]labels.Labels, len(enh.out))
 	}
 
 	outSet := make(map[uint64]struct{}, len(vector))
 	for _, el := range vector {
-		h := el.Metric.Hash()
+		k := firstLabel(el.Metric)
 		var outMetric labels.Labels
-		if l, ok := enh.dmn[h]; ok {
+		if l, ok := enh.dmn[k]; ok {
 			outMetric = l
 		} else {
 			srcVal := el.Metric.Get(src)
@@ -745,7 +745,7 @@ func funcLabelReplace(vals []Value, args Expressions, enh *EvalNodeHelper) Vecto
 			if indexes == nil {
 				// If there is no match, no replacement should take place.
 				outMetric = el.Metric
-				enh.dmn[h] = outMetric
+				enh.dmn[k] = outMetric
 			} else {
 				res := enh.regex.ExpandString([]byte{}, repl, srcVal, indexes)
 
@@ -754,7 +754,7 @@ func funcLabelReplace(vals []Value, args Expressions, enh *EvalNodeHelper) Vecto
 					lb.Set(dst, string(res))
 				}
 				outMetric = lb.Labels()
-				enh.dmn[h] = outMetric
+				enh.dmn[k] = outMetric
 			}
 		}
 
@@ -792,7 +792,7 @@ func funcLabelJoin(vals []Value, args Expressions, enh *EvalNodeHelper) Vector {
 	)
 
 	if enh.dmn == nil {
-		enh.dmn = make(map[uint64]labels.Labels, len(enh.out))
+		enh.dmn = make(map[*labels.Label]labels.Labels, len(enh.out))
 	}
 
 	for i := 3; i < len(args); i++ {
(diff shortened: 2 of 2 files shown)


brian-brazil:
This is a cute optimisation. One of the things I'm planning on doing is reusing evalNodeHelpers across different runs of a query (e.g. recording rules), and I don't think that this PR will interact well with that.

One thing I have considered is either pre-calculating or caching the hash on a labels.Labels.

free:
> One thing I have considered is either pre-calculating or caching the hash on a labels.Labels.

I have actually extended this code with a hash cache similar to `EvalNodeHelper.sigf` (before having read your reply) for use by `VectorBinop` and `resultMetric`. It gives another 10%, but only in simple one to one `a - b` type expressions, over 1000 steps. Everything else is too heavy for it to make a significant difference, so it gets lost in the noise.

Should I merge it into this PR, seeing how it uses the same `firstLabel()` function to map labels to their hash?

I am positively curious to see the `EvalNodeHelper` reused across recorded rules. On my end, I keep thinking about the possibility of label interning (similar to Java String interning) and re-implementing `labels.Labels` as slices of pointers to interned labels, with memoized hash values and possibly interned themselves. But I'm held back by the scale of the project and the possibility it may turn out to be a dud. Plus life, in general.

brian-brazil:
I was thinking something in the Labels class itself rather than in promql.

> On my end, I keep thinking about the possibility of label interning

The problem there is garbage collection.

free:
The problem with doing it in the `Labels` class is that there's an unsafe conversion between tsdb and prom labels that expects `Labels` to be a slice of structs with 2 fields. So you'd need to make the change in TSDB too. Or, now that I think of it, one could take the TSDB slice and put it into a struct (the new prom `Labels`) that also has other fields. Hmm... Lemme see...

> The problem there is garbage collection.

I don't think it's all that necessary: I mean how many different labels is a Prometheus instance going to ever see? And if it turns out to be needed I'd go with something simple: periodically throw away the pool and start from scratch. It can't be worse than what happens now, which is to allocate new labels every time. It's only a cache, after all.

brian-brazil:
I haven't tried implementing it yet :)

>  I mean how many different labels is a Prometheus instance going to ever see?

Churn is a thing, and we're already doing some pretty funky stuff in terms of memory management. I'm not actually sure this optimisation is safe either, you'd have to check every single use of the labels class to be sure.

free:
I've tried to give it a go (making `labels.Labels` an interface, with the default implementation a pointer to a struct wrapping the labels and the hash) but apart from having to edit all range expressions and direct references to a label, I'm running into a couple of problems:
 1. I'm now allocating 2 objects for each `Labels` instance (the slice and the struct).\
 2. There are `var foo labels.Labels` declarations all over the place, which currently result in an empty label set, but with my change will result in a nil pointer and cause segfaults all over.

All in all, it doesn't look like a very promising approach. Did you have a different idea in mind? Something more idiomatic, maybe?

brian-brazil:
In relation to this PR, it depends on an invariant which it is unclear if it is true currently and would be very difficult to maintain in the future, and would also prevent other planned work with much bigger gains, so I don't plan on merging it.

For the other ideas discussed in this PR, we can discuss them in a future PR if they work out.

free:
> In relation to this PR, it depends on an invariant which it is unclear if it is true currently and would be very difficult to maintain in the future, and would also prevent other planned work with much bigger gains, so I don't plan on merging it.

As pointed out in the original post, this does not affect the correctness of the eval logic. Those maps are simply caches, so if you end up with 2 or N instances of the same `Labels`, it's simply not going to be a very efficient cache and `dropMetricName` or the signature function is going to be called more often than necessary. That being said, as things currently stand, the `Labels` instances are actually uniquely identified by their memory location, so this is not even an issue.

When your "reuse `EvalNodeHelper` across evals" change drops, the only change required in order for memory to not blow up (and, again, not for correctness) would be for the caches to be cleared either where the `EvalNodeHelper` instances are currently allocated or, if one wants to minimize memory utilization, by a `defer` statement issued from those locations. Which is probably a wise thing to do anyway, or those caches will grow indefinitely in the face of churn.

free:
I gave the `labels.Labels` as `struct` with a `[]labels.Label` field and a memoized `hash` field a try (see https://github.com/free/prometheus/tree/labels_hash, and particularly [this commit](https://github.com/free/prometheus/commit/c6e0882988dac82660b997f3683fe06ee81fc8ab)).

It's not pretty, it introduces a bunch of edge cases (particularly with regards to testing) and all it provides is a 12-17% improvement to `label_replace` and `label_join` over long ranges while slowing down everything else by 1-2% and using marginally more memory (<1%). Overall I don't think it's worth it.

This PR OTOH, is much more simple and limited in scope, is *correct* (as explained above) and will remain correct even with the changes you envision, and produces performance improvements across the board with zero increase in memory utilization or allocations. Could you please reconsider your decision to not have it merged?

brian-brazil:
That doesn't sound worth it alright, particularly as those functions should not be very common.

There's no proof this PR is correct, as it's susceptible to false positives as memory gets reallocated and reused.

free:
Oh BTW, here are the benchmark results for reference: [before](https://github.com/prometheus/prometheus/files/2228680/before.Jul18.txt) and [after](https://github.com/prometheus/prometheus/files/2228682/after.labels_hash.txt).

The only situation in which this PR might produce incorrect results is if the garbage collector moved the array underlying one `Labels` instance to the exact same address where another `Labels` instance used to be. As long as pointers to these slices exist (in the caches) they will not be garbage collected, so the memory address cannot be reused. And with regards to the garbage collector moving stuff around, I would expect it to update any pointers to data that was moved (or else all hell would break loose), so even in that case you wouldn't end up pointing at a different `Labels` instance than the one you were pointing to originally.

Put differently, if this PR was not correct, you may have the following situation happening in your run-of-the-mill Go program: 
1. Create a pointer to a value within a slice.
2. Lose the reference to the slice, without ever having modified it.
3. (garbage collection magic kicks in)
4. Your pointer now points to a different value at the same memory location.

juliusv:
Agreed to the argument why GC shouldn't be a problem, but it does rely on the invariant that those `labels.Labels` never get mutated and that there are never two `labels.Labels` objects with the same labels. So far that has only been this way by accident, now everyone would have to watch out that this arcane detail holds true in the future. So from a maintenance perspective I'm also not really sure about this.

free:
It does rely on the invariant that `labels.Labels` never get mutated, indeed. This is thus far (AFAICT) respected, as labels are required to be ordered and, as such, can't be trivially mutated.

It does not rely on three not being two 'labels.Labels' instances with the same labels. These are simply caches for "what is this `labels.Labels` with specific labels removed" and "what is the hash of this `labels.Labels` with specific labels removed". There is no harm in having multiple entries that map to equal `labels.Labels` or hash values.

Going back to immutability, this is further guaranteed by the fact that these are label sets returned by queriers or generated by the query engine, and the query engine would be the only one who might mutate them. Thus far it doesn't and this could be verified to be the case in the future. And if someone does introduce a change that breaks it, that's what unit tests are there to guard against.

brian-brazil:
The question isn't really immutability, it's reusing a labels.Label that's part of one labels.Labels in another labels.Labels. labels.Labels are used as lists in various places, and for example a more efficient version of dropMetricName could use this. The TSDB could also do this to save memory, and it's a sufficiently subtle breakage that unittests may not catch it.

I don't believe the risk and complexity is worth the relatively small potential performance gains.

free:
A `labels.Label` that's part of a `labels.Labels` is a `struct`. Reusing it in a different `labels.Labels` means making a copy of it (and *not* taking a reference). I don't see how that could be a problem.

And going back to whether `labels.Labels` instances are immutable or not: a memoized hash would have the exact same problem as using a pointer to the instance, if said instance is mutated. As such, compared with a memoized hash (which, as far as I could tell, doesn't help performance at all) the only 2 downsides of using a reference to the `labels.Labels` instance (which is what this `*labels.Label` pointer is, a pointer to the beginning of the slice's underlying storage) are:
1. The possibility of ending up with a larger cache than strictly necessary (and the corresponding increase in the number of "costly" function calls). (Although I'm having a hard time picturing an instance of a query or eval correctly producing multiple time series with equal labels.)
2. (Speculative) When `EvalNodeHelper` reuse across evals is implemented, it is likely that some of the expected benefits of reuse (namely those derived from reusing the 2 caches) will be lost to some extent. I.e. the caches won't be of use if the querier returns a different but equal `labels.Labels` instance, but this will be offset to some extent by not having to compute the hash (which turns out to be surprisingly costly, as the benchmarks on this PR show). At that point, we can revisit this change and measure again whether caching by instance is still faster than caching by hash.

free:
Ping?

To summarize the discussion so far:
* There exists no way for this change to produce incorrect results: the garbage collector will not release the `labels.Labels` instance while a pointer points anywhere within it, and so no other `labels.Labels` instance with the same address may be allocated.
* Labels are immutable within the query engine (and if they weren't, any optimization using a memoized hash would break).
* Multiple `labels.Labels` instances with the same value do not affect the correctness, only the performance of the cache.
* `labels.Labels` is a slice of `labels.Label` objects, not pointers. As a result, "reusing" a `labels.Label` within another `labels.Labels` results in a copy (with a different memory address).

Regarding your last comment, there is no complexity to speak of: the change is less than 30 lines of code, a third of which is the function that takes the address of the first label. Also, verifying whether this change is still beneficial or not after your planned change to reuse node helpers across rule evals is trivial and I'm happy to do it.

And finally, I would argue the performance benefits are non-negligible: I can't think of another way of wringing out 20-25% of the eval time of a large vector binary op (maybe apart from fully rewriting the way TSDB samples are exposed to the query engine, which is definitely not a trivial undertaking). The 20-25% reductions in query times occur for range queries with hundreds to thousands of steps (i.e. the kind of `query_range` requests that tend to bring down otherwise well-behaved Prometheus instances) and reducing the time window during which these may overlap, should improve the situation.

PR #4272, opened 2018-06-14
Log the line when failing a PromQL test

free:
Makes it easier to track down failures in large test files.


brian-brazil:
Thanks!

PR #4259, opened 2018-06-12
web: remove security headers

fabxc:
Sorry. Of course I forgot the right merge base again...

PR #4274, opened 2018-06-15
Merge 2.3 fixes to master

simonpasquier:
cc @brian-brazil @krasi-georgiev 

The fix removing the security headers already exists in master as #4258 (#4259 on the release-2.3 branch) but reading the [Wiki](https://github.com/prometheus/prometheus/wiki/HOWTO-cut-a-new-release#branch-management-and-versioning-strategy), I understand that there's nothing special to do about it.

### DIFF
--- a/discovery/kubernetes/endpoints_test.go
+++ b/discovery/kubernetes/endpoints_test.go
@@ -19,6 +19,7 @@ import (
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/discovery/targetgroup"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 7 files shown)


brian-brazil:
Thanks!

All this branch messing is a bit confusing.

grobie:
Wait, did you squash these commits @brian-brazil?? This rewrote the history, instead of merging the branch into master. This completely defeats the purpose of having branches.

grobie:
But rebasing rewrites the history. This must not be done when merging release branches back into master, as that will duplicate every commit.

brian-brazil:
If we want to avoid divergent branches, maybe we should work only on master.

grobie:
We have the release branch concept established for years. Following it would remove the need for any "no more features can be merged" moratoriums and provide a clean and readable diff and history between releases.

PR #4275, opened 2018-06-15
Avoid infinite loop on duplicate NaN values.

### LINKED ISSUE #4254: 2.3.0 significatnt memory usage increase.
## Bug Report
**What did you do?**
Upgraded to 2.3.0

General improvements.

Memory usage, possibly driven by queries, has considerably increased. Upgrade at 09:27, the memory usage drops on the graph after then are from container restarts due to OOM.

container_memory_usage_bytes

image

**Environment**

Prometheus in kubernetes 1.9

* System information:
Standard docker containers, on docker kubelet on linux.

* Prometheus version:
2.3.0
	insert output of `prometheus --version` here

brian-brazil:
Fixes #4254

NaNs don't equal themselves, so a duplicate NaN would
always hit the break statement and never get popped.

We should not be returning multiple data point for the same
timestamp, so don't compare values at all.

### DIFF
--- a/storage/fanout.go
+++ b/storage/fanout.go
@@ -450,10 +450,10 @@ func (c *mergeIterator) Next() bool {
 		return false
 	}
 
-	currt, currv := c.At()
+	currt, _ := c.At()
 	for len(c.h) > 0 {
-		nextt, nextv := c.h[0].At()
-		if nextt != currt || nextv != currv {
+		nextt, _ := c.h[0].At()
+		if nextt != currt {
 			break
 		}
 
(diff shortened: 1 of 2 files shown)


gouthamve:
LGTM

tomwilkie:
LGTM - nice find.

juliusv:
:+1:

(basically, "screw the possibility of two equal timestamps with different values?" otherwise you could test for NaN-ness of course)

brian-brazil:
> screw the possibility of two equal timestamps with different values?

That'd break things anyway, which exact way it breaks isn't too relevant :)

> otherwise you could test for NaN-ness of course

A bitwise comparison would have been the way to do it.

PR #4276, opened 2018-06-15
Set display: inline-block for .label class

### LINKED ISSUE #3972: Improve Prometheus UI Target page by hiding empty fields / or by reverting back to dynamic alignment.
**What did you do?**
I visited Target page on Prometheus v2.2.X UI

Every target with readable label sets.

Badly wrapped label sets that obfuscates the view ):

image

Almost all my targets have more than 4 labels which ends up with random wrapping like this.

I already started discussion about it here: https://github.com/prometheus/prometheus/issues/3373
Not sure if that change actually introduced weird wrappings, but making fixed align here is not really helping. There is less space for things that are usually wider.

I guess, there are several options to fix this:
1) Fix wrapping logic - unfortunately that will make each entry quite long in height (because of fixed align)
2) Revert back to dynamic align (esthetics vs clearness) - that will make more space for things that are wide.

I am not necessarily a frontend guy, so can anybody more experienced (any design wizards here?) help us with this? (:  Any suggestions? @lovisas @caarlos0

**Environment**
* System information:
 Chrome, linux

* Prometheus version:
	v2.2.1

lae:
Overrides bootstrap style, fixes #3972

This goes for option 1 mentioned in that issue but to be honest, it doesn't make that much of a difference in height.

### DIFF
--- a/web/ui/static/css/prometheus.css
+++ b/web/ui/static/css/prometheus.css
@@ -23,4 +23,5 @@ body {
 
 .label {
   white-space: normal;
+  display: inline-block;
 }


lae:
Not sure why the build failed after force pushing a commit message edit - https://travis-ci.org/prometheus/prometheus/builds/392795022 passed fine.

brian-brazil:
Do you have before/after screenshot to compare?

simonpasquier:
Fixed by https://github.com/prometheus/prometheus/pull/5226.

PR #4278, opened 2018-06-18
fix minor issues in custom SD example

cstyan:
@brian-brazil noticed some issues with the SD implementation example while writing up the blog post.


gouthamve:
LGTM, but DCO is missing.

cstyan:
thanks @gouthamve, I really should just alias `commit -s` :)

brian-brazil:
Thanks!

PR #4279, opened 2018-06-18
Saner defaults for remote-write

gouthamve:
Signed-off-by: Goutham Veeramachaneni <cs14btech11014@iith.ac.in>

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -516,9 +516,10 @@ func (s *shards) sendSamples(samples model.Samples) {
 // sendSamples to the remote storage with backoff for recoverable errors.
 func (s *shards) sendSamplesWithBackoff(samples model.Samples) {
 	backoff := s.qm.cfg.MinBackoff
+	req := ToWriteRequest(samples)
+
 	for retries := s.qm.cfg.MaxRetries; retries > 0; retries-- {
 		begin := time.Now()
-		req := ToWriteRequest(samples)
 		err := s.qm.client.Store(s.ctx, req)
 
 		sentBatchDuration.WithLabelValues(s.qm.queueName).Observe(time.Since(begin).Seconds())
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -97,12 +97,12 @@ var (
 		},
 		[]string{queue},
 	)
-	queueCapacity = prometheus.NewGaugeVec(
+	shardCapacity = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Namespace: namespace,
 			Subsystem: subsystem,
-			Name:      "queue_capacity",
-			Help:      "The capacity of the queue of samples to be sent to the remote storage.",
+			Name:      "shard_capacity",
+			Help:      "The capacity of each shard of the queue used for parallel sending to the remote storage.",
 		},
 		[]string{queue},
 	)
@@ -123,7 +123,7 @@ func init() {
 	prometheus.MustRegister(droppedSamplesTotal)
 	prometheus.MustRegister(sentBatchDuration)
 	prometheus.MustRegister(queueLength)
-	prometheus.MustRegister(queueCapacity)
+	prometheus.MustRegister(shardCapacity)
 	prometheus.MustRegister(numShards)
 }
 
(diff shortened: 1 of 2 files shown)


tomwilkie:
LGTM

brian-brazil:
This is not a bugfix so we'll wait on merging this one.

gouthamve:
@brian-brazil Sorry for highjacking this PR, did not realise and committed to the same branch. While not a bug-fix, this would hugely help those facing issues with prometheus OOMing when remote has problems.

We are retrying each batch 10 times which is causing the queue to backup and prometheus to fall-over.

gouthamve:
Queue length doesn't increase during queue-resharding and so does memory. Allocs fell due to retry-backoff. This shows that each sample in the queue takes `~700B` of memory, and the current defaults mean a prometheus ingesting 50K samples quickly climbs to 10G Resident memory usage when upstream is down.

storage/remote/queue_manager.go
```diff
@@ -97,12 +97,12 @@ var (
...
 	)
-	queueCapacity = prometheus.NewGaugeVec(
+	shardCapacity = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Namespace: namespace,
 			Subsystem: subsystem,
-			Name:      "queue_capacity",
-			Help:      "The capacity of the queue of samples to be sent to the remote storage.",
+			Name:      "shard_capacity",
+			Help:      "The capacity of each shard of the queue used for parallel sending to the remote storage.",
```
tomwilkie: We can always just ditch this change from this PR, its not important to the overall changes.

storage/remote/queue_manager.go
```diff
@@ -516,9 +516,10 @@ func (s *shards) sendSamples(samples model.Samples) {
 // sendSamples to the remote storage with backoff for recoverable errors.
 func (s *shards) sendSamplesWithBackoff(samples model.Samples) {
 	backoff := s.qm.cfg.MinBackoff
+	req := ToWriteRequest(samples)
+
 	for retries := s.qm.cfg.MaxRetries; retries > 0; retries-- {
 		begin := time.Now()
-		req := ToWriteRequest(samples)
```
tomwilkie: Did you get anywhere with pulling the snappy & proto out of the loop too?

gouthamve: Couple of issues meant that it needs a little more thought, while we could push retries into the client, this would mean the client now logs the errors, which seemed sub-optimal.

And the allocs from proto and snappy did not show up in pprof, so I deferred this change to another PR to keep this PR simple enough to make tomorrows release.

tomwilkie: Cool, on change at a time.

tomwilkie:
How do those graphs look post these changes?

gouthamve:
Hit 1.5G but stabilised at 800M

https://i.imgur.com/uevDBae.jpg

Some deductions as to why queue-length is small:

With reduced retries, the shards stabilised at 197 here. I guess we drop 100 samples to the floor every 30 + 60 + 90 = 180ms, i.e 555.55 samples dropped per second to the floor per shard. Which means 100K samples dropped across all shards, which is much higher than `55451` samples per second being ingested.

With no samples in the queue, memory also came down.

storage/remote/queue_manager.go
gouthamve: Turns out, they do show up in pprof after these changes :) I'll try to get that done before Brian wakes up tomorrow.

gouthamve: I made the change but there was not enough change in allocs but added a lot of plumbing which I don't think manadates a change here.

gouthamve:
This is ready for another review..

tomwilkie:
LGTM

This fixes something a lot of people are hitting with remote write (Prometheus OOMing if the endpoint goes down), so I think this should probably be merged into 2.3 branch from the next minor release.   @brian-brazil still think we should hold this?

tomwilkie:
> This is a breaking change that could lose data.

You need to provide more exposition here Brian.

brian-brazil:
I had mentioned it on IRC. This drops outage resilience from 100s to 10s, if someone was depending on that then they can now lose data where they wouldn't before.

tomwilkie:
Thanks for explaining, not all of us have the time to follow every conversation on IRC.

tomwilkie:
@gouthamve lets merge into master; even though we've got ideas on reducing memory usage without sacrificing the 100s buffering, they'll be too invasive for a patch release.

PR #4283, opened 2018-06-18
Update vendoring of Prometheus Go client

si74:
This is to pickup changes from
https://github.com/prometheus/client_golang/pull/414. It leads to
better error output in promtool.


brian-brazil:
We're only merging bugfixes at the moment, so this will be on hold for a bit.

You should update all of the client, not just the api bits.

si74:
@brian-brazil Ahoy! I did vendor everything in the go-client now. Thanks!

brian-brazil:
Thanks!

PR #4285, opened 2018-06-19
Merge release-2.3 back to master

brian-brazil:


RichiH:
One of the tests failed, but it's only merging back the release and in line with our moratorium so LGTM.

PR #4287, opened 2018-06-19
Fix some (valid) lint errors

juliusv:
To be merged after the non-bugfix moratorium.


krasi-georgiev:
:+1:

PR #4290, opened 2018-06-19
Handle timeouts during PromQL series population

### LINKED ISSUE #4289: Promql won't time-out queries in expandSeriesSet
## Bug Report
**What did you do?**
Someone sent a large query such as `{__name__=~".+"}[5d]` to our prometheus host.

I expect the query to timeout before completing, while consuming some memory and CPU along the way.

The query runs until it exhausts all available memory on the host and OOM killer get is. I have set it up in a repro env locally and the query has been running for ~60m (its still going) even though I have the timeout set to the default 2m.

**Environment**

* Prometheus version:

2.2

Same symptoms as #4288

juliusv:
Fixes #4289

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -493,7 +493,7 @@ func (ng *Engine) populateSeries(ctx context.Context, q storage.Queryable, s *Ev
 				level.Error(ng.logger).Log("msg", "error selecting series set", "err", err)
 				return err
 			}
-			n.series, err = expandSeriesSet(set)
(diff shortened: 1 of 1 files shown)


juliusv:
@brian-brazil @jacksontj

juliusv:
Of course this still doesn't help if a single series takes too long, but better than the current state.

juliusv:
Retracting in favor of duplicate implementation in https://github.com/prometheus/prometheus/pull/4291

PR #4294, opened 2018-06-20
Simplify BufferedSeriesIterator usage

free:
Allow for `BufferedSeriesIterator` instances to be created without an underlying iterator, to simplify their usage.

Signed-off-by: Alin Sinpalean <alin.sinpalean@gmail.com>


brian-brazil:
That looks cleaner (and will save a bit of memory for federation). We can merge this after the bugfix moratorium, and you've fixed the failing tests.

free:
Sorry for the tests, I had copied a test file into the root dir and was getting a confusing build failure for it (which I've now figured out) and had ascribed the real test failures to the same mystery cause. Fixed now.

brian-brazil:
Don't worry about it, it happens to us all.

storage/buffer.go
```diff
@@ -26,9 +26,25 @@ type BufferedSeriesIterator struct {
...
 }
 
+type nopIterator struct{}
+
+func (nopIterator) At() (int64, float64) { return math.MinInt64, 0 }
+func (nopIterator) Seek(t int64) bool    { return false }
+func (nopIterator) Next() bool           { return false }
+func (nopIterator) Err() error           { return nil }
+
+var nopIt = nopIterator{}
```
tomwilkie: You could put this in `storage/noop.go` with the other no-op implementation of the storage interfaces...

free: Done.

brian-brazil:
Thanks!

Issue #4293, opened 2018-06-20
When bad configuration defined prometheus should exit with non zero exit-code.


k0ste:
## Bug Report
**What did you do?**

Prometheus config have a error.

**What did you expect to see?**

Startup failed -> non zero exit-code.
This is important for automation tools and `systemd`.

**What did you see instead? Under which circumstances?**

Startup failed -> zero exit-code.

**Environment**

* System information:

```
Linux 4.11.12-1-nfcustom x86_64
```

* Prometheus version:

```
prometheus, version 2.3.1 (branch: non-git, revision: non-git)
  build user:       someman@linux01
  build date:       20180620-05:11:40
  go version:       go1.10.3
```

* Prometheus configuration file:
```yaml
- mybad:
```

* Logs:
```shell
[root@linux01 prometheus]# prometheus 
level=info ts=2018-06-20T05:43:40.583393016Z caller=main.go:222 msg="Starting Prometheus" version="(version=2.3.1, branch=non-git, revision=non-git)"
level=info ts=2018-06-20T05:43:40.583468472Z caller=main.go:223 build_context="(go=go1.10.3, user=someman@linux01, date=20180620-05:11:40)"
level=info ts=2018-06-20T05:43:40.58386063Z caller=main.go:224 host_details="(Linux 4.11.12-1-nfcustom #1 SMP PREEMPT Fri Nov 3 13:29:36 +07 2017 x86_64 linux01 (none))"
level=info ts=2018-06-20T05:43:40.584102954Z caller=main.go:225 fd_limits="(soft=1024, hard=4096)"
level=info ts=2018-06-20T05:43:40.585073298Z caller=main.go:514 msg="Starting TSDB ..."
level=info ts=2018-06-20T05:43:40.585125536Z caller=web.go:415 component=web msg="Start listening for connections" address=0.0.0.0:9090
level=info ts=2018-06-20T05:43:40.593193403Z caller=main.go:524 msg="TSDB started"
level=info ts=2018-06-20T05:43:40.593257652Z caller=main.go:603 msg="Loading configuration file" filename=prometheus.yml
level=info ts=2018-06-20T05:43:40.593524321Z caller=main.go:402 msg="Stopping scrape discovery manager..."
level=info ts=2018-06-20T05:43:40.59355105Z caller=main.go:416 msg="Stopping notify discovery manager..."
level=info ts=2018-06-20T05:43:40.593566994Z caller=main.go:438 msg="Stopping scrape manager..."
[...]

PR #4296, opened 2018-06-21
return error exit status in prometheus

mikeykhalil:
This should close this issue : https://github.com/prometheus/prometheus/issues/4293

### DIFF
--- a/cmd/prometheus/main_test.go
+++ b/cmd/prometheus/main_test.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


cmd/prometheus/main_test.go
```diff
@@ -155,3 +156,20 @@ func TestComputeExternalURL(t *testing.T) {
 		}
 	}
 }
+
+// Provide an invalid configuration file and verify the exit status indicates the error
```
brian-brazil: Comments should be full sentences and end in a full stop.

brian-brazil:
Thanks!

cmd/prometheus/main_test.go
brian-brazil: This is actually something that even varies across the Prometheus team, I wouldn't worry about it - especially for a tiny PR like this.

mikeykhalil:
Thanks for the quick review!

PR #4300, opened 2018-06-21
Check for timeout in each iteration of matrixSelector

### LINKED ISSUE #4288: Promql won't time-out queries in matrixSelector
## Bug Report
**What did you do?**
Someone sent a large query such as `{__name__=~".+"}[5d]` to our prometheus host.

I expect the query to timeout before completing, while consuming some memory and CPU along the way.

The query runs until it exhausts all available memory on the host and OOM killer get is. I have set it up in a repro env locally and the query has been running for ~60m (its still going) even though I have the timeout set to the default 2m.

**Environment**

* Prometheus version:

2.2

jacksontj:
Fixes #4288

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -1044,6 +1044,9 @@ func (ev *evaluator) matrixSelector(node *MatrixSelector) Matrix {
 
 	var it *storage.BufferedSeriesIterator
 	for i, s := range node.series {
+		if err := contextDone(ev.ctx, "expression evaluation"); err != nil {
+			ev.error(err)
+		}
 		if it == nil {
 			it = storage.NewBuffer(s.Iterator(), durationMilliseconds(node.Range))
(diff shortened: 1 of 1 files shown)


promql/engine.go
```diff
@@ -1044,6 +1044,10 @@ func (ev *evaluator) matrixSelector(node *MatrixSelector) Matrix {
 
 	var it *storage.BufferedSeriesIterator
 	for i, s := range node.series {
+		// check for timeout/cancellation here.
```
brian-brazil: No need for the comment

jacksontj:
Comment removed, CI looks good (except for the one travis test that has been failing for at least the last week).

brian-brazil:
Thanks!

We're aware of that, it's a resource issue that we're in the process of getting resolved.

Issue #2648, opened 2017-04-21
Remote queue manager should deal with blips in upstream latency


tomwilkie:
Currently the queue is per shard and only buffers a 100k samples; if you have a remote target that is usually very fast, but periodically slow, we will have one shard that can easily overwhelm its queue.

We could consider having a globally fixed length queue that gets partitioned by the number of shards.   Or we could make resharding happen when the queue is overwhelmed.  Or we could just accept dropping samples in this case.

danielshiplett:
More notes on what I was seeing.  The time to call my remote write was typically 80ms (0.999), 20ms (0.99), and 5ms (0.9).  Occasionally I would see blips where my 0.999 quantile was over 1s.

Since I've switched to using Undertow as my embedded server, I'm seeing far fewer blips to 1s.  I've also added more scrape targets and I would say that Prometheus is now actively adjusting shard counts.  I'm seeing no dropped records.  If this is due to the fewer blips, the change to Undertow, or the more active sharding in Prometheus, I have no idea.

I'll report more some next week after this has had a chance to soak in for a while.

danielshiplett:
So things are generally running better now.  I have 500,000 timeseries and I see about 300 remote write requests per second (almost 2 million samples per minute).  I still see occasional blips but they are occurring much less frequently.  My shards seem to bounce around between 4-6.

However, I've also noticed that when I do a configuration reload (POST to /-/reload) that my shards gets reset to 1.  Does Prometheus not save this information between runs?  When it gets reset to 1 I go through the process of losing more write requests until it works its shards back up again.

juliusv:
Yeah, config reloads are still fairly intrusive for the remote write path,
as that all gets completely flushed/reloaded on config reload. See also
this comment:
https://github.com/prometheus/prometheus/blob/master/storage/remote/write.go#L36-L37

tomwilkie:
We should indeed optimise this bit, it shouldn't be too hard.

poblahblahblah:
It looks like the queue configuration was addressed in https://sourcegraph.com/github.com/prometheus/prometheus/-/commit/454b6611458d01398678baca372a57973df4215f and then documented in https://github.com/prometheus/prometheus/pull/4126

tomwilkie:
I'd say we can close this: the new WAL based remote_write code in 2.8 includes a much more proactive resharding code that responds to upstream latency variations very quickly.

PR #4305, opened 2018-06-22
Reload remote write only on config change

gouthamve:
Right now it is quite disruptive, forgetting the old shard num and
starting from 1 causing prometheus to drop samples.

Related: https://github.com/prometheus/prometheus/issues/2648


storage/remote/storage.go
```diff
@@ -62,37 +63,49 @@ func (s *Storage) ApplyConfig(conf *config.Config) error {
...
 	for i, rwConf := range conf.RemoteWriteConfigs {
-		c, err := NewClient(i, &ClientConfig{
-			URL:              rwConf.URL,
-			Timeout:          rwConf.RemoteTimeout,
-			HTTPClientConfig: rwConf.HTTPClientConfig,
-		})
+		ok := i < len(oldQueues)
+		if ok {
+			qm := oldQueues[i]
+			if reflect.DeepEqual(conf.GlobalConfig.ExternalLabels, qm.externalLabels) &&
```
tomwilkie: I'm fine with using reflect.DeepEqual in a test, but not in actual production code.  model.LabelSet has a Equal function.

storage/remote/storage.go
```diff
@@ -62,37 +63,49 @@ func (s *Storage) ApplyConfig(conf *config.Config) error {
...
+				reflect.DeepEqual(rwConf, qm.rwConf) {
```
tomwilkie: Ditto; isf golang equality not enough? 

> Struct values are comparable if all their fields are comparable. Two struct values are equal if their corresponding non-blank fields are equal.

https://golang.org/ref/spec#Comparison_operators

storage/remote/storage.go
```diff
@@ -62,37 +63,49 @@ func (s *Storage) ApplyConfig(conf *config.Config) error {
 
 	// Update write queues
 
-	newQueues := []*QueueManager{}
+	newQueues := make([]*QueueManager, 0, len(s.queues))
+	oldQueues := s.queues
 	// TODO: we should only stop & recreate queues which have changes,
 	// as this can be quite disruptive.
```
tomwilkie: This TODO should be removed.

storage/remote/storage.go
```diff
@@ -120,6 +133,30 @@ func (s *Storage) ApplyConfig(conf *config.Config) error {
 	return nil
 }
 
+func (s *Storage) newQueue(i int, rwConf *config.RemoteWriteConfig, externalLabels model.LabelSet) (*QueueManager, error) {
+	c, err := NewClient(i, &ClientConfig{
```
tomwilkie: In keeping with the style of this file, can we use long-form variable names (like `client` in this case)?

tomwilkie:
Thanks for tackling this @gouthamve!  Something I've wanted to do for a while.

I think we can do better though: IIUC this PR will still stop & recreate all the queues when just the ordering has changed.  

I'd approach this by finding an identifier for each queue (perhaps the hash of its config), building two maps for old and new queues, and doing set-style arithmetic to stop the ones than have been remove (old - new) and start the new ones (new - old).

You'll also need to fix the identifiers used in the labels metrics exported by the queues: currently they are the index of the queue in the list of queues, which isn't kept constant by this change.  I suggest that you use the same identifier as above.

storage/remote/storage.go
juliusv: Golang equality for structs only works if there are no pointers or slices in the structs, which there are here. We could do a hash of all the fields, but then we have to maintain a hash function that is always updated to contain all the latest fields (including sub fields like relabel configs and HTTP client configs). So `reflect.DeepEqual()` is probably the best option. And as this is not a hot path...

juliusv:
Any overall news on this one?

gouthamve:
Sorry for going quiet here. So, while we could make it work for 100% of the cases, it adds a lot of complexity and there is high chance we might break it when we modify this code-path later (as Julius mentioned, by adding new fields.)

When config changes it is very rarely on the remote path, and when the changes do happen there, I think it is acceptable to restart the queues. This is simple and fixes >99% of the cases, and I'd like to roll with this WDYT @tom?

storage/remote/storage.go
```diff
@@ -62,42 +63,53 @@ func (s *Storage) ApplyConfig(conf *config.Config) error {
 
 	// Update write queues
 
-	newQueues := []*QueueManager{}
-	// TODO: we should only stop & recreate queues which have changes,
-	// as this can be quite disruptive.
+	newQueues := make([]*QueueManager, 0, len(s.queues))
```
juliusv: wouldn't `len(conf.RemoteWriteConfigs)` make more sense here for the new length?

storage/remote/storage.go
```diff
@@ -62,42 +63,53 @@ func (s *Storage) ApplyConfig(conf *config.Config) error {
...
+	oldQueues := s.queues
+
 	for i, rwConf := range conf.RemoteWriteConfigs {
-		c, err := NewClient(i, &ClientConfig{
-			URL:              rwConf.URL,
-			Timeout:          rwConf.RemoteTimeout,
-			HTTPClientConfig: rwConf.HTTPClientConfig,
-		})
+		ok := i < len(oldQueues)
```
juliusv: Why introduce an extra variable for this instead of just saying `if i < len(oldQueues)`?

storage/remote/storage.go
```diff
@@ -62,42 +63,53 @@ func (s *Storage) ApplyConfig(conf *config.Config) error {
...
+		if ok {
+			qm := oldQueues[i]
+			if conf.GlobalConfig.ExternalLabels.Equal(qm.externalLabels) &&
+				reflect.DeepEqual(rwConf, qm.rwConf) {
+
```
juliusv: Remove empty lines at start of block

storage/remote/storage.go
```diff
@@ -62,42 +63,53 @@ func (s *Storage) ApplyConfig(conf *config.Config) error {
...
-		))
+
+		newQueues = append(newQueues, qm)
 	}
 
-	for _, q := range s.queues {
-		q.Stop()
+	for i, nqm := range newQueues {
+		ok := i < len(oldQueues)
+		if ok {
```
juliusv: dito

gouthamve:
> Why introduce an extra variable

No clue what I was thinking then :) Fixed everything @juliusv

gouthamve:
Closing as too old

PR #4306, opened 2018-06-22
Make TestUpdate() do some work

bboreham:
Previously it would set no preconditions and check no postconditions, as the `groups` member was empty.

### DIFF
--- a/rules/fixtures/rules.yaml
+++ b/rules/fixtures/rules.yaml
@@ -1,0 +1,5 @@
+groups:
+  - name: test
+    rules:
+    - record: job:http_requests:rate5m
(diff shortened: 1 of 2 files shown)


rules/fixtures/rules.yaml
```diff
@@ -0,0 +1,5 @@
+groups:
+  - name: test
+    rules:
+    - record: job:http_requests:sum
+      expr: sum(http_requests) by (job)
```
brian-brazil: Let's be a little more realistic:  job:http_requests:rate5m = sum by (job)(rate(http_requests_total[5m]))

bboreham:
I am mystified by that CI failure.

brian-brazil:
Thanks!

PR #4312, opened 2018-06-26
rulemanager: block `Update`s after `Stop`ing

daveworth:
Simply block any further calls to the Rule Manager's `Update` method after `Stop` has been called.

This is a simple but also very naive approach @krasi-georgiev, @bboreham, and @brian-brazil but may work.  I would love to have used the `block` channel instead but I don't think it can block early enough in the `Update` routine to be effective.  Feels on something like this or is there something more sophisticated we'd like?

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -409,10 +409,11 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
 
 // The Manager manages recording and alerting rules.
 type Manager struct {
-	opts   *ManagerOptions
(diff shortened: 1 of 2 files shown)


brian-brazil:
I think this would work, but something up in main.go would be more general. It's likely we have this bug in other managers too.

krasi-georgiev:
+1 for adding this globally in main.go

brian-brazil:
Obseleted by #4321

PR #4315, opened 2018-06-26
Reuse (copy) overlapping matrix samples between range evaluation steps

free:
In many cases (such as recommended with the `rate` function) the evaluation range is significantly larger than one evaluation step and, in this situation, many of the samples between subsequent evaluation steps overlap. Starting with an empty slice at every step and copying over all samples one-by-one is significantly slower than doing a `copy()`, so doing the latter gives a performance boost of between ~25% with an overlap of 5 out of 6 samples (`rate[1m]`, 1000 10s steps, 10s resolution) and ~90% (`rate[1d]`, same step and resolution).

This only produces an improvement if the range covered by `BufferedSeriesIterator` is adjusted to the eval step starting with the second step, otherwise the `Next()` and `At()` calls need to be made for all overlapping samples, and this is where the slowness comes from to begin with.

Here are the [before](https://github.com/prometheus/prometheus/files/2137835/before.txt) and [after](https://github.com/prometheus/prometheus/files/2137838/after.matrix_slice.4.txt) benchmarks, as well as [benchcmp output](https://github.com/prometheus/prometheus/files/2137843/benchcmp.txt).


promql/engine.go
```diff
@@ -1074,6 +1081,23 @@ func (ev *evaluator) matrixIterSlice(it *storage.BufferedSeriesIterator, mint, m
 		}
 	}
 
+	if maxt-mint > ev.interval && len(out) > 0 && out[len(out)-1].T >= mint {
```
brian-brazil: This is a layering violation, you're making presumptions both about what's in `out` and poking around `ev`. This function should be standalone, and any preconditions clearly stated.

promql/engine.go
```diff
@@ -861,6 +866,8 @@ func (ev *evaluator) eval(expr Expr) Value {
 				if len(outVec) > 0 {
 					ss.Points = append(ss.Points, Point{V: outVec[0].Point.V, T: ts})
 				}
+				// Only buffer stepRange milliseconds from the second step on.
```
brian-brazil: This doesn't sound right at all. We need all of the range.

free: Assuming all data from the previous step is still in `points`, then strictly speaking we don't need any of the data before the previous step to also be buffered by `it`. We can "merge" the samples in `points` with the following `stepRange` worth of samples and end up with the same data that `it` with a delta of `selRange` would otherwise provide.

As stated in the PR description, it's iterating over all the overlapping data that slows things down, not so much copying it into `points`. So the intent is to essentially reduce the data buffered by `BufferedSeriesIterator` to the bare minimum (i.e. the minimum between `selRange` and `ev.interval`) from the second step on and complementing that data with what's already in the slice.

Ideally one would pass the iterator's buffer straight through to the eval logic (no copying involved as it's not, strictly speaking, needed) but that's not easily done without major refactoring.

promql/engine.go
free: It's undocumented and not particularly elegant, but I wouldn't go as far as calling it a violation of any kind. Simply put, if any data in `out` overlaps with the requested `[mint, maxt]` range, then it's kept, and any data retrieved from `it` is appended to it (with duplicates dropped).

I'll think of a better way of expressing this (in code and documentation).

free:
I have slightly reworked the change (but only slightly).

I got rid of "poking around `ev`", as you put it, as that was a misguided optimization and thus unnecessary. I have also clearly documented the preconditions for `matrixIterSlice`. I could rename it to `matrixIterSliceAppend`, to make it more obvious, if you think that helps. And finally, I'm buffering the initial delta value in `BufferedSeriesIterator`, so it doesn't have to be passed in on every `Reset` call.

If you have a suggestion for how to wrap this more nicely, I'm happy to oblige.

promql/engine.go
brian-brazil: That looks cleaner now. Still a tad convoluted to have the reducedelta call interacting with the matrixIterSlice.

free: Hmm... Now that you bring it up, I could have `matrixIterSlice` adjust the delta directly instead, but I'm not actually sure that would make it cleaner. Or safer.

brian-brazil: Yeah, as you say a deeper change so we don't have to do all this copying would be better but a bit much for now.

promql/bench_test.go
```diff
@@ -131,6 +135,9 @@ func BenchmarkRangeQuery(b *testing.B) {
 		{
 			expr: "label_join(a_X, 'l2', '-', 'l', 'l')",
 		},
+		{
+			expr: "max_over_time(a_X[10s])",
```
brian-brazil: This is duplicate with the rate() tests.

free: It's not, actually. It was meant as a benchmark of the reuse logic in this PR (to test whether there was significant slowdown vs. the previous implementation when there was no actual overlap between steps and thus the newly added code was purely overhead. (There was no significant difference, even though it doesn't appear in the benchmark outputs I had attached above.)

Since that was specific to this PR though, I've now dropped it altogether.

free:
Conflicts resolved.

I did add a comment to #4226 though, on what I perceived to be a regression introduced by that change (which I noticed during merging).

brian-brazil:
Thanks!

PR #4316, opened 2018-06-27
Add services metadata labels to consul labels

roidelapluie:


roidelapluie:
Not sure how I can test it because we populate that on watch() and watch() is not run in unit tests

roidelapluie:
+65,452 −299
 I am speechless.

PR #4012, opened 2018-03-25
api v1 alerts/rules json endpoint

mg03:

PR #4318, opened 2018-06-27
api/v1: Expose rules and alerts

mxinden:
This PR builds on top of @mg03 work and is a follow up to https://github.com/prometheus/prometheus/pull/4012:
- Addressing outstanding comments on https://github.com/prometheus/prometheus/pull/4012
- Adjusting to Prometheus coding style
- Differentiating between alerting and recording rules
- ...

The first commit contains @mg03 original changes rebased to current `master`. The second commit contains all follow up changes by me.

Let me know what you think.

//CC @brancz @kyoto

### DIFF
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -363,6 +363,103 @@ $ curl http://localhost:9090/api/v1/targets
 }
 ```
 
+
+## Rules
+
+The `/rules` API endpoint returns a list of alerting and recording rules that
+are currently loaded. In addition it returns the currently active alerts fired
+by the Prometheus instance of each alerting rule.
+
+As the `/rules` endpoint is fairly new, it does not have the same stability
+guarantees as the overarching API v1.
+
+```
+GET /api/v1/rules
+```
+
+```json
+$ curl http://localhost:9090/api/v1/rules
+
+{
+    "data": {
+        "groups": [
+            {
+                "rules": [
+                    {
+                        "alerts": [
+                            {
+                                "activeAt": "2018-07-04T20:27:12.60602144+02:00",
+                                "annotations": {
+                                    "summary": "High request latency"
+                                },
+                                "labels": {
+                                    "alertname": "HighRequestLatency",
+                                    "severity": "page"
+                                },
+                                "state": "firing",
+                                "value": 1
+                            }
+                        ],
+                        "annotations": {
+                            "summary": "High request latency"
+                        },
+                        "duration": 600,
+                        "labels": {
+                            "severity": "page"
+                        },
+                        "name": "HighRequestLatency",
+                        "query": "job:request_latency_seconds:mean5m{job=\"myjob\"} > 0.5",
+                        "type": "alerting"
+                    },
+                    {
+                        "name": "job:http_inprogress_requests:sum",
+                        "query": "sum(http_inprogress_requests) by (job)",
+                        "type": "recording"
+                    }
+                ],
+                "file": "/rules.yaml",
+                "interval": 60,
+                "name": "example"
+            }
+        ]
+    },
+    "status": "success"
+}
+```
+
+
+## Alerts
+
+The `/alerts` endpoint returns a list of all active alerts.
+
+As the `/alerts` endpoint is fairly new, it does not have the same stability
+guarantees as the overarching API v1.
+
+```
+GET /api/v1/alerts
+```
+
+```json
+$ curl http://localhost:9090/api/v1/alerts
+
+{
+    "data": {
+        "alerts": [
+            {
+                "activeAt": "2018-07-04T20:27:12.60602144+02:00",
+                "annotations": {},
+                "labels": {
+                    "alertname": "my-alert"
+                },
+                "state": "firing",
+                "value": 1
+            }
+        ]
+    },
+    "status": "success"
+}
+```
+
 ## Querying target metadata
 
 The following endpoint returns metadata about metrics currently scraped by targets.
--- a/rules/alerting.go
+++ b/rules/alerting.go
@@ -126,44 +126,29 @@ func NewAlertingRule(name string, vec promql.Expr, hold time.Duration, lbls, ann
 	}
 }
 
-// Name returns the name of the alert.
+// Name returns the name of the alerting rule.
 func (r *AlertingRule) Name() string {
 	return r.name
 }
 
-// Query returns the query expression of the alert.
+// Query returns the query expression of the alerting rule.
 func (r *AlertingRule) Query() promql.Expr {
 	return r.vector
 }
 
-// Duration returns the hold duration of the alert.
+// Duration returns the hold duration of the alerting rule.
 func (r *AlertingRule) Duration() time.Duration {
 	return r.holdDuration
 }
 
-// Labels returns the labels of the alert.
+// Labels returns the labels of the alerting rule.
 func (r *AlertingRule) Labels() labels.Labels {
 	return r.labels
 }
 
-// Annotations returns the annotations of the alert.
+// Annotations returns the annotations of the alerting rule.
 func (r *AlertingRule) Annotations() labels.Labels {
 	return r.annotations
-}
-
-// Alertinfo return an array of alerts
-func (r *AlertingRule) Alertinfo() []*Alert {
-	activealerts := &r.active
-	alertsarr := make([]*Alert, 0)
-	if len(*activealerts) > 0 {
-		for _, a := range *activealerts {
-			if a.ResolvedAt.IsZero() {
-				alertsarr = append(alertsarr, a)
-			}
-		}
-		return alertsarr
-	}
-	return nil
 }
 
 func (r *AlertingRule) equal(o *AlertingRule) bool {
--- a/rules/alerting.go
+++ b/rules/alerting.go
@@ -131,6 +131,41 @@ func (r *AlertingRule) Name() string {
 	return r.name
 }
 
+// Query returns the query expression of the alert.
+func (r *AlertingRule) Query() promql.Expr {
+	return r.vector
+}
+
+// Duration returns the hold duration of the alert.
+func (r *AlertingRule) Duration() time.Duration {
+	return r.holdDuration
+}
+
+// Labels returns the labels of the alert.
+func (r *AlertingRule) Labels() labels.Labels {
+	return r.labels
+}
+
+// Annotations returns the annotations of the alert.
+func (r *AlertingRule) Annotations() labels.Labels {
+	return r.annotations
+}
+
+// Alertinfo return an array of alerts
+func (r *AlertingRule) Alertinfo() []*Alert {
+	activealerts := &r.active
+	alertsarr := make([]*Alert, 0)
+	if len(*activealerts) > 0 {
+		for _, a := range *activealerts {
+			if a.ResolvedAt.IsZero() {
+				alertsarr = append(alertsarr, a)
+			}
+		}
+		return alertsarr
+	}
+	return nil
+}
+
 func (r *AlertingRule) equal(o *AlertingRule) bool {
 	return r.name == o.name && labels.Equal(r.labels, o.labels)
 }
--- a/rules/recording.go
+++ b/rules/recording.go
@@ -52,6 +52,16 @@ func (rule *RecordingRule) Name() string {
 	return rule.name
 }
 
+// Query returns the rule query expression.
+func (rule *RecordingRule) Query() promql.Expr {
+	return rule.vector
+}
+
+// Labels returns the rule labels.
+func (rule *RecordingRule) Labels() labels.Labels {
+	return rule.labels
+}
+
 // Eval evaluates the rule and then overrides the metric names and labels accordingly.
 func (rule *RecordingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc, _ *url.URL) (promql.Vector, error) {
 	vector, err := query(ctx, rule.vector.String(), ts)
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -41,6 +41,7 @@ import (
 	"github.com/prometheus/prometheus/pkg/timestamp"
 	"github.com/prometheus/prometheus/prompb"
 	"github.com/prometheus/prometheus/promql"
+	"github.com/prometheus/prometheus/rules"
 	"github.com/prometheus/prometheus/scrape"
 	"github.com/prometheus/prometheus/storage"
 	"github.com/prometheus/prometheus/storage/remote"
@@ -95,6 +96,19 @@ type alertmanagerRetriever interface {
 	DroppedAlertmanagers() []*url.URL
 }
 
+type alertRetreiver interface {
+	AlertingRules() []*rules.AlertingRule
+}
+
+type rulesRetreiver interface {
+	RuleGroups() []*rules.Group
+}
+
+type alertsrulesRetreiver interface {
+	alertRetreiver
+	rulesRetreiver
+}
+
 type response struct {
 	Status    status      `json:"status"`
 	Data      interface{} `json:"data,omitempty"`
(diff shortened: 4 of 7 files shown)


rules/alerting.go
```diff
@@ -131,6 +131,26 @@ func (r *AlertingRule) Name() string {
...
+func (r *AlertingRule) Query() promql.Expr {
+	return r.vector
+}
+
+// Duration returns the hold duration of the alert.
+func (r *AlertingRule) Duration() time.Duration {
+	return r.holdDuration
+}
+
+// Labels returns the labels of the alert.
```
brian-brazil: alerting rule

alert labels are a bit different due to templating

rules/recording.go
```diff
@@ -52,6 +52,16 @@ func (rule *RecordingRule) Name() string {
 	return rule.name
 }
 
+// Query returns the rule query expression
```
brian-brazil: Full stop

web/api/v1/api.go
```diff
@@ -574,6 +585,130 @@ func (api *API) alertmanagers(r *http.Request) (interface{}, *apiError, func())
 	return ams, nil, nil
 }
 
+// AlertDiscovery has info for all alerts
+type AlertDiscovery struct {
+	AlertGroups []*AlertGroup `json:"alertGroups"`
+}
+
+// Alert has info for a alert
```
brian-brazil: an alert

web/api/v1/api.go
```diff
@@ -574,6 +585,130 @@ func (api *API) alertmanagers(r *http.Request) (interface{}, *apiError, func())
...
+type Alert struct {
+	Labels      labels.Labels `json:"labels"`
+	Annotations labels.Labels `json:"annotations"`
+	State       string        `json:"state"`
+	ActiveAt    *time.Time    `json:"activeAt,omitempty"`
```
brian-brazil: We might want to be a little careful with this one, its semantics may change over time.

web/api/v1/api.go
```diff
@@ -574,6 +585,130 @@ func (api *API) alertmanagers(r *http.Request) (interface{}, *apiError, func())
...
+}
+
+// AlertGroup has info for alerts part of a group
+type AlertGroup struct {
+	Name        string        `json:"name"`
+	Query       string        `json:"query"`
+	Duration    string        `json:"duration"`
+	Annotations labels.Labels `json:"annotations,omitempty"`
```
brian-brazil: No labels?

web/api/v1/api.go
```diff
@@ -574,6 +585,130 @@ func (api *API) alertmanagers(r *http.Request) (interface{}, *apiError, func())
...
+// Alert has info for a alert
+type Alert struct {
+	Labels      labels.Labels `json:"labels"`
+	Annotations labels.Labels `json:"annotations"`
+	State       string        `json:"state"`
+	ActiveAt    *time.Time    `json:"activeAt,omitempty"`
+}
+
+// AlertGroup has info for alerts part of a group
+type AlertGroup struct {
```
brian-brazil: An alert group is something we have in the alertmanager. This is the output of an alerting rule.

web/api/v1/api.go
```diff
@@ -574,6 +585,130 @@ func (api *API) alertmanagers(r *http.Request) (interface{}, *apiError, func())
...
+}
+
+func (api *API) alerts(r *http.Request) (interface{}, *apiError, func()) {
+	alertingRules := api.rulesRetriever.AlertingRules()
+	var alertGroups []*AlertGroup
+
+	for _, alertingRule := range alertingRules {
+		alertGroup := &AlertGroup{
+			Name:        alertingRule.Name(),
+			Query:       fmt.Sprintf("%v", alertingRule.Query()),
```
brian-brazil: Do you really need the sprintf here?

mxinden:
> The docs need updating.

@brian-brazil Where would you like the two endpoints to be documented? `querying/api.md` seems to only be related to the `/query` api endpoint.

In addition, should I add a comment to `stability.md` mentioning that this endpoint is subject to change even though it is part of the v1 API?

brian-brazil:
api.md is where we put everything. A comment in api.md is sufficient.

web/api/v1/api.go
mxinden: Good point. Should I rename this to `AlertingRule` or should we flatten (/denormalize) the endpoint and expose a list of alerts right away?

brian-brazil: Flattening is probably easiest.

web/api/v1/api.go
mxinden: Good point. Added below. :heavy_check_mark:

brian-brazil:
That's insufficient to identify a rule, you can have multiple of the same name in a group.

> Tying an alert and an alerting rule together seems like a valid use case to me.

Noone has ever asked for it that I can recall.

> Adding filename and rule group name to each alert exposed on the /alerts endpoint.

That could get rather large size wise.

> Internally adding a filename and groupname label to each alert, as we already do with the alertname.

This would be a breaking change for both ALERTs and alerting generally. From a semantic standpoint it doesn't make sense, you don't care where the alert came from just that it is firing.

brancz:
The ability to tell which rule produced which alert seems reasonable to me (and is a use case we have).

How about we don't attach that information by default but make it available via alert relabeling, similar to optional discovery information as `__meta` labels? If I understand you correctly @brian-brazil your comment on this being a breaking change only refers to the set of labels that make up an alert (which I agree we absolutely cannot break the default behavior).

mxinden:
What are your thoughts on option number three?

> Nesting all active alerts inside the `/rules` endpoint in each respective rules object.

This would result in a rather deep nesting, but enables us to tie each alert to its alerting rule.

brian-brazil:
> How about we don't attach that information by default but make it available via alert relabeling, similar to optional discovery information as __meta labels?

I don't see how that helps, this should never hit the alertmanager.

brancz:
> I don't see how that helps, this should never hit the alertmanager.

Hmm I see your point. It's easy enough to identify the source Prometheus via external labels anyways. And then it would only need to be possible to have Prometheus tell which rule produced an alert (but never sending this information to Alertmanager). I think ideally it would be nice if (within Prometheus) we could tell the alert(s) a specific rule is creating and vice versa which alert was produced by which rule.

Returning the respective rule to an alert could be optional, if we're worried about the amount of data returned.

mxinden:
Flattened `/alerts` endpoint and nested active alerts of each rule into the `/rules` endpoint.

You find a sample endpoint payload in the `api.md` changes.

web/api/v1/api.go
```diff
@@ -574,6 +585,120 @@ func (api *API) alertmanagers(r *http.Request) (interface{}, *apiError, func())
 	return ams, nil, nil
 }
 
+// AlertDiscovery has info for all active alerts.
+type AlertDiscovery struct {
+	ActiveAlerts []*Alert `json:"activeAlerts"`
```
mxinden: Happy to rename this back to `Alerts`. `ActiveAlerts` gives the user a hint, that resolved alerts are not included.

web/api/v1/api.go
```diff
@@ -574,6 +585,120 @@ func (api *API) alertmanagers(r *http.Request) (interface{}, *apiError, func())
...
+	Interval       time.Duration   `json:"interval"`
+}
+
+type alertingRule struct {
+	Name         string        `json:"name"`
+	Query        string        `json:"query"`
+	Duration     time.Duration `json:"duration"`
+	Labels       labels.Labels `json:"labels"`
+	Annotations  labels.Labels `json:"annotations"`
+	ActiveAlerts []*Alert      `json:"activeAlerts"`
```
mxinden: Same as above: Happy to rename this to `Alerts`. `ActiveAlerts` gives the user a hint, that resolved alerts are not included.

web/api/v1/api_test.go
```diff
@@ -98,6 +101,74 @@ func (t testAlertmanagerRetriever) DroppedAlertmanagers() []*url.URL {
...
+	)
+	rule2 := rules.NewAlertingRule(
+		"test_metric4",
+		expr2,
+		time.Second,
+		labels.Labels{},
+		labels.Labels{},
+		log.NewNopLogger(),
+	)
+	// TODO: Should also contain test for recording rule?
```
simonpasquier: Is it still valid?

mxinden:
@simonpasquier @brian-brazil @brancz thanks for your comments. Would you mind taking another look?

docs/querying/api.md
```diff
@@ -363,6 +363,91 @@ $ curl http://localhost:9090/api/v1/targets
...
+```
+
+```json
+$ curl http://localhost:9090/api/v1/rules
+
+{
+    "data": {
+        "groups": [
+            {
+                "alertingRules": [
```
brian-brazil: There is an ordering across all rules, the alert recording and recording rules need to be in the same list.

docs/querying/api.md
```diff
@@ -363,6 +363,91 @@ $ curl http://localhost:9090/api/v1/targets
...
+                        ],
+                        "annotations": {},
+                        "duration": 0,
+                        "labels": {},
+                        "name": "my-alert",
+                        "query": "vector(1)"
+                    }
+                ],
+                "file": "",
+                "interval": 60000000000,
```
brian-brazil: we use seconds

web/api/v1/api.go
brian-brazil: Yes. Users also depend on these values in ways that aren't correct, these are implementation details more than anything.

docs/querying/api.md
mxinden: But giving up the `alertingRule` and recordingRule` bucket prevents us from differentiating the two, thereby we are not able to expose details like `Duration`, `Annotations`, active `Alerts`, ...

In which use case is the ordering of rules important?

brian-brazil: In any use case where there are interdependent rules and alerts.

web/api/v1/api.go
```diff
@@ -574,6 +585,121 @@ func (api *API) alertmanagers(r *http.Request) (interface{}, *apiError, func())
...
+		)
+	}
+
+	res := &AlertDiscovery{Alerts: alerts}
+
+	return res, nil, nil
+}
+
+func rulesAlertsToAPIAlerts(rulesAlerts []*rules.Alert) []*Alert {
+	apiAlerts := []*Alert{}
```
simonpasquier: super minor but it could be `apiAlerts := make([]*Alert{}, len(rulesAlerts))`

docs/querying/api.md
simonpasquier: I think @brian-brazil has a point. I see several options.

1) add a `type` field that would be `alerting` or `recording`.

```json
{
    "data": {
        "groups": [
            {
                "rules": [
                    {
                        "alerts": [
                            {
                                "activeAt": "2018-07-04T20:27:12.60602144+02:00",
                                "annotations": {},
                                "labels": {
                                    "alertname": "my-alert"
                                },
                                "state": "firing",
                                "value": 1
                            }
                        ],
                        "annotations": {},
                        "duration": 0,
                        "labels": {},
                        "name": "my-alert",
                        "query": "vector(1)",
                        "type": "alerting"
                    }, {
                        "duration": 0,
                        "labels": {},
                        "name": "my-record",
                        "query": "vector(0)",
                        "type": "recording"
                    }
                ],
                "file": "",
                "interval": 60,
                "name": ""
            }
        ]
    },
    "status": "success"
}
```
2) Keep your proposal and add an `index` field.

```json
{
    "data": {
        "groups": [
            {
                "alertingRules": [
                    {
                        "alerts": [
                            {
                                "activeAt": "2018-07-04T20:27:12.60602144+02:00",
                                "annotations": {},
                                "labels": {
                                    "alertname": "my-alert"
                                },
                                "state": "firing",
                                "value": 1
                            }
                        ],
                        "annotations": {},
                        "duration": 0,
                        "index": 0,
                        "labels": {},
                        "name": "my-alert",
                        "query": "vector(1)"
                    }
                ],
                "file": "",
                "interval": 60,
                "name": "",
                "recordingRules": [{
                        "duration": 0,
                        "index": 1,
                        "labels": {},
                        "name": "my-record",
                        "query": "vector(0)",
                        "type": "recording"
                    }
                ]
            }
        ]
    },
    "status": "success"
}
```

I'd go with the first option as alerting rules are a superset of recording rules.

mxinden:
@brian-brazil @simonpasquier Thanks for the hint on the ordering. Both alerting and recording rules are exposed in the same `rules` array now, differentiating the two via a `type` property.

Further comments / suggestions?

docs/querying/api.md
```diff
@@ -363,6 +363,98 @@ $ curl http://localhost:9090/api/v1/targets
...
+                        "annotations": {},
+                        "duration": 0,
+                        "labels": {},
+                        "name": "my-alerting-rule",
+                        "query": "vector(1)",
+                        "type": "alerting"
+
+                    },
+                    {
+                        "labels": {},
```
brian-brazil: Should we have an omitempty on this? Labels aren't typical on recording rules, and having this here could be a bit confusing.

docs/querying/api.md
```diff
@@ -363,6 +363,98 @@ $ curl http://localhost:9090/api/v1/targets
...
+                        "name": "my-recording-rule",
```
brian-brazil: Can we use a valid metric name?

docs/querying/api.md
```diff
@@ -363,6 +363,98 @@ $ curl http://localhost:9090/api/v1/targets
...
+        "groups": [
+            {
+                "rules": [
+                    {
+                        "alerts": [
+                            {
+                                "activeAt": "2018-07-04T20:27:12.60602144+02:00",
+                                "annotations": {},
+                                "labels": {
+                                    "alertname": "my-alert"
```
brian-brazil: This is inconsistent with the `name`. MyAlertingRule is more typical.

web/api/v1/api.go
```diff
@@ -574,6 +585,132 @@ func (api *API) alertmanagers(r *http.Request) (interface{}, *apiError, func())
...
+type rule interface{}
+
+type alertingRule struct {
+	Name        string        `json:"name"`
+	Query       string        `json:"query"`
+	Duration    time.Duration `json:"duration"`
+	Labels      labels.Labels `json:"labels"`
+	Annotations labels.Labels `json:"annotations"`
+	Alerts      []*Alert      `json:"alerts"`
+	// Type of an alertingRule is always "alerting"
```
brian-brazil: Full stop

mxinden:
@brian-brazil thanks for the comments! I very much appreciated the help.

I included the following changes:https://prometheus.io/docs/prometheus/latest/configuration/
- Updated the examples now derived from the [Prometheus docs](https://prometheus.io/docs/prometheus/latest/configuration/).
- Add `omitempty` struct tag to `Labels` of a recording rule.
- Change `alertingRule.Duration` from `time.Duration` to `float64` using base unit seconds.

Let me know if you have further suggestions.

In addition, we need to think about how we want to proceed in regards to the DCO check. As the first commit is not authored by me, I don't want to sign it. I would suggest a PR approval by @mg03. What do you think @brian-brazil?

brian-brazil:
That sounds reasonable to me. You will need to adjust the commit message.

mxinden:
> You will need to adjust the commit message.

@brian-brazil Do you suggest me to add `Signed-off-by: Max Leonard Inden <IndenML@gmail.com>` to the first commit authored by @mg03? That sounds a bit odd to me.

brian-brazil:
No, signed off by mg03's email if they approve.

mg03:
@mxinden @brian-brazil : I approve. Thanx.

Let me know if anything else is required

mxinden:
Thanks @mg03!

I have adjusted your commit message:

```git
commit 31f8ca0dfb497f12404aba3108440f79b3015659
Author: mg03 <mgeng03@gmail.com>
Date:   Sun Mar 25 09:50:34 2018 -0700

    api v1 alerts/rules json endpoint
    
    Signed-off-by: mg03 <mgeng03@gmail.com>
```

@brian-brazil @simonpasquier anything else missing here?

simonpasquier:
LGTM thanks @mg03 and @mxinden for the hard work!

mxinden:
@brian-brazil looks good to you as well? Can we include this in `v2.4.0`?

brian-brazil:
:+1: 

Everything in master gets included in the next minor release.

mxinden:
Thanks @simonpasquier @brian-brazil and @mg03.

danny311296:
Is there any way to get a lost of all times of alert occurrences from the API.
For example it an alert1 occurred at time t1,t2 and t3,
some sort of API that returns the json with the various timestamps at which rules where triggered.
My use case requires to use prometheus as a time series memory store with a flask backend to handle connecting to prometheus, querying, plotting alerts, ect

PR #4321, opened 2018-06-28
Reorder startup and shutdown to prevent panics.

### LINKED ISSUE #4133: Rules manager Panic caused by closing a channel more than once.
caused when sending a SIGHUP. and  SIGTERM right after that.
I looked briefly and seems like an easy one to solve.

```
panic: close of closed channel

goroutine 86182 [running]:
github.com/prometheus/prometheus/rules.(*Group).stop(...)
      ..../src/github.com/prometheus/prometheus/rules/manager.go:238
github.com/prometheus/prometheus/rules.(*Manager).Update.func1(0xc58eeacc01, 0xc42842c480, 0xc4200c5bc0, 0xc53ad459e0, 0xc44f874a20)
      ..../src/github.com/prometheus/prometheus/rules/manager.go:501 +0x80
created by github.com/prometheus/prometheus/rules.(*Manager).Update
      ..../src/github.com/prometheus/prometheus/rules/manager.go:499 +0x512
panic: close of closed channel

goroutine 86140 [running]:
github.com/prometheus/prometheus/rules.(*Group).stop(...)
      ..../src/github.com/prometheus/prometheus/rules/manager.go:238
github.com/prometheus/prometheus/rules.(*Manager).Update.func1(0xc4fab69b01, 0xc428228b40, 0xc4200c5bc0, 0xc53ad459e0, 0xc4e381def0)
      ..../src/github.com/prometheus/prometheus/rules/manager.go:501 +0x80
created by github.com/prometheus/prometheus/rules.(*Manager).Update
      ..../src/github.com/prometheus/prometheus/rules/manager.go:499 +0x512
```

### LINKED ISSUE #4262: remote_write: send on closed channel
## Bug Report

Running Prometheus 2.3.0 with graphite-remote-adapter and stopping it:
```
level=info ts=2018-06-13T06:49:24.26401728Z caller=queue_manager.go:258 component=remote queue="0:http://prometheus-meta-graphite-remote-adapter/write?graphite.default-prefix=sre.network.services.prometheus." msg="Stopping remote storage..."
level=info ts=2018-06-13T06:49:25.680587091Z caller=queue_manager.go:266 component=remote queue="0:http://prometheus-meta-graphite-remote-adapter/write?graphite.default-prefix=sre.network.services.prometheus." msg="Remote storage stopped."
level=info ts=2018-06-13T06:49:25.680647131Z caller=manager.go:464 component="rule manager" msg="Stopping rule manager..."
panic: send on closed channel

goroutine 932 [running]:
github.com/prometheus/prometheus/storage/remote.(*shards).enqueue(0xc42017b0e0, 0xc42e688c00, 0x5)
        /go/src/github.com/prometheus/prometheus/storage/remote/queue_manager.go:438 +0x8b
github.com/prometheus/prometheus/storage/remote.(*QueueManager).Append(0xc4202e41e0, 0xc439343918, 0x8, 0xc437844090)
        /go/src/github.com/prometheus/prometheus/storage/remote/queue_manager.go:222 +0x26e
github.com/prometheus/prometheus/storage/remote.(*Storage).Add(0xc420193420, 0xc429089100, 0x5, 0x8, 0x163f7e7f3cc, 0x0, 0x0, 0x0, 0x0)
        /go/src/github.com/prometheus/prometheus/storage/remote/write.go:32 +0x11a
github.com/prometheus/prometheus/storage.(*fanoutAppender).Add(0xc426cafa40, 0xc429089100, 0x5, 0x8, 0x163f7e7f3cc, 0x0, 0x1c85b, 0x0, 0x0)
        /go/src/github.com/prometheus/prometheus/storage/fanout.go:140 +0x131
github.com/prometheus/prometheus/rules.(*Group).Eval.func1(0x1decdc0, 0xc420042118, 0x275dbd39, 0xed2b2b573, 0x2afd680, 0xc4208fa870, 0x10, 0x1df69e0, 0xc42109fcc0)
        /go/src/github.com/prometheus/prometheus/rules/manager.go:365 +0x6f8
github.com/prometheus/prometheus/rules.(*Group).Eval(0xc4208fa870, 0x1decdc0, 0xc420042118, 0x275dbd39, 0xed2b2b573, 0x2afd680)
        /go/src/github.com/prometheus/prometheus/rules/manager.go:406 +0xad
github.com/prometheus/prometheus/rules.(*Group).run.func1()
        /go/src/github.com/prometheus/prometheus/rules/manager.go:206 +0xc0
github.com/prometheus/prometheus/rules.(*Group).run(0xc4208fa870, 0x1decdc0, 0xc420042118)
        /go/src/github.com/prometheus/prometheus/rules/manager.go:235 +0x2e6
github.com/prometheus/prometheus/rules.(*Manager).Update.func1.1(0xc420179340, 0xc4208fa870)
        /go/src/github.com/prometheus/prometheus/rules/manager.go:509 +0x60
created by github.com/prometheus/prometheus/rules.(*Manager).Update.func1
        /go/src/github.com/prometheus/prometheus/rules/manager.go:504 +0x56
```

**Environment**

```
Linux 4.13.0-45-generic x86_64
prometheus, version 2.3.0 (branch: HEAD, revision: 290d71791a507a5057b9a099c9d48703d86dc941)
  build user:       root@d539e167976a
  build date:       20180607-08:46:54
  go version:       go1.10.2
```

brian-brazil:
Start rule manager only after tsdb and config is loaded.
Stop rule manager before tsdb to avoid writing to closed storage.
Wait for any in-progress reloads to complete before shutting
down rule manager, so that rule manager doesn't get updated after
being shut down.

Remove incorrect comment around shutting down query enginge.
Log when config reload is completed.

Fixes #4133
Fixes #4262

I haven't given this a lot of testing, but the idea seems right.

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -472,7 +472,9 @@ func main() {
 
 			},
 			func(err error) {
-				close(cancel)
+				// Wait for any in-progress reloads to complete to avoid
+				// reloading things after they have been shutdown.
+				cancel <- struct{}{}
 			},
 		)
 	}
@@ -502,6 +504,23 @@ func main() {
 				return nil
 			},
 			func(err error) {
+				close(cancel)
+			},
+		)
+	}
+	{
+		// Rule manager.
+		// TODO(krasi) refactor ruleManager.Run() to be blocking to avoid using an extra blocking channel.
+		cancel := make(chan struct{})
+		g.Add(
+			func() error {
+				<-reloadReady.C
+				ruleManager.Run()
+				<-cancel
+				return nil
+			},
+			func(err error) {
+				ruleManager.Stop()
 				close(cancel)
 			},
 		)
@@ -547,27 +566,7 @@ func main() {
 				return nil
 			},
 			func(err error) {
-				// Keep this interrupt before the ruleManager.Stop().
-				// Shutting down the query engine before the rule manager will cause pending queries
-				// to be canceled and ensures a quick shutdown of the rule manager.
 				cancelWeb()
-			},
-		)
-	}
-	{
-		// Rule manager.
-
-		// TODO(krasi) refactor ruleManager.Run() to be blocking to avoid using an extra blocking channel.
-		cancel := make(chan struct{})
-		g.Add(
-			func() error {
-				ruleManager.Run()
-				<-cancel
-				return nil
-			},
-			func(err error) {
-				ruleManager.Stop()
-				close(cancel)
 			},
 		)
 	}
@@ -626,6 +625,7 @@ func reloadConfig(filename string, logger log.Logger, rls ...func(*config.Config
 	if failed {
 		return fmt.Errorf("one or more errors occurred while applying the new configuration (--config.file=%s)", filename)
 	}
+	level.Info(logger).Log("msg", "Completed loading of configuration file", "filename", filename)
 	return nil
 }
 


cmd/prometheus/main.go
```diff
@@ -472,7 +472,9 @@ func main() {
 
 			},
 			func(err error) {
-				close(cancel)
+				// Wait for any in-progress reloads to complete to avoid
+				// reloading things after they have been shutdown.
+				cancel <- struct{}{}
```
fabxc: I don't quite understand what this changes. This will make this interrupt function block until the above function actually reached the point where it reads from `cancel`. But what benefit does that give us?

brian-brazil: It ensures that there will never be a reload after that read.

fabxc: I don't follow. Once the above function receives something from cancel it returns immediately.
Whether that happens through closing `cancel` or sending a value on it doesn't matter.

brian-brazil: It's an unbuffered channel, so if there's a reload in progress it'll block.

brian-brazil: Interrupt functions are called sequentially, without concurrency.

fabxc: That's an implementation detail, and I'd think @peterbourgon would object that this is something to rely on. Maybe he can provide more context here.

https://godoc.org/github.com/oklog/run#Group.Run

fabxc: I take this is also the reason why you changed the order in which `Group.Add`s are called?

brian-brazil: Yes, to control the shutdown order.

juliusv: I also think that we shouldn't depend on the implementation details of `Group.Run()` and same for the ordering of `Group.Add()`. That's too implicit to me and easy to screw up when not paying close attention.

It'd be better to impose those ordering constraints explicitly externally, like we do with some of the other ordering channels already.

brian-brazil: I was under the impression that the whole point of switching to this library was to get rid of the ad-hoc approaches we used previously. Are you proposing we get rid of this library and go back to doing all this ourselves?

juliusv: Maybe it's a good opportunity for someone to create a library that allows more explicit modeling of ordering constraints (half serious) :)

I like the library usage here already, but when it comes to ordering, it just doesn't give you any tools (besides implementation details), so we do still have to do that ourselves.

brian-brazil: Reading the docs more closely, this is the intended behaviour. The first line is "Package run implements an actor-runner with *deterministic teardown*", which you can only offer if the interrupts are sequential and synchronous.

peterbourgon: Before I get into detail, one clarification: it looks like you're trying to prevent subsequent components, like ruleManager, from _being interrupted_ until the reloadHandler's execute func has returned. Is that correct?

brian-brazil: Basically yes. The rulemanager shouldn't shutdown until the reload handler has.

peterbourgon: OK, thanks. This is a super gray area.

First, I didn't mean _deterministic_ as in _in a specific order_ but rather _in any order, but guaranteed_ — I'll push a doc update to make that more clear. So you are indeed relying on implementation details at the moment. This implementation detail is the only sane way to accomplish the work, so it could feasibly be lifted into the contract of the API. But I'm hesitant to do that, because of situations precisely like this one.

From my admittedly rather strict perspective, if ruleManager shouldn't terminate until reloadHandler has terminated, then that causal dependency should be expressed explicitly in the components themselves. A strawman for argument's sake:

```go
type ruleManager struct {
    reloadHandler interface{ Done() <-chan struct{} } // or whatever
    // ...
}

func (rm *ruleManager) Run() {
    defer func() {
        <-rm.reloadHandler.Done()
    }()
    // ...
}
```

I think this is strictly superior, especially for future maintainers, who now understand the required causality rules explicitly, without having to build a mental model that relies on e.g. source code order and guessing at the behavior of the run.Group black box.

But I say it's a gray area, because sometimes the practicalities of building software outweigh a theoretical soundness. If you were whipping up a prototype and wanted to ensure a certain startup/shutdown order, maybe it's fine to do that by leveraging details of your component lifecycle manager package. Maybe that gets the job done reliably, and you can move on to the next thing.

Which kind of brings me full-circle, actually. If you're writing that prototype, then implementation details are a fine-enough "contract" for you to leverage in your code. And if you're writing a more substantial project with lots of maintainers, then I don't want to encourage you to encode dependency relationships via the order of calling functions on an opaque type. It's certainly better to lift these important details of your implementation into the implementation itself, and make them explicit.

With all of this typed out, I'm open to hearing the counterpoints.

brian-brazil: > then that causal dependency should be expressed explicitly in the components themselves.

That'd be a layering violation, these components don't know that each other exist - nor should they. All the glue code is meant to live inside this file, and I'm trying to enforce that reloads don't happen after shutdown has commenced.

peterbourgon: > That'd be a layering violation, these components don't know that each other exist - nor should they. All the glue code is meant to live inside this file, and I'm trying to enforce that reloads don't happen after shutdown has commenced.

OK, that makes sense to me. I think the current state of things, while a compromise in the ways I've laid out above, is about as good as it gets. I can't imagine a goroutine lifecycle manager with more features that would be a net benefit in clarity, and I think what you have now is clearly superior to rolling it all by hand.

juliusv: IMO the layering violation is less bad because it's very explicit and visible to any new (or old) developer. We already do that for other concerns in this group of actors as well (like `reloadReady` and `dbOpen`), so it's an established pattern.

brian-brazil: reloadReady and dbOpen exist entirely in this file, not in the object they're managing.

krasi-georgiev: I looked at the problem only briefly but I suspect that the main problem comes from how we handle the config management and reloading.  These kind of issues happened few times already and although the components are very well separated still depend on a global config state.
Last time i looked had the impression that it could be solved by introducing another config  layer/manager

juliusv: > reloadReady and dbOpen exist entirely in this file, not in the object they're managing.

I'm not sure I understand how this would mean introducing anything outside of this file? I was thinking of e.g. a channel that gets closed when the config reloader has exited so that others only shut down when this has occurred. But now actually if termination handlers are executed sequentially, this wouldn't help us because it'd either be superfluous or lead to a deadlock if the ordering of groups is wrong.

So that seems like a good argument why the external channel-based coordination only works for startup, not shutdown, because actors are started in parallel, but shut down sequentially.

If we go for the group ordering solution here, it would be great to also have a big fat comment at the top of each actor that explains ordering requirements though.

fabxc: Question: we want TSDB to be started first and stopped after everything else terminated. We also want the first config load to complete successfully. Can we just pull those out of `run.Group` in regular `main` and only use the group for things that are actually independent?

brian-brazil: I think we're getting a bit sidetracked here. I propose we get this in now as an immediate fix so we can move towards ending the moratorium, and changes beyond that can be done later.

juliusv: @brian-brazil Ok.

brian-brazil: Is there anyone who objects to merging this as-is for now?

fabxc:
:+1:

juliusv:
:+1: for now

PR #4323, opened 2018-06-28
vSphere Service Discovery

dvonthenen:
This implements service discovery for objects in a vSphere environment. Currently, supported objects that are surfaced through service discovery are hosts (ESX), storage, and virtual machines. Native service discovery for vSphere objects in Prometheus enables multiple use-cases including:

1) similar to the OpenStack service discovery, allows obtaining metrics for on-premise environments.
2) it enables metrics collection for those users running [vSphere in the Amazon cloud](https://aws.amazon.com/vmware/)
3) since we can discover vSphere related objects, we can start to tie application metrics to the infrastructure that runs underneath it. Example, resource contention because two VMs running kubernetes are under stress since they share the same physical host. The VM can be migrated to a different host via alerting rules.
4) the vSphere service discovery can easily scale to handle multiple isolated vSphere environments. Adding new environments is as easy as defining a new scape job.

The Prometheus documentation was updated to include configuration information for using the service discovery as well as providing an example YAML.


simonpasquier:
@dvonthenen thanks for suggesting this change. Unfortunately, we are currently not accepting new integrations.

That being said, you will find a [post](https://prometheus.io/blog/2018/07/05/implementing-custom-sd/) on the project's blog on how you can write your custom SD mechanism and integrate it with the file SD.

dvonthenen:
@simonpasquier well that's unfortunate. Thanks for taking the time to look at and update with a response.

simonpasquier:
@dvonthenen thanks for acknowledging. If you need any help to move forward with the custom SD integation, feel free to ping me. Since you've already done most of the work, it shouldn't be complicated to refactor it as a standalone project.

PR #4328, opened 2018-07-01
rules: Minor naming/comment cleanups

juliusv:
Mainly the `evaluationTime` name confused me while reading the code because it sounds like a timestamp.


brian-brazil:
:+1:

Issue #4327, opened 2018-06-29
Kubernetes ingress discovery sets invalid scheme


nailgun:
## Bug Report
**What did you do?**
I have a single ingress object with multiple backends. One is https, second is http.

```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    kubernetes.io/ingress.class: nginx
    prometheus.io/path: healthz
  name: web-v1
  namespace: myns
spec:
  rules:
  - host: api.smp.io
    http:
      paths:
      - backend:
          serviceName: web-v1
          servicePort: 80
        path: /apps/v1/
  - host: smp.ingress.svc.cluster.local
    http:
      paths:
      - backend:
          serviceName: web-v1
          servicePort: 80
        path: /apps/v1/
  tls:
  - hosts:
    - api.smp.io
    secretName: api.smp.io-tls
```

**What did you expect to see?**
api.smp.io - httpS
smp.ingress.svc.cluster.local - http

Ingress controller (in my case - nginx) works this way.

**What did you see instead? Under which circumstances?**
Prometheus sets `__meta_kubernetes_ingress_scheme` to https for both backends.

**Environment**

* System information: GKE cluster
* Prometheus version: v2.3.1
* Prometheus configuration file:
```yaml
    # Scrape config for probing ingresses via the Blackbox Exporter
    - job_name: ingress
      kubernetes_sd_configs:
      - role: ingress

      metrics_path: /probe
      params:
        module: [http_2xx]

      scrape_timeout: 30s

      relabel_configs:
      - source_labels: [__meta_kubernetes_ingress_annotation_prometheus_io_probe]
        action: drop
        regex: false
      - source_labels: [__meta_kubernetes_ingress_path]
        action: drop
        regex: /.well-known/acme-challenge
      - source_labels: [__meta_kubernetes_ingress_scheme, __address__, __meta_kubernetes_ingress_path, __meta_kubernetes_ingress_annotation_prometheus_io_path]
        regex: (.+);(.+);(.+);(.*)
        replacement: ${1}://${2}${3}${4}
        target_label: __param_target
      - source_labels: [__meta_kubernetes_ingress_annotation_prometheus_io_blackbox_module]
        action: replace
        target_label: __param_module
      - target_label: __address__
        replacement: blackbox-exporter
      - source_labels: [__param_target]
        target_label: instance
      - action: labelmap
        regex: __meta_kubernetes_ingress_label_(.+)
[...]

PR #4329, opened 2018-07-01
discovery/kubernetes/ingress: fix TLS discovery

nailgun:
Add **failing** test for #4327

### DIFF
--- a/discovery/kubernetes/ingress.go
+++ b/discovery/kubernetes/ingress.go
@@ -176,12 +176,23 @@ func (s *Ingress) buildIngress(ingress *v1beta1.Ingress) *targetgroup.Group {
 	}
 	tg.Labels = ingressLabels(ingress)
 
-	schema := "http"
+	tlsHosts := make(map[string]struct{})
 	if ingress.Spec.TLS != nil {
-		schema = "https"
+		for _, tls := range ingress.Spec.TLS {
+			for _, host := range tls.Hosts {
+				tlsHosts[host] = struct{}{}
+			}
+		}
 	}
+
 	for _, rule := range ingress.Spec.Rules {
 		paths := pathsFromIngressRule(&rule.IngressRuleValue)
+
+		schema := "http"
+		_, isTLS := tlsHosts[rule.Host]
+		if isTLS {
+			schema = "https"
+		}
 
 		for _, path := range paths {
 			tg.Targets = append(tg.Targets, model.LabelSet{
--- a/discovery/kubernetes/ingress.go
+++ b/discovery/kubernetes/ingress.go
@@ -177,11 +177,9 @@ func (s *Ingress) buildIngress(ingress *v1beta1.Ingress) *targetgroup.Group {
 	tg.Labels = ingressLabels(ingress)
 
 	tlsHosts := make(map[string]struct{})
-	if ingress.Spec.TLS != nil {
-		for _, tls := range ingress.Spec.TLS {
-			for _, host := range tls.Hosts {
-				tlsHosts[host] = struct{}{}
-			}
+	for _, tls := range ingress.Spec.TLS {
+		for _, host := range tls.Hosts {
+			tlsHosts[host] = struct{}{}
 		}
 	}
 
--- a/discovery/kubernetes/ingress_test.go
+++ b/discovery/kubernetes/ingress_test.go
@@ -23,8 +23,16 @@ import (
 	"k8s.io/client-go/pkg/apis/extensions/v1beta1"
 )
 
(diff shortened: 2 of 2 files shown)


nailgun:
Should I fix DCO check failure?

brancz:
This lgtm, but yes we will need the DCO to be signed.

nailgun:
I have rebuilded branch with signed commits. But Travis fails on files I didn't changed.

krasi-georgiev:
that is strange. I re triggered and it still fails.
Do these fail locally for you?
If they fail locally could you run github bisect and find out which commit broke this test.

nailgun:
My PR is based on latest commit in master. And Travis test has not been run for that commit.
image
I think this is a new check that was added recently and it fails from that time. All latest PRs fail this check.
I think someone with write access should run `gofmt` or something in master and this will be fixed.

nailgun:
But hey, Travis build succeeds for go 1.10.x and fails for 1.x in style check. It looks like there is different versions of `gofmt` in Travis (maybe `gofmt` was changed in 1.10.x).

brancz:
If that's the case then there is no way we can get green tests for both jobs ...

krasi-georgiev:
@nailgun looks like you are right -  it is Travis related as all recent PR fail with the same error.

@SuperQ any idea for a fix?

krasi-georgiev:
I just upgraded to go version go1.10.3 linux/amd64 and all tests pass locally.

brancz:
It's not the tests that are failing, it's the formatting step and the version that is failing is `go version go1.11beta1 linux/amd64`. Try installing that version and just run `make`.

krasi-georgiev:
yeah I just noticed that and trying to find out what changed in 1.11 or whether this is a bug.

simonpasquier:
This seems to be introduced by this change to gofmt in go1.11: 
https://github.com/golang/go/commit/542ea5ad91367d2b40589942cc5757a4d5f43f97

nailgun:
Maybe enable style check only for supported Go version which is used to build official release?

krasi-georgiev:
looks like a bug and here is the issue report
https://github.com/golang/go/issues/26180

simonpasquier:
Reading at the gimme [documentation](https://github.com/travis-ci/gimme#asking-gimme-what-a-version-is) which resolves which go version to use, this is a bug on the TravisCI side:

> Use of .x will not find release candidates, alphas, betas or other non-release versions: it's only for finding the last stable release.

krasi-georgiev:
I think @nailgun is right we should be running the style check only against the latest stable golang release.

simonpasquier:
@krasi-georgiev using `stable` instead of `1.x` returns the correct version (see #4331). Back to your question, I don't see a point of running against releases which end up being the same.

simonpasquier:
I've also filed a bug against gimme since either the doc or the code is wrong: https://github.com/travis-ci/gimme/issues/152

nailgun:
Can I re-run Travis build without pushing any commits?

SuperQ:
@nailgun If you rebase your change against master, it will fix the Travis issue.

nailgun:
@SuperQ Thanks for the hint!

nailgun:
Unfortunate PR... Now SIGSEGV during testing of `prometheus/rules`. My changes should not cause this. Locally tests succeed. I have `go version go1.10.3 darwin/amd64`.

nailgun:
Hurray!

discovery/kubernetes/ingress.go
```diff
@@ -176,13 +176,24 @@ func (s *Ingress) buildIngress(ingress *v1beta1.Ingress) *targetgroup.Group {
 	}
 	tg.Labels = ingressLabels(ingress)
 
-	schema := "http"
+	tlsHosts := make(map[string]struct{})
 	if ingress.Spec.TLS != nil {
```
simonpasquier: very minor thing but this test can be removed since the `range ingress.Spec.TLS` will yield nothing when `ingress.Spec.TLS` is nil.

nailgun:
You are right, thank you. I have removed this check.

simonpasquier:
@nailgun thanks! The CircleCI error is transient (they say they have issues with GCE).

nailgun:
Is there a chance for it to be included in the next release?

brancz:
This lgtm.

Yes as this is a bug fix, this should land in the 2.3.2 patch release.

Maybe @krasi-georgiev or @simonpasquier can give this a review as well and then we can merge this if they give their lgtm.

simonpasquier:
Yes, I'm planning to test it today.

simonpasquier:
:+1: for me

Issue #4322, opened 2018-06-28
Add OCI Service Discovery


sw-samuraj:
# Proposal
**Add OCI Service Discovery support**

[Oracle Service Infrastructure](https://cloud.oracle.com/cloud-infrastructure) (OCI) is a cloud IaaS, similar to _EC2_, _GCE_, or _OpenStack_ which already have support for _Prometheus_ service discovery. It would be suitable to have _OCI_ service discovery as well.

I've already forked _Prometheus_ and implemented the first version of _OCI_ service discovery (branch [oci-service-discovery](https://github.com/sw-samuraj/prometheus/tree/oci-service-discovery)), therefore after some polishing, I could provide a pull request.

We are currently building a solution build on _OCI_ infrastructure where _Prometheus_ monitoring is an important component - as such, we are developing and testing _OCI_ service discovery in a real _OCI_ environment (I'm mentioning this because it's probably not so straightforward to have _OCI_ access).

krasi-georgiev:
Thanks for the suggestion. Unfortunately, we are currently not accepting new integrations.

We can only provide the stability and performance we want to provide if we can properly maintain the codebase. This includes, test integrations in an automated and scalable fashion. For this reason, we are suggesting people to integrate via [file SD](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#%3Cfile_sd_config%3E) and [here](https://github.com/prometheus/prometheus/tree/master/documentation/examples/custom-sd) is a very good example to show this in practise. We would be more than happy to list this on our [integrations page](https://prometheus.io/docs/operating/integrations/).

Even if existing integrations can not be tested in an automated fashion, we will not remove them for reasons of compatibility. This also means that any additions we take on, or any changes to existing integrations we make or accept, will mean maintaining and testing these until at least the next major version, realistically even beyond that.

Feel free to question this answer on our [developer mailing list](https://groups.google.com/forum/#!forum/prometheus-developers), but be aware it's unlikely that you will get a different answer.

sw-samuraj:
I've posted a question to the developer mailing list: [Is it a service discovery via file_sd efficient enough to handle hundreds of instances](https://groups.google.com/forum/#!topic/prometheus-developers/EbLFqoXIjJg).

PR #4330, opened 2018-07-02
OCI service discovery

sw-samuraj:
Implements the proposal #4322.

This PR adds a new service discovery for OCI (Oracle Cloud Infractructure).


krasi-georgiev:
as per my comment in the original proposal I would have to close this as everyone on the team agrees that we can't take any more integrations. sorry for not replying earlier before you had spent the time for the PR.

PR #4331, opened 2018-07-02
travis: remove testing with go 1.x

### LINKED ISSUE #4332: Failing gofmt tests
@brancz , @SuperQ do you remember why we run `make test` against .1.10 and 1.x? 

It seems that we always update the second target (currently 1.10.x) to latest stable anyway so we are running against the same versions on both targets most of the time.

There are some gofmt changes in 1.13 and Travis has some bug in travis that returns 1.13 for 1.x where it should return latest stable, but my question is should we just run 1.10 or at least run `make style` only for the laster golang stable.?

simonpasquier:
https://github.com/travis-ci/gimme/issues/152 is the upstream issue (`gimme` was returning a beta version for `1.x`) but in any case, it makes sense to run TravisCI with the same go version as what is used by CircleCI to build the artifacts (eg the latest `1.10.x`).

Closes #4332.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,8 @@ version: 2
 
 jobs:
   test:
+    # Whenever the Go version is updated here, .travis.yml should also be
+    # updated.
     docker:
     - image: circleci/golang:1.10
     working_directory: /go/src/github.com/prometheus/prometheus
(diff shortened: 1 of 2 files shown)


krasi-georgiev:
LGTM , can merge on green

nailgun:
It's green :)

PR #4333, opened 2018-07-02
EC2 DS: Allow to set a custom endpoint

jannickfahlbusch:
Allowing to set a custom endpoint makes it easy to monitor targets on non AWS providers with EC2 compliant APIs.

### DIFF
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -415,6 +415,9 @@ See below for the configuration options for EC2 discovery:
 # The AWS Region.
 region: <string>
 
+# Custom endpoint to be used.
+[ endpoint: <string> ]
+
 # The AWS API keys. If blank, the environment variables `AWS_ACCESS_KEY_ID`
 # and `AWS_SECRET_ACCESS_KEY` are used.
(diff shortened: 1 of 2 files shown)


jannickfahlbusch:
Hi @brian-brazil or @fabxc could you take a look at this?

jannickfahlbusch:
Hi @brian-brazil or @fabxc could you take a look at this?

simonpasquier:
LGTM

docs/configuration/configuration.md
```diff
@@ -415,6 +415,9 @@ See below for the configuration options for EC2 discovery:
 # The AWS Region.
 region: <string>
 
+# Custom endpoint to be used.
```
brian-brazil: What happens if both endpoint and region is specified?

jannickfahlbusch: A custom endpoint has also regions. 

The region needs to be specified regardless of the endpoint to be used. In case of the AWS endpoints you have a set of predefined regions while with a custom endpoint you have the ability to specify any string

brian-brazil:
Thanks!

PR #4347, opened 2018-07-04
Build prometheus inside docker

sylr:
Use a docker golang image to build prometheus and copy the output to make the prometheus docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,22 @@
+FROM golang:1.10-alpine3.7 AS builder
+
+ADD . /go/src/github.com/prometheus/prometheus
+WORKDIR /go/src/github.com/prometheus/prometheus
+
+RUN apk update && apk upgrade && apk add --no-cache git alpine-sdk
+
+RUN make build
+
+# -----------------------------------------------------------------------------
+
 FROM        quay.io/prometheus/busybox:latest
 LABEL maintainer "The Prometheus Authors <prometheus-developers@googlegroups.com>"
 
-COPY prometheus                             /bin/prometheus
-COPY promtool                               /bin/promtool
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. Our build system is over in https://github.com/prometheus/promu which already does this for us, but if there's improvements you think that could be made you can propose them there.

sylr:
@brian-brazil the purpose of this PR is to be able to use 3rd party services like docker hub or quay.io and build automatically our development branches like here : https://hub.docker.com/r/sylr/prometheus/builds/bzeilxxtn29jxqg8k7bkawr/

brian-brazil:
We already have our own infrastructure for that for our public builds, you're free to develop your own Dockerfiles for internal use though.

sylr:
I'm sure you have your own infrastructure but this Dockerfile makes it effortless for anyone (which might not have access to said infrastructure) to build the prometheus docker image.

It's a general improvement over the current one which requires you to build on your machine (which might have not all the build dependencies) before generating the image.

This makes no sense to refuse it.

brian-brazil:
You change could interfere with our build infrastructure. While I understand that this is useful for you, it does not belong upstream. If you wish to propose we rework our build system to work in this fashion, you are free to propose it in the promu repo.

sdurrheimer:
One of the goal of our current build system is to deliver the exact same binary as tarballs and through our docker images.
We also crossbuild the project for several platforms, and one of them goes to the docker image.
We also had issues (DNS resolution etc) when building our projects with alpine, thats why today we crossbuild with debian's based docker images with proper libc.
Keep in mind we have the same release process for all our golang's based projects.

It might be okay to have second `Dockerfile.multi-stage`, but we would then need to maintain it and troubleshoot issues coming from it.

Issue #1695, opened 2016-05-31
General purpose rule/alert testing tool


tamsky:
https://groups.google.com/d/msg/prometheus-developers/h6Vp2pf619A/AJsKJsJTHQAJ
DL;DR (very edited) summary:

> Q on (8/7/15) by @swsnider :  [https://github.com/prometheus/prometheus/promql/test.go](https://github.com/prometheus/prometheus/blob/master/promql/test.go) -- Is it appropriate to write some wrapper code to turn this into a general-purpose rule/alert testing tool?
> 
> A by @fabxc :
> It is absolutely on our roadmap (should probably be added to the actual one on the website).
> What's blocking it is not so much implementation but specification.

I would like to help get the specification ball rolling here.

@fabxc Has there been any work done to spec this out since that post?

fabxc:
Yes, getting the discussion started would help a lot and make it easier for people willing to contribute to start working on something. So far this area has not progressed.

For these kinds of things an RFC in a shared Google doc generally works best.

tamsky:
Let me get the ball rolling with a few ideas here, a google doc can follow.

> (in the same [groups thread](https://groups.google.com/d/msg/prometheus-developers/h6Vp2pf619A/AJsKJsJTHQAJ)) @fabxc wrote:
> the syntax for evaluation is something that strongly bothers me. Specifying time (ranges) is fine for the current testing but not suitable for a more general purpose language.

Perhaps you can put into a few more words what bothers you regarding the current time (range) stuff, and alternatives that might help?

The current test syntax (to me) appears to be very simple, which is good:
    `load <step:duration>)`
    `eval[_fail|_ordered] instant [at <offset:duration>] <query>`

Concepts I think are missing:
- ability to `include ../*.rules`
- `NOW <date:dateTime>|<[-+]offset:duration>`
- `evaluation_interval`  (allows rule testing tool to interact with the global config param)
- `eval_vector`  (accepts an instant-vector, and drops all labels)
- `eval range`
- a way to express missed scrapes in `load` which should affect metric staleness

`NOW` would provide a way to set the date+time for all actions that follow, simplifying `eval` commands.  Example ::
```
clear
evaluation_interval 5m
include "../rules/*.rules"

NOW 2016-06-02T00:00:00
load 5m
    http_errors{result="404"} 0+10x12 120x12   # steady error rate for 1h, zero additional for 1h

    # eval without an 'at' clause implies clock == NOW command above
[...]

PR #4350, opened 2018-07-04
Unit testing for rules

codesome:
Issue: https://github.com/prometheus/prometheus/issues/1695

Design Doc: [Prometheus - Unit testing for Rules](https://docs.google.com/document/d/1vH5nxFcdVCXlTuUfXwPoRghiE79Ce9_eq0Je3a5g2Vs/edit?usp=sharing)

Usage: `./promtool test rules test1.yml test2.yml test3.yml ...`

Example for valid test file: [Prometheus - Unit Testing for Rules - Example test files](https://gist.github.com/codesome/f4426d28244ee9fac128031ed2f9af70)


cmd/promtool/unittest.go
```diff
@@ -0,0 +1,507 @@
...
+	for _, art := range tg.AlertRuleTest {
+		pos := func() int {
+			for i, d := range alertEvalTimes {
+				if d > art.EvalTime {
+					return i
+				}
+			}
+			return len(alertEvalTimes)
+		}()
+		alertEvalTimes = append(alertEvalTimes[:pos], append([]time.Duration{art.EvalTime}, alertEvalTimes[pos:]...)...)
```
gouthamve: Can we not sort the `alertEvalTimes` via: `sort.Slice()` after getting an unsorted slice? That looks more readable.

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,507 @@
...
+		} else {
+			alertTests[art.EvalTime] = append(alertTests[art.EvalTime], art)
+		}
+	}
+
+	// Current index in alertEvalTimes what we are looking at.
+	curr := 0
+
+	var errs []error
+	for ts := mint; maxt.Sub(ts) > 0; ts = ts.Add(evalInterval) {
```
gouthamve: `ts.Before(maxt)` is better

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,507 @@
...
+			g.Eval(suite.Context(), ts)
+		}
+
+		for curr < len(alertEvalTimes) && ts.Sub(mint) <= alertEvalTimes[curr] && alertEvalTimes[curr] < ts.Add(evalInterval).Sub(mint) {
+			// We need to check alerts for this time.
+			// If 'ts <= `eval_time=alertEvalTimes[curr]` < ts+evalInterval'
+			// then we compare alerts with the Eval at `ts`.
+			t := alertEvalTimes[curr]
+
+			presentNames := presentInTest[t]
```
gouthamve: `presentAlerts` better name?

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,507 @@
...
+				}
+			}
+			return len(alertEvalTimes)
+		}()
+		alertEvalTimes = append(alertEvalTimes[:pos], append([]time.Duration{art.EvalTime}, alertEvalTimes[pos:]...)...)
+
+		if _, ok := presentInTest[art.EvalTime]; !ok {
+			presentInTest[art.EvalTime] = make(map[string]struct{})
+		}
+		presentInTest[art.EvalTime][art.Alertname] = struct{}{}
```
gouthamve: `alertsInTests` better?

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,507 @@
...
+			t := alertEvalTimes[curr]
+
+			presentNames := presentInTest[t]
+			got := make(map[string]labelsAndAnnotations)
+
+			// Same Alert name can be present in multiple groups.
+			// Hence we collect them all to check against expected alerts.
+			for _, g := range groups {
+				grules := g.Rules()
+
```
gouthamve: Extra whitespace is not needed.

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,507 @@
...
+					for _, a := range ar.ActiveAlerts() {
+						if a.State == rules.StateFiring {
+							alerts = append(alerts, labelAndAnnotation{
+								Labels:      append(labels.Labels{}, a.Labels...),
+								Annotations: append(labels.Labels{}, a.Annotations...),
+							})
+						}
+					}
+
+					if _, ok := got[ar.Name()]; ok {
```
gouthamve: You could simply do: `got[ar.Name()] = append(got[ar.Name()], alerts...)`

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,507 @@
...
+			return len(alertEvalTimes)
+		}()
+		alertEvalTimes = append(alertEvalTimes[:pos], append([]time.Duration{art.EvalTime}, alertEvalTimes[pos:]...)...)
+
+		if _, ok := presentInTest[art.EvalTime]; !ok {
+			presentInTest[art.EvalTime] = make(map[string]struct{})
+		}
+		presentInTest[art.EvalTime][art.Alertname] = struct{}{}
+
+		if _, ok := alertTests[art.EvalTime]; ok {
```
gouthamve: Could just be: `alertTests[art.EvalTime] = append(alertTests[art.EvalTime], art)`

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,507 @@
...
+						got[ar.Name()] = append(got[ar.Name()], alerts...)
+					} else {
+						got[ar.Name()] = alerts
+					}
+				}
+			}
+
+			for _, testCase := range alertTests[t] {
+				// Checking alerts.
+				gotAlerts, ok := got[testCase.Alertname]
```
gouthamve: Just `gotAlerts := got[testCase.Alertname]` should be fine.

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,507 @@
...
+
+					if _, ok := got[ar.Name()]; ok {
+						got[ar.Name()] = append(got[ar.Name()], alerts...)
+					} else {
+						got[ar.Name()] = alerts
+					}
+				}
+			}
+
+			for _, testCase := range alertTests[t] {
```
gouthamve: `testcase` should be fine. No need to camelcase that :)

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,507 @@
...
+		return nil, fmt.Errorf("rule result is not a vector or scalar")
+	}
+}
+
+type labelsAndAnnotations []labelAndAnnotation
+
+func (la labelsAndAnnotations) Len() int      { return len(la) }
+func (la labelsAndAnnotations) Swap(i, j int) { la[i], la[j] = la[j], la[i] }
+func (la labelsAndAnnotations) Less(i, j int) bool {
+	l1h, l2h := la[i].Labels.Hash(), la[j].Labels.Hash()
```
gouthamve: `labels.Compare` is better here.

codesome:
@gouthamve All the reviews fixed. Learn few stuff about Golang :)

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,489 @@
...
+		s = s[:len(s)-2]
+	}
+	return s
+}
+
+// orderedGroups returns a slice of *rules.Group from `groupsMap` which follows the order
+// mentioned by `groupOrderMap`. The returned slice starts with the groups that need to be in order,
+// followed by rest of the groups in undetermined order.
+func orderedGroups(groupsMap map[string]*rules.Group, groupOrderMap map[string]int) []*rules.Group {
+	var groups []*rules.Group
```
gouthamve: Okay, this is too complex. Can we do this instead:

```go
groups := make([]*rules.Group, 0, len(groupsMaps))
for _, g := range groupsMap {
  groups = append(groups, g)
}

sort.Slice(groups, func(i, j int) bool {
  return groupOrderMap[groups[i].Name()] < groupOrderMap[groups[j].Name()]
})

return groups
```

codesome: Didn't think of that, thanks! I have updated.

codesome:
@brian-brazil PTAL when you get free. Also, where would the docs for this go in?

brian-brazil:
I won't get to look until next week. A new sibling page to Recording/Alerting rules is probably the right place.

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,467 @@
...
+	}
+	m := rules.NewManager(opts)
+	groupsMap, ers := m.LoadGroups(tg.Interval, ruleFiles...)
+	if ers != nil {
+		return ers
+	}
+	groups := orderedGroups(groupsMap, groupOrderMap)
+
+	// Pre-processing some data for testing alerts.
+	// All this preparating so that we can test the alerts as and when we evaluate
```
brian-brazil: All this preparation is so that we can test alerts as we evaluate the rules. This avoids storing them in memory, as the number of evals might be high.

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,467 @@
...
+	sort.Slice(alertEvalTimes, func(i, j int) bool {
+		return alertEvalTimes[i] < alertEvalTimes[j]
+	})
+
+	// Current index in alertEvalTimes what we are looking at.
+	curr := 0
+
+	var errs []error
+	for ts := mint; ts.Before(maxt); ts = ts.Add(evalInterval) {
+		// Collects the alerts asked in unit testing.
```
brian-brazil: asked for

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,467 @@
...
+		for _, g := range groups {
+			g.Eval(suite.Context(), ts)
+		}
+
+		for curr < len(alertEvalTimes) && ts.Sub(mint) <= alertEvalTimes[curr] && alertEvalTimes[curr] < ts.Add(evalInterval).Sub(mint) {
```
brian-brazil: This is a bit long of a for expression, I'd suggest moving it into an if inside the for

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,467 @@
...
+		}
+	}
+
+	// Checking promql expressions.
+	for _, testCase := range tg.PromqlExprTests {
+		got, err := query(suite.Context(), testCase.Expr, mint.Add(testCase.EvalTime),
+			suite.QueryEngine(), suite.Queryable())
+		if err != nil {
+			errs = append(errs, fmt.Errorf("    expr:'%s', time:%s, err:%s", testCase.Expr,
+				testCase.EvalTime.String(), err.Error()))
```
brian-brazil: Shouldn't we continue here?

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,467 @@
...
+		}
+
+		var expSamples []parsedSample
+		parseError := false
+		for _, s := range testCase.ExpSamples {
+			lb, err := promql.ParseMetric(s.Labels)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("    expr:'%s', time:%s, err:%s", testCase.Expr,
+					testCase.EvalTime.String(), err.Error()))
+				parseError = true
```
brian-brazil: Why not break out here, rather than adding a new bool?

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,467 @@
...
+func (tg *testGroup) seriesLoadingString() string {
+	result := ""
+	result += "load " + shortDur(tg.Interval) + "\n"
+	for _, is := range tg.InputSeries {
+		result += "  " + is.Series + " " + is.Values + "\n"
+	}
+	return result
+}
+
+// https://stackoverflow.com/a/41336257/6219247
```
brian-brazil: Do you have rights to use this code as apache2?

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,467 @@
...
+
+	if unitTestInp.EvaluationInterval == 0 {
+		unitTestInp.EvaluationInterval = 1 * time.Minute
+	}
+
+	// Bounds for evaluating the rules.
+	mint := time.Unix(0, 0)
+	maxd := unitTestInp.maxEvalTime()
+	maxt := mint.Add(maxd)
+	// Rounding off to nearest Eval time (> maxt).
```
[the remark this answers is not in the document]
codesome: https://github.com/codesome/prometheus/blob/5576d8e6312d07431d2db5e8d76ac78908510968/cmd/promtool/unittest.go#L192

If we don't add some extra time, the above loop can miss the last eval if `maxt` happens to be exactly the last eval. I think adding one `evalInterval` would make more sense than rounding off.

cmd/promtool/unittest.go
codesome: So that I can skip checking for `got ==  exp` [here](https://github.com/codesome/prometheus/blob/5576d8e6312d07431d2db5e8d76ac78908510968/cmd/promtool/unittest.go#L301) in case it was an error. So I am using a bool to keep track of that.

cmd/promtool/unittest.go
codesome: I am unaware of licenses, but what I see is, if I remove that comment, the code is general enough that it can be used. Maybe change it a bit you say?

cmd/promtool/unittest.go
brian-brazil: If you break out of the outer for, you don't need that

codesome: I have made the changes to continue to the outer for.

cmd/promtool/unittest.go
```diff
@@ -0,0 +1,476 @@
...
+	"gopkg.in/yaml.v2"
+
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/promql"
+	"github.com/prometheus/prometheus/rules"
+	"github.com/prometheus/prometheus/storage"
+)
+
+// RulesUnitTest does unit testing of rules based on the unit testing files provided.
+// An example file as input can be found here: https://gist.github.com/codesome/f4426d28244ee9fac128031ed2f9af70
```
brian-brazil: We should add proper docs for this

codesome: Yes. I will add docs for this. So I will remove that comment from here as docs will be added for this.

codesome: Oh, I was trying to add docs in https://github.com/prometheus/docs. Just found that there is a `docs/` directory in this repo. I will add them.

codesome: I have added the docs.

docs/configuration/unit_testing_rules.md
```diff
@@ -0,0 +1,213 @@
...
+
+### `<series>`
+
+```yaml
+# This follows the usual series notation (x{a="b", c="d"}).
+series: <string>
+
+# This uses expanding notation.
+# Expanding notation: 
+#     a+bxc ~ a a+b a+(2*b) a+(3*b) … a+(c*b)
```
brian-brazil: The meaning of the ~ here is not clear, I'd suggest using becomes instead

docs/configuration/unit_testing_rules.md
```diff
@@ -0,0 +1,213 @@
...
+# List of expected alerts which are firing under the given alertname at 
+# given evaluation time. If you want to test if an alerting rule should 
+# not be firing, then you can mention the above fields and leave 'exp_alerts' empty.
+exp_alerts:
+  [ - <alert> ]
+```
+
+### `<alert>`
+
+Remember, this alert shoud be firing.
```
brian-brazil: This comes across weirdly, I'd remove it.

docs/configuration/unit_testing_rules.md
```diff
@@ -0,0 +1,213 @@
...
+# Expected samples at the given evaluation time.
+exp_samples:
+  [ - <sample> ]
+```
+
+### `<sample>`
+
+```yaml
+# Labels of the sample in series notation.
+labels: <series_notation>
```
brian-brazil: What's series notation?

codesome: I updated it with examples. I meant to say `<metric name>{<label name>=<label value>, ...}`.

docs/configuration/unit_testing_rules.md
```diff
@@ -0,0 +1,217 @@
...
+
+# Expected samples at the given evaluation time.
+exp_samples:
+  [ - <sample> ]
+```
+
+### `<sample>`
+
+```yaml
+# Labels of the sample in ususal series notation '<metric name>{<label name>=<label value>, ...}'
```
brian-brazil: usual

codesome: Done.

brian-brazil:
Thanks!

PR #4359, opened 2018-07-06
Log errors encountered when marshalling and writing responses.

tomwilkie:
This was masking errors in Cortex - I would get an empty response with headers or body.  And took me an embarrassingly long time to find...

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -824,9 +824,6 @@ func mergeLabels(primary, secondary []*prompb.Label) []*prompb.Label {
 }
 
 func (api *API) respond(w http.ResponseWriter, data interface{}) {
-	w.Header().Set("Content-Type", "application/json")
-	w.WriteHeader(http.StatusOK)
-
 	json := jsoniter.ConfigCompatibleWithStandardLibrary
 	b, err := json.Marshal(&response{
 		Status: statusSuccess,
@@ -834,8 +831,12 @@ func (api *API) respond(w http.ResponseWriter, data interface{}) {
 	})
 	if err != nil {
 		level.Error(api.logger).Log("msg", "error marshalling json response", "err", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
 	if n, err := w.Write(b); err != nil {
 		level.Error(api.logger).Log("msg", "error writing response", "n", n, "err", err)
 	}
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -28,6 +28,8 @@ import (
 	"time"
 	"unsafe"
 
+	"github.com/go-kit/kit/log"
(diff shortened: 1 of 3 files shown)


web/api/v1/api.go
```diff
@@ -829,12 +833,15 @@ func respond(w http.ResponseWriter, data interface{}) {
 		Data:   data,
 	})
 	if err != nil {
+		level.Error(api.logger).Log("msg", "error marshalling json response", "err", err)
 		return
 	}
-	w.Write(b)
+	if n, err := w.Write(b); err != nil {
+		level.Error(api.logger).Log("msg", "error writing response", "n", n, "err", err)
+	}
```
DanCech: Is there a reason for sending the header before the body has been successfully encoded?  It seems like it would make more sense to encode the data first, then send the Content-Type (and potentially Content-Length as it will be known) just before calling Write.  The explicit `w.WriteHeader(http.StatusOK)` is technically not needed as `w.Write` will do that implicitly.

If this is done, then an encoding error could be handled by returning an error to the client.

web/api/v1/api.go
```diff
@@ -864,7 +872,13 @@ func respondError(w http.ResponseWriter, apiErr *apiError, data interface{}) {
 	if err != nil {
 		return
 	}
-	w.Write(b)
+	if err != nil {
+		level.Error(api.logger).Log("msg", "error marshalling json response", "err", err)
+		return
```
DanCech: Should this be updated to match the new flow in `respond`?

web/api/v1/api.go
```diff
@@ -864,7 +872,13 @@ func respondError(w http.ResponseWriter, apiErr *apiError, data interface{}) {
 	if err != nil {
```
gouthamve: This needs to go? We're handling the same error twice.

web/api/v1/api.go
```diff
@@ -819,23 +823,38 @@ func mergeLabels(primary, secondary []*prompb.Label) []*prompb.Label {
...
 		return
 	}
-	w.Write(b)
-}
 
-func respondError(w http.ResponseWriter, apiErr *apiError, data interface{}) {
 	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	if n, err := w.Write(b); err != nil {
+		level.Error(api.logger).Log("msg", "error writing response", "n", n, "err", err)
```
gouthamve: `n` will look odd. Can we do `bytesWritten`?

web/api/v1/api.go
```diff
@@ -852,19 +871,12 @@ func respondError(w http.ResponseWriter, apiErr *apiError, data interface{}) {
...
-		ErrorType: apiErr.typ,
-		Error:     apiErr.err.Error(),
-		Data:      data,
-	})
-	if err != nil {
-		return
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(code)
+	if n, err := w.Write(b); err != nil {
+		level.Error(api.logger).Log("msg", "error writing response", "n", n, "err", err)
```
gouthamve: Here too!

PR #4360, opened 2018-07-07
docs: added undocumented step api parameter format

FUSAKLA:
@juliusv 
So it does't look I'm only complaining :)

Documented second allowed format for the `step` parameter in the HTTP API for `query_range` allowing to specify the `step` using number of seconds in as float.

### DIFF
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -132,7 +132,7 @@ URL query parameters:
 - `query=<string>`: Prometheus expression query string.
 - `start=<rfc3339 | unix_timestamp>`: Start timestamp.
 - `end=<rfc3339 | unix_timestamp>`: End timestamp.
-- `step=<duration>`: Query resolution step width.
+- `step=<duration | float>`: Query resolution step width in `duration` format or float number of seconds.
 - `timeout=<duration>`: Evaluation timeout. Optional. Defaults to and
    is capped by the value of the `-query.timeout` flag.
 


docs/querying/api.md
```diff
@@ -132,7 +132,7 @@ URL query parameters:
 - `query=<string>`: Prometheus expression query string.
 - `start=<rfc3339 | unix_timestamp>`: Start timestamp.
 - `end=<rfc3339 | unix_timestamp>`: End timestamp.
-- `step=<duration>`: Query resolution step width.
+- `step=<duration | float>`: Query resolution step width in `duration` format or float describing number of seconds..
```
brian-brazil: s/describing/

FUSAKLA: Thanks, should be fixed and amended. 
Also removed the double trailing dot.

brian-brazil:
Thanks!

docs/querying/api.md
```diff
@@ -132,7 +132,7 @@ URL query parameters:
 - `query=<string>`: Prometheus expression query string.
 - `start=<rfc3339 | unix_timestamp>`: Start timestamp.
 - `end=<rfc3339 | unix_timestamp>`: End timestamp.
-- `step=<duration>`: Query resolution step width.
+- `step=<duration | float>`: Query resolution step width in `duration` format or float number of seconds.
```
juliusv: Instead of changing the documentation for the `step` parameter, the documentation for `<duration>` should be changed. There is one other parameter which is a `<duration>` too (the timeout) and it is parsed the same way.

brian-brazil: Durations are used in other places, such as in range vectors, which do not allow floats.

juliusv: That's a different kind of `<duration>`. On https://prometheus.io/docs/prometheus/latest/querying/api/ it's defined separately and only used for `step` and `timeout`, so the definition there should be updated instead of the individual fields.

brian-brazil: We shouldn't have types with different definitions on different pages, that'll cause confusion for users.

juliusv: That's a larger issue. My point stands.

The larger issue is that Prometheus itself parses duration values differently depending on context (HTTP API arg vs. config field or PromQL range selector) , so we'll either need to change Prometheus, name these duration types differently, or just have separate definitions with the same name in different contexts. I think for now the latter is fine.

brian-brazil: I think what we currently have is fine.

As it stands we're not exhaustively defining every type on every relevant page, so there's more to it than just adding an extra definition in some places. And defining the same thing differently would still cause unnecessary confusion.

PR #4363, opened 2018-07-09
add unused point slices to the pool

### LINKED ISSUE #4362: Unused point slices are not added back to the pool by the matrix selector
A recent change to promql/engine.go moved the call to `putPointSlice` from the evaluator finalizers in the `matrixSelector` method to elsewhere:

https://github.com/prometheus/prometheus/commit/dd6781add2390eb2f2251e108f2ef835eba44b05#diff-83539908220a826e39b7ba3a4f27d5d6L862

However, now only the used point slices are returned to the pool:
https://github.com/prometheus/prometheus/blob/dd6781add2390eb2f2251e108f2ef835eba44b05/promql/engine.go#L732

I may be missing something here, but afaict unused (ie: empty) point slices should be returned to the pool immediately, since no reference to them is held after the current iteration in `matrixSelector` anyway.

roganartu:
Fixes #4362

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -1060,6 +1060,8 @@ func (ev *evaluator) matrixSelector(node *MatrixSelector) Matrix {
 
 		if len(ss.Points) > 0 {
 			matrix = append(matrix, ss)
+		} else {
+			putPointSlice(ss.Points)
 		}
 	}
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks!

PR #4370, opened 2018-07-11
Release 2.3.2

brian-brazil:


CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 2.3.1 / 2018-07-12
```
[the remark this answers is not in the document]
brian-brazil: Fixed, thanks

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 2.3.2 / 2018-07-12
+
+* [BUGFIX] Fix various tsdb bugs (#4369)
```
juliusv: Can we settle on one formatting for the linked PRs? It was without parens in the 2.3.1 Changelog. With parens is better, but maybe update the 2.3.1 one then too?

juliusv: Btw., these don't automatically turn into links, they have to be linked manually...

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
...
+* [BUGFIX] Reorder startup and shutdown to prevent panics. (#4321)
+* [BUGFIX] Exit with non-zero code on error (#4296)
+* [BUGFIX] discovery/kubernetes/ingress: fix scheme discovery (#4329)
+* [BUGFIX] Fix race in zookeeper sd (#4355)
+* [BUGFIX] Better timeout handling in promql (#4291 #4300)
+* [BUGFIX] Bubble up errors to promql from populating iterators (#4136)
```
juliusv: From a user's perspective it's not really clear what (user-visible) bug has been fixed here. Maybe something about how there was a silently ignored error resulting in partial data for a PromQL query?

CHANGELOG.md
brian-brazil: This is what was easiest to copy&paste, if the people pushing for this change want to offer tooling for it I'll consider using it. They are auto-linked, see https://github.com/prometheus/prometheus/releases/tag/v2.3.1

CHANGELOG.md
brian-brazil: I've tweaked it. This is a very subtle thing that I'd be surprised if anyone has ever hit, I don't think it needs much detail here.

grobie:
Great work, thanks! :+1:

CHANGELOG.md
juliusv: Ah, good to know it shows there. It doesn't show as links on https://github.com/prometheus/prometheus/blob/master/CHANGELOG.md

Kubernetes does real links, see https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.9.md#changelog-since-v198. But if that's too much work to do for now, we can also wait for tooling.

juliusv: I'd still at least keep the format consistent with 2.3.1 (with my pref being updating 2.3.1 to the parenthesized format).

Issue #3228, opened 2017-09-28
Several unbounded ioutil.ReadAll calls with HTTP request/responses


mdlayher:
Did a quick `grep` and filtered out vendored code and tests, and found a few instances of unbounded `ioutil.ReadAll` calls.  Since some of these accept HTTP request and response bodies, it may be a good idea to set some sane upper bound to avoid limitless memory consumption.

I suspect a few MiB of data might be okay in the worst case, but I'm not super familiar with e.g. the remote read/write traffic.

```
$ grep -r "ioutil.ReadAll" | grep -v vendor | grep -v "_test.go"
documentation/examples/remote_storage/remote_storage_adapter/main.go:           compressed, err := ioutil.ReadAll(r.Body)
documentation/examples/remote_storage/remote_storage_adapter/main.go:           compressed, err := ioutil.ReadAll(r.Body)
documentation/examples/remote_storage/remote_storage_adapter/opentsdb/client.go:        buf, err = ioutil.ReadAll(resp.Body)
documentation/examples/remote_storage/example_write_adapter/server.go:          compressed, err := ioutil.ReadAll(r.Body)
web/web.go:     text, err := ioutil.ReadAll(file)
discovery/triton/triton.go:     data, err := ioutil.ReadAll(resp.Body)
discovery/marathon/marathon.go: body, err := ioutil.ReadAll(resp.Body)
storage/remote/codec.go:        compressed, err := ioutil.ReadAll(r.Body)
storage/remote/client.go:       compressed, err = ioutil.ReadAll(httpResp.Body)
```

If this isn't a concern, feel free to close the issue.  Just wanted to make sure it was noted.

Gouthamve:
This is especially interesting for remote-read as it has the highest potential to nuke Prometheus.

mdlayher:
Indeed.  Ping @tomwilkie for thoughts since I noticed a couple of these in remote storage.

tomwilkie:
Yeah we should bound them with a `LimitedReader`.  Prometheus doesn't do a lot to protect itself, but thats not an excuse not to improve things!

I want to get 2.0 merged in before fixing this kind of bug though, as I don't want it to get lost in the merge.

mdlayher:
Makes sense for me. Just so it's tracked here, do you know what a sane upper bound could be?

tomwilkie:
I'll gather some stats from our Cortex but I don't image more than 10Mb.

mattbostock:
[...]

PR #4376, opened 2018-07-12
wrap remote triton and marathon reads in an io.LimitReader

adamdecaf:
Previously we allowed unbounded reads from these remote sources, which
can lead to performance and stability issues with prometheus.

Issue: https://github.com/prometheus/prometheus/issues/3228

### DIFF
--- a/discovery/marathon/marathon.go
+++ b/discovery/marathon/marathon.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"math/rand"
 	"net"
@@ -57,6 +58,9 @@ const (
 
 	// Constants for instrumentation.
 	namespace = "prometheus"
+
+	// Maximum bytes to read in remote response
+	maxRemoteReadBytes = 10 * 1024 * 1024
 )
 
 var (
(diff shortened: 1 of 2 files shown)


discovery/marathon/marathon.go
```diff
@@ -57,6 +58,9 @@ const (
 
 	// Constants for instrumentation.
 	namespace = "prometheus"
+
+	// Maximum bytes to read in remote response
+	maxRemoteReadBytes = 10 * 1024 * 1024
```
adamdecaf: I couldn't find a share spot to define these. I tried `discovery/config` and the root `config` package, but both cause import cycles. 

Do we want them configurable in yaml?

brian-brazil:
This should not be configurable. How did you determine that 10MB is enough for any possible user?

adamdecaf:
@brian-brazil I didn't, but that's what @tomwilkie suggested on the original issue. https://github.com/prometheus/prometheus/issues/3228#issuecomment-332913431 I don't know of any evidence to support that though. 25mb?

brian-brazil:
He was talking about a different use case. I lack the domain knowledge to know how big these can get. These are outbound requests in any case.

adamdecaf:
@brian-brazil The two calls I'm changing in this PR are for reading the marathon/triton response. Setting _some_ limit on those seems like a good idea.

I don't know enough to suggest a limit on the remote storage.

bwplotka:
Thanks for this @adamdecaf - however issue https://github.com/prometheus/prometheus/issues/3228 was for all discoveries and we don't know if it's still relevant. Closing this for now, we can always reopen. Thanks!

PR #4377, opened 2018-07-12
Fix missing 'msg' in remote storage adapter main.go .Log info message

silentpete:
Signed-off-by: Peter Gallerani <peter.gallerani@gmail.com>

@juliusv, lets see if this gets passed DCO


juliusv:
:+1: Thanks!

PR #4378, opened 2018-07-12
add prefix "common-" to make target names

pgier:
The node_exporter build is printing warnings due to overriding target names:
```
Makefile:74: warning: overriding recipe for target 'test'
Makefile.common:61: warning: ignoring old recipe for target 'test'
Makefile:105: warning: overriding recipe for target 'docker'
Makefile.common:89: warning: ignoring old recipe for target 'docker'
```

This change allows rules to be overridden without getting warnings about conflicting target names.  The make targets can still be called using their normal names such as "make build", however the actual target name in Makefile.common is "common-build".  And "common-build" will be skipped if there is a "build" target in Makefile. 

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -41,13 +41,18 @@ DOCKER_REPO             ?= prom
 .PHONY: all
 all: style staticcheck unused build test
 
-.PHONY: style
-style:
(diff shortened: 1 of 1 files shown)


pgier:
See also: https://github.com/prometheus/node_exporter/pull/999

PR #4382, opened 2018-07-13
the go test -race option is only supported on amd64

pgier:
Adds checks for current OS and arch to determine whether the '-race' option
should be used with go test.
These changes were mostly copied from the node_exporter Makefile.


krasi-georgiev:
LGTM ,Thanks @pgier

simonpasquier:
Paul, would you be able to resolve the conflicts?

pgier:
Changes have been rebased.

Issue #4383, opened 2018-07-13
Proposal: Add ability to (approximately) limit memory use by queries


roganartu:
## Proposal
Add an ability to set an (approximate) per-query limit on the amount of memory a query can consume and kill queries exceeding it, returning an error.

## Problem
Prometheus queries are currently able to consume as much memory as the system has available. It is possible for uneducated (or experimenting) users to issue a pathological query that takes down an entire node. Since there is no separation of querying and scraping components, a query OOMing prom results in scrape downtime. With a large WAL this may be in the order of 5-10 mins or longer as the service restarts and the WAL is read.

This is operationally difficult to support when designing for high availability while simultaneously encouraging new users to explore available metrics and create dashboards for their own use. Even if you run multiple prometheus nodes scraping the same targets with a load balancer in front, a pathological query being refreshed by the user can easily take down multiple nodes. The likelihood of this is perhaps increased by the fact that pathological queries are slow, making them more likely to be refreshed by the end user while the old query continues running until it times out.

The existing maximum query count protection doesn't help against this type of query, and other system-level protections like cgroups will still result in an OOM of the process. The hard limit on number of points per series in a result (11k atm I believe?) helps somewhat, but it is still possible to load significant amounts of data in a subquery with eg large ranges as this depends more on scrape intervals than anything configurable at the query level. The query timeout helps a little insofar as it becomes a little more difficult to load in large amounts of data with the default 2 minute timeout, but it is still very much possible to OOM a 100GB server inside that window.

### Example

```
increase(node_interrupts_total[1h])
```

Over a period of a day or so is a good example of a common metric that has high cardinality (and hence probably a lot of data). If you have thousands of machines with many CPUs each you could easily have 100k unique labelsets per time point which quickly adds up to tens of GB of data or more depending on your scrape interval. There are many others, and since a common process for metric exploration is to use the Prometheus query page for adhoc querying before moving to Grafana and refining, it is very easy to accidentally issue a bad query against these kinds of inherently high-cardinality metrics. If you have a default query timeout configured (2 minutes) and any reasonable number of multi-cpu machines, this query will very likely timeout, but depending on how big your prom node is it might OOM first. I've seen it take upwards of 40GB of RAM before timing out on one of our prom nodes.

### Summary
[...]

PR #4384, opened 2018-07-13
Add option to (approximately) limit memory use by queries.

roganartu:
Implements #4383 

Adds a cmd option to set an approximate query size limit. This limit includes input and output matrices and other assorted objects used during the evaluation of a query.
If a query exceeds this limit at any point during execution it is killed with a (new) oversize error.

Aditionally adds three new metrics:
- prometheus_engine_queries_size_max_bytes: gauge indicating the currently configured query size limit
- prometheus_engine_query_size_bytes_bucket: histogram of query sizes, up to and including the configured limit
- prometheus_engine_queries_oversize_total: counter of the number of queries that returned an error due to exceeding the configured limit

Signed-off-by: Tony Lee <tl@hudson-trading.com>

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -92,6 +92,7 @@ func main() {
 		webTimeout          model.Duration
 		queryTimeout        model.Duration
 		queryConcurrency    int
+		querySizeLimit      int64
 		RemoteFlushDeadline model.Duration
 
 		prometheusURL string
@@ -176,6 +177,9 @@ func main() {
 	a.Flag("query.timeout", "Maximum time a query may take before being aborted.").
 		Default("2m").SetValue(&cfg.queryTimeout)
 
+	a.Flag("query.max-size", "Maximum size of a single query can be (in bytes).").
+		Default("-1").Int64Var(&cfg.querySizeLimit)
+
 	a.Flag("query.max-concurrency", "Maximum number of queries executed concurrently.").
 		Default("20").IntVar(&cfg.queryConcurrency)
 
@@ -249,6 +253,7 @@ func main() {
 			prometheus.DefaultRegisterer,
 			cfg.queryConcurrency,
 			time.Duration(cfg.queryTimeout),
+			cfg.querySizeLimit,
 		)
 
 		ruleManager = rules.NewManager(&rules.ManagerOptions{
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -201,17 +201,19 @@ func Compare(a, b Labels) int {
 
 // Builder allows modifiying Labels.
 type Builder struct {
-	base Labels
-	del  []string
-	add  []Label
+	base     Labels
+	del      []string
+	add      []Label
+	lastSize int64
 }
 
 // NewBuilder returns a new LabelsBuilder
 func NewBuilder(base Labels) *Builder {
 	return &Builder{
-		base: base,
-		del:  make([]string, 0, 5),
-		add:  make([]Label, 0, 5),
+		base:     base,
+		del:      make([]string, 0, 5),
+		add:      make([]Label, 0, 5),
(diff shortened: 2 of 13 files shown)


brian-brazil:
Thank you for this PR, this is something affecting many users. I'm unclear what the actual tracking design here is, as there appears to be no real consistency and at least double counting in a places. The change is also fairly invasive, and going to a lot of effort to track amounts of memory that don't matter. The goal is to differentiate between 100MB and 1GB, not 100MB and 200MB - especially as your approach is looking at lengths rather than the actual memory taken by capacities. I'd suggest simplifying a lot, the only places that should need notable changes are eval and rangeEval.

roganartu:
> I'm unclear what the actual tracking design here is, as there appears to be no real consistency and at least double counting in a places.

This doesn't surprise me too much, as this is the result of some trial and error on our systems over some time, occasionally coming across queries very clearly exceeding the configured limit and then doing some profiling to find out why. I tried to clean it up a bit before submitting, but the conflict with the recent promql/engine.go changes made that much harder.

> The change is also fairly invasive, and going to a lot of effort to track amounts of memory that don't matter. The goal is to differentiate between 100MB and 1GB, not 100MB and 200MB - especially as your approach is looking at lengths rather than the actual memory taken by capacities.

I found that many of the capacities that are allocated were done pre-emptively for things like `Series` and `Sample` where the actual space for the labels isn't alloc'd until the `append`. In these cases I used the `ev.assertFreeSpace()` func before the initial capacity make and then actually incremented the size when the alloc including the labels was made later, eg: https://github.com/prometheus/prometheus/pull/4384/files#diff-83539908220a826e39b7ba3a4f27d5d6R1141

> I'd suggest simplifying a lot, the only places that should need notable changes are eval and rangeEval.

Do you have any specific examples of things that I've double counted or that you don't believe need to be counted? Going through the PR again I don't recall why the calls to `incSize` in `VectorBinop` exist (https://github.com/prometheus/prometheus/pull/4384/files#diff-83539908220a826e39b7ba3a4f27d5d6R1321) but I think the counting in `vectorSelector`, `matrixSelector`, `matrixIterSlice` , and `aggregation` are all necessary in practice (though could probably be simplified or improved).

My initial approach to this just tracked counts through `eval` and `rangeEval`, but I quickly found that the calls these functions made were often costly themselves and, especially for pathological queries, would often not return control the `eval`/`rangeEval` before timing out (after taking significant volumes of memory of course) so the expansion into other functions so queries could be better tracked and killed grew naturally from there.

brian-brazil:
> I found that many of the capacities that are allocated were done pre-emptively for things like Series and Sample where the actual space for the labels isn't alloc'd until the append.

That's still not taking into account how capacities work. Which is an okay tradeoff, if you don't then try to go to a lot of effort to track other things which would end up being of a similar magnitude. Capacities can make you off by a factor of 2, if you're okay with not tracking that then there's a lot of other things you don't need to track either.

> Do you have any specific examples of things that I've double counted or that you don't believe need to be counted?

I'd have to go through the entire PR again, but for example you're counting the strings for labels even though they're only allocated once and I'm not sure you're deallocating the argument matrixes when a node is complete.
https://github.com/prometheus/prometheus/pull/4384/files#diff-83539908220a826e39b7ba3a4f27d5d6R1141 is also incorrect, as it it doesn't make sense to test for space sequentially like that.

As it stands this code it too hard to understand, yet alone maintain and keep correct. There's a lot of non-obvious dependencies here between functions. A simpler solution could have similar benefits and not get in the way of future development.

roganartu:
fwiw I (obviously, since this hasn't been updated in a while) haven't had time to revisit this.

I noticed in the most recent release candidate that #4513 was included, which appears to follow a similar structure and achieve an approximation of my goal with this. As such, I'm unlikely to revisit this PR and am closing it.

PR #4386, opened 2018-07-14
Change max/min over_time to handle NaNs properly

### LINKED ISSUE #4385: max_over_time treats NaN as the max
## Bug Report
**What did you do?**
a simple query like `max_over_time(foo[30m])`

I expected to see the largest value

If there was a single `NaN` in the time range then the value returned is NaN. I made an example test case (which fails) to showcase the issue: https://github.com/jacksontj/prometheus/commit/a36f35153d4c230847bbfe992992b1f78f492f37

**Environment**

* Prometheus version:

	verified on 2.2, 2.3, and master

jacksontj:
We only want to return a NaN if the NaN is the only value

Fixes #4385


brian-brazil:
You can usual the usual PromQL tests for this, no need to write code. You should also test a series with just NaNs.

jacksontj:
Updated with tests in the `functions.test`

brian-brazil:
Thanks!

PR #4346, opened 2018-07-04
add query labels command to promtool

infoverload:
Signed-off-by: Daisy T <daisyts@gmx.com>

PR #4389, opened 2018-07-16
Allow injection of arbitrary headers in promtool, for auth etc.

tomwilkie:

### DIFF
--- a/cmd/promtool/main.go
+++ b/cmd/promtool/main.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


brian-brazil:
Should we really be putting passwords on the command line?

juliusv:
Btw., there's a bunch of open promtool PRs that add more querying which are waiting for the feature moratorium to end, this will have to be integrated with all of them as well:

https://github.com/prometheus/prometheus/pull/4346
https://github.com/prometheus/prometheus/pull/4308
https://github.com/prometheus/prometheus/pull/4247

tomwilkie:
@juliusv will wait; really want to refactor this a little so server, headers etc can be specified once.  Is the promtool CLI included in out compatibility guarantee?

brian-brazil:
Promtool is not mentioned either way, so breaking changes should be avoided without a good reason I'd say.

roidelapluie:
"As long as you are not using any features marked as experimental/unstable, an upgrade within a major version can usually be performed without any operational adjustments and very little risk that anything will break"

PR #4390, opened 2018-07-16
Adding URL support for file_sd_configs

matthiaslee:
The attached changes allow `file_sd_configs` to retrieve json/yaml target files from http(s) locations.
This is especially useful for service discovery of non-supported or custom service discovery providers.

Simply add a URL to the `files` list and it will be picked up every time the `refresh_interval` triggers, alternatively add a separate `urls` list. 

```yaml
scrape_configs:
- job_name: dynamic-prom-targets
  file_sd_configs:
   - files:
     - https://internal-backend.acme.org/dynamic-targets/prom-targets.json
```

### DIFF
--- a/discovery/file/file.go
+++ b/discovery/file/file.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -37,6 +38,7 @@ import (
 
 var (
 	patFileSDName = regexp.MustCompile(`^[^*]*(\*[^/]*)?\.(json|yml|yaml|JSON|YML|YAML)$`)
+	patURLSDName  = regexp.MustCompile(`^http[s]?://[^*]*(\*[^/]*)?\.(json|yml|yaml|JSON|YML|YAML)$`)
 
 	// DefaultSDConfig is the default file SD configuration.
 	DefaultSDConfig = SDConfig{
@@ -47,6 +49,7 @@ var (
 // SDConfig is the configuration for file based discovery.
 type SDConfig struct {
 	Files           []string       `yaml:"files"`
+	Urls            []string       `yaml:"urls,omitempty"`
 	RefreshInterval model.Duration `yaml:"refresh_interval,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thanks for your PR. We intend to have exactly one generic interface at each hook point, and we already have file_sd_configs. If you want to work from a URL you can always setup wget in a cronjob to feed into file_sd_configs..

brian-brazil:
You're basically bundling two distinct SDs into one, and we already have a generic SD and a moratorium on new SDs. If you feel our existing SDs are not sufficient you are free to integrate via file_sd_configs.

a2ndrade:
@brian-brazil would you consider accepting a new `url_sd_configs` service discovery configuration instead? Basically, reworking this PR so that that file and URL-based configurations are not conflated.

cstyan:
@matthiaslee have you tried this https://prometheus.io/blog/2018/07/05/implementing-custom-sd/ ?

brian-brazil:
No, as I said we already have a generic SD and we've also a moratorium. See https://github.com/prometheus/prometheus/tree/master/discovery#service-discovery

matthiaslee:
@cstyan I have not, but I'll take a peek, thanks.

krasi-georgiev:
I will post the canned reply as it highlight all the important points why we can't accept any new providers and the link to the example that will enable you to implement the same without any code changes in Prometheus itself.
I had quick chat with Brian on IRC and indeed at the moment even something as small as a http url_sd_configs would mean more bugs and more maintenance so we would rather focus on stability rather than adding more new stuff especially when there is already a good way to achieve the same as per the example in the link below. 


We can only provide the stability and performance we want to provide if we can properly maintain the codebase. This includes, test integrations in an automated and scalable fashion. For this reason, we are suggesting people to integrate via [file SD](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#%3Cfile_sd_config%3E) and [here](https://github.com/prometheus/prometheus/tree/master/documentation/examples/custom-sd) is a very good example to show this in practise. We would be more than happy to list this on our [integrations page](https://prometheus.io/docs/operating/integrations/).

roidelapluie:
Another approach is being implemented in #8839

PR #4398, opened 2018-07-18
Update internal architecture diagram

juliusv:


brian-brazil:
I don't think your arrow directions are consistent. What is the direction meant to represent?

juliusv:
They are meant to signal initiation, but I just noticed that I missed one in the remote storage. Will fix that later!

juliusv:
See also https://twitter.com/juliusvolz/status/1019586280026501120 :)

brian-brazil:
I had seen that, which is why I carefully checked :)

scrape manager/notifier -> sd should go the other way, as they create the SD they use.

juliusv:
> sd should go the other way, as they create the SD they use.

The SD manager is created in `main.go` for each, and the sync channel is passed in to the scrape manager and notifier there too. But even if it was the other way around this would IMO be a bit too much of an implementation detail? Like conceptually the SD manager pokes the other end when it has new targets (which is in line with the "update targets" level of abstraction on the arrow text).

Of course that opens a whole other can of worms about the arrow direction between SD manager and the outside SD service, but at least the initial request there will always come from the SD manager and that is less on the implementation-detail-level.

brian-brazil:
Okay, that one's just a bit messy then.

juliusv:
Ok, updated just the arrow direction on the remote read then.

juliusv:
Also added note about the arrow direction in `internal_architecture.md`.

brian-brazil:
:+1:

PR #4399, opened 2018-07-18
Only add LookbackDelta to vector selectors

jacksontj:

Related to #4226

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -482,13 +482,14 @@ func (ng *Engine) populateSeries(ctx context.Context, q storage.Queryable, s *Ev
 	Inspect(s.Expr, func(node Node, path []Node) error {
 		var set storage.SeriesSet
 		params := &storage.SelectParams{
-			Start: timestamp.FromTime(s.Start.Add(-LookbackDelta)),
+			Start: timestamp.FromTime(s.Start),
 			End:   timestamp.FromTime(s.End),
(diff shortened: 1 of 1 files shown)


promql/engine.go
```diff
@@ -482,13 +482,14 @@ func (ng *Engine) populateSeries(ctx context.Context, q storage.Queryable, s *Ev
...
 		params := &storage.SelectParams{
-			Start: timestamp.FromTime(s.Start.Add(-LookbackDelta)),
+			Start: timestamp.FromTime(s.Start),
 			End:   timestamp.FromTime(s.End),
 			Step:  int64(s.Interval / time.Millisecond),
 		}
 
 		switch n := node.(type) {
 		case *VectorSelector:
+			params.Start = params.Start - durationMilliseconds(LookbackDelta)
```
brian-brazil: We should probably be setting end here too for consistency

jacksontj: As it stands how start/end/step are all set before the switch, and then the attributes that change are set below. I think it could definitely make sense to move the Start down there, but the End is the same regardless of type (vector or matrix). So IMO we either leave the shared initialization above, or make each type switch have the entire creation local to the switch case.

brian-brazil: Ah I see it now, that's what I get for reviewing code just before bed.

brian-brazil:
Thanks!

PR #4403, opened 2018-07-23
Added public IP label to Azure SD targets.

### LINKED ISSUE #3970: azure sd with public IP
**What did you do?**

trying to scrape azure instances with their public IP

metrics

the current azure implementation https://github.com/prometheus/prometheus/blob/master/discovery/azure/azure.go#L38 does not allow for Public Azure IP's to be scraped thus every azure location/ ressource group would need their own prometheus instances 

* Prometheus version:
any

h7kanna:
@brian-brazil 

Hi,
I added Public IP information to Azure SD labels.
I needed this to do some load tests on Public Azure VMs.
Resolves #3970

### DIFF
--- a/discovery/azure/azure.go
+++ b/discovery/azure/azure.go
@@ -44,6 +44,8 @@ const (
 	azureLabelMachineOSType        = azureLabel + "machine_os_type"
 	azureLabelMachineLocation      = azureLabel + "machine_location"
 	azureLabelMachinePrivateIP     = azureLabel + "machine_private_ip"
+	azureLabelMachinePublicIP      = azureLabel + "machine_public_ip"
+	azureLabelMachinePublicDNSName = azureLabel + "machine_public_dns_name"
 	azureLabelMachineTag           = azureLabel + "machine_tag_"
 )
 
@@ -150,6 +152,7 @@ func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
 // azureClient represents multiple Azure Resource Manager providers.
 type azureClient struct {
 	nic network.InterfacesClient
+	ip  network.PublicIPAddressesClient
 	vm  compute.VirtualMachinesClient
 }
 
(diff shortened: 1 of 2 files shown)


discovery/azure/azure.go
```diff
@@ -291,6 +297,27 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 							labels[azureLabelMachinePrivateIP] = model.LabelValue(*ip.Properties.PrivateIPAddress)
 							address := net.JoinHostPort(*ip.Properties.PrivateIPAddress, fmt.Sprintf("%d", d.port))
 							labels[model.AddressLabel] = model.LabelValue(address)
+							// Check if Public IP is also present
```
brian-brazil: Full stop

discovery/azure/azure.go
```diff
@@ -291,6 +297,27 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
...
+							// Get the Public IP address information via separate call to the network provider.
```
brian-brazil: Azure has a quite low rate limit for API requests, so an additional request per instance let alone per network interface is problematic. Is there a way to do this without any additional requests?

h7kanna: Unfortunately there is no other way as public IP is a separate resource in Azure RM API and only referenced by ID in the network interface.

brian-brazil: That's unfortunate, this isn't the first PR we've had to reject for this. Hopefully Azure will have higher API limits or more denormalised APIs in the future.

discovery/azure/azure.go
h7kanna: Done

discovery/azure/azure.go
h7kanna: Yes I understand. But I was thinking this network call will be triggered only for VMs which have a public IP. Most VMs should have only Private IP(internal). 
Thanks for quick review. Will close this.

PR #4405, opened 2018-07-23
discovery/ec2: Maintain order of subnet_id label

xose:
The current method for deduplicating Subnet IDs on EC2 VPC instances makes the resulting order nondeterministic because it ranges over keys from a map. This is OK for filtering, but causes issues if the meta label from discovery is relabeled on the resulting metrics.

This PR changes the way the IDs are deduplicated to maintain the order of the network interfaces as returned by the EC2 API, which should be constant. It also sets the Instance's main Subnet as the first subnet on the list, to aid with relabeling if the user is only interested in the main Subnet.

### DIFF
--- a/discovery/ec2/ec2.go
+++ b/discovery/ec2/ec2.go
@@ -257,13 +257,14 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 				if inst.VpcId != nil {
 					labels[ec2LabelVPCID] = model.LabelValue(*inst.VpcId)
 
+					// Deduplicate VPC Subnet IDs maintaining the order of the network interfaces returned by EC2.
+					var subnets []string
 					subnetsMap := make(map[string]struct{})
 					for _, eni := range inst.NetworkInterfaces {
-						subnetsMap[*eni.SubnetId] = struct{}{}
-					}
-					subnets := []string{}
-					for k := range subnetsMap {
-						subnets = append(subnets, k)
+						if _, ok := subnetsMap[*eni.SubnetId]; !ok {
+							subnetsMap[*eni.SubnetId] = struct{}{}
+							subnets = append(subnets, *eni.SubnetId)
+						}
 					}
 					labels[ec2LabelSubnetID] = model.LabelValue(
 						subnetSeparator +
--- a/discovery/ec2/ec2.go
+++ b/discovery/ec2/ec2.go
@@ -37,18 +37,19 @@ import (
 )
 
 const (
-	ec2Label              = model.MetaLabelPrefix + "ec2_"
-	ec2LabelAZ            = ec2Label + "availability_zone"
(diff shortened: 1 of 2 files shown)


xose:
Updated the PR with a comment.

discovery/ec2/ec2.go
```diff
@@ -257,13 +257,20 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 				if inst.VpcId != nil {
 					labels[ec2LabelVPCID] = model.LabelValue(*inst.VpcId)
 
+					// Deduplicate VPC Subnet IDs maintaining the order of the network interfaces returned by EC2
+					// The Subnet ID from the instance metadata (if available) will always be first on the list
+					// This can help with relabeling if the user is only interested on the main Subnet
```
simonpasquier: nit picking but all sentences need to end with a dot.

xose: Done.

discovery/ec2/ec2.go
```diff
@@ -257,13 +257,20 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 				if inst.VpcId != nil {
 					labels[ec2LabelVPCID] = model.LabelValue(*inst.VpcId)
 
+					// Deduplicate VPC Subnet IDs maintaining the order of the network interfaces returned by EC2.
+					// The Subnet ID from the instance metadata (if available) will always be first on the list.
+					// This can help with relabeling if the user is only interested on the main Subnet.
```
brian-brazil: interested in

discovery/ec2/ec2.go
```diff
@@ -257,13 +257,20 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 				if inst.VpcId != nil {
 					labels[ec2LabelVPCID] = model.LabelValue(*inst.VpcId)
 
+					// Deduplicate VPC Subnet IDs maintaining the order of the network interfaces returned by EC2.
+					// The Subnet ID from the instance metadata (if available) will always be first on the list.
```
brian-brazil: It sounds like the instance metadata subnet should be a separate label, rather than overloading the meaning of this one.

xose: The subnet returned in the metadata is the subnet of the primary network interface of the instance. The primary interface cannot be changed or detached from an instance, so the subnet will always be part of the resulting list.

The primary subnet should always be present for VPC instances so the nil check should not be necessary, but I didn't want to risk a panic if there's some edge case where it's missing.

So this PR is not overloading the meaning of the label, just rearranging the list. I can move it to a separate label if you think it's better that way.

xose:
PR updated based on suggestions. I've split the addition of the new label to a separate commit.

brian-brazil:
Thanks!

PR #4409, opened 2018-07-24
promtool - add option to check rule from slice

Danm87:
I've added an option to check rules directly from a byte slice instead of a file
this helps me creating a service that can verify rules at runtime through using promtool as a lib

### DIFF
--- a/cmd/promtool/main.go
+++ b/cmd/promtool/main.go
@@ -285,6 +285,20 @@ func checkRules(filename string) (int, []error) {
 	return numRules, nil
 }
 
+func CheckRule(rule []byte) (int, []error) {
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. To keep things maintainable we do not add code to internal APIs that we don't need ourselves, as that makes things difficult to maintain as it creates a defacto public API where we've stated that there explicitly is no such API and can break it at will. I'd suggest shelling out to promtool instead.

roidelapluie:
We have looked at this PR during our bug scrub and decided to close it because we do not add code in Prometheus that we do not use, like specified above.

PR #4410, opened 2018-07-24
/config: Added `Copy to clipboard` button for /config

staff0rd:
Allows easy copying of `/config` to clipboard via button click because drag-selecting is boring.

/cc @juliusv


simonpasquier:
Would you be able to resolve the conflicts? It seems like a useful addition to me.

staff0rd:
@simonpasquier done

image

web/ui/static/css/config.css
```diff
@@ -0,0 +1,3 @@
+.btn {
+    border-radius: 0;
+}
```
juliusv: Please end text files with a newline (same below).

staff0rd: resolved in 9349188

juliusv: Hmm, now the files end with two newlines. I just meant not removing the trailing newline that should be there, as is standard in Unix/POSIX text files, see: https://stackoverflow.com/a/729795

All common editors will have a configuration setting to not remove the final newline character upon saving.

juliusv:
👍 thanks!

PR #4413, opened 2018-07-24
Pass Authorization header to RR backend

henridf:
Using a PR as the way to discuss this, given that it's pretty simple and I needed it anyway. 

An alternate approach could be to have a configurable whitelist of headers that are passed-on to the backend, but I'm not convinced we need to do that unless we have a need for it.

### DIFF
--- a/storage/remote/client.go
+++ b/storage/remote/client.go
@@ -30,6 +30,7 @@ import (
 
 	config_util "github.com/prometheus/common/config"
 	"github.com/prometheus/prometheus/prompb"
+	"github.com/prometheus/prometheus/web/api/auth"
 )
 
 const maxErrMsgLen = 256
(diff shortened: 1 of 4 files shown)


brian-brazil:
Remote read already supports bearer tokens from the config file, see https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_read

henridf:
Yes I know, but this is for a scenario (SaaS-like) where different clients are making queries, each with their own token.

brian-brazil:
Prometheus explicitly doesn't offer authorisation related features, nor should the same query give different results for different people. A more typical way to approach this is via a label that you ensure is present on all selectors.

henridf:
> A more typical way to approach this is via a label that you ensure is present on all selectors.

Do you have any examples of anyone doing that? It seems like a pretty bad user experience to have to maintain tokens in your PromQL queries (and questionable security-wise).

> Prometheus explicitly doesn't offer authorisation related features

Well, it  _does_ offer some authorisation-related features, such as the one you pointed out in your first comment above. My understanding is that Prometheus doesn't want to authenticate or participate in auth processes (which makes sense), but this PR is  just about relaying a header.

brian-brazil:
> Do you have any examples of anyone doing that?

I believe Cortex does this. The label is never visible to the user, it's injected into their PromQL for them via whatever API server they're using once they're authenticated. These sort of setups often use Prometheus components as a library, rather than as a binary.

> Well, it does offer some authorisation-related features, such as the one you pointed out in your first comment above.

That's client authentication, we don't offer any server auth.

> this PR is just about relaying a header.

It's still an authorisation feature, and also a layering violation (e.g. it doesn't work with the local tsdb), and a new way users could complicate their setup that would make supporting them more difficult. Any proposal involves passing opaque data from PromQL down to remote read is going to have the same issues. We don't allow collectors to get information from the original HTTP /metrics request for similar reasons.

henridf:
Thanks for the explanations, makes sense.

PR #4415, opened 2018-07-25
[DON'T MERGE] add owners file to enable Prow benchmarks

krasi-georgiev:

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,21 @@
+approvers:
+- sipian
+- beorn7
+- brancz
+- brian-brazil
+- Conorbro
+- discordianfish
+- fabxc
+- gouthamve
+- grobie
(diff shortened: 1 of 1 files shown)


brian-brazil:
This list doesn't match -team, and it'd be good not to have extra places to maintain the list of -team.

krasi-georgiev:
This is only for some initial tests.
The filename is required by Prow to decide who can trigger benchmark tests via comments.

I think it is hard coded, but in general I agree that it is a good idea to maintain a single file so will think best way to handle this.

krasi-georgiev:
maybe we can't change it so that when someone requests /benchmark.... Prow will make a request to check the repo members and if the commenter is also a repo member it will trigger a new job.

Issue #4407, opened 2018-07-23
Rules page returns error in V2.3.2


BartVerc:
## Bug Report
**What did you do?**
Opening prometheus in a browser on `/rules`.
**What did you expect to see?**
A page with all rules defined in the configuration.
**What did you see instead? Under which circumstances?**
I got the response: `error executing template rules.html: template: rules.html:79:57: executing "content" at <.GetEvaluationTime.S...>: can't evaluate field GetEvaluationTime in type *rules.Group`
**Environment**
Dockerhub: prom/prometheus:v2.3.2 
version 2.3.2 
go1.10.3
* System information:
Linux 4.13.0-41-generic x86_64
* Prometheus version:
version 2.3.2

simonpasquier:
Hmm, that is weird. Can you share your Prometheus logs and full configuration (including rules)?

BartVerc:
Logs:
[logs_prom_v232](https://github.com/prometheus/prometheus/files/2227019/logs_prom_232.txt)
Config:
[prometheus.yml](https://github.com/prometheus/prometheus/files/2227015/prometheus.yml.txt)
[rules.yml](https://github.com/prometheus/prometheus/files/2227016/rules.yml.txt)

juliusv:
Confirmed broken in master, working in 2.3.2. `git bisect`-ing now.

juliusv:
Badumm-tssss: I broke this in https://github.com/prometheus/prometheus/commit/9e3171f6e3ff36c820b2c987194dd1baf17e401a

Renamed some functions in there, but wasn't aware they were being called from a template!

juliusv:
(gonna work on a fix)

BartVerc:
Thanks guys!

PR #4416, opened 2018-07-25
Update method name in rules template, fix rendering

juliusv:
Fixes https://github.com/prometheus/prometheus/issues/4407

### DIFF
--- a/web/ui/bindata.go
+++ b/web/ui/bindata.go
@@ -126,7 +126,7 @@ func webUiTemplates_baseHtml() (*asset, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


juliusv:
Now this raises the general question of testing web templates. For us to get reasonably confident that they all render correctly and refer to the right Go-internal names, we would have to do a lot of the construction steps that `main()` currently does to get all the relevant backing data structures / objects up and running. Maybe it would make sense (not only for this) to refactor `main.go` to be more reusable and allow you to spin up an entire Prometheus server instance internally without spawning a new process?

brian-brazil:
:+1: 

An e2e test might be easier.

simonpasquier:
LGTM. Thanks for fixing this!

juliusv:
Filed issue for web template tests here: https://github.com/prometheus/prometheus/issues/4417

PR #4421, opened 2018-07-25
Send "Accept-Encoding" header in read request

henridf:


storage/remote/client.go
```diff
@@ -83,6 +83,7 @@ func (c *Client) Store(ctx context.Context, req *prompb.WriteRequest) error {
 		return err
 	}
 	httpReq.Header.Add("Content-Encoding", "snappy")
+	httpReq.Header.Add("Accept-Encoding", "snappy")
```
brian-brazil: This is the write request, the read request is below.

henridf: Urgh, so much for making quick changes in the GitHub UI. 
Fixed.

brian-brazil:
Thanks!

PR #3421, opened 2017-11-07
[WIP] return query samples if at least one backend has returned some data

krasi-georgiev:
fix for #2573

cc @brian-brazil , @tomwilkie 

when `expandSeriesSet` is called on `mergeIterator` , `mergeIterator.Err` returns an error even when not all iteratros have errored.

This change will return an error only if all iterators have errored.
This way  `expandSeriesSet` returns the remaining samples even when one backend is misbehaving.

@brian-brazil mentioned to add some extra info in the returned sample to specify that there was an error, so if you think this is needed please provide a sample with an error in it.

Tested with the `influxdb` remote read and write
It would be great if I test it with some more different config examples  to make sure I am not braking any use-case.

PR #4426, opened 2018-07-26
Return whatever data is available when there is a failed remote read

mknapphrt:
The goal is to return all the data that can be retrieved from storage, even if there is an error doing a remote read. Normally, the error would have caused there to be no data sent from an api call. Now, a "RemoteReadError" is propagated back to the api call. In addition, as per the discussions https://github.com/prometheus/prometheus/issues/2573 and https://github.com/prometheus/prometheus/pull/3421, if there is a remote read error, a new field is added (called extraMessage) that indicates that there was an issue doing the remote read. This way the failure doesn't get silenced.

I'm not exactly sure if the "extra message" is being placed in the best spot/has the best name, so any suggestions would be great. Thanks guys!


brian-brazil:
Thanks for looking into this. This seems like a layering violation, the query engine shouldn't have to care about how the storage is implemented.

What you want to do is in the fanout if one request for a storage fails, make it as though all requests to that storage return nothing. Returning partial data is to be avoided, it's either all or nothing.

mknapphrt:
I'm not sure I really follow, are you suggesting keeping the behavior as is then? Because that's what currently happens, if any storage fails, then it returns nothing.

mknapphrt:
Ok, I get it now. I don't see any way though to stop the the query engine from having to know about the storage in some way. For the API to to know about a remote storage error, that information would have to pass through the query engine, would it not? I agree that it does feel like a layering violation, but I don't really see how to avoid it.

brian-brazil:
You could remember that the error happened in the fanout storage, and the api check for it afterwards.

mknapphrt:
It would have to be stored in the querier though, wouldn't it? Because a new querier is generated for each query. Otherwise if it was remembered in the fanout storage, a different simultaneous query might see that it was "remembered" and report it, even if it didn't fail for that query.

brian-brazil:
No, you could provide a new fanout storage. We'll want the ability to turn on/off remote storage on a per query basis at some point.

juliusv:
> If a remote storage fails one any single selector, then we should act as though that remote storage returned nothing for all selectors. If local storage fails for any selector, we fail the entire query.

It feels a bit weird to me for a system to fail silently this way. With little local history relevant for a query, that could make results become completely unpredictable in the face of remote errors, without any errors being surfaced at all. Shouldn't this at least be opt-in via a setting? Or at least somewhere have an indicator in the result metadata that the result is partial?

brian-brazil:
> Shouldn't this at least be opt-in via a setting?

This shouldn't be optional. A Prometheus should still be able to work based on local data alone.

> Or at least somewhere have an indicator in the result metadata that the result is partial?

That's the idea.

juliusv:
> This shouldn't be optional. A Prometheus should still be able to work based on local data alone.

For rules, yes. For interactive queries though, should it just drop data by default when there are remote errors? I would imagine a query option to explicitly allow not failing on partial errors.

>> Or at least somewhere have an indicator in the result metadata that the result is partial?
>
> That's the idea.

Ok, still not ideal because we can probably wait a long time until existing API consumers get that particular memo. On the other hand they would also have to implement features needed for explicitly enabling partial results, so either case is not great.

brian-brazil:
>  For interactive queries though, should it just drop data by default when there are remote errors?

Yes, it's the only sane thing to do semantically. Consider `a unless b` where `b` failed.

juliusv:
>> For interactive queries though, should it just drop data by default when there are remote errors?
>
> Yes, it's the only sane thing to do semantically. Consider `a unless b` where `b` failed.

With "drop data by default" I meant acting and succeeding on partial data. Your example sounds like you interpreted it the other way around, or am I mistaken? You don't meant to say that you want to continue evaluating `a unless b` when `b` fails, or? Otherwise I might need more elaboration on the reasoning of that particular query. I mean it's always semantically safer to just return an error by default.

brian-brazil:
> With "drop data by default" I meant acting and succeeding on partial data.

My intention is that a query will always succeed as long as local storage succeeds. If a remote storage fails, we need to discard any successful data we got back from it before the failure though.

juliusv:
Ah ok, you were talking about some remote results succeeding, others failing. I was talking about the case where all remote queries fail intermittently (or one fails, and the data for other remote ones is dropped), leaving the caller with weird and unpredictably flip-flopping results for the older time ranges and without any of the current API users knowing that they have to look somewhere deep into result error metadata to detect that a partial error happened. That's why I considered whether it'd be better to fail the entire query by the default on any partial error anywhere, and only enabling partial evaluation when explicitly enabled.

brian-brazil:
> (or one fails, and the data for other remote ones is dropped), 

I'd expect each remote storage to be considered independently.

> That's why I considered whether it'd be better to fail the entire query by default on any partial error anywhere, and only enabling partial evaluation when explicitly enabled.

That's kinda what we currently do, and it leads to broken graphs when remote storage is broken - which is not what we should be doing reliability wise.

juliusv:
> That's kinda what we currently do, and it leads to broken graphs when remote storage is broken - which is not what we should be doing reliability wise.

Yes, I was suggesting to keep the current behavior as the default, but allowing partial failures when explicitly enabled. The question is what's worse: completely but obviously broken graphs or graphs that look like they loaded correctly, but that are actually partially wrong (and might look completely different on the next refresh). Not that I'm sure of it myself, but it seems worthy of some thought it seems?

brian-brazil:
I think it's pretty clear that we should be changing how this works by default, and I'm not sure we should even consider giving users an option to make it hard fail in this case. If they care they can look at the warnings field in the response.

> The question is what's worse: completely but obviously broken graphs or graphs that look like they loaded correctly,

Consider the case where remote storage is hard down. Completely broken alerting and a no dashboards to debug it are worse. In a standard setup older data will be missing, recent data is fine.

juliusv:
> I think it's pretty clear that we should be changing how this works by default

I don't think it's *that* clear, given that both options are suboptimal and we aim to fail hard and fast in almost all other circumstances.

> Consider the case where remote storage is hard down. Completely broken alerting and a no dashboards to debug it are worse. In a standard setup older data will be missing, recent data is fine.

Alerting rules would be exempt as they should be changed to only query local data by default anyway. Completely broken dashboards in a failure case are a better argument for allowing partial evals by default.

I would be ok to go this way, but we do have to at least recognize that it's going to be problematic, non-obvious, and can hurt people's trust in Prometheus (data in graphs flapping in and out of existence). If we do allow this partial failure eval, then we need to push really hard on Grafana folks to implement some kind of obvious indicator for the new partial failure metadata field.

brian-brazil:
> I don't think it's that clear, given that both options are suboptimal and we aim to fail hard and fast in almost all other circumstances.

I'm not sure that's true. We work off stale data rather than hard failing for SD for example.

> Completely broken dashboards in a failure case are a better argument for allowing partial evals by default.

I disagree. One of our core principles is reliability. I would not class hard failing because a non-critical component failed as being reliable. Prometheus is not meant to have a hard dependency on any external resources. The whole point of the remote APIs is to isolate us from such failures.

juliusv:
> I'm not sure that's true. We work off stale data rather than hard failing for SD for example.

I was careful to say "almost all" :)

>> Completely broken dashboards in a failure case are a better argument for allowing partial evals by default.
>
> I disagree.

I was trying to make a case *for* your proposal with that sentence :) I agree that in such a case it would suck to have completely broken dashboards and not be able to debug anything. Thus the "I would be ok to go this way, but...".

mknapphrt:
@brian-brazil The error happens at the querier level, AFAIK there's no reference to the storage from the querier. I don't see a way of remembering an error in the storage if it's happening in the querier. It could certainly be cleaned up a bit but I feel the query engine has to be involved in some way.

And as far as failing all selectors if 1 fails, how should it handle an error in the middle of getting all the selectors? If it fails in the middle then some selectors have already been added appended. Should the API go back through and remove all series that have the same instance as the failed remote storage or something? That feels like way too hacky of an approach, but I don't see another way of getting rid of those series.

mknapphrt:
`populateSeries` is already called just before the evaluation occurs, so I'm not sure where you'd want to move expandSeriesSet to. Also, i'm sure it's just me misunderstanding, but 

> and return empty results for previous queries from that storage

Makes it sound like you want to return empty for a query that already happened, i'm not sure what you mean here.

brian-brazil:
> so I'm not sure where you'd want to move expandSeriesSet to

Move it into the evaluation proper, or more exactly move the logic into the places it's used.

> Makes it sound like you want to return empty for a query that already happened, i'm not sure what you mean here.

The result only becomes "real" once we call expandSeriesSet, so if we delay that then we can not mention "missing" data as needed when the expansion starts.

mknapphrt:
The problem I'm having right now is being able to trace a SeriesSet back to the source it came from and knowing it was the same source that errored before. The remote storage(s) and local storage get merged together for each series, and the source is lost. 

One idea I had that would be easier to implement would be to return an error for the whole query if a remote storage errors in the middle of the query, and if it was down the entire time then it would exclude data only from that remote storage. A remote storage failing in the middle of a query should only be a rare event. I know it would be more ideal to be able to handle this event, I just wanted to get your opinion, because i'm finding it's going to take a good amount of reworking to be able to handle it.

brian-brazil:
> A remote storage failing in the middle of a query should only be a rare event.

All data from a remote storage is requested (in populateIterators) before the query is evaluated, so this is impossible.

brian-brazil:
Yes, and we can handle that entirely in the fanout storage.

mknapphrt:
Should the behavior be the same for the `/series` endpoint that doesn't use expandSeriesSet?

brian-brazil:
Yes, that should be an easier one to handle anyway as it's only a single select.

mknapphrt:
Well, I managed to really screw up rebasing, but the changes I've made are here. I think the code is kind of messy and hard to follow, so any suggestion would really be appreciated. I still wasn't exactly sure where you wanted expandSeriesSet, I moved it to a later point in populateSeries so that failed remote sources would be identified before expanding everything.

storage/fanout.go
```diff
@@ -66,23 +74,26 @@ func (f *fanout) Querier(ctx context.Context, mint, maxt int64) (Querier, error)
...
 	if err != nil {
 		return nil, err
 	}
-	queriers = append(queriers, querier)
+	queriers = append(queriers, primaryQuerier)
 
 	// Add secondary queriers
 	for _, storage := range f.secondaries {
 		querier, err := storage.Querier(ctx, mint, maxt)
+		if mergeQuerier, ok := querier.(*mergeQuerier); ok {
```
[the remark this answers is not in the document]
mknapphrt: I needed a reference from the querier back to the fanout storage so that if the querier had a remote error, it could notify the fanout storage and the fanout would remember which queriers had remote errors.

web/api/v1/api.go
```diff
@@ -263,11 +286,18 @@ func (api *API) query(r *http.Request) (interface{}, *apiError, func()) {
 		qs = stats.NewQueryStats(qry.Stats())
 	}
 
+	var remoteError *apiError
+	if fanout, ok := newQueryable.(storage.FanoutStorage); ok {
+		if fanout.HadRemoteError() {
```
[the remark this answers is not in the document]
mknapphrt: It's not actually being returned as an error, it's simply being used so that the wrap function can identify that there was a remote read error. It seemed like the simplest way to propagate that warning to the outermost json body. It feels like overkill to modify the apiFunc type to return an extra message when only a few methods may ever need it.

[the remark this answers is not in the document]
brian-brazil: It'd still be best to return the actual error that occurred back to the user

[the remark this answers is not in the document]
mknapphrt: Ok, in the case of multiple remote reads failing, should it be an array of all the error messages sent back?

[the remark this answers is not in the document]
brian-brazil: That's one way of doing it.

storage/fanout.go
```diff
@@ -220,12 +245,26 @@ func (q *mergeQuerier) Select(params *SelectParams, matchers ...*labels.Matcher)
 	seriesSets := make([]SeriesSet, 0, len(q.queriers))
 	for _, querier := range q.queriers {
 		set, err := querier.Select(params, matchers...)
+		if remoteSet, ok := set.(RemoteSet); ok {
+			remoteSet.SetQuerier(querier)
```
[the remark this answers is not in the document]
mknapphrt: I don't think having the map in the mergeQuerier would be accessible then from the set. Once it gets to iterating through the mergeSeriesSet there isn't currently a way to get access to the mergeQuerier. Maybe putting the map in the remoteSeriesSet would work too?

web/api/v1/api.go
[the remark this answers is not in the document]
mknapphrt: Do you have a different suggestion?

storage/fanout.go
```diff
@@ -338,11 +379,28 @@ func (c *mergeSeriesSet) Next() bool {
...
 		set := heap.Pop(&c.heap).(SeriesSet)
+		if remoteSeriesSet, ok := set.(RemoteSet); ok {
+			if contains(c.storage.failedRemoteQueriers, remoteSeriesSet.GetQuerier()) {
+				continue
+			}
+		}
 		c.currentSets = append(c.currentSets, set)
 	}
+	if len(c.currentSets) == 0 {
+		return c.Next()
```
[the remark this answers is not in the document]
mknapphrt: Why would that cause a deep recursion? Or any recursion for that matter? If the first query succeeded, then the series would get popped off the heap and added to currentSets, and it wouldn't even recursively call next then because the length would no longer be 0.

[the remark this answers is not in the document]
mknapphrt: Unless I misunderstood. Do you mean 2 queries for different remote storages on the same set of label matchers? Or 2 queries for the same remote storage on a different set of label matchers? I assume the former before. If it's the latter, then it still shouldn't cause a deep recursion because the seriesSet was popped off the heap, so the next time that Next() gets called, the series set isn't in the heap or the currentSets, so it wouldn't try to check it again.

brian-brazil:
Can you rebase this properly? It's a bit difficult to review currently with all the unrelated changes.

mknapphrt:
Besides the error coming from `TestRecoverParserRuntime()`, this should reflect all the changes now. I'm not really sure what's causing that error.

web/api/v1/api.go
```diff
@@ -174,7 +175,22 @@ func (api *API) Register(r *route.Router) {
 			if err != nil {
 				api.respondError(w, err, data)
 			} else if data != nil {
-				api.respond(w, data)
+				var warnings []error
+				switch result := data.(type) {
```
brian-brazil: You should try to avoid special casing here

web/api/v1/api.go
```diff
@@ -250,7 +267,11 @@ func (api *API) query(r *http.Request) (interface{}, *apiError, func()) {
 		defer cancel()
 	}
 
-	qry, err := api.QueryEngine.NewInstantQuery(api.Queryable, r.FormValue("query"), ts)
+	newQueryable := api.Queryable
+	if fanout, ok := newQueryable.(storage.FanoutStorage); ok {
```
brian-brazil: Using a factory function for api.Queryable would seem cleaner than checking the type here

web/api/v1/api.go
```diff
@@ -274,10 +295,16 @@ func (api *API) query(r *http.Request) (interface{}, *apiError, func()) {
 		qs = stats.NewQueryStats(qry.Stats())
 	}
 
+	var warnings []error
+	if fanout, ok := newQueryable.(storage.FanoutStorage); ok {
+		warnings = fanout.GetRemoteErrors()
```
brian-brazil: Warnings should probably be part of the main interface

web/api/v1/api.go
```diff
@@ -1016,6 +1066,32 @@ func (api *API) respondError(w http.ResponseWriter, apiErr *apiError, data inter
 	}
 }
 
+func (api *API) respondWarnings(w http.ResponseWriter, warnings []error, data interface{}) {
+	seen := make(map[string]struct{}, len(warnings))
+	var warningStrings []string
+	for _, warning := range warnings {
+		// Deduplicate warnings
```
brian-brazil: This loses information, I'd just concatenate them.

This shouldn't need a new function.

storage/fanout.go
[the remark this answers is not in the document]
brian-brazil: What if the underlying querier isn't a mergeQuerier though?

storage/fanout.go
```diff
@@ -210,7 +233,9 @@ func NewMergeQuerier(queriers []Querier) Querier {
 		return filtered[0]
```
brian-brazil: This isn't including the primaryQuerier

storage/fanout.go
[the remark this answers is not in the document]
brian-brazil: This line is calling itself, and I'm not really clear on what this logic is doing.

web/api/v1/api.go
mknapphrt: The warnings are passed to the main interface, and then removed from queryData later. I did this to try and avoid editing the apiFunc interface. Would it be preferable to add a new value to apiFunc? ie. `type apiFunc func(r *http.Request) (interface{}, *apiError, []error, func())` where the []error is the warnings returned? It would remove the type checking at line 179 then too.

web/api/v1/api.go
mknapphrt: Why would this lose information? It's just removing duplicate warnings. If you query several metrics, each one is going to return a warning about the remote read issue, but all for the same source and same warning.

storage/fanout.go
mknapphrt: The input `queriers` still includes the primary querier, and subsequently so does filtered. primaryQuerier is just passed in addition to be able to identify the primaryQuerier

web/api/v1/api.go
brian-brazil: What if two different remote reads produce the same error?

storage/fanout.go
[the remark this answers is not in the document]
mknapphrt: Then it shouldn't have to worry about knowing the storage source. What I found was that if theres >1 remote storage source, then those remote storages get combined into a mergeQuerier, which is nested inside of another mergeQuerier that contains the querier for the remote sources and the querier for the primary source. So I wanted a reference from that mergeQuerier back to the fanout storage too, so that those remote errors would get remembered.

The alternative is that there's 1 remote storage, and that isn't merged into a mergeQuerier. In that case, remote storage errors are identified in the top level mergeQuerier already so there's already a reference to the fanout storage. I hope that makes sense

web/api/v1/api.go
mknapphrt: Is that possible? The errors I get back are like `error sending request: Post http://localhost:9999/api/v1/read: dial tcp 127.0.0.1:9999: connect: connection refused`. If it were a different remote read wouldn't the ip/port have to be different too?

brian-brazil: Not necessarily, there's a few more options involved that could be in play.

web/api/v1/api.go
brian-brazil: That sounds like a better approach. We should probably change that to a struct while we're at it, 4 return values is pushing it.

web/api/v1/api.go
mknapphrt: Ok, I assumed they'd have to be from the same source.

brian-brazil: In a typical setup yes, but there's more esoteric possibilities.

storage/fanout.go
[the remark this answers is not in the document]
mknapphrt: The idea is that if for the current labels, all the series sets came from failed remote storages, then you'd end up with an empty currentSets, so you wouldn't want to return true. But it wouldn't want to return false either because the next labels pulled from the heap could have sets from non-failed remote sources (or the primary source). So it should check Next(). If it returned false then subsequent sets would be missed.

web/api/v1/api.go
mknapphrt: I'm not quite following, do you want to add a function to the Queryable interface that would do this?

mknapphrt:
I just pushed some of the changes, but for some reason they aren't showing up here on the PR (Any ideas? https://github.com/mknapphrt/prometheus/commits/failed_remote_read). I was still waiting on a few answers, but I pushed the updates to the apiFunc interface/added a new return type struct, and removed special casing checks for apiFunc responses.

storage/fanout.go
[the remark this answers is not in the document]
brian-brazil: This is still a recursion, an iterative version would be safer. This logic would also benefit from comments.

web/api/v1/api.go
```diff
@@ -170,16 +178,18 @@ func (api *API) Register(r *route.Router) {
...
-			data, err, finalizer := f(r)
-			if err != nil {
-				api.respondError(w, err, data)
-			} else if data != nil {
-				api.respond(w, data)
+			result := f(r)
+			if result.err != nil {
+				api.respondError(w, result.err, result.data)
+			} else if result.warnings != nil {
+				api.respondWarnings(w, result.warnings, result.data)
```
brian-brazil: Can this be handled by respond rather than having to handle warnings separately?

mknapphrt:
@brian-brazil I pushed some changes that remove all the type checks, It adds a lot of code extra code to accomplish it, I don't know if this is what you're looking for.

mknapphrt:
I don't think I follow, which merge-related state stuff are you talking about? The list of failed queriers/errors that's in fanout? And I don't understand what this new class is supposed to do or what you mean by reusing mergeQuerier?

brian-brazil:
> The list of failed queriers/errors that's in fanout? 

Yes

> And I don't understand what this new class is supposed to do

Contain that state, rather than mixing it it in with Storage.

> what you mean by reusing mergeQuerier?

Instead of creating a whole new class for this, mergeQuerier might suffice.

mknapphrt:
I need some way of accessing that 'Container', whether it be in the mergerQuerier or somewhere else. The api doesn't have access to the querier that the fanout storage creates. How would it get the state if it were contained in the mergeQuerier (or a new object either)?

brian-brazil:
The API has access to the Queryable, which you've currently got a "GetRemoteErrors" method on - which does that. Though GetWarnings would be more generic.

mknapphrt:
Right, GetRemoteErrors is a function on the fanout storage, not the mergeQuerier though. I agree that GetWarnings would be more appropriate.

mknapphrt:
@brian-brazil I'm still now sure how you want to access the warnings, do you think you could explain your idea more or give some suggestions? I don't see how the mergeQuerier would be able to store it and still be accessible from the api.

brian-brazil:
mergeQuerier is a Querier, so will have a GetWarnings method. It you pass the mergeQuerier down to NewMergeSeriesSet, you can then keep the state in there.

mknapphrt:
Can you explain again why it's unreasonable to pass these warnings through the engine and return them as a result of the query? It seems like this would be a much easier solution if `Select` returned a set, an error, and warnings and the warnings got passed up exactly the same way the errors did. I don't see how this is would be a layering violation.

PR #4430, opened 2018-07-27
Replace go-bindata by vfsgen

### LINKED ISSUE #2411: Evaluate alternative static file embedder
The tool we use to embed assets into the binary has issues around timestamping. That's the reason why we don't always trigger it on build (commit pollution), which often leads to it being forgotten on actual changes.

https://github.com/rakyll/statik might be an alternative.

simonpasquier:
Closes #2411.

Looking at https://tech.townsourced.com/post/embedding-static-files-in-go/ (which was mentioned in the issue), [vfsgen](https://github.com/shurcooL/vfsgen) has all the needed features.

In particular:
* Reproducible builds (no issue with timestamping).
* Well maintained and relatively popular.
* Integration with `go generate`.
* Self-contained (no external dependency).


grobie:
With go generate support, there isn't any support anymore to check in the generated files, right? That would remove the need to remind contributors to run `make assets` in pull requests.

simonpasquier:
@grobie I'm not sure. AFAICT it is more a personal preference: some projects like kubernetes still commit all the generated code. I don't have any personal preference though.

simonpasquier:
@brian-brazil can you have a look?

web/ui/README.md
```diff
@@ -1,13 +1,13 @@
...
+using the vfsgen library (c.f. Makefile).
 
 During development it is more convenient to always use the files on disk to
 directly see changes without recompiling.
-To make this work, set the environment variable `DEBUG=1`, run `make assets`, and then `make build`.
- 
-This will put `go-bindata` in DEBUG mode where it serves from your local filesystem.
+To make this work, add `-tags dev` to the `flags` entry in `.promu.yml`, and then `make build`.
+
+This will serves all files from your local filesystem.
```
brian-brazil: serve

Makefile
```diff
@@ -23,13 +23,12 @@ STATICCHECK_IGNORE = \
...
-assets:
+assets: $(VFSGENDEV)
 	@echo ">> writing assets"
-	@$(GO) get -u github.com/jteeuwen/go-bindata/...
-	@go-bindata $(bindata_flags) -pkg ui -o web/ui/bindata.go -ignore '(.*\.map|bootstrap\.js|bootstrap-theme\.css|bootstrap\.css)'  web/ui/templates/... web/ui/static/...
-	@$(GO) fmt ./web/ui
+	cd $(PREFIX)/web/ui && go generate
+
+$(VFSGENDEV):
+	cd $(PREFIX)/vendor/github.com/shurcooL/vfsgen/ && go install ./cmd/vfsgendev/...
```
brian-brazil: I'm not sure we should be modifying the user's environment like this

simonpasquier: That's a good point. I've replaced `vfsgen`by a custom helper program that does the same (just a couple of lines).

Makefile
```diff
@@ -23,13 +23,8 @@ STATICCHECK_IGNORE = \
...
-	bindata_flags = -debug
-endif
-
 .PHONY: assets
 assets:
 	@echo ">> writing assets"
-	@$(GO) get -u github.com/jteeuwen/go-bindata/...
-	@go-bindata $(bindata_flags) -pkg ui -o web/ui/bindata.go -ignore '(.*\.map|bootstrap\.js|bootstrap-theme\.css|bootstrap\.css)'  web/ui/templates/... web/ui/static/...
-	@$(GO) fmt ./web/ui
+	$(GO) get -u github.com/shurcooL/vfsgen
```
brian-brazil: Is this still needed given its vendored?

simonpasquier: Yes. I have removed it from `vendor/` in the last version because otherwise `govendor ilst +unused` complains that it isn't used.

simonpasquier: It has a `// +build ignore` constraint https://github.com/prometheus/prometheus/pull/4430/files#diff-eb481951f2846eb52fb4b11bfa5b8afdR14

simonpasquier: Using a blank import has made `govendor` happy.

brian-brazil:
:+1:

grobie:
> AFAICT it is more a personal preference: some projects like kubernetes still commit all the generated code. I don't have any personal preference though.

I'm not aware that anyone is actively reviewing the code generated by go-bindata / vfsgen, so I don't see a use case for actually checking in that file. As it's now possible to easily generate the files during compilation time on-the-fly, I'd vote to remove the need to ask every contributor to remember executing `make assets`.

simonpasquier:
> I'd vote to remove the need to ask every contributor to remember executing make assets

@brian-brazil @juliusv what are your views on this?

brian-brazil:
If we've gotten the system to a place where it's no longer required, I see no issue with removing it.

grobie:
Ah, I know remember the issue with `go generate`, it's not part of go build, so one couldn't longer do `go get github.com/prometheus/prometheus/cmd/prometheus` to install Prometheus. In that case, maybe leave it as it is right now @simonpasquier.

simonpasquier:
@grobie yep it seems to be the less disruptive way for now. Also when the generated code isn't committed, we need to add the `assets` prerequisite to a few targets (eg `build`, `test`, ...) as well as modify the CircleCI and Travis CI configurations.

brian-brazil:
I already gave a +1

grobie:
Thanks, great work @simonpasquier!

Issue #4428, opened 2018-07-27
Feat Req: Remote read hint for /api/v1/series


mtanda:
## Proposal
**Use case. Why is this important?**
Currently, `GET /api/v1/series` will read datapoints from remote read endpoint.
If remote read endpoint has "label" index, the endpoint don't need actual datapoints, just return matched label pair is enough to work the API.

juliusv:
Related: https://github.com/prometheus/prometheus/issues/4057

gouthamve:
This is very similar in scope to #4057, closing this in our bug scrub as duplicate.

Issue #4057, opened 2018-04-06
Using `Select()` for both sample+metadata queries is unsuitable for remote storage implementations


juliusv:
The `storage.Querier` interface has a `Select()` method which can be used to retrieve either only metadata or bulk sample data of series. In the local storage case, this works nicely, as sample data is loaded lazily when accessed, and thus metadata-only queries never over-fetch sample data. This breaks down for remote storage implementations of the same interface, where we cannot afford to lazily fetch sample data only as it's accessed (that would require more remote round trips). We need some way of knowing beforehand (when we call `Select()`) whether a query is metadata-only or whether it actually needs sample data. Either a separate method or query parameter would work.

Context: Cortex reuses Prometheus's web API and PromQL packages to offer comparable functionality based on a different storage engine. In the past, we were able to solve this issue because the Prometheus web API packages allowed injecting two different queriers: one that would be used as part of the PromQL engine (the full sample querier) and one that would be used for metadata queries. This got broken in https://github.com/prometheus/prometheus/commit/7ccd4b39b89e936bf0f9c47811e3fe2b20d1fe3b#diff-d81f5cda89ea7b129ba708b586c2bc83L132, due to a PromQL engine restructuring (the querier is not part of the engine anymore, it gets passed in for every query, and the web API only knows about one querier now, because we cannot encapsulate the other one in the engine anymore).

The issue about this on the Cortex side is https://github.com/weaveworks/cortex/issues/787

juliusv:
I already mentioned the option of adding a separate querier method or `Select()` parameter.

The other option would be to modify the web API package to allow explicitly passing in two different `Queryable`s, one that's used only for metadata queries and another that's used for full (PromQL and remote read) queries.

jacksontj:
I have this same issue for [promxy](https://github.com/jacksontj/promxy. I could also live with a flag in the `SelectParams` but that also seems a bit hacky. To me it seems like having a second interface that you could optionally implement (something like `MetadataQuerier`) which would be called for just metadata queries would be clean.

krasi-georgiev:
@juliusv does anything need to be done on the tsdb side? 
If not would you mind removing the local storage label so that it doesn't appear i my queue :stuck_out_tongue_winking_eye:

juliusv:
@krasi-georgiev I changed the labels from local to remote storage.

krasi-georgiev:
thanks

Harkishen-Singh:
[...]

PR #4431, opened 2018-07-28
Add series match query hint

mtanda:
Implement https://github.com/prometheus/prometheus/issues/4428


mtanda:
@brian-brazil How do you think this change?

brian-brazil:
This is more @tomwilkie's area.

brian-brazil:
> The inclusion of startMs and endMs is a bit strange, as the query itself has that.

This is to do with whole range vs. range offset needs. This makes sense consistency wise.

mtanda:
> IIUC we've been using the absence of a ReadHints to signify a metadata-only fetch, so I think this is an improvement.

Sorry, I didn't know the detail of remote read.
I can use same method to detect metadata-only fetch if the interface is never changed.

> And the nagging question is whether or not this should be a separate RPC altogether? I'm not a big fan of overloading the interface like this.

I need same functionality for remote read, which Prometheus provide.
`GET /api/v1/label/<label_name>/values` seems to not supported in remote read.
To support the API, current remote read interface can't handle, so need separete RPC.

The remote read interface can't handle all Prometheus API, and `GET /api/v1/series` interface and `GET /api/v1/query_range` is little differ, I think adding RPC is good decision.

juliusv:
I also prefer making each type of query explicit along the entire path (new method on the storage / querier, new remote read endpoint or request message type). The whole read hints business is getting out of hand and is confusing IMO. We should make sure though that we have some kind of consistent vision for remote read should be going, and when we're fine breaking it (since it's strictly still marked experimental, but of course we don't want to break it all the time).

roidelapluie:
We have looked at this during our bug scrub and decided to close this. There is further discussions needed before we can decide what is the best approach there. See #4057 and #5547  for the discussions.

PR #4432, opened 2018-07-28
Change strings to named consts

niya3:
@brian-brazil 

via https://go-critic.github.io/overview#namedConst-ref

Signed-off-by: ildar nizamov <il.nizamov@ya.ru>


brian-brazil:
Thanks for your PR. With this change our tests would no longer fail if the constant were to be changed. We want to test that the output is an exact value, not that it matches some internal constant.

PR #4433, opened 2018-07-28
Change if-else-chain to switch-case

niya3:
@brian-brazil

via https://go-critic.github.io/overview#ifElseChain-ref

Signed-off-by: ildar nizamov <il.nizamov@ya.ru>

### DIFF
--- a/documentation/examples/remote_storage/remote_storage_adapter/influxdb/client.go
+++ b/documentation/examples/remote_storage/remote_storage_adapter/influxdb/client.go
@@ -289,13 +289,14 @@ func mergeSamples(a, b []*prompb.Sample) []*prompb.Sample {
 	result := make([]*prompb.Sample, 0, len(a)+len(b))
 	i, j := 0, 0
 	for i < len(a) && j < len(b) {
-		if a[i].Timestamp < b[j].Timestamp {
+		switch {
+		case a[i].Timestamp < b[j].Timestamp:
 			result = append(result, a[i])
 			i++
-		} else if a[i].Timestamp > b[j].Timestamp {
+		case a[i].Timestamp > b[j].Timestamp:
 			result = append(result, b[j])
 			j++
-		} else {
+		default:
 			result = append(result, a[i])
 			i++
 			j++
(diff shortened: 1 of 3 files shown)


brian-brazil:
Thanks for the PR, why do you want to change this? This seems more like a personal preference than a problem.

Quasilyte:
https://golang.org/doc/effective_go.html#switch

```
Go's switch is more general than C's. The expressions need not be constants or even integers, the cases are evaluated top to bottom until a match is found, and if the switch has no expression it switches on true. It's therefore possible—and idiomatic—to write an if-else-if-else chain as a switch. 
```

It's more idiomatic.
There are more static checks on switches.
There are more optimizations for switch statements.
They are easier to read for like 70% of Go programmers.

brian-brazil:
This is harder to read for me, as I'm spending wondering why it's using a switch in an unusual fashion rather than a more typical if. Nor have I come across other Go code using switch statements in this way.

I think you're reading that doc a bit broadly, as the example presented is simpler than these.

> There are more static checks on switches.
> There are more optimizations for switch statements.

Style should not be determined by whatever the optimisations or static checks of the day are.

brian-brazil:
I do see a case for switching the numScrapes to a more standard switch statement.

I appreciate that you're trying to improve Go codebases and I see that you've detected real bugs in other projects, but not all things that a static checker detects are automatically problems.

Quasilyte:
@brian-brazil, style checks are very specific subject in general.
Maybe it's a good idea to avoid them during audit of external projects.
Hope no one is offended. It's easy to become emotional when style-related things are concerned.

Quasilyte:
On the bright side, we found 0 bugs in prometheus so far.
We'll come back with more diagnostic checks rather than style checks. :wink:

brian-brazil:
> style checks are very sensitive subject in general.

They can be alright. Things that are clear improvements are good, for example one of the more recent set of style improvements we had go through was removing redundant type casts.

> On the bright side, we found 0 bugs in prometheus so far.

That's good to hear.

PR #4436, opened 2018-07-30
Add tracing spans to promql

gouthamve:
@tomwilkie @bboreham This'll help in prometheus when we add proper tracing support, but currently we'd like to use this for Cortex.

screen shot 2018-07-30 at 15 32 33

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -161,6 +161,35 @@ func contextDone(ctx context.Context, env string) error {
 	}
 }
 
+// spanTimer unifies tracing and timing, to reduce repetition.
+type spanTimer struct {
+	timer     *stats.Timer
+	observers []prometheus.Summary
+
+	span opentracing.Span
+}
+
+func newSpanTimer(ctx context.Context, operation string, timer *stats.Timer, observers ...prometheus.Summary) (*spanTimer, context.Context) {
+	span, ctx := opentracing.StartSpanFromContext(ctx, operation)
+	timer.Start()
+
+	return &spanTimer{
+		timer:     timer,
+		observers: observers,
+
+		span: span,
+	}, ctx
+}
+
+func (s *spanTimer) Finish() {
+	s.timer.Stop()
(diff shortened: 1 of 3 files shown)


promql/engine.go
```diff
@@ -310,25 +339,25 @@ func (ng *Engine) exec(ctx context.Context, q *query) (Value, error) {
...
+	queueSpanTimer, _ := newSpanTimer(ctx, "promqlExecQueue", q.stats.GetTimer(stats.ExecQueueTime), ng.metrics.queryQueueTime)
 
 	if err := ng.gate.Start(ctx); err != nil {
 		return nil, err
 	}
 	defer ng.gate.Done()
 
-	queueTimer.Stop()
-	ng.metrics.queryQueueTime.Observe(queueTimer.ElapsedTime().Seconds())
+	queueSpanTimer.Finish()
```
tomwilkie: Is it a problem that the if the `gate.Start` fails, this won't be stopped?

gouthamve: Hmm, yep. But it wasn't stopped before either....

promql/engine.go
```diff
@@ -355,23 +384,25 @@ func durationMilliseconds(d time.Duration) int64 {
...
-	ng.metrics.queryPrepareTime.Observe(prepareTimer.ElapsedTime().Seconds())
-
-	// XXX(fabxc): the querier returned by populateSeries might be instantiated
-	// we must not return without closing irrespective of the error.
-	// TODO: make this semantically saner.
-	if querier != nil {
-		defer querier.Close()
-	}
+	{
+		// New block because we want to throw away this ctx. Propagating it further makes every call a child of prepare in traces.
```
juliusv: Or just call it `ctxPrepare` / `prepareCtx`?

juliusv:
:+1: Thanks.

(I think in the long it'd make sense to move the promql-specific timer names back to the promql package, but keep the generic timer code in a separate package)

PR #4438, opened 2018-07-30
promtool - Use API querier with built-in timeout

chyeh:
In this PR, `prometheusAPIQuerier` with built-in timeout is introduced for eliminating duplicate code.


chyeh:
This PR is the followup of #4247. In #4247, @juliusv mentioned:

>Since you are introducing that new API HTTP client with built-in timeout in http.go, should it be used here and in the other function below as well?

I found out I can't just use `prometheusHTTPClient` type variable as an argument for `v1.NewAPI` since it doesn't implement the `Client` interface in `github.com/prometheus/client_golang/api`. The behaviour of these two are fundamentally different. However, the duplicate `ctx` code can still be eliminated.

roidelapluie:
Thanks for your PR.

We have looked at that during our bug scrub and decided to close this, as the PR is now outdated by the many changes in the codebase itself. The API client takes context which enables us to nicely play with timeouts while using golang's defaults http clients.

Also, we noticed that reflect should be avoided wherever possible.

PR #4452, opened 2018-08-02
Allow '.' and '-' in metric names.

josefs:
There are several standards which uses metric names that are not allowed
in Prometheus. For example:

* The 3GPP standard for mobile telephony uses '.' in metric names
  (See TS 32.404, chapter 3.3, bullet “e”)

* YANG data models should use '-' as a separator for all names, including
  metric names.
  (See rfc6087-bis: https://tools.ietf.org/id/draft-ietf-netmod-rfc6087bis-13.html#rfc.section.4.3.1)

This patch makes it possible to use Prometheus out of the box with these
standards, without having to do any name conversion.

### DIFF
--- a/pkg/textparse/lex.l
+++ b/pkg/textparse/lex.l
@@ -43,7 +43,8 @@ func (l *lexer) Lex() token {
 
 D     [0-9]
 L     [a-zA-Z_]
-M     [a-zA-Z_:]
+M     [a-zA-Z_:.-]
+N     [a-zA-Z_:.-]
 C     [^\n]
 
 %x sComment sMeta1 sMeta2 sLabels sLValue sValue sTimestamp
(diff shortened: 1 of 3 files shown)


brian-brazil:
Thanks for your PR. The Prometheus data model is pretty heavily baked into the software, and the ecosystem. While there are other formats out there, supporting them directly would cause large amounts of pain to our users. If you have another format you'll need to write an exporter to convert it.

josefs:
Thanks for the quick response @brian-brazil! Just to be clear, are you saying this change will break other things in the Prometheus eco system?

brian-brazil:
Yes, our data model is that metric names do not contain . or -. This would break promql for example.

josefs:
I'm confused. My patch updates promql to accept the new names. So I don't expect that to break.

My understanding is that nothing will break as long as users keep using the old names. What I expect to happen is that if someone starts using . and - in their names, then those metrics will be dropped in certain tools. But that's what's happening right now, except Prometheus is dropping the metrics instead of some other tool.

I understand you're not keep on accept this change, but I'd still like to understand better what problems it might cause.

brian-brazil:
Consider how is `a-b` parsed now, and how it'd be parsed after your change. That's just one of the many reasons this would be a major breaking change.

> But that's what's happening right now, except Prometheus is dropping the metrics instead of some other tool.

If you provide data in an invalid format, Prometheus will consider the entire scrape failed.

josefs:
Ok, thanks. That's clearly an oversight on my part.

Last question: would you consider adding just . as a valid character in metric names?

brian-brazil:
No, with the same reasoning. It's not valid in our data model.

PR #4453, opened 2018-08-02
Use circle-ci builds to make docker images for PR

sylr:
I changed the docker image used from circleci/golang:1.10 to golang:1.10 because,
surprisingly enough, there seems to be an incompatibility of the tar and gzip tools
between the machine executor and the docker images built by circle-ci. It makes it
impossible to attach a workspace inside a container which as been persisted from
a machine executor.

New/Changed jobs:

- promu: factorizes promu building for following build and crossbuild steps
- crossbuild: replaces current build job which does a promu crossbuild
- crossbuild-linux-amd64: calls promu crossbuild to only build linux/amd64

New/Changed workflow:

- prometheus-pr: builds pull requests

```
 -- promu -+- crossbuild-linux-amd64 -+- docker_push_pr
           |                          |
           +- test                    +- crossbuild
```

crossbuild step lasts 1 hour so we first use build to only make the binary needed
for the docker image and we build that docker image. This allows to have a docker
image ready to test in a few minutes. Then we crossbuild.

Docker tags are named `pr-<github-pr-#>` and pushed to quay.io/prometheus repository.

- prometheus-master-and-releases: replaces prometheus workflow

```
 -- promu --- test --- crossbuild -+- docker_hub_master_and_releases
                                   |
                                   +- docker_hub_release_tags
```

This does exactly what old prometheus workflow used to except it also builds a
master like image for each release-.* branches.

Signed-off-by: Sylvain Rabot <s.rabot@lectra.com>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,48 +1,103 @@
 ---
 version: 2
 
+docker-install: &docker-install |
+  set -x && \
+  export DOCKER_VERSION=$(curl --silent --fail --retry 3 https://download.docker.com/linux/static/stable/x86_64/ | grep -o -e 'docker-[.0-9]*-ce\.tgz' | sort -r | head -n 1) \
+  && DOCKER_URL="https://download.docker.com/linux/static/stable/x86_64/${DOCKER_VERSION}" \
+  && curl --silent --show-error --location --fail --retry 3 --output /tmp/docker.tgz "${DOCKER_URL}" \
+  && tar -xz -C /tmp -f /tmp/docker.tgz \
+  && mv /tmp/docker/* /usr/bin
+
 jobs:
-  test:
-    # Whenever the Go version is updated here, .travis.yml should also be
-    # updated.
-    docker:
-    - image: circleci/golang:1.10
-    working_directory: /go/src/github.com/prometheus/prometheus
-    resource_class: large
-
-    steps:
-    - checkout
-    - run: make promu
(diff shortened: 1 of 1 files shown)


sylr:
It seems that random Joes like myself can't access the `$QUAY_LOGIN` & `$QUAY_PASSWORD` and thus can't push on quay.io .

I also came to think that it would be best to have a different repo for PR to avoid polluting the official one, `quay.io/prometheus/prometheus-pr` maybe ?

sylr:
@SuperQ @simonpasquier Could you review this ?

SuperQ:
Interesting idea, I need to think about this a bit.

One thing I was considering was a way to improve the crossbuild speed by spawning multiple parallel builds. That would reduce the complication added by separating amd64 build. We also want to be able to produce non-amd64 docker images eventually which would mean we need the full crossbuild anyway.

sylr:
@SuperQ did you think about this ?

simonpasquier:
I see how it could be useful, for instance to help with prombench (see the description of #4685 too) but I'm a bit afraid that it would clutter the list of tags and history in the registries.

https://hub.docker.com/r/prom/prometheus/tags/
https://quay.io/repository/prometheus/prometheus?tab=tags
https://quay.io/repository/prometheus/prometheus?tab=history

sylr:
@simonpasquier Yes that's why I think we could push the PR docker image to a separate dedicated registry.

PR #4454, opened 2018-08-03
promtool - Adding --step flag to 'query range' subcommand

hairyhenderson:
Sometimes I want to do a range query and the default step size is just not quite right... So I've implemented a `--step` flag, that works like this:

```console
$ promtool query range --start 1533219360 --end 1533220976 --step 60s http://localhost:9090 'count(up)'
...
```

The default behaviour remains unchanged (when `--step` is `0` or not specified).

I've added some very basic tests.

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

### DIFF
--- a/cmd/promtool/main.go
+++ b/cmd/promtool/main.go
@@ -74,6 +74,7 @@ func main() {
 	queryRangeExpr := queryRangeCmd.Arg("expr", "PromQL query expression.").Required().String()
 	queryRangeBegin := queryRangeCmd.Flag("start", "Query range start time (RFC3339 or Unix timestamp).").String()
 	queryRangeEnd := queryRangeCmd.Flag("end", "Query range end time (RFC3339 or Unix timestamp).").String()
+	queryRangeStep := queryRangeCmd.Flag("step", "Query step size (duration).").Duration()
 
 	querySeriesCmd := queryCmd.Command("series", "Run series query.")
 	querySeriesServer := querySeriesCmd.Arg("server", "Prometheus server to query.").Required().URL()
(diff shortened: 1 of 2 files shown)


cmd/promtool/main.go
```diff
@@ -74,6 +74,7 @@ func main() {
 	queryRangeExpr := queryRangeCmd.Arg("expr", "PromQL query expression.").Required().String()
 	queryRangeBegin := queryRangeCmd.Flag("start", "Query range start time (RFC3339 or Unix timestamp).").String()
 	queryRangeEnd := queryRangeCmd.Flag("end", "Query range end time (RFC3339 or Unix timestamp).").String()
+	queryRangeStep := queryRangeCmd.Flag("step", "Query step size (duration - specify unit).").Duration()
```
simonpasquier: (nit) I'd remove ` - specify unit`.

hairyhenderson: @simonpasquier sure - I waffled on that one.

cmd/promtool/main_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+	"time"
+)
+
+func TestQueryRange(t *testing.T) {
+	s, getURL := mockServer(200, `{"status": "success", "data": {"resultType": "matrix", "result": []}}`)
```
juliusv: `defer s.Close()` would be nice here

juliusv:
:+1: Thanks!

PR #4457, opened 2018-08-04
Show rule evaluation errors on rules page

### LINKED ISSUE #4446: Show rule evaluation errors on rules page
## Proposal

On the `/rules` UI page, show the last evaluation status and any errors for each rule. This is similar to how the status and errors for targets are shown on `/targets`.

Humans are not perfect, and sometimes rule evaluation fails. This cannot be fully solved by testing beforehand, since e.g. many-to-many matching errors and timeouts are data dependent.

Right now, we need to find and look at the logs to see this error. Surfacing the last error would make this much easier. It's not a great help for fleetingly broken rules but most of the time when they break they stay broken for a while.

* Logs:
```
2018-08-01_11:42:43.52603 level=warn ts=2018-08-01T11:42:43.525846289Z caller=manager.go:343 component="rule manager" group=<redacted> msg="Evaluating rule failed" rule="record: <redacted> \nexpr: (<redacted> / on(a, b) <redacted>)\n  * 100\n" err="many-to-many matching not allowed: matching labels must be unique on one side"
```

noqcks:
This is my initial stab at #4446 

image

I am able to get the happy path working, but I can't seem to replicate a query error.. I'm not seeing one in my logs when using the expr given by @matthiasr in the linked issue. Is there a query I can run locally to get an error that someone might have off hand? Thanks!

Fixes #4446

cc: @brian-brazil


brian-brazil:
Thanks for this. If you configure a remote read that doesn't work/exist, it'll cause your rules to fail.

I don't think we need the state, that's more a target thing.

rules/alerting.go
```diff
@@ -109,6 +109,12 @@ type AlertingRule struct {
 	// only after the restoration.
 	restored bool
 
+	// the health of the alerting rule
```
juliusv: Style nit: comments should start upper-cased and end with a period (sentence-style), here and everywhere else.

rules/recording.go
```diff
@@ -112,6 +123,16 @@ func (rule *RecordingRule) SetEvaluationDuration(dur time.Duration) {
 	rule.evaluationDuration = dur
 }
 
+// Health returns the current health of the recording rule
+func (rule *RecordingRule) Health() RuleHealth {
+	return rule.health
```
juliusv: I think this and `LastError()` also need to be protected against concurrent access via the mutex, like in `GetEvaluationDuration()`?

While the web UI gets a copy of the list of rules, it's a copy of pointers.

noqcks:
@matthiasr fixed up some code comments and added mutex as per comment from @juliusv. Could you review again? 

I also prefer the state field for the same reasons you mentioned.

rules/alerting.go
```diff
@@ -137,6 +144,34 @@ func (r *AlertingRule) Name() string {
 	return r.name
 }
 
+// SetLastError sets the current error seen by the alerting rule
```
juliusv: Same nit for all the method comments ;)

juliusv:
Looks generally good to me besides last nits.

I don't feel strongly on having the state or just the error from the UI side. From the code side, having just the error would be less code, which is always good, but `¯\_(ツ)_/¯`.

noqcks:
seems like the tests are hanging? 🤔

juliusv:
@noqcks The Travis tests are causing trouble for us unfortunately :( I re-prodded them, but the other tests are already good.

rules/recording.go
```diff
@@ -31,9 +31,16 @@ import (
...
-	labels             labels.Labels
+	name   string
+	vector promql.Expr
+	labels labels.Labels
+
+	// The health of the recording rule.
+	health RuleHealth
+
+	// The last error seen by the recording rule.
+	lastError error
```
simonpasquier: Same here.

rules/alerting.go
```diff
@@ -109,6 +109,12 @@ type AlertingRule struct {
 	// only after the restoration.
 	restored bool
 
+	// The health of the alerting rule.
+	health RuleHealth
+
+	// The last error seen by the alerting rule.
+	lastError error
+
```
simonpasquier: nit but since `health` and `lastError` are protected by the mutex, it would be better to have them after `mtx`.

mxinden:
>  Also @mxinden we probably want this reflected in the Rules API?

@simonpasquier Good idea, thanks for the link. If there are no objections, I can follow up with that, once this PR is merged.

juliusv:
:+1: now, will merge after the (non-Travis, at least) tests pass

juliusv:
Oh drat, that's what we get for not fixing our Travis tests that fail all the time for other reasons. Thought somehow that one of the other CIs also ran tests. @noqcks, would you have time to fix this at the moment or should I take a look?

fabxc:
Yea, it's a bit confusing with all our different CIs doing different stuff.
It's all good though, just needed 2 lines changed to fix it, which I just folded into #4471.

juliusv:
@fabxc Oh great, thanks!

PR #4462, opened 2018-08-04
Change snapshot file permissions from 777 -> 750

juliusv:
This should be more secure and was pointed out by the `gosec` static
analysis tool.


juliusv:
TSDB also does `0777`. So I guess maybe just leave it as it is and leave it up to the user's umask?

PR #4463, opened 2018-08-04
promql: Propagate Inspect() errors to the user

juliusv:
...instead of logging them.

### DIFF
--- a/promql/ast.go
+++ b/promql/ast.go
@@ -336,6 +336,6 @@ func (f inspector) Visit(node Node, path []Node) (Visitor, error) {
 // Inspect traverses an AST in depth-first order: It starts by calling
 // f(node, path); node must not be nil. If f returns a nil error, Inspect invokes f
 // for all the non-nil children of node, recursively.
-func Inspect(node Node, f inspector) {
-	Walk(inspector(f), node, nil)
+func Inspect(node Node, f inspector) error {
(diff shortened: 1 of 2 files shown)


brian-brazil:
:+1:

juliusv:
Note that the tests here are failing because they and the API v1 still expect something around an `ErrStorage` error type, but we never actually ever create errors of that sort anywhere (also not sure how they are currently propagated in the test). That seems broken, will have a look...

juliusv:
(DON'T MERGE)

juliusv:
Aha! The `ErrStorage` error type was only ever used by Cortex (see https://github.com/weaveworks/cortex/search?q=errstorage&unscoped_q=errstorage). The existing test ensures that it is still propagated correctly. So I will have to preserve that... fix incoming.

juliusv:
Ok, so it turns out that the storage errors were already propagated correctly through an `err` variable in the outer function which the `Inspect()` calls close over... this works, but I wonder if it wouldn't be better to not do that kind of convoluted thing and do it by returning the error from `Walk()` instead.

brian-brazil:
Yeah, doing it explicitly is a lot clearer.

codesome:
This seems still relevant. @juliusv could you rebase the PR?

juliusv:
Closing, as it doesn't seem that errors can happen in this path anymore.

PR #4464, opened 2018-08-04
rules: HTML-escape rule YAML marshal errors

juliusv:
This was pointed out by `gosec`.


brian-brazil:
This probably deserved a unittest.

juliusv:
@brian-brazil I'm not sure how to provoke a YAML marshalling error though...

brian-brazil:
:+1: 

As I've said before, I'd prefer we weren't hand assembling HTML and rather left it up the Go templates escaping.

Issue #4441, opened 2018-07-31
Panic in EC2 discovery: NetworkInterfaces


nhinds:
## Bug Report
**What did you do?**
Ran Prometheus 1.8.2 with EC2 service discovery across many EC2 instances in unknown states.

**What did you expect to see?**
EC2 service discovery works without crashing.

**What did you see instead? Under which circumstances?**
Prometheus panicked about a nil pointer dereference:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xa262b2]

goroutine 18710894 [running]:
github.com/prometheus/prometheus/discovery/ec2.(*Discovery).refresh.func2(0xc43bf84560, 0x1, 0x0)
        /go/src/github.com/prometheus/prometheus/discovery/ec2/ec2.go:190 +0x7f2
github.com/prometheus/prometheus/vendor/github.com/aws/aws-sdk-go/service/ec2.(*EC2).DescribeInst
ancesPages.func1(0x187aaa0, 0xc43bf84560, 0x1, 0xc47d437da0)
        /go/src/github.com/prometheus/prometheus/vendor/github.com/aws/aws-sdk-go/service/ec2/api
.go:6785 +0x49
github.com/prometheus/prometheus/vendor/github.com/aws/aws-sdk-go/aws/request.(*Request).EachPage
(0xc49ec08000, 0xc4815cbb60, 0x2, 0x2)
        /go/src/github.com/prometheus/prometheus/vendor/github.com/aws/aws-sdk-go/aws/request/request_pagination.go:98 +0xa1
github.com/prometheus/prometheus/vendor/github.com/aws/aws-sdk-go/service/ec2.(*EC2).DescribeInstancesPages(0xc4821b45e8, 0x0, 0xc4815cbca0, 0x0, 0x0)
        /go/src/github.com/prometheus/prometheus/vendor/github.com/aws/aws-sdk-go/service/ec2/api.go:6784 +0x11a
```

This looks similar to #3001, but this time the the nil pointer was not in EC2 tags, it was in the ENI code: https://github.com/prometheus/prometheus/blob/v1.8.2/discovery/ec2/ec2.go#L190
```
					for _, eni := range inst.NetworkInterfaces {
						subnetsMap[*eni.SubnetId] = struct{}{}
					}
```

[...]

PR #4469, opened 2018-08-07
Handle nil pointer in EC2 discovery

noqcks:
This handles a nil pointer that was being accessed in EC2 discovery https://github.com/prometheus/prometheus/blob/master/discovery/ec2/ec2.go#L265-L270

Fixes: #4441


brian-brazil:
Thanks!

PR #4471, opened 2018-08-07
vendor: update TSDB

fabxc:
@brian-brazil


brian-brazil:
The build failure is real.

fabxc:
Fixed it right in this PR since it's small enough.

brian-brazil:
:+1: once CI is good.

krasi-georgiev:
@fabxc reduced memory usage, didn't expect that.

http://prombench.prometheus.io/grafana/d/7gmLoNDmz/prombench?orgId=1&var-RuleGroup=All&var-pr-number=4471

PR #4473, opened 2018-08-07
discovery: start runUpdater() only once

### LINKED ISSUE #4470: discovery.Manager.runUpdater() only needs to run once
A provider runs runUpdater()，but they do same things, it maybe a bug.

simonpasquier:
Closes #4470

### DIFF
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -90,6 +90,7 @@ type Manager struct {
 
 // Run starts the background processing
 func (m *Manager) Run() error {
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
At first glance it seems we can simplify this by putting the logic of `runProvider` directly in `Run()`

simonpasquier:
Holding off this PR since #4523 touches this part too.

krasi-georgiev:
with #4523 we don't even need this one. Will add in the comment that it also  fixes #4470

PR #4474, opened 2018-08-07
Prometheus Monitoring Mixin for Prometheus itself.

tomwilkie:
Right now just defines a set of alerts that (IMO) everyone running Prometheus should use.

Will be adding:
- [x] documentation on how to use
- [x] dashboard


documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,92 @@
...
+            alert: 'PromScrapeFailed',
+            expr: |||
+              up != 1
+            |||,
+            'for': '15m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: 'Prometheus failed to scrape a target {{ $labels.job }} / {{ $labels.instance }}',
```
brian-brazil: This is not an alert about Prometheus itself

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,92 @@
...
+        name: 'prometheus',
+        rules: [
+          {
+            alert: 'PromScrapeFailed',
+            expr: |||
+              up != 1
+            |||,
+            'for': '15m',
+            labels: {
+              severity: 'warning',
```
brian-brazil: Ticket and page are our standard severities.

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,92 @@
...
+            alert: 'PromAlertsFailed',
+            expr: |||
+              sum(increase(alertmanager_notifications_failed_total{%(alertmanagerSelector)s}[5m])) by (namespace) > 0
+            ||| % $._config,
+            'for': '5m',
+            labels: {
+              severity: 'critical',
+            },
+            annotations: {
+              message: 'Alertmanager failed to send an alert.',
```
brian-brazil: I don't think a single failed notification should result in paging someone, you don't want a page every time pagerduty has a blip.

documentation/prometheus-mixin/alerts.libsonnet
brancz: hmm .. looking at all the sets of alerts out there my feeling is that `warning` / `critical` is more commonly used. I remember we've had discussions and did not really come to an agreement. Maybe these should be configurable and default to one or the other?

brian-brazil: Looking through our existing docs, "page" is what we use. No other values are present.

tomwilkie: There are good arguments in both directions; I'm not super opinionated about this.

But lets not use this PR to discuss this particular issue as it can be divisive; either link to previous discussion where this was decided, start a new discussion on the -dev list, or use the dev summit. 

I'm happy to hold this PR until this is decided.

brian-brazil: I can't seem to find it among my email. As it stands our existing usage is "page" in the docs, so if that's to be changed that would need discussion.

tomwilkie: We discussed at the dev summit and settled on critical and warning:

https://docs.google.com/document/d/1-C5PycocOZEVIPrmM1hn8fBelShqtqiAmFptoG4yK70/edit

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,79 @@
...
+            'for': '15m',
+            labels: {
+              severity: 'critical',
+            },
+            annotations: {
+              mesage: 'Prometheus failed to reload config, see container logs',
+            },
+          },
+          {
+            alert: 'PromAlertmanagerBadConfig',
```
brancz: This should be in the Alertmanager repo, no?

tomwilkie: I think you mentioned @mxinden was working on a mixin for Alertmanager - so I wait for that or JFDI?

tomwilkie:
> If you want you can also already include the ones that we have additionally here

Will do!

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,176 @@
...
+              severity: 'critical',
+            },
+            annotations: {
+              message: 'Prometheus failed to reload config, see container logs',
+            },
+          },
+          {
+            alert: 'PrometheusNotificationQueueRunningFull',
+            expr: |||
+              predict_linear(prometheus_notifications_queue_length{%(prometheusSelector)s}[5m], 60 * 30)
```
brian-brazil: This should be indented by 2

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,176 @@
...
+              >
+              prometheus_notifications_queue_capacity{%(prometheusSelector)s}
```
brian-brazil: And this - plus many below

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,176 @@
...
+            ||| % $._config,
+            'for': '15m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: "Prometheus' alert notification queue is running full for {{$labels.namespace}}/{{ $labels.pod}}",
```
brian-brazil: This doesn't describe the alert

Also, summary and description are what we use.

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,176 @@
...
+            annotations: {
+              message: 'Prometheus {{ $labels.namespace }}/{{ $labels.pod}} is not connected to any Alertmanagers',
+            },
+          },
+          {
+            alert: 'PrometheusTSDBReloadsFailing',
+            expr: |||
+              increase(prometheus_tsdb_reloads_failures_total{%(prometheusSelector)s}[2h]) > 0
+            ||| % $._config,
+            'for': '12h',
```
brian-brazil: This should probably alert faster

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,176 @@
...
+            annotations: {
+              message: '{{$labels.job}} at {{$labels.instance}} had {{$value | humanize}} compaction failures over the last four hours.',
+            },
+          },
+          {
+            alert: 'PrometheusTSDBWALCorruptions',
+            expr: |||
+              tsdb_wal_corruptions_total{%(prometheusSelector)s} > 0
+            ||| % $._config,
+            'for': '4h',
```
brian-brazil: This doesn't really jive with the expression you're using, I'd expect a much shorter for

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,176 @@
...
+            annotations: {
+              message: '{{$labels.namespace}}/{{$labels.pod}} has many samples rejected due to duplicate timestamps but different values',
+            },
+          },
+          {
+            alert: 'PrometheusRemoteStorageFailures',
+            expr: |||
+              (rate(prometheus_remote_storage_failed_samples_total{%(prometheusSelector)s}[1m]) * 100)
+                /
+              (rate(prometheus_remote_storage_failed_samples_total{%(prometheusSelector)s}[1m]) + rate(prometheus_remote_storage_succeeded_samples_total{%(prometheusSelector)s}[1m]))
```
brian-brazil: We should fix this metric

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,176 @@
...
+            annotations: {
+              message: "Prometheus' alert notification queue is running full for {{$labels.namespace}}/{{ $labels.pod}}",
+            },
+          },
+          {
+            alert: 'PrometheusErrorSendingAlerts',
+            expr: |||
+              100 * rate(prometheus_notifications_errors_total{%(prometheusSelector)s}[5m])
+              /
+              rate(prometheus_notifications_sent_total{%(prometheusSelector)s}[5m]) > 1
```
brian-brazil: This makes it seem as though the > 1 is only filtering the rate.

```
    100 * rate(prometheus_notifications_errors_total{%(prometheusSelector)s}[5m])
  /
    rate(prometheus_notifications_sent_total{%(prometheusSelector)s}[5m])
> 
  1
```

Is the standard indentation, which makes this clear.

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,176 @@
...
+              severity: 'warning',
+            },
+            annotations: {
+              message: '{{$labels.namespace}}/{{$labels.pod}} has many samples rejected due to duplicate timestamps but different values',
+            },
+          },
+          {
+            alert: 'PrometheusRemoteStorageFailures',
+            expr: |||
+              (rate(prometheus_remote_storage_failed_samples_total{%(prometheusSelector)s}[1m]) * 100)
```
brian-brazil: I'd use at least 5m rates, the scrape interval might be 1m.

documentation/prometheus-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,96 @@
...
+            'Value #B': { alias: 'Uptime' },
+          })
+        )
+      )
+      .addRow(
+        g.row('Discovery')
+        .addPanel(
+          g.panel('Target Sync') +
+          g.queryPanel('sum(rate(prometheus_target_sync_length_seconds_sum{job=~"$job",instance=~"$instance"}[2m])) by (scrape_job) * 1e3', '{{scrape_job}}') +
+          { yaxes: g.yaxes('ms') }
```
brian-brazil: Why not use seconds?

documentation/prometheus-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,96 @@
...
+        )
+        .addPanel(
+          g.panel('Targets') +
+          g.queryPanel('count(up{})', 'Targets') +
```
brian-brazil: This is not the number of targets the Prometheus has - it's the number in the meta Prometheus

documentation/prometheus-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,96 @@
...
+          g.stack
+        )
+      )
+      .addRow(
+        g.row('Retrieval')
+        .addPanel(
+          g.panel('Target Scrape Duration') +
+          g.queryPanel('1e3 * sum(scrape_duration_seconds) / count(scrape_duration_seconds)', 'Average') +
```
brian-brazil: Ditto

And you could use avg

documentation/prometheus-mixin/alerts.libsonnet
tomwilkie: The jsonnet language doesn't support that; the first line indentation dictates the rest of the block.  So we can't support your style guide I'm afraid.

documentation/prometheus-mixin/alerts.libsonnet
tomwilkie: > summary and description are what we use.

This isn't something we've agreed on - the duplication between summary and description is particularly annoying.  Lets take this to the list.

documentation/prometheus-mixin/dashboards.libsonnet
tomwilkie: The way grafana renders them is sub-optimal - you'll get '0.1s' when '100ms' is more useful.  With `ms` you'll still get '1.3s' instead of '1300ms'.

documentation/prometheus-mixin/alerts.libsonnet
brian-brazil: ||| doesn't support it, jsonnet does if you use any of the other forms of string: https://jsonnet.org/ref/spec.html

documentation/prometheus-mixin/alerts.libsonnet
brian-brazil: It is, there was a big discussion on it a few years back.

documentation/prometheus-mixin/jsonnetfile.json
```diff
@@ -0,0 +1,14 @@
+{
+    "dependencies": [
+        {
+            "name": "grafana-builder",
+            "source": {
+                "git": {
+                    "remote": "https://github.com/kausalco/public",
```
mxinden: Would this be `https://github.com/monitoring-mixins/docs`? //CC @metalmatze

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,192 @@
...
+            },
+            annotations: {
+              message: 'Prometheus remote write is {{ printf "%.1f" $value }}s behind.',
+            },
+          },
+          {
+            alert: 'PrometheusRuleFailures',
+            'for': '15m',
+            expr: |||
+              rate(prometheus_rule_evaluation_failures_total{%(prometheusSelector)s}[1m]) > 0
```
jtlisi:
```suggestion
              rate(prometheus_rule_evaluation_failures_total{%(prometheusSelector)s}[2m]) > 0
```
or make this range configurable because it needs to be larger than the rule eval interval

beorn7:
I'd like to revive this PR. I'll go through all the open comments ASAP and try to address them.

In the meantime, a question for @brancz : You suggested to include the Prometheus metamon alerts from the prometheus operator repo. If I got that right, the ksonnet stuff has moved to https://github.com/coreos/kube-prometheus/blob/master/jsonnet/kube-prometheus/alerts/prometheus.libsonnet . Could you confirm that this is the file you would suggest to merge into the PR here? Are there other files in the repo you would recommend to look at?

brancz:
That's all yes :)

beorn7:
OK, cool. Working on it now…

documentation/prometheus-mixin/alerts.libsonnet
beorn7: This is gone, so I resolve this conversation.

documentation/prometheus-mixin/alerts.libsonnet
beorn7: It's not that easy.

The following is legal jsonnet:
```jsonnet
            expr: '
                predict_linear(prometheus_notifications_queue_length{%(prometheusSelector)s}[5m], 60 * 30)
              >
                prometheus_notifications_queue_capacity{%(prometheusSelector)s}
            ' % $._config,
```

However, if you run `jsonnetfmt` on it, it will transform it into something quite unreadable. But even if you do not run `jsonnetfmt` on it, then the same unreadable form will still end up in the YAML file:
```yaml
    "expr": "\n                predict_linear(prometheus_notifications_queue_length{job=\"prometheus\"}[5m], 60 * 30)\n              >\n                prometheus_notifications_queue_capacity{job=\"prometheus\"}\n            "
```

What we would need is something like YAML's `|2`. But that's missing from jsonnet.

The only way out I could come up with is the following:
```jsonnet
            expr: |||
              # (indentation hint)
                predict_linear(prometheus_notifications_queue_length{%(prometheusSelector)s}[5m], 60 * 30)
              >
                prometheus_notifications_queue_capacity{%(prometheusSelector)s}
            ||| % $._config,
```

If nobody has a better idea, I'll go with that one.

beorn7: Or how about this one?
```jsonnet
            expr: |||
              (
                predict_linear(prometheus_notifications_queue_length{%(prometheusSelector)s}[5m], 60 * 30)
              >
                prometheus_notifications_queue_capacity{%(prometheusSelector)s}
              )
            ||| % $._config,
```

documentation/prometheus-mixin/alerts.libsonnet
beorn7: The outcome of the discussion a few years back was that setups are so different that we don't want to mandate anything and thus made the annotations completely free-form. At SoundCloud, we went into the habit of making `summary` something fairly static so that it would not vary between different alerts ending up in the same notification group, while `description` was tailored to the individual alert, describing in human-readable form what was going on. In that way, you could use `summary` as the headline for a notification group (e.g. mail subject) and then iterate through individual alerts e.g. in the mail text:
```
DiskRunningFull [2 alerts firing]:
- 98% full on foo.
- 95% full on bar.
```

If that's what everybody could live with, I'd apply it here.

documentation/prometheus-mixin/alerts.libsonnet
beorn7: I'll go through all the indentations and resolve this one.

documentation/prometheus-mixin/alerts.libsonnet
beorn7: I went for the parentheses method as it appears the least invasive and most readable. Sad that jsonnet's block quoting is not at par with yaml's one here.
I'll resolve this conversation, but feel free to come up with other ideas.

documentation/prometheus-mixin/alerts.libsonnet
beorn7: I chose a slightly different indentation scheme (the outdented operator on its own line for vector-vector operations, operator with scalar on the same line for vector-scalar operations).

Is the "standard indentation" documented anywhere?

brian-brazil: It's documented by example at https://prometheus.io/docs/practices/rules/

documentation/prometheus-mixin/alerts.libsonnet
beorn7: I went for 4h.

beorn7: Thinking about it, the 2h range for the `increase` could fit into a 2h window of not doing anything (if a compaction is triggered slightly late). How about 3h window for 4h? This should essentially trigger if something goes wrong for two cycles in a row.

What does everyone think about it?

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,210 @@
...
+            annotations: {
+              message: '{{$labels.job}} at {{$labels.instance}} had {{$value | humanize}} reload failures over the last four hours.',
+            },
+          },
+          {
+            alert: 'PrometheusTSDBCompactionsFailing',
+            expr: |||
+              increase(prometheus_tsdb_compactions_failed_total{%(prometheusSelector)s}[2h]) > 0
+            ||| % $._config,
+            'for': '12h',
```
beorn7: I changed it into range of 3h for 4h, similar to above.

documentation/prometheus-mixin/alerts.libsonnet
beorn7: Is there any reason to have a for clause at all if we want to alert on any WAL corruptions having occurred ever?

On the other hand, once the corrupted WAL is obsolete and no corruptions have happened in the current cycle, this alert is not actionable anymore.

How about an increase over a 3h window with a 4h for clause? We'll get alerted if at least two cycles happen consecutively with a WAL corruption, and the alert will stop firing if WALs are created fine again.

documentation/prometheus-mixin/alerts.libsonnet
beorn7: I think 5m is a safe bet. And it doesn't really has to be shorter if we only alert after 15m anyway.

documentation/prometheus-mixin/alerts.libsonnet
brian-brazil: It's always wise to have some for clause, in case of weirdness. I'd use something like 5-10m.

The alert isn't actionable one way or the other.

documentation/prometheus-mixin/dashboards.libsonnet
beorn7: `sum(prometheus_sd_discovered_targets{job=~"$job",instance=~"$instance"})` should do the trick.

documentation/prometheus-mixin/dashboards.libsonnet
beorn7: I changed it to display the average actual scrape interval.

documentation/prometheus-mixin/alerts.libsonnet
beorn7: "Weirdness" doesn't sound like a good reason to have a for clause.

This is a warning. If I regularly get WAL corruptions, I should start to wonder about a bug or a broken disk. I also should avoid restarting the server while the WAL is corrupted.

documentation/prometheus-mixin/alerts.libsonnet
beorn7: Interesting. It also uses the parentheses hack, although Yaml has a better way of handing the indent in the 1st line.

In any case, the examples do not contain vector-to-scalar ops.

documentation/prometheus-mixin/alerts.libsonnet
beorn7: Done.

documentation/prometheus-mixin/jsonnetfile.json
beorn7: I don't think so. Why?

documentation/prometheus-mixin/alerts.libsonnet
beorn7: Yes, we should, but that's outside of the scope of this PR.

beorn7:
I still have to merge this with https://github.com/coreos/kube-prometheus/blob/master/jsonnet/kube-prometheus/alerts/prometheus.libsonnet . And I need to resolve the `message` vs `description`/`summary` controversy.

While doing so, I'll apply some scrutiny to the alerts here in general, but I do not aim to make everything perfect in this PR. I think having mixin here as a "technology demonstration" has a certain value even without perfection. We can iterate on these over time. (They will be used in production at Grafana, so there will be improvements driven by practical usage, pushed back here.)

brancz:
I agree. We didn’t the same with the kubernetes mixin. Of course we won’t catch everything here, but the whole point is that as a community we can all use and improve them over time.

documentation/prometheus-mixin/jsonnetfile.json
lukmdo: maybe @mxinden that is `https://github.com/grafana/jsonnet-libs` ?

beorn7: Yes, that seems to be the updated path. Thanks @lukmdo.

documentation/prometheus-mixin/alerts.libsonnet
beorn7: As the kube-prometheus alert mixins are using summary/description too (with approximately the semantics I have explained above), I go with them for now.

beorn7:
This is now ready for a re-review. @brancz @brian-brazil @tomwilkie 

As said, I'm not aiming for alert-philosophical perfection or completion at this time but more for not doing anything blatantly wrong and creating a reasonable starting point.

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,252 @@
...
+            },
+          },
+          {
+            alert: 'PrometheusNotConnectedToAlertmanagers',
+            expr: |||
+              prometheus_notifications_alertmanagers_discovered{%(prometheusSelector)s} < 1
+            ||| % $._config,
+            'for': '10m',
+            labels: {
+              severity: 'warning',
```
brancz: This could potentially result in not being alerted at all (and also this one is likely to be hard to be sent at all), should this be critical or do we expect users to have dead-mans-switch type alerting setup (if so we should probably at least mention it somewhere)?

documentation/prometheus-mixin/config.libsonnet
```diff
@@ -0,0 +1,15 @@
+{
+  _config+:: {
+    // Selectors are inserted between {} in Prometheus queries.
+    prometheusSelector: 'job="prometheus"',
+    alertmanagerSelector: 'job="alertmanager"',
```
brancz: this is not unused

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,252 @@
+{
+  prometheusAlerts+:: {
+    groups+: [
+      {
+        name: 'prometheus',
+        rules: [
+          {
+            alert: 'PrometheusBadConfig',
+            expr: |||
+              prometheus_config_last_reload_successful{%(prometheusSelector)s} == 0
```
brian-brazil: You should use max_over_time/min_over_time here, otherwise a failed scrape will reset the alert due to staleness. Similarly with any other gauge-based alerts.

documentation/prometheus-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,150 @@
+local g = import 'grafana-builder/grafana.libsonnet';
+
+{
+  dashboards+: {
+    'prometheus.json':
+      g.dashboard('Prometheus')
+      .addMultiTemplate('job', 'prometheus_build_info', 'job')
+      .addMultiTemplate('instance', 'prometheus_build_info', 'instance')
+      # Prometheus is quite commonly configured with honor_labels set to true;
```
brian-brazil: Huh? The Prometheus server is not something to use honor_labels true with.

Pgw, cloudwatch, statsd, graphite and influxdb exporters are the ones I know of.

brancz: Agreed.

documentation/prometheus-mixin/alerts.libsonnet
beorn7: “Production-ready” meta-monitoring should not be done by a Prometheus server monitoring itself. Thus, this alert will fire if the metamon-prometheus detects other Prometheus servers without any discovered Alertmanagers.

Of course, the metamon-prometheus itself still needs to be able to send alerts. That's where a dead-man-switch like setup enters the game. In general, I think the metamon-prometheus is a good location to test the whole alerting chain with a dead-man-switch like setup. But I believe discussing that aspect of meta-monitoring is out of scope of this mixin example.

documentation/prometheus-mixin/config.libsonnet
beorn7: Indeed, removed.

beorn7: And improved the comment.

documentation/prometheus-mixin/alerts.libsonnet
beorn7: Ooof. That's a very good point. All nicely written up in https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 , but which I totally forgot about. This is probably a good opportunity to spread awareness. I'll rework all gauge-based alerts.

documentation/prometheus-mixin/dashboards.libsonnet
beorn7: Yeah, weird. I'll remove it.

beorn7:
All done. Please resume reviewing. :o)

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,260 @@
...
+              )
+              * 100
+              > 1
+            ||| % $._config,
+            'for': '15m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              summary: 'Prometheus encounters more than 1% errors sending alerts to a specific Alertmanager.',
```
brian-brazil: encountered

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,260 @@
...
+            alert: 'PrometheusDuplicateTimestamps',
+            expr: |||
+              rate(prometheus_target_scrapes_sample_duplicate_timestamp_total{%(prometheusSelector)s}[5m]) > 0
+            ||| % $._config,
+            'for': '10m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              summary: 'Prometheus drops samples with duplicate timestamps.',
```
brian-brazil: is dropping

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,260 @@
...
+              )
+              * 100
+              > 1
+            ||| % $._config,
+            'for': '15m',
+            labels: {
+              severity: 'critical',
+            },
+            annotations: {
+              summary: 'Prometheus fails to send samples to remote storage.',
```
brian-brazil: is failing

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,260 @@
...
+            alert: 'PrometheusMissingRuleEvaluations',
+            expr: |||
+              increase(prometheus_rule_group_iterations_missed_total{%(prometheusSelector)s}[5m]) > 0
+            ||| % $._config,
+            'for': '15m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              summary: 'Prometheus misses rule evaluations due to slow rule group evaluation.',
```
brian-brazil: is missing

documentation/prometheus-mixin/alerts.libsonnet
beorn7: Perhaps "has encountered" is even better? (As in "started in the past and is ongoing", if I remember my grammar lessons correctly.)

beorn7:
Adjusted the tenses. (@brian-brazil you didn't comment on any passive-voice issues. Did you just mean tenses, or is there more you intended to comment on?)

documentation/prometheus-mixin/alerts.libsonnet
brian-brazil: That works too

beorn7:
I guess that's a :+1: then. If not, it's just documentation and we can follow up.
Thanks everyone for the feedback.

PR #4479, opened 2018-08-07
change time.Duration to model.Duration for standardization purposes

infoverload:
In the configuration struct in Prometheus, some of the struct fields are of type time.Duration while others are of type model.Duration.  I thought it would be good to standardize the duration types to model.Duration. 

Signed-off-by: Daisy T <daisyts@gmx.com>

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -89,7 +89,7 @@ var (
 	// DefaultAlertmanagerConfig is the default alertmanager configuration.
 	DefaultAlertmanagerConfig = AlertmanagerConfig{
 		Scheme:  "http",
-		Timeout: 10 * time.Second,
+		Timeout: model.Duration(10 * time.Second),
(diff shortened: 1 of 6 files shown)


juliusv:
Looks good, thanks!

@brian-brazil Technically this would be a breaking change since `model.Duration` has slightly different YAML parsing behavior than `time.Duration`, but it's probably fair to consider this a bug that not all duration types were `model.Duration` yet, while the most common ones were already. https://prometheus.io/docs/prometheus/latest/configuration/configuration/ also only defines one `<duration>` type and uses that for all duration fields. Would you agree?

brian-brazil:
Yes, we were meant to switch everything to model.Duration a good while back.

juliusv:
Ok, cool.

:+1: Thanks!

PR #4486, opened 2018-08-10
scrape configs as files

knope:
Additional logic to handle scrape configs as files, as in rule file blobs.
Builds as is, requires testing.

   -- for consideration only, please hold on merge.


brian-brazil:
Thank you for PR. In Prometheus we generally only have one way to do thing, and try to leave configuration management problems to configuration management tools rather than clashing with them. Accordingly you should handle this with templating within your configuration management system, such as Ansible's assemble module.

knope:
meh i just think its funny that for rules, the `separate files` are the `one way` and for scrapes the `all in config.yml` is the `one way` ... doesn't really seem like `one way` to me :P ...

PR #4492, opened 2018-08-13
Clarify that {a="b",a!="c"} is possible.

brian-brazil:

### DIFF
--- a/docs/querying/basics.md
+++ b/docs/querying/basics.md
@@ -95,8 +95,9 @@ For example, this selects all `http_requests_total` time series for `staging`,
 
     http_requests_total{environment=~"staging|testing|development",method!="GET"}
 
-Label matchers that match empty label values also select all time series that do
-not have the specific label set at all. Regex-matches are fully anchored.
+Label matchers that match empty label values also select all time series that
+do not have the specific label set at all. Regex-matches are fully anchored. It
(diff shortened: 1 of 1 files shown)


docs/querying/basics.md
```diff
@@ -95,8 +95,9 @@ For example, this selects all `http_requests_total` time series for `staging`,
 
     http_requests_total{environment=~"staging|testing|development",method!="GET"}
 
-Label matchers that match empty label values also select all time series that do
-not have the specific label set at all. Regex-matches are fully anchored.
+Label matchers that match empty label values also select all time series that
+do not have the specific label set at all. Regex-matches are fully anchored. It
+is possible to have multiple matchers for the same label name.
```
juliusv: This still doesn't clarify that they are AND-ed together vs. OR-ed. I also think the AND property should be explicitly stated somewhere for matchers in general, not just for two matchers for the same label name.

brian-brazil: That's already covered a few paragraphs up.

juliusv: I was thinking of stating it more explicitly rather than just as an example. But that would perhaps imply making this whole document look more like a formal reference rather than a casual explanation. Maybe a task for another day...

:+1: then

Issue #4493, opened 2018-08-13
Breaking files of the main package makes it difficult to test using `go run` 


krasi-georgiev:
## Proposal
**Use case. Why is this important?**
Run and debug Prometheus locally. 
It is important so that we can debug quickly :)

I wish we could organise the files so that the `cmd/prometheus/` contains only `main.go` and `main_test.go` so that I can run and debug only using `go run main.go`.

currently we have so many files in there so that it  is no easy to follow which files need to be passed to the `go run` command.

I guess that mean putting all the other files in their dedicated packages which seems the idiomatic go way anyway.

roidelapluie:
is `go run *.go` such a big deal?

krasi-georgiev:
I wish it was that easy :)
 this picks up the `main_test.go` and fails. I tried many different combinations and nothing worked. 

opening a PR in a sec as this was very easy to fix by moving these in a dedicated package.

PR #4494, opened 2018-08-13
move runtime debug funcs in own package

krasi-georgiev:
fixes: https://github.com/prometheus/prometheus/issues/4493

To make local debuging with `go run` easyer moved all files into a
dedicate package `runtime`.
This allows running prometheus just by using `go run main.go` instead of
passing mani files like `go run main.go limits_default.go ...`

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -39,6 +39,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/common/version"
+	prom_runtime "github.com/prometheus/prometheus/pkg/runtime"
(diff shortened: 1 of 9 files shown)


brancz:
I think this is reasonable :+1:

cmd/prometheus/main.go
```diff
@@ -39,6 +39,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/common/version"
+	promRuntime "github.com/prometheus/prometheus/pkg/runtime"
```
juliusv: So far for multi-word package aliases we have always used under_scores rather than camelCase.

brancz:
lgtm on green

PR #4495, opened 2018-08-13
Add debounce when changing range

biglovisa:
Copy of #3743


gouthamve:
I'd like to test this later today, but just wanted to point out that you added `promu` as a submodule :)

biglovisa:
Oops! that was not intended. I'll clean it up later today. Thanks for taking a look!

PR #4496, opened 2018-08-13
[WIP] Teach label_replace to replace multiple labels at once

sylr:
I'm taking a wild stab in the dark here.

I wanted to do a binary operations on metrics of different sources and to do so I had to rename several labels.

Encapsulating several `label_replace()` calls felt not right so here an attempt to make label_replace prompt to accept several sets of replacements.

Comments are very welcome.

PS: Hide whitespace changes in the diff to make it easier to comprehend

### DIFF
--- a/promql/functions.go
+++ b/promql/functions.go
@@ -711,63 +711,76 @@ func funcChanges(vals []Value, args Expressions, enh *EvalNodeHelper) Vector {
 	return enh.out
 }
 
-// === label_replace(Vector ValueTypeVector, dst_label, replacement, src_labelname, regex ValueTypeString) Vector ===
+// === label_replace(Vector ValueTypeVector, dst_label, replacement, src_labelname, regex ValueTypeString...) Vector ===
 func funcLabelReplace(vals []Value, args Expressions, enh *EvalNodeHelper) Vector {
 	var (
 		vector   = vals[0].(Vector)
-		dst      = args[1].(*StringLiteral).Val
-		repl     = args[2].(*StringLiteral).Val
-		src      = args[3].(*StringLiteral).Val
-		regexStr = args[4].(*StringLiteral).Val
+		dst      string
+		repl     string
+		src      string
+		regexStr string
 	)
 
-	if enh.regex == nil {
-		var err error
-		enh.regex, err = regexp.Compile("^(?:" + regexStr + ")$")
-		if err != nil {
-			panic(fmt.Errorf("invalid regular expression in label_replace(): %s", regexStr))
-		}
-		if !model.LabelNameRE.MatchString(dst) {
-			panic(fmt.Errorf("invalid destination label name in label_replace(): %s", dst))
-		}
-		enh.dmn = make(map[uint64]labels.Labels, len(enh.out))
+	if (len(args)-1)%4 != 0 {
+		panic(fmt.Errorf("invalid number of arguments in label_replace(): %d", len(args)))
 	}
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
Can you explain more about your use case? With this change it's not less complex than multiple calls, and label_replace has a non-trivial invocation already. label_replace should be a last resort, preferring to either fix the data at source or use relabelling.

sylr:
I want to mix metrics coming from cadvisor and kube-state-metrics, unfortunately one use `container_name` and `pod_name` where the other user `container` and `pod`.

Currently
```
avg(
  label_replace(
    label_replace(
      container_memory_max_usage_bytes{namespace=~"$k8snamespace",container_name!="POD"},
      "container",
      "$1",
      "container_name",
      "(.*)"
    ),
    "pod",
    "$1",
    "pod_name",
    "(.*)"
  )
) by (namespace, container)
/
avg(kube_pod_container_resource_requests_memory_bytes{namespace=~"$k8snamespace",container!="POD"}) by (namespace, container) > 1.3
```

Aiming for:

```
avg(
  label_replace(
    container_memory_max_usage_bytes{namespace=~"$k8snamespace",container_name!="POD"},
    "container",
    "$1",
    "container_name",
    "(.*)",
    "pod",
    "$1",
    "pod_name",
    "(.*)"
  )
) by (namespace, container)
/
avg(kube_pod_container_resource_requests_memory_bytes{namespace=~"$k8snamespace",container!="POD"}) by (namespace, container) > 1.3
```

brian-brazil:
The new way is harder to read, and it's not anything that you can't already do with label_replace.

You want metric_relabel_configs here, or waiting for cadvisor to fix their labels (it's in progress, but takes a while due to k8 stability guarantees)

sylr:
No, I don't want metric_relabel_configs, I'm not going to relabel every time metrics have labels who do not match, it would be endless and I'm not willing to sacrifice my historical data.

I'm not gonna argue the harder/easier readability part, it would be pointless as it is totally subjective.

But I've tried not to change the default behaviour nor to add any astonishing complexity to the current implementation.

So why not let the end user choose what suits him best?

@juliusv @beorn7 @brancz @fabxc could we have your opinion on the matter ?

brian-brazil:
> I'm not going to relabel every time metrics have labels who do not match

This is why the recommendation is to fix the source data. label_replace and metric_relabel_configs are only temporary workarounds until that can be done.

> So why not let the end user choose what suits him best?

I see where you're coming from, but having multiple ways to do the same thing ultimately makes the system harder to learn and understand for all users. 

There is already a way to do what you want with PromQL. The way to compose functions is to compose functions, not to make the functions themselves more complex. Our goal is to make things possible, not to exactly meet the preferred way that each individual user would like things to work, as that would not produce an understandable system.

juliusv:
> So why not let the end user choose what suits him best?

This is the eternal question - the answer is: any feature in a project makes the project worse for users that do not need that feature (even if it's great for people who need it), because you have to go through the mental efforts to read about it and understand that you do not need it and file it away. Every extra feature also requires maintenance and can contain bugs.

> could we have your opinion on the matter

I would prefer to keep things as they are as `label_replace()` is already a niche use case, the current invocation is already hard to understand, and doing multiple nested label replaces is even more uncommon. That is not to invalidate your use case, but from an overall project view IMO it's best to keep `label_replace()` as it is.

sylr:
Understood

PR #4499, opened 2018-08-14
marathon-sd - change port gathering strategy, support for container networking

ti-mo:
Hi all

While testing out the Marathon discovery of the current release, I noticed only our services that use host networking would be picked up, all others were just ignored. Taking a look at the source code, it became clear why:

```go
 if len(t.Ports) == 0 {
	continue 
```

I'm currently using Marathon 1.6 (through DC/OS 1.11), and only host-networked services get their `Ports` fields populated. As far as I know, this has been the case since Marathon 1.4; since the design of this discovery plugin dates from way before that (~2 years ago?), I decided to give it a re-write using a different strategy. Instead of requiring a port to appear in `ports`, scan the app for port definitions in the following order on a first-found basis:

- `container.portMappings`
- `container.docker.portMappings`
- `portDefinitions`
- `ports` as a fallback in case there are no `portMappings` or `portDefinitions` present

The reasons I chose for this method:
- `portDefinitions` and `portMappings` cannot co-exist; Marathon will decline the app definition when posted - older versions accept wrong app definitions, but silently re-write them in the background
- `ports` is deprecated, replaced by `portDefinitions`
- `ports` is only used with host networking, where specifying `portMappings` is invalid
- if no mapping is desired (eg. with container networking), `portDefinitions` should be used anyway

I believe that locking in to the first-found ports definition is the best way to preserve backwards-compatibility (and forwards-compatibility!), because it doesn't rely on more than one type of definition being present. There is quite some legacy behaviour to deal with, so the more complexity we can remove from the equation, the better.

Also, while working on this, I noticed there was no support for [container networking](https://github.com/mesosphere/marathon/blob/master/changelog.md#app-json-fields-changed-or-moved) (previously called ip-per-task pre-1.5), so I decided to add that as well. I decided _not_ to support the old ip-per-task (using `ipAddresses.x.networkName` iirc), because it has been omitted from this discovery implementation so far, it didn't make sense to include it.

Other changes:

- removed unnecessary error check on HTTPClientConfig.Validate()
- renamed PortDefinitions and PortMappings to PortDefinition and PortMapping respectively
- extended data model for extra parsed fields from Marathon json
- support container networking on Marathon 1.5+ (target Task.IPAddresses.x.Address)
- expanded test suite to cover all new cases
- test: cancel context when reading from doneCh before returning from function
- test: split test suite into Ports/PortMappings/PortDefinitions

### DIFF
--- a/discovery/marathon/marathon.go
+++ b/discovery/marathon/marathon.go
@@ -107,11 +107,7 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if (len(c.HTTPClientConfig.BearerToken) > 0 || len(c.HTTPClientConfig.BearerTokenFile) > 0) && (len(c.AuthToken) > 0 || len(c.AuthTokenFile) > 0) {
 		return fmt.Errorf("marathon_sd: at most one of bearer_token, bearer_token_file, auth_token & auth_token_file must be configured")
 	}
-	if err := c.HTTPClientConfig.Validate(); err != nil {
-		return err
-	}
-
-	return nil
+	return c.HTTPClientConfig.Validate()
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


brian-brazil:
We don't strictly need to maintain backwards compatibility in SD config wise, such as for when issues like this come to light. If the underlying system has changed, we should change to match it.

ti-mo:
@brian-brazil Alright, good to know! Marathon itself tries to juggle backwards-compatibility, so this code is still correct today regardless, it was written for 1.6 (current) first.. Can't speak for anything pre-1.5, but as an added benefit, this should be (mostly) compatible with earlier versions as well.

ti-mo:
@brian-brazil Any interest in merging this?

discovery/marathon/marathon.go
```diff
@@ -403,46 +421,108 @@ func createTargetGroup(app *App) *targetgroup.Group {
...
-			}
-			// In Marathon 1.5.x the container.docker.portMappings object was moved
-			// to container.portMappings.
-			if i < len(app.Container.PortMappings) {
-				for ln, lv := range app.Container.PortMappings[i].Labels {
-					ln = portMappingLabelPrefix + strutil.SanitizeLabelName(ln)
+
+			// Gather all port labels and set them on the current target.
+			// Skip if there are no Marathon labels set on this port.
+			if len(labels) > 0 {
```
brian-brazil: You don't need this if

discovery/marathon/marathon.go
```diff
@@ -403,46 +421,108 @@ func createTargetGroup(app *App) *targetgroup.Group {
...
+		// to be the same across all tasks. If we haven't gathered any ports by now,
+		// use the task's ports as the port list.
+		if len(ports) == 0 && len(t.Ports) != 0 {
+			ports = t.Ports
 		}
-		for i := 0; i < len(t.Ports); i++ {
-			targetAddress := targetForTask(&t, i)
+
+		// Iterate over the ports we gathered using one of the methods above.
+		for i := 0; i < len(ports); i++ {
```
brian-brazil: Using a range on ports would be more usual here

brian-brazil:
Thanks!

PR #4501, opened 2018-08-14
web/api: Expose rule health and last error

mxinden:
Expose rule health and last evaluation error on `/api/v1/rules`.

//CC @noqcks

Follow up to https://github.com/prometheus/prometheus/pull/4457#pullrequestreview-143589537.

### DIFF
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -406,6 +406,7 @@ $ curl http://localhost:9090/api/v1/rules
                             "summary": "High request latency"
                         },
                         "duration": 600,
+                        "health": "ok",
                         "labels": {
                             "severity": "page"
                         },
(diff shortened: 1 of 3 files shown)


mxinden:
Ping @juliusv @fabxc @brian-brazil

gouthamve:
Actually, just noticed a minor race condition: https://github.com/prometheus/prometheus/blob/master/rules/alerting.go#L368-L369

These should use `r.SetHealth` and `r.SetLastError` else there is a race while accessing the same via the API.

simonpasquier:
@gouthamve in fact not, see https://github.com/prometheus/prometheus/commit/b7e2f407ded76644edc97957c6e8e611a64ea29d

mxinden:
@gouthamve @simonpasquier thanks for the feedback. Any further concerns?

gouthamve:
Nope, this is good to go :)

gouthamve:
I think two reviews from `-team` would be enough. Usually people won't approve the change unless they are pretty confident about it.

Having a maintainer approve every single PR in a repo like `prometheus/prometheus` would be too much to ask, if you ask me.

brian-brazil:
> Usually people won't approve the change unless they are pretty confident about it.

That's not the case for all of -team. In this case I believe you have gotten sufficient review.

mxinden:
Thanks for clarifying!

PR #4507, opened 2018-08-15
discovery/dns: add force_tcp option

rhysm:
Adds a configurable option to dns_sd_config which forces the use of DNS lookups over TCP.

I have been experiencing issues with DNS SD via CoreDNS and an issue relating to the truncation bit. Basically when there are a certain number of records returned via SRV lookup the additional section is truncated but the TC bit is not set on the response and Prometheus doesn't attempt to retry via TCP.

The following errors show in the logs:

```
level=warn ts=2018-08-15T22:38:18.755247522Z caller=dns.go:301 component="discovery manager scrape" discovery=dns msg="DNS resolution failed" server=redacted name=redacted. err="dns: overflow unpacking uint16"

level=warn ts=2018-08-15T22:38:18.86874238Z caller=dns.go:301 component="discovery manager scrape" discovery=dns msg="DNS resolution failed" server=redacted name=redacted. err="dns: overflow unpacking uint16"

level=error ts=2018-08-15T22:38:18.868800607Z caller=dns.go:162 component="discovery manager scrape" discovery=dns msg="Error refreshing DNS targets" err="could not resolve \"redacted\": all servers responded with errors to at least one search domain"
```

There is a PR open to the DNS library for CoreDNS at https://github.com/miekg/dns/pull/708
which I believe will fix the underlying issue if/when it is merged.

This change would be useful in order to workaround the issue on the Prometheus side.

### DIFF
--- a/discovery/dns/dns.go
+++ b/discovery/dns/dns.go
@@ -56,6 +56,7 @@ var (
 	DefaultSDConfig = SDConfig{
 		RefreshInterval: model.Duration(30 * time.Second),
 		Type:            "SRV",
+		ForceTCP:        false,
 	}
 )
 
@@ -65,6 +66,7 @@ type SDConfig struct {
 	RefreshInterval model.Duration `yaml:"refresh_interval,omitempty"`
 	Type            string         `yaml:"type"`
 	Port            int            `yaml:"port"` // Ignored for SRV records
+	ForceTCP        bool           `yaml:"force_tcp"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
@@ -104,6 +106,7 @@ type Discovery struct {
 	port     int
 	qtype    uint16
 	logger   log.Logger
+	forceTcp bool
 }
 
 // NewDiscovery returns a new Discovery which periodically refreshes its targets.
@@ -127,6 +130,7 @@ func NewDiscovery(conf SDConfig, logger log.Logger) *Discovery {
 		qtype:    qtype,
 		port:     conf.Port,
 		logger:   logger,
+		forceTcp: conf.ForceTCP,
 	}
 }
 
@@ -165,7 +169,7 @@ func (d *Discovery) refreshAll(ctx context.Context, ch chan<- []*targetgroup.Gro
 }
 
 func (d *Discovery) refresh(ctx context.Context, name string, ch chan<- []*targetgroup.Group) error {
-	response, err := lookupWithSearchPath(name, d.qtype, d.logger)
+	response, err := lookupWithSearchPath(name, d.qtype, d.logger, d.forceTcp)
 	dnsSDLookupsCount.Inc()
 	if err != nil {
 		dnsSDLookupFailuresCount.Inc()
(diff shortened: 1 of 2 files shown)


rhysm:
Travis CI failed but tests pass locally and on CircleCI.

grobie:
The additional section can be removed without the need to set the truncation bit. In fact, the DNS spec mandates to not set truncation bit in that case.

I feel unsure about introducing a new config option just to work around an issue between CoreDNS and Prometheus.

What version of CoreDNS and Prometheus are you using?

@miekg do you have any idea what's going on here?

brian-brazil:
It sounds like we shouldn't be relying on the additional section at all here.

miekg:
Tobias is right here. CoreDNS returns a valid packet even if it has been
"truncated", because otherwise alpine Linux can't deal with it.

If you expect the additional section to be complete, your going to have a
bad time. You need to fetch that data yourself then. Assuming we're taking
resolving srv records, that going to be a lot a extra queries to get all
the addresses.

rhysm:
I've made a repo at https://github.com/rhysm/prometheus-dns-sd-reproduce which reproduces a similar issue. I couldn't get the uint16 overflow, I get "dns: overflowing header size" in Prometheus so I assume it's the same root cause.

The test to reproduce adds 50 records, 49 and 51 records work fine and Prometheus has no errors. It seems to be when the response is slightly over the 4096 byte mark, I can see in Wireshark the problematic response is 4105 bytes.

Have also put hex dump of DNS packet into repo at https://github.com/rhysm/prometheus-dns-sd-reproduce/blob/master/problem.hex

Instructions to reproduce are in the repo.

miekg:
@rhysm "I can see in Wireshark the problematic response is 4105 bytes.", this means you get a larger buffer even though you set edns to 4096? Can you repro without prometheus? I.e. just coredns? If so we are doing something stupid

miekg:
[ Quoting <notifications@github.com> in "Re: [prometheus/prometheus] discove..." ]
>> @rhysm "I can see in Wireshark the problematic response is 4105 bytes.", this means you get a larger buffer even though you set edns to 4096? Can you repro without prometheus? I.e. just coredns? If so we are doing something stupid
>
>@miekg when I use dig against CoreDNS there is no error (see output below). The ANSWER and ADDITIONAL count at the top are both 50 however the actual number of ADDITIONAL displayed in the dig output is 49. Response size also shows 4105.

yes, the OPT RR is the extra one here.

>
>When I use nslookup against CoreDNS (output below) it retries in TCP but looking at Wireshark it seems to only get a 500 byte UDP response so I assume it requests a smaller buffer by default?

yes, and it does more things automatically.

--
Miek Gieben

codesome:
@rhysm this is not needed with correct DNS implementation and is out of scope Prometheus. We discussed this in our bug scrub and decided to close this.

nemobis:
> There is a PR open to the DNS library for CoreDNS at miekg/dns#708
> which I believe will fix the underlying issue if/when it is merged.

Do do we know if this was fixed upstream in the end?

https://github.com/miekg/dns/pull/708#issuecomment-451381134 :

> will be done differently.

but no issue or alternative PR tracking it AFAIK.

roidelapluie:
https://github.com/miekg/dns/pull/854

PR #4508, opened 2018-08-16
Wait for all internal discoveries are done before exiting.

cofyc:


discovery/kubernetes/kubernetes.go
```diff
@@ -370,6 +370,8 @@ func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
 	}
 
 	d.Unlock()
+
+	wg.Wait()
 	<-ctx.Done()
```
krasi-georgiev: shouldn't this we fine even without `<-ctx.Done()` ?
`wg.Wait()` will block

krasi-georgiev:
LGTM

gouthamve:
LGTM

Issue #4414, opened 2018-07-25
Limit number of data points returned in a query


tomwilkie:
We currently limit the number of steps to 11k, but if you query a very high cardinality metric we seem to be able OOM even a very large Prometheus.  I propose we limit the number of points we return (step * timeseries).  

Plumbing this through might be a little tricky though, I though @gouthamve had some ideas?

brian-brazil:
The plan is to limit the number of points that a PromQL node can return. Limiting the ultimate output could be too late.

tomwilkie:
That sounds reasonable.  Is this discussed anywhere?

brian-brazil:
It came up on the developers list a while back.

tomwilkie:
Found it: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/prometheus-developers/bwcBD-JBXck/BiFO_eKbCgAJ

> I'm planning on adding a returned sample limit for each node in a promql evaluation, which we can do with the new promql design and would be good enough for most purposes.

When you said "the plan" you meant "my plan", right?  :-)

Is this something you have started working on or plan on working on soon?  If not I'll take a stab.

brian-brazil:
I've not started working on it yet, there's also #4384 which is taking a more complicated approach.

tomwilkie:
Roger, thanks for the pointer.

Do you have any thoughts on reasonable limits?  IIRC we consider 100k to be a reasonable cardinality limit for a metric, and therefore each aggregation should be able to consider 100k samples for a given stpe.  I guess a limit would also be appropriate on the output of aggregations (specifically, less than 100k - maybe 1k?) and on the number of samples in a vector selector (maybe a relatively small limit - 6k would be 1 day at 15s).

Even after we do all this, it still sounds like queries that returns lots of timeseries over a long range could be a problem, if they are very simple - I'm thinking something as basic as `up` over 1k targets and 11k datapoints.

Hence why I think we need to limit `step * #timeseries`.  WDYT?

brian-brazil:
My main thought was that `sum by job({__name__=~".+"})` needs to keep working in the vast majority of setups, as we don't currently have anything else to help debug cardinality issues. That'd imply at least 20M which works out as ~1GB in the worst case (3 nodes worth of data can be in memory at once for a binary operator, at 16B per sample).

> Hence why I think we need to limit step * #timeseries.

That's why I'm proposing limiting output samples of a node, as it's more accurate than that.

[...]

PR #4513, opened 2018-08-16
WIP: keep track of samples per query, set a max # of samples

cstyan:
For #4414 

Mostly looking for feedback on whether or not I'm on the right track here, naive solution for now. 
@brian-brazil @tomwilkie 

Signed-off-by: Callum Stytan <callumstyan@gmail.com>

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -96,6 +96,7 @@ func main() {
 		webTimeout          model.Duration
 		queryTimeout        model.Duration
 		queryConcurrency    int
+		queryMaxSamples     int
 		RemoteFlushDeadline model.Duration
 
 		prometheusURL string
@@ -197,6 +198,8 @@ func main() {
 
 	a.Flag("query.max-concurrency", "Maximum number of queries executed concurrently.").
 		Default("20").IntVar(&cfg.queryConcurrency)
+	a.Flag("query.max-samples", "Maximum number of samples a single query can load into memory. Note that queries will fail if they would load more samples than this into memory, so this also limits the number of samples a query can return.").
+		Default("50000000").IntVar(&cfg.queryMaxSamples)
 
 	promlogflag.AddFlags(a, &cfg.logLevel)
 
@@ -264,12 +267,14 @@ func main() {
 
 		scrapeManager = scrape.NewManager(log.With(logger, "component", "scrape manager"), fanoutStorage)
 
-		queryEngine = promql.NewEngine(
-			log.With(logger, "component", "query engine"),
-			prometheus.DefaultRegisterer,
-			cfg.queryConcurrency,
-			time.Duration(cfg.queryTimeout),
-		)
+		opts = promql.EngineOpts{
+			Logger:        log.With(logger, "component", "query engine"),
+			Reg:           prometheus.DefaultRegisterer,
+			MaxConcurrent: cfg.queryConcurrency,
+			MaxSamples:    cfg.queryMaxSamples,
+			Timeout:       time.Duration(cfg.queryTimeout),
+		}
+		queryEngine = promql.NewEngine(opts)
 
 		ruleManager = rules.NewManager(&rules.ManagerOptions{
 			Appendable:      fanoutStorage,
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -702,14 +702,11 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 	numSteps := int((ev.endTimestamp-ev.startTimestamp)/ev.interval) + 1
 	matrixes := make([]Matrix, len(exprs))
 	origMatrixes := make([]Matrix, len(exprs))
-<<<<<<< HEAD
-	tempNumSamples := ev.currentSamples
-=======
+
 	originalNumSamples := ev.currentSamples
 
 	var tempNumSamples int
 
->>>>>>> c6453509... asdf
 	for i, e := range exprs {
 		// Functions will take string arguments from the expressions, not the values.
 		if e != nil && e.Type() != ValueTypeString {
@@ -780,7 +777,7 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 				s.Point.T = ts
 				mat[i] = Series{Metric: s.Metric, Points: []Point{s.Point}}
 			}
-			ev.currentSamples = tempNumSamples + mat.TotalSamples()
+			ev.currentSamples = originalNumSamples + mat.TotalSamples()
 			return mat
 		}
 
@@ -811,7 +808,7 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 	for _, ss := range seriess {
 		mat = append(mat, ss)
 	}
-	ev.currentSamples = tempNumSamples + mat.TotalSamples()
+	ev.currentSamples = originalNumSamples + mat.TotalSamples()
 	return mat
 }
 
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -733,9 +733,10 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 	}
 	enh := &EvalNodeHelper{out: make(Vector, 0, biggestLen)}
 	seriess := make(map[uint64]Series, biggestLen) // Output series by series hash.
+	tempNumSamples = ev.currentSamples
 	for ts := ev.startTimestamp; ts <= ev.endTimestamp; ts += ev.interval {
 		// Reset num. of samples in memory after each timestamp.
-		tempNumSamples = ev.currentSamples
+		ev.currentSamples = tempNumSamples
 		// Gather input vectors for this timestamp.
 		for i := range exprs {
 			vectors[i] = vectors[i][:0]
@@ -795,7 +796,6 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 			seriess[h] = ss
 
 		}
-		ev.currentSamples = tempNumSamples
 	}
 	// Reuse the original point slices.
 	for _, m := range origMatrixes {
@@ -1009,7 +1009,6 @@ func (ev *evaluator) eval(expr Expr) Value {
 		}
 
 	case *NumberLiteral:
-
 		return ev.rangeEval(func(v []Value, enh *EvalNodeHelper) Vector {
 			return append(enh.out, Sample{Point: Point{V: e.Val}})
 		})
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -43,6 +43,7 @@ const (
 	namespace = "prometheus"
 	subsystem = "engine"
 	queryTag  = "query"
+	env       = "query execution"
 
 	// The largest SampleValue that can be converted to an int64 without overflow.
 	maxInt64 = 9223372036854774784
@@ -69,13 +70,22 @@ type (
 	ErrQueryTimeout string
 	// ErrQueryCanceled is returned if a query was canceled during processing.
 	ErrQueryCanceled string
+	// ErrTooManySamples is returned if a query would woud load more than the maximum allowed samples into memory.
+	ErrTooManySamples string
 	// ErrStorage is returned if an error was encountered in the storage layer
 	// during query handling.
 	ErrStorage error
 )
 
-func (e ErrQueryTimeout) Error() string  { return fmt.Sprintf("query timed out in %s", string(e)) }
-func (e ErrQueryCanceled) Error() string { return fmt.Sprintf("query was canceled in %s", string(e)) }
+func (e ErrQueryTimeout) Error() string {
+	return fmt.Sprintf("query timed out in %s", string(e))
+}
+func (e ErrQueryCanceled) Error() string {
+	return fmt.Sprintf("query was canceled in %s", string(e))
+}
+func (e ErrTooManySamples) Error() string {
+	return fmt.Sprintf("query processing would load too many samples into memory in %s", string(e))
+}
 
 // A Query is derived from an a raw query string and can be run against an engine
 // it is associated with.
@@ -166,19 +176,29 @@ func contextErr(err error, env string) error {
 	}
 }
 
+// EngineOpts contains configuration options used when creating a new Engine.
+type EngineOpts struct {
+	Logger        log.Logger
+	Reg           prometheus.Registerer
+	MaxConcurrent int
+	MaxSamples    int
+	Timeout       time.Duration
+}
+
 // Engine handles the lifetime of queries from beginning to end.
 // It is connected to a querier.
 type Engine struct {
-	logger  log.Logger
-	metrics *engineMetrics
-	timeout time.Duration
-	gate    *gate.Gate
+	logger             log.Logger
+	metrics            *engineMetrics
+	timeout            time.Duration
+	gate               *gate.Gate
+	maxSamplesPerQuery int
 }
 
 // NewEngine returns a new engine.
-func NewEngine(logger log.Logger, reg prometheus.Registerer, maxConcurrent int, timeout time.Duration) *Engine {
-	if logger == nil {
-		logger = log.NewNopLogger()
+func NewEngine(opts EngineOpts) *Engine {
+	if opts.Logger == nil {
+		opts.Logger = log.NewNopLogger()
 	}
 
 	metrics := &engineMetrics{
@@ -223,10 +243,10 @@ func NewEngine(logger log.Logger, reg prometheus.Registerer, maxConcurrent int, 
 			ConstLabels: prometheus.Labels{"slice": "result_sort"},
 		}),
 	}
-	metrics.maxConcurrentQueries.Set(float64(maxConcurrent))
+	metrics.maxConcurrentQueries.Set(float64(opts.MaxConcurrent))
 
-	if reg != nil {
-		reg.MustRegister(
+	if opts.Reg != nil {
+		opts.Reg.MustRegister(
 			metrics.currentQueries,
 			metrics.maxConcurrentQueries,
 			metrics.queryQueueTime,
@@ -236,10 +256,11 @@ func NewEngine(logger log.Logger, reg prometheus.Registerer, maxConcurrent int, 
 		)
 	}
 	return &Engine{
-		gate:    gate.New(maxConcurrent),
-		timeout: timeout,
-		logger:  logger,
-		metrics: metrics,
+		gate:               gate.New(opts.MaxConcurrent),
+		timeout:            opts.Timeout,
+		logger:             opts.Logger,
+		metrics:            metrics,
+		maxSamplesPerQuery: opts.MaxSamples,
 	}
 }
 
@@ -331,9 +352,8 @@ func (ng *Engine) exec(ctx context.Context, q *query) (Value, error) {
 	defer q.cancel()
 
 	const env = "query execution"
-
-	evalSpanTimer, ctx := q.stats.GetSpanTimer(ctx, stats.EvalTotalTime)
-	defer evalSpanTimer.Finish()
+	evalTimer := q.stats.GetTimer(stats.EvalTotalTime).Start()
+	defer evalTimer.Stop()
 
 	// The base context might already be canceled on the first iteration (e.g. during shutdown).
 	if err := contextDone(ctx, env); err != nil {
@@ -384,6 +404,7 @@ func (ng *Engine) execEvalStmt(ctx context.Context, query *query, s *EvalStmt) (
 			endTimestamp:   start,
 			interval:       1,
 			ctx:            ctx,
+			maxSamples:     ng.maxSamplesPerQuery,
 			logger:         ng.logger,
 		}
 		val, err := evaluator.Eval(s.Expr)
@@ -424,6 +445,7 @@ func (ng *Engine) execEvalStmt(ctx context.Context, query *query, s *EvalStmt) (
 		endTimestamp:   timeMilliseconds(s.End),
 		interval:       durationMilliseconds(s.Interval),
 		ctx:            ctx,
+		maxSamples:     ng.maxSamplesPerQuery,
 		logger:         ng.logger,
 	}
 	val, err := evaluator.Eval(s.Expr)
@@ -575,11 +597,12 @@ type evaluator struct {
 	ctx context.Context
 
 	startTimestamp int64 // Start time in milliseconds.
+	endTimestamp   int64 // End time in milliseconds.
+	interval       int64 // Interval in milliseconds.
 
-	endTimestamp int64 // End time in milliseconds.
-	interval     int64 // Interval in milliseconds.
-
-	logger log.Logger
+	maxSamples     int
+	currentSamples int
+	logger         log.Logger
 }
 
 // errorf causes a panic with the input formatted into an error.
@@ -673,15 +696,24 @@ func (enh *EvalNodeHelper) signatureFunc(on bool, names ...string) func(labels.L
 
 // rangeEval evaluates the given expressions, and then for each step calls
 // the given function with the values computed for each expression at that
-// step.  The return value is the combination into time series of  of all the
+// step.  The return value is the combination into time series of all the
 // function call results.
 func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ...Expr) Matrix {
 	numSteps := int((ev.endTimestamp-ev.startTimestamp)/ev.interval) + 1
 	matrixes := make([]Matrix, len(exprs))
 	origMatrixes := make([]Matrix, len(exprs))
+<<<<<<< HEAD
+	tempNumSamples := ev.currentSamples
+=======
+	originalNumSamples := ev.currentSamples
+
+	var tempNumSamples int
+
+>>>>>>> c6453509... asdf
 	for i, e := range exprs {
 		// Functions will take string arguments from the expressions, not the values.
 		if e != nil && e.Type() != ValueTypeString {
+			// ev.currentSamples will be updated to the correct value within the ev.eval call.
 			matrixes[i] = ev.eval(e).(Matrix)
 
 			// Keep a copy of the original point slices so that they
@@ -705,16 +737,21 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 	enh := &EvalNodeHelper{out: make(Vector, 0, biggestLen)}
 	seriess := make(map[uint64]Series, biggestLen) // Output series by series hash.
 	for ts := ev.startTimestamp; ts <= ev.endTimestamp; ts += ev.interval {
+		// Reset num. of samples in memory after each timestamp.
+		tempNumSamples = ev.currentSamples
 		// Gather input vectors for this timestamp.
 		for i := range exprs {
 			vectors[i] = vectors[i][:0]
 			for si, series := range matrixes[i] {
 				for _, point := range series.Points {
-					if point.T == ts {
+					if point.T == ts && ev.currentSamples < ev.maxSamples {
 						vectors[i] = append(vectors[i], Sample{Metric: series.Metric, Point: point})
 						// Move input vectors forward so we don't have to re-scan the same
 						// past points at the next step.
 						matrixes[i][si].Points = series.Points[1:]
+						ev.currentSamples++
+					} else if ev.currentSamples >= ev.maxSamples {
+						ev.error(ErrTooManySamples(env))
 					}
 					break
 				}
@@ -728,6 +765,14 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 			ev.errorf("vector cannot contain metrics with the same labelset")
 		}
 		enh.out = result[:0] // Reuse result vector.
+
+		// In some cases we only have the result of 'f' without having tracked any samples
+		// in the above for loop that built the input vectors.
+		ev.currentSamples += len(result)
+		if ev.currentSamples > ev.maxSamples {
+			ev.error(ErrTooManySamples(env))
+		}
+
 		// If this could be an instant query, shortcut so as not to change sort order.
 		if ev.endTimestamp == ev.startTimestamp {
 			mat := make(Matrix, len(result))
@@ -735,9 +780,10 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 				s.Point.T = ts
 				mat[i] = Series{Metric: s.Metric, Points: []Point{s.Point}}
 			}
+			ev.currentSamples = tempNumSamples + mat.TotalSamples()
 			return mat
 		}
-		// Add samples in output vector to output series.
+
 		for _, sample := range result {
 			h := sample.Metric.Hash()
 			ss, ok := seriess[h]
@@ -750,7 +796,9 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 			sample.Point.T = ts
 			ss.Points = append(ss.Points, sample.Point)
 			seriess[h] = ss
+
 		}
+		ev.currentSamples = tempNumSamples
 	}
 	// Reuse the original point slices.
 	for _, m := range origMatrixes {
@@ -758,11 +806,12 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 			putPointSlice(s.Points)
 		}
 	}
-	// Assemble the output matrix.
+	// Assemble the output matrix. By the time we get here we know we don't have too many samples.
 	mat := make(Matrix, 0, len(seriess))
 	for _, ss := range seriess {
 		mat = append(mat, ss)
 	}
+	ev.currentSamples = tempNumSamples + mat.TotalSamples()
 	return mat
 }
 
@@ -802,6 +851,7 @@ func (ev *evaluator) eval(expr Expr) Value {
 				})
 			}
 		}
+
 		// Check if the function has a matrix argument.
 		var matrixArgIndex int
 		var matrixArg bool
@@ -828,6 +878,7 @@ func (ev *evaluator) eval(expr Expr) Value {
 			if i != matrixArgIndex {
 				otherArgs[i] = ev.eval(e).(Matrix)
 				otherInArgs[i] = Vector{Sample{}}
+				// (callum) this is just building args for later procesing so I don't think we need to track samples here
 				inArgs[i] = otherInArgs[i]
 			}
 		}
@@ -886,8 +937,14 @@ func (ev *evaluator) eval(expr Expr) Value {
 				// Only buffer stepRange milliseconds from the second step on.
 				it.ReduceDelta(stepRange)
 			}
+			// (callum) I think we still need to track number of samples here
 			if len(ss.Points) > 0 {
-				mat = append(mat, ss)
+				if ev.currentSamples < ev.maxSamples {
+					mat = append(mat, ss)
+					ev.currentSamples += len(ss.Points)
+				} else {
+					ev.error(ErrTooManySamples(env))
+				}
 			}
 		}
 		if mat.ContainsSameLabelset() {
@@ -907,6 +964,7 @@ func (ev *evaluator) eval(expr Expr) Value {
 				mat[i].Metric = dropMetricName(mat[i].Metric)
 				for j := range mat[i].Points {
 					mat[i].Points[j].V = -mat[i].Points[j].V
+					// (callum) shouldn't need to track number of samples here, we're just modifiying the results from ev.eval a few lines above
 				}
 			}
 			if mat.ContainsSameLabelset() {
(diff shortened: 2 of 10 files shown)


promql/engine.go
```diff
@@ -743,9 +747,14 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 					Points: getPointSlice(numSteps),
 				}
 			}
-			sample.Point.T = ts
-			ss.Points = append(ss.Points, sample.Point)
-			seriess[h] = ss
+			if curSamples < maxSamples {
```
cstyan: it seems in some cases that there's the potential for _no_ samples to be added in the above block of for loops, and so they're then added later down here, so we have to track curSamples here as well

promql/engine.go
```diff
@@ -382,7 +384,7 @@ func (ng *Engine) execEvalStmt(ctx context.Context, query *query, s *EvalStmt) (
 			ctx:            ctx,
 			logger:         ng.logger,
 		}
-		val, err := evaluator.Eval(s.Expr)
+		val, err := evaluator.Eval(s.Expr, ng.maxSamplesPerQuery)
```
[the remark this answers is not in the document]
cstyan: Agreed, it's already on my "todo list". I was mostly just exploring how the code was currently working to start.

promql/engine.go
```diff
@@ -1116,14 +1149,20 @@ func (ev *evaluator) matrixIterSlice(it *storage.BufferedSeriesIterator, mint, m
 		}
 		// Values in the buffer are guaranteed to be smaller than maxt.
 		if t >= mint {
-			out = append(out, Point{T: t, V: v})
+			if curSamples < maxSamples {
```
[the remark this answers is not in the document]
cstyan: Failing the query entirely, and returning nothing?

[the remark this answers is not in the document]
brian-brazil: Yes, same as a timeout.

promql/engine.go
cstyan: Okay, so the output is the correct place to do it then? Or should we limit the input as well? The goal is to reduce the chance of OOM by limit the # of samples returned by a query, in this case are we not doubling up on memory allocation with the `vectors` for the input to the functions `f`, and then building the output?

brian-brazil: You could view it a few ways. It probably doesn't hurt to limit both.

promql/engine.go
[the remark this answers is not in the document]
cstyan: interesting

promql/engine.go
cstyan: So in some cases, the input (or rather, output of `f`) is the only input that we consider when building the output vector. We don't want every function passed to `rangeEval` to have to deal with limiting the # of samples right?

brian-brazil: Yes, only matrix functions and raw lookups should need special handling beyond that

promql/engine.go
```diff
@@ -164,14 +164,15 @@ func contextDone(ctx context.Context, env string) error {
...
+	logger             log.Logger
+	metrics            *engineMetrics
+	timeout            time.Duration
+	gate               *queryGate
+	maxSamplesPerQuery int
 }
 
 // NewEngine returns a new engine.
-func NewEngine(logger log.Logger, reg prometheus.Registerer, maxConcurrent int, timeout time.Duration) *Engine {
+func NewEngine(logger log.Logger, reg prometheus.Registerer, maxConcurrent int, maxSamples int, timeout time.Duration) *Engine {
```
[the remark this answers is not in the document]
tomwilkie: Yes please!

promql/engine.go
```diff
@@ -574,11 +578,11 @@ type evaluator struct {
 	ctx context.Context
 
 	startTimestamp int64 // Start time in milliseconds.
+	endTimestamp   int64 // End time in milliseconds.
+	interval       int64 // Interval in milliseconds.
```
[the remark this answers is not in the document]
cstyan: yeah, wish the diff output was smarter here :+1:

cmd/prometheus/main.go
```diff
@@ -186,6 +187,8 @@ func main() {
 
 	a.Flag("query.max-concurrency", "Maximum number of queries executed concurrently.").
 		Default("20").IntVar(&cfg.queryConcurrency)
+	a.Flag("query.max-samples", "Maximum number of samples a single query can return. Queries will fail if they would return more than this number.").
```
brian-brazil: This message isn't right, it's about samples in memory rather than returned.

Could we express it in bytes? It's a lower bound due to things like labelsets and capacity vs len, but we could improve it over time.

cmd/prometheus/main.go
```diff
@@ -186,6 +187,8 @@ func main() {
...
+		Default("1000000").IntVar(&cfg.queryMaxSamples)
```
brian-brazil: This should be at least 20M

promql/engine.go
```diff
@@ -67,13 +68,22 @@ type (
 	ErrQueryTimeout string
 	// ErrQueryCanceled is returned if a query was canceled during processing.
 	ErrQueryCanceled string
+	// ErrTooManySamples is returned if a query would return more than the maximum allowed samples.
```
brian-brazil: Samples in memory, not returned

promql/engine.go
```diff
@@ -668,16 +689,24 @@ func (enh *EvalNodeHelper) signatureFunc(on bool, names ...string) func(labels.L
...
-// step.  The return value is the combination into time series of  of all the
+// step.  The return value is the combination into time series of all the
 // function call results.
-func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ...Expr) Matrix {
+func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ...Expr) (Matrix, error) {
 	numSteps := int((ev.endTimestamp-ev.startTimestamp)/ev.interval) + 1
 	matrixes := make([]Matrix, len(exprs))
 	origMatrixes := make([]Matrix, len(exprs))
+	// Limit both input and output samples, in some cases we allocate input vectors.
+	// We'll want to know that the input is too large and fail early rather than also
```
brian-brazil: Hmm, shouldn't we already have failed while preparing the inputs?

promql/engine.go
```diff
@@ -705,11 +734,12 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 			vectors[i] = vectors[i][:0]
 			for si, series := range matrixes[i] {
 				for _, point := range series.Points {
-					if point.T == ts {
+					if point.T == ts && inputSamples < ev.maxSamples {
 						vectors[i] = append(vectors[i], Sample{Metric: series.Metric, Point: point})
 						// Move input vectors forward so we don't have to re-scan the same
 						// past points at the next step.
 						matrixes[i][si].Points = series.Points[1:]
+						inputSamples++
```
brian-brazil: That's not right, the input samples are the entire of the input matrix - not jut the bits we've used so far.

promql/engine.go
```diff
@@ -818,14 +861,19 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+				}
+				otherArgs[i] = ret.(Matrix)
 				otherInArgs[i] = Vector{Sample{}}
 				inArgs[i] = otherInArgs[i]
 			}
 		}
 
 		sel := e.Args[matrixArgIndex].(*MatrixSelector)
-		mat := make(Matrix, 0, len(sel.series)) // Output matrix.
+		mat := make(Matrix, 0, ev.maxSamples) // Output matrix.
```
brian-brazil: This would make the matrix way too large, the existing capacity was fine.

promql/engine.go
```diff
@@ -1112,14 +1190,20 @@ func (ev *evaluator) matrixIterSlice(it *storage.BufferedSeriesIterator, mint, m
 		}
 		// Values in the buffer are guaranteed to be smaller than maxt.
 		if t >= mint {
-			out = append(out, Point{T: t, V: v})
+			if curSamples < ev.maxSamples {
```
brian-brazil: This should be erroring

promql/engine.go
cstyan: I'm not sure what you mean, because we already have `matrixes` in memory? We're only appending the samples we end up using to `vectors`, which is then used as `args` to `f`,

brian-brazil: Yes, they're already in memory.

cstyan: So then if the length of all the `Matrix` in `matrixes` is `> maxSamples` should we not just fail right away? Or only if `inputSamples + len(matrixes[i]) > maxSamples`?

brian-brazil: We should have already failed when one of them was being calculated.

cstyan: Right, so if `inputSamples + len matrixes[i] > maxSamples` we should fail, that should happen on line 707

brian-brazil: Yeah around, there. It'd be inside the eval.

promql/engine.go
cstyan: fixed

promql/engine.go
cstyan: Well we're already tracking if each call to eval would load more than `maxSamples`, but not if the result of all the calls in rangeEval to eval would.

cstyan: So it seems if we track the # of samples that way, based on the outputs of the eval's, we don't need to track samples where I am here at all. That, in combination with not tracking samples when adding to the output series (https://github.com/prometheus/prometheus/pull/4513/commits/29e1469c81a1476a51c7b5f7ede128bb61ac7005), are we missing any cases? The engine tests I added would all still pass.

promql/engine.go
```diff
@@ -668,16 +689,21 @@ func (enh *EvalNodeHelper) signatureFunc(on bool, names ...string) func(labels.L
 
 // rangeEval evaluates the given expressions, and then for each step calls
 // the given function with the values computed for each expression at that
-// step.  The return value is the combination into time series of  of all the
+// step.  The return value is the combination into time series of all the
 // function call results.
-func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ...Expr) Matrix {
+func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ...Expr) (Matrix, error) {
```
brian-brazil: You shouldn't need to change this signature, use ev.error

cstyan: we want to panic and recover all error cases? that seems strange to me

brian-brazil: Yes, that's how error handling is done in PromQL.

promql/engine.go
```diff
@@ -818,14 +853,19 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+				}
+				otherArgs[i] = ret.(Matrix)
 				otherInArgs[i] = Vector{Sample{}}
 				inArgs[i] = otherInArgs[i]
 			}
 		}
 
 		sel := e.Args[matrixArgIndex].(*MatrixSelector)
 		mat := make(Matrix, 0, len(sel.series)) // Output matrix.
+		curSamples := 0
```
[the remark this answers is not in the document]
cstyan: are you referring to this? https://github.com/prometheus/prometheus/blob/master/promql/engine.go#L819-L825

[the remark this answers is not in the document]
brian-brazil: Yes

promql/engine.go
```diff
@@ -720,14 +747,19 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 		enh.ts = ts
 		result := f(args, enh)
 		enh.out = result[:0] // Reuse result vector.
+
+		if len(result) > ev.maxSamples {
```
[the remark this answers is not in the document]
brian-brazil: Yes, both need to be considered.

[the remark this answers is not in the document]
brian-brazil: Yes, both are present. If there's binary operators, there may also be input vectors from the LHS of binary operators up the call stack.

cstyan:
sorry for the delay, I think I'm covering all the cases now but I'll have another look over my PR plus the other PR, I also want to simplify the sample tracking by keeping track of it in the evaluator struct rather than via an int in each function

cstyan:
@brian-brazil when you have time please see the latest commit. I've simplified the tracking of the # samples. The # of samples in memory vs the # of samples being returned from an evaluation was tripping me up a bit still, but I think I've got things right this time. I've left comments where I was previously tracking samples but no longer think it's necessary, or where we're not tracking samples in the first place because _I think_ it's not necessary.

I think the metrics around query sizes in the other PR were interesting so I'll add some here. 

If everything's good I'll rebase down to one commit.

promql/engine.go
```diff
@@ -67,13 +68,22 @@ type (
...
-func (e ErrQueryTimeout) Error() string  { return fmt.Sprintf("query timed out in %s", string(e)) }
-func (e ErrQueryCanceled) Error() string { return fmt.Sprintf("query was canceled in %s", string(e)) }
+func (e ErrQueryTimeout) Error() string {
+	return fmt.Sprintf("query timed out in %s", string(e))
+}
+func (e ErrQueryCanceled) Error() string {
+	return fmt.Sprintf("query was canceled in %s", string(e))
+}
+func (e ErrTooManySamples) Error() string {
+	return fmt.Sprintf("query would return too many samples in %s", string(e))
```
brian-brazil: it's more the processing than the end result

promql/engine.go
```diff
@@ -677,7 +700,8 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 	for i, e := range exprs {
 		// Functions will take string arguments from the expressions, not the values.
 		if e != nil && e.Type() != ValueTypeString {
-			matrixes[i] = ev.eval(e).(Matrix)
+			ret := ev.eval(e)
```
brian-brazil: You don't need to change this

promql/engine.go
```diff
@@ -705,11 +729,12 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 			vectors[i] = vectors[i][:0]
 			for si, series := range matrixes[i] {
 				for _, point := range series.Points {
-					if point.T == ts {
+					if point.T == ts && ev.currentSamples < ev.maxSamples {
 						vectors[i] = append(vectors[i], Sample{Metric: series.Metric, Point: point})
 						// Move input vectors forward so we don't have to re-scan the same
 						// past points at the next step.
 						matrixes[i][si].Points = series.Points[1:]
+						ev.currentSamples++
```
brian-brazil: You need to undo this before the next ts iteration

promql/engine.go
```diff
@@ -705,11 +729,12 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 			vectors[i] = vectors[i][:0]
 			for si, series := range matrixes[i] {
 				for _, point := range series.Points {
-					if point.T == ts {
+					if point.T == ts && ev.currentSamples < ev.maxSamples {
```
brian-brazil: This should error when it's hit

promql/engine.go
```diff
@@ -720,6 +745,14 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 		enh.ts = ts
 		result := f(args, enh)
 		enh.out = result[:0] // Reuse result vector.
+
+		// In some cases we only have the result of 'f' without having tracked any samples
+		// in the above for loop that built the input vectors.
+		ev.currentSamples += len(result)
+		if ev.currentSamples > ev.maxSamples {
+			ev.error(error(ErrTooManySamples(env)))
```
brian-brazil: Why do you need the `error()` here?

cstyan: do we not have more samples in memory via the return value of `f`?

promql/engine.go
cstyan: :+1: sorry, leftover from when I modified the function signatures to return errors

promql/engine.go
cstyan: Okay, so in that case I need a query for one of the tests that would load more than a single sample into memory per timestamp. The tests I have at the moment load at most one sample per ts, and I don't think setting the `ev.maxSamples` value to 0 is really a valid test.

promql/engine.go
brian-brazil: I mean specifically, why are you wrapping `ErrTooManySamples` in `error`?

promql/engine.go
brian-brazil: It's more testing across multiple timestamps

promql/engine.go
cstyan: Hmm, at some point I had a compile error with this but you're right they're not needed.

promql/engine.go
cstyan: How so? At the moment none of the tests I've added load more than a single sample into memory per ts.

brian-brazil: The issue is that you're still counting all the inputs from the previous timestamps when doing the current timestamp, but they're no longer in memory.

cstyan: Right, I've fixed that via lines 727/783. What I'm saying is that currently none of the tests have more than a single sample per timestamp. I want to find a query to add for a test that would load say 3 samples per timestamp into memory, and ensure the test fails if we set max samples to 2.

promql/engine.go
```diff
@@ -777,6 +780,7 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 			seriess[h] = ss
 
 		}
+		ev.currentSamples = tempNumSamples
```
brian-brazil: This can return early up a bit, so this needs reworking.

The input vectors to the function also need accounting for when they're no longer in use.

cstyan: > The input vectors to the function also need accounting for when they're no longer in use.

Not sure what you mean here. We should reset the number of samples immediately after the call to `f` rather than later on?

brian-brazil: With recursive calls to eval, once the inner evals are done we don't need to account for their memory.

promql/engine.go
```diff
@@ -697,10 +697,15 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 	numSteps := int((ev.endTimestamp-ev.startTimestamp)/ev.interval) + 1
 	matrixes := make([]Matrix, len(exprs))
 	origMatrixes := make([]Matrix, len(exprs))
+	tempNumSamples := 0
 	for i, e := range exprs {
 		// Functions will take string arguments from the expressions, not the values.
 		if e != nil && e.Type() != ValueTypeString {
+			tempNumSamples = ev.currentSamples
 			matrixes[i] = ev.eval(e).(Matrix)
+			// Once inner calls to eval are done we don't need to account for memory for
```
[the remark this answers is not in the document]
cstyan: Right, the output is saved in `matrixes[i]` and then we're tracking the number of samples in `matrixes` below in the loop for the timestamp. We should be adding `len(matrixes)` to `ev.currentSamples` after resetting to `tempNumSamples` I think.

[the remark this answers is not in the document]
cstyan: I think this is right.

cmd/prometheus/main.go
```diff
@@ -186,6 +187,8 @@ func main() {
 
 	a.Flag("query.max-concurrency", "Maximum number of queries executed concurrently.").
 		Default("20").IntVar(&cfg.queryConcurrency)
+	a.Flag("query.max-samples", "Maximum number of samples a single query can load into memory. Note that queries will fail if they would load more samples than this into memory, so this also limits the number of samples a query can return.").
+		Default("1000000").IntVar(&cfg.queryMaxSamples)
```
brian-brazil: I'd suggest 50M, that's nominally 800MB

promql/engine.go
```diff
@@ -889,12 +944,14 @@ func (ev *evaluator) eval(expr Expr) Value {
 		return ev.eval(e.Expr)
 
 	case *UnaryExpr:
-		mat := ev.eval(e.Expr).(Matrix)
+		ret = ev.eval(e.Expr)
```
brian-brazil: This can be undone

promql/engine.go
```diff
@@ -878,8 +927,14 @@ func (ev *evaluator) eval(expr Expr) Value {
 				// Only buffer stepRange milliseconds from the second step on.
 				it.ReduceDelta(stepRange)
 			}
-			if len(ss.Points) > 0 {
-				mat = append(mat, ss)
+			// (callum) I think we still need to track number of samples here
+			if len(ss.Points) > 0 && ev.currentSamples < ev.maxSamples {
```
brian-brazil: This seems more complicated than it needs to be

promql/engine.go
```diff
@@ -818,8 +865,10 @@ func (ev *evaluator) eval(expr Expr) Value {
 		otherInArgs := make([]Vector, len(e.Args))
 		for i, e := range e.Args {
 			if i != matrixArgIndex {
-				otherArgs[i] = ev.eval(e).(Matrix)
+				ret = ev.eval(e)
+				otherArgs[i] = ret.(Matrix)
```
brian-brazil: This can be undone

promql/engine.go
```diff
@@ -668,16 +691,22 @@ func (enh *EvalNodeHelper) signatureFunc(on bool, names ...string) func(labels.L
...
+	tempNumSamples := 0
 	for i, e := range exprs {
 		// Functions will take string arguments from the expressions, not the values.
 		if e != nil && e.Type() != ValueTypeString {
+			tempNumSamples = ev.currentSamples
 			matrixes[i] = ev.eval(e).(Matrix)
+			// Once inner calls to eval are done we don't need to account for memory for
+			// samples within those calls. However the result of the call to eval is in mem.
+			ev.currentSamples = tempNumSamples
+			ev.currentSamples += len(matrixes[i])
```
brian-brazil: This is the number of series, not the number of samples. I think you're doing the saving of tempNumSamples in the wrong place, and you don't need any code changes to this part as ev.currentSamples will already have the right value.

promql/engine.go
```diff
@@ -700,16 +729,21 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 	enh := &EvalNodeHelper{out: make(Vector, 0, biggestLen)}
 	seriess := make(map[uint64]Series, biggestLen) // Output series by series hash.
 	for ts := ev.startTimestamp; ts <= ev.endTimestamp; ts += ev.interval {
+		// Reset num. of samples in memory after each timestamp.
```
brian-brazil: But the number of samples will have gone up, due to the output.

promql/engine.go
```diff
@@ -668,16 +691,22 @@ func (enh *EvalNodeHelper) signatureFunc(on bool, names ...string) func(labels.L
...
 // rangeEval evaluates the given expressions, and then for each step calls
 // the given function with the values computed for each expression at that
-// step.  The return value is the combination into time series of  of all the
+// step.  The return value is the combination into time series of all the
 // function call results.
 func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ...Expr) Matrix {
 	numSteps := int((ev.endTimestamp-ev.startTimestamp)/ev.interval) + 1
 	matrixes := make([]Matrix, len(exprs))
 	origMatrixes := make([]Matrix, len(exprs))
+	tempNumSamples := 0
```
brian-brazil: Shouldn't this be ev.currentSamples? Ultimately when this function returns ev.currentSamples should be the previous value, plus how ever many samples we're returning.

cstyan: wouldn't that just be whatever the value of `ev.currentSamples` is when we return?

cstyan:
Sorry @brian-brazil but I'm still a bit lost trying to understand exactly how the `ev.currentSamples` reset should work.

promql/engine.go
cstyan: In every case? After the call to eval, we've potentially added to `ev.currentSamples` in  eval via`*Call`, `*VectorSelector` cases, or via `vectorSelector()`/`matrixSelector()` in which cases we wouldn't have reset `ev.currentSamples` afterwards, or we've called `rangeEval()`.

As I said earlier, I'm sill a bit lost understanding how to properly handle all of the cases. Any pointers would be helpful.

brian-brazil: Yes, in every case. 

> After the call to eval, we've potentially added to ev.currentSamples in ...

Yes, and that needs to be undone before we return as those are no longer in memory.

So if we start a rangeEval with say 10 samples used up by higher in the call stack, then any functions we call we will add to that - so say we're now at 30 samples. If the output of the rangeEval is 5 samples then currentSamples should be 15 by time the rangeEval returns.

cstyan: so then here before a call to `ev.eval` we want to be sure `ev.currentSamples` is saved to a temp var, and after the call `ev.currentSamples` should be set to that temp var plus the length of all the series in `matrixes[i]`?

brian-brazil: Yes

promql/engine.go
```diff
@@ -700,16 +726,21 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
...
 			vectors[i] = vectors[i][:0]
 			for si, series := range matrixes[i] {
 				for _, point := range series.Points {
-					if point.T == ts {
+					if point.T == ts && ev.currentSamples < ev.maxSamples {
 						vectors[i] = append(vectors[i], Sample{Metric: series.Metric, Point: point})
 						// Move input vectors forward so we don't have to re-scan the same
 						// past points at the next step.
 						matrixes[i][si].Points = series.Points[1:]
+						ev.currentSamples++
```
cstyan: @brian-brazil so now I assume that since we're adding the # of samples in the matrices returned by the calls to eval, we don't need to add to `ev.currentSamples` here, but we should keep the later block where we add the length of the `result` from `f`. Is that correct?

promql/engine.go
```diff
@@ -720,16 +751,25 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 		enh.ts = ts
 		result := f(args, enh)
 		enh.out = result[:0] // Reuse result vector.
+
+		// In some cases we only have the result of 'f' without having tracked any samples
+		// in the above for loop that built the input vectors.
+		ev.currentSamples += len(result)
```
cstyan: keep this?

promql/engine.go
```diff
@@ -668,16 +691,19 @@ func (enh *EvalNodeHelper) signatureFunc(on bool, names ...string) func(labels.L
...
 // function call results.
 func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ...Expr) Matrix {
 	numSteps := int((ev.endTimestamp-ev.startTimestamp)/ev.interval) + 1
 	matrixes := make([]Matrix, len(exprs))
 	origMatrixes := make([]Matrix, len(exprs))
+	tempNumSamples := ev.currentSamples
 	for i, e := range exprs {
 		// Functions will take string arguments from the expressions, not the values.
 		if e != nil && e.Type() != ValueTypeString {
+			tempNumSamples = ev.currentSamples
```
brian-brazil: This wipes out the tempNumSamples if there's more than one expr, you don't need this line

promql/engine.go
```diff
@@ -668,16 +691,19 @@ func (enh *EvalNodeHelper) signatureFunc(on bool, names ...string) func(labels.L
...
 			matrixes[i] = ev.eval(e).(Matrix)
+			ev.currentSamples = tempNumSamples + matrixes[i].TotalSamples()
```
brian-brazil: ev.currentSamples will already have the right value due to the eval, no need to change it

promql/engine.go
```diff
@@ -700,16 +726,21 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 	enh := &EvalNodeHelper{out: make(Vector, 0, biggestLen)}
 	seriess := make(map[uint64]Series, biggestLen) // Output series by series hash.
 	for ts := ev.startTimestamp; ts <= ev.endTimestamp; ts += ev.interval {
+		// Reset num. of samples in memory after each timestamp.
+		tempNumSamples = ev.currentSamples
```
brian-brazil: You're losing the original value of tempNumSamples here

promql/engine.go
brian-brazil: No, we still need these as it is a (temporary) 2nd copy of some of the data in the input vector.

promql/engine.go
brian-brazil: Yes, this will be part of the output

What you ultimately want is the original tempNumSamples plus all of these in ev.currentSamples when you return

promql/engine.go
cstyan: I don't understand, each call to eval is adding samples to `ev.currentSamples`, but after the call we want the result of `ev.currentSamples` to be the value prior to the `eval` call plus the amount of samples returned.

cstyan: as we discussed here: https://github.com/prometheus/prometheus/pull/4513#discussion_r218979274

brian-brazil: After the call, the value of ev.currentSamples should already be the eval call plus the amount of samples returned - as that's the contract.

brian-brazil:
No, those changes aren't needed.

The goal should be that at the end of the evalRange call (or any other eval call) that ev.currentSample is the number of samples that function is about to return plus the samples held by all calling functions up the call stack.

cstyan:
Okay, I believe the latest commit is correct then.

promql/engine.go
brian-brazil: You're still losing this here, you need a different variable to store this in.

cstyan: I see what you're saying. The number of samples we have going into the function is the value we want to ues (plus the amount of new samples) when returning, but we want to reset to the previous value of `ev.currentSamples` after each timestamp iteration. Latest commit should fix this.

cstyan:
rebased off master and squashed everything but the lastest commit into one commit

promql/engine.go
```diff
@@ -705,16 +734,21 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
...
```
brian-brazil: This can go above the for, and then you can reset it here - which is more natural than doing it at the end of the loop

promql/engine.go
```diff
@@ -728,16 +762,25 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 			ev.errorf("vector cannot contain metrics with the same labelset")
 		}
 		enh.out = result[:0] // Reuse result vector.
+
+		// In some cases we only have the result of 'f' without having tracked any samples
+		// in the above for loop that built the input vectors.
+		ev.currentSamples += len(result)
```
brian-brazil: You want to add this to tempNumSamples as the output data from previous cycles is in memory.

cstyan: Hmmm, sorry I don't understand why. 

In addition to, or instead of adding it to `ev.currentSamples`? It seems to me that it should be in addition to since we check `ev.currentSamples` when determining if we have too many samples in memory.

brian-brazil: Yes. After every iteration we're now holding the results of previous iterations in memory, which needs to be part of tempNumSamples.

promql/engine.go
```diff
@@ -728,16 +763,29 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 			ev.errorf("vector cannot contain metrics with the same labelset")
 		}
 		enh.out = result[:0] // Reuse result vector.
+
+		// In some cases we only have the result of 'f' without having tracked any samples
```
brian-brazil: I don't think this comment helps, one way or the other we should always be adding this.

promql/engine.go
```diff
@@ -1127,14 +1200,22 @@ func (ev *evaluator) matrixIterSlice(it *storage.BufferedSeriesIterator, mint, m
...
+				ev.error(ErrTooManySamples(env))
+			}
 		}
 	}
 	// The seeked sample might also be in the range.
 	if ok {
 		t, v := it.Values()
 		if t == maxt && !value.IsStaleNaN(v) {
-			out = append(out, Point{T: t, V: v})
+			if ev.currentSamples < ev.maxSamples {
```
brian-brazil: This should error

promql/engine.go
```diff
@@ -331,9 +352,8 @@ func (ng *Engine) exec(ctx context.Context, q *query) (Value, error) {
 	defer q.cancel()
 
 	const env = "query execution"
-
-	evalSpanTimer, ctx := q.stats.GetSpanTimer(ctx, stats.EvalTotalTime)
-	defer evalSpanTimer.Finish()
+	evalTimer := q.stats.GetTimer(stats.EvalTotalTime).Start()
```
brian-brazil: Why does the function change here?

cstyan: I think I picked the wrong change when fixing a merge conflict here during rebase, fixing.

promql/engine.go
```diff
@@ -704,17 +734,23 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 	}
 	enh := &EvalNodeHelper{out: make(Vector, 0, biggestLen)}
 	seriess := make(map[uint64]Series, biggestLen) // Output series by series hash.
+	tempNumSamples = ev.currentSamples
 	for ts := ev.startTimestamp; ts <= ev.endTimestamp; ts += ev.interval {
+		// Reset num. of samples in memory after each timestamp.
```
brian-brazil: number

promql/engine.go
```diff
@@ -673,15 +697,21 @@ func (enh *EvalNodeHelper) signatureFunc(on bool, names ...string) func(labels.L
...
-// step.  The return value is the combination into time series of  of all the
+// step.  The return value is the combination into time series of all the
 // function call results.
 func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ...Expr) Matrix {
 	numSteps := int((ev.endTimestamp-ev.startTimestamp)/ev.interval) + 1
 	matrixes := make([]Matrix, len(exprs))
 	origMatrixes := make([]Matrix, len(exprs))
+
+	originalNumSamples := ev.currentSamples
+
```
brian-brazil: There's a bit much vertical whitespace here

promql/engine.go
```diff
@@ -673,15 +697,21 @@ func (enh *EvalNodeHelper) signatureFunc(on bool, names ...string) func(labels.L
...
+	var tempNumSamples int
```
brian-brazil: Why not create this when you assign to it?

promql/engine.go
```diff
@@ -704,17 +734,23 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
...
 			for si, series := range matrixes[i] {
 				for _, point := range series.Points {
-					if point.T == ts {
+					if point.T == ts && ev.currentSamples < ev.maxSamples {
 						vectors[i] = append(vectors[i], Sample{Metric: series.Metric, Point: point})
 						// Move input vectors forward so we don't have to re-scan the same
 						// past points at the next step.
 						matrixes[i][si].Points = series.Points[1:]
+						ev.currentSamples++
+					} else if ev.currentSamples >= ev.maxSamples {
```
brian-brazil: The else here and duplicated inverse condition is a bit odd, it seems like it might be accidentally broken bu a future change. I'd move this inside the above if as a sub if.

promql/engine.go
```diff
@@ -728,16 +764,27 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
...
 		if ev.endTimestamp == ev.startTimestamp {
 			mat := make(Matrix, len(result))
 			for i, s := range result {
 				s.Point.T = ts
 				mat[i] = Series{Metric: s.Metric, Points: []Point{s.Point}}
 			}
+			ev.currentSamples = originalNumSamples + mat.TotalSamples()
 			return mat
 		}
-		// Add samples in output vector to output series.
```
brian-brazil: This should stay

promql/engine.go
```diff
@@ -828,6 +878,7 @@ func (ev *evaluator) eval(expr Expr) Value {
 			if i != matrixArgIndex {
 				otherArgs[i] = ev.eval(e).(Matrix)
 				otherInArgs[i] = Vector{Sample{}}
+				// (callum) this is just building args for later procesing so I don't think we need to track samples here
```
brian-brazil: Can you either clean these up or make them proper comments?

cstyan:
removed most of the comments you pointed out as they were only necessary while I was trying to track where samples were loaded in and out of memory, updated the rest and simplified the if block

brian-brazil:
Thanks!

maozi07:
I think it's better to check the memory prometheus used and have a limit, not the amount of samples. In my situation that's caused some problems for a max sample.Thank you all for your work

ernanirst:
@brian-brazil I read that each sample takes 16 bytes, so if I'm using a VM that has 8GB of RAM available, how would you advice on calculating `query.max-samples`? I'm supposing I can consider 1 GB for OS, 1 GB for Prometheus itself and the remaining 6 GB to the samples? And therefore 6GB / 16 bytes = 3,75 * 10^8 samples?

 Does this math makes sense?

ringerc:
This limits sample counts, but doesn't do any accounting for the labels for each series, right? 

From what I'm seeing in testing, loading a lot of "wide" series due to queries that consume a lot of info-metrics and/or query some high-churn info-metric series can still easily OOM the server even with quite conservative sample limits.

PR #4514, opened 2018-08-17
Expose EC2 instance owner as a discovery label.

jkohen:
This account ID/owner ID is necessary for uniquely identifying a VM instance.
This exposes the OwnerID field of the DescribeInstances response as `__meta_ec2_owner_id`. Other valid names could be `__meta_aws_account_id` and `__meta_ec2_account_id`.

Signed-off-by: Javier Kohen <jkohen@google.com>


fabxc:
:+1:

PR #4523, opened 2018-08-21
simplified SD updates throttling

krasi-georgiev:
as part of troubleshooting the SD updates bug I figured that when the receiver can't take any new updates(channel is full) the current throttling will block any targets updates

this is blocked 

https://github.com/prometheus/prometheus/blob/c66347768889f140253ecc7a0f4358bc051aaa01/discovery/manager.go#L150-L152
because we keep the lock here when the receiving channel is full.
https://github.com/prometheus/prometheus/blob/c66347768889f140253ecc7a0f4358bc051aaa01/discovery/manager.go#L166-L171

it also closes: #4470

### DIFF
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -82,10 +82,6 @@ type Manager struct {
 	targets map[poolKey]map[string]*targetgroup.Group
 	// The sync channels sends the updates in map[targetSetName] where targetSetName is the job value from the scrape config.
 	syncCh chan map[string][]*targetgroup.Group
-	// True if updates were received in the last 5 seconds.
-	recentlyUpdated bool
-	// Protects recentlyUpdated.
-	recentlyUpdatedMtx sync.Mutex
 }
 
 // Run starts the background processing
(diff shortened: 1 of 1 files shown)


brancz:
Could we also run a performance test against v2.3.2, as this code has been particularly fragile in the past.

krasi-georgiev:
this is only one part of the problem the other is that the scrape manager stops the old scrapes one by one so I have refactored the code a bit to fix this.
I will open another PR , but that one is quite involved so I am still trying to brake it in 1-2 smaller ones. than I will run the benchmark.

@grobie I will think about a test, but I think maybe this could be caught only in some bench-marking.

krasi-georgiev:
@brancz decided to run it anyway, all maintainters can also do this with a single comment.

krasi-georgiev:
lets compare it with master since 2.3.2 has too many commits.

krasi-georgiev:
@grobie I tried few different ideas about a test, but nothing that would catch this properly. 
We send updates every 5 seconds (from the discovery manager to the scrape manager) so to catch blocked or no updates would need to make the test last more than that which  leans towards an e2e test which we plan to add  in prombench at some point.

The test will compare a good know version and will scale up and down and check at a regular interval to  make sure that the total targets on both tested instance match at all times.

discovery/manager.go
```diff
@@ -132,43 +128,48 @@ func (m *Manager) startProvider(ctx context.Context, poolKey poolKey, worker Dis
...
-			if m.recentlyUpdated {
-				m.syncCh <- m.allGroups()
-				m.recentlyUpdated = false
+			// Signal that there was an update.
+			select {
+			case updateReceived <- struct{}{}:
+			default:
+			}
+		case <-ticker.C: // Some discoverers send updates too often so we send these to the receiver once every 5 seconds.
+			select {
```
grobie: Isn't it possible that the provider loop gets stuck her forever? I'd have expected a default branch to handle the case that there hasn't been any update / a new message on the `updateReceived` channel. Or what am I missing?

krasi-georgiev: yes you are right, updated. The default branch now also removes the need for `case <-ctx.Done():` as this will be caught  at the top iteration.

PR #4524, opened 2018-08-21
Add rule group update failure metric

povilasv:
Adds rule reloading failure metric. 

This is useful when you have automation which automatically reloads prometheus configs and would like to alert on reloading configuration failures.

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -102,6 +102,13 @@ var (
 		[]string{"rule_group"},
 		nil,
 	)
+	updateFailures = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "rule_group_update_failures_total",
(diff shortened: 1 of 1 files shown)


gouthamve:
Config loading has: `prometheus_config_last_reload_success_timestamp_seconds` and `prometheus_config_last_reload_successful`. Can you make the metric here similar to that for consistency?

povilasv:
@gouthamve thanks, didn't knew about that metric. So maybe we don't really need this metric, as `prometheus_config_last_reload_successful` can be used for unsuccessful reload alerting.

gouthamve:
Actually you are right. I just ran a reload after botching up a rule and `prometheus_config_last_reload_successful` turned to be `0`.

Feel free to close this PR if the present metric satisfies your use-case :)

Issue #4301, opened 2018-06-21
Prometheus configuration reloads slowly


lyonssp:
## Bug Report
I'm reloading Prometheus configurations through a continuous deployment process using ansible.  I'm finding that configurations take such a long time to reload that what seem like sensible timeout thresholds are consistently passed by the reload process.  

What are some factors that could be causing the significant reload times?

**What did you do?**
`curl -X POST localhost:9090/-/reload`

**What did you expect to see?**
Expected Prometheus to reload my configuration in < 90s

**What did you see instead? Under which circumstances?**
Prometheus seems to take ~2.5 minutes to reload the configuration. 

**Environment**

Prometheus is deployed using docker with the configuration mounted from the file system to 
/etc/prometheus/prometheus.yml.  I am loading a new file to the local filesystem and almost immediately triggering a reload of the prometheus configuration via lifecycle APIs.

Containers aren't rebooted frequently.

* System information:
Linux 4.4.0-1057-aws x86_64

* Prometheus version:

prometheus, version 2.2.1
  build date:       20180314-14:15:45
  go version:       go1.10

* Prometheus configuration file:
```
global:
  scrape_interval: 30s
  scrape_timeout: 30s

alerting:
  alertmanagers:
  - file_sd_configs:
    - files:
      - /etc/prometheus/alertmanagers.yml

rule_files:
  - /etc/prometheus/rules/*.yml

scrape_configs:
  - job_name: '<redacted>'
    ec2_sd_configs:
    - region: us-east-1 
      port: 9100
      access_key: <redacted>
      secret_key: <redacted>
    relabel_configs: <redacted>

  - job_name: '<redacted>'
    ec2_sd_configs:
    - region: <redacted>
      port: 9100
      access_key: <redacted>
      secret_key: <redacted>
     relabel_configs: <redacted>

  - job_name: '<redacted>'
    ec2_sd_configs:
    - region: <redacted>
      port: 9100
      access_key: <redacted>
      secret_key: <redacted>
    relabel_configs: <redacted>

  - job_name: '<redacted>'
    ec2_sd_configs:
    - region: <redacted>
      port: 9090
      access_key: <redacted>
      secret_key: <redacted>
    relabel_configs: <redacted>

  - job_name: '<redacted>'
    ec2_sd_configs:
    - region: <redacted>
      port: 9090
      access_key: <redacted>
      secret_key: <redacted>
    relabel_configs: <redacted>

  - job_name: '<redacted>'
[...]

Issue #4124, opened 2018-04-30
K8s SD: Targets are not updated


beorn7:
## Bug Report
**What did you do?**

Scaled new K8s pods up, scaled old ones down.

**What did you expect to see?**

Old pods disappear from targets, new ones show up.

**What did you see instead? Under which circumstances?**

Prometheus doesn't see the new pods, tries to scrape the old ones.

**Environment**

* System information:

	Linux 4.4.10+soundcloud x86_64

* Prometheus version:

	most recently investigated version is revision 2be543e65ac2587ac46d87fef759d80bfb1c68f5 (in the release-2.2 branch). However, the same happened in all 2.2 releases, most notably also before the grand K8s SD refactoring of #4117 .

* Prometheus configuration file:

See the various comments below.

beorn7:
This issue is seen in production at SoundCloud. I'm still working on reproducing it with a minimal setup.

In particular, this issue can __not__ be reproduced with the following setup (cf. https://github.com/prometheus/prometheus/pull/4117#issuecomment-384909820 ):

@krasi-georgiev wrote: "using the configs below seems to work as expected with minikube. 
When scaling up and down the deployment pods the  targets list get updated properly in the prometheus GUI."

```
global:
  scrape_interval:     500ms

scrape_configs:
- job_name: 'kubernetes-pods'
  tls_config:
    insecure_skip_verify: true
  
  kubernetes_sd_configs:
  - api_server: https://192.168.99.101:8443
    role: pod
    namespaces:
      names:
      - default
    tls_config:
      ca_file: /home/krasi/.minikube/ca.crt
      cert_file: /home/krasi/.minikube/client.crt
      key_file: /home/krasi/.minikube/client.key
```

```
apiVersion: apps/v1 
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 2 
  template: 
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.7.9
        ports:
        - containerPort: 80
```

beorn7:
Here is the job config that creates the problem:

```yaml
- job_name: myjob
  kubernetes_sd_configs:
  - api_server: https://api.mycluster.k8s.example.org
    role: pod
    tls_config:
      ca_file: "/srv/prometheus/k8s-certificates/mycluster/ca.crt"
[...]

Issue #4518, opened 2018-08-20
k8s Discoverer never stops sending updates even when there are no changes.


krasi-georgiev:
as part of troubleshooting https://github.com/prometheus/prometheus/issues/4124 I figred that the k8s provider never stops sending updates.

if you put some printf's you will see that it keeps sending updated even when there are no changes in the k8s cluster.
 
https://github.com/prometheus/prometheus/blob/f571b69010d8ea4a9a4c8ecb3a3c07096c7b0238/discovery/manager.go#L143-L153

if you add some printf's in the same corresponding block in v2.0.0 it stops sending updates when there are no changes.
https://github.com/prometheus/prometheus/blob/0a74f98628a0463dddc90528220c94de5032d1a0/discovery/discovery.go#L253-L287

I am thinking something changed at some point when refactoring the k8s code, updating the k8s client or fixing the bug with the namespaces.

ping @cofyc , @simonpasquier

krasi-georgiev:
note: to replicate you need to remove the locking as there is another bug that blocks there.
```
m.recentlyUpdatedMtx.Lock() 
m.recentlyUpdated = true 
m.recentlyUpdatedMtx.Unlock() 
```

krasi-georgiev:
the change happened in https://github.com/prometheus/prometheus/pull/3660/files so that we get updates even when the  deployment is scaled to 0. In other words so that we send an update when all pods have been deleted.

The k8s klient keeps sending updates for the `kube-controller-manager` and the only change is that the `ResourceVersion` has changed.


old
```
[...]

PR #4526, opened 2018-08-22
[WIP] fix for the slow updates of targets changes

krasi-georgiev:
The scrape manage receiver channel now just saves the target sets
and another backgorund runner updates the scrape loops every 5 seconds.
This is so that the scrape manager doesn't block the receiving channel
when it does the  long background reloading of the scrape loops.

Active and dropped targets are now saved in each scrape pool instead of
the scrape manager. This is mainly to avoid races when getting the
targets via the web api.

When reloading the scrape loops now happens in parallel to speed up the
final disared state and this also speeds up the prometheus's shutting
down.

fixes: https://github.com/prometheus/prometheus/issues/4124
fixes: https://github.com/prometheus/prometheus/issues/4301

I will run a benchmark once the https://github.com/prometheus/prometheus/pull/4523 is merged and I rebase it to this PR as well.


krasi-georgiev:
@grobie still polishing few details, but you can already test this version or wait few days to clear all small issues.

The last major problem is with the http2 go client update in https://github.com/golang/net/commit/1c05540f6879653db88113bc4a2b70aec4bd491f , but as long at your k8s cluster doesn't use the default connection stream limit of 1000, you should be fine.
Alternatevely you can disable HTTP2 by running prometheus with 
```
DISABLE_HTTP2=true ./prometheus ....
```

more details about the http2 client issue.
https://github.com/kubernetes/client-go/issues/456

krasi-georgiev:
the benchmark is looking good. When I scale up and down the CPU and memory usage is higher, but this is to be expected as all processing is now done in parallel.

scrape/manager.go
```diff
@@ -50,28 +51,84 @@ type Manager struct {
...
+	m.mtxScrape.Lock()
+	var wg sync.WaitGroup
+	for setName, groups := range m.targetSets {
+		var sp *scrapePool
+		existing, ok := m.scrapePools[setName]
+		if !ok {
+			scrapeConfig, ok := m.scrapeConfigs[setName]
+			if !ok {
+				level.Error(m.logger).Log("msg", "error reloading target set", "err", fmt.Sprintf("invalid config id:%v", setName))
+				return
```
simonpasquier: Could it happen? Does it need to return or it could continue?

Right now the scrape manager's configuration is reloaded before the discovery manager.

https://github.com/prometheus/prometheus/blob/75bd348135019c6bf5bee4350a339964e62f2d8a/cmd/prometheus/main.go#L323-L333

So I guess it can happen when a scrape job is removed from the configuration and the configuration update has been applied to the scrape manager but the discovery manager hasn't picked up yet. I'd say that we can silently drop the item then.

_I see that this code already exists in the previous version so feel free to ignore_

krasi-georgiev: If someone changes the workflow somewhere we might hit this so this way we will know soon enough that something is not right. This is one of the problems with the current implementation of how wet handle the config reloading. 
I have suggested how to improve, but there was no feedback at the time.

krasi-georgiev: Either way this is story for another PR ;)

beorn7:
Hi @krasi-georgiev , I'm back at work and would like to canary all those SD changes. What's the state of this PR? Should this still be included? You said you wanted to rebase this before further checking.

simonpasquier:
@beorn7 IIUC #4124 that you reported on targets not being updated should be fixed now on master by https://github.com/prometheus/prometheus/pull/4523 and  https://github.com/prometheus/prometheus/pull/4556. This PR is more about optimizations that would speed up the recreation of the scrape loops. Also #3912 (which merges identical SD configurations into a single provider and thus avoids hammering the K8S API) has been merged. Finally Krasi has found a couple of issues with the K8S go client (#4518 and #4528) but not sure that they apply to your environment.

In summary, you're more than welcome to test the current master on your clusters :smirk:

krasi-georgiev:
@beorn7 , @grobie would you mind testing this?

simonpasquier:
LGTM but I think it would worth to have unit tests for this.

krasi-georgiev:
@simonpasquier what unit tests did you have in mind? 

This PR fixes the time it takes to reload the scrape loops. 
An e2e with some timings  would be flaky, but suggestions are welcome.

beorn7:
I can run this on one of our servers at my next convenience.

krasi-georgiev:
@beorn7 appreciated!

simonpasquier:
I was thinking about something similar to https://github.com/prometheus/prometheus/pull/4582 although looking at the manager's code it would require some adjustments to be able to skip the effective launch of the scrapers.

beorn7:
Running this now (with current state of master merged in).

Disclaimer: I'm not using race detector (would not work on the heavily loaded test machine), but I recommend to do so before merging this.

I guess this change should result in faster reloads of the config file. I sighup'd the test server and the 2.4.0 production server a few times and couldn't detect a significant improvement, mostly because reload times are vastly different from case to case, from 5s to 30s, more or less the same on both servers.

Restart time is quite long on this server (around 12m (!)) and isn't significantly different with this PR merged in. (Perhaps the new WAL implementation is slower? I don't remember such long startup times from earlier 2.x versions. But perhaps our load has just increased so much in the meantime...)

Altogether, the two servers seem to behave mostly the same.

krasi-georgiev:
I tested this mainly with the k8s discovery.
Using the config from: https://github.com/prometheus/prometheus/issues/4124#issuecomment-412441386

in general 100 jobs with 20-30 targets per jobs.

Than randomly scaling up and down the number of targets from 1 to 20 .

Before this pr it takes minutes to reflect the changes in the  /targets web gui and it can never catch up.

krasi-georgiev:
When I tested locally it also improved the shutdown time as now old scrape loops are stopped in parallel.

simonpasquier:
@krasi-georgiev have you retried your tests with the latest master? I suspect that the various improvements to the SD manager may have change the situation.

beorn7:
@krasi-georgiev I see. I guess our test server doesn't see quickly changing targets that often. I might be able to identify one of those, and then run the canary binary there.

beorn7:
Looking around, with 2.4.0, all our practical uses of K8s SD seem to update fast enough so that any further improvements would not stick out of the noise.

krasi-georgiev:
Maybe @brancz can also give it a try since it is very easy to replicate using his configs even using minikibe.

@simonpasquier this change is in the scrape manager which is completely independent from the sd manger. 
Which changes do you think might be relevant here?

simonpasquier:
>  Which changes do you think might be relevant here

Mostly #3912. It might be that optimizing the discovery part (basically much less requests to the K8S API) reduced the contention and made the problem less acute on the scrape side.

krasi-georgiev:
aah yes that definitely helped, but processing the loops in parallel is another great improvement.

simonpasquier:
Scaling from 0 to 2000 targets back and forth, I definitely see an improvement when down-scaling. WDYT of adding a few tests?

krasi-georgiev:
the only useful test that I can see is making  sure the Scrape manager never blocks the `tsets <-chan map[string][]*targetgroup.Group` 

Is this what you had in mind as well?

simonpasquier:
Yes. Also making sure that data received on the channel triggers scrape updates and that everything is teared down properly on shutdown. I know that it wasn't tested before but it looks like a good opportunity to fix it.

krasi-georgiev:
@simonpasquier yes good idea, I added some tests. 

Just not sure what you had in mind for: 

> everything is teared down properly on shutdown.

scrape/manager.go
```diff
@@ -50,28 +54,83 @@ type Manager struct {
...
+			}
+			sp = newScrapePool(scrapeConfig, m.append, log.With(m.logger, "scrape_pool", setName))
+			m.scrapePools[setName] = sp
+		} else {
+			sp = existing
+		}
+
+		wg.Add(1)
+		// Run the sync in paralel as these take a while and at high load can't catch up.
```
simonpasquier: s/paralel/parallel/

simonpasquier:
:+1: It would be even better to test the logic including `reload()` but it is more involved and wasn't covered before either... 

> everything is teared down properly on shutdown.

That calling `Stop()` does the right thing.

krasi-georgiev:
Yeah I didn't want to take it that far as this would probably end being too high maintenance. 
In general it is not a bad idea, but it is outside the topic of this PR.
Even if someone decides to implement it let's leave it for another PR so we keep this one on topic and don't block it for any longer that is needed. 
The same for the teardown.

PR #4532, opened 2018-08-23
Limit the number of samples remote read can return.

tomwilkie:
Tracking down Prometheus OOMs for a client, initially thought it was PromQL (see #4414), but looks like it is remote read requests.

Quick and dirty limit; probably want this plumbed through as a command line argument?

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -168,6 +168,9 @@ func main() {
 	a.Flag("storage.remote.flush-deadline", "How long to wait flushing sample on shutdown or config reload.").
 		Default("1m").PlaceHolder("<duration>").SetValue(&cfg.RemoteFlushDeadline)
 
+	a.Flag("storage.remote.read-sample-limit", "Maximum overall number of samples to return via the remote read interface, in a single query. 0 means no limit.").
+		Default("5e7").IntVar(&cfg.web.RemoteReadLimit)
+
 	a.Flag("rules.alert.for-outage-tolerance", "Max time to tolerate prometheus outage for restoring 'for' state of alert.").
 		Default("1h").SetValue(&cfg.outageTolerance)
 
@@ -176,9 +179,6 @@ func main() {
 
 	a.Flag("rules.alert.resend-delay", "Minimum amount of time to wait before resending an alert to Alertmanager. Must be lower than resolve_timeout in Alertmanager").
 		Default("1m").SetValue(&cfg.resendDelay)
-
-	a.Flag("storage.remote.read-sample-limit", "Maximum number of samples to return via the remote read interface, in a single query.  0 means no limit.").
-		Default("5e7").IntVar(&cfg.web.RemoteReadLimit)
 
 	a.Flag("alertmanager.notification-queue-capacity", "The capacity of the queue for pending Alertmanager notifications.").
 		Default("10000").IntVar(&cfg.notifier.QueueCapacity)
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -177,6 +177,9 @@ func main() {
 	a.Flag("rules.alert.resend-delay", "Minimum amount of time to wait before resending an alert to Alertmanager. Must be lower than resolve_timeout in Alertmanager").
 		Default("1m").SetValue(&cfg.resendDelay)
 
+	a.Flag("storage.remote.read-sample-limit", "Maximum number of samples to return via the remote read interface, in a single query.  0 means no limit.").
+		Default("5e7").IntVar(&cfg.web.RemoteReadLimit)
+
 	a.Flag("alertmanager.notification-queue-capacity", "The capacity of the queue for pending Alertmanager notifications.").
 		Default("10000").IntVar(&cfg.notifier.QueueCapacity)
 
--- a/storage/remote/codec.go
+++ b/storage/remote/codec.go
@@ -32,6 +32,19 @@ import (
 // decodeReadLimit is the maximum size of a read request body in bytes.
 const decodeReadLimit = 32 * 1024 * 1024
 
+type HTTPError struct {
+	msg    string
+	status int
+}
+
+func (e HTTPError) Error() string {
+	return e.msg
+}
+
+func (e HTTPError) Status() int {
+	return e.status
+}
+
 // DecodeReadRequest reads a remote.Request from a http.Request.
 func DecodeReadRequest(r *http.Request) (*prompb.ReadRequest, error) {
 	compressed, err := ioutil.ReadAll(io.LimitReader(r.Body, decodeReadLimit))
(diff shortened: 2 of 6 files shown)


gouthamve:
This might break some Thanos users. @fabxc @Bplotka

tomwilkie:
Yes, sorry should have said; OOMs seemed to be caused by queries submitted via Thanos.

We'll see if we can remove this limit through a streaming API as discussed at the dev summit, this is just a quick fix.

cmd/prometheus/main.go
```diff
@@ -164,6 +164,9 @@ func main() {
 	a.Flag("storage.remote.flush-deadline", "How long to wait flushing sample on shutdown or config reload.").
 		Default("1m").PlaceHolder("<duration>").SetValue(&cfg.RemoteFlushDeadline)
 
+	a.Flag("storage.remote.read-sample-limit", "Maxium number of samples to return via the remote read interface, in a single query.").
```
brian-brazil: Maximum

cmd/prometheus/main.go
```diff
@@ -164,6 +164,9 @@ func main() {
...
+		Default("20m").IntVar(&cfg.web.RemoteReadLimit)
```
brian-brazil: Usually 0 would mean no limit.

Does m mean milli here?

milesbxf: Default is `"20m"`, but the flag is an `IntVar` - running this results in `Error parsing commandline arguments: strconv.ParseFloat: parsing "20m": invalid syntax`

tomwilkie:
Yeah, should be 2e7.

I can also make 0 mean no limit, but it’s not very safe...

cmd/prometheus/main.go
Bplotka: So.. how to assess what max number of samples Prometheus can handle from remote read API?

storage/remote/codec.go
```diff
@@ -118,14 +118,19 @@ func FromQuery(req *prompb.Query) (int64, int64, []*labels.Matcher, error) {
...
 	resp := &prompb.QueryResult{}
 	for ss.Next() {
 		series := ss.At()
 		iter := series.Iterator()
 		samples := []*prompb.Sample{}
 
 		for iter.Next() {
+			numSamples += 1
+			if numSamples > sampleLimit {
+				return nil, fmt.Errorf("too many samples")
```
Bplotka: Can we also wrap the error with what is the limit?

brian-brazil: We should probably have some metrics around this too.

tomwilkie: And make it return a 400, not a 500...

tomwilkie:
> So.. how to assess what max number of samples Prometheus can handle from remote read API?

I've set the default to 20 million based on discussions on the other PR (#4513), but will happily consider making it higher.

tomwilkie:
> Incidentally is there any spike for streaming API for remote Read @tomwilkie? Can I help with this?

Yes we discussed at the dev summit making a streaming API to help thanos & other systems.  I've just done similar in cortex: https://github.com/weaveworks/cortex/pull/933

Its beyond the scope of this PR - I suggest you start a thread on the -dev mailing list.  We discussed streaming the compressed chunks out, instead of samples; previously I think you + I discussed trying to make a common API that can be used in both Cortex and Thanos.

brian-brazil:
We're in a different context here, so a higher value could be fine. I arbitrarily say that we aim to keep it under 1GB by default.

tomwilkie:
Ignoring labels, 60M * 16 bytes = 1GB, so I'll set it to that.

brian-brazil:
Call it 50M to allow for the labels, and requests for many series with few points. It's also more round.

tomwilkie:
Rebased against master; ready for final review I hope!

storage/remote/codec.go
```diff
@@ -134,14 +147,22 @@ func FromQuery(req *prompb.Query) (int64, int64, []*labels.Matcher, *storage.Sel
...
 		series := ss.At()
 		iter := series.Iterator()
 		samples := []*prompb.Sample{}
 
 		for iter.Next() {
+			numSamples += 1
+			if sampleLimit > 0 && numSamples > sampleLimit {
+				return nil, HTTPError{
+					msg:    fmt.Sprintf("exceeded sample limit (%d)", sampleLimit),
+					status: http.StatusRequestEntityTooLarge,
```
brian-brazil: This isn't a 413, as it's the server response that's too big. I think this should be a 400.

brancz:
Can you actually try a large size message? As far as I know the default size for a proto message is limited to 64Mb (last time I checked they warned of larger sizes as it loosens some security constraints, but not sure of the details).

Bplotka:
To clarify @brancz do you mean "if it's even useful to have this flag because anything more than 50m samples will make protobuf message larger than default proto message (64MB)"?

Good point, not sure of the implications. Also, how you get 64MB? The grpc https://github.com/grpc/grpc/issues/7927 by default allows max 4MB (: But it might be because protocol limitations itself. Would love to know more about implications. 

BTW we are working on streaming as well: https://docs.google.com/document/d/1JqrU3NjM9HoGLSTPYOvR217f5HBKBiJTqikEB9UiJL0/edit?ts=5b8f135a# 
- @tomwilkie how the current flag would fit into streamed response? That limit would be an overall limit? or per frame?

gouthamve:
We don't use gRPC, we marshal into proto and send it over HTTP.
https://github.com/prometheus/prometheus/blob/14b044963748b0d9d95f89409fb928ded0da51c8/storage/remote/codec.go#L68-L81

Not sure where the 64MB limit is documented though.

Bplotka:
I know @gouthamve, just putting this an example

tomwilkie:
> As far as I know the default maximum size for a proto message is limited to 64Mb 

I'm not aware, nor could I find, any limit in the golang proto bindings at https://github.com/golang/protobuf.

> To clarify @brancz do you mean "if it's even useful to have this flag because anything more than 50m samples will make protobuf message larger than default proto message (64MB)"?

Yes it is still useful; we build the datastructure before encoding it, and its the building of the datastructure to which this limits applies.  Even there was a 64MB limit, we still see OOMs for large queries to the remote read endpoint, imply its the building of the datastructure causing the OOM.

> @tomwilkie how the current flag would fit into streamed response? That limit would be an overall limit? or per frame?

Until an agreed upon approach exists for the streaming, no.  I can only speculate that this could apply to a single "frame", but thats just a guess.  Lets take this one offline, we can always deprecate the flag if needs be.

tomwilkie:
> The golang implementations shouldn't actually be affected by this, but it feels like we should follow the best practice especially for an API that is exposed to be consumed by arbitrary implementations.

By setting the limit to 64MB by default, you'd be limiting Thanos from fetching ~8k timeseries per query (64MB / 16bytes * 15s scrape interval / 2hr head block).  I think that limit would be too low, unless my math is wrong.

brancz:
That's fair. Maybe we should improve this step by step instead and only impose smaller limits if we actually manage to get streaming for example.

web/api/v1/api_test.go
```diff
@@ -861,6 +862,10 @@ func TestReadEndpoint(t *testing.T) {
 	recorder := httptest.NewRecorder()
 	api.remoteRead(recorder, request)
 
+	if recorder.Code/100 != 2 {
```
brancz: Can we add a test for the too large case as well?

sksingh20:
Team, I have major issue on grafana dashboard which is increasing step size by itself resulting in substantial loss of data.
https://github.com/grafana/grafana/issues/34775

As per them this is limitation of Prometheus where prometheus is making limit of 11K records.

Could you please confirm if there is any limit enforced for query to return data points and what is that value.

sksingh20:
@brancz Any thoughts on claim by Grafana teams?

PR #4533, opened 2018-08-24
use Welford/Knuth method to compute standard deviation and variance

### LINKED ISSUE #4527: Bug: stddev_over_time returns NaN for constant series; floating point inaccuracy
## Bug Report
In our production ,we use stddev_over_time to aggregate metric series, because the samples of metric series are constant, the aggregate result may be NaN;
because in function funcStddevOverTime :

```go
 func funcStddevOverTime(vals []Value, args Expressions, enh *EvalNodeHelper) Vector {
	return aggrOverTime(vals, enh, func(values []Point) float64 {
		var sum, squaredSum, count float64
		for _, v := range values {
			sum += v.V
			squaredSum += v.V * v.V
			count++
		}
		avg := sum / count
		return math.Sqrt(squaredSum/count - avg*avg)
	})
}
```

`squaredSum/count - avg*avg`  is Negative Value;

we expect the result is 0;

* Prometheus version:

we use prometheus  2.1 

* Logs:
we add logs to see the result:

DEBUG::calc stddev, elems:  [1.5990505637277868 @[1534906200000] 1.5990505637277868 @[1534906500000] 1.5990505637277868 @[1534906800000]]
DEBUG::calc stddev, sum:  4.797151691183361
DEBUG::calc stddev, squaredSum:  7.670888116074458
DEBUG::calc stddev, sub:  -4.440892098500626e-16
DEBUG::calc stddev, stddev:  NaN

DanCech:
Closes #4527 

This PR fixes float precision-related issues encountered when using the naive sum of squares method for computing variance and standard deviation.

It uses the method described in Donald Knuth’s Art of Computer Programming, Vol 2, page 232, 3rd edition to avoid these problems.  For further discussion see #4527

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -1464,12 +1464,12 @@ func intersection(ls1, ls2 labels.Labels) labels.Labels {
 }
 
 type groupedAggregation struct {
-	labels       labels.Labels
-	value        float64
-	varianceMean float64
-	groupCount   int
-	heap         vectorByValueHeap
-	reverseHeap  vectorByReverseValueHeap
+	labels      labels.Labels
+	value       float64
+	mean        float64
+	groupCount  int
+	heap        vectorByValueHeap
+	reverseHeap vectorByReverseValueHeap
 }
 
 // aggregation evaluates an aggregation operation on a Vector.
@@ -1542,6 +1542,7 @@ func (ev *evaluator) aggregation(op ItemType, grouping []string, without bool, p
 			result[groupingKey] = &groupedAggregation{
 				labels:     m,
 				value:      s.V,
+				mean:       s.V,
 				groupCount: 1,
 			}
 			inputVecLen := int64(len(vec))
@@ -1551,7 +1552,6 @@ func (ev *evaluator) aggregation(op ItemType, grouping []string, without bool, p
 			}
 			if op == itemStdvar || op == itemStddev {
 				result[groupingKey].value = 0.0
-				result[groupingKey].varianceMean = s.V
 			} else if op == itemTopK || op == itemQuantile {
 				result[groupingKey].heap = make(vectorByValueHeap, 0, resultSize)
 				heap.Push(&result[groupingKey].heap, &Sample{
@@ -1573,8 +1573,8 @@ func (ev *evaluator) aggregation(op ItemType, grouping []string, without bool, p
 			group.value += s.V
 
 		case itemAvg:
-			group.value += s.V
 			group.groupCount++
+			group.mean += (s.V - group.mean) / float64(group.groupCount)
 
 		case itemMax:
 			if group.value < s.V || math.IsNaN(group.value) {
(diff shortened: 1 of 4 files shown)


DanCech:
Verified the new tests fail on master:
```
promql_test.go:33: error running test testdata/functions.test: error in eval stdvar_over_time(metric[1m]): expected 0 for {} but got -4.440892098500626e-16
```
and with stdvar commented out:
```
promql_test.go:33: error running test testdata/functions.test: error in eval stddev_over_time(metric[1m]): expected 0 for {} but got NaN
```

promql/testdata/functions.test
```diff
@@ -385,6 +385,17 @@ eval instant at 1m stdvar_over_time(metric[1m])
...
 
+# Tests for stddev_over_time and stdvar_over_time #4927.
+clear
+load 10s
+  metric 1.5990505637277868 1.5990505637277868 1.5990505637277868
+
+eval instant at 1m stdvar_over_time(metric[1m])
+  {} 0
+
+eval instant at 1m stddev_over_time(metric[1m])
```
brian-brazil: This is duplicate

promql/testdata/functions.test
```diff
@@ -385,6 +385,17 @@ eval instant at 1m stdvar_over_time(metric[1m])
 eval instant at 1m stddev_over_time(metric[1m])
   {} 3.249615
 
+# Tests for stddev_over_time and stdvar_over_time #4927.
```
brian-brazil: We should probably add a test for stddev too

DanCech: I'm confused, this test both stdvar_over_time and stddev_over_time

brian-brazil: We also have the stddev and stdvar aggregators over in engine.go

DanCech:
Verified new stddev and stdvar tests fail on master:

```
promql_test.go:33: error running test testdata/aggregators.test: error in eval stddev(http_requests): expected 0 for {} but got NaN
```

```
promql_test.go:33: error running test testdata/aggregators.test: error in eval stdvar(http_requests): expected 0 for {} but got -5.684341886080802e-14
```

promql/engine.go
```diff
@@ -1464,12 +1464,12 @@ func intersection(ls1, ls2 labels.Labels) labels.Labels {
...
 type groupedAggregation struct {
-	labels           labels.Labels
-	value            float64
-	valuesSquaredSum float64
-	groupCount       int
-	heap             vectorByValueHeap
-	reverseHeap      vectorByReverseValueHeap
+	labels       labels.Labels
+	value        float64
+	varianceMean float64
```
brian-brazil: We're now calculating the mean and the mean for the stddev differently, which is a bit weird. Can you converge them?

DanCech: When you say converge them, would you like to see `varianceMean` renamed to `mean` and used directly for `itemAvg`? I suspect that will be less efficient than the current implementation as it will involve a division operation for each `s` rather than just once per `groupingKey`.  We require the intermediate mean values in `itemStddev` and `itemStdvar` which is why we perform the calculation that way, but it's not needed for `itemAvg`.

brian-brazil: > When you say converge them, would you like to see varianceMean renamed to mean and used directly for itemAvg? 

That's what I was thinking. There's existing benchmarks if you want to test it out, but I don't imagine it'll make much of a difference as I can't imagine more than 10k items being averaged at once.

DanCech: Do you also want to update `funcAvgOverTime` to make that consistent too?

It seems a little strange to introduce a (minor) performance regression (and imho make it more difficult for someone reading the code to understand it) just to use the same method of calculating the mean when we don't actually need the intermediate values.

brian-brazil: I'm okay with leaving that one, it's not all in the same function.

This method of calculating mean is also more resilient to floating point issues.

DanCech:
I went ahead and updated avg_over_time to keep things consistent across the system, and added a test case to cover it.

brian-brazil:
Thanks!

PR #4534, opened 2018-08-26
Adding unit tests for the util package

tariq1890:
Added unit tests for the util package

### DIFF
--- a/util/strutil/strconv_test.go
+++ b/util/strutil/strconv_test.go
@@ -47,3 +47,19 @@ func TestLink(t *testing.T) {
 		}
 	}
 }
+
+func TestSanitizeLabelName(t *testing.T) {
+	actual := SanitizeLabelName("fooClientLABEL")
+	expected := "fooClientLABEL"
+
+	if actual != expected {
+		t.Errorf("SanitizeLabelName failed for label (%s), want %s got %s", "fooClientLABEL", expected, actual)
+	}
+
+	actual = SanitizeLabelName("barClient.LABEL$$##")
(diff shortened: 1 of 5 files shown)


util/strutil/strconv_test.go
```diff
@@ -36,6 +41,17 @@ var linkTests = []linkTest{
 	},
 }
 
+var sanitizeTests = []sanitizeTest{
```
brian-brazil: Keep this with the function using it. Usually we're do these inline.

tariq1890:
@brian-brazi Thanks for the review. Done!

util/strutil/strconv_test.go
```diff
@@ -23,6 +23,11 @@ type linkTest struct {
 	expectedTableLink string
 }
 
+type sanitizeTest struct {
```
brian-brazil: You don't need this now

promql/engine_test.go
```diff
@@ -198,7 +199,7 @@ func TestQueryError(t *testing.T) {
 }
 
 func TestEngineShutdown(t *testing.T) {
-	engine := NewEngine(nil, nil, 10, 10*time.Second)
+	engine := NewEngine(nil, prometheus.DefaultRegisterer, 10, 10*time.Second)
```
brian-brazil: Why did you make this change? This could interfere with other tests.

tariq1890: My bad. I forgot to remove this. Thanks!

brian-brazil:
Thanks!

PR #4537, opened 2018-08-26
Test template expansion while loading groups

codesome:
Fixes https://github.com/prometheus/prometheus/issues/835

Currently if wrong templates are used for alerting rules, it will error out when alert fires, and not before that. This adds a check while the rules are being loaded.

### DIFF
--- a/pkg/rulefmt/rulefmt.go
+++ b/pkg/rulefmt/rulefmt.go
@@ -148,9 +148,6 @@ func (r *Rule) Validate() (errs []error) {
 
 // testTemplateParsing checks if the templates used in labels and annotations
 // of the alerting rules are parsed correctly.
-// NOTE: It wont detect if a wrong key is used with the variables, like `$labels.wrongKey`.
-//       This will check for wrong variables used, like `$wrongLabelsVariable`.
-//       Also wrong usage of template like {{$labels.quantile * 100}}, etc.
 func testTemplateParsing(rl *Rule) (errs []error) {
 	if rl.Alert == "" {
 		// Not an alerting rule.
@@ -158,12 +155,7 @@ func testTemplateParsing(rl *Rule) (errs []error) {
 	}
 
 	// Trying to parse templates.
-	tmplData := struct {
-		Labels map[string]string
-		Value  float64
-	}{
-		Labels: make(map[string]string),
-	}
+	tmplData := template.AlertTemplateData(make(map[string]string), 0)
 	defs := "{{$labels := .Labels}}{{$value := .Value}}"
 	parseTest := func(text string) error {
 		tmpl := template.NewTemplateExpander(
@@ -178,7 +170,7 @@ func testTemplateParsing(rl *Rule) (errs []error) {
 		return tmpl.ParseTest()
 	}
 
-	// Expanding Labels.
+	// Parsing Labels.
 	for _, val := range rl.Labels {
 		err := parseTest(val)
 		if err != nil {
@@ -186,7 +178,7 @@ func testTemplateParsing(rl *Rule) (errs []error) {
 		}
 	}
 
-	// Expanding Annotations.
+	// Parsing Annotations.
 	for _, val := range rl.Annotations {
 		err := parseTest(val)
 		if err != nil {
--- a/pkg/rulefmt/rulefmt.go
+++ b/pkg/rulefmt/rulefmt.go
@@ -14,11 +14,16 @@
 package rulefmt
 
 import (
+	"context"
+	"fmt"
 	"io/ioutil"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/pkg/timestamp"
 	"github.com/prometheus/prometheus/promql"
+	"github.com/prometheus/prometheus/template"
 	yaml "gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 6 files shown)


brian-brazil:
This may cause false positives due to passing in data which the actual template would never have to deal with (and thus be a breaking change). I think we may want to limit this to parsing, without execution.

codesome:
Can you give an example for more insight?

brian-brazil:
There might be a `query` based on passed labels that would always return something, which is using `first` but now fails as it's returning nothing.

codesome:
Thanks. I have made the changes.

rules/manager.go
```diff
@@ -747,6 +753,62 @@ func (m *Manager) loadGroups(interval time.Duration, filenames ...string) (map[s
...
+					ctx,
+					defs+text,
+					"__alert_"+rl.Alert,
+					tmplData,
+					model.Time(timestamp.FromTime(time.Now())),
+					template.QueryFunc(queryFunc),
+					nil,
+				)
+				err := tmpl.ParseTest()
+				return err
```
gouthamve: `return tmpl.ParseTest()`

rules/manager_test.go
```diff
@@ -651,3 +652,86 @@ func TestUpdate(t *testing.T) {
...
+	suite, err := promql.NewTest(t, "")
+	testutil.Ok(t, err)
+	queryFunc := EngineQueryFunc(suite.QueryEngine(), suite.Storage())
+
+	tests := []struct {
+		ruleString string
+		shouldPass bool
+	}{
+		{
+			ruleString: "" +
```
gouthamve: Can we use:
`
- name: example
  rules:
  - alert: InstanceDown
    expr: up == 0
    for: 5m
    labels:
      severity: "page"
    annotations:
      summary: "Instance {{ $labels.instance }} down"

`

In golang, `string with whatever` is considered a string (i.e, use ` at the beginning and end) https://stackoverflow.com/questions/7933460/how-do-you-write-multiline-strings-in-go

rules/manager.go
gouthamve: Further, can we print which alert is erroring so that it is easy to debug and fix for the user?

rules/manager_test.go
```diff
@@ -651,3 +652,90 @@ func TestUpdate(t *testing.T) {
...
+groups:
+- name: example
+  rules:
+  - alert: InstanceDown
+    expr: up == 0
+    for: 5m
+    labels:
+      severity: \"page\"
+    annotations:
+      summary: \"Instance {{ $labels.instance }} down\"
```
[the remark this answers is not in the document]
codesome: Forgot to remove the escape after using `, I updated it now.

gouthamve:
This should also be in `promtool` because prometheus would now fail on rules that were cleared by `promtool`.

codesome:
As discussed offline, best place to have the check would be `Rule.Validate()` in `rulefmt`. But we cannot use `"query"` function in template expander function map, which needs `ctx` and storage. Is `"query"` ever used in just parsing the template?

brian-brazil:
The query won't be called, that'd only be during execution.

codesome:
I moved template validation from `rules` package to `rulefmt`, in `Validate()`.

codesome:
@brian-brazil The error messages looks something like this
```
level=error ts=2018-08-28T15:56:34.530553452Z caller=manager.go:655 component="rule manager" msg="loading groups failed" err="group \"example\", rule 0, \"InstanceDown\": msg=template: __alert_InstanceDown:1: unexpected \"*\" in operand"
level=error ts=2018-08-28T15:56:34.530604882Z caller=main.go:635 msg="Failed to apply configuration" err="error loading rules, previous rule set restored"
```

```
level=error ts=2018-08-28T15:57:20.402622833Z caller=manager.go:655 component="rule manager" msg="loading groups failed" err="group \"example\", rule 0, \"InstanceDown\": msg=template: __alert_InstanceDown:1: undefined variable \"$labelsinstance\""
level=error ts=2018-08-28T15:57:20.402657667Z caller=main.go:635 msg="Failed to apply configuration" err="error loading rules, previous rule set restored"
```

Does this look right?

pkg/rulefmt/rulefmt.go
```diff
@@ -137,6 +142,58 @@ func (r *Rule) Validate() (errs []error) {
 		}
 	}
 
+	errs = append(errs, testTemplateParsing(r)...)
+	return errs
+}
+
+// testTemplateParsing checks if the templates used in labels and annotations
+// of the alerting rules are parsed correctly.
+// NOTE: It wont detect if a wrong key is used with the variables, like `$labels.wrongKey`.
```
brian-brazil: That's actually fine, that's what missing=zero handles.

pkg/rulefmt/rulefmt.go
```diff
@@ -137,6 +142,58 @@ func (r *Rule) Validate() (errs []error) {
...
+//       This will check for wrong variables used, like `$wrongLabelsVariable`.
+//       Also wrong usage of template like {{$labels.quantile * 100}}, etc.
+func testTemplateParsing(rl *Rule) (errs []error) {
+	if rl.Alert == "" {
+		// Not an alerting rule.
+		return errs
+	}
+
+	// Trying to parse templates.
+	tmplData := struct {
```
brian-brazil: We seem to be duplicating a fair bit of code here, so there might be drift. Could this be refactored to avoid that?

codesome: @brian-brazil I added a function for that in `template.go` and using that here and in `alerting.go`. Does it look right?

codesome:
@brian-brazil PTAL.

brian-brazil:
Thanks!

Issue #2585, opened 2017-04-06
Throttle resends of alerts


brian-brazil:
We currently send alerts at every eval interval. With a 1s eval interval, this would produce a lot of traffic to the AM.

I propose we continue to send new alerts instantly, but throttle the resends to say 60s by default.

brancz:
Generally I think that's ok, we just need to make it explicit in the documentation and make sure that the resolve timeout in the Alertmanager needs to be higher than this value.

brian-brazil:
The resolve timeout is just a default, we can adjust the end time as required inside Prometheus.

brancz:
Yes I realize that I just meant we need to document properly what the relationship between the two is.

reymont:
Any idea for the maximum alarm times? Reaching the "Throttle"(maximum alarm times), do not send alert message any more?

brian-brazil:
Please don't clutter the issue tracker with unrelated questions, use https://groups.google.com/forum/#!aboutgroup/prometheus-users for support.

stuartnelson3:
The current ingest path in AM for alerts has pretty high lock contention, throttling resends would be 👍

PR #4538, opened 2018-08-26
Throttle resends of alerts to 1 minute by default

csmarchbanks:
By default only resend alerts every 1 minute. If the alert is new, or just became resolved, the alert will be sent immediately.

Closes: https://github.com/prometheus/prometheus/issues/2585

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -89,6 +89,7 @@ func main() {
 		notifierTimeout     model.Duration
 		forGracePeriod      model.Duration
 		outageTolerance     model.Duration
+		resendDelay         model.Duration
 		web                 web.Options
 		tsdb                tsdb.Options
 		lookbackDelta       model.Duration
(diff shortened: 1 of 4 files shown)


cmd/prometheus/main.go
```diff
@@ -173,6 +174,9 @@ func main() {
 	a.Flag("rules.alert.for-grace-period", "Minimum duration between alert and restored 'for' state. This is maintained only for alerts with configured 'for' time greater than grace period.").
 		Default("10m").SetValue(&cfg.forGracePeriod)
 
+	a.Flag("rules.alert.resend-delay", "Minimum amount of time to wait before resending an alert to Alertmanager. Must be lower than resolve_timeout in Alertmanager").
```
stuartnelson3: :+1:, thanks for mentioning the resolve_timeout.

brian-brazil: We really should change Prometheus to send the end time appropriately, resolve_timeout should only be a fallback.

csmarchbanks: @brian-brazil Would you like that work done as part of this PR or a separate PR?

What would be the appropriate end time to send for an unresolved alert? Eval time + query staleness time?

brian-brazil: That'd be separate I'd say.

You'd want to leave a few eval intervals for safety, plus allow for this setting.

csmarchbanks: Sounds good. I will plan to work on that after this PR is merged.

brian-brazil:
Thanks!

Issue #4541, opened 2018-08-27
file_sd never stop update 'custom_sd.json' file


earthdiaosi:
## Proposal
**Use case. Why is this important?**
  I want to implement custom service discovery, so i read the follow blog and wirte a demo; 
  *  https://prometheus.io/blog/2018/07/05/implementing-custom-sd/
 
  but i found "custom_sd.json" file was updated each time. so i debug the follow [code](https://github.com/prometheus/prometheus/blob/master/documentation/examples/custom-sd/adapter/adapter.go#L59), and find `key` and `i` is different each time: 
```
// Parses incoming target groups updates. If the update contains changes to the target groups
// Adapter already knows about, or new target groups, we Marshal to JSON and write to file.
func (a *Adapter) generateTargetGroups(allTargetGroups map[string][]*targetgroup.Group) {
	tempGroups := make(map[string]*customSD)
	for k, sdTargetGroups := range allTargetGroups {
		for i, group := range sdTargetGroups {
			newTargets := make([]string, 0)
			newLabels := make(map[string]string)

			for _, targets := range group.Targets {
				for _, target := range targets {
					newTargets = append(newTargets, string(target))
				}
			}

			for name, value := range group.Labels {
				newLabels[string(name)] = string(value)
			}
			// Make a unique key, including the current index, in case the sd_type (map key) and group.Source is not unique.
			key := fmt.Sprintf("%s:%s:%d", k, group.Source, i)
			tempGroups[key] = &customSD{
				Targets: newTargets,
				Labels:  newLabels,
			}
		}
	}
	if !reflect.DeepEqual(a.groups, tempGroups) {
		a.groups = tempGroups
		err := a.writeOutput()
		if err != nil {
			level.Error(log.With(a.logger, "component", "sd-adapter")).Log("err", err)
		}
	}

}
```
  then i search code, fount the [reason](https://github.com/prometheus/prometheus/blob/master/discovery/manager.go#L198): the `tsets` is a map, so the results are different for each iteration;
```
func (m *Manager) allGroups() map[string][]*targetgroup.Group {
	m.mtx.Lock()
	defer m.mtx.Unlock()

	tSets := map[string][]*targetgroup.Group{}
	for pkey, tsets := range m.targets {
		for _, tg := range tsets {
			// Even if the target group 'tg' is empty we still need to send it to the 'Scrape manager'
			// to signal that it needs to stop all scrape loops for this target set.
[...]

PR #4542, opened 2018-08-27
fix file_sd never stop update 'custom_sd.json' file

earthdiaosi:
fixed: https://github.com/prometheus/prometheus/issues/4541


cstyan:
looking into the issue with reporter, but if there is a bug this is the wrong place to fix it imo

PR #4544, opened 2018-08-27
fix(Parser): Add error at parsing when quantile value isn't valid

### LINKED ISSUE #4543: [PromQL] Percentile parsing
## Proposal
When add an incorrect parameter for quantile function, a result is loaded

*Nice to have' is not a good use case :)*
You can test it with: "percentile(2.0, metric...)

## Bug Report
**What did you do?**
Just execute command below on Prometheus

An error message

A result (+inf)

**Environment**

* Prometheus version: 2.3.1

aurrelhebert:
Proposal to fix #4543


brian-brazil:
Scalars can come from expressions, and thus there's no requirement that they be parseable as floats - merely that they evaluate as scalars. I'm afraid that this change doesn't make sense.

brian-brazil:
This would introduce an inconsistency in PromQL, it's documented that we accept values outside of 0 and 1 and will return -Inf/+Inf for these. I'm afraid I cannot accept this change.

PR #4545, opened 2018-08-27
send updates only when there was a change in the targets.

krasi-georgiev:
this will limit the unnecessary CPU cicles when providers keep sending
updates that didn't actually change anything.

closes: https://github.com/prometheus/prometheus/issues/4518

see https://github.com/kubernetes/client-go/issues/454 for more details.

### DIFF
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -16,7 +16,6 @@ package discovery
 import (
 	"context"
 	"fmt"
-	"reflect"
 	"sync"
 	"time"
 
@@ -239,7 +238,7 @@ func (m *Manager) updateGroup(poolKey poolKey, tgs []*targetgroup.Group) (update
 			// "kube-scheduler" even though these don't actually include any targets.
 			// Even if we find a way to filter these in the k8s client
 			// this check is safer for other misbehaving providers.
-			if !reflect.DeepEqual(m.targets[poolKey][tg.Source], tg) {
+			if _, exist := m.targets[poolKey][tg.Source]; !exist || !m.targets[poolKey][tg.Source].Equal(tg) {
 				m.targets[poolKey][tg.Source] = tg
 				updated = true
 			}
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -16,6 +16,7 @@ package discovery
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"sync"
 	"time"
 
@@ -187,16 +188,15 @@ func (m *Manager) updater(ctx context.Context, p *provider, updates chan []*targ
 
 			}
 			for _, s := range p.subs {
-				m.updateGroup(poolKey{setName: s, provider: p.name}, tgs)
-			}
-			if m.updateGroup(poolKey, tgs) {
-				select {
-				case triggerUpdate <- struct{}{}: // Signal that there was an update.
-				default:
+				if m.updateGroup(poolKey{setName: s, provider: p.name}, tgs) {
+					select {
+					case triggerUpdate <- struct{}{}: // Signal that there was an update.
+					default:
+					}
 				}
 			}
 
-		case <-ticker.C: // Some discoverers send updates too often so we send these to the receiver once every 5 seconds.
+		case <-ticker.C: // Some discoverers send updates too often so we throttle these with the ticker.
 			select {
 			case <-triggerUpdate:
 				select {
@@ -223,6 +223,8 @@ func (m *Manager) cancelDiscoverers() {
 	m.discoverCancel = nil
 }
 
+// updateGroup updates or creates targets.
+// Returns false if nothing changed.
 func (m *Manager) updateGroup(poolKey poolKey, tgs []*targetgroup.Group) (updated bool) {
 	m.mtx.Lock()
 	defer m.mtx.Unlock()
@@ -233,11 +235,6 @@ func (m *Manager) updateGroup(poolKey poolKey, tgs []*targetgroup.Group) (update
 				m.targets[poolKey] = make(map[string]*targetgroup.Group)
 			}
 
-			// No need to update when there aren't any actual changes.
-			// The k8s discoverer keeps sending updates for "kube-controller-manager" and
-			// "kube-scheduler" even though these don't actually include any targets.
-			// Even if we find a way to filter these in the k8s client
-			// this check is safer for other misbehaving providers.
 			if _, exist := m.targets[poolKey][tg.Source]; !exist || !m.targets[poolKey][tg.Source].Equal(tg) {
 				m.targets[poolKey][tg.Source] = tg
 				updated = true
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -190,12 +190,14 @@ func (m *Manager) updater(ctx context.Context, p *provider, updates chan []*targ
 			for _, s := range p.subs {
 				m.updateGroup(poolKey{setName: s, provider: p.name}, tgs)
 			}
-
-			select {
-			case triggerUpdate <- struct{}{}:
-			default:
+			if m.updateGroup(poolKey, tgs) {
+				select {
+				case triggerUpdate <- struct{}{}: // Signal that there was an update.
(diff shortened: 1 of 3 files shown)


discovery/manager.go
```diff
@@ -190,7 +186,17 @@ func (m *Manager) updateGroup(poolKey poolKey, tgs []*targetgroup.Group) {
 			if _, ok := m.targets[poolKey]; !ok {
 				m.targets[poolKey] = make(map[string]*targetgroup.Group)
 			}
-			m.targets[poolKey][tg.Source] = tg
+
+			// No need to update when there aren't any actual changes.
+			// The k8s discoverer keeps sending updates for "kube-controller-manager" and "kube-scheduler" even though these don't actually include any targets.
```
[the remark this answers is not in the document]
krasi-georgiev: updated

grobie:
I don't know enough about all the consumer implementations of the manager, and whether some might need such seemingly unnecessary updates to work correctly. Would be good to get another pair of eyes on this @brancz. Also I'd have expected a test to ensure this behavior (duplicates don't get propagated).

krasi-georgiev:
@grobie thanks for the quick feedback. Spot on for both suggestions!

> Instead of adding another channel to the manager struct and increasing the coupling between runProvider and updateGroup functions, I'd have exposed the update status as return value: updateGroup(poolKey poolKey, tgs []*targetgroup.Group) (updated bool).

updated


> Also I'd have expected a test to ensure this behavior (duplicates don't get propagated).

keep forgetting about the tests :stuck_out_tongue_winking_eye:  . Will add one soon.

discovery/manager.go
```diff
@@ -190,9 +191,19 @@ func (m *Manager) updateGroup(poolKey poolKey, tgs []*targetgroup.Group) {
...
 				m.targets[poolKey] = make(map[string]*targetgroup.Group)
 			}
-			m.targets[poolKey][tg.Source] = tg
+
+			// No need to update when there aren't any actual changes.
+			// The k8s discoverer keeps sending updates for "kube-controller-manager" and
+			// "kube-scheduler" even though these don't actually include any targets.
+			// Even if we find a way to filter these in the k8s client
+			// this check is safer for other misbehaving providers.
+			if !reflect.DeepEqual(m.targets[poolKey][tg.Source], tg) {
```
brancz: `reflect.DeepEqual` seems excessive and maybe even incorrect here, as `DeepEqual` returns false if the ordering of `.Targets` has changed within the `Group`, but that does not actually make a difference. I'd propose to implement an `Equal` method on the `Group`.

As a side note, those events are generated (for the better or worse) because the controller manager and scheduler use `Endpoints` objects to perform leader election. They change an annotation to signify that they are still alive, which results in an even for Prometheus. There's nothing we can do and technically this "works as intended" on the Kubernetes side.

krasi-georgiev: I was also crossed for the "DeepEqual" so added an Equal method, but without any sorting for the Targets. Sorting maps would kill all performance gains so the comparison will only work when the targets arrive in the same order. Otherwise we just send a new update.

krasi-georgiev:
@grobie added a new test. Not 100% what I wanted, but a better test would mean somehow exposing the `updateReceived` channel so that the test can listen if it gets notified for a new update.

discovery/targetgroup/targetgroup.go
```diff
@@ -91,3 +91,26 @@ func (tg *Group) UnmarshalJSON(b []byte) error {
 	tg.Labels = g.Labels
 	return nil
 }
+
+// Equal returns true if all fields have exactly the same values and lenghts.
+// To keep a good performance the `Targets []model.LabelSet` is not sorted
+// so if these arrive in a different order the comparing won't work.
```
brancz: I think "if .Targets of the compared objects are of different order they are not equal", would be clearer, but I'm also unsure whether this semantic might be a problem, as I don't think we ask provider implementations to have this consistent, just some (like Kubernetes) happen to report targets in a consistent order.

discovery/manager.go
```diff
@@ -190,9 +190,19 @@ func (m *Manager) updateGroup(poolKey poolKey, tgs []*targetgroup.Group) {
 			if _, ok := m.targets[poolKey]; !ok {
 				m.targets[poolKey] = make(map[string]*targetgroup.Group)
 			}
-			m.targets[poolKey][tg.Source] = tg
+
+			// No need to update when there aren't any actual changes.
+			// The k8s discoverer keeps sending updates for "kube-controller-manager" and
```
brancz: We should drop this explanation about a specific target provider and just keep the first line. If we really want to keep the Kubernetes example we should mention why Kubernetes does this.

discovery/targetgroup/targetgroup.go
krasi-georgiev: another way to  deal with this is to filter these in the k8s provider code, but not sure if another provider might misbehave in the same way.

brancz: Filtering them would be incorrect, the events rightfully end up here, it just happens to be a side effect, it could just as well be a valid update.

grobie:
I'm a bit confused by the discussion, if the annotations of the k8s targets change, will this change ever find any duplicate updates? Shouldn't these all differ in their annotation labels?

discovery/targetgroup/targetgroup.go
krasi-georgiev: `SelfLink:/api/v1/namespaces/kube-system/endpoints/kube-controller-manager`

can this actually include some targets to be scraped? if yes than end of discussion can't filter these.

About the slice order, implementing to order the maps would mean many iterations in a loop so it would be probably worse than just allowing the duplicate updates.

krasi-georgiev:
@grobie  can you include an example as I don't understand the question.

We are discussing the even when receiving 2 identical target groups that have the slice order in a different order we can't catch these as duplicates without ordering the slice(which are maps and sorting these would be too heavy)

https://github.com/krasi-georgiev/prometheus/blob/master/discovery/targetgroup/targetgroup.go#L27

grobie:
My question was more fundamental: do we really receive 2 completely identical target groups? As I understood @brancz explanation above, one annotation label constantly changes, or is this not correct?

krasi-georgiev:
@grobie 

> As a side note, those events are generated (for the better or worse) because the controller manager and scheduler use Endpoints objects to perform leader election. They change an annotation to signify that they are still alive, which results in an even for Prometheus. There's nothing we can do and technically this "works as intended" on the Kubernetes side.

at my observation the only thing that changes is "renewTime" and "ResourceVersion" and we don't care about these so once they hit the Discovery manager any of these are detected as duplicates.

brancz:
Yes and I've also seen those services used to actually discover the controller-manager and scheduler so events from those objects can potentially be real updates, and the resourceVersion changing implies that the apiserver actually accepted a real (non-nop) change, so we can't ignore it.

krasi-georgiev:
@brancz @grobie any final thoughts on this?

simonpasquier:
IMO this PR relies on something (eg sorted labels) that wasn't explicitly expected from the `Discoverer` interface. 

https://github.com/prometheus/prometheus/blob/abf6fe0a983ca116f1c3be8875193458174c6748/discovery/manager.go#L56-L63

It also says `Whenever a discovery provider detects a potential change, it sends the TargetGroup through its channel` which I read like "the manager will forward whatever it receives from the provider".

We could enforce the sorting for all built-in SD providers but now that we encourage people to build their own service discovery, it could be breaking for them.

brancz:
I agree with @simonpasquier. How about instead of checking for equality we hash all the targets with fnv and then hash all those hashes together again, then we have a single comparison value that is not order sensitive.

krasi-georgiev:
after some brainstorming came up with this solution. will run a benchmark to see if it will actually make any performance gains.

krasi-georgiev:
based on the benchmarking I don't see any gains so at this point unless anyone has other ideas I am closing this and can revisit again at some point if needed.

PR #4546, opened 2018-08-27
Add SelectParams to Select for federation

jacksontj:
When prom2 came out the storage querier interface consolidated to a
single Select() method. While doing this it makes it impossible as the
implementer of the querier to know if you are being called for metadata
or actual data. The workaround has been to check if the SelectParams are
nil, which the federation call is always nil. This has 2 negative
consequences (1) remote implementations interpret this as a metadata
call, which makes the federation endpoint return nothing. (2) this means
that the storage implementations don't get the same information passed
down to them as far as SelectParams goes.

This diff simply adds SelectParams to the Select() call in the
federation handler

Mitigation for #4057

### DIFF
--- a/web/federate.go
+++ b/web/federate.go
@@ -76,9 +76,14 @@ func (h *Handler) federation(w http.ResponseWriter, req *http.Request) {
 
 	vec := make(promql.Vector, 0, 8000)
 
+	params := &storage.SelectParams{
+		Start: mint,
+		End:   maxt,
+	}
+
 	var sets []storage.SeriesSet
 	for _, mset := range matcherSets {
-		s, err := q.Select(nil, mset...)
+		s, err := q.Select(params, mset...)
 		if err != nil {
 			federationErrors.Inc()
 			http.Error(w, err.Error(), http.StatusInternalServerError)


brian-brazil:
Hmm, to be honest I don't think federation should be using remote storage at all - it should only query local storage. That we blindly use remote storage everywhere when querying is something that is on the todo list to be fixed.

jacksontj:
Until that is fixed I think this would be a good one to add in. As it stands today it just comes back blank, which isn't right either :/

jacksontj:
To be clear, this isn't specifically a remote-storage issue, this is actually a web API issue. To create the web endpoint you only need to pass something that implements the querier interface. The issue that this patch is handling is that those `Select()` calls aren't being passed the `SelectParams`. In my specific case this is especially bad because I'm using my own implementation of the `Querier` interface, but even in regular prom this means that queries to TSDB don't include `SelectParams` -- which makes these queries different than others-- which isn't good. We should pass the `SelectParams` always when we do the query-- and if we want to *not* call some storage engines then we can do so, but any engine we call should get the same set of args.

PR #4550, opened 2018-08-28
Send EndsAt along with alerts

csmarchbanks:
Send a reasonable value for EndsAt to the Alertmanager for alerts. Since either eval loops, or sending to alertmanager can fail, allow for a failure of whichever one happens less often before marking an alert as resolved.

As discussed: https://github.com/prometheus/prometheus/pull/4538#discussion_r213025949

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -174,7 +174,7 @@ func main() {
 	a.Flag("rules.alert.for-grace-period", "Minimum duration between alert and restored 'for' state. This is maintained only for alerts with configured 'for' time greater than grace period.").
 		Default("10m").SetValue(&cfg.forGracePeriod)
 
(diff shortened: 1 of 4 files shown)


brian-brazil:
This looks okay to me. Would you mind doing a quick check against a 0.13.0 AM or older to make sure nothing breaks? There was a change in 0.14.0 that touched this, though I believe it just affects a metric.

csmarchbanks:
Tested against 0.13.0, still got an alert with a reasonable end time value:
```
"startsAt":"2018-08-28T08:39:13.637186683-06:00"
"endsAt":"2018-08-28T08:43:14.637186683-06:00"
```
This was from the second send, so a 4 minute difference between start and end makes sense.

brian-brazil:
Thanks!

PR #4554, opened 2018-08-28
web: clean up api/v2

simonpasquier:
The `api_v2.API` struct have some fields that aren't used anywhere.


gouthamve:
Wow, nice change! 👍

krasi-georgiev:
if all test pass I would say it is safe to assume that these aren't really used anywhere :+1: 
I think that was a copy and paste error when first adding the v2 api 

LGTM

PR #4555, opened 2018-08-28
fix bug in SD adapter that would result in writing the output even if target groups had not changed

cstyan:
fixes:  #4541 

use the previous data + a hash of the target group as a key now
afaict this was a result of not accounting for maps being unsorted 


cstyan:
see #4567

Issue #4551, opened 2018-08-28
Race condition starting HEAD prometheus with static scrape configs


csmarchbanks:
## Bug Report
**What did you do?**
Ran prometheus locally, with some basic static configs

**What did you expect to see?**
Prometheus comes up and my targets are scraped. 

**What did you see instead? Under which circumstances?**
Most of the time none of my static targets are scraped. Approximately 5-10% of the time my targets are scraped properly.

**Environment**

* System information:

	Darwin 17.7.0 x86_64

* Prometheus version:

	latest master (87f1dad16da760aa48f8a41089c0d1a20b56689c)

* Prometheus configuration file:
```
global:
  scrape_interval:     15s
  evaluation_interval: 15s

scrape_configs:
  - job_name: 'prometheus'
    static_configs:
      - targets: ['localhost:9090']
  - job_name: 'node-exporter'
    static_configs:
      - targets: ['localhost:9100']
```

* Logs:
```
$ go run cmd/prometheus/main.go --config.file=prometheus.yml --log.level=debug
level=info ts=2018-08-28T03:39:41.803111877Z caller=main.go:235 msg="Starting Prometheus" version="(version=, branch=, revision=)"
level=info ts=2018-08-28T03:39:41.803208011Z caller=main.go:236 build_context="(go=go1.10, user=, date=)"
level=info ts=2018-08-28T03:39:41.803227773Z caller=main.go:237 host_details=(darwin)
level=info ts=2018-08-28T03:39:41.803248042Z caller=main.go:238 fd_limits="(soft=4864, hard=unlimited)"
level=info ts=2018-08-28T03:39:41.803265702Z caller=main.go:239 vm_limits="(soft=unlimited, hard=unlimited)"
level=info ts=2018-08-28T03:39:41.803878491Z caller=main.go:551 msg="Starting TSDB ..."
level=info ts=2018-08-28T03:39:41.80393762Z caller=web.go:395 component=web msg="Start listening for connections" address=0.0.0.0:9090
level=info ts=2018-08-28T03:39:41.831430515Z caller=main.go:561 msg="TSDB started"
level=info ts=2018-08-28T03:39:41.831492958Z caller=main.go:621 msg="Loading configuration file" filename=prometheus.yml
level=info ts=2018-08-28T03:39:41.832040894Z caller=main.go:647 msg="Completed loading of configuration file" filename=prometheus.yml
[...]

PR #4556, opened 2018-08-28
tidy up the discovery logs,updating loops and selects

krasi-georgiev:
few objects renamings

removed a very noise debug log on the k8s discovery. It would be usefull
to show some summary rather than every update as this is impossible to
follow.

added most comments as debug logs so each block becomes self
explanatory.

when the discovery receiving channel is full will retry again on the
next cycle. fixes: https://github.com/prometheus/prometheus/issues/4551

### DIFF
--- a/discovery/kubernetes/kubernetes.go
+++ b/discovery/kubernetes/kubernetes.go
@@ -383,7 +383,6 @@ func send(ctx context.Context, l log.Logger, role Role, ch chan<- []*targetgroup
 	if tg == nil {
 		return
 	}
-	level.Debug(l).Log("msg", "kubernetes discovery update", "role", string(role), "tg", fmt.Sprintf("%#v", tg))
 	select {
 	case <-ctx.Done():
 	case ch <- []*targetgroup.Group{tg}:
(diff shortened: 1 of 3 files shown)


krasi-georgiev:
@csmarchbanks can you confirm if this works as expected now?

the main change is to trigger a new update when the receiver fails to accept the updates.
https://github.com/prometheus/prometheus/pull/4556/files#diff-8d649638bc4ce3d86e318a4a0bd1abfdR169

krasi-georgiev:
Thanks , updated.

> Next step is to improve the unit tests as you mentioned.

yeah maybe it would be even best to add it to this PR, but  I won't be able to do it until next week.
@simonpasquier if you have an idea how to add test feel free to open a PR against my branch(krasi-georgiev:retry-when-failed-update) and I will review.

issue https://github.com/prometheus/prometheus/issues/4565

gouthamve:
👍 As master is essentially broken now, I don't think it makes sense to block this anymore. I'd rather get this in sooner rather than later, and have a clear issue for the test.

@simonpasquier Thoughts?

simonpasquier:
@krasi-georgiev yep, working on it.

simonpasquier:
@gouthamve agree, I would merge it now.

gouthamve:
Thanks! You should still write that test ;)

Issue #4356, opened 2018-07-06
`promql.parseSeriesDesc` does not parse negative values


kevinjqiu:
## Bug Report
**What did you do?**

I'm building a [tool](https://github.com/kevinjqiu/pat) that piggy-backs on the promql's testing facility to allow custom test cases for alert rules to be generated from a user defined yaml file.

However, it appears that `promql.parseSeriesDesc` function does not parse negative values correctly, e.g.,

`parseSeriesDesc("my_metric{a=\"b\"} 1 -1")` - I expect to see the parsed values to be `[1, -1]`, however, it throws a parser error:

```
could not parse: parse error at char 22: unexpected end of input in series values, expected "x"
```

If the negative value appears first in the list, the series can be correctly parsed:

`parseSeriesDesc("my_metric{a=\"b\"} -1 1)` <- this parses correctly.

I have written [a test case](https://github.com/prometheus/prometheus/compare/master...kevinjqiu:negative-series-value-parser-error) that demonstrated this parser error, and here's the test output:

```
=== RUN   TestParseSeries
--- FAIL: TestParseSeries (0.00s)
	parse_test.go:1778: error in input: 
		
		my_metric{a="b"} 1 -1
	parse_test.go:1779: could not parse: parse error at char 22: unexpected end of input in series values, expected "x"
FAIL
```

EDIT: Add link to the test code that demonstrate this bug

codesome:
Adding to above description, this happens not only for `-`, but also for `+`.  `'1 +1'` would also error out.

Basically for `-` or `+`, the parser is expecting notation like `1 -1x3`, `1 +1x4` (except for the first number).

PR #4564, opened 2018-08-30
[bugfix] Parse negative value in PromQL

codesome:
Fixes https://github.com/prometheus/prometheus/issues/4356

The issue was that, after seeing `+` or `-` parser was expecting `num [+-] num x num` notation and not looking for individual values. 

Signed-off-by: Ganesh Vernekar <cs15btech11018@iith.ac.in>


codesome:
@brian-brazil PTAL

codesome:
Unfortunately yes. That would be hard to tackle as we ignore white spaces.

brian-brazil:
Hmm, given this is going to be a publicly exposed api we may want to change that.

codesome:
So don't ignore white spaces in lexer and handle them in parsing?

brian-brazil:
We should probably consider it for this bit of parsing.

codesome:
@brian-brazil I have made the changes to consider spaces between values.

brian-brazil:
Thanks!

Issue #4509, opened 2018-08-16
PROPOSAL: Add tracing backend support


gouthamve:
Now that we have some spans in prometheus now, I think it's time we added some backend support to prometheus. This would help people see how each of their query performs. We run cortex-querier with sampling rate as `1` and make sure to get a trace for every query.

As the number of queries hitting prometheus would be fairly low, it works and also gives us an insight into which queries take how long. I'm sure people would love the same information for the prometheus server.

We know users want it and ideally would love it in logs: https://github.com/prometheus/prometheus/issues/1315 but this could be an alternate avenue for them until we figure out the logging situation.

/cc @brancz also expressed interest in this.

brian-brazil:
I think it'd be useful to have an in-memory backend so users could view both recent and long queries. I'm not sure at all about remote backends, as that seems like we'd end up having to support a pile of systems.

gouthamve:
Hmm, fair point. Could we just support jaeger and stop at that? Seems implausible but would be great. Implementing an in-memory backend looks a lot more complex, but would be great to have. @tom_wilkie, you were looking into this last, how complex is the in-mem backend?

We could also support OpenCensus, which should solve the too many backends problem, but would require instrumenting everything again with it along side open-tracing. Sigh.

brian-brazil:
Why should we support jaeger and not anything else? Is there some generic protocol we can support?

Given that OpenTracing is a code-level API, we may have to leave this up to users to compile in themselves as needed.

tomwilkie:
>  @tomwilkie , you were looking into this last, how complex is the in-mem backend?

Non-trivial, but not hard.  We really just want enough to render a simple /tracez style page.  I have a branch somewhere, but realistically I won't have time to see this through.

>  We could also support OpenCensus, which should solve the too many backends problem, but would require instrumenting everything again with it along side open-tracing. Sigh.

There was some movement a while back on an opentracing to opencensus bridge, that would allow us to stick with opentracing and get the z pages from opencensus.  Don't know where it went though: https://github.com/census-instrumentation/opencensus-go/pull/505

> Is there some generic protocol we can support?

There is not, see https://github.com/opentracing/specification/issues/64

> Given that OpenTracing is a code-level API, we may have to leave this up to users to compile in themselves as needed.

[...]

PR #4568, opened 2018-09-01
Add support for Opencensus z-pages for in-memory tracing

sipian:
#4509 

This PR replaces [opentracing](https://opentracing.io ) tracing spans with [opencensus](https://opencensus.io/) tracing spans.
As discussed in #4509 , this PR adds an in-memory backend in Prometheus, in the form of [opencensus z-pages](https://opencensus.io/core-concepts/z-pages/).

A running Prometheus server can be found at http://35.237.43.34/graph.
It's zpages can be found at http://35.237.43.34/trace/tracez.
To see traces of alerting-rules `(span-name:rule)`, an alerting-rule is also added http://35.237.43.34/alerts.

The load on Prometheus is similar to the one used by [prombench](https://github.com/prometheus/prombench/blob/master/components/prombench/manifests/benchmark/3_prometheus-test.yaml#L7).

### govendor details
- Opentracing packages are removed
- Opencensus is compatible with [grpc-v1.6.0](https://github.com/grpc/grpc-go/tree/v1.6.x). Prometheus used grpc-v1.5.0
- Opencensus is compatible with [grpc/metadata-v1.10.0](https://github.com/grpc/grpc-go/tree/v1.10.x/metadata)

/cc @brian-brazil @gouthamve @tomwilkie @brancz


sipian:
I have opened an issue https://github.com/census-instrumentation/opencensus-go/issues/893
in opencensus about the `promu crossbuild -v` error in `ci/circleci:build`

codesome:
http://35.198.182.203/trace/tracez?zspanname=GET%20%2ftrace%2fpublic%2fopencensus.css&ztype=2&zsubtype=0

Seems like `/trace/public/opencensus.css` not found.

sipian:
I doubt it.
The sample page given [here](https://opencensus.io/core-concepts/z-pages/) is similar to the one deployed.

tomwilkie:
:-1:

We're pretty dependant on the existing opentracing instrumentation.  If there are way we can have an adapter from one to the other?

gouthamve:
I've spoken @sipian offline, and the `zpages` don't provide us the trace, just the traceid. We'll still need to have a backend to view it. With that, I wonder if opencensus is worth the switch.

I think he's currently investigating building an in-mem backend. If the effort is the same, I'd prefer if we didn't make the switch to opencensus.

brancz:
I was playing around with this, and frankly I don't see the use if we just see the trace IDs, I was actually wildly clicking around the UI expecting to see more.

sipian:
@tomwilkie @brancz 
opencensus z-pages do not provide a complete view of the traces (as seen in Jaeger).
It seems implementing an in-memory backend is the only option left. 
(I didn't find any existing tool, which support a complete in-memory visualization of traces)

bboreham:
OpenTracing is an implementation-independent API - it seems likely there could be an adapter behind the scenes.

Having looked at the [sample page](http://35.198.182.203/trace/tracez) (and figured out you have to scroll to the bottom after clicking on something) I would have found it really useful, before we (Weaveworks) set up Jaeger. But now we have machinery in place to control Jaeger sampling I don't particularly want to re-do that for OpenCensus.

simonpasquier:
Since [go 1.11](https://golang.org/doc/go1.11#trace), the `runtime/time` package from go 1.11 has a [user annotation](https://golang.org/pkg/runtime/trace/#hdr-User_annotation) API.

I've hacked something quickly (see https://github.com/simonpasquier/prometheus/commit/db97bfbbbc193928ff4de43c293521aac62bcfb1) and it is quite promising. You need to download a trace profile and pass it to `go tool trace` instead of having the information available in the Prometheus UI but overall it provides the same level of information.

## Summary page with all user-defined tasks

image

## List of all traces for one task 

image

## Detailed view of a trace

image

brian-brazil:
That looks promising. It's better than nothing, doesn't require much code, and should mainly cover the need. Hopefully we can get something in the server itself at some point.

sipian:
@simonpasquier 
:+1: I'll try this out and open a PR.

gouthamve:
I really like this proposal, but this is not user friendly at all. The whole point was for users to find out slow queries and asking them to `pprof` and run this tool for finding it is a little too much.

I think this is orthogonal to the issue and still merits adding it to prometheus, but doesn't solve the problem.

PR #4569, opened 2018-09-03
Adding support for multiple azure environments

tariq1890:
Currently, the azure discovery assumes the environment as AzurePublicCloud. Other supported Azure Environments won't with the current code.

This PR changes that.  Minor code improvements have been made as well.

### DIFF
--- a/discovery/azure/azure.go
+++ b/discovery/azure/azure.go
@@ -65,11 +65,13 @@ var (
 	DefaultSDConfig = SDConfig{
 		Port:            80,
 		RefreshInterval: model.Duration(5 * time.Minute),
+		Environment:     azure.PublicCloud.Name,
 	}
 )
 
 // SDConfig is the configuration for Azure based service discovery.
 type SDConfig struct {
+	Environment     string             `yaml:"environment,omitempty"`
 	Port            int                `yaml:"port"`
 	SubscriptionID  string             `yaml:"subscription_id"`
 	TenantID        string             `yaml:"tenant_id,omitempty"`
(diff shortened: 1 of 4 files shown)


discovery/azure/azure.go
```diff
@@ -157,29 +159,45 @@ type azureClient struct {
 	vmssvm compute.VirtualMachineScaleSetVMsClient
 }
 
+// getAzureEnvironment returns an instance of azure.Environment based on envName.
+// If an invalid/unrecognized value of envName is passed, azurePublicCloud is returned by default.
```
brian-brazil: This should be an error, not silently failing.

tariq1890:
@brian-brazil Thanks for the review. I will fix it in the morning. Is there any other comment that you for the PR?

tariq1890:
Done!

docs/configuration/configuration.md
```diff
@@ -272,6 +272,8 @@ See below for the configuration options for Azure discovery:
 
 ```yaml
 # The information to access the Azure API.
+# The Azure environment.
+environment: <string>
```
grobie: This seems to be `[ environment: <string> | default = "AzurePublicCloud" ]` :-)

tariq1890: oh yes :(. Thanks for pointing it out.

discovery/azure/azure.go
```diff
@@ -157,29 +159,48 @@ type azureClient struct {
 	vmssvm compute.VirtualMachineScaleSetVMsClient
 }
 
+// getAzureEnvironment returns an instance of azure.Environment based on envName.
+func getAzureEnvironment(envName string) (azure.Environment, error) {
+	env, err := azure.EnvironmentFromName(envName)
```
grobie: The next 4 lines are the same as `return azure.EnvironmentFromName(envName)`. In the end I don't think we need this function, just use `azure.EnvironmentFromName(cfg.Environment)` below in `createAzureClient()`.

tariq1890: Ah yes. Done

PR #4575, opened 2018-09-05
[ServiceDiscovery] Unregister source when the target is empty

### LINKED ISSUE #4573: [Custom_sd] Impossible to remove targets from the discovery
## Bug Report

**What did you do?**
Write code to implement custom service discovery;

I expect that the targets that don't exist are removed in the custom_sd.json.

When a target is removed in the custom discovery, there is no way to remove the targetGroup in the updated channel 

**Environment**

* System information:

  * Windows10 64bit, 
  * Darwin 16.7.0 x86_64

* Prometheus version:

2.3.2

* GO Version :

  * 1.10.1 windows/amd64
  * 1.11 darwin/amd64

## Demo

I write a short demo to reproduce the bug

```golang
type File struct {
	Source      string `yaml:"source"`
	Address     string `yaml:"address"`
	Application string `yaml:"application"`
}

func readFile() ([]*File, error) {
	b, err := ioutil.ReadFile("./data.yaml")

	if err != nil {
		return nil, err
	}
	var res []*File
	err = yaml.UnmarshalStrict(b, &res)
	return res, err
}

func (d *discovery) parseServiceNodes(f *File) *targetgroup.Group {
	tgroup := targetgroup.Group{
		Source: f.Source,
		Labels: make(model.LabelSet),
	}

	tgroup.Targets = make([]model.LabelSet, 0, 1)

	target := model.LabelSet{model.AddressLabel: model.LabelValue(f.Address)}
	labels := model.LabelSet{
		model.AddressLabel:                    model.LabelValue(f.Address),
		model.LabelName("__meta_application"): model.LabelValue(f.Application),
		model.LabelName("__meta_source"):      model.LabelValue(f.Source),
	}

	tgroup.Labels = labels
	tgroup.Targets = append(tgroup.Targets, target)
	return &tgroup
}

func (d *discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
	for c := time.Tick(time.Duration(d.refreshInterval) * time.Second); ; {
		file, err := readFile()
		if err != nil {
			level.Error(d.logger).Log("msg", "Error reading services list", "err", err)
			time.Sleep(time.Duration(d.refreshInterval) * time.Second)
			continue
		}

		var tgs []*targetgroup.Group

		for _, f := range file {

			tgs = append(tgs, d.parseServiceNodes(f))
		}

		ch <- tgs
		// Wait for ticker or exit when ctx is closed.
		select {
		case <-c:
			continue
		case <-ctx.Done():
			return
		}
	}
}
```

The data.yaml file is "a service" that giving me all the targets that should be scraped.

When the code is running change the source in the file "data.yaml" like the following step :

1. With this data.yaml :

```yaml
- source: "source2"
  address: "localhost:1954"
  application: "PYM"
```

We have this custom_sd.json : 

```json
[
    {
        "targets": [
            "localhost:1954"
        ],
        "labels": {
            "__address__": "localhost:1954",
            "__meta_application": "PYM",
            "__meta_source": "source2"
        }
    }
]
```

2. Now I just change the source in the file data.yaml: 

```yaml
- source: "source1"
  address: "192.168.1.12:9090"
  application: "FOO"
```

And I have this file as result : 

```json
[
    {
        "targets": [
            "localhost:1954"
        ],
        "labels": {
            "__address__": "localhost:1954",
            "__meta_application": "PYM",
            "__meta_source": "source2"
        }
    },
    {
        "targets": [
            "192.168.1.12:9090"
        ],
        "labels": {
            "__address__": "192.168.1.12:9090",
            "__meta_application": "FOO",
            "__meta_source": "source1"
        }
    }
]
```

Instead of having this one : 

```json
[
    {
        "targets": [
            "192.168.1.12:9090"
        ],
        "labels": {
            "__address__": "192.168.1.12:9090",
            "__meta_application": "FOO",
            "__meta_source": "source1"
        }
    }
]
```

## Crappy Workaround:

The only way that I found to remove the target in the file is to send an update of the targetGroup.source with all the other parameter to nil. But it's creating an empty target group as below

```json
[
    {
        "targets": [],
        "labels": {}
    },
    {
        "targets": [
            "192.168.1.12:9090"
        ],
        "labels": {
            "__address__": "192.168.1.12:9090",
            "__meta_application": "FOO",
            "__meta_source": "source1"
        }
    }
]
```

Nexucis:
close #4573 

We could do a similar things in the Adapter (in the method generateTargetGroups for example),  but it's cleaner to unregister the target in the Manager level. In this way it wil reduce he size of the map `targets`

### DIFF
--- a/documentation/examples/custom-sd/adapter-usage/main.go
+++ b/documentation/examples/custom-sd/adapter-usage/main.go
@@ -200,7 +200,7 @@ func (d *discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
 			tgs = append(tgs, tg)
 			newSourceList[tg.Source] = true
 		}
-		// when targetGroup disappear, send an update with empty targetList
+		// When targetGroup disappear, send an update with empty targetList.
 		for key := range d.oldSourceList {
 			if !newSourceList[key] {
 				tgs = append(tgs, &targetgroup.Group{
--- a/documentation/examples/custom-sd/adapter-usage/main.go
+++ b/documentation/examples/custom-sd/adapter-usage/main.go
@@ -94,6 +94,7 @@ type discovery struct {
 	clientDatacenter string
 	tagSeparator     string
 	logger           log.Logger
+	oldSourceList    map[string]bool
 }
 
 func (d *discovery) parseServiceNodes(resp *http.Response, name string) (*targetgroup.Group, error) {
@@ -180,6 +181,8 @@ func (d *discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
 		// list of targets simply because there may have been a timeout. If the service is actually
 		// gone as far as consul is concerned, that will be picked up during the next iteration of
 		// the outer loop.
+
+		newSourceList := make(map[string]bool)
 		for name := range srvs {
 			if name == "consul" {
 				continue
@@ -195,7 +198,17 @@ func (d *discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
 				break
 			}
 			tgs = append(tgs, tg)
+			newSourceList[tg.Source] = true
 		}
+		// when targetGroup disappear, send an update with empty targetList
+		for key := range d.oldSourceList {
+			if !newSourceList[key] {
+				tgs = append(tgs, &targetgroup.Group{
+					Source: key,
+				})
+			}
+		}
+		d.oldSourceList = newSourceList
 		if err == nil {
 			// We're returning all Consul services as a single targetgroup.
 			ch <- tgs
@@ -216,6 +229,7 @@ func newDiscovery(conf sdConfig) (*discovery, error) {
 		refreshInterval: conf.RefreshInterval,
 		tagSeparator:    conf.TagSeparator,
 		logger:          logger,
+		oldSourceList:   make(map[string]bool),
 	}
 	return cd, nil
 }
--- a/documentation/examples/custom-sd/adapter/adapter.go
+++ b/documentation/examples/custom-sd/adapter/adapter.go
@@ -18,14 +18,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/go-kit/kit/log"
-	"github.com/go-kit/kit/log/level"
-	"github.com/prometheus/prometheus/discovery"
-	"github.com/prometheus/prometheus/discovery/targetgroup"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"reflect"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/prometheus/discovery"
+	"github.com/prometheus/prometheus/discovery/targetgroup"
 )
 
 type customSD struct {
--- a/documentation/examples/custom-sd/adapter/adapter.go
+++ b/documentation/examples/custom-sd/adapter/adapter.go
@@ -60,6 +60,12 @@ func (a *Adapter) generateTargetGroups(allTargetGroups map[string][]*targetgroup
 	tempGroups := make(map[string]*customSD)
 	for k, sdTargetGroups := range allTargetGroups {
 		for i, group := range sdTargetGroups {
+
+			// There is no target, so no need to keep it
+			if len(group.Targets) <= 0 {
+				continue
+			}
+
 			newTargets := make([]string, 0)
 			newLabels := make(map[string]string)
 
--- a/documentation/examples/custom-sd/adapter/adapter.go
+++ b/documentation/examples/custom-sd/adapter/adapter.go
@@ -18,15 +18,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
-	"os"
-	"path/filepath"
-	"reflect"
-
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/prometheus/discovery"
 	"github.com/prometheus/prometheus/discovery/targetgroup"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"reflect"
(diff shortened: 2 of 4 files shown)


simonpasquier:
The scrape manager still needs to receive the information that the targets are gone. See comment below on L251-252:

https://github.com/prometheus/prometheus/blob/457e4bb58e0fd3c10e16ecaabfeb04fb2c41bac5/discovery/manager.go#L239-L251

cstyan:
@Nexucis As Simon mentioned, the Manager needs to to know that the targets are gone. I think there are two issues here. 

1) The way that the channels are set up, generateAllTarget groups seems to get all target groups, even if that target group has disappeared. For example, I tweaked `main.go` in `adapter-usage` to not track consul itself. If the consul's config contains prometheus, the adapter picks it up and writes it, but then if we remove the consul config file and tell consul to reload it's config the next update in `generateTargetGroups` still thinks prometheus exists as a target group.

2) Within a target group itself we'd need to keep track of individual targets disappearing.

Nexucis:
Hello,

@cstyan since what @simonpasquier mention in this PL, I search how to properly remove the target without impact the ScrapManager.

I still think that it's not really proper to do it at the adapter level, but seems that it is the only way to handle this minor issue.
So I'll update this PR with the appropriate changes soon.

I'm sorry to not replying immediately after what @simonpasquier said. 
Don't worry I didn't give up this PR :)

cstyan:
@Nexucis  No problem, I'll review again when you push new changes :)

The reason the changes should be in the adapter are because this is an adapter specific issue. The existing SD mechanisms track their targets and send updates with empty data for targets that disappear, as Simon pointed out. Manager properly handles these updates.

Since the adapter is essentially an out of process wrapper around another instance of manager, it's the adapter code that isn't handling targets that go away properly.

documentation/examples/custom-sd/adapter/adapter_test.go
```diff
@@ -0,0 +1,151 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package adapter
+
+import (
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/discovery/targetgroup"
```
simonpasquier: need to separate the imports

```go
import (
        "reflect"
        "testing"

	"github.com/prometheus/prometheus/discovery/targetgroup"
)
```

documentation/examples/custom-sd/adapter/adapter_test.go
```diff
@@ -0,0 +1,151 @@
...
+				},
+			},
+		},
+	}
+
+	for _, testCase := range testCases {
+		result := generateTargetGroups(testCase.targetGroup)
+
+		if !reflect.DeepEqual(result, testCase.expectedCustomSD) {
+			t.Errorf("%v :\nresult produced %v\nmismatch the expected customSD: %v",
```
simonpasquier: (nit) the wording could be simplified IMO

```go
t.Errorf("%q failed\ngot: %#v\nexpected: %v",
...
```

documentation/examples/custom-sd/adapter/adapter_test.go
cstyan: can be 
```
import (
        "reflect"
        "testing"

	"github.com/prometheus/prometheus/discovery/targetgroup"
)
```
imo

Nexucis:
Thanks @cstyan for the explanation. I'll try to find an appropriate solution next week :)

cstyan:
:+1: sounds good

Nexucis:
@cstyan I'm just trying to implement the last use case that you found. 
I was wondering if applying the "crappy workaround" talked in the issue on adapter-usage (custom_consul_sd) in addition of the current diff would fix this use case ?

I updated the PR, you can see the diff on the adapter usage

documentation/examples/custom-sd/adapter-usage/main.go
```diff
@@ -195,7 +198,17 @@ func (d *discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
 				break
 			}
 			tgs = append(tgs, tg)
+			newSourceList[tg.Source] = true
+		}
+		// when targetGroup disappear, send an update with empty targetList
```
simonpasquier: (nit) comments start with uppercase and end with a dot.

documentation/examples/custom-sd/adapter/adapter.go
```diff
@@ -54,12 +54,16 @@ func mapToArray(m map[string]*customSD) []customSD {
...
-// Parses incoming target groups updates. If the update contains changes to the target groups
-// Adapter already knows about, or new target groups, we Marshal to JSON and write to file.
-func (a *Adapter) generateTargetGroups(allTargetGroups map[string][]*targetgroup.Group) {
-	tempGroups := make(map[string]*customSD)
+func generateTargetGroups(allTargetGroups map[string][]*targetgroup.Group) map[string]*customSD {
+	groups := make(map[string]*customSD)
 	for k, sdTargetGroups := range allTargetGroups {
 		for i, group := range sdTargetGroups {
+
+			// There is no target, so no need to keep it
```
simonpasquier: missing dot

documentation/examples/custom-sd/adapter/adapter_test.go
```diff
@@ -0,0 +1,152 @@
...
+import (
+	"reflect"
+	"testing"
+
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/discovery/targetgroup"
+)
+
+// TestGenerateTargetGroups checks that the target is correctly generated.
+// It covers the case when the target is empty
```
simonpasquier: missing dot

simonpasquier:
Thanks @Nexucis for the PR and @cstyan for the review!

PR #4579, opened 2018-09-05
proto: Allow reproducible proto generation.

bwplotka:
Current way of setting prometheus project to generate code from proto files is tedious.
If you just go install packages described in README.md, it ends up with this (non buildable code)
```
../../../prompb/remote.pb.go:61:34: undefined: "github.com/prometheus/prometheus/vendor/github.com/gogo/protobuf/proto".InternalMessageInfo
../../../prompb/remote.pb.go:109:33: undefined: "github.com/prometheus/prometheus/vendor/github.com/gogo/protobuf/proto".InternalMessageInfo
../../../prompb/remote.pb.go:164:34: undefined: "github.com/prometheus/prometheus/vendor/github.com/gogo/protobuf/proto".InternalMessageInfo
../../../prompb/remote.pb.go:214:27: undefined: "github.com/prometheus/prometheus/vendor/github.com/gogo/protobuf/proto".InternalMessageInfo
../../../prompb/remote.pb.go:284:33: undefined: "github.com/prometheus/prometheus/vendor/github.com/gogo/protobuf/proto".InternalMessageInfo
```
The issue is that the guide presented in proto folder is asking to install some GO packages. It does not mention from what version and there is strong mismatch between those used to generate proto and those that are vendored and used to run the code later...
The particular outdated packages in vendor are:
`github.com/gogo/protobuf` and `google.golang.org/grpc`.

Just updating vendor would be not enough, as it would only tmp fix.
I propose adding script that will parse version from vendor.json and install correct version.

Changes:
- Added script that uses packages in *exactly* the same version the Go code is build against later.
- Regenerated proto with those pinned deps.
- Updated REAMDE.md
- Added `make proto`

I have mixed feeling about install script.
* It's nice that it requires manual step (we could attach this script to genproto). It checkouts existing packages in GOPATH as well, which could be counted as .. unsafe?
* I am strongly suggesting leaving it somewhere... Otherwise it's just extremely hard to have reproducible proto generation.

Any thoughts?

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -123,6 +123,11 @@ common-docker-tag-latest:
 promu:
 	GOOS= GOARCH= $(GO) get -u github.com/prometheus/promu
 
+.PHONY: proto
+proto:
+	@echo ">> generating code from proto files"
+	@./scripts/genproto.sh
+
 .PHONY: $(STATICCHECK)
 $(STATICCHECK):
 	GOOS= GOARCH= $(GO) get -u honnef.co/go/tools/cmd/staticcheck
--- a/prompb/README.md
+++ b/prompb/README.md
@@ -2,13 +2,8 @@ The compiled protobufs are version controlled and you won't normally need to
 re-compile them when building Prometheus.
 
 If however you have modified the defs and do need to re-compile, run
-`./scripts/genproto.sh` from the parent dir.
+`make proto` from the parent dir.
 
-In order for the script to run, you'll need `protoc` (version 3.5) in your
-PATH, and the following Go packages installed:
+In order for the script to run, you'll need `protoc` (version 3.5.1) in your
+PATH.
 
-- github.com/gogo/protobuf
-- github.com/gogo/protobuf/protoc-gen-gogofast
-- github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway/
-- github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger
-- golang.org/x/tools/cmd/goimports
--- a/prompb/remote.pb.go
+++ b/prompb/remote.pb.go
@@ -330,24 +330,6 @@ func (m *QueryResult) MarshalTo(dAtA []byte) (int, error) {
 	return i, nil
 }
 
-func encodeFixed64Remote(dAtA []byte, offset int, v uint64) int {
-	dAtA[offset] = uint8(v)
-	dAtA[offset+1] = uint8(v >> 8)
-	dAtA[offset+2] = uint8(v >> 16)
-	dAtA[offset+3] = uint8(v >> 24)
-	dAtA[offset+4] = uint8(v >> 32)
-	dAtA[offset+5] = uint8(v >> 40)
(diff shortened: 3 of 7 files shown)


Bplotka:
Some work to be done on this.. not everything builds still

simonpasquier:
Another alternative would be to have all the correct versions required by `genproto.sh` checked in `vendor/`.

Bplotka:
PTAL @brancz @brian-brazil @tomwilkie  - this helped me to find correct versions of deps to properly generate our proto, worth to allow other as well to not waste time on this.

Bplotka:
@simonpasquier Agree, but we use `govendor` (and maybe someday something else) that cares `only` about generating deps from code, don't care things required to build plugins for proto build.

Also we cannot vendor binaries (as we would end up to vendor installation for all platforms etc)

simonpasquier:
Ah right, sorry for the noise.

scripts/install_genproto_deps.sh
```diff
@@ -0,0 +1,50 @@
...
+do
+    # This regex helps to find "root" for Golang package which is always in form <url>/<author/org>/<package>
+    rootPackageRegex="([a-z.-]*/[a-z.-]*/[a-z.-]*)/.*"
+    if [[ "${dep}" =~ $rootPackageRegex ]]; then
+        rootPackage="${BASH_REMATCH[1]}"
+    else
+        rootPackage="${dep}"
+    fi
+
+    version=$(jq '[.package[] | select(.path|test("'"${rootPackage}"'."))][1].revision' ./vendor/vendor.json | tr -d '"')
```
brancz: Do we require `jq` anywhere else in our builds today? If not it's technically a new dependency. I don't care myself, just noting it. If no one cares, I'm happy with merging with it.

brian-brazil: We don't, and it's not a standard tool.

simonpasquier: From my POV, this way of retrieving the package versions is brittle anyhow. FWIW etcd is pinning the exact dependencies in their `genproto.sh` script + get them in a scratch GOPATH directory. Although it is duplication from `vendor/vendor.json`, it is also more explicit and we don't update the protobuf dependencies that often.

https://github.com/etcd-io/etcd/blob/a18b46711125d08de37d0b71e9c0d42db363ed9e/scripts/genproto.sh#L21-L24

Bplotka: I am fine to move to more explicit version, without auto deduction of deps version. Maybe instead we could add circle CI step to generate proto and compare with what is commited?

cstyan:
:+1: I lost a few hours to this issue today :disappointed:

simonpasquier:
Definitely +1 on following the etcd example.

brancz:
The script lgtm, and the (re-)generated code as well as CI passed. :+1:

bwplotka:
so.. are we ready to merge this? (:

brancz:
I'm happy with merging this, but would like to wait for the final 2.5 release to land. (my prediction is that that will happen on monday unless there is a major bug report)

bwplotka:
Thx

PR #4581, opened 2018-09-06
Add updaterTicker that can be overwritten in tests.

### LINKED ISSUE #4565: Make the Discovery throtling configurable to allow easier testing.
https://github.com/prometheus/prometheus/blob/75bd348135019c6bf5bee4350a339964e62f2d8a/discovery/manager.go#L134

Sending the discovery updates are hard coded to throttle at 5 seconds. 
This makes creaating test for these parts of the code needing to wait 5 seconds so not good for unit testing. We need to make this somehow configurable so that we can adjust when creating unit tests.

With this change to the same PR  can also add a test for:

 - blocked or no updates when the receiver's channel is full - https://github.com/prometheus/prometheus/pull/4523#issuecomment-415967435
 - send updates only when there was an actual change in the targets - https://github.com/prometheus/prometheus/pull/4545

janickic:
Fixes #4565 
I wasn't sure if this should be configurable via command line option or in the config file, but if it's only for testing within the discovery package, I didn't want to expose modification of the value outside the package.
@krasi-georgiev

### DIFF
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -93,6 +93,7 @@ func NewManager(ctx context.Context, logger log.Logger) *Manager {
 		targets:        make(map[poolKey]map[string]*targetgroup.Group),
 		discoverCancel: []context.CancelFunc{},
 		ctx:            ctx,
+		updaterInterval: 5*time.Second,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


simonpasquier:
Thanks for working on this! I'm working on a PR (should be sent soon) that will fix this as well as adding more unit tests for the discovery manager.

simonpasquier:
This is #4582. I've added you as a co-author as I "stole" from your PR for the ticker field.

janickic:
Thanks for reviewing @simonpasquier! Should I close this PR?

simonpasquier:
Yes please.

simonpasquier:
And again thanks for your work!

PR #4582, opened 2018-09-06
discovery: add more tests

### LINKED ISSUE #4565: Make the Discovery throtling configurable to allow easier testing.

simonpasquier:
Closes #4565

When I run this PR without #4556, it fails as expected.


discovery/manager_test.go
```diff
@@ -868,40 +885,289 @@ scrape_configs:
...
+								Source:  "tg2",
+								Targets: []model.LabelSet{{"__instance__": "2"}},
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			title: "The receiver gets an update when a target group is gone",
```
krasi-georgiev: doesn't this and all the tests below already duplicate with the tests in `TestTargetUpdatesOrder` ?

simonpasquier:
Thanks for the review @krasi-georgiev 

I've added an explicit test (`TestCoordinationWithEmptyProvider`) when the provider closes the update channel although it was implicitly tested by `TestTargetSetRecreatesTargetGroupsEveryRun`. I've also modified the tests in `TestCoordinationWithReceiver` to fail if they didn't receive anything after 5 seconds while something is still expected.

> doesn't this and all the tests below already duplicate with the tests in TestTargetUpdatesOrder ?

Somehow yes but this test is more end-to-end since it reads from the SD manager update channel while `TestTargetUpdatesOrder` bypasses `updater()`. I should probably rewrite `TestTargetUpdatesOrder` to check the output of the update channel. WDYT?

Finally regarding
> https://github.com/prometheus/prometheus/blob/068eaa5dbfce6c08f3d05d3d3c0bfd96267cfed2/discovery/manager.go#L204-L207

This is covered by the `"Receiver should get updates even when the channel is blocked"` test. See https://github.com/prometheus/prometheus/pull/4582/files?utf8=%E2%9C%93&diff=unified#diff-f7fd37c9f5b407b3f4f1bd3ef08c624dR902

krasi-georgiev:
> I've added an explicit test (TestCoordinationWithEmptyProvider) when the provider closes the update channel although it was implicitly tested by TestTargetSetRecreatesTargetGroupsEveryRun. I've also modified the tests in TestCoordinationWithReceiver to fail if they didn't receive anything after 5 seconds while something is still expected.

I had something else in mind.
Start a provider ->  send some updates -> close the channel before the discovery manager's `updatert` 
this ensures that any updates before `updatert` are sent even when the provider closes its channel

krasi-georgiev:
>>  doesn't this and all the tests below already duplicate with the tests in TestTargetUpdatesOrder ?

> Somehow yes but this test is more end-to-end since it reads from the SD manager update channel while TestTargetUpdatesOrder bypasses updater(). I should probably rewrite TestTargetUpdatesOrder to check the output of the update channel. WDYT?

yeah I would say that these test fit better in the `TestTargetUpdatesOrder` so lets refactor.

discovery/manager_test.go
```diff
@@ -868,40 +885,354 @@ scrape_configs:
...
+		tgs   map[string][]*targetgroup.Group
+	}
+
+	testCases := []struct {
+		title     string
+		providers map[string]Discoverer
+		expected  []expect
+	}{
+		{
+			title: "Receiver should get an empty map even when the provider sends nothing and closes the channel",
```
krasi-georgiev: why do we need to test for this? Was there a bug for this?

discovery/manager_test.go
```diff
@@ -868,40 +885,354 @@ scrape_configs:
...
+								Source:  "tg2",
+								Targets: []model.LabelSet{{"__instance__": "2"}},
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			title: "Receiver should get updates even when the channel is blocked",
```
krasi-georgiev: What about:
"Receiver should get all updates even when the channel was blocked"

discovery/manager_test.go
```diff
@@ -868,40 +885,354 @@ scrape_configs:
...
+				"empty": &onceProvider{},
+			},
+			expected: []expect{
+				{
+					tgs: map[string][]*targetgroup.Group{},
+				},
+			},
+		},
+		{
+			title: "Receiver should get updates even when one provider closes its channel",
```
krasi-georgiev: What about: "Receiver should get all updates even when one provider closes its channel"

discovery/manager_test.go
simonpasquier: Not really. This is the result of some exploratory testing but agreed that it isn't necessary.

simonpasquier:
BTW I've found something that could be optimized in the manager: there are as many `updater()` goroutines as discoverers meaning that the updates can be sent to the channel more often than every 5 seconds when you have many discoverers.

It is trivial to fix it but I wonder if I miss something here? I've got a local branch that moves the sending of the updates to the channel from `updater()` to a single goroutine and it passes the new tests.

krasi-georgiev:
ooh yeah I think we discussed this somewhere.
Lets merge this one to keep it small and can open a PR for that optimisation as well.

simonpasquier:
> Lets merge this one to keep it small

Sure even though this PR isn't really small already :smirk:

simonpasquier:
I should have addressed all your comments. I'll refactor `TestTargetUpdatesOrder` with/after the optimization for dispatching the updates only once otherwise the tests aren't be reliable with multiple providers.

krasi-georgiev:
Yep that is even better.

LGTM, Thanks!

simonpasquier:
@gouthamve any opinion on this PR?

PR #4586, opened 2018-09-07
Stop removing the final dot from rooted DNS names

### LINKED ISSUE #3196: DNS discovery should not trim final dot

bboreham:
Removing a final dot changes the meaning of the name and can cause extra DNS lookups as the resolver traverses its search path.

Fixes #3196

### DIFF
--- a/discovery/dns/dns.go
+++ b/discovery/dns/dns.go
@@ -181,9 +181,6 @@ func (d *Discovery) refresh(ctx context.Context, name string, ch chan<- []*targe
 		target := model.LabelValue("")
 		switch addr := record.(type) {
 		case *dns.SRV:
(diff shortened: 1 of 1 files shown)


simonpasquier:
:+1:

@gouthamve does it qualify for merging now knowing that 2.4.0-rc.0 is out?

gouthamve:
No, master is frozen until a stable release is made. Only critical bug-fixes for `2.4.0-rc.x` will be accepted until a stable is out.

gouthamve:
👍

grobie:
This is a breaking change and will change/invalidate relabel rules based on `__address__` labels afaics. While I see the motivation for this change, it has serious side effects for users of the DNS discovery.

PR #4588, opened 2018-09-07
Use WAL to gather samples for remote write

cstyan:
Example of WAL tailing. Still need to do something with what we read out of the WAL.

cc @csmarchbanks @tomwilkie

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -307,7 +307,7 @@ func main() {
 
 	var (
 		localStorage  = &tsdb.ReadyStorage{}
-		remoteStorage = remote.NewStorage(log.With(logger, "component", "remote"), localStorage.StartTime, time.Duration(cfg.RemoteFlushDeadline))
+		remoteStorage = remote.NewStorage(log.With(logger, "component", "remote"), prometheus.DefaultRegisterer, localStorage.StartTime, cfg.localStoragePath, time.Duration(cfg.RemoteFlushDeadline))
 		fanoutStorage = storage.NewFanout(logger, localStorage, remoteStorage)
 	)
 
--- a/storage/remote/client_test.go
+++ b/storage/remote/client_test.go
@@ -26,7 +26,6 @@ import (
 
 	config_util "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
-	"github.com/prometheus/prometheus/prompb"
 )
 
 var longErrMessage = strings.Repeat("error message", maxErrMsgLen)
@@ -74,7 +73,7 @@ func TestStoreHTTPErrorHandling(t *testing.T) {
 			t.Fatal(err)
 		}
 
-		err = c.Store(context.Background(), &prompb.WriteRequest{})
+		err = c.Store(context.Background(), []byte{})
 		if !reflect.DeepEqual(err, test.err) {
 			t.Errorf("%d. Unexpected error; want %v, got %v", i, test.err, err)
 		}
--- a/storage/remote/codec.go
+++ b/storage/remote/codec.go
@@ -80,28 +80,6 @@ func EncodeReadResponse(resp *prompb.ReadResponse, w http.ResponseWriter) error 
 	return err
 }
 
-// ToWriteRequest converts an array of samples into a WriteRequest proto.
-func ToWriteRequest(samples []*model.Sample) *prompb.WriteRequest {
-	req := &prompb.WriteRequest{
-		Timeseries: make([]prompb.TimeSeries, 0, len(samples)),
-	}
-
-	for _, s := range samples {
-		ts := prompb.TimeSeries{
-			Labels: MetricToLabelProtos(s.Metric),
-			Samples: []prompb.Sample{
-				{
-					Value:     float64(s.Value),
-					Timestamp: int64(s.Timestamp),
-				},
-			},
-		}
-		req.Timeseries = append(req.Timeseries, ts)
-	}
-
-	return req
-}
-
 // ToQuery builds a Query proto.
 func ToQuery(from, to int64, matchers []*labels.Matcher, p *storage.SelectParams) (*prompb.Query, error) {
 	ms, err := toLabelMatchers(matchers)
@@ -364,21 +342,6 @@ func fromLabelMatchers(matchers []*prompb.LabelMatcher) ([]*labels.Matcher, erro
 	return result, nil
 }
 
-// MetricToLabelProtos builds a []*prompb.Label from a model.Metric
-func MetricToLabelProtos(metric model.Metric) []prompb.Label {
-	labels := make([]prompb.Label, 0, len(metric))
-	for k, v := range metric {
-		labels = append(labels, prompb.Label{
-			Name:  string(k),
-			Value: string(v),
-		})
-	}
-	sort.Slice(labels, func(i int, j int) bool {
-		return labels[i].Name < labels[j].Name
-	})
-	return labels
-}
-
 // LabelProtosToMetric unpack a []*prompb.Label to a model.Metric
 func LabelProtosToMetric(labelPairs []*prompb.Label) model.Metric {
 	metric := make(model.Metric, len(labelPairs))
@@ -400,6 +363,26 @@ func labelProtosToLabels(labelPairs []prompb.Label) labels.Labels {
 	return result
 }
 
+func labelsetToLabelsProto(ls model.LabelSet) []prompb.Label {
+	result := make([]prompb.Label, 0, len(ls))
+	keys := make([]string, 0, len(ls))
+
+	for k := range ls {
+		keys = append(keys, string(k))
+	}
+	sort.Strings(keys)
+
+	for _, k := range keys {
+		ln := model.LabelName(k)
+		result = append(result, prompb.Label{
+			Name:  k,
+			Value: string(ls[ln]),
+		})
+	}
+
+	return result
+}
+
 func labelsToLabelsProto(labels labels.Labels) []prompb.Label {
 	result := make([]prompb.Label, 0, len(labels))
 	for _, l := range labels {
@@ -409,12 +392,4 @@ func labelsToLabelsProto(labels labels.Labels) []prompb.Label {
 		})
 	}
 	return result
-}
-
-func labelsToMetric(ls labels.Labels) model.Metric {
-	metric := make(model.Metric, len(ls))
-	for _, l := range ls {
-		metric[model.LabelName(l.Name)] = model.LabelValue(l.Value)
-	}
-	return metric
 }
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -16,6 +16,7 @@ package remote
 import (
 	"context"
 	"math"
+	"strconv"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -24,12 +25,16 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+	"github.com/gogo/protobuf/proto"
+	"github.com/golang/snappy"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/config"
 	pkgrelabel "github.com/prometheus/prometheus/pkg/relabel"
 	"github.com/prometheus/prometheus/prompb"
 	"github.com/prometheus/prometheus/relabel"
+	"github.com/prometheus/tsdb"
 )
 
 // String constants for instrumentation.
@@ -66,7 +71,16 @@ var (
 			Namespace: namespace,
 			Subsystem: subsystem,
 			Name:      "failed_samples_total",
-			Help:      "Total number of samples which failed on send to remote storage.",
+			Help:      "Total number of samples which failed on send to remote storage, non-recoverable errors.",
+		},
+		[]string{queue},
+	)
+	retriedSamplesTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "retried_samples_total",
+			Help:      "Total number of samples which failed on send to remote storage but were retried because the send error was recoverable.",
 		},
 		[]string{queue},
 	)
@@ -75,7 +89,16 @@ var (
 			Namespace: namespace,
 			Subsystem: subsystem,
 			Name:      "dropped_samples_total",
-			Help:      "Total number of samples which were dropped due to the queue being full.",
+			Help:      "Total number of samples which were dropped after being read from the WAL before being sent via remote write.",
+		},
+		[]string{queue},
+	)
+	enqueueRetriesTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "enqueue_retries_total",
+			Help:      "Total number of times enqueue has failed because a shards queue was full.",
 		},
 		[]string{queue},
 	)
@@ -89,12 +112,30 @@ var (
 		},
 		[]string{queue},
 	)
-	queueLength = prometheus.NewGaugeVec(
+	queueLastSendTimestamp = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Namespace: namespace,
 			Subsystem: subsystem,
-			Name:      "queue_length",
-			Help:      "The number of processed samples queued to be sent to the remote storage.",
+			Name:      "queue_last_send_timestamp",
+			Help:      "Timestamp of the last successful send by this queue.",
+		},
+		[]string{queue},
+	)
+	queueHighestSentTimestamp = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "queue_highest_sent_timestamp",
+			Help:      "Timestamp from a WAL sample, the highest timestamp successfully sent by this queue.",
+		},
+		[]string{queue},
+	)
+	queuePendingSamples = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "pending_samples",
+			Help:      "The number of samples pending in the queues shards to be sent to the remote storage.",
 		},
 		[]string{queue},
 	)
@@ -121,9 +162,13 @@ var (
 func init() {
 	prometheus.MustRegister(succeededSamplesTotal)
 	prometheus.MustRegister(failedSamplesTotal)
+	prometheus.MustRegister(retriedSamplesTotal)
 	prometheus.MustRegister(droppedSamplesTotal)
+	prometheus.MustRegister(enqueueRetriesTotal)
 	prometheus.MustRegister(sentBatchDuration)
-	prometheus.MustRegister(queueLength)
+	prometheus.MustRegister(queueLastSendTimestamp)
+	prometheus.MustRegister(queueHighestSentTimestamp)
+	prometheus.MustRegister(queuePendingSamples)
 	prometheus.MustRegister(shardCapacity)
 	prometheus.MustRegister(numShards)
 }
@@ -132,37 +177,54 @@ func init() {
 // external timeseries database.
 type StorageClient interface {
 	// Store stores the given samples in the remote storage.
-	Store(context.Context, *prompb.WriteRequest) error
+	Store(context.Context, []byte) error
 	// Name identifies the remote storage implementation.
 	Name() string
 }
 
 // QueueManager manages a queue of samples to be sent to the Storage
-// indicated by the provided StorageClient.
+// indicated by the provided StorageClient. Implements writeTo interface
+// used by WAL Watcher.
 type QueueManager struct {
 	logger log.Logger
 
-	flushDeadline  time.Duration
-	cfg            config.QueueConfig
-	externalLabels model.LabelSet
-	relabelConfigs []*pkgrelabel.Config
-	client         StorageClient
-	queueName      string
-	logLimiter     *rate.Limiter
+	flushDeadline              time.Duration
+	cfg                        config.QueueConfig
+	externalLabels             model.LabelSet
+	relabelConfigs             []*pkgrelabel.Config
+	client                     StorageClient
+	queueName                  string
+	logLimiter                 *rate.Limiter
+	watcher                    *WALWatcher
+	lastSendTimestampMetric    prometheus.Gauge
+	highestSentTimestampMetric prometheus.Gauge
+	pendingSamplesMetric       prometheus.Gauge
+	enqueueRetriesMetric       prometheus.Counter
 
-	shardsMtx   sync.RWMutex
+	lastSendTimestamp    int64
+	highestSentTimestamp int64
+	timestampLock        sync.Mutex
+
+	highestTimestampIn *int64 // highest timestamp of any sample ingested by remote storage via scrape (Appender)
+
+	seriesMtx            sync.Mutex
+	seriesLabels         map[uint64][]prompb.Label
+	seriesSegmentIndexes map[uint64]int
+	droppedSeries        map[uint64]struct{}
+
 	shards      *shards
 	numShards   int
 	reshardChan chan int
-	quit        chan struct{}
-	wg          sync.WaitGroup
+
+	quit chan struct{}
+	wg   sync.WaitGroup
 
 	samplesIn, samplesOut, samplesOutDuration *ewmaRate
 	integralAccumulator                       float64
 }
 
 // NewQueueManager builds a new QueueManager.
-func NewQueueManager(logger log.Logger, cfg config.QueueConfig, externalLabels model.LabelSet, relabelConfigs []*pkgrelabel.Config, client StorageClient, flushDeadline time.Duration) *QueueManager {
+func NewQueueManager(logger log.Logger, walDir string, samplesIn *ewmaRate, highestTimestampIn *int64, cfg config.QueueConfig, externalLabels model.LabelSet, relabelConfigs []*pkgrelabel.Config, client StorageClient, flushDeadline time.Duration, startTime int64) *QueueManager {
 	if logger == nil {
 		logger = log.NewNopLogger()
 	} else {
@@ -177,16 +239,29 @@ func NewQueueManager(logger log.Logger, cfg config.QueueConfig, externalLabels m
 		client:         client,
 		queueName:      client.Name(),
 
+		highestTimestampIn: highestTimestampIn,
+
+		seriesLabels:         make(map[uint64][]prompb.Label),
+		seriesSegmentIndexes: make(map[uint64]int),
+		droppedSeries:        make(map[uint64]struct{}),
+
 		logLimiter:  rate.NewLimiter(logRateLimit, logBurst),
 		numShards:   cfg.MinShards,
 		reshardChan: make(chan int),
 		quit:        make(chan struct{}),
 
-		samplesIn:          newEWMARate(ewmaWeight, shardUpdateDuration),
+		samplesIn:          samplesIn,
 		samplesOut:         newEWMARate(ewmaWeight, shardUpdateDuration),
 		samplesOutDuration: newEWMARate(ewmaWeight, shardUpdateDuration),
 	}
-	t.shards = t.newShards(t.numShards)
+
+	t.lastSendTimestampMetric = queueLastSendTimestamp.WithLabelValues(t.queueName)
+	t.highestSentTimestampMetric = queueHighestSentTimestamp.WithLabelValues(t.queueName)
+	t.pendingSamplesMetric = queuePendingSamples.WithLabelValues(t.queueName)
+	t.enqueueRetriesMetric = enqueueRetriesTotal.WithLabelValues(t.queueName)
+	t.watcher = NewWALWatcher(logger, t, walDir, startTime)
+	t.shards = t.newShards()
+
 	numShards.WithLabelValues(t.queueName).Set(float64(t.numShards))
 	shardCapacity.WithLabelValues(t.queueName).Set(float64(t.cfg.Capacity))
 
@@ -194,77 +269,158 @@ func NewQueueManager(logger log.Logger, cfg config.QueueConfig, externalLabels m
 	sentBatchDuration.WithLabelValues(t.queueName)
 	succeededSamplesTotal.WithLabelValues(t.queueName)
 	failedSamplesTotal.WithLabelValues(t.queueName)
-	droppedSamplesTotal.WithLabelValues(t.queueName)
+	retriedSamplesTotal.WithLabelValues(t.queueName)
+	// Reset pending samples metric to 0.
+	t.pendingSamplesMetric.Set(0)
 
 	return t
 }
 
-// Append queues a sample to be sent to the remote storage. It drops the
-// sample on the floor if the queue is full.
-// Always returns nil.
-func (t *QueueManager) Append(s *model.Sample) error {
-	snew := *s
-	snew.Metric = s.Metric.Clone()
+// Append queues a sample to be sent to the remote storage. Blocks until all samples are
+// enqueued on their shards or a shutdown signal is received.
+func (t *QueueManager) Append(s []tsdb.RefSample) bool {
+	type enqueuable struct {
+		ts  prompb.TimeSeries
+		ref uint64
+	}
 
-	for ln, lv := range t.externalLabels {
-		if _, ok := s.Metric[ln]; !ok {
-			snew.Metric[ln] = lv
+	tempSamples := make([]enqueuable, 0, len(s))
+	t.seriesMtx.Lock()
+	for _, sample := range s {
+		// If we have no labels for the series, due to relabelling or otherwise, don't send the sample.
+		if _, ok := t.seriesLabels[sample.Ref]; !ok {
+			droppedSamplesTotal.WithLabelValues(t.queueName).Inc()
+			if _, ok := t.droppedSeries[sample.Ref]; !ok && t.logLimiter.Allow() {
+				level.Info(t.logger).Log("msg", "dropped sample for series that was not explicitly dropped via relabelling", "ref", sample.Ref)
+			}
+			continue
+		}
+		tempSamples = append(tempSamples, enqueuable{
+			ts: prompb.TimeSeries{
+				Labels: t.seriesLabels[sample.Ref],
+				Samples: []prompb.Sample{
+					prompb.Sample{
+						Value:     float64(sample.V),
+						Timestamp: sample.T,
+					},
+				},
+			},
+			ref: sample.Ref,
+		})
+	}
+	t.seriesMtx.Unlock()
+
+	backoff := t.cfg.MinBackoff
+outer:
+	for _, sample := range tempSamples {
+		// This will result in spin/busy waiting if the queues are being resharded
+		// or shutting down.  TODO backoff.
+		for {
+			select {
+			case <-t.quit:
+				return false
+			default:
+			}
+
+			if t.shards.enqueue(sample.ref, sample.ts) {
+				continue outer
+			}
+			t.enqueueRetriesMetric.Inc()
+			time.Sleep(time.Duration(backoff))
+			backoff = backoff * 2
+			if backoff > t.cfg.MaxBackoff {
+				backoff = t.cfg.MaxBackoff
+			}
 		}
 	}
-
-	snew.Metric = model.Metric(
-		relabel.Process(model.LabelSet(snew.Metric), t.relabelConfigs...))
-
-	if snew.Metric == nil {
-		return nil
-	}
-
-	t.shardsMtx.RLock()
-	enqueued := t.shards.enqueue(&snew)
-	t.shardsMtx.RUnlock()
-
-	if enqueued {
-		queueLength.WithLabelValues(t.queueName).Inc()
-	} else {
-		droppedSamplesTotal.WithLabelValues(t.queueName).Inc()
-		if t.logLimiter.Allow() {
-			level.Warn(t.logger).Log("msg", "Remote storage queue full, discarding sample. Multiple subsequent messages of this kind may be suppressed.")
-		}
-	}
-	return nil
-}
-
-// NeedsThrottling implements storage.SampleAppender. It will always return
-// false as a remote storage drops samples on the floor if backlogging instead
-// of asking for throttling.
-func (*QueueManager) NeedsThrottling() bool {
-	return false
+	return true
 }
 
 // Start the queue manager sending samples to the remote storage.
 // Does not block.
 func (t *QueueManager) Start() {
+	t.shards.start(t.numShards)
+	t.watcher.Start()
+
 	t.wg.Add(2)
 	go t.updateShardsLoop()
 	go t.reshardLoop()
-
-	t.shardsMtx.Lock()
-	defer t.shardsMtx.Unlock()
-	t.shards.start()
 }
 
 // Stop stops sending samples to the remote storage and waits for pending
 // sends to complete.
 func (t *QueueManager) Stop() {
 	level.Info(t.logger).Log("msg", "Stopping remote storage...")
+	defer level.Info(t.logger).Log("msg", "Remote storage stopped.")
+
 	close(t.quit)
+	t.shards.stop()
+	t.watcher.Stop()
 	t.wg.Wait()
+}
 
-	t.shardsMtx.Lock()
-	defer t.shardsMtx.Unlock()
-	t.shards.stop(t.flushDeadline)
+func (t *QueueManager) Name() string {
+	return t.queueName
+}
 
-	level.Info(t.logger).Log("msg", "Remote storage stopped.")
+// Find out which series are dropped after relabelling and make sure we have a metric label for them.
+func (t *QueueManager) diffKeys(ref uint64, original, relabelled model.LabelSet) {
+	numDropped := len(original) - len(relabelled)
+	if numDropped == 0 {
+		return
+	}
+}
+
+// StoreSeries keeps track of which series we know about for lookups when sending samples to remote.
+func (t *QueueManager) StoreSeries(series []tsdb.RefSeries, index int) {
+	temp := make(map[uint64][]prompb.Label, len(series))
+	for _, s := range series {
+		ls := make(model.LabelSet, len(s.Labels))
+		for _, label := range s.Labels {
+			ls[model.LabelName(label.Name)] = model.LabelValue(label.Value)
+		}
+		t.processExternalLabels(ls)
+		rl := relabel.Process(ls, t.relabelConfigs...)
+
+		t.diffKeys(s.Ref, ls, rl)
+		if len(rl) == 0 {
+			t.droppedSeries[s.Ref] = struct{}{}
+			continue
+		}
+		temp[s.Ref] = labelsetToLabelsProto(rl)
+	}
+
+	t.seriesMtx.Lock()
+	defer t.seriesMtx.Unlock()
+	for ref, labels := range temp {
+		t.seriesLabels[ref] = labels
+		t.seriesSegmentIndexes[ref] = index
+	}
+}
+
+// SeriesReset is used when reading a checkpoint. WAL Watcher should have
+// stored series records with the checkpoints index number, so we can now
+// delete any ref ID's lower than that # from the two maps.
+func (t *QueueManager) SeriesReset(index int) {
+	t.seriesMtx.Lock()
+	defer t.seriesMtx.Unlock()
+
+	// Check for series that are in segments older than the checkpoint
+	// that were not also present in the checkpoint.
+	for k, v := range t.seriesSegmentIndexes {
+		if v < index {
+			delete(t.seriesLabels, k)
+			delete(t.seriesSegmentIndexes, k)
+		}
+	}
+}
+
+func (t *QueueManager) processExternalLabels(ls model.LabelSet) {
+	for ln, lv := range t.externalLabels {
+		if _, ok := ls[ln]; !ok {
+			ls[ln] = lv
+		}
+	}
 }
 
 func (t *QueueManager) updateShardsLoop() {
@@ -275,6 +431,12 @@ func (t *QueueManager) updateShardsLoop() {
 	for {
 		select {
 		case <-ticker.C:
+			now := time.Now().Unix()
+			threshold := int64(time.Duration(2 * t.cfg.BatchSendDeadline).Seconds())
+			if now-t.lastSendTimestamp > threshold {
+				level.Debug(t.logger).Log("msg", "Skipping resharding, last successful send was beyond threshold")
+				continue
+			}
 			t.calculateDesiredShards()
 		case <-t.quit:
 			return
@@ -351,107 +513,150 @@ func (t *QueueManager) reshardLoop() {
 	for {
 		select {
 		case numShards := <-t.reshardChan:
-			t.reshard(numShards)
+			// We start the newShards after we have stopped (the therefore completely
+			// flushed) the oldShards, to guarantee we only every deliver samples in
+			// order.
+			t.shards.stop()
+			t.shards.start(numShards)
 		case <-t.quit:
 			return
 		}
 	}
 }
 
-func (t *QueueManager) reshard(n int) {
-	numShards.WithLabelValues(t.queueName).Set(float64(n))
-
-	t.shardsMtx.Lock()
-	newShards := t.newShards(n)
-	oldShards := t.shards
-	t.shards = newShards
-	t.shardsMtx.Unlock()
-
-	oldShards.stop(t.flushDeadline)
-
-	// We start the newShards after we have stopped (the therefore completely
-	// flushed) the oldShards, to guarantee we only every deliver samples in
-	// order.
-	newShards.start()
-}
-
-type shards struct {
-	qm      *QueueManager
-	queues  []chan *model.Sample
-	done    chan struct{}
-	running int32
-	ctx     context.Context
-	cancel  context.CancelFunc
-}
-
-func (t *QueueManager) newShards(numShards int) *shards {
-	queues := make([]chan *model.Sample, numShards)
-	for i := 0; i < numShards; i++ {
-		queues[i] = make(chan *model.Sample, t.cfg.Capacity)
-	}
-	ctx, cancel := context.WithCancel(context.Background())
+func (t *QueueManager) newShards() *shards {
 	s := &shards{
-		qm:      t,
-		queues:  queues,
-		done:    make(chan struct{}),
-		running: int32(numShards),
-		ctx:     ctx,
-		cancel:  cancel,
+		qm:   t,
+		done: make(chan struct{}),
 	}
 	return s
 }
 
-func (s *shards) start() {
-	for i := 0; i < len(s.queues); i++ {
-		go s.runShard(i)
+// Check and set highestSentTimestamp
+func (t *QueueManager) setHighestSentTimestamp(highest int64) {
+	t.timestampLock.Lock()
+	defer t.timestampLock.Unlock()
+	if highest > t.highestSentTimestamp {
+		t.highestSentTimestamp = highest
+		t.highestSentTimestampMetric.Set(float64(t.highestSentTimestamp))
 	}
 }
 
-func (s *shards) stop(deadline time.Duration) {
-	// Attempt a clean shutdown.
-	for _, shard := range s.queues {
-		close(shard)
+func (t *QueueManager) setLastSendTimestamp(now time.Time) {
+	t.timestampLock.Lock()
+	defer t.timestampLock.Unlock()
+	t.lastSendTimestampMetric.Set(float64(now.UnixNano()) / 1e9)
+	t.lastSendTimestamp = now.Unix()
+}
+
+type shards struct {
+	mtx sync.RWMutex // With the WAL, this is never actually contended.
+
+	qm     *QueueManager
+	queues []chan prompb.TimeSeries
+
+	// Emulate a wait group with a channel and an atomic int, as you
+	// cannot select on a wait group.
+	done    chan struct{}
+	running int32
+
+	// Soft shutdown context will prevent new enqueues and deadlocks.
+	softShutdown chan struct{}
+
+	// Hard shutdown context is used to terminate outgoing HTTP connections
+	// after giving them a chance to terminate.
+	hardShutdown context.CancelFunc
+}
+
+// start the shards; must be called before any call to enqueue.
+func (s *shards) start(n int) {
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	newQueues := make([]chan prompb.TimeSeries, n)
+	for i := 0; i < n; i++ {
+		newQueues[i] = make(chan prompb.TimeSeries, s.qm.cfg.Capacity)
+	}
+
+	s.queues = newQueues
+
+	var hardShutdownCtx context.Context
+	hardShutdownCtx, s.hardShutdown = context.WithCancel(context.Background())
+	s.softShutdown = make(chan struct{})
+	s.running = int32(n)
+	s.done = make(chan struct{})
+	for i := 0; i < n; i++ {
+		go s.runShard(hardShutdownCtx, i, newQueues[i])
+	}
+	numShards.WithLabelValues(s.qm.queueName).Set(float64(n))
+}
+
+// stop the shards; subsequent call to enqueue will return false.
+func (s *shards) stop() {
+	// Attempt a clean shutdown, but only wait flushDeadline for all the shards
+	// to cleanly exit.  As we're doing RPCs, enqueue can block indefinately.
+	// We must be able so call stop concurrently, hence we can only take the
+	// RLock here.
+	s.mtx.RLock()
+	close(s.softShutdown)
+	s.mtx.RUnlock()
+
+	// Enqueue should now be unblocked, so we can take the write lock.  This
+	// also ensures we don't race with writes to the queues, and get a panic:
+	// send on closed channel.
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+	for _, queue := range s.queues {
+		close(queue)
 	}
 	select {
 	case <-s.done:
 		return
-	case <-time.After(deadline):
+	case <-time.After(s.qm.flushDeadline):
 		level.Error(s.qm.logger).Log("msg", "Failed to flush all samples on shutdown")
 	}
 
 	// Force an unclean shutdown.
-	s.cancel()
+	s.hardShutdown()
 	<-s.done
 }
 
-func (s *shards) enqueue(sample *model.Sample) bool {
-	s.qm.samplesIn.incr(1)
-
-	fp := sample.Metric.FastFingerprint()
-	shard := uint64(fp) % uint64(len(s.queues))
+// enqueue a sample.  If we are currently in the process of shutting down or resharding,
+// will return false; in this case, you should back off and retry.
+func (s *shards) enqueue(ref uint64, sample prompb.TimeSeries) bool {
+	s.mtx.RLock()
+	defer s.mtx.RUnlock()
 
 	select {
+	case <-s.softShutdown:
+		return false
+	default:
+	}
+
+	shard := uint64(ref) % uint64(len(s.queues))
+	select {
+	case <-s.softShutdown:
+		return false
 	case s.queues[shard] <- sample:
 		return true
-	default:
-		return false
 	}
 }
 
-func (s *shards) runShard(i int) {
+func (s *shards) runShard(ctx context.Context, i int, queue chan prompb.TimeSeries) {
 	defer func() {
 		if atomic.AddInt32(&s.running, -1) == 0 {
 			close(s.done)
 		}
 	}()
 
-	queue := s.queues[i]
+	shardNum := strconv.Itoa(i)
 
 	// Send batches of at most MaxSamplesPerSend samples to the remote storage.
 	// If we have fewer samples than that, flush them out after a deadline
 	// anyways.
-	pendingSamples := model.Samples{}
+	pendingSamples := []prompb.TimeSeries{}
 
+	max := s.qm.cfg.MaxSamplesPerSend
 	timer := time.NewTimer(time.Duration(s.qm.cfg.BatchSendDeadline))
 	stop := func() {
 		if !timer.Stop() {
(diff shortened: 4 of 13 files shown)


documentation/examples/tail-wal/main.go
```diff
@@ -0,0 +1,153 @@
...
+	var (
+		dec      tsdb.RecordDecoder
+		series   []tsdb.RefSeries
+		samples  []tsdb.RefSample
+		nSeries  = 0
+		nSamples = 0
+	)
+	for {
+		ok := r.Next()
+		rec := r.Record()
```
csmarchbanks: If you check ok before calling r.Record() I think all the defer/recover stuff can be avoided. Using a `for r.Next()` loop would also simplify everything, then just static returns (true if in the for, false after the loop)

documentation/examples/tail-wal/main.go
```diff
@@ -0,0 +1,153 @@
...
+
+	// Read the existing segments except the last one.
+	var segment *wal.Segment
+	for ; m < n; m++ {
+		segment, err = wal.OpenReadSegment(wal.SegmentName(*walDir, m))
+		if err != nil {
+			level.Error(log.With(logger)).Log("err", err)
+		}
+		defer segment.Close()
+		for {
```
csmarchbanks: This could just be `for decode(segment) {}`, though even better would be perhaps decode only returns when it is done reading the entire segment regardless of a decoding error? Or else errors need to be handled somewhere?

documentation/examples/tail-wal/main.go
cstyan: yep, for a while I wanted to change how `r.Next` worked to have it return an error but for now I don't think it's necessary

storage/remote/storage.go
```diff
@@ -120,6 +111,186 @@ func (s *Storage) ApplyConfig(conf *config.Config) error {
...
+					// level.Info(logger).Log("event", "write", "file", event.Name)
+					s.logger.Log("event", "write", "file", event.Name)
+					s.decodeSegment(segment)
+				}
+				if event.Op&fsnotify.Create == fsnotify.Create {
+					// TODO: callum, we should check if the file name is a new segment and not a checkpoint or something else
+					// level.Info(logger).Log("event", "create", "file", event.Name)
+					s.logger.Log("event", "create", "file", event.Name)
+
+					segment.Close()
```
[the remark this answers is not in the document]
cstyan: Yes, they are, I just hadn't gotten around to resetting the map yet :)

storage/remote/storage.go
```diff
@@ -120,6 +111,186 @@ func (s *Storage) ApplyConfig(conf *config.Config) error {
...
+func (s *Storage) storeSeries(series []tsdb.RefSeries) {
+	for i := 0; i < len(series); i++ {
+		_, ok := s.series[series[i].Ref]
+		if !ok {
+			s.series[series[i].Ref] = series[i].Labels
+		}
+	}
+}
+
+func (s *Storage) storeSamples(samples []tsdb.RefSample) {
```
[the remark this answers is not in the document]
cstyan: Probably, but I just wanted to get something working first and then see what kind of throughput we can get.

storage/remote/storage.go
```diff
@@ -120,6 +111,186 @@ func (s *Storage) ApplyConfig(conf *config.Config) error {
...
+		metric := make(model.Metric, len(s.series[sample.Ref]))
+		for _, v := range s.series[sample.Ref] {
+			metric[model.LabelName(v.Name)] = model.LabelValue(v.Value)
+		}
+		c := &model.Sample{
+			Metric:    metric,
+			Value:     model.SampleValue(sample.V),
+			Timestamp: model.Time(sample.T),
+		}
+		sa = append(sa, c)
```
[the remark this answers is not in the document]
cstyan: :+1: yep, we should reuse the code in queue_manager, client, write, etc. but I just wanted to be sure I could convert from the record/refSamples to the protobuf for now

storage/remote/storage.go
```diff
@@ -120,6 +111,186 @@ func (s *Storage) ApplyConfig(conf *config.Config) error {
...
+					s.decodeSegment(segment)
+				}
+				if event.Op&fsnotify.Create == fsnotify.Create {
+					// TODO: callum, we should check if the file name is a new segment and not a checkpoint or something else
+					// level.Info(logger).Log("event", "create", "file", event.Name)
+					s.logger.Log("event", "create", "file", event.Name)
+
+					segment.Close()
+					n++
+					segment, err = wal.OpenReadSegment(wal.SegmentName(s.walDir, n))
```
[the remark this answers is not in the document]
cstyan: Yep, we'll have to decide how we want to handle this. Whether it's just a pointer into the wal, or if the watcher simply forwards samples to another goroutine via a channel. Or something else.

cstyan:
@tomwilkie could use a sanity check review here. Also @csmarchbanks if you're interested :)

Some thoughts:
- The way I'm setting up and starting the go routine that's reading from the WAL feels wrong. One idea is that main could start remote storage and pass it a channel. Remote storage would start the routine immediately, and it would wait on this channel to recv a string containing the path to the WAL dir, then proceed as it does now.
- If we need to obey the Appender interface we could take in a second channel that sends data we read from the WAL to a go routine(s) that just does writes. Or potentially relabelling and writes. Similar to the way queues are setup now. Except rather than Appender `Add()` being called as a result of scraping, we would call it in `decodeSegment` or somewhere around there.
- I'm still not sure how to handle/if we want to read the WAL segments via some type of pointer into the WAL, or periodically polling the current segment, rather than the current setup of using fsnotify. I think we still want to use fsnotify to get file creation events.
- Relabeling seems expensive. Maybe the global config external labels bit is applied to samples in tsdb on ingestion and so we don't need to do it here? Still need to investigate here.

csmarchbanks:
No real opinions on the first two points yet.

> I'm still not sure how to handle/if we want to read the WAL segments via some type of pointer into the WAL, or periodically polling the current segment, rather than the current setup of using fsnotify. I think we still want to use fsnotify to get file creation events.

I think some type of pointer will be necessary for when a remote write endpoint goes down. We will need to not move forward in the WAL for that remote write until it comes back up. fsnotify could then be used to keep track of if there are new segments for the pointer to move to when it finishes reading the current segment. Also, if the current segment is deleted the pointer would have to be moved to the next segment.

> Relabeling seems expensive. Maybe the global config external labels bit is applied to samples in tsdb on ingestion and so we don't need to do it here? Still need to investigate here.

Isn't relabeling done on every sample now? So it shouldn't be any more expensive than the current implementation? Since you can specify relabel configs for each remote write I don't know how you can avoid it.

cstyan:
> I think some type of pointer will be necessary for when a remote write endpoint goes down. We will need to not move forward in the WAL for that remote write until it comes back up. fsnotify could then be used to keep track of if there are new segments for the pointer to move to when it finishes reading the current segment. Also, if the current segment is deleted the pointer would have to be moved to the next segment

Right, I keep forgetting that there can be multiple remote write URL's.

> Isn't relabeling done on every sample now? So it shouldn't be any more expensive than the current implementation? Since you can specify relabel configs for each remote write I don't know how you can avoid it.

I just want to confirm whether or not the external labels are applied on sample ingestion into TSDB. If they are, we don't need to attempt to apply them again when reading out of the tsdb. See: https://github.com/prometheus/prometheus/pull/4588/files#diff-4a762c727e88ccd4b3454b9312afcccdR46

brian-brazil:
> I just want to confirm whether or not the external labels are applied on sample ingestion into TSDB.

They aren't. They're added as the samples are sent out by remote_write, before write_relabel_configs.

csmarchbanks:
> The way I'm setting up and starting the go routine that's reading from the WAL feels wrong.

I don't think main should start the wal watcher. That will need to be inside the remote write code, and done for each remote write config (like how we currently have a QueueManager for each).

What would be the reason for delaying starting the wal watcher using a channel? Seems like it would actually be good to have the wal reader read to the end of the wal (without sending any of the samples) before scraping starts to avoid sending duplicate samples?

cstyan:
> I don't think main should start the wal watcher. That will need to be inside the remote write code, and done for each remote write config (like how we currently have a QueueManager for each).

Agreed.

> What would be the reason for delaying starting the wal watcher using a channel? Seems like it would actually be good to have the wal reader read to the end of the wal (without sending any of the samples) before scraping starts to avoid sending duplicate samples?

We need to know the location of the tsdb/wal, which isn't necessarily set at the time that remote storage is created.

cstyan:
Some updates @csmarchbanks @tomwilkie we now do relabeling once per series per segment during storeSeries, and we store the series labels as `prompb.Label` rather than tsdb or pkg `Label` so that we don't need to recreate the labels as those types on every call to `storeSamples`.

cstyan:
Latest commit made some changes to the proto definitions, made it so that certain fields can't be null which results in `[]type` rather than `[]*type` and therefore less object allocations.

thanks to @bwplotka for this PR: https://github.com/prometheus/prometheus/pull/4579

cstyan:
pretty significant refactor in the latest commit

- there is now a single WAL watcher rather than one per remote write config
- make use of Tom's original work (queue manager) to do parallelization, move things like the cache of series labels from WAL watcher to the queue managers
- the WAL reading code was failing regularly under heavy (for my laptop) load if we got anything other than a tsdb series or sample record, we now handle the rest and seem to decode invalid records quite often

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,247 @@
...
+
+			for _, q := range w.storage.queues {
+				for _, s := range samples {
+					q.Append(s)
+				}
+			}
+			nSamples += len(samples)
+		case tsdb.RecordTombstones:
+			w.logger.Log("msg", "got tsdb tombstones")
+		case tsdb.RecordInvalid:
```
cstyan: @gouthamve or @fabxc any pointers on what would be causing us to read back invalid records? Should I be worried that we're reading back invalid/corrupt or incomplete data?

storage/remote/storage.go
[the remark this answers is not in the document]
cstyan: @csmarchbanks do you remember where we found out about this?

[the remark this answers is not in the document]
csmarchbanks: Now that I look at it again I am not sure. Looks like series and samples are actually recorded in two different `log` calls, so could potentially be across two different segments? https://github.com/prometheus/tsdb/blob/043e3bb5f97b07681bcf9d1d2d8f295fa5718f1d/head.go#L724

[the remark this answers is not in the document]
cstyan: Yeah, I could have sword I read somewhere that the labels were repeated, but after a conversation with @gouthamve that doesn't make any sense and I can't find anything that confirms labels are repeated.

prompb/remote.proto
```diff
@@ -40,5 +41,5 @@ message Query {
 
 message QueryResult {
   // Samples within a time series must be ordered by time.
-  repeated prometheus.TimeSeries timeseries = 1;
+  repeated prometheus.TimeSeries timeseries = 1 [(gogoproto.nullable) = false];
```
tomwilkie: I'd avoid making unnecessary changes for now.

storage/remote/client_test.go
```diff
@@ -74,7 +74,7 @@ func TestStoreHTTPErrorHandling(t *testing.T) {
 			t.Fatal(err)
 		}
 
-		err = c.Store(context.Background(), &prompb.WriteRequest{})
+		err = c.Store(context.Background(), prompb.WriteRequest{})
```
tomwilkie: Lets keep the top-level proto messages as pointers-to-structs, as `proto.Marshall` expects that.

storage/remote/codec.go
```diff
@@ -80,10 +81,30 @@ func EncodeReadResponse(resp *prompb.ReadResponse, w http.ResponseWriter) error
...
+		req.Timeseries = append(req.Timeseries, ts)
+	}
+	return req
+}
+
 // ToWriteRequest converts an array of samples into a WriteRequest proto.
-func ToWriteRequest(samples []*model.Sample) *prompb.WriteRequest {
-	req := &prompb.WriteRequest{
-		Timeseries: make([]*prompb.TimeSeries, 0, len(samples)),
+func ToWriteRequest(samples []*model.Sample) prompb.WriteRequest {
```
tomwilkie: Is this still used, or can it be removed?  

If its still used, make it `[]model.Sample`.

tomwilkie: Yeah looks like its unused.

storage/remote/codec.go
```diff
@@ -365,10 +386,10 @@ func fromLabelMatchers(matchers []*prompb.LabelMatcher) ([]*labels.Matcher, erro
 }
 
 // MetricToLabelProtos builds a []*prompb.Label from a model.Metric
-func MetricToLabelProtos(metric model.Metric) []*prompb.Label {
-	labels := make([]*prompb.Label, 0, len(metric))
+func MetricToLabelProtos(metric model.Metric) []prompb.Label {
```
tomwilkie: I think this can be removed too (or moved to the tests).

storage/remote/queue_manager.go
```diff
@@ -266,6 +272,48 @@ func (t *QueueManager) Stop() {
...
 
+// Keep track of which series we know about for lookups when sending samples to remote.
+func (t *QueueManager) storeSeries(series []tsdb.RefSeries) {
+	for i := 0; i < len(series); i++ {
+		var ls = make(model.LabelSet)
+		for _, label := range series[i].Labels {
+			ls[model.LabelName(label.Name)] = model.LabelValue(label.Value)
+		}
+		_, ok := t.series[series[i].Ref]
+		if !ok {
```
tomwilkie: Under what circumstances will we find a label set already in the cache?

And can you move this above the building of the `model.LabelSet` please?

Finally, generally in go we aim for the least nesting possible; so we favour:

```golang
if _, ok := t.series[series[i].Ref]; ok {
  continue
}
```

storage/remote/queue_manager.go
```diff
@@ -266,6 +272,48 @@ func (t *QueueManager) Stop() {
...
+			t.processExternalLabels(ls, series[i].Labels)
+			relabel.Process(ls, t.relabelConfigs...)
+			t.series[series[i].Ref] = labelsToLabelsProto(labelsetToLabels(ls))
```
tomwilkie: Consider combining `labelsToLabelsProto` and `labelsetToLabels` into a single function.

Also, labels are sorted, which is lost when you convert to model.LabelSet.  Need to sort again.

storage/remote/queue_manager.go
```diff
@@ -428,11 +476,11 @@ func (s *shards) stop(deadline time.Duration) {
...
 }
 
-func (s *shards) enqueue(sample *model.Sample) bool {
+func (s *shards) enqueue(sample tsdb.RefSample) bool {
 	s.qm.samplesIn.incr(1)
 
-	fp := sample.Metric.FastFingerprint()
-	shard := uint64(fp) % uint64(len(s.queues))
+	// fp := sample.Metric.FastFingerprint()
+	shard := uint64(sample.Ref) % uint64(len(s.queues))
```
tomwilkie: How is the ref generated?  Is in only ever incrementing?  I just wonder if this will balance across queues correctly.

storage/remote/queue_manager.go
```diff
@@ -428,11 +476,11 @@ func (s *shards) stop(deadline time.Duration) {
...
 
 	select {
 	case s.queues[shard] <- sample:
```
tomwilkie: Need to remove the `select` and `default` - this should block if the queue is full.

storage/remote/queue_manager.go
```diff
@@ -514,9 +562,9 @@ func (s *shards) sendSamples(samples model.Samples) {
 }
 
 // sendSamples to the remote storage with backoff for recoverable errors.
-func (s *shards) sendSamplesWithBackoff(samples model.Samples) {
+func (s *shards) sendSamplesWithBackoff(samples []tsdb.RefSample) {
 	backoff := s.qm.cfg.MinBackoff
-	req := ToWriteRequest(samples)
+	req := toWriteRequest(samples, s.qm.series)
```
tomwilkie: Should access to s.qm.series be locked?  And won't this introduce contention?  Perhaps each shard should have it own series map?

storage/remote/queue_manager.go
```diff
@@ -514,9 +562,9 @@ func (s *shards) sendSamples(samples model.Samples) {
...
 
 	for retries := s.qm.cfg.MaxRetries; retries > 0; retries-- {
```
tomwilkie: We shouldn't drop samples any more - MaxRetires should be infinite.

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,253 @@
...
+)
+
+// WALWatcher watches the TSDB WAL for a given StorageClient (remote write config).
+type WALWatcher struct {
+	storage *Storage
+	logger  log.Logger
+
+	currentSegment *wal.Segment
+	walDir         string
+	series         map[uint64][]prompb.Label
```
tomwilkie: Not used anymore.

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,253 @@
...
+	return
+}
+
+func labelsetToLabels(ls model.LabelSet) labels.Labels {
+	var l labels.Labels
+	for k, v := range ls {
+		l = append(l, labels.Label{Name: string(k), Value: string(v)})
+	}
+	return l
+}
```
tomwilkie: If you keep this, move it to codec.go please.

storage/remote/write.go
```diff
@@ -28,15 +27,15 @@ func (s *Storage) Appender() (storage.Appender, error) {
...
+	// 	if err := q.Append(&model.Sample{
+	// 		Metric:    labelsToMetric(l),
+	// 		Timestamp: model.Time(t),
+	// 		Value:     model.SampleValue(v),
+	// 	}); err != nil {
+	// 		panic(err) // QueueManager.Append() should always return nil as per doc string.
+	// 	}
+	// }
 	return 0, nil
 }
```
tomwilkie: Drop all this code now.

tomwilkie:
Need to make the queue lengths much shorter now - maybe only 10 samples.  That should bound memory usage when the remote endpoint goes down.

Also, not sure the dynamic sharding will work anymore - it needs to measure the number of samples being scraped, and size the number of queues based on that.  It used to do this by using the calls to append.  Now that this code path is called by the WALWatcher, which has back pressure if the queue is full, we won't measure the true rate.  Some need something before the writes to the WAL - perhaps we still need to implement Appended, but ignore the values and just count them.  WDYT?

storage/remote/storage.go
```diff
@@ -28,14 +29,22 @@ import (
...
 // Storage represents all the remote read and write endpoints.  It implements
 // storage.Storage.
 type Storage struct {
 	logger log.Logger
 	mtx    sync.RWMutex
 
 	// For writes
-	queues []*QueueManager
+	watcher *WALWatcher
+	walDir  string
```
tomwilkie: I don't see why this needs to be remembered in this struct.

storage/remote/storage.go
```diff
@@ -28,14 +29,22 @@ import (
...
+	dirChan chan string
```
tomwilkie: Whats this used for?

tomwilkie: OIC

tomwilkie:
Finally, whilst we're here - can we have the code convert, encoded and compress each request once, and not once per retry?

storage/remote/queue_manager.go
cstyan: the series reference id's are atomically incremented: https://github.com/prometheus/tsdb/blob/master/head.go#L1109-L1110

prompb/remote.proto
cstyan: this is unnecessary because it just saves memory allocations? not required for changing how remote write receives samples I guess :+1:

storage/remote/queue_manager.go
cstyan: Previously I thought series records were repeated per segment file, but I was wrong. I can remove the check for now, I don't think there's a case where we would receive another series record for a reference ID that's already in the cache.

storage/remote/queue_manager.go
cstyan: Do you mean they have to be sorted in order to be treated as the same series?

storage/remote/write.go
cstyan: `Add` and `AddFast` could just be no-ops now.

cmd/prometheus/main.go
```diff
@@ -321,8 +321,9 @@ func main() {
 	}
 
 	// Depends on cfg.web.ScrapeManager so needs to be after cfg.web.ScrapeManager = scrapeManager
+	fmt.Println("starting web handler")
 	webHandler := web.New(log.With(logger, "component", "web"), &cfg.web)
-
+	fmt.Println("started web handler")
```
tomwilkie: Remove pls.

cmd/prometheus/main.go
```diff
@@ -566,13 +567,20 @@ func main() {
 					prometheus.DefaultRegisterer,
 					&cfg.tsdb,
 				)
+
+				pwd, _ := os.Getwd()
+				walDir := fmt.Sprintf("%s/%swal/", pwd, db.Dir())
+				fmt.Println("setting wal dir")
+				remoteStorage.SetWALDir(walDir)
+				fmt.Println("set wal dir")
```
tomwilkie: Clean up.

cmd/prometheus/main.go
```diff
@@ -566,13 +567,20 @@ func main() {
 					prometheus.DefaultRegisterer,
 					&cfg.tsdb,
 				)
+
+				pwd, _ := os.Getwd()
+				walDir := fmt.Sprintf("%s/%swal/", pwd, db.Dir())
```
tomwilkie: path.join?

tomwilkie: do you need to deal with absolute paths for db.Dir?

storage/remote/queue_manager.go
```diff
@@ -161,21 +179,26 @@ type QueueManager struct {
 }
 
 // NewQueueManager builds a new QueueManager.
-func NewQueueManager(logger log.Logger, cfg config.QueueConfig, externalLabels model.LabelSet, relabelConfigs []*config.RelabelConfig, client StorageClient, flushDeadline time.Duration) *QueueManager {
+func NewQueueManager(logger log.Logger, s *Storage, cfg config.QueueConfig, externalLabels model.LabelSet, relabelConfigs []*config.RelabelConfig, client StorageClient, flushDeadline time.Duration) *QueueManager {
```
tomwilkie: Please don't pass storage, you only need the EWMA - so wrap it in a simple little struct and use that?

storage/remote/queue_manager.go
```diff
@@ -282,7 +331,10 @@ func (t *QueueManager) updateShardsLoop() {
 }
 
 func (t *QueueManager) calculateDesiredShards() {
-	t.samplesIn.tick()
+	if t.storage == nil {
+		level.Warn(t.logger).Log("msg", "queue manager does not have storage pointer, can't calculate shards", "queue", t.queueName)
+	}
```
tomwilkie: Redundant, never nil.

storage/remote/queue_manager.go
```diff
@@ -376,17 +428,17 @@ func (t *QueueManager) reshard(n int) {
...
 	ctx     context.Context
 	cancel  context.CancelFunc
 }
 
 func (t *QueueManager) newShards(numShards int) *shards {
-	queues := make([]chan *model.Sample, numShards)
+	queues := make([]chan tsdb.RefSample, numShards)
 	for i := 0; i < numShards; i++ {
-		queues[i] = make(chan *model.Sample, t.cfg.Capacity)
+		queues[i] = make(chan tsdb.RefSample, 10)
```
tomwilkie: Lets either make this a const at the top of file, of put it back to t.cfg.Capacity and set that to 10 by default.

storage/remote/queue_manager.go
```diff
@@ -424,22 +476,30 @@ func (s *shards) stop(deadline time.Duration) {
...
 	return
 }
 
-func (s *shards) enqueue(sample *model.Sample) bool {
+func (s *shards) enqueue(sample tsdb.RefSample) bool {
 	s.qm.samplesIn.incr(1)
 
-	fp := sample.Metric.FastFingerprint()
-	shard := uint64(fp) % uint64(len(s.queues))
+	// fp := sample.Metric.FastFingerprint()
```
tomwilkie: Remove pls.

storage/remote/queue_manager.go
```diff
@@ -424,22 +476,30 @@ func (s *shards) stop(deadline time.Duration) {
...
+	shard := uint64(sample.Ref) % uint64(len(s.queues))
 
-	select {
-	case s.queues[shard] <- sample:
-		return true
-	default:
+	if int(shard) >= len(s.queues) {
+		level.Error(s.qm.logger).Log("err", "we would be chosing a non-existant shard", "shard", shard, "numQueues", len(s.queues))
+	}
```
tomwilkie: Erm...?

storage/remote/queue_manager.go
```diff
@@ -424,22 +476,30 @@ func (s *shards) stop(deadline time.Duration) {
...
+
+	if len(s.queues[shard]) == cap(s.queues[shard]) {
+		if s.qm.logLimiter.Allow() {
+			// This still seems to log frequently.
+			level.Info(s.qm.logger).Log("msg", "Shard queue is full. Multiple subsequent messages of this kind may be suppressed.", "shard", shard)
+		}
 		return false
 	}
```
tomwilkie: No logging here pls: far to noisy.  Shards being full is common case and is out back pressure mechanism.

storage/remote/queue_manager.go
```diff
@@ -424,22 +476,30 @@ func (s *shards) stop(deadline time.Duration) {
...
+	s.queues[shard] <- sample
```
tomwilkie: Is this your shutdown problem? `select` with `ctx.Done`?

storage/remote/queue_manager.go
```diff
@@ -454,7 +514,7 @@ func (s *shards) runShard(i int) {
 	// Send batches of at most MaxSamplesPerSend samples to the remote storage.
 	// If we have fewer samples than that, flush them out after a deadline
 	// anyways.
-	pendingSamples := model.Samples{}
+	var pendingSamples []tsdb.RefSample
```
tomwilkie: We prefer `:=`

storage/remote/queue_manager.go
```diff
@@ -485,14 +544,13 @@ func (s *shards) runShard(i int) {
 			queueLength.WithLabelValues(s.qm.queueName).Dec()
 			pendingSamples = append(pendingSamples, sample)
 
-			if len(pendingSamples) >= s.qm.cfg.MaxSamplesPerSend {
-				s.sendSamples(pendingSamples[:s.qm.cfg.MaxSamplesPerSend])
-				pendingSamples = pendingSamples[s.qm.cfg.MaxSamplesPerSend:]
+			if len(pendingSamples) >= 100 {
```
tomwilkie: Don't hard code this, use cfg.

storage/remote/queue_manager.go
```diff
@@ -503,41 +561,55 @@ func (s *shards) runShard(i int) {
...
+		return
+	}
 
 	// These counters are used to calculate the dynamic sharding, and as such
 	// should be maintained irrespective of success or failure.
 	s.qm.samplesOut.incr(int64(len(samples)))
 	s.qm.samplesOutDuration.incr(int64(time.Since(begin)))
+	if s.qm.storage != nil {
+		level.Debug(s.qm.logger).Log("pending", s.qm.storage.samplesIn.newEvents-s.qm.samplesOut.newEvents, "out", s.qm.samplesOut.newEvents, "in", s.qm.storage.samplesIn.newEvents)
+	}
```
tomwilkie: Remove this pls.

storage/remote/queue_manager.go
```diff
@@ -503,41 +561,55 @@ func (s *shards) runShard(i int) {
...
-			succeededSamplesTotal.WithLabelValues(s.qm.queueName).Add(float64(len(samples)))
-			return
-		}
+	s.qm.seriesMtx.Lock()
+	req := toWriteRequest(samples, s.qm.series)
+	s.qm.seriesMtx.Unlock()
+	for {
+		select {
+		case <-s.ctx.Done():
+			return nil
```
tomwilkie: `return ctx.Err()`

storage/remote/queue_manager.go
```diff
@@ -503,41 +561,55 @@ func (s *shards) runShard(i int) {
...
+		default:
+			begin := time.Now()
```
tomwilkie: Can we unnest this and just have an empty default?

storage/remote/queue_manager.go
```diff
@@ -503,41 +561,55 @@ func (s *shards) runShard(i int) {
...
+			err := s.qm.client.Store(s.ctx, req)
```
tomwilkie: Can we move the proto.Marshal and snappy.Compress into `buildWriteRequest` pls?

storage/remote/storage.go
```diff
@@ -120,6 +157,28 @@ func (s *Storage) ApplyConfig(conf *config.Config) error {
...
 }
 
+// TODO: callum, should this always be called via a go routine or...?
+func (s *Storage) SetWALDir(dir string) {
+	s.walDir = dir
+	s.dirChan <- s.walDir
+	// Wait for the WAL watcher to read to the end of the current WAL segment when
+	// we start Prometheus from an existing WAL.
+	<-s.dirChan
+}
```
tomwilkie: Can we do this statically at config time?

storage/remote/storage.go
```diff
@@ -120,6 +157,28 @@ func (s *Storage) ApplyConfig(conf *config.Config) error {
...
+
+func (s *Storage) SetLabels() {
+	// Head exists after TSDB.Open().
+	labels := s.db.GetAllLabels()
+	for q := range s.queues {
+		q.db = s.db
+		q.StoreSeriesLabels(labels)
+	}
+}
```
tomwilkie: This is going to go, right?

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,278 @@
...
+	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/tsdb"
+	"github.com/prometheus/tsdb/wal"
+	fsnotify "gopkg.in/fsnotify/fsnotify.v1"
+)
+
+// WALWatcher watches the TSDB WAL for a given StorageClient (remote write config).
+type WALWatcher struct {
+	storage *Storage
```
tomwilkie: We're going to go with each queue tailing the WAL because:
- when we have 0 remote writes, we don't want to tail
- when we have a single remote endpoint down, we don't to to block all

tomwilkie: Also, lets consider decoupling the WALWatcher from the Queue using an interface, as I suspect other people will use this.

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,278 @@
...
+	go w.watch()
+}
+
+func (w *WALWatcher) Stop() {
+	// It seems that sometimes the quit channel is already closed.
+	defer func() {
+		if r := recover(); r != nil {
+			level.Error(log.With(w.logger)).Log("err", r)
+		}
+	}()
```
tomwilkie: This can't be committed.

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,278 @@
...
+
+	currentSegment  *wal.Segment
+	walDir          string
+	lastForwardedTs prometheus.Gauge
+
+	dirChan        chan string
+	ctx            context.Context
+	cancel         context.CancelFunc
+	quit           chan struct{}
+	exitForwarding chan struct{}
```
tomwilkie: Do you need this?  Just use `quit` everywhere?

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,278 @@
...
+	return
+}
+
+// When starting the WAL watcher, there is potentially an existing WAL. In that case, we
+// should read to the end of the latest segment before listening for file events.
+func (w *WALWatcher) readToEnd(s *wal.Segment) {
+	r := wal.NewReader(s)
+	for r.Next() {
+	}
+}
```
tomwilkie: return `r.Err()` here?

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,278 @@
...
+
+func (w *WALWatcher) watch() {
+	// The WAL dir won't be set when Prometheus first starts up, wait to recv.
+	w.walDir = <-w.dirChan
+
+	var err error
+	nw, err := wal.New(nil, nil, w.walDir)
+	if err != nil {
+		w.logger.Log("err", err)
+	}
```
tomwilkie: Is it safe to continue?

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,278 @@
...
+	_, n, err := nw.Segments()
+
+	if err != nil {
```
tomwilkie: Remove newline above, but one before `_, n, err := nw.Segments()`

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,278 @@
...
+		w.logger.Log("err", err)
+	}
+	if n == -1 {
+		w.logger.Log("err", "no segments found")
+		return
+	}
+	watcher, err := fsnotify.NewWatcher()
```
tomwilkie: New line above please.

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,278 @@
...
+	var err error
+	nw, err := wal.New(nil, nil, w.walDir)
+	if err != nil {
+		w.logger.Log("err", err)
+	}
+	_, n, err := nw.Segments()
+
+	if err != nil {
+		w.logger.Log("err", err)
+	}
```
tomwilkie: Safe to continue on error here?

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,278 @@
...
+	if err != nil {
+		w.logger.Log("err", err)
+		return
+	}
+	w.readToEnd(segment)
+	// Send a message back to signal that we've read to the end of the current WAL segment and are starting to watch
+	// writes and new WAL segment files.
+	w.dirChan <- ""
+	var currentSegment = fmt.Sprintf("%08d", n)
+	defer segment.Close()
```
tomwilkie: This `defer` will run when you exit, and captures the current value of `segment` I think.

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,278 @@
...
+	segmentRegex := regexp.MustCompile(`\d{8}`)
+	checkpointRegex := regexp.MustCompile(`checkpoint.\d{6}`)
```
tomwilkie: Global pls.

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,278 @@
...
+
+			}
+		}
+	}
+	return nil
+}
+
+func (w *WALWatcher) decodeSegment(segment *wal.Segment) {
+	// todo: callum, is there an easy way to detect if r.Next() has returned the last possible record in a segment.
+	// For now just recover from the panic that is thrown if we call r.Next() when we shouldn't have
```
tomwilkie: re-reading something that EOF'd should just EOF; never panic in normal operations.

storage/remote/wal_watcher.go
tomwilkie: Put segment handling in a separate function to make closing semantics clear; then call that function in a loop.

storage/remote/queue_manager.go
cstyan: exporting `Capacity` on the `QueueConfig` struct makes it configurable by the user correct? we don't want it to be configurable anymore

storage/remote/queue_manager.go
cstyan: same question, should this still be configurable? can we just have it as a constant value when creating a new shard?

storage/remote/queue_manager.go
```diff
@@ -149,6 +161,12 @@ type QueueManager struct {
 	queueName      string
 	logLimiter     *rate.Limiter
 
+	series         map[uint64][]*prompb.Label
```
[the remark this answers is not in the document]
cstyan: What was the reasoning here? We're worried about lock contention? I can't remember what we were discussing.

storage/remote/queue_manager.go
```diff
@@ -266,6 +265,56 @@ func (t *QueueManager) Stop() {
...
+	t.seriesMtx.Lock()
+	defer t.seriesMtx.Unlock()
+
+	refs := t.db.AllSeriesIDs()
+	newSeries := make(map[uint64][]*prompb.Label)
+	for _, k := range refs[:] {
+		newSeries[k] = t.series[k]
+	}
+	t.series = newSeries
+}
```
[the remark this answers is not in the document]
cstyan: @tomwilkie @gouthamve do you remember what the plan was here? Having the map values be a struct with a field to tell us the segment they were in? When we get a file created event and it's a checkpoint, ideally we don't want to read all of the checkpoint and every segment after it up to where we currently are in the latest segment just to get all the series records. I'll double check what the checkpoint format is as well. Maybe it just contains series records for all active series at the time of the checkpoint?

storage/remote/queue_manager.go
cstyan: I seem to have broken the shutdown again

storage/remote/queue_manager.go
cstyan: Just select on the ctx so it's non-blocking? Is there any benefit to not using the default case, or downside to using it?

storage/remote/wal_watcher.go
cstyan: this will require a PR for tsdb, I'll work on that

storage/remote/wal_watcher.go
cstyan: tsdb pr required

cstyan:
Now that we retry everywhere, I'm having a hard time getting the shutdown logic to work properly. Or rather, keep it working properly. I had it working Friday after Tom pointed out a missing select, and now I've broken it with some changes to the Append function.

- for shutdowns, should we now be attempting to flush in all the functions that retry, not just in `sendSamples`/`sendSamplesWithBackoff`?
- it seems to me like we don't need the failed samples metric anymore? we either drop samples because we had no labels for them (due to relabeling, I need to update the help status there) but we retry forever, so we should never fail to send any.
- do we need to worry about persisting the segment struct/pointer across restarts right now? for example, if someone changes their config and causes ApplyConfig to be called, we might lose our spot? not sure if data can be written to the WAL during that time
- not really related to my changes, but it seems like there's lots of functions within remote storage that are exported outside the package that seem like they don't need to be, do we do that by default for using packages in other prometheus projects?

storage/remote/queue_manager.go
```diff
@@ -97,6 +98,15 @@ var (
...
 	)
+	queueSentTs = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "queue_last_sent_timestamp",
+			Help:      "Timestamp of the last successful send by this queue.",
+		},
+		[]string{queue},
+	)
```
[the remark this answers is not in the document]
cstyan: do you mean `Add` from `storage/remote/write.go`?

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,328 @@
...
+}
+
+// WALWatcher watches the TSDB WAL for a given StorageClient (remote write config).
+type WALWatcher struct {
+	writer  WriteTo
+	logger  log.Logger
+	walDir  string
+	watcher *fsnotify.Watcher
+
+	segment         *wal.Segment
```
tomwilkie: Not keen on the quasi-global state of the `segment` field in the WALWatcher; feels like this should be passed between functions as needed?  Its mainly just used in `watch()`.  That should make the lifecycle of the object clearer.

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,328 @@
...
+}
+
+// Blocks  until the sample is sent to all remote write endpoints
+func (w *WALWatcher) forwardSamples(s []tsdb.RefSample) error {
+	for {
+		select {
+		case <-w.quit:
+			return fmt.Errorf("exit when attempting to forward samples")
+		default:
+			ok := w.writer.Append(s)
```
tomwilkie: Now that Append retries forever, does it make sense to return an error?  If not, this whole loop can go away.

tomwilkie:
> for shutdowns, should we now be attempting to flush in all the functions that retry, not just in sendSamples/sendSamplesWithBackoff?

You should try and make sure we're only retrying in one place; and that should probably be sendSamples(withBackoff).  When you have a golang program thats not exiting cleanly, send it SIGQUIT (I think) and it will dump its goroutine stacks.  That should tell you which goroutine is blocked.

> it seems to me like we don't need the failed samples metric anymore? we either drop samples because we had no labels for them (due to relabeling, I need to update the help status there) but we retry forever, so we should never fail to send any.

I'd keep it (to see how many are failing and being retried).  I'd have a separate metric for dropped due to relabelling.

> do we need to worry about persisting the segment struct/pointer across restarts right now? for example, if someone changes their config and causes ApplyConfig to be called, we might lose our spot? not sure if data can be written to the WAL during that time

Not yet; thats the next PR.

> not really related to my changes, but it seems like there's lots of functions within remote storage that are exported outside the package that seem like they don't need to be, do we do that by default for using packages in other prometheus projects?

Not explicitly, some could very much be made private.  But don't make unnecessary changes for now.

cstyan:
> You should try and make sure we're only retrying in one place; and that should probably be sendSamples(withBackoff). When you have a golang program thats not exiting cleanly, send it SIGQUIT (I think) and it will dump its goroutine stacks. That should tell you which goroutine is blocked.

What about the potential samples backed up in WAL watcher because the queue was full? we drop those for now?

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,328 @@
...
+	err = w.watcher.Add(w.walDir)
+	if err != nil {
+		level.Error(w.logger).Log("err", err)
+		return
+	}
+
+	for {
+		select {
+		case <-w.quit:
+			return
```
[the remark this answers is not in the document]
cstyan: yeah this actually isn't necessary if `watch` returns a bool or error and continues to select on the quit channel

prompb/remote.proto
tomwilkie: Only reason to say this is because its read path not write path; was this necessary because of changes on the write path sharing helping functions? If not, try and avoid it.

storage/remote/queue_manager.go
tomwilkie: Yes, thats what I mean.

storage/remote/queue_manager.go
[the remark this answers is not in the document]
tomwilkie: Make `Appender()` create a new thing; every `Add()` should compare & update that; `Commit()` should then propagate that value to a metric.

tomwilkie:
> What about the potential samples backed up in WAL watcher because the queue was full? we drop those for now?

Yeah I think thats fine for now.

storage/remote/queue_manager.go
cstyan: :+1: makes sense. I am sorting the labels in order of `LabelName` in `labelsetToLabels`. I can look at combining `labelsToLabelsProto` and `labelsetToLabels`, but `labelsToLabelsProto` is also used by `ToQueryResult` and the change seems low priority for now.

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,362 @@
...
+	"github.com/prometheus/tsdb"
+	"github.com/prometheus/tsdb/wal"
+	fsnotify "gopkg.in/fsnotify/fsnotify.v1"
+)
+
+var (
+	segmentRegex              = regexp.MustCompile(`\d{8}`)
+	checkpointRegex           = regexp.MustCompile(`checkpoint.\d{6}`)
+	watcherSamplesRecordsRead = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
```
brian-brazil: This sounds like a counter

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,362 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/tsdb"
+	"github.com/prometheus/tsdb/wal"
+	fsnotify "gopkg.in/fsnotify/fsnotify.v1"
+)
+
+var (
+	segmentRegex              = regexp.MustCompile(`\d{8}`)
+	checkpointRegex           = regexp.MustCompile(`checkpoint.\d{6}`)
+	watcherSamplesRecordsRead = prometheus.NewGaugeVec(
```
brian-brazil: This sounds like a counter

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,362 @@
...
+	return w
+}
+
+func (w *WALWatcher) Start() {
+	go w.runWatcher()
+}
+
+func (w *WALWatcher) Stop() {
+	close(w.quit)
+	return
```
brian-brazil: This is redundant

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,362 @@
...
+				return nil
+			}
+		case <-segmentTicker.C:
+			w.readSegment(reader)
+		}
+	}
+}
+
+func (w *WALWatcher) runWatcher() {
+	// The WAL dir may not exist when Prometheus first starts up.
```
brian-brazil: You could do this with one for loop

storage/remote/wal_watcher.go
cstyan: :+1: yes it likely should be. When I was testing locally whether or not the read rate was correct I needed some of the metrics as gauges so that they explicitly started at 0 on each restart.

cstyan:
@brian-brazil thanks for the review!

> I only looked at the wal reading bits. This doesn't seem to be handling incomplete writes correctly yet, and I think you can avoid copying so much code into LiveReader.Next

Yep, it isn't handling them correctly. I see that in the # of active series/ingest rate on the remote end. For now, at least it's not failing when we do read an incomplete record, stuck in a state where all the records afterwards are read as garbage until we roll over to another segment file.

storage/remote/queue_manager.go
```diff
@@ -16,19 +16,22 @@ package remote
...
 	"sync/atomic"
 	"time"
 
-	"golang.org/x/time/rate"
-
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+	"github.com/gogo/protobuf/proto"
+	"github.com/golang/snappy"
 	"github.com/prometheus/client_golang/prometheus"
```
tomwilkie: Need to split out prometheus from non-prometheus imports pls.

storage/remote/queue_manager.go
```diff
@@ -266,6 +286,44 @@ func (t *QueueManager) Stop() {
...
+	}
+	t.seriesMtx.Lock()
+	defer t.seriesMtx.Unlock()
+	for ref, labels := range temp {
+		t.series[ref] = labels
+	}
+}
+
+// Clear all series stored in series labels cache.
+func (t *QueueManager) ClearSeries() {
```
tomwilkie: I think this could race with already enqueued samples; before you clear the series, you need to ensure everything has been flushed.

tomwilkie: Or you could look up the labels for a series in `Append`, which is called on the same goroutine as this so can't race.

storage/remote/queue_manager.go
```diff
@@ -482,30 +575,38 @@ func (s *shards) runShard(i int) {
...
 			pendingSamples = append(pendingSamples, sample)
+			queuePendingSamples.WithLabelValues(s.qm.queueName, shardNum).Set(float64(len(pendingSamples)))
 
-			if len(pendingSamples) >= s.qm.cfg.MaxSamplesPerSend {
-				s.sendSamples(pendingSamples[:s.qm.cfg.MaxSamplesPerSend])
-				pendingSamples = pendingSamples[s.qm.cfg.MaxSamplesPerSend:]
+			if len(pendingSamples) >= max {
+				s.sendSamples(pendingSamples[:max])
+				pendingSamples = pendingSamples[max:]
+				queuePendingSamples.WithLabelValues(s.qm.queueName, shardNum).Set(float64(len(pendingSamples)))
```
tomwilkie: `WithLabelValues` takes a lock, might want to cache that.

storage/remote/queue_manager.go
```diff
@@ -514,30 +615,48 @@ func (s *shards) sendSamples(samples model.Samples) {
...
+
+			sentBatchDuration.WithLabelValues(s.qm.queueName).Observe(time.Since(begin).Seconds())
+			if err != nil {
+				failedSamplesTotal.WithLabelValues(s.qm.queueName).Add(float64(len(samples)))
+			}
+
+			if err == nil {
+				succeededSamplesTotal.WithLabelValues(s.qm.queueName).Add(float64(len(samples)))
+				now := time.Now()
+				s.qm.sentTimestampMetric.Set(float64(now.UnixNano()) / 1e9)
```
tomwilkie: Shouldn't this be the time of the sample?  So we can compare the last scraped timestamp to the last sent timestamp, and alert if there is a big backlog,

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,359 @@
...
+func (w *WALWatcher) readSeriesRecords(r *wal.LiveReader) {
+	var (
+		dec    tsdb.RecordDecoder
+		series []tsdb.RefSeries
+	)
+
+	for {
+		err := r.Next()
+		if err != nil {
+			if err == io.EOF {
```
tomwilkie: Collapse this into one line `if err == io.EOF`.

You continue if err == that isn't EOF.

tomwilkie: Discussed: if this only returns EOF, probably want type to be bool?

storage/remote/write.go
```diff
@@ -26,17 +25,8 @@ func (s *Storage) Appender() (storage.Appender, error) {
...
-		if err := q.Append(&model.Sample{
-			Metric:    labelsToMetric(l),
-			Timestamp: model.Time(t),
-			Value:     model.SampleValue(v),
-		}); err != nil {
-			panic(err) // QueueManager.Append() should always return nil as per doc string.
-		}
-	}
+	s.samplesIn.incr(1)
+	s.samplesInMetric.Inc()
```
tomwilkie: Need to track the last scraped time stamp, for monitoring backlog.

vendor/github.com/prometheus/tsdb/wal/wal.go
```diff
@@ -861,6 +861,157 @@ func (r *Reader) Record() []byte {
...
+			r.err = errors.Wrap(err, "read first header byte")
+		}
+		read += n
+		if read == len(buf) {
+			break
+		} else if read != 0 {
+			fmt.Println("read partial record, the next read should finish it")
+			r.tempBuf = temp
+		}
+		time.Sleep(500 * time.Millisecond)
```
tomwilkie: ????

storage/remote/queue_manager.go
brian-brazil: That wouldn't work, as we can take samples back in time. If we want to track that the number of pending bytes would be better.

storage/remote/queue_manager.go
cstyan: samples are now enqueued with their labels

storage/remote/queue_manager.go
cstyan: this metric is specifically the time at which we last successfully sent samples, there's another metric for the highest timestamp from a sample that we have successfully sent

storage/remote/queue_manager.go
```diff
@@ -76,7 +91,7 @@ var (
 			Name:      "dropped_samples_total",
 			Help:      "Total number of samples which were dropped due to the queue being full.",
 		},
-		[]string{queue},
+		[]string{queue, "ref_id"},
```
tomwilkie: Whats `ref_id` and what is its cardinality?

storage/remote/queue_manager.go
```diff
@@ -115,39 +148,66 @@ var (
...
 		[]string{queue},
 	)
+	droppedSeries = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "dropped_series",
+			Help:      "Dropped series metric, the labels are what we care about here, compare to labels for dropped samples metric.",
+		},
+		[]string{queue, "ref_id"},
```
tomwilkie: Whats `ref_id` and whats its cardinality?

storage/remote/queue_manager.go
```diff
@@ -266,6 +360,70 @@ func (t *QueueManager) Stop() {
...
+		return
+	}
+	droppedSeries.WithLabelValues(t.queueName, strconv.FormatUint(ref, 10))
+}
+
+// Keep track of which series we know about for lookups when sending samples to remote.
+func (t *QueueManager) StoreSeries(series []tsdb.RefSeries, index int) {
+	temp := make(map[uint64][]prompb.Label)
+	for i := 0; i < len(series); i++ {
+		var ls = make(model.LabelSet)
```
tomwilkie: This could be initialised with the right size to save on allocs.

storage/remote/queue_manager.go
```diff
@@ -266,6 +360,70 @@ func (t *QueueManager) Stop() {
...
+	numDropped := len(original) - len(relabelled)
+	if numDropped == 0 {
+		return
+	}
+	droppedSeries.WithLabelValues(t.queueName, strconv.FormatUint(ref, 10))
+}
+
+// Keep track of which series we know about for lookups when sending samples to remote.
+func (t *QueueManager) StoreSeries(series []tsdb.RefSeries, index int) {
+	temp := make(map[uint64][]prompb.Label)
```
tomwilkie: This could be initialised with the right size to save on allocs.

storage/remote/queue_manager.go
```diff
@@ -266,6 +360,70 @@ func (t *QueueManager) Stop() {
...
+	}
+	t.seriesMtx.Lock()
+	defer t.seriesMtx.Unlock()
+	for ref, labels := range temp {
+		t.seriesLabels[ref] = labels
+		t.seriesSegmentIndexes[ref] = index
+	}
+}
+
+// Used when reading a checkpoint. Clears the current seri
```
tomwilkie: Unfinished comment?

storage/remote/queue_manager.go
```diff
@@ -428,17 +615,14 @@ func (s *shards) stop(deadline time.Duration) {
 	return
 }
 
-func (s *shards) enqueue(sample *model.Sample) bool {
-	s.qm.samplesIn.incr(1)
-
-	fp := sample.Metric.FastFingerprint()
-	shard := uint64(fp) % uint64(len(s.queues))
-
+func (s *shards) enqueue(sample enqueuable) error {
```
tomwilkie: Maybe just pass ref and timeseries?  And make enquable type local to Append.

vendor/github.com/prometheus/tsdb/wal/wal.go
```diff
@@ -861,6 +861,209 @@ func (r *Reader) Record() []byte {
...
+	return r.hdr
+}
+
+// Returns true if r.rec will contain a full record.
+func (r *LiveReader) Next() bool {
+	for {
+		if r.writeIndex != 0 {
+			r.shiftBuffer()
+		}
+		r.fillBuffer()
```
tomwilkie: No checking error after this, should you be?

vendor/github.com/prometheus/tsdb/wal/wal.go
```diff
@@ -861,6 +861,209 @@ func (r *Reader) Record() []byte {
...
+}
+
+func (r *LiveReader) Header() [7]byte {
+	return r.hdr
+}
+
+// Returns true if r.rec will contain a full record.
+func (r *LiveReader) Next() bool {
+	for {
+		if r.writeIndex != 0 {
```
tomwilkie: `if r.readIndex != 0` would make more sense, no?

vendor/github.com/prometheus/tsdb/wal/wal.go
```diff
@@ -861,6 +861,209 @@ func (r *Reader) Record() []byte {
...
+			r.shiftBuffer()
+		}
+		r.fillBuffer()
+
+		// Do we have data to read?
+		if r.readIndex < r.writeIndex {
```
tomwilkie: Invert this and return early?

vendor/github.com/prometheus/tsdb/wal/wal.go
```diff
@@ -861,6 +861,209 @@ func (r *Reader) Record() []byte {
...
+				}
+			}
+
+			if r.readIndex+recordHeaderSize-1 > r.writeIndex {
+				// Treat this the same as an EOF, it's an error we would expect to see.
+				return false
+			}
+
+			copied := copy(r.hdr[1:], r.buf[readSave:readSave+len(r.hdr[1:])])
+			if copied != recordHeaderSize-1 {
```
tomwilkie: I don't believe this can ever fail.

vendor/github.com/prometheus/tsdb/wal/wal.go
```diff
@@ -899,6 +899,210 @@ func (r *Reader) Offset() int64 {
...
+			return false
+		}
+		r.rec = append(r.rec, temp...)
+		if fullRec {
+			r.index = 0
+			return true
+		}
+		// Only increment i for non-zero records since we use it
+		// to determine valid content record sequences.
+		r.index++
```
tomwilkie: If you pull this logic out into a separate layer, it will be much easier to follow.

vendor/github.com/prometheus/tsdb/wal/wal.go
```diff
@@ -899,6 +899,210 @@ func (r *Reader) Offset() int64 {
...
+	)
+	readTo := length + uint16(readIndex)
+	// safer to cast write index? check if we have enough bytes to read the record data
+	if readTo > uint16(len(buf)) {
+		// Treat this the same as an EOF, it's an error we would expect to see.
+		return nil, 0, io.EOF
+	}
+
+	temp := make([]byte, length)
+	copied := copy(temp, buf[readIndex:readTo])
```
tomwilkie: Why is this a copy?

vendor/github.com/prometheus/tsdb/wal/wal.go
```diff
@@ -899,6 +899,210 @@ func (r *Reader) Offset() int64 {
...
+	total += int64(recordHeaderSize - 1)
+	var (
+		length = binary.BigEndian.Uint16(header[1:])
+		crc    = binary.BigEndian.Uint32(header[3:])
+	)
+	readTo := length + uint16(readIndex)
+	// safer to cast write index? check if we have enough bytes to read the record data
+	if readTo > uint16(len(buf)) {
+		// Treat this the same as an EOF, it's an error we would expect to see.
+		return nil, 0, io.EOF
```
tomwilkie: Why is this EOF? I don't follow.

vendor/github.com/prometheus/tsdb/wal/wal.go
```diff
@@ -899,6 +899,210 @@ func (r *Reader) Offset() int64 {
...
+func (r *LiveReader) Header() [7]byte {
+	return r.hdr
+}
+
+// Returns true if r.rec will contain a full record.
+func (r *LiveReader) Next() bool {
+	r.err = nil
+	for {
+		if r.readIndex != 0 {
+			r.shiftBuffer()
```
tomwilkie: I don't think we need to shift the buffer every time; I think we only need to shift every page size.

vendor/github.com/prometheus/tsdb/wal/wal.go
cstyan: I've broken up the record reading some more, we now have `readRecord` and `buildRecord` used by `Next`.

storage/remote/wal_watcher.go
cstyan: @brian-brazil these metrics seem more useful to me as gauges that reset to 0 when prometheus is restarted or the remote write config is applied again. Let me know if you disagree.

storage/remote/wal_watcher.go
cstyan: Not sure what you're saying here; do the entire function in a single for loop?

storage/remote/wal_watcher.go
cstyan: though they should at least be of the same type as the samples in and samples succeeded metrics, those are both counters atm

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,542 @@
...
+						w.writer.Append(samples)
+						ret = true
+						w.samplesSentPreTailing.Inc()
+					}
+				}
+				if ret {
+					level.Info(w.logger).Log("msg", "found a sample with a timestamp after the WAL watcher start")
+					level.Info(w.logger).Log("msg", "read all series records in segment/checkpoint", "index", index)
+					return
+				}
```
tomwilkie: Won't this massively spam the logs?

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,542 @@
...
+			// because this is the first sample written to the WAL after the WAL watcher was started.
+			typ := dec.Type(rec)
+			if typ == tsdb.RecordSamples {
+				samples, err := dec.Samples(rec, samples[:0])
+				if err != nil {
+					continue
+				}
+				for _, s := range samples {
+					if s.T > w.startTime {
+						w.writer.Append(samples)
```
tomwilkie: I think I'm missing something here; is `samples` contains N samples, won't this try and append them N times?

storage/remote/wal_watcher.go
tomwilkie: Also, why the return here?  Don't we want to try and read more samples?

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,542 @@
...
+	level.Info(w.logger).Log("msg", "starting WAL watcher", "queue", w.writer.Name())
+	go w.runWatcher()
+}
+
+func (w *WALWatcher) Stop() {
+	level.Info(w.logger).Log("msg", "stopping WAL watcher", "queue", w.writer.Name())
+	close(w.quit)
+}
+
+// TODO: fix the exit logic for this function
```
tomwilkie: Is this TODO still valid?

storage/remote/wal_watcher.go
tomwilkie: Oh I see this function is only used at startup.

storage/remote/wal_watcher.go
cstyan: Yes, it should be `Append(s)` :man_facepalming:

storage/remote/wal_watcher.go
```diff
@@ -0,0 +1,565 @@
...
+
+	split := strings.Split(checkpointDir, ".")
+	if len(split) != 2 {
+		return errors.Errorf("checkpoint dir name is not in the right format: %s", checkpointDir)
+	}
+
+	i, err := strconv.Atoi(split[1])
+	if err != nil {
+		i = w.currentSegment - 1
+	}
```
tomwilkie: This will need some exposition as a comment, I don't understand it right now.

cstyan: I don't know what I'm doing here. We should probably be returning that error and ensuring `w.lastCheckpoint` wasn't updated.

tomwilkie:
Done a final pass on this - LGTM!

tomwilkie:
(Going to manually squash some of the changes so we don't spam the commit log)

PR #4589, opened 2018-09-09
Add duplicate-labelset check for range/instant vectors

### LINKED ISSUE #4562: PromQL bug with query_range and metrics name change
## Bug Report
**What did you do?**

* Upgrade node_exporter
* run with query_range: `sum(rate({__name__=~'node_network_receive_drop_total|node_network_receive_drop',env='prod',instance="nip-beprd51"}[5m]) > 0) by(instance,cluster)`

a constant result

when the query range overlaps the metric name change, I git no result when we switch to the new metric

**Environment**

* Prometheus version:

prometheus, version 2.3.2 (branch: HEAD, revision: 71af5e29e815795e9dd14742ee7725682fa14b7b)
  build user:       root@5258e0bd9cc1
  build date:       20180712-14:02:52
  go version:       go1.10.3

**Raw data**

```
{"status":"success","data":{"resultType":"matrix","result":[{"metric":{"__name__":"node_network_receive_drop","cluster":"default","device":"bond0","env":"prod","instance":"nip-beprd51","job":"node_exporters"},"values":[[1535541130.395,"107315172"],[1535541160.395,"107315217"]]},{"metric":{"__name__":"node_network_receive_drop_total","cluster":"default","device":"bond0","env":"prod","instance":"nip-beprd51","job":"node_exporters"},"values":[[1535541190.396,"107315254"],[1535541220.395,"107315281"],[1535541250.395,"107315334"],[1535541280.395,"107315405"],[1535541310.395,"107315457"],[1535541340.395,"107315529"],[1535541370.395,"107315598"],[1535541400.395,"107315649"],[1535541430.395,"107315697"]]}]}}
```

Query range queries with sum() that cross 1535541430-1535541431 are stopped in the middle

query3
query4

sipian:
Closes #4562 

This PR adds a duplicate-label set check for range-vector & instant-vectors.
Signed-off-by: Harsh Agarwal <cs15btech11019@iith.ac.in>

cc @brian-brazil

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -719,6 +719,9 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 		// Make the function call.
 		enh.ts = ts
 		result := f(args, enh)
+		if result.ContainsSameLabelset() {
+			ev.errorf("vector cannot contain metrics with the same labelset")
+		}
 		enh.out = result[:0] // Reuse result vector.
 		// If this could be an instant query, shortcut so as not to change sort order.
 		if ev.endTimestamp == ev.startTimestamp {
@@ -726,9 +729,6 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 			for i, s := range result {
 				s.Point.T = ts
 				mat[i] = Series{Metric: s.Metric, Points: []Point{s.Point}}
-			}
-			if mat.ContainsSameLabelset() {
-				ev.errorf("vector cannot contain metrics with the same labelset")
 			}
 			return mat
 		}
@@ -757,9 +757,6 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 	mat := make(Matrix, 0, len(seriess))
 	for _, ss := range seriess {
 		mat = append(mat, ss)
-	}
-	if mat.ContainsSameLabelset() {
-		ev.errorf("vector cannot contain metrics with the same labelset")
 	}
 	return mat
 }
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -905,7 +905,7 @@ func (ev *evaluator) eval(expr Expr) Value {
 				}
 			}
 			if mat.ContainsSameLabelset() {
-				panic(fmt.Errorf("vector cannot contain metrics with the same labelset"))
+				ev.errorf("vector cannot contain metrics with the same labelset")
 			}
 		}
 		return mat
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -719,9 +719,6 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 		// Make the function call.
 		enh.ts = ts
 		result := f(args, enh)
-		if result.ContainsSameLabelset() {
-			panic(fmt.Errorf("vector cannot contain metrics with the same labelset"))
-		}
 		enh.out = result[:0] // Reuse result vector.
 		// If this could be an instant query, shortcut so as not to change sort order.
 		if ev.endTimestamp == ev.startTimestamp {
@@ -729,6 +726,9 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 			for i, s := range result {
 				s.Point.T = ts
 				mat[i] = Series{Metric: s.Metric, Points: []Point{s.Point}}
+			}
+			if mat.ContainsSameLabelset() {
+				ev.errorf("vector cannot contain metrics with the same labelset")
 			}
 			return mat
 		}
@@ -757,6 +757,9 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 	mat := make(Matrix, 0, len(seriess))
 	for _, ss := range seriess {
 		mat = append(mat, ss)
+	}
+	if mat.ContainsSameLabelset() {
+		ev.errorf("vector cannot contain metrics with the same labelset")
 	}
 	return mat
 }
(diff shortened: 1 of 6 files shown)


promql/value.go
```diff
@@ -140,6 +140,27 @@ func (vec Vector) String() string {
...
+// Such a behaviour is semantically undefined
+// https://github.com/prometheus/prometheus/issues/4562
+func (vec Vector) ContainsSameLabelset() bool {
+	var hashVector []uint64
+	for _, ss := range vec {
+		hashVector = append(hashVector, ss.Metric.Hash())
+	}
+	l := len(vec)
+
+	for i := 0; i < l; i++ {
```
brian-brazil: This is quadratic, use a map to make it nlogn

promql/functions.go
```diff
@@ -278,7 +290,12 @@ func funcHoltWinters(vals []Value, args Expressions, enh *EvalNodeHelper) Vector
 func funcSort(vals []Value, args Expressions, enh *EvalNodeHelper) Vector {
 	// NaN should sort to the bottom, so take descending sort with NaN first and
 	// reverse it.
-	byValueSorter := vectorByReverseValueHeap(vals[0].(Vector))
+	vec := vals[0].(Vector)
+	if vec.ContainsSameLabelset() {
```
brian-brazil: Why are you checking inputs? All the changes needed should be in engine.go

sipian:
I got confused when you mentioned `range vector functions` in [this](https://github.com/prometheus/prometheus/issues/4562#issuecomment-419524190) comment.
I have amended my commit. I'll add any further checks after another review.

Should this check also be added in the [args](https://github.com/prometheus/prometheus/blob/master/promql/engine.go#L717) passed to the promql functions?

brian-brazil:
You still need to check the output of range vector functions.

There's no need to check inputs, as you'll already be checking the outputs.

promql/engine.go
```diff
@@ -719,6 +719,9 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 		// Make the function call.
 		enh.ts = ts
 		result := f(args, enh)
+		if result.ContainsSameLabelset() {
+			panic(fmt.Errorf("vector cannot contain metrics with the same labelset"))
```
brian-brazil: You can use ev.errorf here

promql/engine.go
```diff
@@ -871,6 +874,9 @@ func (ev *evaluator) eval(expr Expr) Value {
 				enh.ts = ts
 				// Make the function call.
 				outVec := e.Func.Call(inArgs, e.Args, enh)
+				if outVec.ContainsSameLabelset() {
```
brian-brazil: OutVec only ever contains at most one element. You want to do this two fors out

promql/engine.go
```diff
@@ -719,9 +719,6 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 		// Make the function call.
 		enh.ts = ts
 		result := f(args, enh)
-		if result.ContainsSameLabelset() {
```
brian-brazil: Having this here was fine

sipian: When I removed the duplicate label-set check in funcLabelReplace to test ContainsSameLabelset.
```
  testmetric{src="source-value-10",dst="original-destination-value"} 0
  testmetric{src="source-value-20",dst="original-destination-value"} 1

  eval_fail instant at 0m label_replace(testmetric, "src", "", "", "")
```
For above test case  adding this check here was not detecting the error because the metrics were passed to the function in separate iterations.
That's why I moved the check outside the loop https://github.com/prometheus/prometheus/pull/4589/commits/fc9efa414cd03de5634c35505e5712385c305d49#diff-83539908220a826e39b7ba3a4f27d5d6R761

sipian:
What is your opinion on removing the duplicate labelset check from label_replace and label_join?
They will anyways be checked later in rangeEval.

promql/engine.go
```diff
@@ -727,6 +727,9 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 				s.Point.T = ts
 				mat[i] = Series{Metric: s.Metric, Points: []Point{s.Point}}
 			}
+			if mat.ContainsSameLabelset() {
```
brian-brazil: You should check result, rather than doing this in two places.

sipian: The tests are failing if I am just checking `result`.
Errors are being caught when the `combined result` is checked.

promql/testdata/functions.test
```diff
@@ -514,3 +514,11 @@ eval instant at 0m days_in_month(vector(1454284800))
 eval instant at 0m days_in_month(vector(1485907200))
   {} 28
 
+clear
+
+# Test duplicate labelset in promql output.
+load 5m
+  testmetric1{src="a",dst="b"} 0
```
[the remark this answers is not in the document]
sipian: No labelset does not contain `__name__`.
Acc to https://www.robustperception.io/whats-in-a-__name__

> If I multiply process_cpu_seconds_total by 0 or take a rate(), the resultant metric is no longer the amount of CPU time in seconds that a process has used since startup - so process_cpu_seconds_total no longer makes sense as a metric name. PromQL doesn't know what name would make sense, so the metric name (contained in the __name__ label) is removed.

promql/engine.go
```diff
@@ -897,6 +904,9 @@ func (ev *evaluator) eval(expr Expr) Value {
 					mat[i].Points[j].V = -mat[i].Points[j].V
 				}
 			}
+			if mat.ContainsSameLabelset() {
+				panic(fmt.Errorf("vector cannot contain metrics with the same labelset"))
```
brian-brazil: ev.errorf

promql/testdata/functions.test
```diff
@@ -514,3 +514,11 @@ eval instant at 0m days_in_month(vector(1454284800))
...
+  testmetric2{src="a",dst="b"} 1
+
+eval_fail instant at 0m changes({__name__=~'testmetric1|testmetric2'}[5m])
```
brian-brazil: Can we test unary minus and a non-range vector function too so we cover all three cases?

sipian:
@brian-brazil PTAL

brian-brazil:
Thanks!

PR #4591, opened 2018-09-10
remote: Added support for stream remote read.

bwplotka:
Design doc: https://docs.google.com/document/d/1JqrU3NjM9HoGLSTPYOvR217f5HBKBiJTqikEB9UiJL0

Changes:
- Changed proto according to design (queryNum and responseType in request)
- Commented to ensure the (Series/Query)Set interface contract is clear
- Chagned remote read handler to support streamed and non streamed responses based on content negotiation (encoders,decoders)
- Changed remote storage to support both types and prefer RAW_STREAMED.
- Put sample limiting inside encoders

Potentially TODO in next PR: Make sure we do pure streaming in `remote.Client.Read`.  Right, now we buffer as we were initially been doing. This is to decrease risk in this PR.

Depends on: https://github.com/prometheus/prometheus/pull/4579

PTAL @brian-brazil @brancz @tomwilkie 

Sorry for big PR, let me know if you want me to split it, but this is full swap to `encoders/decoders` that allows to negotiate stream/nonstream and it seems to work.


prompb/remote.proto
```diff
@@ -24,6 +24,12 @@ message WriteRequest {
 
 message ReadRequest {
   repeated Query queries = 1;
+
+  enum Response {
```
Bplotka: Potentially we could move to `Accept` header negotiation, but it will introduce much more code to ensure backward/forwad compatibility. We have that for free in protobuf instead.

storage/interface.go
```diff
@@ -92,6 +92,29 @@ type Appender interface {
 }
 
 // SeriesSet contains a set of series.
+// It's caller responsibility to ensure it is always emptied (`Next()` is invoked until false)
+// or error is return from `Err()`. This ensures that all used resources are released.
+//
+// Example usage:
```
Bplotka: I hope this was a rule for this interface all the time. Otherwise there is no trival way to implement pure streamed iterators with this that will properly release resources (e.g http body which holds few go routines and TCP conn until you read all and close!). Any thoughts? I feel like this caller responsibility is sane and was that from very beginning. @fabxc ?

storage/remote/client.go
```diff
@@ -148,30 +151,41 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
...
-		return nil, fmt.Errorf("error reading response: %v", err)
-	}
-
-	uncompressed, err := snappy.Decode(nil, compressed)
-	if err != nil {
-		return nil, fmt.Errorf("error reading response: %v", err)
-	}
+	querySet := DecodeReadResponse(httpResp.Header, httpResp.Body)
+
+	// TODO(bplotka): Do not buffer response in concreteSeriesSet. Add caller support for remote.QuerySet or
```
Bplotka: It feels like we should attack this later. We need to ensure callers supports query set or serieSet that you should always `empty` in case of any non series error.

storage/remote/codec.go
```diff
@@ -146,156 +131,44 @@ func FromQuery(req *prompb.Query) (int64, int64, []*labels.Matcher, *storage.Sel
...
+// ToTimeSeries builds a repeated TimeSeries proto.
+func ToTimeSeries(ss storage.SeriesSet) ([]*prompb.TimeSeries, error) {
+	var resp []*prompb.TimeSeries
+
 	for ss.Next() {
-		series := ss.At()
-		iter := series.Iterator()
-		samples := []*prompb.Sample{}
-
-		for iter.Next() {
```
Bplotka: This is done in encoders, based on the info if it's buffered or streamed.

prompb/remote.proto
tomwilkie: We use the headers for versions right now (https://github.com/prometheus/prometheus/blob/master/storage/remote/client.go#L87), so I'd prefer to stick to that for consistency - for instance, we do the correct handling per version in Cortex: https://github.com/weaveworks/cortex/blob/master/pkg/distributor/http_server.go#L17

tomwilkie:
In general I feel like we didn't reach consensus on the design work yet.  I'm still not convinced we can't stream chunks, and that we've given enough thought to horizontal vs vertical "batching".  Perhaps best to do a call with interested parties?

prompb/remote.proto
Bplotka: Can agree, but it won't be that verbose ):

Bplotka:
Cool, we can schedule some call... @tomwilkie 

**In the meantime, regarding consensus:**
a) stream chunks vs stream samples: The initial consensus was that we can always have both. We can extend this to chunks as well in next iterations. But let's discuss it more. Since I explored and implemented the non chunk version, maybe it's  good moment time to dig more the chunk bit https://docs.google.com/document/d/1JqrU3NjM9HoGLSTPYOvR217f5HBKBiJTqikEB9UiJL0/edit?disco=AAAACH4ldEU 
b) horiontal / vertical I created some pros & cons specially for you @tomwilkie feel free to put your opinion on this (:   https://docs.google.com/document/d/1JqrU3NjM9HoGLSTPYOvR217f5HBKBiJTqikEB9UiJL0/edit?disco=AAAACH4ldEY

PR #4600, opened 2018-09-12
changelog: /s/backward compatible/forward compatible/

### LINKED ISSUE #4597: Backward vs forward compatibility
[CHANGELOG.md](https://github.com/prometheus/prometheus/blob/v2.4.0/CHANGELOG.md#240--2018-09-11) currently says:

> Further, the WAL implementation has been re-written so the storage is not backward compatible. Prometheus 2.3 storage will work on 2.4 but not vice-versa.

The first sentence there contradicts the second one. Wikipedia explains [Backward compatibility](https://en.wikipedia.org/wiki/Backward_compatibility#Software) as
> A data format is said to be backward compatible with its predecessor if every message or file that is valid under the old format is still valid, retaining its meaning under the new format.

So storage _is_ backward compatible.

gouthamve:
Fixes #4597 @AlekSi 

Signed-off-by: Goutham Veeramachaneni <gouthamve@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,13 @@
 ## 2.4.0 / 2018-09-11
 
-This release includes multiple bugfixes and features. Further, the WAL implementation has been re-written so the storage is not backward compatible. Prometheus 2.3 storage will work on 2.4 but not vice-versa.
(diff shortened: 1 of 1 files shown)


brian-brazil:
:+1:

gouthamve:
@AlekSi Done

PR #4607, opened 2018-09-14
Make endpoint type configurable in openstack service discovery

d-kuhn:


discovery/openstack/instance.go
```diff
@@ -45,21 +45,22 @@ const (
...
-	region   string
-	interval time.Duration
-	logger   log.Logger
-	port     int
+	authOpts     *gophercloud.AuthOptions
+	region       string
+	interval     time.Duration
+	logger       log.Logger
+	port         int
+	endpointType string
```
brian-brazil: The underlying setting is called Availability, why is this different?

discovery/openstack/instance.go
```diff
@@ -113,8 +114,18 @@ func (i *InstanceDiscovery) refresh() (*targetgroup.Group, error) {
...
 		return nil, fmt.Errorf("could not create OpenStack session: %s", err)
 	}
+
+	availability := gophercloud.AvailabilityPublic
+	switch i.endpointType {
+	case "internal":
+		availability = gophercloud.AvailabilityInternal
+	case "admin":
+		availability = gophercloud.AvailabilityAdmin
+
```
brian-brazil: This should error if there's an invalid value, preferably at config load time.

discovery/openstack/openstack.go
```diff
@@ -96,12 +97,21 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err != nil {
 		return err
 	}
+
+	if c.EndpointType != "" &&
```
simonpasquier: You can set `EndpointType` to `public` in `DefaultSDConfig` and remove this check.

discovery/openstack/instance.go
d-kuhn: I think endpoint_type is the more usual term. It is used for example in the nova command line client. For openstack users it is probably more familiar than availability.

discovery/openstack/instance.go
d-kuhn: I added a check for valid values in 1a24f529a6a1bb3a894cca9ba437bb0d04f787ce

simonpasquier: you can remove this check now that it is enforced when unmarshaling the configuration.

discovery/openstack/openstack.go
```diff
@@ -96,12 +98,18 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err != nil {
 		return err
 	}
+
+	if c.EndpointType != "public" && c.EndpointType != "admin" && c.EndpointType != "internal" {
```
simonpasquier: can you add a test case like in https://github.com/prometheus/prometheus/pull/4594/files?

docs/configuration/configuration.md
```diff
@@ -535,6 +535,9 @@ region: <string>
 # The port to scrape metrics from. If using the public IP address, this must
 # instead be specified in the relabeling rule.
 [ port: <int> | default = 80 ]
+
+# The endpoint type to connect to. Must be one of public, admin or internal.
+ [ endpoint_type <string> | default = public ]
```
simonpasquier: leading space to be removed

discovery/openstack/openstack.go
simonpasquier: You could use the `gophercloud.Availability*` constants directly instead of hard-coded string values.

```go
availability := gophercloud.Availability(c.EndpointType)
if availability != gophercloud.AvailabilityPublic && availability != gophercloud.AvailabilityAdmin && availability != gophercloud.AvailabilityInternal {
    return fmt.Errorf("unknown endpoint_type %q, must be one of %s, %s or %s", c.EndpointType, gophercloud.AvailabilityAdmin, gophercloud.AvailabilityInternal, gophercloud.AvailabilityPublic)
}
```

simonpasquier:
@d-kuhn you've got conflicts.

d-kuhn:
Obsolete because of https://github.com/prometheus/prometheus/pull/7494

PR #4612, opened 2018-09-17
Change discovery subpackages to not use testify in tests

janickic:
#3242 Discovery package

Signed-off-by: Camille Janicki <camille.janicki@gmail.com>

### DIFF
--- a/discovery/kubernetes/kubernetes_test.go
+++ b/discovery/kubernetes/kubernetes_test.go
@@ -22,7 +22,7 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/prometheus/prometheus/discovery/targetgroup"
-	"github.com/stretchr/testify/require"
+	"github.com/prometheus/prometheus/util/testutil"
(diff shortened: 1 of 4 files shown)


janickic:
Need help understanding why CI tests are failing.

janickic:
@gouthamve Would you be able to review? Thanks

grobie:
Thanks a lot for this change @janickic!

It seems testify/suite is still referenced in `vendor/vendor.json`. We use [govendor](https://github.com/kardianos/govendor) to manage our dependencies. Have a look at the `govendor remove` command which should help to completely remove a dependency.

janickic:
Thanks for reviewing @grobie! 

I don't see testify/suite being referenced in the vendor/vendor.json file at all. I reset my last commit and ran `govendor remove github.com/stretchr/testify/suite` and the same files were removed as per my second commit.

grobie:
Oh, you're right. My bad, it should have been in there actually. Anyway, thanks for your contribution!

PR #4619, opened 2018-09-17
Adding alert rule filtering function

huangzhiyong:
The current alert rule is global, and is not very friendly for some special cases. For example, if the CPU idle of the A machine is less than 10, the CPU idle of the B machine is less than 20, and the others are less than 30, we need to write the alert rule three times.

I added the filter function, the tag returned for the global expr matches the filter tag (the match draws on the Alertmanager match idea and some code) and the value returned by expr is compared with the filter to provide a new value, once again filtered. In this way, we can write filter matching (support regular) rules for special cases.

Currently, the filter function has been added, performance testing has not been performed, and test cases have not been perfected.

new alert rule yaml eg:

```
groups:
- name: alertOs
  rules:
  - alert: LowerCPUIdle
    expr: clamp_max(avg by (instance) (sum by (instance,cpu)((clamp_max(rate(node_cpu_seconds_total{mode="idle", job="linux"}[2m]),1)))),1)*100 < 30
    for: 1m
    labels:
      severity: High
    annotations:
      summary: Low CPU Idle On {{ $labels.instance }}
      description: "{{ $labels.instance }} of cpu idle usage has been less than the threshold (current value: {{ $value }}s)"
    filters:
      match:
      - label:
          instance: prometheus
        v: 20
        op: lt
      - label:
          instance: 192.168.1.2
        v_range:
          min: 1
          max: 10

      match_re:
      - label_re:
          instance: op-bigdata.*
        v: 10
        op: lt
      - label_re:
          instance: op-bigdata2.*
          job: a
        v: 15
        op: '<='
```

### DIFF
--- a/discovery/kubernetes/service_test.go
+++ b/discovery/kubernetes/service_test.go
@@ -86,8 +86,8 @@ func TestServiceDiscoveryAdd(t *testing.T) {
 				Targets: []model.LabelSet{
 					{
 						"__meta_kubernetes_service_port_protocol": "TCP",
-						"__address__":                             "testservice.default.svc:30900",
-						"__meta_kubernetes_service_port_name":     "testport",
+						"__address__":                         "testservice.default.svc:30900",
+						"__meta_kubernetes_service_port_name": "testport",
 					},
 				},
 				Labels: model.LabelSet{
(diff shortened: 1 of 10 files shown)


brian-brazil:
I think you're looking for https://www.robustperception.io/using-time-series-as-alert-thresholds

huangzhiyong:
This is good advice.
I'll try。
Thank you

PR #4620, opened 2018-09-17
Update to current client_golang

beorn7:
@gouthamve @brancz This is mostly to get the new client_golang code into our own main repos before releasing it as v0.9.

There was only one use of a deprecated type, and that only in an example. I fixed it in this PR, too.

We _could_ now, of course, prefix the TSDB metrics with `prometheus_` here (by wrapping) and remove it in prometheus/tsdb. @gouthamve as you are the expert of using prometheus/tsdb as a library, is this something desired, i.e. give users of prometheus/tsdb the freedom of prefixing their TSDB metrics with something else (or nothing, i.e. leave it at `tsdb_...`)? One might also argue every user of prometheus/tsdb should use the same metric names. The crucial question here is: Do all the `prometheus_tsdb_...` metrics mean the same, even if prometheus/tsdb is used as a library in something that is not Prometheus?


gouthamve:
Oh wow, https://godoc.org/github.com/prometheus/client_golang/prometheus#WrapRegistererWithPrefix has been something I have been looking forward to since ages!
I think the only reason we have the `prometheus_` prefix down in TSDB is because this function is missing and I think it only makes sense to remove it there and wrap it here.

> Do all the `prometheus_tsdb_...` metrics mean the same, even if prometheus/tsdb is used as a library in something that is not Prometheus?
Yes, but prepending it with `prometheus_` could be confusing if people are using it for generic applications.

👍 to this PR though!

beorn7:
> > Do all the `prometheus_tsdb_...` metrics mean the same, even if prometheus/tsdb is used as a library in something that is not Prometheus?
>
> Yes, but prepending it with `prometheus_` could be confusing if people are using it for generic applications.

What was the reason then for prefixing them with `prometheus_`? (Real question. I'd like to understand the reasoning so that we can make an informed decision if we want `tsdb_...` only in the prometheus/tsdb repo but then still prefix with `prometheus_...` for Prometheus as a user of the TSDB library.)

gouthamve:
IIRC it was to make sure that all metrics that are prometheus internal are prefixed with `prometheus_`. It was initially `tsdb_` but the change was made in this PR: https://github.com/prometheus/tsdb/pull/184

Maybe @fabxc @brancz can clarify.

brancz:
I don't think `_tsdb` is sufficient for namespacing, while `prometheus_tsdb_` clearly tells you this is concerning a Prometheus tsdb. I think there are arguments to both sides, but I think there are going to be less confusions with the `prometheus_tsdb_` prefix, plus the metrics have been around for almost a year now and it would be nice not to break them :slightly_smiling_face: .

brancz:
this lgtm

beorn7:
I'll create a separate PR in prometheus/tsdb for the un-prefixing. Objections can still be raised there.

Merging this for now.

grobie:
@beorn7 the prefixing was the result of a longer discussion during release and the amount of confusion and support requests it caused when users couldn't find Prometheus metrics. I don't think we should reverse that decision.

beorn7:
The resulting metrics will still look like `prometheus_tsdb_...` on the Prometheus server. The only question is if it should be hardcoded into the TSDB library or added by the actual Prometheus server code using https://godoc.org/github.com/prometheus/client_golang/prometheus#WrapRegistererWithPrefix

brian-brazil:
> The only question is if it should be hardcoded into the TSDB library

It should be hardcoded into the TSDB library. Applications should not prefix metrics of the libraries they use, except in extremely rare circumstances as that prevents horizontal monitoring - that's what target labels are for.

PR #4623, opened 2018-09-17
Update vendoring of prometheus/tsdb

beorn7:
The new version has no prometheus_ metrics prefix hardcoded anymore,
so this commit adds it using the new WrapRegistererWithPrefix
function.

Signed-off-by: beorn7 <beorn@soundcloud.com>

@gouthamve @brancz @grobie as promised. The metrics look the same as before.

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -555,7 +555,10 @@ func main() {
 				db, err := tsdb.Open(
 					cfg.localStoragePath,
 					log.With(logger, "component", "tsdb"),
-					prometheus.DefaultRegisterer,
+					prometheus.WrapRegistererWithPrefix(
+						"prometheus_",
+						prometheus.DefaultRegisterer,
+					),
 					&cfg.tsdb,
 				)
 				if err != nil {
--- a/vendor/github.com/prometheus/tsdb/block.go
+++ b/vendor/github.com/prometheus/tsdb/block.go
@@ -15,6 +15,7 @@
 package tsdb
 
 import (
+	"encoding/binary"
 	"encoding/json"
 	"io/ioutil"
 	"os"
@@ -248,6 +249,10 @@ type Block struct {
 	dir  string
 	meta BlockMeta
 
+	// Symbol Table Size in bytes.
+	// We maintain this variable to avoid recalculation everytime.
+	symbolTableSize uint64
+
 	chunkr     ChunkReader
 	indexr     IndexReader
 	tombstones TombstoneReader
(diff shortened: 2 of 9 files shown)


brancz:
The metrics changes lgtm, but someone with a closer look on tsdb has to decide about the other tsdb changes that this pulls in.

beorn7:
Metric excerpt:

```
# HELP prometheus_build_info A metric with a constant '1' value labeled by version, revision, branch, and goversion from which prometheus was built.
# TYPE prometheus_build_info gauge
prometheus_build_info{branch="beorn7/client_golang",goversion="go1.10.2",revision="e0be02b12cbe51fc1a13824e9bdde377f395ed3d",version="2.4.0"} 1
# ...
# HELP prometheus_tsdb_blocks_loaded Number of currently loaded data blocks
# TYPE prometheus_tsdb_blocks_loaded gauge
prometheus_tsdb_blocks_loaded 182
# HELP prometheus_tsdb_compaction_chunk_range_seconds Final time range of chunks on their first compaction
# TYPE prometheus_tsdb_compaction_chunk_range_seconds histogram
# ...
```

cmd/prometheus/main.go
```diff
@@ -555,7 +555,10 @@ func main() {
 				db, err := tsdb.Open(
 					cfg.localStoragePath,
 					log.With(logger, "component", "tsdb"),
-					prometheus.DefaultRegisterer,
+					prometheus.WrapRegistererWithPrefix(
```
brian-brazil: I am against this, applications should not add prefixes to the libraries they happen to use - that's what target labels are for.

beorn7: In the discussions that preceded this change (see https://github.com/prometheus/tsdb/pull/383 and in particular #4620), the understanding of the people in charge of this (@brancz @gouthamve @grobie) seems to be that the metrics provided by the TSDB library mean sufficiently different things if it is used for Prometheus vs. high frequency trading or unicorn counting so that they should have different prefixes. If the meaning changes, the metric name should change, not the target label. Thus, the important question is if the meaning actually changes or rather not, and I [asked it in that PR](https://github.com/prometheus/prometheus/pull/4620#issue-215945608) (last sentence).

I don't really have an own idea about the answer to this question, which I'd like to defer to @brancz @gouthamve @grobie @fabxc , who have all the history. Do you want to change the answer given in #4620 in light of the concern @brian-brazil has raised?

brian-brazil: I'd view it as the same as process_.

The case for adding a prefix is when you've got sharding (vertical or horizontal) and are likely also adding labels to distinguish the various different tsdb instances.

grobie: To be honest I don't really care and my only comments were about requesting not to change the `prometheus_tsdb_` prefix currently present (which you never planned to change in the end).

I see arguments for both sides, but would ultimately agree with @brian-brazil and leave the prefix in place.

brancz: I also think @brian-brazil's argument is fair.

beorn7: Thanks for all the feedback. I haven't heard from @gouthamve yet, but I guess if he felt strongly, he would have chimed in by now.
I'll close this PR and revert the change in prometheus/tsdb.

PR #4625, opened 2018-09-18
fix k8s config example

CaoShuFeng:
tls_config is an element of kubernetes_sd_configs:
https://github.com/prometheus/prometheus/blob/cfbd72b4f1b5257ab7e0074bfab0631168747556/discovery/kubernetes/kubernetes.go#L91

### DIFF
--- a/documentation/examples/prometheus-kubernetes.yml
+++ b/documentation/examples/prometheus-kubernetes.yml
@@ -20,27 +20,26 @@ scrape_configs:
 
   kubernetes_sd_configs:
   - role: endpoints
+    # This TLS & bearer token file config is used to connect to the actual scrape
+    # endpoints for cluster components. This is separate to discovery auth
+    # configuration because discovery & scraping are two separate concerns in
+    # Prometheus. The discovery auth config is automatic if Prometheus runs inside
+    # the cluster. Otherwise, more config options have to be provided within the
+    # <kubernetes_sd_config>.
+    tls_config:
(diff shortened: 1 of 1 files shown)


jojohappy:
Maybe it is not correctly. The `TLSConfig` in `kubernetes_sd_config` will be used with the `api_server` provided if the Prometheus runs out of the Kubernetes cluster. 

The `TLSConfig` in the example will be used to the `scrape` requests with SSL if the scheme of `target` is **https**, IMO, the example is used when the Prometheus integrates in Kubernetes cluster, so it is right now.

CaoShuFeng:
> The TLSConfig in kubernetes_sd_config

Yes, TLSConfig is an element of kubernetes_sd_config. But now TLSConfig and kubernetes_sd_config are in the same level. After I try to convert the yml to json, we could see it clearly:
```json
            "tls_config": {
                "ca_file": "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
            }, 
            "kubernetes_sd_configs": [
                {
                    "role": "endpoints"
                }
            ]
```

The right format should be:
```json
            "kubernetes_sd_configs": [
                {
                    "role": "endpoints", 
                    "tls_config": {
                        "ca_file": "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
                    }
                }
            ], 

```

simonpasquier:
Thanks @jojohappy for the detailed explanation which is absolutely correct. 
I'm closing it for now. If you have further questions, please use our [user mailing list](https://groups.google.com/forum/#!forum/prometheus-users), which you can also search.

PR #4626, opened 2018-09-18
*: move to go 1.11

simonpasquier:


gouthamve:
While I'm not against it, it is worth noting that this is going to break a lot of kubernetes users badly (including me). See: https://github.com/golang/go/issues/27546

grobie:
@gouthamve Why? Prometheus doesn't use the native go DNS client.

gouthamve:
@grobie I didn't know that. We upgraded Cortex and that broke some stuff, best test it out though. I'll deploy this one soon.

simonpasquier:
Yeah, I would definitely be cautious with this change: DNS + Go is a finicky topic.

sipian:
@simonpasquier 
A big refactoring was merged into prombench just now. https://github.com/prometheus/prombench/pull/139
Cheers for testing it successfully. :raised_hands:

simonpasquier:
Regarding the DNS service discovery, @grobie noted correctly that it uses https://github.com/miekg/dns instead of the native Go client so it should be immune to https://github.com/golang/go/issues/27546. I did a test with a domain query returning a bunch of SRV records and it worked fine. Again not saying that it is sufficient and additional testing would be much appreciated.

simonpasquier:
As mentioned on IRC, Prometheus v2.4.3 is already built with go1.11.1 as promu had [switched](https://github.com/prometheus/promu/pull/108) its default Go version to 1.11 before v2.4.3 was released. To prevent this from happening again, this PR also sets the Go version in `.promu.yml`.

simonpasquier:
@hoffie thanks for the pointer! I double-checked and the benchmark didn't anything particular wrt virtual memory utilization:

http://prombench.prometheus.io/prometheus-meta/graph?g0.range_input=6h&g0.end_input=2018-10-09%2014%3A00&g0.expr=process_virtual_memory_bytes%7Bjob%3D%22prometheus%22%2C%20namespace%3D%22prombench-4626%22%7D&g0.tab=0

simonpasquier:
As noted [above](https://github.com/prometheus/prometheus/pull/4626#issuecomment-428217498), v2.4.3 has already been built with go1.11.1 so I'd say that this PR is ok to go.

@gouthamve @brian-brazil WDYT?

PR #2572, opened 2017-04-04
Add promql.ErrStorage, which the API propagates as a 500.

tomwilkie:
Fixes #2571 - see there for discussion.

PR #4628, opened 2018-09-18
Make ErrStorage a concrete type not an interface

### LINKED ISSUE #4629: All errors come back as internal errors from api.query()
[This type assertion](https://github.com/prometheus/prometheus/blob/3e87c04/web/api/v1/api.go#L265-L271) matches `ErrorStorage` for all types of error, because `ErrorStorage` is an alias to the error interface.

https://golang.org/ref/spec#Type_assertions

~~I believe this is the cause of https://github.com/weaveworks/cortex/issues/780~~, though I actually found it while chasing something else.

bboreham:
Fixes #4629 

Since it is [used in a type assertion](https://github.com/prometheus/prometheus/blob/3e87c04/web/api/v1/api.go#L265-L271), having `ErrStorage` as an alias to the error interface is the same as saying `error`, i.e. it succeeds for all types of error.  Change to a struct which is a concrete type and the type assertion will only succeed if the type is identical.

https://golang.org/ref/spec#Type_assertions

This change will break downstream projects that are using the type, but I argue they can't be using it to any great benefit as it stands.

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -74,7 +74,7 @@ type (
 	ErrTooManySamples string
 	// ErrStorage is returned if an error was encountered in the storage layer
 	// during query handling.
-	ErrStorage error
+	ErrStorage struct{ error }
 )
 
 func (e ErrQueryTimeout) Error() string {
@@ -85,6 +85,9 @@ func (e ErrQueryCanceled) Error() string {
 }
 func (e ErrTooManySamples) Error() string {
 	return fmt.Sprintf("query processing would load too many samples into memory in %s", string(e))
+}
+func (e ErrStorage) Error() string {
+	return e.error.Error()
 }
 
 // A Query is derived from an a raw query string and can be run against an engine
--- a/promql/engine_test.go
+++ b/promql/engine_test.go
@@ -193,7 +193,7 @@ func TestQueryError(t *testing.T) {
 		Timeout:       10 * time.Second,
 	}
 	engine := NewEngine(opts)
-	errStorage := ErrStorage(fmt.Errorf("storage error"))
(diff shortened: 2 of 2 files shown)


simonpasquier:
:+1:

gouthamve:
@bboreham This actually ends up fixing https://github.com/cortexproject/cortex/issues/1056 Could you rebase?

I could take over if you're busy.

gouthamve:
Further, this needs to be updated too: https://github.com/prometheus/prometheus/blob/3e87c04/web/api/v1/api.go#L338-L346

A `returnApiError` function to dedupe the logic maybe?

bboreham:
> this needs to be updated too

I mentioned in #4629 it is inconsistent; that is surely a different issue though?

gouthamve:
Yup, but an issue nonetheless :)

bboreham:
I will leave it to someone who thinks they know what the requirements are to specify that other issue.

gouthamve:
@brian-brazil I see that https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L268-L278 and https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L342-L350 differ. I think they should be the same.

As your were the last to touch those lines, could you comment on why they differ?

brian-brazil:
This is actually from #2572, it appears to be an oversight.

gouthamve:
@bboreham ping?

gouthamve:
LGTM. I'll open another PR for the inconsistency.

tomwilkie:
Unfortunately `struct ErrStorage{ error }` can't be used from outside the promql package, as `error` is an unexported field...

PR #4632, opened 2018-09-19
release 2.4.1

gouthamve:
We have an enhancement in there because we don't have a release branch in `prometheus/tsdb`. Will be fixed for the 2.5 release. Not sure if we should mention that though.

depends on #4631


brancz:
I have little time this week to look at anything, can we just make sure that the tsdb metrics are still correct, as I'm unsure whether we've cleaned up everything in prometheus/prometheus after deciding to revert one of the PRs @beorn7 did against prometheus/tsdb.

beorn7:
The changes in prometheus/prometheus were never merged. We should be good metrics-wise.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.4.1 / 2018-09-19
+
+* [ENHANCEMENT] New TSDB metrics prometheus/tsdb#375 prometheus/tsdb#363
+* [BUGFIX] Render UI correctly for windows #4616
```
simonpasquier: s/windows/Windows/

gouthamve: Fixed, thanks!

PR #4639, opened 2018-09-19
Move to testutil in the remaining packages that use testify

### LINKED ISSUE #3242: tests: move tests to testutil
tsdb is using the require library of [testify](https://github.com/stretchr/testify) and its much less verbose and cleaner. Should we also migrate the tests everywhere?

/cc @brian-brazil @grobie @fabxc

csmarchbanks:
Related to but does not close (packages like discovery still use `t.Errorf`): #3242

The only changes to behavior are:
* Pulled NewStorage out of `testutil` and into its own package (`util/teststorage`) to avoid a circular dependency between `storage` and `testutil`. This is in its own commit.
* Added a way to add messages to testutil.Assert since we used this functionality in several places, very similar to how prometheus/tsdb does it.


csmarchbanks:
Bump for a review. Perhaps @brian-brazil or @gouthamve?

util/testutil/testing.go
```diff
@@ -62,9 +63,16 @@ func NotOk(tb TB, err error, format string, a ...interface{}) {
...
-		tb.Fatalf("\033[31m\nexp: %#v\n\ngot: %#v\033[39m\n", exp, act)
+		tb.Fatalf("\033[31m\nexp: %#v\n\ngot: %#v%s\033[39m\n", exp, act, formatMessage(msgAndArgs))
 	}
 }
+
+func formatMessage(msgAndArgs []interface{}) string {
+	if len(msgAndArgs) == 0 {
+		return ""
+	}
+	return fmt.Sprintf(fmt.Sprintf("\n\nmsg: %s", msgAndArgs[0]), msgAndArgs[1:]...)
```
brian-brazil: This looks odd, why are you taking the first argument as a format string?

csmarchbanks: The goal was to have an optional message that could have formatters in it, similar to the [tesitfy equal](https://github.com/stretchr/testify/blob/master/assert/assertions.go#L328).

I find that much of the time a message is not needed for an `Equals`, but sometimes you want to add the extra info, and passing an empty string everywhere that uses `Equals` is not a pattern I like.

If that pattern is ok, perhaps I should add a comment to Equals explaining how to format messages?

csmarchbanks:
> Did you forget to push the change moving the package?
Looks like it is there. Github has the package move displayed as a file move "util/testutil/storage.go → util/teststorage/storage.go"

util/testutil/testing.go
```diff
@@ -62,9 +63,16 @@ func NotOk(tb TB, err error, format string, a ...interface{}) {
 }
 
 // Equals fails the test if exp is not equal to act.
-func Equals(tb TB, exp, act interface{}) {
+func Equals(tb TB, exp, act interface{}, msgAndArgs ...interface{}) {
 	tb.Helper()
 	if !reflect.DeepEqual(exp, act) {
-		tb.Fatalf("\033[31m\nexp: %#v\n\ngot: %#v\033[39m\n", exp, act)
+		tb.Fatalf("\033[31m\nexp: %#v\n\ngot: %#v%s\033[39m\n", exp, act, formatMessage(msgAndArgs))
```
codesome: I feel printing the message before `exp` and `got` would be better here.

csmarchbanks: Added a commit to move the message first, similarly to tsdb

himanshu-dixit:
@codesome @csmarchbanks Is this still being worked upon?

csmarchbanks:
I resurrected this PR, please take another look when you get a chance!

PR #4642, opened 2018-09-20
WIP: support get alerts by some filters

knightXun:
support filter alerts when try to get them by /api/v1/alerts

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -611,10 +611,18 @@ type Alert struct {
 }
 
 func (api *API) alerts(r *http.Request) (interface{}, *apiError, func()) {
+	alertName := r.FormValue("alertname")
 	alertingRules := api.rulesRetriever.AlertingRules()
+	instanceName := r.FormValue("instance")
 	alerts := []*Alert{}
 
 	for _, alertingRule := range alertingRules {
+		if alertName != "" && alertingRule.Name() != alertName {
+			continue
+		}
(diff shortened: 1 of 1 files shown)


knightXun:
Are this alerts be stored in tsdb? I also want to support historic alerts searching

knightXun:
This is existed in api query.Alert is already stored in tsdb

simonpasquier:
Alerts are stored in the `ALERTS` metric. Not sure if we want to increase the scope of `/api/v1/alerts` (cc @mxinden) as you could also filter client-side. But if we agree to, one should be able to filter on any label, not just `instance`.

Also please don't close & open a PR as it creates artificial noise.

knightXun:
I think api for query history alerts is needed. The prometheus web only display the firing alerts.Users can't see historic alerts. I am working on this now. If community aggree this feature, I will be helpful.
/cc @simonpasquier @mxinden

mxinden:
As this is not a very data intensive endpoint I prefer client side filtering, but this is not a strong opinion.

In regards to the history of alerts, as @simonpasquier mentioned above, you might want to look into the `ALERTS` metric, which exposes historical data on firing alerts.

codesome:
> As this is not a very data intensive endpoint I prefer client side filtering, but this is not a strong opinion.

We decided during the bug scrub that we don't need this for the above reason. If there is a good use case for the filter, please open an issue with the details.

PR #4647, opened 2018-09-21
Do not override GOOS and GOARCH

jannickfahlbusch:
The two variables should be in control by the user, not the Makefile.

I wanted to compile Prometheus on a Mac for Linux but was unable to do this, as the Makefile overrides my variables.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -121,12 +121,12 @@ common-docker-tag-latest:
 
 .PHONY: promu
 promu:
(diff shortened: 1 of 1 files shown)


grobie:
Our current build tooling requires these env settings, so we can't merge this change. I suggest to use promu which supports cross compilation already.

jannickfahlbusch:
Okay, that is sad :(

I'll close this PR then.

PR #4649, opened 2018-09-24
discovery/openstack: discover all interfaces

### LINKED ISSUE #4339: openstack_sd chosing random network-interface
## Bug Report
**What did you do?**

I set up openstack_sd with the below config. 

openstack_sd should discover nodes and use consistently use the same network-interface (10.x.x.x) on all servers.

openstack_sd successfully added all available nodes, but the chose network interface seems arbitrary. it also changes on each discovery run (every 60 seconds)

out of 44 nodes, between 6 and 10 get "re-discovered" with a new __private_ip__ every 60 seconds.

** examples **

before:

before

__address__="10.150.x.x:9100"
__meta_openstack_instance_flavor="1310a6bc-3b91-47d0-ad50-9a50b9cf743f"
__meta_openstack_instance_id="1cda4a6c-fc4a-4de5-9cc5-13bba908333f"
__meta_openstack_instance_name="xxx"
__meta_openstack_instance_status="ACTIVE"
__meta_openstack_private_ip="10.150.x.x"
__metrics_path__="/metrics"
__scheme__="http"
job="openstack_dc1"
	
after:

after

(marked in yellow are the two servers from the 'before' picture)

__address__="100.100.x.x:9100"
__meta_openstack_instance_flavor="1310a6bc-3b91-47d0-ad50-9a50b9cf743f"
__meta_openstack_instance_id="1cda4a6c-fc4a-4de5-9cc5-13bba908333f"
__meta_openstack_instance_name="xxx"
__meta_openstack_instance_status="ACTIVE"
__meta_openstack_private_ip="100.100.x.x"
__metrics_path__="/metrics"
__scheme__="http"
job="openstack_dc1"
	

**Environment**

* System information:

Linux 4.4.70 x86_64`

* Prometheus version:

Version 	1.8.1
Revision 	3a7c51ab70fc7615cd318204d3aa7c078b7c5b20
Branch 	HEAD
BuildUser 	root@ab78fb101775
BuildDate 	20171023-15:50:57
GoVersion 	go1.9.1

* Prometheus configuration file: (relevant part)
```
- job_name: 'openstack_dc1'
    scrape_interval: 1m
    scrape_timeout: 10s
    metrics_path: /metrics
    scheme: http
    tls_config:
     insecure_skip_verify: true
    openstack_sd_configs:
    - identity_endpoint: 'https://xxx:5000/v2.0'
      username: <secret>
      userid: <secret>
      password: '<secret>'
      project_name: xxx
      project_id: xxx
      domain_name: default
      domain_id: default
      role: instance
      region: RegionOne
      refresh_interval: 1m
      port: 9100
```

simonpasquier:
Closes #4339 

This PR discovers one target per network interface of Nova instance instead of picking one arbitrary IP address. I've tested with a local install and it works fine. If a private IP address is associated to a floating address, `__meta_openstack_public_ip` will be set. The only trick is to skip any floating IP address that may be listed in the `Addresses` map.

cc @alfarme that filed the corresponding issue.

### DIFF
--- a/discovery/openstack/instance.go
+++ b/discovery/openstack/instance.go
@@ -129,7 +129,7 @@ func (i *InstanceDiscovery) refresh() (*targetgroup.Group, error) {
 	// https://developer.openstack.org/api-ref/compute/#list-floating-ips
 	pagerFIP := floatingips.List(client)
 	floatingIPList := make(map[floatingIPKey]string)
-	fipPresent := make(map[string]struct{})
+	floatingIPPresent := make(map[string]struct{})
 	err = pagerFIP.EachPage(func(page pagination.Page) (bool, error) {
 		result, err := floatingips.ExtractFloatingIPs(page)
 		if err != nil {
@@ -141,7 +141,7 @@ func (i *InstanceDiscovery) refresh() (*targetgroup.Group, error) {
 				continue
 			}
 			floatingIPList[floatingIPKey{id: ip.InstanceID, fixed: ip.FixedIP}] = ip.IP
-			fipPresent[ip.IP] = struct{}{}
+			floatingIPPresent[ip.IP] = struct{}{}
 		}
 		return true, nil
 	})
@@ -169,10 +169,10 @@ func (i *InstanceDiscovery) refresh() (*targetgroup.Group, error) {
 			}
 
 			labels := model.LabelSet{
-				openstackLabelInstanceID: model.LabelValue(s.ID),
+				openstackLabelInstanceID:     model.LabelValue(s.ID),
+				openstackLabelInstanceStatus: model.LabelValue(s.Status),
+				openstackLabelInstanceName:   model.LabelValue(s.Name),
 			}
-			labels[openstackLabelInstanceStatus] = model.LabelValue(s.Status)
-			labels[openstackLabelInstanceName] = model.LabelValue(s.Name)
 
 			id, ok := s.Flavor["id"].(string)
 			if !ok {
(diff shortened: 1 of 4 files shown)


simonpasquier:
@brian-brazil please have a look.

discovery/openstack/instance.go
```diff
@@ -123,17 +128,20 @@ func (i *InstanceDiscovery) refresh() (*targetgroup.Group, error) {
 	// OpenStack API reference
 	// https://developer.openstack.org/api-ref/compute/#list-floating-ips
 	pagerFIP := floatingips.List(client)
-	floatingIPList := make(map[string][]string)
+	floatingIPList := make(map[floatingIPKey]string)
+	fipPresent := make(map[string]struct{})
```
brian-brazil: does f here stand for floating or fixed?

discovery/openstack/instance.go
```diff
@@ -155,13 +163,27 @@ func (i *InstanceDiscovery) refresh() (*targetgroup.Group, error) {
...
+				level.Info(i.logger).Log("msg", "Got no IP address", "instance", s.ID)
+				continue
+			}
+
 			labels := model.LabelSet{
 				openstackLabelInstanceID: model.LabelValue(s.ID),
 			}
-			if len(s.Addresses) == 0 {
-				level.Info(i.logger).Log("msg", "Got no IP address", "instance", s.ID)
+			labels[openstackLabelInstanceStatus] = model.LabelValue(s.Status)
```
brian-brazil: These can go in the above initilisation

simonpasquier:
> If a user just wanted one of the IPs to be monitored, how would they filter it down?

I imagine that they would use a relabel_config rule that matches the prefix of the IP address.

brian-brazil:
> I imagine that they would use a relabel_config rule that matches the prefix of the IP address.

That's not going to work too nicely if they have CIDR prefixes.

PR #4653, opened 2018-09-25
[WIP] Block time ranges

krasi-georgiev:
testing what(if anything) will break when we create blocks with exact time ranges.

In relation to https://github.com/prometheus/tsdb/issues/346

~will add the exact use case a bit later.~ the added test explains the use case.

### DIFF
--- a/vendor/github.com/prometheus/tsdb/db.go
+++ b/vendor/github.com/prometheus/tsdb/db.go
@@ -382,41 +382,27 @@ func (db *DB) compact() (changes bool, err error) {
 			return changes, nil
 		default:
 		}
-
-		minHead := db.head.MinTime()
-		maxHead := db.head.MaxTime()
-		maxBlock := minHead + db.opts.BlockRanges[0]
-
-		// // Make sure we have at least 30% of db.opts.BlockRanges[0] remains in the head
-		// // so we can append samples with timestamp in the recent past.
-		if maxHead-minHead <= db.opts.BlockRanges[0]+(db.opts.BlockRanges[0]/3) {
-			break
-		}
-
 		// The head has a compactable range if 1.5 level 0 ranges are between the oldest
 		// and newest timestamp. The 0.5 acts as a buffer of the appendable window.
-		// if db.head.MaxTime()-db.head.MinTime() <= db.opts.BlockRanges[0]/2*3 {
-		// 	break
-		// }
-
-		// mint, maxt := rangeForTimestamp(db.head.MinTime(), db.opts.BlockRanges[0])
+		if db.head.MaxTime()-db.head.MinTime() <= db.opts.BlockRanges[0]/2*3 {
+			break
+		}
+		mint := db.head.MinTime()
+		_, maxt := rangeForTimestamp(mint, db.opts.BlockRanges[0])
 
 		// Wrap head into a range that bounds all reads to it.
 		head := &rangeHead{
 			head: db.head,
-			mint: minHead,
-			// mint: mint,
+			mint: mint,
 			// We remove 1 millisecond from maxt because block
 			// intervals are half-open: [b.MinTime, b.MaxTime). But
 			// chunk intervals are closed: [c.MinTime, c.MaxTime];
 			// so in order to make sure that overlaps are evaluated
 			// consistently, we explicitly remove the last value
 			// from the block interval here.
-			maxt: maxBlock - 1,
-			// maxt: maxt - 1,
+			maxt: maxt - 1,
 		}
-		if _, err = db.compactor.Write(db.dir, head, minHead, maxBlock, nil); err != nil {
-			// if _, err = db.compactor.Write(db.dir, head, mint, maxt, nil); err != nil {
+		if _, err = db.compactor.Write(db.dir, head, mint, maxt, nil); err != nil {
 			return changes, errors.Wrap(err, "persist head block")
 		}
 		changes = true
@@ -834,7 +820,6 @@ func (db *DB) Querier(mint, maxt int64) (Querier, error) {
 func rangeForTimestamp(t int64, width int64) (mint, maxt int64) {
 	mint = (t / width) * width
 	return mint, mint + width
-
 }
 
 // Delete implements deletion of metrics. It only has atomicity guarantees on a per-block basis.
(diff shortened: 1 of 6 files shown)


krasi-georgiev:
@bwplotka , @brancz so far it seems that it all works as expected.

 the only difference is that the compaction happens less often, but I will change few more lines to make the compaction happen at the same intervals as before.

Next I will run tests to see how long it will take to convert the wal into a block.


Logs
2h block
https://pastebin.com/raw/8aAN3thS
blocks with exact sizes.
https://pastebin.com/raw/0br8u8A3

http://prombench.prometheus.io/grafana/d/7gmLoNDmz/prombench?panelId=34&orgId=1&var-RuleGroup=All&var-pr-number=4653&from=now-2d&to=now

brancz:
I think it would be a good idea to get a review on this by @fabxc as as I recall there were some intentions to why the times were not exact, but I might be mis-remembering.

fabxc:
Can I get some context here on what those changes intend to do? I see two critical pieces of code commented out entirely, which certainly raises some alarms.

krasi-georgiev:
@fabxc in relation to https://github.com/prometheus/tsdb/issues/346

 - making sure that the wal contains only samples after the maxtime of the last block.
 - making sure that the compaction creates blocks with mintime of the first sample in the WAL


this is to achieve to following.
1. Stop Prometheus (scale down) - Last block is from 14.00-16.00
1. Convert the wal to a block (the wal needs to contain only data after 16.00 to avoid the created block to  overlap). The converted block contains data from 16.00 to 17.20
1. Upload the blocks.
1. Delete the wal.
1. Start Prometheus at 17.30. Compaction should create a block from the first sample of the wal to avoid overlapping with the uplloaded 16.00 - 17.20 block.

I have run prombench with all these changes and all worked as expected.

vendor/github.com/prometheus/tsdb/db.go
```diff
@@ -382,26 +382,41 @@ func (db *DB) compact() (changes bool, err error) {
 			return changes, nil
 		default:
 		}
-		// The head has a compactable range if 1.5 level 0 ranges are between the oldest
-		// and newest timestamp. The 0.5 acts as a buffer of the appendable window.
-		if db.head.MaxTime()-db.head.MinTime() <= db.opts.BlockRanges[0]/2*3 {
+
+		minHead := db.head.MinTime()
+		maxHead := db.head.MaxTime()
+		maxBlock := minHead + db.opts.BlockRanges[0]
```
fabxc: This will cause blocks to cross the virtual block boundaries aligned at each `db.opts.BlockRanges[0]`. This discards a very central paradigm that lot of TSDB assumes as an invariant.

I _think_ the compactor today might just skip such blocks and then later, at a higher-level compaction catch those blocks. 
But I'm by no means sure about that. TSDB certainly wasn't build with this in mind and it will almost guaranteed cause trouble somewhere.

fabxc:
> @fabxc in relation to [prometheus/tsdb#346](https://github.com/prometheus/tsdb/issues/346)
> 
>     * making sure that the wal contains only samples after the maxtime of the last block.

How do these changes ensure that?
I'm seeing that checkpointing now applies to all but the currently active segment. That's equivalent to a full rewrite on every call to `Truncate` (that comes with a cost but may be ok).

This does _not_ ensure that the WAL never contains samples after the most recent block. The active segment is not considered and for a small server, it may very well contain data for a few blocks into the past.

> * making sure that the compaction creates blocks with mintime of the first sample in the WAL

The compaction with changes I see here still creates blocks with the min time of the current head, not the WAL (which doesn't have a getter for min time I think).

> this is to achieve to following.
> 
>     1. Stop Prometheus (scale down) - Last block is from 14.00-16.00
> 
>     2. Convert the wal to a block (the wal needs to contain only data after 16.00 to avoid the created block to  overlap). The converted block contains data from 16.00 to 17.20

This would be done by a CLI tool? I don't follow what part those TSDB runtime changes cover then.

Why can't we just have the procedure look at the most recent block, see that it ends at 16:00, and ignore WAL samples after that?

>     3. Upload the blocks.
> 
>     4. Delete the wal.

>     5. Start Prometheus at 17.30. Compaction should create a block from the first sample of the wal to avoid overlapping with the uplloaded 16.00 - 17.20 block.

I assume in 4. we also deleted the blocks? Otherwise it would still apply that we just ignore samples before the most recent blocks, which TSDB does today I believe.

> I have run prombench with all these changes and all worked as expected.

It's a benchmark, not a comprehensive test suite AFAIK. Does it cover any kind of restart edge cases for example, which these changes ask to trigger.

---

So you primarily want to prevent that after a backup + wipe of all previous data, the ranges of blocks produced after a clean restart overlap. Did I get that right?

I'd try to avoid making blocks cover arbitrary ranges for that – not that it's a bad idea in general, but it's just not what TSDB was made for and the implications will reach further than you want to go for this feature.

Wouldn't it be sufficient to say that a block always gets the `minTime` in `meta.json` of its earliest sample, but keep blocks being terminated at the intended boundaries?
Maybe only do it if no previous block exists, so we keep our nicely aligned min and max times for the regular case.

krasi-georgiev:
big thanks for spending the time to review, appreciated!

to keep this short I will sum up.

at the moment tsdb always creates a block bound to the  the db.opts.BlockRanges[0] range. So if we start prometheus at 16.30 the next block would be 16.00-18.00 regardless of whether we already have a loaded block that covers 16.00-16.30 
https://github.com/prometheus/tsdb/blob/master/db.go#L818

> Wouldn't it be sufficient to say that a block always gets the minTime in meta.json of its earliest sample, but keep blocks being terminated at the intended boundaries?
Maybe only do it if no previous block exists, so we keep our nicely aligned min and max times for the regular case.

yes this will work, all the other changes are just for optimisation and will work even without these.

Although I didn't see any problems in the bench and manually doing different scenarios restart/stop , delete wal etc, I understand that this might cause issues elsewhere.

Do you remember the main reason to NOT create blocks bound to  head mint or the last block mint?
Can you point me to some docs that describes the guarantee about the maximum delay of a sample?
At high load keeping so much samples in the past increase the memory quite a bit.

krasi-georgiev:
updated. These are the minimum changes required to make it possible. (ignore the debug print code)

krasi-georgiev:
btw this also fixes https://github.com/prometheus/tsdb/issues/446

krasi-georgiev:
opened a proper PR on tsdb so closing this one.
https://github.com/prometheus/tsdb/pull/461

PR #4656, opened 2018-09-25
Limit concurrent remote reads.

tomwilkie:

Also aimed at preventing OOMs

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -30,6 +30,7 @@ import (
 	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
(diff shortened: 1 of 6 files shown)


promql/engine.go
```diff
@@ -1700,21 +1700,21 @@ func shouldDropMetricName(op ItemType) bool {
...
 
-// A queryGate controls the maximum number of concurrently running and waiting queries.
-type queryGate struct {
+// A QueryGate controls the maximum number of concurrently running and waiting queries.
+type QueryGate struct {
 	ch chan struct{}
 }
 
-// newQueryGate returns a query gate that limits the number of queries
+// NewQueryGate returns a query gate that limits the number of queries
```
brian-brazil: We should probably move this over to pkg

brian-brazil:
:+1: 

You've a conflict.

PR #4663, opened 2018-09-26
EC2 Platform meta label

sgissi:
Set __meta_ec2_platform label with the instance platform string. Set to 'windows' on Windows servers and absent otherwise. This is the only way to distinguish Linux from Windows EC2 instances.

Signed-off-by: Silvio Gissi <silvio@gissilabs.com>


sgissi:
@fabxc @brian-brazil Can one of you review if you have time? Pretty small change.

sgissi:
@brian-brazil PTAL. Thanks for the review.

brian-brazil:
Thanks!

Issue #4522, opened 2018-08-21
any way to support external user defined functions


alburthoffman:
## Proposal
**Use case. Why is this important?**
We has to support some validation functions, which are limited to our company only. 
like the metrics will be ignored if the host is in maintenance mode, and will be used when host is back.
We also have other private requirements.

it's not a good idea to add those functions to functions.go. communicate does-not need them and we will have to maintain our own prometheus version.

the ideal solution would be we can do something like:
`apply_external("plugin.function", metrics)`
then it can check functions defined in plugin.so, which can be preload or loaded dynamically.

brian-brazil:
Thank you for your suggestion. I don't think it's wise to allow any external code inside the Prometheus process, as that'll lead to bug reports caused by code in a non-standard setup which we have little to know ability to debug.

The use case you describe is already possible within PromQL.

brian-brazil:
> you mean ignoring metrics from maintenance mode hosts?

Yes, expose a metric indicating it's in maintenance. Then whatever in PromQL to use that.

> it's very limited to only support functions

That doesn't make a difference, it's still arbitrary code outside our control.

> it would be great to have plugins, then plugins can be maintained by users themself. they can also share them out if wanted.

Go plugins don't help for the above reasons. We already have a number of integration points.

RalphSu:
Hmm, probably related, i had some function want it available to my user in PromQL, since i'm exposing the PromQL to them. 
It's something like do calculation similiar https://en.wikipedia.org/wiki/Bollinger_Bands which sound like https://prometheus.io/docs/prometheus/latest/querying/functions/#holt_winters(), expecting syntax like bolling_band() that user can use.

Currently, it looks i had to fork the code, do the code change, compile and deploy, which looks invasive one. @brian-brazil  any suggestion on this kind of extension?

brian-brazil:
That's already possible within PromQL.

RalphSu:
> That's already possible within PromQL.
Do you mean extension is already possible? That would be cool, can you point me a bit of reference? Not find in site docs if i don't miss anything

pzou1974:
This is a very useful feature.   
Influxdb supports custom-functions.
https://docs.influxdata.com/flux/v0.24/functions/custom-functions/

inossidabile:
How about allowing people write things in LUA? This way it's properly isolated and can be controlled well. The language is promoted by Redis so we are in good company.

[...]

PR #4664, opened 2018-09-26
Allow PromQL to register custom funcs

vjsamuel:
This PR attempts to achieve what is stated in https://github.com/prometheus/prometheus/issues/4522

As seen in the test case, this would allow us to register a custom function on PromQL and achieve use-cases that are not generic enough to be committed back into Prometheus. 

Kindly take a look and provide feedback.

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -623,7 +623,7 @@ type EvalNodeHelper struct {
 	out Vector
 
 	// Caches.
-	// dropMetricName and label_*.
+	// DropMetricName and label_*.
 	dmn map[uint64]labels.Labels
 	// signatureFunc.
 	sigf map[uint64]uint64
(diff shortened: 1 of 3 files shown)


brian-brazil:
Thanks for your PR, Per #4522, this is not something we intend to support on maintenance/support grounds.

vjsamuel:
@brian-brazil, I understand the concern but as I understand and please correct me if I am wrong, the community would support only folks relying on the stock code right?

m2s:
@brian-brazil, @fabxc As I understand, the recommendation is to keep this out as this feature doesn’t benefit core prometheus.  While that makes sense for keeping low maintenance, it would hurts adoption due to inflexibility.  

If you look at other successful communities, we find ability to expand via plugin like model (e.g., Kubernetes) and registration mechanism.  When using promql as library, there would be use cases to extend it to support additional functions and extend it for supporting events and other functions?

RalphSu:
+1 for @m2s 
This PR appeals for me. That would be helpful for extension from developer point of view who use prometheus as library.

Having new function like BB runs as preprocessing before prometheus is leaky, it's similar to holt_winters.

@brian-brazil merged discussion in #4522 to here, let me know if it's already supported.

> That's already possible within PromQL.

Do you mean extension is already possible? That would be cool, can you point me a bit of reference? Not find in site docs if i don't miss anything

shaoxt:
I like this proposal.  Why don't we make prometheus extensible for this useful feature?

brian-brazil:
Obseleted by #6434

PR #4667, opened 2018-09-27
discovery: add metrics + send updates from one goroutine only

simonpasquier:
The added metrics are:

* prometheus_sd_discovered_targets
* prometheus_sd_received_updates_total
* prometheus_sd_updates_delayed_total
* prometheus_sd_updates_total

cc @krasi-georgiev

### DIFF
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -118,7 +118,7 @@ func NewManager(ctx context.Context, logger log.Logger) *Manager {
 		discoverCancel: []context.CancelFunc{},
 		ctx:            ctx,
 		updatert:       5 * time.Second,
-		trigger:        make(chan struct{}, 1),
+		triggerSend:    make(chan struct{}, 1),
 	}
 }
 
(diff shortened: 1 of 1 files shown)


discovery/manager.go
```diff
@@ -110,24 +135,28 @@ type Manager struct {
...
-	// The sync channels sends the updates in map[targetSetName] where targetSetName is the job value from the scrape config.
+	// The sync channel sends the updates as a map where the key is the job value from the scrape config.
 	syncCh chan map[string][]*targetgroup.Group
 
 	// How long to wait before sending updates to the channel. The variable
 	// should only be modified in unit tests.
 	updatert time.Duration
+
+	// The trigger channel signals to the manager that new updates have been received from providers.
+	trigger chan struct{}
```
krasi-georgiev: how about `triggerSend`

discovery/manager.go
```diff
@@ -171,43 +200,48 @@ func (m *Manager) startProvider(ctx context.Context, p *provider) {
...
 			select {
-			case triggerUpdate <- struct{}{}:
+			case m.trigger <- struct{}{}:
 			default:
 			}
+		}
+	}
+}
+
+func (m *Manager) sendUpdates() {
```
krasi-georgiev: how about just `sender()` ?

simonpasquier:
@krasi-georgiev done!

simonpasquier:
The benchmark doesn't show significant differences on resource usage before and after this PR but if master had the `prometheus_sd_updates_total` metric, we would probably have seen a decrease of the metric's value with the PR.

krasi-georgiev:
LGTM

krasi-georgiev:
don't forget to squash on merge as otherwise we end up with non build-able commits.

the merge button has a drop-down so you can select it by default.

PR #4638, opened 2018-09-19
[FIX] EC2 DS: Check for existence of OwnerID

jannickfahlbusch:
Commit 1c89984778fa78eda1b89d7fc2d30e025c7d3b9a introduced the ability to expose the owner of the instance. However, this breaks Prometheus if there is no OwnerID in the reservation (Eg. if you are using a private EC2-API introduced by #4333)

Signed-off-by: Jannick Fahlbusch <git@jf-projects.de>

PR #4672, opened 2018-09-28
[FIX] EC2 DS: Check for existence of OwnerID

jannickfahlbusch:
Commit 1c89984 introduced the ability to expose the owner of the instance.
However, this breaks Prometheus if there is no OwnerID in the reservation (Eg. if you are using a private EC2-API introduced by #4333)

This is the reopened PR for #4638 (https://github.com/prometheus/prometheus/pull/4638#issuecomment-425393366)

Closes https://github.com/prometheus/prometheus/pull/4672

### DIFF
--- a/discovery/ec2/ec2.go
+++ b/discovery/ec2/ec2.go
@@ -241,9 +241,13 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 					continue
 				}
 				labels := model.LabelSet{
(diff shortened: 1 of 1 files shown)


gouthamve:
Hi, this still targets `master`..

jannickfahlbusch:
Ah, forgot to change the branch :(

Not sure why all those commits are here... I just branched off `origin/release-2.4`

PR #4677, opened 2018-09-29
Adding new metric type to track in-flight remote read queries.

### LINKED ISSUE #4661: Metric for number of remote_read queries in-flight
We have a metric that shows the number in-flight queries on the promql engine: `prometheus_engine_queries`, with the introduction of the concurrency limits #4656, a similar metric should be added for in-flight remote-read queries.

tariq1890:
Add new metric type to track number of remote read queries in the remote storage module of prometheus.

### DIFF
--- a/storage/remote/read.go
+++ b/storage/remote/read.go
@@ -16,14 +16,30 @@ package remote
 import (
 	"context"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/storage"
 )
 
+var remoteReadQueries = prometheus.NewGaugeVec(
+	prometheus.GaugeOpts{
+		Namespace: namespace,
+		Subsystem: subsystem,
+		Name:      "remote_read_queries",
+		Help:      "The number of in-flight remote read queries.",
+	},
+	[]string{"client"},
+)
+
+func init() {
+	prometheus.MustRegister(remoteReadQueries)
+}
+
 // QueryableClient returns a storage.Queryable which queries the given
 // Client to select series sets.
 func QueryableClient(c *Client) storage.Queryable {
+	remoteReadQueries.WithLabelValues(c.Name())
 	return storage.QueryableFunc(func(ctx context.Context, mint, maxt int64) (storage.Querier, error) {
 		return &querier{
 			ctx:    ctx,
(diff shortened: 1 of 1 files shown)


tariq1890:
@simonpasquier Thanks! Review comments addressed.

simonpasquier:
Thanks for the update! I'll let the last word to @gouthamve as he filed the PR originally.

storage/remote/read.go
```diff
@@ -16,15 +16,31 @@ package remote
...
+
+func init() {
+	prometheus.MustRegister(remoteReadQueries)
+}
+
 // QueryableClient returns a storage.Queryable which queries the given
 // Client to select series sets.
 func QueryableClient(c *Client) storage.Queryable {
 	return storage.QueryableFunc(func(ctx context.Context, mint, maxt int64) (storage.Querier, error) {
+		remoteReadQueries.WithLabelValues(c.Name())
```
gouthamve: We can move this up, i.e,

```go
// QueryableClient returns a storage.Queryable which queries the given
// Client to select series sets.
func QueryableClient(c *Client) storage.Queryable {
	remoteReadQueries.WithLabelValues(c.Name())

	return storage.QueryableFunc(func(ctx context.Context, mint, maxt int64) (storage.Querier, error) {
		return &querier{
			ctx:    ctx,
			mint:   mint,
			maxt:   maxt,
			client: c,
		}, nil
	})
}
```

gouthamve:
Hmm, this looks good to me and should be included, but the original issue is for the read endpoint exposed by prometheus. See: https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L763:17

Feel free to open an new PR for it, or updating this PR. Sorry for not being clear in the issue.

storage/remote/read.go
gouthamve: The reason for it is until we use the `Queryable` we won't be initialising the metric.

tariq1890: oh yes! Thanks for pointing that out.

tariq1890:
@gouthamve Thanks for the review. I have made the review comments and updated the PR description since it does not fix the issue you filed. I'll ask you for clarifications in the issue page.

tariq1890:
@gouthamve Can you have a look at this? I have addressed your comments

tariq1890:
@gouthamve Ping!

gouthamve:
👍 Thanks!

PR #4685, opened 2018-10-01
circleci: store built binary as an artifact

simonpasquier:
It might be convenient to store the resulting binary of a PR as an artifact. This way, people that want to test/try a PR can just download the binary from CircleCI instead of checking out and rebuilding locally.

Another use case is [prombench](https://github.com/prometheus/prombench/). For instance, we can't really evaluate the benefits of using go1.11 (cf. https://github.com/prometheus/prometheus/pull/4626) as prombench rebuilds the binary from sources and it currently uses go1.10.2. Instead it could use the artifact of the PR to build the container image.

Finally this is already related to https://github.com/prometheus/prometheus/pull/4453 which aims at building Docker images for every PR but this PR has a much smaller scope.

cc @SuperQ @krasi-georgiev


SuperQ:
What about storing the output of the `build` step as well, that way we get all architecture artifacts.

simonpasquier:
I just wanted to test the water but yes, it makes sense too.

SuperQ:
Also, maybe `promtool`?

SuperQ:
Yup, this is a great idea. We should do this everywhere.

SuperQ:
Here's an [artifact output](https://circleci.com/gh/prometheus/prometheus/8538#artifacts/containers/0). Looks good!

PR #4687, opened 2018-10-01
Add documentation for `quantile()`

RichiH:
Somehow, we missed that...

Signed-off-by: Richard Hartmann <richih@richih.org>


brian-brazil:
Quantile is an aggregator, not a function, and it is documented.

Issue #4684, opened 2018-10-01
EC2 service discovery panics when OwnerID is nil


simonpasquier:
_This only affects installations specifying the `endpoint` parameter in the EC2 SD configuration which means it doesn't concern users on AWS EC2._

https://github.com/prometheus/prometheus/blob/1c89984778fa78eda1b89d7fc2d30e025c7d3b9a/discovery/ec2/ec2.go#L243-L246

When `r.OwnerId` is `nil`, the server will crash.

simonpasquier:
Associated PR https://github.com/prometheus/prometheus/pull/4672

gouthamve:
Closed in #4672

PR #4693, opened 2018-10-04
Adding private_dns_name to the list of ec2 labels which can be used i…

sanicheev:
…n node naming for dynamic environments

### DIFF
--- a/discovery/ec2/ec2.go
+++ b/discovery/ec2/ec2.go
@@ -45,6 +45,7 @@ const (
 	ec2LabelOwnerID         = ec2Label + "owner_id"
 	ec2LabelPublicDNS       = ec2Label + "public_dns_name"
 	ec2LabelPublicIP        = ec2Label + "public_ip"
+	ec2LabelPrivateDNS      = ec2Label + "private_dns_name"
 	ec2LabelPrivateIP       = ec2Label + "private_ip"
 	ec2LabelPrimarySubnetID = ec2Label + "primary_subnet_id"
 	ec2LabelSubnetID        = ec2Label + "subnet_id"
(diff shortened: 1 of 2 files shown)


sanicheev:
@codesome can you please re-run ci/circleci: build?

codesome:
@sanicheev Sorry, I don't have access to it. @krasi-georgiev Can you please restart the build?

discovery/ec2/ec2.go
```diff
@@ -245,6 +246,7 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 					ec2LabelInstanceID: model.LabelValue(*inst.InstanceId),
 				}
 				labels[ec2LabelPrivateIP] = model.LabelValue(*inst.PrivateIpAddress)
+				labels[ec2LabelPrivateDNS] = model.LabelValue(*inst.PrivateDnsName)
```
simonpasquier: couldn't `inst.PrivateDnsName` be `nil`?

sanicheev: i'm not sure, i think if private ip address is set it should not be nil. i need to check this

sanicheev: According to aws-sdk comments same conditions applied as for PublicDnsName.  But in case of PrivateDnsName private ip address is always assigned.

simonpasquier: We had a recent issue (https://github.com/prometheus/prometheus/issues/4684) where the EC2 service discovery crashed when used with EC2-like APIs so I'd prefer to be safe and add a check.

sanicheev: Agree. I've added a check.

simonpasquier:
@brian-brazil PTAL

brian-brazil:
Thanks!

PR #4696, opened 2018-10-04
promql: fill_absent function support

sumkincpp:
## Feature request ##
Ability to fill absent(empty) values with desired value, e.g. querying with
```
fill_absent(up)
```
will fill absent values with zeroes(by default).

Or explicitly with custom value:
```
fill_absent(up, 100)
```

This similar to what `clamp_max`, `clamp_min` and `absent` do.

Also, labels for empty values should be retained.

### DIFF
--- a/docs/querying/functions.md
+++ b/docs/querying/functions.md
@@ -109,6 +109,11 @@ Special cases are:
 * `Exp(+Inf) = +Inf`
 * `Exp(NaN) = NaN`
 
+## `fill_absent()`
+
+`fill_absent(v instant-vector, val=0 scalar)` fills all absent values in `v`
+to have all value of `val`.
+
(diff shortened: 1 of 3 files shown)


brian-brazil:
Can you explain more about what you're trying to do?

sumkincpp:
@brian-brazil Implementing a function that will fill all absent(empty) values in metric with a desired value.

It's currently not implemented as separate function. Using ```up + absent(up)`` tricks will drop all labels.

sumkincpp:
@brian-brazil any idea why tests are failing with non-existent metric? Should it be somewhere defined?

roidelapluie:
As @brian-brazil said, this PR can be closed in favour of `up or on() vector(0)`.

PR #4699, opened 2018-10-05
Adding new metric type to track in-flight queries via the remote read API endpoint.

### LINKED ISSUE #4661: Metric for number of remote_read queries in-flight

tariq1890:
Fixes #4661 @gouthamve


web/api/v1/api.go
```diff
@@ -28,9 +28,11 @@ import (
 	"time"
 	"unsafe"
 
+	"github.com/prometheus/client_golang/prometheus"
+
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
-	jsoniter "github.com/json-iterator/go"
+	"github.com/json-iterator/go"
```
simonpasquier: This change isn't needed, I suspect that the alias is there to make obvious that `github.com/json-iterator/go` == `jsoniter`

tariq1890:
@simonpasquier Thanks for reviewing. I have addressed your comments :)

web/api/v1/api.go
```diff
@@ -28,6 +28,8 @@ import (
 	"time"
 	"unsafe"
 
+	"github.com/prometheus/client_golang/prometheus"
```
simonpasquier: The package import should go with the other 3rd party packages just below.

web/api/v1/api.go
```diff
@@ -781,10 +795,12 @@ func (api *API) remoteRead(w http.ResponseWriter, r *http.Request) {
 		}
 
 		querier, err := api.Queryable.Querier(r.Context(), from, through)
+		remoteReadQueries.Inc()
```
simonpasquier: This should be done at the beginning of the function. By the way, the metric's update could be delegated to `api.remoteReadGate` as we want to increase the gauge at the beginning of `api.remoteReadGate.Start()` and decrease it when `api.remoteReadGate.End()` returns.

@gouthamve WDYT?

cstyan:
I think including the remote URL as a label in the metrics would be useful. This could be done in a similar way to how we include the queue, which contains some info about that queues remote endpoint: https://github.com/prometheus/prometheus/blob/master/storage/remote/queue_manager.go#L189

web/api/v1/api.go
cstyan: this makes sense to me

tariq1890:
Done. I've also implemented Gauge vector as per @cstyan's suggestions.

web/api/v1/api.go
```diff
@@ -762,7 +778,10 @@ func (api *API) serveFlags(r *http.Request) (interface{}, *apiError, func()) {
 
 func (api *API) remoteRead(w http.ResponseWriter, r *http.Request) {
 	api.remoteReadGate.Start(r.Context())
+	remoteReadQueries.WithLabelValues(r.RequestURI).Inc()
```
gouthamve: What does a typical `r.RequestURI` look like? Isn't it going to be same every time?

If it's going to have the params or selectors too, then we'll have the infinite-cardinality issue.

web/api/v1/api.go
```diff
@@ -78,6 +85,15 @@ var corsHeaders = map[string]string{
...
 }
 
+var remoteReadQueries = prometheus.NewGaugeVec(
+	prometheus.GaugeOpts{
+		Namespace: namespace,
+		Subsystem: subsystem,
+		Name:      "remote_read_queries",
+		Help:      "The current number of remote read queries being executed or waiting.",
+	},
+	[]string{urlLabel})
```
gouthamve: No need to make a new variable. Just out `[]string{"url"}` here.

tariq1890: alright

web/api/v1/api.go
tariq1890: Yes. you are definitely correct. What would you recommend as a label ? In my previous commit I just added this metric without a label. Let me know if we are better off not having a label for this

tariq1890: Done @gouthamve

web/api/v1/api.go
```diff
@@ -1085,6 +1101,7 @@ func parseDuration(s string) (time.Duration, error) {
 
 func init() {
```
gouthamve: Can you move the init to the top?

gouthamve:
Hi @tariq1890, thanks for keeping at it, could you address @simonpasquier's comment about the `jsoniter` import and this is good to go!

tariq1890:
Ah that keeps getting cleared lol due to auto-formatting by my IDE.

web/api/v1/api.go
tariq1890: Sure

tariq1890:
@gouthamve Done :)

gouthamve:
Thanks!

PR #4701, opened 2018-10-05
Add ability to filter triton_sd targets by pre-defined groups

richardkiene:
Signed-off-by: Richard Kiene <richard.kiene@joyent.com>

As Triton and Manta installs have become substantially larger and customer reliance upon Prometheus has become more important, it has become painfully obvious that filtering is a necessary feature of Triton Service Discovery. This pull request leverages the newly added group filtering support that we've added to [Triton CMON](https://github.com/joyent/triton-cmon).

//cc: @fabxc @brian-brazil

### DIFF
--- a/discovery/triton/triton.go
+++ b/discovery/triton/triton.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/url"
+	"strings"
 	"time"
 
 	"github.com/go-kit/kit/log"
@@ -33,6 +35,7 @@ import (
 
 const (
 	tritonLabel             = model.MetaLabelPrefix + "triton_"
+	tritonLabelGroups       = tritonLabel + "groups"
 	tritonLabelMachineID    = tritonLabel + "machine_id"
 	tritonLabelMachineAlias = tritonLabel + "machine_alias"
 	tritonLabelMachineBrand = tritonLabel + "machine_brand"
@@ -65,6 +68,7 @@ type SDConfig struct {
 	Account         string                `yaml:"account"`
 	DNSSuffix       string                `yaml:"dns_suffix"`
 	Endpoint        string                `yaml:"endpoint"`
+	Groups          []string              `yaml:"groups,omitempty"`
 	Port            int                   `yaml:"port"`
 	RefreshInterval model.Duration        `yaml:"refresh_interval,omitempty"`
 	TLSConfig       config_util.TLSConfig `yaml:"tls_config,omitempty"`
@@ -102,11 +106,12 @@ func init() {
 // DiscoveryResponse models a JSON response from the Triton discovery.
 type DiscoveryResponse struct {
 	Containers []struct {
-		ServerUUID  string `json:"server_uuid"`
-		VMAlias     string `json:"vm_alias"`
-		VMBrand     string `json:"vm_brand"`
-		VMImageUUID string `json:"vm_image_uuid"`
-		VMUUID      string `json:"vm_uuid"`
+		Groups      []string `json:"groups"`
+		ServerUUID  string   `json:"server_uuid"`
+		VMAlias     string   `json:"vm_alias"`
+		VMBrand     string   `json:"vm_brand"`
+		VMImageUUID string   `json:"vm_image_uuid"`
+		VMUUID      string   `json:"vm_uuid"`
 	} `json:"containers"`
 }
 
@@ -187,6 +192,11 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 	}()
 
 	var endpoint = fmt.Sprintf("https://%s:%d/v%d/discover", d.sdConfig.Endpoint, d.sdConfig.Port, d.sdConfig.Version)
+	if len(d.sdConfig.Groups) > 0 {
+		groups := url.QueryEscape(strings.Join(d.sdConfig.Groups, ","))
+		endpoint = fmt.Sprintf("%s?groups=%s", endpoint, groups)
+	}
+
 	tg = &targetgroup.Group{
 		Source: endpoint,
 	}
@@ -219,6 +229,12 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 		}
 		addr := fmt.Sprintf("%s.%s:%d", container.VMUUID, d.sdConfig.DNSSuffix, d.sdConfig.Port)
 		labels[model.AddressLabel] = model.LabelValue(addr)
+
+		if len(container.Groups) > 0 {
+			name := "," + strings.Join(container.Groups, ",") + ","
+			labels[tritonLabelGroups] = model.LabelValue(name)
+		}
+
 		tg.Targets = append(tg.Targets, labels)
 	}
 
--- a/discovery/triton/triton_test.go
+++ b/discovery/triton/triton_test.go
@@ -55,6 +55,16 @@ var (
 			CertFile:           "shouldnotexist.cert",
 		},
 	}
+	groupsconf = SDConfig{
+		Account:         "testAccount",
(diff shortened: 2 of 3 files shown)


brian-brazil:
I don't see the groups among the current metadata, usually we use relabelling to do filtering rather than having every SD implement it's own ad-hoc mechanisms. Could you add this information to the metadata returned and work from that?

brian-brazil:
> that is already possible with other metadata, but at this point we're throwing away 10's of thousands of targets that come back from the endpoint with that method.

I don't see the group among the current data. Is this actually causing performance problems, or is it more theoretical?

richardkiene:
@brian-brazil Yes it is actually causing performance problems. Groups were very recently added to support filtering in this exact way, so I'm not sure why they would also be in the metadata.

brian-brazil:
Can you add in the metadata in this change then also please?

I'd also suggest waiting until the support in triton-cmon is actually merged.

richardkiene:
@brian-brazil I've updated the PR to include support for groups in metadata. The corresponding changes for triton-cmon will be merged to master sometime during business hours on Monday Oct 8th 2018 Pacific Time. If you'd like to track that process you can watch https://cr.joyent.us/#/c/4888/. That said, I've tested these changes against an older version of triton-cmon and they should not break users who are using an older versions.

discovery/triton/triton.go
```diff
@@ -102,11 +105,12 @@ func init() {
 // DiscoveryResponse models a JSON response from the Triton discovery.
 type DiscoveryResponse struct {
 	Containers []struct {
-		ServerUUID  string `json:"server_uuid"`
-		VMAlias     string `json:"vm_alias"`
-		VMBrand     string `json:"vm_brand"`
-		VMImageUUID string `json:"vm_image_uuid"`
-		VMUUID      string `json:"vm_uuid"`
+		Groups      []string `json:"groups:`
```
brian-brazil: The tag is off here

discovery/triton/triton.go
```diff
@@ -187,6 +191,10 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 	}()
 
 	var endpoint = fmt.Sprintf("https://%s:%d/v%d/discover", d.sdConfig.Endpoint, d.sdConfig.Port, d.sdConfig.Version)
+	if d.sdConfig.Groups != nil {
+		endpoint = fmt.Sprintf("%s?groups=%s", endpoint, strings.Join(d.sdConfig.Groups, ","))
```
brian-brazil: Is there any escaping needed here?

discovery/triton/triton.go
```diff
@@ -219,6 +227,13 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 		}
 		addr := fmt.Sprintf("%s.%s:%d", container.VMUUID, d.sdConfig.DNSSuffix, d.sdConfig.Port)
 		labels[model.AddressLabel] = model.LabelValue(addr)
+
+		if container.Groups != nil {
+			for _, v := range container.Groups {
+				labels[tritonLabelGroups+"_"+model.LabelName(v)] = model.LabelValue(v)
```
brian-brazil: The way we handle arrays is to have a single label where they're joined with commas, and also surrounded with commas.

discovery/triton/triton.go
richardkiene: I assume you're talking about the `:` at the end of `json:"groups:` and the missing `"`? If so it's amazing this actually works with it like that :). I'll fix this right now.

richardkiene: Fixed

discovery/triton/triton.go
richardkiene: I added escaping anyway, so this should be fixed.

discovery/triton/triton.go
richardkiene: done

richardkiene:
@brian-brazil I think I've fixed everything you pointed out, thanks.

discovery/triton/triton.go
```diff
@@ -187,6 +192,11 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 	}()
 
 	var endpoint = fmt.Sprintf("https://%s:%d/v%d/discover", d.sdConfig.Endpoint, d.sdConfig.Port, d.sdConfig.Version)
+	if d.sdConfig.Groups != nil {
+		groups := html.EscapeString(strings.Join(d.sdConfig.Groups, ","))
```
brian-brazil: url.QueryEscape

discovery/triton/triton.go
```diff
@@ -187,6 +192,11 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 	}()
 
 	var endpoint = fmt.Sprintf("https://%s:%d/v%d/discover", d.sdConfig.Endpoint, d.sdConfig.Port, d.sdConfig.Version)
+	if d.sdConfig.Groups != nil {
```
brian-brazil: `len(d.sdConfig.Groups) > 0` would be more typical

discovery/triton/triton_test.go
```diff
@@ -112,6 +136,7 @@ func TestTritonSDRefreshMultipleTargets(t *testing.T) {
 	var (
 		dstr = `{"containers":[
 		 	{
+                                "groups":["foo","bar","baz"],
```
brian-brazil: indentation is off

discovery/triton/triton.go
```diff
@@ -219,6 +229,12 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 		}
 		addr := fmt.Sprintf("%s.%s:%d", container.VMUUID, d.sdConfig.DNSSuffix, d.sdConfig.Port)
 		labels[model.AddressLabel] = model.LabelValue(addr)
+
+		if container.Groups != nil && len(container.Groups) > 0 {
+			name := strings.Join(container.Groups, ",")
+			labels[tritonLabelGroups+"_"+model.LabelName(","+name+",")] = model.LabelValue(name)
```
brian-brazil: The name should be fixed, look at how Consul tags are handled

discovery/triton/triton_test.go
richardkiene: This appears to "look" incorrect because of the way GitHub formats it. I double checked the source code, re-ran `go fmt`, and also ran `make style` without issue.

discovery/triton/triton.go
richardkiene: Will do

discovery/triton/triton.go
richardkiene: Will do

discovery/triton/triton.go
richardkiene: Okay, I understand now. Fixing this currently.

docs/configuration/configuration.md
```diff
@@ -947,6 +949,14 @@ dns_suffix: <string>
 # often the same value as dns_suffix.
 endpoint: <string>
 
+# The group(s) to filter discovery results by.
```
brian-brazil: This is a bit wordy, Consul has "A list of services for which targets are retrieved. If omitted, all services
 are scraped."

richardkiene: Done.

richardkiene:
@brian-brazil I've updated the commit message to reflect the additional changes. I also addressed your concern about the documentation being to wordy regarding the groups configuration option. I will ping you again once CMON support for this change has landed in master.

Thanks for the review!

richardkiene:
@brian-brazil the corresponding changes to triton-cmon have been [merged](https://github.com/joyent/triton-cmon/commit/f837f3e96bdc0e47ba2c3ad063b244a8092b06c0) into master. Thanks!

brian-brazil:
Thanks!

PR #4712, opened 2018-10-09
Merge release-2.4

simonpasquier:
cc @brian-brazil @gouthamve


simonpasquier:
(sigh) DCO is again confused by slightly different addresses for @gouthamve and @jannickfahlbusch.
@brian-brazil should we just force it to pass?

brian-brazil:
Yeah, these are false positives.

simonpasquier:
Because I'm used to submit forked PRs so I didn't even think about using the GH interface directly... Replaced by #4730

PR #4719, opened 2018-10-10
[FIX] Azure DS: Return error when request failed

jannickfahlbusch:
This fixes the issue that the error is swallowed when the request failed.

### DIFF
--- a/discovery/azure/azure.go
+++ b/discovery/azure/azure.go
@@ -551,13 +551,12 @@ func (client *azureClient) getNetworkInterfaceByID(networkInterfaceID string) (*
 
 	resp, err := client.nic.GetSender(req)
 	if err != nil {
-		result.Response = autorest.Response{Response: resp}
 		return nil, autorest.NewErrorWithError(err, "network.InterfacesClient", "Get", resp, "Failure sending request")
 	}
 
 	result, err = client.nic.GetResponder(resp)
 	if err != nil {
-		err = autorest.NewErrorWithError(err, "network.InterfacesClient", "Get", resp, "Failure responding to request")
+		return nil, autorest.NewErrorWithError(err, "network.InterfacesClient", "Get", resp, "Failure responding to request")
 	}
 
 	return &result, nil


discovery/azure/azure.go
```diff
@@ -502,5 +502,5 @@ func (client *azureClient) getNetworkInterfaceByID(networkInterfaceID string) (n
 		err = autorest.NewErrorWithError(err, "network.InterfacesClient", "Get", resp, "Failure responding to request")
 	}
 
-	return result, nil
+	return result, err
```
simonpasquier: Thanks for spotting it! It works but I'd prefer to keep the usual flow like this:

```go
  result, err = client.nic.GetResponder(resp)
  if err != nil {
    return result, autorest.NewErrorWithError(err, "network.InterfacesClient", "Get", resp, "Failure responding to request")
  }

  return result, nil
```

Also L496 is useless and can be removed while you are it.

jannickfahlbusch: Fixed, thanks for the hint!

simonpasquier:
@brian-brazil PTAL

discovery/azure/azure.go
```diff
@@ -493,13 +493,12 @@ func (client *azureClient) getNetworkInterfaceByID(networkInterfaceID string) (n
...
 	resp, err := client.nic.GetSender(req)
 	if err != nil {
-		result.Response = autorest.Response{Response: resp}
 		return result, autorest.NewErrorWithError(err, "network.InterfacesClient", "Get", resp, "Failure sending request")
 	}
 
 	result, err = client.nic.GetResponder(resp)
 	if err != nil {
-		err = autorest.NewErrorWithError(err, "network.InterfacesClient", "Get", resp, "Failure responding to request")
+		return result, autorest.NewErrorWithError(err, "network.InterfacesClient", "Get", resp, "Failure responding to request")
```
brian-brazil: It's usual to return nil for the result if there's an error.

jannickfahlbusch: I just wanted to be consistent with the style of the rest of the file.

What do you mean? Should I fix it in just this place or in the entire file?

simonpasquier: Looking at the `github.com/prometheus/prometheus/discovery/azure` package, there are quite a few places where we could do `return nil, err`. I'd lean toward "fixing" these parts in a separate PR.

jannickfahlbusch: So whats the status then?

Should this be merged or are we waiting on the follow-up change?

tariq1890:
@simonpasquier @brian-brazil Requesting your attention on this :)

simonpasquier:
I'm fine with this change especially if @tariq1890 fixes the `return result, ...` => `return nil, ...` in #5015. Note that v2.7.0-rc.0 being released, we're supposedly having a master branch freeze. cc @gouthamve

tariq1890:
@simonpasquier You have a deal ;)

jannickfahlbusch:
Sounds like we can finally merge this PR in the near future 🎉 😄

tariq1890:
@jannickfahlbusch Can you rebase this?

@brian-brazil FYI

discovery/azure/azure.go
tariq1890:
```suggestion
		return nil, autorest.NewErrorWithError(err, "network.InterfacesClient", "Get", resp, "Failure responding to request")
```

jannickfahlbusch: Done

jannickfahlbusch:
Rebased and ready to be merged

brian-brazil:
Thanks!

PR #4722, opened 2018-10-10
Updates to /targets and /rules (scrape duration, last evaluation time)

### LINKED ISSUE #4706: Show last execution time of rules
## Proposal

In the page of '/rules', there is only `EvaluationDuration` of groups and rules. `LastExecutionTime` is required to show the last execution time a group. Then we can check if rules are executing normally. This field is like `LastScrapeTime`.

Also `/targets` could have a filed to show the scrape duration of a target. Then we can see if a target have too many metrics or the network is not stable.

wbh1:
This is to resolve #4706. 

Some notes:
- Instead of printing the whole time stamp with milliseconds and all for each group in /rules, I chose to just print the basic timestamp with time zone (fine with changing it if that's preferred, it just doesn't look nice in my opinion). The individual rules still have a full timestamp.

- In manager.go, there was already an `evalTimestamp` variable but it did not reflect `time.Now()` so it was not used. However, the property added to `Group` struct is `evaluationTimestamp` to keep with the naming scheme of the other property, `evaluationDuration`. Since the name is similar to the existing variable, it's possible that some confusion could be caused.


wbh1:
cc @juliusv @simonpasquier

juliusv:
Thanks, taking a look!

rules/alerting.go
```diff
@@ -120,6 +120,8 @@ type AlertingRule struct {
 	annotations labels.Labels
 	// Time in seconds taken to evaluate rule.
 	evaluationDuration time.Duration
+	// Unix timestamp of last evaluation of rule
```
juliusv: Style nit: add a period at end of comment (we do sentence-style comments).

rules/alerting.go
```diff
@@ -120,6 +120,8 @@ type AlertingRule struct {
...
+	evaluationTimestamp time.Time
```
juliusv: The mutex has a comment `// Protects the below.`, but now there's already several fields above it in the struct that it also protects. Doesn't have to be done in this PR, but it would be good to reorganize this all to make it clear which fields the mutex protects.

rules/manager.go
```diff
@@ -175,6 +177,7 @@ type Group struct {
 	opts                 *ManagerOptions
 	evaluationDuration   time.Duration
 	mtx                  sync.Mutex
+	evaluationTimestamp  time.Time
```
juliusv: This field should be grouped with `evaluationDuration`, probably under `mtx` (similar mutex clarity problem as above).

rules/manager.go
```diff
@@ -226,9 +229,9 @@ func (g *Group) run(ctx context.Context) {
 	iter := func() {
 		iterationsScheduled.Inc()
 
-		start := time.Now()
+		g.SetEvaluationTimestamp(time.Now())
 		g.Eval(ctx, evalTimestamp)
-		timeSinceStart := time.Since(start)
+		timeSinceStart := time.Since(g.evaluationTimestamp)
```
juliusv: While I don't see that a single group can be executed twice concurrently, you're using the mutex-protected function to set `g.evaluationTimestamp`, but then accessing it directly here without mutex to read it back.

IMO I would just keep the `start` variable as it is and additionally add the `g.SetEvaluationTimestamp(start)` call.

One thing to think about: do you really want to already set the evaluation timestamp *before* the rule evaluation happens or only after it completes? We only set the eval duration afterwards (naturally), and probably it'd be less confusing to also set the timestamp only afterwards?

rules/alerting.go
wbh1: Missed this one. I'll update :)

rules/manager.go
wbh1: I'll add the start variable back. The idea with adding the timestamp prior to to completing the evaluation was to represent the last time that rules were attempted to be evaluated (at the start of evaluation). I think you're right that it would be more consistent to set the timestamp afterwards, though.

I think other monitoring tools (e.g. Nagios) also use the check (i.e. rule evaluation) completion time rather than start time to signify the last evaluation, so it'd be more consistent for people coming from other products, too.

rules/manager.go
```diff
@@ -381,6 +398,7 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
 
 			evalTotal.Inc()
 
+			rule.SetEvaluationTimestamp(time.Now())
```
juliusv: Same as for groups, should the rule eval timestamp only be set after a successful evaluation, together with the duration (in the `defer` above?).

juliusv: (or not necessarily *successful* duration)

rules/recording.go
```diff
@@ -38,6 +38,8 @@ type RecordingRule struct {
 	mtx sync.Mutex
 	// The health of the recording rule.
 	health RuleHealth
+	// Unix timestamp of last evaluation of rule
```
juliusv: Style nit: Dito, period at the end.

`time.Time` is not a Unix timestamp, so I'd drop the `Unix`.

rules/recording.go
```diff
@@ -156,6 +158,20 @@ func (rule *RecordingRule) GetEvaluationDuration() time.Duration {
 	return rule.evaluationDuration
 }
 
+// SetEvaluationTimestamp updates evaluationTimestamp to the Unix timestamp of when the rule was last evaluated.
```
juliusv: Same comment here about `Unix timestamp` and elsewhere.

web/ui/templates/targets.html
```diff
@@ -32,6 +32,7 @@ <h2 class="job_header{{if lt $healthy $total}} danger{{end}}">
             <th>State</th>
             <th>Labels</th>
             <th>Last Scrape</th>
+            <th>Scape Duration</th>
```
juliusv: Scape -> Scrape

rules/manager.go
```diff
@@ -311,6 +314,20 @@ func (g *Group) SetEvaluationDuration(dur time.Duration) {
...
+	g.mtx.Lock()
+	defer g.mtx.Unlock()
+	g.evaluationTimestamp = ts
+}
+
+// GetEvaluationTimestamp returns the time the evaluation took place.
+func (g *Group) GetEvaluationTimestamp() string {
+	g.mtx.Lock()
+	defer g.mtx.Unlock()
+	return g.evaluationTimestamp.UTC().Format("15:04:06 MST")
```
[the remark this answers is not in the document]
brian-brazil: We should also not expose anything other than UTC.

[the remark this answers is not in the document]
juliusv: It's UTC here, the MST is just there because that's what the Go time formatter expects as a reference time zone.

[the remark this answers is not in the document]
wbh1: I've fixed everything else based on your feedback, but left this part the for the time being while a decision is made.

web/ui/templates/rules.html
```diff
@@ -34,6 +36,7 @@ <h2>Rules</h2>
               {{end}}
             </td>
             <td>{{humanizeDuration .GetEvaluationDuration.Seconds}}</td>
+            <td>{{.GetEvaluationTimestamp.UTC}}</h2></td>
```
kdada: There is a reference for this cell:
```html
            <td class="last-scrape">
              {{if .LastScrape.IsZero}}Never{{else}}{{since .LastScrape}} ago{{end}}	              {{if .LastScrape.IsZero}}Never{{else}}{{since .LastScrape}} ago{{end}}
            </td>	            </td>
```

And `Timestamp` should prior to `Duration`.

rules/manager.go
```diff
@@ -232,6 +235,7 @@ func (g *Group) run(ctx context.Context) {
 
 		iterationDuration.Observe(timeSinceStart.Seconds())
 		g.SetEvaluationDuration(timeSinceStart)
+		g.SetEvaluationTimestamp(time.Now())
```
kdada: This line should be `g.SetEvaluationTimestamp(start)`.

rules/manager.go
```diff
@@ -377,6 +395,7 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
 				sp.Finish()
 				evalDuration.Observe(time.Since(t).Seconds())
 				rule.SetEvaluationDuration(time.Since(t))
+				rule.SetEvaluationTimestamp(time.Now())
```
kdada: This line should be `rule.SetEvaluationTimestamp(t)`.

rules/manager.go
wbh1: Hey Wei,

That is similar to how I originally had it, but after discussing with Julius we agreed that it makes more sense and is more consistent to set the evaluation timestamp to the time that evaluation completed rather than when it started.

rules/manager.go
wbh1: Same comment as above.

rules/manager.go
juliusv: There may have been a misunderstanding here: I meant we should still set it to the start / eval timestamp, but only after the evaluation has finished.

juliusv: Basically, this will tell people the timestamp for which they can expect rule result samples to be in the TSDB already, which I think is the most useful timestamp to expose for users.

rules/manager.go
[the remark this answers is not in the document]
simonpasquier: What if we display the time since the last evaluation as it is done for scrape targets?

rules/manager.go
wbh1: Ah, yes -- I misunderstood. I can update this today to use the `start` variable, then.

rules/manager.go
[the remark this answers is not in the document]
juliusv: Yeah, we could do that too and it would seem more consistent with the targets page. Then it would also solve our template function problem as we can just use `humanizeDuration` again :)

@wbh1 What do you think, is there a need for the specific timestamp or would a duration since the last evaluation also be sufficient?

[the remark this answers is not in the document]
wbh1: That works for me.

Is everyone okay with the column header "Last Evaluation" and using the "\<seconds\> ago" format?

(Like how /targets uses: ```{{if .LastScrape.IsZero}}Never{{else}}{{since .LastScrape}} ago{{end}}``` )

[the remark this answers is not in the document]
wbh1: This has been updated (and it looks nicer, too. Thanks, @simonpasquier ! 😄)

kdada:
LGTM

web/ui/templates/rules.html
```diff
@@ -19,6 +20,7 @@ <h2>Rules</h2>
             <td style="font-weight:bold">State</td>
             <td style="font-weight:bold">Error</td>
             <td style="font-weight:bold">Evaluation Time</td>
+            <td style="font-weight:bold">Last Evaluation</td>
```
juliusv: Could you move the "Last Evaluation" before "Evaluation Time"? It would be more consistent with how we show the last scrape and scrape duration for the target.

juliusv:
:+1: besides last nit

web/ui/templates/rules.html
wbh1: Updated 😃

wbh1:
@simonpasquier I've re-run `make assets` and I think it should be good now. All tests in `make test` passed, too. Let me know if I need to do anything else.

simonpasquier:
For anybody wanting to give it a try, the build output is https://8821-6838921-gh.circle-artifacts.com/0/build/prometheus

simonpasquier:
:+1:

simonpasquier:
Thanks a lot for your contribution!

PR #4723, opened 2018-10-10
Make prometheus more GoLand and VSCode friendly

tariq1890:
This PR removes minor inconveniences for GoLand and VSCode users


gouthamve:
Hmm, iirc, this has come up before, and we decided not to support editor specific stuff. See: https://github.com/prometheus/prometheus/pull/3869/files

I think the reason was that there are just too many editor configs to support and that would be untenable. Let me know your thoughts on this.

tariq1890:
That certainly makes sense . Thanks @gouthamve . Closing this

PR #4731, opened 2018-10-12
*: check that assets are up-to-date

simonpasquier:
It happens very often that contributors forget to run `make assets`. This PR adds a `check_assets` target to the Makefile that fails if the generated code doesn't match with the source. It updates the CI configurations accordingly.

`vfsgen` can't generate deterministic content when executed from different systems (because file timestamps aren't stable). To circumvent this issue, I've taken existing code from AlertManager (https://github.com/prometheus/alertmanager/pull/1532) and copy/paste it in this repository (namely `pkg/modtimevfs/modtimevfs.go`). Another option would be to move this code to prometheus/common but https://github.com/prometheus/common/issues/147 would probably needs to be resolved first.

cc @juliusv


simonpasquier:
@brancz @brian-brazil please take a look.

brancz:
lgtm :+1:

PR #4743, opened 2018-10-15
Drop 'else' condition when 'if' ends with a return

vandewilly:
This commit drops the else condition when the if block ends with a
return statement.


brian-brazil:
Thanks!

PR #4746, opened 2018-10-16
Moving wrong positioned comment in test code

gangseok514:
"// Range queries." comment in incorrect line.

Moving it to above "metric[20s]"

### DIFF
--- a/promql/engine_test.go
+++ b/promql/engine_test.go
@@ -323,6 +323,7 @@ load 10s
 			},
 			Start: time.Unix(1, 0),
 		},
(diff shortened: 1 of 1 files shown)


brian-brazil:
These are correct, it's the end time that makes it a range query.

gangseok514:
sorry. the "metric[20s]" is not range query?

gangseok514:
i misunderstood.sorry for your inconvenience :)

brian-brazil:
`metric[20s]` is a range vector selector.

PR #4750, opened 2018-10-16
add a metric for the last time we appended metrics to local storage

cstyan:
@krasi-georgiev as discussed on https://github.com/prometheus/tsdb/pull/401

I don't know how we could solve https://github.com/prometheus/tsdb/issues/395 because we kind of have to get that metric from within TSDB itself. We have no information about Head/WAL truncation and what the oldest metric timestamp is from Prometheus itself.

### DIFF
--- a/storage/tsdb/tsdb.go
+++ b/storage/tsdb/tsdb.go
@@ -35,8 +35,9 @@ var ErrNotReady = errors.New("TSDB not ready")
 // ReadyStorage implements the Storage interface while allowing to set the actual
 // storage at a later point in time.
 type ReadyStorage struct {
-	mtx sync.RWMutex
-	a   *adapter
+	mtx                sync.RWMutex
+	a                  *adapter
+	lastSamplesAddTime prometheus.Gauge
 }
 
 // Set the storage.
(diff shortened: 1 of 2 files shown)


storage/tsdb/tsdb.go
```diff
@@ -236,6 +251,8 @@ func (a appender) Add(lset labels.Labels, t int64, v float64) (uint64, error) {
 	case tsdb.ErrOutOfBounds:
 		return 0, storage.ErrOutOfBounds
 	}
+	m := *(a.lastSamplesAddTime)
+	m.SetToCurrentTime()
```
brian-brazil: This is pretty expensive to do on every single ingested sample.

cstyan: Agreed. If you have suggestions for another way to get the same or similar data I'm open to it. The use case is in https://github.com/prometheus/tsdb/pull/401, but the short of it is that I would like to be able to know how far behind the local storage (in terms of storing metrics) a given remote write is by comparing the last time we ingested samples vs the last time we successfully sent samples.

brian-brazil: Queue size would be the usual way to do that.

cstyan: The problem with that is that the queue mechanics are changing as part of my remote write via WAL PR.

krasi-georgiev:
Just realised that you can use prometheus_tsdb_head_max_time (soon to be renamed to prometheus_tsdb_head_max_time_milliseconds)

krasi-georgiev:
ping @cstyan

cstyan:
@krasi-georgiev sorry I'm still in the middle of the remote write stuff and I haven't been keeping track of what metrics I've added there that I actually need. This can be closed for now and I'll either reopen or open a PR with the correct changes.

PR #4751, opened 2018-10-17
Reordering fields of String datetype

gangseok514:
Other types has timestamp as first field, but only String type has value as first field.

Signed-off-by: gangseok.lee <gangseok.lee@samsung.com>


brian-brazil:
Thanks!

PR #4752, opened 2018-10-17
web: deduplicate handler label for HTTP metrics

simonpasquier:
Some HTTP handlers for the UI and the API v1 have identical names (eg `alerts`, `rules`, ...) but the `handler` label for HTTP metrics didn't distinguish between them.

Before:

image

After:

image

### DIFF
--- a/web/web.go
+++ b/web/web.go
@@ -171,6 +171,14 @@ type Options struct {
 	RemoteReadConcurrencyLimit int
 }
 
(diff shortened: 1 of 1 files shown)


simonpasquier:
@gouthamve PTAL

simonpasquier:
@brian-brazil @gouthamve please take a look

brian-brazil:
:+1:

PR #4753, opened 2018-10-17
discovery: add config label to discovered targets metric

simonpasquier:
I was hacking on the discovery part and read about @SuperQ request to get the number of discovered targets per scrape config (cc @brian-brazil too).

In any case, I've noticed that the current metric is broken because Prometheus spawns 2 discovery managers (one for scraping and the other one for alerting) and the value actually depends on which manager has updated the metric last.

### DIFF
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -47,11 +47,12 @@ var (
 			Help: "Total number of service discovery configurations that failed to load.",
 		},
 	)
-	discoveredTargets = prometheus.NewGauge(
+	discoveredTargets = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "prometheus_sd_discovered_targets",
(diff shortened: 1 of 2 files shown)


discovery/manager.go
```diff
@@ -106,12 +107,12 @@ type provider struct {
 	config interface{}
 }
 
-// NewManager is the Discovery Manager constructor
-func NewManager(ctx context.Context, logger log.Logger) *Manager {
+// NewManager is the Discovery Manager constructor.
+func NewManager(ctx context.Context, logger log.Logger, options ...func(*Manager)) *Manager {
```
simonpasquier: I didn't want to change the signatute of `NewManager()` as it is imported by file-based discovery systems. I stole the idea of variadic option functions from https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

PR #4763, opened 2018-10-19
use rangehead for queries and pre allocate chunk slices.

krasi-georgiev:
benching some tsdb optimisations from https://github.com/prometheus/tsdb/pull/351

### DIFF
--- a/vendor/github.com/prometheus/tsdb/db.go
+++ b/vendor/github.com/prometheus/tsdb/db.go
@@ -793,7 +793,11 @@ func (db *DB) Querier(mint, maxt int64) (Querier, error) {
 		}
 	}
 	if maxt >= db.head.MinTime() {
-		blocks = append(blocks, db.head)
+		blocks = append(blocks, &rangeHead{
+			head: db.head,
+			mint: mint,
+			maxt: maxt,
+		})
 	}
 
 	sq := &querier{
--- a/vendor/github.com/prometheus/tsdb/querier.go
+++ b/vendor/github.com/prometheus/tsdb/querier.go
@@ -504,6 +504,14 @@ func (s *baseChunkSeries) Next() bool {
 		err      error
 	)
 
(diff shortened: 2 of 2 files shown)


krasi-georgiev:
@free so far doesn't show a any difference in the mem/cpu usage. Maybe need to wait a bit more to accumulate more data.

This is the current config for the queries.
https://github.com/prometheus/prombench/blob/master/componenquerier:s/prombench/manifests/benchmark/2_loadgen.yaml#L8

free:
[Here](http://prombench.prometheus.io/prometheus-meta/graph?g0.range_input=6h&g0.expr=rate(go_memstats_alloc_bytes_total%7Bjob%3D%22prometheus%22%2Cnamespace%3D%22prombench-4763%22%7D%5B5m%5D)&g0.tab=0&g1.range_input=6h&g1.expr=rate(go_memstats_mallocs_total%20%7Bjob%3D%22prometheus%22%2Cnamespace%3D%22prombench-4763%22%7D%5B5m%5D)&g1.tab=0)'s the only material difference I could find: a 10-20% reduction in allocations and allocated bytes. For whatever reason, it is not reflected in CPU utilization, though. I imagine it's outweighed by all the scraping and queries.

free:
FYI, there appears to be an [up to 10% difference](http://prombench.prometheus.io/prometheus-meta/graph?g0.range_input=7h&g0.end_input=2018-10-19%2016%3A02&g0.step_input=60&g0.expr=rate(prometheus_engine_query_duration_seconds_count%7Bjob%3D%22prometheus%22%2Cslice%3D%22inner_eval%22%2Cnamespace%3D%22prombench-4763%22%2Cprometheus%3D~%22(test%2B).*%22%7D%5B1h%5D)&g0.tab=0) in query load between the 2 Prometheus instances (first graph), varying over time (including which server gets more queries). I don't think that's conducive to a relevant comparison.

Issue #2119, opened 2016-10-25
Limit number of metrics displayed in prometheus UI


discordianfish:
IMO it's too easy to crash the browser by entering a expression returning too many elements. It should be straight forward to only display the first X elements a query returns and print a warning.
X should be really high so the limit only kicks in in cases where displaying the data would takes several gigs of memory.

Gouthamve:
I think this is related to #2253

Fixing one will fix another (by using a parameter)

brian-brazil:
That's not related.

fabxc:
Yes, #2253 talks about the horizontal axis, i.e. how many datapoints are
evaluated for any particular expression. That affects any number of time
series this expression may select.
The problem described here talks about the vertical axis of how many result
time series there can be.

Probably both should be configurable by the user to be honest. There are
many ways how a query can be made expensive enough to hog a Prometheus.
Having a hardcoded limit on a single one of them doesn't seem particularly
useful.

brian-brazil:
I ran into a case a while back where labels were being put in metric names, and this was taking out my browser (plus chewing through CPU for prometheus). I think a hardcoded limit on the order of 10-100k should do the trick.

EyesBear:
Is this issue still up for grabs? I am interested.

brian-brazil:
Noone is working on this currently. This is the the v1 label API.

PrincipalsOffice:
Should the limit be implemented on the v1 instead of v2 api then?

applegrain:
Even if the results aren't crashing your browser, the way the timeseries are rendered today (both on the `Graph` and `Console` view) isn't very easy to read or skim. In addition to limiting the results returned, perhaps better filtering would be useful here as well? For example, you could only display timeseries if certain labels are present, or even specific label-value pairs.

@PrincipalsOffice are you still working on this? If not, I have time to work on this.

brian-brazil:
Such features are more for Grafana. This issue is about avoiding taking out browsers.

frittentheke:
This is still very much an issue with Prometheus >= 2.2.1. Even on a beefy machine with a current browser the UI fetches such large results when doing a character by character typeahead it sometimes even crashes.

Maybe a quick and dirty solution (before applying a limit on the number of results via the API) is to only start the type-ahead after like 3 or 5 characters or to allow disabling it all-together and only start the search after I press enter?

bechtoldt:
[...]

PR #4765, opened 2018-10-20
Limit number of metrics displayed in prometheus UI

potomak:
Limits the number of of metrics displayed in prometheus UI to 10k and
displays a warning message in case the API returns more than 10k
results.

#### TODO

- [ ] run `make assets` to re-generate `web/ui/assets_vfsdata.go`

See https://github.com/prometheus/prometheus/issues/2119.

### DIFF
--- a/web/ui/static/js/graph/index.js
+++ b/web/ui/static/js/graph/index.js
@@ -2,6 +2,7 @@ var Prometheus = Prometheus || {};
 var graphTemplate;
 
 var SECOND = 1000;
+var METRICS_MAX_SIZE = 10000;
 
 /**
  * Graph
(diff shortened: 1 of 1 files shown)


codesome:
If I understand correctly, no. of metrics mean the no. of series displayed on the UI. This doesn't seem to be working. I don't think that is the right place to add that check.

PS: You need to run `make assets` and build prometheus to test any changes in UI elements.

PPS: 
* I get the warning without even running any query.
* I set the limit to `10` and tested. `{__name__=~".+"}`  displayed lot more than 10 series.

potomak:
@codesome thanks for the feedback.

> PS: You need to run make assets and build prometheus to test any changes in UI elements.

I'll make sure to run `make assets` to re-generate `web/ui/assets_vfsdata.go` before merging the PR if approved.

> If I understand correctly, no. of metrics mean the no. of series displayed on the UI. This doesn't seem to be working. I don't think that is the right place to add that check.

Maybe I misunderstood the problem. I'll add a comment in the original issue to get more feedback.

> I get the warning without even running any query.

Yes, that's "by design". The query is _static_, meaning that it just searches in a list of metrics that's loaded when a new graph is created. For this reason I thought it was more appropriate to display a single warning alert at the top of the first chart.

This approach has two problems though:

1. If you remove the first chart, there's no way to get alerted anymore, because the code prepends the alert element to `#graph_wrapper0` (id: 0)
2. If the error persists and you add a new chart, a duplicate alert is prepended, see attached image

Note: both of these problems existed already for the time drift warning message.

#### Example duplicate warning error

screen shot 2018-10-24 at 9 56 34 am

simonpasquier:
@potomak would you have time to address my comments?

vn-ki:
@potomak Are you still working on this?

Ownercz:
@potomak any update on this issue, please?

codesome:
@vn-ki go ahead!

PR #4766, opened 2018-10-21
Revert #4586

gouthamve:
From #4586, @grobie:

> This is a breaking change and will change/invalidate relabel rules based on __address__ labels afaics. While I see the motivation for this change, it has serious side effects for users of the DNS discovery.

While is indeed breaking some people, it is also technically a bug. Not sure if we should keep it or revert it.

### DIFF
--- a/discovery/dns/dns.go
+++ b/discovery/dns/dns.go
@@ -181,6 +181,9 @@ func (d *Discovery) refresh(ctx context.Context, name string, ch chan<- []*targe
 		target := model.LabelValue("")
 		switch addr := record.(type) {
 		case *dns.SRV:
+			// Remove the final dot from rooted DNS names to make them look more usual.
+			addr.Target = strings.TrimRight(addr.Target, ".")
+
 			target = hostPort(addr.Target, int(addr.Port))
 		case *dns.A:
 			target = hostPort(addr.A.String(), d.port)


gouthamve:
From @grobie on IRC:
> not sure that xkcd is too relevant here gouthamve[m]. We are not relying on some undefined obscure side effect, but the intended interface of the dns discovery. as the DNS root dot is practically always omitted in actual usage, it was a purposefully stripped in the discovery implementation.

> the question boils down to the extent of breaking changes we find acceptable in minor releases.
one thing to keep in mind, while for soundcloud this change "only" affects relabel rules and epxressions (which we can probably all handle server side together with the upgrade), this change can break scraping for people using reverse proxies in front of their targets. not many setups handle the root dot

> see https://prometheus.io./ ...

> personally, I think this slight optimiziation is not worth such breakage, but that's my 2 cents

Given `prometheus.io.` is broken in my browser, I am kinda inclined to revert the change. But I am no expert in DNS.

simonpasquier:
Yeah, it is safer to revert for now. From the standpoint of the standards, the final dot shouldn't be breaking stuff but it looks like in practice many systems aren't ready to handle this change.

brian-brazil:
> the final dot shouldn't be breaking stuff but it looks like in practice many systems aren't ready to handle this change.

That's been the case for over a decade. This might be better handled by the user adjusting ndots.

simonpasquier:
Yeah, it is safer to revert for now. From the standpoint of the standards, the final dot shouldn't be breaking stuff but it looks like in practice many systems aren't ready to handle this change.

cc @bboreham that committed the initial PR.

grobie:
We agree that ideally we would not alter the DNS name before doing another lookup for the returned target. The problem is that the returned identifier is not only used for additional DNS lookups, but also in other contexts outside of DNS, e.g. as HTTP Host header, as TLS server name, and generally as Prometheus instance label. While the final dot makes sense in the DNS context, it doesn't make sense in any other context. 

While stripping will only result in additional lookups in some environments, keeping the dot will break setups with existing relabel rules, and setups where TLS/HTTP reverse proxies in front of targets don't handle the unexpected dot.

There are at least two options to avoid the additional DNS lookups: 1) as @brian-brazil mentioned, change the `/etc/resolv.conf` `ndots` option accordingly to the length of your target names. 2) configure your DNS recursors / caches to immediately return a NXDOMAIN for known impossible lookups (basic example `\.(com|net|org)\.dc[0-9]\.company\.org\.` matches all requests were top level domain lookups are made against our `dc4.company.org` search domain). I've done that at SoundCloud and cut down our DNS lookups to a small percentage of the original rate. Both of these workarounds will cover all applications, and not only the lookups made by Prometheus.

Weighting the arguments, I'm in favor of keeping the behavior of the last release and merge this PR.

juliusv:
:+1: I don't know what the right thing here is long-term, but from a breakage perspective we probably want to get this into 2.5.0 to not break a bunch of users.

grobie:
I agree we should think about and work on a DNS discovery implementation which respects the target names returned by the SRV lookup. It's not as easy as it looks, just stopping to strip the final dot has too many side effects.

PR #4674, opened 2018-09-28
web/api/v1: fix optional skip_head for snapshot

simonpasquier:
Closes #4671

I started to add a unit test but this would involve quite a bit of refactoring in the unit tests. It would be beneficial to add more tests for the other admin actions too so I propose to tackle it in `master` once this gets merged back.

PR #4767, opened 2018-10-22
web/v1/api: add tests for admin actions

simonpasquier:
As promised in https://github.com/prometheus/prometheus/pull/4674#issuecomment-425446511, I've added tests for the admin endpoints that weren't tested at all.


simonpasquier:
cc @gouthamve

simonpasquier:
@gouthamve friendly ping.

PR #4770, opened 2018-10-22
Fix promql benchmarks

csmarchbanks:
Previously every benchmark was failing since they query more than 10 samples. MaxSamples now matches the default value used when starting Prometheus.


brian-brazil:
Thanks!

PR #4772, opened 2018-10-23
improve remote storage enqueue performance

fengyuceNv:
Signed-off-by: fyc <fyc22788@ly.com>
The writes of the remote store slows down the local storing performance, because the remote uses the write lock for enqueue. Read lock is better in this case as the metrics enqueue is not blocked. My app has benefited from the performance after using the read lock.

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -149,7 +149,7 @@ type QueueManager struct {
 	queueName      string
 	logLimiter     *rate.Limiter
 
-	shardsMtx   sync.Mutex
+	shardsMtx   sync.RWMutex
 	shards      *shards
 	numShards   int
 	reshardChan chan int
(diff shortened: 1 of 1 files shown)


fengyuceNv:
before using the read lock, remote storage sync affect local storage scrape ,our scrape interval is 1 min,but most targets scrape need 7min
image
after using the read lock , all targets complete in a min
image

BertHartm:
I can confirm this issue still exists on 2.5.0, and that this branch fixes it. Our performance drops off at about 80k samples/s on 2.4.3 and about 90k samples/s on 2.5.0, and this branch lets us get to a much more stable 125k, and drops all the lag we've seen around scraping and rule evaluation

Right now remote_write is effectively unusable for us because of this issue.

@tomwilkie would you be the right person to review this?

tomwilkie:
Thanks @fengyuceNv!  I'll give this a review today.

tomwilkie:
@gouthamve beat me to it; LGTM from me too though!

Issue #3030, opened 2017-08-07
vendor.json in dev-2.0 is broken


AlekSi:
Please look there: https://github.com/prometheus/prometheus/blob/dev-2.0/vendor/vendor.json Search for `84bc9597164f671c0130543778228928d6865c5c`. You will see it for too many import paths. This commit exists in cockroachdb repository (https://github.com/cockroachdb/cockroach/commit/84bc9597164f671c0130543778228928d6865c5c), but in https://github.com/petermattis/goid, for example. Notice also that those origin paths contain `/vendor/`. I can only guess that old version of Go or old version of govendor was used to update vendor.json.

The end result is that `govendor sync` fails with Go 1.8.3 and govendor 1.0.8.

Refs #3022.

grobie:
Fixed in #3400.

PR #4781, opened 2018-10-24
Fix errors in the vendor/ directory

simonpasquier:
This PR fixes several inconsistencies in the actual vendoring. This is a prerequisite to switch from `govendor` to `go mod`.

* Some indirect dependencies were tracked from the vendored repository (eg `k8s.io/client-go/1.5/vendor/github.com/davecgh/go-spew/spew`). Those dependencies are now tracking the same version of the code but from the original repository (eg `github.com/davecgh/go-spew/spew`).
* Some packages were still referenced in `vendor/vendor.json` but not vendored because they weren't needed anymore (for instance, `github.com/dgryski/go-bits`).
* Fix invalid revisions for a number of packages pulled by `github.com/cockroachdb/cockroach` (see #3030).

The end result is that after this PR, `rm -rf vendor/*/* && govendor sync` shouldn't fail and `git diff` returns no difference.


simonpasquier:
@brancz @brian-brazil PTAL.

simonpasquier:
cc @SuperQ too

brancz:
lgtm :+1:

PR #4785, opened 2018-10-25
Add documentation for new marathon-sd functionality

ti-mo:
Hi all

Since #4499 (bound to drop in 2.5.0(?)), the functionality has expanded somewhat, and the documentation leaves it pretty much up to the user to scour the (very few) resources on the internet on how to configure port labels and making it all work together. This can be quite daunting as a first-timer, so this example configuration might help.

This might not be in the correct format, but I tried following the convention of the existing Kubernetes example. It's dense on comments, but I figured it was a better place than putting it in `configuration.md` itself.

Please let me know how this can be improved, or if there's anything missing.

### DIFF
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -874,8 +874,12 @@ tls_config:
 
 By default every app listed in Marathon will be scraped by Prometheus. If not all
 of your services provide Prometheus metrics, you can use a Marathon label and
-Prometheus relabeling to control which instances will actually be scraped. Also
-by default all apps will show up as a single job in Prometheus (the one specified
+Prometheus relabeling to control which instances will actually be scraped.
+See [the Prometheus marathon-sd configuration file](/documentation/examples/prometheus-marathon.yml)
+for a practical example on how to set up your Marathon app and your Prometheus
+configuration.
+
+By default, all apps will show up as a single job in Prometheus (the one specified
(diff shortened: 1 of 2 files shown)


docs/configuration/configuration.md
```diff
@@ -874,8 +874,12 @@ tls_config:
 
 By default every app listed in Marathon will be scraped by Prometheus. If not all
 of your services provide Prometheus metrics, you can use a Marathon label and
-Prometheus relabeling to control which instances will actually be scraped. Also
-by default all apps will show up as a single job in Prometheus (the one specified
+Prometheus relabeling to control which instances will actually be scraped.
+See [the Prometheus marathon-sd configuration file](/documentation/examples/prometheus-marathon.yml)
```
brancz: @simonpasquier @grobie do you know if this link would actually work in the docs?

simonpasquier: It should work as it is already used this way in the `kubernetes_sd_config` section:
https://github.com/prometheus/prometheus/blame/36143be2349f414623a45ffe6143bc9ae71cb4d4/docs/configuration/configuration.md#L804

brancz: Ack. Will merge and cherrypick into release branch, as I feel it should in there.

brancz:
lgtm :+1:

PR #4792, opened 2018-10-26
Makefile.common: use gotestsum if present

simonpasquier:
Using gotestsum with CircleCI, we can gain more insights on failing tests. See https://circleci.com/docs/2.0/collect-test-data/ and https://circleci.com/blog/level-up-go-test-with-gotestsum/ for more details.

~~This can't be used currently for the Prometheus repository itself because of resource limitations in CircleCI but it is beneficial for the other repositories.~~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,6 +51,8 @@ jobs:
         key: v1-npm-deps-{{ checksum "web/ui/react-app/yarn.lock" }}
         paths:
         - web/ui/react-app/node_modules
+    - store_test_results:
+        path: test-results
   test_windows:
     executor: win/default
     working_directory: /go/src/github.com/prometheus/prometheus
--- a/Makefile.common
+++ b/Makefile.common
@@ -69,6 +69,15 @@ else
 	GO_BUILD_PLATFORM ?= $(GOHOSTOS)-$(GOHOSTARCH)
 endif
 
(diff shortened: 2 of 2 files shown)


simonpasquier:
cc @SuperQ

SuperQ:
I don't think we need to skip with `CIRCLE_JOB`, because we don't run `test` at all in circle.

simonpasquier:
> we don't run test at all in circle.

Yes for Prometheus itself but I want to use this for alertmanager and other projects that run tests with CircleCI so I need to make the change here before propagating to the other repositories.

SuperQ:
@simonpasquier But for other projects, we might want to run this in CircleCI, as they're not as resource intensive.

simonpasquier:
@SuperQ please have another look. I've resurrected this PR since we now use Circle CI to run Prometheus tests.

simonpasquier:
@SuperQ would you mind having a look at this?

roidelapluie:
@simonpasquier will it still work with the new go 1.14 test output?

simonpasquier:
@roidelapluie which new output? I couldn't find anything in the release [notes](https://tip.golang.org/doc/go1.14).

roidelapluie:
`go test -v now streams t.Log output as it happens, rather than at the end of all tests. `

but indeed it does not affect output.

PR #2201, opened 2016-11-18
discovery: terminate senders before closing channel

fabxc:
Fixes #2200

@juliusv

PR #4796, opened 2018-10-27
Included default flags into Dockerfile entrypoint

### LINKED ISSUE #4776: Use Docker ENTRYPOINT
## Proposal

Use Docker `ENTRYPOINT`[0] rather than `CMD` for setting the default command line flag values.

When CMD is used, any additional flags cause the default flags to be dropped.

This makes starting up Prometheus with a simple additional flag like `--storage.tsdb.retention=15d` to need the full list of flags added.

**Environment**

Docker

[0]: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#entrypoint

DrakeW:
Fixes #4776  
Changes:
1. Moved common default flags into ENTRYPOINT to avoid re-entering all flags when custom flags is needed
2. Left `--config.file` flag in CMD since it's a flag that's likely to be changed and referred to by the user at a custom location.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,5 +18,5 @@ WORKDIR    /prometheus
 ENTRYPOINT [ "/bin/prometheus", \
              "--storage.tsdb.path=/prometheus", \
              "--web.console.libraries=/usr/share/prometheus/console_libraries", \
-             "--web.console.templates=/usr/share/prometheus/consoles"]
+             "--web.console.templates=/usr/share/prometheus/consoles" ]
 CMD        [ "--config.file=/etc/prometheus/prometheus.yml" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,8 +15,8 @@ USER       nobody
 EXPOSE     9090
 VOLUME     [ "/prometheus" ]
 WORKDIR    /prometheus
-ENTRYPOINT [ "/bin/prometheus" ]
-CMD        [ "--config.file=/etc/prometheus/prometheus.yml", \
+ENTRYPOINT [ "/bin/prometheus", \
              "--storage.tsdb.path=/prometheus", \
              "--web.console.libraries=/usr/share/prometheus/console_libraries", \
-             "--web.console.templates=/usr/share/prometheus/consoles" ]
+             "--web.console.templates=/usr/share/prometheus/consoles"]
+CMD        [ "--config.file=/etc/prometheus/prometheus.yml" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,6 @@ VOLUME     [ "/prometheus" ]
 WORKDIR    /prometheus
 ENTRYPOINT [ "/bin/prometheus", \
              "--storage.tsdb.path=/prometheus", \
-             "--web.console.libraries=/usr/share/prometheus/console_libraries", \
-             "--web.console.templates=/usr/share/prometheus/consoles" ]
-CMD        [ "--config.file=/etc/prometheus/prometheus.yml" ]
+             "--web.console.libraries=/etc/prometheus/console_libraries", \
+             "--web.console.templates=/etc/prometheus/consoles", \
+             "--config.file=/etc/prometheus/prometheus.yml" ]


Dockerfile
```diff
@@ -15,8 +15,8 @@ USER       nobody
 EXPOSE     9090
 VOLUME     [ "/prometheus" ]
 WORKDIR    /prometheus
-ENTRYPOINT [ "/bin/prometheus" ]
-CMD        [ "--config.file=/etc/prometheus/prometheus.yml", \
+ENTRYPOINT [ "/bin/prometheus", \
              "--storage.tsdb.path=/prometheus", \
              "--web.console.libraries=/usr/share/prometheus/console_libraries", \
-             "--web.console.templates=/usr/share/prometheus/consoles" ]
+             "--web.console.templates=/usr/share/prometheus/consoles"]
```
skorpy2009: Style difference `"]` => `" ]`
```suggestion
             "--web.console.templates=/usr/share/prometheus/consoles" ]
```

DrakeW: good catch and applied suggestion!

simonpasquier:
> Left --config.file flag in CMD since it's a flag that's likely to be changed and referred to by the user at a custom location.

I'm not sure what it buys us. If I want to change a random flag, I would have to specify the `--config.file` flag even though I'm fine with the default value.

DrakeW:
@simonpasquier - that's a valid point, I don't quite remember why I left it in `CMD` but it might has something to do with the args list when I inspect a container that was run with entrypoint flags overwritten like below (there's duplicates flags, but I guess `prometheus` by itself take the latest appearance)
```
"Path": "/bin/prometheus",
 "Args": [
            "--storage.tsdb.path=/prometheus",
            "--web.console.libraries=/usr/share/prometheus/console_libraries",
            "--web.console.templates=/usr/share/prometheus/consoles",
            "--abc=123",
            "--storage.tsdb.path=456"
],
```
I'll move it to `ENTRYPOINT`

simonpasquier:
You're right: the container runtime will append (and not merge) the `CMD` arguments to the arguments defined in `ENTRYPOINT`. The caveat is that with this patch one can't override any of the following CLI arguments: `storage.tsdb.path`, `--web.console.libraries` and `--web.console.templates` but I think it is acceptable.

@SuperQ thoughts?

SuperQ:
I don't see any major reason to need to change those paths inside the standard image. Typically the user will only need to control the external path via volume configuration.

SuperQ:
One thing we may want to do is adjust the defaults to point to `/etc/prometheus/consoles`, this way it would be easier for users to volume map in a whole tree of `/etc/prometheus` configuration.

I guess I don't know what the Docker standard is for this kind of thing.

Dockerfile
```diff
@@ -15,8 +15,8 @@ USER       nobody
 EXPOSE     9090
 VOLUME     [ "/prometheus" ]
 WORKDIR    /prometheus
-ENTRYPOINT [ "/bin/prometheus" ]
-CMD        [ "--config.file=/etc/prometheus/prometheus.yml", \
+ENTRYPOINT [ "/bin/prometheus", \
              "--storage.tsdb.path=/prometheus", \
              "--web.console.libraries=/usr/share/prometheus/console_libraries", \
              "--web.console.templates=/usr/share/prometheus/consoles" ]
+CMD        [ "--config.file=/etc/prometheus/prometheus.yml" ]
```
SuperQ: This should also be in the `ENTRYPOINT`.

Dockerfile
```diff
@@ -15,8 +15,8 @@ USER       nobody
 EXPOSE     9090
 VOLUME     [ "/prometheus" ]
 WORKDIR    /prometheus
-ENTRYPOINT [ "/bin/prometheus" ]
-CMD        [ "--config.file=/etc/prometheus/prometheus.yml", \
+ENTRYPOINT [ "/bin/prometheus", \
              "--storage.tsdb.path=/prometheus", \
              "--web.console.libraries=/usr/share/prometheus/console_libraries", \
```
SuperQ: I think we should move the console templates/libraries into `/etc/prometheus` for consistency.

Dockerfile
DrakeW: updated

Dockerfile
DrakeW: good point, updated

DrakeW:
@SuperQ - Awesome! Please feel free to merge it since I don't have write access

matthiaslee:
This change is/will be breaking most charts and operators deploying prometheus, since most are passing in a set of [arguments](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#container-v1-core), instead of overriding the command. Since prometheus doesn't handle repeat arguments, it just crashes, including the `coreos/prometheus-operator` ([#2201](https://github.com/coreos/prometheus-operator/issues/2201))

```
Error parsing commandline arguments: flag 'config.file' cannot be repeated
```

redbaron:
Any chance to pull it out? it is going to have long lasting ripple effect across whole ecosystem :-/

matthiaslee:
@redbaron, looks like the answer is yes: https://github.com/prometheus/prometheus/pull/4976

simonpasquier:
For everyone looking at this issue, #4976 has been merged and v2.6.0 shouldn't break existing deployments. #4776 is still fixed meaning that it isn't required to pass the full list of flags when you want to change only one flag.

PR #4798, opened 2018-10-27
Bump Go versions and use '.x' to always get latest minor versions

HaraldNordgren:


simonpasquier:
We want to control which Go version is used to test & build Prometheus because we've already been hit by changes/regressions. Also this isn't the only place where the change would have to be made.

HaraldNordgren:
@simonpasquier Understood. Close this PR if you want to.

PR #4806, opened 2018-10-30
web: display job label in targets page

simonpasquier:
#4526 incidentally changed the layout of the targets page as targets are now grouped by scrape configurations instead of jobs (see https://github.com/prometheus/prometheus/pull/4526/files#diff-73cbd2009b0e4f156e8ca0f47e95b016L677). From discussions on IRC with @beorn7 and @brian-brazil, this isn't a bad thing but the `job` label shouldn't be filtered anymore.

cc @brancz


simonpasquier:
After the change:

image

brian-brazil:
Are you sure that's the right image? There's no job label there.

simonpasquier:
@brian-brazil oops my bad! I've updated the comment.

brian-brazil:
:+1:

brancz:
lgtm :+1:

PR #4810, opened 2018-10-31
Cherrypick 4806 4809

brancz:
@simonpasquier @brian-brazil @beorn7 @SuperQ @krasi-georgiev


beorn7:
LGTM

(Just FTR: Like the EU, cherrypicking makes me sad, but I guess that's what you have settled on…)

PR #4816, opened 2018-11-01
For tsdb#441

brian-brazil:


krasi-georgiev:
@brian-brazil why do you think the bench doesn't show any improvements?

brian-brazil:
On that Prometheus there's only around 150MB of indexes extra loaded during a compaction, the effect would be more pronounced if there were cardinality issues.

sipian:
@brian-brazil 
If you are done with the benchmarking you can cancel it to delete the resources.

krasi-georgiev:
not sure if you noticed, but looks like the v2.4.3 went into some sort of crash loop

http://prombench.prometheus.io/grafana/d/7gmLoNDmz/prombench?orgId=1&var-RuleGroup=All&var-pr-number=4816&from=1541732236144&to=1541759216649

krasi-georgiev:
another reason why we need to keep the logs somewhere

https://github.com/prometheus/prombench/issues/157

PR #4819, opened 2018-11-02
discovery: send empty group on empty SD config

### LINKED ISSUE #4812: Prometheus continues to scrape removed targets from empty scrape configs
## Bug Report

**What did you do?**

Removed last entry of a `static_configs` directive. Our template generator omits empty config blocks from being written out to disk. The written out scrape configuration didn't have any `*_sd_configs` or `static_configs` configuration block.

Either Prometheus rejects scrape configurations without a single discovery config (neither static nor any `*_sd_configs`), or it handles such empty scrape configurations correctly.

Prometheus didn't show any targets in the targets page, but continued to scrape the target from the last static_configs (with up == 0 as the target was deprovisioned).

**Environment**

* System information:

	Linux 4.18

* Prometheus version:

	2.4.2

@krasi-georgiev

simonpasquier:
Closes #4812 

cc @krasi-georgiev @grobie

### DIFF
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -408,7 +408,9 @@ func (m *Manager) registerProviders(cfg sd_config.ServiceDiscoveryConfig, setNam
 		})
 	}
 	if !added {
-		// Add an empty target group to force the refresh of the corresponding scrape pool.
+		// Add an empty target group to force the refresh of the corresponding
+		// scrape pool and to notify the receiver that this target set has no
+		// current targets.
 		add(setName, func() (Discoverer, error) {
 			return &StaticProvider{TargetGroups: []*targetgroup.Group{&targetgroup.Group{}}}, nil
 		})
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -314,11 +314,13 @@ func (m *Manager) allGroups() map[string][]*targetgroup.Group {
 }
 
 func (m *Manager) registerProviders(cfg sd_config.ServiceDiscoveryConfig, setName string) {
+	var added bool
 	add := func(cfg interface{}, newDiscoverer func() (Discoverer, error)) {
 		t := reflect.TypeOf(cfg).String()
 		for _, p := range m.providers {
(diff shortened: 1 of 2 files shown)


simonpasquier:
Friendly ping for @grobie and @krasi-georgiev

discovery/manager.go
```diff
@@ -401,7 +404,13 @@ func (m *Manager) registerProviders(cfg sd_config.ServiceDiscoveryConfig, setNam
 	}
 	if len(cfg.StaticConfigs) > 0 {
 		add(setName, func() (Discoverer, error) {
-			return &StaticProvider{cfg.StaticConfigs}, nil
+			return &StaticProvider{TargetGroups: cfg.StaticConfigs}, nil
+		})
+	}
+	if !added {
+		// Add an empty target group to force the refresh of the corresponding scrape pool.
```
krasi-georgiev: also add some more explicit comments that is is required for when reloading with an empty StaticConfigs

discovery/manager_test.go
```diff
@@ -794,6 +795,64 @@ scrape_configs:
 	verifyPresence(t, discoveryManager.targets, poolKey{setName: "prometheus", provider: "string/0"}, "{__address__=\"bar:9090\"}", false)
 }
 
+func TestTargetSetRecreatesEmptyTargetGroup(t *testing.T) {
```
krasi-georgiev: how about something like.
```suggestion
// TestTargetSetRecreatesEmptyStaticConfigs ensures that reloading a config file after 
// removing all targets from the static_configs sends an update with empty targetGroups.
// This is required to signal the receiver that this target set has no current targets.
func TestTargetSetRecreatesEmptyStaticConfigs(t *testing.T) {
```

krasi-georgiev:
LGTM

as part of reviewing this I found a bug in the yaml package which doesn't dstinguish between nil and empty config so this might be able to be handled better if the yaml package changes its behaviour.
https://github.com/go-yaml/yaml/issues/417

Issue #4754, opened 2018-10-18
XSS vulnerable bootstrap version needs to be upgraded to 4.0.0


harche:
## Proposal
**Use case. Why is this important?**

Bootstrap version included is vulnerable for XSS attack, https://snyk.io/test/npm/bootstrap/3.3.1 

It needs to be upgraded to 4.0.0  

## Bug Report
**What did you do?**
Everything works fine but the bootstrap js lib included has XSS vulnerability so needs to be upgraded

**What did you expect to see?**
N/A

**What did you see instead? Under which circumstances?**
N/A

**Environment**
N/A

* System information:
N/A

* Prometheus version:
The XSS vulnerable bootstrap js lib exists even in master branch, https://github.com/prometheus/prometheus/tree/master/web/ui/static/vendor/bootstrap-3.3.1

* Alertmanager version:
N/A

* Prometheus configuration file:
```
N/A
```

* Alertmanager configuration file:
```
N/A
```

* Logs:
```
N/A
```

harche:
Duplicate of #3494?

ksherryBAE:
Does this issue still require fixing?

harche:
@ksherryBAE yes, very much. It's the known vulnerability that's getting shipped with Prometheus.

ksherryBAE:
Ok, I will have a look at it.

harche:
@ksherryBAE thanks.

harche:
@ksherryBAE thanks, that sounds awesome. Also, would it be possible to backport the fix to release [2.3.1](https://github.com/prometheus/prometheus/releases/tag/v2.3.1 )?

simonpasquier:
@harche FYI we may only backport patches to the current stable release.

harche:
@simonpasquier No problem. Thanks.

ksherryBAE:
Just to quickly check, glyphicons are no longer supported in bootstrap 4.0.0 however there is a work around utilising https://github.com/Darkseal/bootstrap4-glyphicons
Would it be suitable to use this? Or would something else be more appropriate?

achiuBAE:
Hi, i've also been contributing to this with ksherryBAE; i'll be making the pull request soon.

dbambro:
The following are bootstrap 3.3.1 XSS vulnerabilities:
CVE-2018-14042: In Bootstrap before 4.1.2, XSS is possible in the data-container property of tooltip.
[...]

PR #4764, opened 2018-10-20
Support to query elements by a specific time

kdada:
Implement #1246.

UI:
image

There are two options in URL query string:
- g[index].moment_input: date string. ex. 2018-10-20 10:11:12
- g[index].timestamp: unix timestamp. ex. 1540012198.141

If time and timestamp are provided, timestamp is choosen. After page loading, timestamp will be applied to moment_input and removed from URL.

Examples of URL:
```
/graph?g0.expr=go_memstats_alloc_bytes_total&g0.tab=1&g0.moment_input=2018-10-20 05:10:08
/graph?g0.expr=go_memstats_alloc_bytes_total&g0.tab=1&g0.timestamp=1540013881.663
```

Issue #3321, opened 2017-10-20
Autocomplete on /graph interface slow with large numbers of metricnames


jacksontj:
**What did you do?**
I have a cluster which returns ~150k names from `/api/v1/label/__name__/values`. This takes a decent amount of time for the browser to load, but then when you go to type it is slow to the point of unusable. For me at least (in both chrome and firefox) after I finish typing 2 characters it immediately pegs a CPU core for ~2m or so, rending the browser useless.

**What did you expect to see?**
I expect it to keep up :) A great example is from the grafana guys. They recently added ace editor support for promql (https://github.com/grafana/grafana/pull/9118) and that javascript typeahead doesn't have problems with my 150k names

brian-brazil:
Dupe of #2119

jacksontj:
This is actually not a duplicate, this is a separate issue.

The linked issue (#2119) is causing the browser to crash and requesting an addition of a limit to the API.

This issue is not showing as the browser crashing, but just that the javascript in the UI is slow. More interestingly (and as mentioned in the original issue description) the same number of metrics showing up in the response works fine performance-wise in grafana.

So, to re-iterate: this issue is client-side javascript slowness as the typeahead javascript library currently in use is horribly inefficient at handling large numbers of metrics (where there are many others-- including one example linked above) that handles it just fine.

alkar:
Experiencing the same, running v2.0.0.

Stono:
+1 to this, can we get it reopened?

juliusv:
Indeed, this is not a dupe of #2119 at all (that was about expression query results), reopening.

brian-brazil:
The discussion on #2119 is about autocomplete.

juliusv:
It was about limiting expression query results originally, but then it seems further down people started misunderstanding it and adding their thoughts about the metrics list and autocomplete. That seems like a separate issue though, despite it being similar.

PR #4821, opened 2018-11-05
web: Update from bootstrap 3.3.1 to bootstrap 4.0.0

achiuBAE:
implemented: #4754 

- Updated the html template files to be bootstrap 4.0.0 compatible, ensuring that all the functionality and ui is the same.
vendor: 
-reimplemented glyphicons from bootstrap 3 to work with 4.0.0. used: https://github.com/Darkseal/bootstrap4-glyphicons.
-jquery to 3.3.1. 
-bootstrap3-typeahead updated to work with bootstrap 4
css:
- minor css update to alerts.css, graph.css, prometheus.css. navbar and glyphicon button highlighting color inheritance. 
note: commits show up as nov 5th, as we didn't do commit signoff messages in initial fork. 

cc @juliusv


simonpasquier:
Thanks for looking into it! You've got conflicts.

web/ui/templates/_base.html
```diff
@@ -1,12 +1,12 @@
...
-        <title>Prometheus Time Series Collection and Processing Server</title>
-        <link rel="shortcut icon" href="{{ pathPrefix }}/static/img/favicon.ico?v={{ buildVersion }}">
-        <script src="{{ pathPrefix }}/static/vendor/js/jquery-3.3.1.min.js?v={{ buildVersion }}"></script>    
-        <script src="{{ pathPrefix }}/static/vendor/bootstrap-4.0.0/js/bootstrap.min.js?v={{ buildVersion }}"></script>
+  <head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+    <meta name="robots" content="noindex,nofollow">
+    <title>{{ pageTitle }}</title>
+    <link rel="shortcut icon" href="{{ pathPrefix }}/static/img/favicon.ico?v={{ buildVersion }}">
+    <script src="{{ pathPrefix }}/static/vendor/js/jquery.min.js?v={{ buildVersion }}"></script>
```
simonpasquier: Hmm, are you sure about this line?

web/ui/templates/_base.html
```diff
@@ -1,12 +1,12 @@
...
+    <script src="{{ pathPrefix }}/static/vendor/bootstrap-3.3.1/js/bootstrap.min.js?v={{ buildVersion }}"></script>
```
simonpasquier: Same remark here.

simonpasquier:
Thanks! I'll have a look next week.

web/ui/templates/_base.html
achiuBAE: ill take a look at the build problems next week, apologies. its trying to load some bootstrap.scss and also not loading the correct type-ahead file for some reason.

achiuBAE:
Thanks Simon. Hopefully the build passes now, i'll check in next week :)

simonpasquier:
I tried it quickly and it looks good albeit a few remarks. I've tested on Firefox (63.0.1 on Fedora).

* Somehow you're missing https://github.com/prometheus/prometheus/pull/4764.
* There's not enough margin because the top menu bar and the main header.
* Alert rows have no background color.
* The built-in console templates are broken by this change (check `http://<prometheus_URL>/consoles/prometheus.html` for instance).
* The button color is lighter, I don't have a strong opinion on this though so I'm fine keeping the new color.

image

image

mkumatag:
@simonpasquier @achiuBAE Any deadline for this task, if it gets merged then when can we expect this into a release?

simonpasquier:
@mkumatag Prometheus has adopted a time-based release schedule so there should a new Prometheus release every 6 weeks. You can check the schedule [here](https://github.com/prometheus/prometheus/blob/master/RELEASE.md#release-schedule).

@achiuBAE there are new conflicts.

ksherryBAE:
@simonpasquier All conflicts have been resolved, if there is still an issue with the margins clearing your cache should then ensure you're now viewing the latest css.

simonpasquier:
Thanks a lot for doing this! This looks very good. A few minor comments:

* `web/ui/static/vendor/bootstrap-4.1.3/css/.DS_Store` should be removed.
* I still don't see the UI change introduced by https://github.com/prometheus/prometheus/pull/4764.

We're currently in feature freeze waiting for the final v2.6.0 so in any case this PR will have to wait a bit before being merged.

cc @juliusv for feedback.

juliusv:
Couple of things I noticed when playing with this, before looking at code:

- The control buttons in the Graph view are no longer same height and style as their input elements, and no longer connected to them: 
buttons
- In the Graph view's typeahead autocomplete dropdown, you can no longer select elements with KeyUp and KeyDown.
- On the `/targets` page, there is very little spacing between the "All | Unhealthy" selector at the top and the first job group.

jritchieBAE:
> * In the Graph view's typeahead autocomplete dropdown, you can no longer select elements with KeyUp and KeyDown.

@juliusv Two of the three issues you mentioned have been resolved in our latest build. The issue above is due a change in how dropdowns are handled differently in bootstrap4, and there is not a straightforward fix for it. Would you be prepared to accept the build as it is to allow the update to bootstrap4, and maybe have this raised as a seperate new issue if it is something that wants to be added back in?

juliusv:
@jritchieBAE Thanks for the update and the fixes! Unfortunately the arrow key behavior is really fundamental though and not having it would be very frustrating, so I don't think we can afford it being broken. Are there maybe other dropdown/autocomplete libraries that would work with Bootstrap 4? The current one also has the problem that it gets really slow if you have many metric names, see https://github.com/prometheus/prometheus/issues/3321 and https://github.com/prometheus/prometheus/issues/5145. Maybe that would be an opportunity to improve that as well, although that is of course not required for just the Bootstrap update (it should just at least not get worse).

simonpasquier:
Hopefully the autocomplete slowness should be fixed by https://github.com/prometheus/prometheus/pull/5139 which is in good shape from my POV.

juliusv:
@jritchieBAE Just gave this an actual try and there's still some kinks here and there (spacing, styling), but I guess that stuff can be easily fixed. The missing autosuggest keyboard support is an actual problem of course. I'm currently also rebuilding the query browser page in React (with Bootstrap 4 already, so I can base it on this PR), so maybe I will find a dropdown component that works well. Or let me know if you do. I've tried out a couple already, but some are clearly not up for the task (slow with many items) or need a lot of custom work to behave like the one we had.

jritchieBAE:
@juliusv I managed to narrow the issue down. The behaviour of keyboard arrows in the typeahead was being interfered with by the jquery.hotkeys plugin. Disabling this plugin makes the typeahead work as desired. As far as I can tell there are actually no references in the code to the hotkeys file so no functionality is lost.

juliusv:
@jritchieBAE Great, the keyUp/keyDown works now, but I don't see the currently selected dropdown item being highlighted anymore. This makes it hard to see where in the list you are currently navigating. Do you think this is something that can be added back with this autosuggest picker?

jritchieBAE:
@juliusv I'll look into it, hopefully just a CSS change

juliusv:
@jritchieBAE Great. After that, could you rebase on the latest master and rebuild the assets (there's currently a conflict)? And would you be ok with re-opening this PR against a `bootstrap4` branch instead of `master` so I can do some last visual tweaks on it before it goes out?

juliusv:
@jritchieBAE With your latest changes (when rebuilding assets) I get the correct item highlighting now! There is still one odd behavior: if you select one of the items with the cursor keys and press Return to select it, the input field loses focus. This does not happen in current Prometheus, where you want to keep typing in the expression box after selecting a metric (it keeps the focus in the expression box).

jritchieBAE:
> @jritchieBAE With your latest changes (when rebuilding assets) I get the correct item highlighting now! There is still one odd behavior: if you select one of the items with the cursor keys and press Return to select it, the input field loses focus. This does not happen in current Prometheus, where you want to keep typing in the expression box after selecting a metric (it keeps the focus in the expression box).

@juliusv I have been unable to reproduce that issue - for me the input field retains focus. I have tested on multiple browsers.
Also, if you create an empty branch for me to make the new pull request into, I will do so :+1:

juliusv:
@jritchieBAE Hm weird, the loss of focus only happens the second time I select something from the dropdown. E.g.:

1. Go to empty /graph page.
2. Type `node_cpu`.
3. With cursor keys, select `node_cpu_seconds_total` in the dropdown -> focus remains
4. Erase the expression input.
5. Do the same thing again -> no focus on the input field anymore.

I created a new `bootstrap4` branch from `master` against which you should be able to reopen the PR.

jritchieBAE:
In light of the new pull request #5192 , could this pull request be closed?

juliusv:
Yes, thanks, closing here.

PR #4824, opened 2018-11-05
Add __meta_kubernetes_pod_phase label in discovery

### LINKED ISSUE #4823: Please add __meta_kubernetes_pod_status so 'completed' pods can be dropped from scrape
## Proposal

Pods that are created by Kubernetes Job objects will go to `Completed` state if they finish without error.
Without a way to stop scraping, the Prometheus `up` state goes to 0 and (in my current environment) this triggers an alert.

The Kubernetes SD inserts various labels such as `__meta_kubernetes_pod_annotation...` which allow relabel config to drop some targets.  I would like to extend this to a meta label which would hold the `Completed` status, so I can drop those pods from the targets.

bboreham:
This lets you add a relabel rule to drop scrapes for pods which are not running.

Fixes #4823


brian-brazil:
The docs also need updating.

bboreham:
> The docs also need updating.

Done.

brian-brazil:
Thanks!

Issue #1227, opened 2015-11-18
Sub query support for range selection.


fabxc:
Currently one can only do range selection on vector selectors and has to use recording rules to range-select expressions.

brian-brazil:
An interesting question is how we choose the interval for this, currently it's not needed as everything in calculated at a point. We should also look at how to discourage/prevent use of this in rules/alerts on performance grounds.

grobie:
:+1:

Andor:
Currently we can run query like `rate(hit[1m]) + rate(miss[1m])`, which is not very convenient.
IMHO better syntax is something like this: `rate(hit+miss [1m])` or `rate((hit+miss) [1m])`

brian-brazil:
That's incorrect usage, you need to do the rate separately first or you'll get the wrong answer around counter resets. I'd also suggest that the exporter do either total&hit or total&miss, as that's a bit easier to work with.

guoshimin:
Being able to run range queries on arbitrary expressions is valuable for ad-hoc explorations, either for sanity checking before making recorded rules, or for analyzing past data where you didn't already have recorded rules.

roman-vynar:
Will leave one of the references with discussion on this topic https://groups.google.com/forum/#!topic/prometheus-users/JOVfqQsVRl8

virusdave:
Is there any ticket tracking this feature request (historical function evaluation)?

brian-brazil:
This is the issue for it. I wouldn't be too hopeful of it ever getting implemented.

virusdave:
OK.  Is that due to a philosophical issue, or a eng-resource-scarcity problem?

brian-brazil:
It's a semantic issue, and an operational one. It's unclear how to do this correctly, and even then its use would have to be greatly limited for safety.

pparth:
I do not understand why this is not considered important at all. The requirement is simple: we want to generate a range-vector from a function that returns an instant-vector. For example i want to execute something like: 

sum_over_time(((sum(rate(server_request_hit{service_name=~"[[serviceName]]",http_response_status=~"5.."}[1m]))) / (sum(rate(server_request_hit{service_name=~"[[serviceName]]"}[1m]))) * vector(100)) > bool vector(0.1))

[...]

PR #4831, opened 2018-11-06
Subquery

codesome:
Fixes https://github.com/prometheus/prometheus/issues/1227

Based on this design doc: https://docs.google.com/document/d/1P_G87zN88YvmMr4iwLWygChMTZhai1L7S_c0awu1CAE/edit

**Checklist**
- [x] Lexer, with unit tests
- [x] Parser, with unit tests
- [x] Query engine
- [x] Unit tests for the working of query
- [ ] ~~Intensive real world tests~~


promql/parse.go
```diff
@@ -1125,6 +1161,12 @@ func (p *parser) checkType(node Node) (typ ValueType) {
 			p.errorf("unary expression only allowed on expressions of type scalar or instant vector, got %q", documentedType(t))
 		}
 
+	case *SubqueryExpr:
+		ty := p.checkType(n.Expr)
+		if ty != ValueTypeScalar && ty != ValueTypeVector {
+			p.errorf("expected scalar or vector selector for subquery %q, got %s instead", n.String(), ty)
```
brian-brazil: Scalars don't make sense for subqueries, it has to be an instant vector

roidelapluie: Why? We could do a subquery with scalar()

codesome:
@brian-brazil I see 2 potential places to execute subqueries
1. In `Engine.populateSeries(...)` we execute subquery and populate the series.
2. In `evaluator.eval(...)` we execute the expression of subquery using another evaluator and return it.

Am I missing something here? Which one is preferred? (My choice is 1).

brian-brazil:
It should be in eval I think.

codesome:
While trying to device a solution in `eval`, I see 1 complication. 

Consider this expression
```
rate(baz[1m]) +  rate( ( rate(foo[5m]) + rate(bar[10m]) )[15m:5s] )
```

As we are populating all the vector and matrix selectors beforehand, we would ideally like to load up to `25m` back in time (`15m` of subquery and `10m` of matrix selector inside it).

But as the `maxOffset` will be taken same for the entire expression, loading up to `25m` back in time for `baz` is unnecessary, as it needs only up to `1m`.

As there can be more nesting inside subqueries, this this added overhead fine?

brian-brazil:
> As there can be more nesting inside subqueries, this this added overhead fine?

This is not a problem for local storage.

codesome:
@brian-brazil I am stuck at choosing default step for expressions of the form `expr[5m:]`, where step is optional.

For range queries I am able to take the interval given with it. But I am not sure what is to be done in the case of an instant query.

brian-brazil:
> For range queries I am able to take the interval given with it. But I am not sure what is to be done in the case of an instant query.

In both cases, use the global evaluation interval.

codesome:
I was unable to find out where the global evaluation interval is 

I believe it is _not_ this https://github.com/prometheus/prometheus/blob/b1e779d0edca9c7b3db29a8b73345a94597108c6/config/config.go#L268

brian-brazil:
That is it.

codesome:
@brian-brazil I think this is good for a review now. I will be adding tests in the meanwhile.

cmd/prometheus/main.go
```diff
@@ -655,6 +656,7 @@ func reloadConfig(filename string, logger log.Logger, rls ...func(*config.Config
 	if failed {
 		return fmt.Errorf("one or more errors occurred while applying the new configuration (--config.file=%q)", filename)
 	}
+	promql.GlobalEvaluationInterval = time.Duration(conf.GlobalConfig.EvaluationInterval)
```
brian-brazil: Does this need locking?

promql/engine.go
```diff
@@ -863,6 +900,15 @@ func (ev *evaluator) eval(expr Expr) Value {
 				matrixArg = true
 				break
 			}
+			// SubqueryExpr can be used in place of MatrixSelector.
+			subq, ok := a.(*SubqueryExpr)
+			if ok {
```
brian-brazil: You can combine this into the above line

promql/engine.go
```diff
@@ -1048,6 +1094,25 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+			maxSamples:     ev.maxSamples,
+			logger:         ev.logger,
+		}
+		if ev.startTimestamp < newEv.startTimestamp {
+			newEv.startTimestamp = ev.startTimestamp
+		}
+		if e.StepExists {
+			newEv.interval = int64(e.Step / time.Millisecond)
+		} else if newEv.interval == 0 {
+			newEv.interval = int64(time.Duration(config.DefaultGlobalConfig.EvaluationInterval) / time.Millisecond)
```
brian-brazil: This can never happen, it's already set above

promql/engine.go
```diff
@@ -1048,6 +1094,25 @@ func (ev *evaluator) eval(expr Expr) Value {
 			panic(fmt.Errorf("cannot do range evaluation of matrix selector"))
 		}
 		return ev.matrixSelector(e)
+
+	case *SubqueryExpr:
+		newEv := &evaluator{
+			startTimestamp: ev.endTimestamp - int64(e.Range/time.Millisecond),
```
brian-brazil: startTimestamp?

This should also be aligned with the step

promql/lex_test.go
```diff
@@ -429,6 +429,120 @@ var tests = []struct {
...
+			{itemNEQRegex, 12, `!~`},
+			{itemString, 14, `"b:ar"`},
+			{itemRightBrace, 20, `}`},
+			{itemLeftBracket, 21, `[`},
+			{itemDuration, 22, `4m`},
+			{itemColon, 24, `:`},
+			{itemRightBracket, 25, `]`},
+		},
+	}, { // Nested Subquery.
+		input: `min_over_time(sum(rate(testname{foo="bar"}[5m])[4m:4s])[5m:5s])`,
```
brian-brazil: The nested subquery isn't needed here, do you have a better example?

promql/parse.go
```diff
@@ -494,13 +494,17 @@ func (p *parser) unaryExpr() Expr {
 	return e
 }
 
-// rangeSelector parses a Matrix (a.k.a. range) selector based on a given
-// Vector selector.
+// subqueryOrRangeSelector parses a Subquery based on given Expr (or)
+// a Matrix (a.k.a. range) selector based on a given Vector selector.
 //
-//		<Vector_selector> '[' <duration> ']'
+//		<Vector_selector> '[' <duration> ']' | (<Vector_selector>|<Scalar_expr>) '[' <duration> ':' [<duration>] ']'
```
brian-brazil: You can't have scalars

promql/testdata/subquery.test
```diff
@@ -0,0 +1,182 @@
+## The following tests are from selectors.test
```
brian-brazil: It'd be best to focus on aspects of the implementation, rather than things that test the internals of other functionality.

For example, how does offset work with this?

promql/lex_test.go
codesome: Are we not planning to support nested subquery? (I haven't used promql much, so I am unaware of use cases)

promql/testdata/subquery.test
codesome: I had not added support for offset for subquery. Should I add it?

promql/engine.go
codesome: If I am understanding this correctly, `startTimestamp = endTimestamp - (some_int * interval)`? 

If so, should `startTimestamp` be before or after `(end - Range)`.

brian-brazil: Remember that this is a range query evaluation, so you need data from before the startTimestamp.

In addition the startTimestamp should be aligned with the step of the subquery, so the query/query_range invairant isn't broken.

codesome: I have made some changes to align it with the step. Is it right, or am I still missing something?

promql/engine.go
```diff
@@ -58,9 +57,23 @@ var (
...
+// SetGlobalEvaluationInterval sets GlobalEvaluationInterval.
+func SetGlobalEvaluationInterval(ev time.Duration) {
+	globalEvaluationIntervalMtx.Lock()
+	defer globalEvaluationIntervalMtx.Unlock()
+	GlobalEvaluationInterval = ev
+}
+
+func getGlobalEvaluationInterval() time.Duration {
+	globalEvaluationIntervalMtx.RLock()
+	defer globalEvaluationIntervalMtx.RUnlock()
```
brian-brazil: This allocates memory, you'd be better to do the locking manually.

promql/engine.go
```diff
@@ -1048,6 +1105,33 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+			maxSamples:   ev.maxSamples,
+			logger:       ev.logger,
+		}
+		if e.StepExists {
+			newEv.interval = int64(e.Step / time.Millisecond)
+		}
+
+		// Calculate number of steps of the subquery evaluation.
+		rangeMillis := (int64(e.Range / time.Millisecond))
+		numSteps := (rangeMillis + (newEv.interval / 2)) / newEv.interval
```
brian-brazil: Why are you adding in half?

promql/engine.go
```diff
@@ -1048,6 +1105,33 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+		if ev.startTimestamp < (newEv.endTimestamp - rangeMillis) {
+			// If the original evaluator has startTimestamp before subquery startTimestamp,
+			// have starting step on/before that.
+			numSteps = (newEv.endTimestamp - ev.startTimestamp + (newEv.interval / 2)) / newEv.interval
+		}
+
+		// Align the startTimestamp with the subquery step.
+		correctedRange := newEv.interval * numSteps
+		newEv.startTimestamp = newEv.endTimestamp - correctedRange
```
brian-brazil: This is aligning relative to the end time, not the step

promql/engine.go
```diff
@@ -1048,6 +1105,33 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+			logger:       ev.logger,
+		}
+		if e.StepExists {
+			newEv.interval = int64(e.Step / time.Millisecond)
+		}
+
+		// Calculate number of steps of the subquery evaluation.
+		rangeMillis := (int64(e.Range / time.Millisecond))
+		numSteps := (rangeMillis + (newEv.interval / 2)) / newEv.interval
+		if ev.startTimestamp < (newEv.endTimestamp - rangeMillis) {
```
brian-brazil: This can never happen

promql/engine.go
codesome: I am a little confused about the step now. What I understood was, the start time should be integral steps away from end time.

What do we mean by aligning with the step?

brian-brazil: If the step is say 5s, then the evals should be at 00:00:00, 00:00:05, 00:00:10 etc. irrelevant of what the start/end time are for the query.

promql/engine.go
codesome: What I understood was, the start time should be integral steps away from end time. And that it should either overlap with the (end-range), or should be the one just behind it if it doesn't exactly overlap. So adding that half takes case of just behind it.

But oh well, I just realised adding that half won't take care of it, I guess I mixed something here. I will fix it. But is my above explanation right?

promql/engine.go
codesome: Subquery starts at time 0 always?

codesome: Does it mean 
```
startTimestamp = 0
endTimestamp = numSteps * interval
```

brian-brazil: The start time will be the first aligned step at or after the start time.

codesome: What is aligned step here? Also, shouldn't it be like the end time of subquery is the first aligned step on or before the end time, and then we take the steps back?

brian-brazil: The evaluator works from the start to the end, so I'd expect it to work from the start. We can then leave the endTimestamp as-is.

codesome:
I think I am doing something wrong here. I was taking the ranges of the evaluator and using that to evaluate the subquery. But when I study the code more, I see that the subquery would be called at a particular timestamp, and I should not use the start and end time of the evaluator directly.

So if I am not wrong, I should be catching that particular timestamp for which the subquery is called and evaluate based on it. Am I correct?

brian-brazil:
That'd be inefficient, what you want to do is generate the entire range vector at once. This way points used in more than one step can be reused, rather than recalculated.

codesome:
With your comment and some more digging into code, now things look a bit more clear to me. I did not realise that I had to load the entire matrix at once.

promql/engine.go
codesome: Ignore my previous comment. Now things are more clear. I did not realise that we are loading the entire matrix at once.

But I don't think I still understand what do you mean by aligned step. With my current understanding, the start time should be `ev.startTimestamp - offset - range`.

brian-brazil: The problem with that is that it'd break the invariant that query range is just syntactic sugar over query, such as if the query range step and subquery step didn't match up.

By aligning the subquery with it's step, independent of the query/query range settings, we avoid that.

codesome: Sorry for so much confusion, what is `it` in "subquery with it's step"?

Can you give me an example of the original step (w.r.t to the variables in the code / query) with whom we want to align the subquery's start time.

brian-brazil: The subquery's step, `newEv.interval`

codesome:
@brian-brazil is the start time aligned correctly now?
Also does this unit test look right https://github.com/prometheus/prometheus/blob/683545dd1414b2624d45bcb2054ceac48b7524b2/promql/engine_test.go#L697-L707

brian-brazil:
The code doesn't look right. Does the test get the same result if the start time is different? It's meant to.

codesome:
No it doesn't give the same result. In that case, w.r.t. which start time do we need to align the `newEv.startTimestamp`?

brian-brazil:
newEv.startTimestamp itself.

codesome:
`newEv.startTimestamp` will be 0 initially. 

So does it mean that, if we run the subquery at any point of time - say evaluation sometime in the future, sometime in the past,  all evaluations in a query range - all of them would return the same result?

Shouldn't it be somewhat connected to when we are querying?

brian-brazil:
Yes, it'd be restricted to the time range we actually need.

codesome:
> How will this interact with sample counting and sample slice pooling?

With the latest commit, the sample counting will be carried forward from the parent evaluator, i.e. parent query + subquery will be counted in total.

And about sample slice pooling, looks like subquery will re-use the pool of the parent query.

PS: @brian-brazil what's you opinion on above comment on start time?

brian-brazil:
> should I be keeping newEv.startTimestamp as the first aligned step before (ev.startTimestamp - offset - range)?

The first after is enough. The first before will not be needed.

codesome:
@brian-brazil PTAL

promql/engine.go
```diff
@@ -1051,11 +1103,43 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+
+		if e.StepExists {
+			newEv.interval = durationToInt64Millis(e.Step)
+		} else {
+			globalEvaluationIntervalMtx.RLock()
+			newEv.interval = durationToInt64Millis(GlobalEvaluationInterval)
+			globalEvaluationIntervalMtx.RUnlock()
+		}
+
+		newEv.startTimestamp = newEv.interval * ((ev.startTimestamp - offsetMillis - rangeMillis) / newEv.interval)
```
gouthamve: Can you add a comment what is happening here and why we are doing this?

promql/lex.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2015 The Prometheus Authors
+// Copyright 2018 The Prometheus Authors
```
gouthamve: This date is about when the file is added and not when the file is changed.

promql/lex_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2015 The Prometheus Authors
+// Copyright 2018 The Prometheus Authors
```
gouthamve: Same here

promql/engine.go
```diff
@@ -477,23 +494,36 @@ func (ng *Engine) execEvalStmt(ctx context.Context, query *query, s *EvalStmt) (
 	return mat, nil
 }
 
+// cumulativeSubqueryOffset returns the sum of range of all subqueries in the path.
+func (ng *Engine) cumulativeSubqueryOffset(path []Node) time.Duration {
```
gouthamve: We're not adding the offsets of each subquery here but should be right?

cmd/prometheus/main.go
```diff
@@ -233,6 +233,7 @@ func main() {
 	}
 
 	promql.LookbackDelta = time.Duration(cfg.lookbackDelta)
+	promql.SetGlobalEvaluationInterval(time.Duration(config.DefaultGlobalConfig.EvaluationInterval))
```
juliusv: Can we avoid a global variable here (well, you always can, just a question of how much work)? AFAICS it would mainly need to be passed into all places that create an engine, and from the engine into the evaluator?

Btw. in the `promql` package we should not call this `GlobalEvaluationInterval`, since that is a rule-related concept that has nothing to do with PromQL per se. How about `DefaultEvaluationInterval` instead? That's all that PromQL needs to know.

promql/ast.go
```diff
@@ -115,6 +115,15 @@ type MatrixSelector struct {
 	series []storage.Series
 }
 
+// SubqueryExpr represents a subquery.
+type SubqueryExpr struct {
+	Expr       Expr
+	Range      time.Duration
+	Offset     time.Duration
+	Step       time.Duration
+	StepExists bool
```
juliusv: Don't feel strongly about this one, but I would have probably just used the zero value of `Step` to signal a non-specified step, unless there's a problem with that that I'm overlooking. It's pretty common to check if something is 0 and if yes, set it to some default.

promql/engine.go
```diff
@@ -1051,11 +1103,51 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+
+		if e.StepExists {
+			newEv.interval = durationToInt64Millis(e.Step)
+		} else {
+			globalEvaluationIntervalMtx.RLock()
+			newEv.interval = durationToInt64Millis(GlobalEvaluationInterval)
+			globalEvaluationIntervalMtx.RUnlock()
+		}
+
+		// We want to align the start time of the subquery with it's step.
```
juliusv: nit: "it's" -> "its"

cmd/prometheus/main.go
codesome: As it is passed down through Engine, it would need passing from all the places where we create Engine - that might need access to the global config. I would say it's better if we set it once, and it is used thereafter. If there is any benefit of passing down through Engine, I can do that.

I will change the name.

promql/value.go
```diff
@@ -255,3 +255,63 @@ func (r *Result) String() string {
...
+}
+
+func (ssi *storageSeriesIterator) At() (t int64, v float64) {
+	p := ssi.points[ssi.curr]
+	return p.T, p.V
+}
+
+func (ssi *storageSeriesIterator) Next() bool {
+	ssi.curr++
+	if ssi.curr == len(ssi.points) {
```
brian-brazil: If next is called repeatedly, this will start to return true again. >= would be safer.

codesome:
I have added some tests for nested subquery and for the points at the boundaries. Also for offset.

The more I think of tests, it is boiling down to using it in the same way as in the current tests, but with different function calls. Anything more that I need to cover?

promql/engine.go
```diff
@@ -1077,11 +1129,50 @@ func (ev *evaluator) eval(expr Expr) Value {
...
+		}
+
+		// We want to align the start time of the subquery with its step.
+		// Aligned steps are `0 1*step 2*step 3*step ...`
+		// We choose start time as the first aligned step just after (or equal to)
+		// 'S = (ev.startTimestamp - offset - range)'
+
+		// numSteps = S / step
+		// newEv.startTimestamp = step * numSteps
+		newEv.startTimestamp = newEv.interval * ((ev.startTimestamp - offsetMillis - rangeMillis) / newEv.interval)
```
free: You can get rid of the `if` below if you write this as:
```suggestion
		newEv.startTimestamp = ((ev.startTimestamp - offsetMillis - rangeMillis - 1) / newEv.interval + 1) * newEv.interval
```

I would also replace all the comments on the few lines above with "Start from the first timestamp at or after \`ev.startTimestamp - offset - range\` that is a multiple of \`interval\`.

promql/engine.go
```diff
@@ -839,6 +869,21 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 	return mat
 }
 
+// subqIntoMatrixSelector evaluates given SubqueryExpr and returns an equivalent
+// evaluated MatrixSelector in its place. Note that the Name and LabelMatchers are not set.
+func (ev *evaluator) subqIntoMatrixSelector(subq *SubqueryExpr) *MatrixSelector {
```
free: I think a better name would be `evalSubquery`. From the current name it is not obvious that you are actually evaluating the subquery, and that is arguably the most important part of what you're doing here, not merely mapping the range and offset into an equivalent `MatrixSelector`.

promql/value.go
```diff
@@ -255,3 +255,63 @@ func (r *Result) String() string {
...
+
+func newStorageSeriesIterator(series Series) *storageSeriesIterator {
+	return &storageSeriesIterator{
+		points: series.Points,
+		curr:   -1,
+	}
+}
+
+func (ssi *storageSeriesIterator) Seek(t int64) bool {
+	for i := ssi.curr; i < len(ssi.points); i++ {
```
free: This will blow up in your face if you call `Seek` before having called `Next`.

promql/engine.go
```diff
@@ -51,6 +51,23 @@ const (
...
 )
 
+var (
+	// LookbackDelta determines the time since the last sample after which a time
+	// series is considered stale.
+	LookbackDelta = 5 * time.Minute
+
+	// DefaultEvaluationInterval is the value of EvaluationInterval in the GlobalConfig.
+	DefaultEvaluationInterval    time.Duration
+	defaultEvaluationIntervalMtx sync.RWMutex
```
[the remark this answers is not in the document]
codesome: Makes sense. Atomic + pass it down the evaluator.

promql/engine.go
codesome: I was testing this, and noticed that when the start time goes negative because of offset and range, this equation wont hold good.

codesome:
@brian-brazil I have added tests for edge data and offsets. The tests for queries that would return a matrix are in Go, rest all are in testdata directory.

What more do I need to add?

promql/parse.go
```diff
@@ -1000,6 +1042,12 @@ func (p *parser) checkType(node Node) (typ ValueType) {
 			p.errorf("unary expression only allowed on expressions of type scalar or instant vector, got %q", documentedType(t))
 		}
 
+	case *SubqueryExpr:
+		ty := p.checkType(n.Expr)
+		if ty != ValueTypeVector {
+			p.errorf("expected vector selector for subquery %q, got %s instead", n.String(), ty)
```
brian-brazil: It needs to be an instant vector, not only a vector selector

codesome: Yes, updated the error message.

brian-brazil:
You've got a test failure.

brian-brazil:
The docs also need updating.

brian-brazil:
Yeah, I think that makes most sense. Just before Operators.

docs/querying/basics.md
```diff
@@ -169,6 +169,20 @@ The same works for range vectors. This returns the 5-minutes rate that
...
+
+Syntax: `<instant_query> '[' <range> ':' [<resolution>] ']' [ offset <duration> ]`
+
+* `<resolution>` is optional. Default is the global evaluation interval.
+
+Examples
+
+    1. `rate(http_requests_total[5m])[30m:1m]` - This returns 5-minute rate of `http_requests_total` metric for the past 30 minutes, at a resolution of 1 minute.
+
+    2. `rate(sum_over_time(http_requests_total[30s:10s])[50s:])` - This is an example of nested subquery and using default resolution for outer subquery.
```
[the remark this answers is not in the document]
brian-brazil: Yes, that doesn't make sense. A `max_over_time(rate(...))` would be a good example.

docs/querying/basics.md
```diff
@@ -169,6 +169,20 @@ The same works for range vectors. This returns the 5-minutes rate that
 
     rate(http_requests_total[5m] offset 1w)
 
+## Subquery
+
+Subquery allows you to run an instant query for a given range and resolution. Like how range vector selector is for vector selector, subquery is for an instant query. The result of a subquery is a matrix.
```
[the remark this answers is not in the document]
brian-brazil: Yip, matrix is only visible to users if they look at HTTP responses these days.

brian-brazil: This comparison to range vector selectors is confusing. It'd be instant vector expressions.

docs/querying/basics.md
```diff
@@ -169,6 +169,20 @@ The same works for range vectors. This returns the 5-minutes rate that
...
 
+## Subquery
+
+Subquery allows you to run an instant query for a given range and resolution. Like how range vector selector is for vector selector, subquery is for an instant query. The result of a subquery is a matrix. 
+
+Syntax: `<instant_query> '[' <range> ':' [<resolution>] ']' [ offset <duration> ]`
+
+* `<resolution>` is optional. Default is the global evaluation interval.
+
+Examples
```
brian-brazil: There's a separate examples page

docs/querying/examples.md
```diff
@@ -38,6 +38,16 @@ To select all HTTP status codes except 4xx ones, you could run:
...
 
+## Subquery
+
+This query returns 5-minute rate of `http_requests_total` metric for the past 30 minutes, at a resolution of 1 minute.
+
+    rate(http_requests_total[5m])[30m:1m]
+
+This is an example of nested subquery. The subquery for the rate uses default resolution. Note that using subqueries unnecessarily is unwise.
+
+    max_over_time(rate(http_requests_total[30s:5s])[10m:])
```
brian-brazil: This isn't a good example, you should use a range vector selector with rate.

codesome: I thought so, will update it.

codesome: `max_over_time(rate(rate(distance_covered_total[5s])[30s:5s])[10m:])` - does this look sane? Looks like max acceleration over time.

brian-brazil: A rate of a rate doesn't make sense, as rate can only be applied to a counter and the output of rate is a gauge. A deriv of a rate would make sense though.

codesome: So does `max_over_time(deriv(rate(distance_covered_total[5s])[30s:5s])[10m:])` look fine for the doc?

brian-brazil: Yes, that's fine.

brian-brazil:
Thanks!

juliusv:
Wow, congrats on getting this done!

roidelapluie:
this is huge!! congratz!!

codesome:
Thanks @brian-brazil for all the patient reviews!

PR #4832, opened 2018-11-06
Returns available data when doing remote read and warnings if remote read fails.

mknapphrt:
This is a remake of: https://github.com/prometheus/prometheus/pull/4426
Reasonably significant changes were made since that PR and I feel as though the comments and whatnot aren't very relevant anymore, so I remade it. 

The purpose of this PR is to fix an issue with remote read where if any remote read source failed, then the entire query would fail. With this change, any query errors that occur that aren't from the primary data source (the local tsdb) are instead returned as a list of warnings and the query continues. In addition, if a query fails midway through, no data from that data source is returned, keeping the all-or-nothing approach used in most cases.

On thing that I still think should be done is displaying the warnings on the web UI in the same manner that errors are displayed. But I'd like to get this up for review while I work on this. @brian-brazil has given me a lot of feedback on the other PR and I'm hoping this solution is a better approach.


brian-brazil:
> On thing that I still think should be done is displaying the warnings on the web UI in the same manner that errors are displayed.

That would be good, but it doesn't have to happen all in the one PR.

web/api/v1/api.go
```diff
@@ -1043,11 +1053,19 @@ func mergeLabels(primary, secondary []*prompb.Label) []*prompb.Label {
 	return result
 }
 
-func (api *API) respond(w http.ResponseWriter, data interface{}) {
+func (api *API) respond(w http.ResponseWriter, data interface{}, warnings storage.Warnings) {
+	statusMessage := statusSuccess
+	var warningStrings []string
+	if warnings != nil && len(warnings) != 0 {
```
brian-brazil: len(warnings) != 0  is sufficient

storage/fanout.go
```diff
@@ -241,6 +261,15 @@ func (q *mergeQuerier) LabelValues(name string) ([]string, error) {
 	return mergeStringSlices(results), nil
 }
 
+func (q *mergeQuerier) IsValidSet(set SeriesSet) bool {
+	for _, failedQuerier := range q.failedQueriers {
```
brian-brazil: Should failedQueriers be a map?

storage/fanout.go
```diff
@@ -311,34 +342,55 @@ func NewMergeSeriesSet(sets []SeriesSet) SeriesSet {
...
+	// failed remote storage sources, we want to keep trying with the next label set.
+	for {
+		// Firstly advance all the current series sets.  If any of them have run out
+		// we can drop them, otherwise they should be inserted back into the heap.
+		for _, set := range c.currentSets {
+			if set.Next() {
+				heap.Push(&c.heap, set)
+			}
+		}
+		if c.heap == nil || len(c.heap) == 0 {
```
brian-brazil: len(c.heap) == 0 is sufficient

storage/fanout.go
```diff
@@ -241,6 +261,15 @@ func (q *mergeQuerier) LabelValues(name string) ([]string, error) {
 	return mergeStringSlices(results), nil
 }
 
+func (q *mergeQuerier) IsValidSet(set SeriesSet) bool {
```
brian-brazil: IsFailedSet would be clearer

promql/engine.go
```diff
@@ -884,6 +880,14 @@ func (ev *evaluator) eval(expr Expr) Value {
 		}
 
 		sel := e.Args[matrixArgIndex].(*MatrixSelector)
+		if sel.series == nil {
```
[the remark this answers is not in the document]
mknapphrt: So I had to change this around from what I originally did because it didn't quite work. Originally I just had the unexpanded series set stored in the selector and expanded it when I needed to use it. But this became an issue when doing a range query because it tries to expand the series set multiple times, which is a problem because once the series is expanded once it "uses up" the iterator. I don't know how to phrase that better, but the second time it tries to expand it the iterator has already gone through to the end. So instead, after I expand it the first time, I store it in sel.series and reuse that value. So it should be nil the first time through but only the first time through.

web/api/v1/api.go
```diff
@@ -1056,7 +1056,7 @@ func mergeLabels(primary, secondary []*prompb.Label) []*prompb.Label {
 func (api *API) respond(w http.ResponseWriter, data interface{}, warnings storage.Warnings) {
 	statusMessage := statusSuccess
 	var warningStrings []string
-	if warnings != nil && len(warnings) != 0 {
+	if len(warnings) != 0 {
```
brian-brazil: You don't need this if, range handles nil gracefully

storage/fanout.go
```diff
@@ -261,13 +263,12 @@ func (q *mergeQuerier) LabelValues(name string) ([]string, error) {
...
 }
 
-func (q *mergeQuerier) IsValidSet(set SeriesSet) bool {
-	for _, failedQuerier := range q.failedQueriers {
-		if q.setQuerierMap[set] == failedQuerier {
-			return false
-		}
+func (q *mergeQuerier) IsFailedSet(set SeriesSet) bool {
+	_, isFailedQuerier := q.failedQueriers[q.setQuerierMap[set]]
+	if isFailedQuerier {
```
brian-brazil: return isFailedQuerier

promql/engine.go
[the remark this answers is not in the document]
brian-brazil: Okay. This is repeated a few times, it should probably be factored out

mknapphrt:
@brian-brazil I'm seeing now that there's some new merge conflicts, but I was wondering if you had a chance to look over this again? Thanks

promql/engine.go
```diff
@@ -544,20 +543,36 @@ func (ng *Engine) populateSeries(ctx context.Context, q storage.Queryable, s *Ev
...
-				level.Error(ng.logger).Log("msg", "error expanding series set", "err", err)
-				return err
+			n.unexpandedSeriesSet = set
+		}
+		return nil
+	})
+	Inspect(s.Expr, func(node Node, path []Node) error {
+		switch n := node.(type) {
+		case *VectorSelector:
+			if n.unexpandedSeriesSet != nil {
```
brian-brazil: In what circumstances is this nil?

It could also be combined with the above, or you could go back to doing it at eval time (which is a bit more memory efficient).

mknapphrt: `n.unexpandedSeriesSet` would be nil any time that `querier.Select` returns an error. I put it here because we wanted to factor out the expansion. Unless I misunderstood, did you just mean to take that portion of code put it in a function instead of repeating it a few times?

brian-brazil: If querier.Select returns an error, the query has already failed.

> did you just mean to take that portion of code put it in a function instead of repeating it a few times?

Yes

mknapphrt:
Apologies, I didn't realize you were talking about literal "TODO" comments in the code, I thought you meant stuff that you had requested. I'm not sure what the best way to handle warnings in the federation is. Right now there's a metric for the number of errors occurring, would it be appropriate to do the same for the number of warnings?

brian-brazil:
Log it at debug level and bumping a metric sounds good to me. Really federation should only be using local data, so long term it won't matter.

mknapphrt:
TODO's are now TODONE and the merge conflicts are resolved @brian-brazil

brian-brazil:
You've got a conflict still.

mknapphrt:
That must have come up over the weekend, resolved now. @brian-brazil

brian-brazil:
One last thing, the API docs need updating for the new field.

mknapphrt:
I wasn't sure how in depth you wanted the description, let me know what you think @brian-brazil

docs/querying/api.md
```diff
@@ -24,6 +24,10 @@ and one of the following HTTP response codes:
 Other non-`2xx` codes may be returned for errors occurring before the API
 endpoint is reached.
 
+An array of warnings will be returned if there is an error while querying
+remote storage sources. All data from those sources are excluded, but data
```
brian-brazil: Let's not tie this to remote storage, this part of the docs is the general response format and we may have other uses for this in future (e.g. ignoring corrupted local chunks).

mknapphrt:
@brian-brazil I fixed a merge conflict and changed the description in the docs, what do you think now?

docs/querying/api.md
```diff
@@ -24,6 +24,10 @@ and one of the following HTTP response codes:
 Other non-`2xx` codes may be returned for errors occurring before the API
 endpoint is reached.
 
+An array of warnings may be returned if there are errors that do
+not inhibit the query execution. All of the data that was successfully
```
brian-brazil: s/query/request/.

Not all queries are requests.

juliusv:
It would be good to document this, I added a docs issue for that: https://github.com/prometheus/docs/issues/1243

Issue #2038, opened 2016-09-28
Implement HTTP API for getting all label names


juliusv:
Would be relevant for https://github.com/grafana/grafana/issues/6038 (cc @bergquist).

Endpoint would be:

`GET /api/v1/labels`

Output would be:

``` json
{
   "status" : "success",
   "data" : [
      "__name__",
      "job",
      "instance",
      "method",
      "status",
      "path"
   ]
}
```

This isn't a trivial change, as it requires plumbing all the way down to the storage layer and the LevelDB index accessor functions.

opsnull:
any updating ?

brancz:
I don't think we'll tackle this before 2.0 as LevelDB is going away completely for that and this feature would have to be rewritten entirely. @fabxc what is your stance on getting something like this into the 2.0 dev branch? However I wouldn't expect this to be too difficult with the 2.0 storage given that we are already keeping indices of labels and their values, right @fabxc ?

fabxc:
Yea, that should be reasonably easy to do. I also want to investigate composite indices against metric names then, i.e. "what are all possible values of label `instance` for metric `requests_total`".

juliusv:
@fabxc @Gouthamve This just came up again in a discussion with @davkal (it could be useful for the UIs he's building). Do you have any more insight into how to do this efficiently on top of the new storage? To do it properly with minimal overhead, it'd probably need a new TSDB interface function?

Gouthamve:
So the `tsdb.Querier` already exposes it: https://github.com/prometheus/tsdb/blob/master/querier.go#L34-L35

We need to add it to the `Storage` interface I guess.

shubheksha:
@Gouthamve, that returns all the values, not names. A new function will be needed. See https://github.com/prometheus/tsdb/issues/279

krasi-georgiev:
the work on tsdb side is done so removed the `local storage` label.

deeTEEcee:
Is there a way to get the list of unique label names based on the metric?

PR #4835, opened 2018-11-07
/api/v1/labels endpoint for getting all label names

codesome:
Closes https://github.com/prometheus/prometheus/issues/2038

See https://github.com/prometheus/tsdb/pull/369


brancz:
I think we said we wanted the returned labels to be filter-able by a selector, no? (we said on the tsdb PR that it's not worth having in tsdb, but here it still makes sense functionality wise)

codesome:
The added complexity wont add much value.

As said in this comment https://github.com/prometheus/tsdb/pull/369#issuecomment-430547460
> How many unique label names are we really looking at – maybe 1000? The caller can just loop over them and filter with whatever they want.

It perfectly makes sense, label names are not large in number. In future if we do want to add matchers, it wont break this API call.

brancz:
Right, but now we're not talking about the tsdb interface, but the Prometheus API, and I'd argue that for proxies hiding data, or user interfaces this would make sense on this level.

codesome:
Okay, now I get it. So add the same matchers here which was dropped in tsdb?

codesome:
Added.
```
$ curl 'localhost:9090/api/v1/labels'
{
  "status": "success",
  "data": [
    "__name__",
    "call",
    "code",
    "config",
    "dialer_name",
    "endpoint",
    "event",
    "goversion",
    "handler",
    "instance",
    "interval",
    "job",
    "le",
    "listener_name",
    "name",
    "quantile",
    "reason",
    "role",
    "scrape_job",
    "slice",
    "version"
  ]
}
```
```
$ curl 'localhost:9090/api/v1/labels?match[]=call|le|.*name'
{
  "status": "success",
  "data": [
    "call",
    "dialer_name",
    "le",
    "listener_name",
    "name"
  ]
}
```
```
$ curl 'localhost:9090/api/v1/labels?match[]=.*name&match[]=dial.*'
{
  "status": "success",
  "data": [
    "dialer_name"
  ]
}
```

brancz:
Actually @gouthamve and I were just talking and he made a good point. Now we just have the string list returned from TSDB, whereas the point for auto-complete or something was that we could limit based on existing selections. That type of selection is just not possible at this stage anymore. Can we maybe rethink that decision for TSDB again? /cc @fabxc

codesome:
Yes. Would that need indexing for composite label names?

gouthamve:
Well, I think we have the series API there. Ideally, it would be great if we could add some matchers to it and it returns only those labels which are in the series that match them.

But right now, it'll just mean querying TSDB, getting the series and then getting the label values. Which is basically the series API with server side filtering. I think for now, using the series API and filtering client side is better.

Further, it won't be breaking if we add matcher support later to this API and this is not super required, as we'll want to auto-complete the label values also and we'll be calling the series API anyways for that.

gouthamve:
I'm saying **drop the matchers**. And I don't think there is any point in having regex filters either. I'd rather start with a super clean API with nothing, and then add matchers/filters based on the feature requests.

Also given it's just going to be small list, I'd rather the client side do it's regex-filtering anyways.

And in my previous comment, I was just saying right now, if we were to implement matchers, it would be like hitting the series endpoint and doing label-name dedupe server side, and this is something the client can do.

codesome:
And it has been removed.

codesome:
@gouthamve Done.

web/api/v1/api.go
```diff
@@ -373,6 +375,20 @@ func (api *API) queryRange(r *http.Request) (interface{}, *apiError, func()) {
 	}, nil, qry.Close
 }
 
+func (api *API) labelNames(r *http.Request) (interface{}, *apiError, func()) {
```
gouthamve: Colocate this with `labelValues`?

`returnApiError` is used in `queryRange` above and is best near that. Super minor nit though.

codesome: Done.

bwplotka:
I think some matching here will end up being neccesary to limit cardinality for example. (: 

so as @gouthamve said:
> start with a super clean API with nothing, and then add matchers/filters based on the feature requests.

Potential reason why it would good to have: https://github.com/grafana/grafana/issues/14718#issuecomment-451471960

bwplotka:
Defintiely! but still it could be unnecessary big. But maybe small enough to avoid matchers, true (:

brian-brazil:
If you've more than about 100 label names, you're doing something very wrong.

bwplotka:
Yes, you are all right, label names should be quite static. Ignore me then (:

PR #4837, opened 2018-11-08
Add in-memory tracing back-end support 

### LINKED ISSUE #4509: PROPOSAL: Add tracing backend support

sipian:
Closes #4509 

This implementes an in-memory backend for traces using zipkin-ui.
Design Doc: https://docs.google.com/document/d/1NpfEkXRswmhcuh-BsmtfOVRJ0r7Ft7S9_JIo0NWYoGk/edit?usp=sharing

~~A working POC can be seen at http://35.244.22.86:9090/traces/~~

@brian-brazil @gouthamve @tomwilkie


brian-brazil:
At a high level, we might want to put this code somewhere a bit more general. I want to use it in the blackbox exporter at least.

sipian:
The entire tracer code is in a separate [directory](https://github.com/sipian/prometheus/tree/loki-opentracing/tracer).
The only place this is linked in prometheus code is [web.go](https://github.com/prometheus/prometheus/pull/4837/commits/b597f9eecd6dc4b5ed625da8e69be02fe56296f2#diff-73cbd2009b0e4f156e8ca0f47e95b016) to register a new tracer HTTP handler.
(we could also create a new repo for this in the prometheus org)

sipian:
@brian-brazil 
Benchmarking allocs is a bit unclear to me. Could you please elaborate a bit more on how to proceed with it. Will be helpful.

brian-brazil:
https://methane.github.io/2015/02/reduce-allocation-in-go-code/ is one guide, I think Beorn has a few good talks covering it too.

tracer/model/tracer.go
```diff
@@ -33,21 +32,12 @@ const (
...
-	defer seededIDLock.Unlock()
-
-	x := seededIDGen.Uint64()
-	for x == 0 {
-		x = seededIDGen.Uint64()
-	}
-
-	return x
+func newID() uint64 {
+	atomic.AddUint64(&startID, 1)
```
brian-brazil: Return this, otherwise you have a race.

codesome:
After recent discussion, we don't need this anymore as we have Jaeger support right now and we plan to support Open Telementary in the long term.

PR #4840, opened 2018-11-08
Correctly skip mismatching targets

fabxc:
@simonpasquier

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -547,6 +547,15 @@ func (api *API) targets(r *http.Request) (interface{}, *apiError, func()) {
 	return res, nil, nil
 }
 
+func matchLabels(lset labels.Labels, matchers []*labels.Matcher) bool {
+	for _, m := range matchers {
+		if !m.Matches(lset.Get(m.Name)) {
+			return false
+		}
+	}
+	return true
+}
+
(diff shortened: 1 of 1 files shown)


fabxc:
Suggestions for testing?

This often comes up with bugs triggered by map iteration randomization. In the end it seems the only way to validate this with some reliability is looping a test 1000 times – and even then it might not happen.
Doing that for every place where we iterate over map would get a bit out of hand.

simonpasquier:
> In the end it seems the only way to validate this with some reliability is looping a test 1000 times – and even then it might not happen

That was my idea. Agree that it might give a feeling of false confidence...

jkohen:
> Suggestions for testing?

Drive-by comment. At the extreme, break it by design: implement TargetsActive so that it usually returns elements in a different order. May not be practical here, but it could be applied to testing, if you control the implementation in the test.

fabxc:
Alternate proposal, let's factor out the behavior into a function so we can skip on the label, which is easy to misplace.
I think that prevents the issue from happening in the first place in a very obvious way. I'd prefer that to a randomized (and likely flaky) test that tries to fight the runtime. Those have their place, e.g. fuzzing our sample compression. But it just doesn't seem like a scalable approach for every map iteration in our code base.

simonpasquier:
@fabxc works for me!

Issue #4683, opened 2018-10-01
Allow customizing the server name


austin987:
Currently, it's hardcoded:
https://github.com/prometheus/prometheus/blob/9ee4b05d16de0b1ecd8bdaa644c2d72cdb8e41ee/web/ui/templates/_base.html#L6

as `Prometheus Time Series Collection and Processing Server`

I'd like the ability to customize it with my own title. In particular, I'd like to show the current environment, so I can distinguish between different instances.

So for example:
`'Dev Prometheus Server`
`Prod Prometheus Server`

Noticed while running app-metrics/prometheus-2.3.2, but the relevant code is unchanged in 2.4.2.

achiuBAE:
Hi, is anyone working on this one? If not, could I work on it?

achiuBAE:
What would be the optimal way to implement this? Through the config file, or through the flag? I see the flags are purely for configuring immutable system parameters, so i'm guessing the flag wouldn't be the way?
Or even through the front end?

Thanks

austin987:
IMO config file or flag is fine. Front end would not be preferred, as that's harder to automate ;)

austin987:
Fixed by https://github.com/prometheus/prometheus/commit/a9050c45f68f00cd9907df9bab96afd1908db934

PR #4841, opened 2018-11-08
Allow setting the Prometheus instance document title through a flag. 

achiuBAE:
implemented: #4683 
- The HTML title can now be set via a flag: --web.ui-title="[custom name]"

cc @juliusv


juliusv:
I can actually see this being useful if you have a bunch of tabs with different Prometheus servers open and you want to quickly differentiate them by, say, datacenter name. However, I want to make sure that others think it's a good idea to add this as a flag as well.

@brian-brazil and others?

brian-brazil:
I'm not sure about this one either way to be honest.

achiuBAE:
Hey has there been any decision made or this or does anything else need changing? :)

cmd/prometheus/main.go
```diff
@@ -150,6 +150,9 @@ func main() {
 	a.Flag("web.console.libraries", "Path to the console library directory.").
 		Default("console_libraries").StringVar(&cfg.web.ConsoleLibrariesPath)
 
+	a.Flag("web.ui-title", "Document title of Prometheus instance.").
```
juliusv: I think I'd call this `web.page-title` or `web.html-title`, with preference towards the former.

simonpasquier:
@achiuBAE you need to run `make assets`, commit the changes and push again.

simonpasquier:
LGTM

PR #4843, opened 2018-11-08
Use release binary of promu

SuperQ:
Download and extract a release binary of promu, rather than pull from
master.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -33,6 +33,12 @@ STATICCHECK  := $(FIRST_GOPATH)/bin/staticcheck
 GOVENDOR     := $(FIRST_GOPATH)/bin/govendor
 pkgs          = ./...
 
+GO_VERSION        ?= $(shell $(GO) version)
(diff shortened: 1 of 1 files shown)


Makefile.common
```diff
@@ -120,8 +126,13 @@ common-docker-tag-latest:
 	docker tag "$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)" "$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME):latest"
 
 .PHONY: promu
-promu:
-	GOOS= GOARCH= $(GO) get -u github.com/prometheus/promu
+promu: $(PROMU)
+
+.PHONY: $(PROMU)
```
sdurrheimer: @SuperQ Looks like the curl is done for each platforms during the crossbuild. Maybe because of this line ?

SuperQ: Yup, that was it. Fixed.

PR #4847, opened 2018-11-09
Use a fake SeriesIterator for BufferedSeriesIterator benchmarking

free:
Use a fake `SeriesIterator` implementation (that generates samples on the fly instead of using a slice) for `BufferedSeriesIterator` benchmarking, to reduce the variance of benchmark results due to memory pressure.

Before this change, benchmark results for `BenchmarkBufferedSeriesIterator-4` would randomly jump around between 19 and 35 ns/op on my machine. Now it's pretty stable at between 16.7 and 17.1 ns/op.

Signed-off-by: Alin Sinpalean <alin.sinpalean@gmail.com>


brian-brazil:
Thanks!

PR #4848, opened 2018-11-09
[promtool] Support writing output as json

stuartnelson3:
Oftentimes I'll want to execute something based on
the output from promtool, and supporting json
makes it easy to pull out values with a supporting
tool such as jq.

Signed-off-by: stuart nelson <stuartnelson3@gmail.com>

### DIFF
--- a/cmd/promtool/main.go
+++ b/cmd/promtool/main.go
@@ -30,6 +30,7 @@ import (
 	"github.com/prometheus/client_golang/api"
 	"github.com/prometheus/client_golang/api/prometheus/v1"
 	config_util "github.com/prometheus/common/config"
+	"github.com/prometheus/common/model"
 	"github.com/prometheus/common/version"
 	"github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/pkg/rulefmt"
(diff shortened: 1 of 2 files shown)


stuartnelson3:
@krasi-georgiev @simonpasquier @brancz @brian-brazil any thoughts?

brancz:
How different from calling curl is that really?

stuartnelson3:
As a result? Not very. As far as UX goes, I think it's a substantial improvement:

```
./promtool query instant http://prometheus-server.example.net 'histogram_quantile(0.9,
    sum by(job, instance, le) (rate(coredns_dns_request_duration_seconds_bucket[1m])))
    * 1000 and on(job, instance) sum by(job, instance) (rate(coredns_dns_request_count_total[5m]))' -o json
```

vs.

```
curl 'http://prometheus-prodeng.x-stn.s-cloud.net/api/v1/query?query=histogram_quantile(0.9%2C%0A%20%20%20%20sum%20by(job%2C%20instance%2C%20le)%20(rate(coredns_dns_request_duration_seconds_bucket%5B1m%5D)))%0A%20%20%20%20*%201000%20and%20on(job%2C%20instance)%20sum%20by(job%2C%20instance)%20(rate(coredns_dns_request_count_total%5B5m%5D))&time=1542103601.27&_=1542103590601'
```

brancz:
That's fair. I'm fine with having json support. We could also have a `--dry-run` flag or something that just spits out the request URL, that could then be used for composition, but I think either works - no strong opinion.

stuartnelson3:
Used the printer interface you suggested, though I'm not sure it's worth it :man_shrugging:

stuartnelson3:
Any more comments?

cmd/promtool/main.go
```diff
@@ -57,6 +59,7 @@ func main() {
 	checkMetricsCmd := checkCmd.Command("metrics", checkMetricsUsage)
 
 	queryCmd := app.Command("query", "Run query against a Prometheus server.")
+	queryCmdFmt := queryCmd.Flag("format", "Output format of the query.").Short('o').Default("promql").Enum("promql", "json")
```
simonpasquier: Not sure why the help text doesn't tell that valid values for this flag are `promql` or `json`.

stuartnelson3: :/

PR #4852, opened 2018-11-10
Add metric prometheus_rule_group_last_evaluation for recording and alerting

kdada:
In Prometheus 2.5, it supports last evaluation time of rules. But only metric `prometheus_rule_group_last_duration_seconds` is outputted by rule manager. I think `prometheus_rule_group_last_evaluation` is useful for users to know when a rule group was running.


kdada:
@juliusv

rules/manager.go
```diff
@@ -818,17 +824,28 @@ func (m *Manager) AlertingRules() []*AlertingRule {
...
 	ch <- groupInterval
 }
 
 // Collect implements prometheus.Collector.
 func (m *Manager) Collect(ch chan<- prometheus.Metric) {
 	for _, g := range m.RuleGroups() {
+		lastEvaluationTime := g.GetEvaluationTimestamp()
+		lastEvaluationTimestamp := 0.0
+		if !lastEvaluationTime.IsZero() {
+			lastEvaluationTimestamp = float64(timestamp.FromTime(lastEvaluationTime)) / 1e3
```
juliusv: Not sure why the indirection through `timestamp.FromTime()` is needed here. Couldn't you just do:

```suggestion
			lastEvaluationTimestamp = lastEvaluationTime.Seconds()
```

rules/manager.go
```diff
@@ -818,17 +824,28 @@ func (m *Manager) AlertingRules() []*AlertingRule {
...
+	ch <- lastEvaluation
 	ch <- lastDuration
 	ch <- groupInterval
 }
 
 // Collect implements prometheus.Collector.
 func (m *Manager) Collect(ch chan<- prometheus.Metric) {
 	for _, g := range m.RuleGroups() {
+		lastEvaluationTime := g.GetEvaluationTimestamp()
+		lastEvaluationTimestamp := 0.0
```
juliusv: I think `NaN` or `-Inf` would make more sense than 0 here. If I think about how I'd use it in an alert (alerting on a group not having been evaluated in a too long time), my expression would be:

    time() - prometheus_ rule_group_last_evaluation_timestamp_seconds > x

...combined with a `for ...` clause. In that scenario, `-Inf` would be most useful, as I wouldn't have to add another check for it staying in `NaN` for too long. @brian-brazil agreed?

rules/manager.go
```diff
@@ -90,6 +90,12 @@ var (
 		Name:      "rule_group_iterations_total",
 		Help:      "The total number of scheduled rule group evaluations, whether executed or missed.",
 	})
+	lastEvaluation = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, "", "rule_group_last_evaluation"),
```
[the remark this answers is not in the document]
kdada: OK.

rules/manager.go
kdada: The durations of evaluations are in milliseconds.  So `lastEvaluationTimestamp` keeps three decimal fractions. WDYT?

rules/manager.go
kdada: `-Inf` looks more useful.

rules/manager.go
kdada: The result of `lastEvaluationTime` is type of `time.Time`. Its `Seoncds()` only returns time in seconds. I think it should be changed to ` float64(lastEvaluationTime.Second()) + float64(lastEvaluationTime.Nanosecond())/1e9`. Just likes what `time.Duration.Seconds()` does. WDYT?

kdada: My fault. `lastEvaluationTime.Second()` is not correct.

But `lastEvaluationTime` has no method named `Seconds`.
This is correct:
```go
lastEvaluationTimestamp = float64(lastEvaluationTime.UnixNano()) / 1e9
```

juliusv: Ah hurrdurr, I was thinking of the `time.Duration` type, which does have the `Seconds()` method. Yep, your suggestion makes total sense.

juliusv:
👍 Thanks!

PR #4858, opened 2018-11-12
Fix goroutine leak in lexer/parser.

brian-brazil:
When there was an error in the parser, the
lexer goroutine was left running.

Also make runtime panic test actually test things.


juliusv:
:+1:

Issue #3778, opened 2018-02-01
Please update the Dockerfile on Dockerhub for the Prometheus image


CharlesWinter:
Just a quick note to please update the Dockerfile example shown on Dockerhub to reflect the new command line arguments that are required.

kaytrance:
Can you specify them? I cannot launch prometheus docker image

CharlesWinter:
@kaytrance Because the flags to Prometheus have changed in the latest version, I had to use the following CMD in my Dockerfile to get prometheus v2.1.0 to run.

```
CMD        [ "--config.file=/etc/prometheus/prometheus.yml", \
             "--storage.tsdb.path=/prometheus", \
             "--web.console.libraries=/etc/prometheus/console_libraries", \
             "--web.console.templates=/etc/prometheus/consoles", \
             "--web.external-url=https://YOURDOMAIN/prometheus", \
             "--web.route-prefix=/prometheus" ]
```

hairyhenderson:
If it's https://hub.docker.com/r/prom/prometheus/~/dockerfile/ you're talking about, that's not an example, that's the Dockerfile used for the automated build. Unfortunately it looks like the last time an automated build was run was ~2yrs ago (see https://hub.docker.com/r/prom/prometheus/builds/). 

Since then, it appears that docker images have been pushed [as part of the CircleCI build](https://github.com/prometheus/prometheus/blob/master/circle.yml#L62). So it's not _really_ an automated build...

I don't think there's a way to alter the listed Dockerfile manually in DockerHub - to update it must be run automatically again.

simonpasquier:
@kristaxox I'd rather remove the Dockerfile from Docker Hub as it is easy to get it from GitHub. I'm trying to get that done btw.

brian-brazil:
@matthiasr found https://github.com/christian-korneck/docker-pushrm which may be of use.

beorn7:
Hello from the bug scrub.

Is this still relevant? @simonpasquier have you done what you planned to do 5½ years ago?

Closing this, but please re-open if it still relevant.

PR #4864, opened 2018-11-14
Migrate all Docker image references to quay.io

### LINKED ISSUE #4665: Choose quay.io or docker hub and use it in docs & readme
The [README.md](https://github.com/prometheus/prometheus/blob/master/README.md#docker-images) points to quay.io for docker images, while the [installation page](https://github.com/prometheus/prometheus/blob/master/docs/installation.md#using-docker) points to Docker Hub.

As a person who is not familiar with whole docker image ecosystem it took me a while to figure out which one was the up-to-date version and more *correct* to use. It turns out both are kept up-to-date by the CI system. Additionally due to https://github.com/prometheus/prometheus/issues/3778#issuecomment-375965987 , one can think that there's actually a difference between both builds after some sleuthing on Dockerfiles.

Can you decide one of these Docker repos, and list only that one in documentation?

noqcks:
Based on the thread [here](https://groups.google.com/forum/#!msg/prometheus-developers/PGOASKQfJhQ/WiQcAeHKAQAJ), it seems the informal consensus is that DockerHub _and_ Quay.io will both remain. I updated all the image references to use quay.io to keep things consistent. If Docker Hub is preferred, I can change it.

**Changes**: 
- Migrate all Docker image references to use quay.io
- Explain in docs that Docker images are available on both Quay.io and Docker Hub

Something that we should do in addition to this PR that I can't change personally is remove the [Dockerfile from Docker Hub](https://hub.docker.com/r/prom/prometheus/~/dockerfile/) as mentioned in [this](https://github.com/prometheus/prometheus/issues/3778#issuecomment-375965987) comment. 

Fixes #4665

cc: @simonpasquier


simonpasquier:
Thanks for tackling this. I don't have any preference whether the examples should use quay.io or Docker Hub so this looks good to me.

brian-brazil:
Docker Hub is what people are used to, and we've seen reliability issues with Quay.

sdurrheimer:
IIRC Quay was used in order to have access to security analysis of our images.

noqcks:
Now using Docker Hub as default based on feedback from @brian-brazil 

I agree Docker Hub is standard and quay could confuse people.

noqcks:
What's the process for getting it merged in as a contributor? 

I noticed some other PRs get approved but don't get merged in immediately: https://github.com/prometheus/prometheus/pull/4719

simonpasquier:
@noqcks we usually wait for 2 approvals before merging. If it takes too long, it is ok to ping on IRC or here.

noqcks:
Ahhh good to know, thanks.

PR #4865, opened 2018-11-14
cmd/prometheus: buffer signal chans

squat:
According to the GoDoc for os.Signal [0]:

> Package signal will not block sending to c: the caller must ensure that
> c has sufficient buffer space to keep up with the expected signal rate.
> For a channel used for notification of just one signal value, a buffer
> of size 1 is sufficient.

[0] https://golang.org/pkg/os/signal/#Notify

Signed-off-by: Lucas Serven <lserven@gmail.com>

cc @brancz @simonpasquier

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -395,7 +395,7 @@ func main() {
 	var g group.Group
 	{
 		// Termination handler.
-		term := make(chan os.Signal)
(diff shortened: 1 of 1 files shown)


brancz:
Interesting. I've been using this incorrectly a lot.

lgtm :+1:

PR #4868, opened 2018-11-15
feat(azure_sd): Allow filtering VMs and scale sets by tags

hedgepigdaniel:
Currently with Azure service discovery there is no way to filter the VMs and scale sets that are scraped - the only option is to always scrape all servers.

This change extends `azure_sd_config` with a `tags` field. If tags are specified, only VMs and scale sets who's tags are a superset of the configured tags are scraped. This allows to scrape services only where they are running (e.g. to scrape the webserver stats only from VMs which are webservers).

I'm new to this codebase and to Go, so treat my code with suspicion :rofl:

### DIFF
--- a/discovery/azure/azure.go
+++ b/discovery/azure/azure.go
@@ -78,6 +78,7 @@ type SDConfig struct {
 	ClientID        string             `yaml:"client_id,omitempty"`
 	ClientSecret    config_util.Secret `yaml:"client_secret,omitempty"`
 	RefreshInterval model.Duration     `yaml:"refresh_interval,omitempty"`
+	Tags            map[string]*string `yaml:"tags,omitempty"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
@@ -249,7 +250,7 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 		return tg, fmt.Errorf("could not create Azure client: %s", err)
 	}
 
-	machines, err := client.getVMs()
+	machines, err := client.getVMs((*d.cfg).Tags)
 	if err != nil {
 		return tg, fmt.Errorf("could not get virtual machines: %s", err)
 	}
@@ -257,7 +258,7 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 	level.Debug(d.logger).Log("msg", "Found virtual machines during Azure discovery.", "count", len(machines))
 
 	// Load the vms managed by scale sets.
-	scaleSets, err := client.getScaleSets()
+	scaleSets, err := client.getScaleSets((*d.cfg).Tags)
 	if err != nil {
 		return tg, fmt.Errorf("could not get virtual machine scale sets: %s", err)
 	}
@@ -358,7 +359,7 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 	return tg, nil
 }
 
-func (client *azureClient) getVMs() ([]virtualMachine, error) {
+func (client *azureClient) getVMs(filterTags map[string]*string) ([]virtualMachine, error) {
 	var vms []virtualMachine
 	result, err := client.vm.ListAll()
 	if err != nil {
@@ -366,7 +367,9 @@ func (client *azureClient) getVMs() ([]virtualMachine, error) {
 	}
 
 	for _, vm := range *result.Value {
-		vms = append(vms, mapFromVM(vm))
+		if IncludesTags(filterTags, vm.Tags) {
+			vms = append(vms, mapFromVM(vm))
+		}
 	}
 
 	// If we still have results, keep going until we have no more.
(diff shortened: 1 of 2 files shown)


brian-brazil:
You can already filter based on tag with relabelling and __meta_azure_machine_tag_

hedgepigdaniel:
Sure, you can filter the scraped metrics like that - but the aim here is to filter the targets before they are scraped, not the metrics after they are scraped.

For background, I'm scraping from a setup which involves VMs with many distinct purposes, e.g. webservers, cache servers, database servers, etc. For example, our webservers have nginx, Django/uwsgi (12 worker processes each), and celery (4 worker processes each) - so 17 ports to scrape. Different services are on different ports, since in our staging environments we use each server for more services.

If I scrape all ports from all VMs and then filter the metrics afterwards:
- Many scrape requests are made to machines that aren't listening or exporting metrics. Not a disaster, but also not good. I might want to use those ports for another purpose, and not get regular requests to `/metrics`
- The targets page in the prometheus web UI is permanently full of red failures. This makes it hard to see if all the scrape targets are working, since even when they are its still a sea of red
- If there are different naming conventions for different types of servers and I want to make the labels in prometheus consistent, the `relabel_configs` become complex and brittle, since I need to ensure that in all cases the right set of regexes match - that they don't overlap, or that they are all compatible.

The solution I'm using is to tag each server with its deployment name and the services it exposes. For example, a webserver might be tagged as follows:
- `deployment_live`: ''
- `service_django_celery_uwsgi`: ''
- `service_nginx`: ''

```yml
scrape_configs:

  {% for deployment in prometheus_deployments %} # Jinja2

  # This job is to scrape all the VMs which are part of scale sets
  - job_name: 'azure-scale-set-{{ deployment }}'
    relabel_configs:

      # Set the unique instance to "<machine_name>_<scaleset-instance>_<port>"
      - source_labels:
          - __meta_azure_machine_name
          - __address__
        regex: ^(.*-.*)-[0-9]+-[0-9]+-[0-9a-f]+-scaleset_(\d+);.*:([0-9]+)$
        target_label: instance
        replacement: ${1}_instance-${2}_port-${3}

      # Set the "server" label to the scale set name and the instance number
      - source_labels: [__meta_azure_machine_name]
        regex: ^(.*-.*)-[0-9]+-[0-9]+-[0-9a-f]+-scaleset_(\d+)$
        target_label: server
        replacement: ${1}_instance-${2}

      # Tag these metrics with the deployment
      - action: replace
        target_label: deployment
        replacement: '{{ deployment }}'

azure_sd_configs:
      {% for port in range(8020, 8020 + uwsgi_workers_per_server + celery_workers_per_server) %}

      - subscription_id: {{ subscription_id }}
        tenant_id: {{ tenant_id }}
        client_id: {{ client_id }}
        client_secret: {{ client_secret }}
        port: {{ port }}
        tags:
          service_django_celery_uwsgi: ''
          deployment_{{ deployment }}: ''

      {% endfor %}

      # nginxlog-exporter
      ### FOR ARGUMENTS SAKE:
      # I might want to put this in a different job, and use different relabel_configs to
      # account for a different naming convention (without having to make the regexes
      # compatible with all the other options, or scraping all the servers a second time)
      - subscription_id: {{ subscription_id }}
        tenant_id: {{ tenant_id }}
        client_id: {{ client_id }}
        client_secret: {{ client_secret }}
        port: 4040
        tags:
          service_nginx: ''
          deployment_{{ deployment }}: ''
```

The equivalent for `static_config` would be simply to specify multiple sets of servers - at the moment I don't think the same is possible with `azure_sd_config`.

brian-brazil:
The existing metadata will allow you to do that.

hedgepigdaniel:
Could you be more specific?

I'm trying to filter the scrape targets themselves, not the scraped metrics, for reasons explained above.

It is not clear to me how metadata from the already scraped targets helps me to do that.

hedgepigdaniel:
The `filter ` option for Google cloud seems to serve a similar function: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#<gce_sd_config>

hedgepigdaniel:
For [ec2](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#<ec2_sd_config>) again there is a "filters" option.

The Azure service discovery seems to be the outlier, in that there is no way to discover a subset of VMs.

brian-brazil:
This is a usage question, I'd suggest taking this to the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users). On the mailing list, more people are available to potentially respond to your question, and the whole community can benefit from the answers provided.

hedgepigdaniel:
Am I wrong that relabel configs are evaluated after discovery and scraping? That "_meta_azure_machine_tag" is only available AFTER discovery/scraping, so too late?

I'll ask on the mailing list, but since you can clearly see what I've missed, if you could point me in the direction of what my misunderstanding is, I'd be grateful.

simonpasquier:
you're looking for [`relabel_configs`](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#%3Crelabel_config%3E) which happens after discovery but before scraping. This is equivalent to what you're attempting in this PR. Other SD mechanisms have filtering options because it is supported by the EC2/GCP/... API.

hedgepigdaniel:
Right, that makes sense, thanks!

Not sure why I missed that. Potentially tweaking the first sentence in the relabel configs docs might help to avoid such confusion:

"Relabeling is a powerful tool to dynamically rewrite the label set of a target before it gets scraped"

That doesn't sound like it's possible to prevent the target from being scraped at all.

PR #4869, opened 2018-11-15
*: remove use of golang.org/x/net/context

simonpasquier:


brian-brazil:
:+1: Looks like the tests hung

simonpasquier:
I've rewritten the [test](https://github.com/prometheus/prometheus/pull/4869/files#diff-9d64f310300ba30206ddf9eb6efaae71) since `scrape()` doesn't return `context.Canceled` anymore on context cancellation.

PR #4870, opened 2018-11-15
add a wmi_exporter node-overview console template

nbirnel:
@brian-brazil 

This is so the next person who wants a console for Windows nodes doesn't have to write it themself.

Despite #3099, I am unable to find anything on mailing list suggesting that consoles are deprecated, so I believe this is worthwhile. https://groups.google.com/forum/?fromgroups#!searchin/prometheus-developers/console|sort:date/prometheus-developers/VzmcWJKRQxY/mUVBSzVCGQAJ is an older conversation, but I see nothing more recent on the subject.

If the prometheus team does not wish to maintain console templates, but will not be removing them, perhaps the docs could instead point to a separate repo to avoid duplication of effort for those who want to use them.

### DIFF
--- a/consoles/wmi-overview.html
+++ b/consoles/wmi-overview.html
@@ -1,0 +1,140 @@
+{{ template "head" . }}
+
+{{ template "prom_content_head" . }}
+  <h1>Node Overview - {{ reReplaceAll "(.*?://)([^:/]+?)(:\\d+)?/.*" "$2" .Params.instance }}</h1>
+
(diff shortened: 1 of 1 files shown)


simonpasquier:
This should go in the [wmi-exporter](https://github.com/martinlindhe/wmi_exporter) repository as the decision has already been made that this repository can't contain templates for all possible exporters.

nbirnel:
Thanks @simonpasquier, I am closing this and opening a PR over there.

PR #4875, opened 2018-11-16
Populate series in the PromQL engine in two phases.

tomwilkie:
This allows other storage implementations (for instance, where Select might take more time to execute) to process multiple selects in the background and speed up queries.


brian-brazil:
Sure. Hopefully the other will go in early next week.

PR #4878, opened 2018-11-19
Add delay for rule evaluation to allow metric lateness

garrettlish:
Sometimes we are using metric source time rather than Prometheus scrape time as metric timestamp, it means that we might have some delays during rule evaluation. Although we have staleness mechanism to fallback to 5 minutes if no sample found, however, we cannot get accurate result with this fallback mechanism while delay happens. So the idea is adding delay for rule evaluation to allow metric lateness.

```
groups:
  - name: test
    delay: 30s
    rules:
    - record: job:http_requests:rate5m
      expr: sum by (job)(rate(http_requests_total[5m]))
```

### DIFF
--- a/promql/quantile.go
+++ b/promql/quantile.go
@@ -115,7 +115,25 @@ func bucketQuantile(q float64, buckets buckets) float64 {
 func coalesceBuckets(buckets buckets) buckets {
 	last := buckets[0]
 	i := 0
+	var next float64
 	for _, b := range buckets[1:] {
+		// use next count if current count is smaller than previous one which means data loss
+		if b.count < last.count {
+			if next >= last.count {
+				b.count = next
+			} else {
+				for j := 1; ; j++ {
+					if i+j > len(buckets)-1 {
+						break
+					}
+					if buckets[i+j].count >= last.count {
+						next = buckets[i+j].count
+						b.count = next
+						break
+					}
+				}
+			}
+		}
 		if b.upperBound == last.upperBound {
 			last.count += b.count
 		} else {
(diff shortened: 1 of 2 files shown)


brian-brazil:
How would this interact with ALERTS and ALERT_FOR_STATE?

garrettlish:
@brian-brazil ALERTS is generated if *evaluation time* subtract activeAt timestamp > hold duration, here we *shift the evaluation time to past time* to allow metric lateness - both evaluation time and activeAt shift to past time, the ALERTS condition won't be changed and impacted, only change is that the ALERTS will get delayed. ALERT_FOR_STATE is the same.

garrettlish:
We are sending shifted timestamp (metric source timestamp) to alertmanager, restoring alerts is also using metric source timestamp (for example - delay 1 minute: we will send 2018-11-20 10:00:00 metric at 2018-11-20 10:00:01 to alertmanager). Do you mean metric source timestamp is incorrect timestamp (2018-11-20 10:00:00 is incorrect in the above example?)?

brian-brazil:
> Do you mean metric source timestamp is incorrect timestamp (such as 2018-11-20 10:00:00 is incorrect in the above example?)?

It'd cause for example the alert to be considered resolved early, and restoration of alerts not to happen when it should due to the data being too far back.

garrettlish:
@brian-brazil thanks, you are correct, alert restoration will be impacted if the data is too far back, apply delay timestamp shift for both evaluation and restoration

brian-brazil:
I'm not sure this is the right approach, I'd expect ALERTS to have the evaluation time rather than the adjusted time. Otherwise it's not an alert history.

garrettlish:
@brian-brazil from my perspective, if user expects metric lateness, it means that user prefer using metric source time rather than rule evaluation time (process time). Then the adjusted time rather than evaluation time (process time) should be the proper value for alert history.

brian-brazil:
I'd disagree, ALERTS is what's currently firing.

PR #4879, opened 2018-11-19
Add replay to re-build alert state

garrettlish:
If Prometheus crash or restarted, the on-the-fly alert state will be lost, we need to replay the rules within the holding duration to re-build alert state.

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -365,6 +365,8 @@ func main() {
 				}
 				files = append(files, fs...)
 			}
+			// Add replay to re-build alert state
+			ruleManager.Replay(time.Duration(cfg.GlobalConfig.EvaluationInterval), files)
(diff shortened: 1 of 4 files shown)


simonpasquier:
Hmm, I'm not sure I understand exactly what you are trying to accomplish but at first glance, I'd say that re-evaluating rules when Prometheus was down/crashed isn't very consistent since the raw data will be missing.

simonpasquier:
Since 2.4.0, Prometheus should keep track of the pending and firing alerts across restarts (see https://github.com/prometheus/prometheus/issues/422). In your example, the "new" Prometheus should fire the alert after the first evaluation.

garrettlish:
Thanks @simonpasquier, I am using 2.3.0, will upgrade to 2.4.0 (https://github.com/prometheus/prometheus/pull/4061). Close it since it is already implemented.

Issue #4882, opened 2018-11-19
prometheus_tsdb_head_max_time and prometheus_tsdb_head_min_time values is milliseconds


AndorCS:
## Proposal
**Use case. Why is this important?**
Since prometheus use standard values in seconds, having `prometheus_tsdb_head_max_time` and `prometheus_tsdb_head_min_time` in milliseconds is a bit confusing.

## Bug Report
**What did you do?**
Query `prometheus_tsdb_head_max_time - prometheus_tsdb_head_min_time`
**What did you expect to see?**
`1339.077`
**What did you see instead? Under which circumstances?**
`1339077`
**Environment**
any
* System information:
any
* Prometheus version:
```
prometheus_build_info{branch="HEAD",goversion="go1.11.1",revision="167a4b4e73a8eca8df648d2d2043e21bdb9a7449",version="2.4.3"}
```

* Alertmanager version:
N/A
* Prometheus configuration file:
N/A
* Alertmanager configuration file:
N/A
* Logs:
N/A

PR #4884, opened 2018-11-19
Add TSDB timestamp metrics that use base units (seconds)

bobmshannon:
- Adds TSDB metrics `prometheus_tsdb_start_time_seconds`, `prometheus_tsdb_head_max_time_seconds`, and `prometheus_tsdb_head_min_time_seconds` metrics.

Addresses https://github.com/prometheus/tsdb/issues/395 and https://github.com/prometheus/prometheus/issues/4882.

If this gets merged, then I think we may want to deprecate the old metrics without the base units: `prometheus_tsdb_lowest_timestamp`, `prometheus_tsdb_head_max_time`, and `prometheus_tsdb_head_min_time` because with this change they are redundant and likely to cause more confusion for users.

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -50,6 +50,7 @@ import (
 	"github.com/prometheus/prometheus/discovery"
 	sd_config "github.com/prometheus/prometheus/discovery/config"
 	"github.com/prometheus/prometheus/notifier"
+	"github.com/prometheus/prometheus/pkg/timestamp"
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
thanks for spending the time on this. We worked on the same issue and the other PR got approved and merged so I am gonna have to close this one, sorry.

I have already opened a PR to remove/rename these in tsdb https://github.com/prometheus/tsdb/pull/457

PR #4891, opened 2018-11-21
remote: Set User-Agent header in requests

daenney:
Currently Prometheus requests show up with a UA of Go-http-client/1.1
which isn't super helpful. Though the X-Prometheus-Remote-* headers
exist they need to be explicitly configured when logging the request in
order to be able to deduce this is a request originating from
Prometheus. By setting the header we remove this ambiguity and make
default server logs just a bit more useful.

### DIFF
--- a/storage/remote/client.go
+++ b/storage/remote/client.go
@@ -26,12 +26,15 @@ import (
 	"github.com/gogo/protobuf/proto"
 	"github.com/golang/snappy"
 	"github.com/prometheus/common/model"
+	"github.com/prometheus/common/version"
 
 	config_util "github.com/prometheus/common/config"
 	"github.com/prometheus/prometheus/prompb"
 )
 
 const maxErrMsgLen = 256
+
+var userAgent = fmt.Sprintf("Prometheus/%s", version.Version)
 
 // Client allows reading and writing from/to a remote HTTP endpoint.
 type Client struct {
@@ -83,6 +86,7 @@ func (c *Client) Store(ctx context.Context, req *prompb.WriteRequest) error {
 	}
 	httpReq.Header.Add("Content-Encoding", "snappy")
 	httpReq.Header.Set("Content-Type", "application/x-protobuf")
+	httpReq.Header.Set("User-Agent", userAgent)
 	httpReq.Header.Set("X-Prometheus-Remote-Write-Version", "0.1.0")
 	httpReq = httpReq.WithContext(ctx)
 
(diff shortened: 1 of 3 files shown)


storage/remote/client.go
```diff
@@ -26,13 +26,16 @@ import (
...
 	"github.com/prometheus/common/model"
+	"github.com/prometheus/common/version"
 
 	config_util "github.com/prometheus/common/config"
 	"github.com/prometheus/prometheus/prompb"
 )
 
 const maxErrMsgLen = 256
 
+var userAgent = fmt.Sprintf("prometheus/%s", version.Version)
```
juliusv: During scrape, we capitalize the "Prometheus": https://github.com/prometheus/prometheus/blob/bd100182b2b9f4c2a632ec98f96d5d1a152a09f9/scrape/scrape.go#L455

And then in k8s discovery, we set it to something else entirely, "prometheus/discovery", without version: https://github.com/prometheus/prometheus/blob/b75ec7e6ef3666695edfa7bc1d9fccd1fdaae82a/discovery/kubernetes/kubernetes.go#L241

Makes me wonder if we shouldn't standardize all this, but for now it might be good to at least capitalize "prometheus" here too.

daenney: I'm more than happy to change the capitalization. As to what the scheme should be, I generally use something like "Name/version".

MDN suggests the format for UA should be `<product> / <product-version> <comment>`. So I suppose in the case of k8s discovery for example it would make more sense to have it as `Prometheus/x.x.x k8s-discovery`, and follow a similar pattern everywhere else too.

juliusv: That makes sense, thanks! I think for now just the capitalization change would be good to be consistent. Also, in case you want to add the same for the Prom->AM communication (that should cover all outgoing communication then, besides SD mechanisms), that would be awesome, but not required.

daenney: AM... AlertManager? I'm sure I can dig that up too while we're at it. Consistency is nice. I'll fix the capitalisation for now and spawn a separate PR for the global UA thing after that.

juliusv: Yep, Alertmanager: https://github.com/prometheus/prometheus/blob/master/notifier/notifier.go#L497

That'd be nice, thanks!

daenney:
@juliusv I think the new commit includes everything we talked about. I'll get cracking on a PR for common tomorrow and send a second PR to get that all in place.

simonpasquier:
LGTM

juliusv:
👍 Thanks!

PR #4894, opened 2018-11-22
fix deadlock in scrape manager

kdada:
Scrape manager will fall in deadlock when we reload configs frequently.

Situation:
We don't use prometheus discovery to re-config our targets. As instead, we implemented an external discoverer to generate static scrape config. So it will reload prometheus when some configs are changed. 
When it reloads prometheus in a high frequency (~ 20/min), prometheus gets stuck occasionally.

The logic of the deadlock is:
```go
func (m *Manager) reload() {
	m.mtxScrape.Lock()
	var wg sync.WaitGroup
	// m.targetSets is from discovery. It's from old config.
	for setName, groups := range m.targetSets {
		var sp *scrapePool
		// m.scrapePools and m.scrapeConfigs are from new config.
		// New config has deleted some static scrape job. So it causes error.
		existing, ok := m.scrapePools[setName]
		if !ok {
			scrapeConfig, ok := m.scrapeConfigs[setName]
			if !ok {
				level.Error(m.logger).Log("msg", "error reloading target set", "err", "invalid config id:"+setName)
				return
			}
			sp = newScrapePool(scrapeConfig, m.append, log.With(m.logger, "scrape_pool", setName))
			m.scrapePools[setName] = sp
		} else {
			sp = existing
		}

		wg.Add(1)
		go func(sp *scrapePool, groups []*targetgroup.Group) {
			sp.Sync(groups)
			wg.Done()
		}(sp, groups)

	}
	m.mtxScrape.Unlock()
	wg.Wait()
}
```

### DIFF
--- a/scrape/manager.go
+++ b/scrape/manager.go
@@ -110,6 +110,7 @@ func (m *Manager) reload() {
 			scrapeConfig, ok := m.scrapeConfigs[setName]
 			if !ok {
 				level.Error(m.logger).Log("msg", "error reloading target set", "err", "invalid config id:"+setName)
+				m.mtxScrape.Unlock()
 				return
 			}
 			sp = newScrapePool(scrapeConfig, m.append, log.With(m.logger, "scrape_pool", setName))


scrape/manager.go
```diff
@@ -110,6 +110,7 @@ func (m *Manager) reload() {
 			scrapeConfig, ok := m.scrapeConfigs[setName]
 			if !ok {
 				level.Error(m.logger).Log("msg", "error reloading target set", "err", "invalid config id:"+setName)
+				m.mtxScrape.Unlock()
```
simonpasquier: Thanks for catching it! A `defer m.mtxScrape.Unlock()` following L104 would be easier to read and maintain IMO.

kdada: This method has a `sync.WaitGroup` to wait some goroutines finishing. So it should release `m.mtxScrape` as soon as possible.

simonpasquier: Hmm, in fact the problem is that we shouldn't return in the middle of the loop but rather continue to the next item. @krasi-georgiev?

kdada: In this case, we only can check if `m.targetSets` and `m.scrapeConfigs` has same keys. But it still does not guarantee that their configs are consistent.

simonpasquier: IIUC the scrape manager configuration has been reloaded but the discovery manager hasn't been yet. So it could send target updates for scraping configurations that don't exist anymore. It is just a transient discrepancy between the 2 managers and I would assume that it is safe to skip the group as the scrape manager knows the latest configuration.

krasi-georgiev: yes replacing `return` with `continue` would be most appropriate here.

krasi-georgiev:
```suggestion
```

scrape/manager.go
```diff
@@ -110,6 +110,7 @@ func (m *Manager) reload() {
...
 				return
```
krasi-georgiev:
```suggestion
				continue
```

kdada:
PTAL @krasi-georgiev @simonpasquier 

I replaced `return` with `continue`. But I think it is not a final solution. 

The execution order of reloading:

[User triggers `/-/reload`] --> reloadConfig[main.go:634] --> executing reloaders in order[main.go334]
```go
	reloaders := []func(cfg *config.Config) error{
		...
		scrapeManager.ApplyConfig,
		func(cfg *config.Config) error {
			c := make(map[string]sd_config.ServiceDiscoveryConfig)
			for _, v := range cfg.ScrapeConfigs {
				c[v.JobName] = v.ServiceDiscoveryConfig
			}
			return discoveryManagerScrape.ApplyConfig(c)
		},
		...
	}
```
When `scrape.Manager.reload` finds inconsistent configs, it can safely ignore this reloading. Because discoveryManagerScrape will trigger a new reloading after a while.

krasi-georgiev:
sorry I didn't quite understand your point. Could you explain with a bit more details.

kdada:
Reloading stack:
1. `syscall.SIGHUP` or `/-/reload` calls `main.reloadConfig`(main.go:643)
2. `main.reloadConfig` calls `reloaders`(main.go:334) in order. `scrapeManager.ApplyConfig` runs before `discoveryManagerScrape.ApplyConfig`.
3. `scrapeManager.ApplyConfig` stops deleted jobs and reloads changed jobs.
```go
// ApplyConfig resets the manager's target providers and job configurations as defined by the new cfg.
func (m *Manager) ApplyConfig(cfg *config.Config) error {
	m.mtxScrape.Lock()
	defer m.mtxScrape.Unlock()

	c := make(map[string]*config.ScrapeConfig)
	for _, scfg := range cfg.ScrapeConfigs {
		c[scfg.JobName] = scfg
	}
	// Update to latest scrape config.
	m.scrapeConfigs = c

	// Cleanup and reload pool if config has changed.
	for name, sp := range m.scrapePools {
		if cfg, ok := m.scrapeConfigs[name]; !ok {
			sp.stop()
			delete(m.scrapePools, name)
		} else if !reflect.DeepEqual(sp.config, cfg) {
			// The `reload` does not reload targets. It only create new client with new HTTPClientConfig.
			sp.reload(cfg)
		}
	}

	return nil
}
```
4. After `3`, `scrapeManager` has latest `scrapeConfigs` but does not update scrape targets. Then `discoveryManagerScrape.ApplyConfig` is called by `main.reloadConfig`.
5. `discoveryManagerScrape.ApplyConfig` gets `ServiceDiscoveryConfig`s from configs and creates a lot of providers. providers generate scrape groups/targets to `discoveryManager`.  Then `discoveryManager` sends these groups/targets to `scrapeManager.Run`(manager.go:66).
6. `scrapeManager.Run` updates `scrapeManager.targetSets` and triggers `scrapeManager.reload`.
7. `scrapeManager.reload` syncs groups/targets and creates new scraper.
```go
func (m *Manager) reload() {
	m.mtxScrape.Lock()
	var wg sync.WaitGroup
	for setName, groups := range m.targetSets {
		var sp *scrapePool
		existing, ok := m.scrapePools[setName]
		if !ok {
			scrapeConfig, ok := m.scrapeConfigs[setName]
			if !ok {
				// If there is no trouble happened, it won't come here.
				level.Error(m.logger).Log("msg", "error reloading target set", "err", "invalid config id:"+setName)
				continue
			}
			sp = newScrapePool(scrapeConfig, m.append, log.With(m.logger, "scrape_pool", setName))
			m.scrapePools[setName] = sp
		} else {
			sp = existing
		}

		wg.Add(1)
		// Run the sync in parallel as these take a while and at high load can't catch up.
		go func(sp *scrapePool, groups []*targetgroup.Group) {
			sp.Sync(groups)
			wg.Done()
		}(sp, groups)

	}
	m.mtxScrape.Unlock()
	wg.Wait()
}
```

The above is the complete logic of updating scrape configs. `scrapeManager.scrapeConfigs` is always same as or newer than `scrapeManager.targetSets`. If `scrapeManager.targetSets` is older than `scrapeManager.scrapeConfigs`, `discoveryManagerScrape` must be in the situation of waiting for sending new groups/targets to `scrapeManager`. So I think we can simply ignore current reloading and wait for the next `scrapeManager.targetSets`.

krasi-georgiev:
if I understand your point correctly you are saying the we can replace `continue` with `break`. I think that would be fine as well.

simonpasquier:
IMO it doesn't make much difference since the discovery manager should send updates right after. However the code prior to #4526 had a `continue` so I'd stick to it.

https://github.com/prometheus/prometheus/blob/ad4c33c1ff40553a8fb75615ab0ea3a74c3baf91/scrape/manager.go#L149-L172

kdada:
@krasi-georgiev Yes. 
@simonpasquier The two ways lead to same result. My opinion only is we can skip some useless operations. 

Anyway, changing it to `continue` solves this issue. Is there anything else needs to modify?

simonpasquier:
Thanks a lot @kdada!

PR #4895, opened 2018-11-22
discovery/dns: add trim_final_dot option for SRV records

### LINKED ISSUE #3196: DNS discovery should not trim final dot

simonpasquier:
Closes #3196

Following @juliusv's [suggestion](https://github.com/prometheus/prometheus/issues/3196#issuecomment-432663344), I've implemented an additional configuration option (`trim_final_dot`) that controls the removal of the leading dot. The default value is `true` to preserve the existing behavior.

cc @bboreham @grobie

### DIFF
--- a/discovery/dns/dns_test.go
+++ b/discovery/dns/dns_test.go
@@ -100,10 +100,37 @@ func TestDNS(t *testing.T) {
 			},
 		},
 		{
-			name: "SRV record query",
+			name: "SRV record query without trimming dot",
 			config: SDConfig{
 				Names:           []string{"_mysql._tcp.db.example.com."},
 				RefreshInterval: model.Duration(time.Minute),
+				TrimFinalDot:    false,
+			},
+			lookup: func(name string, qtype uint16, logger log.Logger) (*dns.Msg, error) {
+				return &dns.Msg{
+						Answer: []dns.RR{
+							&dns.SRV{Port: 3306, Target: "db1.example.com."},
+							&dns.SRV{Port: 3306, Target: "db2.example.com."},
+						},
+					},
+					nil
+			},
+			expected: []*targetgroup.Group{
+				&targetgroup.Group{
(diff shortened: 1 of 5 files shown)


discovery/dns/dns_test.go
```diff
@@ -0,0 +1,177 @@
...
+					return
+				}
+				t.Fatal("Empty channel")
+			}
+
+			if len(tg) != len(tc.expected) {
+				t.Fatalf("Invalid number of target groups, exp: %d, got: %d", len(tc.expected), len(tg))
+			}
+			for i := range tg {
+				if !tg[i].Equal(tc.expected[i]) {
```
juliusv: In many other places in Prometheus we either use https://github.com/prometheus/prometheus/blob/3b5dea4e6db822533a2e3d2eb1d253ec04853493/util/testutil/testing.go#L64-L70 or `testify`'s `require.Equal()`, like here: https://github.com/prometheus/prometheus/blob/2bd510a63e48ac6bf4971d62199bdb1045c93f1a/storage/fanout_test.go#L36

The benefit of the latter is nice built-in diff formatting when something isn't equal. Can we just use that here? Then we also wouldn't need the new `Equal()` method on the target group.

simonpasquier: Sure I'll update the PR.

simonpasquier:
@juliusv @bboreham @grobie any feedback on this patch?

juliusv:
@simonpasquier I think it's good. I was wondering for a while if the option name should be more semantic than technical, but couldn't think of a great name... so :+1:

brian-brazil:
Something that occurred to me yesterday is that this it make sense to have metadata for the raw version, and relabel from there.

codesome:
During the discussions in the bug scrub we decided that it is better as a metadata. Hence closing this PR.

PR #4897, opened 2018-11-22
collect cpu and trace profiles with the promtool debug command

krasi-georgiev:

### DIFF
--- a/cmd/promtool/main.go
+++ b/cmd/promtool/main.go
@@ -510,18 +510,22 @@ func parseTime(s string) (time.Time, error) {
 	return time.Time{}, fmt.Errorf("cannot parse %q to a valid timestamp", s)
 }
 
+var profilesList = map[string]string{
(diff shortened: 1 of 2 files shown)


cmd/promtool/main.go
```diff
@@ -510,18 +510,22 @@ func parseTime(s string) (time.Time, error) {
 	return time.Time{}, fmt.Errorf("cannot parse %q to a valid timestamp", s)
 }
 
+var profilesList = map[string]string{
+	"/debug/pprof/profile":      "cpu.pb",
+	"/debug/pprof/trace":        "trace.pb",
```
simonpasquier: IIUC this will only capture 1 second of trace data. Do we want a higher default value and some option to tweak it?

simonpasquier: Yeah hard coding doesn't bother me.

PR #4898, opened 2018-11-22
web: group targets by job then instance

### LINKED ISSUE #4853: Consul job rename/relabel not working with 2.5.0
## Proposal
Grouping of exporters when using consul is broken for job rename.

## Bug Report
**What did you do?**
Upgraded to latest 2.5 version. All exporters discovered with consul ended up with a single job name instead of rename for each exporter which was working flawlessly with 1.8.0

Continued renaming of job name, so all discovered exporters are sorted correctly in the prometheus view. Downgrading to 2.4.3 resolved the issue.

* Prometheus version:

2.5.0

simonpasquier:
Closes #4853

After #4526 the targets page no longer groups the targets by job label but by scrape configuration name. This change ensures that targets inside a scrape configuration are sorted by job then instance.

### DIFF
--- a/web/web.go
+++ b/web/web.go
@@ -55,7 +55,6 @@ import (
 
 	"github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/notifier"
-	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/rules"
 	"github.com/prometheus/prometheus/scrape"
@@ -702,7 +701,8 @@ func (h *Handler) targets(w http.ResponseWriter, r *http.Request) {
 	tps := h.scrapeManager.TargetsActive()
 	for _, targets := range tps {
 		sort.Slice(targets, func(i, j int) bool {
-			return targets[i].Labels().Get(labels.InstanceName) < targets[j].Labels().Get(labels.InstanceName)
+			return targets[i].Labels().Get(model.JobLabel) < targets[j].Labels().Get(model.JobLabel) ||
+				targets[i].Labels().Get(model.InstanceLabel) < targets[j].Labels().Get(model.InstanceLabel)
 		})
 	}
 


simonpasquier:
@juliusv @brian-brazil does it make sense for you?

juliusv:
This topic came up at the Berlin Prometheus Meetup a couple of weeks ago. @beorn7 presented this UI change as a good example for how the job label is losing significance, and he liked it. I pointed out that it will effectively break existing users who have one large scrape config that relabels targets into different jobs (common e.g. with Kubernetes SD). Even if we normally don't treat UI changes as breaking, this one has more far-reaching usability consequences than your usual visual "fix".

In the end we all agreed that it would be best to let the user choose a grouping in the UI, but I don't see anyone jumping up to implement that just yet. IMO for now it would be better to un-break this and revert to the old behavior as you are proposing here, but curious about what others think.

juliusv:
@simonpasquier if I understand correctly, this change would still keep the scrape configs `job_name` field as the major UI grouping, and only change the sorting within that group, without actually re-introducing a full group for each effective `job` label?

brian-brazil:
:+1: 

The original change aligned the SD page with the targets page, which likely reduces confusion.

juliusv:
@brian-brazil recently I told a client that they should relabel into the `job` label instead of something equivalent but differently named (like `service` or so), since it would give them a better grouping on the `/targets` page. And plenty of others were already relying on that old behavior as well. I'm not sure if this change un-breaks that change sufficiently? (i.e. I would still see the number of targets up/down per scrape config, not per job, etc.)

simonpasquier:
> if I understand correctly, this change would still keep the scrape configs job_name field as the major UI grouping, and only change the sorting within that group, without actually re-introducing a full group for each effective job label?

Yes.

brian-brazil:
> I'm not sure if this change un-breaks that change sufficiently? (i.e. I would still see the number of targets up/down per scrape config, not per job, etc.)

I'd say it's good enough. Users can always build dashboards for this.

juliusv:
Ok. I think it's going to be painful to use Prometheus for some existing users now, but at least this change moves in a right direction 👍

juliusv:
@simonpasquier or rather, by any custom combination of labels, like in the Alertmanager.

zbindenren:
@juliusv I agree. We have about 2000 services discovered with `file_sd_config`. The old behavior was definitively better for us.

PR #4904, opened 2018-11-23
Use `USER 99`instead of 'USER nobody' in Dockerfile

edganiukov:
User "nobody" has `id=99`, for Kubernetes, in order to run container as low privileged user, necessary to specify numeric `USER` in docker image, because Kubernetes can't verify non numeric users.


edganiukov:
```      
 securityContext:
   runAsUser: 99
```
solves the problem

PR #4905, opened 2018-11-23
Cherrypick 4840

simonpasquier:
#4840 was merged in the `release-2.5` branch. Because of generated assets, merging it back in master generates too many conflicts while cherry-picking is straightforward. Note that no v2.5.1 has been released but v2.6.0 is only 2 weeks from now so it might not be worth the effort.

cc @brancz


brancz:
Good catch! lgtm :+1:

PR #4906, opened 2018-11-24
ui: Fix hover when metric has null value.

dcoles:
When a metric has a null value, number formatters like
`humanizeNoSmallPrefix` will throw "Uncaught TypeError: Cannot read
property 'toPrecision' of null".

This is fixed by explicitly checking for `null` and returning the string
"null".

Note: This is usually not seen as rickshaw doesn't show annotations for
null values, but still calls the formatter.

@juliusv for web-ui.


juliusv:
👍 Thanks. Note that this is not part of the main web UI, but the JS library that can be used by custom console templates served by the Prometheus server.

dcoles:
Ah. Sorry about the confusion. Thanks for reviewing anyway.

PR #4590, opened 2018-09-10
discovery: add support for Managed Identity authentication in Azure SD

marratj:
Azure supports Managed Identity authentication for a while now: https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview

This PR is to add support in the Azure SD for Managed Identity auth as an alternative to OAuth with an Azure AD Service Principal.

PR #4908, opened 2018-11-26
add logic to check if an azure VM is deallocated or not

tariq1890:
This PR contains the verified method to confirm if an Azure VM has been deallocated or not through azure-sdk-for-go.

@brian-brazil FYI


discovery/azure/azure.go
```diff
@@ -286,6 +287,15 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 				return
 			}
 
+			// We check if the virtual machine has been deallocated/stopped.
+			// If so, we skip them in service discovery.
+			if strings.EqualFold(vm.PowerStateCode, "PowerState/deallocated") ||
+				strings.EqualFold(vm.PowerStateCode, "PowerState/stopped") {
```
brian-brazil: Can a stopped machine be started again?

For EC2 we still list the stopped machines, as someone may want to monitor them still and leave dropping them to relabelling.

discovery/azure/azure.go
```diff
@@ -286,6 +287,15 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
...
+				level.Debug(d.logger).Log("msg", "Skipping deallocated virtual machine", "machine", vm.Name)
```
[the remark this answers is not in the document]
tariq1890: Sure

tariq1890:
@simonpasquier @brian-brazil Ready for another round of review :)

discovery/azure/azure.go
```diff
@@ -508,3 +509,17 @@ func (client *azureClient) getNetworkInterfaceByID(networkInterfaceID string) (n
 
 	return result, nil
 }
+
+func getPowerStatusFromVMInstanceView(instanceView *compute.VirtualMachineInstanceView) string {
```
simonpasquier: It could rewritten a bit for a better [line of sight](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88):

```go
func getPowerStatusFromVMInstanceView(instanceView *compute.VirtualMachineInstanceView) string {
    if instanceView.Statuses == nil {
        return ""
    }
    var powerStatusCode string
    for _, ivs := range *instanceView.Statuses {
        if strings.HasPrefix(*ivs.Code, "PowerState") {
            powerStatusCode = *ivs.Code
        }
    }
    return powerStatusCode
}
```

discovery/azure/azure.go
```diff
@@ -286,6 +287,14 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 				return
 			}
 
+			// We check if the virtual machine has been deallocated/stopped.
```
simonpasquier: Please remove `stopped`.

brian-brazil:
Is this already in the metadata, or do we need to add it?

discovery/azure/azure.go
tariq1890: Sure!

tariq1890:
@brian-brazil I am not sure I follow. Do you mean by adding Machine PowerState to the metadata?

simonpasquier:
Thanks for the update. The last thing would be to update the documentation at https://github.com/prometheus/prometheus/blob/master/docs/configuration/configuration.md with the new metadata label.

tariq1890:
@simonpasquier Done :)

brian-brazil:
> @brian-brazil I am not sure I follow. Do you mean by adding Machine PowerState to the metadata?

Yes, if it's not already there.

tariq1890:
Looks like we need to retrigger the Circle CI build

docs/configuration/configuration.md
```diff
@@ -267,6 +267,7 @@ The following meta labels are available on targets during relabeling:
 * `__meta_azure_machine_resource_group`: the machine's resource group
 * `__meta_azure_machine_tag_<tagname>`: each tag value of the machine
 * `__meta_azure_machine_scale_set`: the name of the scale set which the vm is part of (this value is only set if you are using a [scale set](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/))
+* `__meta_azure_machine_power_state`: the current power state of the machine
```
simonpasquier: small nit but it would be nice if the list were sorted.

tariq1890: Sure thing :)

simonpasquier:
@tariq1890 thanks a lot for the quick update!

tariq1890:
Can this be merged?

brian-brazil:
Thanks!

discovery/azure/azure.go
```diff
@@ -508,3 +511,16 @@ func (client *azureClient) getNetworkInterfaceByID(networkInterfaceID string) (n
 
 	return result, nil
 }
+
+func getPowerStateFromVMInstanceView(instanceView *compute.VirtualMachineInstanceView) (powerState string) {
```
marratj: While rebasing my PR #4590 I found this function to always panic Prometheus for me with

```
panic: runtime error: invalid memory address or nil pointer dereference                                    
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x14fd62e]
```

It seems that `instanceView.Statuses == nil` is the culprit here, because when checking instead `instanceView == nil` it works.

marratj: This happens not only in my PR branch, but also in master for me :(

tariq1890: Hi @marratj Thanks for reporting this. I am surprised that we have a case of nil instanceView :( . That is not to happen. Could you please email me at tariq.ibrahim@microsoft.com , so that I can investigate that further?

I will , in the mean time, work on adding the nil check here.

tariq1890: PR #4977 created

PR #4910, opened 2018-11-26
cmd/prometheus: add tests for sendAlerts

simonpasquier:
~~This is #3879~~ with tests.

cc @roidelapluie


simonpasquier:
@brian-brazil please take a look.

brian-brazil:
Thinking about this now, I'm not as sure this makes sense. What we send to the AM is about the active alert, not Prometheus-internal state.

roidelapluie:
Yes and it would be a really big breaking change..

simonpasquier:
No problem, I've just kept the tests which are still useful IMO.

PR #4915, opened 2018-11-26
discovery/marathon: fix leaked connections

### LINKED ISSUE #4914: Buggy Discovery behavior
Sorry for the terrible bug report. I don't have more time to spend on this and it's not incredibly critical.

1. Seems the Discovery instance (at least for marathon) does not close response so underlying transport leaks sockets. This was observed on a server trying to scrape an endpoint returning 5xx errors. After some period of time you can observe many sockets in CLOSE_WAIT state

2. After removing the faulty Discoverer from config and reloading the leaked sockets are not cleaned up (why?)

simonpasquier:
Closes #4914

### DIFF
--- a/discovery/marathon/marathon.go
+++ b/discovery/marathon/marathon.go
@@ -352,30 +352,18 @@ func fetchApps(client *http.Client, url string) (*AppList, error) {
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


brian-brazil:
:+1:

PR #4918, opened 2018-11-26
Makefile.common: fix format rule

simonpasquier:
`go fmt` doesn't support the `-mod=vendor` flag.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -115,7 +115,7 @@ common-test:
 .PHONY: common-format
 common-format:
 	@echo ">> formatting code"
-	GO111MODULE=$(GO111MODULE) $(GO) fmt $(GOOPTS) $(pkgs)
+	GO111MODULE=$(GO111MODULE) $(GO) fmt $(pkgs)
 
 .PHONY: common-vet
 common-vet:


Makefile.common
```diff
@@ -115,7 +115,7 @@ common-test:
 .PHONY: common-format
 common-format:
 	@echo ">> formatting code"
-	GO111MODULE=$(GO111MODULE) $(GO) fmt $(GOOPTS) $(pkgs)
+	GO111MODULE=$(GO111MODULE) $(GO) fmt $(pkgs)
```
tariq1890: Question : Can we use `goimports` instead? Since `goimports` does what `go fmt` does and more.

beorn7: The problem with `goimports` is that it doesn't flag ambiguities. It just picks one of the possible imports if there are multiple possibilities to fix a missing import. Thus, its usage is not entirely uncontroversial.

tariq1890: Got it! Thanks for the clarification :)

simonpasquier:
@SuperQ I forgot to tag you but I just saw that you mentioned the issue on IRC.

PR #4924, opened 2018-11-28
Introduce min_shards for remote write to set minimum number of shards.

ryotarai:
After starting Prometheus, number of shards is set to 1 initially and it causes `Remote storage queue full, discarding sample` in heavy-load environment before scaling out shards.
(I have initially implemented `initial_shards` but just after starting Prometheus, it scales down to 1 shard automatically)

This PR introduces `min_shards` parameter that is used as an initial and minimum number of shards.


tomwilkie:
Thanks @ryotarai! Change looks good to me.

Issue #1546, opened 2016-04-11
Avoid HA pair hitting targets at the same time


brian-brazil:
Currently we determine the phase of a target's scrape by it's labels and params. This means that a HA pair will both try to scrape targets at exactly the same time, causing a bigger load spike and potential locking issues.

In order to avoid this we need to offset them somehow. I propose also adding in `external_labels` to the hash, as that should be different between the pair.

juliusv:
AFAICT the phasing is not based on absolute time anyways, but relative to the latest scrape (https://github.com/prometheus/prometheus/blob/master/retrieval/target.go#L127). Given that, won't things distribute enough anyways?

grobie:
I can't confirm this, our scrape times are independent from each other. I believe you'll run into this issue more often if you reload/restart both prometheis at the exact same time though.

OT: What external labels do you usually set? Our pairs are all 100% identical actually, including external_labels.

brian-brazil:
It's relative to absolute time, so that if a target disappears and reappears (e.g. race in SD) it'll keep its offset.

brian-brazil:
> OT: What external labels do you usually set? Our pairs are all 100% identical actually, including external_labels.

That'll cause you some fun with remote storage, and possibly also federation depending on exact what you're doing.

grobie:
@brian-brazil How so? Remote storage does not really exist so far so it's hard to plan for that, and federation will add an instance label which will distinguish the results of the two nodes. What labels do you suggest, something like `role=ha1`/`role=ha2`?

brian-brazil:
> Remote storage does not really exist so far so it's hard to plan for that

There's some aspects that can be planned, from experience you'll want both in storage and distinguished.

> and federation will add an instance label which will distinguish the results of the two nodes.

You should be setting `honor_labels` for federation, otherwise you'll mess up your labels.

Andor:
And how we can call query like that?
http://i.imgur.com/JENmJpE.png

juliusv:
@Andor There's no query-time federation at this point. You'd have to federate the necessary aggregated data into your higher-level Prometheus first and then query the data there.

brian-brazil:
@Andor That's unrelated to this issue, and HA. This isn't possible currently.

brian-brazil:
A simpler way of doing this might be adding the hostname to the hash, as it's unlikely it'll be the same in a HA pair (at least if they're in the same datacenter).

RichiH:
[...]

PR #4926, opened 2018-11-29
bug fixed: wait for interval-now%interval to make sure target will be collected at a fixed point within an interval

F4ncyMooN:
Issue:
each target will not be collect at a fixed position in each interval and this position will be determined randomly according to the prometheus start time.
Senario: 
Suppose interval = 1m, now= 11:00:21, offset = 10s, this target will be collect at 11:00:52(wait 21+10s), but I doubt if it makes sense. But this target should be at collect at 11:01:10.

### DIFF
--- a/scrape/target.go
+++ b/scrape/target.go
@@ -129,7 +129,7 @@ func (t *Target) offset(interval time.Duration) time.Duration {
 	now := time.Now().UnixNano()
 
 	var (
-		base   = now % int64(interval)
+		base   = int64(interval) - now%int64(interval)
 		offset = t.hash() % uint64(interval)
 		next   = base + int64(offset)
 	)


gouthamve:
Sorry, I misunderstood the PR. I think the reason we also include the start time into the equation is so that when we have HA pairs, they don't hit the target at once.

brian-brazil:
Actually we haven't implemented that yet. A given Prometheus should be scraping consistently across restarts. So this is a regression.

F4ncyMooN:
Thanks for your comments.
Here are the reasons that I want to do the change.

I agree that a prometheus instance should crape targets consistently after restart. For me, consistent means a fixed target should only be scraped with a fixed interval. Without this PR one target with 30m interval may be scraped at 00:00:07 and then scraped at 00:05:14 after a prometheus restart, which is NOT consistent.

And with HA pairs, I think secondary prometheus should do exactly the same things with the primary. But scrape time is influenced by the starting time of prometheus. What if secondary prometheus scrape one instance at :17 and primary scrape at :14? This seems like unconsistent to me.

brian-brazil:
This change looks correct to me, and may explain some issues we've seen around users reloading at a high rate. We do want HA pairs to scrape at different times, see #1546

brian-brazil:
Thanks!

nikosmeds:
Can someone confirm the randomness actually works as intended?

My two Prometheus servers are scraping in sync. I've tried stop/start the `prometheus` service on one (multiple times now), and yet they continue to scrape targets at the exact same time.

SuperQ:
@nikosmeds This change does not introduce any randomness. As @brian-brazil said, multi-server offsets are not implemented yet.

nikosmeds:
Sorry guys - I completely misread the first comment.

I understand how this change is desirable (quickly restarting a Prometheus server has caused issues with scrape timing for me in the past), but this change in current state causes issues for me as well.

Some context: running https://github.com/czerwonk/junos_exporter to scrape a number of Juniper switches. Our more basic switches don't handle being scraped by both Prometheus servers simultaneously very well (large CPU spikes, increased scrape times). Using Ansible to configure two Prometheus servers in serial, which causes them to restart/reload service at different intervals (not ideal, but was a working solution). This change stops that behavior, and now I'm unable to get both Prometheus servers to scrape asynchronously.

I'll need to downgrade my Prometheus servers for now.

Would devs support a command line flag which disables this change?
Otherwise I'll be stuck at older Prometheus version while waiting for the randomness to be implemented.

Thanks!

SuperQ:
@nikosmeds One option is to put a strict http caching proxy in front of the junos_exporter. I know some people like @RichiH do this kind of thing to force caching of the snmp_exporter to avoid similar issues. This still lets you get HA from the Prometheus perspective, without the additional load.

PR #4927, opened 2018-11-29
update client-go to v10.0.0 and other k8s deps to v1.13.1

tariq1890:
Updating client-go to v10.0.0. This version should be compatible with kubernetes releases 1.13.*

The following dependencies have been updated

client-go : v10.0.0
k8s.io/api: kubernetes-1.13.1
k8s.io/apimachinery: kubernetes-1.13.1


tariq1890:
@simonpasquier requesting your review on this :)

simonpasquier:
I don't feel strongly about updating the k8s client because the k8s SD uses API resources that are stable. Are you aware if it fixes any known issue we might have? I've spotted https://github.com/kubernetes/kubernetes/pull/67359 (periodically reload InClusterConfig token) that could be interesting but at the same time, it has caused an issue for some consumers (https://github.com/kubernetes/kubernetes/issues/69234). From my understanding, we shouldn't affected by this issue though.

cc @brancz

brancz:
We make heavy use of the in-cluster config so it would be a feature worth having here I think. In terms of API resources I agree with @simonpasquier, but this seems like a api-machinery update that would be worth it (I do also know that in the 1.13 client code there is improved timeout handling for various api-machinery that we are using, so we may just want to wait and update directly to that once 1.13 is out? It's scheduled to be released on 3rd of December.)

simonpasquier:
In any case lets trigger a benchmark run of this PR.

/benchmark

simonpasquier:
At least the benchmark didn't show any regression.

/benchmark cancel

tariq1890:
@simonpasquier @brancz If it's okay with you guys, I can wait until tomorrow for the latest client-go release to update the PR. We can then run the regression and benchmark tests against it. Thoughts?

brancz:
Sounds good to me.

tariq1890:
The same make job passed in my local machine. Odd.

discovery/kubernetes/client_metrics.go
```diff
@@ -136,6 +136,22 @@ var (
...
+			Namespace: workqueueMetricsNamespace,
+			Name:      "unfinished_work_seconds",
+			Help:      "How long an item has remained unfinished in the work queue.",
+		},
+		[]string{"queue_name"},
+	)
+	clientGoWorkqueueLongestRunningProcessorMetricVec = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: workqueueMetricsNamespace,
+			Name:      "longest_running_processor_micro_seconds",
```
simonpasquier: The metric in Prometheus should be in seconds. See how it is done with `clientGoWorkqueueLatencyMetricVec` for instance.

tariq1890: @simonpasquier I am still researching on how to do this for the Gauge vector. ( i think I need to pass in a GaugeFunc somehow). The other places which do the conversion are for Summary metric.

tariq1890: @simonpasquier any tips on this? I am still looking for a way to do this conversion before setting it to Gauge

simonpasquier: I haven't looked into the details. I'll do soon.

sipian:
@simonpasquier 
Looks like benchmarking did not stop for this. Could you comment cancel again?

discovery/kubernetes/client_metrics.go
simonpasquier: You could create a type that would satisfy the `workqueue.SettableGaugeMetric` interface and do the conversion from microseconds to seconds like this:

```diff
diff --git a/discovery/kubernetes/client_metrics.go b/discovery/kubernetes/client_metrics.go
index 19dc7023..6d76a74d 100644
--- a/discovery/kubernetes/client_metrics.go
+++ b/discovery/kubernetes/client_metrics.go
@@ -258,8 +258,17 @@ func (f *clientGoWorkqueueMetricsProvider) NewWorkDurationMetric(name string) wo
                metric.Observe(v / 1e6)
        })
 }
+
+type settableGaugeFunc func(float64)
+
+func (f settableGaugeFunc) Set(value float64) {
+       f(value)
+}
 func (f *clientGoWorkqueueMetricsProvider) NewUnfinishedWorkSecondsMetric(name string) workqueue.SettableGaugeMetric {
-       return clientGoWorkqueueUnfinishedWorkSecondsMetricVec.WithLabelValues(name)
+       m := clientGoWorkqueueUnfinishedWorkSecondsMetricVec.WithLabelValues(name)
+       return settableGaugeFunc(func(v float64) {
+               m.Set(v / 1e6)
+       })
 }
 func (f *clientGoWorkqueueMetricsProvider) NewLongestRunningProcessorMicrosecondsMetric(name string) workqueue.SettableGaugeMetric {
        return clientGoWorkqueueLongestRunningProcessorMetricVec.WithLabelValues(name)
```

tariq1890: Yes. That was what had come to mind. I was wondering if that was okay. Is there a particular location I should declare this type at (similar types seemed to be declared in the client_golang repo)? Or is it to declare the type in `kubernetes\client_metrics.go` ?

simonpasquier: Having it in `client_metrics.go` is ok.

tariq1890:
PR feedback addressed and rebased

discovery/kubernetes/client_metrics.go
tariq1890: Thanks:)

tariq1890:
I have updated `k8s.io/api` to the latest patch release. Ready for another round of review.

discovery/kubernetes/client_metrics.go
simonpasquier: should be `longest_running_processor_seconds`

simonpasquier:
Thanks @tariq1890! I'll give run another benchmark once we're settled.

tariq1890:
Something up with the Travis-CI ?

simonpasquier:
Probably an issue with Travis, I've restarted the job.

tariq1890:
@simonpasquier I have rebased this PR using `go mod` along with `go version 1.11.4`. The only delta was in the `go.sum` after rebase.

discovery/kubernetes/client_metrics.go
```diff
@@ -136,6 +136,22 @@ var (
...
+			Name:      "unfinished_work_seconds",
+			Help:      "How long an item has remained unfinished in the work queue.",
+		},
+		[]string{"queue_name"},
+	)
+	clientGoWorkqueueLongestRunningProcessorMetricVec = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: workqueueMetricsNamespace,
+			Name:      "longest_running_processor_seconds",
+			Help:      "The duration (in microseconds) of the longest running processor in the work queue.",
```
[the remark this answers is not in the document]
tariq1890: Ahh my bad. Sorry to have missed this :(

krasi-georgiev:
/benchmark cancel

It should have enough data so cancelling to keep the cost down.

tariq1890:
Can this PR be merged?

simonpasquier:
LGTM. Note that my availability is limited until beginning of next year :smirk:

simonpasquier:
As Kubernetes v1.13 switched from glog to klog, we need to update the current replacement. I've created [klog-gokit](https://github.com/simonpasquier/klog-gokit) which mimics what [glog-gokit](https://github.com/kubermatic/glog-gokit) did.

See https://github.com/simonpasquier/prometheus/commit/0104baf40bce92c6df80a2d44a9d45f616d96bce for what needs to be updated.

tariq1890:
@simonpasquier Done! thanks :)

tariq1890:
image

brancz:
lgtm thanks for bearing with us

simonpasquier:
Thanks a lot @tariq1890

erikh:
Yes. Thanks!!!

PR #4928, opened 2018-11-29
Add sequential rule evaluation in a group for remote storage backend

garrettlish:
Sometimes remote write service is synchronized. However, remote storage queue manager in prometheus doesn't wait remote write response and run next rule in the group immediately which means that we cannot ensure sequential rule evaluation in a group if it is remote storage backend although the remote write service is sychronized. This PR will make sure rule evaluation for one group wait the remote write response before evaluating next rule in the group.

Signed-off-by: Garrett Li

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -474,6 +474,9 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
 			} else {
 				g.seriesInPreviousEval[i] = seriesReturned
 			}
+
+			// commit sync to wait remote storage call completed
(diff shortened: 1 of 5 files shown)


brian-brazil:
Thank you for your PR, I don't think this makes sense for Prometheus. Remote storage is an optional extra which can fail, and shouldn't delay things when it does so. The local storage is all you should really be depending on. In addition this will not work with the new WAL-based work being done.

garrettlish:
ic, thanks @brian-brazil, close it

Issue #3236, opened 2017-10-03
How to get Kubernetes log messages from inside Prometheus?


bboreham:
`prometheus/discovery/kubernetes` calls Kubernetes code; Kubernetes code logs via `glog`; there doesn't seem to be a way to set the glog level inside Prometheus.

glog does set up command-line options like `-v` via `flag.Var()`, but because nobody is calling `flag.Parse()`  they don't help.

fabxc:
In the 2.0 branch we are overwriting the error handler and handle the logging of errors from the k8s code base ourselves: https://github.com/prometheus/prometheus/blob/dev-2.0/cmd/prometheus/main.go#L205-L212

bboreham:
That appears to be yet another thing.

I'm talking about for instance https://github.com/prometheus/prometheus/blob/dev-2.0/vendor/k8s.io/client-go/tools/cache/reflector.go#L236

fabxc:
I see, ultimately, this is a severe mis-design of Kubernetes libraries in general and we currently don't and cannot provide flags for a basically arbitrarily chosen and unswappable logging library somewhere in our dependencies.

Why are you interested in seeing those debug logs?

bboreham:
Got an error from discovery; didn't understand it; wanted more context what the k8s library was doing. 

More specifically it was a customer running the program so I preferred to say "run with `-v=8` and send me the full log" - even though most of the information is useless it is good to get it all in one hit rather than go back 20 times "ok, now try this...".

fabxc:
Yes, that's right. It would be great to just show those messages when debug logging is enabled but with k8s' current approach I don't see a sane way for us to do it =/

bboreham:
Request to fix Kubernetes http://github.com/kubernetes/kubernetes/issues/61006

simonpasquier:
`--log.level=debug` displays the Kubernetes logs since #4931.

PR #4931, opened 2018-11-29
cmd/prometheus: replace glog by glog-gokit

### LINKED ISSUE #4795: Prometheus crashes if it can't write logs to /tmp
## Bug Report
**What did you do?**

We copy the Prometheus binary into a from-scratch Docker container and run it as a non-root user to reduce our footprint and threat surface. This is a practice we follow with all applications we run in our Kubernetes environment. Recently we upgraded to prometheus v2.4.2 from v2.3.0, where in both cases we were running in containers using this method.

Prometheus should run fine, as v2.3.0 seemed to.

Prometheus is repeatedly crashing with this error:

```
W1026 14:54:21.157256       1 reflector.go:341] github.com/prometheus/prometheus/discovery/kubernetes/kubernetes.go:306: watch of *v1.Pod ended with: too old resource version: 19430593 (19430983)
log: exiting because of error: log: cannot create log: open /tmp/prometheus.prometheus-75f465d4d4-hjztt.unknownuser.log.WARNING.20181026-145421.1: no such file or directory
```

This seems to be a glog error message which occurs when glog can't write a logfile. However, I don't want glog writing to any files at all; it should write all logs to stderr, since we are running in a Docker container and want to delegate log management to Docker. However, I see no way to configure Prometheus or glog to not log to files.

glog apparently has a `--logtostderr` flag, but it's not being exposed by Prometheus.

Note that the crash is ALWAYS preceded by a warning log from Kubernetes service discovery, so this may be the source of the bug.

This error was also reported on coreos/prometheus-operator here: https://github.com/coreos/prometheus-operator/issues/1932 Notice that the crash in that report is ALSO preceded by a warning from Kubernetes service discovery.

**Environment**

* System information: no `uname` in a scratch container.

* Prometheus version:

```
prometheus, version 2.4.2 (branch: HEAD, revision: c305ffaa092e94e9d2dbbddf8226c4813b1190a0)
  build user:       root@dcde2b74c858
  build date:       20180921-07:22:29
  go version:       go1.10.3
```

* Logs:
```
W1026 14:54:21.157256       1 reflector.go:341] github.com/prometheus/prometheus/discovery/kubernetes/kubernetes.go:306: watch of *v1.Pod ended with: too old resource version: 19430593 (19430983)
log: exiting because of error: log: cannot create log: open /tmp/prometheus.prometheus-75f465d4d4-hjztt.unknownuser.log.WARNING.20181026-145421.1: no such file or directory
```

simonpasquier:
Closes #4795 #3236

cc @metalmatze


brancz:
Out of curiosity, can we test this somewhere just to make sure the "6" verbosity is not over-doing it?

simonpasquier:
@brancz yep I meant to provide some snippets of logs anyhow.

simonpasquier:
Prometheus logs with level 6:

https://paste.fedoraproject.org/paste/ZbTSh5FuyAaoKSQ7M7EaFg

Prometheus logs with level 7 (I didn't redacted the tokens since this cluster is already gone):

https://paste.fedoraproject.org/paste/iKb7LBrFJm1IeA9kjVutZQ

simonpasquier:
@brancz friendly ping

brancz:
Did we have this verbosity level previously? I don't recall seeing messages like that.

Generally looks good to me though.

brancz:
Oooohhhh... I see now, yes that makes perfect sense. In that case lgtm :+1:

PR #4947, opened 2018-12-04
Fix variable 'notifier' collides with imported package name.

aixeshunter:
Fix variable 'notifier' collides with imported package name 'github.com/prometheus/prometheus/notifier', changed to 'notifierManager'.


brian-brazil:
Thanks!

PR #4953, opened 2018-12-04
update tsdb to version v0.3.0

krasi-georgiev:


simonpasquier:
LGTM

brian-brazil:
This seems to be changing a lot more than the tsdb. Was that intended?

simonpasquier:
Yes this is because of how Go modules work. tsdb v0.3.0 is the first version with `go.mod` so updating to it bumps the dependencies shared by prometheus to match the minimum requirements of tsdb as defined in https://github.com/prometheus/tsdb/blob/master/go.mod.

The main updates are:
* github.com/go-kit/kit updated from an older (untagged) revision to the latest stable version (v0.8.0).
* github.com/gogo/protobuf updated from an older (untagged) revision to v1.1.1.
* github.com/prometheus/client_golang updated from an older (untagged but relatively recent) revision to v0.9.1.
* gopkg.in/alecthomas/kingpin.v2 updated from v2.2.5 to v2.2.6.

IMHO those should be relatively safe updates.

PR #4957, opened 2018-12-04
update remote write path proto so that Labels/Timeseries can't be nil

cstyan:
cc: @tomwilkie 

pulling this out, currently part of #4588 
also update the genproto script to not call `go get` since that updates go modules (cc @simonpasquier )


tomwilkie:
> also update the genproto script to not call go get since that updates go modules (cc @simonpasquier )

How are people supposed to know what tools to install?  We want building the protos to be easy.  Perhaps `go install` has the behaviour you want?

LGTM otherwise.

simonpasquier:
I'm looking at the `genproto.sh` script as it should be able to leverage Go modules and get rid of the `go get ...`/`git reset ...` dance.

cstyan:
@simonpasquier @tomwilkie confirmed that with a clean repo and removing the `go get/git reset` bits of the script, `genproto` will still run properly. Assuming it's pulling the libraries it needs from vendor, and those are the same versions listed in `go mod` (gogo proto 1.1.1, and grpc-gateway at commit `e4b8a938efae14de11fd97311e873e989896348c`) then all is good.

The only other difference is that the script is currently pulling gogo proto `971cbfd2e72b513a28c74af7462aee0800248d69` which is between the 1.1.1 and 1.2.0 release. I can update that in the go modules to 1.2.0?

tomwilkie:
> confirmed that with a clean repo and removing the go get/git reset bits of the script, genproto will still run properly

This is because you already have the tools installed in you $GOPATH/bin.  It will fail if you don't.  Also, I don't think the `genproto` is capable of pulling from vendor - its not a go tool.

These lines are needed, can you see if `go install` has the right effect and doesn't update the go modules?

cstyan:
:man_facepalming: I totally blanked that these were executables @tomwilkie 

Not sure what the right fix is here. We shouldn't assume people already have those packages in their `GOPATH`. We could just `git checkout go.mod go.sum` at the end of the script? \

That works, but simply commenting out the `go get ...` lines does not. We end up with the right version of the grpc-gateway but the build still fails.

Another option set `GO111MODULE=off` in `genproto.sh`. @simonpasquier any opinions? I'm still not up to date on modules best practices.

simonpasquier:
My plan is to update the `genproto.sh` script to rely only on Go modules. This way the packages used by the code generators and the runtime (gRPC gateway + client) would be the same versions. I'll try to send out something today.

cstyan:
@simonpasquier any progress? want me to have another go at it?

simonpasquier:
@cstyan I've rebased my changes on master and currently testing it. It should be ready soon.

simonpasquier:
See https://github.com/prometheus/prometheus/pull/5075

cstyan:
bump @tomwilkie

tomwilkie:
LGTM!

bboreham:
Pulling this comment from the original PR since it's not obvious here why this change was made.

> changes to the proto definitions, made it so that certain fields can't be null which results in `[]type` rather than `[]*type` and therefore less object allocations.

PR #4445, opened 2018-08-01
Add a side Dockerfile.multi-stage which will build prometheus inside …

sylr:
This will allow anyone to build prometheus easily without any building environment.

Combined this with a quay.io account and you'll have ready to test images at your disposal
for all your development branches.

Signed-off-by: Sylvain Rabot <s.rabot@lectra.com>

PR #4959, opened 2018-12-05
Add go stage for building inside docker

sanchezpaco:
Changes:

- Add new golang stage to Dockerfile in order to build Prometheus in docker

Update: Sorry the thing about deleting the branch/PR, I had a pebcac moment


sanchezpaco:
I just saw this: #4445

PR #4961, opened 2018-12-05
rule manager: Moved metric registration to custom registerer which is already available.

bwplotka:
Hi :wave:  

This is to start discussion. We started notion of custom registerer in this package for some metrics calculated by manager `Collect` method itself, but we did not go full way with out other metrics.

Since in Thanos we don't use global registry, can we use Manager registry to register those as well?

TODO in this PR, move metrics itself to some metric struct owned by Manager itself.
  
### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -110,14 +110,6 @@ var (
 	)
 )
 
-func init() {
-	prometheus.MustRegister(iterationDuration)
-	prometheus.MustRegister(iterationsScheduled)
-	prometheus.MustRegister(iterationsMissed)
-	prometheus.MustRegister(evalFailures)
-	prometheus.MustRegister(evalDuration)
-}
-
 // QueryFunc processes PromQL queries.
 type QueryFunc func(ctx context.Context, q string, t time.Time) (promql.Vector, error)
 
@@ -646,6 +638,11 @@ func NewManager(o *ManagerOptions) *Manager {
 		logger: o.Logger,
 	}
 	if o.Registerer != nil {
+		o.Registerer.MustRegister(iterationDuration)
+		o.Registerer.MustRegister(iterationsScheduled)
+		o.Registerer.MustRegister(iterationsMissed)
+		o.Registerer.MustRegister(evalFailures)
+		o.Registerer.MustRegister(evalDuration)
 		o.Registerer.MustRegister(m)
 	}
 	return m
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -53,55 +53,6 @@ const (
 const namespace = "prometheus"
 
 var (
-	evalDuration = prometheus.NewSummary(
-		prometheus.SummaryOpts{
-			Namespace: namespace,
-			Name:      "rule_evaluation_duration_seconds",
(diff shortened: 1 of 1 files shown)


brancz:
> TODO in this PR, move metrics itself to some metric struct owned by Manager itself.

What benefit does that have?

Otherwise this looks good to me.

brian-brazil:
> What benefit does that have?

The PR would not be correct otherwise, as you'd be sharing metrics across managers.

bwplotka:
> The PR would not be correct otherwise, as you'd be sharing metrics across managers.

exactly.

However creating multiple managers with same register will still return error (panic) from register, but I think that's expected, right?

brian-brazil:
Yes, that's expected.

brancz:
Whops. I'm so used to only having the descriptions as vars these days, that I read this code the way you meant it to be. In that case your suggestion makes sense, let's change it to do that.

brian-brazil:
You still need to make that change.

bwplotka:
I am aware, thanks for ping - just moving all of those to `prometheus.NewDesc(` is quite tedious (: lot's of setters and getters needed for `RuleGroup` wonder if we can avoid this. Probably not.

rules/manager.go
```diff
@@ -110,12 +61,85 @@ var (
...
-	prometheus.MustRegister(evalFailures)
-	prometheus.MustRegister(evalDuration)
+type groupMetrics struct {
+	evalDuration        prometheus.Summary
+	evalFailures        prometheus.Counter
+	evalTotal           prometheus.Counter
+	iterationDuration   prometheus.Summary
+	iterationsMissed    prometheus.Counter
+	iterationsScheduled prometheus.Counter
+	lastEvaluation      *prometheus.GaugeVec
```
brian-brazil: Why are these pointers and the others not?

rules/manager.go
```diff
@@ -307,34 +336,6 @@ func (g *Group) hash() uint64 {
 	return l.Hash()
 }
 
-// GetEvaluationDuration returns the time in seconds it took to evaluate the rule group.
-func (g *Group) GetEvaluationDuration() time.Duration {
```
brian-brazil: These are used in the UI

rules/manager.go
bwplotka: This is struct, rest are interfaces.

rules/manager.go
bwplotka: Hm... why CI is green then?

brian-brazil: CI doesn't test those.

bwplotka:
Ok, reverted interface change, but things I don't really like and most likely needs to be fixed in this PR:
* Group is used like rule but not. No one ever use it as `Rule` interface, however it tries to implement all boilerplate methods. It is super misleading.
* No one knows what methods UI uses, so you never know until you manually run UI and check - I think some unit test for compatibility against `rules.html` template is must have.

brian-brazil:
Thanks!

> I think some unit test for compatibility against rules.html template is must have.

That's been the case for a while.

PR #4962, opened 2018-12-05
test chunk pool

krasi-georgiev:
testing a tsdb PR

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -129,4 +129,7 @@ require (
 	launchpad.net/gocheck v0.0.0-20140225173054-000000000087 // indirect
 )
 
-replace github.com/golang/glog => github.com/kubermatic/glog-gokit v0.0.0-20181129151237-8ab7e4c2d352
+replace (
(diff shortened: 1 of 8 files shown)


csmarchbanks:
Looking at the graphs it doesn't look like prombench is doing any queries?

krasi-georgiev:
hm yeah the the dashboards don't show any query timings but I double checked and the querier does indeed show some logs that it runs different queries.

@csmarchbanks maybe run some heavy ones manually on both as well.
PR - prombench.prometheus.io/4962/prometheus-pr
master - prombench.prometheus.io/4962/prometheus-release

@sipian can you see why the the query dashboards are empty? maybe changed metrics names?
http://prombench.prometheus.io/grafana/d/7gmLoNDmz/prombench?orgId=1&var-RuleGroup=All&var-pr-number=4962&from=now-3h&to=now

sipian:
@krasi-georgiev I seem to have found the issue. Testing a fix now.

krasi-georgiev:
will restart after the queerer is fixed

/benchmark cancel

Issue #4944, opened 2018-12-03
Promtool duplicates error message if multiple alertnames are evaluated at same time


mwalters-workmarket:
Minimal example:

rules.yaml:
```
groups:
  - name: MyRules
    rules:
      - alert: AnAlert
        expr: abc{} > 2
        for: 1m
      - alert: AnotherAlert
        expr: def{} > 2
```

test.yaml:
```
rule_files:
  - rules.yaml
evaluation_interval: 1m
tests:
  - interval: 1m
    input_series:
      - series: 'abc{}'
        values: '1'
      - series: 'def{}'
        values: '3'
    alert_rule_test:
      - alertname: AnAlert
        eval_time: 1m
      - alertname: AnotherAlert
        eval_time: 1m
```

when I execute `promtool test rules test.yaml`
I expect to see output:

```
Unit Testing:  test.yaml
  FAILED:
    alertname:AnotherAlert, time:1m0s, 
        exp:"[]", 
        got:"[Labels:{alertname=\"AnotherAlert\"} Annotations:{}]"
```

But the actual output is:

```
Unit Testing:  test.yaml
  FAILED:
    alertname:AnotherAlert, time:1m0s, 
        exp:"[]", 
        got:"[Labels:{alertname=\"AnotherAlert\"} Annotations:{}]"
    alertname:AnotherAlert, time:1m0s, 
        exp:"[]", 
        got:"[Labels:{alertname=\"AnotherAlert\"} Annotations:{}]"

```

It looks like the issue is occurring because in the test file, the eval_time for both rules is the same (1m)

brian-brazil:
@codesome

codesome:
Thanks @mwalters-workmarket for reporting this. Will look at it.

PR #4964, opened 2018-12-06
Avoid duplicate tests for alert unit tests

codesome:
Closes https://github.com/prometheus/prometheus/issues/4944

With this PR, alert unit tests will be run only once at a timestamp. Before this PR if multiple alert unit tests had same timestamp, the tests were duplicated for that timestamp.

### DIFF
--- a/cmd/promtool/unittest.go
+++ b/cmd/promtool/unittest.go
@@ -160,14 +160,14 @@ func (tg *testGroup) test(mint, maxt time.Time, evalInterval time.Duration, grou
 	// All this preparation is so that we can test alerts as we evaluate the rules.
 	// This avoids storing them in memory, as the number of evals might be high.
 
-	// All the `eval_time` for which we have unit tests.
-	var alertEvalTimes []time.Duration
+	// All the `eval_time` for which we have unit tests for alerts.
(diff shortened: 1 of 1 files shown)


cmd/promtool/unittest.go
```diff
@@ -160,14 +160,14 @@ func (tg *testGroup) test(mint, maxt time.Time, evalInterval time.Duration, grou
 	// All this preparation is so that we can test alerts as we evaluate the rules.
 	// This avoids storing them in memory, as the number of evals might be high.
 
-	// All the `eval_time` for which we have unit tests.
-	var alertEvalTimes []time.Duration
+	// All the `eval_time` for which we have unit tests for alerts.
+	alertEvalTimesMap := make(map[time.Duration]struct{})
```
brian-brazil: You don't need the make here

codesome: It panics if I don't have that. Without `make` its a `nil` map.

brian-brazil: `map[time.Duration]struct{}{}` should work.

codesome: I changed it.

codesome:
@brian-brazil is this good to be merged?

brian-brazil:
Thanks!

PR #4969, opened 2018-12-07
Add Azure tenant and subscription ID labels

fabxc:
This copies subscription and tenant ID e discovered labels of all Azure targets.
We do the same for project ID's in the GCE discovery. So this should be a consistent change.

@simonpasquier, since this is a minor change it would be great to still get this into 2.6.

### DIFF
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -259,6 +259,8 @@ Azure SD configurations allow retrieving scrape targets from Azure VMs.
 
 The following meta labels are available on targets during relabeling:
 
+* `__meta_azure_subscription_id`: the subscription ID
+* `__meta_azure_tenant_id`: the tenant ID
 * `__meta_azure_machine_id`: the machine ID
 * `__meta_azure_machine_location`: the location the machine runs in
 * `__meta_azure_machine_name`: the machine name
--- a/discovery/azure/azure.go
+++ b/discovery/azure/azure.go
@@ -26,19 +26,19 @@ import (
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/Azure/go-autorest/autorest/azure"
-
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
 	config_util "github.com/prometheus/common/config"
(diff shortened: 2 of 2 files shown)


simonpasquier:
@fabxc I'd rather stick with the release documentation which says that only critical bug fixes are accepted once the first release candidate has been issued.

fabxc:
Fine.

But why do the new release docs specify that pre-releases are cut from master, which remains frozen for the entire time?
Sometimes our RCs last for 2-4 weeks. That doesn't seem like a reasonable time frame to let feature PRs pile up and result in a lot of merge conflicts.

beorn7:
I guess the idea is that the RC period usually wouldn't last long. See https://github.com/prometheus/prometheus/blob/master/RELEASE.md#release-shepard-responsibilities . But I, too, was wondering why the `release-<major>.<minor>` branch is only created once the final release is out. Why not create that branch the moment the release candidate is out? Then we don't have to freeze master at all. (This is not meant passive-aggressively but as a real question. I wasn't present at the dev summit where all of this was discussed and decided. I'm sure there is a rationale behind this that I just don't know.)

simonpasquier:
> Why not create that branch the moment the release candidate is out?

I asked myself the same question during 2.5.0 RC before checking the release documentation. I don't recall the exact rationale from the dev summit to be honest and the process can definitely be improved as we do more frequent releases.

Additional details here: https://docs.google.com/document/d/1RQgGTwAgkdtZu7DelMwqlmDqmpSH1e8Y6IcL3fD2WQI/edit#

tariq1890:
Lgtm

tariq1890:
Now that 2.6.0 has been released, can this be merged?

docs/configuration/configuration.md
```diff
@@ -259,6 +259,8 @@ Azure SD configurations allow retrieving scrape targets from Azure VMs.
 
 The following meta labels are available on targets during relabeling:
 
+* `__meta_azure_subscription_id`: the subscription ID
+* `__meta_azure_tenant_id`: the tenant ID
```
brian-brazil: Keep it sorted

juliusv:
:+1: but CircleCI tests are currently failing again due to a Go modules checksum mismatch that is probably unrelated?

simonpasquier:
You'd need to rebase on master to fix the checksum issue.

PR #4976, opened 2018-12-07
Fix Dockerfile entrypoint

SuperQ:
Fix WORKDIR and use symlink to setup working defaults for config flags.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,13 +10,10 @@ COPY consoles/                              /usr/share/prometheus/consoles/
 RUN ln -s /usr/share/prometheus/console_libraries /usr/share/prometheus/consoles/ /etc/prometheus/
 RUN mkdir -p /prometheus && \
     chown -R nobody:nogroup etc/prometheus /prometheus
+RUN ln -s /prometheus /etc/prometheus/data
 
 USER       nobody
 EXPOSE     9090
 VOLUME     [ "/prometheus" ]
-WORKDIR    /prometheus
-ENTRYPOINT [ "/bin/prometheus", \
-             "--storage.tsdb.path=/prometheus", \
-             "--web.console.libraries=/etc/prometheus/console_libraries", \
-             "--web.console.templates=/etc/prometheus/consoles", \
-             "--config.file=/etc/prometheus/prometheus.yml" ]
+WORKDIR    /etc/prometheus
+ENTRYPOINT [ "/bin/prometheus" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,10 +7,10 @@ COPY documentation/examples/prometheus.yml  /etc/prometheus/prometheus.yml
 COPY console_libraries/                     /usr/share/prometheus/console_libraries/
 COPY consoles/                              /usr/share/prometheus/consoles/
 
-RUN ln -s /usr/share/prometheus/console_libraries /usr/share/prometheus/consoles/ /etc/prometheus/
-RUN mkdir -p /prometheus && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -10,13 +10,13 @@ COPY consoles/                              /usr/share/prometheus/consoles/
 RUN ln -s /usr/share/prometheus/console_libraries /usr/share/prometheus/consoles/ /etc/prometheus/
 RUN mkdir -p /prometheus && \
     chown -R nobody:nogroup etc/prometheus /prometheus
+RUN ln -s /prometheus /data
+RUN ln -s /etc/prometheus/prometheus.yml /prometheus.yml
+RUN ln -s /etc/prometheus/console_libraries /console_libraries
```
simonpasquier: `/usr/share/prometheus/console_libraries` rather?

SuperQ: I thought we were intending to move those to `/etc/prometheus/...`

simonpasquier: Sorry I'm confused

SuperQ: We already setup a symlinks from 1/usr/share/prometheus/...` to `/etc/prometheus/...` above on line 10.

SuperQ:
cc @yesac

Dockerfile
```diff
@@ -10,13 +10,10 @@ COPY consoles/                              /usr/share/prometheus/consoles/
 RUN ln -s /usr/share/prometheus/console_libraries /usr/share/prometheus/consoles/ /etc/prometheus/
 RUN mkdir -p /prometheus && \
     chown -R nobody:nogroup etc/prometheus /prometheus
+RUN ln -s /prometheus /etc/prometheus/data
```
squeed: Every docker `RUN` command creates a new layer; you should combine this in to a single command.

simonpasquier:
@SuperQ I've tested quickly with the Prometheus Helm chart and the image from this PR deployed just fine :+1:

Dockerfile
SuperQ: O.o WTF, Thanks docker.

brancz:
lgtm :+1:

roman-vynar:
This breaks the setup for those having data dir and configs in the same folder.
For example:
```
# ll /data/prometheus/
total 64
drwxr-x---  3 nobody nogroup    58 Dec 18 10:52 ./
drwxr-xr-x  7 root   root      107 Jul 13 09:23 ../
-rw-r-----  1 root   nogroup 11579 Dec 11 15:18 alerts.yml
drwxr-x--- 16 nobody nogroup  4096 Dec 18 11:00 data/
-rw-r-----  1 root   nogroup 48003 Dec 11 15:14 prometheus.yml
```
```
    command: >
      --web.listen-address=127.0.0.1:9090
      --storage.tsdb.retention=30d
...
    volumes:
      - /data/prometheus:/prometheus
```

The following change to Dockerfile    "ln -s /prometheus /etc/prometheus/data" puts everything into /etc/prometheus/data along with the configs resulting in a missing config and writing wal and lock file to the local /data/prometheus/ instead of /data/prometheus/data.

So this is not `[ENHANCEMENT] Remove default flags from the container's entrypoint. #4976` but a broken change :-/

simonpasquier:
@roman-vynar sorry to hear that it broke your setup. If we had the report before that would have been included as a `[CHANGE]` obviously.

xanonid:
This would have broken also our setup. Would it be possible to adapt the changelog and add a "incompatible change" notice to https://github.com/prometheus/prometheus/releases ?

Starefossen:
This severely broke our setup! Please consider this a braking change in the future!

PR #4980, opened 2018-12-09
IMPORTANT: Revert "add logic to check if an azure VM is deallocated or not (#4908)"

tariq1890:
This is required because InstanceView struct is only exposed and never populated. This is an issue the azure-go-sdk vendor of prometheus.

The solution to get around this problem would be to update azure-go-sdk dependencies and then attempt to retrieve the VM powerstate via InstanceView.

@simonpasquier @brian-brazil FYI. My apologies for this oversight and I have created this PR to make the required amends.

The v2.6.0 release notes will have to be amended as well to reflect this reversion.


tariq1890:
Is this good to merge?

Issue #3697, opened 2018-01-17
better PromQL syntax documentation


rogpeppe:
I recently tried to form a query that failed with a parse error, but I'm not sure why it did.
As the query language documentation is mostly by example, I have no clear idea what the syntax of the query language actually is.

My specific example was that it's not clear what syntactic role an instance vector selector takes and hence how it can be applied. I had a vector expression `rate(jem_monitor_errors_count[1m])` that I wanted only one component of, so I thought that `rate(jem_monitor_errors_count[1m]){instance="instance-name"}` would work, but I get `parse error at char 37: could not parse remaining input \"{instance=\\\"in\"..."`, which I find odd because `jem_monitor_errors_count{instance="instance-name"}` works fine.

It would be great to have the PromQL grammar available as a reference doc (ideally in some kind of BNF notation), so people like me don't come bugging the prometheus issues with probably-stupid questions like this.

PS I realised that I could fix the issue by applying the instance vector selector to the inner expression, but I think my point remains - the syntax is not clearly documented so we have to guess.

deeTEEcee:
I figure I'd just add to this for more syntax/documentation-related issues. So we had just started using prometheus and I added a few alerts, one of the using multiple conditoins with the AND keyword. Then we started having magical alert issues and I wasn't sure why. After analyzing graphs for days, I realized that queries were mixed with the labels (but I had wrongfully assumed that Prometheus alerts would isolate the labels) So then I learned about the "on" keyword (if you google for this, you won't find it anywhere.)

Related page: https://prometheus.io/docs/prometheus/latest/querying/operators/
My suggestion is to improve that page and in my case, would it make sense to put the keywords `ignoring` and `on` under the "Logical/set binary operators" section? Because I figure that's where it's important.

brian-brazil:
Yip, looks like we forgot to document many-to-many matching.

douardda:
I agree that a proper syntax definition is missing. In my case, reading the doc, I see no reason why 
```(a + b)[5m]``` fails with a syntax error. Reading the [unit tests code](https://github.com/prometheus/prometheus/blob/458455245fd38fa5df73d13d74e2fa0b51d4cbb4/promql/parse_test.go#L1056), this case is explicitly a fail but there is no way I could figure it out from the doc. As a side note, I don't understand how to do such a thing as building a range vector from an instant vector resulting from a binary op.

brian-brazil:
[...]

PR #4989, opened 2018-12-11
more explicit doc for the range vector syntax

douardda:
Be more explicit on which object the range vector 'constuctor' applies to

a very partial response to #3697


brian-brazil:
Thanks for your PR, this is already covered in the first paragraph of this section so I don't think we should duplicate that.

codesome:
We are closing this for the above reason mentioned.

PR #4998, opened 2018-12-13
Makefile.common: fix promu download path for arm32

pgier:
See also https://github.com/prometheus/node_exporter/issues/1194


simonpasquier:
LGTM

PR #4999, opened 2018-12-13
Add context parameter and return error for ruleRetriever interface

jtlisi:
This work does not directly relate to prometheus. It allows the prometheus API to be reused in the [cortex project](https://github.com/cortexproject/cortex) for the rule evaluation component. If this is acceptable that would be great, and make matching the prometheus API much easier. Let me know if there is anything I can do to help make this possible.


brian-brazil:
I'm not sure this makes sense in Prometheus terms, retrieving rules is always local and fast.

PR #5007, opened 2018-12-17
Fix go.sum after Go1.11.4 bump

simonpasquier:
The latest Go version (1.11.4) breaks the build process because it changes the checksum of the `github.com/hashicorp/go-rootcerts` and `github.com/coreos/etcd` modules. See https://github.com/golang/go/issues/29278#issuecomment-447537558 for the details.

As the CircleCI Go image hasn't updated to 1.11.4 yet, some tests have to be skipped in CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,9 @@ jobs:
     steps:
     - checkout
     - run: make promu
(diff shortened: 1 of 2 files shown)


brian-brazil:
:+1:

aixeshunter:
:+1:

Issue #3043, opened 2017-08-09
External Labels not available in alert Annotations


byrnedo:
[NOTICE]: <> (If your question is around usage and not a bug in Prometheus please use: https://groups.google.com/forum/#!forum/prometheus-users)

**What did you do?**

Defined an external label in prometheus config:
```
global:
  scrape_interval: 15s
  scrape_timeout: 10s
  evaluation_interval: 15s
  external_labels:
    monitor: acme-logs-dev
```
In a rule I had: 
```
ALERT container_eating_memory
  IF sum(container_memory_rss{container_label_com_docker_swarm_task_name=~".+"}) BY (instance, name) > 2500000000
  FOR 5m
  ANNOTATIONS {
    description="{{ $labels.container_label_com_docker_swarm_task_name }} is eating up a LOT of memory. Memory consumption of {{ $labels.container_label_com_docker_swarm_task_name }} is at {{ humanize $value}}.", 
    summary="{{$labels.monitor}} - HIGH MEMORY USAGE WARNING: TASK '{{ $labels.container_label_com_docker_swarm_task_name }}' on '{{ $labels.instance }}'"}
```

**What did you expect to see?**
The templated alert printing the monitor label.

But the alert in both prometheus alert view and the alert that ended up in slack were both missing the external label

**What did you see instead? Under which circumstances?**

No label listed in prometheus alerts page and empty string in slack

**Environment**

* System information:

* Prometheus version: 

	1.7.1

* Alertmanager version:

	0.8.0

* Prometheus configuration file:
```
global:
  scrape_interval: 15s
  scrape_timeout: 10s
  evaluation_interval: 15s
  external_labels:
    monitor: acme-logs-dev
alerting:
  alertmanagers:
  - static_configs:
    - targets:
      - alertmanager.service.acme:9093
    scheme: http
    timeout: 10s
rule_files:
- /etc/prometheus/tasks.rules
- /etc/prometheus/host.rules
- /etc/prometheus/containers.rules
scrape_configs:
- job_name: cadvisor
  scrape_interval: 5s
  scrape_timeout: 5s
  metrics_path: /metrics
  scheme: http
  dns_sd_configs:
  - names:
    - tasks.cadvisor
    refresh_interval: 30s
    type: A
    port: 8080
- job_name: node-exporter
  scrape_interval: 5s
  scrape_timeout: 5s
  metrics_path: /metrics
  scheme: http
  dns_sd_configs:
  - names:
[...]

PR #5009, opened 2018-12-17
Populate external labels for annotations expansion 

sylr:
Attempt to fix https://github.com/prometheus/prometheus/issues/3043


brian-brazil:
As noted in #3043, this is by design. In addition it'd be weird if external_labels were applied twice, and also weird if they ended up in ALERTS.

sylr:
> external_labels were applied twice

@brian-brazil what do you mean ?

sylr:
@fabxc @simonpasquier could I have your thoughts on this ?

poblahblahblah:
We would **love** to see this merged in.

beorn7:
I think we should do this, and I suspect the vast majority of prometheus-team thinks so, too. We should get clarity on this first, though, before actually implementing it. I have re-opened #3043 for now.

beorn7:
I think this implementation is doing the right thing. Would be good to have a test for it, though (in particular about the case where a label already exists, which must then _not_ be replaced by the external label).

It might be good to use the same code path for adding the external labels to templates and to the externally served metrics.

beorn7:
And yet another thought: If there are really concerns about the breakage this change might cause, we could hide this feature behind a flag (and then switch the default or remove it for Prom 3).

sylr:
@beorn7 Would you prefer I unexport `*CurrentConfig` and add `GetCurrentConfig() *config` method ?

beorn7:
@sylr Let's discuss implementation details once we have a decision on #3043 .

beorn7:
I guess we have a decision in #3043 . @sylr Do you want to adjust this PR? I could totally understand if that long discussion and all the implications (like unit tests and console templates) is a bit overwhelming. In which case somebody else can take care of it...

sylr:
I've been overwhelmed at work recently so I did not follow through the discussion but from I've been reading the concensus is to have a .ExternalLabels property ?

If I find the time I'll try to give it a look.

beorn7:
So the plan is:
- Make the external-labels map from the config available in templates (alert annotations, alert labels, and console templates) as `.ExternalLabels`. (Pay attention that the template expansion also works on the alert status page.)
- To make it work within ranges, add a convenience variable `$external_labels` for alert templates.
- Add a way to feed external labels to rule unit testing (with `promtool`). Either by adding them to the test definition, or by passing in a Prometheus config file via the promtool command line, or both. (To be discussed.)

If you find time to work on this, it would be much appreciated.

brian-brazil:
I wouldn't consider the last item in scope for this PR.

In relation to $ that would lead to an inconsistency where it's present in
alerts but not console templates. Or where that is in console templates,
but none of the others are. I believe all or nothing will resolve it, and
note that there's not much call for using this in a range.

poblahblahblah:
Is there anything I can do to help push this change along?

beorn7:
@sylr do you think you'll work on this in the foreseeable future?

Otherwise, this is just up for grabs for whoever feels qualified to do it.

beorn7:
First step is really only introducing `.ExternalLabels`. The other points I listed don't have to be in this PR.

sylr:
I'll take a look this afternoon.

simonpasquier:
@sylr you can rebase on `master` to pass the CI.

sylr:
@beorn7 PTAL

rules/alerting.go
```diff
@@ -299,15 +301,25 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc,
...
+
+		// Labels.
 		for _, lbl := range smpl.Metric {
 			l[lbl.Name] = lbl.Value
 		}
 
-		tmplData := template.AlertTemplateData(l, smpl.V)
+		// External labels.
+		el := make(map[string]string)
+		if config.CurrentConfig != nil && (*config.CurrentConfig).GlobalConfig.ExternalLabels != nil {
```
brian-brazil: There's a data race here

pkg/rulefmt/rulefmt.go
```diff
@@ -155,8 +155,8 @@ func testTemplateParsing(rl *Rule) (errs []error) {
 	}
 
 	// Trying to parse templates.
-	tmplData := template.AlertTemplateData(make(map[string]string), 0)
-	defs := "{{$labels := .Labels}}{{$value := .Value}}"
+	tmplData := template.AlertTemplateData(make(map[string]string), make(map[string]string), 0)
+	defs := "{{$labels := .Labels}}{{$external_labels := .ExternalLabels}}{{$value := .Value}}"
```
brian-brazil: It feels odd to have Go style names for fields, and Javascript style names for variables.

sylr: My first choice was `$externalLabels` but I felt it looked bad.

rules/alerting.go
sylr: If GlobalConfig is nil ?

sylr:
> 
> 
> This doesn't cover console templates.

No it does not, to be honnest I didn't even know about it.

rules/alerting.go
sylr: Should be fixed now.

pkg/rulefmt/rulefmt.go
sylr: What do you propose ?

brian-brazil: externalLabels would be my thought.

sylr: Done.

beorn7:
I think the CI failure was due to not being able to download staticcheck. Re-running…

beorn7:
:tada: Build is green.
@brian-brazil is there anything left to improve in this PR?

config/config.go
```diff
@@ -66,6 +67,10 @@ func LoadFile(filename string) (*Config, error) {
 
 // The defaults applied before parsing the respective config sections.
 var (
+	// CurrentConfig is a pointer to the current configuration.
```
brian-brazil: A global isn't how we pass config around, see https://github.com/prometheus/prometheus/blob/master/cmd/prometheus/main.go#L401

brian-brazil:
> I think the CI failure was due to not being able to download staticcheck. Re-running…

That happens a bit unfortunately.

config/config.go
sylr: I know.

sylr:
> I'm still not seeing console templates covered.

Can you give me pointers ?

brian-brazil:
Look for `ExpandHTML`.

poblahblahblah:
Looks like the build is green. Can this make it's way into 2.8?

brian-brazil:
I don't see this making it into 2.8, the code is still not right.

beorn7:
Sorry for not following this review in the necessary depth to figure out why this is getting stale.

@sylr are you aware what makes @brian-brazil think this is not quite there yet? Or alternatively, @brian-brazil , could you explain in more detail?

sylr:
@beorn7 I believe @brian-brazil does not like the config being accessed through a global var.

beorn7:
Do you think it's a reasonable concern? Do you know how to address it?
(Just trying to unblock you two here.)

sylr:
I don't know, more experienced go developer should answer that. What I know is that it provided an easy enough solution to implement that did not require to touch the current configuration reloading system which seems complex to me.

brian-brazil:
I'm expecting the current system to be used as it's our standard way of doing things, and is better for those reusing our code.

sylr:
What could be done to make it prettier is to:

- unexport config.CurrentConfig and add the needed function to config to set and get the current config
- move the config.CurrentConfigMutex inside the Config struct and add Public method to Lock/Unlock.

brian-brazil:
RuleManager is already hooked into config updated, extend what's there.

sylr:
So reverting this: https://github.com/prometheus/prometheus/commit/bd9f7460eb6fac35a103036d8fc5cd63c42ddd63

cc @fabxc

brian-brazil:
You don't have to revert that.

poblahblahblah:
Hey ya'll.

Is there anything we can do to help push this feature over the line in time for Prometheus 2.9.0?

beorn7:
I hoped @sylr and @brian-brazil could shepherd this into a merge on their own. I would dive in here and help, but I am very short on time right now. Next week hopefully will be better...

beorn7:
I finally got some time to look into this. After I found myself writing a lot of code changes myself, I decided to create #5463 on top of this. Please have a look there.

beorn7:
Superseded by #5463.

Issue #5010, opened 2018-12-18
CORS: Leave user configure CORS in prometheus.yml


aloababa:
According to prometheus code Allowed Origin is always set to '*'.

`
var corsHeaders = map[string]string{
	"Access-Control-Allow-Headers":  "Accept, Authorization, Content-Type, Origin",
	"Access-Control-Allow-Methods":  "GET, OPTIONS",
	"Access-Control-Allow-Origin":   "*",
	"Access-Control-Expose-Headers": "Date",
}
`

So what if i want a single allowed origin ?

It would be nice to allow user configure allowed origins or disable cors in prometheus.yml

brian-brazil:
Per previous discussions, this would make more sense as a flag.

geekodour:
I'll try to send a PR for this one today.

cstyan:
looks like the two PR's fix this, can we close?

PR #5011, opened 2018-12-18
[WIP] Added CORS Origin flag

geekodour:
ref #5010 
I don't know if this is the best way to implement this flag but it seems to work, and I didn't change the test as the implementation is not confirmed.

In this PR:
- added `--web.cors.origin` flag
- removed `corsHeaders` variable
- added `headers` field to `/web/api/v1/api.go:API` and `/web/web.go:Handler`
- headers are initilized in `/web/web.go:New` instead

@brian-brazil `/web/api/v1/api.go` and `/web/web.go` have a function `setCORS` will it be a good idea to change it to `setHeaders`? 
this signature of the function is also changed in this PR.

https://github.com/prometheus/prometheus/blob/68f0787c37d530394328171a1e9e86d336f2cfb3/web/api/v1/api.go#L133-L137

https://github.com/prometheus/prometheus/blob/68f0787c37d530394328171a1e9e86d336f2cfb3/web/web.go#L353-L357

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 	"sync"
@@ -50,6 +51,7 @@ import (
 	"github.com/prometheus/prometheus/discovery"
 	sd_config "github.com/prometheus/prometheus/discovery/config"
 	"github.com/prometheus/prometheus/notifier"
+	"github.com/prometheus/prometheus/pkg/relabel"
 	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/rules"
 	"github.com/prometheus/prometheus/scrape"
@@ -99,7 +101,8 @@ func main() {
 		queryMaxSamples     int
 		RemoteFlushDeadline model.Duration
 
-		prometheusURL string
+		prometheusURL   string
+		corsRegexString string
 
 		promlogConfig promlog.Config
 	}{
@@ -205,8 +208,8 @@ func main() {
 	a.Flag("query.max-samples", "Maximum number of samples a single query can load into memory. Note that queries will fail if they would load more samples than this into memory, so this also limits the number of samples a query can return.").
 		Default("50000000").IntVar(&cfg.queryMaxSamples)
 
-	a.Flag("web.cors.origin", "CORS origin to use").
-		PlaceHolder("<URL>").StringsVar(&cfg.web.CORSOrigins)
+	a.Flag("web.cors.origin", "Regex for CORS origin").
+		Default(".*").StringVar(&cfg.corsRegexString)
 
 	promlogflag.AddFlags(a, &cfg.promlogConfig)
 
@@ -220,6 +223,12 @@ func main() {
 	cfg.web.ExternalURL, err = computeExternalURL(cfg.prometheusURL, cfg.web.ListenAddress)
 	if err != nil {
 		fmt.Fprintln(os.Stderr, errors.Wrapf(err, "parse external URL %q", cfg.prometheusURL))
+		os.Exit(2)
+	}
+
+	cfg.web.CORSOrigin, err = compileCORSRegexString(cfg.corsRegexString)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, errors.Wrapf(err, "could not compile regex string %q", cfg.corsRegexString))
 		os.Exit(2)
 	}
 
@@ -664,6 +673,15 @@ func reloadConfig(filename string, logger log.Logger, rls ...func(*config.Config
 func startsOrEndsWithQuote(s string) bool {
 	return strings.HasPrefix(s, "\"") || strings.HasPrefix(s, "'") ||
 		strings.HasSuffix(s, "\"") || strings.HasSuffix(s, "'")
+}
+
+// compileCORSRegexString compiles corsRegexString and adds anchors
+func compileCORSRegexString(s string) (*regexp.Regexp, error) {
+	r, err := relabel.NewRegexp(s)
+	if err != nil {
+		return nil, err
+	}
+	return r.Regexp, nil
 }
 
 // computeExternalURL computes a sanitized external URL from a raw input. It infers unset
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -226,9 +226,9 @@ func main() {
 		os.Exit(2)
 	}
 
-	cfg.web.CORSOrigin, err = compileRegexString(cfg.corsRegexString)
+	cfg.web.CORSOrigin, err = compileCORSRegexString(cfg.corsRegexString)
 	if err != nil {
-		fmt.Fprintln(os.Stderr, errors.Wrapf(err, "could not compile regex string %q", cfg.corsRegexString))
+		fmt.Fprintln(os.Stderr, errors.Wrapf(err, "could not compile CORS regex string %q", cfg.corsRegexString))
 		os.Exit(2)
 	}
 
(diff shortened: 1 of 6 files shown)


brian-brazil:
I don't believe this behaviour is correct, you need to look at the request's Origin header.

brian-brazil:
Yes, that'll be needed. Yes, that's one way to do it.

geekodour:
@brian-brazil 
I made the changes, please review the changes, 

According to the [CORS spec:6.1](https://www.w3.org/TR/cors/#resource-proce1ssing-model) a `simple request` works without having to respond 
with `Access-Control-Allow-Methods` header, the current implementation sends it with every `simple request`. 

The only case for a `preflighted request` was this:
https://github.com/prometheus/prometheus/blob/d304269ac4875d9305fafc12b8eb58a1591f048d/web/api/v1/api.go#L244

which is an unimplemented function:
https://github.com/prometheus/prometheus/blob/d304269ac4875d9305fafc12b8eb58a1591f048d/web/api/v1/api.go#L511-L513

In the PR, I added `DELETE` to `Access-Control-Allow-Methods` for `preflighted` requests to be successful and `POST` just for the sake of completion. 

Additionally while working on this, I found that **cors requests** to resources with unsupported methods result in CORS errors, rather than informing that the method is not supported. (This is reproducible on master). This indicates that the CORS headers were not added yet. ( I am not sure if a fix for this is necessary )

web/web.go
```diff
@@ -342,17 +344,32 @@ func New(logger log.Logger, o *Options) *Handler {
...
+		"Access-Control-Allow-Methods":  "GET, POST, OPTIONS, DELETE",
+		"Access-Control-Expose-Headers": "Date",
+	}
+
+	for k, v := range headers {
+		w.Header().Set(k, v)
+	}
+
+	if len(o) == 0 {
+		w.Header().Set("Access-Control-Allow-Origin", "*")
```
brian-brazil: For clarity, this should return here

web/web.go
```diff
@@ -342,17 +344,32 @@ func New(logger log.Logger, o *Options) *Handler {
...
-	"Access-Control-Allow-Methods":  "GET, OPTIONS",
-	"Access-Control-Allow-Origin":   "*",
-	"Access-Control-Expose-Headers": "Date",
-}
-
 // Enables cross-site script calls.
-func setCORS(w http.ResponseWriter) {
-	for h, v := range corsHeaders {
-		w.Header().Set(h, v)
+func setCORS(w http.ResponseWriter, o []string, r *http.Request) {
```
brian-brazil: I don't like that we're duplicating security related code. Could you move this under pkg somewhere?

web/web.go
```diff
@@ -342,17 +344,32 @@ func New(logger log.Logger, o *Options) *Handler {
...
+	origin := r.Header.Get("Origin")
+	if origin == "" {
+		return
+	}
+
+	headers := map[string]string{
```
brian-brazil: You need a bit more: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin#CORS_and_caching

cmd/prometheus/main.go
```diff
@@ -205,6 +205,9 @@ func main() {
 	a.Flag("query.max-samples", "Maximum number of samples a single query can load into memory. Note that queries will fail if they would load more samples than this into memory, so this also limits the number of samples a query can return.").
 		Default("50000000").IntVar(&cfg.queryMaxSamples)
 
+	a.Flag("web.cors.origin", "CORS origin to use").
+		PlaceHolder("<URL>").StringsVar(&cfg.web.CORSOrigins)
```
brian-brazil: Hmm, I wonder if this would be best as a regex.

web/api/v1/api.go
```diff
@@ -130,9 +123,31 @@ type apiFuncResult struct {
...
-		w.Header().Set(h, v)
+func setCORS(w http.ResponseWriter, o []string, r *http.Request) {
+	origin := r.Header.Get("Origin")
+	if origin == "" {
+		return
+	}
+
+	headers := map[string]string{
+		"Access-Control-Allow-Headers":  "Accept, Authorization, Content-Type, Origin",
+		"Access-Control-Allow-Methods":  "GET, POST, OPTIONS, DELETE",
```
brian-brazil: I'm not sure we want to allow DELETE

web/web.go
geekodour: will `/util/httputil` package be a better place to put this? I am not sure which will be the best place under `pkg`

brian-brazil: That sounds good.

util/httputil/cors_test.go
```diff
@@ -0,0 +1,68 @@
...
+	tearDown := setup()
+	defer tearDown()
+	client := &http.Client{}
+
+	ch := getCORSHandlerFunc()
+	mux.Handle("/any_path", ch)
+
+	dummyOrigin := "https://foo.com"
+
+	// OPTIONS WITH LEGIT ORIGIN
```
brian-brazil: No need to shout.

cmd/prometheus/main.go
geekodour: @brian-brazil 
regex as in like this?:
`--web.cors.origin='^(http|https)://(domain1|domain2).(com)$'`

brian-brazil: Yes, though we anchor regexes for users

geekodour:
@brian-brazil I think the failing test is unrelated?

brian-brazil:
That's unrelated, haven't seen that one before.

cmd/prometheus/main.go
```diff
@@ -220,6 +226,12 @@ func main() {
 		os.Exit(2)
 	}
 
+	cfg.web.CORSOrigin, err = compileCORSRegexString(cfg.corsRegexString)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, errors.Wrapf(err, "could not compile regex string %q", cfg.corsRegexString))
```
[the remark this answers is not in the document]
geekodour: I think you meant "... should probably **not** mention ..." , I made the changes accordingly.
Should I update the docs here as well? 

https://prometheus.io/docs/operating/security/

> For non-mutating endpoints, you may wish to set CORS headers such as Access-Control-Allow-Origin in your reverse proxy to prevent XSS or use the `web.cors.origin` option.

cmd/prometheus/main.go
```diff
@@ -205,6 +208,9 @@ func main() {
 	a.Flag("query.max-samples", "Maximum number of samples a single query can load into memory. Note that queries will fail if they would load more samples than this into memory, so this also limits the number of samples a query can return.").
 		Default("50000000").IntVar(&cfg.queryMaxSamples)
 
+	a.Flag("web.cors.origin", `Regex for CORS origin. It is anchored in both ends by prometheus. Eg. 'https?://(domain1|domain2)\.com'`).
```
brian-brazil: no need for "by Prometheus". "fully anchored" is how we usually say that.

cmd/prometheus/main.go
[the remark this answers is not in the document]
brian-brazil: I meant what I said.

> Should I update the docs here as well?

You can once this gets released, it's over in the docs repo.

brian-brazil:
Thanks!

andrew-musoke:
Hi. Is there any documentation on how to use this flag? I cannot seem to find any. I am trying to consume the API values to use in my UI graphs. But I am getting CORS issues.

roidelapluie:
It makes more sense to ask questions like this on the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users) rather than in a GitHub issue. On the mailing list, more people are available to potentially respond to your question, and the whole community can benefit from the answers provided.

PR #5012, opened 2018-12-19
Variable 'labels' collides with imported package name

aixeshunter:
Declarations of variables that overlap with the used import. Variable 'labels' collides with imported package name "github.com/prometheus/prometheus/pkg/labels".


brian-brazil:
Thanks!

PR #5018, opened 2018-12-20
import aliases 'yaml' that can be omitted

aixeshunter:
The import `gopkg.in/yaml.v2` that package also named `yaml`(package yaml), the aliases `yaml` is redundant.


brian-brazil:
Thanks!

You can send all of these small things as one PR if you want.

PR #5015, opened 2018-12-19
Use latest version azure go sdk and go-autorest

tariq1890:
Update Azure-Go-SDK to 23.2.0
Update Go-Autorest to 11.2.8

The current build of Azure Go SDK is over 2 years old. It would be in the best interests to keep the Azure deps updated. Also, the querying of PowerStates requires us to use more recent. versions of the Go SDK (refer to #4908 )

Please note that is a WIP. Feedback is appreciated :). The changes worth noting can be seen in `azure.go` and `azure_test.go`. There are changes in the method signatures and type structure which I thought might be of interest to the reviewers.

PR #5020, opened 2018-12-20
Upgrade miekg dns

erikh:
This upgrades miekg/dns to 1.1.1, which seems small but combined with https://github.com/prometheus/prometheus/pull/5015 and https://github.com/prometheus/prometheus/pull/4927 means that large parts of kubernetes 1.13.1 can coexist in the same vendor tree with prometheus, which would make my employer very happy. :)

I tested everything and tried to annotate the commits appropriately. Please let me know if anything needs to change, happy to assist in any way!

BTW, two modules -- one from hashicorp and etcd -- had messed up checksums in the go.sum, so I regenerated them. I hope that's also ok!


erikh:
I'll purge my gopath and try this again.

simonpasquier:
> BTW, two modules -- one from hashicorp and etcd -- had messed up checksums in the go.sum, so I regenerated them. I hope that's also ok!

You probably have a problem with the modules cache on your machine (`go clean -modcache`) or you don't use Go1.11.4 (as it computes different hashes compared to Go1.11.3).

erikh:
looks that way. thanks!

erikh:
@simonpasquier done!

erikh:
let me correct that, I omitted it because it wasn't related to my changes.

vendor/modules.txt
```diff
@@ -179,7 +179,7 @@ github.com/julienschmidt/httprouter
 github.com/kr/logfmt
 # github.com/matttproud/golang_protobuf_extensions v1.0.1
 github.com/matttproud/golang_protobuf_extensions/pbutil
-# github.com/miekg/dns v1.0.4
+# github.com/miekg/dns v1.1.1
```
tariq1890: can we upgrade to 1.1.3

tariq1890:
```suggestion
# github.com/miekg/dns v1.1.3
```

discovery/dns/dns.go
```diff
@@ -317,7 +317,7 @@ func askServerForName(name string, queryType uint16, client *dns.Client, servAdd
 	}
 
 	response, _, err := client.Exchange(msg, servAddr)
-	if err == dns.ErrTruncated {
+	if err != nil || response.Truncated {
```
tariq1890:
```suggestion
	if response.Truncated {
```

the `err != nil` check is being done in L328. If an error other than truncated message is thrown, then that error's details will never be shown with this conditional block.

tariq1890: Pinging @simonpasquier as he LGTM'ed this. Please let me know if this review comment holds.

tariq1890:
@erikh Any updates on this?

erikh:
Hi friends. I will address this over the weekend, I haven't felt well this
week.

tariq1890:
Sure! Hope you recover soon :)

discovery/dns/dns.go
tariq1890: Sure! I am editing my comment to remove the GH suggestion.

erikh:
Getting to this now. Sorry for the delay.

discovery/dns/dns.go
erikh: I'm not seeing where errid is needed or where it comes from, but I've made the rest of the changes; committing now.

erikh:
this is ready for review at this point I think; Not sure what ya'll mean by the ErrID business though. If someone wants to patch it or send me a diff, I'll incorporate it.

discovery/dns/dns.go
```diff
@@ -317,17 +317,19 @@ func askServerForName(name string, queryType uint16, client *dns.Client, servAdd
 	}
 
 	response, _, err := client.Exchange(msg, servAddr)
-	if err == dns.ErrTruncated {
+	if err != nil {
```
tariq1890:
```suggestion
	if err != nil {
	       if err == dns.ErrId {
	            return nil, fmt.Errorf("DNS ID mismatch, request: %d, response: %d", msg.Id, response.Id)
	       }     
```

discovery/dns/dns.go
```diff
@@ -317,17 +317,19 @@ func askServerForName(name string, queryType uint16, client *dns.Client, servAdd
...
 		}
 
 		client.Net = "tcp"
 		return askServerForName(name, queryType, client, servAddr, false)
 	}
-	if err != nil {
-		return nil, err
-	}
+
 	if msg.Id != response.Id {
```
tariq1890: As per @simonpasquier , you can now remove this. This check is now redundant. See https://github.com/miekg/dns/blob/56be65265e34e731425e0269a301774938827c60/client.go#L432 for clarity on this.

tariq1890:
@erikh I have expanded on Simon's review comment. Hopefully I have understood what he meant and that it makes sense

discovery/dns/dns.go
erikh: Ah I see now. I will resolve it. Thanks.

erikh:
peek now; I dove the source as well... the response can be nil so I guarded against that too.

discovery/dns/dns.go
```diff
@@ -317,19 +317,26 @@ func askServerForName(name string, queryType uint16, client *dns.Client, servAdd
 	}
 
 	response, _, err := client.Exchange(msg, servAddr)
-	if err == dns.ErrTruncated {
+	if err != nil {
+		if err == dns.ErrId {
+			if response != nil { // response can be nil here, just be paranoid
```
tariq1890: This check is of little consequence since we return `nil` anyway. I think we are better off without this extra if check.

tariq1890: From the prometheus SD perspective.

erikh:
The id deref in the error message will panic otherwise.

discovery/dns/dns.go
erikh: oh the conditional is reversed, oops.

erikh: there, the conditional is fixed. as mentioned prior, if we remove this and the response is nil, then the `response.Id` dereference will panic the running goroutine.

discovery/dns/dns.go
```diff
@@ -317,19 +317,26 @@ func askServerForName(name string, queryType uint16, client *dns.Client, servAdd
 	}
 
 	response, _, err := client.Exchange(msg, servAddr)
-	if err == dns.ErrTruncated {
+	if err != nil {
+		if err == dns.ErrId {
+			if response == nil { // response can be nil here, just be paranoid
```
tariq1890: @erikh , I still think this should be removed. Look at L431 of `dns/client.go` [here](https://github.com/miekg/dns/blob/a288d199dc686dca98438c25c90c16647a6e28df/client.go#L431). For the outcome of ErrId, a nil of check of `response` is already being done without a guard. If `response` was nil , then the panic would originate from the `miekg/dns` itself. Now if it so happens that an unguarded nil check is being in that library, then that fix should go upstream :).

erikh:
Sure, no problem. I was just of the mindset that it's much more
conventional to return nil with an error, so perhaps relying on that
outlier would not be worth it.

I'll make the patch later tonight.

tariq1890:
Thank you for your patience with this :).

discovery/dns/dns.go
tariq1890: So based on my comment above: if `response` was indeed nil, the `panic` would happen anyway and originate from the library. This guard will not protect us unfortunately.

erikh:
ok, it's patched in.

tariq1890:
@brian-brazil Requesting your review on this.

discovery/dns/dns.go
```diff
@@ -317,19 +317,22 @@ func askServerForName(name string, queryType uint16, client *dns.Client, servAdd
 	}
 
 	response, _, err := client.Exchange(msg, servAddr)
-	if err == dns.ErrTruncated {
+	if err != nil {
+		if err == dns.ErrId {
```
brian-brazil: Do we really need to special case this?

erikh: if we don't, the response may be nil and inappropriately dereferenced; this is one of many solutions. Happy to take whatever the maintainers agree on.

simonpasquier: I'm ok having only:

```go
        response, _, err := client.Exchange(msg, servAddr)
	if err != nil {
		return nil, err
	}
```

I suspect that the previous `if msg.Id != response.Id {...}` block was there for historical reasons (eg the dns library didn't return `dns.ErrId` at that time).

erikh: Not an issue, but I'd like to request a quorum of agreement before I make any more changes to this. Again, happy to, just optimizing the process a bit I hope.

brian-brazil: @simonpasquier That's also my suspicion.

tariq1890: (Chipping in as a non-maintainer): I completely understand the @brian-brazil 's comment and I think it should be fine even if we stop special casing Id missing error. To explain my review comments, I had suggested we special case that error only because it was already being done before and this PR was essentially made for upgrading the `miekg/dns` dependency. (Moreover, the new `ErrId` wraps a string that says `Id mismatch` rather than show the two mistmatching Ids).

In any case, I think the changes should be fine either way. Can we request consensus among the maintainers so that @erikh is able to complete this PR?

erikh:
Not to add fuel but this is the last patch in a large patch set we've been
carrying that hasn't been contributed to the community yet. We'd really
like to get it in.

>

simonpasquier:
The consensus is to remove the `if err == dns.ErrId {...}` block as it is redundant. Other than that the PR is good to go from my POV.

simonpasquier:
@erikh would you able to adjust the PR? Thanks for the work and sorry for the back and forth.

erikh:
Yeah sorry I've just been distracted this week. I'll get to it today.

erikh:
Chipping at this now; requesting reviews.

erikh:
looks like a circleCI issue, not the build. I'll see if a FP fixes it.

beorn7:
Merging as per discussion in #5264.

PR #5024, opened 2018-12-20
discovery/gce: oauth2.NoContext is deprecated, replace with context.Background()

erikh:
small patch! woo!

Signed-off-by: Erik Hollensbe <github@hollensbe.org>


brian-brazil:
Thanks!

PR #4955, opened 2018-12-04
Removed config dependency from `relabel` package.

bwplotka:
Changes:
* Moved RelabelConfig to `pkg/relabel` package
* Fixed all usages.
* Adapted top dir relabel to use pkg relabel structs.

Removal of top level `relabel` package in a separate tracked here: https://github.com/prometheus/prometheus/issues/3647

PTAL @krasi-georgiev @brian-brazil 

Signed-off-by: Bartek Plotka <bwplotka@gmail.com>

PR #5025, opened 2018-12-20
notifier: Moved config package dependency from notifier package.

bwplotka:
Follow up from https://github.com/prometheus/prometheus/pull/4955 to make notifier independent of config package.

~TODO in next PR: Move notifier to `pkg/notifier`, @bbrazil what do you think?~

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -22,11 +22,11 @@ import (
 	"strings"
 	"time"
 
-	"github.com/prometheus/prometheus/pkg/relabel"
-
 	config_util "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
 	sd_config "github.com/prometheus/prometheus/discovery/config"
+	"github.com/prometheus/prometheus/notifier"
+	"github.com/prometheus/prometheus/pkg/relabel"
 	"gopkg.in/yaml.v2"
 )
 
@@ -89,8 +89,7 @@ var (
 
 	// DefaultAlertmanagerConfig is the default alertmanager configuration.
 	DefaultAlertmanagerConfig = AlertmanagerConfig{
-		Scheme:  "http",
-		Timeout: model.Duration(10 * time.Second),
+		NotifierConfig: notifier.DefaultConfig,
 	}
 
 	// DefaultRemoteWriteConfig is the default remote write configuration.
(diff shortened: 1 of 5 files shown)


brian-brazil:
This looks okay from a first glance. I don't see a reason we'd move this to pkg, it's not a utility library.

bwplotka:
Hm.. I thought it is a golang convention to have all in `pkg` and thought we are moving towards it. Ignore me then, removed this todo from description.

config/config.go
```diff
@@ -456,10 +445,10 @@ func (c *AlertmanagerConfig) UnmarshalYAML(unmarshal func(interface{}) error) er
 		return err
 	}
 
-	// The UnmarshalYAML method of HTTPClientConfig is not being called because it's not a pointer.
+	// The UnmarshalYAML method of NotifierConfig.HTTPClientConfig is not being called because it's not a pointer.
 	// We cannot make it a pointer as the parser panics for inlined pointer structs.
 	// Thus we just do its validation here.
-	if err := c.HTTPClientConfig.Validate(); err != nil {
+	if err := c.NotifierConfig.HTTPClientConfig.Validate(); err != nil {
```
brian-brazil: Shouldn't this check move down?

notifier/notifier.go
```diff
@@ -553,7 +564,7 @@ type alertmanagerSet struct {
 	logger     log.Logger
 }
 
-func newAlertmanagerSet(cfg *config.AlertmanagerConfig, logger log.Logger) (*alertmanagerSet, error) {
+func newAlertmanagerSet(cfg Config, logger log.Logger) (*alertmanagerSet, error) {
```
brian-brazil: You're passing around full structs rather than pointers, just pointers would be more efficient.

bwplotka:
PTAL @brian-brazil  (:  This is last (known) piece that stops us from upgrading Prometheus dependency in Thanos codebase.

brian-brazil:
I was looking at it honest, but paused to take a shower :)

config/config.go
brian-brazil: I meant moved down to NotifierConfig, due to Law of Demeter

bwplotka:
Definitely stopped doing this. Help wanted to continue this :hugs: 

Let's also update an issue with latest for removing config deps or rather some idea of indirection for config.

bwplotka:
Actually this is no longer needed if https://github.com/prometheus/prometheus/issues/7107 will be done.

PR #5026, opened 2018-12-21
marathon-sd - use Tasks.Ports instead of PortDefinitions.Ports if RequirePorts is false

### LINKED ISSUE #5022: marathon_sd connect to wrong ports
## Bug Report

**What did you do?**

Tried to get metrics of marathon applications.

Metrics were scraped by their correct port.

Got connection refused error because of marathon_sd connect to wrong ports. marathon_sd should use `Tasks.Ports` instead of `PortDefinitions.Ports` if `RequirePorts` is `false` according to [document of marathon](http://mesosphere.github.io/marathon/1.6/docs/networking.html#port-definition).

**Environment**

* System information:

	`Darwin 16.5.0 x86_64`

* Prometheus version:

	```
  prometheus, version 2.6.0 (branch: HEAD, revision: dbd1d58c894775c0788470944b818cc724f550fb)
  build user:       root@localhost
  build date:       20181220-09:35:49
  go version:       go1.11.4
  ```

  <details>
  <summary>Click to expand</summary>
  <pre>
  <code>
  global:
  scrape_interval:     15s
  evaluation_interval: 15s
  alerting:
  alertmanagers:
  - static_configs:
    - targets:
  rule_files:
  scrape_configs:
  - job_name: 'prometheus'
    static_configs:
    - targets: ['localhost:9090']
  - job_name: 'marathon'
    scrape_interval: 15s
    scrape_timeout: 10s
    metrics_path: /metrics
    scheme: http
    marathon_sd_configs:
      - servers:
        - http://localhost:8080
        refresh_interval: 30s
    relabel_configs:
    - source_labels:
    ...
  </code>
  </pre>
  </details>

* Logs:

  ```
  level=debug ts=2018-12-20T09:43:09.582113Z caller=scrape.go:825 component="scrape manager" scrape_pool=marathon target=http://192.168.0.1:10006/metrics msg="Scrape failed" err="Get http://192.168.0.1:10006/metrics: dial tcp 192.168.0.1:10012: connect: connection refused"
  ```

  In this case, correct port was 31690.

  <details>
  <summary>Click to expand</summary>
  <pre>
  <code>
  $ curl -s http://localhost:8080/v2/apps/test/httpserver | jq .
  {
    "app": {
      "id": "/test/httpserver",
      ...
      "container": {
        "type": "DOCKER",
        "volumes": [],
        "docker": {
          "image": "centos",
          "network": "HOST",
          "portMappings": [],
          "privileged": false,
          "parameters": [],
          "forcePullImage": false
        }
      },
      ...
      "ports": [
        10006
      ],
      "portDefinitions": [
        {
          "port": 10006,
          "protocol": "tcp",
          "name": "http",
          "labels": {}
        }
      ],
      "requirePorts": false,
      ...
      "tasks": [
        {
          "ipAddresses": [
            {
              "ipAddress": "192.168.0.1",
              "protocol": "IPv4"
            }
          ],
          ...
          "state": "TASK_RUNNING",
          "ports": [
            31690
          ],
          ...
          "host": "192.168.0.1"
        }
      ]
    }
  }
  </code>
  </pre>
  </details>

tommarute:
Fixes #5022


discovery/marathon/marathon.go
```diff
@@ -427,7 +428,7 @@ func targetsForApp(app *App) []model.LabelSet {
 		ports, labels = extractPortMapping(app.Container.Docker.PortMappings, app.isContainerNet())
 		prefix = portMappingLabelPrefix
 
-	} else if len(app.PortDefinitions) != 0 {
+	} else if app.RequirePorts && len(app.PortDefinitions) != 0 {
```
ti-mo: This will effectively disable label gathering when `requirePorts` is false, which is the default. This should have broken at least a test or two, please take a look at this, I will dive in as well. This might be missing from coverage.

In any case, I would wrap `ports[i] = app.PortDefinitions[i].Port` below in an `if app.RequirePorts` instead of this. This will make sure `ports` is still initialized to an array of zeroes of the correct length in both cases, making your change on line 462 unnecessary. (a zero will cause a port lookup in the task)
Make sure to add a comment, eg.
```
// When requirePorts is false, this port becomes the 'servicePort', not the listen port.
// In this case, the port needs to be taken from the task instead of the app.
```

discovery/marathon/marathon.go
```diff
@@ -458,7 +459,7 @@ func targetsForApp(app *App) []model.LabelSet {
 
 			// A zero port can appear in a portMapping, which means it is auto-generated.
 			// The allocated port appears at the corresponding index in the task's 'ports' array.
-			if port == 0 && len(t.Ports) == len(ports) {
+			if (!app.RequirePorts || port == 0) && len(t.Ports) == len(ports) {
```
ti-mo: Undo this change, an upstream zero port will cause this code to do The Right Thing. (tm)

discovery/marathon/marathon.go
```diff
@@ -458,7 +459,7 @@ func targetsForApp(app *App) []model.LabelSet {
 
 			// A zero port can appear in a portMapping, which means it is auto-generated.
```
ti-mo: Let's update this comment to:
```
// A zero port here means that either the portMapping has a zero port defined,
// or there is a portDefinition with requirePorts set to false. This means the port
// is auto-generated by Mesos and needs to be looked up in the task.
```

discovery/marathon/marathon_test.go
```diff
@@ -747,6 +822,7 @@ func marathonTestAppListWithContainerNetworkAndPortDefinition(labels map[string]
 			Container:       container,
 			Networks:        networks,
 			PortDefinitions: portDefinitions,
+			RequirePorts:    true,
```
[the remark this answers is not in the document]
tommarute: I get it. Thanks.

By the way, Should I delete this and `TestMarathonSDSendGroupWithContainerNetworkAndPortDefinition`? It seems that apps in container networks doesn't have not only `requirePorts` but also `portDefinitions` as far as I confirmed my configuration of marathon.

discovery/marathon/marathon.go
tommarute: Awthanks!

discovery/marathon/marathon_test.go
```diff
@@ -358,6 +358,7 @@ func marathonTestAppListWithPortDefinitions(labels map[string]string, runningTas
 				{Labels: make(map[string]string), Port: 31000},
 				{Labels: labels, Port: 32000},
 			},
+			RequirePorts: true,
```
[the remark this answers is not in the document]
tommarute: I removed it. But test failed. I think that I need to add `Ports: []uint32{31000, 32000},` to a `task`.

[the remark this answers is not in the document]
ti-mo: Correct, but I wouldn't use those ports, since they already correspond to the expected test results. I've pushed a commit here: https://github.com/prometheus/prometheus/commit/4a4a442def69f18c81272e72a92d8af81a53cab2

ti-mo:
@tommarute I went over your changes, and to the best of my knowledge, this implementation should be correct. Please include https://github.com/prometheus/prometheus/pull/5026#discussion_r245950481.

tommarute:
@ti-mo I have included your commit. I really appreciate your help in resolving the issue.

ti-mo:
@tommarute Thanks for the patience, sorry for the breakage.

@simonpasquier This looks good to me. Needs to be fixed on master as well. What does the 2.7 release window look like?

discovery/marathon/marathon.go
```diff
@@ -435,7 +436,11 @@ func targetsForApp(app *App) []model.LabelSet {
 
 		for i := 0; i < len(app.PortDefinitions); i++ {
 			labels[i] = app.PortDefinitions[i].Labels
-			ports[i] = app.PortDefinitions[i].Port
+			// When requirePorts is false, this port becomes the 'servicePort', not the listen port.
+			// In this case, the port needs to be taken from the task instead of the app.
+			if app.RequirePorts {
```
brian-brazil: You should prefer relabelling to adding business logic to SD. If there's metadata missing you should add it.

ti-mo: @brian-brazil This is the label/port gathering stage. The 'port' field means either 'data' port or 'service' port based on the state of this bool.

While I agree with your stance, I can't think of a way to handle this in relabeling without complicating the user's relabeling rule set. Additionally, the user will need to be aware of this idiosyncracy, one that we, the people working on this feature, were not even aware of until this issue popped up.

From the point of view of reducing maintenance cost and amount of incidents/issues, you are completely right, but this way we can at least fix it once, for all users. Unfortunately, consuming this API is hard to get right..

brian-brazil: The fact that you need to add a boolean configuration option that the user has to care about indicates that this belongs in relabelling. It's better to have everything in relabelling, than each SD slowly reinventing it.

simonpasquier: Hmm, `RequirePorts` is returned by the Marathon API, not exposed in the SD configuration if this is what you meant.

ti-mo: @brian-brazil This is not a configuration option, it's a variable set in the Marathon app definition, obtained by querying the Marathon API.

brian-brazil: Ah, I'd missed that.

ti-mo: No worries, thanks for taking the time to review. :+1:

simonpasquier:
@brian-brazil PTAL

brian-brazil:
Thanks!

PR #5027, opened 2018-12-21
Fix error strings, slice declaration and expression 'c <= 'ÿ'' is always 'true'

aixeshunter:
1. Error string should not be capitalized or end with punctuation, fix in `storage/remote/codec.go`.
2. Slice declaration with empty literal initializer instead of nil, change the declaration to avoid an unnecessary allocation
3.Expression 'c <= 'ÿ'' is always 'true', delete them in `pkg/textparse/promlex.l.go`.

### DIFF
--- a/pkg/textparse/promlex.l.go
+++ b/pkg/textparse/promlex.l.go
@@ -263,7 +263,7 @@ yystart21:
 		goto yyrule9
 	case c == '\t' || c == ' ':
 		goto yystate23
-	case c >= '\x01' && c <= '\b' || c >= '\v' && c <= '\x1f' || c >= '!' && c <= 'ÿ':
+	case c >= '\x01' && c <= '\b' || c >= '\v' && c <= '\x1f' || c >= '!':
 		goto yystate22
 	}
 
@@ -272,7 +272,7 @@ yystate22:
 	switch {
 	default:
 		goto yyrule9
-	case c >= '\x01' && c <= '\t' || c >= '\v' && c <= 'ÿ':
+	case c >= '\x01' && c <= '\t' || c >= '\v':
 		goto yystate22
 	}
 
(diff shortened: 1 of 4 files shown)


storage/remote/read_test.go
```diff
@@ -122,12 +123,12 @@ func TestSeriesSetFilter(t *testing.T) {
 			toRemove: model.LabelSet{"foo": "bar"},
 			in: &prompb.QueryResult{
 				Timeseries: []*prompb.TimeSeries{
-					{Labels: labelsToLabelsProto(labels.FromStrings("foo", "bar", "a", "b")), Samples: []prompb.Sample{}},
+					{Labels: labelsToLabelsProto(labels.FromStrings("foo", "bar", "a", "b")), Samples: samples},
```
codesome: Instead of this, `Samples: nil` would work.

aixeshunter: haha, yes, it would work.

pkg/textparse/promlex.l.go
```diff
@@ -386,7 +386,7 @@ yystate34:
 	switch {
 	default:
 		goto yyrule17
-	case c >= '\x01' && c <= '\b' || c >= '\v' && c <= '\x1f' || c >= '!' && c <= 'z' || c >= '|' && c <= 'ÿ':
+	case c >= '\x01' && c <= '\b' || c >= '\v' && c <= '\x1f' || c >= '!' && c <= 'z' || c >= '|':
```
fabxc: This is generated code and must not be changed.

fabxc:
As @codesome said, this will be overwritten with the next run of the generator. You could try to fix this in the generating library. But I doubt it has any notable impact on performance.

storage/remote/read_test.go
juliusv: Or even just omitting `Samples` completely, in which case it will take its zero value `nil` automatically.

codesome:
The fix for error string is already upstream and the other changes are not relevant. Hence closing this.

PR #5029, opened 2018-12-21
tune the "Wal segment size" with a flag

glutamatt:
Add WALSegmentSize as an option, and the corresponding flag "storage.tsdb.wal-segment-size" to tune the max size of wal segment files.

The addressed base problem is to reduce the disk space used by wal segment files : on a raspberry pi, for instance, we often want to reduce write load of the sd card, then, the wal directory is mounted on a memory (space limited) partition.

the default value of the segment max file size, pushed the size of directory to 128 MB for each segment , which is too much ram consumption on a rasp.

the initial discussion is at https://github.com/prometheus/tsdb/pull/450

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -165,7 +165,7 @@ func main() {
 		Hidden().PlaceHolder("<duration>").SetValue(&cfg.tsdb.MaxBlockDuration)
 
 	a.Flag("storage.tsdb.wal-segment-size",
-		"Size at which to split the tsdb WAL segment files(e.g. 100MB)").
+		"Size at which to split the tsdb WAL segment files (e.g. 100MB)").
 		Hidden().PlaceHolder("<bytes>").BytesVar(&cfg.tsdb.WALSegmentSize)
 
 	a.Flag("storage.tsdb.retention", "How long to retain samples in storage.").
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -164,6 +164,10 @@ func main() {
 		"Maximum duration compacted blocks may span. For use in testing. (Defaults to 10% of the retention period).").
 		Hidden().PlaceHolder("<duration>").SetValue(&cfg.tsdb.MaxBlockDuration)
 
+	a.Flag("storage.tsdb.wal-segment-size",
+		"Maximum Size for tsdb Walk Ahead Log segment files").
+		Hidden().IntVar(&cfg.tsdb.WALSegmentSize)
+
 	a.Flag("storage.tsdb.retention", "How long to retain samples in storage.").
 		Default("15d").SetValue(&cfg.tsdb.Retention)
 
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -166,7 +166,7 @@ func main() {
 
 	a.Flag("storage.tsdb.wal-segment-size",
 		"Maximum Size for tsdb Walk Ahead Log segment files").
-		Hidden().IntVar(&cfg.tsdb.WALSegmentSize)
+		Hidden().PlaceHolder("<bytes>").BytesVar(&cfg.tsdb.WALSegmentSize)
 
 	a.Flag("storage.tsdb.retention", "How long to retain samples in storage.").
 		Default("15d").SetValue(&cfg.tsdb.Retention)
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -165,7 +165,7 @@ func main() {
 		Hidden().PlaceHolder("<duration>").SetValue(&cfg.tsdb.MaxBlockDuration)
 
 	a.Flag("storage.tsdb.wal-segment-size",
-		"Maximum Size for tsdb Walk Ahead Log segment files").
+		"Size at which to split the tsdb WAL segment files(e.g. 100MB)").
 		Hidden().PlaceHolder("<bytes>").BytesVar(&cfg.tsdb.WALSegmentSize)
 
 	a.Flag("storage.tsdb.retention", "How long to retain samples in storage.").
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -563,6 +563,11 @@ func main() {
 		g.Add(
 			func() error {
 				level.Info(logger).Log("msg", "Starting TSDB ...")
+				if cfg.tsdb.WALSegmentSize != 0 {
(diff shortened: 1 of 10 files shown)


cmd/prometheus/main.go
```diff
@@ -164,6 +164,10 @@ func main() {
 		"Maximum duration compacted blocks may span. For use in testing. (Defaults to 10% of the retention period).").
 		Hidden().PlaceHolder("<duration>").SetValue(&cfg.tsdb.MaxBlockDuration)
 
+	a.Flag("storage.tsdb.wal-segment-size",
+		"Maximum Size for tsdb Walk Ahead Log segment files").
+		Hidden().IntVar(&cfg.tsdb.WALSegmentSize)
```
krasi-georgiev: I think we should accept this with an unit and parse it.
100KB,100MB,10GB

brian-brazil:
Is this really something we should allow a user to configure? I can see users messing around with it (like they do with min/max block sizes) and just causing themselves problems.

krasi-georgiev:
I think if we add some sane restrictions for min and max it should be fine and if not can always revert it at some point.

cmd/prometheus/main.go
```diff
@@ -164,6 +164,10 @@ func main() {
 		"Maximum duration compacted blocks may span. For use in testing. (Defaults to 10% of the retention period).").
 		Hidden().PlaceHolder("<duration>").SetValue(&cfg.tsdb.MaxBlockDuration)
 
+	a.Flag("storage.tsdb.wal-segment-size",
+		"Maximum Size for tsdb Walk Ahead Log segment files").
```
krasi-georgiev:
```suggestion
		"Size at which to split the tsdb WAL segment files(e.g. 10MB,100MB)").
```

krasi-georgiev:
@glutamatt can you add some limits. Maybe: min 10mb and max 256mb

krasi-georgiev:
also can you check if we can add a test to ensure that setting this flag creates the correct cfg.tsdb options.

glutamatt:
> also can you check if we can add a test to ensure that setting this flag creates the correct cfg.tsdb options.

the only way i found of checking this is to :
1. dispatch a startup log about  WALSegmentSize
2. test in running a prometheus main command with a fixed flag value
3. check the logs of the runned command to verify if the final value in bytes is printed in logs

i'm trying to code this but i'm not sure it's the right way of testing it

krasi-georgiev:
yeah doesn't sound a good way to test it. 
Just add the min/max limits and that should do it for now.

storage/tsdb/tsdb.go
```diff
@@ -117,6 +121,9 @@ type Options struct {
 	// The maximum timestamp range of compacted blocks.
 	MaxBlockDuration model.Duration
 
+	// The maximum size of each WAL segment files
```
juliusv: "files" -> "file"

nit: add period at end of comment.

cmd/prometheus/main.go
```diff
@@ -164,6 +164,10 @@ func main() {
 		"Maximum duration compacted blocks may span. For use in testing. (Defaults to 10% of the retention period).").
 		Hidden().PlaceHolder("<duration>").SetValue(&cfg.tsdb.MaxBlockDuration)
 
+	a.Flag("storage.tsdb.wal-segment-size",
+		"Size at which to split the tsdb WAL segment files(e.g. 100MB)").
```
juliusv: nit: add space before opening parenthesis.

glutamatt:
> Just add the min/max limits and that should do it for now.

Done

cmd/prometheus/main.go
glutamatt: Done

storage/tsdb/tsdb.go
glutamatt: Done

storage/tsdb/tsdb.go
```diff
@@ -180,8 +185,16 @@ func Open(path string, l log.Logger, r prometheus.Registerer, opts *Options) (*t
 		}
 	}
 
+	walSegmentSize := int(opts.WALSegmentSize)
+	if walSegmentSize != 0 {
+		if walSegmentSize < 10*1024*1024 || walSegmentSize > 256*1024*1024 {
+			return nil, errors.New("tsdb.Options.WALSegmentSize must be set between 10MB and 256MB")
```
krasi-georgiev: maybe best to have this after this line.
https://github.com/prometheus/prometheus/blob/master/cmd/prometheus/main.go#L562

glutamatt: yep i agress, done

glutamatt: Maybe i could add a test to check if the app stop the right way if the flag value is set out of authorized bounds ?

krasi-georgiev: yeah that is not a bad idea.

glutamatt: Done !

krasi-georgiev:
LGTM

storage/tsdb/tsdb.go
```diff
@@ -19,6 +19,8 @@ import (
 	"time"
 	"unsafe"
 
+	"github.com/alecthomas/units"
+
```
simonpasquier: The blank line is not needed.

storage/tsdb/tsdb.go
```diff
@@ -117,6 +119,9 @@ type Options struct {
 	// The maximum timestamp range of compacted blocks.
 	MaxBlockDuration model.Duration
 
+	// The maximum size of each WAL segment file.
+	WALSegmentSize units.Base2Bytes
```
simonpasquier: I'd rather use `int` here.

storage/tsdb/tsdb.go
glutamatt: done thanks

storage/tsdb/tsdb.go
glutamatt: this is what i used before allowing the flag value to be a byte unit value thanks to https://github.com/alecthomas/units package. Then, i need WALSegmentSize to be the type units.Base2Bytes to make the parser works as expected

glutamatt: this is the doc of https://godoc.org/github.com/alecthomas/kingpin#ArgClause.BytesVar i used here https://github.com/prometheus/prometheus/pull/5029/files#diff-fc1d26df0adf6edc1b9f25c878ab9ccaR169

simonpasquier: Ah right I missed that `Options` was used directly in the main func.

PR #5031, opened 2018-12-21
Docker images for ARM32v7 and ARM64v8

johanneswuerbach:
Build and push docker images for linux arm32v7 and arm64v8.

Fixes https://github.com/prometheus/promu/issues/89

**Building**
The ARM images are building using `binfmt_misc`, which is the same mechanism also used by Docker for Mac to support building ARM images, some details are here https://www.ecliptik.com/Cross-Building-and-Running-Multi-Arch-Docker-Images/

While the circle docker executor doesn't natively support building ARM images, the machine executor has almost all the necessary dependencies pre-installed and allows privileged execution, which is required to setup `binfmt_misc`

The setup itself is done via`docker run --privileged linuxkit/binfmt:v0.6`, which is the same script also used inside Docker for Mac. (Source https://github.com/linuxkit/linuxkit/tree/master/pkg/binfmt)

As the default machine image is fairly old and `binfmt_misc` requires a Kernel 4.8+, use a more recent machine image https://circleci.com/docs/2.0/configuration-reference/#machine

**Publishing**

The two ARM images would be published to `prom/prometheus-arm32v7-linux` and `prom/prometheus-arm64v8-linux`, so those 2 docker hub / quay.io repositories would need to be created.

To simplify image consumption a multi-arch manifest is created using https://docs.docker.com/edge/engine/reference/commandline/manifest/, which allows `docker pull prometheus/prometheus` to pull the respective amd64, arm32v7 and arm64v8 image automatically.

Requires https://github.com/prometheus/busybox/pull/19 to have an arm32v7 and arm64v8 base image.

**Test**

Build images locally and pushed them to my registry using:
```
promu crossbuild -v -p "linux/amd64 linux/armv7 linux/arm64"
make docker DOCKER_REPO=johanneswuerbach
make docker-publish DOCKER_REPO=johanneswuerbach
```

Using them on macOS via Docker for Mac:
```
$ docker pull johanneswuerbach/prometheus:docker-arm && docker run --rm -it johanneswuerbach/prometheus:docker-arm --version
docker-arm: Pulling from johanneswuerbach/prometheus
Digest: sha256:c8212d84d2d79f32d9b824d9235d3b4ab178e21c3a762a643c8abe12145a2d22
Status: Image is up to date for johanneswuerbach/prometheus:docker-arm
prometheus, version 2.6.0 (branch: master, revision: 2e725a195a17155dfd1e172a1e1205fc7d6986ec)
  build user:       root@682d9280fbbf
  build date:       20181221-21:56:26
  go version:       go1.11.4
```

Using them on a Raspberry Pi 3:
```
$ docker pull johanneswuerbach/prometheus:docker-arm && docker run --rm -it johanneswuerbach/prometheus:docker-arm --version
docker-arm: Pulling from johanneswuerbach/prometheus
Digest: sha256:c8212d84d2d79f32d9b824d9235d3b4ab178e21c3a762a643c8abe12145a2d22
Status: Image is up to date for johanneswuerbach/prometheus:docker-arm
prometheus, version 2.6.0 (branch: master, revision: 2e725a195a17155dfd1e172a1e1205fc7d6986ec)
  build user:       root@682d9280fbbf
  build date:       20181221-22:07:33
  go version:       go1.11.4
```

while the current `latest` fails as expected:

```
$ docker pull prom/prometheus && docker run --rm -it prom/prometheus --version
Using default tag: latest
latest: Pulling from prom/prometheus
Digest: sha256:1ffbf5d3c6476384905e8f57c98ac0611f328af68bedb909ec3f350d7e18b134
Status: Image is up to date for prom/prometheus:latest
standard_init_linux.go:190: exec user process caused "exec format error"
```

Feedback more then welcome :-)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,11 +53,11 @@ jobs:
 
     steps:
     - checkout
-    - setup_remote_docker
+    - setup_remote_docker:
+        version: 18.06.0-ce
+    - run: docker run --privileged linuxkit/binfmt:v0.6
     - attach_workspace:
         at: .
-    - run: ln -s .build/linux-amd64/prometheus prometheus
-    - run: ln -s .build/linux-amd64/promtool promtool
     - run: make docker
     - run: make docker DOCKER_REPO=quay.io/prometheus
     - run: docker images
@@ -65,13 +65,17 @@ jobs:
     - run: docker login -u $QUAY_LOGIN -p $QUAY_PASSWORD quay.io
     - run: make docker-publish
     - run: make docker-publish DOCKER_REPO=quay.io/prometheus
+    - run: make docker-manifest
+    - run: make docker-manifest DOCKER_REPO=quay.io/prometheus
 
   docker_hub_release_tags:
     executor: golang
 
     steps:
     - checkout
-    - setup_remote_docker
+    - setup_remote_docker:
+        version: 18.06.0-ce
+    - run: docker run --privileged linuxkit/binfmt:v0.6
     - run: mkdir -v -p ${HOME}/bin
     - run: curl -L 'https://github.com/aktau/github-release/releases/download/v0.7.2/linux-amd64-github-release.tar.bz2' | tar xvjf - --strip-components 3 -C ${HOME}/bin
     - run: echo 'export PATH=${HOME}/bin:${PATH}' >> ${BASH_ENV}
@@ -84,19 +88,23 @@ jobs:
     - store_artifacts:
         path: .tarballs
         destination: releases
-    - run: ln -s .build/linux-amd64/prometheus prometheus
-    - run: ln -s .build/linux-amd64/promtool promtool
     - run: make docker DOCKER_IMAGE_TAG=$CIRCLE_TAG
     - run: make docker DOCKER_IMAGE_TAG=$CIRCLE_TAG DOCKER_REPO=quay.io/prometheus
     - run: docker login -u $DOCKER_LOGIN -p $DOCKER_PASSWORD
     - run: docker login -u $QUAY_LOGIN -p $QUAY_PASSWORD quay.io
+    - run: make docker-publish DOCKER_IMAGE_TAG="$CIRCLE_TAG"
+    - run: make docker-publish DOCKER_IMAGE_TAG="$CIRCLE_TAG" DOCKER_REPO=quay.io/prometheus
+    - run: make docker-manifest DOCKER_IMAGE_TAG="$CIRCLE_TAG"
+    - run: make docker-manifest DOCKER_IMAGE_TAG="$CIRCLE_TAG" DOCKER_REPO=quay.io/prometheus
     - run: |
         if [[ "$CIRCLE_TAG" =~ ^v[0-9]+(\.[0-9]+){2}$ ]]; then
           make docker-tag-latest DOCKER_IMAGE_TAG="$CIRCLE_TAG"
           make docker-tag-latest DOCKER_IMAGE_TAG="$CIRCLE_TAG" DOCKER_REPO=quay.io/prometheus
+          make docker-publish DOCKER_IMAGE_TAG="latest"
+          make docker-publish DOCKER_IMAGE_TAG="latest" DOCKER_REPO=quay.io/prometheus
+          make docker-manifest DOCKER_IMAGE_TAG="latest"
+          make docker-manifest DOCKER_IMAGE_TAG="latest" DOCKER_REPO=quay.io/prometheus
         fi
-    - run: make docker-publish
-    - run: make docker-publish DOCKER_REPO=quay.io/prometheus
 
 workflows:
   version: 2
--- a/Makefile.common
+++ b/Makefile.common
@@ -80,6 +80,12 @@ BIN_DIR                 ?= $(shell pwd)
 DOCKER_IMAGE_TAG        ?= $(subst /,-,$(shell git rev-parse --abbrev-ref HEAD))
 DOCKER_REPO             ?= prom
 
+DOCKER_ARCHS            ?= amd64
+
+BUILD_DOCKER_ARCHS = $(addprefix common-docker-,$(DOCKER_ARCHS))
+PUBLISH_DOCKER_ARCHS = $(addprefix common-docker-publish-,$(DOCKER_ARCHS))
+TAG_DOCKER_ARCHS = $(addprefix common-docker-tag-latest-,$(DOCKER_ARCHS))
+
 ifeq ($(GOHOSTARCH),amd64)
         ifeq ($(GOHOSTOS),$(filter $(GOHOSTOS),linux freebsd darwin windows))
                 # Only supported on amd64
@@ -178,17 +184,28 @@ common-tarball: promu
 	@echo ">> building release tarball"
 	$(PROMU) tarball --prefix $(PREFIX) $(BIN_DIR)
 
-.PHONY: common-docker
-common-docker:
-	docker build -t "$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)" .
+.PHONY: common-docker $(BUILD_DOCKER_ARCHS)
+common-docker: $(BUILD_DOCKER_ARCHS)
+$(BUILD_DOCKER_ARCHS): common-docker-%:
+	docker build -t "$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME)-linux-$*:$(DOCKER_IMAGE_TAG)" \
+		--build-arg ARCH="$*" \
+		--build-arg OS="linux" \
+		.
 
-.PHONY: common-docker-publish
-common-docker-publish:
-	docker push "$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME)"
+.PHONY: common-docker-publish $(PUBLISH_DOCKER_ARCHS)
+common-docker-publish: $(PUBLISH_DOCKER_ARCHS)
+$(PUBLISH_DOCKER_ARCHS): common-docker-publish-%:
+	docker push "$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME)-linux-$*:$(DOCKER_IMAGE_TAG)"
 
-.PHONY: common-docker-tag-latest
-common-docker-tag-latest:
-	docker tag "$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)" "$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME):latest"
+.PHONY: common-docker-tag-latest $(TAG_DOCKER_ARCHS)
+common-docker-tag-latest: $(TAG_DOCKER_ARCHS)
+$(TAG_DOCKER_ARCHS): common-docker-tag-latest-%:
+	docker tag "$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME)-linux-$*:$(DOCKER_IMAGE_TAG)" "$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME)-linux-$*:latest"
+
+.PHONY: common-docker-manifest
+common-docker-manifest:
+	DOCKER_CLI_EXPERIMENTAL=enabled docker manifest create -a "$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)" $(foreach ARCH,$(DOCKER_ARCHS),$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME)-linux-$(ARCH):$(DOCKER_IMAGE_TAG))
+	DOCKER_CLI_EXPERIMENTAL=enabled docker manifest push "$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)"
 
 .PHONY: promu
 promu: $(PROMU)
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,5 @@ data/
 .tarballs/
 
 !.build/linux-amd64/
+!.build/linux-armv7/
+!.build/linux-arm64/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,12 @@
-FROM        quay.io/prometheus/busybox:latest
+ARG ARCH="amd64"
+ARG OS="linux"
+FROM quay.io/prometheus/busybox-${OS}-${ARCH}:latest
 LABEL maintainer="The Prometheus Authors <prometheus-developers@googlegroups.com>"
 
-COPY prometheus                             /bin/prometheus
-COPY promtool                               /bin/promtool
+ARG ARCH="amd64"
+ARG OS="linux"
+COPY .build/${OS}-${ARCH}/prometheus        /bin/prometheus
+COPY .build/${OS}-${ARCH}/promtool          /bin/promtool
 COPY documentation/examples/prometheus.yml  /etc/prometheus/prometheus.yml
 COPY console_libraries/                     /usr/share/prometheus/console_libraries/
 COPY consoles/                              /usr/share/prometheus/consoles/
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Needs to be defined before including Makefile.common to auto-generate targets
+DOCKER_ARCHS ?= amd64 armv7 arm64
+
 include Makefile.common
 
 STATICCHECK_IGNORE = \


brian-brazil:
This only seems to be creating docker images, this should be consistent with other parts of our builds such as tarballs on github. We should also avoid providing 32bit versions of Prometheus, as that won't work well.

johanneswuerbach:
@brian-brazil I'm not entirely sure I can follow. This PR mainly builds and publishes the already existing linux/armv7 (32bit) and linux/arm64 as docker images and the tarballs already exist for those.

The main purpose of this PR is to provide docker images for architectures besides amd64 (the only one provided today) as requested for example here https://github.com/prometheus/promu/issues/89

I added ARM images here as a starting point, but other architectures should be fairly easy to add.

> avoid providing 32bit versions of Prometheus

The linux/armv7 binary is already being published today and this mainly makes using prometheus easier on a Raspberry Pi 3 as now users can just pull prom/prometheus and don't have to fallback to a prom raspberry fork like https://hub.docker.com/r/napnap75/rpi-prometheus/, which is actually the main reason why I created this PR.

.circleci/config.yml
```diff
@@ -41,15 +41,14 @@ jobs:
         destination: /build
 
   docker_hub_master:
-    executor: golang
+    machine:
+      image: circleci/classic:201808-01
 
     steps:
     - checkout
-    - setup_remote_docker
```
discordianfish: Dropping setup_remote_docker means the docker builds now run locally in the primary container. I don't think this has any downsides beside maybe worse caching. But not sure and something to keep in mind.

discordianfish:
Nice, LGTM but see my code comment.

Makefile.common
```diff
@@ -175,15 +175,38 @@ common-tarball: promu
 
 .PHONY: common-docker
 common-docker:
-	docker build -t "$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)" .
+	docker build -t "$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)" \
+		--build-arg BASE_IMAGE="quay.io/prometheus/busybox:latest" \
+		--build-arg PROMETHEUS_PATH=".build/linux-amd64/prometheus" \
```
brian-brazil: This won't work for other repos, Makefile.common is used elsewhere too.

discordianfish: This would require changes to all other repos too, that's right. But I still feel it's reasonable to do that.

brian-brazil: That's the whole point of Makefile.common, we copy it directly to other repos.

.circleci/config.yml
johanneswuerbach: As far as I know caching is anyway a paid opt-in feature and would require setting `docker_layer_caching: true` as described here https://circleci.com/docs/2.0/building-docker-images/

With the prometheus docker images only being small wrappers around the already built go binaries based on ~5mb busybox, I don't think re-pulling and re-building on every run is really a concern here.

Makefile.common
johanneswuerbach: I'm happy to change it to whatever direction you prefer, but I'm currently not sure whom to follow.

The building itself could we just multiple calls to `make docker` with `BASE_IMAGE`, `BUILD_DIR`  and `DOCKER_IMAGE_NAME` as parameters. Building the manifest itself might be a bit more tricky as it requires knowledge of all included architectures, but nothing unsolvable.

johanneswuerbach:
> only a few GB of data

I would assume this isn't really a concern for most 32bit ARM use-cases as those are mainly Raspberry PIs and similar small compute platforms afaik, which have limited processing and storage capabilities anyway.

E.g. I'm currently running prom ARM32v7 on a Raspberry PI 3 with storage provided by a 16 GB sd card to collect some home automation stats and don't think I'll actually ever reach > 1 gb of collected data, but still find the prom platform with the variety of exporters providing a lot of value for my small use-case.

ARM setups with potentially more data might be in datacenter / cloud environments, but for example those new AWS ARM instances https://aws.amazon.com/blogs/aws/new-ec2-instances-a1-powered-by-arm-based-aws-graviton-processors/ are ARM64v8 afaik, similar to https://www.scaleway.com/virtual-cloud-servers/#anchor_arm and https://www.packet.com/cloud/servers/c1-large-arm/

Makefile.common
johanneswuerbach: @brian-brazil I pushed https://github.com/prometheus/node_exporter/pull/1207 with much more make "magic" to dynamically generate targets based on a per projects `DOCKER_ARCHS` variable, let me know whether which approach you prefer.

brian-brazil: @simonpasquier and @SuperQ are the ones more looking after this area.

discordianfish: @johanneswuerbach The goal here is to have the Makefile.common work without changes for all prometheus repositories, so we can't hardcode build args that are specific to a repo (like PROMETHEUS_PATH).

Not sure what the best way to solve this would be. Something like `BIN_PATH` in prometheus/node_exporter#1207 might work. Somebody would need to make sure this works for all other repos.

discordianfish:
To follow up on this:

1.) I think we should provide docker images for both platforms
2.) This is the right place to add the functionality
3.) ..but it needs to be added in a way that works for all repos

johanneswuerbach:
@discordianfish I pushed https://github.com/prometheus/prometheus/pull/5031/commits/8071a3fcf1e7a57febc3d5c5e57997eb2de706e1, which changes this PR to also use an approach similar to prometheus/node_exporter#1207

As a test https://github.com/prometheus/node_exporter/pull/1207 pushes docker images for 4 different architectures, while this PR only pushes three of them (no ppc64le) and the only difference is in `DOCKER_ARCHS` https://github.com/prometheus/prometheus/pull/5031/files#diff-b67911656ef5d18c4ae36cb6741b7965R15, while the `circle.yml` and `Makefile.common` are identical.

As `DOCKER_ARCHS` defaults to `AMD64`, other repositories should support the same `Makefile.common` without any changes, except calling `make docker-publish` explicitly for each to be released tag in CI. I can happily revert this, but I find the implicit pushing of multiple tags a bit dangerous and accompanying `make docker-manifest` only works with an explicit tag.

> 3.) ..but it needs to be added in a way that works for all repos

I can open PRs for more repos if you think the above looks good.

Makefile.common
```diff
@@ -173,17 +195,28 @@ common-tarball: promu
...
-	docker tag "$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)" "$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME):latest"
+.PHONY: common-docker $(BUILD_DOCKER_ARCHS)
+common-docker: $(BUILD_DOCKER_ARCHS)
+$(BUILD_DOCKER_ARCHS): common-docker-%:
+	docker build -t "$(DOCKER_REPO)/$(DOCKER_$*_IMAGE_NAME):$(DOCKER_IMAGE_TAG)" \
+		--build-arg BASE_IMAGE="$(DOCKER_$*_BASE_IMAGE)" \
+		--build-arg BIN_PATH="$(DOCKER_$*_BIN_PATH)" \
+		.
+
+.PHONY: common-docker-publish $(PUBLISH_DOCKER_ARCHS)
```
discordianfish: Not much of a Makefile expert but I don't think `$(PUBLISH_DOCKER_ARCHS)` should be here, right? It should only mark the target itself as PHONY?

Makefile.common
```diff
@@ -173,17 +195,28 @@ common-tarball: promu
...
+$(PUBLISH_DOCKER_ARCHS): common-docker-publish-%:
+	docker push "$(DOCKER_REPO)/$(DOCKER_$*_IMAGE_NAME):$(DOCKER_IMAGE_TAG)"
+
+.PHONY: common-docker-tag-latest $(TAG_DOCKER_ARCHS)
+common-docker-tag-latest: $(TAG_DOCKER_ARCHS)
+$(TAG_DOCKER_ARCHS): common-docker-tag-latest-%:
+	docker tag "$(DOCKER_REPO)/$(DOCKER_$*_IMAGE_NAME):$(DOCKER_IMAGE_TAG)" "$(DOCKER_REPO)/$(DOCKER_$*_IMAGE_NAME):latest"
+
+.PHONY: common-docker-manifest
+common-docker-manifest:
```
discordianfish: What's the rational for 'docker manifest'? Haven't used it so not sure we need this.

discordianfish:
Nice! In general this looks good to me, see inline comments though.
Also agree on calling `make docker-publish` explicitly with the ARCHS to build for.

The only thing I'm not sure about is the BIN_PATH stuff.. That seems a bit brittle. Maybe we should get rid of that, pass ARCH (and maybe OS) as build-arg and leave it to the Dockerfile to figure out where to get the binaries from, e.g:

```
ARG ARCH="amd64"
ARG OS="linux"
FROM quay.io/prometheus/busybox-${ARCH}-${OS}:latest
...
COPY .build/${OS}-${ARCH}/prometheus  /bin/prometheus
COPY .build/${OS}-${ARCH}/promtool       /bin/promtool
```

To make this nice we'd need to rename the default busybox image for this though.

Alternatively we could only use the build args for the base image and set docker context to `.build/[os-arch]`, e.g: `docker build -f ./Dockerfile .build/linux-amd64`. But since we have to use the build args to figure out the base image somehow, it probably makes sense to use the former.

Makefile.common
johanneswuerbach: The manifest contains references to all arch images and allows the client to just use prometheus/prometheus:tag to get the respective arch image automatically.

Docs: https://docs.docker.com/edge/engine/reference/commandline/manifest/

Makefile.common
johanneswuerbach: All of those are actually targets, which could also be called individually.

E.g. common-docker-publish-AMD64

And therefor I think it makes sense to make the all and the sub targets phony

Makefile.common
discordianfish: Ah interesting.. It also sounds like Kubernetes uses the same mechanism? https://kubernetes.io/docs/concepts/containers/images/#building-multi-architecture-images-with-manifests But it's not very clear on that.. but I suppose in that case we'd want to have it. Thanks for clarifying!

johanneswuerbach:
When choosing those names, my main reasoning was staying consistent with the official docker naming https://github.com/docker-library/official-images#architectures-other-than-amd64, but I can happily adjust it to be inline with promu.

Is there an example repo, where you think the `BIN_PATH` approach would break?

discordianfish:
@johanneswuerbach I just find it nice if it's up to the Dockerfile to reference the things it copies in. I'm fine either way though but not maintainer of this repo. So up to @brian-brazil @juliusv and @fabxc

brian-brazil:
I defer to you on Docker-related stuff, @SuperQ and @simonpasquier may care on the build front.

johanneswuerbach:
To recap: the remaining issue is, whether the released docker images should follow the promu style `GOOS-GOARCH` naming, which would allow to only use two build args `OS` and `ARCH`.

Example resulting docker file:
```
ARG ARCH="arm64"
ARG OS="linux"
FROM quay.io/prometheus/busybox-${ARCH}-${OS}:latest
...
COPY .build/${OS}-${ARCH}/prometheus  /bin/prometheus
COPY .build/${OS}-${ARCH}/promtool       /bin/promtool
```

or whether the docker images should follow the official docker image naming https://github.com/docker-library/official-images#architectures-other-than-amd64 (e.g. armv8 vs arm64), which would require to inject three build args: os, promu arch and docker arch.

Example resulting docker file:
```
ARG PROMU_ARCH="arm64"
ARG DOCKER_ARCH="armv8"
ARG OS="linux"
FROM quay.io/prometheus/busybox-${DOCKER_ARCH}-${OS}:latest
...
COPY .build/${OS}-${PROMU_ARCH}/prometheus  /bin/prometheus
COPY .build/${OS}-${PROMU_ARCH}/promtool       /bin/promtool
```

I don't really have a strong opinion (or any concerns) with any of the two and will happily change this PR to the preferred solution.

discordianfish:
I tend towards using only OS/ARCH. If I understand this correctly, with the docker manifest stuff this doesn't really matter much anyway.

johanneswuerbach:
@discordianfish , last open question. What would you do for amd64?

Re-push them as `prometheus-amd64-linux` or leave them as they are today?

johanneswuerbach:
@discordianfish, any update on the above?

simonpasquier:
LGTM for the Makefile/CI parts. Thanks for all the efforts @johanneswuerbach!

discordianfish:
Don't care much about prometheus-amd64-linux vs keeping. Keeping it is probably less involved so lets do that.

simonpasquier:
@SuperQ any feedback on this PR?

.circleci/config.yml
SuperQ: We should switch this back to remote_docker, like we did in busybox: https://github.com/prometheus/busybox/pull/20

Makefile.common
```diff
@@ -79,6 +79,28 @@ BIN_DIR                 ?= $(shell pwd)
...
 DOCKER_REPO             ?= prom
 
+DOCKER_ARCHS            ?= AMD64
+
+DOCKER_AMD64_IMAGE_NAME ?= $(DOCKER_IMAGE_NAME)
+DOCKER_AMD64_BASE_IMAGE ?= quay.io/prometheus/busybox:latest
+DOCKER_AMD64_BIN_PATH   ?= .build/linux-amd64
+
+DOCKER_ARM32V7_IMAGE_NAME ?= $(DOCKER_IMAGE_NAME)-arm32v7-linux
+DOCKER_ARM32V7_BASE_IMAGE ?= quay.io/prometheus/busybox-arm32v7-linux:latest
```
SuperQ: Maybe we should add a `busybox-amd64-linux` image as well, then we can get rid of this variable copy-pasta.

.circleci/config.yml
johanneswuerbach: Done

Makefile.common
johanneswuerbach: @SuperQ reworked here https://github.com/prometheus/busybox/pull/23

vielmetti:
Is there anything remaining in your review before this is ready to merge, @simonpasquier ?

brian-brazil:
:+1:

johanneswuerbach:
Quick update, we discovered an issue with quay.io not supporting v2 docker manifests https://github.com/prometheus/busybox/issues/25#issuecomment-467329736 and therefor need to special-case amd64 (at least on quay.io)

New proposed layout:

- registry/prometheus (manifest on docker hub, amd64 image on quay.io)
- registry/prometheus-linux-amd64
- registry/prometheus-linux-armv7
- registry/prometheus-linux-armv8

Eventually the amd64 image on https://quay.io/repository/prometheus/prometheus can be replaced by a docker https://docs.docker.com/edge/engine/reference/commandline/manifest/ so client can fetch to respective architecture automatically.

SuperQ:
@johanneswuerbach Thanks, that proposal seems fine. It's too bad we have to special-case quay.io. :disappointed:

Adirio:
Once this gets merged, will all the kube-prometheus project images have multi-architecture images? I'm specially interested in ARM32v7.

Is there anything I can do to help finish this?

johanneswuerbach:
@Adirio for other projects the common Makefile needs to be updated and respective docker repositories to be created, but that is the plan I think.

Adirio:
Should that plan be tracked somewhere? Maybe in prometheus/promu#89?

alexellis:
/subscribe

Looking forward to seeing this for ARMHF/ARM64 for both Prometheus and AlertManager so the OpenFaaS community can deprecate our on-device builds of both -> https://github.com/openfaas/faas/tree/master/contrib

johanneswuerbach:
This should be ready now, but before merging this the following repositories need to be created on docker hub / quay.

- prometheus-linux-amd64
- prometheus-linux-armv7
- prometheus-linux-arm64

Test images generated using

```
promu crossbuild -v -p "linux/amd64 linux/armv7 linux/arm64"
DOCKER_REPO=johanneswuerbach DOCKER_IMAGE_TAG=master make docker docker-publish docker-manifest
```

are available as `docker pull johanneswuerbach/prometheus:master` https://cloud.docker.com/repository/registry-1.docker.io/johanneswuerbach/prometheus/tags

and are correctly detected as multi-arch

image

Adirio:
Why are you using different repositories? I thought that multi-arch images supported refering to different labels in the same repository. Doesn't that structure make more sense?

johanneswuerbach:
tl;dr: There are multi ways to achieve this, but this seems to be most inline with is done in docker hub.

Generally its fairly modular so you can have multiple repositories, per arch tags or just the new manifests.

For just manifests I couldn't find a way to cross-build as `FROM x` needs to target a specific arch to trigger a cross-build and there is no explicit selection with manifests.

I found per arch tags confusion as this would mean a single tag would be here for each architecture https://hub.docker.com/r/prom/prometheus/tags

Therefor I went with multiple repositories as in https://github.com/docker-library/official-images#architectures-other-than-amd64 and a manifest in prom/prometheus, so using `docker pull prom/prometheus` on amd64 or a raspberry pi just works ™️ , but you can still pull a specific architecture using the per-arch repository if you need that.

Adirio:
There are 3 different levels: organization, repository and tag. And also 3 different category types to differentiate: library (prometheus, busybox, ...), architecture and versions.

The official libraries use organizations to separate architectures and repositories to hold each library. Conceptually this is a bit strange, but taking into account that they have one order of magnitude more libraries than architectures this may make sense. Tags are then used for versions.

Right now there was no multi-arch support for prometheus so there was a single organation (prom) that used repositories for each library and tags for versions. Now, when including multi-architecture there are several options:

1. Use organizations for architectures as the official libraries: this will keep one level per category type at the cost of having to manage multiple organizations. 
2. Keep a single organization: one of the levels will hold 2 category types but only an organization would need to be managed.

If choosing the second option there are another 2 options:

1. Use repositories for architectures.
2. Use tags for architectures.

Your choice was 2.1 which will have less tags per repository as you said but multiple repositories per library. 2.2 would have more tags per repository but just 1 repository per library.

johanneswuerbach:
I don't really have strong options on either, so it's up to you (the prometheus collaborators) to decide. So far that point never came up, so I'm a bit hesitant to change all the things, again, but can obviously do that for the greater good.

@Adirio what would be the best way to have such a decision and come to a final agreement?

Adirio:
I'm not a Prometheus member either so it is not up to me.

I just mentioned it as when I was designing one of my own repositories the first thing that came to my mind was something like 2.2 and when I read your comment I realized you where going on a different route that I didn't even think about so I listed all the options.

SuperQ:
@Adirio I'm no Docker expert, so your expertise / recommendations are appreciated.

Adirio:
The main difference is from a management perspective. Organizations need to be created and a password for them stored/saved/memorized. Repositories need to be created but no password required. Labels get auto-created so no need to manage them.

From the user point of view the only difference is how he should refer to the images as I wrote before:
> Option 1 would look like `prom-ARCH/LIBRARY:VERSION` (e.g. `prom-arm32v7/prometheus:latest`).
> Option 2.1 would look like `prom/LIBRARY-ARCH:VERSION` (e.g. `prom/prometheus-arm32v7:latest`).
> Option 2.2 would look like `prom/LIBRARY:ARCH-VERSION` (e.g. `prom/prometheus:arm32v7-latest`).
> In any of the options there would be the multi-arch manifests under `prom/LIBRARY:VERSION` (e.g. `prom/prometheus:latest`) that will be used most of the times.
The most used option will be the common `prop/prometheus:latest` that is valid for the 3 options and when needing to specify the architecture or the version any of the 3 options works fine if it is documented.

Both are two trivial differences so I would decide based on the following criteria:
- If following the official aproach is the most important criteria of these 3, go for option 1.
- If avoiding to create new organizations/repositories is the most imporatant criteria of these 3, go for option 2.2.
- If none of the above criterias are that important, stick with option 2.1 as this PR would not need to be changed.

SuperQ:
Thanks for the explanation.
* Creating new orgs would be a huge pain I would like to avoid. No to Option 1
* Creating libraries is only slightly annoying.

In either Option 2.1, the user has to specify non-amd64 architecture in the pull. This is annoying for the user, but I don't see there being much difference for the user between 2.1 and 2.2.

Adirio:
> Thanks for the explanation.

You are welcome

> * Creating new orgs would be a huge pain I would like to avoid. No to Option 1
> * Creating libraries is only slightly annoying.

Option 2.1 & 2.2 remaining

> In either Option 2.1, the user has to specify non-amd64 architecture in the pull. This is annoying for the user, but I don't see there being much difference for the user between 2.1 and 2.2.

I guess you mean either Option 2. And thats not right, I guess I was not clear enough. In either Option 2 you would have the `prom/prometheus/latest` with the multi-arch manifest, so no need to specify the arch. But if for any reason someone wanted to specify the arch they would need to write it the way I presented.

P.S.: I can not think of a single use case of referencing the specific archs directly when you can specify the multi-arch manifest and forget about it. You still need to have them so that the multi-arch can refer to them, but the user will always use the multi-arch manifest. You can consider `prom/prometheus:latest` as the proxy that redirects to either `prom/prometheus-arm32v7:latest` or `prom/prometheus:arm32v7-latest` depending on which option you choose.

johanneswuerbach:
> P.S.: I can not think of a single use case of referencing the specific archs directly when you can specify the multi-arch manifest and forget about it.

cross-building is a least one use-case, which doesn't work with only a multi-arch manifest. E.g. in this PR `FROM quay.io/prometheus/busybox-${OS}-${ARCH}:latest` needs to be done and using just `FROM quay.io/prometheus/busybox:latest` won't work as there is no way to force a different architecture then the local one afaik.

kfox1111:
Mirroring also benefits from being able to reference it directly. Though it may be easier now.

Thanks,
Kevin
________________________________

Adirio:
I was thinking from the user perspective, the developer of other images would need to target the specific archs thats true.

Kevin, prometheus is actually mirrored in quay.io and docker hub, and obviously both the specific images for each architectures and the multi arch manifest are mirrored. But I don't see that as being a use case.

What I was talking is that no user should target the specific images over the multi arch manifest.

simonpasquier:
Pondering on the discussion and I think that @johanneswuerbach has a point with users leveraging the Prometheus image as a base image for building another image. As Ben noted, creating per-arch repositories is a bit painful but it shouldn't happen too frequently. Hence my inclination would be to go with option 2.1.

Adirio:
> Pondering on the discussion and I think that @johanneswuerbach has a point with users leveraging the Prometheus image as a base image for building another image.

I think my point was not clear. I was never suggesting not having arch specific images, you need them for the multi-arch manifest. I was just saying that the most common case will be linking to the multi-arch manifest and letting the node determine which image it wants. So the differences writting 1, 2.1 or 2.2 are minimal. As you all said when building a docker image you need to reference the specific architecture but image developers are expected to have a stronger background than the common user.

discordianfish:
If docker registries would be just http servers.. Remember these discussions vividly :).
Either way, looks like we need to decide on something so I say let's go with one org and one repo per app-arch as @johanneswuerbach initially suggested.

discordianfish:
Docker seems to have allowed 'lazy' repo creation by just pushing to some name, so I don't think we need to create anything anymore, correct? So I'd be all up for merging this now. We can still iterate on this later to come up with an even more convenient way.

kfox1111:
yeah. pushing to a non existing repo creates it.

Adirio:
> Docker seems to have allowed 'lazy' repo creation by just pushing to some name, so I don't think we need to create anything anymore, correct? So I'd be all up for merging this now. We can still iterate on this later to come up with an even more convenient way.

I think that the issue with repo creation was on quay.io not on dockerhub

I would also merge though

SuperQ:
Ok, I created the required repos on quay.io. I think this is ready for merge.

alexellis:
Hi all, I can't seem to find the arm/arm64 images.

Can you tell me where they are going to be published?

Thanks,

Alex

no-arm

simonpasquier:
@alexellis you'd have to wait for v2.10.0 as this PR has been merged after v2.9.0 got released. Multi-arch images are already available for master though:

image

PR #5054, opened 2019-01-02
Move err to be the last return value from storage.Select.

### LINKED ISSUE #4956: Querier.Select now returns a third thing 'warnings' - can't pass golint...
As of #4832 we've added a third parameter to `Querier.Select`, but this makes it impossible to implement this interface in a golint clean way:

```
./pkg/querier/querier.go:114:1: error should be the last type when returning multiple items
```

We'd like to stay golint clean in Cortex; anyone object to me making error last in that interface?  And perhaps adding golint to our CI...

tomwilkie:
Fixes #4956 


brian-brazil:
:+1:

PR #5056, opened 2019-01-03
Add test case for promtool

MIBc:
Add test case TestQueryLabels, TestQueryInstant, TestQuerySeries for promtool.


codesome:
@simonpasquier could you take a look at this?

bwplotka:
It looks good and it sounds like it's still relevant, but it might be easier to start from scratch given all changes, so closing.

Really sorry for massive lag, we'll try to do better next time! 🤗 

PR #5057, opened 2019-01-03
*: use latest release of staticcheck

simonpasquier:
The latest release of staticcheck (2019.1) now supports Go modules so we don't have to rely on the `next` branch anymore. Note that this version introduces more checks as staticcheck also includes the gosimple, unused, and megacheck programs (see [changelog](https://staticcheck.io/changes/2019.1)). It is also possible to tune the [configuration](https://staticcheck.io/docs/#configuration) with a local `staticcheck.conf` file as well as defining exceptions directly in the code (using `//lint:ignore ...` comments).

The PR also fixes a couple of things in the code flagged by the additional checks.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -73,6 +73,8 @@ endif
 
 PROMU_VERSION ?= 0.2.0
 PROMU_URL     := https://github.com/prometheus/promu/releases/download/v$(PROMU_VERSION)/promu-$(PROMU_VERSION).$(GO_BUILD_PLATFORM).tar.gz
+STATICCHECK_VERSION ?= 2019.1
+STATICCHECK_URL     := https://github.com/dominikh/go-tools/releases/download/$(STATICCHECK_VERSION)/staticcheck_$(GOHOSTOS)_$(GOHOSTARCH)
 
 PREFIX                  ?= $(shell pwd)
 BIN_DIR                 ?= $(shell pwd)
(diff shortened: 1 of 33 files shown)


web/federate.go
```diff
@@ -90,7 +90,7 @@ func (h *Handler) federation(w http.ResponseWriter, req *http.Request) {
 		s, wrns, err := q.Select(params, mset...)
 		if wrns != nil {
 			level.Debug(h.logger).Log("msg", "federation select returned warnings", "warnings", wrns)
-			federationErrors.Add(float64(len(wrns)))
+			federationWarnings.Add(float64(len(wrns)))
```
discordianfish: Why? Or was the regex too greedy?

simonpasquier: staticcheck reported correctly that `federationWarnings` was unused. Looking at the [commit](https://github.com/prometheus/prometheus/commit/f0e9196dca9633e77a200be2d8805672f9255cfe#diff-c5eb08a58adae3a6dddb169f8523fc83) that introduced the variable, it looks like there was a typo.

promql/engine.go
```diff
@@ -592,7 +592,7 @@ func (ng *Engine) populateSeries(ctx context.Context, q storage.Queryable, s *Ev
 		}
 		return nil
 	})
-	return querier, err, warnings
+	return querier, warnings, err
```
brian-brazil: Hmm, is your branch up to date? All these were already meant to have been fixed.

web/federate.go
discordianfish: Ah, so this actually fixes a bug and we were counting the warnings as errors before? Okay, makes sense!

simonpasquier:
The latest [commit](https://github.com/prometheus/prometheus/pull/5057/commits/d7501c7107d1944bed46fd5631d51bca787ee0a0) gets the `staticcheck` binary from upstream instead of rebuilding it locally. This is similar to what we do for `promu`. Also because `staticcheck` isn't SemVer compliant (and [doesn't plan](https://github.com/dominikh/go-tools/issues/328#issuecomment-449020278) to be any time soon), the `go get` command pulls the `master` branch instead of the latest tag which can be annoying as new checks being added without notice.

I've also added a hack to avoid the spurious errors (as reported in https://github.com/prometheus/node_exporter/pull/1214 and https://github.com/prometheus/statsd_exporter/issues/173): before running staticcheck, we run `go list` to populate the local cache.

SuperQ:
@simonpasquier Thanks, I was going to propose doing the same thing. :grin:

Makefile.common
```diff
@@ -138,8 +140,12 @@ common-vet:
 .PHONY: common-staticcheck
 common-staticcheck: $(STATICCHECK)
 	@echo ">> running staticcheck"
+	chmod +x $(STATICCHECK)
 ifdef GO111MODULE
-	GO111MODULE=$(GO111MODULE) $(STATICCHECK) -ignore "$(STATICCHECK_IGNORE)" -checks "SA*" $(pkgs)
+# 'go list' needs to be executed before staticcheck to prepopulate the modules cache.
```
SuperQ: Can you indent these comments to match the lines below?

PR #5059, opened 2019-01-03
Expose rules.NewGroupMetrics and rules.Metrics.

### LINKED ISSUE #5058: No way to use rules.NewGroup without registering metrics twice.
Since #4961, there is no way to use `rules.NewGroup` without registering the metrics twice, because the `metrics` field of `ManagerOptions` is not public.

This is a problem for Cortex, see https://github.com/cortexproject/cortex/pull/1154

Any objections to making `metrics` and `newGroupMetrics` public?

tomwilkie:
Fixes #5058 

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -61,7 +61,8 @@ var (
 	)
 )
 
-type metrics struct {
+// Metrics for rule evaluation.
+type Metrics struct {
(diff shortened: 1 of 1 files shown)


brian-brazil:
:+1:

tomwilkie:
Thanks!

PR #5061, opened 2019-01-04
Fix typo in comment

MIBc:
I think it is a typo in comment of Alert.ResolvedAt .

### DIFF
--- a/notifier/notifier.go
+++ b/notifier/notifier.go
@@ -93,7 +93,7 @@ func (a *Alert) Resolved() bool {
 	return a.ResolvedAt(time.Now())
 }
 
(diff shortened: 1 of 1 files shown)


notifier/notifier.go
```diff
@@ -93,7 +93,7 @@ func (a *Alert) Resolved() bool {
 	return a.ResolvedAt(time.Now())
 }
 
-// ResolvedAt returns true off the activity interval ended before
+// ResolvedAt returns true iff the activity interval ended before
```
codesome:
```suggestion
// ResolvedAt returns true if the activity interval ended before
```

simonpasquier: iff = if and only if so the change is correct.

Issue #5064, opened 2019-01-04
Promtool test rules - out of order sample on v2.6.0 but worked on v.2.5.0


kentsanggds:
## Bug Report
**What did you do?**
Ran `promtool test rules <rule test file>`

Rules file - https://github.com/alphagov/prometheus-aws-configuration-beta/blob/4dfb2058b805bb6dee32eb4fbc129d4be14bf70c/terraform/modules/app-ecs-services/config/alerts-tests/test-observe-alerts.yml

**What did you expect to see?**

Tests passing, they were passing on Prometheus v2.5.0

**What did you see instead? Under which circumstances?**

Tests failed with - `out of order sample`

**Environment**

* System information:

	insert output of `uname -srm` here

`Darwin 17.7.0 x86_64`

* Prometheus version:

	insert output of `prometheus --version` here

```
prometheus, version 2.6.0 (branch: non-git, revision: non-git)
  build user:       brew@HighSierra.local
  build date:       20181220-02:53:59
  go version:       go1.11.4
```

codesome:
https://github.com/prometheus/prometheus/pull/4851 might possible be the reason, not sure, will have a look at it soon.

kentsanggds:
has this been fixed?

codesome:
@kentsanggds Yes, [the fix](https://github.com/prometheus/prometheus/pull/5069) is included in [`v2.6.1`](https://github.com/prometheus/prometheus/releases/tag/v2.6.1) (released yesterday).

kentsanggds:
Great, will ty it out

PR #5069, opened 2019-01-05
Empty slice after last append in promql test LazyLoader

codesome:
Fixes https://github.com/prometheus/prometheus/issues/5064

Apparently the slice of samples to be loaded into the storage was not emptied after the last sample was added. This would cause any future call to `WithSamplesTill` to attempt append again, and error out.


beorn7:
:+1: 

With @simonpasquier being the shepherd for 2.6, I guess that's sufficient to merge this.

(I'll canary this at SC, and will also take care of the forward-merge into master.)

Issue #5067, opened 2019-01-05
Marathon_SD Host Target Address  when using Virtual Networks


Smithx10:
## Proposal
While using the Marathon_SD configuration with a prometheus server external to Mesosphere Cluster ( _which uses marathon_ ) I ran into a situation where the Port's Target Address would use the Virtual IP instead of the HOST IP,  which is fair if the Prometheus Server was internal to the DC/OS cluster.  

I added an external option to the config which will bias the ports Target Address to the Host IP which is routable externally.  If omitted,  marathon_sd behaves the same as before.  So this should be a non breaking change.

https://github.com/Smithx10/prometheus/commit/2780ee59e33fb4d3365193035ad3b90327a85d56

In IRC someone suggested relabelling, but I don't believe there is any meta labels that would facilitate this behaviour. 

```
__meta_marathon_app: the name of the app (with slashes replaced by dashes)
__meta_marathon_image: the name of the Docker image used (if available)
__meta_marathon_task: the ID of the Mesos task
__meta_marathon_app_label_<labelname>: any Marathon labels attached to the app
__meta_marathon_port_definition_label_<labelname>: the port definition labels
__meta_marathon_port_mapping_label_<labelname>: the port mapping labels
__meta_marathon_port_index: the port index number (e.g. 1 for PORT1)
```
https://prometheus.io/docs/prometheus/latest/configuration/configuration/#marathon_sd_config

PR #5074, opened 2019-01-07
added meta_marathon_host label

Smithx10:
This pull request should address the criticism received in a prior pull request that hard coded logic in the marathon_sd configuration.  

It addresses the following issue:
https://github.com/prometheus/prometheus/issues/5067

### DIFF
--- a/discovery/marathon/marathon.go
+++ b/discovery/marathon/marathon.go
@@ -51,6 +51,7 @@ const (
 	taskLabel model.LabelName = metaLabelPrefix + "task"
 	// hostLabel contains the mesos task host address of the app instance
 	hostLabel model.LabelName = metaLabelPrefix + "host"
+
 	// portMappingLabelPrefix is the prefix for the application portMappings labels.
 	portMappingLabelPrefix = metaLabelPrefix + "port_mapping_label_"
 	// portDefinitionLabelPrefix is the prefix for the application portDefinitions labels.
--- a/discovery/marathon/marathon.go
+++ b/discovery/marathon/marathon.go
@@ -49,7 +49,8 @@ const (
 	portIndexLabel model.LabelName = metaLabelPrefix + "port_index"
 	// taskLabel contains the mesos task name of the app instance.
 	taskLabel model.LabelName = metaLabelPrefix + "task"
-
+	// hostLabel contains the mesos task host address of the app instance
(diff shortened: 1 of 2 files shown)


simonpasquier:
You'd need to update https://github.com/prometheus/prometheus/blob/master/docs/configuration/configuration.md too. And sign your commit for DCO.

discovery/marathon/marathon.go
```diff
@@ -49,7 +49,8 @@ const (
 	portIndexLabel model.LabelName = metaLabelPrefix + "port_index"
 	// taskLabel contains the mesos task name of the app instance.
 	taskLabel model.LabelName = metaLabelPrefix + "task"
-
+	// hostLabel contains the mesos task host address of the app instance
```
brian-brazil: full stop

Smithx10: I believe the full stop was used to group port labels together. 

So the stop probably should be after the hostLabel Line correct?

See the following:
https://github.com/prometheus/prometheus/blob/master/discovery/marathon/marathon.go#L52

brian-brazil: The style in Go is to have comments as full sentences, which includes a full stop. So all of these should have a full stop.

docs/configuration/configuration.md
```diff
@@ -832,6 +832,7 @@ The following meta labels are available on targets during [relabeling](#relabel_
 * `__meta_marathon_app`: the name of the app (with slashes replaced by dashes)
 * `__meta_marathon_image`: the name of the Docker image used (if available)
 * `__meta_marathon_task`: the ID of the Mesos task
+* `__meta_marathon_host`: the Host Address of the Mesos task
```
brian-brazil: host address, neither are proper nouns

roidelapluie:
@Smithx10 Are you still interested by this?

Smithx10:
@roidelapluie,

Nope.  Luckily I no longer have to use DC/OS.  :)

codesome:
This is still relevant. @Smithx10 could you fix the review comments and the DCO? Thanks!

bwplotka:
Sounds like this is relevant, but author is not interested to pursue this. Also Mesos is on path to deprecation. 

If anyone wants to use it and fix it, this closed PR will be a good inspiration (: 

Thanks @Smithx10 anyway!

PR #5090, opened 2019-01-11
Add metric rule_group_rules_loaded to get the number of rules loaded

### LINKED ISSUE #5084: Feature request: export the number of rules loaded by prometheus
## Proposal

It would be nice to have a guage exported by prometheus for the number of rules that are loaded. It would be useful when you are verifying large changes. Did you lose rules, or are the rules added like you expected.  I was bit surprised that prometheus doesn't already export this information.

vn-ki:
@brian-brazil 

Closes #5084 

I'm not 100% whether this is the way to solve it, but I tried out with a few rule files and it seems to be working correctly.

I could not find any tests for the metrics testing, so I did not add tests.

Signed-off-by: Vishnunarayan K I <appukuttancr@gmail.com>


rules/manager.go
```diff
@@ -127,6 +128,14 @@ func NewGroupMetrics(reg prometheus.Registerer) *Metrics {
 			},
 			[]string{"rule_group"},
 		),
+		groupRulesLoaded: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Namespace: namespace,
+				Name:      "rule_group_rules_loaded",
```
brian-brazil: I don't think we need the _loaded

vn-ki:
@brian-brazil I have made the required changes

PR #5091, opened 2019-01-11
Fixing unchecked error

saromanov:
Hi! I've just fixed unchecked error. Please check it

### DIFF
--- a/discovery/azure/azure.go
+++ b/discovery/azure/azure.go
@@ -552,7 +552,7 @@ func (client *azureClient) getNetworkInterfaceByID(networkInterfaceID string) (n
 
 	result, err = client.nic.GetResponder(resp)
 	if err != nil {
-		err = autorest.NewErrorWithError(err, "network.InterfacesClient", "Get", resp, "Failure responding to request")
+		return result, autorest.NewErrorWithError(err, "network.InterfacesClient", "Get", resp, "Failure responding to request")
 	}
 
 	return result, nil


discovery/azure/azure.go
```diff
@@ -552,7 +552,7 @@ func (client *azureClient) getNetworkInterfaceByID(networkInterfaceID string) (n
 
 	result, err = client.nic.GetResponder(resp)
 	if err != nil {
-		err = autorest.NewErrorWithError(err, "network.InterfacesClient", "Get", resp, "Failure responding to request")
+		return result, autorest.NewErrorWithError(err, "network.InterfacesClient", "Get", resp, "Failure responding to request")
```
aixeshunter: I think it's unnecessary.

gouthamve: Hmm, can you explain why? This looks like a valid change to make and if nothing, we should atleast document why the error is being ignored.

From what I see, when it errors, the result will be empty.

cstyan: Agreed, looks to me like this is an error that is slipping through the cracks right now. Might be nicer to declare `var err error` at the top like `result` and keep the `err = autorest...` line as is but change the return at the end of the function to `return result, err`?

simonpasquier:
This is a duplicate of #4719

discovery/azure/azure.go
pytimer: I see the code that invoke this method and this result unused if err is not nil. So i think is it fixes `return result,...` to `return nil, autorest.NewErrorWithError()` will be better?

jannickfahlbusch:
#4719 just got merged. You can close this PR. 

Thanks for your effort :)

Issue #5093, opened 2019-01-14
Zookeeper connection leak in Discovery


ioanvapi:
**What did you do?**
Used prometheus with zookeeper for service discovery then reloaded prometheus (SIGHUP) several times.

**What did you expect to see?**
From pprof perspective, prometheus keeps the same amount of goroutines and a similar number of process file descriptors related to the zookeeepr connections.

**What did you see instead? Under which circumstances?**
After several prometheus service reload ("killall -SIGHUP prometheus") we observed the number of files descriptors for zookeeper connections are increased. Also, the number of goroutines related to the Discovery.Run() increased in /debug/pprof/goroutine.
We created a cron that reloads the prometheus service at every 20 mins and we observed the above behaviour. 

**Environment**
* System information:

	Linux 3.10.0-862.14.4.el7.x86_64 x86_64

* Prometheus version:

  prometheus, version 2.6.0 (branch: HEAD, revision: dbd1d58c894775c0788470944b818cc724f550fb)
  build user:       root@bf5760470f13
  build date:       20181217-15:14:46
  go version:       go1.11.3

* Code in prometheus/discovery/zookeeper/zookeeper.go:
```
func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
	defer func() {
		for _, tc := range d.treeCaches {
			tc.Stop()
		}
		// Drain event channel in case the treecache leaks goroutines otherwise.
		for range d.updates {
		}
		d.conn.Close()
	}()
```

* Pprof/goroutines
```
7 @ 0x42dfab 0x43df5d 0x14e1213 0x45bcc1
#	0x14e1212	github.com/prometheus/prometheus/discovery/zookeeper.(*Discovery).Run+0x122	/app/discovery/zookeeper/zookeeper.go:175
```

* Description
In the above code we observe the conn.Close() is invoked if the above range over the channel updates is terminated. Since the d.updates channel is never closed in the code then the following happens:
  -  d.conn.Close() is never invoked and the result is a leak of zookeeper connections
  -  defer func() is never existed and the result is a leak of goroutine that invoked the Run() method

ioanvapi:
I have replaced the below code in func (d *Discovery) Run():
```
        for range d.updates {
         }
```
with the code: 
```
         L:
		for {
			select {
			case <-d.updates:
			default:
				break L
			}
		}
		d.conn.Close()
	}()
``` 
[...]

PR #2669, opened 2017-04-30
Fix reload of ZooKeeper service discovery config

StephanErb:
Rational:

* When the config is reloaded and the provider context is canceled, we need to
  exit the current ZK `TargetProvider.Run` method as a new provider will be
  instantiated.
* In case `Stop` is called on the `ZookeeperTreeCache`, the update/events
  channel may not be closed as it is shared by multiple caches and would
  thus be double closed.
* Stopping all `zookeeperTreeCacheNode`s on teardown ensures all associated
  watcher go-routines will be closed eagerly rather than implicityly on
  connection close events.

Verification:
* Manually verified using ZooKeeper serversets generated by Apache Aurora.

Fixes #2481

PR #5095, opened 2019-01-14
fix for zookeeper connection leak

ioanvapi:
issue: https://github.com/prometheus/prometheus/issues/5093#issue-398814606


discovery/zookeeper/zookeeper.go
```diff
@@ -166,7 +166,13 @@ func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
 			tc.Stop()
 		}
 		// Drain event channel in case the treecache leaks goroutines otherwise.
-		for range d.updates {
+	L:
+		for {
+			select {
+			case <-d.updates:
```
brian-brazil: Why is this channel not getting closed?

ioanvapi: It was closed in a previous version (treecache.go) but at that moment it has generate a panic because of multiple close on the same channel. The same updates channel created in Discovery constructor has beep passed to multiple ZookeeperTreeCaches objects. 
When an stop event has come one tree cache has close the updates channel but the same will try the others tree caches and they will generate panic. Anyway, it was an unhappy solution to close a channel outside the code that created it when the channel is distributed to multiple goroutines.

That solution is explained in PR #2669.

brian-brazil: This seems racy, I think a better approach would be for Stop to wait until things are stopped.

jjneely:
I've been using this patch in production for several months now and it has definitely solved some ZK connection leak issues for us.  Love to see some progress on getting this merged.

PR #5096, opened 2019-01-14
Update WAL directory structure, and mention that the last three are always kept

Daniel15:
- Update directory structure to reflect how the WAL is actually a separate directory (my guess is that this changed as part of https://github.com/prometheus/tsdb/issues/131)
 - Mention that Prometheus will always keep the three most recent write ahead logs

### DIFF
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -13,7 +13,9 @@ Prometheus's local time series database stores time series data in a custom form
 
 ### On-disk layout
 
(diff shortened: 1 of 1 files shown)


SuperQ:
I would clarify this slightly. In the case of higher traffic Prometheus servers, there may be more than 3 WAL files. It needs to keep enough WAL for the last two hours in order to recover from a restart.

How about: "Prometheus keeps two hours of write-ahead-log files. It will also keep at least three or more write-ahead-log files."

brian-brazil:
Always a max of 128MB

SuperQ:
How about split up the paragraph a bit. It's a bit long.

One paragraph for the TSDB dirs. One for the WAL dir.

Daniel15:
I split the WAL info into a separate paragraph... Let me know what you think.

brian-brazil:
Thanks!

Issue #4790, opened 2018-10-26
Matching labels with remote_read does not work


lgogolin:
## Bug Report
**What did you do?**
Having prometheus-proxy with remote_read configuration that read data from two other prometheus. 

**What works ?**
```
up{project="monitoring",region="us-east-1"}
up{project="monitoring",region="eu-central-1"}
up{project="monitoring",region=~".*"}
```

**What does not work?**
```
up{project="monitoring",region=~"eu-central-1"}
up{project="monitoring",region=~"us-east-1"}
up{project="monitoring",region=~"us-east-1|eu-central-1"}
```

**Environment**

* Prometheus version:
```
prometheus, version 2.4.3 (branch: HEAD, revision: 167a4b4e73a8eca8df648d2d2043e21bdb9a7449)
  build user:       root@1e42b46043e9
  build date:       20181004-08:42:02
  go version:       go1.11.1
```

* Prometheus configuration file:
prometheus-proxy
```
remote_read:
        # monitoring - eu-central-1
        - url: http://prometheus-eu-central-1:9090/api/v1/read
          read_recent: true
          required_matchers:
            project: monitoring
        # # monitoring - us-eas-1
        - url: http://prometheus-us-east-1/api/v1/read
          read_recent: true
          required_matchers:
            project: monitoring
```
prometheus-eu-central-1:

```
global:
        external_labels:
          project: monitoring
          region: "eu-central-1"
          env: production

scrape_configs:
  ...
```

prometheus us-east-1:
```
global:
        external_labels:
          project: monitoring
          region: "us-east-1"
          env: production

scrape_configs:
  ...
```

brian-brazil:
The issue here is that the data inside the readee has no region label, it's only removed/added by the remote read logic and that can only deal with exact matches.

Federation is probably a better choice here.

lgogolin:
It's even weirder. After I restart POD with prometheus-proxy, first query works:

```
up{project="monitoring",region="us-east-1"}
```

But trying next even with other label, like:

```
up{project="monitoring",region="eu-central-1"}
```

is not returning data anymore

lgogolin:
Any update regarding this bug ?

brian-brazil:
We've looked at this as part of our bug scrub, this seems to be working as expected. If you've further questions you can 
ask them on the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users).

PR #4970, opened 2018-12-07
fix typo imeplements -> implements

SenXuDC:
fix a typo : imeplements -> implements
Signed-off-by: SenXuDC sen.xu@daocloud.io

PR #5104, opened 2019-01-17
Implement full matching on external labels for remote read

csmarchbanks:
This will allow users trying to do a remote read to a prometheus with
external labels to properly do matches besides external_label="value".

According to the docs
> In communication with external systems, they [external_labels] are always applied only when a time series does not have a given label yet and are ignored otherwise.

Which led to the development of the following logic:

For each matcher in a query, check if it pertains to an external label.
If so then following table shows how the matcher needs to be transformed
before querying local storage.

| Matcher | Matched external label | Did not match external label | 
| --- | --- | --- |
| =   | label=value OR label=""    | label=value AND label!="" |
| !=  | label!=value OR label=""   | label!=value AND label!="" |
| =~ | label=~value OR label="" | label=~value AND label!="" |
| !~  | label!~value OR label=""   | label!~value AND label!="" |

This can be simplified to:
* external label is matched, then the internal label must match the matcher OR be empty
* external label is not matched, then the internal label must match the matcher AND not be empty

Each time an OR is done, twice as many queries are done (one for each of the cases), then the results are merged together. If there is a better way of accomplishing an OR, I am happy to hear it!

Closes https://github.com/prometheus/prometheus/issues/4790

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -850,6 +850,7 @@ func (api *API) remoteRead(w http.ResponseWriter, r *http.Request) {
 	resp := prompb.ReadResponse{
 		Results: make([]*prompb.QueryResult, len(req.Queries)),
 	}
+	externalLabels := api.config().GlobalConfig.ExternalLabels.Clone()
 	for i, query := range req.Queries {
 		from, through, matchers, selectParams, err := remote.FromQuery(query)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thanks for your PR, I don't think this makes sense semantically. #4790 is the intended behaviour.

csmarchbanks:
What exactly does not make sense semantically? I am happy to change this around. I tried to make sure that external_labels do not override any internal labels as the docs suggest.

I have tested with queries similar to those found in #4790 and they have started working with this PR. In fact I was doing something similar to that when I found this bug.

brian-brazil:
We shouldn't be 2nd guessing the query the user has provided. In addition i don't think we should be adding features to support abuses of remote read.

csmarchbanks:
I don't agree that this is 2nd guessing the query the user has provided. It is making the query actually work. If I provide a query like `{region!="europe"}`, and a series comes back with region="europe" that is not what I expect as a user. But it is exactly what happens right now.

What would be a solution that doesn't support abuse cases to you? Allowing regular promql matchers to work doesn't seem abusive to me.

brian-brazil:
That sounds like you're doing cross-region remote read, which is not the intended use case. Do you have a use case that applies in the horizontal-sharding case?

csmarchbanks:
I don't personally do this, but I remember a presentation/talking to people at PromCon that do 1 prometheus per team. It could be nice to have a central prometheus that could do ad hoc queries across multiple teams. 

Do I personally absolutely need this fix? Probably not. But it is not much fun when promql does not behave how I expect when testing out remote read.

brian-brazil:
> It could be nice to have a central prometheus that could do ad hoc queries across multiple teams.

That's a good way to simultaneously take out all of your monitoring, and is thus explicitly not what remote read is for.

csmarchbanks:
I am going to try to refocus this conversation.

I was experimenting with remote read and discovered that promql was broken when trying to query using external_labels. I looked online, and found #4790, which means other people are trying similar queries to me. I had some time, so I decided to fix the issue. From your first comment, you agree that #4970 is the expected behavior, which is what I am trying to implement. To move forward:

1. Reject this PR/behavior change on the grounds of it supporting abusive remote read cases. That is your/the Prometheus Team's prerogative. I would ask that the associated issue is also closed with a comment about not fixing this behavior for future reference.
2. Please provide specific feedback on the implementation of this fix. I chose a method that emphasized correctness of results according to the tests and documentation I could find. I am happy to consider alternative implementation ideas.

brian-brazil:
> From your first comment, you agree that #4970 is the expected behavior, which is what I am trying to implement.

To clarify, I meant that the output the user was seeing was as expected.

cboggs:
I'm also curious - if remote_read is not intended for standard Prometheus queries, what _is_ it intended for?

brian-brazil:
> I'm also curious - if remote_read is not intended for standard Prometheus queries, what is it intended for?

Transition from Prometheus 1.x, and horizontal sharding.

cboggs:
There's a good chance I'm missing some nuance here, but it seems that "horizontal sharding" is something that would require the ability to query shards in the same way one would query the local store, yea? If I have 3 "shards" in my DC, and a query that *should* return 12 series only returns 4, that seems broken to me.

brian-brazil:
> is something that would require the ability to query shards in the same way one would query the local store, yea?

The external label setup is such that this is not an issue.

> I'm also not sure why your concern about "taking down all your monitoring at once" doesn't apply equally to the sharding use case as it presumably would to the multi-region sharding use case.

It does, but it's more restricted. It's a tradeoff against using federation for the same.

PR #5088, opened 2019-01-10
sd: Fix stuck Azure service discovery

sylr:
Our prometheus instances had its Azure SD stuck.

After some analyses with @simonpasquier we think it's because nothing is sent to the channel and we keep waiting for something.

Here the go routine dump of the stuck instance : https://gist.github.com/sylr/8c1e7f78c160fad84b0aea10a16f5ec1

PR #5106, opened 2019-01-17
 discovery/azure: more robust handling of go routines

simonpasquier:
The existing code led to issues that are quite hard to detect in code reviews (see #5088). This change refactors the logic to use a `sync.WaitGroup` instead of assuming that all goroutines will return something to the receiving channel.

cc @sylr @brian-brazil


brian-brazil:
:+1:

cstyan:
:+1: nice

PR #5109, opened 2019-01-18
Add flag for size based retention

gouthamve:
@krasi-georgiev Could you commit the vendor update, it's taking ages upon ages for me

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -177,8 +177,8 @@ func main() {
 	a.Flag("storage.tsdb.retention.time", "How long to retain samples in storage.").
 		Default("15d").SetValue(&cfg.tsdb.RetentionDuration)
 
-	a.Flag("storage.tsdb.retention.size", "[EXPERIMENTAL] Maximum number of bytes that can be stored for blocks. This flag is experimental and can be changed in future releases.").
-		Default("0").Int64Var(&cfg.tsdb.MaxBytes)
+	a.Flag("storage.tsdb.retention.size", "[EXPERIMENTAL] Maximum number of bytes that can be stored for blocks. Units supported: KB, MB, GB, TB, PB. This flag is experimental and can be changed in future releases.").
+		Default("0").BytesVar(&cfg.tsdb.MaxBytes)
 
 	a.Flag("storage.tsdb.no-lockfile", "Do not create lockfile in data directory.").
 		Default("false").BoolVar(&cfg.tsdb.NoLockfile)
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -171,7 +171,10 @@ func main() {
 		"Size at which to split the tsdb WAL segment files (e.g. 100MB)").
 		Hidden().PlaceHolder("<bytes>").BytesVar(&cfg.tsdb.WALSegmentSize)
 
-	a.Flag("storage.tsdb.retention", "How long to retain samples in storage.").
+	a.Flag("storage.tsdb.retention", "[DEPRECATED] How long to retain samples in storage. This flag has been deprecated, use \"storage.tsdb.retention.time\" instead").
+		Default("15d").SetValue(&cfg.tsdb.Retention)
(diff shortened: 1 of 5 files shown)


krasi-georgiev:
decided to update tsdb in a separate PR to keep the git history separate and reverted the tsdb update commits

https://github.com/prometheus/prometheus/pull/5110

storage/tsdb/tsdb.go
```diff
@@ -195,6 +208,27 @@ func Open(path string, l log.Logger, r prometheus.Registerer, opts *Options) (*t
 	return db, nil
 }
 
+// ChooseRetention is some roundabout code to support both RetentionDuration and Retention (for different flags).
+// If Retention is 15d, then it means that the default value is set and the value of RetentionDuration is used.
+func ChooseRetention(oldFlagDuration, newFlagDuration model.Duration) model.Duration {
+	defaultDuration, err := model.ParseDuration("15d")
```
codesome: I see `"15d"` being used in 4 places now (1 more in test). Should be have a single variable for this?

brancz:
--storage.tsdb.retention can be deprecated but must continue to work. The new flag can have precedence, but the old one must continue to work until 3.0.

brancz:
Agreed with Simon.

gouthamve:
Yup, @brancz that is what is happening. And yup, I moved the logic into main.go. It's better, thanks! @simonpasquier

storage/tsdb/tsdb.go
```diff
@@ -33,6 +33,8 @@ import (
 // ErrNotReady is returned if the underlying storage is not ready yet.
 var ErrNotReady = errors.New("TSDB not ready")
 
+const DefaultRetentionDuration = "15d"
```
simonpasquier: not needed anymore

cmd/prometheus/main.go
```diff
@@ -244,15 +257,25 @@ func main() {
...
-		cfg.tsdb.MaxBlockDuration = cfg.tsdb.Retention / 10
+		cfg.tsdb.MaxBlockDuration = cfg.tsdb.RetentionDuration / 10
 	}
 
 	promql.LookbackDelta = time.Duration(cfg.lookbackDelta)
 	promql.SetDefaultEvaluationInterval(time.Duration(config.DefaultGlobalConfig.EvaluationInterval))
 
 	logger := promlog.New(&cfg.promlogConfig)
 
+	defaultDuration, err := model.ParseDuration(tsdb.DefaultRetentionDuration)
```
simonpasquier: As we do the same in `chooseRetention()`, I'd use a global `mode.Duration` variable and move this to the `init()` function.

cmd/prometheus/main.go
```diff
@@ -757,3 +780,24 @@ func sendAlerts(s sender, externalURL string) rules.NotifyFunc {
...
+		panic(err)
+	}
+
+	retention := oldFlagDuration
+	if retention == defaultDuration {
+		retention = newFlagDuration
+	}
+
+	// Further if both the flags are set, then RetentionDuration takes precedence.
+	if newFlagDuration != defaultDuration {
```
brancz: I'm not sure this yields the expected behavior, when we say the new flag takes precedence. If the new flag explicitly specifies 15 days and the old flag specifies 30 days, then the old one will be chosen.

gouthamve: Hmm, fair. Not sure how to deal with the case though :/ Detecting if the flag has been set or if it's just the default value is next to impossible.

brancz: I think just setting the flag description to explain this rectifies the situation.

"If this flag is set to anything other than the default, it takes precedence."

gouthamve: Done

brancz:
lgtm :+1:

gouthamve:
@simonpasquier @brancz Feedback addressed, PTAL.

Thanks for all the rounds of reviews, sorry for the trouble!

PR #5110, opened 2019-01-18
Udpate tsdb to 0.4

krasi-georgiev:
cc @gouthamve 	
I decided to do it in a separate PR to keep the git history separate.

### DIFF
--- a/vendor/github.com/prometheus/tsdb/db.go
+++ b/vendor/github.com/prometheus/tsdb/db.go
@@ -44,7 +44,6 @@ import (
 // DefaultOptions used for the DB. They are sane for setups using
 // millisecond precision timestamps.
 var DefaultOptions = &Options{
-	WALFlushInterval:  5 * time.Second,
 	WALSegmentSize:    wal.DefaultSegmentSize,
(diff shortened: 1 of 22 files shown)


vendor/github.com/prometheus/tsdb/db.go
```diff
@@ -174,18 +178,14 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
 		Help: "Total number of skipped compactions due to disabled auto compaction.",
 	})
-	m.cutoffs = prometheus.NewCounter(prometheus.CounterOpts{
-		Name: "prometheus_tsdb_retention_cutoffs_total",
-		Help: "Number of times the database cut off block data from disk.",
-	})
-	m.cutoffsFailed = prometheus.NewCounter(prometheus.CounterOpts{
-		Name: "prometheus_tsdb_retention_cutoffs_failures_total",
-		Help: "Number of times the database failed to cut off block data from disk.",
-	})
```
skozlov: This is a breaking change, but it seems it is not mentioned in the changelog.

PR #5112, opened 2019-01-18
Protect retention from overflowing

### LINKED ISSUE #5111: Overflow in storage.tsdb.max-block-duration
## Bug Report
**What did you do?**

Set `--storage.tsdb.retention=500y`

`storage.tsdb.max-block-duration` in /flags would be "50y"

`storage.tsdb.max-block-duration` returns `-267874407370ms`

**Environment**

* System information:

`Linux 4.18.0-13-generic x86_64`

* Prometheus version:

```
prometheus, version 2.6.0 (branch: HEAD, revision: dbd1d58c894775c0788470944b818cc724f550fb)
  build user:       root@bf5760470f13
  build date:       20181217-15:14:46
  go version:       go1.11.3
```

gouthamve:
Also sanitise the max block duration to max a week.

Fixes #5111

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -19,6 +19,7 @@ import (
 	"crypto/md5"
 	"encoding/json"
 	"fmt"
+	"math"
 	"net"
 	"net/http"
 	_ "net/http/pprof" // Comment this line to disable pprof endpoint.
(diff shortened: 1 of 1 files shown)


cmd/prometheus/main.go
```diff
@@ -244,8 +245,22 @@ func main() {
...
 	if cfg.tsdb.MaxBlockDuration == 0 {
 		cfg.tsdb.MaxBlockDuration = cfg.tsdb.Retention / 10
+
+		weekLong, err := model.ParseDuration("7d")
+		if err != nil {
+			panic(err)
+		}
+
+		if cfg.tsdb.MaxBlockDuration > weekLong {
+			cfg.tsdb.MaxBlockDuration = weekLong
```
brian-brazil: How was this value determined? I'd be looking at something in the one month to one year range.

brian-brazil: That was before the memory improvements I made in the last release. Having too small blocks will actually increase memory usage.

SuperQ: I still see 20% bump in memory compacting 7d blocks on a typical server with 100k samples/sec.

What is the the overhead of having more blocks?

brian-brazil: It's duplicating the index memory structures, plus the various other bits and pieces. It'll also slow down and require more memory for queries across blocks.

gouthamve:
@brian-brazil PTAL, made it a month.

cmd/prometheus/main.go
SuperQ: Right, but how much is that in terms of bytes or percent of server? I see a nearly linear percent burst based on the size of the compaction, so if we went with 28d blocks on my server, I would see an 80% burst in memory use on compaction.

gouthamve:
Nope, those blocks that are already large would be ignored. Should break anything.

cmd/prometheus/main.go
brian-brazil: With the current code, I'd expect it to always be around 10% as it'd be creating an index for up to 10% of data.

gouthamve:
Should not* :P edited the original comment also.

simonpasquier:
:+1: 

I hope it isn't a Freudian slip :smirk:

PR #3369, opened 2017-10-30
Add Eureka Service Discovery

kangwoo:
Added Eureka Service Discovery. 
The Eureka client used a 'https://github.com/kangwoo/go-eureka-client' library

PR #5113, opened 2019-01-18
add service discovery support for eureka

rmalchow:
i added basic support for service discovery through eureka. this adds "eureka_sd_configs" to the prometheus config. works like a charm with (for example) springboot applications that expose both the prometheus endpoint and are eureka clients. 

i would really appreciate it if this could go in, however, i am not especially proficient in go ... so, if there are any changes you need me to make, let me know. also, i guess it would be nice to add eureka cluster support. 

```
scrape_configs:
  - job_name: 'services_from_eureka'
    metrics_path: '/actuator/prometheus'
    eureka_sd_configs:
    - eureka: http://127.0.0.1:8761/eureka/apps
```

### DIFF
--- a/discovery/eureka/eureka.go
+++ b/discovery/eureka/eureka.go
@@ -15,11 +15,11 @@ package eureka
 
 import (
 	"context"
+	"bytes"
 	"fmt"
 	"time"
 	"net"
 	"net/http"
-	"bytes"
 	"encoding/json"
 
 	"github.com/go-kit/kit/log"
@@ -195,8 +195,6 @@ func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
 		return nil, fmt.Errorf("could not parse JSON response from eureka (%s)", err)
 	}
 
-	bytes, _ := json.Marshal(er)
-	
 	tg = &targetgroup.Group{
 		Source: d.eureka,
 	}
--- a/discovery/eureka/eureka.go
+++ b/discovery/eureka/eureka.go
@@ -1,0 +1,222 @@
+// Copyright 2015 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


simonpasquier:
Thanks for suggesting this change. Unfortunately, we are currently not accepting new integrations. For Eureka, you can have a look at https://github.com/twinformatics/eureka-consul-adapter which is a service that exposes an API compatible with `consul_sd_configs` (not tried it myself though).

rmalchow:
hi. 

we've tried the consul adapter, and it is unfortunately not passing on as much information as we would like to see. another way i was thinking of was to beef up what the filewatcher can do, to the point where it can load stuff (i.e. proper prometheus configs) through HTTP. ultimately, given the (not so short) list of existing sd integrations, that eureka would be in good company ... and given the brevity of my addition (it's just 220 lines ... ). another thought i had was to run this in parallel, and then let the file watcher do the work. this, however, doesn't really play well with containerized environments. 

could you maybe discuss this further? as i said, i am perfectly fine with doing additional work if you need. 

.rm

roidelapluie:
Thank you for your work. Currently I am considering #3369 as it is using an Eureka library, which is less error-prone regarding unmarshalling etc.

discovery/eureka/eureka.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+	return &Discovery{
+		interval: time.Duration(conf.RefreshInterval),
+		eureka:   conf.Eureka,
+		logger:   logger,
+	}
+}
+
+// Run implements the Discoverer interface.
+func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
```
roidelapluie, line 131: This is now obsolete in prometheus. You should use the refresh package.

discovery/eureka/eureka.go
```diff
@@ -0,0 +1,220 @@
...
+func (d *Discovery) refresh() (tg *targetgroup.Group, err error) {
+	t0 := time.Now()
+	defer func() {
+		eurekaSDRefreshDuration.Observe(time.Since(t0).Seconds())
+		if err != nil {
+			eurekaSDRefreshFailuresCount.Inc()
+		}
+	}()
+
+	client := &http.Client{}
```
roidelapluie, line 176: Generaly we create a client from the HTTPClientConfig

roidelapluie:
mmh I intended to review #3369 :)

Sorry!

I will close this PR now and work on the other one.

PR #3205, opened 2017-09-21
Make all rule links link to the "Console" tab rather than "Graph"

beorn7:
Clicking on a rule, either the name or the expression, opens the rule
result (or the corresponding expression, repsectively) in the
expression browser. This should by default happen in the console tab,
as, more often than not, displaying it in the graph tab runs into a
timeout.

@grobie we talked about this. I wanted to fix it for ages, and finally it should go into the bug fix release.
IIRC @brian-brazil said he thought this has been fixed long ago. Thus, it's definitely the expected behavior, and what we have now is a bug.

PR #5115, opened 2019-01-18
Default link for expression in alerts

lilred:
I have a Slack integration that uses `{{ .GeneratorURL }}` to link to a graph. I expected the relevant URL to lend itself to appending parameters, for example adding `&g0.tab=0` to link to the "Graph" tab in the Prometheus web interface. However, the generated URL has `&g0.tab=1` at the end of it. It's not possible to override that by adding another parameter after.

This is my first time dealing with Go code. I can't get this to build locally, unfortunately I get the following output:

```
$ make
>> checking code style
>> running staticcheck
chmod +x /Users/me/go/bin/staticcheck
GO111MODULE=on go list -e -compiled -test=true -export=false -deps=true -find=false -tags= -- ./... > /dev/null
go: downloading github.com/cockroachdb/cockroach v0.0.0-20170608034007-84bc9597164f
go: downloading github.com/hashicorp/serf v0.0.0-20161007004122-1d4fa605f6ff
go: downloading github.com/hashicorp/go-rootcerts v0.0.0-20160503143440-6bb64b370b90
go: verifying github.com/hashicorp/go-rootcerts@v0.0.0-20160503143440-6bb64b370b90: checksum mismatch
        downloaded: h1:9HVkPxOpo+yO93Ah4yrO67d/qh0fbLLWbKqhYjyHq9A=
        go.sum:     h1:VBj0QYQ0u2MCJzBfeYXGexnAl17GsH1yidnoxCqqD9E=
make: *** [common-staticcheck] Error 1
```

Signed-off-by: Carl Patenaude-Poulin <carlpaten@gmail.com>


brian-brazil:
Thanks for your PR, this is on purpose per #3205

lilred:
Hi @brian-brazil,

I think this PR and the one you linked are entirely compatible! This change here preserves the (intentional) default behaviour, the only difference is that it becomes possible to override it. I guess I could see "don't make the default overrideable" as a legitimate design decision, but I'd hope that's not what you have in mind because I would be very happy if I could override the default

brian-brazil:
Ah, I see what you're doing now. Try reReplaceAll on the Alertmanager side.

lilred:
Thanks for the workaround, that's awesome. I still think this PR may have merit, but I'll let you decide. (I'll fix the broken tests tomorrow.)

brian-brazil:
This is technically a breaking change (though one we're allowed make), so I'd prefer not to change it if we don't have to.

codesome:
We discussed this during our bug scrub and decided to close this for this reason: https://github.com/prometheus/prometheus/pull/5115#issuecomment-455709170

PR #5116, opened 2019-01-18
expose kubernetes node metrics on discovered kubernetes pods

pdbogen:
We use kubernetes discovery with `role: pod` as our primary way to discover kubernetes services. Unfortunately, pod labels / annotations do not include properties of the node like instance type or region.

This change adds node labels to pods, so that discovered pods include information about the node that runs them.

### DIFF
--- a/discovery/kubernetes/kubernetes.go
+++ b/discovery/kubernetes/kubernetes.go
@@ -93,6 +93,7 @@ type SDConfig struct {
 	BearerTokenFile    string                 `yaml:"bearer_token_file,omitempty"`
 	TLSConfig          config_util.TLSConfig  `yaml:"tls_config,omitempty"`
 	NamespaceDiscovery NamespaceDiscovery     `yaml:"namespaces,omitempty"`
+	PodNodeLabels      bool                   `yaml:"pod_node_labels,omitempty"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
@@ -116,6 +117,9 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		(c.BasicAuth != nil || c.BearerToken != "" || c.BearerTokenFile != "" ||
 			c.TLSConfig.CAFile != "" || c.TLSConfig.CertFile != "" || c.TLSConfig.KeyFile != "") {
 		return fmt.Errorf("to use custom authentication please provide the 'api_server' URL explicitly")
+	}
+	if c.Role != "pod" && c.PodNodeLabels {
+		return fmt.Errorf("pod_node_labels can only be used with role=pod")
 	}
 	return nil
 }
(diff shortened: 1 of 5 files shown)


simonpasquier:
Hmm, IIUC Prometheus needs additional permissions to query the node's information. I don't think it is practical...

cc @brancz

pdbogen:
Hmm, yeah, kubernetes rbac is definitely capable of restricting which resources are readable. Is there a recommended configuration that needs to be updated?

There's certainly nothing fundamentally wrong with the change; a build compiled from this worked fine with an unmodified cluster, unmodified prom config.

Maybe it should accept a failure to read nodes and log it, but otherwise ignore it? So existing use cases that do narrowly restrict Prometheus would still work.

Or, the functionality could be enabled by a config flag?

brancz:
Judging by the managed Kubernetes offerings I’ve been in touch with, list/watch on nodes is a very high permission and is not given lightly, so if at all, this functionality must be opt in. Beyond that I feel your use case is better covered by labeling your targets with namespace and pod and join the rest of the metrics you want onto it using the metrics exposed by kube-state-metrics.

pdbogen:
Because our targets are pods, we found it non-trivial to dynamically label them with even the pod name.

However, I agree that it'd be really bad UX to break existing use cases, so I've added a configuration option to the kube sd config to enable this functionality.

brancz:
I'm not too sure about this one. My feeling is that this is a case of "to make the easy things easy, and the hard things possible", and extending service discovery with this feature introduces inconsistencies such as what does `podNodeLabels` do when enabled for `node`, `ingress` or `service` roles?

pdbogen:
I'm not quite sure I understand the second part; but knowing about the runtime environment of a workload is very important for monitoring, since otherwise we might totally miss issues that are correlated to availability zone, for example.

I think that _should_ be easy, not merely possible; and needing another scraped third party cluster component plus relatively complex join queries everywhere is not easy.

On the second point, the flag is only valid for pods, and is rejected otherwise:

```go
 		if c.Role != "pod" && c.PodNodeLabels {
				return fmt.Errorf("pod_node_labels can only be used with role=pod")
			}
```

brancz:
I don't know, I don't think joining is hard, and node metadata may change over time, while the target is actually static. I think the target should only have target identifying labels, everything is additional metadata that should be added at query time.

pdbogen:
While I understand the reasoning behind your opinion, it does not seem consistent with the Prometheus documentation on labels:

> Use labels to differentiate the characteristics of the thing that is being measured:

Identifying properties of the location where a workload is being run seem like important differentiating characteristics.

Furthermore, the Node a Pod is running on never changes, so a given named pod will always run on the same node. The node's AZ never changes, so a given node will always have the same AZ. Which is to say, many of the node labels are in fact constant. Not all, of course, but that's true also of pods, or prometheus targets in general.

In fact, for the use cases I have in mind, this won't even result in an increase in cardinality.

codesome:
@brancz can this be closed?

PR #5117, opened 2019-01-19
corrected regex string check for anyorigin(*)

geekodour:
forgot to change this with #5011

currently it will return with whatever `Origin` header the request has because the default regex is `.*` and will match any Origin anyway.

but default behavior of returning `*` for `Access-Control-Allow-Origin` is changed. 

this PR fixes that.


brian-brazil:
@gouthamve Bugfix for 2.7 - how do you want to handle this?

brian-brazil:
It looks fine to me, this is more about our regex handling.

simonpasquier:
LGTM

PR #5127, opened 2019-01-23
add debounce support in prometheus

garrettlish:
You can specify either hold duration or debounce in alerting rule for field. Here `for: 3/4` represents that alerting if 3 matches in last 4 round evaluations.
```
- alert: test1
  expr: up > 0
  for: 3/4
- alert: test2
  expr: up > 0
  for: 3m
```

### DIFF
--- a/pkg/rulefmt/rulefmt.go
+++ b/pkg/rulefmt/rulefmt.go
@@ -95,7 +95,7 @@ type Rule struct {
 	Record      string            `yaml:"record,omitempty"`
 	Alert       string            `yaml:"alert,omitempty"`
 	Expr        string            `yaml:"expr"`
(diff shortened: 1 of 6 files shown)


codesome:
While I leave the decision about this feature to other members, I _feel_ having this config under `for` would be confusing.

beorn7:
How about the following:
```Yaml
- alert: test1
  expr: avg_over_time(up[4m]) < 0.75
```

Especially with the new sub-query feature, I think you can implement this kind of alerts out of the box.

brian-brazil:
I agree with @beorn7, this is already possible via PromQL.

garrettlish:
agree, sub-query looks good for debounce

PR #5130, opened 2019-01-23
Use the context associated with the inner evaluation span

bboreham:
So any spans or logs attached in code beneath this point will appear in the right place.


brian-brazil:
Thanks!

PR #5131, opened 2019-01-23
 Check for cancellation on every step of a range evaluation

bboreham:
The background to this is I have an end-user sending a query that runs for 17 seconds and cancelling it after 4 seconds.

I know, I should find a way to make them stop, but I couldn't see (a) why this isn't checked already or (b) where this has been discussed.

In case it's a fear of an expensive `select` call I simplified that, although I can't see either change show up in benchmarks.

Justification for the `contextDone()` change from the `Context` docs:

	// If Done is not yet closed, Err returns nil.
	// If Done is closed, Err returns a non-nil error

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -791,6 +791,9 @@ func (ev *evaluator) rangeEval(f func([]Value, *EvalNodeHelper) Vector, exprs ..
 	seriess := make(map[uint64]Series, biggestLen) // Output series by series hash.
 	tempNumSamples := ev.currentSamples
 	for ts := ev.startTimestamp; ts <= ev.endTimestamp; ts += ev.interval {
+		if err := contextDone(ev.ctx, "expression evaluation"); err != nil {
+			ev.error(err)
+		}
 		// Reset number of samples in memory after each timestamp.
 		ev.currentSamples = tempNumSamples
 		// Gather input vectors for this timestamp.
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -181,12 +181,10 @@ func (q *query) Exec(ctx context.Context) *Result {
 
 // contextDone returns an error if the context was canceled or timed out.
 func contextDone(ctx context.Context, env string) error {
-	select {
-	case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


brian-brazil:
I think this will only cost a few microseconds. Can you run the promql benchmarks to verify?

bboreham:
I don't stable results from benchmark timings.  Initially I thought this was because I was running them on my laptop, but switching to a cloud server doing nothing else I still get variations between +12% and -25% between runs of the same code.

Do you have any tips?

brian-brazil:
That's a bit suprising, they're fairly constant for me. Try increasing `-benchtime`.

bboreham:
I tried increasing `-benchtime`, and apart from making the whole batch take a really long time it behaved the same.

Next idea was to increase `-count` - this just made it really clear what I'm talking about when two successive runs of the same test have radically different stats:

```
BenchmarkRangeQuery/expr=sum_without_(l)(h_hundred),steps=1000-4           	       5	 335913807 ns/op	16007358 B/op	   94085 allocs/op
BenchmarkRangeQuery/expr=sum_without_(l)(h_hundred),steps=1000-4           	       3	 335528891 ns/op	14789061 B/op	   94008 allocs/op
...
BenchmarkRangeQuery/expr=sum_by_(le)(h_hundred),steps=1000-4               	       3	 344485515 ns/op	11442114 B/op	   71977 allocs/op
BenchmarkRangeQuery/expr=sum_by_(le)(h_hundred),steps=1000-4               	       5	 314056918 ns/op	 9007076 B/op	   71832 allocs/op
```

I've never seen the bytes and allocs numbers change much between runs of a Go benchmark, before now.

I do see [other people get similar outcomes](https://github.com/prometheus/tsdb/pull/499#issuecomment-455552204) - byte number differences range from -23% to +4% and time differences range from -20% to +2.5% there, on a change that is unlikely to cause any extra allocation.

Can you share any more details about your method to get stable results?

brian-brazil:
That's quite odd, I'd only ever expect time to vary.

PR #5132, opened 2019-01-24
allow string arguments for the function which applied to matrix

garrettlish:
Assume we have one function which accept matrix as 1st arguments and url as 2nd arguments, the function will make PRC call to the url by passing the matrix. To support the custom function, we need to make the change to non-matrix argument handler in eval, otherwise we will get `unhandled expression of type` error.

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -28,7 +28,7 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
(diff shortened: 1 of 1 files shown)


brian-brazil:
This information is already passed to functions, so you can do the cast there which is what e.g. label_replace does.

> the function will make PRC call to the url by passing the matrix.

That's not how it works, it'll be called per range per series so you could easily be looking at millions of function calls. Remote read is the general way to handle this.

PR #5139, opened 2019-01-27
Limit number of merics in prometheus UI

### LINKED ISSUE #2119: Limit number of metrics displayed in prometheus UI

### LINKED ISSUE #3321: Autocomplete on /graph interface slow with large numbers of metricnames

vn-ki:
Closes #2119 
Closes #3321

Displays only first 10000 series in the UI.
If there are more than 10000 series, show a warning.

Display only the first 100 metrics in the typeahead. I chose 100 because it did not make sense to scroll beyond that in a suggestion box and can be changed in case more is required.

Make the typeahead scrollable for easy navigation of suggestions. 

Ping: @simonpasquier @codesome 

Signed-off-by: Vishnunarayan K I <appukuttancr@gmail.com>


web/ui/static/js/graph/index.js
```diff
@@ -1143,6 +1145,25 @@ function escapeHTML(string) {
   });
 }
 
+function limitMetrics(result) {
+  var MAX_METRICS_NUM = 10;
```
brian-brazil: Is this the right value?

I'd suggest going with 10k, which is still a reasonable number of metrics to have.

vn-ki: @brian-brazil Done.

codesome:
Gave a quick run. The warning doesn't go away when I query something later which fetches fewer metrics.

vn-ki:
@codesome Fixed.

web/ui/static/js/graph/index.js
```diff
@@ -1143,6 +1145,26 @@ function escapeHTML(string) {
...
 
+function limitMetrics(result) {
+  var MAX_METRICS_NUM = 10000;
+  var message = "<strong>Warning!</strong> Fetched " +
+    result.length + " metrics, but displaying only first " +
+    MAX_METRICS_NUM + " ones.";
+
+  if ($("#max-metrics-warning")[0]) {
+    // If message already exists
+    $("#max-metrics-warning").remove();
```
codesome: From a UX perspective, I think this should be done when you submit a query.

vn-ki: Done

vn-ki: Please don't merge it yet. I'm refactoring it a bit.

vn-ki: Done

web/ui/static/js/graph/index.js
```diff
@@ -887,6 +903,20 @@ Prometheus.Graph.prototype.formatKMBT = function(y) {
   }
 }
 
+Prometheus.Graph.prototype.limitMetrics = function (result) {
+  var self = this;
+  var MAX_METRICS_NUM = 10000;
+  var message = "<strong>Warning!</strong> Fetched " +
+    result.length + " metrics, but displaying only first " +
+    MAX_METRICS_NUM + " ones.";
```
codesome:
```suggestion
    MAX_METRICS_NUM + ".";
```

web/ui/static/js/graph/index.js
```diff
@@ -803,6 +817,7 @@ Prometheus.Graph.prototype.handleConsoleResponse = function(data, textStatus) {
       tBody.append("<tr><td colspan='2'><i>no data</i></td></tr>");
       return;
     }
+    data.result = self.limitMetrics(data.result);
```
juliusv: Could we change the wording from "metrics" to "series" here, since "metrics" often in most Prometheus contexts refers to a metric name or all series belonging to a given metric name. Same below in the error message.

web/ui/static/js/graph/index.js
```diff
@@ -286,7 +287,7 @@ Prometheus.Graph.prototype.initTypeahead = function(self) {
   self.expr.typeahead({
     autoSelect: false,
     source: source,
-    items: "all",
+    items: 100,
```
juliusv: Can we make this limit higher, like 1000? When exploring metrics, I often just type e.g. `node_` in the expression input to see what exists with that prefix. But on [http://demo.robustperception.io:9090/graph?g0.range_input=1h&g0.expr=count(count%20by(__name__)%20(%7B__name__%3D~%22node_.*%22%7D))&g0.tab=1](the RobustPerception demo server), this already returns 169 different metric names. It would be possible to use the dropdown selector for some of these use cases, but that is more cumbersome.

vn-ki:
Done @codesome @juliusv

web/ui/static/js/graph/index.js
```diff
@@ -887,6 +903,20 @@ Prometheus.Graph.prototype.formatKMBT = function(y) {
   }
 }
 
+Prometheus.Graph.prototype.limitSeries = function (result) {
+  var self = this;
+  var MAX_METRICS_NUM = 10000;
```
juliusv: This is still "metrics" rather than "series".

web/ui/static/js/graph/index.js
```diff
@@ -887,6 +903,20 @@ Prometheus.Graph.prototype.formatKMBT = function(y) {
   }
 }
 
+Prometheus.Graph.prototype.limitSeries = function (result) {
```
juliusv: Mini style nit: `function (result)` -> `function(result)`

vn-ki: @juliusv Done

juliusv:
:+1: thanks!

PR #5140, opened 2019-01-28
tests: Remove the prepended watch reactor from the fake k8s client

tariq1890:
This PR improves the test suite by removing the unnecessary watch reactor. The latest client-go supports the adding of events automatically CRUD events, so the reactor should not be required anymore. 

This change should be safe (hopefully) as only tests cases are modified. Hurray to new client-go libs ;)


tariq1890:
FYI @simonpasquier @brancz

PR #5144, opened 2019-01-28
Close idle scrape connections after reload and shutdown

mknapphrt:
When doing a reload and when shutting down tcp connections were left open until they timed out from the `IdleConnTimeout` field in Transport. Which means that they persist for 5 minutes when they will never be used again. In the case of a reload, we were seeing a doubling of tcp connections for a few minutes as a new client was setup on reload and both the new and the old client had established connections.

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -222,6 +222,7 @@ func (sp *scrapePool) stop() {
 		delete(sp.activeTargets, fp)
 	}
 	wg.Wait()
+	sp.client.Transport.(*http.Transport).CloseIdleConnections()
 }
 
 // reload the scrape pool with the given scrape configuration. The target state is preserved
@@ -232,6 +233,8 @@ func (sp *scrapePool) reload(cfg *config.ScrapeConfig) {
 
 	sp.mtx.Lock()
 	defer sp.mtx.Unlock()
+
+	oldClient := sp.client
 
 	client, err := config_util.NewClientFromConfig(cfg.HTTPClientConfig, cfg.JobName)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


scrape/scrape.go
```diff
@@ -222,6 +222,7 @@ func (sp *scrapePool) stop() {
 		delete(sp.activeTargets, fp)
 	}
 	wg.Wait()
+	sp.client.Transport.(*http.Transport).CloseIdleConnections()
```
brian-brazil: Digging through the types feels wrong here. Could this be more of a direct method?

mknapphrt: It could be more direct, but I would need to propagate the method back up from `NewRoundTripperFromConfig` and then `NewClientFromConfig`. It's not a huge change, it would just require changing the signature of those methods which I was trying to avoid.

brian-brazil: There's a few ways that it could be done, but it should be done. This code isn't safe for example if any of the auth methods are used, as they wrap the transport.

mknapphrt: I could wrap it in a type check for safety. I don't really want to modify the `NewRoundTripperFromConfig` and `NewClientFromConfig` methods if I can avoid it because it would require changes in prometheus/common. What other ways are you thinking of?

mknapphrt: Ok, I made a PR https://github.com/prometheus/common/pull/169, what's the proper way to update go vendor? `govendor fetch ...`?

simonpasquier: We're using Go modules now. See https://github.com/prometheus/prometheus/blob/master/CONTRIBUTING.md#dependency-management

brian-brazil:
Go 1.12 has a Client.CloseIdleConnections now, so we can simplify greatly once that's out.

brian-brazil:
1.12 is out.

simonpasquier:
That's been addressed by #5344 (https://github.com/prometheus/prometheus/pull/5344/files#diff-6a976999031a4966ff9d1a2734f15e5d)

PR #5151, opened 2019-01-29
update Consul vendor code so that catalog.ServiceMultipleTags can be

### LINKED ISSUE #4786: Several tags for filtering targets in consul_sd_config
## Proposal
According to documentation,
```
# An optional tag used to filter nodes for a given service.
[ tag: <string> ]
```
it is not possible to specify several tags to filter targets in consul_sd_config. One can use relabel_configs with action keep or drop but anyway it causes metrics to be pulled.

According to consul documentation it is possible to specify multiple tags in request:
```
tag (string: "") - Specifies the tag to filter on. This is specified as part of the URL as a query parameter. Can be used multiple times for additional filtering, returning only the results that include all of the tag values provided.
```

cstyan:
Fixes #4786 

cc: @tariq1890 you've touched this code recently also @simonpasquier 


tariq1890:
Hi @cstyan thanks for letting me know about this PR. I don't remember working on consul SD and If any of my changes caused a bug, I apologize. Could you let me know where exactly I made changes?

In any case, always happy to help review ^__^

cstyan:
:man_facepalming: my bad @tariq1890, wrong person 

cc: @iksaif

discovery/consul/consul.go
```diff
@@ -113,9 +113,8 @@ type SDConfig struct {
 	// The list of services for which targets are discovered.
 	// Defaults to all services if empty.
 	Services []string `yaml:"services,omitempty"`
-	// An optional tag used to filter instances inside a service. A single tag is supported
-	// here to match the Consul API.
-	ServiceTag string `yaml:"tag,omitempty"`
+	// An optional tag used to filter instances inside a service.
```
brian-brazil: Is it an and or an or?

cstyan: Based on #4786 it's an `AND` on all the tags from here: https://www.consul.io/api/catalog.html#tag

I'll update the comment.

iksaif:
Good to know that consul now supports this :). Do we really want to break compatibility with the old config here?

simonpasquier:
I agree with @iksaif that although this SD mechanism isn't covered by the [stability guarantees](https://prometheus.io/docs/prometheus/latest/stability/), it would be nicer to not break users immediately. Maybe support both `tag: <string> ` and `tags: [ <string>, ... ]` (`tags` taking precedence), advertise `tag` as deprecated and remove it after a couple of releases.

brian-brazil:
I'd prefer to avoid cruft building up in the code. This is a fairly new feature, and should only be in use by advanced users.

cstyan:
@iksaif @simonpasquier good point

@brian-brazil it's unclear whether you're saying we shouldn't support searching for services with multiple tags, or just to not include `tags` and `tag` in the config

brian-brazil:
I'm saying don't include both.

cstyan:
Any opinions on how to proceed? I don't see a way to not break peoples old config for this change. We can't marshal a single string into an array of strings with go-yaml.

brian-brazil:
I'd suggest breaking it, as new features come along in SDs I'm not sure we should be spending too much effort on avoiding breakage. Particularly for more advanced features.

cstyan:
Fine by me. It's a pretty minor change required in users config files anyways.

cstyan:
Do we want to move ahead with these changes as is?

simonpasquier:
:+1: as @brian-brazil is ok with the breaking change.

cstyan:
okay, seems no one is opposed

cstyan:
@brian-brazil bump

discovery/consul/consul.go
```diff
@@ -238,13 +237,15 @@ func (d *Discovery) shouldWatchFromName(name string) bool {
...
+	if len(d.watchedTags) == 0 {
 		return true
 	}
 
 	for _, tag := range tags {
-		if d.watchedTag == tag {
-			return true
+		for _, wtag := range d.watchedTags {
+			if wtag == tag {
+				return true
```
brian-brazil: This is only checking for the presence of one tag

cstyan:
@brian-brazil @iksaif updated, a service must now contain all tags we're watching to be added to watched services.

cc: @ashepelev

discovery/consul/consul.go
```diff
@@ -238,16 +237,20 @@ func (d *Discovery) shouldWatchFromName(name string) bool {
 // *all* services. Details in https://github.com/prometheus/prometheus/pull/3814
 func (d *Discovery) shouldWatchFromTags(tags []string) bool {
 	// If there's no fixed set of watched tags, we watch everything.
-	if d.watchedTag == "" {
+	if len(d.watchedTags) == 0 {
 		return true
 	}
 
+	a := struct{}{}
+	found := make(map[string]struct{})
```
brian-brazil: This seems a bit inefficient, and won't work if a tag is repeated for some reason. What I'd do is have the inner loop continue the outer if it's found, and return false if it's not.

discovery/consul/consul.go
```diff
@@ -444,10 +447,10 @@ func (d *Discovery) watchService(ctx context.Context, ch chan<- []*targetgroup.G
 
 // Get updates for a service.
 func (srv *consulService) watch(ctx context.Context, ch chan<- []*targetgroup.Group, catalog *consul.Catalog, lastIndex *uint64) error {
-	level.Debug(srv.logger).Log("msg", "Watching service", "service", srv.name, "tag", srv.tag)
+	level.Debug(srv.logger).Log("msg", "Watching service", "service", srv.name, "tag", srv.tags)
```
simonpasquier: s/"tag"/"tags"/

discovery/consul/consul_test.go
```diff
@@ -68,6 +68,50 @@ func TestConfiguredServiceWithTag(t *testing.T) {
 	}
 }
 
+func TestConfiguredServiceWithTags(t *testing.T) {
```
simonpasquier: Table-driven tests would be more readable here but this is my personal taste.

discovery/consul/consul.go
cstyan: Repeated tags is a weird case. I would assume that looking for the same tag multiple times is a configuration issue, and not something people would actually have on purpose.

But yes, the continue you're suggesting is better :+1:

discovery/consul/consul_test.go
cstyan: :+1: agreed, done

brian-brazil:
:+1: 

You've conflicts

cstyan:
@brian-brazil thanks, fixed. 

I really need to check if there's a way to have GH notify me when there's a merge conflict.

cstyan:
@brian-brazil PTAL :)

brian-brazil:
Thanks!

PR #5158, opened 2019-01-30
In histogram_quantile merge buckets with equivalent le values

brian-brazil:
This makes things generally more resilient, and will
help with OpenMetrics transitions (and inconsistencies).

### DIFF
--- a/promql/quantile.go
+++ b/promql/quantile.go
@@ -75,15 +75,17 @@ func bucketQuantile(q float64, buckets buckets) float64 {
 	if q > 1 {
 		return math.Inf(+1)
 	}
-	if len(buckets) < 2 {
-		return math.NaN()
-	}
 	sort.Sort(buckets)
 	if !math.IsInf(buckets[len(buckets)-1].upperBound, +1) {
 		return math.NaN()
 	}
 
+	buckets = coalesceBuckets(buckets)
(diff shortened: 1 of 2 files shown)


promql/quantile.go
```diff
@@ -83,6 +83,7 @@ func bucketQuantile(q float64, buckets buckets) float64 {
 		return math.NaN()
 	}
 
+	buckets = coalesceBuckets(buckets)
```
juliusv: This is done after the `len(buckets) > 2` check, but `coalesceBuckets()` may remove buckets, causing the number of buckets to be <2 again. It seems due to the extra `+Inf` check, this can only happen if the input bucket list is an edge case like this:

`["+inf", "+Inf"]`

Not sure if the following code then actually has problems with this, but adding a test case for this would be good.

brian-brazil: Good point, fixed.

juliusv:
:+1:

beorn7:
Academic note: This PR assumes that `sum` is the only reasonably aggregation to be done before feeding the histogram to `histogram_quantile` (as it sums up equivalent buckets). I think that's a reasonable assumption, but it raises the question why `histogram_quantile` isn't an aggregation operator itself, sparing users the effort to `sum` explicitly.

PR #5159, opened 2019-01-30
Fix multiple response.WriteHeader calls error in remote read adapter

qingwave:
1. **Error in remote read adapter**
Error in `/documentation/examples/remote_storage/remote_storage_adapter/main.go`. When [w.write](https://github.com/prometheus/prometheus/blob/a2ef8cf2f509b58d140acf9e37f1e906b28582cd/documentation/examples/remote_storage/remote_storage_adapter/main.go#L298) return err, log shows `http: multiple response.WriteHeader calls`. `w.write` and `http.Error(w, err.Error(), http.StatusInternalServerError)` will write header twice，triggers the error.

2. **Bug fix**
Replace the  `http.Error(w, err.Error(), http.StatusInternalServerError)` with log.

### DIFF
--- a/documentation/examples/remote_storage/remote_storage_adapter/main.go
+++ b/documentation/examples/remote_storage/remote_storage_adapter/main.go
@@ -296,8 +296,7 @@ func serve(logger log.Logger, addr string, writers []writer, readers []reader) e
 
 		compressed = snappy.Encode(nil, data)
 		if _, err := w.Write(compressed); err != nil {
-			level.Warn(logger).Log("msg", "Resonse write error", "storage", reader.Name(), "err", err)
-			return
+			level.Warn(logger).Log("msg", "Error writing response", "storage", reader.Name(), "err", err)
 		}
 	})
 
--- a/documentation/examples/remote_storage/remote_storage_adapter/main.go
+++ b/documentation/examples/remote_storage/remote_storage_adapter/main.go
@@ -296,7 +296,7 @@ func serve(logger log.Logger, addr string, writers []writer, readers []reader) e
 
 		compressed = snappy.Encode(nil, data)
 		if _, err := w.Write(compressed); err != nil {
-			http.Error(w, err.Error(), http.StatusInternalServerError)
+			level.Warn(logger).Log("msg", "Resonse write error", "storage", reader.Name(), "err", err)
 			return
 		}
 	})


documentation/examples/remote_storage/remote_storage_adapter/main.go
```diff
@@ -296,7 +296,7 @@ func serve(logger log.Logger, addr string, writers []writer, readers []reader) e
 
 		compressed = snappy.Encode(nil, data)
 		if _, err := w.Write(compressed); err != nil {
-			http.Error(w, err.Error(), http.StatusInternalServerError)
+			level.Warn(logger).Log("msg", "Resonse write error", "storage", reader.Name(), "err", err)
```
simonpasquier: It would be a better fix to move L294-295 after this block.

qingwave: The error is not isolated to L294-295.
 Function [w.Write](https://github.com/golang/go/blob/889aa5eb98ff2a5134f5b075525479da399e5c4f/src/net/http/server.go#L367) will WriteHeader and function [http.Error](https://github.com/golang/go/blob/889aa5eb98ff2a5134f5b075525479da399e5c4f/src/net/http/server.go#L2004) will WriteHeader too.
Function Write
```golang
func (cw *chunkWriter) Write(p []byte) (n int, err error) {
	if !cw.wroteHeader {
		cw.writeHeader(p)
	}
	...
}
```
Function Error
```golang
func Error(w ResponseWriter, error string, code int) {
	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
	w.Header().Set("X-Content-Type-Options", "nosniff")
	w.WriteHeader(code)
	fmt.Fprintln(w, error)
}
```

simonpasquier: My bad, I overlooked the code. 

```go
level.Warn(logger).Log("msg", "Error writing response", "storage", reader.Name(), "err", err)
```

And you can remove `return` at L300.

qingwave: Okay, I have removed the useless return as your  suggest.

simonpasquier: Thanks! Can you also change `"Resonse write error"` to `"Error writing response"` please?

qingwave: My fault! Changed.

PR #5181, opened 2019-02-04
scrape: Add global jitter for HA server

xjewer:
Covers issue in https://github.com/prometheus/prometheus/pull/4926#issuecomment-449039848
where the HA setup become a problem for targets unable to be scraped simultaneously.
The new jitter per server relies on a hostname and external labels which necessarily to be uniq.

As before, scrape offset will be calculated with regard the absolute time, so even
restart/reload doesn't change scrape time per scrape target + prometheus instance.

Jitter interval was chosen within the boundaries 0-59 seconds, which should be enough
to bring entropy for the scrapes across HA setup.

### DIFF
--- a/scrape/manager.go
+++ b/scrape/manager.go
@@ -14,8 +14,10 @@
 package scrape
 
 import (
+	"encoding"
 	"fmt"
 	"hash/fnv"
+	"net"
 	"os"
 	"reflect"
 	"sync"
@@ -57,7 +59,7 @@ type Manager struct {
 	append    Appendable
 	graceShut chan struct{}
 
-	jitterSeed    uint64     // Global jitterSeed seed is using to spread scrape workload across HA setup.
+	jitterSeed    uint64     // Global jitterSeed seed is used to spread scrape workload across HA setup.
 	mtxScrape     sync.Mutex // Guards the fields below.
 	scrapeConfigs map[string]*config.ScrapeConfig
 	scrapePools   map[string]*scrapePool
@@ -135,14 +137,14 @@ func (m *Manager) reload() {
 	wg.Wait()
 }
 
-// setJitterSeed calculates global jitterSeed per server relying on extra label set.
+// setJitterSeed calculates a global jitterSeed per server relying on extra label set.
 func (m *Manager) setJitterSeed(labels model.LabelSet) error {
 	h := fnv.New64a()
-	hostname, err := os.Hostname()
+	hostname, err := getFqdn()
 	if err != nil {
 		return err
 	}
-	if _, err := h.Write([]byte(hostname + labels.String())); err != nil {
+	if _, err := fmt.Fprintf(h, "%s%s", hostname, labels.String()); err != nil {
 		return err
 	}
 	m.jitterSeed = h.Sum64()
@@ -237,4 +239,46 @@ func (m *Manager) TargetsDropped() map[string][]*Target {
 		targets[tset] = sp.DroppedTargets()
 	}
 	return targets
+}
+
+// getFqdn returns a fqdn if it's possible, otherwise falls back a hostname.
+func getFqdn() (string, error) {
+	hostname, err := os.Hostname()
+	if err != nil {
+		return "", err
+	}
+
+	ips, err := net.LookupIP(hostname)
+	if err != nil {
+		return hostname, err
+	}
+
+	lookup := func(ipStr encoding.TextMarshaler) (string, error) {
+		ip, err := ipStr.MarshalText()
+		if err != nil {
+			return "", err
+		}
+		hosts, err := net.LookupAddr(string(ip))
+		if err != nil || len(hosts) == 0 {
+			return "", err
+		}
+		return hosts[0], nil
+	}
+
+	for _, addr := range ips {
+		if ip := addr.To4(); ip != nil {
+			if fqdn, err := lookup(ip); err == nil {
+				return fqdn, nil
+			}
+
+		}
+
+		if ip := addr.To16(); ip != nil {
+			if fqdn, err := lookup(ip); err == nil {
+				return fqdn, nil
+			}
+
+		}
+	}
+	return hostname, nil
 }
--- a/scrape/manager.go
+++ b/scrape/manager.go
@@ -15,6 +15,8 @@ package scrape
 
 import (
 	"fmt"
+	"hash/fnv"
+	"os"
 	"reflect"
 	"sync"
 	"time"
@@ -22,6 +24,7 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 
+	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/discovery/targetgroup"
 	"github.com/prometheus/prometheus/storage"
@@ -54,6 +57,7 @@ type Manager struct {
 	append    Appendable
 	graceShut chan struct{}
 
+	jitterSeed    uint64     // Global jitterSeed seed is using to spread scrape workload across HA setup.
 	mtxScrape     sync.Mutex // Guards the fields below.
 	scrapeConfigs map[string]*config.ScrapeConfig
 	scrapePools   map[string]*scrapePool
@@ -111,7 +115,7 @@ func (m *Manager) reload() {
 				level.Error(m.logger).Log("msg", "error reloading target set", "err", "invalid config id:"+setName)
 				continue
 			}
-			sp, err := newScrapePool(scrapeConfig, m.append, log.With(m.logger, "scrape_pool", setName))
+			sp, err := newScrapePool(scrapeConfig, m.append, m.jitterSeed, log.With(m.logger, "scrape_pool", setName))
 			if err != nil {
 				level.Error(m.logger).Log("msg", "error creating new scrape pool", "err", err, "scrape_pool", setName)
 				continue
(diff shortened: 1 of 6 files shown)


scrape/target.go
```diff
@@ -125,13 +125,15 @@ func (t *Target) hash() uint64 {
...
+// It includes the global server jitter for scrapes of HA setup to be spread across the time interval.
+func (t *Target) offset(interval, jitter time.Duration) time.Duration {
 	now := time.Now().UnixNano()
 
+	// base is a pinned to a absolute time, no matter how often offset is called.
 	var (
 		base   = int64(interval) - now%int64(interval)
 		offset = t.hash() % uint64(interval)
-		next   = base + int64(offset)
+		next   = base + int64(offset) + int64(jitter)
```
brian-brazil: This could delay things by up to a minute, which is undesirable. You should augment the above hash, rather than adding a new time period to worry about.

xjewer: it could be more than the interval itself, see [below](https://github.com/prometheus/prometheus/pull/5181/files#diff-71e6b8df5f2eeb6a786cb5cacaa4187dR139) :

```
	if next > int64(interval) {
		next -= int64(interval)
	}
```

so it won't add the more diff than offset from hash `t.hash() % uint64(interval)`, it just adds an entropy throughout the HA setup.

brian-brazil: If the interval is 1ms, then it can delay things by a minute.

xjewer:
>It's be good to also include hostname, as that'll also catch test machines using the same config.

could you elaborate on it? what machines?

scrape/target.go
xjewer: fair, will change, thanks for catch.

brian-brazil:
> could you elaborate on it? what machines?

Any other machine on which someone copies over a production config verbatim.

xjewer:
>someone copies over a production config

what about docker setup, for those who doesn't specify hostname it will move around the offset on re-runs. 🤔

brian-brazil:
That's an acceptable tradeoff.

xjewer:
updated with `hostname`, but smth went wrong with an import:

```
go: gopkg.in/vmihailenco/msgpack.v2@v2.9.1: unrecognized import path "gopkg.in/vmihailenco/msgpack.v2" (parse https://gopkg.in/vmihailenco/msgpack.v2?go-get=1: no go-import meta tags ())
```

could you restart the CI test build?

xjewer:
I applied this patch to our 2.6.1 HA setup and it seems to work as expected:

image

scrape/target.go
```diff
@@ -125,12 +125,14 @@ func (t *Target) hash() uint64 {
...
-func (t *Target) offset(interval time.Duration) time.Duration {
+// It includes the global server jitter for scrapes of HA setup to be spread across the time interval.
+func (t *Target) offset(interval, jitter time.Duration) time.Duration {
 	now := time.Now().UnixNano()
 
+	// base is a pinned to a absolute time, no matter how often offset is called.
 	var (
 		base   = int64(interval) - now%int64(interval)
-		offset = t.hash() % uint64(interval)
+		offset = (t.hash() + uint64(jitter)) % uint64(interval)
```
brian-brazil: Usually you'd use `^` to combine these

scrape/manager.go
```diff
@@ -130,6 +134,20 @@ func (m *Manager) reload() {
...
+func (m *Manager) setServerJitter(labels model.LabelSet) error {
+	h := fnv.New64a()
+	hostname, err := os.Hostname()
+	if err != nil {
+		return err
+	}
+	if _, err := h.Write([]byte(hostname + labels.String())); err != nil {
+		return err
+	}
+	m.serverJitter = time.Duration(h.Sum64()%60) * time.Second
```
brian-brazil: You've still got the 60s in here.

scrape/manager.go
```diff
@@ -130,6 +134,20 @@ func (m *Manager) reload() {
 	wg.Wait()
 }
 
+// setServerJitter calculates global jitter per server relying on extra label set.
+func (m *Manager) setServerJitter(labels model.LabelSet) error {
+	h := fnv.New64a()
+	hostname, err := os.Hostname()
```
brian-brazil: Could we use the fqdn here?

scrape/manager.go
xjewer: that's a trade-off between distribution and collisions. 
It could affect scrape intervals more than minute, but it's what target hash does in here: https://github.com/prometheus/prometheus/pull/5181/files#diff-71e6b8df5f2eeb6a786cb5cacaa4187dR135 it appends jitter up to scrape interval.

scrape/manager.go
xjewer: what kind of problem we want to resolve with this? hostname is just an extra seed. I can add lookup, but it seems to be over-engineering in here.

scrape/manager.go
brian-brazil: There's no need for it though.

xjewer: you mean modulo operation?

Initially I wanted to pass to the `scrape` interface an offset as a time.Duration, that's why I called it a jitter and implemented as a time offset
https://github.com/prometheus/prometheus/pull/5181/files#diff-6a976999031a4966ff9d1a2734f15e5dR439

Otherwise passing a hash seed would be vague and cause questions like: why is this hash in here.

I might should have add more clear comment, that jitter in interval [0-59) will not affect postpone scrapes on more than the interval time itself.

But I hope with a current implementation: jitter speaks for itself what is this for.

brian-brazil: Passing the hash as an int would be better, it'll keep all the math in one place.

bwplotka:
@xjewer Nice! This graph you provide is for what?

xjewer:
>@xjewer Nice! This graph you provide is for what?

this one is for http probes

xjewer:
fixed merge conflicts and remarks, ready for review

xjewer:
any chance to be merged into next release? cc @brian-brazil

scrape/manager.go
```diff
@@ -54,6 +59,7 @@ type Manager struct {
 	append    Appendable
 	graceShut chan struct{}
 
+	jitterSeed    uint64     // Global jitterSeed seed is using to spread scrape workload across HA setup.
```
brian-brazil: used

scrape/manager.go
```diff
@@ -14,14 +14,19 @@
...
+	"net"
+	"os"
 	"reflect"
 	"sync"
 	"time"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 
+	"github.com/prometheus/common/model"
```
brian-brazil: We're trying to get rid of usage of this library.

scrape/manager.go
```diff
@@ -216,3 +240,45 @@ func (m *Manager) TargetsDropped() map[string][]*Target {
 	}
 	return targets
 }
+
+// getFqdn returns fqdn if it's possible, otherwise falls back hostname
```
brian-brazil: fullstop

scrape/target.go
```diff
@@ -125,12 +125,14 @@ func (t *Target) hash() uint64 {
 }
 
 // offset returns the time until the next scrape cycle for the target.
-func (t *Target) offset(interval time.Duration) time.Duration {
+// It includes the global server jitterSeed for scrapes of HA setup to be spread across the time interval.
+func (t *Target) offset(interval time.Duration, jitterSeed uint64) time.Duration {
 	now := time.Now().UnixNano()
 
+	// base is a pinned to a absolute time, no matter how often offset is called.
```
brian-brazil: Base

scrape/target.go
```diff
@@ -125,12 +125,14 @@ func (t *Target) hash() uint64 {
 }
 
 // offset returns the time until the next scrape cycle for the target.
-func (t *Target) offset(interval time.Duration) time.Duration {
+// It includes the global server jitterSeed for scrapes of HA setup to be spread across the time interval.
```
brian-brazil: Scrapes from multiple Prometheus to try to be at different times.

scrape/target.go
```diff
@@ -125,12 +125,14 @@ func (t *Target) hash() uint64 {
...
+func (t *Target) offset(interval time.Duration, jitterSeed uint64) time.Duration {
 	now := time.Now().UnixNano()
 
+	// base is a pinned to a absolute time, no matter how often offset is called.
 	var (
 		base   = int64(interval) - now%int64(interval)
-		offset = t.hash() % uint64(interval)
+		offset = (t.hash() ^ uint64(jitterSeed)) % uint64(interval)
```
brian-brazil: jitterSeed is already a uint64

scrape/manager.go
xjewer: Any suggestions?

xjewer: should I use `github.com/prometheus/prometheus/pkg/labels` instead?

I can add this to GlobalConfig:
```
// GetExternalLabels wraps external labels as a labels.Labels.
func (c *GlobalConfig) GetExternalLabels() labels.Labels {
	labelSet := make(map[string]string, len(c.ExternalLabels))
	for n, v := range c.ExternalLabels {
		labelSet[string(n)] = string(v)
	}

	return labels.FromMap(labelSet)
}
```

but as this isn't related to the Jitter, let's leave it as is and do clean-up in follow up PRs, I can help you out with it.

Didn't find any issue regarding liquidation the `github.com/prometheus/common/model` library.

image

brian-brazil: model.Labelset is already a map[string]string, so you can use FromMap directly. You can also re-use its Hash method.

xjewer: the problem is, the `model.Labelset` isn't a `map[string]string`, it's `map[LabelName]LabelValue` even if `LabelName` and `LabelValue` are strings, so it's not the same and golang can't do type conversion.

brian-brazil: You can cast it

xjewer: what am I doing wrong?
`labelSet := map[string]string(cfg.GlobalConfig.ExternalLabels)`

```
cannot convert cfg.GlobalConfig.ExternalLabels (type model.LabelSet) to type map[string]string
```

https://play.golang.org/p/EwQjDtB00dJ 

```
prog.go:13:23: cannot convert ls2 (type LabelSet2) to type map[string]string
```
where as LabelSet works as mentioned before.

brian-brazil: Right you are.

brian-brazil:
Thanks!

PR #5184, opened 2019-02-05
Fix prometheus_rule_group_last_evaluation_timestamp_seconds

brian-brazil:
It should be a unix timestamp, not the seconds in the minute.


beorn7:
I have created #5186 to fix.

PR #5187, opened 2019-02-06
Merge v2.7 bugfixes into master

beorn7:


gouthamve:
I need to review this before this PR: https://github.com/prometheus/tsdb/pull/514 I'll get around to it today

beorn7:
DCO tests complains about the merge commit not being DCO'd. But we usually don't do that for merge commits.

beorn7:
This is just to make the things in release-2.7 available in master, too. It's kind of a formality. I only created a PR to let the CI run on it.
In the past, I often just merged back from the release branch directly without a PR.

beorn7:
Note that the actual code difference is just one line. All the commits above show up because cherrypicking has happened before (which we like to avoid).

gouthamve:
> All the commits above show up because cherrypicking has happened before (which we like to avoid)

That was because of a conflict. Both master and the PR changed the UI bundle.

beorn7:
Fine. But can we merge this now?

beorn7:
:bowing_man: Thank you.

PR #5189, opened 2019-02-07
show list of offending labels in the error message in many-to-many scenarios

### LINKED ISSUE #4608: Show offending labels in many-to-many error
## Proposal

When showing the error message

```
Error executing query: many-to-many matching not allowed: matching labels must be unique on one side
```

include the labels, and potentially even the whole metric, that caused it.

I just spent half an hour staring at a query trying to fix this issue; sometimes a very large set of time series has duplication in only a handful of them. Knowing more about the reason for the many-to-many error would have helped a lot.

This message is generated here: https://github.com/prometheus/prometheus/blob/576ee4d3099131e12aa56f7562de416393ad86f4/promql/engine.go#L1222

tariq1890:
FIxes #4608 

This is an initial stab. Reviews appreciated :) @juliusv @brian-brazil

### DIFF
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -81,6 +81,25 @@ func (ls *Labels) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// MatchLabels returns a subset of Labels that matches/does not match with the provided label names based on the 'on' boolean.
+// If on is set to true, it returns the subset of labels that match with the provided label names and its inverse when 'on' is set to false.
+func (ls Labels) MatchLabels(on bool, names ...string) Labels {
+	matchedLabels := Labels{}
+
+	nameSet := map[string]struct{}{}
+	for _, n := range names {
+		nameSet[n] = struct{}{}
+	}
+
+	for _, v := range ls {
+		if _, ok := nameSet[v.Name]; on == ok {
+			matchedLabels = append(matchedLabels, v)
+		}
+	}
+
+	return matchedLabels
+}
+
 // Hash returns a hash value for the label set.
 func (ls Labels) Hash() uint64 {
 	b := make([]byte, 0, 1024)
--- a/pkg/labels/labels_test.go
+++ b/pkg/labels/labels_test.go
@@ -1,0 +1,104 @@
+// Copyright 2019 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package labels
+
+import (
+	"testing"
+)
+
+func TestLabels_MatchLabels(t *testing.T) {
+	labels := Labels{
+		{
+			Name:  "__name__",
+			Value: "ALERTS",
+		},
+		{
+			Name:  "alertname",
+			Value: "HTTPRequestRateLow",
+		},
+		{
+			Name:  "alertstate",
+			Value: "pending",
+		},
+		{
+			Name:  "instance",
+			Value: "0",
+		},
+		{
+			Name:  "job",
+			Value: "app-server",
+		},
+		{
+			Name:  "severity",
+			Value: "critical",
+		},
+	}
+
+	providedNames := []string{
+		"__name__",
+		"alertname",
+		"alertstate",
+		"instance",
+	}
+
+	got := labels.MatchLabels(true, providedNames...)
+	expected := Labels{
+		{
+			Name:  "__name__",
+			Value: "ALERTS",
+		},
+		{
+			Name:  "alertname",
+			Value: "HTTPRequestRateLow",
+		},
+		{
+			Name:  "alertstate",
+			Value: "pending",
+		},
+		{
+			Name:  "instance",
+			Value: "0",
+		},
+	}
+
+	assertSlice(t, got, expected)
+
+	// Now try with 'on' set to false.
+	got = labels.MatchLabels(false, providedNames...)
+
+	expected = Labels{
+		{
+			Name:  "job",
+			Value: "app-server",
+		},
+		{
+			Name:  "severity",
+			Value: "critical",
+		},
+	}
+
+	assertSlice(t, got, expected)
+}
+
+func assertSlice(t *testing.T, got, expected Labels) {
+	if len(expected) != len(got) {
+		t.Errorf("expected the length of matched label names to be %d, but got %d", len(expected), len(got))
+	}
+
+	for i, expectedLabel := range expected {
+		if expectedLabel.Name != got[i].Name {
+			t.Errorf("expected to get Label with name %s, but got %s instead", expectedLabel.Name, got[i].Name)
+		}
+	}
+}
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -1434,9 +1434,16 @@ func (ev *evaluator) VectorBinop(op ItemType, lhs, rhs Vector, matching *VectorM
 		sig := sigf(rs.Metric)
 		// The rhs is guaranteed to be the 'one' side. Having multiple samples
 		// with the same signature means that the matching is many-to-many.
-		if _, found := rightSigs[sig]; found {
+		if duplSample, found := rightSigs[sig]; found {
+			// oneSide represents which side of the vector represents the 'one' in the many-to-one relationship.
+			oneSide := "right"
+			if matching.Card == CardOneToMany {
+				oneSide = "left"
+			}
+			matchedLabels := rs.Metric.MatchLabels(matching.On, matching.MatchingLabels...)
 			// Many-to-many matching not allowed.
-			ev.errorf("many-to-many matching not allowed: matching labels must be unique on one side")
+			ev.errorf("found duplicate series for the match group %s on the %s hand-side of the operation: [%s, %s]"+
+				";many-to-many matching not allowed: matching labels must be unique on one side", matchedLabels.String(), oneSide, rs.Metric.String(), duplSample.Metric.String())
 		}
 		rightSigs[sig] = rs
 	}


pkg/labels/labels.go
```diff
@@ -81,6 +81,27 @@ func (ls *Labels) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// MatchLabels returns a subset of Labels that matches/does not match with the provided label names based on the 'on' boolean.
+// If on is set to true, it returns the subset of labels that match with the provided label names and its inverse when 'on' is set to false.
+func (ls Labels) MatchLabels(on bool, names ...string) Labels {
+	matchedLabels := make(Labels, 0)
```
juliusv: Nit: We try to avoid `make` for stylistic reasons wherever it's not actually necessary (such as zero-length allocations). So in this case, `var matchedLabels Labels` or `matchedLabels := Labels{}` would be closer to the rest of the codebase.

promql/engine.go
```diff
@@ -1435,8 +1435,9 @@ func (ev *evaluator) VectorBinop(op ItemType, lhs, rhs Vector, matching *VectorM
 		// The rhs is guaranteed to be the 'one' side. Having multiple samples
 		// with the same signature means that the matching is many-to-many.
 		if _, found := rightSigs[sig]; found {
+			matchedLabels := rs.Metric.MatchLabels(matching.On, matching.MatchingLabels...)
 			// Many-to-many matching not allowed.
-			ev.errorf("many-to-many matching not allowed: matching labels must be unique on one side")
+			ev.errorf("duplicates found in labels: %s, many-to-many matching not allowed: matching labels must be unique on one side", matchedLabels.String())
```
juliusv: I'm thinking about how to make this error message as clear as possible to inexperienced users. Currently when I do:

    node_cpu_seconds_total / on(instance, job) node_cpu_seconds_total

...I get:

    Error executing query: duplicates found in labels: {instance="localhost:9100", job="node"}, many-to-many matching not allowed: matching labels must be unique on one side

How about changing it to something like:

    Error executing query: found multiple series for the match group {instance="localhost:9100", job="node"} on the right hand-side of the operation [or left-hand side, determine this based on matching.Card]: [include the two series here]; many-to-many matching not allowed: matching labels must be unique on one side

pkg/labels/labels.go
tariq1890: Done!

tariq1890:
@juliusv Thanks for reviewing. I have updated the PR.

pkg/labels/labels_test.go
```diff
@@ -0,0 +1,111 @@
...
+			Name:  "severity",
+			Value: "critical",
+		},
+	}
+
+	assertSlice(t, got, expected)
+}
+
+func assertSlice(t *testing.T, got, expected Labels) {
+	sort.Sort(got)
```
codesome: Sorting is not necessary here (for both `got` and `expected`).
https://github.com/prometheus/prometheus/blob/fc10f6d8144b6a34d07a2cc22448c42375d87de9/pkg/labels/labels.go#L41-L43

pkg/labels/labels_test.go
```diff
@@ -0,0 +1,111 @@
...
+	sort.Sort(expected)
+
+	if len(expected) != len(got) {
+		t.Errorf("expected the length of matched label names to be %d, but got %d", len(expected), len(got))
+	}
+
+	for i, expectedLabel := range expected {
+		expectedLabelName := expectedLabel.Name
```
codesome: I think you can use `expectedLabel.Name` and `got[i].Name` directly without assigning it to another variable.

pkg/labels/labels_test.go
```diff
@@ -0,0 +1,111 @@
...
+		},
+		{
+			Name:  "instance",
+			Value: "0",
+		},
+	}
+
+	assertSlice(t, got, expected)
+
+	// Now try with on set to false.
```
codesome:
```suggestion
	// Now try with 'on' set to false.
```

pkg/labels/labels_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+package labels
+
+import (
+	"sort"
+	"testing"
+)
+
+func TestLabels_MatchLabels(t *testing.T) {
+
```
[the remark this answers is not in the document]
tariq1890: Done.

pkg/labels/labels.go
```diff
@@ -81,6 +81,27 @@ func (ls *Labels) UnmarshalJSON(b []byte) error {
...
+
+	nameSet := map[string]struct{}{}
+
+	for _, n := range names {
+		nameSet[n] = struct{}{}
+	}
+
+	for _, v := range ls {
+		_, ok := nameSet[v.Name]
+		if on == ok {
```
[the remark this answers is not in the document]
tariq1890: Yes! Thanks for pointing that out :)

pkg/labels/labels.go
```diff
@@ -81,6 +81,27 @@ func (ls *Labels) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// MatchLabels returns a subset of Labels that matches/does not match with the provided label names based on the 'on' boolean.
+// If on is set to true, it returns the subset of labels that match with the provided label names and its inverse when 'on' is set to false.
+func (ls Labels) MatchLabels(on bool, names ...string) Labels {
+	matchedLabels := Labels{}
+
+	nameSet := map[string]struct{}{}
+
```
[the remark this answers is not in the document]
tariq1890: Done

pkg/labels/labels_test.go
tariq1890: Ah yes! Thanks

pkg/labels/labels_test.go
tariq1890: Done

pkg/labels/labels_test.go
tariq1890: Sure

tariq1890:
PR comments addressed. Thanks @codesome ! Ready for more rounds of review .

promql/engine.go
```diff
@@ -1435,8 +1435,11 @@ func (ev *evaluator) VectorBinop(op ItemType, lhs, rhs Vector, matching *VectorM
 		// The rhs is guaranteed to be the 'one' side. Having multiple samples
 		// with the same signature means that the matching is many-to-many.
 		if _, found := rightSigs[sig]; found {
+			matchedLabels := rs.Metric.MatchLabels(matching.On, matching.MatchingLabels...)
 			// Many-to-many matching not allowed.
-			ev.errorf("many-to-many matching not allowed: matching labels must be unique on one side")
+			ev.errorf("found multiple series for the match group %s on the right hand-side of the operation "+
+				"[or left-hand side, determine this based on matching.Card]: [include the two series here]; "+
```
juliusv: Heh, I didn't meant to literally paste the parts in `[...]`, but to replace them with computed values...

tariq1890: Whoops my bad :).

tariq1890:
@juliusv I will take a deeper look and improve this PR

tariq1890:
@juliusv I have updated the PR to show the duplicate series against the matched group of labels. 

I get something like the following 👍 

`Error executing query: found multiple series for the match group {instance="localhost:9100"} on the right hand-side of the operation : [{__name__="node_cpu_seconds_total", cpu="0", instance="localhost:9100", job="node", mode="nice"}, {__name__="node_cpu_seconds_total", cpu="0", instance="localhost:9100", job="node", mode="idle"}] ;many-to-many matching not allowed: matching labels must be unique on one side`

pkg/labels/labels.go
```diff
@@ -81,6 +81,25 @@ func (ls *Labels) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// MatchLabels returns a subset of Labels that matches/does not match with the provided label names based on the 'on' boolean.
+// If on is set to true, it returns the subset of labels that match with the provided label names and its inverse when 'on' is set to false.
+func (ls Labels) MatchLabels(on bool, names ...string) Labels {
```
tariq1890: I think this method may be unnecessary for this PR.

tariq1890: Thoughts ?

juliusv: Yeah, how about just using `Labels.Has()` instead of the custom set and then putting the rest directly into the branch where you handle the many-to-many error?

promql/engine.go
```diff
@@ -1434,9 +1434,16 @@ func (ev *evaluator) VectorBinop(op ItemType, lhs, rhs Vector, matching *VectorM
...
+		if duplSample, found := rightSigs[sig]; found {
+			// oneSide represents which side of the vector represents the 'one' in the many-to-one relationship.
+			oneSide := "right"
+			if matching.Card == CardOneToMany {
+				oneSide = "left"
+			}
+			matchedLabels := rs.Metric.MatchLabels(matching.On, matching.MatchingLabels...)
 			// Many-to-many matching not allowed.
-			ev.errorf("many-to-many matching not allowed: matching labels must be unique on one side")
+			ev.errorf("found duplicate series for the match group %s on the %s hand-side of the operation : [%s, %s] "+
```
juliusv: Much better already. Some nits: remove spaces before punctuation like ":" or ";", have spaces *after* them instead.

pkg/labels/labels.go
tariq1890: > then putting the rest directly into the branch where you handle the many-to-many error?

I don't think I understand this

promql/engine.go
tariq1890: This is done!

pkg/labels/labels.go
tariq1890: Also using `Labels.Has()` we assume that `matching.On` will be true. We also have to account for the other case correct?

juliusv: Good point. Yeah, you probably still want the `nameSet` then.

Regarding the confusion: I thought your initial question was about whether you really need this method (since it's only needed in one place and somewhat special) or whether you could just put all of that code in the place where you use it (in `engine.go` when that error occurs). So that's what I meant.

tariq1890: Got it! Thanks for clarifying. 

I guess keeping this method is okay then? I guess MatchLabels is faster but at the cost of space for maintaining a temporary set.

juliusv: I don't have strong feelings either way regarding keeping the method or putting all that code into the place where you use it, so do whatever you like most :) You already have unit tests for the method now, so...

We really don't care about anything regarding speed or efficiency here since it's not a hot path, but just used for that one error message.

tariq1890: Sure! I can vote for keeping the method unless another prometheus maintainer objects :)

juliusv:
:+1: Thanks!

PR #5192, opened 2019-02-07
Update to Bootstrap 4.1.3

jritchieBAE:
This is an updated pull request as discussed in issue #4821 and has been rebased against the current Prometheus master.


juliusv:
Thanks, I'll merge this here into the `bootstrap4` branch and then we can both work on fixing some final bits (in case you do manage to figure out the focus issue from https://github.com/prometheus/prometheus/pull/4821#issuecomment-461486602, let me know).

PR #5199, opened 2019-02-10
 Fix fmt.Errorf error message

JoeWrightss:
1. "disovery" to "discovery" in line 218.
2. "simgle" to "single" in line 883.


codesome:
Thanks!

Issue #4599, opened 2018-09-12
Remember choice "Unhealthy" on target-overview when reloading page


tom932:
## Proposal

Not really a big problem, but nice if it is easy and fast to accomplish:

Today, I analyzed and tried to fix problems with some scrapes. Therefore, to see all the problematic scrape-targets, I used the UI: Status -> Targets.

There, I clicked on `Unhealthy`, which showed the sections where the `State` is `DOWN`.
When I refresh the page, it again shows `All`, and I have to click `Unhealthy` again, which is annoying when I have to refresh the page mutiple times.

It would be more user-friendly, I if the site would go to `Unhealthy` again after a page reload/refresh.

* Prometheus version:

	v2.4.0 docker version

PR #5201, opened 2019-02-10
Web interface Quality of Life improvements

MashaNemt:
Fixes https://github.com/prometheus/prometheus/issues/4599 and introduces a small QoL improvement.

1. Resize text-input area on Graph page not only on text input, but on mouseclick as well. It helps to avoid tediously inputting something in the fields of interest on page reload;
2. Remember "All"/"Unhealthy" choice on target-overview when reloading page.


zuzzas:
@juliusv, care to take a look? Seems really useful.

web/ui/static/js/graph/index.js
```diff
@@ -62,21 +62,26 @@ Prometheus.Graph.prototype.initialize = function() {
...
-
-    // Auto-resize the text area on input.
-    var offset = this.offsetHeight - this.clientHeight;
-    var resizeTextarea = function(el) {
-        $(el).css('height', 'auto').css('height', el.scrollHeight + offset);
-    };
     $(this).on('keyup input', function() { resizeTextarea(this); });
   });
+
+  self.expr.click(function(e) {
```
juliusv: Ha funny, I also just worked on this today in the rewrite of this UI in React that I'm doing. But there it works to just bind on the `input` event alone to catch any changes to the field, then I think the click handler wouldn't be needed anymore in the future. But good for now.

web/ui/static/js/targets.js
```diff
@@ -21,6 +21,15 @@ function showUnhealthy(_, container) {
 }
 
 function init() {
+  if (!localStorage.selectedTab || localStorage.selectedTab == "all-targets"){
+    $("div label:nth-child(1)").addClass("active");
+    $(".table-container").each(showAll);
+  }
+  else if (localStorage.selectedTab == "unhealthy-targets") {
```
juliusv: Style-wise, could you put this on the same line as the brace above it? (`} else if (...`).

web/ui/static/js/targets.js
```diff
@@ -21,6 +21,15 @@ function showUnhealthy(_, container) {
 }
 
 function init() {
+  if (!localStorage.selectedTab || localStorage.selectedTab == "all-targets"){
+    $("div label:nth-child(1)").addClass("active");
```
juliusv: Maybe slightly better to find the labels by ID (`all-targets` / `unhealthy-targets`) and set the parent active?

MashaNemt: Done.

web/ui/static/js/targets.js
MashaNemt: Done.

web/ui/static/js/graph/index.js
MashaNemt: Looks like there is nothing for me to do then. :)

juliusv:
:+1: Thanks!

PR #5206, opened 2019-02-12
Various WAL remote write updates.

tomwilkie:
- Export timestamps in seconds.
- Refactor WAL watcher to remove some duplication
- Factor out logging ratelimit & dedupe middleware.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,7 @@ require (
 	github.com/getsentry/raven-go v0.1.0 // indirect
 	github.com/go-ini/ini v1.21.1 // indirect
 	github.com/go-kit/kit v0.8.0
+	github.com/go-logfmt/logfmt v0.4.0
 	github.com/go-ole/go-ole v1.2.1 // indirect
 	github.com/go-sql-driver/mysql v1.4.0 // indirect
 	github.com/gogo/protobuf v1.2.0
--- a/pkg/logging/dedupe.go
+++ b/pkg/logging/dedupe.go
@@ -1,0 +1,126 @@
+// Copyright 2019 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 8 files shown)


tomwilkie:
ping @cstyan

cstyan:
@tomwilkie we used to have `readToEnd` but now we just start reading from the beginning of the WAL in start to rebuild the labels cache, and only start sending samples after the starting timestamp.

>     * Start following the last segment at startup <- is this safe?
We need to add in some way to replay the existing WAL if we make this change.

tomwilkie:
@cstyan can we rebuild all this from the latest checkpoint?

I added it as the restart time is too slow - esp. when we start pushing to ~5 remote endpoints, and each one needs to replay.

cstyan:
@tomwilkie iiuc the checkpoints only have the series records for active series that started prior to the existing oldest segment, so we still need to replay the segments. 

I think the right fix is to allow remote storage to replay the WAL sooner, or to start from the newest segment but have go routines for each older segment/checkpoint that populate the cache. With the latter we may drop a few samples right on start up.

tomwilkie:
> @tomwilkie iiuc the checkpoints only have the series records for active series that started prior to the existing oldest segment, so we still need to replay the segments.

Isn't that exactly what we need?  We currently read the entire WAL to get the series records for series started prior to the oldest segment.

tomwilkie:
Yeah, thats exactly what TSDB does: https://github.com/prometheus/tsdb/blob/master/head.go#L467

cstyan:
In this whole block: https://github.com/prometheus/tsdb/blob/master/head.go#L466-L490

We get `LastCheckpoint`, and then replay the segments starting from the first one after the checkpoint. `LastCheckpoint` returns the directory and index of the checkpoint, which is why we increment `startFrom` if we successfully read the checkpoint.

So in this example WAL dir:
```
00006132
00006133
00006134
00006135
00006136
00006137
00006138
00006139
00006140
00006141
checkpoint.005089.tmp
checkpoint.006131
```

Last checkpoint would return `checkpoint.06131` and `6131`. And then head `Init` will read the checkpoint followed by segments 6132-6141.

tomwilkie:
(Closing and opening from the grafana repo, so @cstyan can collaborate)

PR #5207, opened 2019-02-13
replay existing WAL segments concurrently

cstyan:
TODO:

- [ ] remove tail var in runWatcher

This should have been obvious to do before but we can replay the WAL segments concurrently in the WAL watcher.

Also adds a benchmark test with testdata.

So for the added `replayWal` function, this is the benchmark result:
```
2m 52s ➜ go test -bench=. -test.benchtime=20s -run=zzz
goos: linux
goarch: amd64
pkg: github.com/prometheus/prometheus/storage/remote
Benchmark_ReplayWAL-8   	      20	1678919242 ns/op
PASS
ok  	github.com/prometheus/prometheus/storage/remote	35.042s
```

and removing all the `go func...` and `WaitGroup` bits from that function the benchmark result:
```
prometheus/storage/remote branch  callum-fix-replay-time [$!?] go v1.11.5 at ☸ops-tools1-us-east4 took 2m 51s ➜ go test -bench=. -test.benchtime=60s -run=zzz
goos: linux
goarch: amd64
pkg: github.com/prometheus/prometheus/storage/remote
Benchmark_ReplayWAL-8   	      20	3876434749 ns/op
PASS
ok  	github.com/prometheus/prometheus/storage/remote	81.758s
```

@tomwilkie @gouthamve 

I guess I should reduce the size of the segment files in testdata before we actually merge this?


brian-brazil:
I'm not sure this is safe, prior segments can contain time series creations that you won't have in subsequent segments.

cstyan:
@brian-brazil when replaying we only care about series records. If I understand correctly, series records don't depend on previous series records, so this is safe.

brian-brazil:
Yes, that'd be safe.

sylr:
Are we willing to add those test WAL to the git ? That's +500MB.

cstyan:
> Are we willing to add those test WAL to the git ? That's +500MB.

No, if we keep the test I'll rewrite it and remove the tesdata files. 

It's unclear if we're going to use this PR anyways so I'm just leaving it for now.

cstyan:
@tomwilkie thoughts here? I still think we should use this; I've updated `StoreSeries` based on our discussion at the end of last week. 

The test still needs to change obviously, I'll delete the data files.

cstyan:
Not needed anymore with the recent commits in #5214

PR #5211, opened 2019-02-13
discovery/kubernetes: fix support for password_file and bearer_token_file

### LINKED ISSUE #5210: password_file ignored in kubernetes_sd_configs
## Bug Report
**What did you do?**
Attempted to use password_file in stead of plaintext password in kubernetes_sd_configs

Authorized http calls

Unauthorized/401 codes thrown in response

In kubernetes_sd_configs only:
* if I use password, authorized, it works
* if I use password_file, unauthorized, it doesn't work
* if I use password and password_file at the same time, no error thrown, but it works

it appears that the password_file variable is ignored completely 

**Environment**

* System information:
Linux 4.15.0-45-generic x86_64

* Prometheus version:
prometheus, version 2.7.1 (branch: HEAD, revision: 62e591f928ddf6b3468308b7ac1de1c63aa7fcf3)
  build user:       root@f9f82868fc43
  build date:       20190131-11:16:59
  go version:       go1.11.5

```
# Global Prometheus Configuration
global:
  scrape_interval: 30s     # By default, scrape targets every 30 seconds.
  evaluation_interval: 30s # By default, scrape targets every 30 seconds.

scrape_configs:

- job_name: kubernetes-apiservers-pre
  kubernetes_sd_configs:
  - role: endpoints
    api_server: https://**********:443
    tls_config:
      insecure_skip_verify: true
    basic_auth:
      username: prometheus
      password_file: /var/run/secrets/prometheus/pre
  scheme: https
  tls_config:
    insecure_skip_verify: true
  basic_auth:
    username: prometheus
    password_file: /var/run/secrets/prometheus/pre
  relabel_configs:
  - source_labels: [__meta_kubernetes_namespace, __meta_kubernetes_service_name, __meta_kubernetes_endpoint_port_name]
    action: keep
    regex: default;kubernetes;https
```

* Logs:
```
prometheus_1  | level=debug ts=2019-02-13T11:16:30.775409678Z caller=klog.go:70 component=k8s_client_runtime func=Infof msg="GET https://********:443/api/v1/endpoints?limit=500&resourceVersion=0 401 Unauthorized in 2 milliseconds"
prometheus_1  | level=error ts=2019-02-13T11:16:30.775560501Z caller=klog.go:94 component=k8s_client_runtime func=ErrorDepth msg="/app/discovery/kubernetes/kubernetes.go:300: Failed to list *v1.Endpoints: Unauthorized"
```

### LINKED ISSUE #5221: bearer_token_file option not working in kubernetes_sd_configs
## Bug Report
**What did you do?**
Attempted to use bearer_token_file in stead of plaintext bearer_token in kubernetes_sd_configs

Authorized http calls

net/http: invalid header field value \"Bearer *******************\\n\" for key Authorization

In kubernetes_sd_configs only:
* if bearer_token_file set, loads data, but throws errors (see logs below). It looks like there is a newline character added even though the file does not contain one.

**Environment**

* System information:
Linux 4.15.0-45-generic x86_64

* Prometheus version:
prometheus, version 2.7.1 (branch: HEAD, revision: 62e591f928ddf6b3468308b7ac1de1c63aa7fcf3)
  build user:       root@f9f82868fc43
  build date:       20190131-11:16:59
  go version:       go1.11.5

```
# Global Prometheus Configuration
global:
  scrape_interval: 30s     # By default, scrape targets every 30 seconds.
  evaluation_interval: 30s # By default, scrape targets every 30 seconds.

scrape_configs:

- job_name: kubernetes-apiservers-pre
  kubernetes_sd_configs:
  - role: endpoints
    api_server: https://**********:443
    tls_config:
      insecure_skip_verify: true
      bearer_token_file: /var/run/secrets/prometheus/pre
  scheme: https
  tls_config:
    insecure_skip_verify: true
  bearer_token_file: /var/run/secrets/prometheus/pre
  relabel_configs:
  - source_labels: [__meta_kubernetes_namespace, __meta_kubernetes_service_name, __meta_kubernetes_endpoint_port_name]
    action: keep
    regex: default;kubernetes;https
```

* Logs:
```
prometheus_1         | level=error ts=2019-02-14T16:40:10.193074594Z caller=klog.go:94 component=k8s_client_runtime func=ErrorDepth msg="github.com/prometheus/prometheus/discovery/kubernetes/kubernetes.go:306: Failed to list *v1.Endpoints: Get https://*********:443/api/v1/endpoints?limit=500&resourceVersion=0: net/http: invalid header field value \"Bearer *******************\\n\" for key Authorization"
prometheus_1         | level=debug ts=2019-02-14T16:40:10.19404676Z caller=klog.go:53 component=k8s_client_runtime func=Verbose.Infof msg="Listing and watching *v1.Service from github.com/prometheus/prometheus/discovery/kubernetes/kubernetes.go:307"
prometheus_1         | level=debug ts=2019-02-14T16:40:10.194199661Z caller=klog.go:70 component=k8s_client_runtime func=Infof msg="GET https://*********:443/api/v1/services?limit=500&resourceVersion=0  in 0 milliseconds"

```

simonpasquier:
Closes #5210 
Closes #5221

### DIFF
--- a/discovery/kubernetes/kubernetes.go
+++ b/discovery/kubernetes/kubernetes.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"sync"
 	"time"
 
@@ -233,8 +234,13 @@ func New(l log.Logger, conf *SDConfig) (*Discovery, error) {
 		kcfg.BearerToken = token
 
 		if conf.BasicAuth != nil {
-			kcfg.Username = conf.BasicAuth.Username
-			kcfg.Password = string(conf.BasicAuth.Password)
+			kcfg.WrapTransport = func(rt http.RoundTripper) http.RoundTripper {
+				return config_util.NewBasicAuthRoundTripper(
+					conf.BasicAuth.Username,
+					conf.BasicAuth.Password,
+					conf.BasicAuth.PasswordFile,
+					rt)
+			}
 		}
 	}
 
--- a/discovery/kubernetes/kubernetes.go
+++ b/discovery/kubernetes/kubernetes.go
@@ -16,8 +16,6 @@ package kubernetes
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
-	"net/http"
 	"sync"
 	"time"
 
@@ -92,6 +90,7 @@ type SDConfig struct {
 	BasicAuth          *config_util.BasicAuth `yaml:"basic_auth,omitempty"`
 	BearerToken        config_util.Secret     `yaml:"bearer_token,omitempty"`
 	BearerTokenFile    string                 `yaml:"bearer_token_file,omitempty"`
+	ProxyURL           config_util.URL        `yaml:"proxy_url,omitempty"`
 	TLSConfig          config_util.TLSConfig  `yaml:"tls_config,omitempty"`
 	NamespaceDiscovery NamespaceDiscovery     `yaml:"namespaces,omitempty"`
 }
@@ -207,40 +206,29 @@ func New(l log.Logger, conf *SDConfig) (*Discovery, error) {
 		if conf.TLSConfig.CertFile != "" || conf.TLSConfig.KeyFile != "" {
 			level.Warn(l).Log("msg", "Configured TLS client certificate is ignored when using pod service account")
 		}
-		if conf.BearerToken != "" {
+		if conf.BearerToken != "" || conf.BearerTokenFile != "" {
 			level.Warn(l).Log("msg", "Configured auth token is ignored when using pod service account")
 		}
 		if conf.BasicAuth != nil {
 			level.Warn(l).Log("msg", "Configured basic authentication credentials are ignored when using pod service account")
 		}
+		if conf.ProxyURL.URL != nil {
+			level.Warn(l).Log("msg", "Configured proxy URL ignored when using pod service account")
+		}
 	} else {
+		rt, err := config_util.NewRoundTripperFromConfig(config_util.HTTPClientConfig{
+			BasicAuth:       conf.BasicAuth,
+			BearerToken:     conf.BearerToken,
+			BearerTokenFile: conf.BearerTokenFile,
+			ProxyURL:        conf.ProxyURL,
+			TLSConfig:       conf.TLSConfig,
+		}, "kubernetes_sd")
+		if err != nil {
+			return nil, err
+		}
 		kcfg = &rest.Config{
-			Host: conf.APIServer.String(),
-			TLSClientConfig: rest.TLSClientConfig{
-				CAFile:   conf.TLSConfig.CAFile,
-				CertFile: conf.TLSConfig.CertFile,
-				KeyFile:  conf.TLSConfig.KeyFile,
-				Insecure: conf.TLSConfig.InsecureSkipVerify,
-			},
-		}
-		token := string(conf.BearerToken)
-		if conf.BearerTokenFile != "" {
-			bf, err := ioutil.ReadFile(conf.BearerTokenFile)
-			if err != nil {
-				return nil, err
-			}
-			token = string(bf)
-		}
-		kcfg.BearerToken = token
-
-		if conf.BasicAuth != nil {
-			kcfg.WrapTransport = func(rt http.RoundTripper) http.RoundTripper {
-				return config_util.NewBasicAuthRoundTripper(
-					conf.BasicAuth.Username,
-					conf.BasicAuth.Password,
-					conf.BasicAuth.PasswordFile,
-					rt)
-			}
+			Host:      conf.APIServer.String(),
+			Transport: rt,
 		}
 	}
 
--- a/discovery/kubernetes/kubernetes.go
+++ b/discovery/kubernetes/kubernetes.go
@@ -16,6 +16,7 @@ package kubernetes
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"sync"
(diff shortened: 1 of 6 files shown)


discovery/kubernetes/kubernetes.go
```diff
@@ -233,8 +234,13 @@ func New(l log.Logger, conf *SDConfig) (*Discovery, error) {
 		kcfg.BearerToken = token
 
 		if conf.BasicAuth != nil {
-			kcfg.Username = conf.BasicAuth.Username
-			kcfg.Password = string(conf.BasicAuth.Password)
+			kcfg.WrapTransport = func(rt http.RoundTripper) http.RoundTripper {
```
brian-brazil: Why are we doing this by hand rather than relying on config_util for all of this?

simonpasquier: There is no way to pass a `http.Client` to the Kubernetes client...

brian-brazil: How about using NewRoundTripperFromConfig?

gouthamve:
Shouldn't this target the `release-2.7` branch? I'll release a `2.7.2` with this fix.

simonpasquier:
@gouthamve this isn't a regression from the previous minor release so I'd say that it can wait for v2.8.0 but it is up to you if you want to get it fixed in `release-2.7`.

discovery/kubernetes/kubernetes.go
simonpasquier: I've done that but I'm a bit worried that it will change some connection settings under the hood. At least we can run prombench to check that it doesn't introduce some regression.

brian-brazil: All the settings are in the transport, the client is a plain wrapper.

discovery/kubernetes/kubernetes.go
```diff
@@ -206,35 +206,29 @@ func New(l log.Logger, conf *SDConfig) (*Discovery, error) {
...
 		}
-		token := string(conf.BearerToken)
-		if conf.BearerTokenFile != "" {
-			bf, err := ioutil.ReadFile(conf.BearerTokenFile)
-			if err != nil {
-				return nil, err
-			}
-			token = string(bf)
+	} else {
+		rt, err := config_util.NewRoundTripperFromConfig(config_util.HTTPClientConfig{
```
brian-brazil: Inline it like we do in https://github.com/prometheus/prometheus/blob/12708acd1516e0d1756006a54684ccaa76f1aaf2/config/config.go#L344, this would also be fragile.

discovery/kubernetes/kubernetes.go
simonpasquier: Still the `client-go` code does some tuning which would be "lost" if we pass our own `Transport`.

https://github.com/prometheus/prometheus/blob/12708acd1516e0d1756006a54684ccaa76f1aaf2/vendor/k8s.io/client-go/transport/cache.go#L85-L98

https://github.com/prometheus/prometheus/blob/12708acd1516e0d1756006a54684ccaa76f1aaf2/vendor/k8s.io/apimachinery/pkg/util/net/http.go#L104-L117

brian-brazil: That doesn't seem to be doing much we're not already doing ourselves.

simonpasquier: `Dialer.Timeout`, `Dialier.KeepAlive` and `TLSHandshakeTimeout` values aren't set by `NewRoundTripperFromConfig`. Wouldn't it make sense to initialize them to sane defaults?

brian-brazil: We use these with contexts, rather than having individual timeouts for each part of a connection. I'd presume k8 does the same.

simonpasquier: I've double checked in the client-go code and done some tests with a local environment and indeed relying only on context should be safe.

brian-brazil:
:+1:

tkornai:
Hi @gouthamve, do you still plan to release a new minor version with this fix by any chance? Or can we expect a 2.8 release any time soon? Thanks!

simonpasquier:
@tkornai the first RC of 2.8 is slated for the end of the month, see https://github.com/prometheus/prometheus/blob/master/RELEASE.md#release-schedule

adamdecaf:
Is `api_server` required now? Prometheus's kubernetes discovery works fine without specifying this, but `promtool` complains that `api_server` isn't set. 

```
FAILED: parsing YAML file infra/14-prometheus.yml: to use custom HTTP client configuration please provide the 'api_server' URL explicitly
```

```yaml
    global:
      scrape_interval: 15s
      scrape_timeout: 10s
      evaluation_interval: 15s
    rule_files:
      - /opt/prometheus-rules/*.yml
    scrape_configs:
      - job_name: 'kubernetes-pods'
        kubernetes_sd_configs:
          - role: pod
        relabel_configs:
          - action: labelmap
            regex: __meta_kubernetes_pod_label_(.+)
          - source_labels: [__meta_kubernetes_namespace]
            action: replace
            target_label: kubernetes_namespace
          - source_labels: [__meta_kubernetes_pod_name]
            action: replace
            target_label: kubernetes_pod_name
          - action: labeldrop
            regex: '(pod_template_hash)'
```

simonpasquier:
> Is `api_server` required now?
No this shouldn't change with this PR. I'll have a look...

simonpasquier:
https://github.com/prometheus/prometheus/pull/5256 should fix it. Thanks for reporting @adamdecaf!

PR #5214, opened 2019-02-13
Various WAL remote_write improvements.

tomwilkie:
- Export timestamps in seconds since epoch. 
- Refactor WAL watcher to remove some duplication.
- Factor out logging ratelimit & dedupe middleware. 
- Dealing with failure: if reading the WAL fails, try again from the next checkpoint.
- Read from the segment after the index for the last checkpoint, not the first segment (should be the same in practice)
- Remove some 'global' state, moving segment numbers to parameters.
- Combine the record decoding metrics into one
- Break out garbage collection into a separate function.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,7 @@ require (
 	github.com/getsentry/raven-go v0.1.0 // indirect
 	github.com/go-ini/ini v1.21.1 // indirect
 	github.com/go-kit/kit v0.8.0
+	github.com/go-logfmt/logfmt v0.4.0
 	github.com/go-ole/go-ole v1.2.1 // indirect
 	github.com/go-sql-driver/mysql v1.4.0 // indirect
 	github.com/gogo/protobuf v1.2.0
--- a/go.mod
+++ b/go.mod
@@ -90,7 +90,7 @@ require (
 	github.com/prometheus/client_golang v0.9.1
 	github.com/prometheus/client_model v0.0.0-20180712105110-5c3871d89910
 	github.com/prometheus/common v0.0.0-20181119215939-b36ad289a3ea
-	github.com/prometheus/tsdb v0.4.0
+	github.com/prometheus/tsdb v0.4.1-0.20190219143357-77d5a7d47a52
 	github.com/rcrowley/go-metrics v0.0.0-20181016184325-3113b8401b8a // indirect
 	github.com/rlmcpherson/s3gof3r v0.5.0 // indirect
 	github.com/rubyist/circuitbreaker v2.2.1+incompatible // indirect
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -126,7 +126,7 @@ var (
 			Namespace: namespace,
 			Subsystem: subsystem,
 			Name:      "queue_highest_sent_timestamp",
-			Help:      "Timestamp from a WAL sample, the highest timestamp successfully sent by this queue.",
+			Help:      "Timestamp from a WAL sample, the highest timestamp successfully sent by this queue, in seconds since epoch.",
 		},
 		[]string{queue},
 	)
@@ -524,7 +524,7 @@ func (t *QueueManager) setHighestSentTimestamp(highest int64) {
 	defer t.timestampLock.Unlock()
 	if highest > t.highestSentTimestamp {
 		t.highestSentTimestamp = highest
-		t.highestSentTimestampMetric.Set(float64(t.highestSentTimestamp))
+		t.highestSentTimestampMetric.Set(float64(t.highestSentTimestamp) / 1000.)
 	}
 }
 
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -216,7 +216,7 @@ type QueueManager struct {
 }
 
 // NewQueueManager builds a new QueueManager.
-func NewQueueManager(logger log.Logger, walDir string, samplesIn *ewmaRate, highestTimestampIn *int64, cfg config.QueueConfig, externalLabels model.LabelSet, relabelConfigs []*pkgrelabel.Config, client StorageClient, flushDeadline time.Duration, startTime int64) *QueueManager {
+func NewQueueManager(logger log.Logger, walDir string, samplesIn *ewmaRate, highestTimestampIn *int64, cfg config.QueueConfig, externalLabels model.LabelSet, relabelConfigs []*pkgrelabel.Config, client StorageClient, flushDeadline time.Duration) *QueueManager {
 	if logger == nil {
 		logger = log.NewNopLogger()
 	} else {
@@ -250,7 +250,7 @@ func NewQueueManager(logger log.Logger, walDir string, samplesIn *ewmaRate, high
 	t.highestSentTimestampMetric = queueHighestSentTimestamp.WithLabelValues(t.queueName)
 	t.pendingSamplesMetric = queuePendingSamples.WithLabelValues(t.queueName)
 	t.enqueueRetriesMetric = enqueueRetriesTotal.WithLabelValues(t.queueName)
-	t.watcher = NewWALWatcher(logger, client.Name(), t, walDir, startTime)
+	t.watcher = NewWALWatcher(logger, client.Name(), t, walDir)
 	t.shards = t.newShards()
 
 	numShards.WithLabelValues(t.queueName).Set(float64(t.numShards))
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -21,8 +21,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"golang.org/x/time/rate"
-
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/gogo/protobuf/proto"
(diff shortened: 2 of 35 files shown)


go.mod
```diff
@@ -29,6 +29,7 @@ require (
 	github.com/getsentry/raven-go v0.1.0 // indirect
 	github.com/go-ini/ini v1.21.1 // indirect
 	github.com/go-kit/kit v0.8.0
+	github.com/go-logfmt/logfmt v0.4.0
```
cstyan: missed go.sum?

storage/remote/wal_watcher.go
```diff
@@ -138,148 +99,153 @@ type WALWatcher struct {
...
-
-	if last == -1 {
-		level.Error(w.logger).Log("err", err)
-		return
+		return err
 	}
 
 	// Backfill from the checkpoint first if it exists.
-	dir, _, err := tsdb.LastCheckpoint(w.walDir)
+	lastCheckpoint, nextIndex, err := tsdb.LastCheckpoint(w.walDir)
```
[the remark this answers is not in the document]
cstyan: this works, nevermind

cstyan:
Okay @tomwilkie I think I've tracked down the issue, at least within the Prometheus instance we looked at earlier. 

It seems that there is still some cases in the way TSDB writes to segments where we can end up reading a page at a time and still not have enough data to read the current record. Because we currently only shift the buffer in `wal.LiveReader` when we've read to the end of the buffered page we get stuck in an infinite loop of not being able to build the record.

The `io.EOF` error isn't propagated from `LiveReader` to `WALWatcher`, so we didn't have any indication that was happening.

Can we wrap `io.EOF` in `LiveReader` with more info? And then we can either change the buffer shfit logic to also shift on EOF, or we can choose to handle EOF for segments we're replaying vs segments we're tailing differently, where for replaying segments we could treat it as some kind of WAL corruption and move on to replaying the next segment?

Thoughts?

storage/remote/wal_watcher.go
```diff
@@ -138,148 +99,153 @@ type WALWatcher struct {
...
+			return err
 		}
 
-		w.currentSegment++
-		w.currentSegmentMetric.Set(float64(w.currentSegment))
+		currentSegment++
+	}
+}
+
+func (w *WALWatcher) findSegmentForIndex(index int) (int, error) {
```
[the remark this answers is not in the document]
tomwilkie: Done.

go.mod
tomwilkie: Doesn't seem to need it?  I ran `go mod tidy` and it didn't add it.  Not sure what I'm suppose to do really.

tomwilkie:
Thanks for fixing the test Callum! 

> It seems that there is still some cases in the way TSDB writes to segments where we can end up reading a page at a time and still not have enough data to read the current record.

Yes that makes sense - I think the prometheus instance we're on was uncleanly shutdown, so the WAL might not have been cleanly written.

> The `io.EOF` error isn't propagated from `LiveReader` to `WALWatcher`

I think we need to make sure that cleanly propagates, and that there is no chance of an infinite loop in the LiveReader.  Then we can deal with the error in the WALWatcher.

tomwilkie:
Ahh, looked at this a bit more I see the infinite loop.  If there is not enough data in the buf, `readRecord` returns EOF, which gets put in the r.err.  If this happens when the buf is already at pageLen, we loop.  This implies a record that erroneously spans a page.

go.mod
simonpasquier: You need to run both `go mod tidy` and `go mod vendor`.
https://github.com/prometheus/prometheus/blob/master/CONTRIBUTING.md#dependency-management

storage/remote/wal_watcher.go
```diff
@@ -133,153 +95,159 @@ type writeTo interface {
...
+			return r, nil
 		}
-		reader = wal.NewLiveReader(segment)
 	}
+
+	return -1, errors.New("failed to find segment for index")
 }
 
-// Use tail true to indicate that the reader is currently on a segment that is
+// Use tail true to indicate thatreader is currently on a segment that is
```
cstyan: > that reader
?

storage/remote/wal_watcher.go
```diff
@@ -133,153 +95,159 @@ type writeTo interface {
...
-
-	if last == -1 {
-		level.Error(w.logger).Log("err", err)
-		return
+		return errors.Wrap(err, "wal.Segments")
 	}
 
 	// Backfill from the checkpoint first if it exists.
-	dir, _, err := tsdb.LastCheckpoint(w.walDir)
+	lastCheckpoint, nextIndex, err := tsdb.LastCheckpoint(w.walDir)
```
cstyan: nit: the name `nextIndex` could be misleading

storage/remote/wal_watcher.go
```diff
@@ -353,59 +291,125 @@ func (w *WALWatcher) watch(wl *wal.WAL, reader *wal.LiveReader, tail bool) error
...
+	}
+
+	if index >= segmentNum {
+		level.Debug(w.logger).Log("msg", "current segment is behind the checkpoint, skipping reading of checkpoint", "current", fmt.Sprintf("%08d", segmentNum), "checkpoint", dir)
+		return nil
+	}
+
+	level.Debug(w.logger).Log("msg", "new checkpoint detected", "new", dir, "currentSegment", segmentNum)
+
+	// This potentially takes a long time, should we run it in another go routine?
```
cstyan: I left this comment a while ago; do we want to do this?

storage/remote/wal_watcher.go
```diff
@@ -110,6 +110,9 @@ type WALWatcher struct {
 
 	quit chan struct{}
 	done chan struct{}
+
+	// For testing, stop when we hit this segment.
+	maxSegment int
```
tomwilkie: We should probably initialise this to -1 so it doesn't fail on a fresh Prometheus.

storage/remote/queue_manager_test.go
```diff
@@ -378,3 +380,26 @@ func (c *TestBlockingStorageClient) NumCalls() uint64 {
...
+	logger = log.With(logger, "caller", log.DefaultCaller)
+
+	for n := 0; n < b.N; n++ {
+		var temp int64
+		c := NewTestBlockedStorageClient()
+		m := NewQueueManager(logger, dir,
+			newEWMARate(ewmaWeight, shardUpdateDuration),
+			&temp, config.DefaultQueueConfig, nil, nil, c, 1*time.Minute)
+		m.watcher.startTime = math.MaxInt64
+		m.watcher.maxSegment = 6158 // n-1
```
tomwilkie: Need to find the max-1 automatically so the test isn't too brittle.

storage/remote/wal_watcher.go
tomwilkie: Nah, it doesn't take that long.  Will remove comment.

storage/remote/storage.go
```diff
@@ -71,7 +71,7 @@ func NewStorage(l log.Logger, reg prometheus.Registerer, stCallback startTimeCal
 		}),
 		highestTimestampMetric: prometheus.NewGauge(prometheus.GaugeOpts{
 			Name: "prometheus_remote_storage_highest_timestamp_in",
```
gouthamve: `_seconds` suffix

storage/remote/wal_watcher.go
```diff
@@ -133,153 +95,171 @@ type writeTo interface {
...
-		// If we've replayed the existing WAL, start tailing.
-		if w.currentSegment == last {
-			tail = true
-		}
-		if tail {
-			level.Info(w.logger).Log("msg", "watching segment", "segment", w.currentSegment)
-		} else {
-			level.Info(w.logger).Log("msg", "replaying segment", "segment", w.currentSegment)
-		}
+	level.Debug(w.logger).Log("msg", "tailing WAL", "lastCheckpoint", lastCheckpoint, "checkpointIndex", checkpointIndex, "currentSegment", currentSegment, "last", last)
```
gouthamve: `last` --> `lastSegment`?

storage/remote/queue_manager.go
```diff
@@ -455,11 +451,12 @@ func (t *QueueManager) calculateDesiredShards() {
 
 	var (
 		timePerSample = samplesOutDuration / samplesOut
-		desiredShards = (timePerSample * (samplesIn + samplesPending + t.integralAccumulator)) / float64(time.Second)
+		desiredShards = (timePerSample * (samplesIn - samplesDropped + samplesPending + t.integralAccumulator)) / float64(time.Second)
```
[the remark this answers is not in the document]
cstyan: @tomwilkie can you help us out here? the desired shards calculation is still a bit confusing.
@gouthamve is thinking that even if we remove samples pending, we would still be able to upshard even if `samplesDropped` is 0 because of the integral accumulator.

tomwilkie:
Yup, will look into it tomorrow.

storage/remote/storage.go
```diff
@@ -59,7 +59,7 @@ func NewStorage(l log.Logger, reg prometheus.Registerer, stCallback startTimeCal
 	}
 	shardUpdateDuration := 10 * time.Second
 	s := &Storage{
-		logger:                 l,
+		logger:                 logging.Dedupe(l, 1*time.Minute),
```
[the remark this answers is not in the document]
tomwilkie: Its mostly for failures coming from servers; we had dedup logic before, but it was inconsistent.  This is much cleaner IMO.

storage/remote/queue_manager.go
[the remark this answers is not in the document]
tomwilkie: Lets look at these calculations in another PR; its important we get this one in.

PR #5215, opened 2019-02-14
Using shebang in shell script

truongnh1992:
Adding `#!/usr/bin/env bash` in order to indicate that scripts use
bash shell for interpreting

Signed-off-by: Nguyen Hai Truong <nguyenhaitruonghp@gmail.com>


cstyan:
Hi @truongnh1992,

The file you've edited is a vendored file from a library we use, and so it shouldn't be edited manually.

Issue #5213, opened 2019-02-13
storage.tsdb.retention.size not being respected >15d


TimSimmons:
**What did you do?**

Started an existing Prometheus instance (previously with 15d retention on v2.3.1 with `--storage.tsdb.retention.time=15d`) on v2.7.1 with `--storage.tsdb.retention.size 473436089549B`

**What did you expect to see?**

Retention slowly creep upwards until it used that many bytes of disk.

**What did you see instead? Under which circumstances?**

Retention inched up past 15d and then went back down. Never exceeding 16 days.

**Environment**

* System information:

	`Linux 4.4.0-112-generic x86_64`

**Prometheus version**:

```
prometheus, version 2.7.1 (branch: HEAD, revision: 62e591f928ddf6b3468308b7ac1de1c63aa7fcf3)
  build user:       root@f9f82868fc43
  build date:       20190131-11:16:59
  go version:       go1.11.5
```

**Prometheus configuration file**:
```
global:
  scrape_interval: 5m
  scrape_timeout: 30s
  evaluation_interval: 1m
alerting:
  alertmanagers:
  - static_configs:
    - targets:
      - <not important>
    scheme: http
    timeout: 10s
rule_files:
- /opt/prometheus/rules/*
- /opt/prometheus/staticrules/*
scrape_configs:
- job_name: <not important>
  scrape_interval: 5m
  scrape_timeout: 30s
  metrics_path: /metrics
  scheme: http
  file_sd_configs:
  - files:
    - /opt/prometheus/services/<not important>.json
    refresh_interval: 5m
- job_name: <not important>
  scrape_interval: 5m
  scrape_timeout: 30s
  metrics_path: /metrics
  scheme: http
  file_sd_configs:
  - files:
    - /opt/prometheus/services/<not important>.json
    refresh_interval: 5m
```

**Logs**:

Nothing out of the ordinary
```
February 13th 2019, 15:05:26.749	compact blocks
February 13th 2019, 15:03:26.172	WAL checkpoint complete
February 13th 2019, 15:02:51.239	head GC completed
February 13th 2019, 15:02:28.742	write block
February 13th 2019, 13:03:39.905	WAL checkpoint complete
February 13th 2019, 13:03:04.751	head GC completed
February 13th 2019, 13:02:40.954	write block
February 13th 2019, 11:03:21.124	WAL checkpoint complete
February 13th 2019, 11:02:48.191	head GC completed
February 13th 2019, 11:02:27.659	write block
[...]

PR #5216, opened 2019-02-14
use the default time retention value only when no size retention is set

krasi-georgiev:
fixes https://github.com/prometheus/prometheus/issues/5213

Now that we have time and size base retention time bases should not have a default value. A default is set only when both - time and size flags are not set.

This change will not affect current installations that rely on the default time based value, and will avoid confusions when only the size retention is set and it is expected that the default time based setting would be no longer in place.


gouthamve:
Should we release this with `2.7.2`?

brian-brazil:
This level of behaviour change feels more like a minor version thing to me.

cmd/prometheus/main.go
```diff
@@ -187,13 +187,13 @@ func main() {
 		Hidden().PlaceHolder("<bytes>").BytesVar(&cfg.tsdb.WALSegmentSize)
 
 	a.Flag("storage.tsdb.retention", "[DEPRECATED] How long to retain samples in storage. This flag has been deprecated, use \"storage.tsdb.retention.time\" instead").
-		Default(defaultRetentionString).SetValue(&oldFlagRetentionDuration)
+		SetValue(&oldFlagRetentionDuration)
 
 	a.Flag("storage.tsdb.retention.time", "How long to retain samples in storage. Overrides \"storage.tsdb.retention\" if this flag is set to anything other than default.").
-		Default(defaultRetentionString).SetValue(&newFlagRetentionDuration)
+		SetValue(&newFlagRetentionDuration)
```
beorn7: "other than default" in the previous line doesn't make sense anymore from the user's perspective as the flag has no default value anymore.

Internally, the value will be the zero value of `time.Duration`, but that has no meaning to the user.

Suggestion:
```
Overrides \"storage.tsdb.retention\" if this flag is set to any non-zero value. If neither this flag nor \"storage.tsdb.retention\" nor "\storage.tsdb.retention.size\" is set, the retention time defaults to 15d.
```

cmd/prometheus/main.go
```diff
@@ -265,36 +267,46 @@ func main() {
...
+		// When the new flag is set it takes precedence.
+		if newFlagRetentionDuration != 0 {
+			cfg.tsdb.RetentionDuration = newFlagRetentionDuration
+		}
 
-	if cfg.tsdb.MaxBlockDuration == 0 {
-		cfg.tsdb.MaxBlockDuration = cfg.tsdb.RetentionDuration / 10
+		if cfg.tsdb.RetentionDuration == 0 && cfg.tsdb.MaxBytes == 0 {
+			cfg.tsdb.RetentionDuration = defaultRetentionDuration
+			level.Warn(logger).Log("msg", "no time or size retention was set so using the default time retention", "duration", defaultRetentionDuration)
```
simonpasquier: IMO it shouldn't a warning. A debug message is probably enough.

brian-brazil: I'd suggest Info, and mentioning what the default is. We've had quite a few user questions around this over the years.

krasi-georgiev: This will only appear when no time or size retention is set and I prefer to be more distinct than just an info message so I would rather keep it as a warning.

@brian-brazil  	the default value is already in the message.

cmd/prometheus/main.go
krasi-georgiev: updated

cmd/prometheus/main.go
simonpasquier: As a user, I find it confusing to get a warning log while I just run `./prometheus`

krasi-georgiev: Confusing in what way?

for me a `warn` is something out of the ordinary. Something you should pay attention to. 
In this case since no explicit flag or a setting was set a warning is raised to notify that a default of 15d retention is automatically applied. Without this it is easy to miss along the other info messages.

simonpasquier: Because the warning wasn't present in previous versions and we aren't changing anything to the way retention settings work. I assume that a significant fraction of users run Prometheus without specifying the retention period and they might be confused seeing a warning log.

krasi-georgiev: well there is a change in the behaviour. Before this was a default and now it is only applied when no time/size retention is set. 


Anyway it is not that important so lets have one more person give an opinion and will leave or change accordingly.

beorn7:
I vote for INFO.

cmd/prometheus/main.go
beorn7: I vote for INFO.

krasi-georgiev: updated, thanks for the feedback.

simonpasquier: Thanks Krasi!

krasi-georgiev:
anymore comments , changes?

brian-brazil:
:+1:

krasi-georgiev:
Thanks Brian.

PR #5226, opened 2019-02-17
Update to Bootstrap 4

juliusv:
@jritchieBAE This is the PR to merge your Bootstrap 4 conversion into master. Most of the changes are from you / BAE, but I added some fix commits on top of that. The /graph page still doesn't look ideal, but neither did it before and also I plan to replace it completely, so I didn't spend more energy on adjusting that.


juliusv:
Hmm, the failing DCO check is because of:

> Commit sha: b8f0a41, Author: jritchieBAE, Committer: Julius Volz; Expected "jritchieBAE 46932687+jritchiebae@users.noreply.github.com", but got "Andrew Chiu andrew.chiu2@baesystems.com".

That originated from squashing the original PR into the `bootstrap4` branch, which already mixed multiple authors together. Maybe not worth trying to fix that now?

web/ui/static/vendor/bootstrap3-typeahead/bootstrap3-typeahead.min.js
```diff
@@ -1,21 +1 @@
-/* =============================================================
- * bootstrap3-typeahead.js v3.1.0
```
juliusv: @jritchieBAE I notice that the typeahead's license header and info got removed in https://github.com/prometheus/prometheus/pull/5192. I think we probably want to keep (re-add) that, so I added another commit to add the latest license header from https://github.com/bassjobsen/Bootstrap-3-Typeahead/blob/master/bootstrap3-typeahead.js.

Issue #4811, opened 2018-10-31
Better error reporting for missing query string parameters


candlerb:
## Proposal
**Use case. Why is this important?**

Currently a missing query parameter is reported as if the query parameter was present with empty string.  Furthermore, the error message does not include the *name* of the faulty parameter.  This makes it hard to identify the problem with the query.

This affects people experimenting with the API for the first time.

## Bug Report
**What did you do?**

```
curl -s 'http://localhost:9090/api/v1/query_range?query=ifHCInOctets%7bifDescr%3d"pppoe-out1"%7d' | python3 -mjson.tool
```

**What did you expect to see?**

Something like:

```
"error": "missing parameter 'start'"
```

**What did you see instead? Under which circumstances?**

```
{
    "status": "error",
    "errorType": "bad_data",
    "error": "cannot parse \"\" to a valid timestamp"
}
```

**Environment**

* System information:

        Linux 4.4.0-138-generic x86_64

* Prometheus version:

    ```
    prometheus, version 2.4.3 (branch: HEAD, revision: 167a4b4e73a8eca8df648d2d2043e21bdb9a7449)
      build user:       root@1e42b46043e9
      build date:       20181004-08:42:02
      go version:       go1.11.1
    ```

BrianHa94:
@brian-brazil I would like to take this one up

codwu:
@BrianHa94  Are you still working on it?  I would like to open a PR for this issue.

BrianHa94:
@codwu Yes, I have a fix. I just need to test before opening a PR.

gouthamve:
Hi @BrianHa94 could you submit the patch? This is still an issue on 2.7.1

palash25:
Hi @gouthamve I would like to work on this

Just to be sure this change would require using `errors.New` with appropriate messages and adding it to the `apiError` struct wherever params are checked like in this snippet `start, err := parseTime(r.FormValue("start"))` right? and ofcourse changing the error response in these [test cases](https://github.com/prometheus/prometheus/blob/master/web/api/v1/api_test.go#L498-L525)

geekodour:
@gouthamve @palash25 Would it be useful if we set `start` to `minTime` and `end` to `time.Now` or something because otherwise I think the error message is helpful only?

codesome:
[...]

PR #5229, opened 2019-02-18
queryRange: Add more descriptive error messages

palash25:
Fixes: https://github.com/prometheus/prometheus/issues/4811

cc @codesome @gouthamve

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -266,6 +266,7 @@ func (api *API) query(r *http.Request) apiFuncResult {
 		var err error
 		ts, err = parseTime(t)
 		if err != nil {
+			err = fmt.Errorf("invalid parameter 'time': %s", err)
 			return apiFuncResult{nil, &apiError{errorBadData, err}, nil, nil}
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


web/api/v1/api.go
```diff
@@ -310,10 +310,12 @@ func (api *API) query(r *http.Request) apiFuncResult {
 func (api *API) queryRange(r *http.Request) apiFuncResult {
 	start, err := parseTime(r.FormValue("start"))
 	if err != nil {
+		err = errors.New("missing parameter 'start'")
```
juliusv: Currently this does `fmt.Errorf("cannot parse %q to a valid timestamp", s)` in `parseTime()`, so completely overwriting that error here would lose the actual timestamp value that couldn't be parsed. Also, the parameter is not technically missing, but invalid (can't be parsed). How about just:

`err = fmt.Errorf("invalid parameter 'start': %s", err)`

Similar for the instance below.

juliusv:
Thanks! The DCO check is currently failing, could you follow https://github.com/prometheus/prometheus/pull/5229/checks?check_run_id=64963847 to add a signoff to your commit?

palash25:
Thanks for the quick review @juliusv 

I have pushed the suggested changes with the commit signed-off

juliusv:
@palash25 Thanks! Oh, one thing I overlooked yesterday is that the function above `queryRange` (`query`), also parses similar parameters and to be consistent we should make the error reporting the same there.

palash25:
@juliusv requested changes made, please review

juliusv:
:+1: Thanks!

PR #5230, opened 2019-02-18
discovery/consul: pass current context to Consul queries

simonpasquier:
The Consul client should abort sooner than later when the context is canceled by the SD manager.

cc @iksaif


simonpasquier:
@brian-brazil PTAL

brian-brazil:
:+1:

PR #5231, opened 2019-02-18
discovery/openstack: pass context to the OpenStack client

simonpasquier:
I've also refactored the `Run()` method that was similar for instances and hypervisors service discovery.


simonpasquier:
I've also sent a PR to the upstream library to support contexts in the client (https://github.com/gophercloud/gophercloud/pull/1480).

simonpasquier:
Updated PR to include https://github.com/gophercloud/gophercloud/pull/1480.

brian-brazil:
:+1:

PR #5236, opened 2019-02-19
Makefile.common: add check_license by default.

krasi-georgiev:
for repos that fail on the check_license check can override it in the main Makefile.

Not sure about removing the build target. Do you know why it was added to run by default?

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -93,7 +93,7 @@ endif
 %: common-% ;
 
 .PHONY: common-all
-common-all: precheck style check_license staticcheck unused test
+common-all: precheck style check_license staticcheck unused test build
 
 .PHONY: common-style
 common-style:
--- a/Makefile.common
+++ b/Makefile.common
@@ -93,7 +93,7 @@ endif
 %: common-% ;
 
 .PHONY: common-all
-common-all: precheck style check_license staticcheck unused test build
+common-all: precheck style check_license staticcheck unused build test
 
 .PHONY: common-style
 common-style:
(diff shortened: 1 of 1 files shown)


SuperQ:
We shouldn't remove build, `make all` should produce a binary.

krasi-georgiev:
added again.


> We shouldn't remove build, `make all` should produce a binary.

@SuperQ 	why is this?

SuperQ:
Because this is the standard for what people expect out of make.

Makefile.common
```diff
@@ -87,14 +87,14 @@ ifeq ($(GOHOSTARCH),amd64)
...
-.PHONY: all
-all: precheck style staticcheck unused build test
-
 # This rule is used to forward a target like "build" to "common-build".  This
 # allows a new "build" target to be defined in a Makefile which includes this
 # one and override "common-build" without override warnings.
 %: common-% ;
 
+.PHONY: common-all
+common-all: precheck style check_license staticcheck unused test build
```
simonpasquier: At least for AlertManager, `build` needs to happen before `test`.

SuperQ: That's odd, why is that? Maybe the alertmanager needs some dependency fixes?

simonpasquier: Because some tests spawn the `alertmanager` binary. Eventually it should be fixed but that's how it works today.

krasi-georgiev: updated. Thanks Simon for the info.

SuperQ: Yea, that's valid.

Maybe we should make test depend on build explicitly by default.

krasi-georgiev: I would say keep it generic and if some repo (like the alertmanager) needs this that it can add this to the main Makefile.

PR #5250, opened 2019-02-22
Add guidance on using `for` vs range duration.

jsha:
I found the distinction rather confusing and had to do a fair bit of
research to understand them better. Hopefully this will help other
users.

Fixes https://github.com/prometheus/docs/issues/1262.

This is based on reading Chapter 5 of the SRE Workbook,
and help from @beorn7 in https://github.com/prometheus/docs/issues/1262.

### DIFF
--- a/docs/configuration/alerting_rules.md
+++ b/docs/configuration/alerting_rules.md
@@ -31,14 +31,44 @@ groups:
       summary: High request latency
 ```
 
-The optional `for` clause causes Prometheus to wait for a certain duration
-between first encountering a new expression output vector element and counting an alert as firing for this element. In this case, Prometheus will check that the alert continues to be active during each evaluation for 10 minutes before firing the alert. Elements that are active, but not firing yet, are in the pending state.
-
 The `labels` clause allows specifying a set of additional labels to be attached
 to the alert. Any existing conflicting labels will be overwritten. The label
 values can be templated.
 
 The `annotations` clause specifies a set of informational labels that can be used to store longer additional information such as alert descriptions or runbook links. The annotation values can be templated.
+
+The optional `for` clause causes Prometheus to wait for a certain duration
+between first encountering a new expression output vector element and counting an alert as firing for this element. In this case, Prometheus will check that the alert continues to be active during each evaluation for 10 minutes before firing the alert. Elements that are active, but not firing yet, are in the pending state.
+
+#### Range duration vs `for`
+
+As a rule of thumb, if an alert is firing too often, it's better to change the
+the range duration rather than adding a `for` clause. For instance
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. These are reference docs, so this sort of guidance does not belong on this page. I think it also needs a bit more nuance, all alerts should have a for of at least 5m for example.

jsha:
Hi! Thanks for the review. Where would be a better place for this guidance? I can tell you that as a user, this was the main place I looked, since it appeared to be the authoritative documentation on the `for` clause. I think it would be ideal for there to be at least a link from here to wherever the more detailed guidance winds up.

beorn7:
I'd say this might very well be worth an entry in the _Best Practices_ section of the site.
The need becomes even more pronounced with [chapter 5](https://landing.google.com/sre/workbook/chapters/alerting-on-slos/) of the SRE workbook.

A whole article could also convey the nuance @brian-brazil was referring to and could explain the various approaches including why the authors of chapter 5 of the SRE workbook advice against `for` clauses while we generally recommend them. (FTR: I, too, believe `for` clauses should be used in many cases, but I follow a more flexible approach than “at least 5m on all alerts”.)

@juliusv what would you think would be the best location for this kind of documentation?

juliusv:
So we have https://prometheus.io/docs/practices/alerting/ already, but that is a high-level doc of what to alert on for certain systems, so this detail does not fit well in there.

I think we should create a new Best Practices docs page about more detailed alerting considerations. Besides the `for` best practices, this could include some other points from the `Alerting` section of this best practices talk: https://docs.google.com/presentation/d/1xhE-pStT7mKmsGKUQpv08eAN0LWx8-FcimSIzUA3xpI/edit#slide=id.g24d86afc62_0_63

It could just start out with the `for` bit though, as we can't expect from @jsha to write up all the other stuff as well.

And yes, we should link to it from the reference docs, as we do for other topics (e.g. https://prometheus.io/docs/instrumenting/pushing/ links to https://prometheus.io/docs/practices/pushing/).

brian-brazil:
We could expand the existing doc, go from what to how.

juliusv:
My concern there would be mixing two layers of abstraction in one doc, but if we separate them clearly enough it could be fine.

beorn7:
As a data point: I finally got https://developers.soundcloud.com/blog/alerting-on-slos published. It is a case study with very long ranges in the rate calculation and a modest deliberate use of `for` clauses (see especially the explanation in the last bullet point, just before the conclusion).

roidelapluie:
We have looked at it during our bug scrub and decided to close this PR as per above comments. This particular advice is also not generally applicable, even if now we have a best practices part.

Feel free to reopen the PR in the "best practices" section : https://prometheus.io/docs/practices/alerting/ and we can see how we can tune it.

Issue #4890, opened 2018-11-21
prometheus_sd_configs_failed_total is not actionable, gauge would be better


juliusv:
The metric `prometheus_sd_configs_failed_total` (https://github.com/prometheus/prometheus/blob/a9050c45f68f00cd9907df9bab96afd1908db934/discovery/manager.go#L44-L49) currently tracks the total number of SD configurations that were not able to be applied during configuration (re)load. However, the actual rate of that counter depends highly on the configuration reload rate, so a bad SD config might stick around forever without actually causing any counter increments. This means that this metric is not very useful for alerts. We should introduce a gauge metric that tells us how many SD configurations are currently invalid / were not able to be instantiated.

infoverload:
I would like to work on this

geekodour:
@infoverload , are you working on this?

infoverload:
@geekodour hey, yes I am!

palash25:
@juliusv I don't see any referenced PR by the original claimer on this issue, would it be okay if I take a crack at this?

juliusv:
@palash25 Yes, please go ahead, I haven't heard anything in a while either from @infoverload about this.

nevill:
@beorn7 This one can be closed since my PR got merged.

PR #5254, opened 2019-02-22
Change prometheus_sd_configs_failed_total to Gauge

nevill:
It's a fix to #4890. Hope it can satisfy @juliusv requirement.

### DIFF
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -195,12 +195,13 @@ func (m *Manager) ApplyConfig(cfg map[string]sd_config.ServiceDiscoveryConfig) e
 	m.providers = nil
 	m.discoverCancel = nil
 
-	failedConfigs.WithLabelValues(m.name).Set(0)
-
+	failedCount := 0
 	for name, scfg := range cfg {
-		m.registerProviders(scfg, name)
+		failedCount += m.registerProviders(scfg, name)
 		discoveredTargets.WithLabelValues(m.name, name).Set(0)
 	}
+	failedConfigs.WithLabelValues(m.name).Set(float64(failedCount))
+
 	for _, prov := range m.providers {
 		m.startProvider(m.ctx, prov)
 	}
@@ -320,8 +321,12 @@ func (m *Manager) allGroups() map[string][]*targetgroup.Group {
 	return tSets
 }
 
-func (m *Manager) registerProviders(cfg sd_config.ServiceDiscoveryConfig, setName string) {
-	var added bool
+// registerProviders returns a number of failed SD config.
+func (m *Manager) registerProviders(cfg sd_config.ServiceDiscoveryConfig, setName string) int {
+	var (
+		failedCount int
+		added       bool
+	)
 	add := func(cfg interface{}, newDiscoverer func() (Discoverer, error)) {
 		t := reflect.TypeOf(cfg).String()
 		for _, p := range m.providers {
@@ -335,7 +340,7 @@ func (m *Manager) registerProviders(cfg sd_config.ServiceDiscoveryConfig, setNam
 		d, err := newDiscoverer()
 		if err != nil {
 			level.Error(m.logger).Log("msg", "Cannot create service discovery", "err", err, "type", t)
-			failedConfigs.WithLabelValues(m.name).Inc()
+			failedCount++
 			return
 		}
 
@@ -424,6 +429,7 @@ func (m *Manager) registerProviders(cfg sd_config.ServiceDiscoveryConfig, setNam
 			return &StaticProvider{TargetGroups: []*targetgroup.Group{{}}}, nil
 		})
 	}
+	return failedCount
 }
 
 // StaticProvider holds a list of target groups that never change.
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -41,10 +41,10 @@ import (
 )
 
 var (
-	failedConfigs = prometheus.NewCounterVec(
-		prometheus.CounterOpts{
-			Name: "prometheus_sd_configs_failed_total",
-			Help: "Total number of service discovery configurations that failed to load.",
+	failedConfigs = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "prometheus_sd_failed_configs",
+			Help: "Current number of service discovery configurations that failed to load.",
 		},
 		[]string{"name"},
 	)
@@ -194,6 +194,9 @@ func (m *Manager) ApplyConfig(cfg map[string]sd_config.ServiceDiscoveryConfig) e
 	m.targets = make(map[poolKey]map[string]*targetgroup.Group)
 	m.providers = nil
 	m.discoverCancel = nil
+
+	failedConfigs.WithLabelValues(m.name).Set(0)
+
 	for name, scfg := range cfg {
 		m.registerProviders(scfg, name)
 		discoveredTargets.WithLabelValues(m.name, name).Set(0)
--- a/discovery/manager_test.go
+++ b/discovery/manager_test.go
@@ -25,6 +25,8 @@ import (
 	"time"
 
 	"github.com/go-kit/kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/config"
 	sd_config "github.com/prometheus/prometheus/discovery/config"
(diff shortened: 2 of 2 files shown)


discovery/manager.go
```diff
@@ -41,8 +41,8 @@ import (
 )
 
 var (
-	failedConfigs = prometheus.NewCounterVec(
-		prometheus.CounterOpts{
+	failedConfigs = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
 			Name: "prometheus_sd_configs_failed_total",
```
simonpasquier: This should be `prometheus_sd_failed_configs` now and the help text needs to be reworded too.

discovery/manager.go
```diff
@@ -191,6 +191,7 @@ func (m *Manager) ApplyConfig(cfg map[string]sd_config.ServiceDiscoveryConfig) e
 		}
 	}
 	m.cancelDiscoverers()
+	failedConfigs.WithLabelValues(m.name).Set(0)
```
juliusv: You'll also need to change the line that currently increments the counter when a failed config is encountered. Instead, it should always be set to the current number of failed SD configs, not a cumulative count over time.

discovery/manager.go
nevill: Sure, wish the rename doesn't break anything.

simonpasquier: We don't have stability guarantees on metrics.

discovery/manager.go
nevill: Not quite sure about your question. Can you check my test within the PR, to see if that's your requirement ?

And there was the call to `Inc()`, I just keep it.
(https://github.com/prometheus/prometheus/blob/e4a741cb7dfa24978097971c2c8d5c8cc12e6cef/discovery/manager.go#L335)

discovery/manager.go
juliusv: Yeah, we don't have stability guarantees on metrics, and in this case this metric changes both its type and semantic meaning, so breaking it is good in any case :) (and gauges shouldn't have a `_total` suffix according to our conventions at https://prometheus.io/docs/practices/naming/)

nevill: No problem, will do it.

discovery/manager.go
juliusv: Ah, ok. I was confused because it's uncommon to use `Inc()` instead of `Set()` on a gauge, but in this case it looks like the right thing because you're setting it to `0` every time the config is applied, and `m.registerProviders()` is only called from `ApplyConfig()`, so it should be fine.

discovery/manager_test.go
```diff
@@ -949,6 +950,82 @@ scrape_configs:
...
 }
 
+func TestGaugeFailedConfigs(t *testing.T) {
+	cfgOneText := `
+scrape_configs:
+- job_name: prometheus
+  consul_sd_configs:
+  - server: "foo:8500"
+    tls_config:
+      cert_file: "/tmp/not_existed"
```
juliusv: `not_existet` -> `non_existent` here and below

discovery/manager_test.go
```diff
@@ -949,6 +950,82 @@ scrape_configs:
...
+	for _, v := range cfgOne.ScrapeConfigs {
+		c[v.JobName] = v.ServiceDiscoveryConfig
+	}
+
+	discoveryManager.ApplyConfig(c)
+	<-discoveryManager.SyncCh()
+
+	metricOne := &dto.Metric{}
+	if counter, err := failedConfigs.GetMetricWithLabelValues(discoveryManager.name); err != nil {
+		t.Fatalf("%s", err)
```
juliusv: This can be just `t.Fatal(err)` (if no message is included).

discovery/manager_test.go
```diff
@@ -949,6 +950,82 @@ scrape_configs:
...
+	} else {
```
juliusv: Omit the `else` (common pattern in Go on error exits), since you are `Fatal()`-ing above, you can just out-dent again here and continue as normal.

discovery/manager_test.go
```diff
@@ -949,6 +950,82 @@ scrape_configs:
...
+		counter.Write(metricOne)
+	}
+
+	if metricOne.GetGauge().GetValue() != 3 {
+		t.Fatalf("Expected to have 3 failed configs")
```
juliusv: Would be good to include the received count as well in this message: "Expected 3 failed configs, got X."

discovery/manager_test.go
```diff
@@ -949,6 +950,82 @@ scrape_configs:
...
+		c[v.JobName] = v.ServiceDiscoveryConfig
+	}
+
+	discoveryManager.ApplyConfig(c)
+	<-discoveryManager.SyncCh()
+
+	metricTwo := &dto.Metric{}
+	if counter, err := failedConfigs.GetMetricWithLabelValues(discoveryManager.name); err != nil {
+		t.Fatalf("%s", err)
+	} else {
```
juliusv: Same comment about `else` and same about `Fatalf` -> `Fatal(err)`.

discovery/manager.go
brian-brazil: That seems a bit racy.

discovery/manager_test.go
nevill: Well, because I use `:=` to initialize variable `counter` twice here and after, it's only available within the `if ... else` scope.

nevill: Well, I'll declare the variable ahead, then use it anyway, check my update soon.

nevill:
@juliusv can you review it again?

nevill:
Well, I can resolve the conflicts and keep everything up to date.

joe-elliott:
Sure, whatever path forward works for me.  I was just looking for a way to contribute to prometheus.

nevill:
Now, it's up to date. @juliusv can you merge it or give me further suggestion ?

discovery/manager_test.go
```diff
@@ -949,6 +951,91 @@ scrape_configs:
...
+
+	metricOne := &dto.Metric{}
+	fcGauge, err = failedConfigs.GetMetricWithLabelValues(discoveryManager.name)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	fcGauge.Write(metricOne)
+
+	failedCount := metricOne.GetGauge().GetValue()
```
beorn7: There is https://godoc.org/github.com/prometheus/client_golang/prometheus/testutil#ToFloat64 to save you the boilerplate code here.

discovery/manager.go
beorn7: Indeed, the problem is that a scrape could happen while the Prometheus server is iterating over configs. It will then underreport the number of failed configs. On the other hand, we are doing something similar with `discoveredTargets` below.

It would be better to let `registerProviders` return the number of failed configs and set them once after the loop in L203.

nevill: I guess you mean ? @beorn7 
```go
failedCount := 0

for name, scfg := range cfg {
	failedCount = m.registerProviders(scfg, name)
	discoveredTargets.WithLabelValues(m.name, name).Set(0)
}

failedConfigs.WithLabelValues(m.name).Set(failedCount)
```

beorn7: That might work. But you want `failedCount += m.registerProviders(scfg, name)`, don't you?

nevill: Yeah, exactly `+=` 😸

discovery/manager_test.go
nevill: Good to know about it. But, I'd like to have another PR for this whole test file `manager_test.go`, what do you think ?

discovery/manager.go
nevill: Due to the signature change of `registerProviders` (will add a return value), I've checked all the references of this method, only used in one place in this file.🤞

beorn7: Sounds all good to me. @cstyan , would be great if you could sanity-check this once @nevill has updated the PR.

And many thanks @nevill for following up.

discovery/manager_test.go
cstyan: If you mean another PR to refactor `manager_test.go` to make use of testutil, I think that's reasonable.

discovery/manager_test.go
```diff
@@ -949,6 +951,91 @@ scrape_configs:
...
+		t.Fatalf("Unable to load YAML config cfgOne: %s", err)
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	discoveryManager := NewManager(ctx, log.NewNopLogger())
+	discoveryManager.updatert = 100 * time.Millisecond
+	go discoveryManager.Run()
+
+	c := make(map[string]sd_config.ServiceDiscoveryConfig)
+	for _, v := range cfgOne.ScrapeConfigs {
```
cstyan: You can just create the `map[string]ServiceDiscoveryConfig` instead of unmarshalling text.

discovery/manager_test.go
nevill: > If you mean another PR to refactor `manager_test.go` to make use of testutil, I think that's reasonable.

Exactly, I will ask you for a review once I create another PR to refactor it.

discovery/manager_test.go
nevill: I will do that with some other tests during refactor.

nevill:
@beorn7 Please review it again

discovery/manager.go
```diff
@@ -320,8 +321,12 @@ func (m *Manager) allGroups() map[string][]*targetgroup.Group {
 	return tSets
 }
 
-func (m *Manager) registerProviders(cfg sd_config.ServiceDiscoveryConfig, setName string) {
-	var added bool
+// registerProviders returns a number of failed SD config
```
cstyan: nit, should end with a `.`

beorn7:
I can totally do recreational code reviewing while enjoying the beaches. :o)

@nevill, this looks good to me. Will merge once you have fixed the nit @cstyan pointed out. (Yeah, just a nit, but we love to get those fixed.)

nevill:
@beorn7 nit is fixed, you can merge now.

PR #5260, opened 2019-02-23
rules: Fix sorting of result from (*Manager).RuleGroups

dsymonds:
The previous code was defective in that it never sorted groups within a
file due to doing a multi-key sort incorrectly.

Signed-off-by: David Symonds <dsymonds@gmail.com>


dsymonds:
@juliusv because I noticed it in the web UI causing rule groups to jump around due to non-determinism.

It looks like this has been bad since 5ff283a7b (June 2017).

juliusv:
:+1: Thanks!

Issue #5263, opened 2019-02-24
Upgrade github.com/miekg/dns


sylr:
Latest version of `github.com/miekg/dns` does not have the ErrTruncated anymore, see: https://github.com/miekg/dns/pull/815

The commit message says:

```
Any program using ErrTruncated will fail to compile when they update to
this version: this is by design: you're doing it wrong. For checking if
a message was truncated you should checked the msg.Truncated boolean;
assuming the unpack didn't fail.
```

Thanos use a recent version of `github.com/miekg/dns` and it is not an issue now but it might become (for Thanos) if https://github.com/prometheus/prometheus/pull/5009 makes it to master as the dependency tree of prometheus within thanos will become bigger as Thanos imports `github.com/prometheus/rules` (indirectly I guess) which will import `github.com/prometheus/config` which itself imports a lot of libs needed for discovery like `github.com/miekg/dns`.

cc @bwplotka

simonpasquier:
Closed by #5020

PR #5264, opened 2019-02-24
Upgrade github.com/miekg/dns

sylr:
Attempt at upgrading github.com/miekg/dns to fix https://github.com/prometheus/prometheus/issues/5263

@stuartnelson3 @beorn7 I blame it and your name appeared around those lines so if you could review it it would be nice.


simonpasquier:
Duplicates https://github.com/prometheus/prometheus/pull/5020 which should be merged soon.

beorn7:
It looks like @simonpasquier is right. I'll close this and merge #5020 as delaying the latter has now caused @sylr to replicate work already done. If there are any concerns, we can still fix/undo before we release.

PR #5267, opened 2019-02-25
Default to bigger remote_write sends

### LINKED ISSUE #5166: proposal: increase default value for `max_samples_per_send`
## Proposal

Default value for `max_samples_per_send` - 100 - is too low for any non-idle Prometheus setup with `remote_write` enabled. It results in too frequent requests to remote storage if Prometheus scrapes more than a few hundred metrics per second. High requests' frequency wastes resources on both Prometheus and remote storage sides, so users have to increase `max_samples_per_send` after the first attempt to write metrics to remote storage.

It would be great if default value for `max_samples_per_send` is increased from 100 to 1000 or even 10K. This would simplify `remote_write` configuration for the majority of users.

bboreham:
Fixes #5166 

Raise the default `MaxSamplesPerSend` to amortise the cost of remote calls across more samples. Lower `MaxShards` to keep the expected max memory usage roughly the same.

Some adjustment of comment wording to make them less easy to misunderstand.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -105,15 +105,15 @@ var (
 
 	// DefaultQueueConfig is the default remote queue configuration.
 	DefaultQueueConfig = QueueConfig{
-		// With a maximum of 1000 shards, assuming an average of 100ms remote write
-		// time and 100 samples per batch, we will be able to push 1M samples/s.
-		MaxShards:         1000,
+		// With a maximum of 200 shards, assuming an average of 100ms remote write
+		// time and 500 samples per batch, we will be able to push 1M samples/s.
+		MaxShards:         200,
 		MinShards:         1,
-		MaxSamplesPerSend: 100,
+		MaxSamplesPerSend: 500,
 
-		// Each shard will have a max of 500 samples pending in it's channel, plus the pending
-		// samples that have been enqueued. Theoretically we should only ever have about 600 samples
-		// per shard pending. At 1000 shards that's 600k.
+		// Each shard will have a max of 500 samples pending in its channel, plus the pending
+		// samples that have been enqueued. Theoretically we should only ever have about 1000 samples
+		// per shard pending. At 200 shards that's 200k.
 		Capacity:          500,
 		BatchSendDeadline: model.Duration(5 * time.Second),
 
--- a/config/config.go
+++ b/config/config.go
@@ -111,10 +111,10 @@ var (
 		MinShards:         1,
 		MaxSamplesPerSend: 500,
 
-		// Each shard will have a max of 500 samples pending in its channel, plus the pending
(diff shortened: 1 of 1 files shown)


tomwilkie:
Not against this per se, but we need to do the maths wrt to all the other params to make sure it makes sense.  For instance, the queue wait time might need tuning.  Plus with bigger batches I think you might be more likely to hit a quasi live lock in the sharding code when one shard gets stuck.  Lets get 2.8 release out in the next few days then come back to this.

beorn7:
Is there any update on this, dear remote-write experts?

ashutoshgngwr:
@beorn7 I don't think this is ever getting merged but applying the suggested defaults using the Prometheus config file did bring its memory usage in check.

beorn7:
I'm really not qualified when it comes to remote-write. @tomwilkie @bboreham @cstyan @csmarchbanks (semi-randomly picked experts on the matter) WDYT about this PR ?

beorn7:
(Context: We reached this during our weekly bug scrub, and the people in the room couldn't make a call here on their own.)

csmarchbanks:
You can see some recent conversation in https://github.com/prometheus/prometheus/issues/5166 around what default values to choose. I proposed 500, as I have seen some systems start to break down with 1000 samples per send. Once we have a better default decided Bryan can feel free to update this PR, or someone can make a new one.

csmarchbanks:
@bboreham Do you have any interest in resurrecting this PR and setting the default to 500 samples per send/200 max shards as discussed in #5166?

bboreham:
Sure.

bboreham:
I have updated this PR to 200 shards and 500 samples.

config/config.go
```diff
@@ -105,15 +105,15 @@ var (
...
-		MaxSamplesPerSend: 100,
+		MaxSamplesPerSend: 500,
 
-		// Each shard will have a max of 500 samples pending in it's channel, plus the pending
-		// samples that have been enqueued. Theoretically we should only ever have about 600 samples
-		// per shard pending. At 1000 shards that's 600k.
+		// Each shard will have a max of 500 samples pending in its channel, plus the pending
+		// samples that have been enqueued. Theoretically we should only ever have about 1000 samples
+		// per shard pending. At 200 shards that's 200k.
 		Capacity:          500,
```
csmarchbanks, line 117: Would you also increase the default Capacity to 2500 so it stays at 5 full requests being buffered?

csmarchbanks (approved):
Thanks!

roidelapluie:
This seems a big change -- since it has been in the making for a very long time and would require a RC it will be for prometheus 2.22.

PR #5274, opened 2019-02-26
Insert .idea/ in file .gitignore

truongnh1992:
Ignore folder `.idea/` when open project Prometheus with
IntelliJ IDEA.

Signed-off-by: Nguyen Hai Truong <truongnh@vn.fujitsu.com>


brian-brazil:
Thanks for your PR. Our policy is that such things belong in your global gitignore, rather than us ending up either all possible editors listed in these files.

Issue #3726, opened 2018-01-23
Support differentiation between empty label and non-existing labels in k8


discordianfish:
**What did you do?**
I've tried to use the Kubernetes SD and relabling to only scrape master nodes. In Kubernetes, master nodes are tagged with the label: `node-role.kubernetes.io/master: ""` which becomes `__meta_kubernetes_node_label_node_role_kubernetes_io_master=""` in Prometheus after sanitizing.

Unfortunately this doesn't seem to be possible with relabling because it's not possible to differentiate between a nonexisting and empty label. This for example will always match, since if the key doesn't exist the regexp gets matched against "".

* Prometheus version: 2.1

brian-brazil:
This is not something we can really handle in relabelling, however service discovery could provide additional metadata to make this possible.

cstyan:
What would the solution look like here? A new `Role` in `discovery/kubernetes.go` to get nodes but drop ones without `ode-role.kubernetes.io/master: ""`? `RoleMaster`? Or should we set an explicit value `true` for that meta label?

brian-brazil:
We'd either add a new metadata entry that is a list with a list of all labels/annotations that are present, or  a label per k8 label that has the value true.

discordianfish:
Not sure how the "label per k8 label that has the value true" should work (the value is "" for these labels) but a entry with a list of all labels/annotations sounds like a good idea. :+1:

brian-brazil:
I'm saying we could represent this as either a list, or a map. A map may be easier for users to use, but is more verbose.

discordianfish:
Ah got it.. hrm Isee, yeah checking if something is set or not would be indeed easier to do with a map. I'm fine either way.

Quentin-M:
+1

This makes it hard to monitor master and worker nodes separately.

brancz:
At least for all the "official" `*.kubernetes.io` type annotations/labels, this should be ok to have as metadata in the SD.

nevill:
@brancz Unfortunately, `node-role.kubernetes.io/master` is not a member of well_known_labels ([kubelet](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/well_known_labels.go), [scheduler](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/api/well_known_labels.go), etc). It's usually set during `mark control plane phase` when kubeadm is running. Shall we still consider it as **official** ?

[...]

PR #5283, opened 2019-02-28
Add meta role label

Labbs:
Add meta role label like : __meta_kubernetes_node_role="worker"

### DIFF
--- a/discovery/kubernetes/node.go
+++ b/discovery/kubernetes/node.go
@@ -163,12 +163,10 @@ func nodeLabels(n *apiv1.Node) model.LabelSet {
 			}
 		case k == roleLabel && v != "":
 			ls[nodeRoleLabel] = lv(v)
+		default:
+			ln := strutil.SanitizeLabelName(nodeLabelPrefix + k)
+			ls[model.LabelName(ln)] = lv(v)
 		}
-	}
-
-	for k, v := range n.Labels {
-		ln := strutil.SanitizeLabelName(nodeLabelPrefix + k)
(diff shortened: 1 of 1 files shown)


discovery/kubernetes/node.go
```diff
@@ -140,16 +141,31 @@ func nodeSourceFromName(name string) string {
...
+	rolePrefix = "node-role.kubernetes.io/"
+	roleLabel  = "kubernetes.io/role"
 )
 
 func nodeLabels(n *apiv1.Node) model.LabelSet {
 	ls := make(model.LabelSet, len(n.Labels)+len(n.Annotations)+1)
 
 	ls[nodeNameLabel] = lv(n.Name)
 
+	for k, v := range n.Labels {
```
brian-brazil: We already have this under node_label as far as I can see.

Labbs: yes exactly, i change that

brian-brazil: My point is that this is already supported and available to relabelling, so your change isn't needed.

Labbs: yes when we use `kubernetes.io/role="worker"`, it's simple.
But with `node-role.kubernetes.io/worker=""`, how would you do with this syntax? I tried different things without success.

simonpasquier: Hmm, I suspect that you're hitting https://github.com/prometheus/prometheus/issues/3726.

roidelapluie:
We have decided to close this PR during our bug scrub per above comments.

PR #5286, opened 2019-02-28
Refactor and fix queue resharding

tomwilkie:
Right now just refactoring to try and simplify the code.  Then I'll expose a lot more of the resharding logic as metrics, then try and fix. /cc @cstyan

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -190,11 +190,9 @@ type QueueManager struct {
 func NewQueueManager(logger log.Logger, walDir string, samplesIn *ewmaRate, cfg config.QueueConfig, externalLabels model.LabelSet, relabelConfigs []*pkgrelabel.Config, client StorageClient, flushDeadline time.Duration) *QueueManager {
 	if logger == nil {
 		logger = log.NewNopLogger()
-	} else {
-		logger = log.With(logger, "queue", client.Name())
 	}
 	t := &QueueManager{
-		logger:         logger,
+		logger:         log.With(logger, "queue", client.Name()),
 		flushDeadline:  flushDeadline,
 		cfg:            cfg,
 		externalLabels: externalLabels,
@@ -414,11 +412,15 @@ func (t *QueueManager) calculateDesiredShards() {
 
 	var (
 		timePerSample = samplesOutDuration / samplesOut
-		desiredShards = (timePerSample * (samplesIn - samplesDropped + samplesPending + t.integralAccumulator)) / float64(time.Second)
+		desiredShards = (timePerSample * (samplesPending + t.integralAccumulator)) / float64(time.Second)
 	)
 	level.Debug(t.logger).Log("msg", "QueueManager.caclulateDesiredShards",
-		"samplesIn", samplesIn, "samplesDropped", samplesDropped,
-		"samplesOut", samplesOut, "samplesPending", samplesPending,
+		"samplesIn", samplesIn,
+		"samplesOut", samplesOut,
+		"samplesDropped", samplesDropped,
+		"samplesPending", samplesPending,
+		"samplesOutDuration", samplesOutDuration,
+		"timePerSample", timePerSample,
 		"desiredShards", desiredShards)
 
 	// Changes in the number of shards must be greater than shardToleranceFraction.
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -396,15 +396,13 @@ func (t *QueueManager) calculateDesiredShards() {
 	// (received - send) so we can catch up with any backlog. We use the average
 	// outgoing batch latency to work out how many shards we need.
 	var (
-		samplesIn      = t.samplesIn.rate()
-		samplesOut     = t.samplesOut.rate()
-		samplesDropped = t.samplesDropped.rate()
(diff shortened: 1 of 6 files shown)


storage/remote/queue_manager.go
```diff
@@ -731,17 +678,14 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
 
 		if err == nil {
 			succeededSamplesTotal.WithLabelValues(s.qm.queueName).Add(float64(len(samples)))
-			now := time.Now()
-			s.qm.setLastSendTimestamp(now)
-			s.qm.setHighestSentTimestamp(highest)
+			s.qm.highestSentTimestampMetric.Set(float64(highest / 1000))
```
cstyan: are you sure we don't want to check that highest > the value we last set for `highestSentTimestampMetric` first?

cstyan: ah I see, `maxGauge` :+1:

storage/remote/queue_manager.go
```diff
@@ -425,7 +389,6 @@ func (t *QueueManager) updateShardsLoop() {
 }
 
 func (t *QueueManager) calculateDesiredShards() {
-	t.samplesIn.tick()
```
gouthamve: Is this the fix for the issue?

storage/remote/queue_manager.go
```diff
@@ -731,17 +678,14 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
...
-			s.qm.setHighestSentTimestamp(highest)
+			s.qm.highestSentTimestampMetric.Set(float64(highest / 1000))
 			return nil
 		}
 
 		if _, ok := err.(recoverableError); !ok {
 			return err
 		}
 		retriedSamplesTotal.WithLabelValues(s.qm.queueName).Add(float64(len(samples)))
-		level.Error(s.qm.logger).Log("err", err)
```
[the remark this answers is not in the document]
tomwilkie: Yeah, some retryable errors won't be logged.  I thought it was a bit spammy.

storage/remote/queue_manager.go
tomwilkie: This is a correctness fix; we shouldn't be ticking this N times, one for each queue.  This isn't the main fix though.

tomwilkie:
> Changes look sane but what was the issue and how was it fixed by this? As far as I can see this is just moving and adding some metrics around.

Initial changes were just moving stuff around, sorry.  Made it easier for me to see the real bug.

The real issue was that using two rates (one measure before appending to the WAL, one after) was never going to work - if you get behind, the reading from the WAL rate will go high and the writing to rate will stay the same, and indicate you need to scale down - which is not what you want to do when you're behind!

What we do now is use the two timestamps we track - the max appending to the WAL and the max sent.  We multiply the difference by the rate at which we are appending to WAL to estimate the number of pending samples. 

Another fix was to measure the ratio of dropped samples and not the number of dropped samples, and then use this in calculating the pending samples.  This fixes queue that make heavy use of relabelling.

tomwilkie:
screenshot 2019-03-01 at 08 55 43

storage/remote/queue_manager.go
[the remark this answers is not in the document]
gouthamve: Can we make it a debug log? This is going to be useful when debugging upstream issues.

storage/remote/queue_manager.go
```diff
@@ -437,9 +398,11 @@ func (t *QueueManager) calculateDesiredShards() {
 	var (
 		samplesIn          = t.samplesIn.rate()
 		samplesOut         = t.samplesOut.rate()
-		samplesDropped     = t.samplesDropped.rate()
-		samplesPending     = samplesIn - samplesDropped - samplesOut
+		samplesKeptRatio   = samplesOut / (t.samplesDropped.rate() + samplesOut)
 		samplesOutDuration = t.samplesOutDuration.rate()
+		highestSent        = t.highestSentTimestampMetric.value
```
gouthamve: This is protected by a mutex, need to use that. Maybe add a `Get` method to `maxGauge`

gouthamve:
LGTM apart from the two comments.

storage/remote/queue_manager.go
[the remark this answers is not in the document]
tomwilkie: Done

storage/remote/queue_manager.go
```diff
@@ -437,9 +398,11 @@ func (t *QueueManager) calculateDesiredShards() {
...
-		samplesDropped     = t.samplesDropped.rate()
-		samplesPending     = samplesIn - samplesDropped - samplesOut
+		samplesKeptRatio   = samplesOut / (t.samplesDropped.rate() + samplesOut)
 		samplesOutDuration = t.samplesOutDuration.rate()
+		highestSent        = t.highestSentTimestampMetric.Get()
+		highestRecv        = highestTimestamp.Get()
+		samplesPending     = (highestRecv - highestSent) * samplesIn * samplesKeptRatio
 	)
 
 	// We use an integral accumulator, like in a PID, to help dampen oscillation.
```
cstyan: I think we can remove the integral accumulator now

PR #5289, opened 2019-03-01
Return multiple errors when fanout close

MIBc:
Use tsdb.MultiError to return multiple errors when fanout close.
Currently, it returns only one error.

Signed-off-by: MIBc <lvjiawei@cmss.chinamobile.com>

### DIFF
--- a/storage/fanout.go
+++ b/storage/fanout.go
@@ -24,6 +24,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/tsdb"
 )
 
 type fanout struct {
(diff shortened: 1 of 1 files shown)


storage/fanout.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/tsdb"
```
roidelapluie:
```suggestion
	"github.com/prometheus/prometheus/tsdb"
```

roidelapluie:
This PR has been obsoleted by recent refactoring of the TSDB. We now correctly return multiple errors.

Thanks!

PR #5291, opened 2019-03-01
*: cut 2.7.2

gouthamve:
@beorn7 @brian-brazil


beorn7:
:+1: if we really don't want to update TSDB here (which is not an easy call as it fixes a performance regression introduced with 2.7).

gouthamve:
I'd rather not include the TSDB changes as it needs to pull in a ton of other changes. We don't cut release branches for TSDB.

PR #5292, opened 2019-03-01
updated tsdb to 0.6.0

krasi-georgiev:
as part of the update also added the new
storage.tsdb.allow.overlapping.block flag

cc @codesome 	


cmd/prometheus/main.go
```diff
@@ -198,6 +198,9 @@ func main() {
 	a.Flag("storage.tsdb.no-lockfile", "Do not create lockfile in data directory.").
 		Default("false").BoolVar(&cfg.tsdb.NoLockfile)
 
+	a.Flag("storage.tsdb.allow.overlapping.block", "Allow overlapping blocks which in-turn enables vertical compaction and vertical query merge.").
```
codesome: I see dots are used to denoting a subsection and not to separate the words.
```suggestion
	a.Flag("storage.tsdb.allow-overlapping-blocks", "Allow overlapping blocks which in-turn enables vertical compaction and vertical query merge.").
```

krasi-georgiev: yeah I am also thinking it should be plural everywhere - AllowOverlappingBlock -> AllowOverlappingBlocks

krasi-georgiev: here is the change and I will update it here after it is merged
https://github.com/prometheus/tsdb/pull/537

krasi-georgiev: updated

cmd/prometheus/main.go
```diff
@@ -198,6 +198,9 @@ func main() {
 	a.Flag("storage.tsdb.no-lockfile", "Do not create lockfile in data directory.").
 		Default("false").BoolVar(&cfg.tsdb.NoLockfile)
 
+	a.Flag("storage.tsdb.allow-overlapping-blocks", "Allow overlapping blocks which in-turn enables vertical compaction and vertical query merge.").
```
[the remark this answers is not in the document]
krasi-georgiev: experimental is logical, but why hidden? What complications do you think it might bring?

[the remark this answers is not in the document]
codesome: @gouthamve is on a long flight, so adding my thoughts here.

Unusual overlaps will go undetected if there is an actual bug in tsdb. So keeping this hidden might partially make sure that the user knows what he/she is doing if enabled.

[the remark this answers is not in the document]
krasi-georgiev: for me this explains the reason to make it experimental. 
Hidden OTOH for me means should be only used for development, tests and troubleshooting.

PR #5293, opened 2019-03-01
Expose the WAL watcher WriteTo interface

csmarchbanks:
In order for other applications to easily use the WALWatcher it would be nice for this interface to be exposed. Is there any reason this interface is not exposed right now?

cc: @cstyan / @tomwilkie / @gouthamve

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -155,7 +155,7 @@ type StorageClient interface {
 }
 
 // QueueManager manages a queue of samples to be sent to the Storage
-// indicated by the provided StorageClient. Implements writeTo interface
+// indicated by the provided StorageClient. Implements WriteTo interface
 // used by WAL Watcher.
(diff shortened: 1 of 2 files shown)


cstyan:
@csmarchbanks the WAL watcher is still subject to change, and is changing regularly (daily sometimes). We plan to move the WAL watcher code to tsdb itself, I think at that point it would make more sense to expose the the `WriteTo` interface.

That said, I'm not against exposing it now.

csmarchbanks:
Thanks for the info! My personal opinion is that if it is exposed sooner perhaps more people will play around with the WALWatcher to flush out any issues. I am hoping to be one of those people, but we are still doing some design.

It is possible to use the WALWatcher without the WriteTo interface exposed - just harder to extend interfaces and such.

csmarchbanks:
This isn't necessary and will be moved to tsdb

PR #5296, opened 2019-03-03
docs: minor improvements to the service discovery README.md

tariq1890:
i) Increased the size of the Service Discovery Readme title
ii)  Changed `TargetGroups` to "target groups" as it has been relocated and renamed to another package.

### DIFF
--- a/discovery/README.md
+++ b/discovery/README.md
@@ -1,4 +1,4 @@
-### Service Discovery
+# Service Discovery
 
 This directory contains the service discovery (SD) component of Prometheus.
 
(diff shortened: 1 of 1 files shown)


juliusv:
:+1: on the title, not so sure about the `TargetGroups` quoting though, since I would expect anything in backticks to be a literal quote of something existing like that in code, but `TargetGroups` does not exist as a type. Actually now I just noticed that the type is not even called `TargetGroup` anymore, but has been factored out into a separate package, so it'd be `targetgroup.Group` now. The link to the type in https://github.com/prometheus/prometheus/blob/master/discovery/README.md#the-sd-interface is also broken because of that.

How about fixing the link, but then just using "target groups" (without code quotes) as a term in the rest of the doc? The code snippets that mention `[]config.TargetGroup` would also need updating.

tariq1890:
@juliusv Makes sense. will do!

tariq1890:
Review comment addressed & Commit/PR titles changed.

juliusv:
Wonderful, thanks!

PR #5303, opened 2019-03-04
Alert API Values as Strings

dtrejod:
This patch changes the `api/v1/alert` endpoint to return string alert values mimicking the `api/v1/query` return values. The change ensures `null`, and `+-Inf` values can be properly parsed.

We are using the [`client_golang`](https://github.com/prometheus/client_golang) library in our applications to interact with the `api/v1/alert` endpoint. We've recently hit an issue where the library failed to parse the return values from a `+Inf` alert value. We traced the problem back to the alert api not using string return values. We noticed `NaN` values are also affected. 

**Problem**
Create an alert with a `+Inf` return value; for example ```count(up)/0 != 0```. When curling the api endpoint you can see the returned value is as follows:

```
$ curl -s http://$(hostname)/prometheus/api/v1/rules |jq '.data.groups[] |select(.name == "breaking")'
{
  "name": "breaking",
  "file": "rules.yml",
  "rules": [
    {
      "name": "InfRule",
      "query": "count(up) / 0 != 0",
      "duration": 30,
      "labels": {
        "severy": "p2"
      },
      "annotations": {
        "decription": " This rule should break ",
        "summary": " Breaking +INF rule "
      },
      "alerts": [
        {
          "labels": {
            "alertname": "InfRule",
            "severy": "p2"
          },
          "annotations": {
            "decription": " This rule should break ",
            "summary": " Breaking +INF rule "
          },
          "state": "firing",
          "activeAt": "2019-02-26T12:43:21.527666941-05:00",
          "value": 1.7976931348623157e+308
        }
      ],
      "health": "ok",
      "type": "alerting"
    }
  ],
  "interval": 60
}
```

Running the same query against the `query` api results in:

```
$ curl -s  -g 'http://$(hostname)/prometheus/api/v1/query?query=count(up)/0'|jq .
{
  "status": "success",
  "data": {
    "resultType": "vector",
    "result": [
      {
        "metric": {},
        "value": [
          1551461908.309,
          "+Inf"
        ]
      }
    ]
  }
}
```

The resulting error when parsing using the `client_golang` library is:
```
bad_response: invalid character '+' looking for beginning of value
```

**Solution**
This patch converts the return values for the `alert` api to strings. Running the same query now results with:

```
{                                                           
  "name": "breaking",    
  "file": "rules.yml",
  "rules": [
    {                
      "name": "InfRule",
      "query": "count(up) / 0 != 0",
      "duration": 30,
      "labels": {       
        "severity": "p2"                          
      },             
      "annotations": {
        "decription": " This rule should break {{ $labels.value }}",
        "summary": " Breaking +INF rule "
      },              
      "alerts": [                                                  
        {                               
          "labels": {
            "alertname": "InfRule",
            "severity": "p2"
          },         
          "annotations": {         
            "decription": " This rule should break ",
            "summary": " Breaking +INF rule "
          },              
          "state": "firing",                         
          "activeAt": "2019-03-04T09:39:21.527666941-05:00",
          "value": "+Inf"
        }                   
      ],                                                    
      "health": "ok",   
      "type": "alerting"
    },  
    {                
      "name": "NanRule",
      "query": "(count(up) - count(up)) / 0 != 0",
      "duration": 30,
      "labels": {
        "severity": "p2"
      },
      "annotations": {                                             
        "decription": " This rule should break {{ $labels.value}}",
        "summary": " Breaking Nan rule "
      },                           
      "alerts": [           
        {            
          "labels": {              
            "alertname": "NanRule",                  
            "severity": "p2"                 
          },              
          "annotations": {                           
            "decription": " This rule should break ",       
            "summary": " Breaking Nan rule "
          },                
          "state": "firing",                                
          "activeAt": "2019-03-04T09:39:21.527666941-05:00",
          "value": "NaN"
        }
      ],             
      "health": "ok",   
      "type": "alerting"                          
    }                
  ],             
  "interval": 60        
}               
```
**Prometheus Version:** 2.6.0

### DIFF
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -450,7 +450,10 @@ $ curl http://localhost:9090/api/v1/rules
                                     "severity": "page"
                                 },
                                 "state": "firing",
-                                "value": 1
+                                "value": [
+                                  1552085678.2,
+                                  "1"
+                                ]
                             }
                         ],
                         "annotations": {
@@ -507,7 +510,10 @@ $ curl http://localhost:9090/api/v1/alerts
                     "alertname": "my-alert"
                 },
                 "state": "firing",
-                "value": 1
+                "value": [
+                  1552085678.2,
+                  "1"
+                ]
             }
         ]
     },
--- a/rules/alerting.go
+++ b/rules/alerting.go
@@ -85,11 +85,12 @@ type Alert struct {
 	Value float64
 	// The interval during which the condition of this alert held true.
 	// ResolvedAt will be 0 to indicate a still active alert.
-	ActiveAt   time.Time
-	FiredAt    time.Time
-	ResolvedAt time.Time
-	LastSentAt time.Time
-	ValidUntil time.Time
+	ActiveAt    time.Time
+	FiredAt     time.Time
+	ResolvedAt  time.Time
+	LastSentAt  time.Time
+	ValidUntil  time.Time
+	EvaluatedAt time.Time
 }
 
 func (a *Alert) needsSending(ts time.Time, resendDelay time.Duration) bool {
@@ -348,6 +349,7 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc, 
 		if alert, ok := r.active[h]; ok && alert.State != StateInactive {
 			alert.Value = smpl.V
 			alert.Annotations = annotations
+			alert.EvaluatedAt = ts
 			continue
 		}
 
@@ -355,6 +357,7 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc, 
 			Labels:      lbs,
 			Annotations: annotations,
 			ActiveAt:    ts,
+			EvaluatedAt: ts,
 			State:       StatePending,
 			Value:       smpl.V,
 		}
--- a/rules/alerting.go
+++ b/rules/alerting.go
@@ -85,12 +85,11 @@ type Alert struct {
 	Value float64
 	// The interval during which the condition of this alert held true.
 	// ResolvedAt will be 0 to indicate a still active alert.
-	ActiveAt    time.Time
-	FiredAt     time.Time
-	ResolvedAt  time.Time
-	LastSentAt  time.Time
-	ValidUntil  time.Time
-	EvaluatedAt time.Time
+	ActiveAt   time.Time
+	FiredAt    time.Time
+	ResolvedAt time.Time
+	LastSentAt time.Time
+	ValidUntil time.Time
 }
 
 func (a *Alert) needsSending(ts time.Time, resendDelay time.Duration) bool {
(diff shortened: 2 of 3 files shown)


simonpasquier:
Thanks for spotting this! Tracking it down, the culprit is the [json-iterator](github.com/json-iterator/go) library which doesn't produce 100% compliant data. Another [option](https://github.com/capnproto/capnproto/pull/262) would be to encode such values as `null` but I like your proposal better as it is consistent with the query API endpoints.

In general the v1 API is considered stable but we have a disclaimer in the documentation for [rules](https://prometheus.io/docs/prometheus/latest/querying/api/#rules) and [alerts](https://prometheus.io/docs/prometheus/latest/querying/api/#alerts).

cc @mxinden @brian-brazil

brian-brazil:
What does the standard Go JSON library do with these?

simonpasquier:
It returns an error.

brian-brazil:
Okay, using the same approach we do for the query apis seems best then. I think the code can be cleaner, we should do whatever we do in the query api.

dtrejod:
Willing to rework the PR if need be. It looks like the other Marshalling functions used by the `query` api are under https://github.com/prometheus/prometheus/blob/master/promql/value.go

Would adding a reusable `Value` definition there be cleaner?

brian-brazil:
A Scalar should be enough here.

dtrejod:
Latest commit uses Scalars. Apologies for the mess of commits as I pushed the wrong branch originally. 

Let me know if all looks good.

docs/querying/api.md
```diff
@@ -450,7 +450,10 @@ $ curl http://localhost:9090/api/v1/rules
                                     "severity": "page"
                                 },
                                 "state": "firing",
-                                "value": 1
+                                "value": [
```
brian-brazil: That's not right, it should be the value as a string - not a list with a timestamp.

dtrejod: This is using the Scalar type as suggested which is a timestamp + value. If we are okay with modifying the API to return a different value type I think this approach is the best solution as it mimics what the query API currently returns.

brian-brazil: We're technically allowed to make breaking changes here, however if we wanted the timestamp I think we'd put it in a different field.

dtrejod: Why not keep is the same as the query API? 

```
$ curl -s  -g 'http://$(hostname)/prometheus/api/v1/query?query=count(up)/0'|jq .
{
  "status": "success",
  "data": {
    "resultType": "vector",
    "result": [
      {
        "metric": {},
        "value": [
          1551461908.309,
          "+Inf"
        ]
      }
    ]
  }
}
```

brian-brazil: That'd break existing users unnecessarily.

dtrejod: Okay I understand. I thought we were moving forward with breaking the API given the clause. 

Also, sorry I've been busy with other work. Hopefully I can revisit sometime next week. If someone wants to implement this before me, feel free.

simonpasquier:
@dtrejod are you willing to carry on the PR?

brian-brazil:
Superseded by #5582

PR #5304, opened 2019-03-04
Add honor_timestamps

### LINKED ISSUE #5302: Discard timestamps at scrape time
## Proposal

When Prometheus is federating another monitoring system, it could be that that monitoring system is exposing timestamps.
In the case of federation with timestamps, Prometheus will not mark the metric as stale if it disappear. Instead, it will be present up to the query.loopback value.

Considering that Open Metrics will not help
https://github.com/OpenObservability/OpenMetrics/issues/37

Could we have a *job option* honor_timestamps to discard timestamps? That way power user can deal with broken third parties.

roidelapluie:
Fixes #5302

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -82,9 +82,10 @@ var (
 	DefaultScrapeConfig = ScrapeConfig{
 		// ScrapeTimeout and ScrapeInterval default to the
 		// configured globals.
-		MetricsPath: "/metrics",
-		Scheme:      "http",
-		HonorLabels: false,
+		MetricsPath:     "/metrics",
+		Scheme:          "http",
+		HonorLabels:     false,
+		HonorTimestamps: true,
 	}
 
 	// DefaultAlertmanagerConfig is the default alertmanager configuration.
@@ -334,6 +335,8 @@ type ScrapeConfig struct {
 	JobName string `yaml:"job_name"`
 	// Indicator whether the scraped metrics should remain unmodified.
 	HonorLabels bool `yaml:"honor_labels,omitempty"`
+	// Indicator whether the scraped timestamps should be respected.
+	HonorTimestamps bool `yaml:"honor_timestamps"`
 	// A set of query parameters with which the target is scraped.
 	Params url.Values `yaml:"params,omitempty"`
 	// How frequently to scrape the targets of this scrape config.
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -121,9 +121,10 @@ var expectedConf = &Config{
 		{
 			JobName: "prometheus",
 
-			HonorLabels:    true,
-			ScrapeInterval: model.Duration(15 * time.Second),
-			ScrapeTimeout:  DefaultGlobalConfig.ScrapeTimeout,
+			HonorLabels:     true,
+			HonorTimestamps: true,
+			ScrapeInterval:  model.Duration(15 * time.Second),
+			ScrapeTimeout:   DefaultGlobalConfig.ScrapeTimeout,
 
 			MetricsPath: DefaultScrapeConfig.MetricsPath,
 			Scheme:      DefaultScrapeConfig.Scheme,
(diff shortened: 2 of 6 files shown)


roidelapluie:
missing: doc and tests

roidelapluie:
Added doc and tests. Ready for review.

config/config.go
```diff
@@ -604,3 +609,8 @@ func (c *RemoteReadConfig) UnmarshalYAML(unmarshal func(interface{}) error) erro
 	// Thus we just do its validation here.
 	return c.HTTPClientConfig.Validate()
 }
+
+// IsZero is used to display false boolean values in the prometheus config page, as true is the default.
```
brian-brazil: What's this for?

docs/configuration/configuration.md
```diff
@@ -135,6 +135,20 @@ job_name: <job_name>
...
+# the target, providing the ability to discard timestamps. While timestamps
+# are most of the time desirable, some edge cases might require you to discard
+# the timestamps.
+#
+# If honor_timestamps is set to "true", the timestamps of the metrics exposed
+# by the target will be used.
+#
+# If honor_timestamps is set to "false", the timestamp of all the metrics will
+# be the timestamp of the scrape. As a consequence, prometheus staleness rules
+# will apply on all the metrics. Prometheus will not discard old metrics as long
```
brian-brazil: Retention has nothing to do with ingestion, this is confusing.

scrape/scrape.go
```diff
@@ -189,7 +189,7 @@ func newScrapePool(cfg *config.ScrapeConfig, app Appendable, logger log.Logger)
 		loops:         map[uint64]loop{},
 		logger:        logger,
 	}
-	sp.newLoop = func(t *Target, s scraper, limit int, honor bool, mrc []*relabel.Config) loop {
+	sp.newLoop = func(t *Target, s scraper, limit int, honorLabels, honorTimestamps bool, mrc []*relabel.Config) loop {
```
brian-brazil: I think we're at the point where we should have a struct for all these options.

config/config.go
roidelapluie: This used so that when you look the configuration in the webui you only see the honor_timestamps: false. Without this you see only the honor_timestamps: true

brian-brazil: You can remove the omitempty to show a false.

scrape/scrape.go
roidelapluie: Right. I will do it in two steps: one pr with the sctuct only and I will rebase this PR on top of that one.

config/config.go
roidelapluie: I want to hide honor_timestamps: true.

My preference would have been to hide it, as we do for honor_label, because it is not globally overridable and there is a default value.

I would also like to hide 

  metrics_path: /metrics
  scheme: http

in a future pr.

It will improve readability of the config page.

brian-brazil: This would be inconsistent with how we render things elsewhere, and the YAML library we're using plans to add the ability to keep comments which will make this all much nicer in future.

roidelapluie: ack, thanks

roidelapluie:
Pushed an update.

brian-brazil:
You've conflicts now

docs/configuration/configuration.md
```diff
@@ -135,6 +135,18 @@ job_name: <job_name>
 # when a time series does not have a given label yet and are ignored otherwise.
 [ honor_labels: <boolean> | default = false ]
 
+# honor_timestamps controls how prometheus respects the timestamps exposed by
+# the target, providing the ability to discard timestamps. While timestamps
+# are most of the time desirable, some edge cases might require you to discard
```
brian-brazil: This is misleading, the timestamps on exposition are generally not desirable.

roidelapluie: I rephrased.

config/config_test.go
```diff
@@ -121,7 +121,9 @@ var expectedConf = &Config{
 		{
 			JobName: "prometheus",
 
-			HonorLabels:    true,
+			HonorLabels:     true,
+			HonorTimestamps: true,
+
```
brian-brazil: I don't think we need all this new vertical whitespace.

docs/configuration/configuration.md
```diff
@@ -135,6 +135,16 @@ job_name: <job_name>
 # when a time series does not have a given label yet and are ignored otherwise.
 [ honor_labels: <boolean> | default = false ]
 
+# honor_timestamps controls whether prometheus respects the timestamps present
```
brian-brazil: Prometheus

brian-brazil:
Thanks!

PR #5309, opened 2019-03-05
Add debounce to input key handling

### LINKED ISSUE #5308: Master UI is unusably slow
Running a very recent build (from yesterday), editing text in the query/graph UI is unusably slow.

Try putting this query in:
```
  sum by (pod_name, namespace) (container_memory_rss) 
/ 
  sum by (pod_name, namespace) (
    label_replace(
      kube_pod_container_resource_requests_memory_bytes, 
      "pod_name", "$1", "pod", "(.*)"
    )
  ) 
```
And adding some text after `"(.*)"`.

Heres a profile:

screenshot 2019-03-05 at 14 10 01

Took ~2s to handle the keypress.  I don't know whats going on though.  @davkal ?

davkal:
- input key handler causes 2 layout cycles on each keypress which can
clog up browser rendering when typing quickly
- this change adds a debounce to the key press handler of 500ms

Fixes #5308

### DIFF
--- a/web/ui/static/js/graph/index.js
+++ b/web/ui/static/js/graph/index.js
@@ -1,6 +1,7 @@
 var Prometheus = Prometheus || {};
 var graphTemplate;
 
+var INPUT_DEBOUNCE_WAIT = 500; // ms
 var SECOND = 1000;
 
 /**
@@ -75,7 +76,9 @@ Prometheus.Graph.prototype.initialize = function() {
       self.queryForm.submit();
       e.preventDefault();
     }
-    $(this).on('keyup input', function() { resizeTextarea(this); });
+    $(this).on('keyup input', debounce(function() {
+      resizeTextarea(this);
+    }, INPUT_DEBOUNCE_WAIT));
   });
 
   self.expr.click(function(e) {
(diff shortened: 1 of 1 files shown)


tomwilkie:
Was this caused by https://github.com/prometheus/prometheus/pull/5201?

davkal:
> Was this caused by #5201?

Don't think so. The key input handler was added before that PR.

davkal:
Just took another look at the code. The debounce was not needed after all. `height` was being set twice, causing a two layout cycles each time a key was pressed. I've removed the setting to `auto`, so now it just sets to usually the same value, which is idempotent and wont cause a layout cycle.

juliusv:
@davkal With this change, the expression input doesn't shrink again when removing lines. It did before. Is it possible to keep that behavior?

davkal:
Can confirm, removing the `auto` setter breaks shrinking. I'll add a comment to the code to make clear what its purpose is. I'll bring back the debounce to fix the main issue.

davkal:
Reinstated the original fix. PTAL

juliusv:
@davkal Interesting, I can't reproduce the original issue #5308 at all locally (as in, I entered @tomwilkie's example expression and started typing around in it, no lag). Is it that browser-version-dependent? Just want to make sure that @tomwilkie says this fixes the lag for him :)

codesome:
I tried on my local machine and was able to reproduce. Steps:
1. Paste the query in the textarea and expand it.
2. Move the cursor outside the textarea and start typing

And this PR seems to fix it. I would like to merge this if it is working fine and go ahead with 2.8.

juliusv:
Interesting, doesn't happen for me, but good that it's confirmed!

:+1:

davkal:
For posterity: I was testing on Mac 10.14 with Chrome 72.

juliusv:
Chrome 72.0.3626.109 on Ubuntu Linux 18.04 here.

Issue #5242, opened 2019-02-20
promtool check metrics does not validate uniqueness


carlpett:
## Bug Report
**What did you do?**
`curl -s ... | promtool check metrics`

**What did you expect to see?**
According to the specs:
>Each line must have a unique combination of a metric name and labels. Otherwise, the ingestion behavior is undefined.

We had an endpoint that served duplicates for some name/label sets, but promtool does not complain.

**What did you see instead? Under which circumstances?**
Nothing

**Environment**
promtool from `master`/c8a1a5a9

An independent repro without our faulty exporter is 
```cat <<EOF | promtool check metrics
# HELP foo testing
# TYPE counter
foo{bar="1"} 1
foo{bar="1"} 2
EOF
```

* System information:
Darwin 18.2.0 x86_64

* Prometheus version:
git c8a1a5a9

geekodour:
will send the fix if I can reproduce the issue on master.

beorn7:
Hello from the bug scrub. Still trying to rally help for the open PR #5312 to fix this.

PR #5312, opened 2019-03-06
Added lintDuplicateMetric to promtool

geekodour:
see #5242

### DIFF
--- a/util/promlint/promlint.go
+++ b/util/promlint/promlint.go
@@ -17,6 +17,7 @@ package promlint
 import (
 	"fmt"
 	"io"
+	"reflect"
 	"sort"
 	"strings"
 
@@ -101,6 +102,7 @@ func lint(mf dto.MetricFamily) []Problem {
 		lintMetricUnits,
 		lintCounter,
 		lintHistogramSummaryReserved,
+		lintNotUniqueMetric,
 	}
 
 	var problems []Problem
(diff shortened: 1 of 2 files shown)


util/promlint/promlint.go
```diff
@@ -101,6 +102,7 @@ func lint(mf dto.MetricFamily) []Problem {
 		lintMetricUnits,
 		lintCounter,
 		lintHistogramSummaryReserved,
+		lintNotUniqueMetric,
```
tariq1890:
```suggestion
		lintDuplicateMetric,
```

gouthamve:
Hey, we do want this, but we should check for the timestamps as well and we should make sure that the timestamps are in ascending order (if there are timestamps).

Sorry for waiting so long to get back to this. Do you want to pick this up again?

geekodour:
@gouthamve sure, I'll update it. :) 

Just to confirm this is the format for timestamped metrics right? (Since I've not come across any timestamped metric yet :p)

```
a_total{foo="4"} 1 12345678901234567890.1234567890
```

brian-brazil:
It's a integer number of milliseconds.

bwplotka:
Hello from the Bug Scrub!

Sounds like this needs rebase and @gouthamve comment needs to be addressed https://github.com/prometheus/prometheus/pull/5312#issuecomment-668057564

@bboreham was nice enough to mention he will try to complete this work 💪🏽

beorn7:
Hello from the bug scrub once again.

@bboreham are you still going to work on this?

bboreham:
The code has since moved to a different repo, so I created https://github.com/prometheus/client_golang/pull/1472.

@gouthamve's comment is suggesting a completely different check, so I didn't do that.

I believe there is nothing left to do in this PR; if the new PR is merged, promtool will get updated shortly after.

PR #5313, opened 2019-03-07
[Consul] Add health label to metrics

ShimmerGlass:
Label metrics with the target health using consul's /health endpoint.

### DIFF
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -51,6 +51,8 @@ const (
 	tagsLabel = model.MetaLabelPrefix + "consul_tags"
 	// serviceLabel is the name of the label containing the service name.
 	serviceLabel = model.MetaLabelPrefix + "consul_service"
+	// healthLabel is the name of the label containing the health of the service instance
+	healthLabel = model.MetaLabelPrefix + "consul_health"
 	// serviceAddressLabel is the name of the label containing the (optional) service address.
 	serviceAddressLabel = model.MetaLabelPrefix + "consul_service_address"
 	//servicePortLabel is the name of the label containing the service port.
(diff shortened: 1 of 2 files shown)


discovery/consul/consul.go
```diff
@@ -100,6 +102,10 @@ type SDConfig struct {
 	Username     string             `yaml:"username,omitempty"`
 	Password     config_util.Secret `yaml:"password,omitempty"`
 
+	// Only scrap services nodes this healthy, inclusive. Possible values are "critical", "warning",
+	// "passning".
+	// Defaults to "citical" (all services nodes) if empty.
+	MinHealth string `yaml:"min_health,omitempty"`
```
brian-brazil: This can be done with relabelling, and it is recommended to scrape nodes that are apparently unhealthy.

Aestek:
Removed the `min_health` parameter. This change now only labels the metrics with the target health.

discovery/consul/consul.go
```diff
@@ -481,48 +484,48 @@ func (srv *consulService) watch(ctx context.Context, ch chan<- []*targetgroup.Gr
...
 		// Add all key/value pairs from the service's metadata as their own labels.
-		for k, v := range node.ServiceMeta {
+		for k, v := range serviceNode.Service.Meta {
 			name := strutil.SanitizeLabelName(k)
 			labels[serviceMetaDataLabel+model.LabelName(name)] = model.LabelValue(v)
 		}
 
 		// Add all key/value pairs from the service's tagged addresses as their own labels.
-		for k, v := range node.TaggedAddresses {
+		for k, v := range serviceNode.Node.TaggedAddresses {
```
brian-brazil: This doesn't seem to exist for the Health call

Aestek: Hmm, it does : https://www.consul.io/api/health.html#sample-response-2

brian-brazil: Ah, I was looking in the Service rather than the Node

ShimmerGlass:
Any update on this ?

brian-brazil:
Thanks!

PR #5315, opened 2019-03-08
ui: Hide default scheme and metrics path

roidelapluie:


brian-brazil:
I think this would be unexpected, and this approach is also in general not scalable. We've a lot of default properties across our various repos.

PR #5316, opened 2019-03-08
Naive string iterning for labels & values in the remote_write path.

### LINKED ISSUE #5364: v2.8.0 OOM with remote write
2.8.0 version OOM always， update from 2.7.2

tomwilkie:
~Currently no way to "forget" a string once it has been interned.  Open to suggestions...~ 

We went with ref counting the strings.

Fixes #5364


brian-brazil:
This doesn't seem right, that's already a string. Using the labels package rather than the one from model will likely help here.

tomwilkie:
Actually doesn't make that much of a difference:

screenshot 2019-03-08 at 15 07 58

brian-brazil:
Which, this PR or switching from model? We should be switching from model in any case.

tomwilkie:
> This doesn't seem right, that's already a string. Using the labels package rather than the one from model will likely help here.

Not sure I quite follow, but its `model` here because its the output of relabelling.

tomwilkie:
> Which, this PR or switching from model?

The PR.

brian-brazil:
If you use pkg/relabel that'll all line up, and save a little bit of CPU at least.

tomwilkie:
Ah perfect, I hadn't seen that.  Thanks!

tomwilkie:
I've half done this, @cstyan is going to pick it up now.

brian-brazil:
You could do reference counts, or more simply every time you see a checkpoint get rid of every label it has which is more expensive memory wise but not by much.

I don't think we should release a version with a known memory leak.

storage/remote/itern.go
```diff
@@ -0,0 +1,65 @@
...
+	p.mtx.RLock()
+	interned, ok := p.pool[s]
+	p.mtx.RUnlock()
+	if ok {
+		return interned
+	}
+
+	p.mtx.Lock()
+	defer p.mtx.Unlock()
+	if interned, ok := p.pool[s]; ok {
```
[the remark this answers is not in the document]
tomwilkie: This is the standard read-write mutex pattern; the first check is under the read lock and therefore concurrent, but you still need a second check it under write lock just in case someone got in there first and added the string between you dropping the read lock and taking the write lock.

cstyan:
We've leaked at least 100mb of memory over 3 days: 
2019-03-11-184932_2424x913_scrot

Going to run a version with the latest commit (adding ref counting and removing things with 0 refs from the interning map) for a similar amount of time to see if it helps.

storage/remote/read_test.go
```diff
@@ -33,13 +33,16 @@ func mustNewLabelMatcher(mt labels.MatchType, name, val string) *labels.Matcher
 	return m
 }
 
+/*
```
tomwilkie: Need to fix this!

storage/remote/queue_manager.go
```diff
@@ -343,8 +343,12 @@ func (t *QueueManager) StoreSeries(series []tsdb.RefSeries, index int) {
 	t.seriesMtx.Lock()
 	defer t.seriesMtx.Unlock()
 	for ref, labels := range temp {
-		t.seriesLabels[ref] = labels
 		t.seriesSegmentIndexes[ref] = index
+
+		if orig, ok := t.seriesLabels[ref]; ok {
```
brian-brazil: This takes a bit to understand. A comment would help.

config/config.go
```diff
@@ -298,6 +299,12 @@ func (c *GlobalConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		return err
 	}
 
+	for _, l := range gc.ExternalLabels {
```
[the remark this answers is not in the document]
cstyan: Not sure what you're asking for a test of? We already have `testdata/labelname.bad.yml` and `testdata/conf.good.yml` which cover unmarshal'ing of `external_labels`. I can add a test to check if the label value is valid.

storage/remote/read_test.go
cstyan: looks like getting all of the read code to use `labels.Labels` is going to be a bit of work, can we just turn external labels back into `model.LabelSet` for now and do that in another PR?

config/config.go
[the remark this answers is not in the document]
brian-brazil: Ah right, this was existing logic.

storage/remote/intern.go
```diff
@@ -0,0 +1,99 @@
...
+	p.mtx.Lock()
+	defer p.mtx.Unlock()
+	if atomic.LoadInt64(&interned.refs) != 0 {
+		return
+	}
+	delete(p.pool, s)
+}
+
+// StrPack returns a new instance of s which is tightly packed in memory.
+// It is intended for avoiding the situation where having a live reference
```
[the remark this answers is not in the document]
tomwilkie: Yeah these are never subsetted, so we could remove this.  I'm in favour of belt and braces just in case we start using this code elsewhere (such as in the scrape parser), but don't feel strongly.  Want me to remove it?

storage/remote/queue_manager.go
tomwilkie: Done.

tomwilkie:
> What happens when the external_labels/write_relabel_configs change?

Good catch.  When any config changes we stop/start all the queues, so we need to release the labels on stop.  Done.

storage/remote/read_test.go
tomwilkie: I updated it to use `labels.Labels`;  we're trying to move away from `model`.

tomwilkie:
One test failing, then we should get good to go.

storage/remote/intern.go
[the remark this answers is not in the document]
brian-brazil: The scrape parser has it's own thing already. I'd remove it for simplicity.

storage/remote/intern.go
```diff
@@ -92,7 +92,7 @@ func (p *pool) release(s string) {
 
 // StrPack returns a new instance of s which is tightly packed in memory.
 // It is intended for avoiding the situation where having a live reference
-// to a string slice over an unreferenced biger underlying string keeps the biger one
+// to a string slice over an unreferenced bigger underlying string keeps the biger one
```
brian-brazil: bigger :)

tomwilkie: BIGGER

tomwilkie:
```
______ _____ _____ _____  ___________ 
| ___ \_   _|  __ \  __ \|  ___| ___ \
| |_/ / | | | |  \/ |  \/| |__ | |_/ /
| ___ \ | | | | __| | __ |  __||    / 
| |_/ /_| |_| |_\ \ |_\ \| |___| |\ \ 
\____/ \___/ \____/\____/\____/\_| \_|
```

tomwilkie:
```
▀█████████▄   ▄█     ▄██████▄     ▄██████▄     ▄████████    ▄████████ 
  ███    ███ ███    ███    ███   ███    ███   ███    ███   ███    ███ 
  ███    ███ ███▌   ███    █▀    ███    █▀    ███    █▀    ███    ███ 
 ▄███▄▄▄██▀  ███▌  ▄███         ▄███         ▄███▄▄▄      ▄███▄▄▄▄██▀ 
▀▀███▀▀▀██▄  ███▌ ▀▀███ ████▄  ▀▀███ ████▄  ▀▀███▀▀▀     ▀▀███▀▀▀▀▀   
  ███    ██▄ ███    ███    ███   ███    ███   ███    █▄  ▀███████████ 
  ███    ███ ███    ███    ███   ███    ███   ███    ███   ███    ███ 
▄█████████▀  █▀     ████████▀    ████████▀    ██████████   ███    ███ 
                                                           ███    ███ 
```

codesome:
I think this should go into release-2.8 branch and there should be 2.8.1 with this improvement. What do you think?

tomwilkie:
I was going to ask you the same question!  It's a little big & risky for a patch release IMO, but its a good improvement.

tomwilkie:
I was going to ask you the same question!  It's a little big & risky for a patch release IMO, but its a good improvement.

config/config.go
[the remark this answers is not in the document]
cstyan: We should still be checking if the value is valid though probably.

cstyan:
Memory usage looks more stable with ref counting, no leak over the past 24+ hours.

2019-03-13-222318_2422x522_scrot

tomwilkie:
@brian-brazil any more feedback?

config/config.go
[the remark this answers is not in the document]
brian-brazil: Yes, we should. I must have missed this one when I was doing all those.

cstyan:
@brian-brazil test with a config like this does not fail:
```
global:
  external_labels:
    name: \xff
```
doesn't matter if it's `\xff`, `'\xff'`, `"\xff"`, etc.

brian-brazil:
With a bit of research, you want `!!binary "/w=="`

cstyan:
@brian-brazil thanks, not great with yaml, wasn't able to find `!!binary` in my searching.

brian-brazil:
:+1:

storage/remote/intern_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	testutil.Equals(t, ok, true)
+	testutil.Assert(t, interned.refs == 1, fmt.Sprintf("expected refs to be 1 but it was %d", interned.refs))
+
+	interner.release(testString)
+	_, ok = interner.pool[testString]
+	testutil.Equals(t, ok, false)
+}
+
+func TestIntern_MultiRef_Concurrent(t *testing.T) {
```
cstyan: There's probably a better test for this case (multiple go routines releasing and interning the same string).

tomwilkie: I'm happy with this; we've had it running for ~2 weeks and not seen any issues.

PR #5318, opened 2019-03-08
ui: remove extra table tag in service discovery

roidelapluie:

### DIFF
--- a/web/ui/templates/service-discovery.html
+++ b/web/ui/templates/service-discovery.html
@@ -17,8 +17,7 @@
   <div class="container-fluid">
 
     <h1>Service Discovery</h1>
-    <div class="table-container">
-      <table class="table table-sm table-bordered table-striped table-hover">
+    <div>
       <ul>
         {{range $i, $job := .Index}}
           <li>


brian-brazil:
:+1: 

I'm surprised that didn't cause problems.

roidelapluie:
Here is the "problem" (space between federate_1 and federate_2) :
prob

roidelapluie:
(present in last stable too:  http://demo.robustperception.io:9090/service-discovery )

web/ui/templates/service-discovery.html
```diff
@@ -18,7 +18,6 @@
 
     <h1>Service Discovery</h1>
     <div class="table-container">
```
tijmenb: Do we still need to add this div/class? `table-container` seems only used to [add functionality and styles to tables](https://github.com/prometheus/prometheus/search?utf8=%E2%9C%93&q=table-container&type=).

roidelapluie: Indeed!

codesome:
Thanks!

PR #5319, opened 2019-03-08
ui: Remove time picker borders

roidelapluie:
2.7:
xxtn

2.8-rc0:
xxtn22

With this patch:
btn


brian-brazil:
@codesome Do you think this is worth including in 2.8?

codesome:
If https://github.com/prometheus/prometheus/pull/5322 calls for 2.8.0-rc.1, I think no harm in including this too.

codesome:
Thanks!

PR #5320, opened 2019-03-08
ui: Expand expression_select to 220px

roidelapluie:
- insert metric at cursor -  is not fully visible in 2.8

Prometheus 2.7 (chrome/firefox)
crome3
ff3

Prometheus 2.8-rc0
chrom1
ff1

With this patch
ff2
crome2


codesome:
If https://github.com/prometheus/prometheus/pull/5322 calls for 2.8.0-rc.1, we might as well include it in 2.8.0 as one of multiple UI fixes.

Issue #5321, opened 2019-03-08
retention shown in runtime page set to 0s


roidelapluie:
In Prometheus 2.8-rc0 the runtime page shows '0s' retention.

2.7.1:
![retention](https://user-images.githubusercontent.com/291750/54060292-7579ee80-41fc-11e9-9c74-d00667ba333d.png)

2.8-rc0:
![retention2](https://user-images.githubusercontent.com/291750/54060359-b6720300-41fc-11e9-83bc-fa316efd143f.png)

Bisect:
a3c41f425699565c880fb529af6d58cf0cb254d5 is the first bad commit
commit a3c41f425699565c880fb529af6d58cf0cb254d5
Author: Krasi Georgiev <krasi-georgiev@users.noreply.github.com>
Date:   Tue Feb 19 13:53:43 2019 +0200

    use the default time retention value only when no size retention is set (#5216)

    fixes https://github.com/prometheus/prometheus/issues/5213

    Now that we have time and size base retention time bases should not have a default value. A default is set only when both - time and size flags are not set.

    This change will not affect current installations that rely on the default time based value, and will avoid confusions when only the size retention is set and it is expected that the default time based setting would be no longer in place.

    Signed-off-by: Krasi Georgiev <kgeorgie@redhat.com>

:040000 040000 e9acc051e5fd1bfc7262fbd60e675f59976ffa3e 35e8966e63113b7bfc3a61432bbdab27d5a006f6 M      cmd

roidelapluie:
Not sure if that is a bug or not but it is definitively confusing. Maybe we should add to the "Runtime" page the actual retention values.

roidelapluie:
cc @krasi-georgiev

krasi-georgiev:
what flags did you use when starting Prometheus?

krasi-georgiev:
just replicated it, the retention is set to 15d as before,but the GUI shows incorrect info. Will fix soon. Thanks.

roidelapluie:
Since now the flag is optional, I think maybe we could have in runtime page: "TSDB retention: 15d"

roidelapluie:
The values printed out could be:

- 15d
- 15d or 12 GB
- 12GB

krasi-georgiev:
I opened a PR with the fix.

> Since now the flag is optional, I think maybe we could have in runtime page: "TSDB retention: 15d"

not sure what you mean by this comment.

roidelapluie:
Oh nevermind I had a more complex fix in mind

PR #5322, opened 2019-03-09
Display correct values for the retention in the flags web gui.

krasi-georgiev:
fixes https://github.com/prometheus/prometheus/issues/5321

### DIFF
--- a/web/ui/templates/status.html
+++ b/web/ui/templates/status.html
@@ -46,12 +46,8 @@
           <td>{{.GOGC}}</td>
         </tr>
         <tr>
-          <th>Storage Time Retention</th>
-          <td>{{.RetentionTime}}</td>
-        </tr>
-        <tr>
-          <th>Storage Size Retention</th>
-          <td>{{.RetentionSize}}</td>
+          <th>Storage Retention</th>
+          <td>{{.StorageRetention}}</td>
         </tr>
       </tbody>
     </table>
--- a/web/ui/templates/status.html
+++ b/web/ui/templates/status.html
@@ -45,6 +45,14 @@
           <th>GOGC</th>
           <td>{{.GOGC}}</td>
         </tr>
+        <tr>
+          <th>Storage Time Retention</th>
+          <td>{{.RetentionTime}}</td>
+        </tr>
+        <tr>
+          <th>Storage Size Retention</th>
+          <td>{{.RetentionSize}}</td>
+        </tr>
       </tbody>
     </table>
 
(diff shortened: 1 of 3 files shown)


codesome:
This goes in 2.8.0

roidelapluie:
This will change the value in the flags. But it is confusing: if I add a maxbytes retention, the retention flag will change even if I do not change the retention flag.


I would propose to move that information to the Runtime and Build info page, by creating two metrics:

tsdb_retention_bytes
And
tsdb_retention_seconds

What do you think?

That could go in 2.9 ofc.

krasi-georgiev:
> This will change the value in the flags. But it is confusing: if I add a maxbytes retention, the retention flag will change even if I do not change the retention flag.

the flag will change only when no time or size retention is set and the default of 15d is applied. The behaviour hasn't changed since 2.7

```
if cfg.tsdb.RetentionDuration == 0 && cfg.tsdb.MaxBytes == 0 {
```



```
I would propose to move that information to the Runtime and Build info page, by creating two metrics:

tsdb_retention_bytes
And
tsdb_retention_seconds

What do you think?
```

I don't see why would we move it under Runtime and Build. 
These are flags and looks logical to stay under the /flags web page.

brian-brazil:
I'd expect flags to be what the user passed in, and runtime to show current status (i.e. what actually applies).

krasi-georgiev:
> I'd expect flags to be what the user passed in, and runtime to show current status (i.e. what actually applies).

aaah yeah that is a good way to think about it. otherwise if you miss the logs it is weird to see 15d of retention when you didn't pass anything as a flag.

I like the idea and will look into it.

roidelapluie:
Yes @bbrazil explained exactly what I had in mind.

krasi-georgiev:
updated!

now the /flags page will show exactly what was passed where the /status will show what is the actual retention values.

Flags page
Screenshot_20190311_104324

Status Page
Screenshot_20190311_104254

web/ui/templates/status.html
```diff
@@ -45,6 +45,14 @@ <h2 id="runtime">Runtime Information</h2>
...
           <td>{{.GOGC}}</td>
         </tr>
+        <tr>
+          <th>Storage Time Retention</th>
+          <td>{{.RetentionTime}}</td>
+        </tr>
+        <tr>
+          <th>Storage Size Retention</th>
+          <td>{{.RetentionSize}}</td>
+        </tr>
```
roidelapluie: Can we simplify this and have just one line?

(exepmle values that would be. printed: 
- 15d
- 15d or 10GB
- 19GB
)

krasi-georgiev: yep , not a bad idea. I updated the code.

krasi-georgiev: Screenshot_20190311_110834
Screenshot_20190311_110908

PR #5324, opened 2019-03-09
Add GODEBUG to the runtime page

roidelapluie:
This is useful, e.g. with go 1.12 to see if TLS1.3 has been enabled.


roidelapluie:
Note: Maybe this (and GOGC) should be hidden when empty?

brian-brazil:
I'd say it's better to always be present. Makes it clearer what's going on.

roidelapluie:
rebased

brian-brazil:
Thanks!

PR #5328, opened 2019-03-10
Prefix tsdb metrics with prometheus_

roidelapluie:

### DIFF
--- a/storage/tsdb/tsdb.go
+++ b/storage/tsdb/tsdb.go
@@ -188,7 +188,8 @@ func Open(path string, l log.Logger, r prometheus.Registerer, opts *Options) (*t
 		}
 	}
 
-	db, err := tsdb.Open(path, l, r, &tsdb.Options{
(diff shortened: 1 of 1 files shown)


roidelapluie:
See https://github.com/prometheus/tsdb/pull/546

storage/tsdb/tsdb.go
```diff
@@ -188,7 +188,8 @@ func Open(path string, l log.Logger, r prometheus.Registerer, opts *Options) (*t
 		}
 	}
 
-	db, err := tsdb.Open(path, l, r, &tsdb.Options{
+	pr := prometheus.WrapRegistererWithPrefix("prometheus_", r)
```
brian-brazil: We already discussed this, this is an inappropriate use of this feature. Metric names from a library should not change depending on what binary they happen to be in.

roidelapluie: ok

PR #5329, opened 2019-03-10
Exponentation operator to drop metric name in result of op operation

### LINKED ISSUE #5327: Exponentation (`^`) operator does not drop metric name
See http://demo.robustperception.io:9090/graph?g0.range_input=1h&g0.expr=up%20%5E%202&g0.tab=1

Like any other binary arithmetic operator, `^` should drop the metric name.

Reason is this line: https://github.com/prometheus/prometheus/blob/d390497280d61918aa8792f87a2f458e5e231ec2/promql/engine.go#L1896

infoverload:
Fixes #5327

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -1893,7 +1893,7 @@ func btos(b bool) float64 {
 // result of the op operation.
 func shouldDropMetricName(op ItemType) bool {
 	switch op {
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thanks!

PR #5335, opened 2019-03-11
.travis.yml: download modules in advance

simonpasquier:
cc @codesome

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,5 +8,6 @@ go:
 go_import_path: github.com/prometheus/prometheus
 
 script:
+- GO111MODULE=on travis_retry go mod download
 - make check_license style unused test staticcheck check_assets
 - git diff --exit-code


.travis.yml
```diff
@@ -8,5 +8,6 @@ go:
 go_import_path: github.com/prometheus/prometheus
 
 script:
+- GO111MODULE=on travis_retry go mod download
```
tariq1890: Is this for warming the `go mod cache` so as to allow for more  consistent `go mod` operations in Travis CI ?

codesome: Yes, currently travis is failing for `go mod tidy`.

tariq1890: Ah! Thanks for the clarification. This would come in handy in other repos as well.

codesome: `travis_retry` retries if downloading of modules failed, that's it. Because we don't retry on `go mod tidy` in our scripts, this is the travis way to do it.

simonpasquier:
I'll send a cleaner patch to follow up so it can be integrated in other repositories.

PR #5336, opened 2019-03-11
cmd/promtool: resolve relative paths in alert test files

dsymonds:
Like `promtool check config <path/to/foo.yaml>`, which resolves relative
paths inside foo.yaml to be relative to `path/to`, this now makes
`promtool test rules <path/to/test.yaml>` do the same thing.

Signed-off-by: David Symonds <dsymonds@gmail.com>

### DIFF
--- a/cmd/promtool/unittest.go
+++ b/cmd/promtool/unittest.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"reflect"
 	"sort"
 	"strconv"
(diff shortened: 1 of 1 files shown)


codesome:
You can already use relative path for rule files.

codesome:
:+1:

dsymonds:
@juliusv It'd be nice to have this in 2.8 if possible. Low risk, but a nice usability gain.

codesome:
@dsymonds we generally add only bug fixes to 2.8.x once 2.8.0 has been released. This would go in 2.9.

dsymonds:
Ah, I thought only 2.8.0rc1 had come out. I guess this can go in whenever then. Arguably it's a bug fix though, so maybe 2.8.1?

codesome:
I would not actually call it a bug (or a major bug which needs another release), also this behavior has been from the start, unfortunately. Having a behavioral change in a minor release might not be a good idea, adding this in 2.9 sounds better.

dsymonds:
So is there anything holding this back, @juliusv?

juliusv:
@dsymonds Nope, looks good!

PR #5339, opened 2019-03-12
Makefile.common: add deps target

simonpasquier:
Follow up of #5335 

Due to Travis CI having intermittent issues, we warm up the modules cache before running the tests. This will also be useful for prometheus/common and prometheus/tsdb that already have a similar target in their Makefile.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,12 @@ go:
 
 go_import_path: github.com/prometheus/prometheus
 
+# This ensures that the local cache is filled before running the CI.
+# travis_retry retries the command 3 times if it fails as we've experienced
+# random issues on Travis.
+before_install:
+- travis_retry make deps
+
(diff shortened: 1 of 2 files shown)


brian-brazil:
> LGTM, shouldn't this be unnecessary as we vendor everything?

Our libraries (common/client_golang) don't vendor. Otherwise this should not be needed.

simonpasquier:
`go mod tidy` verifies that `go.sum` and `go.mod` match with the remote repositories which is why it needs to download stuff from the Internet. `build` using `-mod=vendor` won't use the downloaded packages so I don't it's necessary to add a per-requisite there.

This is still needed for libraries (eg common, tsdb and client_golang) as they're not vendoring anything.

SuperQ:
Thanks @simonpasquier

PR #5346, opened 2019-03-12
fix redundant err!=nil check in web.go

tariq1890:
The err != nil check to retrieve the tsdb instance is unnecessary and can be reduced to one statement.


brian-brazil:
Thanks!

PR #5350, opened 2019-03-13
discovery/azure: make local virtualMachine struct more generic by removing the go sdk field reference

tariq1890:
Removing an anti-pattern where we directly inject a Go SDK struct reference into the `virtualMachine` struct when it's intended to be as generic as possible. This struct provides a wrapper around VMs and VM Scale sets.


tariq1890:
@brian-brazil @simonpasquier FYI

tariq1890:
Requesting your review @brian-brazil

PR #5358, opened 2019-03-14
Expose lexer item types

juliusv:
We have generally agreed to expose AST types / values that are necessary
to make sense of the AST outside of the promql package. Currently the
`UnaryExpr`, `BinaryExpr`, and `AggregateExpr` AST nodes store the lexer
item type to indicate the operator type, but since the individual item
types aren't exposed, an external user of the package cannot determine
the operator type. So this PR exposes them.

Although not all item types are required to make sense of the AST (some
are really only used in the lexer), I decided to expose them all here to
be somewhat more consistent. Another option would be to not use lexer
item types at all in AST nodes.

The concrete motivation is my work on the PromQL->Flux transpiler, but
this ought to be useful for other cases as well.

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -36,7 +36,6 @@ import (
 	"github.com/prometheus/prometheus/pkg/timestamp"
 	"github.com/prometheus/prometheus/pkg/value"
 	"github.com/prometheus/prometheus/storage"
-
 	"github.com/prometheus/prometheus/util/stats"
 )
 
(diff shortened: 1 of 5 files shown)


brian-brazil:
:+1: 

You've test failures.

juliusv:
Doh. Fixed.

mxinden:
Thanks @juliusv. I am looking into a [language server](https://microsoft.github.io/language-server-protocol/) implementation for PromQL right now. This should help.

juliusv:
@mxinden Oohhh, looking forward to that!

PR #5363, opened 2019-03-14
discovery/azure:optimize iteration logic for VMScalesets, VMScalesetVMs, and VMs

tariq1890:
This PR optimizes the iteration logic into something saner. There is unnecessary duplication.

### DIFF
--- a/discovery/azure/azure.go
+++ b/discovery/azure/azure.go
@@ -420,22 +420,15 @@ func (client *azureClient) getVMs(ctx context.Context) ([]virtualMachine, error)
 	var vms []virtualMachine
 	result, err := client.vm.ListAll(ctx)
 	if err != nil {
-		return nil, fmt.Errorf("could not list virtual machines: %s", err)
+		return nil, fmt.Errorf("could not list virtual machines: %s", err.Error())
 	}
-
-	for _, vm := range result.Values() {
-		vms = append(vms, mapFromVM(vm))
-	}
-
-	// If we still have results, keep going until we have no more.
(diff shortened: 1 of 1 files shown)


tariq1890:
This has been tested and it works as-is.

discovery/azure/azure.go
```diff
@@ -419,24 +419,16 @@ func (d *Discovery) refresh(ctx context.Context) (tg *targetgroup.Group, err err
 func (client *azureClient) getVMs(ctx context.Context) ([]virtualMachine, error) {
 	var vms []virtualMachine
 	result, err := client.vm.ListAll(ctx)
-	if err != nil {
```
simonpasquier: Swallowing the error returned by `ListAll()` doesn't feel very idiomatic IMO. I'd keep this first `if err != nil { ... }` block for readability. And it would also make the for loop more readable as the following error check can happen right after the `NextWithContext()` call.

```go
	for result.NotDone() {
		for _, vm := range result.Values() {
			vms = append(vms, mapFromVM(vm))
		}
		err = result.NextWithContext(ctx)
		if err != nil {
			return nil, fmt.Errorf("could not list virtual machines: %s", err)
		}
	}
```

tariq1890: Makes sense. I can do that

tariq1890:
@simonpasquier Done!

tariq1890:
@brian-brazil Pinging you for final approval.

brian-brazil:
Thanks!

PR #5366, opened 2019-03-15
scrape: fallback to hostname if lookup fails

simonpasquier:
Without it, I can't start Prometheus on my local machine...

### DIFF
--- a/scrape/manager.go
+++ b/scrape/manager.go
@@ -250,7 +250,7 @@ func getFqdn() (string, error) {
 
 	ips, err := net.LookupIP(hostname)
 	if err != nil {
-		return hostname, err
+		return hostname, nil
 	}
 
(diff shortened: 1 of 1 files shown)


scrape/manager.go
```diff
@@ -250,7 +250,7 @@ func getFqdn() (string, error) {
 
 	ips, err := net.LookupIP(hostname)
 	if err != nil {
-		return hostname, err
+		return hostname, nil
```
brian-brazil: Can you add a comment here?

simonpasquier: Sure, it was already mentioned in the function header comment FWIW.

brian-brazil: More that this looks wrong at first glance, and we don't want people accidentally undoing it.

Issue #5193, opened 2019-02-08
Rule evaluation failure after reload with newly added rules


jsravn:
## Bug Report
**What did you do?**

1. `docker run -p 9090:9090 prom/prometheus:v2.7.1`
2. Edit `/etc/prometheus/prometheus.yml`, enable the `- "first_rules.yml"` file.
3. Create `/etc/prometheus/first_rules.yml`:
```yaml
groups:       
- name: testgroup
  rules:      
  - record: test            
    expr: 22222         
    labels:             
      instance: my-second
```
4. Reload prometheus - `kill -SIGHUP 1`. Everything is fine.
5. Update `/etc/prometheus/first_rules.yml`:
```yaml
groups:       
- name: testgroup
  rules:      
  - record: test            
    expr: 11111         
    labels:             
      instance: my-first
  - record: test         
    expr: 22222          
    labels:              
      instance: my-second
```
6. Reload prometheus - `kill -SIGHUP 1`.
7. Wait for the next rule evaluation. An error is produced:
```
level=info ts=2019-02-08T10:28:36.906865554Z caller=main.go:695 msg="Loading configuration file" filename=/etc/prometheus/prometheus.yml
level=info ts=2019-02-08T10:28:36.907330758Z caller=main.go:722 msg="Completed loading of configuration file" filename=/etc/prometheus/prometheus.yml
level=warn ts=2019-02-08T10:28:43.395453805Z caller=manager.go:514 component="rule manager" group=testgroup msg="Error on ingesting results from rule evaluation with different value but same timestamp" numDropped=1
```
8. Data for the `instance: my-second` rule is missing at that timestamp.

**What did you expect to see?**

I expected to see no errors, and data for `instance: my-second` at the evaluation time. Although they have the same identifier `test`, they should be unique time series as they have different labels.

**What did you see instead? Under which circumstances?**

Log error, and missing data for the `instance: my-second` rule at the evaluation time.

**More details**

* This only happens when the recording rule identifier matches, and the new rule is prepended before the other ones.
* All the subsequent recording rules with the same identifier fail in the next rule evaluation. After that, they succeed.

**Environment**

docker container

* System information:

```
/prometheus $ uname -srm
Linux 4.20.6-arch1-1-ARCH x86_64
```

* Prometheus version:

```
/prometheus $ prometheus --version
prometheus, version 2.7.1 (branch: HEAD, revision: 62e591f928ddf6b3468308b7ac1de1c63aa7fcf3)
[...]

PR #5368, opened 2019-03-15
reload: copy state on both name and labels

jsravn:
Fix https://github.com/prometheus/prometheus/issues/5193

Using just name causes the linked issue - if new rules are inserted with
the same name (but different labels), the reordering will cause stale
markers to be inserted in the next eval for all shifted rules, despite
them not being stale.

Ideally we want to avoid stale markers for time series that still exist
in the new rules, with name and labels being the unique identifer.

This change adds labels to the internal map when copying the old rule
data to the new rule data. This prevents the problem of staling rules
that simply shifted order.

If labels change, it is a new time series and the old series will stale
regardless. So it should be safe to always match on name and labels when
copying state.

Can you have a look @fabxc?

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -189,6 +189,8 @@ func EngineQueryFunc(engine *promql.Engine, q storage.Queryable) QueryFunc {
 // interval and acted upon (currently either recorded or used for alerting).
 type Rule interface {
 	Name() string
+	// Labels of the rule.
+	Labels() labels.Labels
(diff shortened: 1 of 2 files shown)


brian-brazil:
The CopyState comment needs updating.

jsravn:
Okay, updated it.

brian-brazil:
Thanks! It doesn't solve the problem generally, but should help.

PR #5371, opened 2019-03-15
web/ui: Fix padding in navbar container

metalmatze:
Running `v2.8.0-rc.0` on my cluster I noticed the slightly off Prometheus heading in the navbar. That's just too annoying to leave it like that.
The navbar and the container inside of that both add a horiztonal padding of ~15px. This PR removes the inner padding of the container in this specific case.

Before:
prometheus_off

Afterwards:
prometheus_on

### DIFF
--- a/web/ui/static/css/prometheus.css
+++ b/web/ui/static/css/prometheus.css
@@ -24,3 +24,8 @@ body {
 .label {
   white-space: normal;
 }
+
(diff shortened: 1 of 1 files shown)


simonpasquier:
LGTM

juliusv:
Hehe :+1:

PR #5372, opened 2019-03-15
promparse: sort all labels when parsing

GiedriusS:
Some label names might start with an uppercase letter in which case it
needs to come before `__name__`. Without this, it means that we are not
maintaining the same invariant that all label names should be sorted in
such cases.

Amend the tests to check this problem automatically. Without the change
the tests do not pass.


brian-brazil:
Thanks!

jjneely:
We were hoping to see this in 2.8.1 to address the label ordering bugs we have with Thanos.  Is it possible to have this in a future 2.8 release, or does this need to wait for 2.9?

brian-brazil:
2.9 will start next week, so I'm not sure there's a point in putting out a bugfix for this.

Issue #5370, opened 2019-03-15
Display rules as defined in configuration


sylr:
## Proposal
**Use case. Why is this important?**

Prometheus rules can be quite complex thus the use of comments, new lines, uppercased keywords ... etc can help make it easier to understand.

E.G:

```
(
    # Pod is down and rename label "kubernetes_pod_name" to "pod"
    label_replace(up{job="kubernetes-pods"}, "pod", "$1", "kubernetes_pod_name", "(.*)")
    *
    # Add node label to up{}
    on (pod) group_left(node) avg(kube_pod_info) by (pod, node)
) == 0
AND
# Node is ready
on (node) kube_node_status_condition{condition="Ready",status="true"} == 1
```

Unfortunately the Prometheus UI does not display the rules as they were written but like if a "sanitizer" have been applied, i.e.:

```
(label_replace(up{job="kubernetes-pods"},
  "pod", "$1", "kubernetes_pod_name", "(.*)") * on(pod)
  group_left(node) avg by(pod, node) (kube_pod_info)) == 0 and on(node) kube_node_status_condition{condition="Ready",status="true"}
  == 1
)
```
The latter is really quite indigest so it would be nice if you could keep the formatting as it.

brian-brazil:
This is a limitation of the YAML parser we use, once it's better we can do better.

sylr:
The YAML parser seems capable to handle comments and multiline strings quite all right:

```go
package main

import (
	"fmt"

	yaml "gopkg.in/yaml.v2"
)

type T struct {
	A string `yaml:"a"`
}

var (
	conf = `a: |
  (
    # Pod is down and rename label "kubernetes_pod_name" to "pod"
    label_replace(up{job="kubernetes-pods"}, "pod", "$1", "kubernetes_pod_name", "(.*)")
    *
    # Add node label to up{}
    on (pod) group_left(node) avg(kube_pod_info) by (pod, node)
  ) == 0
  AND  # Node is ready
  on (node) kube_node_status_condition{condition="Ready",status="true"} == 1`
)

func main() {
	t := T{}
	yaml.UnmarshalStrict([]byte(conf), &t)
	fmt.Printf("%s\n-----------\n", conf)
	fmt.Printf("%s\n", t.A)
}
```

brian-brazil:
Ah, you're looking at the rules page. Those are what the PromQL printer produces. On a page like that it's best to have the actual AST, rather than what may be a confusingly written expression.

[...]

PR #5373, opened 2019-03-16
Display rules as defined in configuration

sylr:
Attempt at implementing https://github.com/prometheus/prometheus/issues/5370


brian-brazil:
Per #5370, this should display a canonical view of what Prometheus actually parsed - not whatever the user happened to enter.

Issue #5343, opened 2019-03-12
Allow reloading with non POST methods


eirc:
The `/-/reload` endpoint only works with the `POST` method. It would possibly make more sense to allow other methods too like `PUT`.

https://github.com/prometheus/prometheus/blob/7c8f0b08647f1ac5f9608069678b4a01c6e79866/web/web.go#L313

The reasoning is that `POST` by definition produces different side effects each time it is run (like create new resources under `REST`) compared to `PUT` which you can supposedly safely fire multiple times.

Technically my problem is when trying to reload prometheus with a python program and `urllib3` refuses to retry `POST` requests when there's a read error (specifically) since resending a `POST` when the server _may_ have already received is a no no although it's not really a problem in config reloading.

https://github.com/urllib3/urllib3/blob/2a0957ea27e966166f81f88693af8e2f87d19fb6/src/urllib3/util/retry.py#L97-L104

PS: I'm also suggesting to just add the `PUT` instead of replacing the `POST` just for compatibility reasons, I'm sure many would take issue with such an API change.

brian-brazil:
Longer term we plan on switching this to PUT for XSRF reasons, so supporting both now would make sense.

> Technically my problem is when trying to reload prometheus with a python program and urllib3 refuses to retry POST requests when there's a read error

Is that supported by the RFCs?

eirc:
TBH I have no idea what RFCs say about this but it does make sense IMHO.

bharaththiruveedula:
@brian-brazil can I work on this?

brian-brazil:
Sure

Danielv123:
With the PR merged, this issue should be considered for closing.

PR #5376, opened 2019-03-16
Support non POST methods for Lifecycle and Admin APIs

bharaththiruveedula:
Signed-off-by: Bharath Thiruveedula <bharath_ves@hotmail.com>

### DIFF
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -694,6 +694,7 @@ It will optionally skip snapshotting data that is only present in the head block
 
 ```
 POST /api/v1/admin/tsdb/snapshot?skip_head=<bool>
+PUT /api/v1/admin/tsdb/snapshot?skip_head=<bool>
 ```
 
 ```json
@@ -705,10 +706,9 @@ $ curl -XPOST http://localhost:9090/api/v1/admin/tsdb/snapshot
   }
 }
 ```
-
 The snapshot now exists at `<data-dir>/snapshots/20171210T211224Z-2be650b6d019eb54`
 
-*New in v2.1*
+*New in v2.1 and supports PUT from v2.9*
 
 ### Delete Series
 DeleteSeries deletes data for a selection of series in a time range. The actual data still exists on disk and is cleaned up in future compactions or can be explicitly cleaned up by hitting the Clean Tombstones endpoint.
@@ -717,6 +717,7 @@ If successful, a `204` is returned.
 
 ```
 POST /api/v1/admin/tsdb/delete_series
+PUT /api/v1/admin/tsdb/delete_series
 ```
 
 URL query parameters:
@@ -733,7 +734,7 @@ Example:
 $ curl -X POST \
   -g 'http://localhost:9090/api/v1/admin/tsdb/delete_series?match[]=up&match[]=process_start_time_seconds{job="prometheus"}'
 ```
-*New in v2.1*
+*New in v2.1 and supports PUT from v2.9*
 
 ### Clean Tombstones
 CleanTombstones removes the deleted data from disk and cleans up the existing tombstones. This can be used after deleting series to free up space.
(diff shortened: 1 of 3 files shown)


bharaththiruveedula:
Implements #5343

brian-brazil:
This should apply to the other admin endpoints too.

bharaththiruveedula:
@brian-brazil you mean  for /quit also?

brian-brazil:
Both.

prompb/rpc.pb.gw.go
```diff
@@ -45,6 +45,23 @@ func request_Admin_TSDBSnapshot_0(ctx context.Context, marshaler runtime.Marshal
 
 }
 
+var (
```
brian-brazil: There's no need to update all this, v1 is where it's at.

bharaththiruveedula:
> The docs should also be updated.

No Need to update v2?

prompb/rpc.pb.gw.go
bharaththiruveedula: No Need to update V2 apis?

brian-brazil: No need.

bharaththiruveedula:
> The docs should also be updated.
Can you point me which docs to be updated. Can I update accordingly this doc https://github.com/prometheus/prometheus/blob/master/docs/configuration/configuration.md

brian-brazil:
It's docs/querying/api.md

bharaththiruveedula:
> It's docs/querying/api.md

Done

docs/querying/api.md
```diff
@@ -733,6 +745,12 @@ Example:
 $ curl -X POST \
   -g 'http://localhost:9090/api/v1/admin/tsdb/delete_series?match[]=up&match[]=process_start_time_seconds{job="prometheus"}'
 ```
+
+```json
+$ curl -X PUT \
+  -g 'http://localhost:9090/api/v1/admin/tsdb/delete_series?match[]=up&match[]=process_start_time_seconds{job="prometheus"}'
+```
+
```
brian-brazil: You should add a note here that PUT was added in 2.9

docs/querying/api.md
```diff
@@ -733,6 +745,12 @@ Example:
 $ curl -X POST \
   -g 'http://localhost:9090/api/v1/admin/tsdb/delete_series?match[]=up&match[]=process_start_time_seconds{job="prometheus"}'
 ```
+
+```json
+$ curl -X PUT \
```
brian-brazil: You can have just the POST example

docs/querying/api.md
```diff
@@ -750,4 +769,45 @@ This takes no parameters or body.
 $ curl -XPOST http://localhost:9090/api/v1/admin/tsdb/clean_tombstones
 ```
 
+```json
+$ curl -XPUT http://localhost:9090/api/v1/admin/tsdb/clean_tombstones
+```
+
 *New in v2.1*
+
+## Prometheus Lifecycle APIs
```
brian-brazil: This page is scoped to /v1, which these aren't under. And apparently quit isn't documented anywhere. This seems like something to leave for another PR, as we'll need to find a space for these.

bharaththiruveedula: Okay. I will remove this section

docs/querying/api.md
bharaththiruveedula: I will remove PUT example

docs/querying/api.md
bharaththiruveedula: Done

docs/querying/api.md
```diff
@@ -750,4 +752,4 @@ This takes no parameters or body.
 $ curl -XPOST http://localhost:9090/api/v1/admin/tsdb/clean_tombstones
 ```
 
-*New in v2.1*
+*New in v2.1*
```
brian-brazil: The notes are generally put here (presuming we're following Python), so it'd be PUT is new in v2.9 or similar.

bharaththiruveedula: Fixed

brian-brazil:
Thanks!

PR #5377, opened 2019-03-17
Update vendoring

SuperQ:
* Update hashicorp.
* Update gophercloud.


go.mod
```diff
@@ -86,51 +68,43 @@ require (
...
 	gopkg.in/fsnotify.v1 v1.4.7 // indirect
 	gopkg.in/fsnotify/fsnotify.v1 v1.3.0
-	gopkg.in/inf.v0 v0.9.0 // indirect
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
-	gopkg.in/vmihailenco/msgpack.v2 v2.9.1 // indirect
 	gopkg.in/yaml.v2 v2.2.2
 	k8s.io/api v0.0.0-20181213150558-05914d821849
 	k8s.io/apimachinery v0.0.0-20181127025237-2b1284ed4c93
-	k8s.io/client-go v2.0.0-alpha.0.0.20181121191925-a47917edff34+incompatible
+	k8s.io/client-go v8.0.0+incompatible
```
tariq1890: i) Prometheus currently uses v10.0.0 of client-go (equivalent to k8s 1.13). I don't think we should be downgrading
ii) If we are to update client-go, api and apimachinery would need to be updated as well.

SuperQ: According to this diff, Prometheus is using `v2.0.0-alpha`, not v10.

roidelapluie:
This pull request is outdated and can be closed. Updates of dependencies has also been added to RELEASE.md, ensuring it happens on regular basis.

PR #5381, opened 2019-03-18
discovery: factorize for SD based on refresh

simonpasquier:
Many of the SD mechanisms refresh the targets at fixed interval. This PR factorizes the `Run()` loop across those mechanisms.

### DIFF
--- a/discovery/azure/azure.go
+++ b/discovery/azure/azure.go
@@ -29,7 +29,6 @@ import (
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
-	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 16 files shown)


discovery/azure/azure.go
```diff
@@ -126,60 +128,35 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	d := &Discovery{
+		cfg:    cfg,
+		port:   cfg.Port,
+		logger: logger,
+	}
+	d.Discovery = refresh.NewDiscovery(
+		logger,
+		time.Duration(cfg.RefreshInterval),
+		d.refresh,
+		refresh.WithDuration(refreshDuration),
```
brian-brazil: This is a smell, it'd be better to define it in the refresh package and use labels.

brian-brazil:
:+1:

PR #5382, opened 2019-03-18
scrape: stop scrape loops ASAP on config reload

simonpasquier:
The context passed to the scraper was derived from the scrape pool's
context instead of the scrape loop's context. On scrape configuration
change, the scrape pool stopped the loop by cancelling the scrape loop's
context but it didn't cancel its own context. The scraper could wait up
to the scrape timeout before returning and effectively stopping the
loop. In practice with unresponsive targets, a configuration reload
could take as long as the highest scrape timeout to complete.

This change wraps the scraper's context with the loop's context which
ensures that the scrape is canceled immediately irrespective of whether
it comes from the loop or the pool. We tentatively write the stale
markers when the loop is canceled but not on the pool's cancellation.

This change also introduces a new histogram metric
"prometheus_config_reload_duration_seconds" for tracking the
configuration reload duration.

I tested this change with a configuration including unreachable targets and a scrape interval & timeout of 30s. Before this PR:

```
$ time curl -XPOST localhost:9090/-/reload
real    0m30.240s
user    0m0.006s
sys     0m0.004s
```

After this PR:

```
$ time curl -XPOST localhost:9090/-/reload
real    0m0.030s
user    0m0.011s
sys     0m0.006s
```


cmd/prometheus/main.go
```diff
@@ -71,6 +71,11 @@ var (
 		Name: "prometheus_config_last_reload_success_timestamp_seconds",
 		Help: "Timestamp of the last successful configuration reload.",
 	})
+	configDuration = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Name:    "prometheus_config_reload_duration_seconds",
+		Help:    "Histogram of configuration reload durations.",
+		Buckets: []float64{.1, 1, 5, 10},
```
brian-brazil: The defaults are fine with these sort of values.

scrape/scrape.go
```diff
@@ -895,6 +898,12 @@ mainLoop:
 			}
 		}
 
+		select {
+		case <-sl.runCtx.Done():
```
brian-brazil: This needs a comment

scrape/helpers_test.go
```diff
@@ -46,25 +49,36 @@ func (a *collectResultAppender) AddFast(m labels.Labels, ref uint64, t int64, v
...
 	if a.next == nil {
 		return 0, nil
 	}
 	return a.next.Add(m, t, v)
 }
 
-func (a *collectResultAppender) Commit() error   { return nil }
-func (a *collectResultAppender) Rollback() error { return nil }
+func (a *collectResultAppender) samples() []sample {
+	a.mtx.Lock()
```
roidelapluie: Should that be a read lock?

simonpasquier: I've used a normal mutex as the code is only used in tests so we have no particular performance concern.

simonpasquier:
@brian-brazil what do you think of this PR?

brian-brazil:
This would still make scrapes fail due to a reload.

simonpasquier:
On the other hand, an unresponsive target blocks the reload process until it timeouts currently.

brian-brazil:
That's better than never getting a scrape completed if someone is reloading several times a minute.

roidelapluie:
We have looked at it during the bug scrub, and stopping scrapes should only happen during shutdown.

PR #5386, opened 2019-03-21
fix query with offset incorrect end time issue

garrettlish:
Now query with offset, promql engine will shift query start time to be `offset` ago, however, it doesn't shift query end time accordingly. It will cause query too long time range. Although later evaluation will drop the out of boundary metrics, it still brings too much pressure to tsdb.

For example: `metric offset 1w`, without the fix, it will query 1w+5m time range for vector and 1w+ time range for matrix.


garrettlish:
incorrect, will refine it

PR #5388, opened 2019-03-21
refine error handling in prometheus

tariq1890:
i) Uses the more idiomatic `Wrap` and `Wrapf` methods for creating nested errors.
ii) Fixes some incorrect usages of `fmt.Errorf` where the error messages don't have any formatting directives.
iii) Does away with the use of `fmt` package for errors in favour of pkg/errors


cmd/prometheus/main.go
```diff
@@ -41,10 +41,10 @@ import (
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/common/version"
 	prom_runtime "github.com/prometheus/prometheus/pkg/runtime"
-	"gopkg.in/alecthomas/kingpin.v2"
+	kingpin "gopkg.in/alecthomas/kingpin.v2"
```
tariq1890: These were added because of running `goimports`. Harmless change.

tariq1890:
@juliusvolz requesting your review on this

tariq1890:
@juliusv requesting your review on this

util/treecache/treecache.go
```diff
@@ -16,6 +16,7 @@ package treecache
 import (
 	"bytes"
 	"fmt"
+	"github.com/pkg/errors"
```
juliusv: We put stdlib and non-stdlib in separate import groups usually.

On the other hand, in many of the other files, there's no reason to add `github.com/pkg/errors` into its completely separate import group, it can go together with all the other non-stdlib packages.

template/template.go
```diff
@@ -317,7 +318,7 @@ func (te Expander) ExpandHTML(templateFiles []string) (result string, resultErr
 			var ok bool
 			resultErr, ok = r.(error)
 			if !ok {
-				resultErr = fmt.Errorf("panic expanding template %v: %v", te.name, r)
+				resultErr = errors.Wrapf(r.(error), "panic expanding template %v", te.name)
```
juliusv: This branch is reached if `r` couldn't be coerced to an `error` type, so doing `r.(error)` here would fail. In the case where it's not an `error`, we can just print it via `%v` I believe.

util/treecache/treecache.go
tariq1890: Goimports enforces that. Since it has a Github.com prefix, it is recognized as one needing a separate group.

template/template.go
tariq1890: Whoops thanks for catching that :)

util/treecache/treecache.go
tariq1890: I forgot to goimports that file. Looks like we could add goimports to prometheus' staticcheck job.

juliusv: > Since it has a Github.com prefix, it is recognized as one needing a separate group.

Hmm, but all our other non-stdlib packages are also grouped together just fine. My editor also has goimports, but it doesn't put every GH import into its separate section.

tariq1890: So I moved pkg/errors into the other import groups and reran `goimports`. After that, it didn't move it out into a separate group.

`goimports` is strange. I'll make the necessary corrections.

juliusv: Ah right, it does create a new group when adding a completely new import, but when something is already present and in a group it leaves it in that group.

tariq1890:
@juliusv Thank you very much for your review :). I have addressed the comments (comes with an added bonus of fixing imports that were already out of order before this PR :P)

Ready for another round of review.

tariq1890:
@beorn7 Requesting your review as well on this.

config/config.go
```diff
@@ -22,13 +22,14 @@ import (
 	"strings"
 	"time"
 
-	"github.com/prometheus/prometheus/pkg/labels"
-	"github.com/prometheus/prometheus/pkg/relabel"
-
+	"github.com/pkg/errors"
```
juliusv: still some separate groups?

juliusv:
Besides the last nits this looks good to me, but since it touches all of Prometheus I would like at least one other person to agree :) Maybe @brian-brazil or @brancz?

config/config.go
tariq1890: This isn't separate I think?

tariq1890: Or maybe I misunderstood.

juliusv: Oh sorry, misread the diff.

tariq1890: Ok. Resolving comment :)

beorn7:
I'm fine with the changes around errors (but I don't have the closest relationship to the codebase as a whole). 

However, this PR also changes grouping of imports a bit. Originally, we had one group for std imports, one for non-std imports and one for imports from `github.com/prometheus/prometheus/...` (possibly even separate groups for named imports). I guess this wasn't consistently followed over time. Perhaps we could quickly agree on one consistent way and at least do whatever this PR touches consistently?

tariq1890:
Sure I can create a separate group for the `github.com/prometheus/...` imports. That's definitely a  practice I've seen being followed. I can do either of the following

i) Implement consistent grouping of imports across the codebase in a followup and even have the linter recognize `github.com/prometheus/...` as local imports and needing to be separated from other named imports.

ii) Implement consistent grouping of the imports touched by this PR as suggested by @beorn7.

juliusv:
@tariq1890 I would just go for ii) for now - we made quite an effort to have even more import groups than that initially (stdlib, renamed stdlib, third-party, renamed third-party, same repo), but honestly I think nobody cares *that* much anymore :)

tariq1890:
Cool! A question though - `github.com/pkg/errors` is considered a non std lib yes?

discovery/openstack/instance.go
```diff
@@ -125,11 +126,11 @@ func (i *InstanceDiscovery) refresh(ctx context.Context) (*targetgroup.Group, er
 	}
 	err = pager.EachPage(func(page pagination.Page) (bool, error) {
 		if ctx.Err() != nil {
-			return false, fmt.Errorf("could not extract instances: %s", ctx.Err())
+			return false, errors.Wrap(err, "could not extract instances")
```
simonpasquier: ctx.Err()

juliusv:
> `github.com/pkg/errors` is considered a non std lib, yes?

Yep.

discovery/openstack/instance.go
tariq1890: Thanks!

tariq1890:
Phew! Done :). Addressed the comments by @beorn7 & @simonpasquier. Ready for the next round of review.

juliusv:
Thanks, hopefully should be good now :+1:

PR #5403, opened 2019-03-25
cmd: Warn rather than Info when retention time wraps

brian-brazil:

@beorn7


beorn7:
Sweet. Thanks.

PR #5414, opened 2019-03-28
Handle scrape cache better for failed scrapes

brian-brazil:
@brancz This should let us have a much more correct idea of churn, though the exhaustion mitigation messes that up a little.

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -125,6 +125,12 @@ var (
 			Help: "Total number of samples rejected due to timestamp falling outside of the time bounds",
 		},
 	)
+	targetScrapeCacheFlushForced = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "prometheus_target_scrapes_cache_flush_forced_total",
+			Help: "How many times a scrape cache was flushed due to getting big while scrapes are failing.",
+		},
+	)
 )
 
 func init() {
@@ -140,6 +146,7 @@ func init() {
 	prometheus.MustRegister(targetScrapeSampleDuplicate)
 	prometheus.MustRegister(targetScrapeSampleOutOfOrder)
 	prometheus.MustRegister(targetScrapeSampleOutOfBounds)
+	prometheus.MustRegister(targetScrapeCacheFlushForced)
 }
 
 // scrapePool manages scrapes for sets of targets.
@@ -606,6 +613,9 @@ type scrapeLoop struct {
 type scrapeCache struct {
 	iter uint64 // Current scrape iteration.
 
+	// How many series and metadata entries there were at the last success.
+	successfulCount int
+
 	// Parsed string to an entry with information about the actual label set
 	// and its storage reference.
 	series map[string]*cacheEntry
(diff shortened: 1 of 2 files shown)


scrape/scrape.go
```diff
@@ -1184,7 +1185,7 @@ loop:
 		return total, added, err
 	}
 
-	sl.cache.iterDone()
+	sl.cache.iterDone(len(b) > 0)
```
brancz: this deserves a comment explaining why this condition makes it right to drop cache, also extracting it here into a variable that makes it obvious already what this condition does would be helpful as well

scrape/scrape.go
```diff
@@ -643,28 +646,44 @@ func newScrapeCache() *scrapeCache {
...
-		if c.iter-e.lastIter > 2 {
-			delete(c.series, s)
+func (c *scrapeCache) iterDone(flushCache bool) {
+	c.metaMtx.Lock()
+	count := len(c.series) + len(c.droppedSeries) + len(c.metadata)
+	c.metaMtx.Unlock()
+
+	if flushCache {
+		c.successfulCount = count
+	} else if count > c.successfulCount*2+100 {
```
brancz: My understanding is that this is a heuristic, that aims for tolerating targets that drop and replace all metrics in a consecutive scrape plus the metadata (which should be somewhat consistent in either case). At the very least this needs a comment describing the heuristic, but I'd also like us to consider different heuristics. For example as I said metadata is probably irrelevant, at least in terms of churn. So maybe just x2 or x3, without metadata is simpler and should allow us to catch close to all churn situations. Please correct me if I misunderstood any of this, which may very well have happened.

brian-brazil: This is more that if there's a small bit of churn don't lose anything. The 100 is to somewhat handle the case where this happens before there's been a successful scrape.

Metadata churn is a possibility. We've never seen it, but may aswell protect against it while we're at it.

scrape/scrape.go
brian-brazil: Done

scrape/scrape.go
```diff
@@ -643,28 +646,44 @@ func newScrapeCache() *scrapeCache {
...
+		// If a target had varying labels in scrapes that ultimately failed,
+		// the caches would grow indefinitely. Force a flush when this happens.
```
brancz: Force flushing should probably have a metric for one to be able to track if this happens often.

scrape/scrape.go
brancz: I’d like us to at least document the heuristic and maybe explain that flushing the cache on initial iter is ok (as successfulCount will still be 0 and count is likely to be higher than 100)

brian-brazil: Done, and bumped it to 1k.

scrape/scrape.go
brian-brazil: Done

scrape/scrape_test.go
```diff
@@ -779,6 +779,146 @@ func TestScrapeLoopRunCreatesStaleMarkersOnParseFailure(t *testing.T) {
...
+	}
+
+	// 1 successfully scraped sample, 1 stale marker after first fail, 4 report samples for
+	// each scrape successful or not.
+	if len(appender.result) != 22 {
+		t.Fatalf("Appended samples not as expected. Wanted: %d samples Got: %d", 22, len(appender.result))
+	}
+}
+
+func TestScrapeLoopCacheMemoryExhaustionProtection(t *testing.T) {
```
brancz: This test seems to fail consistently on CI

brian-brazil: I had changed the constant without updating the test, should work now.

brancz:
lgtm

Issue #5397, opened 2019-03-22
Prometheus stop timeout


wlay:
## Proposal
**Use case. Why is this important?**
Prometheus process can't exit gracefully.

*“Nice to have” is not a good use case. :)*

## Bug Report
**What did you do?**
sudo systemctl stop prometheus.service
**What did you expect to see?**
The process exit gracefully. And show "See you next time!"
**What did you see instead? Under which circumstances?**
The process didn't show the message "See you next time!" until it timeout.

**Environment**
centos
* System information:

	Linux 3.10.0-514.el7.x86_64 x86_64

* Prometheus version:

	2.7.1
* Alertmanager version:

	didn't deploy it.

* Prometheus configuration file:
```
insert configuration here
```

* Alertmanager configuration file:
```
insert configuration here (if relevant to the issue)
```

* Logs:
```
prometheus log:
level=warn ts=2019-03-22T08:35:46.127409912Z caller=main.go:464 msg="Received SIGTERM, exiting gracefully..."
level=info ts=2019-03-22T08:35:46.127525054Z caller=main.go:489 msg="Stopping scrape discovery manager..."
level=info ts=2019-03-22T08:35:46.127559332Z caller=main.go:503 msg="Stopping notify discovery manager..."
level=info ts=2019-03-22T08:35:46.127573097Z caller=main.go:525 msg="Stopping scrape manager..."
level=info ts=2019-03-22T08:35:46.127571547Z caller=main.go:485 msg="Scrape discovery manager stopped"
level=info ts=2019-03-22T08:35:46.127623261Z caller=main.go:499 msg="Notify discovery manager stopped"
level=info ts=2019-03-22T08:35:50.821469229Z caller=manager.go:745 component="rule manager" msg="Stopping rule manager..."
level=info ts=2019-03-22T08:35:50.821504484Z caller=main.go:519 msg="Scrape manager stopped"
level=info ts=2019-03-22T08:35:50.821581696Z caller=manager.go:751 component="rule manager" msg="Rule manager stopped"
level=info ts=2019-03-22T08:35:50.834491455Z caller=queue_manager.go:259 component=remote queue="0:https://xxxx/write?
db=xxxxx&rp=autogen&precision=s" msg="Stopping remote storage..."
level=info ts=2019-03-22T08:35:50.834598573Z caller=queue_manager.go:267 component=remote queue="0:https://xxxx/write?
db=xxxxx&rp=autogen&precision=s" msg="Remote storage stopped."
[...]

PR #5420, opened 2019-03-29
discovery/file: log slow drains

simonpasquier:
Related to https://github.com/prometheus/prometheus/issues/5397
If for some reason Prometheus can't drain the fsnotify events, it blocks the shutdown/reload and nothing will warn about it.


simonpasquier:
> And I'm curious in what case Prometheus isn't able to drain the fsnotify events.

I've got no concrete example: the OP of #5397 mentioned this issue initially but redacted later on. I'd say we can forget about this PR for now.

PR #5422, opened 2019-03-30
Add support for POSTing to /series endpoint

### LINKED ISSUE #5417: [Feature Request] Support POST method for series endpoint
## Proposal
Some user queries `/api/v1/series` with many label values, it will trigger the GET url length limit.

If support POST method for `/api/v1/series` like `/api/v1/query` endpoint, the use case will be fine.

bobmshannon:
Closes #5417

The [ParseForm()](https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L446) method in [net/http](https://golang.org/pkg/net/http/#Request.ParseForm) supports url-encoded parameters in the request body if the method is `POST`, with a built-in 10 MB limit.

This change should mitigate issues where a user runs into query string length limits when providing a large number of selectors using the `GET` method inline in the URL query string.

### DIFF
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -74,6 +74,7 @@ The following endpoint evaluates an instant query at a single point in time:
 
 ```
 GET /api/v1/query
+POST /api/v1/query
 ```
 
 URL query parameters:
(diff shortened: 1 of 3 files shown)


eahydra:
maybe need modify web/api/v1/api_test.go L838, support POST unit test for series endpoint.

eahydra:
LGTM

docs/querying/api.md
```diff
@@ -201,10 +201,11 @@ $ curl 'http://localhost:9090/api/v1/query_range?query=up&start=2015-07-01T20:10
 
 ### Finding series by label matchers
 
-The following endpoint returns the list of time series that match a certain label set.
+The following endpoints return the list of time series that match a certain label set.
```
brian-brazil: It's still one endpoint, but multiple verbs

bobmshannon:
I was going to suggest adding POST support to the query endpoints for https://github.com/prometheus/client_golang/issues/428, but it looks like that was already implemented. Instead, I've just updated the API documentation instead.

brian-brazil:
Thanks!

PR #5432, opened 2019-04-04
discovery/dns: fix slice with wrong length

simonpasquier:
Introduced by #5381. The bug is harmless because the SD manager drops `nil` items but better to have it fixed anyway.

https://github.com/prometheus/prometheus/blob/81c4248081d51d628916cb214ae7e9bceb17be94/discovery/manager.go#L293


brian-brazil:
:+1:

PR #5433, opened 2019-04-04
Fix reporting wrong default value for retention flags

vitkovskii:
Hey :) As you added the feature of size-based retention you changed some logic in the command line parameters parsing. And it leads to reporting wrong flags values on `/flags` page and in api `/api/v1/status/flags`.

**Steps to reproduce**
1) Build prometheus `2.8.1`
2) Run prometheus locally with empty config `./prometheus --config.file=empty.yaml`
3) It logs `no time or size retention was set so using the default time retention`
4) Go to `localhost:9090/flags`
5) You should see `storage.tsdb.retention.time`/`storage.tsdb.retention` = `15d`, but it shows `0s`

**Why it happens?**
Because default value for retention duration is set simply to cfg: `cfg.tsdb.RetentionDuration = defaultRetentionDuration`

But source values for `/flags` and `/api/v1/status/flags` are taken from kingping flag values:
```
`a.Flag("storage.tsdb.retention.time", "...").SetValue(&oldFlagRetentionDuration)`
`a.Flag("storage.tsdb.retention.time", "...").SetValue(&newFlagRetentionDuration)`
 ```
Setting value to `cfg.tsdb.RetentionDuration` aren't affect `oldFlagRetentionDuration` and `newFlagRetentionDuration` so you see `0s` instead of `15d`.

@brian-brazil

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -35,14 +35,14 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
-	conntrack "github.com/mwitkow/go-conntrack"
+	"github.com/mwitkow/go-conntrack"
 	"github.com/oklog/oklog/pkg/group"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/common/promlog"
 	"github.com/prometheus/common/version"
-	kingpin "gopkg.in/alecthomas/kingpin.v2"
+	"gopkg.in/alecthomas/kingpin.v2"
 	"k8s.io/klog"
 
 	promlogflag "github.com/prometheus/common/promlog/flag"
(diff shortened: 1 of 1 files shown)


brian-brazil:
This is something we've actually already discussed, the /flags is what the user has passed in - not what we subsequently do with it.

vitkovskii:
That would be the logic. But I don't pass `0s` to for `storage.tsdb.retention`. And don't pass `2h` for `storage.tsdb.min-block-duration` too. Why setting the default value in one case is shown as value and don't shown in other case? If kingpin sets the default value it will be on the page, but if you set it by hand it won't.

vitkovskii:
Anyway thanks for the fast reply and for the explanation :)

roidelapluie:
Note: we added retention info in the status page.

roidelapluie:
You do not pass it but it is still the actual value.

roidelapluie:
As the information is available in the status page, we have decided to close this PR.

PR #5437, opened 2019-04-08
*: enable golangci-lint

simonpasquier:
[golangci-lint](https://github.com/golangci/golangci-lint) runs many Go linters at the same time. In addition to [staticcheck](https://staticcheck.io/) that we already use in mostly all the repositories, it can also run [errcheck](https://github.com/kisielk/errcheck) (currently used by AlertManager), [govet](https://golang.org/cmd/vet/) and [more](https://github.com/golangci/golangci-lint#enabled-by-default-linters). The main advantage of using golangci-lint over each individual linter is that it's one-step installation and we can easily turn on/tune linters on a per-repository basis. It seems to have sane defaults and is also quite good at detecting and skipping generated code. I've checked with a few other repositories (tsbd, alertmanager, client_golang, common) and the update would be trivial.

There's also a GitHub app but again it requires broad access to the repositories (https://github.com/golangci/golangci/issues/32) which I'm not comfortable with.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,5 +14,5 @@ before_install:
 - travis_retry make deps
 
 script:
-- make check_license style unused test staticcheck check_assets
+- make check_license style unused test lint check_assets
 - git diff --exit-code
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,7 @@ go build ./cmd/prometheus/
 make test         # Make sure all the tests pass before you commit and push :)
 ```
 
-We use `golangci-lint`[https://github.com/golangci/golangci-lint] for linting the code. If it reports an issue and you think that it deserves an exception, you can add a special comment `//nolint:linter1[,linter2,...]` before the offending line. Use this sparingly though, fixing the code to comply with the linter's recommendation is in general the preferred course of action.
+We use `golangci-lint`[https://github.com/golangci/golangci-lint] for linting the code. If it reports an issue and you think that the warning needs to be disregarded or is a false-positive, you can add a special comment `//nolint:linter1[,linter2,...]` before the offending line. Use this sparingly though, fixing the code to comply with the linter's recommendation is in general the preferred course of action.
 
 All our issues are regularly tagged so that you can also filter down the issues involving the components you want to work on. For our labeling policy refer [the wiki page](https://github.com/prometheus/prometheus/wiki/Label-Names-and-Descriptions).
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,6 +40,8 @@ go build ./cmd/prometheus/
 make test         # Make sure all the tests pass before you commit and push :)
 ```
 
+We use `golangci-lint`[https://github.com/golangci/golangci-lint] for linting the code. If it reports an issue and you think that it deserves an exception, you can add a special comment `//nolint:linter1[,linter2,...]` before the offending line. Use this sparingly though, fixing the code to comply with the linter's recommendation is in general the preferred course of action.
+
 All our issues are regularly tagged so that you can also filter down the issues involving the components you want to work on. For our labeling policy refer [the wiki page](https://github.com/prometheus/prometheus/wiki/Label-Names-and-Descriptions).
 
 ## Pull Request Checklist
--- a/Makefile.common
+++ b/Makefile.common
@@ -166,6 +166,10 @@ else
 endif
 endif
 
+# For backward-compatibility.
+.PHONY: common-staticcheck
+common-staticcheck: lint
+
 .PHONY: common-unused
 common-unused: $(GOVENDOR)
 ifdef GOVENDOR
--- a/Makefile.common
+++ b/Makefile.common
@@ -72,14 +72,13 @@ endif
 PROMU_VERSION ?= 0.3.0
 PROMU_URL     := https://github.com/prometheus/promu/releases/download/v$(PROMU_VERSION)/promu-$(PROMU_VERSION).$(GO_BUILD_PLATFORM).tar.gz
 
-STATICCHECK :=
+GOLANGCI_LINT :=
+GOLANGCI_LINT_VERSION ?= v1.16.0
 # staticcheck only supports linux, freebsd, darwin and windows platforms on i386/amd64
 # windows isn't included here because of the path separator being different.
 ifeq ($(GOHOSTOS),$(filter $(GOHOSTOS),linux freebsd darwin))
 	ifeq ($(GOHOSTARCH),$(filter $(GOHOSTARCH),amd64 i386))
-		STATICCHECK := $(FIRST_GOPATH)/bin/staticcheck
-		STATICCHECK_VERSION ?= 2019.1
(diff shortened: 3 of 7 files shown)


simonpasquier:
Irrespective of whether we want golangci-lint or not, govet would be useful. I've just tried it and it spotted at least 2 mistakes:

```
web/federate.go:139:26: nilness: nil dereference in dynamic method call (govet)
                http.Error(w, err.Error(), http.StatusInternalServerError)
                                       ^
```
It should be `set.Err()` and not `err`

https://github.com/prometheus/prometheus/blob/f7184978f4ecd847aa0708a59aed701b0bd08c83/web/federate.go#L137-L141

```
discovery/kubernetes/endpoints.go:346:9: nilness: impossible condition: nil != nil (govet)
        if err != nil {
               ^
```

https://github.com/prometheus/prometheus/blob/f7184978f4ecd847aa0708a59aed701b0bd08c83/discovery/kubernetes/endpoints.go#L342-L348

brian-brazil:
How is it in terms of false positives? That's the usual issue with these
things.

beorn7:
IIRC `go vet` is supposed to not have false positives. A recent go release has auto-included `go vet` in tests (which, still assuming I'm remembering this correctly, we should actually run `go vet` implicitly by running tests).

simonpasquier:
> How is it in terms of false positives? That's the usual issue with these things.

golangci-lint claims to have few false-positives. In case we hit one (or if we found the linter too strict), it is possible to disable the check with a preceding comment (eg `//nolint:govet`). Which I should document in the contributing guide btw.

> A recent go release has auto-included go vet in tests (which, still assuming I'm remembering this correctly, we should actually run go vet implicitly by running tests).

You're right @beorn7. From the Go [documentation](https://golang.org/cmd/go/#hdr-Test_packages):

> As part of building a test binary, go test runs go vet on the package and its test source files to identify significant problems. If go vet finds any problems, go test reports those and does not run the test binary. Only a high-confidence subset of the default go vet checks are used. That subset is: 'atomic', 'bool', 'buildtags', 'nilfunc', and 'printf'. You can see the documentation for these and other vet tests via "go doc cmd/vet". To disable the running of go vet, use the -vet=off flag. 

So `go vet` is executed during tests but only with a subset of "safe'" checks.

.travis.yml
```diff
@@ -14,5 +14,5 @@ before_install:
 - travis_retry make deps
 
 script:
-- make check_license style unused test staticcheck check_assets
+- make check_license style unused test lint check_assets
```
krasi-georgiev: Why are we still using travis and circle? I thought there was some discussion to use only one of them?

Makefile.common
```diff
@@ -72,14 +72,13 @@ endif
 PROMU_VERSION ?= 0.3.0
 PROMU_URL     := https://github.com/prometheus/promu/releases/download/v$(PROMU_VERSION)/promu-$(PROMU_VERSION).$(GO_BUILD_PLATFORM).tar.gz
 
-STATICCHECK :=
+GOLANGCI_LINT :=
+GOLANGCI_LINT_VERSION ?= v1.16.0
 # staticcheck only supports linux, freebsd, darwin and windows platforms on i386/amd64
 # windows isn't included here because of the path separator being different.
 ifeq ($(GOHOSTOS),$(filter $(GOHOSTOS),linux freebsd darwin))
 	ifeq ($(GOHOSTARCH),$(filter $(GOHOSTARCH),amd64 i386))
```
krasi-georgiev: are these checks still needed?

Makefile.common
```diff
@@ -153,18 +152,17 @@ common-vet:
...
-ifdef STATICCHECK
-	@echo ">> running staticcheck"
-	chmod +x $(STATICCHECK)
+.PHONY: common-lint
+common-lint: $(GOLANGCI_LINT)
+ifdef GOLANGCI_LINT
+	@echo ">> running golangci-lint"
 ifdef GO111MODULE
 # 'go list' needs to be executed before staticcheck to prepopulate the modules cache.
 # Otherwise staticcheck might fail randomly for some reason not yet explained.
```
krasi-georgiev: Is this comment still needed?

beorn7:
@simonpasquier thanks for the explanation. Now we know why running the tests didn't find all the `go vet` issues. And also that `go vet` has checks that it feels more confident about than others. Still, I'd say we should have `go vet` on by default and exclude checks where needed. I'd say the checks are less noisy than some other tools we run by default.

.travis.yml
simonpasquier: We hit memory limits when running the tests in Circle CI which is why Travis is still used.

Makefile.common
simonpasquier: I've removed freebsd since it isn't supported by golangci-lint.

Makefile.common
simonpasquier: AFAICT yes because golangci-lint will run staticcheck behind the scenes.

.travis.yml
krasi-georgiev: bummer :+1:

simonpasquier:
FYI these are the changes required to enable the default linters:

https://github.com/simonpasquier/prometheus/compare/golangci-lint...simonpasquier:enable-default-linters

As I mentioned earlier some of the changes are actual bugs, others remove dead code or simplify the code structure.

tariq1890:
You beat me to this @simonpasquier :). I was about to implement this.

In regards to your latest comment, I wouldn't worry too much about `errcheck` and I would much rather not have `errcheck` than add `nolint` directives to code while keeping it. My two cents: `nolints` should be used a last resort and as a means of unblocking.

CONTRIBUTING.md
```diff
@@ -40,6 +40,8 @@ go build ./cmd/prometheus/
 make test         # Make sure all the tests pass before you commit and push :)
 ```
 
+We use `golangci-lint`[https://github.com/golangci/golangci-lint] for linting the code. If it reports an issue and you think that it deserves an exception, you can add a special comment `//nolint:linter1[,linter2,...]` before the offending line. Use this sparingly though, fixing the code to comply with the linter's recommendation is in general the preferred course of action.
```
tariq1890: I feel that `it deserves an exception` isn't too clear. Something along the lines of "if the linter's warning needs to be disregarded or is a false positive", perhaps?

tariq1890:
> How is it in terms of false positives? That's the usual issue with these things.

I've used `golangci-lint` for a long time and I can confidently say that it is very good about dealing with false positives (if there were any that is). The maintainer is active and very responsive to issues.

simonpasquier:
> In regards to your latest comment, I wouldn't worry too much about errcheck and I would much rather not have errcheck than add nolint directives to code while keeping it.

What we've done in AlertManager is to disable `errcheck` for tests and maintain a list of functions/methods that are excluded (see https://github.com/prometheus/alertmanager/blob/master/scripts/errcheck_excludes.txt). So far it worked quite well...

tariq1890:
:+1:

SuperQ:
After this is deployed to the other repos, we can go hunting for use of the `staticcheck` in CI and switch them over to `lint` and drop the compatibility mode.

tariq1890:
Can this be merged?

PR #5440, opened 2019-04-09
code optimize for config/config.go

legendtkl:
## before
In config/config.go, the data tranfered between func#LoadFile and func#Load is type converted unnecessary, ie:
```go
line 61:    cfg, err := Load(string(content))   // []byte -> string
line 47:    err := yaml.UnmarshalStrict([]byte(s), cfg)     //string -> []byte
```
Can we optimize these code as this PR?

## after
refer the commit in this PR

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -37,14 +37,14 @@ var (
 )
 
 // Load parses the YAML input s into a Config.
-func Load(s string) (*Config, error) {
+func Load(s []byte) (*Config, error) {
 	cfg := &Config{}
 	// If the entire config body is empty the UnmarshalYAML method is
 	// never called. We thus have to set the DefaultConfig at the entry
 	// point as well.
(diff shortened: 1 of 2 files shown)


brian-brazil:
What you're proposing is different, it would no longer make a copy of the default config.

legendtkl:
hi, @brian-brazil ,  I just update the pr before seeing your reply. Would you help check again? 

thanks.

brian-brazil:
How much of a difference would this make? Reloading configs happens rarely.

legendtkl:
Yes, you are right. In the view of performance,  there is nearly no difference. 
But on the side of pursuit of perfection for coding, this might be better.

roidelapluie:
We have looked at this during our bug scrub and decided not to merge this PR as the performance gain is close to zero.

PR #5441, opened 2019-04-09
Put myself down for 2.9 release.

brian-brazil:
Update docs on where to email release announcements.

Google Groups does not allow one group to be a member of another, so for now document our existing practice of emailing both -users and -ann.


brancz:
:+1:

PR #5443, opened 2019-04-09
discovery/kubernetes: add "...present_<...>" labels for Kubernetes labels and annotations

### LINKED ISSUE #3726: Support differentiation between empty label and non-existing labels in k8

simonpasquier:
Closes #3726


discovery/kubernetes/pod_test.go
```diff
@@ -178,18 +178,20 @@ func TestPodDiscoveryBeforeRun(t *testing.T) {
...
-					"__meta_kubernetes_pod_host_ip":                   "2.3.4.5",
-					"__meta_kubernetes_pod_ready":                     "true",
-					"__meta_kubernetes_pod_phase":                     "Running",
-					"__meta_kubernetes_pod_uid":                       "abc123",
-					"__meta_kubernetes_pod_controller_kind":           "testcontrollerkind",
-					"__meta_kubernetes_pod_controller_name":           "testcontrollername",
+					"__meta_kubernetes_pod_name":                              "testpod",
+					"__meta_kubernetes_namespace":                             "default",
+					"__meta_kubernetes_pod_label_testlabel":                   "testvalue",
+					"__meta_kubernetes_pod_label_testlabel_present":           "true",
```
brian-brazil: This will clash with a label of that name, and we currently have all dynamic bits of label names at the end for metadata.

simonpasquier: It shouldn't clash as a label `foo_present` would generate `__meta_kubernetes_pod_label_foo_present` and `__meta_kubernetes_pod_label_present_present` but I agree that having dynamic values surrounded by static prefix and suffix is confusing.
Are you ok with `__meta_kubernetes_pod_label_present_<label>` then? If not, happy to update with any alternative.

brian-brazil: That'd also clash.

`__meta_kubernetes_pod_labelpresent_<label>` maybe?

simonpasquier: Done deal :-)

brian-brazil: Did you forget to push?

PR #5445, opened 2019-04-10
Remote Write: Queue Manager specific metrics shouldn't exist if the queue no longer exists

### LINKED ISSUE #5395: Remote write should remove metrics for queues that have gone away
This is so alerts on queues being behind don't fire when the queues go away.

cstyan:
Unregister remote write queue manager specific metrics when stopping the queue manager.

fixes #5395
@tomwilkie

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -189,7 +189,7 @@ type QueueManager struct {
 	sentBatchDuration          prometheus.Observer
 	succeededSamplesTotal      prometheus.Counter
 	retriedSamplesTotal        prometheus.Counter
-	queueMetrics               []prometheus.Collector
+	shardCapacity              prometheus.Gauge
 }
 
 // NewQueueManager builds a new QueueManager.
@@ -220,31 +220,10 @@ func NewQueueManager(logger log.Logger, walDir string, samplesIn *ewmaRate, cfg 
 		samplesDropped:     newEWMARate(ewmaWeight, shardUpdateDuration),
 		samplesOut:         newEWMARate(ewmaWeight, shardUpdateDuration),
 		samplesOutDuration: newEWMARate(ewmaWeight, shardUpdateDuration),
-
-		highestSentTimestampMetric: &maxGauge{
-			Gauge: queueHighestSentTimestamp.WithLabelValues(name),
-		},
-		pendingSamplesMetric:  queuePendingSamples.WithLabelValues(name),
-		enqueueRetriesMetric:  enqueueRetriesTotal.WithLabelValues(name),
-		droppedSamplesTotal:   droppedSamplesTotal.WithLabelValues(name),
-		numShardsMetric:       numShards.WithLabelValues(name),
-		failedSamplesTotal:    failedSamplesTotal.WithLabelValues(name),
-		sentBatchDuration:     sentBatchDuration.WithLabelValues(name),
-		succeededSamplesTotal: succeededSamplesTotal.WithLabelValues(name),
-		retriedSamplesTotal:   retriedSamplesTotal.WithLabelValues(name),
-	}
-	// We can't add sentBatchDuration because Observer doesn't implement Collector interface.
-	t.queueMetrics = []prometheus.Collector{
-		t.highestSentTimestampMetric.Gauge, t.pendingSamplesMetric, t.enqueueRetriesMetric,
-		t.droppedSamplesTotal, t.numShardsMetric, t.failedSamplesTotal, t.succeededSamplesTotal, t.retriedSamplesTotal,
 	}
 
 	t.watcher = NewWALWatcher(logger, name, t, walDir)
 	t.shards = t.newShards()
-
-	// Initialise some metrics.
-	shardCapacity.WithLabelValues(name).Set(float64(t.cfg.Capacity))
-	t.pendingSamplesMetric.Set(0)
 
 	return t
 }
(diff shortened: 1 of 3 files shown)


brian-brazil:
I'm not sure this will work, what happens if the queue comes back?

A custom collector may be better here.

beorn7:
Registering and unregistering is working per se, but other things are messed up here.
IIRC, it won't work to register and unregister individual metrics retrieved from `somethingVec.WithLabelValues`. Those will only ever register the whole vector. But I'll verify.

cstyan:
@beorn7 yeah you're right, looks like it unregisters the entire vector.

beorn7:
The solution is actually very easy and clean. All the vectors are autoregistering. so you won't act on the register level at all. You create the applicable metrics in `NewQueueManager` with `WithLabelValues`, so you can simply delete them from the vectors with `Delete`, see https://godoc.org/github.com/prometheus/client_golang/prometheus#CounterVec.Delete and friends.

storage/remote/queue_manager.go
```diff
@@ -332,6 +338,10 @@ func (t *QueueManager) Stop() {
 	for _, labels := range t.seriesLabels {
 		release(labels)
 	}
+	// Un-register metrics so we don't have alerts for queues that are gone.
+	for _, m := range t.queueMetrics {
+		prometheus.Unregister(m)
```
beorn7: Essentially, just repeat every `WithLabelaValues` call here, but use `Delete` instead of it.

cstyan: This works, except now if there's a config reload (which would call stop for all the queues) but the remote write config hasn't changed, the same queues come back up but there are no metrics for them. 

If the remote write config changes on each reload (adding a queue, removing a queue, etc.) then the metrics are always correct.

cstyan: I would expect that `metricVec.WithLabelValues` would panic here ^ when the metrics don't seem to exist.

beorn7: `metricVec.WithLabelValues` recreates the metric if it has been deleted before.
Whenever you take the `Delete` codepath, you have to make sure that the `WithLabelValues` codepath is taken, too, if the queue comes back up.

brian-brazil: That'd reset the metric value too, which is possibly not what we want.

cstyan: I found the cause of the problem. We in remote storage `ApplyConfig` we create new `QueueManagers`, then stop the old ones. When we stop the old ones we remove the metrics even though we might then be starting the exact same queue's in terms of config. Working on a fix right now.

beorn7: You should think about what @brian-brazil said: deleting and recreating the metrics will reset them, and you might not want that. I guess to avoid that, we need to track if a queue is created in the exact same way.

beorn7: BTW: You can always do the `WithLabelValues` call as it will simply return the existing metric if there is already one. If possible, just don't do the `DeleteLabelValues` if you know you'll recreate a queue with the same name.

storage/remote/wal_watcher.go
```diff
@@ -140,13 +140,26 @@ func NewWALWatcher(logger log.Logger, name string, writer writeTo, walDir string
 // Start the WALWatcher.
 func (w *WALWatcher) Start() {
 	level.Info(w.logger).Log("msg", "starting WAL watcher", "queue", w.name)
+
+	// Setup the WAL Watchers metrics. We do this here rather than in the
+	// constructor because of the ordering of creating Queue Managers's,
+	// stopping them,and then starting new ones in storage/remote/storage.go ApplyConfig.
```
beorn7: Nit: blank missing in "them,and".

beorn7: This whole comment is actually not needed anymore, is it?

cstyan:
@beorn7 still thinking about this. If we had multiple remote writes for the same endpoint the way I'm deleting the metrics might not work as we want it to.

cstyan:
Could use @tomwilkie's input on whether resetting the metrics values is going to be an issue. 

At the moment we're using most of these to get differences between metrics (`samples_in` and `succeeded_samples`) so as long as both are reset (they are) we may be okay.

cstyan:
@brian-brazil @beorn7 talked to Tom. The metrics values resetting is the behaviour we want to go with right now. It makes sense to tie the metrics life to the lifecyle of the queue manager/WAL watcher, and right now we don't have a good way to uniquely identify a queue across restarts and only restart those with changes.

See: https://github.com/prometheus/prometheus/blob/master/storage/remote/storage.go#L82-L83

brian-brazil:
Will that produce sane results if someone is reloading very frequently?

cstyan:
It should, yes. More sane than the current implementation imo.

All the metrics will be created or deleted with the QueueManager's lifecycle now. This means that if you have 3 queues `0:url0`, `1:url1`, `2:url2` and if you delete the remote write from your config for the queue `1`, you will now have `0:url0` and `1:url2`. Right now you would have metrics for the original 3 queues as well, which is confusing.

For those who are reloading config frequently with or without changes to the remote write config it will just look like counters resetting to 0. For most of these metrics they're only really useful when comparing their rate to the rate of another metric.

cstyan:
Do you have any suggestions? This seems to me like the only option until we rework remote storage `ApplyConfig` and have a reliable way to identify queues across config reloads.

Consider the case where someone has two remote write configs for the same endpoint with different queue configs, relabel configs, etc. Right now the only way to identify series tied to a given remote write queue is via their queue label which is the index of that queues remote write config in the config file and the URL of the endpoint to send to. If they remove one of those from their config and reload, the queue that was previously index 1 is now index 0, and both had the same url, so now activity in that queue will update metrics for `0:url`. 

IMO, it's going to be easier to reason about what's happening with a given remote write queue across restarts if the metrics reset.

brian-brazil:
The main thing to do would that if the remote_write config isn't changing, don't reset the counters.

cstyan:
A queue's config not changing doesn't necessarily mean that it's index in the overall config hasn't changed. The index is what we use as the label.

AFAICT in `client_golang` we don't have a way to grab the value of a metric so that we could copy a queue's metric values, delete the metrics, and then recreate the metrics with the copied values.

brian-brazil:
I'm presuming that we'd check the entire remote_write config as one.

cstyan:
@brian-brazil yeah we can definitely do that. We could store the config in the `Storage` struct , or hash the configs, and compare the current config to the "new" config in `ApplyConfig`. I'd prefer to do that in a separate PR though.

brian-brazil:
SD does similar, so you can take whatever approach is it taking.

cstyan:
> SD does similar, so you can take whatever approach is it taking.

Are you just referring to using the `md5sum` of each Alertmanager config here: https://github.com/prometheus/prometheus/blob/master/cmd/prometheus/main.go#L423-L434

We're not using that to only restart Discoverers whose config has changed. Discovery Manager `ApplyConfig` cancels all Discoverers:
- https://github.com/prometheus/prometheus/blob/master/discovery/manager.go#L193
- https://github.com/prometheus/prometheus/blob/master/discovery/manager.go#L279-L286
- https://github.com/prometheus/prometheus/blob/master/discovery/manager.go#L216-L225

brian-brazil:
Thanks!

Issue #3579, opened 2017-12-13
[Feature request] Group alerts by group name on /alerts


zemek:
It would be nice to have the `/alerts` page on Prometheus which shows all of the defined alerts to be grouped by the group name. Sort of like how `/targets` is grouped by job.

beorn7:
I don't quite understand. Alerts are already grouped by alert name. But what is the group name? Alerts are only grouped in the Alertmanager. (And displaying the configured groups is indeed a problem right now, see https://github.com/prometheus/alertmanager/issues/868 .)

brian-brazil:
I believe they mean the name of the rule group.

aditya-konarde:
Came here looking for the same. 

The reasoning I think behind this feature request is: 

Lets say I have an app named `foo` and I define 100 rules for the app named `foo`.

As a user, I want to be able to see all the rules for that app grouped by the group I have defined the rules in. 

Can one of the maintainers please take a look?

pbhudiaBAE:
Can this issue be closed now since #5448 has been merged

simonpasquier:
Sure, thanks!

kormotodor:
Oh my god why did you made it.. It looks so bad now without that alerts sorting fancy method (firing to top from all bunch of rules)

pbhudiaBAE:
The grouping of alerts was a requested feature in multiple issues. There has already been some changes #5664 #5655 to make it more friendly. If you feel that this is still an issue you might want to consider opening a new issue.

Issue #1371, opened 2016-02-04
Improve display of alerts in /alerts


grobie:
The alerts on /alerts are currently ordered by their order in the rule files. As rule file names are not visible in that table, the order appears to be random. One way would be to sort all alerts alphabetically.

juliusv:
...or alternatively, group them by rule file. This could be better because thematically related alerts that are not necessarily alphabetically adjacent will stay grouped together.

brian-brazil:
By rule group would make the most sense, once we have those.

juliusv:
@brian-brazil Since rule groups cannot span files, it could even be nested. File -> Group.

pbhudiaBAE:
I can have a look into this

pbhudiaBAE:
![alerts](https://user-images.githubusercontent.com/46529282/55617556-e5d84900-5783-11e9-8454-066547134635.png)

I know this feature/issue has been mentioned a few times #3579 

Here's what I've managed to do so far.
The alerts are grouped by their group name. The group name row isn't a chevron but just the name of the group. When I refresh the page however the Group rows shuffle (I haven't changed/added any code on how alerts within groups should be organised since I am unsure of how it works exactly, it might be a change in web.go).

Here's a link to my branch
[https://github.com/BAESystemsAI/prometheus/tree/SortingAlertsv1](https://github.com/BAESystemsAI/prometheus/tree/SortingAlertsv1)

Any suggestions/ help would be really appreciated.

juliusv:
@pbhudiaBAE Looks like a great start! For the group sorting, you see how we use a `byAlertStateAndNameSorter` to sort alerts - we'd need something similar for the groups in web.go.

pbhudiaBAE:
![webln500 go](https://user-images.githubusercontent.com/46529282/55868689-822f9080-5b74-11e9-9a6c-36b856bfc47a.png)
![webgoln900](https://user-images.githubusercontent.com/46529282/55868693-8491ea80-5b74-11e9-8686-a2d345cc1de8.png)

@juliusv Thanks. Having trouble sorting the groups. I tried to implement a similar method as to how alerts are grouped but I get the following error:

web/web.go:501:33: cannot use alerts (type *rules.AlertingGroups) as type []*rules.AlertingGroups in field value
web/web.go:503:9: cannot use groupSorter.Groups (type []*rules.AlertingGroups) as type *rules.AlertingGroups in assignment
[...]

PR #5448, opened 2019-04-10
Sorting alerts by group name in /alerts

pbhudiaBAE:
As mentioned in #1371 and #3579, this is a draft pull request for sorting alerts by group name on the alerts page. As it stands I need some help sorting the groups alphabetically, when the page is refreshed the groups shuffle position.

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -18,7 +18,6 @@ import (
 	"errors"
 	"math"
 	"net/url"
-	"path/filepath"
 	"sort"
 	"sync"
 	"time"
@@ -268,7 +267,7 @@ func NewGroup(name, file string, interval time.Duration, rules []Rule, shouldRes
 func (g *Group) Name() string { return g.name }
 
 // File returns the group's file.
-func (g *Group) File() string { return filepath.Base(g.file) }
+func (g *Group) File() string { return g.file }
 
 // Rules returns the group's rules.
 func (g *Group) Rules() []Rule { return g.rules }
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"math"
 	"net/url"
+	"path/filepath"
 	"sort"
 	"sync"
 	"time"
@@ -267,7 +268,7 @@ func NewGroup(name, file string, interval time.Duration, rules []Rule, shouldRes
 func (g *Group) Name() string { return g.name }
 
 // File returns the group's file.
-func (g *Group) File() string { return g.file }
+func (g *Group) File() string { return filepath.Base(g.file) }
 
 // Rules returns the group's rules.
 func (g *Group) Rules() []Rule { return g.rules }
@@ -903,9 +904,8 @@ func (m *Manager) RuleGroups() []*Group {
 		rgs = append(rgs, g)
 	}
 
-	// Sort rule groups by file, then by name.
 	sort.Slice(rgs, func(i, j int) bool {
-		if rgs[i].name == rgs[j].name {
+		if rgs[i].file != rgs[j].file {
 			return rgs[i].file < rgs[j].file
 		}
 		return rgs[i].name < rgs[j].name
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -361,7 +361,7 @@ func (g *Group) hash() uint64 {
 	return l.Hash()
 }
 
-// AlertingRules returns the list of the group's alerting rules
+// AlertingRules returns the list of the group's alerting rules.
 func (g *Group) AlertingRules() []*AlertingRule {
 	g.mtx.Lock()
 	defer g.mtx.Unlock()
@@ -380,7 +380,7 @@ func (g *Group) AlertingRules() []*AlertingRule {
 	return alerts
 }
 
-// HasAlertingRules returns true if the group contains at least one AlertingRule
+// HasAlertingRules returns true if the group contains at least one AlertingRule.
 func (g *Group) HasAlertingRules() bool {
 	g.mtx.Lock()
 	defer g.mtx.Unlock()
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -361,6 +361,38 @@ func (g *Group) hash() uint64 {
 	return l.Hash()
 }
 
+// AlertingRules returns the list of the group's alerting rules
+func (g *Group) AlertingRules() []*AlertingRule {
+	g.mtx.Lock()
+	defer g.mtx.Unlock()
+
+	var alerts []*AlertingRule
+	for _, rule := range g.rules {
+		if alertingRule, ok := rule.(*AlertingRule); ok {
+			alerts = append(alerts, alertingRule)
+		}
+	}
+	sort.Slice(alerts, func(i, j int) bool {
+		return alerts[i].State() > alerts[j].State() ||
+			(alerts[i].State() == alerts[j].State() &&
+				alerts[i].Name() < alerts[j].Name())
+	})
+	return alerts
+}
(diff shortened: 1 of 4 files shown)


rules/manager.go
```diff
@@ -907,6 +907,34 @@ func (m *Manager) AlertingRules() []*AlertingRule {
...
+		alerts.Groups = append(alerts.Groups, group)
+	}
+	return alerts
+}
+
+type AlertingGroups struct {
+	Groups []*AlertingGroup
+}
+
+type AlertingGroup struct {
```
juliusv: Hmm, I think it might be better to just return the existing `rule.Group` structs here and add `Name()` and `AlertingRules()` methods on that type directly.

rules/manager.go
```diff
@@ -907,6 +907,34 @@ func (m *Manager) AlertingRules() []*AlertingRule {
...
+			if alertingRule, ok := rule.(*AlertingRule); ok {
+				group.Rules = append(group.Rules, alertingRule)
+			}
+		}
+		alerts.Groups = append(alerts.Groups, group)
+	}
+	return alerts
+}
+
+type AlertingGroups struct {
```
juliusv: I'm not sure we need this intermediary type. Just return a slice of `[]*rule.Group` directly from the method that returns the rule groups?

rules/manager.go
```diff
@@ -907,6 +907,34 @@ func (m *Manager) AlertingRules() []*AlertingRule {
...
+	Groups []*AlertingGroup
+}
+
+type AlertingGroup struct {
+	Name                 string
+	Rules                []*AlertingRule
+	AlertStateToRowClass map[AlertState]string
```
juliusv: This map is only relevant for display purposes and shouldn't live in the rule code or be attached to every individual group. Best to keep it part of the top-level data value that is passed into the HTML template.

web/web.go
```diff
@@ -497,19 +497,21 @@ func (h *Handler) Run(ctx context.Context) error {
 }
 
 func (h *Handler) alerts(w http.ResponseWriter, r *http.Request) {
-	alerts := h.ruleManager.AlertingRules()
-	alertsSorter := byAlertStateAndNameSorter{alerts: alerts}
-	sort.Sort(alertsSorter)
-
-	alertStatus := AlertStatus{
-		AlertingRules: alertsSorter.alerts,
-		AlertStateToRowClass: map[rules.AlertState]string{
```
juliusv: You'll probably want to keep the `AlertStatus` struct as the top-level value that is passed into the HTML template, but change the type to take alerting groups instead of individual alerting rules. In any case, `AlertStateToRowClass` should only be instantiated once, in this top-level field of that struct.

rules/manager.go
```diff
@@ -362,6 +362,38 @@ func (g *Group) hash() uint64 {
 	return l.Hash()
 }
 
+// AlertingRules returns the list of the group's alerting rules
```
squat: tiny nit: for consistency, godoc comments should be full sentences ending with a `.`

pbhudiaBAE: Done :)

pbhudiaBAE:
I think I've addressed the comments from @juliusv let me know if I've missed something or you want something changing. The "Fixed group sorting and comments" commit didn't pass the checks because I didn't run gofmt on manager.go and web.go. The new "Fixed group sorting and comments with gofmt" commit passes the checks.

I've also changed how the RuleGroups() method sorts groups, before if two groups had the same name then refreshing the page would occasionally shuffle the group, so I added code to check if the group name is the same then sort by file name so that the Group position is more likely to be fixed. 

I welcome any  feedback/ suggestions :+1:

rules/manager.go
```diff
@@ -871,7 +903,7 @@ func (m *Manager) RuleGroups() []*Group {
 
 	// Sort rule groups by file, then by name.
 	sort.Slice(rgs, func(i, j int) bool {
-		if rgs[i].file != rgs[j].file {
+		if rgs[i].name == rgs[j].name {
```
juliusv: It kinda seems odd to sort by file as if it's lower in the file->group->rule hierarchy than the group... how about always sorting by file first, then by group name, but surfacing the file name in the group header somehow (maybe `{{.Name}} ({{.File}})` instead of just `{{.Name}}`)? I don't think we should introduce a new dropdown level in the UI for just the file, but making it visible in an unobtrusive way could be good and make the file->group sorting behavior less confusing for users. Maybe even make the group name grey/italic/small or something, so it pops out less, but is still there?

juliusv:
Looks great besides the one comment, thanks :)

pbhudiaBAE:
// Sort rule groups by file, then by name.
	sort.Slice(rgs, func(i, j int) bool {
		if rgs[i].file != rgs[j].file {
			return rgs[i].file < rgs[j].file
		}
		return rgs[i].name < rgs[j].name
	})
	return rgs
}

You mean keep the sorting of groups like how it was before (above)?

I changed the File method on the Group struct to return **filepath.Base(g.file)** instead of **g.file** (so you only get the file name and not the whole path) and made the group name italic in the HTML.

Here's a screenshot of it so far.
GroupandFile

Some suggestions for styling:
- Move the File Name to the very end of the Group Header row.
- Make the Group Header row thinner than the normal alert rows.
- Make the Group header text size smaller than the alert text size.

pbhudiaBAE:
Here's a view with setting the Group Header row padding to 2px, which style looks better to you @juliusv. Should there be a label like Group:(Group Name) and File:(File Name) in the row so users know what they mean?
SmallerGroupHeader

pbhudiaBAE:
FileAndGroupName

I swapped the File and Group name since it is ordered by File name first, and made the group name smaller so it is still visible. If there aren't any more changes you would like I can make the commit and make the pull request an Open one.

juliusv:
Gah, so sorry for the delay again! Taking another look now!

rules/manager.go
juliusv: Hmm, your screenshots and updates say that the ordering now happens by file first, but the code here still looks like it's sorting by name first?

juliusv: Ah, I think a code push is missing?

juliusv:
Ah ok, was confused by the code not being there yet, but you said "I can make the commit", sorry :)

Regarding the formatting I might find something like "foo-rules.yml > alerting" more visually pleasing than the more explicit "File: foo-rules.yml  Group: alerting" (maybe in that case, even keep it the same font). Otherwise, great.

pbhudiaBAE:
File Group

@juliusv The latest commit changes the formatting to File Name > Group Name.

rules/manager.go
```diff
@@ -267,7 +268,7 @@ func NewGroup(name, file string, interval time.Duration, rules []Rule, shouldRes
 func (g *Group) Name() string { return g.name }
 
 // File returns the group's file.
-func (g *Group) File() string { return g.file }
+func (g *Group) File() string { return filepath.Base(g.file) }
```
juliusv: I think it's ok to preserve the full path here (this function is already used by `/rules` as well, where it already shows the whole path, see e.g. http://demo.robustperception.io:9090/rules).

juliusv:
Looks good besides the last comment, thanks! Could you mark the PR as ready for review as well?

juliusv:
(and as for what exactly we display how, I might still go through some iterations / tweaks of it after having experienced it for a while :))

juliusv:
I allowed myself to mark the PR as ready for review and will merge - thank you :)

sirikhum:
Is there any configurable parameter to have /alerts sorted by status as in the 2.9.x?

Issue #5451, opened 2019-04-10
Queue Manager: panic: close of closed channel 


bismarck:
## Bug Report
**What did you do?**
Ran prometheus with remote write enabled.
**What did you expect to see?**
No panic
**What did you see instead? Under which circumstances?**
Got a panic for close of closed channel .  I suspect it occurs because `(t *QueueManager) Stop()` is called in the middle of `(t *QueueManager) reshardLoop()`.
**Environment**

* System information:

	Linux 4.15.0-1037-azure x86_64

* Prometheus version:

	prometheus, version 2.8.1 (branch: remote-write-fix, revision: 4a913999f7e6f50ac8186f589561498c07e6dd53)
  build user:       user@machine
  build date:       20190405-17:39:10
  go version:       go1.12.1

This was a build of PR #5429 and Brian's TSDB changes 1099999.  I'm pretty sure this issue is present in master.

* Prometheus configuration file:
```
remote_write:
  - url: "http://HOST:7201/api/v1/prom/remote/write"
    queue_config:
      capacity: 100000
      min_shards: 1
      max_retries: 10
```

* Logs:
```
Apr 06 14:39:49 my-host prometheus[19149]: ts=2019-04-06T14:39:49.5887504Z caller=dedupe.go:111 component=remote level=info queue=0:http://10.113.11.11:7201/api/v1/prom/remote/w
Apr 06 14:39:49 my-host prometheus[19149]: panic: close of closed channel
Apr 06 14:39:49 my-host prometheus[19149]: goroutine 183 [running]:
Apr 06 14:39:49 my-host prometheus[19149]: github.com/prometheus/prometheus/storage/remote.(*shards).stop(0xc03ee227e0)
Apr 06 14:39:49 my-host prometheus[19149]:         /Users/my.user/Work/github/prometheus/storage/remote/queue_manager.go:585 +0x59
Apr 06 14:39:49 my-host prometheus[19149]: github.com/prometheus/prometheus/storage/remote.(*QueueManager).Stop(0xc06a8da340)
Apr 06 14:39:49 my-host prometheus[19149]:         /Users/my.user/Work/github/prometheus/storage/remote/queue_manager.go:325 +0x236
Apr 06 14:39:49 my-host prometheus[19149]: github.com/prometheus/prometheus/storage/remote.(*Storage).ApplyConfig(0xc0001525b0, 0xc0020d6270, 0x0, 0x0)
[...]

PR #5460, opened 2019-04-15
Prevent reshard concurrent with calling stop

vsliouniaev:
We should wait util the reshard loop and reshard calculations are complete before calling stop, otherwise the reshard stop, start can happen concurrently.

In practice this happens when doing a Prometheus reload at the same time as a reshard loop is happening

fixes: https://github.com/prometheus/prometheus/issues/5451

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -322,9 +322,9 @@ func (t *QueueManager) Stop() {
 	defer level.Info(t.logger).Log("msg", "Remote storage stopped.")
 
 	close(t.quit)
+	t.wg.Wait()
 	t.shards.stop()
 	t.watcher.Stop()
-	t.wg.Wait()
 
 	// On shutdown, release the strings in the labels from the intern pool.
 	t.seriesMtx.Lock()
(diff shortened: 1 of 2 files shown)


brian-brazil:
@cstyan You were looking into this.

storage/remote/queue_manager.go
```diff
@@ -322,9 +322,9 @@ func (t *QueueManager) Stop() {
 	defer level.Info(t.logger).Log("msg", "Remote storage stopped.")
 
 	close(t.quit)
+	t.wg.Wait()
```
cstyan: Can we leave a comment here? Something like
```suggestion
	// Wait for all QueueManager routines to end before stopping shards and WAL watcher. This
	// is to ensure we don't end up executing a reshard and shards.stop() at the same time, which
	// causes a closed channel panic.
```

vsliouniaev:
I initially also tried a locking approach, but depending on where locks were added, I would end up with hard to reproduce deadlocks, due to the comms all being done through asynchronous channels.

brian-brazil:
This looks like the right fix to me. Thanks!

PR #5463, opened 2019-04-15
Follow-up on #5009

beorn7:
@sylr I finally found some time to look in depth at #5009. As some things had changed upstream in the meantime, and @brian-brazil 's requests required quite invasive changes, things got really involved. Thus, I decided to take your commits and put my own on top in a new branch instead. I hope you'll understand.

@brian-brazil please have a look if this fits your ideas.

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -748,6 +748,12 @@ func reloadConfig(filename string, logger log.Logger, rls ...func(*config.Config
 		return errors.Errorf("one or more errors occurred while applying the new configuration (--config.file=%q)", filename)
 	}
 	promql.SetDefaultEvaluationInterval(time.Duration(conf.GlobalConfig.EvaluationInterval))
+
+	// Register conf as current config.
+	config.CurrentConfigMutex.Lock()
+	config.CurrentConfig = conf
+	config.CurrentConfigMutex.Unlock()
+
 	level.Info(logger).Log("msg", "Completed loading of configuration file", "filename", filename)
 	return nil
 }
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -443,7 +443,11 @@ func main() {
 				}
 				files = append(files, fs...)
 			}
-			return ruleManager.Update(time.Duration(cfg.GlobalConfig.EvaluationInterval), files)
+			return ruleManager.Update(
+				time.Duration(cfg.GlobalConfig.EvaluationInterval),
+				files,
+				cfg.GlobalConfig.ExternalLabels,
+			)
 		},
 	}
 
@@ -747,13 +751,8 @@ func reloadConfig(filename string, logger log.Logger, rls ...func(*config.Config
 	if failed {
 		return errors.Errorf("one or more errors occurred while applying the new configuration (--config.file=%q)", filename)
 	}
+
 	promql.SetDefaultEvaluationInterval(time.Duration(conf.GlobalConfig.EvaluationInterval))
-
-	// Register conf as current config.
-	config.CurrentConfigMutex.Lock()
-	config.CurrentConfig = conf
-	config.CurrentConfigMutex.Unlock()
-
 	level.Info(logger).Log("msg", "Completed loading of configuration file", "filename", filename)
 	return nil
 }
--- a/discovery/openstack/mock_test.go
+++ b/discovery/openstack/mock_test.go
@@ -247,7 +247,7 @@ func (m *SDMock) HandleHypervisorListSuccessfully() {
 		testHeader(m.t, r, "X-Auth-Token", tokenID)
 
 		w.Header().Add("Content-Type", "application/json")
-		fmt.Fprintf(w, hypervisorListBody)
+		fmt.Fprint(w, hypervisorListBody)
 	})
 }
 
@@ -544,7 +544,7 @@ func (m *SDMock) HandleServerListSuccessfully() {
 		testHeader(m.t, r, "X-Auth-Token", tokenID)
 
 		w.Header().Add("Content-Type", "application/json")
-		fmt.Fprintf(w, serverListBody)
+		fmt.Fprint(w, serverListBody)
 	})
 }
 
@@ -583,6 +583,6 @@ func (m *SDMock) HandleFloatingIPListSuccessfully() {
 		testHeader(m.t, r, "X-Auth-Token", tokenID)
 
 		w.Header().Add("Content-Type", "application/json")
-		fmt.Fprintf(w, listOutput)
+		fmt.Fprint(w, listOutput)
 	})
 }
--- a/rules/alerting.go
+++ b/rules/alerting.go
@@ -28,6 +28,7 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/config"
 
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/pkg/rulefmt"
@@ -79,8 +80,9 @@ func (s AlertState) String() string {
 type Alert struct {
 	State AlertState
 
-	Labels      labels.Labels
-	Annotations labels.Labels
+	Labels         labels.Labels
+	ExternalLabels labels.Labels
+	Annotations    labels.Labels
 
 	// The value at the last evaluation of the alerting expression.
 	Value float64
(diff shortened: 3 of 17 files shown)


beorn7:
OK, that squashing went horribly wrong. Trying again...

beorn7:
OK, looks better.

discovery/openstack/mock_test.go
```diff
@@ -583,6 +583,6 @@ func (m *SDMock) HandleFloatingIPListSuccessfully() {
 		testHeader(m.t, r, "X-Auth-Token", tokenID)
 
 		w.Header().Add("Content-Type", "application/json")
-		fmt.Fprintf(w, listOutput)
+		fmt.Fprint(w, listOutput)
```
brian-brazil: I think you need to rebase.

web/web.go
```diff
@@ -541,19 +541,39 @@ func (h *Handler) consoles(w http.ResponseWriter, r *http.Request) {
 	for k, v := range rawParams {
 		params[k] = v[0]
 	}
+
+	externalLabels := map[string]string{}
+	h.mtx.RLock()
+	els := h.config.GlobalConfig.ExternalLabels
+	h.mtx.RUnlock()
+	for _, el := range els {
```
brian-brazil: I think it would be wise to have a quick unittest either here or for the alerting rules to check that nil hasn't been passed to the templating.

beorn7: Yes, definitely. Will add at my next convenience.

discovery/openstack/mock_test.go
beorn7: I have rebased now, but the problem fixed here (as marked by staticcheck) is currently in master.

brian-brazil: Weird, @simonpasquier did you not fix these already?

cmd/prometheus/main.go
```diff
@@ -443,7 +443,11 @@ func main() {
 				}
 				files = append(files, fs...)
 			}
-			return ruleManager.Update(time.Duration(cfg.GlobalConfig.EvaluationInterval), files)
+			return ruleManager.Update(
+				time.Duration(cfg.GlobalConfig.EvaluationInterval),
+				files,
+				cfg.GlobalConfig.ExternalLabels,
```
sylr: I would have passed the entire `cfg.GlobalConfig` not just a subset. Like that in the future if you need to access new global configurations you wouldn't have to redo this tedious work.

rules/alerting.go
```diff
@@ -140,17 +143,27 @@ type AlertingRule struct {
 }
 
 // NewAlertingRule constructs a new AlertingRule.
-func NewAlertingRule(name string, vec promql.Expr, hold time.Duration, lbls, anns labels.Labels, restored bool, logger log.Logger) *AlertingRule {
+func NewAlertingRule(
+	name string, vec promql.Expr, hold time.Duration,
+	labels, annotations, externalLabels labels.Labels,
```
sylr: ditto

cmd/prometheus/main.go
beorn7: Yeah, I was thinking about that, too. Some of the `reloaders` do it that way, some not. It has pros and cons. If you pass down the whole `cfg.GlobalConfig`, you create a tight coupling of the rule manager to the config. On the other hand, the more parts of the config you are using, the more sense it makes.

I don't feel strongly and decided to stick with the way it has been done so far (which even includes the globbing logic). One might argue the one more thing to pass down flips the balance, but I don't think so.

@brian-brazil what do you think?

rules/alerting.go
beorn7: Here I really think we should not introduce a dependency on the config struct.

brian-brazil: Agreed, this should be separate.

cmd/prometheus/main.go
brian-brazil: I don't really mind either way, but given the code is reused by other projects it's probably best to keep the coupling minimal.

juliusv: Yeah, ideally each package defines its own config structs that are relevant to it and then the global config only ties those together. At least we were moving into that direction, even if not everything is doing that yet.

beorn7:
Now we have a test for external labels expanded in a label template of an alerting rule.

rules/alerting_test.go
```diff
@@ -142,3 +143,96 @@ func TestAlertingRuleLabelsUpdate(t *testing.T) {
...
+	)
+	result := promql.Vector{
+		{
+			Metric: labels.FromStrings(
+				"__name__", "ALERTS",
+				"alertname", "ExternalLabelDoesNotExist",
+				"alertstate", "pending",
+				"instance", "0",
+				"job", "app-server",
+				"external_label", "",
```
brian-brazil: Hmm, does this mean we're sending an empty label rather than no label?

beorn7: No, it means that accessing a non-existent key in a Go map results in the zero value of the value (in this case an empty string). I think my naming of the label is causing the misunderstanding here. I'll change it and add some more things to the template to test more and better demonstrate what we are testing.

beorn7:
Test is improved now, hopefully clarifying what it is testing.

WRT the staticcheck warnings: They show up for me in master, too. I assume the warning in promql/parse.go:330:3 is a new check, while the warnings in discovery/openstack/mock_test.go might only show up now because staticcheck might have decided only recently to also check test files? This is all just speculation. The warnings are all legit, and the fixes I included here should just work. Not including the fixes will make the pipeline red.

I would prefer to not hold back merging because of the staticcheck warnings. My fixes are in a separate commit, so they can be reverted easily for forensics or investigations (or on the off-chance that anything is wrong with them). Holding back merging will potentially cause additional work and risk if a merge conflict has to be resolved later. Also, early adopters (who build from master) will be prevented from trying out the new feature before we do a RC.

brian-brazil:
That's not how labels.FromStrings works as it's backed by an array. The
label should be missing if it is empty. The name is fine.

beorn7:
I am well aware how `labels.FromStrings` works. However, we don't pass in the backing array. That would be very awkward to work with in the templating (and also different to work with than `$labels`). The array is converted into a map before it is passed into the templating.

discovery/openstack/mock_test.go
simonpasquier: I suspect that @beorn7 runs `staticcheck` from master which introduces those new checks. In the CI, we always run a specific version of `staticcheck` (currently 2019.1).

brian-brazil: Grand, not stepping on anyone else's toes then.

brian-brazil:
I've dug around more, and there seems to be an existing bug in a corner case here: `{a=""}` is not the same as `{}` when you hash it, so could cause issues if the config changes. Likely tolerable. The alertmanager will deal with it okay, as it explicitly removes empty labels. As does scrape ingestion.

However the same issue exists on the path to the TSDB, so we could miss time series that are actually duplicate if the labels are templated.

beorn7:
I don't quite follow what the problem is here. Could you explain it in more detail? (Does it have to do with this PR at all or does "existing bug" implies that we have the problem already?)

brian-brazil:
> Could you explain it in more detail?

The two labels implementations in tsdb and Prometheus do not correctly handle the idea that a missing label is the same as an empty label, affecting hashing at least on which many other bits of logic depend on.

> Does it have to do with this PR at all or does "existing bug" implies that we have the problem already?

The problem existed already, your initial unittest was what let me spot it as I wouldn't have expected that test to pass. Nor does the change make it more likely to come up, label templating is fairly rare in the first place.

I'd suggest this be merged as-is, which I'll leave to you in case there's additional squashing you want to do.

beorn7:
Thanks for the explanation. If I understand correctly, the problem you see is that "copying" a label via templating (i.e. `new_label: "{{$labels.old_label}}"`) will always create `new_label` even if `old_label` doesn't exist. I'm still not sure that's actually a problem.

But anyway, following your suggestion, I'm merging this now.

Thanks everybody involved for your contributions.

brian-brazil:
> If I understand correctly,

That's not the problem per-se, the problem is that when creating ALERTS that the presence of the empty `new_label` will affect the hash.

beorn7:
So the problem would occur if you create two alerts that are all the same with the only exception of one alert having `new_label` in the labels section. If that label happens to be empty, we would have two equivalent alerts that would not be detected as duplicates by the TSDB. Is that the case you have in mind?

(What happens if I just set `new_label: ""` in the labels section? Perhaps we don't even need templating to tickle the problem.)

brian-brazil:
Yes, that's one case. Another is changing the alert definition over time. I'll look at fixing this later in the week.

PR #5467, opened 2019-04-16
Make sure subquery range is taken into account for selection

gouthamve:


gouthamve:
This would make sure that `remote_read` and other downstream library users get the right `SelectParams` to select the data.

promql/engine.go
```diff
@@ -552,6 +552,16 @@ func (ng *Engine) populateSeries(ctx context.Context, q storage.Queryable, s *Ev
 			Step:  durationToInt64Millis(s.Interval),
 		}
 
+		// We need to make sure we select the timerange selected by the subquery.
+		// TODO(gouthamve): cumulativeSubqueryOffset gives the sum of range and the offset
+		// we can optimise it by separating out the range and offsets, and substracting the offsets
+		// from end also.
+		subqOffset := ng.cumulativeSubqueryOffset(path)
+		if subqOffset > 0 {
```
brian-brazil: You don't need this if

brian-brazil:
This should go against master.

gouthamve:
@brian-brazil Addressed all comments.

promql/engine_test.go
```diff
@@ -227,6 +227,167 @@ func TestQueryError(t *testing.T) {
 	}
 }
 
+// paramCheckerQuerier implements storage.Querier	which checks the start and end times
```
brian-brazil: stray tab here before which

promql/engine_test.go
```diff
@@ -227,6 +227,167 @@ func TestQueryError(t *testing.T) {
...
+		query: "foo[2m] offset 2m",
+		start: 300,
+
+		paramStart: 60,
+		paramEnd:   180,
+	}, {
+		query: "foo[2m:1s]",
+		start: 300,
+
+		paramStart: 175, // 300 - 120 - 5(!)
```
brian-brazil: The -5 is correct here. We may need that data to calculate `foo` at the very start of the subquery.

promql/engine_test.go
```diff
@@ -227,6 +227,167 @@ func TestQueryError(t *testing.T) {
...
+		paramEnd:   300,
+	}, {
+		query: "count_over_time((foo offset 10s)[2m:1s] offset 10s)",
+		start: 300,
+
+		paramStart: 155, // 300 - 120 - 5 - 10 - 10
+		paramEnd:   290,
+	},
+		// Range queries now.
+		{
```
brian-brazil: Indentation doesn't seem right here

gouthamve:
Done :)

PR #5475, opened 2019-04-17
Add azure public ip label

EarthmanT:
Add a public IP label to Azure config


brian-brazil:
Can you add a DCO and update the docs?

EarthmanT:
> Can you add a DCO and update the docs?

Added Docs and rebased with DCO.

brian-brazil:
That's failing to build. Are you sure you've the right field name?

brian-brazil:
Thanks!

PR #5479, opened 2019-04-18
Exhaust every request body before closing it (#5166)

elwinar:
From the documentation:
> The default HTTP client's Transport may not
> reuse HTTP/1.x "keep-alive" TCP connections if the Body is
> not read to completion and closed.

This effectively enable keep-alive for the fixed requests.


brian-brazil:
Thanks!

Issue #5167, opened 2019-02-01
prometheus still send alert to alertmanager/api/v1


ArcherLinXin:
log is : caller=notifier.go:481 component=notifier alertmanager=http://*******/api/v1/alerts

roidelapluie:
Alertmanager v2 api is still fairly new and still has a few bugs.

I think it is safer to keep using v1 for the moment

simonpasquier:
For the record you can always set Prometheus to use the v2 api using `relabel_configs` like this:

```yaml
alerting:
  alertmanagers:
    relabel_configs:
      replacement: "/api/v2/alerts"
      target_label: __alerts_path__
   ...
```

As for when/if we switch Prometheus to the v2 API by default, this is a good question...

PR #5482, opened 2019-04-18
config&notifier: Add option to use Alertmanager API v2

mxinden:
With v0.16.0 Alertmanager introduced a new API (v2). This patch adds a
configuration option for Prometheus to send alerts to the v2 endpoint
instead of the defautl v1 endpoint.

Fixes https://github.com/prometheus/prometheus/issues/5167


notifier/notifier.go
```diff
@@ -490,13 +513,40 @@ func (n *Manager) sendAll(alerts ...*Alert) bool {
 				wg.Done()
 			}(ams, am)
 		}
-		ams.mtx.RUnlock()
```
mxinden: This would unlock the read lock, even though the spawned go routines still had references to `ams`, resulting in possible race conditions. Instead, the lock id dropped at the end of the function via `defer`. This is safe as the function waits for all spawned goroutines before exiting.

notifier/notifier.go
```diff
@@ -30,21 +30,22 @@ import (
...
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+	"github.com/go-openapi/strfmt"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	config_util "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/common/version"
 
+	"github.com/prometheus/alertmanager/api/v2/models"
```
mxinden: This patch vendors the Alertmanager API v2 model. From my point of view, this gives us a strong contract between the two APIs. Let me know what you think.

config/config.go
```diff
@@ -454,6 +454,45 @@ func (c *AlertingConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
...
+
+// UnmarshalYAML implements the yaml.Unmarshaler interface.
+func (v *AlertmanagerAPIVersion) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	*v = AlertmanagerAPIVersion("")
+	type plain AlertmanagerAPIVersion
+	if err := unmarshal((*plain)(v)); err != nil {
+		return err
+	}
+
+	// Default to Alertmanager API v1.
```
brian-brazil: This isn't how we generally implement defaults.

notifier/notifier.go
simonpasquier: Another option would be to still release the lock here and pass the necessary inputs to the goroutine:

```golang
ams.mtx.RLock()
for _, am := range ams.ams {
   ...
   go func(c *http.Client, u string) {
       if err := n.sendOne(ctx, c, u, b); err != nil {
       ...
      }
   }(ams.client, am.url().String())
}
ams.mtx.RUnlock()
```

notifier/notifier.go
```diff
@@ -467,7 +462,35 @@ func (n *Manager) sendAll(alerts ...*Alert) bool {
...
 	for _, ams := range amSets {
+		var (
+			b   []byte
+			err error
+		)
+
 		ams.mtx.RLock()
+		defer ams.mtx.RUnlock()
+
+		switch ams.cfg.APIVersion {
```
simonpasquier: Maybe it could be cached to avoid serializing several times the same data.

notifier/notifier.go
simonpasquier: Looks good to me.

notifier/notifier.go
mxinden: That's even better, thanks @simonpasquier.

notifier/notifier.go
mxinden: Good point. Done.

mxinden:
@brian-brazil I updated _docs/configuration/configuration.md_. Any other docs that I am missing?

config/config.go
mxinden: Changed to the default way of doing defaults. Thanks.

brian-brazil:
There's Alerting -> Clients, but that's not in this repo.

mxinden:
@simonpasquier @brian-brazil any further thoughts on this?

brian-brazil:
:+1:

mxinden:
Tested this on a larger setup as well.

Thanks for the help.

PR #5485, opened 2019-04-18
Check if RemoteWrite config has changed in ApplyConfig

cstyan:
@brian-brazil from our discussion on #5445 

Don't stop, recreate, and start remote storage QueueManagers if the remote write config hasn't changed at all.

We can make more changes in the future to uniquely identify each remote write queue in the config.

cc @tomwilkie


brian-brazil:
:+1:

PR #4458, opened 2018-08-04
feat: added more k8s SD metrics

FUSAKLA:
@simonpasquier 

So there are 3 sources of metrics from the k8s client-go library I managed to find and added.
- HTTP requests 
- Cache metrics
- Workqueue metrics

I suppose HTTP requests are the main source of information we want to get. The cache and workqueue metrics I leave up to you to decide if we want to provide.

Hypothetically could be switched on/off in configuration of the SD definition?  

I tested it in our cluster and there seems to be no issues.
Here is link to example output: https://pastebin.com/raw/k3YiEmnc

(I'm newbie to Golang so please excuse me if this is somehow twisted approach)

PR #5486, opened 2019-04-18
update client-go,api,api-machinery and klog dependencies

tariq1890:
Bringing the dependencies to `v1.14.1` and klog to `v0.3.0`


tariq1890:
Can we run benchmarks against this PR please :) ?

discovery/kubernetes/client_metrics.go
```diff
@@ -251,31 +252,51 @@ func (f *clientGoWorkqueueMetricsProvider) NewDepthMetric(name string) workqueue
...
 		metric.Set(v / 1e6)
 	})
-
 }
 func (clientGoWorkqueueMetricsProvider) NewRetriesMetric(name string) workqueue.CounterMetric {
 	// Retries are not used so the metric is omitted.
 	return noopMetric{}
 }
+func (clientGoWorkqueueMetricsProvider) NewDeprecatedDepthMetric(name string) workqueue.GaugeMetric {
+	return clientGoWorkqueueDepthMetricVec.WithLabelValues(name)
```
simonpasquier: Shouldn't we return `noopMetric{}` here as well as for the other deprecated metrics?

discovery/kubernetes/client_metrics.go
```diff
@@ -251,31 +252,51 @@ func (f *clientGoWorkqueueMetricsProvider) NewDepthMetric(name string) workqueue
...
+func (f *clientGoWorkqueueMetricsProvider) NewWorkDurationMetric(name string) workqueue.HistogramMetric {
+	return clientGoWorkqueueWorkDurationMetricVec.WithLabelValues(name)
 }
 func (f *clientGoWorkqueueMetricsProvider) NewUnfinishedWorkSecondsMetric(name string) workqueue.SettableGaugeMetric {
 	return clientGoWorkqueueUnfinishedWorkSecondsMetricVec.WithLabelValues(name)
 }
-func (f *clientGoWorkqueueMetricsProvider) NewLongestRunningProcessorMicrosecondsMetric(name string) workqueue.SettableGaugeMetric {
+func (f *clientGoWorkqueueMetricsProvider) NewLongestRunningProcessorSecondsMetric(name string) workqueue.SettableGaugeMetric {
 	metric := clientGoWorkqueueLongestRunningProcessorMetricVec.WithLabelValues(name)
 	return gaugeSetFunc(func(v float64) {
```
simonpasquier: IIUC the conversion from microseconds to seconds should be removed.

discovery/kubernetes/client_metrics.go
```diff
@@ -127,12 +128,12 @@ var (
 		},
 		[]string{"queue_name"},
 	)
-	clientGoWorkqueueLatencyMetricVec = prometheus.NewSummaryVec(
```
simonpasquier: Originally, we [decided](https://github.com/prometheus/prometheus/pull/4458#pullrequestreview-148651324) on quantile-less summaries to cut down on the number of time-series. But it might be worth adding a comment at the top of the `var()` block.

discovery/kubernetes/client_metrics.go
tariq1890: Yes. I applied the change wrongly to WorkDuration. Correcting that now. Thanks.

discovery/kubernetes/client_metrics.go
tariq1890: I thought it would be a safer bet to keep the implementation as-is. Since it's many of the previously implemented methods in k8s `1.13` that got deprecated in `k8s 1.14`.

I can change it to `noopMetric{}` if you think that's more appropriate.

discovery/kubernetes/client_metrics.go
tariq1890: Then I'll revert this change. It'll still work if it's declared as a Summary.

tariq1890:
I've addressed all of review comments but one @simonpasquier. Waiting on your answer for the last one.

krasi-georgiev:
do you still need the bench running on this one?

discovery/kubernetes/client_metrics.go
simonpasquier: For gauges, it shouldn't matter much (although it is extra work) but the counters and summaries will double counted.

tariq1890: Makes sense. I’ll make it noopMetric for all the deprecated methods then.

tariq1890:
@simonpasquier Done!

tariq1890:
@brancz Requesting your review.

tariq1890:
@simonpasquier Who can provide a second review for this PR?

brancz:
Sorry I missed that notification. I’ll review on Monday :)

brancz:
lgtm :+1: thanks!

Issue #5480, opened 2019-04-18
Panic: 2.8.0+ When prometheus is asked for a reload we almost every time get  `panic: released unknown string`


beorn-:
## Bug Report

Since 2.8.x when we trigger a prometheus daemon reload the prometheus instance crashes. 

the instance sustains a stable 150k datapoints/s stream

We are running on linux stretch amd64.

* System information:

	Linux 4.18.0-0.bpo.1-amd64 x86_64

* Prometheus version:

  prometheus, version 2.9.1 (branch: HEAD, revision: ad71f2785fc321092948e33706b04f3150eee44f)
  build user:       root@09f919068df4
  build date:       20190416-17:50:04
  go version:       go1.12.4

* Prometheus configuration file:
```
global:
  scrape_interval:     15s # Set the scrape interval to every 15 seconds. Default is every 1 minute.
  evaluation_interval: 15s # Evaluate rules every 15 seconds. The default is every 1 minute.
  external_labels:
      monitor: 'prometheus'

rule_files:
  - "rules/*.yml"
  - "rules/app/*.yml"

scrape_configs:

[..]

remote_write:
  - url: "http://<redacted>:8000/write"
    queue_config:
      capacity: 1000000
      max_shards: 2000
```

* Logs:
```
Apr 18 10:39:59 srv01 systemd[1]: Reloading Monitoring system and time series database.
Apr 18 10:39:59 srv01 prometheus[8124]: level=info ts=2019-04-18T10:39:59.297Z caller=main.go:724 msg="Loading configuration file" filename=/etc/prometheus/prometheus.yml 
Apr 18 10:39:59 srv01 systemd[1]: Reloaded Monitoring system and time series database.
Apr 18 10:39:59 srv01 prometheus[8124]: ts=2019-04-18T10:39:59.302Z caller=dedupe.go:111 component=remote level=info queue=0:http://lts01/url msg="Stopping remote storage..."
Apr 18 10:39:59 srv01 prometheus[8124]: ts=2019-04-18T10:39:59.459Z caller=dedupe.go:111 component=remote level=info queue=0:http://lts01/url msg="WAL watcher stopped" queue=0:http://lts01/url
Apr 18 10:40:00 srv01 prometheus[8124]: ts=2019-04-18T10:40:00.444Z caller=dedupe.go:111 component=remote level=info queue=0:http://lts01/url msg="Remote storage stopped."
[...]

PR #5487, opened 2019-04-18
Remote Storage: string interner should not panic in release

cstyan:
Need to dig a bit more to find the call stack that lead to the panic but this will mean we don't panic at least.

@brian-brazil @beorn-

cc @tomwilkie

### DIFF
--- a/storage/remote/intern.go
+++ b/storage/remote/intern.go
@@ -21,9 +21,18 @@ package remote
 import (
 	"sync"
 	"sync/atomic"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
 var interner = newPool()
+var noReferenceReleases = promauto.NewCounter(prometheus.CounterOpts{
+	Namespace: namespace,
+	Subsystem: subsystem,
+	Name:      "string_interner_zero_reference_releases_total",
+	Help:      "The number of times release has been called for strings that are not interned.",
+})
 
 type pool struct {
 	mtx  sync.RWMutex
@@ -73,7 +82,8 @@ func (p *pool) release(s string) {
 	p.mtx.RUnlock()
 
 	if !ok {
-		panic("released unknown string")
+		noReferenceReleases.Inc()
+		return
 	}
 
 	refs := atomic.AddInt64(&interned.refs, -1)
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -339,6 +339,10 @@ func (t *QueueManager) Stop() {
 
 // StoreSeries keeps track of which series we know about for lookups when sending samples to remote.
 func (t *QueueManager) StoreSeries(series []tsdb.RefSeries, index int) {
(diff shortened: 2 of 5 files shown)


storage/remote/intern.go
```diff
@@ -73,7 +83,7 @@ func (p *pool) release(s string) {
 	p.mtx.RUnlock()
 
 	if !ok {
-		panic("released unknown string")
+		p.noReferenceReleases.Inc()
```
brian-brazil: Should we return here?

storage/remote/intern.go
```diff
@@ -38,6 +42,12 @@ type entry struct {
 func newPool() *pool {
 	return &pool{
 		pool: map[string]*entry{},
+		noReferenceReleases: prometheus.NewCounter(prometheus.CounterOpts{
```
simonpasquier: The metric isn't registered anywhere?

cstyan:
PTAL

cc @simonpasquier I needed the `testutil` package from `client_golang`, which wasn't already vendored. I ran `go get github.com/prometheus/client_golang/prometheus/testutil@65d3a96fbaa7c8c9535d7133d6d98cd50eed4db8` and then `go mod tidy && go mod vendor`. Let me know if it doesn't look correct. I've noticed that our `go.mod` says that we're using `client_golang` at `v0.9.3` for some commit even though the latest actual release was `v0.9.2`. I assume go mod updates the version in the files itself.

cstyan:
fixes #5480

simonpasquier:
> I needed the testutil package from client_golang, which wasn't already vendored. I ran go get github.com/prometheus/client_golang/prometheus/testutil@65d3a96fbaa7c8c9535d7133d6d98cd50eed4db8 and then go mod tidy && go mod vendor. 

You wouldn't even need to `go get .../client_golang` as it was already present. Importing the testutil package and running `go mod tidy/go mod vendor` would have been enough to update the vendor directory.

> I've noticed that our go.mod says that we're using client_golang at v0.9.3 for some commit even though the latest actual release was v0.9.2. I assume go mod updates the version in the files itself.

This is because we're referencing a commit after the last known tag, see https://tip.golang.org/cmd/go/#hdr-Pseudo_versions.

storage/remote/intern.go
```diff
@@ -36,9 +40,17 @@ type entry struct {
...
+	p := &pool{
 		pool: map[string]*entry{},
+		noReferenceReleases: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "string_interner_zero_reference_releases_total",
+			Help:      "The number of times release has been called for strings that are not interned.",
+		}),
 	}
+	prometheus.Register(p.noReferenceReleases)
```
brian-brazil: This is unsafe, both the metric and the registration should be at file level.

cstyan:
I'd like to make the interner and it's metric not package level variables but I'll do that later in another PR

brian-brazil:
Thanks!

beorn-:
Thank you very much !

Issue #5326, opened 2019-03-10
Vector/vector filter comparisons return match group labels rather than LHS labels


juliusv:
See http://demo.robustperception.io:9090/graph?g0.range_input=1h&g0.expr=node_cpu_seconds_total%20%3D%3D%20on(cpu%2C%20instance%2C%20job%2C%20mode)%20node_cpu_seconds_total&g0.tab=1&g1.range_input=1h&g1.expr=node_cpu_seconds_total%20%3D%3D%20node_cpu_seconds_total&g1.tab=1

When adding an `on()` modifier to a vector-to-vector filter, the metric name is dropped. Without the modifier, the metric name is retained. According to the docs at https://prometheus.io/docs/prometheus/latest/querying/operators/#comparison-binary-operators, the metric name should always be kept for filter operators (when not using the `bool` modifier).

juliusv:
Btw., when including `__name__` in the `on()` modifier, the metric name is retained.

juliusv:
To generalize, when including an `on()` clause, only labels mentioned in that clause are retained. E.g. `node_cpu_seconds_total == on(cpu, instance, mode, __name__) node_cpu_seconds_total` will drop the `job` label from the result.

roidelapluie:
sounds like a breaking change.

brian-brazil:
Yeah, that'd be breaking. I'd need to think on this one.

brian-brazil:
`node_cpu_seconds_total == bool node_cpu_seconds_total` also leaves the metric name in place, that's definitely a bug.

juliusv:
So we've had bug fixes in the past that technically changed behavior, but since they were bug fixes they were deemed ok most of the time. Probably it depends on how much we expect people to rely on this behavior already and how breaking the change would likely be for those people. On the other hand if we don't fix it, we need to document that it's broken.

brian-brazil:
Okay, we've no unittests covering this. The documented behaviour is what makes sense consistency wise, so the question is how breaking is this. `__name__` I'm not worried about, however starting to return other labels could cause issues.

roidelapluie:
For the time being, users who are affected by this can get the desired behavior by adding group_left()

brian-brazil:
True, we could define this to be the correct behaviour and leave that as a workaround.

juliusv:
[...]

PR #5494, opened 2019-04-22
Fix binop output labels when using on()/ignoring()

juliusv:
The entire label set from the LHS should be returned, not just the
supplied matching labels.

Fixes https://github.com/prometheus/prometheus/issues/5326

### DIFF
--- a/promql/testdata/operators.test
+++ b/promql/testdata/operators.test
@@ -176,10 +176,10 @@ eval instant at 50m http_requests{group="canary"} unless on(job, instance) http_
 	http_requests{group="canary", instance="1", job="app-server"} 800
 
 eval instant at 50m http_requests{group="canary"} / on(instance,job) http_requests{group="production"}
-	{instance="0", job="api-server"} 3
-	{instance="0", job="app-server"} 1.4
-	{instance="1", job="api-server"} 2
-	{instance="1", job="app-server"} 1.3333333333333333
+	{group="canary", instance="0", job="api-server"} 3
+	{group="canary", instance="0", job="app-server"} 1.4
+	{group="canary", instance="1", job="api-server"} 2
+	{group="canary", instance="1", job="app-server"} 1.3333333333333333
 
 eval instant at 50m http_requests{group="canary"} unless ignoring(group, instance) http_requests{instance="0"}
 
@@ -188,10 +188,10 @@ eval instant at 50m http_requests{group="canary"} unless ignoring(group) http_re
 	http_requests{group="canary", instance="1", job="app-server"} 800
 
 eval instant at 50m http_requests{group="canary"} / ignoring(group) http_requests{group="production"}
-	{instance="0", job="api-server"} 3
-	{instance="0", job="app-server"} 1.4
-	{instance="1", job="api-server"} 2
-	{instance="1", job="app-server"} 1.3333333333333333
+	{group="canary", instance="0", job="api-server"} 3
+	{group="canary", instance="0", job="app-server"} 1.4
+	{group="canary", instance="1", job="api-server"} 2
+	{group="canary", instance="1", job="app-server"} 1.3333333333333333
 
 # https://github.com/prometheus/prometheus/issues/1489
 eval instant at 50m http_requests AND ON (dummy) vector(1)
@@ -398,7 +398,7 @@ load 5m
 
 # On with no labels, for metrics with no common labels.
 eval instant at 5m random + on() metricA
-  {} 5
+  {foo="bar"} 5
 
 # Ignoring with no labels is the same as no ignoring.
 eval instant at 5m metricA + ignoring() metricB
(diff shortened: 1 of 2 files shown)


promql/testdata/operators.test
```diff
@@ -398,7 +398,7 @@ load 5m
 
 # On with no labels, for metrics with no common labels.
 eval instant at 5m random + on() metricA
-  {} 5
+  {foo="bar"} 5
```
brian-brazil: This isn't right, the fixed logic should only apply to comparison operators.

juliusv: Hmm, though #5326 was only about comparison operators I implicitly ended up assuming that it should behave the same for arithmetic operators (though the current behavior feels the most wrong in the filtering case). Unfortunately for arithmetic operators we also never specified the behavior properly in the docs at https://prometheus.io/docs/prometheus/latest/querying/operators/#arithmetic-binary-operators.

Naively I would expect the matching clause (`on()` / `ignoring()`) to be orthogonal to what labels are included in the output? I can see how it is slightly arbitrary to take the LHS vs. the RHS labels then, but then at least it'd be consistent with filtering ops?

brian-brazil: > Naively I would expect the matching clause (on() / ignoring()) to be orthogonal to what labels are included in the output?

For me the matching indicates the output labels, as that's the group you're working with.

juliusv: Interesting. That is a different viewpoint from where I'm coming from, which is more like treating `on()` like the `WHERE` in a SQL join and that being independent of the `SELECT *`.

brian-brazil: If you want that behaviour, there's group_left.

juliusv: True. But should it not be the default? I think of using `group_left` usually only in situations where there's more actual dimensionality/cardinality on the left side.

brian-brazil: That'd be a pretty massive breaking change.

juliusv: Hm yes. It's a breaking change also for filtering, but that's smaller, so we just accept it there? Just out of interest then, would we want to make this change if breaking-ness wouldn't be an issue (in Prometheus 3.0, say...)?

brian-brazil: For filtering it's a lot smaller, and in line with the intended behaviour. That's not so for general math.

juliusv: @brian-brazil I added a fixup commit to change arithmetic binops back to the old behavior. One thing I'm wondering about: if we have `CardOneToMany` for a compop, then the LHS/RHS are swapped and thus we choose the RHS labels for the output. But then the `group_x(...)` which always adds labels from the RHS (which is always the "many" side) is basically a no-op in that case, since all the RHS labels are already present. That's fine, right?

So cases like this, where the "one" side coincides with the actual initial LHS:

```
foo / on(x) group_right(y) bar
```

(`bar` would become the LHS, `foo` the RHS)

brian-brazil: I'd expect that to (usually) return an error as the `foo` is the labels that are returned, so will likely have duplicate series. I presume you meant to use a comparison operator in that example.

juliusv: Sorry, yes, I meant to use a comparison operator. Good point about the error.

For:

```
sum without(cpu) (node_cpu_seconds_total) == on(mode) group_right(job) node_cpu_seconds_total
```

I get `Error executing query: multiple matches for labels: grouping labels must ensure unique matches` with my change, while it works on `master`: http://demo.robustperception.io:9090/graph?g0.range_input=1h&g0.expr=sum%20without(cpu)%20(node_cpu_seconds_total)%20%3D%3D%20on(mode)%20group_right(job)%20node_cpu_seconds_total&g0.tab=1

It makes sense of course since now we don't add the additional `cpu` label from the right anymore. Maybe for `OneToMany` with comparison operators we still should, to not break things too much? Gah, this is getting complicated...

brian-brazil: > Gah, this is getting complicated...

Yeah, I've previously wondered if we should remove group_right. I'm now wondering if we want to change this logic at all.

juliusv: Yeah. Actually producing an error where there was none before is more breaking than I anticipated.

Of course one could say that in the case of one-to-many / many-to-one, we always choose the labels from the "many" side as the output, even if it happens to be on the RHS. So normally comp ops would always return the LHS labels, except for the one-to-many case.

Either that, or keep everything as is, I think.

brian-brazil: That'd be an exception on an exception, which is getting weird for something that should very rarely come up.

Our current behaviour is defendable on consistency grounds, so I'm tending towards leaving as-is currently.

juliusv: Yeah. In a greenfield world I'd lean towards just having this case produce an error, but given the breakingness of it I'll close it for now. Good to have explored this though.

brian-brazil: We should clarify the docs though.

juliusv: Here's an attempt: https://github.com/prometheus/prometheus/pull/5512

PR #5505, opened 2019-04-24
fix graph Y axis on stacked mode

### LINKED ISSUE #5017: Stacked graphs in /graph have increasingly incorrect max values
## Bug Report

Generated a stacked graph that was many small aggregates over a long window (a week).  As I increased the window size from 1 hour to 1 week, the max Y value of the graph was increasingly inaccurate, and the stacked data went from filling 90% of the min -> max Y value to filling only 5-10% of the vertical area of the graph.  There were no outlier values that I am aware of above that limit.  If I disable stacking on the long window, the series occupy most of the range of the vertical area.

I expected as I increased the window that stacked would continue to take up most of the Y range.

Unstacked at 2w

Stacked at 2w

Stacked at 1d

Stacked at 2h

It looks as if it is estimating the max Y incorrectly the larger the range.

**Environment**

* Prometheus version:

```
2.3.2
71af5e29e815795e9dd14742ee7725682fa14b7b
20180712-14:02:52
go1.10.3
```

shaikatz:
Closes #5017

We were accidentally summing up all the max Y of each time series instead of taking the highest Y value from all the series together.

@simonpasquier

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -443,7 +443,11 @@ func main() {
 				}
 				files = append(files, fs...)
 			}
-			return ruleManager.Update(time.Duration(cfg.GlobalConfig.EvaluationInterval), files)
+			return ruleManager.Update(
+				time.Duration(cfg.GlobalConfig.EvaluationInterval),
+				files,
+				cfg.GlobalConfig.ExternalLabels,
+			)
 		},
 	}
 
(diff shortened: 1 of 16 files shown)


web/ui/static/js/graph/index.js
```diff
@@ -711,13 +711,11 @@ Prometheus.Graph.prototype.updateGraph = function() {
     // When stacked is toggled
     var max = 0;
     self.data.forEach(function(timeSeries) {
-      var currSeriesMax = 0;
       timeSeries.data.forEach(function(dataPoint) {
-        if (dataPoint.y > currSeriesMax && dataPoint.y != null) {
-          currSeriesMax = dataPoint.y;
+        if (dataPoint.y > max && dataPoint.y != null) {
```
simonpasquier: Hmm I don't think it is correct... If you have 2 timeseries with values `[t0=1, t1=2, t3=3]` and `[t1=2, t2=2, t3=2]`, the computed max value would be `3` while it should be `5`.

PR #5515, opened 2019-04-25
More promlint rules

### LINKED ISSUE #5453: Expand promlint rules
## Proposal

Right now there are prometheus naming best practices documented in both [the docs](https://prometheus.io/docs/practices/naming/) and blog posts such as [Robust Perception](https://www.robustperception.io/on-the-naming-of-things). I propose that as many of these conventions as possible should be codified in the `promtool check metrics` tool.

Currently the promlint code includes rules for:
* Help text
* Using base units
* Using misleading/inappropriate suffixes

I propose that in addition there should be rules for:
* Don't put the metric type in the name, such as `request_duration_seconds_histogram`
* Don't use colons in metric names (reserved for recording rules)
* Use `snake_case` not `camelCase`
* Detection on `_s_` instead of `seconds`, `_b_` or `_kb_` instead of `_bytes_`

I would love to hear other people's opinion on rules that should be added and am happy to change the list around.

If people think some or all of these new rules are a good idea I plan to do the implementation.

Thanks for thinking about this proposal!

ellisda:
Here are a few extra linting rules for metric names, based on #5453.

Closes #5453 

Signed-off-by: David Ellis <ellisda@gmail.com>


util/promlint/promlint_test.go
```diff
@@ -478,7 +444,116 @@ go_gc_duration_seconds_count 5962
...
+		genTest("instance_memory_limit_pb"),
+
+		genTest("request_duration_s"),
+		genTest("request_duration_sec"),
+		genTest("request_duration_sec_summary"),
+	}
+	runTests(t, tests)
+}
 
+func runTests(t *testing.T, tests []test) {
```
simonpasquier: You can add [`t.Helper()`](https://golang.org/pkg/testing/#T.Helper) at the beginning of the function.

ellisda: add t.Helper() as suggested

rscot231:
@simonpasquier friendly bump on behalf of @ellisda. He and I worked together on this.

util/promlint/promlint.go
```diff
@@ -205,6 +210,54 @@ func lintHistogramSummaryReserved(mf dto.MetricFamily) []Problem {
 	return problems
 }
 
+// lintMetricTypeInName detects when the type of the metric is included in the metric name
```
simonpasquier: Missing dot at the end of the sentence and the same for the other comments.

util/promlint/promlint.go
```diff
@@ -205,6 +210,54 @@ func lintHistogramSummaryReserved(mf dto.MetricFamily) []Problem {
...
+func lintMetricTypeInName(mf dto.MetricFamily) []Problem {
+	t := mf.GetType()
+	if t == dto.MetricType_UNTYPED {
+		return nil
+	}
+
+	var problems problems
+	n := strings.ToLower(mf.GetName())
+
+	typename := strings.ToLower(dto.MetricType_name[int32(t)])
```
simonpasquier: `dto.MetricType_name[int32(t)]` could be simplified to `t.String()`

rscot231: Fixed.

util/promlint/promlint.go
rscot231: Fixed.

util/promlint/promlint.go
```diff
@@ -205,6 +210,54 @@ func lintHistogramSummaryReserved(mf dto.MetricFamily) []Problem {
...
+	return problems
+}
+
+// lintUnitAbbreviations detects abbreviated units in the metric name.
+func lintUnitAbbreviations(mf dto.MetricFamily) []Problem {
+	var problems problems
+	n := strings.ToLower(mf.GetName())
+	for _, s := range unitAbbreviations {
+		if strings.Contains(n, "_"+s+"_") || strings.HasSuffix(n, "_"+s) {
+			problems.Add(mf, "metric names should not used abbreviated units")
```
MarkusTeufelberger:
```suggestion
			problems.Add(mf, "metric names should not contain abbreviated units")
```

util/promlint/promlint.go
```diff
@@ -265,4 +318,16 @@ var (
 		"peta",
 		"pebi",
 	}
+
+	// Common abbreviations that we'd like to discourage.
+	unitAbbreviations = []string{
```
MarkusTeufelberger: Maybe abbreviations for minutes, hours, days, weeks, milliseconds, microseconds and nanoseconds could/should be added to the list?

csmarchbanks: I think adding `min`, `ms`, `us`, and `ns` would be a good idea. I think hours, days, and weeks are pretty rare.

csmarchbanks: Though minutes, hours, days, weeks are not base units. Perhaps there should be a different check for non-base units? That should also perhaps be addressed in a separate PR since there are lots of non base units.

util/promlint/promlint.go
ellisda: Fixed. (also, I had to to overwrite your commit to add sign-off as per DCO)

util/promlint/promlint.go
ellisda: I agree with the non-base unit argument. I've added `ms, us, ns` and omitted `min, hr, day, week`

util/promlint/promlint.go
```diff
@@ -205,6 +210,54 @@ func lintHistogramSummaryReserved(mf dto.MetricFamily) []Problem {
...
+	if strings.Contains(mf.GetName(), ":") {
+		problems.Add(mf, "metric names should not contain ':'")
+	}
+	return problems
+}
+
+// lintCamelCase detects metric names written in camelCase.
+func lintCamelCase(mf dto.MetricFamily) []Problem {
+	var problems problems
+	re := regexp.MustCompile(`[a-z][A-Z]`)
```
mdlayher: As a nit: I'd pull this out to a package level var like 'var camelCaseRe = regexp.MustCompile(`[a-z][A-Z]`)' and then use it in this function.

bcotton:
@brian-brazil, per our conversation at KubeCon, is there someone else who should approve this? Thanks

util/promlint/promlint.go
```diff
@@ -205,6 +210,54 @@ func lintHistogramSummaryReserved(mf dto.MetricFamily) []Problem {
...
+	if t == dto.MetricType_UNTYPED {
+		return nil
+	}
+
+	var problems problems
+	n := strings.ToLower(mf.GetName())
+
+	typename := strings.ToLower(t.String())
+	if strings.Contains(n, "_"+typename+"_") || strings.HasSuffix(n, "_"+typename) {
+		problems.Add(mf, fmt.Sprintf(`%s metrics should not include the type in metric name`, typename))
```
brian-brazil: in the

util/promlint/promlint.go
```diff
@@ -205,6 +210,54 @@ func lintHistogramSummaryReserved(mf dto.MetricFamily) []Problem {
 	return problems
 }
 
+// lintMetricTypeInName detects when the type of the metric is included in the metric name.
+func lintMetricTypeInName(mf dto.MetricFamily) []Problem {
+	t := mf.GetType()
+	if t == dto.MetricType_UNTYPED {
```
brian-brazil: We could check for any type in any name

util/promlint/promlint.go
```diff
@@ -205,6 +210,54 @@ func lintHistogramSummaryReserved(mf dto.MetricFamily) []Problem {
...
+		problems.Add(mf, fmt.Sprintf(`%s metrics should not include the type in metric name`, typename))
+	}
+	return problems
+}
+
+// lintReservedChars detects colons in metric names.
+func lintReservedChars(mf dto.MetricFamily) []Problem {
+	var problems problems
+	if strings.Contains(mf.GetName(), ":") {
+		problems.Add(mf, "metric names should not contain ':'")
```
brian-brazil: metric names can contain :, but not from exporters/direct instrumentation. So it's fine from /federate for example (and indeed should be the case).

util/promlint/promlint.go
```diff
@@ -205,6 +210,54 @@ func lintHistogramSummaryReserved(mf dto.MetricFamily) []Problem {
...
+	}
+	return problems
+}
+
+// lintCamelCase detects metric names written in camelCase.
+func lintCamelCase(mf dto.MetricFamily) []Problem {
+	var problems problems
+	re := regexp.MustCompile(`[a-z][A-Z]`)
+	if re.FindString(mf.GetName()) != "" {
+		problems.Add(mf, "metric names should be written in 'snake_case' not 'camelCase'")
```
brian-brazil: label names too?

util/promlint/promlint.go
csmarchbanks: I completely agree. My thoughts are currently that this tool should be targeted for exporters/direct instrumentation so ':' should fail, but I do not hold that opinion tightly. Perhaps the description should contain more information?

util/promlint/promlint.go
ellisda: :+1: ,  added check for label names

util/promlint/promlint.go
ellisda: added check for all except untyped

util/promlint/promlint.go
ellisda: I want to keep the locality - definition near the use, but happy to pull out to package var

util/promlint/promlint.go
```diff
@@ -205,6 +210,64 @@ func lintHistogramSummaryReserved(mf dto.MetricFamily) []Problem {
...
+// lintReservedChars detects colons in metric names.
+func lintReservedChars(mf dto.MetricFamily) []Problem {
+	var problems problems
+	if strings.Contains(mf.GetName(), ":") {
+		problems.Add(mf, "metric names should not contain ':'")
+	}
+	return problems
+}
+
+var camelCase *regexp.Regexp = regexp.MustCompile(`[a-z][A-Z]`)
```
mdlayher:
```suggestion
var camelCase = regexp.MustCompile(`[a-z][A-Z]`)
```

ellisda: Thanks @mdlayher . Also, that was FAST response :+1:  ha ha

util/promlint/promlint.go
csmarchbanks: @brian-brazil I think this is the only unresolved conversation left. Do you feel strongly that linting should work with `/federate` or is it ok to target exporters/direct instrumentation?

csmarchbanks:
This has two approvals from Prometheus team members. Is there anything else that needs to be done before it can be merged?

Perhaps @simonpasquier or @mdlayher could take one final look?

util/promlint/promlint.go
brian-brazil: It should work with both, but mainly exporters. What we don't want is is continuous bugs filed by users asking why /federate doesn't pass.

util/promlint/promlint.go
brian-brazil: Hours and days aren't that unusual.

csmarchbanks: I agree that they are not unusual, and should be linted. However, the linter does not check for any sort of non-base unit right now. To be complete we should also check for things like fahrenheit, kelvin, rankine, inches, yards, bits, calories, pounds, ounces, etc. Is it ok to do that as part of a separate PR?

brian-brazil: We're already checking for MB and kB, how is that different from minutes, hours and days?

csmarchbanks: Fair point, I expanded the unit linting to check several non base units.

util/promlint/promlint.go
csmarchbanks: Right now it looks like /federate will not pass the linting in master due to /federate not presenting help text. Since there are not continuous bugs filed by users right now I don't think it will be a problem, but I can remove the `:` check if you feel better about that.

brian-brazil: We can leave it in then

util/promlint/promlint.go
```diff
@@ -232,17 +295,41 @@ func metricUnits(m string) (unit string, base string, ok bool) {
...
-		"moles",
-		"seconds",
+	// map a unit to the appropriate base unit.
+	units = map[string]string{
+		// Base units.
+		"amperes": "amperes",
+		"bytes":   "bytes",
+		"candela": "candela",
+		"grams":   "grams",
+		"celsius": "celsius", // Allow both kelvin and celsius.
```
brian-brazil: Celsius is the base unit here.

util/promlint/promlint.go
```diff
@@ -232,17 +295,41 @@ func metricUnits(m string) (unit string, base string, ok bool) {
...
+		"kelvin":  "kelvin",  // Both plural and non-plural form allowed.
+		"kelvins": "kelvins",
+		"meters":  "meters", // Both American and international spelling permitted.
+		"metres":  "metres",
+		"moles":   "moles",
```
brian-brazil: Wouldn't the base unit be the number of molecules?

Time and data size are the main ones, I wouldn't worry about the rest too much.

MarkusTeufelberger: https://en.wikipedia.org/wiki/Mole_(unit)

> The mole is the base unit of amount of substance ("number of substance") in the International System of Units or System International (SI) [...]

util/promlint/promlint.go
csmarchbanks: :+1:

util/promlint/promlint.go
brian-brazil: That's basically a prefix, and we don't use the SI unit for mass for example as it has a prefix.

I don't expect many users to be doing chemistry with Prometheus, let's not get beyond what you're likely to see in a standard datacenter.

util/promlint/promlint.go
```diff
@@ -232,17 +295,43 @@ func metricUnits(m string) (unit string, base string, ok bool) {
...
+		"rankine":    "celsius",
+		// Length.
+		"inches": "meters",
+		"yards":  "meters",
+		"miles":  "meters",
+		// Bytes.
+		"bits": "bytes",
+		// Energy.
+		"calories": "joules",
+		// Weight.
```
brian-brazil: Technically weight is newtons :)

util/promlint/promlint.go
csmarchbanks: Does it seem reasonable to lint the classes listed in https://prometheus.io/docs/practices/naming/#base-units and delete the rest (in this case moles and candela)?

util/promlint/promlint.go
csmarchbanks: My physics professors probably all just disowned me :smile:.

Shall I also make the docs technically correct?

csmarchbanks:
Are there any concerns around merging this as is? If any issues crop up in the future we are happy to iterate on it.

util/promlint/promlint.go
brian-brazil: Yes

csmarchbanks: :+1: I removed moles and candela.

brian-brazil:
:+1:

PR #5525, opened 2019-04-30
web/ui: Consistent Content-Type for js and css

mrasu:
Currently, When `/etc/mime.types` has a unusual mime type, web of prometheus uses the type and you may get unexpected result.
With this change, web returns consistent Content-Type header for static js and css files

To reproduce:
1. Add a type at the end of `/etc/mime` like `text/x-js js`
2. Run prometheus
3. Request js file like `http://localhost:9090/static/vendor/js/jquery.min.js`
4. You will see Content-Type of the response is `text/x-js` instead of `application/javascript`

Relates to https://github.com/prometheus/prometheus/issues/4246

Signed-off-by: mrasu <m.rasu.hitsuji@gmail.com>

### DIFF
--- a/web/web.go
+++ b/web/web.go
@@ -294,11 +294,7 @@ func New(logger log.Logger, o *Options) *Handler {
 
 	router.Get("/consoles/*filepath", readyf(h.consoles))
 
-	router.Get("/static/*filepath", func(w http.ResponseWriter, r *http.Request) {
-		r.URL.Path = path.Join("/static", route.Param(r.Context(), "filepath"))
-		fs := http.FileServer(ui.Assets)
-		fs.ServeHTTP(w, r)
-	})
+	router.Get("/static/*filepath", h.static)
 
 	if o.UserAssetsPath != "" {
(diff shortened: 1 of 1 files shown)


mrasu:
Because simulating a change of `etc/mime.types` is not easy for me and existing tests for web doesn't test behavior deeply, I believe no test is needed.
But if it is needed, I will

simonpasquier:
Maybe this should be moved to https://github.com/prometheus/common/ as AlertManager and pushgateway also rely on `http.FileServer`. @brian-brazil WDYT?

brian-brazil:
Makes sense, pgw probably needs it too.

@beorn7 FYI

mrasu:
Thanks for the feedback!
I made at common https://github.com/prometheus/common/pull/191

beorn7:
Thanks for the heads up. I'll follow the implementation in prometheus/common and will adopt it in the PGW once done.

brian-brazil:
:+1:

beorn7:
prometheus/common got a new release, so I guess it's best to close this PR (or rework into one making prometheus/prometheus use the file server from prometheus/common).

PR #5527, opened 2019-04-30
Add more noop types

jacksontj:

### DIFF
--- a/storage/noop.go
+++ b/storage/noop.go
@@ -14,10 +14,55 @@
 package storage
 
 import (
+	"context"
 	"math"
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
Is there more to this PR? We only accept code that we need ourselves, so if you only need these in your repo you should keep it there.

jacksontj:
I need them in mine, I figured they'd potentially be useful here -- since there are other noops there. If you don't find these helpful feel free to close this PR.

brian-brazil:
If we need them for unittests at some point, we can always add them in then.

PR #5535, opened 2019-05-06
some minor fixes:

YaoZengzeng:
1. If alerts is empty after `relabelAlerts`, just return to avoid
subsequent unnecessary operations

2. minor fix in notifier's test case

3. minor fix in comment


brian-brazil:
Thanks!

PR #5544, opened 2019-05-08
import aliases that can be omitted

aixeshunter:
import aliases that can be omitted, like `kingpin `, `conntrack`, ...

Signed-off-by: aixeshunter <aixeshunter@gmail.com>


brian-brazil:
I think we should leave these as-is for readability, what Golang is using as the name isn't exactly obvious in these cases.

cmd/promtool/main.go
```diff
@@ -29,11 +29,11 @@ import (
 	"github.com/google/pprof/profile"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/api"
-	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
```
beorn7: I tend to agree with what @brian-brazil said, but this alias can actually go.

aixeshunter: I agree it.

tariq1890:
I agree with @brian-brazil. This is simply how `goimports` functions in the latest go runtimes.

roidelapluie:
Can you please rebase this PR? Thanks!

bwplotka:
Sounds like goimports places this behaviour of requiring alias on "v.*" package names as it collides with module versions (https://github.com/golang/go/issues/30051). Closing this then. 

Thanks @aixeshunter anyway!

PR #5549, opened 2019-05-09
Remove bits about substring matching from the docs

semyonslepov:
Substring matching for labels doesn't work anymore.
These parts are misleading, and contradict with the rest of the
documentation.


brian-brazil:
Thanks!

PR #5551, opened 2019-05-09
Consistent Content-Type for js and css

mrasu:
Alternative of https://github.com/prometheus/prometheus/pull/5525

Currently, When `/etc/mime.types` has a unusual mime type, web of prometheus uses the type and you may get unexpected result.
With this change, web returns consistent Content-Type header for static js and css files

To reproduce:
1. Add a type at the end of `/etc/mime` like `text/x-js js`
2. Run prometheus
3. Request js file like `http://localhost:9090/static/vendor/js/jquery.min.js`
4. You will see Content-Type of the response is `text/x-js` instead of `application/javascript`

Relates to https://github.com/prometheus/prometheus/issues/4246

Signed-off-by: mrasu <m.rasu.hitsuji@gmail.com>


brian-brazil:
Thanks!

PR #5552, opened 2019-05-09
Correctly handle {__name__="a"}

brian-brazil:
This can cause problems in alerts.

### DIFF
--- a/promql/printer.go
+++ b/promql/printer.go
@@ -179,8 +179,8 @@ func (node *UnaryExpr) String() string {
 func (node *VectorSelector) String() string {
 	labelStrings := make([]string, 0, len(node.LabelMatchers)-1)
 	for _, matcher := range node.LabelMatchers {
-		// Only include the __name__ label if its no equality matching.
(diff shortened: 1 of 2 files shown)


promql/printer.go
```diff
@@ -180,7 +180,7 @@ func (node *VectorSelector) String() string {
 	labelStrings := make([]string, 0, len(node.LabelMatchers)-1)
 	for _, matcher := range node.LabelMatchers {
 		// Only include the __name__ label if its no equality matching.
```
beorn7: Perhaps amend the comment with the new additional condition? (Something like: "... or it is a selection via `__name__=`.)

brian-brazil: Good point, fixed.

juliusv:
Makes me wonder if we should keep the `.Name` field on selector nodes in the first place, since it's redundant and introduces the possibility of a mismatch. Or do we care that `{__name__="a"}` gets rendered out as that string again vs. always being normalized to `a`.

brian-brazil:
Yeah, the code might need a rethink but I just wanted to fix the immediate problem here.

PR #5555, opened 2019-05-10
Fix remote storage config not updating correctly

csmarchbanks:
This PR fixes some issues introduced when code was put in to not update the remote write queues upon a config reload (#5485).

1. Remote read config is not updated unless you also update the remote write config.
1. Updating external labels does not add them to the remote write queues.


csmarchbanks:
@brian-brazil could you take a look at this when you get a chance please?

brian-brazil:
Thanks!

PR #5564, opened 2019-05-14
Remote write sidecar

### LINKED ISSUE #5539: Add ability to run remote write as a sidecar
## Proposal

Now that remote write is reading from the WAL, there is no requirement for remote write to share a process with Prometheus. Running remote write as a sidecar would have the following advantages:

1. A crash in remote write will not take down all of Prometheus
1. Ability to apply different resource constraints to each component

If a sidecar is desired, would it be better to have a separate binary or a mode from within the Prometheus binary?

Also, to be clear I am proposing this as an additive change, not a replacement of the in process mode.

cc: @cstyan curious what you think since you did the WAL reading work.

csmarchbanks:
This PR adds a binary that implements the remote write protocol that can be run as a sidecar to the Prometheus process.

The configuration will be identical to the current remote write configuration, the biggest difference being that the sample rate for calculating sharding will come from a second WALWatcher running instead of from samples being scraped.

Closes #5539 

@gouthamve I would love a review from you to make sure this would work for your customers as well.

Current state: Everything is working, I have been running it in a cluster without issue for a couple of days now without issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 
 /prometheus
 /promtool
+/remotewrite
 benchmark.txt
 /data
 /cmd/prometheus/data
--- a/.promu.yml
+++ b/.promu.yml
@@ -10,6 +10,8 @@ build:
           path: ./cmd/prometheus
         - name: promtool
           path: ./cmd/promtool
+        - name: remotewrite
+          path: ./cmd/remotewrite
     flags: -mod=vendor -a -tags netgo
     ldflags: |
         -X github.com/prometheus/common/version.Version={{.Version}}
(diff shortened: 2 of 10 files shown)


csmarchbanks:
@gouthamve / @tomwilkie / @cstyan Now that KubeCon is over is anyone interested in taking a look at this?

csmarchbanks:
Thanks for the review! I will take a look at having a new type rather than a flag.

Another option for where the `main.go` lives would be the (soon to be?) newly revived [prometheus-community](https://github.com/prometheus-community). Then just have this PR do the refactoring to make it easier to use from there.

My goal is to have an easy to use image published somewhere, but I am not attached to it living in this repo. I think the only downside of having it in a separate repo is making sure to stay on top of updates to the remote write code, but that is not a huge deal.

cstyan:
It's worth pointing out that the `WALWatcher` code will move into the WAL package in the near future as well.

csmarchbanks:
I am closing this PR in favor of just doing the refactor and providing the main code/binary in a separate repo. See https://github.com/prometheus/prometheus/pull/5624

PR #5568, opened 2019-05-15
Fix minor punctuation and language issues in flag doc strings

beorn7:
This is mostly to create consistency, not because the one or the other
way would be wrong. A few actual corrections are also included.

My OCD commands me to do this…

I guess @juliusv is the one who will appreciate this most.


juliusv:
:+1:

PR #5569, opened 2019-05-15
ignore deallocated VMs during azure discovery

### LINKED ISSUE #4340: Azure Discovery incorrectly discovering de-allocated VM NICs
Hi

Looks like the Azure Discovery code is using the IP address to determine whether a VM is de-allocated or not.  Dynamic Private IP addresses are not released back to the pool when a VM is deallocated.  The comment in the code reads

```
// Unfortunately Azure does not return information on whether a VM is deallocated.
// This information is available via another API call however the Go SDK does not
// yet support this. On deallocated machines, this value happens to be nil so it
// is a cheap and easy way to determine if a machine is allocated or not.
```

This is no longer the case.  It would make more sense now to use the `MacAddress` property of the NIC, as this will is NULL when the VM is allocated.

dmitsh:
This PR removes deallocated VMs (both individual VMs and the instances of VMSS) from the target list during azure discovery.
Closes #4340

### DIFF
--- a/discovery/azure/azure.go
+++ b/discovery/azure/azure.go
@@ -54,6 +54,9 @@ const (
 
 	authMethodOAuth           = "OAuth"
 	authMethodManagedIdentity = "ManagedIdentity"
+
+	powerStatePrefix  = "PowerState/"
+	powerStateRunning = powerStatePrefix + "running"
 )
 
 // DefaultSDConfig is the default Azure SD configuration.
@@ -147,10 +150,11 @@ type azureClient struct {
 	vm     compute.VirtualMachinesClient
 	vmss   compute.VirtualMachineScaleSetsClient
 	vmssvm compute.VirtualMachineScaleSetVMsClient
+	logger log.Logger
 }
 
 // createAzureClient is a helper function for creating an Azure compute client to ARM.
-func createAzureClient(cfg SDConfig) (azureClient, error) {
+func createAzureClient(cfg SDConfig, logger log.Logger) (azureClient, error) {
 	env, err := azure.EnvironmentFromName(cfg.Environment)
 	if err != nil {
 		return azureClient{}, err
@@ -200,13 +204,16 @@ func createAzureClient(cfg SDConfig) (azureClient, error) {
 	c.vmssvm = compute.NewVirtualMachineScaleSetVMsClientWithBaseURI(resourceManagerEndpoint, cfg.SubscriptionID)
 	c.vmssvm.Authorizer = bearerAuthorizer
 
+	c.logger = logger
+
 	return c, nil
 }
 
 // azureResource represents a resource identifier in Azure.
 type azureResource struct {
-	Name          string
 	ResourceGroup string
+	Name          string
+	Subname       string
 }
 
 // virtualMachine represents an Azure virtual machine (which can also be created by a VMSS)
@@ -234,16 +241,22 @@ func newAzureResourceFromID(id string, logger log.Logger) (azureResource, error)
 		return azureResource{}, err
 	}
 
+	var subname string
+	if len(s) == 11 {
+		subname = strings.ToLower(s[10])
+	}
+
 	return azureResource{
-		Name:          strings.ToLower(s[8]),
 		ResourceGroup: strings.ToLower(s[4]),
+		Name:          strings.ToLower(s[8]),
+		Subname:       subname,
 	}, nil
 }
 
 func (d *Discovery) refresh(ctx context.Context) ([]*targetgroup.Group, error) {
 	defer level.Debug(d.logger).Log("msg", "Azure discovery completed")
 
-	client, err := createAzureClient(*d.cfg)
+	client, err := createAzureClient(*d.cfg, d.logger)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not create Azure client")
 	}
@@ -324,18 +337,9 @@ func (d *Discovery) refresh(ctx context.Context) ([]*targetgroup.Group, error) {
 					continue
 				}
 
-				// Unfortunately Azure does not return information on whether a VM is deallocated.
-				// This information is available via another API call however the Go SDK does not
-				// yet support this. On deallocated machines, this value happens to be nil so it
-				// is a cheap and easy way to determine if a machine is allocated or not.
-				if networkInterface.Primary == nil {
-					level.Debug(d.logger).Log("msg", "Skipping deallocated virtual machine", "machine", vm.Name)
-					return
-				}
-
 				if *networkInterface.Primary {
 					for _, ip := range *networkInterface.IPConfigurations {
-						if ip.PublicIPAddress != nil {
+						if ip.PublicIPAddress != nil && ip.PublicIPAddress.PublicIPAddressPropertiesFormat != nil {
 							labels[azureLabelMachinePublicIP] = model.LabelValue(*ip.PublicIPAddress.IPAddress)
 						}
 						if ip.PrivateIPAddress != nil {
@@ -372,6 +376,48 @@ func (d *Discovery) refresh(ctx context.Context) ([]*targetgroup.Group, error) {
 	return []*targetgroup.Group{&tg}, nil
 }
 
+// isRunningVM return true if the VM is up and running
+// ref: https://docs.microsoft.com/en-us/azure/virtual-machines/windows/states-lifecycle
+func (client *azureClient) isRunningVM(ctx context.Context, vm *compute.VirtualMachine) bool {
+	r, err := newAzureResourceFromID(*vm.ID, client.logger)
+	if err != nil {
+		return false
+	}
+	instanceView, err := client.vm.InstanceView(ctx, r.ResourceGroup, *vm.Name)
+	if err != nil {
+		level.Error(client.logger).Log("msg", "failed to get InstanceView", "VM name", *vm.Name, "err", err)
+		return false
+	}
+	for _, status := range *instanceView.Statuses {
+		if strings.HasPrefix(*status.Code, powerStatePrefix) && strings.Compare(*status.Code, powerStateRunning) != 0 {
+			level.Info(client.logger).Log("msg", "virtual machine is not running", "VM name", *vm.Name, "status", *status.Code)
+			return false
+		}
+	}
+	return true
+}
+
+// isRunningScaleSetVM return true if the VM is up and running
+// ref: https://docs.microsoft.com/en-us/azure/virtual-machines/windows/states-lifecycle
+func (client *azureClient) isRunningScaleSetVM(ctx context.Context, vmss *compute.VirtualMachineScaleSet, vm *compute.VirtualMachineScaleSetVM) bool {
+	r, err := newAzureResourceFromID(*vm.ID, client.logger)
+	if err != nil {
+		return false
+	}
+	instanceView, err := client.vmssvm.GetInstanceView(ctx, r.ResourceGroup, *vmss.Name, r.Subname)
+	if err != nil {
+		level.Error(client.logger).Log("msg", "failed to get InstanceView", "VMSS name", *vmss.Name, "VM name", *vm.Name, "err", err)
+		return false
+	}
+	for _, status := range *instanceView.Statuses {
+		if strings.HasPrefix(*status.Code, powerStatePrefix) && strings.Compare(*status.Code, powerStateRunning) != 0 {
+			level.Info(client.logger).Log("msg", "virtual machine is not running", "VM name", *vm.Name, "status", *status.Code)
+			return false
+		}
+	}
+	return true
+}
+
 func (client *azureClient) getVMs(ctx context.Context) ([]virtualMachine, error) {
 	var vms []virtualMachine
 	result, err := client.vm.ListAll(ctx)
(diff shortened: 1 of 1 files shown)


discovery/azure/azure.go
```diff
@@ -324,18 +337,9 @@ func (d *Discovery) refresh(ctx context.Context) ([]*targetgroup.Group, error) {
...
-				// is a cheap and easy way to determine if a machine is allocated or not.
-				if networkInterface.Primary == nil {
-					level.Debug(d.logger).Log("msg", "Skipping deallocated virtual machine", "machine", vm.Name)
-					return
-				}
-
 				if *networkInterface.Primary {
 					for _, ip := range *networkInterface.IPConfigurations {
-						if ip.PublicIPAddress != nil {
+						if ip.PublicIPAddress != nil && ip.PublicIPAddress.PublicIPAddressPropertiesFormat != nil {
```
simonpasquier: This should go in another PR for clarity. See #5570

discovery/azure/azure.go
```diff
@@ -372,6 +376,48 @@ func (d *Discovery) refresh(ctx context.Context) ([]*targetgroup.Group, error) {
...
+		return false
+	}
+	instanceView, err := client.vm.InstanceView(ctx, r.ResourceGroup, *vm.Name)
+	if err != nil {
+		level.Error(client.logger).Log("msg", "failed to get InstanceView", "VM name", *vm.Name, "err", err)
+		return false
+	}
+	for _, status := range *instanceView.Statuses {
+		if strings.HasPrefix(*status.Code, powerStatePrefix) && strings.Compare(*status.Code, powerStateRunning) != 0 {
+			level.Info(client.logger).Log("msg", "virtual machine is not running", "VM name", *vm.Name, "status", *status.Code)
```
simonpasquier: This should be debug

discovery/azure/azure.go
```diff
@@ -372,6 +376,48 @@ func (d *Discovery) refresh(ctx context.Context) ([]*targetgroup.Group, error) {
...
+	r, err := newAzureResourceFromID(*vm.ID, client.logger)
+	if err != nil {
+		return false
+	}
+	instanceView, err := client.vm.InstanceView(ctx, r.ResourceGroup, *vm.Name)
+	if err != nil {
+		level.Error(client.logger).Log("msg", "failed to get InstanceView", "VM name", *vm.Name, "err", err)
+		return false
+	}
+	for _, status := range *instanceView.Statuses {
```
simonpasquier: check for `instanceView.Statuses == nil` before?

discovery/azure/azure.go
```diff
@@ -372,6 +376,48 @@ func (d *Discovery) refresh(ctx context.Context) ([]*targetgroup.Group, error) {
...
+		if strings.HasPrefix(*status.Code, powerStatePrefix) && strings.Compare(*status.Code, powerStateRunning) != 0 {
```
simonpasquier: check for `status.Code == nil`. I'm also not sure why we would need to check both for the prefix and the full string: `if status.Code != nil && *status.Code == powerStateRunning {` should be enough.

discovery/azure/azure.go
```diff
@@ -372,6 +376,48 @@ func (d *Discovery) refresh(ctx context.Context) ([]*targetgroup.Group, error) {
...
+			level.Info(client.logger).Log("msg", "virtual machine is not running", "VM name", *vm.Name, "status", *status.Code)
+			return false
+		}
+	}
+	return true
+}
+
+// isRunningScaleSetVM return true if the VM is up and running
+// ref: https://docs.microsoft.com/en-us/azure/virtual-machines/windows/states-lifecycle
+func (client *azureClient) isRunningScaleSetVM(ctx context.Context, vmss *compute.VirtualMachineScaleSet, vm *compute.VirtualMachineScaleSetVM) bool {
```
simonpasquier: Same remarks as for `isRunningVM()`.

discovery/azure/azure.go
```diff
@@ -200,13 +204,16 @@ func createAzureClient(cfg SDConfig) (azureClient, error) {
 	c.vmssvm = compute.NewVirtualMachineScaleSetVMsClientWithBaseURI(resourceManagerEndpoint, cfg.SubscriptionID)
 	c.vmssvm.Authorizer = bearerAuthorizer
 
+	c.logger = logger
```
simonpasquier: This could go at L166 where `c` is declared.

discovery/azure/azure.go
brian-brazil: For EC2 we just expose the state, without filtering it as depending on use case there you might want to notice that an instance is stopped.

tariq1890:
@dmitsh  I would advocate for tweaking the approach here a bit so that the power state can be added as a prometheus label. You can check this PR (it was reverted) https://github.com/prometheus/prometheus/pull/4908 for more details. This PR did not work because the `InstanceView` field of the Azure SDK object was always nil and depended on the `GetInstanceView` API which you have implemented in this PR.

dmitsh:
> @dmitsh I would advocate for tweaking the approach here a bit so that the power state can be added as a prometheus label. You can check this PR (it was reverted) #4908 for more details. This PR did not work because the `InstanceView` field of the Azure SDK object was always nil and depended on the `GetInstanceView` API which you have implemented in this PR.

I don't have strong opinion, but we cannot collect any metrics from non-running nodes (beside the fact that they are not running). You can always get the list of running nodes from the target endpoint.
This PR doesn't change existing logics, but fixes a bug.
Let's open a separate issue whether we want to discover nodes in a bad state.

brian-brazil:
> we cannot collect any metrics from non-running nodes (beside the fact that they are not running).

That is a relevant fact, and a node in the process of shutting down may have metrics too.

tariq1890:
I was not suggesting that we collect metrics from non-running nodes. Azure has multiple power states iiuc and I thought it'd be useful to potentially query across the various states of VMs.

dmitsh:
OK, let me add power state label

dmitsh:
@simonpasquier @brian-brazil @tariq1890 
I've refactored the code to add VM power state label, as was intended in #4908, and successfully tested it against real Azure VM/VMSS.
Please take another look.

discovery/azure/azure_test.go
```diff
@@ -201,3 +277,50 @@ func TestMapFromVMScaleSetVMWithTags(t *testing.T) {
...
+		VirtualMachineScaleSetVMProperties: properties,
+	}
+
+	actual := getPowerState(testVM.VirtualMachineScaleSetVMProperties.InstanceView.Statuses)
+	expected := "running"
+	if actual != expected {
+		t.Errorf("expected powerStatus %s, but got %s instead", expected, actual)
+	}
+
+	// Noq we test a virtualMachine with an empty InstanceView struct.
```
simonpasquier: s/Noq/Now/?

discovery/azure/azure_test.go
```diff
@@ -201,3 +277,50 @@ func TestMapFromVMScaleSetVMWithTags(t *testing.T) {
 		t.Errorf("Expected %v got %v", expectedVM, actualVM)
 	}
 }
+
+func TestGetPowerStatusFromVM(t *testing.T) {
```
simonpasquier: I'd use [table-driven tests](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) here as they are easier to read/maintain IMO.

discovery/azure/azure.go
```diff
@@ -288,6 +309,13 @@ func (d *Discovery) refresh(ctx context.Context) ([]*targetgroup.Group, error) {
 				return
 			}
 
+			// Skip service discovery for deallocated VM
```
simonpasquier: Missing dot at the end of the sentence.

discovery/azure/azure.go
```diff
@@ -288,6 +309,13 @@ func (d *Discovery) refresh(ctx context.Context) ([]*targetgroup.Group, error) {
...
+			if strings.EqualFold(vm.PowerState, powerStateDeallocated) {
+				level.Debug(d.logger).Log("msg", "Skipping virtual machine", "VM name", vm.Name, "power state", vm.PowerState)
+				ch <- target{}
```
simonpasquier: this line can be removed.

discovery/azure/azure.go
```diff
@@ -380,6 +442,14 @@ func (client *azureClient) getVMs(ctx context.Context) ([]virtualMachine, error)
 	}
 	for result.NotDone() {
 		for _, vm := range result.Values() {
+			if vm.InstanceView == nil {
```
simonpasquier: It might be better to do this in the `for i, vm := range machines { ... }` loop as the calls to the Azure API will be done in parallel instead of sequentially here.

brian-brazil: Is there a way to do this without additional RPCs?

Azure has really low api limits, and so far we've had to reject all PRs that added more RPCs due to this.

dmitsh: @simonpasquier We cannot do it inside `for i, vm := range machines { ... }` because `machines` contains both individual VMs and VMSS' VMs. There are different API calls to get InstanceView for each of the types.

@brian-brazil Unfortunately that's the way Azure has implemented it. I will keep monitoring the updates for azure-sdk-for-go, in case they would address this issue.

I've resolved the rest of the comments and updated the PR. Please take a look.

brian-brazil: This PR proposes to roughly double the RPCs we send to Azure, which is a showstopper unfortunately as it halves the (already relatively small) number of machines that can be monitored. Is there any other way to get this, or cut other RPCs?

dmitsh: @brian-brazil 
> This PR proposes to roughly double the RPCs we send to Azure
> Is there any other way to get this, or cut other RPCs?

To be precise, before this PR we were making one API call `result, err := client.vm.ListAll(ctx)` to get the list of all VMs.
In this PR I added iterating over the VMs and extracting InstanceView for every one of them.
I agree that this is not optimal, but `ListAll` does not populate InstanceView.
We have to make a separate API call for each VM.
There is an open issue in azure-sdk https://github.com/Azure/azure-sdk-for-python/issues/573
Azure team is supposed to resolve this problem at some point, but I don't know when.

I would suggest that we accept this PR, as it resolves a bug and provides additional metrics.
I will follow up with the azure team regarding their plans to fix this problem, and I will open a new PR once it's ready.

brian-brazil: We also need a call per machine to get network stats currently.

dmitsh: > We also need a call per machine to get network addresses currently.

Correct, I'm aware of that.
But there is currently no other way to get power state (and disregard deallocated VMs) without sending a separate InstanceView request for every VM.

brian-brazil: CloudNativeCon is this week, hopefully I can find someone with Azure to have a chat about this. This is far from the first PR we've had to reject on these grounds :(

dmitsh: Could you elaborate your concern about azure low API limits?

According to this doc (https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-request-limits), we can make 12,000 read API calls per hour.

The discovery runs every 5 minutes, or 12 times in an hour.
Assuming we have 100 VMs in the cluster, we would be changing the number of read API calls from 12*100=1200 to 12*2*100=2400.
Unless a customer has over 500 VMs, we are not hitting the limit.
What is the typical/maximal Azure cluster size, in your experience?

brian-brazil: 500 VMs is not that many, and 5m delay is already quite a lot. 1m is what we use for other SDs. When we initially chose those values we thought there were 15k requests per hour.

dmitsh: I totally understand.
The only alternative I'm aware of is the Azure Instance Metadata Service, but it is available on the VM itself, not from a remote endpoint.
https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service

brian-brazil: Had a chat with someone at the Azure stand. They didn't have anyone from that particular team there, but will put them in touch with me.

silencev: hi @brian-brazil , any update on this?
My team govern around 5500 VMs in Azure and located in 90+ subscriotions.
I'm thinking of having one additional REST API call to get power state of VM is fine.

brian-brazil:
Have you come up with a way of doing this that doesn't involve additional RPCs? Otherwise we'll unfortunately have to close this as infeasible.

Azure never got back to me.

dmitsh:
> Have you come up with a way of doing this that doesn't involve additional RPCs? Otherwise we'll unfortunately have to close this as infeasible.
> 
> Azure never got back to me.

Let me reach out to them again...

roidelapluie:
We have looked at this during our bug scrub and we can't merge this for now as it is making new RPC calls to azure and the rate limiting is quite low for Azure.

dmitsh:
@brian-brazil , this is a very old PR.
Is it still relevant?
I updated it to reduce number of API calls.

gburton1:
Just discovering this now...I think this is still nice to have. It makes alerting on the `up` metric for instances more manageable when all of the stopped and deallocated cruft can be filtered out somehow.

We have been struggling with the 12k subscription-wide Azure limits for a while (there are many things besides Prometheus in the system that use up the rate limit budget quickly), so it makes sense to be very careful when introducing new calls. It looks like you now get the statuses for all VMs in one call, which seems like it addressed earlier concerns 👍

gburton1:
Hi @brian-brazil @simonpasquier, do you think we have a clear path to merging this? It appears that past concerns about excessive Azure usage have been addressed. And this is the only thing blocking us from having very clean host level checks on the `up` metric of the service discovery targets. As it stands right now, we'll have too much noise from VMs that are shut down for normal reasons.

roidelapluie:
We have update the azure SDK. Can you check the current state, if we need to add a new RPC call or if the new SDK provides us the info without extra RPC calls.

PR #5574, opened 2019-05-16
minor doc improvement in migration.md sample

tariq1890:
This PR improves the documentation for running prometheus as a non-root user. Hoping to make it more beginner-friendly. Also the example seems to be inspired by [this](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container). So thought it would make sense to follow suit.


tariq1890:
FYI @brancz

PR #5576, opened 2019-05-17
Update prometheus/client_golang and prometheus/tsdb modules

beorn7:
Promtool test is failing. I still have to look into it.

Wanted to put this out already. Perhaps it's a known issue for somebody.


simonpasquier:
LGTM

beorn7:
@krasi-georgiev I'm fairly sure this is because the API cilent is now using POST. Is that intended/ok? If so, I'll adjust the test.

beorn7:
Test fixed. It should pass now.

beorn7:
I'll merge to get more time with the dependency updated (before cutting the RC).
@krasi-georgiev if my assumptions here are wrong, please let me know and we can rectify things.

PR #5582, opened 2019-05-20
web api: handle alert with Infinity/NaN values

holycheater:
The problem:
- `json-iterator/go` library encodes `math.Inf` and `math.NaN` as is (which is invalid json spec), where standard `encoding/json` raises an error: https://github.com/json-iterator/go/issues/365

The solution I see is to return 0 instead of Infinity/NaN values.

There is a pull request (https://github.com/prometheus/prometheus/pull/5303), but it is not correct and seems to be abandoned

### DIFF
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -465,7 +465,7 @@ $ curl http://localhost:9090/api/v1/rules
                                     "severity": "page"
                                 },
                                 "state": "firing",
-                                "value": 1
+                                "value": "1e+00"
                             }
                         ],
                         "annotations": {
@@ -522,7 +522,7 @@ $ curl http://localhost:9090/api/v1/alerts
                     "alertname": "my-alert"
                 },
                 "state": "firing",
-                "value": 1
+                "value": "1e+00"
             }
         ]
     },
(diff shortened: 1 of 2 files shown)


brian-brazil:
5303 would be the preferred way of handling this, if you'd like to take it
over.

simonpasquier:
This isn't the correct change: the solution would be to return a string value, see https://github.com/prometheus/prometheus/pull/5303#pullrequestreview-212945209.

holycheater:
You mean return number if it is valid or string if it is not?

brian-brazil:
Always return a string.

holycheater:
It will still break clients which expect a number, I can fix go client, but I don’t know about others. Are you ok with that?

brian-brazil:
Yes, that's acceptable here and in line with what we do for the query api.

holycheater:
I've updated api to return alert value as a string.
Also made a PR in `client_golang`: https://github.com/prometheus/client_golang/pull/585/files
The resulting value can be parsed with `strconv.ParseFloat` if neccessary

brian-brazil:
Looks good, docs also need updating,

brian-brazil:
Thanks!

Issue #5570, opened 2019-05-16
panic in azure sd with VMs with public IPs


dmitsh:
## Bug Report
**What did you do?**
1. Deploy Linux VM with public IP in Azure
2. Start Prometheus with the config file containing `azure_sd_config`

**What did you expect to see?**
Normal execution

**What did you see instead? Under which circumstances?**
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xf08eb5]
```
**Environment**

* System information:
Linux 4.15.0-50-generic x86_64

* Prometheus version:
prometheus, version 2.9.2 (branch: upstream-master, revision: 0a343996115d7063b81022a11e5aa9cf2c30fb14)
  build user:       vagrant@ubuntu-bionic
  build date:       20190516-03:22:56
  go version:       go1.12.4

* Prometheus configuration file:
```
# my global config
global:
  scrape_interval:     15s
  evaluation_interval: 30s
  # scrape_timeout is set to the global default (10s).

scrape_configs:
- job_name: service-azure
  azure_sd_configs:
    - environment: AzurePublicCloud
      authentication_method: OAuth
      subscription_id: **********
      tenant_id: **********
      client_id: **********
      client_secret: **********
      port: 9100
```

* Logs:
```
level=info ts=2019-05-16T03:12:37.431Z caller=main.go:614 msg="Server is ready to receive web requests."
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xf08eb5]

goroutine 205 [running]:
github.com/prometheus/prometheus/discovery/azure.(*Discovery).refresh.func1(0xc000d20510, 0xc00074ae40, 0xc0001b9e60, 0xc0000bb400, 0x33f2900, 0xc0006d5500, 0x0, 0xc000397980, 0x7d, 0xc000cfe4d4, ...)
	/home/vagrant/gopath/src/github.com/prometheus/prometheus/discovery/azure/azure.go:339 +0x1015
created by github.com/prometheus/prometheus/discovery/azure.(*Discovery).refresh
	/home/vagrant/gopath/src/github.com/prometheus/prometheus/discovery/azure/azure.go:283 +0x803
exit status 2
```

dmitsh:
I'm looking into this issue. Will post a PR soon.

dmitsh:
Fixed with #5587

PR #5587, opened 2019-05-21
fix nil pointer dereference in azure discovery

dmitsh:
Azure service discovery fails with `panic: runtime error: invalid memory address or nil pointer dereference` on a VM with public IP
This PR adds a simple pointer check.
Refers to #5570

### DIFF
--- a/discovery/azure/azure.go
+++ b/discovery/azure/azure.go
@@ -335,7 +335,7 @@ func (d *Discovery) refresh(ctx context.Context) ([]*targetgroup.Group, error) {
 
 				if *networkInterface.Primary {
 					for _, ip := range *networkInterface.IPConfigurations {
-						if ip.PublicIPAddress != nil {
+						if ip.PublicIPAddress != nil && ip.PublicIPAddress.PublicIPAddressPropertiesFormat != nil {
 							labels[azureLabelMachinePublicIP] = model.LabelValue(*ip.PublicIPAddress.IPAddress)
 						}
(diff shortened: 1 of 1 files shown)


dmitsh:
@brian-brazil @simonpasquier @alfred-landrum please take a look

discovery/azure/azure.go
```diff
@@ -335,7 +335,7 @@ func (d *Discovery) refresh(ctx context.Context) ([]*targetgroup.Group, error) {
 
 				if *networkInterface.Primary {
 					for _, ip := range *networkInterface.IPConfigurations {
-						if ip.PublicIPAddress != nil {
+						if ip.PublicIPAddress != nil && ip.PublicIPAddress.PublicIPAddressPropertiesFormat != nil {
```
brian-brazil: Shouldn't this be ip.PublicIPAddress.IPAddress?

dmitsh: Either way. `IPAddress` is an anonymous field defined in `PublicIPAddressPropertiesFormat`:

```
// PublicIPAddress public IP address resource.
type PublicIPAddress struct {
	autorest.Response `json:"-"`
	// Sku - The public IP address SKU.
	Sku *PublicIPAddressSku `json:"sku,omitempty"`
	// PublicIPAddressPropertiesFormat - Public IP address properties.
	*PublicIPAddressPropertiesFormat `json:"properties,omitempty"`
...
```

brian-brazil:
Thanks!

PR #5592, opened 2019-05-23
Compress WAL

csmarchbanks:
This is a PR in order to run benchmarks for compressing the WAL (see https://github.com/prometheus/tsdb/pull/609).

@krasi-georgiev would you be willing to start a benchmark run when you get a chance?

Currently WALCompression is defaulted to true for the benchmarks, I will switch it to false before marking this ready to merge.


krasi-georgiev:
Will run for a day and will also check the WAL size savings.

csmarchbanks:
Looks like compression ratios in the 1.8 - 2.8 range ([prom query](http://prombench.prometheus.io/prometheus-meta/graph?g0.range_input=6h&g0.expr=rate(prometheus_tsdb_wal_raw_record_bytes_total%5B15m%5D)%20%2F%20rate(prometheus_tsdb_wal_compressed_record_bytes_total%5B15m%5D)&g0.tab=0))

Looks good to me so far, WAL fsync latencies are also down significantly. Is it possible to restart each of the Prometheus instances? It would be interesting to see how reading the WAL at startup differs.

krasi-georgiev:
restarted. 

not sure if there is a metric for the time it took to read the logs so let me know and I can probably figure out the timings from the logs.

csmarchbanks:
Thanks! I can get an estimate based on when samples start getting ingested again:
master: 00:10:10 -> 00:24:55 = 14 minutes, 45 seconds
pr: 00:18:10 -> 00:33:00 = 14 minutes, 50 seconds

Almost no difference, so seems like decoding speed is not something to be worried about either.

krasi-georgiev:
amazing difference in the WAL size

no compression
```
7.9G    /mnt/disks/ssd0/wal/checkpoint.001516
63G     /mnt/disks/ssd0/wal/
```

compressed
```
828M    /mnt/disks/ssd0/wal/checkpoint.000798
30G     /mnt/disks/ssd0/wal/
```

krasi-georgiev:
do we need more data or can I shut it down?

csmarchbanks:
That’s all the data I want, feel free to shut it down.

bwplotka:
@csmarchbanks TSDB https://github.com/prometheus/tsdb/releases/tag/v0.9.1 is released. (:

csmarchbanks:
@brancz @bwplotka @krasi-georgiev This is ready for a review now. 

Did a quick smoke test of remote write in both compressed and non-compressed modes and everything is looking good.

csmarchbanks:
I broke it into two commits for ease of review - one for a minimal tsdb update, one for adding the WAL compression flag.

storage/remote/wal_watcher.go
```diff
@@ -78,6 +78,7 @@ var (
 		},
 		[]string{queue},
 	)
+	liveReaderMetrics = wal.NewLiveReaderMetrics(prometheus.DefaultRegisterer)
```
brancz: I'd love to see us refactor this eventually to not use global :) not a blocker here though

storage/remote/wal_watcher_test.go
```diff
@@ -505,5 +505,5 @@ func TestCheckpointSeriesReset(t *testing.T) {
 	// If you modify the checkpoint and truncate segment #'s run the test to see how
 	// many series records you end up with and change the last Equals check accordingly
 	// or modify the Equals to Assert(len(wt.seriesLabels) < seriesCount*10)
-	testutil.Equals(t, 14, wt.checkNumLabels())
+	testutil.Equals(t, 13, wt.checkNumLabels())
```
csmarchbanks: This change is due to the WAL compression causing a different number of segments.

storage/remote/wal_watcher_test.go
```diff
@@ -171,7 +171,7 @@ func TestReadToEndNoCheckpoint(t *testing.T) {
 	err = os.Mkdir(wdir, 0777)
 	testutil.Ok(t, err)
 
-	w, err := wal.NewSize(nil, nil, wdir, 128*pageSize)
+	w, err := wal.NewSize(nil, nil, wdir, 128*pageSize, true)
```
krasi-georgiev: is this intentional for making the wal compressed?

storage/remote/wal_watcher.go
csmarchbanks: Would also love to see that, I think the struct has to be exported to enable that and then we would have to pass it down from somewhere pretty high. Seemed like a more invasive change than making it global for now.

brancz: Definitely, just evangelizing this, this pattern is pretty spread around the codebase right now, so definitely something to solve here :)

storage/remote/wal_watcher_test.go
csmarchbanks: I wanted the tests to run using the compressed WAL since that is the future. Uncompressed is the default right now though so I could change them to false for now?

brancz: I would probably make sense to test both cases here. I do agree compressed should some day become the default, but for now uncompressed is the default and the default should be tested.

krasi-georgiev: I don't mind either way.

brancz: (or for this PR I'd be ok with just testing the default and adding the compressed case as a follow up)

csmarchbanks: Adding the compressed case is pretty easy will be done in a couple minutes. If it looks too complicated I can quickly revert it.

csmarchbanks: Added t.Run statements for both compressed and uncompressed.

csmarchbanks:
> In terms of test. Isn't wotth to test... both?

Yep, I agree. I updated all the tests to test both cases similar to the tsdb tests.

storage/remote/wal_watcher.go
cstyan: I can do this as part of prometheus/tsdb#606

PR #5593, opened 2019-05-23
Change dropdown hover color to white

nowke:
Change Dropdown item color in navigation bar (Status) on hover from black to white. Much more readable in `steelblue` background.

Change from,

before

Change to,

after

@simonpasquier


juliusv:
Sorry, didn't notice this earlier. This is for the old UI and will at least need a rebase (the conflicting file doesn't exist anymore, since asset embedding now happens differently), but happy to merge after that.

The new React UI already has different colors here, so should be fine.

roidelapluie:
Hi @nowke, we are still interested in this. Can you rebase this change? Thanks!

nowke:
@roidelapluie I have rebased the change.

roidelapluie:
Thanks! I have merged it as per Julius' comment:

> This is for the old UI and will at least need a rebase, but happy to merge after that.

PR #5601, opened 2019-05-28
Add a dashboard for remote write to prometheus-mixin

cstyan:
Didn't want to cramp the existing Prometheus dashboard so I added a new one specifically for remote write.
cc @tomwilkie @gouthamve 

### DIFF
--- a/documentation/prometheus-mixin/dashboards.libsonnet
+++ b/documentation/prometheus-mixin/dashboards.libsonnet
@@ -91,6 +91,60 @@ local g = import 'grafana-builder/grafana.libsonnet';
           { yaxes: g.yaxes('ms') } +
           g.stack,
         )
-      )
-  },
+      ),
+    # Remote write specific dashboard.
+    'prometheus-remote-write.json':
+      g.dashboard('Prometheus Remote Write')
+        .addMultiTemplate('instance', 'prometheus_build_info', 'instance')
+        .addMultiTemplate('cluster', 'kube_pod_container_info{image=~".*prometheus.*"}', 'cluster')
+          .addRow(
+            g.row('Timestamps')
+            .addPanel(
+              g.panel('Highest Timestamp In vs. Highest Timestamp Sent') +
+              g.queryPanel('prometheus_remote_storage_highest_timestamp_in_seconds{cluster=~"$cluster", instance=~"$instance"} - ignoring(queue) group_right(instance) prometheus_remote_storage_queue_highest_sent_timestamp_seconds{cluster=~"$cluster", instance=~"$instance"}', '{{cluster}}:{{instance}}-{{queue}}') +
+              { yaxes: g.yaxes('s') }
+            )
+            .addPanel(
+              g.panel('Rate[5m]') +
+              g.queryPanel('rate(prometheus_remote_storage_highest_timestamp_in_seconds{cluster=~"$cluster", instance=~"$instance"}[5m])  - ignoring (queue) group_right(instance) rate(prometheus_remote_storage_queue_highest_sent_timestamp_seconds{cluster=~"$cluster", instance=~"$instance"}[5m])', '{{cluster}}:{{instance}}-{{queue}}')
+            )
+          )
+          .addRow(
+            g.row('Samples')
+            .addPanel(
+              g.panel('Rate, in vs. succeeded or dropped [5m]') +
+              g.queryPanel('rate(prometheus_remote_storage_samples_in_total{cluster=~"$cluster", instance=~"$instance"}[5m])- ignoring(queue) group_right(instance) rate(prometheus_remote_storage_succeeded_samples_total{cluster=~"$cluster", instance=~"$instance"}[5m]) - rate(prometheus_remote_storage_dropped_samples_total{cluster=~"$cluster", instance=~"$instance"}[5m])', '{{cluster}}:{{instance}}-{{queue}}')
+            )
+          )
+          .addRow(
+            g.row('Shards')
+            .addPanel(
+              g.panel("Num. Shards") +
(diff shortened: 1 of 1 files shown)


documentation/prometheus-mixin/dashboards.libsonnet
```diff
@@ -91,6 +91,63 @@ local g = import 'grafana-builder/grafana.libsonnet';
...
+            .addPanel(
+              g.panel('Rate[5m]') +
+              g.queryPanel('rate(prometheus_remote_storage_highest_timestamp_in_seconds{instance=~"$instance"}[5m])  - ignoring (queue) group_right(instance) rate(prometheus_remote_storage_queue_highest_sent_timestamp_seconds{instance=~"$instance"}[5m])', '{{queue}}')
+            )
+          )
+          .addRow(
+            g.row('Samples')
+            .addPanel(
+              g.panel('Samples In vs. Samples Out') +
+              g.queryPanel('prometheus_remote_storage_samples_in_total{instance=~"$instance"} - ignoring (queue) group_right(instance) prometheus_remote_storage_succeeded_samples_total{instance=~"$instance"}', '{{queue}}')
```
brian-brazil: You shouldn't do math on counters, if you want this better to use a gauge.

documentation/prometheus-mixin/dashboards.libsonnet
```diff
@@ -91,6 +91,63 @@ local g = import 'grafana-builder/grafana.libsonnet';
...
+      ),
+    # Remote write specific dashboard.
+    'prometheus-remote-write.json':
+      g.dashboard('Prometheus Remote Write')
+        .addMultiTemplate('instance', 'prometheus_build_info', 'instance')
+          .addRow(
+            g.row('Timestamps')
+            .addPanel(
+              g.panel('Highest Timestamp In vs. Highest Timestamp Sent') +
+              g.queryPanel('prometheus_remote_storage_highest_timestamp_in_seconds{instance=~"$instance"} - ignoring(queue) group_right(instance) prometheus_remote_storage_queue_highest_sent_timestamp_seconds{instance=~"$instance"}', '{{queue}}') +
```
brian-brazil: Prefer group_left to group_right.

cstyan: That's wouldn't work, `prometheus_remote_storage_highest_timestamp_in_seconds` doesn't have the `queue` label but we'd want the graph to display the difference in seconds for each remote write queue. I don't know promql very well so if you have a suggestion for an alternative query let me know.

brian-brazil: Flip the RHS and LHS

cstyan: That would give a negative value, I guess then I'd just `* -1`?

brian-brazil: Yeah. I partially regret adding group_right (for symmetry), this is the first case where I'd seen it kinda making sense.

cstyan:
@gouthamve updated the legends for the remote write dashboard

documentation/prometheus-mixin/dashboards.libsonnet
```diff
@@ -91,6 +91,63 @@ local g = import 'grafana-builder/grafana.libsonnet';
...
+            .addPanel(
+              g.panel('Rate[5m]') +
+              g.queryPanel('rate(prometheus_remote_storage_highest_timestamp_in_seconds{instance=~"$instance"}[5m])  - ignoring (queue) group_right(instance) rate(prometheus_remote_storage_queue_highest_sent_timestamp_seconds{instance=~"$instance"}[5m])', '{{instance}}-{{queue}}')
+            )
+          )
+          .addRow(
+            g.row('Samples')
+            .addPanel(
+              g.panel('Samples In vs. Samples Out') +
+              g.queryPanel('prometheus_remote_storage_samples_in_total{instance=~"$instance"} - ignoring (queue) group_right(instance) prometheus_remote_storage_succeeded_samples_total{instance=~"$instance"}', '{{instance}}-{{queue}}')
```
beorn7: This is still math on a bare counter. I guess we can be pretty sure that those two counters are always reset at the same time, but then we would rely on an implementation detail that is left open by the spec. (Counters may be reset at any time, even without a binary restart.)

The semantics of the panel is also a bit weird as it is implicitly "since binary restarted". Howe about using `rate` or `increase` over a timeframe that makes sense (1h? 1d?)?

beorn7:
CI failures should go away after merging in master. (Go modules checksum mismatch.)

documentation/prometheus-mixin/dashboards.libsonnet
```diff
@@ -91,6 +91,63 @@ local g = import 'grafana-builder/grafana.libsonnet';
...
+            )
+            .addPanel(
+              g.panel('Rate[5m]') +
+              g.queryPanel('rate(prometheus_remote_storage_samples_in_total{instance=~"$instance"}[5m])  - ignoring (queue) group_right(instance) rate(prometheus_remote_storage_succeeded_samples_total{instance=~"$instance"}[5m])', '{{instance}}-{{queue}}')
```
beorn7: Spurious blank in `[5m])  - ignoring`

cstyan:
oh boy, rebased off master, one sec

documentation/prometheus-mixin/dashboards.libsonnet
beorn7: Note: After some in-person discussion, we decided to not have this particular panel at all.

beorn7:
:+1: will merge now (it's anyway not into master, but the mixin branch).

PR #5603, opened 2019-05-28
Allow swapping out PromQL engine in web API

juliusv:
Currently we allow swapping out the Queryable that the *promql.Engine uses, but
if one wants to implement PromQL evaluation by an entirely different mechanism
than the *promql.Engine, this upper layer also needs to be swappable. The
concrete motivation for this is for adding a Prometheus API endpoint to
InfluxDB which handles PromQL queries by transpiling them to Flux and then
executing the resulting Flux query instead.

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -133,11 +133,16 @@ type TSDBAdmin interface {
 	Snapshot(dir string, withHead bool) error
 }
 
+type PromQLEngine interface {
+	NewInstantQuery(q storage.Queryable, qs string, ts time.Time) (promql.Query, error)
+	NewRangeQuery(q storage.Queryable, qs string, start, end time.Time, interval time.Duration) (promql.Query, error)
+}
+
 // API can register a set of endpoints in a router and handle
 // them using the provided storage and query engine.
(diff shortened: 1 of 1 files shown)


roidelapluie:
Since we import promql anyway, should we move the interface to the promql package and use that in:

cmd/promtool/
web/api/v1/
rules/

?

juliusv:
@roidelapluie So a common pattern in Go is to define an interface at a usage site (rather than at the site of an implementation) if you need a narrower interface than the actual implementation type, in the sense that the interface here only contains the methods required by the API. The concrete `*promql.Engine` has more methods than just these two (actually just one more at the moment, but could be more).

But if that narrower interface is required in multiple places, it could make sense to move it to a shared place, yeah. If we moved it to the `promql` package, then we should use interface composition there, with one narrow interface just defining the two methods required here (and in promtool, etc.), and another interface embedding that, plus adding the one more other method.

But I'm not sure that's needed at the moment. Initially I had filed this PR for a specific reason related to the PromQL->Flux transpiler, I believe. But I think the reason became obsolete. I'll close it for now.

PR #5604, opened 2019-05-28
Update queue config documentation

cstyan:
Update queue config documentation to reflect default value change for capacity. It's now a queue length (shard capacity) of 10 by default (see #4588).


cstyan:
@gouthamve

PR #5609, opened 2019-05-30
Add symlink commands back in circle-ci config so that Dockerfile will work locally.

cstyan:
Changes in #5031 broke the locker `make docker` locally, 
```Step 6/19 : ARG OS="linux"
 ---> Using cache
 ---> 762f78671055
Step 7/19 : COPY .build/${OS}-${ARCH}/prometheus        /bin/prometheus
COPY failed: stat /var/lib/docker/tmp/docker-builder535576532/.build/linux-amd64/prometheus: no such file or directory
make: *** [Makefile.common:212: common-docker-amd64] Error 1```

cc @SuperQ @johanneswuerbach 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,6 +62,8 @@ jobs:
     - run: docker run --privileged linuxkit/binfmt:v0.6
     - attach_workspace:
         at: .
+    - run: ln -s .build/linux-amd64/prometheus prometheus	
+    - run: ln -s .build/linux-amd64/promtool promtool
     - run: make docker
     - run: make docker DOCKER_REPO=quay.io/prometheus
     - run: docker images
@@ -89,6 +91,8 @@ jobs:
     - store_artifacts:
         path: .tarballs
         destination: releases
+    - run: ln -s .build/linux-amd64/prometheus prometheus	
+    - run: ln -s .build/linux-amd64/promtool promtool
     - run: make docker DOCKER_IMAGE_TAG=$CIRCLE_TAG
     - run: make docker DOCKER_IMAGE_TAG=$CIRCLE_TAG DOCKER_REPO=quay.io/prometheus
     - run: docker login -u $DOCKER_LOGIN -p $DOCKER_PASSWORD
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,8 @@ LABEL maintainer="The Prometheus Authors <prometheus-developers@googlegroups.com
 
 ARG ARCH="amd64"
 ARG OS="linux"
-COPY .build/${OS}-${ARCH}/prometheus        /bin/prometheus
-COPY .build/${OS}-${ARCH}/promtool          /bin/promtool
+COPY prometheus        /bin/prometheus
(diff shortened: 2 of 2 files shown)


cstyan:
oops, forgot to ping @SuperQ @johanneswuerbach

johanneswuerbach:
This would break cross image building, could you try to running „make docker-amd64“ instead?

simonpasquier:
> This would break cross image building, could you try to run „make docker-amd64“ instead?

I guess that @cstyan's workflow is `make build docker` (eg no cross-build) so it doesn't work either IIUC.

johanneswuerbach:
Afaik „make build docker-amd64“ should just work :tm: if an amd64 binary was built.

It might be worthwhile to either default „make docker“ to build amd64 images, detect the current system arch or build all supported architectures, but that would require additional work.

How is promu handling this case? Is it by default just building all architectures?

simonpasquier:
> Afaik „make build docker-amd64“ should just work tm

It doesn't, at least for me.

> How is promu handling this case? Is it by default just building all architectures?

`promu build` (hence `make build`) generates the binary in the root directory of the project for the current system while `promu crossbuild` puts things in `.build/<OS>-<ARCH>/`.

cstyan:
> Afaik „make build docker-amd64“ should just work tm

Nope, any local build will fail because the `.build` directory doesn't exist.

johanneswuerbach:
Circling in @SuperQ and @discordianfish as you both were involved in the cross-building PR.

To make `make build docker` work I see the following options:

1. Change `promu build` to also write artefacts to `.build/<OS>-<ARCH>/` 
2. Have a pre-flight step in `make docker` symlinking/copying artefacts to `.build/<OS>-<ARCH>/` 
3. Change `Dockerfile` to always look into the root and copy to per-arch binaries before building the images from `.build/<OS>-<ARCH>/`

Overall I would favour 1. as this would make things more consistent, but I've no context why binaries were output into the project root in the first place. 2. seems also fine, but I think with option 3 there is a significant risk that a docker image accidentally contains the wrong binaries.

For any of the above options we would also need to decide:

1. detect the local arch automatically and use that
2. always default to amd64

What do you think?

simonpasquier:
> Change promu build to also write artefacts to .build/<OS>-<ARCH>/

As long as the binaries are also present in the root directory, I'm ok with this. For local development, I usually do `make build` then `./prometheus`.

cstyan:
> As long as the binaries are also present in the root directory, I'm ok with this. For local development, I usually do make build then ./prometheus.

Same. I do this, and then if I want to test that build in one of our environments I also run `make docker`.

discordianfish:
@johanneswuerbach Yes that seems like the best option to me.
I'd suggest though to change promu build to only write to .build/<OS>-<ARCH> and have the Makefile copy the file to the root directory. Seems cleaner to me, but not feeling strongly if somebody has objections.

cstyan:
Master seems to build properly now, see: https://github.com/prometheus/prombench/pull/196

PR #5613, opened 2019-05-30
Add cancel button on graph ui

alileza:
While exploring metrics with sometimes it took a while, and I decided to modify or just abort the query. It happened to me a few times that I look for a cancel button, so I thought it would be nice to have one.

**Screenshot**

Signed-off-by: Ali Reza <alirezayahya@gmail.com>


codesome:
This would not cancel the query on the prometheus server. This UI change might give a wrong impression that query has stopped in the prometheus server too.

alileza:
Yep, I'm aware of that, if you want the server to cancel the query, you also need to handle on the server side. While reading the code I also just realize, by re-execute the query, it'll abort the request as well. Feel free to close the PR if it doesn't seem to be acceptable~

brian-brazil:
Shouldn't contexts cause it to be cancelled?

alileza:
Ah right, didn't notice that! then I don't see any downside of having this button.

roidelapluie:
cc @juliusv did you see this PR?

juliusv:
Sorry for the long inactivity from my side. I think if we implement something like this, then:

- It should go into the new React UI now
- It should be visually designed to be as unobtrusive as possible, e.g. turning an existing button (maybe the "Execute" one?) into a cancellation button while the query is running

alileza:
I think that make sense in a user experience perspective to turn the execute button to cancel. But anyway since we're moving to react ui, I think this won't be useful, I'll be closing it.

PR #5614, opened 2019-05-30
Remote Write Allocs Improvements

csmarchbanks:
A few different commits to improve allocations when running remote write.
1. Remove an unneeded temp variable
1. Only allocate `pendingSamples` once per shard
1. Allocate the `send` slice far fewer times
1. Use a mask rather than always copy samples into a temporary slice
1. Allocate a snappy buffer per shard rather than create one per request.

Added benchmark:
```
BenchmarkSampleDelivery-4     1206794       1141762       -5.39%

BenchmarkSampleDelivery-4     7152           7121           -0.43%

BenchmarkSampleDelivery-4     909406        610669        -32.85%
```
Much of the remaining allocs are from proto encoding/decoding.

Reviewing commit by commit will have all changes in isolation.

@cstyan Have you been running any benchmarks for these? I only see one that requires a real WAL somewhere.

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -361,7 +361,6 @@ func (t *QueueManager) StoreSeries(series []tsdb.RefSeries, index int) {
 	t.seriesMtx.Lock()
 	defer t.seriesMtx.Unlock()
 
-	temp := make(map[uint64][]prompb.Label, len(series))
 	for _, s := range series {
 		ls := processExternalLabels(s.Labels, t.externalLabels)
 		rl := relabel.Process(ls, t.relabelConfigs...)
@@ -369,19 +368,16 @@ func (t *QueueManager) StoreSeries(series []tsdb.RefSeries, index int) {
 			t.droppedSeries[s.Ref] = struct{}{}
 			continue
 		}
-		temp[s.Ref] = labelsToLabelsProto(rl)
-	}
-
-	for ref, labels := range temp {
-		t.seriesSegmentIndexes[ref] = index
+		t.seriesSegmentIndexes[s.Ref] = index
+		labels := labelsToLabelsProto(rl)
 
 		// We should not ever be replacing a series labels in the map, but just
 		// in case we do we need to ensure we do not leak the replaced interned
 		// strings.
-		if orig, ok := t.seriesLabels[ref]; ok {
+		if orig, ok := t.seriesLabels[s.Ref]; ok {
 			release(orig)
 		}
-		t.seriesLabels[ref] = labels
+		t.seriesLabels[s.Ref] = labels
 	}
 }
 
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -657,9 +657,9 @@ func (s *shards) runShard(ctx context.Context, i int, queue chan prompb.TimeSeri
 	// Send batches of at most MaxSamplesPerSend samples to the remote storage.
 	// If we have fewer samples than that, flush them out after a deadline
 	// anyways.
-	pendingSamples := []prompb.TimeSeries{}
-
 	max := s.qm.cfg.MaxSamplesPerSend
+	pendingSamples := make([]prompb.TimeSeries, 0, max)
+
 	timer := time.NewTimer(time.Duration(s.qm.cfg.BatchSendDeadline))
 	stop := func() {
 		if !timer.Stop() {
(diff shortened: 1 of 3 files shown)


storage/remote/queue_manager.go
```diff
@@ -231,40 +231,30 @@ func NewQueueManager(logger log.Logger, walDir string, samplesIn *ewmaRate, cfg
...
-		})
 	}
 	t.seriesMtx.Unlock()
 
+	j := 0
 outer:
-	for _, sample := range tempSamples {
+	for i, sample := range s {
+		// Filter out any samples marked as ignored.
+		for ; j < len(ignored) && ignored[j] <= i; j++ {
```
brian-brazil: This smells n^2, how about using a map?

storage/remote/queue_manager.go
```diff
@@ -661,9 +656,9 @@ func (s *shards) runShard(ctx context.Context, i int, queue chan prompb.TimeSeri
 	// Send batches of at most MaxSamplesPerSend samples to the remote storage.
 	// If we have fewer samples than that, flush them out after a deadline
 	// anyways.
-	pendingSamples := []prompb.TimeSeries{}
-
 	max := s.qm.cfg.MaxSamplesPerSend
+	pendingSamples := make([]prompb.TimeSeries, 0, max)
```
cstyan: I think this has the potential to cause more backup problems.

csmarchbanks: Could you explain this comment more? I am not seeing how this would cause backup problems.

cstyan: Sorry, I was thinking this would enforce there being no more than `maxSamplesPerSend` pending samples, but that's not how slices work :man_facepalming:  

Backup meaning the shards not being able to send samples at the rate their being scrapped and written to the WAL. At the moment we send as soon as we reach `maxSamplesPerSend` or the flush timeout. At the moment we can potentially have some multiple of `maxSamplesPerSend` samples pending, meaning that after a successful send there's theoretically less time before we have enough samples for the next send. If we enforced only have `maxSamplesPerSend` pending it could take longer to fill up `pendingSample` given that we block on the shards queue.

storage/remote/queue_manager.go
csmarchbanks: This was still O(n) due to not resetting j during each loop. Definitely not the most obvious code I have ever written though. Also, my newest change makes it irrelevant.

csmarchbanks:
When looking further through the code I am pretty convinced that `seriesMtx` is unnecessary which allows me simplify `Append` to only have one loop and no `tempSamples`. I have been running remote write locally for awhile now with `-race`, and only found the race condition in `maxGauge` that I fixed with the most recent commit.

csmarchbanks:
Update: I ran this branch all weekend in a cluster without issue. I am reasonably sure that the mutex is not needed.

storage/remote/queue_manager.go
```diff
@@ -734,9 +707,10 @@ func (s *shards) sendSamples(ctx context.Context, samples []prompb.TimeSeries) {
 }
 
 // sendSamples to the remote storage with backoff for recoverable errors.
-func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.TimeSeries) error {
+func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.TimeSeries, buf *[]byte) error {
 	backoff := s.qm.cfg.MinBackoff
-	req, highest, err := buildWriteRequest(samples)
+	req, highest, err := buildWriteRequest(samples, *buf)
+	*buf = req
```
tomwilkie: Do you think this assignment to the pointer should be in `buildWriteRequest`?

tomwilkie:
> Update: I ran this branch all weekend in a cluster without issue. I am reasonably sure that the mutex is not needed.

I think you're right, as there is a single go routine in the watcher driving all these interactions.  This will become more obvious when we refactor the WAL Watcher.

LGTM from me but I'll hold off on merging until @cstyan has had a chance to run this in our env.

storage/remote/queue_manager.go
csmarchbanks: I put it here because it is the last place that deals with `*shards` and I wanted to keep the mutation isolated to those functions. I am happy to move it down to `buildWriteRequest` if you have a strong preference.

csmarchbanks:
@tomwilkie I know @cstyan ran this for a little bit in one of your environments. Do you need to run it longer or is this good to go?

tomwilkie:
Let me ask @cstyan.  Sorry for dropping ball on this.

cstyan:
I just deployed a new build of this image to some test prometheus instances, it will run overnight at the least and I'll check resource usage in the morning.

tomwilkie:
LGTM 
Screenshot 2019-06-27 at 19 47 25

PR #5630, opened 2019-06-04
Add region to ec2 discovery labels

yershalom:


brian-brazil:
Thanks for your PR, this can already be done via relabelling. If there's an explicit field for this coming from the API we can use it though.

yershalom:
Hi @brian-brazil ,
Thanks for the fast response.. the ec2 discovery is calling to `Instances []*Instance`, which don't have the region...
Regard the relabelling, I tried to do it there but I didn't manage and after few tries I decided just to added this, but if you have some example for this I would be more than happy. I know it's sound stupid but I didn't manage to remove last character in the relabelling....

simonpasquier:
This should work:

```yaml
  - source_labels: [__meta_ec2_availability_zone]
    regex: (.+).
    target_label: region
```

yershalom:
Thanks a lot! Do you mind if I'll add this to the docs for ec2_sd_config?

brian-brazil:
Our docs are reference docs for EC2, not general relabelling/regex examples to avoid the docs becoming cluttered. You could put out a quick blog post somewhere to show this.

yershalom:
Awesome, thanks!

PR #5632, opened 2019-06-05
Add next three release shepherds

juliusv:

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -15,7 +15,10 @@ Release cadence of first pre-releases being cut is 6 weeks.
 | v2.8           | 2019-02-27                                 | Ganesh Vernekar (GitHub: @codesome)         |
 | v2.9           | 2019-04-10                                 | Brian Brazil (GitHub: @brian-brazil)        |
 | v2.10          | 2019-05-22                                 | Björn Rabenstein (GitHub: @beorn7)          |
-| v2.11          | 2019-07-03                                 | **searching for volunteer**                 |
+| v2.11          | 2019-07-03                                 | Ganesh Vernekar (GitHub: @codesome)         |
+| v2.12          | 2019-08-14                                 | Julius Volz (GitHub: @juliusv)              |
+| v2.13          | 2019-09-25                                 | Krasi Georgiev (GitHub: @krasi-georgiev)    |
+| v2.14          | 2019-11-06                                 | **searching for volunteer**                 |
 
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -15,7 +15,7 @@ Release cadence of first pre-releases being cut is 6 weeks.
 | v2.8           | 2019-02-27                                 | Ganesh Vernekar (GitHub: @codesome)         |
 | v2.9           | 2019-04-10                                 | Brian Brazil (GitHub: @brian-brazil)        |
 | v2.10          | 2019-05-22                                 | Björn Rabenstein (GitHub: @beorn7)          |
-| v2.11          | 2019-07-03                                 | **searching for volunteer**                 |
+| v2.11          | 2019-07-03                                 | Julius Volz (GitHub: @juliusv)              |
```
brian-brazil: We should add the next few slots too

juliusv:
Gonna add more slots and also change myself to a later slot.

RELEASE.md
brancz: I'd be ok with going again after, if nobody else wants to

beorn7: Krasi should do one, as well. 

Tom is probably too busy and will just delegate to his reportees.

We were thinking perhaps @codesome could do the next one.

So how about 2.11 @codesome , 2.12 @juliusv , 2,13 @krasi-georgiev , and then we check again later.

brancz: @codesome did 2.8

juliusv: Updated. @codesome and @krasi-georgiev, please confirm.

krasi-georgiev: confirmed. I can't tell about my schedule around the time for 2.13, but will try to plan for it.

codesome: I would probably skip 2.11 as I have been heavily out of touch since past couple of month due to academics. I can do any release after that.

beorn7: Yes, we know. @brancz Are you volunteering for 2.11?

beorn7: "Yes we know' was for @brancz comment. Concurrency, how does it work…

brancz: Yeah I'm ok with doing 2.11.

beorn7: \o/ I'll update this PR.

beorn7:
This should be it. Merging before somebody gets second thoughts. :o)

Issue #2357, opened 2017-01-21
Support for environment variable substitution in configuration file


dopuskh3:
I think that would be a good idea to substitute environment variables in the configuration file.

That could be done really easily using os.ExpandEnv on configuration string when loading configuration string. 

That would be much easier to substitute environment variables only on configuration values. go -ini provides a valueMapper but yaml.v2 doesn't have such mechanism.

brian-brazil:
We've previously discussed several times and decided against. For simplicity we keep to one way to configure a given thing, otherwise we'd end up with 4-5 ways.

alexellis:
This could be a particularly powerful way to configure Prometheus in a cluster without the burdensome task of a separate config file for every configuration you can ever deploy.

nutchalum:
It's a nice to have.
Now I've encountered a problem with `external_label`.
I've deploy prometheus `Deployment` and `ConfigMap` via kubernetes federation.
All of them have the same configuration, **except** `external_label`.
I can't think a way to injected this `external_label` (like source: clusterA)
So I have to deploy `ConfigMap` in the old way.
(I have around 7 clusters, that means 7 times instead of 1 time via Federation)

Nick-Triller:
Environment variable substitution would allow to keep sensitive data like passwords used for [remote read/write integration](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#%3Cremote_write%3E) out of the config file, too. 

Influx [recently mentioned ](https://www.influxdata.com/blog/influxdb-now-supports-prometheus-remote-read-write-natively/) environment variable substitution in config files would be a useful feature in their eyes.

ababushkin:
I don't think it would cause usability issues guys, its pretty common to use environment variable substitution in config files like these. Any chance to revise that decision?

Here are some benefits:
- Keeps sensitive configuration out of config files
- Simplifies build / deployment process, especially in containerised environments
- Inline with 12factor principles (https://12factor.net/)

Am I missing something here?

boeboe:
Any update on this?

Bplotka:
We have similar issues when we run Prometheus in 2-replica statefulset which reuses single configmap. To support HA we have `external_label` called `replica` that suppose to have unique replica name (e.g hostname). We end up doing this:
```
spec:
      containers:
      - args:
        - -c
        - sed s/'@@HOSTNAME@@'/${HOSTNAME}/g /etc/prometheus/prometheus.yaml.tmpl
[...]

PR #5633, opened 2019-06-05
Use `CONSUL_HTTP_ADDR` adress when available in environment for consul discovery

pierresouchay:
By default consul tools use `CONSUL_HTTP_ADDR` to guess the schemes/server ip/port.

Use the same naming convention, so exporting http://consul.machine.acme.org:8500
will be used for consul discovery instead of localhost:8500.

This also detect the scheme and apply similar rules as Consul if scheme is not
setup.

### DIFF
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"os"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


brian-brazil:
To keep things understandable and have only one source of truth, we do not read configuration from the environment. That way you can look at the file and know what's happening.

roidelapluie:
The prometheus developers have voted against env variable substitution in #2357. Even if this is not stricto sensu env variable substitution, I think that the vote applies to pull requests like this too.

This pull request can be closed.

brian-brazil:
If a library under us happens to do it that's one thing, but we should never be explicitly reading configuration from environment variables in our own code.

PR #5634, opened 2019-06-05
Update instructions for the release shepherd

beorn7:
This codifies how 2.10 was released. It removes the inconsistency of
freezing master for pre-releases but handling post-release bug fixes
in a separate branch.

The previous instructions came from a time where master was often in
bad shape. However, that's a problem of its own, which should be
avoided at all times and not only when a release is imminent. On other
words, freezing master can still happen if it is in bad shape and we
need a break from feature development and just fix the bugs for a
while. However, it should not happen as a formal step during the
release of a release candidate. On the contrary, a release candidate
is not really a release candidate if we already know it is in such a
bad shape that we need bug fixes. On the other hand, if we truly
believe the RC could be identical to the final release, there is no
need to freeze master.

I can explain more of the concept if there is still a need for
clarification.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,6 +1,6 @@
 # Releases
 
-This page describes the release process and the currently planned schedule for upcoming releases as well as the respective release shepherd. Release shepherd are chosen on a voluntary basis.
+This page describes the release process and the currently planned schedule for upcoming releases as well as the respective release shepherd. Release shepherds are chosen on a voluntary basis.
 
 ## Release schedule
 
@@ -24,14 +24,12 @@ If you are interested in volunteering please create a pull request against the [
 
 ## Release shepherd responsibilities
 
-The release shepherd is responsible for the entire release series of a minor release, meaning all pre- and patch releases of a minor release. The process starts with the initial pre-release.
+The release shepherd is responsible for the entire release series of a minor release, meaning all pre- and patch releases of a minor release. The process formally starts with the initial pre-release, but some preparations should be done a few days in advance.
 
-* The first pre-release is scheduled according to the above schedule.
-* With the pre-release the release shepherd is responsible for running and monitoring a benchmark run of the pre-release for 3 days, after which, if successful, the pre-release is promoted to a stable release.
-* Once a pre-release has been released, the `master` branch of the repository is frozen for any feature work, only critical bug fix work concerning the minor release is merged.
-* Pre-releases are done from `master`, after pre-releases are promoted to the stable release a `release-major.minor` branch is created.
-
-_Experimental change of the above procedure for the v2.10 release: The `release-2.10` branch is created at the time the first pre-release is cut. All releases of the series including pre-releases are cut from the `release-2.10` branch. `master` will not be frozen for feature work._
+* We aim to keep the master branch in a working state at all times. In principle, it should be possible to cut a release from master at any time. In practice, things might not work out as nicely. A few days before the pre-release is scheduled, the shepherd should check the state of master. Following their best judgement, the shepherd should try to expedite bug fixes that are still in progress but should make it into the release. On the other hand, the shepherd may hold back merging last-minute invasive and risky changes that are better suited for the next minor release.
+* On the date listed in the table above, the release shepherd cuts the first pre-release (using the suffix `-rc.0`) and creates a new branch called  `release-<major>.<minor>` starting at the commit tagged for the pre-release. In general, a pre-release is considered a release candidate (that's what `rc` stands for) and should therefore not contain any known bugs that are planned to be fixed in the final release.
+* With the pre-release, the release shepherd is responsible for running and monitoring a benchmark run of the pre-release for 3 days, after which, if successful, the pre-release is promoted to a stable release.
+* If regressions or critical bugs are detected, they need to get fixed in the release branch (and later merged into master) before cutting a new pre-release (called `-rc.1`, `-rc.2`, etc.).
 
 See the next section for details on cutting an individual release.
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -45,7 +45,7 @@ We maintain a separate branch for each minor release, named `release-<major>.<mi
 
 The usual flow is to merge new features and changes into the master branch and to merge bug fixes into the latest release branch. Bug fixes are then merged into master from the latest release branch. The master branch should always contain all commits from the latest release branch. As long as master hasn't deviated from the release branch, new commits can also go to master, followed by merging master back into the release branch.
 
-If a bug fix got accidentally merged into master, cherry-pick commits have to be created in the latest release branch, which then have to be merged back into master. Try to avoid that situation.
+If a bug fix got accidentally merged into master after non-bug-fix changes in master, the bug-fix commits have to be cherry-picked into the release branch, which then have to be merged back into master. Try to avoid that situation.
 
 Maintaining the release branches for older minor releases happens on a best effort basis.
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -29,7 +29,7 @@ The release shepherd is responsible for the entire release series of a minor rel
 * We aim to keep the master branch in a working state at all times. In principle, it should be possible to cut a release from master at any time. In practice, things might not work out as nicely. A few days before the pre-release is scheduled, the shepherd should check the state of master. Following their best judgement, the shepherd should try to expedite bug fixes that are still in progress but should make it into the release. On the other hand, the shepherd may hold back merging last-minute invasive and risky changes that are better suited for the next minor release.
 * On the date listed in the table above, the release shepherd cuts the first pre-release (using the suffix `-rc.0`) and creates a new branch called  `release-<major>.<minor>` starting at the commit tagged for the pre-release. In general, a pre-release is considered a release candidate (that's what `rc` stands for) and should therefore not contain any known bugs that are planned to be fixed in the final release.
 * With the pre-release, the release shepherd is responsible for running and monitoring a benchmark run of the pre-release for 3 days, after which, if successful, the pre-release is promoted to a stable release.
-* If regressions or critical bugs are detected, they need to get fixed in the release branch (and later merged into master) before cutting a new pre-release (called `-rc.1`, `-rc.2`, etc.).
+* If regressions or critical bugs are detected, they need to get fixed before cutting a new pre-release (called `-rc.1`, `-rc.2`, etc.). 
 
 See the next section for details on cutting an individual release.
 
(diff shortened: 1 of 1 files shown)


beorn7:
Assigned reviewers are the next three release shepherds.

RELEASE.md
```diff
@@ -15,7 +15,7 @@ Release cadence of first pre-releases being cut is 6 weeks.
 | v2.8           | 2019-02-27                                 | Ganesh Vernekar (GitHub: @codesome)         |
 | v2.9           | 2019-04-10                                 | Brian Brazil (GitHub: @brian-brazil)        |
 | v2.10          | 2019-05-22                                 | Björn Rabenstein (GitHub: @beorn7)          |
-| v2.11          | 2019-07-03                                 | Frederic Branczyk (GitHub: @brancz)         |
+| v2.11          | 2019-07-03                                 | nFrederic Branczyk (GitHub: @brancz)         |
```
brancz: this looks accidental

RELEASE.md
```diff
@@ -24,14 +24,12 @@ If you are interested in volunteering please create a pull request against the [
...
+* We aim to keep the master branch in a working state at all times. In principle, it should be possible to cut a release from master at any time. In practice, things might not work out as nicely. A few days before the pre-release is scheduled, the shepherd should check the state of master. Following their best judgement, the shepherd should try to expedite bug fixes that are still in progress but should make it into the release. On the other hand, the shepherd may hold back merging last-minute invasive and risky changes that are better suited for the next minor release.
+* On the date listed in the table above, the release shepherd cuts the first pre-release (using the suffix `-rc.0`) and creates a new branch called  `release-<major>.<minor>` starting at the commit tagged for the pre-release. In general, a pre-release is considered a release candidate (that's what `rc` stands for) and should therefore not contain any known bugs that are planned to be fixed in the final release. If need be, the release has to be delayed.
```
brian-brazil: > should therefore not contain any known bugs that are planned to be fixed in the final release. If need be, the release has to be delayed.

This is not our standard practice, and risks us getting back into a place where releases are delayed months while one more bug needs to be fixed.

RELEASE.md
```diff
@@ -24,14 +24,12 @@ If you are interested in volunteering please create a pull request against the [
...
+* With the pre-release, the release shepherd is responsible for running and monitoring a benchmark run of the pre-release for 3 days, after which, if successful, the pre-release is promoted to a stable release.
+* If regressions or critical bugs are detected, they need to get fixed in the release branch (and later merged into master) before cutting a new pre-release (called `-rc.1`, `-rc.2`, etc.).
```
brian-brazil: Bug fixes going into master is also practiced, as noted below. This is more resilient to human error.

RELEASE.md
juliusv: I guess you mean that there can be known bugs which should not block a release. But isn't the "that are planned to be fixed in the final release" qualification sufficient to make sure it's only about a subset of known bugs?

RELEASE.md
juliusv: Yeah, so master first, then cherry-pick into release branch?

RELEASE.md
beorn7: Oops.

RELEASE.md
beorn7: I don't understand. The instruction call it an accident if a bugfix goes into master first.

RELEASE.md
beorn7: If we agree that a known bug is fine to have in the final release, it's OK to also to an RC with it. But if we have a bug we want to fix before the final release is out, we should in general not cut an RC that includes it. (Exceptions are possible, that's why the word "should" is used above.)

What I describe here is what I assume always has been our standard practice. Deviations were either deliberate exceptions or accidents.

What do all the others think?

RELEASE.md
beorn7: > Yeah, so master first, then cherry-pick into release branch?

No. That's explicitly against the recommended practice. It's called an accident, and that was always the case, even in the times when these instructions were still in the wiki.

brian-brazil: That's as the text has changed back and forth several times on this point, as there's no consensus here. I think we should let each shepard follow their own preferences.

RELEASE.md
brian-brazil: We've disobeyed this more often than obeyed, even 2.10 had a commit that wasn't in the rc - which is fine. What I'm worried about is that this indicates that it's more important to delay things to get them perfect, than to get a release out on a reasonable timeframe.

RELEASE.md
beorn7: I don't think the text has ever changed back and forth. Do you have a commit with the claimed change?

I think this needs consistency. Otherwise we'll never get into the habit. I suggest to follow the practice that was documented as such for ages. If there is a disagreement, perhaps we should have a vote about it.

RELEASE.md
beorn7: We can soften the wording even more than "if need be". However, I don't want to convey the notion that you throw a buggy pile of mess at the users as a "release" candidate. That's not why we cut them. We cut them to discover unknown bugs before we do the final release .

RELEASE.md
juliusv: I've always preferred merging everything into master first and cherry-picking into release branches, even if it means a change (not sure if the doc ever stated doing it like I would prefer). But that's just my vote :)

brian-brazil: > Do you have a commit with the claimed change?

https://github.com/prometheus/prometheus/wiki/HOWTO-cut-a-new-release/_compare/e59a6e4de82994aa86bcac1b658cf0c0a7b9fb98...ffc51407eee4f17b4e43734868aff6aad4389ced

> I think this needs consistency.

We don't seem to be having issues from this as-is.

> I suggest to follow the practice that was documented as such for ages

It was written down, but never discussed that I'm aware. I suspect that it might have been discussed in the SoundCloud building some years ago, but it was never discussed on email/IRC - from my standpoint it seems to just have appeared on the wiki one day.

RELEASE.md
brian-brazil: It's the final sentence with "delayed" that worries me more tbh.

RELEASE.md
beorn7: > https://github.com/prometheus/prometheus/wiki/HOWTO-cut-a-new-release/_compare/e59a6e4de82994aa86bcac1b658cf0c0a7b9fb98...ffc51407eee4f17b4e43734868aff6aad4389ced

That's not a "back and forth", that's going from "we haven't documented at all how we handle bugfix releases" to "this is how we do it". And ever since, the documented behavior was consistent.

> We don't seem to be having issues from this as-is.

I don't have an issue with the documented procedure "as is". I do have an issue that selected few decided not to follow it.

If you don't like the documented release procedure, please open a debate to change it, and don't just disregard the documentation.

This PR doesn't change the procedures WRT how bug fixes are handled. It only changes if we should freeze master or not for an RC. 

Conflating both things is just more complicated.

Could we handle in this PR what it is about?

RELEASE.md
beorn7: I got that. I'll just remove that sentence if that removes the controversy.

beorn7:
I removed the sentence with the "delay" and the "n" prefix from Frederic's name.

RELEASE.md
brian-brazil: > This PR doesn't change the procedures WRT how bug fixes are handled. It only changes if we should freeze master or not for an RC.

It changes several things since we last discussed releases at the last dev summit, including freezing of master (https://docs.google.com/document/d/1-C5PycocOZEVIPrmM1hn8fBelShqtqiAmFptoG4yK70/edit#heading=h.hvmkzmz0lp70)

The only thing approaching consensus that I'm aware of around bug fixes (https://github.com/prometheus/docs/pull/1169) is that it's up to the person doing the release. It appears that change got accidentally lost when switching PRs though.

juliusv: I'd also be generally for leaving that debate out of this PR, but existing text below in this document says as an option: "Whether merging master back into a release branch makes more sense is left up to the shepherd's judgement.". At the same time, this new bullet point does not mention that as a possibility, so one or the other needs to be changed or amended to keep things consistent.

beorn7: That's a different issue. This note is _not_ about cherry-picking, but about _merging_ master back into the release branch. That's possible as long as no feature development has been added in master. But it is completely orthogonal to the debato of cherry-picking things back from master vs. first do then in the release branch and then merge into master.

beorn7: Or in other words: As long as you can simply merge master back into the release branch, there is no cherry-picking required, and the debate is moot. The problem only occurs once master has feature development that you do not want to have in the release branch.

From that perspective, that sentence is just redundant and could be removed. But again, we don't have to do that in this PR.

juliusv: Yes I know, the extisting text is not about cherry-picking, but you writing "they need to get fixed in the release branch (and later merged into master)" seems to conflict with the possibility of fixing things in master first and merging master into the release branch.

beorn7: Got it. 

Of course, we could also add that bug fixes can happen in master for as long as nothing else has happened in master. Happy to add this, but is is kind of confusing and bloating the text. By just consistently merging bug fixes into the release branch first, things are just very smooth and straight, and we avoid all the complications, even before we have to revert to cherry-picking.

juliusv: I'd be happy with either way for now, as long as there's internal consistency in the document :) So you may have to remove that existing remark if you want to advocate for merging bugfixes into the release branch first.

beorn7: To hit two birds with one stone, I now _simplified_ this section (we are pointing to the remaining document for details anyway) and tried to clarify the "merge back into release branch" sentence. (I think is so clear now that we don't even need "shepherd's judgement" anymore. It is just equivalent, so no real judgement needed.)

I did it in a separate commit for clarity.

juliusv:
:+1:

RELEASE.md
brian-brazil: The wording there was odd alright.

> As long as master hasn't deviated from the release branch, new commits can also go to master

As written this doesn't really allow for bug fixes to master. I'd suggest "New commits could also go to master, followed by cherry-picking the commits into the release branch."

beorn7: That's _exactly_ what the procedure so far clearly marked as "accident".

As said above, if you want to make cherry-picking a regular thing, not to be avoided, please open a separate debate/PR.

beorn7: It's not just marked as an accident. It even says (and always has said ever since we documented the flow for bug fixes): "Try to avoid that situation."

I don't think we have to discuss this particular aspect anymore in this PR. It's just a different discussion.

beorn7:
Clarified in the "try to avoid" sentence that cherry-picking only applies after non-bug-fix changes in master.

brian-brazil:
:+1: 

It's slightly better than it was.

PR #5635, opened 2019-06-05
DOCKERFILE_PATH added to Makefile.common

geekodour:
See https://github.com/prometheus/prombench/pull/196

This will allow creation of docker images for Dockerfiles inside subdirectories.

cc @simonpasquier


simonpasquier:
LGTM

Issue #5585, opened 2019-05-21
Prometheus stops scraping when insufficient memory 


woodsaj:
**What did you do?**
The node that the Prometheus pod was running on ran out of memory (due to another job consuming it all).  

**What did you expect to see?**
An error (not warning) log message when memory could not be allocated.
Instead of silently failing, Prometheus should either try and recover or crash.

**What did you see instead? Under which circumstances?**
When the memory exhaustion was encountered prometheus stopped scraping metrics, but was otherwise online and responding.

**Environment**
Running as a pod in Kubernetes

* Prometheus version:

|   |   |
|-------|--------|
| Version     | 2.9.2 |
| Revision   | d3245f15022551c6fc8281766ea62db4d71e2747 |
| Branch      | HEAD   |
| BuildUser  | root@1d43b6951e8f |
| BuildDate  | 20190424-15:32:31 |
| GoVersion | go1.12.4 |

* Logs:
```
level=warn ts=2019-05-21T05:08:22.701Z caller=scrape.go:937 component="scrape manager" scrape_pool=kubernetes-pods target=http://10.40.1.20:80/metrics msg="append failed" err="write to WAL: log samples: write /prometheus/data/wal/00002922: cannot allocate memory"
```

gouthamve:
We decided to close this as a wont-fix. You should use meta-monitoring to catch this (alerting on scrapes failing for ex.) This is ideally done by having another Prometheus monitoring it.

PR #5637, opened 2019-06-06
Responding to #5585: Check for syscall errors

pswica:
Responding to https://github.com/prometheus/prometheus/issues/5585

I try to log an Error, instead of a warning, if there's a syscall error in `scrape.go`. Please let me know wherever I can improve.

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -23,6 +23,7 @@ import (
 	"io/ioutil"
 	"math"
 	"net/http"
+	"os"
 	"sync"
 	"time"
 	"unsafe"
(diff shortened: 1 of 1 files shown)


brian-brazil:
Can you explain what you're trying to do?

pswica:
Thanks so much for reviewing this. I'm trying to log an Error (rather than Warning) if the application catches a syscall error--that is, something like a ENOMEM (enough memory) error.

brian-brazil:
It's already a warning, and this seems to be tackling things at the wrong level. What exact syscall is returning this?

pswica:
I changed it to Error as this seemed to be a wanted/requested feature (instead of silently failing when the application is out of memory). I chose syscall errors to generate Error as I think most seem to warrant that.

I'm not sure which syscall (any call which asks for more memory when there is none could generate it), but would definitely be ENOMEM which has error code 12.
```
errno 12
>>ENOMEM 12 Cannot allocate memory
```

This matches the error message that @woodsaj reported: `write /prometheus/data/wal/00002922: cannot allocate memory`. 

Do you have any suggestions on how to go about this (I'm still new to open source, so I would love to learn more)? Also, let me know what further information you need.

brian-brazil:
Changing the severity of a log message doesn't solve any problems. The question here is why this isn't either panicing or OOMing.

pswica:
Oh I see what you are saying, yes that's definitely a better approach. Maybe this weirdness is due to using:

```
func yoloString(b []byte) string {
	return *((*string)(unsafe.Pointer(&b)))
}
```

Do you think a good approach would be replacing this with the safer golang construct `string([]byte)`? Let me know if this is something you'd welcome and I can implement this. Otherwise, please feel free to close this PR as I completely agree with your logic and thoughts.

brian-brazil:
That's nothing to do with it, that'd cause different problems.

PR #5640, opened 2019-06-06
web: add 'code' label to HTTP metrics

### LINKED ISSUE #5618: Instrument /api/v1 handlers for response codes
## Proposal: Instrument the /api/v1 handlers 
Currently, we have instrumentation for response codes on the `/api/v1/metrics` handler. This is `promhttp_metric_handler_requests_total`.

The proposal is that we instrument the other handlers like `/api/v1/query`, so that we can visualize and/or alert on these if we need to.

simonpasquier:
Closes #5618


web/web.go
```diff
@@ -96,7 +96,7 @@ var (
 			Help:    "Histogram of latencies for HTTP requests.",
 			Buckets: []float64{.1, .2, .4, 1, 3, 8, 20, 60, 120},
 		},
-		[]string{"handler"},
+		[]string{"handler", "code"},
```
brian-brazil: That's getting into high cardinality, and breaking out latency by success/failure can be problematic. I'd suggest a new counter metric.

simonpasquier: Makes sense, I'll update accordingly.

krasi-georgiev:
I think these metrics should also have a test. These are easy to write and houls be just few extra lines.

simonpasquier:
I'm not against having tests for metrics but I don't think it is a current practice in the prometheus repository so I'll defer to @brian-brazil.

brancz:
In my opinion, http requests total and duration are worth to have tests for. This is literally coming out of a regression (as we used to have these metrics), where we accidentally removed these metrics, so that's also a case for is :slightly_smiling_face: .

brian-brazil:
https://www.robustperception.io/when-should-you-unit-test-instrumentation are my general thoughts. If you want to add them to avoid a regression you can.

brancz:
Fully agree with the article, and these two are indeed the among the very few types of metrics where I think it makes sense.

simonpasquier:
@brian-brazil friendly ping, I've added a test for `prometheus_http_requests_total`.

brian-brazil:
:+1:

PR #5642, opened 2019-06-07
Change snapshot description to be consistent with rest of api.md

mkenigs:
Don't know if there's a reason to leave it the way it is for snapshot, but every other description puts parameters in this format. @juliusv


codesome:
:+1:

brian-brazil:
Thanks!

PR #5647, opened 2019-06-07
resolve race condition in maxGauge

dmitsh:

### DIFF
--- a/storage/remote/max_gauge.go
+++ b/storage/remote/max_gauge.go
@@ -35,5 +35,7 @@ func (m *maxGauge) Set(value float64) {
 }
 
 func (m *maxGauge) Get() float64 {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
 	return m.value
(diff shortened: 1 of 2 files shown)


csmarchbanks:
Thanks for the cc. Reminder to myself to remove this from https://github.com/prometheus/prometheus/pull/5614 if this PR is merged first.

dmitsh:
Could this one be merged? cc @brian-brazil

storage/remote/max_gauge.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 type maxGauge struct {
-	mtx   sync.Mutex
```
brian-brazil: A normal mutex is fine. RWMutexes are only suitable when there's many many more reads than writes.

dmitsh:
@brian-brazil addressed

brian-brazil:
Thanks!

PR #5648, opened 2019-06-10
labelkeep action should ignore __name__ label

ddongchen:
the labelkeep action is used to keep just some labels, and the '__name__' is also one of
the labels, while few people will add the '__name__' label when use labelkeep, which may waste
much time to debug

### DIFF
--- a/pkg/relabel/relabel.go
+++ b/pkg/relabel/relabel.go
@@ -243,6 +243,10 @@ func relabel(lset labels.Labels, cfg *Config) labels.Labels {
 		}
 	case LabelKeep:
 		for _, l := range lset {
+			// Should ignore metrics name
+			if l.Name == labels.MetricName {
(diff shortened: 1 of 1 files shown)


codesome:
This would add more confusion I believe. I prefer it to be handled in the config itself.

brian-brazil:
I agree, and relabelling is used more than with metrics.

ddongchen:
> This would add more confusion I believe. I prefer it to be handled in the config itself.

or maybe we fix it in other part? or shall we tell people that they should be careful with "labelkeep" in docs, and should add built-in labels, such as "__name__".

roidelapluie:
We have looked at this during our bug scrub. `__name__` should just be considered as other labels. We also now check for empty `__name__` to prevent mistakes with labelkeep/labeldrop.

Issue #5636, opened 2019-06-05
Remote write not honoring max_retries setting


murphyzlaw:
## Bug Report
**What did you do?**
Set up Prometheus to scrape some external metrics (e.g., Avalanche).
Configured remote write with max_retries explicitly as 2.
Set up a remote write test server based on documentation/examples/remote_storage/example_write_adapter but with an HTTP 500 response being returned. Following are the source changes to server.go

```
diff --git a/documentation/examples/remote_storage/example_write_adapter/server.go b/documentation/examples/remote_storage/example_write_adapter/server.go
index b94ef8fc..a737ea13 100644
--- a/documentation/examples/remote_storage/example_write_adapter/server.go
+++ b/documentation/examples/remote_storage/example_write_adapter/server.go
@@ -18,15 +18,26 @@ import (
        "io/ioutil"
        "log"
        "net/http"
+  "time"
 
        "github.com/gogo/protobuf/proto"
        "github.com/golang/snappy"
        "github.com/prometheus/common/model"
 
        "github.com/prometheus/prometheus/prompb"
+       kingpin "gopkg.in/alecthomas/kingpin.v2"
+)
+
+var (
+  allMetrics = kingpin.Flag("all", "Show all metrics.").Short('a').Bool()
 )
 
 func main() {
+       kingpin.Version("0.1-rad")
+       kingpin.CommandLine.Help = "Example Prometheus remote write metrics server"
+
+       kingpin.Parse()
+
        http.HandleFunc("/receive", func(w http.ResponseWriter, r *http.Request) {
                compressed, err := ioutil.ReadAll(r.Body)
                if err != nil {
@@ -51,10 +62,26 @@ func main() {
                        for _, l := range ts.Labels {
                                m[model.LabelName(l.Name)] = model.LabelValue(l.Value)
                        }
-                       fmt.Println(m)
-
+      
                        for _, s := range ts.Samples {
-                               fmt.Printf("  %f %d\n", s.Value, s.Timestamp)
+        unixsec := s.Timestamp / 1000
+        unixnsec := (s.Timestamp % 1000) * 1000000
+        timestamp := time.Unix(unixsec, unixnsec)
+        now := time.Now()
+        bufferPastDuration := 5 * time.Minute
+        pastLimit := now.Add(-1*bufferPastDuration)
+        if !pastLimit.Before(timestamp) {
+          msg := "ERROR too far in the past"
+          fmt.Println(msg)
+          fmt.Println(m)
[...]

PR #5649, opened 2019-06-10
Remove max_retries from queue_config

cstyan:
Re #5636, remote write no longer uses the max retries config option. We retry endlessly for recoverable errors.

cc @tomwilkie @brian-brazil

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -116,8 +116,7 @@ var (
 		Capacity:          10,
 		BatchSendDeadline: model.Duration(5 * time.Second),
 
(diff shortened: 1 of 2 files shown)


brian-brazil:
:+1:

Kampe:
Except most errors are not recoverable in practice: https://github.com/thanos-io/thanos/issues/4831

cstyan:
@Kampe Prometheus' remote write receiver treats `ErrOutOfOrder` as non-recoverable. If Thanos' remote write receiver code is not doing the same you should open an issue there.

PR #5652, opened 2019-06-11
Improve kubernetes scrape config example

sylr:
Use the following annotations:

 - prometheus.io/scrape
 - prometheus.io/port
 - prometheus.io/path
 - prometheus.io/scheme
 - prometheus.io/param-<name>

Signed-off-by: Sylvain Rabot <s.rabot@lectra.com>

### DIFF
--- a/documentation/examples/prometheus-kubernetes.yml
+++ b/documentation/examples/prometheus-kubernetes.yml
@@ -138,33 +138,40 @@ scrape_configs:
   - role: endpoints
 
   relabel_configs:
-  # Example relabel to scrape only endpoints that have
-  # "example.io/should_be_scraped = true" annotation.
-  #  - source_labels: [__meta_kubernetes_service_annotation_example_io_should_be_scraped]
-  #    action: keep
-  #    regex: true
-  #
-  # Example relabel to customize metric path based on endpoints
-  # "example.io/metric_path = <metric path>" annotation.
-  #  - source_labels: [__meta_kubernetes_service_annotation_example_io_metric_path]
-  #    action: replace
-  #    target_label: __metrics_path__
-  #    regex: (.+)
-  #
-  # Example relabel to scrape only single, desired port for the service based
-  # on endpoints "example.io/scrape_port = <port>" annotation.
-  #  - source_labels: [__address__, __meta_kubernetes_service_annotation_example_io_scrape_port]
-  #    action: replace
-  #    regex: ([^:]+)(?::\d+)?;(\d+)
-  #    replacement: $1:$2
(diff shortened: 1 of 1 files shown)


brian-brazil:
We purposefully removed these, as users were taking them to be The way to do things rather than merely an example of something you could do. @bwplotka

sylr:
Intended or not I think those became standard and I don’t see how it improves the user experience to replace them by a placeholder especially given that most of helm charts use those (https://github.com/helm/charts/search?utf8=%E2%9C%93&q=prometheus.io%2Fscrape&type=).

I think a config example working out of the box and compatible with most deployments and helm charts out there has more value for the users.

I’m glad I had theses when I started, god only knows what kind of messed up config I would have end up with otherwise.

bwplotka:
Hi @sylr :wave: (: Feel free to use those, but I don't think Prometheus should keep this as an example and we already talked about this multiple times. 

Essentially it gives false impression that `prometheus.io` is somehow magically used by prometheus, whereas it's not until you use it in relabelling on your own. This created much more confusion in the past. We also did not remove any examples. You have the example how you do annotation based scrape choice it if you want, we just moved `prometheus.io` to `example.io` to explicitly show that you can put in label name whatever you want and that user is responsible for handling this in relabelling. (:

We are defintely happy to discuss potential improvement in examples and docs, but we agreed to not confuse users with hardcoded `prometheus.io`. 

> I think a config example working out of the box and compatible with most deployments and helm charts out there has more value for the users.

**Personally I think it's not. If this is now a standard I would advice to revisit it as it does not support multiple containers.** 

You can find overview of your options here: https://docs.google.com/document/d/1S6O1czHtjR2DGfK2zeZDLr88wxyLRXB2JpH5g7YM1J8/edit?usp=sharing

sylr:
Hi @bwplotka I understand your stance, however I still feel giving a turnkey configuration is something that new users would appreciate.

I believe that the less time you spend getting things running the more you'll appreciate a software. Also I think it's easier to understand things when you have a working example. 

The `prometheus.io/...` annotations is enough for 95% of kubernetes deployments out there. We could warn that the example would not work in a scenario where pods have several containers exposing metrics.

fischerman:
Hi @bwplotka, if Prometheus doesn't not encourage any sort of convention (or even standard) around `prometheus.io/...` annotations, than I agree that they should not be in the examples. However, you can find the annotations all over the internet. As such, I was expecting to find some documentation about them in the Prometheus docs. I think the docs would greatly benefit from mentioning them. Even if it is just to say that their use is discouraged.

PR #5654, opened 2019-06-12
 Update quickly testing commands

qudongfang:
1. Update cmd in contributing.


cstyan:
Seems sane to me, not sure about the changelog changes though. I think if we wanted those URLs to act as links we would have done so?

cc @brian-brazil

brian-brazil:
Github already makes them links. I'd rather not make bulk changes to the CHANGELOG.

qudongfang:
Thanks for your review. I closed this PR.

PR #5658, opened 2019-06-13
Work around golangci-lint issues

matthiasr:
The version on goreleaser is broken and different from the one in
Github: golangci/golangci-lint#575 – use the direct Github URL as a
workaround.

Additionally, the installer script attempts to `install -d /go/bin` on
CircleCI, which doesn't work due to permissions. Patch that line out –
we don't need it because we make sure the directory exists anyway.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -248,7 +248,9 @@ proto:
 ifdef GOLANGCI_LINT
 $(GOLANGCI_LINT):
 	mkdir -p $(FIRST_GOPATH)/bin
-	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(FIRST_GOPATH)/bin $(GOLANGCI_LINT_VERSION)
+	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/$(GOLANGCI_LINT_VERSION)/install.sh \
+		| sed -e '/install -d/d' \
+		| sh -s -- -b $(FIRST_GOPATH)/bin $(GOLANGCI_LINT_VERSION)
 endif
 
 ifdef GOVENDOR


matthiasr:
Ok, looking at this with tracing shows that at least this gets us further along … now what fails is `install -d /go/bin`. How did anything ever work? 🤔

matthiasr:
`/go/bin` already exists and is owned by root. it has mode rwxrwxrwx. the golangci-lint installer runs as `circleci`, and `install -d` tries to "fix" the permissions but can't.

simonpasquier:
> How do Makefile.common changes get propagated?

There's a [schedule job](https://github.com/prometheus/prometheus/blob/55f3a9fe4ae40a3b09f6176511f516efcae06958/.circleci/config.yml#L146-L156) running in Circle CI

Makefile.common
```diff
@@ -247,8 +248,12 @@ proto:
 
 ifdef GOLANGCI_LINT
 $(GOLANGCI_LINT):
+	-ls -lah $(FIRST_GOPATH)/bin
 	mkdir -p $(FIRST_GOPATH)/bin
-	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(FIRST_GOPATH)/bin $(GOLANGCI_LINT_VERSION)
+	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh \
```
simonpasquier: someone on the upstream issue commented that one could use `https://raw.../$(GOLANGCI_LINT_VERSION)/install.sh` to be even safer.

matthiasr: 👌

matthiasr:
I consolidated the changes and updated the PR description, PTAL!

Issue #5655, opened 2019-06-12
Display active alerts on top


kustodian:
## Proposal
https://github.com/prometheus/prometheus/pull/5448 introduced grouping of alerts by file. This makes alerts page look nice and tidy, but the problem is that it's very hard to see which alerts are currently active because if you have a lot of alerts you have to scroll down to find an active alert. We mostly used the alerts page to see at a glance which alerts are still active.

mszczygiel:
Please see my comment https://github.com/prometheus/prometheus/pull/5664#issuecomment-502809410
I think it may be better to add filtering instead of changing order of groups as it may be annoying to see groups in effectively random order.

TyBrown:
+1 on this.

We've got 285 alerts across 44 different alert groups... the change to group by alert groups has made the Alerts page intractable for us, and we would seriously consider staying on a version of Prometheus which doesn't have this change, except that there are some important ZookeeperSD fixes in later versions that we are eager to take advantage of.

We are very excited about filtering the alerts page, however, it would be nice if we could disable the alert group headings altogether as well, since we run Prometheus under Mesos with native cgroups, so the filesystem paths are not friendly on the Web UI.

![image](https://user-images.githubusercontent.com/7005728/62576768-1d0a5400-b863-11e9-98df-939b79e81415.png)

Transwert:
@TyBrown @codesome @mszczygiel @mszczygiel   can you state about this issue, if this has been solved, if yes, can you close it. and if not, can you state, what is needed to be done.

simonpasquier:
@Transwert you can follow #5758 which should fix this issue.

beorn7:
We are currently running a bug scrub with the team.  We assume the filtering we have in current Prometheus versions (implemented in #5758) is good enough to fix this issue. Please follow-up if you disagree.

PR #5664, opened 2019-06-14
Display active alerts on top

mszczygiel:
Closes https://github.com/prometheus/prometheus/issues/5655

### DIFF
--- a/web/web.go
+++ b/web/web.go
@@ -508,6 +508,10 @@ func (h *Handler) alerts(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	sort.Slice(groups, func(i, j int) bool {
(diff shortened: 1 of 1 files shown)


brian-brazil:
This would make the order of rule groups basically random if you're trying to look at a specific group.

mszczygiel:
I agree that following a particular group can be annoying now. Maybe it could be done with simple filtering (like on `Targets` tab where there are `All/Unhealthy` buttons). I could put buttons like `All`, `Active`, `Firing` on the Alerts tab where `Active` would contain both pending and active alerts. What do you think?

kustodian:
Filtering would be great. Also IMO the default should be to only display active alerts.

mszczygiel:
I will close this PR in favour of https://github.com/prometheus/prometheus/pull/5758 where instead of sorting I added filters.

PR #5669, opened 2019-06-14
added dockerfile path and build context

geekodour:
`DOCKERFILE_PATH` was added to `Makefile.common` in #5635

This PR changes that to be passed to the `-f` flag and adds a new variable `DOCKERBUILD_CONTEXT` which specifies the build context for the docker build.

more info:
https://docs.docker.com/engine/reference/commandline/build/#build-with-path

cc  @simonpasquier  @SuperQ 


geekodour:
ping @SuperQ

PR #5670, opened 2019-06-14
Added humanizePercentage formatting to templates

JensErat:
Lots of alerts are based on ratios (eg. disk usage), and humans are used
to values in percentage in textual descriptions.

Signed-off-by: Jens Erat <email@jenserat.de>

@brian-brazil


brian-brazil:
Thanks!

Issue #5668, opened 2019-06-14
/api/v1/series HTTP API does not have dedicated method in Querier interface


dmitsh:
## Proposal
**Use case. Why is this important?**
There is no dedicated method for `/api/v1/series` API in the `Querier` interface. It is implemented via `Select` method, similar to `/api/v1/query_range`.
For TSDB it seems not to be a problem, as the `SeriesIterator` is not used by `/api/v1/series`, therefore the samples are not pulled.
However, for `remote_read` operation `Select` returns the entire time-series, including the samples, which are later discarded when serving `/api/v1/series`
This results in significant overhead in traffic and data processing from both Prometheus server and `remote_read` endpoint.

My proposal would be to add a new `Series` method to `Querier` interface.
I understand that this is a breaking change, and I suggest to make it in two steps.
First step is to add the `Series` method and temporarily implement it by calling `Select` under the hood. That will provide backward compatibility, and expose updated `Querier` interface to the community.
Second step would require support of `Series` method by exiting `remote_read` implementations, after which we can replace the temporary implementation.

dmitsh:
I have a working prototype, which I would gladly turn into a PR, if there are no objections to this proposal

jacksontj:
This seems like a dupe of #4057 (for more context).

brian-brazil:
Yip, that's a dupe. I don't believe anyone has worked on this yet, so a PR would be appreciated.

dmitsh:
Thanks @jacksontj , I agree, the idea is the same, although both issues are open, and #4057 has diverted a bit.
I'm totally OK with closing this issue in favor of #4057, but I'd like to continue the discussion.
I understand that there were two proposals:
1. add a flag in `SelectParams` to prevent sample from being sent
2. add a new method in `Querier` to query series metadata

My proposal aligns with the second approach.
I agree that the first approach is simpler but a bit "hacky".
Second approach is cleaner, but introduces a breaking change.

@brian-brazil, any preferences?

PR #5672, opened 2019-06-14
setting SelectParams.Metadata=true in Querier.Select() for /api/v1/series HTTP API call

dmitsh:
setting SelectParams.Func="metadata" in Querier.Select()
for /api/v1/series HTTP API call to instruct remote_read endpoint
to omit time-series samples in the response.


dmitsh:
This PR addresses  issues #4057 and #5668

Based on the discussion, there were two proposals:

1. adding a new method in Querier interface to query time-series metadata
2. adding a flag in SelectParams to omit time-series sample in the response to /api/v1/series

While the first approach is clean, it introduces breaking changes in remote_read API.
The second approach is simpler, but a bit hacky.

In this PR I'm trying to make second approach less hacky.
I'm proposing to pass new `metadata` function as `ReadHints.Func` in `Select()` call for /api/v1/series.
This change preserves backward compatibility, and provides flexibility to remote_read maintainers.

jacksontj:
IMO this `Func` thing is in itself hacky, the `Func` field is used to annotate which aggregation/function call the query is in (https://github.com/prometheus/prometheus/blob/master/storage/interface.go#L73) -- so using it to signal "metadata" seems hacky.

If you want to stick with this #2 approach there is already a workaround in-place-- the SelectParams are nil. And for the remote-read calls this means the `ReadHints` are nil (https://github.com/prometheus/prometheus/blob/0c0638b080cf1565fc4e5b7ee3fd35c36ae5832a/storage/remote/codec.go#L85).

IMO this change (using the func) is just as "destructive" as #1 since all current users of the current workaround would be broken. In your use-case is checking for nil ReadHints not viable? Or just unknown (it is an undocumented workaround AFAIK).

brian-brazil:
An explicit field in select params would make more sense. The general idea with SelectParams is that we don't know what remote storages want, so we'll add things that make sense as needed.

dmitsh:
@jacksontj I agree with you, having `SelectParams` equal to `nil` is a good indication that it was set in web/api/v1/api.go:series()
@brian-brazil I've addressed your comment

web/api/v1/api.go
```diff
@@ -491,7 +491,8 @@ func (api *API) series(r *http.Request) apiFuncResult {
 	var sets []storage.SeriesSet
 	var warnings storage.Warnings
 	for _, mset := range matcherSets {
-		s, wrn, err := q.Select(nil, mset...) //TODO
+		// SelectParams.Metadata=true should instruct remote_read endpoint to send labels only (without samples)
+		s, wrn, err := q.Select(&storage.SelectParams{Metadata: true}, mset...)
```
jacksontj: My concern is that this line specifically will cause anyone using the old check (if nil) to start fetching all data.

brian-brazil: We can mention it as a CHANGE in the changelog.

jacksontj: IMO simply changing from having to check whether its a nil to checking a single field for a bool doesn't seem like enough improvement (if any) to justify the breaking change. If this were going to start adding *more* information I could maybe be onboard, but at this point its moving a boolean check from the current "is nil" to this "metadata" field -- which IMO seems like an unnecessary change (which isn't providing any value-- at least not that I'm seeing in this PR).

dmitsh: I tend to agree with @jacksontj . The main benefit of this commit is that it brings more clarity to what the remote_read endpoint would expect to do. There is no functional change in the code.

Originally I intended to introduce a new method to the `Querier` interface for querying the metrics (labels). I actually have that change in my other branch. But then I realized that the new method will have almost identical signature with the `Select` method: it will receive a list of `labels.Matcher`s and return a `SeriesSet`. So I dropped that idea.

@brian-brazil @jacksontj leaving it up to you guys.

jacksontj: Sounds like maybe the solution for now is to better document the `ReadHints` + `SelectParams` nil meaning?

brian-brazil: That makes sense.

PR #5673, opened 2019-06-14
Add storage.Warnings to LabelValues and LabelNames

### LINKED ISSUE #5661: LabelValues and LabelNames missing storage.Warnings

I've been working on adding the Warnings responses to the go client, in doing so I have noticed that the `LabelValues` and `LabelNames` API calls do not return warnings -- but seem like they should. The `Warnings` are currently returning any time there is an error doing some remote read (and presumably for other non-fatal errors)-- which IMO means that this field should exist for *any* API call that requires coordinating with remote read endpoints -- and `LabelValues` and `LabelNames` both do.

jacksontj:
Fixes #5661

### DIFF
--- a/promql/engine_test.go
+++ b/promql/engine_test.go
@@ -173,9 +173,9 @@ type errQuerier struct {
 func (q *errQuerier) Select(*storage.SelectParams, ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
 	return errSeriesSet{err: q.err}, nil, q.err
 }
-func (*errQuerier) LabelValues(name string) ([]string, error) { return nil, nil }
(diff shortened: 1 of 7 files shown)


brian-brazil:
Thanks!

A thought I had yesterday: To avoid breaking the Go API in the future, might it be better to have a general extra information struct returned instead of warnings? That way if we e.g. add timings at some point in the future don't have to add an extra return value.

jacksontj:
Some sort of "result metadata", I think that makes some sense. I'd initially define it as all data related to the request which shouldn't have any impact on the result data (e.g. non-fatal errors, timings, etc.) -- nothing like errors etc. should be in there.

PR #5675, opened 2019-06-15
Fix zk connection leak

meridional:
Fixing the same issue mentioned in #5095 and #5093.

This works by creating a wait group for each tree cache. And from within the Stop function, we close the updates channel after all members of the wait group, who are basically the only producers to the updates channel, have exited. With that, this loop https://github.com/prometheus/prometheus/blob/master/discovery/zookeeper/zookeeper.go#L170 can terminate and the connection to zk can be properly shut down.

I tested this by running prometheus against a local zookeeper, and reloading prometheus does not increment the watch count after this patch.

### DIFF
--- a/util/treecache/treecache.go
+++ b/util/treecache/treecache.go
@@ -121,14 +121,16 @@ func (tc *ZookeeperTreeCache) Stop() {
 	}()
 	go func() {
 		tc.wg.Wait()
-		// close the tc.head.events after all go rountines have exited so that
-		// the one above will exit
+		// close the tc.head.events after all members of the wait group have exited
+		// this makes the go routine above exit
 		close(tc.head.events)
 		close(tc.events)
 	}()
 }
 
 func (tc *ZookeeperTreeCache) loop(path string) {
+	defer tc.wg.Done()
+
 	failureMode := false
 	retryChan := make(chan struct{})
 
@@ -205,7 +207,6 @@ func (tc *ZookeeperTreeCache) loop(path string) {
 			// stop head as well
 			tc.head.done <- struct{}{}
 			tc.recursiveStop(tc.head)
-			tc.wg.Done()
 			return
 		}
 	}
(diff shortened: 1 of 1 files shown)


util/treecache/treecache.go
```diff
@@ -243,7 +264,9 @@ func (tc *ZookeeperTreeCache) recursiveNodeUpdate(path string, node *zookeeperTr
 		}
 	}
 
+	tc.wg.Add(1)
 	go func() {
+		defer tc.wg.Done()
 		numWatchers.Inc()
 		// Pass up zookeeper events, until the node is deleted.
 		select {
```
brian-brazil: Hmm, are we missing a for loop here?

util/treecache/treecache.go
```diff
@@ -243,7 +264,9 @@ func (tc *ZookeeperTreeCache) recursiveNodeUpdate(path string, node *zookeeperTr
 		}
 	}
 
+	tc.wg.Add(1)
 	go func() {
+		defer tc.wg.Done()
```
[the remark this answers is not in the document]
meridional: Fixed.

util/treecache/treecache.go
meridional: Are you talking about a for loop around the select statement?

I don't think a for loop is needed here because `childWatcher` and `dataWatcher` will only fire once. The `loop` function takes care of re-watching the zk nodes.

jjneely:
I can give this a spin....it will take a couple days though....

meridional:
> I can give this a spin....it will take a couple days though....
We are also testing this in our staging environment, and so far so good.

util/treecache/treecache.go
```diff
@@ -94,24 +96,41 @@ func NewZookeeperTreeCache(conn *zk.Conn, path string, events chan ZookeeperTree
...
 func (tc *ZookeeperTreeCache) Stop() {
 	tc.stop <- struct{}{}
+	go func() {
+		// drain tc.head.events so that go routines can make progress and exit
+		for range tc.head.events {
+		}
+	}()
+	go func() {
+		tc.wg.Wait()
+		// close the tc.head.events after all members of the wait group have exited
```
brian-brazil: Comments should start with a capital letter and end with a full stop.

brian-brazil:
Thanks!

PR #5681, opened 2019-06-17
Merge master into mixin

beorn7:
Merge master into mixin to get rid of CI failures.


beorn7:
@cstyan @tomwilkie Does it make sense this way? We could also rebase the mixin branch on top of master. (But that will confuse everybody with the mixin branch checked out and being worked on.)

We could also just not do it and live with the CI errors.

DCO fails inevitably because master has some un-signed commits.

brancz:
Fwiw the mixin is probably consumed via jsonnet-bundler in which case a rebase wouldn’t make a difference, so at most people who have existing commits on top of the existing branch would be affected (I realize there is already a PR against this one. In my opinion we should just merge the remote write dashboard directly into this and rebase everything altogether. Just my 2 cents, ultimately whatever works :) .

beorn7:
PR #5601 is merged now. @cstyan @tomwilkie your pick: Merge this PR, rebase mixin on top of master, or do nothing?

beorn7:
Talked to @cstyan and decided to take the conservative option and merge master in instead of rebasing.

PR #5694, opened 2019-06-20
Integrate beginning of React UI

juliusv:
This integrates the beginning of a new React web UI, for now only serving a new graph page at `/static/graph-new/app.html`. I'm not linking it from anywhere yet for now.

There's still a lot to do here, but I want to get this merged so other people can help fix / clean up / complete it.


NOTICE
```diff
@@ -85,3 +85,8 @@ go-zookeeper - Native ZooKeeper client for Go
 https://github.com/samuel/go-zookeeper
 Copyright (c) 2013, Samuel Stauffer <samuel@descolada.com>
 See https://github.com/samuel/go-zookeeper/blob/master/LICENSE for license details.
+
+We also include code from a variety of npm packages. For details, see
```
juliusv: Waiting on further input from @swinslow here on how to properly attribute npm package dependencies.

juliusv:
Whoops, still some demo.robustperception.io server mentions in the code, will update.

juliusv:
...and I'll need to teach CircleCI about npm.

juliusv:
Hmm, in order to get the CircleCI assets diff test not to fail, we would either have to get reproducible builds working (https://spin.atomicobject.com/2016/12/16/reproducible-builds-npm-yarn/ suggests switching from npm to yarn for that) or exempt the generated React assets from the diffing for now.

juliusv:
@brancz

juliusv:
Exempted the generated React files from the diffing for now, but that obviously doesn't help since the assets_vfsdata.go contains the diffs as well. Grumble.

simonpasquier:
I'm not at all a JS expert but I've seen colleagues [switching](https://groups.google.com/forum/#!topic/kiali-dev/wGaxznuC20A) from npm to yarn for similar reasons. I can ping them for advice if needed.

juliusv:
Thanks! Trying out migrating to yarn now and seeing if that helps.

juliusv:
Migrated to yarn, but another problem is that `react-scripts build` also considers the absolute source filenames for the output file hashes, so the output files are still different on CI :( Copying everything to a predictable location before building (like `/tmp/react-build`) might work, but seems icky.

juliusv:
Coming back to this PR after a while.

The problems left here are:

* It doesn't seem feasible to do 100% reproducible builds even using Yarn, since the full local build directory path is baked into the build, and that directory is different on different machines.
* We need to include licenses for all the ~3000-ish npm packages in `node_modules` somehow.

I see two options:

* Check in built assets (resulting from `yarn build`), but exempt them from CI build & diffing. Drawback: we can be less sure that the built result really corresponds to the source TypeScript files. Also, this would require us to distribute npm dependency licenses in our code (not just the release tarball).
* Only build final assets during the build phase. Drawbacks: requires Node during build, making it impossible to build Prometheus with just Go, and also makes the build process take longer and depend more on the internet. Pro: we know the result is built off the right source TypeScript, and we only need to include npm dep licenses in the release tarball.

If I read things correctly, Grafana follows the latter approach. You can still build it with just Go without building frontend assets, but the frontend will not work correctly without it. @davkal can you confirm?

juliusv:
@csmarchbanks Yeah, good point. I had totally forgotten about the automatically created initial test :) I just made that minimal test pass now and created a new `react-app-test` target in the main Makefile. I'm skipping React tests if `yarn` is not installed because I don't want to annoy developers more than needed before the React app becomes more relevant. What do you think?

juliusv:
@csmarchbanks Oops. The `react-app-test` target needs to depend on the node_modules target, so that needs yarn too, so to keep things simpler I just didn't make the `test` target depend on `react-app-test` for now, so one has to run it separately. But I added it to the `all` target and Travis & CircleCI... let's see if they pass...

boyskila:
Hi! I'm not sure whether this is the right place for this topic (if its not please point me to the right place where I can ask questions :)) but do you think guys to get rid of jquery at some point? As far I can see is mostly used as a datepicker and for data vizualization. There are pretty good alternatives I think, Yes, no so tiny as JQ but I guess some kb more are not going to be a problem. What do you think?

juliusv:
@boyskila It would be nice to get rid of jQuery 100%, but I haven't managed to yet. For example, the native React datepickers I tried out in February (when I did most of this work) seemed not quite what I was looking for (either in terms of stylability or behavior). Also, the graphing library (Flot) that Grafana uses and that I'm switching to here as well, requires jQuery.

Overall I'm fine with jQuery still being used in isolated components here and there, as long as the overall app is React-managed. But if someone does find fitting native React components to replace the jQuery ones, that's good too.

Makefile.common
```diff
@@ -109,7 +109,7 @@ endif
 %: common-% ;
 
 .PHONY: common-all
-common-all: precheck style check_license lint unused build test
+common-all: precheck style check_license lint unused build test react-app-test
```
csmarchbanks: Isn't this file propagated to other repos that would not have `react-app-test`?

juliusv: Whoops of course - I carefully avoided modifications to it up until the last changes where I stopped thinking :D Will look & fix tomorrow.

juliusv:
@csmarchbanks I just removed the local modification to `Makefile.common`. That means that CircleCI won't run the React tests for now, but at least Travis will. Long-term that should be changed somehow...

juliusv:
FTR, the `prometheus` binary grows by around 1.5% with these changes (78598920 vs. 79795185 bytes).

.travis.yml
```diff
@@ -12,8 +12,11 @@ go_import_path: github.com/prometheus/prometheus
...
 before_install:
 - travis_retry make deps
+- . $HOME/.nvm/nvm.sh
+- nvm install stable
+- nvm use stable
 - if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then choco install make; fi
 
 script:
-- make check_license style unused test lint check_assets
+- make check_license style unused test react-app-test lint
```
simonpasquier: I have https://github.com/prometheus/prometheus/pull/5932 to "fix" the Makefile default target and have Circle CI run the `check_assets` target too. Any reason why the React tests can't run on Circle? Eventually I'd like us to keep only one CI system (eg Circle CI since it manages releases too).

juliusv: The React tests actually run fine on Circle, but the Prometheus orb on Circle just runs `make test`, and I didn't want to drag in the React tests into the normal `test` target yet to not unnecessarily annoy developers. Maybe I just should, since `make build` also requires node + yarn?

Makefile
```diff
@@ -25,19 +31,44 @@ include Makefile.common
...
 
-.PHONY: check_assets
-check_assets: assets
-	@echo ">> checking that assets are up-to-date"
-	@if ! (cd $(PREFIX)/web/ui && git diff --exit-code); then \
-		echo "Run 'make assets' and commit the changes to fix the error."; \
-		exit 1; \
-	fi
+.PHONY: npm_licenses
+npm_licenses:
```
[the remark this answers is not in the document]
juliusv: Hm yeah, I left it out because `make tarball` and `make docker` also already just assume everything they require is already present. But I don't really see a reason to follow that, so added the dependency now.

Makefile
```diff
@@ -25,19 +31,44 @@ include Makefile.common
...
+	@echo ">> bundling npm licenses"
+	rm -f $(REACT_APP_NPM_LICENSES_TARBALL)
+	find $(REACT_APP_NODE_MODULES_PATH) -iname "license*" | tar cfj $(REACT_APP_NPM_LICENSES_TARBALL) --transform 's/^/npm_licenses\//' --files-from=-
+
+.PHONY: tarball
+tarball: npm_licenses
+	$(MAKE) common-tarball
```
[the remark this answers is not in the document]
juliusv: Oh yeah. Is there a specific reason for using a double-colon rule here? It seems to work with a single one as well, and not sure why it would need a double.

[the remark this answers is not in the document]
simonpasquier: The double-colon avoids the warning message when you redefine the target.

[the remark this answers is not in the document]
juliusv: Hm, I'm not getting any warning messages with a single colon, or do you mean in case we define it multiple times in the future?

.travis.yml
simonpasquier: Hmm, I don't think the Prometheus orb is involved here. I'm fine with adding an additional `make react-app-test` step to the Circle CI job if we don't want to extend `make test`.

https://github.com/prometheus/prometheus/blob/3c7028480271a70fac2d1958d6bb2fdf47b2b7de/.circleci/config.yml#L22-L38

web/ui/react-app/src/ExpressionInput.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  componentDidMount() {
+    const $exprInput = $(this.exprInputRef.current!);
+    $exprInput.on('input', () => {
+      const el = $exprInput.get(0);
+      const offset = el.offsetHeight - el.clientHeight;
+      $exprInput.css('height', 'auto').css('height', el.scrollHeight + offset);
+    });
+  }
+
```
boyskila: I think height should be set on DidMount as well since now on page refresh Input is taking its initial size and text remains hidden. Btw, this can be replaced very easily with
`() => this.setState({ height: 'auto' }, () => this.setState({ height: this.inputRef.current.scrollHeight }))` attached as onInput handler and height as style attr

Makefile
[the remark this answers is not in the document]
simonpasquier: My bad, `::` isn't needed, single colon is fine.

.travis.yml
juliusv: Ah right, it just calls `make` directly in the checked-in config. I did now just re-add the React tests to the normal `test` target since anyone building Prometheus will now need to have Node+Yarn installed anyway. I also removed the reference to `check_assets` that you added in a recent PR, since that target no longer exists.

juliusv:
@simonpasquier I believe all oustanding things have been addressed for now and all CI is passing - good to merge? :)

web/ui/react-app/src/ExpressionInput.tsx
```diff
@@ -0,0 +1,159 @@
...
+                  style: {
+                    backgroundColor:
+                      downshift.highlightedIndex === index ? 'lightgray' : 'white',
+                    fontWeight: downshift.selectedItem === item ? 'bold' : 'normal',
+                  },
+                })}
+              >
+                {/* TODO: Find better way than setting inner HTML dangerously. We just want the <strong> to not be escaped.
+                    This will be a problem when we save history and the user enters HTML into a query. */}
+                <span dangerouslySetInnerHTML={{__html: item.string}}></span>
```
codingdash: To address this concern [sanitize-html](https://www.npmjs.com/package/sanitize-html) library can be used which allows whitelisting certain HTML elements to be allowed.

juliusv: Nice! Would you be up for doing a PR after this one? The only thing we need to watch out for is that it still performs well for thousands of list entries, but I hope that will be fine.

At the moment, there is no query history search in the new UI yet, so it's not an immediate vulnerability.

codingdash: Yes, I'd love to.

juliusv: 🙌 Can't wait for people who actually know frontend / React stuff to contribute and make this all better :)

web/ui/react-app/src/ExpressionInput.tsx
boyskila: Yep, it works fine but overall is better to delegate this job to react since another issue appears here. If the user clears the input (which contains multiple lines) and then the input loses its focus the query will appear again without input to be resized. This is happening because Downshift set its selectedItem with the value coming from this.props which in this case never changes. So, this can be resolved by making the input controlled and value to be saved in the local state together with the height of the input. I can make PR for this if you don't have mind :)
Also I was wonder whether redux can be integrated in the App. I'm not sure how much is going to grow but for sure will simplify the components and later on will be easier to extend end test

simonpasquier:
LGTM!

web/ui/react-app/src/ExpressionInput.tsx
juliusv: @boyskila Yeah, it would be great to get a PR for improving this :) Getting Downshift to work just the way I wanted required a bit of hackery and dirtiness, and I hope there's better / cleaner / more efficient ways of doing it.

As for Redux, I've heard different opinions from React people, so at least I would like to consult more people before introducing it.

beorn7:
Aaaaand security warnings about react components coming… :smile:

juliusv:
@beorn7 Nice. Will check them out! Most of this code is from February, so I hope it's mostly about module upgrades being necessary.

web/ui/react-app/src/ExpressionInput.tsx
codingdash: Created PR which addresses this - https://github.com/prometheus/prometheus/pull/6165/files

boyskila:
@juliusv, any plans to add "enzyme"?

juliusv:
@boyskila As someone new (again) to the frontend, I've never heard of it :) I think I need to get a decent overview of the different testing frameworks that exist for React apps, and their pros and cons.

boyskila:
Yes, sorry my question wasn't correct. I was wonder whether the testing environment gonna be enhanced with some utilities like "Enzyme". I'm also new to the open-source and became interesting in how things happen. So, im looking forward to seeing what's gonna happen with this and for now, i'll skip writing tests :)

web/ui/react-app/src/ExpressionInput.tsx
boyskila: @juliusv this PR https://github.com/prometheus/prometheus/pull/6174 is for the ExpressionInput clearing issue on blur.

.promu.yml
```diff
@@ -12,7 +12,7 @@ build:
           path: ./cmd/promtool
         - name: tsdb
           path: ./tsdb/cmd/tsdb
-    flags: -mod=vendor -a -tags netgo
+    flags: -mod=vendor -a -tags netgo,builtinassets
```
martingrambow: this creates
> cmd/go: -tags space-separated list contains comma
when running make build

juliusv: You'll need Go 1.13 as specified in this file as well, see https://github.com/prometheus/prometheus/pull/6308

martingrambow: Thx

PR #5695, opened 2019-06-20
web: fix flaky TestHTTPMetrics()

simonpasquier:
I spotted a transient failure in the CI with `TestHTTPMetrics`. Obviously testing metrics with parallel tests enabled and using the global registry at the same time isn't a great idea...


brancz:
lgtm :+1:

would love to see all of prometheus codebase using injected registries, like tsdb already does and thanos :)

simonpasquier:
Yep this is one step forward :)

PR #5698, opened 2019-06-21
Add label of a group name to rule_group_iterations_missed_total metric 

homelessnessbo:
Signed-off-by: Ivan Kiselev <kiselev_ivan@pm.me>

Right now it's impossible to understand what group of rules is struggling without manually analyzing config. Adding `rule_group ` should make this process a bit easier.

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -66,7 +66,7 @@ type Metrics struct {
 	evalFailures        prometheus.Counter
 	evalTotal           prometheus.Counter
 	iterationDuration   prometheus.Summary
-	iterationsMissed    prometheus.Counter
+	iterationsMissed    *prometheus.CounterVec
 	iterationsScheduled prometheus.Counter
 	groupLastEvalTime   *prometheus.GaugeVec
 	groupLastDuration   *prometheus.GaugeVec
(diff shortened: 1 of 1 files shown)


rules/manager.go
```diff
@@ -102,16 +102,19 @@ func NewGroupMetrics(reg prometheus.Registerer) *Metrics {
 			Help:       "The duration of rule group evaluations.",
 			Objectives: map[float64]float64{0.01: 0.001, 0.05: 0.005, 0.5: 0.05, 0.90: 0.01, 0.99: 0.001},
 		}),
-		iterationsMissed: prometheus.NewCounter(prometheus.CounterOpts{
-			Namespace: namespace,
-			Name:      "rule_group_iterations_missed_total",
-			Help:      "The total number of rule group evaluations missed due to slow rule group evaluation.",
-		}),
 		iterationsScheduled: prometheus.NewCounter(prometheus.CounterOpts{
```
bwplotka: Can we change this one as well?

brian-brazil:
This counts as a breaking change, you can already figure this out from rule_group_last_duration_seconds.

homelessnessbo:
@brian-brazil then, to come up with an actual alert - another metric is needed that actually reflects the evaluation interval for certain group

brian-brazil:
rule_group_interval_seconds covers that.

homelessnessbo:
yep. You're totally right, the combination of those two gives the opportunity to alert. 

Thanks!

bwplotka:
Wow, I just noticed it's not a Thanos project 0.o I think this is a good sign ;p

PR #5699, opened 2019-06-21
add take relabel action

gitwillsky:
Add `take` relabel action for scrape whitelist metrics

Usage:
```yaml
    metric_relabel_configs:
      - source_labels: ['__name__', 'cpu', 'mode']
        regex: 'node_cpu_seconds_total;0;user'
        action: 'take'
      - source_labels: ['__name__', 'cpu', 'mode']
        regex: 'node_cpu_seconds_total;1;idle'
        action: 'take'
```

### DIFF
--- a/pkg/relabel/relabel.go
+++ b/pkg/relabel/relabel.go
@@ -46,6 +46,8 @@ const (
 	Keep Action = "keep"
 	// Drop drops targets for which the input does match the regex.
 	Drop Action = "drop"
+	// Take takes targets for which the input does match the regex.
+	Take Action = "take"
 	// HashMod sets a label to the modulus of a hash of labels.
 	HashMod Action = "hashmod"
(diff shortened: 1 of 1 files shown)


brian-brazil:
Can you explain more about your use case? This seems to be handled by the existing keep/drop actions.

gitwillsky:
Forgive my poor english, :)
If user only want to scrape some metrics, like this:
```bash
a {cpu="cpu0", mode="idle"}
b {type="1",mode="user"}
```
`keep` action config will like this:
```yaml
 - source_labels: ['__name__', 'mode']
        regex: '(?:a|b);(?:idle|user)'
        action: 'keep'
```
But when a metric like this:

```bash
b {type="1",mode="idle"}
```
It will also be keep

brian-brazil:
What you want is `a;idle|b;user`

gitwillsky:
Suddenly realized，thank you.

Issue #4517, opened 2018-08-19
Add streaming to Remote Read endpoint


gouthamve:
Remote read is quite inefficient right now and could be made much better through streaming. This has been discussed in the dev-summit: https://docs.google.com/document/d/1-C5PycocOZEVIPrmM1hn8fBelShqtqiAmFptoG4yK70/edit#heading=h.5d2mc4i0bnnp

Bplotka:
I can take it if no one started working on this.

Design doc: https://docs.google.com/document/d/1JqrU3NjM9HoGLSTPYOvR217f5HBKBiJTqikEB9UiJL0/edit?usp=sharing

Dev-mailing topic: https://groups.google.com/forum/#!topic/prometheus-developers/79b6j_6F_Vc

harshal-shah:
According to https://github.com/prometheus/prometheus/pull/4655 it seems this has been resolved in prometheus 2.5.0

rshivane:
I would like to add that, as a vendor implementing the server-side of remote read API, the lack of streaming adds similar memory pressure on the marshaling side.

@bwplotka I notice that your already have a PR for streaming support in #4591. Whenever this is approved for merge, we would be happy to implement it on our storage and validate.

bwplotka:
Yes. This is working code for streaming with raw samples, but we have some internal discussion how it should be done. I closed it to reduce confusion for now: https://github.com/prometheus/prometheus/pull/4591

TL;DR: It would be nice to have remote read stream with Prometheus chunk encoded samples (instead of row samples), to avoid unnecessary decoding & encoding. Does it 
The plan is to:
1. Change the doc to propose this idea (right now it proposes something in between)
2. Setup a call to gather feedback on this idea.

Did not have much time to jump into it yet, but I am happy to hear that you are ok to help with this!

rshivane:
>  It would be nice to have remote read stream with Prometheus chunk encoded samples (instead of row samples), to avoid unnecessary decoding & encoding.

Is that essentially ["Alternative 1" in the design doc](https://docs.google.com/document/d/1JqrU3NjM9HoGLSTPYOvR217f5HBKBiJTqikEB9UiJL0/edit#heading=h.12fpmrn0ayj) ?

bwplotka:
Exactly, but need to reorder ;p

bwplotka:
[...]

PR #5701, opened 2019-06-21
remote-read: [1] Extended remote read proto with negotiation and chunked response.

bwplotka:
First part of chained PRs to add streamed, chunked remote read. 

Part of: https://github.com/prometheus/prometheus/issues/4517 and https://github.com/improbable-eng/thanos/issues/488

## Notes for reviewers:
To fully understand decisions and how usage will look like feel free to see all changes together [here](https://github.com/prometheus/prometheus/pull/5703).

.. and caller side changes (Thanos): https://github.com/improbable-eng/thanos/pull/1268

I thought it will be much easier to discuss changes in much smaller chunks. 

PTAL @brian-brazil @tomwilkie @gouthamve @krasi-georgiev and others. Feedback very welcome.

Concerns from my side:
* Should we support non-streamed XOR Chunk. IMO only streamed makes sense.
* Should we support streamed Sampled response? IMO not.
* Should we do content negotiation based on request headers instead? I think proto is way more explicit and will help in maintaing compatibility.
* Should we split responseType and negotiate `streaming` `encoding` and `compression` in three different enums / `ReadRequest` parameters? It will be awful on implementation site and user will expect every permutation to work. That's why I left only two options: `SAMPLES` and `STREAMED_XOR_CHUNKS`

Initial benchmarks are available here: https://github.com/prometheus/prometheus/pull/5703

More details on some decisions: https://docs.google.com/document/d/1JqrU3NjM9HoGLSTPYOvR217f5HBKBiJTqikEB9UiJL0/edit# 


bwplotka:
Pinged Cortex team for feedback: https://github.com/cortexproject/cortex/issues/1472

prompb/remote.proto
```diff
@@ -43,3 +65,12 @@ message QueryResult {
...
   repeated prometheus.TimeSeries timeseries = 1;
 }
+
+// ChunkedReadResponse is a response when response_type equals STREAMED_XOR_CHUNKS.
+// We strictly stream full series by series. Single frame can contain partion of the single series.
+message ChunkedReadResponse {
+  repeated prometheus.ChunkedSeries chunked_series = 1;
+
+  // This is used when we have more than one ReadRequest.queries.
+  int64 query_index = 2;
```
brian-brazil: Shouldn't it always be set, even when there's only one query?

prompb/remote.proto
```diff
@@ -23,10 +23,32 @@ message WriteRequest {
   repeated prometheus.TimeSeries timeseries = 1 [(gogoproto.nullable) = false];
 }
 
+// ReadRequest represents remote read request.
```
juliusv: represents -> represents a

prompb/remote.proto
```diff
@@ -23,10 +23,32 @@ message WriteRequest {
...
 message ReadRequest {
   repeated Query queries = 1;
+
+  enum ResponseType {
+    // Server will return ReadResponse with raw samples.
```
juliusv: return -> return a

prompb/remote.proto
```diff
@@ -23,10 +23,32 @@ message WriteRequest {
...
+    //
+    // Response headers:
+    // Content-Type: "application/x-protobuf"
+    // Content-Encoding: "snappy"
+    SAMPLES = 0;
+
+    // Server will stream varint delimited ChunkedReadResponse message that contains XOR encoded chunks for single series.
```
juliusv: stream -> stream a
for -> for a

prompb/remote.proto
```diff
@@ -23,10 +23,32 @@ message WriteRequest {
...
+    //
+    // Response headers:
+    // Content-Type: "application/x-streamed-protobuf; proto=prometheus.ChunkedReadResponse"
+    // Content-Encoding: ""
+    STREAMED_XOR_CHUNKS = 1;
+  }
+  // response_type allows negotiating content type of response.
+  // This does not guarantee the returned response. For servers that does not support this field, SAMPLES response type
```
juliusv: does -> do
SAMPLES -> the SAMPLES

prompb/remote.proto
```diff
@@ -23,10 +23,32 @@ message WriteRequest {
...
+    SAMPLES = 0;
+
+    // Server will stream varint delimited ChunkedReadResponse message that contains XOR encoded chunks for single series.
+    //
+    // Response headers:
+    // Content-Type: "application/x-streamed-protobuf; proto=prometheus.ChunkedReadResponse"
+    // Content-Encoding: ""
+    STREAMED_XOR_CHUNKS = 1;
+  }
+  // response_type allows negotiating content type of response.
```
juliusv: content type -> the content type

prompb/remote.proto
```diff
@@ -43,3 +65,12 @@ message QueryResult {
   // Samples within a time series must be ordered by time.
   repeated prometheus.TimeSeries timeseries = 1;
 }
+
+// ChunkedReadResponse is a response when response_type equals STREAMED_XOR_CHUNKS.
+// We strictly stream full series by series. Single frame can contain partion of the single series.
```
juliusv: Single -> A single
partion -> partition

juliusv: Not sure what "We strictly stream full series by series" means. I thought it means that each single series wouldn't be broken up, but then the next sentence says the opposite?

prompb/remote.proto
squat: `does` is correct here. `do` would be for plural, ie `these`

prompb/types.proto
```diff
@@ -56,3 +57,23 @@ message ReadHints {
   int64 start_ms = 3; // Start time in milliseconds.
   int64 end_ms = 4;   // End time in milliseconds.
 }
+
+// Chunk represents TSDB chunk.
```
squat: Should be “Chunk represents _a_ TSDB chunk.”

prompb/remote.proto
juliusv: Oh I meant the second "does" on this line, "For servers that does".

prompb/remote.proto
bwplotka: Wasted 64 bits in most cases, but otherwise yes - let's simplify this and set always.

prompb/remote.proto
bwplotka: `partition`: I meant rather `portion` here but partition might make sense as well (:

bwplotka: It rather means that we expect series to be be sequentially send. 

So you cannot send series ABC [1 10] then XYZ [1-20] then ABC [11-20] but you can ABC [1 10] then ABC [11-20]  then  XYZ [1-20]  (:

bwplotka: Tried to be more explicit. PTAL. Also not 100% sure if we want to be that strict on interface side.

bwplotka:
Comments fixed, PTAL.

prompb/remote.proto
brian-brazil: Proto is more efficient than that on the wire.

prompb/remote.proto
```diff
@@ -23,10 +23,32 @@ message WriteRequest {
...
+    //
+    // Response headers:
+    // Content-Type: "application/x-streamed-protobuf; proto=prometheus.ChunkedReadResponse"
+    // Content-Encoding: ""
+    STREAMED_XOR_CHUNKS = 1;
+  }
+  // response_type allows negotiating the content type of response.
+  // This does not guarantee the returned response. For servers that do not support this field, the SAMPLES response type
+  // is used.
+  ResponseType response_type = 2;
```
bwplotka: > https://github.com/prometheus/prometheus/pull/5703#discussion_r296321539

@brian-brazil commented: 

> Shouldn't this be a list, in order of preference?

Yes more of "accepted" response types might make more sense, happy to address that.  I would propose
`repeated ResponseType accepted_response_types = 2;`

bwplotka: I guess, if empty - we assume `[Samples]` though, right?

prompb/remote.proto
bwplotka: True, agree then.

prompb/remote.proto
brian-brazil: Yes, that'd be a sane default.

bwplotka: Cool, changing to that approach then (:

bwplotka:
@brian-brazil Moved to `accepted_response_types` (: Updating other PRs in chain.

prompb/remote.proto
```diff
@@ -23,10 +23,34 @@ message WriteRequest {
...
+    // Response headers:
+    // Content-Type: "application/x-streamed-protobuf; proto=prometheus.ChunkedReadResponse"
+    // Content-Encoding: ""
+    STREAMED_XOR_CHUNKS = 1;
+  }
+
+  // accepted_response_types allows negotiating the content type of the response.
+  // Response types are taken from the list in the FIFO order. If no response type is implemented by server, error is returned.
+  // For servers that do not support this field, the SAMPLES response type is used strictly.
+  // If field is empty, the server should assume SAMPLES response type is accepte by the caller.
```
brian-brazil: accepted

bwplotka: fixed (:

bwplotka:
@tomwilkie any feedback on this?

prompb/remote.proto
```diff
@@ -43,3 +67,14 @@ message QueryResult {
...
 }
+
+// ChunkedReadResponse is a response when response_type equals STREAMED_XOR_CHUNKS.
+// We strictly stream full series after series, optionally split by time. This means that a single frame can contain
+// partition of the single series, but once a new series is started to be streamed it means that no more chunks will
+// be sent for previous one.
+message ChunkedReadResponse {
+  repeated prometheus.ChunkedSeries chunked_series = 1;
+
+  // query_index representsa an index of the query from ReadRequest.queries this results relates to.
```
GiedriusS: Typo here as well: `s/representsa/represents/`

prompb/remote.proto
```diff
@@ -23,10 +23,34 @@ message WriteRequest {
...
+    // Content-Type: "application/x-streamed-protobuf; proto=prometheus.ChunkedReadResponse"
+    // Content-Encoding: ""
+    STREAMED_XOR_CHUNKS = 1;
+  }
+
+  // accepted_response_types allows negotiating the content type of the response.
+  // Response types are taken from the list in the FIFO order. If no response type is implemented by server, error is returned.
+  // For servers that do not support this field, the SAMPLES response type is used strictly.
+  // If field is empty, the server should assume SAMPLES response type is accepted by the caller.
+  repeated ResponseType accepted_response_types = 2;
```
tomwilkie: Given that empty implies SAMPLES, and we assume all clients must accept a SAMPLES response, do we really need the SAMPLES enum value?

prompb/types.proto
```diff
@@ -56,3 +57,23 @@ message ReadHints {
...
 }
+
+// Chunk represents a TSDB chunk.
+message Chunk {
+  int64 min_time_ms = 1;
+  int64 max_time_ms = 2;
+
+  // We require this to match chunkenc.Encoding.
+  enum Encoding {
+    UNKNOWN = 0;
```
tomwilkie: When is UNKOWN ever used?

prompb/remote.proto
```diff
@@ -43,3 +67,14 @@ message QueryResult {
   // Samples within a time series must be ordered by time.
   repeated prometheus.TimeSeries timeseries = 1;
 }
+
+// ChunkedReadResponse is a response when response_type equals STREAMED_XOR_CHUNKS.
+// We strictly stream full series after series, optionally split by time. This means that a single frame can contain
+// partition of the single series, but once a new series is started to be streamed it means that no more chunks will
+// be sent for previous one.
```
tomwilkie: Whilst this might be true, it seems like an unnecessary restriction that we might want to change in the future (esp for very long queries).  Should we consider not having this restriction as part of the interface?

Can we say anything about the order of the series we return?

prompb/types.proto
```diff
@@ -56,3 +57,23 @@ message ReadHints {
...
+    XOR     = 1;
+  }
+  Encoding type  = 3;
+  bytes data     = 4;
+}
+
+// ChunkedSeries represents single, encoded time series.
+message ChunkedSeries {
+  repeated Label labels = 1 [(gogoproto.nullable) = false];
+  repeated Chunk chunks = 2 [(gogoproto.nullable) = false];
```
tomwilkie: Is there anything to be said for the order of the chunks we return?  Will they overlap?

prompb/types.proto
```diff
@@ -56,3 +57,23 @@ message ReadHints {
...
+    UNKNOWN = 0;
+    XOR     = 1;
+  }
+  Encoding type  = 3;
+  bytes data     = 4;
+}
+
+// ChunkedSeries represents single, encoded time series.
+message ChunkedSeries {
+  repeated Label labels = 1 [(gogoproto.nullable) = false];
```
tomwilkie: We tried elsewhere in the remote interfaces to guarantee ordering of the labels IIRC.

prompb/remote.proto
brian-brazil: It's something that the other end can rely on for efficiency, and if there's a new TSDB implementation at some point where a different way makes sense we can always have a new type covering it. For now I'd say let users depend on this, rather than sacrificing this to an unknown future change.

prompb/types.proto
bwplotka: no, but please see comment, it helps to match `chunkenc.Encoding` exact values of enum (int).

prompb/remote.proto
bwplotka: Hm. I can agree with @tomwilkie that it will change soon and then we will need to do deprecation process or different type which might be painful, but without that restriction as @brian-brazil says callers and receivers cannot assume that useful invariant and gain benefits which would have bad conequences on Prometheus PromQL current shape.

This means that I think I am bit more on restricting it as @brian-brazil suggests, but maybe it makes sense then be prepared or at least have an idea how we will do it if we will need to change that restriction.

Maybe leave the current shape and once new restriction come in, add some `enum SeriesOrder` to `ReadRequest`? In theory we can do that gradually once the different restriction arise? And keep it YAGNI for now.

prompb/types.proto
bwplotka: Good point. I think the Prometheus server implementation will not give overlaps, but not sure if all known caller implementation assumes. I think both Prometheus and Thanos handles overlaps well.

In terms of sorting :man_shrugging: need to double check.

prompb/types.proto
bwplotka: yes. Also the in most other places in TSDB. Happy to add such comment and check implementation on later chained PRs.

However, what do you think about having more complex logic - if we send consequtive chunks for exactly the same series in multiple frames it is enough to send labels just on first one, first? But might be optimization we can add later (which will be another interface change ;p)

bwplotka:
Addressed  comments @tomwilkie @brian-brazil

prompb/remote.proto
brian-brazil: I'm still not seeing how we'd be changing this, and we can always extend the proto.

prompb/types.proto
brian-brazil: I'd say that they shouldn't overlay, and will be sorted.

prompb/types.proto
brian-brazil: That'd require a bit more state on the reader side for a pretty small saving, I'd keep it simple and always include it.

prompb/types.proto
bwplotka: commented.

bwplotka:
Closing -> the squashed commits are now part of this PR: https://github.com/prometheus/prometheus/pull/5703

PR #5702, opened 2019-06-21
remote-read: [2] Added chunked writer and reader.

bwplotka:
Second part of chained PRs to add streamed, chunked remote read as proposed here: https://docs.google.com/document/d/1JqrU3NjM9HoGLSTPYOvR217f5HBKBiJTqikEB9UiJL0/edit#

Part of https://github.com/prometheus/prometheus/issues/4517 and https://github.com/improbable-eng/thanos/issues/488

This is inspired by https://github.com/kythe/kythe/blob/master/kythe/go/platform/delimited/delimited.go but bit improved and tailored to our needs.

### DIFF
--- a/storage/remote/stream.go
+++ b/storage/remote/stream.go
@@ -19,8 +19,8 @@ import (
 	"io"
 )
 
-// StreamWriter is an io.Writer wrapper that allows streaming by adding uvarint delimiter before each write in form of length of
-// the corresponded byte array.
+// StreamWriter is an io.Writer wrapper that allows streaming by adding uvarint delimiter before each write in a form
+// of length of the corresponded byte array.
 type StreamWriter struct {
 	wrapped io.Writer
 }
@@ -31,7 +31,7 @@ func NewStreamWriter(w io.Writer) *StreamWriter {
 }
 
 // Write writes given bytes to the stream. It adds uvarint delimiter before each message.
-// Returned bytes number are for sent bytes for given buffer, it does not include delimiter bytes.
+// Returned bytes number represents sent bytes for a given buffer. The number does not include delimiter bytes.
 func (w *StreamWriter) Write(b []byte) (int, error) {
 	if len(b) == 0 {
 		return 0, nil
@@ -68,7 +68,7 @@ func NewStreamReader(r io.Reader) *StreamReader {
 // there are no more records available. Returns io.ErrUnexpectedEOF if a short
 // record is found, with a length of n but fewer than n bytes of data.
 //
-// NOTE: The slice returned is valid only until a subsequent call to Next. It's a caller responsibility to copy the
+// NOTE: The slice returned is valid only until a subsequent call to Next. It's a caller's responsibility to copy the
 // returned slice if needed.
 func (r *StreamReader) Next() ([]byte, error) {
 	size, err := binary.ReadUvarint(r.b)
@@ -89,7 +89,7 @@ func (r *StreamReader) Next() ([]byte, error) {
 }
 
 // NextProto consumes the next available record by calling r.Next, and decodes
-// it into pb with proto.Unmarshal.
+// it into the protobuf with proto.Unmarshal.
 func (r *StreamReader) NextProto(pb proto.Message) error {
 	rec, err := r.Next()
 	if err != nil {
--- a/storage/remote/stream.go
+++ b/storage/remote/stream.go
@@ -1,0 +1,99 @@
+// Copyright 2019 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package remote
+
+import (
+	"bufio"
+	"encoding/binary"
+	"github.com/gogo/protobuf/proto"
+	"io"
+)
+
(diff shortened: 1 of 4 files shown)


storage/remote/stream.go
```diff
@@ -0,0 +1,99 @@
...
+// NewStreamReader constructs a StreamReader.
+func NewStreamReader(r io.Reader) *StreamReader {
+	return &StreamReader{b: bufio.NewReader(r)}
+}
+
+// Next returns the next length-delimited record from the input, or io.EOF if
+// there are no more records available. Returns io.ErrUnexpectedEOF if a short
+// record is found, with a length of n but fewer than n bytes of data.
+//
+// NOTE: The slice returned is valid only until a subsequent call to Next. It's a caller responsibility to copy the
```
brian-brazil: the caller's

bwplotka:
Fixed comments PTAL (:

storage/remote/stream.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	var buf [binary.MaxVarintLen64]byte
+	v := binary.PutUvarint(buf[:], uint64(len(b)))
+
+	if _, err := w.wrapped.Write(buf[:v]); err != nil {
+		return 0, err
+	}
+
+	n, err := w.wrapped.Write(b)
```
GiedriusS: We could simplify this to a simple: `return w.wrapped.Write(b)`.

storage/remote/stream.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		return n, err
+	}
+
+	return n, nil
```
tomwilkie: Its hard to know without seeing the entire implementation, but I think you'll need a flush to ensure the data isn't buffered in the server.

tomwilkie:
The HTTP server is probably going to be using HTTP chunked encoding under the hood here (hard to know for sure without the entire implementation), so is a custom framing format necessary?

bwplotka:
> The HTTP server is probably going to be using HTTP chunked encoding under the hood here (hard to know for sure without the entire implementation), so is a custom framing format necessary?

Very good point, missed that, will explore this.

bwplotka:
Closing until the alternative is checked (:

bwplotka:
https://tools.ietf.org/html/rfc2616#section-3.6.1 :tada:

bwplotka:
ok reopening, as after debugging a bit the HTTP can be chunked or whatever, but we have no control on how it chunks and IF it chunks so I think we need application layer chunk boundaries / delimiters like in this PR.

See: https://groups.google.com/d/msg/golang-nuts/y8CsrzluZmA/ioG5p07AwlcJ

bwplotka:
That's why https://golang.org/pkg/encoding/json/#Decoder streaming decoder exists and there is nothing standard for protobuf

bwplotka:
Also looked on gRPC internals and sure - they operate on TCP connection directly: https://github.com/grpc/grpc-go/blob/master/rpc_util.go#L592 https://github.com/grpc/grpc-go/blob/master/internal/transport/http2_server.go#L898 and put data frames as designed in HTTP/2.

If we choose to base on Chunked format, for HTTP golang client there are thousands edge cases on how it is chunked and for our protobuf parsing we need to know *exactly* number of bytes in each frame as it's the only info we have to tell what bytes to parse. And this info is actually deep in http `transferWire` struct: https://golang.org/src/net/http/transfer.go#L535

I am bit blocked on even simple MVP as not sure how to proceed as client response Body has to know exactly how much bytes to buffer in given buffer (as io.Reader) and we don't know that as we don't have access to underlying delimiter as specified in HTTP/1 rfc... unless we will operate directly on TCP conn instead HTTP Client which is overengineering IMO. ;p

brian-brazil:
Sounds like we should do our own framing then.

bwplotka:
Yea. Essentially the implementation for caller would be quite crazy for Golang, as Golang `http.Transport` does guarantee the `Read` will return chunk by chunk (although usually that's the case).   Short tests confirms it. With flush I can see the internal `ChunkedReader` is reading chunk by chunk (numbers are lenght of byte writes), but `response.Body.Read` is not giving me chunk by chunk strictly:
```
write  86
flush
write  87
flush
write  87
flush
(inside chunkedReader) READING  86
response.Body.Read 86
chunked_series:<labels:<name:"__name__" value:"test_metric1" > labels:<name:"b" value:"c" > labels:<name:"baz" value:"qux" > labels:<name:"d" value:"e" > labels:<name:"foo" value:"bar" > chunks:<type:XOR data:"\000\001\000?\360\000\000\000\000\000\000\000" > > 
(inside chunkedReader)  READING ..  87
(inside chunkedReader)  READING ..  87
response.Body.Read 174 & err EOF
``` 
Last one violates our protocol as from 174 bytes I have no idea as a caller if it's one protobuf or many.

Also worth to not that for this test I couldn't use `httptest.Recorder` in tests as it does not have any notion of internal chunking ;p 

Full attempt code is here: https://github.com/prometheus/prometheus/compare/master...bwplotka:rr-attempt-chunked-framing?expand=1#diff-c32caf482fb76fa695ceffe3455c6d9aR1059  I would not spend more time on it though.

Alternatively we could dial natively and setup TCP connection and do things directly on connection using https://golang.google.cn/src/net/http/internal/chunked.go but it is overkill and IMO it might be similarly hard for users of our remote read in different languages. 

I would stick to application delimition. Obviously we will rely on HTTP chunking for actual streaming, but the actual chunking is very close to how we "Flush()"  response on server side (but no strictly the same), so I think it's good enough. Varint should be simple for callers to expect as well, and the overhead should be small if we keep frames not too small.

storage/remote/stream.go
bwplotka: Yes. I planned to do it on caller side, but happy to ensure this here. Let's see what is more clear.

bwplotka:
PTAL @tomwilkie @brian-brazil @SuperQ addressed comments. Moved to `ChunkedWriter/Reader` and added flushing.

bwplotka:
Closing -> the squashed commits are now part of this PR: https://github.com/prometheus/prometheus/pull/5703

PR #5703, opened 2019-06-21
remote-read: streamed chunked server side; Extended protobuf; Added chunked, checksumed reader 

bwplotka:
Part of: https://github.com/prometheus/prometheus/issues/4517 and https://github.com/improbable-eng/thanos/issues/488

Changes:
* Extended protobuf
* Chunked Writer/Reader.
* Server side implementation.

Intentionally there is:
* No optimization for older blocks against reencoding.
* Remote read client code

Those will be added in separate PR.

### Before:
* Avg request time: 25s (10s just Prometheus)
* Allocations vs Memory: 

image

### After:
* This really depends on configuration, but for now:
  * 120 samples in XOR chunk
  * Max 100 chunks per frame (probably too low).
* Avg request time 15s
* Frames:  160000
* Allocations vs Memory:

image

CPU us 3x but we got here quite low chunk number, I am pretty sure we should try with bigger number.

More details on some decisions: https://docs.google.com/document/d/1JqrU3NjM9HoGLSTPYOvR217f5HBKBiJTqikEB9UiJL0/edit# 


storage/remote/codec.go
```diff
@@ -183,6 +183,148 @@ func FromQueryResult(res *prompb.QueryResult) storage.SeriesSet {
...
+		err     error
+
+		isNext     = true
+		numSamples = 0
+		chks       = make([]prompb.Chunk, 0, maxChunks)
+	)
+
+	for isNext {
+		numSamples++
+		if sampleLimit > 0 && numSamples > sampleLimit {
```
brian-brazil: Is it still possible to hit this?

storage/remote/codec.go
```diff
@@ -183,6 +183,148 @@ func FromQueryResult(res *prompb.QueryResult) storage.SeriesSet {
...
+		chk     *chunkenc.XORChunk
+		app     chunkenc.Appender
+		err     error
+
+		isNext     = true
+		numSamples = 0
+		chks       = make([]prompb.Chunk, 0, maxChunks)
+	)
+
+	for isNext {
```
brian-brazil: I think you can do a more standard `for it := it.Next() {` here

prompb/remote.proto
```diff
@@ -23,10 +23,32 @@ message WriteRequest {
...
+    //
+    // Response headers:
+    // Content-Type: "application/x-streamed-protobuf; proto=prometheus.ChunkedReadResponse"
+    // Content-Encoding: ""
+    STREAMED_XOR_CHUNKS = 1;
+  }
+  // response_type allows negotiating content type of response.
+  // This does not guarantee the returned response. For servers that does not support this field, SAMPLES response type
+  // is used.
+  ResponseType response_type = 2;
```
brian-brazil: Shouldn't this be a list, in order of preference?

web/api/v1/api.go
```diff
@@ -853,11 +854,18 @@ func (api *API) remoteRead(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	if req.ResponseType != prompb.ReadRequest_SAMPLES {
+	switch req.ResponseType {
+	case prompb.ReadRequest_SAMPLES: // Also default.
+		api.sampledRemoteRead(ctx, w, req)
+	case prompb.ReadRequest_STREAMED_XOR_CHUNKS:
+		api.streamedChunkedRemoteRead(ctx, w, req)
+	default:
 		http.Error(w, fmt.Sprintf("%s response type is not implemented", req.ResponseType.String()), http.StatusNotImplemented)
```
brian-brazil: This should fallback to SAMPLES

storage/remote/codec.go
bwplotka: Yes, why not? Sample limit is counted overall for all frames, so `sampleLimit` here contains the left samples we can serve. 

With streaming it does not makes sense much maybe as we made remote read quite efficient with this, but still caller would like to limit number of overall samples maybe in all frames together? What do you think?

bwplotka: The below error is wrong though as `sampleLimit` is not overall limit, just the left samples for the chunks here. Will fix the error message.

storage/remote/codec.go
bwplotka: In current version I cannot due to:

> // encodeChunks expects iterator to be ready to use (aka iter.Next() done before invoking).

This simplifies caller code for us, but happy to rearrange it a bit.

prompb/remote.proto
bwplotka: Very good question but will move this comment to https://github.com/prometheus/prometheus/pull/5701

web/api/v1/api.go
bwplotka: hm.. Not sure about it, but if we move to `accepted_response_types` field instead, the logic will be clearer here, right? See https://github.com/prometheus/prometheus/pull/5701/files#r296731573

storage/remote/codec.go
brian-brazil: I'm not seeing that

cmd/prometheus/main.go
```diff
@@ -212,6 +212,9 @@ func main() {
 	a.Flag("storage.remote.read-concurrent-limit", "Maximum number of concurrent remote read calls. 0 means no limit.").
 		Default("10").IntVar(&cfg.web.RemoteReadConcurrencyLimit)
 
+	a.Flag("storage.remote.read-max-chunks-in-frame", "Maximum number of chunks in single frame for STREAMED_XOR_CHUNKS remote read response type. Each chunk corresponds roughly to (~3B * 120 samples) + 32B. Default is 1000 which is roughly (1000 * ~400B) + labelset, so approx. 0.4MB per frame .").
+		Default("1000").IntVar(&cfg.web.RemoteReadMaxChunksInFrame)
+
```
tomwilkie: What happens when a single series has more than chunks than this?

storage/remote/codec.go
```diff
@@ -183,6 +183,150 @@ func FromQueryResult(res *prompb.QueryResult) storage.SeriesSet {
...
+	}
+	if err := ss.Err(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// encodeChunks expects iterator to be ready to use (aka iter.Next() done before invoking).
+func encodeChunks(iter storage.SeriesIterator, maxSamplesPerChunk int, maxChunks int, allowedNumSamples int) ([]prompb.Chunk, int, error) {
```
tomwilkie: If I read this correctly, this is iterating over the samples in the series and rebuilding the chunks.  Can we extend TSDB to just give us the chunks and save the hassle?

storage/remote/codec.go
```diff
@@ -183,6 +183,150 @@ func FromQueryResult(res *prompb.QueryResult) storage.SeriesSet {
...
+
+	for isNext {
+		numSamples++
+		if allowedNumSamples > 0 && numSamples > allowedNumSamples {
+			return nil, numSamples, HTTPError{
+				msg:    fmt.Sprintf("exceeded sample limit"),
+				status: http.StatusBadRequest,
+			}
+		}
+
```
tomwilkie: Do we need a sample limit in the streaming case?

cmd/prometheus/main.go
tomwilkie: I actually don't see where this is enforced...

storage/remote/codec.go
tomwilkie: ps I want this for Cortex too, so I did it: https://github.com/cortexproject/cortex/pull/1427/files#diff-a6d94f56da1ed83021ef693d2dd12b55R58

storage/remote/stream.go
```diff
@@ -0,0 +1,99 @@
...
+		return nil, err
+	}
+
+	if cap(r.data) < int(size) {
+		r.data = make([]byte, size)
+	} else {
+		r.data = r.data[:size]
+	}
+
+	if _, err := io.ReadFull(r.b, r.data); err != nil {
```
tomwilkie: I feel like we probably want to limit the size of this read (or the allocation above).

web/api/v1/api.go
```diff
@@ -927,6 +963,85 @@ func (api *API) remoteRead(w http.ResponseWriter, r *http.Request) {
...
+				matcher, err := labels.NewMatcher(labels.MatchEqual, m.Name, "")
+				if err != nil {
+					http.Error(w, err.Error(), http.StatusInternalServerError)
+					return
+				}
+				filteredMatchers = append(filteredMatchers, matcher)
+			} else {
+				filteredMatchers = append(filteredMatchers, m)
+			}
+		}
```
tomwilkie: Can we factor this code out and share it with the origin remote read implementation?

tomwilkie:
Looking good @bwplotka!  I've left any major points but not really gone through it with a fine tooth comb.

Personal preference for me would be to merge the first PRs into this one and not go this as a series of small PRs.  But thats up to you...

storage/remote/codec.go
brian-brazil: I gather this PR is just the first step.

web/api/v1/api.go
bwplotka: Not sure, maybe for compatiblity falling back to sampled might make sense here

storage/remote/codec.go
bwplotka: Hard to show you the issue, but this was the most clear implementation to me. Alternatives:

a)   passing `stream io.Write` into encodeChunks directly and stream straight away once chunks built up.

b)  duplicating `cut` process so it will end up like this:
```
// encodeChunks expects iterator to be ready to use (aka iter.Next() done before invoking).
func encodeChunks(iter storage.SeriesIterator, maxSamplesPerChunk int, maxChunks int, allowedNumSamples int) ([]prompb.Chunk, int, error) {
	var (
		chkMint int64
		chkMaxt int64
		chk     *chunkenc.XORChunk
		app     chunkenc.Appender
		err     error

		numSamples = 0
		chks       = make([]prompb.Chunk, 0, maxChunks)
	)

	for iter.Next() {
		numSamples++
		if allowedNumSamples > 0 && numSamples > allowedNumSamples {
			return nil, numSamples, HTTPError{
				msg:    fmt.Sprintf("exceeded sample limit"),
				status: http.StatusBadRequest,
			}
		}

		if chk == nil {
			chk = chunkenc.NewXORChunk()
			app, err = chk.Appender()
			if err != nil {
				return nil, numSamples, err
			}
			chkMint, _ = iter.At()
		}

		app.Append(iter.At())
		chkMaxt, _ = iter.At()

		if chk.NumSamples() < maxSamplesPerChunk {
			continue
		}

		// Cut the chunk.
		chks = append(chks, prompb.Chunk{
			MinTimeMs: chkMint,
			MaxTimeMs: chkMaxt,
			Type:      prompb.Chunk_Encoding(chk.Encoding()),
			Data:      chk.Bytes(),
		})
		chk = nil

		if maxChunks >= len(chks) {
			break
		}
	}
	if iter.Err() != nil {
		return nil, numSamples, errors.Wrap(iter.Err(), "iter TSDB series")
	}

	if chk != nil {
		// Cut the chunk if exists.
		chks = append(chks, prompb.Chunk{
			MinTimeMs: chkMint,
			MaxTimeMs: chkMaxt,
			Type:      prompb.Chunk_Encoding(chk.Encoding()),
			Data:      chk.Bytes(),
		})
	}
	return chks, numSamples, nil
}
```

bwplotka: will try b let's see if we like it more

cmd/prometheus/main.go
bwplotka: wait. I think there is misunderstanding. This is not a limit. It's essentially control on how much you want to buffer / how large frames are.

You can have many frames per series. How we can make it more clear here?

storage/remote/codec.go
bwplotka: Yes. I would really really vote for avoiding scope creep and merge working bits by bits.

We can add avoid re-encoding easily in other PRs, it's just internal optimization. I promise I will do it, but let's ship massive win with even this implementation. ;p

storage/remote/codec.go
bwplotka: Good question, what are your thoughts? (: If you accidently ask for 10 years data you will just.. timeout I guess? so timeout might be totally enough.

bwplotka: Happy to remove sample limitation

storage/remote/stream.go
bwplotka: Comment related to: https://github.com/prometheus/prometheus/pull/5702

web/api/v1/api.go
bwplotka: I refactored a bit. I could refactor more (literally only response is different), but since it might change soon (samples vs fetching encoded chunks) I would not try to simplify it more. Let me know your thoughts.

storage/remote/codec.go
bwplotka: Removed, and added info in sample limit flag.

prompb/remote.proto
```diff
@@ -23,10 +23,32 @@ message WriteRequest {
...
+  }
+
+  // accepted_response_types allows negotiating the content type of the response.
+  //
+  // Response types are taken from the list in the FIFO order. If no response type in `accepted_response_types` is
+  // implemented by server, error is returned.
+  // For request that do not support `accepted_response_types` field the non streamed, raw samples response is used.
+  // In such case response headers are as follows:
+  //
+  // Content-Type: "application/x-protobuf"
```
brian-brazil: This should be the first item in the enum.

prompb/types.proto
```diff
@@ -56,3 +57,25 @@ message ReadHints {
   int64 start_ms = 3; // Start time in milliseconds.
   int64 end_ms = 4;   // End time in milliseconds.
 }
+
+// Chunk represents a TSDB chunk.
```
brian-brazil: open/closed should be documented here

prompb/remote.proto
bwplotka: See: https://github.com/prometheus/prometheus/pull/5701#discussion_r297801885

bwplotka:
I might think that chained PRs in this form might be more confusion then benefit ;p

bwplotka:
OK, moved everything to this PR for now.

Updated description with:
```
3 commits:
* Extended protobuf
* Chunked Writer/Reader.
* Server side implementation.

Intentionally there is:
* No optimization for older blocks against reencoding.
* Remote read client code

Those will be added in separate PR.
```

prompb/remote.proto
bwplotka: Essentially the @tomwilkie idea was that SAMPLED is by default when the `accepted...` is empty, so no point in having separate enum item - and I can agree with that. It avoids having multiple options to control one thing.

What are you thoughts @brian-brazil why it should be in enum?

storage/remote/stream.go
bwplotka: Will think about it, thanks!

prompb/remote.proto
brian-brazil: What if I explicitly want to ask for it as a fallback?

bwplotka:
Need to update https://github.com/opentracing-contrib/go-stdlib dependency. In older version it does not support Flusher.

storage/remote/codec.go
```diff
@@ -183,6 +164,154 @@ func FromQueryResult(res *prompb.QueryResult) storage.SeriesSet {
...
+		chks = make([]prompb.Chunk, 0, maxChunksInFrame)
+		err  error
+	)
+
+	for ss.Next() {
+		series := ss.At()
+		iter := series.Iterator()
+		lbls := MergeLabels(labelsToLabelsProto(series.Labels()), sortedExternalLabels)
+
+		// TODO(bwplotka): We send each series in separate frame no matter what. Even if series has only one sample.
```
bwplotka: @brian-brazil @tomwilkie I think this is quite major optimization we can make, but adds extra complexity to the caller. I think it's worth it though.

bwplotka:
Current state with 1000 max chunks:
image

I think we could put *more* chunks into frame (2000)? and add optimization mentioned here: https://github.com/prometheus/prometheus/pull/5703#discussion_r300649511

bwplotka:
Prometheus mem spike is due to reencoding which is planned to be improved in later PRs.

bwplotka:
Build `bplotka/prometheus:v2.11.0-rc.0-rr-streaming` with remote read streaming server support (#5703) if someone want to test it.

storage/remote/codec.go
```diff
@@ -183,6 +164,154 @@ func FromQueryResult(res *prompb.QueryResult) storage.SeriesSet {
...
+	if err := ss.Err(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// encodeChunks expects iterator to be ready to use (aka iter.Next() done before invoking).
+func encodeChunks(iter storage.SeriesIterator, chks []prompb.Chunk, maxChunks int) ([]prompb.Chunk, error) {
+	const maxSamplesInChunk = 120
```
bwplotka: worth to comment that this hardcoded magic value is for XOR only

bwplotka:
Not much changes indeed for `Series interface`  to return Chunks but quite major interface change: https://github.com/cortexproject/cortex/pull/1427/files#r301158203 Can experiment with this a bit a a separate side thing. Still looking for review on this PR if you anyone has some time.

cmd/prometheus/main.go
```diff
@@ -213,12 +213,15 @@ func main() {
 	a.Flag("storage.remote.flush-deadline", "How long to wait flushing sample on shutdown or config reload.").
 		Default("1m").PlaceHolder("<duration>").SetValue(&cfg.RemoteFlushDeadline)
 
-	a.Flag("storage.remote.read-sample-limit", "Maximum overall number of samples to return via the remote read interface, in a single query. 0 means no limit.").
+	a.Flag("storage.remote.read-sample-limit", "Maximum overall number of samples to return via the remote read interface, in a single query. 0 means no limit. This limit is ignored for STREAMED_XOR_CHUNKS response type.").
```
brian-brazil: I'd say "streamed" to be future proof.

cmd/prometheus/main.go
brian-brazil: How about specifying this in bytes, so it works across future types?

storage/remote/chunked.go
```diff
@@ -0,0 +1,103 @@
...
+	writer  io.Writer
+	flusher http.Flusher
+}
+
+// NewChunkedWriter constructs a ChunkedWriter.
+func NewChunkedWriter(w io.Writer, f http.Flusher) *ChunkedWriter {
+	return &ChunkedWriter{writer: w, flusher: f}
+}
+
+// Write writes given bytes to the stream. It adds uvarint delimiter before each message.
```
brian-brazil: Should we add a checksum while we're at it?

storage/remote/codec.go
brian-brazil: I'm not sure it'll make much difference, series ordering should be preserved though.

prompb/remote.proto
tomwilkie: You can't.  It is the always the explicit fallback.  

If you want to ask for it explicitly, don't send any `accepted_response_types` - which is the same as getting a request for a client that doesn't know about this new field.

prompb/types.proto
```diff
@@ -56,3 +57,25 @@ message ReadHints {
...
+  }
+  Encoding type  = 3;
+  bytes data     = 4;
+}
+
+// ChunkedSeries represents single, encoded time series.
+message ChunkedSeries {
+  // labels should be sorted.
+  repeated Label labels = 1 [(gogoproto.nullable) = false];
+  // chunks should be sorted should not overlap in time.
```
tomwilkie: Does this work with backfill? 

Also, Sentence case on comments please.

prompb/remote.proto
brian-brazil: That's not great in negotiation terms, maybe it's too inefficient for what I'm trying to do. I'd rather keep this explicit.

prompb/types.proto
brian-brazil: That's a case where we'd have to reencode I suspect.

storage/remote/chunked.go
tomwilkie: I don't have strong opinions here; it adds complexity, OTOH TCP checksums are pretty useless.  I'm inclined to say not right now, mainly as we have the ability to introduce a new framing protocol with checksums in the future.

storage/remote/chunked.go
```diff
@@ -0,0 +1,103 @@
...
+	size, err := binary.ReadUvarint(r.b)
+	if err != nil {
+		return nil, err
+	}
+
+	if cap(r.data) < int(size) {
+		r.data = make([]byte, size)
+	} else {
+		r.data = r.data[:size]
+	}
```
tomwilkie: We really need to have a limit to the reasonable lengths here; reading a corrupt length (perhaps due to a malicious server) could really hurt here.

web/api/v1/api.go
```diff
@@ -837,8 +840,23 @@ func (api *API) serveFlags(r *http.Request) apiFuncResult {
 	return apiFuncResult{api.flagsMap, nil, nil, nil}
 }
 
+// negotiateResponseType returns first accepted response type that this server supports.
+func negotiateResponseType(accepted []prompb.ReadRequest_ResponseType) prompb.ReadRequest_ResponseType {
+	supported := map[prompb.ReadRequest_ResponseType]struct{}{
+		prompb.ReadRequest_STREAMED_XOR_CHUNKS: {},
+	}
```
tomwilkie: Could probably be moved to the package level.

web/api/v1/api.go
```diff
@@ -853,39 +871,71 @@ func (api *API) remoteRead(w http.ResponseWriter, r *http.Request) {
...
-					return
-				}
-				filteredMatchers = append(filteredMatchers, matcher)
-			} else {
-				filteredMatchers = append(filteredMatchers, m)
+		defer func() {
+			if err := querier.Close(); err != nil {
+				level.Warn(api.logger).Log("msg", "error on querier close", "err", err.Error())
+			}
+		}()
```
tomwilkie: This defer won't be executed until the whole function exits IIUC, meaning we can have multiple queries open per remote_read.  Might want to break the body of the loop out into its own function.

go.mod
```diff
@@ -60,7 +60,7 @@ require (
 	github.com/mwitkow/go-conntrack v0.0.0-20161129095857-cc309e4a2223
 	github.com/oklog/run v1.0.0
 	github.com/olekukonko/tablewriter v0.0.1 // indirect
-	github.com/opentracing-contrib/go-stdlib v0.0.0-20170113013457-1de4cc2120e7
+	github.com/opentracing-contrib/go-stdlib v0.0.0-20190519235532-cf7a6c988dc9
```
tomwilkie: Is this change spurious?

web/api/v1/api.go
tomwilkie: We've had bugs in the selectParams not being honored in the past; super keen for this in particular not to be repeated.

web/api/v1/api.go
```diff
@@ -927,6 +977,82 @@ func (api *API) remoteRead(w http.ResponseWriter, r *http.Request) {
...
+		sortedExternalLabels := make([]prompb.Label, 0, len(externalLabels))
+		for name, value := range externalLabels {
+			sortedExternalLabels = append(sortedExternalLabels, prompb.Label{
+				Name:  string(name),
+				Value: string(value),
+			})
+		}
+		sort.Slice(sortedExternalLabels, func(i, j int) bool {
+			return sortedExternalLabels[i].Name < sortedExternalLabels[j].Name
+		})
```
tomwilkie: Cloud this be done once, and not for each request?

web/api/v1/api.go
```diff
@@ -927,6 +977,82 @@ func (api *API) remoteRead(w http.ResponseWriter, r *http.Request) {
...
+
+		w.Header().Set("Content-Type", "application/x-streamed-protobuf; proto=prometheus.ChunkedReadResponse")
+		// TODO(bwplotka): Should we use snappy? benchmark to see.
+		// w.Header().Set("Content-Encoding", "snappy")
+
+		f, ok := w.(http.Flusher)
+		if !ok {
+			http.Error(w, "internal http.ResponseWriter does not implement http.Flusher interface", http.StatusInternalServerError)
+			return
+		}
```
tomwilkie: When is this the case?

prompb/remote.proto
tomwilkie: Why is that not great?  Seems (a) pretty straight forward and (b) the correct embedding of what we want in what protos provide.

prompb/types.proto
tomwilkie: Why?  

We could relax this constraint to something like "chunks will be in start time order and may overlap." and leave merging to the client.

storage/remote/codec.go
tomwilkie: Yeah this is fine to start, @bwplotka.  Could rephrase for clarity "A most one series per frame; series may be split over multiple frames according to maxChunksInFrame"

cmd/prometheus/main.go
tomwilkie: @bwplotka yup, I misunderstood.  I see how this works now.  Thanks.

storage/remote/codec.go
tomwilkie: Happy to do this optimisation as another PR.  Even happy to do that myself.

go.mod
tomwilkie: > Need to update https://github.com/opentracing-contrib/go-stdlib dependency. In older version it does not support Flusher.

tomwilkie:
@bwplotka looking good.  Give me a ping when you've done another round of review comments, I'm hoping we can get this in for the next release.

prompb/remote.proto
brian-brazil: We'd send a whole pile of data that the other end doesn't want. I can see this as a gotcha in the future, why is the server returning something in a format I didn't ask for?

cmd/prometheus/main.go
bwplotka: Should we do it by bytes then? `storage.remote.read-max-bytes-in-frame` 

I think we might be able to calculate on the stream side, but it might be bit brittle and complex to implement.

storage/remote/chunked.go
bwplotka: :+1:

go.mod
bwplotka: (:

storage/remote/codec.go
bwplotka: to link: https://github.com/prometheus/tsdb/pull/659

But I think iterator would be awesome instead.

prompb/remote.proto
bwplotka: Yea I can see the failure in case of unsupported accepted formats only as the more explicit logic. If the client wants to be backward compatible it might need to explicitly state some `SAMPLES` enum at the end of `accepted_response_type` and this is fair. 

On the other hand @tomwilkie is right that there will be now two ways to ask for SAMPLES flow: either empty `accepted_response_type` or explicit SAMPLES in `accepted_response_type`. I feel like this is less evil though than the bit of the confusion as well as lack of way to disable SAMPLES response totally from client side. And I think it can be fair use case.

I am happy to revert to previous flow with `SAMPLES` and `STREAMED_XOR_CHUNKS` enum once we have consenus on this. Any other thoughts? @gouthamve ?

cmd/prometheus/main.go
bwplotka: good point!

prompb/types.proto
bwplotka: I guess you mean something like `// Time range [min, max] is inclusive.`?

storage/remote/codec.go
bwplotka: actually we explicitly do XOR chunks here, so it's not needed.

web/api/v1/api.go
bwplotka: done

storage/remote/chunked.go
bwplotka: done.

web/api/v1/api.go
bwplotka: :+1:

bwplotka: done

web/api/v1/api.go
bwplotka: When someone refactors other part of the code (wrap it with something that does not implement this).

We have lots of middlewares so it is super likely (: With this we can catch that in unit tests.

bwplotka:
Addressed all @tomwilkie @brian-brazil (: Thanks for detailed reviewed so far! Ready for another one.

Still we need to agree on:
* Do we need checksum. I think we don't. (https://github.com/prometheus/prometheus/pull/5703#discussion_r302631617) 
* Should we have `SAMPLES` in ReadResponse type. I think i agree with @brian-brazil that we should. ( https://github.com/prometheus/prometheus/pull/5703#discussion_r304169343)

storage/remote/chunked.go
bwplotka: isn't that done on [TCP level](https://locklessinc.com/articles/tcp_checksum/)?

brian-brazil: As Tom says, TCP checksums are pretty useless.

prompb/types.proto
brian-brazil: That's extra complexity for every client to handle, and they'll likely only do so when the bug reports come in.

cmd/prometheus/main.go
brian-brazil: We'd only need something that's good enough. The idea is more to stop massive frames, than to get it within the byte.

brian-brazil:
I'm not exactly awake right now, but it looks like just those to decide on.

bwplotka:
@tomwilkie do you have strong opinion about this? (:

>  Should we have SAMPLES in ReadResponse type. I think i agree with @brian-brazil that we should. (https://github.com/prometheus/prometheus/pull/5703#discussion_r304169343)

For the checksum I happy to add it straightaway.

bwplotka:
@tomwilkie @codesome @brian-brazil addressed all comments. 

* Added CRC32 checksum, plus tests.
* Reverted to explicit SAMPLES response type.
* Bit reluctant to change frame limit to be in bytes - I think having this in number of chunks is fair.

prompb/remote.proto
```diff
@@ -23,10 +23,35 @@ message WriteRequest {
...
   repeated Query queries = 1;
+
+  enum ResponseType {
+    // Server will return a single ReadResponse message with matched series that includes list of raw samples.
+    // It's recommended to use streamed response types instead.
+    //
+    // Response headers:
+    // Content-Type: "application/x-protobuf"
+    // Content-Encoding: "snappy"
+    SAMPLED = 0;
```
brian-brazil: SAMPLES

storage/remote/chunked.go
```diff
@@ -0,0 +1,154 @@
...
+// NewChunkedWriter constructs a ChunkedWriter.
+func NewChunkedWriter(w io.Writer, f http.Flusher) *ChunkedWriter {
+	return &ChunkedWriter{writer: w, flusher: f, crc32: crc32.New(castagnoliTable)}
+}
+
+// Write writes given bytes to the stream and flushes it.
+// Each frame includes:
+//
+// 1. uvarint for the size of the data frame.
+// 2. uvarint for the Castagnoli polynomial CRC-32 checksum of the data frame.
```
brian-brazil: Why not fixed size?

cmd/prometheus/main.go
brian-brazil: I still think we should either do this in bytes, or just leave it hardcoded.

storage/remote/codec.go
```diff
@@ -183,6 +161,173 @@ func FromQueryResult(res *prompb.QueryResult) storage.SeriesSet {
...
+		}
+	}
+	if err := ss.Err(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// encodeChunks expects iterator to be ready to use (aka iter.Next() done before invoking).
```
brian-brazil: s/done/called/

web/api/v1/api.go
```diff
@@ -856,78 +860,153 @@ func (api *API) remoteRead(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	resp := prompb.ReadResponse{
-		Results: make([]*prompb.QueryResult, len(req.Queries)),
+	externalLabels := api.config().GlobalConfig.ExternalLabels.Map()
+
+	// Add external labels back in, in sorted order.
```
brian-brazil: This comment isn't right

storage/remote/chunked.go
bwplotka: You are right, it should be fixed - CRC will be always fully filled

bwplotka:
@brian-brazil in terms of limit of bytes per frame. How accurate you would be?

Sum:
* how many bytes labelset has
* add chunks one by one until bytes limit?

That number would probably will be near to the actual memory used or actual protobuf size neither ):

brian-brazil:
That's roughly what I'd do.

bwplotka:
Addressed comments and moved to bytes limt @brian-brazil

bwplotka:
Ping @brian-brazil @tomwilkie @codesome @krasi-georgiev if you have time. (:

cmd/prometheus/main.go
```diff
@@ -219,8 +219,8 @@ func main() {
 	a.Flag("storage.remote.read-concurrent-limit", "Maximum number of concurrent remote read calls. 0 means no limit.").
 		Default("10").IntVar(&cfg.web.RemoteReadConcurrencyLimit)
 
-	a.Flag("storage.remote.read-max-chunks-in-frame", "Maximum number of chunks in single frame for STREAMED_XOR_CHUNKS remote read response type. Each chunk corresponds roughly to (~3B * 120 samples) + 32B. Default is 1000 which is roughly (1000 * ~400B) + labelset, so approx. 0.4MB per frame. Be aware that client might have limit on frame size as well.").
-		Default("1000").IntVar(&cfg.web.RemoteReadMaxChunksInFrame)
+	a.Flag("storage.remote.read-max-bytes-in-frame", "Maximum number of bytes in a single frame for streaming remote read response types before marshalling. Note that client might have limit on frame size as well.").
+		Default("1000").IntVar(&cfg.web.RemoteReadBytesInFrame)
```
brian-brazil: This default needs adjusting.

bwplotka: Yup. Moved to 1MB as probuf suggests that anything over 1MB should be streamed. Will perform some benchmarks with it.

prompb/remote.proto
```diff
@@ -23,10 +23,36 @@ message WriteRequest {
...
+    // Content-Type: "application/x-streamed-protobuf; proto=prometheus.ChunkedReadResponse"
+    // Content-Encoding: ""
+    STREAMED_XOR_CHUNKS = 1;
+  }
+
+  // accepted_response_types allows negotiating the content type of the response.
+  //
+  // Response types are taken from the list in the FIFO order. If no response type in `accepted_response_types` is
+  // implemented by server, error is returned.
+  // For request that do not support `accepted_response_types` field the SAMPLES response type will be used.
```
brian-brazil: s/support/contain/

bwplotka: done (:

bwplotka:
PTAL @brian-brazil @tomwilkie @krasi-georgiev 

I feel like we are close to merge this.

I will provide some benchmarks for this version. Let's merge and then apply `chunkIterator` once merged.

bwplotka:
Basic benchmark:

* 3 sequential requests:
* Used Images:
  * quay.io/thanos/prometheus:v2.12.0-rc.0-rr-streaming (built from this PR https://github.com/prometheus/prometheus/pull/5703/commits/f0af4045650963b02f75af9048e4787a2da59ecc)
  * quay.io/thanos/thanos:v0.6.0-master-rr-streaming (built from https://github.com/thanos-io/thanos/pull/1268)
* Max 1MB in frame.
* Avg request time 15s
* Frames:  10000
* Series: 1000
* Resources:

2019-08-18-14:17:33-Selection

It looks good, and from my side, this PR is ready to go. I plan to make more benchmarks with spam or requests etc while adjusting bytes number vs request concurrency limits, but all of this is in flags, and we can tweak defaults later on.

We will be rolling this on our clusters as well next week.

storage/remote/chunked.go
```diff
@@ -0,0 +1,153 @@
...
+type ChunkedReader struct {
+	b         *bufio.Reader
+	data      []byte
+	sizeLimit uint64
+
+	crc32 hash.Hash32
+}
+
+// NewChunkedReader constructs a ChunkedReader.
+func NewChunkedReader(r io.Reader, sizeLimit uint64) *ChunkedReader {
```
bwplotka: I think I will allow passing  `data []byte` as argument to this constructor to reuse byte slice even more.

web/api/v1/api_test.go
```diff
@@ -1012,10 +1013,188 @@ func TestReadEndpoint(t *testing.T) {
...
+
+func TestStreamReadEndpoint(t *testing.T) {
+	// 3 series each series.
+	// First with 119 samples. We expect 1 frame with 1 chunk.
+	// Second with 121 samples, We expect 1 frame with 2 chunks.
+	// Third with 241 samples. We expect 1 frame with 2 chunks, and 1 frame with 1 chunk for the same series due to bytes limit.
+	suite, err := promql.NewTest(t, `
+		load 1m
+			test_metric1{foo="bar1",baz="qux"} 0+100x119
+            test_metric1{foo="bar2",baz="qux"} 0+100x120
```
krasi-georgiev: Comment says `// Second with 121 samples, We expect 1 frame with 2 chunks.` but here we have 120 why?

```suggestion
            test_metric1{foo="bar2",baz="qux"} 0+100x121
```

web/api/v1/api_test.go
```diff
@@ -1012,10 +1013,188 @@ func TestReadEndpoint(t *testing.T) {
...
+	testutil.Ok(t, suite.Run())
+
+	api := &API{
+		Queryable:   suite.Storage(),
+		QueryEngine: suite.QueryEngine(),
+		config: func() config.Config {
+			return config.Config{
+				GlobalConfig: config.GlobalConfig{
+					ExternalLabels: labels.Labels{
+						{Name: "baz", Value: "a"},
```
krasi-georgiev: why are you adding this external label when it is already part of `test_metric1{foo="bar3",baz="qux"} 0+100x240` ?

web/api/v1/api_test.go
```diff
@@ -1012,10 +1013,188 @@ func TestReadEndpoint(t *testing.T) {
...
+						{Name: "baz", Value: "qux"},
+						{Name: "d", Value: "e"},
+						{Name: "foo", Value: "bar1"},
+					},
+					Chunks: []prompb.Chunk{
+						{
+							Type:      prompb.Chunk_XOR,
+							MaxTimeMs: 7140000,
+							Data:      []byte("\000x\000\000\000\000\000\000\000\000\000\340\324\003\302|\005\224\000\301\254}\351z2\320O\355\264n[\007\316\224\243md\371\320\375\032Pm\nS\235\016Q\255\006P\275\250\277\312\201Z\003(3\240R\207\332\005(\017\240\322\201\332=(\023\2402\203Z\007(w\2402\201Z\017(\023\265\227\364P\033@\245\007\364\nP\033C\245\002t\036P+@e\036\364\016Pk@e\002t:P;A\245\001\364\nS\373@\245\006t\006P+C\345\002\364\006Pk@\345\036t\nP\033A\245\003\364:P\033@\245\006t\016ZJ\377\\\205\313\210\327\270\017\345+F[\310\347E)\355\024\241\366\342}(v\215(N\203)\326\207(\336\203(V\332W\362\202t4\240m\005(\377AJ\006\320\322\202t\374\240\255\003(oA\312:\3202"),
+						},
```
krasi-georgiev: Is there no way to get the expected chunk data from somewhere instead of hard coding it?
I checked briefly , but can't see any way to get the raw chunk bytes, maybe after adding the Chunk iterator?

I am thinking that if the implementation of the test suite changes at some point this would require refactoring the test as well.

prompb/remote.proto
```diff
@@ -43,3 +69,14 @@ message QueryResult {
...
 }
+
+// ChunkedReadResponse is a response when response_type equals STREAMED_XOR_CHUNKS.
+// We strictly stream full series after series, optionally split by time. This means that a single frame can contain
+// partition of the single series, but once a new series is started to be streamed it means that no more chunks will
+// be sent for previous one.
+message ChunkedReadResponse {
+  repeated prometheus.ChunkedSeries chunked_series = 1;
+
+  // query_index represents an index of the query from ReadRequest.queries this results relates to.
```
brian-brazil: s/this results/these chunks/

web/api/v1/api.go
```diff
@@ -856,78 +860,150 @@ func (api *API) remoteRead(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	switch responseType {
+	case prompb.ReadRequest_STREAMED_XOR_CHUNKS:
+		w.Header().Set("Content-Type", "application/x-streamed-protobuf; proto=prometheus.ChunkedReadResponse")
+
+		f, ok := w.(http.Flusher)
+		if !ok {
+			http.Error(w, "internal http.ResponseWriter does not implement http.Flusher interface", http.StatusInternalServerError)
```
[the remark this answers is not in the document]
bwplotka: Isn't that a hard fail? Why panicking whole binary?

web/api/v1/api_test.go
bwplotka: so `0 + X` means X+1 samples (: 

Fixed `119 samples` comment as there are 120 exactly.

web/api/v1/api_test.go
bwplotka: to check if external labels are handled properly

web/api/v1/api_test.go
bwplotka: I know what you mean, but there is not much we can do here. Maybe, later on, we could potentially have a single place for proper encoding of chunks which will be tested in isolation and then use this here? For now, I would vote for leaving it hardcoded it's not that problematic.

web/api/v1/api_test.go
krasi-georgiev: yep that was the confusion, just the comment than, thanks

web/api/v1/api_test.go
krasi-georgiev: properly here means that adding an external label is ignored in this case?
Is this documented somewhere so I read about it? 
Maybe add a comment here as well.

web/api/v1/api_test.go
krasi-georgiev: When we merge the Chunk iterator should solve this, right?

bwplotka: Potentially

web/api/v1/api.go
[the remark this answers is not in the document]
bwplotka: We have test against that: https://github.com/prometheus/prometheus/blob/939d1b4263c26cc8256969a77de5948db01a53ca/web/api/v1/api_test.go#L1079

web/api/v1/api_test.go
bwplotka: Added comment

web/api/v1/api.go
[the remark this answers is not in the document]
krasi-georgiev: true, true!

krasi-georgiev:
:+1:

tomwilkie:
LGTM

PR #5711, opened 2019-06-27
Document behavior of empty ec2_sd_config region

svend:
Document the behavior of an empty `ec2_sd_config` `region` setting. If this is
omitted or blank, the region is discovered from the instance metadata, if available.
If it is blank and instance region metadata is not available, an error will
result ("EC2 SD configuration requires a region").

### DIFF
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -438,8 +438,8 @@ See below for the configuration options for EC2 discovery:
 ```yaml
 # The information to access the EC2 API.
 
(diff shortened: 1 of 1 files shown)


svend:
Here is the code where instance metadata is used: https://github.com/prometheus/prometheus/blob/a38a54fa117442c8d8c80020421941bf5c4145be/discovery/ec2/ec2.go#L91-L102

brian-brazil:
Thanks!

PR #5728, opened 2019-07-02
Clarifying honor_labels documentation

beorn7:
Previously, the wording could be misunderstood as setting honor_labels
to "false" for federation.

This also adds scraping the Pushgateway as a typical use case for
honor_labels=true.


brian-brazil:
:+1:

PR #5734, opened 2019-07-03
Add tests to ensure we can marshal and unmarshal our min/max times

jacksontj:
Related to https://github.com/prometheus/client_golang/issues/614

This simply showcases the exact issue. If this test passes then the issue will be resolved (either moving the min/max time so we can parse them, or implementing parsing for the times we currently have).

cc @brian-brazil  @krasi-georgiev

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -439,8 +439,11 @@ func (api *API) labelValues(r *http.Request) apiFuncResult {
 }
 
 var (
-	minTime = time.Unix(math.MinInt64/1000+62135596801, 0)
-	maxTime = time.Unix(math.MaxInt64/1000-62135596801, 999999999)
+	minTime = time.Unix(math.MinInt64/1000+62135596801, 0).UTC()
+	maxTime = time.Unix(math.MaxInt64/1000-62135596801, 999999999).UTC()
+
+	minTimeFormatted = minTime.Format(time.RFC3339Nano)
+	maxTimeFormatted = maxTime.Format(time.RFC3339Nano)
 )
 
 func (api *API) series(r *http.Request) apiFuncResult {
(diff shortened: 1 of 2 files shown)


web/api/v1/api.go
```diff
@@ -1162,6 +1165,12 @@ func parseTime(s string) (time.Time, error) {
 	if t, err := time.Parse(time.RFC3339Nano, s); err == nil {
 		return t, nil
 	}
+	switch s {
```
brian-brazil: Can you add a comment here?

jacksontj:
@brian-brazil comment updated, tests passed.

web/api/v1/api.go
```diff
@@ -1162,6 +1165,15 @@ func parseTime(s string) (time.Time, error) {
 	if t, err := time.Parse(time.RFC3339Nano, s); err == nil {
 		return t, nil
 	}
+
+	// Stdlib's time parser can only handle 4 digit years. As a workaround until
+	// that is fixed we want to at least support our own boundary times
```
brian-brazil: Full stop at the end of comments. Can you link the upstream issue?

jacksontj:
@brian-brazil Comment updated again, I included links to both issues.

jacksontj:
Unfortunately this requires a prometheus release+deploy for anyone impacted by this issue (server-side fix not just a client-side fix), will we be able to get this into 2.11 ?

brian-brazil:
:+1: 

@brancz You own 2.11, this would be good to get in.

web/api/v1/api.go
```diff
@@ -441,6 +441,9 @@ func (api *API) labelValues(r *http.Request) apiFuncResult {
 var (
 	minTime = time.Unix(math.MinInt64/1000+62135596801, 0)
 	maxTime = time.Unix(math.MaxInt64/1000-62135596801, 999999999)
+
+	minTimeFormatted = minTime.Format(time.RFC3339Nano)
```
valyala: Probably the time must be converted to `UTC` before formatting?

I get the following values for `minTime` in local time zone and in `UTC` time zones:

```
local: -292273086-05-16T18:49:10+02:02
UTC: -292273086-05-16T16:47:06Z
```

jacksontj: Good catch, I've actually added it to the minTime and maxTime instead -- same outcome though :)

brian-brazil:
Considering that this is going to be handled on client_golang by switching to unix timestamps, I'm just going to merge now and not worry about getting this in to 2.11.

PR #5740, opened 2019-07-06
make targets active parallel

### LINKED ISSUE #5717: Make Manager.TargetsActive parallel
Currently https://github.com/prometheus/prometheus/blob/32c3a1beeffbcf0d308f724931d8ed46e8048018/scrape/manager.go#L222 goes in series through the scrape pools. If there's an update from SD, this can end up blocking on each scrapePool's `Sync` in turn. Doing this in parallel would limit this to the scrape interval.

yeya24:

fix #5717

### DIFF
--- a/scrape/manager.go
+++ b/scrape/manager.go
@@ -213,7 +213,6 @@ func (m *Manager) TargetsAll() map[string][]*Target {
 	targets := make(map[string][]*Target, len(m.scrapePools))
 	for tset, sp := range m.scrapePools {
 		targets[tset] = append(sp.ActiveTargets(), sp.DroppedTargets()...)
-
 	}
 	return targets
 }
@@ -223,10 +222,24 @@ func (m *Manager) TargetsActive() map[string][]*Target {
 	m.mtxScrape.Lock()
 	defer m.mtxScrape.Unlock()
 
+	var (
+		wg  sync.WaitGroup
+		mtx sync.Mutex
+	)
+
 	targets := make(map[string][]*Target, len(m.scrapePools))
+	wg.Add(len(m.scrapePools))
 	for tset, sp := range m.scrapePools {
-		targets[tset] = sp.ActiveTargets()
+		// Running in parallel limits the blocking time of scrapePool to scrape
+		// interval when there's an update from SD.
+		go func(tset string, sp *scrapePool) {
+			mtx.Lock()
+			targets[tset] = sp.ActiveTargets()
+			mtx.Unlock()
+			wg.Done()
+		}(tset, sp)
 	}
+	wg.Wait()
 	return targets
 }
 
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -66,7 +66,7 @@ var (
 	targetScrapePools = prometheus.NewCounter(
 		prometheus.CounterOpts{
 			Name: "prometheus_target_scrape_pools_total",
-			Help: "Total number of scrape pool creation atttempts.",
+			Help: "Total number of scrape pool creation attempts.",
 		},
 	)
 	targetScrapePoolsFailed = prometheus.NewCounter(


scrape/manager.go
```diff
@@ -223,10 +223,18 @@ func (m *Manager) TargetsActive() map[string][]*Target {
 	m.mtxScrape.Lock()
 	defer m.mtxScrape.Unlock()
 
+	var wg sync.WaitGroup
 	targets := make(map[string][]*Target, len(m.scrapePools))
 	for tset, sp := range m.scrapePools {
-		targets[tset] = sp.ActiveTargets()
+		wg.Add(1)
+		// Run in parallel limit the blocking time of scrapePool to scrape
```
codesome:
```suggestion
		// Runing in parallel limits the blocking time of scrapePool to scrape
```

scrape/manager.go
```diff
@@ -223,10 +223,18 @@ func (m *Manager) TargetsActive() map[string][]*Target {
 	m.mtxScrape.Lock()
 	defer m.mtxScrape.Unlock()
 
+	var wg sync.WaitGroup
 	targets := make(map[string][]*Target, len(m.scrapePools))
 	for tset, sp := range m.scrapePools {
-		targets[tset] = sp.ActiveTargets()
+		wg.Add(1)
```
codesome: You can have a single `wg.Add(len(m.scrapePools))` call before the `for` loop.

scrape/manager.go
codesome: `Runing` -> `Running`

scrape/manager.go
yeya24: Done.

scrape/manager.go
yeya24: Done.

scrape/manager.go
```diff
@@ -223,10 +223,18 @@ func (m *Manager) TargetsActive() map[string][]*Target {
...
 
+	var wg sync.WaitGroup
 	targets := make(map[string][]*Target, len(m.scrapePools))
+	wg.Add(len(m.scrapePools))
 	for tset, sp := range m.scrapePools {
-		targets[tset] = sp.ActiveTargets()
+		// Running in parallel limits the blocking time of scrapePool to scrape
+		// interval when there's an update from SD.
+		go func(tset string, sp *scrapePool) {
+			targets[tset] = sp.ActiveTargets()
```
brian-brazil: Do we need some locking here?

brian-brazil: Did you have a chance to look at this?

yeya24: Yes, I will finish it  soon.

scrape/manager.go
```diff
@@ -61,6 +61,7 @@ type Manager struct {
 	graceShut chan struct{}
 
 	jitterSeed    uint64     // Global jitterSeed seed is used to spread scrape workload across HA setup.
+	mtxTargets    sync.Mutex // Guards for active targets.
```
brian-brazil: This should be a local variable, it's only protecting a local variable.

scrape/manager.go
sylr: I might be wrong but isn't this a case of https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables ?

```
go func(tset string, sp *scrapePool) {
 	...
}(tset, &m.scrapePools[tset])
```

https://play.golang.org/p/FTB20cdWLXu

brian-brazil: The code looks fine to me

sylr: I think we do need locking. As far as I understand you can't have several goroutines writing in the same map at the same time (even if they modify different keys of the map).

brian-brazil: The locking has already been added

sylr: > The code looks fine to me

It looks fine but the goroutines are stacked and can be started after the end of the iteration of the `for loop` (especially if there is nothing else in the for loop body that the goroutine calls).

If you pass a pointer as argument of the goroutine function closure you might end up on calling all the goroutines with the same pointer, the one of the last element of the map you iterate on.

See the following example: https://play.golang.org/p/FTB20cdWLXu

brian-brazil: We're not passing `&sp` though, we're passing `sp`

sylr: Indeed, it works in this case: https://play.golang.org/p/XuBiUEu59ci

brian-brazil:
Thanks!

scrape/manager.go
```diff
@@ -223,10 +222,24 @@ func (m *Manager) TargetsActive() map[string][]*Target {
...
+	)
+
 	targets := make(map[string][]*Target, len(m.scrapePools))
+	wg.Add(len(m.scrapePools))
 	for tset, sp := range m.scrapePools {
-		targets[tset] = sp.ActiveTargets()
+		// Running in parallel limits the blocking time of scrapePool to scrape
+		// interval when there's an update from SD.
+		go func(tset string, sp *scrapePool) {
+			mtx.Lock()
```
bboreham, line 236: Does this really run in parallel?  As far as I can see it's all serialized by the mutex - what did you gain?

PR #5741, opened 2019-07-08
bugfix: pass error up if WALWathcer.segments() return err

YaoZengzeng:

### DIFF
--- a/storage/remote/wal_watcher.go
+++ b/storage/remote/wal_watcher.go
@@ -234,7 +234,7 @@ func (w *WALWatcher) run() error {
 func (w *WALWatcher) findSegmentForIndex(index int) (int, error) {
 	refs, err := w.segments(w.walDir)
 	if err != nil {
-		return -1, nil
+		return -1, err
 	}
 
 	for _, r := range refs {
@@ -249,7 +249,7 @@ func (w *WALWatcher) findSegmentForIndex(index int) (int, error) {
 func (w *WALWatcher) firstAndLast() (int, int, error) {
 	refs, err := w.segments(w.walDir)
 	if err != nil {
-		return -1, -1, nil
+		return -1, -1, err
 	}
 
 	if len(refs) == 0 {


tomwilkie:
ping @cstyan

cstyan:
@YaoZengzeng did you run into an actual bug that was caused by this error not being bubbled up? If so, can you add a test?

Otherwise lgtm.

YaoZengzeng:
@cstyan 
I have not encountered it in practice actually.

But in theory, if the segments are not sequential, the segments() would return error and we should bubble up the error ASAP, instead of replaying the WAL until encounter the missed segments.

In addition, the functions with similar function in TSDB bubbles up the error when list segments failed.
https://github.com/prometheus/tsdb/blob/master/wal/wal.go#L634

So it's reasonable to bubble up the error here :)

cstyan:
Yes, the error should be bubbled up. I was just wondering if you'd come across an actual bug.

> But in theory, if the segments are not sequential, the segments() would return error and we should bubble up the error ASAP, instead of replaying the WAL until encounter the missed segments.

In this particular case Prometheus will exit with an error on startup before the WAL watcher code even starts :)

YaoZengzeng:
@cstyan 
I just read through the WAL and TSDB related code to try to figure out why memory surge when WAL watcher and remote write are enabled in Prometheus v2.8.0

And find this bug by accident, haha...

cstyan:
2.8 included a rewrite of remote write that will use more memory in the happy path than previous versions, but in failure cases it should use less memory than before, and not lose samples or OOM your Prometheus. 

There have been numerous small improvements since then though, if you're having issues with memory usage when using remote write try 2.10/2.11 and then open an issue (and ping me!) if you're still seeing something problematic.

YaoZengzeng:
@cstyan 
Yes , thanks!

Also, I'd to contribute if there are issues to be fixed :joy:

tomwilkie:
LGTM from me too.

PR #5742, opened 2019-07-09
*: cut v2.11.0

brancz:
[Benchmarks](http://prombench.prometheus.io/grafana/d/7gmLoNDmz/prombench?orgId=1&var-RuleGroup=All&var-pr-number=5729) showing no performance regressions, no new bugs reported. Cutting final v2.11.0. :tada: 

@brian-brazil @beorn7 @gouthamve @juliusv @simonpasquier @krasi-georgiev @cstyan @codesome


brian-brazil:
:+1:

simonpasquier:
:+1:

beorn7:
I recommend to point out the new flags to activate optional features in the announcement mails.

beorn7:
Sorry, that should be a flag (`--storage.tsdb.wal-compression`) for WAL compression and a config setting (`api_version` in the alertmanager config) for the Alertmanager API.

brancz:
Good point! Will do! We should probably also have docs for the wal compression, which I don't think we do right now.

csmarchbanks:
> We should probably also have docs for the wal compression, which I don't think we do right now.

Good idea, I will open a PR with some documentation for the wal compression today.

PR #1987, opened 2016-09-13
Switch back to protos over HTTP, instead of grpc.

tomwilkie:
Don't worry @brian-brazil, they're still snappy compressed.

Fixes #1982

PR #5743, opened 2019-07-09
remove the dependency on cockroachdb in prometheus

tariq1890:
I am not sure why we are using cockroachdb for just a mere json.Marshaler based interface. I think it would be best to remove such a heavyweight dependency if we can help it


brancz:
lgtm :+1: thanks!

tariq1890:
Thanks for the review @brancz. Requesting a review from @brian-brazil as well since he was involved in this PR that may be related : https://github.com/prometheus/prometheus/pull/1987

I had a look at the discussion as well the relevant vendor codes and these changes seemed fine to me.

brian-brazil:
I was just waiting for the tests to pass, and that PR seems unrelated. I was wondering why we pulled this in.

brancz:
I'm assuming we accidentally took it when we introduced this in the first place so we missed the dependency mess. I followed the code and this seems to be identical as the previously vendored code, so it looks good to me.

beorn7:
I guess that was an go-imports snafu. (It just grabs the first thing it finds matching the package and type name.) Funny how we shipped all those dependencies just because of a single mis-import.

bwplotka:
Wow, that was interesting. Thanks for fixing @tariq1890 !

simonpasquier:
:tada:

PR #5749, opened 2019-07-09
Fix potential panic when prometheus is watching multiple zookeeper paths

meridional:
#5675 introduced a serious bug: if prometheus is configured to watch multiple zk paths, multiple tree caches will share the same updates channel. When reloaded, each treecache will call `close` on the `updates` channel, causing a panic.

This patch fixes the issue by using fan-in pattern, where each treecache gets its own update channel, and in zookeeper library, we merge the events from each channel into `zookeeper.Discovery.updates`.

Sorry for the interruptions caused by #5675!


brian-brazil:
:+1: 

@brancz Bugfix for 2.11.0

beorn7:
@brian-brazil As you correctly identified this as a bugfix for v2.11, you could as well have changed the target branch to be release-2.11. New @brancz needed to cherry-pick, which we want to avoid.

brancz:
@beorn7 you're right. My mistake.

brian-brazil:
I was leaving this up to the release owner, which is why I didn't merge it.

beorn7:
As cherry-picking is explicitly marked as something we'd like to avoid in the instructions, it would have been a friendly move to point out that the base branch should not be master. A :+1: by an experienced core developer conveys the meaning that all is good with the PR.

PR #5753, opened 2019-07-11
Fix prometheus-mixin dashboards to use grafanaDashboards

metalmatze:
Updating to the latest version of this prometheus-mixin, I found that it doesn't show up in my Grafana.

Instead of using `grafanaDashboards+::` the mixin currently uses `dashboards+::` which is wrong.
The confusion is probably coming from the original [Monitoring Mixin Design Doc](https://docs.google.com/document/d/1A9xvzwqnFVSOZ5fD3blKODXfsat5fg6ZhnKu9LK3lB4/edit#) which first talks about `dashboards+::` and then further down actually uses `grafanaDashboards+::` (search for Proposed Schema).

Here are a few examples using `grafanaDashboards+::`:
[kubernetes-mixin](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/dashboards/use.libsonnet#L4)
[etcd-mixin](https://github.com/etcd-io/etcd/blob/master/Documentation/etcd-mixin/mixin.libsonnet#L198)

### DIFF
--- a/documentation/prometheus-mixin/dashboards.jsonnet
+++ b/documentation/prometheus-mixin/dashboards.jsonnet
@@ -1,4 +1,4 @@
-local dashboards = (import 'mixin.libsonnet').dashboards;
+local dashboards = (import 'mixin.libsonnet').grafanaDashboards;
 
 {
   [name]: dashboards[name]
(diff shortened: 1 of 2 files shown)


brancz:
lgtm

PR #5756, opened 2019-07-11
Only check last directory when discovering checkpoint number

dtrejod:
We keep our Prometheus TSDB data under a hidden directory which breaks `remote_writes` since the function that finds the checkpoint number simply splits on `.`. 

This PR ensures that we only look at the base directory when discovery the checkpoint number.


cstyan:
You were getting the error case then? `invalid checkpoint dir...`? This is the correct fix :+1: Thank you!

Can you add a comment to go along with the checkpoint format comment?

dtrejod:
Added comments. 

Yes were we hitting the error case. Made the bug easy to find 😄

tomwilkie:
LGTM too.

PR #5759, opened 2019-07-12
Mark deleted rule's series as stale on next evaluation.

### LINKED ISSUE #5755: Updating information through recording rules
## Bug Report
**What did you do?**
We use recording rules to create metrics with information about system. Something like location and status to be able to pass this information to alertmanager. This information is updated through an automated process and it reloads prometheus.
After the update to version 2.10 we see the metric with old and new labels.

e.g.
```
asset_info{instance="test1",status="a",location="a"}
asset_info{instance="test1",status="b",location="a"}
```

In version 2.8.0 there was only one metric after the prometheus reload.

The query we use to trigger this looks like this.
```
up{exported_instance="i"} * on(exported_instance) group_left(status, location) asset_info
```

After reload the asset information is updated and not reported twice.

The asset metric is shown twice instead of one after an update of a value.

**Environment**

* Prometheus version:
```
# prometheus --version
prometheus, version 2.10.0 (branch: HEAD, revision: d20e84d0fb64aff2f62a977adc8cfb656da4e286)
  build user:       root@a49185acd9b0
  build date:       20190525-12:28:13
  go version:       go1.12.5
```

* Testcase
```
#!/bin/bash

function cleanup() {
    kill $WEB
    kill $PROMETHEUS
}

trap cleanup SIGINT SIGTERM

TMP=$(mktemp -d)

echo "changing to $TMP"

cd $TMP
echo 'up{instance="i"} 1' > metrics


cat > prometheus.yml <<EOF
global:
  scrape_interval: 10s
  scrape_timeout: 5s
  evaluation_interval: 10s
  external_labels:
    prometheus_scraper: test

rule_files:
  - rules.yml

scrape_configs:
  - job_name: 'example1'
    metrics_path: '/metrics'
    static_configs:
      - targets: [127.0.0.1:8000]
EOF

cat > rules.yml <<EOF
groups:
  - name: asset_info
    interval: 10s
    rules:
      - record: asset_info
        expr: 1
        labels:
          exported_instance: "i"
          asset_status: "test"
          asset_location: "test"
          testlabel: "a"

  - name: linux
    interval: 10s
    rules:
      - alert: TestAlert
        expr: up{exported_instance="i"} * on(exported_instance) group_left(asset_status, asset_location) asset_info
        for: 2m
EOF

python -m SimpleHTTPServer &
WEB=$!
prometheus &
PROMETHEUS=$!

sleep 15

sed -e 's/testlabel: ".*"/testlabel: "'$RANDOM'"/' -i rules.yml
kill -SIGHUP $PROMETHEUS

wait
```

If you run this, you will see evaluation errors after some seconds.

**Affected version**
I have bisected it and found that change e15d8c5802f98b6a508022500acf44e91adc7d87 is the culprit here. It is also the case in master branch.

brian-brazil:
Fixes #5755

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -223,6 +223,7 @@ type Group struct {
 	interval             time.Duration
 	rules                []Rule
 	seriesInPreviousEval []map[string]labels.Labels // One per Rule.
+	staleSeries          []labels.Labels
 	opts                 *ManagerOptions
 	mtx                  sync.Mutex
 	evaluationDuration   time.Duration
@@ -480,6 +481,18 @@ func (g *Group) CopyState(from *Group) {
 			ar.active[fp] = a
 		}
 	}
+
+	// Handle deleted and unmatched duplicate rules.
+	g.staleSeries = from.staleSeries
+	for fi, fromRule := range from.rules {
+		nameAndLabels := nameAndLabels(fromRule)
+		l := ruleMap[nameAndLabels]
+		if len(l) != 0 {
+			for _, series := range from.seriesInPreviousEval[fi] {
+				g.staleSeries = append(g.staleSeries, series)
+			}
+		}
+	}
 }
 
 // Eval runs a single evaluation cycle in which all rules are evaluated sequentially.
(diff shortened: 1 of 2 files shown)


rules/manager.go
```diff
@@ -480,6 +481,18 @@ func (g *Group) CopyState(from *Group) {
...
 	}
+
+	// Handle deleted and unmatched duplicate rules.
+	g.staleSeries = from.staleSeries
+	for fi, fromRule := range from.rules {
+		nameAndLabels := nameAndLabels(fromRule)
+		l := ruleMap[nameAndLabels]
+		if len(l) != 0 {
+			for _, series := range from.seriesInPreviousEval[fi] {
+				g.staleSeries = append(from.staleSeries, series)
```
csmarchbanks: Is this supposed to be `g.staleSeries = append(g.staleSeries, series)`?

rules/manager.go
```diff
@@ -575,6 +588,32 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
...
+
+	if len(g.staleSeries) != 0 {
+		app, err := g.opts.Appendable.Appender()
+		if err != nil {
+			level.Warn(g.logger).Log("msg", "creating appender failed", "err", err)
+			return
+		}
+		for _, s := range g.staleSeries {
+			// Rule that produced series no longer configured, mark it stale.
+			_, err = app.Add(s, timestamp.FromTime(ts), math.Float64frombits(value.StaleNaN))
```
csmarchbanks: It might be nice to factor out the append and special error handling for out of order/duplicate samples to one place.

rules/manager.go
brian-brazil: Fixed

rules/manager.go
brian-brazil: We need different handling in different places though

csmarchbanks: :+1: It's just a different message, but I agree it wouldn't make anything clearer.

juliusv:
:+1:

Issue #5627, opened 2019-06-03
Remote Write Shard Flapping


BertHartm:
## Proposal
Since upgrading to 2.8 / 2.9, it's become very hard to tune the number of shards, and the shards seem to flap between min and max shards if there's any sort of issue.

I'm currently mitigating by trying to set all of my prometheus instances to have enough shards to not need to scale up in normal circumstances, and leaving a low max_shards so as to not go crazy if there's an issue (flapping to 1000 was not fun).

This additionally causes issues because resharding causes http connection churn and smaller than normal batches, both of which lead to higher latency / more flapping on other boxes with the same remote write target.

## Bug Report
**What did you do?**
Set up remote write, in this case with the following settings 
```
- url: redacted
  remote_timeout: 30s
  queue_config:
    capacity: 10000
    max_shards: 10
    min_shards: 6
    max_samples_per_send: 5000
    batch_send_deadline: 5m
    max_retries: 3
    min_backoff: 50ms
    max_backoff: 1s
```

**What did you expect to see?**
Smooth movement between 6 and 10 shards as needed to send the data.

**What did you see instead? Under which circumstances?**
Sharp jumps between 6 and 10 shards whenever there's any sort of network latency.
Graph, 15s resolution, # shards in green, remote write network time in yellow `(rate(prometheus_remote_storage_sent_batch_duration_seconds_sum{instance=~"$instance"}[$__interval])`
![image](https://user-images.githubusercontent.com/508811/58824718-bc645e00-860a-11e9-9105-fa6840d176ea.png)

**Environment**

* System information: Linux 4.15.0-1045-azure x86_64

* Prometheus version:
```
prometheus, version 2.9.2 (branch: HEAD, revision: d3245f15022551c6fc8281766ea62db4d71e2747)
  build user:       root@1d43b6951e8f
  build date:       20190424-15:32:31
  go version:       go1.12.4
```

* Alertmanager version: N/A

* Prometheus configuration file:

```
remote_write:
- url: redacted
  remote_timeout: 30s
  queue_config:
    capacity: 10000
    max_shards: 10
    min_shards: 6
    max_samples_per_send: 5000
    batch_send_deadline: 5m
    max_retries: 3
    min_backoff: 50ms
    max_backoff: 1s
```

* Alertmanager configuration file:
N/A

* Logs:
```
[...]

PR #5763, opened 2019-07-14
Improve desired shards calculation

csmarchbanks:
This includes three changes that help drastically with remote write shards flapping.
1. `samplesPending` needs to be a rate like the other measurements. Otherwise the units when doing math don't work out. Previously the units for desired shards were `s * count`, giving seconds, now they are `s * count/s` giving a count.
1. Accounting for `samplesIn` and `samplesPending` on line 461. Previously, whenever the queue caught up with data it would instantly think it needed zero shards which caused the crazy changes from min_shards to max_shards.
1. Making the integral accumulator used again let the calculation avoid flapping between two moderate values.

cc: @cstyan @tomwilkie This works well in my testing, but would appreciate this exercised with more data.

Fixes https://github.com/prometheus/prometheus/issues/5627


cstyan:
Awesome, thank you @csmarchbanks! I was starting to get disappointed I hadn't been able to look into this more myself yet.

I'll pick one of our instances that's has reshard oscillation tomorrow and deploy these changes there.

tomwilkie:
Thanks @csmarchbanks! Can you post before any after effect on num_shards?

csmarchbanks:
Right now this seems to be finding the correct number of shards for the incoming sample rate very well, but does not always catch up with old samples as well as I would like. Investigating...

csmarchbanks:
Number of shards looks much much lower in general on this test Prometheus sending ~50k samples per second.
Shards Before and After

Still a bit more refinement to go though.

csmarchbanks:
It appears that increasing the capacity on the queue greatly helps with the sharding calculation (and shard throughput). When the capacity was 10, the timePerSample calculation was off as the shard was spending a reasonable amount of time waiting for samples, but when the capacity is equal to max samples per send then the next request can go out very quickly and dominates the time. 

WDYT about making the default capacity equal to whatever the user's max samples per send is?

storage/remote/queue_manager.go
```diff
@@ -440,36 +443,50 @@ func (t *QueueManager) calculateDesiredShards() {
...
 	var (
 		timePerSample = samplesOutDuration / samplesOut
-		desiredShards = (timePerSample * samplesPending) / float64(time.Second)
+		desiredShards = timePerSample * (samplesInRate + t.integralAccumulator)
 	)
 	level.Debug(t.logger).Log("msg", "QueueManager.caclulateDesiredShards",
-		"samplesIn", samplesIn,
-		"samplesOut", samplesOut,
+		"samplesInRate", samplesInRate,
+		"samplesOutRate", samplesOut,
```
YaoZengzeng: Maybe it's better to mv `samplesOut` with `samplesOutRate` to keep the consistent.

And what if when the wal watcher start , the network to remote storage is broken , so `samplesOut` and `samplesOutDuration` both will be 0, then timePerSample will be NaN ?

csmarchbanks: Good catch, I thought I renamed that variable, but clearly not.

If `samplesOut` is 0 then this entire function is returned early so NaN's will be avoided.

YaoZengzeng: Hoo.. I missed the judgement of of `SamplesOut` above.

Other code LGTM and I have deployed the code of this PR in my environment.

I'll have a look tomorrow morning to check if it works well, hahaha..

cstyan:
2019-07-16-202406_2408x392_scrot
Left is with the capacity changes, right is before those changes (when I first deployed this branch)

storage/remote/queue_manager.go
csmarchbanks: @YaoZengzeng How did this end up working for you?

YaoZengzeng:
@csmarchbanks 

It seems work fine for me.

image

The memory usage of prometheus server is stable. 

However in this case, network is stable and the sample input is not very big, so the shard number is stable.

storage/remote/queue_manager.go
```diff
@@ -440,36 +443,50 @@ func (t *QueueManager) calculateDesiredShards() {
...
 		highestRecv        = highestTimestamp.Get()
-		samplesPending     = (highestRecv - highestSent) * samplesIn * samplesKeptRatio
+		samplesPending     = (highestRecv - highestSent) * samplesInRate * samplesKeptRatio
 	)
 
-	// We use an integral accumulator, like in a PID, to help dampen oscillation.
-	t.integralAccumulator = t.integralAccumulator + (samplesPending * 0.1)
-
-	if samplesOut <= 0 {
+	if samplesOutRate <= 0 {
```
YaoZengzeng: Actually I find some problems here :joy: 

If the network between prometheus and remote storage is shaking and shard number become 1000 after reshard.

However the network traffic of 1000 shard is a disaster for remote storage which can't be suffered and it crushed.

Then remote storage will restart again and again, because the sampleOutRate is 0, and the number of shard still be 1000 which result a deadlock.

So if the samplesOutRate is 0, maybe we should return shard number back to min to make the network recover.

csmarchbanks: That's an interesting problem, if you are backed up far enough then most users would want to catch up as quickly as possible, so scaling to max shards makes sense to me. If the remote storage cannot keep up, it is likely you will want a smaller `max_shards` that the remote storage can handle. Curious what others think as well.

bboreham: I have always thought 1,000 is way too high a default. However people demanded proof.

brian-brazil: We should ensure the default can handle a 1M/s instance, with slack for catchup.

cstyan: I don't know if there's really a nice way that resharding can account for issues on the remote storage's end. Maybe we could take into account retries somehow, but reconfiguring Prometheus and the remote storage is going to yield a better result. 

If the amount of network traffic and sheer number of requests is too much for the remote storage to handle you can either scale up your remote storage or mess with your remote write config.

csmarchbanks:
More research/thoughts on shard calculations and capacity:

What is happening now (when you have relatively low capacities configured) is that one shard will backup, say by processing a p99 or p999 request. Those requests can take 10-100x as long as a p50 request, so even with a capacity equal to `max_samples_per_send` reading from the WAL will backup on that one blocked shard.

The desired shard calculation is based on the time it takes to send a sample, only including the time spent doing the remote call. All the time spent blocked is not accounted for, and if we do try to account for blocked time it is hard to know if it is blocked due to another shard being slow, or if it is blocked waiting for a sample to appear on the WAL (in which case we should scale down). This can be addressed by increasing the capacity of each shard so that one slow request does not block all of the shards.

My recommendations are to either put capacity back to a large number (perhaps 1000, still less than the 10,000 before the remote write rewrite), or to default capacity to `10 * max_samples_per_send` in order to bias towards throughput and accurate shard calculations that will work for most users. I would love to hear what other remote write users think though!

csmarchbanks:
@tomwilkie (and anyone else interested) This is ready for a review. I decided to keep it simple, and just have a better default capacity and some notes in the documentation. A capacity of 500 seems to be working pretty well for 100 samples per send in my test Prometheus.

cstyan:
@csmarchbanks okay I've tested this on some of our larger instances and everything seems to be running smoothly, better throughput and more stable in terms of # of shards. 

I think the default capacity of 500 is too high though. What do you think of a default of 1.5x or 2.x the default max_samples_per_send?

cc @tomwilkie @gouthamve this has :+1: from me

csmarchbanks:
Glad to hear it is working well for you!

I am curious why you think a default capacity of 500 is too high? Are you concerned about memory usage or something else?

How I got 5x is that the p99 of my remote write request latency compared to the average varies from 3x to 10x depending on the environment/queue. I would prefer 10x, but put up 5x as a compromise. I would not like to go much below 3x (p95 / avg in my clusters). Query I use:
```
histogram_quantile(0.99, 
  rate(prometheus_remote_storage_sent_batch_duration_seconds_bucket[24h])
) / (
  rate(prometheus_remote_storage_sent_batch_duration_seconds_sum[24h])
  /  
  rate(prometheus_remote_storage_sent_batch_duration_seconds_count[24h])
)
```
I would be interested to see how this query behaves in your systems.

If worried about memory:

If https://github.com/prometheus/prometheus/pull/5849 is accepted, a queue will use 40 bytes per sample. Quick math gives `1000 shards * 600 samples * 40 bytes per sample = 22.9 MB`. 

The current implementation uses 128 bytes per sample so all queues would use 73.2 MB, either number doesn't seem like too much for a default to me.

storage/remote/queue_manager.go
tomwilkie: Yeah, there are strong reasons for 1k shards (which can each do 1k samples/s), and we've made the limit configurable to support other usecases.

I don't wonder wether we can improve the backoff so that we only ever for 10 QPS per shard, to prevent us DoSing the endpoint when it is down... but thats a separate issue.

storage/remote/queue_manager_test.go
```diff
@@ -362,6 +362,11 @@ func (c *TestStorageClient) waitForExpectedSampleCount() {
 }
 
 func (c *TestStorageClient) Store(_ context.Context, req []byte) error {
+	defer func() {
+		if r := recover(); r != nil {
+			fmt.Println("recovered")
+		}
+	}()
```
tomwilkie: This look spurious?

csmarchbanks: Yep, was used for debugging a test failure. Removed for now, but will work on making test error more obvious in the future.

PR #5765, opened 2019-07-15
Benchmark tsdb master

codesome:
**DO NOT MERGE**

## Benchmark 1

Benchmark the following PRs against 2.11.1

1. For queries: https://github.com/prometheus/tsdb/pull/642
2. For compaction: https://github.com/prometheus/tsdb/pull/643 https://github.com/prometheus/tsdb/pull/654 https://github.com/prometheus/tsdb/pull/653 
3. Opening block: https://github.com/prometheus/tsdb/pull/645

**Results**

Did not test for compaction from on-disk blocks. Could not really see the allocation optimizations in compaction, that might be because the savings are mostly in the number of allocations and not the size of allocation (size is what is showed in the dashboards). That would mean CPU to be saved, but couldn't make a huge difference, but a slight increase in gap during compaction.

The gains looked good in 
1. Allocations
2. CPU (because of allocations?)
3. RSS was also lower (upto 10 GiB lower! ~60 vs ~70).
4. Also a tiny-good improvement in query `inner_eval` times.
5. Compaction time (this should help the increase in compaction time that https://github.com/prometheus/tsdb/pull/627 is going to bring).
6. System load.

And bad in
1. `result_sort` for the queries. Not sure why.

## Benchmark 2

Benchmark https://github.com/prometheus/tsdb/pull/627 (which includes all the PRs from above Benchmark 1) against 2.11.1


codesome:
I am stopping the benchmark (did not test for compaction from on-disk blocks). Could not really see the allocation optimizations in compaction, that might be because the savings are mostly in the number of allocations and not the size of allocation (size is what is showed in the dashboards). That would mean CPU to be saved, but couldn't make a huge difference, but a slight increase in gap during compaction.

The gains looked good in 
1. Allocations
2. CPU (because of allocations?)
3. RSS was also lower (upto 10 GiB lower! ~60 vs ~70).
4. Also a tiny-good improvement in query `inner_eval` times.
5. Compaction time (this should help the increase in compaction time that https://github.com/prometheus/tsdb/pull/627 is going to bring).
6. System load.

And bad in
1. `result_sort` for the queries. Not sure why.

codesome:
Removed old vendoring and added https://github.com/prometheus/tsdb/pull/627. Will benchmark memory savings in compaction now.

codesome:
Last 2 benchmarks were a failure. One had [a bug](https://github.com/prometheus/tsdb/pull/627#issuecomment-511768986). In the next, I fixed the bug first and then induced the bug again trying to optimize the bugfix. Reverted the last change and I think I have finally fixed the bug.

I will leave the benchmark running overnight.

codesome:
The last test never triggered compaction for this PR and was OOM killed, strange. The only change from the last test was a modification on how we compact, and not when we compact, so compaction should have triggered. Investigating.

codesome:
Head compaction is stuck and never exits. Adding more logs to debug.

codesome:
Finally fixed it. Previously the compaction was going on forever due to inefficiency in `compactionMerger` in removing the ref. I fixed it in the last commit (fix was to not remove the ref from the `seriesMap` at all when there are 0 chunks). 
Memory savings are noticeable for the head compaction http://prombench.prometheus.io/grafana/d/7gmLoNDmz/prombench?orgId=1&var-RuleGroup=All&var-pr-number=5765&from=1563871789157&to=1563874170975 , waiting for the compaction from disk blocks.

codesome:
Re-started the benchmark as in the previous benchmark the prometheus@master build was of yesterday because the build after https://github.com/prometheus/prometheus/pull/5793 was merged was not ready yet. It's ready now.

codesome:
The last benchmark uncovered yet another bug (related to all previous bugs, but the solution was not proper before). Hopefully, I have finally fixed it.

codesome:
Status: Blocks from disk are never compacted again and OOM. Will add more logs in investigate.

codesome:
The benchmark shows that memory consumption is not reduced. But the profiles show that the master allocates way more memory during compaction than this PR, but it is not reflected in the "Allocations" dashboard. Investigating.

PR #5772, opened 2019-07-17
promtool command to obfuscate symbols of index file

codesome:
While working on postings compression with @naivewong we needed real-world index file to run tests on it. This command will help obfuscate the index file to hide any sensitive data.

This might also help in sharing index files for debugging without worrying about sensitive data.

### DIFF
--- a/cmd/promtool/main.go
+++ b/cmd/promtool/main.go
@@ -23,6 +23,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -35,6 +36,9 @@ import (
 	config_util "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/common/version"
+	"github.com/prometheus/tsdb/chunks"
+	"github.com/prometheus/tsdb/index"
+	"github.com/prometheus/tsdb/labels"
 	kingpin "gopkg.in/alecthomas/kingpin.v2"
 
 	"github.com/prometheus/prometheus/config"
@@ -102,6 +106,16 @@ func main() {
 		"The unit test file.",
 	).Required().ExistingFiles()
 
+	obfuscateIndexCmd := app.Command("obfuscate_index", "Obfuscate index.")
+	obfuscateIndexInput := obfuscateIndexCmd.Arg(
+		"input",
+		"The input index file.",
+	).Required().ExistingFile()
+	obfuscateIndexOutput := obfuscateIndexCmd.Arg(
+		"output",
+		"The output index file.",
+	).Required().String()
+
 	parsedCmd := kingpin.MustParse(app.Parse(os.Args[1:]))
 
 	var p printer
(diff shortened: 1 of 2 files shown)


cmd/promtool/main.go
```diff
@@ -641,3 +659,185 @@ func (j *jsonPrinter) printLabelValues(v model.LabelValues) {
...
+	sort.Strings(names)
+
+	for _, n := range names {
+		labelValuesBuf = labelValuesBuf[:0]
+		for v := range values[n] {
+			labelValuesBuf = append(labelValuesBuf, v)
+		}
+		sort.Strings(labelValuesBuf)
+
+		if mappedName, ok := symbolMap[n]; ok {
```
naivewong: I think this lookup should be done after `posts, err := indexr.Postings(n, v)`. Otherwise, the `n` will be the new symbol and `posts, err := indexr.Postings(n, v)` will return an empty postings without any error (because the error will be nil when the name/value lookup fails).

codesome: Yes, you are right. Thanks for pointing out!

csmarchbanks:
Out of curiosity, why is this part of `promtool` and not the `tsdb` tool? Is there any criteria for putting new features in one tool or the other?

cmd/promtool/main.go
```diff
@@ -641,3 +659,185 @@ func (j *jsonPrinter) printLabelValues(v model.LabelValues) {
...
+			posts, err := indexr.Postings(n, v)
+			if err != nil {
+				return err
+			}
+
+			if mappedVal, ok := symbolMap[v]; ok {
+				if err := indexw.WritePostings(n, mappedVal, posts); err != nil {
+					return err
+				}
+			} else {
```
bwplotka: You can almost always avoid `else` by just returning on `!ok` immdiately.

cmd/promtool/main.go
```diff
@@ -641,3 +659,185 @@ func (j *jsonPrinter) printLabelValues(v model.LabelValues) {
 	//nolint:errcheck
 	json.NewEncoder(os.Stdout).Encode(v)
 }
+
+// ObfuscateIndexFile renames the symbols of the given index file
+// and re-writes the index file at given output path.
+func ObfuscateIndexFile(inputFilePath, outputFilePath string) int {
```
bwplotka: Can we return `error` and not invent yet another way of returning errors? (:  Also can we put this in separate package, so we can use it in Thanos/Cortex/XYZ at some point programmatically? (:

codesome:
> Out of curiosity, why is this part of `promtool` and not the `tsdb` tool?

I had actually not thought about it. But as @bwplotka said, after TSDB is merged with this repo, all would be in `promtool`? Merging tsdb with prometheus is in my todo list (given no-one is planning to do it).

cmd/promtool/main.go
codesome: Sounds fine. I can do that.

roidelapluie:
Hello @codesome can you look at this?

codesome:
I don't think I can work further on this. Should I close this and open an issue so that someone else can pick it up?

PR #5780, opened 2019-07-18
Add k8s 1.15.1 as a vendor dependency.

tariq1890:
~In the interest of getting with the times, we should be using `networking/v1beta1` for ingress and not `extensions/v1beta1`.~

Updating the k8s vendor to v1.15.1


brancz:
I think @LiliC looked into this before, and identified that this API was only introduced recently, so this change would impact our backward compatibility. My understanding is that the Ingress API as part of the extensions API is sticking around until 1.18.

tariq1890:
Yes you are right. It won't be removed until k8s v1.18 .

brancz:
Then we should wait until the end of 1.17 time-frame to upgrade to the new API.

tariq1890:
I'll undo the api changes and just vendor in the latest k8s library then.

tariq1890:
@simonpasquier Requesting your review as well.

PR #5782, opened 2019-07-19
Cloudmap discovery

adjkilgore:


simonpasquier:
Thanks for suggesting this change. Unfortunately, we are currently not accepting new integrations.

adjkilgore:
No problem - we forked this anyway, and will maintain ourselves.  This PR was created accidentally as I was merging into our fork.

It’s a great product, and we understand the need to be selective on how you approach plugins / integrations

PR #5787, opened 2019-07-19
Add metrics for max/min/desired shards to queue manager.

cstyan:
Add metrics for max shards, min shards, and desired shards to queue manager. This will help remote write users determine whether they to raise the max shards in their config.

@csmarchbanks 
cc @tomwilkie @gouthamve

### DIFF
--- a/documentation/prometheus-mixin/alerts.libsonnet
+++ b/documentation/prometheus-mixin/alerts.libsonnet
@@ -226,6 +226,27 @@
             },
           },
           {
+            alert: 'PrometheusRemoteWriteDesiredShards',
+            expr: |||
+              # Without max_over_time, failed scrapes could create false negatives, see
+              # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+              (
+                max_over_time(prometheus_remote_storage_shards_desired{%(prometheusSelector)s}[5m])
+              > on(job, instance) group_right
+                max_over_time(prometheus_remote_storage_shards_max{%(prometheusSelector)s}[5m])
+              )
+              == 1
+            ||| % $._config,
+            'for': '15m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              summary: 'Prometheus remote write desired shards calculation wants to run more than configured max shards.',
+              description: 'Prometheus %(prometheusName)s remote write is {{ printf "%%.1f" $value }}s behind for queue {{$labels.queue}}.' % $._config,
+            },
+          },
+          {
             alert: 'PrometheusRuleFailures',
             expr: |||
               increase(prometheus_rule_evaluation_failures_total{%(prometheusSelector)s}[5m]) > 0
--- a/documentation/prometheus-mixin/alerts.libsonnet
+++ b/documentation/prometheus-mixin/alerts.libsonnet
@@ -235,7 +235,6 @@
               > on(job, instance) group_right
                 max_over_time(prometheus_remote_storage_shards_max{%(prometheusSelector)s}[5m])
               )
-              == 1
             ||| % $._config,
             'for': '15m',
             labels: {
@@ -243,7 +242,7 @@
             },
             annotations: {
               summary: 'Prometheus remote write desired shards calculation wants to run more than configured max shards.',
-              description: 'Prometheus %(prometheusName)s remote write is {{ printf "%%.1f" $value }}s behind for queue {{$labels.queue}}.' % $._config,
+              description: 'Prometheus %(prometheusName)s remote write desired shards calculation wants to run {{ printf $value }} shards, which is more than the max of {{ printf `prometheus_remote_storage_shards_max{instance="%%s",%(prometheusSelector)s}` $labels.instance | query | first | value }}.' % $._config,
             },
           },
           {
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -143,6 +143,33 @@ var (
 		},
 		[]string{queue},
 	)
+	maxNumShards = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "shards_max",
+			Help:      "The maximum number of shards that the queue is allowed to run.",
+		},
+		[]string{queue},
+	)
+	minNumShards = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "shards_min",
+			Help:      "The minimum number of shards that the queue is allowed to run.",
+		},
+		[]string{queue},
+	)
+	desiredNumShards = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "shards_desired",
+			Help:      "The number of shards that the queues shard calculation wants to run based on the rate of samples in vs. samples out.",
+		},
+		[]string{queue},
+	)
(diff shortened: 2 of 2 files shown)


csmarchbanks:
I don't think this adds much as a Debug log since there is detailed information about the desired shards calculation just a couple lines up. Is it too noisy above the debug level?

cstyan:
@csmarchbanks yeah I suppose this would make sense as `warning` level as well, especially since the existing debug logging already takes place prior to the check and assignment of `desiredShards = maxShards`.

csmarchbanks:
I like the idea of using metrics rather than a warning log. The issue only needs to be brought to an operator's attention if desired shards is greater than max shards for a significant amount of time, so I think individual warnings could be a bit misleading.

cstyan:
I still think logging is useful to more easily see how many shards the calculation wanted to run. I'll add a metric to count the # of times the calculation has been > maxShards.

beorn7:
Is it important how often the calculation came up with that result? If you had a gauge "desired shards" (updated after each shard calculation) and a fixed-value metric with the configured "max shards", wouldn't that be sufficient? You could still look at it over time.

(I'm not a remote-write expert. My suggestion might be completely off.)

cstyan:
If we were to alert on a metric for this it would likely be something like if "desired shards is greater than max shards" over some period of time. For example we only alert if remote write is behind by a minute or two for more than 15 minutes.

Sounds like a gauge would be sufficient. I don't have a preference either way.

beorn7:
I would also add a metric for the configured max shards, even if that seems redundant. In some setup, it's hard to get this value from config management, but you want to have it easily available in alerts or even dashboards.

cstyan:
Added metrics, removed the logging.

storage/remote/queue_manager.go
```diff
@@ -143,6 +143,33 @@ var (
 		},
 		[]string{queue},
 	)
+	maxNumShards = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "max_shards",
```
brian-brazil: shards_max is the usual naming

storage/remote/queue_manager.go
```diff
@@ -143,6 +143,33 @@ var (
...
+			Help:      "The maxiumum number of shards that the queue is allowed to run.",
```
beorn7: maxiumum → maximum

storage/remote/queue_manager.go
beorn7: `min_shards` and `desired_shards` has to be changed accordingly.

storage/remote/seriesTest/main.go
```diff
@@ -0,0 +1,111 @@
+// Copyright 2019 The Prometheus Authors
```
beorn7: Is this file meant to be part of the commit?

storage/remote/queue_manager.go
cstyan: didn't know that, thanks!

storage/remote/queue_manager.go
```diff
@@ -293,10 +323,16 @@ func (t *QueueManager) Start() {
...
+	t.maxNumShards = maxNumShards.WithLabelValues(name)
+	t.minNumShards = minNumShards.WithLabelValues(name)
+	t.desiredNumShards = desiredNumShards.WithLabelValues(name)
 
 	// Initialise some metrics.
 	t.shardCapacity.Set(float64(t.cfg.Capacity))
 	t.pendingSamplesMetric.Set(0)
+	t.maxNumShards.Set(float64(t.cfg.MaxShards))
+	t.minNumShards.Set(float64(t.cfg.MinShards))
+	t.desiredNumShards.Set(0)
```
csmarchbanks: Should desired shards be initialized to `t.cfg.MinShards` since that is where sharding starts from?

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -225,6 +225,27 @@
...
+            alert: 'PrometheusRemoteWriteDesiredShards',
+            expr: |||
+              # Without max_over_time, failed scrapes could create false negatives, see
+              # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+              (
+                max_over_time(prometheus_remote_storage_shards_desired{%(prometheusSelector)s}[5m])
+              > on(job, instance) group_right
+                prometheus_remote_storage_shards_max{%(prometheusSelector)s})
+              )
+              == 1
```
gouthamve: `== 1` requires a bool operator no? I don't think a simple `>` would work here.

brian-brazil: Only for scalar/scalar comparisons. > should work in here. the LHS however also needs a max_over_time.

cstyan: :+1: added

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -225,6 +225,27 @@
...
+            ||| % $._config,
+            'for': '15m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              summary: 'Prometheus remote write desired shards calculation wants to run more than configured max shards.',
+              description: 'Prometheus %(prometheusName)s remote write is {{ printf "%%.1f" $value }}s behind for queue {{$labels.queue}}.' % $._config,
```
beorn7: This description is from a different alert.

cstyan:
@beorn7 PTAL at the alert. I tested it locally by hardcoding the return value of `calculateDesiredShards` and setting `max_shards` to 1 in my config, then checking the annotations on the alerts page. It looks correct to me.

documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -225,6 +225,26 @@
...
+                max_over_time(prometheus_remote_storage_shards_max{%(prometheusSelector)s}[5m])
+              )
+            ||| % $._config,
+            'for': '15m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              summary: 'Prometheus remote write desired shards calculation wants to run more than configured max shards.',
+              description: 'Prometheus %(prometheusName)s remote write desired shards calculation wants to run {{ printf $value }} shards, which is more than the max of {{ printf `prometheus_remote_storage_shards_max{%(prometheusSelector)s}` | query | first | value }}.' % $._config,
```
beorn7: The query will just pick the first result, not necessarily the one for the Prometheus server affected. I believe it has to be written in the following way:

```yaml
              description: 'Prometheus %(prometheusName)s remote write desired shards calculation wants to run {{ printf $value }} shards, which is more than the max of {{ printf `prometheus_remote_storage_shards_max{instance="%%s",%(prometheusSelector)s}` $labels.instance | query | first | value }}.' % $._config,
```

cstyan: Hmmm okay, is that not what `prometheusSelectors` is supposed to be doing? I assumed those would contain the label selectors for the instance that triggered the alert.

beorn7: `prometheusSelector` selects _all_ Prometheus servers to monitor.

beorn7: In this case, it would intersect with the alerts firing, i.e. if only one Prometheus server fires this alerts, all is good, but as soon as more than one of the monitored Prometheus servers are affected by this alert, you'll get multiple results in the query.

cstyan: :+1: got it, thank you!

Issue #1315, opened 2016-01-14
Add option to log slow queries and recording rules


AlekSi:
Having something like [PostgreSQL's log_min_duration_statement](http://www.postgresql.org/docs/current/static/runtime-config-logging.html) would be useful to debug performance problems.

spinus:
+1 for that feature

grobie:
I was thinking a few times that it would be great to collect detailed query information, like how many chunks were necessary to compute the result and how many had to be loaded from disk.

fabxc:
+1 for the more detailed reporting – especially those feeding into metrics.
We are collecting how long various stages of a query evaluation take but
are not making use of it.

beorn7:
We used to spit out a lot of query information on debug log level. But that's gone now.
I'm currently trying to hunt down an abusive query, but it's not possible anymore.

Just bringing back individual query logging on debug level would already be quite helpful.
If we could also create some useful metrics (mysqld_exporter style), even better.

noushi:
+1

beorn7:
Suggestion for logging:

- On INFO level, log the API accesses, just the URL, response time, status code. (This could nicely feed into log analysis. An ELK stack and Prometheus could then look after each other... ;)
- On DEBUG level, bring back (and improve) the comprehensive analysis of query time break up.

Rationale: API calls are not _that_ frequent. And whoever doesn't like to log them, goes to WARN level in logging. On the other hand, we want the comprehensive analysis on DEBUG level. So we need a level for the "classical" query logging you might want to routinely feed into log analysis. If it's not INFO, what would it be then?

brian-brazil:
> On INFO level, log the API accesses, just the URL, response time, status code. (This could nicely feed into log analysis. An ELK stack and Prometheus could then look after each other... ;)

This would be way too spammy. API calls are frequent, a few dashboards auto-refreshing every minute and you could easily be hitting 1 qps. Application logging should aim for no more than a few log lines per minute, and even then that's high.

If these are all to be logged, they belong in some other logging mechanism more suited for request logs - not the application logs.

beorn7:
WARN level could be the default. What else, in a completely re-thought logging, would be on INFO level? Perhaps INFO is already the place for "spam" and should only be activated explicitly by those interested. (Here at SC, I'd love to have API logging on all our Prometheus servers. There are way bigger log producers around...)

A separate logging mechanism would be a solution, too, but requires plenty of configuration.

[...]

PR #5794, opened 2019-07-23
Added query logging for prometheus. Issue #1315

You-NeverKnow:
Flags added:
1) active.queries.filepath: Filename where queries will be recorded
2) active.queries.filesize: Size of the file where queries will be recorded.

Functionality added:
All active queries are now logged in a file. If prometheus crashes unexpectedly, these queries are also printed out on stdout in the rerun.
[Here's a flow chart depicting how its implemented](https://imgur.com/a/rq3r81I)

Queries are written concurrently to an mmaped file, and removed once they are done. Their positions in the file are reused. They are written in json format. However, due to dynamic nature of application, the json has an extra comma after the last query, and is missing an ending ']'. There may also null bytes in the tail of file.

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -358,7 +358,7 @@ func main() {
 			MaxConcurrent:      cfg.queryConcurrency,
 			MaxSamples:         cfg.queryMaxSamples,
 			Timeout:            time.Duration(cfg.queryTimeout),
-			ActiveQueryTracker: promql.NewActiveQueryTracker(cfg.queryConcurrency, log.With(logger, "component", "activeQueryTracker")),
+			ActiveQueryTracker: promql.NewActiveQueryTracker(cfg.localStoragePath, cfg.queryConcurrency, log.With(logger, "component", "activeQueryTracker")),
 		}
 
 		queryEngine = promql.NewEngine(opts)
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -119,9 +119,6 @@ func main() {
 		corsRegexString string
 
 		promlogConfig promlog.Config
-
-		activeQueryLogFilename string
-		activeQueryLogFileSize int
 	}{
 		notifier: notifier.Options{
 			Registerer: prometheus.DefaultRegisterer,
@@ -249,12 +246,6 @@ func main() {
 	a.Flag("query.max-samples", "Maximum number of samples a single query can load into memory. Note that queries will fail if they try to load more samples than this into memory, so this also limits the number of samples a query can return.").
 		Default("50000000").IntVar(&cfg.queryMaxSamples)
 
-	a.Flag("active.queries.filepath", "File that will log active queries").
-		Default("queries.log").StringVar(&cfg.activeQueryLogFilename)
-
-	a.Flag("active.queries.filesize", "Filesize of active query log").
-		Default("20000").IntVar(&cfg.activeQueryLogFileSize)
-
 	promlogflag.AddFlags(a, &cfg.promlogConfig)
 
 	_, err := a.Parse(os.Args[1:])
@@ -347,11 +338,6 @@ func main() {
 		fanoutStorage = storage.NewFanout(logger, localStorage, remoteStorage)
 	)
 
-	queryLogger := promql.QueryLogger{}
-	if cfg.activeQueryLogFilename != "" {
-		queryLogger = promql.NewQueryLogger(cfg.activeQueryLogFilename, cfg.activeQueryLogFileSize, log.With(logger, "component", "queryLogger"))
-	}
-
 	var (
 		ctxWeb, cancelWeb = context.WithCancel(context.Background())
 		ctxRule           = context.Background()
@@ -367,12 +353,12 @@ func main() {
 		scrapeManager = scrape.NewManager(log.With(logger, "component", "scrape manager"), fanoutStorage)
 
 		opts = promql.EngineOpts{
-			Logger:        log.With(logger, "component", "query engine"),
-			Reg:           prometheus.DefaultRegisterer,
-			MaxConcurrent: cfg.queryConcurrency,
-			MaxSamples:    cfg.queryMaxSamples,
-			Timeout:       time.Duration(cfg.queryTimeout),
-			QueryLogger:   queryLogger,
+			Logger:             log.With(logger, "component", "query engine"),
+			Reg:                prometheus.DefaultRegisterer,
+			MaxConcurrent:      cfg.queryConcurrency,
+			MaxSamples:         cfg.queryMaxSamples,
+			Timeout:            time.Duration(cfg.queryTimeout),
+			ActiveQueryTracker: promql.NewActiveQueryTracker(cfg.queryConcurrency, log.With(logger, "component", "activeQueryTracker")),
 		}
 
 		queryEngine = promql.NewEngine(opts)
@@ -422,7 +408,7 @@ func main() {
 		cfg.web.Flags[f.Name] = f.Value.String()
 	}
 
-	// Depends on cfg.web.ScrapeManager so needs to be after cfg.web.ScrapeManager = scrapeManager
+	// Depends on cfg.web.ScrapeManager so needs to be after cfg.web.ScrapeManager = scrapeManager.
 	webHandler := web.New(log.With(logger, "component", "web"), &cfg.web)
 
 	// Monitor outgoing connections on default transport with conntrack.
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -119,6 +119,9 @@ func main() {
 		corsRegexString string
 
 		promlogConfig promlog.Config
+
+		activeQueryLogFilename string
+		activeQueryLogFileSize int
 	}{
 		notifier: notifier.Options{
 			Registerer: prometheus.DefaultRegisterer,
@@ -246,6 +249,12 @@ func main() {
 	a.Flag("query.max-samples", "Maximum number of samples a single query can load into memory. Note that queries will fail if they try to load more samples than this into memory, so this also limits the number of samples a query can return.").
 		Default("50000000").IntVar(&cfg.queryMaxSamples)
 
+	a.Flag("active.queries.filepath", "File that will log active queries").
+		Default("queries.log").StringVar(&cfg.activeQueryLogFilename)
+
+	a.Flag("active.queries.filesize", "Filesize of active query log").
+		Default("20000").IntVar(&cfg.activeQueryLogFileSize)
+
 	promlogflag.AddFlags(a, &cfg.promlogConfig)
 
 	_, err := a.Parse(os.Args[1:])
@@ -338,6 +347,11 @@ func main() {
 		fanoutStorage = storage.NewFanout(logger, localStorage, remoteStorage)
 	)
 
+	queryLogger := promql.QueryLogger{}
+	if cfg.activeQueryLogFilename != "" {
+		queryLogger = promql.NewQueryLogger(cfg.activeQueryLogFilename, cfg.activeQueryLogFileSize, log.With(logger, "component", "queryLogger"))
+	}
+
 	var (
 		ctxWeb, cancelWeb = context.WithCancel(context.Background())
 		ctxRule           = context.Background()
@@ -358,7 +372,9 @@ func main() {
 			MaxConcurrent: cfg.queryConcurrency,
 			MaxSamples:    cfg.queryMaxSamples,
 			Timeout:       time.Duration(cfg.queryTimeout),
+			QueryLogger:   queryLogger,
 		}
+
 		queryEngine = promql.NewEngine(opts)
 
 		ruleManager = rules.NewManager(&rules.ManagerOptions{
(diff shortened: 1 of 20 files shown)


cmd/prometheus/main.go
```diff
@@ -246,6 +249,12 @@ func main() {
 	a.Flag("query.max-samples", "Maximum number of samples a single query can load into memory. Note that queries will fail if they try to load more samples than this into memory, so this also limits the number of samples a query can return.").
 		Default("50000000").IntVar(&cfg.queryMaxSamples)
 
+	a.Flag("active.queries.filepath", "File that will log active queries").
```
brian-brazil: I don't think either of these need to be configurable.

promql/query_logger.go
```diff
@@ -0,0 +1,217 @@
...
+	"fmt"
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"os"
+	"strings"
+	"syscall"
+	"time"
+)
+
+type QueryLogger struct {
```
brian-brazil: This isn't logging, it's current queries.

promql/query_logger.go
```diff
@@ -0,0 +1,217 @@
...
+func (logger QueryLogger) generateLastIndex(maxSize int) {
+	for i := 1; i <= maxSize-entrySize; i += entrySize {
+		logger.lastIndex <- i
+	}
+	close(logger.lastIndex)
+}
+
+func (logger QueryLogger) LogQuery(query string, queryCompleted <-chan struct{}) {
+	defer func() {
+		if r := recover(); r != nil {
```
brian-brazil: How can this happen?

promql/query_logger.go
```diff
@@ -0,0 +1,217 @@
...
+
+	err, fileAsBytes := getMMapedFile(filename, filesize, logger)
+	if err != nil {
+		return QueryLogger{}
+	}
+
+	copy(fileAsBytes, "[")
+	queryLogger = QueryLogger{
+		mmapedFile:   fileAsBytes,
+		getNextIndex: make(chan int, (1<<16)-1),
```
brian-brazil: Why 32k? We know the maximum number of running queries from flags.

promql/query_logger.go
You-NeverKnow: It can't. I kept it as fail safe mechanism to keep prometheus from crashing in case something changes in this code in the future. Should I remove it?

cmd/prometheus/main.go
You-NeverKnow: What would you like to be the defaults? I picked both of them at random: 100 queries at a time, and queries.log as filename.

promql/query_logger.go
You-NeverKnow: Yeah.. I'll change that

promql/query_logger.go
You-NeverKnow: Oh didn't think of that. I'll fix it.

promql/query_logger.go
brian-brazil: Yes, remove it. If there's a bug we'll deal with it when it's discovered.

cmd/prometheus/main.go
brian-brazil: That's as good as anything.

promql/engine.go
```diff
@@ -176,7 +176,21 @@ func (q *query) Exec(ctx context.Context) *Result {
 		span.SetTag(queryTag, q.stmt.String())
 	}
 
+	// Log query in active log
```
brian-brazil: Comments should end in full stops

promql/engine.go
```diff
@@ -176,7 +176,21 @@ func (q *query) Exec(ctx context.Context) *Result {
...
+	queryCompleted := make(chan struct{})
+
+	go func() {
```
brian-brazil: Why is this in a goroutine?

promql/query_logger.go
```diff
@@ -0,0 +1,217 @@
...
+type QueryLogger struct {
+	mmapedFile   []byte
+	getNextIndex chan int
+	lastIndex    chan int
+	logger       log.Logger
+	Ok           bool
+}
+
+const (
+	entrySize int = 200
```
brian-brazil: I'd be generous here, there won't be that many queries running at once.

promql/query_logger.go
```diff
@@ -0,0 +1,217 @@
...
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func newJsonEntry(query string, timestamp int64, logger log.Logger) []byte {
+	// 35 bytes is the size of json entry for empty string. Including a comma for every entry, max query size is entrySize-36
+	entry := Entry{query[:min(len(query), entrySize-36)], timestamp}
```
brian-brazil: You're measuring characters here, not bytes

promql/query_logger.go
```diff
@@ -0,0 +1,217 @@
...
+	copy(fileAsBytes, "[")
+	queryLogger = QueryLogger{
+		mmapedFile:   fileAsBytes,
+		getNextIndex: make(chan int, (1<<16)-1),
+		lastIndex:    make(chan int),
+		logger:       logger,
+		Ok:           true,
+	}
+
+	go queryLogger.generateLastIndex(filesize)
```
brian-brazil: This doesn't need to be a goroutine

promql/engine.go
You-NeverKnow: To decouple active query tracking from query execution, and make it faster since it would be run in parallel, if possible. I haven't done any measurements vs. its sequential counterpart though. I'll check which one's faster on my side, and let you know.

brian-brazil: Keep it simple, the whole query execution will use far more than this.

promql/query_logger.go
You-NeverKnow: The smallest entry would be an empty string query, and its size after json.Marshal() turned out to be 35. Since I also want a comma in an entry, and comma is part of UTF-8, (entrySize-36) would be the maximum size of the query, no? Alternatively, I could introduce another variable that repeats this procedure to compute smallest query size in bytes.

brian-brazil: That's still mixing bytes and characters though.

You-NeverKnow: Yes. I spotted my mistake. Using this now: 
```
func trimStringSize(str string, size int) string {
	bytesStr := []byte(str)
	return string(bytesStr[:min(len(str), size)])
}
```

promql/bench_test.go
```diff
@@ -16,6 +16,8 @@ package promql
 import (
 	"context"
 	"fmt"
+	"github.com/go-kit/kit/log"
```
brian-brazil: This should be in a separate group, it's not part of Go itself

promql/engine_test.go
```diff
@@ -87,12 +90,14 @@ func TestQueryConcurrency(t *testing.T) {
...
-		Reg:           nil,
-		MaxConcurrent: 20,
-		MaxSamples:    10,
-		Timeout:       5 * time.Millisecond,
+		Logger:             nil,
+		Reg:                nil,
+		MaxConcurrent:      20,
+		MaxSamples:         10,
+		Timeout:            5 * time.Millisecond,
+		ActiveQueryTracker: NewActiveQueryTracker(20, logger),
```
brian-brazil: Hmm, maybe this should be allowed to be nil?

promql/query_logger.go
```diff
@@ -0,0 +1,208 @@
+// Copyright 2013 The Prometheus Authors
```
brian-brazil: 2019

promql/query_logger.go
```diff
@@ -0,0 +1,208 @@
...
+			level.Error(logger).Log("msg", "Failed to read query log file", "err", err)
+			return
+		}
+
+		queriesExist, queries := parseBrokenJson(brokenJson, logger)
+		if !queriesExist {
+			return
+		}
+		level.Info(logger).Log("msg", "These queries didn't finish in prometheus' last run:")
+		fmt.Println(queries)
```
brian-brazil: log these normally

promql/query_logger.go
```diff
@@ -0,0 +1,208 @@
...
+	if err != nil {
+		level.Error(logger).Log("msg", "Error mmapping file", "file", filename, "err", err)
+		return err, []byte{}
+	}
+
+	return err, fileAsBytes
+}
+
+func NewActiveQueryTracker(maxQueries int, logger log.Logger) (activeQueryTracker ActiveQueryTracker) {
+	defer func() {
```
brian-brazil: This should go

promql/query_logger.go
```diff
@@ -0,0 +1,208 @@
...
+			level.Error(logger).Log("msg", "Panic while creating ActiveQueryTracker", "err", r)
+			activeQueryTracker = ActiveQueryTracker{}
+		}
+	}()
+
+	filename, filesize := "queries.active", 1+maxQueries*entrySize
+	logUnfinishedQueries(filename, filesize, logger)
+
+	err, fileAsBytes := getMMapedFile(filename, filesize, logger)
+	if err != nil {
```
brian-brazil: We should error out here

promql/query_logger.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	activeQueryTracker.generateIndices(filesize)
+
+	return activeQueryTracker
+}
+
+type Entry struct {
+	Query     string `json:"query"`
+	Timestamp int64  `json:"timestamp"`
```
brian-brazil: The unit should be mentioned here

promql/query_logger.go
```diff
@@ -0,0 +1,208 @@
...
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func trimStringByBytes(str string, size int) string {
+	bytesStr := []byte(str)
+	return string(bytesStr[:min(len(bytesStr), size)])
```
brian-brazil: How does this handle a character split in the middle of its bytes? https://stackoverflow.com/questions/52323044/split-string-based-on-byte-length-in-golang

promql/query_logger.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+func newJsonEntry(query string, timestamp int64, logger log.Logger) []byte {
+	// 35 bytes is the size of json entry for empty string.
```
brian-brazil: Wouldn't that vary based on the timestamp?

promql/query_logger.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/go-kit/kit/log/level"
+	"os"
+	"strings"
+	"syscall"
+	"time"
+)
+
+type ActiveQueryTracker struct {
+	mmapedFile   []byte
+	getNextIndex chan int
```
brian-brazil: Shouldn't this channel be enough?

You-NeverKnow: yep!

promql/query_logger.go
You-NeverKnow: I didn't know about that. I assumed all unix timestamps would be 10 characters long. I'll add a function to dynamically compute max byte length possible for the string.

promql/query_logger.go
You-NeverKnow: It doesn't. I'll fix it

brian-brazil:
I think you forgot to push.

promql/query_logger.go
You-NeverKnow: Do you mean something like this?
```
level.Info(logger).Log("msg", "Some queries didn't finish in prometheus' last run", "queries", queries)
```

Because I am difficulty printing a formatted json with logger. Instead of true newlines and indents, I am getting a bunch of \n and \t mixed with rest of the json in my shell output.

You-NeverKnow: Done!

promql/query_logger.go
```diff
@@ -0,0 +1,177 @@
...
+
+	err, fileAsBytes := getMMapedFile(filename, filesize, logger)
+	if err != nil {
+		panic("Unable to create mmap-ed active query log")
+	}
+
+	copy(fileAsBytes, "[")
+	activeQueryTracker := ActiveQueryTracker{
+		mmapedFile:   fileAsBytes,
+		getNextIndex: make(chan int, maxQueries+(filesize-1)/entrySize),
```
brian-brazil: Shouldn't this be just maxQueries?

You-NeverKnow: Yep!

brian-brazil:
The file should be under the data directory, as we know we can write there, rather than pwd.

You-NeverKnow:
> The file should be under the data directory, as we know we can write there, rather than pwd.

Should I put it under its own folder, like _data/active_queries/queries.active_, or just _data/queries.active_ ?

brian-brazil:
I'd put it directly under.

You-NeverKnow:
Could you run benchmark on it?  
And I could rebase it down to one commit if this is satisfactory.

brian-brazil:
Is there some particular reason you want a benchmark? I would expect no noticeable change from this PR, and we'd catch it at release if it was an issue.
GitHub lets me do the squash on merge these days.

You-NeverKnow:
No particular reason. Just wanted to do it as a sanity check.

brian-brazil:
Thanks!

PR #5796, opened 2019-07-24
Fix typo

izeye:
This PR fixes a typo.


brian-brazil:
Hmm, to make it clear can you add the scrape_configs at the top of the example?

Can you also add the DCO?

izeye:
@brian-brazil Thanks for the quick feedback! I've changed as you suggested.

brian-brazil:
Thanks!

PR #5807, opened 2019-07-29
Fix log config setup

fser:
The return value of Set() was not checked.
Hence, the typo ("al" instead of "af") wasn't catched.


brian-brazil:
Thanks!

PR #5812, opened 2019-07-30
Use GOPROXY to speed up builds

sylr:
Signed-off-by: Sylvain Rabot <sylvain@abstraction.fr>


simonpasquier:
I'm not sure we need to add this now since Go 1.13 is almost there and it will use `https://proxy.golang.org/` as the default `GOPROXY` value (https://tip.golang.org/doc/go1.13#introduction).
In other news, you'd have to update your CircleCI settings as your pull requests bypass the CI currently. See https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization-, I wish they had a better story than this...

PR #5821, opened 2019-08-01
removed docreference to meta label that doesnt exist

cuotos:
In the configuration documentation page its stated the kubernetes_sd (service role) provides 

__meta_kubernetes_service_port_number: Number of the service port for the target.

This is not the case. Only name and protocol are provided by the kubernetes_sd.


brian-brazil:
Thanks!

PR #5823, opened 2019-08-02
remove the wrongly adding when errSampleLimit in scrape/scrape.go

naivewong:
https://github.com/prometheus/prometheus/blob/ead0933dd9065602e234ae907eee17a6dd208601/scrape/target.go#L268-L273
When `errSampleLimit` is returned, the `Add` or `AddFast` of `Appender` is not executed. I think the `added++` should be removed.


codesome:
According to the comment
```
// Keep on parsing output if we hit the limit, so we report the correct
// total number of samples scraped.
```

PR #5825, opened 2019-08-02
prometheus-mixin: remove PrometheusTSDBWALCorruptions

simonpasquier:
This is for discussion as I'm not sure what's the proper way to handle. Maybe we need to reconsider the metric itself?

The counter is only increased when tsdb.Open() is called which
Prometheus does only once in its lifetime (when it initializes). If the
corruption can't be recovered, tsdb.Open() returns an error and
Prometheus exits. Hence the metric is either 0 (no corruption) or 1
(corruption detected and repaired).

The metric name is 'prometheus_tsdb_wal_corruptions_total', not
'tsdb_wal_corruptions_total'.

### DIFF
--- a/documentation/prometheus-mixin/alerts.libsonnet
+++ b/documentation/prometheus-mixin/alerts.libsonnet
@@ -125,20 +125,6 @@
             },
           },
           {
-            alert: 'PrometheusTSDBWALCorruptions',
-            expr: |||
-              increase(tsdb_wal_corruptions_total{%(prometheusSelector)s}[3h]) > 0
-            ||| % $._config,
-            'for': '4h',
-            labels: {
-              severity: 'warning',
-            },
-            annotations: {
-              summary: 'Prometheus is detecting WAL corruptions.',
-              description: 'Prometheus %(prometheusName)s has detected {{$value | humanize}} corruptions of the write-ahead log (WAL) over the last 3h.' % $._config,
-            },
-          },
-          {
             alert: 'PrometheusNotIngestingSamples',
             expr: |||
               rate(prometheus_tsdb_head_samples_appended_total{%(prometheusSelector)s}[5m]) <= 0


brian-brazil:
Can't this also happen when creating checkpoints?

I'd also not have an alert that turns itself off, either this is a problem that requires intelligent human intervention or it isn't.

simonpasquier:
> Can't this also happen when creating checkpoints?

Looking at the code, I'd say no but I'll let @krasi-georgiev chime in. Checkpoint operations have their own metrics: `prometheus_tsdb_checkpoint_creations_failed_total` and `prometheus_tsdb_checkpoint_creations_total`.

> I'd also not have an alert that turns itself off, either this is a problem that requires intelligent human intervention or it isn't.

Fair enough. As I wrote in the description, I'm not sure we need the metric/alert since if corruption happens, it is either repaired automatically or Prometheus can't restart. I can see that someone wants to know about the "event" and investigate if data has been lost but the only way to resolve the alert would be to restart the server.

brian-brazil:
That a case for logging, not alerting.

krasi-georgiev:
I just double checked and we even have a bug as this metric is registered twice
https://github.com/prometheus/tsdb/blob/b40cc43958a563aa06a73b2c8d9090e66109ad1b/wal.go#L72-L75

https://github.com/prometheus/tsdb/blob/6f9bbc7253cb1cd550eeba2c3cc4c3015abbc91c/head.go#L169-L171

and yes when I was working on these parts I also thought it is weird and not very useful metric to have so can be removed I think.

krasi-georgiev:
@brancz mentioned in a call that this metric might be useful to know that a corruption happened and maybe can keep a copy of the corrupted data for potential repair, but from previous discussions I think the overall impression is that even if some data can be recovered it would make it impossible to use because of missing data points, series etc. Or in other words querying this recovered data would lead to misleading results.

brancz:
I meant that it'd be good to keep the corrupt data around for us to analyze if we potentially could have repaired more/better. I think the metric is still useful to have, but the alerting rule is questionable.

simonpasquier:
Given the current feedback, I've amended the PR to remove the alert.

beorn7:
From the discussion so far I conclude that everyone agrees removing the alert is an improvement.
Therefore I'll merge this.

Issue #5731, opened 2019-07-03
Check for duplicate label names in remote read response


xinyu7git:
## Bug Report
**What did you do?**
Using remote_read and remote_write config, but Prometheus did not merge data from local storage and remote storage , i get two same points. so many function can't be used.

![image](https://user-images.githubusercontent.com/3701997/60559780-23973e80-9d80-11e9-86df-98072060bdd2.png)

**What did you expect to see?**
Prometheus should merge same points to one point.

**What did you see instead? Under which circumstances?**

**Environment**
   Start Command: /data/prometheus/prometheus --config.file=/data/prometheus/prometheus.yml --storage.tsdb.retention.time=2h --storage.tsdb.retention.size=2MB --log.level=debug

* System information:

	insert output of `uname -srm` here

         Linux 5.1.5-1.el7.elrepo.x86_64 x86_64
* Prometheus version:

	insert output of `prometheus --version` here
        `[root@10-0-0-14 prometheus]# ./prometheus --version
          prometheus, version 2.10.0 (branch: HEAD, revision: d20e84d0fb64aff2f62a977adc8cfb656da4e286)
          build user:       root@a49185acd9b0
          build date:       20190525-12:28:13
          go version:       go1.12.5`

* Alertmanager version:

	insert output of `alertmanager --version` here (if relevant to the issue)

* Prometheus configuration file:
```
insert configuration here
# my global config
global:
  scrape_interval:     15s # Set the scrape interval to every 15 seconds. Default is every 1 minute.
  evaluation_interval: 15s # Evaluate rules every 15 seconds. The default is every 1 minute.
  # scrape_timeout is set to the global default (10s).

# Alertmanager configuration
alerting:
  alertmanagers:
  - static_configs:
    - targets:
      # - alertmanager:9093

# Load rules once and periodically evaluate them according to the global 'evaluation_interval'.
rule_files:
  # - "first_rules.yml"
  # - "second_rules.yml"

# A scrape configuration containing exactly one endpoint to scrape:
# Here it's Prometheus itself.
scrape_configs:
  # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
  - job_name: 'prometheus'

    # metrics_path defaults to '/metrics'
    # scheme defaults to 'http'.

    static_configs:
    - targets: ['10.0.0.14:9090']

  - job_name: 'node_exporter'
    scrape_interval: 10s
    static_configs:
    - targets: ['localhost:9100']

[...]

PR #5829, opened 2019-08-04
Check for duplicate label names in remote read

vmartsynovskyy:
In #5731 a bug was discovered where there would sometimes be a timeseries from local storage and a timeseries from a remoteRead that were duplicates of eachother. This happened in cases where a timeseries coming from a remoteRead had duplicate label names. This caused `mergeSeriesSet` to not merge the remoteRead timeseries with the local storage timeseries.

This PR fixes this issue by adding a function to check for and remove duplicate label names coming from a remoteRead. There are also tests that ensure this is really happening.

CC @brian-brazil

### DIFF
--- a/storage/remote/codec.go
+++ b/storage/remote/codec.go
@@ -280,8 +280,7 @@ func (c *concreteSeriesIterator) Err() error {
 // validateLabelsAndMetricName validates the label names/values and metric names returned from remote read,
 // also making sure that there are no labels with duplicate names
 func validateLabelsAndMetricName(ls labels.Labels) error {
-	var prevLabelName string
-	for _, l := range ls {
+	for i, l := range ls {
 		if l.Name == labels.MetricName && !model.IsValidMetricName(model.LabelValue(l.Value)) {
 			return errors.Errorf("invalid metric name: %v", l.Value)
 		}
@@ -291,10 +290,9 @@ func validateLabelsAndMetricName(ls labels.Labels) error {
 		if !model.LabelValue(l.Value).IsValid() {
 			return errors.Errorf("invalid label value: %v", l.Value)
 		}
-		if l.Name == prevLabelName {
+		if i > 0 && l.Name == ls[i-1].Name {
 			return errors.Errorf("duplicate label with name: %v", l.Name)
 		}
-		prevLabelName = l.Name
 	}
 	return nil
 }
--- a/storage/remote/codec.go
+++ b/storage/remote/codec.go
@@ -280,7 +280,7 @@ func (c *concreteSeriesIterator) Err() error {
 // validateLabelsAndMetricName validates the label names/values and metric names returned from remote read,
 // also making sure that there are no labels with duplicate names
 func validateLabelsAndMetricName(ls labels.Labels) error {
-	existingLabelNames := make(map[string]bool, len(ls))
+	var prevLabelName string
 	for _, l := range ls {
 		if l.Name == labels.MetricName && !model.IsValidMetricName(model.LabelValue(l.Value)) {
 			return errors.Errorf("invalid metric name: %v", l.Value)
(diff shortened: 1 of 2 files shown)


storage/remote/codec.go
```diff
@@ -352,6 +353,21 @@ func LabelProtosToMetric(labelPairs []*prompb.Label) model.Metric {
 	return metric
 }
 
+// uniqueLabelNameProtos creates a []prompb.Label with duplicate label names removed
```
brian-brazil: We should error in this case, not hide the issue.

vmartsynovskyy: Good point. I'll add a check for duplicate labels in `validateLabelsAndMetricName` and update the tests I wrote accordingly.

storage/remote/codec_test.go
```diff
@@ -145,3 +147,59 @@ func TestConcreteSeriesClonesLabels(t *testing.T) {
 	gotLabels = cs.Labels()
 	require.Equal(t, lbls, gotLabels)
 }
+
+func TestDuplicateLabelsSimple(t *testing.T) {
```
cstyan: you could turn this and `TestDuplicateLabelsDoNothing` into a single test by using a subtest and turning the input/expected output into a struct

vmartsynovskyy: I removed these tests and replaced them with more test cases in `TestValidateLabelsAndMetricName`. I'll see if I can refactor that test to use subtests though.

vmartsynovskyy: I refactored `TestValidateLabelsAndMetricName` to use subtests, let me know if it looks right to you.

storage/remote/codec.go
```diff
@@ -277,8 +277,10 @@ func (c *concreteSeriesIterator) Err() error {
 	return nil
 }
 
-// validateLabelsAndMetricName validates the label names/values and metric names returned from remote read.
+// validateLabelsAndMetricName validates the label names/values and metric names returned from remote read,
+// also making sure that there are no labels with duplicate names
 func validateLabelsAndMetricName(ls labels.Labels) error {
+	existingLabelNames := make(map[string]bool, len(ls))
```
brian-brazil: Labels will be in order, so you only need to compare with the previous one.

vmartsynovskyy: Nice catch. I didn't notice that `labelProtosToLabels` sorts the labels.

I updated the validation function and added a corresponding test case (as well as updating the tests to actually check that expectedErr matches err).

storage/remote/codec.go
```diff
@@ -289,6 +291,10 @@ func validateLabelsAndMetricName(ls labels.Labels) error {
 		if !model.LabelValue(l.Value).IsValid() {
 			return errors.Errorf("invalid label value: %v", l.Value)
 		}
+		if l.Name == prevLabelName {
```
brian-brazil: You could avoid having prevLabelName by getting the index from the for, checking that it's >0 and then looking at `ls[i-1]`

vmartsynovskyy: Sure, I guess its cleaner that way.

storage/remote/codec_test.go
```diff
@@ -44,57 +48,96 @@ func TestValidateLabelsAndMetricName(t *testing.T) {
...
 		},
 		{
 			input: labels.FromStrings(
 				"__name__", "name",
 				"labelName", string([]byte{0xff}),
 			),
-			expectedErr: "Invalid label value: " + string([]byte{0xff}),
+			expectedErr: "invalid label value: " + string([]byte{0xff}),
 			shouldPass:  false,
+			description: "label value is 0xff",
```
brian-brazil: Invalid UTF=8 value

vmartsynovskyy: Updated

brian-brazil:
Thanks!

PR #5839, opened 2019-08-06
Use Eval for Rule Group Instrumentation

jtlisi:
This PR nests the instrumentation associated with rule groups inside the eval function, instead of in an anonymous function. This allows any downstream user of the Group struct to benefit from the instrumentation even if the Manager is not used to schedule evaluations.

@gouthamve @tomwilkie @cstyan @beorn7
Signed-off-by: Jacob Lisi <jacob.t.lisi@gmail.com>

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -287,25 +287,13 @@ func (g *Group) run(ctx context.Context) {
 		return
 	}
 
-	iter := func() {
-		g.metrics.iterationsScheduled.Inc()
-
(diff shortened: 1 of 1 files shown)


brian-brazil:
I'm not sure this makes sense in Prometheus terms, in Prometheus only the manager schedules things and the metrics don't mean the same thing if that's not the case.

I think you probably want your own metrics here.

Issue #1893, opened 2016-08-15
Make rule evaluation and federation consistent in time


beorn7:
Currently, if samples get ingested _during_ a rule is evaluated or a federation request is processed, only a subset of samples with the same timestamp may be taken into account during the evaluation.

Making ingestion of all the samples from a single scrape atomic would solve the problem, but the problem might be solved in an easier way by "simply" ignoring samples that came in after a certain watermark.

For reference: discussion started in https://github.com/prometheus/docs/issues/514 and already referenced #398 as a possible part of the solution.

beorn7:
Also #1887 for the reported incident that caused the discussion in the first place.

lmb:
I reported #2598, which was closed as a duplicate of this issue. I think this issue should mention that this behaviour breaks aggregating histograms using histogram_quantile in the currently released version. That should prevent more dupes.

Gouthamve:
AFAICS, this is fixed 2.0. Can this be closed @fabxc @brian-brazil

brian-brazil:
This is not fixed in 2.0 (yet).

brian-brazil:
I've got a basic version of this working over in the tsdb repo, it needs cleanup and performance work.

krasi-georgiev:
@beorn7 , @brian-brazil , @gouthamve  is this issue still unsolved?

brian-brazil:
Yes, there's an old PR over in TSDB for this. I believe @gouthamve was planning on working on it soon.

krasi-georgiev:
aah that must be the isolation PR https://github.com/prometheus/tsdb/pull/306

beorn7:
Based on some recent discussions I had, this seems to hurt quite a few people, especially for rules calculating histogram quantiles aggregating many targets.

The “almost done” [PR](https://github.com/prometheus/tsdb/pull/306) somehow became stale. @gouthamve doesn't have time right now to work on it. The PR is for the archived TSDB repo, so it has to be lifted over into prometheus/prometheus and then made ready for a final review. If anybody wants to do that, please speak up here in the issue. Otherwise, I'll add it to my long list of urgent things to do.

For reference: @gouthamve created [slides](https://docs.google.com/presentation/d/1-ICg7PEmDHYcITykD2SR2xwg56Tzf4gr8zfz1OerY5Y/edit?usp=sharing) to explain the design. Slides 14, 15, 16 have most of the process.

beorn7:
[...]

PR #5844, opened 2019-08-08
enhance histogram quantile function to handle data loss

garrettlish:
Assume we have quantile calculation expression like `histogram_quantile(0.99, sum by (le) (irate(read_request_latency_bucket{app="thanos"}[5m])))`, and there are some data loss for partial buckets (which means that higher bucket count migh be smaller than lower bucket count since irate is the per second rate of change at the end of the time window and higher bucket and lower bucket timestamp are not aligned). To avoid generating abnormal histogram quantile value, we can enchance the function by replacing the higher bucket count with next bucket count if it is smaller than lower bucket count.

### DIFF
--- a/promql/quantile.go
+++ b/promql/quantile.go
@@ -115,7 +115,27 @@ func bucketQuantile(q float64, buckets buckets) float64 {
 func coalesceBuckets(buckets buckets) buckets {
 	last := buckets[0]
 	i := 0
+	var next float64
 	for _, b := range buckets[1:] {
+		// use next count if current count is smaller than previous one which means data loss
+		if b.count < last.count {
+			if next >= last.count {
+				b.count = next
+			} else {
+				for j := 1; ; j++ {
+					if i+j > len(buckets)-1 {
+						next = last.count
+						b.count = next
+						break
+					}
+					if buckets[i+j].count >= last.count {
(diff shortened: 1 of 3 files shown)


brian-brazil:
How is this data loss occurring?

garrettlish:
Different buckets are different time series (le=0.01,le=0.025,le=0.05,le=0.1,etc.), one of the time series loss could happen while scraping or tsdb has problems.

brian-brazil:
I'm interested in how this happens, as scrape ingestion is atomic. Let's fix the root cause, rather than a symptom.

beorn7:
Which client library (and version thereof) has been used to create the histogram on the target side?
(There had been issues with non-atomic updates of the buckets. In client_golang, they were fixed in [v0.9](https://github.com/prometheus/client_golang/blob/master/CHANGELOG.md#090--2018-10-15).)

beorn7:
I'm closing this for lack of follow-up.

The problem that this PR tries to address should be solved at its root. The most likely candidate is (if not a buggy client library as discussed above) the lack of atomicity during ingestion. That would be #1893, and the stale PR to solve it is https://github.com/prometheus/tsdb/pull/306 . I'll talk to @gouthamve and see how we can revive it.

PR #5845, opened 2019-08-08
Correctly handle empty labels from alert templates.

brian-brazil:
Fixes https://github.com/prometheus/common/issues/36

Move logic handling this into the labels package,
so all the cases are handled in one place and we're
less likely to have this come up again.

@beorn7

### DIFF
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -285,20 +285,26 @@ type Builder struct {
 	add  []Label
 }
 
-// NewBuilder returns a new LabelsBuilder
+// NewBuilder returns a new LabelsBuilder.
 func NewBuilder(base Labels) *Builder {
-	return &Builder{
-		base: base,
-		del:  make([]string, 0, 5),
-		add:  make([]Label, 0, 5),
+	b := &Builder{
+		del: make([]string, 0, 5),
+		add: make([]Label, 0, 5),
 	}
+	b.Reset(base)
+	return b
 }
 
-// Reset clears all current state for the builder
+// Reset clears all current state for the builder.
 func (b *Builder) Reset(base Labels) {
 	b.base = base
 	b.del = b.del[:0]
 	b.add = b.add[:0]
+	for _, l := range b.base {
+		if l.Value == "" {
+			b.del = append(b.del, l.Name)
+		}
+	}
 }
 
 // Del deletes the label of the given name.
@@ -316,6 +322,10 @@ func (b *Builder) Del(ns ...string) *Builder {
 
 // Set the name/value pair as a label.
 func (b *Builder) Set(n, v string) *Builder {
+	if v == "" {
+		// Empty labels are the same as missing labels.
+		return b.Del(n)
+	}
 	for i, a := range b.add {
 		if a.Name == n {
 			b.add[i].Value = v
--- a/rules/alerting_test.go
+++ b/rules/alerting_test.go
@@ -49,8 +49,7 @@ func TestAlertingRuleLabelsUpdate(t *testing.T) {
 	testutil.Ok(t, err)
 	defer suite.Close()
 
-	err = suite.Run()
-	testutil.Ok(t, err)
+	testutil.Ok(t, suite.Run())
 
 	expr, err := promql.ParseExpr(`http_requests < 100`)
 	testutil.Ok(t, err)
@@ -152,8 +151,7 @@ func TestAlertingRuleExternalLabelsInTemplate(t *testing.T) {
 	testutil.Ok(t, err)
 	defer suite.Close()
 
-	err = suite.Run()
-	testutil.Ok(t, err)
+	testutil.Ok(t, suite.Run())
 
 	expr, err := promql.ParseExpr(`http_requests < 100`)
 	testutil.Ok(t, err)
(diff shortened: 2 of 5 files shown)


pkg/labels/labels.go
```diff
@@ -287,18 +287,24 @@ type Builder struct {
...
-		add:  make([]Label, 0, 5),
+	b := &Builder{
+		del: make([]string, 0, 5),
+		add: make([]Label, 0, 5),
 	}
+	b.Reset(base)
+	return b
 }
 
 // Reset clears all current state for the builder
```
bwplotka: Missing trailing period.

pkg/labels/labels.go
```diff
@@ -316,6 +322,9 @@ func (b *Builder) Del(ns ...string) *Builder {
 
 // Set the name/value pair as a label.
 func (b *Builder) Set(n, v string) *Builder {
+	if v == "" {
```
bwplotka: I would add a short sentence about this behavior in the comment.

rules/alerting_test.go
```diff
@@ -236,3 +236,59 @@ func TestAlertingRuleExternalLabelsInTemplate(t *testing.T) {
...
+
+func TestAlertingRuleEmptyLabelFromTemplate(t *testing.T) {
+	suite, err := promql.NewTest(t, `
+		load 1m
+			http_requests{job="app-server", instance="0"}	75 85 70 70
+	`)
+	testutil.Ok(t, err)
+	defer suite.Close()
+
+	err = suite.Run()
```
bwplotka: Why not inline? `testutil.Ok(t, suite.Run())`

scrape/scrape.go
```diff
@@ -449,20 +449,11 @@ func mutateSampleLabels(lset labels.Labels, target *Target, honor bool, rc []*re
 		}
 	} else {
 		for _, l := range target.Labels() {
-			lv := lset.Get(l.Name)
-			if lv != "" {
-				lb.Set(model.ExportedLabelPrefix+l.Name, lv)
-			}
+			lb.Set(model.ExportedLabelPrefix+l.Name, lset.Get(l.Name))
```
bwplotka: Maybe minor and not a blocker but in this case it is quite confusing to add two labels and base on the fact that one of those is empty. Also we unnecessarily add a couple of `adds` and `dels` to the builder just to avoid one if. Sound unnecessary to me. I would leave as this maybe?

```
lb.Set(l.Name, l.Value)
lv := lset.Get(l.Name)
if lv != "" {	
    lb.Set(model.ExportedLabelPrefix+l.Name, lv)	
}
```

scrape/scrape.go
```diff
@@ -449,20 +449,11 @@ func mutateSampleLabels(lset labels.Labels, target *Target, honor bool, rc []*re
 		}
```
bwplotka: can we change `lset` to `extLset` to be clear? Was quite confused until I figured out those are external labels.

scrape/scrape.go
bwplotka: Essentially it quite hard to read this code, maybe comment would help? as well?

bwplotka: Even better, can we use `mutateReportSampleLabels` here? We duplicate code essentially.

scrape/scrape.go
brian-brazil: It's not the external labels though, they're not involved in scraping. These are the sample labels.

scrape/scrape.go
brian-brazil: > base on the fact that one of those is empty

That's not what this logic is doing, both values can be non-empty.

> can we use mutateReportSampleLabels here?

I think that'd be less clear, there's different business logic for each.

brian-brazil:
Addressed all your other comments.

scrape/scrape.go
beorn7: In different news, while thinking through the code, I realised we have edge cases that are underspecified in our docs:
1. What happens if the target labels contain both `foo="a"` and `exported_foo="b"` and then a scrape contains its own label `foo="c"`? End result could be:
   1. `foo="a",exported_foo="b"
   1. `foo="a",exported_foo="c"
   1. `foo="a",exported_foo="b",exported_exported_foo="c"
2. What happens if a scrape contains both `foo="a"` and `exported_foo="b"` and then there is a target label `foo="c"`? End result could be:
   1. `foo="c",exported_foo="a"
   2. `foo="c",exported_foo="b"
   3. `foo="c",exported_foo="a",exported_exported_foo="b"

In case 2, the code will do i. But as said, that's not precisely specified in the docs. One might also expect iii.

Case 1 is more interesting, as the behavior depends on the order the target labels show up in the iteration. It could be i. or iii. That's in any case an undesired behavior.

For double fun, case 1 and case 2 could be combined (both the scrape and the target labels contain both a `foo` and an `exported_foo` label).

:boom:

brian-brazil: Yes, this is unspecified. Noone has complained about it after many years, so I haven't worried about it thus far.

Don't do this ;)

beorn7: We should at least document that the behavior is undefined if labels with the `exported_` prefix are part of the game.

I think, however, that we can do better. Especially the more or less random choice of behavior in case 2 worries me.

brian-brazil: Made your suggested code change.

The proper way would be iii, and that any exported_ scraped label would always be prefixed. It's technically a breaking change.

beorn7: Yeah, just by reading the docs, I'd expect iii., although it's not very clearly expressed. We could still claim we are just bringing the implementation in line with the (vaguely) specified behavior. Thus, bug fix, not breaking change. https://xkcd.com/1172/

I'll file a separate issue to keep track.

PR #5847, opened 2019-08-09
Make a rule's evaluation timestamp available to alert templates

seveas:
We have a fairly complex monitoring/alerting pipeline, that includes several monitoring systems all routing into a homegrown alert router and out to various endpoints. I am measuring latency of all components by sending synthetic alerts throught the system, recording ingress and egress timestamps of all subsystems in the alert itself. For prometheus, I can record an ingress timestamp because the metric it is fetching is a magical one that is always a timestamp-like value. But a sensible definition of "latency inside prometheus" should take into account the fact that metric ingress and evaluation of a rule happen at different times, so I would really like to record the evaluation time as well.

This seemed simple enough to do: prometheus already is recording this information, I just had to change it to record it a bit earlier, and then add it to the template data. I hope this approach is acceptable.

### DIFF
--- a/pkg/rulefmt/rulefmt.go
+++ b/pkg/rulefmt/rulefmt.go
@@ -156,7 +156,7 @@ func testTemplateParsing(rl *Rule) (errs []error) {
 	}
 
 	// Trying to parse templates.
-	tmplData := template.AlertTemplateData(map[string]string{}, map[string]string{}, 0)
+	tmplData := template.AlertTemplateData(map[string]string{}, map[string]string{}, 0, time.Now())
 	defs := []string{
 		"{{$labels := .Labels}}",
 		"{{$externalLabels := .ExternalLabels}}",
--- a/rules/alerting.go
+++ b/rules/alerting.go
@@ -318,7 +318,7 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc, 
 			l[lbl.Name] = lbl.Value
 		}
 
-		tmplData := template.AlertTemplateData(l, r.externalLabels, smpl.V)
+		tmplData := template.AlertTemplateData(l, r.externalLabels, smpl.V, r.evaluationTimestamp)
 		// Inject some convenience variables that are easier to remember for users
 		// who are not used to Go's templating system.
 		defs := []string{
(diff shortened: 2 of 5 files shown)


brian-brazil:
The evaluation timestamp is already available as `time()`.

Your PR is doing something different, and would make alerts vary in execution time which is not desirable.

seveas:
Hi Brian,

Thanks for yout response! If I try to use `time()` in an alert template, I get an error saying `function "time" not defined`. I see there is a query function with that name, but it's rather cumbersome to use in a template: `{{ (index (query \"time()\") 0).Value | printf \"%.0f\" }}`.

If the current approach is not acceptable, would adding a template function named `time` that returns a `time.Time` obejct be good?

Could you also help me understand how this makes alerts vary in execution time? All it does is move a SetEvaluationTimestamp call from after to before evaluating the alert.

brian-brazil:
> I see there is a query function with that name, but it's rather cumbersome to use in a template

That's how you access any PromQL in templates. It's not that common to do it.

> Could you also help me understand how this makes alerts vary in execution time?

What you're proposing is walltime, not evaluation time. We explicitly don't provide walltime to PromQL. You'd also be changing semantics of the status pages to something less useful.

seveas:
>> I see there is a query function with that name, but it's rather cumbersome to use in a template
> That's how you access any PromQL in templates. It's not that common to do it.

That's a bit of a circular argument. I'd rather not use promql, all I need is the evaluation timestamp of the rule that's triggering the alert for which I'm writing a template. I did that by adding a field to the template data, the only reason I tried it now with promql is because you said there's a `time()` function, and that function only exists as a promql function, not a template function.

>> Could you also help me understand how this makes alerts vary in execution time?
> What you're proposing is walltime, not evaluation time. We explicitly don't provide walltime to PromQL. You'd also be changing semantics of the status pages to something less useful.

Sorry, but that does not explain how this makes execution times vary. Moreover, I'm *not* proposing walltime in promql anywhere, or changing any status pages. All this change does is making the alert evaluation timestamp available to the templates that are used for alerts caused by the rule. And to do that, it needs to set it a bit earlier, but the actual value of the timestamp does not change.

brian-brazil:
> Sorry, but that does not explain how this makes execution times vary. 

It'd now vary based on walltime.

> And to do that, it needs to set it a bit earlier,

Which changes the status page, in a misleading way.

roidelapluie:
As there is `{{ (index (query \"time()\") 0).Value | printf \"%.0f\" }}` available, we have decided during our bug scrub to close this PR, . Thanks!

PR #5849, opened 2019-08-09
Cache labels.Labels to Identify Series in Remote Write

csmarchbanks:
This is an experiment to have the remote write component use `labels.Labels` instead of `[]prompb.Label` when passing samples around/identifying series. A `label.Label` uses less space so steady state memory usage should be less, with the tradeoff of transforming `label.Labels` to `[]prompb.Label` for each remote request.

Results after removing some extra sources of allocations look very good. This is for a server with three remote-write targets, and ~8k samples per second ingestion.

### Benchmarks
```
BenchmarkSampleDelivery-8     998914        1007751       +0.88%

BenchmarkSampleDelivery-8     7117           6117           -14.05%

BenchmarkSampleDelivery-8     688391        640400        -6.97%
```

### Heap profiles
Heap profiles show a 50% reduction in memory used by StoreSeries as expected.

#### v2.11.1:
v2 11 1

#### rw-use-labels branch
rw-use-labels

### Other metrics
CPU usage is very similar between the servers. However, there is a reduction in allocations:
ReducedAllocations

### DIFF
--- a/storage/remote/codec.go
+++ b/storage/remote/codec.go
@@ -131,7 +131,7 @@ func ToQueryResult(ss storage.SeriesSet, sampleLimit int) (*prompb.QueryResult, 
 		}
 
 		resp.Timeseries = append(resp.Timeseries, &prompb.TimeSeries{
-			Labels:  labelsToLabelsProto(series.Labels()),
+			Labels:  labelsToLabelsProto(series.Labels(), nil),
 			Samples: samples,
 		})
 	}
@@ -192,6 +192,7 @@ func StreamChunkedReadResponses(
 ) error {
 	var (
 		chks     []prompb.Chunk
+		lbls     []prompb.Label
 		err      error
 		lblsSize int
 	)
@@ -199,7 +200,7 @@ func StreamChunkedReadResponses(
 	for ss.Next() {
 		series := ss.At()
 		iter := series.Iterator()
-		lbls := MergeLabels(labelsToLabelsProto(series.Labels()), sortedExternalLabels)
+		lbls = MergeLabels(labelsToLabelsProto(series.Labels(), lbls), sortedExternalLabels)
 
 		lblsSize = 0
 		for _, lbl := range lbls {
@@ -522,8 +523,13 @@ func labelProtosToLabels(labelPairs []prompb.Label) labels.Labels {
 	return result
 }
 
-func labelsToLabelsProto(labels labels.Labels) []prompb.Label {
-	result := make([]prompb.Label, 0, len(labels))
+// labelsToLabelsProto transforms labels into prompb labels. The buffer slice
+// will be used to avoid allocations if it is big enough to store the labels.
+func labelsToLabelsProto(labels labels.Labels, buf []prompb.Label) []prompb.Label {
+	result := buf[:0]
+	if cap(buf) < len(labels) {
+		result = make([]prompb.Label, 0, len(labels))
+	}
 	for _, l := range labels {
 		result = append(result, prompb.Label{
 			Name:  l.Name,
--- a/storage/remote/codec.go
+++ b/storage/remote/codec.go
@@ -526,8 +526,8 @@ func labelsToLabelsProto(labels labels.Labels) []prompb.Label {
 	result := make([]prompb.Label, 0, len(labels))
 	for _, l := range labels {
 		result = append(result, prompb.Label{
-			Name:  interner.intern(l.Name),
-			Value: interner.intern(l.Value),
+			Name:  l.Name,
+			Value: l.Value,
 		})
 	}
 	return result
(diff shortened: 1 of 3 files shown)


storage/remote/queue_manager.go
```diff
@@ -552,13 +555,19 @@ func (s *shards) start(n int) {
...
-	newQueues := make([]chan prompb.TimeSeries, n)
+	newQueues := make([]chan sample, n)
 	for i := 0; i < n; i++ {
-		newQueues[i] = make(chan prompb.TimeSeries, s.qm.cfg.Capacity)
+		newQueues[i] = make(chan sample, s.qm.cfg.Capacity)
 	}
 
 	s.queues = newQueues
 
+	s.tsPool = sync.Pool{
```
brian-brazil: Does this really help? Changing the number of shards isn't that often.

storage/remote/codec.go
```diff
@@ -372,9 +372,16 @@ func labelsToLabelsProto(labels labels.Labels) []prompb.Label {
 	result := make([]prompb.Label, 0, len(labels))
```
brian-brazil: You can be smarter here I think, I presume this is where the allocations are.

storage/remote/queue_manager.go
csmarchbanks: Benchmarks improved some when I first added it. The pool is used for every http request, not just the number of shards, though it could probably be done by keeping a single `[]prompb.TimeSeries` per shard instead of using a pool, similar to the compression buffer.

storage/remote/codec.go
csmarchbanks: :+1: Many were here, and some still in buildWriteRequest from making `[]prompb.Sample`. Latest update helps both of these.

storage/remote/queue_manager.go
```diff
@@ -748,16 +752,17 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
...
 
-func buildWriteRequest(samples []prompb.TimeSeries, buf []byte) ([]byte, int64, error) {
+func (s *shards) buildWriteRequest(samples []sample, buf []byte) ([]byte, int64, error) {
 	var highest int64
-	for _, ts := range samples {
-		// At the moment we only ever append a TimeSeries with a single sample in it.
-		if ts.Samples[0].Timestamp > highest {
-			highest = ts.Samples[0].Timestamp
+	req := s.writeRequests.get(len(samples))
+	defer s.writeRequests.put(req)
```
brian-brazil: defer causes an alloc. I'd avoid sync.Pool entirely, it has interesting interactions with GC and in this case we know exactly how much memory we need.

csmarchbanks: :+1: Removed the pool.

brian-brazil:
That's looking okay. What do the benchmarks say now?

csmarchbanks:
I removed the merge conflict, and updated the benchmarks, slightly fewer allocations than last time, CPU jumps around between runs but is about the same before/after.

I don't expect the graphs to change too much, but can repost them in a few hours if interested.

csmarchbanks:
@brian-brazil would you mind taking another look when you have a moment?

brian-brazil:
:+1:

storage/remote/codec.go
```diff
@@ -192,14 +192,15 @@ func StreamChunkedReadResponses(
...
+		lbls     []prompb.Label
 		err      error
 		lblsSize int
 	)
 
 	for ss.Next() {
 		series := ss.At()
 		iter := series.Iterator()
-		lbls := MergeLabels(labelsToLabelsProto(series.Labels()), sortedExternalLabels)
+		lbls = MergeLabels(labelsToLabelsProto(series.Labels(), lbls), sortedExternalLabels)
```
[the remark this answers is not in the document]
cstyan: looks like remote read only uses `labelsToLabelsProto` in tests so it should be okay

storage/remote/queue_manager.go
```diff
@@ -545,11 +540,17 @@ func (t *QueueManager) newShards() *shards {
 	return s
 }
 
+type sample struct {
```
cstyan: Hmmm, would be nice if we could use a different name. To me sample implies just ts and value, but I can't think of anything better right now.

storage/remote/codec.go
```diff
@@ -522,13 +523,25 @@ func labelProtosToLabels(labelPairs []prompb.Label) labels.Labels {
...
-			Name:  interner.intern(l.Name),
-			Value: interner.intern(l.Value),
+			Name:  l.Name,
+			Value: l.Value,
 		})
 	}
 	return result
 }
+
+func internLabels(lbls labels.Labels) {
```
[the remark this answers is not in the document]
cstyan: release is there too

storage/remote/queue_manager.go
csmarchbanks: I could call it something like `queueSample` to differentiate it a bit, but I don't know if that gains anything.

storage/remote/codec.go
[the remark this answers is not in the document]
csmarchbanks: :+1:

PR #5852, opened 2019-08-10
cleanup: correct func name in log message

ethan-daocloud:
Signed-off-by: Guangming Wang <guangming.wang@daocloud.io>


brian-brazil:
Thanks!

Issue #5871, opened 2019-08-13
tsdb: SeriesIterator.Seek() is vaguely defined


free:
Context: I was looking into cleaning up the iterator code, as I'm trying to add a separate `SeekBefore()` method, that would return the value at time `t` as defined by Prometheus without always parsing and iterating over a whole chunk covering the previous 5 minutes (when the looked up value is toward the beginning of a chunk).

However, I ran into some issues regarding how chunked series and iterators should actually work. There are a couple of gray areas that are only partially covered by comments or test code, making it impossible to figure out what is the expected correct behavior:

1. `blockQuerier` and the resulting `chunkedSeries` and `chunkSeriesIterator` all have `mint` and `maxt` fields, which (along with `(*chunkSeriesIterator).Seek()` implementation details) suggest that iterators should only return samples with timestamps `mint <= t <= maxt`. However, there are bugs (such as the one I'm attempting to fix with prometheus/tsdb#327) and even [confusing/confused unit tests](https://github.com/prometheus/tsdb/blob/54205903f79fdff71d52682b09dd760259bca201/querier_test.go#L888) -- which appear to create an iterator with `mint = 5`, `maxt = 8` and expect `(*chunkSeriesIterator).Seek()` to return a value with timestamp `3`.<p>Are`mint` and `maxt` supposed to be "hard" limits on the respective iterators or only on which chunks are selected?
2. I gather (from the implementation again) that `SeriesIterator.Seek()` is only supposed to seek forward (which doesn't work as expected, as it will also seek backward within the same chunk). Is this assumption correct?
4. Is `Seek()` supposed to always advance the iterator or should a second `Seek()` call with the same timestamp as the immediately preceding one leave the iterator positioned where it was before?
3. Is it actually a good idea for a `Seek()` with a timestamp `t < mint` to ever succeed? If so a poorly defined series e.g. `[now - 10m, now]` will "successfully" produce a value of the series at `now - 1y` equal to the first value after `now - 10m`.

I'm glad to do the work (including properly documenting the code) regardless of what the answers are, I'd just like to get some "authoritative" input. Thanks.

brian-brazil:
@bwplotka @free Was this resolved in the end?

bboreham:
Is there anything left to do here?

beorn7:
Hello from the bug scrub.

> Is there anything left to do here?

We take three years of silence as "no". Closing.

PR #5882, opened 2019-08-13
Add ChunksIterator method to Series interface.

bwplotka:
Continuation from: https://github.com/prometheus/tsdb/pull/665

## Changes:

* Adjusts Iterator Interface of Series e.g Seek
* Reuse single interface instead of 3 same ones spread across 3 packages.
* Refactored tests
* Added ChunkIterator to iterate over chunks if requested by the client. 

Fixes: https://github.com/prometheus/prometheus/issues/5871

### DIFF
--- a/storage/generic.go
+++ b/storage/generic.go
@@ -1,0 +1,88 @@
+package storage
+
+import "github.com/prometheus/prometheus/pkg/labels"
+
+// Boilerplate on purpose. Generics some day...
+
+type genericQuerierAdapter struct {
+	baseQuerier
+
+	// Union.
+	q  Querier
+	cq ChunkedQuerier
+}
+
+func (q *genericQuerierAdapter) Select(params *SelectParams, matchers ...*labels.Matcher) (genericSeriesSet, Warnings, error) {
+	if q.q != nil {
+		s, w, err := q.q.Select(params, matchers...)
+		return s.(genericSeriesSet), w, err
+	}
+	s, w, err := q.cq.Select(params, matchers...)
+	return s.(genericSeriesSet), w, err
+}
+func (q *genericQuerierAdapter) SelectSorted(params *SelectParams, matchers ...*labels.Matcher) (genericSeriesSet, Warnings, error) {
+	if q.q != nil {
+		s, w, err := q.q.SelectSorted(params, matchers...)
+		return s.(genericSeriesSet), w, err
+	}
+	s, w, err := q.cq.SelectSorted(params, matchers...)
+	return s.(genericSeriesSet), w, err
+}
+
+func newGenericQuerierFrom(q Querier) genericQuerier {
+	return &genericQuerierAdapter{baseQuerier: q, q: q}
+}
+
+func newGenericQuerierFromChunked(cq ChunkedQuerier) genericQuerier {
+	return &genericQuerierAdapter{baseQuerier: cq, cq: cq}
+}
+
+type genericQuerier interface {
+	baseQuerier
+	Select(*SelectParams, ...*labels.Matcher) (genericSeriesSet, Warnings, error)
+	SelectSorted(*SelectParams, ...*labels.Matcher) (genericSeriesSet, Warnings, error)
+}
+
+type genericSeriesSet interface {
+	Next() bool
+	At() Labeled
+	Err() error
+}
+
+type querierAdapter struct {
+	genericQuerier
+}
+
+func (q *querierAdapter) Select(params *SelectParams, matchers ...*labels.Matcher) (SeriesSet, Warnings, error) {
+	s, w, err := q.Select(params, matchers...)
+	return s.(SeriesSet), w, err
+}
+func (q *querierAdapter) SelectSorted(params *SelectParams, matchers ...*labels.Matcher) (SeriesSet, Warnings, error) {
+	s, w, err := q.SelectSorted(params, matchers...)
+	return s.(SeriesSet), w, err
+}
+
+type chunkedQuerierAdapter struct {
+	genericQuerier
+}
+
+func (q *chunkedQuerierAdapter) Select(params *SelectParams, matchers ...*labels.Matcher) (ChunkedSeriesSet, Warnings, error) {
+	s, w, err := q.Select(params, matchers...)
+	return s.(ChunkedSeriesSet), w, err
+}
+func (q *chunkedQuerierAdapter) SelectSorted(params *SelectParams, matchers ...*labels.Matcher) (ChunkedSeriesSet, Warnings, error) {
+	s, w, err := q.SelectSorted(params, matchers...)
+	return s.(ChunkedSeriesSet), w, err
+}
+
+type genericSeriesMerger interface {
+	NewMergedSeries(...Labeled) Labeled
+}
+
+func newGenericSeriesMergerFrom(q SeriesMerger) genericSeriesMerger {
+
+}
+
+func newGenericSeriesMergerFromChunked(q ChunkedSeriesMerger) genericSeriesMerger {
+
+}
--- a/storage/interface.go
+++ b/storage/interface.go
@@ -20,7 +20,6 @@ import (
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/tsdb/chunkenc"
 	"github.com/prometheus/prometheus/tsdb/chunks"
-	"github.com/prometheus/prometheus/tsdb/tombstones"
 )
 
 // The errors exposed.
@@ -54,17 +53,10 @@ type Storage interface {
 type Queryable interface {
 	// Querier returns a new Querier on the storage.
 	Querier(ctx context.Context, mint, maxt int64) (Querier, error)
+	ChunkedQuerier(ctx context.Context, mint, maxt int64) (ChunkedQuerier, error)
 }
 
-// Querier provides querying access over time series data of a fixed
-// time range.
-type Querier interface {
-	// Select returns a set of series that matches the given label matchers.
-	Select(*SelectParams, ...*labels.Matcher) (SeriesSet, Warnings, error)
-
-	// SelectSorted returns a sorted set of series that matches the given label matchers.
-	SelectSorted(*SelectParams, ...*labels.Matcher) (SeriesSet, Warnings, error)
-
+type baseQuerier interface {
 	// LabelValues returns all potential values for a label name.
 	// It is not safe to use the strings beyond the lifefime of the querier.
 	LabelValues(name string) ([]string, Warnings, error)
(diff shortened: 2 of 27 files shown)


tsdb/querier.go
```diff
@@ -872,22 +889,14 @@ func (s *chunkSeries) Labels() labels.Labels {
...
-	// after t.
-	Seek(t int64) bool
-	// At returns the current timestamp/value pair.
-	At() (t int64, v float64)
-	// Next advances the iterator by one.
-	Next() bool
-	// Err returns the current error.
-	Err() error
+func (s *chunkSeries) ChunkIterator() ChunkIterator {
+	// TODO(bwplotka): This is wrong. We probably have to be strict in terms of min/maxt and tombstones.
```
bwplotka: TODO/TO decide:

When we want to apply tombstones in `ChunkIterator`? :/ 

TBH it might be the most efficient to pass intervals upstream e.g in streamed remote read protocol as well (???)

cc @brian-brazil

bwplotka:
Addressed all comments from the previous PR: https://github.com/prometheus/tsdb/pull/665 PTAL @brian-brazil @krasi-georgiev

tsdb/querier.go
bwplotka: Two options:

A) Naive: Rewrite chunk on the fly applying `mint,maxt` and `dranges`. It is must have for `Select` interface which assumes things are scoped to given time range and tombstones are tolerated.
B) Ignore mint,maxt level. Maybe allow ChunkIterator to return `deleted` intervals as well. This probably means extending the remote read chunked protocol [here](https://github.com/prometheus/prometheus/pull/5703) for passing tombstones intervals as well -> let client side to apply intervals.. However, we leak "deleted" metrics. If they were deleted because of GDPR or something, this is kind of wrong (: 

I think overall we need to do A here. Thoughts @brian-brazil @tomwilkie @codesome ?

brian-brazil: Definitely A, for general data security we don't want to be returning deleted data. It'd also push that complexity onto all readers, and they'd probably forget it.

tsdb/head.go
```diff
@@ -1617,6 +1617,8 @@ func newMemSeries(lset labels.Labels, id uint64, chunkRange int64) *memSeries {
 		ref:        id,
 		chunkRange: chunkRange,
 		nextAt:     math.MinInt64,
+		// This is to make sure iterator.At without advancing returns default values.
```
brian-brazil: It's not valid to call At without advancing.

tsdb/chunkenc/chunk.go
```diff
@@ -56,11 +57,20 @@ type Appender interface {
 	Append(int64, float64)
 }
 
-// Iterator is a simple iterator that can only get the next value.
+// Iterator iterates over the samples of a time series.
 type Iterator interface {
+	// Seek advances the iterator forward to the sample with the timestamp t or first value after t.
+	// Seek has no effect if requested timestamp is the same or lower than the current iterator position.
+	// Seek returns false if there is no such sample with the timestamp equal or larger than t.
+	// Iterator is exhausted when the Seek returns false.
```
krasi-georgiev: Is that desired behavior? 
I mean does it make  sense to exhaust the iterator if the current time is 100 the iterator max time is 200 and you request Seek to 10? 
Logically I would think that exhausting the iterator would happen when you request 210 for example.

```suggestion
	// Iterator is exhausted when the requested Seek is higher than the iterator maximum timestamp.
```

tsdb/querier.go
```diff
@@ -971,42 +988,56 @@ func (s *verticalChainedSeries) Labels() labels.Labels {
...
-		b: newVerticalMergeSeriesIterator(s[1:]...),
+		a:    s[0].Iterator(),
+		b:    newVerticalMergeSeriesIterator(s[1:]...),
+		curT: math.MinInt64,
 	}
 }
 
 func (it *verticalMergeSeriesIterator) Seek(t int64) bool {
+	if it.initialized && it.curT >= t {
+		return true
```
krasi-georgiev: hm this doesn't look right. This would return true when `it.curT` is 150 and the requested time is 100

I thought we want the following behaviour:
Iterator with `minT:100`, `maxT:200`, `curT:150`

Seek(150) - returns true and the Seek is noop, doesn't advance.
Seek(100) - returns false and is noop.
Seek(160) - returns true and advances the iterator to 160 or the next available timestamp.
Seek(210) - returns false and exhausts the iterator.
When it.Err is not nil return false in all cases.

Maybe putting an example like this in the iterator description would make things a bir more clear for what is expected in the different scenarios.

tsdb/querier.go
```diff
@@ -1098,38 +1268,56 @@ func (it *chunkSeriesIterator) resetCurIterator() {
...
 
-func (it *chunkSeriesIterator) Seek(t int64) (ok bool) {
+func (it *chunkSeriesIterator) Seek(t int64) bool {
+	if it.Err() != nil || it.i > len(it.chunks)-1 {
+		return false
+	}
+
 	if t > it.maxt {
+		// Exhaust iterator.
+		it.i = len(it.chunks)
```
krasi-georgiev: maybe it is just my style, but I would think it will be easier to follow with an explicit var `it.exhausted = true`. Can leave as it if you prefer.

tsdb/querier.go
```diff
@@ -1138,22 +1326,17 @@ func (it *chunkSeriesIterator) Next() bool {
...
-
-			return t <= it.maxt
 		}
-		if t > it.maxt {
-			return false
-		}
-		return true
+		return t <= it.maxt
 	}
 	if err := it.cur.Err(); err != nil {
```
krasi-georgiev:
```suggestion
	if it.cur.Err() != nil {
```

tsdb/querier.go
```diff
@@ -940,7 +957,7 @@ func (it *chainedSeriesIterator) Next() bool {
 	if it.cur.Next() {
 		return true
 	}
-	if err := it.cur.Err(); err != nil {
+	if it.cur.Err() != nil {
```
krasi-georgiev: I completely forgot we can do that :1st_place_medal: nice reminder, thanks!

tsdb/querier_test.go
```diff
@@ -1053,8 +1304,9 @@ func TestChunkSeriesIterator_DoubleSeek(t *testing.T) {
 	}
 
 	res := newChunkSeriesIterator(chkMetas, nil, 2, 8)
-	testutil.Assert(t, res.Seek(1) == true, "")
-	testutil.Assert(t, res.Seek(2) == true, "")
+	testutil.Assert(t, res.Seek(1), "")
```
krasi-georgiev: don't think we even need this test anymore as we can cover it in `TestSeriesIterators`

tsdb/querier.go
bwplotka: >  It'd also push that complexity onto all readers, and they'd probably forget it.

You would NOT put it to readers you mean? 

Makes sense to me.. Let's do it then here.

bwplotka:
Need to get back to this soon (:

bwplotka:
Getting back to this, expect changes soon: WIP.

@brian-brazil @codesome do you know why we have two places with storage interfaces? 
tsdb/Querier have some: https://github.com/prometheus/prometheus/blob/master/tsdb/querier.go#L33
And storage has almost the same: https://github.com/prometheus/prometheus/blob/master/storage/interface.go

Since we merged TSDB with Prometheus, do you think it makes sense to unify those together?

brian-brazil:
It's historical, but we probably want to think about the impact on the remote storage code too.

bwplotka:
We can literally remove and deduplicate lot's of iterators.. expect big PR :cry: 

Should be ready for review EOW

bwplotka:
This PR is getting too big. I split it by function into ~3 @pracucci @pstibrany 

The first one in line is https://github.com/prometheus/prometheus/pull/6777. It is ready to review - I am fixing the last unit test hopefully.

This is very breaking PR for Thanos and Cortex projects as we kill and unify several important interfaces. Nothing hidden though, go build will tell you what to change (:

bwplotka:
In heavy progress.

tsdb/chunks/chunks.go
```diff
@@ -67,6 +67,20 @@ type Meta struct {
 	MinTime, MaxTime int64
 }
 
+// Iterator iterates over the chunk of a time series.
+type Iterator interface {
+	//// Seek advances the iterator forward to the given timestamp.
+	//// It advances to the chunk with min time at t or first chunk with min time after t.
+	//Seek(t int64) bool
+	// At returns the current meta.
+	// It depends on implementation if the chunk is populated or not.
```
krasi-georgiev: why is that? Why not just return non empty chunks?

storage/series.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/prometheus/prometheus/tsdb/tsdbutil"
+)
+
+type ConcreteSeries struct {
+	labels           labels.Labels
+	SampleIteratorFn func() chunkenc.Iterator
+	ChunkIteratorFn  func() chunks.Iterator
+}
+
+func NewTestSeries(lset labels.Labels, samples ...[]tsdbutil.Sample) Series {
```
krasi-georgiev: this is only for testing right? so why not put in a test package or a test file?

storage/generic.go
```diff
@@ -0,0 +1,88 @@
...
+
+type genericQuerier interface {
+	baseQuerier
+	Select(*SelectParams, ...*labels.Matcher) (genericSeriesSet, Warnings, error)
+	SelectSorted(*SelectParams, ...*labels.Matcher) (genericSeriesSet, Warnings, error)
+}
+
+type genericSeriesSet interface {
+	Next() bool
+	At() Labeled
```
krasi-georgiev:
```suggestion
	At() Series
```

bwplotka: Nope, we need Labelled, but also because it's actually `genericSeries` ;p

storage/series.go
bwplotka: Because we want to reuse this across all unit test, even those not in this package.

tsdb/chunks/chunks.go
bwplotka: I don't think you know what I mean here. Populated is if you have only `Ref` or if you have actually `Bytes` of chunk there (:

storage/interface.go
```diff
@@ -54,26 +53,43 @@ type Storage interface {
...
 	Select(*SelectParams, ...*labels.Matcher) (SeriesSet, Warnings, error)
 
 	// SelectSorted returns a sorted set of series that matches the given label matchers.
 	SelectSorted(*SelectParams, ...*labels.Matcher) (SeriesSet, Warnings, error)
+}
 
-	// LabelValues returns all potential values for a label name.
-	// It is not safe to use the strings beyond the lifefime of the querier.
-	LabelValues(name string) ([]string, Warnings, error)
+// ChunkedQuerier provides querying access over time series' chunks of a fixed
```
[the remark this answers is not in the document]
bwplotka: Can do (:

storage/generic.go
```diff
@@ -0,0 +1,88 @@
...
+type querierAdapter struct {
+	genericQuerier
+}
+
+func (q *querierAdapter) Select(params *SelectParams, matchers ...*labels.Matcher) (SeriesSet, Warnings, error) {
+	s, w, err := q.Select(params, matchers...)
+	return s.(SeriesSet), w, err
+}
+func (q *querierAdapter) SelectSorted(params *SelectParams, matchers ...*labels.Matcher) (SeriesSet, Warnings, error) {
+	s, w, err := q.SelectSorted(params, matchers...)
```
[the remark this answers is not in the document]
bwplotka: yes, lol

bwplotka:
> label querier?

> Labeled

I like `labelled` as it is "something" that has labels (:

bwplotka:
OK, let's close this in favor of clean version of same thing: https://github.com/prometheus/prometheus/pull/6990

Issue #5868, opened 2019-08-13
tsdb: [feature request]: support index file size >64Gi


davidquarles:
We're actually running into this on our largest cluster and it's brought thanos compaction to a standstill. I'm whittling away at cardinality, sharding our scrape targets where possible and backing down our retention policies.  Hopefully all of the above will do the trick in the near-term, but it would be great to not have to worry about this in the future.

https://github.com/prometheus/tsdb/blob/e7436e13f07511ee669876905f5147ae1713f5bd/index/index.go#L233

codesome:
Fix for this is here https://github.com/prometheus/prometheus/pull/5884

beorn7:
During our bug scrub, we realized that #5884 was caught in review limbo for years now. We are wondering how much this feature is actually needed and if the benefits outweigh the complications. See [here](https://github.com/prometheus/prometheus/issues/9078#issuecomment-1572752630) for a line of argument why we shouldn't do this.

We would like to get an idea about the relevance of the problem, so if you still run into this problem, feel free to comment on this issue.

rtksamsmith:
My personal opinion:

Once you're big enough to hit this (outrageously high) limit, _all_ mitigations are complicated or expensive. Scale is a multiplier of complexity. As you know, at this scale, the necessary rearchitecture to work around something like this is months of work. That's compounded by the fact that it's a hidden limitation - with no warning or info level logs suggesting that you may be approaching a limit - so it ends up saddling teams with a significant, unplanned lift to prevent critical, production-level impacts. That's, like, the trifecta of "uh oh."

Back in my day, we were lucky to have 100 MiB of storage, and my day wasn't _that_ long ago. Arbitrarily deciding "that ought to be enough bytes for anybody" has been consistently proven to show lack of foresight. Maybe Prometheus stacks reaching 64GiB blocks are rare today, but will they be rare next year? The year after? If Prometheus continues to hold the spot as the industry leading platform, the number of teams impacted will increase rapidly as a result of two simple factors: "we're at scale, so we should be using the industry leader" + "our observability is working, let's do more of it."

This limit will be patched out at some point. It will have to be. I'd urge maintainers to opt for "sooner" rather than "later" given the severity and suddenness of impact. If you wait until some critical mass of users have hit this limit, you'll be too late, because nobody runs `latest` in prod. Better to have been fixed a couple major versions ago than need to bump a major and a dozen minors for remediation.

nixfloyd:
[...]

PR #5884, opened 2019-08-13
[tsdb] Add 64 bit prefixCompressedPostings

naivewong:
Closes https://github.com/prometheus/prometheus/issues/5868

Refer to this [tsdb/pr672](https://github.com/prometheus/tsdb/pull/672) and this [doc](https://docs.google.com/document/d/1FX4wWfcYI4eE9CmdPHdfjecmp2jzyWKloKuqK5suxls/edit#heading=h.fhoqupzha3ie).

### DIFF
--- a/tsdb/index/postings_test.go
+++ b/tsdb/index/postings_test.go
@@ -808,13 +808,6 @@ func BenchmarkPostingsOneBlock(b *testing.B) {
 		binary.BigEndian.PutUint32(b, ls1[i])
 	}
 
-	// bigEndian64Postings for ls1.
-	bufBE641 := make([]byte, num*8)
-	for i := 0; i < num; i++ {
-		b := bufBE641[i*8 : i*8+8]
-		binary.BigEndian.PutUint64(b, uint64(ls1[i]))
-	}
-
 	// prefixCompressedPostings for ls1.
 	bufPCP1 := encoding.Encbuf{}
 	temp := make([]uint64, len(ls1))
@@ -828,13 +821,6 @@ func BenchmarkPostingsOneBlock(b *testing.B) {
 	for i := 0; i < num; i++ {
 		b := bufBE2[i*4 : i*4+4]
 		binary.BigEndian.PutUint32(b, ls2[i])
-	}
-
-	// bigEndian64Postings for ls2.
-	bufBE642 := make([]byte, num*8)
-	for i := 0; i < num; i++ {
-		b := bufBE642[i*8 : i*8+8]
-		binary.BigEndian.PutUint64(b, uint64(ls2[i]))
 	}
 
 	// prefixCompressedPostings for ls2.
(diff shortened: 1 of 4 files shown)


codesome:
Not sure what is happening but the query inner_eval times are 2x for this PR even before any compactions - which means only the head block which this PR doesn't touch.

http://prombench.prometheus.io/grafana/d/7gmLoNDmz/prombench?orgId=1&var-RuleGroup=All&var-pr-number=5884&from=1565719211808&to=1565720984397
(There is even some difference in RSS)

Cancelled the benchmark to investigate this first before testing the PR.

codesome:
In the last benchmark run the query inner_eval times seemed to be comparable. Not sure what was wrong during the first run. Anyway, now the RSS of this PR was slightly higher than master even before any blocks were written to the disk (and the PR touches only the postings of the disk). Not sure how is that possible as this PR doesn't touch the head. @naivewong and I will be investigating the downloaded profiles.

codesome:
The last benchmark had less difference in RSS compared to the previous benchmark (without any code changes). Looking at the profiles I infer that it is coming from the queries, and mainly because of the difference in the number of chunks (small diff in startup time causes it?).

As the change in postings encoding doesn't affect the count of chunks, I believe the perf looks fine. I would like someone to verify the benchmark.

The last benchmark (need to zoom in at the required time to see it properly): http://prombench.prometheus.io/grafana/d/7gmLoNDmz/prombench?orgId=1&var-RuleGroup=All&var-pr-number=5884&from=1565972281954&to=1566117298424

tsdb/index/postings_test.go
```diff
@@ -718,6 +719,847 @@ func TestBigEndian(t *testing.T) {
...
+		binary.BigEndian.PutUint32(b, ls1[i])
+	}
+
+	// prefixCompressedPostings for ls1.
+	bufPCP1 := encoding.Encbuf{}
+	temp := make([]uint64, len(ls1))
+	for i, x := range ls1 {
+		temp[i] = uint64(x)
+	}
+	writePrefixCompressedPostings(&bufPCP1, temp)
```
krasi-georgiev: using internal functions makes it impossible to run proper benchmarks - new vs old code.
I mean the internal implementation changes quite frequently so this also means refactoring the bench test as well, making it impossible to run the same bench test on the new and the old code.

Can you implement these using only external API so that even when the internal implementation changes we can still run these  without any changes to the bench tests.

This will also allow us to use the automated /funcbench testing project that we have been working on.

codesome:
\<poke for review\>

naivewong:
add bigEndian64Postings, the Next() is worse than prefixCompressedPostings, the Seek() is similar

codesome:
Alright. You can undo the last commit in that case.

^^ @gouthamve

Lemmons:
Any chance of this being merged soon?

ipstatic:
Has progress stalled on this PR? With one of our larger Prometheus instances we are hitting the 64 gigabyte limit with Thanos Compactor.

GiedriusS (commented):
Thanks for this work! It would be cool to also implement fuzzing for this new format to make sure we haven't missed any edge cases.

I've been personally looking into S4-BP128-D4 which can also be sped up by SIMD instructions. I didn't have time to finish it, yet. I think we should try to use a format that can be easily sped up by SIMD instructions.

bwplotka:
Hello from the Bug Scrub again!

We updated the underlying issue https://github.com/prometheus/prometheus/issues/5868

Since there is not activity and we proposed a simpler fix to the underlying issue, let's close this for now. Thanks everyone!

Issue #5771, opened 2019-07-17
storage.tsdb.retention.size exceeds storage


Armatorix:
## Proposal
Prometheus flag is not working as it should. When setting storage.tsdb.retention.size=X, I'm convinced that prometheus won't use more data then X.

## Bug Report
**What did you do?**
I've setup Prometheus with flag
```
      - "--storage.tsdb.retention.size=300MB"
```
left it for some time, next, I've changed scrape interval from 15s to 1s.
So I started to generate more metrics than previously.
**What did you expect to see?**
I expected to see the storage use of Prometheus below 300MB. So I would like to see that old data be deleted as used storage comes close to limit. 

**What did you see instead? Under which circumstances?**
I see that used space is exceeded a lot. It easily takes 750MB and still growing. 
It was caused by an extreme change of config.
It probably would have similar effect if I add more instances to scrape in my service discovery.
**Environment**

* System information:
`Linux 4.15.0-54-generic x86_64`

* Prometheus version:
docker image
`prom/prometheus:v2.11.1`

* Prometheus configuration file:
```
global:
  scrape_interval: 1s
  external_labels:
      monitor: 'local'

scrape_configs:

  - job_name: 'consul'
    consul_sd_configs:
      - server: "localhost:8500"
    relabel_configs:  
    - source_labels: [__meta_consul_service]
      separator: ;
      regex: (.*)_exporter
      target_label: job
      replacement: $1
      action: replace
```

codesome:
That is possible right now because WAL size is still not considered for the size limit, but only the blocks as of now. Work is in progress here https://github.com/prometheus/tsdb/pull/651 to also consider WAL size in the storage size limit.

Additionally, the flag tells when the retention jobs have to be triggered. So even after the WAL addition above, you can expect the storage size to grow little beyond the size mentioned in the flag until the old data is removed.

Armatorix:
Emmm... So this flag is not working as you described... It's just exceeding infinitely now, not "little beyond", I've got limit of 300MB, after each 4h It's deleting only small chunks of data.
If it's working as you described that problem is probably in "retention job".

PR #5886, opened 2019-08-13
Compute WAL size and use it during retention size checks

dipack95:
Signed-off-by: Dipack P Panjabi <dpanjabi@hudson-trading.com>

Closes: https://github.com/prometheus/prometheus/issues/5771

Compute the size of the WAL and use it while calculating if exceeding the max retention size limit.

(Port of https://github.com/prometheus/tsdb/pull/651)

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -933,7 +933,11 @@ func (db *DB) beyondSizeRetention(blocks []*Block) (deleteable map[ulid.ULID]*Bl
 	}
 
 	deleteable = make(map[ulid.ULID]*Block)
-	blocksSize := int64(0)
+
+	walSize, _ := db.Head().wal.Size()
+	// Initializing size counter with WAL size,
+	// as that is part of the retention strategy.
+	blocksSize := walSize
 	for i, block := range blocks {
 		blocksSize += block.Size()
 		if blocksSize > db.opts.MaxBytes {
--- a/tsdb/db_test.go
+++ b/tsdb/db_test.go
@@ -1115,7 +1115,6 @@ func TestSizeRetention(t *testing.T) {
 
 	headBlocks := []*BlockMeta{
 		{MinTime: 700, MaxTime: 800},
-		{MinTime: 800, MaxTime: 900},
 	}
 
 	// Add some data to the WAL.
@@ -1123,18 +1122,10 @@ func TestSizeRetention(t *testing.T) {
 	for _, m := range headBlocks {
 		series := genSeries(100, 10, m.MinTime, m.MaxTime)
 		for _, s := range series {
-			var err error
-			ref := uint64(0)
 			it := s.Iterator()
 			for it.Next() {
 				tim, v := it.At()
-				if ref != 0 {
-					err := headApp.AddFast(ref, tim, v)
-					if err == nil {
-						continue
-					}
-				}
-				ref, err = headApp.Add(s.Labels(), tim, v)
+				_, err := headApp.Add(s.Labels(), tim, v)
 				testutil.Ok(t, err)
 			}
 			testutil.Ok(t, it.Err())
--- a/tsdb/db_test.go
+++ b/tsdb/db_test.go
@@ -1113,15 +1113,57 @@ func TestSizeRetention(t *testing.T) {
 		createBlock(t, db.Dir(), genSeries(100, 10, m.MinTime, m.MaxTime))
 	}
 
+	headBlocks := []*BlockMeta{
+		{MinTime: 700, MaxTime: 800},
+		{MinTime: 800, MaxTime: 900},
+	}
+
+	// Add some data to the WAL.
+	headApp := db.Head().Appender()
+	for _, m := range headBlocks {
+		series := genSeries(100, 10, m.MinTime, m.MaxTime)
+		for _, s := range series {
+			var err error
+			ref := uint64(0)
+			it := s.Iterator()
+			for it.Next() {
+				tim, v := it.At()
+				if ref != 0 {
+					err := headApp.AddFast(ref, tim, v)
+					if err == nil {
+						continue
+					}
+				}
+				ref, err = headApp.Add(s.Labels(), tim, v)
+				testutil.Ok(t, err)
+			}
+			testutil.Ok(t, it.Err())
+		}
+	}
+	testutil.Ok(t, headApp.Commit())
+
 	// Test that registered size matches the actual disk size.
-	testutil.Ok(t, db.reload())                                       // Reload the db to register the new db size.
-	testutil.Equals(t, len(blocks), len(db.Blocks()))                 // Ensure all blocks are registered.
-	expSize := int64(prom_testutil.ToFloat64(db.metrics.blocksBytes)) // Use the the actual internal metrics.
+	testutil.Ok(t, db.reload())                                         // Reload the db to register the new db size.
+	testutil.Equals(t, len(blocks), len(db.Blocks()))                   // Ensure all blocks are registered.
+	blockSize := int64(prom_testutil.ToFloat64(db.metrics.blocksBytes)) // Use the the actual internal metrics.
 	walSize, err := db.Head().wal.Size()
 	testutil.Ok(t, err)
 	// Expected size should take into account block size + WAL size
-	expSize += walSize
+	expSize := blockSize + walSize
 	actSize, err := fileutil.DirSize(db.Dir())
+	testutil.Ok(t, err)
+	testutil.Equals(t, expSize, actSize, "registered size doesn't match actual disk size")
+
+	// Create a WAL checkpoint, and compare sizes.
+	first, last, err := db.Head().wal.Segments()
+	testutil.Ok(t, err)
+	_, err = wal.Checkpoint(db.Head().wal, first, last-1, func(x uint64) bool { return false }, 0)
+	testutil.Ok(t, err)
+	blockSize = int64(prom_testutil.ToFloat64(db.metrics.blocksBytes)) // Use the the actual internal metrics.
+	walSize, err = db.Head().wal.Size()
+	testutil.Ok(t, err)
+	expSize = blockSize + walSize
+	actSize, err = fileutil.DirSize(db.Dir())
 	testutil.Ok(t, err)
 	testutil.Equals(t, expSize, actSize, "registered size doesn't match actual disk size")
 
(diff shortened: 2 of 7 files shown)


krasi-georgiev:
unless someone has a better idea, this LGTM after the conflit resolut.

dipack95:
@krasi-georgiev I've corrected the merge conflicts!

krasi-georgiev:
LGTM

ping @bwplotka @codesome

codesome:
Sorry I have not been following PRs lately. 

I had only 1 comment in the old PR
> I am wondering if it would error out if we are calculating size while files are being added or removed

dipack95:
@codesome In your scenario, are you referring to when WAL chunks are added/removed? If so, I tested for that particular scenario, and it does not crash.

codesome:
> WAL chunks are added/removed

Not chunks, but when new files are being added/removed (mostly the case of removing files while we are already iterating the items in the directory).

dipack95:
From what I can tell, the size computation logic is a blocking call, so no files are added/removed while it iterates over the contents of the `wal/` dir.

krasi-georgiev:
@codesome did you double check about this race?

codesome:
I don't think I will be able to investigate that race right now.

tsdb/db.go
```diff
@@ -890,7 +890,11 @@ func (db *DB) beyondSizeRetention(blocks []*Block) (deleteable map[ulid.ULID]*Bl
 	}
 
 	deleteable = make(map[ulid.ULID]*Block)
-	blocksSize := int64(0)
+
+	walSize, _ := db.Head().wal.Size()
```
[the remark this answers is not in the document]
cstyan: Sorry, not sure I understand what you're saying.

So, with this PR we're now going to include the size of the WAL in the retention size calculation. Let's say that after this for loop that we need to delete N blocks to stay within the configured retention size. However, the call to this function `beyondSizeRetention` in `db.Reload` we also call `db.head.Truncate`, which will delete some WAL segments.

What I'm trying to point out is that after truncating the wall we may have only needed to remove less than N blocks to remain within size retention.

WAL segments are 128M, looking at the size of on disk data for some of our Prometheus instances that's the size of at least 2 blocks.

I think either deleting blocks needs to happen after truncating the WAL, or size retention should not account for WAL size. The retention flags are meant for storage of metrics long term on disk from my understanding.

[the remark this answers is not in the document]
codesome: > after truncating the wal we may have only needed to remove less than N blocks to remain within size retention

The analysis sounds right. Quickly looking at the code it should be simple to take care of that. The `reload()` call for head compaction can be made into 2 `reload()` calls, one before and after `Truncate(...)`, with the first call skipping the size retention.

> The retention flags are meant for storage of metrics long term on disk from my understanding.

I realize that we have not discussed a lot about what the flag is meant to be. My understanding was it should be any disk space that the Prometheus is using, and WAL can grow very big so it won't be good to leave that out of equation especially for small setups with limited storage. @krasi-georgiev thoughts on this?

codesome:
Had a look now, the possibility of a race is very slim here and not critical. There wont be any issue with deletion of segments as it doesn't happen while the size is being computed (as size of WAL is checked inside `reload()` which doesn't happen in parallel with WAL truncation). And addition of WAL segment should not be a problem for `filepath.Walk` (I guess?).

> the size computation logic is a blocking call, so no files are added/removed while it iterates over the contents of the `wal/` dir.

Having gone through the souce of `filepath.Walk`, not sure if it is blocking to changes on the disk. But not a problem in this case now.

tsdb/db.go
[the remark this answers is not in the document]
krasi-georgiev: yes logically the flag should control the overall size of HDD that Prometheus uses.

@cstyan thanks for mentioning the WAL logic. @dipack95 We should definitely add a test for this. 
Something like: generate enough wal files to cause a checkpoint and check exp vs act disk size.
With the wal size option you can set it small enough to cause creating few WAL files that a reload will cause a wal checkoint.

krasi-georgiev:
@dipack95 just need to add a test that involves checkpointing and it is ready for merging.

https://github.com/prometheus/prometheus/pull/5886/files#r327057583

tsdb/db.go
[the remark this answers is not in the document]
dipack95: So the test could be something like:
1. Write a bunch of records to the WAL
2. Note size of WAL
3. Create a checkpoint
4. Truncate WAL segments that have been checkpointed
5. Check to see if the new current WAL size, and WAL size at step 2 are equal?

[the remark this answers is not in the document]
krasi-georgiev: yes something like that. 

1. `wal.NewSize` and add records using `wal.Log`
3. Check that there are enough wal files to trigger Truncation. Can use the `prometheus_tsdb_wal_segment_current` metric.
4. Check the wal size - act vs exp
5. Run wal.Trucate
4. Check that wal.Trucate happened. (can use the `prometheus_tsdb_wal_truncations_total` metric)
4. Check that `prometheus_tsdb_wal_truncations_failed_total` is zero.
4. Check the wal size again - act vs exp


for using the metrics to check the state can borrow the code for a test I did recently in
https://github.com/thanos-io/thanos/pull/1587/files#diff-e97a3cbd2df9c75cfeb6ff231a4ef37fR78-R98

krasi-georgiev:
can you check if my suggestion would remove all that boilerplate than?

krasi-georgiev:
ping @dipack95

dipack95:
Hey sorry, I haven't had the time for the last week. I should have the time to update this, and #5887 in a few days.

krasi-georgiev:
Thanks.  Appreciated

krasi-georgiev:
ping @dipack95

dipack95:
@krasi-georgiev When I create a `Head`, and try to go about it that way, it runs into an import cycle. Looks like we might have to keep the test the way it is right now, ugly as it is!

dipack95:
@krasi-georgiev That makes sense, I've added the WAL size to the expected size as well.

However, I've noticed that over the course of that test, it directly writes blocks to the disk instead of using the Head (and consequently the WAL), and thus the WAL size is always zero. Thus the test never failed with the introduction of this PR.

krasi-georgiev:
yep you will need to add some records to the wal as well. Look at some other tests for ideas.

dipack95:
@krasi-georgiev Please take a look, I've updated the test.

tsdb/wal/wal_test.go
```diff
@@ -475,3 +481,75 @@ func BenchmarkWAL_Log(b *testing.B) {
 		})
 	}
 }
+
+func TestWAL_Size(t *testing.T) {
```
krasi-georgiev: do we still need this test?

dipack95: You're right, this one does do an additional truncate, but that isn't really something that needs to be tested here. I'll remove it.

tsdb/db_test.go
```diff
@@ -1111,11 +1113,58 @@ func TestSizeRetention(t *testing.T) {
 		createBlock(t, db.Dir(), genSeries(100, 10, m.MinTime, m.MaxTime))
 	}
 
+	headBlocks := []*BlockMeta{
+		{MinTime: 700, MaxTime: 800},
+		{MinTime: 800, MaxTime: 900},
```
krasi-georgiev: this doesn't create 2 wal files.

you can use with the WALSegmentSize to ensure that appending created more than a single file and it will best to add to the test that after appending to the wal it results in 2 files on disk.

dipack95: Any particular reason you think that having more WAL files will help the test? I don't expect the behaviour of the WAL size calculations will change as such, considering when creating the checkpoint it takes into account multiple files and dirs

krasi-georgiev: oh, I thought you have headBlocks multiple heads because you want to test the checkpointing with multiple WAL files

So why do you have 2 headBlocks than?

dipack95: No particular reason, although I do see now how that could cause some confusion. Do you think I should bring it down to just 1 head block?

krasi-georgiev: yep, and probbaly no need to use addfast for such small amount of samples.

dipack95:
@krasi-georgiev Please take a look now.

krasi-georgiev:
LGTM, Thanks

unless @codesome , @brian-brazil or anyone else has any other comments will merge this in few days.

life0215:
Not sure if it's only me finding that the WAL size is still not used during retention size checks:

```sh
/prometheus $ prometheus --version
prometheus, version 2.17.1 (branch: HEAD, revision: ae041f97cfc6f43494bed65ec4ea4e3a0cf2ac69)
  build user:       root@806b02dfe114
  build date:       20200326-16:18:19
  go version:       go1.13.9
/prometheus $ 
/prometheus $ ps
PID   USER     TIME  COMMAND
    1 1000      8d02 /bin/prometheus --web.console.templates=/etc/prometheus/consoles --web.console.libraries=/etc/prometheus/console_libraries --storage.tsdb.retention.size=15GB --config.file=/etc/prometheus/config_out/prometheus.env.yaml --
   79 1000      0:00 sh
   87 1000      0:00 sh
   99 1000      0:00 ps
/prometheus $ du -h -d 1
79.5M   ./01EB7D47PPV49QCNY2MZVJCYTE.tmp
524.5M  ./01EBBR0J3VHNPNBYM6CZXM56T9.tmp
62.1M   ./01EJFJF03NFF5J29VE9JKEE2MS
512.5M  ./01EBBPV9FS1Z0QY4EK9QC7RMQA.tmp
1.2G    ./01EJ1DCVBP2C3XS5WVJE6H85AE
1.0G    ./01EHG166DB6JYRB79VEMWME7Z6
158.7M  ./01EJFJG3Y1CQ3PRRPDQH3NACNQ
512.5M  ./01EBAM8DHP0GAKCWWS6P3F7D0E.tmp
1.1G    ./01EHVM027KCG4CAT3VGZW6VMJ6
516.5M  ./01EBBQ7Q7CG0J99RHM57FE0TF0.tmp
57.5M   ./01EJFSAQGFA3PT9BW8JMDKN824
1.0G    ./01EGYN0SAP6C43V8RF335SNFFW
1.1G    ./01EHA7SGSXJ4CR88GQ4B936SNZ
79.5M   ./01EB7CYVEXJENABPGR037E14P8.tmp
453.4M  ./01EJEXYPAFCM86Y7T5REJXNJHA
8.0K    ./01EBCB2SVA5XVV84EC54GR2S0T.tmp
8.0K    ./01EBBS9M0417RNJVVTCYEK33RH.tmp
1.3G    ./01EJD06W4FTYPFAKYVM7FZNE01
1.1G    ./01EH4EE1ZHCXEWP80NJJJJFSW5
516.5M  ./01EBAM3540NXPQM49TSAPHSDBH.tmp
1.1G    ./01EHNTJTJ7CGDRPSB4H6TA7YE0
516.5M  ./01EBATZ99KJ4PX2FG44TW0SZAE.tmp
1.0G    ./01EGRVM1JBVNQWD1Y7N5FYN4GN
1.9G    ./wal
1.2G    ./01EJ76T4CNYNDX6EJ5KMWXPKMZ
16.8G   .
```

Please note that we've set **--storage.tsdb.retention.size=15GB**, but the actual storage usage is 16.8G

codesome:
I dont think the `*.tmp` directories count towards the size calculation. If they are leftovers from any errors before, they can be deleted. Ideally we should have a cleanup of those temporary directories on startup if any exists.

PR #5893, opened 2019-08-14
Removed extra tsdb/testutil after merge.

bwplotka:


codesome:
While this does remove some duplicate code, it also makes tsdb depend on prometheus now and is no longer independent. I guess that was going to happen after the merger anyway.

bwplotka:
Mhm only on testing helpers, so it is ok I think.

brancz:
Package encapsulation still holds. I would deem this ok.

krasi-georgiev:
LGTM

PR #5897, opened 2019-08-15
tsdb: refactor PostingsForMatchers()

YaoZengzeng:
Actually some code in `PostingsForMatchers()` is redundant or never to be run and some comments are misleading.

So refactor it to make it clean.

### DIFF
--- a/tsdb/querier.go
+++ b/tsdb/querier.go
@@ -328,6 +328,7 @@ func findSetMatches(pattern string) []string {
 func PostingsForMatchers(ix IndexReader, ms ...labels.Matcher) (index.Postings, error) {
 	var its, notIts []index.Postings
 	// See which label must be non-empty.
+	// Optimization for case like {l=~".", l!="1"}.
 	labelMustBeSet := make(map[string]bool, len(ms))
 	for _, m := range ms {
 		if !m.Matches("") {
@@ -336,9 +337,9 @@ func PostingsForMatchers(ix IndexReader, ms ...labels.Matcher) (index.Postings, 
 	}
 
 	for _, m := range ms {
-		matchesEmpty := m.Matches("")
-		if labelMustBeSet[m.Name()] || !matchesEmpty {
+		if labelMustBeSet[m.Name()] {
 			// If this matcher must be non-empty, we can be smarter.
+			matchesEmpty := m.Matches("")
 			nm, isNot := m.(*labels.NotMatcher)
 			if isNot && matchesEmpty { // l!="foo"
 				// If the label can't be empty and is a Not and the inner matcher
@@ -444,7 +445,7 @@ func postingsForMatcher(ix IndexReader, m labels.Matcher) (index.Postings, error
 	return index.Merge(rit...), nil
 }
 
-// inversePostingsForMatcher eeturns the postings for the series with the label name set but not matching the matcher.
+// inversePostingsForMatcher returns the postings for the series with the label name set but not matching the matcher.
 func inversePostingsForMatcher(ix IndexReader, m labels.Matcher) (index.Postings, error) {
 	tpls, err := ix.LabelValues(m.Name())
 	if err != nil {


YaoZengzeng:
@brian-brazil PTAL

tsdb/querier.go
```diff
@@ -364,7 +350,9 @@ func PostingsForMatchers(ix IndexReader, ms ...labels.Matcher) (index.Postings,
 				}
 				its = append(its, it)
 			}
-		} else { // l=""
+		} else { // l!="foo" or l=""
```
codesome: As we are using `inversePostingsForMatcher` for `l!="foo"` here (instead of `postingsForMatcher` above), this could make queries including `l!="foo"` to perform worse than before.

But looking at the refactoring, it seems that the code always fell back here for `l!="foo"`.

Can you add another `} else if isNot {...}` case here which includes the above code that was removed (which includes `postingsForMatcher`)?

YaoZengzeng: @codesome Make sense, updated.

tsdb/querier.go
```diff
@@ -326,50 +326,39 @@ func findSetMatches(pattern string) []string {
 // PostingsForMatchers assembles a single postings iterator against the index reader
 // based on the given matchers.
 func PostingsForMatchers(ix IndexReader, ms ...labels.Matcher) (index.Postings, error) {
+	var err error
```
codesome: Can you combine these variables inside a single `var (...)`

tsdb/querier.go
```diff
@@ -326,50 +326,39 @@ func findSetMatches(pattern string) []string {
...
-				its = append(its, it)
+				it, err = postingsForMatcher(ix, m)
 			}
-		} else { // l=""
+			if err != nil {
+				return nil, err
+			}
+			its = append(its, it)
+		} else { // l!="foo" or l=""
+			// If the label can be empty, subtract all not matched series out.
```
codesome: These comments can be put individually inside the corresponding `if` and `else` blocks.

YaoZengzeng:
@codesome Updated.

codesome:
I would like another maintainer to verify this refactoring before merging.

tsdb/querier.go
```diff
@@ -326,50 +326,41 @@ func findSetMatches(pattern string) []string {
 // PostingsForMatchers assembles a single postings iterator against the index reader
 // based on the given matchers.
 func PostingsForMatchers(ix IndexReader, ms ...labels.Matcher) (index.Postings, error) {
-	var its, notIts []index.Postings
-	// See which label must be non-empty.
-	labelMustBeSet := make(map[string]bool, len(ms))
```
brian-brazil: Removing this removes an optimisation. We want to be smart for things like {a=~".",a!="1"}

YaoZengzeng: Thanks for reminding me. Finally I know why the code design like this 😂 

If we could specify same label for multiple times, then everything make sense.

But if query like `{a="1", a="2"}` reach here, then the result will be incorrect, so PromQL will filter query like this ?

Hmm.. then this PR seems unnecessary, maybe could add some comment to make the code more clear 😄

brian-brazil: The label must be set in that case, though it'll match nothing.

YaoZengzeng: Yes, forget the final intersect :)

YaoZengzeng:
@brian-brazil Add some comment and refactor a little to make the code more readable.

PTAL if it make sense.

tsdb/querier.go
```diff
@@ -336,9 +337,9 @@ func PostingsForMatchers(ix IndexReader, ms ...labels.Matcher) (index.Postings,
 	}
 
 	for _, m := range ms {
-		matchesEmpty := m.Matches("")
-		if labelMustBeSet[m.Name()] || !matchesEmpty {
+		if labelMustBeSet[m.Name()] {
```
brian-brazil: You're changing the logic here, which will be slower

YaoZengzeng: If `labelMustBeSet[m.Name()]` is true, then `!matchesEmpty` is not need to judge.

If `labelMustBeSet[m.Name()]` is false, then `!matchesEmpty` must be false.

The judgment of `!matchesEmpty` is redundant, so the refactor is fine and make the code more clear. WDYT?

brian-brazil: That makes sense.

brian-brazil:
:+1:

YaoZengzeng:
@brian-brazil This PR could get merged? Otherwise I would close it 😂

PR #5900, opened 2019-08-15
cleanup tsdb/head.go, correct misspelled words

beautytiger:
Signed-off-by: Guangming Wang <guangming.wang@daocloud.io>

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -773,7 +773,7 @@ func (h *Head) Appender() Appender {
 func (h *Head) appender() *headAppender {
 	return &headAppender{
 		head: h,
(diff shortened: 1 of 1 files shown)


tsdb/head.go
```diff
@@ -1665,7 +1665,7 @@ func (s *memSeries) chunksMetas() []chunks.Meta {
 	return metas
 }
 
-// reset re-initialises all the variable in the memSeries except 'lset', 'ref',
+// reset re-initializes all the variable in the memSeries except 'lset', 'ref',
```
tariq1890: This is not really a typographical error.

ethan-daocloud: well learned, thank you!

PR #5902, opened 2019-08-15
[prometheus_sd/kubernetes]add new node address types for discover

tariq1890:
Please see here: https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-apiserver/app/options/options.go#L104

### DIFF
--- a/discovery/kubernetes/node.go
+++ b/discovery/kubernetes/node.go
@@ -196,9 +196,11 @@ func (n *Node) buildNode(node *apiv1.Node) *targetgroup.Group {
 
 // nodeAddresses returns the provided node's address, based on the priority:
 // 1. NodeInternalIP
-// 2. NodeExternalIP
-// 3. NodeLegacyHostIP
-// 3. NodeHostName
+// 2. NodeInternalDNS
+// 3. NodeExternalIP
+// 4. NodeExternalDNS
+// 5. NodeLegacyHostIP
+// 6. NodeHostName
 //
 // Derived from k8s.io/kubernetes/pkg/util/node/node.go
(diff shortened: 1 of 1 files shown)


tariq1890:
@brian-Brazil. Requesting your review

discovery/kubernetes/node.go
```diff
@@ -207,9 +209,15 @@ func nodeAddress(node *apiv1.Node) (string, map[apiv1.NodeAddressType][]string,
 		m[a.Type] = append(m[a.Type], a.Address)
 	}
 
+	if addresses, ok := m[apiv1.NodeInternalDNS]; ok {
```
brian-brazil: I think we should prefer IP to DNS, as that avoids having to do additional DNS lookups.

tariq1890: Oh yes, that makes sense. Will do.

discovery/kubernetes/node.go
```diff
@@ -207,9 +209,15 @@ func nodeAddress(node *apiv1.Node) (string, map[apiv1.NodeAddressType][]string,
...
+		return addresses[0], m, nil
+	}
 	if addresses, ok := m[apiv1.NodeInternalIP]; ok {
 		return addresses[0], m, nil
 	}
+	if addresses, ok := m[apiv1.NodeExternalDNS]; ok {
```
brancz: same here

tariq1890:
Review comments addressed.

brancz:
lgtm :+1: leaving last call up to @brian-brazil

brian-brazil:
You've got a stray file here.

tariq1890:
Thanks for catching that @brian-brazil!

brian-brazil:
Thanks!

PR #5904, opened 2019-08-16
template: remove redundant parentheses.

johncming:
Remove unnecessary parentheses to achieve uniform format.


brian-brazil:
Thanks!

PR #5905, opened 2019-08-16
rules: Use Camel format

johncming:
Use Camel format.

### DIFF
--- a/rules/alerting.go
+++ b/rules/alerting.go
@@ -21,7 +21,7 @@ import (
 	"sync"
 	"time"
 
-	html_template "html/template"
+	htmlTemplate "html/template"
 
 	yaml "gopkg.in/yaml.v2"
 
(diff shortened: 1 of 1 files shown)


codesome:
I don't think using camel cases for package names is the convention. https://blog.golang.org/package-names and some other blog posts tell me that using `_` is also not a good practice in Go. `html_template` should be just `htmltemplate`, and similarly many other renames in this codebase (for example `tsdb_errors`).

johncming:
> I don't think using camel cases for package names is the convention. https://blog.golang.org/package-names and some other blog posts tell me that using `_` is also not a good practice in Go. `html_template` should be just `htmltemplate`, and similarly many other renames in this codebase (for example `tsdb_errors`).

Yes, you're right, both camel and underline are not recommended in go.

PR #5908, opened 2019-08-16
Regulate the package name.

johncming:
Regulate the package name.

There are two types of non-standard package names (https://blog.golang.org/package-names):

1. myPackage
2. my_package

There is also the case where the aliases for the same package are different.

Modified principle：

1. myPackage  -> mypackage
2. my_package -> mypackage

There are exceptions:

`github.com/prometheus/common/config` use `commonconfig` instead of `configutil`, because `configutil` is more confusing.

See diff results for more detailed modifications

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -48,10 +48,10 @@ import (
 	promlogflag "github.com/prometheus/common/promlog/flag"
 	"github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/discovery"
-	sd_config "github.com/prometheus/prometheus/discovery/config"
+	sdconfig "github.com/prometheus/prometheus/discovery/config"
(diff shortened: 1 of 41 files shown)


simonpasquier:
I'm all for consistency but if it isn't enforced by the tooling, it will drift very quickly. Do you see any way that it could be checked automatically?

johncming:
> I'm all for consistency but if it isn't enforced by the tooling, it will drift very quickly. Do you see any way that it could be checked automatically?

There are no automated tools, and the common golint and staticchecks have no special checks.

beorn7:
First of all, apologies for letting this become stale.

We just discussed this in the bug scrub and agreed that we only would like to get this consistent if it's something that `golint` and friends detect. Since this is passing by common linters, we don't want to ban it.

PR #5911, opened 2019-08-17
tsdb: Put constants under the package.

johncming:
Put constants under the package.

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -57,6 +57,13 @@ var (
 	emptyTombstoneReader = newMemTombstones()
 )
 
+const (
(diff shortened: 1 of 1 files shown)


brian-brazil:
It's a heuristic rather than an invariant, nothing would break break if it were different.

johncming:
> Thanks for looking into this. What are the reasons for this change? If the constant is used only in single place, it's nice to have it scoped.
> 
> However, it's true that this is quite important constant and invariant we have in other places as well. Can we make this constant then to `chunkenc` near `Xor` implementation? (`xor.go`). If anything that should belong to `Xor` code.

The semantics of how many samples each chunk looks like are more appropriate for the package than for the `memseries.append` function.

PR #5915, opened 2019-08-18
Fixes service-discovery view when accessed from unhealthy targets

### LINKED ISSUE #5677: When you pick "unhealthy" in targets the service discovery page is empty
## Proposal

we want to see SD even when on the other page we pick unhealty

## Bug Report
**What did you do?**

Pick unhealthy in targets
Go to service discovery page

The service discovery details

Empty page

**Environment**

* Prometheus version:

2.9.2

Harkishen-Singh:
This PR fixes #5677 

**Explanation**
To retain the previous screen (reported by #4599 ), PR [link](https://github.com/prometheus/prometheus/pull/5201) was merged. However, this did not add any check for the current `url` and hence assigned `display: none` to the `table-container` which should only be done in the `targets` screen and not in the `service-discovery` screen. 

This PR fixes the issue by adding the URL check to find out in which screen the current view-port is and based on that access the `localStorage`.

**Before fixing the bug [gif]**
before

**After fixing the bug [gif]**
after

Signed-off-by: Harkishen Singh harkishensingh@hotmail.com


Harkishen-Singh:
please review @juliusv @simonpasquier @brian-brazil

web/ui/static/js/targets.js
```diff
@@ -21,12 +21,16 @@ function showUnhealthy(_, container) {
 }
 
 function init() {
-  if (!localStorage.selectedTab || localStorage.selectedTab == "all-targets"){
-    $("#all-targets").parent().addClass("active");
+  if (document.URL.includes('targets')) {
```
simonpasquier: Rather that checking the URL, it should be possible to check `$("#unhealthy-targets").length`: if > 0 then apply `localStorage.selectedTab` preferences otherwise show everything. What do you think?

Harkishen-Singh: sure. that would work in a better way too. Thanks for the suggestion. Implementing..

Harkishen-Singh:
@simonpasquier done! please have a look

Harkishen-Singh:
@simonpasquier  Done!

simonpasquier:
Thanks! It could be merged on green CI.

Issue #5529, opened 2019-05-02
Random value in recording rule with multiple identical series result


FUSAKLA:
## Bug Report
**What did you do?**
Not sure if proposal of bug report but when recording rule expression results to multiple series with different values and has the `labels` section specified which will cause the two series having exact same labels. This results to recorded serie which will have randomly value of one of those values because of random ordering.

Simple example:
```yaml
 groups:
     - name: "test"
       interval: "1s"
       rules:
         - record: test_metric
           expr: "0"
           labels:
               label: "value"
         - record: "test_metric"
           expr: "1"
         - record: "result"
           expr: "max(test_metric) by (label)"
           labels:
               label: "value"
```

Will result to:
![image](https://user-images.githubusercontent.com/6112562/57088855-73677400-6d03-11e9-840e-82fab955c964.png)

I understand that it is users responsibility to make sure the resulting metrics won't be identical, but on the other hand I'd expect this to result as a failed rule evaluation or at least log something or increment metric about it?
This way user has no notice that the data he has recorded are invalid.

brian-brazil:
What version of Prometheus are you using?

FUSAKLA:
Sorry forgot to mention. 

It's current master 03b90b46

brian-brazil:
The issue here is that your configuration is invalid, as you're producing two identical time series. That's checked down in PromQL, but not for the label addition in rules.

FUSAKLA:
Yes, this is what I'm aware of as I wrote in the issue description, I know that it produces invalid data. 

The thing is first of all there is no warning in case of recording rules the same way it's for the label replace in the docs
> Care must be taken with labeldrop and labelkeep to ensure that metrics are still uniquely labeled once the labels are removed.

But the main thing is that Prometheus decides to use random value without letting user know about this. This is my main concern. Sorry if I didn't make myself clear before.

Transwert:
@brian-brazil @FUSAKLA can any of you close this issue, as this has been solved,

brian-brazil:
This issue has not been resolved.

Transwert:
@brian-brazil isn't this PR : https://github.com/prometheus/prometheus/pull/6025 solved the issue, and if it doesn't ,can you brief out about possible correction needed which can be implemented though!

brian-brazil:
That added a warning about something that can cause this to promtool. We still need to check for this and error out in the recording rule code.

rothsa:
[...]

PR #5916, opened 2019-08-18
Fixes issue #5529

JBerny:
This is supposed to fix [Issue 5529] (https://github.com/prometheus/prometheus/issues/5529) by checking that rules are not repeated in the same group,

### DIFF
--- a/pkg/rulefmt/rulefmt.go
+++ b/pkg/rulefmt/rulefmt.go
@@ -18,6 +18,7 @@ import (
 	"io/ioutil"
 	"strings"
 	"time"
+	"fmt"
 
 	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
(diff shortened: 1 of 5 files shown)


brian-brazil:
This fix isn't right, this is a semantic issue rather than a syntactic one. This can only be caught at runtime.

JBerny:
Isn't the semantic expressed through syntax?

As far as I understand from the issue, the user is warning that Prometheus is behaving in an unsafe way because it allows the same record rule to be set multiple times and then takes a random value when querying. I thought it would be better to prevent this unpredictable behavior at the root.

But if you think this can be done in a better way, and if you would like to give me some more directions, I'm happy to help.

brian-brazil:
This is something that you can't determine the correctness of purely by inspection, you can only tell once you have the result of the query and then see the clashes.

JBerny:
Ok as far as I understand, the syntax is actually valid. It is allowed to specify the same metric multiple times with different labels. 
The problem is that at the moment of the evaluation of the last rule cannot return a consistent result. Probably time series are not sorted and are stored in a map. I will make another attempt.

PR #5924, opened 2019-08-19
Show the warnings during label query

### LINKED ISSUE #5885: promtool: show warnings for label value queries
An update of the Go client library dependency done in https://github.com/prometheus/prometheus/pull/5883#pullrequestreview-274297321 now provides warnings for label value queries. Figure out if / how to display those in promtool.

sayanchowdhury:
This patch loops through the warnings while querying the label and spits the
output to stderr.

Fixes #5885

Signed-off-by: Sayan Chowdhury <sayan.chowdhury2012@gmail.com>

### DIFF
--- a/cmd/promtool/main.go
+++ b/cmd/promtool/main.go
@@ -501,9 +501,12 @@ func QueryLabels(url *url.URL, name string, p printer) int {
 	// Run query against client.
 	api := v1.NewAPI(c)
 	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
-	// TODO: Show warnings (second return value) somehow? See https://github.com/prometheus/prometheus/issues/5885
-	val, _, err := api.LabelValues(ctx, name)
+	val, warn, err := api.LabelValues(ctx, name)
 	cancel()
+
+	for _, v := range warn {
+		fmt.Fprintln(os.Stderr, "query warning:", v)
+	}
 
 	if err != nil {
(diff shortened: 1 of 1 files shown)


sayanchowdhury:
@juliusv can you have a look at the PR?

cmd/promtool/main.go
```diff
@@ -501,10 +501,15 @@ func QueryLabels(url *url.URL, name string, p printer) int {
 	// Run query against client.
 	api := v1.NewAPI(c)
 	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
-	// TODO: Show warnings (second return value) somehow? See https://github.com/prometheus/prometheus/issues/5885
-	val, _, err := api.LabelValues(ctx, name)
+	val, warn, err := api.LabelValues(ctx, name)
 	cancel()
 
+	if warn != nil {
```
juliusv: How about omitting the `if` clause completely because the loop over a `nil` or empty slice will just do nothing anyway.

juliusv:
:+1: besides the comment, thanks!

cmd/promtool/main.go
sayanchowdhury: Yeah, makes sense.

sayanchowdhury:
@juliusv i've updated the PR.

juliusv:
:+1: Thanks!

PR #5928, opened 2019-08-21
bugfix:  The read operation uses a read lock

johncming:
This is a read operation and the lock should use a read lock.


roidelapluie:
LGTM

brian-brazil:
Thanks!

PR #5937, opened 2019-08-22
[prometheus_sd/kubernetes] add labels to the client-go req metrics

tariq1890:
Using all of the provided interface parameters for added dimensionality.

### DIFF
--- a/discovery/kubernetes/client_metrics.go
+++ b/discovery/kubernetes/client_metrics.go
@@ -37,7 +37,7 @@ var (
 			Name:      "http_request_total",
 			Help:      "Total number of HTTP requests to the Kubernetes API by status code.",
 		},
-		[]string{"status_code"},
+		[]string{"status_code", "method", "host"},
 	)
 	clientGoRequestLatencyMetricVec = prometheus.NewSummaryVec(
 		prometheus.SummaryOpts{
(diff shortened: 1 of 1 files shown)


simonpasquier:
Do you have an idea of how many additional timeseries it will create?

tariq1890:
Hmmmm, I can do a test run against my k8s environment and get back on what I find.

discovery/kubernetes/client_metrics.go
```diff
@@ -46,7 +46,7 @@ var (
 			Help:       "Summary of latencies for HTTP requests to the Kubernetes API by endpoint.",
 			Objectives: map[float64]float64{},
 		},
-		[]string{"endpoint"},
+		[]string{"verb", "endpoint"},
```
brian-brazil, line 49: Why is this verb here and method above? Will it ever not be GET?

beorn7:
Currently going through old PRs in the bug scrub...

@tariq1890 did you do the test runs?

Let's discuss this in a bit more detail. @brancz do you perhaps have an opinion?

discovery/kubernetes/client_metrics.go
tariq1890: The labels are merely following the naming patterns of the arguments in the Client-go interfaces.

You do bring up a good point though; it might be prudent to stick to a standardised name for `verb`/`method`

tariq1890:
@beorn7 Thanks for following up. I haven't got around to doing this yet. Looking forward to others' thoughts though.

discovery/kubernetes/client_metrics.go
brancz: I think this should be fixed upstream.

PR #5940, opened 2019-08-24
promql: Modify the order of return values.

johncming:
Modify the order of return values and return value.

https://github.com/prometheus/prometheus/blob/04cb37fea8a05b1c2287d128e63c06b80c12865d/promql/query_logger.go#L78

1. Error is usually the last of several return values
2.  It's more common that slice returns nil in error case.
3. format import orders(goimports).

### DIFF
--- a/promql/query_logger.go
+++ b/promql/query_logger.go
@@ -15,14 +15,15 @@ package promql
 
 import (
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thanks!

Issue #5901, opened 2019-08-15
Flaky TestFileSD test


bwplotka:
e.g https://circleci.com/gh/prometheus/prometheus/14622

```
?   	github.com/prometheus/prometheus/discovery/ec2	[no test files]
--- FAIL: TestFileSD (16.64s)
    file_test.go:113: Unexpected target groups [], we expected a failure here.
FAIL
FAIL	github.com/prometheus/prometheus/discovery/file	16.661s
```

YaoZengzeng:
I find this flaky bug yesterday as well, let me dig into it.

simonpasquier:
I had work in progress to fix it, I'll send a PR very soon.

PR #5942, opened 2019-08-24
discovery/file: Fix the bug mentioned in #5901

johncming:
fix the bug mentioned in #5901 

## Root Cause

read an empty file, no error, return an empty `[]*targetgroup.Group`

https://github.com/prometheus/prometheus/blob/ec94df49d45baa8e680d81e732294eb04fff00de/discovery/file/file.go#L368

## Why this error doesn't happen very often

Because of the following code:

https://github.com/prometheus/prometheus/blob/ec94df49d45baa8e680d81e732294eb04fff00de/discovery/file/file_test.go#L70

If out channel action occurs earlier , there is no error. But if out channel later, there is an error.

https://github.com/prometheus/prometheus/blob/ec94df49d45baa8e680d81e732294eb04fff00de/discovery/file/file_test.go#L113

### DIFF
--- a/discovery/file/file.go
+++ b/discovery/file/file.go
@@ -367,6 +367,10 @@ func (d *Discovery) readFile(filename string) ([]*targetgroup.Group, error) {
 		return nil, err
 	}
 
+	if len(content) == 0 {
+		return nil, errors.New("empty file")
+	}
+
 	info, err := fd.Stat()
 	if err != nil {
 		return nil, err


simonpasquier:
See also https://github.com/prometheus/prometheus/pull/5948. I've taken a different approach and decided to fix the tests rather than modify the main code.

johncming:
> See also #5948. I've taken a different approach and decided to fix the tests rather than modify the main code.

Although the test case is a bit long, however the scenario of testing the code is also a real scenario.

roidelapluie:
This pull request can be closed as #5948 has been merged.

PR #5943, opened 2019-08-24
pkg/rulefmt: check annotations under alert rule

johncming:
Annotations are checked only in the alert rule case.

https://github.com/prometheus/prometheus/blob/38d518c0fe8abeeb8dcdb0472510cece3dabfd0e/pkg/rulefmt/rulefmt.go#L140-L144

### DIFF
--- a/pkg/rulefmt/rulefmt.go
+++ b/pkg/rulefmt/rulefmt.go
@@ -137,9 +137,11 @@ func (r *Rule) Validate() (errs []error) {
 		}
 	}
 
-	for k := range r.Annotations {
-		if !model.LabelName(k).IsValid() {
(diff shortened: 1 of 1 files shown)


pkg/rulefmt/rulefmt.go
```diff
@@ -137,9 +137,11 @@ func (r *Rule) Validate() (errs []error) {
 		}
 	}
 
-	for k := range r.Annotations {
-		if !model.LabelName(k).IsValid() {
-			errs = append(errs, errors.Errorf("invalid annotation name: %s", k))
+	if r.Alert != "" {
```
simonpasquier: this isn't really needed since L119 checks that recording rules have 0 annotations.

johncming: yes. thanks

PR #5946, opened 2019-08-25
Omit honor_timestamps directive when false

gebn:
This PR makes the `honor_timestamps` directive consistent with others in having the `omitempty` flag. This means `honor_timestamps: false`, which is the default, will be omitted from config files.


brian-brazil:
Unlike the other fields, false is not the default for this one so it should be displayed.

gebn:
Head -> desk. Thanks @brian-brazil

PR #5947, opened 2019-08-26
fix-up-typo prometheus

tanjunchen:


simonpasquier:
Thanks for the PR. Files in the the vendor/ and web/ui/static directories shouldn't be modified as they're copied from external repositories.

tanjunchen:
> Thanks for the PR. Files in the the vendor/ and web/ui/static directories shouldn't be modified as they're copied from external repositories.

Thanks

PR #5948, opened 2019-08-26
discovery/file: fix flaky tests

### LINKED ISSUE #5901: Flaky TestFileSD test

simonpasquier:
Closes #5901 

This change refactors the test code to make it more deterministic. It also implements different tests to check different cases (initial update, update, deletion, ...) rather than having a big function combining everything.


discovery/file/file_test.go
```diff
@@ -15,158 +15,456 @@ package file
...
+		T:       t,
+		dir:     tmpDir,
+		ch:      make(chan []*targetgroup.Group),
+		done:    make(chan struct{}),
+		stopped: make(chan struct{}),
+		tgs:     make(map[string]*targetgroup.Group),
 	}
+}
+
+// copyFile copies atomically a file to the runner's directory.
```
brian-brazil: atomically copies

discovery/file/file_test.go
```diff
@@ -15,158 +15,456 @@ package file
...
+
+	f, err := os.Open(src)
+	testutil.Ok(t, err)
+
+	_, err = io.Copy(newf, f)
+	testutil.Ok(t, err)
+	testutil.Ok(t, f.Close())
+
+	dst = filepath.Join(t.dir, filepath.Base(dst))
+	err = os.Rename(newf.Name(), dst)
```
brian-brazil: Aren't you assuming here that the tmpdir and dst are on the same filesystem?

discovery/file/file_test.go
```diff
@@ -15,158 +15,456 @@ package file
...
 	}
-	defer newf.Close()
+	ctx, cancel := context.WithCancel(context.Background())
+	t.cancelSD = cancel
+	go func() {
+		NewDiscovery(
+			&SDConfig{
+				Files: files,
+				// Setting a high refresh interval to make sure that the tests only
+				// relies on file watches.
```
brian-brazil: rely

discovery/file/file_test.go
simonpasquier: They should be since the temporary file is created in `t.dir` which itself is located in the default temporary directory.

simonpasquier:
@brian-brazil PTAL

discovery/file/file_test.go
```diff
@@ -15,158 +15,456 @@ package file
...
+		T:       t,
+		dir:     tmpDir,
+		ch:      make(chan []*targetgroup.Group),
+		done:    make(chan struct{}),
+		stopped: make(chan struct{}),
+		tgs:     make(map[string]*targetgroup.Group),
 	}
+}
+
+// copyFile atomically copies a file to the runner's directory.
```
cstyan: if the it's always the runners directory can't `filepath.Base(tc.dir)` just be a value on `testRunner`?

discovery/file/file_test.go
```diff
@@ -15,158 +15,456 @@ package file
...
+	b2, err := json.Marshal(got)
+	if err != nil {
+		panic(err)
+	}
+
+	testutil.Equals(t, string(b1), string(b2))
+}
+
+// validTg() maps to fixtures/valid.{json,yml}.
+func validTg(file string) []*targetgroup.Group {
```
cstyan: Whats the motivation behind checking the target groups via a function like this rather than expected return value vars in a test case struct?

discovery/file/file_test.go
cstyan: it's not clear without looking at the calls to `copyFile` that `dst` is always `t.dir`

discovery/file/file_test.go
simonpasquier: This isn't checking the target groups, just returning a `[]*targetgroup.Group`. The reason is that the `fileSDFilepathLabel` label depends on the location of the SD file so it is tedious to write in test case struct.

simonpasquier:
Thanks for the comments @cstyan. I've updated the PR: there's now `*testRunner.copyFile(src string)` when the destination name is the same as the source file (most of the cases) and `*testRunner.copyFileTo(src string, name string)` otherwise.

PR #5957, opened 2019-08-27
tsdb/wal: pull out wal metrics separately as tsdb.DB

johncming:
Extract wal's metrics for better encapsulation, like tsdb.DB

https://github.com/prometheus/prometheus/blob/5ecef3542d67f7af5fa519eb73261b07d167416a/tsdb/db.go#L121

### DIFF
--- a/tsdb/wal/wal.go
+++ b/tsdb/wal/wal.go
@@ -169,12 +169,59 @@ type WAL struct {
 	compress    bool
 	snappyBuf   []byte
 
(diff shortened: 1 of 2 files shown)


brian-brazil:
It's still the tsdb library.

beorn7:
@krasi-georgiev is there anything left to do?

krasi-georgiev:
All good, thanks!

PR #5963, opened 2019-08-27
Allow to control behaviour when there is a failed remote read

svagner:
Currently if any query errors that occur while we try to read from remote are returned as a list of warnings and the query continues.
Sometimes it can return partition result (data from all other not failed remote readers). Related MR - https://github.com/prometheus/prometheus/pull/4426
After this patch user can define if he wants to stop processing and return error while remote reader error occurs.
New option in `remote_read` config section is `fail_on_error`. By default option is false, so we are not changing current prometheus behaviour.

Also warning that provided is not very helpful:
`{"status":"success","data":{"resultType":"matrix","result":[]},"warnings":["Remote server returned HTTP status 400 Bad Request"]}`

After those change error will be:
`{"status":"success","data":{"resultType":"matrix","result":[]},"warnings":["Remote server http://127.0.0.1:10000/api/v1/read returned HTTP status 400 Bad Request: exceeded sample limit (5)"]}`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -642,6 +642,7 @@ type RemoteReadConfig struct {
 	URL           *config_util.URL `yaml:"url"`
 	RemoteTimeout model.Duration   `yaml:"remote_timeout,omitempty"`
 	ReadRecent    bool             `yaml:"read_recent,omitempty"`
+	FailOnError   bool             `yaml:"fail_on_error,omitempty"`
 	// We cannot do proper Go type embedding below as the parser will then parse
 	// values arbitrarily into the overflow maps of further-down types.
 	HTTPClientConfig config_util.HTTPClientConfig `yaml:",inline"`
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -1356,6 +1356,11 @@ basic_auth:
 tls_config:
   [ <tls_config> ]
 
+# By default any query errors that occur that aren't from the primary data source (the local tsdb)
+# are returned as a list of warnings and the query continues. If `fail_on_error` is set to `true`
+# for remote reader we will stop query and return error instead query continue.
+[ fail_on_error: <boolean> | default = false ]
+
 # Optional proxy URL.
 [ proxy_url: <string> ]
 ```
--- a/promql/engine_test.go
+++ b/promql/engine_test.go
@@ -176,6 +176,7 @@ func (q *errQuerier) Select(*storage.SelectParams, ...*labels.Matcher) (storage.
 func (*errQuerier) LabelValues(name string) ([]string, storage.Warnings, error) { return nil, nil, nil }
 func (*errQuerier) LabelNames() ([]string, storage.Warnings, error)             { return nil, nil, nil }
 func (*errQuerier) Close() error                                                { return nil }
+func (*errQuerier) FailOnError() bool                                           { return true }
 
 // errSeriesSet implements storage.SeriesSet which always returns error.
 type errSeriesSet struct {
@@ -247,6 +248,7 @@ func (*paramCheckerQuerier) LabelValues(name string) ([]string, storage.Warnings
 }
 func (*paramCheckerQuerier) LabelNames() ([]string, storage.Warnings, error) { return nil, nil, nil }
 func (*paramCheckerQuerier) Close() error                                    { return nil }
+func (*paramCheckerQuerier) FailOnError() bool                               { return true }
 
 func TestParamsSetCorrectly(t *testing.T) {
 	opts := EngineOpts{
--- a/storage/fanout.go
+++ b/storage/fanout.go
@@ -240,7 +240,7 @@ func (q *mergeQuerier) Select(params *SelectParams, matchers ...*labels.Matcher)
 		if err != nil {
 			q.failedQueriers[querier] = struct{}{}
 			// If the error source isn't the primary querier, return the error as a warning and continue.
-			if querier != q.primaryQuerier {
+			if !querier.FailOnError() {
 				warnings = append(warnings, err)
 				continue
 			} else {
@@ -250,6 +250,10 @@ func (q *mergeQuerier) Select(params *SelectParams, matchers ...*labels.Matcher)
 		seriesSets = append(seriesSets, set)
 	}
 	return NewMergeSeriesSet(seriesSets, q), warnings, nil
+}
+
+func (q *mergeQuerier) FailOnError() bool {
+	return true
 }
 
 // LabelValues returns all potential values for a label name.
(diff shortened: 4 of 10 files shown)


brian-brazil:
Remote storage should never be a hard dependency, you always have local storage to work off. If you want to fail queries that have warnings, you should do that in the client rather than making your monitoring more fragile.

The warning improvements would be good though.

svagner:
> Remote storage should never be a hard dependency, you always have local storage to work off. If you want to fail queries that have warnings, you should do that in the client rather than making your monitoring more fragile.
> 
> The warning improvements would be good though.

We have installation with a lot of separate instances, so we querying data from prometheus UI and grafana. Both clients expect real error (not warnings) for show it for client. It is first reason of this change - it is pretty difficult to change code of any clients. So I understand that not all users have the same use case, it is why I've added it as optional with saving preview behavior.  Second reason of that change: we receive `no data` while those kind of errors (currently warnings). In that case it brings incorrect data (not full set). In our case (and may be case of some other users) fullness of data is very important. So I offer to keep it optional with default behavior as it was defined in https://github.com/prometheus/prometheus/pull/4426

brian-brazil:
> so we querying data from this remote reader via prometheus UI and grafana

That's really not a good idea, that'll have both performance and reliability issues. I'd suggest looking at Thanos or Cortex.

> In our case (and may be case of some other users) fullness of data is very important.

If having all data from all remote read clients is more important than your monitoring working, then I'd suggest changing your client to handle warnings as errors.

svagner:
> > so we querying data from this remote reader via prometheus UI and grafana
> 
> That's really not a good idea, that'll have both performance and reliability issues. I'd suggest looking at Thanos or Cortex.
> 
For users queries (that happens not too often + read samples limit) performance degradation don't bring some problems. Reliability issues in that case only with remote instances, but for us prometheus service more important for monitoring, so queries from users are just for troubleshooting or look on some short history.

> > In our case (and may be case of some other users) fullness of data is very important.
> 
> If having all data from all remote read clients is more important than your monitoring working, then I'd suggest changing your client to handle warnings as errors.

Generally our monitoring processed on instances with local storage. I think set monitoring based on remote readers is bad idea. For monitoring I see only 2 stable way: use local storage or use federation. Our case for monitoring is local storage on each instance + federated for monitoring infra-wide (across all instances) data.

brian-brazil:
> Here is another problem - users see incorrect data (not full data set) while troubleshoot and receive alerts based on full data set.

It'd be recommended to look at the Prometheus with the data you want, rather than trying to proxy it all through a central Prometheus. Reliability in Prometheus terms means that alerts and graphs work, and if some older data from a few weeks back is temporarily unavailable that's not the end of the world.

> Our case for monitoring is local storage on each instance + federated for monitoring infra-wide (across all instances) data.

That's the recommended setup. Pulling graphs through one Prometheus via remote_read is not.

svagner:
> > Here is another problem - users see incorrect data (not full data set) while troubleshoot and receive alerts based on full data set.
> 
> It'd be recommended to look at the Prometheus with the data you want, rather than trying to proxy it all through a central Prometheus. Reliability in Prometheus terms means that alerts and graphs work, and if some older data from a few weeks back is temporarily unavailable that's not the end of the world.

Very often users shouldn't know about 100 instances of prometheuses and what data stored on what instances. I absolutely agree that if data for few weeks back unavailable it's not end of the world. But I think that end of the world if you ask for fresh data and receive incorrect data collection. As example I want to see sum of queries to all services. In reality I have 100b queries, but prometheus in that case will return 50b. I know that for queries like that best way is federated, but if this query happens one time in month I think federated with aggs isn't good idea (just waste disk space).
I'm really confusing about reliability here.  So if users query instances directly and UI of instance unavailable there is no reliability anyway. Thanos and Cortex are trying to solve reliability problem but anyway with some trade off (ex. performance degradation). So we should find trade off in any cases.

> 
> > Our case for monitoring is local storage on each instance + federated for monitoring infra-wide (across all instances) data.
> 
> That's the recommended setup. Pulling graphs through one Prometheus via remote_read is not.

brian-brazil:
> if this query happens one time in month I think federated with aggs isn't good idea (just waste disk space).

That sort of thing is generally quite cheap. By the time it's federated, the aggregation should have greatly reduced the cardinality.

> So if users query instances directly and UI of instance unavailable there is no reliability anyway.

Yes, but only for that instance (and presuming you don't have a redundant one). Every other instance will still work though.

svagner:
> > if this query happens one time in month I think federated with aggs isn't good idea (just waste disk space).
> 
> That sort of thing is generally quite cheap. By the time it's federated, the aggregation should have greatly reduced the cardinality.
For one query I agree. But for any variants of queries it would not be cheap + it makes usage of service more difficult (they should write aggregations for all queries before they will use it in future).
> 
> > So if users query instances directly and UI of instance unavailable there is no reliability anyway.
> 
> Yes, but only for that instance (and presuming you don't have a redundant one). Every other instance will still work though.

It still will require from users to have knowledge about all instances.

I just wander about remote reader using in that case. As we discussed above remote reader is bad idea for any variants (overhead, reliability for query data etc.), but we still continue to support it. It will return incorrect data but will be reliable (more then one remote_read server), it can not return data at all, but still will available (one remote_read server). I think I don't follow idea of remote reader in that case at all...

brian-brazil:
> I think I don't follow idea of remote reader in that case at all...

The primary intended use case is remote storage, so the local instance has complete data for say a few weeks and anything further back which that instance used to know can come from remote storage. If a big time chunk is missing that shouldn't matter much, bar for the very small number of alerts that are using old historical data.

svagner:
> > I think I don't follow idea of remote reader in that case at all...
> 
> The primary intended use case is remote storage, so the local instance has complete data for say a few weeks and anything further back which that instance used to know can come from remote storage. If a big time chunk is missing that shouldn't matter much, bar for the very small number of alerts that are using old historical data.

Ok, I got it. I still don't understand use case when we support remote reader from another prometheus. But it is another discussion I think. So can we leave improve error message text from this MR and can I add warnings show in UI then?

brian-brazil:
> I still don't understand use case when we support remote reader from another prometheus. 

We used it for the 1.x->2.x transition, and there's another rarer use case too.

> So can we leave improve error message text from this MR and can I add warnings show in UI then?

Sure, though the UI one is probably better done as a separate PR.

svagner:
> > I still don't understand use case when we support remote reader from another prometheus.
> 
> We used it for the 1.x->2.x transition, and there's another rarer use case too.
> 
> > So can we leave improve error message text from this MR and can I add warnings show in UI then?
> 
> Sure, though the UI one is probably better done as a separate PR.

Ok, I'll create it in separate MR

PR #5964, opened 2019-08-28
Show warnings in UI if query have returned some warnings

svagner:
- show warnings in UI if query have returned some warnings
- improve warning (error) text if query to remote was finished with error

As follow up of MR https://github.com/prometheus/prometheus/pull/5963

### DIFF
--- a/web/ui/static/js/graph/index.js
+++ b/web/ui/static/js/graph/index.js
@@ -513,6 +513,10 @@ Prometheus.Graph.prototype.submitQuery = function() {
           return;
         }
 
+        if ("warnings" in json && json.warnings.length > 0) {
+          self.showWarning(json.warnings.join('<br>'));
+        }
+
         queryHistory.handleHistory(self);
         success(json.data, textStatus);
       },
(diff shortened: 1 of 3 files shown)


svagner:
@brian-brazil Could you please take a look to this MR?

web/ui/static/js/graph/index.js
```diff
@@ -513,6 +513,10 @@ Prometheus.Graph.prototype.submitQuery = function() {
           return;
         }
 
+        if ("warnings" in json && json.warnings.length > 0) {
+          self.showWarning(json.warnings.join('<br>'));
```
brian-brazil: We should probably prefix this with an indication that it's a remote read issue.

svagner: May be it wpuld be better to add it within error from storage/remote like fmt.Sprintf("remote_read: %s", err). How do you think?

brian-brazil: Makes sense.

svagner: Prefix was added

brian-brazil:
Thanks!

PR #5966, opened 2019-08-28
building individual binaries with make build

geekodour:
allows: `make build PROMU_BINARIES="prometheus,promtool"`

@simonpasquier I am working on a pr to use circle ci artifacts for peombench test then we probably won't need this, but we need this for https://github.com/prometheus/prombench/pull/271 now, otherwise will need another `make promu` step but I think this is still an useful feature to have.

cc @SuperQ

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -201,7 +201,7 @@ endif
 .PHONY: common-build
 common-build: promu
 	@echo ">> building binaries"
-	GO111MODULE=$(GO111MODULE) $(PROMU) build --prefix $(PREFIX) $(BUILD_NAMES)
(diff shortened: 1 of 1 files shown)


Makefile.common
```diff
@@ -201,7 +201,7 @@ endif
 .PHONY: common-build
 common-build: promu
 	@echo ">> building binaries"
-	GO111MODULE=$(GO111MODULE) $(PROMU) build --prefix $(PREFIX)
+	GO111MODULE=$(GO111MODULE) $(PROMU) build --prefix $(PREFIX) $(BUILD_NAMES)
```
simonpasquier: I'd rather use `BUILD_BINARIES` or `PROMU_BINARIES`.

simonpasquier:
LGTM

Issue #5956, opened 2019-08-27
Add some methods to wal page


johncming:
## Proposal

**Why**

Better encapsulation.

**Action**

`page.buf` abstracts three methods

1. `func (*page) alloc() []byte`

Include the following codes

https://github.com/prometheus/prometheus/blob/7d43feb03ffca86276bf6dd47b618b6986c6583a/tsdb/wal/wal.go#L489

2. `func (*page) remaining() []byte`

Include the following codes

https://github.com/prometheus/prometheus/blob/7d43feb03ffca86276bf6dd47b618b6986c6583a/tsdb/wal/wal.go#L611

3. `func (*page) reset()`

Include the following codes

https://github.com/prometheus/prometheus/blob/7d43feb03ffca86276bf6dd47b618b6986c6583a/tsdb/wal/wal.go#L497-L501

beorn7:
Based on the outcome of the discussion in #5967, I'm closing this. Please speak up if you disagree.

PR #5967, opened 2019-08-28
Add methods to fetch page's buf data in tsdb WAL

erbesharat:
This PR adds methods like `bufalloc` or `bufremaining` to fetch page's buf data and a `reset` method to reset page's `alloc` and `buf` fields in tsdb WAL.

Related Issue: #5956

### DIFF
--- a/tsdb/wal/wal.go
+++ b/tsdb/wal/wal.go
@@ -64,6 +64,22 @@ func (p *page) full() bool {
 	return pageSize-p.alloc < recordHeaderSize
 }
 
+func (p *page) bufalloc() []byte {
+	return p.buf[p.flushed:p.alloc]
+}
+
+func (p *page) bufremaining() []byte {
+	return p.buf[p.alloc:]
+}
+
+func (p *page) reset() {
+	for i := range p.buf {
+		p.buf[i] = 0
+	}
+	p.alloc = 0
+	p.flushed = 0
+}
+
 // Segment represents a segment file.
 type Segment struct {
 	*os.File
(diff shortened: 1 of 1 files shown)


erbesharat:
The force push is because I didn't know about the DCO check.

codesome:
I personally don't see a need for this given all of them are used in a single place and all are simple enough to be inlined.

krasi-georgiev:
maybe the reset one makes it a bit nicer to read, but the rest yeah , don't see how it helps.

erbesharat:
Yeah, I agree that only the `reset` makes the code cleaner but thought since others are mentioned in the issue, they have to be implemented too.

codesome:
>  since others are mentioned in the issue, they have to be implemented too

Not necessary. Could you just keep the `reset` method?

krasi-georgiev:
the failing test is about a data race that is not related to this PR so merging.

Thanks

brancz:
> the failing test is about a data race that is not related to this PR so merging.

Even if that is the case, I think we agreed that we at least want to see a green CI by re-running if that's the case. Otherwise we don't truly know if it's just a flake.

krasi-georgiev:
I don't believe I have seen an invalid race condition so I have opened an issues.
https://github.com/prometheus/prometheus/issues/5977

I thought to leave it as is just to have the race details, but I copied the logs in the issues and now have triggered the CI run again.

krasi-georgiev:
all green in CI now.

brancz:
Opening an issue is even better. Thanks! :)

PR #5970, opened 2019-08-29
Change the global variable 'name' to a local variable so that it can not populate the global space.

Zhuzhenghao:
Change the global variable 'name' to a local variable so that it can not populate the global space.


simonpasquier:
For the CI to pass, you have to run `make assets` and commit the changes to `web/ui/assets_vfsdata.go`.

Zhuzhenghao:
> make assets

thx, I will do that immediately.

csmarchbanks:
Thanks!

Issue #5926, opened 2019-08-19
Allow Prometheus to use ReadRequest_STREAMED_XOR_CHUNKS when calling remote read API.


bwplotka:
https://github.com/prometheus/prometheus/pull/5703 extended remote read protocol and added `ReadRequest_STREAMED_XOR_CHUNKS`

Prometheus implements remote read API, but it can also be configured to **use/call** other remote read API. Enable streaming remote read if the server supports it.

AC:
* Prometheus remote read client is capable to use `ReadRequest_STREAMED_XOR_CHUNKS`

Similar client code for inspiration: https://github.com/thanos-io/thanos/pull/1268

YaoZengzeng:
I'd like to fix this issue 😃 @bwplotka 

Maybe we can solve this problem by following steps:
1. Extend the `Read` method of `Client` in `storage/remote/client.go` to support both SAMPLE and STREAM, the return type changed to `*prompb.QueryResult` to `*http.Response` would be better.

2. Refactor `FromQueryResult` to decode both type of response and expose unified `storage.SeriesSet`.

3. Encapsulate the chunks into some struct to make it implement `storage.SeriesSet`. Actually the implementation method is similar to `chunkSeries` stuff in `tsdb/query.go`. In addition, this step may easy to be implemented if #5882 being merged :) 
( At first I‘d encapsulate the chunks to make the code run)

My initial idea is probably like this, any comments? @bwplotka

bwplotka:
Sorry for lag and thanks @YaoZengzeng 

Yes that makes sense to me :+1: 

I think step 2 would be the most work, especially packing SAMPLED method to `SeriesSet`, but I guess we can't really avoid this.

For Streaming be careful as we might need to implement something like [this we did in Thanos](https://github.com/thanos-io/thanos/blob/master/pkg/query/iter.go#L39) in SeriesSet.

YaoZengzeng:
@bwplotka Thanks for your comments and I'll fix this issue step by step 😃

YaoZengzeng:
@bwplotka Please review #5974 again, otherwise the fixing of this issue can't move on 😂

bwplotka:
Sorry for massive lag, proposed solution (: Also reach me on IRC bwplotka or [CNCF slack channel](https://slack.cncf.io/) bwplotka if you need to ping me - bit spammed on email.

bwplotka:
[...]

PR #5974, opened 2019-08-30
Test remote read client

YaoZengzeng:

### DIFF
--- a/storage/remote/client_test.go
+++ b/storage/remote/client_test.go
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package remote
+package remote_test
 
 import (
 	"context"
@@ -25,11 +25,19 @@ import (
 	"github.com/pkg/errors"
 	config_util "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
+	"github.com/prometheus/common/route"
 
+	"github.com/prometheus/prometheus/config"
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/prompb"
+	"github.com/prometheus/prometheus/promql"
+	"github.com/prometheus/prometheus/storage"
+	. "github.com/prometheus/prometheus/storage/remote"
 	"github.com/prometheus/prometheus/util/testutil"
+	apiv1 "github.com/prometheus/prometheus/web/api/v1"
 )
 
-var longErrMessage = strings.Repeat("error message", maxErrMsgLen)
+var longErrMessage = strings.Repeat("error message", MaxErrMsgLen)
 
 func TestStoreHTTPErrorHandling(t *testing.T) {
 	tests := []struct {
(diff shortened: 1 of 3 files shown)


YaoZengzeng:
It is the test case for SAMPLED remote read client.

Also it laid a solid foundation for testing STREAM remote read client.

@bwplotka PTAL

ref #5926

storage/remote/client_test.go
```diff
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package remote
+package remote_test
```
bwplotka: So depends on context, but usually this is antipattern - I think on Prometheus we stick for tests to have exactly the same package. Can we revert this and same with exporting `recoverableError`?

storage/remote/client_test.go
```diff
@@ -82,3 +90,87 @@ func TestStoreHTTPErrorHandling(t *testing.T) {
...
+func TestReadHandling(t *testing.T) {
+	suite, err := promql.NewTest(t, `
+		load 1m
+			test_metric1{foo="bar",baz="qux"} 1
+	`)
+	testutil.Ok(t, err)
+
+	defer suite.Close()
+
+	err = suite.Run()
```
bwplotka: you can inline this in `testutil.Ok(t, suite.Run())`

storage/remote/client_test.go
```diff
@@ -82,3 +90,87 @@ func TestStoreHTTPErrorHandling(t *testing.T) {
 		server.Close()
 	}
 }
+
+func TestReadHandling(t *testing.T) {
```
bwplotka: Maybe 
```suggestion
func TestReadClient(t *testing.T) {
```

?

storage/remote/client_test.go
brian-brazil: Yes, and that's standard Go practice AFAIK

YaoZengzeng: Actually it's easiest way to avoid `import cycle` (client test -> api -> client).

Also, why I can't receive your comments from email @bwplotka , really weird 😂

bwplotka: No idea. So you were not notified about my review?

bwplotka: Also yea, we need to fix the cycle in different way then.

YaoZengzeng: Could from Github page, but not email 😄 

If we want to reuse the package `web/api/v1`, then the cycle can not be avoided.

Maybe we could simulate an prometheus instance like thanos do ?

But it seems a big change, so we could do it in another PR (or several PR) ?

@bwplotka  WDYT?

YaoZengzeng: ping @bwplotka

bwplotka: Sorry for delay. I looked through and we have 2 options:
* Mock server in some other way
* Move client code to separate package e.g `storage/remote/client`

I really like the second option because as far as I can see there is literally no import for `remote` stuff from `client.go` or `client_test.go` so it is actually might be more natural way.
Plus we can use exact api code for server side for both remote read and write :+1:  What do you think?

bwplotka: `storage/remote ` is not using anything from `storage/remote/client/client.go` and `storage/remote/client/client_test.go` AFAIK  only `api/v1` is using this. 

> > Plus we can use exact api code for server side for both remote read and write +1 What do you think?
> Could you describe it in more detail? :)

Sure, I mean that the extra benefit for moving client to separate package is that we can in `api/v1_test.go` use exact server remote read code instead of just mocking something that *should* behave like server code, right?

YaoZengzeng: Yes, use exact api code for server side for both remote read and write is great.

But `storage/remote` surely use client related code for remote write and read 😂 

https://github.com/prometheus/prometheus/blob/master/storage/remote/storage.go#L72

https://github.com/prometheus/prometheus/blob/master/storage/remote/write.go#L135

bwplotka: Yea, sorry, I missed that - it's quite tightly together. Still we have few options. I think making testing package external is not really preferable as it's a just hack for the potentially improper composition of packages. I would also try to avoid mocking server. 

In this case, I would say, let's just test client e2e in `api/v1_test.go` - let's not spend to much time on this!

The long term solution would be to allow `remote/storage.go` and `write.go` to work on `RemoteClient` interface rather and have `main` to put `remote/client` implementation (:

YaoZengzeng: Actually after more digging into the code, I find the remote part of `TestEndpoints` in `web/api/v1/api_test.go` almost do the same thing as this PR 😱 Maybe after adding ExternalLabels part, it would be enough for us.

Sad to spend so much time to do the duplicate thing 😂

bwplotka:
Can we close this one then?

YaoZengzeng:
Yes, this PR will be closed.

Issue #5878, opened 2017-05-22
tsdb: Use context.Context for transaction lifecycle


fabxc:
Currently `Appender` and `Querier` both have `Commit/Rollback` and `Close`respectively.
Especially Rollback/Close are easy to forget to do properly, causing deadlocks.

One solution would be to replace the explicit aborts with contexts cancelation. 

The interfaces would change as follows:

```go
type Queryable interface {
    Querier(context.Context, mint, maxt int64)
}

type Appednable interface {
    Appender(context.Context, mint, maxt int64)
}
```

On context cancelation `Rollback` and `Close` behaviour would be called. A `Commit` would error if the context is already canceled.

This would make sense to pass to `Delete` as well @Gouthamve (after the main work is in).
It's the only operation that is actually executing a long-running work in a blocking manner and particularly because it may be stuck waiting for the compaction lock. That may take longer than a related HTTP request remains open.

Problem: 
Querying returns iterators over storage memory, which must only be accessed while the transaction is still ongoing.
We cannot really guarantee that client code receives the context's cancelation signal before the storage. For example, the passed context may have a parent context with a timeout. So the storage might return locks before we stop accessing memory. 
Strictly speaking, all data in memory and persisted blocks should be append-only. But that would need some careful checking again.

The problem here really is the high-frequency of our operations. We cannot check `<-ctx.Done` after every sample we added or iterated over as `select` is rather expensive. 

So this is more of an issue for feature consideration and shouldn't be added lightly.

brian-brazil:
We look at this as part of the bug scrub, and currently queries are fast enough in the tsdb that a lack of contexts is not an issue so closing.

PR #5978, opened 2019-09-01
Support for context cancellation in Querier (tsdb)

Harkishen-Singh:

Refers to #5878 

This PR adds context cancellation for `Querier` **(tsdb)** in the transaction lifecycle.

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -269,6 +269,12 @@ type DBReadOnly struct {
 	closed  chan struct{}
 }
 
+// Queryable implements a layer for context cancelations over the Querier.
+// Cancelation of any query would lead to rollback and close
+type Queryable interface {
+	Querier(ctx context.Context, mint, maxt int64) (Querier, error)
+}
+
 // OpenDBReadOnly opens DB in the given directory for read only operations.
 func OpenDBReadOnly(dir string, l log.Logger) (*DBReadOnly, error) {
 	if _, err := os.Stat(dir); err != nil {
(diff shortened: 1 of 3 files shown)


tsdb/db.go
```diff
@@ -269,6 +269,12 @@ type DBReadOnly struct {
 	closed  chan struct{}
 }
 
+// Queryable implements a layer for context cancelations over the Querier.
+// Cancelation of any query would lead to rollback and close
```
brian-brazil: Queries are read-only. What are you trying to do?

Harkishen-Singh: @brian-brazil  please correct me if I have misunderstood, the issue #5878 asks to implement cancellation in order to avoid issues due to any failure to any higher-level task. Hence, I wrapped the Querier with context cancellation layer which could be stopped at the higher-level if any error arises and closes the db.

Harkishen-Singh: have I understood the issue correctly?

brian-brazil: What you propose doesn't cover the bits of the api that matter, opening the DB is not the interesting bit. It's the actual querying/appending that's of interest.

Last we talked about this, we managed to get things fast enough that contexts weren't needed yet and we'll add them once that's not the case.

@jacksontj @krasi-georgiev  FYI

krasi-georgiev: I think there were still few places that would benefit a context cancellation, but this would need testing and if context can be avoided by further optimizations than yes we should avoid it. 

Btw I don't see how does this PR can stop a long running query.

Harkishen-Singh: @krasi-georgiev  I was about to implement content cancellation in the entire package including the appending part intensively, but reading the suggestions, I felt that the importance of content implementation is not of much use as of now. I can resume the implementation if we are in need of this. What do you think? @brian-brazil  @krasi-georgiev @jacksontj

krasi-georgiev: yes It would be best to implement this with a case where an optimization doesn't solve the problem.

Harkishen-Singh: okay @krasi-georgiev , would be moving ahead as suggested and modify the PR accordingly. Thanks

krasi-georgiev: would you mind closing this one as by the time this is needed the PR would be very outdated so not worth keeping it open.

Harkishen-Singh: sure

PR #5983, opened 2019-09-04
Trim benchmarkIterator() memory usage

bboreham:
Previously it was allocating millions of chunks, all containing the same 250 samples.  Above some ratio of CPU performance to available memory, the benchmark cannot run.

Make 250 a const and just allocate one chunk which we iterate repeatedly till we reach the benchmark count.

~Also avoid calling `testutil.Ok()` every time round the benchmark loop - the cost of this call is about 20% of the entire loop, and it seems to be causing some memory allocation.~

### DIFF
--- a/tsdb/chunkenc/chunk_test.go
+++ b/tsdb/chunkenc/chunk_test.go
@@ -110,12 +110,13 @@ func testChunk(t *testing.T, c Chunk) {
 }
 
 func benchmarkIterator(b *testing.B, newChunk func() Chunk) {
+	const samplesPerChunk = 250
 	var (
 		t   = int64(1234123324)
 		v   = 1243535.123
 		exp []pair
 	)
-	for i := 0; i < b.N; i++ {
+	for i := 0; i < samplesPerChunk; i++ {
 		// t += int64(rand.Intn(10000) + 1)
 		t += int64(1000)
 		// v = rand.Float64()
@@ -123,11 +124,9 @@ func benchmarkIterator(b *testing.B, newChunk func() Chunk) {
 		exp = append(exp, pair{t: t, v: v})
 	}
 
-	var chunks []Chunk
-	for i := 0; i < b.N; {
-		c := newChunk()
-
-		a, err := c.Appender()
+	chunk := newChunk()
+	{
+		a, err := chunk.Appender()
 		if err != nil {
 			b.Fatalf("get appender: %s", err)
 		}
@@ -137,32 +136,27 @@ func benchmarkIterator(b *testing.B, newChunk func() Chunk) {
 				break
 			}
 			a.Append(p.t, p.v)
-			i++
 			j++
 		}
-		chunks = append(chunks, c)
 	}
 
 	b.ReportAllocs()
 	b.ResetTimer()
 
-	b.Log("num", b.N, "created chunks", len(chunks))
-
-	res := make([]float64, 0, 1024)
-
+	var res float64
 	var it Iterator
-	for i := 0; i < len(chunks); i++ {
-		c := chunks[i]
-		it := c.Iterator(it)
+	for i := 0; i < b.N; {
+		it := chunk.Iterator(it)
 
 		for it.Next() {
 			_, v := it.At()
-			res = append(res, v)
+			res = v
+			i++
 		}
 		if it.Err() != io.EOF {
 			require.NoError(b, it.Err())
 		}
-		res = res[:0]
+		_ = res
 	}
 }
 


tsdb/chunkenc/chunk_test.go
```diff
@@ -120,32 +119,27 @@ func benchmarkIterator(b *testing.B, newChunk func() Chunk) {
...
-		c := chunks[i]
-		it := c.Iterator(it)
+	for i := 0; i < b.N; {
+		it := chunk.Iterator(it)
 
 		for it.Next() {
 			_, v := it.At()
-			res = append(res, v)
+			res = v
+			i++
```
krasi-georgiev: why do we even need this?

I think the only useful bench here is that passing an existing `it` to  `it := chunk.Iterator(it)` saves memory allocation.

bboreham: Which line(s) are you questioning?
 * `i++` is needed so we run the benchmark an appropriate number of times.
 * `res = v` is not needed, but I wanted to give some nod to the previous code, perhaps it was hoping to avoid the work being optimised out.

krasi-georgiev: wouldn' something like this do the same job?

```
	for i := 0; i < b.N; i++ {
```

krasi-georgiev: hm I tested it and it returns a completely different results.

krasi-georgiev: with your version the results are:

```
BenchmarkXORIterator-8           1000000              1075 ns/op               0 B/op          0 allocs/op
```

and with my syggestion:

```
BenchmarkXORIterator-8              5000            274289 ns/op              96 B/op          1 allocs/op
```

bboreham: Well, yes, mine is trying to do the same amount of work as before, just without gigabytes of memory in the set-up.
Yours is perfectly reasonable, but not comparable between versions.

krasi-georgiev: Strange, looking at the code:`i` only determines how many times to run the loop and it would stop executing only when few iterations produce similar/same results right? What am I missing?

```
it := chunk.Iterator(it) ---> it = chunk.Iterator(it)
```
to avoid allocating a new iterator.

bboreham: In the original code and my version, `b.N` determines the number of samples to read.
In your version, `b.N` chunks are read, and each chunk has 250 samples so it does 250x as much work per iteration.

beorn7:
@krasi-georgiev are you still ready to review this?

tsdb/chunkenc/chunk_test.go
```diff
@@ -110,24 +110,23 @@ func testChunk(t *testing.T, c Chunk) {
 }
 
 func benchmarkIterator(b *testing.B, newChunk func() Chunk) {
+	const samplesPerChunk = 250
```
bwplotka, line 113: Why not taking 120 as it's the usual value?

tsdb/chunkenc/chunk_test.go
```diff
@@ -110,24 +110,23 @@ func testChunk(t *testing.T, c Chunk) {
...
 	var (
 		t   = int64(1234123324)
 		v   = 1243535.123
 		exp []pair
 	)
-	for i := 0; i < b.N; i++ {
+	for i := 0; i < samplesPerChunk; i++ {
 		// t += int64(rand.Intn(10000) + 1)
```
bwplotka, line 120: Do we need commented code?

tsdb/chunkenc/chunk_test.go
bwplotka: Makes sense to me.

@krasi-georgiev to answer your question, b.N will be limit for samples not chunks, just different unit - should be good enough.

bwplotka:
If you are still intersted @bboreham to look on 1y old improvement. Our review latency is quite high here ;p

krasi-georgiev:
fighting bugs on some other fronts so feel free to merge without my review

tsdb/chunkenc/chunk_test.go
bboreham: The number 250 comes from this line: https://github.com/prometheus/prometheus/pull/5983/files#diff-37dd1fb1ccb424c10039dd47e623d671L136

Probably I should remove that line since it doesn't do anything now.

Possibly I should replace the other `250` in the file.

tsdb/chunkenc/chunk_test.go
bboreham: Not relevant to this PR.

bboreham:
I have rebased and removed the second commit from this PR - solved upstream at https://github.com/stretchr/testify/pull/933

roidelapluie:
@bwplotka can we get this in now? :)))

bboreham:
Anything else needed here?  Looks like CI got stuck but I don't have permission to re-run it.

tsdb/chunkenc/chunk_test.go
```diff
@@ -137,32 +136,27 @@ func benchmarkIterator(b *testing.B, newChunk func() Chunk) {
...
-	for i := 0; i < len(chunks); i++ {
-		c := chunks[i]
-		it := c.Iterator(it)
+	for i := 0; i < b.N; {
+		it := chunk.Iterator(it)
 
 		for it.Next() {
 			_, v := it.At()
-			res = append(res, v)
+			res = v
```
codesome, line 153: What does this `res` achieve compared to just `_, _ = it.At()`

tsdb/chunkenc/chunk_test.go
```diff
@@ -110,24 +110,23 @@ func testChunk(t *testing.T, c Chunk) {
...
 
 func benchmarkIterator(b *testing.B, newChunk func() Chunk) {
+	const samplesPerChunk = 250
 	var (
 		t   = int64(1234123324)
 		v   = 1243535.123
 		exp []pair
 	)
-	for i := 0; i < b.N; i++ {
+	for i := 0; i < samplesPerChunk; i++ {
```
codesome, line 119: A non-blocking comment for this PR: we could actually merge this loop with chunk creation loop below. `exp` is not required at all.

codesome (approved):

codesome:
> Looks like CI got stuck but I don't have permission to re-run it.

I tried to start the workflow from start, but it did not do anything. Can you try pushing an empty commit? Or maybe a rebase with main branch.

tsdb/chunkenc/chunk_test.go
bboreham: Not much.  I wanted to make it similar to the previous code, which collected results in `res`.

bboreham:
Rebased, and CI passed now.

PR #5984, opened 2019-09-04
Update to disallow new Go 1.13 float formats.

brian-brazil:
Don't inadvertantly expand the acceptable ways to specify
a float.

@beorn7 
@SuperQ @RichiH We haven't explicitly discussed this for OM, but I don't think there'll be any disagreement here. We want something easily parsed by standard float parsers.

### DIFF
--- a/pkg/textparse/openmetricsparse.go
+++ b/pkg/textparse/openmetricsparse.go
@@ -20,7 +20,6 @@ import (
 	"io"
 	"math"
 	"sort"
-	"strconv"
 	"strings"
 	"unicode/utf8"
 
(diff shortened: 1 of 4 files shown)


RichiH:
We shouldn't be changing the OM spec just to follow one specific
programming language, agreed.

brian-brazil:
Yeah, it's unfortunate. I realised last night we probably also need to check for upper case, so I've updated my PR and will merge this evening unless there's an objection.

beorn7:
If we want to go wild, we could copy over the parsing code from the standard library and rip out the new parts. We could then also make it act on `[]byte` and get rid of the scary `yoloString`.

(This is less crazy as it sounds. It was suggested by Russ Cox himself [here](https://github.com/golang/go/issues/2632#issuecomment-252725945).)

brian-brazil:
If someone cares about it performance wise they can do that, though we may then end up losing future performance improvements from upstream.

PR #5990, opened 2019-09-05
Reposition "remove graph" close to query area

### LINKED ISSUE #5496: "Remove graph" link should be close to query area
## Proposal
Move "Remove graph" link to the top of query section on Prometheus `/graph` page.

Sometimes badly composed query could return a lot of result rows. To remove that result user needs to scroll to the bottom of the page and press on "Remove graph" link. It could be a problem is query returned couple of thousands of result rows.

*“Nice to have” is not a good use case. :)*
It would be nice to place that link near to query area, so user could clear results without scrolling the page.

Harkishen-Singh:
Fixes #5496

Screenshot from 2019-09-05 18-18-54


roidelapluie:
It is the opportunity to rephrase that too:

Remove Graph and Expression ??  What would be better?

Harkishen-Singh:
@simonpasquier please review

Harkishen-Singh:
> It is the opportunity to rephrase that too:
> 
> Remove Graph and Expression ?? What would be better?

@roidelapluie sure. or probably just `Remove` since the button would be almost in the center vertically. So `Remove` would correspond to the removal of that particular component.  Let's see what others suggest.

Harkishen-Singh:
@simonpasquier  pls have a look. Thank you.

beorn7:
Apologies for letting this become stale.

@Harkishen-Singh could you resolve the merge conflict? (Probably only need to delete the vfsdata file.) This should be ready to get merged then.

Harkishen-Singh:
Sure @beorn7. I will resolve within few hours.

Harkishen-Singh:
Done @beorn7.

beorn7:
Thanks. Based on our discussion in the bug scrub, I'll merge this.

I think this can be refined further, as @roidelapluie suggested, e.g. calling it "Remove expression", but then we should also change the butten "Add Graph" to "Add Expression", and perhaps make the "Remove" link a proper button, too (possibly reddish in color…). 

But before we get lost in details, let's remember that there is a new UI already available, so that we should limit the effort that goes into refining the old UI.

PR #5994, opened 2019-09-06
tsdb: add block meta version constant

zhulongcheng:
## Changes

- [x] Add  `metaVersion1` constant
- [x] Replace number `1` with `metaVersion1`


krasi-georgiev:
Thanks!

Issue #5435, opened 2019-04-05
*: upgrade to gopkg.in/yaml.v3


mdlayher:
https://godoc.org/gopkg.in/yaml.v3 is now available: https://blog.ubuntu.com/2019/04/05/api-v3-of-the-yaml-package-for-go-is-available.

Here's an issue for tracking the upgrade, and any changes required to perform it.

brian-brazil:
We could also likely simplify things a bit. I'd suggest starting with a
simpler repo, like say blackbox or SNMP.

brian-brazil:
It appears that this may not have strict unmarshalling, and also Cortex ran into issues that we'll run into ourselves - so both need to be resolved before we go further.

Issue #21, opened 2013-01-04
Rule formatting tool.


juliusv:
Like "gofmt" for Go, we ought to have a "promfmt" for Prometheus since we have a syntax tree. The idea being that the system produces uniform style that minimizes deviation and learning curve.

**Update after we have totally moved to YAML rule files**: In addition to formatting the PromQL expressions, we also want to format the YAML files to have a fixed structure, while preserving comments for both PromQL expressions and the YAML file.

grobie:
:+1: * :100:

fabxc:
Yep, that'd be great. Preserving comments is the biggest (and basically only) problem.

brian-brazil:
Aren't we in the middle of adding this to promtool?

mattbostock:
Related: https://github.com/prometheus/prometheus/pull/1779

beorn7:
Some observations:
- With the 2.x rule format (YAML embedding PromQL), we want this to act on YAML files, too. 
- It should preserve both comments in YAML and in PromQL.
- Indentation is crucial (not just line breaks).
- It has to be seen if certain line breaks in the input should be preserved in the output (cf. how `gofmt` does it).
- Wherever a rule is rendered, the same formatting should apply, in particular on the _Alerts_ and _Rules_ tabs of the Prometheus web UI. This implies that the AST needs to keep comments (and possibly certain line breaks, see previous point).

brian-brazil:
> It should preserve both comments in YAML

Our current YAML library doesn't do this, but is hoping to "soon".

geekodour:
I'd like to include this in my gsoc proposal. @brian-brazil, can you link to the YAML library you're referring to?

brian-brazil:
https://github.com/go-yaml/yaml, but I'd scope this to just the PromQL as this stands.

juliusv:
@brian-brazil But since the PromQL from rules is always embedded in YAML files, it at least has to preserve YAML comments, right? Otherwise the tool wouldn't be very useful as nobody wants to lose their comments the benefit of formatting.

brian-brazil:
That'd be nice, but we can get benefits without that. I'd rather also not depend on something with an unclear timeline for a gsoc project.

beorn7:
Let's start with the PromQL part. The YAML part should be almost trivial once the library supports comment nodes.

juliusv:
Ok!

brian-brazil:
[...]

PR #5996, opened 2019-09-07
Upgrade yaml version to v3

geekodour:
See #5435 
See #21 

This PR just upgrades the yaml.v2 package to yaml.v3, preserving of comments is not added yet, I think it'll be good to do that in a separate pr. 

As discussed in riot /w @simonpasquier about the following snippet,  a "null" value for regex is equivalent to the zero-value string (eg. the empty string) which is true for yaml.v2
```
  relabel_configs:
  - regex:
    replacement:   static
    target_label:  abc
``` 
but with `yaml.v3` it is interpreted as null value,  so had to set `""` to `regex` explicitly in `/config/testdata/conf.good.yml`


simonpasquier:
There's no official release of `gopkg.in/yaml.v3` (https://github.com/go-yaml/yaml/issues/487) so I would refrain on using it for now. The behavior change regarding null vs. empty strings deserves more investigation too IMO.

brian-brazil:
We've had fun in the past with nulls and yaml, plus that technically a breaking change.

slrtbtfs:
Another issue with yaml.v3 is, that is does behave in unpredictable ways when there are multiple YAML docs in one file.

brian-brazil:
We only ever have one YAML doc, so that's not a concern.

bwplotka:
It's been a while since we saw movement here, so I would close this - we can always reopen. Thanks @geekodour!

We use v3 already a bit, but we would love to use yaml.v3 just for consistency at some point--low priority though - we have https://github.com/prometheus/prometheus/issues/21 to track the motivation for it.

PR #5999, opened 2019-09-10
Move WAL watcher code to tsdb/wal package.

cstyan:
The TSDB port happened while I was away on vacation so I'm reopening this as a prometheus/prometheus PR. 

Moving the WAL watcher code into the WAL package requires moving some code from the root tsdb package. Specifically the Watcher uses RefSeries and RefSample. Moving those requires a bunch of other code to be moved as well.

Pinging the same people that reviewed last time @csmarchbanks @bwplotka @codesome 

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -998,7 +998,7 @@ func (h *Head) Delete(mint, maxt int64, ms ...labels.Matcher) error {
 		// Delete only until the current values and not beyond.
 		t0, t1 = clampInterval(mint, maxt, t0, t1)
 		if h.wal != nil {
-			stones = append(stones, tombstones.Stone{p.At(), tombstones.Intervals{{t0, t1}}})
+			stones = append(stones, tombstones.Stone{Ref: p.At(), Intervals: tombstones.Intervals{{Mint: t0, Maxt: t1}}})
 		}
 		if err := h.chunkRewrite(p.At(), tombstones.Intervals{{Mint: t0, Maxt: t1}}); err != nil {
 			return errors.Wrap(err, "delete samples")
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -33,6 +33,8 @@ import (
 	"github.com/prometheus/prometheus/tsdb/encoding"
 	"github.com/prometheus/prometheus/tsdb/index"
 	"github.com/prometheus/prometheus/tsdb/labels"
+	"github.com/prometheus/prometheus/tsdb/record"
+	"github.com/prometheus/prometheus/tsdb/tombstones"
 	"github.com/prometheus/prometheus/tsdb/wal"
 )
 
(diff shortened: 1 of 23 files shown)


codesome:
Pointing to the previous PR: https://github.com/prometheus/tsdb/pull/606

It was a :+1: from me for that PR. Will take a look at this soon!

tsdb/tombstones/tombstones.go
```diff
@@ -30,7 +32,7 @@ import (
 	"github.com/prometheus/prometheus/tsdb/fileutil"
 )
 
-const tombstoneFilename = "tombstones"
+const Filename = "tombstones"
```
codesome: This can be confusing to use within the package. Rather keep it `TombstoneFilename`?

tsdb/head.go
```diff
@@ -508,8 +511,10 @@ func (h *Head) Init(minValidTime int64) error {
 
 	level.Info(h.logger).Log("msg", "replaying WAL, this may take awhile")
 	// Backfill the checkpoint first if it exists.
-	dir, startFrom, err := LastCheckpoint(h.wal.Dir())
-	if err != nil && err != ErrNotFound {
+	dir, startFrom, err := wal.LastCheckpoint(h.wal.Dir())
+	// We need to compare err to record.ErrNotFound as that's what
+	// wal.LastCheckpoint would return, not tsdb.ErrNotFound.
+	if err != nil && err != record.ErrNotFound {
```
bwplotka: I think that is quite clear so I would skip this comment.

tsdb/head.go
```diff
@@ -908,18 +928,21 @@ func (a *headAppender) log() error {
 func (a *headAppender) Commit() error {
 	defer a.head.metrics.activeAppenders.Dec()
 	defer a.head.putAppendBuffer(a.samples)
+	defer a.head.putSeriesBuffer(a.sampleSeries)
```
bwplotka: Hm, what's the goal of this change? Sorry I might be  missing the context, but I think this change is totally orthogonal to extracting the code to separate packages, right? Thus, it might be the best to keep those in separate PRs.

csmarchbanks: I was pretty against exposing `memSeries` in the last PR, and this was a way to be able to move `RefSample` to the `record` package without also having to move and expose `memSeries`. You can see some context starting with [this comment](https://github.com/prometheus/tsdb/pull/606#issuecomment-499331881), and a handful of comments afterward.

bwplotka: Cool, that's a valid use case, I will looks closer to those non-move-code changes then :+1:

tsdb/head.go
```diff
@@ -908,18 +928,21 @@ func (a *headAppender) log() error {
...
 
 	if err := a.log(); err != nil {
 		return errors.Wrap(err, "write to WAL")
 	}
 
 	total := len(a.samples)
+	var series *memSeries
```
bwplotka: minor nit: is this new line needed?

tsdb/head.go
bwplotka: LGTM

cstyan:
@bwplotka @csmarchbanks @codesome Let me know if there's any other comments, I fixed a few nits and rebased them into the first commit.

PR #6004, opened 2019-09-10
Bump golangci-lint to v1.18.0

simonpasquier:
It is required to support Go 1.13 (see https://github.com/golangci/golangci-lint/issues/652).


simonpasquier:
Merging despite the fuzzit job failing as it will be fixed by https://github.com/prometheus/prometheus/pull/6005.

PR #6006, opened 2019-09-10
tsdb: allow readonly DB to flush WAL

squat:
This PR gives the readonly DB the ability to create blocks from the WAL.
In order to implement this, we modify DBReadOnly.Blocks() to return an
empty slice and no error if no blocks are found.
xref: https://github.com/prometheus/tsdb/issues/346#issuecomment-520786524

Signed-off-by: Lucas Servén Marín <lserven@gmail.com>

cc @krasi-georgiev

### DIFF
--- a/tsdb/db_test.go
+++ b/tsdb/db_test.go
@@ -2359,3 +2359,76 @@ func TestDBReadOnlyClosing(t *testing.T) {
 	_, err = db.Querier(0, 1)
 	testutil.Equals(t, err, ErrClosed)
 }
+
+func TestDBReadOnly_FlushWAL(t *testing.T) {
+	var (
+		dbDir  string
+		logger = log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
+		err    error
+		maxt   int
+	)
+
+	// Boostrap the db.
+	{
+		dbDir, err = ioutil.TempDir("", "test")
+		testutil.Ok(t, err)
+
+		defer func() {
+			testutil.Ok(t, os.RemoveAll(dbDir))
+		}()
+
+		// Append data to the WAL.
+		db, err := Open(dbDir, logger, nil, nil)
+		testutil.Ok(t, err)
+		db.DisableCompactions()
+		app := db.Appender()
+		maxt = 1000
+		for i := 0; i < maxt; i++ {
+			_, err := app.Add(labels.FromStrings(defaultLabelName, "flush"), int64(i), 1.0)
+			testutil.Ok(t, err)
+		}
+		testutil.Ok(t, app.Commit())
+		defer func() { testutil.Ok(t, db.Close()) }()
+	}
+
+	// Flush WAL.
+	db, err := OpenDBReadOnly(dbDir, logger)
+	testutil.Ok(t, err)
+
+	flush, err := ioutil.TempDir("", "flush")
+	testutil.Ok(t, err)
+
+	defer func() {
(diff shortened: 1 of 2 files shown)


squat:
Looks like the test failure is an unrelated SD issue:
```
--- FAIL: TestFileSD (16.61s)
    file_test.go:113: Unexpected target groups [], we expected a failure here.
FAIL
FAIL	github.com/prometheus/prometheus/discovery/file	16.629s
```

tsdb/db_test.go
```diff
@@ -2359,3 +2359,85 @@ func TestDBReadOnlyClosing(t *testing.T) {
...
+		db.DisableCompactions()
+		app := db.Appender()
+		mint := int64(dbBlocks[len(dbBlocks)-1].MaxTime)
+		maxt = mint + 1000
+		for i := 0; i < 1000; i++ {
+			_, err := app.Add(labels.FromStrings(defaultLabelName, "0"), mint+int64(i), 1.0)
+			testutil.Ok(t, err)
+		}
+		testutil.Ok(t, app.Commit())
+		testutil.Ok(t, app.Rollback())
```
krasi-georgiev: Why the rollback call here?

krasi-georgiev:
> Looks like the test failure is an unrelated SD issue:

yes I will re-trigger the CI

tsdb/db_test.go
squat: ah, not sure; I saw it on the other tests and thought it was boilerplate. I'll remove it and rebase the PR :+1:

squat:
Dang, this time there was a problem downloading the deps:
```
go: gopkg.in/yaml.v2@v2.2.2: git -c protocol.version=0 fetch --unshallow -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /go/pkg/mod/cache/vcs/4f939cf8921e8662ef2d8cf78b0db3bcacccda0d79c9e1b4cef06c640b4a8cef: exit status 128:
	error: RPC failed; HTTP 502 curl 22 The requested URL returned error: 502 
	fatal: the remote end hung up unexpectedly
```
Can we re-kick the test?

krasi-georgiev:
finally all green.

@bwplotka mind having a quick final look?

Added this to make the Thanos receiver snapshots safer in https://github.com/thanos-io/thanos/pull/1480

krasi-georgiev:
yes I think that the wal shouldn't be deleted and there is no issue to remain in the folder even with the new block as all wal samples with times lower than the last block maxt are ignored anyway.

bwplotka:
> with times lower than the last block maxt are ignored anyway.

Ignore by running Prometheus even? Sure then it's fine :+1:

krasi-georgiev:
@bwplotka 	with the current implementation running a flush over a running server is still racy. Imagine running a flush at the same time a compaction happens. you will end up with 2 almost identical blocks.
Don't think this should be intended on running it along another prometheus using the same folder and it is commented in the method:
https://github.com/prometheus/prometheus/pull/6006/files#diff-19345ac9885e2f1a8062b757506c91d4R291-R292

bwplotka:
I am not suggesting that. I'm rather exploring case of:

1. Run TSDB
2. Stop TSDB
3. Run ReadOnlyTSDB.WalToBlock
4. Run Prometheus again on same dir (will it work)?

Anyway, let's merge this PR in the current form (:  LGTM

krasi-georgiev:
@bwplotka yes that is the intended workflow.

PR #6012, opened 2019-09-11
Remove extra symbols

yuxiaobo96:

### DIFF
--- a/promql/test.go
+++ b/promql/test.go
@@ -461,7 +461,7 @@ func (t *Test) exec(tc testCommand) error {
 		}
 
 		// Check query returns same result in range mode,
-		/// by checking against the middle step.
+		// by checking against the middle step.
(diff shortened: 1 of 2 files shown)


promql/test.go
```diff
@@ -461,7 +461,7 @@ func (t *Test) exec(tc testCommand) error {
 		}
 
 		// Check query returns same result in range mode,
```
cstyan: I think the `,` at the end could be removed as well.

yuxiaobo96: Sorry, I can't understand what you mean.

cstyan: I don't think the `,` at the end of `... result in range mode,` is necessary. The previous comment for this function was different so maybe this one was incomplete or edited a few times but this was left in unintentionally. cc @brian-brazil

brian-brazil: I think either way works

yuxiaobo96: Ok, I have deleted this line of comments.

brian-brazil: This comment should remain

yuxiaobo96: Ok, I am going to modify it now.

brian-brazil:
Thanks!

PR #6017, opened 2019-09-12
added EC2 InstanceLifecycle info as label (new label called lifecycle)

aesiliato:
Signed-off-by: aesiliato

NB:
AWS API shows InstanceLifecycle in ec2 describe instances result only if the instance is a spot one. The new feature set value of lifecycle label = normal if InstanceLifecycle doesn't exist on describe instances results.


beorn7:
Superseded by #6914

PR #6020, opened 2019-09-13
*: bump Go version to 1.13

simonpasquier:


brian-brazil:
Something is up with the k8 dep.

simonpasquier:
`k8s.io/client-go@v12.0.0+incompatible` is an invalid reference for Go 1.13. When bumping k8s.io/client-go we should now point to `k8s.io/client-go@kubernetes-X.Y.Z` (from https://github.com/kubernetes/client-go/blob/master/INSTALL.md#add-client-go-as-a-dependency).

PR #6021, opened 2019-09-13
Garbage collect asynchronously in the WAL Watcher

### LINKED ISSUE #6007: Remote write backs up during compaction
## Bug Report
**What did you do?**
Ran remote_write code.

Continuous sending of samples to remote storage.

During a compaction samples will backup causing a several minute delay in some cases.

CompactionRWLag

**Environment**

* System information: Docker/Kubernetes
* Prometheus version: v2.12.0

csmarchbanks:
The WAL Watcher replays a checkpoint after it is created in order to
garbage collect series that no longer exist in the WAL. Currently the
garbage collection process is done serially with reading from the tip of
the WAL which can cause large delays in writing samples to remote
storage just after compaction occurs.

This also fixes a memory leak where dropped series are not cleaned up as
part of the SeriesReset process.

Closes #6007 

cc: @cstyan @tomwilkie

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -194,6 +194,7 @@ type QueueManager struct {
 	client         StorageClient
 	watcher        *wal.Watcher
 
+	seriesMtx            sync.Mutex
 	seriesLabels         map[uint64]labels.Labels
 	seriesSegmentIndexes map[uint64]int
 	droppedSeries        map[uint64]struct{}
@@ -264,6 +265,7 @@ func NewQueueManager(reg prometheus.Registerer, logger log.Logger, walDir string
 func (t *QueueManager) Append(samples []record.RefSample) bool {
 outer:
 	for _, s := range samples {
+		t.seriesMtx.Lock()
 		lbls, ok := t.seriesLabels[s.Ref]
 		if !ok {
 			t.droppedSamplesTotal.Inc()
@@ -271,8 +273,10 @@ outer:
 			if _, ok := t.droppedSeries[s.Ref]; !ok {
 				level.Info(t.logger).Log("msg", "dropped sample for series that was not explicitly dropped via relabelling", "ref", s.Ref)
 			}
+			t.seriesMtx.Unlock()
 			continue
 		}
+		t.seriesMtx.Unlock()
 		// This will only loop if the queues are being resharded.
 		backoff := t.cfg.MinBackoff
 		for {
@@ -356,9 +360,11 @@ func (t *QueueManager) Stop() {
 	t.watcher.Stop()
 
 	// On shutdown, release the strings in the labels from the intern pool.
+	t.seriesMtx.Lock()
 	for _, labels := range t.seriesLabels {
 		releaseLabels(labels)
 	}
+	t.seriesMtx.Unlock()
 	// Delete metrics so we don't have alerts for queues that are gone.
 	name := t.client.Name()
 	queueHighestSentTimestamp.DeleteLabelValues(name)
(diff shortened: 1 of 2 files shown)


storage/remote/queue_manager.go
```diff
@@ -193,9 +193,11 @@ type QueueManager struct {
 	client         StorageClient
 	watcher        *WALWatcher
 
-	seriesLabels         map[uint64]labels.Labels
-	seriesSegmentIndexes map[uint64]int
-	droppedSeries        map[uint64]struct{}
+	seriesLabels            map[uint64]labels.Labels
+	seriesLabelsMtx         sync.Mutex
```
YaoZengzeng: `sync.RWMutex` would be better?

csmarchbanks: I do not believe it is possible for multiple readers to be accessing any of these maps concurrently (there is only ever one `Append` call happening at a time), so a `sync.RWMutex` is unneeded.

storage/remote/wal_watcher.go
```diff
@@ -326,9 +327,17 @@ func (w *WALWatcher) watch(segmentNum int, tail bool) error {
 		case <-checkpointTicker.C:
 			// Periodically check if there is a new checkpoint so we can garbage
 			// collect labels. As this is considered an optimisation, we ignore
-			// errors during checkpoint processing.
-			if err := w.garbageCollectSeries(segmentNum); err != nil {
-				level.Warn(w.logger).Log("msg", "error process checkpoint", "err", err)
+			// errors during checkpoint processing, and do the process asynchronously.
```
cstyan: Can you leave another comment, something like `locking of the series labels within garbageCollectSeries isn't done until SeriesReset is called, so we can still process the current WAL segment while reading the checkpoint`.

csmarchbanks: Hmmm, where do you think the best place for a comment would be? Perhaps on the `writeTo` interface and say something about `SeriesReset` might be called concurrently with the other two?

csmarchbanks: I added a comment here, and at the `WriteTo` interface noting that the functions might be called concurrently. Note that garbage collection calls both `SeriesReset` and `StoreSeries` so there will be lots of concurrent calls happening.

csmarchbanks:
Red line shows lag every two hours before deploying this branch, green line shows no lag when using this branch.
PromRemoteWriteStability

tsdb/wal/watcher.go
```diff
@@ -345,9 +349,19 @@ func (w *Watcher) watch(segmentNum int, tail bool) error {
...
+			// errors during checkpoint processing. Doing the process asynchronously
+			// allows the current WAL segment to be processed while reading the
+			// checkpoint.
+			select {
+			case gcSem <- struct{}{}:
+				go func() {
+					if err := w.garbageCollectSeries(segmentNum); err != nil {
+						level.Warn(w.logger).Log("msg", "error process checkpoint", "err", err)
+					}
+					<-gcSem
```
beorn7: Perhaps use `defer` for this one? It's not needed now, but it might one day, and then this might get unnoticed here.

storage/remote/queue_manager.go
```diff
@@ -402,13 +410,16 @@ func (t *QueueManager) StoreSeries(series []record.RefSeries, index int) {
...
+	t.seriesMtx.Lock()
+	defer t.seriesMtx.Unlock()
 	// Check for series that are in segments older than the checkpoint
 	// that were not also present in the checkpoint.
 	for k, v := range t.seriesSegmentIndexes {
 		if v < index {
 			delete(t.seriesSegmentIndexes, k)
 			releaseLabels(t.seriesLabels[k])
 			delete(t.seriesLabels, k)
+			delete(t.droppedSeries, k)
```
beorn7: Was the absence of this line a bug (even before this commit)?

csmarchbanks: Yes, I have a note about it in the commit message/pull request description. Should I break it out into a different commit or even PR?

beorn7: No need, I just didn't read the description. Thanks.

PR #6023, opened 2019-09-14
Rule manager easy extend

liguozhong:
Background：
       Our team provided Prometheus as a service product. When we integrated rules, we encountered a problem. The rule of Prometheus is strongly dependent on the file system. Our product is strongly bound to k8s, and our rule is the PrometheusRule CRD of CoreOS. We have made some simple adjustments to rule manager without modifying the original logic of prometheus.

Change detail：
      Change the parameters of the ruleManager. Update method; make ruleManager easier to use. This makes it easier for other team products (prometheus as a service) to integrate Prometheus master source code.

Only minor changes are required in the main.go entry, and the original logic does not change：
image
But with minor modifications, our team will be able to integrate the Recoding Rule of the k8s CRD, rather than through the file system：
image
image
Other background：
        Our team leader want our team's R&D staff to modify the version of Prometheus and merge with the master. They don't want our team's R&D staff to develop independently. 
        It would be difficult for us to support our customers if the rulemanager did not modify the parameters of the Update method. They currently encounter Ip label with dimension explosion, which needs to be solved by Recoding Rule of prometheus. 
        we hope the prometheus Community can seriously consider the compatibility of this ruleManager.
       Please help review our PR. Thank you very much.


liguozhong:
@simonpasquier  Please help review our PR. Thank you very much.

liguozhong:
image
image
The github.com/prometheus/prometheus/discovery/file 33.333s in the make common-test script
In my development environment, common-test script can be tested and passed. But committing PR to GitHub failed to execute CI

liguozhong:
case：
There are 3.2 billion samples in our storage, and Prometheus reacts very slowly or even times out when we query data for a day or seven days. After a week's survey, we found that the data is too large and time-consuming in the data scanning process. Finally, we fix the problem by removing Ip label through recording rule

Recordrule is a very important feature for our Prometheus scenario. Our group has super-many servers, so Ip is super-many. Data is collected through the endpoint of each Ip, but when viewed by grafana, metrics information is only needed according to the application dimension.

brian-brazil:
Thanks for your PR, I'm not sure this makes sense in Prometheus internal API terms as it's making this component less of a black box. I'd suggest having something copy the files to disk.

3.6B samples is also pretty small, that's only 2.5k/s with 15d retention.

liguozhong:
@brian-brazil Thank you very much for review our PR.

Our team uses remote storage API to store our Prometheus data in the storage engine developed by our company, so as to achieve long-term storage. Our company needs to save 1 year of monitoring data, in order to better carry out the fault review and root cause analysis and historical data comparison, as well as some machine learning needs. In addition, we also have an old monitoring system through the combination of traditional storm+hbase+... , which stores data for one year.

At present, we only monitor the success rate, failure rate, Response time's 99 95 quantile, At present, only a few hundred servers are deployed, and there are super many servers that do not use Prometheus model monitoring solution. When all machines are monitored using Prometheus, the usage of storage will be much larger than that. When querying with the Remote storage api, the speed of data scanning becomes very, very slow

We really need this Recording Rule to solve the pre-aggregation of Ip labels brought about by the current super multi-server. Otherwise, it will be very difficult for our company to use Prometheus on a large scale to replace the original monitoring scheme.

Finally, there is a very important background.Using Prometheus as the next generation monitoring system is the top-down decision of the company's top level. And We want to be able to move forward with master's code, and we don't want to maintain a branch of code independently. 
We look forward to the community's core developers thinking carefully about giving Prometheus more development capabilities. Make Prometheus more open. Not as a black box. Like CoreOS's k8s Operator solution.Make Prometheus more open please.

brian-brazil:
Prometheus is already open, and the standard interface for this and several other parts of the code are files on disk. This is what the Operator does for example. I'd suggest working with this, rather than making source changes.

liguozhong:
We received it. Before finding a more open source approach, our team will temporarily maintain a small branch to ensure that our customers can use the recordingRule feature properly. 
Thank you for your patience.

PR #6026, opened 2019-09-15
fix recording inconsistency due to duplicate labels

### LINKED ISSUE #5529: Random value in recording rule with multiple identical series result

Harkishen-Singh:

Refers #5529. It **warns** the user from inconsistent data which might occur if more than one labels in the same group are found.

### DIFF
--- a/cmd/promtool/main.go
+++ b/cmd/promtool/main.go
@@ -304,72 +304,40 @@ func checkRules(filename string) (int, []error) {
 		numRules += len(rg.Rules)
 	}
 
-	showWarning := false
-	dLabels, dMetrics, nLabels, nMetrics := checkDuplicates(rgs.Groups)
-	if dLabels != 0 {
-		showWarning = true
-		fmt.Printf("%d duplicate label(s) found.\n", dLabels)
-		for _, n := range nLabels {
-			fmt.Println(n)
+	dRules := checkDuplicates(rgs.Groups)
+	if len(dRules) != 0 {
+		fmt.Printf("%d duplicate rules(s) found.\n", len(dRules))
+		for _, n := range dRules {
+			fmt.Printf("Metric: %s\nLabel(s):\n", n.metric)
+			for i, l := range n.label {
+				fmt.Printf("\t%s: %s\n", i, l)
+			}
 		}
-	}
-	if dMetrics != 0 {
-		showWarning = true
-		fmt.Printf("%d duplicate metric(s) found.\n", dMetrics)
-		for _, n := range nMetrics {
-			fmt.Println(n)
-		}
-	}
-	if showWarning {
 		fmt.Println("Might cause inconsistency while recording expressions.")
 	}
 
 	return numRules, nil
 }
 
-func checkDuplicates(r []rulefmt.RuleGroup) (int, int, []string, []string) {
-	duplicateLabels, duplicateMetrics := 0, 0
-	labelNames, metricName := []string{}, []string{}
-	for _, rule := range r {
-		var stackLabels []map[string]string
-		var stackMetrics []string
-		for _, props := range rule.Rules {
-			for _, labels := range stackLabels {
-				if reflect.DeepEqual(labels, props.Labels) {
-					duplicateLabels++
-					found := false
-					for _, inst := range labelNames {
-						if inst == labels["label"] {
-							found = true
-							break
-						}
-					}
-					if !found {
-						labelNames = append(labelNames, labels["label"])
-					}
-				}
-			}
-			stackLabels = append(stackLabels, props.Labels)
+type duplicateRules struct {
+	metric string
+	label  map[string]string
+}
 
-			for _, mname := range stackMetrics {
-				if reflect.DeepEqual(mname, props.Record) {
-					duplicateMetrics++
-					found := false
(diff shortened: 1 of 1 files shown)


Harkishen-Singh:
@brian-brazil  please have a look

cmd/promtool/main.go
```diff
@@ -303,9 +304,30 @@ func checkRules(filename string) (int, []error) {
 		numRules += len(rg.Rules)
 	}
 
+	duplicateLabels := checkDuplicateLabels(rgs.Groups)
+	if duplicateLabels != 0 {
+		fmt.Printf("%d duplicate label(s) found. Might cause inconsistency while recording expressions.\n", duplicateLabels)
```
brian-brazil: It'd be more useful to print the relevant labels. You should also check the metric name.

Harkishen-Singh: Sure. I will implement it in the following commit.

Harkishen-Singh:
@brian-brazil  implemented the suggestions! please have a look.

cmd/promtool/main.go
```diff
@@ -303,9 +304,74 @@ func checkRules(filename string) (int, []error) {
...
 }
 
+func checkDuplicates(r []rulefmt.RuleGroup) (int, int, []string, []string) {
+	duplicateLabels, duplicateMetrics := 0, 0
+	labelNames, metricName := []string{}, []string{}
+	for _, rule := range r {
+		var stackLabels []map[string]string
+		var stackMetrics []string
+		for _, props := range rule.Rules {
+			for _, labels := range stackLabels {
```
brian-brazil: A map would be more efficient, and easier to understant

cmd/promtool/main.go
```diff
@@ -303,9 +304,74 @@ func checkRules(filename string) (int, []error) {
...
+						}
+					}
+					if !found {
+						labelNames = append(labelNames, labels["label"])
+					}
+				}
+			}
+			stackLabels = append(stackLabels, props.Labels)
+
+			for _, mname := range stackMetrics {
```
brian-brazil: You need to check the metric names and labels at the same time, not independently.

cmd/promtool/main.go
Harkishen-Singh: sure

cmd/promtool/main.go
Harkishen-Singh: sure

Harkishen-Singh:
@brian-brazil  done! please have a look.

cmd/promtool/main.go
```diff
@@ -303,9 +304,42 @@ func checkRules(filename string) (int, []error) {
...
+	var stack, stackDuplicates []duplicateRules
+
+	for _, rule := range r {
+		for _, props := range rule.Rules {
+			for _, inst := range stack {
+				if reflect.DeepEqual(inst, duplicateRules{metric: props.Record, label: props.Labels}) {
+					stackDuplicates = append(stackDuplicates, duplicateRules{metric: props.Record, label: props.Labels})
+				}
+			}
+			stack = append(stack, duplicateRules{metric: props.Record, label: props.Labels})
```
brian-brazil: Building this up is odd, I'd suggest using an index variable in the for.

Harkishen-Singh: sure

Harkishen-Singh:
@brian-brazil, implemented index values in the necessary *for loops*. Please have a look.

cmd/promtool/main.go
brian-brazil: More exactly, get rid of the stack - you don't need it. Nested iteration over the rules will do it.

Harkishen-Singh: @brian-brazil, stack array is used only to check whether the duplicate rule has been already scanned in the earlier iteration since we aim to print the duplicate rule only once in the warning. 

Please note that stack does not implement the stack concept. It is surely misleading and i will change in the following commit.

Suppose if I remove the `stack array` (which is used for maintaining a check) and do simple nesting, then we might print a duplicated rule twice in the output like in the case when the label value is repeated more than twice.

Please correct me if I am getting it wrong.

Harkishen-Singh: @brian-brazil , done! pleae have a look.

Harkishen-Singh:
@brian-brazil , as suggested, I have removed the `stack array` and implemented the mechanism via nesting of rules! please review.

brian-brazil:
Thanks!

PR #6027, opened 2019-09-15
Allow users of scrape manager to configure context and custom roundtripper

bogdandrutu:
Please let me know if you are ok with this, and what kind of tests you would like to see for this.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -375,10 +375,10 @@ type ScrapeConfig struct {
 	MetricRelabelConfigs []*relabel.Config `yaml:"metric_relabel_configs,omitempty"`
 
 	// Ctx is the context that will be used to make all http scrape requests.
-	Ctx context.Context `-`
+	Ctx context.Context `yaml:"-"`
 	// WrapTransport will be invoked for custom HTTP behavior after the underlying
 	// transport is initialized.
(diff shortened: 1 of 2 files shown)


brian-brazil:
What are you trying to do here? This is the user-facing config file, and users can't specify either of those object types.

bogdandrutu:
I am using just the scraper component of the prometheus server, and would like to be able to configure extra things (like tracing, or special headers) for every request. Also I did use "-" to clarify that these configs are only possible when you start your own scraper.Manager

brian-brazil:
We don't generally add code that we don't use, nor have non-config things in our config structs, in order to keep things maintainable. If there's something that makes sense in Prometheus terms, it could be added directly.

bogdandrutu:
To make sure I understand your comment. Are you fine with adding these as options to the NewManager instead?

beorn7:
We revisited this during the bug scrub. In general, hooking in contexts and such is something that we could accept even if it currently not used by core Prometheus. However, things like changing the config struct are a bit too invasive.

Sorry for letting this PR rot. At this point, if this is still a need of yours, please follow up here so that we can see what next steps could be. For now, we are closing this.

PR #6030, opened 2019-09-17
vendor: Update json-iterator

gouthamve:
Also add explicit support for NaN/Inf

Signed-off-by: Goutham Veeramachaneni <gouthamve@gmail.com>

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -1266,10 +1266,23 @@ func marshalPointJSON(ptr unsafe.Pointer, stream *jsoniter.Stream) {
 	}
 	stream.WriteMore()
 	stream.WriteRaw(`"`)
-	stream.WriteFloat64(p.V)
+
+	// Taken from https://github.com/json-iterator/go/blob/master/stream_float.go#L71 as a workaround
+	// to https://github.com/json-iterator/go/issues/365 (jsoniter, to follow json standard, doesn't allow inf/nan).
+	buf := stream.Buffer()
+	abs := math.Abs(p.V)
+	fmt := byte('f')
+	// Note: Must use float32 comparisons for underlying float32 value to get precise cutoffs right.
+	if abs != 0 {
+		if abs < 1e-6 || abs >= 1e21 {
+			fmt = 'e'
+		}
+	}
+	buf = strconv.AppendFloat(buf, p.V, fmt, -1, 64)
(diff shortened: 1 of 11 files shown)


web/api/v1/api.go
```diff
@@ -1266,10 +1266,23 @@ func marshalPointJSON(ptr unsafe.Pointer, stream *jsoniter.Stream) {
 	}
 	stream.WriteMore()
 	stream.WriteRaw(`"`)
-	stream.WriteFloat64(p.V)
+
+	// Taken from https://github.com/json-iterator/go/blob/master/stream_float.go#L71 as a workaround
+	// to https://github.com/json-it
```
juliusv: Was this supposed to be a link to a GitHub issue? It looks truncated.

juliusv:
Could you elaborate on "Also add explicit support for NaN/Inf", how/where that's achieved?

juliusv:
(just wondering because the last time I tried to update this json-iter, it resulted in NaN/Inf/-Inf causing marshaling errors, so I downgraded it again)

juliusv:
Ah I guess that's just generally the manual `strconv.AppendFloat()`, taking that out of the hands of the now-buggy json-iter. But was there any other reason to update to the buggy json-iter version?

bboreham:
From my point of view it's to match https://github.com/prometheus/client_golang/pull/570

However the jsoniter change was deliberate so the other choice would be to fork it?

brian-brazil:
:+1:

web/api/v1/api.go
```diff
@@ -1266,10 +1266,23 @@ func marshalPointJSON(ptr unsafe.Pointer, stream *jsoniter.Stream) {
 	}
 	stream.WriteMore()
 	stream.WriteRaw(`"`)
-	stream.WriteFloat64(p.V)
+
+	// Taken from https://github.com/json-iterator/go/blob/master/stream_float.go#L71 as a workaround
```
bwplotka: Is this tested properly in `api_test.go`?

brian-brazil: I find your lack of faith... disturbing: https://github.com/prometheus/prometheus/blob/76769d4b617989c6c60f61a3ba16261f812538c2/web/api/v1/api_test.go#L1705

bwplotka: :cry: I am too lazy

PR #6033, opened 2019-09-18
tsdb: update chunks format

zhulongcheng:
https://github.com/prometheus/prometheus/blob/8744afdd1ea068bb6026a527983467952c7dbf5d/tsdb/chunks/chunks.go#L181-L190

there are 3 bytes padding after the format version.


YaoZengzeng:
Maybe you should rebase the code to pass CI :)

krasi-georgiev:
Thanks I noticed this padding too.

PR #6035, opened 2019-09-19
discovery/dns: add test case for SDConfig.UnmarshalYAML.

johncming:
No test code coverage:

https://github.com/prometheus/prometheus/blob/4f47806a7d33507dc9ff89c96cc41c91d3fd1d77/discovery/dns/dns.go#L74-L94

### DIFF
--- a/discovery/dns/dns_test.go
+++ b/discovery/dns/dns_test.go
@@ -23,9 +23,9 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/miekg/dns"
 	"github.com/prometheus/common/model"
-
 	"github.com/prometheus/prometheus/discovery/targetgroup"
 	"github.com/prometheus/prometheus/util/testutil"
+	"gopkg.in/yaml.v2"
 )
 
 func TestDNS(t *testing.T) {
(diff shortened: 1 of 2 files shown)


discovery/dns/dns_test.go
```diff
@@ -16,6 +16,7 @@ package dns
 import (
 	"context"
 	"fmt"
+	"gopkg.in/yaml.v2"
```
simonpasquier: could you move it the second group of imports (eg `github.com/go-kit/kit/log`)?

discovery/dns/dns_test.go
```diff
@@ -178,3 +179,96 @@ func TestDNS(t *testing.T) {
...
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			var config SDConfig
+			d := marshal(c.input)
+			err := config.UnmarshalYAML(unmarshal(d))
+			if (err != nil) != c.expectErr {
+				t.Errorf("test case %q: UnmarshalYAML expect error: %v, got: %v; %v", c.name, c.expectErr, err != nil, err)
```
simonpasquier: no need to repeat `c.name` in the error message since it is already passed to `t.Run`.

johncming: ok.

discovery/dns/dns_test.go
johncming: ok.

johncming:
> Thanks! A few nits otherwise LGTM.

updated.

discovery/dns/dns_test.go
```diff
@@ -23,9 +23,9 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/miekg/dns"
 	"github.com/prometheus/common/model"
-
 	"github.com/prometheus/prometheus/discovery/targetgroup"
 	"github.com/prometheus/prometheus/util/testutil"
+	"gopkg.in/yaml.v2"
```
simonpasquier: nit but it should be listed in this group:

```
	"github.com/go-kit/kit/log"
	"github.com/miekg/dns"
```

simonpasquier: the rule is

```
import (
  <stdlib packages>

  <3rd party packages>

  <github.com/prometheus/prometheus packages>
)
```

johncming: ok

discovery/dns/dns_test.go
```diff
@@ -178,3 +179,94 @@ func TestDNS(t *testing.T) {
 		})
 	}
 }
+
+func TestSDConfigUnmarshalYAML(t *testing.T) {
+
```
cstyan: nit, don't need newline here

johncming: ok.

johncming:
updated.

simonpasquier:
Thanks!

PR #6036, opened 2019-09-19
discovery/dns: Add code coverage.

johncming:
no test code coverage:

https://github.com/prometheus/prometheus/blob/4f47806a7d33507dc9ff89c96cc41c91d3fd1d77/discovery/dns/dns.go#L125-L126

add it.


simonpasquier:
could you do the change as part of #6035?

johncming:
> #6035

ok.

PR #6037, opened 2019-09-19
config: add test case for global scrape timeout.

johncming:
no test code coverage:

https://github.com/prometheus/prometheus/blob/a6a55c433cb9735923eebe52292b18db289c7f8a/config/config.go#L257


johncming:
> I would suggest to check that the scrape timeout of the job is indeed no larger than the scrape interval.

Added a test case.

simonpasquier:
Thanks!

PR #5991, opened 2019-09-05
tsdb: return err instead of panic for corrupted chunk

zhulongcheng:
Update https://github.com/prometheus/prometheus/issues/4128

Related to https://github.com/thanos-io/thanos/issues/1467

PR #6040, opened 2019-09-19
return err instead of panic for corrupted chunk 

krasi-georgiev:
check that the chunk segment has enough data to read all chunk pieces.

fixes: #5991 
fixes: https://github.com/thanos-io/thanos/issues/1467

@zhulongcheng - while reviewing your pr in https://github.com/prometheus/prometheus/pull/5991 there were many things I didn't understand so had to refactor and rename few variables to make the workflow more clear. Feel free to copy and paste the code from here or just review this PR.

Again sorry for hijacking the PR, I just did soo many changes to understand the code properly that it only made sense to open a separate PR


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Unreleased
+
+* [BUGFIX] Fix tsdb panic when querying corrupted chunks. #5991
```
zhulongcheng: pr number should be `#6040`

krasi-georgiev: yep, thanks.

tsdb/chunks/chunks.go
```diff
@@ -445,28 +480,50 @@ func (s *Reader) Size() int64 {
...
+		return nil, errors.Errorf("segment index %d out of range", sgmIndex)
 	}
-	chkS := s.bs[sgmSeq]
 
-	if sgmOffset >= chkS.Len() {
-		return nil, errors.Errorf("offset %d beyond data size %d", sgmOffset, chkS.Len())
+	chkBytes := s.bs[sgmIndex]
+
+	if sgmChunkOffset+maxChunkLengthFieldSize > chkBytes.Len()-SegmentHeaderSize {
+		return nil, errors.Errorf("segment doesn't include enough bytes to read the chunk size data field - required:%v, available:%v", sgmChunkOffset+maxChunkLengthFieldSize+SegmentHeaderSize, chkBytes.Len())
```
zhulongcheng: I think this should be 
```
if sgmChunkOffset+maxChunkLengthFieldSize > chkBytes.Len() {
```

tsdb/chunks/chunks.go
```diff
@@ -445,28 +480,50 @@ func (s *Reader) Size() int64 {
...
 	}
 	// With the minimum chunk length this should never cause us reading
 	// over the end of the slice.
-	chk := chkS.Range(sgmOffset, sgmOffset+binary.MaxVarintLen32)
-
-	chkLen, n := binary.Uvarint(chk)
+	c := chkBytes.Range(sgmChunkOffset, sgmChunkOffset+binary.MaxVarintLen32)
```
zhulongcheng: Use the `maxChunkLengthFieldSize` const?

tsdb/block_test.go
```diff
@@ -125,6 +128,45 @@ func TestCorruptedChunk(t *testing.T) {
...
+			nil,
+		},
+		"chunk not enough bytes to read the chunk length": {
+			func(f *os.File) {
+				// Truncate one byte after the segment header.
+				err := f.Truncate(chunks.SegmentHeaderSize + 1)
+				testutil.Ok(t, err)
+			},
+			nil,
+			errors.New("segment doesn't include enough bytes to read the chunk size data field - required:21, available:9"),
```
zhulongcheng: Would you mind clarifying why 21 bytes are required?  

Are 13 bytes enough to read the chunk size?  
// SegmentHeaderSize (8 bytes) + binary.MaxVarintLen32 (5 bytes)

related to comment https://github.com/prometheus/prometheus/pull/6040/files#r326134782

zhulongcheng:
Nice, and thanks for help this. 👍 

(just added some comments, hope these comments have a little help. I am sorry if these comments disturb the pr)

tsdb/chunks/chunks.go
krasi-georgiev: Aaah actually this needs to be 
```
if sgmChunkOffset+maxChunkLengthFieldSize > chkBytes.Len()+SegmentHeaderSize {
```

`chkBytes.Len()` doesn't include the size of the segment header. The minimum offset is always 8bytes which is the size of the segment header and need to account for that.
Here is the calculation with an offset 8.
8+5 > 5+8

It will hit the error if the chunk bytes are 4 which is not enough to read the chunks length field

krasi-georgiev: aaah I just double checked and you are right `chkBytes.Len()` does include the segment header so your calculation is correct. Thanks

tsdb/chunks/chunks.go
krasi-georgiev: Yep, thanks

tsdb/block_test.go
krasi-georgiev: yes correct, updated.

krasi-georgiev:
@zhulongcheng updated, thanks for the review!

tsdb/chunks/chunks.go
```diff
@@ -285,17 +295,35 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
...
+			reqSpace += maxChunkLengthFieldSize     // The data length is a variable length field so use the maximum possible value.
+			reqSpace += chunkEncodingSize           // The chunk encoding.
+			reqSpace += int64(len(c.Chunk.Bytes())) // The data itself.
+			reqSpace += crc32.Size                  // The 4 bytes of crc32
+		}
+
+		// Cutting a segment will not help here as
+		// this chunk will not fit even in a new segment.
+		if reqSpace > w.segmentSize {
+			return errors.New("chunks size larger than a full segment file")
```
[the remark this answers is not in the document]
krasi-georgiev: yep that is a good idea, I will refactor.

tsdb/chunks/chunks.go
```diff
@@ -445,28 +482,52 @@ func (s *Reader) Size() int64 {
...
+	if chkEnd > sgmBytes.Len() {
+		return nil, errors.Errorf("segment doesn't include enough bytes to read the chunk - required:%v, available:%v", chkEnd, sgmBytes.Len())
+	}
+
+	sum := sgmBytes.Range(chkEnd-crc32.Size, chkEnd)
+	s.crc32.Reset()
+	if _, err := s.crc32.Write(sgmBytes.Range(chkEncStart, chkDataEnd)); err != nil {
+		return nil, err
+	}
+	if s := s.crc32.Sum(s.buf[:0]); !bytes.Equal(s, sum) {
```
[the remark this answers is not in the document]
krasi-georgiev: suggestions?

tsdb/chunks/chunks.go
```diff
@@ -285,17 +295,35 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
...
+	// First chunk so no need for other checks.
+	if w.wbuf == nil {
+		cutChunk = true
+	} else {
+		// Calculate maximum required space and cut a new segment when
+		// the chunks won't fit in the current segment.
+
+		// Each segment contains: a header + many chunks.
+		// Each chunk contains: data length + encoding + the data itself + crc32
+		reqSpace := int64(SegmentHeaderSize) // The segment header.
```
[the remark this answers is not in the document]
krasi-georgiev: the segment available size is not the full `w.segmentSize`, but always `w.segmentSize - SegmentHeaderSize` , that is why if all chunks total at 100 and the `segmentSize` is 100 they will not fit in this segment because 8 bytes are already used by the segment header.

I think it will be more clear to remove it from here and put it like:
```
if reqSpace > w.segmentSize - SegmentHeaderSize {
```

Let me know if your understanding is different.

krasi-georgiev:
> If I understand correctly, the panic is avoided by...

correct

tsdb/chunks/chunks.go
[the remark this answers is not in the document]
codesome: We are checking `w.n+reqSpace > w.segmentSize`, and the segment size is already included inside `w.n`, so we are double-counting it there. And yes, `if reqSpace > w.segmentSize - SegmentHeaderSize {` would be better.

tsdb/chunks/chunks.go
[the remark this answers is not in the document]
codesome: actSum, expSum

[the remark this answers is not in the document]
codesome: Mainly because `s` also corresponds to the `Reader` (no idea why it is `s` for the `Reader`, should have been `r` instead :stuck_out_tongue:).

tsdb/chunks/chunks.go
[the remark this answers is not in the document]
krasi-georgiev: yes but this is different in one case cutting a new segment will help and in the other it won't, but now with your idea to add as many chunks as possible and than cut a new segment there will be no need for this check. very good suggestion and I will refactor soon.

tsdb/chunks/chunks.go
```diff
@@ -445,28 +482,52 @@ func (s *Reader) Size() int64 {
 // Chunk returns a chunk from a given reference.
 func (s *Reader) Chunk(ref uint64) (chunkenc.Chunk, error) {
 	var (
-		sgmSeq    = int(ref >> 32)
-		sgmOffset = int((ref << 32) >> 32)
+		// Get the upper 4 bytes.
+		// These contain the segment index.
```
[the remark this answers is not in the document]
krasi-georgiev: I am refering to the `4 bytes` that is why it is plural

tsdb/chunks/chunks.go
```diff
@@ -445,28 +482,52 @@ func (s *Reader) Size() int64 {
...
+		sgmIndex = int(ref >> 32)
+		// Get the lower 4 bytes.
+		// These contain the segment offset where the data for this chunk starts.
```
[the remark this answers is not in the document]
krasi-georgiev: the same here

tsdb/chunks/chunks.go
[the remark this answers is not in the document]
krasi-georgiev: renamed to `act` , it is only used in that block so it is ok to have it short.

tsdb/chunks/chunks.go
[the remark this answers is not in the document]
krasi-georgiev: after working on this today I realized that we don't have to error even if a single chunk is bigger than the segment size. The segment size is just advisory and it is ok to be bigger than the maximum set size. And it is quite unlikely that a single chunk would be that big to pass the max size of a segment.

krasi-georgiev:
ping @codesome

krasi-georgiev:
@codesome @zhulongcheng would appreciate one final review before merging this.

codesome:
Taking a look at this today

tsdb/chunks/chunks.go
```diff
@@ -445,28 +495,52 @@ func (s *Reader) Size() int64 {
...
 	if n <= 0 {
 		return nil, errors.Errorf("reading chunk length failed with %d", n)
 	}
-	chk = chkS.Range(sgmOffset+n, sgmOffset+n+1+int(chkLen))
 
-	return s.pool.Get(chunkenc.Encoding(chk[0]), chk[1:1+chkLen])
+	chkEncStart := sgmChunkStart + n
+	chkEnd := chkEncStart + chunkEncodingSize + int(chkDataLen) + crc32.Size
+	chkDataStart := chkEncStart + chunkEncodingSize
+	chkDataEnd := chkEnd - crc32.Size
```
codesome: Took some time to understand how these fit below, but definitely makes it more readable now.

tsdb/chunks/chunks.go
```diff
@@ -445,28 +495,52 @@ func (s *Reader) Size() int64 {
...
+	}
+
+	sum := sgmBytes.Range(chkEnd-crc32.Size, chkEnd)
+	s.crc32.Reset()
+	if _, err := s.crc32.Write(sgmBytes.Range(chkEncStart, chkDataEnd)); err != nil {
+		return nil, err
+	}
+	if act := s.crc32.Sum(s.buf[:0]); !bytes.Equal(act, sum) {
+		return nil, errors.Errorf("unexpected checksum %x, expected %x", act, sum)
+	}
```
codesome: This might have some effect on performance, but I am not sure why it was not checked before.

tsdb/chunks/chunks.go
```diff
@@ -284,18 +299,44 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
...
+	for _, chk := range chks {
+		// Each chunk contains: data length + encoding + the data itself + crc32
+		chksBatchSize += int64(maxChunkLengthFieldSize) // The data length is a variable length field so use the maximum possible value.
+		chksBatchSize += chunkEncodingSize              // The chunk encoding.
+		chksBatchSize += int64(len(chk.Chunk.Bytes()))  // The data itself.
+		chksBatchSize += crc32.Size                     // The 4 bytes of crc32
+
+		end++
+		if chksBatchSize+w.n > w.segmentSize {
+			if err := w.writeChunks(chks[:end], true); err != nil {
```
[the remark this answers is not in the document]
krasi-georgiev: > A non-blocker observation here: we could write the chunks till end-1 (or till end) without cutting a new segment. Given that logic exists only here, we could take out the cutChunk boolean from writeChunks, write till end-1 (or end) to the existing segment, and have w.cut() in this function (WriteChunks) after writing the segments.


it complicated the logic a bit as you would also need to check for `w.wbuf == nil` and doing this in the parent complicates the workflow a bit.

> If the last chunk crossed the segmentSize, shouldn't we consider chunks till end-1 such that it fits within given segment size?

> If the last chunk crossed the segmentSize, shouldn't we consider chunks till end-1 such that it fits within given segment size?

this complicated the logic a bit as well. Imaging that the first chunk is bigger than the segment size than you would have `chks[0:0]`

tsdb/chunks/chunks.go
krasi-georgiev: yes I agree and if you have suggestions how to simplify please comment.

tsdb/chunks/chunks.go
```diff
@@ -284,18 +299,44 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
...
+	return w.writeChunks(chks, false)
+}
+
+func (w *Writer) writeChunks(chks []Meta, cutChunk bool) error {
+	if len(chks) == 0 {
+		return nil
+	}
+	// w.wbuf == nil means it is the first chunk
+	// so need to start a new segment.
+	if cutChunk || w.wbuf == nil {
```
zhulongcheng: `chks` should be written to the current segment.

if `cutChunk` is true, cut after writing `chks`?

tsdb/db_test.go
```diff
@@ -2434,3 +2436,109 @@ func TestDBReadOnly_FlushWAL(t *testing.T) {
...
+	}{
+		// Last chunk ends at the segment boundary so
+		// all chunks should fit in a single segment.
+		{
+			chks: []chunks.Meta{
+				ch,
+				ch,
+				ch,
+			},
+			segmentSize:      chunks.SegmentHeaderSize + 3*len(ch.Chunk.Bytes()),
```
zhulongcheng: `segmentSize` should  include `data length`,  `encoding` and  `crc32`.

tsdb/chunks/chunks.go
codesome: Yes that will be done, see https://github.com/krasi-georgiev/prometheus/pull/4

tsdb/db_test.go
krasi-georgiev: yep , I also just noticed that and refactored.

krasi-georgiev:
@codesome is on leave and won't be able to review soon so @zhulongcheng waiting for your final review and will merge.

tsdb/chunks/chunks.go
```diff
@@ -47,8 +47,8 @@ const (
 // Chunk fields constants.
 const (
 	// The data length is a variable length field so use the maximum possible value.
-	maxChunkLengthFieldSize = binary.MaxVarintLen32
-	chunkEncodingSize       = 1
+	MaxChunkLengthFieldSize = binary.MaxVarintLen32
+	ChunkEncodingSize       = 1
```
zhulongcheng: Add comments for exported constants?

tsdb/chunks/chunks.go
```diff
@@ -338,8 +338,12 @@ func (w *Writer) WriteChunks(chks ...Meta) error {
...
 			chksBatchSize = 0
 			end = 0
-			if err := w.cut(); err != nil {
-				return err
+			// Cut a new segment only when there are more chunks to write.
+			// This avoids creating a new empty segment.
+			if len(chks) > 0 {
+				if err := w.cut(); err != nil {
+					return err
+				}
```
zhulongcheng: Nice!

tsdb/chunks/chunks.go
krasi-georgiev: thanks, updated

tsdb/chunks/chunks.go
```diff
@@ -284,27 +301,75 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
...
+		chksBatchSize += crc32.Size                     // The 4 bytes of crc32
+
+		end++
+		if chksBatchSize+w.n > w.segmentSize {
+			if end > 1 {
+				// Don't include the last chunk only if there are >1 chunks.
+				// This will keep segment size within the configured limit.
+				// If a single chunks is bigger than the configured limit,
+				// we cannot do much.
+				end--
```
zhulongcheng: Because we try to keep the segment size within the configured limit.  
So, if `end == 1` and if the current segment has chunks, then cut a new segment for new chunks.

tsdb/chunks/chunks.go
```diff
@@ -284,27 +301,75 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
...
+		chksBatchSize += ChunkEncodingSize              // The chunk encoding.
+		chksBatchSize += int64(len(chk.Chunk.Bytes()))  // The data itself.
+		chksBatchSize += crc32.Size                     // The 4 bytes of crc32
+
+		end++
+		if chksBatchSize+w.n > w.segmentSize {
+			if end > 1 {
+				// Don't include the last chunk only if there are >1 chunks.
+				// This will keep segment size within the configured limit.
+				// If a single chunks is bigger than the configured limit,
```
zhulongcheng: ` a single chunk` ?

tsdb/chunks/chunks.go
krasi-georgiev: sorry I didn't understand what you mean.

zhulongcheng: sorry, I will try to make it clear. (:

zhulongcheng: For example:    
    1. `w.segmentSize`: 1000   // configured limit
    2. `w.n`:  950                         // segment offset, there are some chunks in the current segment 
    3. the size of a new chunk:  120  

should write the new chunk to the current segment, or write to a new segment?

zhulongcheng: If our goal is to keep the segment size within the configured limit, the new chunk should be written to a new segment in this example.  right?

krasi-georgiev: Yep you are right and I have added tests to cover this condition as well and unfortunately this adds a lot more logic to the implementation. I don't like the implementation right now as it is too complex, so will look into it in few days with some fresh perspective. Please feel free to comment if you can see a way to simplify it and still pass all tests.

If we can't find a way to simplify the current implementation I would prefer to be not so strict about the segment size if this makes the implementation easier to understand.

zhulongcheng: Yeah it makes the implementation more difficult to understand.

> If we can't find a way to simplify the current implementation I would prefer to be not so strict about the segment size if this makes the implementation easier to understand.

Fully agree with you.

zhulongcheng: How do you think about this way?
```
		if chksBatchSize+w.n > w.segmentSize {
			if end == 1 && w.n > SegmentHeaderSize {
				if err := w.cut(); err != nil {
					return err
				}
				end--
			} else if end > 1 {
				end--
			}
			// . . .

```

https://github.com/zhulongcheng/prometheus/commit/7700541485589c65b768c54bc241917ba3b34426

krasi-georgiev: I just tried it and the tests fail, do the test pass at your end?

zhulongcheng: I noticed that we can't reset the `chks` inside for loop.

krasi-georgiev:
even after @zhulongcheng's cleanup it is still a nightmare to read. 

I am also thinking that if the implementation would be easier to read maybe the code shouldn't be so strict about the segment size. I don't see any real benefit in trying to keep the segments size so close to the maximum size. Chunks aren't so big so I don't see any problem in having few extra MBs over the max segment size.

zhulongcheng:
Yeah, easy to read is important. 

In TSDB, the sample number of a full chunk is about 120. Even if we write one more chunk into a segment file, the segment size will not increase too much.

So let's further simplify the code implementation?
I think we only need to remove the range handing inside the for loop.

krasi-georgiev:
@zhulongcheng , @codesome I now have a version that I believe is not that hard to understand. Let me know what you think.

@csmarchbanks maybe you can have a quick look as well if you have an idea how to simplify.

codesome:
In my list to review this week

krasi-georgiev:
Thanks appreciated.

krasi-georgiev:
@codesome just resolved the conflict with master to ready for a review when you find the time.

krasi-georgiev:
ping @zhulongcheng @codesome

tsdb/chunks/chunks.go
```diff
@@ -284,27 +301,85 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
...
+			cutNewBatch = batchSize+w.n > w.segmentSize
+			if cutNewBatch {
+				firstBatch = false
+			}
+		}
+
+		if cutNewBatch {
+			batchStart = i
+			batches = append(batches, []Meta{})
+			batchID++
```
zhulongcheng: reset `batchSize` ?

zhulongcheng: Add more chunks to test case 1:
```
		// 1:Two chunks can fit in a single segment can fit so the last one should result in a new segment.
		{
			chks: [][]chunks.Meta{
				[]chunks.Meta{
					chk1,
					chk2,
					chk3,
					chk4,
					chk5,
				},
			},
		// . . .
		},
```
batches: 
```
[[{Ref:0 Chunk:0xc00015a300 MinTime:1 MaxTime:1} {Ref:0 Chunk:0xc00015a320 MinTime:1 MaxTime:1}] [{Ref:0 Chunk:0xc00015a340 MinTime:1 MaxTime:1}] [{Ref:0 Chunk:0xc00015a360 MinTime:1 MaxTime:1}] [{Ref:0 Chunk:0xc00015a380 MinTime:1 MaxTime:1}]] 
```

krasi-georgiev: :) and when I thought I fixed all cases.
Updated.

tsdb/db_test.go
```diff
@@ -2502,30 +2502,32 @@ func TestChunkWriter(t *testing.T) {
...
+		// 		[]chunks.Meta{
+		// 			chk1,
+		// 			chk2,
+		// 			chk3,
+		// 		},
+		// 	},
+		// 	segmentSize:      3 * chunkSize,
+		// 	expSegmentSizes:  []int{3 * chunkSize},
+		// 	expSegmentsCount: 1,
+		// },
```
zhulongcheng: enable this test case?

tsdb/chunks/chunks.go
zhulongcheng: Nice!

tsdb/db_test.go
krasi-georgiev: yep

krasi-georgiev:
all comments addressed, thanks!

krasi-georgiev:
@codesome if you see any other problems please ping me and will open another PR

codesome:
Was planning to review today, but the changes looked fine earlier and the above approvals should be enough :)

krasi-georgiev:
It is not late :) if you see anything let me know and will open another PR.

tsdb/chunks/chunks.go
```diff
@@ -284,27 +311,87 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
...
+			batchSize = chkSize
+		}
+		batches[batchID] = chks[batchStart : i+1]
```
codesome: This should ideally be done when (1) We cut a new batch (2) When its the last chunk. Else it is going to create a new slide header for every chunk.

Additionally (not a blocker), we could write the chunks as soon as we hit this case instead of collecting multiple batches, what say?

krasi-georgiev: > This should ideally be done when (1) We cut a new batch (2) When its the last chunk. Else it is going to create a new slide header for every chunk.

hm, I am not sure I understand the idea, can you give an example or maybe even test it and if it passes the tests just open a PR and I will review quickly.


> Additionally (not a blocker), we could write the chunks as soon as we hit this case instead of collecting multiple batches, what say?

I think I tried this, but there was some other problem there, can't remember exactly. Maybe again try it and if it passes the tests than I will review quickly.

codesome: I thought getting a sub-slice in every iteration would cause extra allocations, but running `BenchmarkCompaction` shows me that apparently it does not. Additionally, I see some regression in performance in `ns/op` of nearly 7-9%, but I cannot say if it is this PR itself, so that would need some pprof action I suppose.

krasi-georgiev: Thanks for running the bench test. I did it this way as it was easyer to follow, and I also did a prombench test and didn't see any difference in the performance.

the regression might be due to the extra cheksum checking.

PR #6041, opened 2019-09-19
Refactor remote write queue tracking/restarts

cstyan:
Track remote write queues via a map so we don't have to track the index in the remote write array. The queue names in metrics now contain the hash as a string instead of the index in the config array, so they look like this:
```
prometheus_remote_storage_dropped_samples_total{queue="4d50f478186c47dfcffff29ffbc4b5d4:http://localhost:1235/receive"} 0
prometheus_remote_storage_dropped_samples_total{queue="97f072436c72649a7547e91184322b58:http://localhost:1234/receive"} 0
```

@csmarchbanks @YaoZengzeng 
cc @tomwilkie


cstyan:
closing in favour of #6043

PR #6042, opened 2019-09-20
extend remote read client to support stream type

YaoZengzeng:
fixes : #5926

Signed-off-by: YaoZengzeng <yaozengzeng@huawei.com>

### DIFF
--- a/storage/remote/client.go
+++ b/storage/remote/client.go
@@ -121,12 +121,16 @@ func (c Client) Name() string {
 }
 
 // Read reads from a remote endpoint.
-func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryResult, error) {
+func (c *Client) StartRead(ctx context.Context, query *prompb.Query) (*http.Response, error) {
 	req := &prompb.ReadRequest{
 		// TODO: Support batching multiple queries into one read request,
 		// as the protobuf interface allows for it.
 		Queries: []*prompb.Query{
 			query,
+		},
+		AcceptedResponseTypes: []prompb.ReadRequest_ResponseType{
+			prompb.ReadRequest_STREAMED_XOR_CHUNKS,
+			prompb.ReadRequest_SAMPLES,
 		},
 	}
 	data, err := proto.Marshal(req)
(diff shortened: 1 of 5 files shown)


YaoZengzeng:
@bwplotka PTAL:)

storage/remote/read.go
```diff
@@ -69,12 +76,88 @@ func (q *querier) Select(p *storage.SelectParams, matchers ...*labels.Matcher) (
...
+		res := &prompb.ChunkedReadResponse{}
+		err := stream.NextProto(res)
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, nil, fmt.Errorf("get next proto from stream failed: %v", err)
+		}
+
+		// TODO: log if return non 1 series in frame?
```
bwplotka: I think it's fine to skip it - it is server implementation detail that it limits to one for now - unless you algorithm will collapse if this invariant is not met..

storage/remote/read.go
```diff
@@ -69,12 +76,88 @@ func (q *querier) Select(p *storage.SelectParams, matchers ...*labels.Matcher) (
...
+				tmp = append(tmp, l.String())
+			}
+			currSeries = strings.Join(tmp, ";")
+			if currSeries != lastSeries {
+				lastSeries = currSeries
+				seriesNum++
+				seriesSet.set = append(seriesSet.set, &chunkSeries{labels: series.Labels, mint: q.mint, maxt: q.maxt})
+			}
+
+			seriesSet.set[seriesNum].chunks = append(seriesSet.set[seriesNum].chunks, series.Chunks...)
```
bwplotka: Do we really need to maintain `seriesNum`? It's as easy as getting last item in the slice, right?

storage/remote/client.go
```diff
@@ -145,41 +148,16 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
...
-	defer cancel()
-
 	httpResp, err := c.client.Do(httpReq.WithContext(ctx))
 	if err != nil {
 		return nil, errors.Wrap(err, "error sending request")
 	}
-	defer func() {
-		io.Copy(ioutil.Discard, httpResp.Body)
-		httpResp.Body.Close()
-	}()
```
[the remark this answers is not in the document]
YaoZengzeng: Actually close the http body here is not an appropriate way, because STREAM should keep it open.

So for SAMPLE, I close the http body when `handleSampledResponse` return.

But for STREAM, close the http body after reading all the series or encounter error.

WDYT? ;)

YaoZengzeng:
@bwplotka PTAL

storage/remote/client.go
```diff
@@ -145,41 +149,16 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
 	httpReq.Header.Set("User-Agent", userAgent)
 	httpReq.Header.Set("X-Prometheus-Remote-Read-Version", "0.1.0")
 
-	ctx, cancel := context.WithTimeout(ctx, c.timeout)
-	defer cancel()
-
-	httpResp, err := c.client.Do(httpReq.WithContext(ctx))
+	httpResp, err := c.client.Do(httpReq)
```
YaoZengzeng: Actually I don't encapsulate the httpReq with ctx. Because we need to keep the http connection alive, the ctx will be closed before we finished read all series.

YaoZengzeng:
Ping @bwplotka :)

web/api/v1/api_test.go
```diff
@@ -314,6 +308,24 @@ func TestEndpoints(t *testing.T) {
 		}
 
 		testEndpoints(t, api, false)
+	}
+
+	// Run all the API tests against a API that is wired to forward queries via
```
bwplotka: I find it a bit confusing, but I think it will work.. (: More opinions on those tests would be nice.

bwplotka:
Hello @YaoZengzeng are you planning to continue this? (:

roidelapluie:
There is an issue with this.

https://github.com/prometheus/prometheus/issues/6747
https://github.com/prometheus/prometheus/blob/fe76ccbfe3bb2a7549bbb1454c3a51ff4544c6d6/web/web.go#L594
If prometheis server start to use straming remote read, the other instance might never be able to shut down. Let's find an acceptable solution for the first issue.

Sudhar287:
@roidelapluie Do you mean that this issue should be taken forward only after #6747 is fixed?

bwplotka:
That's correct @Sudhar287 

But it is as easy as having extra context propagated that can be cancelled before `grpcSrv.GracefulStop() ` or something

bwplotka:
Let's close it as @YaoZengzeng said, @Sudhar287 can continue this. Ref: https://github.com/prometheus/prometheus/issues/5926#issuecomment-609533756

Feel free @Sudhar287 to use @YaoZengzeng commits for a good start :hugs: 

I don't think anything is blocked. I don't get @roidelapluie request. We don't even use gRPC here.

roidelapluie:
> I don't think anything is blocked. I don't get @roidelapluie request. We don't even use gRPC here.

Yeah, what did that guy say? Apparently this was a comment for another issue :)

bwplotka:
No worries Julien :smile:

Issue #5136, opened 2019-01-25
Catch when two targets are relabelled to the same labels


gouthamve:
When two different targets end up with the same labels after relabelling, two scrape loops are created and OOO ingestion errors are logged.

We should ideally catch this and log it. And create only one scrape loop per target labelset. 

This might break some setups as people might be abusing it already and are _safe_ if the two targets don't have any common metrics.

roidelapluie:
I would rather not scrape the targets than creating just one scrape config, as two instances can get the same labelset if you overwrite the metrics.

Also, we can have the issue with honor_label: true anyway ..

machine424:
Hello from the bug scrub!

The linked PR https://github.com/prometheus/prometheus/pull/9589 was closed.
We're still open to give this a try behind a feature flag, if someone is willing to pick up the PR (it was in its final stages).

shilohlee98:
Hi, I can give it a try and pick up the PR

PR #6043, opened 2019-09-20
Add config option for remote job name

cstyan:
Track remote write queues via a map so we don't have to track the index in the remote write array. The queue names in metrics now contain the hash as a string instead of the index in the config array, so they look like this:

```
prometheus_remote_storage_dropped_samples_total{queue="4d50f478186c47dfcffff29ffbc4b5d4:http://localhost:1235/receive"} 0
prometheus_remote_storage_dropped_samples_total{queue="97f072436c72649a7547e91184322b58
```

Also adds the option of setting a job name for remote write/read in config. If not set, the name will default to the first 6 characters of the config's `md5.Sum()`. The config parsing does not allow for duplicate job names, the same as scrape jobs.

Note: the base commit is from another PR (#6041)

@csmarchbanks

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -14,6 +14,9 @@
 package config
 
 import (
+	"crypto/md5"
+	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net/url"
@@ -626,6 +629,19 @@ func (c *RemoteWriteConfig) UnmarshalYAML(unmarshal func(interface{}) error) err
 	return c.HTTPClientConfig.Validate()
 }
 
+// ToHash generates a hash of the config struct minus it's name, as we consider configs
+// that are identical in all but name to still be identical.
+func (c *RemoteWriteConfig) ToHash() (string, error) {
+	temp := *c
+	temp.Name = ""
+	bytes, err := json.Marshal(temp)
+	if err != nil {
+		return "", err
+	}
+	hash := md5.Sum(bytes)
+	return hex.EncodeToString(hash[:]), nil
+}
+
 // QueueConfig is the configuration for the queue used to write to remote
 // storage.
 type QueueConfig struct {
@@ -680,4 +696,18 @@ func (c *RemoteReadConfig) UnmarshalYAML(unmarshal func(interface{}) error) erro
 	// We cannot make it a pointer as the parser panics for inlined pointer structs.
 	// Thus we just do its validation here.
 	return c.HTTPClientConfig.Validate()
+}
+
+// ToHash generates a hash of the config struct minus it's name, as we consider configs
+// that are identical in all but name to still be identical.
+// Copy of RemoteWriteConfigs ToHash
+func (c *RemoteReadConfig) ToHash() (string, error) {
+	temp := *c
+	temp.Name = ""
+	bytes, err := json.Marshal(temp)
+	if err != nil {
+		return "", err
+	}
+	hash := md5.Sum(bytes)
+	return hex.EncodeToString(hash[:]), nil
 }
--- a/config/config.go
+++ b/config/config.go
@@ -14,9 +14,6 @@
 package config
 
 import (
-	"crypto/md5"
-	"encoding/hex"
-	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net/url"
@@ -629,19 +626,6 @@ func (c *RemoteWriteConfig) UnmarshalYAML(unmarshal func(interface{}) error) err
 	return c.HTTPClientConfig.Validate()
 }
 
-// ToHash generates a hash of the config struct minus it's name, as we consider configs
-// that are identical in all but name to still be identical.
-func (c *RemoteWriteConfig) ToHash() (string, error) {
-	temp := *c
-	temp.Name = ""
-	bytes, err := json.Marshal(temp)
-	if err != nil {
-		return "", err
-	}
-	hash := md5.Sum(bytes)
-	return hex.EncodeToString(hash[:]), nil
-}
-
 // QueueConfig is the configuration for the queue used to write to remote
 // storage.
 type QueueConfig struct {
@@ -696,18 +680,4 @@ func (c *RemoteReadConfig) UnmarshalYAML(unmarshal func(interface{}) error) erro
 	// We cannot make it a pointer as the parser panics for inlined pointer structs.
 	// Thus we just do its validation here.
 	return c.HTTPClientConfig.Validate()
-}
-
-// ToHash generates a hash of the config struct minus it's name, as we consider configs
-// that are identical in all but name to still be identical.
-// Copy of RemoteWriteConfigs ToHash
-func (c *RemoteReadConfig) ToHash() (string, error) {
-	temp := *c
-	temp.Name = ""
-	bytes, err := json.Marshal(temp)
-	if err != nil {
-		return "", err
-	}
-	hash := md5.Sum(bytes)
-	return hex.EncodeToString(hash[:]), nil
 }
--- a/config/config.go
+++ b/config/config.go
@@ -265,27 +265,27 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		}
 		jobNames[scfg.JobName] = struct{}{}
 	}
-	rwJobNames := map[string]struct{}{}
+	rwNames := map[string]struct{}{}
 	for _, rwcfg := range c.RemoteWriteConfigs {
 		if rwcfg == nil {
 			return errors.New("empty or null remote write config section")
 		}
-		// Skip empty job names, we fill their name with their config hash in remote write code.
-		if _, ok := rwJobNames[rwcfg.JobName]; ok && rwcfg.JobName != "" {
-			return errors.Errorf("found multiple remote write configs with job name %q", rwcfg.JobName)
+		// Skip empty names, we fill their name with their config hash in remote write code.
+		if _, ok := rwNames[rwcfg.Name]; ok && rwcfg.Name != "" {
+			return errors.Errorf("found multiple remote write configs with job name %q", rwcfg.Name)
 		}
-		rwJobNames[rwcfg.JobName] = struct{}{}
+		rwNames[rwcfg.Name] = struct{}{}
 	}
-	rrJobNames := map[string]struct{}{}
+	rrNames := map[string]struct{}{}
 	for _, rrcfg := range c.RemoteReadConfigs {
 		if rrcfg == nil {
 			return errors.New("empty or null remote read config section")
 		}
-		// Skip empty job names, we fill their name with their config hash in remote read code.
-		if _, ok := rrJobNames[rrcfg.JobName]; ok && rrcfg.JobName != "" {
-			return errors.Errorf("found multiple remote read configs with job name %q", rrcfg.JobName)
+		// Skip empty names, we fill their name with their config hash in remote read code.
+		if _, ok := rrNames[rrcfg.Name]; ok && rrcfg.Name != "" {
+			return errors.Errorf("found multiple remote read configs with job name %q", rrcfg.Name)
 		}
-		rrJobNames[rrcfg.JobName] = struct{}{}
+		rrNames[rrcfg.Name] = struct{}{}
 	}
 	return nil
 }
@@ -596,7 +596,7 @@ type RemoteWriteConfig struct {
 	URL                 *config_util.URL  `yaml:"url"`
 	RemoteTimeout       model.Duration    `yaml:"remote_timeout,omitempty"`
 	WriteRelabelConfigs []*relabel.Config `yaml:"write_relabel_configs,omitempty"`
-	JobName             string            `yaml:"write_job_name,omitempty"`
+	Name                string            `yaml:"name,omitempty"`
 
 	// We cannot do proper Go type embedding below as the parser will then parse
 	// values arbitrarily into the overflow maps of further-down types.
@@ -655,7 +655,7 @@ type RemoteReadConfig struct {
 	URL           *config_util.URL `yaml:"url"`
 	RemoteTimeout model.Duration   `yaml:"remote_timeout,omitempty"`
 	ReadRecent    bool             `yaml:"read_recent,omitempty"`
-	JobName       string           `yaml:"write_job_name,omitempty"`
+	Name          string           `yaml:"name,omitempty"`
 
 	// We cannot do proper Go type embedding below as the parser will then parse
 	// values arbitrarily into the overflow maps of further-down types.
--- a/config/config.go
+++ b/config/config.go
@@ -265,15 +265,27 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		}
 		jobNames[scfg.JobName] = struct{}{}
 	}
+	rwJobNames := map[string]struct{}{}
 	for _, rwcfg := range c.RemoteWriteConfigs {
 		if rwcfg == nil {
 			return errors.New("empty or null remote write config section")
 		}
+		// Skip empty job names, we fill their name with their config hash in remote write code.
+		if _, ok := rwJobNames[rwcfg.JobName]; ok && rwcfg.JobName != "" {
+			return errors.Errorf("found multiple remote write configs with job name %q", rwcfg.JobName)
+		}
+		rwJobNames[rwcfg.JobName] = struct{}{}
 	}
+	rrJobNames := map[string]struct{}{}
 	for _, rrcfg := range c.RemoteReadConfigs {
 		if rrcfg == nil {
 			return errors.New("empty or null remote read config section")
 		}
+		// Skip empty job names, we fill their name with their config hash in remote read code.
+		if _, ok := rrJobNames[rrcfg.JobName]; ok && rrcfg.JobName != "" {
+			return errors.Errorf("found multiple remote read configs with job name %q", rrcfg.JobName)
+		}
+		rrJobNames[rrcfg.JobName] = struct{}{}
 	}
 	return nil
 }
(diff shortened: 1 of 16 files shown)


brian-brazil:
By the name I would expect this to have semantic meaning, in particular the same meaning as `job_name`. What's wrong with indexes?

cstyan:
Does it not have semantic meaning? A name can now be provided as a more descriptive way of identifying a remote write queue. The index is really only useful because of the ordering in the config file, which I've never particularly liked. It feels brittle. 

This way, as a user with remote configs that write to the same endpoint but that have different relabel or queue configs, I can name them appropriately. For example, the remote write config that uses relabelling to drop everything except dev metrics can have a name of `dev`.

The PR is also branched off another change I've submitted which would remove the index as part of the queue name in metrics, instead including a hash of the config.

brian-brazil:
> Does it not have semantic meaning?

No, it has no semantic effect as it's only used in metrics. Whereas `job_name` is a default for the `job` label.

cstyan:
The metric label `queue`, which is the combo of the `write_job_name` and the remote write `url`, is applied to all metrics related to a remote write queue. I suppose that could be changed to two labels `job` and `url`.

Do you disagree that having a name is not providing more value than just the index?

csmarchbanks:
I think being able to name queues would be helpful to identify a queue in metrics/logs rather than looking up the order in your configs. It is very similar to the name of a rule group, no real meaning, but nice for display and other internal purposes.

Rather than calling it `write_job_name`, perhaps just `name`, or `queue_name` to avoid any confusion with the idea of a job in Prometheus?

storage/remote/storage.go
```diff
@@ -68,8 +68,21 @@ func (s *Storage) ApplyConfig(conf *config.Config) error {
...
+	for _, rrConf := range conf.RemoteReadConfigs {
+		hash, err := toHash(rrConf)
+		if err != nil {
+			return nil
+		}
+
+		// Set the queue name to the config hash if the user has not set
+		// a name in their remote write config so we can still differentiate
+		// between queues that have the same remote write endpoint.
+		name := string(hash[:6])
```
csmarchbanks: I think the default value for the name should be the index. The index will convey much more information than a hash.

storage/remote/codec.go
```diff
@@ -538,3 +541,13 @@ func labelsToLabelsProto(labels labels.Labels, buf []prompb.Label) []prompb.Labe
 	}
 	return result
 }
+
+// Used for hashing configs and diff'ing hashes in ApplyConfig.
+func toHash(data interface{}) (string, error) {
```
csmarchbanks: `codec.go` seems like a weird place for this function, perhaps it should just go next to `ApplyConfig` in `storage.go`?

config/config.go
```diff
@@ -265,15 +265,27 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		}
 		jobNames[scfg.JobName] = struct{}{}
 	}
+	rwNames := map[string]struct{}{}
 	for _, rwcfg := range c.RemoteWriteConfigs {
 		if rwcfg == nil {
 			return errors.New("empty or null remote write config section")
 		}
+		// Skip empty names, we fill their name with their config hash in remote write code.
+		if _, ok := rwNames[rwcfg.Name]; ok && rwcfg.Name != "" {
```
csmarchbanks: What would you think of setting the default name while parsing? Then the remote code could depend on there always being a name.

Might make testing or re-use tricky since a name would always have to be present though...

storage/remote/codec.go
cstyan: Just put it there because we're using it in multiple places, `storage.go`, `write.go`, `write_test.go`, and `client_test.go`. Not a big deal to me where the function goes though.

storage/remote/storage.go
cstyan: I don't have a strong opinion. I don't tend to rely on the index.

config/config.go
cstyan: That would be fine imo.

storage/remote/storage.go
cstyan: We actually shouldn't really use the index, since reordering the individual RW/RR configs in the config file will no longer result in restarts, but the index will have changed.

config/config.go
cstyan: This would require the `toHash` function to be available in `config/config.go` by imports or declaring the function in `config`.

storage/remote/storage.go
csmarchbanks: If the index is used then reordering the config would need to result in restarting the queues. The restart-free reordering would only be available if a user specifies a name.

config/config.go
csmarchbanks: Good point. Lets keep the defaulting logic in the remote code.

storage/remote/storage.go
cstyan: One of the goals was to reduce the number of situations in which we restart queues when we don't really need to, so unless you or someone else feels really strongly about having the index as the default name then I think we should proceed as is.

storage/remote/client.go
```diff
@@ -117,7 +117,7 @@ func (c *Client) Store(ctx context.Context, req []byte) error {
 
 // Name identifies the client.
 func (c Client) Name() string {
-	return fmt.Sprintf("%d:%s", c.index, c.url)
+	return fmt.Sprintf("%s:%s", c.name, c.url)
```
csmarchbanks: Should this be split into two labels now, one `queue`, and one `url`?

cstyan: Yes I think with the addition of the queue name it makes sense for the current label to be split into two. `name` makes more sense as the label name I think, or maybe `rw_name`? Something that would match with the config option.

How much do we care about breaking people's dashboards that use the current label? At the very least the mixin would need to be updated as well.

beorn7:
@csmarchbanks how does this look to you now?

I feel not qualified to judge about the detailed implementation (unless I invested time to get into it), but I'd like to note that changing the queue names just because the config was reordered or a queue was added at the top rather than the bottom of the config. The queue name shows up in metrics, which is another reason why it is handy to have a meaning.

storage/remote/storage.go
csmarchbanks: We chatted for awhile and decided to stick with the hash as the default for the name.

csmarchbanks:
> I'd like to note that changing the queue names just because the config was reordered or a queue was added at the top rather than the bottom of the config.

Is this sentence missing a conclusion?

Generally this implementation seems correct to me, trying to think through some edge cases that might break. Right now:
1. If you have two duplicate remote write configs they would validate parsing then blow up metrics due to having the same hash. This seems pretty far-fetched in practice, should it be protected against?

cstyan:
@csmarchbanks Hmm, so right now we don't do anything to stop people from having multiple remote write queues with the exact same config. I think that should be changed.

It doesn't cause any kind of error/panic with the metrics, the values are just double (or slightly higher) than what they would be with two separate queues running.

I've added a check for duplicate configs, note that configs that are the same in everything except name are still considered duplicates, and a test for this.

storage/remote/write_test.go
```diff
@@ -15,16 +15,58 @@ package remote
...
+func TestNoDuplicateWriteConfigs(t *testing.T) {
+	dir, err := ioutil.TempDir("", "TestNoDuplicateWriteConfigs")
+	testutil.Ok(t, err)
+	defer os.RemoveAll(dir)
+
+	// Test that configs that are identical in all but name are considered identical.
+	cfg1 := config.RemoteWriteConfig{
+		Name: "write-1",
+		URL: &config_util.URL{
+			&url.URL{
```
beorn7: And here again (etc.), I'm afraid it needs to repeat the `URL` key.

storage/remote/client_test.go
```diff
@@ -65,11 +65,15 @@ func TestStoreHTTPErrorHandling(t *testing.T) {
...
+		conf := &ClientConfig{
 			URL:     &config_util.URL{URL: serverURL},
 			Timeout: model.Duration(time.Second),
-		})
+		}
+		hash, err := toHash(conf)
+		if err != nil {
+			t.Fatal(err)
+		}
+		c, err := NewClient(hash, conf)
```
csmarchbanks: Since this is just a test could a hardcoded string as a name work?

storage/remote/queue_manager.go
```diff
@@ -230,8 +227,8 @@ func NewQueueManager(reg prometheus.Registerer, logger log.Logger, walDir string
 		logger = log.NewNopLogger()
 	}
 
-	name := client.Name()
-	logger = log.With(logger, "queue", name)
+	// name := client.Name()
```
csmarchbanks: Extra commented code.

storage/remote/write.go
```diff
@@ -88,20 +87,17 @@ func (rws *WriteStorage) ApplyConfig(conf *config.Config) error {
 	rws.mtx.Lock()
 	defer rws.mtx.Unlock()
 
-	// Remote write queues only need to change if the remote write config or
-	// external labels change. Hash these together and only reload if the hash
-	// changes.
-	cfgBytes, err := json.Marshal(conf.RemoteWriteConfigs)
+	configHash, err := toHash(conf.RemoteWriteConfigs)
```
csmarchbanks: Is this necessary anymore or should we just let the loop go through a few times?

config/config.go
```diff
@@ -666,3 +697,17 @@ func (c *RemoteReadConfig) UnmarshalYAML(unmarshal func(interface{}) error) erro
 	// Thus we just do its validation here.
 	return c.HTTPClientConfig.Validate()
 }
+
+// ToHash generates a hash of the config struct minus it's name, as we consider configs
```
csmarchbanks: Why is name not included? Wouldn't we want to restart the component so the metrics pick up the new name?

cstyan: Hmm, yeah I need to change this a bit. The idea is that two remote write configs that are identical except for their names should also be considered duplicates.

storage/remote/write.go
cstyan: It's a pretty cheap short circuit to doing the rest of the loop, which is also pretty cheap. I'd keep it personally, but move it to `storage.go`'s `ApplyConfig` and include a similar check for the remote read config.

storage/remote/client_test.go
cstyan: not sure what you mean

config/config.go
cstyan: @csmarchbanks PTAL, still think we might be able to simplify the hashing/comparison but my brain is dead atm. The current check works.

csmarchbanks: Why do you want two remote write configs that are identical to be disallowed? I guess it doesn't make much sense to run that way, but moving Name logic to different places is extra complexity.

storage/remote/client_test.go
```diff
@@ -65,11 +65,15 @@ func TestStoreHTTPErrorHandling(t *testing.T) {
...
 		}
-
-		c, err := NewClient(0, &ClientConfig{
+		conf := &ClientConfig{
 			URL:     &config_util.URL{URL: serverURL},
 			Timeout: model.Duration(time.Second),
-		})
+		}
+		hash, err := toHash(conf)
+		if err != nil {
```
csmarchbanks: Can you use `testutil` in here rather than err checks?

storage/remote/client_test.go
csmarchbanks: I am not sure anymore...

config/config.go
cstyan: Minimize the chance for users to end up with bad config. If we allow duplicates, but with different names, it means both queues would be writing to the same endpoint with the exact same samples. At the very least they'd end up with lots of write errors in their logs.

Needing both hashing functions is unfortunate, but to me it's better than maintaining a second version of the config struct that just removes the name field, and making one generic hash function. I could rename this function to `HashWithoutName` so it's clearer that they're different functions?

csmarchbanks: :+1: Nit: `ToHash` -> `Hash` to be consistent with labels and such.

brian-brazil: > The idea is that two remote write configs that are identical except for their names should also be considered duplicates.

We don't consider that an issue for scrape configs, so I don't think we should consider it a problem here.

cstyan: I think what I mentioned above is reason enough to consider duplicates an issue for remote write even though they may not be for scrape.

brian-brazil: That's pretty much exactly what'd also happen with scrape configs, unless you know what you're up to.

beorn7: > We don't consider that an issue for scrape configs

That's not quite correct.

For one, we enforce a different job name for each scrape config, and by default, the job name becomes the `job` label for the scraped metric, so there isn't really a problem.

If, however, the config relabels the target labels so that they become identical, we _literally_ consider that an issue, namely [issue #5136](https://github.com/prometheus/prometheus/issues/5136).

I remote write config that differs in nothing but its name will create problems by definition, while a scrape config that differs in nothing but its name can very well be fine. That's a huge difference.

brian-brazil: > we literally consider that an issue, namely issue #5136.

That is a GIGO issue for the user, but not a problem for Prometheus.

beorn7: > That is a GIGO issue for the user, but not a problem for Prometheus.

I disagree. Of course, you are entitled to have your opinion. But please don't say "we" whan you mean "I".

I'm not sure if your statement is only about #5136.

I think it makes sense to reject duplicate remote-write configs with a different name. Do we have consensus on that?

brian-brazil: Every other place that it's been proposed to add something like that it's turned out not to be what we ended up doing, as either there's subtle cases where it's valid or it's a lot more complicated than it seems to catch all the corner cases. For example this could be relied upon for loadtesting, or there could be some weird load balancer involved.

I don't think we should start rejecting semantically valid configurations merely because they aren't necessarily sane.

#5136 is also about something subtly different, two targets with the same identity but not necessarily the same configuration.

beorn7: > Every other place that it's been proposed to add something like that it's turned out not to be what we ended up doing, as either there's subtle cases where it's valid or it's a lot more complicated than it seems to catch all the corner cases. For example this could be relied upon for loadtesting, or there could be some weird load balancer involved.

I leave it to the remote-write experts if they see anything problematic here. I don't see a corner case here is the configs are plainly identical. (We are not trying to catch cases where somebody happens to create different relabeling rules that happen to create the same output.) Loadtesting with all the samples duplicated would be problematic (I would loadtest the error detection), and I don't see how a loadbalancer could get involved in a reasonable way.

> I don't think we should start rejecting semantically valid configurations merely because they aren't necessarily sane.

Semantics is for us to define. I'd define identical configs as semantically invalid.

@cstyan @csmarchbanks what do you think?

brian-brazil: > Loadtesting with all the samples duplicated would be problematic

That depends on what you're loadtesting, it doesn't have to be the storage itself. It might be something like how well you deal with duplicate samples.

> I don't see how a loadbalancer could get involved in a reasonable way.

I said weird :)

> I'd define identical configs as semantically invalid.

I see no semantic issues here, nothing about it breaks our data, querying, or remote storage model. It works today and causes no problems for the Prometheus server. It may not have any "normal" use cases or be efficient, but I don't think we should be 2nd guessing what users put in their config file.

beorn7: > It might be something like how well you deal with duplicate samples.

This line of arguments would require the removal of all kind of correctness check. Should I now allow duplicates in expositions by client_golang because somebody might want to test how well a consumer deals with duplicate samples?

> It works today and causes no problems for the Prometheus server.

Again, with this line of argument, we could justify all kind of insane output of the Prometheus server.

I expect remote-write config duplication to happen by accident, and I want to protect my users from it. While accidental config duplication is likely to happen, valid use cases for duplicate config is all but speculation at this point. The course of action appears clear to me from that.

cstyan: I agree, I think it's semantically incorrect to accept identical configurations in the case of remote write when duplicates have implications for the downstream remote storage.

brian-brazil: > in the case of remote write when duplicates have implications for the downstream remote storage.

That's something that we get to define rather than the remote storages, and as we're going for at-least-once delivery remote storages already have to handle this semantically. Efficiency is not the same as semantics.

csmarchbanks: I think that providing some sanity checking of configuration would be nice, but I also think it will be weird that two configs with a different backoff is counted as "valid" but two configs with different names are rejected. As far as this PR goes, my vote would be that the name is included in any hash, and required to be unique due to the metrics we export. That would allow a user to just change the name of the queue in case they want duplicate samples being sent, but would guard against simple copy-paste without changing anything mistakes.

I am not opposed to the current implementation, just a preference to include the name in the hash. Finally, perhaps it would be good to open a separate issue/PR for sanity checking effectively duplicate configs more thoroughly?

storage/remote/write.go
cstyan: :point_up: This can be it's own PR

storage/remote/read_test.go
```diff
@@ -15,15 +15,99 @@ package remote
...
 	"github.com/prometheus/prometheus/storage"
+	"github.com/prometheus/prometheus/util/testutil"
 )
 
+func TestNoDuplicateReadConfigs(t *testing.T) {
+	dir, err := ioutil.TempDir("", "TestNoDuplicateReadConfigs")
+	testutil.Ok(t, err)
+	defer os.RemoveAll(dir)
+
+	// Test that configs that are identical in all but name are considered identical.
```
csmarchbanks: I think this comment is outdated.

storage/remote/write_test.go
```diff
@@ -15,16 +15,146 @@ package remote
...
+	},
+	QueueConfig: config.DefaultQueueConfig,
+}
+
+func TestNoDuplicateWriteConfigs(t *testing.T) {
+	dir, err := ioutil.TempDir("", "TestNoDuplicateWriteConfigs")
+	testutil.Ok(t, err)
+	defer os.RemoveAll(dir)
+
+	// Test that configs that are identical in all but name are considered identical.
```
csmarchbanks: I think this one is also outdated.

cstyan:
@csmarchbanks thanks, fixed. If everything passes I'll merge it.

PR #6044, opened 2019-09-21
scrape: close manager gracefully at end.

johncming:
`Manager` did not stop gracefully at the end and add test coverage.

https://github.com/prometheus/prometheus/blob/be9de64cda450fa2cb0c84a995fb0a76efbe4a36/scrape/manager_test.go#L337-L342

### DIFF
--- a/scrape/manager_test.go
+++ b/scrape/manager_test.go
@@ -339,6 +339,7 @@ func TestManagerTargetsUpdates(t *testing.T) {
 
 	ts := make(chan map[string][]*targetgroup.Group)
 	go m.Run(ts)
+	defer m.Stop()
 
 	tgSent := make(map[string][]*targetgroup.Group)
 	for x := 0; x < 10; x++ {


scrape/manager_test.go
```diff
@@ -336,6 +336,7 @@ scrape_configs:
 
 func TestManagerTargetsUpdates(t *testing.T) {
 	m := NewManager(nil, nil)
+	defer m.Stop()
```
juliusv: From a readability perspective it might be clearer to put this line right after `go m.Run(ts)`, since it's related to that and not needed for the scrape managers in the other tests that are constructed, but where `Run()` is never called.

johncming: yes. it looks like better to put stop after the run method

johncming: > From a readability perspective it might be clearer to put this line right after `go m.Run(ts)`, since it's related to that and not needed for the scrape managers in the other tests that are constructed, but where `Run()` is never called.

updated.

juliusv:
:+1: Thanks!

PR #6045, opened 2019-09-23
delete duplicate words and fix spelling

hwdef:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -332,7 +332,7 @@ func (c *GlobalConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -332,7 +332,7 @@ func (c *GlobalConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return nil
 }
 
-// isZero returns true iff the global config is the zero value.
+// isZero returns true if the global config is the zero value.
```
codesome: `iff` = `if and only if`. Same for the other `iff` in this PR. So not need to change them.

hwdef: ok.I will close this PR.

PR #6054, opened 2019-09-24
Refactor on discovery/manager_test.go

nevill:
As discussed with @cstyan [here](https://github.com/prometheus/prometheus/pull/5254#discussion_r326863103), I created this PR.
Changes including
- Use `testutil.ToFloat64` to collect testing metrics
- Declare `ServiceDiscoveryConfig` directly instead of calling `Unmarshal` on a piece of YAML

### DIFF
--- a/discovery/manager_test.go
+++ b/discovery/manager_test.go
@@ -25,11 +25,13 @@ import (
 	"time"
 
 	"github.com/go-kit/kit/log"
-	"github.com/prometheus/client_golang/prometheus"
-	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	common_config "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/config"
 	sd_config "github.com/prometheus/prometheus/discovery/config"
+	"github.com/prometheus/prometheus/discovery/consul"
+	"github.com/prometheus/prometheus/discovery/file"
 	"github.com/prometheus/prometheus/discovery/targetgroup"
 	"gopkg.in/yaml.v2"
 )
@@ -749,27 +751,33 @@ func verifyPresence(t *testing.T, tSets map[poolKey]map[string]*targetgroup.Grou
 }
 
 func TestTargetSetRecreatesTargetGroupsEveryRun(t *testing.T) {
-	cfg := &config.Config{}
-
-	sOne := `
-scrape_configs:
- - job_name: 'prometheus'
-   static_configs:
-   - targets: ["foo:9090"]
-   - targets: ["bar:9090"]
-`
-	if err := yaml.UnmarshalStrict([]byte(sOne), cfg); err != nil {
-		t.Fatalf("Unable to load YAML config sOne: %s", err)
-	}
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	discoveryManager := NewManager(ctx, log.NewNopLogger())
 	discoveryManager.updatert = 100 * time.Millisecond
 	go discoveryManager.Run()
 
-	c := make(map[string]sd_config.ServiceDiscoveryConfig)
-	for _, v := range cfg.ScrapeConfigs {
-		c[v.JobName] = v.ServiceDiscoveryConfig
+	c := map[string]sd_config.ServiceDiscoveryConfig{
+		"prometheus": sd_config.ServiceDiscoveryConfig{
+			StaticConfigs: []*targetgroup.Group{
+				&targetgroup.Group{
+					Source: "0",
+					Targets: []model.LabelSet{
+						model.LabelSet{
+							model.AddressLabel: model.LabelValue("foo:9090"),
+						},
+					},
+				},
+				&targetgroup.Group{
+					Source: "1",
+					Targets: []model.LabelSet{
+						model.LabelSet{
+							model.AddressLabel: model.LabelValue("bar:9090"),
+						},
+					},
+				},
+			},
+		},
 	}
 	discoveryManager.ApplyConfig(c)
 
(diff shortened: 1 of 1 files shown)


nevill:
I didn't replace `yaml.UnmarshalStrict` in `TestApplyConfigDoesNotModifyStaticProviderTargets`, instead of implementing a way to do deep clone, current implementation is still a good way.

@cstyan Please review it.

discovery/manager_test.go
```diff
@@ -749,46 +751,50 @@ func verifyPresence(t *testing.T, tSets map[poolKey]map[string]*targetgroup.Grou
...
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	discoveryManager := NewManager(ctx, log.NewNopLogger())
 	discoveryManager.updatert = 100 * time.Millisecond
 	go discoveryManager.Run()
 
 	c := make(map[string]sd_config.ServiceDiscoveryConfig)
-	for _, v := range cfg.ScrapeConfigs {
-		c[v.JobName] = v.ServiceDiscoveryConfig
+	c["prometheus"] = sd_config.ServiceDiscoveryConfig{
```
simonpasquier: (nit) you can collapse L760 and L761:

```
c := map[string]sd_config.ServiceDiscoveryConfig{
    "prometheus": sd_config.ServiceDiscoveryConfig{
     ...
```

The same comment applies to the rest of the change.

discovery/manager_test.go
```diff
@@ -749,46 +751,50 @@ func verifyPresence(t *testing.T, tSets map[poolKey]map[string]*targetgroup.Grou
...
+		StaticConfigs: []*targetgroup.Group{
```
cstyan: since the later `StaticConfigs` is just this `[]Group` without the second `Group`, can you define each once and then just reference them when assigning to `c["prometheus"]?

discovery/manager_test.go
```diff
@@ -801,43 +807,32 @@ scrape_configs:
...
+			&targetgroup.Group{
```
cstyan: same here, this matches the target group on lines 763 and 789, it could be defined once outside the tests

discovery/manager_test.go
```diff
@@ -937,99 +933,73 @@ scrape_configs:
...
-`
-	cfgTwo := &config.Config{}
-	if err := yaml.UnmarshalStrict([]byte(cfgTwoText), cfgTwo); err != nil {
-		t.Fatalf("Unable to load YAML config cfgTwo: %s", err)
-	}
-	c = make(map[string]sd_config.ServiceDiscoveryConfig)
-	for _, v := range cfgTwo.ScrapeConfigs {
-		c[v.JobName] = v.ServiceDiscoveryConfig
+	c["prometheus"] = sd_config.ServiceDiscoveryConfig{
+		StaticConfigs: []*targetgroup.Group{
```
cstyan: same

discovery/manager_test.go
nevill: When I play with the tests, I will change the config here and there. I don't think it requires to be compact here.
e.g. I may change this one to `foo1:9090` and later one to `foo2:9090`, it's a good idea to duplicate them here.

discovery/manager_test.go
nevill: you mean a common variable for tow tests ?
```go
var TG = targetgroup.Group {
    // targets: "foo:9090"
}

func TestA() {
    // use TG here
}
func TestB() {
    // use TG here
}
```
May not be a good idea.

cstyan: Something like this 
```
var fooTarget = &targetgroup.Group{
    Source: "0",
    Targets: []model.LabelSet{
        model.LabelSet{
	    "__address__": "foo:9090",
        },
    },
}

...

func TestTargetSetRecreatesEmptyStaticConfigs(t *testing.T) {
    // some setup
    c["prometheus"] = sd_config.ServiceDiscoveryConfig{
        StaticConfigs: []*targetgroup.Group{
            &fooTg,
        },
    }

    // check that the TG is present
    // change c["prometheus"] to an empty target group
    // check that length of targetgroups in discoveryManager is now 0
}

File: discovery/manager_test.go (hunk 2 above)
Comment (cstyan) [reply discovery/manager_test.go]: I think we should collapse the target group definitions until there's a need within the tests for them to actually be different.

File: discovery/manager_test.go (hunk 3 above)
Comment (nevill) [reply discovery/manager_test.go]: And also use `fooTarget` here ?
```go
func TestTargetSetRecreatesTargetGroupsEveryRun(t *testing.T) {
// ...
	c["prometheus"] = sd_config.ServiceDiscoveryConfig{
		StaticConfigs: []*targetgroup.Group{

			&fooTarget,  // <----

			&targetgroup.Group{
				Source: "1",
				Targets: []model.LabelSet{
					model.LabelSet{
						"__address__": "bar:9090",
					},
				},
			},
		},
	}
// ...
}
```
If yes, I don't think it's a good idea. The point is, these two tests will depend on the same target. What if I change it for `TestTargetSetRecreatesEmptyStaticConfigs`? I may break another test.

Plus, we have a lot duplication of this kind, I think they are reasonable.
e.g.
https://github.com/prometheus/prometheus/blob/08c55c119f39093e18b2bb9cba5c5619dc4ea0e1/discovery/manager_test.go#L111-L115
and
https://github.com/prometheus/prometheus/blob/08c55c119f39093e18b2bb9cba5c5619dc4ea0e1/discovery/manager_test.go#L141-L145

nevill: Apparently, we have some different ideas here, @beorn7 can you take a review please ?

simonpasquier: Personally the duplication in tests doesn't bother me so much.

cstyan: I don't feel strongly about changing it. This is good to go imo cc @simonpasquier

discovery/manager_test.go
beorn7: @nevill could you address this nit?

discovery/manager_test.go
```diff
@@ -25,11 +25,13 @@ import (
 	"time"
 
 	"github.com/go-kit/kit/log"
-	"github.com/prometheus/client_golang/prometheus"
-	dto "github.com/prometheus/client_model/go"
+	client_testutil "github.com/prometheus/client_golang/prometheus/testutil"
+	config_util "github.com/prometheus/common/config"
```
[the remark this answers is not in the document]
beorn7: Agreed. @nevill could you change this? It's just a nit…

discovery/manager_test.go
```diff
@@ -25,11 +25,13 @@ import (
 	"time"
 
 	"github.com/go-kit/kit/log"
-	"github.com/prometheus/client_golang/prometheus"
-	dto "github.com/prometheus/client_model/go"
+	client_testutil "github.com/prometheus/client_golang/prometheus/testutil"
```
beorn7: Is there really a need for a named import? `testutil` is not used as a name in this file, and I think it fits nicely in here.

nevill:
@beorn7 Thank you for help. Everything gets fixed now, please check and merge.

Issue #6053, opened 2019-09-24
Shouldn't PUT /-/quit and /-/reload return 403 Lifecycle APIs are not enabled ?


snebel29:
# Bug Report

I was working on https://github.com/prometheus/pushgateway/issues/136 and noticed that sending PUT vs POST to `/-/quit` and `/-/reload` when flag `--web.enable-lifecycle` is not enabled have different behavior

I Ran prometheus without `--web.enable-lifecycle` flag enabled and the [configuration](https://prometheus.io/docs/prometheus/latest/getting_started/#configuring-prometheus-to-monitor-itself) to scrape itself
```
$ ./prometheus
```
**What did you expect to see?**

I'm assuming that the desired behavior when you call `/-/quit` and `/-/reload` using PUT and POST method Prometheus without admin-lifecycle enabled should always return `403` and `Lifecycle APIs are not enabled`

**What did you see instead?**
When you chose to use PUT, you get default router `405` and `Method Not Allowed`
```
$ curl -XPUT http://localhost:9090/-/quit
Method Not Allowed
```
```
$ curl -XPUT http://localhost:9090/-/reload
Method Not Allowed
```
**Proposed solution**
Assuming that we want to have `403` and `Lifecycle APIs are not enabled`, we could simply add the handler for the missing verbs within the else clause below
https://github.com/prometheus/prometheus/blob/e36ab7e192b9a18e92f56ec51346997e67526811/web/web.go#L341-L355

I could create a PR if that's the case,

Thanks

beorn7:
This totally makes sense to me. The behavior should be consistent between verbs.
(Nit: Why is it "APIs" and not just "API"? Two endpoints can be part of the same API. I would call it "the lifecycle API" in singular.)

snebel29:
I guess you're technically correct on API vs APIs for this case, we could change the message from `Lifecycle APIs are not enabled` to `Lifecycle API is not enabled.`

PR #6057, opened 2019-09-25
Fix management api return code when lifecycle API is disabled

snebel29:
Fixes https://github.com/prometheus/prometheus/issues/6053


snebel29:
Not that this is of high importance but, is there any reason not to merge it yet?

PR #6059, opened 2019-09-25
Checkpoints with series chunks

codesome:
I am currently exploring the idea of having the chunks of series written down into the checkpoint rather than the records and checkpointing regularly (every 10-15m).

This is similar to what I am doing in https://github.com/cortexproject/cortex/pull/1103, and I have seen some good WAL replay times with this and without much increase in avg disk writes.

### DIFF
--- a/tsdb/chunkenc/xor.go
+++ b/tsdb/chunkenc/xor.go
@@ -62,7 +62,7 @@ func NewXORChunk() *XORChunk {
 
 // NewXORChunk returns a new chunk with XOR encoding of the given size.
 func NewXORChunkFromBytes(b []byte) *XORChunk {
-	return &XORChunk{b: bstream{stream: b, count: 0}}
+	return &XORChunk{b: newBReader(b)}
 }
 
 // Encoding returns the encoding type.
--- a/tsdb/chunkenc/xor.go
+++ b/tsdb/chunkenc/xor.go
@@ -60,6 +60,11 @@ func NewXORChunk() *XORChunk {
 	return &XORChunk{b: bstream{stream: b, count: 0}}
 }
 
+// NewXORChunk returns a new chunk with XOR encoding of the given size.
+func NewXORChunkFromBytes(b []byte) *XORChunk {
+	return &XORChunk{b: newBReader(b)}
+}
+
 // Encoding returns the encoding type.
 func (c *XORChunk) Encoding() Encoding {
 	return EncXOR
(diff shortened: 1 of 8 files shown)


csmarchbanks:
I think checkpointing the chunks could definitely be interesting and speed things up. However, checkpointing every 10-15 minutes would be a large change to remote write, and mean you only have 10-15 minutes to fix a remote target before Prometheus begins to drop data.

codesome:
Ah, I didn't know exactly how connected were TSDB's WAL and remote-write, now I know. In that case, we could skip truncation of WAL after the new checkpoint and rather truncate the WAL in the similar fashion as now. Additionally, add some info in the checkpoint which will tell the Head block from which WAL segment to start reading from in case of recovery.

codesome:
For anyone taking a peek at this while this is still WIP, here are some points about the current status:
1. I have temporarily removed reading of old checkpoint while I test the new checkpoint. If the new checkpoint makes it to the upstream, it would be with reading the old checkpoint.
2. Currently, I am able to write and read the new checkpoint successfully. But due to some edge case which I don't know yet, all the parameters are not being set properly after reading the new checkpoint and the queries get stuck. I am investigating this at the moment.

codesome:
Funnily, just after posting the above comment, the query started running with some corruption in data. So writing/reading is not happening properly yet :)

codesome:
Happy to say that with the last commit the writing and reading of the WAL is happening properly and I am able to see the recovery via queries on the UI.

Next up:
1. Write some unit tests for checkpoint write and recovery.
2. Make the processing of checkpoint records concurrent.
3. Don't delete the WAL segments and keep track of last segment in some other way in checkpoint as per [this comment](https://github.com/prometheus/prometheus/pull/6059#issuecomment-536406168) to avoid [this case](https://github.com/prometheus/prometheus/pull/6059#issuecomment-536208130)
4. Compare replay time of old checkpoint+WAL vs the new checkpoint+WAL
5. Estimate the increase in disk write and see the potential of spreading the checkpoint creation.

krasi-georgiev:
don't forget to do some research about SSD wear as I remember that there was some mentioning that the current the design was adjusted to reduce the wear. 

did a quick search and this came up
https://engineering.linkedin.com/blog/2016/05/designing-ssd-friendly-applications-for-better-application-perfo

brian-brazil:
The question there is if 1.6 bytes * number of checkpoints is going to be smaller than the ~12-13 bytes we do now.

codesome:
Some analysis on the disk writes that this new checkpoint would take: https://docs.google.com/spreadsheets/d/16SMmHAUAmS5eGVPGFCCQ2ou6bdXGjZgFTkjA7uEih8Y/edit?usp=sharing

The `MiB/s` estimations are just an average over the checkpoint duration, though the checkpointing would happen in a burst (but it can be spread throughout the checkpoint duration if we don't want to have disk write spikes).

Next up will be the comparison with current checkpoint disk writes.

codesome:
I will be attempting this on top of Head chunk mmapping to test out any gains in restart time. Closing this PR.

PR #6069, opened 2019-09-28
 fix the  wrong word

tanjunchen:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,7 @@ jobs:
     - run:
         command: make
         environment:
-          # Run garbage collection more aggresively to avoid getting OOMed during the lint phase.
+          # Run garbage collection more aggressively to avoid getting OOMed during the lint phase.
           GOGC: "20"
           # By default Go uses GOMAXPROCS but a Circle CI executor has many
           # cores (> 30) while the CPU and RAM resources are throttled. If we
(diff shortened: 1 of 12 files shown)


promql/engine.go
```diff
@@ -1350,7 +1350,7 @@ func (ev *evaluator) matrixIterSlice(it *storage.BufferedSeriesIterator, mint, m
 			ev.currentSamples++
 		}
 	}
-	// The seeked sample might also be in the range.
+	// The sought sample might also be in the range.
```
csmarchbanks: Seeked is a non-standard, but correct past tense of seek, and in these cases easier to read than changing the word to sought.

tsdb/querier_test.go
```diff
@@ -1062,7 +1062,7 @@ func TestChunkSeriesIterator_DoubleSeek(t *testing.T) {
 	testutil.Equals(t, float64(2), v)
 }
 
-// Regression when seeked chunks were still found via binary search and we always
+// Regression when sought chunks were still found via binary search and we always
```
csmarchbanks: Same as above, seeked is fine.

promql/engine.go
tanjunchen: OK Thank You

tsdb/querier_test.go
tanjunchen: OK Thank You

promql/engine.go
tanjunchen: I Changed it Any problem?

tanjunchen:
@csmarchbanks

PR #6074, opened 2019-09-29
Add DNS (TXT) auto discovery

gregoiremahe:
Hello,

Today, Prometheus has DNS auto discovery feature with A, AAAA and SRV record. These kind of record works well to discover layer4 and layer5 services/equipments, based on IPs or FQDN. (i.e. test.com)

The "problem" is, we can't do DNS auto discovery with protocol or sub path URL, like blackbox do to probe for example HTTP services with path.

So that's impossible to auto discover targets like http://test.com/api/v2 with DNS Service Discovery, which can be usefull for blackbox for example. (layer 7 services)
`GET 127.0.0.1:9115/probe?target=http://test.com/api/v2&module=http`

Having sub URLs and protocol is possible by using TXT records.

So I just added the TXT record to retrieve text from dns and be able to use it as target parameter with relabel_configs.

I have done the feature this way, but we are open to discuss about the changes I proposed if necessary.

### DIFF
--- a/discovery/dns/dns.go
+++ b/discovery/dns/dns.go
@@ -185,12 +185,21 @@ func (d *Discovery) refreshOne(ctx context.Context, name string, ch chan<- *targ
 		return model.LabelValue(net.JoinHostPort(a, fmt.Sprintf("%d", p)))
 	}
 
+	validTxtRecord := false
+	firstRecord := true
 	for _, record := range response.Answer {
 		var target model.LabelValue
 		var targetLabel model.LabelValue
+
 		switch addr := record.(type) {
 		case *dns.TXT:
-			targetLabel = model.LabelValue(addr.Txt[0])
+			if firstRecord || (!firstRecord && validTxtRecord) {
+				validTxtRecord = true
+				targetLabel = model.LabelValue(addr.Txt[0])
+			} else {
+				level.Warn(d.logger).Log("msg", "Invalid TXT record", "record", record)
+				continue
+			}
 		case *dns.SRV:
 			// Remove the final dot from rooted DNS names to make them look more usual.
 			addr.Target = strings.TrimRight(addr.Target, ".")
@@ -204,12 +213,20 @@ func (d *Discovery) refreshOne(ctx context.Context, name string, ch chan<- *targ
 			level.Warn(d.logger).Log("msg", "Invalid SRV record", "record", record)
 			continue
 		}
-		tg.Targets = append(tg.Targets, model.LabelSet{
-			model.AddressLabel:                target,
-			dnsNameLabel:                      model.LabelValue(name),
-			model.LabelName("__param_target"): targetLabel,
-			model.LabelName("instance"):       targetLabel,
-		})
+		if validTxtRecord {
+			tg.Targets = append(tg.Targets, model.LabelSet{
+				model.AddressLabel:                target,
+				dnsNameLabel:                      model.LabelValue(name),
+				model.LabelName("__param_target"): targetLabel,
+				model.LabelName("instance"):       targetLabel,
+			})
+		} else {
+			tg.Targets = append(tg.Targets, model.LabelSet{
+				model.AddressLabel: target,
+				dnsNameLabel:       model.LabelValue(name),
+			})
+		}
+		firstRecord = false
 	}
 
 	tg.Source = name
--- a/discovery/dns/dns.go
+++ b/discovery/dns/dns.go
@@ -82,6 +82,7 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		return errors.New("DNS-SD config must contain at least one SRV record name")
 	}
 	switch strings.ToUpper(c.Type) {
+	case "TXT":
 	case "SRV":
 	case "A", "AAAA":
 		if c.Port == 0 {
@@ -124,6 +125,8 @@ func NewDiscovery(conf SDConfig, logger log.Logger) *Discovery {
 		qtype = dns.TypeAAAA
 	case "SRV":
 		qtype = dns.TypeSRV
+	case "TXT":
+		qtype = dns.TypeTXT
 	}
 	d := &Discovery{
 		names:    conf.Names,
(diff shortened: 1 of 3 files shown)


brian-brazil:
Can you show that this is a standard SD mechanism? I've not come across this for SD before, SRV and A records are the official way to do this.

Generally we don't invent SD mechanisms. We also don't do business logic inside SD, that's for relabelling. Finally there is no ordering on RRs, so taking the first one is going to be non-deterministic.

gregoiremahe:
@brian-brazil : 
TXT is a also a standard to do SD : https://ns1.com/dns-service-discovery 
The pros of that is we are not limited to layer4 on OSI model. We can have "https://example.com/api/v2" for example, we are not limited to "example.com" only.

Then, i'm actually not taking the first record, this is just a check to be sure the first record was TXT. The boolean expression is "if this is either the first, either not but valid"

Okay for relabelling, I can remove it, the most important is to reuse TXT record result to use it as target parameter for blackbox

brian-brazil:
Having URLs in TXT records doesn't sounds like a standard SD mechanism to me, I'd suggest using file_sd here.

gregoiremahe:
We actually can have URLs in TXT records
- http://dns-sd.org/ServerStaticSetup.html
- http://sabre.io/dav/service-discovery/
- https://www.infoq.com/articles/rest-discovery-dns/

So this behavior is normalized.

Our need is to use dns service discovery to avoid redeploying our monitoring system on each equipment adding/removal on our infrastructure, which can be very recurrent

brian-brazil:
That's not what you've implemented, and we don't support the DNS-SD standard.

As this is a bespoke SD I'd suggest using file_sd to handle this.

gregoiremahe:
I can change the way I done it, that's not a problem as I said.
We just have to be agreed on the way to do it, but we really need dns sd to do this, not file_sd

SuperQ:
This is an interesting use case for targets because SRV records can't contain full URL paths, which would be necessary to have something like `target=example.com/foo` sent as a param.

discovery/dns/dns.go
```diff
@@ -198,10 +213,20 @@ func (d *Discovery) refreshOne(ctx context.Context, name string, ch chan<- *targ
...
 		}
-		tg.Targets = append(tg.Targets, model.LabelSet{
-			model.AddressLabel: target,
-			dnsNameLabel:       model.LabelValue(name),
-		})
+		if validTxtRecord {
+			tg.Targets = append(tg.Targets, model.LabelSet{
+				model.AddressLabel:                target,
+				dnsNameLabel:                      model.LabelValue(name),
+				model.LabelName("__param_target"): targetLabel,
```
SuperQ: This isn't something you should do in the code here. This makes the discovery method inflexible.

Instead, you should pass normal targets, and use relabel_configs to deal with the param.

docs/configuration/configuration.md
```diff
@@ -412,7 +412,35 @@ names:
...
+    metrics_path: /probe
+    params:
+      module: [http_2xx]
+    dns_sd_configs:
+      - names:
+         - your-record.example.com.
+        type: TXT
+        refresh_interval: 10s
+    relabel_configs:
+      - source_labels: []
```
SuperQ: This relabel config should be the same as the [standard example](https://github.com/prometheus/blackbox_exporter#prometheus-configuration). The TXT records provide the target results that get relabeled into `__param_target`.

brian-brazil:
> We just have to be agreed on the way to do it, but we really need dns sd to do this, not file_sd

The standard way for users to support bespoke SDs is via file_sd, we can't support everything natively and aren't accepting new SDs currently.

SuperQ:
After thinking about this some more, I agree that we shouldn't be inventing a new discovery method. But extending the existing dns_sd to support the TXT records in RFC 6763 DNS-SD, this would be a good idea.

There are a couple ways to do this.

1. Have the existing type "SRV" also query the TXT record.
2. Have a new "DNS-SD" type that does the SRV+TXT behavior.

brian-brazil:
I had similar thoughts, however it's unlikely that the information in DNS-SD would be useful metadata. Plus to be useful I think we need something in the multicast space which we were previously reluctant to touch.

And this still wouldn't help in this particular case, I don't think printer queues are something we need metadata for.

gregoiremahe:
I think using metadata in SRV+TXT should be useful. i.e. : 
_http._tcp.my-service-apis.example.com. 86400 IN SRV 0 5 5060 api1.example.com.
_http._tcp.my-service-apis.example.com. 86400 IN SRV 0 5 5060 api2.example.com.
_http._tcp.my-service-apis.example.com. 86400 IN SRV 0 5 5060 api3.example.com.
_http._tcp.my-service-apis.example.com.  IN TXT path=/api/version

will autodiscover : 
- http://api1.example.com/api/version
- http://api2.example.com/api/version
- http://api3.example.com/api/version

It's not exactly what we wanted, but doing this can match/fulfill our need. So let's do that.

The DNS Service Discovery via SRV implemented on Prometheus is not RFC compliant, by let's do something fully and stricly RFC compliant.

brian-brazil:
That's not too useful in Prometheus terms though, as you want the /metrics endpoint rather than the serving endpoint. I'd prefer any changes here be based on a more typical use case.

gregoiremahe:
@brian-brazil and what about prometheus/blackbox-exporter case ? It is not a matter of /metrics

brian-brazil:
mDNS was what I was thinking of in terms of multicase (that also seems to use DNS-SD), and https://github.com/msiebuhr/prometheus-mdns-sd covers that.

> and what about prometheus/blackbox-exporter case ? It is not a matter of /metrics

If you've already got to have a scrape_config per SRV record, you can add the path you want while you're at it.

simonpasquier:
I agree with @SuperQ that supporting DNS-based service discovery (RFC 6763) seems like a legit use case. I'd lean towards [option 2](https://github.com/prometheus/prometheus/pull/6074#issuecomment-539434264) (eg new type option, not confused with SRV).

beorn7:
Thanks for the discussion here. This PR is stale by now, and as it looks it wouldn't be the way forward anyway. In the bug scrub we decided to close this, which should not prevent implementation of one of the options discussed as viable above. But that would need to happen in a new PR.

PR #6075, opened 2019-09-29
add: check for rule files existence

haibeey:
Promtool outputs _got:nil_ when rule files are typed wrong or rules file does not exist. This is misleading as the message does not point developers to the issue. This commit aims to inform developers using promtool when rules files doesn’t matched any file.

Closes https://github.com/prometheus/prometheus/pull/6182

### DIFF
--- a/cmd/promtool/unittest.go
+++ b/cmd/promtool/unittest.go
@@ -145,6 +145,9 @@ func resolveAndGlobFilepaths(baseDir string, utf *unitTestFile) error {
 		if err != nil {
 			return err
 		}
+		if len(m) <= 0 {
+			fmt.Fprintln(os.Stderr, "  WARNING: no file match pattern", rf)
+		}
 		globbedFiles = append(globbedFiles, m...)
 	}
 	utf.RuleFiles = globbedFiles


cmd/promtool/unittest.go
```diff
@@ -145,6 +145,9 @@ func resolveAndGlobFilepaths(baseDir string, utf *unitTestFile) error {
 		if err != nil {
 			return err
 		}
+		if len(m) <= 0 {
```
codesome: I don't think returning an error in this case would be right as it is not necessary to always find a file for the pattern. A warning log would be better instead.

haibeey: returning an error forces the developer to check the file path or the glob pattern again since there was no match. just making a warning log might eventually get ignored. A developer would mostly want to match a specific file(s) with the glob pattern or file path.

codesome: I still feel we should not return an error here. Thoughts @brian-brazil?

simonpasquier:
I got bitten by this when playing with alert tests but I'm on the fence whether it should be a hard error or a warning. Whichever solution we choose, adding a test would be good.

brian-brazil:
Warning, as Prometheus doesn't error out on this.

haibeey:
Okay. Would make update ASAP.

simonpasquier:
LGTM

PR #6076, opened 2019-09-30
minor fix for making map

YaoZengzeng:
Signed-off-by: YaoZengzeng <yaozengzeng@huawei.com>

### DIFF
--- a/discovery/kubernetes/ingress.go
+++ b/discovery/kubernetes/ingress.go
@@ -142,7 +142,8 @@ const (
 )
 
 func ingressLabels(ingress *v1beta1.Ingress) model.LabelSet {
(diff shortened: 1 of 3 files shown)


cstyan:
why is the `2*` needed?

YaoZengzeng:
@cstyan because each label and annotation will create two key-value pair in the map.

bwplotka:
Maybe some commentary would be nice then, but otherwise LGM

cstyan:
:point_up: agreed, a comment would help

YaoZengzeng:
@codesome @bwplotka @cstyan Updated.

But why we need another key like `XXX_labelpresent_XXX=true` ?

PR #6077, opened 2019-09-30
Update Yoda station for a place

yuxiaobo96:

Delete redundant alias and update Yoda station for a place.

### DIFF
--- a/tsdb/index/postings_test.go
+++ b/tsdb/index/postings_test.go
@@ -841,5 +841,5 @@ func TestMemPostings_Delete(t *testing.T) {
 	deleted := p.Get("lbl1", "b")
 	expanded, err = ExpandPostings(deleted)
 	testutil.Ok(t, err)
-	testutil.Assert(t, 0 == len(expanded), "expected empty postings, got %v", expanded)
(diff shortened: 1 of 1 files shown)


cmd/prometheus/main.go
```diff
@@ -35,14 +35,14 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
-	conntrack "github.com/mwitkow/go-conntrack"
```
codesome: These aliases help in knowing what name to be use to refer the package (same goes for kingpin and yaml). Because those package names are not clear by just looking at the import path.

yuxiaobo96: Oh, I understand it.
Thank you.

yuxiaobo96: The Yoda station need to update in the document tsdb/index/postings_test.go.

yuxiaobo96: ？？

csmarchbanks:
The current code works fine, and there are no guidelines for avoiding yoda conditions in golang to my knowledge. IMO it is not worth merging this PR.

PR #6081, opened 2019-10-01
Benchmark for loading WAL

codesome:


codesome:
Makes sense. I have added those 3 cases.

codesome:
I would prefer not adding it in `tsdb/cmd/tsdb/main.go` as the cmd is user-facing and benchmarking only the WAL doesn't add much value there (does it? also benchmarking on the go while developing is easier this way). Also, it would be merged with promtool now and there is already benchmark to test the entire storage. Merging this for now and can move it to the cmd later if it is required.

PR #6083, opened 2019-10-01
replace chinese code  with english

yasongxu:
replace chinese code  with english
PTAL


brian-brazil:
We're purposefully testing UTF-8 here.

bo-er:
> We're purposefully testing UTF-8 here.

May I humbly ask why UTF-8 is not supported, and what will happen if we add Chinese characters into a prometheus alert rule file?

From my understanding, it's fine as long as I don't put UTF-8 characters into metrics, right?
Is it OK if I put some extra Chinese labels into an alert message?

roidelapluie:
UTF8 is supported in labels and annotation values. Alert names are label values.

PR #6084, opened 2019-10-01
Replaced t.Fatalf() with testutil.Assert() in buffer_test.go

joe-elliott:
- Replaced t.Fataf() calls with testutil.Assert()'s
- Work toward finishing https://github.com/prometheus/prometheus/issues/3242

### DIFF
--- a/storage/buffer_test.go
+++ b/storage/buffer_test.go
@@ -77,8 +77,12 @@ func TestSampleRing(t *testing.T) {
 						break
 					}
 				}
-				testutil.Assert(t, sold.t < s.t-c.delta || found, "%d: expected sample %d to be in buffer but was not; buffer %v", i, sold.t, buffered)
-				testutil.Assert(t, sold.t >= s.t-c.delta || !found, "%d: unexpected sample %d in buffer; buffer %v", i, sold.t, buffered)
+
+				if found {
+					testutil.Assert(t, sold.t >= s.t-c.delta, "%d: unexpected sample %d in buffer; buffer %v", i, sold.t, buffered)
+				} else {
+					testutil.Assert(t, sold.t < s.t-c.delta, "%d: expected sample %d to be in buffer but was not; buffer %v", i, sold.t, buffered)
+				}
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


storage/buffer_test.go
```diff
@@ -78,10 +78,10 @@ func TestSampleRing(t *testing.T) {
 					}
 				}
 				if sold.t >= s.t-c.delta && !found {
-					t.Fatalf("%d: expected sample %d to be in buffer but was not; buffer %v", i, sold.t, buffered)
+					testutil.Fatal(t, "%d: expected sample %d to be in buffer but was not; buffer %v", i, sold.t, buffered)
```
csmarchbanks: This change isn't quite right. The point of the helper is that you move the assertion into the same line. In this case, you should be using `testutil.Assert()`

util/testutil/testing.go
```diff
@@ -71,6 +71,12 @@ func Equals(tb TB, exp, act interface{}, msgAndArgs ...interface{}) {
 	}
 }
 
+// Fatal fails the test immediately.
+func Fatal(tb TB, msgAndArgs ...interface{}) {
```
csmarchbanks: I don't think this function should be needed. May as well just use `t.Fatalf`

storage/buffer_test.go
```diff
@@ -77,12 +77,8 @@ func TestSampleRing(t *testing.T) {
 						break
 					}
 				}
-				if sold.t >= s.t-c.delta && !found {
-					t.Fatalf("%d: expected sample %d to be in buffer but was not; buffer %v", i, sold.t, buffered)
-				}
-				if sold.t < s.t-c.delta && found {
-					t.Fatalf("%d: unexpected sample %d in buffer; buffer %v", i, sold.t, buffered)
-				}
+				testutil.Assert(t, sold.t < s.t-c.delta || found, "%d: expected sample %d to be in buffer but was not; buffer %v", i, sold.t, buffered)
```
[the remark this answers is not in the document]
joe-elliott: Agreed.  Thanks for the suggestion.

PR #6096, opened 2019-10-03
discovery/kubernetes: Expose label_selector and field_selector to config

shuttie:
The main motivation behind this PR is the performance/GC issue within grafana/loki (https://github.com/grafana/loki/issues/1102), which uses Prometheus discovery mechanism to pull new targets from Kubernetes API.

Currently Prometheus has the following k8s discovery flow:
* pull everything from k8s API
* register watch to receive updates
* during relabeling, drop all targets which are out of scope (for example, based on an annotation)

For Prometheus on a large Kubernetes clusters this may be a problem, so in case when you need to monitor only a small subset of pods, you still need to pull a giant blob of all the pods in all the states. In our case we have 2k running pods and about 10k completed pods,  which can easily take 1G of RAM.

For Loki, this is a bit different problem, as there is a `promtail` agent running on all the nodes, which is periodically polling for all the running pods in a cluster, and after that dropping all the targets not belonging to a current node. This options will allow to limit the logging agent to receive only events belonging to the current node.


discovery/kubernetes/kubernetes.go
```diff
@@ -91,6 +91,9 @@ type SDConfig struct {
 	Role               Role                         `yaml:"role"`
 	HTTPClientConfig   config_util.HTTPClientConfig `yaml:",inline"`
 	NamespaceDiscovery NamespaceDiscovery           `yaml:"namespaces,omitempty"`
+	LabelSelector      string                       `yaml:"label_selector,omitempty"`
+	FieldSelector      string                       `yaml:"field_selector,omitempty"`
+	ResyncPeriod       model.Duration               `yaml:"resync_period,omitempty"`
```
brian-brazil: Why does this need to be configurable?

docs/configuration/configuration.md
```diff
@@ -850,11 +850,25 @@ tls_config:
...
+
+# Optional label and filter selectors to limit the discovery process to a subset of available resources. 
+[ label_selector: <string> ]
+[ field_selector: <string> ]
 ```
 
 Where `<role>` must be `endpoints`, `service`, `pod`, `node`, or
 `ingress`.
 
+By default `label_selector` and `field_selector` are empty, so the discovery pulls all resources
```
brian-brazil: This is a bit more wordy than it needs to be, see what we say in consul. It'd be better to link the docs by the fields.

discovery/kubernetes/kubernetes.go
shuttie: @brian-brazil the ResyncPeriod?

brian-brazil: Yes

discovery/kubernetes/kubernetes.go
```diff
@@ -229,27 +235,39 @@ func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
 			e := d.client.CoreV1().Endpoints(namespace)
 			elw := &cache.ListWatch{
 				ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+					options.FieldSelector = d.fieldSelector
+					options.LabelSelector = d.labelSelector
```
brian-brazil: What happens if the endpoints match, but not the associated services or pods?

shuttie: As far as I understand, endpoint discovery is absolutely separate from pods/services, so it's up to the user what filters to apply. In a case when there are two blocks of discovery roles, like for endpoints and pods, then they will be discovered as separate targets.

brian-brazil: We could potentially miss the service/pod metadata though when doing endpoint discovery which would be unexpected for users I think.

shuttie: Can you please explain a bit a case when this situation is possible? 

As far as I understand, there are multiple roles for kubernetes discovery in Prometheus: node, service, pod, endpoints, ingress. Each of these roles maps 1-to-1 to a group of kubernetes resources. 

For endpoints, it queries the `/api/v1/namespaces/<ns>/endpoints/<name>` kubernetes API route. Endpoint is not a pod or service, it's a special type of internal intermediate resource used to map pods to services. The mapping is internal to kubernetes and not exposed to anywhere. So it can have it's own set of labels and fields we can filter with (however they are usually the same as for the corresponding pod).

In these particular lines of code we subscribe to all the updates for the `endpoints` route, `pods` and `services` are not touched anywhere here.

The only issue I see here is that user may have (for some reason) different set of selectors for pods and endpoints of the single service, and be surprised that the discovered IPs are different. But it seems quite logical for me: different filters => different discovered targets.

If you have an example when this filtering will not work or will confuse user, I would be really happy to take it into account and change the implementation.

brian-brazil: Read the rest of the case statement, just below this.

shuttie: Oops, got it, thanks for the hint. Will update the PR tomorrow to address this issue.

shuttie:
@brian-brazil I've changed the selector config syntax in a way, so you can now set per-role label/field selectors. Then, for the case you've described with service/pod/endpoint having different set of labels and fields, a user may set pod/service/endpoint specific selectors. Or just don't do anything and the default behavior will be intact.

brian-brazil:
Hmm, I think this might be a bit confusing for users. It may not be obvious that endpoint uses pod. Some better wording would help at least.

What do others think?

shuttie:
Maybe we should mention this in the docs? As it was also not really obvious even for me.

simonpasquier:
This looks useful to have. Having said that, I don't have a better idea to expose the settings. Maybe enforce at YAML unmarshaling that the selectors configuration are supported for the `role` (eg `role: pod` should allow `pod` selectors only, `role: endpoint` should allow `pod`, `endpoint` and `service`, ...).

This would also deserve some additional unit tests in [discovery/kubernetes](https://github.com/prometheus/prometheus/tree/master/discovery/kubernetes) IMO.

shuttie:
I've updated the implementation with the following modifications:
* there is now a proper validation process for selectors, so a user won't be able to supply pod selector for a node role.
* good/bad tests for all the possible roles and selectors.

As for additional discovery/kubernetes discovery tests, I have found no proper way to test the implementation, as the filtering is made server-side in kubernetes-apiserver (and with etcd somewhere deep inside). Here we only put additional filters in the API request.

discovery/kubernetes/kubernetes.go
```diff
@@ -111,6 +125,30 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	case "node":
+		if len(c.Selectors.Service.Field) > 0 || len(c.Selectors.Endpoints.Field) > 0 || len(c.Selectors.Ingress.Field) > 0 || len(c.Selectors.Pod.Field) > 0 {
+			return errors.Errorf("node role supports only node selectors")
+		}
+	case "ingress":
+		if len(c.Selectors.Service.Field) > 0 || len(c.Selectors.Endpoints.Field) > 0 || len(c.Selectors.Node.Field) > 0 || len(c.Selectors.Pod.Field) > 0 {
+			return errors.Errorf("ingress role supports only ingress selectors")
+		}
+	default:
+		return errors.Errorf("role is not one of: pod, service, endpoints, node, ingress")
```
brian-brazil: Can we remove the check on 118 now?

docs/configuration/configuration.md
```diff
@@ -850,6 +850,16 @@ tls_config:
 namespaces:
   names:
     [ - <string> ]
+
+# Optional label and field selectors to limit the discovery process to a subset of available resources. 
+# See https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/
+# and https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/ to know more about the possible
```
brian-brazil: s/know/learn/

discovery/kubernetes/kubernetes.go
shuttie: Nice catch, thanks!

docs/configuration/configuration.md
```diff
@@ -850,6 +850,16 @@ tls_config:
 namespaces:
   names:
     [ - <string> ]
+
+# Optional label and field selectors to limit the discovery process to a subset of available resources. 
+# See https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/
+# and https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/ to learn more about the possible 
+# filters that can be used. Endpoints role supports pod, service and endpoints selectors, other roles
+# only support selectors matching the role itself (e.g. node role can only contain node selectors).
```
brancz: I think this needs to document more the trade-off people are making here. Defining a label or field selector prevents Prometheus from re-using a single list/watch for all scrape configs (Prometheus already does this optimization today). Meaning that if there are two scrape configs each doing Prometheus discovery, then in reality Prometheus only opens 1 list/watch and processes the items through both relabeling configs, when this is used, the trade-off we're doing is we take the load off of Prometheus and put it onto the Kubernetes API, and create a list/watch per selector combination. It depends on the individual situation what may be the "best" approach. So I think this should have more thorough documentation about this trade-off.

brian-brazil: That's the case for other discovery mechanisms too, so a more general note up top could do it.

simonpasquier:
Given that we already have a good test suite in `discovery/kubernetes`, I'd suggest to add tests there too. We should also check the validity of the selectors using [`k8s.io/apimachinery/pkg/fields.ParseSelector`](https://godoc.org/k8s.io/apimachinery/pkg/fields#ParseSelector) when loading the configuration.

cyriltovena:
@shuttie Do you need help to move this forward ? I'd like to have this for Loki.

StevenYCChou:
@shuttie I'm interested in this feature as well, and I'd also like to offer help to move this forward.

x13n:
@shuttie Are you still working on this? Seems like there is some demand for that feature, so it would make sense for someone else to pick this up if you don't have the time.

simonpasquier:
If @shuttie doesn't have time to address the comments then I'm fine with having someone completing the work. Just make sure to credit the original contributor for the work (e.g. `Co-Authored-By` in the commit message).

olagacek:
I've picked up this PR and I am willing to complete this. My PR is https://github.com/prometheus/prometheus/pull/6807

roidelapluie:
Thank you @shuttie and @olagacek . This is now merged in master.

docs/configuration/configuration.md
roidelapluie: It has been addressed inside this part only. I am open to move that as a general note but that would deserve a dedicated discussion as now the documentations is basically 'flat', tls_config is at the same level as `kube_sd_config`, so that would probably take some refactoring.

Issue #5919, opened 2019-08-19
UI gets freezed on api/v1/label/__name__/values?_=1566195610386


kshitijcode:
I tried using the prometheus UI . 

On loading the endpoint, it got stalled executing the request 
api/v1/label/__name__/values?_=1566197849138. The response was 200OK but the browser is freezing. 

We do have a large number of metrics/labels in our cluster.

I saw a previous issue where limit was reduced to 10k but I guess thats not helping. Is there any way to configure this? 

Using prometheus version : 2.10.0

simonpasquier:
How many time series do you have?

FUSAKLA:
This issue was encountered using Thanos which uses the same GUI and theoretically can have unlimited number or label values even when following best practices. Not saying Prometheus UI should be able to handle amount of data this big but option to at least disable the hinting at all would be great possibly?

brian-brazil:
I thought we had something in the javascript to limit this?

FUSAKLA:
It was discussed on Thanos slack before https://cloud-native.slack.com/archives/CK5RSSC10/p1565939498149400

from the discussion:
> Yes, we do have large amount of metrics/labels. And yes the response is 200 OK. Can we disable or restrict /limit the values somehow?
The response is 200OK but it says request is not finished yet. The resource size download is about 125mb and request takes around 24seconds. 

The amount of data in this case is probably really big and JS can limit this only once it is loaded so in cases like this might be good to omit theese request at all? There is already the checkbox for enable/disable query history so the same way the hinting could be disabled maybe?

brian-brazil:
Uhm, how do you have several million metric names?

ramshankarjaiswal:
Current status of our prometheus - monitoring 2262 targets

sum((scrape_samples_scraped)) = 2172660
Output of ({__name__=~".+"}) -
Warning! Fetched 2177672 series, but displaying only first 10000.

Are the above two queries pretty much giving me the same answer? i.e. total current time series? and how do they differ from prometheus_tsdb_head_series < this gives us unique time series over last 2-3 hours?

In the past we've had some applications misbehave and expose 100k to 300k metrics each from about a 10+ servers .. that is why we have so many metric names.

This causes the browser to freeze after few seconds of running any query.. same thing happens when we use thanos for global query view.

Thanks to your book @brian-brazil I've come across https://www.robustperception.io/using-sample_limit-to-avoid-overload which I will now implement to avoid this in future.

[...]

PR #6100, opened 2019-10-04
Add a limit in the graph code for number of metrics

dgl:
Plus a few optimisations, this improves the situation for #5919.


GiedriusS:
Any updates on this? On the Thanos side, this kind of code makes sense where we easily have terabytes of data and solves the original problem mentioned here but I also think it's understandable if on the Prometheus' side you don't want to have such logic embedded in the UI's code.

brian-brazil:
@juliusv is this still relevant?

dgl:
Closing, would need updating for the new UI anyway (which mostly seems faster, although the fuzzy matching still seems expensive).

PR #6102, opened 2019-10-04
merge Release 2.13 branch into master

krasi-georgiev:
@bwplotka , @simonpasquier @brancz @brian-brazil @juliusv @SuperQ @codesome @csmarchbanks


.promu.yml
```diff
@@ -10,6 +10,8 @@ build:
           path: ./cmd/prometheus
         - name: promtool
           path: ./cmd/promtool
+        - name: tsdb
```
brian-brazil: Does this change make sense? @SuperQ

SuperQ: The change is fine, but I think it's already in master, so I'm not sure why there is a diff here.

krasi-georgiev: I was wondering as well, but it doesn't break anything so didn't spend much time to find out the cause.

krasi-georgiev:
superseded by https://github.com/prometheus/prometheus/pull/6117

PR #6103, opened 2019-10-04
promql: Move tests to testutil

dzeban:
Part of the fix for #3242 

### DIFF
--- a/promql/engine_test.go
+++ b/promql/engine_test.go
@@ -15,11 +15,11 @@ package promql
 
 import (
 	"context"
+	"errors"
 	"testing"
 	"time"
 
 	"github.com/go-kit/kit/log"
-	"github.com/pkg/errors"
 
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/storage"
@@ -104,9 +104,9 @@ func TestQueryTimeout(t *testing.T) {
 
 	res := query.Exec(ctx)
 	testutil.NotOk(t, res.Err, "expected timeout error but got none")
-	if _, ok := res.Err.(ErrQueryTimeout); res.Err != nil && !ok {
-		t.Fatalf("expected timeout error but got: %s", res.Err)
-	}
+
+	var e ErrQueryTimeout
+	testutil.Assert(t, errors.As(res.Err, &e), "expected timeout error but got: %s", res.Err)
 }
 
 const errQueryCanceled = ErrQueryCanceled("test statement execution")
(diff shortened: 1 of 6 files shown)


dzeban:
I think that the `errors.As` part is optional but, in my opinion, it reads easier then clunky `if` with type assertion.

promql/parse_test.go
```diff
@@ -1590,48 +1589,29 @@ func TestParseExpressions(t *testing.T) {
...
-		if test.fail && err != nil {
+		if !test.fail {
+			testutil.Ok(t, err)
+			testutil.Equals(t, expr, test.expected, "error on input '%s'", test.input)
+		} else {
+			testutil.NotOk(t, err)
 			if !strings.Contains(err.Error(), test.errMsg) {
 				t.Errorf("unexpected error on input '%s'", test.input)
 				t.Fatalf("expected error to contain %q but got %q", test.errMsg, err)
 			}
```
csmarchbanks: What would you/others think of adding a `testutil.Contains` function?

dzeban: As far as I understand, current mood about testutil is to keep it as small as possible. Moreover, I would remove this string comparison because it's essentially comparing errors and doing it via string comparison is fragile. I would create dedicated parser error types (maybe wrapped) and then matched those types in tests.

promql/engine_test.go
```diff
@@ -104,14 +103,14 @@ func TestQueryTimeout(t *testing.T) {
...
-	if res.Err == nil {
-		t.Fatalf("expected timeout error but got none")
-	}
-	if _, ok := res.Err.(ErrQueryTimeout); res.Err != nil && !ok {
-		t.Fatalf("expected timeout error but got: %s", res.Err)
-	}
+	testutil.NotOk(t, res.Err, "expected timeout error but got none")
+
+	var e ErrQueryTimeout
+	testutil.Assert(t, errors.As(res.Err, &e), "expected timeout error but got: %s", res.Err)
```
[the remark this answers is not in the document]
dzeban: As per [go wiki](https://github.com/golang/go/wiki/ErrorValueFAQ) `errors.Is` is to compare error values and `errors.As` is for checking error type. Here we look for error type and that's why it's `errors.As`.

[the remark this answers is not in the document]
csmarchbanks: :+1:

promql/parse_test.go
csmarchbanks: In this case string comparisons are fine because it is also testing that a reasonable error message is going to be displayed to the user. For this PR how about putting the contains function into a `testutil.Assert` to remove the two `t.` lines?

dzeban: OK, done.

PR #6108, opened 2019-10-07
fix: README Makefile target: build binaries

kormotodor:
According to https://github.com/prometheus/prometheus/pull/6089

Obvious fix.
Signed-off-by: Georgy hypno.cat40000@gmail.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ You can also clone the repository yourself and build using `make`:
 
 The Makefile provides several targets:
 
-  * *build*: build the `prometheus` and `promtool` binaries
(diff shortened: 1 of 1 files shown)


brian-brazil:
This is only a temporary state, we'll be merging tsdb into promtool.

kormotodor:
And how should we build binaries from source with Makefile when directory with name `tsdb` already exists?
I think it will ouputs something like this:

> go build github.com/prometheus/prometheus/tsdb/cmd/tsdb: build output .../src/github.com/prometheus/prometheus/tsdb" already exists and is a directory

Issue #6095, opened 2019-10-03
remote_write unreliable (stops writing) after prolonged connection issue to remote, during resharding


zytek:
## Bug Report
**What did you do?**

Set up remote_write to Victoria Metrics. Default queue config.
Prometheus is deployed via prometheus-operator on kubernetes.

**Expectations vs reality**

After connectivity issues to remote - either via unstable connection (AWS to AWS China) or because Remote had issues (EC2 failure, took more than 30 minutes to heal and be available again) Prometheus gives up. 

Once remote becomes available again Prometheus does not retry.

Logs suggest that resharding might be a culprit here - as resharding means flushing data from old shards which was impossible due to connection issues.

I consider this a reliability bug - as it causes me to restart Prometheus. 
Either retry logic or default configuration issue (need advice here) or resharding logic.

My expected behaviour would be:
1. Backoff increases to some high, but reasonable value (30s - 5 minutes)
2. Retry is indefinite
3. Each WAL rotation means I lose data, but this is acceptable 
4. Resharding is either cancelled or retried if buffers could not be flushed 

My workaround was to se min_shards to 1000, so resharding would not take place. I will perform more tests later to check if this actually helps. 

**Environment**

* System information:

Prometheus 2.10.0 deployed on Kubernetes 1.11 via prometheus-operator.

```
level=info ts=2019-10-02T14:41:22.597Z caller=main.go:322 msg="Starting Prometheus" version="(version=2.10.0, branch=HEAD, revision=d20e84d0fb64aff2f62a977adc8cfb656da4e286)"
level=info ts=2019-10-02T14:41:22.597Z caller=main.go:323 build_context="(go=go1.12.5, user=root@a49185acd9b0, date=20190525-12:28:13)"
level=info ts=2019-10-02T14:41:22.597Z caller=main.go:324 host_details="(Linux 4.4.0-1079-aws #89-Ubuntu SMP Tue Mar 26 15:25:52 UTC 2019 x86_64 prometheus-prometheus-operator-prometheus-0 (none))"
level=info ts=2019-10-02T14:41:22.597Z caller=main.go:325 fd_limits="(soft=1048576, hard=1048576)"
level=info ts=2019-10-02T14:41:22.597Z caller=main.go:326 vm_limits="(soft=unlimited, hard=unlimited)"
```

* Logs:
```
[...]

PR #6111, opened 2019-10-08
Refactor calculateDesiredShards + don't reshard if we're having issues sending samples.

cstyan:
This refactors calculateDesiredShards to return an int instead of sending on the `reshardChan` itself. It also re-introduces logic to skip resharding if we haven't successfully sent samples in `2*cfg.BatchSendDeadline`, which I believe was mistakenly removed in this commit: https://github.com/prometheus/prometheus/commit/67da8e7b460ffaba049334961b09569866c3e02e#diff-b3f7e4329d3b8d9712dd10086f0c27d9L414-L419 as part of refactoring metrics in remote write.

I'll think about any possible addition of tests overnight and add them in the morning.

@csmarchbanks I think the later might have been part of the issue in #6095 

cc @tomwilkie

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -186,6 +186,9 @@ type StorageClient interface {
 // indicated by the provided StorageClient. Implements writeTo interface
 // used by WAL Watcher.
 type QueueManager struct {
+	// https://golang.org/pkg/sync/atomic/#pkg-note-BUG
+	lastSendTimestamp int64
+
 	logger         log.Logger
 	flushDeadline  time.Duration
 	cfg            config.QueueConfig
@@ -207,8 +210,6 @@ type QueueManager struct {
 	samplesIn, samplesDropped, samplesOut, samplesOutDuration *ewmaRate
 	integralAccumulator                                       float64
 	startedAt                                                 time.Time
-	lastSendTimestampLock                                     sync.RWMutex
-	lastSendTimestamp                                         int64
 
 	highestSentTimestampMetric *maxGauge
 	pendingSamplesMetric       prometheus.Gauge
@@ -498,18 +499,6 @@ func (t *QueueManager) calculateDesiredShards() int {
 	t.samplesDropped.tick()
 	t.samplesOutDuration.tick()
 
-	// We shouldn't reshard if Prometheus hasn't been able to send to the
-	// remote endpoint successfully within some period of time.
-	now := time.Now().Unix()
-	threshold := int64(time.Duration(2 * t.cfg.BatchSendDeadline).Seconds())
-	t.lastSendTimestampLock.RLock()
-	diff := now - t.lastSendTimestamp
-	t.lastSendTimestampLock.RUnlock()
-	if diff > threshold {
-		level.Warn(t.logger).Log("msg", "Skipping resharding, last successful send was beyond threshold")
-		return t.numShards
-	}
-
 	// We use the number of incoming samples as a prediction of how much work we
 	// will need to do next iteration.  We add to this any pending samples
 	// (received - send) so we can catch up with any backlog. We use the average
@@ -541,6 +530,15 @@ func (t *QueueManager) calculateDesiredShards() int {
 		t.integralAccumulator = integralGain * samplesPending / float64(elapsed)
 	}
 	t.integralAccumulator += samplesPendingRate * integralGain
+
+	// We shouldn't reshard if Prometheus hasn't been able to send to the
+	// remote endpoint successfully within some period of time.
+	lts := atomic.LoadInt64(&t.lastSendTimestamp)
+	skip := (time.Now().Unix() - lts) > int64(2*time.Duration(t.cfg.BatchSendDeadline)/time.Second)
+	if skip {
+		level.Warn(t.logger).Log("msg", "Skipping resharding, last successful send was beyond threshold")
+		return t.numShards
+	}
 
 	var (
 		timePerSample = samplesOutDuration / samplesOutRate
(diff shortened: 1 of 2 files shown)


storage/remote/queue_manager.go
```diff
@@ -468,18 +470,46 @@ func (t *QueueManager) updateShardsLoop() {
...
 	t.samplesOutDuration.tick()
 
+	// We shouldn't reshard if Prometheus hasn't been able to send to the
+	// remote endpoint successfully within some period of time.
+	now := time.Now().Unix()
+	threshold := int64(time.Duration(2 * t.cfg.BatchSendDeadline).Seconds())
+	t.lastSendTimestampLock.RLock()
+	diff := now - t.lastSendTimestamp
+	t.lastSendTimestampLock.RUnlock()
+	if diff > threshold {
```
beorn7: With `time.Time` for `t.lastSendTimestamp` (then renamed `t.lastSend`), this could read:
```Go
t.lastSendLock.RLock()
skip := time.Since(t.lastSend) > 2 * t.cfg.BatchSendDeadline
t.lastSendLock.RUnlock()
if skip {
```

storage/remote/queue_manager.go
```diff
@@ -468,18 +470,46 @@ func (t *QueueManager) updateShardsLoop() {
...
+		level.Warn(t.logger).Log("msg", "Skipping resharding, last successful send was beyond threshold")
+		return t.numShards
```
csmarchbanks: By breaking out of the loop before trying to add to the integral accumulator you are not be accounting for tons of pending samples when Prometheus starts being able to send again. This will cause future shard calculations to under provision the number of necessary shards.

storage/remote/queue_manager.go
```diff
@@ -207,6 +207,8 @@ type QueueManager struct {
 	samplesIn, samplesDropped, samplesOut, samplesOutDuration *ewmaRate
 	integralAccumulator                                       float64
 	startedAt                                                 time.Time
+	lastSendTimestampLock                                     sync.RWMutex
+	lastSendTimestamp                                         int64
```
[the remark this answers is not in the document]
csmarchbanks: Note that if going the atomic route, the variable should be moved to the top of the struct to avoid breaking some builds: https://golang.org/pkg/sync/atomic/#pkg-note-BUG

cstyan:
@beorn7 @csmarchbanks thanks for the reviews, I went with the atomic int and moved the skip check to after the integral accumulator block

storage/remote/queue_manager.go
```diff
@@ -512,6 +531,15 @@ func (t *QueueManager) calculateDesiredShards() {
 	}
 	t.integralAccumulator += samplesPendingRate * integralGain
 
+	// We shouldn't reshard if Prometheus hasn't been able to send to the
+	// remote endpoint successfully within some period of time.
+	lts := atomic.LoadInt64(&t.lastSendTimestamp)
+	skip := (time.Now().Unix() - lts) > int64(2*time.Duration(t.cfg.BatchSendDeadline)/time.Second)
+	if skip {
```
beorn7: Just a nit about making this more readable:

I only used the `skip` bool in my suggestion because it needed to be pulled out of the code path protected by the mutex. And the whole condition in one line only read nicely because it used `time.Since(lastSend)` and such.

Now that we don't need a mutex and don't use `time.Time` for `lastSendTimestamp`, how about the following:

```Go
minSendTimestamp := time.Now().Add(-2*time.Duration(t.cfg.BatchSendDeadline)).Unix()
if atomic.LoadInt64(&t.lastSendTimestamp) < minSendTimestamp {
```

storage/remote/queue_manager_test.go
```diff
@@ -281,6 +281,49 @@ func TestReleaseNoninternedString(t *testing.T) {
...
+			m.lastSendTimestamp = m.lastSendTimestamp - int64(3*time.Duration(config.DefaultQueueConfig.BatchSendDeadline)/time.Second)
+		}
+		m.Start()
+		desiredShards := m.calculateDesiredShards()
+		m.Stop()
+		if !c.reshard {
+			testutil.Assert(t, desiredShards == m.numShards, "expected calculateDesiredShards to not want to reshard, wants to change from %d to %d shards", m.numShards, desiredShards)
+		} else {
+			testutil.Assert(t, desiredShards != m.numShards, "expected calculateDesiredShards to want to reshard, wants to change from %d to %d shards", m.numShards, desiredShards)
+
```
csmarchbanks: Nit: extra empty line.

storage/remote/queue_manager_test.go
```diff
@@ -281,6 +281,49 @@ func TestReleaseNoninternedString(t *testing.T) {
...
+		{
+			startingShards: 5,
+			reshard:        true,
+		},
+	}
+	for _, c := range cases {
+		client := NewTestStorageClient()
+		m := NewQueueManager(nil, nil, "", newEWMARate(ewmaWeight, shardUpdateDuration), config.DefaultQueueConfig, nil, nil, client, defaultFlushDeadline)
+		m.numShards = c.startingShards
+		// Fake the samples in/out rates and the last sent timestamp so the reshard calculation has some numbers to use.
```
csmarchbanks: This comment isn't necessary.

storage/remote/queue_manager_test.go
```diff
@@ -281,6 +281,49 @@ func TestReleaseNoninternedString(t *testing.T) {
...
+		m.samplesIn.incr(1000)
```
csmarchbanks: Consider including samplesIn and samplesOut in the testcases. Right now it would be hard to write new test cases with these hard coded values.

cstyan:
@beorn7 @csmarchbanks I think this is good for another review.

I left an assignment of the `atomic.LoadInt` call to var in the block that checks if we should reshard so that the value could be used when logging without having to load again.

PR #6119, opened 2019-10-09
Add additional shards/segment graphs to remote write dashboard.

cstyan:
2.13.0 added new metrics for min/max/desired shards as well as the TSDB current segment and WAL watcher current segment. 

This PR adds panels to the remote write dashboard in the mixin so that you can at a glance check if desired shards is > min or > max shards for any queue, and how far behind (in segments) a given queue may be.

cc @csmarchbanks 

### DIFF
--- a/documentation/prometheus-mixin/dashboards.libsonnet
+++ b/documentation/prometheus-mixin/dashboards.libsonnet
@@ -1,5 +1,12 @@
 local g = import 'grafana-builder/grafana.libsonnet';
-
+local grafana = import 'grafonnet/grafana.libsonnet';
+local dashboard = grafana.dashboard;
+local row = grafana.row;
+local singlestat = grafana.singlestat;
+local prometheus = grafana.prometheus;
+local graphPanel = grafana.graphPanel;
+local tablePanel = grafana.tablePanel;
+local template = grafana.template;
 {
   grafanaDashboards+:: {
     'prometheus.json':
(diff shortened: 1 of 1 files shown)


cstyan:
waiting to hear from grafana folks about any scale limits for the first panel re: max_shards

cstyan:
@beorn7 PTAL; I rewrote the dashboard using grafonnet and I've made the two separate panels you suggested. For now I've only made the first panel repeat on the `cluster` label. 

I'm not sure how I feel about the repeat feature, we're always refining the dashboard when we view it via the template variables. I feel like I'd rather not have the repeat and now have to scroll as much when all the template variables are set to `All`, but I think we should proceed as is and see how we feel when using the dashboard and if we get feedback from anyone else.

beorn7:
Repeating on everything will blow up things way too much, indeed. But I would prefer to have `queue` as the one variable to repeat on. If I want to analyze queue sharding and capacity, I'm most likely looking at a specific Prometheus instance, for which I want to see the queues next to each other. Also, queue numbers are kind of limited by nature. (You won't have dozens or hundreds of queues on a single Prometheus server, and in particular the number of queues doesn't really grow with the size of your org.) While the number of clusters and number of Prometheus instances grows with your org size and can thus blow up the dashboard size easily if on repeat.

I generally agree we should let this bake and see how useful it is before discussing further details. But then again, I think repeating on `queue` is the more conservative choice here as this might be used by orgs with dozens of clusters and hundreds of Prometheus instances.

PR #6122, opened 2019-10-10
Clarify the branch in preparing the release

codesome:


RELEASE.md
```diff
@@ -56,7 +56,7 @@ For a patch release, work in the branch of the minor release you want to patch.
 
 For a new major or minor release, create the corresponding release branch based on the master branch.
 
-Bump the version in the `VERSION` file and update `CHANGELOG.md`. Do this in a proper PR as this gives others the opportunity to chime in on the release in general and on the addition to the changelog in particular.
+Bump the version in the `VERSION` file and update `CHANGELOG.md`. Do this in a proper PR pointing to the release branch as this gives others the opportunity to chime in on the release in general and on the addition to the changelog in particular.
```
codesome: Else it is unclear if the `CHANGELOG` and `VERSION` have to be put into master first and then create a branch, or start all the changes from the release branch and the changes to those files goes into master only after the release.

beorn7: There is the scenario where master already moves on while the `VERSION`/`CHANGELOG.md` PR is still being discussed. Branching off first allows the release shepherd to explicitly pick the point in the commit history that they wish to release.

While both ways are valid, I'd prefer we have the one with the least possible complications in the instructions. So I'm :+1: for this change.

bwplotka: Happy with this restriction as well.

PR #6123, opened 2019-10-10
Correct spelling mistakes

yuxiaobo96:

### DIFF
--- a/scrape/scrape_test.go
+++ b/scrape/scrape_test.go
@@ -550,12 +550,12 @@ func TestScrapeLoopRun(t *testing.T) {
 	select {
 	case <-signal:
 	case <-time.After(5 * time.Second):
-		t.Fatalf("Cancelation during initial offset failed")
+		t.Fatalf("Cancellation during initial offset failed")
 	case err := <-errc:
 		t.Fatalf("Unexpected error: %s", err)
 	}
 
-	// The provided timeout must cause cancelation of the context passed down to the
+	// The provided timeout must cause cancellation of the context passed down to the
 	// scraper. The scraper has to respect the context.
 	scraper.offsetDur = 0
 
@@ -607,7 +607,7 @@ func TestScrapeLoopRun(t *testing.T) {
 	case err := <-errc:
 		t.Fatalf("Unexpected error: %s", err)
 	case <-time.After(3 * time.Second):
-		t.Fatalf("Loop did not terminate on context cancelation")
+		t.Fatalf("Loop did not terminate on context cancellation")
 	}
 }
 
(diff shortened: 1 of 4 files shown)


scrape/scrape_test.go
```diff
@@ -550,12 +550,12 @@ func TestScrapeLoopRun(t *testing.T) {
 	select {
 	case <-signal:
 	case <-time.After(5 * time.Second):
-		t.Fatalf("Cancelation during initial offset failed")
+		t.Fatalf("Cancellation during initial offset failed")
```
codesome: A quick google search tells me that `cancelation` is also fine.

tsdb/chunkenc/xor.go
```diff
@@ -14,7 +14,7 @@
 // The code in this file was largely written by Damian Gryski as part of
 // https://github.com/dgryski/go-tsz and published under the license below.
 // It was modified to accommodate reading from byte slices without modifying
-// the underlying bytes, which would panic when reading from mmaped
+// the underlying bytes, which would panic when reading from mapped
```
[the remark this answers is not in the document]
yuxiaobo96: OK.

scrape/scrape_test.go
brian-brazil: That's Americans forgetting their own spellings, it's meant to be two l's.

tsdb/compact_test.go
```diff
@@ -163,7 +163,7 @@ func TestNoPanicFor0Tombstones(t *testing.T) {
 }
 
 func TestLeveledCompactor_plan(t *testing.T) {
-	// This mimicks our default ExponentialBlockRanges with min block size equals to 20.
+	// This mimics our default ExponentialBlockRanges with min block size equals to 20.
```
brian-brazil: both spellings are correct

yuxiaobo96: Thanks.

tsdb/chunkenc/xor.go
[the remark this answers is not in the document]
beorn7: How about "mmap'd"? It would make it more obvious that we are conjugating a technical term here (which isn't really a verb).

scrape/scrape_test.go
beorn7: Oh that debate again… :o) One day, we might need a formal vote of prometheus-team to solve it once and forever.

Personally, my proposal would be to ban flip-flopping between the two spellings in PRs so that we are not lured into wasting time on it ever again. Both spellings exist, and we have other problems than deciding for the world which one to use. There isn't even any ambiguity involved (like with Oxford comma or "Prometheus'" vs. "Prometheus's").

For the interested: https://en.wiktionary.org/wiki/cancelation probably summarizes it neatly.

tsdb/chunkenc/xor.go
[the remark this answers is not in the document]
yuxiaobo96: I think this way is better.
I will update it.

scrape/scrape_test.go
brian-brazil: Ah, I'd not seen that it was one of those explicit US changes.

PR #6125, opened 2019-10-10
Head Cardinality Status Page

sharadgaur:
The goal of this pull request to analyze TSDB Head statistics and display on Status page. 
Please feel free to provide comments or suggestion.


brian-brazil:
How heavy is this? Anything on the status page needs to be pretty light, which is why tsdb analyze is offline.

tsdb/index/postings.go
```diff
@@ -79,6 +79,70 @@ func (p *MemPostings) SortedKeys() []labels.Label {
...
+	var size uint64
+	p.mtx.RLock()
+
+	for n, e := range p.m {
+		if n == "" {
+			continue
+		}
+		labels = append(labels, HeadStats{Name: n, Count: uint64(len(e))})
+		size = 0
+		if n == "__name__" {
```
brian-brazil: The TSDB shouldn't know that __name__ is special.

sharadgaur:
@brian-brazil thanks for suggestion. 
By default status page is not sending any Head Stats query until user clicks on "Query Cardinality Status" button.
image

Currently we are running load test in our dev environment. I will update you soon on statistics like page load time and query time.

tsdb/index/postings.go
sharadgaur: Removed the ` __name__ ` from TSDB.

brian-brazil:
To be clear, anything accessible from the UI needs to be light. Especially if taking mutexes on the critical path.

tsdb/index/postings.go
```diff
@@ -79,6 +79,70 @@ func (p *MemPostings) SortedKeys() []labels.Label {
...
+	LabelValueStats         []HeadStats
+	LabelValuePairsStats    []HeadStats
+}
+
+// Calculating Cardinality Stats
+func (p *MemPostings) CardinalityStats(label string) *Stats {
+	metrics := make([]HeadStats, 0)
+	labels := make([]HeadStats, 0)
+	labelValueLenght := make([]HeadStats, 0)
+	labelValuePairs := make([]HeadStats, 0)
```
csmarchbanks: These either need to be initialized more intelligently or replaced with a heap/slice of 10 entries that are continuously updated.

docs/configuration/configuration.md
```diff
@@ -20,6 +20,8 @@ A configuration reload is triggered by sending a `SIGHUP` to the Prometheus proc
 sending a HTTP POST request to the `/-/reload` endpoint (when the `--web.enable-lifecycle` flag is enabled).
 This will also reload any configured rule files.
 
+To enable TSDB head statistics (cardinality stats) on Prometheus web UI (Status>Runtime & Build Information) set `--web.enable-cardinality-stats=true`
```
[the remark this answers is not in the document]
sharadgaur: how about `--web.disable-cardinality-stats`?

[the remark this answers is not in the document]
brian-brazil: This sort of thing should be always on, if it's a risk we shouldn't have it in the first place.

sharadgaur:
Benchmark for postings stats functions its taking less than ~12 mill seconds with 6 MB of memory for 200K+ unique label value pairs.
Head block with 4+ millions series was calculated in less than 70 milliseconds.

sharadgaur:
@brian-brazil posted the benchmark tests results above. It look very lightweight.

tsdb/head.go
```diff
@@ -231,6 +235,26 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
...
 }
 
+const cardinalityCacheExperationTime int64 = 30 //seconds
+
+//Return top 10 Highest Cardinality stats for Labels and Value Names
+func (h *Head) PostingsCardinalityStats(statsByLabelName string) *index.PostingsStats {
+	h.cardinalityMutex.Lock()
+	defer h.cardinalityMutex.Unlock()
+
+	seconds := time.Now().Unix() - h.lastPostingsStatsCall
```
csmarchbanks: It would be nice to have all of this logic using `time.Time` and `time.Duration` rather than various `int64`s.

tsdb/index/postingsstats.go
```diff
@@ -0,0 +1,72 @@
...
+	"math"
+	"sort"
+)
+
+type Stat struct {
+	Name  string
+	Count uint64
+}
+
+type MaxHeap struct {
```
csmarchbanks: I don't think this struct or methods should be exposed. It is pretty specific to only the postings stats use case.

tsdb/index/postingsstats.go
```diff
@@ -0,0 +1,72 @@
...
+	Items     []Stat
+}
+
+func (m *MaxHeap) Init(len int) {
+	m.maxLength = len
+	m.minValue = math.MaxUint64
+	m.Items = make([]Stat, 0, len)
+}
+
+func (m *MaxHeap) Len() int {
```
csmarchbanks: Is this function used anywhere?

web/ui/templates/status.html
```diff
@@ -101,5 +95,26 @@ <h2 id="alertmanagers">Alertmanagers</h2>
...
+        </tr>
+        <tr>
+          <td scope="row">{{ .NumSeries}}</th>
+          <td>{{.ChunkCount}}</td>  
+          <td>{{ .MaxTime}}</td>
+          <td>{{ .MinTime}}</td>
+        </tr>
+      </tbody>
+    </table>
+    <input class="btn btn-primary" type="submit" value="Query Cardinality Stats" id="get_stats"></div>
```
csmarchbanks: Is the button necessary here or should the query just be done on page load? Since the query is cached it won't update if pressed frequently anyway.

tsdb/head.go
```diff
@@ -231,6 +235,26 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 	return m
 }
 
+const cardinalityCacheExperationTime int64 = 30 //seconds
```
csmarchbanks: s/Experation/Expiration

Since this is only used in one place it could be inlined if you wanted.

tsdb/head.go
```diff
@@ -231,6 +235,26 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
...
+
+//Return top 10 Highest Cardinality stats for Labels and Value Names
```
csmarchbanks: Comments should have a space after the `//` and end in a full stop (needs to be updated in a few places).

tsdb/head.go
bwplotka: Also if the unit is not clear it should be embedded in variable name OR `time.Duration` should be used. Comment can be lost easily. I would vote for `time.Duration`, thoughts?

web/ui/static/js/status.js
```diff
@@ -0,0 +1,20 @@
+
+function init() {
+  
+
+  $("#get_stats").click(function() {
+    // $("#get_stats").attr("disabled", true);
```
bwplotka: Some commented code ):

web/web.go
```diff
@@ -635,7 +637,51 @@ func (h *Handler) graph(w http.ResponseWriter, r *http.Request) {
...
+	stats := db.Head().PostingsCardinalityStats("__name__")
+	data := struct {
+		Stats    *index.PostingsStats
+		Duration string
+	}{
+		Stats:    stats,
+		Duration: fmt.Sprintf("%.3f", float64(time.Now().UnixNano()-startTime)/float64(1e9)),
+	}
+
+	f, err := ui.Assets.Open(path.Join("/templates", "headstats.html"))
```
bwplotka: Can't we follow `executeTemplate` method?

sharadgaur:
@bwplotka  Here are the banchmark results:

goos: darwin
goarch: amd64
pkg: github.com/prometheus/prometheus/tsdb/index
BenchmarkPostings_Stats-12    	      81	  14950851 ns/op	 6903856 B/op	  200816 allocs/op
PASS
ok  	github.com/prometheus/prometheus/tsdb/index	25.356s
Success: Benchmarks passed.

krasi-georgiev:
instead of this isn't it possible to keep some metrics/state when adding and removing postings?

krasi-georgiev:
These bench results assume that no other process is holding the mutex lock so don't think are an actual presentation of what to expect in real installations.

Anyway did you try instead of this to keep some metrics/state when adding and removing postings?
 


> @bwplotka Here are the banchmark results:
> 
> goos: darwin
> goarch: amd64
> pkg: github.com/prometheus/prometheus/tsdb/index
> BenchmarkPostings_Stats-12 81	14950851 ns/op	6903856 B/op	200816 allocs/op
> PASS
> ok github.com/prometheus/prometheus/tsdb/index	25.356s
> Success: Benchmarks passed.

sharadgaur:
@krasi-georgiev thank for review.
Initially, I thought of doing the same but the only reason I hesitated, it will take dedicated resources all the time if we start collecting stats at the time of postings add/remove operations.
Current path is more like on-demand collection of stats.

csmarchbanks:
/benchmark

Let's run a benchmark for a bit and see how long it takes to load with that data. Since it is a read lock I would imagine it would not be much different than the latency added to any query for when the write lock is taken. With the benchmark we mostly wanted to see how long it would stop a write lock from being taken.

csmarchbanks:
/benchmark cancel

Will run benchmark again once https://github.com/prometheus/prombench/issues/299 is fixed.

csmarchbanks:
/benchmark cancel

On the test data loading the cardinality stats was taking sub 1ms. likely because there were not very many distinct label names so the postings. Our internal tests on a reasonably abused instance showed up to 70ms. No significant memory usage in benchmark or our internal tests.

How lightweight does this need to be to be merged/what else could we do to test it?

sharadgaur:
@krasi-georgiev any suggestion? It looks lightweight.

tsdb/index/postings.go
```diff
@@ -79,6 +79,56 @@ func (p *MemPostings) SortedKeys() []labels.Label {
...
 }
 
+type PostingsStats struct {
+	CardinalityMetricsStats []Stat
+	CardinalityLabelStats   []Stat
+	LabelValueStats         []Stat
+	LabelValuePairsStats    []Stat
+}
+
+// Calculating Cardinality Stats
```
csmarchbanks: Format of this comment should be something like `// Stats calculates the cardinality statistics for postings.`

tsdb/index/postingsstats.go
csmarchbanks: `MaxHeap` -> `maxHeap` still needs to be changed.

tsdb/index/postingsstats.go
```diff
@@ -0,0 +1,68 @@
+// Copyright 2017 The Prometheus Authors
```
csmarchbanks: Date should be 2019.

tsdb/index/postingsstats_test.go
```diff
@@ -0,0 +1,77 @@
+// Copyright 2017 The Prometheus Authors
```
csmarchbanks: Date should be 2019 here as well.

tsdb/head.go
```diff
@@ -231,6 +235,26 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 	return m
 }
 
+const cardinalityCacheExpirationTime = time.Duration(30) * time.Second
+
+// Returns top 10 Highest Cardinality stats By label and value names.
```
csmarchbanks: Comment format should be like `// PostingsCardinalityStats returns the top 10 highest cardinality stats by label and value names.`

tsdb/index/postings.go
```diff
@@ -79,6 +79,56 @@ func (p *MemPostings) SortedKeys() []labels.Label {
 	return keys
 }
 
+type PostingsStats struct {
```
csmarchbanks: Exported type should have a godoc comment. Perhaps `// PostingsStats contains cardinality based statistics for postings.`

tsdb/index/postingsstats.go
```diff
@@ -0,0 +1,68 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package index
+
+import (
+	"math"
+	"sort"
+)
+
+type Stat struct {
```
csmarchbanks: Exported type should have a godoc comment. Perhaps `// Stat holds values for a single cardinality statistic.`

sharadgaur:
thank you @csmarchbanks  and @bwplotka

juliusv:
For this much new information it might make sense to introduce a new TSDB stats page. But let's do that in the new UI...

sharadgaur:
+1 for the new TSDB stats page in React.

PR #6128, opened 2019-10-11
[ENHANCEMENT] service discovery: Fewer configuration hash shard service discovery

liguozhong:
##background：
   When we have 50 jobs, when using relabel+hashmod to do hash scrape, the prometheus.yaml configuration file becomes basically unmaintainable. So we want hash scrape to be simpler and less likely to make mistakes.

##solution：
old config：https://www.robustperception.io/scaling-and-federating-prometheus
image
new config：
image

##detail
Relabel is a very powerful ETL feature, but when used as a hash scrape, it is difficult to operate, easy to make mistakes, and configured too long.

Our users need to understand the full lifecycle of a label (service discovery -> relabel -> scrape -> metric relabel. At the same time, scrape uses the label __addreass__ to fetch data, so there must be this label). Understand the configuration behind this relabel+hashmod configuration. And most of our customer scenarios can be done without relabel. So most of our users don't understand the configuration of hash scrape.

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -443,6 +443,9 @@ func main() {
 				}
 				c[fmt.Sprintf("%x", md5.Sum(b))] = v.ServiceDiscoveryConfig
 			}
+
+			discoveryManagerScrape.Shards = cfg.GlobalConfig.Shards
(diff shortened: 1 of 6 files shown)


brian-brazil:
We looked at this as part of the bug scrub, and are closing per the above comments plus that this would only work in situations when every single scrape config requires sharding.

PR #6129, opened 2019-10-11
fix crash when out of sequence wal segments

### LINKED ISSUE #6080: Remote_write unable to send due to `segments are not sequential` error
## Bug Report
**What did you do?**
Restarted prometheus after a crash

Prometheus start and continue to send metrics to remote_write endpoint

Prometheus starts and scrapes metrics, but the logs are littered with:
```
prometheus[19563]: level=error ts=2019-10-01T02:43:23.492Z caller=wal_watcher.go:176 component=remote queue=1:<REMOTE_WRITE_ENDPOINT> msg="error tailing WAL" err="wal.Segments: segments are not sequential"
```

When I look at the wal directory I can see that I have `00000007.repair` in the wal directory.  This means in the [listSegments](https://github.com/prometheus/prometheus/blob/master/tsdb/wal/wal.go#L762) call it is skipped, meaning it reads 6 then 8 and then complains about this error. I then attempted to restart prometheus (thinking it would re-write the WAL or finish the repair) -- but unfortunately it now won't start -- with the same error (`segments are not sequential`). So this means that now I can't do anything with prometheus until I manually clear the WAL file -- which seems not-so-good :)

So with that in mind, it seems that we should probably add some option (env var, cli flag) to allow for partial WAL files -- otherwise if *any* file is corrupt the only way to get prometheus to start is to delete them all or manually rename all the files.

**Environment**

* Prometheus version: 2.12.0

Harkishen-Singh:

Fixes #6080 

On encountering any unsequential wal segments _(gap in the sequence_), the wal segments after the gap are ignored and deleted. This prevents Prometheus from stopping when encountering such situations.


Harkishen-Singh:
@brian-brazil please review.

brian-brazil:
@krasi-georgiev @codesome A TSDB change for you.

Harkishen-Singh:
@krasi-georgiev @codesome please review.

krasi-georgiev:
@Harkishen-Singh thanks for picking up this.

I think `listSegments` should return an error like `ErrorUnsequential` and than the repair should happen in `(w *WAL) Repair`

also we would need to add a test for `(w *WAL) Repair` to ensure this behavior.
 - create some WAL files
 - remove a file in the middle
 - run `tsdb.Open` to ensure the wal is repaired.

Harkishen-Singh:
@krasi-georgiev sure. Will implement the same and inform you accordingly.

Harkishen-Singh:
@krasi-georgiev implemented the suggestions. Please have a look.

Harkishen-Singh:
@krasi-georgiev sure!

Harkishen-Singh:
@krasi-georgiev pls have a look and suggest improvements if required.

krasi-georgiev:
Please try if you can just propagate the error to 
```
if initErr := db.head.Init(minValidTime); initErr != nil {
....

and handle the error in the  repair just after that block

**Comment by Harkishen-Singh** [comment]
@krasi-georgiev sure!

**Comment by Harkishen-Singh** [comment]
@krasi-georgiev implemented the suggestions. Please have a look.

File: tsdb/db.go [hunk 1]
```diff
@@ -538,6 +538,15 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db
 			segmentSize = opts.WALSegmentSize
 		}
 		wlog, err = wal.NewSize(l, r, filepath.Join(dir, "wal"), segmentSize, opts.WALCompression)
+		if errors.Cause(err).Error() == wal.ErrorUnsequentialSegments {
+			if err = wlog.Repair(errors.Cause(err)); err != nil {
```
krasi-georgiev: should increase the metric and also show a log for this repair, similiar to the other repair block, and for this the wal repair metric needs to be moved in the db struct
```
db.metrics.walCorruptionsTotal.Inc()
level.Warn(db.logger).Log("msg", "encountered WAL segment sequence error, attempting repair", "err", err)
```

tsdb/wal/wal.go
```diff
@@ -335,6 +342,11 @@ func (w *WAL) Repair(origErr error) error {
...
 	// Maybe as an extra mode in the future if mid-WAL corruptions become
 	// a frequent concern.
+
+	if origErr.Error() == ErrorUnsequentialSegments {
+		return w.repairUnsequentialSegments()
+	}
+
 	err := errors.Cause(origErr) // So that we can pick up errors even if wrapped.
 
 	cerr, ok := err.(*CorruptionErr)
```
krasi-georgiev: now that the Repair handles different errors I would move everything after this in its own func

tsdb/wal/wal.go
```diff
@@ -757,23 +769,52 @@ func listSegments(dir string) (refs []segmentRef, err error) {
...
-			return nil, errors.New("segments are not sequential")
+
+		if len(refs) > 0 && k > last+1 && inSequence {
+			inSequence = false
 		}
 		refs = append(refs, segmentRef{name: fn, index: k})
 		last = k
 	}
+	if !inSequence {
+		return refs, errors.New(ErrorUnsequentialSegments)
```
krasi-georgiev: It is quite unusual that `refs` is also returned with the error
I would just return nil here and the repair can read the segment numbers and do the repair complitely separately. It would duplicate some code, but would be easier to follow.

tsdb/wal/wal.go
```diff
@@ -757,23 +769,52 @@ func listSegments(dir string) (refs []segmentRef, err error) {
...
 	sort.Slice(refs, func(i, j int) bool {
 		return refs[i].index < refs[j].index
 	})
 	return refs, nil
 }
 
+// Returns the sequential segments after deleting the out of sequence wals after the sequence gap.
+func (w *WAL) repairUnsequentialSegments() error {
+	segments, _ := listSegments(w.dir)
+	level.Warn(w.logger).Log("msg", "deleting unsequential segments")
```
krasi-georgiev: I would also add the last good segment index and the segments that would be deleted. which means moving this log a bit further down.

tsdb/wal/wal.go
```diff
@@ -757,23 +769,52 @@ func listSegments(dir string) (refs []segmentRef, err error) {
...
+	last, outofSequence := -1, false
+
+	for _, seg := range segments {
+		k, err := strconv.Atoi(seg.name)
+		if err != nil {
+			return errors.Wrap(err, "faulty segment name")
```
krasi-georgiev:
```suggestion
			return errors.Wrap(err, "invalid segment name")
```

tsdb/wal/wal.go
```diff
@@ -757,23 +769,52 @@ func listSegments(dir string) (refs []segmentRef, err error) {
...
 	for _, fn := range files {
 		k, err := strconv.Atoi(fn)
 		if err != nil {
 			continue
 		}
-		if len(refs) > 0 && k > last+1 {
-			return nil, errors.New("segments are not sequential")
+
+		if len(refs) > 0 && k > last+1 && inSequence {
+			inSequence = false
```
krasi-georgiev:
```suggestion
			return nil, errors.New(ErrorUnsequentialSegments)```

File: tsdb/wal/wal.go [hunk 7]
```diff
@@ -757,23 +769,52 @@ func listSegments(dir string) (refs []segmentRef, err error) {
...
+	segments, _ := listSegments(w.dir)
+	level.Warn(w.logger).Log("msg", "deleting unsequential segments")
+	last, outofSequence := -1, false
+
+	for _, seg := range segments {
+		k, err := strconv.Atoi(seg.name)
+		if err != nil {
+			return errors.Wrap(err, "faulty segment name")
+		}
+		if k > last+1 || outofSequence {
```
krasi-georgiev: When the segments are ordered by name you can just simply delete all the segments after the last good segment.
Something like:

```
if k > last+1 {
	log which is the last good segment and the segments which will be deleted.
	_,seg:=range segments[k:]{
		err := os.Remove(SegmentName(w.dir, seg.index))
		if err != nil {
			return errors.Wrapf(err, "delete segment: %s", seg.name)
		}
	}
	return nil	
}

```

tsdb/wal/wal_test.go
```diff
@@ -204,6 +208,21 @@ func TestWALRepair_ReadingError(t *testing.T) {
 	}
 }
 
+func TestUnsequentialSegments(t *testing.T) {
+	segsN := 5 * pageSize
```
krasi-georgiev: this determines how big is the wal and not how many wal files it will create.

tsdb/wal/wal_test.go
```diff
@@ -204,6 +208,20 @@ func TestWALRepair_ReadingError(t *testing.T) {
...
+	dir := "../testdata/repair_unsequential_wals"
+	w, err := NewSize(log.NewNopLogger(), nil, dir, segsN, false)
+
+	// send unsequential segments error to trigger the segment scan
+	w.Repair(errors.Cause(err))
+	f, err := exec.Command("ls", "../testdata/repair_unsequential_wals").Output()
```
krasi-georgiev: use ioutil.ReadDir

krasi-georgiev:
also please make sure all tests are passing, Thanks

Harkishen-Singh:
@krasi-georgiev sure!

tsdb/wal/wal.go
```diff
@@ -258,13 +264,14 @@ func NewSize(logger log.Logger, reg prometheus.Registerer, dir string, segmentSi
 		stopc:       make(chan chan struct{}),
 		compress:    compress,
 	}
-	w.metrics = newWALMetrics(w, reg)
 
 	_, last, err := w.Segments()
 	if err != nil {
-		return nil, errors.Wrap(err, "get segment range")
+		return w, errors.Wrap(err, "get segment range")
```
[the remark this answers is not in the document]
Harkishen-Singh: we need a directory path in order to repair/delete the corrupted segments which is done inside `repairUnsequentialSegments()`. If I return nil, then I might not be able to pass the directory since the `Repair()` calls the `repairUnsequentialSegments()`. Hence I use the interface to pass on the path of the directory.

tsdb/wal/wal.go
```diff
@@ -258,13 +264,14 @@ func NewSize(logger log.Logger, reg prometheus.Registerer, dir string, segmentSi
...
 	}
 
+	w.metrics = newWALMetrics(w, reg)
```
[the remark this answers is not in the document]
Harkishen-Singh: This is moved in order to prevent the `MustRegister()` else, it will throw an error `dulpicate wals`

tsdb/db.go
Harkishen-Singh: sure

tsdb/wal/wal.go
Harkishen-Singh: sure

Harkishen-Singh:
@krasi-georgiev implemented the suggestions. Please have a look.

I have also fixed the tests. However, one test keeps failing which I guess, checks the wals added for unsequential tests. They were not in the original repo, so that may be the reason.
Please help me fix that test. Thank you!

tsdb/db.go
Harkishen-Singh: this is giving me `panic: runtime error: invalid memory address or nil pointer dereference`.

tsdb/wal/wal.go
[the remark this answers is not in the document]
Harkishen-Singh: is the explanation satisfactory?

tsdb/wal/wal_test.go
Harkishen-Singh: This is done in order to prevent `error: invalid segment` which is due to `segsN%pageSize != 0`.

Harkishen-Singh:
@krasi-georgiev implemented the suggestions. Please have a look.

tsdb/db.go
```diff
@@ -538,6 +538,17 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db
 			segmentSize = opts.WALSegmentSize
 		}
 		wlog, err = wal.NewSize(l, r, filepath.Join(dir, "wal"), segmentSize, opts.WALCompression)
+		if err != nil && errors.Cause(err).Error() == wal.ErrorUnsequentialSegments {
+			level.Warn(db.logger).Log("msg", "encountered WAL segment sequence error, attempting repair", "err", err)
+			level.Warn(db.logger).Log("msg", "deleting unsequential segments")
+			if err = wlog.Repair(errors.Cause(err), filepath.Join(dir, "wal")); err != nil {
```
krasi-georgiev: instead of calling `wlog.Repair` can call `wal.RemoveUnsequentialSegments(dir)`

tsdb/wal/wal.go
```diff
@@ -774,6 +784,52 @@ func listSegments(dir string) (refs []segmentRef, err error) {
...
+		if err != nil {
+			return nil, err
+		}
+		refs = append(refs, segmentRef{name: fn, index: k})
+	}
+	sort.Slice(refs, func(i, j int) bool {
+		return refs[i].index < refs[j].index
+	})
+	return refs, nil
+}
```
codesome: Can we avoid duplication of code? `listSegments` could call `getSegments` and check for the sequence.

tsdb/wal/wal.go
```diff
@@ -774,6 +784,52 @@ func listSegments(dir string) (refs []segmentRef, err error) {
...
+
+// RepairUnsequentialSegments repairs the out of sequence segments by deleting the onces after the sequence gap.
+// Returns the sequential segments after deleting the out of sequence wals after the sequence gap.
```
codesome: I got a couple of questions here for maintainers:
1. Should we just recover data from all segments without deleting any of them?
2. Why delete the segments after the gap and not before the gap?

/cc @krasi-georgiev @brian-brazil

tsdb/wal/wal.go
```diff
@@ -258,13 +264,15 @@ func NewSize(logger log.Logger, reg prometheus.Registerer, dir string, segmentSi
 		stopc:       make(chan chan struct{}),
 		compress:    compress,
 	}
-	w.metrics = newWALMetrics(w, reg)
 
+	// Prevent MustRegister if segments corrupted
```
codesome: No need for this comment here. The tests should capture this.

tsdb/db.go
```diff
@@ -538,6 +538,17 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db
 			segmentSize = opts.WALSegmentSize
 		}
 		wlog, err = wal.NewSize(l, r, filepath.Join(dir, "wal"), segmentSize, opts.WALCompression)
+		if err != nil && errors.Cause(err).Error() == wal.ErrorUnsequentialSegments {
+			level.Warn(db.logger).Log("msg", "encountered WAL segment sequence error, attempting repair", "err", err)
+			level.Warn(db.logger).Log("msg", "deleting unsequential segments")
+			if err = wlog.RepairUnsequentialSegments(filepath.Join(dir, "wal")); err != nil {
+				return nil, err
+			}
+			// Get new WAL after the repair of unsequential segments
```
codesome: `.` at the end of all the comments.

tsdb/db.go
```diff
@@ -538,6 +538,17 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db
 			segmentSize = opts.WALSegmentSize
 		}
 		wlog, err = wal.NewSize(l, r, filepath.Join(dir, "wal"), segmentSize, opts.WALCompression)
+		if err != nil && errors.Cause(err).Error() == wal.ErrorUnsequentialSegments {
+			level.Warn(db.logger).Log("msg", "encountered WAL segment sequence error, attempting repair", "err", err)
```
codesome:
```suggestion
			level.Warn(db.logger).Log("msg", "encountered unsequential WAL segments, attempting repair", "err", err)
```

tsdb/db.go
```diff
@@ -538,6 +538,17 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db
...
+			level.Warn(db.logger).Log("msg", "deleting unsequential segments")
```
codesome: This message is not clear as to which segments are being deleted. Before or after the gap.

tsdb/wal/wal_test.go
```diff
@@ -204,6 +204,56 @@ func TestWALRepair_ReadingError(t *testing.T) {
...
+	testutil.Ok(t, err)
+	defer func() {
+		testutil.Ok(t, os.RemoveAll(dir))
+	}()
+
+	var (
+		logger         = testutil.NewLogger(t)
+		segmentSize    = pageSize * 3
+		recordSize     = pageSize - recordHeaderSize
+		segments       = 20
```
codesome: `numSegments` would sound better here, I confused it with actual segments below.

tsdb/wal/wal_test.go
```diff
@@ -204,6 +204,56 @@ func TestWALRepair_ReadingError(t *testing.T) {
...
+
+	// Check if the total created segments have number equal to the segments
+	testutil.Assert(t, len(f) == segments, "Segments not created as expected")
+
+	// Delete a segment in order to create a gap or unsequential
+	deletionSegName := f[deletePosition].Name()
+	correctSegmentsCount := segments - deletePosition
+
+	// Create a gap in the middle of the sequence
+	os.Remove(dir + "/" + deletionSegName)
```
codesome: The returned error is not checked here. Also, after the deletion, you need to verify that there is 1 less than original segments on disk.

tsdb/wal/wal_test.go
```diff
@@ -204,6 +204,56 @@ func TestWALRepair_ReadingError(t *testing.T) {
...
+	w, err := Open(logger, nil, dir)
+	testutil.Ok(t, err)
```
codesome: We are not checking anywhere if the error was detected. A better test would be calling `NewSize` and verifying that the error cause was `ErrorUnsequentialSegments` and then go for repair.

tsdb/wal/wal_test.go
```diff
@@ -204,6 +204,56 @@ func TestWALRepair_ReadingError(t *testing.T) {
...
+
+	w.RepairUnsequentialSegments(dir)
```
codesome: Error should not be ignored here

tsdb/wal/wal_test.go
```diff
@@ -204,6 +204,56 @@ func TestWALRepair_ReadingError(t *testing.T) {
...
+
+	df, err := ioutil.ReadDir(dir)
+	testutil.Ok(t, err)
+	testutil.Assert(t, len(df) == correctSegmentsCount, "Unable to repair unsequential segments")
```
codesome: We also need to verify that the sequence is fine after the repair and that `NewSize` returns no error.

Additionally, verify which segments are present after the repair according to the expected behaviour. The number of segments would not be enough.

tsdb/wal/wal.go
```diff
@@ -774,6 +784,52 @@ func listSegments(dir string) (refs []segmentRef, err error) {
...
+}
+
+// RepairUnsequentialSegments repairs the out of sequence segments by deleting the onces after the sequence gap.
+// Returns the sequential segments after deleting the out of sequence wals after the sequence gap.
+func (w *WAL) RepairUnsequentialSegments(dir string) error {
+	segments, err := getSegments(dir)
+	if err != nil {
+		return errors.Wrap(err, "getSegments")
+	}
+	last := -1
```
codesome: Segments won't start with 0 always, this should be set to the 1st segment and start the loop from 2nd segment, else it will end up deleting all the segments according to the below logic,

krasi-georgiev:
> Thanks for the PR. Left few comments. As Krasi said we also need to tackle the underlying problem, but that would be for another PR.

I think we had few reports about the same thing so there is a bug somewhere. If we merge this PR that deleted all segments after the missing one this means ignoring the bug

krasi-georgiev:
The way I see it is that if this get merged the actual bug might never get fixed as Prometheus will just delete evertyhing after the missging segment and continue as normal meaning that it is highly unlekily that someone will actually report it again.

krasi-georgiev:
@brian-brazil I have seen few reports about this already and noone confirmed manual deletion.

I even [asked](https://github.com/prometheus/prometheus/issues/6080#issuecomment-559871635) @jacksontj about the fs and he confirmed that this happened on local storage

Harkishen-Singh:
@krasi-georgiev why can't we have both, the gap cleaning and the fix for the bug? We can have a prioritized issue for the bug.

krasi-georgiev:
Because just doesn't feel right to ignore the fact that there might be a bug and changing the code to deal with the consequences.

Harkishen-Singh:
@codesome do we plan to work on this, as I came across https://github.com/prometheus/prometheus/pull/6679#discussion_r372831557 or should we move ahead with closing this PR?

roidelapluie:

cc @codesome @jesusvazquez can you please have a look and update us on the state of this pull request?

PR #6130, opened 2019-10-11
testutil: Fail tests in ErrorEqual

dzeban:
ErrorEqual is renamed to ErrorsEqual and converted to fail tests just
like other functions in testuil package (e.g. `Ok` and `Assert`). This
allow to simplify test code and also fixes vain check in
scrape/manager_test.go.

### DIFF
--- a/promql/engine_test.go
+++ b/promql/engine_test.go
@@ -755,9 +755,8 @@ func TestRecoverEvaluatorRuntime(t *testing.T) {
 	//nolint:govet
 	a[123] = 1
 
-	if err.Error() != "unexpected error" {
-		t.Fatalf("wrong error message: %q, expected %q", err, "unexpected error")
-	}
+	unexpectedErr := errors.New("unexpected error")
+	testutil.ErrorsEqual(t, unexpectedErr, err)
 }
 
 func TestRecoverEvaluatorError(t *testing.T) {
@@ -767,9 +766,7 @@ func TestRecoverEvaluatorError(t *testing.T) {
 	e := errors.New("custom error")
 
 	defer func() {
-		if err.Error() != e.Error() {
-			t.Fatalf("wrong error message: %q, expected %q", err, e)
-		}
+		testutil.ErrorsEqual(t, e, err)
 	}()
 	defer ev.recover(&err)
 
--- a/util/testutil/error.go
+++ b/util/testutil/error.go
@@ -13,15 +13,11 @@
 
 package testutil
 
-// ErrorEqual compares Go errors for equality.
-func ErrorEqual(left, right error) bool {
-	if left == right {
-		return true
+// ErrorsEqual fails when Go errors are not equal.
+func ErrorsEqual(tb TB, exp, act error) {
+	if exp != nil && act != nil {
+		if exp.Error() != act.Error() {
+			tb.Fatalf("\033[31munexp error; want %v, got %v\033[39m\n", exp, act)
+		}
 	}
-
-	if left != nil && right != nil {
-		return left.Error() == right.Error()
-	}
-
-	return false
 }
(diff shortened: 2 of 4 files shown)


util/testutil/error.go
```diff
@@ -13,15 +13,11 @@
...
 package testutil
 
-// ErrorEqual compares Go errors for equality.
-func ErrorEqual(left, right error) bool {
-	if left == right {
-		return true
-	}
-
+// ErrorsEqual fails when Go errors are not equal.
+func ErrorsEqual(tb TB, left, right error) {
```
cstyan: It might be useful to also rename the the parameters to `expected, actual error`

dzeban: I renamed it to `exp` and `act` just like in `Equals` function.

util/testutil/error.go
```diff
@@ -13,15 +13,11 @@
...
 
-// ErrorEqual compares Go errors for equality.
-func ErrorEqual(left, right error) bool {
-	if left == right {
-		return true
+// ErrorsEqual fails when Go errors are not equal.
+func ErrorsEqual(tb TB, exp, act error) {
+	if exp != nil && act != nil {
+		if exp.Error() != act.Error() {
+			tb.Fatalf("\033[31munexp error; want %v, got %v\033[39m\n", exp, act)
```
codesome:
```suggestion
			tb.Fatalf("\033[31munexpected error; want %v, got %v\033[39m\n", exp, act)
```

util/testutil/error.go
```diff
@@ -13,15 +13,11 @@
 
 package testutil
 
-// ErrorEqual compares Go errors for equality.
-func ErrorEqual(left, right error) bool {
-	if left == right {
-		return true
+// ErrorsEqual fails when Go errors are not equal.
+func ErrorsEqual(tb TB, exp, act error) {
+	if exp != nil && act != nil {
```
codesome: We should fail even when one of them is `nil` and the other is not. You could follow the old implementation and just replace the `return false` cases with `tb.Fatalf`.

util/testutil/error.go
dzeban: Fair point :-)

util/testutil/error.go
dzeban: Yes, good point. Fixed.

promql/engine_test.go
```diff
@@ -755,9 +755,8 @@ func TestRecoverEvaluatorRuntime(t *testing.T) {
 	//nolint:govet
 	a[123] = 1
 
-	if err.Error() != "unexpected error" {
-		t.Fatalf("wrong error message: %q, expected %q", err, "unexpected error")
-	}
+	unexpectedErr := errors.New("unexpected error")
+	testutil.ErrorsEqual(t, err, unexpectedErr)
```
codesome:
```suggestion
	testutil.ErrorsEqual(t, unexpectedErr, err)
```

dzeban: Hmm, I thought that the other way would be more usual - we usually write `if err != nil` and not `if nil != err`. Also all `testutil.Assert` conditions are written this way. 

While go allows writing yoda condition (`if nil != err`) we don't do this in `testutil.Assert`. Take, for example, this piece from tsdb/db_test.go:

```go
	testutil.Equals(t, 0, len(lns))
	testutil.Equals(t, 0, len(lvs))
	testutil.Assert(t, res.Next() == false, "")
```
In case of `Equals` the tested variable is on the right, but in `Assert` it's on the left. For me it just seems incosistent.

I don't mind changing it as you're suggesting but I would rather do it the other way - change all invocations of `testutil.Equals` to read as ordinary go condition.

Anyway, this may be a subject of another PR, so I'm doing the suggested change here.

codesome:
Hi @dzeban, sorry for dropping the ball on this. Can you fix the conflicts?

dzeban:
> Hi @dzeban, sorry for dropping the ball on this. Can you fix the conflicts?

Yes, sure. Should be updated this week.

dzeban:
Actually, @roidelapluie fixed everything in 31700a05df64c2b4e32bb0ecd8baa25279144778, so I'm closing this PR

PR #6134, opened 2019-10-14
Update storage considerations documentation

SuperQ:
* Improve clarity of internal storage durability.
* Add notes about remote storage.

### DIFF
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -43,7 +43,9 @@ The directory structure of a Prometheus server's data directory will look someth
 ```
 
 
-Note that a limitation of the local storage is that it is not clustered or replicated. Thus, it is not arbitrarily scalable or durable in the face of disk or node outages and should thus be treated as more of an ephemeral sliding window of recent data. However, if your durability requirements are not strict, you may still succeed in storing up to years of data in the local storage.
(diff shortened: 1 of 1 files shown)


docs/storage.md
```diff
@@ -43,7 +43,9 @@ The directory structure of a Prometheus server's data directory will look someth
 ```
 
 
-Note that a limitation of the local storage is that it is not clustered or replicated. Thus, it is not arbitrarily scalable or durable in the face of disk or node outages and should thus be treated as more of an ephemeral sliding window of recent data. However, if your durability requirements are not strict, you may still succeed in storing up to years of data in the local storage.
+Note that a limitation of the local storage is that it is not clustered or replicated. Thus, it is not arbitrarily scalable or durable in the face of disk or node outages and should be treated as you would any other kind of single node database. Using RAID for disk availiablity, [snapshots](https://prometheus.io/docs/prometheus/latest/querying/api/#snapshot) for backups, capacity planning, etc, is recommended for improved durability. With proper storage durability and planning storing years of data in the local storage is possible.
```
brian-brazil: etc.

SuperQ: I'm not sure what you're suggesting.

brian-brazil: You're missing the period after etc

SuperQ: No, it's part of the complete sentence of `Using X,Y,Z, etc, is recommended...`.

SuperQ: How about `Using X,Y, and Z are recommended...`?

PR #6137, opened 2019-10-14
Step down and propose Ganesh as TSDB maintainer

gouthamve:
We should ideally move this up to the root MAINTAINERS.md, but the clean up of that is up for discussion at Promcon, so I'd rather postpone the move until after that discussion.

prometheus-developers thread: https://groups.google.com/d/msg/prometheus-developers/p8Ymwz5RWBw/jCfBPEMRCQAJ


gouthamve:
Done

beorn7:
Given all the approval an the mailing list thread, I'll merge this.

We'll talk about a general update of the MAINTAINERS.md file at the dev summit happening after PromCon on 2019-11-09.

PR #6138, opened 2019-10-14
Adding unit test for target group

jbleduigou:
This PR is about adding a unit test for target group for unmarshalling to yaml.


brian-brazil:
Thanks!

PR #6141, opened 2019-10-15
Adding unit test for target group

jbleduigou:
This PR is about adding a unit test for target group for marshalling to yaml.
I also improved slightly the test case for JSON.

If there are some other areas that need more tests I'd be happy to help!


discovery/targetgroup/targetgroup_test.go
```diff
@@ -17,32 +17,154 @@ import (
...
+		},
+		{
+			//brackets syntax
+			yaml:                    "labels:\n  my:  label\ntargets:\n  ['localhost:9090', 'localhost:9191']",
+			expectedNumberOfTargets: 2,
+			expectedNumberOfLabels:  1,
+			expectedReply:           nil,
+		},
+		{
+			//hyphen syntax
```
simonpasquier: We don't need to test every variant of YAML format so I'd remove this test case. It is assumed that the underlying YAML library manages it properly.

discovery/targetgroup/targetgroup_test.go
```diff
@@ -17,32 +17,154 @@ import (
...
+			expectedNumberOfLabels:  0,
+			expectedReply:           &yaml.TypeError{Errors: []string{"line 3: cannot unmarshal !!str `localho...` into []string"}},
+		},
+	}
+
+	for _, test := range tests {
+		tg := Group{}
+		actual := tg.UnmarshalYAML(unmarshal([]byte(test.yaml)))
+		testutil.Equals(t, test.expectedReply, actual)
+		testutil.Equals(t, test.expectedNumberOfTargets, len(tg.Targets))
```
simonpasquier: I could check that the resulting `tg` equals to what we expect like in `TestTargetGroupStrictJsonUnmarshal`.

jbleduigou:
Hi @simonpasquier,

I have addressed your comments.
However it seems that the imports are now misplaced.

discovery/targetgroup/targetgroup_test.go
```diff
@@ -17,75 +17,147 @@ import (
...
+	"github.com/prometheus/prometheus/util/testutil"
 )
 
 func TestTargetGroupStrictJsonUnmarshal(t *testing.T) {
 	tests := []struct {
-		json          string
-		expectedReply error
+		json            string
+		expectedReply   error
+		expectedTargets []model.LabelSet
```
simonpasquier: While you're at it, it might be worth extending the test to also cover `Group.Labels`. So instead of checking  `Group.Targets`, check that `Group` instances are equal.

discovery/targetgroup/targetgroup_test.go
```diff
@@ -17,75 +17,147 @@ import (
...
 			return yaml.Unmarshal(d, o)
 		}
 	}
 	tests := []struct {
-		yaml                    string
-		expectedNumberOfTargets int
-		expectedNumberOfLabels  int
-		expectedReply           error
+		yaml                   string
+		expectedTargets        []model.LabelSet
```
simonpasquier: Same here, check for `Group` equality.

discovery/targetgroup/targetgroup_test.go
jbleduigou: Yes, good idea.

discovery/targetgroup/targetgroup_test.go
jbleduigou: Sure no problem

simonpasquier:
:+1:

simonpasquier:
Thanks!

PR #6142, opened 2019-10-16
fix ignore file

stewart-yu:
fix ignore file


simonpasquier:
This should go to your [global](https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore) `.gitignore` file.

stewart-yu:
like that: https://github.com/kubernetes/kubernetes/blob/master/.gitignore . If we add once, other guys no need do setting in local ENV.

simonpasquier:
Our policy so far has been to include only binary files, configuration files and data directories in `.gitignore`. I don't think we want to change this rule for now.

Issue #6139, opened 2019-10-15
Error: http: superfluous response.WriteHeader call


bandesz:
## Bug Report

**What did you see?**

Multiple errors of `http: superfluous response.WriteHeader call` in the logs - during normal operation.

```
level=info ts=2019-10-14T11:01:57.535Z caller=head.go:666 component=tsdb msg="WAL checkpoint complete" first=90900 last=90995 duration=58.045866315s
level=error ts=2019-10-14T11:02:03.505Z caller=stdlib.go:89 component=web caller="http: superfluous response.WriteHeader call from github.com/opentracing-contrib/go-stdlib/nethttp.(*statusCodeTracker).WriteHeader (status-code-tracker.go:19" msg=)
level=error ts=2019-10-14T11:02:06.599Z caller=stdlib.go:89 component=web caller="http: superfluous response.WriteHeader call from github.com/opentracing-contrib/go-stdlib/nethttp.(*statusCodeTracker).WriteHeader (status-code-tracker.go:19" msg=)
level=error ts=2019-10-14T11:02:07.848Z caller=stdlib.go:89 component=web caller="http: superfluous response.WriteHeader call from github.com/opentracing-contrib/go-stdlib/nethttp.(*statusCodeTracker).WriteHeader (status-code-tracker.go:19" msg=)
[...]
```

**Environment**

* System information:

	busybox Docker image running the Prometheus binary.

* Prometheus version:

	```level=info ts=2019-10-09T12:24:15.591Z caller=main.go:332 msg="Starting Prometheus" version="(version=2.13.0, branch=HEAD, revision=6ea4252299f542669aca11860abc2192bdc7bede)"```

simonpasquier:
This is strange. Are you able to pinpoint a particular type of HTTP requests that would trigger it?

kfdm:
I've noticed the same error on many of my own servers. I'll see if I can narrow down the trigger as well.

beorn7:
This could be caused by a bug in prometheus/client_golang addressed here: https://github.com/prometheus/client_golang/pull/634
It is fixed in v1.2.0 (released yesterday), but Prometheus v2.13.0 only uses v1.1.0.

beorn7:
This is a plausible course of action:

[...]

PR #6145, opened 2019-10-16
Cut release 2.13.1

krasi-georgiev:


VERSION
```diff
@@ -1 +1 @@
-2.13.0
\ No newline at end of file
+2.13.1
```
beorn7: Shouldn't this have a newline at the end? This one is the only VERSION file in any Prometheus project without a trailing newline.

simonpasquier: I'm guessing it is an artifact of https://github.com/prometheus/promu/pull/170

beorn7: I'd suggest to fix it in `promu bump` then.

simonpasquier: In progress though `promu bump` doesn't exist yet ;-)

krasi-georgiev: > Thinking about #6139 : Perhaps it will help to bump prometheus/client_golang to v1.2.0.

IIUC the patch release should add only include bug fixes introduced since the last version right? 
Is this one of those or has been there before that?

krasi-georgiev:
> Thinking about #6139 : Perhaps it will help to bump prometheus/client_golang to v1.2.0.

IIUC the patch release should add only include bug fixes introduced since the last version right? 
Is this one of those or has been there before that?

beorn7:
I'm not sure when the bug was introduced.
But I think all relevant bug fixes should make it into bug fix releases. That's what they are for. (So that users who want the bug fixes don't have to deal with non-bug-fix changes and their impact, which comes with minor and major releases.)

krasi-georgiev:
> But I think all relevant bug fixes should make it into bug fix releases. That's what they are for. (So that users who want the bug fixes don't have to deal with non-bug-fix changes and their impact, which comes with minor and major releases.)

true, true, but that means cutting a new patch release every week which I wouldn't have the time to do so.

beorn7:
But you are cutting a patch release right now. That's why I suggest to do the easy bump of client_golang to get a bugfix in, which might even fix #6139 . Low risk, low effort, nice potential benefit. And it avoids cutting another patch release next week.

beorn7:
Many thanks.

But now it looks you need to run a `go get` and then `go mod tidy` and `go mod vendor` again.
(Or something like that, still getting into that module stuff.)

krasi-georgiev:
yep I did all of that and there is still some problem. Troubleshooting now.

krasi-georgiev:
@beorn7 all good now, do you have any other comments?

PR #6149, opened 2019-10-17
Fix solaris build.

ptribble:
Signed-off-by: Peter Tribble <peter.tribble@gmail.com>

### DIFF
--- a/pkg/runtime/statfs.go
+++ b/pkg/runtime/statfs.go
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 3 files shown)


brian-brazil:
Does statfs_linux_386.go also need updating?

brian-brazil:
Can we remove the line with netbsd etc in that file then?

ptribble:
I think you can (and the other excludes). I would say that it's not part of this change, and I have no access to those platforms to test it either.

brian-brazil:
The CI will cover it.

brian-brazil:
Thanks!

PR #6088, opened 2019-10-02
Add tsdb commands as promtool tsdb subcommand

yeya24:
Signed-off-by: yeya24 <yb532204897@gmail.com>

fixes #6087

PR #6150, opened 2019-10-17
Include the tsdb tool in docker images

askoriy:


simonpasquier:
The `tsdb` tool is to be included with `promtool`. See https://github.com/prometheus/prometheus/pull/6088

roidelapluie:
tsdb will be included in promtool the next release. Thanks for your PR anyway!

PR #6154, opened 2019-10-17
Update the golang client to v1.2.1

krasi-georgiev:


beorn7:
It's probably still good to merge this, but I wouldn't let it alone trigger a new bugfix release.

PR #6165, opened 2019-10-17
React UI: Sanitize HTML string to allow only supported elements

codingdash:


web/ui/react-app/src/ExpressionInput.tsx
```diff
@@ -72,9 +73,9 @@ class ExpressionInput extends Component<ExpressionInputProps> {
                   },
                 })}
               >
-                {/* TODO: Find better way than setting inner HTML dangerously. We just want the <strong> to not be escaped.
-                    This will be a problem when we save history and the user enters HTML into a query. */}
-                <span dangerouslySetInnerHTML={{__html: item.string}}></span>
+                <SanitizeHTML inline={true}>
```
juliusv: Could we explicitly only allow the `<strong>` tag?

web/ui/react-app/src/ExpressionInput.tsx
```diff
@@ -95,63 +96,68 @@ class ExpressionInput extends Component<ExpressionInputProps> {
...
-                  placeholder="Expression (press Shift+Enter for newlines)"
-                  innerRef={this.exprInputRef}
-                  {...downshift.getInputProps({
-                    onKeyDown: (event: React.KeyboardEvent): void => {
-                      switch (event.key) {
-                        case 'Home':
-                        case 'End':
-                          // We want to be able to jump to the beginning/end of the input field.
-                          // By default, Downshift otherwise jumps to the first/last suggestion item instead.
+      <Downshift
```
juliusv: Did anything change in this block besides formatting? It's hard to see with the whole block being reindented.

web/ui/react-app/src/ExpressionInput.tsx
```diff
@@ -8,8 +8,9 @@ import {
   Input,
 } from 'reactstrap';
 
-import Downshift, { ChildrenFunction, ControllerStateAndHelpers, DownshiftInterface } from 'downshift';
```
juliusv: I just fixed this and other linter errors in master - could you rebase?

codingdash: This is done.

web/ui/react-app/src/ExpressionInput.tsx
codingdash: Yes, improved `SanitizeHTML` component to accept props `allowedTags` using which it's whitelisting tags.

juliusv:
:+1: Wonderful, thank you :)

juliusv:
One observation though: this PR added about 300kb to the final size of the `prometheus` binary (80155524 vs. 79840132 bytes). That seems quite heavy to me for some HTML sanitization...

codingdash:
@juliusv If I have correctly understood by `prometheus` binary you mean the React app assets sizes? The size diff I see following-

**[Before adding sanitize-html](https://circleci.com/gh/prometheus/prometheus/17076)**
```
File sizes after gzip:

  204.06 KB  build/static/js/2.17185238.chunk.js
  23.12 KB   build/static/css/2.164b004b.chunk.css
  7.2 KB     build/static/js/main.e6039d22.chunk.js
  912 B      build/static/css/main.2ac95e7f.chunk.css
  763 B      build/static/js/runtime~main.d653cc00.js
```
**[After adding sanitize-html](https://circleci.com/gh/prometheus/prometheus/17172)**
```
File sizes after gzip:

  279.21 KB  build/static/js/2.b257cdc5.chunk.js
  23.12 KB   build/static/css/2.bae5dac8.chunk.css
  7.49 KB    build/static/js/main.091b61ea.chunk.js
  913 B      build/static/css/main.73fd8119.chunk.css
  769 B      build/static/js/runtime-main.56ad181a.js
```

The difference here is ~75.44KB, but still it's more for just sanitization. Unfortunately, [sanitize-html](https://github.com/apostrophecms/sanitize-html/blob/master/package.json) depends on some heavy libs like [chalk](https://www.npmjs.com/package/chalk) and [htmlparser2](https://www.npmjs.com/package/htmlparser2) to sanitize lib. But at the moment this is just one library which does this job perfectly.

juliusv:
@codingdash I meant the actual size of the `prometheus` binary file. Interesting that the underlying bundle size so much less! When you run `make build`, it generates a `web/ui/assets_vfsdata.go` file that gets compiled into the binary, but it just contains byte slices with the file contents to serve. So I'm not sure how this can trigger the blowup. ¯\_(ツ)_/¯

codingdash:
Ah, thanks :) I wasn't aware of this.

PR #6176, opened 2019-10-18
.circleci/config.yml: fix publish jobs

simonpasquier:
As a quick fix, the publish jobs will use an image with npm/yarn installed. Eventually we could use the license tarball from the `build` job.


juliusv:
:+1: Thanks!

PR #6181, opened 2019-10-18
.circleci/config.yml: cache dependencies

simonpasquier:


.circleci/config.yml
```diff
@@ -21,6 +22,12 @@ jobs:
 
     steps:
     - prometheus/setup_environment
+    - go/load-cache:
+        key: v1
+    - restore_cache:
+        keys:
+        - v1-npm-deps-{{ checksum "web/ui/react-app/yarn.lock" }}
+        - v1-npm-deps-
```
juliusv: Do we need the fallback, since `yarn.lock` should always be there?

juliusv:
:+1: Either way

.circleci/config.yml
SuperQ: +1, I'm not sure exactly how this work.

juliusv: Ah ok, makes sense. :+1:

PR #6187, opened 2019-10-20
scrape: Move tests to testutil

dzeban:

### DIFF
--- a/scrape/scrape_test.go
+++ b/scrape/scrape_test.go
@@ -23,7 +23,6 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
-	"reflect"
 	"strings"
 	"sync"
 	"testing"
@@ -196,17 +195,11 @@ func TestScrapePoolStop(t *testing.T) {
 	}
 
 	mtx.Lock()
-	if len(stopped) != numTargets {
-		t.Fatalf("Expected 20 stopped loops, got %d", len(stopped))
-	}
+	testutil.Equals(t, numTargets, len(stopped), "Unexpected number of stopped loops")
 	mtx.Unlock()
 
-	if len(sp.activeTargets) > 0 {
-		t.Fatalf("Targets were not cleared on stopping: %d left", len(sp.activeTargets))
-	}
-	if len(sp.loops) > 0 {
-		t.Fatalf("Loops were not cleared on stopping: %d left", len(sp.loops))
-	}
+	testutil.Assert(t, len(sp.activeTargets) == 0, "Targets were not cleared on stopping: %d left", len(sp.activeTargets))
+	testutil.Assert(t, len(sp.loops) == 0, "Loops were not cleared on stopping: %d left", len(sp.loops))
 }
 
 func TestScrapePoolReload(t *testing.T) {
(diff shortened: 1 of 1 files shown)


scrape/scrape_test.go
```diff
@@ -196,17 +195,11 @@ func TestScrapePoolStop(t *testing.T) {
 	}
 
 	mtx.Lock()
-	if len(stopped) != numTargets {
-		t.Fatalf("Expected 20 stopped loops, got %d", len(stopped))
-	}
+	testutil.Equals(t, len(stopped), numTargets)
```
cstyan: Can you include the message here so we know which value was expected vs what the output was in the test, you can include a string after `numTargets`

dzeban: Good point, thanks! I've update the PR.

scrape/scrape_test.go
```diff
@@ -744,15 +716,10 @@ func TestScrapeLoopRunCreatesStaleMarkersOnFailedScrape(t *testing.T) {
...
-	if appender.result[0].v != 42.0 {
-		t.Fatalf("Appended first sample not as expected. Wanted: %f Got: %f", appender.result[0].v, 42.0)
-	}
-	if !value.IsStaleNaN(appender.result[6].v) {
-		t.Fatalf("Appended second sample not as expected. Wanted: stale NaN Got: %x", math.Float64bits(appender.result[5].v))
-	}
+	testutil.Equals(t, 27, len(appender.result), "Appended samples not as expected")
+	testutil.Equals(t, 42.0, appender.result[0].v, "Appended first sample not as expected")
+	testutil.Assert(t, value.IsStaleNaN(appender.result[6].v),
+		"Appended second sample not as expected. Wanted: stale NaN Got: %x", math.Float64bits(appender.result[5].v))
```
csmarchbanks: Possibly a bug in the previous code, but shouldn't both be `result[6]`?

scrape/scrape_test.go
```diff
@@ -806,25 +773,19 @@ func TestScrapeLoopRunCreatesStaleMarkersOnParseFailure(t *testing.T) {
...
-	if appender.result[0].v != 42.0 {
-		t.Fatalf("Appended first sample not as expected. Wanted: %f Got: %f", appender.result[0].v, 42.0)
-	}
-	if !value.IsStaleNaN(appender.result[6].v) {
-		t.Fatalf("Appended second sample not as expected. Wanted: stale NaN Got: %x", math.Float64bits(appender.result[6].v))
-	}
+	testutil.Equals(t, 17, len(appender.result), "Appended samples not as expected")
+	testutil.Equals(t, 42.0, appender.result[0].v, "Appended first sample not as expected")
+	testutil.Assert(t, value.IsStaleNaN(appender.result[6].v),
+		"Appended second sample not as expected. Wanted: stale NaN Got: %x", math.Float64bits(appender.result[5].v))
```
csmarchbanks: Should be `result[6]` here.

scrape/scrape_test.go
dzeban: Yes, sure. Fixed that.

scrape/scrape_test.go
dzeban: Also fixed.

Issue #6156, opened 2019-10-17
React UI: Re-add query stats information


juliusv:
The [new React UI](https://github.com/prometheus/prometheus/pull/5694) does not yet show query stats (execution time, resolution, and number of result time series) yet. It would be good to re-add this feature somehow, but care should be taken to not clutter the new UI unnecessarily. Maybe we can find a way to only display it on request (like collapsing it first or similar).

CSTDev:
I'll take this on if no one else is...

CSTDev:
Initial thoughts, uses a reactstrap collapse. 
Have just thought actually that having it there will be quite jarring as it moves everything below it down and up as it's expanded/collapsed. Will re-think.

![query-stats-1](https://user-images.githubusercontent.com/3914110/67109706-07050200-f1c9-11e9-98ee-f67631f371bc.PNG)
![query-stats-1-expanded](https://user-images.githubusercontent.com/3914110/67109708-079d9880-f1c9-11e9-9653-87936bf56a5b.PNG)

juliusv:
Yeah, another inspiration could be Grafana's explore mode. See this screenshot, where some stats appear very unobtrusively at the far right of the expression box:

![](https://grafana.com/docs/img/docs/v60/explore_prometheus.png)

That breaks down at a certain amount of stats of course, and one would have to take care that it doesn't interfere with long expression lines.

Not saying it should be like this, but it's one more idea of a possible direction.

juliusv:
@CSTDev Here's two more mockup variants that I find relatively unobtrusive, making use of the unused space of the Graph/Table tabpane on the right:

![variant2](https://user-images.githubusercontent.com/538008/67158315-fd011180-f336-11e9-82e8-bcb7994dce11.png)

I hacked this by putting this div into the HTML:

```html
<div style="
    position: absolute;
    right: 18px;
    top: 43px;
    font-size: 9px;
    color: #999;
">Query time: 234ms<br>Time series: 1023<br>Resolution: 13s</div>
```

And:

![variant1](https://user-images.githubusercontent.com/538008/67158323-0ee2b480-f337-11e9-9875-f1dcd023851a.png)

...using:

```html
<div style="
    position: absolute;
    right: 16px;
    top: 42px;
    font-size: 9px;
    color: #999;
">Query time: 234ms - Time series: 1023 - Resolution: 13s</div>
```

What do you think? It always shows the information, but still doesn't seem too clutter-y IMO. But happy for other options too :)

[...]

PR #6190, opened 2019-10-21
React ui add query stats

CSTDev:
Puts query stats in a useful but unobtrusive place as per #6156


web/ui/react-app/src/QueryStats.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+class QueryStats extends PureComponent<QueryStatsProps>{
+
+    render() {
+        const stats = this.props.stats;
+        const loading = this.props.loading;
+
+        return (
+            <div className="query-stats">
```
boyskila: @CSTDev is it needed this component to be created at all if !loading && stats !== null?

web/ui/react-app/src/QueryStats.tsx
```diff
@@ -0,0 +1,29 @@
...
+interface QueryStatsProps {
+    stats: {
+        loadTime: number;
+        resolution: number;
+        totalTimeSeries: number;
+    } | null;
+    loading: boolean;
+}
+
+class QueryStats extends PureComponent<QueryStatsProps>{
```
boyskila: I would suggest you convert it to Functional component since is not maintaining any state or dealing with any of the lifecycles hooks. If you want to preserve it like PureComponent just use memo from React

web/ui/react-app/src/Panel.tsx
```diff
@@ -36,7 +37,11 @@ interface PanelState {
   } | null;
   loading: boolean;
   error: string | null;
-  stats: null; // TODO: Stats.
+  stats: {
```
boyskila: I would suggest you create interface PaneStats or something like that and to use it here and in QueryStats component as well

web/ui/react-app/src/Panel.tsx
```diff
@@ -161,6 +175,11 @@ class Panel extends Component<PanelProps, PanelState> {
           endTime: endTime,
           resolution: resolution,
         },
+        stats: {
+          loadTime: Date.now() - queryStart,
+          resolution: resolution,
```
boyskila: Just resolution is enough, since key: value have same names, same for totalTimeSeries

web/ui/react-app/src/QueryStats.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+    render() {
+        const stats = this.props.stats;
```
boyskila: Just a personal preference. I would destruct the values
const { stats, loading } = this.props
const { loadTime, resolution, totalTimeSeries } = stats

web/ui/react-app/src/Panel.tsx
```diff
@@ -287,7 +307,7 @@ class Panel extends Component<PanelProps, PanelState> {
         <Row>
           <Col>
             <Button className="float-right" color="link" onClick={this.props.removePanel} size="sm">Remove Panel</Button>
-          </Col>
+          </Col>
```
boyskila: unnecessary spaces

web/ui/react-app/src/Panel.tsx
```diff
@@ -153,6 +158,15 @@ class Panel extends Component<PanelProps, PanelState> {
         throw new Error(json.error || 'invalid response JSON');
       }
 
+      let totalTimeSeries = 0;
+      if (json.data !== undefined) {
```
boyskila: I think this will fail if json.data is null

web/ui/react-app/src/Panel.tsx
```diff
@@ -153,6 +158,15 @@ class Panel extends Component<PanelProps, PanelState> {
...
+        if (json.data.resultType === "scalar") {
+          totalTimeSeries = 1;
+        } else if(json.data.result !== null) {
```
boyskila: json.data.result !== null will fail too if json.data.result is undefined

boyskila: Maybe some utility function like isPresent(someValue) will be useful, where isPresent will perform a check for someValue !== null && someValue !== undefined

web/ui/react-app/src/QueryStats.tsx
CSTDev: Fair point will do the check earlier.

web/ui/react-app/src/Panel.tsx
CSTDev: Changed it to if(json.data.result) which should evaluate to false if it's null or undefined.

boyskila: 0 as well :) I think proper check here would (be value !== null && value !== undefined)

web/ui/react-app/src/Panel.tsx
CSTDev: Did some renaming, and shared it from QueryStats.

web/ui/react-app/src/Panel.tsx
CSTDev: Added a check on length too in case it's an empty object.

web/ui/react-app/src/Panel.tsx
```diff
@@ -157,29 +153,29 @@ class Panel extends Component<PanelProps, PanelState> {
...
-
+      
       let totalTimeSeries = 0;
-      if (json.data !== undefined) {
+      if (json.data) {
         if (json.data.resultType === "scalar") {
           totalTimeSeries = 1;
-        } else if(json.data.result !== null) {
+        } else if(json.data.result && json.data.result.length > 0) {
           totalTimeSeries = json.data.result.length;
```
boyskila: Zero value shouldn't be displayed?

web/ui/react-app/src/Panel.tsx
```diff
@@ -157,29 +153,29 @@ class Panel extends Component<PanelProps, PanelState> {
       if (json.status !== 'success') {
         throw new Error(json.error || 'invalid response JSON');
       }
-
+
```
boyskila: extra spaces :) Not sure wheter is IDE setting or tslint.json should be added to the project to catch such errors. @juliusv

web/ui/react-app/src/QueryStatsView.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import './QueryStatsView.css';
+
+export interface QueryStats{
+    loadTime: number;
+    resolution: number;
+    totalTimeSeries: number;
+}
+
+export default function QueryStatsView(props: QueryStats){
```
boyskila: I think this should looks like
```
const QueryStatsView = FC<QueryStats> = ({ loadTime, resolution, totalTimeSeries}) => { .... }
```

web/ui/react-app/src/QueryStatsView.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import './QueryStatsView.css';
+
+export interface QueryStats{
```
boyskila: space before bracket is missing.

web/ui/react-app/src/Panel.tsx
```diff
@@ -157,29 +153,29 @@ class Panel extends Component<PanelProps, PanelState> {
       if (json.status !== 'success') {
         throw new Error(json.error || 'invalid response JSON');
       }
-
+      
       let totalTimeSeries = 0;
-      if (json.data !== undefined) {
+      if (json.data) {
         if (json.data.resultType === "scalar") {
```
boyskila: You can destruct all data values like const { resultType, result } json.data

web/ui/react-app/src/Panel.tsx
CSTDev: No they shouldn't and so shouldn't ever get here if there are 0 results, but it's just in case something odd happens, it then displays a 0, which would be correct.

web/ui/react-app/src/QueryStatsView.tsx
CSTDev: Hmm, not see that form before and I can't get anything like it working? No point expanding them out as I created an interface to define them that's then shared too.
Going for:
`const QueryStatsView = (props: QueryStats) => {...}`

boyskila: Yep, That's fine too

web/ui/react-app/src/Panel.tsx
juliusv: Yeah, I'm all for automating this stuff :) https://github.com/prometheus/prometheus/issues/6199

web/ui/react-app/src/Panel.tsx
```diff
@@ -153,14 +154,29 @@ class Panel extends Component<PanelProps, PanelState> {
         throw new Error(json.error || 'invalid response JSON');
```
juliusv: When there's a query error, we should either hide the previously shown query stats or show stats for the errored query (like the old UI does, see http://demo.robustperception.io:9090/graph?g0.range_input=1h&g0.expr=-invalid-&g0.tab=1).

web/ui/react-app/src/QueryStatsView.tsx
```diff
@@ -0,0 +1,20 @@
...
+    resolution: number;
+    totalTimeSeries: number;
+}
+
+const QueryStatsView = (props: QueryStats) => {
+    const {loadTime, resolution, totalTimeSeries} = props;
+
+    return (
+        <div className="query-stats">
+            <span className="float-right">Load Time: {loadTime}ms &ensp; Resolution: {resolution} &ensp; Total Time Series: {totalTimeSeries}</span>
```
juliusv: Here's where my specific sense of aesthetics kicks in ;)

Could we change this to:

```suggestion
            <span className="float-right">Load time: {loadTime}ms &ensp; Resolution: {resolution}s &ensp; Result series: {totalTimeSeries}</span>
```

(shorter wording, no title case, and adding a unit to the resolution)

web/ui/react-app/src/Panel.tsx
```diff
@@ -153,14 +154,29 @@ class Panel extends Component<PanelProps, PanelState> {
...
       }
 
+      let totalTimeSeries = 0;
+      if (json.data) {
+        const { resultType, result} = json.data;
```
juliusv: Whitespace consistency nit:

```suggestion
        const { resultType, result } = json.data;
```

web/ui/react-app/src/Panel.tsx
```diff
@@ -153,14 +154,29 @@ class Panel extends Component<PanelProps, PanelState> {
...
+        if (resultType === "scalar") {
+          totalTimeSeries = 1;
+        } else if(result && result.length > 0) {
```
juliusv: Whitespace consistency nit:

```suggestion
        } else if (result && result.length > 0) {
```

juliusv:
Looks great besides nits, thanks :)

web/ui/react-app/src/QueryStatsView.tsx
CSTDev: 🤣 I looked at it and thought about the case, but went with this one. Will change it.

web/ui/react-app/src/Panel.tsx
CSTDev: Ah yep good point. Have sorted that.

web/ui/react-app/src/QueryStatsView.tsx
CSTDev: Figured it out, so done it the way you suggested.

juliusv:
:+1: Awesome, looks good now, thanks!

PR #6192, opened 2019-10-21
adding additional unit tests for getDataCenter() in consul

jbleduigou:
Signed-off-by: Jean-Baptiste Le Duigou <jb.leduigou@gmail.com>


discovery/consul/consul_test.go
```diff
@@ -297,3 +297,89 @@ func TestAllOptions(t *testing.T) {
 	checkOneTarget(t, <-ch)
 	cancel()
 }
+
+func TestGetDatacenter_ShouldUseAgentInfo(t *testing.T) {
```
simonpasquier: This path should already be covered by the existing tests AFAICT.

discovery/consul/consul_test.go
```diff
@@ -297,3 +297,89 @@ func TestAllOptions(t *testing.T) {
...
+	// second call should use value cached
+	err = d.getDatacenter()
+
+	// no error expected here
+	testutil.Equals(t, nil, err)
+	// should be value of agent info once initialized
+	testutil.Equals(t, "test-dc", d.clientDatacenter)
+}
+
+func TestGetDatacenter_ShouldReturnErrorGivenInternalServerError(t *testing.T) {
```
simonpasquier: You can use table-driven tests to group this test and the next one in a single test since most of the code is common.

jbleduigou:
I have updated the comments with proper style.
I will be looking at the table-driven test.

jbleduigou:
Update negative scenarios with a single table-driven test.

jbleduigou:
> You can take it even one step further.
> 
> ```go
> for _, tc := range []struct{
>   handler func(http.ResponseWriter, *http.Request)
>   errMsg string
> } {
>     {
>       handler: ...,
>       errMsg: "...",
>     },
>     {
>       handler: ...,
>       errMsg: "...",
>     },
> } {
>   stub := httptest.NewServer(http.HandlerFunc(tc.handler))
>   ...
> }
> ```

Yes indeed, that is a much cleaner syntax

jbleduigou:
Hi @simonpasquier,
Could you please review the latest changes and let me know if I need to make additional changes?
Thanks

discovery/consul/consul_test.go
```diff
@@ -297,3 +297,72 @@ func TestAllOptions(t *testing.T) {
...
+	// second call should use value cached.
+	err = d.getDatacenter()
+
+	// No error expected here.
+	testutil.Equals(t, nil, err)
+	// Should still be value of agent info.
+	testutil.Equals(t, "test-dc", d.clientDatacenter)
+}
+
+func TestGetDatacenter_ShouldReturnError(t *testing.T) {
```
simonpasquier: `TestGetDatacenterShouldReturnError` as per https://golang.org/doc/effective_go.html#mixed-caps

discovery/consul/consul_test.go
simonpasquier: As said above, it should already be covered so I'd rather remove this test.

discovery/consul/consul_test.go
jbleduigou: Sure, I will fix that.

discovery/consul/consul_test.go
jbleduigou: Test was removed.

simonpasquier:
:+1: 
Thanks!

jbleduigou:
Thanks for the review!

PR #6193, opened 2019-10-21
migrate query history

boyskila:


boyskila:
@juliusv I think it's working fine. Also, I found and fix two small issues in the Expression input, so now should be even better :)

web/ui/react-app/src/App.css
```diff
@@ -10,6 +10,24 @@ button.classic-ui-btn {
     margin-bottom: 20px;
 }
 
+.checkbox{
```
prymitive: New UI uses reactstrap, which uses bootstrap and both have style definitions for checkboxes. Shouldn't the new UI re-use bootstrap styles for basic elements like checkboxes?
There's also support for cross-OS consistient style - https://getbootstrap.com/docs/4.3/components/forms/#checkboxes-and-radios-1

web/ui/react-app/src/App.css
```diff
@@ -10,6 +10,24 @@ button.classic-ui-btn {
...
+    display: inline-flex;
+    align-items: center;
+    cursor: pointer;
+    margin: 0 0 13px 3px;
+}
+
+.checkbox.checked {
+    color: #286090
```
juliusv: I think this color doesn't fit anymore to the bootstrap version that the new UI is using. Also, I added a "Return to classic UI" at the top right and in the old UI made it share the same row as the query history checkbox. Could we adjust colors (same as "Return to classic UI" link), font sizes, and positioning to be more like that? Dirty mockup using float left/right for query history + back link in a parent container:

query_history_mockup

web/ui/react-app/src/App.css
juliusv: Consistency nit: space before `{`.

web/ui/react-app/src/App.tsx
```diff
@@ -7,9 +7,19 @@ import {
...
 } from 'reactstrap';
 
+import './App.css';
 import PanelList from './PanelList';
+import { library } from '@fortawesome/fontawesome-svg-core';
+import { faSearch, faSpinner, faCheck } from '@fortawesome/free-solid-svg-icons';
+import { faSquare, faCheckSquare } from '@fortawesome/free-regular-svg-icons';
 
-import './App.css';
+library.add(
```
juliusv: There's already a couple of `library.add()` statements in the app, but they are local to the components that need them (`TimeInput`, `ExpressionInput`, and `GraphControls`). Wouldn't it make sense to keep that pattern where every component adds what it needs? I also see this call adding some icons that other `library.add()` calls add (and which the checkbox doesn't need), but without removing those other calls?

web/ui/react-app/src/Checkbox.tsx
```diff
@@ -0,0 +1,23 @@
...
+  }
+  return (
+    <span {...rest} onClick={handleClick} className={`checkbox${(selected ? ' checked' : '')}`}>
+      <FontAwesomeIcon size={size} icon={['far', (selected ? 'check-square' : 'square')]}/>
+      <span className="checkbox-label">{children}</span>
+    </span>
+  )
+};
+
+export default memo(Checkbox);
```
juliusv: Nit: please configure your editor to always end files with a terminating newline character (e.g. `"files.insertFinalNewline": true` in VS Code).

web/ui/react-app/src/Checkbox.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { FC, useState, HTMLProps, memo } from 'react';
+import { FontAwesomeIcon, FontAwesomeIconProps } from '@fortawesome/react-fontawesome';
+
+interface CheckboxProps extends Omit<HTMLProps<HTMLSpanElement>, 'size' | 'onClick'> {
+  onToggle: (selected: boolean) => void
+}
+
+const Checkbox: FC<CheckboxProps & Pick<FontAwesomeIconProps, 'size'>> = props => {
```
juliusv: I learned new TypeScript + React things in this file, thanks :laughing:

web/ui/react-app/src/Panel.tsx
```diff
@@ -100,6 +101,7 @@ class Panel extends Component<PanelProps, PanelState> {
   }
 
   executeQuery = (expr: string): void => {
+    this.props.onExecuteQuery(expr)
```
juliusv: Nit: no semicolon (can we automate this via linter checks?)

web/ui/react-app/src/PanelList.tsx
```diff
@@ -17,7 +18,7 @@ interface PanelListState {
 
 class PanelList extends Component<any, PanelListState> {
   private key: number = 0;
-
+  private initialMetricsNames: string[] = [];
```
juliusv: I think it'd just be "metric names" rather than "metrics names".

web/ui/react-app/src/App.css
boyskila: Thanks, I'll check later but definitely sounds better than SVG icons from @fortawesome/react-fontawesome

web/ui/react-app/src/PanelList.tsx
```diff
@@ -75,6 +79,40 @@ class PanelList extends Component<any, PanelListState> {
...
+    localStorage.setItem('enable-query-history', `${enabled}`);
+    this.setMetrics();
+  }
+
+  setMetrics = () => {
+    if (this.isHistoryEnabled()) {
+      const historyItems = this.getHistoryItems();
+      const { length } = historyItems;
+      this.setState({
+        metricNames: [...historyItems.slice(length - 50, length), ...this.initialMetricsNames],
```
juliusv: I think it'd be good to keep query history items separate from metric names somehow. We could either pass in two different properties (`metricNames` and `pastQueries`) or a single object in which autocompletion items have different types (`autocomplete={{metricNames: [...], pastQueries: [...]}}`). Because I'd like to differentiate them in the dropdown later. Super ugly mockup for demonstration purposes:

differentiated_autocomplete

web/ui/react-app/src/App.tsx
```diff
@@ -7,9 +7,19 @@ import {
   Row,
 } from 'reactstrap';
 
+import './App.css';
 import PanelList from './PanelList';
+import { library } from '@fortawesome/fontawesome-svg-core';
+import { faSearch, faSpinner, faCheck } from '@fortawesome/free-solid-svg-icons';
+import { faSquare, faCheckSquare } from '@fortawesome/free-regular-svg-icons';
```
juliusv: Hmm, while I would like to just keep using the solid icons like in the rest of the app, the solid square looks a bit stupid. So I guess this is fine until we find something better :)

web/ui/react-app/src/App.css
juliusv: Ah, I didn't see this comment before I posted my entire review (multiple comments at once). But yes, using the native reactstrap styles probably makes way more sense!

web/ui/react-app/src/App.css
boyskila: Yep, will make those adjustments, thanks

web/ui/react-app/src/App.tsx
boyskila: So, as per library docs this should be placed in the root Component in order icons to be available in the whole app and it make sense if same icons are used in multiple places. Thus this kind of setup will be in one place but generally, I would suggest finding another, more simple solution instead of library.add(), dom.watch(), importing more than one package, and so on :) WDYT?

web/ui/react-app/src/Checkbox.tsx
boyskila: Thanks, I'll check that setting

web/ui/react-app/src/Checkbox.tsx
boyskila: Yes, generally this will go away since I'm going to replace it with reactstrap Input

web/ui/react-app/src/Panel.tsx
boyskila: Yes, I can make such a setup.

web/ui/react-app/src/PanelList.tsx
boyskila: thanks, will fix the typo

web/ui/react-app/src/PanelList.tsx
boyskila: Yeah, this looks pretty organized.

web/ui/react-app/src/App.css
boyskila: This can be a tslint rule

web/ui/react-app/src/App.tsx
prymitive: Do you really need to use the `library.add()`? 
I was under the impression that you can get tree shaking (if that's what you're after) by importing each icon manually and passing it "raw" with:

```
import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
import { faTrash } from "@fortawesome/free-solid-svg-icons/faTrash";
[...]
<FontAwesomeIcon icon={faTrash} />
```

boyskila: Yes, this pattern definitely looks better and no additional setup is needed :)

web/ui/react-app/src/Checkbox.tsx
prymitive: prettier is a pretty decent tool to enforce consistent style in es code (gofmt for js), you can hook it into eslint via https://prettier.io/docs/en/integrating-with-linters.html
there are also plugins for editors so it's run on file save

web/ui/react-app/src/App.tsx
juliusv: That seems even better, yeah :)

I bet when I did this back in February (while learning TypeScript + React at the same time), I just saw a page like https://fontawesome.com/how-to-use/with-the-api/setup/library and assumed that one should use the library.

web/ui/react-app/src/Checkbox.tsx
juliusv: Thanks, I filed https://github.com/prometheus/prometheus/issues/6199 for someone to improve the linter situation :)

web/ui/react-app/src/Panel.tsx
juliusv: See https://github.com/prometheus/prometheus/issues/6199

web/ui/react-app/src/PanelList.tsx
boyskila: @juliusv, I'm going to implement this in separate PR since too much things going on here :)

web/ui/react-app/src/App.css
```diff
@@ -1,185 +1,191 @@
 body {
-    padding-top: 10px; /* TODO remove */
+  padding-top: 10px; /* TODO remove */
```
juliusv: Hmm, could we omit all the unrelated CSS formatting changes in this PR?

web/ui/react-app/src/PanelList.tsx
```diff
@@ -38,34 +39,37 @@ class PanelList extends Component<any, PanelListState> {
...
-      if (delta >= 30) {
-        throw new Error('Detected ' + delta + ' seconds time difference between your browser and the server. Prometheus relies on accurate time and time drift might cause unexpected query results.');
-      }
-    })
-    .catch(error => this.setState({ timeDriftError: error.message }));
+        if (delta >= 30) {
+          throw new Error('Detected ' + delta + ' seconds time difference between your browser and the server. Prometheus relies on accurate time and time drift might cause unexpected query results.');
+        }
+      })
+      .catch(error => this.setState({ timeDriftError: error.message }));
```
juliusv: Hmm seems like here there are also a lot of unrelated formatting changes?

web/ui/react-app/src/utils/func.ts
```diff
@@ -0,0 +1 @@
+export const uuidGen = () => '_' + Math.random().toString(36).substr(2, 9);
```
juliusv: nit: missing newline at end of file

juliusv:
Somehow with this PR, the spacing between the expression input field and the Table/Graph tabs completely disappears. It's there on `master`, but I couldn't figure out yet what makes it appear/disappear.

web/ui/react-app/src/App.css
```diff
@@ -1,185 +1,191 @@
...
+  padding: 10px 0;
 }
 
 .panel {
-    margin-bottom: 20px;
+  margin-bottom: 20px;
 }
 
-.expression-input {
-    margin-bottom: 10px;
```
juliusv: Ah, this was what I was looking for - removing this makes the spacing between the expression input and the tab pane disappear.

boyskila: Yeah, will revert the file and I'll apply only checkbox CSS

Yesterday I spend some time experimenting with eslint end prettier and probably "autofix on save" option makes all those changes

juliusv: Yep, that's what I guessed :)

web/ui/react-app/src/Checkbox.tsx
```diff
@@ -0,0 +1,22 @@
...
+        id={`checkbox_${id}`}
+        placeholder="password placeholder" />
+      <Label style={{ userSelect: 'none' }} className="custom-control-label" for={`checkbox_${id}`}>
+        {children}
+      </Label>
+    </FormGroup>
+  )
+}
+
+export default memo(Checkbox)
```
juliusv: nit: missing newline :)

Hmm, something with the memo-ing is off. Toggling the checkbox causes the random id to change due to some re-render. I guess that's not what you intended, even though it probably doesn't do much harm?

boyskila: Actually I knew about that and I was thinking to make it a class component in order to preserve the id but as you notice, it's not going to be a problem.

web/ui/react-app/yarn.lock
```diff
@@ -927,6 +927,13 @@
   dependencies:
     "@fortawesome/fontawesome-common-types" "^0.2.25"
 
+"@fortawesome/free-regular-svg-icons@^5.7.1":
```
juliusv: Sorry, didn't explicitly call it out earlier, but the yarn.lock entry can now be removed too :)

juliusv:
Looks good now besides last yarn.lock nit!

juliusv:
@boyskila I was maybe a bit trigger-happy there, but would be surprised if the minimal last changes caused new test errors. In case they do, we'll see :)

web/ui/react-app/src/PanelList.tsx
boyskila: @juliusv https://github.com/prometheus/prometheus/pull/6211 - splitting autocomplete dropdown into groups

PR #6194, opened 2019-10-21
bump go version in go.mod

tariq1890:
Bumping go mod's go version to `1.13` since prometheus recommends the use of golang 1.13.


tariq1890:
@simonpasquier requesting your review.

Issue #5560, opened 2019-05-13
Prometheus goes to a OOM loop and states "unknown series references"


vaibhavkhurana2018:
## Bug Report
**What did you do?**

Running prometheus as a docker container on a stand-alone box.

**What did you expect to see?**
Prometheus running without any issues. 
**What did you see instead? Under which circumstances?**

Prometheus is going in an endless crashing state and getting the below error, once it restarts. 

```
level=warn ts=2019-05-13T09:02:15.270Z caller=head.go:454 component=tsdb msg="unknown series references" count=1085
```

**Environment**

Prometheus version:

2.9.2

Alertmanager version: N/A

Prometheus configuration file:

```
global:
  scrape_interval: 1m
  scrape_timeout: 10s
  evaluation_interval: 1m
scrape_configs:
- job_name: nodeexporter
  honor_timestamps: true
  scrape_interval: 1m
  scrape_timeout: 10s
  metrics_path: /metrics
  scheme: http
  ec2_sd_configs:
  - endpoint: ""
    region: ap-south-1
    refresh_interval: 1m
    port: 9100
    filters: []
  relabel_configs:
  - source_labels: [__meta_ec2_tag_monitoring]
    separator: ;
    regex: "true"
    replacement: $1
    action: keep
  - source_labels: [__meta_ec2_tag_Name]
    separator: ;
    regex: (.*)
    target_label: asg_group
    replacement: $1
    action: replace
- job_name: elasticexporter
  honor_timestamps: true
  scrape_interval: 1m
  scrape_timeout: 10s
  metrics_path: /_prometheus/metrics
  scheme: http
  ec2_sd_configs:
  - endpoint: ""
    region: ap-south-1
    refresh_interval: 1m
    port: 9200
    filters: []
  relabel_configs:
  - source_labels: [__meta_ec2_tag_elasticsearch]
    separator: ;
    regex: "true"
    replacement: $1
    action: keep
- job_name: prod-blue-traefik-external
  honor_timestamps: true
  scrape_interval: 1m
  scrape_timeout: 10s
  metrics_path: /metrics
  scheme: http
  static_configs:
  - targets:
    - <endpoint>
- job_name: prod-blue-traefik-internal
  honor_timestamps: true
  scrape_interval: 1m
  scrape_timeout: 10s
  metrics_path: /metrics
  scheme: http
  static_configs:
  - targets:
    - <endpoint>
- job_name: prod-blue-traefik-concierge
  honor_timestamps: true
[...]

PR #6195, opened 2019-10-21
Change bounds for WALSegmentSize flag value

afnanenayet:
This allows the flag sets `WALSegmentSize` to be negative, which allows the user to disable the WAL. I saw that in tsdb/db.go:60 there is a comment stating that setting the segment size to anything less than zero will actually disable the WAL, but in the main file, the flag is restricted to a value between 10 and 256 MB. I wanted to expose this setting through the flag because the WAL can use a significant amount of RAM and for some users I think it may be worth disabling the WAL to avoid the OOM crash cycle as described in this comment: https://github.com/prometheus/prometheus/issues/5560#issuecomment-537210222

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -655,8 +655,10 @@ func main() {
 		g.Add(
 			func() error {
 				level.Info(logger).Log("msg", "Starting TSDB ...")
-				if cfg.tsdb.WALSegmentSize > 256*1024*1024 {
-					return errors.New("flag 'storage.tsdb.wal-segment-size' must be less than 256MB")
+				if cfg.tsdb.WALSegmentSize > 0 {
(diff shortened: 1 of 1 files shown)


brian-brazil:
I don't think this is something we should expose, in fact with the current
code I don't think we need this flag at all.

afnanenayet:
I apologize because I think I may have been unclear and framed this as an addition rather than a bugfix (not trying to add a new flag or anything).

We can see here that the `WALSegmentSize` value already accepts negative numbers to disable the WAL, and I just wanted to update main so it accurately reflects the WAL settings since it is already documented in the WAL options and it seems like the command line flag doesn't accurately reflect that.

https://github.com/prometheus/prometheus/blob/5f1be2cf455409d3577ac9ddb43144ec2d39c90b/tsdb/db.go#L59-L62

brian-brazil:
It's not sane to run the Prometheus binary without a WAL, and if you're OOMing at startup you were going to OOM anyway.

afnanenayet:
Gotcha, makes sense. Closing

Issue #2855, opened 2017-06-16
FR: Add commenting to promQL


virusdave:
Feature request: Add some means of commenting or disabling portions of a query in promql.

This is especially useful when using the `/graph` playground to explore and compose queries.  Currently one almost needs to keep a separate text file or state somewhere to compose query snippets.  It would be very helpful if i could "comment out" portions of a query while trying to construct a query or debug one that's not working as expected.

virusdave:
Apparently this already exists: `#` will comment until end-of-line.

PR #6198, opened 2019-10-22
doc: Add documentation about comments in PromQL

slrtbtfs:

TIL that prometheus actually supports comments, while reading the parser code. This is  also mentioned  in #2855. 

This PR documents this.

However this feature isn't tested in  the parser tests, so it might make sense to add such tests before telling everybody about it.


brian-brazil:
Adding tests first would be good, and I'd put this down the end of the docs as it's quite rarely used.

brian-brazil:
Thanks!

PR #6200, opened 2019-10-23
Build result of tsdb should be ignored by git

kevinfeng:
Signed-off-by: Feng Kun <kevinkfeng@tencent.com>


.gitignore
```diff
@@ -6,6 +6,7 @@
 
 /prometheus
 /promtool
+/tsdb/tsdb
```
codesome: The tsdb binary is actually built at `/tsdb/cmd/tsdb/tsdb`

kevinfeng: I tried `make build` and it generates /tsdb/tsdb by promu:
```
➜  prometheus git:(master) ✗ make build
>> writing assets
# Un-setting GOOS and GOARCH here because the generated Go code is always the same,
# but the cached object code is incompatible between architectures and OSes (which
# breaks cross-building for different combinations on CI in the same container).
cd web/ui && GO111MODULE=on GOOS= GOARCH= go generate -x -v  -mod=vendor
doc.go
go run -mod=vendor assets_generate.go
writing assets_vfsdata.go
ui.go
>> building binaries
GO111MODULE=on /Users/kevinkfeng/code/go/bin/promu build --prefix /Users/kevinkfeng/code/go/src/github.com/prometheus/prometheus
 >   prometheus
 >   promtool
 >   tsdb
```
Is there something wrong?

codesome: Apparently yes. `make build_tsdb` seems to build it in `/tsdb/cmd/tsdb/tsdb`. Could you modify `build_tsdb` to also build as `/tsdb/tsdb` to be consistent (@krasi-georgiev sounds good?) and incorporate https://github.com/prometheus/prometheus/pull/6204 in it?

simonpasquier:
Note that the `tsdb` binary should go away, see https://github.com/prometheus/prometheus/pull/6088

krasi-georgiev:
yep closing this one in favor of, if needed please add your review in that PR
https://github.com/prometheus/prometheus/pull/6088

thanks

codesome:
Oops, was not in sync with ongoing stuff, sorry!

PR #6205, opened 2019-10-23
promql: Refactor abstract syntax tree traversion

slrtbtfs:
This PR adds a `Childs()` function to the AST `Node` interface. This makes traversing the syntax tree less complex and prevents code duplication.

Note that these changes are implemented on top of #6061, so it doesn't make sense to merge it until the other PR is merged. Also only the last three commits are actually new here.


slrtbtfs:
Closed in favour of #6362

PR #6206, opened 2019-10-23
WIP: React UI Linting rules

CSTDev:
Mostly default rules, a lot of white space changes, some commas, semi-colons and single quotes. There's currently a lot of warnings too mostly about missing return types on some React lifecycle functions that I'll look to address.

Initial thoughts?
@juliusv @boyskila ..anyone else who's interested

Will look into CI steps


juliusv:
Cool! First reactions:

From the changes I see, I love:
- enforcing quote consistency
- enforcing semicolons
- enforcing consistent spacing around operators and braces/brackets/parens within a line
- enforcing const/let/... as applicable
- consistent indentation of existing lines
- removing unnecessary parens

I'm unsure about the larger reformattings that enforce multiple lines, etc. - Maybe a bit aggressive for the beginning?

Regarding trailing commas for last elements in arrays and objects, I'd prefer reversing the rule and always having them. (I'm a huge fan of commas after the last element because it makes both editing and diffing better)

boyskila:
Cool! Looking forward to this to be integrated. It's really hard to follow all the best practices without such a tool.

CSTDev:
@juliusv  How about trailing commas after imports? 
```
import { Button, InputGroup, InputGroupAddon, InputGroupText, Input, } from 'reactstrap';
```

boyskila:
> @juliusv How about trailing commas after imports?
> 
> ```
> import { Button, InputGroup, InputGroupAddon, InputGroupText, Input, } from 'reactstrap';
> ```

Is there a rule which added it only when is multiline import? I think is useful since adding a comma manually will result as a change in the diff

CSTDev:
> > @juliusv How about trailing commas after imports?
> > ```
> > import { Button, InputGroup, InputGroupAddon, InputGroupText, Input, } from 'reactstrap';
> > ```
> 
> Is there a rule which added it only when is multiline import? I think is useful since adding a comma manually will result as a change in the diff

Yes there is. Will try that one.

juliusv:
@CSTDev Yeah, those commas only make sense to me for multi-line imports / objects / arrays.

web/ui/react-app/.eslintrc.json
```diff
@@ -0,0 +1,24 @@
...
+      "comma-dangle": ["error", {
+        "arrays": "always-multiline",
+        "objects": "always-multiline",
+        "imports": "always-multiline"
+    }]
+    },
+    "plugins": [
+      "prettier"
+    ]
+  }
```
juliusv: missing newline file termination here (not sure if we can have a linter setting for the linter config file itself :))

web/ui/react-app/src/DataTable.tsx
```diff
@@ -59,45 +64,69 @@ class DataTable extends PureComponent<QueryResult> {
...
     let limited = false;
-    switch(data.resultType) {
+    switch (data.resultType) {
       case 'vector':
-        rows = (this.limitSeries(data.result) as InstantSample[])
-          .map((s: InstantSample, index: number): ReactNode => {
-            return <tr key={index}><td><SeriesName labels={s.metric} format={false}/></td><td>{s.value[1]}</td></tr>;
-          });
+        rows = (this.limitSeries(data.result) as InstantSample[]).map(
+          (s: InstantSample, index: number): ReactNode => {
```
juliusv: Do we want to enforce this `.map()` to happen at the end of the line rather at the beginning of a new one where it's more visible? I don't care much either way, just curious if an explicit rules causes this in particular, or it's the side effect of a more general reformatting rule?

web/ui/react-app/src/DataTable.tsx
```diff
@@ -59,45 +64,69 @@ class DataTable extends PureComponent<QueryResult> {
...
+        );
         limited = rows.length !== data.result.length;
         break;
       case 'matrix':
-        rows = (this.limitSeries(data.result) as RangeSamples[])
-          .map((s, index) => {
-            const valueText = s.values.map((v) => {
+        rows = (this.limitSeries(data.result) as RangeSamples[]).map((s, index) => {
+          const valueText = s.values
+            .map(v => {
```
juliusv: Ah interesting, here it did the opposite and put `.map()` at the beginning of the line where it was previously at the end...

web/ui/react-app/src/DataTable.tsx
```diff
@@ -59,45 +64,69 @@ class DataTable extends PureComponent<QueryResult> {
 
     let rows: ReactNode[] = [];
     let limited = false;
-    switch(data.resultType) {
+    switch (data.resultType) {
       case 'vector':
-        rows = (this.limitSeries(data.result) as InstantSample[])
-          .map((s: InstantSample, index: number): ReactNode => {
-            return <tr key={index}><td><SeriesName labels={s.metric} format={false}/></td><td>{s.value[1]}</td></tr>;
```
juliusv: Since this line length was below 125 chars, I guess there's also an explicit rule that disallows putting any nested JSX tags on the same line anywhere? Just kinda want to understand the implications... I guess this is ok though.

web/ui/react-app/src/Panel.tsx
```diff
@@ -1,16 +1,6 @@
...
-  Button,
-  Col,
-  Nav,
-  NavItem,
-  NavLink,
-  Row,
-  TabContent,
-  TabPane,
-} from 'reactstrap';
+import { Alert, Button, Col, Nav, NavItem, NavLink, Row, TabContent, TabPane } from 'reactstrap';
```
juliusv: Not sure if we need this multi- to single-line reformatting of import statements. Maybe better to leave this decision to the developer depending on what they find more reasonable on a case-by-case basis? If it's hard to turn this off individually though, it's ok too.

juliusv:
What I see now looks mostly good besides some comments / questions. Some things might be a bit aggressive, but we can always redecide later if we get too annoyed by some rules, I guess.

We should probably give @boyskila a chance to get https://github.com/prometheus/prometheus/pull/6211 merged first, as it's easier to re-autogenerate this PR here one ontop of that one than the other way around.

web/ui/react-app/src/DataTable.tsx
CSTDev: I have no idea why it's doing this I've tried to make them look the same, but it keeps doing one one way and the other another 😕 Seems to be a prettier thing again....

web/ui/react-app/src/Panel.tsx
CSTDev: Agreed I'd rather leave it up to the developer.

CSTDev: After looking into it seems that you can't tell it not too, without putting a `//prettier-ignore` comment in front of it which is a bit nasty.

juliusv: @CSTDev It's probably due to their stance on configurability (https://prettier.io/docs/en/option-philosophy.html). I guess it's fine then, we can live with it.

Makefile
```diff
@@ -47,8 +47,13 @@ assets: $(REACT_APP_OUTPUT_DIR)
 	cd web/ui && GO111MODULE=$(GO111MODULE) GOOS= GOARCH= $(GO) generate -x -v $(GOOPTS)
 	@$(GOFMT) -w ./web/ui
 
+.PHONY: react-app-lint
+react-app-lint:
```
juliusv: Could you add a `react-app-lint-fix` target as well that runs `yarn lint`? I think running it manually for now via `make react-app-lint-fix` would be fine, so other targets don't have to depend on it.

web/ui/react-app/.eslintrc.json
```diff
@@ -0,0 +1,23 @@
...
+        "arrays": "always-multiline",
+        "objects": "always-multiline",
+        "imports": "always-multiline"
+      }]
+    },
+    "plugins": [
+      "prettier"
+    ]
+  }
+
```
juliusv: Now the file ends with a space, not a newline :P

juliusv:
@CSTDev any news on this one? Would be good to get this thing in soon.

Makefile
CSTDev: Will add

web/ui/react-app/.eslintrc.json
CSTDev: Whoops, have removed it, I tried to get the linter to check it but seems to be a bit of a PITA to get it to do its own config

CSTDev:
@juliusv Should be there now, I've pulled in the current master and linted it. Hopefully all the checks pass.

juliusv:
@CSTDev Looks good now, thanks! :)

PR #6207, opened 2019-10-23
disable github actions

krasi-georgiev:
github actions runs the comment validation for every comment which in
turn downloads and runs a new container on every execution.

This is wasting many resources and also there is a big delay when
starting a test so need to think for a different design and for now will
disable GH actions.

An addition to this people are receiving an email for each posted
comment that fails the validation

Signed-off-by: Krasi Georgiev <8903888+krasi-georgiev@users.noreply.github.com>


krasi-georgiev:
cc @geekodour

Issue #6160, opened 2019-10-17
React UI: Implement more sophisticated autocomplete


juliusv:
It would be great to have more sophisticated expression field autocompletion in the [new React UI](https://github.com/prometheus/prometheus/pull/5694).

Currently it only autocompletes metric names, and only when the expression field doesn't contain any other sub-expressions yet.

Things that would be nice to autocomplete:

- metric names anywhere within an expression
- label names
- label values
- function names
- etc.

For autocomplete functionality not to annoy users, it needs to be as highly performant, correct, and unobtrusive as possible. Grafana does many things right here already, but they also have a few really annoying bugs, like inserting closing parentheses in incorrect locations of an expression.

Currently @slrtbtfs has indicated interest in building a language-server-based autocomplete implementation.

slrtbtfs:
I'm currently in the process of building that language server. It lives at <https://github.com/slrtbtfs/promql-lsp>.

The features you named are all intended to land there. Support for completing metric and function names will probably be implemented in the coming two weeks.

Other features include:
* Showing syntax error messages while typing (done)
* Showing documentation and type signatures on hover (in progress)
* PromQL Syntax highlighting (not part of the language server, but exists already)

The part that has to be done on the React UI side to support the language server would be to provide a language client. A good candidate for that seems to be the monaco editor which is also used by vscode and is surprisingly fast and responsive when included in a web UI (See <https://microsoft.github.io/monaco-editor/playground.html>).

cc @squat

juliusv:
@slrtbtfs Wonderful! Really looking forward to this :) Do you know if you'll have capacity to work on the React UI integration as well?

slrtbtfs:
It would make sense to do so. Especially as there probably is a lot of overlap with the PromQL vscode extension (<https://github.com/slrtbtfs/vscode-prometheus>).

I'll discuss this with my managers.

juliusv:
@slrtbtfs One point to consider: Monaco is not exactly small. Even just the minified [`editor.main.js`](https://microsoft.github.io/monaco-editor/node_modules/monaco-editor/min/vs/editor/editor.main.js) is >2MB large. That doesn't make me immediately rule it out, but if we could do what we need with a significantly smaller download size, that would be much better of course.

slrtbtfs:
[...]

PR #6211, opened 2019-10-24
split autocomplete dropdown in to groups

boyskila:

### DIFF
--- a/web/ui/react-app/src/App.css
+++ b/web/ui/react-app/src/App.css
@@ -75,7 +75,6 @@ button.execute-btn {
     top: 100%;
     left: 56px;
     float: left;
-    padding: .5rem 1px .5rem 1px;
     margin: -5px;
     list-style: none;
 }
@@ -88,6 +87,14 @@ button.execute-btn {
     background-color: transparent;
     border: 0;
     display: block;
+}
+
+.autosuggest-dropdown .card-header {
+    background: rgba(240, 230, 140, 0.4);
+    height: 30px;
+    display: flex;
+    align-items: center;
+    text-transform: uppercase;
 }
 
 .graph-controls, .table-controls {
--- a/web/ui/react-app/src/Checkbox.tsx
+++ b/web/ui/react-app/src/Checkbox.tsx
@@ -1,18 +1,15 @@
-import React, { FC, HTMLProps, memo } from 'react';
-import { FormGroup, Label, Input } from 'reactstrap';
-import { uuidGen } from './utils/func';
+import React, { FC, memo, CSSProperties } from 'react';
+import { FormGroup, Label, Input, InputProps } from 'reactstrap';
 
-const Checkbox: FC<HTMLProps<HTMLSpanElement>> = ({ children, onChange, style }) => {
-  const id = uuidGen();
+interface CheckboxProps extends InputProps {
+  wrapperStyles?: CSSProperties;
+}
+
+const Checkbox: FC<CheckboxProps> = ({ children, wrapperStyles, id, ...rest }) => {
   return (
-    <FormGroup className="custom-control custom-checkbox" style={style}>
-      <Input
-        onChange={onChange}
-        type="checkbox"
-        className="custom-control-input"
-        id={`checkbox_${id}`}
-        placeholder="password placeholder" />
-      <Label style={{ userSelect: 'none' }} className="custom-control-label" for={`checkbox_${id}`}>
+    <FormGroup className="custom-control custom-checkbox" style={wrapperStyles}>
+      <Input {...rest} id={id} type="checkbox" className="custom-control-input" />
+      <Label style={{ userSelect: 'none' }} className="custom-control-label" for={id}>
         {children}
       </Label>
     </FormGroup>
--- a/web/ui/react-app/src/ExpressionInput.tsx
+++ b/web/ui/react-app/src/ExpressionInput.tsx
@@ -32,7 +32,6 @@ interface ExpressionInputState {
 }
 
 class ExpressionInput extends Component<ExpressionInputProps, ExpressionInputState> {
-  private prevNoMatchValue: string | null = null;
   private exprInputRef = React.createRef<HTMLInputElement>();
 
   constructor(props: ExpressionInputProps) {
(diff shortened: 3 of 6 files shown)


web/ui/react-app/src/Checkbox.tsx
```diff
@@ -1,18 +1,15 @@
-import React, { FC, HTMLProps, memo } from 'react';
-import { FormGroup, Label, Input } from 'reactstrap';
-import { uuidGen } from './utils/func';
```
juliusv: You can remove the whole `./utils/func` dir then too?

web/ui/react-app/src/Checkbox.tsx
```diff
@@ -1,18 +1,15 @@
...
-    <FormGroup className="custom-control custom-checkbox" style={style}>
-      <Input
-        onChange={onChange}
-        type="checkbox"
-        className="custom-control-input"
-        id={`checkbox_${id}`}
-        placeholder="password placeholder" />
-      <Label style={{ userSelect: 'none' }} className="custom-control-label" for={`checkbox_${id}`}>
+    <FormGroup className="custom-control custom-checkbox" style={wrapperStyles}>
+      <Input {...rest} id={id} type="checkbox" className="custom-control-input" />
```
juliusv: Yeah, good call to just pass in the `id` and have it be a constant in the caller, as there's just one of these anyway on the page.

web/ui/react-app/src/PanelList.tsx
```diff
@@ -6,19 +6,23 @@ import Panel, { PanelOptions, PanelDefaultOptions } from './Panel';
 import { decodePanelOptionsFromQueryString, encodePanelOptionsToQueryString } from './utils/urlParams';
 import Checkbox from './Checkbox';
 
+export type MetricGroup = { title: string; items: string[] };
```
juliusv: `MetricGroup` seems like a misnomer since queries are not (always) metrics / metric names. How about `CompletionGroup` (here and in comparable variable names)?

However, I wonder if we really have to abstract it this far, since I think there will only be two autocompletion groups for the foreseeable future (and I'm for adding abstraction only when it becomes necessary).

If we kept it simpler and just had separate top-level `pastQueries` and `metricNames` properties in the `PanelListState` (i.e. not grouped under `metricGroups`), then it would also simplify the code that reacts to new queries being saved in the history - it wouldn't have to ever have to refer to the `metricNames` at all, since those only get set once when the page loads. It would also have to reach fewer levels deep (`this.state.metricGroups.queries.items` -> `this.state.pastQueries`). And then maybe even leave the specific string title for each group up to the expression input.

What do you think?

web/ui/react-app/src/ExpressionInput.tsx
```diff
@@ -179,7 +191,10 @@ class ExpressionInput extends Component<ExpressionInputProps, ExpressionInputSta
                 </Button>
               </InputGroupAddon>
             </InputGroup>
-            {downshift.isOpen && this.renderAutosuggest(downshift)}
+            {downshift.isOpen &&
+              <div {...downshift.getMenuProps()} className="autosuggest-dropdown">
+                { this.creatAutocompleteSection(downshift) }
```
juliusv: Not sure what causes this, but with this PR there's a new bug where the dropdown opens very briefly and then closes immediately on any new input, even if the current string doesn't match anything in the history. E.g. type `zzzzzzzzz` and you'll see the (empty) dropdown briefly flash open after every typed letter.

web/ui/react-app/src/PanelList.tsx
```diff
@@ -88,20 +94,24 @@ class PanelList extends Component<any, PanelListState> {
...
+    let queryItems: string[] = [];
     if (this.isHistoryEnabled()) {
       const historyItems = this.getHistoryItems();
       const { length } = historyItems;
-      this.setState({
-        metricNames: [...historyItems.slice(length - 50, length), ...this.initialMetricNames],
-      });
-    } else {
-      this.setState({ metricNames: this.initialMetricNames });
+      queryItems = historyItems.slice(length - 50, length)
```
juliusv: Currently we don't seem to be keeping a correct rolling window of the past N (e.g. 50) queries here and in the local storage. The local storage in `handleQueryHistory()` currently keeps everything without limit, but at the same time, the autocomplete doesn't keep the last 50 correctly somehow.

Simple example: Try executing as queries the numbers 1 ... 60. You'll see all 60 of them in the local storage, but if you type `6`, you will not see an autocomplete entry for `60`.

I think what we want:
- always limit the local storage to the last N queries
- of course still don't store or show duplicate queries twice (currently correct)
- if a past query gets executed again, update it to be the latest one (so it gets kept in the history longer)

web/ui/react-app/src/App.css
```diff
@@ -90,6 +89,14 @@ button.execute-btn {
     display: block;
 }
 
+.autosuggest-dropdown .card-header {
```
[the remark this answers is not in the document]
boyskila: I'm more interested in functional part and implementing mockups but not creating them :) So thats why I trusted you.
Play as much as you want of course :)

web/ui/react-app/src/PanelList.tsx
boyskila: I think is more flexible now. If you decided to add another group someday(who knows) it's gonna be easier than if you have separate properties. Eventually only the PanelsList will be changed. The only downside I can see of this approach is the long path to the items.
```this.state.metricGroups.queries.items VS this.state.pastQueries```

Comment (juliusv) [reply web/ui/react-app/src/PanelList.tsx]: I don't feel too strongly about it in this case, just generally over the years of maintaining code I've become more and more of a fan of YAGNI (https://www.martinfowler.com/bliki/Yagni.html), KISS, and avoiding premature abstraction. IMO it makes code easier to read/understand/keep correct before an abstraction is needed.

My guess is that we will likely not add a third category before we change to completely a different solution that can do context-sensitive autocompletion (like auto-completing label names in `by` clauses, auto-completing function names, etc.). And if we do, we can still change things then.

That's just to outline my general thinking about "this will be more flexible in the future" questions, though again I don't feel too strongly about this particular case. If you'd rather like to keep it like it is, it's ok too.

File: web/ui/react-app/src/ExpressionInput.tsx (hunk 4 above)
Comment (boyskila) [reply web/ui/react-app/src/ExpressionInput.tsx]: here is the fix https://github.com/prometheus/prometheus/pull/6211/commits/c9d2220113ef16d1d48b64387da78ae9f488d587

File: web/ui/react-app/src/PanelList.tsx (hunk 5 above)
Comment (boyskila) [reply web/ui/react-app/src/PanelList.tsx]: image
@juliusv Everything seems to be fine. Did you refresh the page after setting the items in the storage?

Comment (juliusv) [reply web/ui/react-app/src/PanelList.tsx]: So the first issue is that local storage seems unbounded:

- Delete `history` and `enable-query-history` local storage items via Chrome's inspector (Application->Local Storage)
- Enter >50 different queries.
- Watch the local storage entry grow to >50 entries. This should be limited to not grow infinitely.

The other issue with suggestions now showing up I get with:

- Delete both local storage items again, enable query history.
- Enter and execute `1` ... `10` as 10 different queries.
- Enter `1` and get no autocomplete results (neither `1` nor `10`). The numbers `2` ... `9` do give me autocomplete results in this case btw.
- When I reload the page, then I also get `1` and `10` as autocomplete results, but that shouldn't be needed.

File: web/ui/react-app/src/ExpressionInput.tsx [hunk 7]
```diff
@@ -67,54 +71,62 @@ class ExpressionInput extends Component<ExpressionInputProps, ExpressionInputSta
...
+  getSearchMatches = (input: string, expressions: string[]) => {
+    return fuzzy.filter(input.replace(/ /g, ''), expressions, {
+      pre: "<strong>",
+      post: "</strong>",
+    });
+  }
+
+  creatAutocompleteSection = (downshift: ControllerStateAndHelpers<any>) => {
+    const { inputValue, closeMenu, highlightedIndex } = downshift;
     if (!inputValue || (this.prevNoMatchValue && inputValue.includes(this.prevNoMatchValue))) {
```
boyskila: @juliusv what is the purpose of ```(this.prevNoMatchValue && inputValue.includes(this.prevNoMatchValue))```
Seems like everything is fine when I remove it

juliusv: I added this hack back then to skip the fuzzy-matching calculation, which can create noticeable delays while typing when there are a lot of metrics to match against.

When we already know that the previous input didn't have a match entry and the new input value is a superset of the old one, it will definitely also not match anything.

juliusv:
Cool, the two points I mentioned seem fixed now, but I found a new bug where selecting a metric entry from the autocomplete closes the dropdown, but doesn't set the input. Steps to reproduce:

1. Run Prometheus with the following config:
```yml
global:
  scrape_interval: 5s

scrape_configs:
  - job_name: "node"
    static_configs:
      - targets: ['demo.robustperception.io:9100']
```
2. Open UI with blank local storage
3. Enable query history
4. Type `node_arp_entries` and execute the query.
5. Type `node_arp`, navigate down to the `node_arp_entries` autocomplete entry, and press `Return`.
6. The dropdown closes and the input field still says `node_arp`.

The problem seems to be that the entry is also in the query history. Selecting another autocompleted metric name which wasn't executed (and saved) before works.

I think in general, with the new section split, pure metric names that have been queried should probably just show up in both the history and the metric names sections.

boyskila:
> Cool, the two points I mentioned seem fixed now, but I found a new bug where selecting a metric entry from the autocomplete closes the dropdown, but doesn't set the input. Steps to reproduce:
> 
> 1. Run Prometheus with the following config:
> 
> ```yaml
> global:
>   scrape_interval: 5s
> 
> scrape_configs:
>   - job_name: "node"
>     static_configs:
>       - targets: ['demo.robustperception.io:9100']
> ```
> 
> 1. Open UI with blank local storage
> 2. Enable query history
> 3. Type `node_arp_entries` and execute the query.
> 4. Type `node_arp`, navigate down to the `node_arp_entries` autocomplete entry, and press `Return`.
> 5. The dropdown closes and the input field still says `node_arp`.
> 
> The problem seems to be that the entry is also in the query history. Selecting another autocompleted metric name which wasn't executed (and saved) before works.
> 
> I think in general, with the new section split, pure metric names that have been queried should probably just show up in both the history and the metric names sections.

Not sure why but Downshift not trigger onChange every time when item is selected. Fortunately, onSelect does it, so here is the fix https://github.com/prometheus/prometheus/pull/6211/commits/c1b619a63d0efa4897086c90e1855c39066a3028

slrtbtfs:
@boyskila There are plans to let use a language server for things like auto complete in the React UI. See #6160 . So it might make sense to implement these things there.

juliusv:
@slrtbtfs Yeah, we're purposefully not taking it much further here than what the old UI already does. I assume though that the LS integration will still take some time? And https://github.com/prometheus/prometheus/issues/6160#issuecomment-544239224 is a big question IMO.

web/ui/react-app/src/ExpressionInput.tsx
```diff
@@ -67,62 +70,69 @@ class ExpressionInput extends Component<ExpressionInputProps, ExpressionInputSta
...
+    return fuzzy.filter(input.replace(/ /g, ''), expressions, {
       pre: "<strong>",
       post: "</strong>",
     });
+  }
 
-    if (matches.length === 0) {
-      this.prevNoMatchValue = inputValue;
-      setTimeout(downshift.closeMenu);
+  creatAutocompleteSection = (downshift: ControllerStateAndHelpers<any>) => {
```
juliusv: `creat` -> `create`

web/ui/react-app/src/PanelList.tsx
juliusv: Assuming you are keeping the structure as it is for now, could you still rename it to `CompletionGroup`, `AutocompleteGroup`, `SearchGroup`, or something like that, since queries are not metrics?

boyskila:
I'm going to change to what you propose... I just don't have enough time these days :)

juliusv:
@boyskila Thanks, seems to look good now :) Is it ready to merge from your side?

boyskila:
Yeah, I think so. Definitely looks better now, thanks!

juliusv:
Cool! I'll merge and work on the styling a bit then :)

PR #6215, opened 2019-10-24
promql: Eliminate dead code

slrtbtfs:
This PR removes some dead code from the parser.

Reason why the code is dead:
peek() already ensures to not return a ItemComment so checking for this again
is redundant.


juliusv:
I only capitalized it so it's visible from outside the package :) But LGTM.

PR #6217, opened 2019-10-24
Fix Windows support for custom-sd adapter

### LINKED ISSUE #6208: custom-sd writeOuput fails on Windows due to filehandle semantics
It is not possible to run the `custom-sd` adapter on Windows:

```text
err="rename C:\\Users\\bgagnon\\AppData\\Local\\Temp\\sd-adapter518011287 custom_sd.json: 
The process cannot access the file because it is being used by another process."
```

This is because of the `defer` on the tmpfile handle in the `writeOutput()` function:

```go
 	defer tmpfile.Close()

	_, err = tmpfile.Write(b)
	if err != nil {
		return err
	}

	err = os.Rename(tmpfile.Name(), a.output)
```
On Windows, the file must be closed _before_ it is moved.

I understand support for Windows is not a priority, but in this case the fix looks simple enough. Developing simple things like a custom_sd is convenient to do with the native tooling of a workstation.

I think this problem was alluded to in #6079

bgagnon:
Fixes #6208, with test.

### DIFF
--- a/documentation/examples/custom-sd/adapter/adapter.go
+++ b/documentation/examples/custom-sd/adapter/adapter.go
@@ -128,6 +128,9 @@ func (a *Adapter) writeOutput() error {
 		return err
 	}
 
+	// Close the file immediately for platforms (eg. Windows) that cannot move
+	// a file while a process is holding a file handle
+	tmpfile.Close()
 	err = os.Rename(tmpfile.Name(), a.output)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


documentation/examples/custom-sd/adapter/adapter.go
```diff
@@ -128,6 +128,9 @@ func (a *Adapter) writeOutput() error {
 		return err
 	}
 
+	// Close the file immediately for platforms (eg. Windows) that cannot move
+	// a file while a process is holding a file handle
```
simonpasquier: (nit) missing dot

bgagnon:
@simonpasquier is anything else needed for this PR?

simonpasquier:
Thanks!

PR #6219, opened 2019-10-25
tsdb: remove unnecessary WALSegmentSize check

huanggze:
WALSegmentSize is always non-negative. The outer if statement seems unnecessary.

Signed-off-by: huanggze <loganhuang@yunify.com>


huanggze:
I didn't find the entry to disable wal or the way to make WALSegmentSize negative. Is this a feature in plan? If I didn't misunderstand, I think we are good to remove the check for negative value.

huanggze:
That's true, thanks

PR #6224, opened 2019-10-25
[ENHANCEMENT] scrape : Scrape fail retry

liguozhong:
Background：
When we encountered the network jitter during the pressure test, the scrape failed, which eventually caused the graffana's dashboard to be completely unavailable. I look forward to providing scrape's retry mechanism to deal with the short-term unavailability of network devices.

image
up==0
image
PR config
image
debug 
image


brian-brazil:
Thanks for your PR. We purposefully do not retry, as putting more load on
an overload system is unwise. In general you should be gracefully be able
to handle a failed scrape.

liguozhong:
Thank you for your review. 
In our team's IDC, there are hundreds of thousands of servers connected via network devices. Sporadic unavailability of network devices is a common phenomenon, but our business systems can avoid request failures by retrying. We very much hope that the monitoring system Prometheus can also have the ability to tolerate errors like retrying like a business system.

liguozhong:
The cpu of the collected target is very low, but the network router occasionally has problems causing scrape to fail.

Please help me solve this problem, I am working full-time on prometheus in a company, if it is confirmed that prometheus can not be used in our production environment monitoring. Then maybe our company's prometheus will have to go offline.

liguozhong:
image

We use prometheus to monitor istio envoy. The failure of scrape makes our business development judge the prometheus dashboard is not reliable. This problem is very troublesome for us.

liguozhong:
image
The problem of scrape failure happened again in the early hours of the morning.

brian-brazil:
Retries were explicitly removed many years ago, there's no plans to re-add them. If you're looking for 100% accuracy, then Prometheus is probably not for you.

liguozhong:
This is really a bad news for us, but thank you very much for your guidance.

roidelapluie:
I encourage you to put a proxy between Prometheus and the target if you really need this. But you have to be aware of the implications.

liguozhong:
> I encourage you to put a proxy between Prometheus and the target if you really need this. But you have to be aware of the implications.

Thank you very much for your suggestion. The logic of the scrape module retrying 3 times has been running stably for 3 weeks. Although it is very simple to modify, it makes our prometheus server a reliable and reliable monitoring platform. 
Our business side is currently very satisfied with prometheus and is using this prometheus to participate in Taobao's 11.11 global promotion festival.

brian-brazil:
We looked at this in the bug scrub, and this does not make sense for the reasons given above so we're closing it.

PR #6225, opened 2019-10-26
React UI: Simplify query history setting code

juliusv:
The metric names only get loaded once initially, so there is no reason
to mix them up with the handling of ongoing query history.


juliusv:
@boyskila small simplification PR

boyskila:
Yes, it's better now since is just a setState call

juliusv:
I take that as a :+1: :)

PR #6227, opened 2019-10-26
promtool: Add check for expression validation

weeezes:
Option to validate expression syntax through promtool, so a running Prometheus isn't required for syntax checking.

### DIFF
--- a/cmd/promtool/main.go
+++ b/cmd/promtool/main.go
@@ -40,6 +40,7 @@ import (
 
 	"github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/pkg/rulefmt"
+	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/util/promlint"
 )
 
@@ -63,6 +64,9 @@ func main() {
 	).Required().ExistingFiles()
 
 	checkMetricsCmd := checkCmd.Command("metrics", checkMetricsUsage)
+
+	checkExpressionCmd := checkCmd.Command("expression", "Check validity of PromQL expression")
+	checkExpr := checkExpressionCmd.Arg("expr", "PromQL query expression.").Required().String()
 
 	queryCmd := app.Command("query", "Run query against a Prometheus server.")
 	queryCmdFmt := queryCmd.Flag("format", "Output format of the query.").Short('o').Default("promql").Enum("promql", "json")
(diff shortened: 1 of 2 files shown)


cmd/promtool/main.go
```diff
@@ -377,6 +384,18 @@ func CheckMetrics() int {
 	return 0
 }
 
+// CheckExpression validates if the input expression is a valid PromQL expression
+func CheckExpression(expr string) int {
```
bwplotka: I know we would break consistency, but can we just return error here? This function is super basic. This will allow us to unit test it without this 
```
defer func() { os.Stderr = stderr }()
os.Stderr = os.NewFile(0, os.DevNull)
```

But I guess this would mean changing interface for all commands here..

cmd/promtool/main_test.go
```diff
@@ -78,3 +79,19 @@ func mockServer(code int, body string) (*httptest.Server, func() *http.Request)
 	}
 	return server, f
 }
+
+func TestCheckExpression(t *testing.T) {
+	// Temporarily move error output from CheckExpression calls
+	// to /dev/null, only return value is relevant
+	stderr := os.Stderr
+	defer func() { os.Stderr = stderr }()
```
bwplotka: Alternatively (to the comment above) I would just remove this. I think it's ok to log something on unit test.

cmd/promtool/main.go
weeezes: I could have a go at it at some point later, to refactor all of it to return an error instead of just an integer! I can report an issue to track it if that helps :). Doing it in the same PR would either make this one a lot bigger (if changing all of them) or introduce inconsistency (if just changing this single one) :sweat_smile: .

cmd/promtool/main_test.go
weeezes: It makes the test output quite confusing when testing that the error case is handled properly:

```
Error while validating: parse error at char 18: unexpected end of input inside braces
PASS
ok  	github.com/prometheus/prometheus/cmd/promtool	0.020s
```

Can ofc remove the lines and allow it, but to me that looks quite dirty :smile: . Let me know if that's still OK and I'll make it log :).

brian-brazil:
@simonpasquier Is there anything blocking merging this?

Can you fix the conflicts?

roidelapluie:

You can now do `promtool promql format` to achieve this, so we are closing this pull request.

Issue #6163, opened 2019-10-17
React UI: Ensure compliance with external URL and route prefix settings


juliusv:
We need to ensure that the [new React UI](https://github.com/prometheus/prometheus/pull/5694) works correctly with different values set for the `--web.external-url` and `--web.route-prefix` Prometheus command-line flags.

juliusv:
To elaborate on the problem here: the entire React UI (using create-react-app) is built and bundled expecting a particular path to be served on, then the resulting bundle is compiled into the Prometheus Go binary. But Prometheus needs to be able to dynamically change the path that the React app is expecting to be served on.

E.g. if someone sets the `--web.route-prefix=/my/prefix` flag, then the React app (and all legacy assets/endpoints) should be served under `/my/prefix/new` instead of `/new`. I'm not sure how to dynamically inject the expected serving path into the finished + compiled-in React app bundle.

juliusv:
One option: have Prometheus dynamically inject something like `<script>PATH_PREFIX = "/my/prefix";</script>` into the React bundle's cryptic `index.html` file and then refer to that global variable from the TypeScript files (needs a `declare var PATH_PREFIX: any;` to make TypeScript understand that this global variable exists).

juliusv:
On this.

roidelapluie:
@juliusv do you know if that is in 2.14.0?

juliusv:
@roidelapluie Should be! But if you found a bug, let me know :)

roidelapluie:
Yes that is simply not working

roidelapluie:
It does work only when I do add a trailing slash. /new -> NOK -- /new/ -> OK

juliusv:
Hmm, `http://localhost:9090/foo/bar/new` works totally fine for me (with `--web.external-url=http://localhost:9090/foo/bar`). Or did you also set `--web.route-prefix`?

roidelapluie:
Yes we have both

roidelapluie:
--web.external-url=https://XXXX/prometheus/XXXX2/ --web.route-prefix=/

juliusv:
Yeah, can reproduce. I did:

```
./prometheus --web.external-url=http://localhost:9090/foo/bar --web.route-prefix=/
```

With an nginx config of:

```
http {
 server {
   listen 0.0.0.0:19090;
   location /foo/bar/ {
     proxy_pass http://localhost:9090/;
   }
 }
}
events {
}
```

Two separate issues, it seems:

[...]

PR #6229, opened 2019-10-26
React UI: Serve React UI under /new

juliusv:
This makes React UI URLs look nicer than the previous
/static/graph-new/app.html, but internally still serves all React UI
files from the compiled-in static assets directory.

Also, to allow future usage of the React / Reach router, we need to
serve the main React app's index.html on certain sub-paths that
correspond to current Prometheus's UI pages, instead of trying to serve
actual files that match the provided path name.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,5 +18,5 @@ benchmark.txt
 /documentation/examples/remote_storage/example_write_adapter/example_writer_adapter
 
 npm_licenses.tar.bz2
-/web/ui/static/graph-new
+/web/ui/static/react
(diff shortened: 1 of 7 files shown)


juliusv:
@simonpasquier 

FYI @boyskila

juliusv:
@SuperQ changed the base path from `/react` to `/new` and renamed the PR + commit.

juliusv:
FWIW we're going to have a fun time with https://github.com/prometheus/prometheus/issues/6163

PR #6230, opened 2019-10-27
Decode WAL in Separate Goroutine

csmarchbanks:
This PR moves the decoding of WAL records into a separate goroutine which passes the decoded results back using a channel. When looking at startup performance in a handful of Prometheus instances, startup time is reasonably evenly split between decoding records and processing the result so this has the potential to reduce WAL replay time by up to 50%.

Benchmarks show a reasonable performance improvement.

WAL compression off:
```
benchmark                                                                  old ns/op     new ns/op     delta
BenchmarkLoadWAL/batches=10,seriesPerBatch=100,samplesPerSeries=7200-8     442345548     390425328     -11.74%
BenchmarkLoadWAL/batches=10,seriesPerBatch=10000,samplesPerSeries=50-8     896853353     852906578     -4.90%
BenchmarkLoadWAL/batches=10,seriesPerBatch=1000,samplesPerSeries=480-8     303195989     275009282     -9.30%
```
WAL compression on:
```
benchmark                                                                  old ns/op     new ns/op     delta
BenchmarkLoadWAL/batches=10,seriesPerBatch=100,samplesPerSeries=7200-8     475893407     439551460     -7.64%
BenchmarkLoadWAL/batches=10,seriesPerBatch=10000,samplesPerSeries=50-8     997983033     852942115     -14.53%
BenchmarkLoadWAL/batches=10,seriesPerBatch=1000,samplesPerSeries=480-8     329612836     289762604     -12.09%
```

The first commit on this PR improves the benchmark for replaying the WAL, trying to use more reasonable values for the number of samples per series, and being sure to use `b.N` so the reporting is correct.

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -365,9 +365,6 @@ func (h *Head) loadWAL(r *wal.Reader, multiRef map[uint64]uint64) (err error) {
 
 	var (
 		dec       record.Decoder
-		series    []record.RefSeries
-		samples   []record.RefSample
-		tstones   []tombstones.Stone
 		allStones = tombstones.NewMemTombstones()
 		shards    = make([][]record.RefSample, n)
 	)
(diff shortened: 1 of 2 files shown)


tsdb/head_test.go
```diff
@@ -173,7 +173,9 @@ func BenchmarkLoadWAL(b *testing.B) {
 				b.ResetTimer()
 
 				// Load the WAL.
-				h.Init(0)
+				for i := 0; i < b.N; i++ {
+					h.Init(0)
```
codesome: Samples would be added to head only in the first iteration and the rest will be discarded. Should we have creation of a new head inside the loop?

csmarchbanks: :+1: Good call, done.

codesome: What were the benchmark results after this change?

csmarchbanks:
I am planning to let the WAL build to a reasonable size and the manually restart the Prometheus instances to see how long startup takes.

tsdb/head_test.go
csmarchbanks: The results in the description were done after this change.

csmarchbanks:
/benchmark cancel

Just restarted each Prometheus. Timing numbers are from the log line for starting to replay the WAL to the last segment loaded.
```
master       7m59s
pr           5m52s
improvement: 23.3%
```

geekodour:
> we don't restart the Prometheus 

There's a half baked PR to prombench on restarting prometheus at random intervals, will try to add it soon.

cc @codesome

cstyan:
Looks reasonable to me, I'll read through again later today. Is the memory usage when replaying any different with the use of the `sync.Pool`'s?

csmarchbanks:
> s the memory usage when replaying any different with the use of the sync.Pool's?

I didn't notice much of a difference. The WAL is replayed faster, so memory growth is also faster as expected. The pool shouldn't contain more than 10 items, so it is pretty limited in size.

tsdb/head.go
```diff
@@ -376,21 +373,82 @@ func (h *Head) loadWAL(r *wal.Reader, multiRef map[uint64]uint64) (err error) {
...
-
-		switch dec.Type(rec) {
-		case record.Series:
-			series, err = dec.Series(rec, series)
-			if err != nil {
-				return &wal.CorruptionErr{
-					Err:     errors.Wrap(err, "decode series"),
+
+	var (
+		decoded    = make(chan interface{}, 10)
```
codesome: Does increasing the channel size change the speed of loading WAL? (I don't know where the bottleneck is)

tsdb/head.go
```diff
@@ -463,15 +508,19 @@ func (h *Head) loadWAL(r *wal.Reader, multiRef map[uint64]uint64) (err error) {
...
 			}
+			//lint:ignore SA6002 relax staticcheck verification.
+			tstonesPool.Put(v)
 		default:
-			return &wal.CorruptionErr{
-				Err:     errors.Errorf("invalid record type %v", dec.Type(rec)),
-				Segment: r.Segment(),
-				Offset:  r.Offset(),
-			}
+			panic(fmt.Errorf("unexpected decoded type: %T", d))
```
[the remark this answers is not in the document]
csmarchbanks: I used panic because this is an issue where there is a severe bug in the code, rather than something like data corruption. Does that sound like a good case for panic?

tsdb/head.go
csmarchbanks: I haven't played with channel size  much. Where the bottleneck is depends highly on shape of your data, hardware, and settings (such as WAL compression) so I think any value is going to be a guess. I chose a value pretty arbitrarily that gives a bit of headroom, but didn't make me concerned about increased memory usage...

tsdb/head.go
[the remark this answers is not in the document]
codesome: Yeah makes sense

tsdb/head.go
codesome: I tried with channel size 100 in my machine and no difference in the benchmark numbers. 10 sounds good then.

PR #6231, opened 2019-10-27
Replace time.After with time.NewTimer

lelenanam:
If the rate of compaction is high we might leave a lot of stray timers.

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -588,9 +588,9 @@ func (db *DB) run() {
 			return
 		case <-time.After(backoff):
 		}
-
+		timer := time.NewTimer(1 * time.Minute)
 		select {
-		case <-time.After(1 * time.Minute):
+		case <-timer.C:
 			select {
 			case db.compactc <- struct{}{}:
 			default:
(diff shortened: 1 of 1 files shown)


tsdb/db.go
```diff
@@ -588,9 +588,9 @@ func (db *DB) run() {
 			return
 		case <-time.After(backoff):
 		}
-
+		timer := time.NewTimer(1 * time.Minute)
```
codesome: Ticker makes more sense here. A timer would run the compaction only once.

tsdb/db.go
```diff
@@ -611,8 +611,10 @@ func (db *DB) run() {
 			}
 			db.autoCompactMtx.Unlock()
 		case <-db.stopc:
+			timer.Stop()
```
codesome: This could be in a `defer` statement just after initialization.

tsdb/db.go
lelenanam: Considering backoff logic it isn't that easy to convert per-iteration timer to one ticker.

tsdb/db.go
lelenanam: defer will hold a reference for each timer in that case and that will be a memory leak.

tsdb/db.go
codesome: Ah sorry, I didn't notice the loop enclosed this. In that case, the current implementation looks simpler and `time.After` uses `time.NewTimer` internally. 1 object per minute doesn't hurt much.

PR #6232, opened 2019-10-27
reach router integration initial commit

boyskila:


boyskila:
@juliusv so far reach router is integrated and paths are mapped to all pages we have. The UI is similar to the old one and improvements can be made in separate PR(if is needed) and I think here we can just verify the correctness of the paths.

web/ui/react-app/src/App.css
```diff
@@ -10,6 +10,26 @@ button.classic-ui-btn {
     margin-bottom: 20px;
 }
 
+.navbar a {
```
juliusv: I think none of these custom CSS styles should be needed, as we just want the default Bootstrap dark navbar (https://getbootstrap.com/docs/4.3/components/navbar/#color-schemes). When I remove the custom styles and add the `navbar-dark` class to the navbar, it starts looking (almost) like the one in the classic UI.

web/ui/react-app/src/App.css
```diff
@@ -10,6 +10,26 @@ button.classic-ui-btn {
     margin-bottom: 20px;
```
juliusv: GitHub doesn't let me comment on the right line, but you can now remove the body padding at the very top (line 1-3) to remove the extra space between the top of the page and the beginning of the navbar.

web/ui/react-app/src/Navbar.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+const DropdownNavLink = (props: any) => (
+  <Link getProps={isActiveDropDown} {...props} />
+);
+
+const Navigation = () => {
+  const [isOpen, setIsOpen] = useState(false);
+  const toggle = () => setIsOpen(!isOpen);
+  return (
+    <Navbar className="mb-3" color="dark" expand="md">
```
juliusv: This this change, the custom CSS is no longer needed:

```suggestion
    <Navbar className="mb-3" dark color="dark" expand="md">
```

web/ui/react-app/src/Navbar.tsx
```diff
@@ -0,0 +1,68 @@
...
+              <DropdownItem tag={DropdownNavLink} to="/react/rules">Rules</DropdownItem>
+              <DropdownItem tag={DropdownNavLink} to="/react/targets">Targets</DropdownItem>
+              <DropdownItem tag={DropdownNavLink} to="/react/service-discovery">Service Discovery</DropdownItem>
+            </DropdownMenu>
+          </UncontrolledDropdown>
+          <NavItem>
+            <NavLink href="https://prometheus.io/docs/prometheus/latest/getting_started/">Help</NavLink>
+          </NavItem>
+        </Nav>
+        <Link to="../../graph">Return to classic UI</Link>
```
juliusv: I think it's fine to just make this a normal `NavItem` like the other menu items and just call it `Classic UI`. It's going to go away at some point anyway :)

web/ui/react-app/src/App.css
boyskila: Yep, you are right. "dark" replace all custom CSS, except the active state of the tabs. I'll check whether reactstrap doesn't provide something to replace the isActive wrapper from the NavBar

web/ui/react-app/src/Navbar.tsx
boyskila: Yep, ok

juliusv:
:+1: Thanks!

juliusv:
I'm going to give the Config page a shot so I can practice some React again :)

PR #6233, opened 2019-10-27
Scrape error warn log

liguozhong:


brian-brazil:
Thank you for your PR. Scrape failures are a normal expected thing, and do
not indicate an issue with the Prometheus server and thus should not spam
the logs.

juliusv:
Yeah, I'll have to agree with @brian-brazil, sorry, but thanks!

Closing.

PR #6235, opened 2019-10-27
pass the value to the input instead to downshift

boyskila:

### DIFF
--- a/web/ui/react-app/src/ExpressionInput.test.tsx
+++ b/web/ui/react-app/src/ExpressionInput.test.tsx
@@ -28,8 +28,8 @@ describe('ExpressionInput', () => {
   const expressionInput = mount(<ExpressionInput {...expressionInputProps} />);
 
   it('renders a downshift component', () => {
-    const downshift = expressionInput.find(Downshift);
-    expect(downshift.prop('inputValue')).toEqual('node_cpu');
+    const downshift = expressionInput.find(Input);
+    expect(downshift.prop('value')).toEqual('node_cpu');
   });
 
   it('renders an InputGroup', () => {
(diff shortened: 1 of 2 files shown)


juliusv:
@boyskila Could you rebase this on master? (https://github.com/prometheus/prometheus/pull/6206 got merged in the meantime)

boyskila:
> @boyskila Could you rebase this on master? (#6206 got merged in the meantime)

Sure, but a bit later. I don't have access to my local machine now

juliusv:
:+1: will merge when CI passes

juliusv:
@boyskila React tests seem to need adjustments (see Travis log).

web/ui/react-app/src/ExpressionInput.test.tsx
```diff
@@ -28,8 +28,8 @@ describe('ExpressionInput', () => {
   const expressionInput = mount(<ExpressionInput {...expressionInputProps} />);
 
   it('renders a downshift component', () => {
-    const downshift = expressionInput.find(Downshift);
-    expect(downshift.prop('inputValue')).toEqual('node_cpu');
+    const downshift = expressionInput.find(Input);
+    expect(downshift.prop('value')).toEqual('node_cpu');
```
juliusv: While this makes the test pass, it no longer fits to the test goal of `it('renders a downshift component', ...)`.

juliusv: (I'm fine changing it in either direction, whichever you think makes more sense)

boyskila: Yes, will change it. One thing I notice here is the shared instance of ExpressionInput between tests. I think each test should operate over a fresh instance of the Input and tests not to depend on each other.

juliusv: Yeah, makes sense to me, even though in this instance it's probably not super important. /cc @hooten

hooten: That sounds good to me!

hooten:
Question (for perhaps another pr; asking here bc `value` is in play):

 It seems like `PanelList` is actually managing the state of `value` [here](https://github.com/prometheus/prometheus/blob/master/web/ui/react-app/src/pages/PanelList.tsx#L128) and [here](https://github.com/prometheus/prometheus/blob/master/web/ui/react-app/src/pages/PanelList.tsx#L36). But there is still a React [Ref](https://github.com/prometheus/prometheus/blob/master/web/ui/react-app/src/ExpressionInput.tsx#L34) that also seems to play a role. Why do we need that ref?

boyskila:
@hooten to control the height of the input. If you press shift + enter the input(actually is a text area) will grow in height. In this case ```offsetHeight, clientHeight, scrollHeight``` can be taken from event.target but there are few other cases where no event exist..so ref to the rescue :)

juliusv:
@boyskila Sorry, could you do one more conflict resolution? I promise I'll remember to merge it then :)

boyskila:
@juliusv branch is rebased with master and no changes are pushed since last commit

juliusv:
:+1: thanks!

PR #6236, opened 2019-10-27
React UI: Implement /config page

juliusv:
Putting the clipboard button into the `<h2>` is maybe a bit dirty, but it's that way in the old UI and I decided to keep it like that for now.


juliusv:
@boyskila

web/ui/react-app/src/pages/Config.tsx
```diff
@@ -1,6 +1,46 @@
...
 
-const Config: FC<RouteComponentProps> = () => <div>Config page</div>
+import './Config.css';
+
+const Config: FC<RouteComponentProps> = () => {
+  const [config, setConfig] = useState(null);
+  const [error, setError] = useState("");
+  const [copied, setCopied] = useState(false);
+
+  useEffect(() => {
```
boyskila: You have to specify empty array as a second arg of useEffect in order to prevent API calls on every component update.
Thus it will act as componentDidMount

web/ui/react-app/src/pages/Config.tsx
```diff
@@ -1,6 +1,46 @@
...
+          <Button color="light" disabled={!config}>
+            {copied ? 'Copied' : 'Copy to clipboard'}
+          </Button>
+        </CopyToClipboard>
+      </h2>
+
+      {error
+        ? <Alert color="danger"><strong>Error:</strong> Error fetching configuration: {error}</Alert>
+        : config
+          ? <pre className="config-yaml">{config !== null ? config : "Loading..."}</pre>
```
boyskila: Hm, "config" should always be !== null, right?

web/ui/react-app/src/pages/Config.tsx
```diff
@@ -1,6 +1,46 @@
...
+      .catch(error => setError(error.message));
+  });
+
+  return (
+    <>
+      <h2>
+        Configuration&nbsp;
+        <CopyToClipboard
+          text={config ? config! : ''}
+          onCopy={() => { setCopied(true); setTimeout(() => setCopied(false), 1500);}}
```
boyskila: Isn't better to pass "result" param in case copy is not successful? ```(text, result) => setCopied(result)...```
Also, you can omit the extra callback in the timeout ```setTimeout(setCopied, 1500)``` should be enough

web/ui/react-app/src/pages/Config.tsx
```diff
@@ -1,6 +1,46 @@
...
+        >
+          <Button color="light" disabled={!config}>
+            {copied ? 'Copied' : 'Copy to clipboard'}
+          </Button>
+        </CopyToClipboard>
+      </h2>
+
+      {error
```
boyskila: I think Alert should be returned before everything else in case of error, wdyt?

web/ui/react-app/src/pages/Config.tsx
juliusv: Ah cool, didn't know about this. This PR was my first time ever using hooks :)

web/ui/react-app/src/pages/Config.tsx
juliusv: It's set to `null` via `useState()` before it's loaded (and the spinner works).

web/ui/react-app/src/pages/Config.tsx
juliusv: Ah, I just copied the example from https://www.npmjs.com/package/react-copy-to-clipboard, didn't realize it had a `result` param, thanks! And yeah :)

web/ui/react-app/src/pages/Config.tsx
boyskila: I mean this statement ```config !== null ? config : "Loading...``` - it should always be different than null, otherwise will not reach this line
``` config // is not null -> let's render <pre className=.....>{ here config cannot be null, so it always render config and never  will render 'Loading' }</pre>```

web/ui/react-app/src/pages/Config.tsx
juliusv: I had played around with this and preferred the error to appear in the place where the spinner is, and where the config would have appeared if there was no error. Then at least there's still the headline that tells you it's supposed to be the config page.

web/ui/react-app/src/pages/Config.tsx
juliusv: Ahhh, sorry, I totally forgot to delete that part after adding the spinner. Oops!

juliusv:
Gonna merge this for now...

PR #6238, opened 2019-10-28
Adjust React query stats font

SuperQ:
* Use root relative font size rather than px to avoid hidpi issues.
* Darken to 50% saturation of base font color.


SuperQ:
Screenshot of new font render

juliusv:
:+1:

PR #6239, opened 2019-10-28
React UI: Update try link

SuperQ:
Update try new UI link to send users directly to the new graph interface
rather than a blank UI page.

Misc: Add /tsdb/tsdb binary to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 
 /prometheus
 /promtool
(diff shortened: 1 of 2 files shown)


juliusv:
@SuperQ Ah yeah, but I think we should just change the React UI's router to forward `/new` to `/new/graph`, like the main UI does.

PR #6242, opened 2019-10-28
Fix PrometheusRemoteWriteDesiredShards

beorn7:
This rule has the queue label on both sides. `group_right` doesn't
help, we just need to do the comparison on `queue`, too.

@cstyan

### DIFF
--- a/documentation/prometheus-mixin/alerts.libsonnet
+++ b/documentation/prometheus-mixin/alerts.libsonnet
@@ -218,7 +218,7 @@
               # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
               (
                 max_over_time(prometheus_remote_storage_shards_desired{%(prometheusSelector)s}[5m])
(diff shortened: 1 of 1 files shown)


documentation/prometheus-mixin/alerts.libsonnet
```diff
@@ -218,7 +218,7 @@
               # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
               (
                 max_over_time(prometheus_remote_storage_shards_desired{%(prometheusSelector)s}[5m])
-              > on(job, instance) group_right
+              > on(job, instance, queue)
```
csmarchbanks: Is the `on` clause necessary at all?

beorn7: Right. There is actually no reason why these two should have different labels. and we want to keep all labels anyway. Will fix in next push.

PR #6243, opened 2019-10-29
WIP: status page - API and UI

boyskila:
TODO
 - [x] prepare API endpoints for runtime and build information
- [ ] api tests
- [x] react status page
- [x] status page unit tests


boyskila:
@juliusv I have some concerns about the API implementation and I decided to stop until it's not ready. I'll be happy for any feedback.

web/api/v1/api.go
```diff
@@ -108,6 +108,31 @@ type rulesRetriever interface {
...
+	GoroutineCount      int
+	GOMAXPROCS          int
+	GOGC                string
+	GODEBUG             string
+	CorruptionCount     int64
+	ChunkCount          int64
+	TimeSeriesCount     int64
+	LastConfigTime      time.Time
+	ReloadConfigSuccess bool
+	StorageRetention    string
```
juliusv: Keeping this as a string seems a bit odd in an API, on the other hand we already have it in a more structured way in the flags API, so not sure.

web/api/v1/api.go
```diff
@@ -108,6 +108,31 @@ type rulesRetriever interface {
...
+	Version   string `json:"version"`
+	Revision  string `json:"revision"`
+	Branch    string `json:"branch"`
+	BuildUser string `json:"buildUser"`
+	BuildDate string `json:"buildDate"`
+	GoVersion string `json:"goVersion"`
+}
+
+type RuntimeInfo struct {
+	Birth               time.Time
```
juliusv: `Birth` was always a bit funny term, initially from 2013 :) How about we rename it to `StartTime` in the API?

web/api/v1/api.go
```diff
@@ -108,6 +108,31 @@ type rulesRetriever interface {
...
+type PrometheusVersion struct {
+	Version   string `json:"version"`
+	Revision  string `json:"revision"`
+	Branch    string `json:"branch"`
+	BuildUser string `json:"buildUser"`
+	BuildDate string `json:"buildDate"`
+	GoVersion string `json:"goVersion"`
+}
+
+type RuntimeInfo struct {
```
juliusv: Add a doc comment and JSON struct tags for this as well?

web/api/v1/api.go
```diff
@@ -242,6 +273,8 @@ func (api *API) Register(r *route.Router) {
 	r.Get("/alertmanagers", wrap(api.alertmanagers))
 
 	r.Get("/status/config", wrap(api.serveConfig))
+	r.Get("/runtimeinfo", wrap(api.serveRuntimeInfo))
+	r.Get("/buildinfo", wrap(api.serveBuildInfo))
```
juliusv: Probably these would also be:

- `/status/runtime`
- `/status/buildinfo`

web/ui/react-app/src/pages/Status.tsx
```diff
@@ -1,6 +1,82 @@
-import React, { FC } from 'react';
+import React, { FC, Fragment } from 'react';
 import { RouteComponentProps } from '@reach/router';
+import { Table, Alert } from 'reactstrap';
+import useFetch from './useFetch';
 
-const Status: FC<RouteComponentProps> = () => <div>Status page</div>;
+const ENDPOINTS = ['../api/v1/runtimeinfo', '../api/v1/buildinfo', '../api/v1/alertmanagers'];
+const sectionsTitles = ['Runtime Information', 'Build Information', 'Alertmanagers'];
```
juliusv: sectionsTitles -> sectionTitles

web/ui/react-app/src/pages/Status.tsx
```diff
@@ -1,6 +1,82 @@
...
 
-export default Status;
+interface StatusConfig {
+  [k: string]: { title: string; normalizeValue?: (v: any) => any };
+}
+
+type StatusPageState = Array<{ [k: string]: string }>;
+
+const normalizeAlertManagerValue = (alertMgrs: { url: string }[]) => {
```
juliusv: AlertManager -> Alertmanager (no logic behind it except being consistent with how we write Alertmanager :))

web/ui/react-app/src/pages/Status.tsx
```diff
@@ -1,6 +1,82 @@
...
+  return alertMgrs.map(({ url }) => (
+    <a key={url} className="mr-2" href={url}>
+      {url}
+    </a>
+  ));
+};
+
+const statusConfig: StatusConfig = {
+  Birth: { title: 'Uptime', normalizeValue: (v: string) => new Date(v).toUTCString() },
```
juliusv: Instead of `Uptime` (which implies a duration, not a timestamp), we should also change this to say `Start time` in the UI.

web/ui/react-app/src/pages/Status.tsx
```diff
@@ -1,6 +1,82 @@
...
+  CWD: { title: 'Working Directory' },
```
juliusv: Let's change the new status page to be consistent in Title Casing (or rather not title casing), I'd just go for `Working Directory` -> `Working directory` here (and below for storage retention).

web/ui/react-app/src/pages/Status.tsx
```diff
@@ -1,6 +1,82 @@
...
+    );
+  } else if (spinner) {
+    return spinner;
+  }
+  return data
+    ? data.map((statuses, i) => {
+        return (
+          <Fragment key={i}>
+            <h2>{sectionsTitles[i]}</h2>
+            <Table size="sm" key={i} borderless>
```
juliusv: Does the table need a `key` although it is not a list item within its parent?

web/ui/react-app/src/pages/Status.tsx
```diff
@@ -1,6 +1,82 @@
...
+              <tbody>
+                {Object.entries(statuses).map(([k, v], index: number) => {
+                  const { title = k, normalizeValue = (val: any) => val } = statusConfig[k] || {};
+                  return (
+                    <tr key={k} className={index % 2 ? 'bg-white' : 'bg-light'}>
```
juliusv: No custom calculation needed for the row coloring, just set the `striped` attribute on the table (which sets the `table-striped` Bootstrap class.

web/api/v1/api.go
boyskila: I decided to omit the "status" part since alertmanager route doesn't use it. Should I change the alertmanagers path as well?

web/ui/react-app/src/pages/Status.tsx
boyskila: Oh, nice....didn't know that, thanks!

web/api/v1/api.go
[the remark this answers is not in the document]
boyskila: @brian-brazil Thank you for your feedback! Can you please share your thoughts about using string Map instead of struct. Generally, I took it from the current status Handler (responsible for executing status.html template in web.go) and use it as a type of what API should receive.

web/api/v1/api.go
juliusv: We can't break the existing v1 API, but I guess `/api/v1/alertmanagers` is also far-enough removed from the main status of the Prometheus server (like `/api/v1/targets`, `/api/v1/service-discovery`, and the like) that I wouldn't really expect it to be under `/api/v1/status/...`. I guess someone still decided to put it onto the `/status` page because they didn't feel creating an extra web page for that little information was worth it :man_shrugging: 

But runtime and build info are infos that are closely tied to the server itself, so I think they will best live under `/api/status/...`

web/api/v1/api.go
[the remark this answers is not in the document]
juliusv: I don't see how it would work well for `RuntimeInfo` to be a map of strings (at least for the values), given that some values are not strings, and should be represented as non-strings in JSON. For `PrometheusVersion` that might work better, but I think a Go struct is fine there too.

I also share the concern of "this is a big bag of values, you can't rely on these staying the same over Prometheus versions". Not sure how to signal that besides documenting it clearly (and/or moving this endpoint outside of `/api/v1`, but I'd rather keep it as part of the v1 API I think).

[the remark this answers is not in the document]
brian-brazil: This belongs in v1 one way or the other.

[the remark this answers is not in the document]
juliusv: Yeah, would think so as well. So we'll just document that the fields can change with a big warning then.

boyskila:
@juliusv @hooten @csmarchbanks just add snapshot testing https://github.com/prometheus/prometheus/pull/6243/commits/d6bd23415f23dae8ed44f6d2dd778634de710e19 WDYT?

web/ui/react-app/src/test/Status.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+          .text()
+      ).toEqual('Build Information');
+      expect(
+        wrapper
+          .find('h2')
+          .at(2)
+          .text()
+      ).toEqual('Alertmanagers');
+    });
+    it('should match table snapshot', () => {
```
juliusv: Yeah, I like the snapshot test for this, but then I don't think we need the two tests above anymore, as they're just testing a subset of the snapshot test?

web/web.go
```diff
@@ -744,6 +739,48 @@ func (h *Handler) status(w http.ResponseWriter, r *http.Request) {
...
+	if h.options.TSDBCfg.MaxBytes != 0 {
+		if status.StorageRetention != "" {
+			status.StorageRetention = status.StorageRetention + " or "
+		}
+		status.StorageRetention = status.StorageRetention + h.options.TSDBCfg.MaxBytes.String()
+	}
+
+	metrics, err := h.gatherer.Gather()
+	if err != nil {
+		fmt.Println(err)
```
juliusv: We (almost) never want to `fmt.Println()` anything (we use a logging library instead), but in this case we should either return an error to the caller or panic. Since I'm not 100% sure that none of the Prometheus-internal metric collectors ever return errors during gathering, returning an error is probably the way to go to be safe.

juliusv:
@boyskila if you're moving tests to a separate directory, shouldn't that be all existing tests as well? But I kind of like test files being in the same directory (again, I'm so used to it from Go...). I guess the motivation for the move was the snapshots directory, but is it really that disturbing?

boyskila:
Basically I always put tests in a dedicated folder but if you prefer to remain like it is I'll remove it(the snapshot folder will be created where other tests are).

boyskila:
Yes, that's why I didn't move the tests there. It's better to be done in separate PR

web/ui/react-app/src/pages/Status.tsx
boyskila: Done!

web/ui/react-app/src/test/Status.test.tsx
boyskila: Done

web/ui/react-app/src/pages/Status.tsx
```diff
@@ -1,6 +1,94 @@
...
+    normalizeValue: normalizeAlertmanagerValue,
+  },
+  droppedAlertmanagers: {
+    title: 'Dropped',
+    normalizeValue: normalizeAlertmanagerValue,
+  },
+};
+
+const Status = () => {
+  const { response: data, error, isLoading } = useFetch<StatusPageState[]>(ENDPOINTS);
```
hooten: You could consider partially rendering the page when each response comes back. Since each section of the page seems to be built on a corresponding api call, it might work nicely, be easy to implement, and render quicker.

web/ui/react-app/src/pages/Status.tsx
```diff
@@ -1,6 +1,94 @@
...
+          <Fragment key={i}>
+            <h2>{sectionTitles[i]}</h2>
+            <Table className="h-auto" size="sm" borderless striped responsive>
+              <tbody>
+                {Object.entries(statuses).map(([k, v]) => {
+                  const { title = k, normalizeValue = (val: any) => val } = statusConfig[k] || {};
+                  return (
+                    <tr key={k}>
+                      <td style={{ width: '35%' }}>{title}</td>
+                      <Row tag="td" className="m-0">
```
hooten: I'm curious: why not just use a `<td className="m-0">` here?

juliusv: Yeah, and `<th>` for the title cells? (just looking how it is in the old UI)

boyskila: Opps...resolve the conversation by mistake... The idea is to use Row styles in order to split the content into multiple rows otherwise content overflow the page and horizontal scrollbar appears

boyskila: > Yeah, and `<th>` for the title cells? (just looking how it is in the old UI)

Yes, will change it

web/ui/react-app/src/pages/Status.tsx
boyskila: Currently, the page is loading for a few ms and I guess adding spinners for each section can lead to bad user experience. But even if it takes more I still prefer to see one spinner. Also managing multiple requests separately will lead to more complexity in the code. It's just a matter of what the requirements are. What do YT? @csmarchbanks  @juliusv

juliusv: So I really wasn't sure about this, so I tried out a version locally where it loads every section separately with its own spinner. In the end I think I prefer having the single load / spinner vs. multiple ones, since the requests are really fast and the individual loading that appears but disappears almost too fast to see looks less good than the full-page load.

web/ui/react-app/src/pages/Status.tsx
```diff
@@ -1,6 +1,94 @@
-import React, { FC } from 'react';
+import React, { FC, Fragment } from 'react';
 import { RouteComponentProps } from '@reach/router';
+import { Table, Alert, Row } from 'reactstrap';
+import useFetch from '../hooks/useFetches';
```
juliusv: `useFetch` -> `useFetches`

web/ui/react-app/src/hooks/useFetches.tsx
```diff
@@ -0,0 +1,36 @@
+import { useState, useEffect } from 'react';
+
+const useFetches = <R extends any>(urls: string[], options?: RequestInit) => {
```
juliusv: Btw., this API turned out to be a bit dangerous: when you want to use it to just load a single endpoint and do `useFetches([myEndpoint])`, the page ends up in an infinite fetch loop and you can't even close the tab anymore. E.g. try changing the `Status` handler function to:

```typescript
const Status = () => {
  const ep0 = useFetch<StatusPageState>([ENDPOINTS[0]]);

  return (
    <h2>Hello</h2>
  )
};
```

It will load the runtimeinfo endpoint in an infinite loop, I guess because React thinks it's a new array every time? I could solve it by creating the single-element array as a const outside of the function and reusing that. Not sure how that can be improved for a multi-fetch function though.

boyskila: Yep, I found this yesterday when writing tests for it. The same case is with the second argument - options. Try ```const ep0 = useFetch<StatusPageState>(ENDPOINTS, {});```

web/ui/react-app/src/pages/Status.tsx
juliusv: I'm not sure I understand. Whether I use this code or just change it to a normal `<td>`, I get scrollbars when I make the browser window narrow enough, and nothing is split into rows?

boyskila: Yes, this should work for multiple URLs splited by an interval but will not fit it if it's too long. Just tried with text-break class form bootstrap and looks fine

web/ui/react-app/src/hooks/useFetches.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+  });
+  it('should isLoading remains true on empty response', async done => {
+    fetchMock.mockResponse(jest.fn());
+    const { result, waitForNextUpdate } = renderHook(useFetches, { initialProps: ['/foo/bar'] });
+    expect(result.current.isLoading).toEqual(true);
+    await waitForNextUpdate();
+    setTimeout(() => {
+      expect(result.current.isLoading).toEqual(true);
+      done();
+    }, 1000);
```
juliusv: Is the `setTimeout()` here still needed after the `waitForNextUpdate()`? It seems to work without as well (but maybe just because the value is `true` to begin with)...

boyskila: Yes, It is working without the setTimeout. I just wanted to be clear that isLoading is checked after the response.

web/ui/react-app/src/hooks/useFetches.tsx
juliusv: Reading more, I wonder if we are correctly obeying the rules of hooks? https://reactjs.org/docs/hooks-overview.html#rules-of-hooks

"Only call Hooks at the top level. Don’t call Hooks inside loops, conditions, or nested functions."

I mean, it seems to work...

juliusv: From the more verbose explanation at https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level it makes sense that we don't have a direct problem here, since we don't have variability in ordering or similar, but it's harder to ensure this way of course.

boyskila: I think we are not breaking any of the rules so far but yeah hooks are a bit dangerous

web/api/v1/api.go
```diff
@@ -1176,6 +1222,7 @@ func (api *API) respondError(w http.ResponseWriter, apiErr *apiError, data inter
 		Error:     apiErr.err.Error(),
 		Data:      data,
 	})
+	fmt.Println(apiErr)
```
juliusv: Left-over debug Println

web/ui/react-app/test/Status.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+  });
+  it('should fetch proper API endpoints', () => {
+    const useFetchSpy = jest.spyOn(useFetch, 'default');
+    shallow(<Status />);
+    expect(useFetchSpy).toHaveBeenCalledWith(['../api/v1/status/runtimeinfo', '../api/v1/status/buildinfo', '../api/v1/alertmanagers']);
+  });
+  describe('Snapshot testing', () => {
+    const response = [
+      {
+        StartTime: '2019-10-30T22:03:23.247913868+02:00',
```
juliusv: It probably doesn't matter for the tests, but in the actual API response, these fields are differently cased now.

juliusv:
@boyskila Thanks for all the patience :) I'm going to merge and any follow-up we can do separately, if needed.

PR #6247, opened 2019-10-29
[docs] Correct the docker run volume parameter

mattrussi:
* Fixes prometheus/docs#1418

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -39,15 +39,19 @@ two examples.
 Bind-mount your `prometheus.yml` from the host by running:
 
 ```bash
-docker run -p 9090:9090 -v /tmp/prometheus.yml:/etc/prometheus/prometheus.yml \
-       prom/prometheus
+docker run \
+    -p 9090:9090 \
+    -v /tmp/prometheus.yml:/etc/prometheus/prometheus.yml \
+    prom/prometheus
 ```
 
 Or use an additional volume for the config:
 
 ```bash
-docker run -p 9090:9090 -v /prometheus-data \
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -39,15 +39,20 @@ two examples.
...
 ```
 
 Or use an additional volume for the config:
 
 ```bash
-docker run -p 9090:9090 -v /prometheus-data \
-       prom/prometheus --config.file=/prometheus-data/prometheus.yml
+docker run \
+    -p 9090:9090 \
+    -v /path/to/data:/prometheus-data \
```
simonpasquier: I suppose the idea was to show how to [start the container with a volume](https://docs.docker.com/storage/volumes/#start-a-container-with-a-volume). Using a volume name instead of a directory path would be less confusing:

```
docker run \
  -v prometheus:/prometheus-data
  ...
```

cc @SuperQ

SuperQ: No, I prefer to show the local path mapping.

I also think we should remove the `--config` option from the example, and show how to pass the config into the default location (`/etc/prometheus/prometheus.ym`) with no additional flags.

By including the one flag override, it breaks the other default flags current `CMD` config.

docs/installation.md
```diff
@@ -39,15 +39,20 @@ two examples.
...
+    prom/prometheus \
+    --config.file=/prometheus-data/prometheus.yml
```
SuperQ: Let's remove this, and provide an example for volume mapping the config. Something like `-v /path/to/config:/etc/prometheus`.

mattrussi: Sounds good! Updated with your suggestion.

Issue #6164, opened 2019-10-17
React UI: Convert remainder of Prometheus web UI to React


juliusv:
The [new React UI](https://github.com/prometheus/prometheus/pull/5694) only implements the graphing/querying interface so far. It is based on [create-react-app](https://github.com/facebook/create-react-app), which only supports the full page being a React app (vs. React just being part of the page). At least if we decide to stay with vanilla create-react-app, we'll need to convert the rest of the Prometheus UI to be React-based as well.

For some of the information on the current UI (like `/rules`), we don't have fitting APIs yet to fetch (and then render) the required information. We need to decide:

- Whether to really go single-page app or move away from create-react-app
- If yes: Whether to create / modify existing APIs to get the required information, or what other possibilities exist

boyskila:
How about to set an environment variable and if is React to send JSON response, otherwise to render the template
![image](https://user-images.githubusercontent.com/10514821/67384076-25884600-f599-11e9-9495-bbac4da919db.png)

brian-brazil:
I thought we'd added all the APIs by now. Which are missing?

juliusv:
At a glance, I see a couple of things missing:

- `/rules` page - `/api/v1/status/rules` API endpoint
  - missing last evaluation time
  - missing last evaluation duration
- `/status` page with general build and runtime information
  - no equivalent API endpoint seems to exist at all, except for the Alertmanager info at `/api/v1/alertmanagers`
- `/targets` page - `/api/v1/targets` API endpoint
  - AFAICS the API endpoint doesn't include any grouping info about the scrape config a target came from
  - We are probably missing the extra bit here where the `/targets` page sometimes shows a different scrape URL than it links to (when the server detects a URL containing a localhost representation, the link gets changed to be to the external IP instead)
- `/service-discovery` page - `/api/v1/targets` API endpoint
  - API is missing scrape config grouping info
- `/alerts` - `/api/v1/alerts` API endpoint
  - API is missing association of active alerts to alerting rules

simonpasquier:
I'm all for extending the API endpoints. This would also benefit other web apps that want to integrate with Prometheus.

juliusv:
@simonpasquier Yeah, me too. I will miss immediately-there & greppable HTML pages somewhat, but oh well (one can still grep / jq the API endpoints)...

simonpasquier:
I won't miss the Go templates :stuck_out_tongue:

juliusv:
@boyskila
[...]

PR #6248, opened 2019-10-29
Implement the /flags page in react

csmarchbanks:
The page itself was straightforward, but I would especially like a review/discussion on the test. I couldn't find any great prior art around using enzyme to test a mocked asynchronous call so I did my best. If anyone else has a better idea I am all ears!

Added a dependency on [jest-fetch-mock](https://www.npmjs.com/package/jest-fetch-mock).

cc: @juliusv @boyskila @hooten

`/flags` page for https://github.com/prometheus/prometheus/issues/6164


boyskila:
Cool! I would use spyOn over fetch
```
      jest.spyOn(global, 'fetch').mockImplementation(() => Promise.resolve({
        json: () => Promise.resolve(sampleAlertmanagerResponse),
      }));
```
And then you can check
```
expect(global.fetch).toHaveBeenCalledWith('../api/v1/status/flags')
        expect(global.fetch).toHaveBeenCalledTimes(1);
```
And for the error handling I guess you can return new Error() instead of resolving the Promise

web/ui/react-app/src/pages/Flags.tsx
```diff
@@ -1,6 +1,52 @@
...
+  const [error, setError] = useState('');
+
+  useEffect(() => {
+    fetch('../api/v1/status/flags')
+      .then(res => res.json())
+      .then(res => setFlags(res.data))
+      .catch(error => setError(error.message));
+  }, []);
+
+  const body = (error: string, flags: { [key: string]: any } | null) => {
```
hooten: It looks like the type here is slightly different from the type of the `flag` state

web/ui/react-app/src/pages/Flags.tsx
```diff
@@ -1,6 +1,52 @@
...
+    if (error) {
+      return (
+        <Alert color="danger">
+          <strong>Error:</strong> Error fetching flags: {error}
```
hooten: Is the error something we want to display? Does it need any formatting?

web/ui/react-app/src/pages/Flags.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+      const flags = shallow(<Flags />);
+      const icon = flags.find(FontAwesomeIcon);
+      expect(icon.prop('icon')).toEqual(faSpinner);
+      expect(icon.prop('spin')).toEqual(true);
+    });
+  });
+
+  describe('when data is returned', () => {
+    it('renders a table', async () => {
+      const mock = fetch.mockResponse(JSON.stringify(sampleAlertmanagerResponse));
```
hooten: I think this is pretty cool. It seems slightly magical, but I like that your Flag component doesn't have to cater to the test code.

web/ui/react-app/src/pages/Flags.tsx
```diff
@@ -1,6 +1,52 @@
-import React, { FC } from 'react';
+import React, { FC, useEffect, useState } from 'react';
 import { RouteComponentProps } from '@reach/router';
+import { Alert, Table } from 'reactstrap';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faSpinner } from '@fortawesome/free-solid-svg-icons';
 
-const Flags: FC<RouteComponentProps> = () => <div>Flags page</div>;
+const Flags: FC<RouteComponentProps> = () => {
```
hooten: Yay React Hooks! I think this is the best way to manage state and fetches

web/ui/react-app/src/pages/Flags.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+    'web.route-prefix': '/',
+    'web.user-assets': '',
+  },
+};
+
+describe('Flags', () => {
+  beforeEach(() => {
+    fetch.resetMocks();
+  });
+
```
boyskila: I would test whether fetch is called with proper endpoind ....something like
```expect(fetch).toBeCalledWith('../../api/v1/flags')```

File: web/ui/react-app/src/pages/Flags.tsx [hunk 6]
```diff
@@ -1,6 +1,52 @@
...
+  const [flags, setFlags] = useState<{ [key: string]: string } | null>(null);
```
boyskila: Here null can be omitted and flags param in "body" function should be set as optional.

web/ui/react-app/src/pages/Flags.tsx
```diff
@@ -1,6 +1,52 @@
...
+        </Table>
+      );
+    }
+    return <FontAwesomeIcon icon={faSpinner} spin />;
+  };
+
+  return (
+    <>
+      <h2>Command-Line Flags</h2>
+      {body(error, flags)}
```
boyskila: I think you don't need to pass the values. They are visible in the scope

web/ui/react-app/src/pages/Flags.tsx
boyskila: https://github.com/prometheus/prometheus/pull/6243/files - useFetch example

web/ui/react-app/package.json
```diff
@@ -64,9 +64,12 @@
     "@types/flot": "0.0.31",
     "@types/moment-timezone": "^0.5.10",
     "@types/reactstrap": "^8.0.5",
+    "@types/sinon": "^7.5.0",
```
juliusv: Nothing in this file changed except for alphabetic sorting of packages. Does anyone know if this depends on the yarn versions or something? Would be good to get this consistent for everyone so we don't get extra diff noise all the time.

juliusv: Ah sorry, `jest-fetch-mock` is new after all, but the rest just got reordered. My comment about that still stands.

web/ui/react-app/src/pages/Flags.tsx
juliusv: Yeah, I like that approach!

web/ui/react-app/src/pages/Flags.test.tsx
```diff
@@ -0,0 +1,111 @@
+import * as React from 'react';
+import { mount, shallow, ReactWrapper } from 'enzyme';
+import { act } from 'react-dom/test-utils';
+import Flags from './Flags';
+import { Alert, Table } from 'reactstrap';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faSpinner } from '@fortawesome/free-solid-svg-icons';
+
+const sampleAlertmanagerResponse: {
```
juliusv: `sampleAlertmanagerResponse` -> `sampleFlagsResponse` (it has nothing to do with Alertmanager, I think you probably accidentally copied this from the example in another comment).

web/ui/react-app/src/pages/Flags.tsx
juliusv: Yeah, this is the standard way of how we display errors so far, I think it's a good idea...

web/ui/react-app/src/pages/Flags.test.tsx
csmarchbanks: ... I think I was talking about alertmanager with hooten while writing this and got my thoughts crossed. Ooops.

web/ui/react-app/src/pages/Flags.tsx
csmarchbanks: :+1: Good idea! I implemented this and adapted the config page to use it. Check it out in the second commit.

csmarchbanks:
@boyskila If I use the `jest.spyOn` route, would I have to also mock out the whole response (things like `ok`, and `statusText` now that the custom react hook is testing those as well? If so would that be any better than using a pre-built mock library?

web/ui/react-app/src/pages/Flags.tsx
boyskila: Awesome! Looks like I have adopted another version of the custom hook since logic in the status page requires multiple requests at once. We should somehow take from both :) @csmarchbanks  can you please check https://github.com/prometheus/prometheus/blob/f21da6bba7f332eb750d412d313af27e22acf675/web/ui/react-app/src/pages/useFetch.tsx :)

boyskila:
> @boyskila If I use the `jest.spyOn` route, would I have to also mock out the whole response (things like `ok`, and `statusText` now that the custom react hook is testing those as well? If so would that be any better than using a pre-built mock library?

Yes, yes, I agree that using a library is more convenient especially for testing, where adding new libraries has no impact on bundle. I mention that I'd use "spyOn" instead of a library because I think after creating the custom hook for "Fetch" we make all other components independent and we can test them as "dumb" components, ie. we just need to pass some mock data and to check whether they are rendered properly. This is good because now we can test fetch in isolation and since the custom hook is not something special we can just create a mock implementation of fetch and to test only logic we are using in it. WDYT?

hooten:
I've used this [`fetch-suspense`](https://github.com/CharlesStover/fetch-suspense) which has a `useFetch` and has a pretty neat pattern. Not sure if it's best to add a dependency or manage the functionality ourselves.

csmarchbanks:
> I mention that I'd use "spyOn" instead of a library because I think after creating the custom hook for "Fetch" we make all other components independent and we can test them as "dumb" components, ie. we just need to pass some mock data and to check whether they are rendered properly. This is good because now we can test fetch in isolation and since the custom hook is not something special we can just create a mock implementation of fetch and to test only logic we are using in it. WDYT?

I am always a bit torn on situations like this. Personally, I prefer having mocks in as few places as possible, and at as low a level as possible (in this case at just the `fetch` level). I prefer fewer mocks because it provides me more confidence that everything is wired up correctly even after a refactor. If others strongly disagree I am ok changing it.

> I've used this fetch-suspense library before. It has a useFetch and a pretty neat pattern using the Suspense component. Not sure if it's best to add a dependency or manage the functionality ourselves.

The component looks interesting, not sure I want to implement it quite yet, but I will keep it in mind.

web/ui/react-app/src/pages/Flags.tsx
csmarchbanks: Thank you for the best practice advice, I missed that! I have updated my function.

I guess I think it's the component's responsibility to decide what to render depending on the data received so far rather than encouraging people to group everything together. For example, if loading two parts of the same page, it would be nice to have partial information in many cases while the second part (possibly a more expensive request) loads.

boyskila: > I guess I think it's the component's responsibility to decide what to render depending on the data received so far rather than encouraging people to group everything together. For example, if loading two parts of the same page, it would be nice to have partial information in many cases while the second part (possibly a more expensive request) loads.

Being able to group requests or passing a single URL doesn't limit you at all. In my case, I want to request all the data and then to render it. If useFetch doesn't support it I should implement it anyway ( in my component ).
Regarding partial information on the page, I'm 100% agree. Then requesting different endpoints separately make more sense. Definitely depends on the situation and I'm more for supporting both options. Kind of polymorphism I would say :)

csmarchbanks: I would like an opinion from @juliusv or @hooten. Either way works, it is just whether we want to have the complexity in the helper function or the component (where you already have to deal with no-data states). If they think having the multiple url pattern is helpful I am happy to use that.

That said I am very much against returning a spinner as that is conflating data fetching with display concerns.

boyskila: > That said I am very much against returning a spinner as that is conflating data fetching with display concerns.

I agree for the spinner. This is kind of proof of concept and I decided to try it anyway. For the logic behind useFetch, I'm don't mind to support both options as long as useFetch is doing one thing - fetching :)

hooten: I agree that the return value of `useFetch` should not return a spinner, but rather something like `{ response, error, loading }` where `loading` is a `boolean`

I understand the concern of wanting to wait for multiple fetches to return before rendering, but it might be cleaner to use function composition rather than polymorphism to accomplish that. That is, maybe we need a `useFetches` that maps over `useFetch` with an array of fetches and awaits their responses.

hooten: It could also return a `setFetch` so that the client can rerun a fetch (e.g., on a user event). Just an idea!

boyskila: I just want to pass my string URL and to receive the data :D

juliusv: This module shows another approach, having the display logic as a child function of a `<Fetch>` component: https://www.npmjs.com/package/react-fetch-component

Ultimately I don't care much about the exact approach (already too many voices here :)), but generally I always prefer to err on the side of code understandability / simplicity / pragmatism vs. too much abstraction and avoiding any repetition (that's the Go side of me).

juliusv: So... we have https://github.com/prometheus/prometheus/pull/6243 flying in parallel with this one. We also have my existing Config page implementation (which has no abstraction at all yet).

I'm fine merging different approaches for now, but we should all converge on the same way of doing data fetching for these pages... since https://github.com/prometheus/prometheus/pull/6243 is already adding a `useFetch()` as described here, that's probably a good candidate unless someone has objections?

csmarchbanks: This PR includes updating the Config page implementation so those will be in sync. I liked @hooten's idea of having two functions `useFetch` and `useFetches` to make it clear what is being done, so maybe this PR keeps this `useFetch` and #6243 builds the `useFetches` function? Does that sound reasonable?

juliusv: Reading this thread again, I think I like that idea most too (as in, I hope I'll be less confused by functional composition of `useFetch`/`useFetches` than by the TypeScript polymorphism at which I had to stare for a while until I understood what was going on :))

@boyskila Is that ok with you?

boyskila: Ahh, guys..sorry. Just back home. I like that approach too and definitely is easier to understand what functions are doing just by reading their names. So I'm going to rename my version of useFetch to useFetches and ... I think It was constructive discussion :)

csmarchbanks: 👍 Thank you all, I definitely learned some things. Resolving this conversation!

csmarchbanks:
@juliusv I think this is ready to go, is there anything else I should take a look at?

juliusv:
@csmarchbanks I already mentioned to @boyskila earlier today that it may have been a mistake to encourage multiple people to work on the various status pages in parallel before settling the data fetching debate, since now everyone has built slightly different mechanisms.

The `useFetches()` in https://github.com/prometheus/prometheus/pull/6243 seems a bit more general in that it allows multiple requests. Not sure what to do now, I could merge this one and we can consolidate later, or what do you think?

csmarchbanks:
I would say if the code looks ok to you lets merge this one and consolidate later. That way having at least one `useFetch` will allow other pages to get worked on while #6234 is iterated on. We can decide later if consolidating to one function is worthwhile or not.

csmarchbanks:
Also, if the status page ends up being the only page that needs multiple urls all returning before a render, that would feed into my opinion on whether a multi-url option should exist at all or if the complexity should be encapsulated by that component.

juliusv:
Yeah, merging this for now, but then we should definitely consolidate in some way.

hooten:
Nice work, Chris

PR #6249, opened 2019-10-30
clean up indirects in go.mod manifest

tariq1890:
go mod adds indirect dependencies to version pin transitively imported projects that don't have a go.mod in them.

go mod tidy will not remove the indirect project version pins even if the project starts to use go.mod (for very understandable reasons). In this PR, I've cleaned the transitive indirects up and re-run the go mod tidy.

This was done with `go 1.13.3`


brian-brazil:
Thanks!

PR #6262, opened 2019-11-02
Document new status API endpoints

juliusv:
These were added in https://github.com/prometheus/prometheus/pull/6243
for the React UI.

### DIFF
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -696,6 +696,8 @@ $ curl http://localhost:9090/api/v1/status/flags
 }
 ```
 
+*New in v2.2*
+
 ### Runtime Information
 
 The following endpoint returns various runtime information properties about the Prometheus server:
@@ -729,6 +731,8 @@ $ curl http://localhost:9090/api/v1/status/runtimeinfo
 
 **NOTE**: The exact returned runtime properties may change without notice between Prometheus versions.
 
+*New in v2.14*
+
 ### Build Information
 
 The following endpoint returns various build information properties about the Prometheus server:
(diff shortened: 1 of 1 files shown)


juliusv:
@brian-brazil (FYI @boyskila)

docs/querying/api.md
```diff
@@ -696,6 +696,66 @@ $ curl http://localhost:9090/api/v1/status/flags
...
+    "goroutineCount": 48,
+    "GOMAXPROCS": 4,
+    "GOGC": "",
+    "GODEBUG": "",
+    "storageRetention": "15d"
+  }
+}
+```
+
+**NOTE**: The exact returned runtime properties may change without notice between Prometheus versions!
```
brian-brazil: I don't think we need the exclamation mark.

docs/querying/api.md
```diff
@@ -696,6 +696,66 @@ $ curl http://localhost:9090/api/v1/status/flags
...
+    "buildUser": "julius@desktop",
+    "buildDate": "20191102-16:19:59",
+    "goVersion": "go1.13.1"
+  }
+}
+```
+
+**NOTE**: The exact returned build properties may change without notice between Prometheus versions!
+
 *New in v2.2*
```
brian-brazil: Needs adding/updating/moving.

juliusv: Hmmm... or how about I just add equivalent `New in v2.14` notices for the new endpoints and keep this as is?

docs/querying/api.md
juliusv: I thought the exclamation mark would keep the API stability lawyers away ;) Replaced with a period.

docs/querying/api.md
juliusv: Oh, the "new in" referred to the section above it, not below it, right?

juliusv: Hopefully fixed now.

docs/querying/api.md
brian-brazil: We need to mention it, not shout it from the rooftops :)

brian-brazil:
:+1:

PR #6263, opened 2019-11-03
React UI: Simpler and relative-only query URL building

juliusv:
I prefer just keeping it dumb (and it helps with a current Netlify
deploy hack I have).


boyskila:
+1

PR #6264, opened 2019-11-03
React UI: Support custom path prefixes

juliusv:
The challenge was that the path prefix can be set dynamically as a flag
on Prometheus, but the React app bundle is statically compiled in to
expect a given path prefix. By adding a placeholder value to the React
app's index.html and replacing it in Prometheus with the right path
prefix during serving, this injects Prometheus's path prefix into the
React app via a global const.

Threading the path prefix into the different React components could have
been done with React's Contexts (https://reactjs.org/docs/context.html),
but I found the consumer side of context values to be a bit cumbersome
(wrapping entire components in context consumers), so I ended up
preferring direct threading of the path prefix values to components that
needed them. Also, using contexts in tests is more verbose than just
passing in path prefix values directly.

Fixes https://github.com/prometheus/prometheus/issues/6163

### DIFF
--- a/web/ui/react-app/src/App.test.tsx
+++ b/web/ui/react-app/src/App.test.tsx
@@ -7,15 +7,17 @@ import { Router } from '@reach/router';
 import { Alerts, Config, Flags, Rules, Services, Status, Targets, PanelList } from './pages';
 
 describe('App', () => {
-  const app = shallow(<App />);
+  const app = shallow(<App pathPrefix="/path/prefix" />);
 
   it('navigates', () => {
     expect(app.find(Navigation)).toHaveLength(1);
   });
   it('routes', () => {
-    [Alerts, Config, Flags, Rules, Services, Status, Targets, PanelList].forEach(component =>
-      expect(app.find(component)).toHaveLength(1)
-    );
+    [Alerts, Config, Flags, Rules, Services, Status, Targets, PanelList].forEach(component => {
+      const c = app.find(component);
+      expect(c).toHaveLength(1);
+      expect(c.prop('pathPrefix')).toBe('/path/prefix');
+    });
     expect(app.find(Router)).toHaveLength(1);
     expect(app.find(Container)).toHaveLength(1);
   });
--- a/web/ui/react-app/src/index.tsx
+++ b/web/ui/react-app/src/index.tsx
@@ -4,4 +4,16 @@ import ReactDOM from 'react-dom';
 import App from './App';
 import 'bootstrap/dist/css/bootstrap.min.css';
 
-ReactDOM.render(<App />, document.getElementById('root'));
+// Declared/defined in public/index.html, value replaced by Prometheus when serving bundle.
+declare const PATH_PREFIX: string;
+
+let prefix = PATH_PREFIX;
+if (PATH_PREFIX === 'PATH_PREFIX_PLACEHOLDER' || PATH_PREFIX === '/') {
+  // Either we are running the app outside of Prometheus, so the placeholder value in
+  // the index.html didn't get replaced, or we have a '/' prefix, which we also need to
+  // normalize to '' to make concatenations work (prefixes like '/foo/bar/' already get
+  // their trailing slash stripped by Prometheus).
+  prefix = '';
+}
+
+ReactDOM.render(<App pathPrefix={prefix} />, document.getElementById('root'));
--- a/web/ui/react-app/src/pages/Status.tsx
+++ b/web/ui/react-app/src/pages/Status.tsx
@@ -5,8 +5,9 @@ import useFetches from '../hooks/useFetches';
 
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faSpinner } from '@fortawesome/free-solid-svg-icons';
+import PathPrefixProps from '../PathPrefixProps';
 
-const ENDPOINTS = ['../api/v1/status/runtimeinfo', '../api/v1/status/buildinfo', '../api/v1/alertmanagers'];
+const ENDPOINTS = ['/api/v1/status/runtimeinfo', '/api/v1/status/buildinfo', '/api/v1/alertmanagers'];
 const sectionTitles = ['Runtime Information', 'Build Information', 'Alertmanagers'];
 
 interface StatusConfig {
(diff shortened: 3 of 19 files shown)


juliusv:
@csmarchbanks from the Prometheus team side

@boyskila for general React comments :)

web/ui/react-app/src/index.tsx
```diff
@@ -4,4 +4,16 @@ import ReactDOM from 'react-dom';
...
+
+let prefix = PATH_PREFIX;
+if (PATH_PREFIX === 'PATH_PREFIX_PLACEHOLDER' || PATH_PREFIX === '/') {
+  // Either we are running the app outside of Prometheus, so the placeholder value in
+  // the index.html didn't get replaced, or we have a '/' prefix, which we also need to
+  // normalize to '' to make concatenations work (prefixes like '/foo/bar/' already get
+  // their trailing slash stripped by Prometheus).
+  prefix = '';
+}
+
```
boyskila: Just an idea - instead of passing prefix down to all components can you just export from here all the endpoints
For example, Status page endpoints will look like:
```
export const statusPagePaths = [
    '/api/v1/status/runtimeinfo',
    '/api/v1/status/buildinfo',
    '/api/v1/alertmanagers'
].map(p => `${prefix}${p}`);

```

I think this will reduce a lot the number of changes. @juliusv  WDYT?

juliusv: Yeah, but I would very much like to avoid global state whereever possible (that's why I only use the global path prefix here in the `index.tsx` and then never again). Passing in dependencies into components is more verbose, but in exchange we get unambiguous explicitness, with dependencies enumerated as properties to components.

boyskila: Another approach can be to pass the prefix(explicitly) as Link state prop. Then prefix is available through props in each component ruled by Reach router

juliusv: Chatted with @boyskila to find out what was meant here. https://reach.tech/router/api/Link provides a way to attach location state to the Reach router, but it doesn't seem to be quite for this kind of use case.

csmarchbanks: I also would like to avoid global state whenever possible, and like to see dependencies clearly defined in the props.

web/ui/react-app/src/App.test.tsx
```diff
@@ -7,15 +7,17 @@ import { Router } from '@reach/router';
...
   it('navigates', () => {
     expect(app.find(Navigation)).toHaveLength(1);
   });
   it('routes', () => {
-    [Alerts, Config, Flags, Rules, Services, Status, Targets, PanelList].forEach(component =>
-      expect(app.find(component)).toHaveLength(1)
-    );
+    [Alerts, Config, Flags, Rules, Services, Status, Targets, PanelList].forEach(component => {
+      const c = app.find(component);
+      expect(c).toHaveLength(1);
```
csmarchbanks: Is this assertion needed? Seems like just checking that the prefix is correct is enough, and will not cause the test to be changed if any pages need more props in the future.

web/ui/react-app/src/pages/Status.tsx
```diff
@@ -54,8 +55,18 @@ export const statusConfig: StatusConfig = {
...
-  const { response: data, error, isLoading } = useFetches<StatusPageState[]>(ENDPOINTS);
+const endpointsMemo: { [prefix: string]: string[] } = {};
+
+const Status: FC<RouteComponentProps & PathPrefixProps> = ({ pathPrefix = '' }) => {
+  if (!endpointsMemo[pathPrefix]) {
+    // TODO: REMOVE THIS SUPER UGLY HACK IMMEDIATELY.
+    // The problem is that there's an infinite reload loop if the endpoints array is
+    // reconstructed on every render, as the dependency checking in useFetches()
+    // then thinks that something has changed... the whole useFetches() should
+    // probably removed and solved differently (within the component?) somehow.
```
csmarchbanks: Just curious on your thoughts to why it should be removed, too many small unexpected edge cases?

web/ui/react-app/src/pages/Status.tsx
```diff
@@ -73,36 +84,40 @@ const Status = () => {
...
+                      );
+                    })}
+                  </tbody>
+                </Table>
+              </Fragment>
+            );
+          })
+        : null}
+    </>
+  );
```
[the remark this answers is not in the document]
juliusv: `data.map()` returns an array of React elements, which TypeScript doesn't allow for function components (the function wasn't properly typed before). I thought TypeScript also didn't allow returning nulls from function components (the TS error message was... let's say hard to read), but this is not actually the case, only the `data.map()` portion was the problem.

I changed it now so that the fragment only encloses the array, not the null.

web/ui/react-app/src/pages/Status.tsx
juliusv: I think I went overboard with my comment here. It's not that bad, but basically in the actual app (vs. tests etc.), there's only ever one path prefix throughout the lifetime of the application, and thus these endpoints are basically constants. Caching them in a global mutable object with provisions for different path prefixes seems like more than should be needed, except that constructing them each time on the fly in the render function causes infinite render loops due to  `useEffect()` array dependency comparisons to fail.

web/ui/react-app/src/App.test.tsx
juliusv: This just tests that there is exactly one component (Alerts, Config, ...) of this type, it doesn't check the component's props.

juliusv:
Merging for now due to the general :+1: (want to build ontop of it), but happy to change things if there are followup issues.

web/ui/react-app/src/pages/Status.tsx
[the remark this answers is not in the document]
boyskila: ```const Status = ({ pathPrefix = '' }: RouteComponentProps & PathPrefixProps) => {```
I think this will eliminate fragments and the component will remain properly typed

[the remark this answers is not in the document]
juliusv: Then I get the following error in `App.tsx` because it's not fully typed as a function component anymore:

`JSX element type 'Element | Element[] | null' is not a constructor function for JSX elements.
  Type 'Element[]' is missing the following properties from type 'Element': type, props, key`

I think adding the one extra `<></>` is fine though, and with `FC<...>` it's clearer what it is.

Removing `data ?` I get `Object is possibly 'undefined'.`. I could instead do `data!`, but not sure if that's nicer.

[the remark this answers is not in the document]
csmarchbanks: I think the extra `<></>` is fine, I just wanted to understand the need for it and now I do, thanks!

PR #6267, opened 2019-11-04
Replace fetching hooks with class render prop component

boyskila:
- [x] implement fetcher component
- [x] replace fetching hooks
- [x] unit tests

@juliusv  @csmarchbanks @hooten


web/ui/react-app/src/api/Fetcher.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  handleResponse = async (cb: () => Promise<any>) => {
+    try {
+      this.setState({ data: await cb() });
+    } catch (error) {
+      this.setState({ error });
+    }
+  };
+  get = (options: RequestInit = {}) => async (url: string) => {
```
juliusv: nit: add empty line above `get`

web/ui/react-app/src/pages/Config.tsx
```diff
@@ -4,43 +4,47 @@ import { Alert, Button } from 'reactstrap';
...
-            setCopied(result);
-            setTimeout(setCopied, 1500);
-          }}
-        >
-          <Button color="light" disabled={!config}>
-            {copied ? 'Copied' : 'Copy to clipboard'}
-          </Button>
-        </CopyToClipboard>
-      </h2>
+    <Fetcher url={`${pathPrefix}/api/v1/status/config`}>
```
juliusv: Extracting the children into their own component would make it even easier to test the display logic separately from the fetching, but maybe that's for a second step.

web/ui/react-app/src/api/Fetcher.tsx
```diff
@@ -0,0 +1,68 @@
+import { Component } from 'react';
+
+export interface FetcherState<T> {
```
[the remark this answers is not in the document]
boyskila: Yes, yes! Sounds better to me as well, thanks!

web/ui/react-app/src/api/Fetcher.tsx
```diff
@@ -0,0 +1,68 @@
...
+      this.handleResponse(async () => await Promise.all(urls.map(this.get(options)).filter(Boolean)));
+    } else if (url) {
+      this.handleResponse(() => this.get(options)(url));
+    } else {
+      throw new Error('URL/s is Missing');
+    }
+  }
+
+  componentDidUpdate(nextProps: FetcherProps) {
+    const { url: nextURL, urls: nextURLS, options } = nextProps;
```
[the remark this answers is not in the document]
boyskila: Hm, typo -  it supposes to be nextURLs. Will change it, thanks!

web/ui/react-app/src/api/Fetcher.tsx
```diff
@@ -0,0 +1,68 @@
...
+    const { url, urls } = this.props;
+    if (nextURLS && !urlsEqual(urls, nextURLS)) {
+      this.handleResponse(async () => await Promise.all(nextURLS.map(this.get(options)).filter(Boolean)));
```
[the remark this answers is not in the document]
boyskila: If for some reason undefined or null appears, filter(Boolean) will remove them.
"Wouldn't one always expect to get back data for each URL" - yes, I guess so but I want to be sure that nothing unexpected will happen after that.

web/ui/react-app/src/api/Fetcher.tsx
boyskila: Maybe Eslint can help here....will check, thanks!

web/ui/react-app/src/pages/Config.tsx
boyskila: Good point. I'm gonna think of how this can be done without introducing one more layer of components. Maybe a local functional component..

web/ui/react-app/src/pages/Status.tsx
```diff
@@ -55,26 +55,14 @@ export const statusConfig: StatusConfig = {
...
-    // TODO: Come up with a nicer solution for this?
-    //
-    // The problem is that there's an infinite reload loop if the endpoints array is
-    // reconstructed on every render, as the dependency checking in useFetches()
-    // then thinks that something has changed... the whole useFetches() should
-    // probably removed and solved differently (within the component?) somehow.
-    endpointsMemo[pathPrefix] = ENDPOINTS.map(ep => `${pathPrefix}${ep}`);
-  }
-  const { response: data, error, isLoading } = useFetches<StatusPageState[]>(endpointsMemo[pathPrefix]);
+export const statusRenderer = ({ data, error }: FetchState<StatusPageState>) => {
```
juliusv: I'd call this either `renderStatus`, noun names seem to be more for classes or at least full components?

web/ui/react-app/src/pages/Status.tsx
```diff
@@ -116,7 +104,9 @@ const Status: FC<RouteComponentProps & PathPrefixProps> = ({ pathPrefix = '' })
         );
       })}
     </>
-  ) : null;
+  );
 };
 
-export default Status;
+export default ({ pathPrefix = '' }: RouteComponentProps & PathPrefixProps) => (
+  <Fetch urls={endpoints.map(e => `${pathPrefix}${e}`)}>{statusRenderer}</Fetch>
```
juliusv: Nice.

web/ui/react-app/src/pages/Status.tsx
boyskila: It has a strange status - it is a function (when is used as a children prop in Fetch) but can also act as a Component (see unit tests). Maybe I should unify it and rename it as "Status" as it was before. WDYT?

web/ui/react-app/src/pages/Status.tsx
boyskila: I'm accepting this as 👍 to refactor other components as well :)

web/ui/react-app/src/pages/Status.tsx
juliusv: With unify, do you mean merging it together again so it can't be used separately? I do like having it separately, but I think either making it a full separate component or just calling it `status` would be fine? Not sure...

boyskila: Yes, naming it in such a way to become clear that is one thing.

web/ui/react-app/src/pages/Config.tsx
```diff
@@ -42,6 +43,12 @@ const Config: FC<RouteComponentProps & PathPrefixProps> = ({ pathPrefix }) => {
...
+}
 
-export default Config;
+export default ({ pathPrefix }: RouteComponentProps & PathPrefixProps) => {
+  return (
+    <Fetch url={`${pathPrefix}/api/v1/status/config`}>
+      {(state: FetchState<YamlConfig>) => <Config {...state} />}
+    </Fetch>
+  )
+}
```
juliusv: Hm, does the linter not catch missing end-of-file newlines yet?

boyskila: Linter is fine, I just forgot to "npm run lint" before commit

web/ui/react-app/src/pages/Flags.tsx
```diff
@@ -1,51 +1,41 @@
...
-  };
-
-  return (
-    <>
-      <h2>Command-Line Flags</h2>
-      {body()}
-    </>
-  );
-};
+export const Flags = ({ error, data: flags }: FetchState<FlagMap>) => {
```
juliusv: Thinking more, I see value in being as explicit as possible and just making it a full functional component, like:

```suggestion
export const Flags: FC<FetchState<FlagMap>> = ({ error, data: flags }) => {
```

Fewer linter warnings about undefined return values, and clearer to the reader that it is (or can be used) as a component.

I'd do the same typing below wrapped component, though I don't know how you can do the `FC<...>` typing inline after `export default`. But assigning it to another name first is cool too.

For naming, maybe `RenderFlags` and for the exported one, just `Flags`?

boyskila:
I have some concerns about StatusIndicator component but decided to implement it anyway. The goal is to make the components clean as possible and to make them responsible only for one thing - all suffixed as "Content" are just data renderers. Error Alert and Loader are displaying what the current status of the response is but I'm not sure whether is OK to act like a guard of its children. So any feedback will be useful @juliusv @hooten @csmarchbanks

web/ui/react-app/src/pages/Status.tsx
```diff
@@ -116,7 +88,19 @@ const Status: FC<RouteComponentProps & PathPrefixProps> = ({ pathPrefix = '' })
...
     </>
-  ) : null;
+  );
 };
 
+const Status: FC<RouteComponentProps & PathPrefixProps> = ({ pathPrefix = '' }) => (
+  <Fetch urls={endpoints.map(e => `${pathPrefix}${e}`)}>
+    {({ error, data }) => {
+      return (
+        <StatusIndicator error={error && `Error fetching status: ${error.message}`} hasData={data && data.length}>
```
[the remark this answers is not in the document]
boyskila: Thanks for the idea. Such a component will be really useful.

hooten:
Nice work. I have some concerns about this pattern though. I think that creating another layer of abstraction over `useFetch` is putting the cart before the horse, in a sense. I don't know if there are enough cases of data fetching in this react app to justify the creation of this pattern. There might be other cases in the future that do a variety of fetches of partial data. There might be other cases we haven't imagined yet. It might be wise to wait until there are more cases that we know about before creating a pattern we want everyone to follow. I think the hooks pattern is great for now

hooten:
Nice work. I have some concerns about this pattern though. I think that creating another layer of abstraction over `useFetch` is putting the cart before the horse, in a sense. I don't know if there are enough cases of data fetching in this react app to justify the creation of this pattern. There might be other cases in the future that do a variety of fetches of partial data. There might be other cases we haven't imagined yet. It might be wise to wait until there are more cases that we know about before creating a pattern we want everyone to follow. I think the hooks pattern is great for now

boyskila:
@hooten thank you for your feedback! Actually there is no abstraction over useFetch/es. This is a completely new approach which aims to replace both hooks and to unify fetching. I think so far we have enough cases that require a common approach (at least 3 for now ) and I guess a few more will appear so the ```fetchWithStatus``` pattern looks it can be used extensively.
About the future cases - You are absolutely right that they might appear and that's why we should try to make the API extensible.

juliusv:
Yeah, I think there's a couple of cases where this is helpfully applicable already, but @hooten in case you see complexity or design problems with it, please let us know!

web/ui/react-app/src/api/FetchWithStatus.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { FC, ComponentType } from 'react';
+import { Fetch, FetchState } from '../api/Fetch';
+import { StatusIndicator } from '../StatusIndicator';
+import PathPrefixProps from '../PathPrefixProps';
+
+export const fetchWithStatus = <T extends {}, DataType extends {}>(
```
juliusv: Hm, why not `fetchWithStatus` -> `FetchWithStatus` like other components?

...and I'm still trying to understand the crazy TypeScript signature :laughing:

juliusv: Ah ok, because it's a HOC, which explains the casing and the signature...

boyskila: Yeah, so by convention in React HOC names are starting with the prefix 'with' and with lower case. In this situation, I omit the prefix since it would sound strange ```withFetchWithStatus```

juliusv: I wonder though, why a HOC and not a normal component that just uses `<Fetch>`, `<StatusIndicator>`, and the content (as children)?

boyskila:
```
const FetchWithStatus = ({ children }) => {
  return (
    <Fetch>
      {({ err, data }) => {
        return (
          <StatusIndicator>
            {children} // the issue is here - no way to pass the data to the children
          </StatusIndicator>
        )
      }}
    </Fetch>
  )
}
```
Basically there is a way to make it work. If I pass the data to StatusIndicator and StatusIndicator to pass it back to children(data) which is odd since children have access to data already in the parent scope....too much complexity...maybe there is another approach which I cannot see

hooten:
Good point that this might unify fetching. But I personally find the `useFetch` hook to be simpler to maintain and easier to reason about.

web/ui/react-app/src/api/Fetch.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { fetchAPI } from './FetchAPI';
+
+export interface FetchState<T> {
+  data?: T;
+  error?: Error;
+}
+
+interface FetchProps {
+  url?: string;
+  urls?: string[];
```
hooten: I think this should be more strongly typed. Could you just have a `urls` field? People who just want to fetch data at one url can just pass in `[url]`

web/ui/react-app/src/api/Fetch.tsx
```diff
@@ -0,0 +1,75 @@
...
+  options?: RequestInit;
+  children: (state: any) => any;
```
hooten: Any way to make `any` more strongly typed? I'm thinking about how I might use this as a client. `any` tends to obscure usage for me.

web/ui/react-app/src/api/Fetch.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export class Fetch extends Component<FetchProps, FetchState<any>> {
+  state = {
+    data: undefined,
+    error: undefined,
+  };
+
+  componentDidMount() {
+    const { url, urls, options } = this.props;
+    if (Boolean(url) && Boolean(urls)) {
```
hooten: Stronger typing would make this check unnecessary

web/ui/react-app/src/api/Fetch.tsx
boyskila: I think no..any data can be passed to the consumer. I mean I don't know whether the type can be predicted. Any ideas?

web/ui/react-app/src/api/Fetch.tsx
hooten: and the ones below

web/ui/react-app/src/api/Fetch.tsx
```diff
@@ -0,0 +1,75 @@
...
+      this.setState({ data: await resultCb() });
+    } catch (error) {
+      this.setState({ error });
+    }
+  };
+
+  render() {
+    return this.props.children(this.state);
+  }
+}
```
hooten: why not use hooks in this component? the component lifecycle methods you have are the ones a hook can replace

web/ui/react-app/src/api/Fetch.tsx
boyskila: They have types ```url: string urls: string[]```, I'm just checking here whether user not passing both of them by mistake

hooten: There should not be two ways to pass in urls. There should only be a `urls: string[]` field. Clients with one url can just pass it in as a single-element array.

web/ui/react-app/src/api/Fetch.tsx
hooten: You could do something like 
```
interface FetchProps<T, S> {
   ...
   children: (state: S) => T;
}

or something like that

File: web/ui/react-app/src/api/Fetch.tsx (hunk 15 above)
Comment (boyskila) [reply web/ui/react-app/src/api/Fetch.tsx]: In hooks you really don't have a mechanism to compare dependency array values (like ```this.props.foo === nextProps.foo```) which is needed when you expect new instances every time but not changes in the data

Comment (hooten) [reply web/ui/react-app/src/api/Fetch.tsx]: How often do you expect these urls to change? I guess I'm having a hard time understanding how I would use this mechanism

File: web/ui/react-app/src/api/Fetch.tsx (hunk 14 above)
Comment (boyskila) [reply web/ui/react-app/src/api/Fetch.tsx]: Yes, I already think about this and I came to a conclusion that is non-intuitive. Also, it brings more complexity in the code

File: web/ui/react-app/src/api/Fetch.tsx (hunk 15 above)
Comment (hooten) [reply web/ui/react-app/src/api/Fetch.tsx]: Actually i think i understand now... eg if params for a fetch change

File: web/ui/react-app/src/api/Fetch.tsx (hunk 14 above)
Comment (hooten) [reply web/ui/react-app/src/api/Fetch.tsx]: I disagree

File: web/ui/react-app/src/api/Fetch.tsx (hunk 13 above)
Comment (boyskila) [reply web/ui/react-app/src/api/Fetch.tsx]: How can I use this here ```...extends Component<FetchProps<???>>?

File: web/ui/react-app/src/api/Fetch.tsx (hunk 15 above)
Comment (hooten) [reply web/ui/react-app/src/api/Fetch.tsx]: hooks return a setState for managing the url, which can be used to re-trigger an effect

File: web/ui/react-app/src/api/Fetch.tsx (hunk 13 above)
Comment (hooten) [reply web/ui/react-app/src/api/Fetch.tsx]: Yeah I'm not totally sure either. No worries if it's too much of a pain. Would be cool if you can figure it out

File: web/ui/react-app/src/api/Fetch.tsx (hunk 14 above)
Comment (boyskila) [reply web/ui/react-app/src/api/Fetch.tsx]: From user perspective or dev?

File: web/ui/react-app/src/api/Fetch.tsx (hunk 13 above)
Comment (boyskila) [reply web/ui/react-app/src/api/Fetch.tsx]: I'll try, thanks!

File: web/ui/react-app/src/api/Fetch.tsx (hunk 14 above)
Comment (hooten) [reply web/ui/react-app/src/api/Fetch.tsx]: Both, i think. As a dev, I have to maintain two options. As a user, I still have to read the code to understand what i'm supposed to pass in since there are two similar options.

Comment (boyskila) [reply web/ui/react-app/src/api/Fetch.tsx]: From a dev perspective, you should maintain  both options again since you don' t want to execute ```Promise.all``` logic for single URL. You have to check again the array whether it has a single element an if not - take ```Promise.all``` branch otherwise take the other direction...
From a user perspective I guess it would be easier, yes

File: web/ui/react-app/src/api/Fetch.tsx (hunk 15 above)
Comment (boyskila) [reply web/ui/react-app/src/api/Fetch.tsx]: The issue is not with the dependencies updates (url, options). The issue is when the Effect is triggered without those values to be changed and this is happening when on every render new instance is passed of those values. To overcome this problem you have to implement logic similar to the one which class components gives you anyway.

Comment (boyskila) [reply web/ui/react-app/src/api/Fetch.tsx]: Any ideas on how I can improve this? Tried @hooten suggestion but I'm a bit unhappy with that the user is forced to pass array for a single request. On the other side, @hooten is right that a single prop will be less confusing. I can just define it as ```endpoint: string | string[]``` but this will turn into the same situation because of which we split useFetches into two hooks. WDYT? @juliusv @csmarchbanks

Comment (juliusv) [reply web/ui/react-app/src/api/Fetch.tsx]: @boyskila Was the last comment meant to be posted on the other sub-thread https://github.com/prometheus/prometheus/pull/6267#discussion_r342867962?

Both options (always passing an array vs. having two ways of passing URLs) seem equally bad to me, but I don't see a better way, unless we get rid of the multi-fetch functionality completely. So maybe we just have to live with the ugliness unless someone has another design idea.

Comment (boyskila) [reply web/ui/react-app/src/api/Fetch.tsx]: Oh, my bad. It is for https://github.com/prometheus/prometheus/pull/6267#discussion_r342867962

Comment (hooten) [reply web/ui/react-app/src/api/Fetch.tsx]: I think I agree with @juliusv that both options are less than ideal. Maybe it would be better to remove the multi-fetch functionality. It might be an anti-pattern to wait for multiple remote calls to return before beginning rendering.

@csmarchbanks do you have any thought on this (since you wrote the previous fetch code)?

Comment (boyskila) [reply web/ui/react-app/src/api/Fetch.tsx]: Other option could be to create a new route in the api.go to handle all the information at once
``` /stats ``` and thus we will not need multi fetch anymore

File: web/ui/react-app/src/api/Fetcher.tsx (hunk 5 above)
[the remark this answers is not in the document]
Comment (csmarchbanks) [reply web/ui/react-app/src/api/Fetcher.tsx]: Does this mean that if for some reason an api returns a falsey value then we just silently discard the result and therefore change the ordering of the results? If so that seems very bad.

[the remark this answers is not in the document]
Comment (boyskila) [reply web/ui/react-app/src/api/Fetcher.tsx]: Yes, maybe it's good the consumer of the response to be aware of that. Will remove it from here and will check into component instead, thanks!

File: web/ui/react-app/src/api/Fetch.tsx (hunk 15 above)
Comment (csmarchbanks) [reply web/ui/react-app/src/api/Fetch.tsx]: I have been giving the multi-fetch functionality some thought over the last few days and generally think it should go away for the following reasons.
1. IMO Lumping all requests together and returning an error if any of them errors is an anti-pattern. Errors from each call should be handled separately with as much information exposed to the user as possible from any calls that do succeed.
1. The function is complicated (I am not convinced there aren't unexpected edge cases that will bite us) and only used in one spot. Doing three `useFetch` calls is not much more code, and will be much more readable in the future. No need for a new route either.
1. What about passing extra options such as if doing multiple POST requests? Do we now need to support arrays of both urls and options? Or some wrapping object?

At that point, this component becomes of dubious value, perhaps FetchWithStatus is still useful to reduce a little bit of code. The Go developer inside of me still wants to make the page explicitly handle errors though.

Comment (boyskila) [reply web/ui/react-app/src/api/Fetch.tsx]: @csmarchbanks Thank you for your feedback!

For point 1 I would say that it depends on the requirements.
In the Status page, I want to show the error message if some of the 3 requests fail but probably not in such a way. You are right that it is good to pass all the information back to the user and the user to decide how to handle the errors. So, for example, if we fetch ```[A, B, C]``` and the response is ```[success, fail, success]``` is good to return back the response objects instead of the data only.

For point 2:  This is a basic implementation covering the needs of the app. I'm sure that I'm far from what it should be to cover all edge cases and scenarios. Not sure how the three useFetches will handle the requirements for the status page?

For point 3: Currently, you can pass only one ```RequestInit``` object which will not work for multiple requests as you mention but I guess it can be easily implemented just by sending not ```string[]``` but ```{url: string, opt: RequestInit}[]```

Can you please explain why do you think the function is complicated? IMO, at least for now, is not doing so much  :)

Comment (csmarchbanks) [reply web/ui/react-app/src/api/Fetch.tsx]: Here is how the status page could use three fetches to accomplish exactly what `useFetches` does today: https://github.com/csmarchbanks/prometheus/commit/a5fd258404520e51c3664631153c183490426e0f

As far as complexity goes: 
Currently in master the `useFetches` function is not too complicated - but also requires the hacky workaround for dependencies. In this PR you have to create an entire component just to fetch some data which seems like a lot, especially for components that have a different lifecycle than just a page load.

Comment (boyskila) [reply web/ui/react-app/src/api/Fetch.tsx]: Thanks for the suggestion, I like the approach.

File: web/ui/react-app/src/pages/Config.tsx [hunk 16]
```diff
@@ -1,47 +1,53 @@
...
+interface ConfigContentProps {
+  error?: Error;
+  data?: YamlConfig;
+}
+
+const YamlContent = ({ yaml }: YamlConfig) => <pre className="config-yaml">{yaml}</pre>;
+YamlContent.displayName = 'Config';
 
-  const config = response && response.data.yaml;
+const ConfigWithResponseIndicator = withStatusIndicator(YamlContent);
```
juliusv: I'd call it `ConfigWithStatusIndicator` for consistency and because it's clearer. Same for the other pages.

web/ui/react-app/src/pages/Config.tsx
```diff
@@ -1,47 +1,53 @@
...
+
+export const ConfigContent: FC<ConfigContentProps> = ({ error, data }) => {
+  const [copied, setCopied] = useState(false);
+  const config = data && data.yaml ? data.yaml : '';
+  const hasConfig = Boolean(config && config.length);
```
juliusv: Can the config ever be legitimately empty when it's not undefined? I.e. can't this just be `hasConfig = config !== undefined`, which would avoid the bit dirty `Boolean` cast as well?

boyskila: Is it possible the yaml itself to be an empty string?

juliusv: No, the YAML content as rendered by the server will never be empty.

juliusv: (and if it could be, I guess we would still just want to display an empty box)

juliusv:
@boyskila Is this ready to be merged now from your side (just because it still has WIP in the title)?

boyskila:
@juliusv Yes, I think it's ready, although the test coverage is not the desired (we can postpone it for another PR if you agree).

juliusv:
Thanks, I think for now it's ok to merge then :)

PR #6270, opened 2019-11-04
Fix Promtool showing false duplicate rule warnings

### LINKED ISSUE #6109: In Prometheus 2.13.0, promtool misleadingly complains about rules that reuse the same label values
## Bug Report

In Prometheus 2.13.0, `promtool` will complain if you have rules with the same label and the same label value, which is a routine occurrence if you're using per-rule labels to control various things or add additional information. On top of that, the complaint is misleading; it reports a duplicate rule, instead of duplicate labels. Sample output (for the example rule file in this bug report):

````
$ promtool check rules testrule.yaml 
Checking testrule.yaml
1 duplicate rules(s) found.
Metric: 
Label(s):
        dest: email-low
Might cause inconsistency while recording expressions.
  SUCCESS: 2 rules found
````

In our live configuration, we only have alerting rules, not recording rules, and this reports for them. This may be part of the bug, since alerting rules do not have a metric and the code appears to be comparing only the metric name, not the metric name and the alert name. Possibly cmd/promtool/main.go should include `alert` in the `compareRuleType` struct it uses to check for duplication.

(I'm assuming here that duplicate `alert` values are at least bad form and worth `promtool` warning about. Perhaps it should have a specific check and warning for that.)

If this check is supposed to apply only to recording rules, the obvious change might be to skip any rule that has a blank `Record` field or perhaps a non-blank `Alert` field.

**Environment**

* System information:

	Linux 4.15.0-52-generic x86_64

* Prometheus version:

	prometheus, version 2.13.0 (branch: HEAD, revision: 6ea4252299f542669aca11860abc2192bdc7bede)
          build user:       root@f30bdad2c3fd
          build date:       20191004-11:25:34
          go version:       go1.13.1

* Example rules file:
```
groups:
- name: show-bug
  rules:
  - alert: Test1
    expr: (node_time_seconds - node_boot_time_seconds) < 60
    labels:
      dest: email-low
  - alert: Test2
    expr: node_load15 > 0.5
    labels:
      dest: email-low
```

### PR DESCRIPTION
**Pull request by csmarchbanks**
Alert rules do not use the Record field, so any alerts with the same
labels and different names would be counted as being duplicates.
Promtool will now consider either field when finding duplicates.

Closes #6109

### DIFF
--- a/cmd/promtool/main.go
+++ b/cmd/promtool/main.go
@@ -324,19 +324,19 @@ type compareRuleType struct {
 	label  map[string]string
 }
 
-func checkDuplicates(r []rulefmt.RuleGroup) []compareRuleType {
+func checkDuplicates(groups []rulefmt.RuleGroup) []compareRuleType {
 	var duplicates []compareRuleType
 
-	for rindex := range r {
-		for index, props := range r[rindex].Rules {
+	for _, group := range groups {
+		for index, rule := range group.Rules {
 			inst := compareRuleType{
-				metric: props.Record,
-				label:  props.Labels,
+				metric: ruleMetric(rule),
+				label:  rule.Labels,
 			}
 			for i := 0; i < index; i++ {
 				t := compareRuleType{
-					metric: r[rindex].Rules[i].Record,
-					label:  r[rindex].Rules[i].Labels,
+					metric: ruleMetric(group.Rules[i]),
+					label:  group.Rules[i].Labels,
 				}
 				if reflect.DeepEqual(t, inst) {
 					duplicates = append(duplicates, t)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cmd/promtool/main.go [hunk 1]
```diff
@@ -324,21 +324,22 @@ type compareRuleType struct {
...
 			}
 			for i := 0; i < index; i++ {
 				t := compareRuleType{
-					metric: r[rindex].Rules[i].Record,
-					label:  r[rindex].Rules[i].Labels,
+					metric: ruleMetric(group.Rules[i]),
+					label:  group.Rules[i].Labels,
 				}
 				if reflect.DeepEqual(t, inst) {
+					fmt.Println(t, inst)
```
flyhard: Maybe this should contain more info what this is or maybe removed if it is not needed anymore.

brian-brazil:
:+1:

PR #6274, opened 2019-11-05
React UI: Display small numbers correctly

### LINKED ISSUE #6272: React UI: Regression over legend and small values
## Proposal

This is a regression compared to the current UI

*“Nice to have” is not a good use case. :)*

## Bug Report
**What did you do?**

Run go_gc_duration_seconds

legend-old

legend-new

Legend says 0.00, 0.00 , 0.00 ..  instead of 0.003, 0.004, which is strange because it also displays 1.00m, so I would expect to see 3.00m...

**Environment**

* System information:

	linux

* Prometheus version:

prometheus, version 2.13.1 (branch: master, revision: e94503ff5c412590ce7616accdd3c62a2189bcd3)
  build user:       roidelapluie@oxygen
  build date:       20191105-10:36:21
  go version:       go1.13.1

* node version:
v13.0.1

roidelapluie:
Closes #6272

nl

### DIFF
--- a/web/ui/react-app/src/Graph.test.tsx
+++ b/web/ui/react-app/src/Graph.test.tsx
@@ -101,5 +101,64 @@ describe('Graph', () => {
       expect(innerdiv).toHaveLength(1);
       expect(legend).toHaveLength(1);
     });
+    it('formats tick values correctly', () => {
+      const graph = new Graph();
+      [
+        { input: 2e24, output: '2.00Y' },
+        { input: 2e23, output: '200.00Z' },
+        { input: 2e22, output: '20.00Z' },
+        { input: 2e21, output: '2.00Z' },
+        { input: 2e19, output: '20.00E' },
+        { input: 2e18, output: '2.00E' },
+        { input: 2e17, output: '200.00P' },
+        { input: 2e16, output: '20.00P' },
+        { input: 2e15, output: '2.00P' },
+        { input: 1e15, output: '1.00P' },
+        { input: 2e14, output: '200.00T' },
+        { input: 2e13, output: '20.00T' },
+        { input: 2e12, output: '2.00T' },
+        { input: 2e11, output: '200.00G' },
+        { input: 2e10, output: '20.00G' },
+        { input: 2e9, output: '2.00G' },
+        { input: 2e8, output: '200.00M' },
+        { input: 2e7, output: '20.00M' },
+        { input: 2e6, output: '2.00M' },
+        { input: 2e5, output: '200.00k' },
+        { input: 2e4, output: '20.00k' },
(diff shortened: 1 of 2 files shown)


hooten:
Could be cool if there were a unit test for this function

roidelapluie:
Indeed! There is more to fix! wait for it :)

roidelapluie:
@juliusv @hooten this is ready for review.

roidelapluie:
Screnshots
before:
o1

after:
o2

web/ui/react-app/src/Graph.test.tsx
```diff
@@ -101,5 +101,60 @@ describe('Graph', () => {
...
+      expect(graph.formatValue(2e-18)).toBe('2.00a');
+      expect(graph.formatValue(2e-19)).toBe('0.20a');
+      expect(graph.formatValue(2e-20)).toBe('0.02a');
+      expect(graph.formatValue(1e-21)).toBe('1.00z');
+      expect(graph.formatValue(2e-21)).toBe('2.00z');
+      expect(graph.formatValue(2e-22)).toBe('0.20z');
+      expect(graph.formatValue(2e-23)).toBe('0.02z');
+      expect(graph.formatValue(2e-24)).toBe('2.00y');
+      expect(graph.formatValue(2e-25)).toBe('0.20y');
+      expect(graph.formatValue(2e-26)).toBe('0.02y');
```
juliusv: How about putting the key/value pairs into an object and then mapping over them?

roidelapluie: Indeed, I have done the change.

web/ui/react-app/src/Graph.test.tsx
```diff
@@ -101,5 +101,64 @@ describe('Graph', () => {
...
+        { input: 2e-20, output: '0.02a' },
+        { input: 1e-21, output: '1.00z' },
+        { input: 2e-21, output: '2.00z' },
+        { input: 2e-22, output: '0.20z' },
+        { input: 2e-23, output: '0.02z' },
+        { input: 2e-24, output: '2.00y' },
+        { input: 2e-25, output: '0.20y' },
+        { input: 2e-26, output: '0.02y' },
+      ].map(function(t) {
+        expect(graph.formatValue(t.input)).toBe(t.output);
```
juliusv: Heh, I thought about just using the input as the key and the output as the value, but explicit `input` / `output` properties work too :)

juliusv:
:+1: thanks!

web/ui/react-app/src/Graph.test.tsx
roidelapluie: As in input is not string I made keys.

PR #6276, opened 2019-11-05
React UI: Implement /targets page

### LINKED ISSUE #6269: React UI: Build targets page
This is a more specific issue filed for the discussion started on https://github.com/prometheus/prometheus/issues/6164#issuecomment-549375737.

Let's discuss targets-page specific things here.

hooten:
This PR implements the `/targets` page in React. It ports the existing page without any UI design changes. 

It also extends the `/api/v1/targets` api by adding two fields:
  + `scrapePool` - the name of the pool (e.g, `"blackbox"`) grouping several targets together
  + `lastScrapeDuration` - the duration in nanoseconds of the last scrape
and by adding a query parameter:
  + `state` - `active` or `dropped` (or `any`), which limits the api response payload to `activeTargets` or `droppedTargets`, respectively. The default behavior is still to return both. This will go a long way in improving render time of the page, which does not require `droppedTargets` at this point.

Fixes #6269


hooten:
@juliusv @csmarchbanks @boyskila I'd appreciate a review when you get a chance. Thanks in advance!

web/ui/react-app/src/ExpressionInput.tsx
```diff
@@ -97,7 +97,7 @@ class ExpressionInput extends Component<ExpressionInputProps, ExpressionInputSta
                           },
                         });
                         return (
-                          <SanitizeHTML tag="li" {...itemProps} allowedTags={['strong']}>
+                          <SanitizeHTML key={string} tag="li" {...itemProps} allowedTags={['strong']}>
```
juliusv: This fix seems unrelated, could you put it into a separate PR?

web/ui/react-app/src/Loader.tsx
```diff
@@ -0,0 +1,7 @@
+import React, { FC } from 'react';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faSpinner } from '@fortawesome/free-solid-svg-icons';
+
+const Loader: FC<{}> = () => <FontAwesomeIcon icon={faSpinner} spin />;
```
juliusv: @boyskila is refactoring these fetching / status indicator components in https://github.com/prometheus/prometheus/pull/6267. Probably it makes sense for that one to be merged first (feel free to leave input there!) and then adjust this here accordingly?

web/ui/react-app/src/pages/Targets.tsx
```diff
@@ -1,7 +0,0 @@
-import React, { FC } from 'react';
```
juliusv: If we're moving components used by the targets page into a subdirectory, we'll also want to move other components in the main `/src` directory into their respective page / utils / etc. directories, but let's not do the latter in this PR yet.

web/ui/react-app/src/pages/targets/EndpointLink.tsx
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  const { host, pathname, protocol, searchParams }: URL = url;
+  const params = Array.from(searchParams.entries());
+
+  return (
+    <>
+      <a href={endpoint}>{`${protocol}//${host}${pathname}`}</a>
+      {params.length > 0 ? <br /> : null}
+      {params.map(([labelKey, labelValue]: [string, string]) => {
```
juliusv: If you're going for `labelKey` / `labelValue`, then the Prometheus terminology for the key would be label name, thus `labelName`.

web/ui/react-app/src/pages/targets/SeriesLabels.test.tsx
```diff
@@ -0,0 +1,52 @@
+/* eslint @typescript-eslint/camelcase: 0 */
```
juliusv: I assume this is for `__metrics_path__`. Would quoting that object key (`"__metrics_path__": ...`) avoid the linter problem without turning that check off completely?

web/ui/react-app/src/pages/targets/Filter.module.css
```diff
@@ -0,0 +1,5 @@
+.btn {
```
juliusv: You realize that by introducing CSS modules as yet another way to do CSS in this project, you just volunteered yourself for converting everything else (that is applicable) to CSS modules as well :P

web/ui/react-app/src/pages/targets/Filter.module.css
```diff
@@ -0,0 +1,5 @@
...
+  border-radius: 0px !important;
+  margin-top: 0.6em;
+  margin-bottom: 15px;
+}
```
juliusv: Btw. I don't think we should keep the styling of the old UI's buttons verbatim. Let's just use the standard Bootstrap style (rounded corners and all)?

web/ui/react-app/src/Loader.tsx
hooten: I have some concerns about the Fetch component. I'll bring them up in the other PR review. I would prefer to keep `useFetch`

web/ui/react-app/src/pages/Targets.tsx
hooten: Cool. I think it would be nice to have page-specific components in the `src/pages/<page>/` directory and shared components in `src/components`

web/ui/react-app/src/pages/targets/EndpointLink.tsx
hooten: will do

web/ui/react-app/src/pages/targets/JobDetails.tsx
```diff
@@ -0,0 +1,52 @@
...
+              <td className={styles.endpoint}>
+                <EndpointLink endpoint={scrapeUrl} />
+              </td>
+              <td className={styles.state}>
+                <StateIndicator health={health} />
+              </td>
+              <td className={styles.labels}>
+                <SeriesLabels discoveredLabels={discoveredLabels} labels={labels} />
+              </td>
+              <td className={styles['last-scrape']}>{formatRelative(lastScrape, now())}</td>
```
juliusv: `now()` is cool for now (pun intended), though it's possible that the user's clock is totally off from the server's clock and these durations will be totally wrong. Long run we could maybe get the server's reference "now" timestamp via the `Date` header that Prometheus returns on each API call (just a wild idea without having to add more infos to APIs). But forget that for now :)

web/ui/react-app/src/pages/targets/SeriesLabels.test.tsx
hooten: Unfortunately, it does not. I can disable eslint for the line if you prefer

web/ui/react-app/src/pages/targets/SeriesLabels.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const toggle = (): void => setTooltipOpen(!tooltipOpen);
+
+  return (
+    <>
+      <div id="series-labels" className="series-labels-container">
+        {Object.keys(labels).map(key => {
+          return <SeriesLabel key={key} labelKey={key} labelValue={labels[key]} />;
+        })}
+      </div>
+      <Tooltip isOpen={tooltipOpen} target="series-labels" toggle={toggle}>
```
juliusv: The tooltip only works on the first target on the page for me, and it's flickery when it loads, I think it might be rendering all tooltips on top of each other when hovering over the first target?

web/ui/react-app/src/pages/targets/Filter.module.css
hooten: Sure, I can do that. Or I can just make this a single style sheet. What would you prefer? I can do whatever

web/ui/react-app/src/pages/targets/Filter.module.css
hooten: sure, sounds good

web/ui/react-app/src/pages/targets/SeriesLabels.tsx
hooten: i will look into this, thanks for the heads up!

web/ui/react-app/src/pages/targets/StateIndicator.tsx
```diff
@@ -0,0 +1,30 @@
...
+  }
+};
+
+const StateIndicator: React.FC<StateIndicatorProps> = ({ health, message }) => {
+  const alertProps = {
+    className: styles.message,
+    color: getColor(health),
+  };
+
+  return <Alert {...alertProps}>{message ? message : health.toUpperCase()}</Alert>;
```
juliusv: Let's just do default-styled `<Badge>` elements in the new UI with `success` / `warning` / `danger` colors, instead of the old `<Alert>` elements with custom styles.

docs/querying/api.md
```diff
@@ -430,6 +432,37 @@ $ curl http://localhost:9090/api/v1/targets
...
+          "__address__": "127.0.0.1:9090",
+          "__metrics_path__": "/metrics",
+          "__scheme__": "http",
+          "job": "prometheus"
+        },
+        "labels": {
+          "instance": "127.0.0.1:9090",
+          "job": "prometheus"
+        },
+        "scrapeJob": "prometheus",
```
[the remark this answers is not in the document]
hooten: okay, i will change it to `scrapePool`

web/api/v1/api.go
```diff
@@ -577,51 +580,70 @@ type DroppedTarget struct {
 
 // TargetDiscovery has all the active targets.
 type TargetDiscovery struct {
-	ActiveTargets  []*Target        `json:"activeTargets"`
-	DroppedTargets []*DroppedTarget `json:"droppedTargets"`
+	ActiveTargets  []*Target        `json:"activeTargets,omitempty"`
```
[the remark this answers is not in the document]
hooten: okay, i will remove the omitempty

web/ui/react-app/src/pages/targets/JobPanel.module.css
```diff
@@ -0,0 +1,3 @@
+.container {
+  margin-top: -12px;
+}
```
juliusv: nit: Missing newline at end of file (I think our linter catches these already for TSX files, but maybe not for CSS)

web/api/v1/api.go
```diff
@@ -577,51 +580,70 @@ type DroppedTarget struct {
...
 			res = append(res, targets[k]...)
 		}
 		return res
 	}
 
-	tActive := flatten(api.targetRetriever.TargetsActive())
-	tDropped := flatten(api.targetRetriever.TargetsDropped())
-	res := &TargetDiscovery{ActiveTargets: make([]*Target, 0, len(tActive)), DroppedTargets: make([]*DroppedTarget, 0, len(tDropped))}
+	state := strings.ToLower(r.URL.Query().Get("state"))
+	showActive := state == "" || strings.Contains(state, "active")
```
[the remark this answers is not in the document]
hooten: great, i will do that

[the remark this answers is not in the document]
hooten: and update the docs

web/ui/react-app/src/pages/targets/SeriesLabels.test.tsx
juliusv: Huh weird, ok. Meh, since it's a test, I think just keep it.

web/ui/react-app/src/pages/targets/Filter.module.css
juliusv: @hooten To be able to judge that better I would first have to learn about CSS modules (eager to do that actually, just not right now for me), how they work, pros/cons, etc.

I'm fine with them being introduced now, but it does mean another way of doing things, and we should understand that we *should* eventually converge on fewer ways.

juliusv:
@hooten This is amazing, thanks! I only got halfway through with the review today, but hope to review the rest soon (maybe on the train to PromCon tomorrow, if I get a place with table and internet) :)

hooten:
Great! Thanks so much for the review! I'll work on the things you suggested

web/ui/react-app/src/pages/targets/__testdata__/testdata.ts
```diff
@@ -0,0 +1,220 @@
+/* eslint @typescript-eslint/camelcase: 0 */
```
boyskila: This file exports mutable objects which can be modified at some point from somewhere. Maybe you should export freezed Objects instead.

web/ui/react-app/src/pages/targets/StateIndicator.tsx
```diff
@@ -0,0 +1,30 @@
...
+    case 'up':
+      return 'success';
+    case 'down':
+      return 'danger';
+    default:
+      return 'warning';
+  }
+};
+
+const StateIndicator: React.FC<StateIndicatorProps> = ({ health, message }) => {
```
boyskila: You can use named imports for react interfaces ```React.FC -> FC ```

web/ui/react-app/src/pages/targets/StateIndicator.module.css
```diff
@@ -0,0 +1,4 @@
+.message {
+  padding: 0 4px 0 4px !important;
```
boyskila: I'm curious why !important ?

web/ui/react-app/src/pages/targets/SeriesLabels.tsx
```diff
@@ -0,0 +1,41 @@
...
+interface Labels {
+  [key: string]: string;
+}
+
+export interface SeriesLabelsProps {
+  discoveredLabels: Labels;
+  labels: Labels;
+}
+
+const formatLabels = (labels: Labels): string[] => Object.keys(labels).map(key => `${key}="${labels[key]}"`);
```
boyskila: You can omit the return type ```string[]``` since typescript infer it from the returning value

web/ui/react-app/src/pages/targets/SeriesLabels.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const SeriesLabels: React.FC<SeriesLabelsProps> = ({ discoveredLabels, labels }: SeriesLabelsProps) => {
+  const [tooltipOpen, setTooltipOpen] = React.useState(false);
+
+  const toggle = (): void => setTooltipOpen(!tooltipOpen);
+
+  return (
+    <>
+      <div id="series-labels" className="series-labels-container">
```
boyskila: is id needed? I found such a css class but not an id

web/ui/react-app/src/pages/targets/SeriesLabel.module.css
```diff
@@ -0,0 +1,3 @@
+.series-label {
```
boyskila: boostrap has built-in CSS classes for margins. Please refer to https://getbootstrap.com/docs/4.0/utilities/spacing/. It might be useful for such small things

web/ui/react-app/src/pages/targets/JobsList.tsx
```diff
@@ -0,0 +1,47 @@
...
+          .map((scrapeJob: string) => {
+            const targetGroupProps = {
+              scrapeJob,
+              targetGroup: targetGroups[scrapeJob],
+            };
+            return <JobPanel key={scrapeJob} {...targetGroupProps} />;
+          })}
+      </>
+    );
+  } else {
```
boyskila: You can omit else block

web/ui/react-app/src/pages/targets/SeriesLabels.tsx
juliusv: A constant `id` definitely seems problematic for a repeated element. I bet this is why all the tooltips show up when hovering over the first target's labels.

web/ui/react-app/src/pages/targets/JobHeader.module.css
```diff
@@ -0,0 +1,22 @@
...
+.normal {
+  composes: title;
+}
+
+.danger {
+  composes: title;
+  color: rgb(242, 65, 65);
+}
+
+.expand-btn {
```
boyskila: I guess all ```!important``` are needed to override default button styles, right?

hooten: @boyskila I think so. It seems that if not, the default styles get applied... unless you know of another way? I'm all ears!

web/ui/react-app/src/pages/targets/SeriesLabels.tsx
hooten: yes, for the Tooltip i think

web/ui/react-app/src/pages/targets/SeriesLabels.tsx
hooten: I thought there was a eslint rule about specifying return types?

web/ui/react-app/src/pages/targets/StateIndicator.module.css
hooten: Otherwise they won't be applied because the default styles get applied instead. I couldn't figure out a better way of doing this. 😭

web/ui/react-app/src/pages/targets/__testdata__/testdata.ts
hooten: Good point. It's just test data, but freezing is a good practice

hooten:
Thanks for the review @boyskila ! I really appreciate it!

web/ui/react-app/src/pages/targets/SeriesLabels.tsx
juliusv: Yeah, but then the `id` has to vary between targets :)

web/ui/react-app/src/pages/targets/JobHeader.module.css
boyskila: yes, inline styles will override the default styles unless they are not marked as ```!important``` as well. ```!important``` is dangerous and I try to avoid it if it's possible

web/ui/react-app/src/pages/targets/SeriesLabels.tsx
boyskila: I knew that we didn't have to delete the idGen function :D

web/ui/react-app/src/pages/targets/StateIndicator.module.css
boyskila: yes, you can add it like a style attr ```style={{ padding:....}}```

hooten: ah yes that's true

hooten: i wish there were a way to specify that my stylesheet takes precedence without having to put the styles directly in the js code

web/ui/react-app/src/pages/targets/SeriesLabels.tsx
hooten: ahhh i see now

web/ui/react-app/src/pages/targets/StateIndicator.module.css
boyskila: You can try to make stronger selector than the default one. If the default one is 
```.foo.bar { default styles here }```
you selector should be ```.your-css-class .foo.bar { your styles here }```

File: web/ui/react-app/src/pages/targets/SeriesLabels.tsx (hunk 18 above)
Comment (hooten) [reply web/ui/react-app/src/pages/targets/SeriesLabels.tsx]: Screen Shot 2019-11-05 at 2 55 45 PM

Comment (hooten) [reply web/ui/react-app/src/pages/targets/SeriesLabels.tsx]: Screen Shot 2019-11-05 at 2 55 51 PM

Comment (hooten) [reply web/ui/react-app/src/pages/targets/SeriesLabels.tsx]: Yeah, turns out there is. I see your point though. On the other hand, it's nice to specify types

Comment (juliusv) [reply web/ui/react-app/src/pages/targets/SeriesLabels.tsx]: FWIW we have a gazillion missing ones currently and are just ignoring those warnings. I think that's fine for now, but also not opposed to adding return types to get rid of the warnings.

File: web/ui/react-app/src/pages/targets/JobHeader.module.css (hunk 22 above)
Comment (hooten) [reply web/ui/react-app/src/pages/targets/JobHeader.module.css]: okay i went ahead and removed all the `!important`s. Thanks for the insight @boyskila !

File: web/ui/react-app/src/pages/targets/SeriesLabel.module.css (hunk 20 above)
Comment (hooten) [reply web/ui/react-app/src/pages/targets/SeriesLabel.module.css]: done!

File: web/ui/react-app/src/pages/targets/StateIndicator.module.css (hunk 17 above)
Comment (hooten) [reply web/ui/react-app/src/pages/targets/StateIndicator.module.css]: used the inline styles attr here

File: web/ui/react-app/src/pages/targets/__testdata__/testdata.ts (hunk 15 above)
Comment (hooten) [reply web/ui/react-app/src/pages/targets/__testdata__/testdata.ts]: frozen

File: web/ui/react-app/src/pages/targets/SeriesLabels.tsx (hunk 9 above)
Comment (hooten) [reply web/ui/react-app/src/pages/targets/SeriesLabels.tsx]: this is fixed now :)

File: web/ui/react-app/src/pages/targets/SeriesLabels.tsx (hunk 19 above)
Comment (hooten) [reply web/ui/react-app/src/pages/targets/SeriesLabels.tsx]: fixed

File: web/ui/react-app/src/pages/targets/SeriesLabel.tsx [hunk 23]
```diff
@@ -0,0 +1,17 @@
+import React, { FC } from 'react';
```
boyskila: I would use the Badge component without abstracted it away.

web/ui/react-app/src/pages/targets/SeriesLabels.tsx
```diff
@@ -0,0 +1,47 @@
...
+export interface SeriesLabelsProps {
+  discoveredLabels: Labels;
+  labels: Labels;
+  idx: number;
+  scrapePool: string;
+}
+
+const formatLabels = (labels: Labels): string[] => Object.keys(labels).map(key => `${key}="${labels[key]}"`);
+
+const SeriesLabels: FC<SeriesLabelsProps> = ({ discoveredLabels, labels, idx, scrapePool }: SeriesLabelsProps) => {
```
juliusv: Thinking about this, `<SeriesLabel>` and `<SeriesLabels>` is a misnomer, since it's not about series yet, but about targets (what we call target labels). This component is even coded specifically around targets in that it has a tooltip with labels from before relabeling, which only applies to targets, not series. Can we rename it `<TargetLabel>` / `<TargetLabels>`?

web/ui/react-app/src/pages/targets/SeriesLabel.tsx
juliusv: Yeah, arguably this is a bit too much abstraction for a little thing that can just appear verbatim anywhere it's used.

web/ui/react-app/src/ErrorAlert.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { FC } from 'react';
+import { Alert } from 'reactstrap';
+
+interface ErrorProps {
+  summary: string;
+  message?: string;
+}
+
+const Error: FC<ErrorProps> = ({ summary, message }) => {
```
juliusv: `<Error>` is another case where I wonder if it's worth it to abstract it, but if we do, it would be good to change all the existing error messages to use this too.

Also, some of the summaries start with `Error fetching ...`, which stutters after we already say `Error:`. How about doing `<strong>{summary}</strong}: {message}`, and default `summary` to `Error:` when it's missing?

web/ui/react-app/src/pages/targets/JobsList.tsx
```diff
@@ -0,0 +1,46 @@
...
+interface JobsListProps {
+  filter: FilterData;
+}
+
+const filterByHealth = ({ metadata, targets }: TargetGroup, { showHealthy, showUnhealthy }: FilterData): boolean => {
+  const isHealthy = metadata.up === targets.length;
+  return (isHealthy && showHealthy) || (!isHealthy && showUnhealthy);
+};
+
+const JobsList: FC<JobsListProps & PathPrefixProps> = ({ filter, pathPrefix }) => {
```
juliusv: Although there's often a 1:1 relation between jobs and scrape configs / scrape pools, it can also be 1:n or m:1 or even n:m in some cases. So technically it would be more correct to call it a `ScrapePoolList` or something like that.

Btw., in the future we might want to have arbitrary labels by which we can group things on the /targets page, but let's think of that when we get there.

web/ui/react-app/src/pages/targets/Targets.tsx
```diff
@@ -0,0 +1,21 @@
...
+import PathPrefixProps from '../../PathPrefixProps';
+
+const Targets: FC<RouteComponentProps & PathPrefixProps> = ({ pathPrefix }) => {
+  const [filter, setFilter] = useState({ showHealthy: true, showUnhealthy: true });
+  const filterProps = { filter, setFilter };
+  const jobsListProps = { filter, pathPrefix };
+
+  return (
+    <>
+      <h1>Targets</h1>
```
juliusv: This should be `<h2>`, in the old UI that page is wrong / inconsistent with the other pages.

web/ui/react-app/src/Loader.tsx
juliusv: Actually when this component was used, I thought it was loading something, but it's just a loading indicator. Could you rename it to `<Spinner>`, `<LoadingSpinner>`, `<LoadingIndicator>` or something like that to make usage code clearer?

juliusv: (though I still think we need to come to a conclusion about the abstractions introduced in that other PR, and whether they should be used here - I think at least after this wave of PRs we should focus on settling some things and making them consistent)

web/ui/react-app/src/pages/targets/target.ts
```diff
@@ -0,0 +1,42 @@
...
+  labels: Labels;
+  scrapePool: string;
+  scrapeUrl: string;
+  lastError: string;
+  lastScrape: string;
+  lastScrapeDuration: number;
+  health: string;
+}
+
+export interface TargetGroupMetadata {
```
juliusv: This seems like premature abstraction, I'd just add an `upCount` field directly to the `TargetGroup`.

web/api/v1/api.go
```diff
@@ -582,46 +585,69 @@ type TargetDiscovery struct {
...
-			DiscoveredLabels: t.DiscoveredLabels().Map(),
-		})
+				res.ActiveTargets = append(res.ActiveTargets, &Target{
+					DiscoveredLabels:   target.DiscoveredLabels().Map(),
+					Labels:             target.Labels().Map(),
+					ScrapePool:         key,
+					ScrapeURL:          target.URL().String(),
+					LastError:          lastErrStr,
+					LastScrape:         target.LastScrape(),
+					LastScrapeDuration: float64(target.LastScrapeDuration()) / float64(time.Second),
```
juliusv: Just do:

```suggestion
					LastScrapeDuration: target.LastScrapeDuration().Seconds(),
```

web/api/v1/api.go
```diff
@@ -582,46 +585,69 @@ type TargetDiscovery struct {
...
+					Health:             target.Health(),
+				})
+			}
+		}
+	} else {
+		res.ActiveTargets = make([]*Target, 0)
```
juliusv: style nit: use `make()` only when necessary:

```suggestion
		res.ActiveTargets = []*Target{}
```

web/api/v1/api.go
```diff
@@ -582,46 +585,69 @@ type TargetDiscovery struct {
...
+	if showDropped {
+		tDropped := flatten(api.targetRetriever.TargetsDropped())
+		res.DroppedTargets = make([]*DroppedTarget, 0, len(tDropped))
+		for _, t := range tDropped {
+			res.DroppedTargets = append(res.DroppedTargets, &DroppedTarget{
+				DiscoveredLabels: t.DiscoveredLabels().Map(),
+			})
+		}
+	} else {
+		res.DroppedTargets = make([]*DroppedTarget, 0)
```
juliusv:
```suggestion
		res.DroppedTargets = []*DroppedTarget{}
```

web/ui/react-app/src/pages/targets/SeriesLabel.tsx
hooten: Cool, will do

web/ui/react-app/src/pages/targets/SeriesLabels.tsx
hooten: Yes will do!

web/ui/react-app/src/Loader.tsx
hooten: I will remove the abstraction for the time being

web/ui/react-app/src/ErrorAlert.tsx
hooten: I've removed the abstraction. I think you all are right that it's not necessary :)

web/ui/react-app/src/pages/targets/Targets.tsx
hooten: done

web/ui/react-app/src/pages/targets/Filter.tsx
```diff
@@ -7,6 +7,10 @@ export interface FilterData {
   showUnhealthy: boolean;
 }
 
+interface Blah<T> {
```
boyskila: I'm not fully aware of this PR functionality but what is Blah?

web/ui/react-app/src/pages/targets/Filter.tsx
```diff
@@ -7,6 +7,10 @@ export interface FilterData {
...
+  fn: (t: T) => T;
+}
+
 export interface FilterProps {
   filter: FilterData;
   setFilter: React.Dispatch<React.SetStateAction<FilterData>>;
```
boyskila: You can omit React namespace here and use named imports instead

web/ui/react-app/src/pages/targets/JobsList.tsx
```diff
@@ -20,9 +21,17 @@ const JobsList: FC<JobsListProps & PathPrefixProps> = ({ filter, pathPrefix }) =
   const { response, error } = useFetch(`${pathPrefix}/api/v1/targets?state=active`);
 
   if (error) {
-    return <Error summary="Error fetching targets" message={error.message} />;
+    return (
+      <Alert color="danger">
+        <strong>Error fetching targets:</strong> {error.message}
+      </Alert>
+    );
   } else if (response && response.status !== 'success') {
```
boyskila: I think this shouldn't be handled by ```Joblist```

web/ui/react-app/src/pages/targets/TargetLabels.test.tsx
```diff
@@ -21,32 +20,32 @@ describe('SeriesLabels', () => {
...
+    expect(targetLabels.find(Badge)).toHaveLength(3);
   });
 
   it('renders a tooltip for discovered labels', () => {
-    const tooltip = seriesLabels.find(Tooltip);
+    const tooltip = targetLabels.find(Tooltip);
     expect(tooltip).toHaveLength(1);
     expect(tooltip.prop('isOpen')).toBe(false);
     expect(tooltip.prop('target')).toEqual('series-labels-cortex/node-exporter_group/0-1');
     expect(tooltip.prop('isOpen')).toBe(false);
```
boyskila: Is it needed isOpen to be checked twice? Btw you can use ```toBeFalsy()```

web/ui/react-app/src/pages/targets/TargetLabels.tsx
```diff
@@ -16,7 +15,7 @@ export interface SeriesLabelsProps {
 
 const formatLabels = (labels: Labels): string[] => Object.keys(labels).map(key => `${key}="${labels[key]}"`);
 
-const SeriesLabels: FC<SeriesLabelsProps> = ({ discoveredLabels, labels, idx, scrapePool }: SeriesLabelsProps) => {
+const TargetLabels: FC<TargetLabelsProps> = ({ discoveredLabels, labels, idx, scrapePool }: TargetLabelsProps) => {
```
boyskila: I guess Typescript already know about props type since the component is declared as ```FC<TargetLabelProps>```

web/ui/react-app/src/pages/targets/TargetLabels.tsx
```diff
@@ -25,23 +24,25 @@ const SeriesLabels: FC<SeriesLabelsProps> = ({ discoveredLabels, labels, idx, sc
...
+              {`${labelName}="${labels[labelName]}"`}
+            </Badge>
+          );
         })}
       </div>
-      <Tooltip isOpen={tooltipOpen} target={id} toggle={toggle} style={{maxWidth: 'none', textAlign: 'left'}}>
+      <Tooltip isOpen={tooltipOpen} target={id} toggle={toggle} style={{ maxWidth: 'none', textAlign: 'left' }}>
         <b>Before relabeling:</b>
         {formatLabels(discoveredLabels).map((s: string, idx: number) => (
           <React.Fragment key={idx}>
```
boyskila: ```React``` namespace can be omitted

web/ui/react-app/src/pages/targets/Targets.test.tsx
```diff
@@ -10,9 +10,9 @@ describe('Targets', () => {
   };
   const targets = shallow(<Targets {...defaultProps} />);
   it('renders a header', () => {
-    const h1 = targets.find('h1');
-    expect(h1).toHaveLength(1);
-    expect(h1.text()).toEqual('Targets');
+    const h2 = targets.find('h2');
+    expect(h2).toHaveLength(1);
```
boyskila: I would split this test on two - one test to check the count and one the content

web/ui/react-app/src/pages/targets/JobsList.test.tsx
```diff
@@ -19,9 +20,12 @@ describe('Flags', () => {
...
 
   describe('before data is returned', () => {
-    it('renders a Loader', () => {
+    it('renders a spinner', () => {
       const jobsList = shallow(<JobsList {...defaultProps} />);
-      expect(jobsList.find(Loader)).toHaveLength(1);
+      const spinner = jobsList.find(FontAwesomeIcon);
+      expect(spinner).toHaveLength(1);
+      expect(spinner.prop('icon')).toEqual(faSpinner);
+      expect(spinner.prop('spin')).toBe(true);
```
boyskila: You can use ```toBeTruthy()``` ... Spinner count, I would check in a separate test like -```it('should not render multiple spinners')``` or something like that

web/ui/react-app/src/Loader.tsx
hooten: done

web/ui/react-app/src/pages/targets/JobPanel.tsx
```diff
@@ -0,0 +1,33 @@
...
+import styles from './JobPanel.module.css';
+
+interface JobPanelProps {
+  scrapeJob: string;
+  targetGroup: TargetGroup;
+}
+
+const JobPanel: FC<JobPanelProps> = ({ scrapeJob, targetGroup }) => {
+  const [expanded, setExpanded] = useState(true);
+  const jobHeaderProps = {
```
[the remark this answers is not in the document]
hooten: I grouped a bunch of code together under ScrapePoolPanel so that there is less intermediary code. Thanks for the suggestion! I like it better too

web/ui/react-app/src/pages/targets/SeriesLabels.tsx
hooten: done

web/ui/react-app/src/pages/targets/JobsList.tsx
hooten: This is now `ScrapePoolList` which contains a number of `ScrapePoolPanel`s.

I think that the `groupTargets` function could be extended (in a future pr) to enable different groupings.

web/ui/react-app/src/pages/targets/target.ts
hooten: done

web/ui/react-app/src/pages/targets/TargetLabels.test.tsx
hooten: Good catch, thanks. I prefer using `toBe(false)` because it is a stronger and more precise assertion than `toBeFalsy()` which would also accept 0, null, etc.

web/ui/react-app/src/pages/targets/JobsList.tsx
hooten: Do you mean the line in particular, or all the error and loading handling? I can move it somewhere else, like Targets.tsx if you think that would be better :)

hooten: ...Or somewhere else. Let me know! Thanks!

web/ui/react-app/src/pages/targets/Filter.tsx
hooten: oops this was a mistake. I've removed it now.

web/ui/react-app/src/pages/targets/Filter.tsx
hooten: done! Thanks

web/ui/react-app/src/pages/targets/JobsList.test.tsx
hooten: Again, i prefer `toBe(true)`. But i will split it into multiple tests as you've suggested. Thanks!

hooten: done

web/ui/react-app/src/pages/targets/Targets.test.tsx
hooten: done

web/ui/react-app/src/pages/targets/JobsList.tsx
boyskila: No no, I mean that components should not be aware of the response status. IMO they should be aware only of their data and nothing else. Actually this is what I've tried to solve here https://github.com/prometheus/prometheus/pull/6267 - creating strongly specific components responsible for doing one thing

web/ui/react-app/src/pages/targets/JobsList.test.tsx
boyskila: Yes, didn't take into account that toBeFalsy/Truthy will handle values different than true/false.

web/ui/react-app/src/pages/targets/Filter.tsx
```diff
@@ -0,0 +1,39 @@
...
+  showHealthy: boolean;
+  showUnhealthy: boolean;
+}
+
+export interface FilterProps {
+  filter: FilterData;
+  setFilter: React.Dispatch<React.SetStateAction<FilterData>>;
+}
+
+const Filter: React.FC<FilterProps> = ({ filter, setFilter }) => {
```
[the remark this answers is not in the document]
hooten: This is done! See `useLocalStorage` function

web/ui/react-app/src/pages/targets/JobsList.tsx
hooten: Okay, I understand what you mean. I can change it later to integrate with whatever Fetch logic exists in the future. For now, only `useFetch` exists.

web/ui/react-app/src/pages/targets/TargetLabels.tsx
hooten: done, thanks!

web/ui/react-app/src/pages/targets/TargetLabels.tsx
hooten: done, thanks for the catch!

web/ui/react-app/src/pages/targets/JobsList.tsx
boyskila: Yes, better to leave it for now :)

hooten: cool thanks!

juliusv:
Just FYI, PromCon has swallowed me up whole, but I plan on being back to reviews in a useful way on Monday at the latest :)

juliusv:
> If it's ready to merge in your view, I'll merge!

Uh, after the last nit fixes, I meant of course :)

docs/querying/api.md
```diff
@@ -431,6 +433,39 @@ $ curl http://localhost:9090/api/v1/targets
 }
 ```
 
+The `state` query parameter allows the caller to filter by active or dropped targets,
+(e.g., `state=active`, `state=dropped`, `state=active,dropped`). 
+Other values are ignored.
+
+```json
+$ curl http://localhost:9090/api/v1/targets?state=active
```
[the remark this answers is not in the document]
hooten: done!

docs/querying/api.md
```diff
@@ -431,6 +433,39 @@ $ curl http://localhost:9090/api/v1/targets
...
+{
+  "status": "success",
+  "data": {
+    "activeTargets": [
```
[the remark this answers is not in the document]
hooten: Good call. I added a line in the docs as well to note this.

hooten:
@juliusv thanks again for the feedback! I think this is ready to merge

juliusv:
:ship:

juliusv:
Thanks :)

hooten:
Awesome! Thanks for giving me the chance to contribute!

PR #6277, opened 2019-11-05
React UI: Mark not-yet-done pages as under construction

juliusv:
This is a quick change because the release is about to be cut in a few
hours and it would be good to clarify to users what the empty pages in
the new UI are about.


juliusv:
@csmarchbanks

PR #6291, opened 2019-11-07
Added examples for clamp_max, max_over_time, sum_over_time functions.

HemantNegi:
Added examples for below functions:
- clamp_max
- max_over_time
- sum_over_time

### DIFF
--- a/docs/querying/examples.md
+++ b/docs/querying/examples.md
@@ -23,7 +23,7 @@ making it a range vector:
     http_requests_total{job="apiserver", handler="/api/comments"}[5m]
 
 Note that an expression resulting in a range vector cannot be graphed directly,
-but viewed in the tabular ("Console") view of the expression browser.
+but viewed in the tabular ("Console") view of the expression browser, also this output will not be sorted by timestamp.
 
 Using regular expressions, you could select time series only for jobs whose
 name match a certain pattern, in this case, all jobs that end with `server`:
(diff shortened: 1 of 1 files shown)


docs/querying/examples.md
```diff
@@ -95,3 +95,25 @@ Assuming this metric contains one time series per running instance, you could
 count the number of running instances per application like this:
 
     count by (app) (instance_cpu_time_ns)
+  
+To find out the maximum value of a metric over a period of time, for eg. last 
+6 hours, you can query like this:
+
+    max_over_time(instance_cpu_time_ns[6h])
```
brian-brazil: This is not good query, the metric isn't following bestpractices for naming and it looks like a counter.

docs/querying/examples.md
```diff
@@ -95,3 +95,25 @@ Assuming this metric contains one time series per running instance, you could
...
+clamp them to the threshold value like this:
+
+    clamp_max(instance_cpu_time_ns, 100)
+    
+You can calculate the rolling sum of value of a metric over a period of time, 
+say 6 hours like below:
+
+    sum_over_time(some_mertic_name[6h])
+    
+Lets assume we have a deduped metric `processing_time_ns` denoting the time
```
brian-brazil: This isn't what this query does (plus it's a syntax error), and this sort of thing requires much more discussion than makes sense on a page like this.

docs/querying/examples.md
HemantNegi: I don't understand. its the same metric used in other examples also. 

> it looks like a counter
Why do you think so?

docs/querying/examples.md
HemantNegi: I fixed the syntax error. thanks for pointing that out.
Yeah, I know this is a complex example. But this query counts the number of 1's within the given time period.

docs/querying/examples.md
brian-brazil: Hmm, we should fix those examples then. This sort of data is usually exposed as a counter - which the other examples show.

docs/querying/examples.md
brian-brazil: There's more to this than what's here, it's far too complex to include here. It's a blog post in itself.

brian-brazil:
We looked at this in the bug scrub, and the comments above have not been addressed. We're going to close this for now, but if you would like to add additional better examples to this page feel free to send in a PR.

PR #6292, opened 2019-11-07
Add exemplar support to the openmetrics parser

shreyassrivatsan:
The change adds support for exemplars in the openmetrics exposition format. It includes changes to the grammar and parsing of the expression along with tests. 

It does change the behavior of the grammar. Now we expect an exemplar after the `#` after a metric. The earlier grammar allow for random comments that were consumed by the parser.


pkg/textparse/openmetricsparse.go
```diff
@@ -300,50 +350,103 @@ func (p *OpenMetricsParser) Next() (Entry, error) {
...
 		}
 
 		// The openMetricsLexer ensures the value string is quoted. Strip first
 		// and last character.
-		p.offsets = append(p.offsets, p.l.start+1, p.l.i-1)
+		offsets = append(offsets, p.l.start+1, p.l.i-1)
+	}
+}
 
+func (p *OpenMetricsParser) getFLoatValue(t token, after string) (float64, error) {
```
robskillington: `getFLoatValue` -> `getFloatValue`

pkg/textparse/openmetricsparse.go
```diff
@@ -85,10 +87,19 @@ type OpenMetricsParser struct {
...
+	exemplar      []byte
+	exemplarVal   float64
+	exemplarTs    int64
+	hasExemplarTs bool
 }
 
 // NewOpenMetricsParser returns a new parser of the byte slice.
 func NewOpenMetricsParser(b []byte) Parser {
+	if len(b) > 0 && b[len(b)-1] != '\n' {
+		b = append(b, '\n')
```
robskillington: Ideally we wouldn't need to append an extra endline - can we add this support without this requirement?

If it adds an extra character the entire buffer might need re-allocation, also not guaranteed that this is safe to do either if it's a subslice of a bigger slice passed to this function.

pkg/exemplar/exemplar.go
```diff
@@ -0,0 +1,23 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package exemplar
+
+import "github.com/prometheus/prometheus/pkg/labels"
+
+// Exemplar is additional information associated with a metric.
```
brian-brazil: Time series

pkg/textparse/interface.go
```diff
@@ -50,6 +51,10 @@ type Parser interface {
 	// It returns the string from which the metric was parsed.
 	Metric(l *labels.Labels) string
 
+	// Exemplar writes the exemplar of the current sample into the passed
+	// exemplar. It returns the string from which the metric was parsed.
```
brian-brazil: exemplar

pkg/textparse/openmetricsparse.go
```diff
@@ -170,6 +178,34 @@ func (p *OpenMetricsParser) Metric(l *labels.Labels) string {
...
+// Exemplar writes the exemplar of the current sample into the passed
+// exemplar. It returns the string from which the metric was parsed.
+func (p *OpenMetricsParser) Exemplar(e *exemplar.Exemplar) string {
+	// Allocate the full immutable string immediately, so we just
+	// have to create references on it below.
+	s := string(p.exemplar)
+
+	e.Value = p.exemplarVal
+	if p.hasExemplarTs {
+		e.Ts = &p.exemplarTs
```
brian-brazil: Won't this be lost when we parse the next one?

pkg/textparse/openmetricsparse.go
```diff
@@ -170,6 +178,34 @@ func (p *OpenMetricsParser) Metric(l *labels.Labels) string {
...
+	for i := 0; i < len(p.eOffsets); i += 4 {
+		a := p.eOffsets[i] - p.start
+		b := p.eOffsets[i+1] - p.start
+		c := p.eOffsets[i+2] - p.start
+		d := p.eOffsets[i+3] - p.start
+
+		e.Labels = append(e.Labels, labels.Label{Name: s[a:b], Value: s[c:d]})
+	}
+
+	// Sort labels. We can skip the first entry since the metric name is
```
brian-brazil: This comment is outdated.

pkg/textparse/openmetricsparse.go
```diff
@@ -286,11 +327,17 @@ func (p *OpenMetricsParser) Next() (Entry, error) {
...
+			case tComment:
+				if err := p.parseComment(); err != nil {
+					return EntryInvalid, err
+				}
+			default:
+				return EntryInvalid, parseError("expected next entry after timestamp", t3)
 			}
 		default:
-			return EntryInvalid, parseError("expected timestamp or new record", t)
+			return EntryInvalid, parseError("expected timestamp or comment", t2)
```
brian-brazil: It's not a comment, it's a hash symbol

pkg/textparse/openmetricsparse_test.go
```diff
@@ -386,6 +411,14 @@ func TestOpenMetricsParseErrors(t *testing.T) {
 			input: "foo 0 1_2\n",
 			err:   "unsupported character in float",
 		},
+		{
```
brian-brazil: We should check that the metric name suffix is _bucket or _total

pkg/textparse/openmetricsparse.go
shreyassrivatsan: I do not follow. Can we have multiple exemplars per metric? If not this should be ok I think as the same applies to all the information stored in the parser, no?

brian-brazil: You're taking a pointer which is static relative to the parser.

shreyassrivatsan: Ok, yes. That's right. Its a problem above for `p.Ts` too. Fixed both.

pkg/textparse/openmetricsparse.go
```diff
@@ -25,10 +27,13 @@ import (
 
 	"github.com/pkg/errors"
 
+	"github.com/prometheus/prometheus/pkg/exemplar"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/pkg/value"
 )
 
+var allowedSuffixes = []string{"_total", "_bucket"}
```
robskillington: Perhaps make these `[][]byte{[]byte("_total"), []byte("_bucket")}` so that don't need to do a `[]byte(strValue)` in the validate code later?

pkg/textparse/openmetricsparse.go
```diff
@@ -170,6 +182,34 @@ func (p *OpenMetricsParser) Metric(l *labels.Labels) string {
 	return s
 }
 
+// Exemplar writes the exemplar of the current sample into the passed
+// exemplar. It returns the string from which the metric was parsed.
+func (p *OpenMetricsParser) Exemplar(e *exemplar.Exemplar) string {
```
brian-brazil: How do you distinguish between a series with no exemplar, and one with no labels and a value of 0?

brian-brazil:
Hmm, can you rebase to try and get rid of that test failure?

shreyassrivatsan:
> Hmm, can you rebase to try and get rid of that test failure?

Rebased and got rid of the test failures.

brian-brazil:
Thanks!

PR #6294, opened 2019-11-08
Link example when the main config file is first referenced in the README

rothsa:
I didn't have a configuration file handy, and it was not immediately obvious to me where to find one. I only knew that it was specifying the main config as I've used prometheus before, so it would be a good idea to point new folks in the right direction. :)

Contributing asks me to link a maintainer for minor changes, so, cc @juliusv


juliusv:
Thanks! \o/ I think it's ok to add it in that section for now.

Even better for the future would be if we could disentangle the information in those sections a bit more and separate installing/building/running more cleanly (except in the Docker case, that's basically one step). Then there'd be a "Running" section that only covers that, and that would be the best place to mention the config file.

PR #6062, opened 2019-09-26
cmd/promtool: add tests to alert unit testing

simonpasquier:
The change also refactors a bit the code to move some validation and setup logic from the test execution to YAML deserialization.

PR #6297, opened 2019-11-09
Support extended durations in promtool unit tests (Fixes #6285)

### LINKED ISSUE #6285: eval_time does not actually support all possible <duration> values
## Bug Report
**What did you do?**
I tried to set `eval_time` as for an alert rule test as follows, but none of them worked:
```
eval_time: 1d
eval_time: 1w
eval_time: 1y
```

I expected no errors, [as per documentation](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#configuration-file):
> `<duration>`: a duration matching the regular expression `[0-9]+(ms|[smhdwy])`

```
yaml: unmarshal errors:
  line 27: cannot unmarshal !!str `1d` into time.Duration

# same for "1w" and "1d"
```

**Environment**

* System information:
`Darwin 19.0.0 x86_64`

* Prometheus version:
```
prometheus, version 2.13.1 (branch: non-git, revision: non-git)
  build user:       brew@Catalina.local
  build date:       20191018-01:35:32
  go version:       go1.13.1
```

```
promtool, version 2.13.1 (branch: non-git, revision: non-git)
  build user:       brew@Catalina.local
  build date:       20191018-01:35:32
  go version:       go1.13.1
```

* Alertmanager version:
N/A

N/A

N/A

* Logs:
N/A

neufeldtech:
Fixes #6285 

This PR updates the promtool unit testing framework to parse the yaml files with `model.Duration` instead of `time.Duration`. This allows the code to support durations such as `1d` or `1w` and brings this implementation closer to what a user would expect as it matches the supported durations of other prometheus config files.

I'm open to feedback on if there is a more go idiomatic way of performing this fix, rather then doing `time.Duration(model.Duration)` in several places in the code as I've done here.

### DIFF
--- a/cmd/promtool/unittest.go
+++ b/cmd/promtool/unittest.go
@@ -371,15 +371,15 @@ Outer:
 
 // seriesLoadingString returns the input series in PromQL notation.
 func (tg *testGroup) seriesLoadingString() string {
-	result := ""
-	result += "load " + shortDuration(time.Duration(tg.Interval)) + "\n"
+
+	result := fmt.Sprintf("load %v\n", shortDuration(tg.Interval))
 	for _, is := range tg.InputSeries {
-		result += "  " + is.Series + " " + is.Values + "\n"
+		result += fmt.Sprintf("  %v %v\n", is.Series, is.Values)
 	}
 	return result
 }
 
-func shortDuration(d time.Duration) string {
+func shortDuration(d model.Duration) string {
 	s := d.String()
 	if strings.HasSuffix(s, "m0s") {
 		s = s[:len(s)-2]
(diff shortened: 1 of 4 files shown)


cmd/promtool/unittest.go
```diff
@@ -405,13 +407,13 @@ func orderedGroups(groupsMap map[string]*rules.Group, groupOrderMap map[string]i
 func (tg *testGroup) maxEvalTime() time.Duration {
 	var maxd time.Duration
 	for _, alert := range tg.AlertRuleTests {
-		if alert.EvalTime > maxd {
-			maxd = alert.EvalTime
+		if time.Duration(alert.EvalTime) > maxd {
```
brian-brazil: Change the type of maxd, so you don't need to convert

cmd/promtool/unittest.go
```diff
@@ -370,7 +372,7 @@ Outer:
 // seriesLoadingString returns the input series in PromQL notation.
 func (tg *testGroup) seriesLoadingString() string {
 	result := ""
-	result += "load " + shortDuration(tg.Interval) + "\n"
+	result += "load " + shortDuration(time.Duration(tg.Interval)) + "\n"
```
brian-brazil: Hmm, can we simplify here?

cmd/promtool/unittest_test.go
```diff
@@ -0,0 +1,42 @@
+// Copyright 2018 The Prometheus Authors
```
codesome: There is already a PR for unit tests for unit test https://github.com/prometheus/prometheus/pull/6062, I would like to avoid duplication of effort. Can you add these test cases to that PR instead?

codesome: Or more like it would be better if we merge that PR first and merge this with the additional test case.

neufeldtech: I'm fine with this PR going after #6062 is merged. I had created the one basic test to prove that deserialization of durations such as `1d` was now working

brian-brazil: The other PR seems to have stalled, so I'll just go ahead and merge this.

brian-brazil:
Thanks!

PR #6303, opened 2019-11-11
Fix target metadata API for empty match_targets

slrtbtfs:
According to the documentation, the target metadata API accepts it,
if no value for match_target has been provided. This was not the case
in the implementation.

This commit make the API behave as described in the docs.


juliusv:
Seems good to me, thanks! It would be nice to also have tests for this endpoint, but since there are none at all for it yet, l I won't force it in this PR.

PR #6308, opened 2019-11-12
fix: .promu.yml

kormotodor:
build: flags: -tags should be separated by space instead by comma

Obvious fix.

### DIFF
--- a/.promu.yml
+++ b/.promu.yml
@@ -12,7 +12,7 @@ build:
           path: ./cmd/promtool
         - name: tsdb
           path: ./tsdb/cmd/tsdb
-    flags: -mod=vendor -a -tags netgo,builtinassets
+    flags: -mod=vendor -a -tags 'netgo builtinassets'
     ldflags: |
         -X github.com/prometheus/common/version.Version={{.Version}}
         -X github.com/prometheus/common/version.Revision={{.Revision}}


kormotodor:
Little more about bad conditions in https://github.com/prometheus/prometheus/blob/master/.promu.yml:
1. if `flags: -mod=vendor -a -tags netgo,builtinassets`; then:
```
 >   prometheus
cmd/go: -tags space-separated list contains comma
...
github.com/prometheus/common/version.BuildDate=20191112-09:43:22  -extldflags '-static' -mod=vendor -a -tags netgo,builtinassets github.com/prometheus/prometheus/cmd/prometheus: exit status 2
```
2. if `flags: -mod=vendor -a -tags netgo builtinassets`; then:
```
 >   prometheus
can't load package: package builtinassets: cannot find package "."
...
github.com/prometheus/common/version.BuildDate=20191112-11:14:32  -extldflags '-static' -mod=vendor -a -tags netgo builtinassets github.com/prometheus/prometheus/cmd/prometheus: exit status 1
```

juliusv:
I think this would break due to https://github.com/prometheus/promu/issues/172, but with Go 1.13 the comma syntax should work as well. Are you trying with Go 1.12?

kormotodor:
Yes, those builds was under Go 1.12.

juliusv:
This promu file defines the Go version to be 1.13 at the top, so if you already change that to 1.12 for yourself locally, wouldn't it be sufficient to also make any other changes locally? As I mentioned in https://github.com/prometheus/promu/issues/172, promu currently doesn't handle quotes correctly, so I think this fix wouldn't actually help (it doesn't actually error, but it doesn't set the correct tags).

kormotodor:
Agreed

PR #6309, opened 2019-11-13
Exemplars store in-memory storage interface

shreyassrivatsan:
Initial pass at passing exemplars through the scrape logic. 

This is one option we could take by changing the `Appender` interface `Add`, `AddFast` methods. The current draft shows the changes involved at a high level. Still need to deal with optimally handling the no exemplar case. 

An alternate approach we could take is to not change the current Appender interface, but add a new interface `AppenderWithExemplars` and then upcast the appender provided.. `appender, ok := Appender.(AppenderWithExemplar)` and decide whether to handle exemplars based on what type of appender has been configure.. 

There are few cases to take care of here

1. Metric does not have exemplar, appender does not support exemplar
2. Metric has exemplar, appender does not support exemplars - we should ideally not even parse and propagate the exemplar.
3. Metric has exemplar, appender supports exemplars - push the exemplars through the scraper

### DIFF
--- a/pkg/exemplar/interface.go
+++ b/pkg/exemplar/interface.go
@@ -17,6 +17,15 @@ import (
 	"github.com/prometheus/prometheus/pkg/labels"
 )
 
+// TsExemplar is an exemplar alongside a timestamp.
+type TsExemplar struct {
+	Ts       int64
+	Exemplar Exemplar
+}
+
+// SeriesExemplar is a time series of exemplar values.
+type SeriesExemplar []TsExemplar
+
 // The following are similar to those of `storage/interfaces.go`:
 // https://github.com/prometheus/prometheus/blob/91d7175eaac18b00e370965f3a8186cc40bf9f55/storage/interface.go#L31-L65
 
@@ -26,9 +35,10 @@ type Exemplars interface {
 	// and timestamp provided (so this can be O(1) lookup).
 	Get(l labels.Labels, t int64) (Exemplar, bool, error)
 
-	// Query is a query to find an exemplars with exact matches on labels
-	// and timestamp ranges provided (this is a range query).
-	Query(l labels.Labels, start, end int64) ([]Exemplar, error)
+	// Query is a query to find an exemplars with partial matches on labels
+	// and timestamp ranges provided (this is a range query). Query returns
+	// a list of SeriesExemplar for each label match.
+	Query(l labels.Labels, start, end int64) ([]SeriesExemplar, error)
 
 	// Add an exemplar, it will be retained based on configuration at
 	// the construction of the exmplar storage.
--- a/pkg/exemplar/interface.go
+++ b/pkg/exemplar/interface.go
@@ -36,4 +36,4 @@ type Exemplars interface {
 
 	// Close the storage.
 	Close() error
-} 
+}
--- a/pkg/exemplar/interface.go
+++ b/pkg/exemplar/interface.go
@@ -17,28 +17,16 @@ import (
 	"github.com/prometheus/prometheus/pkg/labels"
 )
 
-// TsExemplar is an exemplar alongside a timestamp.
-type TsExemplar struct {
-	Ts       int64
-	Exemplar Exemplar
-}
-
-// SeriesExemplar is a time series of exemplar values.
-type SeriesExemplar []TsExemplar
-
-// The following are similar to those of `storage/interfaces.go`:
-// https://github.com/prometheus/prometheus/blob/91d7175eaac18b00e370965f3a8186cc40bf9f55/storage/interface.go#L31-L65
-
 // Exemplars is an exemplar storage backend.
 type Exemplars interface {
 	// Get is a fast lookup to find an exemplar with an exact match on labels
 	// and timestamp provided (so this can be O(1) lookup).
 	Get(l labels.Labels, t int64) (Exemplar, bool, error)
 
-	// Query is a query to find an exemplars with partial matches on labels
+	// GetRange is a query to find exemplars with exact matches on labels
 	// and timestamp ranges provided (this is a range query). Query returns
-	// a list of SeriesExemplar for each label match.
-	Query(l labels.Labels, start, end int64) ([]SeriesExemplar, error)
+	// a list of Exemplar for the label match.
+	GetRange(l labels.Labels, start, end int64) ([]Exemplar, error)
 
 	// Add an exemplar, it will be retained based on configuration at
 	// the construction of the exmplar storage.
--- a/pkg/exemplar/interface.go
+++ b/pkg/exemplar/interface.go
@@ -1,0 +1,39 @@
+// Copyright 2019 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package exemplar
+
+import (
+	"github.com/prometheus/prometheus/pkg/labels"
+)
+
+// The following are similar to those of `storage/interfaces.go`:
+// https://github.com/prometheus/prometheus/blob/91d7175eaac18b00e370965f3a8186cc40bf9f55/storage/interface.go#L31-L65
+
+// Exemplars is an exemplar storage backend.
+type Exemplars interface {
+	// Get is a fast lookup to find an exemplar with an exact match on labels
+	// and timestamp provided (so this can be O(1) lookup).
+	Get(l labels.Labels, t int64) (Exemplar, bool, error)
+
+	// Query is a query to find an exemplars with exact matches on labels
+	// and timestamp ranges provided (this is a range query).
+	Query(l labels.Labels, start, end int64) ([]Exemplar, error)
+
+	// Add an exemplar, it will be retained based on configuration at
+	// the construction of the exmplar storage.
+	Add(l labels.Labels, t int64, e Exemplar) error
+
+	// Close the storage.
+	Close() error
+} 


brian-brazil:
Exemplars aren't going to the TSDB, so shouldn't affect the appender interface.

juliusv:
[FYI: the build on this is broken due to an unrelated TypeScript error, looking into that]

shreyassrivatsan:
> Exemplars aren't going to the TSDB, so shouldn't affect the appender interface.

Is the plan to only store them in-memory in the scraper? It seems like it would be nice to be able to store them in a TSDB or a more persistent store..

brian-brazil:
Yes. The data volumes make storing on disk impractical.

shreyassrivatsan:
> Yes. The data volumes make storing on disk impractical.

@brian-brazil does this jive with your thoughts on the updated approach to pass through exemplars? https://github.com/prometheus/prometheus/pull/6309/files#diff-3bbf5d3205ec4ba63d9de26d3e31c776R24

The implementation itself for now is just a placeholder..

brian-brazil:
They shouldn't hit the storage interface at all.

robskillington:
@brian-brazil I think the new proposal from @shreyassrivatsan is just for in-memory collection. The existing storage/queryable interfaces are not changed/modified/touched.  The words "Storage" and "Queryable" don't need to be a part of the interfaces proposed, it's just they are generic names picked for reuse due to their similarity to other parts of the code base.

i.e. the proposal is some simple struct that receives exemplars, can cycle them out of memory based on a fixed size (or other attributes), and can be used to query/retrieve them for a matching label set.

brian-brazil:
We should avoid clashing with the names of existing important interfaces.

beorn7:
@shreyassrivatsan could you rebase this now that #6292 is merged?

beorn7:
Which should also fix test errors.

shreyassrivatsan:
> @shreyassrivatsan could you rebase this now that #6292 is merged?

Rebased and pushed just the interface changes..

pkg/exemplar/interface.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// The following are similar to those of `storage/interfaces.go`:
+// https://github.com/prometheus/prometheus/blob/91d7175eaac18b00e370965f3a8186cc40bf9f55/storage/interface.go#L31-L65
+
+// Exemplars is an exemplar storage backend.
+type Exemplars interface {
+	// Get is a fast lookup to find an exemplar with an exact match on labels
+	// and timestamp provided (so this can be O(1) lookup).
+	Get(l labels.Labels, t int64) (Exemplar, bool, error)
```
cstyan: `Get(...)` should return `[]Exemplar`, multiple exemplars could match `l` if not many labels are passed.

brian-brazil:
> How come an exemplar have it's own label set when it's associated with an existing timeseries?

It's for things a trace id for the exemplar.

pkg/exemplar/interface.go
shreyassrivatsan: So the idea was it has to be a full equality match on labels. The labels here describe the exact metric including all the labels and the name, so there should only be one. Otherwise the lookup is not that useful. We need to be able to find the exemplar corresponding to the metric.

cstyan: I can't speak for everyone else but I think we should follow the same pattern as existing queries; return all matches for the labels provided. That doesn't exclude users from passing labels that would match only a single exemplar, but is better than `error; query for exemplar would return more than one result` in the case of the current functions IMO.

shreyassrivatsan: Ok.. But in that case the return value will have to change.. It will additionally have to include information of the actual series matched as the exemplar is not useful without the series itself. So returns something like
```
type SeriesExemplar struct {
    l labels.Labels
    e Exemplar
}
```

and the method becomes `Get(l labels.Labels, t int64) ([]SeriesExemplar, bool, error)`

beorn7: I assume @shreyassrivatsan 's original idea was to keep things really as simple as possible. With exact matches, we can just have all the exemplars in a simple hash map. With label matching, we'll find ourselves once more in a situation where we have to implement a database (or do full scans of the exemplars in memory). Is that part of the plan?

Could we perhaps quickly think through the typical query pattern for exemplars? Will we commonly need label matching at all?

shreyassrivatsan: Correct. I was going for 1 - exact matches, so we can do simple hash lookups. I do not really see a use case of looking up more exemplars based on fewer labels. If one needs to do that it would be in the context of a metric query, so it would be implemented something like..

```
ls := querySeries() // returns []labels.Labels
for _, l := range ls {
  queryExemplar(l)
}
```

beorn7: Yeah, that's my thinking, too. But perhaps we are missing something. Let's see what the others say once they are awake again.

cstyan: I think a Grafana dashboard is going to be a pretty common usecase. If a dashboard panel has a single query for something like `http_request_duration_seconds`, without enough labels to narrow the query down to a single series, I don't think it's reasonable for Grafana to have to do the work to split all the results into individual label sets and make a query per unique series.

There's also the issue that many histogram queries aggregate away labels that we'd need to uniquely identify which exemplar to select.

@shreyassrivatsan can you elaborate on your use case and what M3 needs from the API, and hopefully we can get some more details from @davkal as well.

shreyassrivatsan: Ah. So I did not mean that grafana makes a query and then retrieves the exemplars. I meant a prometheus query will first query the series, get the exemplars and then apply functions.. I am simplifying a lot here, but essentially `/api/v1/query` and `/api/v1/query_range` will do

```
func query(query string) series {
  ls := querySeries() // returns []labels.Labels
  if exemplarsRequested {
    for _, l := range ls {
      queryExemplar(l)
    }
  }
  // apply functions
}
```

The resulting response will probably contain something maybe like the below 
```
{
   "status" : "success",
   "data" : {
      "resultType" : "vector",
      "result" : [
         {
            "metric" : {
               "__name__" : "up",
               "job" : "prometheus",
               "instance" : "localhost:9090"
            },
            "value": [ 1435781451.781, "1", {<exemplar>} ]
         }
      ]
  }
}
```


Eventually for full exemplar support , functions would have to be updated to handle exemplars. For example, aggregation functions can have a pick one policy.

beorn7: I think this approach makes a lot of sense. However, it requires to intertwine the existing API with exemplar support. My impression from the Prometheus side of things was that the plan is to implement something experimental that is a completely separate API so that consumers of that API can play with it and prove or disprove its usefulness. We had a few chats now with @juliusv and @davkal, wherein we realized that it is actually very hard for a visualization frontend like Grafana to craft dedicated queries for exemplars (for example in case the user clicks on a peak of a latency graph or on a tile of a histogram heatmap).

Returning exemplars alongside normal queries, if so requested, is solving that problem, but it's quite invasive (as the exemplar support has to be piped through the whole query engine, including smart decisions what to pick for aggregations). If we went down that alley, the exact-match method as suggested here would indeed make sense as illustrated by your pseudo code. (We might want to use a time range and not just a single timestamp in the signature then.)

While I like the idea, I'm pretty sure amending everything in the query engine is not feasible for an experimental feature. I'm still thinking about ways how to do something for the MVP that leaves the current query API and engine as is…

shreyassrivatsan: Ok, that makes sense. And yes, this is going more towards what will be the right solution if we decide to pipe this through the query engine. I still do think the exact timestamp makes for a faster implementation as we will have access to that as a part of the results for a time series query. 

For an experimental implementation using a time range and partial label match is probably a better approach, but implementing the backing store will be some more work. I made some changes to the interface. `Get` still does exact matches for a given timestamp. `Query` does partial label matches, takes a time range and returns a list of `SeriesExemplars`.

pkg/exemplar/interface.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+// TsExemplar is an exemplar alongside a timestamp.
+type TsExemplar struct {
+	Ts       int64
+	Exemplar Exemplar
+}
+
+// SeriesExemplar is a time series of exemplar values.
+type SeriesExemplar []TsExemplar
```
beorn7: Perhaps this should be a `struct` then?
```Go
type SeriesExemplar struct{
    Labels      labels.Labels
    TsExemplars []TsExemplar
}
```

In that way, the consumer can actually see what labels matched the labels in the query.

pkg/exemplar/interface.go
```diff
@@ -0,0 +1,49 @@
...
+// Exemplars is an exemplar storage backend.
+type Exemplars interface {
+	// Get is a fast lookup to find an exemplar with an exact match on labels
+	// and timestamp provided (so this can be O(1) lookup).
+	Get(l labels.Labels, t int64) (Exemplar, bool, error)
+
+	// Query is a query to find an exemplars with partial matches on labels
+	// and timestamp ranges provided (this is a range query). Query returns
+	// a list of SeriesExemplar for each label match.
+	Query(l labels.Labels, start, end int64) ([]SeriesExemplar, error)
```
beorn7: To really implement a generic query API, we needed regexp matchers and negative matchers, too.

That's where I'm thinking we should perhaps stay with the `Get` only approach, just that it takes a time range, not just one timestamp, and returns a time series of matchers for that one exact label set. The implementation of the externally visible API would need to hit the normal storage first to resolve the label matchers into concrete series first. I think that's fine for an MVP.

cstyan: Couldn't we do this via `Query(start, end int64, matchers ...*labels.Matcher)`?

pkg/exemplar/interface.go
beorn7: My idea for how the (MVP) query flow and UI in Grafana could look like:

The Grafana user creating a dashboard would be required to pair each graphing query with an additional "query", which would be required to just be a PromQL label selector without any aggregations or functions, e.g. `request_latency_bucket{code=~"2..",job="foo"}`. The exemplar API would take exactly that query string (i.e. a naked label matcher in PromQL syntax) and a time range to return exemplars, possibly sampled to avoid too many return values. Grafana could then use the provided string to query for exemplars upon a click event (could include the `le` label when in a histogram heatmap, but would also work without it when clicking on a latency graph). Further server-side filtering by exemplar or specific downsampling value would not be in the MVP but could be done in the browser for now.

A selection of exemplars with links to logs or traces could then be shown in a pop-up, or exemplars could show up as clickable "stars" in a latency graph.

@davkal does that make sense?

pkg/exemplar/interface.go
beorn7: Sure, but then we are really talking about implementing a whole query engine.
By doing this via a call to the existing query engine, we would waste a bit of resources (as we only want to get the matching label sets and would never access the actual time series data), but it would be much simpler.

beorn7: Also, we didn't need what I commented above (the separate `struct` to show in the answer which exact label set is attached to each returned series of exemplars).

shreyassrivatsan: One more round of changes. I got rid of `Query` for the time being as query in a way indicates partial matches. Added a `GetRange` which takes a range and returns a list of exemplars based on exact label matches.

beorn7:
Based on a couple of chats some people involved here had, I have to conclude that there are quite different opinions how this interface should look like. I was aiming for "as simple as possible, to be used as the ultimate underlying storage primitive", but one might as well make the argument that the storage interface should be rich in features and then the implementation might decide to re-use existing indexing and matching functionality or run their own.

It might just be wrong to start with the design of the internal interface. Perhaps it's better to get a document together for the big picture how everything plays together (exposition → ingestion → storage update on the one hand, and then Grafana panel → query for exemplars → storage retrieval).

brian-brazil:
I'd personally start with the API that Grafana wants, and then build the simplest possible thing to provide that. https://github.com/prometheus/prometheus/pull/6309#discussion_r358482680 sounds about right to me at first glance.

For an MVP implementation I'd suggest having them tied to targets similarly to how we do metadata, and brute forcing the lookup from there. We won't really know what Grafana wants until we can get an API they can play with, so I'd go for throwing something together over lots of up-front design.

brian-brazil:
@cstyan Where are we with this review?

cstyan:
@brian-brazil The interface here doesn't really prevent us from implementing the query API that was discussed with the Grafana team, but it makes the implementation of the API not as nice. I think it also limits options for those building long term storage ecosystem projects that may intend to store exemplars "forever".

I think what I've implemented in https://github.com/prometheus/prometheus/pull/6635 makes more sense given the end goals of how we see exemplars being useful to a system like Grafana dashboards.

codesome:
Superseded by #6635

PR #6313, opened 2019-11-13
React UI: Add /tsdb-status to React-handled routes

juliusv:


juliusv:
@boyskila @sharadgaur

sharadgaur:
Oh, I missed that.
 LGTM I think we need to update all the `ts` files where we are using useFeatch function. Thank you @juliusv

sharadgaur:
question: can we make type conversion via `const stats: TSDBMap = response.data;` this instead of  using `as` keywork like `const stats = response.data as TSDBMap;`

juliusv:
@sharadgaur Ah yep, @boyskila already did that as part of https://github.com/prometheus/prometheus/pull/6314/files, but thanks!

juliusv:
@csmarchbanks cool to merge?

PR #6314, opened 2019-11-13
unit test fixes

boyskila:


boyskila:
@juliusv

juliusv:
:+1: Thanks!

web/ui/react-app/src/pages/targets/ScrapePoolList.tsx
```diff
@@ -23,22 +23,22 @@ const filterByHealth = ({ upCount, targets }: ScrapePool, { showHealthy, showUnh
...
+  const { response, error } = useFetch<TargetsResponse>(`${pathPrefix}/api/v1/targets?state=active`);
 
   if (error) {
     return (
       <Alert color="danger">
         <strong>Error fetching targets:</strong> {error.message}
       </Alert>
     );
-  } else if (response && response.status !== 'success') {
+  } else if (response && response.status !== 'success' && response.status !== 'start fetching') {
```
juliusv: This seems a bit odd by the way. Shouldn't we just return a null or undefined response as long as it's not done fetching?

boyskila: Yes, indeed. useFetch initial state should be ```{ status: undefined, data: undefined }``` but returning undefined as a response will force us to make chained checks everywhere is used.

juliusv: Hm yeah, or would it not make sense to allow the whole `response` to be null as long as it't not fetched yet? E.g. via:

`response: APIResponse<T>;` ---> `response: APIResponse<T> | null;`

Seems like we check for `response` to be truthy anyway in the caller.

boyskila: If we initialize the state this way ```useState({ status: undefined, data: undefined })``` and
when the response is recieved we fallback again to this initial state in case response is undefined
```
const json: APIResponse<T> = await res.json();
setResponse(json || { status: undefined, data: undefined });
```

then I guess we can spread the response ```return { ...response, error, isLoading };``` without worry too much about what the status or data is.

After that, the caller will be able to consume whatever is returned without the need to check whether response exists and whether response exists and data is inside and so on

juliusv: That seems ok, but one more idea: Wouldn't it be nice if we could type the whole response in a way that TypeScript knows that `data` contains data exactly when `status` has the value `success`, and not otherwise? Like I did here for the query result differentiation: https://github.com/prometheus/prometheus/blob/12d347e4db04c926dab54f37e46eaca42f2ea685/web/ui/react-app/src/DataTable.tsx#L7-L26

Yeah, then the caller will still need to check whether the response exists, etc. but I think it would happen naturally anyway through the 4 different display sections:

```typescript
if (error) {
  // Render error
} else if (!response) {
  // Render spinner
} else if (response.status === 'error') {
  // Render API error.
} else {
  // Render data (access response.data directly, since TypeScript knows it has to be set for response.status !== 'error').
}
```

boyskila: Hm, not sure I understand that. Why I need to know the status value to know that data exist. Isn't enough just to check what data contains?

juliusv: Just checking what `data` contains is sufficient, but once one knows the value of `status`, one (and TypeScript) also knows whether data contains something or not, so one doesn't even need to check anymore whether `data` contains anything. (if the API response is typed in a way that the `QueryResult` example is, with the `status: 'success'` alternative having a `data` field and `status: 'error'` not having one.)

PR #6316, opened 2019-11-13
React UI: Fix tests harder

juliusv:
Again not sure why this passed last time (?), but now I was getting an
error about 'NaN' not being a valid value to assign to the 'height'
property of the input element. This changes it so that only the blur()
function is actually mocked out on the active input element.

### DIFF
--- a/web/ui/react-app/src/ExpressionInput.test.tsx
+++ b/web/ui/react-app/src/ExpressionInput.test.tsx
@@ -178,7 +178,7 @@ describe('ExpressionInput', () => {
       const downshift = expressionInput.find(Downshift);
       const instance: any = expressionInput.instance();
       const spyBlur = jest.fn();
-      instance.exprInputRef.current = { blur: spyBlur };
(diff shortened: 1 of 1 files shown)


juliusv:
@boyskila

juliusv:
Seems like this only affected my tests locally actually, since the tests over in https://github.com/prometheus/prometheus/pull/6313 are already passing without this, hmmmmm. Still good to fix it I guess.

boyskila:
Oh, this is not an real issue but is a wrong test setup instead. Good catch. Looks like doing this
```instance.exprInputRef.current = { blur: spyBlur }``` I'm overriding the whole object instead just to spy on blur.

juliusv:
Yep, exactly.

boyskila:
Does it really need of  "!" ```!blur``` I'm not facing any issue without it

juliusv:
Whoops, I thought it did, but it doesn't. Thanks, removed the `!`.

juliusv:
@boyskila this one good to merge?

PR #6320, opened 2019-11-14
React UI: Make link to classic UI work

juliusv:
It being a Reach Router <Link> caused the Reach router to not actually
leave the React app, even though the destination path was not a path
handled by the Reach Router.


juliusv:
@boyskila

PR #6321, opened 2019-11-14
React UI: Graph legend

boyskila:
This PR covers the graph legend functionalities toggle/hover and UI


boyskila:
@juliusv

web/ui/react-app/src/App.css
```diff
@@ -135,16 +135,42 @@ div.time-input {
...
+    display: inline-block;
+    cursor: pointer;
+}
+
+.legend-item {
+    display: flex;
+    align-items: center;
+    padding: 0 5px;
+    border-radius: 3px;
+    user-select: none;
```
juliusv: Can we still allow user-select? It was sometimes useful in the old UI when you want to copy a time series name.

web/ui/react-app/src/App.css
```diff
@@ -135,16 +135,42 @@ div.time-input {
...
 .graph-legend {
-    margin: 15px 0 15px 25px;
-    font-size: 0.8em;
+    margin: 15px 0 15px 55px;
+    font-size: 0.75em;
+    background: #222;
+    color: white;
+    padding: 10px 5px;
+    display: inline-block;
+    cursor: pointer;
```
juliusv: Hmm, I don't think we should be trying to copy the classic UI's legend styling here, just getting the series selection behavior to work. The old legend always looked a bit out of place in the non-dark page design. I think the only reason it looked like it did is because originally I copied it from a Rickshaw graphing library example like https://tech.shutterstock.com/rickshaw/examples/legend.html.

Of course we have to add some hover styles etc., but maybe just adjust them for the light design?

web/ui/react-app/src/Graph.tsx
```diff
@@ -28,11 +37,20 @@ interface GraphProps {
     resolution: number;
   } | null;
 }
+interface GraphState {
```
juliusv: Nit: Add empty line between interface definitions.

web/ui/react-app/src/Graph.tsx
```diff
@@ -19,6 +19,15 @@ function getGraphID() {
   return graphID++;
 }
 
+export type GraphSeriesValue = number | null;
+
+export interface GraphSeriesXY {
```
juliusv: I think the `XY` part of the name is unnecessary, maybe just `GraphSeries`?

web/ui/react-app/src/Graph.tsx
```diff
@@ -260,12 +291,27 @@ class Graph extends PureComponent<GraphProps> {
...
       <div className="graph">
-        <ReactResizeDetector handleWidth onResize={() => this.plot()} />
+        <ReactResizeDetector handleWidth onResize={this.plot} />
         <div className="graph-chart" ref={this.chartRef} />
-        <Legend series={this.getData()} />
+        <div className="graph-legend" onMouseOut={canUseHover ? this.onLegendMouseOut : undefined}>
+          {series.map(({ index, color, labels }) => (
+            <div
+              style={{ opacity: selectedSeriesIndex > -1 && index !== selectedSeriesIndex ? 0.4 : 1 }}
+              onClick={series.length > 1 ? this.onSeriesSelect(index) : undefined}
```
juliusv: I believe renaming `onSeriesSelect` -> `handleSeriesSelect` would be more in line with React event handler naming conventions in general and in this codebase (same for `onSeriesHover`). The other convention I've seen is to omit the `on` / `handle` completely and just say `selectSeries`.

web/ui/react-app/src/Graph.tsx
```diff
@@ -28,11 +37,20 @@ interface GraphProps {
     resolution: number;
   } | null;
 }
+interface GraphState {
+  selectedSeriesIndex: number;
```
juliusv: It'd seem cleaner to me to use `number | null` or an optional property than using a special sentinel value `-1` of the same numeric domain for checking whether there is a series index.

web/ui/react-app/src/Graph.tsx
```diff
@@ -217,25 +237,25 @@ class Graph extends PureComponent<GraphProps> {
...
-    if (this.chartRef.current === null) {
+  plot = () => {
+    if (!this.chartRef.current) {
       return;
     }
+    const selectedData = this.getData()[this.state.selectedSeriesIndex];
     this.destroyPlot();
-    $.plot($(this.chartRef.current!), this.getData(), this.getOptions());
-  }
+    $.plot($(this.chartRef.current), selectedData ? [selectedData] : this.getData(), this.getOptions());
```
juliusv: By just plotting the selected series like this, the Y axis range changes to the range of the single selected series. In the old UI, the Y axis didn't update. I think either behavior can have an advantage at different times (e.g. zooming in into a small series vs. not having the Y axis jump around so you don't lose your orientation).

I'm not sure yet what to do here, but we might also want to keep the old behavior here to keep the graph range fixed to the entire data range, independent of individual series visibility.

web/ui/react-app/src/Graph.tsx
```diff
@@ -28,11 +37,20 @@ interface GraphProps {
     resolution: number;
   } | null;
 }
+interface GraphState {
+  selectedSeriesIndex: number;
+  hoveredSeriesIndex: number;
```
juliusv: I see hover and single-series selection features implemented here. The old UI also had the feature to turn off selected series (this is what the checkmark symbol was for in the old UI). Were you still planning on adding that? I wonder if people actually use it... in case we decide to not add that feature, I don't think we need the checkmarks.

web/ui/react-app/src/App.css
boyskila: Yes, sure. I will remove it.

web/ui/react-app/src/App.css
boyskila: Ok, let's style it in another PR.

web/ui/react-app/src/Graph.tsx
boyskila: Ok, I'll rename it.

web/ui/react-app/src/Graph.tsx
boyskila: Make sense, yes. This is why a style guide is needed :)

web/ui/react-app/src/Graph.tsx
boyskila: Hm, I didn't notice the "off" state on checkbox click. I thought is related to the "selected" state of the row, but yes, ok, I'll added it if you think it is usable.

juliusv: Hm yeah, I'm also not sure whether we still need it. How about we omit that feature for now, and we can still add it later if users really ask for it. Less features is always good when they aren't used :)

But then I think we don't need the checkmark icons at all for now.

web/ui/react-app/src/Graph.tsx
roidelapluie: Personnally the old behaviour always annoyed me :)

juliusv: @roidelapluie Thanks, that's the feedback I was looking for :) I also agree, let's just keep it with the new behavior for now, which is likely actually going to be more useful for people. And if people really complain, we can see.

web/ui/react-app/src/App.css
juliusv: I don't think we should be introducing a completely new styling in this PR though just to roll it back again. Or do you mean you were still planning on removing the new dark style in this PR?

boyskila: Ok, I'll revert the old styles. Are there any plans to be styled differently?

web/ui/react-app/src/types/index.d.ts
```diff
@@ -0,0 +1,64 @@
+declare namespace jquery.flot {
```
juliusv: Cool!

web/ui/react-app/src/types/types.ts
```diff
@@ -0,0 +1,22 @@
+export type Nullable<T> = T | null;
+
+export interface GraphSeries {
```
juliusv: I generally wonder what the motivation is to move some of these types to a different file - as long as they are only used in one file (Graph.tsx), wouldn't they be better / easier kept and maintained there?

web/ui/react-app/src/App.css
juliusv: No, I mean we only need new styles for the new functionality (hover highlight, selected series, etc.), but not change the whole design otherwise.

web/ui/react-app/src/Graph.tsx
```diff
@@ -3,94 +3,86 @@ import React, { PureComponent } from 'react';
...
+  };
 
-  formatValue = (y: number | null): string => {
+  formatValue = (y: Nullable<number>): string => {
     if (y === null) {
       return 'null';
     }
-    const abs_y = Math.abs(y);
-    if (abs_y >= 1e24) {
+    const absY = Math.abs(y);
```
juliusv: Heh yeah, these somehow made it over from the old code with old naming :P Thanks for fixing.

web/ui/react-app/src/Graph.tsx
```diff
@@ -160,45 +157,45 @@ class Graph extends PureComponent<GraphProps> {
...
         }
       }
+      const { r, g, b } = colors[index];
 
       return {
         labels: ts.metric !== null ? ts.metric : {},
-        data: data,
-        color: colors[index],
-        index: index,
+        color: `rgba(${r}, ${g}, ${b}, ${hoveredSeriesIndex === -1 || hoveredSeriesIndex === index ? 1 : 0.5})`,
```
juliusv: `=== -1` needs to be `=== null` now.

web/ui/react-app/src/types/types.ts
boyskila: Yes, I'm kind of experimenting. I was wonder whether would it not be better if we move those interfaces (not Props and State) in dedicated file. Thus we will have a better look what we have as interfaces and components will shrink a bit. WDYT?

web/ui/react-app/src/types/types.ts
```diff
@@ -0,0 +1,22 @@
...
+  labels: { [key: string]: string };
+  color: string;
+  data: Nullable<number>[][]; // [x,y][]
+  index: number;
+}
+
+export interface GraphMetric {
+  __name__: string;
```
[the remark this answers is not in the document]
boyskila: Ah, ok. I guess Metric interface in DataTable can be used as well

web/ui/react-app/src/types/types.ts
```diff
@@ -0,0 +1,22 @@
+export type Nullable<T> = T | null;
```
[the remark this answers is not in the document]
boyskila: Not sure whether is worth in this case but I found declarative approach more elegant and readable.

web/ui/react-app/src/Graph.tsx
boyskila: Fixed.Thanks!

web/ui/react-app/src/Graph.tsx
boyskila: Yeah, and eslint it would take my eyes out :D

web/ui/react-app/src/types/types.ts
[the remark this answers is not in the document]
juliusv: Yes, right, `Metric` in `DataTable.tsx` would actually be a good example for a shared type that can live in a separate file like this, if we reuse it from multiple places. We also have `Labels` in `targets.ts`, which is the same type, and can be consolidated. Of course one could also argue that even defining such a simple type (`{ [key: string]: string }`) is over-architecting things, but I think either way is ok.

web/ui/react-app/src/types/types.ts
juliusv: Generally I think I'm a fan of keeping types local to a component if they are only used by the component, and only separate them out when they are really shared.

web/ui/react-app/src/types/types.ts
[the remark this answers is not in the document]
juliusv: Both approaches are declarative IMO. `T | null` has the advantage of not introducing a new type definition that I have to think about (how is it implemented, let me check it, is it correct, does it do anything more than just OR-ing?). `Nullable<T>` has the advantage of being more verbal about the intent. I guess I prefer `T | null`, but don't feel strongly either way, so happy to defer to your judgement for now.

web/ui/react-app/src/Graph.tsx
```diff
@@ -160,45 +157,46 @@ class Graph extends PureComponent<GraphProps> {
...
         }
       }
+      const { r, g, b } = colors[index];
 
       return {
         labels: ts.metric !== null ? ts.metric : {},
-        data: data,
-        color: colors[index],
-        index: index,
+        color: `rgba(${r}, ${g}, ${b}, ${hoveredSeriesIndex === null || hoveredSeriesIndex === index ? 1 : 0.5})`,
```
juliusv: Playing with it a bit more, let's do `0.5` -> `0.3` to make highlighted series stand out even more?

juliusv:
Loving the behavior now besides nits!

juliusv:
:+1: (ignoring broken Travis tests because those seem to be Travis' fault)

PR #6323, opened 2019-11-15
Add tsdb/tsdb to .gitignore

beorn7:
This is built this way by the Makefile. One might ask if that's the
correct location for the `tsdb` tool or if it should be placed
top-level together with `prometheus` and `promtool`. But for now, it
should be excluded from Git like the latter two.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 
 /prometheus
 /promtool
+/tsdb/tsdb
 benchmark.txt
 /data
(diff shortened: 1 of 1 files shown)


simonpasquier:
See https://github.com/prometheus/prometheus/pull/6088 which should move the `tsdb` commands to `promtool`. Unfortunately this is depending on #5887 (tsdb import command).

PR #6324, opened 2019-11-15
SD moratorium has been lifted.

cstyan:
As per the dev summit, we agreed to lift the moratorium on new SD implementations provided that new implementations have a committed maintainer who has push access to prometheus/prometheus.

cc @brian-brazil @SuperQ 

### DIFF
--- a/discovery/README.md
+++ b/discovery/README.md
@@ -2,10 +2,6 @@
 
 This directory contains the service discovery (SD) component of Prometheus.
 
-There is currently a moratorium on new service discovery mechanisms being added
-to Prometheus due to a lack of developer capacity. In the meantime `file_sd` 
-remains available.
-
 ## Design of a Prometheus SD
 
 There are many requests to add new SDs to Prometheus, this section looks at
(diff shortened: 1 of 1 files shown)


discovery/README.md
```diff
@@ -2,9 +2,8 @@
 
 This directory contains the service discovery (SD) component of Prometheus.
 
-There is currently a moratorium on new service discovery mechanisms being added
-to Prometheus due to a lack of developer capacity. In the meantime `file_sd` 
-remains available.
+Note: new service discovery implementations will be accepted if they have a
```
brian-brazil: That's only one requirement, we still have all the sanity checks listed below. I'd add this as one more requirement, rather than making it seem like the only requirement.

cstyan: Updated to clarify that it's an additional requirement.

discovery/README.md
```diff
@@ -19,6 +15,10 @@ use across multiple organizations. It should allow discovering of machines
 and/or services running somewhere. When exactly an SD is popular enough to
 justify being added to Prometheus natively is an open question.
 
+Note: As part of lifting the past moratorium on new SD implementations it was
```
brian-brazil: i.e.

Do we need to even mention the moratorium here? It's historical now.

cstyan: I think there's value in keeping mention of it for some period of time. Even if we had an announcement on the mailing list not everyone's going to read it, explicitly saying the moratorium is over and there's a new requirement for new SD implementations is clearer.

brian-brazil: You still need to fix i.e.

PR #6325, opened 2019-11-15
React UI: Change the dynamic react key with static id

boyskila:
In this PR the dynamic react keys for the panels on the graph page are changed with static ID's. During refactoring I found another bug in parseParams which causes removing on the last panel. This is fixed as well. All other changes are just cosmetic changes and nothing functional is touched there. @juliusv

### DIFF
--- a/web/ui/react-app/src/utils/urlParams.test.ts
+++ b/web/ui/react-app/src/utils/urlParams.test.ts
@@ -1,5 +1,6 @@
-import { decodePanelOptionsFromQueryString, encodePanelOptionsToQueryString } from './urlParams';
+import { decodePanelOptionsFromQueryString, encodePanelOptionsToQueryString, parseOption, toQueryString } from './urlParams';
 import { PanelType } from '../Panel';
+import moment from 'moment';
 
 const panels: any = [
   {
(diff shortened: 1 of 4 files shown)


web/ui/react-app/src/utils/urlParams.test.ts
```diff
@@ -33,7 +34,61 @@ describe('decodePanelOptionsFromQueryString', () => {
...
+  describe('step_input', () => {
+    it('should return step_input parsed if > 0', () => {
+      expect(parseOption('step_input=2')).toEqual({ resolution: 2 });
+    });
+    it('should return empty object if step is equal 0', () => {
+      expect(parseOption('step_input=0')).toEqual({});
+    });
+  });
+  describe('range_input', () => {
+    it('should return range parsed if its ot null', () => {
```
juliusv: "its ot" -> "it's not"

web/ui/react-app/src/utils/urlParams.test.ts
```diff
@@ -33,7 +34,61 @@ describe('decodePanelOptionsFromQueryString', () => {
...
+    it('should return empty object for invalid value', () => {
+      expect(parseOption('range_input=h')).toEqual({});
+    });
+  });
+  describe('Parse type param', () => {
+    it('should return "graph" if tab value is 0', () => {
+      expect(parseOption('tab=0')).toEqual({ type: 'graph' });
+    });
+    it('should return "table" if tab value is 1', () => {
+      expect(parseOption('tab=0')).toEqual({ type: 'graph' });
```
juliusv: Here you still need to adjust the values copied from the previous test to be table instead of graph.

juliusv: Better to use `PanelType.Graph` and `PanelType.Table` here instead of direct strings?

web/ui/react-app/src/utils/urlParams.test.ts
```diff
@@ -33,7 +34,61 @@ describe('decodePanelOptionsFromQueryString', () => {
...
+    });
+  });
+});
+
+describe('toQueryString', () => {
+  it('should generate query string from panels options', () => {
```
juliusv: nit: "panels options" -> "panel options"

web/ui/react-app/src/utils/urlParams.ts
```diff
@@ -1,127 +1,85 @@
...
+export const toQueryString = ({ key, options }: PanelMeta) => {
+  const formatWithKey = formatParam(key);
+  const { expr, type, stacked, range, endTime, resolution } = options;
+  const time = isPresent(endTime) ? formatTime(endTime) : false;
+  const urlParams = [
+    formatWithKey('expr', expr),
+    formatWithKey('tab', type === PanelType.Graph ? 0 : 1),
+    formatWithKey('stacked', stacked ? 1 : 0),
+    formatWithKey('range_input', formatRange(range)),
+    time ? `${formatWithKey('end_input', time)}&${formatWithKey('moment_input', time)}` : '',
```
juliusv: Hmm I know it's been this way before (in the new UI, not in classic UI), but I wonder if we actually want this behavior that the table and graph end times are synced. But ok to keep it like this for now.

juliusv:
Cool, seems to work great, thanks! Just a few nit comments.

boyskila:
Thanks for the review @juliusv ! I think is ready to be merged.

juliusv:
:+1: Thanks!

PR #6326, opened 2019-11-16
Port tsdb to use pkg/labels.

tomwilkie:
This removes a bunch of conversions and duplicate code.


pkg/labels/labels.go
```diff
@@ -201,6 +201,23 @@ func (ls Labels) Has(name string) bool {
 	return false
 }
 
+// WithoutEmpty returns the labelset without empty labels.
+// May return the same labelset.
+func (ls Labels) WithoutEmpty() Labels {
+	for _, v := range ls {
+		if v.Value == "" {
```
bwplotka: minor: 
```
if v.Value != "" {
   continue
}
```
.. would make a complex path the "main one" in this function (aka filter boring cases first) (:

pkg/labels/matcher.go
```diff
@@ -86,3 +95,18 @@ func (m *Matcher) Matches(s string) bool {
...
+	case MatchEqual:
+		return NewMatcher(MatchNotEqual, m.Name, m.Value)
+	case MatchNotEqual:
+		return NewMatcher(MatchEqual, m.Name, m.Value)
+	case MatchRegexp:
+		return NewMatcher(MatchNotRegexp, m.Name, m.Value)
+	case MatchNotRegexp:
+		return NewMatcher(MatchRegexp, m.Name, m.Value)
+	}
+	panic("labels.Matcher.Matches: invalid match type")
```
bwplotka:
```suggestion
	panic("labels.Matcher.Inverse: invalid match type")
```

bwplotka: Also not sure why panic if we have error handling piped already (:

brian-brazil: It's a coding bug, so a panic isn't inappropriate.

Issue #6333, opened 2019-11-18
Remote Write Meta Issue


cstyan:
I'd like to collect issues around the current status of remote write so that we can better track what's still a problem, what we might want to add or improve, and dis

- [ ] 1. Revisit remote write/queue config defaults, bump samples per send/lower max shards/bump min shards.
    - [ ] add metric for remote write bandwidth
- [ ] 2. Resharding still appears to not behave correctly after restart if WAL is large.
    - [ ] add more logging around remote write's replaying of the WAL
    - [ ] come up with testing/reproduction strategy
- [ ] 3. Keep a checkpoint per remote write queue so we can attempt restart from that point in the WAL on restarts.
- [ ] 4. Documentation work; I've already started a blog post for troubleshooting remote write and turning knobs on the queue config.
- [ ] 5. Streaming remote write. 

More details: 

1. We'll want to be able to maintain the same max samples/s output with the new defaults, but bumping the max samples per send would lower the shard requirements, and upsharding can lead to DDOS like issues since we have no back pressure mechanism (related issues: https://github.com/prometheus/prometheus/issues/5166)
2. My current theory is that, if Prometheus restarts just prior to a time at which it would write a new block and truncate the WAL, the WAL is large enough that replay within remote write takes long enough that the build up of samples (since scraping has started) is enough that remote write will always be minutes behind the WAL. I believe the weight of the samples pending within the calculation isn't enough, but again this is just my theory.
3. Right now remote write starts sending samples with a timestamp after `time.Now()` when Prometheus started (more or less). On restart or config reload if remote write wasn't able to flush it's pending samples cleanly or was behind for some reason there will be samples missing in the remote storage. If remote write always starts from a checkpoint in some kind of checkpoint file for it's queue name/config hash we can avoid that situation.
4. There's still open issues that we could address via just providing better/more easily accessible documentation for remote write (related issues: https://github.com/prometheus/prometheus/issues/5709).
5. Is this something we want to do? Tom brought up that if we're going to do this we might as well just stream the WAL entirely, since the goal is to reduce network overhead this would eliminate the need to send labels with every sample. The downside to this work would be that we're putting more implementation work on the server side instead of in the client. Existing implementations may no longer work. Would this be remote write v2?

@csmarchbanks @tomwilkie @gouthamve @bboreham

brian-brazil:
@valyala had a request around 1 iirc.

[...]

PR #6337, opened 2019-11-19
Log the start and end of the WAL replay within the WAL watcher.

cstyan:
Part of #6333

Without having to turn on debug logging, it would be useful to know how long WAL replay took within the WAL watcher, especially relative to replaying the WAL for TSDB head block

@csmarchbanks 

### DIFF
--- a/tsdb/wal/watcher.go
+++ b/tsdb/wal/watcher.go
@@ -68,8 +68,9 @@ type Watcher struct {
 	metrics        *WatcherMetrics
 	readerMetrics  *liveReaderMetrics
 
-	StartTime   int64
-	lastSegment int
+	startTime      time.Time
+	startTimestamp int64 // the start time as a Prometheus timestamp
+	sendSamples    bool
 
 	recordsReadMetric       *prometheus.CounterVec
 	recordDecodeFailsMetric prometheus.Counter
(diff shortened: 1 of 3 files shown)


tsdb/wal/watcher.go
```diff
@@ -483,6 +488,10 @@ func (w *Watcher) readSegment(r *LiveReader, segmentNum int, tail bool) error {
 			}
 			for _, s := range samples {
 				if s.T > w.StartTime {
+					if !sentSamples && segmentNum == w.lastSegment {
```
csmarchbanks: Why the check for `segmentNum == w.lastSegment` here? I would have thought that would be controlled by `tail`?

cstyan: `w.lastSegment` is the last/newest segment that existed on startup. We only want to log `done replaying WAL` once, when we find a sample with a timestamp `s.T > w.StartTime` in the newest segment that existed when the watcher started. It's possible that new segments were created during the replay, but even if there weren't we'd still end up logging this for every new segment that's created while tailing the WAL.

csmarchbanks: I think it would be simpler to store `sentSamples` rather than `lastSegment` on the `Watcher` struct. Then there is only one check that is more obvious. WDYT?

Otherwise, I think `lastSegment` needs to be named something else as I would expect a variable with that name to change over time as more segments are created.

tsdb/wal/watcher.go
```diff
@@ -483,6 +488,10 @@ func (w *Watcher) readSegment(r *LiveReader, segmentNum int, tail bool) error {
...
+						sentSamples = true
+						level.Info(w.logger).Log("msg", "done replaying WAL")
```
csmarchbanks: Do you think it is worthwhile to log the elapsed time here rather than having to look for two log entries?

cstyan: yep!

tsdb/wal/watcher.go
cstyan: I changed it to just `sendSamples`, which is set to true when we find a sample after the `StartTime`.

PR #6339, opened 2019-11-19
Added an enabled monitor label to file_sd

chicknsoup:
Sometimes, we want to temporary disable the scrape on some specific targets. The idea is to add a label check to file_sd, whenever a label named enabled has the value "false", the targetGroup will be ignored.

### DIFF
--- a/discovery/file/file.go
+++ b/discovery/file/file.go
@@ -71,6 +71,7 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 }
 
 const fileSDFilepathLabel = model.MetaLabelPrefix + "filepath"
+const enabledLabel = "enabled"
 
 // TimestampCollector is a Custom Collector for Timestamps of the files.
 type TimestampCollector struct {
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. This is business logic, which doesn't belong in an SD implementation. You can do this with relabelling.

chicknsoup:
But relabeling will not work. For example, in one file we have hundreds of targets with different sets of labels, how would you relabel to disable scraping some targets for some time (1hour, 1 day, etc.). Moreover, relabeling requires configuration reloading or prometheus restarting while changing properties of labels does not.

chicknsoup:
> 
> 
> Thanks for your PR. This is business logic, which doesn't belong in an SD implementation. You can do this with relabelling.

But relabeling will not work. For example, in one file we have hundreds of targets with different sets of labels, how would you relabel to disable scraping some targets for some time (1hour, 1 day, etc.). Moreover, relabeling requires configuration reloading or prometheus restarting while changing properties of labels does not.

brian-brazil:
This is all possible, but this isn't the best place to discuss it.

It makes more sense to ask questions like this on the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users) rather than in a GitHub issue. On the mailing list, more people are available to potentially respond to your question, and the whole community can benefit from the answers provided.

PR #6344, opened 2019-11-19
Add a metric to track total bytes sent per remote write queue.

cstyan:
It would be useful for troubleshooting and tracking performance of remote write to have a metric to track the bandwidth of a remote write queue. For #6333.

@csmarchbanks 

cc @tomwilkie

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -170,6 +170,15 @@ var (
 		},
 		[]string{queue},
 	)
+	bytesSent = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "sent_bytes_total",
+			Help:      "The total number of bytes sent by the queue.",
+		},
+		[]string{queue},
+	)
 )
 
 // StorageClient defines an interface for sending a batch of samples to an
@@ -224,6 +233,7 @@ type QueueManager struct {
 	maxNumShards               prometheus.Gauge
 	minNumShards               prometheus.Gauge
 	desiredNumShards           prometheus.Gauge
+	bytesSent                  prometheus.Counter
 }
 
 // NewQueueManager builds a new QueueManager.
@@ -331,6 +341,7 @@ func (t *QueueManager) Start() {
 	t.maxNumShards = maxNumShards.WithLabelValues(name)
 	t.minNumShards = minNumShards.WithLabelValues(name)
 	t.desiredNumShards = desiredNumShards.WithLabelValues(name)
+	t.bytesSent = bytesSent.WithLabelValues(name)
 
 	// Initialise some metrics.
 	t.shardCapacity.Set(float64(t.cfg.Capacity))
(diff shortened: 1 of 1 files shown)


storage/remote/queue_manager.go
```diff
@@ -170,6 +170,15 @@ var (
 		},
 		[]string{queue},
 	)
+	bytesSent = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "bytes_sent_total",
+			Help:      "The total number of bytes sent by the queue",
```
csmarchbanks: Full stop here.

storage/remote/queue_manager.go
```diff
@@ -170,6 +170,15 @@ var (
 		},
 		[]string{queue},
 	)
+	bytesSent = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "bytes_sent_total",
```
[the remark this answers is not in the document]
juliusv: Yeah, usually try to do `<foo>_<unit>_total` (for counters) / `<foo>_<unit>` (for gauges).

storage/remote/queue_manager.go
```diff
@@ -801,23 +812,24 @@ func (s *shards) sendSamples(ctx context.Context, samples []prompb.TimeSeries, b
 	// should be maintained irrespective of success or failure.
 	s.qm.samplesOut.incr(int64(len(samples)))
 	s.qm.samplesOutDuration.incr(int64(time.Since(begin)))
+	s.qm.bytesSent.Add(float64(n))
```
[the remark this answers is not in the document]
cstyan: :man_facepalming: yeah, moving this

cstyan:
> A couple small comments. I am also curious, what does tracking bytes buy us over tracking the number of samples sent?

Samples sent doesn't give us any insight in to the actual amount of network traffic remote write is using. For troubleshooting or tuning remote write to match the resources available, it could be useful to know what kind of network bandwidth remote write needs to keep up with the samples you want to send.

It would also be useful to, if we implement streaming remote write, compare any difference in total bytes sent. For example with streaming we may end up with an implementation that no longer requires us to send labels with every sample.

storage/remote/queue_manager.go
```diff
@@ -827,6 +838,7 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
 
 		if err == nil {
 			s.qm.succeededSamplesTotal.Add(float64(len(samples)))
+			s.qm.bytesSent.Add(float64(len(*buf)))
```
csmarchbanks: Nit: since we send `req` not `*buf` could we use that here as well?

cstyan: Sure. 

Do you think there's enough value in having two metrics, one using `*buf` and one usin `req`, so that we could see the difference in overhead per sample with any config changes or future streaming remote write?

csmarchbanks: Are you suggesting tracking `cap(req) - len(req)`? If so I don't think it's worth it (yet?).

Should this be a histogram rather than a counter so that it is also possible to see the size of requests?

cstyan: I was suggesting we track the size of the data vs the size of the data wrapped up as a write request, though that actually wouldn't be as simple change as I'd have liked because of the compression.

And we're actually setting `*buf = req`.

csmarchbanks: Ahh ok, yeah I was confused about tracking `*buf` and `req` since they are the same. I think this PR is good to start with, the nit is just that `*buf` is only being used at the top of the function which is kinda nice.

csmarchbanks:
Travis has been removed, so merging despite build failure there.

PR #6347, opened 2019-11-20
Removed compilation step inside main_test.

jkohen:
Advantages:
* Speeds up the test slightly.
* One binary for the `main` function and test, which is how every other test works, and it ensures they are compiled identically.
* The test code is simpler.

### DIFF
--- a/cmd/prometheus/main_test.go
+++ b/cmd/prometheus/main_test.go
@@ -15,7 +15,6 @@ package main
 
 import (
 	"context"
-	"flag"
 	"fmt"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


fabxc:
@brian-brazil could you take a look at this?

brian-brazil:
I think this might be better controlled by a command line flag, which we can strip off before calling main()

cmd/prometheus/main_test.go
```diff
@@ -31,37 +30,17 @@ import (
...
-	build := exec.Command("go", "build", "-o", promPath)
-	output, err := build.CombinedOutput()
-	if err != nil {
-		fmt.Printf("compilation error :%s \n", output)
-		os.Exit(1)
-	}
-
-	exitCode := m.Run()
-	os.Remove(promPath)
-	os.RemoveAll(promData)
```
roidelapluie: That should still happen.

brian-brazil:
Done by #6658

jkohen:
Thanks all for the review.

Issue #5861, opened 2019-08-13
Prefixing exported labels with `exported_` has problematic edge cases


beorn7:
This is extracted from #5845 .

When Prometheus adds target labels to a scrape, conflicts with existing labels may arise and have to be solved. The [docs](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config) specify the following: “… label conflicts are resolved by renaming conflicting labels in the scraped data to "exported_<original-label>" (for example "exported_instance", "exported_job") and then attaching server-side labels.”

1. What happens if the target labels contain both `foo="a"` and `exported_foo="b"` and then a scrape contains its own label `foo="c"`? End result could be:
   1. `foo="a",exported_foo="b"
   1. `foo="a",exported_foo="c"
   1. `foo="a",exported_foo="b",exported_exported_foo="c"
2. What happens if a scrape contains both `foo="a"` and `exported_foo="b"` and then there is a target label `foo="c"`? End result could be:
   1. `foo="c",exported_foo="a"
   2. `foo="c",exported_foo="b"
   3. `foo="c",exported_foo="a",exported_exported_foo="b"

In case 2, the current code will do i. But that's not precisely specified in the docs. One might also expect iii.

Case 1 is more interesting, as the behavior depends on the order the target labels show up in the iteration. It could be i. or iii. That's in any case an undesired behavior.

For double fun, case 1 and case 2 could be combined (both the scrape and the target labels contain both a `foo` and an `exported_foo` label).

While this is unlikely to ever show up, we should do something about it. The most conservative approach would be to specify in the docs that the behavior is undefined if exported labels and/or target labels already have a prefix `exported_`. The most invasive change would be to go for case iii., essentially prefixing `exported_` to exported label as often as necessary to avoid any name collisions. That's technically a breaking change, but one might argue that's what the docs so far suggest (in a very vague fashion).

For the record, this has never reported to be a problem in many years of widespread Prometheus usage.

Transwert:
hey @beorn7 @brian-brazil  can i work on this issue, and it will be good awesome thing, if you can give me pointer on how to work on it.

beorn7:
Sure, your contribution would be very much appreciated. I'm not sure, though, if this is a good first issue if you have never worked with the Prometheus codebase (despite the "low hanging fruit" label that Brian added).

[...]

PR #6350, opened 2019-11-20
Fix relabaling collision when using exported labels

geobeau:
Linked to #5845 

When merging target labels and sample labels, we now check first if the label exists before creating the exported one. The previous behaviour was causing a label for no obvious reason.

The issue was:
```
target_labels : {foo: "val", exported_foo: "another_val"}
sample_labels: {}
```
After ingestion it became:
```
sample_labels : {foo: "val"}
```

This is due to the fact that Prometheus will first process `exported_foo`:
```
sample_labels : {foo: "val", exported_foo: "another_val", exported_foo: ""}
```
then foo:
```
sample_labels : {foo: "val", exported_foo: "", exported_foo: ""}
```
resulting in:
```
sample_labels : {foo: "val"}
```

Note that since labels are computed in order, the labels wouldn't be impacted for this set of labels:
```
target_labels : {bar: "val", exported_bar: "another_val"}
```

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -451,9 +451,9 @@ func mutateSampleLabels(lset labels.Labels, target *Target, honor bool, rc []*re
 		for _, l := range target.Labels() {
 			// existingValue will be empty if l.Name doesn't exist.
 			existingValue := lset.Get(l.Name)
-			// Because setting a label with an empty value is a no-op,
-			// this will only create the prefixed label if necessary.
-			lb.Set(model.ExportedLabelPrefix+l.Name, existingValue)
+			if existingValue != "" {
+				lb.Set(model.ExportedLabelPrefix+l.Name, existingValue)
+			}
 			// It is now safe to set the target label.
 			lb.Set(l.Name, l.Value)
 		}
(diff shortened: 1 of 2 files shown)


beorn7:
This is a bit more complicated, see #5861 .

geobeau:
Thanks for your answer! I missed this issue.

While it doesn’t fix the edge cases, it should fix the breaking change introduced in 2.12

beorn7:
@brian-brazil might have a more elaborated opinion here.

brian-brazil:
:+1: 

It looks like this is technically better, while not solving the whole problem.

geobeau:
What do you think @beorn7, should we wait for a fix of #5861 ?

scrape/scrape.go
```diff
@@ -1106,6 +1106,7 @@ loop:
 
 			// Hash label set as it is seen local to the target. Then add target labels
 			// and relabeling and store the final label set.
+
```
beorn7: Was this blank line added on purpose?

geobeau: Done

beorn7:
Thanks.

PR #6351, opened 2019-11-21
Add a maintainers file to each discovery method

SuperQ:
Add Brian as the initial maintainer for each of the service discovery
packages.

Followup to https://github.com/prometheus/prometheus/pull/6324


brian-brazil:
We'll have a top-level maintainers covering everything, as I agree with @beorn7 on that point. Sorting this properly is already near the top of my todo list.

PR #6355, opened 2019-11-21
Add Benchmark for the PromQL Parser

slrtbtfs:

This PR adds Benchmarks for the PromQL parser.

These will allow to better estimate how upcoming parser changes affect performance.

### DIFF
--- a/promql/bench_test.go
+++ b/promql/bench_test.go
@@ -220,3 +220,47 @@ func BenchmarkRangeQuery(b *testing.B) {
 		})
 	}
 }
+
+func BenchmarkParser(b *testing.B) {
+	cases := []string{
+		"a",
+		"metric",
+		"1",
+		"1 >= bool 1",
+		"1 + 2/(3*1)",
+		"foo or bar",
+		"foo and bar unless baz or qux",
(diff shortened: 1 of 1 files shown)


promql/bench_test.go
```diff
@@ -220,3 +220,47 @@ func BenchmarkRangeQuery(b *testing.B) {
...
+	}
+	for _, c := range errCases {
+		name := fmt.Sprintf("%s (should fail)", c)
+		b.Run(name, func(b *testing.B) {
+			b.ReportAllocs()
+			for i := 0; i < b.N; i++ {
+				ParseMetric(c)
+			}
+		})
+	}
```
codesome: Benchmarking the error cases seems unnecessary

slrtbtfs: There are several reasons for benchmarking error cases:
* In case there is an error the control flow looks different (There is a panic - recover somewhere). So one cannot easily guess the performance in error cases from the regular case.
* The planned parser rewrite will change error handling significantly. It is useful to see how that affects performance.
* For the language server use case, fast parsing is desired, regardless of whether there is an error or not, especially for quickly showing diagnostics and error messages while typing.

brian-brazil:
Thanks!

PR #6356, opened 2019-11-21
Improve PromQL parser performance by making it non-concurrent

slrtbtfs:
Before this commit, the PromQL parser ran in two goroutines:
* The lexer goroutine that splits the input into tokens and sent them over a channel to
* the parser goroutine which produces the abstract syntax tree

The Problem with this approach is that the parser spends more time on goroutine creation
and synchronisation than on actual parsing.

This commit removes that concurrency and replaces the channel by a slice based buffer.

Benchmarks show that this makes the parser up to 7 times faster than before.

Benchmark results (based on the #6355): https://gist.github.com/slrtbtfs/24a4d92de43ad4ae871714840dab2d76

### DIFF
--- a/promql/lex.go
+++ b/promql/lex.go
@@ -317,13 +317,13 @@ type Pos int
 
 // lexer holds the state of the scanner.
 type lexer struct {
-	input   string    // The string being scanned.
-	state   stateFn   // The next lexing function to enter.
-	pos     Pos       // Current position in the input.
-	start   Pos       // Start position of this item.
-	width   Pos       // Width of last rune read from input.
-	lastPos Pos       // Position of most recent item returned by nextItem.
-	items   chan item // Channel of scanned items.
+	input   string  // The string being scanned.
+	state   stateFn // The next lexing function to enter.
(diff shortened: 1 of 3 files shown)


brian-brazil:
Thanks!

I didn't even know we were doing this, it doesn't make sense to me at all.

juliusv:
Yeah, I think it was based on Rob Pike's goroutine+channel-based lexer+parser design (https://www.youtube.com/watch?v=HxaD_trXwRE), which didn't really care much about performance. Not that it matters too much here either way given how little time is spent in this code path in general.

slrtbtfs:
Thanks for the quick merge and the pointer to the talk.

> Not that it matters too much here either way given how little time is spent in this code path in general.
For Prometheus itself it doesn't. When just using the normal PromQL engine benchmarks this gets indeed lost in the noise.

However, for the language server, when working with large YAML files with thousands of PromQL expressions, the parser performance starts to matter.

juliusv:
That's a good point. And generally totally for getting rid of this weird artefact :)

PR #6360, opened 2019-11-22
promql: Clean up parser struct

slrtbtfs:
The parser struct used two have two somewhat misused fields:

peekCount int
token     [3]item

By reading the code carefully one notices, that peekCount always has the value 0 or 1 and that only the first element of token is ever accessed.

To make this clearer, this commit replaces the token array with a single variable and the peekCount int with a boolean.

### DIFF
--- a/promql/parse.go
+++ b/promql/parse.go
@@ -32,9 +32,9 @@ import (
 )
 
 type parser struct {
-	lex       *lexer
-	token     [3]item
-	peekCount int
+	lex     *lexer
+	token   item
+	peeking bool
 }
 
 // ParseErr wraps a parsing error with line and position context.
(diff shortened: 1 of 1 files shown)


promql/parse.go
```diff
@@ -245,41 +245,41 @@ func (p *parser) typecheck(node Node) (err error) {
 
 // next returns the next token.
 func (p *parser) next() item {
-	if p.peekCount > 0 {
-		p.peekCount--
+	if p.peeking {
+		p.peeking = false
```
brian-brazil: You could unconditionally set this to false

slrtbtfs: True, thanks!

brian-brazil:
Thanks!

PR #6362, opened 2019-11-22
PromQL: Reduce Code duplication in AST traversion

slrtbtfs:
This PR adds a new Children() function that returns the Children of an AST Node. 

By changing existing AST traversing code to use this new function, code duplication is reduced.

This new function is also used by the language server.

This PR is the better version of #6205, which will be closed in favour of this one.


brian-brazil:
Thanks!

PR #6363, opened 2019-11-22
WIP: Add checkpointing within remote write; 

cstyan:
For #6333, this PR adds a file per queue with the highest timestamp from a sample that has been sent by that queue. This allows the queue to restart from that point in the event of a restart, potentially decreasing the number of samples lost in the case of remote write being far behind or not able to flush when it shutdown last.

Feel free to ignore `queue_manager_test.go` for now. Once https://github.com/prometheus/prometheus/pull/6043 is merged the naming of the queue and therefore the naming of it's checkpoint file will be much simpler.

@csmarchbanks 

### DIFF
--- a/storage/remote/checkpoint.go
+++ b/storage/remote/checkpoint.go
@@ -1,0 +1,81 @@
+// Copyright 2019 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package remote
+
+import (
+	"io/ioutil"
+	"os"
+	"strconv"
+)
+
+// A checkpoint file for remote write to use on restart/config reload
+// in attempt to minimize the chance of missing data in the remote system.
+type rwCheckpoint struct {
+	f *os.File
+}
+
+// NewRemoteWriteCheckpoint creates a new file for the
+// queue's remote write checkpoint if it doesn't already
+// have one, otherwise it returns the timestamp that was
+// contained in the existing file.
+func NewRemoteWriteCheckpoint(name string) (rwCheckpoint, int64, error) {
+	var c rwCheckpoint
+
+	_, err := os.Stat(name)
+	if err == nil {
+		f, err := os.OpenFile(name, os.O_WRONLY, 0666)
+		if err != nil {
+			return rwCheckpoint{}, 0, err
+		}
+		c.f = f
+		ts, err := c.readCheckpoint()
+		return c, ts, err
+	}
+	f, err := os.OpenFile(name, os.O_CREATE|os.O_RDWR, 0666)
+	if err != nil {
+		return rwCheckpoint{}, 0, err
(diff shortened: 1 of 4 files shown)


brian-brazil:
How did this interact with there being no ordering of samples by timestamp
within the WAL?

cstyan:
@brian-brazil If this is merged the feature should be labelled as best effort. 

If we sent a sample with t1 from one series but there was still unsent samples from another series at t0, reading the checkpoint would mean we only send samples after t1. It seemed overkill to track the timestamp per series.

brian-brazil:
One of the points of the new design is that we don't lose data on things
like restarts, so I'm not sure that's something we'd want to be doing.

cstyan:
Right now we potentially lose more data on restarts than we would with this
PR.

brian-brazil:
Why not checkpoint the segment number when we complete reading a segment?

cstyan:
Because we don't want to retransmit samples we've already sent. Segments are 128MB. 

If we're 90% of the way through a segment B with some more buffered but the remote endpoint goes down, if we just use the segment number A as a checkpoint we would have to start from the beginning of B on a restart. Currently we wouldn't do that, nor we would attempt to transmit from part of the way through B for what was missing on shutdown. We simply take the `time.Now()` of when remote write started up and only send samples with timestamps higher than that.

brian-brazil:
I'd think it'd be better to retransmit than to drop samples.

csmarchbanks:
During regular operation we try to do at least once delivery, and I think the goal should be to keep that behavior across restarts as well. If a segment number is too big, you could consider checkpointing the segment and offset.

cstyan:
> I'd think it'd be better to retransmit than to drop samples.

Before we implement something that could result in (potentially) an entire segment being re-transmitted I'd want to confirm whether we documented an intention to not do something of that nature anywhere in the past, and whether current remote write implementations are handling duplicate samples properly.



> If a segment number is too big, you could consider checkpointing the segment and offset.

For this approach I'd need to go back and do some refactoring of the way remote write tails the WAL and in addition I don't know if there's any guarantees for segments/offsets in a segment to be consistent across restarts. Repair and truncation could occur when Prometheus restarts.

I went with this implementation for now because I believe it gets us most of the benefit (not losing any samples) without adding a lot more code to maintain within remote write. If there's consensus that doing this checkpointing in a _best effort_ isn't enough I'll close this and start on a design doc for other options.

csmarchbanks:
I think if we are to implement checkpointing we should try to do it correctly. This implementation is better than what we have now, but will need to be updated in the future. Plus, we would probably still want to read this checkpoint too meaning multiple versions of checkpoint code.

I believe that once a page in a segment is written it is immutable, so the offset should be stable across restarts. If you look at the stackdriver sidecar they use an offset for their checkpoint.

One other thing to consider in any approach is the sharding calculation, it will need to be completely redone as the expectation that `samplesIn >= samplesOut` will no longer be valid.

cstyan:
> I think if we are to implement checkpointing we should try to do it correctly. This implementation is better than what we have now, but will need to be updated in the future. Plus, we would probably still want to read this checkpoint too meaning multiple versions of checkpoint code.
> 

Fair, I'll start a doc and we can discuss there/the mailing list.

> I believe that once a page in a segment is written it is immutable, so the offset should be stable across restarts. If you look at the stackdriver sidecar they use an offset for their checkpoint.
> 

My only concern is ensuring this is also true when the WAL is truncated/repaired when Prometheus is restarted. I think it is, since repair deletes the data after the corruption, and we always start writing to a new clean when we start.
> One other thing to consider in any approach is the sharding calculation, it will need to be completely redone as the expectation that `samplesIn >= samplesOut` will no longer be valid.

Good point.

PR #6365, opened 2019-11-23
Add time units to storage.tsdb.retention.size flag

bswinnerton:
In an effort to reduce confusion with the `m` option of the [`ParseDuration()`](https://github.com/prometheus/common/blob/17f5ca1748182ddf24fc33a5a7caaaf790a52fcc/model/time.go#L186-L216) function, this commit adds the available time units to the `storage.tsdb.retention.time` flag to help showcase that there is no option for months (which could be assumed to be `m`).

If someone were looking to set the retention to six months, they may mistakenly do so with `6m`, which would reduce their retention to six minutes.

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -195,7 +195,7 @@ func main() {
 	a.Flag("storage.tsdb.retention", "[DEPRECATED] How long to retain samples in storage. This flag has been deprecated, use \"storage.tsdb.retention.time\" instead.").
 		SetValue(&oldFlagRetentionDuration)
 
-	a.Flag("storage.tsdb.retention.time", "How long to retain samples in storage. When this flag is set it overrides \"storage.tsdb.retention\". If neither this flag nor \"storage.tsdb.retention\" nor \"storage.tsdb.retention.size\" is set, the retention time defaults to "+defaultRetentionString+". Units Supported: y,w,d,h,m,s,ms.").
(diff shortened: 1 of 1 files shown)


cmd/prometheus/main.go
```diff
@@ -195,7 +195,7 @@ func main() {
 	a.Flag("storage.tsdb.retention", "[DEPRECATED] How long to retain samples in storage. This flag has been deprecated, use \"storage.tsdb.retention.time\" instead.").
 		SetValue(&oldFlagRetentionDuration)
 
-	a.Flag("storage.tsdb.retention.time", "How long to retain samples in storage. When this flag is set it overrides \"storage.tsdb.retention\". If neither this flag nor \"storage.tsdb.retention\" nor \"storage.tsdb.retention.size\" is set, the retention time defaults to "+defaultRetentionString+".").
+	a.Flag("storage.tsdb.retention.time", "How long to retain samples in storage. When this flag is set it overrides \"storage.tsdb.retention\". If neither this flag nor \"storage.tsdb.retention\" nor \"storage.tsdb.retention.size\" is set, the retention time defaults to "+defaultRetentionString+". Units Supported: y,w,d,h,m,s,ms.").
```
simonpasquier: `Units Supported: y,w,d,h,m,s,ms` -> `Units supported: y, w, d, h, m, s, ms`

brian-brazil:
Hmm, github never got my reply email.

I don't suppose there's a way to do this for all the time-based flags centrally?

cmd/prometheus/main.go
bswinnerton: 👍 Addressed in 27463de.

bswinnerton:
> I don't suppose there's a way to do this for all the time-based flags centrally?

@brian-brazil not that I'm aware of, but happy to submit a follow up commit if you have any ideas.

brian-brazil:
Thanks!

This is the important one anyway, it was worth asking.

PR #6367, opened 2019-11-24
React UI: Improve graph legend hover performance

boyskila:

### DIFF
--- a/web/ui/react-app/src/Graph.tsx
+++ b/web/ui/react-app/src/Graph.tsx
@@ -37,6 +37,7 @@ interface GraphState {
 class Graph extends PureComponent<GraphProps, GraphState> {
   private chartRef = React.createRef<HTMLDivElement>();
   private $chart = {} as jquery.flot.plot;
+  private rafID = 0;
 
   state = {
     selectedSeriesIndex: null,
@@ -247,18 +248,21 @@ class Graph extends PureComponent<GraphProps, GraphState> {
   }
 
   plotSetAndDraw = (data: GraphSeries[] = this.state.chartData) => {
-    this.$chart.setData(data);
-    this.$chart.draw();
+    if (this.rafID) {
+      cancelAnimationFrame(this.rafID);
+    }
+    this.rafID = requestAnimationFrame(() => {
+      this.$chart.setData(data);
+      this.$chart.draw();
+    });
   };
 
   handleSeriesSelect = (index: number) => () => {
     this.setState(({ selectedSeriesIndex, chartData }) => {
-      if (selectedSeriesIndex === index) {
-        this.plotSetAndDraw(chartData.map(this.toHoverColor(index)));
-        return { selectedSeriesIndex: null };
-      }
-      this.plotSetAndDraw(chartData.slice(index, index + 1));
-      return { selectedSeriesIndex: index };
+      this.plotSetAndDraw(
+        selectedSeriesIndex === index ? chartData.map(this.toHoverColor(index)) : chartData.slice(index, index + 1)
+      );
+      return { selectedSeriesIndex: selectedSeriesIndex === index ? null : index };
     });
   };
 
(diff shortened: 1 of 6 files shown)


web/ui/react-app/src/Graph.tsx
```diff
@@ -238,64 +236,78 @@ class Graph extends PureComponent<GraphProps, GraphState> {
...
+      this.$chart.setData(data);
+      this.$chart.draw();
+    });
   };
 
-  handleSeriesHover = (index: number) => () => {
-    this.setState({ hoveredSeriesIndex: index });
+  handleSeriesSelect = (index: number) => () => {
+    this.setState(({ selectedSeriesIndex, chartData }) => {
+      this.plotSetAndDraw(
```
juliusv: Is it "normal" to put side effects like this into the `setState()` updater function? Just for my understanding, you probably had a good reason to call `this.plotSetAndDraw(...)` directly in `setState()` rather than outside of it?

boyskila: Yes, so this is same as
```
const { selectedSeriesIndex, chartData } = this.state;
this.plotSetAndDraw(
      selectedSeriesIndex === index
          ? chartData.map(this.toHoverColor(index))
           : chartData.slice(index, index + 1)
);
this.setState(
  { selectedSeriesIndex: selectedSeriesIndex === index ? null : index }
)
```
I just decided to use the callback version of setState which receives the prev state values and since ```plotSetAndDraw``` is not depending (in this case) on the next state is absolutely fine to be triggered from there. Thus looks more like a single unit.
Do you have any concerns about this?

web/ui/react-app/src/Graph.tsx
```diff
@@ -238,64 +236,78 @@ class Graph extends PureComponent<GraphProps, GraphState> {
...
     this.destroyPlot();
-    $.plot($(this.chartRef.current), selectedData ? [selectedData] : this.getData(), this.getOptions());
+    this.$chart = $.plot($(this.chartRef.current), this.state.chartData, this.getOptions());
   };
 
   destroyPlot() {
-    const chart = $(this.chartRef.current!).data('plot');
-    if (chart !== undefined) {
-      chart.destroy();
+    const $chart = $(this.chartRef.current!).data('plot');
```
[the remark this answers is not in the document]
boyskila: yes, I was looking for ```destroy``` method in plot types but didn't find it and I thought is a jquery one but now console log it and I see is a part of the plot object. I guess some of the plugins added it. I'll add it to our types in order to be accessible from this.$chart

web/ui/react-app/src/Graph.tsx
juliusv: When I approach this kind of question I always think mostly about whoever reads this code and what they would expect, and what would surprise them.

In this case, it's surprising for a reader to have extra side effects in a `setState()` call, which is (to my knowledge) normally only used to do direct state transformations on `this.state`, but nothing else besides that. So I'd prefer separating the two concerns and doing the plotting outside of `setState()`. Never be afraid of a tiny bit of extra code if it's less surprising :)

web/ui/react-app/src/Graph.tsx
```diff
@@ -238,64 +236,80 @@ class Graph extends PureComponent<GraphProps, GraphState> {
...
+    this.$chart = $.plot($(this.chartRef.current), this.state.chartData, this.getOptions());
   };
 
-  destroyPlot() {
-    const chart = $(this.chartRef.current!).data('plot');
-    if (chart !== undefined) {
-      chart.destroy();
-    }
+  destroyPlot = () => {
+    this.$chart.destroy && this.$chart.destroy();
```
juliusv: Shouldn't this be:

```suggestion
    this.$chart && this.$chart.destroy();
```

...as if there is a chart, it will always have a `destroy()` method, but there may not always be a chart?

boyskila: The approach is totally wrong. Initially, $chart should be undefined and not {}, thus checks here and there should be added but it will be more clear. I'll fix it.

web/ui/react-app/src/Graph.tsx
boyskila: Yes, I agree. I was thinking from another point of view but your suggestion makes more sense.

web/ui/react-app/src/Graph.tsx
```diff
@@ -242,21 +243,24 @@ class Graph extends PureComponent<GraphProps, GraphState> {
   };
 
   destroyPlot = () => {
-    this.$chart.destroy && this.$chart.destroy();
+    if (isPresent(this.$chart)) {
+      this.$chart.destroy && this.$chart.destroy();
```
juliusv: Do you still need the `this.$chart.destroy` check? Shouldn't `destroy` always exist on an existent chart?

boyskila: Nope....still sleeping :)

juliusv:
:+1: Really nice and fast now, thanks :) :heart_eyes:

Issue #6256, opened 2019-10-31
Proposal: Replacing the PromQL Parser with a generated parser


slrtbtfs:
# Replacing the PromQL Parser with a Generated Parser

### Current State of the Parser

Currently, the PromQL query engine utilizes a handwritten, recursive-descent parser, that was written five years ago and hasn't seen large changes since then.
It seems that none of the active members of the Prometheus project has a deep understanding of the parser code and is willing to maintain it or review complex PRs on it (based on the discussion in <https://github.com/prometheus/prometheus/pull/6061>).

There also seem to be some hidden features and edge cases that nobody is aware of (see <https://github.com/prometheus/prometheus/pull/6198> and <https://github.com/prometheus/prometheus/issues/6065>).

The code seems to be far more complex than necessary and would require significant refactoring before continuing to develop new features on it.

### Using the Parser for the Upcoming Language Server 

The upcoming PromQL language Server (see [proposal](https://github.com/slrtbtfs/promql-lsp/blob/master/doc/proposals/2019_promql-language-server.md)) requires a parser in order to analyze PromQL queries.

For purposes of compatibility, maintainability, and code reuse, it is intended for the language server to use the same PromQL parse as Prometheus.
To enable that use case, some enhancements to the existing parser have been [proposed](https://github.com/slrtbtfs/promql-lsp/blob/master/doc/proposals/2019_promql-parser-improvements.md).
These mainly concern the _output_ of the parser and have already been partly implemented.
However, some of the features described there, such as error recovery, would be difficult to implement in the existing parser.

#### Producing a Deeper Syntax Tree

Consider the following PromQL query:

    metric{label=="value", foo~="ba."}[5m] offset 3h

The corresponding syntax tree produced by the current parser looks something like the following:

    |---- MatrixSelector: metric{label=="value", foo~="ba."}[5m] offset 3h

This syntax tree is sufficient if the only goal is evaluating PromQL queries.
In fact, the PromQL query engine expects this simplified format.

However, it is insufficient for more involved use cases, such as: auto-completion that works almost anywhere in the code, generating helpful error messages, showing function signatures and documentation while editing queries, etc.
Instead, a syntax tree structured as follows would be more appropriate:

    |---- OffsetModifier: metric{label=="value", foo~="ba."}[5m] offset 3h
    . . . |---- MatrixSelector: metric{label=="value", foo~="ba."}[5m]
    . . . . . . |---- VectorSelector: metric{label=="value", foo~="ba."}
[...]

PR #6370, opened 2019-11-25
promql: Implement yyLexer interface

slrtbtfs:
This is the first step towards a generated lexer as described in #6256.

It adds methods to the parser struct, that make it implement the yyLexer interface required by a yacc generated parser, as descibed here: https://godoc.org/golang.org/x/tools/cmd/goyacc

The yyLexer interface is implemented by the parser struct instead of the lexer struct for the following reasons:

* Both parsers have a lookahead that the lexer does not know about. This solution makes it possible to synchronize these lookaheads when switching parsers.
* The routines to handle parser errors are not accessible to the lexer.

### DIFF
--- a/promql/parse.go
+++ b/promql/parse.go
@@ -337,6 +337,33 @@ func (p *parser) recover(errp *error) {
 	p.lex.close()
 }
 
+// yySymType is the Type the yacc generated parser expects for lexer items.
+//
+// For more information, see https://godoc.org/golang.org/x/tools/cmd/goyacc.
+type yySymType item
+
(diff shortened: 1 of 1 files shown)


promql/parse.go
```diff
@@ -337,6 +337,33 @@ func (p *parser) recover(errp *error) {
 	p.lex.close()
 }
 
+// yySymType is the Type the yacc generated parser expects for lexer items
```
brian-brazil: You're missing full stops throughout.

slrtbtfs: Wow, indeed. Fixed that now.

brian-brazil:
Thanks!

PR #6371, opened 2019-11-25
promql: Use capitalized names for item types

slrtbtfs:
For yacc generated parsers there is the convention to capitalize the names of item types provided by the lexer, which makes it easy to distinct lexer tokens (capitalized) from nonterminal symbols (not capitalized) in language grammars.

This convention is also followed by the (non generated) go compiler (see https://golang.org/pkg/go/token/#Token).

Part of the parser rewrite described in #6256.

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -1070,7 +1070,7 @@ func (ev *evaluator) eval(expr Expr) Value {
 
 	case *UnaryExpr:
 		mat := ev.eval(e.Expr).(Matrix)
-		if e.Op == ItemSUB {
+		if e.Op == SUB {
 			for i := range mat {
 				mat[i].Metric = dropMetricName(mat[i].Metric)
 				for j := range mat[i].Points {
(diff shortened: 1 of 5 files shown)


brian-brazil:
Is there anywhere that a 3rd party depending on this might want the name of a terminal, which would now be private if following this convention?

@juliusv

slrtbtfs:
> Is there anywhere that a 3rd party depending on this might want the name of a non-terminal?

Yes, the language server.

But this change won't make anything private. If this PR is merged and the parser is rewritten there would be:

* `LEXER_TOKENS / TERMINAL_SYMBOLS` that are produced by the lexer and are consumed by the parser (exported).
* `AstNodes` which are produced by the parser and are consumed by the engine and 3rd parties (exported).
* `grammar_rules / nonterminal_symbols` which are written in the yacc grammar and consumed by the parser generator. They won't be written in the generated go code, so talking about private/exported does not make sense here.
* `astNodeConstructors`, i.e. functions that are called by the generated parser to create the 
AST (not exported).

brian-brazil:
Thanks!

juliusv:
@brian-brazil As far as I can see, no exports are changed here, so I don't think it would be a problem for anyone?

brian-brazil:
Yeah, I wasn't sure initially and thought it might affect your Flux work.

slrtbtfs:
The names of the exported item types are changed by this PR. So if anybody is using them,  this change could indeed break things. 

But since @juliusv only exported them recently https://github.com/prometheus/prometheus/commit/8155cc49924cd48ec552506b4e5519ebafa8b722#diff-8d5817a793b0a130f76aa345d2d56614 , I doubt they are used anywhere except possibly flux.

juliusv:
Ah yeah, they are definitely used in Flux, but InfluxData has their own fork of the whole package now due to Go modules. Of course they should update at some point, but as long as they only have to change some names and the names don't get unexported completely, that's no problem at all.

PR #6372, opened 2019-11-25
DO NOT MERGE: fix memory overallocs in parts of PromQL engine

ppanyukov:
# DRAFT, DO NOT MERGE

### Summary

Fixes memory overallocations in parts of PromQL engine, specifically for `case *VectorSelector:` in `evaluator.eval(expr Expr)` in `promql/engine.go`.

**This reduced memory in use from `6.03GB` to `228.18MB`, which is broadly 26x/98% improvement** for my sample data and query, if we trust pprof and Go's memory stats :) 

NOTE: this somewhat fixes the https://github.com/prometheus/prometheus/pull/4513 which probably incorrectly accounts for the samples *used* in the query rather than *allocated* samples. If so, perhaps the fix should be within that space.

NOTE2: the savings depend on the query range, downsampling interval (step), and number of "holes" in data. I think.

NOTE3: the current PR includes instrumentation code, please ignore.

### Root cause and patch

The gist of the issue is best shown in code based on https://github.com/prometheus/prometheus/blob/master/promql/engine.go#L1129

```golang
case *VectorSelector:
    checkForSeriesSetExpansion(ev.ctx, e)
    mat := make(Matrix, 0, len(e.series))
    it := storage.NewBuffer(durationMilliseconds(LookbackDelta))
    for i, s := range e.series {
        it.Reset(s.Iterator())
        ss := Series{
            Metric: e.series[i].Labels(),

            // ==> This pre-allocates ss.Points with capacity equal to `numSteps`.
            Points: getPointSlice(numSteps),
        }

            _, v, ok := ev.vectorSelectorSingle(it, e, ts)
            if ok {
                if ev.currentSamples < ev.maxSamples {
                    // ==> This is where we actually add stuff to ss.Points
                    ss.Points = append(ss.Points, Point{V: v, T: ts})
                    ev.currentSamples++
                } else {
                    ev.error(ErrTooManySamples(env))
                }
            }
        }

        // ==> At this point the cap(ss.Points) may be very much larger than len(ss.Points).
        // ==> Since we keep the fully allocated array for the duration of the query eval,
        // ==> and since the number of series can be quite large, this can lead to excessive
        // ==> memory usage over what we need.
        if len(ss.Points) > 0 {
            mat = append(mat, ss)
        }

        // ==> The patch is to shrink ss.Points by reallocating to a smaller array.
        // ==> Here I do so if we overallocated more than 20% (arbitrary number).
        {
            pointsOverallocLimit := 1.2
            pointsOverallocRatio := float64(cap(ss.Points)) / float64(len(ss.Points))
            if pointsOverallocRatio > pointsOverallocLimit {
                pointsCopy := make([]Point, len(ss.Points))
                copy(ss.Points, pointsCopy)
                ss.Points = pointsCopy
            }
        }

    }
    return mat
```

### Impact of overallocation and the patch

Running 1-month ranged query over synthetic 1-week data:

Data:
* Thanos-bench generator profile: `realistic-k8s-1w-small`
* mint: 2019-10-11T03:00:15+01:00
* maxt: 2019-10-18T03:00:15+01:00

Query:
* query: `count({__name__=~".+"}) by (__name__)`
* mint: 2019-10-01T00:00:0+01:00
* maxt: 2019-11-01T00:00:0+01:00
* resolution: step=4838

```
// BEFORE PATCH:

PromQL: PROMQL_ALLOC_PATCH_OFF: true                                <=| Patch disabled
PromQL: Expression: {__name__=~".+"}
PromQL: VectorSelector: Series Count: 845000
PromQL: VectorSelector: pointsAllocSize: 468975000; Size: 7503.60M  <=| This is what we allocated
PromQL: VectorSelector: pointsNeededSize: 670000; Size: 10.72M      <=| This is what's actually needed
PromQL: VectorSelector: pointsOverAllocRatio: 699.96x               <=| This is how much we over-allocated
PromQL: VectorSelector: Current Samples: 670000

// memory impact, including garbage
MEM STATS DIFF:   	promql-eval 	promql-eval - AFTER 	-> Delta
    HeapAlloc  : 	804.21M 	7.26G 			-> 6.46G
    HeapSys    : 	1.01G 		8.12G 			-> 7.11G
    HeapObjects: 	12.60M 		14.53M 			-> 1.94M
```

```
// AFTER PATCH:
// memory impact, including garbage
MEM STATS DIFF:   	promql-eval 	promql-eval - AFTER 	-> Delta
    HeapAlloc  : 	985.59M 	925.55M 		-> -60.05M
    HeapSys    : 	1.00G 		2.15G 			-> 1.14G
    HeapObjects: 	14.69M 		14.53M 			-> -151.87K
```

Pprof figures:

```
// BEFORE PATCH:
Showing nodes accounting for 6.03GB, 99.71% of 6.05GB total

// AFTER PATCH:
Showing nodes accounting for 228.18MB, 100% of 228.18MB total

// DELATA: 
-5.8GB --> -96%
```

### Pprof charts

**Original bytes in use:**

orig

**Patch bytes in use:**

patch

Diff bytes in use:

diff

### Pprof files

[eval-end-orig.pb.gz](https://github.com/prometheus/prometheus/files/3888115/eval-end-orig.pb.gz)
[eval-end-patch.pb.gz](https://github.com/prometheus/prometheus/files/3888116/eval-end-patch.pb.gz)
[eval-start-orig.pb.gz](https://github.com/prometheus/prometheus/files/3888117/eval-start-orig.pb.gz)
[eval-start-patch.pb.gz](https://github.com/prometheus/prometheus/files/3888118/eval-start-patch.pb.gz)


easyCZ:
Neat!

brian-brazil:
What does it look like with a more typical query? This change will increase allocs, and cpu usage. The existing PromQL benchmarks would also be useful.

ppanyukov:
> What does it look like with a more typical query? 

Overallocs look broadly similar. I think the interesting thing here is that surprising amounts of memory can be over-allocated depending on the data and query controlled by the user: range, step. I managed to contrust a query over the same data which wanted to allocated **65GB (!)** when the actual need was like 256M.

Here are the stats for `k8s_app_metric0` with everything else being exactly same including data and other query parameters:

Query:

* query: `k8s_app_metric0`
* mint: 2019-10-01T00:00:0+01:00
* maxt: 2019-11-01T00:00:0+01:00
* resolution: step=4838

```
// BEFORE PATCH
PromQL: PROMQL_ALLOC_PATCH_OFF: true
PromQL: Expression: k8s_app_metric0
PromQL: VectorSelector: Series Count: 16900
PromQL: VectorSelector: pointsAllocSize: 9379500; Size: 150.07M   <=| This is what we allocated
PromQL: VectorSelector: pointsNeededSize: 13400; Size: 0.21M      <=| This is what's actually needed
PromQL: VectorSelector: pointsOverAllocRatio: 699.96x             <=| This is how much we over-allocated
PromQL: VectorSelector: Current Samples: 13400

// memory impact, including garbage
MEM STATS DIFF:   	promql-eval 	promql-eval - AFTER 	-> Delta
    HeapAlloc  : 	25.70M 		151.65M 		-> 125.96M
    HeapSys    : 	65.90M 		200.11M 		-> 134.22M
    HeapObjects: 	323.45K 	294.41K 		-> -29.05K

```

```
// AFTER PATCH:
// memory impact, including garbage
MEM STATS DIFF:   	promql-eval 	promql-eval - AFTER 	-> Delta
    HeapAlloc  : 	25.44M 		25.00M 			-> -440.93K
    HeapSys    : 	65.90M 		65.90M 			-> 0
    HeapObjects: 	319.56K 	294.47K 		-> -25.09K
```

Pprof figures:

```
// BEFORE PATCH:
Showing nodes accounting for 127.61MB, 100% of 127.61MB total

AFTER PATCH:
Showing nodes accounting for 5620.68kB, 100% of 5620.68kB total
```

Note that overallo ratio is exactly same here at `699.96x`, even if absolute figures are different.

----

> This change will increase allocs, and cpu usage. 

Here is my current thinking on this:

- We allocate only 1 extra array per loop cycle, **but** that object can be several orders of magnitude smaller to replace large ones. Overall the I think this should have a positive impact reducing GC pressure etc and that one extra array alloc per loop is a drop in a ocean in the grand scheme of things.

- Also we don't do anything if we are overallocated less than 20% (arbitrary figure at the moment). So for some scenarios this patch can have zero impact.

- CPU cycles. Yes true but should we worry about it? `memcpy` of contigious areas of memory is **fast**. To be fair, I didn't do any benchmarks so there is a bit of hand-waving there :)  If people insist, I can do some perf benchmarks around this.


----

> The existing PromQL benchmarks would also be useful.

More than happy to run these if people could point me in the right direction and guide a bit. I'm not at all familiar with this side of the project.

brian-brazil:
> More than happy to run these if people could point me in the right direction and guide a bit.

Run all benchmarks in the promql directory.

ppanyukov:
So... I removed all instrumentation and run the benchmarks.

Unless I did something completely dumb, I'm inclined to interpret results as "near-zero performance impact with marginal differenced bordered on error" but I may be mistaken.


* Original: https://github.com/prometheus/prometheus/commit/7293c859abce40ae8c526c0f0134bcb894f282c7
* Patch: this PR https://github.com/prometheus/prometheus/pull/6372/commits/f9e2ab03bba09b6ca6c1711b28d349b5ddf026d0

```
go test -run=NONE -bench=. ./promql/... | tee old.txt
go test -run=NONE -bench=. ./promql/... | tee new.txt
benchcmp old.txt new.txt  > diff.txt
```

Did I run the correct thing?

<details>
<summary>Highlights:</summary>

* ns/op
```
benchmark                                                                                                              old ns/op      new ns/op      delta
# deltas border or error
BenchmarkRangeQuery/expr=a_hundred,steps=1000-16                                                                       6434642        6480290        +0.71%
BenchmarkRangeQuery/expr=changes(a_hundred[1d]),steps=1000-16                                                          2587642563     2574245027     -0.52%

# some take longer
BenchmarkRangeQuery/expr=a_one,steps=100-16                                                                            25623          27219          +6.23%
BenchmarkRangeQuery/expr=a_one,steps=1000-16                                                                           78269          82208          +5.03%

# but some are equally faster
BenchmarkRangeQuery/expr=a_one_-_b_one,steps=100-16                                                                    112892         105559         -6.50%
BenchmarkRangeQuery/expr=a_one_-_b_one,steps=1000-16                                                                   785382         728117         -7.29%
BenchmarkRangeQuery/expr=a_ten_-_b_ten,steps=1000-16                                                                   7162044        6463838        -9.75%
BenchmarkRangeQuery/expr=a_ten_-_b_ten,steps=10000-16                                                                  77495123       63063953       -18.62%
```

* allocs/op

```
# *most* zero deltas
BenchmarkRangeQuery/expr=a_one,steps=1-16                                                                              107            107            +0.00%

# only three show small icrease
BenchmarkRangeQuery/expr=a_hundred_-_b_hundred,steps=1000-16                                                           17895          18575          +3.80%
BenchmarkRangeQuery/expr=a_ten_-_b_ten,steps=10000-16                                                                  53296          54529          +2.31%
BenchmarkRangeQuery/expr=a_hundred_-_b_hundred,steps=10000-16                                                          155247         160374         +3.30%
```

* bytes/op

```
# *most* are near-zero deltas
BenchmarkRangeQuery/expr=a_one,steps=1000-16                                                                           8328          8323          -0.06%
BenchmarkRangeQuery/expr=a_ten,steps=1-16                                                                              13307         13306         -0.01%
BenchmarkRangeQuery/expr=a_ten,steps=10-16                                                                             13306         13306         +0.00%
BenchmarkRangeQuery/expr=a_ten,steps=100-16                                                                            14026         14026         +0.00%
BenchmarkRangeQuery/expr=a_ten,steps=1000-16                                                                           21066         21075         +0.04%
BenchmarkRangeQuery/expr=a_hundred,steps=1-16                                                                          80003         80206         +0.25%

# two strange ones showing some increase
BenchmarkRangeQuery/expr=a_ten_-_b_ten,steps=10000-16                                                                  3201351       3550912       +10.92%
BenchmarkRangeQuery/expr=a_hundred_-_b_hundred,steps=1000-16                                                           2811254       3035373       +7.97%

# equally some showing decrease
BenchmarkRangeQuery/expr=changes(a_one[1d]),steps=1000-16                                                              578778        547794        -5.35%
BenchmarkRangeQuery/expr=a_ten_-_b_ten,steps=1000-16                                                                   359869        348921        -3.04%

```
</details>

Attached are the benchmark files. Of course it would be nice if someone could verify my results independently.

[old.txt](https://github.com/prometheus/prometheus/files/3893593/old.txt)

[new.txt](https://github.com/prometheus/prometheus/files/3893594/new.txt)

[diff.txt](https://github.com/prometheus/prometheus/files/3893595/diff.txt)

brian-brazil:
Your PR only helps with sparse series, and the existing benchmarks use data which is not sparse at all. We'll likely need some new benchmarks here.

ppanyukov:
OK, I made a Big Dumb Thing with `copy` by confusing source and destination, which I have now fixed. Thank you `go test` for picking this up :) 

The memory benchmarks still stand though! (basically I was replacing the samples with all zeros, doh!).

I have re-run the benchmarks with this fix though, and results are interesting, and also a bit confusing. Maybe because it's late in the evening or I did some Dumb Thing again.

Basically, benchmarks show that eval is now *faster*. A lot of tests show 7-20% faster. Why? is there a bug somewhere either in the PR or the way I ran benchmarks? Some background activities on my laptop skewed this? Is it some GC-related thing or GO's malloc thing? 

For now, see the updated benchmarks.

[old.txt](https://github.com/prometheus/prometheus/files/3893811/old.txt)

[new.txt](https://github.com/prometheus/prometheus/files/3893812/new.txt)

[diff.txt](https://github.com/prometheus/prometheus/files/3893816/diff.txt)

ppanyukov:
> Your PR only helps with sparse series, and the existing benchmarks use data which is not sparse at all. We'll likely need some new benchmarks here.

Yes I guess so. It would be extremely helpful if someone either created these sparse benchmarks or helped me do so because I really have no idea what I’m doing when it comes to PromQL :)

It would be nice if benchmarks were already in place somewhere so I could use them.

brian-brazil:
> A lot of tests show 7-20% faster. Why?

It's possible some of our existing benchmarks are not actually sparse, which should be fixed if that's the case.

ppanyukov:
It would also be nice if someone did a really careful review of this patch for correctness and as a bonus did both men and perf validation. I was a bit rocked by that “copy” fiasco and don’t trust myself at the moment!

brian-brazil:
One thing to watch on laptops as well is power saving stuff which can vary the frequencies of the cores, I'd suggest turning that off. Though that usually shouldn't affect microbenchmarks like these.

brian-brazil:
Also, if you broke something the unittests should catch it.

ppanyukov:
> Also, if you broke something the unittests should catch it.

Yes they did :) Tests are green now though. Quick approve and merge ;)

On a serious note, if all is green and if people are happy with this fix and benchmarks as they stand, will this be a LGTM and a merge? If not, what would I/we need to do to make this happen?

brian-brazil:
There'll be more benchmarks and tweaks required anyway, but first we should determine if it's actually a win.

ppanyukov:
Ok let’s start with the win.

“6GB -> 0.2GB” == win

Right?

brian-brazil:
That's only one case, it might make other cases slower or use more memory.

ppanyukov:
> That's only one case, it might make other cases slower or use more memory.

The perf benchmarks suggest it’s all good? Of course this assertion is only as good as the benchmarks but in the absence of any other measurement, how do we make a judgement if this PR is ok? What do we need to do to assure ourselves it is indeed the case?

ppanyukov:
I think what I’m asking for, to be a bit blunt perhaps, is this:

* what are the concrete steps I need to take for the merge to happen? 

More benchmarks? If so what kind and can anyone help?

Explaining and scrutinising the method and data I used? Happy to do that if that’s what blocks this PR.

Over alloc ratio or 1.2 is questionable? Which ratio is good then? 1.5? 1.8? 2.0? Let’s discuss.

Real life test ran by volunteers? That would be great?

Anything else? A list of concrete steps would help.

I think I provided all the pprof files and perf benchmarks (ran on a laptop of course) to show there is a benefit for some cases without perf degradation as far as the tests show.

So what do I need to do to move this forward? Concrete steps?

bwplotka:
Nice! 

> Real life test ran by volunteers? That would be great?

That would do as well, but probably some docker image would be nice to have for this, we can ask on prometheus-dev mailing list if needed as well.

> So what do I need to do to move this forward? Concrete steps?

I think we need to justify this with microbenchmarks, so we would probably need to add a case to benchmarks for sparse series, so series that are living, for example, 1/(30*24) of the requested time range like in your case, which is sometimes realistic. We definitely over-allocate in those cases as we allocate assuming all samples for each step for all series are present and I totally see some fix like this to make sense.

I cannot still understand though how shrinking helps here though as we already allocated this memory in `return make([]Point, 0, sz)` as I understand and once we put it properly we just retrieve this from the sync pool or allocate the too large slice again. (: Do we have idea why shrinking helps here?

codesome:
How about running prombench for few hours (6-10) on this?

brian-brazil:
> what are the concrete steps I need to take for the merge to happen?

We don't exactly have an ISO-9000 certified process, so we'll likely go back and forth a bit as we ensure the change is good and ways this may cause issues occur to us.

> Over alloc ratio or 1.2 is questionable? Which ratio is good then? 1.5? 1.8? 2.0? Let’s discuss.

I'd go for 2 by default barring data indicating otherwise, to match how Go slices generally work.

> How about running prombench for few hours (6-10) on this?

Microbenchmarks will be the most relevant here, and we'll want relevant ones added anyway to judge any future changes to this code.

> Do we have idea why shrinking helps here?

We're also not considering the interaction with sync pool, nor using it for the new slices, so that means more garbage, thus more memory pressure and more CPU to collect it.

ppanyukov:
> I cannot still understand though how shrinking helps here though as we already allocated this memory in return make([]Point, 0, sz) as I understand and once we put it properly we just retrieve this from the sync pool or allocate the too large slice again. (: Do we have idea why shrinking helps here?

A very good question @bwplotka!  Also thanks @brian-brazil, helpful comments (that ISO comment made my chuckle) and yes, sync pool looks to be the key.

I dug around trying to understand what's going on and I found some interesting things.

**TL;DR:** The `sync.Pool` makes things fun and intersting, I've no idea what I'm measuring anymore, but the patch still seems to help!


<details>

<summary>Points are not always put back into pool. It depends on the query. Is it a bug or by design?</summary>

The way I did this is here: https://github.com/ppanyukov/prometheus/compare/master...ppanyukov:feature/promql-pool-debug

```
expr               : k8s_app_metric0
poolGetCount       : 16900
poolGetSuccessCount: 0
poolPutCount       : 0          <= not putting back

expr               : k8s_app_metric0
poolGetCount       : 33800
poolGetSuccessCount: 0
poolPutCount       : 0          <= not putting back

expr               : k8s_app_metric0
poolGetCount       : 50700
poolGetSuccessCount: 0
poolPutCount       : 0          <= not putting back

expr               : count by(__name__) ({__name__=~".+"})
poolGetCount       : 895750
poolGetSuccessCount: 0
poolPutCount       : 670000     <= this does put back

expr               : k8s_app_metric0
poolGetCount       : 912650
poolGetSuccessCount: 16900
poolPutCount       : 670000

expr               : k8s_app_metric0
poolGetCount       : 929550
poolGetSuccessCount: 33800
poolPutCount       : 670000

expr               : k8s_app_metric0
poolGetCount       : 946450
poolGetSuccessCount: 50700
poolPutCount       : 670000
```
</details>


<details>
<summary>Calling `runtime.GC()` wipes `sync.Pool`</summary>

In instrumented version I was writing pprof heap dumps. This, for some reason, called `runtime.GC()`. Which, as it turns out, seems to wipe whatever we have in `sync.Pool`.

This certainly would have an impact on any measurements. I have removed the call to `runtime.GC()` and things became much more interesting and fun after that!

</details>


<details>
<summary>Use of `sync.Pool` seems to make pprof heap profiling near-useless?</summary>

It looks like whatever is put into `sync.Pool` does not show up as "bytes in use" in pprof heap profiles. Unless I am doing it wrong? I wouldn't be surprised TBH. If anyone can shed some light on this that'd be great.

</details>



<details>
<summary>Use of `sync.Pool` shows different figures depending on how many times you run queries</summary>

With `sync.Pool` the more we run things, the more we reuse byte slices. Correct? And since things in `sync.Pool` are not shown as "bytes in use" in heap profile (see above), we get different profiles depending how many times we run queries. Combine this with the fact that some queries do not put anything back into the pool. Now we get different figures depending which queries we run in which order and how many times.

Questions: what am I measuring here?? And how to do it right?

</details>


<details>
<summary>The `HeapSys` stat seems to be most useful metric with `sync.Pool`?</summary>

I found that the only reliable figure here is `HeapSys`.

What I also found is that it "stabilises" at some point when I run queries, that's the effect of `sync.Pool` I presume.

</details>


<details>
<summary>But! The patch still seems to help!</summary>

If we assume that `HeapSys` is reliable, I found this patch still helps with memory utilisation.

First, the `sync.Pool` is still working (I observe things being put and successfully taken from it).
Second, the `HeapSys` is way lower with the patch. On my machine it was `13.8GB` without patch vs `2.2GB` with the patch.

So it does seem to help. I think the way this works is:

- we eiher alloc new array or get it from pool;
- we shrink it if required;
- the shrunk array is put back in the pool (sometimes);

The question remains: are we doing too many allocs and put pressure on GC with this patch?

</details>

This is about it for now. I'm very confused at this point and don't know what/if we want to do next :) 

Probably would be nice to confirm why things are not being put back into the pool first? And fix that if it's a bug?

ppanyukov:
> I'd go for 2 by default barring data indicating otherwise, to match how Go slices generally work.

Yes, seems reasonable, I considered exactly same number too. How about maybe even go to 2.2? Since we are using sync.Pool we perhaps don't want to be overaggressive/precise in the trimming?

More general question: what kind of tests/benchmarks can we have to help us confirm 2.0 or 2.2 are good numbers?  If tha's possible at all.

brian-brazil:
> Points are not always put back into pool. It depends on the query. Is it a bug or by design?

Hmm, they're meant to. Are you closing the query properly when it's complete? I'd have to see the specific cases, but it's probably an oversight. Let's not worry about it right now if we don't have to.

> Calling `runtime.GC()` wipes `sync.Pool`

That depends on your Go version, it's only half wiped now I believe.

> And since things in sync.Pool are not shown as "bytes in use" in heap profile (see above)

They are in use on the heap, so should be showing.

> Use of `sync.Pool` seems to make pprof heap profiling near-useless?

It does make performance profiling fun alright.

> Questions: what am I measuring here?? And how to do it right?

That's the big question here. In an ideal world we wouldn't need to have the pool for performance, but that's not how Go works currently.

> More general question: what kind of tests/benchmarks can we have to help us confirm 2.0 or 2.2 are good numbers? If tha's possible at all.

To be honest I wouldn't worry about getting it perfect in this PR. Once we've figured out what benchmarks etc. we need, finding the optimal value can be done later.

brian-brazil:
It occurs to me that what we probably want benchmark wise here is to manually set the contents of the syncpool before each iteration of the tests for this. As against the existing tests (which we can re-use the test cases for) where we want the sync pool to mature naturally to simulate the best case with regular refreshes/evaluations.

ppanyukov:
Yes excellent idea. If we could come up with a good set of tests and benchmarks for this, it would be generally a good thing regardless of the fate of this PR :)

Brian do you think it’s worth to sync up offline on slack or something at some point for a more fluid discussion and to avoid noise here and then post the summary here?

brian-brazil:
We're on #prometheus-dev on Freenode if you want to chat more realtime, but I think my above idea will be sufficient benchmark wise.

ppanyukov:
Quick update on this. I'm a bit sidetracked with other things at the mo but this PR is not forgotten. We are going to run some tests of this patch on the real data sets using Thanos and we will see what it shows.

The biggest blocker for me right now is "what microbenchmarks exactly do I need to create?", as I'm not knowledgable enough of Prometheus and existing benchmarks etc so it will take some learning.  If someone could take a stab at these in a separate PR, that would be super awesome.

mjd95:
We did some tests with this patch in the Thanos querier, using real data.  We just hit the querier with various heavy queries, with a couple of minutes between each query.

We first tried `sum(up)` over an 8 week period.  We only have data for about 4 of those weeks in this environment so it's partial results, but I think this kind of query is somewhat normal anyway.

With the patch **off**:
```
level=debug name=<querier_name> ts=2019-12-06T11:16:07.105677359Z caller=proxy.go:376 queryTotalSize=420.09MB queryLocalSize=160.02MB
level=debug name=<querier_name> ts=2019-12-06T11:16:07.10942091Z caller=proxy.go:376 queryTotalSize=440.09MB queryLocalSize=220.06MB
level=debug name=<querier_name> ts=2019-12-06T11:16:07.109481432Z caller=proxy.go:191 queryTotalSize=440.09MB
...
PromQL: PROMQL_ALLOC_PATCH_ON: false
PromQL: Expression: up
PromQL: VectorSelector: Series Count:    286829
PromQL: VectorSelector: Current Samples: 20112
PromQL: VectorSelector: pointsAllocInitial:   71994079; Size: 1151.91M
PromQL: VectorSelector: pointsAllocNeeded:    20112; Size: 0.32M
PromQL: VectorSelector: pointsAllocFinal:     71994079; Size: 1151.91M
PromQL: VectorSelector: pointsOverallocLimit:        2.00
PromQL: VectorSelector: pointsOverAllocRatioInitial: 3579.66x
PromQL: VectorSelector: pointsOverAllocRatioFinal:   3579.66x
PromQL: VectorSelector: pointsOverAllocDelta:        0.00x
expr               : sum(up)
poolGetCount       : 1080832
poolGetSuccessCount: 15
poolPutCount       : 106400
MEM STATS DIFF:         promql-eval     promql-eval - AFTER     -> Delta
    HeapAlloc  :        885.60M         1.76G                   -> 874.95M
    HeapSys    :        1.81G           1.88G                   -> 66.72M
    HeapObjects:        9.51M           10.37M                  -> 855.72K
```

And with the patch **on**:
```
level=debug name=<querier_name> ts=2019-12-06T11:23:44.44175381Z caller=proxy.go:376 queryTotalSize=420.05MB queryLocalSize=160.02MB
level=debug name=<querier_name> ts=2019-12-06T11:23:44.441867962Z caller=proxy.go:376 queryTotalSize=440.05MB queryLocalSize=220.02MB
level=debug name=<querier_name> ts=2019-12-06T11:23:44.441949684Z caller=proxy.go:191 queryTotalSize=440.05MB
...
PromQL: PROMQL_ALLOC_PATCH_ON: true
PromQL: Expression: up
PromQL: VectorSelector: Series Count:    286829
PromQL: VectorSelector: Current Samples: 20112
PromQL: VectorSelector: pointsAllocInitial:   71994079; Size: 1151.91M
PromQL: VectorSelector: pointsAllocNeeded:    20112; Size: 0.32M
PromQL: VectorSelector: pointsAllocFinal:     20112; Size: 0.32M
PromQL: VectorSelector: pointsOverallocLimit:        2.00
PromQL: VectorSelector: pointsOverAllocRatioInitial: 3579.66x
PromQL: VectorSelector: pointsOverAllocRatioFinal:   1.00x
PromQL: VectorSelector: pointsOverAllocDelta:        -3578.66x
expr               : sum(up)
poolGetCount       : 286830
poolGetSuccessCount: 0
poolPutCount       : 15055
MEM STATS DIFF:         promql-eval     promql-eval - AFTER     -> Delta
    HeapAlloc  :        1.24G           1.40G                   -> 164.35M
    HeapSys    :        1.27G           1.68G                   -> 402.65M
    HeapObjects:        11.03M          9.55M                   -> -1.49M
```

--------

We also tried the pathological query `count({__name__=~".+"}) by (__name__)` over a 6 day period, again covering a partial data time
Patch off:
```
level=debug name=<querier_name> ts=2019-12-06T11:55:39.537519487Z caller=proxy.go:376 queryTotalSize=3900.84MB queryLocalSize=3900.84MB
level=debug name=<querier_name> ts=2019-12-06T11:55:39.53760166Z caller=proxy.go:191 queryTotalSize=3900.84MB
….
PromQL: PROMQL_ALLOC_PATCH_ON: false
PromQL: Expression: {__name__=~".+"}
PromQL: VectorSelector: Series Count:    2696665
PromQL: VectorSelector: Current Samples: 8744867
PromQL: VectorSelector: pointsAllocInitial:   676859462; Size: 10829.75M
PromQL: VectorSelector: pointsAllocNeeded:    8744867; Size: 139.92M
PromQL: VectorSelector: pointsAllocFinal:     676859462; Size: 10829.75M
PromQL: VectorSelector: pointsOverallocLimit:        2.00
PromQL: VectorSelector: pointsOverAllocRatioInitial: 77.40x
PromQL: VectorSelector: pointsOverAllocRatioFinal:   77.40x
PromQL: VectorSelector: pointsOverAllocDelta:        0.00x
expr               : count by(__name__) ({__name__=~".+"})
poolGetCount       : 2697090
poolGetSuccessCount: 269
poolPutCount       : 322
MEM STATS DIFF:         promql-eval     promql-eval - AFTER     -> Delta
    HeapAlloc  :        8.65G           18.64G                  -> 9.99G
    HeapSys    :        8.79G           19.19G                  -> 10.40G
    HeapObjects:        97.48M          110.30M                 -> 12.82M
```
Patch on:
```
level=debug name=<querier_name> ts=2019-12-06T11:44:34.426667882Z caller=proxy.go:376 queryTotalSize=3900.84MB queryLocalSize=3900.84MB
level=debug name=<querier_name> ts=2019-12-06T11:44:34.426772731Z caller=proxy.go:191 queryTotalSize=3900.84MB
…. 
PromQL: PROMQL_ALLOC_PATCH_ON: true
PromQL: Expression: {__name__=~".+"}
PromQL: VectorSelector: Series Count:    2696665
PromQL: VectorSelector: Current Samples: 8744867
PromQL: VectorSelector: pointsAllocInitial:   676862915; Size: 10829.81M
PromQL: VectorSelector: pointsAllocNeeded:    8744867; Size: 139.92M
PromQL: VectorSelector: pointsAllocFinal:     8760984; Size: 140.18M
PromQL: VectorSelector: pointsOverallocLimit:        2.00
PromQL: VectorSelector: pointsOverAllocRatioInitial: 77.40x
PromQL: VectorSelector: pointsOverAllocRatioFinal:   1.00x
PromQL: VectorSelector: pointsOverAllocDelta:        -76.40x
expr               : count by(__name__) ({__name__=~".+"})
poolGetCount       : 4955190
poolGetSuccessCount: 542
poolPutCount       : 917824
MEM STATS DIFF:         promql-eval     promql-eval - AFTER     -> Delta
    HeapAlloc  :        8.92G           17.35G                  -> 8.43G
    HeapSys    :        9.06G           18.05G                  -> 9.00G
    HeapObjects:        99.13M          111.68M                 -> 12.55M

```
Honestly, these look kind of mixed to me.  Of course there is the issue of when GCs happened as well.

It's notable that we are not actually extracting things from the pool that often.  I would be inclined to try getting rid of the pool altogether (as in https://github.com/prometheus/prometheus/pull/6427) to make this easier to reason about.  But in any case, the pool is only partially related to whether we should do the shrink.

ppanyukov:
Re escape analysis:

* Escape analysis is rewritten in 1.14: https://go-review.googlesource.com/c/go/+/170322/
* Go 1.14 Beta announced today: https://twitter.com/golang/status/1207027024893792256?s=20

Don't know if this would cause anything to escape less w/o running tests and deving into details of this change.

brian-brazil:
If it happens to make PromQL without the pooling no longer do heap allocations for samples, then it's of interest here.

ppanyukov:
Yes, I will try out 1.14 and see what happens. 

In the meantime, we've been running some tests on the real data on the real clusters to see what positive impact (if any) various patches may have. Doing so, we've come up with an idea which *should* make things better.

Idea 1:

> When we make a new correctly-sized slice and copy points there, we are dropping the over-allocated slice on the floor. This becomes garbage. Which is wrong. Why don't we reuse it?

Also, considering there is "let's not use sync.Pool" conversation floating around, here is another idea:

Idea 2:

> Switch off sync.Pool. But keep the resizing patch. **And** in addition to that, keep the overallocated slice for reuse in the next loop iteration.


I did some local test runs on OS X with synthetic data to see how that works. Summary:

* Using `HeapSys` is probably more reliable metric than anything else?
* Most numbers seem to show improvement
* Both "sync.Pool" and "no sync.Pool" versions showed comparable numbers 
* But some numbers are confusing, esp `HeapObjects`
* Also I don't trust local tests running on OS X, we've seen things work differently on Linux in k8s.
* Will need to run benchmarks and more realistic tests


```
# All numbers are after running same query 3 times

ORIGINAL (uses sync.Pool)
    HeapSys    : 	11.54G

PATCH1 (uses sync.Pool + resize overalloced slice)
    HeapSys    : 	2.28G

PATCH1+PATCH2 (uses sync.Pool + resize overalloced slice + put overalloced slice back to sync.Pool)
    HeapSys    : 	1.88G

PATCH3.1 (Original code, No Sync.Pool at all)
    HeapSys    : 	13.62G

PATCH3.2: No Sync.Pool at all, + resize + keep overalloced slice for next loop iteration
    HeapSys    : 	1.81G

```

<details>
<summary>Full logs</summary>

```
ORIGINAL (uses sync.Pool)
=========

real	1m19.310s
user	0m0.014s
sys	0m0.017s

PromQL: PROMQL_ALLOC_PATCH_ON : false
PromQL: PROMQL_ALLOC_PATCH2_ON: false
PromQL: Expression: {__name__=~".+"}
PromQL: VectorSelector: Series Count:    845000
PromQL: VectorSelector: Current Samples: 670000
PromQL: VectorSelector: pointsAllocInitial:   468975000; Size: 7503.60M
PromQL: VectorSelector: pointsAllocNeeded:    670000; Size: 10.72M
PromQL: VectorSelector: pointsAllocFinal:     468975000; Size: 7503.60M
PromQL: VectorSelector: pointsOverallocLimit:        2.00
PromQL: VectorSelector: pointsOverAllocRatioInitial: 699.96x
PromQL: VectorSelector: pointsOverAllocRatioFinal:   699.96x
PromQL: VectorSelector: pointsOverAllocDelta:        0.00x
expr               : count by(__name__) ({__name__=~".+"})
poolGetCount       : 2535150
poolGetSuccessCount: 1864997
poolPutCount       : 2535000

MEM STATS DIFF:   	promql-eval 	promql-eval - AFTER 	-> Delta
    HeapAlloc  : 	10.43G 		11.46G 			-> 1.03G
    HeapSys    : 	10.53G 		11.54G 			-> 1.01G
    HeapObjects: 	55.59M 		68.35M 			-> 12.76M


PATCH1 (uses sync.Pool + resize overalloced slice)
=======
real	1m11.209s
user	0m0.014s
sys	0m0.016s

PromQL: PROMQL_ALLOC_PATCH_ON : true
PromQL: PROMQL_ALLOC_PATCH2_ON: false
PromQL: Expression: {__name__=~".+"}
PromQL: VectorSelector: Series Count:    845000
PromQL: VectorSelector: Current Samples: 670000
PromQL: VectorSelector: pointsAllocInitial:   468975000; Size: 7503.60M
PromQL: VectorSelector: pointsAllocNeeded:    670000; Size: 10.72M
PromQL: VectorSelector: pointsAllocFinal:     670000; Size: 10.72M
PromQL: VectorSelector: pointsOverallocLimit:        2.00
PromQL: VectorSelector: pointsOverAllocRatioInitial: 699.96x
PromQL: VectorSelector: pointsOverAllocRatioFinal:   1.00x
PromQL: VectorSelector: pointsOverAllocDelta:        -698.96x
expr               : count by(__name__) ({__name__=~".+"})
poolGetCount       : 2535150
poolGetSuccessCount: 0
poolPutCount       : 2010000

MEM STATS DIFF:   	promql-eval 	promql-eval - AFTER 	-> Delta
    HeapAlloc  : 	974.87M 	1.12G 			-> 143.49M
    HeapSys    : 	2.28G 		2.28G 			-> 0
    HeapObjects: 	14.50M 		14.55M 			-> 48.07K



PATCH1+PATCH2 (uses sync.Pool + resize overalloced slice + put overalloced slice back to sync.Pool)
=============

real	1m6.615s
user	0m0.014s
sys	0m0.015s

PromQL: PROMQL_ALLOC_PATCH_ON : true
PromQL: PROMQL_ALLOC_PATCH2_ON: true
PromQL: Expression: {__name__=~".+"}
PromQL: VectorSelector: Series Count:    845000
PromQL: VectorSelector: Current Samples: 670000
PromQL: VectorSelector: pointsAllocInitial:   468975000; Size: 7503.60M
PromQL: VectorSelector: pointsAllocNeeded:    670000; Size: 10.72M
PromQL: VectorSelector: pointsAllocFinal:     670000; Size: 10.72M
PromQL: VectorSelector: pointsOverallocLimit:        2.00
PromQL: VectorSelector: pointsOverAllocRatioInitial: 699.96x
PromQL: VectorSelector: pointsOverAllocRatioFinal:   1.00x
PromQL: VectorSelector: pointsOverAllocDelta:        -698.96x
expr               : count by(__name__) ({__name__=~".+"})
poolGetCount       : 2535150
poolGetSuccessCount: 2534997
poolPutCount       : 4545000

MEM STATS DIFF:   	promql-eval 	promql-eval - AFTER 	-> Delta
    HeapAlloc  : 	802.34M 	1.55G 			-> 745.27M
    HeapSys    : 	1.81G 		1.88G 			-> 67.11M
    HeapObjects: 	12.55M 		22.66M 			-> 10.11M


PATCH3.1 (Original code, No Sync.Pool at all)
=============

real	1m23.233s
user	0m0.015s
sys	0m0.018s

PromQL: SYNC_POOL_IS_OFF: true
PromQL: PROMQL_ALLOC_PATCH_ON (sync.Pool) : false
PromQL: Expression: {__name__=~".+"}
PromQL: VectorSelector: Series Count:    845000
PromQL: VectorSelector: Current Samples: 670000
PromQL: VectorSelector: pointsAllocInitial:   468975000; Size: 7503.60M
PromQL: VectorSelector: pointsAllocNeeded:    670000; Size: 10.72M
PromQL: VectorSelector: pointsAllocFinal:     468975000; Size: 7503.60M
PromQL: VectorSelector: pointsOverallocLimit:        2.00
PromQL: VectorSelector: pointsOverAllocRatioInitial: 699.96x
PromQL: VectorSelector: pointsOverAllocRatioFinal:   699.96x
PromQL: VectorSelector: pointsOverAllocDelta:        0.00x
expr               : count by(__name__) ({__name__=~".+"})
poolGetCount       : 0
poolGetSuccessCount: 0
poolPutCount       : 0

MEM STATS DIFF:   	promql-eval 	promql-eval - AFTER 	-> Delta
    HeapAlloc  : 	9.39G 		8.06G 			-> -1.34G
    HeapSys    : 	11.81G 		13.62G 			-> 1.81G
    HeapObjects: 	38.23M 		16.30M 			-> -21.93M


PATCH3.2: No Sync.Pool at all, + resize + keep overalloced slice for next loop iteration
=============

real	1m8.178s
user	0m0.014s
sys	0m0.015s


PromQL: SYNC_POOL_IS_OFF: true
PromQL: PROMQL_ALLOC_PATCH_ON (sync.Pool) : true
PromQL: Expression: {__name__=~".+"}
PromQL: VectorSelector: Series Count:    845000
PromQL: VectorSelector: Current Samples: 670000
PromQL: VectorSelector: pointsAllocInitial:   468975000; Size: 7503.60M
PromQL: VectorSelector: pointsAllocNeeded:    670000; Size: 10.72M
PromQL: VectorSelector: pointsAllocFinal:     670000; Size: 10.72M
PromQL: VectorSelector: pointsOverallocLimit:        2.00
PromQL: VectorSelector: pointsOverAllocRatioInitial: 699.96x
PromQL: VectorSelector: pointsOverAllocRatioFinal:   1.00x
PromQL: VectorSelector: pointsOverAllocDelta:        -698.96x
expr               : count by(__name__) ({__name__=~".+"})
poolGetCount       : 0
poolGetSuccessCount: 0
poolPutCount       : 0

MEM STATS DIFF:   	promql-eval 	promql-eval - AFTER 	-> Delta
    HeapAlloc  : 	871.80M 	1.56G 			-> 683.37M
    HeapSys    : 	1.81G 		1.88G 			-> 67.11M
    HeapObjects: 	13.52M 		22.38M 			-> 8.86M

```

</details>

brian-brazil:
HeapAlloc is generally what I'd use. I'd be interested in dropping the pool, even if it was a small performance hit. It makes optimisation hard.

ppanyukov:
> I'd be interested in dropping the pool

Just it benchmarks with/without pool. Killing the pool makes some benchmarks a bit slower, slightly more allocs in some of them. But it addition bytes/op show some crazy increases.



```
benchmark                                                                                                              old bytes     new bytes     delta
BenchmarkRangeQuery/expr=a_one,steps=1000-16                                                                           8468          24817         +193.07%
BenchmarkRangeQuery/expr=a_ten,steps=100-16                                                                            14171         31769         +124.18%
BenchmarkRangeQuery/expr=a_ten,steps=1000-16                                                                           21219         184729        +770.58%
BenchmarkRangeQuery/expr=a_hundred,steps=100-16                                                                        86337         262313        +203.82%
BenchmarkRangeQuery/expr=a_hundred,steps=1000-16                                                                       146745        1781832       +1114.24%
BenchmarkRangeQuery/expr=rate(a_ten[1m]),steps=1000-16                                                                 24738         188476        +661.89%
BenchmarkRangeQuery/expr=rate(a_hundred[1m]),steps=1000-16                                                             176835        1812550       +925.00%
BenchmarkRangeQuery/expr=rate(a_one[1m]),steps=10000-16                                                                40259         203686        +405.94%
BenchmarkRangeQuery/expr=rate(a_ten[1m]),steps=10000-16                                                                139119        1779392       +1179.04%
BenchmarkRangeQuery/expr=rate(a_hundred[1m]),steps=10000-16                                                            1156695       17518945      +1414.57%
```

Interesting?

brian-brazil:
That's not unexpected, it's those allocs that the pool is preventing. Any increases in allocations would need to be a lot smaller than that.

ppanyukov:
Short update.

I've added to the current patch to put the overallocated slice back into pool in the resize path, so that we reuse it in the next loop iteration.

In the absence of benchmarks for sparse datasets, the current benchmarks do not show much of interest:

- Performance: some bits faster, some bits slower. Overall the patched version seems consistently faster to run all benchmarks (`orig: 414.809s` vs `patch: 411.111s`). Considering this is all on a laptop with CPU stepping etc (I ran each kind twice and took second result), variation is probably within margins?

- Allocs: absolutely zero diffs. Except two tests which show reduction of `-1.48%` and `-12.97%`

- Bytes/op: broadly no difference except a couple of cases.

I also tried bechmarks with 1.14Beta, no significant differences in anything there either.

image

image

image

Files:

[orig.txt](https://github.com/prometheus/prometheus/files/3988504/orig.txt)

[patch.txt](https://github.com/prometheus/prometheus/files/3988506/patch.txt)

[diff.txt](https://github.com/prometheus/prometheus/files/3988507/diff.txt)

PS. I may have a look into new benchmarks over xmas if I get bored with day time TV :)

ppanyukov:
Also quick question:

What does make this `_, v, ok := ev.vectorSelectorSingle(it, e, ts)` return OK or NOT OK?

Is it possible to figure out up-front how often we wil get OK? So that we can size the points array better?

I looked at the code and it's a bit complicated :) 

Full extract:

```
for i, s := range e.series {
        _, v, ok := ev.vectorSelectorSingle(it, e, ts)  // <-- this
        if ok {
            if ev.currentSamples < ev.maxSamples {
                ss.Points = append(ss.Points, Point{V: v, T: ts})
                ev.currentSamples++
            } else {
                ev.error(ErrTooManySamples(env))
            }
        }
    }
}
```

brian-brazil:
False means no sample was returned. You can presume that it's always true.

ppanyukov:
> False means no sample was returned. You can presume that it's always true.

Yes, I get it :) 

What I don't get is what would lead to samples being found or not. There is some correlation between data, query, step, and range. But I can't quite figure out what exactly it is.

OK, let's make it even simpler. Given same data, query, and range. Just varying "step" parameter I observe differences in how often the sample data is found. Why? What drives this?

brian-brazil:
A series may have gone stale, or target disappeared.

brian-brazil:
Did you get a chance to work on benchmarks to cover this?

roidelapluie:

Given the lack of updates from the author, we have decided to close this pull request.

PR #6373, opened 2019-11-25
fix subquery with offset handling in rules

bjakubski:
Hi,
Currently any rule that contains subquery with offset is incorrectly executed and displayed.
The offset part is completely missing from query in such case.

This PR seems to fix this

@fabxc  @bwplotka

### DIFF
--- a/promql/printer.go
+++ b/promql/printer.go
@@ -128,7 +128,11 @@ func (node *SubqueryExpr) String() string {
 	if node.Step != 0 {
 		step = model.Duration(node.Step).String()
 	}
-	return fmt.Sprintf("%s[%s:%s]", node.Expr.String(), model.Duration(node.Range), step)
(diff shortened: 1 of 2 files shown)


brian-brazil:
Why is it incorrect executed? I thought we'd fixed that when the printer was off.

bjakubski:
This may be incorrect on my part. I was about to write that is is *shown* incorrectly, but decided to check the execution. I've only now realised that I've checked execution on 2.12, so execution may be ok in current version.

context: I've encountered this general problem when I was debugging why rules in thanos ruler are incorrect

brian-brazil:
Thanks!

Hmm, looks like we're still going via string and back so this was never fixed systemically. We should probably add something to hang off exisiting unittests to catch this, as this is far from the first time this has come up.

PR #6375, opened 2019-11-26
support rule format with json

00arthur00:
support json format

### DIFF
--- a/pkg/rulefmt/rulefmt.go
+++ b/pkg/rulefmt/rulefmt.go
@@ -96,7 +96,7 @@ type Rule struct {
 	Record      string            `yaml:"record,omitempty" json:"record,omitempty"`
 	Alert       string            `yaml:"alert,omitempty" json:"alert,omitempty"`
 	Expr        string            `yaml:"expr" json:"expr"`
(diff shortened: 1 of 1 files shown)


brian-brazil:
Can you explain what you're trying to do? This file is part of the internals of Prometheus, which only uses YAML.

00arthur00:
We use consul to manage our promethues rule configuration, internally we use json. So add it.
Although yaml is an option, yaml's indentation is challengable for consul-template writer. JSON doesn't have this promblem. It's good for us.
I think others may have this requirement.

00arthur00:
@brian-brazil

brian-brazil:
These are internal structs, we don't generally change them to support things that aren't features of Prometheus itself as any such change would either constrain us or get removed in passing by some other change. The most we'll usually do is make something public, which these already are.

For these configs JSON will parse as YAML, so you shouldn't need to change anything.

roidelapluie:
I agree with Brian. Lots of people actually generate Json for alerts and it just works out of the box because json is a subset of yaml.

00arthur00:
Thanks. @brian-brazil  @roidelapluie 
Will close it.

PR #6378, opened 2019-11-26
Change desired shards metric to report raw calculated value

csmarchbanks:
Output desired shards as the calculated float rather than the integer that may be incorrect by up to 30%.

Also, the reshard log line was trying to log the metric rather than the value which was not working.

cc: @cstyan

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -500,7 +500,7 @@ func (t *QueueManager) updateShardsLoop() {
 			// to stay close to shardUpdateDuration.
 			select {
 			case t.reshardChan <- desiredShards:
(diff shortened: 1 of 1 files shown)


csmarchbanks:
@cstyan I would love a review when you have a moment :)

storage/remote/queue_manager.go
```diff
@@ -500,7 +500,7 @@ func (t *QueueManager) updateShardsLoop() {
 			// to stay close to shardUpdateDuration.
 			select {
 			case t.reshardChan <- desiredShards:
-				level.Info(t.logger).Log("msg", "Remote storage resharding", "from", t.numShards, "to", numShards)
+				level.Info(t.logger).Log("msg", "Remote storage resharding", "from", t.numShards, "to", desiredShards)
```
cstyan: :+1:

PR #6380, opened 2019-11-26
[ENHANCEMENT] scrape: The scrape module use too much memory 21.8GB

liguozhong:
image

More information is being collected.
We will try to optimize this problem.


liguozhong:
We are trying to modify the scrape code. When we confirm that the code is valid and can reduce the memory, we will submit the changed code.

brian-brazil:
When there's something to review, please send a PR.

PR #6381, opened 2019-11-26
promql: Allow injecting fake tokens into the generated parser

slrtbtfs:
Yacc grammars do not support having multiple start symbols.

To work around that restriction, it is possible to inject fake tokens into the lexer stream,
as described here https://www.gnu.org/software/bison/manual/html_node/Multiple-start_002dsymbols.html .

This is part of the parser rewrite effort described in #6256.

### DIFF
--- a/promql/lex.go
+++ b/promql/lex.go
@@ -193,6 +193,10 @@ const (
 	GROUP_RIGHT
 	BOOL
 	keywordsEnd
+
+	startSymbolsStart
+	// Start symbols for the generated parser.
+	startSymbolsEnd
 )
 
 var key = map[string]ItemType{
--- a/promql/parse.go
+++ b/promql/parse.go
@@ -378,9 +378,14 @@ func (p *parser) Error(e string) {
 // https://www.gnu.org/software/bison/manual/html_node/Multiple-start_002dsymbols.html .
 // Only the Lex function used by the generated parser is affected by this injected item.
 // Trying to inject when a previously injected item has not yet been consumed will panic.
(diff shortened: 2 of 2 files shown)


brian-brazil:
I'm confused, how is this even a problem?

slrtbtfs:
Right now the parser already has several entry points:

* `ParseExpr()` accepts any valid PromQL expression.
* `ParseMetric()` accepts only a single metric. I guess its used for scraping metrics.
* `ParseMetricSelector()` accepts only a single metric and returns a set of label matchers.

So technically  the PromQL parser parses 3 different formal languages, that happen to share some syntax rules.

By using this start symbol trick, these can actually share the same grammar.

---

While the parser rewrite is in progress there will be even more entry points, e.g. one of metric selectors, so parts of the parsing can be done with a generated parser while the the old parser is used for the rest.

slrtbtfs:
There already was something similar in the old generated parser: https://github.com/prometheus/prometheus/blob/0.10.0/rules/parser.y#L37

promql/lex.go
```diff
@@ -193,6 +193,10 @@ const (
 	GROUP_RIGHT
 	BOOL
 	keywordsEnd
+
+	startSymbolsStart
+	// start symbols for the generated parser
```
brian-brazil: Capital letter and full stop.

slrtbtfs: Oh no, not again! :man_facepalming:

brian-brazil: Thought the bowl of petunias.

brian-brazil:
Thanks!

PR #6382, opened 2019-11-26
React UI: Graph refactoring

boyskila:
In this PR established as helper functions form ```Graph.tsx``` are moved into file called ```GraphHelpers.ts```. Their corresponding test are already part of a file called ```GraphHelpers.test.tsx``` and all Graph related components are grouped in a folder ```graph```
Overall nothing functional is touched, just props from Graph.tsx on which they depend on are now passed as params.


boyskila:
@juliusv can you please review

juliusv:
:+1: Thanks :)

PR #6389, opened 2019-11-28
Update and re-work maintainers file.

### LINKED ISSUE #4714: Update MAINTAINERS.md
@fabxc 's email address (fabian.reinartz@coreos.com) seems outdated.

Beyond that formally required update, does the current state of MAINTAINERS.md still represent reality?

Local storage is still just missing here (which is worse in combination with https://github.com/prometheus/tsdb/issues/403 ).

Do the other focus areas still make sense? Both their layout as well as the maintainers assigned.

As an example, SD was seriously broken for a long time (rendering 2.1, 2.2, and 2.3 useless in larger production scenarios). My vague memory tells me that things got rectified mostly by @krasi-georgiev with no noticeable shepherding or mentoring by the maintainers.

And then there are rather trivial formalities like #3399 that might still have nasty consequences, given the visibility and importance of this project. It would be great if somebody felt responsible for those.

brian-brazil:
For sanity I'm not going to try to distinguish the things I'm maintainer
for because I'm the right person (e.g. template) vs things I'm the
maintainer for because someone has to be (e.g. pkg). Also add a general
warning to handle that this is more nuanced than it's worth trying to
capture, and relatedly always going to be out of date.

I didn't put Bartek down for remote read, as particular functions of
particlar files seems a a bit fine grained.

Fixes #4714

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,10 +1,21 @@
-Maintainers of this repository with their focus areas:
+@brian-brazil is the main/default maintainer, some parts of the codebase have other maintainers:
 
-* Brian Brazil <brian.brazil@robustperception.io> @brian-brazil: Console templates; semantics of PromQL, service discovery, and relabeling.
-* Fabian Reinartz <freinartz@google.com> @fabxc: PromQL parsing and evaluation; implementation of retrieval, alert notification, and service discovery.
-* Julius Volz <julius.volz@gmail.com> @juliusv: Web UI.
-* Krasi Georgiev <kgeorgie@redhat.com> @krasi-georgiev: TSDB - the storage engine.
-* Ganesh Vernekar <cs15btech11018@iith.ac.in> @codesome: TSDB - the storage engine.
-* Chris Marchbanks <csmarchbanks@gmail.com> @csmarchbanks: Remote write integration.
-* Callum Styan <callumstyan@gmail.com> @cstyan: Remote write integration.
-* Bartłomiej Płotka <bwplotka@gmail.com> @bwplotka: Remote read integration.
+* `cmd`
+  * `promtool`: @simonpasquier
+* `discovery`
+  * `k8s`: @brancz
+* `documentation`
+  * `prometheus-mixin`: @beorn7
+* `storage`
+  * `remote`: @csmarchbanks, @cstyan
+* `tsdb`: @codesome, @krasi-georgiev
+* `web`
+  * `ui`: @juliusv
+* `Makefile` and related build configuration: @simonpasquier, @SuperQ
+
+For the sake of brevity all subtrees are not explicitly listed. Due to the size
+of this repository, the natural changes in focus of maintainers over time, and
+nuances of where particular features live, this list will always be incomplete
+and out of date. However the listed maintainer(s) should be able to direct a
+PR/question to the right person.
+


beorn7:
I think we should try go get a :+1: from everyone whose GH handle shows up in this change.

roidelapluie:
Should we move to a CODEOWNERS file?

beorn7:
The convention is to have a MAINTAINERS.md file in every repo. Changing that would be confusing. What would that change bring in value?

roidelapluie:
https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners

Automatically assign pull requests to people

beorn7:
Interesting. We should keep it in mind. But I think currently, the mapping is not complete or strict enough to really benefit. Perhaps we'll get there.

MAINTAINERS.md
```diff
@@ -1,10 +1,21 @@
-Maintainers of this repository with their focus areas:
+@brian-brazil is the main/default maintainer.
```
juliusv: How about making the list below clearer:

```suggestion
@brian-brazil is the main/default maintainer. Some parts of the codebase have other maintainers:
```

MAINTAINERS.md
```diff
@@ -1,10 +1,21 @@
...
+
+* cmd
```
juliusv: I'd put direct path / file names into code backticks.

MAINTAINERS.md
```diff
@@ -1,10 +1,21 @@
...
+  * promtool: @simonpasquier
+* discovery
+  * k8s: @brancz
+* documentation
+  * prometheus-mixin: @beorn7
+* storage
+  * remote: @csmarchbanks, @cstyan
+* tsdb: @codesome, @krasi-georgiev
```
bwplotka: Super tiny nit: isn't TSDB a part of storage.

brian-brazil: No, they're two different directories. Storage has all the interfaces (plus remote storage).

MAINTAINERS.md
brian-brazil: Done

MAINTAINERS.md
brian-brazil: I wasn't sure how this would end up rendering, it's still not great after applying this change.

brian-brazil:
> However, I agree with @cstyan that it's similar to the SD implementation.

If the remote read endpoint was in its own directory, I've have put your name on it in this PR. However that's not the case.

MAINTAINERS.md
bwplotka: Ups, it's directory-based indeed, make sense.

beorn7:
Perhaps then list three maintainers for remote storage with their "sub focus area" in parentheses?

```
* `storage`
  * `remote`: @csmarchbanks (write), @cstyan (write), @bwplotka (read)
```

brian-brazil:
The remote read endpoint code lives under `web/api/v1` though.

beorn7:
I guess the partition by directory is not perfect. But perhaps it doesn't have to be.

beorn7:
Or do you mean there is no remote read code at all in `storage/remote`?

brian-brazil:
There's a bit of it in there.

brancz:
lgtm :+1:

beorn7:
Only @krasi-georgiev is still pending. Since his responsibility hasn't really changed in this PR, I'll merge this now.

Thanks everyone.

Now back to work and make sure the issues and PRs in your area of responsibilities get triaged and assigned and such. :o)

krasi-georgiev:
yes sorry for the late reply LGTM

PR #6390, opened 2019-11-28
Allow alerts to explicitly specify the ValidUntil field that is sent to alertmanager

Shadi:
Alerts EndsAt is being set to 3 multiplied by whatever value is larger between resendDelay and interval, this change will allow for explicitly setting EndsAt at an alert level.

The use case we ran into is that we don't want to throttle the alerts being sent from prometheus by a long period, and we have an alert with short evaluation interval, and still we would like alertmanager to consider the alert to be resolved only if it didn't receive it as firing for a period of 10 minutes, before PR#4550 that could have been accomplished by setting resolve_timeout, but after the changes in #4550, it is now coupled with resendDelay and interval, this PR will allow to have an alert with short evaluation interval, and higher validUntil value, without having to use resendDelay option.

Related PRs:
#4538
#4550

Signed-off-by: Shadi Abdelfatah <shadi.abdelfatah@gmail.com>

### DIFF
--- a/pkg/rulefmt/rulefmt.go
+++ b/pkg/rulefmt/rulefmt.go
@@ -97,6 +97,7 @@ type Rule struct {
 	Alert       string            `yaml:"alert,omitempty"`
 	Expr        string            `yaml:"expr"`
 	For         model.Duration    `yaml:"for,omitempty"`
+	ValidFor    model.Duration    `yaml:"validFor,omitempty"`
 	Labels      map[string]string `yaml:"labels,omitempty"`
 	Annotations map[string]string `yaml:"annotations,omitempty"`
 }
@@ -121,6 +122,9 @@ func (r *Rule) Validate() (errs []error) {
 		}
 		if r.For != 0 {
 			errs = append(errs, errors.Errorf("invalid field 'for' in recording rule"))
+		}
+		if r.ValidFor != 0 {
+			errs = append(errs, errors.Errorf("invalid field 'validFor' in recording rule"))
 		}
 		if !model.IsValidMetricName(model.LabelValue(r.Record)) {
 			errs = append(errs, errors.Errorf("invalid recording rule name: %s", r.Record))
--- a/rules/alerting.go
+++ b/rules/alerting.go
@@ -116,6 +116,8 @@ type AlertingRule struct {
 	// The duration for which a labelset needs to persist in the expression
 	// output vector before an alert transitions from Pending to Firing state.
 	holdDuration time.Duration
+	// Explicit value for alert EndsAt field
+	validDuration time.Duration
 	// Extra labels to attach to the resulting alert sample vectors.
 	labels labels.Labels
 	// Non-identifying key/value pairs.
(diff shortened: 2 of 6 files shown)


brian-brazil:
I see no reason that this should ever be user configurable at the alert level (or even at the Prometheus level). This is an internal implementation detail of the Prometheus->Alertmanager protocol that should just work.

> and still we would like alertmanager to consider the alert to be resolved only if it didn't receive it as firing for a period of 10 minutes, 

That's not how it works. This only matters when Prometheus dies and doesn't come back up, which shouldn't happen often.

Shadi:
I might be misunderstanding when an alert is considered resolved by alertmanager, my current understanding is that if an alert is firing, prometheus will keep sending that to alertmanager every evaluation interval, once it is not then it will stop sending anything about it, alert manager still has the EndsAt field and will only consider the alert to be resolved after that period expires.

We have an alert that fires for a short period, so alert manager will send a notification about it, then the condition goes away, alertmanager will send a notification that this alert is resolved, after it is resolved if it fires again then alertmanager will send a new alert, we are trying to avoid this by increasing the EndsAt period, so that the alertmanager will only send a notification that this alert is resolved only if it didn't receive a message about it from prometheus for a specified period that is longer than 3*interval.

hoffie:
> We have an alert that fires for a short period, so alert manager will send a notification about it, then the condition goes away, alertmanager will send a notification that this alert is resolved, after it is resolved if it fires again then alertmanager will send a new alert,

Couldn't you avoid this either by tuning the Prometheus alert rule or by making use of Alertmanager's [grouping features](https://www.robustperception.io/whats-the-difference-between-group_interval-group_wait-and-repeat_interval)?

There was also a [PromCon talk](https://promcon.io/2019-munich/talks/improved-alerting-with-prometheus-and-alertmanager/) by @roidelapluie. It had some clever tricks for special alerting demands (hysteresis; see page 16 in the slides).

Shadi:
@hoffie Thanks for the slides, Hysteresis are similar to what we are aiming at.
in slides expression uses average, but we don't want to be alerted based on average, instead I want to get an alert when the temperature gets to above 27 at any point in time, and I want to the alert manager to consider the alert to be resolved when it has been below 27 for 5 minutes.

Using average or for clause will slow the response time to the alerts.

roidelapluie:
```
(temperature_celcius > 27)
or (max_over_time(temperature_celcius[5m]) > 27 and
count without (alertstate, alertname, priority)
  ALERTS{
  alertstate="firing",
  alertname="temperature is above threshold"
})
```

roidelapluie:
Then your alert should just be (max_over_time(temperature[5m])) > 5m

Shadi:
@roidelapluie Thanks, max_over_time seems to be what I'm looking for.

regarding this PR, I think having it will simplify some of the queries, but if you think it is unnecessary or adds an undesired complexity then feel free to close it.

brian-brazil:
This PR would only make confusion around this worse I think, and likely introduce it's own problems.

Issue #6268, opened 2019-11-04
React UI: Bring back interactive legend


juliusv:
The graph legend in the classic UI has three modes of interaction, see e.g. http://demo.robustperception.io:9090/graph?g0.range_input=1h&g0.expr=rate(node_cpu_seconds_total%5B5m%5D)&g0.tab=0

- Hovering over legend items highlights the hovered series in the graph.
- Clicking on the series name in the legend toggles showing only that series in the graph.
- Clicking on the checkmark of a legend item turns off that series in the graph.

It would be good to bring back some, if not all, of these features in the new React UI.

boyskila:
Anyone working on this? I would like to give it a try, just to finish with my current PR :)

juliusv:
@boyskila Feel free!

roidelapluie:
@juliusv Can we reopen this?

We miss the ability to select multiple lines in legend (and the unselect one of them..)

juliusv:
@roidelapluie Yeah, I ended up telling @boyskila to omit the multi-deselect for now and see if someone really wants it :) If we do want to reintroduce it, I wonder if we can do it better somehow, since the old UI only allowed you to explicitly *de*select series that you didn't want to see, rather than the other way around. I want to avoid adding extra UI elements as much as possible...

Do you have an opinion on how this should best work?

roidelapluie:
I do not agree with you. The old UI allow both.

- Clicking on the labels would select the one you want, then click the checkboxes to add others.
- Deselect checkbox to remove one.

Comparing some metrics is really useful, I rarely see a case where I would just see one.

roidelapluie:
Without addind new UI elements, we could work with SHIFT and CTRL

CTRL+click = exclude
SHIFT+click = include

boyskila:
> Comparing some metrics is really useful, I rarely see a case where I would just see one.

Yes, it's gonna be useful for the user to have the possibility to select 2 or more series.
About the series toggle, I guess just CTRL + click will be enough.

juliusv:
> The old UI allows both.

Today I learned :rofl: The fact that I wrote the old UI and didn't know this (because it was a built-in feature of the old graphing library) is a good sign that this feature wasn't very discoverable from a UX perspective.

I generally like the idea of modifier keys to switch between behaviors. But then we have even worse discoverability features unless we write it out as instructions somewhere. Hm. I'll think about the UI bits a bit more...

[...]

PR #6391, opened 2019-11-28
Graph multi series select

boyskila:
In this PR is implemented multiple series selection from the Graph legend. The behavior is as follows.
   1. Clicking on an item will make it active and all other items will be deselected
   2. Clicking on an item will disable hover effect on the Graph series
   3. Clicking on an item + holding the CTRL key will not deselect other items if there are such
   4. Once multiple selections are made just clicking on an active item(without Ctrl key) will make it 
       inactive but all other selections will remain
Fixes: #6268

### DIFF
--- a/web/ui/react-app/src/graph/Graph.tsx
+++ b/web/ui/react-app/src/graph/Graph.tsx
@@ -2,7 +2,7 @@ import $ from 'jquery';
 import React, { PureComponent } from 'react';
 import ReactResizeDetector from 'react-resize-detector';
 
-import SeriesName from '../SeriesName';
+import { Legend } from './Legend';
 import { Metric, QueryParams } from '../types/types';
 import { isPresent } from '../utils/func';
 import { normalizeData, getOptions, toHoverColor } from './GraphHelpers';
@@ -32,30 +32,29 @@ export interface GraphSeries {
 
 interface GraphState {
   chartData: GraphSeries[];
-  selectedSeriesIndexes: number[];
 }
 
 class Graph extends PureComponent<GraphProps, GraphState> {
   private chartRef = React.createRef<HTMLDivElement>();
   private $chart?: jquery.flot.plot;
   private rafID = 0;
+  private selectedSeriesIndexes: number[] = [];
 
   state = {
     chartData: normalizeData(this.props),
-    selectedSeriesIndexes: [] as number[],
   };
 
   componentDidUpdate(prevProps: GraphProps) {
     const { data, stacked } = this.props;
     if (prevProps.data !== data) {
-      this.setState({ selectedSeriesIndexes: [], chartData: normalizeData(this.props) }, this.plot);
+      this.selectedSeriesIndexes = [];
+      this.setState({ chartData: normalizeData(this.props) }, this.plot);
     } else if (prevProps.stacked !== stacked) {
-      const { selectedSeriesIndexes } = this.state;
       this.setState({ chartData: normalizeData(this.props) }, () => {
-        if (selectedSeriesIndexes.length === 0) {
+        if (this.selectedSeriesIndexes.length === 0) {
           this.plot();
         } else {
-          this.plot(this.state.chartData.filter((_, i) => selectedSeriesIndexes.includes(i)));
+          this.plot(this.state.chartData.filter((_, i) => this.selectedSeriesIndexes.includes(i)));
         }
       });
     }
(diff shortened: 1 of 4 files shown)


boyskila:
@juliusv @roidelapluie please review.

web/ui/react-app/src/graph/Graph.tsx
```diff
@@ -123,6 +151,11 @@ class Graph extends PureComponent<GraphProps, GraphState> {
               <SeriesName labels={labels} format />
             </div>
           ))}
+          {chartData.length > 1 && (
+            <div className="pl-1 mt-1 text-muted" style={{ fontSize: 13 }}>
+              Click: toogle serie, CTRL + click: toogle multiple series
```
juliusv: Can we style this identically to the query stats?

"toogle" -> "toggle"
"serie" -> "series"

juliusv:
@boyskila Tests are still failing, they probably need an update?

juliusv:
Looks great besides the nits!

roidelapluie:
Excellent. I have tested it and have 2 comments:

A) Legend
Click: toggle series, CTRL + click: toggle multiple series
Maybe: Click: *select* series, CTRL + click: toggle multiple series

B) CTRL+Click behaviour when no one is selected
When no timeserie is selected, I think that CTRL+CLICK should "remove" that timeserie (e.g they all become selected but that one).

Good job!

boyskila:
> Excellent. I have tested it and have 2 comments:
> 
> A) Legend
> Click: toggle series, CTRL + click: toggle multiple series
> Maybe: Click: _select_ series, CTRL + click: toggle multiple series
> 
> B) CTRL+Click behaviour when no one is selected
> When no timeserie is selected, I think that CTRL+CLICK should "remove" that timeserie (e.g they all become selected but that one).
> 
> Good job!

Thanks for the feedback @roidelapluie @juliusv  ! I think ```Click: select series``` is more appropriate.
Regarding point B If I understand correctly you suggest on ```Ctrl + click``` to deselect the series but if there is series already selected to continue working as described in point A.
This functionality eventually can be delegated to ```Shift + click``` wdyt?

roidelapluie:
No

B is:

From :

I "Click" federate_1 and I have:




From:

I "CTRL+Click" federate_1 and I have:

boyskila:
Cool! So I understand it correctly. This is what I describe in the comment above :). Just wonder isn't better Shift + click to handle it and to leave Ctrl + click just to toggle multiple series.

roidelapluie:
I prefer ctrl because it is the normal ctrl behaviour, we just need to consider than they are all selected at the beginning

roidelapluie:
ctrl+click on a selected TS deselects it unless they are all selected ; that is a bug

boyskila:
Yeah, I'm Ok with that. @juliusv do you have any comment on that or to proceed?

juliusv:
I trust you two :)

boyskila:
@juliusv @roidelapluie ctrl + click adjustments are done. Please review when you have time, thanks.
P.s did small refactoring as well by extracting the Legend as a component.

roidelapluie:
Great work!

One last issue however:


When I click (without CTRL) the serie gets deselected instead of selected:

boyskila:
> Great work!
> 
> One last issue however:
> 
> 
> When I click (without CTRL) the serie gets deselected instead of selected:
> 

Done

juliusv:
Hmm, it doesn't seem to work as expected currently. E.g. if I graph `rate(node_cpu_seconds_total[5m])` on https://deploy-preview-6391--prometheus-react.netlify.com/ and I click on one of the series in the legend, it initially selects only the one series, clicking on it a second time also shows other series in the graph again, but in light color even after leaving the legend, and a third time does nothing anymore.

juliusv:
Also, when reloading the graph (even with totally different series), it seems to remember the selected series (it should forget it).

boyskila:
@juliusv Yes, thanks for catching those bugs. They appeared after the small refactoring I have made. I think now everything is working as expected.

web/ui/react-app/src/graph/Legend.tsx
```diff
@@ -0,0 +1,76 @@
...
+    const { selectedIndexes } = this.state;
+
+    let selected = [index];
+    if (ev.ctrlKey) {
+      const { chartData } = this.props;
+      if (selectedIndexes.includes(index)) {
+        selected = selectedIndexes.filter(idx => idx !== index);
+      } else {
+        selected =
+          // Flip the logic - In case non is selected ctrl + click should deselect clicked series.
```
juliusv: "non" -> "none"

web/ui/react-app/src/graph/Legend.tsx
```diff
@@ -0,0 +1,76 @@
...
+            : [...selectedIndexes, index]; // Select multiple.
+      }
+    } else if (selectedIndexes.length === 1 && selectedIndexes.includes(index)) {
+      selected = [];
+    }
+
+    this.setState({ selectedIndexes: selected });
+    this.props.onSeriesToggle(selected, index);
+  };
+  render() {
```
juliusv: nit: add empty line above

juliusv:
Seems to behave pretty awesome now, just some nits!

PR #6393, opened 2019-11-29
discovery/kubernetes: fix client metrics

simonpasquier:
The Kubernetes client records workqueue duration and latency metrics as seconds so there's no need to convert the values from microseconds to seconds anymore.

https://github.com/prometheus/prometheus/blob/eba5e0271837b00c6fd5c63ff1a3d0f9bafdaf11/vendor/k8s.io/client-go/util/workqueue/metrics.go#L123

https://github.com/prometheus/prometheus/blob/eba5e0271837b00c6fd5c63ff1a3d0f9bafdaf11/vendor/k8s.io/client-go/util/workqueue/metrics.go#L135

The cache metrics (prometheus_sd_kubernetes_cache_*) are removed because they aren't used anymore by the client though still exposed by its API. See https://github.com/kubernetes/client-go/commit/a1320a3a472a94bbf467d743f845b1269fc0398e.

### DIFF
--- a/discovery/kubernetes/client_metrics.go
+++ b/discovery/kubernetes/client_metrics.go
@@ -19,15 +19,11 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 
-	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/metrics"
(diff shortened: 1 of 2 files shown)


brancz:
lgtm :+1:

Issue #6386, opened 2019-11-27
React UI: Create /service-discovery page


juliusv:
The React UI is still missing the [/service-discovery](http://demo.robustperception.io:9090/service-discovery) page - this still needs to be implemented (and server-side APIs extended as necessary).

Harkishen-Singh:
hi @juliusv! I would like to take up this.

Issue #3843, opened 2018-02-14
Add Alertmanager service discovery to Service Discovery WebUI


jamtur01:
Currently, the Service Discovery WebUI page lists targets found by service discovery. I think it'd be useful if it also listed any Alertmanager targets found if the `alertmanager` configuration uses service discovery.

drholmie:
Hey there! Is there anyone still working on this? If not, I'd like to do so.

brian-brazil:
Not that I know of, go ahead.

drholmie:
Cool, thanks! Will work on it then.

natrimmer:
I believe #6494 should have marked this issue as closed.

krajorama:
Hello from the bug scrub!

Implemented via #6394 in the new UI. But the old UI will be removed in 4.0 anyway. 

Demo link : https://prometheus.demo.prometheus.io/service-discovery

machine424:
https://prometheus.demo.prometheus.io/alertmanager-discovery

PR #6394, opened 2019-11-29
Adds support service discovery page in react ui

Harkishen-Singh:
Fixes: #6386 

**Explanatory GIF**

ezgif-2-04ebae52c80d

[Video Link](https://drive.google.com/file/d/1qamyvUsfmVU1Fmqo-9_JiC4Dd4rY23dC/view)


juliusv:
Awesome, thanks, I'll try to take a look soon (somewhat busy weekend)!

cc @boyskila in case you're interested.

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -1,15 +1,199 @@
...
+                        ))}
+                      </div>
+                    </div>
+                    <div className="col-md-6">
+                      {i === 0 ? <div className="head">Target Labels</div> : null}
+                      <div key={this.props.value[i].labels[i] + '-' + i}>
+                        {Object.keys(this.props.value[i].labels).map((v, _) => (
+                          <div className="label-style" key={this.props.value[i].labels[v]}>
+                            {' '}
+                            {v}={this.props.value[i].labels[v]}{' '}
```
boyskila: Can you please use Badge component here as it is in https://github.com/prometheus/prometheus/blob/master/web/ui/react-app/src/pages/targets/TargetLabels.tsx

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -1,15 +1,199 @@
...
+              </div>
+            );
+          })}
+        </div>
+      </>
+    );
+  }
+}
+
+class Labels extends Component<LabelProps, LabelState> {
```
boyskila: Can you please replace this component with functional one with useState hook?

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -1,15 +1,199 @@
...
+        ) : (
+          <>
+            <button className="btn btn-primary btn-sm" onClick={this.toggleMore}>
+              Less
+            </button>
+            <div>
+              {Object.keys(this.props.value).map((_, i) => {
+                return (
+                  <div key={'inner-layer-' + i} className="row inner-layer">
+                    <div className="col-md-6">
```
boyskila: You can use Col from Reactstrap with md="6", although to be consistent with the other pages the entire structure can be created with a table

web/ui/react-app/src/App.css
```diff
@@ -197,3 +197,43 @@ div.time-input {
...
+    padding: 1%;
+}
+
+.head {
+    font-weight: 600;
+    margin-bottom: 2%;
+    border-bottom: 1px solid #DEE2E6;
+}
+
+.label-style {
```
boyskila: Once you wrap content in a Badge component these styles will become redundant

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -1,15 +1,199 @@
...
+
+interface LabelState {
+  showMore: boolean;
+}
+
+interface LabelProps {
+  value: any;
+}
+
+class Services extends Component<RouteComponentProps & PathPrefixProps, ServicesState> {
```
boyskila: Maybe both classes should live in their own files

boyskila:
Hi @Harkishen-Singh ! I tried to stick only to Labels component but I leave some comments for the Services.tsx as well

roidelapluie:
Can we tacle #3843 at the same time?

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -1,15 +1,199 @@
...
+      .catch(err => {
+        this.setState({ fetchServicesErrorMessage: err });
+      });
+  }
+
+  isHealthy = (health: string) => {
+    return health === 'up';
+  };
+
+  processTargets = (resp: any) => {
```
juliusv: It would be good to type the response similar to how it's done for /new/targets, same further below.

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -1,15 +1,199 @@
...
+        targets[name].total++;
+        if (this.isHealthy(health)) {
+          targets[name].active++;
+        }
+      }
+    }
+    return targets;
+  };
+
+  getLabels = (target: any) => {
```
juliusv: I'd probably just inline this rather than putting it in its own function (since it's trivial and not used in more than one place).

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -1,15 +1,199 @@
...
+    return {
+      discoveredLabels: target.discoveredLabels,
+      labels: target.labels,
+    };
+  };
+
+  processLabels = (resp: any) => {
+    const labels = Object.create({}),
```
juliusv: Any reason why not just:

```suggestion
    const labels = {},
```

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -1,15 +1,199 @@
...
+      activeTargets = resp.data.activeTargets;
```
juliusv: I feel it'd be more readable with two separate `const` definitions rather than using the comma operator and a newline.

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -1,15 +1,199 @@
...
+    fetch(this.props.pathPrefix + '/api/v1/targets')
+      .then(resp => {
+        if (resp.ok) {
+          return resp.json();
+        } else {
+          throw new Error('Unexpected response status when fetching services: ' + resp.statusText); // TODO extract error
+        }
+      })
+      .then(json => {
+        this.setState({ response: json, targets: this.processTargets(json), labels: this.processLabels(json) });
```
juliusv: Do we need to store `response: json` here? It doesn't seem to be used anywhere.

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -1,15 +1,199 @@
...
+      });
+  }
+
+  isHealthy = (health: string) => {
+    return health === 'up';
+  };
+
+  processTargets = (resp: any) => {
+    const activeTargets = resp.data.activeTargets,
+      targets: any = Object.create({});
```
juliusv: Any reason why not just:

```suggestion
      targets: any = {};
```

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -1,15 +1,199 @@
...
+
+    // Get targets of each type along with the total and active end points
+    for (const target of activeTargets) {
+      const name = target.scrapePool,
```
juliusv: I feel it'd be more readable with two separate `const` definitions rather than using the comma operator and a newline (since the indentation with that reads a bit weird). Same in other places.

Harkishen-Singh:
@juliusv @boyskila Thank you for the review. I will implement the suggestions asap. Sorry for the delay.

Harkishen-Singh:
@boyskila please have a look. I have implemented the suggestions.

web/ui/react-app/src/pages/Labels.tsx
```diff
@@ -0,0 +1,61 @@
...
+      {showMore ? (
+        <>
+          <div>
+            {Object.keys(value).map((_, i) => {
+              return (
+                <div key={i} className="row inner-layer">
+                  <div className="col-md-6">
+                    {i === 0 ? <div className="head">Discovered Labels</div> : null}
+                    <div>
+                      {Object.keys(value[i].discoveredLabels).map((labelName, j) => (
```
boyskila: Object entries will give you the key and value as a pair as first arg to the map callback ```map([key, val].....)```. Then this ```{`${labelName}="${value[i].discoveredLabels[labelName]}"`}```
can be shortenet to ```{`${key}="${val}"`}```

web/ui/react-app/src/App.css
```diff
@@ -212,6 +212,42 @@ div.time-input {
     margin-bottom: 20px;
 }
 
+.target-head {
```
boyskila: All styles will become redundant once you start using Reactstrap components

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -2,14 +2,129 @@ import React, { FC } from 'react';
...
+            </li>
+          ))}
+        </ul>
+        <hr />
+        <div className="outer-layer">
+          {Object.keys(labels).map((val: any, i) => {
+            const value = labels[val];
+            return (
+              <div id={val} key={Object.keys(labels)[i]} className="label-component">
+                <span className="target-head">
```
boyskila: I think target-head should be part of the labels Table. Hm, maybe ```Labels.tsx``` should become ```LabelsTable.tsx``` or something

Harkishen-Singh: removing `table-head` would make the job name lighter, hence decreasing the look and feel. However, I have renamed the said. Thanks.

web/ui/react-app/src/App.css
Harkishen-Singh: custom styling for bold fonts

web/ui/react-app/src/pages/Services.tsx
boyskila: What I meant here is not to remove the ```target-head``` but instead to make it part of LabelsTable.tsx. This way you can group it with more/less button and to add some proper styling to the wrapper since now there is no margin in between.
screen png

web/ui/react-app/src/App.css
boyskila: I can see big difference comparing with the old UI
old_vs_new

I think you can remove all the custom css without any worries.

web/ui/react-app/src/pages/LabelsTable.tsx
```diff
@@ -0,0 +1,63 @@
...
+    doToggle(!showMore);
+  };
+
+  return (
+    <>
+      <Button size="sm" color="primary" onClick={toggleMore}>
+        {showMore ? 'More' : 'Less'}
+      </Button>
+      {showMore ? (
+        <Table striped={true} className="table-head">
```
boyskila: I would suggest you to remove the className since it does nothing(or at least I coudn't find anyy css related to it) and to add ```bordered``` to the Table

web/ui/react-app/src/pages/LabelsTable.tsx
```diff
@@ -0,0 +1,63 @@
...
+              <th>Discovered Labels</th>
+              <th>Target Labels</th>
+            </tr>
+          </thead>
+          <tbody>
+            {value.map((_, i) => {
+              return (
+                <tr key={i}>
+                  <td>
+                    {Object.entries(value[i].discoveredLabels).map(([key, value], i) => {
```
boyskila: Suggestion:
```
const printLabels = (labels: Record<string, string>) => {
   return Object.entries(labels).map(([key, value]) => {
        return (
             <div key={key}>
                <Badge color="primary" className='mr-1'>
                     {`${key}="${value}"`}
                 </Badge>
              </div>
          );
    })
}
```
and then you can use it like
```
   <td>
      {printLabels(value[i].discoveredLabels)}
  </td>
   <td>
       {printLabels(value[i].labels)}
   </td>
```

web/ui/react-app/src/pages/LabelsTable.tsx
Harkishen-Singh: done

web/ui/react-app/src/pages/LabelsTable.tsx
Harkishen-Singh: done. thanks.

web/ui/react-app/src/pages/Services.tsx
Harkishen-Singh: fixed this.

web/ui/react-app/src/App.css
Harkishen-Singh: made the UI almost same to that of old.

Harkishen-Singh:
@boyskila please have a look.

Screenshot from 2019-12-21 15-32-19

boyskila:
@juliusv WDYT about the idea to combine name and button similar to Alerts?

image

juliusv:
@boyskila That's a good point to think about, but another thing for now: the SD page is most similar to the targets page, but that one also would need to be changed to this style then. For now I would prefer if the SD page got styled consistently with the targets page. E.g. the "show more" buttons have a different style on SD than on targets. And then we can think later whether we want to change both to the alerts page style.

web/ui/react-app/src/App.css
juliusv: If we adopt more of the styling of the existing /targets page here (which is a very similar page), we can lose more of these custom styles. E.g. on the targets page the headers are simply `h3` tags, and this should be consistent here. The "show less" / "show more" buttons there do have some custom styling, but it would be good to share that somehow between both pages too.

web/ui/react-app/src/App.css
```diff
@@ -212,6 +212,26 @@ div.time-input {
...
+    font-weight: 700;
+    font-size: large;
+}
+
+.label-component {
+    margin-bottom: 1%;
+}
+
+.table-outer-layer {
```
juliusv: I don't see any change to the page at all if I remove these styles (and even cranking the numbers up here doesn't seem to change anything in the table layout?).

web/ui/react-app/src/App.css
```diff
@@ -212,6 +212,26 @@ div.time-input {
...
+    margin: 1%;
+    margin-right: 5%;
+}
+
+.head {
```
juliusv: I don't think the `head` class is used anywhere? But probably these CSS comments will be obsolete anyway when making the structure / styling more consistent with the targets page.

web/ui/react-app/src/pages/LabelsTable.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const [showMore, doToggle] = useState(false);
+
+  const toggleMore = () => {
+    doToggle(!showMore);
+  };
+
+  return (
+    <div className="label-component">
+      <span className="target-head"> {name} </span>
+      <Button size="sm" color="primary" onClick={toggleMore}>
```
juliusv: When a group is expanded, the button touches the table below it without any spacing. But probably another comment that won't matter after adopting the target page header styles :)

web/ui/react-app/src/pages/LabelsTable.tsx
```diff
@@ -0,0 +1,57 @@
+import React, { FC, useState } from 'react';
+import { RouteComponentProps } from '@reach/router';
+import { Button, Badge, Table } from 'reactstrap';
+
+interface LabelProps {
+  value: Record<string, Record<string, string>>[];
+  name: string;
+}
+
+const printLabels = (labels: Record<string, string>) => {
```
juliusv: I know this came from @boyskila but generally I find the prefix `print` confusing for anything that doesn't print anything onto a console or similar. How about `formatLabels`?

web/ui/react-app/src/pages/LabelsTable.tsx
```diff
@@ -0,0 +1,57 @@
...
+  return Object.entries(labels).map(([key, value]) => {
+    return (
+      <div key={key}>
```
juliusv: I think you can omit the extra `div` wrapper here and put the key directly onto the `Badge`.

juliusv: Ah, you have the div for line breaks. Fair enough.

web/ui/react-app/src/pages/LabelsTable.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <Badge color="primary" className="mr-1">
+          {`${key}="${value}"`}
+        </Badge>
+      </div>
+    );
+  });
+};
+
+export const LabelsTable: FC<RouteComponentProps & LabelProps> = ({ value, name }) => {
+  const [showMore, doToggle] = useState(false);
```
juliusv: More customary for `useState` result names would be `showMore` and `setShowMore` (`doToggle` is a bit misleading because it doesn't toggle by itself).

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -2,14 +2,113 @@ import React, { FC } from 'react';
...
 
-const Services: FC<RouteComponentProps & PathPrefixProps> = ({ pathPrefix }) => (
-  <>
-    <h2>Service Discovery</h2>
-    <Alert color="warning">
-      This page is still under construction. Please try it in the <a href={`${pathPrefix}/service-discovery`}>Classic UI</a>.
-    </Alert>
-  </>
-);
+interface DiscoveredLabels {
```
juliusv: I think you can delete this interface (it contains some clearly testing field names, plus you never know what labels there will be anyway) and make any labels just the standard string->string map like in https://github.com/prometheus/prometheus/blob/213a8fe89a7308e73f22888a963cbf9375217cd6/web/ui/react-app/src/pages/targets/target.ts#L1-L3

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -2,14 +2,113 @@ import React, { FC } from 'react';
...
+  metrics_path: string;
+  scheme: string;
+  job: string;
+  my: string;
+  your: string;
+}
+
+interface ActiveTargets {
+  discoveredLabels: DiscoveredLabels[];
+  labels: any;
```
juliusv: Type this properly, like https://github.com/prometheus/prometheus/blob/213a8fe89a7308e73f22888a963cbf9375217cd6/web/ui/react-app/src/pages/targets/target.ts#L1-L3

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -2,14 +2,113 @@ import React, { FC } from 'react';
...
+interface DiscoveredLabels {
+  address: string;
+  metrics_path: string;
+  scheme: string;
+  job: string;
+  my: string;
+  your: string;
+}
+
+interface ActiveTargets {
```
juliusv: This interface is completely duplicating https://github.com/prometheus/prometheus/blob/213a8fe89a7308e73f22888a963cbf9375217cd6/web/ui/react-app/src/pages/targets/target.ts#L5-L14 - could you reuse the existing one?

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -2,14 +2,113 @@ import React, { FC } from 'react';
...
+  labels: any;
+  scrapePool: string;
+  scrapeUrl: string;
+  lastError: string;
+  lastScrape: string;
+  lastScrapeDuration: number;
+  health: string;
+}
+
+interface ServiceMap {
```
juliusv: This one duplicates https://github.com/prometheus/prometheus/blob/213a8fe89a7308e73f22888a963cbf9375217cd6/web/ui/react-app/src/pages/targets/ScrapePoolList.tsx#L11-L14.

We should probably find a good common place for these type definitions, but I guess for now it's ok to duplicate here since it's small. But maybe add a `// TODO: Deduplicate with https://github.com/prometheus/prometheus/blob/213a8fe89a7308e73f22888a963cbf9375217cd6/web/ui/react-app/src/pages/targets/ScrapePoolList.tsx#L11-L14`.

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -2,14 +2,113 @@ import React, { FC } from 'react';
...
+  activeTargets: ActiveTargets[];
+  droppedTargets: any[];
+}
+
+const Services: FC<RouteComponentProps & PathPrefixProps> = ({ pathPrefix }) => {
+  const { response, error } = useFetch<ServiceMap>(`${pathPrefix}/api/v1/targets`);
+  const processTargets = (response: ServiceMap) => {
+    console.warn(response);
```
juliusv: Left over from debugging and can be deleted?

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -2,14 +2,113 @@ import React, { FC } from 'react';
...
+    const activeTargets = response.activeTargets;
```
juliusv: You extract `activeTargets` into a separate const here, but not `droppedTargets`, although they are used in the same way below. Since you're accessing each only once, I'd probably omit this for `activeTargets`.

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -2,14 +2,113 @@ import React, { FC } from 'react';
...
+    for (const target of response.droppedTargets) {
+      const { scrapePool: name } = target;
+      targets[name].total++;
+    }
+
+    return targets;
+  };
+
+  const processLabels = (response: ActiveTargets[]) => {
+    const labels: any = {};
```
juliusv: Type the labels (as mentioned elsewhere).

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -2,14 +2,113 @@ import React, { FC } from 'react';
...
+        </ul>
+        <hr />
+        {Object.keys(labels).map((val: any, i) => {
+          const value = labels[val];
+          return <LabelsTable value={value} name={val} key={Object.keys(labels)[i]} />;
+        })}
+      </>
+    );
+  }
+  return null;
```
juliusv: If there's no data yet, we'll want to show a loading spinner like on the other pages (e.g. https://github.com/prometheus/prometheus/blob/8a9509b0a85430532ee9a01a9f8086ca916275bc/web/ui/react-app/src/pages/Config.tsx#L42).

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -2,14 +2,113 @@ import React, { FC } from 'react';
...
+      }
+      labels[name].push({
+        discoveredLabels: target.discoveredLabels,
+        labels: target.labels,
+      });
+    }
+
+    return labels;
+  };
+  let targets: any;
```
juliusv: I don't see a reason why these need to be defined up here? Can't they just be defined as consts down below where they are first set?

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -2,14 +2,113 @@ import React, { FC } from 'react';
...
+    }
+    for (const target of response.droppedTargets) {
+      const { scrapePool: name } = target;
+      targets[name].total++;
+    }
+
+    return targets;
+  };
+
+  const processLabels = (response: ActiveTargets[]) => {
```
juliusv: Looking at what `processTargets` and `processLabels` do, it would seem clearer to call them:

- `processTargets` -> `processSummary`
- `processLabels` -> `processTargets`

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -2,14 +2,113 @@ import React, { FC } from 'react';
...
+          total: 0,
+          active: 0,
+        };
+      }
+      targets[name].total++;
+      targets[name].active++;
+    }
+    for (const target of response.droppedTargets) {
+      const { scrapePool: name } = target;
+      targets[name].total++;
```
juliusv: When there is a dropped target, the page crashes with `Cannot read property 'total' of undefined` in this line. You'll need to ensure that `targets[name]` gets created for each pool before using, like in the active targets loop above.

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -2,14 +2,113 @@ import React, { FC } from 'react';
...
+            <li key={i}>
+              <a href={'#' + val}>
+                {' '}
+                {val} ({targets[val].active} / {targets[val].total} active targets){' '}
+              </a>
+            </li>
+          ))}
+        </ul>
+        <hr />
+        {Object.keys(labels).map((val: any, i) => {
```
juliusv: This table only shows active targets (if I fix the crash bug for when there's a dropped target). It also needs to show information about dropped targets, like the old UI:

dropped

You can introduce a dropped target for testing by adding the following scrape config to your Prometheus config:

```
  - job_name: 'testjob'
    static_configs:
      - targets: # Targets to probe through the Blackbox Exporter.
        - 'keep-this:8080'
        - 'drop-this:8080'
    relabel_configs:
      - source_labels: [__address__]
        regex: 'drop-this:8080'
        action: drop
```

juliusv:
Alright, added a bunch more comments. There still seems to be a lot to do. I would say let's focus on getting this one fixed up and right before attacking the rules page (which is again pretty complex).

web/ui/react-app/src/App.css
Harkishen-Singh: oh. My earlier versions had some changes with this. But the newer code is independent of this style. Thanks for pointing.

web/ui/react-app/src/pages/LabelsTable.tsx
Harkishen-Singh: I think, this was handled by `table-outer-layer`. Anyways, we could add a custom style to the label button as well.

web/ui/react-app/src/pages/LabelsTable.tsx
Harkishen-Singh: yes :)

web/ui/react-app/src/pages/Services.tsx
Harkishen-Singh: sorry. I had fixed this earlier, but the git revert would have created this again. Thanks for pointing.

web/ui/react-app/src/pages/Services.tsx
Harkishen-Singh: Sure.

web/ui/react-app/src/pages/Services.tsx
Harkishen-Singh: sure.

web/ui/react-app/src/pages/LabelsTable.tsx
boyskila: I think the proper way is to group the ```span``` and ```button``` in a div and just to apply bootstrap class ```mb-2``` to it

web/ui/react-app/src/pages/Services.tsx
boyskila: You can use ```Record<string, string>```

web/ui/react-app/src/pages/Services.tsx
boyskila: Any duplicated interface can be added and consumed from https://github.com/prometheus/prometheus/blob/master/web/ui/react-app/src/types/types.ts

boyskila:
> @boyskila That's a good point to think about, but another thing for now: the SD page is most similar to the targets page, but that one also would need to be changed to this style then. For now I would prefer if the SD page got styled consistently with the targets page. E.g. the "show more" buttons have a different style on SD than on targets. And then we can think later whether we want to change both to the alerts page style.

Yeah, I agree. I just mention that just because I don't like that the buttons are aligned differently. Maybe all pages which in some way looks similarly can be designed in a same way

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -1,15 +1,116 @@
...
+
+export interface TargetLabels {
+  discoveredLabels: Labels;
+  labels: Labels;
+}
+
+const Services: FC<RouteComponentProps & PathPrefixProps> = ({ pathPrefix }) => {
+  const { response, error } = useFetch<ServiceMap>(`${pathPrefix}/api/v1/targets`);
+
+  const processSummary = (response: ServiceMap) => {
```
boyskila: I would suggest you to split ```proccess``` functions into smaller reducer functions, to pull them out from the component and export them in order to test them. Something like:

```
export const activeTargetsSummary = (acc: any, { scrapePool }: Target) => {............};

export const droppedTargetsSummary = (acc: any, { job }: any) => {..........};

```
Then you can use them like:

```
const { activeTargets, droppedTargets } = response.data;
const activeSummary = activeTargtes.reduce(activeTargetsSummary, {});
const allTargetsSummary = droppedTartgets.reduce(droppedTargetsSummary, activeSummary);
```

Same can be applied for labels even I think both (summary and labels) can be combined into structure like this:
```
{
   active: number;
   total: number;
   labels: Array<{ discoveredLabels: Labels; labels: Labels }>
}
```

web/ui/react-app/src/pages/LabelsTable.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const toggleShowMore = () => {
+    setShowMore(!showMore);
+  };
+
+  return (
+    <div className="label-component">
+      <div className="label-button-style">
+        <span className="target-head"> {name} </span>
+        <Button size="sm" color="primary" onClick={toggleShowMore}>
```
boyskila: Button width now is based on the text size. Will look better if it has a fixed width.

web/ui/react-app/src/pages/LabelsTable.tsx
```diff
@@ -0,0 +1,64 @@
...
+          {showMore ? 'show less' : 'show more'}
+        </Button>
+      </div>
+      {showMore ? (
+        <Table striped={true} bordered={true}>
```
boyskila: size="sm" as the other tables in the app

web/ui/react-app/src/pages/LabelsTable.tsx
```diff
@@ -0,0 +1,64 @@
...
+export const LabelsTable: FC<RouteComponentProps & LabelProps> = ({ value, name }) => {
+  const [showMore, setShowMore] = useState(false);
+
+  const toggleShowMore = () => {
+    setShowMore(!showMore);
+  };
+
+  return (
+    <div className="label-component">
+      <div className="label-button-style">
```
boyskila: You don't need class for this since bootstrap provide such - just use class ```mb-2```

web/ui/react-app/src/pages/LabelsTable.tsx
```diff
@@ -0,0 +1,64 @@
...
+              <th>Discovered Labels</th>
+              <th>Target Labels</th>
+            </tr>
+          </thead>
+          <tbody>
+            {value.map((_, i) => {
+              return (
+                <tr key={i}>
+                  <td>{formatLabels(value[i].discoveredLabels)}</td>
+                  {value[i].labels.hasOwnProperty('dropped') ? (
```
boyskila: Can we drop labels at all for dropped targets? I think there is no need to set object with ```dropped: 'yes'``` field in it

web/ui/react-app/src/pages/targets/target.ts
```diff
@@ -13,6 +13,10 @@ export interface Target {
   health: string;
 }
 
+export interface DroppedTargets {
```
juliusv: `DroppedTargets` -> `DroppedTarget` since it's just one.

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -1,15 +1,116 @@
...
-    <h2>Service Discovery</h2>
-    <Alert color="warning">
-      This page is still under construction. Please try it in the <a href={`${pathPrefix}/service-discovery`}>Classic UI</a>.
-    </Alert>
-  </>
-);
+// TODO: Deduplicate with https://github.com/prometheus/prometheus/blob/213a8fe89a7308e73f22888a963cbf9375217cd6/web/ui/react-app/src/pages/targets/ScrapePoolList.tsx#L11-L14
+interface ServiceMap {
+  activeTargets: Target[];
+  droppedTargets: any[];
```
juliusv: `any[]` -> `DroppedTarget[]` since you're importing that now.

web/ui/react-app/src/pages/Services.tsx
```diff
@@ -1,15 +1,116 @@
...
+    }
+
+    for (const target of dropped) {
+      const { job: name } = target.discoveredLabels;
+      if (!labels[name]) {
+        labels[name] = [];
+      }
+      labels[name].push({
+        discoveredLabels: target.discoveredLabels,
+        labels: { dropped: 'yes' },
```
juliusv: A normal (non-dropped) target may include a label with the label name `dropped`, so you cannot just use labels to communicate whether a target has been dropped. You'll probably want to add a `dropped` boolean or something outside of the labels.

Harkishen-Singh: Sure. I have implemented an external boolean property.

juliusv:
So one more thing I discovered: when a target overrides the default job name (so the job name becomes different from the scrape pool name) and is dropped, it won't be associated with the right section anymore.

Config to reproduce:

```yml
scrape_configs:
  - job_name: default_job_name
    static_configs:
      - targets:
        - localhost:9090
        - localhost:12345
        labels:
          job: overridden_job_name
    relabel_configs:
      - action: drop
        source_labels: [__address__]
        regex: 'localhost:12345'
```

Old UI:

oldui

New UI:

newui

Currently the API does not return the name of the scrape pool for dropped targets. We should extend it to do so, so we can refer to it from the UI. But it's IMO ok to do after this PR as well.

web/ui/react-app/src/App.css
```diff
@@ -212,6 +212,19 @@ div.time-input {
     margin-bottom: 20px;
 }
 
+.label-button-style {
```
juliusv: This style is unused (no class of that name anywhere) and can be removed.

web/ui/react-app/src/pages/LabelsTable.tsx
juliusv: My old comment still stands: we should make this look like the same buttons on the /new/targets page by reusing the markup and styling from there.

web/ui/react-app/src/App.css
juliusv: This comment still needs to be addressed as well (more constistency with /new/targets page).

web/ui/react-app/src/App.css
```diff
@@ -212,6 +212,19 @@ div.time-input {
...
+    margin-bottom: 1%;
+}
+
+.target-head {
+    font-weight: 700;
+    font-size: large;
+}
+
+.label-component {
+    margin-bottom: 1%;
```
juliusv: I don't see any difference when removing this (only when I use larger percentages), is this style rule needed?

web/ui/react-app/src/pages/targets/ScrapePoolPanel.tsx
```diff
@@ -14,6 +14,19 @@ interface PanelProps {
   targetGroup: ScrapePool;
 }
 
+export const ToggleButtonProps = {
```
boyskila: If you want to share common styles and behavior its better to create a component instead.

web/ui/react-app/src/pages/targets/ScrapePoolPanel.tsx
```diff
@@ -24,26 +37,16 @@ const ScrapePoolPanel: FC<PanelProps> = ({ scrapePool, targetGroup }) => {
...
-      marginLeft: '1em',
-      verticalAlign: 'baseline',
-    },
-  };
+  ToggleButtonProps.onClick = (): void => setOptions({ expanded: !expanded });
+  ToggleButtonProps.children = `show ${expanded ? 'less' : 'more'}`;
 
   return (
     <div className={styles.container}>
       <h3>
```
boyskila: Maybe the whole h3 container can be a component used from Service discovery page and Targets page with possibility to pass title as children - <a> for targets and <strong> for SD.

Harkishen-Singh: sure

juliusv: For now I would prioritize getting this PR done and merged, since it is taking quite a while already, but happy to have more refactorings later.

web/ui/react-app/src/pages/targets/ScrapePoolPanel.tsx
```diff
@@ -24,27 +25,14 @@ const ScrapePoolPanel: FC<PanelProps> = ({ scrapePool, targetGroup }) => {
...
-      fontSize: '0.375em',
-      marginLeft: '1em',
-      verticalAlign: 'baseline',
-    },
-  };
 
   return (
     <div className={styles.container}>
-      <h3>
+      <ToggleMoreLess onClick={(): void => setOptions({ expanded: !expanded })}>
```
juliusv: The targets are all expanded by default. But now the button says "show less" on /new/targets for an expanded section, and vice versa. You'll probably want to pass in `expanded` from the outside and keep the state outside of the component, as @boyskila noted in https://github.com/prometheus/prometheus/commit/d8a460f1a684068c4b8961aa33c1f11c82d66d7b#r36603580.

web/ui/react-app/src/pages/targets/ToggleMoreLess.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { FC, useState } from 'react';
+import { Button } from 'reactstrap';
+
+interface ToggleProps {
```
juliusv: Usually our naming for props is `<component name>Props`, so `ToggleMoreLessProps` here.
```suggestion
interface ToggleMoreLessProps {
```

web/ui/react-app/src/pages/targets/ToggleMoreLess.tsx
```diff
@@ -0,0 +1,32 @@
...
+  onClick(): void;
+}
+
+export const ToggleMoreLess: FC<ToggleProps> = ({ children, onClick }) => {
+  const [showMoreLess, toggleMoreLess] = useState(false);
```
juliusv: As noted above, better to remove the state inside the component and only pass in an `expanded` prop from the outside which can be controlled from the button via the `onClick` function that is already passed in.

juliusv:
:+1: Thanks, looks mostly good now! I'm going to merge, and if we find more stuff to change (like the marked TODO, https://github.com/prometheus/prometheus/pull/6394#issuecomment-569279774, or others), we can always do it as a follow-up :)

Issue #6400, opened 2019-12-03
Pushdown aggregator group by through read hint


garrettlish:
## Proposal
Assume we have third-party remote storage, pull raw metrics from third-party remote storage and aggregate them in prometheus might NOT be efficient. It might worth pushdown predicate (aggregators, like min, max, avg, sum, etc.) along with group by information to third-party remote storage through read hint and reduce the data set returned from remote storage.

The proposal is add `groups` array in `ReadHints` struct, and populating it while walk through the syntax tree in promQL engine.

Let me give you a concrete example:
Assume we have one metric, it has 1,000,000 series. If there is a query like `max by (host) (sample_metric)`, third-party remote storage return 1,000,000 series and find maximum value in prometheus rule engine is inefficient. It might be better to pushdown `max` and `host` group by information to third-party remote storage and return few series (host number) to prometheus rule engine.

I had already seen one PR (https://github.com/prometheus/prometheus/pull/4265) which proposing the same thing, however, this PR doesn't handle properly (sample: it doesn't handle aggregate against binary operator result - `max by  (foo)(a - b)`).

Another proposal is passing hacked filters through promQL (like: `sum(abc{HINTS="{'by':'url|app'}"}) by (url, app)`), however, this is hacky and relies on user's promQL input and we cannot make sure it always get proper group by information.

brian-brazil:
We're happy to add any hints that a remote storage finds useful, as long as there's a correct plausible use case for them. Note that you'll also need to indicate if it's a without or by.

garrettlish:
Cool, thanks @brian-brazil. You are correct, we need to indicate it is without or by.

bwplotka:
I think even Thanos project might be interested in this initiative - this allows some way of distributed querying on leafs (:

garrettlish:
close it since https://github.com/prometheus/prometheus/pull/6401 is merged already

sudeepknair:
nice work @garrettlish . Thanks for this feature addition 👍

PR #6401, opened 2019-12-03
Pushdown aggregator group by through read hint

garrettlish:
Implement https://github.com/prometheus/prometheus/issues/6400


prompb/types.proto
```diff
@@ -56,6 +56,8 @@ message ReadHints {
   string func = 2;    // String representation of surrounding function or aggregation.
   int64 start_ms = 3; // Start time in milliseconds.
   int64 end_ms = 4;   // End time in milliseconds.
+  repeated string groups = 5; // List of label names of any fields used in a group-by aggregation
```
brian-brazil: Full stops. And there's no such thing as a "group-by".

brian-brazil:
Also we should think a bit about counters here I think, as common use cases look like sum(rate) or historgam_quantile(sum(rate())).

@beorn7 this aspect likely is of interest.

prompb/types.proto
garrettlish: sure, will change it to `group_by`

prompb/types.proto
```diff
@@ -56,6 +56,8 @@ message ReadHints {
   string func = 2;    // String representation of surrounding function or aggregation.
   int64 start_ms = 3; // Start time in milliseconds.
   int64 end_ms = 4;   // End time in milliseconds.
+  repeated string group_by = 5; // List of label names of any fields used in a group-by aggregation
```
brian-brazil: grouping is what we use internally, so probably best to stick with that.

prompb/types.pb.go
```diff
@@ -430,6 +432,20 @@ func (m *ReadHints) GetEndMs() int64 {
...
+		return m.GroupBy
+	}
+	return nil
+}
+
+func (m *ReadHints) GetWithout() bool {
+	if m != nil {
+		return m.Without
+	}
+	return false
```
brian-brazil: If missing this would mean `by()`, which is not the same as not having a grouping. True is the right default here.

garrettlish:
@brian-brazil To support `sum(rate)`, besides the spatial aggregation read hint, we might need to introduce a temporal aggregation read hint and time window, something like `ta=rate:5m`. If so, maybe we can extend the PR to support pushdown TA functions as well (such as `<aggregation>_over_time`).

prompb/types.proto
garrettlish: sure

prompb/types.pb.go
garrettlish: will correct it

brian-brazil: Thinking further, we actually want to change the field to By (rather than Without) and default to false. That way a Prometheus that doesn't support this looks the same as one that does support it but wants no grouping.

garrettlish: Yes, agree, I did change it to `By` and default set to false, please see https://github.com/prometheus/prometheus/pull/6401/commits/95cb9b4ba4b627a903f044b4c8c8b090e609371b

prompb/types.proto
brian-brazil: `Grouping` is what we use elsewhere. Also s/group-by//, this isn't SQL.

storage/interface.go
```diff
@@ -71,6 +71,9 @@ type SelectParams struct {
 
 	Step int64  // Query step size in milliseconds.
 	Func string // String representation of surrounding function or aggregation.
+
+	Groups []string // List of label names of any fields used in a group-by aggregation
+	By     bool     // Indicate whether it is without or by
```
beorn7: Nit: Please end comments with periods, like above.

beorn7:
> @beorn7 this aspect likely is of interest.

I guess because of the upcoming histogram with 40,000 buckets? At the moment, I'd think that needs even deeper integration, but we'll see.

brian-brazil:
> I guess because of the upcoming histogram with 40,000 buckets? At the moment, I'd think that needs even deeper integration, but we'll see.

Yes. In the most optimistic case something like this may turn out to be enough, without having to add layering violations to the interfaces themselves.

brian-brazil:
> Also, does this PR supports without () or rather by () ?

Yes, set `By` to true.

bwplotka:
>  > Also, does this PR supports without () or rather by () ?

> Yes, set By to true.

Ah yes all good. And for `without ()` we don't group as expected.

brian-brazil:
> Ah yes all good.

It only took me two goes to figure it out :)

promql/engine_test.go
```diff
@@ -1028,3 +1029,44 @@ func TestSubquerySelector(t *testing.T) {
...
+	by, groups = inspectExpr("max by (dim1) (avg_over_time(metric[5m]))", extractGroupsFromPath)
+	testutil.Assert(t, len(by) == 0 && len(groups) == 0, "Expected 0 - 0 but got %s - %s", nil, len(by), len(groups))
+
+	by, groups = inspectExpr("(max by (dim1) (metric))[5m:1m]", extractGroupsFromPath)
+	testutil.Assert(t, by[0] == true && reflect.DeepEqual(groups[0], expectedGroupBy), "Expected true - %s but got %t - %s", expectedGroupBy, by, groups)
+
+	by, groups = inspectExpr("(max without (dim1) (metric))[5m:1m]", extractGroupsFromPath)
+	testutil.Assert(t, by[0] == false && reflect.DeepEqual(groups[0], expectedGroupBy), "Expected false - %s but got %t - %s", expectedGroupBy, by, groups)
+
+	by, groups = inspectExpr("max by (dim1) (metric) - min by (dim2) (metric2)", extractGroupsFromPath)
```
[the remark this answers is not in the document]
garrettlish: @bwplotka we will only extract inner max function along with dim1 & dim2 grouping into read hint

promql/engine_test.go
```diff
@@ -1028,3 +1029,44 @@ func TestSubquerySelector(t *testing.T) {
 		}
 	}
 }
+
+func inspectExpr(expr string, f func(p []Node) (bool, []string)) ([]bool, [][]string) {
```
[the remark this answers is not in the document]
garrettlish: make sense, we can using `extractGroupsFromPath` directly rather than a parameter, let me also try using `TestParamsSetCorrectly ` to test grouping parameters. Thanks @bwplotka

garrettlish:
@bwplotka 
> How this will work exactly on sum(rate(.... [5m])? E.g What you will return for such select from some remote storage?

This PR supports spatial aggregation only, to support temporal aggregation, I think we need to introduce another temporal aggregation read hint and time window, something like `ta=rate:5m`. 

> I think the problem is that we might have to say stop at some point, like how to make sure to support queries like: sum(sum by(handler) (sum by (handler, instance, method) (metric))) which in theory is the same as sum(metric)?

This PR only handles the inner aggregations, we only pushdown sum function along with handler, instance, method grouping to read hint in your example.

> I think I like the idea of supporting common queries like sum(rate or histogram ones.

This PR doesn't support sum(rate) & histogram. To support them, I think we need to introduce another temporal aggregation read hint and time window, something like `ta=rate:5m`.

> Also, does this PR supports without () or rather by () ? I think we might miss this case. For this, we might need to add maybe Grouping type for in proto (?) to tell if grouping is actually used or not. Do you mind adding test cases for this?

Yes, we do support both without and by, for without unit test, please refer to https://github.com/prometheus/prometheus/pull/6401/files#diff-97f10a261de0bb743875fe9cc3a969d1R367-R373.

garrettlish:
@brian-brazil @bwplotka @beorn7 comments fixed, besides, I also add temporal aggregation pushdown support, now this PR supports the following 3 scenarios but not support subqueries.
1. TA only, such as `rate(range_vector)`, `sum_over_time(range_vector)` - pushdown func name and time window
2. SA only, such as `sum by (dimensions) (instant_vector)` - pushdown func and group by dimensions
3. SA with TA, such `sum by (dimensions) (rate(range_vector))` - pushdown func, group by dimensions and time window

prompb/types.proto
```diff
@@ -56,6 +56,9 @@ message ReadHints {
   string func = 2;    // String representation of surrounding function or aggregation.
   int64 start_ms = 3; // Start time in milliseconds.
   int64 end_ms = 4;   // End time in milliseconds.
+  repeated string grouping = 5; // List of label names of any fields used in aggregation.
+  bool by = 6; // Indicate whether it is without or by, it will be false if grouping is not used.
+  string ta = 7; // Temporal aggregation read hint include func name and time window, such as rate:5m0s.
```
brian-brazil: No need for the function name, we just need the time and it should be in milliseconds like the other. The field name also doesn't go over the wire, so no need to abbreviate it. `range` is what we'd usually call this.

garrettlish: @brian-brazil If we don't include function name, how can we distinguish `sum by (dim) (rate(range_vector))` and `avg by (dim) (rate(range_vector))`? `sum by (dim) (sum_over_time(range_vector))` and `max by (dim) (sum_over_time(range_vector))`?

brian-brazil: The function name is already in `func`

garrettlish: @brian-brazil In this PR, we'd like to support 1) temporal aggregation, 2) spatial aggregation 3) temporal aggregation. For 1) & 2), the `func` will be temporal aggregation func name (such as `rate`, `sum_over_time`, etc.) name or spatial aggregation func name (such as `sum`, `avg`, etc.). However, for 3), the `func` will be spatial aggregation func name, we don't have a place to record the temporal aggregation func name.
For example:
`sum by (dim) (rate(range_vector))`, if we use `rate` as `func`, then we cannot identify spatial aggregation name `sum`.

brian-brazil: We could make func a list, but we're already quite bit of scope creep here so I'd say let's get this in without worrying about 3.

garrettlish: make sense, let me make the change

prompb/types.proto
```diff
@@ -56,6 +56,9 @@ message ReadHints {
   string func = 2;    // String representation of surrounding function or aggregation.
   int64 start_ms = 3; // Start time in milliseconds.
   int64 end_ms = 4;   // End time in milliseconds.
+  repeated string grouping = 5; // List of label names of any fields used in aggregation.
+  bool by = 6; // Indicate whether it is without or by, it will be false if grouping is not used.
+  int64 range_ms = 7; // Temporal aggregation time window in millisecond.
```
brian-brazil: This is a bit wordy, and we don't use these terms in the docs. "Range vector selector range" covers it.

prompb/types.proto
```diff
@@ -56,6 +56,9 @@ message ReadHints {
   string func = 2;    // String representation of surrounding function or aggregation.
   int64 start_ms = 3; // Start time in milliseconds.
   int64 end_ms = 4;   // End time in milliseconds.
+  repeated string grouping = 5; // List of label names of any fields used in aggregation.
+  bool by = 6; // Indicate whether it is without or by, it will be false if grouping is not used.
```
brian-brazil: No need for anything after the comma.

garrettlish: sure

prompb/types.proto
```diff
@@ -56,6 +56,9 @@ message ReadHints {
   string func = 2;    // String representation of surrounding function or aggregation.
   int64 start_ms = 3; // Start time in milliseconds.
   int64 end_ms = 4;   // End time in milliseconds.
+  repeated string grouping = 5; // List of label names of any fields used in aggregation.
```
brian-brazil: s/of any fields/

PromQL doesn't have fields.

prompb/types.proto
```diff
@@ -56,6 +56,9 @@ message ReadHints {
...
+  bool by = 6; // Indicate whether it is without or by.
+  int64 range_ms = 7; // Range vector selector range in millisecond.
```
brian-brazil: milliseconds

prompb/types.proto
garrettlish: What about `List of label names used in aggregation. `?

prompb/types.proto
garrettlish: sorry, let me correct it.

brian-brazil:
Thanks!

bwplotka:
:+1:

sudeepknair:
👍

Issue #6395, opened 2019-11-29
Proposal: Add API endpoint to retrieve metadata by metric name


beorn7:
_Credits and notes: The below has mostly been written by @gotjosh. He's currently on vacation. That's why I (@beorn7) am filing this now. I helped with the draft, and we also got input from @davkal (Grafana), @slrtbtfs ([LSP implementation for PromQL](https://github.com/slrtbtfs/promql-lsp)) and @jkohen and @StevenYCChou ([Stackdriver Prometheus sidecar](https://github.com/Stackdriver/stackdriver-prometheus-sidecar)). Note, however, that the below mostly caters for the Grafana and LSP use case, while the Stackdriver use case needed more invasive changes to provide metadata from targets that are not active anymore. This more or less historical data would also help Grafana, but is out of scope for this proposal. This proposal is a fairly lightweight (and hopefully uncontroversial) addition to serve as an intermediate solution until we get full metadata support, which realistically I only expect in Prometheus v3._

# Context and motivation

At the moment of writing, Prometheus users and integrations (e.g. Grafana) [make use](https://github.com/grafana/grafana/blob/master/public/app/plugins/datasource/prometheus/metric_find_query.ts#L93) of `/api/v1/label/__name__/values` as a way to expose all the metric names that Prometheus has scraped.

Prometheus currently offers a way to obtain metadata (type, help, etc) for a given metric via the `/api/v1/targets/metadata` endpoint. This metadata helps power features such as query hinting, where users are suggested to use `rate` on counters.

Often, integrations are unhappy with the current state of the API endpoint. Both Grafana and the PromQL LSP would like to obtain metadata in a situation where the target is not relevant or even unknown – in which case the endpoint has to be queried for _all_ targets, resulting in a possibly very large payload, blown up by a factor equal to the number of targets exposing the requested metric name (possibly thousands). Without the metadata, integrations have to fall back to  heuristics to determine it e.g. [detect monotonicity](https://github.com/grafana/grafana/blob/master/public/app/plugins/datasource/prometheus/query_hints.ts#L29) to infer it’s a counter.

# Challenges

## `/api/v1/targets/metadata` is per target

[...]

PR #6409, opened 2019-12-05
Add tests for /api/v1/target/metadata

gotjosh:
As a foundation for #6395, let's add test to the existing metadata API.

The only thing that could be mildly controversial is that I've had to make the interface `MetricMetadataStore` public alongside with its methods. 

Without it, we have no way (to the best of my knowledge) to represent data in the metadata cache of a specific target. The current design of the `scrape` package tells us that the cache should be [assigned to a scrape pool](https://github.com/prometheus/prometheus/blob/master/scrape/scrape.go#L230) instead of individual targets.

I tried _hardly_ to circumvent this challenge and avoid changing the interface but that proved futile. All in all, this is the minimal diff I can put together to make this work - there are other alternatives but I get the feeling these are more involved and would require an overhaul of the current testing structure e.g. Using the `scrape/manager` to set the cache for the pool.

Best viewed commit-by-commit.

TODO

- [ ] Do we need unit tests scrapeCache#GetMetadata and scrapeCache#ListMetadata now that these are public?

### DIFF
--- a/web/api/v1/api_test.go
+++ b/web/api/v1/api_test.go
@@ -43,6 +43,7 @@ import (
 	"github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/pkg/gate"
 	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/pkg/textparse"
 	"github.com/prometheus/prometheus/pkg/timestamp"
 	"github.com/prometheus/prometheus/prompb"
 	"github.com/prometheus/prometheus/promql"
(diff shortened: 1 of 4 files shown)


brian-brazil:
> The only thing that could be mildly controversial is that I've had to make the interface MetricMetadataStore public alongside with its methods.

We're generally fine with making things public, even if we don't need that visibility within Prometheus.

web/api/v1/api_test.go
```diff
@@ -55,55 +56,98 @@ import (
...
+		droppedTargets: droppedTargets,
+	}
+}
 
 var (
 	scrapeStart = time.Now().Add(-11 * time.Second)
 )
 
 func (t testTargetRetriever) TargetsActive() map[string][]*scrape.Target {
-	testTarget := scrape.NewTarget(
```
gotjosh: While this change is technically not needed right now. It will in my subsequent pull request as I need to define various targets at different tests cases dynamically.

Would rather get some feedback on it now and avoid a bigger diff later on.

brian-brazil:
Thanks!

PR #6410, opened 2019-12-05
PromQL: Use generated parser to parse label matchers

slrtbtfs:
This PR adds a generated parser to PromQL.

Currently the generated parser only parses label selectors and is invoked by the existing parser, whenever label selectors are encountered while parsing a Vector selector.

The generated parser should produce exactly the same syntax error messages as the existing parser.

This PR would correspond to step two of what is described in https://github.com/prometheus/prometheus/issues/6256#issuecomment-557535813 .

cc @brian-brazil @squat @simonpasquier


promql/generated_parser.y
```diff
@@ -0,0 +1,155 @@
...
+%token <item> LSS
+%token <item> GTE
+%token <item> GTR
+%token <item> EQL_REGEX
+%token <item> NEQ_REGEX
+%token <item> POW
+%token	operatorsEnd
+
+%token	aggregatorsStart
+/* Aggregators.*/
```
brian-brazil: Spaces in comment formatting are inconsistent.

promql/parse.go
```diff
@@ -356,20 +355,29 @@ type yySymType item
...
+		}
+	}
+
+	return int(typ)
 }
 
 // Error is expected by the yyLexer interface of the yacc generated parser.
 //
+// It is a no-op since the parsers error routines are triggered
+// over mechanisms that allow more fine grained control
```
brian-brazil: s/over/by/ ?

promql/parse.go
```diff
@@ -1127,3 +1136,50 @@ func parseDuration(ds string) (time.Duration, error) {
 	}
 	return time.Duration(dur), nil
 }
+
+// parseGenerated invokes the yacc generated parser.
+// The generated parser gets the provided startSymbol injected into
+// the lexer stream, bases on which different grammars can be used.
```
brian-brazil: based on which grammer will be used

promql/parse.go
```diff
@@ -1127,3 +1136,50 @@ func parseDuration(ds string) (time.Duration, error) {
...
+		matchType = labels.MatchEqual
+	case NEQ:
+		matchType = labels.MatchNotEqual
+	case EQL_REGEX:
+		matchType = labels.MatchRegexp
+	case NEQ_REGEX:
+		matchType = labels.MatchNotRegexp
+	default:
+		// This should never happen, since the error should have been chaught
+		// by the generated parser
```
brian-brazil: Full stop

promql/parse.go
```diff
@@ -933,14 +941,18 @@ func (p *parser) offset() time.Duration {
...
-	var matchers []*labels.Matcher
+	ret := &VectorSelector{
+		Name: name,
+	}
 	// Parse label matching if any.
 	if t := p.peek(); t.typ == LEFT_BRACE {
-		matchers = p.labelMatchers(EQL, NEQ, EQL_REGEX, NEQ_REGEX)
+		p.generatedParserResult = ret
+
+		p.parseGenerated(START_LABELS, []ItemType{RIGHT_BRACE, EOF})
```
brian-brazil: Do we have a unittest covering a } inside a label value?

slrtbtfs: You mean something like `foo{bar="}"}` ?

brian-brazil: Yes

slrtbtfs: AFAIK we don't. But if something like that would cause problems it would be a lexer bug.

The parser only sees the token stream:

    IDENTIFIER "foo"
    LEFT_BRACE "{"
    IDENTIFIER "bar"
    EQL "="
    STRING "\"}\""
    RIGHT_BRACE "}"

brian-brazil: It doesn't hurt to have a unittest to be safe.

slrtbtfs: I've added a test case.

brian-brazil:
Thanks!

slrtbtfs:
Didn't expect it to be merged that quickly.

Thanks!

brian-brazil:
We've a lot of unittests, which helps give confidence.

PR #6415, opened 2019-12-06
tsdb: add error details in log.

johncming:
add error details in log.

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -865,7 +865,7 @@ func openBlocks(l log.Logger, dir string, loaded []*Block, chunkPool chunkenc.Po
 	for _, bDir := range bDirs {
 		meta, _, err := readMetaFile(bDir)
 		if err != nil {
-			level.Error(l).Log("msg", "not a block dir", "dir", bDir)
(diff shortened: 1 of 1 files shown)


tsdb/db.go
```diff
@@ -865,7 +865,7 @@ func openBlocks(l log.Logger, dir string, loaded []*Block, chunkPool chunkenc.Po
 	for _, bDir := range bDirs {
 		meta, _, err := readMetaFile(bDir)
 		if err != nil {
-			level.Error(l).Log("msg", "not a block dir", "dir", bDir)
+			level.Error(l).Log("msg", "not a block dir", "dir", bDir, "err", err)
```
bwplotka: Hm.. In fact this can have many different reasons. We already check if it's block dir in https://github.com/prometheus/prometheus/blob/3bb715031f440eb98cd73bc0bd961becf61c855c/tsdb/db.go#L1321:6 

Can we change it to an overall error?  (: 
```suggestion
			level.Error(l).Log("msg", "failed to read meta.json for a block", "dir", bDir, "err", err)
```

johncming: ok, thanks.

bwplotka:
Thanks!

PR #6418, opened 2019-12-06
Reduce memory used by postings offset table.

brian-brazil:
Rather than keeping the offset of each postings list, instead
keep the nth offset of the offset of the posting list. As postings
list offsets have always been sorted, we can then get to the closest
entry before the one we want an iterate forwards.

I haven't done much tuning on the 32 number, it was chosen to try
not to read through more than a 4k page of data.

Add a bulk interface for fetching postings. Use it to avoid having to
re-read parts of the posting offset table when querying lots of it.

For a index with what BenchmarkHeadPostingForMatchers uses RAM
for r.postings drops from 3.79MB to 80.19kB or about 48x.
Bytes allocated go down by 30%, and suprisingly CPU usage drops by
4-6% for typical queries too.

```
benchmark                                                            old ns/op     new ns/op     delta
BenchmarkHeadPostingForMatchers/n="1"-4                              35220         36959         +4.94%
BenchmarkHeadPostingForMatchers/n="1",j="foo"-4                      578798        615273        +6.30%
BenchmarkHeadPostingForMatchers/j="foo",n="1"-4                      581930        614427        +5.58%
BenchmarkHeadPostingForMatchers/n="1",j!="foo"-4                     583190        615479        +5.54%
BenchmarkHeadPostingForMatchers/i=~".*"-4                            115169232     116079317     +0.79%
BenchmarkHeadPostingForMatchers/i=~".+"-4                            155318554     146743642     -5.52%
BenchmarkHeadPostingForMatchers/i=~""-4                              60655399      49071222      -19.10%
BenchmarkHeadPostingForMatchers/i!=""-4                              49437509      39161630      -20.79%
BenchmarkHeadPostingForMatchers/n="1",i=~".*",j="foo"-4              114361825     116101312     +1.52%
BenchmarkHeadPostingForMatchers/n="1",i=~".*",i!="2",j="foo"-4       113697227     114773786     +0.95%
BenchmarkHeadPostingForMatchers/n="1",i!=""-4                        50452920      39238957      -22.23%
BenchmarkHeadPostingForMatchers/n="1",i!="",j="foo"-4                51290788      41008892      -20.05%
BenchmarkHeadPostingForMatchers/n="1",i=~".+",j="foo"-4              156004011     145612748     -6.66%
BenchmarkHeadPostingForMatchers/n="1",i=~"1.+",j="foo"-4             30318091      30444402      +0.42%
BenchmarkHeadPostingForMatchers/n="1",i=~".+",i!="2",j="foo"-4       155504240     148051405     -4.79%
BenchmarkHeadPostingForMatchers/n="1",i=~".+",i!~"2.*",j="foo"-4     185225863     174089236     -6.01%

benchmark                                                            old allocs     new allocs     delta
BenchmarkHeadPostingForMatchers/n="1"-4                              4              5              +25.00%
BenchmarkHeadPostingForMatchers/n="1",j="foo"-4                      7              9              +28.57%
BenchmarkHeadPostingForMatchers/j="foo",n="1"-4                      7              9              +28.57%
BenchmarkHeadPostingForMatchers/n="1",j!="foo"-4                     15             16             +6.67%
BenchmarkHeadPostingForMatchers/i=~".*"-4                            100010         100012         +0.00%
BenchmarkHeadPostingForMatchers/i=~".+"-4                            200073         200047         -0.01%
BenchmarkHeadPostingForMatchers/i=~""-4                              200076         200048         -0.01%
BenchmarkHeadPostingForMatchers/i!=""-4                              200070         200041         -0.01%
BenchmarkHeadPostingForMatchers/n="1",i=~".*",j="foo"-4              100013         100016         +0.00%
BenchmarkHeadPostingForMatchers/n="1",i=~".*",i!="2",j="foo"-4       100017         100020         +0.00%
BenchmarkHeadPostingForMatchers/n="1",i!=""-4                        200073         200045         -0.01%
BenchmarkHeadPostingForMatchers/n="1",i!="",j="foo"-4                200075         200048         -0.01%
BenchmarkHeadPostingForMatchers/n="1",i=~".+",j="foo"-4              200078         200053         -0.01%
BenchmarkHeadPostingForMatchers/n="1",i=~"1.+",j="foo"-4             111166         111149         -0.02%
BenchmarkHeadPostingForMatchers/n="1",i=~".+",i!="2",j="foo"-4       200082         200057         -0.01%
BenchmarkHeadPostingForMatchers/n="1",i=~".+",i!~"2.*",j="foo"-4     311288         311240         -0.02%

benchmark                                                            old bytes     new bytes     delta
BenchmarkHeadPostingForMatchers/n="1"-4                              264           280           +6.06%
BenchmarkHeadPostingForMatchers/n="1",j="foo"-4                      360           392           +8.89%
BenchmarkHeadPostingForMatchers/j="foo",n="1"-4                      360           392           +8.89%
BenchmarkHeadPostingForMatchers/n="1",j!="foo"-4                     520           537           +3.27%
BenchmarkHeadPostingForMatchers/i=~".*"-4                            1600547       1602997       +0.15%
BenchmarkHeadPostingForMatchers/i=~".+"-4                            24901726      17262925      -30.68%
BenchmarkHeadPostingForMatchers/i=~""-4                              24901882      17260825      -30.68%
BenchmarkHeadPostingForMatchers/i!=""-4                              24900778      17260546      -30.68%
BenchmarkHeadPostingForMatchers/n="1",i=~".*",j="foo"-4              1600632       1601616       +0.06%
BenchmarkHeadPostingForMatchers/n="1",i=~".*",i!="2",j="foo"-4       1600803       1600851       +0.00%
BenchmarkHeadPostingForMatchers/n="1",i!=""-4                        24900967      17259722      -30.69%
BenchmarkHeadPostingForMatchers/n="1",i!="",j="foo"-4                24900971      17259819      -30.69%
BenchmarkHeadPostingForMatchers/n="1",i=~".+",j="foo"-4              24901861      17262861      -30.68%
BenchmarkHeadPostingForMatchers/n="1",i=~"1.+",j="foo"-4             3788509       3143975       -17.01%
BenchmarkHeadPostingForMatchers/n="1",i=~".+",i!="2",j="foo"-4       24902019      17261760      -30.68%
BenchmarkHeadPostingForMatchers/n="1",i=~".+",i!~"2.*",j="foo"-4     28694270      20407961      -28.88%
```

The benchmark commit is something that I hacked together to test this, and should not be merged.


brancz:
cc @pracucci

tsdb/head_bench_test.go
```diff
@@ -50,31 +56,38 @@ func BenchmarkHeadStripeSeriesCreateParallel(b *testing.B) {
...
-	defer func() {
-		testutil.Ok(b, h.Close())
-	}()
-
-	var ref uint64
-
-	addSeries := func(l labels.Labels) {
-		ref++
-		h.getOrCreateWithID(ref, l.Hash(), l)
+	// Make entries ~50B in side
```
juliusv: side -> size :)

tsdb/index/index.go
```diff
@@ -716,19 +723,45 @@ func newReader(b ByteSlice, c io.Closer) (*Reader, error) {
...
+		}
+		valueCount += 1
 		return nil
 	}); err != nil {
 		return nil, errors.Wrap(err, "read postings table")
 	}
+	if lastKey != nil {
+		r.postings[lastKey[0]] = append(r.postings[lastKey[0]], postingOffset{value: allocatedSymbols[lastKey[1]], off: lastOff})
+	}
+	// Trim any extra space in the slices.
```
bwplotka: This is due to double allocs from Go? Can we mention this in comment here?

tsdb/index/index.go
```diff
@@ -716,19 +723,45 @@ func newReader(b ByteSlice, c io.Closer) (*Reader, error) {
 		return nil, errors.Wrap(err, "read label index table")
 	}
 
-	r.postings[""] = map[string]uint64{}
-	if err := ReadOffsetTable(r.b, toc.PostingsTable, func(key []string, off uint64) error {
+	var lastKey []string
+	lastOff := 0
+	valueCount := 0
+	if err := ReadOffsetTable(r.b, r.toc.PostingsTable, func(key []string, _ uint64, off int) error {
```
bwplotka: Can we comment, something like:

`Get partial postings offsets. Keep all label name keys, but only 1/32 of all label values in total.

tsdb/index/index.go
```diff
@@ -749,18 +782,21 @@ type Range struct {
 // for all postings lists.
 func (r *Reader) PostingsRanges() (map[labels.Label]Range, error) {
```
bwplotka: As discussed offline, thanks for making this work. AFAIK it's only Thanos that uses this. As discussed about removing it: I would vote for keeping it as otherwise we need to mmap index twice to get mmaped bytes as it's private in reader ):

tsdb/encoding/encoding.go
```diff
@@ -109,8 +109,11 @@ func NewDecbufAt(bs ByteSlice, off int, castagnoliTable *crc32.Table) Decbuf {
 	b = bs.Range(off+4, off+4+l+4)
 	dec := Decbuf{B: b[:len(b)-4]}
 
-	if exp := binary.BigEndian.Uint32(b[len(b)-4:]); dec.Crc32(castagnoliTable) != exp {
-		return Decbuf{E: ErrInvalidChecksum}
+	if castagnoliTable != nil {
```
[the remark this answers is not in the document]
brian-brazil: Otherwise things get about 100x slower.

tsdb/index/index.go
```diff
@@ -716,19 +723,45 @@ func newReader(b ByteSlice, c io.Closer) (*Reader, error) {
...
+	lastOff := 0
+	valueCount := 0
+	if err := ReadOffsetTable(r.b, r.toc.PostingsTable, func(key []string, _ uint64, off int) error {
 		if len(key) != 2 {
 			return errors.Errorf("unexpected key length for posting table %d", len(key))
 		}
 		if _, ok := r.postings[key[0]]; !ok {
-			r.postings[allocatedSymbols[key[0]]] = map[string]uint64{}
+			// Next label name.
+			r.postings[allocatedSymbols[key[0]]] = []postingOffset{}
```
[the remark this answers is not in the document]
brian-brazil: Not required, but it's a previous optimisation I did to save memory.

tsdb/index/index.go
brian-brazil: That's what this comment is for.

tsdb/index/index.go
brian-brazil: Mmaps are cheap compared to what you're then doing with them.

tsdb/index/index.go

brian-brazil:
Okay, that'd cleaned up and in a submittable state. Comments have been addressed.

krasi-georgiev:
Wouldn't have to review before new year, but really interested to see a prombench results for this PR.

brian-brazil:
The benchmarks above cover this, and I'd like to get this in as I've further significant memory improvements I'd like to do which are based on this.

tsdb/index/index.go
```diff
@@ -802,17 +839,18 @@ func ReadSymbols(bs ByteSlice, version int, off int) ([]string, map[uint32]strin
 
 // ReadOffsetTable reads an offset table and at the given position calls f for each
 // found entry. If f returns an error it stops decoding and returns the received error.
-func ReadOffsetTable(bs ByteSlice, off uint64, f func([]string, uint64) error) error {
+func ReadOffsetTable(bs ByteSlice, off uint64, f func([]string, uint64, int) error) error {
 	d := encoding.NewDecbufAt(bs, int(off), castagnoliTable)
+	startLen := d.Len()
 	cnt := d.Be32()
 
 	// The Postings offset table takes only 2 keys per entry (name and value of label),
```
pracucci: This comment should be moved below, right before the `keys := make([]string, 0, 2)`

tsdb/index/index.go
```diff
@@ -921,25 +959,81 @@ func (r *Reader) Series(id uint64, lbls *labels.Labels, chks *[]chunks.Meta) err
...
-		return nil, errors.Wrap(err, "decode postings")
+	for valueIndex < len(values) {
+		value := values[valueIndex]
+
+		i := sort.Search(len(e), func(i int) bool { return e[i].value >= value })
+		if i == len(e) {
+			// We're past the end.
+			break
+		}
+		// Iterate on the offset table.
```
pracucci: I would suggest to rephrase this comment, explaining that either we hit the expected value or we have to start searching from the previous one.

tsdb/index/index.go
```diff
@@ -921,25 +959,81 @@ func (r *Reader) Series(id uint64, lbls *labels.Labels, chks *[]chunks.Meta) err
...
+
+	if len(values) == 0 {
 		return EmptyPostings(), nil
 	}
-	d := encoding.NewDecbufAt(r.b, int(off), castagnoliTable)
-	if d.Err() != nil {
-		return nil, errors.Wrap(d.Err(), "get postings entry")
+
+	res := make([]Postings, 0, len(values))
+	skip := 0
```
pracucci: Could be more clear renaming it to `skipLabelName`?

tsdb/index/index.go
brian-brazil: Done.

I want to remove all the multi-index code as it's unused and presently confusing.

tsdb/index/index.go
brian-brazil: Rearranged&added.

tsdb/index/index.go
brian-brazil: We're not just skipping a label name though. We're also skipping the keycount.

tsdb/index/index.go
```diff
@@ -921,25 +959,82 @@ func (r *Reader) Series(id uint64, lbls *labels.Labels, chks *[]chunks.Meta) err
...
+			// Need to look from previous entry.
+			i--
+		}
+		// Don't Crc32 the entire postings offset table, this is very slow
+		// so hope any issues were caught at startup.
+		d := encoding.NewDecbufAt(r.b, int(r.toc.PostingsTable), nil)
+		d.Skip(e[i].off)
+
+		// Iterate on the offset table.
+		var postingsOff uint64 // The offset into the postings table.
```
bwplotka:
```suggestion
		var postingsOff uint64 // The offset in the postings table.
```

tsdb/index/index.go
brian-brazil: It did however get in my way, as a function with no uses, no unittests, but I still kinda have to keep working. This is why I really really don't like this sort of thing.

tsdb/index/index.go
```diff
@@ -716,19 +723,45 @@ func newReader(b ByteSlice, c io.Closer) (*Reader, error) {
...
+			if lastKey != nil {
+				// Always include last value for each label name.
+				r.postings[lastKey[0]] = append(r.postings[lastKey[0]], postingOffset{value: allocatedSymbols[lastKey[1]], off: lastOff})
+			}
+			lastKey = nil
+			valueCount = 0
 		}
-		r.postings[key[0]][allocatedSymbols[key[1]]] = off
+		// Keep only every nth posting offset, to save memory.
+		if valueCount%32 == 0 {
```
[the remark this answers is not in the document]
brian-brazil: It's largely an arbitrary number, and I don't generally like splitting constants like this to a far distant part of the file.

tsdb/block.go
```diff
@@ -69,14 +69,14 @@ type IndexReader interface {
...
 
-	// LabelValues returns the possible label values.
+	// LabelValues returns sorted possible label values.
 	LabelValues(names ...string) (index.StringTuples, error)
 
 	// Postings returns the postings list iterator for the label pair.
 	// The Postings here contain the offsets to the series inside the index.
-	// Found IDs are not strictly required to point to a valid Series, e.g. during
-	// background garbage collections.
-	Postings(name, value string) (index.Postings, error)
```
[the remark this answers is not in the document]
brian-brazil: Good idea, done.

tsdb/index/index.go
```diff
@@ -594,15 +594,16 @@ type StringTuples interface {
...
+	toc *TOC
 
 	// Close that releases the underlying resources of the byte slice.
 	c io.Closer
 
 	// Cached hashmaps of section offsets.
 	labels map[string]uint64
-	// LabelName to LabelValue to offset map.
-	postings map[string]map[string]uint64
+	// Map of LabelName to a list of some LabelValues's position in the offset table.
```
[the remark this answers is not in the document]

tsdb/index/index.go
brian-brazil: Into is correct here. In would mean inside.

bwplotka: hmmm , ok (:

tsdb/index/index.go
```diff
@@ -921,25 +960,82 @@ func (r *Reader) Series(id uint64, lbls *labels.Labels, chks *[]chunks.Meta) err
...
+		// Discard values before the start.
+		valueIndex++
 	}
-	_, p, err := r.dec.Postings(d.Get())
-	if err != nil {
-		return nil, errors.Wrap(err, "decode postings")
+	for valueIndex < len(values) {
+		value := values[valueIndex]
+
+		i := sort.Search(len(e), func(i int) bool { return e[i].value >= value })
```
pracucci: If I'm not missing anything, postings values are sorted. Given this function picks multiple values in input (and they're sorted as well), the `values[i+1]` is > `values[i]`, which means the position of `values[i+1]` within `e` will be at a position > than the previous one, which means we may reduce the iterations done by `sort.Search()` with something like (pseudo-code):

```
i := sort.Search(len(e) - previousPos, func(i int) bool { return e[i+previousPos].value >= value })
i += previousPos
```

brian-brazil: Yes that's an optimisation I considered, however the code is pretty finicky so I left it out for now.

tsdb/index/index.go
```diff
@@ -716,19 +724,46 @@ func newReader(b ByteSlice, c io.Closer) (*Reader, error) {
...
+		}
+		if valueCount%32 == 0 {
+			r.postings[key[0]] = append(r.postings[key[0]], postingOffset{value: allocatedSymbols[key[1]], off: off})
+			lastKey = nil
+		} else {
+			lastKey = key
+			lastOff = off
 		}
-		r.postings[key[0]][allocatedSymbols[key[1]]] = off
+		valueCount += 1
```
codesome: Nit
```suggestion
		valueCount++
```

tsdb/index/index.go
```diff
@@ -802,17 +840,18 @@ func ReadSymbols(bs ByteSlice, version int, off int) ([]string, map[uint32]strin
...
-	// Hence setting the size to max of both, i.e. 2.
-	keys := make([]string, 0, 2)
 	for d.Err() == nil && d.Len() > 0 && cnt > 0 {
+		offsetPos := startLen - d.Len()
 		keyCount := d.Uvarint()
-		keys = keys[:0]
+		// The Postings offset table takes only 2 keys per entry (name and value of label),
+		// and the LabelIndices offset table takes only 1 key per entry (a label name).
+		// Hence setting the size to max of both, i.e. 2.
+		keys := make([]string, 0, 2)
```
codesome: Why are we not re-using this slice?

tsdb/index/index.go
```diff
@@ -921,25 +960,82 @@ func (r *Reader) Series(id uint64, lbls *labels.Labels, chks *[]chunks.Meta) err
...
+					}
+					res = append(res, p)
+				}
+				valueIndex++
+				if valueIndex == len(values) {
+					break
+				}
+				value = values[valueIndex]
+			}
+			if i+1 == len(e) || value >= e[i+1].value || valueIndex == len(values) {
```
codesome: Just clarifying: `i+1 == len(e)` because it's the last value and there is no value following it, right? If yes, we could exit early.

tsdb/querier.go
```diff
@@ -413,7 +414,12 @@ func postingsForMatcher(ix IndexReader, m *labels.Matcher) (index.Postings, erro
 	if m.Type == labels.MatchRegexp {
 		setMatches := findSetMatches(m.Value)
 		if len(setMatches) > 0 {
-			return postingsForSetMatcher(ix, m.Name, setMatches)
+			sort.Strings(setMatches)
+			it, err := ix.Postings(m.Name, setMatches...)
+			if err != nil {
+				return nil, err
+			}
+			return it, nil
```
codesome: Why cant this be just `return ix.Postings(m.Name, setMatches...)`?

tsdb/index/index.go

tsdb/index/index.go
brian-brazil: No need to, it'll never make it to the heap.

tsdb/index/index.go
brian-brazil: Yes

tsdb/index/index.go
```diff
@@ -716,19 +724,46 @@ func newReader(b ByteSlice, c io.Closer) (*Reader, error) {
 		return nil, errors.Wrap(err, "read label index table")
 	}
 
-	r.postings[""] = map[string]uint64{}
-	if err := ReadOffsetTable(r.b, toc.PostingsTable, func(key []string, off uint64) error {
+	var lastKey []string
+	lastOff := 0
+	valueCount := 0
+	// For the postings offset table we keep every label name but only every nth
+	// label value (plus the last one), to save memory.
```
bwplotka: Plus the first one! (: 
```suggestion
	// label value (plus the first and last one), to save memory.
```

tsdb/querier.go
brian-brazil: Done

tsdb/index/index_test.go
```diff
@@ -238,6 +243,94 @@ func TestIndexRW_Postings(t *testing.T) {
...
+
+	iw, err := NewWriter(fn)
+	testutil.Ok(t, err)
+
+	// Create a label in the index which has 999 values.
+	symbols := map[string]struct{}{}
+	series := []labels.Labels{}
+	for i := 1; i < 1000; i++ {
+		v := fmt.Sprintf("%03d", i)
+		series = append(series, labels.FromStrings("i", v))
```
bwplotka: Before we merge, can we extend this test to have multi labels series? I think it would be more bit safer.

tsdb/index/index.go
bwplotka: .. as the offsets are sorted by label names and values.

tsdb/index/index_test.go
brian-brazil: Done

PR #6419, opened 2019-12-06
Fix yaml formatting for rules

Marusyk:
Small changes to fix yaml formatting for rules

Please review
Thank you in advance


brian-brazil:
What we have appears to be consistent with the rest of the website.

Marusyk:
What about https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/

brian-brazil:
It appears the alerting rules are the odd one out, if you'd like to fix those instead.

PR #6420, opened 2019-12-06
api: provide per metric metadata

gotjosh:
This adds a new endpoint that provides per metric metadata via the V1 API. Similarly to the `api/v1/target/metadata` the metadata is only available for the lifecycle of the attached target. 

It collapses metadata that is equal across all targets, and aggregates under the same metric name the ones that differ.

An example of a payload looks like:

```bash
curl -s http://localhost:9090/api/v1/metadata?limit=2 | jq
{
  "status": "success",
  "data": {
    "cortex_ring_tokens_total": [
      {
        "type": "gauge",
        "help": "Number of tokens in the ring",
        "unit": ""
      }
    ],
    "jaeger_tracer_sampler_updates_total": [
      {
        "type": "counter",
        "help": "Number of times the Sampler succeeded to retrieve and update sampling strategy",
        "unit": ""
      }
    ]
  }
}
```
I don't think there should be anything highly controversial on what I'm doing but I'll refrain from creating the docs until this is approved.

The next PR coming shortly after, will implement the same endpoint for a specific metric. 

Follow up from #6409 
Part of #6256 

TODO:

- [x] update documentation
- [ ] squash commits

### DIFF
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -569,7 +569,7 @@ $ curl http://localhost:9090/api/v1/alerts
 
 ## Querying target metadata
 
-The following endpoint returns metadata about metrics currently scraped by targets.
+The following endpoint returns metadata about metrics currently scraped from targets.
 This is **experimental** and might change in the future.
 
 ```
@@ -650,6 +650,52 @@ curl -G http://localhost:9091/api/v1/targets/metadata \
     },
     // ...
   ]
+}
+```
+
+## Querying metric metadata
+
+It returns metadata about metrics currently scrapped from targets. However, it does not provide any target information.
+This is considered **experimental** and might change in the future.
+
+```
+GET /api/v1/metadata
+```
+
+URL query parameters:
+
+- `limit=<number>`: Maximum number of metrics to return.
+
+The `data` section of the query result consists of an object where each key is a metric name and each value is a list of unique metadata objects, as exposed for that metric name across all targets.
+
+The following example returns two metrics. Note that the metric `http_requests_total` has more than one object in the list. At least one target has a value for `HELP` that do not match with the rest.
+
+```json
+curl -G http://localhost:9090/api/v1/metadata?limit=2
+
+{
+  "status": "success",
+  "data": {
+    "cortex_ring_tokens": [
+      {
+        "type": "gauge",
+        "help": "Number of tokens in the ring",
+        "unit": ""
+      }
+    ],
+    "http_requests_total": [
+      {
+        "type": "counter",
+        "help": "Number of HTTP requests",
+        "unit": ""
+      },
+      {
+        "type": "counter",
+        "help": "Amount of HTTP requests",
+        "unit": ""
+      }
+    ]
+  }
 }
 ```
 
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -276,6 +276,8 @@ func (api *API) Register(r *route.Router) {
 	r.Get("/targets/metadata", wrap(api.targetMetadata))
 	r.Get("/alertmanagers", wrap(api.alertmanagers))
 
+	r.Get("/metadata", wrap(api.metricMetadata))
+
 	r.Get("/status/config", wrap(api.serveConfig))
 	r.Get("/status/runtimeinfo", wrap(api.serveRuntimeInfo))
 	r.Get("/status/buildinfo", wrap(api.serveBuildInfo))
@@ -688,7 +690,7 @@ func (api *API) targetMetadata(r *http.Request) apiFuncResult {
 
 	metric := r.FormValue("metric")
 
-	var res []metricMetadata
+	res := []metricMetadata{}
 	for _, tt := range api.targetRetriever.TargetsActive() {
 		for _, t := range tt {
 			if limit >= 0 && len(res) >= limit {
@@ -722,9 +724,7 @@ func (api *API) targetMetadata(r *http.Request) apiFuncResult {
 			}
 		}
 	}
-	if len(res) == 0 {
-		return apiFuncResult{nil, &apiError{errorNotFound, errors.New("specified metadata not found")}, nil, nil}
-	}
+
 	return apiFuncResult{res, nil, nil, nil}
 }
 
@@ -803,6 +803,69 @@ func rulesAlertsToAPIAlerts(rulesAlerts []*rules.Alert) []*Alert {
 	}
 
 	return apiAlerts
+}
+
+type metadataSet map[metadata]struct{}
+
+func (ms metadataSet) MarshalJSON() ([]byte, error) {
+	json := jsoniter.ConfigCompatibleWithStandardLibrary
+	m := []metadata{}
+
+	for s := range ms {
+		m = append(m, s)
+	}
+
+	return json.Marshal(m)
+}
+
+type metadata struct {
+	Type textparse.MetricType `json:"type"`
+	Help string               `json:"help"`
+	Unit string               `json:"unit"`
+}
+
+func (api *API) metricMetadata(r *http.Request) apiFuncResult {
+	metrics := map[string]metadataSet{}
+
+	limit := -1
+	if s := r.FormValue("limit"); s != "" {
+		var err error
+		if limit, err = strconv.Atoi(s); err != nil {
+			return apiFuncResult{nil, &apiError{errorBadData, errors.New("limit must be a number")}, nil, nil}
+		}
+	}
+
+	for _, tt := range api.targetRetriever.TargetsActive() {
+		for _, t := range tt {
+			for _, mm := range t.MetadataList() {
+				m := metadata{Type: mm.Type, Help: mm.Help, Unit: mm.Unit}
+				ms, ok := metrics[mm.Metric]
+
+				if !ok {
+					ms = metadataSet{}
+					metrics[mm.Metric] = ms
+				}
+
+				ms[m] = struct{}{}
+			}
+		}
+	}
+
+	if limit >= 0 && len(metrics) >= limit {
+		lm := make(map[string]metadataSet, limit)
+
+		for k, v := range metrics {
+			if len(lm) == limit {
+				break
+			}
+
+			lm[k] = v
+		}
+
+		metrics = lm
+	}
+
+	return apiFuncResult{metrics, nil, nil, nil}
 }
 
 // RuleDiscovery has info for all rules
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -805,19 +805,6 @@ func rulesAlertsToAPIAlerts(rulesAlerts []*rules.Alert) []*Alert {
 	return apiAlerts
 }
 
-type metadataSet map[metadata]struct{}
-
-func (ms metadataSet) MarshalJSON() ([]byte, error) {
-	json := jsoniter.ConfigCompatibleWithStandardLibrary
-	m := []metadata{}
-
-	for s := range ms {
-		m = append(m, s)
-	}
-
-	return json.Marshal(m)
-}
-
 type metadata struct {
 	Type textparse.MetricType `json:"type"`
 	Help string               `json:"help"`
@@ -825,7 +812,7 @@ type metadata struct {
 }
 
 func (api *API) metricMetadata(r *http.Request) apiFuncResult {
-	metrics := map[string]metadataSet{}
+	metrics := map[string]map[metadata]struct{}{}
 
 	limit := -1
 	if s := r.FormValue("limit"); s != "" {
@@ -842,7 +829,7 @@ func (api *API) metricMetadata(r *http.Request) apiFuncResult {
 				ms, ok := metrics[mm.Metric]
 
 				if !ok {
-					ms = metadataSet{}
+					ms = map[metadata]struct{}{}
 					metrics[mm.Metric] = ms
 				}
 
@@ -851,21 +838,23 @@ func (api *API) metricMetadata(r *http.Request) apiFuncResult {
 		}
 	}
 
-	if limit >= 0 && len(metrics) >= limit {
-		lm := make(map[string]metadataSet, limit)
+	res := map[string][]metadata{}
 
-		for k, v := range metrics {
-			if len(lm) == limit {
-				break
-			}
-
-			lm[k] = v
+	for name, set := range metrics {
+		if limit >= 0 && len(res) >= limit {
+			break
 		}
 
-		metrics = lm
+		s := []metadata{}
+
+		for metadata := range set {
+			s = append(s, metadata)
+		}
+
+		res[name] = s
 	}
 
-	return apiFuncResult{metrics, nil, nil, nil}
+	return apiFuncResult{res, nil, nil, nil}
 }
 
 // RuleDiscovery has info for all rules
--- a/web/api/v1/api_test.go
+++ b/web/api/v1/api_test.go
@@ -525,13 +525,14 @@ func testEndpoints(t *testing.T, api *API, tr *testTargetRetriever, testLabelAPI
 	}
 
 	type test struct {
-		endpoint apiFunc
-		params   map[string]string
-		query    url.Values
-		response interface{}
-		errType  errorType
-		sorter   func(interface{})
-		metadata []targetMetadata
+		endpoint    apiFunc
+		params      map[string]string
+		query       url.Values
+		response    interface{}
+		responseLen int
+		errType     errorType
+		sorter      func(interface{})
+		metadata    []targetMetadata
 	}
 
 	var tests = []test{
@@ -1178,7 +1179,7 @@ func testEndpoints(t *testing.T, api *API, tr *testTargetRetriever, testLabelAPI
 		{
 			endpoint: api.metricMetadata,
 			query: url.Values{
-				"limit": []string{"1"},
+				"limit": []string{"2"},
 			},
 			metadata: []targetMetadata{
 				{
@@ -1210,9 +1211,7 @@ func testEndpoints(t *testing.T, api *API, tr *testTargetRetriever, testLabelAPI
 					},
 				},
 			},
-			response: map[string]metadataSet{
-				"go_threads": metadataSet{{textparse.MetricTypeGauge, "Number of OS threads created", ""}: struct{}{}},
-			},
+			responseLen: 2,
 		},
 		// With no available metadata
 		{
@@ -1359,7 +1358,11 @@ func testEndpoints(t *testing.T, api *API, tr *testTargetRetriever, testLabelAPI
 				test.sorter(res.data)
 			}
 
-			assertAPIResponse(t, res.data, test.response)
+			if test.responseLen != 0 {
+				assertAPIResponseLength(t, res.data, test.responseLen)
+			} else {
+				assertAPIResponse(t, res.data, test.response)
+			}
 		}
 	}
 }
@@ -1399,6 +1402,19 @@ func assertAPIResponse(t *testing.T, got interface{}, exp interface{}) {
 			"Response does not match, expected:\n%+v\ngot:\n%+v",
 			string(expectedRespJSON),
 			string(respJSON),
+		)
+	}
+}
+
+func assertAPIResponseLength(t *testing.T, got interface{}, expLen int) {
+	t.Helper()
+
+	gotLen := reflect.ValueOf(got).Len()
+	if gotLen != expLen {
+		t.Fatalf(
+			"Response length does not match, expected:\n%d\ngot:\n%d",
+			expLen,
+			gotLen,
 		)
 	}
 }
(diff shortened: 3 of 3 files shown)


web/api/v1/api.go
```diff
@@ -805,6 +808,93 @@ func rulesAlertsToAPIAlerts(rulesAlerts []*rules.Alert) []*Alert {
 	return apiAlerts
 }
 
+type metricsWithMetadata map[string]metadataSet
+
+type metadataSet map[metadata]struct{}
+
+func (ms metadataSet) MarshalJSON() ([]byte, error) {
```
brian-brazil: You shouldn't need any custom marshalling code, and this doesn't appear to be doing anything non-standard.

web/api/v1/api.go
```diff
@@ -805,6 +808,93 @@ func rulesAlertsToAPIAlerts(rulesAlerts []*rules.Alert) []*Alert {
...
+
+	limit := -1
+	if s := r.FormValue("limit"); s != "" {
+		var err error
+		if limit, err = strconv.Atoi(s); err != nil {
+			return apiFuncResult{nil, &apiError{errorBadData, errors.New("limit must be a number")}, nil, nil}
+		}
+	}
+
+	metrics = make(metricsWithMetadata)
```
brian-brazil: Why not this this all in one line? Why does this type even need to live beyond this function?

web/api/v1/api.go
```diff
@@ -805,6 +808,93 @@ func rulesAlertsToAPIAlerts(rulesAlerts []*rules.Alert) []*Alert {
...
+		for _, t := range tt {
+
+		list:
+			for _, mm := range t.MetadataList() {
+				_, ok := metrics[mm.Metric]
+				m := metadata{Type: mm.Type, Help: mm.Help, Unit: mm.Unit}
+
+				// If the metric name is not in the map, we don't have any metadata yet.
+				if !ok {
+					metrics[mm.Metric] = make(map[metadata]struct{})
```
brian-brazil: Usually we don't use make if we're not using the length parameter

web/api/v1/api.go
```diff
@@ -805,6 +808,93 @@ func rulesAlertsToAPIAlerts(rulesAlerts []*rules.Alert) []*Alert {
...
+					metrics[mm.Metric][m] = struct{}{}
+					continue
+				}
+
+				// If the metric name is in the map, find if the metadata is already in the set.
+				// This collapses the same metadata from different targets.
+				_, ok = metrics[mm.Metric][m]
+				if ok {
+					continue list
```
brian-brazil: You're not in a nested for, no need for the label.

web/api/v1/api.go
```diff
@@ -805,6 +808,93 @@ func rulesAlertsToAPIAlerts(rulesAlerts []*rules.Alert) []*Alert {
...
+				}
+
+				// By now, we know the metadata is not in the set - add it.
+				metrics[mm.Metric][m] = struct{}{}
+			}
+		}
+	}
+
+	if len(metrics) == 0 {
+		return apiFuncResult{nil, &apiError{errorNotFound, errors.New("No metadata available")}, nil, nil}
```
brian-brazil: This is not an error.

web/api/v1/api.go
```diff
@@ -805,6 +808,93 @@ func rulesAlertsToAPIAlerts(rulesAlerts []*rules.Alert) []*Alert {
...
+	}
+
+	if limit >= 0 && len(metrics) >= limit {
+		lm := make(metricsWithMetadata, limit)
+
+		for k, v := range metrics {
```
brian-brazil: You can do `metrics[:limit]`

web/api/v1/api_test.go
```diff
@@ -128,15 +128,15 @@ var (
 	scrapeStart = time.Now().Add(-11 * time.Second)
 )
 
-func (t testTargetRetriever) TargetsActive() map[string][]*scrape.Target {
+func (t *testTargetRetriever) TargetsActive() map[string][]*scrape.Target {
```
brian-brazil: Why does this need a pointer receiver?

web/api/v1/api.go
gotjosh: Completely agree. We treat it as an error in `targetMetadata` too, should I correct both?

brian-brazil: Yes.

web/api/v1/api.go
gotjosh: Metrics is actually a map, if my understanding is correct - I can't slice a map?

web/api/v1/api_test.go
gotjosh: You're 100% correct - I don't need it for Active/Dropped. Thanks!

web/api/v1/api.go
gotjosh: Good catch! Originally I have had a different runtime complexity, no longer needed!

web/api/v1/api.go
gotjosh: Without it, our JSON encoding library complains about it with `unsupported map key type: v1.metadata`

web/api/v1/api.go
gotjosh: Done!

web/api/v1/api.go
brian-brazil: Ah, indeed.

web/api/v1/api.go
brian-brazil: Then use a string as the key.

gotjosh: Sorry if this not a great question, but wouldn't I still need to have custom marshalling code to define how the string should be output?

brian-brazil: No, if you think you need that then it is your data structure which needs improvement.

docs/querying/api.md
```diff
@@ -653,6 +653,52 @@ curl -G http://localhost:9091/api/v1/targets/metadata \
...
+
+The following example returns two metrics, notice that the metric `http_requests_total` has a more than one object in the list. At least one target had a value for `HELP` that did not match with the rest.
+
+```json
+curl -G http://localhost:9090/api/v1/metadata?limit=2
+
+{
+  "status": "success",
+  "data": {
+    "cortex_ring_tokens_total": [
```
brian-brazil: Gauges shouldn't end in _total

docs/querying/api.md
```diff
@@ -653,6 +653,52 @@ curl -G http://localhost:9091/api/v1/targets/metadata \
...
+
+```
+GET /api/v1/metadata
+```
+
+URL query parameters:
+
+- `limit=<number>`: Maximum number of metrics to return.
+
+The `data` section of the query result consists of an object where the keys represent a metric name and the values are a list of the non-matching metadata found on the scraped targets. The metadata that is equal across all targets is collapsed, and aggregates under the same metric name the ones that differ.
```
beorn7: This is a bit verbose, but also confusing. How about the following:

"The `data` section of the query result consists of an object where each key is a metric name and each value is a list of unique metadata objects, as exposed for that metric name across all targets."

docs/querying/api.md
```diff
@@ -653,6 +653,52 @@ curl -G http://localhost:9091/api/v1/targets/metadata \
...
+
+The following example returns two metrics, notice that the metric `http_requests_total` has a more than one object in the list. At least one target had a value for `HELP` that did not match with the rest.
```
beorn7: "…metrics, notice…" → …metrics. Note…"

I would not use the past tense in the 2nd sentence. (This API is only returning the metadata from active targets.)

web/api/v1/api.go
beorn7: This still uses `make`.

Just write `metrics[mm.Metric] = map[metadata]struct{}{}`.

web/api/v1/api.go
```diff
@@ -805,6 +806,84 @@ func rulesAlertsToAPIAlerts(rulesAlerts []*rules.Alert) []*Alert {
...
+
+				// If the metric name is in the map, find if the metadata is already in the set.
+				// This collapses the same metadata from different targets.
+				_, ok = metrics[mm.Metric][m]
+				if ok {
+					continue
+				}
+
+				// By now, we know the metadata is not in the set - add it.
+				metrics[mm.Metric][m] = struct{}{}
```
beorn7: You can write the whole block in a more compact way:

```Go
m := metadata{Type: mm.Type, Help: mm.Help, Unit: mm.Unit}
ms, ok := metrics[mm.Metric]
if !ok { // No metadataSet yet. Create it.
    ms = metadataSet{}
    metrics[mm.Metric] = ms
}
ms[m] = struct{}{}
```

Putting things into the metadataSet is idempotent and cheap. Just do it. Checking first if it is already there and then put it in only if needed is likely more expensive. And a lot more code.

web/api/v1/api.go
beorn7: Sooo… the `metadataSet` has to be the way it is to de-duplicate the metadata, see L847 and following below. There is no way it can be "improved". Go has no native sets, so we use a map with the metadata struct as a key and "nothing" (i.e. an empty struct) as a value. So far so good.

Obviously, we cannot just marshal that `metadataSet`, as the complex key is not compatible with JSON. Plus, we don't even want that representation. We want a JSON array.

Now there are two possibilities:
1. Take the whole, possible very large `metadataSet` and copy it into a newly created `[]metadata`. That would naturally marshal into exactly what we want.
2. Create a custom `MarshalJSON` to teach the `metadataSet` to marshal into what we want.

(1) is probably a bit less code but also less efficient. (2) is the most efficient solution. I'd intuitively prefer (2) but I don't feel strongly. @gotjosh implemented (2). I am inclined to follow that path. Not sure, though, if the code of `MarshalJSON` could be improved? I'm not an expert in writing those methods.

beorn7: I wrote "possible very large `metadataSet`". That's wrong. Most of them will actually only have one element. The concern is that there are possibly very many `metadataSet`s, and we have to convert every single one of them into a slice.

web/api/v1/api.go
```diff
@@ -805,6 +808,93 @@ func rulesAlertsToAPIAlerts(rulesAlerts []*rules.Alert) []*Alert {
...
+type metadataSet map[metadata]struct{}
+
+func (ms metadataSet) MarshalJSON() ([]byte, error) {
+	var res bytes.Buffer
+	json := jsoniter.ConfigCompatibleWithStandardLibrary
+	encoder := json.NewEncoder(&res)
+
+	res.WriteByte('[')
+	for metadata := range ms {
+		if err := encoder.Encode(metadata); err != nil {
```
beorn7: How does this manage commas? (In the tests, it appears to work, but I don't understand why.)

web/api/v1/api.go
brian-brazil: My concern is that we should never have to hand-write JSON marshalling code, it introduces unnecessary risk and makes things more difficult to maintain. The only existing one is due to it's extreme performance criticality.

If you need to convert a map to a list then you should do so before it hits JSON-related code, not inside the JSON code.

gotjosh: I'll amend it @brian-brazil.

docs/querying/api.md
gotjosh: The targets are indeed active but the action (scrape) is in the past, considering we only output data that was scraped beforehand.

I'll correct it as I still think it reads better even if is not _as_ accurate.

docs/querying/api.md
```diff
@@ -653,6 +653,52 @@ curl -G http://localhost:9091/api/v1/targets/metadata \
 }
 ```
 
+## Querying metric metadata
+
+Similarly to the previous endpoint `api/v1/targets/metadata`, it returns metadata about metrics currently scrapped by targets. However, it does not provide any target information.
```
brian-brazil: Let's not refer to other endpoints and their ordering on this page, as that'll change over time.

docs/querying/api.md
```diff
@@ -653,6 +653,52 @@ curl -G http://localhost:9091/api/v1/targets/metadata \
...
+GET /api/v1/metadata
+```
+
+URL query parameters:
+
+- `limit=<number>`: Maximum number of metrics to return.
+
+The `data` section of the query result consists of an object where each key is a metric name and each value is a list of unique metadata objects, as exposed for that metric name across all targets.
+
+The following example returns two metrics. Note that the metric `http_requests_total` has more than one object in the list. At least one target has a value for `HELP` that not match with the rest.
```
brian-brazil: that did not

web/api/v1/api.go
brian-brazil: You can move this into metricMetadata.

web/api/v1/api.go
```diff
@@ -805,6 +805,71 @@ func rulesAlertsToAPIAlerts(rulesAlerts []*rules.Alert) []*Alert {
...
+		if limit, err = strconv.Atoi(s); err != nil {
+			return apiFuncResult{nil, &apiError{errorBadData, errors.New("limit must be a number")}, nil, nil}
+		}
+	}
+
+	for _, tt := range api.targetRetriever.TargetsActive() {
+		for _, t := range tt {
+			for _, mm := range t.MetadataList() {
+				m := metadata{Type: mm.Type, Help: mm.Help, Unit: mm.Unit}
+				ms, ok := metrics[mm.Metric]
```
brian-brazil: It's odd to have the if for this many lines down, and the comment isn't really needed as this is a standard pattern.

docs/querying/api.md
gotjosh: I wanted to avoid confusion a bit as including this one, we now have three sections that refer to metadata. `Querying metadata`, `Querying target metadata`, and `Querying metric metadata`.

The intention was to hint the reader ahead of time what are some of the differences - I'll make the change.

web/api/v1/api.go
gotjosh: Done!

web/api/v1/api.go
gotjosh: Done.

web/api/v1/api.go
gotjosh: If my understanding of the [encoding library](https://github.com/golang/go/blob/64c9ee98b7684cf2156f620cbab4dbb6081b9771/src/encoding/json/encode.go#L729-L765) is correct, Go's JSON encoder is aware of the current state of the buffer. I know we're using `jsoniter` instead of the standard library but they claim  is 100% compatible so the heuristic should still be valid.

docs/querying/api.md
gotjosh: Sorry, not sure I follow the comment. I can't spot the mistake 🤦‍♂.

beorn7: Negations have to happen with "do not" in English instead of just adding a "not" to the full verb like in all other languages. :o)

"that not match with the rest." has to be "that do not match with the rest."

gotjosh: 🤦‍♂ , sorry - I read the text over 10 times and kept adding that do in my mind all time.

web/api/v1/api.go
beorn7: I actually like the state right now most, i.e. have an explicit `MarshalJSON` but it doesn't contain any low-level JSON code generation but only the conversion to a slice.

If we did this without a `MarshalJSON`, we had to create an entire new `metricsForMarshaling := map[string][]metadata{}` and populate it in two nested `for` loops. That's a lot of logic in `metricMetadata`, and it is only about marshaling. That's a layering violation, to use one of @brian-brazil's favorite words.

The current state of this PR gets the best of both worlds: marshaling is well contained, and no hand-written JSON code generation is happening.

docs/querying/api.md
```diff
@@ -653,6 +653,52 @@ curl -G http://localhost:9091/api/v1/targets/metadata \
 }
 ```
 
+## Querying metric metadata
+
+It returns metadata about metrics currently scrapped by targets. However, it does not provide any target information.
```
brian-brazil: scraped from

gotjosh:
The current failure is actually expected, I'm working on making the `limit` test deterministic.

web/api/v1/api.go
brian-brazil: That wouldn't be a layering violation, as you're not reaching into the internals of the JSON-related code or making inappropriate assumptions about how other layers work. It's the same way all standard JSON marshalling works.

However this code does contain layering violation, as it's using jsoniter which means it's depending on a particular JSON marshaller which is a bit of a smell.

I still want to avoid any use of MarshalJSON where it's not absolutely necessary. For example here the function is returning a map of maps, yet somehow the API returns a map of arrays. This is unneeded magic.

> populate it in two nested for loops.

I don't see a problem with making things explicit.

beorn7: The problem here is that what we want is a map of sets. Neither Go nor JSON have that data structure. So Go uses a map of maps while JSON uses a map of arrays.

Brokering between those two different compromises is a marshaling question.

To layer it properly, we needed it in a different function anyway. It could be a separate function called `convertToMapOfSlicesForJSONMarshaling` (slightly exaggerated to make my point), or it could be the `MarshalJSON` method on the Go data structure. The latter makes way more sense to me. It also makes more sense to the author of this PR. You like it differently, fine. It's two to one at this point, and I don't think this is worth an even longer discussion. I'd like to ask you to accept the "majority opinion".

Usage of jsoniter is indeed a smell. It would be better if we could avoid it here. But it is also used throughout this file at various places. It doesn't tip the balance for me to not prefer a separate conversion over a `MarshalJSON` method.

brian-brazil: > To layer it properly, we needed it in a different function anyway.

Not really, it's only a few lines of code. But if a different function ends up containing those lines of code I can live with that.

> convertToMapOfSlicesForJSONMarshaling

Who said anything about JSON? The current endpoint functions don't known that JSON is involved, it could just as well be going to YAML or XML. That we wish to flatten a map into its values isn't related to the marshalling or JSON themselves.

> But it is also used throughout this file at various places.

Only for overall marshalling, not inside any of the functions for specific endpoints. There is currently a separation of concerns, which this PR would remove.

beorn7: As the name says, `MarshalJSON` is specific to JSON marshaling. And the method is not specific to an endpoint but to the data structure. We are essentially doing the same with `promql.Point`, just that this data structure is not from this package, so adding a method to it would be inappropriate.

Converting to a map of slices rather than a map of sets _is_ specific to JSON (as JSON has no sets).

If I had infinite time, I'd happily continue our little argument, but we are way beyond diminishing returns here. I suggest once more to merge this as is.

brian-brazil: > As the name says, MarshalJSON is specific to JSON marshaling.

That's my argument, you'd need to do the exact same thing if it were YAML or most other formats. Thus why this is a smell.

If it's inside the function endpoint then it'd work for everything.

> We are essentially doing the same with promql.Point, just that this data structure is not from this package, so adding a method to it would be inappropriate.

Not quite, and that's on an extremely important hot path (query/query_range) that I couldn't do a breaking change to when I was optimising that. That's one of the reasons I don't want any MarshallJSON, so that this doesn't end up needing more handcoded marshalling in the future. Nor do I want a bad pattern being introduced that other things then follow.

Using MarshallJSON is possibly less performant too, compared to letting jsoniter just handle all of this in the usual way.

>  I suggest once more to merge this as is.

I don't believe that code health concerns should be ignored because consensus hasn't been found. I think a strong argument is needed to introduce an inconsistent pattern which also has been known to be problematic in the past, and the arguments presented thus far do not approach that bar.

gotjosh: > Who said anything about JSON? The current endpoint functions don't know that JSON is involved, it could just as well be going to YAML or XML.

Think this makes a great point. The current functions don't know anything about how the output should be rendered. The introduction of `MarshalJSON` breaks this pattern.

I'll amend it.

gotjosh: The last commit contains the changes made. Can I get (hopefully) one last look? 🙏

brian-brazil:
I've been sitting waiting on the tests to pass for 30 minutes ;)

gotjosh:
Thanks @brian-brazil! 🎉

web/api/v1/api.go
```diff
@@ -805,6 +805,58 @@ func rulesAlertsToAPIAlerts(rulesAlerts []*rules.Alert) []*Alert {
...
+					ms = map[metadata]struct{}{}
+					metrics[mm.Metric] = ms
+				}
+
+				ms[m] = struct{}{}
+			}
+		}
+	}
+
+	res := map[string][]metadata{}
```
beorn7: I'd add a comment that this is needed for marshaling.

gotjosh: I'll add it in the next one which should be up soon.

beorn7:
@brian-brazil after we had an explicit disagreement here, it would have been much nicer to wait for my response instead of just merging.

brian-brazil:
I incorrectly figured that you'd want to get it in, rather than dragging it on further.

There's also some missing periods on the ends of comments in the tests added here.

beorn7:
You correctly figured that I don't want to drag it on further. But just waiting for my "OK then" would not have hurt.

PR #6424, opened 2019-12-08
add flush cmd in tsdbcli

yeya24:

Since readonly db already supported flush https://github.com/prometheus/prometheus/pull/6006, it is nice to have this feature in tsdbcli.


brian-brazil:
You can already access this by taking a snapshot.

yeya24:
Oh, I miss that. But I guess these two are still a little different? Flush only creates a new block from the WAL but snapshot API also links the existing blocks.

krasi-georgiev:
@yeya24 Could you describe the use case that requires this?

yeya24:
@krasi-georgiev  This would be useful to https://github.com/thanos-io/thanos/issues/1849, just creating a block without losing the data in WAL.

krasi-georgiev:
if thas is the case why implement in the tsdb cli and not in the thanos sidecar?

PR #6426, opened 2019-12-09
Match modifier

alburthoffman:
### Proposal
this PR is to provide a match modifier like bool in comparison operators for graphing.

the promql is like: `metrics{} > match 10`. This will return the whole series which have samples larger than 10. So it actually match all time series instead of filtering data points.

by introducing this, we can easily draw matched time series with a threshold line:
image


alburthoffman:
At present it's impossible to do this in one prom query.

brian-brazil:
> This will return the whole series

There's no such things as whole series in PromQL, just range vectors. And an operation on an instant vector changing that into a range vector would violate orthogonality principles.

Per the mailing list, https://www.robustperception.io/graph-top-n-time-series-in-grafana is what you want.

alburthoffman:
this works when using query_range, it is expected to return range vectors.

when using with instant vector, it is as same as doing filters.

Neither topk or bottomk wont help in this case actually.

roidelapluie:
Quoting @brian-brazil in the mailing list:
Try https://www.robustperception.io/graph-top-n-time-series-in-grafana with **max_over_time**(metric[${__range_s}s]) > 10

roidelapluie:
This pull request would violate Prometheus principle that range queries are syntactic sugar over instant queries, as the result would be dependant of the query range.

alburthoffman:
`the result would be dependent of the query range.`

This is true. But it does not violate Prometheus principle, because syntactic sugar does not mean simple, does not mean we only can call step function without any state shared.

It will delegate to the instant query, and do matches based on the results from the instant query.

It is special, because it sounds like function or aggregate, but not the same.

When it comes to graph for alerts, it is very useful because we will be able to show matched lines with threshold. Even only one data point, people are very easy to get what's alerting and its trend.

no other ways can support this except another project called "extended promQL ". but it is more complex than this.

roidelapluie:
> `the result would be dependent of the query range.`
> 
> This is true. But it does not violate Prometheus principle, because syntactic sugar does not mean simple, does not mean we only can call step function without any state shared.

This is exactly what it means.

If that was implemented in some way in instant queries, it would **additionally** violate another core principle: queries don't lookahead in the future.

> It is special, because it sounds like function or aggregate, but not the same.

> When it comes to graph for alerts, it is very useful because we will be able to show matched lines with threshold. Even only one data point, people are very easy to get what's alerting and its trend.

https://www.robustperception.io/graph-top-n-time-series-in-grafana is what you want. I am the one who actually implemented `${__range_s}` in grafana for doing exactly things like this.

> no other ways can support this except another project called "extended promQL ". but it is more complex than this.

At this point you might want to go back to the developers mailing list if you need more clarifications, but it is unlikely that you will get a different answer.

alburthoffman:
it requires a prebuilt dashboard to use the variable `${__range_s}`.  
It is not suitable for all alerts. 
Or can it be support in grafana explorer?

roidelapluie:
It looks like you are missing a feature in Grafana here, in that case their issue tracker might me more suitable.

brian-brazil:
We looked at this in the bug scrub, and per the above this doesn't make sense in terms of PromQL so we're closing it. If you've further questions you can ask on the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users) rather than in a GitHub issue.

PR #6432, opened 2019-12-09
PromQL: Parse label sets using the generated parser

slrtbtfs:
This PR extends the generated part of the PromQL parser to parse label sets.

This allows removing some unneeded parts of the old parser.

### DIFF
--- a/promql/generated_parser.y
+++ b/promql/generated_parser.y
@@ -125,17 +125,10 @@ start           : START_LABELS label_matchers
                      {yylex.(*parser).generatedParserResult.(*VectorSelector).LabelMatchers = $2}
                 | START_LABEL_SET label_set
                      { yylex.(*parser).generatedParserResult = $2 }
-                | error 
-                        { yylex.(*parser).errorf("unknown syntax error after parsing %v", yylex.(*parser).token.desc()) }
+                | error /* If none of the more detailed error messages are triggered, we fall back to this. */
+                        { yylex.(*parser).errorf("unexpected %v", yylex.(*parser).token.desc()) }
                 ;
 
-
-label_match_list:
-                label_match_list COMMA label_matcher
-                        { $$ = append($1, $3)}
-                | label_matcher
-                        { $$ = []*labels.Matcher{$1}}
-                ;
 
 label_matchers  : 
                 LEFT_BRACE label_match_list RIGHT_BRACE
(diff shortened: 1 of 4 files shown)


promql/generated_parser.y
```diff
@@ -151,5 +161,30 @@ match_op        :
...
 
+label_set       :
+                LEFT_BRACE label_set_list RIGHT_BRACE
+                        { $$ = labels.New($2...) }
+                | LEFT_BRACE RIGHT_BRACE
+                        { $$ = labels.New() }
+                ;
+
+label_set_list  :
+                label_set_list COMMA label_set_item
```
brian-brazil: Hmm, do we currently allow trailing commas?

slrtbtfs: No.

What corresponds to `label_set_list` was previously parsed by the `labelMatchers()` function that is removed in this PR and is the same one that is used for label matchers in queries.

brian-brazil: Testing it we do, so this needs fixing and a unittest.

slrtbtfs: Uh, I see.

This means #6410 did break the old behavior, too.

brian-brazil: We were bound to have some breakage through this process, at least it didn't make it into a release.

slrtbtfs: The grammar *should* be fixed now.

I've added another unit test for the label matchers grammar from PR #6410 .

The labels set grammar isn't directly unit tested at all.

This is not as bad as it sounds, since it is only used by the `ParseMetric()` which is only used internally by unit tests. Still such tests should be added, possibly in another PR.

promql/parse_test.go
```diff
@@ -882,6 +882,19 @@ var testExpr = []struct {
 				mustLabelMatcher(labels.MatchEqual, string(model.MetricNameLabel), "foo"),
 			},
 		},
+	}, {
+		input: `foo{a="b", foo!="bar", test=~"test", bar!~"baz",}`,
```
brian-brazil: A simpler example can be used. We should also check that a leading comma is rejected.

slrtbtfs:
I've added some error message improvements.

Will work on proper unit testing tomorrow.

slrtbtfs:
There are unit tests that should trigger all possible error messages inside label matchers.

promql/generated_parser.y
slrtbtfs: Label matchers are now properly unit tested.

I'd consider adding unit tests for triggering error messages inside label sets less important, since label sets are only parsed by internal test routines anyway, and their parsing code looks almost the same as the one for label matchers.

If such tests are desired I can do it anyway.

promql/parse_test.go
slrtbtfs: More tests have been added now, including one with a leading comma.

promql/parse_test.go
```diff
@@ -952,6 +965,22 @@ var testExpr = []struct {
...
 		fail:   true,
 		errMsg: "metric name must not be set twice: \"foo\" or \"bar\"",
+	}, {
+		input:  `foo{__name__= =}`,
+		fail:   true,
+		errMsg: "unexpected <op:=> in label matching, expected string",
+	}, {
+		input:  `foo{,}`,
+		fail:   true,
+		errMsg: "unexpected \",\" in label matching, expected identifier or \"}\"",
```
brian-brazil: You can use ` to avoid all the escaping.

slrtbtfs: True.

I've changed the tests accordingly.

However I'd like to leave it as it is in the yacc grammar, since using ` messes up syntax highlighting in yacc files (tested with Vim and VS Code).

brian-brazil:
Thanks!

PR #6433, opened 2019-12-09
PromQL: Use more standart format for error positions

slrtbtfs:
The most common format (used by go, gcc and clang) for compiler error positions seems to be

`filename:line:char:` or `line:char:` if the filename is unknown.

This PR adapts the PromQL parser to use this convention.


brian-brazil:
Thanks!

PR #6434, opened 2019-12-09
PromQL: export function list

slrtbtfs:
This PR exports the list of supported PromQL functions and their signatures.

The reason for this is that the PromQL language server likes to use that list.


brian-brazil:
Thanks!

PR #6435, opened 2019-12-09
PromQL: Export lexer

slrtbtfs:
This PR exports the the types and functions needed to use the PromQL lexer.

The reason for this is that the PromQL language server would like to use the PromQL lexer for some completion functionality.


brian-brazil:
Thanks!

roidelapluie:
This might be interesting for @davkal  as well

PR #6437, opened 2019-12-09
React UI: Fix issue when changing query then time, the old query executed

### LINKED ISSUE #6405: React UI: When changing query then time, the old query is executed.
How to reproduce:

Take the new ui.

In the expression, write : time()
Click execute

Change the expression:
time()-time()
*do not click execute*

Change the execution time 

**Result**: the result value is time()

**Expected**: the result is 0

Additional note: this is a regression of the new ui

hooten:
This Pull Request fixes a bug in the querying behavior of the new React UI. The bug is described in the linked issue.

The root cause of the bug was the state management strategy in `ExpressionInput`. Previously, the query expression was passed into `ExpressionInput` as a prop, then stored as internal component state. When the query expression was changed, the parent components (which actually do the fetching) were not aware of the state changes. So when they performed a fetch, they sometimes used a stale query expression. Specifically, the query expression was stale whenever a query was executed without using the Execute button.

The solution I propose here is to lift the state management back into the parent components—`Panel` and `PanelList`— so that they know when the query expression changes and fetch using up-to-date values. In addition, I decouple state changes from querying so that each expression input keystroke does _not_ issue a new query. State changes that should perform queries do so in the lifecycle method `componentDidUpdate`. And every time a query is executed, the url is updated. This matches the behavior of the existing classic ui.

Fixes #6405

### DIFF
--- a/web/ui/react-app/src/Panel.tsx
+++ b/web/ui/react-app/src/Panel.tsx
@@ -68,14 +68,15 @@ class Panel extends Component<PanelProps & PathPrefixProps, PanelState> {
     };
   }
 
-  componentDidUpdate(prevProps: PanelProps, prevState: PanelState) {
-    const prevOpts = prevProps.options;
-    const opts = this.props.options;
-    if (prevOpts.type !== opts.type) {
-      // If the other options change, we still want to show the old data until the new
-      // query completes, but this is not a good idea when we actually change between
-      // table and graph view, since not all queries work well in both.
-      this.setState({ data: null });
+  componentDidUpdate({ options: prevOpts }: PanelProps) {
+    const { endTime, range, resolution, type } = this.props.options;
+    if (
+      prevOpts.endTime !== endTime ||
+      prevOpts.range !== range ||
+      prevOpts.resolution !== resolution ||
+      prevOpts.type !== type
+    ) {
+      this.executeQuery();
     }
   }
 
@@ -181,7 +182,6 @@ class Panel extends Component<PanelProps & PathPrefixProps, PanelState> {
 
   handleChangeRange = (range: number): void => {
     this.setOptions({ range: range });
-    this.executeQuery();
   };
 
   getEndTime = (): number | moment.Moment => {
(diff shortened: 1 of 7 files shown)


hooten:
@roidelapluie @juliusv I'd appreciate a review when you get a chance. Thanks!

juliusv:
@hooten Awesome, thanks, didn't even know we had this problem. Will give it a good review soon. One little thing I already spotted: switching between Graph/Table tabs leads to "No data queried yet" instead of (re)executing the query for the new view.

web/ui/react-app/src/Panel.tsx
```diff
@@ -203,14 +193,21 @@ class Panel extends Component<PanelProps & PathPrefixProps, PanelState> {
...
+    this.executeQuery();
   };
 
   handleChangeResolution = (resolution: number | null) => {
     this.setOptions({ resolution: resolution });
+    this.executeQuery();
+  };
+
+  handleChangeType = (type: PanelType) => {
+    this.setOptions({ type: type });
```
juliusv: I guess this is where another `this.executeQuery();` is needed as well.

hooten: Good call. I will work on this

hooten: @juliusv Switching between Graph/Table views now re-executes the query. I had to change a few things around. I think it's better now anyway.

web/ui/react-app/src/pages/PanelList.test.tsx
```diff
@@ -37,4 +39,12 @@ describe('PanelList', () => {
     expect(btn.prop('color')).toEqual('primary');
     expect(btn.children().text()).toEqual('Add Panel');
   });
+
+  it('updatesURL when a query is executed', () => {
```
juliusv: "updatesURL" -> "updates URL"

web/ui/react-app/src/pages/PanelList.tsx
```diff
@@ -181,7 +182,7 @@ class PanelList extends Component<RouteComponentProps & PathPrefixProps, PanelLi
         </Row>
         {panels.map(({ id, options }) => (
           <Panel
-            onExecuteQuery={this.handleQueryHistory}
+            onExecuteQuery={this.onExecuteQuery}
```
juliusv: The convention in React (at least in the docs like https://reactjs.org/docs/handling-events.html) seems to be to call the attribute for handlers `onX()`, but the actual function implementation it points to `handleX()`. So I think we'll want to have `handleExecuteQuery()` here.

hooten:
Thanks for the feedback! Let me know if I should fix anything else

juliusv:
:+1: Looks great now, thanks :)

hooten:
Thanks for merging! Again really happy to contribute

PR #6442, opened 2019-12-10
api: provide metadata for a specified metric

### LINKED ISSUE #6256: Proposal: Replacing the PromQL Parser with a generated parser

### LINKED ISSUE #6395: Proposal: Add API endpoint to retrieve metadata by metric name

gotjosh:
This adds support for a new query param on the new `api/v1/metadata` endpoint that provides metadata for a specified metric via the V1 API. Similarly to the `api/v1/target/metadata` the metadata is only available for the lifecycle of the attached target. 

An example of a payload looks like:

```bash
curl -G http://localhost:9090/api/v1/metadata?metric=http_requests_total | jq

{
  "status": "success",
  "data": {
    "http_requests_total": [
      {
        "type": "counter",
        "help": "Number of HTTP requests",
        "unit": ""
      },
      {
        "type": "counter",
        "help": "Amount of HTTP requests",
        "unit": ""
      }
    ]
  }
}
```

Follow up from #6420 
Closes #6395

### DIFF
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -665,6 +665,7 @@ GET /api/v1/metadata
 URL query parameters:
 
 - `limit=<number>`: Maximum number of metrics to return.
+- `metric=<string>`: A metric name to filter metadata for. All metric metadata is retrieved if left empty.
 
 The `data` section of the query result consists of an object where each key is a metric name and each value is a list of unique metadata objects, as exposed for that metric name across all targets.
 
@@ -683,6 +684,30 @@ curl -G http://localhost:9090/api/v1/metadata?limit=2
         "unit": ""
       }
     ],
+    "http_requests_total": [
+      {
+        "type": "counter",
+        "help": "Number of HTTP requests",
+        "unit": ""
+      },
+      {
+        "type": "counter",
+        "help": "Amount of HTTP requests",
+        "unit": ""
+      }
+    ]
+  }
+}
+```
+
+The following example returns metadata only for the metric `http_requests_total`.
+
+```json
+curl -G http://localhost:9090/api/v1/metadata?metric=http_requests_total
+
+{
+  "status": "success",
+  "data": {
     "http_requests_total": [
       {
         "type": "counter",
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -822,22 +822,39 @@ func (api *API) metricMetadata(r *http.Request) apiFuncResult {
 		}
 	}
 
+	metric := r.FormValue("metric")
+
 	for _, tt := range api.targetRetriever.TargetsActive() {
 		for _, t := range tt {
-			for _, mm := range t.MetadataList() {
-				m := metadata{Type: mm.Type, Help: mm.Help, Unit: mm.Unit}
-				ms, ok := metrics[mm.Metric]
+
+			if metric == "" {
+				for _, mm := range t.MetadataList() {
+					m := metadata{Type: mm.Type, Help: mm.Help, Unit: mm.Unit}
+					ms, ok := metrics[mm.Metric]
+
+					if !ok {
+						ms = map[metadata]struct{}{}
+						metrics[mm.Metric] = ms
+					}
+					ms[m] = struct{}{}
+				}
+				continue
+			}
+
+			if md, ok := t.Metadata(metric); ok {
+				m := metadata{Type: md.Type, Help: md.Help, Unit: md.Unit}
+				ms, ok := metrics[md.Metric]
 
 				if !ok {
 					ms = map[metadata]struct{}{}
-					metrics[mm.Metric] = ms
+					metrics[md.Metric] = ms
 				}
-
 				ms[m] = struct{}{}
 			}
 		}
 	}
 
+	// Put the elements from the pseudo-set into a slice for marshaling.
 	res := map[string][]metadata{}
 
 	for name, set := range metrics {
@@ -846,11 +863,9 @@ func (api *API) metricMetadata(r *http.Request) apiFuncResult {
 		}
 
 		s := []metadata{}
-
 		for metadata := range set {
 			s = append(s, metadata)
 		}
-
 		res[name] = s
 	}
 
--- a/web/api/v1/api_test.go
+++ b/web/api/v1/api_test.go
@@ -1220,6 +1220,73 @@ func testEndpoints(t *testing.T, api *API, tr *testTargetRetriever, testLabelAPI
 			},
 			responseLen: 2,
 		},
+		// When requesting a specific metric that is present.
+		{
+			endpoint: api.metricMetadata,
+			query:    url.Values{"metric": []string{"go_threads"}},
+			metadata: []targetMetadata{
+				{
+					identifier: "test",
+					metadata: []scrape.MetricMetadata{
+						{
+							Metric: "go_threads",
+							Type:   textparse.MetricTypeGauge,
+							Help:   "Number of OS threads created",
+							Unit:   "",
+						},
+					},
+				},
+				{
+					identifier: "blackbox",
+					metadata: []scrape.MetricMetadata{
+						{
+							Metric: "go_gc_duration_seconds",
+							Type:   textparse.MetricTypeSummary,
+							Help:   "A summary of the GC invocation durations.",
+							Unit:   "",
+						},
+						{
+							Metric: "go_threads",
+							Type:   textparse.MetricTypeGauge,
+							Help:   "Number of OS threads that were created.",
+							Unit:   "",
+						},
+					},
+				},
+			},
+			response: map[string][]metadata{
+				"go_threads": []metadata{
+					{textparse.MetricTypeGauge, "Number of OS threads created", ""},
+					{textparse.MetricTypeGauge, "Number of OS threads that were created.", ""},
+				},
+			},
+			sorter: func(m interface{}) {
+				v := m.(map[string][]metadata)["go_threads"]
+
+				sort.Slice(v, func(i, j int) bool {
+					return v[i].Help < v[j].Help
+				})
+			},
+		},
+		// With a specific metric that is not present.
+		{
+			endpoint: api.metricMetadata,
+			query:    url.Values{"metric": []string{"go_gc_duration_seconds"}},
+			metadata: []targetMetadata{
+				{
+					identifier: "test",
+					metadata: []scrape.MetricMetadata{
+						{
+							Metric: "go_threads",
+							Type:   textparse.MetricTypeGauge,
+							Help:   "Number of OS threads created",
+							Unit:   "",
+						},
+					},
+				},
+			},
+			response: map[string][]metadata{},
+		},
 		// With no available metadata.
 		{
 			endpoint: api.metricMetadata,
--- a/web/api/v1/api_test.go
+++ b/web/api/v1/api_test.go
@@ -1182,7 +1182,7 @@ func testEndpoints(t *testing.T, api *API, tr *testTargetRetriever, testLabelAPI
 				})
 			},
 		},
-		// With a limit for the number of metrics returned
+		// With a limit for the number of metrics returned.
 		{
 			endpoint: api.metricMetadata,
 			query: url.Values{
@@ -1220,7 +1220,7 @@ func testEndpoints(t *testing.T, api *API, tr *testTargetRetriever, testLabelAPI
 			},
 			responseLen: 2,
 		},
-		// With no available metadata
+		// With no available metadata.
 		{
 			endpoint: api.metricMetadata,
 			response: map[string][]metadata{},


docs/querying/api.md
```diff
@@ -699,6 +701,41 @@ curl -G http://localhost:9090/api/v1/metadata?limit=2
 }
 ```
 
+### Finding metadata for a single metric
+
+It returns metadata about the provided metric name currently scrapped from targets. However, it does not provide any target information.
+This is considered **experimental** and might change in the future.
+
+```
+GET /api/v1/metadata/<metric_name>
```
brian-brazil: We usually use URL parameters

web/api/v1/api.go
```diff
@@ -838,6 +839,7 @@ func (api *API) metricMetadata(r *http.Request) apiFuncResult {
 		}
 	}
 
+	// Put the elements from the pseudo-set into a slice for marshaling.
```
brian-brazil: Vertical whitespace around the inner for could also be reduced.

docs/querying/api.md
gotjosh: I noticed that, but the main reason for it is to circumvent URL character limits which I feel don't apply to our situation which is very similar to [querying label values](https://prometheus.io/docs/prometheus/latest/querying/api/#querying-label-values) (as these don't also have a limit).

I'm happy to change it if you think otherwise.

web/api/v1/api.go
gotjosh: Done.

docs/querying/api.md
brian-brazil: This doesn't help you get around URL length limits as it's still in the URL, we add POST where that's a concern.

The label values appears to be inconsistent for no good reason.

gotjosh:
> This should probably be part of the existing api.

Do you prefer the same handler to be reused even if it's a separate endpoint or do you want to include the functionality in the same endpoint but via a query param?

I'm not sure about the latter as the proposal described a separate endpoint unless we all agree to deviate from that.

brian-brazil:
It's basically the same code currently, so I'd have it all in one endpoint. It's also what we do elsewhere, we don't want to end up with an endpoint per possible parameter combination.

web/api/v1/api.go
```diff
@@ -822,22 +822,39 @@ func (api *API) metricMetadata(r *http.Request) apiFuncResult {
...
 
+	metric := r.FormValue("metric")
+
 	for _, tt := range api.targetRetriever.TargetsActive() {
 		for _, t := range tt {
-			for _, mm := range t.MetadataList() {
-				m := metadata{Type: mm.Type, Help: mm.Help, Unit: mm.Unit}
-				ms, ok := metrics[mm.Metric]
+
+			if metric == "" {
```
gotjosh: Does this work @brian-brazil? Will update the docs and PR description if agreed.

Was tempted to DRY up the appending of new metadata but refrained from that for now.

brian-brazil: Looks okay, I didn't recall that we had the t.Metadata.

gotjosh: We always did.

brian-brazil: I can't remember every PR to add a new feature that I reviewed :)

brian-brazil:
:+1: 

@beorn7 I believe that comment is what you wanted.

web/api/v1/api.go
gotjosh: sorry if I implied that with my comment was definitely not my intention, I came from an angle of a confirmation.

brian-brazil: Oh no, I like everyone else forget things sometimes and I was joking about that.

PR #6443, opened 2019-12-11
rules: use evalTimestamp more accurate.

johncming:
The `evaluationTimestamp`field of `Group` should use **evaluation timestamp** instead of **start** time. Because **start** time is not the real one to prometheus query engine.

https://github.com/prometheus/prometheus/blob/0a0a228db380c3d68decd720016ba568af8b4b6d/rules/manager.go#L229

https://github.com/prometheus/prometheus/blob/0a0a228db380c3d68decd720016ba568af8b4b6d/rules/manager.go#L290-L299

The estimation time for storing rules in the query engine is better, because the startup time in rule manager is not actually the query time of the metric query engine. Meanwhile, the current value is the latest evaluation time of the group or rule.

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -296,7 +296,7 @@ func (g *Group) run(ctx context.Context) {
 
 		g.metrics.iterationDuration.Observe(timeSinceStart.Seconds())
 		g.setEvaluationDuration(timeSinceStart)
-		g.setEvaluationTimestamp(start)
+		g.setEvaluationTimestamp(evalTimestamp)
 	}
 
 	// The assumption here is that since the ticker was started after having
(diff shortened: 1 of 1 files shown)


roidelapluie:
good catch

brian-brazil:
This is working as expected, this is used as a metric so we want to know when it actually started - not when it should have started.

johncming:
> This is working as expected, this is used as a metric so we want to know when it actually started - not when it should have started.

whether the estimation time in the query engine can be added?

brian-brazil:
I'm not sure that's too relevant. What we only really care about here is if an evaluation was skipped.

roidelapluie:
My first impression is that it was used to set the timestamp of the recorded metrics. It is not the case. Therefore I think the current behaviour is correct. The exact timestamp will still be found on timestamp () of the metrics.

bboreham:
This confused me too.  Perhaps it should be named `lastExecutionStartTimestamp` ?

brian-brazil:
We looked at this in the bug scrub, and per the above the current code is correct. However the field name doesn't match what's in the JSON in the HTTP API, so would you mind making that consistent?

johncming:
> We looked at this in the bug scrub, and per the above the current code is correct. However the field name doesn't match what's in the JSON in the HTTP API, so would you mind making that consistent?

OK. I will fix it.

PR #6444, opened 2019-12-11
skip empty alerts.

johncming:
skip empty alerts.


roidelapluie:
I wich case would you have that?

johncming:
> I wich case would you have that?

Because of the following statement

https://github.com/prometheus/prometheus/blob/0a0a228db380c3d68decd720016ba568af8b4b6d/notifier/notifier.go#L319-L320

brian-brazil:
Thanks!

PR #6445, opened 2019-12-11
Fix promlint order algorithm not work issue

RainbowMango:
The less function provided to `sort.SliceStable` not work well. Since we should  start to compare `Text` **only if** the two `Metric ` are equal.
```golang
	// Ensure deterministic output.
	sort.SliceStable(problems, func(i, j int) bool {
		if problems[i].Metric < problems[j].Metric {
			return true
		} 

		return problems[i].Text < problems[j].Text
	})
```

Signed-off-by: RainbowMango <renhongcai@huawei.com>

### DIFF
--- a/util/promlint/promlint.go
+++ b/util/promlint/promlint.go
@@ -87,6 +87,8 @@ func (l *Linter) Lint() ([]Problem, error) {
 	sort.SliceStable(problems, func(i, j int) bool {
 		if problems[i].Metric < problems[j].Metric {
 			return true
+		} else if problems[i].Metric > problems[j].Metric {
+			return false
 		}
 
 		return problems[i].Text < problems[j].Text


RainbowMango:
/cc @brancz 
I happened to see this issue when I try to lint metrics from Kubernetes.

brancz:
lgtm

RainbowMango:
By the way, I don't think we need to define a new type of `problems` here:
https://github.com/prometheus/prometheus/blob/0a0a228db380c3d68decd720016ba568af8b4b6d/util/promlint/promlint.go#L43-L54

If people using a `Goland` IDE, and Goland always complains `Receiver 'problems' may be 'nil' in call` in place of involving `Add()` function. This is anti-pattern? How do you think?
https://github.com/prometheus/prometheus/blob/0a0a228db380c3d68decd720016ba568af8b4b6d/util/promlint/promlint.go#L126

PR #6447, opened 2019-12-11
fixed typo

sstoner:
fix typo in log info

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -581,7 +581,7 @@ func (h *Head) Init(minValidTime int64) error {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
Can you add the DCO please?

tsdb/head.go
```diff
@@ -581,7 +581,7 @@ func (h *Head) Init(minValidTime int64) error {
 		return nil
 	}
 
-	level.Info(h.logger).Log("msg", "replaying WAL, this may take awhile")
+	level.Info(h.logger).Log("msg", "replaying WAL, this may take a while")
```
codesome: `awhile` is not wrong

codesome: (sorry, I guess in this context it is)

PR #6450, opened 2019-12-11
avoid stopping rule groups if new rule groups are as same as old rule…

alburthoffman:
this is for the discussion in the user mail "[reload rule groups smartly](https://groups.google.com/d/msg/prometheus-users/zfqGXBQTelw/xq4udUhRBgAJ)".

basically enhance the rule groups reload process:

1. skip the old group if new group is equal to the old group
1. stop the old group and start the new group if existing and not equal
1. start the new added groups

The motivation for this is because we meet issues when reload rule groups.
The remove read endpoint is not stable. so sometimes it will take a long long time to stop all rule groups.
But actually we only need to update few rules like 1% of rules running.


rules/alerting.go
```diff
@@ -538,3 +538,56 @@ func (r *AlertingRule) HTMLSnippet(pathPrefix string) html_template.HTML {
...
+
+	if len(r.externalLabels) != len(ar.externalLabels) {
+		return false
+	}
+
+	same := true
+	for k, v := range r.externalLabels {
+		av, ok := ar.externalLabels[k]
+		if !ok {
+			same = false
```
roidelapluie: return false

rules/alerting.go
```diff
@@ -538,3 +538,56 @@ func (r *AlertingRule) HTMLSnippet(pathPrefix string) html_template.HTML {
...
+
+	if r.annotations.Hash() != ar.annotations.Hash() {
+		return false
+	}
+
+	if len(r.externalLabels) != len(ar.externalLabels) {
+		return false
+	}
+
+	same := true
```
roidelapluie: you do not need this

rules/alerting.go
roidelapluie: return false

rules/alerting.go
```diff
@@ -538,3 +538,56 @@ func (r *AlertingRule) HTMLSnippet(pathPrefix string) html_template.HTML {
...
+	for k, v := range r.externalLabels {
+		av, ok := ar.externalLabels[k]
+		if !ok {
+			same = false
+			break
+		}
+		if strings.Compare(v, av) != 0 {
+			same = false
```
roidelapluie: return false

rules/alerting.go
```diff
@@ -538,3 +538,56 @@ func (r *AlertingRule) HTMLSnippet(pathPrefix string) html_template.HTML {
...
+			break
+		}
+	}
+
+	return same
```
roidelapluie: return len(r.externalLabels) == len(ar.externalLabels)

rules/manager.go
```diff
@@ -743,6 +747,53 @@ func (g *Group) RestoreForState(ts time.Time) {
...
+	for _, r := range g.rules {
+		gm[r.Name()] = r
+	}
+
+	ngm := make(map[string]Rule, len(ng.rules))
+	for _, r := range ng.rules {
+		ngm[r.Name()] = r
+	}
+
+	same := true
```
roidelapluie: you do not need this

rules/manager.go
```diff
@@ -743,6 +747,53 @@ func (g *Group) RestoreForState(ts time.Time) {
...
+	for n, nr := range ngm {
+		or, ok := gm[n]
+		if !ok {
+			same = false
```
roidelapluie: return false

rules/manager.go
```diff
@@ -743,6 +747,53 @@ func (g *Group) RestoreForState(ts time.Time) {
...
+			break
+		}
+		if !nr.Equals(or) {
+			same = false
```
roidelapluie: return false

rules/manager.go
```diff
@@ -743,6 +747,53 @@ func (g *Group) RestoreForState(ts time.Time) {
...
+			break
+		}
+	}
+	return same
```
roidelapluie: return len(ngm) == len(gm)

rules/manager.go
```diff
@@ -836,16 +887,21 @@ func (m *Manager) Update(interval time.Duration, files []string, externalLabels
 	m.restored = true
 
 	var wg sync.WaitGroup
-
+	fmt.Println(len(groups))
```
roidelapluie: debug thing?

alburthoffman: yeah. will remove it.

rules/alerting.go
alburthoffman: ok, this is better.

rules/alerting.go
alburthoffman: done

alburthoffman:
@richardkiene please review

rules/alerting.go
```diff
@@ -538,3 +538,49 @@ func (r *AlertingRule) HTMLSnippet(pathPrefix string) html_template.HTML {
...
+
+	if strings.Compare(r.vector.String(), ar.vector.String()) != 0 {
+		return false
+	}
+
+	if r.holdDuration != ar.holdDuration {
+		return false
+	}
+
+	if r.labels.Hash() != ar.labels.Hash() {
```
codesome: Using `labels.Equal` method would be faster. `!labels.Equal(r.labels, ar.labels)`. Same for annotations.

rules/alerting.go
```diff
@@ -538,3 +538,49 @@ func (r *AlertingRule) HTMLSnippet(pathPrefix string) html_template.HTML {
...
+		av, ok := ar.externalLabels[k]
+		if !ok {
+			return false
+		}
+		if strings.Compare(v, av) != 0 {
+			return false
+		}
+	}
+
+	return len(r.externalLabels) == len(ar.externalLabels)
```
codesome: The length check should come before you check the map contents.

rules/manager.go
```diff
@@ -743,6 +746,50 @@ func (g *Group) RestoreForState(ts time.Time) {
...
+
+	gm := make(map[string]Rule, len(g.rules))
+	for _, r := range g.rules {
+		gm[r.Name()] = r
+	}
+
+	ngm := make(map[string]Rule, len(ng.rules))
+	for _, r := range ng.rules {
+		ngm[r.Name()] = r
+	}
```
codesome: I am not sure if rule names can be repeated. Someone has to confirm.

rules/recording.go
```diff
@@ -190,3 +191,28 @@ func (rule *RecordingRule) HTMLSnippet(pathPrefix string) template.HTML {
...
+		return false
+	}
+
+	if strings.Compare(r.vector.String(), rr.vector.String()) != 0 {
+		return false
+	}
+
+	if r.labels.Hash() != rr.labels.Hash() {
+		return false
+	}
```
codesome: Same comments that I had for `AlertingRule`

rules/alerting.go
alburthoffman: yeah. ur right. this is better.

rules/alerting.go
```diff
@@ -538,3 +538,49 @@ func (r *AlertingRule) HTMLSnippet(pathPrefix string) html_template.HTML {
...
+// 4. labels
+// 5. annotations
+// 6. externalLabels
+func (r *AlertingRule) Equals(rule Rule) bool {
+	ar, ok := rule.(*AlertingRule)
+	if !ok {
+		return false
+	}
+
+	if strings.Compare(r.name, ar.name) != 0 {
```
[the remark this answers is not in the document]
alburthoffman: check the latest commit.

rules/recording.go
alburthoffman: check the latest commit.

rules/alerting.go
```diff
@@ -538,3 +538,53 @@ func (r *AlertingRule) HTMLSnippet(pathPrefix string) html_template.HTML {
...
+
+	if !labels.Equal(r.annotations, ar.annotations) {
+		return false
+	}
+
+	if len(r.externalLabels) != len(ar.externalLabels) {
+		return false
+	}
+
+	for k, v := range r.externalLabels {
```
roidelapluie: if !labels.Equal(r.externalLabels, ar.externalLabels) {

would that work??

rules/manager.go
```diff
@@ -24,7 +24,7 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
-	opentracing "github.com/opentracing/opentracing-go"
```
roidelapluie: unrelated?

rules/manager.go
roidelapluie: yes you can have multiple rules with same name

rules/manager.go
```diff
@@ -866,7 +916,6 @@ func (m *Manager) Update(interval time.Duration, files []string, externalLabels
 	for _, oldg := range m.groups {
 		oldg.stop()
 	}
-
```
roidelapluie: unrelated

rules/recording.go
```diff
@@ -21,7 +21,7 @@ import (
 	"sync"
 	"time"
 
-	yaml "gopkg.in/yaml.v2"
```
roidelapluie: unrelated

rules/alerting.go
codesome: It is a `map[string]string` type

roidelapluie:
The tests are probably too small to catch everything

roidelapluie:
@alburthoffman What about instead of trying to compare logically, we would compare the YAML representations? That should catch about every change and would be simples.

rules/manager.go
alburthoffman: not related. it is removed by `go mod tidy`

alburthoffman:
@roidelapluie the problem is manager.Update does not have RuleGroups, which are parsed from yaml files directly. Also manager will have to store the old RuleGroups as well.

The Group struct and Rule struct in manager.Update does not have any yaml annotations yet.

rules/alerting.go
alburthoffman: yeah. it wont work

johncming:
This comparison check  is  not faster than the original, if the update rule is more serious.

why not add Hash function in `AlertingRule` and `RecordingRule` which like `labels.Labels`.

https://github.com/prometheus/prometheus/blob/48d25e6fe7d960c76c01763303c38f11168b1bba/pkg/labels/labels.go#L121-L131

rules/manager.go
alburthoffman: add it back

rules/manager.go
alburthoffman: fix it by using Rule.String() as a key

rules/recording.go
alburthoffman: roll it back

alburthoffman:
@roidelapluie it's clean now. only manager.go gets modified. please check.

@johncming yeah. Hash will be much faster. but we will need to add Hash for AlertingRule and RecordingRule, Also probably Group as well. It's more easier to keep it simple, and performance might not be an issue here.

rules/manager.go
```diff
@@ -743,6 +743,47 @@ func (g *Group) RestoreForState(ts time.Time) {
...
+		return false
+	}
+
+	if len(g.rules) != len(ng.rules) {
+		return false
+	}
+
+	gm := make(map[string]Rule, len(g.rules))
+	for _, r := range g.rules {
+		gm[r.String()] = r
```
roidelapluie: why not just using a slice? and them compare them one by once? changing the order of rules is a different group.

alburthoffman: So the order of the rules also matters, right?

alburthoffman: That would make things much easier

alburthoffman: done.

alburthoffman:
@roidelapluie please review. thx

johncming:
> @roidelapluie it's clean now. only manager.go gets modified. please check.
> 
> @johncming yeah. Hash will be much faster. but we will need to add Hash for AlertingRule and RecordingRule, Also probably Group as well. It's more easier to keep it simple, and performance might not be an issue here.

@alburthoffman 
 There is nothing wrong with the existing way, but for performance: let the unchanged old group keep running. If not for performance, then for what.

alburthoffman:
"not an issue" means that performance might not be so bad that we have to optimize it at present. Here are our two choices for performance decision:

1. compare few fields (integers and strings)
2. calculate hash value and compare

not (stop all VS stop few). So I really do not thing the first choice is super slower than the second one. Honestly we need data for this kind of question, instead of assuming the second one is the best.

Remember that hash calculation is cpu intensive, and string comparison is simply looping...

johncming:
> "not an issue" means that performance might not be so bad that we have to optimize it at present. Here are our two choices for performance decision:
> 
> 1. compare few fields (integers and strings)
> 2. calculate hash value and compare
> 
> not (stop all VS stop few). So I really do not thing the first choice is super slower than the second one. Honestly we need data for this kind of question, instead of assuming the second one is the best.
> 
> Remember that hash calculation is cpu intensive, and string comparison is simply looping...

For the fist method, fewer fields may affect comparison accuracy, which is unacceptable for rule files.

alburthoffman:
In order to calculate the Hash value, you also have to decide which fields are used. Probably same fields as the ones in Equals method. So actually it makes no differences. 

I don't get ur second question. The changes aim to make less effort to archive the goal, like calling Rule.String() method instead of defining Equals in AlertingRule and RecordingRule.

So later, people who change String() wont need to change Equals() as well. Since no String() for Group, so we will have to add code to do it.

My point for performance is that don't rush to optimize it unless it is really needed. Normally we will need to profile and data for this.

alburthoffman:
@roidelapluie please review.

roidelapluie:
We have groups that take 2 minutes to run. I guess that this change will make it not needed to wait for them to stop.

roidelapluie:
Can you add tests?

brian-brazil:
I doubt rule groups will be more than a few MB, equals is fast enough for that. We can always optimise later.

alburthoffman:
> Can you add tests?

sure. tests have been cleanup, so will add tests to manager_test.go

rules/manager.go
```diff
@@ -743,6 +743,37 @@ func (g *Group) RestoreForState(ts time.Time) {
 
 }
 
+// Equals return if two groups have the same
+// 1. name
+// 2. file path
+// 3. interval
+// 4. rule set. and the rules are the same
```
roidelapluie: Maybe simplify thisEquals return if two groups are the same.

alburthoffman: that's ok. the code shows the same things. please check the latest commit .

alburthoffman:
@roidelapluie please view.

alburthoffman:
@brian-brazil are we good with this PR or still need enhancements? Thx

rules/manager.go
```diff
@@ -836,16 +863,20 @@ func (m *Manager) Update(interval time.Duration, files []string, externalLabels
...
 	for _, newg := range groups {
-		wg.Add(1)
-
 		// If there is an old group with the same identifier, stop it and wait for
 		// it to finish the current iteration. Then copy it into the new group.
 		gn := groupKey(newg.name, newg.file)
 		oldg, ok := m.groups[gn]
 		delete(m.groups, gn)
 
+		// if equal, move it to groups
```
brian-brazil: Capital letter and full stop

rules/manager.go
```diff
@@ -743,6 +743,33 @@ func (g *Group) RestoreForState(ts time.Time) {
 
 }
 
+// Equals return if two groups are the same
+func (g *Group) Equals(ng *Group) bool {
```
brian-brazil: This is going to be fragile when we add new fields, I'd suggest marshalling it and comparing that.

rules/manager.go
```diff
@@ -836,16 +863,20 @@ func (m *Manager) Update(interval time.Duration, files []string, externalLabels
 	m.restored = true
 
 	var wg sync.WaitGroup
-
 	for _, newg := range groups {
-		wg.Add(1)
-
 		// If there is an old group with the same identifier, stop it and wait for
```
brian-brazil: Comment needs updating.

rules/manager_test.go
```diff
@@ -703,18 +708,73 @@ func TestUpdate(t *testing.T) {
...
 	}
 
 	err = ruleManager.Update(10*time.Second, files, nil)
 	testutil.Ok(t, err)
-	for _, g := range ruleManager.groups {
+	for h, g := range ruleManager.groups {
 		for _, actual := range g.seriesInPreviousEval {
 			testutil.Equals(t, expected, actual)
 		}
+		// groups are the same because of no updates
```
brian-brazil: And for the rest of the comments

rules/manager.go
roidelapluie: Marshalling would require public fields.

rules/manager.go
alburthoffman: correct this.

rules/manager.go
alburthoffman: correct this.

rules/manager_test.go
alburthoffman: correct this.

rules/manager.go
alburthoffman: there are two options to avoid define extra Equals method:
1. expose public fields and allow marshalling
2. enhance Hash function to include fields.

for #2, it's not a good choice, because the Hash function is used to calculate the evaluation offset.

brian-brazil:
Thanks!

alburthoffman:
Thx

PR #6452, opened 2019-12-11
Reread index series rather than storing in memory.

brian-brazil:
Rather than building up a 2nd copy of all the posting
tables, construct it from the data we've already written
to disk. Benchmarks and more details in individual commits.

This saves memory, and also happens to be slightly faster.

### DIFF
--- a/tsdb/encoding/encoding.go
+++ b/tsdb/encoding/encoding.go
@@ -158,6 +158,14 @@ func NewDecbufUvarintAt(bs ByteSlice, off int, castagnoliTable *crc32.Table) Dec
 	return dec
 }
 
+// NewDecbufRaw returns a new decoding buffer of the given length.
+func NewDecbufRaw(bs ByteSlice, length int) Decbuf {
+	if bs.Len() < length {
+		return Decbuf{E: ErrInvalidSize}
+	}
+	return Decbuf{B: bs.Range(0, length)}
+}
+
 func (d *Decbuf) Uvarint() int     { return int(d.Uvarint64()) }
 func (d *Decbuf) Be32int() int     { return int(d.Be32()) }
 func (d *Decbuf) Be64int64() int64 { return int64(d.Be64()) }
@@ -258,6 +266,20 @@ func (d *Decbuf) Byte() byte {
 	x := d.B[0]
 	d.B = d.B[1:]
 	return x
+}
+
+func (d *Decbuf) EatPadding() {
+	if d.E != nil {
+		return
+	}
+	for len(d.B) > 1 && d.B[0] == '\x00' {
+		d.B = d.B[1:]
+	}
+	if len(d.B) < 1 {
+		d.E = ErrInvalidSize
+		return
+	}
+	return
 }
 
 func (d *Decbuf) Err() error  { return d.E }
--- a/tsdb/encoding/encoding.go
+++ b/tsdb/encoding/encoding.go
@@ -268,7 +268,7 @@ func (d *Decbuf) Byte() byte {
 	return x
 }
 
-func (d *Decbuf) EatPadding() {
+func (d *Decbuf) ConsumePadding() {
 	if d.E != nil {
 		return
 	}
@@ -279,7 +279,6 @@ func (d *Decbuf) EatPadding() {
 		d.E = ErrInvalidSize
 		return
 	}
-	return
 }
 
 func (d *Decbuf) Err() error  { return d.E }
--- a/tsdb/encoding/encoding.go
+++ b/tsdb/encoding/encoding.go
@@ -277,7 +277,6 @@ func (d *Decbuf) ConsumePadding() {
 	}
 	if len(d.B) < 1 {
 		d.E = ErrInvalidSize
-		return
 	}
 }
 
--- a/tsdb/index/index.go
+++ b/tsdb/index/index.go
@@ -610,10 +610,10 @@ func (w *Writer) writePostings() error {
 	}
 	defer f.Close()
 
-	// Write out the special all index.
+	// Write out the special all posting.
 	offsets := []uint32{}
 	d := encoding.NewDecbufRaw(realByteSlice(f.Bytes()), int(w.toc.LabelIndices))
-	d.B = d.B[w.toc.Series:] // dec.Skip not merged yet
+	d.Skip(int(w.toc.Series))
 	for d.Len() > 0 {
 		d.ConsumePadding()
 		startPos := w.toc.LabelIndices - uint64(d.Len())
@@ -622,8 +622,7 @@ func (w *Writer) writePostings() error {
 		}
 		offsets = append(offsets, uint32(startPos/16))
 		// Skip to next series. The 4 is for the CRC32.
-		skip := d.Uvarint() + 4
-		d.B = d.B[skip:]
+		d.Skip(d.Uvarint() + 4)
 		if err := d.Err(); err != nil {
 			return nil
 		}
@@ -643,6 +642,7 @@ func (w *Writer) writePostings() error {
 				break
 			}
 			batchNames = append(batchNames, names[0])
+			c += w.labelNames[names[0]]
 			names = names[1:]
 		}
 
@@ -678,8 +678,7 @@ func (w *Writer) writePostings() error {
 				}
 			}
 			// Skip to next series. The 4 is for the CRC32.
-			skip := l - (startLen - d.Len()) + 4
-			d.B = d.B[skip:]
+			d.Skip(l - (startLen - d.Len()) + 4)
 			if err := d.Err(); err != nil {
 				return nil
 			}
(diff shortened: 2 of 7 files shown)


tsdb/encoding/encoding.go
```diff
@@ -260,6 +268,19 @@ func (d *Decbuf) Byte() byte {
 	return x
 }
 
+func (d *Decbuf) EatPadding() {
```
bwplotka: Eat? ;p

brian-brazil: Do you have a better term?

brian-brazil:
This is ready for review.

tsdb/encoding/encoding.go
Kirides: `Consume` sounds good :)

brian-brazil: Noone for a Devour?

brian-brazil:
So actually I wasn't done with this :)

The performance was good enough compared to the previous implementation, however some quick profiling showed low hanging fruit that means we're overall now ~40% faster than the previous implementation.

tsdb/index/index.go
```diff
@@ -590,76 +594,143 @@ func (w *Writer) writeTOC() error {
...
+		startPos := w.toc.LabelIndices - uint64(d.Len())
+		if startPos%16 != 0 {
+			return errors.Errorf("series not 16-byte aligned at %d", startPos)
 		}
-		if offset > (1<<32)-1 {
-			return errors.Errorf("series offset %d exceeds 4 bytes", offset)
+		offsets = append(offsets, uint32(startPos/16))
+		// Skip to next series. The 4 is for the CRC32.
+		skip := d.Uvarint() + 4
+		d.B = d.B[skip:]
```
pracucci: We could use `d.Skip()` here too.

tsdb/index/index.go
```diff
@@ -590,76 +594,143 @@ func (w *Writer) writeTOC() error {
...
+	f, err := fileutil.OpenMmapFile(w.f.Name())
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	// Write out the special all index.
+	offsets := []uint32{}
+	d := encoding.NewDecbufRaw(realByteSlice(f.Bytes()), int(w.toc.LabelIndices))
+	d.B = d.B[w.toc.Series:] // dec.Skip not merged yet
```
pracucci: Now `d.Skip()` is merged and we could use it.

tsdb/index/index.go
```diff
@@ -590,76 +594,143 @@ func (w *Writer) writeTOC() error {
...
-		offset, ok := w.seriesOffsets[it.At()]
-		if !ok {
-			return errors.Errorf("%p series for reference %d not found", w, it.At())
+	f, err := fileutil.OpenMmapFile(w.f.Name())
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	// Write out the special all index.
```
pracucci: "special all index" sounds unclear to me. Could we improve it?

tsdb/index/index.go
brian-brazil: Though I'd already done that, must have gotten lost. Fixed.

tsdb/index/index.go
brian-brazil: That should say posting.

tsdb/encoding/encoding.go
```diff
@@ -260,6 +268,19 @@ func (d *Decbuf) Byte() byte {
...
+func (d *Decbuf) ConsumePadding() {
+	if d.E != nil {
+		return
+	}
+	for len(d.B) > 1 && d.B[0] == '\x00' {
+		d.B = d.B[1:]
+	}
+	if len(d.B) < 1 {
+		d.E = ErrInvalidSize
+		return
```
codesome: Not needed

tsdb/index/index.go
```diff
@@ -590,76 +594,141 @@ func (w *Writer) writeTOC() error {
...
-		if !ok {
-			return errors.Errorf("%p series for reference %d not found", w, it.At())
+	f, err := fileutil.OpenMmapFile(w.f.Name())
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	// Write out the special all posting.
+	offsets := []uint32{}
```
codesome: If we can keep track of number of series added, we can pre-allocate this and maybe reduce a lot of allocations when there are millions of series.

tsdb/index/index.go
```diff
@@ -590,76 +594,141 @@ func (w *Writer) writeTOC() error {
...
-		return err
-	}
-	w.crc32.Reset()
+	for len(names) > 0 {
+		batchNames := []string{}
+		var c uint64
+		// Try to bunch up label names into one loop, but avoid
+		// using more memory than a single label name can.
+		for len(names) > 0 {
+			if w.labelNames[names[0]]+c > maxPostings {
```
codesome: `c` always seems to be `0`. Did you miss a `c += w.labelNames[names[0]]` below? If yes, then with this change we might want to run the benchmark again.

tsdb/index/index.go
```diff
@@ -590,76 +594,141 @@ func (w *Writer) writeTOC() error {
...
+	maxPostings := uint64(len(offsets)) // No label name can have more postings than this.
 
-	startPos := w.pos
-	// Leave 4 bytes of space for the length, which will be calculated later.
-	if err := w.write([]byte("alen")); err != nil {
-		return err
-	}
-	w.crc32.Reset()
+	for len(names) > 0 {
+		batchNames := []string{}
```
codesome: Maybe we could re-use this?

tsdb/index/index.go
```diff
@@ -590,76 +594,141 @@ func (w *Writer) writeTOC() error {
...
+	d.Skip(int(w.toc.Series))
+	for d.Len() > 0 {
+		d.ConsumePadding()
+		startPos := w.toc.LabelIndices - uint64(d.Len())
+		if startPos%16 != 0 {
+			return errors.Errorf("series not 16-byte aligned at %d", startPos)
 		}
-		if offset > (1<<32)-1 {
-			return errors.Errorf("series offset %d exceeds 4 bytes", offset)
+		offsets = append(offsets, uint32(startPos/16))
```
codesome: Micro-opt: Maybe replace all `/16` with `<<4` if they are done a lot of times.

tsdb/index/index.go
```diff
@@ -590,76 +594,141 @@ func (w *Writer) writeTOC() error {
...
 
-	for _, r := range refs {
-		w.buf1.Reset()
-		w.buf1.PutBE32(r)
-		w.buf1.WriteToHash(w.crc32)
-		if err := w.write(w.buf1.Get()); err != nil {
-			return err
+		for _, name := range batchNames {
+			// Write out postings for this label name.
+			values := make([]uint32, 0, len(postings[w.symbols[name]]))
```
codesome: We could re-use the `offsets` slice for this.

tsdb/index/index.go
brian-brazil: It'll only be ~20 allocations, not much in the grand scheme of things.

tsdb/encoding/encoding.go
brian-brazil: Removed, was unsure whether to go for clarity or succinctness here.

tsdb/index/index.go
brian-brazil: Good catch, though it'll make no difference to this particular benchmark as it always gathers up everything.

tsdb/index/index.go
brian-brazil: There's not going to be that many label names. I'd prefer to leave microoptimsation like this to the future.

tsdb/index/index.go
brian-brazil: The compiler should do that already.

tsdb/index/index.go
brian-brazil: Not really, and once again this is only created once per label name.

brian-brazil:
Okay, speeding things up made the compaction cancellation test fail as it's timing dependant. I fixed that, but could someone review the 4th commit before I merge?

brian-brazil:
@bwplotka had a quick peek in IRC, so off we go.

tsdb/index/index.go
```diff
@@ -590,76 +603,147 @@ func (w *Writer) writeTOC() error {
...
-	w.buf1.PutBE32int(int(w.pos - startPos - 4))
-	if err := w.writeAt(w.buf1.Get(), startPos); err != nil {
+func (w *Writer) writePosting(name, value string, offs []uint32) error {
+	// Align beginning to 4 bytes for more efficient postings list scans.
+	if err := w.addPadding(4); err != nil {
 		return err
 	}
 
+	w.postings = append(w.postings, postingsHashEntry{
+		name:   name,
```
bwplotka: We talked about this offline, but curious if we could store name just once.

bwplotka:
Testing before release if I have permission. 404 so far for me.

tsdb/index/index.go
brian-brazil: That'd be a format change. Getting rid of this line is for a different PR.

tsdb/index/index.go
```diff
@@ -590,76 +594,142 @@ func (w *Writer) writeTOC() error {
...
-		offset, ok := w.seriesOffsets[it.At()]
-		if !ok {
-			return errors.Errorf("%p series for reference %d not found", w, it.At())
+	f, err := fileutil.OpenMmapFile(w.f.Name())
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	// Write out the special all posting.
```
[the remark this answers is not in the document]
brian-brazil: It's special in the context of the TSDB, as it contains all series in the posting.

tsdb/index/index.go
```diff
@@ -590,76 +603,147 @@ func (w *Writer) writeTOC() error {
...
+			numLabels := d.Uvarint()
+			for i := 0; i < numLabels; i++ {
+				lno := uint32(d.Uvarint())
+				lvo := uint32(d.Uvarint())
+
+				if _, ok := nameSymbols[lno]; ok {
+					if _, ok := postings[lno]; !ok {
+						postings[lno] = map[uint32][]uint32{}
+					}
+					if _, ok := postings[lno][lvo]; !ok {
```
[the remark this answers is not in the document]
brian-brazil: Removed in next PR.

tsdb/index/index.go
```diff
@@ -590,76 +603,147 @@ func (w *Writer) writeTOC() error {
...
 
-	w.buf1.Reset()
-	w.buf1.PutBE32int(len(refs))
-	w.buf1.WriteToHash(w.crc32)
-	if err := w.write(w.buf1.Get()); err != nil {
-		return err
-	}
+		nameSymbols := map[uint32]struct{}{}
+		for _, name := range batchNames {
+			nameSymbols[w.symbols[name]] = struct{}{}
```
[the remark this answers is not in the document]
brian-brazil: We could, but I think it's clearer this way.

tsdb/index/index.go
```diff
@@ -590,76 +594,142 @@ func (w *Writer) writeTOC() error {
...
+		d.ConsumePadding()
+		startPos := w.toc.LabelIndices - uint64(d.Len())
+		if startPos%16 != 0 {
+			return errors.Errorf("series not 16-byte aligned at %d", startPos)
 		}
-		if offset > (1<<32)-1 {
-			return errors.Errorf("series offset %d exceeds 4 bytes", offset)
+		offsets = append(offsets, uint32(startPos/16))
+		// Skip to next series. The 4 is for the CRC32.
+		d.Skip(d.Uvarint() + 4)
```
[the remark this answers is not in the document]
brian-brazil: Added to next PR

bwplotka:
/benchmark cancel

It works for me, as a side note, it's funny as testing on this PR vs master gives a weird difference. Master is without this PR so test makes sense:

image

Some difference in number of chunks 0.o 

image

...and memory (probably because of no chunks):

image

Weird, as we scrape the same data AFAIK .

PR #6454, opened 2019-12-12
add testcase for rangeForTimestamp.

johncming:
add testcase for rangeForTimestamp.


tsdb/db_test.go
```diff
@@ -2668,3 +2668,28 @@ func TestChunkWriter(t *testing.T) {
...
 	}
 }
+
+func TestRangeForTimestamp(t *testing.T) {
+	type args struct {
+		t     int64
+		width int64
+	}
+	tests := []struct {
+		name     string
```
brian-brazil: You don't need this, you can use the first value of the range.

johncming: fixed.

tsdb/db_test.go
```diff
@@ -2668,3 +2668,27 @@ func TestChunkWriter(t *testing.T) {
...
+		expected int64
+	}{
+		{args{1, 5}, 5},
+		{args{5, 5}, 10},
+		{args{6, 5}, 10},
+		{args{13, 5}, 15},
+		{args{95, 5}, 100},
+	}
+	for i, tt := range tests {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
```
brian-brazil: Usually we wouldn't use subtests here, these are all very fast tests. It usually only comes up with benchmarks.

tsdb/db_test.go
```diff
@@ -2668,3 +2668,27 @@ func TestChunkWriter(t *testing.T) {
...
+			if got := rangeForTimestamp(tt.args.t, tt.args.width); got != tt.expected {
+				testutil.Equals(t, tt.expected, got)
```
brian-brazil: testutil.Equals does the == for you

johncming: sorry, fixed.

tsdb/db_test.go
johncming: fixed.

tsdb/db_test.go
```diff
@@ -2668,3 +2668,24 @@ func TestChunkWriter(t *testing.T) {
...
+func TestRangeForTimestamp(t *testing.T) {
+	type args struct {
+		t     int64
+		width int64
+	}
+	tests := []struct {
+		args     args
+		expected int64
+	}{
+		{args{1, 5}, 5},
```
bwplotka: Worth to add some edge cases like `-20` or `0`?  (valid timestamps)

johncming: like `t=-20, width=5`?  Is there actually a negative value here?

brian-brazil:
Could you fix the conflicts?

johncming:
> Could you fix the conflicts?

OK. I will do it.

brian-brazil:
The tests are failing.

johncming:
> The tests are failing.

Updated.

brian-brazil:
Thanks!

PR #6455, opened 2019-12-12
*: avoid missed Alertmanager targets

simonpasquier:
This change makes sure that nearly-identical Alertmanager configurations
aren't merged together.

The config's identifier was the MD5 hash of the configuration serialized
to JSON but because `relabel.Regexp` has no public field and doesn't
implement the JSON.Marshaler interface, it was always serialized to
"{}".

In practice, the identifier can be based on the index of the
configuration in the list.

Issue reported in https://github.com/coreos/prometheus-operator/issues/2891

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -16,8 +16,6 @@ package main
 
 import (
 	"context"
-	"crypto/md5"
-	"encoding/json"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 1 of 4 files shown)


brian-brazil:
:+1: 

Are we doing this anywhere else? Secrets also won't be marshalled.

simonpasquier:
I've only found something similar in the remote write, see https://github.com/prometheus/prometheus/pull/6456

PR #6456, opened 2019-12-12
[WIP] storage/remote: use YAML data instead of JSON for MD5 hash

simonpasquier:
The configuration structs don't have JSON field tags so there's no guarantee that different configurations produce different JSON data. This isn't the case for YAML.

This is similar to https://github.com/prometheus/prometheus/pull/6455.

### DIFF
--- a/storage/remote/write_test.go
+++ b/storage/remote/write_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/pkg/relabel"
 	"github.com/prometheus/prometheus/util/testutil"
 )
 
@@ -248,10 +249,18 @@ func TestWriteStorageApplyConfigsPartialUpdate(t *testing.T) {
 	c0 := &config.RemoteWriteConfig{
 		RemoteTimeout: model.Duration(10 * time.Second),
 		QueueConfig:   config.DefaultQueueConfig,
+		WriteRelabelConfigs: []*relabel.Config{
+			&relabel.Config{
+				Regex: relabel.MustNewRegexp(".+"),
+			},
+		},
 	}
 	c1 := &config.RemoteWriteConfig{
 		RemoteTimeout: model.Duration(20 * time.Second),
 		QueueConfig:   config.DefaultQueueConfig,
+		HTTPClientConfig: config_util.HTTPClientConfig{
+			BearerToken: "foo",
+		},
 	}
 	c2 := &config.RemoteWriteConfig{
 		RemoteTimeout: model.Duration(30 * time.Second),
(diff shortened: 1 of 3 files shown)


brian-brazil:
Huh? I don't see how that could be an issue.

That we're not considering secrets is an issue.

simonpasquier:
> Huh? I don't see how that could be an issue.

`relabel.Regexp` are always marshaled to `{}`.

> That we're not considering secrets is an issue.

True. I've added a unit test to catch this.

storage/remote/write_test.go
```diff
@@ -120,31 +136,51 @@ func TestWriteStorageApplyConfigsPartialUpdate(t *testing.T) {
...
+	q0, q1, q2 = s.queues[0], s.queues[1], s.queues[2]
+
+	// Update c1.
+	c1.HTTPClientConfig.BearerToken = "bar"
+	err = s.ApplyConfig(conf)
+	testutil.Ok(t, err)
+	testutil.Equals(t, 3, len(s.queues))
+
+	testutil.Assert(t, q0 == s.queues[0], "Pointer of unchanged queue should have remained the same")
+	testutil.Assert(t, q1 != s.queues[1], "If the index changed, the queue should be stopped and recreated.")
```
csmarchbanks: Shouldn't the message here be around changing the bearer token should restart the queue?

brian-brazil:
> relabel.Regexp are always marshaled to {}.

I don't see what that has to do with field tags.

storage/remote/write_test.go
```diff
@@ -120,31 +136,51 @@ func TestWriteStorageApplyConfigsPartialUpdate(t *testing.T) {
...
 		GlobalConfig:       config.GlobalConfig{},
 		RemoteWriteConfigs: []*config.RemoteWriteConfig{c0, c1, c2},
 	}
-	s.ApplyConfig(conf)
+	err = s.ApplyConfig(conf)
+	testutil.Ok(t, err)
 	testutil.Equals(t, 3, len(s.queues))
 
-	testutil.Assert(t, q == s.queues[1], "Pointer of unchanged queue should have remained the same")
+	testutil.Assert(t, q0 != s.queues[0], "If the index changed, the queue should be stopped and recreated.")
```
cstyan: The message here isn't helpful in my opinion. What index are you referring to? The index of the queue's config in the overall `RemoteWriteConfigs` array? Rearranging those should no longer result in queue restarts. So I think this (and similar lines) should mention which part of the config was changed, and that when changing that field we expect to see a queue restart.

storage/remote/write_test.go
```diff
@@ -19,9 +19,12 @@ import (
 	"testing"
 	"time"
 
+	config_util "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
+
```
cstyan: unintended newline?

johncming:
@simonpasquier

the master branch has updated which use hash instead of index. #6043

storage/remote/write_test.go
simonpasquier: no the convention for package imports are:

```
stdlib packages

3rd party packages

local packages
```

simonpasquier:
As @johncming pointed out, I need to rebase the change but from a quick look, the issues should still exist.

storage/remote/write_test.go
cstyan: They're 3rd party because they're in `prometheus/common`?

cstyan:
> As @johncming pointed out, I need to rebase the change but from a quick look, the issues should still exist.

My bad, I pointed out that the test failure messages should be updated but that PR had only been merged a few hours earlier. It had been open long enough that I just had it in my head that the index to hash change was already merged.

storage/remote/write_test.go
simonpasquier: yes, it's a different Go module.

jutley:
I've run into this bug in two ways, both with regards to write_relabel_configs:
1) When the only difference between two remote_write configs is a regex, they produce the same hash, and Prometheus refuses to start. This is new in 2.15.
2) When a remote_write is updated and the only change is in its regex, the new rules never take effect. This was an issue before 2.15.

I'd really like to see this PR merged. We can create hacky workarounds, but would prefer not to.

csmarchbanks:
@simonpasquier Do you think you will have an opportunity to rebase this soon?

simonpasquier:
I've rebased my change but it only includes the test that demonstrates the issue. I don't have an immediate solution though.

csmarchbanks:
I thought just changing the hashes to use YAML instead of JSON fixed it? Did those changes disappear?

simonpasquier:
It does fix the issue in a way but it breaks in another. More precisely secret fields (bearer tokens for instance) are marshaled as `<secret>` in YAML so [this test](https://github.com/prometheus/prometheus/pull/6456/files#diff-8f7e46a2d0af8c88e81b21172317e3e6R313-R324) won't pass if we switch from JSON to YAML.

jutley:
Would it make sense to update the yaml marshaller to include a hash of the secret? That should fix the toHash function for regexes and secrets without exposing the secret. Not sure if it will break anything.

roidelapluie:
I did think about it but I don't like it... That would make the hash user
visible

Le sam. 1 févr. 2020 à 00:25, Jake Utley <notifications@github.com> a
écrit :

jutley:
@roidelapluie To be clear, what would be the concern with a user-visible hash? I suppose it would allow a user to identify secret re-use. Are there any other concerns?

roidelapluie:
Hash should not be used to hide secrets. They are not meant for security.

brian-brazil:
It depends on the hash, but even with a proper cryptographic hash you could e.g. notice that two secrets were the same.

jutley:
So, it seems like there are 2 requirements here:
1) Configuration marshaling exposes nothing about a secret's value
2) Configuration hashing reflects changes to any field (including regexes and secrets)

In order for these to not conflict, I think we must either 1) salt and hash the secrets when marshaling, or 2) not implement hashing using YAML marshaling.

I already implemented the second option with my original solution in #6546, which uses the JSON marshaler and provides an implementation for Regexp. However, this really feels like an abuse of the fact that there are multiple marshalers available, especially given that the JSON marshalers are never intentionally implemented in this project.

csmarchbanks:
Would it be reasonable to keep the yaml marshaling for the hash, thus having redacted secrets, and specifically update a queue whenever one of the secrets changes? Or else have a way to update the http client on a `remote.Client` so we wouldn't even have to restart the whole queue if the auth changes?

roidelapluie:
What prevents us to use reflects.Deepequal?

csmarchbanks:
Closing as I took this over in #7073

Issue #6406, opened 2019-12-04
React UI: allow for second timestamp


roidelapluie:
In the new ui the timestamp selector is rounded to the minute.

Being able to select the second is a must.

boyskila:
Hi, I'll take a look at this

boyskila:
@juliusv @roidelapluie in the classic UI inc/dec time input threshold is 10s but in the React UI, there is a range input with predefined values which can be either switched by inc/dec buttons or to directly type into it. Regarding this, two questions appear.

1. Should the old behavior remain in terms of the 10s threshold? It would work the same way except if the chosen range is < 10s

2. Since we can type into the input shouldn't be better to remove the predefined values and to leave the user to enter whatever he wants. Of course, some validation needs to be done, something like whether it starts with number and ends with {sec, hours, day/s}

Also in the current inc/dec implementation, the time is formed by the formula
```basetime + (range * 1000) / 2```
which seems to not work as (at least I) expected.
https://github.com/prometheus/prometheus/blob/0a0a228db380c3d68decd720016ba568af8b4b6d/web/ui/react-app/src/TimeInput.tsx#L42-L50

The (incorrect) behavior is most noticeable when the range is in second
So removing the division is the fix but I guess there is a reason to be done this way.

roidelapluie:
@boyskila To answer your questions, I like the way it is implemented in your PR!

PR #6460, opened 2019-12-13
Allow sec timestamp

boyskila:
This PR adds a seconds timestamp to the time picker and fixes the inc/dec behavior
Still have open questions [here](https://github.com/prometheus/prometheus/issues/6406#issuecomment-564411146)
Fixes: https://github.com/prometheus/prometheus/issues/6406
@juliusv @roidelapluie

### DIFF
--- a/web/ui/react-app/src/App.css
+++ b/web/ui/react-app/src/App.css
@@ -108,6 +108,7 @@ button.execute-btn {
 
 .graph-controls .time-input input {
     border-right: none;
+    padding-right: 0;
 }
 
 div.time-input {
--- a/web/ui/react-app/src/App.css
+++ b/web/ui/react-app/src/App.css
@@ -111,7 +111,7 @@ button.execute-btn {
 }
 
 div.time-input {
-    width: 240px !important;
+    width: 250px !important;
 }
 
 .table-controls input {
--- a/web/ui/react-app/src/App.css
+++ b/web/ui/react-app/src/App.css
@@ -98,7 +98,7 @@ button.execute-btn {
     margin-bottom: 10px;
 }
 
-.graph-controls input {
+.graph-controls input, .table-controls input {
     text-align: center;
 }
 
@@ -106,28 +106,23 @@ button.execute-btn {
     width: 50px;
 }
 
-.graph-controls .time-input input {
+.time-input input {
     border-right: none;
-    padding-right: 0;
 }
 
-div.time-input {
-    width: 250px !important;
-}
-
-.table-controls input {
-    text-align: center;
+.time-input {
+    width: 270px !important;
 }
 
 .graph-controls input.resolution-input {
     width: 90px;
 }
 
-.graph-controls .time-input, .graph-controls .resolution-input, .graph-controls .stacked-input {
+.graph-controls > :not(:first-child) {
     margin-left: 20px;
 }
 
-.graph-controls .clear-time-btn {
+    .graph-controls .clear-time-btn, .table-controls .clear-time-btn {
     background: #fff;
     border-left: none;
     border-top: 1px solid #ced4da;
--- a/web/ui/react-app/src/TimeInput.tsx
+++ b/web/ui/react-app/src/TimeInput.tsx
@@ -38,10 +38,7 @@ class TimeInput extends Component<TimeInputProps> {
     return this.props.time || moment().valueOf();
   };
 
-  calcTimeRange = () => {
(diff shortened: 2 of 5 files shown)


roidelapluie:
Should you change the behaviour of "decrease time" buttons? -1h10s is strange

boyskila:
@roidelapluie you mean that 10sek range always persists even the selected range is 1h or something else? If so I can suggest removing this behavior at all

roidelapluie:
oh it is 10ksec?? maybe we could do 10800 to match 3 hours.

boyskila:
@roidelapluie  the correct behaviour, I guess, is just to inc./dec. with the selected value. I think I misunderstood the way is working in the classic UI

roidelapluie:
Which button do you think we speak about?

roidelapluie:
transp3
that button moves 10 s in old ui

boyskila:
I'm talking about for the same time picker but on the graph tab.

roidelapluie:
As 2.15 is approaching , can we get that fixed before? :)

boyskila:
> As 2.15 is approaching , can we get that fixed before? :)

Sure, just to clarify what should be the behavior, because in the classic UI looks broken to me.

web/ui/react-app/src/App.css
```diff
@@ -108,10 +108,11 @@ button.execute-btn {
 
 .graph-controls .time-input input {
     border-right: none;
+    padding-right: 0;
 }
 
 div.time-input {
-    width: 240px !important;
+    width: 250px !important;
```
juliusv: This needs to be a bit wider, in the Table view, the seconds are a bit cut off.

juliusv: Also, the "X" timestamp removal button is differently styled (solid vs. transparent) in the Table view than the Graph view (via `.graph-controls .clear-time-btn` in `App.css`). We should probably change both views to be transparent.

juliusv: (also, other sizings seem a bit off between the two time pickers - would be great to get this consistent and cleaned up)

roidelapluie:
In the graph view: Move by half a graph interval

In the console view: move by 10 seconds(like old ui)?

boyskila:
In the React UI selecting a time range (on graph tab) is affecting the time input on the console tab. Should this remain?

roidelapluie:
oh yes!!

juliusv:
I think neither behavior is clearly the right one, I'd be for keeping the current React UI behavior unless we get a better reasoning whether it should be one way or the other.

web/ui/react-app/src/TimeInput.tsx
```diff
@@ -32,19 +32,24 @@ interface TimeInputProps {
...
-  private $time: any | null = null;
+  private $time: any = null;
 
   getBaseTime = (): number => {
     return this.props.time || moment().valueOf();
   };
 
+  calcTimeRange = () => {
+    const { range } = this.props;
+    return range * 1000 + (range > 10 ? 10000 : 0);
```
juliusv: I don't see how this behaves the way we want. E.g. if I have a range of 60s, my range is now shifted by 70s - what is the purpose of that?

I think we still want the old behavior (at least for the graph) where we shift by 1/2 of the configured range, so that enough (half) of the old portion of a graph is still visible when shifting the data to the left or right.

boyskila: As I mention here https://github.com/prometheus/prometheus/pull/6460#issuecomment-565755035
I misunderstood the way it's working in the old UI so I'll revert all this and just will add the seconds timestamp

juliusv: Ok cool.

web/ui/react-app/src/TimeInput.tsx
```diff
@@ -32,19 +32,21 @@ interface TimeInputProps {
...
 class TimeInput extends Component<TimeInputProps> {
   private timeInputRef = React.createRef<HTMLInputElement>();
-  private $time: any | null = null;
+  private $time: any = null;
 
   getBaseTime = (): number => {
     return this.props.time || moment().valueOf();
   };
 
+  calcTimeRange = () => (this.props.range * 1000) / 2;
```
juliusv:
```suggestion
  calcShiftRange = () => (this.props.range * 1000) / 2;
```

...might be clearer.

juliusv:
:+1: Thanks!

PR #6461, opened 2019-12-13
Load only some offsets into the symbol table into memory.

brian-brazil:
Full details in the commits, but short version is that we save a large amount of RAM by only storing .25bytes per symbol entry instead of a full map of them, at the cost of ~360ms per 1M series returned from PostingsForMatchers.


tsdb/index/index.go
```diff
@@ -1064,77 +1130,72 @@ func (r *Reader) Close() error {
 }
 
 func (r *Reader) lookupSymbol(o uint32) (string, error) {
```
bwplotka: Lot's of repetitions here for label values on series decode, are we sure this is fine overall?

tsdb/index/index.go
```diff
@@ -993,39 +972,126 @@ func (r *Reader) PostingsRanges() (map[labels.Label]Range, error) {
...
+	}
+	return uint32(s.bs.Len() - lastLen), nil
+}
+
+func (s Symbols) All() (map[string]struct{}, error) {
+	d := encoding.NewDecbufAt(s.bs, s.off, castagnoliTable)
+	cnt := d.Be32int()
+	res := make(map[string]struct{}, cnt)
+	for d.Err() == nil && cnt > 0 {
+		res[d.UvarintStr()] = struct{}{}
```
bwplotka: why not using yoloStrings here if we use this method only for `populateAll` :thinking:

bwplotka: Ok we actually change logic here: https://github.com/prometheus/prometheus/pull/6468 makes sense

tsdb/index/index.go
```diff
@@ -993,39 +972,126 @@ func (r *Reader) PostingsRanges() (map[labels.Label]Range, error) {
...
 	var (
-		origLen     = d.Len()
-		cnt         = d.Be32int()
-		basePos     = uint32(off) + 4
-		nextPos     = basePos + uint32(origLen-d.Len())
-		symbolSlice []string
-		symbols     = map[uint32]string{}
+		origLen = d.Len()
+		cnt     = d.Be32int()
+		basePos = off + 4
```
bwplotka: `crc32.Size` ?

tsdb/index/index.go
```diff
@@ -993,39 +972,126 @@ func (r *Reader) PostingsRanges() (map[labels.Label]Range, error) {
...
-// ReadSymbols reads the symbol table fully into memory and allocates proper strings for them.
-// Strings backed by the mmap'd memory would cause memory faults if applications keep using them
-// after the reader is closed.
-func ReadSymbols(bs ByteSlice, version int, off int) ([]string, map[uint32]string, error) {
+type Symbols struct {
+	bs      ByteSlice
+	version int
+	off     int
+
+	index []int
```
bwplotka: I feel it's quite cryptic. Maybe `chunkedOffsets`?

tsdb/index/index.go
brian-brazil: That's not a crc32.

tsdb/index/index.go
brian-brazil: That's the cpu vs ram tradeoff this PR is making.

tsdb/index/index.go
```diff
@@ -993,39 +972,126 @@ func (r *Reader) PostingsRanges() (map[labels.Label]Range, error) {
...
+	seen  int
+}
+
+const SYMBOL_FACTOR = 32
```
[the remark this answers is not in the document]

tsdb/querier_bench_test.go
```diff
@@ -123,3 +124,56 @@ func benchmarkPostingsForMatchers(b *testing.B, ir IndexReader) {
...
+
+				b.ResetTimer()
+				for i := 0; i < b.N; i++ {
+					ss, err := q.Select(matcher)
+					testutil.Ok(b, err)
+					for ss.Next() {
+					}
+					testutil.Ok(b, ss.Err())
+				}
+				q.Close()
```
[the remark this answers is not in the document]
brian-brazil: That'd leave queriers open though in the happy case, which is undesirable. Thus why I did it this way.

tsdb/index/index.go
```diff
@@ -993,39 +972,126 @@ func (r *Reader) PostingsRanges() (map[labels.Label]Range, error) {
...
+
+func NewSymbols(bs ByteSlice, version int, off int) (*Symbols, error) {
```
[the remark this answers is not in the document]

tsdb/index/index.go
brian-brazil: Changed to offsets.

brian-brazil:
> Worried about the benchmarks -> increasing memory for querier select is worrying (from the results in your commit). Does this mean that each request will have 70-30% more footprint (less for bigger number of series)? Are we happy about this?

This is due to creating new strings each time, whereas previously they were interned. I'm hoping to yoloString these, however it requires investigation to find a safe way to do it first in recording rules.

bwplotka:
:+1:

PR #6465, opened 2019-12-16
Add PromQL syntax text with tabs, spaces, trailing commas, singlequotes

roidelapluie:

### DIFF
--- a/promql/testdata/syntax.test
+++ b/promql/testdata/syntax.test
@@ -1,0 +1,25 @@
+clear
+load 1h
+	testmetric{aa="bb"} 1
+	testmetric{a="abb"} 2
+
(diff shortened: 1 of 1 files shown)


brian-brazil:
I'd suggest a comment that some of these are tabs/spaces. Otherwise it might get auto-formatted away at some point - might even be worth testing these ones in Go rather than here.

slrtbtfs:
You might want to include some test with backquotes, too.

brian-brazil:
I guess some weird unicode quotes while we're at it too.

roidelapluie:
OK I will work on that in the coming days

roidelapluie:
Most of this is actually already covered now.

PR #6466, opened 2019-12-16
PromQL: Parse Metrics using generated parser

slrtbtfs:
This PR further extends the generated PromQL parser.

The `promql.ParseMetric()` function is now completely based on the generated parser.

cc @simonpasquier @brian-brazil

### DIFF
--- a/promql/generated_parser.y
+++ b/promql/generated_parser.y
@@ -15,6 +15,8 @@
     package promql
 
     import (
+        "sort"
+
         "github.com/prometheus/prometheus/pkg/labels"
     )
 %}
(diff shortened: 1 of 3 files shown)


promql/generated_parser.y
```diff
@@ -167,13 +170,30 @@ match_op        :
                 | NEQ_REGEX {$$=$1}
                 ;
 
+
+metric          :
+                metric_identifier label_set
+                        { $$ = append($2, labels.Label{Name: labels.MetricName, Value: $1.Val}); sort.Sort($$) }
```
brian-brazil: Why are you sorting these? That'll change any eventual AST.

slrtbtfs: To avoid changing the AST.

The existing implementation does it, too: https://github.com/prometheus/prometheus/blob/master/promql/parse.go#L829

slrtbtfs: Apart from not breaking existing behavior idk any good reason to sort here.

brian-brazil: Let's not break it then without a reason.

brian-brazil:
Thanks!

PR #6469, opened 2019-12-16
Check that rules don't contain metrics with the same labelset

### LINKED ISSUE #5529: Random value in recording rule with multiple identical series result

roidelapluie:
Closes #5529

### DIFF
--- a/rules/alerting.go
+++ b/rules/alerting.go
@@ -311,6 +311,7 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc, 
 	resultFPs := map[uint64]struct{}{}
 
 	var vec promql.Vector
+	var alerts = make(map[uint64]*Alert, len(res))
 	for _, smpl := range res {
 		// Provide the alert information to the template.
 		l := make(map[string]string, len(smpl.Metric))
@@ -361,21 +362,34 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc, 
 		h := lbs.Hash()
 		resultFPs[h] = struct{}{}
 
-		// Check whether we already have alerting state for the identifying label set.
-		// Update the last value and annotations if so, create a new alert entry otherwise.
-		if alert, ok := r.active[h]; ok && alert.State != StateInactive {
-			alert.Value = smpl.V
-			alert.Annotations = annotations
-			continue
+		if _, ok := alerts[h]; ok {
+			err = fmt.Errorf("vector contains metrics with the same labelset after applying alert labels")
+			// We have already acquired the lock above hence using SetHealth and
+			// SetLastError will deadlock.
+			r.health = HealthBad
+			r.lastError = err
+			return nil, err
 		}
 
-		r.active[h] = &Alert{
+		alerts[h] = &Alert{
 			Labels:      lbs,
 			Annotations: annotations,
 			ActiveAt:    ts,
 			State:       StatePending,
 			Value:       smpl.V,
 		}
+	}
+
+	for h, a := range alerts {
+		// Check whether we already have alerting state for the identifying label set.
+		// Update the last value and annotations if so, create a new alert entry otherwise.
+		if alert, ok := r.active[h]; ok && alert.State != StateInactive {
+			alert.Value = a.Value
+			alert.Annotations = a.Annotations
+			continue
+		}
+
+		r.active[h] = a
 	}
 
 	// Check if any pending alerts should be removed or fire now. Write out alert timeseries.
--- a/rules/alerting_test.go
+++ b/rules/alerting_test.go
@@ -14,6 +14,8 @@
 package rules
 
 import (
+	"context"
+	"fmt"
 	"testing"
 	"time"
 
@@ -21,6 +23,7 @@ import (
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/pkg/timestamp"
 	"github.com/prometheus/prometheus/promql"
+	"github.com/prometheus/prometheus/util/teststorage"
 	"github.com/prometheus/prometheus/util/testutil"
 )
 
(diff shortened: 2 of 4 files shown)


juliusv:
Have you looked into how doable it would be to fix this on the TSDB side instead? There are error codes defined there that aren't returned correctly at the moment in this scenario (duplicate series in the same batch), and I wonder if alerting rules (`ALERTS` time series) and ingestion could be affected as well.

brian-brazil:
This is the correct fix from the Prometheus standpoint, the TSDB at most would silently drop some of the data - which is what we're trying to fix here. I'll give a proper review in the morning.

roidelapluie:
It is tricker with alerts - what to do with same alertset ? Fail and let them expire? Fail and don't make them fire?

juliusv:
> This is the correct fix from the Prometheus standpoint, the TSDB at most would silently drop some of the data - which is what we're trying to fix here. I'll give a proper review in the morning.

What's the reason not to fix this in the TSDB though? Is it infeasible? The TSDB already tries to detect this situation (via `ErrAmendSample`), but currently doesn't do so correctly if the conflicting sample is in the same batch. So without understanding more I would have assumed that that's the right place to fix it. But maybe you also want to really ensure here that the entire vector fails if it contains duplicates?

juliusv:
> It is tricker with alerts - what to do with same alertset ? Fail and let them expire? Fail and don't make them fire? As they will end up being the same in TSDB I would let them as is

At least it should detect the situation and log an error so that someone has a chance of catching it. Not sure if other behavioral changes make sense there, yeah.

brian-brazil:
> What's the reason not to fix this in the TSDB though?

Better error reporting in the TSDB wouldn't hurt but doesn't solve the problem here.

> At least it should detect the situation and log an error so that someone has a chance of catching it.

PromQL will report an error if you try to do this sort of thing, however as alert templating is messing about with labels outside of PromQL we need to duplicate that logic here.

This should look the same as any other PromQL error during rule execution.

rules/recording.go
```diff
@@ -93,6 +93,16 @@ func (rule *RecordingRule) Eval(ctx context.Context, ts time.Time, query QueryFu
 
 		sample.Metric = lb.Labels()
 	}
+
+	// Check that the rule does not produce identical metrics after applying
+	// labels #5529
```
brian-brazil: I'd put the reference to the issue in the unittest. This isn't some weird corner case, it's a design thing so feels weird to document by bug.

rules/recording.go
```diff
@@ -93,6 +93,16 @@ func (rule *RecordingRule) Eval(ctx context.Context, ts time.Time, query QueryFu
 
 		sample.Metric = lb.Labels()
 	}
+
+	// Check that the rule does not produce identical metrics after applying
```
brian-brazil: full stop

rules/recording.go
```diff
@@ -93,6 +93,16 @@ func (rule *RecordingRule) Eval(ctx context.Context, ts time.Time, query QueryFu
...
+	// labels #5529
+	if vector.ContainsSameLabelset() {
+		err = fmt.Errorf("resulting vector cannot contain metrics with the same labelset")
```
brian-brazil: I'd make this more in line with the error inside PromQL, but also indicating that this is related to templating and not PromQL per-se.

juliusv:
@brian-brazil Ok yeah, makes sense. Still think it'd be good to fix the TSDB as well, but can be separate.

roidelapluie:
ready for a second pass

rules/alerting.go
```diff
@@ -404,6 +407,16 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc,
 		}
 	}
 
+	// Check that the alert does not produce identical metrics after applying
+	// labels.
+	if res.ContainsSameLabelset() {
+		level.Warn(r.logger).Log("msg", "Ooops")
```
brian-brazil: Debugging?

rules/alerting.go
```diff
@@ -404,6 +407,16 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc,
...
+		err = fmt.Errorf("vector cannot contain metrics with the same labelset after applying extra labels")
```
brian-brazil: vector contains metrics with the same labelset after applying rule/alert labels

rules/alerting.go
```diff
@@ -311,7 +311,9 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc,
 	resultFPs := map[uint64]struct{}{}
 
 	var vec promql.Vector
-	for _, smpl := range res {
+	for i := range res {
```
brian-brazil: Why this change?

roidelapluie: To get the pointer and be able to change the sample in place

brian-brazil: What's that got to do with this PR?

roidelapluie: I need to alter the 'res' vector to look for common labels

brian-brazil: I'm not seeing it, ContainsSameLabelset doesn't alter the vector.

roidelapluie: It does not but we must have a vector to run it against.

brian-brazil: Ah, I see what you're up to now.

rules/alerting_test.go
```diff
@@ -289,3 +292,38 @@ func TestAlertingRuleEmptyLabelFromTemplate(t *testing.T) {
...
+		expr,
+		time.Minute,
+		labels.FromStrings("test", "test"),
+		nil,
+		nil,
+		true, log.NewNopLogger(),
+	)
+	_, err := rule.Eval(ctx, now, EngineQueryFunc(engine, storage), nil)
+	testutil.NotOk(t, err)
+	fmt.Println(err.Error())
```
brian-brazil: This line should go.

roidelapluie: updated

rules/alerting.go
```diff
@@ -404,6 +407,15 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc,
 		}
 	}
 
+	// Check that the alert does not produce identical metrics after applying
```
brian-brazil: What happens to h.active etc. above? We should probably do this check before all that, so this has the same effect as a query failure.

roidelapluie: @brian-brazil in that case I completely changed the way we check for duplicate alerts

rules/alerting.go
```diff
@@ -361,15 +362,16 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc,
...
 		resultFPs[h] = struct{}{}
 
-		// Check whether we already have alerting state for the identifying label set.
-		// Update the last value and annotations if so, create a new alert entry otherwise.
-		if alert, ok := r.active[h]; ok && alert.State != StateInactive {
-			alert.Value = smpl.V
-			alert.Annotations = annotations
-			continue
+		if _, ok := alerts[h]; ok {
+			err = fmt.Errorf("vector contains metrics with the same labelset after applying rule/alert labels")
```
brian-brazil: we're in the alert file, we know it's an alert

roidelapluie: fixed

brian-brazil:
Thanks!

PR #6471, opened 2019-12-17
manager: Improve TestPopulateLabels test

roidelapluie:
testutil.ErrorEqual just returns a bool and needs to be wrapped into an
Assertion.


brian-brazil:
This is a bit unintuitive, should we change how ErrorEqual works?

roidelapluie:
updated

brian-brazil:
Thanks!

Issue #6470, opened 2019-12-17
PromQL: Consider supporting trailing commas in grouping opts


slrtbtfs:
Currently PromQL allows trailing commas in label matchers, but not in the label list of grouping opts.

E.g. `a{b="c",}`,  `a{b="c"}` and `sum (a) by (b)` and are valid PromQL, while `sum (a) by (b,)` is not.

This feels somewhat inconsistent.

It might be worth considering to allow trailing commas in grouping opts, too.

brian-brazil:
This seems reasonable to me.

@tomwilkie any thoughts from the jsonnet side of the house?

beorn7:
Sounds reasonable to me. I can't see how Jsonnet would have a problem with this more relaxed syntax.

brian-brazil:
I can't imagine either, but I could see them appreciating the feature.

beorn7:
Yes, definitely. Let's do it!

slrtbtfs:
Ok, I'll do a PR.

slrtbtfs:
When we're already at it:

We might want to the same thing for functions, too.

brian-brazil:
There's only a small number of functions that take a variable number of arguments, I'm not seeing the need there.

slrtbtfs:
Doing it for functions, too, would make all comma separated lists behave the same, which has some appeal.

brian-brazil:
Functions arguments aren't somewhere where you build up lists of an unknown number of things from disparate sources.

slrtbtfs:
True.

I'm fine with leaving functions in their current state as well.

PR #6472, opened 2019-12-17
PromQL: Parse grouping opts with the generated parser

slrtbtfs:
This PR extends the generated parser to parse grouping opts.

This turned out to be more complex than expected since the lexer will return keyword tokens for labels with names that are also keywords.

To work around this, an additional `maybe_label` grammar rule is introduced, that matches on any token that could be a label.

Based on this PR, implementing #6470 will be a two line change (excluding generated code).

This is part of the parser rewrite effort described in #6410 .

### DIFF
--- a/promql/lex.go
+++ b/promql/lex.go
@@ -123,6 +123,10 @@ func (i ItemType) isRightAssociative() bool {
 
 type ItemType int
 
+// This is a list of all keywords in PromQL.
+// When changing this list, make sure to also change
+// the maybe_label grammar rule in the generated parser
+// to avoid misinterpretation of labels as keywords.
 var key = map[string]ItemType{
 	// Operators.
 	"and":    LAND,
--- a/promql/generated_parser.y
+++ b/promql/generated_parser.y
@@ -28,6 +28,7 @@
     matcher   *labels.Matcher
     label     labels.Label
     labels    labels.Labels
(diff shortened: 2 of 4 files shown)


promql/lex.go
```diff
@@ -123,6 +123,10 @@ func (i ItemType) isRightAssociative() bool {
 
 type ItemType int
 
+// This is a list of all keywords in PromQL
+// When changing this list, make sure to also change
+// the maybe_label grammar rule in the generated parser
+// to avoid misinterpretation of labels as keywords
```
brian-brazil: Full stop.

slrtbtfs: Fixed it.

Given how often this happens to me, I maybe should write some script to check for this specific error automatically.

slrtbtfs:
> I take it you're leaving the commas for a future PR?

Yes, as an immediate follow up.

Thought it would be nicer to have the implementation and behavior changes in separate PRs.

brian-brazil:
:+1: 

CI seems a bit borked.

brian-brazil:
Can you try making a trivial change and force pushing?

slrtbtfs:
Adding an empty commit to trigger the CI worked.

brian-brazil:
Thanks!

PR #6475, opened 2019-12-18
Constant memory compaction

brian-brazil:
Variable compaction memory usage is down to .25 bytes per symbol, plus a few O(labelnames) data structures.

### DIFF
--- a/tsdb/index/index.go
+++ b/tsdb/index/index.go
@@ -111,16 +111,14 @@ func newCRC32() hash.Hash32 {
 // Writer implements the IndexWriter interface for the standard
 // serialization format.
 type Writer struct {
-	ctx  context.Context
-	f    *os.File
-	fbuf *bufio.Writer
-	pos  uint64
+	ctx context.Context
+
+	// For the main index file.
+	f *fileWriter
 
 	// Temporary file for posting offsets table.
-	fPO    *os.File
-	fbufPO *bufio.Writer
-	posPO  uint64
-	cntPO  uint64
+	fPO   *fileWriter
+	cntPO uint64
 
 	toc   TOC
 	stage indexWriterStage
@@ -200,12 +198,12 @@ func NewWriter(ctx context.Context, fn string) (*Writer, error) {
 	}
 
 	// Main index file we are building.
-	f, err := os.OpenFile(fn, os.O_CREATE|os.O_RDWR, 0666)
+	f, err := newFileWriter(fn)
 	if err != nil {
 		return nil, err
 	}
 	// Temporary file for posting offset table.
-	fPO, err := os.OpenFile(fn+"_tmp_po", os.O_CREATE|os.O_RDWR, 0666)
+	fPO, err := newFileWriter(fn + "_tmp_po")
 	if err != nil {
 		return nil, err
 	}
@@ -214,14 +212,10 @@ func NewWriter(ctx context.Context, fn string) (*Writer, error) {
 	}
 
 	iw := &Writer{
-		ctx:    ctx,
-		f:      f,
-		fbuf:   bufio.NewWriterSize(f, 1<<22),
-		pos:    0,
-		fPO:    fPO,
-		fbufPO: bufio.NewWriterSize(fPO, 1<<22),
-		posPO:  0,
-		stage:  idxStageNone,
+		ctx:   ctx,
+		f:     f,
+		fPO:   fPO,
+		stage: idxStageNone,
 
 		// Reusable memory.
 		buf1: encoding.Encbuf{B: make([]byte, 0, 1<<22)},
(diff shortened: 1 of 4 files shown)


tsdb/index/index.go
```diff
@@ -270,37 +331,51 @@ func (w *Writer) ensureStage(s indexWriterStage) error {
 	if w.stage == s {
 		return nil
 	}
+	if w.stage < s-1 {
+		// A stage has been skipped.
+		if err := w.ensureStage(s - 1); err != nil {
```
bwplotka: Am I right that no symbols, no series is the only case when this happens? 

I know we check later:

```
// Populated block is empty, so exit early.
	if meta.Stats.NumSamples == 0 {
		return nil
	}

```

Curious what's changed here in this area: e.g why suddenly we have to add this? Was it a bug that was there all the time? (empty index was not a valid index before?)

tsdb/index/index.go
```diff
@@ -270,37 +331,51 @@ func (w *Writer) ensureStage(s indexWriterStage) error {
...
-	case idxStageLabelIndex:
-		w.toc.LabelIndices = w.pos
+		w.toc.Series = w.f.pos
 
 	case idxStageDone:
-		w.toc.Postings = w.pos
+		w.toc.LabelIndices = w.f.pos
+		// LabelIndices generation depends on the posting offset
+		// table produced at this stage.
+		if err := w.writePostingsToTmpFiles(); err != nil {
```
bwplotka: Nice, the ordering of sections is unfortunate indeed (: Since we have TOC file... is changing the order really a breaking change? ;p

bwplotka: Oh yes, we depend on it multiple times on reader even now, detecting how much bytes we should scan (:

tsdb/index/index.go
```diff
@@ -457,53 +531,88 @@ func (w *Writer) finishSymbols() error {
...
 	// Leave 4 bytes of space for the length, which will be calculated later.
 	if err := w.write([]byte("alen")); err != nil {
 		return err
 	}
 	w.crc32.Reset()
 
 	w.buf1.Reset()
-	w.buf1.PutBE32int(len(names))
-	w.buf1.PutBE32int(valt.Len())
+	w.buf1.PutBE32int(1) // Number of names.
```
bwplotka: We know block composite index from our interface, is it something we want? We never used it, so I think I agree with this, just checking (:

tsdb/index/index.go
```diff
@@ -647,9 +786,10 @@ func (w *Writer) writePostings() error {
 		}
 		offsets = append(offsets, uint32(startPos/16))
 		// Skip to next series.
-		d.Skip(d.Uvarint() + crc32.Size)
+		x := d.Uvarint()
+		d.Skip(x + crc32.Size)
 		if err := d.Err(); err != nil {
-			return nil
```
bwplotka: :scream:

tsdb/index/index.go
```diff
@@ -796,13 +970,17 @@ func (w *Writer) Close() error {
 			return err
 		}
 	}
-	if err := w.fbuf.Flush(); err != nil {
-		return err
+	if w.fP != nil {
+		if err := w.fP.close(); err != nil {
```
bwplotka: Not remove? and below?

tsdb/index/index.go
```diff
@@ -223,39 +234,89 @@ func NewWriter(ctx context.Context, fn string) (*Writer, error) {
...
 		if err != nil {
 			return err
 		}
 		// For now the index file must not grow beyond 64GiB. Some of the fixed-sized
 		// offset references in v1 are only 4 bytes large.
 		// Once we move to compressed/varint representations in those areas, this limitation
 		// can be lifted.
-		if w.pos > 16*math.MaxUint32 {
+		if fw.pos > 16*math.MaxUint32 {
 			return errors.Errorf("exceeding max size of 64GiB")
```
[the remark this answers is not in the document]

tsdb/index/index.go
brian-brazil: I've had to do a few workarounds to the tests as I've been making these changes to prevent this. There's one unittest specifically testing for this.

tsdb/index/index.go
brian-brazil: It's all dead code as far as I'm concerned, but that's for another PR.

tsdb/index/index.go
brian-brazil: The only time it could be removed here is if something went wrong, in which case leaving it behind for debugging with the incomplete main index file seems better than adding more complex logic to remove it.

brian-brazil:
> Buffer names and offset for LabelIndicesOffset: This is where O(names) cames from (:

That's one of the places, however as we already have O(names) already and n is small I don't think it's worth worrying about.

tsdb/index/index.go
brian-brazil: The good news is it only happened when you're in the middle of debugging a broken rewrite of an index writer :)

tsdb/index/index.go
bwplotka: ok

tsdb/index/index.go
bwplotka: Yes, but still part of format

PR #6477, opened 2019-12-18
Refactor WAL.Segments method to be part of the wal package

codesome:
It makes projects like Cortex consume the methods easily. I hope this small change looks fine.

It would be great if it can make it into `v2.15`. /cc @bwplotka

### DIFF
--- a/tsdb/wal/wal.go
+++ b/tsdb/wal/wal.go
@@ -260,7 +260,7 @@ func NewSize(logger log.Logger, reg prometheus.Registerer, dir string, segmentSi
 	}
 	w.metrics = newWALMetrics(w, reg)
 
-	_, last, err := w.Segments()
+	_, last, err := Segments(w.Dir())
 	if err != nil {
 		return nil, errors.Wrap(err, "get segment range")
 	}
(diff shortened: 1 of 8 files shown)


tsdb/wal/wal.go
```diff
@@ -673,8 +673,8 @@ func (w *WAL) log(rec []byte, final bool) error {
 
 // Segments returns the range [first, n] of currently existing segments.
 // If no segments are found, first and n are -1.
-func (w *WAL) Segments() (first, last int, err error) {
```
roidelapluie: I think that this function should now be reordered

tsdb/wal/wal.go
```diff
@@ -260,7 +260,7 @@ func NewSize(logger log.Logger, reg prometheus.Registerer, dir string, segmentSi
 	}
 	w.metrics = newWALMetrics(w, reg)
 
-	_, last, err := w.Segments()
+	_, last, err := Segments(w.dir)
```
roidelapluie: w.Dir()

roidelapluie:
2 comments, otherwise LGTM

brian-brazil:
@codesome Is this still relevant?

codesome:
Yes. I will fix the reviews soon and lets get this merged.

tsdb/wal/wal.go
codesome: @roidelapluie What do you mean by reordering? Move it within the file?

roidelapluie: Yes, it is in the middle of the func (w *WAL)  functions.

codesome:
Merging this on green

PR #6481, opened 2019-12-18
PromQL: Refactor error message generation

slrtbtfs:
This PR introduces a new parser method for generating error messages about unexpected items from the lexer.

Using this method in the generated parser makes the code easier to read and removes the necessity to type the same error messages again and again.


brian-brazil:
Thanks!

PR #6482, opened 2019-12-18
Remove redundant/accidental line in query doc

tjanson:
Removing that line makes it consistent with the other functions.


brian-brazil:
Thanks!

PR #6487, opened 2019-12-19
ui: Display unknown and failing alerts

roidelapluie:

prom

### DIFF
--- a/web/ui/react-app/src/pages/alerts/AlertContents.tsx
+++ b/web/ui/react-app/src/pages/alerts/AlertContents.tsx
@@ -10,7 +10,7 @@ export interface Rule {
   alerts: Alert[];
   annotations: Record<string, string>;
   duration: number;
-  health: string;
+  health: RuleState;
   labels: Record<string, string>;
   name: string;
   query: string;
@@ -22,6 +22,9 @@ export interface RuleStatus<T> {
   firing: T;
   pending: T;
   inactive: T;
+  unknown: T;
+  err: T;
+  ok: T;
 }
 
 export interface AlertsProps {
@@ -35,6 +38,7 @@ interface Alert {
   value: string;
   annotations: Record<string, string>;
   activeAt: string;
+  health: string;
 }
 
 interface RuleGroup {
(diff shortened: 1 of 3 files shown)


web/ui/react-app/src/pages/alerts/AlertContents.tsx
```diff
@@ -10,7 +10,7 @@ export interface Rule {
   alerts: Alert[];
   annotations: Record<string, string>;
   duration: number;
-  health: string;
+  health: RuleState;
```
juliusv: I'd prefer if we could only allow the actually valid values for `health` and `state` here. Maybe just get rid of the `export type RuleState = keyof RuleStatus<any>;` and do a direct `health: 'firing' | 'pending' | 'inactive'` here and `state: 'unknown' | 'err' | 'ok'` below? The alternative would be to have two separate enums / interfaces at the beginning and then joining them into one later on somehow.

juliusv:
Sorry btw., I totally didn't see this. Best to @ me on UI PRs :)

web/ui/react-app/src/pages/alerts/AlertContents.tsx
```diff
@@ -49,6 +53,9 @@ const AlertsContent: FC<AlertsProps> = ({ groups = [], statsCount }) => {
     firing: true,
     pending: true,
     inactive: true,
+    unknown: false,
+    err: false,
+    ok: false,
```
juliusv: Just so I understand correctly, you are introducing three new rule states here, and we have show/hide buttons for the original inactive/pending/firing, but not for the new ones, and it seems the new ones are always shown, with no option of filtering them. I also don't know if we should be adding three new buttons, but curious what your thinking is here.

juliusv:
@roidelapluie Heya, just want to make sure, what's the background for closing this, was it intentional?

roidelapluie:
Yes that means that I stop working on it at the moment and that I think my first approach is not good enough :)

juliusv:
Ah ok, thanks for the clarification!

PR #5223, opened 2019-02-14
Add flag for syntax-only config file checking in promtool

glightfoot:
Fixes #5222 by adding an optional --syntax-only flag to the check config command in promtool. Using this flag skips everything after checking to see if the file can be parsed by config.LoadFile(filename)

@brian-brazil

Issue #5222, opened 2019-02-14
Make promtool better for CI with syntax-only config check


glightfoot:
## Proposal
**Use case. Why is this important?**

*We validate our rules with promtool in CI before deploying, it would be nice to be able to check for config file syntax validity without having to deploy all secrets, certs, etc.*

## Bug Report
**What did you do?**
```promtool check config prometheus.yaml```

**What did you expect to see?**
```
./promtool check config --syntax-only prometheus.yaml                                                                                                                                                                                
Checking prometheus.yaml
  SUCCESS: prometheus.yaml is valid prometheus config file syntax
  SUCCESS: 0 rule files found
```
**What did you see instead? Under which circumstances?**
```
promtool check config prometheus.yaml
Checking prometheus.yaml
  FAILED: error checking bearer token file "/var/run/secrets/kubernetes.io/serviceaccount/token": stat /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory
```

**Environment**
CI (explicitly does not have the token and certs present in the deployment)

* System information:

	Linux 3.10.0-514.26.2.el7.x86_64 x86_64

* Prometheus version:

	2.7.1

roidelapluie:
I do not agree with this. Those files are part of the configuration and I want promtool to check that it can actually read them.

glightfoot:
@roidelapluie - this will not change that behavior. I agree that that is an important check to have available. This will only add the ability to do the tests without those files as a new feature.

`promtool check config prometheus.yaml` - still works as it does now
`promtool check config --syntax-only prometheus.yaml` - only checks syntax

geekodour:
@codesome can I add this with https://github.com/prometheus/prometheus/issues/21 ?

glightfoot:
Is there any way we can move this along? I had syntax bug get checked in that we could have prevented with this...

andbuitra:
I just had this when doing CI on GitLab but it was related to the alertmanager rules using an absolute path on the monitoring server that throws off promtool on the GitLab server. Syntax only checking would be greatly appreciated!

glightfoot:
I opened a new PR (#6488 ) that mirrors what I built locally to fix this. On the last PR, there was concern that not all files were being checked, so if that is still a problem, then I am more than happy to open another issue and fix there.

dt-rush:
Until something is merged to handle this, `promtool check config` is hardly usable unless you only run it in the same environment you deploy to. The tool itself describes `check config` as:

[...]

PR #6488, opened 2019-12-19
[check-syntax] adds a command for checking config file syntax

glightfoot:
In order to make testing syntax in CI easier, we need a way to check syntax without checking if all the files exist. This PR adds a `check syntax` command to promtool that just loads the config from files and returns success if they load fine, or the error if the config is not valid. This replaces #5223 and is the fix for #5222 

Signed-off-by: glightfoot <glightfoot@rsglab.com>

### DIFF
--- a/cmd/promtool/main.go
+++ b/cmd/promtool/main.go
@@ -56,6 +56,12 @@ func main() {
 		"The config files to check.",
 	).Required().ExistingFiles()
 
+	checkSyntaxCmd := checkCmd.Command("syntax", "Check if the config files have valid syntax or not. Does not check if files exist.")
+	syntaxFiles := checkSyntaxCmd.Arg(
+		"config-files",
+		"The config files to check syntax.",
+	).Required().ExistingFiles()
+
 	checkRulesCmd := checkCmd.Command("rules", "Check if the rule files are valid or not.")
 	ruleFiles := checkRulesCmd.Arg(
 		"rule-files",
@@ -117,6 +123,9 @@ func main() {
 	case checkConfigCmd.FullCommand():
 		os.Exit(CheckConfig(*configFiles...))
 
+	case checkSyntaxCmd.FullCommand():
+		os.Exit(CheckSyntax(*syntaxFiles...))
+
 	case checkRulesCmd.FullCommand():
 		os.Exit(CheckRules(*ruleFiles...))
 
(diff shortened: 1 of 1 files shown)


glightfoot:
@simonpasquier

simonpasquier:
Thanks, you'll probably have to wait until next year before I give it a review :smirk:

glightfoot:
Thanks! Not sure if this is the preferred approach or not, but it's definitely useful (we've been building our own promtool with this for a little while)

glightfoot:
@simonpasquier - Can you take a look at this?

dt-rush:
@simonpasquier is there anything preventing us from merging this? As it is, `check config` is actually unusable in CI for non-trivial config files (any use of service-discovery features), and there are several threads around the net where people are running into this same confusion about the meaning of `check config`. I've just run into this myself now. The PR here is minimal, clear in its function, and adds a ton of value.

dt-rush:
For anyone arriving here in between now and when this is fixed, if you use k8s service discovery, I've got things working by faking the file `/var/run/secrets/kubernetes.io/serviceaccount/token`. Here's a cutting from my CI test:

```bash
logmsg "faking deployment env..."
# patch the container to fake being in the deployment environment
# see:
# - https://github.com/prometheus/prometheus/issues/5222
# - https://github.com/prometheus/prometheus/pull/6488
docker exec -i $monitor_id bash <<-EOF
    mkdir -p /var/run/secrets/kubernetes.io/serviceaccount/token
    touch /var/run/secrets/kubernetes.io/serviceaccount/token
EOF
logmsg "✔ faked deployment env."
# check config / rules using promtool
logmsg "using promtool to inspect generated yml's..."
set +e
docker exec $container /bin/sh -c 'promtool check config /etc/prometheus/prometheus.yml || exit $?'
exitcode=$?
set -e
# allow user to inspect files if not passing
if [ "$exitcode" != "0" ]; then
    inspect_command="vim /etc/prometheus/*.yml" && fail "promtool check config failed"
fi
logmsg "✔ promtool check config succeeded"
logmsg "🌱 success!"
```

drcrees:
@simonpasquier I just ran into this as well.  Any idea when this will be pushed through? 👀

roidelapluie:
check config and check syntax might be confusing for users. what about making this a flag of check config?

glightfoot:
@roidelapluie I had that as a pr previously https://github.com/prometheus/prometheus/pull/5223. If that's what you're looking for, I can make similar changes here

glightfoot:
@roidelapluie - any thoughts on the pr mentioned in my previous comment?

brian-brazil:
With recent changes to SD this is now plausible to implement in a maintainable fashion, however requires paying off some technical debt first. All validation of HTTP etc. settings should move down to common, with the file stuff being in a Validate method. Then this could be implemented by unmarshalling without calling Validate.

mal-berbatovci:
Any update on this? Would be nice to have

PR #6491, opened 2019-12-19
promql: fix typo in value.go

roidelapluie:

### DIFF
--- a/promql/value.go
+++ b/promql/value.go
@@ -263,7 +263,7 @@ type StorageSeries struct {
 	series Series
 }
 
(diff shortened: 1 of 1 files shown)


promql/value.go
```diff
@@ -263,7 +263,7 @@ type StorageSeries struct {
 	series Series
 }
 
-// NewStorageSeries returns a StorageSeries fromfor series.
+// NewStorageSeries returns a StorageSeries from series.
```
brian-brazil: from a ?

roidelapluie: changed + capitalized

brian-brazil:
Thanks!

PR #6494, opened 2019-12-20
PromQL: Parse Series descriptions using the generated parser

slrtbtfs:
This PR extends the generated parser (#6256) to parse series descriptions.

The series description format is quite complex, so this PR is relatively large.

However, since series descriptions are only ever parsed inside unit test, there is no risk of breaking any production usage of the parser this time.

### DIFF
--- a/promql/generated_parser.y
+++ b/promql/generated_parser.y
@@ -312,6 +312,8 @@ series_values   :
                         { $$ = append($1, $3...) }
                 | series_values SPACE
                         { $$ = $1 }
+                | error
+                        { yylex.(*parser).unexpected("series values", "") }
                 ;
 
 series_item     :
--- a/promql/generated_parser.y
+++ b/promql/generated_parser.y
@@ -143,7 +143,7 @@ start           : START_LABELS label_matchers
                      { yylex.(*parser).generatedParserResult = $2 }
                 | START_GROUPING_LABELS grouping_labels
                      { yylex.(*parser).generatedParserResult = $2 }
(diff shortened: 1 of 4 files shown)


promql/generated_parser.y
```diff
@@ -15,9 +15,12 @@
     package promql
 
     import (
+        "math"
         "sort"
+        "strconv"
 
         "github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/pkg/value"
```
brian-brazil: Something weird with formatting here

promql/generated_parser.y
```diff
@@ -181,8 +190,8 @@ metric          :
                         { $$ = append($2, labels.Label{Name: labels.MetricName, Value: $1.Val}); sort.Sort($$) }
                 | label_set 
                         {$$ = $1}
-                | error
-                        { yylex.(*parser).errorf("missing metric name or metric selector")}
+              /*| error
```
brian-brazil: Debug code?

promql/generated_parser.y
```diff
@@ -283,4 +292,91 @@ maybe_label     :
                 | BOOL
                 ;
 
+series_description:
```
brian-brazil: A comment that these are only used in promql tests would be good.

promql/generated_parser.y
slrtbtfs: Mixed spaces and tabs.

promql/generated_parser.y
slrtbtfs: Kind of.

slrtbtfs:
Did all the things suggested in the review.

brian-brazil:
Thanks!

PR #6501, opened 2019-12-20
promql: function: fix typo

roidelapluie:


brian-brazil:
Thanks!

It's okay to put tiny things like this into a bigger change, rather than lots of small PRs.

Issue #6397, opened 2019-11-30
Provide default URL scheme for  promtool server addresses


grobie:
## Proposal
**Use case. Why is this important?**

```console
$ promtool query instant prometheus.example.org prometheus_build_info
prometheus_build_info{version="future", ...}
```

Ease of CLI usage. Similar to curl, it would be nice to handle server addresses without an explicit URL scheme gracefully. From a security perspective it would be nice to try `https` first and falling back to `http` otherwise.

*“Nice to have” is not a good use case. :)*

Alright, sorry. Gonna try regardless :upside_down_face:

roidelapluie:
this is super weird.

when to fall back to http? when tcp says no? when certificate is not recognized? how does the user know which proto was picked?

brian-brazil:
Yeah, http is what is expected here with no attempts to fallback.

AdaephonBen:
I'm new to Prometheus and I would like to be assigned to this issue. Just clarifying, all that is required is to prefix http:// to the URL argument if not already present?

brian-brazil:
Yes, add a scheme if there isn't one.

peter-turner-hcl:
Is it free to grab this one? @AdaephonBen

AdaephonBen:
I guess this can be closed, @roidelapluie ?

roidelapluie:
Yes, thanks !

PR #6502, opened 2019-12-21
add check if server url containing http, if not add it

ronshuli:

### DIFF
--- a/cmd/promtool/main.go
+++ b/cmd/promtool/main.go
@@ -126,7 +126,7 @@ func main() {
 
 	case queryInstantCmd.FullCommand():
 		{
-			if !strings.Contains(*queryServer, "http") {
+			if !strings.HasPrefix(*queryServer, "http://") && !strings.HasPrefix(*queryServer, "https://") {
 				*queryServer = "http://" + *queryServer
 			}
 			os.Exit(QueryInstant(*queryServer, *queryExpr, p))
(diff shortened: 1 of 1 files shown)


ronshuli:
Hi this is my first PR,
I fixed this bug https://github.com/prometheus/prometheus/issues/6397
thanks,

roidelapluie:
Thank you for your pull request @ronshuli.

This does not seem to be the right fix however. I think we should do that earlier, at the point where we parse the parameter, so that it will benefit all the commands.

Ping me if you need more assitance to move forward, or when you have updated your pull request, and once again, welcome on the Prometheus project!

ronshuli:
Hello @roidelapluie,
I think I fixed the PR according what you asked,
I would love that you will check this once again and thank for notice.

I'm happy to join Prometheus project :)

roidelapluie:
I will review in a few days ; however I think the approach should be to parse the URL and see if there is a protocol ; instead of checking for a string. I will come back to you :)

Also, did you test this? I was under the impression that we check the URL when we parse the arguments, using prometheus/common

cmd/promtool/main.go
```diff
@@ -124,7 +125,12 @@ func main() {
 		os.Exit(CheckMetrics())
 
 	case queryInstantCmd.FullCommand():
-		os.Exit(QueryInstant(*queryServer, *queryExpr, p))
+		{
+			if !strings.Contains(*queryServer, "http") {
```
codesome: I am not aware of the placement of this check or any other ways, but if we are going with string matching, then I think only the prefix should be checked for `http://` and `https://` and not the entire URL for the existence of `http`.

ronshuli: Hi, thanks for your comment, I change it :)

roidelapluie:
hello,

Somehow this has been addressed in the meantime: #7956

Thanks for your work and sorry this went un-noticed!

Issue #6385, opened 2019-11-27
React UI: Create /rules page


juliusv:
The React UI is still missing the [/rules](http://demo.robustperception.io:9090/rules) page - this still needs to be implemented (and server-side APIs extended as necessary).

Harkishen-Singh:
@juliusv I will like to take up this.

boyskila:
@Harkishen-Singh are you still interested in working on this?

Harkishen-Singh:
@boyskila yes sure, I am still interested. I am in the middle of academic exams so got delayed. I will do ASAP.
Thanks.

juliusv:
In case @boyskila would want to work on it right now, that shouldn't block you IMO if work hasn't started on it and it's still blocked on @Harkishen-Singh's side for some time.

Issue #6302, opened 2019-11-11
React UI: Figure out how to handle "Xs ago" duration displays


juliusv:
Some of the pages in the classic UI (like /targets and /rules) show durations in the format "3.45s ago". In the new React UI, we get only absolute timestamps of events from the server, so if we wanted to compute such a relative duration, we need a relative "now" timestamp to subtract them from. The current approach is to use the browser's current timestamp, but the user's clock might be totally off with regards to the server's. We could instead think about using the `Date` header that the Prometheus server returns on every API request, and which is at least precise to the second (although not lower).

kimuraz:
@juliusv can I work on this issue?

juliusv:
@kimuraz Sure. First thing is to figure out what we want to do though. Via date header or some other way? I don't like any of the options, but maybe the Date header is the least bad?

kimuraz:
@juliusv I did some research but so far I guess the Date header is the "least bad" option indeed, I guess it will guarantee that the calculation is correct since it will have the same reference. If that's the way, what is the Definition of Done for this issue? I mean, what should be delivered in this case?

juliusv:
@kimuraz Yes, I'm thinking the same. Another option would be to replace the "Xs ago" displays with absolute timestamps, but I feel that the relative displays are much more useful for the reader.

So I would say, go for the Date header option.

> what is the Definition of Done for this issue? I mean, what should be delivered in this case?

A pull request that changes the existing React UI code to make available the `Date` header timestamp for relative time computations by the component that fetches data, and then use that on the `/new/targets` and `/new/rules` pages (the latter doesn't exist yet in the React UI). Currently the targets page uses a client-side `now()` timestamp as a reference, which would need to be replaced: https://github.com/prometheus/prometheus/blob/06b47447938a9017e7d1c2ad24df050fd61964df/web/ui/react-app/src/pages/targets/ScrapePoolPanel.tsx#L82

PR #6503, opened 2019-12-21
Adds support for rules screen in react-ui

Harkishen-Singh:
Fixes: #6385
Fixes: #6302  
Initial-version

preview-rules-page-gif

Update-1:
preview-rules-update-1-gif


Harkishen-Singh:
@juliusv @boyskila please have a look.

web/api/v1/api.go
```diff
@@ -933,10 +937,12 @@ func (api *API) rules(r *http.Request) apiFuncResult {
...
 		apiRuleGroup := &RuleGroup{
-			Name:     grp.Name(),
-			File:     grp.File(),
-			Interval: grp.Interval().Seconds(),
-			Rules:    []rule{},
+			Name:           grp.Name(),
+			File:           grp.File(),
+			Interval:       grp.Interval().Seconds(),
+			Rules:          []rule{},
+			EvaluationTime: grp.GetEvaluationDuration().Microseconds(),
```
juliusv: We keep these durations in seconds, so let's do:

```suggestion
			EvaluationTime: grp.GetEvaluationDuration().Seconds(),
```

web/api/v1/api.go
```diff
@@ -967,13 +973,16 @@ func (api *API) rules(r *http.Request) apiFuncResult {
...
-					Labels:    rule.Labels(),
-					Health:    rule.Health(),
-					LastError: lastError,
-					Type:      "recording",
+					Name:           rule.Name(),
+					Query:          rule.Query().String(),
+					Labels:         rule.Labels(),
+					Health:         rule.Health(),
+					LastError:      lastError,
+					EvaluationTime: rule.GetEvaluationDuration().Microseconds(),
```
juliusv: Same here:
```suggestion
					EvaluationTime: rule.GetEvaluationDuration().Seconds(),
```

web/api/v1/api.go
```diff
@@ -967,13 +973,16 @@ func (api *API) rules(r *http.Request) apiFuncResult {
...
+					LastEvaluation: time.Since(rule.GetEvaluationTimestamp()).String(),
+					Type:           "recording",
 				}
+				fmt.Println(enrichedRule)
```
juliusv: Remove left-over debug statement.

web/web.go
```diff
@@ -849,6 +849,8 @@ func (h *Handler) serveConfig(w http.ResponseWriter, r *http.Request) {
 }
 
 func (h *Handler) rules(w http.ResponseWriter, r *http.Request) {
+	fmt.Println(("rules below"))
+	fmt.Println(h.ruleManager.RuleGroups()[0])
```
juliusv: Remove left-over debug print statements.

web/ui/react-app/src/pages/Rules.tsx
```diff
@@ -1,15 +1,108 @@
...
-    </Alert>
-  </>
-);
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faSpinner } from '@fortawesome/free-solid-svg-icons';
+import { Alert, Table } from 'reactstrap';
+import { useFetch } from '../utils/useFetch';
+import { Link } from '@reach/router';
+
+interface Rule {
```
juliusv: There are already rule-API-related types defined here: https://github.com/prometheus/prometheus/blob/ae93bae88fed884f59adef604ea059deff81650b/web/ui/react-app/src/pages/alerts/AlertContents.tsx#L9-L45

...but currently they are pretty alerts-specific, as that is used only on the /alerts page, that filters only by alerting rules (https://github.com/prometheus/prometheus/blob/ae93bae88fed884f59adef604ea059deff81650b/web/ui/react-app/src/pages/alerts/Alerts.tsx#L11).

The best thing to do would be to refactor these types into a common location (probably within https://github.com/prometheus/prometheus/tree/master/web/ui/react-app/src/types) where they support both types of rules properly, but are defined only once.

web/ui/react-app/src/pages/Rules.tsx
```diff
@@ -1,15 +1,108 @@
...
+  name: string;
+  query: string;
+  health: string;
+  evaluationTime: string;
+  lastEvaluation: string;
+  type: string;
+  lastError: string;
+}
+
+interface EnrichRule {
```
juliusv: This does not seem to refer to an enriched rule, but to a group of rules, so it should probably be:

```suggestion
interface RuleGroup {
```

web/api/v1/api.go
```diff
@@ -909,11 +911,13 @@ type alertingRule struct {
 }
```
juliusv: The `alertingRule` type also needs the new `EvaluationDuration` and `LastEvaluation` fields (and have them set below).

web/api/v1/api.go
```diff
@@ -909,11 +911,13 @@ type alertingRule struct {
...
-	Query     string           `json:"query"`
-	Labels    labels.Labels    `json:"labels,omitempty"`
-	Health    rules.RuleHealth `json:"health"`
-	LastError string           `json:"lastError,omitempty"`
+	Name           string           `json:"name"`
+	Query          string           `json:"query"`
+	Labels         labels.Labels    `json:"labels,omitempty"`
+	Health         rules.RuleHealth `json:"health"`
+	LastError      string           `json:"lastError"`
+	EvaluationTime int64            `json:"evaluationTime"`
```
juliusv: We typically use "...Time" for specific timestamps, not for durations, and since we want to export in seconds, we use a float64 for e.g. the target scrape duration. So better would be:

```suggestion
	EvaluationDuration float64            `json:"evaluationDuration"`
```

web/api/v1/api.go
```diff
@@ -933,10 +937,12 @@ func (api *API) rules(r *http.Request) apiFuncResult {
...
-			Name:     grp.Name(),
-			File:     grp.File(),
-			Interval: grp.Interval().Seconds(),
-			Rules:    []rule{},
+			Name:           grp.Name(),
+			File:           grp.File(),
+			Interval:       grp.Interval().Seconds(),
+			Rules:          []rule{},
+			EvaluationTime: grp.GetEvaluationDuration().Microseconds(),
+			LastEvaluation: time.Since(grp.GetEvaluationTimestamp()).String(),
```
juliusv: We'd want to avoid exporting relative durations since the last evaluation here, but just the absolute last evaluation timestamp (like we do for scrapes of targets):

```suggestion
			LastEvaluation: grp.GetEvaluationTimestamp(),
```

...and then make the `LastEvaluation` field a `time.Time` type instead of `string`. Take a look at `LastScrape` and `LastScrapeDuration` for targets, we'll want to use the same existing pattern for rules.

web/api/v1/api.go
```diff
@@ -909,11 +911,13 @@ type alertingRule struct {
...
-	Name      string           `json:"name"`
-	Query     string           `json:"query"`
-	Labels    labels.Labels    `json:"labels,omitempty"`
-	Health    rules.RuleHealth `json:"health"`
-	LastError string           `json:"lastError,omitempty"`
+	Name           string           `json:"name"`
+	Query          string           `json:"query"`
+	Labels         labels.Labels    `json:"labels,omitempty"`
+	Health         rules.RuleHealth `json:"health"`
+	LastError      string           `json:"lastError"`
```
juliusv: You removed the `omitempty` here, but not on the `alertingRule`. Currently if I have any alerting rules configured, the page crashes because it's trying to access this field, but it doesn't exist.

Might be actually best to keep the `omitempty` in both cases, but make the field optional in the TypeScript interface and check for its existence.

web/ui/react-app/src/pages/Rules.tsx
```diff
@@ -1,15 +1,108 @@
...
+            <Table size="sm" striped={true} bordered={true} key={i}>
+              <thead>
+                <tr>
+                  <td colSpan={3}>
+                    <a href={'#' + g.name}>
+                      <h2>{g.name}</h2>
+                    </a>
+                  </td>
+                  <td>
+                    <h2>{roundUp(parseFloat(g.lastEvaluation))}s ago</h2>
```
juliusv: After converting the last evaluation to an absolute timestamp, take a look at how the targets page calculates a relative duration, using `formatRelative()`: https://github.com/prometheus/prometheus/blob/ae93bae88fed884f59adef604ea059deff81650b/web/ui/react-app/src/pages/targets/ScrapePoolPanel.tsx#L82

web/ui/react-app/src/pages/Rules.tsx
```diff
@@ -1,15 +1,108 @@
...
+                  </td>
+                  <td>
+                    <h2>{g.evaluationTime}us</h2>
```
juliusv: After converting (and renaming) the evaluation duration to seconds, take a look at how the targets page formats the duration string: https://github.com/prometheus/prometheus/blob/ae93bae88fed884f59adef604ea059deff81650b/web/ui/react-app/src/pages/targets/ScrapePoolPanel.tsx#L82

web/ui/react-app/src/pages/Rules.tsx
```diff
@@ -1,15 +1,108 @@
...
+                        >
+                          {r.query}
+                        </Link>
+                      </td>
+                      <td>
+                        <Alert style={{ display: 'inline-table' }}>{r.health.toUpperCase()}</Alert>
+                      </td>
+                      <td>{r.lastError.length !== 0 ? r.lastError : null}</td>
+                      <td>{roundUp(parseFloat(r.lastEvaluation))}s ago</td>
+                      <td>{r.evaluationTime}us ago</td>
```
juliusv: Same as above apply here.

web/ui/react-app/src/pages/Rules.tsx
Harkishen-Singh: Done.

web/api/v1/api.go
Harkishen-Singh: should I do the same for the `alertingRules` and `recordingRules` types?

web/api/v1/api.go
Harkishen-Singh: Done.

web/ui/react-app/src/pages/Rules.tsx
Harkishen-Singh: Done.

Harkishen-Singh:
@juliusv I have implemented the suggestions. Please have a look.

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+                              `${pathPrefix}/new/graph?g0.expr=ALERTS{alertname="${r.name}"}&g0.tab=1&g0.stacked=0&g0.range_input=1h`
+                            )}
+                          >
+                            {r.name}
+                          </Link>
+                          <br />
+                          <span className="rules-head">expr:</span>
+                          <Link
+                            style={{ marginLeft: '3%' }}
+                            to={`${pathPrefix}/new/graph?g0.expr=${r.query}&g0.tab=1&g0.stacked=0&g0.range_input=1h`}
```
boyskila: Shouldn't be encoded as it's above?
```suggestion
                            to{encodeURI(`${pathPrefix}/new/graph?g0.expr=${r.query}&g0.tab=1&g0.stacked=0&g0.range_input=1h`)}
```

Same applies for the other links

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+                          >
+                            {r.query}
+                          </Link>
+                        </td>
+                      )}
+                      <td>
+                        <Alert style={{ display: 'inline-table' }}>{r.health.toUpperCase()}</Alert>
```
boyskila:
```suggestion
                        <Alert className="d-inline-block p-1 text-uppercase">{r.health}</Alert>
```

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+                      </td>
+                      <td>{r.hasOwnProperty('lastError') ? r.lastError : null}</td>
```
boyskila:
```suggestion
                      <td>{r.lastError}</td>
```
If it hasn't own property it will not render anyway

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+                  <td className="rules-head">Rule</td>
+                  <td className="rules-head">State</td>
+                  <td className="rules-head">Error</td>
+                  <td className="rules-head">Last Evaluation</td>
+                  <td className="rules-head">Evaluation Time</td>
+                </tr>
+                {g.rules.map((r, i) => {
+                  return (
+                    <tr key={i}>
+                      {r.hasOwnProperty('alerts') ? (
```
boyskila: This check may not work since ```hasOwnProperty``` method will not check through the prototype chain. If this is intentional then fine, if not then it's better simply to check ```r.alerts```

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+      </Alert>
+    );
+  } else if (response.data) {
+    const groups: RuleGroup[] = response.data.groups;
+    return (
+      <>
+        <h2>Rules</h2>
+        {groups.map((g, i) => {
+          return (
+            <Table size="sm" striped={true} bordered={true} key={i}>
```
boyskila:
```suggestion
            <Table size="sm" striped bordered key={i}>
```

web/ui/react-app/src/pages/rules/Rules.tsx
boyskila: Also, these classes doesn't suppress Alert rounded corners

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+interface RulesMap {
+  groups: RuleGroup[];
+}
+
+interface MapProps {
+  map: Record<string, string>;
+  term: string;
+}
+
+const Maps: FC<RouteComponentProps & MapProps> = ({ term, map }) => {
```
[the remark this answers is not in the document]
Harkishen-Singh: I have updated to name as `AlertRuleMaps` for the purpose it does. Hope it is relatable.

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+  return (
+    <div>
+      <span className="rules-head">{term}:</span>
```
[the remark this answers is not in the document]
Harkishen-Singh: Done.

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+        ))}
+      </div>
+    </div>
+  );
+};
+
+const Rules: FC<RouteComponentProps & PathPrefixProps> = ({ pathPrefix }) => {
+  const { response, error } = useFetch<RulesMap>(`${pathPrefix}/api/v1/rules`);
+  console.warn(response);
+  const roundUp = (n: number) => {
```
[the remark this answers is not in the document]
Harkishen-Singh: Done.

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+                  <td className="rules-head">Error</td>
+                  <td className="rules-head">Last Evaluation</td>
+                  <td className="rules-head">Evaluation Time</td>
+                </tr>
+                {g.rules.map((r, i) => {
+                  return (
+                    <tr key={i}>
+                      {r.hasOwnProperty('alerts') ? (
+                        <td>
+                          <span className="rules-head">alert:</span>
```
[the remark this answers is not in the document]
Harkishen-Singh: Done.

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+                  <td>
+                    <h2>{formatRelative(g.lastEvaluation, now())} ago</h2>
+                  </td>
+                  <td>
+                    <h2>{roundUp(parseFloat(g.evaluationTime) * 1000)}ms</h2>
+                  </td>
+                </tr>
+              </thead>
+              <tbody>
+                <tr>
```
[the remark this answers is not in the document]
Harkishen-Singh: Done.

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+                            {r.name}
+                          </Link>
+                          <br />
+                          <span className="rules-head">expr:</span>
+                          <Link
+                            style={{ marginLeft: '3%' }}
+                            to={`${pathPrefix}/new/graph?g0.expr=${r.query}&g0.tab=1&g0.stacked=0&g0.range_input=1h`}
+                          >
+                            {r.query}
+                          </Link>{' '}
```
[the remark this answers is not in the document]
Harkishen-Singh: I think this was generated from the linter.

boyskila:
@Harkishen-Singh  On top of all the suggestions, I would say that there is inconsistency with the old UI in terms of font-size, font-family and text formatting. You can check CSS class ```rule_cell``` in the old UI - it will give you all you need to align both UI's.

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+                          <br />
+                          <AlertRuleMaps map={r.labels} term="labels" />
+                          <AlertRuleMaps map={r.annotations} term="annotations" />
+                        </td>
+                      ) : (
+                        <td>
+                          <span className="rules-head">record:</span> {r.name} <br />
```
boyskila: Shouldn't this also be a link? At least in the old UI is

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+                </tr>
+                {g.rules.map((r, i) => {
+                  return (
+                    <tr key={i}>
+                      {r.alerts ? (
+                        <td>
+                          <strong>alert:</strong>
+                          <Link
+                            className="ml-4"
+                            to={encodeURI(
```
boyskila: This expression is not formatted properly and produce this:
asdf

Harkishen-Singh: I tried on this, but wasn't able to find the correct url. Can you help me with this?

boyskila: You can move this to utils and use it in both places like so:
```<Link to=createExpressionLink(needed_param_here) ......>.....</Link>```

Comment (boyskila) [reply web/ui/react-app/src/pages/rules/Rules.tsx]: As far as I can see this

```
const GraphExpressionLink: FC<{ expr: string; title: string } & PathPrefixProps> = props => {
  return (
    <div>
      <span style={{ width: 60, display: 'inline-block' }}>{props.title}</span>
      <Link
        to={createExpression(props.expr)}
      >
        {props.children}
      </Link>
    </div>
  );
};
```
can be useful component. In Rules page you can used it in atleast 4-5 places and in Alerts page as well.

Comment (juliusv) [reply web/ui/react-app/src/pages/rules/Rules.tsx]: Yup, agreed.

File: web/ui/react-app/src/pages/rules/Rules.tsx [hunk 27]
```diff
@@ -1,15 +1,147 @@
...
+                  <td>Error</td>
+                  <td>Last Evaluation</td>
+                  <td>Evaluation Time</td>
+                </tr>
+                {g.rules.map((r, i) => {
+                  return (
+                    <tr key={i}>
+                      {r.alerts ? (
+                        <td>
+                          <strong>alert:</strong>
```
juliusv: I would prefer if we didn't use CSS/HTML spacing (`margin-left: 3%`, `<br/>`, etc.) in this field and instead kept it all preformatted like in the old UI and use existing normal whitespace for the formatting, so it's properly copyable YAML (at least for the most part).

In the old UI the styling was achieved via:

```css
.rule_cell {
    white-space: pre-wrap;
    background-color: #F5F5F5;
    display: block;
    font-family: monospace;
}
```

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+      </Alert>
+    );
+  } else if (response.data) {
+    const groups: RuleGroup[] = response.data.groups;
+    return (
+      <>
+        <h2>Rules</h2>
+        {groups.map((g, i) => {
+          return (
+            <Table size="sm" striped bordered key={i}>
```
juliusv: 1. This table shouldn't be striped (old UI isn't either), instead, grey coloring should be used for pre-formatted rule cells (see old UI).

2. The size of this table should probably normal like in the old UI (so remove the `size="sm"` option), this also leads to nicer padding in the table cells.

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+                            to={encodeURI(
+                              `${pathPrefix}/new/graph?g0.expr=${r.query}&g0.tab=1&g0.stacked=0&g0.range_input=1h`
+                            )}
+                          >
+                            {r.query}
+                          </Link>
+                        </td>
+                      )}
+                      <td>
+                        <Alert className="d-inline-block p-1 text-uppercase">{r.health.toUpperCase()}</Alert>
```
juliusv: In the new UI on the /targets page we have started using badges instead of alerts for these tiny health status texts. I would do the same here too. See: https://github.com/prometheus/prometheus/blob/2b2eb79e8b7245d2d60001ae08443d31f7b3a5b9/web/ui/react-app/src/pages/targets/ScrapePoolPanel.tsx#L65

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+                      </td>
+                      <td>{r.lastError}</td>
+                      <td>{formatRelative(r.lastEvaluation, now())} ago</td>
+                      <td>{roundUp(parseFloat(r.evaluationTime) * 1000)}ms ago</td>
```
juliusv: Some comment about `humanizeDuration()` here for the dynamic unit.

web/ui/react-app/src/pages/rules/Rules.tsx
juliusv: The color of these health badges needs to change according to the health. E.g. OK = green, ERR = red, and UNKNOWN = yellow/orange.

You can see rules in an unknown state right after starting Prometheus (before a rule is evaluated), and you can reproduce a rule failure easily like this (without requiring special data in your Prometheus):

```yml
groups:
- name: ruleerrors
  rules:
  - record: brokenrule
    expr: label_join(foo, ".", "")
```

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+                            )}
+                          >
+                            {r.query}
+                          </Link>
+                        </td>
+                      )}
+                      <td>
+                        <Alert className="d-inline-block p-1 text-uppercase">{r.health.toUpperCase()}</Alert>
+                      </td>
+                      <td>{r.lastError}</td>
```
juliusv: This need to be highlighted in red when there's an error, like in the old UI. (I would just do a normal `<Alert color="danger">` for this in the new UI, not a `<span ...>` like in the old one.

web/ui/react-app/src/pages/rules/Rules.tsx
```diff
@@ -1,15 +1,147 @@
...
+                          <Link
+                            className="ml-4"
+                            to={encodeURI(
+                              `${pathPrefix}/new/graph?g0.expr=${r.query}&g0.tab=1&g0.stacked=0&g0.range_input=1h`
+                            )}
+                          >
+                            {r.query}
+                          </Link>{' '}
+                          <br />
+                          <AlertRuleMaps map={r.labels} term="labels" />
```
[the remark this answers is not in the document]
Harkishen-Singh: `pathPrefix` can be undefined as well. Hence getting an error.

web/ui/react-app/src/pages/rules/Rules.tsx
Harkishen-Singh: Fixed this.

web/ui/react-app/src/pages/rules/Rules.tsx
Harkishen-Singh: Done. Thanks.

Harkishen-Singh:
@juliusv @boyskila I have implemented the suggestions. Please have a look. Thank you.

web/ui/react-app/src/pages/rules/RulesContent.tsx
```diff
@@ -0,0 +1,131 @@
...
+  rules: Rule[];
+  evaluationTime: string;
+  lastEvaluation: string;
+}
+
+export interface RulesMap {
+  groups: RuleGroup[];
+}
+
+const GraphExpressionLink: FC<{ expr: string; title: string }> = props => {
```
boyskila: This component should look like:
```
export const GraphExpressionLink: FC<{ expr: string; title: string; name: string  & PathPrefixProps}> = memo(({ name, title, expr, pathPrefix }) => {
  return (
    <div>
      {title}:
      <Link className="ml-4" to={createExpressionLink(expr, pathPrefix)}>
        {name}
      </Link>
    </div>
  );
});
```

```name``` and ```expr``` may have different content, so we can't rely on ```expr``` to represent the ```name```
Also, wrapping the ```Link``` and ```title``` in div will remove the need of ```<br/>```. IMO it's better.
Comparing with the old UI titles here are wrapped in ```<strong>```, maybe you should remove the extra tag

Another, more important thing is to pas the ```pathPrefix``` prop down to ```createExpressionLink```. IMO the ``../``` in ```createExpressionLink``` should be passed from outside and the function just to return
``` ${prefix}graph?g0.expr=${encodeURIComponent(expr)}&g0.tab=1&g0.stacked=0&g0.range_input=1h ```
The usage should looks like ```createExpressionLink(expr, pathPrefix + '/../')```

File: web/ui/react-app/src/pages/rules/RulesContent.tsx [hunk 34]
```diff
@@ -0,0 +1,131 @@
...
+                  <td>Rule</td>
+                  <td>State</td>
+                  <td>Error</td>
+                  <td>Last Evaluation</td>
+                  <td>Evaluation Time</td>
+                </tr>
+                {g.rules.map((r, i) => {
+                  return (
+                    <tr key={i}>
+                      {r.alerts ? (
```
boyskila: ```<td>``` content can be reduced to
```
<td className="rule_cell">
      <GraphExpressionLink name={r.name} title={r.alerts ? 'alert' : 'record'} expr={`ALERTS{alertname="${r.name}"}`}/>
      <GraphExpressionLink title="expr" expr={r.query} name={r.query} />
         {r.alerts && (
              <>
                 labels:
                 {Object.entries(r.labels).map(([key, value]) => (
                     <div className="ml-4" key={key}>
                         {key}: {value}
                      </div>
                   ))}
                    annotations:
                    {Object.entries(r.annotations).map(([key, value]) => (
                        <div className="ml-4" key={key}>
                            {key}: {value}
                         </div>
                      ))}
                  </>
             )}
 </td>
```
Some of the divs are redundant and.
Alse, note that first ```GraphExpressionLink``` has as an expression  ```ALERTS{alertname="${r.name}"}```. Passing just ```r.name``` as expr is not enough and its not working

web/ui/react-app/src/utils/index.ts
```diff
@@ -197,3 +197,7 @@ export const toQueryString = ({ key, options }: PanelMeta) => {
 export const encodePanelOptionsToQueryString = (panels: PanelMeta[]) => {
   return `?${panels.map(toQueryString).join('&')}`;
 };
+
+export const createExpressionLink = (expr: string) => {
```
boyskila: Please pass as second argument the ```pathPrefix``` prop and remove ```../``` - this should be passed from outside as well

juliusv:
Thanks! :) There are still a couple of new (and old, https://github.com/prometheus/prometheus/pull/6503#discussion_r368539255) outstanding comments, but I'll merge this for now, and then we can individually refactor/address the remaining few bits.

PR #6505, opened 2019-12-22
OpenMetrics: force input to end with # EOF

### LINKED ISSUE #1169: Deal with incomplete scrapes.
A scrape might be incomplete because of a network error (connection closed, time-out, incomplete HTTP content, ...) or because a client wasn't able to collect all metrics (for whatever reason).

With https://github.com/prometheus/common/issues/14, it will be able to detect incomplete scrapes even without a network error.

At the very least, an incompletely scraped target must be marked as unhealthy, and its `up` metric must be 0. (This might or might not already be the case for network failures.)

Other measures to deal with incomplete scrapes need vetting:

It is easiest to reason about metrics if scrapes are atomic, i.e. either ingest all samples from a complete scrape, or none at all. Implementing an atomic scrape semantics has to be done carefully to not introduce a lot of allocation churn or unpredictable memory consumption (due to buffering until completion of a scrape is confirmed) and to not incur a significant performance cost.

On the other hand, in some situations it might be best to ingest what we can get, even if it is from an incomplete scrape. Via the `up` metric, alerts sensitive to incomplete scrapes can still be protected.

Note that a metric family will always be ingested as a whole from protobuf, even without a termination marker in the exchange format. The same guarantee can be given for the text format once we have a termination marker in the exchange format. However, since appending happens sequentially per time series, a rule evaluation might see some samples from a metric family already ingested but others not - even from a complete scrape.

roidelapluie:
Fixes #1169

### DIFF
--- a/pkg/textparse/openmetricsparse.go
+++ b/pkg/textparse/openmetricsparse.go
@@ -90,6 +90,7 @@ type OpenMetricsParser struct {
 	hasTS   bool
 	start   int
 	offsets []int
+	prev    token
 
 	eOffsets      []int
 	exemplar      []byte
@@ -232,13 +233,18 @@ func (p *OpenMetricsParser) Next() (Entry, error) {
 	p.exemplarVal = 0
 	p.hasExemplarTs = false
 
-	switch t := p.nextToken(); t {
+	t := p.nextToken()
+	defer func() { p.prev = t }()
+	switch t {
 	case tEofWord:
 		if t := p.nextToken(); t != tEOF {
 			return EntryInvalid, errors.New("unexpected data after # EOF")
 		}
 		return EntryInvalid, io.EOF
 	case tEOF:
+		if p.prev != tEofWord {
+			return EntryInvalid, errors.New("data does not end with # EOF")
+		}
 		return EntryInvalid, io.EOF
 	case tHelp, tType, tUnit:
 		switch t := p.nextToken(); t {
@@ -322,7 +328,7 @@ func (p *OpenMetricsParser) Next() (Entry, error) {
 		p.hasTS = false
 		switch t2 := p.nextToken(); t2 {
 		case tEOF:
-			return EntryInvalid, io.EOF
+			return EntryInvalid, errors.New("data does not end with # EOF")
 		case tLinebreak:
 			break
 		case tComment:
@@ -382,7 +388,7 @@ func (p *OpenMetricsParser) parseComment() error {
 	p.hasExemplarTs = false
 	switch t2 := p.nextToken(); t2 {
 	case tEOF:
-		return io.EOF
+		return errors.New("data does not end with # EOF")
 	case tLinebreak:
 		break
 	case tTimestamp:
--- a/pkg/textparse/openmetricsparse_test.go
+++ b/pkg/textparse/openmetricsparse_test.go
@@ -1,4 +1,4 @@
-// Copyright 2017 The OMetheus Authors
+// Copyright 2017 The Prometheus Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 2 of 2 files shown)


pkg/textparse/openmetricsparse.go
```diff
@@ -81,15 +81,17 @@ func (l *openMetricsLexer) Error(es string) {
...
+	series         []byte
+	text           []byte
+	mtype          MetricType
+	val            float64
+	ts             int64
+	hasTS          bool
+	start          int
+	offsets        []int
+	prev           token
+	ignoreEofError bool // only used to simplify tests.
```
brian-brazil: No need to do this, the tests already allow for this.

roidelapluie: Without it all the tests errors are "data does not end with # EOF"

roidelapluie: I have updated the pull request to use the public methods only

brian-brazil: Update the testcases then

roidelapluie: Updated

roidelapluie: Now it looks close to the way we test the Python parser

pkg/textparse/openmetricsparse_test.go
```diff
@@ -272,71 +272,91 @@ func TestOpenMetricsParseErrors(t *testing.T) {
 		err   string
 	}{
 		{
-			input: "",
+			input: "# EOF",
```
brian-brazil: This is valid input.

pkg/textparse/openmetricsparse_test.go
```diff
@@ -272,71 +272,91 @@ func TestOpenMetricsParseErrors(t *testing.T) {
...
 			err:   "EOF",
 		},
+		{
+			input: "# EOF\n",
+			err:   "EOF",
```
brian-brazil: This is also valid input.

pkg/textparse/openmetricsparse_test.go
roidelapluie: Yes, the error is EOF; which means that it is valid. It was like it before.

pkg/textparse/openmetricsparse_test.go
roidelapluie: Same err. We test the error code of the next() and in this case it is EOF.

roidelapluie:
My understanding is that we tested the err returned by Next(). In that situation EOF is returned by Next() ; I did not change the way tests are written.

pkg/textparse/openmetricsparse.go
```diff
@@ -232,13 +233,18 @@ func (p *OpenMetricsParser) Next() (Entry, error) {
...
+	t := p.nextToken()
+	defer func() { p.prev = t }()
+	switch t {
 	case tEofWord:
 		if t := p.nextToken(); t != tEOF {
 			return EntryInvalid, errors.New("unexpected data after # EOF")
 		}
 		return EntryInvalid, io.EOF
 	case tEOF:
+		if p.prev != tEofWord {
```
bwplotka: nit: We tend to capitalize all acronyms on Go e.g `EOF` as described [here](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

pkg/textparse/openmetricsparse_test.go
brian-brazil: This needs a comment then.

pkg/textparse/openmetricsparse.go
roidelapluie: Except that this variable is generated by Golex -- and the link you provide says:

> Code generated by the protocol buffer compiler is exempt from this rule. Human-written code is held to a higher standard than machine-written code.

While this is not protocol buffer compiler, it is machine generated code.

pkg/textparse/openmetricsparse_test.go
roidelapluie: Added

brian-brazil:
Thanks!

PR #6507, opened 2019-12-23
.gitignore: Add tags and swaps

truongnh1992:
*Description:* Add vim swaps and tags to .gitignore
*Risk Level:* low

Signed-off-by: Nguyen Hai Truong <truongnh@vn.fujitsu.com>


brian-brazil:
Thanks for your PR. This would better belong in your global gitignore.

PR #6509, opened 2019-12-23
Cut 2.15.0 release.

bwplotka:
No regression reported and benchmarks look as expected: lower memory usage for loaded blocks/compactions with the tradeoff of compaction/cut head taking more time.

image

image

Tested long-range queries a bit as well and all is fine.

Let's release 2.15.0 :+1: 

Signed-off-by: Bartek Plotka <bwplotka@gmail.com>

PR #6510, opened 2019-12-23
Pull release-2.15 branch to master.

bwplotka:


bwplotka:
Merging commits as it is, so ignoring DCO check. Not sure why it fails though as the failing commit passed DCO check: https://github.com/prometheus/prometheus/pull/6509

PR #6511, opened 2019-12-23
Use samplesPending rather than integral

csmarchbanks:
The integral accumulator in the remote write sharding code is just a
second way of keeping track of the number of samples pending. Remove
integralAccumulator and use the samplesPending value we already
calculate to calculate the number of shards.

This has the added benefit of fixing a bug where the integralAccumulator
was not being initialized correctly due to not taking into account the
number of ticks being counted, causing the integralAccumulator initial
value to be off by an order of magnitude in some cases.

By using `0.2 / shardUpdateDuration` as our gain, the gain is the same as it used to be where we did a sum of `0.2 * rate/sec` every `shardUpdateDuration`. I dropped to `0.1` instead of `0.2` in order to try to catch up in 10 cycles instead of 5 (100 s instead of 50). This helps smooth out the sharding some.

Fixes point 2 of https://github.com/prometheus/prometheus/issues/6333

cc: @cstyan

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -214,8 +214,6 @@ type QueueManager struct {
 	wg          sync.WaitGroup
 
 	samplesIn, samplesDropped, samplesOut, samplesOutDuration *ewmaRate
-	integralAccumulator                                       float64
-	startedAt                                                 time.Time
 
 	highestSentTimestampMetric *maxGauge
 	pendingSamplesMetric       prometheus.Gauge
@@ -316,8 +314,6 @@ outer:
 // Start the queue manager sending samples to the remote storage.
 // Does not block.
 func (t *QueueManager) Start() {
-	t.startedAt = time.Now()
-
 	// Setup the QueueManagers metrics. We do this here rather than in the
 	// constructor because of the ordering of creating Queue Managers's, stopping them,
 	// and then starting new ones in storage/remote/storage.go ApplyConfig.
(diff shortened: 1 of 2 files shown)


csmarchbanks:
@cstyan I added the first stab at a test. I am probably going to do some refactoring to make testing this easier such as removing the dependency on metrics and/or breaking the shard calculation out of the queue manager. However, I want to see if the general approach seems ok first. Does the test make sense as it stands? 2 minute startup, then start sending samples, and make sure our shard count never drops too low.

cstyan:
The test doesn't seem right to me. I think starting up with pending samples and then faking sending the samples is right, but we should be checking the pending samples I think. The issue was that we were never catching up. So if the number of pending samples is similar at the end as it was was at the beginning then the test should fail.

Hacking around a bit, doesn't seem like a test like that will be clean without some refactoring. Ping me when you have time to talk about this.

csmarchbanks:
I agree that pending samples is better, I changed the test some in an additional commit with message "Keep track of pending samples directly rather than approximated by time back" to keep track of the number of samples written. As the message says, I also changed around `ewma` to keep track of a total in addition to the weighted average. Let me know what you think.

storage/remote/queue_manager.go
```diff
@@ -581,6 +581,12 @@ func (t *QueueManager) calculateDesiredShards() int {
 	}
 
 	numShards := int(math.Ceil(desiredShards))
+	// Do not downshard if we are more than ten seconds back.
+	if numShards < t.numShards && samplesPending > 10.0*samplesInRate {
```
cstyan: should use `shardUpdateDuration` here imo

storage/remote/queue_manager.go
```diff
@@ -544,8 +549,9 @@ func (t *QueueManager) calculateDesiredShards() int {
 
 	// When behind we will try to catch up on a proporation of samples per tick.
 	// This works similarly to an integral accumulator in that pending samples
-	// is the result of the error integral.
-	const integralGain = 0.2 / float64(shardUpdateDuration/time.Second)
+	// is the result of the error integral. This effectively means try to catch
+	// up on all pending samples within the next 10 cycles (100 seconds).
```
cstyan: when the default `shardUpdateDuration` of 10s is used, we should mention that just incase the value is ever changed, or just say `within the next <math here> cycles.`

storage/remote/queue_manager.go
csmarchbanks: I am not sure, they happen to be the same right now, but I would be perfectly happy to change this value to something else that doesn't match the `shardUpdateDuration`. In other words, I don't think they are actually coupled.

storage/remote/ewma.go
```diff
@@ -48,23 +49,32 @@ func (r *ewmaRate) rate() float64 {
 	return r.lastRate
 }
 
+// count returns the total events recorded.
```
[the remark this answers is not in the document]
csmarchbanks: Maybe `count returns the total number of events recorded.`? Do you have a different suggestion?

csmarchbanks:
Turns out we cannot use the keep a count method since `samplesIn` will exist from the beginning of the server, and if you add another queue the values will be very very wrong. Current state of this PR:

1. First commit contains a failing test
2. Second commit fixes the test by using the estimated pending samples from the delay rather than trying to incorrectly calculate the pending samples on startup.
3. Third commit is an optional improvement to not shard down while catching up. Happy to break this to a separate PR, but it is pretty simple. Test shows faster catchup, but will still pass with or without this.

Issue #6512, opened 2019-12-23
checksum mismatch breaking Grafana queries


howardjohn:
## Bug Report
**What did you do?**
Updated Prometheus 2.12.0 to 2.15.0. Look at queries in grafana. Sometimes graphs show errors like:

![2019-23-12_14-15-21](https://user-images.githubusercontent.com/623453/71382900-d22d7680-258e-11ea-8882-6e5cedda4b40.png)

Usually seems to happen only when I had a long time window. The same error occurs on all queries intermittently.

**What did you expect to see?**

No errors.

**What did you see instead? Under which circumstances?**

Checksum mismatch error

**Environment**

Prometheus 2.15.0 running from prometheus operator on Kubernetes.

Grafana 6.4.3 and 6.5.2 tested. The problem seems independent of Grafana version. Its highly possible this is a Grafana issue, but because it only is happening with Prometheus 2.15.0 I felt it is more likely Prometheus

* Prometheus version:

	2.15.0

* Prometheus configuration file:
```
    enableAdminAPI: false
    image: docker.io/prom/prometheus:v2.15.0
    podMetadata:
      annotations:
        sidecar.istio.io/inject: "false"
      labels:
        app: prometheus
    resources:
      requests:
        memory: 32Gi
    retention: 72h
    scrapeInterval: 15s
    secrets:
    - istio.prometheus
    securityContext:
      fsGroup: 2000
      runAsNonRoot: true
      runAsUser: 1000
    serviceAccountName: prometheus
    serviceMonitorNamespaceSelector:
      any: true
    serviceMonitorSelector:
      any: true
    storage:
      volumeClaimTemplate:
        selector: {}
        spec:
          accessModes:
          - ReadWriteOnce
          resources:
            requests:
              storage: 500Gi
          storageClassName: ssd
    version: v2.15.0
```

* Logs:
<details>

```
level=info ts=2019-12-23T15:23:36.036Z caller=main.go:330 msg="Starting Prometheus" version="(version=2.15.0, branch=HEAD, revision=ec1868b0267d13cb5967286fd5ec6afff507905b)"
level=info ts=2019-12-23T15:23:36.036Z caller=main.go:331 build_context="(go=go1.13.5, user=root@240f2f89177f, date=20191223-12:03:32)"
[...]

PR #6514, opened 2019-12-24
tsdb: Do not reuse the same crc32 hash between Chunk() calls

roidelapluie:
The same crc32 object was called concurrently between chunks. As this
object is only used in that function, create a new one each time the
function is run.

See #6512

### DIFF
--- a/tsdb/chunks/chunks.go
+++ b/tsdb/chunks/chunks.go
@@ -456,16 +456,15 @@ func (b realByteSlice) Sub(start, end int) ByteSlice {
 type Reader struct {
 	// The underlying bytes holding the encoded series data.
 	// Each slice holds the data for a different segment.
-	bs    []ByteSlice
-	cs    []io.Closer // Closers for resources behind the byte slices.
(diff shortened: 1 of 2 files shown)


roidelapluie:
Now to test this...

roidelapluie:
Now to test it..

roidelapluie:
It does not fix it (test fails)

roidelapluie:
I am not sure that it is enough ; now it runs fine if  r.Chunk(chkExp.Ref) is called concurrently with the same chkExp.Ref but not otherwise.

roidelapluie:
more-complex-than-it-looks

bwplotka:
Thanks for starting this up. Looks like we are also reusing `buf = [binary.MaxVarintLen32]byte{}` This is to reuse the same buffer for calculating the checksum. The easiest would be to just do `Sum(nil)`, but we can consider adding sync.Pool for this. For 5 bytes I think we can + slice header I think it's not worth it.

PR #6517, opened 2019-12-24
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

### DIFF
--- a/notifier/notifier_test.go
+++ b/notifier/notifier_test.go
@@ -299,7 +299,7 @@ func TestHandlerRelabel(t *testing.T) {
 	testutil.Ok(t, alertsEqual(expected, h.queue))
 }
 
-func TestHandlerQueueing(t *testing.T) {
+func TestHandlerQueuing(t *testing.T) {
 	var (
 		expectedc = make(chan []*Alert)
 		called    = make(chan struct{})
--- a/promql/bench_test.go
+++ b/promql/bench_test.go
@@ -5,7 +5,7 @@
 //
 // http://www.apache.org/licenses/LICENSE-2.0
 //
-// Unless required by applicable law or agreed to in writing, softwar
+// Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
--- a/tsdb/db_test.go
+++ b/tsdb/db_test.go
@@ -1176,11 +1176,11 @@ func TestSizeRetention(t *testing.T) {
 	testutil.Ok(t, err)
 	// Expected size should take into account block size + WAL size
 	expSize = blockSize + walSize
-	actRetentCount := int(prom_testutil.ToFloat64(db.metrics.sizeRetentionCount))
+	actRetentionCount := int(prom_testutil.ToFloat64(db.metrics.sizeRetentionCount))
 	actSize, err = fileutil.DirSize(db.Dir())
 	testutil.Ok(t, err)
 
-	testutil.Equals(t, 1, actRetentCount, "metric retention count mismatch")
+	testutil.Equals(t, 1, actRetentionCount, "metric retention count mismatch")
 	testutil.Equals(t, actSize, expSize, "metric db size doesn't match actual disk size")
 	testutil.Assert(t, expSize <= sizeLimit, "actual size (%v) is expected to be less than or equal to limit (%v)", expSize, sizeLimit)
 	testutil.Equals(t, len(blocks)-1, len(actBlocks), "new block count should be decreased from:%v to:%v", len(blocks), len(blocks)-1)
--- a/tsdb/head_test.go
+++ b/tsdb/head_test.go
@@ -1255,7 +1255,7 @@ func TestWalRepair_DecodingError(t *testing.T) {
 }
 
 func TestNewWalSegmentOnTruncate(t *testing.T) {
-	dir, err := ioutil.TempDir("", "test_wal_segemnts")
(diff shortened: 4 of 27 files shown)


roidelapluie:
Thank you for your report. While that is interesting, I am wondering if we should merge this one-off and/or integrate it in our CI jobs.

jsoref:
You might look at:
https://github.com/checkstyle/checkstyle/blob/master/.ci/test-spelling-unknown-words.sh

I'm probably going to work on a version that uses a bucket for storage for our word-list I'm the near future.

Someday I might manage to make this into a GitHub action or similar, but I can't offer a timetable.

promql/generated_parser.y.go
```diff
@@ -981,7 +981,7 @@ yydefault:
 			var err error
 			yyVAL.uint, err = strconv.ParseUint(yyDollar[1].item.Val, 10, 64)
 			if err != nil {
-				yylex.(*parser).errorf("invalid repitition in series values: %s", err)
+				yylex.(*parser).errorf("invalid repetition in series values: %s", err)
```
jsoref: This looks suspiciously like generated code/a third party. Let me know the origin and I'll see about scanning it instead…

promql/bench_test.go
```diff
@@ -5,7 +5,7 @@
 //
 // http://www.apache.org/licenses/LICENSE-2.0
 //
-// Unless required by applicable law or agreed to in writing, softwar
+// Unless required by applicable law or agreed to in writing, software
```
jsoref: This is odd. Generally lawyers get annoyed at things like this.

codesome: This correction looks fine.

promql/generated_parser.y.go
codesome: This is a generated file and should not be changed. The method of generation is in the first line of this file, which is `goyacc`.

jsoref: https://gitlab.com/cznic/goyacc/blob/master/ ? -- I'll have to look...

tsdb/db_test.go
```diff
@@ -1176,11 +1176,11 @@ func TestSizeRetention(t *testing.T) {
 	testutil.Ok(t, err)
 	// Expected size should take into account block size + WAL size
 	expSize = blockSize + walSize
-	actRetentCount := int(prom_testutil.ToFloat64(db.metrics.sizeRetentionCount))
+	actRetentionCount := int(prom_testutil.ToFloat64(db.metrics.sizeRetentionCount))
```
juliusv: This seems to be less of a spelling fix and more writing out an intentionally abbreviated term - but I like the new version more :)

tsdb/head_test.go
```diff
@@ -644,9 +644,9 @@ func TestDeleteUntilCurMax(t *testing.T) {
 	testutil.Assert(t, res.Next(), "series don't exist")
 	exps := res.At()
 	it := exps.Iterator()
-	ressmpls, err := expandSeriesIterator(it)
+	ressamples, err := expandSeriesIterator(it)
```
juliusv: If you're going to unabbreviate `ressmpls` here, I'd probably go for `resSamples` camelCasing.

notifier/notifier_test.go
```diff
@@ -299,7 +299,7 @@ func TestHandlerRelabel(t *testing.T) {
 	testutil.Ok(t, alertsEqual(expected, h.queue))
 }
 
-func TestHandlerQueueing(t *testing.T) {
+func TestHandlerQueuing(t *testing.T) {
```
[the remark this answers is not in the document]
jsoref: Both spellings are currently present. I'm aware both are individually acceptable. A given locale would only use one.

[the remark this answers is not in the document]
brian-brazil: Both appear to be valid US spellings (though it seems the prevalence of queuing is pretty recent). I personally would use queueing, and would consider queuing a misspelling.

promql/generated_parser.y.go
slrtbtfs: Nope, golang.org/x/tools/cmd/goyacc .

The file is generated from `promql/generated_parser.y`.

We probably should document this somewhere.

jsoref: Err, so I should have kept the change https://github.com/prometheus/prometheus/commit/6a6d61f6439e40a109eff25b741bf25d730560fa to: https://github.com/prometheus/prometheus/blob/213a8fe89a7308e73f22888a963cbf9375217cd6/promql/generated_parser.y#L348 ?


(That filename is fairly confusing. I'm going to let the dust settle on this PR and then I can go back and touch it. -- And yes, I understand the tradition of naming a `.y` file and a `.something` file with the same prefix.)

slrtbtfs: > Err, so I should have kept the change [6a6d61f](https://github.com/prometheus/prometheus/commit/6a6d61f6439e40a109eff25b741bf25d730560fa) to:
> 
> https://github.com/prometheus/prometheus/blob/213a8fe89a7308e73f22888a963cbf9375217cd6/promql/generated_parser.y#L348
> ?
Yes, and then regenerate the file.

> (That filename is fairly confusing. I'm going to let the dust settle on this PR and then I can go back and touch it. -- And yes, I understand the tradition of naming a `.y` file and a `.something` file with the same prefix.)

I'm doing a lot of changes in that file at the moment anyway, so I can just fix that typo on the way.

And yes, the naming and generation process should be made more clear.

jsoref:
Do I need to do anything else here?

juliusv:
Looks agreeable enough to me, thanks :) :+1:

promql/generated_parser.y.go
slrtbtfs: This spelling error gets fixed as a side effect of #6548.

PR #6520, opened 2019-12-25
Add query log

### LINKED ISSUE #1315: Add option to log slow queries and recording rules

roidelapluie:
It is set in the configuration file to be changed online.

Fixes #1315


brian-brazil:
Request logs still don't belong in application logs.

roidelapluie:
As stated before in this project, we tend not to reinvent the wheel in this project.

Given that file rotation is usually a solved problem, I think we can:

- Not do it ourselve
- Reopen the log file upon every reload (that would align us with other softwares). That makes us compatible with e.g. logrotate.

brian-brazil:
> Reopen the log file upon every reload 

That seems like a reasonable approach.

config/config.go
```diff
@@ -299,6 +300,10 @@ type GlobalConfig struct {
 	ScrapeTimeout model.Duration `yaml:"scrape_timeout,omitempty"`
 	// How frequently to evaluate rules by default.
 	EvaluationInterval model.Duration `yaml:"evaluation_interval,omitempty"`
+	// Duration after which PromQL queries are logged.
+	QueryLogMinDuration *model.Duration `yaml:"query_log_min_duration,omitempty"`
```
brian-brazil: I think we'd said just to log everything, in any case "Min" in the name could be confused to mean minutes.

docs/configuration/configuration.md
```diff
@@ -61,6 +61,13 @@ global:
   # How frequently to evaluate rules.
   [ evaluation_interval: <duration> | default = 1m ]
 
+  # File to which PromQL queries are logged if <query_log_min_duration> is set.
```
brian-brazil: Why would the duration be required for logging to work?

promql/slow_query_logger.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func newSlowQueriesLogger(w io.Writer) log.Logger {
+	var (
+		timestampFormat = log.TimestampFormat(
+			func() time.Time { return time.Now().UTC() },
+			"2006-01-02T15:04:05.000Z07:00",
+		)
+	)
+	l := log.NewLogfmtLogger(log.NewSyncWriter(w))
```
brian-brazil: This is a hot path, I'm not sure no buffering is wise. We should probably also have this as JSON only.

roidelapluie:
- Updated to remove the "slow" part ; now we will log all queries independently of their duration.
- Added tests to check that we reopen the files *and* keep them open in case of reload failure
- Issue a warning if we can't close previous file (that does not make reload fail)
- Added comments in code

roidelapluie:
I have also added the query log format to the list of things considered unstable so we have the ability to make the format evolve in the future.

promql/slow_query_logger.go
roidelapluie: done

docs/configuration/configuration.md
roidelapluie: remove duration

config/config.go
roidelapluie: Removed duration.

roidelapluie:
Further question @brian-brazil:

Should we add a metric `query_log_enabled` 1 or 0 ?

Should we log as INFO in the application log the slow query log activation / deactivation / filepath change ?

brian-brazil:
> Should we log as INFO in the application log the slow query log activation / deactivation / filepath change ?

Sounds more debug to me offhand.

> Should we add a metric query_log_enabled 1 or 0 ? (could be done as a separated PR)

Makes sense, this may have performance impact.

docs/configuration/configuration.md
```diff
@@ -61,6 +61,10 @@ global:
   # How frequently to evaluate rules.
   [ evaluation_interval: <duration> | default = 1m ]
 
+  # File to which PromQL queries are logged.
+  # Reloading the configuration will reopen the file (e.g. after rotation).
```
brian-brazil: I'd exclude the bit in brackets.

Also put this last, it's very different to the other settings as it's not semantic.

docs/stability.md
```diff
@@ -30,6 +30,7 @@ Things considered unstable for 2.x:
 * HTML generated by the web UI
 * The metrics in the /metrics endpoint of Prometheus itself
 * Exact on-disk format. Potential changes however, will be forward compatible and transparently handled by Prometheus
+* The format of the query log
```
brian-brazil: I'd say logs in general

promql/engine.go
```diff
@@ -180,15 +182,15 @@ func (q *query) Exec(ctx context.Context) *Result {
 	var queryIndex int
 	if q.ng.activeQueryTracker != nil {
 		queryIndex = q.ng.activeQueryTracker.Insert(q.q)
+		defer func() {
```
brian-brazil: You don't need the func() here

promql/query_logger.go
```diff
@@ -184,3 +190,38 @@ func (tracker ActiveQueryTracker) Insert(query string) int {
...
+		prevFile = ql.file
+	}
+
+	if s == "" {
+		return nil, prevFile, nil
+	}
+
+	// We always open the file again; so that a Prometheus reload can be
+	// called by external programs when they rotate the query log file.
+	f, err := os.OpenFile(s, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
brian-brazil: Don't make presumptions about permissions, always 0o666. It's umask's responsibility to limit this.

promql/query_logger.go
```diff
@@ -184,3 +190,38 @@ func (tracker ActiveQueryTracker) Insert(query string) int {
...
+	if err != nil {
+		// In case of failure, we keep the old query logger and make reload fail.
+		return ql, nil, errors.Wrap(err, "can't create query log")
+	}
+
+	lg := log.With(log.NewJSONLogger(f), "ts", timestampFormat)
+	l.logger = lg
```
brian-brazil: Why not create the QueryLogger here?

promql/engine.go
```diff
@@ -307,6 +310,27 @@ func NewEngine(opts EngineOpts) *Engine {
 	}
 }
 
+// ApplyConfig updates the state as the new config requires.
+func (ng *Engine) ApplyConfig(c *config.Config) error {
+	var (
+		err error
+		f   *os.File
+	)
+	ng.queryLogger, f, err = reloadQueryLogger(ng.queryLogger, c.GlobalConfig.QueryLogFile)
```
brian-brazil: You've no locking around ng.queryLogger

promql/engine.go
```diff
@@ -379,6 +403,18 @@ func (ng *Engine) exec(ctx context.Context, q *query) (Value, storage.Warnings,
 	ctx, cancel := context.WithTimeout(ctx, ng.timeout)
 	q.cancel = cancel
 
+	defer func() {
+		// queryLogger might have been swapped or removed since the query
+		// start.
+		if l := ng.queryLogger; l != nil {
+			f := []interface{}{"statement", q.Statement().String()}
```
brian-brazil: It's a query, not a statement.

brian-brazil: Where's the step etc.?

promql/engine.go
```diff
@@ -307,6 +310,27 @@ func NewEngine(opts EngineOpts) *Engine {
...
 }
 
+// ApplyConfig updates the state as the new config requires.
+func (ng *Engine) ApplyConfig(c *config.Config) error {
+	var (
+		err error
+		f   *os.File
+	)
+	ng.queryLogger, f, err = reloadQueryLogger(ng.queryLogger, c.GlobalConfig.QueryLogFile)
+	// The prev file is only returned if the query logger reload is successful,
```
brian-brazil: So we can check err first in the usual way.

docs/configuration/configuration.md
roidelapluie: > Per the dev summit this should include IP addresses when it's coming from the query API.

Does the engine know about it yet?

roidelapluie:
Thank you @brian-brazil for the review.

Do you think that this should contain API requests too ? I was under the impression that Query Log and Access Logs were different. If we want to log AccessLogs then I need to work outside of the engine too -- e.g. logging admin commands such as DELETE.

docs/configuration/configuration.md
brian-brazil: It doesn't, as there's a separation of concerns.

brian-brazil:
> If we want to log AccessLogs then I need to work outside of the engine too -- e.g. logging admin commands such as DELETE.

Hmm, we probably should log those too. Probably out of scope for this particular PR though.

promql/engine.go
```diff
@@ -404,15 +414,22 @@ func (ng *Engine) exec(ctx context.Context, q *query) (Value, storage.Warnings,
...
-		// start.
+		ng.queryLoggerLock.RLock()
 		if l := ng.queryLogger; l != nil {
-			f := []interface{}{"statement", q.Statement().String()}
+			f := []interface{}{"query", q.q}
+			if eq, ok := q.Statement().(*EvalStmt); ok {
+				f = append(f,
+					"start", formatDate(eq.Start),
+					"end", formatDate(eq.End),
+					"interval", eq.Interval.String(),
```
brian-brazil: It's the step to users.

promql/engine.go
```diff
@@ -404,15 +414,22 @@ func (ng *Engine) exec(ctx context.Context, q *query) (Value, storage.Warnings,
...
 
 	defer func() {
-		// queryLogger might have been swapped or removed since the query
-		// start.
+		ng.queryLoggerLock.RLock()
 		if l := ng.queryLogger; l != nil {
-			f := []interface{}{"statement", q.Statement().String()}
+			f := []interface{}{"query", q.q}
+			if eq, ok := q.Statement().(*EvalStmt); ok {
+				f = append(f,
```
brian-brazil: Shouldn't we also indicate if it failed?

roidelapluie:
I added a naive way of tracking the origin of a query.

I made it so that we can further log when queries come from rules , console template, , etc

for now it is a free map[string]string, maybe we directly should go for a proper Origin struct

brian-brazil:
I think using the context would be cleaner here, this is per-request information.

roidelapluie:
Adjusted;  I will do more cleanup and add tests somewhere next week

promql/engine.go
```diff
@@ -379,6 +415,31 @@ func (ng *Engine) exec(ctx context.Context, q *query) (Value, storage.Warnings,
...
+			case *EvalStmt:
+				f = append(f,
+					"start", formatDate(eq.Start),
+					"end", formatDate(eq.End),
+					"step", eq.Interval.String(),
+				)
+			case testStmt:
+				f = append(f, "test", "yes")
+			}
+			if v := ctx.Value(queryCtx("ip")); v != nil {
```
bwplotka: This looks nice, but similarly to the previous comments, it would be maybe nice to allow users of `promql` package to add their own fields e.g duration, tenant ID, additional statistics etc. 

We can do that by: 
A) Allow adding and fetching context value for queryLogFields
B) Handle query log outside of the engine. There is nothing specific/private that we don't have access to outside e.g in webHandler. It can also handle log file reopen.

There is nothing blocking `promql` non-Prometheus users to reimplement similar query log thing with their own fields outside  of this as well, so not a blocker (:

roidelapluie: The stats are private.

promql/engine.go
roidelapluie: This is update now afaik

promql/engine.go
bwplotka: hm.. `q.Stats().ToMap()` is public?

brian-brazil:
> BTW: With some request ID and logline on the start of invoking query, can we drop an active query tracker? thinking

No, as you wouldn't log queries that are active when you OOMed.

promql/engine.go
roidelapluie: Indeed. However recording rules etc also need to be taken into account. I do not import config anymore.

roidelapluie:
I have been thinking on this one and here is how I plan to continue the implementation:

The promql package will get an interface
```
type logger interface{
Log(...interface{}) error
}
```

Then there will be a SetLogger(logger) function added to the engine.

The file opening/gokit logger creation will happen in a third party package, e.g. in pkg/filelogger.

cmd/prometheus/main.go reload functions will include a reload function that will call the filelogger and engine.SetLogger.

That looks to me a good compromise:

- not putting too much logic in cmd/prometheus/main.go as it is hard to reuse
- not dealing too much with things not related to the engine in the promql package
- let third party implement their own logger if needed or reuse our filelogger package
- easy split of file tests and logger tests

Give me a few days to move forward with this.

roidelapluie:
I have updated my pull request, with a better separation of concern. The promql package does not deal with files anymore.

/cc @brian-brazil @bwplotka

web/api/v1/api_test.go
```diff
@@ -1466,9 +1466,12 @@ func testEndpoints(t *testing.T, api *API, tr *testTargetRetriever, testLabelAPI
 		if m == http.MethodPost {
 			r, err := http.NewRequest(m, "http://example.com", strings.NewReader(q.Encode()))
 			r.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+			r.RemoteAddr = "127.0.0.1:20201"
```
brian-brazil: Where is this tested?

promql/engine.go
```diff
@@ -307,6 +334,36 @@ func NewEngine(opts EngineOpts) *Engine {
 	}
 }
 
+// UnsetQueryLogger disables the query logger.
+func (ng *Engine) UnsetQueryLogger() error {
```
brian-brazil: Is this function really needed? It's not really saving anything.

promql/engine.go
brian-brazil: I'm not seeing where you added this.

web/api/v1/api_test.go
roidelapluie: It is not tested but tests fail without this. Not sure how to test this as it is passed to the context in a private type that is not usable in this package (only usable in promql).

promql/engine.go
roidelapluie: ooh now I get it. You want to check if the query fails. On it.

promql/engine.go
```diff
@@ -379,6 +436,36 @@ func (ng *Engine) exec(ctx context.Context, q *query) (Value, storage.Warnings,
...
+			case *EvalStmt:
+				f = append(f,
+					"start", formatDate(eq.Start),
+					"end", formatDate(eq.End),
+					"step", eq.Interval.String(),
+				)
+			case testStmt:
+				f = append(f, "test", "yes")
+			default:
+				panic(errors.Errorf("can't log %v", eq))
```
roidelapluie: @brian-brazil should we panic here ? It might be that third party implementations pass objects that also follow the Query() interface, then the engine would panic for them.

promql/engine.go
roidelapluie: removed

promql/engine.go
brian-brazil: Probably best not to, though it'd be the Statement interface which 3rd parties shouldn't need.

But on the other hand why do you log tests? They can only happen in promtool.

promql/engine.go
roidelapluie: I added it.

web/api/v1/api_test.go
brian-brazil: Why do they fail?

roidelapluie: We fail the api call if net.SplitHostPort(r.RemoteAddr) fails.

promql/engine.go
roidelapluie: Removed logging.

roidelapluie:
Exemple log line:
{"client_ip":"127.0.0.1","end":"2020-01-07T13:54:16.128Z","query":"vector(1)","start":"2020-01-07T13:54:16.128Z","stats":{"timings":{"evalTotalTime":0.000026106,"resultSortTime":0,"queryPreparationTime":0.000012256,"innerEvalTime":0.000007964,"execQueueTime":0.000002085,"execTotalTime":0.000041266}},"step":"0s","ts":"2020-01-07T13:54:16.160Z"}


With error:
{"client_ip":"127.0.0.1","end":"2020-01-07T13:56:19.132Z","error":"found duplicate series for the match group {} on the right hand-side of the operation: [{__name__=\"up\", instance=\"127.0.1.2:9090\", job=\"prometheus\"}, {__name__=\"up\", instance=\"127.0.1.1:9090\", job=\"prometheus\"}];many-to-many matching not allowed: matching labels must be unique on one side","query":"vector(9999) + on () up","start":"2020-01-07T13:56:19.132Z","stats":{"timings":{"evalTotalTime":0.000224988,"resultSortTime":0,"queryPreparationTime":0.000078192,"innerEvalTime":0,"execQueueTime":0.000003185,"execTotalTime":0.000239282}},"step":"0s","ts":"2020-01-07T13:56:19.169Z"}

roidelapluie:
Should we use a dedupe logger for the errors? That would imply to add another logger to the engine, specifically for the query log errors.

brian-brazil:
Dedupe logger loses us line numbers, and disk filling up is fatal to Prometheus unlike remote storage going away.

promql/engine.go
```diff
@@ -180,15 +193,13 @@ func (q *query) Exec(ctx context.Context) *Result {
 	var queryIndex int
 	if q.ng.activeQueryTracker != nil {
 		queryIndex = q.ng.activeQueryTracker.Insert(q.q)
+		// Delete query from active log.
```
brian-brazil: This comment isn't needed now, as all the code is together.

promql/engine_test.go
```diff
@@ -1100,3 +1100,136 @@ func TestSubquerySelector(t *testing.T) {
...
+		ctx, cancelCtx := context.WithCancel(context.Background())
+		defer cancelCtx()
+		query := engine.newTestQuery(func(ctx context.Context) error {
+			return contextDone(ctx, "test statement execution")
+		})
+		res := query.Exec(ctx)
+		testutil.Ok(t, res.Err)
+	}
+
+	// Query works without query log initalized
```
brian-brazil: Full stops

promql/engine.go
roidelapluie: done, I also re-scoped queryIndex.

promql/engine_test.go
roidelapluie: done

brian-brazil:
Thanks!

promql/engine.go
```diff
@@ -307,6 +332,31 @@ func NewEngine(opts EngineOpts) *Engine {
 	}
 }
 
+// SetQueryLogger sets the query logger.
+func (ng *Engine) SetQueryLogger(l QueryLogger) error {
```
yeya24: Can we just drop this return value? Seems it only returns nil

docs/stability.md
```diff
@@ -30,6 +30,7 @@ Things considered unstable for 2.x:
 * HTML generated by the web UI
 * The metrics in the /metrics endpoint of Prometheus itself
 * Exact on-disk format. Potential changes however, will be forward compatible and transparently handled by Prometheus
+* The format of the logs
```
brancz: This being a new feature, do we really want to commit to this immediately?

docs/configuration/configuration.md
```diff
@@ -66,6 +66,10 @@ global:
   external_labels:
     [ <labelname>: <labelvalue> ... ]
 
+  # File to which PromQL queries are logged.
+  # Reloading the configuration will reopen the file.
+  [ query_log_file: <string> ]
```
brancz: If I recall correctly we discussed at the devsummit that we should document how to allow people who want to to have this be part of the stdout log stream and have some docs around it. I assume that it's just a matter of setting this to `/dev/stdout`, I just wonder about stream syncronization.

docs/stability.md
roidelapluie: It is in the list of things that we consider unstable and don't commit to

docs/configuration/configuration.md
brian-brazil: Probably best for other docs. Does `CON` still exist on Windows?

docs/stability.md
brian-brazil: More particularly the list of things we'll never commit to

brancz: My mistake sorry.

docs/configuration/configuration.md
roidelapluie: Is it worth a complete section? under Configuration? under Querying?

op

brian-brazil: Querying is for the PromQL language.

It wouldn't be part of the reference docs at all, this is generic Unix/Windows knowledge.

roidelapluie: Should it be a small post on the blog, with example with logrotate etc ?

roidelapluie: (blog post would be published after the release)

brancz: I think a blog with examples for logrotate and stdout sounds like a great idea, or a user guide.

brancz: I think a blog with examples for logrotate and stdout sounds like a great idea, or a user guide.

roidelapluie:
On 09 Jan 00:54, Frederic Branczyk wrote:
> I think a blog with examples for logrotate and stdout sounds like a great idea, or a user guide.

Oh yes a user guide sounds even better.

Issue #5218, opened 2019-02-14
delta function doing full vector scan


bla-bu:
## Bug Report
**What did you do?**
Running PromQL queries with millions of data points and using delta function.

**What did you expect to see?**
Accordingly to the documentation, “delta(v range-vector) calculates the difference between the first and last value of each time series element in a range vector v, returning an instant vector with the given deltas and equivalent labels.[…] delta should only be used with gauges.”. However, looking at the code (functions.go), it seems that delta is doing full scan anyway:
```
for _, sample := range samples.Points {
  if isCounter && sample.V < lastValue {
	counterCorrection += lastValue
  }
  lastValue = sample.V
}
```
instead of simply taking the last value from the slice. 

I would expect something like this (taken form the issue [https://github.com/prometheus/prometheus/issues/3746](url))
```
if isCounter {
  for i := firstPoint; i < len(points); i++ {
    sample := points[i]
    if sample.V < lastValue {
	counterCorrection += lastValue
    }
    lastValue = sample.V
  }
}
resultValue := points[len(points)-1].V - points[firstPoint].V + counterCorrection
```

**Environment**

* System information:
	Linux 4.15.0-1026-gcp x86_64

* Prometheus version:
	prometheus, version 2.6.0 (branch: HEAD, revision: dbd1d58c894775c0788470944b818cc724f550fb)
	build user:       root@bf5760470f13
	build date:       20181217-15:14:46
	go version:       go1.11.3

beorn7:
IIRC we need the full scan for the extrapolation algorithm, i.e. to find out if a time series starts or stops within the range.

brian-brazil:
The length of the slice, plus the start/stop timestamps should be enough for what we need to do.

What are you using delta for?

bla-bu:
We have metrics representing “budget”, they are gauges. Our users are often interested in how much budget they spent over e.g. last 15 days (single value). The query using delta function looks very natural, however, it is rather slow (in our set up). We are pretty sure we can go with a query where we are simply subtracting two value using offset, the question here is rather whether the delta should really do a full scan (especially when the documentation says “calculates the difference between the first and last value”)

Cheers,

brian-brazil:
What delta does is a bit more complicated, it sounds like you want offset
or increase.

bla-bu:
[...]

PR #6521, opened 2019-12-25
Optimize Delta

roidelapluie:
As described in #5218.

This pull request is open so that we can decide between merging this pull request or closing #5218.

### DIFF
--- a/promql/bench_test.go
+++ b/promql/bench_test.go
@@ -109,6 +109,9 @@ func BenchmarkRangeQuery(b *testing.B) {
 			expr: "changes(a_X[1d])",
 		},
 		{
+			expr: "delta(a_X[1d])",
+		},
+		{
 			expr: "rate(a_X[1d])",
 		},
 		// Unary operators.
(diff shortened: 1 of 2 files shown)


brian-brazil:
I'd ask why we should optimise delta, its use cases seem extremely limited to the point that I'd yet to see someone who shouldn't be using either deriv or increase instead.

roidelapluie:
This pull request makes the code ugly, but it might also optimize the other functions by checking isCounter just once.

 I use `delta(x[1d])` as shortcut for `x-x offer 1d > 0` (e.g. in disk space alerting : `disk_space_available_percent < 20 and delta(disk_space_used[1d]) > 0` so that disks that are full for a long time don't alert ( we have some disks full on purpose) where I don't need the extra deriv calculation).

That being said, we could optimize the other way around, remove if counter from the original loop, always set the counterCorrection and after the loop for:

```
if !isCounter {
counterCorrection = 0
}
```
 Once again the purpose of this pull request is to get #5218 moving as there is little added value in getting this issue open.

brian-brazil:
> I use delta(x[1d]) > 0 as shortcut for x-x offset 1d > 0

Those queries do non-trivially different things, and you almost certainly want the offset version.

> Once again the purpose of this pull request is to get #5218 moving as there is little added value in getting this issue open.

We could close it.

PR #6530, opened 2019-12-28
Function `last_over_time`

inossidabile:
We are managing IoT with Prometheus in a federated way and it works brilliant. However there's still some problem with Staleness. Our retranslator nodes have unstable connectivity and it's expected to loose connectivity to those. When it happens, the latest value immediately gets marked with a tombstone and it's a great solution in general. This function however is intended for use in more complicated cases where unreliability is expected to a certain degree. E.g. by running `last_over_time(metric[1h])` you are allowing a monitored node (federated Prometheus in our case) to go offline for up to 1 hour. Since it's a very explicit way to override staleness we believe it's a good addition to the built-in vector aggregation family.

Please note that we have read tons of discussions in this regard and we are aware of pushgateway and general staleness concept (nodes should have reliable connection and such) however all those ways require sidecars or intermediate retranslators for 0 reason. For instance, we are working with fish farms and they can't have good connectivity. Also because of physical limitation, the IoT gateway node has to be in the field as well. Currently we implemented a workaround by tunneling data to pushgateway via AMQP but that's really too much :).

P.S. Prometheus is amazing guys!

### DIFF
--- a/promql/functions.go
+++ b/promql/functions.go
@@ -415,6 +415,19 @@ func funcMinOverTime(vals []Value, args Expressions, enh *EvalNodeHelper) Vector
 	})
 }
 
+// === last_over_time(Matrix ValueTypeMatrix) Vector ===
+func funcLastOverTime(vals []Value, args Expressions, enh *EvalNodeHelper) Vector {
+	return aggrOverTime(vals, enh, func(values []Point) float64 {
+		last := values[0]
+		for _, v := range values {
+			if !math.IsNaN(v.V) && (v.T > last.T || math.IsNaN(last.V)) {
+				last = v
+			}
+		}
+		return last.V
+	})
+}
+
 // === sum_over_time(Matrix ValueTypeMatrix) Vector ===
 func funcSumOverTime(vals []Value, args Expressions, enh *EvalNodeHelper) Vector {
 	return aggrOverTime(vals, enh, func(values []Point) float64 {
@@ -1092,6 +1105,12 @@ var Functions = map[string]*Function{
 		ArgTypes:   []ValueType{ValueTypeMatrix},
 		ReturnType: ValueTypeVector,
 		Call:       funcMinOverTime,
+	},
+	"last_over_time": {
+		Name:       "last_over_time",
+		ArgTypes:   []ValueType{ValueTypeMatrix},
+		ReturnType: ValueTypeVector,
+		Call:       funcLastOverTime,
 	},
 	"minute": {
 		Name:       "minute",
--- a/promql/parse_test.go
+++ b/promql/parse_test.go
@@ -1507,6 +1507,30 @@ var testExpr = []struct {
 			},
 		},
 	}, {
+		input: `last_over_time(rate(foo{bar="baz"}[2s])[5m:5s])`,
+		expected: &Call{
+			Func: mustGetFunction("last_over_time"),
+			Args: Expressions{
+				&SubqueryExpr{
+					Expr: &Call{
+						Func: mustGetFunction("rate"),
+						Args: Expressions{
+							&MatrixSelector{
+								Name:  "foo",
+								Range: 2 * time.Second,
+								LabelMatchers: []*labels.Matcher{
+									mustLabelMatcher(labels.MatchEqual, "bar", "baz"),
+									mustLabelMatcher(labels.MatchEqual, string(model.MetricNameLabel), "foo"),
+								},
+							},
+						},
+					},
+					Range: 5 * time.Minute,
+					Step:  5 * time.Second,
+				},
+			},
+		},
+	}, {
 		input: `min_over_time(rate(foo{bar="baz"}[2s])[5m:])[4m:3s]`,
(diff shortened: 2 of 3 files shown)


brian-brazil:
The general Prometheus approach is that if a target is down that much, you have bigger problems. Not wishing to run a sidecar is not a sufficient reason to add features to PromQL, particularly for something that's not typical local metrics scraping use case.

I suspect something like Influx may be a better choice here.

inossidabile:
This approach doesn't always fit reality though. Different federation architectures and use-cases often let a node to be offline for certain periods. It makes me wonder why do you choose to ignore them and suggest hugely more sophisticated workaround (hack?) instead of just letting people explicitly decide which case is theirs?

I understand there's always a supporting price and effort involved but in scope of existing average functions - whatever you will rewrite executor to, implementation of this method will never be complex, right? Just math :).

brian-brazil:
We can't make everyone happy, without also in the process making things so complicated that it makes everyone unhappy. Prometheus is explicitly not an event logging system, so features like this which are solely to serve such use cases are not likely to be accepted.

This sounds closest to a batch job use case, and there's already things like the pushgateway for that.

inossidabile:
I love that we are making the same arguments and go to polar conclusions :). This a very common use-case from reality which add 0 complexity, that's exactly my point. However it's your monumental work and it's up to you to decide. If the idea is that you want to entirely ignore the cases with unstable federation branches - I get it. Thanks for checking this out anyway :).

brian-brazil:
Every additional function is complexity and cognitive overhead for users. I don't think that knowing what a metric was an hour ago is useful, considering you've a major network outage to deal with first.

inossidabile:
Since I saw similar requests here and there on internet, sharing some experience to everyone struggling with the same situation. Here are some workarounds (in order of how bad they are - best first):

**Grafana**
Just don't user alertmanager. Use alerts from Grafana over Prometheus serving as a data source. It's one more service to support but way more flexibility with alerting. And you get dashboards for free :).

**Sidecar push-service**
Instead of direct Prometheus -> Prometheus federation, you need to write a small service that polls required metrics at edge Prometheus and sends it to a central Pushgateway. The good thing with this is that you get push-based data gateway out of box.

**Proxy over unreliable metrics sources**
Query metrics via intermediate service that "hides" timeouts and connection errors so metrics don't go stale.

brian-brazil:
> Just don't user alertmanager. Use alerts from Grafana over Prometheus serving as a data source. 

That doesn't sound right, if Prometheus doesn't have the data then Grafana doesn't either. As far as is possible alert from the Prometheus that is scraping the data, not Prometheus servers up the federation hierarchy. The AM's design will tolerate network outages gracefully, as long as you can get to one that can still get to the internet.

inossidabile:
The proble is not absense of data. The problem is that Prometheus doesn't have a way of getting last effective value in, say, one hour. And Grafana can. Prometheus returns NaN as soon as it hits tombstone and there's no way around it. Once there's a problem with connectivity to edge node, the latest value is empty. With Grafana you can load a period vector (which ignores staleness tombstones) and use additional filtering at Grafana level.

And yes, it doesn't sound right. Right way would be to have a simple way to get last known value ignoring tombstones straight on Prometheus. That's what this PR was :D.

inossidabile:
@brian-brazil imagine this. You have 100 edge nodes that are online for 60% of time. And the connection is constantly going up and down. Now there's a incident on one of those nodes. Next time this node goes offline for 1 minute, Prometheus can't poll it and resolves the incident. Then it goes online and incident pops up again. This is extremely misleading and spammy. And you can't do anything if you are on Prometheus and AlertManager only.

roidelapluie:
You could also use recording rules

record: last_value
expr: label_replace{job="fish",
__name__!="last_value"},"orig_name","$1","__name__","(.+)") or
last_value{job="fish"}

Or if you know your important metrics it is a lot easier:

record: last_metric
expr: metric or last_metric

Le sam. 28 déc. 2019 à 18:06, Boris Staal <notifications@github.com> a
écrit :

alburthoffman:
the above recording rule makes things more complex and wont help in all cases.

last_over_time is just like max_over_time or absent_over_time, simple and straight forward.
It's less complex than using above complex recording rules.

In some cases, there are choices to get rid of fall back behaviors like doing detections by querying metrics out from prometheus.

alburthoffman:
@brian-brazil @roidelapluie

roidelapluie:
I'd personaly be open to review the conclusion of this based on the fact that I did need this ; I found a workaround with recording rules (but that is not possible with subqueries) & that this is has been asked multiple times by the community.

roidelapluie:
Let's revisit this if you are interested. I think that your implementation could be reviewed because PromQL would not receive stale NaN's, which would mean that you do not need to fiter out NaN's as they are actual NaN's. PromQL would also receive values ordered by timestamp, so you could just take the last value (I will check this).

roidelapluie:
This is being worked on in #8457

Issue #1758, opened 2016-06-22
promtool: print file/line locations for warnings/errors


juergenhoetzel:
Checking my config I get warnings, but not the file-name + line-number which cause the warnings/errors:

```
→ ./promtool  check-config /srv/prometheus/prometheus.yml
Checking /srv/prometheus/prometheus.yml
WARN[0000] The 'target_groups' option for scrape configurations is deprecated, use 'static_configs' instead  source=config.go:468
WARN[0000] The 'target_groups' option for scrape configurations is deprecated, use 'static_configs' instead  source=config.go:468
WARN[0000] The 'target_groups' option for scrape configurations is deprecated, use 'static_configs' instead  source=config.go:468
  SUCCESS: 1 rule files found
```

These additional info could also improve editor support. I use Emacs and implementing a 
[Flycheck](https://github.com/flycheck/flycheck) checker should be easy when this info is provided.

fabxc:
I agree that it would be handy but with the YAML parsing package we are using this is hardly doable.

serathius:
I created issue for yaml package https://github.com/go-yaml/yaml/issues/343

PR #6533, opened 2019-12-30
Adds support for line-column numbers for invalid rules, promtool

Harkishen-Singh:
Fixes: #1758 [link](https://github.com/prometheus/prometheus/issues/1758)

This PR uses the help of `gopkg.in/yaml.v3` to implement this functionality.

line-column-number-invalid-rules


rules/manager_test.go
```diff
@@ -769,7 +769,7 @@ func reloadAndValidate(rgs *rulefmt.RuleGroups, t *testing.T, tmpFile *os.File,
 	_, err = tmpFile.Write(bs)
 	testutil.Ok(t, err)
 	err = ruleManager.Update(10*time.Second, []string{tmpFile.Name()}, nil)
-	testutil.Ok(t, err)
+	testutil.NotOk(t, err)
```
brian-brazil: Why does this fail now?

pkg/rulefmt/rulefmt.go
```diff
@@ -193,10 +213,14 @@ func testTemplateParsing(rl *Rule) (errs []error) {
...
 }
 
+var grpsTmp ruleGroups
+
 // Parse parses and validates a set of rules.
 func Parse(content []byte) (*RuleGroups, []error) {
 	var groups RuleGroups
-	if err := yaml.UnmarshalStrict(content, &groups); err != nil {
+	err := yaml.Unmarshal(content, &groups)
+	_err := yaml.Unmarshal(content, &grpsTmp)
```
brian-brazil: Why are you doing this? This global also won't be threadsafe.

pkg/rulefmt/rulefmt.go
```diff
@@ -45,19 +46,23 @@ type RuleGroups struct {
...
+
 // Validate validates all rules in the rule groups.
 func (g *RuleGroups) Validate() (errs []error) {
 	set := map[string]struct{}{}
 
-	for _, g := range g.Groups {
+	for j, g := range g.Groups {
 		if g.Name == "" {
-			errs = append(errs, errors.Errorf("Groupname should not be empty"))
+			errs = append(errs, errors.Errorf("Groupname should not be empty, line: %d, column: %d", grpsTmp.Groups[j].Line, grpsTmp.Groups[j].Column))
```
brian-brazil: Do we have to add the line/col in every error?

Harkishen-Singh: yes. At each error, we can find both the line and column number, which has been implemented here.

pkg/rulefmt/rulefmt.go
Harkishen-Singh: we need another struct var containing similar names but with `yaml.Node`. `yaml.Node` is the only way we can print the line and column number. Hence we need a different var as well to get the line numbers for errors like `similar groupnames` and `empty groupname`.

rules/manager_test.go
Harkishen-Singh: Returned now is along with line and column number. I haven't been able to figure this out yet. Can you help me with this?

pkg/rulefmt/rulefmt.go
brian-brazil: This isn't safe, nor is it good practice. Pass it via the objects you're returning.

Harkishen-Singh:
Straight switch to v3 through the entire project will cause some test failures since the new package does not support `UnmarshalStrict`. Files using such will experience CI fails. @brian-brazil

slrtbtfs:
@Harkishen-Singh FYI, when using the [PromQL language server](https://github.com/slrtbtfs/promql-lsp) you already get some of the functionality of this PR.

pkg/rulefmt/rulefmt.go
```diff
@@ -45,19 +46,23 @@ type RuleGroups struct {
...
 // Validate validates all rules in the rule groups.
-func (g *RuleGroups) Validate() (errs []error) {
+func (g *RuleGroups) Validate(node ruleGroups) (errs []error) {
 	set := map[string]struct{}{}
 
-	for _, g := range g.Groups {
+	for j, g := range g.Groups {
 		if g.Name == "" {
-			errs = append(errs, errors.Errorf("Groupname should not be empty"))
+			errs = append(errs, errors.Errorf("Groupname should not be empty, line: %d, column: %d", node.Groups[j].Line, node.Groups[j].Column))
```
slrtbtfs: Prometheus uses the `<line>:<col>: <Error message>` or  `<filename>:<line>:<col>: <Error message>` format for errors in the expression browser now, since that format is used by many compilers. Also, when using that format, VS Code allows to just click on the message to get the cursor to the error position.

Might make sense to do the same here.

Harkishen-Singh:
@slrtbtfs I don't find anything related to rule checking and validation from the readme. Please correct me if I am wrong.

pkg/rulefmt/rulefmt.go
```diff
@@ -101,29 +106,44 @@ type Rule struct {
...
 
-	if r.Expr == "" {
-		errs = append(errs, errors.Errorf("field 'expr' must be set in rule"))
-	} else if _, err := promql.ParseExpr(r.Expr); err != nil {
-		errs = append(errs, errors.Wrap(err, "could not parse expression"))
+	if r.Expr.Value == "" {
+		fmt.Println(r.Expr.Line)
+		errs = append(errs, errors.Errorf("field 'expr' must be set in rule, line: %d, column: %d", r.Expr.Line, r.Expr.Column))
+	} else if _, err := promql.ParseExpr(r.Expr.Value); err != nil {
+		errs = append(errs, errors.Wrapf(err, "could not parse expression, line: %d, column: %d", r.Expr.Line, r.Expr.Column))
```
slrtbtfs: This should end up looking quite strange, since the error returned by the parser will include position information, too.

Normally the parser returns a `promql.ParseErr`. This struct already includes the position information.

What the PromQL language server does here, is to add the line and column where the expression starts to the line and column in the error message and then simply convert the error message to a string

slrtbtfs:
That's what is referred to as "YAML support" in the README.

However you're right that this should be explained better.

Relevant issue, tracking part of what is done in this PR, too: https://github.com/slrtbtfs/promql-lsp/issues/75

To be clear, ATM config file validation is still WIP.

pkg/rulefmt/rulefmt.go
Harkishen-Singh: Done

Harkishen-Singh:
@slrtbtfs I did not understand the purpose of the comment. Language server has a different purpose. Configuration checks by promtool are the general implementation which can be used by anyone, irrespective of the IDEs to ensure the proper functioning of the rules and alerts. Need not everyone to install the Prometheus language server.

pkg/rulefmt/rulefmt.go
Harkishen-Singh: We need the position coordinates to make debugging rules easier for the user, in case of a larger file, increasing the convenience. Similar behaviour is noted in various format checkers as well.

slrtbtfs:
Sure, this PR is useful independently of whether there exists a language server or not.

The point was mainly, that the use case of having better IDE support which was demoed in your screencast is also covered by the language server and pointing out that there exists other code that tries to achieve similar things.

Harkishen-Singh:
Sure. Both have their own purpose. Having said that language-server is optional for a user, its good to have it.

pkg/rulefmt/rulefmt.go
slrtbtfs: That's clear.

The problem with the current approach is that the output will look roughly like
`could not parse expression, line: 3, column: 10: 2:14: Unexpected identifier in label matcher, expected string`. 

It would be better if the output looked like:
`rules.yaml:4:24: could not parse expression: Unexpected identifier in label matcher, expected string`. Note that the positions have been added together here.

Harkishen-Singh: The current implement is, (from live example)

```
rules.yml: Groupname should not be empty, line: 2, column: 5
rules.yml: group "recording-node-exporter", rule 1, "heap": field 'expr' must be set in rule, line: 13, column: 12
```
is this acceptable? @brian-brazil

Harkishen-Singh: can you post your rules file here? @slrtbtfs

slrtbtfs: This wasn't an actual output, but rather behaviour guessed from the code. The cases where the problem I described should occur, is if there is a non-empty, syntactically wrong query inside the file being checked.

slrtbtfs: I've constructed a real example:

Using this `rules.yaml`:

```
groups:
    - name: recording
      rules:
          - expr: |
              go_goroutines}
            record: foo
```

The output with your patches applied is:

```
Checking rules.yaml
  FAILED:
rules.yaml: group "recording", rule 0, "foo": could not parse expression, line: 4, column: 19: 1:14: parse error: could not parse remaining input "}\n"...
```

Harkishen-Singh: thanks. I will have a look over this.

Harkishen-Singh: I have updated. The current version gives the following output for the above config,

```
Checking ../../rules.yml
  FAILED:
../../rules.yml: 6:21 group "recording", rule 0, "foo": 4:19 could not parse expression: 1:14: parse error: could not parse remaining input "}\n"...

```

and for the initial example ,
```
  FAILED:
../../rules.yml: 2:5 Groupname should not be empty
../../rules.yml: 12:15 group "recording-node-exporter", rule 1, "heap": 13:12 field 'expr' must be set in rule
```

pkg/rulefmt/rulefmt.go
```diff
@@ -34,48 +35,54 @@ type Error struct {
 	Rule     int
 	RuleName string
 	Err      error
+	Node     yaml.Node
 }
 
 func (err *Error) Error() string {
```
brian-brazil: Why are we wrapping an error, and then unwrapping it again by returning a string?

rules/manager_test.go
brian-brazil: Not offhand, it'd take debugging.

pkg/rulefmt/rulefmt.go
slrtbtfs: Thanks. Looks better now.

---

> `../../rules.yml: 2:5`

That space should not be there and there should be a colon at the end of the position. To be consistent with conventions from various compilers the position should be of the form
`../../rules.yml:2:5:`

(nit)

----

> `../../rules.yml: 5:21 group "recording", rule 0, "foo": 4:19 could not parse expression: 1:14: parse error: could not parse remaining input "}\n"...`

This one looks really confusing, since a single error message contains three different position coordinates and a verbal description of the position.

Ideally there would be only one position, directly at the beginning of the error line. Something like

 `../../rules.yml:5:28: group "recording", rule 0, "foo": could not parse expression: parse error: could not parse remaining input "}\n"...`

Maybe even something more customized like:
`../../rules.yml:5:28: Failed to parse PromQL expression: parse error: could not parse remaining input "}\n"...`

pkg/rulefmt/rulefmt.go
Harkishen-Singh: errors returned by the parser need a `yaml.Node` wrapper as well. Hence for those cases, I am returning a Node which is then unwrapped to string at the time of return.

pkg/rulefmt/rulefmt.go
Harkishen-Singh: As the design of the rulefmt, this looks tough. If we remove the parser checks and let it handle by itself as earlier, then this can be possible, but again on the right-hand side, towards the ending of the sentence.

slrtbtfs: It should be possible by just typecasting the `err` result of `promql.ParseExpr()` to a `promql.ParseErr` and the using the fields in that data structure.

rules/manager_test.go
Harkishen-Singh: @brian-brazil  When we marshal a struct containing the `yaml.Node` type, the []byte received contains additional information from the node. This is done in the `reloadAndValidate()`. This will create an error from the code since the output would be like:
```
groups:
  - name: test
    interval: 0s
    rules:
      - record:
            kind: 8
            style: 0
            tag: '!!str'
            value: job:http_requests:rate5m
            anchor: ""
            alias: null
            content: []
            headcomment: ""
            linecomment: ""
            footcomment: ""
            line: 4
            column: 15
        expr:
            kind: 8
            style: 0
            tag: '!!str'
            value: sum by (job)(rate(http_requests_total[5m]))
            anchor: ""
            alias: null
            content: []
            headcomment: ""
            linecomment: ""
            footcomment: ""
            line: 5
            column: 13
```
_Note: record and expr are `yaml.Node`._

Hence when we Parse this to rulefmt, this will return an error since `record:` and `expr:` are empty fields here and hence I used `testutil.NotOk()`

pkg/rulefmt/rulefmt.go
Harkishen-Singh: The space before the number is due to `errors.Wrapf()`. I don't think there is any way to remove this space. Any idea @slrtbtfs ?

Leaving that, rest I have given a new implementation of the logic. That solves most of what you said.

Harkishen-Singh:
@brian-brazil please review.

pkg/rulefmt/rulefmt.go
slrtbtfs: By not using `Wrapf()` but instead constructing a custom error message from the data contained in the error.

pkg/rulefmt/rulefmt.go
```diff
@@ -101,55 +116,104 @@ type Rule struct {
...
+	For         model.Duration    `yaml:"for,omitempty"`
+	Labels      map[string]string `yaml:"labels,omitempty"`
+	Annotations map[string]string `yaml:"annotations,omitempty"`
+}
+
 // Validate the rule and return a list of encountered errors.
-func (r *Rule) Validate() (errs []error) {
-	if r.Record != "" && r.Alert != "" {
-		errs = append(errs, errors.Errorf("only one of 'record' and 'alert' must be set"))
+func (r *RuleNode) Validate() (errs []WrapError) {
```
brian-brazil: You only use this as an `error`

rules/manager_test.go
brian-brazil: That sounds like we need a better test then, as this one is no longer doing what it's meant to be doing

pkg/rulefmt/rulefmt.go
Harkishen-Singh: I am wrapping this error with the node values so that we can implement according to what slrbfts said. Do you mean to remove the wrapper logic and use the earlier error logic?

brian-brazil: No, I mean the return type here should be a slice of error.

Harkishen-Singh: @brian-brazil  If I return a slice of error here, then the suggestion given by slrbtfs would not be possible. Please correct me if I am wrong.

rules/manager_test.go
Harkishen-Singh: I have fixed this by formatting the rules to the yaml v2 standards.

Harkishen-Singh:
@brian-brazil please have a look.

roidelapluie:
Do we have the same behaviour as before? e.g. the UnmarshalStrict behaviour?

Harkishen-Singh:
I do think yes. Since when we remove the `yaml.Node` then the entire things works like the v2 yaml.

pkg/rulefmt/rulefmt.go
Harkishen-Singh: @brian-brazil any suggestions?

pkg/rulefmt/rulefmt.go
```diff
@@ -33,48 +33,70 @@ type Error struct {
...
+	return errors.Wrapf(err.Err.err, "group %q, rule %d, %q", err.Group, err.Rule, err.RuleName).Error()
 }
 
 // RuleGroups is a set of rule groups that are typically exposed in a file.
 type RuleGroups struct {
 	Groups []RuleGroup `yaml:"groups"`
 }
 
+// RuleGroupsTest for running tests over rules.
+type RuleGroupsTest struct {
```
brian-brazil: This is unused.

pkg/rulefmt/rulefmt.go
brian-brazil: Error is an interface, it's weird to refer to things as errors that don't meet that interface. So I suggest adjusting things so that your errors are errors.

pkg/rulefmt/rulefmt.go
Harkishen-Singh: This is actually used inside the `manager_test.go`. It helps to convey `yaml.Node` type into default strings type so that Marshalling can be done according to the tests.

brian-brazil: If it's only used in tests, then it belongs in the test file.

Harkishen-Singh: Sure. I will update that.

pkg/rulefmt/rulefmt.go
slrtbtfs: > @brian-brazil If I return a slice of error here, then the suggestion given by slrbtfs would not be possible. Please correct me if I am wrong.

It is possible.

You can return an error and then convert it back to the underlying concrete type when needed.

https://tour.golang.org/methods/15

Harkishen-Singh: > Error is an interface, it's weird to refer to things as errors that don't meet that interface. So I suggest adjusting things so that your errors are errors.

With the current design, the errors are being returned after unwrapping, done by the Error interface function. Only after that, it is an error. I think the `errs` is leading to confusion even though its a slice of wrapped error along with its node value. Is this any case?

pkg/rulefmt/rulefmt.go
brian-brazil: This still seems to be here

pkg/rulefmt/rulefmt.go
```diff
@@ -86,6 +108,13 @@ func (g *RuleGroups) Validate() (errs []error) {
 
 // RuleGroup is a list of sequentially evaluated recording and alerting rules.
 type RuleGroup struct {
+	Name     string         `yaml:"name"`
+	Interval model.Duration `yaml:"interval,omitempty"`
+	Rules    []RuleNode     `yaml:"rules"`
+}
+
+// RuleGroupTest forms a testing struct for running tests over rules.
+type RuleGroupTest struct {
```
brian-brazil: This one too.

Harkishen-Singh: @brian-brazil done. Please have a look.

brian-brazil:
Thanks!

PR #6537, opened 2019-12-31
Rw testability improvements

cstyan:
@csmarchbanks spent yesterday and today trying to write a test that would replicate the situation where we fall behind and never catch up, just get equal in/out rates. Haven't been able to replicate that situation in a test yet, but along the way I ran into a few things that I wanted to overwrite/set in tests.

### DIFF
--- a/storage/remote/queue_manager_test.go
+++ b/storage/remote/queue_manager_test.go
@@ -408,10 +408,10 @@ func (c *TestStorageClient) waitForExpectedSamples(tb testing.TB) {
 	}
 }
 
-func (c *TestStorageClient) expectSampleCount(ss []record.RefSample) {
+func (c *TestStorageClient) expectSampleCount(numSamples int) {
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
-	c.wg.Add(len(ss))
+	c.wg.Add(numSamples)
 }
 
 func (c *TestStorageClient) waitForExpectedSampleCount() {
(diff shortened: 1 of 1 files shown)


storage/remote/queue_manager.go
```diff
@@ -234,7 +236,7 @@ type QueueManager struct {
 }
 
 // NewQueueManager builds a new QueueManager.
-func NewQueueManager(reg prometheus.Registerer, logger log.Logger, walDir string, samplesIn *ewmaRate, cfg config.QueueConfig, externalLabels labels.Labels, relabelConfigs []*relabel.Config, client StorageClient, flushDeadline time.Duration) *QueueManager {
+func NewQueueManager(reg prometheus.Registerer, logger log.Logger, walDir string, samplesIn *ewmaRate, cfg config.QueueConfig, externalLabels labels.Labels, relabelConfigs []*relabel.Config, client StorageClient, flushDeadline time.Duration, shardUpdateDuration time.Duration) *QueueManager {
```
csmarchbanks: It doesn't look like you are actually setting `shardUpdateDuration` on the `QueueManager` anywhere.

cstyan: I'm not in this PR, but I wanted to and was trying to with the test I was talking about in the original PR message. I wanted a test that went through the whole motion of samples being written to the WAL, read from the WAL and queued up, and actually sent. Without lowering some values, such as shard update duration from 10s (to 1s for example) the test would need to run for minutes to generate useful data.

jackzhou121:
offer已经收到，非常期待在贵公司的工作～

Best regards
Jack

在 2020年1月2日星期四，Chris Marchbanks <notifications@github.com> 写道：

storage/remote/queue_manager.go
csmarchbanks: I would say wait until there is a working test using the change before committing it. Otherwise it is a lot of changes that may or may not actually get used in the near future.

cstyan:
okay @csmarchbanks finally had some time to look at this again I removed the`shardUpdateDuration` change and tweaked a few more tests

PR #6538, opened 2020-01-01
Multi index cleanup

brian-brazil:
Some code cleanup, that also happens to be a little faster.


brian-brazil:
Rebased and conflicts handled.

brian-brazil:
Ah, I'd not commited that change.

PR #6539, opened 2020-01-01
Optimise labels.Compare

brian-brazil:
strings.Compare isn't meant to be used, and this way we save one
comparison which is thus very slightly cheaper.

benchmark                                     old ns/op     new ns/op delta
BenchmarkPostingsForMatchers/Head/n="1"-4     236305440     233515705 -1.18%

### DIFF
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -18,7 +18,6 @@ import (
 	"encoding/json"
 	"sort"
 	"strconv"
-	"strings"
 
 	"github.com/cespare/xxhash"
 )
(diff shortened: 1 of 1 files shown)


roidelapluie:
can we also microoptimize
if lb := len(b); lb < l {
 l = lb
}
I doubt that it would make a big difference however.


anyway, LGTM.

pkg/labels/labels.go
```diff
@@ -285,11 +284,19 @@ func Compare(a, b Labels) int {
 	}
 
 	for i := 0; i < l; i++ {
-		if d := strings.Compare(a[i].Name, b[i].Name); d != 0 {
-			return d
+		if a[i].Name != b[i].Name {
+			if a[i].Name < b[i].Name {
+				return -1
+			} else {
```
codesome: We can drop the else block and just have return outside in both places. Not a blocker.

brian-brazil: In this case I think it's clearer this way.

brian-brazil:
> I doubt that it would make a big difference however.

It actually makes it slower, which is surprising.

PR #6540, opened 2020-01-02
Fix findSetMatch for regex query

yeya24:

Matcher.Value is not wrapped by `^(?: )$`.
Add a method to return the regex string.

### DIFF
--- a/pkg/labels/matcher.go
+++ b/pkg/labels/matcher.go
@@ -110,3 +110,11 @@ func (m *Matcher) Inverse() (*Matcher, error) {
 	}
 	panic("labels.Matcher.Matches: invalid match type")
 }
+
+// GetRegexString returns the regex string.
+func (m *Matcher) GetRegexString() string {
+	if m.re == nil {
+		return ""
+	}
+	return m.re.String()
+}
--- a/tsdb/querier.go
+++ b/tsdb/querier.go
@@ -424,7 +424,7 @@ func postingsForMatcher(ix IndexReader, m *labels.Matcher) (index.Postings, erro
 
 	// Fast-path for set matching.
 	if m.Type == labels.MatchRegexp {
-		setMatches := findSetMatches(m.Value)
+		setMatches := findSetMatches(m.GetRegexString())
 		if len(setMatches) > 0 {
 			sort.Strings(setMatches)
 			return ix.Postings(m.Name, setMatches...)
--- a/tsdb/querier_test.go
+++ b/tsdb/querier_test.go
@@ -1641,21 +1641,21 @@ func BenchmarkSetMatcher(b *testing.B) {
 			numSeries:                   1,
 			numSamplesPerSeriesPerBlock: 10,
 			cardinality:                 100,
-			pattern:                     "^(?:1|2|3|4|5|6|7|8|9|10)$",
+			pattern:                     "1|2|3|4|5|6|7|8|9|10",
 		},
 		{
 			numBlocks:                   1,
 			numSeries:                   15,
 			numSamplesPerSeriesPerBlock: 10,
 			cardinality:                 100,
-			pattern:                     "^(?:1|2|3|4|5|6|7|8|9|10)$",
+			pattern:                     "1|2|3|4|5|6|7|8|9|10",
 		},
 		{
 			numBlocks:                   1,
 			numSeries:                   15,
 			numSamplesPerSeriesPerBlock: 10,
 			cardinality:                 100,
-			pattern:                     "^(?:1|2|3)$",
+			pattern:                     "1|2|3",
 		},
 		// Big data sizes benchmarks.
 		{
@@ -1663,14 +1663,14 @@ func BenchmarkSetMatcher(b *testing.B) {
 			numSeries:                   1000,
 			numSamplesPerSeriesPerBlock: 10,
 			cardinality:                 100,
-			pattern:                     "^(?:1|2|3)$",
+			pattern:                     "1|2|3",
 		},
 		{
 			numBlocks:                   20,
 			numSeries:                   1000,
 			numSamplesPerSeriesPerBlock: 10,
 			cardinality:                 100,
-			pattern:                     "^(?:1|2|3|4|5|6|7|8|9|10)$",
+			pattern:                     "1|2|3|4|5|6|7|8|9|10",
 		},
 		// Increase cardinality.
 		{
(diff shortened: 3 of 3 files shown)


roidelapluie:
What are you trying to achieve?

yeya24:
@roidelapluie @codesome  I updated my pr, could you please review it? Thanks

brian-brazil:
If this is so, how did we not spot it in benchmarks?

pkg/labels/matcher.go
```diff
@@ -110,3 +110,8 @@ func (m *Matcher) Inverse() (*Matcher, error) {
 	}
 	panic("labels.Matcher.Matches: invalid match type")
 }
+
+// GetRegex returns the regex string.
+func (m *Matcher) GetRegex() string {
```
codesome: We should check for `m.re` being `nil` and return `""` in that case to avoid any panics.

PS: `GetRegexString` would be better IMO.

yeya24: Done

codesome:
> If this is so, how did we not spot it in benchmarks?

Looking at the `BenchmarkSetMatcher` added in https://github.com/prometheus/tsdb/pull/602 , I see that the anchored string was passed in the matcher. So it turned out to be a double anchor?

@yeya24, in that case, the changes that you did to the tests earlier were fine, the anchors should be removed from `BenchmarkSetMatcher` test cases. Could you also show us a comparison of before and after this change to the benchmark?

brian-brazil:
Sounds like an additional test and/or benchmark to check that we're not applying this code when double anchored (it's safe, but not what the code is meant to do) would be a good idea.

yeya24:
@codesome I tested and there is no big difference.
In prometheus/tsdb#602, we create label matcher like `labels.MustNewMatcher(labels.MatchRegexp, "test", "^(?:1|2|3|4|5|6|7|8|9|10)$")`, so we will get a matcher with a single anchored value and double anchored regex. And in func `findSetMatch` we use the m.Value directly so it works well.

In this pr, I create the label matcher without anchor, and I will finally get a single anchor regex in func `findSetMatch`. So I guess from benchmark it is almost same

yeya24:
I have to say the benchmark result is strange, seems it is not correct.
For double anchor value:
```
goos: darwin
goarch: amd64
pkg: github.com/prometheus/prometheus/tsdb
BenchmarkSetMatcher/nSeries=1,nBlocks=1,cardinality=100,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-16         	  993950	      1234 ns/op	     688 B/op	      12 allocs/op
BenchmarkSetMatcher/nSeries=15,nBlocks=1,cardinality=100,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-16        	  874772	      1177 ns/op	     688 B/op	      12 allocs/op
BenchmarkSetMatcher/nSeries=15,nBlocks=1,cardinality=100,pattern="^(?:1|2|3)$"-16                       	  910926	      1224 ns/op	     688 B/op	      12 allocs/op
BenchmarkSetMatcher/nSeries=1000,nBlocks=20,cardinality=100,pattern="^(?:1|2|3)$"-16                    	   93766	     13872 ns/op	   10896 B/op	     205 allocs/op
BenchmarkSetMatcher/nSeries=1000,nBlocks=20,cardinality=100,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-16     	   87012	     13293 ns/op	   10896 B/op	     205 allocs/op
BenchmarkSetMatcher/nSeries=100000,nBlocks=1,cardinality=100000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-16 	 1000000	      1304 ns/op	     688 B/op	      12 allocs/op
BenchmarkSetMatcher/nSeries=500000,nBlocks=1,cardinality=500000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-16 	  929360	      1331 ns/op	     688 B/op	      12 allocs/op
BenchmarkSetMatcher/nSeries=500000,nBlocks=10,cardinality=500000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-16         	  167353	      7146 ns/op	    5616 B/op	     105 allocs/op
BenchmarkSetMatcher/nSeries=1000000,nBlocks=1,cardinality=1000000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-16        	 1000000	      1260 ns/op	     688 B/op	      12 allocs/op
PASS
ok  	github.com/prometheus/prometheus/tsdb	332.182s
```

For single anchor value, even slower than the previous one. 
```
goos: darwin
goarch: amd64
pkg: github.com/prometheus/prometheus/tsdb
BenchmarkSetMatcher/nSeries=1,nBlocks=1,cardinality=100,pattern="1|2|3|4|5|6|7|8|9|10"-16         	  504404	      2479 ns/op	    1456 B/op	      35 allocs/op
BenchmarkSetMatcher/nSeries=15,nBlocks=1,cardinality=100,pattern="1|2|3|4|5|6|7|8|9|10"-16        	  461400	      2491 ns/op	    1456 B/op	      35 allocs/op
BenchmarkSetMatcher/nSeries=15,nBlocks=1,cardinality=100,pattern="1|2|3"-16                       	  712063	      1515 ns/op	     872 B/op	      19 allocs/op
BenchmarkSetMatcher/nSeries=1000,nBlocks=20,cardinality=100,pattern="1|2|3"-16                    	   58981	     20282 ns/op	   14576 B/op	     345 allocs/op
BenchmarkSetMatcher/nSeries=1000,nBlocks=20,cardinality=100,pattern="1|2|3|4|5|6|7|8|9|10"-16     	   31347	     35989 ns/op	   26256 B/op	     665 allocs/op
BenchmarkSetMatcher/nSeries=100000,nBlocks=1,cardinality=100000,pattern="1|2|3|4|5|6|7|8|9|10"-16 	  461697	      2502 ns/op	    1456 B/op	      35 allocs/op
BenchmarkSetMatcher/nSeries=500000,nBlocks=1,cardinality=500000,pattern="1|2|3|4|5|6|7|8|9|10"-16 	  468208	      2576 ns/op	    1456 B/op	      35 allocs/op
BenchmarkSetMatcher/nSeries=500000,nBlocks=10,cardinality=500000,pattern="1|2|3|4|5|6|7|8|9|10"-16         	   59592	     20202 ns/op	   13296 B/op	     335 allocs/op
BenchmarkSetMatcher/nSeries=1000000,nBlocks=1,cardinality=1000000,pattern="1|2|3|4|5|6|7|8|9|10"-16        	  431367	      2659 ns/op	    1456 B/op	      35 allocs/op
PASS
ok  	github.com/prometheus/prometheus/tsdb	315.843s
```

yeya24:
I re-check about this problem. This bug is introduced in PR #6326, in release 2.14.0 it works well.

brian-brazil:
Thinking on it a unittest using a mock indexreader that tracks if LabelValues is called would cover testing that this is working as expected.

yeya24:
@brian-brazil  Sorry for the delay, I have added the unit test to make sure this work as expected. Could you please give a review? It would be good to include this fix in the new release.

brian-brazil:
Thanks!

Issue #6377, opened 2019-11-26
Flush Head chunks to disk


codesome:
The idea is to flush the head chunks to disk as soon as it is full and use it for the following purposes

1. mmap the flushed chunks in the Head instead of holding it in the memory, hence reducing the memory footprint of Head
2. Consider those chunks as checkpoints and process WAL accordingly, hence reducing the WAL replay time heavily

I had attempted checkpointing the chunks and replaying it [here](https://github.com/prometheus/prometheus/pull/6059), but the idea has evolved since then (as described above).

I will come up with a design doc to comment on in the coming weeks.

/cc @brancz @krasi-georgiev

Edit: The high-level design doc https://docs.google.com/document/d/184urkLQnM7rqLmGvS66I15jU2pyPk_qwd8v_qDlXczo/edit?usp=sharing

brian-brazil:
The completed chunks aren't sufficient as a checkpoint, you'd also need to checkpoint all non-completed chunks. You can then replay the WAL from only that checkpoint - after loading in all the completed chunks.

codesome:
Yes, in my previous attempt I was writing all the chunks (complete and incomplete). But in this case (initial idea), where we write out the chunks in real-time when they are complete (and not like a checkpoint exactly), writing incomplete chunks would be rather expensive to mmap (but can be done only for the replay purpose). I will keep this in mind when working on the design doc, thanks!

brian-brazil:
And we'd still have the complete chunks via mmap hopefully, which would make them off-heap and save RAM.

codesome:
Yes, that's the plan

codesome:
I have jotted down some high-level ideas in this doc https://docs.google.com/document/d/184urkLQnM7rqLmGvS66I15jU2pyPk_qwd8v_qDlXczo/edit?usp=sharing

codesome:
I was looking at the replay part of this work (work going on in this branch https://github.com/codesome/prometheus/tree/mmap-chunks) and I feel the need to change the format of samples record in Head (basically add another one).

If we are to decode the samples record and perform appends only to find out the all the samples are out of order after having the m-mapped chunks already restored, then it is going to be a lot of waste of time.

To take advantage of Head chunks on disk for replay speed, I wish to add uncompressed maxt to all samples record (max time among all the samples in that record), so that we can skip the record whose maxt is less than the smallest maxt among all the series currently loaded.

More ideas are welcome.

codesome:
[...]

PR #6542, opened 2020-01-02
Bring back tombstones to Head block

codesome:
This PR reverts the changes done by https://github.com/prometheus/tsdb/pull/270 which is required for https://github.com/prometheus/prometheus/issues/6377.

One change that I noticed while fixing the tests is that even if all the samples are deleted in the tombstones, the querier would still return series but with 0 samples. Is this behaviour fine or should the querier not return any series at all?


brian-brazil:
That behaviour is okay, we'd need a bit much pre-processing otherwise. PromQL should in principle already deal with this, but add uniitests if you're unsure.

codesome:
Then the current test coverage looks fine for this part where samples are verified.

brian-brazil:
What do benchmarks look like?

It also looks like once we delete data that you can't recreate it in that time range, which makes things simpler (and is reasonable semantically).

codesome:
> What do benchmarks look like?

Query benchmarks? Haven't done any benchmark for this change yet

> It also looks like once we delete data that you can't recreate it in that time range

That was the behavior before prometheus/tsdb#270

brian-brazil:
We don't have any benchmarks covering tombstones in the head?

codesome:
I don't see any, so, nope. What do we want to benchmark for tombstones? Queries/iterators?

brian-brazil:
It's not a use case we particularly care about performance wise, as you shouldn't be deleting from the head outside of rare mistakes you need to clean up so I wouldn't worry about it.

Probably should add a unittest to cover trying to add a sample in a time period that has been tombstones though.

codesome:
@brian-brazil added the test cases.

Ping @krasi-georgiev for a review.

brian-brazil:
:+1:

tsdb/head.go
```diff
@@ -1088,39 +1068,8 @@ func (h *Head) Delete(mint, maxt int64, ms ...*labels.Matcher) error {
...
-
-	ms.reset()
-	for it.Next() {
-		t, v := it.At()
-		ok, _ := ms.append(t, v)
-		if !ok {
-			level.Warn(h.logger).Log("msg", "failed to add sample during delete")
-		}
+	for _, s := range stones {
+		h.tombstones.AddInterval(s.Ref, s.Intervals[0])
```
bwplotka: Can we do that directly in `for p.Next() {` loop above?

codesome: There are cases where things would error out above, in which case we don't want to add any stones for the delete call. So adding here assures that there was no error during the delete process above.

tsdb/head.go
```diff
@@ -1068,13 +1054,7 @@ func (h *Head) Delete(mint, maxt int64, ms ...*labels.Matcher) error {
...
-			stones = append(stones, tombstones.Stone{Ref: p.At(), Intervals: tombstones.Intervals{{Mint: t0, Maxt: t1}}})
-		}
-		if err := h.chunkRewrite(p.At(), tombstones.Intervals{{Mint: t0, Maxt: t1}}); err != nil {
-			return errors.Wrap(err, "delete samples")
-		}
-		dirty = true
+		stones = append(stones, tombstones.Stone{Ref: p.At(), Intervals: tombstones.Intervals{{Mint: t0, Maxt: t1}}})
 	}
 	if p.Err() != nil {
 		return p.Err()
```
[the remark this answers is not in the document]
codesome: Thanks, removed it.

tsdb/head.go
bwplotka: :+1:  Thanks for explanation!

Issue #6527, opened 2019-12-27
[UX] Alert filtering is hard to understand


kustodian:
## Proposal
There are two big UX problems with filtering alerts. The first one is that the filters on top don't look like they are toggles and it's hard to understand what happens when you click on one of the filters. They look more like tabs than checkboxes, so when I click on e.g. Firing, is it going to display only firing alerts, or exclude them. See the image below. Basically all the colleagues that used this filtering told me that every time they use it they have to discover again how it actually works, which is bad. It would be far better that these are checkboxes.
![image](https://user-images.githubusercontent.com/1949229/71514120-7c621280-289d-11ea-9c83-aacbf5b05cfc.png)

The second issue is that filters don't hide groups of alerts (files) that are not selected by the filter. Here is how the list of alerts looks like when I just select Firing. As you can see you have to scroll down to get to the firing alert, it's not visible because of the group names.
![image](https://user-images.githubusercontent.com/1949229/71513684-ac101b00-289b-11ea-97e9-be95fffd6b0b.png)

* Prometheus version:

	prometheus, version 2.15.1

juliusv:
Totally, I always feel exactly the same when looking at the Alerts page, especially about the filters at the top. If someone can propose a better mockup (before implementing), that would be awesome.

boyskila:
Yeah, good point! Using checkboxes is more intuitive and visually is looking way better. So something like this could be used instead.
![Screenshot 2020-01-01 23:20:15](https://user-images.githubusercontent.com/10514821/71646223-539eaa80-2ced-11ea-9c42-6bde95b93c54.png)

juliusv:
I like it! Wondering if it makes sense to group the three new buttons even more, or prefix them with something (e.g. "Show alert types:"), but I'd say do a PR and we can play around :)

juliusv:
Reopening this issue, because the linked fix only addressed the first issue (the checkboxes).

The question of also hiding any groups with no items remains an open one.

PR #6543, opened 2020-01-02
ReactUI: replace togglers tabs with checkboxes

boyskila:
Fixes: https://github.com/prometheus/prometheus/issues/6527
@juliusv

### DIFF
--- a/web/ui/react-app/src/pages/alerts/AlertContents.tsx
+++ b/web/ui/react-app/src/pages/alerts/AlertContents.tsx
@@ -67,15 +67,10 @@ const AlertsContent: FC<AlertsProps> = ({ groups = [], statsCount }) => {
 
   return (
     <>
-      <div className="mb-2 d-flex">
+      <div className="d-flex togglers-wrapper">
         {stateColorTuples.map(([state, color]) => {
           return (
-            <Checkbox
-              defaultChecked
-              wrapperStyles={{ margin: '0 15px 0 0' }}
-              id={`${state}-toggler`}
-              onClick={toggle(state)}
-            >
+            <Checkbox wrapperStyles={{ marginRight: 10 }} defaultChecked id={`${state}-toggler`} onClick={toggle(state)}>
               <Badge color={color} className="text-capitalize">
                 {state} ({statsCount[state]})
               </Badge>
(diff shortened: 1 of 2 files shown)


roidelapluie:
mmmh https://deploy-preview-6543--prometheus-react.netlify.com/new/alerts gives a blank page

boyskila:
Yes, known issue. As far as I'm aware, the version of the API used from netlify is old and alerts from the response are undefined which causes the page to goes down

juliusv:
Yeah, unfortunately. In general, @SuperQ is looking into setting up a Prometheus demo instance that will be kept up to date. Once that exists, we can point the React UI preview at it.

web/ui/react-app/src/pages/alerts/AlertContents.tsx
```diff
@@ -80,6 +71,20 @@ const AlertsContent: FC<AlertsProps> = ({ groups = [], statsCount }) => {
...
         </Checkbox>
+        {alertKinds.map((alertKind, i) => {
+          return (
+            <Checkbox
+              defaultChecked
+              id={`${alertKind}-toggler`}
+              wrapperStyles={{ margin: '0 0 0 15px' }}
+              onClick={toggle(alertKind)}
+            >
+              <Badge color={['success', 'warning', 'danger'][i]} className="text-capitalize">
```
juliusv: This seems sliiiightly unclean because it couples the assumption of a length of 3 to the `alertKinds` list, but only implicitly. How about a map from state to color and iterating over that instead?

web/ui/react-app/src/pages/alerts/AlertContents.tsx
```diff
@@ -80,6 +71,20 @@ const AlertsContent: FC<AlertsProps> = ({ groups = [], statsCount }) => {
         >
           Show annotations
         </Checkbox>
+        {alertKinds.map((alertKind, i) => {
+          return (
+            <Checkbox
+              defaultChecked
+              id={`${alertKind}-toggler`}
+              wrapperStyles={{ margin: '0 0 0 15px' }}
```
juliusv: Currently the state texts don't line up nicely vertically with their checkboxes (they seem multiple pixels too low) - could that be improved?

Screenshot from 2020-01-02 23-32-55

web/ui/react-app/src/pages/alerts/AlertContents.tsx
boyskila: Yep, will refactor this with a map, thanks!

web/ui/react-app/src/pages/alerts/AlertContents.tsx
boyskila: Yes, I also notice that but I decided not to touch it because this is just an idea, not the final solution. Here some more ideas
Screenshot 2020-01-03 00:48:41
Screenshot 2020-01-03 00:55:50
Screenshot 2020-01-03 00:57:27

juliusv:
I like the last one!

juliusv:
Looks pretty good, just the badges are still like 1px (or something like that) too low, in comparison to their checkboxes?

juliusv:
:+1: Merging... would be nice to 100% align things in all zoom situations, but it's already better for now.

PR #6547, opened 2020-01-03
Fix TSDB issues on Windows, add CI

brian-brazil:


roidelapluie:
@brian-brazil does this need to be added to github repo config as well?

brian-brazil:
This is me playing around, we don't have the right sort of CircleCI account for it though.

roidelapluie:
you need to add it to jobs: as well

.circleci/config.yml
```diff
@@ -49,6 +50,13 @@ jobs:
         key: v1-npm-deps-{{ checksum "web/ui/react-app/yarn.lock" }}
         paths:
         - web/ui/react-app/node_modules
+  test_windows:
+    executor: win/default
+    working_directory: /go/src/github.com/prometheus/prometheus
+    steps:
+    - checkout
+    - run:
+      - command: go test ./...
```
roidelapluie: shell: bash.exe

.circleci/config.yml
```diff
@@ -49,6 +50,12 @@ jobs:
         key: v1-npm-deps-{{ checksum "web/ui/react-app/yarn.lock" }}
         paths:
         - web/ui/react-app/node_modules
+  test_windows:
+    executor: win/default
+    working_directory: /go/src/github.com/prometheus/prometheus
+    steps:
+    - checkout
+    - run: go test ./...
```
roidelapluie:
```suggestion
    - run:
        command: go test ./...
        shell: bash.exe
```

roidelapluie: mmh it works with powershell apparently.

tsdb/index/index.go
```diff
@@ -720,6 +720,7 @@ func (w *Writer) writePostingsOffsetTable() error {
 	}
 
 	// Cleanup temporary file.
+	f.Close() // Needed for Windows.
```
bwplotka: Just curious.. why? (:

bwplotka: :+1: 

f is mmap of fP0 so it's important to close this handler. I would say it's sane for non windows as well (:

tsdb/index/index.go
```diff
@@ -962,9 +963,9 @@ type labelIndexHashEntry struct {
 }
 
 func (w *Writer) Close() error {
-	if err := w.ensureStage(idxStageDone); err != nil {
-		return err
-	}
+	// Even if this failes, we need to close all the files.
```
roidelapluie:
```suggestion
	// Even if this fails, we need to close all the files.
```

brian-brazil: Fixed

brian-brazil:
@simonpasquier @SuperQ This is touching CI, so one of ye should have a look at that bit. I did just enough to get it working.

brian-brazil:
This is ready for review.

tsdb/index/index.go
```diff
@@ -720,6 +724,8 @@ func (w *Writer) writePostingsOffsetTable() error {
 	}
 
 	// Cleanup temporary file.
+	f.Close() // Needed for Windows.
```
bwplotka: Do we care about errors on close? (I know we used to not care) :thinking: should we log at least? It's a pretty unlikely error and not related to the bug we fix here, so not really necessary for this PR.

FYI we have this helper on Thanos: https://github.com/thanos-io/thanos/blob/865d5ec710b5651ead08a7dc73864a6f5dbeeaa9/pkg/runutil/runutil.go#L119

brian-brazil: We don't generally care, but I've returned it on the happy path here just in case.

bwplotka: Thanks!

PR #6548, opened 2020-01-03
PromQL: Make parser completely generated

slrtbtfs:
This PR replaces the existing PromQL parser by a completely generated parser, as described in #6256.

~While this passes all unit tests now, there are still a couple of issues with it, so I've marked it as WIP.~


slrtbtfs:
This should now be ready for review.

* Since the file promql/generated_parser.y almost completely changed, it is probably easier to view the new version of the file instead of diffs.
* The unit tests underwent some changes regarding expected error messages. Most of the changed error messages have gotten more helpful. There will be follow ups which further improve error handling.

cc @brian-brazil @brancz @simonpasquier

promql/parse_test.go
```diff
@@ -168,38 +168,40 @@ var testExpr = []struct {
 		input:  "# just a comment\n\n",
 		fail:   true,
 		errMsg: "no expression found in input",
-	}, {
-		input:  "1+",
-		fail:   true,
-		errMsg: "no valid expression found",
+		/*
```
brian-brazil: I think you meant to uncomment this.

promql/generated_parser.y
```diff
@@ -35,120 +36,351 @@ import (
     series    []sequenceValue
     uint      uint64
     float     float64
+    string    string
+    duration  time.Duration
 }
 
 
-%token <item> ERROR 
-%token <item> EOF
```
brian-brazil: I'd leave them like this, it makes history of source control easier.

promql/generated_parser.y
```diff
@@ -35,120 +36,351 @@ import (
...
 
+// Operators are listed with increasing precedence.
+%left LOR
+%left LAND LUNLESS
+%left EQL GTE GTR LSS LTE NEQ
+%left ADD SUB
+%left MUL DIV MOD
+%right POW
+
+// Offset modifiers do not have associativity.
```
brian-brazil: Is there an expression where this even comes up?

promql/generated_parser.y
```diff
@@ -35,120 +36,351 @@ import (
...
+                                Without:  true,
+                        }
+                        }
+                ;
+
+/*
+ * Binary expressions.
+ */
+
+// Operator precedence only works, if each of those is listed separately.
```
brian-brazil: No comma

promql/generated_parser.y
```diff
@@ -35,120 +36,351 @@ import (
...
+
+/*
+ * Subquery and range selectors.
+ */
+
+matrix_selector : expr LEFT_BRACKET duration RIGHT_BRACKET
+                        {
+                        vs, ok := $1.(*VectorSelector)
+                        if !ok{
+                                yylex.(*parser).errorf("matrix selectors only allowed for vector selectors")
```
brian-brazil: Avoid the use of the word matrix in anything user facing, it's archaic in Prometheus. This is a range on a selector.

promql/generated_parser.y
```diff
@@ -35,120 +36,351 @@ import (
...
+                        }
+                        if vs.Offset != 0{
+                                yylex.(*parser).errorf("no offset modifiers allowed before range selector")
```
brian-brazil: Why not?

promql/generated_parser.y
```diff
@@ -35,120 +36,351 @@ import (
...
+                ;
 
-label_matchers  : 
-                LEFT_BRACE label_match_list RIGHT_BRACE
+/*
+ * Unary expressions.
+ */
+
+unary_expr      :
+                /* gives the rule the same prec as POW, not a perfect solution */
```
brian-brazil: Does this matter? I'd expect unary to have the highest precedence.

slrtbtfs: The precedence has to be explicitly specified to make the grammar unambiguous and avoid so called shift/reduce conflicts.

In usual mathematical notation, unary operators actually have the same precedence as multiplication, so `-a^2` evaluates to `-(a^2)`.

However, in PromQL, unary operators have indeed the highest precedence, which was not properly implemented here.

I've rewritten that grammar rule to correct this and added an extra unit test for it.

promql/generated_parser.y
slrtbtfs: Because the existing PromQL parser does not support this. Example:

```
some_metric offset 5s [3m]
```
Will lead to
``` 
parse error at char 26: unexpected "]" in subquery selector, expected ":"
```
with the old parser and to
```
1:26: parse error: no offset modifiers allowed before range selector
```
with the new parser.

promql/generated_parser.y
slrtbtfs: Updated this.

promql/generated_parser.y
slrtbtfs: The point this statement is to instruct yacc to evaluate expressions like `a + b offset 42s` as `a + (b offset 42s)` instead of `(a + b) offset 42s`. The latter would fail at a later stage anyway, but yacc does not know this.

promql/parse_test.go
slrtbtfs: Indeed. Uncommented now.

promql/generated_parser.y
slrtbtfs: Resolved.

promql/generated_parser.y
slrtbtfs: How about this form:
```
%token <item>
TOKEN1
TOKEN2
...
```
This is what I've done now.

promql/generated_parser.y
brian-brazil: The whole of `{}[5m]` is a range selector, not just the bit in square brackets. So being in the middle of it is more the issue.

promql/parse_test.go
```diff
@@ -779,6 +786,10 @@ var testExpr = []struct {
 		input:  "foo == on(bar) 10",
 		fail:   true,
 		errMsg: "vector matching only allowed between instant vectors",
+	}, {
+		input:  "foo + group_left(baz) bar",
```
brian-brazil: Why is this invalid? Looks fine to me.

promql/generated_parser.y
```diff
@@ -321,7 +321,7 @@ matrix_selector : expr LEFT_BRACKET duration RIGHT_BRACKET
                         {
                         vs, ok := $1.(*VectorSelector)
                         if !ok{
-                                yylex.(*parser).errorf("matrix selectors only allowed for vector selectors")
+                                yylex.(*parser).errorf("range selectors only allowed for vector selectors")
```
brian-brazil: ranges only allowed

promql/generated_parser.y
brian-brazil: Hmm, so why didn't unittests catch that? We don't document the precedence (or even existence) of the unary operators.

slrtbtfs: There are a bunch of unit tests for operator precedence, which cover most cases.

However the special case of combining an unary operator with `^` slipped trough.

I've tested it with old Prometheus versions and the unary operators do have a higher precedence than `^` there.

promql/parse_test.go
slrtbtfs: Because `group_left|group_right` are only allowed together with `on|ignoring` by the old parser. This test makes sure that the new parser behaves the same.

http://demo.robustperception.io:9090/graph?g0.range_input=1h&g0.expr=foo%20%2B%20group_left(baz)%20bar&g0.tab=1

promql/generated_parser.y
slrtbtfs: Is the terminology "subquery selector" ok, then?

slrtbtfs: Changed the error message accordingly.

promql/generated_parser.y
brian-brazil: That sounds like an unintentional mistake, we should follow what other languages typically do here.

promql/parse_test.go
brian-brazil: Ah yes, it'd be valid but useless.

promql/generated_parser.y
brian-brazil: There's no such thing as a subquery selector, but I don't see that wording anywhere in this PR.

promql/generated_parser.y
brian-brazil: That's distributivity, not associativity.

promql/generated_parser.y
slrtbtfs: For this PR I think we should rather not intentionally change PromQL behaviour since it's already complex enough.

I'd prefer doing this as a follow up.

slrtbtfs: I've created an issue since I feel this should be discussed before changing the current behaviour.

https://github.com/prometheus/prometheus/issues/6575

For this PR, I'd like to leave it as is.

promql/generated_parser.y
slrtbtfs: The `%nonassoc` here contains more than only associativity information:
 
The important part is that it is listed after the binary operators, which tells yacc that `offset` modifiers have a higher precedence than binary operators.

Every precedence specification in yacc must also specify associativity. Since the concept of associativity does not apply here`%nonassoc` is used.

brian-brazil:
Thanks!

PR #6556, opened 2020-01-04
Don't use hash combination

### LINKED ISSUE #5724: hash collision in engine leads to incorrect "multiple matches for labels" error
## Bug Report
**What did you do?**

Run this test case with `promtool test rules`.  This test case is isolated from a series that failed in our production Prometheus instance.

```yaml
evaluation_interval: 1m
tests:
  - interval: 1s
    input_series:
      - series: 'node_namespace_pod:kube_pod_info:{namespace="observability",node="gke-search-infra-custom-96-253440-fli-d135b119-jx00",pod="node-exporter-l454v"}'
        values: '1'
      - series: 'node_cpu_seconds_total{cpu="10",endpoint="https",instance="10.253.57.87:9100",job="node-exporter",mode="idle",namespace="observability",pod="node-exporter-l454v",service="node-exporter"}'
        values: '449'
      - series: 'node_cpu_seconds_total{cpu="35",endpoint="https",instance="10.253.57.87:9100",job="node-exporter",mode="idle",namespace="observability",pod="node-exporter-l454v",service="node-exporter"}'
        values: '449'
      - series: 'node_cpu_seconds_total{cpu="89",endpoint="https",instance="10.253.57.87:9100",job="node-exporter",mode="idle",namespace="observability",pod="node-exporter-l454v",service="node-exporter"}'
        values: '449'
    promql_expr_test:
      - expr: count by(namespace, pod, cpu) (node_cpu_seconds_total{cpu=~".*",job="node-exporter",mode="idle",namespace="observability",pod="node-exporter-l454v"}) * on(namespace, pod) group_left(node) node_namespace_pod:kube_pod_info:{namespace="observability",pod="node-exporter-l454v"}
        eval_time: 4s
        exp_samples:
          - labels: '{cpu="10",namespace="observability",node="gke-search-infra-custom-96-253440-fli-d135b119-jx00",pod="node-exporter-l454v"}'
            value: 1
          - labels: '{cpu="35",namespace="observability",node="gke-search-infra-custom-96-253440-fli-d135b119-jx00",pod="node-exporter-l454v"}'
            value: 1
          - labels: '{cpu="89",namespace="observability",node="gke-search-infra-custom-96-253440-fli-d135b119-jx00",pod="node-exporter-l454v"}'
            value: 1
```

A success message, with the three output series correctly calculated.

A test failure:
```
  FAILED:
    expr:'count by(namespace, pod, cpu) (node_cpu_seconds_total{cpu=~".*",job="node-exporter",mode="idle",namespace="observability",pod="node-exporter-l454v"}) * on(namespace, pod) group_left(node) node_namespace_pod:kube_pod_info:{namespace="observability",pod="node-exporter-l454v"}', time:4s, err:multiple matches for labels: grouping labels must ensure unique matches
```

**Investigation**

This failure is due to a hash collision in the VectorBinop code in `engine.go`, discovered by @cristi- .  Specifically, combining the hashes of the LHS and RHS at [L1550](https://github.com/prometheus/prometheus/blob/851131b0740be7291b98f295567a97f32fffc655/promql/engine.go#L1550) results in a collision between two of the series in the test file, `cpu="35"` and `cpu="89"`.  Because the hashes collide there, the result metric is incorrectly reused for two series that are not identical, leading to VectorBinop incorrectly reporting non-unique matches.

You can see the hash-mixing function `(lh ^ rh) + lh` collide in isolation with this code ([playground](https://play.golang.org/p/YKRBZru5XHA)):

```go
package main

import (
	"fmt"
)

func h(l, r uint64) uint64 {
	return (l ^ r) + l
}

func main() {
	fmt.Println(h(4717575973110797488, 6876917458372880354))
	fmt.Println(h(9462489420336518194, 6876917458372880354))
}
```

This would be fixed with a different function for mixing hashes that better preserves entropy, or not combining hashes at all and using the calculated result hash throughout.

**Environment**

* System information: Linux 5.1.0-1.el7.elrepo.x86_64 x86_64
* Prometheus version: reproduced on v2.8.1 and master (851131b0740be7291b98f295567a97f32fffc655)
* Alertmanager version: n/a
* Prometheus configuration file: n/a
* Alertmanager configuration file: n/a
* Logs: n/a

roidelapluie:
Fixes #5724

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -759,7 +759,7 @@ type EvalNodeHelper struct {
 	// For binary vector matching.
 	rightSigs    map[uint64]Sample
 	matchedSigs  map[uint64]map[uint64]struct{}
-	resultMetric map[[2]uint64]labels.Labels
+	resultMetric map[uint64]labels.Labels
 }
 
 // dropMetricName is a cached version of dropMetricName.
@@ -1614,11 +1614,14 @@ func signatureFunc(on bool, names ...string) func(labels.Labels) uint64 {
 // binary operation and the matching options.
 func resultMetric(lhs, rhs labels.Labels, op ItemType, matching *VectorMatching, enh *EvalNodeHelper) labels.Labels {
 	if enh.resultMetric == nil {
-		enh.resultMetric = make(map[[2]uint64]labels.Labels, len(enh.out))
+		enh.resultMetric = make(map[uint64]labels.Labels, len(enh.out))
 	}
 	// op and matching are always the same for a given node, so
 	// there's no need to include them in the hash key.
-	h := [2]uint64{lhs.Hash(), rhs.Hash()}
+	// If the lhs and rhs are the same then the xor would be 0,
+	// so add in one side to protect against that.
+	lh := lhs.Hash()
+	h := (lh ^ rhs.Hash()) + lh
 	if ret, ok := enh.resultMetric[h]; ok {
 		return ret
 	}
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -759,7 +759,7 @@ type EvalNodeHelper struct {
 	// For binary vector matching.
 	rightSigs    map[uint64]Sample
 	matchedSigs  map[uint64]map[uint64]struct{}
-	resultMetric map[uint64]labels.Labels
+	resultMetric map[[2]uint64]labels.Labels
 }
 
 // dropMetricName is a cached version of dropMetricName.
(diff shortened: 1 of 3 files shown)


roidelapluie:
```
BenchmarkSignature-8          584           617           +5.65%
BenchmarkSignatureReuse-8     128           154           +20.31%

BenchmarkSignature-8          7              7              +0.00%
BenchmarkSignatureReuse-8     1              1              +0.00%

BenchmarkSignature-8          672           736           +9.52%
BenchmarkSignatureReuse-8     32            32            +0.00%
```

not good

PR #6558, opened 2020-01-05
Add prometheus version into block meta

roidelapluie:
Should we introduce regressions in the blocks at some point in the
future, this new meta option will easily point us the blocks that need
repair/adaptation.

### DIFF
--- a/tsdb/block.go
+++ b/tsdb/block.go
@@ -26,6 +26,7 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/oklog/ulid"
 	"github.com/pkg/errors"
+	"github.com/prometheus/common/version"
 	"github.com/prometheus/prometheus/pkg/labels"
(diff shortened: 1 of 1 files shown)


brian-brazil:
This depends on how Prometheus is compiled, so may be missing. Actual block versioning would be better (and we already have index versioning).

roidelapluie:
Yes but how to make sure that we increase it when needed?

roidelapluie:
Could we add to the release.md file that in addition to change the RELEASE file, you need to increase the block version? And therefore it would be updated on each release?

bwplotka:
I think it should be index version bumb not a prom version in meta if any. However, we have index version 2 right now. I would need to check but maybe version 2 was done after sorting was added? If not, maybe we should bumb it now to 3 and bring back old reader for blocks with index version pre sorting? 

..if we want to be backward compatible with block reading for pre 2.1 blocks. (:

roidelapluie:
ack; can be done outside of this PR

PR #6572, opened 2020-01-07
Several little slice optimization

yeya24:

A small improvement for slice operations.

### DIFF
--- a/tsdb/index/index.go
+++ b/tsdb/index/index.go
@@ -1160,6 +1160,9 @@ func newReader(b ByteSlice, c io.Closer) (*Reader, error) {
 		}
 		// Trim any extra space in the slices.
 		for k, v := range r.postings {
+			if len(v) == cap(v) {
+				continue
+			}
 			l := make([]postingOffset, len(v))
 			copy(l, v)
 			r.postings[k] = l
--- a/tsdb/tombstones/tombstones.go
+++ b/tsdb/tombstones/tombstones.go
@@ -317,12 +317,10 @@ func (itvs Intervals) Add(n Interval) Intervals {
 		}
 
 		if n.Mint < r.Mint {
-			newRange := make(Intervals, i, len(itvs[:i])+1)
-			copy(newRange, itvs[:i])
-			newRange = append(newRange, n)
-			newRange = append(newRange, itvs[i:]...)
-
-			return newRange
+			itvs = append(itvs, Interval{})
+			copy(itvs[i+1:], itvs[i:])
+			itvs[i] = n
+			return itvs
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


tsdb/index/index.go
```diff
@@ -1160,6 +1160,9 @@ func newReader(b ByteSlice, c io.Closer) (*Reader, error) {
 		}
 		// Trim any extra space in the slices.
 		for k, v := range r.postings {
+			if len(v) == cap(v) {
```
brian-brazil: What are the chances that the number of postings is exactly a power of 2? This seems like over-optimisation.

tsdb/tsdbutil/buffer.go
```diff
@@ -215,8 +215,7 @@ func (r *sampleRing) last() (int64, float64, bool) {
 	if r.l == 0 {
 		return 0, 0, false
 	}
-	s := r.buf[r.i]
-	return s.t, s.v, true
+	return r.buf[r.i].t, r.buf[r.i].v, true
```
brian-brazil: The previous version was easier to read, this is a bit line noisy

tsdb/tombstones/tombstones.go
```diff
@@ -317,12 +317,10 @@ func (itvs Intervals) Add(n Interval) Intervals {
 		}
 
 		if n.Mint < r.Mint {
-			newRange := make(Intervals, i, len(itvs[:i])+1)
-			copy(newRange, itvs[:i])
-			newRange = append(newRange, n)
-			newRange = append(newRange, itvs[i:]...)
-
-			return newRange
+			itvs = append(itvs, Interval{})
```
brian-brazil: This is changing the backing store of the itvs slice, rather than creating a new one. Is this safe?

tsdb/index/index.go
yeya24: Although the probability is not that high, I still think this is a valid change to reduce some useless slice copy.

tsdb/tsdbutil/buffer.go
yeya24: OK. I will change it back.

tsdb/tombstones/tombstones.go
yeya24: Is it the same with `itvs = append(itvs, n)`?  I am not sure about this.

tsdb/index/index.go
brian-brazil: It might on very rare occasion save 1ms on a process that takes minutes, it's not worth it.

slrtbtfs:
Are there benchmarks for these changes?

tsdb/tombstones/tombstones.go
brian-brazil: True, we're altering it already.

brian-brazil: This code is not equivalent.

yeya24: What do you mean for not equivalent? Does this change the logic? Or you mean it is unsafe?

brian-brazil: I'm not sure it's doing the same thing.

PR #6574, opened 2020-01-08
WIP: MVP exemplars storage via tsdb memseries

cstyan:
Mostly an experiment, but would be interested in feedback of others who have worked on TSDB code. 

This PR unfortunately touches a lot of code/interfaces we probably don't want to change, but achieves the goals of not having to come up with an additional storage structure or new index for exemplar data by reusing `memSeries`. Goutham also has an idea for getting the same benefits without changing existing TSDB interfaces that I'm also going to try out.

cc @gouthamve 

### DIFF
--- a/tsdb/block.go
+++ b/tsdb/block.go
@@ -26,6 +26,7 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/oklog/ulid"
 	"github.com/pkg/errors"
+	"github.com/prometheus/prometheus/pkg/exemplar"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/tsdb/chunkenc"
 	"github.com/prometheus/prometheus/tsdb/chunks"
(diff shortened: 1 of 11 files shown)


brian-brazil:
From a quick look the general approach is about right, though that Appender change isn't. Let's get the API figured out before making big changes like this.

brancz:
Totally an implementation detail but: We've also thought about having the possibility for metadata to be persisted (I realize this PR is only about memseries, but it's part of the puzzle), I would suggest that we make exemplars have an abstract layer that is essentially just a byte slice storage.

API wise I would have expected this to be a separate querier instead of being part of the same "float time series querier".

brian-brazil:
Metric metadata is more complicated, as it is tied to a metric rather than a series. The TSDB only knows about series.

beorn7:
> Metric metadata is more complicated, as it is tied to a metric rather than a series.

I believe that we have to tie metadata to series eventually, as the "unique metadata per metric" assumptions breaks both over time as between different targets.

brian-brazil:
> I believe that we have to tie metadata to series eventually, as the "unique metadata per metric" assumptions breaks both over time as between different targets.

I'm not saying otherwise, however using an opaque per-series bye slice for exemplars doesn't help with that.

brancz:
And I’m saying exemplars should be arbitrary byte slice samples identified by their set of labels which are the combination of the target+series+exemplar labels instead of tied to series.

cstyan:
-

PR #6579, opened 2020-01-08
PromQL: Fix unary operator precedence

### LINKED ISSUE #6575: PromQL: Unusual unary operator precedence
Consider the expression:
```
-a^b
```
For written mathematics the convention is to interpret this as follows:
```
-(a^b)
```
However PromQL will interpret it as
```
(-a)^b
```

__To be discussed:__ Should PromQL keep the existing behaviour that might be not expected by users or change it to follow mathematical conventions, at the risk of breaking existing queries?

slrtbtfs:
Fixes #6575

### DIFF
--- a/promql/testdata/operators.test
+++ b/promql/testdata/operators.test
@@ -35,6 +35,16 @@ eval instant at 50m - - - SUM(http_requests) BY (job)
 eval instant at 50m - - - 1
   -1
 
+eval instant at 50m -2^---1*3
+  -0.75
+
+eval instant at 50m 2/-2^---1*3+2
+  5
+
+eval instant at 50m -10^3 * - SUM(http_requests) BY (job) ^ -1
+	{job="api-server"} 1
+	{job="app-server"} 0.38461538461538464
+
 eval instant at 50m 1000 / SUM(http_requests) BY (job)
 	{job="api-server"} 1
 	{job="app-server"} 0.38461538461538464
--- a/promql/testdata/operators.test
+++ b/promql/testdata/operators.test
@@ -36,10 +36,10 @@ eval instant at 50m - - - 1
   -1
 
 eval instant at 50m -2^---1*3
-  -0.75
+  -1.5
 
 eval instant at 50m 2/-2^---1*3+2
-  5
+  -10 
 
 eval instant at 50m -10^3 * - SUM(http_requests) BY (job) ^ -1
 	{job="api-server"} 1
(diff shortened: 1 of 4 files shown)


brian-brazil:
Can you add some PromQL tests too?

slrtbtfs:
> Can you add some PromQL tests too?

In the engine?

I've already added a couple in for the parser.

brian-brazil:
In the testdata. Weirdness here is more likely to be spotted at the PromQL level.

slrtbtfs:
Ok, I've added some tests there.

promql/testdata/operators.test
```diff
@@ -35,6 +35,16 @@ eval instant at 50m - - - SUM(http_requests) BY (job)
 eval instant at 50m - - - 1
   -1
 
+eval instant at 50m -2^---1*3
+  -0.75
+
+eval instant at 50m 2/-2^---1*3+2
+  5
+
+eval instant at 50m -10^3 * - SUM(http_requests) BY (job) ^ -1
```
brian-brazil: This doesn't seem right, `-x ^ -1` should be same as `-1 / x `

promql/testdata/operators.test
```diff
@@ -35,6 +35,16 @@ eval instant at 50m - - - SUM(http_requests) BY (job)
 eval instant at 50m - - - 1
   -1
 
+eval instant at 50m -2^---1*3
+  -0.75
```
brian-brazil: Shouldn't this be -1.5?

brian-brazil: Shouldn't thus be -1.5?

promql/testdata/operators.test
slrtbtfs: It is.

You can get to this expression by transforming the below expression:

```
    1000 / SUM(http_requests) BY (job)
=  - 1000 / - SUM(http_requests) BY (job)            
=  - 10^3 / - SUM(http_requests) BY (job)
=  - 10^3 * -1 / SUM(http_requests) BY (job)        | -1 / x = -x ^ -1
=  - 10^3 * - SUM(http_requests) BY (job) ^ -1
```

promql/testdata/operators.test
slrtbtfs: Yes, it should.

slrtbtfs: This proves that math students are bad at calculating. q.e.d.

brian-brazil:
Thanks!

PR #6580, opened 2020-01-08
Cleanup: remove function that does not do anything

slrtbtfs:
This removes a function that became a no-op during the parser rewrite.

### DIFF
--- a/promql/lex.go
+++ b/promql/lex.go
@@ -343,13 +343,6 @@ func (l *Lexer) run() {
 	}
 }
 
-// Release resources used by lexer.
-func (l *Lexer) close() {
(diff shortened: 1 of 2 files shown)


promql/lex.go
```diff
@@ -343,13 +343,6 @@ func (l *Lexer) run() {
 	}
 }
 
-// Release resources used by lexer.
-func (l *Lexer) close() {
-	for range l.Items {
```
slrtbtfs: This used to have an effect, when l.Items was a channel. Now it is a slice, so the loop doesn't do anything.

brian-brazil:
Thanks!

PR #6581, opened 2020-01-08
Cleanup: Remove unused variable

slrtbtfs:

### DIFF
--- a/promql/parse.go
+++ b/promql/parse.go
@@ -29,9 +29,8 @@ import (
 )
 
 type parser struct {
-	lex     *Lexer
-	token   Item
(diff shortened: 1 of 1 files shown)


promql/parse.go
```diff
@@ -29,9 +29,8 @@ import (
 )
 
 type parser struct {
-	lex     *Lexer
-	token   Item
-	peeking bool
```
slrtbtfs: Due to the parser rewrite `peeking` is always `false` now. So it can be as well removed.

brian-brazil:
￼Thanks!

PR #6583, opened 2020-01-08
Cleanup: Remove parser switching logic

slrtbtfs:
During the PromQL parser rewrite there was some logic put in place that allowed switching between the non generated and the generated parser. Since the parser is now fully generated this is not needed anymore.

### DIFF
--- a/promql/parse.go
+++ b/promql/parse.go
@@ -36,8 +36,6 @@ type parser struct {
 	inject    Item
 	injecting bool
 
-	switchSymbols []ItemType
-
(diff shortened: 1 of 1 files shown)


roidelapluie:
Perhaps you should group your cleanups in a single PR?

thanks

slrtbtfs:
I assumed it's better to have multiple smaller PRs. But sure, I can do that.

roidelapluie:
I would do multiple commits in one PR

brian-brazil:
Thanks!

PR #6585, opened 2020-01-08
tsdb: Fixed Symbol Lookup edge case; Added tests.

bwplotka:
Without fix lookup for last reference + 1 returns extra bytes after symbols.

### DIFF
--- a/tsdb/index/index_test.go
+++ b/tsdb/index/index_test.go
@@ -16,6 +16,7 @@ package index
 import (
 	"context"
 	"fmt"
+	"hash/crc32"
 	"io/ioutil"
 	"math/rand"
 	"os"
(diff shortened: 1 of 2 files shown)


tsdb/index/index_test.go
```diff
@@ -505,3 +506,48 @@ func TestNewFileReaderErrorNoOpenFiles(t *testing.T) {
...
 	dir.Close()
 }
+
+func TestSymbols(t *testing.T) {
+	buf := encoding.Encbuf{}
+
+	buf.PutUvarintStr("something")
+
+	symbolsStart := buf.Len()
+	buf.PutBE32int(204)
```
brian-brazil: This could all do with comments so you know what field is what.

tsdb/index/index_test.go
```diff
@@ -505,3 +506,48 @@ func TestNewFileReaderErrorNoOpenFiles(t *testing.T) {
...
+	buf.PutBE32int(100)
+	for i := 0; i < 100; i++ {
+		buf.PutUvarintStr(string(i))
```
brian-brazil: These symbols will not be in order, you'd need fixed size strings.

brian-brazil: Or add your test to TestPostingsMany, which already has all the required infrastructure.

bwplotka: I think a separate unit test for separate symbols structure would be nice (:

Additionally, those symbols are in perfect order, when you print `string(i)`:   `       	 
                     ! " # $ % & ' ( ) * + , - . / 0 1 2 3 4 5 6 7 8 9 : ; < = > ? @ A B C D E F G H I J K L M N O P Q R S T U V W X Y Z [ \ ] ^ _ ` a b c `

They would be not in order if I would do `fmt.Sprint(i)` (:

tsdb/index/index_test.go
bwplotka: :+1:

tsdb/index/index_test.go
brian-brazil: That needs a comment then, I'd intuitively expect it to be the decimal representation of the int

bwplotka: done

tsdb/index/index_test.go
```diff
@@ -505,3 +506,50 @@ func TestNewFileReaderErrorNoOpenFiles(t *testing.T) {
...
+
+	symbolsStart := buf.Len()
+	buf.PutBE32int(204) // Length of symbols table.
+	buf.PutBE32int(100) // Number of symbols.
+	for i := 0; i < 100; i++ {
+		// i represents index in unicode characters table.
+		buf.PutUvarintStr(string(i)) // Symbol.
+	}
+	hash := crc32.Checksum(buf.Get()[symbolsStart+4:], castagnoliTable)
+	buf.PutBE32(hash) // Check sum at the end.
```
brian-brazil: Checksum

PR #6590, opened 2020-01-09
PromQL: AST: Make VectorSelector Children of MatrixSelector

slrtbtfs:
This PR changes the structure of the PromQL Abstract Syntax Tree.

Expressions of the form `foo[42s]` used to be represented as a single AST node.
```
&MatrixSelector{
    Name: "foo",
    Range: 42s,
    ...
}
```

Now this would be represented as:
```
&MatrixSelector{
    &VectorSelector{
        Name: "foo",
        ...
    },
    Range: 42s,
}
```

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -625,8 +625,8 @@ func (ng *Engine) populateSeries(ctx context.Context, q storage.Queryable, s *Ev
 			if maxOffset < n.Range+subqOffset {
 				maxOffset = n.Range + subqOffset
 			}
-			if n.Offset+n.Range+subqOffset > maxOffset {
-				maxOffset = n.Offset + n.Range + subqOffset
+			if n.VectorSelector.Offset+n.Range+subqOffset > maxOffset {
+				maxOffset = n.VectorSelector.Offset + n.Range + subqOffset
 			}
 		}
 		return nil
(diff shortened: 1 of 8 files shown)


brian-brazil:
What's the motivation behind this? It doesn't seem to be making things simpler.

slrtbtfs:
I plan to add position metadata to the AST soon, in order to be able to map positions in the query string to AST nodes.

For this use case it's good to have a more detailed syntax tree.

IMHO this PR makes the AST a closer representation of PromQLs syntax structure.

Also, this approach avoids some Code duplication.

brian-brazil:
> Also, this approach avoids some Code duplication.

I think you're more moving code around, so it's all special cases rather than duplication now. It just feels a bit weird.

slrtbtfs:
The engine changes are indeed a bit weird, and I would have expected it to be nicer when I started working on this PR. It doesn't necessarily make things worse, though.

However in the parser, copying the content of an already generated VectorSelector Node into a `MatrixSelector` Node that looks almost the same felt weird, too.

Especially in the language server, dealing with two Node types that almost are the same generates a lot of needless duplication and complexity.

promql/printer.go
```diff
@@ -112,14 +112,15 @@ func (node *Call) String() string {
...
+	vecSelector := node.VectorSelector
 	offset := ""
-	if node.Offset != time.Duration(0) {
-		offset = fmt.Sprintf(" offset %s", model.Duration(node.Offset))
+	if vecSelector.Offset != time.Duration(0) {
+		offset = fmt.Sprintf(" offset %s", model.Duration(vecSelector.Offset))
 	}
+
+	// Do not print the offset twice.
+	vecSelector.Offset = 0
```
brian-brazil: We shouldn't be mutating the tree in a print function.

slrtbtfs: Fair point.

brian-brazil:
Thanks!

promql/engine.go
```diff
@@ -660,7 +665,16 @@ func (ng *Engine) populateSeries(ctx context.Context, q storage.Queryable, s *Ev
...
+			} else {
+				params.Range = durationMilliseconds(evalRange)
+				// For all matrix queries we want to ensure that we have (end-start) + range selected
+				// this way we have `range` data before the start time
+				params.Start = params.Start - durationMilliseconds(evalRange)
+				evalRange = 0
+			}
+
 			params.Func = extractFuncFromPath(path)
 			params.By, params.Grouping = extractGroupsFromPath(path)
```
slrtbtfs: That line should only be run in the case `evalRange == 0`.

PR #6591, opened 2020-01-09
PromQL: Use a sync.Pool for the generated parser structure

slrtbtfs:
The generated PromQL parser allocates a struct about 4kb in size on every run.

This puts a high load on the garbage collector.

To reduce that load, a `sync.Pool` is used to recycle these structures.

For small queries this makes parsing 2-3 times faster.

### DIFF
--- a/promql/parse.go
+++ b/promql/parse.go
@@ -19,6 +19,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


slrtbtfs:
I've updated the PR to store the `parser` and `Lexer` in that `sync.Pool`, too.

This speeds up the parser by another 10-20%.

brian-brazil:
Thanks!

PR #6592, opened 2020-01-09
Query Log: add origin of the rules

roidelapluie:
We don't set rule name and rule kind because the added value would be
quite low, given we have now the file, the group and the query.

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -287,6 +287,12 @@ func (g *Group) run(ctx context.Context) {
 		return
 	}
 
+	ctx = promql.NewOriginContext(ctx, map[string]string{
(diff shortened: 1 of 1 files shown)


rules/manager.go
```diff
@@ -287,6 +287,12 @@ func (g *Group) run(ctx context.Context) {
 		return
 	}
 
+	ctx = promql.NewOriginContext(ctx, map[string]string{
+		"groupFile":          g.File(),
+		"groupName":          g.Name(),
+		"evaluationInterval": g.Interval().String(),
```
brian-brazil: We don't need this, it's implicit from the file/group. Plus it'll be obvious from the timestamps.

PR #6593, opened 2020-01-09
Query Log: Add source IP from console queries

roidelapluie:


brian-brazil:
I think the path (but not the parameters) would be useful too.

roidelapluie:
Added path.

brian-brazil:
Tests are failing.

brian-brazil:
Thanks!

PR #6594, opened 2020-01-09
Query Logging styling

roidelapluie:
- Fix Json vs JSON in activequerylogger
- Fix SetQueryLogger always returns nil


brian-brazil:
Thanks!

PR #6601, opened 2020-01-10
ReactUI: folder re-organization

boyskila:
This PR aims to improve the React UI file organization with the following structure.
```javascript
-src
   -common
     Checkbox.tsx
      ....etc
   -hooks
     useFetch.ts
     ...etc
   -pages
      -graph
      -alerts
      -targets
      -config
      -flags
      -status
      -rules
      -tsdbStatus
      -serviceDiscovery
   -types
   -utils
```

Eventually, each page can be organized further but IMO unless it does not need it's better to remain as flatten as possible.
@juliusv
@roidelapluie in case you are interested


juliusv:
Thanks, I generally like it! A couple of questions:

- What's the conceptual difference between `common` and `utils`? E.g. why is `withStatusIndicator.tsx` in `common` instead of `utils`? If it's because it's component-like, would it make sense to rename `common` to `components` instead?

- Not sure about `Navbar.tsx` living in `common`, given that it's only used from `App.tsx` and not commonly used anywhere else?

boyskila:
I think of ```common``` more like components related folder i.e common components or shared components. About the ```utils```, I think of it more like helper functions utilities (not .tsx files)

That's why withStatusIndicator IMO should live there.

Navbar is a different case, yeah. It should live where App.tsx is.

juliusv:
@boyskila That sounds like `components` would be a bit better descriptive name for the folder then?

boyskila:
@juliusv thinking more about ```utils``` folder I'm more about the idea to combine all files there into ```func.ts``` (which btw can be renamed to ```utils.ts```).

juliusv:
Yeah, that seems fine to me too.

juliusv:
:+1: Thanks!

Issue #6576, opened 2020-01-08
React UI: Stack graph displays null value as zero


roidelapluie:
## Proposal
**Use case. Why is this important?**

A value not defined is different than 0.
Old UI behaves correctly, as does grafana.

## Bug Report
**What did you do?**

Stack a graph with null values. exemple query:

label_replace(vector(time()) % 300 < 180,"x","x","","") or label_replace(vector(time()) % 300 > 140,"y","y","","") 

![est](https://user-images.githubusercontent.com/291750/71974746-8fb49c80-3212-11ea-996f-84dbb8b109dd.png)

**What did you expect to see?**

Null points are not shown

**What did you see instead? Under which circumstances?**

Null points are shown as 0

**Environment**

* Prometheus version:

	2.15.2

This is known issue in code but not tracked in a bug yet. It is a bug in a third party library we use that forces us to set 0.
cc @juliusv

juliusv:
Thanks for opening an issue for this!

CC @boyskila who has already starting looking a bit into this on the Flot / Grafana Flot fork side.

PR #6603, opened 2020-01-11
ReactUI: adopt grafana flot fix for stacked graphs

boyskila:
Fixes: https://github.com/prometheus/prometheus/issues/6576


juliusv:
The behavior looks good to me, thanks! @roidelapluie could you also double-check at https://deploy-preview-6603--prometheus-react.netlify.com/new/graph?g0.expr=label_replace(vector(time())%20%25%20300%20%3C%20180%2C%22x%22%2C%22x%22%2C%22%22%2C%22%22)%20or%20label_replace(vector(time())%20%25%20300%20%3E%20140%2C%22y%22%2C%22y%22%2C%22%22%2C%22%22)&g0.tab=0&g0.stacked=1&g0.range_input=1h

I think it'd be good to change the PR / commit description to note that we are vendoring Grafana's Flot version, not just adopting a single fix from it (although that is the purpose, which should still be noted).

We also need to do our best regarding copyright notices, since Grafana did some changes on its Flot version after forking it, but didn't add new copyright headers :( @davkal as my go-to Grafana person, what would you advise us to do here?

roidelapluie:
+1 on everything.

davkal:
In my layman's opinion: We didnt rewrite flot, so I think the original copyright notice holds. For maintainability I suggest you add a comment linking to the vendored version in Grafana. 

Also I'm pretty sure stacking is broken if you have timeseries with missing datapoints.

juliusv:
@davkal Grafana does have a bunch of local changes to the files in https://github.com/grafana/grafana/tree/master/public/vendor/flot, so while the original notices should be preserved, it seems that additional ones should be added as well. @boyskila How about also adding a note at the top of each copied file:

> This file was copied into Prometheus from Grafana's vendored fork of Flot (living at https://github.com/grafana/grafana/tree/master/public/vendor/flot), which contains fixes for displaying null values in stacked graphs. The original Flot code was licensed under the MIT license as reproduced below. Additional changes have been contributed to the Grafana fork under an Apache 2 license, see https://github.com/grafana/grafana/blob/master/LICENSE.

I'm not a lawyer either, but Grafana people are friends of Prometheus, and this is hopefully a good-enough effort to show what came from where.

@davkal We actually tested stacking with your vendored fork, and at least as far as we could see (see Prometheus link above), it fixes the null stacking bug that is present in upstream Flot.

boyskila:
Ok, sounds good. Hope tomorrow someone's lawyer not going to knocking on my door :)

web/ui/react-app/src/vendor/flot/jquery.flot.crosshair.js
```diff
@@ -0,0 +1,191 @@
...
+
+    Free the crosshair to move again after locking it.
+*/
+
+/**
+ *
+ * THIS FILE WAS COPIED INTO PROMETHEUS FROM GRAFANA'S VENDORED FORK OF FLOT
+ * (LIVING AT HTTPS://GITHUB.COM/GRAFANA/GRAFANA/TREE/MASTER/PUBLIC/VENDOR/FLOT),
+ * WHICH CONTAINS FIXES FOR DISPLAYING NULL VALUES IN STACKED GRAPHS. THE ORIGINAL
+ * FLOT CODE WAS LICENSED UNDER THE MIT LICENSE AS REPRODUCED BELOW. ADDITIONAL
```
juliusv: I'd put this new notice header at the very top. Then the word "BELOW" also makes sense :)

Not sure why it'd all be capitalized? At least the URLs are broken through the capitalization now.

juliusv: And: actually my "as reproduced below" wasn't correct, since the original files don't contain a full reproduction of the MIT license. So "as reproduced below" -> "as stated below".

juliusv:
:+1: Thanks!

PR #6607, opened 2020-01-13
support modify alert-rule by http post/put/delete

dw130:
privode http post/put/delete method to enable alert rule instead of config map
storage config in mysql

post / put / delete  http://127.0.0.1:9090/api/v1/alertRule
{
	"Id": 0,     //mysql alert_rule id,modify or delete rule through id
	"name": "example",
	"expr": "sum (absent (example{job=\"example\"})) by (job_name)==1",
	"severity": "critical",
	"for": "1m",
	"sendInterval": "2m",
	"description": "example no data alert"
}

"Id": the primary id of the alert-rule which storage in mysql, can modify or delete a rule through it
"for": an alert will stay with pending-state util a "for"-time expire and change to firing
"sendInterval": an alert state duration, will not send util the time expire

when modify a rule alert, the origin alert state will be hold

fetch all rule:  http://127.0.0.1:9090/api/v1/ruleList POST, will look up all the rules match the query below
{
	"Id": 0,     //mysql alert_rule id,modify or delete rule through id
	"name": "example",
	"expr": "sum (absent (example{job=\"example\"})) by (job_name)==1",
	"severity": "critical",
	"for": "1m",
	"sendInterval": "2m",
	"description": "example no data alert"
}

enable the mysql config in prometheus.yml
rule_mysql_config:
  host: ""
  port: ""
  db: ""
  user: ""
  password: ""

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -455,6 +455,9 @@ func main() {
 				cfg.GlobalConfig.ExternalLabels,
 			)
 		},
+		func(cfg *config.Config) error {
+			return ruleManager.StartRule(cfg.MysqlRuleConfig,cfg.GlobalConfig.ExternalLabels)
+		},
 	}
 
 	prometheus.MustRegister(configSuccess)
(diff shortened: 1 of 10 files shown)


brian-brazil:
Thanks for your PR. We purposefully do not attempt to maintain a database of configuration, nor do we depend on external databases. If you wish to build something like this you can do so on top of the yamls files we support.

Issue #6598, opened 2020-01-10
PromQL: Parentheses should work around any node type, in all contexts


juliusv:
EDIT: This should be a syntax error instead of being allowed.

---

ORIGINAL:

Parentheses should always just act like a no-op.

But they lead to query errors when used around some function parameters, like:

    rate((foo[5m]))

...or:

    label_join(foo, ("foo"), "bar")

This leads to errors like:

    Error executing query: unexpected error: interface conversion: promql.Expr is *promql.ParenExpr, not *promql.StringLiteral

The reason is that some function implementations assume concrete implementation types, like here: https://github.com/prometheus/prometheus/blob/64194f7d45cbb7fa2f41c8f219706308f65d511a/promql/functions.go#L66 and https://github.com/prometheus/prometheus/blob/64194f7d45cbb7fa2f41c8f219706308f65d511a/promql/functions.go#L754-L755.

Currently we don't have any ways of generating non-literal strings or matrices without matrix selectors, so I think we only need to ensure that parentheses are skipped around these child nodes before trying to use them.

roidelapluie:
I disagree with this, the only place where () make sense is whith subqueries and ops.

juliusv:
I can do a lot of things with parentheses that don't make sense, like `quantile((1), foo)` or `((((rate(foo[5m])))))`, so I don't see why those would work, but the other situations shouldn't. At the very least, it should be a syntax error if we don't want to allow it, it shouldn't just be an unexpected crash somewhere in the query.

roidelapluie:
() for scalar makes sense so you can do (1+1)*2

((rate(5m]))) also makes sense because you could do (3+(backend_enabled*rate(5m]))..
it is really different from rate((((xx[5m])))).

If we allow (vector) to be converted to vector, then rate((x)[4m]) could become valid? That seems really strange to me.

brian-brazil:
I'm with @roidelapluie on this one.

juliusv:
> () for scalar makes sense so you can do (1+1)*2
> 
> ((rate(5m]))) also makes sense because you could do (3+(backend_enabled*rate(5m]))..
it is really different from rate((((xx[5m])))).

Hm, IMO just because no non-function operations are defined with strings or matrices as operands doesn't mean that parentheses should not work. But if we do not want to allow them, then it should be a syntax error.

> If we allow (vector) to be converted to vector, then rate((x)[4m]) could become valid? That seems really strange to me.

[...]

PR #6612, opened 2020-01-13
PromQL: Fix string and parentheses handling in engine

### LINKED ISSUE #6610: PromQL: Allow engine to return strings
## Problem
`promql.ParseExpr` will parse expressions of all types without syntax errors, but the engine can only evaluate expressions of type `vector` and `matrix`.

## Proposal

Add a `ParseQuery` func that will behave like `ParseExpr` but will panic if the type of the AST root has a type that can not be evaluated by the engine, i.e. is a scalar or string.

This would also solve part of the problem pointed out in https://github.com/prometheus/prometheus/issues/6598#issuecomment-573624531.

slrtbtfs:
Will fix #6610.

There are still problems at the font end side.

The old UI handles string returns on queries like `"foo"` or `("bar")` just fine, however the react UI claims the return type is a scalar.

Could someone who is more familiar with the React UI look into this, maybe @juliusv or @boyskila ?

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -535,6 +535,7 @@ func (ng *Engine) execEvalStmt(ctx context.Context, query *query, s *EvalStmt) (
 			defaultEvalInterval: GetDefaultEvaluationInterval(),
 			logger:              ng.logger,
 		}
+
 		val, err := evaluator.Eval(s.Expr)
 		if err != nil {
 			return nil, warnings, err
@@ -542,10 +543,17 @@ func (ng *Engine) execEvalStmt(ctx context.Context, query *query, s *EvalStmt) (
 
 		evalSpanTimer.Finish()
 
-		mat, ok := val.(Matrix)
-		if !ok {
+		var mat Matrix
+
+		switch result := val.(type) {
+		case Matrix:
+			mat = result
+		case String:
+			return result, warnings, nil
+		default:
 			panic(errors.Errorf("promql.Engine.exec: invalid expression type %q", val.Type()))
 		}
+
 		query.matrix = mat
 		switch s.Expr.Type() {
 		case ValueTypeVector:
@@ -1324,6 +1332,8 @@ func (ev *evaluator) eval(expr Expr) Value {
 		res := newEv.eval(e.Expr)
 		ev.currentSamples = newEv.currentSamples
 		return res
+	case *StringLiteral:
+		return String{V: e.Val, T: ev.startTimestamp}
 	}
 
 	panic(errors.Errorf("unhandled expression of type: %T", expr))
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -1028,6 +1028,7 @@ func (ev *evaluator) eval(expr Expr) Value {
 
 	switch e := expr.(type) {
 	case *AggregateExpr:
+		unwrapParenExpr(&e.Param)
 		if s, ok := e.Param.(*StringLiteral); ok {
 			return ev.rangeEval(func(v []Value, enh *EvalNodeHelper) Vector {
 				return ev.aggregation(e.Op, e.Grouping, e.Without, s.Val, v[0].(Vector), enh)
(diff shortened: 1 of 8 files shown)


roidelapluie:
Can you please add tests?

```diff
diff --git a/promql/testdata/aggregators.test b/promql/testdata/aggregators.test
index 0da80576..7b3ba45b 100644
--- a/promql/testdata/aggregators.test
+++ b/promql/testdata/aggregators.test
@@ -233,6 +233,11 @@ eval instant at 5m count_values("version", version)
 	{version="7"} 2
 	{version="8"} 2
 
+eval instant at 5m count_values(((("version"))), version)
+	{version="6"} 5
+	{version="7"} 2
+	{version="8"} 2
+
 eval instant at 5m count_values without (instance)("version", version)
 	{job="api-server", group="production", version="6"} 3
 	{job="api-server", group="canary", version="8"} 2
```

```
    promql_test.go:32: unexpected error: error evaluating query "count_values((((\"version\"))), version)" (line 236): unexpected error: runtime error: index out of range [0] with length 0
```

slrtbtfs:
Thanks!

So I guess we should get to consensus about how to deal with these parentheses in #6598 first, i.e. figure out whether we should fix that kind of bug in the engine or let it already fail during parse time.

juliusv:
> The old UI handles string returns on queries like "foo" or ("bar") just fine, however the react UI claims the return type is a scalar.
>
> Could someone who is more familiar with the React UI look into this, maybe @juliusv or @boyskila ?

I think it's just a typo in the string handling case: https://github.com/prometheus/prometheus/blob/8cb2ab6a9913d7735c729a03964bf485053aca35/web/ui/react-app/src/DataTable.tsx#L106

Ironically we already even support strings as a root-level type in the UI (except for this typo), probably because the new UI code was partially adapted from the old UI code, and 1.x Prometheus once supported root-level strings already.

slrtbtfs:
Thanks, that fixed it.

brian-brazil:
> 1.x Prometheus once supported root-level strings already.

I don't think we ever did, though some of the code that'd be needed to support it was there.

juliusv:
@brian-brazil We did, I tried.

foostring

roidelapluie:
Can we please have negative tests? rate((foo)[5m]).

slrtbtfs:
Still working on another PR, this one will come after.

slrtbtfs:
This PR now (hopefully) fixes #6598, fixes #6610  and is ready for review.

promql/engine.go
```diff
@@ -1013,6 +1021,7 @@ func (ev *evaluator) eval(expr Expr) Value {
 
 	switch e := expr.(type) {
 	case *AggregateExpr:
+		unwrapParenExpr(&e.Param)
```
juliusv: I'd prefer not to mutate the main (non-cache etc.) fields of the AST during the query. E.g. a user might want to serialize or otherwise use it again after executing.

Could we make this functional?

```suggestion
		param := unwrapParenExpr(&e.Param)
```

juliusv:
:+1: besides the nit

promql/engine.go
slrtbtfs: I agree that functional execution would be nicer.

However, evaluation already mutates the AST in other places, so actually making this functional would be a larger effort.

https://github.com/prometheus/prometheus/blob/master/promql/engine.go#L1056

slrtbtfs: Re-evaluation might still produce correct results, but I wouldn't rely on it.

juliusv: Damn. My worst fears are confirmed ;) I think in this case it'd be really easy to avoid a mutation though, and fewer mutations cause fewer confusions.

slrtbtfs: I quickly skimmed the code and found a couple places where evaluation has side effects on (non-cache) AST contents.

All of them look relatively fixable.

If functional execution is desired, I'd prefer to fix all of them in a separate PR and leave this one as is.

juliusv: Ok, I'd say let's put that on hold for now then and keep things as is here.

slrtbtfs: I've made an issue for this #6627

slrtbtfs:
Merge conflicts are resolved now.

juliusv:
:+1:

PR #6613, opened 2020-01-13
feat promtool: added start_time_offset to rule tests

FUSAKLA:
Hi, we use some recording rules and alerts for alert inhibitions during holidays etc (inspired by https://github.com/roidelapluie/prometheus-timezone-holidays) and would like to write tests for it. Unfortunately to write for example test for Christmas we need to record whole year of data to get it tested which takes quite a while to record whole year of data. This way we can set the offset to the Christmas and omit the whole year of recording the data.
This would allow to shift the evaluation start time from 0 timestamp using added duration.

I know this is probably edge case but does not hurt to have the possibility and by default does not change the behavior.

### DIFF
--- a/cmd/promtool/unittest.go
+++ b/cmd/promtool/unittest.go
@@ -79,7 +79,7 @@ func ruleUnitTest(filename string) []error {
 	}
 
 	// Bounds for evaluating the rules.
-	mint := time.Unix(0, 0)
+	mint := time.Unix(0, 0).Add(unitTestInp.StartTimeOffset)
 	maxd := unitTestInp.maxEvalTime()
 	maxt := mint.Add(maxd)
 	// Rounding off to nearest Eval time (> maxt).
@@ -115,6 +115,7 @@ func ruleUnitTest(filename string) []error {
 type unitTestFile struct {
 	RuleFiles          []string      `yaml:"rule_files"`
 	EvaluationInterval time.Duration `yaml:"evaluation_interval,omitempty"`
+	StartTimeOffset    time.Duration `yaml:"start_time_offset,omitempty"`
 	GroupEvalOrder     []string      `yaml:"group_eval_order"`
 	Tests              []testGroup   `yaml:"tests"`
 }
--- a/docs/configuration/unit_testing_rules.md
+++ b/docs/configuration/unit_testing_rules.md
@@ -25,6 +25,11 @@ rule_files:
 # optional, default = 1m
 evaluation_interval: <duration>
 
+# optional, default = 0m
+# By default the time is mocked and starts from 0 timestamp.
+# Here you can add start time offset duration to shift the evaluation.
+start_time_offset: <duration>
+
 # The order in which group names are listed below will be the order of evaluation of
 # rule groups (at a given evaluation time). The order is guaranteed only for the groups mentioned below.
 # All the groups need not be mentioned below.


brian-brazil:
Can you provide an example test which is too slow?

jirislav:
> Can you provide an example test which is too slow?

Of course, here is an example which tests that the `business_day` recording rule returns `nil` on Christmas.

Note that due to staleness, the `evaluation_interval` must be lower than `5m`, which implies very long time for the test to complete.

```yml
rule_files:
- working-hours.yaml

evaluation_interval: 4m

tests:
- interval: 1m
  promql_expr_test:
  # 1970-12-24T00:00 is Christmas Eve
  - expr: business_day
    eval_time: 8568h
  # 1970-12-25T00:00 is Christmas Day
  - expr: business_day
    eval_time: 8592h
  # 1970-12-26T00:00 is St. Stephen's Day
  - expr: business_day
    eval_time: 8616h

```

FUSAKLA:
Yes the example provided by @jirislav is exactly the case.

Minimal example can be for example:
rules.yaml 
```yaml
groups:
  - name: test
    interval: 30s
    rules:
      - record: test_metric
        expr: 1
```

test.yaml 
```yaml
rule_files:
  - rules.yaml
evaluation_interval: 4m
tests:
- interval: 1m
  promql_expr_test:
  - expr: test_metric
    eval_time: 8568h
    exp_samples:
        - labels: "test_metric"
          value: 1
```

```bash
$ time ./promtool test rules test.yaml 
Unit Testing:  test.yaml
  SUCCESS

real	0m4.230s
user	0m8.090s
sys	0m0.689s
```

4s for one test is quite a lot, considering that with using the time start offset it's almost in no time not mentioning sparing CPU time.

test.yaml
```yaml
rule_files:
- rules.yaml
evaluation_interval: 4m
start_time_offset: 8568h
tests:
- interval: 1m
  promql_expr_test:
  - expr: test_metric
    eval_time: 0m
    exp_samples:
        - labels: "test_metric"
          value: 1
```

```bash
$ time ./promtool test rules test.yaml 
Unit Testing:  test.yaml
  SUCCESS

real	0m0.042s
user	0m0.039s
sys	0m0.011s
```

cmd/promtool/unittest.go
```diff
@@ -79,7 +79,7 @@ func ruleUnitTest(filename string) []error {
 	}
 
 	// Bounds for evaluating the rules.
-	mint := time.Unix(0, 0)
+	mint := time.Unix(0, 0).Add(unitTestInp.StartTimeOffset)
```
brian-brazil: This is only catching the evals, this should affect the series too.

docs/configuration/unit_testing_rules.md
```diff
@@ -25,6 +25,11 @@ rule_files:
 # optional, default = 1m
 evaluation_interval: <duration>
 
+# optional, default = 0m
```
brian-brazil: This isn't our usual way of documenting either optional or default options. Can you line it up with the Prometheus configuration docs?

docs/configuration/unit_testing_rules.md
```diff
@@ -25,6 +25,11 @@ rule_files:
...
+# By default the time is mocked and starts from 0 timestamp.
```
brian-brazil: No need to mention mocking, as among other things we don't actually mock and if we did that'd be an implementation detail.

docs/configuration/unit_testing_rules.md
```diff
@@ -25,6 +25,11 @@ rule_files:
...
+# Here you can add start time offset duration to shift the evaluation.
```
brian-brazil: Describe what it does, not how you could use it. These are reference docs.

cmd/promtool/unittest.go
codesome: PS: While taking care for series, should make sure that the samples should be loaded till mint minus maximum lookback in any rule queries.

brian-brazil:
If you are still interested in this PR, there are unaddressed comments.

roidelapluie:

It seems like the original author did not reply in 3 years, let's close this pull request.

PR #6615, opened 2020-01-13
PromQL: Add position metadata to the AST

### LINKED ISSUE #6597: PromQL: Add position metadata to the AST
## Motivation

For the PromQL language server, it is crucial to be able to map string positions to AST Nodes in PromQL queries.

## Proposal

I'd like to add the following function to the `promql.Node` interface:

```
PositionRange() PositionRange
```

where

```
type PositionRange struct {
    Start int
    End int
}
```
contains the Position of the corresponding AST Node as a byte offset in the input string.

The necessary metadata will be stored in the AST Nodes.

This would essentially be a more lightweight and less complex version of #6061.

---

Would people be ok if I go ahead with these changes as described?

cc @brian-brazil @juliusv @simonpasquier

slrtbtfs:
Fixes #6597 .


promql/ast.go
```diff
@@ -79,11 +82,12 @@ type Expressions []Expr
...
-	Expr     Expr     // The Vector expression over which is aggregated.
-	Param    Expr     // Parameter used by some aggregators.
-	Grouping []string // The labels by which to group the Vector.
-	Without  bool     // Whether to drop the given labels rather than keep them.
+	Op            ItemType // The used aggregation operation.
+	Expr          Expr     // The Vector expression over which is aggregated.
+	Param         Expr     // Parameter used by some aggregators.
+	Grouping      []string // The labels by which to group the Vector.
+	Without       bool     // Whether to drop the given labels rather than keep them.
+	positionRange PositionRange
```
brian-brazil: These should probably be public for the sake of 3rd party users of this code.

promql/ast.go
```diff
@@ -319,3 +337,82 @@ func Children(node Node) []Node {
...
 }
+
+// PositionRange describes a position in the input string of the parser.
+type PositionRange struct {
+	Start Pos
+	End   Pos
+}
+
+// mergeRanges is a helper function the PostionRanges of two Nodes
+// Note that the arguments must already be in the same order as they
```
brian-brazil: must be in the order they occur in the input string


The "already" implies you're passing a slice.

promql/ast.go
```diff
@@ -319,3 +337,82 @@ func Children(node Node) []Node {
...
+// occur in the input string.
+func mergeRanges(first Node, last Node) PositionRange {
+	return PositionRange{
+		Start: first.PositionRange().Start,
+		End:   last.PositionRange().End,
+	}
+}
+
+// Item implement the Node interface
```
brian-brazil: implements

promql/ast.go
```diff
@@ -319,3 +337,82 @@ func Children(node Node) []Node {
...
+}
+func (e *Call) PositionRange() PositionRange {
+	return e.positionRange
+}
+func (e *EvalStmt) PositionRange() PositionRange {
+	return e.Expr.PositionRange()
+}
+func (e Expressions) PositionRange() PositionRange {
+	if len(e) == 0 {
+		// Position undefined
```
brian-brazil: full stop

promql/parse.go
```diff
@@ -41,6 +41,10 @@ type parser struct {
 	inject    ItemType
 	injecting bool
 
+	// Everytime a Item is lexed that could be the end
```
brian-brazil: an Item


"an" before vowels, and vowel sounds.

promql/ast.go
```diff
@@ -319,3 +337,82 @@ func Children(node Node) []Node {
...
 	}
 }
+
+// PositionRange describes a position in the input string of the parser.
+type PositionRange struct {
+	Start Pos
+	End   Pos
+}
+
+// mergeRanges is a helper function the PostionRanges of two Nodes
```
juliusv: This first comment line doesn't parse in my head (and needs a period)? Should it read:

```suggestion
// mergeRanges is a helper function to merge the PositionRanges of two Nodes.
```

promql/ast.go
slrtbtfs: The position data can be read accessed trough the public `PositionRange` interface function.

Are there any 3rd parties that want write access this?

brian-brazil: I don't know, but probably doesn't hurt to let them considering they can write everything else.

promql/ast.go
slrtbtfs: Thanks

promql/ast.go
slrtbtfs: Well, only almost everything.

https://github.com/prometheus/prometheus/blob/201491fd1b6b02fab60f7630dff23218bc3d1067/promql/ast.go#L152

brian-brazil: That's part of the engine implementation, not the AST.

slrtbtfs: I've made them public.

promql/ast.go
```diff
@@ -117,29 +125,37 @@ type SubqueryExpr struct {
...
+	PosRange PositionRange
 }
 
 // UnaryExpr represents a unary operation on another expression.
 // Currently unary operations are only supported for Scalars.
 type UnaryExpr struct {
 	Op   ItemType
 	Expr Expr
+
+	startPos Pos
```
brian-brazil: You missed one

brian-brazil:
Thanks!

PR #6616, opened 2020-01-13
small tsdb fixes

roidelapluie:
That metric was not registered.

I also reordered the metrics in the list.


brian-brazil:
Thanks!

PR #6618, opened 2020-01-13
tsdb: writePostingsToTmpFiles returns nil instead of err

roidelapluie:

@brian-brazil I am less confident about this one.


brian-brazil:
That's a pretty obvious bug, thanks!

PR #6621, opened 2020-01-13
Remove old checkpoint dir if it still exists

### LINKED ISSUE #6619: When checkpoint is not successful, next checkpoint is twice the size.
## Bug Report
**What did you do?**

Start prometheus with aggressive settings (--storage.tsdb.max-block-duration=10m --storage.tsdb.min-block-duration=1m)

Wait for a checkpoint:
data/wal/checkpoint.000040/

Let's say that the checkpoint has a size of 5M

Copy the checkpoint to data/wal/checkpoint.000042.tmp

_It simulates the case where the tmp checkpoint is not written completely_

when checkpoint 0000042 is written, its size is 5M

checkpoint 0000042 is 10M, because it contains the tmp that was not written successfully.

**Environment**

Linux

* Prometheus version:

2.15.2

roidelapluie:

Fix #6619

### DIFF
--- a/tsdb/wal/checkpoint.go
+++ b/tsdb/wal/checkpoint.go
@@ -133,6 +133,10 @@ func Checkpoint(w *WAL, from, to int, keep func(id uint64) bool, mint int64) (*C
 	cpdir := filepath.Join(w.Dir(), fmt.Sprintf(checkpointPrefix+"%06d", to))
 	cpdirtmp := cpdir + ".tmp"
 
+	if err := os.RemoveAll(cpdirtmp); err != nil {
+		return nil, errors.Wrap(err, "remove previous temporary checkpoint dir")
+	}
+
 	if err := os.MkdirAll(cpdirtmp, 0777); err != nil {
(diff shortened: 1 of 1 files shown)


tsdb/wal/checkpoint.go
```diff
@@ -133,6 +133,13 @@ func Checkpoint(w *WAL, from, to int, keep func(id uint64) bool, mint int64) (*C
 	cpdir := filepath.Join(w.Dir(), fmt.Sprintf(checkpointPrefix+"%06d", to))
 	cpdirtmp := cpdir + ".tmp"
 
+	if _, err := os.Stat(cpdirtmp); !os.IsNotExist(err) {
```
brian-brazil: This error check feels off. How about always trying, and then use the error check on the removal?

roidelapluie: ack

roidelapluie: done

brian-brazil:
Thanks!

PR #6626, opened 2020-01-14
Fix curl args for data-urlencode in docs

claytono:
The curl argument `--data-urlencode` doesn't accept equal signs between the arg and the value, but needs whitespace instead.


codesome:
Thanks!

PR #6629, opened 2020-01-14
Unify react fetcher components

boyskila:
With this PR, components fetching data from the server are enhanced with HOC which acts as a guard from eventual errors, undefined data and puts a loader while waiting for the response.

Pages covered

- [x] Targets page
- [x] TSDBStatus page
- [x] Alerts page
- [x] Service Discovery page
- [x] Graph page

### DIFF
--- a/web/ui/react-app/src/pages/graph/PanelList.tsx
+++ b/web/ui/react-app/src/pages/graph/PanelList.tsx
@@ -107,10 +107,11 @@ const opts: RequestInit = {
 const PanelList: FC<RouteComponentProps & PathPrefixProps> = ({ pathPrefix = '' }) => {
   const memoizedFetch = useCallback(useFetch, [pathPrefix]);
   const [delta, setDelta] = useState(0);
+  const [panels, setPanels] = useState<PanelMeta[]>([]);
   const [useLocalTime, setUseLocalTime] = useLocalStorage('use-local-time', false);
   const [enableQueryHistory, setEnableQueryHistory] = useLocalStorage('enable-query-history', false);
 
-  const { response: metricsResponse, error: metricsErr } = memoizedFetch<string[]>(
+  const { response: metricsResponse, error: metricsError } = memoizedFetch<string[]>(
     `${pathPrefix}/api/v1/label/__name__/values`,
     opts
   );
@@ -120,6 +121,11 @@ const PanelList: FC<RouteComponentProps & PathPrefixProps> = ({ pathPrefix = '' 
     `${pathPrefix}/api/v1/query?query=time()`,
     opts
   );
+
+  useEffect(() => {
+    setPanels(decodePanelOptionsFromQueryString(window.location.search));
+  }, []);
+
   useEffect(() => {
     if (timeResponse.data) {
       const serverTime = timeResponse.data.result[0];
@@ -148,18 +154,18 @@ const PanelList: FC<RouteComponentProps & PathPrefixProps> = ({ pathPrefix = '' 
         <Alert color="danger">
           <strong>Warning: </strong>
           {timeError && `Unexpected response status when fetching server time: ${timeError.message}`}
-          {delta > 30 &&
+          {delta >= 30 &&
             `Error fetching server time: Detected ${delta} seconds time difference between your browser and the server. Prometheus relies on accurate time and time drift might cause unexpected query results.`}
         </Alert>
       )}
-      {metricsErr && (
+      {metricsError && (
         <Alert color="danger">
           <strong>Warning: </strong>
-          {`Error fetching metrics list: Unexpected response status when fetching metric names: ${metricsErr.message}`}
+          {`Error fetching metrics list: Unexpected response status when fetching metric names: ${metricsError.message}`}
         </Alert>
       )}
       <PanelListContent
-        panels={decodePanelOptionsFromQueryString(window.location.search)}
+        panels={panels}
         pathPrefix={pathPrefix}
         useLocalTime={useLocalTime}
         metrics={metricsResponse.data}
--- a/web/ui/react-app/src/pages/graph/PanelList.tsx
+++ b/web/ui/react-app/src/pages/graph/PanelList.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useState, useEffect } from 'react';
+import React, { FC, useState, useEffect, useCallback } from 'react';
 import { RouteComponentProps } from '@reach/router';
 import { Alert, Button } from 'reactstrap';
 
@@ -6,7 +6,7 @@ import Panel, { PanelOptions, PanelDefaultOptions } from './Panel';
 import Checkbox from '../../components/Checkbox';
 import PathPrefixProps from '../../types/PathPrefixProps';
 import { generateID, decodePanelOptionsFromQueryString, encodePanelOptionsToQueryString, callAll } from '../../utils';
-import { withStatusIndicator } from '../../components/withStatusIndicator';
+import { useFetch } from '../../hooks/useFetch';
 import { useLocalStorage } from '../../hooks/useLocalStorage';
 
 export type PanelMeta = { key: string; options: PanelOptions; id: string };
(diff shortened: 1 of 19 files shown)


web/ui/react-app/src/pages/serviceDiscovery/Services.tsx
```diff
@@ -20,100 +19,95 @@ export interface TargetLabels {
   isDropped: boolean;
 }
 
-const Services: FC<RouteComponentProps & PathPrefixProps> = ({ pathPrefix }) => {
-  const { response, error } = useFetch<ServiceMap>(`${pathPrefix}/api/v1/targets`);
-
-  const processSummary = (response: ServiceMap) => {
-    const targets: any = {};
+export const processSummary = (activeTargets: Target[], droppedTargets: DroppedTarget[]) => {
```
juliusv: Do `processSummary` and `processTargets` need to be exported? Seems they're only used from within this file.

boyskila: Yes, for testing.

web/ui/react-app/src/pages/serviceDiscovery/Services.tsx
```diff
@@ -20,100 +19,95 @@ export interface TargetLabels {
...
+        ))}
+      </ul>
+      <hr />
+      {mapObjEntries(labels, ([k, v]) => {
+        return <LabelsTable value={v} name={k} key={k} />;
+      })}
+    </>
+  );
+};
+ServiceDiscoveryContent.displayName = 'ServiceDiscoveryContent';
```
juliusv: IMO these display names should better be normal human-readable words like "service discovery information", so the error becomes "Error: Error fetching service discovery information: Failed to fetch" instead of something more cryptic.

web/ui/react-app/src/pages/serviceDiscovery/Services.tsx
juliusv: I don't see them being used in any tests yet though... but it would make sense at that point, yeah.

web/ui/react-app/src/pages/serviceDiscovery/Services.tsx
boyskila: IMO ```displayName``` should be used as intended i.e should display the name of the component.
The default error message template use that name but if we want to provide more than that we can use ```customErrorMsg``` prop

web/ui/react-app/src/pages/serviceDiscovery/Services.tsx
boyskila: yes, yes...exported for the future tests :)

web/ui/react-app/src/pages/serviceDiscovery/Services.tsx
juliusv: But the component name isn't really relevant to the user and just looks cryptic... so yeah, let's at least show the user the normal English wording :)

boyskila: Yeah, for the components which have not user-friendly names I'll create a custom error message.

boyskila:
@juliusv, please review.

web/ui/react-app/src/pages/graph/PanelList.tsx
```diff
@@ -1,219 +1,181 @@
...
+const PanelList: FC<RouteComponentProps & PathPrefixProps> = ({ pathPrefix }) => {
+  const [metricNames, setMetricNames] = useState<string[]>([]);
+  const [timeDriftError, setTimeDriftError] = useState();
+  const [fetchMetricsError, setFetchMetricsError] = useState();
+  const [useLocalTime, setUseLocalTime] = useLocalStorage('use-local-time', false);
+  const [enableQueryHistory, setEnableQueryHistory] = useLocalStorage('enable-query-history', false);
+
+  useEffect(() => {
+    (async () => {
+      const metricsResponse = await fetch(`${pathPrefix}/api/v1/label/__name__/values`, { cache: 'no-store' });
```
juliusv: Meanwhile in master the code changed to also include `credentials: 'same-origin'` in the fetch options (also in useFetch).

web/ui/react-app/src/pages/graph/PanelList.tsx
```diff
@@ -1,219 +1,181 @@
...
-        ))}
-        <Button color="primary" className="add-panel-btn" onClick={this.addPanel}>
-          Add Panel
-        </Button>
-      </>
-    );
-  }
-}
+      const browserTime = new Date().getTime() / 1000;
+      const timeResponse = await fetch(`${pathPrefix}/api/v1/query?query=time()`, {
```
juliusv: IIUC by switching to async/await for two subsequent requests (metric names and time), the requests are now done in sequence instead of in parallel, so maybe the page load will take a tiny bit longer? Any special reason to switch to async/await instead of callbacks here, or just preference?

juliusv:
Could you give this a rebase on latest master? Sorry, I somehow thought we already merged this :grimacing:

web/ui/react-app/src/pages/graph/PanelList.tsx
```diff
@@ -1,219 +1,178 @@
...
+  useEffect(() => {
+    !panels.length && addPanel();
     window.onpopstate = () => {
       const panels = decodePanelOptionsFromQueryString(window.location.search);
       if (panels.length > 0) {
-        this.setState({ panels });
+        setPanels(panels);
       }
     };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
juliusv: In cases like these, could you add a comment why the exception is necessary (what problem would happen otherwise)?

web/ui/react-app/src/pages/graph/PanelList.tsx
```diff
@@ -1,219 +1,178 @@
...
+  useEffect(() => {
+    setPanels(decodePanelOptionsFromQueryString(window.location.search));
+  }, []);
+
+  useEffect(() => {
+    if (timeResponse.data) {
+      const serverTime = timeResponse.data.result[0];
+      setDelta(Math.abs(browserTime - serverTime));
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
juliusv: Same here.

boyskila: I should find a way to disable those warnings officially(in the project). It's annoying because react doesn't leave you to decide what should be the dependencies list. For example, here I don't want to include browserTime since its changed on every re-render but react claims that this should be part of it even it automatically include it on save

boyskila: Or maybe it's not a good idea to disable it globally as per https://github.com/facebook/create-react-app/issues/6880

web/ui/react-app/src/pages/graph/PanelList.tsx
```diff
@@ -1,219 +1,178 @@
...
+      ))}
+      <Button color="primary" onClick={addPanel}>
+        Add Panel
+      </Button>
+    </>
+  );
+};
+
+const opts: RequestInit = {
+  cache: 'no-store',
```
juliusv: We should probably just make `no-store` the default in `useFetch()`, instead of `no-cache`, since we always want to fully re-do the fetches when we execute them.

web/ui/react-app/src/pages/graph/PanelList.tsx
```diff
@@ -1,219 +1,178 @@
...
+    if (timeResponse.data) {
+      const serverTime = timeResponse.data.result[0];
+      setDelta(Math.abs(browserTime - serverTime));
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [timeResponse.data]);
+
+  return (
+    <>
+      <Checkbox
```
juliusv: This changes the checkboxes to be underneath instead of beside each other. Could you re-add the `<Row>` around them that's in master?

web/ui/react-app/src/pages/graph/PanelList.tsx
boyskila: Ok, I'll change the default useFetch options to use ```no-store``` and I'll omit it here, thanks

web/ui/react-app/src/pages/graph/PanelList.tsx
boyskila: Hm, they were like that as far as I remember, but yeah, ok will put them one to each other

web/ui/react-app/src/pages/graph/PanelList.tsx
```diff
@@ -1,219 +1,178 @@
...
+              )
+            )
+          }
+          useLocalTime={useLocalTime}
+          metricNames={metrics}
+          pastQueries={queryHistoryEnabled ? historyItems : []}
+          pathPrefix={pathPrefix}
+        />
+      ))}
+      <Button color="primary" onClick={addPanel}>
```
juliusv: You lost the `className="add-panel-btn"` here, which is required to add the bottom margin for the button (when there's a vertical scrollbar, it looks better if you can scroll a bit further than the button, otherwise it always touches the bottom of the window).

web/ui/react-app/src/pages/graph/PanelList.tsx
```diff
@@ -1,219 +1,178 @@
...
+  const [panels, setPanels] = useState(rest.panels);
+  const [historyItems, setLocalStorageHistoryItems] = useLocalStorage<string[]>('history', []);
+
+  useEffect(() => {
+    !panels.length && addPanel();
     window.onpopstate = () => {
       const panels = decodePanelOptionsFromQueryString(window.location.search);
       if (panels.length > 0) {
-        this.setState({ panels });
+        setPanels(panels);
```
juliusv: Somehow panel options aren't loaded correctly from the query string anymore.

Try `https://deploy-preview-6629--prometheus-react.netlify.com/new/graph?g0.expr=irate(node_cpu_seconds_total%5B1m%5D)&g0.tab=0&g0.stacked=1&g0.range_input=1m&g1.expr=irate(node_cpu_seconds_total%5B1m%5D)&g1.tab=1&g1.stacked=0&g1.range_input=1h`

...it will just change itself to `https://deploy-preview-6629--prometheus-react.netlify.com/new/graph?g0.expr=&g0.tab=1&g0.stacked=0&g0.range_input=1h` and not show any panels.

web/ui/react-app/src/pages/graph/PanelList.tsx
```diff
@@ -1,219 +1,168 @@
...
+
+  const browserTime = new Date().getTime() / 1000;
+  const { response: timeRes, error: timeErr } = useFetch<{ result: number[] }>(`${pathPrefix}/api/v1/query?query=time()`);
+
+  useEffect(() => {
+    if (timeRes.data) {
+      const serverTime = timeRes.data.result[0];
+      setDelta(Math.abs(browserTime - serverTime));
+    }
+    // React wants to include browserTime to useEffect dependencie list which will cause a delta change on every re-render
```
juliusv:
```suggestion
    // React wants to include browserTime to the useEffect dependencies list, which will cause a delta change on every re-render
```

web/ui/react-app/src/pages/graph/PanelList.tsx
```diff
@@ -1,219 +1,168 @@
...
+    // Basically to disable this rule it's not recommended but this is the only way to take control over the useEffect and not to include
```
juliusv:
```suggestion
    // Basically it's not recommended to disable this rule, but this is the only way to take control over the useEffect dependencies and to not include the
```

web/ui/react-app/src/pages/graph/PanelList.tsx
```diff
@@ -1,219 +1,168 @@
...
+  useEffect(() => {
+    !panels.length && addPanel();
     window.onpopstate = () => {
       const panels = decodePanelOptionsFromQueryString(window.location.search);
       if (panels.length > 0) {
-        this.setState({ panels });
+        setPanels(panels);
       }
     };
+    // We want useEffect to act only as componentDidMount, but react still complain about the empty dependencies list.
```
juliusv:
```suggestion
    // We want useEffect to act only as componentDidMount, but react still complains about the empty dependencies list.
```

web/ui/react-app/src/pages/graph/PanelList.tsx
juliusv: If you look at the left in the diff, there was a Row around them :) But inline-block is also fine, yeah

juliusv:
:+1: Thanks!

PR #6634, opened 2020-01-15
PromQL: Use PositionRange in parser errors

slrtbtfs:
This PR changes the `ParseErr` struct of the Prometheus parser to use the new `PositionRange` structure. This is especially useful for the language server, since it allows underlining the wrong parts in expressions with errors.

Additionally type errors now report accurate positions. Before, all type errors had the end of the query as their position.

### DIFF
--- a/promql/parse.go
+++ b/promql/parse.go
@@ -175,12 +175,12 @@ func (p *parser) typecheck(node Node) (err error) {
 }
 
 // errorf formats the error and terminates processing.
-func (p *parser) errorf(positionRange PositionRange, format string, args ...interface{}) {
-	p.error(positionRange, errors.Errorf(format, args...))
+func (p *parser) failf(positionRange PositionRange, format string, args ...interface{}) {
+	p.fail(positionRange, errors.Errorf(format, args...))
 }
 
 // error terminates processing.
-func (p *parser) error(positionRange PositionRange, err error) {
+func (p *parser) fail(positionRange PositionRange, err error) {
 	perr := &ParseErr{
 		PositionRange: positionRange,
 		Err:           err,
@@ -208,7 +208,7 @@ func (p *parser) unexpected(context string, expected string) {
 		errMsg.WriteString(expected)
 	}
 
-	p.error(p.yyParser.lval.item.PositionRange(), errors.New(errMsg.String()))
+	p.fail(p.yyParser.lval.item.PositionRange(), errors.New(errMsg.String()))
 }
 
 var errUnexpected = errors.New("unexpected error")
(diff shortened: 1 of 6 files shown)


promql/parse.go
```diff
@@ -151,19 +175,16 @@ func (p *parser) typecheck(node Node) (err error) {
 }
 
 // errorf formats the error and terminates processing.
-func (p *parser) errorf(format string, args ...interface{}) {
-	p.error(errors.Errorf(format, args...))
+func (p *parser) errorf(positionRange PositionRange, format string, args ...interface{}) {
```
brian-brazil: Having an errorf that doesn't have the usual printf/errorf signature is weird.

promql/parse.go
```diff
@@ -281,12 +302,23 @@ func (p *parser) newBinaryExpression(lhs Node, op Item, modifiers Node, rhs Node
 	ret.RHS = rhs.(Expr)
 	ret.Op = op.Typ
 
+	// opRange returns the PositionRange of the operator part of the BinaryExpr.
+	// This is made a function instead of a variable, so it is lazily evaluated on demand.
+	opRange := func() (r PositionRange) {
+		//Remove whitespace at the beginning and end of the range
```
brian-brazil: Formatting.

promql/parse.go
slrtbtfs: So I should rename it?

promql/parse.go
slrtbtfs: What exactly is wrong with formatting?

The unnecessary line breaks in the for loops are enforced by gofmt.

brian-brazil: No space after the //, no .

promql/parse.go
brian-brazil: That'd be best I think

promql/parse.go
slrtbtfs: Oh, I see.

promql/parse.go
```diff
@@ -151,19 +175,16 @@ func (p *parser) typecheck(node Node) (err error) {
 }
 
 // errorf formats the error and terminates processing.
```
brian-brazil: Comments needs updating.

roidelapluie:
What about errorfAtPos ?

slrtbtfs:
I'm working on another PR which will allow having a list of errors.

`addError` and `addErrorf` might be an option, then.

brian-brazil:
Thanks!

roidelapluie:
I have to say that the result is quite impressive. Thanks!

PR #6641, opened 2020-01-16
scrape: fix typo.

johncming:


brian-brazil:
This is fine as is. We also should be trying to reduce use of model, and I'm not a fan of replacing string constants with indirection.

PR #6642, opened 2020-01-16
React UI: hide non selected alert types

boyskila:
Fixes: https://github.com/prometheus/prometheus/issues/6527#issuecomment-573363059

### DIFF
--- a/web/ui/react-app/src/pages/alerts/AlertContents.tsx
+++ b/web/ui/react-app/src/pages/alerts/AlertContents.tsx
@@ -51,17 +51,17 @@ const stateColorTuples: Array<[RuleState, 'success' | 'warning' | 'danger']> = [
 ];
 
 const AlertsContent: FC<AlertsProps> = ({ groups = [], statsCount }) => {
-  const [state, setState] = useState<RuleStatus<boolean>>({
+  const [filter, setFilter] = useState<RuleStatus<boolean>>({
     firing: true,
     pending: true,
     inactive: true,
   });
   const [showAnnotations, setShowAnnotations] = useState(false);
 
-  const toggle = (ruleState: RuleState) => () => {
-    setState({
-      ...state,
-      [ruleState]: !state[ruleState],
+  const toggleFilter = (ruleState: RuleState) => () => {
+    setFilter({
+      ...filter,
+      [ruleState]: !filter[ruleState],
     });
   };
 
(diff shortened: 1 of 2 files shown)


web/ui/react-app/src/pages/alerts/AlertContents.tsx
```diff
@@ -70,46 +70,57 @@ const AlertsContent: FC<AlertsProps> = ({ groups = [], statsCount }) => {
       <div className="d-flex togglers-wrapper">
         {stateColorTuples.map(([state, color]) => {
           return (
-            <Checkbox wrapperStyles={{ marginRight: 10 }} defaultChecked id={`${state}-toggler`} onClick={toggle(state)}>
+            <Checkbox
+              key={color}
```
juliusv: Might be better to key on `state`, just in case colors are ever not unique in the future.

web/ui/react-app/src/pages/alerts/AlertContents.tsx
```diff
@@ -70,46 +70,57 @@ const AlertsContent: FC<AlertsProps> = ({ groups = [], statsCount }) => {
...
           );
         })}
-        <Checkbox
-          wrapperStyles={{ marginLeft: 'auto' }}
-          id="show-annotations-toggler"
-          onClick={() => setShowAnnotations(!showAnnotations)}
-        >
-          <span style={{ fontSize: '0.9rem', lineHeight: 1.9 }}>Show annotations</span>
-        </Checkbox>
+        {/* hide if all filters are off */}
```
juliusv: I would always show it - I don't see the association of this checkbox to the filter state of the groups, it's more of a global page setting...

boyskila: The association is not with the filters but with the fact that once nothing is left on the screen this checkbox becomes unusable. The other option is to disable it but if you prefer I'll leave it without any modifications

web/ui/react-app/src/pages/alerts/AlertContents.tsx
boyskila: yes, ```state``` would be better, thanks!

web/ui/react-app/src/pages/alerts/AlertContents.tsx
juliusv: You mean it doesn't affect anything that's currently visible (which seems fine to me), or it actually becomes broken somehow in that sense that it can't be clicked properly?

boyskila: Nothing breaks, just become unusable and something unusable makes more sense to me to be prevented from usage at all. Hiding it wasn't the best option yes :) but at least to disable it :)

juliusv: :man_shrugging: I dunno. I'd be for still just showing it TBH.

boyskila: Yeah, ok :)

juliusv:
:+1:

PR #6644, opened 2020-01-16
made stripe size configurable

thorfour:
This changes exposes the `stripeSize` const as a configurable option in tsdb. Not passing in the option will result in the tsdb using the default stripe size that is uses today.

This is an important change for Cortex's new experimental TSDB feature as it allows us to reduce the memory footprint of TSDB's as they are scaled to many nodes. More info about this memory usage can be found [here](https://github.com/cortexproject/cortex/pull/1947)

@codesome @krasi-georgiev

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -57,6 +57,7 @@ var DefaultOptions = &Options{
 	NoLockfile:             false,
 	AllowOverlappingBlocks: false,
 	WALCompression:         false,
+	StripeSize:             DefaultStripeSize,
 }
 
 // Options of the DB storage.
@@ -89,6 +90,9 @@ type Options struct {
 
 	// WALCompression will turn on Snappy compression for records on the WAL.
 	WALCompression bool
+
+	// StripeSize is the size in entries of the series hash map. Reducing the size will save memory but impact perfomance.
+	StripeSize int
 }
 
 // Appender allows appending a batch of data. It must be completed with a
@@ -308,7 +312,7 @@ func (db *DBReadOnly) FlushWAL(dir string) error {
 	if err != nil {
 		return err
 	}
-	head, err := NewHead(nil, db.logger, w, 1)
+	head, err := NewHead(nil, db.logger, w, 1, DefaultStripeSize)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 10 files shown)


tsdb/db.go
```diff
@@ -487,6 +491,9 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db
 	if opts == nil {
 		opts = DefaultOptions
 	}
+	if opts.StripeSize == 0 {
```
codesome: `<=`

tsdb/head.go
```diff
@@ -1537,29 +1537,34 @@ func (m seriesHashmap) del(hash uint64, lset labels.Labels) {
 	}
 }
 
+const (
+	defaultStripeSize = 1 << 14
```
codesome: If we are allowing changing of stripe size, we may as well expose this too.

tsdb/head.go
```diff
@@ -1597,7 +1602,7 @@ func (s *stripeSeries) gc(mint int64) (map[uint64]struct{}, int) {
 				// series alike.
 				// If we don't hold them all, there's a very small chance that a series receives
 				// samples again while we are half-way into deleting it.
-				j := int(series.ref & stripeMask)
+				j := int(series.ref) & (s.size - 1)
```
codesome: `s.size - 1` as the mask assumes that the size is a multiple of 2. To avoid any additional logic to take care of non-multiples of 2, we can document it in `NewHead`.

web/api/v1/api_test.go
```diff
@@ -1875,7 +1875,7 @@ func (f *fakeDB) Dir() string {
 }
 func (f *fakeDB) Snapshot(dir string, withHead bool) error { return f.err }
 func (f *fakeDB) Head() *tsdb.Head {
-	h, _ := tsdb.NewHead(nil, nil, nil, 1000)
+	h, _ := tsdb.NewHead(nil, nil, nil, 1000, 1<<14)
```
codesome: This should be replaced after exposing the constant to avoid any confusion about where `1<<14` came from.

tsdb/db.go
thorfour: Fixed.

tsdb/head.go
thorfour: Good point. Fixed.

tsdb/head.go
thorfour: Added a comment in NewHead. Do you think we should also include a logic guard in NewHead or just leave it as a comment?

web/api/v1/api_test.go
thorfour: Fixed.

tsdb/head.go
codesome: Comment should be enough

thorfour:
@krasi-georgiev any chance you could look at these changes?

krasi-georgiev:
LGTM

PR #6646, opened 2020-01-16
Add pkg logging on windows

roidelapluie:


.circleci/config.yml
```diff
@@ -56,7 +56,7 @@ jobs:
     steps:
     - checkout
     # TSDB is where the most risk is Windows wise, so only test there for now.
```
brian-brazil: Comment needs updating.

roidelapluie:
Actually I wonder if that is needed .. I expected it to fail on windows with the error that the file is ope by another process.. it does not fail

brian-brazil:
If the test passes on Windows now, it doesn't hurt to run it in CI.

Issue #6577, opened 2020-01-08
Export metric for WAL write errors


kormat:
When prometheus is unable to write to the WAL, it logs a warning, but there is no metric exported to allow alerting on this condition. This means that an operator might not know there's a corrupted/truncated WAL until the next compaction cycle runs (assuming they have an alert on `prometheus_tsdb_wal_truncations_failed_total`).

Example log entries:
```
level=warn ts=2020-01-08T11:57:49.195Z caller=scrape.go:945 component="scrape manager" scrape_pool=border_routers target=<redacted>/metrics msg="appending scrape report failed" err="write to WAL: log samples: write data/wal/00000000: no space left on device"
level=warn ts=2020-01-08T11:57:49.195Z caller=manager.go:584 component="rule manager" group=alert.rules msg="rule sample appending failed" err="write to WAL: log samples: write data/wal/00000000: no space left on device"
```

roidelapluie:
What is the value of   prometheus_tsdb_wal_corruptions_total ?

cstyan:
In this particular case you'd probably notice via any alerts regarding disk space, but I see no harm in adding a counter for `WAL writes failed`.

jpmcb:
Hi there! I'd like to work on this issue and get involved contributing.

Reading through the code base, I'd anticipate I'd add a test that asserts a new counter is initilized and increments when there is an error here:

https://github.com/prometheus/prometheus/blob/2ca25d1c96424c896683e8cab65ba9666e8d3b66/tsdb/head.go#L1002-L1004

Thoughts? Any other pointers?

brian-brazil:
I'd usually not even write a unittest for a metric, you wouldn't for an application log.

alippai:
Shouldn't `/-/healthy` and `/-/ready` reflect these failures?

roidelapluie:
> Shouldn't `/-/healthy` and `/-/ready` reflect these failures?

This is discussed in https://github.com/prometheus/prometheus/issues/3807

alippai:
@roidelapluie Thank you

codesome:
Looks like this is fixed by https://github.com/prometheus/prometheus/pull/6647

PR #6647, opened 2020-01-16
Exports metric for WAL write errors

jpmcb:
Related to https://github.com/prometheus/prometheus/issues/6577

Includes a metric for WAL write errors

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -107,7 +107,6 @@ type headMetrics struct {
 	samplesAppended         prometheus.Counter
 	walTruncateDuration     prometheus.Summary
 	walCorruptionsTotal     prometheus.Counter
-	walWritesFailed         prometheus.Counter
 	headTruncateFail        prometheus.Counter
 	headTruncateTotal       prometheus.Counter
 	checkpointDeleteFail    prometheus.Counter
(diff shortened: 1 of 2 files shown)


tsdb/head.go
```diff
@@ -178,6 +179,10 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 		Name: "prometheus_tsdb_wal_corruptions_total",
 		Help: "Total number of WAL corruptions.",
 	})
+	m.walWritesFailed = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_tsdb_wal_writes_failed",
```
cstyan: counters should end with `_total`

jpmcb: Ah yes, thanks for that! Updated!!

brian-brazil:
:+1:

cstyan:
would be easier to move the metric to `wal.go`: https://github.com/prometheus/prometheus/blob/master/tsdb/wal/wal.go#L183-L190

jpmcb:
Thanks for the feedback! Reading through `wal.go`, looks like whenever the page gets flushed, we do a file write. That error block seemed like the best place for the metric, but let me know if I missed anything else!

cstyan:
I think the error should be tracked in `Log(...)`, that's the way external consumers can write to the log and that function can error for reasons other than just writing to disk.

jpmcb:
Oh I see - `head.go` and other outside consumers will call `wal.Log(rec)`. I've made the change, thanks so much for the help on this one and bringing me up to speed!

PR #6651, opened 2020-01-17
Don't sort postings if we only have one block.

brian-brazil:
This will make big queries that only touch the head faster,
though queries that touch both the head and a block will still
be the same speed. This probably won't help much with graphing
unless the range is under an hour, however it should make most
recording rules faster. 

PromQL benchmarks for histograms show only 2-3% improvement, but
they're only over 1k series. Looks like we'd be saving about 1s for a head-only
query that touches 1M series, so a 1M series histogram_quantile(rate(5m)) 
would go from around 12s to 11s.

### DIFF
--- a/tsdb/querier.go
+++ b/tsdb/querier.go
@@ -34,6 +34,9 @@ type Querier interface {
 	// Select returns a set of series that matches the given label matchers.
 	Select(...*labels.Matcher) (SeriesSet, error)
 
+	// SelectSorted returns a sorted set of series that matches the given label matcher.
+	SelectSorted(...*labels.Matcher) (SeriesSet, error)
+
 	// LabelValues returns all potential values for a label name.
 	// It is not safe to use the strings beyond the lifefime of the querier.
 	LabelValues(string) ([]string, error)
@@ -106,14 +109,21 @@ func (q *querier) lvals(qs []Querier, n string) ([]string, error) {
 }
 
 func (q *querier) Select(ms ...*labels.Matcher) (SeriesSet, error) {
+	if len(q.blocks) != 1 {
+		return q.SelectSorted(ms...)
+	}
+	// Sorting Head series is slow, and unneeded when only the
+	// Head is being queried. Sorting blocks is a noop.
+	return q.blocks[0].Select(ms...)
+}
+
+func (q *querier) SelectSorted(ms ...*labels.Matcher) (SeriesSet, error) {
 	if len(q.blocks) == 0 {
 		return EmptySeriesSet(), nil
 	}
 	ss := make([]SeriesSet, len(q.blocks))
-	var s SeriesSet
-	var err error
 	for i, b := range q.blocks {
-		s, err = b.Select(ms...)
+		s, err := b.SelectSorted(ms...)
 		if err != nil {
 			return nil, err
 		}
@@ -142,12 +152,16 @@ func (q *verticalQuerier) Select(ms ...*labels.Matcher) (SeriesSet, error) {
 	return q.sel(q.blocks, ms)
 }
 
+func (q *verticalQuerier) SelectSorted(ms ...*labels.Matcher) (SeriesSet, error) {
+	return q.sel(q.blocks, ms)
+}
+
 func (q *verticalQuerier) sel(qs []Querier, ms []*labels.Matcher) (SeriesSet, error) {
 	if len(qs) == 0 {
 		return EmptySeriesSet(), nil
 	}
 	if len(qs) == 1 {
-		return qs[0].Select(ms...)
+		return qs[0].SelectSorted(ms...)
 	}
 	l := len(qs) / 2
 
@@ -201,6 +215,24 @@ type blockQuerier struct {
 
 func (q *blockQuerier) Select(ms ...*labels.Matcher) (SeriesSet, error) {
 	base, err := LookupChunkSeries(q.index, q.tombstones, ms...)
+	if err != nil {
+		return nil, err
+	}
+	return &blockSeriesSet{
+		set: &populatedChunkSeries{
+			set:    base,
+			chunks: q.chunks,
+			mint:   q.mint,
+			maxt:   q.maxt,
+		},
+
+		mint: q.mint,
+		maxt: q.maxt,
+	}, nil
+}
+
+func (q *blockQuerier) SelectSorted(ms ...*labels.Matcher) (SeriesSet, error) {
+	base, err := LookupChunkSeriesSorted(q.index, q.tombstones, ms...)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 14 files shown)


tsdb/querier.go
```diff
@@ -106,14 +109,29 @@ func (q *querier) lvals(qs []Querier, n string) ([]string, error) {
...
+func (q *querier) SelectSorted(ms ...*labels.Matcher) (SeriesSet, error) {
 	if len(q.blocks) == 0 {
 		return EmptySeriesSet(), nil
 	}
 	ss := make([]SeriesSet, len(q.blocks))
 	var s SeriesSet
 	var err error
 	for i, b := range q.blocks {
-		s, err = b.Select(ms...)
+		if len(q.blocks) == 1 {
```
roidelapluie: When we call this function and q.blocs is 1 ?

tsdb/querier.go
```diff
@@ -217,6 +239,24 @@ func (q *blockQuerier) Select(ms ...*labels.Matcher) (SeriesSet, error) {
 	}, nil
 }
 
+func (q *blockQuerier) SelectSorted(ms ...*labels.Matcher) (SeriesSet, error) {
```
roidelapluie: What is the difference between this and Select ?

roidelapluie: nevermind I have seent he difference

tsdb/querier.go
```diff
@@ -106,14 +109,29 @@ func (q *querier) lvals(qs []Querier, n string) ([]string, error) {
...
 func (q *querier) Select(ms ...*labels.Matcher) (SeriesSet, error) {
+	if len(q.blocks) != 1 {
+		return q.SelectSorted(ms...)
+	}
+	// Sorting Head series is slow, and unneeded when only the
+	// Head is being queried. Sorting blocks is a noop.
+	return q.blocks[0].Select(ms...)
+}
+
+func (q *querier) SelectSorted(ms ...*labels.Matcher) (SeriesSet, error) {
```
roidelapluie: do we need this publicly?

roidelapluie:
This optimization look like implementation details that we could not expose publicly.

tsdb/querier.go
brian-brazil: It's the Query interface that we're using everywhere, having shadow internal interfaces that are slightly different tends to end up being hard to maintain

tsdb/querier.go
brian-brazil: Ah, that's from when I refactored this a bit. Let me fix that.

roidelapluie: if the function is exposed publicly that might still make sense

brian-brazil:
> This optimization look like implementation details that we could not expose publicly.

That's the nature of some optimisations, they require removing a level of abstraction.

tsdb/querier.go
```diff
@@ -106,14 +109,29 @@ func (q *querier) lvals(qs []Querier, n string) ([]string, error) {
...
 	if len(q.blocks) == 0 {
 		return EmptySeriesSet(), nil
 	}
```
roidelapluie: I feek like we could avoid this extra function if we pu the if len ==  1 there and return q.blocks[0].Select(ms...)

tsdb/querier.go
brian-brazil: If you call the Sorted version of the function, the result should still be sorted.

tsdb/querier.go
brian-brazil: Not quite, as the sorting has to happen another function down again.

tsdb/querier.go
roidelapluie: As we just have one block arent the postings sorted already?

tsdb/querier.go
roidelapluie: As we just have one block arent the postings sorted already?

roidelapluie: mmh neverming in the head block it is probaly not sorted

brian-brazil: Exactly

roidelapluie:
Merge conflict + waiting for another tsdb contributor to review this

tsdb/querier.go
bwplotka: Yes please (:

brian-brazil:
> This is because all the GO users of the TSDB who imports code will be surprised

We never guaranteed any particular sorting, and they'd lose the performance benefits.

bwplotka:
Hm, yea I thought it was guaranteed somehow and e.g promql rely on this to produce certain order of results, but looks like we did not guaranteed that indeed, so LGTM (:

bwplotka:
What about remote read did we not guarantee the order of returned series? And it bases on Select. @brian-brazil

brian-brazil:
We have grouping for remote read, but not ordering.

brian-brazil:
Hmm, that's not part of the protocol (and not really sane to have as part of it as the sort order is entirely internal to the tsdb) so that sounds like we need to make the remote read code more robust.

bwplotka:
Let's figure out tomorrow how we can solve this (:

brian-brazil:
@bwplotka updated as discussed.

brian-brazil:
Added.

tsdb/querier_bench_test.go
```diff
@@ -136,7 +136,11 @@ func BenchmarkQuerierSelect(b *testing.B) {
 	}
 	testutil.Ok(b, app.Commit())
 
-	bench := func(b *testing.B, br BlockReader) {
+	emptyHead, err := NewHead(nil, nil, nil, 1000)
+	testutil.Ok(b, err)
+	defer emptyHead.Close()
```
codesome: I don't see this being used anywhere

web/api/v1/api.go
```diff
@@ -1121,7 +1121,12 @@ func (api *API) remoteRead(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 		for i, query := range req.Queries {
-			err := api.remoteReadQuery(ctx, query, externalLabels, func(set storage.SeriesSet) error {
+			err := api.remoteReadQuery(ctx, query, externalLabels, func(querier storage.Querier, selectParams *storage.SelectParams, filteredMatchers []*labels.Matcher) error {
+				// The streaming API provides sorted series.
+				set, _, err := querier.SelectSorted(selectParams, filteredMatchers...)
```
codesome: Why is it only the streaming API that has it sorted?

brian-brazil: Thanos needs it. I don't see a reason to slow down the older endpoint, nor to incorrectly make it seem that it must be sorted.

tsdb/querier_bench_test.go
brian-brazil: Fixed

web/api/v1/api.go
codesome: :+1:

bwplotka: Yes, we rely on this heavily. Thanks!

PR #6653, opened 2020-01-17
Add windows tests

roidelapluie:


.circleci/config.yml
```diff
@@ -55,8 +55,14 @@ jobs:
     working_directory: /go/src/github.com/prometheus/prometheus
     steps:
     - checkout
-    # TSDB is where the most risk is Windows wise, so only test there for now.
-    - run: go test ./tsdb/...
+    - run:
+        shell: bash
+        command: |
+          (cd web/ui && GOOS= GOARCH= go generate -mod=vendor)
+          go test -mod=vendor -test.v `go list ./...|grep -v github.com/prometheus/prometheus/discovery | grep -xv github.com/prometheus/prometheus/config | grep -xv github.com/prometheus/prometheus/web`
```
brian-brazil: You can use -e with grep to have multiple patterns

roidelapluie:
Rebased on top of #6658

cmd/prometheus/main_unix_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"os/exec"
+	"testing"
+	"time"
+)
+
+func stopProcess(p *os.Process) {
+	p.Signal(os.Interrupt)
+}
+
+// As soon as prometheus starts responding to http request should be able to accept Interrupt signals for a graceful shutdown.
```
brian-brazil: requests it should

cmd/prometheus/main_unix_test.go
```diff
@@ -0,0 +1,76 @@
...
+	go func() {
+		done <- prom.Wait()
+	}()
+
+	var startedOk bool
+	var stoppedErr error
+
+Loop:
+	for x := 0; x < 10; x++ {
+		// error=nil means prometheus has started so can send the interrupt signal and wait for the grace shutdown.
```
brian-brazil: so we can

graceful

cmd/prometheus/main_windows_test.go
```diff
@@ -0,0 +1,24 @@
...
+// +build windows
+
+package main
+
+import (
+	"os"
+)
+
+func stopProcess(p *os.Process) {
+	p.Kill()
```
brian-brazil: Why do we not stop all OSes the same way?

discovery/manager_windows_test.go
```diff
@@ -0,0 +1,20 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// +build windows
+
+package discovery
+
+import "time"
+
+var timescale = 10 * time.Millisecond
```
brian-brazil: Please don't indirect basic constants like this, and if one OS needs more time give it to all of them.

util/stats/stats_unix_test.go
```diff
@@ -0,0 +1,55 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// +build !windows
+
+package stats
+
+import "testing"
+
+// TestTimerGroup does not work on Windows since its timer is not accurate
```
brian-brazil: That doesn't sound right, Windows has high resolution timers too. This sounds like a bug in the timer code.

cmd/prometheus/main_windows_test.go
roidelapluie: It is preferrable to stop with Interrupt as it gives Prometheus the ability to cleanly shutdown.

However Interrupt is not supported in Windows.

util/stats/stats_unix_test.go
roidelapluie: Time.Time() returns the same if call consecutively on Win.

Added  Sleep()

cmd/prometheus/main_windows_test.go
brian-brazil: KILL is supported on Unix.

util/stats/stats_unix_test.go
brian-brazil: What version of Windows? It really sounds like we're using the wrong timing API here.

cmd/prometheus/main_windows_test.go
roidelapluie: does it trigger a graceful shutdown of TSDB?

brian-brazil: No, but you have that exact problem on Windows anyway.

util/stats/timer.go
```diff
@@ -95,6 +95,7 @@ func (t Timers) Swap(i, j int) {
 }
 
 func (s byCreationTimeSorter) Less(i, j int) bool {
+	fmt.Printf("i: %v / j: %v\n", s.Timers[i].created, s.Timers[j])
```
brian-brazil: Debug code?

roidelapluie: yes to debug the time.Time issue

roidelapluie: here is a version that does not rely on time.Time

brian-brazil:
Thanks!

PR #6658, opened 2020-01-18
Removed compilation step inside main_test

roidelapluie:
Inspired by https://github.com/prometheus/prometheus/pull/6347 and
https://github.com/prometheus/prometheus/pull/6347#issuecomment-570151979

### DIFF
--- a/cmd/prometheus/main_test.go
+++ b/cmd/prometheus/main_test.go
@@ -15,7 +15,6 @@ package main
 
 import (
 	"context"
-	"flag"
 	"fmt"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


cmd/prometheus/main_test.go
```diff
@@ -31,35 +30,23 @@ import (
...
 var promConfig = filepath.Join("..", "..", "documentation", "examples", "prometheus.yml")
 var promData = filepath.Join(os.TempDir(), "data")
 
 func TestMain(m *testing.M) {
-	flag.Parse()
-	if testing.Short() {
-		os.Exit(m.Run())
+	for i, arg := range os.Args {
+		if arg == "-test.main" {
+			os.Args = append(os.Args[:i], os.Args[i+1:]...)
```
roidelapluie: Note that I have check that it does not panic even if `-test.main` is last:
https://play.golang.org/p/IT0mKhfStGh

brian-brazil:
Thanks!

PR #6661, opened 2020-01-18
web: use chained WithInstrumentation

roidelapluie:
Requires an update of github.com/prometheus/common


brian-brazil:
Thanks!

PR #6662, opened 2020-01-18
go.mod: update github.com/prometheus dependencies

roidelapluie:
Except common, which is handled in #6661 


roidelapluie:
go get -u github.com/prometheus/client_golang
go get -u github.com/prometheus/alertmanager
go get -u github.com/prometheus/client_model
go mod tidy
go mod vendor

brian-brazil:
We generally do this just before a release. Is there a particular one you're looking for?

roidelapluie:
I was doing common for #6661, we are just one week off release, and it was not done for 2.15.  While I was as it, I did it.

brian-brazil:
@cstyan this is yours then.

cstyan:
I missed this, do you want me to put this into a 2.16.0-rc.1?

beorn7:
I'd say let's only do the update and a second RC if we have reasons to believe an update is necessary.

For the routine dependency update alone, I wouldn't do a separace RC. Should the rc.0 have a bug and we have to do an rc.1 anyway, then I would also include the dependency update.

PR #6660, opened 2020-01-18
textparse: reject duplicate labels

roidelapluie:
Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>

PR #6664, opened 2020-01-19
tsdb: error on ingesting series with duplicate labels

roidelapluie:

xx

Alternative to #6660

### DIFF
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -200,6 +200,20 @@ func (ls Labels) Has(name string) bool {
 	return false
 }
 
+// HasDuplicateLabels returns whether ls has duplicate label names.
+// It assumes that the labelset is sorted.
+func (ls Labels) HasDuplicateLabelNames() (string, bool) {
+	for i, l := range ls {
+		if i == 0 {
+			continue
+		}
+		if l.Name == ls[i-1].Name {
+			return l.Name, true
+		}
+	}
+	return "", false
+}
+
 // WithoutEmpty returns the labelset without empty labels.
 // May return the same labelset.
 func (ls Labels) WithoutEmpty() Labels {
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -53,6 +53,10 @@ var (
 	// writable time range.
 	ErrOutOfBounds = errors.New("out of bounds")
 
+	// ErrInvalidSample is returned if an appended sample is not valid and can't
+	// be ingested.
+	ErrInvalidSample = errors.New("invalid sample")
+
 	// emptyTombstoneReader is a no-op Tombstone Reader.
 	// This is used by head to satisfy the Tombstones() function call.
 	emptyTombstoneReader = tombstones.NewMemTombstones()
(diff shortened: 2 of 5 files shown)


pkg/labels/labels.go
```diff
@@ -200,6 +200,18 @@ func (ls Labels) Has(name string) bool {
 	return false
 }
 
+// HasDuplicateLabels returns whether ls has duplicate label names.
+func (ls Labels) HasDuplicateLabelNames() bool {
```
roidelapluie: We can chose here if we want more CPU (iter twice on the labels)
or more RAM (found:=make([]string,0,len(ls)))

brian-brazil: Labels are sorted, so you can do this in O(n) time.

tsdb/head_test.go
```diff
@@ -1290,3 +1290,21 @@ func TestNewWalSegmentOnTruncate(t *testing.T) {
...
+		testutil.Ok(t, os.RemoveAll(dir))
+	}()
+	wlog, err := wal.NewSize(nil, nil, dir, 32768, false)
+	testutil.Ok(t, err)
+
+	h, err := NewHead(nil, nil, wlog, 1000)
+	testutil.Ok(t, err)
+	defer h.Close()
+	app := h.Appender()
+	_, err = app.Add(labels.Labels{{Name: "a", Value: "c"}, {Name: "a", Value: "b"}}, 0, 0)
```
brian-brazil: Test for a=b,a=b too

roidelapluie: Should it error?

mback2k:
> Alternative to #6660

Wouldn't it make prometheus more robust if path PRs were merged?

tsdb/head_test.go
brian-brazil: Yes

roidelapluie:
> > Alternative to #6660
> 
> Wouldn't it make prometheus more robust if path PRs were merged?

No, but it would make it slower.

roidelapluie:
I guess that @brian-brazil already knows it but after reading the code this looks like the good place to add this.

The Add() function will only be called once per metric per target. As metrics should not change a lot between scrapes, it will cause an overhead on the first scrape, but won't re-check for duplicate labels on next scrapes.

Adding that in the parser like in #6660 would mean that we make the work for every metric in every scrape.

roidelapluie:
I guess that @brian-brazil already knows it but after reading the code this looks like the good place to add this.

The Add() function will only be called once per metric per target. As metrics should not change a lot between scrapes, it will cause an overhead on the first scrape, but won't re-check for duplicate labels on next scrapes.

Adding that in the parser like in #6660 would mean that we make the work for every metric in every scrape.

tsdb/head.go
```diff
@@ -921,6 +925,10 @@ func (a *headAppender) Add(lset labels.Labels, t int64, v float64) (uint64, erro
 	// Ensure no empty labels have gotten through.
 	lset = lset.WithoutEmpty()
 
+	if l, dup := lset.HasDuplicateLabelNames(); dup {
+		return 0, errors.Wrap(ErrInvalidSample, fmt.Sprintf(`label "%s" is not unique`, l))
```
brian-brazil: label name

brian-brazil:
We should add a test to confirm that this causes such a scrape to fail.

roidelapluie:
Test added.

brian-brazil:
Thanks!

PR #6666, opened 2020-01-19
Add a note about updating deps

roidelapluie:

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -52,6 +52,29 @@ If a bug fix got accidentally merged into master after non-bug-fix changes in ma
 
 Maintaining the release branches for older minor releases happens on a best effort basis.
 
+### Updating dependencies
+
+A few days before a major or minor release, consider updating the dependencies:
+
+```
+export GO111MODULE=on
+go get -u ./...
+go mod tidy
+go mod vendor
+git add go.mod go.sum vendor
+```
+
+Then create a pull request against the master branch.
+
+Note that after a dependency update, you should look out for any weirdness that
+might have happened. Such weirdnesses include but are not limited to: flaky
+tests, differences in resource usage, panic.
+
+In case of doubt or issues that can't be solved in a reasonable amount of time,
+you can skip the dependency update or only update select dependencies. In such a
+case, you have to create an issue or pull request in the GitHub project for
+later follow-up.
+
 ### Prepare your release
 
 For a patch release, work in the branch of the minor release you want to patch.
@@ -103,4 +126,3 @@ The following changes to the above procedures apply:
 * Tick the _This is a pre-release_ box when drafting the release in the Github UI.
 * Still update `CHANGELOG.md`, but when you cut the final release later, merge all the changes from the pre-releases into the one final update.
 * Run the benchmark for 3 days using the `/benchmark x.y.z` command, `x.y.z` being the latest stable patch release of the previous minor release series.
-


RELEASE.md
```diff
@@ -52,6 +52,19 @@ If a bug fix got accidentally merged into master after non-bug-fix changes in ma
 
 Maintaining the release branches for older minor releases happens on a best effort basis.
 
+### Updating the dependencies
+
+A few days before a major or minor release, update the dependencies:
+
+```
+go get -u ./...
```
beorn7: Just in case, let's first do a `export GO111MODULE=on`?

RELEASE.md
```diff
@@ -52,6 +52,19 @@ If a bug fix got accidentally merged into master after non-bug-fix changes in ma
 
 Maintaining the release branches for older minor releases happens on a best effort basis.
 
+### Updating the dependencies
+
+A few days before a major or minor release, update the dependencies:
```
beorn7: Perhaps use a more cautious language? E.g. "A few days before a major or minor release, consider updating the dependencies:" And then later add a warning to look out for any weirdness that might have happened. We should avoid the impression that a dependency update is always good and always "just works".

roidelapluie: Thank you for the review. I have added such a text now.

brian-brazil: Updating dependencies isn't always safe, but I think we should always be doing it before a release. Otherwise when it eventually does happen it'll be harder to deal with any resulting issues.

roidelapluie: I would propose to rephrase that if some dependencies can't be updated with a reasonable amount of efforts, an issue must be created for followup?

brian-brazil: That's reasonable.

brian-brazil:
:+1:

RELEASE.md
```diff
@@ -52,6 +52,29 @@ If a bug fix got accidentally merged into master after non-bug-fix changes in ma
...
+git add go.mod go.sum vendor
+```
+
+Then create a pull request against the master branch.
+
+Note that after an update of the dependencies, you should look out for any
+weirdness that might have happened. Such weirdnesses include but are not limited
+to: flaky tests, differences in resource usage, panic.
+
+In case of doubt, or issues that can't be solved in a reasonable amount of time,
```
juliusv:
```suggestion
In case of doubt or issues that can't be solved in a reasonable amount of time,
```

RELEASE.md
```diff
@@ -52,6 +52,29 @@ If a bug fix got accidentally merged into master after non-bug-fix changes in ma
...
+you can skip this dependency update or only update select dependencies. In such
+a case, you have to create an issue or pull request in the github project for
```
juliusv:
```suggestion
a case, you have to create an issue or pull request in the GitHub project for
```

juliusv:
:+1: besides nits

RELEASE.md
beorn7: @roidelapluie I like @juliusv 's suggestion. Could you apply it?

roidelapluie: Oops I missed it. Applied.

beorn7:
Sorry for more nitpicking, but the newly picked headline "dependencies update" sounds weird to me. I'd say "dependency update" (but I'm not a native speaker either). Even better "update of dependencies" (as before, just without the article).

brian-brazil:
Updating Dependencies is what I'd personally do, and is in line with other headings.

RELEASE.md
```diff
@@ -52,6 +52,29 @@ If a bug fix got accidentally merged into master after non-bug-fix changes in ma
...
+```
+
+Then create a pull request against the master branch.
+
+Note that after an update of the dependencies, you should look out for any
+weirdness that might have happened. Such weirdnesses include but are not limited
+to: flaky tests, differences in resource usage, panic.
+
+In case of doubt or issues that can't be solved in a reasonable amount of time,
+you can skip the dependencies update or only update select dependencies. In such
```
beorn7: Super sorry, but here is another "dependencies update". I'd say "dependency update" here, but a native speaker might have an even better idea.

brian-brazil: Yeah, dependency update is what I'd say here. It's a car sale, not a cars sale.

RELEASE.md
```diff
@@ -52,6 +52,29 @@ If a bug fix got accidentally merged into master after non-bug-fix changes in ma
...
+export GO111MODULE=on
+go get -u ./...
+go mod tidy
+go mod vendor
+git add go.mod go.sum vendor
+```
+
+Then create a pull request against the master branch.
+
+Note that after an dependency update, you should look out for any weirdness that
```
beorn7: an → a

beorn7:
But now the DCO is missing.

@roidelapluie perhaps you could squash the many incremental commits into one and then also fix the DCO? (I could do that for you, but I guess that would defeat the purpose of the DCO.)

roidelapluie:
done :)

beorn7:
Many thanks once more. I'll merge on green.

PR #6667, opened 2020-01-20
Expose DB.compact() method

codesome:
This change won't affect Prometheus in any way but rather help the projects that are vendoring TSDB. As the compact method is guarded by a mutex, it would also be safe to call it concurrently externally.

Any objections to this?

/cc @brian-brazil @krasi-georgiev @bwplotka @pracucci


roidelapluie:
As DisableCompaction is public, it makes sense to make Compact public.

krasi-georgiev:
LGTM, it might also simplify some of the tests for the compaction.

PR #6670, opened 2020-01-20
Pass over scrape cache to the next scrape

### LINKED ISSUE #3112: Reloading Prometheus 2.0 has significant memory spikes

roidelapluie:
Replaces and closes #3861
Closes #3112

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -315,7 +315,7 @@ func (sp *scrapePool) reload(cfg *config.ScrapeConfig) error {
 	for fp, oldLoop := range sp.loops {
 		var cache *scrapeCache
 		if oc := oldLoop.getCache(); reuseCache && oc != nil {
-			cache = oldLoop.getCache()
+			cache = oc
 		} else {
 			cache = newScrapeCache()
 		}
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -23,6 +23,7 @@ import (
 	"io/ioutil"
 	"math"
 	"net/http"
+	"reflect"
 	"sync"
 	"time"
 	"unsafe"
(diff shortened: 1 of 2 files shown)


brian-brazil:
Is this safe? If relabelling rules change for example, the cache is invalid.

roidelapluie:
Yes I am working on this

roidelapluie:
Attempt number 2

roidelapluie:
This is now reviewable.

scrape/scrape.go
```diff
@@ -1312,3 +1331,25 @@ func (sl *scrapeLoop) addReportSample(app storage.Appender, s string, t int64, v
 		return err
 	}
 }
+
+// zeroConfig returns a new scrape config that only contains configuration items
+// that alter metrics.
+func zeroConfig(c *config.ScrapeConfig) *config.ScrapeConfig {
+	z := *c
+	// We zero out the fields that for sure don't affect scrape.
+	z.HonorTimestamps = false
```
brian-brazil: This affects what we ingest.

scrape/scrape.go
```diff
@@ -1312,3 +1331,25 @@ func (sl *scrapeLoop) addReportSample(app storage.Appender, s string, t int64, v
...
+	z.ScrapeInterval = 0
+	z.ScrapeTimeout = 0
+	z.SampleLimit = 0
+	z.HTTPClientConfig = config_util.HTTPClientConfig{}
+	return &z
+}
+
+// reusableCache compares two scrape config and tells wheter the cache is stoll
```
brian-brazil: still

brian-brazil:
Thanks!

PR #6673, opened 2020-01-21
Make remote client `Store` use passed context

anandsinghkunwar:
* Remote store client's `Store` API currently doesn't use passed
context, but instead just constructs a new `context.Background()`


anandsinghkunwar:
@brian-brazil Can you merge this?

cstyan:
:+1: yes thanks for fixing this! the ctx being passed in is indeed the hard shutdown

PR #6675, opened 2020-01-21
scrape: Add metrics to track bytes and entries in the metadata cache

gotjosh:
At the moment, we have low visibility in the usage and entries of the scrape metadata cache. With the inclusion of these metrics, we aim to understand how do the number of entries and cache size correlate in a Prometheus install given we have no limits around metadata. 

### DIFF
--- a/scrape/manager.go
+++ b/scrape/manager.go
@@ -27,11 +27,78 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/discovery/targetgroup"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/storage"
 )
+
+var targetMetadataCache = newMetadataMetricsCollector()
+
+// MetadataMetricsCollector is a Custom Collector for the metadata cache metrics.
+type MetadataMetricsCollector struct {
+	CacheEntries *prometheus.Desc
+	CacheBytes   *prometheus.Desc
+
+	scrapeManager *Manager
+}
+
+func newMetadataMetricsCollector() *MetadataMetricsCollector {
+	return &MetadataMetricsCollector{
+		CacheEntries: prometheus.NewDesc(
+			"prometheus_target_metadata_cache_entries",
+			"Total number of metric metadata entries in the cache",
+			[]string{"scrape_job"},
+			nil,
+		),
+		CacheBytes: prometheus.NewDesc(
+			"prometheus_target_metadata_cache_bytes",
+			"The number of bytes that are currently used for storing metric metadata in the cache",
+			[]string{"scrape_job"},
+			nil,
+		),
+	}
+}
+
+func (mc *MetadataMetricsCollector) registerManager(m *Manager) {
+	mc.scrapeManager = m
+}
+
+// Describe sends the metrics descriptions to the channel.
+func (mc *MetadataMetricsCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- mc.CacheEntries
+	ch <- mc.CacheBytes
+}
+
+// Collect creates and sends the metrics for the metadata cache.
+func (mc *MetadataMetricsCollector) Collect(ch chan<- prometheus.Metric) {
+	if mc.scrapeManager == nil {
+		return
+	}
+
+	for tset, targets := range mc.scrapeManager.TargetsActive() {
+		var size, length int
+		for _, t := range targets {
+			size += t.MetadataSize()
+			length += t.MetadataLength()
+		}
+
+		ch <- prometheus.MustNewConstMetric(
+			mc.CacheEntries,
+			prometheus.GaugeValue,
+			float64(length),
+			tset,
+		)
+
+		ch <- prometheus.MustNewConstMetric(
+			mc.CacheBytes,
+			prometheus.GaugeValue,
+			float64(size),
+			tset,
+		)
+	}
+}
 
 // Appendable returns an Appender.
 type Appendable interface {
--- a/scrape/manager.go
+++ b/scrape/manager.go
@@ -43,7 +43,7 @@ func NewManager(logger log.Logger, app Appendable) *Manager {
 	if logger == nil {
 		logger = log.NewNopLogger()
 	}
-	return &Manager{
+	m := &Manager{
 		append:        app,
 		logger:        logger,
 		scrapeConfigs: make(map[string]*config.ScrapeConfig),
@@ -51,6 +51,8 @@ func NewManager(logger log.Logger, app Appendable) *Manager {
 		graceShut:     make(chan struct{}),
 		triggerReload: make(chan struct{}, 1),
 	}
+
+	return m
 }
 
 // Manager maintains a set of scrape pools and manages start/stop cycles
@@ -134,6 +136,7 @@ func (m *Manager) reload() {
 		}(m.scrapePools[setName], groups)
 
 	}
+	targetMetadataCache.registerManager(m)
 	m.mtxScrape.Unlock()
 	wg.Wait()
 }
--- a/scrape/manager.go
+++ b/scrape/manager.go
@@ -51,6 +51,7 @@ func NewManager(logger log.Logger, app Appendable) *Manager {
 		graceShut:     make(chan struct{}),
 		triggerReload: make(chan struct{}, 1),
 	}
+	targetMetadataCache.registerManager(m)
 
 	return m
 }
@@ -136,7 +137,6 @@ func (m *Manager) reload() {
 		}(m.scrapePools[setName], groups)
 
 	}
-	targetMetadataCache.registerManager(m)
 	m.mtxScrape.Unlock()
 	wg.Wait()
 }
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -133,21 +133,74 @@ var (
 			Help: "How many times a scrape cache was flushed due to getting big while scrapes are failing.",
 		},
 	)
-	targetMetadataCacheBytes = prometheus.NewGaugeVec(
-		prometheus.GaugeOpts{
-			Name: "prometheus_target_metadata_cache_bytes",
-			Help: "The number of bytes that are currently used for storing metric metadata in the cache",
-		},
-		[]string{"scrape_job"},
-	)
-	targetMetadataCacheEntries = prometheus.NewGaugeVec(
-		prometheus.GaugeOpts{
-			Name: "prometheus_target_metadata_cache_entries",
-			Help: "Total number of metric metadata entries in the cache",
-		},
-		[]string{"scrape_job"},
-	)
+	targetMetadataCache = newMetadataMetricsCollector()
 )
+
+// MetadataMetricsCollector is a Custom Collector for the metadata cache metrics.
+type MetadataMetricsCollector struct {
+	CacheEntries *prometheus.Desc
+	CacheBytes   *prometheus.Desc
+
+	scrapeCaches map[*Target]*scrapeCache
+
+	mtx sync.RWMutex
+}
+
+func newMetadataMetricsCollector() *MetadataMetricsCollector {
+	return &MetadataMetricsCollector{
+		CacheEntries: prometheus.NewDesc(
+			"prometheus_target_metadata_cache_entries",
+			"Total number of metric metadata entries in the cache",
+			nil,
+			nil,
+		),
+		CacheBytes: prometheus.NewDesc(
+			"prometheus_target_metadata_cache_bytes",
+			"The number of bytes that are currently used for storing metric metadata in the cache",
+			nil,
+			nil,
+		),
+		scrapeCaches: map[*Target]*scrapeCache{},
+	}
+}
+
+func (mc *MetadataMetricsCollector) addScrapeCache(t *Target, c *scrapeCache) {
+	mc.mtx.Lock()
+	mc.scrapeCaches[t] = c
+	mc.mtx.Unlock()
+}
+
+func (mc *MetadataMetricsCollector) removeScrapeCache(t *Target) {
+	mc.mtx.Lock()
+	delete(mc.scrapeCaches, t)
+	mc.mtx.Unlock()
+}
+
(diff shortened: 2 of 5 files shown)


scrape/scrape.go
```diff
@@ -685,12 +709,18 @@ func (c *scrapeCache) iterDone(flushCache bool) {
...
 		for m, e := range c.metadata {
 			// Keep metadata around for 10 scrapes after its metric disappeared.
 			if c.iter-e.lastIter > 10 {
 				delete(c.metadata, m)
+				continue
 			}
+			metaSize += len(m) + e.size()
 		}
+
+		targetMetadataCacheBytes.WithLabelValues(jobName).Set(float64(metaSize))
```
gotjosh: I've used `jobName` as the identifier of a target given we do something similar in this file - I tried using the target's labels, but that seems quite tricky because of relabelling. Happy to be guided on what would be best!

scrape/scrape.go
```diff
@@ -589,6 +606,8 @@ type cacheEntry struct {
 }
 
 type scrapeLoop struct {
+	jobName string
```
cstyan: I think I'd make the jobName a field on the scrapePool instead of on every scrapeLoop.

scrape/scrape.go
```diff
@@ -132,6 +132,20 @@ var (
 			Help: "How many times a scrape cache was flushed due to getting big while scrapes are failing.",
 		},
 	)
+	targetMetadataCacheBytes = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "prometheus_target_metadata_cache_bytes",
+			Help: "The number of bytes that are currently used for storing metric metadata of a target",
```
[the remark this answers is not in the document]
gotjosh: Sorry, not sure I follow your comment. Yes, the target information is represented as labels - are we saying that the usage of the word "target" in this context does not make sense because of how it is represented in-code? 

Would the same sentence without the last three words work?

scrape/scrape.go
gotjosh: Yeah, I tried that. The problem with it is that the `scrapeLoop` has no notion of the `scrapePool` and I need the name within the `scrapeCache` (with is part of the loop). Originally, I even tried passing the `jobName` to the `scrapeCache` but it feels the cache should have no notion of a "job".

scrape/scrape.go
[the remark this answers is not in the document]
gotjosh: Reworded, what do you think now?

scrape/scrape.go
gotjosh: sorry @brian-brazil, I'm not sure I follow. Wouldn't the target(s) all belong to the same job under the current grouping? 

Taking another pass at the code, I think the `job` does not uniquely identify the target. Per the [group struct](https://github.com/prometheus/prometheus/blob/master/discovery/targetgroup/targetgroup.go#L25-L27), it seems I should be using the address label?

brian-brazil: A scrape config usually has more than one target.

roidelapluie:
we probaly want to add that metric at the same time as `up`, `scrape_duration_seconds` metric etc

scrape/scrape.go
beorn7: But the scrapeCache is per target, isn't it?

We have to calculate the value differently / somewhere else, if we really want the total size.

roidelapluie: I would hope for simplicity that we add that together with `up` metric

gotjosh: Yep @roidelapluie, your comment makes perfect sense. I'll do that.

brian-brazil: There's quite a high bar to adding a new sample to every single scrape that happens, that adds storage cost for every single Prometheus usage.

This belongs on /metrics, but as it stands not as a scrape metric.

roidelapluie: mmh yeah indeed brian. However then it makes it like we will see it only per job?

brian-brazil: Yes, the same others on /metrics. We'd usually not even break something like this out by job.

roidelapluie: Indeed I am not sure I want 300 new metrics for this

roidelapluie:
Note that since #6670 there is a getCache() function on the loops.

gotjosh:
Thanks @roidelapluie, I managed to see that PR a couple of days ago and thought it would affect this work. I'll make sure to rebase asap.

roidelapluie:
It looks like getCache might ease your work a lot, as the scrapePool will then have access to the caches.

gotjosh:
I have taken the PR in another direction based on @brian-brazil's recommendation. With my current understanding of synthetic metrics (e.g. `up`, `scrape_duration`, etc.) I think this belongs as `/metrics`. Any feedback would be massively appreciated.

roidelapluie:
You could make it more simple by iterating over the loops and using get cache. then you don't need the extra field in the loop and you can retrieve the target in activeTargets if you need it.

scrape/scrape.go
```diff
@@ -133,8 +133,75 @@ var (
...
 	)
+	targetMetadataCache = newMetadataMetricsCollector()
 )
 
+// MetadataMetricsCollector is a Custom Collector for the metadata cache metrics.
+type MetadataMetricsCollector struct {
+	CacheEntries *prometheus.Desc
+	CacheBytes   *prometheus.Desc
+
+	scrapeCaches map[*Target]*scrapeCache
```
brian-brazil: Don't maintain an additional structure, that's going to be fragile. Instead walk what we already have.

gotjosh: As Julien mentioned above, the alternative would be walking over the loops to get the cache. Is that what you have in mind or is it something different? 

We'd still need to hold an additional structure unless I'm missing something.

brian-brazil: That's what I have in mind.

I don't see why you'd need anything beyond a reference to the target manager.

scrape/scrape.go
```diff
@@ -133,8 +133,79 @@ var (
...
+func (mc *MetadataMetricsCollector) Collect(ch chan<- prometheus.Metric) {
+	mc.mtx.Lock()
+	defer mc.mtx.Unlock()
+
+	if mc.scrapeManager == nil {
+		return
+	}
+
+	for _, p := range mc.scrapeManager.scrapePools {
+		var size, length int
```
roidelapluie:
```suggestion
		var size, length float64
```

scrape/manager.go
```diff
@@ -43,14 +43,16 @@ func NewManager(logger log.Logger, app Appendable) *Manager {
 	if logger == nil {
 		logger = log.NewNopLogger()
 	}
-	return &Manager{
+	m := &Manager{
```
roidelapluie:
```suggestion
	return &Manager{
```

scrape/manager.go
```diff
@@ -43,14 +43,16 @@ func NewManager(logger log.Logger, app Appendable) *Manager {
...
 		append:        app,
 		logger:        logger,
 		scrapeConfigs: make(map[string]*config.ScrapeConfig),
 		scrapePools:   make(map[string]*scrapePool),
 		graceShut:     make(chan struct{}),
 		triggerReload: make(chan struct{}, 1),
 	}
+
+	return m
```
roidelapluie:
```suggestion
```

scrape/scrape.go
```diff
@@ -149,6 +220,7 @@ func init() {
 	prometheus.MustRegister(targetScrapeSampleOutOfOrder)
 	prometheus.MustRegister(targetScrapeSampleOutOfBounds)
 	prometheus.MustRegister(targetScrapeCacheFlushForced)
+	prometheus.MustRegister(targetMetadataCache)
```
roidelapluie: Maybe refactor this to call MustRegister once with all the collectors

scrape/scrape.go
```diff
@@ -216,8 +288,7 @@ func newScrapePool(cfg *config.ScrapeConfig, app Appendable, jitterSeed uint64,
 		}
 		opts.target.SetMetadataStore(cache)
 
-		return newScrapeLoop(
-			ctx,
+		return newScrapeLoop(ctx,
```
roidelapluie:
```suggestion
		return newScrapeLoop(
		        ctx,
```

scrape/scrape.go
beorn7: I think `int` is the appropriate choice here. Both `MetadataSize` and `MetadataLen` return `int`, and their sum cannot be larger than the heap size, so the platform dependent `int` size is appropriately sized to not overflow. Conversion to float64 (with precision issues, at least on a 64bit platform) should happen in the end.

scrape/scrape.go
beorn7: Indeed. This code has been written before the variadic `MustRegister` was created.

scrape/scrape.go
beorn7: Yes, that's indeed the Prometheus style.

scrape/scrape.go
gotjosh: done!

gotjosh:
Hopefully, this works @roidelapluie !

scrape/scrape.go
```diff
@@ -133,22 +133,96 @@ var (
...
+
+// Describe sends the metrics descriptions to the channel.
+func (mc *MetadataMetricsCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- mc.CacheEntries
+	ch <- mc.CacheBytes
+}
+
+// Collect creates and sends the metrics for the metadata cache.
+func (mc *MetadataMetricsCollector) Collect(ch chan<- prometheus.Metric) {
+	mc.mtx.Lock()
```
brian-brazil: Why do you do locking here?

scrape/scrape.go
```diff
@@ -133,22 +133,96 @@ var (
...
 	)
+	targetMetadataCache = newMetadataMetricsCollector()
 )
 
+// MetadataMetricsCollector is a Custom Collector for the metadata cache metrics.
+type MetadataMetricsCollector struct {
+	CacheEntries *prometheus.Desc
+	CacheBytes   *prometheus.Desc
+
+	scrapeManager *Manager
```
brian-brazil: Manager-related code goes in manager.go

scrape/scrape.go
```diff
@@ -133,22 +133,96 @@ var (
...
+// Collect creates and sends the metrics for the metadata cache.
+func (mc *MetadataMetricsCollector) Collect(ch chan<- prometheus.Metric) {
+	mc.mtx.Lock()
+	defer mc.mtx.Unlock()
+
+	if mc.scrapeManager == nil {
+		return
+	}
+
+	for _, p := range mc.scrapeManager.scrapePools {
```
brian-brazil: This needs locking. It'd be better to use TargetsActive.

scrape/scrape.go
gotjosh: The way I see it manager is a dependency, the metrics are coming out of the `scrapePool` / `scrapeCache` so I decided to put it closer to those. 

Happy to move it, if you still think this belongs in the manager.

scrape/scrape.go
gotjosh: I was hoping to avoid using `TargetsActive` as I don't need / can't use the serialization into a `Target`.  Furthermore, `Target` interacts with the `scrapeCache` for metadata via an interface which would also require modification.

On second thoughts, I think it's better to use the manager directly for this information. I can mimic most of what `TargetActive` is doing and thus limit the amount of scrape locking. 

Let me give that a try.

scrape/scrape.go
```diff
@@ -133,22 +133,94 @@ var (
...
+}
+
+// Collect creates and sends the metrics for the metadata cache.
+func (mc *MetadataMetricsCollector) Collect(ch chan<- prometheus.Metric) {
+	if mc.scrapeManager == nil {
+		return
+	}
+
+	for tset, targets := range mc.scrapeManager.TargetsActive() {
+		p, ok := mc.scrapeManager.scrapePools[tset]
```
gotjosh: Are we OK with this? Typically, we lock scraping when accessing the scrape pool and I would like to avoid that - If the pool is gone, are we ok with not reporting and assume the target set is gone?

brian-brazil: Locking around scrapePools is not optional.

scrape/scrape.go
brian-brazil: It belongs in the manager imho

scrape/scrape.go
gotjosh: Managed to avoid that, I completely missed that we use the `job_name` as the `target_set` identifier within the pool.

scrape/scrape.go
gotjosh: moved it.

brian-brazil:
Thanks!

gotjosh:
Thanks @brian-brazil for all your help.

juliusv:
Belated hello :) I wonder if it's too late to rename the `scrape_job` label to `scrape_pool` or something like that.

The background: I noticed that there is a frequent notion out there with users that Prometheus has "scrape jobs", due to the fact that the `scrape_config` section in the config file has a `job_name` field. But that field is *not* the name of the scrape config, it refers to the default name of the job (ex-Google lingo for "service") that the scraped instances themselves belong to, see https://prometheus.io/docs/concepts/jobs_instances/. From the scraping side, there are only scrape configs and scrape pools (which are basically the live instantiation of a scrape config), and jobs are an external concept. I'm a bit worried about officially muddying the waters more here terminology-wise, with this label name.

beorn7:
I'm all for less overloading (and generally less usage) of the `job` term.

However, the `scrape_job` label isn't just for the metadata cache metrics, but is also used for a bunch of other scrape metrics (`prometheus_target_scrape_pool_target_limit`, `prometheus_target_scrape_pool_targets`, `prometheus_target_sync_length_seconds`, `prometheus_target_scrape_pool_sync_total`). Those have been in place for 5 years or maybe even longer.

Technically, we can change metrics exposed by Prometheus itself, but I'm not sure if it's worth it in this case. Maybe something to keep in mind for Prometheus 3?

juliusv:
Yeah, maybe you're right and it's gonna cause too much real-world pain to change this in Prometheus 2, even if we're technically allowed to :/

Then at least I want to raise the awareness for the future that there's no such thing as a "scrape job" in Prometheus, even though people often think there is, so let's be cautious about when that pops up again somewhere else.

PR #6687, opened 2020-01-24
Trim TSDB head chunks after being cut, to reduce inuse memory

pracucci:
I'm currently working on a new experimental storage in Cortex based on TSDB (and Thanos). While benchmarking it and comparing the **inuse memory**, I've noticed that for the same exact series TSDB uses more memory than the Cortex chunks storage.

While profiling it I've found that part of such difference is given by an optimization we have in Cortex but doesn't exist in TSDB: chunks are not trimmed after being cut in the TSDB head.

In particular, a chunk byte slice is allocated in TSDB with an initial capacity of 128 bytes and then it double each time it's full:
https://github.com/prometheus/prometheus/blob/master/tsdb/chunkenc/xor.go#L58-L61

When a chunk is cut, the extra capacity of that `[]byte` is never reclaimed and we may end up (theoretically) wasting 2x space. It's not 2x space in the real world, but it may be significant. In the Cortex chunks storage, we check it and compact the `[]byte` if the capacity exceed 32 bytes.

I've applied the same optimization in TSDB and, according to our preliminary tests, it has reduced the memory used by TSDB by about -11% in our test cluster. More accurate benchmarks are required.

The change proposed in this PR is a trade-off between inuse memory and allocations: we're doing more allocations to reduce the inuse memory. For example, running the `BenchmarkLoadWAL` with `-benchmem` we can see:

```
benchmark                                                                  old ns/op      new ns/op      delta
BenchmarkLoadWAL/batches=10,seriesPerBatch=100,samplesPerSeries=7200-4     686987166      697524794      +1.53%
BenchmarkLoadWAL/batches=10,seriesPerBatch=10000,samplesPerSeries=50-4     1113564403     1128349782     +1.33%
BenchmarkLoadWAL/batches=10,seriesPerBatch=1000,samplesPerSeries=480-4     481137676      476783942      -0.90%

benchmark                                                                  old allocs     new allocs     delta
BenchmarkLoadWAL/batches=10,seriesPerBatch=100,samplesPerSeries=7200-4     640057         710074         +10.94%
BenchmarkLoadWAL/batches=10,seriesPerBatch=10000,samplesPerSeries=50-4     2395984        2395953        -0.00%
BenchmarkLoadWAL/batches=10,seriesPerBatch=1000,samplesPerSeries=480-4     569179         609208         +7.03%

benchmark                                                                  old bytes     new bytes     delta
BenchmarkLoadWAL/batches=10,seriesPerBatch=100,samplesPerSeries=7200-4     57155592      71857912      +25.72%
BenchmarkLoadWAL/batches=10,seriesPerBatch=10000,samplesPerSeries=50-4     365385744     364694736     -0.19%
BenchmarkLoadWAL/batches=10,seriesPerBatch=1000,samplesPerSeries=480-4     67557674      75518242      +11.78%
```

### DIFF
--- a/tsdb/chunkenc/chunk.go
+++ b/tsdb/chunkenc/chunk.go
@@ -49,6 +49,7 @@ type Chunk interface {
 	// be re-used or a new iterator can be allocated.
 	Iterator(Iterator) Iterator
 	NumSamples() int
+	Compact()
 }
 
 // Appender adds sample pairs to a chunk.
--- a/tsdb/chunkenc/chunk.go
+++ b/tsdb/chunkenc/chunk.go
@@ -41,14 +41,28 @@ const (
 
 // Chunk holds a sequence of sample pairs that can be iterated over and appended to.
 type Chunk interface {
+	// Bytes returns the underlying byte slice of the chunk.
 	Bytes() []byte
+
+	// Encoding returns the encoding type of the chunk.
 	Encoding() Encoding
+
+	// Appender returns an appender to append samples to the chunk.
 	Appender() (Appender, error)
+
 	// The iterator passed as argument is for re-use.
 	// Depending on implementation, the iterator can
 	// be re-used or a new iterator can be allocated.
 	Iterator(Iterator) Iterator
+
+	// NumSamples returns the number of samples in the chunk.
 	NumSamples() int
+
+	// Compact is called whenever a chunk is expected to be complete (no more
+	// samples appended) and the underlying implementation can eventually
+	// optimize the chunk.
+	// There's no strong guarantee that no samples will be appended once
+	// Compact() is called. Implementing this function is optional.
 	Compact()
 }
 
--- a/tsdb/chunkenc/xor.go
+++ b/tsdb/chunkenc/xor.go
@@ -49,6 +49,10 @@ import (
 	"math/bits"
 )
 
+const (
+	chunkCompactCapacityThreshold = 32
+)
+
 // XORChunk holds XOR encoded sample data.
 type XORChunk struct {
 	b bstream
@@ -73,6 +77,14 @@ func (c *XORChunk) Bytes() []byte {
 // NumSamples returns the number of samples in the chunk.
 func (c *XORChunk) NumSamples() int {
 	return int(binary.BigEndian.Uint16(c.Bytes()))
+}
+
+func (c *XORChunk) Compact() {
+	if l := len(c.b.stream); cap(c.b.stream) > l+chunkCompactCapacityThreshold {
+		buf := make([]byte, l)
+		copy(buf, c.b.stream)
+		c.b.stream = buf
+	}
 }
 
 // Appender implements the Chunk interface.
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -1691,6 +1691,11 @@ func (s *memSeries) cut(mint int64) *memChunk {
 	s.chunks = append(s.chunks, c)
 	s.headChunk = c
 
+	// Remove exceeding capacity from the previous chunk byte slice to save memory.
+	if l := len(s.chunks); l > 1 {
+		s.chunks[l-2].chunk.Compact()
+	}
+
 	// Set upper bound on when the next chunk must be started. An earlier timestamp
 	// may be chosen dynamically at a later point.
 	s.nextAt = rangeForTimestamp(mint, s.chunkRange)


codesome:
Some visible differences in resident memory.

Before merging I would also like to verify if it has any potential race. Consider the following cases
1. `Compact()` is called when a query is iterating that chunk - this might not be a problem as we take a subslice for the iterator already.
2. A similar case to above, but `Compact()` is called when the query is trying to get the iterator for the chunk. Here would the swapping of the byte stream in `Compact()` and taking the subslice in the iterator method cause any race?

pracucci:
Thanks @codesome for the feedback. I'm going to close this PR, cause an extended test running for a couple of days showed no significant benefit on the long run for our use case. We're still seeing TSDB significantly using more memory than Cortex chunks storage for the same series, but we need to spend more time on that in order to correctly pinpoint the causes.

_Thanks for your time so far._

pracucci:
I'm re-opening this PR cause further tests actually showed some benefits in the inuse memory reduction. Looking again at the prombench metrics, the same improvements can be observed there, except that it's less than initially estimated (-11%).

The actual inuse memory reduction is in the range of -3%,-6% when we look at the entire process (in this case Prometheus) and not just the chunks.

## Metrics from Prombench

RSS memory:

rss

Allocated memory picking the max over 5m:
`max_over_time(go_memstats_alloc_bytes{job="prometheus",namespace="prombench-[[pr-number]]",prometheus=~"(test+).*"}[5m])`

allocated

Heap inuse picking the max over 5m:
`max_over_time(go_memstats_heap_inuse_bytes{job="prometheus",namespace="prombench-[[pr-number]]",prometheus=~"(test+).*"}[5m])`

heap

## Other considerations

An argument against this PR is that we may `Compact()` all chunks nearly at the same time, in case of the same scrape interval for all metrics and not churn. In general, this PR obviously increase memory allocations.

The following shows allocations with `rate()` over `[5m]` window to see more smooth metrics which are easier to compare:

allocations

## What's next?

I would like to *get more feedback* and *more critical eyes* on the possible positive/negative effects of such change (ie. are we OK having more memory allocations to save a bit on inuse memory?).

pracucci:
> Before merging I would also like to verify if it has any potential race. Consider the following cases
> 
> 1. `Compact()` is called when a query is iterating that chunk - this might not be a problem as we take a subslice for the iterator already.
> 2. A similar case to above, but `Compact()` is called when the query is trying to get the iterator for the chunk. Here would the swapping of the byte stream in `Compact()` and taking the subslice in the iterator method cause any race?

_I'm not much familiar with all the TSDB internals, so I may miss something._

To my understanding, a query gets an Iterator from a `safeChunk`, which wraps it around the `memSeries.Lock()`:
https://github.com/prometheus/prometheus/blob/1ebcf3d5e479a5d78e8b79ee4425c3b586e17197/tsdb/head.go#L1282-L1287

Likewise, when samples are appended in the `Commit()`, the append is wrapped around the `memSeries.Lock()`:
https://github.com/prometheus/prometheus/blob/1ebcf3d5e479a5d78e8b79ee4425c3b586e17197/tsdb/head.go#L1008-L1011

The `chunk.Compact()` is called in the `memSeries.cut()`, which is called only within the `memSeries.append()` which is wrapped by the lock, so it looks safe to me.

pracucci:
## Metrics from Cortex ingesters with blocks storage

We now have some data points to compare inuse memory in Cortex ingesters (with the experimental blocks storage based on TSDB) with and without this optimization. It looks we get similar results as the Prombench.

heap_inuse

alloc_bytes

tsdb/chunkenc/chunk.go
```diff
@@ -49,6 +49,7 @@ type Chunk interface {
 	// be re-used or a new iterator can be allocated.
 	Iterator(Iterator) Iterator
 	NumSamples() int
+	Compact()
```
bwplotka: I believe we need some comment on what we expect from this method for someone who will add some new implementation soon.

bwplotka: If you will be kind enough to add some simple comments and all of others comments for GoDoc.. it would be amazing :heart: 

[Scout rule](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html)!

bwplotka:
I understand this will be irrelevant once @codesome PR will be merged, but I really like this in the interface as a nice practice that we should do in other places. 

(But maybe making code bit more complex just for this purpose is not worth it)

tsdb/chunkenc/chunk.go
bwplotka: Hm I am also thinking if we can separate the chunk interface to reader and writer ): 

It's too wide to me..

pracucci: > If you will be kind enough to add some simple comments and all of others comments for GoDoc

Sure, done. May you take a look? Feedback is welcome 🙏

pracucci: > Hm I am also thinking if we can separate the chunk interface to reader and writer ):
> It's too wide to me..

Few questions, please:
1. Do you think it's good to couple such refactoring within this optimization PR?
2. If so, may you share more details about your thoughts? Are you thinking about keeping the current `Chunk` and add an "read-only" interface?

bwplotka: No, I think I am ok to merge this and think about it later :+1: 

> If so, may you share more details about your thoughts? Are you thinking about keeping the current Chunk and add a "read-only" interface?

No..  I think about two separate interfaces as this interface is becoming too wide. Usually, things that depend on this interface either only read or only write.

bwplotka:
Thank you @pracucci!

PR #6691, opened 2020-01-24
added marshal function for targetgroup.Group

4xoc:
This PR adds a custom marshal function for JSON to ensure that targetgroup.Group is properly formatted.

Currently the default behavior of json.Marshal() for a targetgroup.Group results in a map with `{"__address__":"<address>"}` which is not parsed back into a targetgroup because a []string is expected. This results in file_sd files generated using targegroup.Group being incompatible with Prometheus. Since yaml has both functions already it is not affected.

### DIFF
--- a/discovery/targetgroup/targetgroup.go
+++ b/discovery/targetgroup/targetgroup.go
@@ -108,7 +108,7 @@ func (tg *Group) MarshalJSON() ([]byte, error) {
 
 	// return `"labels":[]` if LabelSet is empty, would be `"labels":null` instead
 	if g.Labels == nil {
-		g.Labels = make(model.LabelSet, 0)
+		g.Labels = make(model.LabelSet)
 	}
 
 	return json.Marshal(g)
--- a/discovery/targetgroup/targetgroup.go
+++ b/discovery/targetgroup/targetgroup.go
@@ -91,3 +91,25 @@ func (tg *Group) UnmarshalJSON(b []byte) error {
 	tg.Labels = g.Labels
 	return nil
 }
+
+// MarshalJSON implements the json.Marshaler interface.
+func (tg *Group) MarshalJSON() ([]byte, error) {
+	g := struct {
+		Targets []string       `json:"targets"`
+		Labels  model.LabelSet `json:"labels"`
+	}{
+		Targets: make([]string, 0, len(tg.Targets)),
+		Labels:  tg.Labels,
+	}
+
+	for _, t := range tg.Targets {
+		g.Targets = append(g.Targets, string(t[model.AddressLabel]))
+	}
+
+	// return `"labels":[]` if LabelSet is empty, would be `"labels":null` instead
+	if g.Labels == nil {
+		g.Labels = make(model.LabelSet, 0)
+	}
+
+	return json.Marshal(g)
+}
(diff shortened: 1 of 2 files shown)


roidelapluie:
What are you trying to do?

Prometheus interfaces should be marshallable to Yaml as this is what we use, but we do not guarantee json marshalling.

4xoc:
In short, I want to be able to use JSON for https://prometheus.io/docs/guides/file-sd/, generated by an application using targetgroup.Group. Currently that doesn't work because the JSON is not what Prometheus expects when reading the file content.

brian-brazil:
This is internal code of Prometheus, and we don't add dead code that we don't use. You want to work from https://github.com/prometheus/prometheus/tree/master/documentation/examples/custom-sd

4xoc:
I fail to see how this is would be dead code. The same struct is used everywhere and apparently for YAML the proper marshal support is given. This would simply make JSON support equal to YAML and allow direct marshal of json.

roidelapluie:
dead code = not used by Prometheus itself

roidelapluie:
Note that any file_sd that is json-formatted will work by prometheus, as Yaml in its 1.2 version is a superset of json.

4xoc:
Then how about making it live code?

> This customSD struct exists mostly to help us convert the internal Prometheus targetgroup.Group struct into JSON for the file_sd format.
> [...]
> When running, the adapter will listen on a channel for updates from our custom SD implementation. Upon receiving an update, it will parse the targetgroup.Groups into another map[string]*customSD, and compare it with what’s stored in the groups field of Adapter.
from https://prometheus.io/blog/2018/07/05/implementing-custom-sd/

Custom SD still uses targetgroup.Group so everyone creating an custom SD uses targetgroup anyway. Instead of converting it to another struct that resembles targetgroup.Groups anyway, one could directly marshal it into json. I believe it is very counter intuitive to not be able to marshal some type into JSON, especially when at the same time YAML works. I think allowing direct marshal of targetgroup would make it easier for everyone working with targetgroup and file_sd.

brian-brazil:
> Then how about making it live code?

This is internal code. If you've got a reason for changing this, it need to make sense entirely in the context of Prometheus itself.

I'm not seeing such a reason here, so closing. If you've questions about using the sd adpater example, the developers mailing list is the best place to ask.

4xoc:
How is this code internal if it is **the** dedicated struct to define and exchange information about targets in Prometheus? Every file_sd and every custom SD uses it and relies on it. How can it be accepted that it is not possible to use targetgroup.Group to generate a valid JSON string? I seriously fail to see how this is just dismissed on the premise of dead or internal code. Adding support for JSON marshaling does no harm whatsoever but allows easier interoperability between any software and Prometheus. Everyone using files for targets relies on this struct and as long as JSON is a supported way of formatting that information it should obviously be possible to use it to generate valid JSON.

Btw, MarshalYAML() for targetgroup is also nowhere used but still implemented

roidelapluie:
Please bring your concerns to the developers mailing list, not in github issues. On the mailing list, more people will be able to comment and discuss.

> Btw, MarshalYAML() for targetgroup is also nowhere used but still implemented

It is used in the UI to show the prometheus configuration.

Issue #500, opened 2015-02-02
Timezone support in the display layer


LordFPL:
Hi all,

Is there a way configure a timezone ? I'm always in gmt time... i try to change with a new build with a dockerfile like this : 

```
FROM prom/prometheus
RUN echo "Europe/Paris" > /etc/timezone && dpkg-reconfigure -f noninteractive tzdata
```

Not the good way ?

juliusv:
Hi, since servers usually run on GMT and that is the common time for correlating things from monitoring, logs, etc. with another, we do everything explicitly in GMT at the moment. I'm open for suggestions on how time zones could/should be configurable...

beorn7:
I guess we should have an FAQ about the timezone issue.

juliusv:
@LordFPL To clarify, this is mainly a display issue, right? I.e. in the UI you want to be able to choose between GMT and local time?

LordFPL:
@juliusv Yes, it's only a display issue. Choose between gmt/local will be a very good solution, or at least, take the host config. Thx in advance :)

juliusv:
@LordFPL Well, it's a valid feature request, but low priority for us at the moment. But we welcome contributions :) Or maybe I'll get bored one weekend and do it...

LordFPL:
@juliusv This is just a minor feature yes.

My message initially trying to see if it was possible to do it now.

The end result is mainly cosmetic ;)

jack230230:
any follow up?

fabxc:
As before, we are open to review and help with contributions but currently do not have capacity to spend time on it ourselves.

beorn7:
I'm tempted to file a bug against several governments on this planet and add it as a blocker to this feature request: "Your timezone has a design fatal design error: The time inexplicably has a gap of 3600s once per year, and at another time once per year, it jumps back 3600s, breaking monotonicity. That renders your timezone unusable for most purposes of time measurement and display."

LuboVarga:
Hi @beorn7 , there is same issue with GMT if I am not wrong. But that one is irregular and smaller scale. It is called leap second. How do you deal with that? I know that visualization of "time going backward" is an issue, but I would also like prometheus charts in local time (optionally) and I could live with some error message like "In selected period of time, time has stepped back and thus effectively making graph with non-monotonous X-axis unrendable. To visualize things, please switch back to GMT time visualization.".

[...]

PR #6692, opened 2020-01-24
React UI: Support local timezone on /graph

juliusv:
This partially implements
https://github.com/prometheus/prometheus/issues/500 in the sense that it
only addresses the /graph page, and only allows toggling between UTC and
local (browser) time, but no arbitrary timezone selection yet.

### DIFF
--- a/web/ui/react-app/src/pages/graph/Graph.tsx
+++ b/web/ui/react-app/src/pages/graph/Graph.tsx
@@ -19,6 +19,7 @@ export interface GraphProps {
     result: Array<{ metric: Metric; values: [number, string][] }>;
   };
   stacked: boolean;
+  useLocalTime: boolean;
   queryParams: QueryParams | null;
 }
 
@@ -44,7 +45,7 @@ class Graph extends PureComponent<GraphProps, GraphState> {
   };
 
   componentDidUpdate(prevProps: GraphProps) {
-    const { data, stacked } = this.props;
+    const { data, stacked, useLocalTime } = this.props;
     if (prevProps.data !== data) {
       this.selectedSeriesIndexes = [];
       this.setState({ chartData: normalizeData(this.props) }, this.plot);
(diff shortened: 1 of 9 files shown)


juliusv:
cc @boyskila @greenrover

roidelapluie:
We only display the timezome when UTC in tooltip. What about also displaying the browser timezone?

juliusv:
@roidelapluie I try to do that via the format string, and for some timezones (like `America/New_York`) it renders `EST`, for others an empty string. Not sure how to improve that...

juliusv:
I could just render the offset though...

juliusv:
@roidelapluie I changed the format string from `z` to `Z z`, see https://momentjs.com/timezone/docs/#/using-timezones/formatting/

GreenRover:
@juliusv i would just use Z. This is always correct especially for crazy thinks like russian half hour time zones. That not all timezones has an z and some have multiple. Most people dont know their own z.

juliusv:
Alright, changed it to just `Z` (numeric offset).

web/ui/react-app/src/pages/graph/Graph.tsx
```diff
@@ -57,6 +58,10 @@ class Graph extends PureComponent<GraphProps, GraphState> {
         }
       });
     }
+
+    if (prevProps.useLocalTime !== useLocalTime) {
+      this.plot();
```
boyskila: Shouldn't the displayed data (if have such) to be restored? ```this.plot(this.$chart.getData() as GraphSeries[]);```

juliusv: I'm not sure what you mean. `plot()` currently defaults to `this.state.chartData`, which is already stored. Why would I need to get it out of `this.$chart.getData()`?

boyskila: Ah, yes. Just didn't check how ```plot```is implemented, instead I saw how data is passed in ```handleResize``` and I thought it should be done the same way, anyways, ignore it :)

boyskila: Everything else looks good to me :+1:

juliusv:
Thanks, merging then!

PR #6693, opened 2020-01-24
Delete prometheus_rule_group metrics when groups are removed

roidelapluie:
Closes #6693 

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -872,7 +872,7 @@ func (m *Manager) Update(interval time.Duration, files []string, externalLabels 
 		// check if new group equals with the old group, if yes then skip it.
 		// If not equals, stop it and wait for it to finish the current iteration.
 		// Then copy it into the new group.
-		gn := groupKey(newg.name, newg.file)
+		gn := groupKey(newg.file, newg.name)
 		oldg, ok := m.groups[gn]
 		delete(m.groups, gn)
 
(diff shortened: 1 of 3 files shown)


rules/manager.go
```diff
@@ -956,16 +961,16 @@ func (m *Manager) LoadGroups(
...
 	}
 
 	return groups, nil
 }
 
 // Group names need not be unique across filenames.
-func groupKey(name, file string) string {
-	return name + ";" + file
+func groupKey(file, name string) string {
+	return file + ";" + name
```
brian-brazil: Won't this change metrics for users? Lets not change this lightly.

roidelapluie: It does not. Everything was inverted in the code. Some functions called it in the wrong order (for internal maps), this PR makes the internal map consistent with the metric.

roidelapluie:
@brian-brazil Here is the status of master:

## Here we set file, name

https://github.com/prometheus/prometheus/blob/5f27ac3583630ddd2c673b5dd96a4f8fb21bc39f/rules/manager.go#L248-L250
https://github.com/prometheus/prometheus/blob/5f27ac3583630ddd2c673b5dd96a4f8fb21bc39f/rules/manager.go#L429


## Here we set name, file

https://github.com/prometheus/prometheus/blob/5f27ac3583630ddd2c673b5dd96a4f8fb21bc39f/rules/manager.go#L877

## but... the args names are name, file

https://github.com/prometheus/prometheus/blob/5f27ac3583630ddd2c673b5dd96a4f8fb21bc39f/rules/manager.go#L969-L971

brian-brazil:
So the existing code was broken, and tests didn't catch it? We should expand the tests.

roidelapluie:
The existing code is not broken in a testable way.

There were two inversions:

the *name* of the args were inverted (name, file)
But we were calling it with the right order anyway (file,name)

As the arguments are only positional that did not cause issues.

roidelapluie:
And the place were we inverted the args were used for something else , an internal map. Which I now use in this PR as the label value that I delete. Which means that now any change in a way or another will be catched by the new tests.

brian-brazil:
Thanks!

PR #6695, opened 2020-01-24
Fix order of testutil.Equals

roidelapluie:
Equals takes the expected value as first parameter, and the actual value
as second parameter.

### DIFF
--- a/pkg/textparse/promparse_test.go
+++ b/pkg/textparse/promparse_test.go
@@ -209,7 +209,7 @@ testmetric{label="\"bar\""} 1`
 
 		i++
 	}
-	testutil.Equals(t, i, len(exp))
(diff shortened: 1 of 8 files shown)


pkg/textparse/promparse_test.go
```diff
@@ -209,7 +209,7 @@ testmetric{label="\"bar\""} 1`
 
 		i++
 	}
-	testutil.Equals(t, len(exp), i)
+	testutil.Equals(t, i, len(exp))
```
brian-brazil: You sure about this one?

roidelapluie: Indeed, good catch.

brian-brazil:
Thanks!

PR #6696, opened 2020-01-24
Move global /graph settings to navbar

juliusv:
This saves precious space at the top and makes the page look less
awkward.

Screenshots:

**Closed:**

topsettings

**Open:**

topsettings_open

### DIFF
--- a/web/ui/react-app/src/App.tsx
+++ b/web/ui/react-app/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { FC } from 'react';
+import React, { FC, useState } from 'react';
 import Navigation from './Navbar';
 import { Container } from 'reactstrap';
 
@@ -8,9 +8,11 @@ import { Alerts, Config, Flags, Rules, Services, Status, Targets, TSDBStatus, Pa
 import PathPrefixProps from './types/PathPrefixProps';
 
 const App: FC<PathPrefixProps> = ({ pathPrefix }) => {
+  const [extraNavItem, setExtraNavItem] = useState<React.ReactNode>(null);
+
   return (
     <>
-      <Navigation pathPrefix={pathPrefix} />
+      <Navigation pathPrefix={pathPrefix} extraNavItem={extraNavItem} />
       <Container fluid style={{ paddingTop: 70 }}>
         <Router basepath={`${pathPrefix}/new`}>
           <Redirect from="/" to={`${pathPrefix}/new/graph`} />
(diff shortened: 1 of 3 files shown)


juliusv:
@boyskila

roidelapluie:
js tests fail

boyskila:
Interesting! I have another idea if you like to try it. How about to move the checkboxes in the navbar and to dispatch StorageEvent on change like so:
```
window.dispatch(StorageEvent('storage', {
    key: 'enable-query-history',
    newValue: `${e.target.checked}`
}))
```
Then we can  ```window.addEventListener('storage', updateStateFunction)``` whenever we like and thus we will not be restricted to use it only in one place.
Also attaching the ```eventListener``` can be utilized in component, let say ```LocalStorageListener``` which will be responsible to update the storage and to propagate it to its children. Something like this:
```
<LocalStorageListener>
   {storage => {
        
   }}
</LocalStorageListener>
```

roidelapluie:
I don't like what this change makes on mobile, as this is not navbar items, it should not be hidden on twice on mobile.

But besides me during outages, who runs promql queries on mobile? :smile:

boyskila:
> as this is not navbar items

Yeah, should be hidden for other pages as those options are graph page related only.

boyskila:
Ops, the became hidden on unmount...ignore comment above :)

juliusv:
Whoops yeah, I always forget about the JS tests, sorry. Will update them once we agree on what we want.

@boyskila Regarding the event listener: for now I would only want to display the settings on the /graph page, as that's the only thing they affect (even if TZ settings become relevant for other settings, the query history setting is graph-page-specific). I think once we do want to support custom TZs on other pages, we'll have to rethink this all a bit. E.g. if /alerts had custom TZ support, I'd probably want to show this menu at the top too, but only with the relevant TZ option, not the query history one. So then it'd still have to be page-specific, and maybe the way it's currently done here (albeit a bit odd) is easiest for that case? But maybe there's a better way.

juliusv:
Regarding mobile: Yeah, will play around a bit more later to see if I can make it a bit nicer. You can currently still access it under mobile (at least when testing in mobile mode in Chrome), though it's hidden behind the hamburger menu).

boyskila:
I think It'd be good to have such pattern which will allow subscribing for storage change. It'd be much cleaner and relatively easy to use.

gouthamve:
@juliusv is this still relevant / active?

juliusv:
Closing for now.

Issue #6685, opened 2020-01-23
Remember current query paramenters when switching to React UI


kustodian:
## Proposal
Our monitoring alerts have links to Prometheus queries, so when we open those Prometheus links and later press `Try experimental React UI` it doesn't remember any of the query parameters, which makes it harder to try to use it and test it out.

## Bug Report
Type some query on the Graph page and click on `Try experimental React UI`. All query parameters are lost, but they should be remembered so that you can continue the investigation in the new UI where you left of.

* Prometheus version:

	2.15.1

boyskila:
I'll take a look

boyskila:
@kustodian I would add that vice versa is also not working

PR #6697, opened 2020-01-25
Change old UI link to React App to point to graph page

boyskila:
This PR should fix https://github.com/prometheus/prometheus/issues/6685


boyskila:
@juliusv

web/ui/react-app/src/Navbar.tsx
```diff
@@ -31,7 +31,17 @@ const Navigation: FC<PathPrefixProps> = ({ pathPrefix }) => {
             </NavLink>
           </NavItem>
           <NavItem>
-            <NavLink tag={Link} to={`${pathPrefix}/new/graph`}>
+            <NavLink
+              tag={Link}
+              to={`${pathPrefix}/new/graph`}
+              onClick={e => {
```
juliusv: Hmm TBH I've always used the "Graph" link at the top to get to a clean/new graph page without any panels on it, since that's the quickest way (otherwise you have to edit the URL to remove the query string, or navigate somewhere else and back to clear the page). So I'm kind of a fan of keeping that behavior (is there a good use case for the new behavior, since it effectively makes the button a no-op? What does the `navigate(...)` with the query string do btw. - since in that case we are not on the graph page, we also wouldn't have an existing query string, so why append it?

boyskila: Hm, but currently the query string is the only thing that disappears and I thought that this is a bug and this is what ```navigate``` does - prevents this by attaching the QS back. Somehow this can't be achieved by just doing ```to={`${pathPrefix}/new/graph${window.location.search}`}```

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "go.inferGopath": false
+}
```
juliusv: seems like this unrelated change made it into the PR

boyskila: yes, will remove it

boyskila: Done.

juliusv:
:+1:

PR #6698, opened 2020-01-26
Add an ignore_annotations option for unit testing

gduchene:
This changes the behavior of `promtool` so that any `<alert>` that has no `exp_annotations` field set and that is part of a `<test_group>` or an `<alert_test_case>` that has `ignore_annotations` set to `true` will not have annotations checked as part of the test.

The idea is that while it’s generally a good idea to test that both labels and annotations have expected values, it is sometimes impractical to do so, for instance if you have an annotation that contains the value of the signal that triggered an alert or if you have stuff that is very verbose and you just end up copying and pasting things from the alert definition to the test file, which is annoying and kind of defeats the purpose of the test.

Of course, we keep the existing behavior as the default. Thoughts?

### DIFF
--- a/cmd/promtool/testdata/alerts.rules
+++ b/cmd/promtool/testdata/alerts.rules
@@ -1,0 +1,10 @@
+groups:
+  - name: my-group-name
+    rules:
+      - alert: InstanceDown
+        expr: up == 0
+        for: 1m
+        labels:
+          severity: critical
+        annotations:
+          description: "stuff's happening with {{ $labels.service }}"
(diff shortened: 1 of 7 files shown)


brian-brazil:
> it is sometimes impractical to do so

I don't see how copying&pasting is impractical, the output is deterministic.

If you've complex annotations, you should probably be testing them.

roidelapluie:
I agree with Brian, if annotations output is not deterministic and can't be tested that is a buotherwise it looks trivial to copy and paste from the error output to the test file if you don't want to deal with them upfront.

gouthamve:
Thanks a lot for opening the PR and using the testing framework, but sadly I agree with Brian and Julien. We looked at this in the bug scrub. With the deterministic inputs, the output should be deterministic as well, so we don't see how the annotations could change between runs. Could you give us an instance where this could be the case and we can re-open this once we have the use-case.

PR #6701, opened 2020-01-26
promql: make active query tracker context-aware

### LINKED ISSUE #6700: pkg/gate is not really used anymore
I have digged into Promql code today.

The pkg/gate package is not really useful in PromQL for prometheus as we now have the active query tracker.

https://github.com/prometheus/prometheus/blob/d996ba20ec9c7f1808823a047ed9d5ce96be3d8f/promql/engine.go#L470-L473

The active query tracker will block the query.Exec until there is enough capacity. Therefore pkg.gate, called in engine.exec can never fail.

https://github.com/prometheus/prometheus/blob/d996ba20ec9c7f1808823a047ed9d5ce96be3d8f/promql/engine.go#L193-L196

https://github.com/prometheus/prometheus/blob/d996ba20ec9c7f1808823a047ed9d5ce96be3d8f/promql/query_logger.go#L178

It is also worth noting that now queries will no longer fail in "wait queue" for timeout, as the active query tracker is called before we have set a deadline on the context.

https://github.com/prometheus/prometheus/blob/d996ba20ec9c7f1808823a047ed9d5ce96be3d8f/promql/engine.go#L193-L199

https://github.com/prometheus/prometheus/blob/d996ba20ec9c7f1808823a047ed9d5ce96be3d8f/promql/engine.go#L430-L435

roidelapluie:
Closes #6700 

For the release notes: This will enable queries in the queue to time out again. not sure how to phrase it the best.


promql/query_logger_test.go
```diff
@@ -77,16 +78,16 @@ func TestIndexReuse(t *testing.T) {
 	}
 
 	queryLogger.generateIndices(3)
-	queryLogger.Insert("TestQuery1")
-	queryLogger.Insert("TestQuery2")
-	queryLogger.Insert("TestQuery3")
+	queryLogger.Insert(context.TODO(), "TestQuery1")
```
brian-brazil: You can use background for tests. It's not as if we'll be changing it in future.

roidelapluie: I will fix it once I have added proper tests

roidelapluie:
@brian-brazil I am considering making the ActiveQueryTracker an interface. One which would be implemented with the mmapped file, and one which would just have the `gate`, which would be selected if we dont have the mmapped file.

WDYT?

brian-brazil:
We always have the file, so I don't see a need to complicate things.

roidelapluie:
We do but people who import the promql package don't.

brian-brazil:
We shouldn't be adding complication to keep around code that is dead to us.

roidelapluie:
In this case, removing gate means that I could adapt the existing test :)

promql/engine.go
```diff
@@ -466,12 +457,16 @@ func (ng *Engine) exec(ctx context.Context, q *query) (v Value, w storage.Warnin
 	defer execSpanTimer.Finish()
 
 	queueSpanTimer, _ := q.stats.GetSpanTimer(ctx, stats.ExecQueueTime, ng.metrics.queryQueueTime)
-
-	if err := ng.gate.Start(ctx); err != nil {
-		return nil, nil, contextErr(err, "query queue")
+	// Log query in active log. The active log guarantees that we don't run over
+	// MaxConcurrent queries concurrently.
```
brian-brazil: concurrently is redundant here.

roidelapluie: fixed

brian-brazil: Did you forget to push?

roidelapluie: indeed

brian-brazil:
Thanks!

PR #6703, opened 2020-01-27
Add Github Actions workflow to run Prombench

geekodour:
Moving from [Prow](https://github.com/kubernetes/test-infra/tree/master/prow), this PR adds Github Actions workflow file `prombench.yml` that allows starting, stopping and restarting of benchmarks.

cc: @krasi-georgiev 

Signed-off-by: Hrishikesh Barman <plain.hrishikeshbman@gmail.com>

### DIFF
--- a/.github/workflows/prombench.yml
+++ b/.github/workflows/prombench.yml
@@ -82,7 +82,7 @@ jobs:
         PROJECT_ID: macro-mile-203600
         CLUSTER_NAME: prombench
         ZONE: europe-west3-a
-        TEST_INFRA_REPO: https://github.com/testpromorg/prombench.git
+        TEST_INFRA_REPO: https://github.com/prometheus/prombench.git
         PR_NUMBER: ${{ github.event.client_payload.PR_NUMBER }}
       with:
         args: >-
--- a/.github/workflows/prombench.yml
+++ b/.github/workflows/prombench.yml
@@ -1,0 +1,172 @@
+on: repository_dispatch
+name: Prombench Workflow
+jobs:
+  benchmark_start:
+    name: Benchmark Start
+    if: github.event.action == 'prombench_start'
+    runs-on: ubuntu-latest
+    steps:
+    - name: Update status to pending
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


roidelapluie:
It looks like this will only allow to test pull requests which have a branch  in the prometheus repo itself (not forks) which seems annoying.

roidelapluie:
The GITHUB_TOKEN secrets seems to be only used to set the status of the pull request, which is something I would expect github action to do out of the box (like DCO action we have)

krasi-georgiev:
> It looks like this will only allow to test pull requests which have a branch in the prometheus repo itself (not forks) which seems annoying.

This hasn't changed from the setup with Prow. Do you mean that you want to test 2  branches in your own fork?  Meaning both branches are in your own fork.

Ideally the test would be most flexible if we could just provide 2 docker images  instead of building it from github, but couldn't find an easy way automate the build and cleanup of docker images directly from a PR. The unclea part was cleaning up those when we don't need them. 

> The GITHUB_TOKEN secrets seems to be only used to set the status of the pull request, which is something I would expect github action to do out of the box (like DCO action we have)

@geekodour can comment on this, but I don't think that GH actions allow this right now.

geekodour:
> The GITHUB_TOKEN secrets seems to be only used to set the status of the pull request, which is something I would expect github action to do out of the box (like DCO action we have).

Yes, Github Actions would not know which stage of the benchmark we're in, so this has to be done manually. I [created a Github Action](https://github.com/marketplace/actions/custom-github-status) to make it look a little bit cleaner, but later decided using curl will provide more clarity.

About the link to the status, currently it just points to the actions page of the repository. According to this discussion on the[ GH Community Forum,](https://github.community/t5/GitHub-Actions/Getting-the-run-id-of-a-run-in-Github-Actions/m-p/41270/highlight/true#M4553) Github actions will be adding env vars for run id of the action. Will update the status link accordingly then.

roidelapluie:
> > It looks like this will only allow to test pull requests which have a branch in the prometheus repo itself (not forks) which seems annoying.
> 
> This hasn't changed from the setup with Prow. Do you mean that you want to test 2 branches in your own fork? Meaning both branches are in your own fork.


This would only allow to test two branches in the prometheus repo I think, due to the secret GITHUB_TOKEN.

krasi-georgiev:
let actually test it if works and can open another PR for fixes and improvments.

PR #6704, opened 2020-01-27
Update README.md

krasi-georgiev:


codesome:
Benchmark is still running after the cancellation? I can see about 3k active head series.

geekodour:
@codesome I deleted it manually. There were 3 `/prombench master` comments and two `/prombench cancel` I think there was some confusion about the status update, looking into it.

geekodour:
> Benchmark tests are running for 3 days! If this is intended ignore this message otherwise you can cancel it by commenting: /prombench cancel

When the benchmark was stopped manually, i forgot to delete the namespace. running `/prombench cancel` again on this PR will not work as `/prombench cancel` only starts cleaning once the nodepools are created.

Deleting the namespace for this PR manually to avoid getting alerts for this again.

krasi-georgiev:
at least we know that the notification works :)

PR #6706, opened 2020-01-27
consul: support reading token from a file

### LINKED ISSUE #6140: feature: Support loading file for Consul Token
## Proposal

I use vault as a secret storage solution and I think it could be useful to support loading the consul token from a file like the `bearer_token_file`. For exemple, the Marathon SD already support it.

What do you think ?
If you like it, I could start working on a PR.

Thanks

roidelapluie:
Closes #6140 

### DIFF
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -16,6 +16,7 @@ package consul
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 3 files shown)


brian-brazil:
What's the use case for this? I wouldn't think these would be rotated automatically.

roidelapluie:
If you use hashicorp vault, as described in the issue, they would be rotated.

roidelapluie:
In particular, https://www.vaultproject.io/docs/agent/index.html with a file sink.

Note that in particular, vault and consul are from the same company so the chances are high that you use both.

brian-brazil:
Hmm, can those actually be tied together? If this were possible and common I'd expect to be able to find a guide on how to do it, which I can't.

It also looks like Consul supports standard bearer tokens now, so we should check when that happened and see if we can switch.

roidelapluie:
they support both but I decided to use X-Consul-Token as this is what the native go client uses.

here is the documentation to have vault creating short term rotating consul tokens:

https://www.vaultproject.io/docs/secrets/consul/

brian-brazil:
We already have code to deal with normal bearer tokens though.

roidelapluie:
Ok I will normalize on normal bearer

wpjunior:
@roidelapluie Could I help you to finalize this issue ?

PR #6707, opened 2020-01-27
scrape: error on duplicate conflicting labels

roidelapluie:
cc @beorn7 

This is my proposal for #5861: error upon conflicting exported_ labels 

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -182,7 +182,7 @@ type scrapeLoopOptions struct {
 
 const maxAheadTime = 10 * time.Minute
 
-type labelsMutator func(labels.Labels) labels.Labels
+type labelsMutator func(labels.Labels) (labels.Labels, error)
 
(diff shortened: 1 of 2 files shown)


brian-brazil:
I don't think we should ever be erroring here.

roidelapluie:
Then we are back on  #5861 which does not have a clear solution proposal

beorn7:
I think the solution in #5861 is “clearly” sketched out (for certain definition of “clear”). Just the implementation will be a bit more involved. I'm happy to take that issue once I have some idle cycles in my schedule (which won't be happening very soon). Luckily, the practical relevance is very very low (but it's one of those issues where it irritates me immensely that we don't have defined and workable behavior here).

roidelapluie:
I did not find it clear. I am happy to implement if you can write down some tests/examples.

beorn7:
See https://github.com/prometheus/prometheus/issues/5861#issuecomment-584239460

Issue #4158, opened 2018-05-11
rules html doesn't always generate parsable configs


adamdecaf:
## Bug Report
**What did you do?**

Recording rules can have `:` on either the front of end of their name. Without quotes in yaml this leads to a parsing error.

Example: 
<img width="998" alt="screen shot 2018-05-11 at 8 43 25 am" src="https://user-images.githubusercontent.com/120951/39929147-ef4d690a-54fc-11e8-8f90-b41d4d403974.png">

```
- record: ':kube_pod_info_node_count:'
  expr: sum(min by(node) (kube_pod_info))
```

Running `promtool` (or prometheus) over the following gives an error: 

```
$ cat /tmp/mapping-bad.rules.yaml 
groups:
  - name: "node.rules"
    rules:
      - record: :kube_pod_info_node_count:
        expr: sum(min by(node) (kube_pod_info))

$ promtool check rules /tmp/mapping-bad.rules.yaml 
Checking /tmp/mapping-bad.rules.yaml
  FAILED:
yaml: line 3: mapping values are not allowed in this context
```

**What did you expect to see?**

I expected for the HTML rules output to be a valid config. 

**Environment**

* Prometheus version:`2.2.1`

```
# promtool 
$ git log -n1 
commit b424eb42e328f13eb32149225f020a4159decb17 (HEAD -> master, origin/master, origin/HEAD, adamdecaf/master)
Author: Daisy T <infoverload@users.noreply.github.com>
Date:   Mon Apr 30 20:08:45 2018 +0200

    document remote write queue parameters (#4126)
```

adamdecaf:
A fix for this could be to wrap everything in quotes, or just those with a leading (or trailing) `:`.

brian-brazil:
This smells like a bug in the yaml library we use.

juliusv:
Probably that is because the YAML field value itself doesn't start with a colon, but with an HTML tag, only the browser then shows it as starting with a colon, since it interprets the HTML. Maybe we need to add some quoting ourselves around the actual text in the HTML.

juliusv:
Relevant code piece: https://github.com/prometheus/prometheus/blob/b3ff5f6b0ee4cef9923d3ec534abfa862486a579/rules/recording.go#L112-L132

brian-brazil:
Ah, we're doing our own escaping rather than relying on Go templates.  We should probably switch this to Go templates, as they have protections for this sort of thing.

juliusv:
[...]

PR #6708, opened 2020-01-27
Rule panel

boyskila:


web/ui/react-app/src/pages/rules/RulePanel.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const mergedStyles = { background: '#f5f5f5', ...styles };
+  if (tag) {
+    customTagProps = { style: mergedStyles };
+  }
+  return (
+    <Tag {...customTagProps}>
+      <pre className="m-0" style={tag ? undefined : mergedStyles}>
+        <code>
+          <div>
+            {rule.alerts ? 'name: ' : 'record: '}
```
juliusv: - IMO better to check for `rule.type === 'alert'` here, as its more explicit.

- `name:` should be `alert:`

web/ui/react-app/src/pages/rules/RulePanel.tsx
```diff
@@ -0,0 +1,53 @@
...
+            <Link to={createExpressionLink(`ALERTS{alertname="${name}"}`)}>{name}</Link>
```
juliusv: Btw. alert names can contain any characters (though it's uncommon). E.g. try naming your alert `Foo"Bar`, and the link will be broken. Not related to this PR, but we should escape quotes in the alert name (`"` -> `\"`).

web/ui/react-app/src/pages/rules/RulePanel.tsx
```diff
@@ -0,0 +1,53 @@
...
+          </div>
+          <div>
+            expr: <Link to={createExpressionLink(query)}>{query}</Link>
+          </div>
+          {alerts && (
```
juliusv: Labels also exist on recording rules, and alerts also need their `for` (duration) field displayed (e.g. `for: 5m`).

I realize now that I overlooked some of this in both the earlier /rules and /alerts PRs, good to catch it now.

boyskila:
@juliusv last comments are already addressed, please review.

web/ui/react-app/src/pages/alerts/CollapsibleAlertPanel.tsx
```diff
@@ -25,27 +25,10 @@ const CollapsibleAlertPanel: FC<CollapsibleAlertPanelProps> = ({ rule, showAnnot
     <>
       <Alert fade={false} onClick={() => toggle(!open)} color={alertColors[rule.state]} style={{ cursor: 'pointer' }}>
         <FontAwesomeIcon icon={open ? faChevronDown : faChevronRight} fixedWidth />
-        <strong>{rule.name}</strong> ({`${rule.alerts.length} active`})
+        <strong>{replaceWith(rule.name, `"`, `'`)}</strong> ({rule.alerts.length} active)
```
juliusv: This seems incorrect, since it would display an alert with the name `foo"bar` as having the name `foo'bar`.

I think in this place here you don't actually need to escape or change anything about the name. But in places where the name appears in double quotes (like in the graph link), you will need to escape double quotes in the name `"` -> `\"`.

web/ui/react-app/src/pages/rules/RulePanel.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const { name, query, labels, annotations, type, duration } = rule;
+  const style = { background: '#f5f5f5', ...styles };
+  const ruleName = replaceWith(name, `"`, `'`);
+  return (
+    <Tag {...(tag ? { style } : {})}>
+      <pre className="m-0" style={tag ? undefined : style}>
+        <code>
+          <div>
+            {/* Type of the rule is either alerting or recording. */}
+            {type.replace('ing', ': ')}
```
juliusv: This seems a bit too "clever" and relying on an accidental fact rather than a deeper domain invariant. I'd say let's not be afraid of being more explicit, even if longer:

```suggestion
            {type === 'record' ? 'recording' : 'alerting'}:{' '}
```

web/ui/react-app/src/pages/rules/RulePanel.tsx
```diff
@@ -0,0 +1,54 @@
...
+            <>
+              <div>labels:</div>
+              {Object.entries(labels).map(([key, value]) => (
+                <div className="ml-4" key={key}>
+                  {key}: {value}
+                </div>
+              ))}
+            </>
+          )}
+          {labels && (
```
juliusv:
```suggestion
          {annotations && (
```

web/ui/react-app/src/pages/rules/RulePanel.tsx
```diff
@@ -0,0 +1,54 @@
...
+        <code>
+          <div>
+            {/* Type of the rule is either alerting or recording. */}
+            {type.replace('ing', ': ')}
+            <Link to={createExpressionLink(`ALERTS{alertname="${ruleName}"}`)}>{ruleName}</Link>
+          </div>
+          <div>
+            expr: <Link to={createExpressionLink(query)}>{query}</Link>
+          </div>
+          {isPresent(duration) && <div>for: {duration}</div>}
```
juliusv: We also want to omit the `for:` field when it's present, but the value is 0.

(see https://github.com/prometheus/prometheus/blob/17d8c49919d4f26c422745b7a0a0ae5be4d63b72/pkg/rulefmt/rulefmt.go#L116, it has the `omitempty` tag set, so normal YAML rendering would omit it when it has its Go zero value (0))

web/ui/react-app/src/pages/alerts/CollapsibleAlertPanel.tsx
boyskila: I was playing with that yesterday and found that query like ```foo"bar``` wrapped in quotes will not work since it became ```"foo"bar"```
Adding an escaping slash also will not work since it just prevents the error - the final result also will be ```"foo"bar"``` which causes the response from the query to be an error as well which isn't correct wright? It should return ```Empty query result```

web/ui/react-app/src/pages/rules/RulePanel.tsx
boyskila: I'm relying on the comments left in both rules structs regarding their type
https://github.com/prometheus/prometheus/blob/17d8c49919d4f26c422745b7a0a0ae5be4d63b72/web/api/v1/api.go#L908-L936

web/ui/react-app/src/pages/rules/RulePanel.tsx
boyskila: Ok, will check for zero value

roidelapluie:
Are we reintroducing this bug: https://github.com/prometheus/prometheus/issues/4158 or do we fix it in this PR?

web/ui/react-app/src/pages/rules/RulePanel.tsx
juliusv: Yeah, I'd still be more explicit though...

web/ui/react-app/src/pages/alerts/CollapsibleAlertPanel.tsx
juliusv: Hmm, not sure what you mean. The final query should become `ALERTS{alertname="foo\"bar"}`, which correctly produces an empty result (because no alert with this name is currently actually firing or even existing).

http://demo.robustperception.io:9090/graph?g0.range_input=1h&g0.expr=ALERTS%7Balertname%3D%22foo%5C%22bar%22%7D&g0.tab=1

So just escaping quotes with a slash should be fine, when used as a label value in a selector surrounded by quotes?

juliusv:
@roidelapluie This PR neither reintroduces nor fixes it, since the "YAML" generated by the new UI is already broken in the sense that it doesn't actually preserve whitespace properly or know anything about quoting special values, etc. Not sure what to do in that regard. To do it 100% correctly I guess we'd have to either use a client-side YAML library or generate HTML snippets on the server-side again. (or not even pretend to render it out in valid YAML in the UI)

web/ui/react-app/src/pages/rules/RulePanel.tsx
boyskila: Yep, thas fine. Will change it

web/ui/react-app/src/pages/rules/RulePanel.tsx
boyskila: copy-paste bug

web/ui/react-app/src/pages/alerts/CollapsibleAlertPanel.tsx
boyskila: Seems I was trying something different. I will revisit it.

brian-brazil: Having to manually escape HTML is a smell, do he libraries we use not take care of this for us?

juliusv: This is not about escaping HTML, it's about escaping quotes in PromQL strings (which should be a simple `"` -> `\"`).

brian-brazil: If that's the case, why is this in a line that includes HTML tags?

boyskila: The actual query doesn't contain any HTML tags. Those you mention are part of the UI

juliusv: Yep that's why I said in this place there doesn't need to be any escaping (on the HTML or PromQL level at least). HTML escaping is already automatically being done by React. But we do need to escape quotes in label values when they become part of a PromQL string like in the `ALERTS{alertname="foo\"bar"}` example that I mentioned (the relevant part of the code for that is furtehr below).

We do, however, strictly still have a bug here when it comes to YAML escaping. E.g. when the alert name contains any characters that have special meaning in YAML, we would need to detect that and quote it somehow. But at the moment the YAML is not copy-and-pastable anyway. See https://github.com/prometheus/prometheus/pull/6708#issuecomment-580221755

web/ui/react-app/src/pages/rules/RulePanel.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const { name, query, labels, annotations, type, duration } = rule;
+  const style = { background: '#f5f5f5', ...styles };
+  return (
+    <Tag {...(tag ? { style } : {})}>
+      <pre className="m-0" style={tag ? undefined : style}>
+        <code>
+          <div>
+            {/* Type of the rule is either alerting or recording. */}
+            {type === 'alerting' ? 'alert: ' : 'record: '}
+            <Link to={createExpressionLink(`ALERTS{alertname="${name.replace(/([/"])/g, "\\$1")}"}`)}>{name}</Link>
```
juliusv: Wouldn't you just need:
```suggestion
            <Link to={createExpressionLink(`ALERTS{alertname="${name.replace(/"/g, '\\"')}"}`)}>{name}</Link>
```

boyskila: yes, thanks - it's better

web/ui/react-app/src/pages/rules/RulePanel.tsx
```diff
@@ -0,0 +1,50 @@
...
+  return (
+    <Tag {...(tag ? { style } : {})}>
+      <pre className="m-0" style={tag ? undefined : style}>
+        <code>
+          {isAlertinRule(rule) ? 'alert: ' : 'record: '}
+          <Link to={createExpressionLink(`ALERTS{alertname="${name.replace(/"/g, '\\"')}"}`)}>{name}</Link>
+          <div>
+            expr: <Link to={createExpressionLink(query)}>{query}</Link>
+          </div>
+          {isAlertinRule(rule) && rule.duration > 0 && <div>for: {formatRange(rule.duration)}</div>}
```
juliusv: Alertin -> Alerting

web/ui/react-app/src/utils/index.ts
```diff
@@ -210,3 +211,11 @@ export const callAll = (...fns: Array<(...args: any) => void>) => (...args: any)
   // eslint-disable-next-line prefer-spread
   fns.filter(Boolean).forEach(fn => fn.apply(null, args));
 };
+
+export const replaceWith = (str: string, needle: string, replace: string) => {
```
juliusv: This seems unused now.

web/ui/react-app/src/utils/index.ts
```diff
@@ -210,3 +211,11 @@ export const callAll = (...fns: Array<(...args: any) => void>) => (...args: any)
...
+  return str.replace(new RegExp(needle, 'g'), `${replace}`);
+};
+
+export const isAlertinRule = (rule: Rule): rule is AlertingRule => {
```
juliusv: I don't think it's worth putting this into its own method, but we could e.g. have a string enum type for the rule types, if you prefer not to use direct string comparisons in conditions.

juliusv: I think it's not worth having a function just for this comparison, but if you prefer abstracting away the raw strings, we could add a string enum type for the rule types?

juliusv: Oh my idea was simply to have an enum like:

```ts
export enum RuleType {
  recording = 'recording';
  alerting = 'alerting';
}
```

...and then use that for the field type.

Just to not rely on direct string literals everywhere, although TS still checks in either case whether a rule can possibly have a given type (literal or not).

The first type guard example you linked (with `isFish`) checks for the existence of a field, not the value... which I find a bit more indirect than directly checking the type. I think I would use that pattern if I had multiple things that could swim / fly / etc., and I didn't really care what concrete type they are, just that they can do a certain behavior (like with interfaces in Go). If I have two concrete types that I want to render quite differently, I think I prefer what we currently do (comparison on the `type` field). I would like the later example with `typeof` type guards (https://www.typescriptlang.org/docs/handbook/advanced-types.html#typeof-type-guards), but those string comparisons only seem to work for [number, string, boolean, symbol] types.

web/ui/react-app/src/pages/rules/RulePanel.tsx
```diff
@@ -0,0 +1,50 @@
...
+export const RulePanel: FC<RulePanelProps> = ({ rule, styles = {}, tag }) => {
+  const Tag = tag || Fragment;
+  const { name, query, labels } = rule;
+  const style = { background: '#f5f5f5', ...styles };
+  return (
+    <Tag {...(tag ? { style } : {})}>
+      <pre className="m-0" style={tag ? undefined : style}>
+        <code>
+          {rule.type === 'alerting' ? 'alert: ' : 'record: '}
+          <Link to={createExpressionLink(`ALERTS{alertname="${name.replace(/"/g, '\\"')}"}`)}>{name}</Link>
```
juliusv: A recording rule does not produce `ALERTS` series, it records directly into the name given after `record:`, so for recording rules you will want to point the expression link directly at that metric name, without the `ALERTS{alertname="..."}` around it.

web/ui/react-app/src/pages/rules/RulePanel.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const { name, query, labels } = rule;
+  const style = { background: '#f5f5f5', ...styles };
+  return (
+    <Tag {...(tag ? { style } : {})}>
+      <pre className="m-0" style={tag ? undefined : style}>
+        <code>
+          {rule.type === 'alerting' ? 'alert: ' : 'record: '}
+          <Link
+            to={createExpressionLink(
+              rule.type === 'alerting' ? `ALERTS{alertname="${name.replace(/"/g, '\\"')}"}` : name.replace(/"/g, '\\"')
```
juliusv: The quote escaping in the case of `ALERTS{alertname="..."}` is necessary because the alert name may contain quotes and appears in the context of a label value, which is a string enclosed in quotes. Neither of those two contextual conditions apply in the case of a recording rule, where the recorded metric name cannot contain quotes, nor does it appear as part of a string. So in that case we can just omit the escaping:

```suggestion
              rule.type === 'alerting' ? `ALERTS{alertname="${name.replace(/"/g, '\\"')}"}` : name
```

juliusv:
:+1: besides last nit

web/ui/react-app/src/pages/rules/RulePanel.tsx
boyskila: Yeah, make sense, thanks. Just push the suggested change.

juliusv:
:+1: will merge after CI passes

PR #6712, opened 2020-01-28
Remove MaxConcurrent from the PromQL engine opts

roidelapluie:
Since we use ActiveQueryTracker to check for concurrency in
d992c36b3a08470532fb4fb36a768a20ec8f2e12 it does not make sense to keep
the MaxConcurrent value as an option of the PromQL engine.

This pull request removes it from the PromQL engine options, sets the
max concurrent metric to -1 if there is no active query tracker, and use
the value of the active query tracker otherwise.

It removes dead code and also will inform people who import the promql
package that we made that change, as it breaks the EngineOpts struct.


brian-brazil:
Thanks!

bwplotka:
Thanks!

PR #6716, opened 2020-01-29
Storage concurrently

liguozhong:


storage/fanout.go
```diff
@@ -231,22 +232,38 @@ func NewMergeQuerier(primaryQuerier Querier, queriers []Querier) Querier {
...
-				continue
-			} else {
-				return nil, nil, err
+	wg := sync.WaitGroup{}
+	wg.Add(len(q.queriers))
+	queryResult := make(map[int]SeriesSet)
+	var priErr error
+	isPriError := false
+	for idx, qr := range q.queriers {
+		go func(index int, querier Querier) {
```
brian-brazil: If there's no remote storage configured do we end up creating goroutines?

storage/fanout.go
```diff
@@ -231,22 +232,38 @@ func NewMergeQuerier(primaryQuerier Querier, queriers []Querier) Querier {
...
+			defer wg.Done()
+			set, wrn, err := querier.Select(params, matchers...)
+			q.setQuerierMap[set] = querier
```
brian-brazil: You need locking on all these.

storage/fanout.go
```diff
@@ -231,22 +232,38 @@ func NewMergeQuerier(primaryQuerier Querier, queriers []Querier) Querier {
...
+				} else {
+					isPriError = true
+					priErr = err
+				}
+			}
+			queryResult[index] = set
+		}(idx, qr)
+	}
+	wg.Wait()
+	if isPriError {
```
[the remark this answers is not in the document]
liguozhong: Thank you. According to your suggestion, two variables can be reduced to one

storage/fanout.go
liguozhong: Thank you for your review. I have modified the code according to your suggestion .

storage/fanout.go
liguozhong: Thank you for your review. According to my understanding, only 'q.setqueriermap' has the risk of concurrent modification, so the lock is only added to 'q.setqueriermap'

Lock.Lock ()
q.setQuerierMap[set] = querier
lock.Unlock()

brian-brazil: It's not the only one.

liguozhong: image
I checked the go syntax. Waitgroup.done is not thread safe either. Two locks were added specially to ensure that the slowest query is not affected by the lock performance

brian-brazil: Huh? Waitgroup is threadsafe.

storage/fanout.go
```diff
@@ -231,23 +232,66 @@ func NewMergeQuerier(primaryQuerier Querier, queriers []Querier) Querier {
...
-				return nil, nil, err
+	if len(q.queriers) > 1 {
+		wg := sync.WaitGroup{}
+		wg.Add(len(q.queriers))
+		queryResult := make(map[int]SeriesSet)
+		var priErr error = nil
+		var lock sync.Mutex
+		var wgLock sync.Mutex
+		for idx, qr := range q.queriers {
+			go func(index int, querier Querier) {
```
brian-brazil: You could simplify and reduce duplication if you did all the processing of the result outside the goroutine.

liguozhong: image
Thank you for your patient guidance. I removed some duplicate code according to your suggestion

storage/fanout.go
```diff
@@ -231,27 +232,57 @@ func NewMergeQuerier(primaryQuerier Querier, queriers []Querier) Querier {
...
-			} else {
+	if len(q.queriers) > 1 {
+		wg := sync.WaitGroup{}
+		wg.Add(len(q.queriers))
+		var priErr error = nil
+		var lock sync.Mutex
+		for _, qr := range q.queriers {
+			go func(querier Querier) {
+				defer wg.Done()
+				set, wrn, selectError := querier.Select(params, matchers...)
```
brian-brazil: I was thinking more that you take this result and send it down a channel.

liguozhong: done,move to https://github.com/prometheus/prometheus/pull/6768

brian-brazil:
Superseded by #6770

PR #6720, opened 2020-01-29
Scrape: test samples_post_metric_relabeling when metrics are dropped

roidelapluie:


brian-brazil:
Thanks!

PR #6726, opened 2020-01-30
expose last sent timestamp from storage/remote

rfratto:
I'm writing a tool that needs access to the last successfully written timestamp from a given remote write config. This PR adds a new method to expose it from `remote.Storage`. 

/cc @cstyan

### DIFF
--- a/storage/remote/storage.go
+++ b/storage/remote/storage.go
@@ -20,9 +20,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/go-kit/kit/log"
+	"github.com/pkg/errors"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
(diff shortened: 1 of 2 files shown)


brian-brazil:
I don't see where you're using this code, we only add and keep code that Prometheus itself needs.

rfratto:
This would be used by a new project that imports Prometheus as a library. The problem I'm dealing with is having multiple instances of remote storage and getting the last written timestamp for each of them. Each remote storage instance may share various subsets of remote writes with the same hashes. 

If adding a new method is out of the question, another way I could tackle this would be to pass through the currently [unused](https://github.com/prometheus/prometheus/blob/master/storage/remote/storage.go) registerer down to the queue manager and have all of its metrics be registered with that rather than using promauto. That would then allow me to provide my own registerer and scrape the metrics I want through there. WDYT?

roidelapluie:
Honestly if the registerer is unused we will probably remove it.

cstyan:
It should be passed to the queue manager, probably was mistakenly removed in some refactoring cc @csmarchbanks

csmarchbanks:
Looks like the registerer was never passed down to the queue manager, it used to be used to register a couple of metrics but those got moved to promauto. I would be for passing the register down since that is the direction we want to move anyway.

roidelapluie:
My personal taste is to avoid promauto too and register to the registerer.

rfratto:
I'm going to close this. I have a second PR almost ready to register the queue manager metrics to the registerer. After I do some local testing, I'll submit that instead.

PR #6728, opened 2020-01-30
pass registerer from storage to queue manager for its metrics

rfratto:
Continuing the discussion from #6726, this PR uses the previously unused registerer passed to NewStorage to create the metrics in the queue manager. 

Tests had to be modified to create new registerers so they didn't register the metrics twice.

Signed-off-by: Robert Fratto <robert.fratto@grafana.com>

/cc @roidelapluie @csmarchbanks

### DIFF
--- a/storage/remote/read_test.go
+++ b/storage/remote/read_test.go
@@ -93,7 +93,7 @@ func TestNoDuplicateReadConfigs(t *testing.T) {
 	}
 
 	for _, tc := range cases {
-		s := NewStorage(nil, prometheus.DefaultRegisterer, nil, dir, defaultFlushDeadline)
+		s := NewStorage(nil, prometheus.NewRegistry(), nil, dir, defaultFlushDeadline)
 		conf := &config.Config{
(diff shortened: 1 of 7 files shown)


storage/remote/read_test.go
```diff
@@ -93,7 +93,7 @@ func TestNoDuplicateReadConfigs(t *testing.T) {
 	}
 
 	for _, tc := range cases {
-		s := NewStorage(nil, prometheus.DefaultRegisterer, nil, dir, defaultFlushDeadline)
```
cstyan: was this change required for something?

rfratto: yeah, actually. `NewStorage` always calls `NewWriteStorage` even if there aren't any remote write configs, so the metrics were getting re-registered for subsequent iterations of the loop here

cstyan: Ah right I see now. This is another place you could just pass nil for the registerer if we don't care about any metric values during the test.

rfratto:
@cstyan Everything passes when using nil for the registerer, PTAL

PR #6734, opened 2020-01-31
rules: Cleanup unused function alert.Duration

roidelapluie:
The function HoldDuration and Duration did the exact same thing.

Let's only keep HoldDuration() as Duration() is more confusing.


brian-brazil:
Thanks!

PR #6737, opened 2020-02-01
Escape invalid css selector characters in tooltip ID

### LINKED ISSUE #6732: React UI: Targets page doesn't work if there job name with special characters
**What did you do?**

Opened `/new/targets` page

List of targets

Blank page. Every time.
Opening Chrome console reveal JavaScript error:
```
react-dom.production.min.js:4482 DOMException: Failed to execute 'querySelectorAll' on 'Document': 'series-labels-prometheus/test' is not a valid selector.
```

**Environment**

* Prometheus version:
prometheus, version 2.15.2 (branch: HEAD, revision: d9613e5c466c6e9de548c4dae1b9aabf9aaf7c57)
  build user:       root@688433cf4ff7
  build date:       20200106-14:50:51
  go version:       go1.13.5

Just have a job with name containing e.g. forward slash:

```
scrape_configs:
  - job_name: 'prometheus/test'
    static_configs:
      - targets: ['docker.for.win.localhost:9090']
```

hooten:
This pull request fixes a bug that caused the React targets page to fail when a special character was used.

The root cause was this. The HTML `id` elements for the discovered label `Tooltip`s could contain characters that would make the `id` and invalid css selector. As a result, when the Reactstrap `Tooltip` code called `querySelectorAll` on the `id`, an error was thrown.

For for information, see https://stackoverflow.com/questions/37270787/uncaught-syntaxerror-failed-to-execute-queryselector-on-document

The result was to escape the scrape pool name in the html ID. I used a [polyfill](https://www.npmjs.com/package/css.escape) to import `CSS.escape` into [browsers that don't natively support it](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape).

I modified a react page-level integration test to use the job name of `prometheus/test` reported in the bug. This reproduced the issue. Escaping the characters fixed the issue.

Fixes #6732


hooten:
@juliusv I'd appreciate a review whenever you get a chance

web/ui/react-app/src/pages/targets/TargetLabels.tsx
```diff
@@ -15,11 +15,15 @@ export interface TargetLabelsProps {
 
 const formatLabels = (labels: Labels): string[] => Object.keys(labels).map(key => `${key}="${labels[key]}"`);
 
+const sanitizeIDSequence = (scrapePool: string): string => {
+  return scrapePool.replace(/\W/g, (substring: string): string => substring.charCodeAt(0).toString());
```
juliusv: I wonder if I'm being tooooo nitpicky here, but if someone actually does also have a job name that has the number as part of its name, we could produce non-unique IDs here. Can we either properly escape things (using any escape char that is valid in CSS selectors), or alternatively just generate other kinds of unique IDs?

hooten: That's a good point! I can just escape every character.

hooten: Actually, I'll look into both options

hooten: I updated the PR to use `CSS.escape` (with polyfill). Thoughts?

juliusv:
:+1: Awesome, thanks :)

hooten:
Thanks for the review and merge!

PR #6738, opened 2020-02-01
PromQL engine refactoring

### LINKED ISSUE #6739: Remove LookbackDelta and DefaultEvaluationInterval from the global variables of promql
## Proposal

The behaviour of a PromQL engine should not depend on global variables of the PromQL package, especially when we alter those global variables at runtime.

https://github.com/prometheus/prometheus/blob/0e912faf4f52ed3f803cc18d0ebbd4eb0ba5a8bd/promql/engine.go#L55-L63

Therefore I would like to move them as EngineOpts

https://github.com/prometheus/prometheus/blob/0e912faf4f52ed3f803cc18d0ebbd4eb0ba5a8bd/promql/engine.go#L216-L223

roidelapluie:
## Change 1

Following https://golang.org/pkg/context/:
 Do not store Contexts inside a struct type; instead, pass a Context
explicitly to each function that needs it. The Context should be the
first parameter, typically named ctx.

I have seen no obvious reason not to do so in the engine.

## Change 2 and 3

Remove LookbackDelta and DefaultEvaluationInterval from the global variables of promql

closes  #6739 

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -52,23 +52,6 @@ const (
 	minInt64 = -9223372036854775808
 )
 
-var (
-
-	// DefaultEvaluationInterval is the default evaluation interval of
-	// a subquery in milliseconds.
-	DefaultEvaluationInterval int64
-)
-
-// SetDefaultEvaluationInterval sets DefaultEvaluationInterval.
-func SetDefaultEvaluationInterval(ev time.Duration) {
-	atomic.StoreInt64(&DefaultEvaluationInterval, durationToInt64Millis(ev))
-}
-
-// GetDefaultEvaluationInterval returns the DefaultEvaluationInterval as time.Duration.
-func GetDefaultEvaluationInterval() int64 {
-	return atomic.LoadInt64(&DefaultEvaluationInterval)
-}
-
 type engineMetrics struct {
 	currentQueries       prometheus.Gauge
 	maxConcurrentQueries prometheus.Gauge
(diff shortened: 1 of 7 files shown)


brian-brazil:
Do you know why it's suggested not to store it in a struct? We use one struct per query, so I don't see a problem with what we're currently doing.

roidelapluie:
I guess that they do not want a context to cancel a parent function's context.

roidelapluie:
Also I would like to see how we could implement time out when reading TSDB. Currently that part of ther query can never time out, in such a situation we want to use context correctly to start with.

roidelapluie:
> https://github.com/prometheus/prometheus/issues/5878

promql/engine.go
```diff
@@ -350,6 +340,21 @@ func (ng *Engine) SetQueryLogger(l QueryLogger) {
 	}
 }
 
+// GetLookbackDelta returns the lookback delta of the query engine.
+func (ng *Engine) GetLookbackDelta() time.Duration {
```
juliusv: Plain getters that don't require any mutexes / special atomic code etc. are unusual in Go - instead, just exporting the struct field is more common / preferred, unless there's a special reason to abstract it behind a function.

juliusv:
For contexts, despite that advice I think it's totally fine to keep them in struct fields, as long as those structs are request/query-scoped, or whatever the scope of that context should be.

roidelapluie:
looks like there is no real concensus now

juliusv:
@roidelapluie FWIW I liked the bits that got rid of global variables, but would maybe be best to submit that separately.

roidelapluie:
@juliusv TBH LookbackDelta should really go away from global because we pass it in the evaluator struct anyway. Will open a PR.

juliusv:
@roidelapluie Since it should almost never be changed, how about keeping a global `const DefaultLookbackDelta` (instead of `var`) and then if a user leaves the `LookbackDelta` to its zero value in the options struct, set it to the default value?

roidelapluie:
> @roidelapluie Since it should almost never be changed, how about keeping a global `const DefaultLookbackDelta` (instead of `var`) and then if a user leaves the `LookbackDelta` to its zero value in the options struct, set it to the default value?

Why would we remove the ability to disable the loopback delta?

juliusv:
@roidelapluie Do you mean you actually want users to be able to set it to `0` and make vector selectors not select anything?

roidelapluie:
It would select if the timestamp is right

juliusv:
Hmmm yeah... I wonder if there's a use case for that. But maybe still good to not rule it out for now, and always make the setting mandatory then. However, it might unknowingly break existing users if they just update the PromQL package and aren't aware that they need to set a new field in the options or otherwise get no lookback :)

roidelapluie:
> Hmmm yeah... I wonder if there's a use case for that. But maybe still good to not rule it out for now, and always make the setting mandatory then. However, it might unknowingly break existing users if they just update the PromQL package and aren't aware that they need to set a new field in the options or otherwise get no lookback :)

let's continue in #6746 . I have used a pointer to check if the value is defined

PR #6744, opened 2020-02-03
Split promql package into parser and engine

### LINKED ISSUE #6742: Split promql package into parser and engine
## Motivation

Currently the `promql` package includes both the PromQL parser and the PromQL engine.

The PromQL language server imports this package, but actually only needs the parser part.

Importing the full PromQL package, brings a several disadvantages, compared to only importing the parser:

* A lot of unnecessary dependencies are included. This makes the language server binary almost three times larger, than it would be when only importing the parser part.
* The engine transitively depends on several operating system specifics, such as mmap or unix system calls. This makes it impossible to compile the language server for certain platforms, like Web Assembly.

## Proposal

It is proposed to create a new `parser` subpackage in the `promql` package that only includes the parser.

cc @brian-brazil @juliusv

slrtbtfs:
Fixes #6742 .

Introduces API breakage, since things are moved from `promql` to `promql/parser`.


promql/engine.go
```diff
@@ -394,24 +395,10 @@ func (ng *Engine) newQuery(q storage.Queryable, expr Expr, start, end time.Time,
 	return qry
 }
 
-// testStmt is an internal helper statement that allows execution
-// of an arbitrary function during handling. It is used to test the Engine.
-type testStmt func(context.Context) error
```
roidelapluie: I don't think this belongs to the parser

slrtbtfs: Not really.

However, `testStmt` is of type `Statement`, which can only be implemented in the parser package because of https://github.com/slrtbtfs/prometheus/blob/split_parser/promql/parser/ast.go#L54 .


The alternative to what I'm doing would be making `Statement.stmt()` public, so `testStmt` could live in the engine, but that kind of defeats the point of  `Statement.stmt()`.

Since having one more `Statement` type in the parser doesn't really hurt I decided to go that way.

slrtbtfs:
This should be in a mergeable state now.

bwplotka:
Please review, but we will potentially merge only after full v2.16.0 is out cc @slrtbtfs

promql/value.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2017 The Prometheus Authors
+// Copyright 2020 The Prometheus Authors
```
brian-brazil: Don't bump these

promql/functions.go
```diff
@@ -827,353 +821,103 @@ func dateWrapper(vals []Value, enh *EvalNodeHelper, f func(time.Time) float64) V
...
-		ReturnType: ValueTypeVector,
-		Call:       funcYear,
-	},
-}
-
-// getFunction returns a predefined Function object for the given name.
-func getFunction(name string) (*Function, bool) {
-	function, ok := Functions[name]
-	return function, ok
+var FunctionCalls = map[string]FunctionCall{
```
brian-brazil: Having the list of functions in two places feels wrong.

slrtbtfs: The reason for that is that the implementation of the function calls shouldn't belong to the parser package, but the type information should.

Do you have an idea for another way to achieve this?

slrtbtfs:
Merge conflicts should be resolved now.

slrtbtfs:
There have been some new merge conflicts, which are resolved now.

promql/functions.go
brian-brazil: Not offhand, but at the least we should have a test that the two have the same elements.

slrtbtfs: I've added such tests.

promql/functions_test.go
```diff
@@ -56,3 +58,24 @@ func TestDeriv(t *testing.T) {
...
+			panic(fmt.Sprintf("function %s exists in promql package, but not in parser package", i))
+		}
+
+	}
+
+	for i := range parser.Functions {
+		_, ok := FunctionCalls[i]
+
+		if !ok {
+			panic(fmt.Sprintf("function %s exists in promql package, but not in parser package", i))
```
roidelapluie:
```suggestion
			panic(fmt.Sprintf("function %s exists in parser package, but not in promql package", i))
```

slrtbtfs: Thanks!

promql/engine.go
```diff
@@ -1038,62 +1025,64 @@ func (ev *evaluator) eval(expr Expr) Value {
...
 			// Matrix evaluation always returns the evaluation time,
 			// so this function needs special handling when given
 			// a vector selector.
-			vs, ok := e.Args[0].(*VectorSelector)
+			vs, ok := e.Args[0].(*parser.VectorSelector)
 			if ok {
-				return ev.rangeEval(func(v []Value, enh *EvalNodeHelper) Vector {
-					return e.Func.Call([]Value{ev.vectorSelector(vs, enh.ts)}, e.Args, enh)
+				return ev.rangeEval(func(v []parser.Value, enh *EvalNodeHelper) Vector {
+					return funcTimestamp([]parser.Value{ev.vectorSelector(vs, enh.ts)}, e.Args, enh)
```
roidelapluie: why not using e.Func.Call ?

slrtbtfs: After the package split it would be `FunctionCalls[e.Func.Name]`.

There's no particularly good reason to use the hard coded `funcTimestamp` here, so I've now changed it back to actually look up the function in this case, too.

Changed it accordingly.

promql/functions_test.go
```diff
@@ -56,3 +58,20 @@ func TestDeriv(t *testing.T) {
 	testutil.Assert(t, len(vec) == 1, "Expected 1 result, got %d", len(vec))
 	testutil.Assert(t, vec[0].V == 0.0, "Expected 0.0 as value, got %f", vec[0].V)
 }
+
+func TestFunctionList(t *testing.T) {
+	// Test that Functions and parser.Functions list the same functions
```
bwplotka: Missing trailing period.

promql/functions_test.go
```diff
@@ -56,3 +58,20 @@ func TestDeriv(t *testing.T) {
...
+	for i := range FunctionCalls {
+		_, ok := parser.Functions[i]
+		if !ok {
+			panic(fmt.Sprintf("function %s exists in promql package, but not in parser package", i))
```
bwplotka: I think we should use `testutil` not panic.

slrtbtfs:
One of the windows tests (github.com/prometheus/prometheus/rules) timed out in one of the test runs.

On a rerun everything was fine.

Doesn't seem to be related to this PR, tough.

PR #6745, opened 2020-02-03
rules: Make deleted rule series as stale after a reload

### LINKED ISSUE #5195: Deleted rules don't go stale after a reload
## Bug Report
**What did you do?**

I noticed this while digging into https://github.com/prometheus/prometheus/issues/5193.

If you delete a recording rule from the end of your rules and reload, the rule doesn't go stale.

After deleting a rule and reloading, I expect it to go stale in the next eval cycle.

The recording rule time series sticks around for 5m.

**Environment**

docker container

* System information:

```
/prometheus $ uname -srm
Linux 4.20.6-arch1-1-ARCH x86_64
```

* Prometheus version:

```
/prometheus $ prometheus --version
prometheus, version 2.7.1 (branch: HEAD, revision: 62e591f928ddf6b3468308b7ac1de1c63aa7fcf3)
  build user:       root@f9f82868fc43
  build date:       20190131-11:16:59
  go version:       go1.11.5
```

```
/etc/prometheus $ cat prometheus.yml
# my global config
global:

# Alertmanager configuration
alerting:
  alertmanagers:
  - static_configs:
    - targets:
      # - alertmanager:9093

rule_files:                
- "first_rules.yml"
  # - "second_rules.yml"

scrape_configs:
  - job_name: 'prometheus'

    # scheme defaults to 'http'.

    static_configs:
    - targets: ['localhost:9090']
```

```
/etc/prometheus $ cat first_rules.yml 
groups:
- name: testgroup
  rules:
  - record: test
    expr: 11111
    labels:
      name: my-first
  - record: test
    expr: 22222
    labels:
      name: my-second
```

roidelapluie:
Fix #5195

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -329,12 +329,12 @@ func (g *Group) run(ctx context.Context) {
 		if !s || g.opts.LookbackDelta < g.interval {
 			return
 		}
-		now := time.Now()
-		// Wait for two extra intervals to give the opportunity to renamed rules
-		// to insert new series in the tsdb.
-		select {
-		case <-g.managerDone:
-		case <-tick.C:
+		go func(now time.Time) {
+			// Wait for 2 intervals to give the opportunity to renamed rules
+			// to insert new series in the tsdb. At this point if there is a
+			// renamed rule, it should already be started.
+			tick := time.NewTicker(2 * g.interval)
+			defer tick.Stop()
 			select {
 			case <-g.managerDone:
 			case <-tick.C:
@@ -345,7 +345,7 @@ func (g *Group) run(ctx context.Context) {
 				}
 				g.cleanupStaleSeries(now)
 			}
-		}
+		}(time.Now())
 	}
 
 	iter()
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -333,11 +333,9 @@ func (g *Group) run(ctx context.Context) {
 			// Wait for 2 intervals to give the opportunity to renamed rules
 			// to insert new series in the tsdb. At this point if there is a
 			// renamed rule, it should already be started.
-			tick := time.NewTicker(2 * g.interval)
-			defer tick.Stop()
 			select {
 			case <-g.managerDone:
-			case <-tick.C:
+			case <-time.After(2 * g.interval):
 				for _, rule := range g.seriesInPreviousEval {
 					for _, s := range rule {
 						g.staleSeries = append(g.staleSeries, s)
@@ -898,7 +896,7 @@ func (m *Manager) Stop() {
 	}
 
 	// Shut down the groups waiting multiple evaluation intervals to write
-	// staleness marlers.
+	// staleness markers.
 	close(m.done)
 
 	level.Info(m.logger).Log("msg", "Rule manager stopped")
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -326,7 +326,7 @@ func (g *Group) run(ctx context.Context) {
 	defer tick.Stop()
 
 	makeStale := func(s bool) {
-		if !s || g.opts.LookbackDelta < g.interval {
+		if !s {
 			return
 		}
 		go func(now time.Time) {
@@ -855,7 +855,6 @@ type ManagerOptions struct {
 	OutageTolerance time.Duration
 	ForGracePeriod  time.Duration
 	ResendDelay     time.Duration
-	LookbackDelta   time.Duration
 
 	Metrics *Metrics
 }
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -381,7 +381,6 @@ func (g *Group) run(ctx context.Context) {
 func (g *Group) stopAndMakeStale() {
 	g.done <- true
 	<-g.terminated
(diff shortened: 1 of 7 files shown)


rules/manager.go
```diff
@@ -901,6 +905,12 @@ func (m *Manager) Update(interval time.Duration, files []string, externalLabels
 	// Stop remaining old groups.
 	for n, oldg := range m.groups {
 		oldg.stop()
+		for _, rule := range oldg.seriesInPreviousEval {
+			for _, s := range rule {
+				oldg.staleSeries = append(oldg.staleSeries, s)
+			}
+		}
+		oldg.cleanupStaleSeries(time.Now())
```
brian-brazil: You're doing this too early. What if a new rules is created with the same name before when the next eval interval will be?

roidelapluie: Done. It will make reloads longer but now is right.

roidelapluie: Note that to be completely safe we probably should wait 2*time.interval because the rule groups are delayed by a seed.

rules/manager.go
```diff
@@ -922,16 +931,16 @@ func (m *Manager) Update(interval time.Duration, files []string, externalLabels
 	// Stop remaining old groups.
 	wg.Add(len(m.groups))
 	for n, oldg := range m.groups {
-		go func() {
-			oldg.stop()
+		go func(n string, g *Group) {
+			g.stopAndMakeStale()
+			if m := g.metrics; m != nil {
+				m.groupInterval.DeleteLabelValues(n)
```
roidelapluie: @brian-brazil is this thread safe to your knowledge?

roidelapluie:
I still need to add a test to see that we don't insert stale maker on stopping

roidelapluie:
I think I will move the stale markers to a goroutine. That way it will not impact the reload and it should be safe.

roidelapluie:
I think I underestimated the task but hey now here we go ... ready for review ... i might still create helpers function to shorten the storage creation etc in tests, WDYT?

rules/manager.go
brian-brazil: Looking at the code in client_golang, there's a mutex in use.

rules/manager.go
```diff
@@ -314,14 +325,36 @@ func (g *Group) run(ctx context.Context) {
 	tick := time.NewTicker(g.interval)
 	defer tick.Stop()
 
+	makeStale := func(s bool) {
+		if !s || g.opts.LookbackDelta < g.interval {
```
brian-brazil: Why is LookbackDelta being considered here? Whether we write stale markers should have nothing to do with how a subsequent query happens to be configured.

brian-brazil:
> I think I underestimated the task but hey now here we go

They way I've looked at is that if you're deleting a rule group it's you no longer care about recording rules and their output, so there's not much point in getting the staleness just right in that case.

rules/manager.go
roidelapluie: it is a safety if someone deletes a rule with interval: 30m so that we don't have go routine out there for 1 hour.

brian-brazil: That's already how it works for scraping, let's keep things simple.

rules/manager.go
roidelapluie: https://github.com/prometheus/client_golang/blob/76dd6c581988366a52807465d426f83e776128ad/prometheus/vec.go#L253

great!

rules/manager.go
roidelapluie: I have removed that bit and simplified the tests.

rules/manager.go
```diff
@@ -314,14 +325,36 @@ func (g *Group) run(ctx context.Context) {
...
+		}
+		go func(now time.Time) {
+			// Wait for 2 intervals to give the opportunity to renamed rules
+			// to insert new series in the tsdb. At this point if there is a
+			// renamed rule, it should already be started.
+			select {
+			case <-g.managerDone:
+			case <-time.After(2 * g.interval):
+				for _, rule := range g.seriesInPreviousEval {
+					for _, s := range rule {
```
brian-brazil: You're shadowing `s` here. Why not move the top if out of this function?

roidelapluie: Done, I have also set g.seriesInPreviousEval to nil so it can be garbage collected before 2 * g.interval

rules/manager.go
```diff
@@ -330,17 +330,19 @@ func (g *Group) run(ctx context.Context) {
 			return
 		}
 		go func(now time.Time) {
+			for _, rule := range g.seriesInPreviousEval {
+				for _, r := range rule {
```
brian-brazil: You can do `g.staleSeries = append(g.staleSeries, rule...)`

roidelapluie: cannot use rules (type map[string]labels.Labels) as type labels.Labels in append

roidelapluie: I will improve it

roidelapluie: Well not sure how to improve that

brian-brazil:
:+1:

PR #6746, opened 2020-02-03
Make lookbackDelta a option of QueryEngine

roidelapluie:

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -232,7 +232,7 @@ type Engine struct {
 	activeQueryTracker *ActiveQueryTracker
 	queryLogger        QueryLogger
 	queryLoggerLock    sync.RWMutex
-	lookbackDelta      time.Duration
+	LookbackDelta      time.Duration
 }
 
 // NewEngine returns a new engine.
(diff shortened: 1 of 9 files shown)


roidelapluie:
@juliusv WDYT?

roidelapluie:
I will think this again. That does not look good.

juliusv:
@roidelapluie It'd be fine with me (modulo the comment about trivial getter methods from the other PR).

@brian-brazil what do you think?

roidelapluie:
@juliusv the getter is there to make sure that other packages can not set the value

juliusv:
Hm ok, I'm not too worried about that, but seems like a valid enough reason.

roidelapluie:
I have used your Zero idea. Setting th query lookback to zero is anyway not helpful in prometheus normal use case. We will log to debug if it is set to 0 and changed to the default.

brian-brazil:
:+1:

promql/engine.go
```diff
@@ -305,6 +306,13 @@ func NewEngine(opts EngineOpts) *Engine {
 		metrics.maxConcurrentQueries.Set(-1)
 	}
 
+	if opts.LookbackDelta == 0 {
+		opts.LookbackDelta = defaultLookbackDelta
+		if l := opts.Logger; l != nil {
+			level.Debug(l).Log("msg", "lookback delta is not zero, setting to default value", "value", defaultLookbackDelta)
```
juliusv: I think you meant:

```suggestion
			level.Debug(l).Log("msg", "lookback delta is zero, setting to default value", "value", defaultLookbackDelta)
```

juliusv:
:+1: besides last comment

roidelapluie:
Fixed your comment @juliusv

juliusv:
Thanks!

PR #6748, opened 2020-02-04
Labels: avoid looping to create a new labelset

roidelapluie:
Looks like this loop has few added value


roidelapluie:
msg: error on input '{} 1 2 3'
        
        exp: labels.Labels{}
        
        got: labels.Labels(nil)

roidelapluie:
So it looks like that loop is only there to avoid nil Labels

PR #6679, opened 2020-01-22
M-map full chunks of Head from disk

codesome:
tl-dr desc for the PR from @krasi-georgiev 

-----------------
When appending to the head and a chunk is full it is flushed to the disk and m-mapped (memory mapped) to free up memory

Prom startup now happens in these stages
 - Iterate the m-maped chunks from disk and keep a map of series reference to its slice of mmapped chunks.
- Iterate the WAL as usual. Whenever we create a new series, look for it's mmapped chunks in the map created before and add it to that series.

If a head chunk is corrupted the currpted one and all chunks after that are deleted and the data after the corruption is recovered from the existing WAL which means that a corruption in m-mapped files results in NO data loss.

[Mmaped chunks format](https://github.com/prometheus/prometheus/blob/master/tsdb/docs/format/head_chunks.md)  - main difference is that the chunk for mmaping now also includes series reference because there is no index for mapping series to chunks.
[The block chunks](https://github.com/prometheus/prometheus/blob/master/tsdb/docs/format/chunks.md) are accessed from the index which includes the offsets for the chunks in the chunks file - example - chunks of series ID have offsets 200, 500 etc in the chunk files.
In case of mmaped chunks, the offsets are stored in memory and accessed from that. During WAL replay, these offsets are restored by iterating all m-mapped chunks as stated above by matching the series id present in the chunk header and offset of that chunk in that file.

**Prombench results**

_WAL Replay_

1h Wal reply time
30% less wal reply time - 4m31 vs 3m36
2h Wal reply time
20% less wal reply time - 8m16 vs 7m

_Memory During WAL Replay_

[High Churn](http://prombench.prometheus.io/grafana/d/7gmLoNDmz/prombench?orgId=1&var-RuleGroup=All&var-pr-number=6679&from=1588177559275&to=1588182978872)	
10-15% less RAM -  32gb vs 28gb
20% less RAM after compaction 34gb vs 27gb
[No Churn](http://prombench.prometheus.io/grafana/d/7gmLoNDmz/prombench?orgId=1&var-RuleGroup=All&var-pr-number=6679&from=1588199791787&to=1588201716740)
20-30% less RAM -  23gb vs 18gb
40% less RAM after compaction 32.5gb vs 20gb

Screenshots are in [this comment](https://github.com/prometheus/prometheus/pull/6679#issuecomment-621678932)

-----------------------

Prerequisite: https://github.com/prometheus/prometheus/pull/6830 (Merged)

[...]

PR #6752, opened 2020-02-05
Add WAL interface and use noopWAL in all tests

codesome:
This is a broken down piece of https://github.com/prometheus/prometheus/pull/6679

1. The interface is required so that we can pass a `noopWAL` to head.
2. We need WAL for all the Head instances as the mmapped chunks logic depends on the dir that it gets from the WAL. I had not added the logic of assigning the noopWAL inside `NewHead` as the temp dir has to come from tests, and doing it like this will avoid passing both dir and WAL in `NewHead`.

### DIFF
--- a/tsdb/block_test.go
+++ b/tsdb/block_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/prometheus/prometheus/tsdb/chunks"
 	"github.com/prometheus/prometheus/tsdb/fileutil"
 	"github.com/prometheus/prometheus/tsdb/tsdbutil"
+	"github.com/prometheus/prometheus/tsdb/wal"
 	"github.com/prometheus/prometheus/util/testutil"
 )
 
@@ -303,7 +304,13 @@ func TestReadIndexFormatV1(t *testing.T) {
 
 // createBlock creates a block with given set of series and returns its dir.
 func createBlock(tb testing.TB, dir string, series []Series) string {
-	return createBlockFromHead(tb, dir, createHead(tb, series))
+	w, close := newTestNoopWal(tb)
+	defer close()
+	head := createHead(tb, series, w)
+	defer func() {
+		testutil.Ok(tb, head.Close())
+	}()
+	return createBlockFromHead(tb, dir, head)
 }
 
 func createBlockFromHead(tb testing.TB, dir string, head *Head) string {
@@ -319,10 +326,9 @@ func createBlockFromHead(tb testing.TB, dir string, head *Head) string {
 	return filepath.Join(dir, ulid.String())
 }
 
-func createHead(tb testing.TB, series []Series) *Head {
-	head, err := NewHead(nil, nil, nil, 2*60*60*1000, DefaultStripeSize)
+func createHead(tb testing.TB, series []Series, w wal.WAL) *Head {
+	head, err := NewHead(nil, nil, w, 2*60*60*1000, DefaultStripeSize)
 	testutil.Ok(tb, err)
-	defer head.Close()
 
 	app := head.Appender()
 	for _, s := range series {
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -542,7 +542,7 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db 
 	}
 	db.compactCancel = cancel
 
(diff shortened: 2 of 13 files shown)


tsdb/block_test.go
```diff
@@ -303,7 +304,13 @@ func TestReadIndexFormatV1(t *testing.T) {
...
 // createBlock creates a block with given set of series and returns its dir.
 func createBlock(tb testing.TB, dir string, series []Series) string {
-	return createBlockFromHead(tb, dir, createHead(tb, series))
+	w, close := newTestNoopWal(tb)
+	defer close()
+	head := createHead(tb, series, w)
+	defer func() {
+		testutil.Ok(tb, head.Close())
+	}()
+	return createBlockFromHead(tb, dir, head)
```
codesome: This refactoring is to close the Head only after the block is created. Closing first and then accessing the data will cause problem for mmapped chunks.

bwplotka: This is quite complex. Why `createHead` cannot close it properly?

tsdb/db.go
```diff
@@ -360,7 +360,8 @@ func (db *DBReadOnly) Querier(mint, maxt int64) (Querier, error) {
 		blocks[i] = b
 	}
 
-	head, err := NewHead(nil, db.logger, nil, 1, DefaultStripeSize)
+	w := wal.NewNoopWAL(db.dir)
```
bwplotka: Not a fan of this.. I think we can make it proper inside it.

tsdb/block_test.go
codesome: If `createHead` closes the Head, then the mmap chunk reader is also closed and we can no longer read the mmapped chunks again. Hence we need to close the Head only after finishing all the operations with the Head.

tsdb/db.go
codesome: Head does not take any explicit directory currently, so taking care of it in Head would mean passing an additional directory along with WAL. But thinking about it again, at this particular place, we dont need a WAL/mmapped chunks at all, so I will remove it.

tsdb/block_test.go
```diff
@@ -303,7 +304,13 @@ func TestReadIndexFormatV1(t *testing.T) {
 
 // createBlock creates a block with given set of series and returns its dir.
 func createBlock(tb testing.TB, dir string, series []Series) string {
-	return createBlockFromHead(tb, dir, createHead(tb, series))
+	w, close := newTestNoopWal(tb)
```
bwplotka: If it's noop and also test WAL, why even need to close it then? ;p

tsdb/wal/wal.go
```diff
@@ -886,6 +910,25 @@ func (r *segmentBufReader) Read(b []byte) (n int, err error) {
...
 // Computing size of the WAL.
 // We do this by adding the sizes of all the files under the WAL dir.
-func (w *WAL) Size() (int64, error) {
+func (w *WALImplementation) Size() (int64, error) {
 	return fileutil.DirSize(w.Dir())
 }
+
+// NewNoopWAL returns a no-op WAL.
+func NewNoopWAL(dir string) WAL {
+	return &noopWAL{dir: dir}
```
bwplotka: Do we ned dir here? Why not just using `""`

tsdb/block_test.go
codesome: `close` is for cleaning up the temporary dir used for the WAL. WAL is noop, but the chunks will be written to the disk by the Head in the mmap PR :)

tsdb/wal/wal.go
codesome: So in https://github.com/prometheus/prometheus/pull/6679 (the mmap chunk PR), the directory for the chunks is taken from the WAL. In tests, that leads to a lot of dirs left behind if we don't pass a temp dir. Hence a dir here.

codesome:
After some offline discussion with @bwplotka, we came to a conclusion that passing the chunks dir as an argument to `NewHead` (in https://github.com/prometheus/prometheus/pull/6679) would be better than taking the dir from the WAL which required all this refactoring in this PR.

As adding the new argument doesn't make any sense right now, I am close this PR and will make the changes in https://github.com/prometheus/prometheus/pull/6679 itself.

PR #6753, opened 2020-02-05
Add OpenMmapFileWithSize method

codesome:
This is a broken down piece of #6679

`OpenMmapFileWithSize` will help open a file with an arbitrary size. This is required for the file in which chunks are being appended in the PR mentioned above.


roidelapluie:
According to our standards the name should be OpenMMAPFileWithSize

roidelapluie:
But I see we did not use that for openMMAP...

codesome:
I guess `Mmap` is fine here

PR #6755, opened 2020-02-05
WIP: go mod compatible

sylr:
This is a attempt to make prometheus go module compatible within the v2 branch.

This is not for merging.

I'm only creating the PR to use the circle-ci test suites.


roidelapluie:
Then we should click the right button in circle ci orbs to allow you to run that only in your forks.

sylr:
I made it work with my circle-ci account but it wouldn't run the `test_windows` job because I'm out of credits.

gouthamve:
Hi @sylr, we looked at this in the bugscrub and as per previous discussions around Go mod, we decided not to go down this route.

Issue #5862, opened 2019-08-02
tsdb: Avoid chunks with >120 samples in MergeOverlappingChunks


codesome:
Currently, during vertical compaction, we directly merge the overlapping chunks and the samples can exceed the limit of 120 in a chunk. Chunk needs to be broken down into smaller chunks if it crosses 120 samples.

The piece of code where the fix goes: https://github.com/prometheus/tsdb/blob/d5b3f0704379a9eaca33b711aa0097f001817fc2/chunks/chunks.go#L208-L240

nidhidhamnani:
I would like to work on this issue

codesome:
@nidhidhamnani go ahead!

Sudhar287:
Can you please elaborate on why this modification is necessary?

bwplotka:
Sure. @Sudhar287 the reason is that our compression algorithm at the moment is designed to have the best compression ration statistically with 120 samples. Anything more than that is introducing high latency (and memory used) for querying because of decoding (time, cpu and mem) without really improving stored size. 

That's why stick to max 120.

Sudhar287:
Okay understood. Thanks @bwplotka

brancz:
I think max 120 alone might not really reflect the entirety of what we want. At the same time we want to avoid chunks that have a low amount of samples as they extrapolated in the long run cause larger disk space and thus space that needs to be mapped into memory. I think we need heuristics for both lower and upper bound merge decisions.

codesome:
@brancz yes, there is another issue tracking the merging of small chunks https://github.com/prometheus/prometheus/issues/6332

bboreham:
Is there a paper somewhere that details how 120 was obtained as the right figure?

brancz:
That would be the gorilla paper: https://www.vldb.org/pvldb/vol8/p1816-teller.pdf

That said while we know the compression statistically gets optimal there, I’m not sure 120 samples exactly the right heuristic here. As I prefer to have a buffer than slip into likely not optimal space.

hdost:
So it looks like this is some sort of cursed issue. There have been multiple PRs all of which have been closed. Is the consensus still that the samples should be split into separate chunks?

codesome:
@hdost Consensus, yes. They were closed because of different reasons :). More than splitting, it's more about avoiding bigger chunks when merging (we can leave alone the already bigger chunks, not worth breaking them down now for the additional complexity that adds).

[...]

PR #6760, opened 2020-02-05
added chunk breaking functionality

Sudhar287:
Breaking a chunk if its greater than a particular size. The best compression ratio is with a maximum of 120 samples per chunk.

With respect to [this](https://github.com/prometheus/prometheus/issues/5862) issue.

### DIFF
--- a/tsdb/chunks/chunks.go
+++ b/tsdb/chunks/chunks.go
@@ -22,6 +22,7 @@ import (
 	"hash/crc32"
 	"io"
 	"io/ioutil"
+	"math"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -51,6 +52,10 @@ const (
 	MaxChunkLengthFieldSize = binary.MaxVarintLen32
 	// ChunkEncodingSize defines the size of the chunk encoding part.
 	ChunkEncodingSize = 1
+	// MaxSamplesInChunk defines the maximum number of samples that should be in a Chunk.
+	// 120 is used becuase it results in the best compression ratio. Anything more will
+	// introduce higher latency while decoding.
+	MaxSamplesInChunk = 120
 )
 
 // Meta holds information about a chunk of data.
@@ -241,34 +246,31 @@ func MergeOverlappingChunks(chks []Meta) ([]Meta, error) {
 	}
 	newChks := make([]Meta, 0, len(chks)) // Will contain the merged chunks.
 	newChks = append(newChks, chks[0])
-	last := 0
 	for _, c := range chks[1:] {
 		// We need to check only the last chunk in newChks.
-		// Reason: (1) newChks[last-1].MaxTime < newChks[last].MinTime (non overlapping)
-		//         (2) As chks are sorted w.r.t. MinTime, newChks[last].MinTime < c.MinTime.
-		// So never overlaps with newChks[last-1] or anything before that.
-		if c.MinTime > newChks[last].MaxTime {
+		// Reason: (1) newChks[len(newChks)-2].MaxTime < newChks[len(newChks)-1].MinTime (non overlapping)
+		//         (2) As chks are sorted w.r.t. MinTime, newChks[len(newChks)-1].MinTime < c.MinTime.
+		// So never overlaps with newChks[len(newChks)-1-1] or anything before that.
+		if c.MinTime > newChks[len(newChks)-1].MaxTime {
 			newChks = append(newChks, c)
-			last++
 			continue
 		}
-		nc := &newChks[last]
-		if c.MaxTime > nc.MaxTime {
-			nc.MaxTime = c.MaxTime
-		}
-		chk, err := MergeChunks(nc.Chunk, c.Chunk)
+		dividedChks, err := MergeChunks(newChks[len(newChks)-1].Chunk, c.Chunk)
 		if err != nil {
 			return nil, err
 		}
-		nc.Chunk = chk
+		// Appending the first len(newChks)-2 chunks with the divided chunks.
+		// The len(newChks)-1 th chunk would have been merged and hence is overwritten.
+		newChks = append(newChks[:len(newChks)-1], dividedChks...)
 	}
-
 	return newChks, nil
 }
 
 // MergeChunks vertically merges a and b, i.e., if there is any sample
 // with same timestamp in both a and b, the sample in a is discarded.
-func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
+// Also breaks chunks that are larger than MaxSamplesInChunk.
+func MergeChunks(a, b chunkenc.Chunk) ([]Meta, error) {
+	var newMetas []Meta
 	newChunk := chunkenc.NewXORChunk()
 	app, err := newChunk.Appender()
 	if err != nil {
@@ -276,31 +278,57 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
 	}
 	ait := a.Iterator(nil)
 	bit := b.Iterator(nil)
+	numSamples := 0
+	var mint, maxt int64
+	mint = int64(math.MinInt64)
+	var at, bt int64
+	var av, bv float64
 	aok, bok := ait.Next(), bit.Next()
-	for aok && bok {
-		at, av := ait.At()
-		bt, bv := bit.At()
+	for {
+		if !aok && !bok {
+			break
+		} else if !aok {
+			at = math.MaxInt64
+			bt, bv = bit.At()
+		} else if !bok {
+			bt = math.MaxInt64
+			at, av = ait.At()
+		} else {
+			at, av = ait.At()
+			bt, bv = bit.At()
+		}
 		if at < bt {
 			app.Append(at, av)
+			maxt = at
 			aok = ait.Next()
 		} else if bt < at {
 			app.Append(bt, bv)
+			maxt = bt
 			bok = bit.Next()
 		} else {
 			app.Append(bt, bv)
+			maxt = bt
 			aok = ait.Next()
 			bok = bit.Next()
 		}
-	}
-	for aok {
-		at, av := ait.At()
-		app.Append(at, av)
-		aok = ait.Next()
-	}
-	for bok {
-		bt, bv := bit.At()
-		app.Append(bt, bv)
-		bok = bit.Next()
+		if mint == math.MinInt64 {
+			if at < bt {
+				mint = at
+			} else {
+				mint = bt
+			}
+		}
+		numSamples++
+		if numSamples >= MaxSamplesInChunk {
+			newMetas = append(newMetas, Meta{Chunk: newChunk, MinTime: mint, MaxTime: maxt})
+			newChunk = chunkenc.NewXORChunk()
+			app, err = newChunk.Appender()
+			if err != nil {
+				return nil, err
+			}
+			numSamples = 0
+			mint = int64(math.MinInt64)
+		}
 	}
 	if ait.Err() != nil {
 		return nil, ait.Err()
@@ -308,7 +336,10 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
 	if bit.Err() != nil {
 		return nil, bit.Err()
 	}
-	return newChunk, nil
+	if newChunk.NumSamples() != 0 {
+		newMetas = append(newMetas, Meta{Chunk: newChunk, MinTime: mint, MaxTime: maxt})
+	}
+	return newMetas, nil
 }
 
 // WriteChunks writes as many chunks as possible to the current segment,
--- a/tsdb/compact_test.go
+++ b/tsdb/compact_test.go
@@ -737,6 +737,126 @@ func TestCompaction_populateBlock(t *testing.T) {
 				},
 			},
 		},
+		{
+			title: "Break chunks that are greater than 120 in size.",
+			inputSeriesSamples: [][]seriesSamples{
+				{
+					{
+						lset: map[string]string{"a": "b"},
+						chunks: func() [][]sample {
+							chunkOne := make([]sample, 120)
+							for i := 0; i < 120; i++ {
+								chunkOne[i] = sample{t: int64(i + 1)}
+							}
+							chunkTwo := make([]sample, 42)
+							for i := 0; i < 42; i++ {
+								chunkTwo[i] = sample{t: int64(i + 1 + 200)}
+							}
+							chunkThree := make([]sample, 70)
+							for i := 0; i < 70; i++ {
+								chunkThree[i] = sample{t: int64(i + 1 + 300)}
+							}
+							chunkFour := make([]sample, 120)
+							for i := 0; i < 120; i++ {
+								chunkFour[i] = sample{t: int64(i + 1 + 500)}
+							}
+							chunkFive := make([]sample, 120) // tricky test case of 3 chunks overlapping 1:1 with more than 120 samples
+							for i := 0; i < 60; i++ {
+								chunkFive[i] = sample{t: int64(i + 1 + 940)}
+							}
+							for i, j := 60, 0; i < 70; i, j = i+1, j+1 {
+								chunkFive[i] = sample{t: int64(j + 1 + 1100)}
+							}
+							for i, j := 70, 0; i < 120; i, j = i+1, j+1 {
+								chunkFive[i] = sample{t: int64(j + 1 + 1220)}
+							}
+							chunks := [][]sample{chunkOne, chunkTwo, chunkThree, chunkFour, chunkFive}
+							return chunks
+						}(),
+					},
+				},
+				{
+					{
+						lset: map[string]string{"a": "b"},
+						chunks: func() [][]sample {
+
+							chunkOne := make([]sample, 70)
+							for i := 0; i < 70; i++ {
+								chunkOne[i] = sample{t: int64(i + 1 + 300 + 60)}
+							}
+							chunkTwo := make([]sample, 120)
+							for i := 0; i < 120; i++ {
+								chunkTwo[i] = sample{t: int64(i + 1 + 500 + 118)}
+							}
+							chunkThree := make([]sample, 120)
+							for i := 0; i < 120; i++ {
+								chunkThree[i] = sample{t: int64(i + 1 + 1000)}
+							}
+							chunks := [][]sample{chunkOne, chunkTwo, chunkThree}
+							return chunks
+						}(),
+					},
+				},
+				{
+					{
+						lset: map[string]string{"a": "b"},
+						chunks: func() [][]sample {
+							chunkOne := make([]sample, 120)
+							for i := 0; i < 120; i++ {
+								chunkOne[i] = sample{t: int64(i + 1 + 1100)}
+							}
+							chunks := [][]sample{chunkOne}
+							return chunks
+						}(),
+					},
+				},
+			},
+			expSeriesSamples: []seriesSamples{
+				{
+					lset: map[string]string{"a": "b"},
+					chunks: func() [][]sample {
+						chunkOne := make([]sample, 120)
+						for i := 0; i < 120; i++ {
+							chunkOne[i] = sample{t: int64(i + 1)}
+						}
+						chunkTwo := make([]sample, 42)
+						for i := 0; i < 42; i++ {
+							chunkTwo[i] = sample{t: int64(i + 1 + 200)}
+						}
+						chunkThree := make([]sample, 120)
+						for i := 0; i < 120; i++ {
+							chunkThree[i] = sample{t: int64(i + 1 + 300)}
+						}
+						chunkFour := make([]sample, 10)
+						for i := 0; i < 10; i++ {
+							chunkFour[i] = sample{t: int64(i + 1 + 300 + 120)}
+						}
+						chunkFive := make([]sample, 120)
+						for i := 0; i < 120; i++ {
+							chunkFive[i] = sample{t: int64(i + 1 + 500)}
+						}
+						chunkSix := make([]sample, 118)
+						for i := 0; i < 118; i++ {
+							chunkSix[i] = sample{t: int64(i + 1 + 620)}
+						}
+						chunkSeven := make([]sample, 120)
+						for i := 0; i < 120; i++ {
+							chunkSeven[i] = sample{t: int64(i + 1 + 940)}
+						}
+						chunkEight := make([]sample, 120)
+						for i := 0; i < 120; i++ {
+							chunkEight[i] = sample{t: int64(i + 1 + 1060)}
+						}
+						chunkNine := make([]sample, 90)
+						for i := 0; i < 90; i++ {
+							chunkNine[i] = sample{t: int64(i + 1 + 1180)}
+						}
+						chunks := [][]sample{chunkOne, chunkTwo, chunkThree, chunkFour, chunkFive, chunkSix, chunkSeven, chunkEight, chunkNine}
+						return chunks
+					}(),
+				},
+			},
+		},
 	}
 
 	for _, tc := range populateBlocksCases {


tsdb/chunks/chunks.go
```diff
@@ -262,7 +264,10 @@ func MergeOverlappingChunks(chks []Meta) ([]Meta, error) {
 		}
 		nc.Chunk = chk
 	}
-
+	newChks, err := BreakLargeChunks(newChks, MaxSamplesInChunk)
```
bwplotka: Nice! All looks good but I am pretty certain it would be more efficient do it in `MergeChunks` instead. We can reuse exactly the same loop iteration (:

tsdb/chunks/chunks.go
```diff
@@ -311,6 +316,48 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
 	return newChunk, nil
 }
 
+// BreakLargeChunks splits a chunk into many if it has more than a certain number of samples.
+func BreakLargeChunks(chks []Meta, limit int) ([]Meta, error) {
```
bwplotka: Does it have to be public?

tsdb/compact_test.go
```diff
@@ -737,6 +737,29 @@ func TestCompaction_populateBlock(t *testing.T) {
 				},
 			},
 		},
+		{
+			title: "Break chunks that are greater than 120 in size.",
+			inputSeriesSamples: [][]seriesSamples{
+				{
```
bwplotka: You can increase readability by removing some of newlines

tsdb/compact_test.go
```diff
@@ -786,6 +809,57 @@ func TestCompaction_populateBlock(t *testing.T) {
 	}
 }
 
+func GetOversizedChunk() [][]sample {
```
bwplotka: I would just inline those with proper comment and definitely make it private.

Sudhar287: Sure. Thanks for the review.

tsdb/chunks/chunks.go
Sudhar287: No it doesn't. I'll change it. Thanks for the review.

tsdb/chunks/chunks.go
Sudhar287: Hello bwplotka. Thanks for taking the time to review my code.
Can you please provide some more added clarification about your comment here?
I think what you have pointed out is that we are iterating through the []Meta once to just do the `MergeChunks` and then once again inside the function I wrote. Is your proposal to change this nature of the code? 

Including my function within `MergeChunks` would require changes to the function signature itself. It is a public function and I'm not too sure if thats what we want to achieve here.

tsdb/compact_test.go
Sudhar287: I adhered to the styling that was in all the previous test cases. Does this change require modifying the style of all the previous test cases?

tsdb/compact_test.go
Sudhar287: Hey bwplotka can you please clarify by what you mean when you say inline?
Do you mean something like this? `for i := 0; i < 200; i++ { chunkOne[i] = sample{t: int64(i + 1)}}`. This wont go through because gofmt will change the style anyways...

tsdb/compact_test.go
bwplotka: Ideally yes. See https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html

However it is too complex, affect too many things feel free to ignore my comment, it's a minor nit.

tsdb/compact_test.go
bwplotka: I meant that I would suggest copying the whole thing to the place you use it. It can be multiline, but having a separate function just for one usage is anti-pattern.

You can use anonymous functions `func() { ... }`

tsdb/chunks/chunks.go
bwplotka: > I think what you have pointed out is that we are iterating through the []Meta once to just do the MergeChunks and then once again inside the function I wrote. Is your proposal to change this nature of the code?

Yes.

> Including my function within MergeChunks would require changes to the function signature itself. It is a public function and I'm not too sure if thats what we want to achieve here.

This is fine. As Prometheus, we don't guarantee stability of internal package API. All of our users user versioned package of Prometheus (e.g project I maintain: [Thanos](http://thanos.io/)). Someday we might want proper semantic versioning for our packages, but not for now.

So it's totally fine to change signature. If you would silently change logic / behaviour, that would be worse. So go for it :+1:

Sudhar287: Okay Sure. Your comments have been incorporated in my latest PR. Please do have a look when possible.

tsdb/chunks/chunks.go
```diff
@@ -308,7 +318,66 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
 	if bit.Err() != nil {
 		return nil, bit.Err()
 	}
-	return newChunk, nil
+	newChks, err := breakLargeChunks(newChunk, MaxSamplesInChunk)
```
bwplotka: I think what you did is essentially moved this function to inside the `MergeChunk`. 

What about actually checking the num samples while appending to the new chunk in `for aok && bok {` loop?

I think we can avoid building the same chunk twice (!)

codesome: +1 on this, build 2 chunks in the fly.

tsdb/chunks/chunks.go
```diff
@@ -256,19 +258,27 @@ func MergeOverlappingChunks(chks []Meta) ([]Meta, error) {
...
 	}
-
 	return newChks, nil
 }
 
 // MergeChunks vertically merges a and b, i.e., if there is any sample
 // with same timestamp in both a and b, the sample in a is discarded.
-func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
+// Also breaks chunks that are larger than a particular limit.
+func MergeChunks(a, b chunkenc.Chunk) ([]chunkenc.Chunk, error) {
```
codesome: We can expect at most 2 chunks as a result. So what about having 2 chunks being returned instead of a slice? It would also simplify the logic in `MergeOverlappingChunks`.

bwplotka:
@codesome Please next time be open and clear that she is working on it. The truth is that we couldn't find any related work done after a month of her initial comment that  @nidhidhamnani so @Sudhar287 assumed it the work got stale. Just commenting "I am on it" and not actually doing anything outside can lead to issues like this (: 

E.g where we can find @nidhidhamnani PR? 

Anyway, super happy this being solved, should not matter who does it first, it's not a competition (: 
However, we could work on communication for sure to avoid duplicated work @codesome (: Still there was some learning involved so nothing is wasted (: 

Looks like @codesome you are on top of this, so it's up to you want the version to choose, just please be public and open about all of it. It's open-source after all :hugs:

codesome:
I have no issues with whichever gets merged :sweat_smile: . We can go ahead with this PR as it was opened first. I was just trying to follow the [contributing guidelines](https://github.com/prometheus/prometheus/blob/master/CONTRIBUTING.md#steps-to-contribute) that we have, and as @Sudhar287's comment didn't feel like he was working on it (else I would have intimated there itself that she is working on it :slightly_smiling_face:, not hiding anything), I was trying to be fair with @nidhidhamnani as she is new to Go and open source in general and had been putting efforts in limited free time since claiming that issue. And yes, we could work on some communication :)

bwplotka:
Cool 👍 Thanks @codesome! Both contributors are new to Go, Prometheus and open source so it's up to us to guide them.

nidhidhamnani:
Hi, I had less free time in the past 1 month and most of the time went into understanding Prometheus, though I have the fix now, hence it got delayed, sorry! I will pick some other issue then :)

Sudhar287:
Hello reviewers @bwplotka and @codesome! Is my PR not being looked into because its failed circle CI? 
I ran it on my local machine and there were no problems with the test cases that are failing online. Do you have any advice on what I should so next?

codesome:
There were some discussions on #prometheus-dev chat if we actually require this as apparently beyond 120 samples though there is no additional benefit, there is no harm too. @bwplotka had some thoughts around small chunks being beneficial to Thanos, did you come to any conclusion?

bwplotka:
Yes, I still think we need this. There is harm in too long chunks IMO, that's why I would accept this PR.

Reasons why we should split huge chunks to max 120 samples:
* To get samples for short period (e.g minutes, hours or days for downsampled data), we need to decode whole chunk. Increased size just makes it harder, introduces more latency and CPU time. Especially for downsampling.
* It helps maintain deterministic characteristics of the chunk.

Overall especially in Thanos and Cortex we will lean more on vertical compaction so it gets more important.

tsdb/chunks/chunks.go
```diff
@@ -51,6 +51,8 @@ const (
 	MaxChunkLengthFieldSize = binary.MaxVarintLen32
 	// ChunkEncodingSize defines the size of the chunk encoding part.
 	ChunkEncodingSize = 1
+	// MaxSamplesInChunk defines the maximum number of samples that should be in a Chunk.
```
bwplotka: We should elaborate more why then (: We can put my arguments from comment.

tsdb/chunks/chunks.go
```diff
@@ -252,63 +254,131 @@ func MergeOverlappingChunks(chks []Meta) ([]Meta, error) {
 			last++
 			continue
 		}
-		nc := &newChks[last]
-		if c.MaxTime > nc.MaxTime {
-			nc.MaxTime = c.MaxTime
-		}
-		chk, err := MergeChunks(nc.Chunk, c.Chunk)
+		dividedChunkOne, dividedChunkTwo, err := MergeChunks(newChks[last].Chunk, c.Chunk, MaxSamplesInChunk)
```
bwplotka:
```suggestion
		chk1, chk2, err := MergeChunks(newChks[last].Chunk, c.Chunk, MaxSamplesInChunk)
```
Is fine really (:

bwplotka: Also.. why two?  I would expect slice of chunks TBH.

tsdb/chunks/chunks.go
```diff
@@ -252,63 +254,131 @@ func MergeOverlappingChunks(chks []Meta) ([]Meta, error) {
...
+	appPtr := &app
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	ait := a.Iterator(nil)
 	bit := b.Iterator(nil)
 	aok, bok := ait.Next(), bit.Next()
+	numSamples := 0
+	appendAndReset := func(newMetaArrayPtr *[]Meta, newChunk *chunkenc.XORChunk, mint int64, maxt int64) (*chunkenc.XORChunk, *chunkenc.Appender, error) {
```
bwplotka: this is way to complex. Also why not making it separate function then?

tsdb/chunks/chunks.go
```diff
@@ -252,63 +254,131 @@ func MergeOverlappingChunks(chks []Meta) ([]Meta, error) {
...
+			}
+			numSamples = 0
+			first = true
+		}
 		aok = ait.Next()
 	}
 	for bok {
 		bt, bv := bit.At()
-		app.Append(bt, bv)
+		(*appPtr).Append(bt, bv)
```
bwplotka: This code is repeating a lot. I think we can simplify it more.

tsdb/chunks/chunks.go
Sudhar287: because you had told me to use anonymous functions [before](https://github.com/prometheus/prometheus/pull/6760#discussion_r376163601) too.
P.S. Thanks for the review!

tsdb/chunks/chunks.go
bwplotka: In this sense I don't agree with @codesome here: https://github.com/prometheus/prometheus/pull/6760#discussion_r376347819

tsdb/chunks/chunks.go
bwplotka: Well, don't take everything so explicit. Try to have your own common sense here (: If something is readable or not (:

tsdb/chunks/chunks.go
Sudhar287: Also, do I need to write test cases to include the event if users change the chunk size to what they want (instead of using a fixed 120)?

codesome: I was assuming that we will have max 120 samples in the incoming chunk. But yes, that is not necessary. But where do we expect way more than 120 samples? :)
It would be interesting to see if there are any performance implications for returning a slice. If it is very less, then it's a no brainer, else would want to consider tolerating a small spillover for the 2nd chunk.

codesome: I was assuming that we will have max 120 samples in the incoming chunk. But yes, that is not necessary. But where do we expect way more than 120 samples? :)
It would be interesting to see if there are any performance implications for returning a slice. If it is very less, then it's a no brainer, else would want to consider tolerating a small spillover for the 2nd chunk.

codesome: > Also, do I need to write test cases to include the event if users change the chunk size to what they want (instead of using a fixed 120)?

I don't think so. It is not user configurable.

Sudhar287:
Hey @bwplotka! I hope you're doing well! Please do look into my CR when possible!

tsdb/chunks/chunks.go
```diff
@@ -252,34 +257,58 @@ func MergeOverlappingChunks(chks []Meta) ([]Meta, error) {
 			last++
```
bwplotka: We don't need this variable we can just use `len(newChks)-1` right?

tsdb/chunks/chunks.go
```diff
@@ -252,34 +257,58 @@ func MergeOverlappingChunks(chks []Meta) ([]Meta, error) {
...
-		if c.MaxTime > nc.MaxTime {
-			nc.MaxTime = c.MaxTime
-		}
-		chk, err := MergeChunks(nc.Chunk, c.Chunk)
+		dividedChks, err := MergeChunks(newChks[last].Chunk, c.Chunk, MaxSamplesInChunk)
 		if err != nil {
 			return nil, err
 		}
-		nc.Chunk = chk
+		newChks[last] = dividedChks[0]
```
bwplotka: All of this can be replaced with single line (: 

```
newChks = append(newChks[:len(newChks)-1], dividedChks...)
```

tsdb/chunks/chunks.go
```diff
@@ -252,34 +257,58 @@ func MergeOverlappingChunks(chks []Meta) ([]Meta, error) {
 			last++
 			continue
 		}
-		nc := &newChks[last]
-		if c.MaxTime > nc.MaxTime {
-			nc.MaxTime = c.MaxTime
-		}
-		chk, err := MergeChunks(nc.Chunk, c.Chunk)
+		dividedChks, err := MergeChunks(newChks[last].Chunk, c.Chunk, MaxSamplesInChunk)
```
bwplotka: I think we are fine, but worth to add some tricky cases of 3 chunks overlapping 1:1 with more than 120 samples to our tests?

To check if we are still working well with the comment from line 251.
```
// We need to check only the last chunk in newChks.
// Reason: (1) newChks[last-1].MaxTime < newChks[last].MinTime (non overlapping)
//         (2) As chks are sorted w.r.t. MinTime, newChks[last].MinTime < c.MinTime.
// So never overlaps with newChks[last-1] or anything before that.
```

tsdb/chunks/chunks.go
```diff
@@ -252,34 +257,58 @@ func MergeOverlappingChunks(chks []Meta) ([]Meta, error) {
...
-
 	return newChks, nil
 }
 
 // MergeChunks vertically merges a and b, i.e., if there is any sample
 // with same timestamp in both a and b, the sample in a is discarded.
-func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
+// Also breaks chunks that are larger than a particular limit.
+func MergeChunks(a, b chunkenc.Chunk, limit int) ([]Meta, error) {
+	newMetaArray := make([]Meta, 0)
```
bwplotka: This is not needed if we don't allocate any  memory

tsdb/chunks/chunks.go
Sudhar287: Yes. I was trying to adhere to the existing code but Okay! :)

tsdb/chunks/chunks.go
Sudhar287: Yes, you are right. Thank you!

tsdb/chunks/chunks.go
Sudhar287: Each chunk from upstream will not have more than 120 samples right?
I'm going to assume you meant 1:1 overlapping for a number of samples and then a non-overlapping extension (which would have to be broken because it will be greater than 120). If I'm wrong then please let me know. :)

bwplotka: It can have more than 120 as you just create code that splits. Before that people were merging chunks to XXXXX samples without any logic, right? ;p

bwplotka: > I'm going to assume you meant 1:1 overlapping for a number of samples and then a non-overlapping extension (which would have to be broken because it will be greater than 120). 

I mean 3x120 samples block for times [0  10] [0 10] [0 10] from 3 different blocks. (:

Sudhar287: Done. Please have a look when you're available :)

tsdb/chunks/chunks.go
```diff
@@ -241,45 +246,63 @@ func MergeOverlappingChunks(chks []Meta) ([]Meta, error) {
...
-			nc.MaxTime = c.MaxTime
-		}
-		chk, err := MergeChunks(nc.Chunk, c.Chunk)
+		dividedChks, err := MergeChunks(newChks[len(newChks)-1].Chunk, c.Chunk, MaxSamplesInChunk)
 		if err != nil {
 			return nil, err
 		}
-		nc.Chunk = chk
+		newChks = append(newChks[:len(newChks)-1], dividedChks...)
+		// Appending the first len(newChks)-2 chunks with the divided chunks.
```
bwplotka: Usually we comment before thing is happening.

tsdb/chunks/chunks.go
```diff
@@ -241,45 +246,63 @@ func MergeOverlappingChunks(chks []Meta) ([]Meta, error) {
...
 	}
-
 	return newChks, nil
 }
 
 // MergeChunks vertically merges a and b, i.e., if there is any sample
 // with same timestamp in both a and b, the sample in a is discarded.
-func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
+// Also breaks chunks that are larger than a particular limit.
+func MergeChunks(a, b chunkenc.Chunk, limit int) ([]Meta, error) {
```
bwplotka: do we really need `limit` variable?

tsdb/chunks/chunks.go
```diff
@@ -291,24 +314,33 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
...
 	}
 	if ait.Err() != nil {
 		return nil, ait.Err()
 	}
 	if bit.Err() != nil {
 		return nil, bit.Err()
 	}
-	return newChunk, nil
+	if newChunk.NumSamples() != 0 {
+		newMetaArray = append(newMetaArray, Meta{Chunk: newChunk, MinTime: mint, MaxTime: maxt})
```
bwplotka: We know it is array because we all use modern IDEs, so maybe `newMetas`?

Also it's not array, but slice (:

tsdb/chunks/chunks.go
```diff
@@ -291,24 +314,33 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
...
-		bt, bv := bit.At()
-		app.Append(bt, bv)
-		bok = bit.Next()
+		if first {
+			mint = minOf(at, bt)
+			first = false
+		}
+		maxt = minOf(at, bt)
+		numSamples++
+		if numSamples == limit {
```
bwplotka:
```suggestion
		if numSamples >= limit {
```
Is usually a more readable choice

tsdb/chunks/chunks.go
```diff
@@ -241,45 +246,63 @@ func MergeOverlappingChunks(chks []Meta) ([]Meta, error) {
...
 	newChunk := chunkenc.NewXORChunk()
 	app, err := newChunk.Appender()
 	if err != nil {
 		return nil, err
 	}
 	ait := a.Iterator(nil)
 	bit := b.Iterator(nil)
+	numSamples := 0
+	var mint, maxt int64
+	first := true // flag needed to check if looking at the first sample of the chunk and subsequenty set mint
```
bwplotka: Comment is not a full sentence.

bwplotka: I think we don't need this variable as well. What about doing setting `mint` to `math.MinMaxInt64` and relying on that?

tsdb/chunks/chunks.go
```diff
@@ -291,24 +314,33 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
...
-	}
-	for bok {
-		bt, bv := bit.At()
-		app.Append(bt, bv)
-		bok = bit.Next()
+		if first {
+			mint = minOf(at, bt)
+			first = false
+		}
+		maxt = minOf(at, bt)
```
bwplotka: This is weird as we already did `if at < bt` so we already compared at with bt. Can we reuse this fact?

tsdb/chunks/chunks.go
```diff
@@ -291,24 +314,33 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
...
-		at, av := ait.At()
-		app.Append(at, av)
-		aok = ait.Next()
-	}
-	for bok {
-		bt, bv := bit.At()
-		app.Append(bt, bv)
-		bok = bit.Next()
+		if first {
+			mint = minOf(at, bt)
```
bwplotka: same here

tsdb/chunks/chunks.go
```diff
@@ -241,45 +246,63 @@ func MergeOverlappingChunks(chks []Meta) ([]Meta, error) {
...
 		return nil, err
 	}
 	ait := a.Iterator(nil)
 	bit := b.Iterator(nil)
+	numSamples := 0
+	var mint, maxt int64
+	first := true // flag needed to check if looking at the first sample of the chunk and subsequenty set mint
+	var at, bt int64
+	var av, bv float64
+	minOf := func(a, b int64) int64 {
```
bwplotka: I think we don't need this, see comment above

tsdb/chunks/chunks.go
Sudhar287: I thought I would be explicit with the `first` variable. But Okay!

tsdb/chunks/chunks.go
Sudhar287: One of the previous reviews stated that we can give the uses the freedom to change the number of samples in a chunk. But Okay! :)

tsdb/chunks/chunks.go
```diff
@@ -241,74 +246,100 @@ func MergeOverlappingChunks(chks []Meta) ([]Meta, error) {
...
+	var mint, maxt int64
+	mint = int64(math.MinInt64)
+	var at, bt int64
+	var av, bv float64
 	aok, bok := ait.Next(), bit.Next()
-	for aok && bok {
-		at, av := ait.At()
-		bt, bv := bit.At()
+	for {
+		if !aok && !bok {
```
bwplotka: Can we put this in `for` as it was? What's the argument for this?

Sudhar287: It was [this](https://github.com/prometheus/prometheus/pull/6760#discussion_r383771923).

Sudhar287:
Although the news is a bit disappointing, I'm absolutely up for contributing! Thanks for sharing the steps about the task. I'll work on it. :)

PR #6761, opened 2020-02-05
Add conditional rendering of Navlink for Consoles

### LINKED ISSUE #6705: React UI: Implement "Consoles" link
The old UI shows a "Consoles" link in the nav at the top when [user-defined console templates](https://prometheus.io/docs/visualization/consoles/) are present. See for example the top nav at http://demo.robustperception.io:9090/graph.

The new UI does not need any changes regarding serving console templates themselves (this will still be handled by existing functionality in Prometheus that is not part of the "old" UI), but we do need to determine whether console templates are present or not, in order to know whether we should display the "Consoles" link at the top. Here is where that determination was made in the old UI:

https://github.com/prometheus/prometheus/blob/ddd49b743c1c03527064d38f4830258b4a6ccb27/web/ui/templates/_base.html#L44-L47

https://github.com/prometheus/prometheus/blob/ddd49b743c1c03527064d38f4830258b4a6ccb27/web/web.go#L963

There's two ways we could do this in the new UI:

- Add an API call that gives us this info. Drawback is that initially on page load, the "Consoles" link will not be shown, and then it appears.
- Inject a global variable into the index.html that gives us the information we need. We already do this with the global path prefix here: https://github.com/prometheus/prometheus/blob/ddd49b743c1c03527064d38f4830258b4a6ccb27/web/ui/react-app/public/index.html#L17 and here: https://github.com/prometheus/prometheus/blob/ddd49b743c1c03527064d38f4830258b4a6ccb27/web/web.go#L384

I like the second option more because we already have that ugly placeholder pattern anyway, and it immediately gives us the right nav at the top.

drumilpatel2000:
Signed-off-by: Drumil Patel <drumilpatel720@gmail.com>

Fixes #6705 
#### When no files are present in consoles
Screenshot from 2020-02-06 02-26-23
Screenshot from 2020-02-06 02-44-37

#### When Files are added in console
Screenshot from 2020-02-06 02-27-24
Screenshot from 2020-02-06 02-27-11

### DIFF
--- a/web/ui/react-app/public/index.html
+++ b/web/ui/react-app/public/index.html
@@ -10,17 +10,17 @@
     <meta name="theme-color" content="#000000" />
 
     <!--
-      PATH_PREFIX constants placeholder magic value is replaced during serving by Prometheus
+      The GLOBAL_PATH_PREFIX placeholder magic value is replaced during serving by Prometheus
       and set to Prometheus's external URL path. It gets prepended to all links back
       to Prometheus, both for asset loading as well as API accesses.
 
-      CONSOLES_LINK constants placeholder magic value is replaced during serving by Prometheus
-      and set to console link if it exists. It will render Console link on navbar when 
-      it is non empty. 
+      The GLOBAL_CONSOLES_LINK placeholder magic value is replaced during serving by Prometheus
+      and set to the consoles link if it exists. It will render a "Consoles" link in the navbar when 
+      it is non-empty. 
     -->
     <script>
-      const PATH_PREFIX='PATH_PREFIX_PLACEHOLDER';
-      const CONSOLES_LINK='CONSOLES_LINK_PLACEHOLDER';
+      const GLOBAL_PATH_PREFIX='PATH_PREFIX_PLACEHOLDER';
+      const GLOBAL_CONSOLES_LINK='CONSOLES_LINK_PLACEHOLDER';
     </script>
 
     <!--
--- a/web/ui/react-app/public/index.html
+++ b/web/ui/react-app/public/index.html
@@ -10,11 +10,18 @@
     <meta name="theme-color" content="#000000" />
 
     <!--
-      This constant's placeholder magic value is replaced during serving by Prometheus
+      PATH_PREFIX constants placeholder magic value is replaced during serving by Prometheus
       and set to Prometheus's external URL path. It gets prepended to all links back
       to Prometheus, both for asset loading as well as API accesses.
+
+      CONSOLES_LINK constants placeholder magic value is replaced during serving by Prometheus
+      and set to console link if it exists. It will render Console link on navbar when 
+      it is non empty. 
     -->
-    <script>const PATH_PREFIX='PATH_PREFIX_PLACEHOLDER';</script>
+    <script>
+      const PATH_PREFIX='PATH_PREFIX_PLACEHOLDER';
+      const CONSOLES_LINK='CONSOLES_LINK_PLACEHOLDER';
+    </script>
 
     <!--
       manifest.json provides metadata used when your web app is added to the
--- a/web/ui/react-app/src/Navbar.test.tsx
+++ b/web/ui/react-app/src/Navbar.test.tsx
@@ -3,8 +3,8 @@ import { shallow } from 'enzyme';
 import Navigation from './Navbar';
 import { NavItem, NavLink } from 'reactstrap';
 
-describe('Navbar should contains console Link', () => {
-  it('with non-empty consoleslinks', () => {
+describe('Navbar should contain console Link', () => {
+  it('with non-empty consoleslink', () => {
     const app = shallow(<Navigation pathPrefix="/path/prefix" consolesLink="/path/consoles" />);
     expect(
       app.contains(
@@ -16,8 +16,8 @@ describe('Navbar should contains console Link', () => {
   });
 });
 
-describe('Navbar should not contains consoles link', () => {
-  it('with empty string in consolesLinks ', () => {
+describe('Navbar should not contain consoles link', () => {
+  it('with empty string in consolesLink', () => {
     const app = shallow(<Navigation pathPrefix="/path/prefix" consolesLink="" />);
     expect(
       app.contains(
--- a/web/ui/react-app/src/Navbar.test.tsx
+++ b/web/ui/react-app/src/Navbar.test.tsx
@@ -1,0 +1,30 @@
+import * as React from 'react';
+import { shallow } from 'enzyme';
+import Navigation from './Navbar';
+import { NavItem, NavLink } from 'reactstrap';
+
+describe('Navbar should contains console Link', () => {
+  it('with non-empty consoleslinks', () => {
+    const app = shallow(<Navigation pathPrefix="/path/prefix" consolesLink="/path/consoles" />);
+    expect(
+      app.contains(
+        <NavItem>
+          <NavLink href="/path/consoles">Consoles</NavLink>
+        </NavItem>
+      )
+    ).toBeTruthy();
+  });
+});
+
+describe('Navbar should not contains consoles link', () => {
+  it('with empty string in consolesLinks ', () => {
+    const app = shallow(<Navigation pathPrefix="/path/prefix" consolesLink="" />);
+    expect(
+      app.contains(
+        <NavItem>
+          <NavLink>Consoles</NavLink>
+        </NavItem>
+      )
+    ).toBeFalsy();
+  });
+});
--- a/web/ui/react-app/src/Navbar.tsx
+++ b/web/ui/react-app/src/Navbar.tsx
@@ -14,11 +14,11 @@ import {
 } from 'reactstrap';
 import PathPrefixProps from './types/PathPrefixProps';
 
-interface NavbarConsolesLink {
+interface NavbarProps {
   consolesLink: string | null;
 }
 
-const Navigation: FC<PathPrefixProps & NavbarConsolesLink> = ({ pathPrefix, consolesLink }) => {
+const Navigation: FC<PathPrefixProps & NavbarProps> = ({ pathPrefix, consolesLink }) => {
   const [isOpen, setIsOpen] = useState(false);
   const toggle = () => setIsOpen(!isOpen);
   return (
--- a/web/ui/react-app/src/Navbar.tsx
+++ b/web/ui/react-app/src/Navbar.tsx
@@ -1,4 +1,3 @@
-import './globals';
 import React, { FC, useState } from 'react';
 import { Link } from '@reach/router';
 import {
@@ -15,15 +14,7 @@ import {
 } from 'reactstrap';
 import PathPrefixProps from './types/PathPrefixProps';
 
-// Declared/defined in public/index.html, value replaced by Prometheus when serving bundle.
-declare const CONSOLES_LINK: string;
-
-let consoleLink = CONSOLES_LINK;
-if (CONSOLES_LINK === 'CONSOLES_LINK_PLACEHOLDER' || CONSOLES_LINK === '') {
-  consoleLink = '';
-}
-
-const Navigation: FC<PathPrefixProps> = ({ pathPrefix }) => {
+const Navigation: FC<PathPrefixProps> = ({ pathPrefix, consolesLink }) => {
   const [isOpen, setIsOpen] = useState(false);
   const toggle = () => setIsOpen(!isOpen);
   return (
@@ -34,9 +25,9 @@ const Navigation: FC<PathPrefixProps> = ({ pathPrefix }) => {
       </Link>
       <Collapse isOpen={isOpen} navbar style={{ justifyContent: 'space-between' }}>
         <Nav className="ml-0" navbar>
-          {consoleLink !== '' && (
+          {consolesLink !== '' && (
             <NavItem>
-              <NavLink href={`${pathPrefix + consoleLink}`}>Consoles</NavLink>
+              <NavLink href={`${consolesLink}`}>Consoles</NavLink>
             </NavItem>
           )}
           <NavItem>
--- a/web/ui/react-app/src/Navbar.tsx
+++ b/web/ui/react-app/src/Navbar.tsx
@@ -14,11 +14,11 @@ import {
 } from 'reactstrap';
 import PathPrefixProps from './types/PathPrefixProps';
 
-interface ConsolesLinkProps {
+interface NavbarConsolesLink {
   consolesLink: string | null;
 }
 
-const Navigation: FC<PathPrefixProps & ConsolesLinkProps> = ({ pathPrefix, consolesLink }) => {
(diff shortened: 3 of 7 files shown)


web/ui/react-app/src/Navbar.tsx
```diff
@@ -14,6 +15,14 @@ import {
 } from 'reactstrap';
 import PathPrefixProps from './types/PathPrefixProps';
 
+// Declared/defined in public/index.html, value replaced by Prometheus when serving bundle.
```
drumilpatel2000: @juliusv Can you please suggest a better way than this

web/ui/react-app/public/index.html
```diff
@@ -10,11 +10,18 @@
     <meta name="theme-color" content="#000000" />
 
     <!--
-      This constant's placeholder magic value is replaced during serving by Prometheus
+      PATH_PREFIX constants placeholder magic value is replaced during serving by Prometheus
```
juliusv:
```suggestion
      The PATH_PREFIX placeholder magic value is replaced during serving by Prometheus
```

web/ui/react-app/public/index.html
```diff
@@ -10,11 +10,18 @@
...
       and set to Prometheus's external URL path. It gets prepended to all links back
       to Prometheus, both for asset loading as well as API accesses.
+
+      CONSOLES_LINK constants placeholder magic value is replaced during serving by Prometheus
```
juliusv:
```suggestion
      The CONSOLES_LINK placeholder magic value is replaced during serving by Prometheus
```

web/ui/react-app/public/index.html
```diff
@@ -10,11 +10,18 @@
...
+      and set to console link if it exists. It will render Console link on navbar when
```
juliusv:
```suggestion
      and set to the consoles link if it exists. It will render a "Consoles" link in the navbar when 
```

web/ui/react-app/public/index.html
```diff
@@ -10,11 +10,18 @@
...
 
     <!--
-      This constant's placeholder magic value is replaced during serving by Prometheus
+      PATH_PREFIX constants placeholder magic value is replaced during serving by Prometheus
       and set to Prometheus's external URL path. It gets prepended to all links back
       to Prometheus, both for asset loading as well as API accesses.
+
+      CONSOLES_LINK constants placeholder magic value is replaced during serving by Prometheus
+      and set to console link if it exists. It will render Console link on navbar when 
+      it is non empty.
```
juliusv:
```suggestion
      it is non-empty. 
```

web/ui/react-app/src/Navbar.tsx
juliusv: I'm not sure we can do better, but I would prefer if we keep this super magic placeholder stuff to `index.tsx` (where the existing path prefix magic is as well), and then hand it down to the App and the Navbar using a prop.

web/ui/react-app/src/Navbar.tsx
```diff
@@ -14,6 +15,14 @@ import {
...
+declare const CONSOLES_LINK: string;
+
+let consoleLink = CONSOLES_LINK;
```
juliusv:
```suggestion
let consolesLink = CONSOLES_LINK;
```

web/ui/react-app/src/Navbar.tsx
```diff
@@ -25,6 +34,11 @@ const Navigation: FC<PathPrefixProps> = ({ pathPrefix }) => {
       </Link>
       <Collapse isOpen={isOpen} navbar style={{ justifyContent: 'space-between' }}>
         <Nav className="ml-0" navbar>
+          {consoleLink !== '' && (
+            <NavItem>
+              <NavLink href={`${pathPrefix + consoleLink}`}>Consoles</NavLink>
```
juliusv: The consoles link already includes the path prefix, so `pathPrefix` should not be needed here.

web/ui/react-app/src/Navbar.tsx
```diff
@@ -25,6 +34,11 @@ const Navigation: FC<PathPrefixProps> = ({ pathPrefix }) => {
       </Link>
       <Collapse isOpen={isOpen} navbar style={{ justifyContent: 'space-between' }}>
         <Nav className="ml-0" navbar>
+          {consoleLink !== '' && (
```
juliusv: A simple test for the existence / non-existence of the extra nav item would be good to have. We don't have a `Navbar.test.tsx` yet, but you can get inspired by e.g. https://github.com/prometheus/prometheus/blob/master/web/ui/react-app/src/pages/flags/Flags.test.tsx

boyskila: Can we just do ```CONSOLES_LINK === 'CONSOLES_LINK_PLACEHOLDER' || CONSOLES_LINK === '' ? null : <NavItem>...</NavItem>``` directly instead of introducing an extra variable?
What if the result of https://github.com/prometheus/prometheus/pull/6761/files#diff-73cbd2009b0e4f156e8ca0f47e95b016R385 is null? Maybe we should check against that as well

web/ui/react-app/public/index.html
```diff
@@ -10,11 +10,18 @@
...
       and set to Prometheus's external URL path. It gets prepended to all links back
       to Prometheus, both for asset loading as well as API accesses.
+
+      CONSOLES_LINK constants placeholder magic value is replaced during serving by Prometheus
+      and set to console link if it exists. It will render Console link on navbar when 
+      it is non empty. 
     -->
-    <script>const PATH_PREFIX='PATH_PREFIX_PLACEHOLDER';</script>
+    <script>
+      const PATH_PREFIX='PATH_PREFIX_PLACEHOLDER';
```
boyskila: Maybe is not a bad idea to prefix those variables with ```GLOBAL_```

web/ui/react-app/src/Navbar.tsx
juliusv: Combined with my other comment about not having the global variables in this file, I think the extra `consolesLink` variable would make sense, but as a prop, and as type `string | null` (or `string | undefined`, we should at some point standardize on one).

boyskila: Yeah, that's why I suggested this https://github.com/prometheus/prometheus/pull/6761#discussion_r375800896
IMO the confusion around those 'magic' variables will disappear this way.

juliusv: Yeah, fine to add a `GLOBAL_` prefix, but I still think any global state shouldn't even be used outside of index.html.

web/ui/react-app/src/index.tsx
```diff
@@ -5,15 +5,21 @@ import App from './App';
...
+if (GLOBAL_PATH_PREFIX === 'PATH_PREFIX_PLACEHOLDER' || GLOBAL_PATH_PREFIX === '/') {
   // Either we are running the app outside of Prometheus, so the placeholder value in
   // the index.html didn't get replaced, or we have a '/' prefix, which we also need to
   // normalize to '' to make concatenations work (prefixes like '/foo/bar/' already get
   // their trailing slash stripped by Prometheus).
   prefix = '';
 }
 
-ReactDOM.render(<App pathPrefix={prefix} />, document.getElementById('root'));
+if (GLOBAL_CONSOLES_LINK === 'CONSOLES_LINK_PLACEHOLDER' || GLOBAL_CONSOLES_LINK === '') {
```
boyskila: I think is good to add null/undefined check here as well. Please check https://github.com/prometheus/prometheus/blob/1a1859417696170f2ceaff34222942bd95bbda78/web/ui/react-app/src/utils/index.ts#L14

web/ui/react-app/src/types/PathPrefixProps.ts
```diff
@@ -1,5 +1,6 @@
 interface PathPrefixProps {
   pathPrefix?: string;
+  consolesLink?: string | null;
```
boyskila: @juliusv in response to the last part of this comment https://github.com/prometheus/prometheus/pull/6761#discussion_r375805037
Is there an option to be undefined? I guess the API will return null either way if something unexpected happens
I'm for making them ```string | null``` and to remove the optional operator (```?```)

drumilpatel2000: @boyskila Can you explain the particular problem in detail?

boyskila: Not sure whether I want to call it a problem....just thinking over what @juliusv mention in https://github.com/prometheus/prometheus/pull/6761#discussion_r375805037 about
```string | null``` vs ```?``` ie. ```string | undefined```

Comment (boyskila) [reply web/ui/react-app/src/types/PathPrefixProps.ts]: And as I said I'm more for 
```suggestion
  pathPrefix: string | null;
  consolesLink: string | null;
```

drumilpatel2000: @boyskila We can acheive it by adding a new type file `ConsolesLinkType`
Screenshot from 2019-12-12 04-33-29
Screenshot from 2020-02-07 14-56-49
Screenshot from 2020-02-07 14-57-11
Screenshot from 2020-02-07 14-57-28
@juliusv @boyskila Any thoughts on this one

web/ui/react-app/src/Navbar.test.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
+import { shallow } from 'enzyme';
+import Navigation from './Navbar';
+import { NavItem, NavLink } from 'reactstrap';
+
+describe('Navbar should contains console Link', () => {
```
juliusv: contains -> contain

web/ui/react-app/src/Navbar.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+      app.contains(
+        <NavItem>
+          <NavLink href="/path/consoles">Consoles</NavLink>
+        </NavItem>
+      )
+    ).toBeTruthy();
+  });
+});
+
+describe('Navbar should not contains consoles link', () => {
```
juliusv: contains -> contain

web/ui/react-app/src/Navbar.test.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
+import { shallow } from 'enzyme';
+import Navigation from './Navbar';
+import { NavItem, NavLink } from 'reactstrap';
+
+describe('Navbar should contains console Link', () => {
+  it('with non-empty consoleslinks', () => {
```
juliusv: consolesLinks -> consolesLink

(similar below)

web/ui/react-app/src/index.tsx
juliusv: I don't think we need to check null/undefined here, since that would be an error (since it is always set to something in index.html). We don't do it for the path prefix either.

web/ui/react-app/src/types/PathPrefixProps.ts
juliusv: Yep, `consolesLink?: string | null` would imply double optionality (it can be either `undefined` or `null`, not just one of them). I'd be for making it required, but nullable (coming from Go, I like things explicit :)).

juliusv: As an aside, I don't think `consolesLink` belongs into `PathPrefixProps`, since it doesn't relate to the path prefix, and is only needed by the Nav (and App for pass-through). Since it's only required in two places, I would probably not even make a shared type for it, but just make separate props types for the App and the Nav (in the files where those components live), that each expect this field?

boyskila: Not sure whether worth to split into separate interfaces. The point was just to type both variables explicitly to string | null. Btw, what about calling this interface ```GlobalProps``` or something. WDYT?

juliusv: @boyskila Seems our comments overlapped on that, I'd prefer https://github.com/prometheus/prometheus/pull/6761/files#r376299510 since I don't want anyone instantiating any of the unrelated components to also have to pass a completely unrelated `consolesLink` to all components that have a path prefix... like let's keep it localized to the one component that really needs it?

web/ui/react-app/src/Navbar.tsx
```diff
@@ -14,7 +14,11 @@ import {
 } from 'reactstrap';
 import PathPrefixProps from './types/PathPrefixProps';
 
-const Navigation: FC<PathPrefixProps> = ({ pathPrefix }) => {
+interface ConsolesLinkProps {
```
juliusv: Last nit: convention-wise, if this were a shared props type defined in one place only and reused from multiple components, then we would name it `ConsolesLinkProps`. Otherwise if it's a props type defined specifically for one component, we follow the naming convention `<Component>Props`, so `NavigationProps` here. There could be a case for making a shared props type out of it (like for `PathPrefixProps`), but IMO as long as it's only really used in one place (+ App pass-through), I prefer it defined twice, but with the component props naming convention.

drumilpatel2000: @juliusv Will `NavbarConsolesLink` and `AppConsolesLink` work fine?

web/ui/react-app/src/index.tsx
boyskila: If null is coming out from the API then what value will be passed to <App> for consolesLink? I think it's going to be null because line 22 will never be reached and as far I can see the check in the navbar is ```consolesLink !== ''``` which will return true and the navbar item will be rendered incorrectly

drumilpatel2000: @boyskila `!isPresent` should be used for both the global value.

web/ui/react-app/src/Navbar.tsx
```diff
@@ -14,7 +14,11 @@ import {
 } from 'reactstrap';
 import PathPrefixProps from './types/PathPrefixProps';
 
-const Navigation: FC<PathPrefixProps> = ({ pathPrefix }) => {
+interface NavbarConsolesLink {
```
boyskila: This is a duplicate of the one in ```App.tsx```

juliusv: Yeah, so what I meant was calling it `<Component>Props` if we don't make it a shared type, so it should be `NavbarProps` here, not `NavbarConsolesLink` (since it would over time also contain other props for the navbar, as necessary).

We could also make it a shared props type for it (similar to `PathPrefixProps`), but then it should live in one place and just be named `ConsolesLinkProps`.

Since it's only *really* needed by one component though, I'm kind of for just going for the `NavbarProps` variant that includes minor duplication...

juliusv:
:+1: Thanks, just getting a CI error now:

```
/home/circleci/project/web/ui/react-app/src/index.tsx
  6:36  error  Insert `;`  prettier/prettier
```

Could you run `make react-app-lint-fix`?

drumilpatel2000:
@juliusv Any reviews

juliusv:
Yup, I was just sleeping :)

:+1: Thanks, all green now, will merge! I'll send one tiny follow-up change that you'll see :)

PR #6764, opened 2020-02-05
Keep consul service metrics in global variables

yeya24:

Each call of `WithLabelValues` needs to re-calculate the hash of the labels. So I think it is okay to use global variables to keep the metrics which labels are already known

### DIFF
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -83,6 +83,10 @@ var (
 		[]string{"endpoint", "call"},
 	)
 
+	// Initialize metric vectors.
+	servicesRPCDuraion = rpcDuration.WithLabelValues("catalog", "services")
+	serviceRPCDuraion  = rpcDuration.WithLabelValues("catalog", "service")
+
 	// DefaultSDConfig is the default Consul SD configuration.
 	DefaultSDConfig = SDConfig{
 		TagSeparator:    ",",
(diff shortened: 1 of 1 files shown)


yeya24:
Same problem exists in Kubernetes SD metrics. If this change is in okay, I can include them in the same PR

``` Go
	e.endpointsInf.AddEventHandler(cache.ResourceEventHandlerFuncs{
		AddFunc: func(o interface{}) {
			eventCount.WithLabelValues("endpoints", "add").Inc()
			e.enqueue(o)
		},
		UpdateFunc: func(_, o interface{}) {
			eventCount.WithLabelValues("endpoints", "update").Inc()
			e.enqueue(o)
		},
		DeleteFunc: func(o interface{}) {
			eventCount.WithLabelValues("endpoints", "delete").Inc()
			e.enqueue(o)
		},
	})

```

brian-brazil:
May I ask what inspired you to change this? It is faster, but only by maybe 100ns.

yeya24:
> May I ask what inspired you to change this? It is faster, but only by maybe 100ns.

I just think if it is an improvement? Even only a little...

brian-brazil:
Thanks!

yeya24:
> Thanks!

Sorry, I was intended to change the metrics in Kubernets SD to global in this PR as well. So is it OK for me to open another one @brian-brazil

brian-brazil:
Of course.

PR #6767, opened 2020-02-06
Storage concurrently v4

liguozhong:
from https://github.com/prometheus/prometheus/pull/6716
Storage concurrently


liguozhong:
merge fail

Issue #6714, opened 2020-01-28
Reading from storage sources concurrently


mknapphrt:
## Proposal
**Use case. Why is this important?**
Currently the storage sources are being read from sequentially and I think there's potentially room for improvement here. Doing bigger queries where series are being pulled from more than 1 source (say 2 sharded remote read targets or something) means that there's significant time waiting on each response, which could be done at the same time. I believe this would require running each select in a thread as part of `(q *mergeQuerier) SelectSorted`.

https://github.com/prometheus/prometheus/blob/9adad8ad3045a93fff03400970f1aa77cc034cb0/storage/fanout.go#L238-L263

I know most people aren't using remote storage sources but I don't believe this should have much if any impact on performance in a local storage only setup. Does this seem like a reasonable thing to pursue? Any opinions on this?

PR #6770, opened 2020-02-06
Storage concurrently

liguozhong:
Storage concurrentl from #6716
issue https://github.com/prometheus/prometheus/issues/6714

### DIFF
--- a/storage/fanout.go
+++ b/storage/fanout.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"sort"
 	"strings"
-	"sync"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
@@ -242,17 +241,13 @@ func (q *mergeQuerier) SelectSorted(params *SelectParams, matchers ...*labels.Ma
 	seriesSets := make([]SeriesSet, 0, len(q.queriers))
 	var warnings Warnings
 
-	wg := sync.WaitGroup{}
-	wg.Add(len(q.queriers))
-	queryResultChan := make(chan *queryResult, len(q.queriers))
+	queryResultChan := make(chan *queryResult)
 	for _, querier := range q.queriers {
 		go func(qr Querier) {
-			defer wg.Done()
 			set, wrn, err := qr.SelectSorted(params, matchers...)
 			queryResultChan <- &queryResult{qr: qr, set: set, wrn: wrn, selectError: err}
 		}(querier)
 	}
-	wg.Wait()
 	for i := 0; i < len(q.queriers); i++ {
 		qryResult := <-queryResultChan
 		q.setQuerierMap[qryResult.set] = qryResult.qr
--- a/storage/fanout.go
+++ b/storage/fanout.go
@@ -241,15 +241,23 @@ func (q *mergeQuerier) SelectSorted(params *SelectParams, matchers ...*labels.Ma
 	seriesSets := make([]SeriesSet, 0, len(q.queriers))
 	var warnings Warnings
 
+	var priErr error = nil
 	queryResultChan := make(chan *queryResult)
 	for _, querier := range q.queriers {
 		go func(qr Querier) {
+			if priErr != nil {
+				queryResultChan <- &queryResult{qr: qr, set: nil, wrn: nil, selectError: nil}
+				return
+			}
 			set, wrn, err := qr.SelectSorted(params, matchers...)
 			queryResultChan <- &queryResult{qr: qr, set: set, wrn: wrn, selectError: err}
 		}(querier)
 	}
 	for i := 0; i < len(q.queriers); i++ {
 		qryResult := <-queryResultChan
+		if priErr != nil {
+			continue
+		}
 		q.setQuerierMap[qryResult.set] = qryResult.qr
 		if qryResult.wrn != nil {
 			warnings = append(warnings, qryResult.wrn...)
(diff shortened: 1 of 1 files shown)


storage/fanout.go
```diff
@@ -240,27 +241,44 @@ func (q *mergeQuerier) Select(params *SelectParams, matchers ...*labels.Matcher)
 func (q *mergeQuerier) SelectSorted(params *SelectParams, matchers ...*labels.Matcher) (SeriesSet, Warnings, error) {
 	seriesSets := make([]SeriesSet, 0, len(q.queriers))
 	var warnings Warnings
+
+	wg := sync.WaitGroup{}
+	wg.Add(len(q.queriers))
```
brian-brazil: You don't need the waitgrup any more, and the channel here doesn't even need to be buffered

liguozhong: Thanks for your review , done.
Thank you for your help. I have less than 1 year experience in go language. My previous work experience is using Java

image

storage/fanout.go
```diff
@@ -240,27 +240,40 @@ func (q *mergeQuerier) Select(params *SelectParams, matchers ...*labels.Matcher)
...
-				return nil, nil, err
+				return nil, nil, qryResult.selectError
 			}
 		}
-		seriesSets = append(seriesSets, set)
 	}
 	return NewMergeSeriesSet(seriesSets, q), warnings, nil
 }
 
+type queryResult struct {
```
brian-brazil: Put this inside the function, as you're only using it there.

storage/fanout.go
```diff
@@ -240,27 +240,40 @@ func (q *mergeQuerier) Select(params *SelectParams, matchers ...*labels.Matcher)
...
+			q.failedQueriers[qryResult.qr] = struct{}{}
 			// If the error source isn't the primary querier, return the error as a warning and continue.
-			if querier != q.primaryQuerier {
-				warnings = append(warnings, err)
-				continue
+			if qryResult.qr != q.primaryQuerier {
+				warnings = append(warnings, qryResult.selectError)
 			} else {
-				return nil, nil, err
+				return nil, nil, qryResult.selectError
```
brian-brazil: If we return early here then the channel isn't emptied. We'll need a little more work here.

liguozhong: done. A "priRrr" variable is added to ensure that the data in the channel can be consumed

image

storage/fanout.go
liguozhong: done. 

image

storage/fanout.go
```diff
@@ -240,23 +240,46 @@ func (q *mergeQuerier) Select(params *SelectParams, matchers ...*labels.Matcher)
...
+		selectError error
+	}
+	queryResultChan := make(chan *queryResult)
 	for _, querier := range q.queriers {
-		set, wrn, err := querier.SelectSorted(params, matchers...)
-		q.setQuerierMap[set] = querier
-		if wrn != nil {
-			warnings = append(warnings, wrn...)
+		go func(qr Querier) {
+			if priErr != nil {
```
brian-brazil: You've still a data race here, you can remove this if entirely.

storage/fanout.go
```diff
@@ -240,23 +240,46 @@ func (q *mergeQuerier) Select(params *SelectParams, matchers ...*labels.Matcher)
...
+				queryResultChan <- &queryResult{qr: qr, set: nil, wrn: nil, selectError: nil}
+				return
+			}
+			set, wrn, err := qr.SelectSorted(params, matchers...)
+			queryResultChan <- &queryResult{qr: qr, set: set, wrn: wrn, selectError: err}
+		}(querier)
+	}
+	for i := 0; i < len(q.queriers); i++ {
+		qryResult := <-queryResultChan
+		if priErr != nil {
```
brian-brazil: We may aswell collect all the warnings when there's an error in the primary.

liguozhong: done.
image

storage/fanout.go
liguozhong: Thank you for your review.
done.
image

storage/fanout.go
```diff
@@ -240,23 +240,43 @@ func (q *mergeQuerier) Select(params *SelectParams, matchers ...*labels.Matcher)
...
-		if wrn != nil {
-			warnings = append(warnings, wrn...)
+		go func(qr Querier) {
+			set, wrn, err := qr.SelectSorted(params, matchers...)
+			queryResultChan <- &queryResult{qr: qr, set: set, wrn: wrn, selectError: err}
+		}(querier)
+	}
+	for i := 0; i < len(q.queriers); i++ {
+		qryResult := <-queryResultChan
+		if qryResult.wrn != nil {
```
brian-brazil: You can simplify by removing this and the next `if`. The performance cost will be negligible.

liguozhong: done.Thank you for your review.
image

brian-brazil:
Thanks!

liguozhong:
> Thanks!

I really very much appreciate your guidance and help.

mknapphrt:
I tried building this to test it out and I get no results. Maybe I'm missing something but where are you actually appending the query results to `seriesSets `? AFAICT it makes an empty map at the beginning of the method and results that same empty map.

brian-brazil:
Hmm, that looks to be the case and unittests didn't catch it either. Would you like to send a PR to fix it?

mknapphrt:
I'll gladly PR the fix, all I did was
```
                qryResult := <-queryResultChan
                q.setQuerierMap[qryResult.set] = qryResult.qr
+              seriesSets = append(seriesSets, qryResult.set)
                if qryResult.wrn != nil {
                        warnings = append(warnings, qryResult.wrn...)
                }
```
but I don't really have time right now to work on the unit tests to go along with it.

liguozhong:
I'll  work on the unit tests

PR #6772, opened 2020-02-06
Update TestWALSegmentSizeOptions to skip directories

codesome:
This is a broken down piece of https://github.com/prometheus/prometheus/pull/6679

`TestWALSegmentSizeOptions` should be skipping the directories when checking the WAL segments. Currently, there can be checkpoint directory inside the `wal`, but not in this test. But a chunk directory will be introduced in #6679.


krasi-georgiev:
This PR is a bit strange on its own, I also prefer to break down big PRs into smaller ones, but I think that this change only makes sence with the main PR itself.

codesome:
Yeah, a bit not obvious when done without mmapped chunks. I will close it then and have it in the main PR.

PR #6773, opened 2020-02-06
Refactor TestGCChunkAccess and TestGCSeriesAccess to create chunks by appending samples

codesome:
This is a broken down piece of https://github.com/prometheus/prometheus/pull/6679.

This just changes the way of initializing chunks for the mentioned tests. It's going to be important for mmapped chunks where we cannot add just to the slice.


krasi-georgiev:
LGTM

Issue #6771, opened 2020-02-06
Performance of v2.16.0-rc.0


roidelapluie:
**What did you do?**

This issue tracks some unexpected performance differences between 2.16.0-rc.0 and 2.15.2.

![image](https://user-images.githubusercontent.com/291750/73934760-66556200-48df-11ea-9664-04d3b9e91466.png)
Allocated is pretty much the same, however RSS is higher (which is harmless)

Allocations are a lot higher
![image](https://user-images.githubusercontent.com/291750/73934908-ae748480-48df-11ea-8e9c-8ac65e9e5493.png)

Response time is higher
![image](https://user-images.githubusercontent.com/291750/73934945-c21feb00-48df-11ea-8075-949b0db10995.png)

did we break these metrics?
![image](https://user-images.githubusercontent.com/291750/73935110-1dea7400-48e0-11ea-8abc-fb813281b207.png)

**Environment**

https://github.com/prometheus/prometheus/pull/6729
http://prombench.prometheus.io/grafana/d/7gmLoNDmz/prombench?orgId=1&var-pr-number=6729

slrtbtfs:
Some notes about the changes that PromQL underwent since the last release:

* The parser was completely rewritten. Since the parsing takes very little time compared to the rest of the engine, it is however relatively unlikely that the performance regressions come from the parser itself.
* There have been two PRs from me, that potentially have an performance impact on the engine: #6612 and #6590. I've reviewed both of them again and didn't see anything that would explain these differences.
* There seem to have been some changes regarding logging. Not sure how these affect things. I guess @roidelapluie and @brian-brazil are more knowledgeable here.

roidelapluie:
If the logging is not enabled there should be no noticeable change.

slrtbtfs:
I've had a look at the PromQL engine benchmarks and both the speed and memory usage seem to have slightly improved, compared to 2.15.2.

bwplotka:
Keep in mind that benchmarks might be missing certain test cases or some
real factors, thus real results might be different. What I am saying that
it is some fact, but we should not completely rule our PromQL changes.

Kind Regards,
Bartek

cstyan:
So after a bunch of prombench runs and bisecting, while looking only at the allocs graphs, it looks like #6591 might be causing the allocation issues. 

[...]

PR #6776, opened 2020-02-06
PromQL: Fix params for series expansion

slrtbtfs:
One of the refactorings in #6590 causes the parameters for series expansion in the engine to be set incorrectly.

This might be (part of) the cause for #6771.


roidelapluie:
@slrtbtfs do we need/want this? should it be rebased to master? Does it have end users impact?

slrtbtfs:
As far as I know this doesn't actually have an notable impact, expect a handful of saved CPU cycles in certain cases.

There are syntax tree instances that behave differently, depending on whether this is merged or not, but I can't think of a way such an instance could pass type checking.

So merging doesn't hurt, but probably won't make a difference.

PR #6777, opened 2020-02-06
Import TSDB directly, without adapter; Unify Iterators, Querier, Series interfaces.

bwplotka:
The interface separation was required as we were prepared for multiple TSDB implementations. We have one that just got merged into the same repository, so let's simplify everything. By accident, because of all of this, we have two complex MergeSeriesSet function that does the same thing. This will be fixed in next PRs.

This is part of https://github.com/prometheus/prometheus/pull/5882 that has to be done to simplify things.
All todos I added will be fixed in follow up PRs.

## Changes

* querier.Querier, querier.Appender, querier.SeriesSet, and querier.Series interfaces merged
with storage interface.go. All imports that.
* querier.SeriesIterator replaced by chunkenc.Iterator
* Added chunkenc.Iterator.Seek method and tests for xor implementation (?)
* Since we properly handle SelectParams for Select methods I adjusted min max
based on that. This should help in terms of performance for queries with functions like offset.
* added Seek to deletedIterator and test.
* storage/tsdb was removed as it was only a unnecessary glue with incompatible structs.

No logic was changed, an only different sources of abstractions, so no need for benchmarks.


brian-brazil:
Looks okay. Someone from Cortex should also take a look.

bwplotka:
All should be ready for review @pracucci @pstibrany 

Let me know if that's ok from Cortex perspective. Thanks!

tsdb/db_test.go
```diff
@@ -2203,20 +2214,16 @@ func TestBlockRanges(t *testing.T) {
...
 
 	// Test that the compactor doesn't create overlapping blocks
 	// when a non standard block already exists.
 	firstBlockMaxT := int64(3)
 	createBlock(t, dir, genSeries(1, 1, 0, firstBlockMaxT))
-	db, err := Open(dir, logger, nil, DefaultOptions)
-	if err != nil {
-		t.Fatalf("Opening test storage failed: %s", err)
-	}
+	db, err := open(dir, logger, nil, DefaultOptions(), []int64{10000})
```
krasi-georgiev: why not usre the public Open method to be consistent?

tsdb/db_test.go
```diff
@@ -2692,3 +2699,38 @@ func TestChunkReader_ConcurrentReads(t *testing.T) {
...
 }
+
+func TestTimeMetrics(t *testing.T) {
+	db, closeFn := openTestDB(t, nil, nil)
+	defer closeFn()
+
+	metrics := &dto.Metric{}
+
+	// Check initial values.
+	testutil.Ok(t, db.metrics.minTime.Write(metrics))
```
krasi-georgiev: just a reminder that can also use testutil.ToFloat64
"github.com/prometheus/client_golang/prometheus/testutil"

tsdb/db.go
```diff
@@ -93,31 +101,13 @@ type Options struct {
 
 	// StripeSize is the size in entries of the series hash map. Reducing the size will save memory but impact perfomance.
 	StripeSize int
-}
 
-// Appender allows appending a batch of data. It must be completed with a
```
pstibrany: This documentation is lost, existing storage.Appender doesn't have it anymore.

tsdb/db.go
```diff
@@ -66,20 +77,17 @@ type Options struct {
...
 
 	// Maximum number of bytes in blocks to be retained.
 	// 0 or less means disabled.
 	// NOTE: For proper storage calculations need to consider
 	// the size of the WAL folder which is not added when calculating
 	// the current size of the database.
-	MaxBytes int64
-
-	// The sizes of the Blocks.
-	BlockRanges []int64
```
pstibrany: It's not clear to me what was this replaced with.

pstibrany:
> I will try to integrate it into Cortex to make sure if I haven't missed anything.

I was able to easily integrate it into Cortex and make unit tests run. (I cannot fully compile and run, as Cortex also depends on Thanos...)

bwplotka:
Thanks for all the detailed comments! :heart: 

Addressing soon.

tsdb/chunkenc/chunk.go
```diff
@@ -72,10 +73,22 @@ type Appender interface {
...
+// Iterator iterates over the samples of a time series.
 type Iterator interface {
+	// Seek advances the iterator forward to the sample with the timestamp t or first value after t.
+	// Seek has no effect if requested timestamp is the same or lower than the current iterator position.
+	// Seek returns false if there is no such sample with the timestamp equal or larger than t.
+	// Iterator is exhausted when the Seek returns false.
+	// TODO(bwplotka): Verify above statements on all implementations with unit test.
+	Seek(t int64) bool
+	// At returns the current timestamp/value pair.
+	// At returns (math.MinInt64, 0.0) before the iterator has advanced.
```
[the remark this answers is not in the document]
bwplotka: So undefined or panic? ((: 

Yes, it just nice to have to have it **consistent**. I vote for this instead of panic, and test in unit test against it. WDYT?

tsdb/chunkenc/chunk.go
```diff
@@ -72,10 +73,22 @@ type Appender interface {
...
+	// TODO(bwplotka): Verify above statement on all implementations with unit test.
 	At() (int64, float64)
-	Err() error
+	// Next advances the iterator by one.
```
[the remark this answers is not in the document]
bwplotka: You can use `Seek` before  `Next` so I would not be that strict here, but happy to do so (:

storage/interface.go
```diff
@@ -49,7 +56,8 @@ type Queryable interface {
 	Querier(ctx context.Context, mint, maxt int64) (Querier, error)
```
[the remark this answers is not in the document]
bwplotka: Let's do that in separate PR

storage/interface.go
```diff
@@ -49,7 +56,8 @@ type Queryable interface {
...
 }
 
-// Querier provides reading access to time series data.
+// Querier provides querying access over time series data of a fixed
+// time range.
 type Querier interface {
 	// Select returns a set of series that matches the given label matchers.
```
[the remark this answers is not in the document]
bwplotka: ditto

tsdb/chunkenc/chunk.go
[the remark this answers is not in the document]
pstibrany: I think panic falls under "undefined" :)

If you put that into specification, then all custom iterators need to implement it, because there can be a piece of code relying on it. If there is a good reason for it, that's fine. If not, it's extra burden on implementations.

storage/interface.go
```diff
@@ -58,6 +66,7 @@ type Querier interface {
 	SelectSorted(*SelectParams, ...*labels.Matcher) (SeriesSet, Warnings, error)
 
 	// LabelValues returns all potential values for a label name.
+	// It is not safe to use the strings beyond the lifefime of the querier.
```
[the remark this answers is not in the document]
bwplotka: No, old one from the similar interface in the `tsdb.Querier`

tsdb/db.go
```diff
@@ -341,7 +363,7 @@ func (db *DBReadOnly) FlushWAL(dir string) error {
 
 // Querier loads the wal and returns a new querier over the data partition for the given time range.
 // Current implementation doesn't support multiple Queriers.
-func (db *DBReadOnly) Querier(mint, maxt int64) (Querier, error) {
+func (db *DBReadOnly) Querier(_ context.Context, mint, maxt int64) (storage.Querier, error) {
```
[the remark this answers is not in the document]
bwplotka: :+1:

tsdb/db_test.go
bwplotka: Because it does not allow overridding range

tsdb/db_test.go
bwplotka: Yes, I just copied what was there before.. should I change?

tsdb/db.go
bwplotka: Important, thanks!

tsdb/db.go
bwplotka: It is not controlled by Min and Max Block sizes. It can be overrided by tests as well.

storage/interface.go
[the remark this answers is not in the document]
pstibrany: OK, thanks!

tsdb/chunkenc/chunk.go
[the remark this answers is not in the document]
pstibrany: From a quick look, for example `xorIterator` currently doesn't do this, and would require extra code to handle this requirement.

tsdb/chunkenc/chunk.go
```diff
@@ -72,10 +73,23 @@ type Appender interface {
...
+	// Seek returns true, if such sample exists, false otherwise.
+	// Iterator is exhausted when the Seek returns false.
+	// TODO(bwplotka): Verify above statements on all implementations with unit test.
+	Seek(t int64) bool
+	// At returns the current timestamp/value pair.
+	// At returns (math.MinInt64, 0.0) before the iterator has advanced.
+	// TODO(bwplotka): Verify above statement on all implementations with unit test.
 	At() (int64, float64)
+	// Err returns the current error.
+	// Err can return undefined value before the iterator has advanced.
```
pstibrany: I would be specific and say "before calling Next or Seek", because "iterator has advanced" is likely not true. But that's really just nitpicking.

tsdb/db_test.go
krasi-georgiev: doesn't matter, not a big difference.

tsdb/chunkenc/chunk.go
bwplotka: It is fair, fixed it (:

krasi-georgiev:
just need to fix the failing tests now :)

bwplotka:
It took ages, but finally, it should be ready for final review! @krasi-georgiev @pracucci @pstibrany @brian-brazil Thanks!

bwplotka:
Please review, but we will merge only after full v2.16.0 is out cc @brian-brazil

storage/interface.go
```diff
@@ -90,14 +99,27 @@ func (f QueryableFunc) Querier(ctx context.Context, mint, maxt int64) (Querier,
...
 
-	AddFast(l labels.Labels, ref uint64, t int64, v float64) error
+	// AddFast adds a sample pair for the referenced series. It is generally
+	// faster than adding a sample by providing its full label set.
+	AddFast(ref uint64, t int64, v float64) error
 
 	// Commit submits the collected samples and purges the batch.
 	Commit() error
 
+	// Rollback rolls back all modifications made in the appender so far.
```
brian-brazil: This makes it seem as though you can keep on using the Appender after that.

tsdb/chunkenc/chunk.go
```diff
@@ -72,10 +73,23 @@ type Appender interface {
...
+	// Seek returns true, if such sample exists, false otherwise.
+	// Iterator is exhausted when the Seek returns false.
+	// TODO(bwplotka): Verify above statements on all implementations with unit test.
+	Seek(t int64) bool
+	// At returns the current timestamp/value pair.
+	// At returns (math.MinInt64, 0.0) before the iterator has advanced.
+	// TODO(bwplotka): Verify above statement on all implementations with unit test.
 	At() (int64, float64)
+	// Err returns the current error.
+	// Err can return undefined value before calling `Next` or `Seek`.
```
brian-brazil: There's no `undefined` value in Go

tsdb/db.go
```diff
@@ -45,19 +48,27 @@ import (
 
 // Default duration of a block in milliseconds - 2h.
 const (
-	DefaultBlockDuration = int64(2 * 60 * 60 * 1000)
+	DefaultBlockDuration = int64(2 * 60 * 60 * 1000) // 2h in miliseconds.
```
brian-brazil: This comment duplicates the one above

tsdb/db.go
```diff
@@ -45,19 +48,27 @@ import (
...
-	BlockRanges:            ExponentialBlockRanges(DefaultBlockDuration, 3, 5),
-	NoLockfile:             false,
-	AllowOverlappingBlocks: false,
-	WALCompression:         false,
-	StripeSize:             DefaultStripeSize,
+func DefaultOptions() *Options {
+	return &Options{
+		WALSegmentSize:         wal.DefaultSegmentSize,
+		RetentionDuration:      15 * 24 * model.Duration(time.Hour),
+		MinBlockDuration:       model.Duration(time.Duration(DefaultBlockDuration) * time.Millisecond),
```
brian-brazil: model has no place in tsdb, among other things the tsdb does not know anything about the units it is being used with.

tsdb/db.go
```diff
@@ -66,20 +77,17 @@ type Options struct {
 	// WALSegmentSize = 0, segment size is default size.
 	// WALSegmentSize > 0, segment size is WALSegmentSize.
 	// WALSegmentSize < 0, wal is disabled.
-	WALSegmentSize int
+	WALSegmentSize units.Base2Bytes
```
brian-brazil: Keep this as an int, no need for additinoal dependencies

tsdb/db.go
```diff
@@ -66,20 +77,17 @@ type Options struct {
...
 
 	// Duration of persisted data to keep.
-	RetentionDuration uint64
+	RetentionDuration model.Duration
```
brian-brazil: This is an opaque number, not a specific duration.

tsdb/db.go
```diff
@@ -245,7 +238,28 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 		Name: "prometheus_tsdb_size_retentions_total",
 		Help: "The number of times that blocks were deleted because the maximum number of bytes was exceeded.",
 	})
-
+	m.minTime = prometheus.NewGaugeFunc(prometheus.GaugeOpts{
+		Name: "prometheus_tsdb_lowest_timestamp_seconds",
```
brian-brazil: You don't know the unit the tsdb is working with, these metrics are they way they are for good reasons.

web/api/v1/api.go
```diff
@@ -1123,11 +1123,19 @@ func (api *API) remoteRead(w http.ResponseWriter, r *http.Request) {
...
 				if err != nil {
 					return err
 				}
 
+				if len(ws) > 0 {
+					msg := ""
+					for _, w := range ws {
+						msg += w.Error() + ";"
+					}
+					level.Warn(api.logger).Log("remote read warnings", "warnings", msg)
```
brian-brazil: This could be very spammy, this should not be in application logs.

tsdb/chunkenc/chunk.go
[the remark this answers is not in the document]
pstibrany: This statement is still not true for some iterators in Prometheus (eg. `mergeIterator` will simply panic). I (very strongly) prefer to keep `At()` behaviour before any `Next` or `Seek` operation to be unspecified, so that clients cannot rely on it.

tsdb/chunkenc/chunk.go
```diff
@@ -85,7 +99,8 @@ func NewNopIterator() Iterator {
 
 type nopIterator struct{}
```
pstibrany: Can this be replaced with noopSeriesIterator  from storage package?

tsdb/chunkenc/xor.go
```diff
@@ -253,6 +253,19 @@ type xorIterator struct {
 	err    error
 }
 
+func (it *xorIterator) Seek(t int64) bool {
+	if it.err != nil {
+		return false
+	}
+
+	for t > it.t || it.numRead == 0 {
```
pstibrany: Shouldn't logic be opposite here? We only want to advance, if `t < it.t`. Looks like unit test here would be a good idea.

tsdb/chunkenc/chunk.go
bwplotka: Yes, I rather meant it can return whatever

storage/interface.go
bwplotka: Will add comment.

tsdb/chunkenc/chunk.go
bwplotka: Will clarify.

tsdb/db.go
bwplotka: I see.... I can make that happen, but why tsdb cannot import `model`?

bwplotka: And thanks to that have much nicer UX? (units are embedded in type)

tsdb/db.go
bwplotka: Hm. It is used like time retention all the time...

bwplotka: If we revert this we would need to name it `RetentionDurationMiliseconds`

bwplotka: Oh.. . you want to totally abstract from units. Why is that useful? Isn't it just easier to agree on unit?

tsdb/db.go
bwplotka: I guess that's the reason: https://github.com/prometheus/prometheus/pull/6777#discussion_r377715609

tsdb/db.go
bwplotka: Hm.. Can we make TSDB unit-aware?

web/api/v1/api.go
bwplotka: How would you handle this then?

tsdb/chunkenc/chunk.go
bwplotka: Nope, as you will have cyclic import.

tsdb/chunkenc/xor.go
bwplotka: There is already: https://github.com/prometheus/prometheus/blob/07b7090dd78fafafbe6c229f77fc469d4bca4fd0/tsdb/chunkenc/chunk_test.go#L91

But I double check this.

tsdb/chunkenc/chunk.go
pstibrany: Then perhaps the other way around?

bwplotka: We can do that I believe :+1:

tsdb/chunkenc/xor.go
bwplotka: so `t > it.t` is fine as  we only want to advance if required t is smaller than iterator t (:

bwplotka: When I applied your suggestion, test failed , double checked if test makes sesne as well.

bwplotka:
> This statement is still not true for some iterators in Prometheus (eg. mergeIterator will simply panic). I (very strongly) prefer to keep At() behaviour before any Next or Seek operation (that returns true!) to be unspecified, so that clients cannot rely on it.

@pstibrany Might be that not all iterators comply to this but I believe there is an advantage in being consistent and specified. I will try to update all implementation and TODO is added to make sure with unit test if this works.

bwplotka:
Also I will fix `mergeIterator` in next PR (probably even merge with duplicate one)

web/api/v1/api.go
brian-brazil: It's in the query logs if you really want that.

bwplotka: kk, adding TODO

bwplotka:
All addressed @brian-brazil :+1:

tsdb/chunkenc/chunk.go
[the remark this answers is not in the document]
pstibrany: I’ll just add... I don’t mind if you change implementations of this interface in Prometheus to be consistent and return (MinInt64, 0.0). (It would be ok to mention that as well, something like “All implementations in Prometheus return (math.MinInt64, 0.0) before first successful Next/Seek, but this isn’t required by external implementations.”

What I object to is adding that requirement to the interface, because then ALL external implementations of this interface (eg. in Thanos or Cortex, and possibly other projects) must do this as well – because it is required by this contract after all. I would even say that it is an incompatible change of API.

bwplotka:
You are totally right @pstibrany, will remove the new requirement from the interfaces, however I would vote for making it consistent for Prometheus implementation in form of maxt, 0 e.g.

pstibrany:
> You are totally right @pstibrany, will remove the new requirement from the interfaces,

Thanks. I would go little further, and say that At() can only be called if Seek or Next returned true. Otherwise its return value is not specified.

> however I would vote for making it consistent for Prometheus implementations in form of maxt, 0 e.g.

Sounds good to me.
Thank you!

tsdb/chunkenc/chunk.go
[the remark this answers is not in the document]
brian-brazil: We're allowed make incompatible changes to APIs, this PR is full of them.

tsdb/chunkenc/chunk.go
```diff
@@ -72,10 +73,21 @@ type Appender interface {
...
+	Next() bool
+	// Seek advances the iterator forward to the first sample with the timestamp equal or greater than t.
+	// If current sample found by previous `Next` or `Seek` operation already has this property, Seek has no effect.
+	// Seek returns true, if such sample exists, false otherwise.
+	// Iterator is exhausted when the Seek returns false.
+	Seek(t int64) bool
+	// At returns the current timestamp/value pair.
+	// Before the iterator has advanced At behaviour is unspecified.
 	At() (int64, float64)
+	// Err returns the current error. It should be used only after iterator is
```
brian-brazil: Grammer is off here, it makes it seem like calling this will make Next/Seek return false.

s/so/that is/

web/web.go
```diff
@@ -212,11 +212,41 @@ func (h *Handler) ApplyConfig(conf *config.Config) error {
...
+		RetentionDuration:      int64(time.Duration(opts.RetentionDuration) / time.Millisecond),
+		MaxBytes:               int64(opts.MaxBytes),
+		NoLockfile:             opts.NoLockfile,
+		AllowOverlappingBlocks: opts.AllowOverlappingBlocks,
+		WALCompression:         opts.WALCompression,
+		StripeSize:             opts.StripeSize,
+		MinBlockDuration:       int64(time.Duration(opts.MinBlockDuration) / time.Millisecond),
+		MaxBlockDuration:       int64(time.Duration(opts.MaxBlockDuration) / time.Millisecond),
+
+		ConvertTimeToSecondsFn: func(i int64) float64 { return float64(i / 1000) },
```
brian-brazil: This shouldn't be here. The user's metrics are their own problem. Leave the metrics where they were.

tsdb/chunkenc/chunk.go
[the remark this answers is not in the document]
pstibrany: That's obviously true, but this change would be very subtle, and doesn't seem to be required to impose on implementors.

web/web.go
bwplotka: I think I disagree.. The previous stuff had globals. ): I will try to put them in separate place somehow.

tsdb/chunkenc/chunk.go
bwplotka: :+1: Thanks.

web/web.go
bwplotka: I think I found a good solution that will satisfiy all (:

bwplotka:
All addressed @brian-brazil @krasi-georgiev @codesome @brancz 

Let's merge this so I can move forward with https://github.com/prometheus/prometheus/pull/5882 :+1:

bwplotka:
Kind reminder @brian-brazil

cmd/prometheus/main.go
```diff
@@ -656,6 +656,7 @@ func main() {
 	}
 	{
 		// TSDB.
+		opts := cfg.tsdb.ToTSDBOptions()
```
brian-brazil: It feels weird to me that we're delegating flag parsing for the tsdb in main to the web handler.

tsdb/db.go
```diff
@@ -1393,3 +1435,88 @@ func exponential(d, min, max time.Duration) time.Duration {
 	}
 	return d
 }
+
+// ReadyStorage implements the Storage interface while allowing to set the actual
```
brian-brazil: Where is this used?

bwplotka: https://github.com/prometheus/prometheus/blob/5efee4f980ea140a4a6972ab06da79060e4302ce/cmd/prometheus/main.go#L337

cmd/prometheus/main.go
bwplotka: I agree, happy to move it to main

tsdb/db.go
brian-brazil: This seems like something that'd better live in main. startTimeMargin also seems a bit dubious in pure DB terms.

cmd/prometheus/main.go
bwplotka: done.

tsdb/db.go
bwplotka: Cool, will look it through.

bwplotka: done :+1:

bwplotka:
All addressed @brian-brazil  (:

tsdb/config/config.go
```diff
@@ -0,0 +1,50 @@
...
+	WALCompression         bool
+	StripeSize             int
+	MinBlockDuration       model.Duration
+	MaxBlockDuration       model.Duration
+}
+
+func (opts Options) ToTSDBOptions() tsdb.Options {
+	return tsdb.Options{
+		WALSegmentSize:         int(opts.WALSegmentSize),
+		RetentionDuration:      int64(time.Duration(opts.RetentionDuration) / time.Millisecond),
```
brian-brazil: I'm not sure about this, this is putting units into the tsdb again.

The TSDB does not need this code. Can we not handle this in main?

bwplotka:
Rebased... again. (:

tsdb/config/config.go
brian-brazil: This comment hasn't been addressed.

bwplotka: We cannot, `web` needs this.

bwplotka: Also it's not `tsdb` just separate package within `tsdb/` dir (:

brian-brazil: This logic doesn't belong under tsdb one way or the other, as nothing under tsdb should care about units nor should any of tsdb depend on model.

I think any rendering-time web conversions should be done in web.

brian-brazil: > Also it's not tsdb just separate package within tsdb/ dir (:

That's still tsdb as far as I'm concerned.

bwplotka: It's not only rendering time, it's also convenient flag parsing based on units. 

So, where would be the correct place for it? As separate `pkg/` package?

brian-brazil: That'd be one option. Possibly we want to split the parsing and the rendering here though.

bwplotka: cool, will take a look what we can do.

bwplotka: I think I found nice solution, let me know what you think. The struct with units is now fully in main.

brian-brazil: That looks good.

brian-brazil:
:+1:

bwplotka:
:tada: WIll merge on the green. Thanks all for review, was quite long PR.

PR #6778, opened 2020-02-06
ensure RuleGroups can be marshaled into YAML

jtlisi:
In [Cortex](https://github.com/cortexproject/cortex/pull/2088) we need to be able to marshal RuleGroups from the `rulefmt` package back into YAML. With the recent upgrade to utilize the `yaml.Node` struct in the `rulefmt.RuleNode` we are unable to do this.

This PR adds a `MarshalYAML` function to the `rulefmt.RuleNode` in order to make it possible to encode it back into yaml. A corresponding unit test that ensures the reencoded yaml matches the original has been added as well.

Signed-off-by: Jacob Lisi <jacob.t.lisi@gmail.com>

### DIFF
--- a/pkg/rulefmt/rulefmt.go
+++ b/pkg/rulefmt/rulefmt.go
@@ -128,6 +128,19 @@ type RuleNode struct {
 	Annotations map[string]string `yaml:"annotations,omitempty"`
 }
 
+// MarshalYAML implements yaml.Marshaler. It is required to encode the RuleGroup
+// struct back into yaml.
+func (r RuleNode) MarshalYAML() (interface{}, error) {
+	return &Rule{
+		Record:      r.Record.Value,
+		Alert:       r.Alert.Value,
+		Expr:        r.Expr.Value,
+		For:         r.For,
+		Labels:      r.Labels,
+		Annotations: r.Annotations,
+	}, nil
+}
+
 // Validate the rule and return a list of encountered errors.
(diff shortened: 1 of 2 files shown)


brian-brazil:
These are internal interface for Prometheus, we don't add things that Prometheus itself doesn't need as that could lead to code bloat and ossification. Is there somewhere Prometheus itself needs this?

roidelapluie:
What Brian said. Especially everything is public here so it should be trivial for cortex to re-create a rule and marshal it.

pstibrany:
> What Brian said. Especially everything is public here so it should be trivial for cortex to re-create a rule and marshal it.

This is what we currently do in Cortex' [PR 2088](https://github.com/cortexproject/cortex/pull/2088) – "recreate" it by simply copying older version of the package. It just seems that given how small this change is to make YAML serialization to work again, it would be nicer if we didn't have to do it.

brian-brazil:
This is merely the first place we've updated to the new yaml library, having to do this in all the other places would be a non-trivial burden and would often get forgotten. This smells like an issue with the upstream library if using the new features breaks the default serialisation.

pstibrany:
> This smells like an issue with the upstream library if using the new features breaks the default serialisation.

[It does indeed](https://github.com/go-yaml/yaml/issues/575).

brian-brazil:
Let's get that fixed upstream then, and then we can update our vendoring.

jtlisi:
Closing in favor of making the change upstream

PR #6781, opened 2020-02-06
OpenMetrics parse: avoid setting prev token

roidelapluie:
We can avoid setting a prev token in the OM parser. The previous
coundition that checked for prev was unreacheable.


brian-brazil:
Thanks!

roidelapluie:
Note that this has been merged to a rc0 branch as there is still no release branch.

beorn7:
I have now created https://github.com/prometheus/prometheus/tree/release-2.16 , which includes this commit. ← @cstyan FYI

PR #6787, opened 2020-02-07
scrape: use a RWMutex for metamtx

roidelapluie:
In #6675 we start to use metaMtx in each scrape.

However now it would make sense to use a RWLock for these, as multiple
functions just need to read the slice, including the new ones.


roidelapluie:
I let @cstyan appreciate whether or not this goes to 2.16.

brian-brazil:
RWMutexes make sense when you're doing many more reads than writes, which I
don't think is the case here.

roidelapluie:
ack , that is just 5 functions. let's close for now.

PR #6788, opened 2020-02-07
Update dependencies

roidelapluie:
This is just an attempt. go modules are quite messy...


roidelapluie:
I think I will take a more atomic version for this. We are very late and there are a bunch of things that don't work out 'out of the box'

PR #6789, opened 2020-02-08
[ENHANCEMENT] remote storage: When the matcher does not contain regular expressions, convert the value of type from RE to EQ

liguozhong:
Improve the query performance of remote storage. When the matcher does not contain regular expressions, convert the value of type from RE to EQ 

example：
ori promQL ：node_load5{instance=~"192.168.0.127:9100"}
new promQL：node_load5{instance="192.168.0.127:9100"}

case：
The multiple selection box in the upper left corner of grafana often has one or more conditions, which leads to invalid regular expression query in the following chart when a condition exists. This is generally very man-made

why？
The code modification in PR reduces our query performance of grafana from 10 seconds to 1 second.
Because the performance of regular expressions in each remote storage is very low.
This PR should also be meaningful to other users of remote storage.

image

### DIFF
--- a/storage/remote/codec.go
+++ b/storage/remote/codec.go
@@ -19,7 +19,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"sort"
-	"strings"
+	"unicode/utf8"
 
 	"github.com/gogo/protobuf/proto"
 	"github.com/golang/snappy"
@@ -123,27 +123,28 @@ func optMatcher(matchers []*prompb.LabelMatcher) {
 	}
 }
 
-func isRegexString(text string) bool {
-	if !strings.Contains(text, ".*") &&
-		!strings.Contains(text, "^") &&
-		!strings.Contains(text, "?") &&
-		!strings.Contains(text, "$") &&
-		!strings.Contains(text, "-") &&
-		!strings.Contains(text, ".+") &&
-		!strings.Contains(text, "?") &&
-		!strings.Contains(text, "\\") &&
-		!strings.Contains(text, "[") &&
-		!strings.Contains(text, "]") &&
-		!strings.Contains(text, "{") &&
-		!strings.Contains(text, "}") &&
-		!strings.Contains(text, "*") &&
-		!strings.Contains(text, "|") {
-		return true
+//from tsdb querier.go
+func isRegexString(pattern string) bool {
+	for i := 4; i < len(pattern)-2; i++ {
+		if isRegexMetaCharacter(pattern[i]) {
+			return true
+		}
 	}
 	return false
 }
 
-// ToQueryResult builds a QueryResult proto.
+var regexMetaCharacterBytes [16]byte
+
+func isRegexMetaCharacter(b byte) bool {
+	return b < utf8.RuneSelf && regexMetaCharacterBytes[b%16]&(1<<(b/16)) != 0
+}
+
+func init() {
+	for _, b := range []byte(`.+*?()|[]{}^$`) {
+		regexMetaCharacterBytes[b%16] |= 1 << (b / 16)
+	}
+}
+
 func ToQueryResult(ss storage.SeriesSet, sampleLimit int) (*prompb.QueryResult, error) {
 	numSamples := 0
 	resp := &prompb.QueryResult{}
--- a/storage/remote/codec.go
+++ b/storage/remote/codec.go
@@ -132,6 +132,11 @@ func isRegexString(text string) bool {
 		!strings.Contains(text, ".+") &&
 		!strings.Contains(text, "?") &&
 		!strings.Contains(text, "\\") &&
+		!strings.Contains(text, "[") &&
+		!strings.Contains(text, "]") &&
+		!strings.Contains(text, "{") &&
+		!strings.Contains(text, "}") &&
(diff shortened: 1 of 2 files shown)


roidelapluie:
Thank you for your pull request.

I get the idea. However, I think we should not implement it. Regexes come in many forms, and it is hard to get them all. There are too much risks there.

Just by looking at the PR there is `[]{}*` etc missing. In addition that, characters in the list could be escaped and in this case your code would simply produce bad results.

users should instead fix their queries.

roidelapluie:
Please note that this is NOT equivalent:

ori promQL ：node_load5{instance=~"192.168.0.127:9100"}
new promQL：node_load5{instance="192.168.0.127:9100"}

The first one will match 192a168.0.127:9100, the second one will not.

As I said, it is hard to get it right and I am not sure that we want this in prometheus.

liguozhong:
> Please note that this is NOT equivalent:
> 
> ori promQL ：node_load5{instance=~"192.168.0.127:9100"}
> new promQL：node_load5{instance="192.168.0.127:9100"}
> 
> The first one will match 192a168.0.127:9100, the second one will not.
> 
> As I said, it is hard to get it right and I am not sure that we want this in prometheus.

You are right. This optimization is easy to cause bugs, but it is limited to remote storage. We can do this optimization well through the research and development students in the community. Our team has 500 + customers using this optimization. At present, there is no user feedback about the unavailability caused by this optimization. I've added the five symbols of "[] {} *" in your feedback.

90% of our customers use grafana to view the data. When grafana uses the selection box, only "~ =" can be used in promql. Otherwise, when the user selects two filter conditions, the data cannot be returned. In most cases, the user only selects one filter condition. At this time, the performance is far worse than "=".

In addition, our team's work is mainly to help customers deploy operation and maintenance Prometheus. Our team believes that we should give back some effective features of our production environment to the Prometheus community.

We may not write perfect optimization code for this problem, but we hope that this optimization idea can also be implemented by more researchers as a topic in the community

The read-write performance of remote storage is very important for our type of team. We help our customers to host the operation and maintenance deployment of Prometheus

roidelapluie:
I fully understand. You should not that Grafana will escape the . by default and replace them with \\. , so I wonder why this works for you.

In general this should be a feature request towards grafana, they should be able to chose between =~ and =. In that case, even standard promql queries would benefit from this.

liguozhong:
Thank you very much for your review and suggestions.
 Our team is not very familiar with the grafana project, only staying in the use stage, we will try to contact the grafana community

roidelapluie:
You can also look at https://github.com/prometheus/tsdb/pull/602

roidelapluie:
And TBH if your approach is more close to what we do in that pull request it could be accepted, but I suggest that you check with our remote write team before investing time on this (see MAINTAINERS file).

liguozhong:
> You can also look at [prometheus/tsdb#602](https://github.com/prometheus/tsdb/pull/602)

Wow, great. Thank you very much for your guidance, please help me to review again.
image

roidelapluie:
Regardless of the code you don't have any tests. Can you y add some?

brian-brazil:
This would also cause issues with external labels, equals and regex are not the same thing when it comes to remote read.

Any optimisations on this front should be performed in the remote read endpoint.

liguozhong:
> Regardless of the code you don't have any tests. Can you y add some?

done.

image

liguozhong:
> This would also cause issues with external labels, equals and regex are not the same thing when it comes to remote read.
> 
> Any optimisations on this front should be performed in the remote read endpoint.

I agree, but when labelmatcher does not contain regular symbols, converting re to EQ has no effect on the result of remote read.
 If Prometheus server can provide this optimization by default, it is meaningful for dozens of remote storage implementations.At least in our case, the read time can be reduced from 10 seconds to 1 second.
Thank you for your review

brian-brazil:
This is false, it can and will change the result.

liguozhong:
OK, I close this pr

roidelapluie:
@brian-brazil I would welcome a better explanation about this and why an optimization we run on the tsdb querier can't happen in the remote read. Maybe a concrete example would help me understand. Thanks.

liguozhong:
When our customer uses Prometheus to monitor k8s clusters of more than 500 nodes, this invalid regular expression will lead to queries of more than 10 seconds. 
This feature is very important for our customer. Through the optimization of this regular transformation, the query is reduced to 1 second.

roidelapluie:
You could also optimize on the remote read end

liguozhong:
> You could also optimize on the remote read end

thanks . 
Our Prometheus branch has made a lot of modifications in response to customer feedback in the online environment.
There are many complaints from online customers that Prometheus is not available, such as scratch retry, regular conversion, excessive memory consumption, etc.
 At present, our modifications are only in Prometheus.So for our team, we don't need to make modifications in remote storage for the moment.
The purpose of this PR is to help users who meet remote storage read performance like us.For example, we should not modify the source code of the remote storage of the influxdb.

brian-brazil:
Because there's no requirement that a remote storage act like a tsdb.

roidelapluie:
> Because there's no requirement that a remote storage act like a tsdb.

That's what I was thinking too. Thx for the clarification

PR #6790, opened 2020-02-08
React UI: Use null value for determining consoles link display

juliusv:


juliusv:
@drumilpatel2000 When reading the last PR one time more this morning I noticed that we didn't actually make use of the explicit `null` value as intended (but didn't want to hold off on merging), so here's a small fix for that.

juliusv:
Thanks!

Issue #6301, opened 2019-11-11
React UI: Figure out how to "normalize" localhost URLs


juliusv:
The classic UI replaces the linked URL for targets that have a localhost-ish URL with the global URL, see: https://github.com/prometheus/prometheus/blob/37d666949c9475e298666cc8e0e787d76fa8fd77/web/web.go#L947-L984

We need to figure out how to do this with the new React UI as well.

brian-brazil:
Should we consider getting rid of this behaviour?

juliusv:
I think it depends on how ugly it would be to implement it for the new UI, since it is definitely useful, but not critical either.

Harkishen-Singh:
I would like to take up this. Thank you.

PR #6794, opened 2020-02-08
Adds normalization of localhost urls in targets page react

Harkishen-Singh:
Fixes: #6301 

targets_globalurl


juliusv:
Hmm, interesting approach, thanks! My initial idea was to inject yet another global variable into the app (like we do in https://github.com/prometheus/prometheus/blob/master/web/ui/react-app/public/index.html#L12-L24), and then only do the URL transformations on the client side.

However, now that I see the global URL added as another field in the targets API, I wonder if that might not make sense too, since other consumers than the UI might want to externally probe the metrics endpoints. So it would have value beyond the UI. OTHO one could assume that API consumers know the external URL of the Prometheus server, so they could also do that kind of transformation themselves anyway. @brian-brazil what do you think?

Harkishen-Singh:
> Hmm, interesting approach, thanks! My initial idea was to inject yet another global variable into the app (like we do in https://github.com/prometheus/prometheus/blob/master/web/ui/react-app/public/index.html#L12-L24), and then only do the URL transformations on the client side.
> 
> However, now that I see the global URL added as another field in the targets API, I wonder if that might not make sense too, since other consumers than the UI might want to externally probe the metrics endpoints. So it would have value beyond the UI. OTOH one could assume that API consumers know the external URL of the Prometheus server, so they could also do that kind of transformation themselves anyway. @brian-brazil what do you think?

@juliusv how about sending the globalURL in scrapeURL, if we do not have any problem showing it in the UI, a bit different from the old UI?

juliusv:
> how about sending the globalURL in scrapeURL, if we do not have any problem showing it in the UI, a bit different from the old UI?

@Harkishen-Singh We are not allowed to do that because it would be a breaking change to the targets API. But even if we were allowed, I think it might be misleading to display a URL that is different from the one that Prometheus actually scrapes. So I think either your current way or the global injected variable way are ok. Just would like to get a second opinion on adding this new field to the API.

web/web.go
```diff
@@ -289,6 +289,9 @@ func New(logger log.Logger, o *Options) *Handler {
 
 		ready: 0,
 	}
+	fmt.Println(o.ListenAddress)
```
juliusv: Remove debug statements

web/web.go
```diff
@@ -297,6 +300,11 @@ func New(logger log.Logger, o *Options) *Handler {
 			return *h.config
 		},
 		o.Flags,
+		api_v1.Options{
```
juliusv: `Options` is a bit generic of a name in this context, given that all the other parameters to `NewAPI()` are also options of sorts. Maybe eventually it'd be good to put everything into one big options struct, like we do in `web.go` for the web handler. But how about for now, calling it `GlobalURLOptions` or something like that?

web/api/v1/api.go
```diff
@@ -641,12 +683,22 @@ func (api *API) targets(r *http.Request) apiFuncResult {
...
-					Labels:             target.Labels().Map(),
-					ScrapePool:         key,
-					ScrapeURL:          target.URL().String(),
-					LastError:          lastErrStr,
+					DiscoveredLabels: target.DiscoveredLabels().Map(),
+					Labels:           target.Labels().Map(),
+					ScrapePool:       key,
+					ScrapeURL:        target.URL().String(),
+					GlobalURL:        globalURL.String(),
+					LastError: func() string {
```
juliusv: This seems like an unrelated change that shouldn't be in this PR?

web/web.go
```diff
@@ -297,6 +300,11 @@ func New(logger log.Logger, o *Options) *Handler {
...
+			Address: o.ListenAddress,
```
juliusv: Let's keep it called `ListenAddress` in the new struct as well.

web/api/v1/api.go
```diff
@@ -175,6 +176,7 @@ type API struct {
 	config                func() config.Config
 	flagsMap              map[string]string
 	ready                 func(http.HandlerFunc) http.HandlerFunc
+	option                Options
```
juliusv: `option` -> `options`, but see my other comment about naming it generally more specifically.

web/api/v1/api.go
```diff
@@ -602,6 +607,43 @@ type TargetDiscovery struct {
 	DroppedTargets []*DroppedTarget `json:"droppedTargets"`
 }
 
+// Options contains fields for globalURl required in local targets.
```
juliusv: I'd suggest:

```suggestion
// Options contains fields used for deriving the global URL for local targets.
```

web/web.go
Harkishen-Singh: Sure. Done.

web/api/v1/api.go
Harkishen-Singh: we already have a func `options` that implements an interface. Hence, it would be conflicting. how about `opts`?

web/api/v1/api.go
```diff
@@ -602,6 +607,43 @@ type TargetDiscovery struct {
...
+	Scheme        string
+}
+
+func getGlobalURL(u *url.URL, opts GlobalURLOptions) (*url.URL, error) {
+	host, port, err := net.SplitHostPort(u.Host)
+	if err != nil {
+		return u, err
+	}
+
+	for _, lhr := range []string{"127.0.0.1", "localhost"} {
```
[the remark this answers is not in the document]
juliusv: FWIW this was adapted from https://github.com/prometheus/prometheus/blob/40dd13b07420a044cc1b0ca57f639c572583d9c1/web/web.go#L976, maybe it'd be good to share more code here, but I didn't press for that because we assume that the old UI (along with its code) is going away anyway.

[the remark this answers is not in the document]
juliusv: (if we only end up keeping the new version of this code though, it would be good to keep the comments from the original)

web/api/v1/api.go
juliusv: Yeah, but `options` was too generic anyway, so now it'd be `globalURLOptions`?

web/ui/react-app/src/pages/targets/EndpointLink.tsx
```diff
@@ -22,7 +23,7 @@ const EndpointLink: FC<EndpointLinkProps> = ({ endpoint }) => {
 
   return (
     <>
-      <a href={endpoint}>{`${protocol}//${host}${pathname}`}</a>
+      <a href={globalUrl}>{`${protocol}//${host}${pathname}`}</a>
```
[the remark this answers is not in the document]
juliusv: Nope, only the link target should be changed, as in the old UI, so this should be correct.

We want to display what the Prometheus server has configured, but if it actually leads to the Prometheus server itself (localhost), we want to un-break that link for the user.

web/api/v1/api.go
Harkishen-Singh: This is actually done to capture the errors from the `globalURL` as well. Without this, it might be a `go-lint` error since we are not catching the error reported by the `net` library.

web/api/v1/api.go
Harkishen-Singh: sure.

web/api/v1/api.go
```diff
@@ -200,6 +205,7 @@ func NewAPI(
 	ar alertmanagerRetriever,
 	configFunc func() config.Config,
 	flagsMap map[string]string,
+	option GlobalURLOptions,
```
juliusv: This `option` parameter still needs to be renamed to `globalURLOptions`.

web/api/v1/api.go
```diff
@@ -602,6 +610,43 @@ type TargetDiscovery struct {
...
 }
 
+// GlobalURLOptions contains fields used for deriving the global URL for local targets.
+type GlobalURLOptions struct {
+	ListenAddress string
+	Host          string
+	Scheme        string
+}
+
+func getGlobalURL(u *url.URL, opts GlobalURLOptions) (*url.URL, error) {
```
juliusv: As mentioned aside in https://github.com/prometheus/prometheus/pull/6794#discussion_r377932616, could you keep (re-add) the comments from the original implementation of this function?

Likely only this copy of the function will survive in the long run, so it would be good to have the comments here as well.

web/web_test.go
```diff
@@ -391,8 +395,14 @@ func TestDebugHandler(t *testing.T) {
...
 		opts := &Options{
-			RoutePrefix: tc.prefix,
+			RoutePrefix:   tc.prefix,
+			ListenAddress: "somehost:9090",
+			ExternalURL: &url.URL{
+				Host:   "localhost.localdomain:9090",
+				Scheme: "http",
+			},
 		}
+		fmt.Println("opts is ", opts)
```
juliusv: Remove debug statement.

web/api/v1/api.go
juliusv: Ahhh, I see. The `LastError` field encodes scrape-related errors, so I don't think it's the right place for communicating errors related to computing the global URL (which is only relevant for UI purposes so far, and not scrape-related).

From what I can see, we pre-process both the external URL (in main.go) and the target's URL (in config.go) in such a way already that errors in `getGlobalURL()` are basically impossible (hopefully). So how about just failing the entire API call here in that case with an Internal Server Error, like in this example: https://github.com/prometheus/prometheus/blob/88af97366366fcbfff0af0f33d2da196a2afe629/web/api/v1/api.go#L1018-L1020

web/api/v1/api.go
Harkishen-Singh: Done

web/api/v1/api.go
Harkishen-Singh: Sure.

juliusv:
:+1: Thanks, looks good!

I'll merge, but we'll also need to update the example response in the documentation in https://github.com/prometheus/prometheus/blob/master/docs/querying/api.md#targets - would you be up for a follow-up PR to do that?

roidelapluie:
> +1 Thanks, looks good!
> 
> I'll merge, but we'll also need to update the example response in the documentation in https://github.com/prometheus/prometheus/blob/master/docs/querying/api.md#targets - would you be up for a follow-up PR to do that?

@juliusv @Harkishen-Singh Did that followup PR happen? We cut the RC in 3 days. Thanks.

juliusv:
@roidelapluie I don't think it ever happened, I'll just do it.

PR #6800, opened 2020-02-10
Cutting v2.16.0-rc.1

bwplotka:
@cstyan want to help a bit and cutting rc.1 with @roidelapluie fix. I also want to retry prombench to check if we can reproduce. So far we could not find the root cause of OOMs.

@krasi-georgiev can you confirm that nothing changed on prombench side between now and 2.15? Did we shrink the GKE instance sizes?


roidelapluie:
Please include https://github.com/prometheus/prometheus/pull/6793 in the release

krasi-georgiev:
> @krasi-georgiev can you confirm that nothing changed on prombench side between now and 2.15? Did we shrink the GKE instance sizes?

we decriesed the machine size, but this doesn't matter as both prometheus versions are started on the same machine size. In other words although the macine sizes are smaller both versions run on identical hardware so this shouldn't affect the memory usage.

bwplotka:
> we decriesed the machine size, but this doesn't matter as both prometheus versions are started on the same machine size. In other words although the macine sizes are smaller both versions run on identical hardware so this shouldn't affect the memory usage.

Well so we have our reason @krasi-georgiev  Memory usage does not need to have 100% characteristics. It might that now 2.16 exploded first because GC was a second slower, than 2.15, but second one will explode as well minutes after.

I think we need to get bigger machines back. I know that cost is bigger, but if we want to have reliable software for 10k (or more) users of Prometheus, we need to find sponsors for this.

bwplotka:
@roidelapluie I merged your fix and rebased. Let's run prombench.

/prombench v2.15.2

bwplotka:
cc @cstyan feel free to commandeer this PR once you wake up (:

yeya24:
Hi, is it okay to also include https://github.com/prometheus/prometheus/pull/6540 in the release? What do you think?

bwplotka:
Is it regression introduced by 2.16.0 or it was like this in 2.15.0 as well?

cstyan:
from the comments it looks like the issue fixed in #6540 was introduced in #6326 which would have gone into 2.15

bwplotka:
So let's cut this without fix I think. It's not critical.

cstyan:
I don't think the memory usage really looks like anything worth delaying the release any longer over. If anyone thinks otherwise or would like to see other queries I can look again in the morning, but after looking at the dashboards and other queries over today I think we're good to go :+1: 

2020-02-11-210818_2425x518_scrot
2020-02-11-210539_2439x526_scrot

bwplotka:
Yes, it looks definitely good to go. The difference is if anything extremely small. 👍 Let's run one day more then cancel prombench.

bwplotka:
All looks good based on https://matrix.to/#/!WaUKIfoqfiyWQhenET:matrix.org/$1581582672359244cpOdE:matrix.org?via=matrix.org

I think we are happy to cancel `prombench` once you are ready @cstyan

PR #6765, opened 2020-02-06
Keep kubernetes metrics in global vars

yeya24:
Signed-off-by: yeya24 <yb532204897@gmail.com>

Follow up PR for https://github.com/prometheus/prometheus/pull/6764. This PR changes the Kubernetes related metrics to global vars.

PR #6802, opened 2020-02-10
Use protobuf type in k8s client option

yeya24:

Protobuf has better performance than JSON

### DIFF
--- a/discovery/kubernetes/kubernetes.go
+++ b/discovery/kubernetes/kubernetes.go
@@ -130,13 +130,6 @@ func (c *NamespaceDiscovery) UnmarshalYAML(unmarshal func(interface{}) error) er
 func init() {
 	prometheus.MustRegister(eventCount)
 
-	// Initialize metric vectors.
-	for _, role := range []string{"endpoints", "node", "pod", "service", "ingress"} {
-		for _, evt := range []string{"add", "delete", "update"} {
-			eventCount.WithLabelValues(role, evt)
-		}
-	}
-
 	var (
 		clientGoRequestMetricAdapterInstance     = clientGoRequestMetricAdapter{}
 		clientGoWorkqueueMetricsProviderInstance = clientGoWorkqueueMetricsProvider{}
@@ -200,6 +193,8 @@ func New(l log.Logger, conf *SDConfig) (*Discovery, error) {
 	}
 
 	kcfg.UserAgent = "Prometheus/discovery"
+	kcfg.AcceptContentTypes = "application/vnd.kubernetes.protobuf,application/json"
+	kcfg.ContentType = "application/vnd.kubernetes.protobuf"
 
 	c, err := kubernetes.NewForConfig(kcfg)
 	if err != nil {


discovery/kubernetes/kubernetes.go
```diff
@@ -130,13 +130,6 @@ func (c *NamespaceDiscovery) UnmarshalYAML(unmarshal func(interface{}) error) er
 func init() {
 	prometheus.MustRegister(eventCount)
 
-	// Initialize metric vectors.
-	for _, role := range []string{"endpoints", "node", "pod", "service", "ingress"} {
-		for _, evt := range []string{"add", "delete", "update"} {
-			eventCount.WithLabelValues(role, evt)
-		}
-	}
```
yeya24: This is a clean up for #6765. Now k8s sd use global variables so we don't need to init them here

yeya24:
Any idea about this change?
I tested the memory between this PR and master branch. They have configured the same service monitor.
That's the memory usage, protobuf (the green one) uses less memory than JSON (the red one)

image

roidelapluie:
Hi there,

The upstream go-client is not using protobuf by default. One of the concerns rised was:

> "I'd probably prefer to continue to treat protobuf as an internal format, with the less strict guarantees we give around perfect compatibility (like we do for JSON)"
https://github.com/kubernetes/client-go/issues/76#issuecomment-278862622

https://github.com/kubernetes/client-go/issues/76
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#serialization-format


It turns out that as this looks like a great optimization, it might end up being a maintenance nightmare on our side as this could tie prometheus with specific versions of kubernetes.

yeya24:
> Thank you for your contribution.
> 
> The upstream go-client is not using protobuf by default. One of the concerns rised was:
> 
> > "I'd probably prefer to continue to treat protobuf as an internal format, with the less strict guarantees we give around perfect compatibility (like we do for JSON)"
> > [kubernetes/client-go#76 (comment)](https://github.com/kubernetes/client-go/issues/76#issuecomment-278862622)
> 
> [kubernetes/client-go#76](https://github.com/kubernetes/client-go/issues/76)
> https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#serialization-format
> 
> It turns out that as this looks like a great optimization, it might end up being a maintenance nightmare on our side as this could tie prometheus with specific versions of kubernetes.

Thanks for your reply!
From https://github.com/kubernetes/client-go/issues/76#issuecomment-274412872, it seems client-go supports protobuf format since 1.3. I am wondering do we still need to keep compatibility for the k8s version before 1.3?

Or maybe we can add a flag to enable protobuf, which defaults to true, what do you think? @roidelapluie

roidelapluie:
It does not mean that the kube team is willing to maintain protobuf compatibity accross releases, as stated in the quote in my previous comment.

simonpasquier:
@yeya24 can you move the clean-up to another PR? I'd agree with @roidelapluie that it seems risky to switch unconditionally to protobuf.
Maybe @brancz could chime in.

yeya24:
Thanks for the review. I agree with you this might be risky. So I am going to close this PR and open another PR to do the clean up.

roidelapluie:
Thank you. Please note that this pull request could have broken the proxy_url, basic_auth, tls options. I did not dig further in this direction, just adding this so we can check if we want to revisit in the future.

brancz:
I think using proto could be an explicit opt-in option at one point, that has best effort guarantees. I think closing this for now was the right choice. I wouldn't be against an opt in mechanism that has the guarantees clearly documented.

roidelapluie:
Thank you for your input @brancz :)

PR #6807, opened 2020-02-11
discovery/kubernetes: Expose label_selector and field_selector to config [Co-Authored-By @shuttie]

olagacek:
Its' based on  @shuttie PR https://github.com/prometheus/prometheus/pull/6096
I've picked up changes from that PR and extended documentation and added selectors validation


docs/configuration/configuration.md
```diff
@@ -854,6 +854,23 @@ tls_config:
...
+# only support selectors matching the role itself (e.g. node role can only contain node selectors).
+
+# Note: When making decision about using field/label selector make sure that this 
+# is the best approach - it will prevent Prometheus from reusing single list/watch
+# for all scrape configs. This might result in a bigger load on the Kubernetes API,
+# because per each selector combination there will be additional LIST/WATCH. On the other hand,
+# if you just want to monitor small subset of pods in large cluster it's recommended to use selectors.
+# Decision, if selectors should be used or not depends on the particular situation.
+selectors:
+  <role>:
```
roidelapluie: Should that be a list, to match what we have elsewhere:
```
[ -  kind: <string>
   [ label: <string> ] 
   [ field: <string> ] ]
```

roidelapluie:
I think we should we error if we specify twice the same role.

discovery/kubernetes/kubernetes.go
```diff
@@ -101,16 +122,94 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+			}
+
+			err = validateSelectors(selector, foundSelectorRoles)
+			if err != nil {
+				return err
+			}
+		}
+	default:
+		return errors.Errorf("role missing (one of: pod, service, endpoints, node, ingress)")
+	}
```
roidelapluie: I would move the check for duplicates here with a `for _, selector := range c.Selectors {`.

docs/configuration/configuration.md
```diff
@@ -854,6 +854,23 @@ tls_config:
...
+# only support selectors matching the role itself (e.g. node role can only contain node selectors).
+
+# Note: When making decision about using field/label selector make sure that this 
+# is the best approach - it will prevent Prometheus from reusing single list/watch
+# for all scrape configs. This might result in a bigger load on the Kubernetes API,
+# because per each selector combination there will be additional LIST/WATCH. On the other hand,
+# if you just want to monitor small subset of pods in large cluster it's recommended to use selectors.
+# Decision, if selectors should be used or not depends on the particular situation.
+[ selectors:
+  [ role: <string>
```
roidelapluie: This is no longer correct, you miss -.

discovery/kubernetes/kubernetes.go
```diff
@@ -91,6 +92,26 @@ type SDConfig struct {
 	Role               Role                         `yaml:"role"`
 	HTTPClientConfig   config_util.HTTPClientConfig `yaml:",inline"`
 	NamespaceDiscovery NamespaceDiscovery           `yaml:"namespaces,omitempty"`
+	Selectors          []ResourceSelectorConfigRaw  `yaml:"selectors,omitempty"`
+}
+
+type RoleSelectorConfig struct {
```
simonpasquier: AFAICT `RoleSelectorConfig` doesn't need to be exported and it doesn't need "yaml" field tags either.

discovery/kubernetes/kubernetes.go
```diff
@@ -91,6 +92,26 @@ type SDConfig struct {
...
+	Ingress   ResourceSelectorConfig `yaml:"ingress,omitempty"`
+}
+
+type ResourceSelectorConfigRaw struct {
+	Role  Role   `yaml:"role,omitempty"`
+	Label string `yaml:"label,omitempty"`
+	Field string `yaml:"field,omitempty"`
+}
+
+type ResourceSelectorConfig struct {
```
simonpasquier: same here?

discovery/kubernetes/kubernetes.go
olagacek: Check moved

olagacek:
> One comment and also:
> 
> If we know that in almost all cases the selectors are going to match the role, can we make them optional, and defaults them to the role? That would improve the user experience I think.

I'm not sure if it would improve user experience, what if user specifies two selectors, one with proper role and one without? should we return duplicated error? it might be confusing in some situations

discovery/kubernetes/kubernetes.go
```diff
@@ -101,19 +122,91 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+
+			err = validateSelectors(selector)
+			if err != nil {
+				return err
+			}
+		}
+	default:
+		return errors.Errorf("role missing (one of: pod, service, endpoints, node, ingress)")
+	}
+	foundSelectorRoles := make(map[Role]bool)
```
simonpasquier: (nit) I would use a map[Role]struct{} for `foundSelectorRoles` as it's more commonly used for detecting duplicates.
Also maybe you could validate selectors in the `for _, selector := range c.Selectors` loop instead of repeating the same code in each `switch` block?

discovery/kubernetes/kubernetes.go
```diff
@@ -101,19 +122,91 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+				return errors.Errorf("ingress role supports only ingress selectors")
+			}
+
+			err = validateSelectors(selector)
+			if err != nil {
+				return err
+			}
+		}
+	default:
+		return errors.Errorf("role missing (one of: pod, service, endpoints, node, ingress)")
```
simonpasquier: I would say `errors.Errorf("invalid role %q: expecting one of pod, servie, endpoints, node or ingress", c.Role)`

discovery/kubernetes/kubernetes.go
```diff
@@ -91,6 +92,26 @@ type SDConfig struct {
 	Role               Role                         `yaml:"role"`
 	HTTPClientConfig   config_util.HTTPClientConfig `yaml:",inline"`
 	NamespaceDiscovery NamespaceDiscovery           `yaml:"namespaces,omitempty"`
+	Selectors          []ResourceSelectorConfigRaw  `yaml:"selectors,omitempty"`
```
simonpasquier: Maybe rename the structs as follows:
* `ResourceSelectorConfigRaw` -> `SelectorConfig`
* `roleSelectorConfig` -> `roleSelector`
* `resourceSelectorConfig` -> `resourceSelector`

discovery/kubernetes/kubernetes.go
```diff
@@ -101,19 +122,70 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	default:
+		return errors.Errorf("invalid role: %q, expecting one of: pod, service, endpoints, node or ingress", c.Role)
+	}
+	foundSelectorRoles := make(map[Role]struct{})
+	for _, selector := range c.Selectors {
+		if _, ok := foundSelectorRoles[selector.Role]; ok {
+			return errors.Errorf("duplicated selector role: %s", selector.Role)
+		}
+		foundSelectorRoles[selector.Role] = struct{}{}
+		err := validateSelectors(selector)
```
simonpasquier: (nit) you can inline the code from `validateSelector()` here IMO

docs/configuration/configuration.md
```diff
@@ -854,6 +854,23 @@ tls_config:
...
+# only support selectors matching the role itself (e.g. node role can only contain node selectors).
+
+# Note: When making decision about using field/label selector make sure that this 
+# is the best approach - it will prevent Prometheus from reusing single list/watch
+# for all scrape configs. This might result in a bigger load on the Kubernetes API,
+# because per each selector combination there will be additional LIST/WATCH. On the other hand,
+# if you just want to monitor small subset of pods in large cluster it's recommended to use selectors.
+# Decision, if selectors should be used or not depends on the particular situation.
+[ selectors:
+  [ - role: <string>
```
roidelapluie:
```suggestion
  [ - role: <role>  
```

discovery/kubernetes/kubernetes.go
```diff
@@ -101,19 +122,70 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
-	}
 	err = c.HTTPClientConfig.Validate()
 	if err != nil {
 		return err
 	}
 	if c.APIServer.URL == nil && !reflect.DeepEqual(c.HTTPClientConfig, config_util.HTTPClientConfig{}) {
 		return errors.Errorf("to use custom HTTP client configuration please provide the 'api_server' URL explicitly")
 	}
+	switch c.Role {
+	case "pod":
```
roidelapluie: Should we use the RoleNode here? same for the others.

discovery/kubernetes/kubernetes.go
```diff
@@ -101,19 +122,70 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
-		return errors.Errorf("role missing (one of: pod, service, endpoints, node, ingress)")
-	}
 	err = c.HTTPClientConfig.Validate()
 	if err != nil {
 		return err
 	}
 	if c.APIServer.URL == nil && !reflect.DeepEqual(c.HTTPClientConfig, config_util.HTTPClientConfig{}) {
 		return errors.Errorf("to use custom HTTP client configuration please provide the 'api_server' URL explicitly")
 	}
+	switch c.Role {
```
simonpasquier: it could be shortened a bit using a map like this:

```go
var allowedSelectors = map[Role][]string{
   RolePod: []string{RolePod},
   RoleEndpoint: []string{RolePod, RoleEndpoint, RoleService},
   ...
}

...

for _, selector := range c.Selectors {
  if selector.Field != "" {
    break
  }
  var allowed bool
  for _, role := range allowedSelectors[c.Role] {
    if role == selector.Role {
      allowed = true
      break
    }
  }
  if !allowed {
    return errors.Errorf("%s role supports only %s selectors", c.Role, strings.Join(allowedSelectors[c.Role], ","))
  }
}
```

roidelapluie:
Thank you! Just two last comments on my side.

discovery/kubernetes/kubernetes.go
```diff
@@ -101,19 +122,70 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	case "pod":
+		for _, selector := range c.Selectors {
+			if selector.Role != RolePod && selector.Field != "" {
```
simonpasquier: shouldn't we test for `selector.Label` too?

olagacek: Actually I think that the incorrect role being present is enough to return error, no matter if field and label are empty, so I've reimplemented it so that we check only role correctness

discovery/kubernetes/kubernetes.go
```diff
@@ -101,16 +123,53 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err != nil {
 		return err
 	}
-	if c.Role == "" {
```
roidelapluie: was that removed by mistake?

olagacek: Yes, my bad. Thanks for pointing out :)

roidelapluie:
Thank you @olagacek 

I will merge this in the coming hours, I first want to test it on a kube cluster.

roidelapluie:
Thank you @olagacek ! I have rebased and squashed this in two commits (to maintain ownership) and pushed to master.

Commit: 8e53c19f9cff78fc9a14a8d372fcfef126e51612

Note: My manual test showed that the config files from the test will not work 'out of the box', e.g. the field selector should be `field: "status.phase=Running"`, not `field: "metadata.status=Running"`.

roidelapluie:
cc @brancz You might be interested to know that this PR has been merged and would be a nice addition to the prometheus operator.

olagacek:
Thank you @roidelapluie for merging! Can you tell me when can I expect it to be part of some release?

roidelapluie:
Yes I can. We make a release candidate every 6 weeks and the proper release a few days after.

The release candidate will be available around  2020-03-11. It will be prometheus v2.17.0.

simonpasquier:
@roidelapluie IMO it would have been nice to merge the PR rather than pushing directly to master...

image

roidelapluie:
> @roidelapluie IMO it would have been nice to merge the PR rather than pushing directly to master...
> 
> image

Yes, next time I will open another pull request with the two rebased commits, or ask the original author to squash the commits. Won't do it again. I don't know why #266 is indicated here however.

PR #6808, opened 2020-02-12
Storage concurrently tests and bug fix

liguozhong:
Storage concurrently tests and bug fix
pr https://github.com/prometheus/prometheus/pull/6770
issue #6714

### DIFF
--- a/storage/fanout/fanout_test.go
+++ b/storage/fanout/fanout_test.go
@@ -1,4 +1,4 @@
-// Copyright 2017 The Prometheus Authors
+// Copyright 2020 The Prometheus Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -31,10 +31,8 @@ func TestSelectSorted(t *testing.T) {
 
 	inputTotalSize := 0
 
-	//0: build test data
 	priStorage := teststorage.New(t)
 	defer priStorage.Close()
-	// A time series that has two samples and then goes stale.
 	app1, _ := priStorage.Appender()
 	app1.Add(inputLabel, 0, 0)
 	inputTotalSize++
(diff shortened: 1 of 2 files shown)


liguozhong:
A new fanout directory is created to add tests, because using teststorage.new (T) in the storage directory will cause the following error

# github.com/prometheus/prometheus/storage
FAIL	github.com/prometheus/prometheus/storage [setup failed]
import cycle not allowed in test
package github.com/prometheus/prometheus/storage (test)
	imports github.com/prometheus/prometheus/storage/tsdb
	imports github.com/prometheus/prometheus/storage

liguozhong:
pr #6770
Through the test case, it is found that in the last PR, one line of code is missing, resulting in the result being empty



image

liguozhong:
tests code
image

liguozhong:
tests PASS
image

storage/fanout/fanout_test.go
```diff
@@ -0,0 +1,107 @@
+// Copyright 2017 The Prometheus Authors
```
brian-brazil: 2020

storage/fanout/fanout_test.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+func TestSelectSorted(t *testing.T) {
+
+	inputLabel := labels.FromStrings(model.MetricNameLabel, "a")
+	outputLabel := labels.FromStrings(model.MetricNameLabel, "a")
+
+	inputTotalSize := 0
+
+	//0: build test data
```
brian-brazil: These comments don't add anything and should be removed.

This style of comment is usually also only when you're doing multiple tests inside one function.

storage/fanout/fanout_test.go
```diff
@@ -0,0 +1,107 @@
...
+	priStorage := teststorage.New(t)
+	defer priStorage.Close()
+	// A time series that has two samples and then goes stale.
```
brian-brazil: No need for these comments, plus they're incorrect.

storage/fanout/fanout_test.go
liguozhong: done

storage/fanout/fanout_test.go
liguozhong: done

storage/fanout/fanout_test.go
liguozhong: done

brian-brazil:
Thanks!

PR #6810, opened 2020-02-12
Merge fanout module into fanout directory

liguozhong:
Merge fanout module into fanout directory. 
When I was writing the test class, I found that fanout.go was easy to cause the problem of "import cycle", so I created a "fanout" folder. 
This PR is to move fanout from "storage" to "fanout" folder. Otherwise, there will be two files with the same name of "fanout_test.go"

image


brian-brazil:
This doesn't seem right. How were the tests passing if there was a cycle preventing a build?

liguozhong:
> This doesn't seem right. How were the tests passing if there was a cycle preventing a build?

storage/fanout_test.go->util/teststorage.go.New->storage/storage.Storage
image
image

liguozhong:
> This doesn't seem right. How were the tests passing if there was a cycle preventing a build?

The four methods of fanout.go (add addfast commit rollback) currently have no test classes, which should also be caused by the failure to import "teststorage. Go.New"

gouthamve:
Hi @liguozhong, thanks for the PR, but unfortunately a lot of the code around this has been refactored. Could you see if the change you want is still relevant with the new code? If yes, could you open another PR for it?

PR #6812, opened 2020-02-12
scrape: fix goroutine leak in test

BurtonQin:
There are 7 test functions sharing a similar problem: some goroutines in these functions will be leaked in certain cases (like timeout).

Although t.Fatal() is often called in these certain cases, it won't stop the leaking, because "[Calling FailNow does not stop those other goroutines.](https://golang.org/pkg/testing/#T.FailNow)"

This patch adds one buffer to the channel, making the send operation never block the child goroutine, and it will not change semantic.

Note that I also remove `close(errc)` on L1425, which is useless.

### DIFF
--- a/scrape/scrape_test.go
+++ b/scrape/scrape_test.go
@@ -427,7 +427,7 @@ func nopMutator(l labels.Labels) labels.Labels { return l }
 
 func TestScrapeLoopStop(t *testing.T) {
 	var (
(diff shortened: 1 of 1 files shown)


brian-brazil:
The tests are failing.

brian-brazil:
Thanks!

PR #6820, opened 2020-02-14
Break DB.Compact and DB.CompactHead and DB.CompactBlocks

codesome:
This just modularises the components in the `DB.Compact()` while having no changes in the logic. This will help external projects like Cortex get access to individual methods (mainly `CompactHead`).

/cc @thorfour @pracucci


thorfour:
These changes look great! Thank you!

PR #6827, opened 2020-02-16
storage/remote: remove redundant RLock

lzhfromustc:
***Description***
The problem is in methods of struct `*shards`. I believe the `s.mtx.RLock()` in `stop()` can be removed. A RLock is useful only when there may be other goroutines holding Lock at the same time. However, `s.mtx.Lock()` is only used in `start()` and `stop()`. `start()` has annotation: "start the shards; must be called before any call to enqueue", so it can't be running at the same time as `stop()`. Also `stop()` cannot run in parallel with another `stop()`, because that will trigger a panic due to closing a channel twice.

`start()`:
https://github.com/prometheus/prometheus/blob/384cba98fec550052292dcc3095ed16fbf197087/storage/remote/queue_manager.go#L686-L689

`stop()`:
https://github.com/prometheus/prometheus/blob/384cba98fec550052292dcc3095ed16fbf197087/storage/remote/queue_manager.go#L709-L723

Tests still pass after this commit


roidelapluie:
(In addition to that closing a channel is not a read operation)

cstyan:
I'm not sure this is going to be correct, I'll have to spend some more time thinking about it, but just because the tests are passing doesn't mean this couldn't be introducing a subtle race condition.

There are multiple go routines at play here; the `QueueManager`'s `Append` and therefore `shards.enqueue` will be called as a result of the WAL `Watcher` reading samples from a WAL segment, and `shards.stop` can happen as part of resharding or `QueueManager.Stop` (which can be called as a result of a config reload or as part of Prometheus shutting down).

gouthamve:
@cstyan Could you take a look at this again?

cstyan:
Looking again this appears to be safe, for the reasons Chris has mentioned, but I'd really prefer to see a refactor/general clean up that makes the resharding and shutdown paths easier to read and reason about instead of just a removal of this read lock call.

roidelapluie:

Given @cstyan 's comment, we have decided to close this pull request, however, @cstyan , can  you please asses if you want the rewrite/simplification, and if so, describe it  an issue to keep track of this work.

Issue #6178, opened 2019-10-18
Allow filtering label values API with matchers


brancz:
## Proposal

I would like to propose allowing our label values API to be filter-able by matchers, just like we [already do with the series API](https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers).

**Use case. Why is this important?**

In particular we have two use cases: 

* Autocompletion: a language server implementation (that we are actively working on) can perform requests against a live system for advanced auto-complete, this could help us create a really nice auto complete for label values, as we can cut down the possible values using the already provided selectors.
* Cutting down on cognitive overload: we have a bunch of rather large Prometheus servers that are shared by teams and the mix of concerns is causing even just metric autocompletion to be kind of a nightmare, so the ability to isolate this for teams would allow us to provide a way better experience to users.

Beyond that, it would be great if our APIs were consistent in what they offer so a third point :slightly_smiling_face: .

I brought this up on IRC and multiple people seemed to be for this idea so I felt it was worth opening an issue on.

krasi-georgiev:
+1 this should also help for other Thanos use cases

brancz:
This doesn’t have anything to do with Thanos per se, but whatever we end up doing here Thanos should do the same for compatibility with the Prometheus API.

krasi-georgiev:
I think for good performance should also pass a time range.

For example in the case of auto completion when you make an instant query you are interested in labels only from the last scrape and for a range query only for the time that this range query covers.

Otherwise getting label names/values for the entire DB involves reading all index files of all current blocks including the head block. 

Or a different approach would be using some cache but again for auto completion it might make more sense to show only relevant label names/values based on the query that the user would want to run.

brancz:
Yeah the series API allows this as well. I think that also makes sense here.

valyala:
FYI, VictoriaMetrics supports passing `match[]`, `start` and `end` args to `/api/v1/label/.../values` since `v1.24.0` - see [the corresponding commit](https://github.com/VictoriaMetrics/VictoriaMetrics/commit/99e048c9dfae9a551bafd96855472449f8ea5c3f).
For instance, the following query would return only metric names for `{job="node_exporter"}`:
```
http://localhost:8428/api/v1/label/__name__/values -d 'match[]={job="node_exporter"}'
```

krasi-georgiev:
@valyala any chance to contribute the same to Prometheus?

valyala:
[...]

PR #6837, opened 2020-02-18
WIP: Filter labels for matching series by name

boyskila:
Fixes: https://github.com/prometheus/prometheus/issues/6178

I'm going to label this PR as WIP since IMO the filter can be extended to handle more scenarios like those described [here](https://docs.google.com/document/d/1sbR7WLWrh1f6ptxX3-92sNM4aP1skc5R44LxnikxGVY/edit) and for this reason, I'm skipping tests and docs examples for now.
@krasi-georgiev @juliusv @brancz What is your opinion?

UPDATE [@krasi-georgiev ]
Here is a simple doc with few examples
https://docs.google.com/document/d/1sbR7WLWrh1f6ptxX3-92sNM4aP1skc5R44LxnikxGVY/edit

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -477,6 +477,41 @@ func (api *API) labelNames(r *http.Request) apiFuncResult {
 	return apiFuncResult{names, nil, warnings, nil}
 }
 
+var (
+	minTime = time.Unix(math.MinInt64/1000+62135596801, 0).UTC()
+	maxTime = time.Unix(math.MaxInt64/1000-62135596801, 999999999).UTC()
+
+	minTimeFormatted = minTime.Format(time.RFC3339Nano)
+	maxTimeFormatted = maxTime.Format(time.RFC3339Nano)
+)
+
+func parseTimeParam(r *http.Request, paramName string, defaultValue time.Time) (time.Time, error) {
+	val := r.FormValue(paramName)
+	if val != "" {
+		result, err := parseTime(val)
+		if err != nil {
+			return result, errors.Wrapf(err, "Invalid parameter '%s'", paramName)
+		}
+		return result, nil
+	}
+	return defaultValue, nil
+}
+
+func selectSeriesByMatchers(q storage.Querier, matcherSets [][]*labels.Matcher) ([]storage.SeriesSet, error, storage.Warnings) {
+	var sets []storage.SeriesSet
+	var warnings storage.Warnings
+	for _, mset := range matcherSets {
+		s, wrn, err := q.Select(nil, mset...)
+		warnings = append(warnings, wrn...)
+		if err != nil {
+			return nil, err, warnings
+		}
+		sets = append(sets, s)
+	}
+
+	return sets, nil, nil
+}
+
 func (api *API) labelValues(r *http.Request) apiFuncResult {
 	ctx := r.Context()
 	name := route.Param(ctx, "name")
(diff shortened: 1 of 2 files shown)


web/api/v1/api.go
```diff
@@ -477,13 +477,93 @@ func (api *API) labelNames(r *http.Request) apiFuncResult {
...
 
+var (
+	minTime = time.Unix(math.MinInt64/1000+62135596801, 0).UTC()
+	maxTime = time.Unix(math.MaxInt64/1000-62135596801, 999999999).UTC()
+
+	minTimeFormatted = minTime.Format(time.RFC3339Nano)
+	maxTimeFormatted = maxTime.Format(time.RFC3339Nano)
+)
+
+func parseTimeParam(r *http.Request, paramName string, defaultValue time.Time) (time.Time, error) {
```
krasi-georgiev: to speed up the reviews such changes that can be indipendent from the main PR could be opened as a separate PR

boyskila: Ok, understand. What do you suggest - to revert this back and to open another PR?

krasi-georgiev: I think it might be fine, the PR is small enough as is.

krasi-georgiev: why not return timestamps here to avoid the duplicate call to `timestamp.FromTime`

web/api/v1/api.go
```diff
@@ -477,13 +477,93 @@ func (api *API) labelNames(r *http.Request) apiFuncResult {
...
+		}
+
+		q, err := api.Queryable.Querier(ctx, timestamp.FromTime(start), timestamp.FromTime(end))
+		defer q.Close()
+
+		if err != nil {
+			return apiFuncResult{nil, &apiError{errorExec, err}, nil, nil}
+		}
+
+		matches, err, warnings := selectSeriesByMatchers(q, [][]*labels.Matcher{selector})
```
krasi-georgiev:
```suggestion
		sets, err, warnings := selectSeriesByMatchers(q, [][]*labels.Matcher{selector})
```

web/api/v1/api.go
```diff
@@ -477,13 +477,93 @@ func (api *API) labelNames(r *http.Request) apiFuncResult {
...
+			return apiFuncResult{nil, &apiError{errorExec, err}, warnings, nil}
+		}
+
+		series := storage.NewMergeSeriesSet(matches, nil)
+		result := make(map[string][]string)
+		for series.Next() {
+			label := series.At().Labels().Get(name)
+			if label != "" && result[label] == nil {
+				result[name] = append(result[name], label)
+				result[label] = result[name] // Prevent from adding same label value twice.
```
krasi-georgiev: I think we should be able to simplify this a bit, but lets have a look at the tests and will revisit

web/api/v1/api.go
boyskila: Yes, good catch, thanks!

boyskila: From the other side returning ```Time``` is more flexible - you can take advantage of its methods but I guess opposite conversion is also possible(from ```int64``` to ```Time```)

gouthamve:
@krasi-georgiev Could you take another look at this again?

krasi-georgiev:
That is quite outdated now, I think https://github.com/prometheus/prometheus/pull/7496 takes over this one, but I don't think I will be able to spend any time for a review - working on some Thanos tasks.

PR #6838, opened 2020-02-18
discovery/kubernetes: EndpointSlice discovery

brancz:
Kubernetes is introducing a new API for discovering endpoints: the `EndpointSlice` API. The difference between this and the existing `Endpoints` API, which we already have a discovery mechanism for, is that at most 100 endpoints are in one `EndpointSlice` object, which makes a different with large scale deployments where if a single Endpoint changes only the `EndpointSlice` that endpoint is part of is updated and communicated, in this case to Prometheus, instead of having to communicate all potentially 100s of thousands of endpoints in one object at most a small object with 100 endpoints is transferred.

The third commit contains a fix to a race that is present in master as well, but became pretty apparent as I was testing these changes.

@simonpasquier @beorn7 @brian-brazil @roidelapluie

### DIFF
--- a/discovery/kubernetes/kubernetes.go
+++ b/discovery/kubernetes/kubernetes.go
@@ -82,7 +82,7 @@ func (c *Role) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		return err
 	}
 	switch *c {
-	case RoleNode, RolePod, RoleService, RoleEndpoint, RoleIngress:
+	case RoleNode, RolePod, RoleService, RoleEndpoint, RoleEndpointSlice, RoleIngress:
 		return nil
 	default:
 		return errors.Errorf("unknown Kubernetes SD role %q", *c)
--- a/discovery/kubernetes/kubernetes.go
+++ b/discovery/kubernetes/kubernetes.go
@@ -27,6 +27,7 @@ import (
 	config_util "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
 	apiv1 "k8s.io/api/core/v1"
+	disv1beta1 "k8s.io/api/discovery/v1beta1"
 	"k8s.io/api/networking/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
@@ -67,11 +68,12 @@ type Role string
 
 // The valid options for Role.
 const (
-	RoleNode     Role = "node"
-	RolePod      Role = "pod"
-	RoleService  Role = "service"
-	RoleEndpoint Role = "endpoints"
-	RoleIngress  Role = "ingress"
+	RoleNode          Role = "node"
+	RolePod           Role = "pod"
+	RoleService       Role = "service"
+	RoleEndpoint      Role = "endpoints"
+	RoleEndpointSlice Role = "endpointslice"
+	RoleIngress       Role = "ingress"
 )
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
@@ -97,11 +99,12 @@ type SDConfig struct {
 }
 
 type roleSelector struct {
-	node      resourceSelector
-	pod       resourceSelector
-	service   resourceSelector
-	endpoints resourceSelector
-	ingress   resourceSelector
+	node          resourceSelector
+	pod           resourceSelector
+	service       resourceSelector
+	endpoints     resourceSelector
+	endpointslice resourceSelector
+	ingress       resourceSelector
 }
 
 type SelectorConfig struct {
@@ -124,7 +127,7 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		return err
 	}
 	if c.Role == "" {
-		return errors.Errorf("role missing (one of: pod, service, endpoints, node, ingress)")
+		return errors.Errorf("role missing (one of: pod, service, endpoints, endpointslice, node, ingress)")
 	}
 	err = c.HTTPClientConfig.Validate()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


discovery/kubernetes/kubernetes.go
```diff
@@ -136,11 +139,12 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
 	foundSelectorRoles := make(map[Role]struct{})
 	allowedSelectors := map[Role][]string{
-		RolePod:      {string(RolePod)},
-		RoleService:  {string(RoleService)},
-		RoleEndpoint: {string(RolePod), string(RoleService), string(RoleEndpoint)},
-		RoleNode:     {string(RoleNode)},
-		RoleIngress:  {string(RoleIngress)},
+		RolePod:           {string(RolePod)},
+		RoleService:       {string(RoleService)},
+		RoleEndpointSlice: {string(RolePod), string(RoleService), string(RoleEndpointSlice)},
```
roidelapluie:
```suggestion
		RoleEndpointSlice: {string(RolePod), string(RoleService), string(RoleEndpointSlice)},
```
Are you sure about this one?

brancz: Yes I'm sure the EndpointSlice role should have EndpointSlice selector. Maybe I'm misunderstanding your question. Can you be more specific?

roidelapluie:
Question: is that something we want explicit or should we switch to that if the version of kubernetes supports it?

discovery/kubernetes/kubernetes.go
roidelapluie: See my other question :)

brancz:
> Question: is that something we want explicit or should we switch to that if the version of kubernetes supports it?

I was thinking about this a lot, and while I would have liked a transparent migration, I don't think it's possible as the new API is different enough from the standpoint of the meta labels it produces, that I don't think we can or should do that.

roidelapluie:
> > Question: is that something we want explicit or should we switch to that if the version of kubernetes supports it?
> 
> I was thinking about this a lot, and while I would have liked a transparent migration, I don't think it's possible as the new API is different enough from the standpoint of the meta labels it produces, that I don't think we can or should do that.

Thank you. I notice that the documentation has not been updated yet. I will review this later.

brancz:
> I notice that the documentation has not been updated yet.

You're totally right. I'll make sure to add that.

brian-brazil:
My question is the same as Julien's. I think we should treat this as an optimisation that we do under the covers. If it can't produce the same labels, that sounds like something to take up with upstream.

brancz:
This is an intentionally new API and the existing Kubernetes API is not going anywhere. They are just similar for historic reasons. I think what we can do, since, the Kubernetes SD is marked experimental, once Kubernetes marks that API as stable we drop support for the "legacy" Endpoints API.

roidelapluie:
Can you comment about how different are the labels?

roidelapluie:
@brancz I pushed a commit to your branch to make this runnable.

roidelapluie:
Notes: We should not hide endpointslices behind endpoints:

> Note: Although EndpointSlices may eventually replace Endpoints, many Kubernetes components still rely on Endpoints. For now, enabling EndpointSlices should be seen as an addition to Endpoints in a cluster, not a replacement for them.

https://kubernetes.io/docs/tasks/administer-cluster/enabling-endpointslices/

roidelapluie:
I am wondering wheter we should use endpointslices or endpoint_slices in strings.

roidelapluie:
I need to dig deeper; it looks like I don't have the same amount of endpoints and endpointslices:

ep1


Config:

```
- job_name: kubernetes-endpoints
  kubernetes_sd_configs:
  - role: endpoints
- job_name: kubernetes-endpointslice
  kubernetes_sd_configs:
  - role: endpointslice
```

@brancz are you sure we loop into the endpointslices?

ep2

roidelapluie:
@brancz censoring has been updated in master. That also requires a cherry pick of your race fix. Thank you.

brian-brazil:
> Notes: We should not hide endpointslices behind endpoints:

I'm not seeing how you're coming to that conclusion from that text.

roidelapluie:
The text says we should not threat that as a replacement. So let's not do this?

Additionally, they use different rbac rules and even are in a non-core API group. so that should not happen behind our users' back I think.

brian-brazil:
It sounds like it's a replacement to me, just not there yet. If we can opportunistically use it and get all the same metadata, I think we should.

roidelapluie:
The sentence indeed says 'for now'. What about then still having a flag in the config that we remove when it is stable?

brian-brazil:
I'm not sure that really works, some users may be running older versions of k8 that don't support it for quite some time.

brancz:
Replace is meant as the other API may be deprecated in its favor. It’s not and was never meant to be a full drop in replacement. If that had been possible then the changes would have been done in the endpoints API itself.

brancz:
I realized another important factor why we can't just transparently switch the endpoints discovery underneath the users: it requires different/additional permissions to be granted to the Prometheus server, which risks breaking user's setups without them noticing/knowing.

@roidelapluie have you been able to discover more targets in another try? (clusters v1.16 and older have to explicitly enable the feature gates on the controller manager, even if the API is available)

roidelapluie:
Yes I noticed the different rbac rules too: https://github.com/prometheus/prometheus/pull/6838#issuecomment-590834805

But I did not dig deeper into why I did not have 1-to-1 targets between both mechanisms.

roidelapluie:
I rebased on top of master

roidelapluie:
cc @howardjohn

discovery/kubernetes/kubernetes.go
```diff
@@ -82,7 +82,7 @@ func (c *Role) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		return err
 	}
 	switch *c {
-	case RoleNode, RolePod, RoleService, RoleEndpoint, RoleIngress:
+	case RoleNode, RolePod, RoleService, RoleEndpoint, RoleEndpointSlice, RoleIngress:
```
roidelapluie, line 85: @brancz You were missing this.

brancz:
I guess since we both did work on this PR it would be good to get a third person to review this.

roidelapluie:
Yes, even if my work was only rebasing to master due to k8s dependency updates.

PR #6840, opened 2020-02-18
Don't return error in ContextFromRequest function.

pstibrany:
Previously `httputil.ContextFromRequest` function could return error if `http.Request.RemoteAddr` field didn't have correct format, but since this field explicitly has no defined format, that was little too strict.

We have run into this issue when integrating [Prometheus 2.16.0 into Cortex](https://github.com/cortexproject/cortex/pull/2088).

### DIFF
--- a/util/httputil/context.go
+++ b/util/httputil/context.go
@@ -34,23 +34,20 @@ func ContextWithPath(ctx context.Context, path string) context.Context {
 // ContextFromRequest returns a new context with identifiers of
 // the request to be used later when logging the query.
 func ContextFromRequest(ctx context.Context, r *http.Request) context.Context {
-	reqCtxVal := map[string]string{
-		"method": r.Method,
+	var ip string
+	if r.RemoteAddr != "" {
+		// r.RemoteAddr has no defined format, so don't return error if we cannot split it into IP:Port.
+		ip, _, _ = net.SplitHostPort(r.RemoteAddr)
 	}
-
-	// r.RemoteAddr has no defined format, so don't return error if we cannot split it into IP:Port.
-	ip, _, _ := net.SplitHostPort(r.RemoteAddr)
-	if ip != "" {
-		reqCtxVal["clientIP"] = ip
-	}
-
 	var path string
 	if v := ctx.Value(pathParam); v != nil {
 		path = v.(string)
-		reqCtxVal["path"] = path
 	}
-
 	return promql.NewOriginContext(ctx, map[string]interface{}{
-		"httpRequest": reqCtxVal,
+		"httpRequest": map[string]string{
+			"clientIP": ip,
+			"method":   r.Method,
+			"path":     path,
+		},
 	})
 }
--- a/util/httputil/context.go
+++ b/util/httputil/context.go
@@ -31,7 +31,7 @@ func ContextWithPath(ctx context.Context, path string) context.Context {
 	return context.WithValue(ctx, pathParam, path)
 }
 
-// ContextFromRequest returns a new context from a requests with identifiers of
+// ContextFromRequest returns a new context with identifiers of
 // the request to be used later when logging the query.
 func ContextFromRequest(ctx context.Context, r *http.Request) context.Context {
 	reqCtxVal := map[string]string{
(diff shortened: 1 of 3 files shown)


util/httputil/context.go
```diff
@@ -33,20 +33,24 @@ func ContextWithPath(ctx context.Context, path string) context.Context {
 
 // ContextFromRequest returns a new context from a requests with identifiers of
 // the request to be used later when logging the query.
-func ContextFromRequest(ctx context.Context, r *http.Request) (context.Context, error) {
-	ip, _, err := net.SplitHostPort(r.RemoteAddr)
-	if err != nil {
-		return ctx, err
+func ContextFromRequest(ctx context.Context, r *http.Request) context.Context {
+	m := map[string]string{
```
bwplotka: why `m`? `reqCtxVal` maybe to be more explicit?

pstibrany: Given how small the function is, `m` looked enough to me, but I can change it to `reqCtxVal` as well.

util/httputil/context.go
```diff
@@ -33,20 +33,24 @@ func ContextWithPath(ctx context.Context, path string) context.Context {
 
 // ContextFromRequest returns a new context from a requests with identifiers of
```
simonpasquier: LGTM. Can you just fix this comment?

```suggestion
// ContextFromRequest returns a new context from a request with identifiers of
```

pstibrany: what about?

```suggestion
// ContextFromRequest returns a new context with identifiers of 
// the request to be used later when logging the query.
```

roidelapluie:
Can you please fix DCO?

pstibrany:
> Can you please fix DCO?

Done. Thanks, missed the last commit.

roidelapluie:
Is there a real-life situation where that can error?

roidelapluie:
This looks strange to me. I would like @brian-brazil view on this one. I think that that error should not happen in real situation. If that can happen in real life, we might trigger a 1.15.3 release.

From my understanding however, this is more about crafting http requests that don't go over http. In which case at the moment you craft the request, you could manually set a RemoteAddr, e.g. 127.0.0.2:0 or something like that.

pstibrany:
> Is there a real-life situation where that can error?

This field is documented to have undefined format, and only uses “IP:port” when request object is created by HTTP server in net/http package.

In our case, we construct request objects manually (based on requests in the frontend queue), and pass them to prometheus v1.API instance. As we don’t keep RemoteAddr, we need to fake this value now, otherwise API instance complains.

brian-brazil:
> This field is documented to have undefined format, and only uses “IP:port” when request object is created by HTTP server in net/http package.

So that's a no then in Prometheus terms. Having the code simpler and more resilient doesn't hurt though.

pstibrany:
> So that's a no then in Prometheus terms. Having the code simpler and more resilient doesn't hurt though.

I understand that Prometheus doesn't need this change, but given the size, code simplification (why should context creation fail?) and the fact that it helps your Cortex friends, perhaps we can still get it in?

roidelapluie:
> > So that's a no then in Prometheus terms. Having the code simpler and more resilient doesn't hurt though.
> 
> I understand that Prometheus doesn't need this change, but given the size, code simplification (why should context creation fail?) and the fact that it helps your Cortex friends, perhaps we can still get it in?

Do we want to run a query if we are not able to log the client IP? Is there a security implication?

pstibrany:
> Do we want to run a query if we are not able to log the client IP? Is there a security implication?

I assume this question was not for me. Just a note about client IPs -- what you really get here is possibly the address of last intermediate proxy, and not real client IP, so it's already best-effort only.

pstibrany:
> But indeed we are changing logic bit here: Maybe someone relies on clientIP?

In Prometheus, `RemoteAddr` will always be in IP:port form, so there should be no change here.

brian-brazil:
> I think @brian-brazil meant that we are happy to have this in, even though it does not help Prometheus in current moment (:

Yes, that's my meaning. This was actually a niggle when I was reviewing it, but I didn't bring it up.

Issue #4580, opened 2018-09-05
Float rounding in recording doesn't match runtime


bobrik:
## Bug Report

**What did you do?**

I added a recording rule:

```
# Total CPU usage by mode and instance
- record: instance_mode:node_cpu:sum
  expr: sum(node_cpu) without (cpu)
```

**What did you expect to see?**

Recording rule matches expression at run time, counter never goes back.

**What did you see instead? Under which circumstances?**

Runtime calculation:

```json
{
  "status": "success",
  "data": {
    "resultType": "matrix",
    "result": [
      {
        "metric": {},
        "values": [
          [
            1536185100,
            "34438.13999999998"
          ],
          [
            1536185130,
            "34438.13999999998"
          ],
          [
            1536185160,
            "34438.13999999998"
          ],
          [
            1536185190,
            "34438.13999999998"
          ],
          [
            1536185220,
            "34438.13999999998"
          ],
          [
            1536185250,
            "34438.13999999998"
          ],
          [
            1536185280,
            "34438.13999999998"
          ],
          [
            1536185310,
            "34438.13999999998"
          ],
          [
            1536185340,
            "34438.14999999998"
          ],
          [
            1536185370,
            "34438.14999999998"
          ],
          [
            1536185400,
            "34438.14999999998"
          ],
          [
            1536185430,
            "34438.14999999998"
          ],
          [
            1536185460,
            "34438.14999999998"
          ]
        ]
      }
    ]
  }
}
```

Rule based calculation:

```json
{
  "status": "success",
  "data": {
    "resultType": "matrix",
    "result": [
      {
        "metric": {
          "__name__": "instance_mode:node_cpu:sum",
          "colo_id": "28",
          "colo_name": "sea01",
          "instance": "myhost",
          "job": "node_exporter",
          "mode": "nice",
          "node_type": "metal",
          "region": "wnam"
        },
        "values": [
          [
            1536185100,
            "34438.139999999985"
          ],
          [
            1536185130,
            "34438.139999999985"
          ],
          [
            1536185160,
            "34438.13999999999"
          ],
          [
            1536185190,
            "34438.13999999999"
          ],
          [
            1536185220,
            "34438.13999999997"
          ],
          [
            1536185250,
            "34438.13999999997"
          ],
          [
            1536185280,
            "34438.14"
          ],
          [
            1536185310,
            "34438.14"
          ],
          [
            1536185340,
            "34438.149999999994"
          ],
          [
            1536185370,
[...]

PR #6841, opened 2020-02-18
Port isolation from old TSDB PR

beorn7:
The original PR was https://github.com/prometheus/tsdb/pull/306 .

I tried to carefully adjust to the new world order, but please give this a very careful review, especially around iterator reuse (marked with a TODO).

On the bright side, I definitely found and fixed a bug in txRing.

### DIFF
--- a/tsdb/db_test.go
+++ b/tsdb/db_test.go
@@ -87,6 +87,10 @@ func query(t testing.TB, q storage.Querier, matchers ...*labels.Matcher) map[str
 		}
 		testutil.Ok(t, it.Err())
 
+		if len(samples) == 0 {
+			continue
+		}
+
 		name := series.Labels().String()
 		result[name] = samples
 	}
@@ -1276,20 +1280,29 @@ func TestNotMatcherSelectsLabelsUnsetSeries(t *testing.T) {
 		testutil.Ok(t, err)
 		testutil.Equals(t, 0, len(ws))
 
-		lres, err := expandSeriesSet(ss)
+		lres, _, err := expandSeriesSet(ss)
 		testutil.Ok(t, err)
-
 		testutil.Equals(t, c.series, lres)
 	}
 }
 
-func expandSeriesSet(ss storage.SeriesSet) ([]labels.Labels, error) {
-	result := []labels.Labels{}
+// expandSeriesSet returns the raw labels in the order they are retrieved from
+// the series set and the samples keyed by Labels().String().
+func expandSeriesSet(ss storage.SeriesSet) ([]labels.Labels, map[string][]sample, error) {
+	resultLabels := []labels.Labels{}
+	resultSamples := map[string][]sample{}
 	for ss.Next() {
-		result = append(result, ss.At().Labels())
+		series := ss.At()
+		samples := []sample{}
+		it := series.Iterator()
+		for it.Next() {
+			t, v := it.At()
+			samples = append(samples, sample{t: t, v: v})
+		}
+		resultLabels = append(resultLabels, series.Labels())
+		resultSamples[series.Labels().String()] = samples
 	}
-
-	return result, ss.Err()
+	return resultLabels, resultSamples, ss.Err()
 }
 
 func TestOverlappingBlocksDetectsAllOverlaps(t *testing.T) {
@@ -2475,6 +2488,136 @@ func TestDBReadOnly_FlushWAL(t *testing.T) {
 	}
 	testutil.Ok(t, seriesSet.Err())
 	testutil.Equals(t, 1000.0, sum)
+}
+
+func TestDBCannotSeePartialCommits(t *testing.T) {
+	tmpdir, _ := ioutil.TempDir("", "test")
+	defer os.RemoveAll(tmpdir)
+
+	db, err := Open(tmpdir, nil, nil, nil)
+	testutil.Ok(t, err)
+	defer db.Close()
+
+	stop := make(chan struct{})
+	firstInsert := make(chan struct{})
+
+	// Insert data in batches.
+	go func() {
+		iter := 0
+		for {
+			app := db.Appender()
+
+			for j := 0; j < 100; j++ {
+				_, err := app.Add(labels.FromStrings("foo", "bar", "a", strconv.Itoa(j)), int64(iter), float64(iter))
+				testutil.Ok(t, err)
+			}
+			err = app.Commit()
+			testutil.Ok(t, err)
+
+			if iter == 0 {
+				close(firstInsert)
+			}
+			iter++
+
+			select {
+			case <-stop:
+				return
+			default:
+			}
+		}
+	}()
+
+	<-firstInsert
+
+	// This is a race condition, so do a few tests to tickle it.
+	// Usually most will fail.
+	inconsistencies := 0
+	for i := 0; i < 10; i++ {
+		func() {
+			querier, err := db.Querier(context.Background(), 0, 1000000)
+			testutil.Ok(t, err)
+			defer querier.Close()
+
+			ss, _, err := querier.Select(nil, labels.MustNewMatcher(labels.MatchEqual, "foo", "bar"))
+			testutil.Ok(t, err)
+
+			_, seriesSet, err := expandSeriesSet(ss)
+			testutil.Ok(t, err)
+			values := map[float64]struct{}{}
+			for _, series := range seriesSet {
+				values[series[len(series)-1].v] = struct{}{}
+			}
+			if len(values) != 1 {
+				inconsistencies++
+			}
+		}()
+	}
+	stop <- struct{}{}
+
+	testutil.Equals(t, 0, inconsistencies, "Some queries saw inconsistent results.")
+}
+
+func TestDBQueryDoesntSeeAppendsAfterCreation(t *testing.T) {
+	tmpdir, _ := ioutil.TempDir("", "test")
+	defer os.RemoveAll(tmpdir)
+
+	db, err := Open(tmpdir, nil, nil, nil)
+	testutil.Ok(t, err)
+	defer db.Close()
+
+	querierBeforeAdd, err := db.Querier(context.Background(), 0, 1000000)
+	testutil.Ok(t, err)
+	defer querierBeforeAdd.Close()
+
+	app := db.Appender()
+	_, err = app.Add(labels.FromStrings("foo", "bar"), 0, 0)
+	testutil.Ok(t, err)
+
+	querierAfterAddButBeforeCommit, err := db.Querier(context.Background(), 0, 1000000)
+	testutil.Ok(t, err)
+	defer querierAfterAddButBeforeCommit.Close()
+
+	// None of the queriers should return anything after the Add but before the commit.
+	ss, _, err := querierBeforeAdd.Select(nil, labels.MustNewMatcher(labels.MatchEqual, "foo", "bar"))
+	testutil.Ok(t, err)
+	_, seriesSet, err := expandSeriesSet(ss)
+	testutil.Ok(t, err)
+	testutil.Equals(t, map[string][]sample{}, seriesSet)
+
+	ss, _, err = querierAfterAddButBeforeCommit.Select(nil, labels.MustNewMatcher(labels.MatchEqual, "foo", "bar"))
+	testutil.Ok(t, err)
+	_, seriesSet, err = expandSeriesSet(ss)
+	testutil.Ok(t, err)
+	testutil.Equals(t, map[string][]sample{}, seriesSet)
+
+	// This commit is after the queriers are created, so should not be returned.
+	err = app.Commit()
+	testutil.Ok(t, err)
+
+	// Nothing returned for querier created before the Add.
+	ss, _, err = querierBeforeAdd.Select(nil, labels.MustNewMatcher(labels.MatchEqual, "foo", "bar"))
+	testutil.Ok(t, err)
+	_, seriesSet, err = expandSeriesSet(ss)
+	testutil.Ok(t, err)
+	testutil.Equals(t, map[string][]sample{}, seriesSet)
+
+	// Series exists but has no samples for querier created after Add.
+	ss, _, err = querierAfterAddButBeforeCommit.Select(nil, labels.MustNewMatcher(labels.MatchEqual, "foo", "bar"))
+	testutil.Ok(t, err)
+	_, seriesSet, err = expandSeriesSet(ss)
+	testutil.Ok(t, err)
+	testutil.Equals(t, map[string][]sample{`{foo="bar"}`: []sample{}}, seriesSet)
+
+	querierAfterCommit, err := db.Querier(context.Background(), 0, 1000000)
+	testutil.Ok(t, err)
+	defer querierAfterCommit.Close()
+
+	// Samples are returned for querier created after Commit.
+	ss, _, err = querierAfterCommit.Select(nil, labels.MustNewMatcher(labels.MatchEqual, "foo", "bar"))
+	testutil.Ok(t, err)
+	_, seriesSet, err = expandSeriesSet(ss)
+	testutil.Ok(t, err)
+	testutil.Equals(t, map[string][]sample{`{foo="bar"}`: []sample{{t: 0, v: 0}}}, seriesSet)
 }
 
 // TestChunkWriter_ReadAfterWrite ensures that chunk segment are cut at the set segment size and
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -81,18 +81,18 @@ type Head struct {
 
 	symMtx  sync.RWMutex
 	symbols map[string]struct{}
-	values  map[string]stringset // label names to possible values
+	values  map[string]stringset // Label names to possible values.
 
 	deletedMtx sync.Mutex
 	deleted    map[uint64]int // Deleted series, and what WAL segment they must be kept until.
 
-	postings *index.MemPostings // postings lists for terms
+	postings *index.MemPostings // Postings lists for terms.
 
 	tombstones *tombstones.MemTombstones
 
 	cardinalityMutex      sync.Mutex
-	cardinalityCache      *index.PostingsStats // posting stats cache which will expire after 30sec
-	lastPostingsStatsCall time.Duration        // last posting stats call (PostingsCardinalityStats()) time for caching
+	cardinalityCache      *index.PostingsStats // Posting stats cache which will expire after 30sec.
+	lastPostingsStatsCall time.Duration        // Last posting stats call (PostingsCardinalityStats()) time for caching.
 }
 
 type headMetrics struct {
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -90,6 +90,8 @@ type Head struct {
 
 	tombstones *tombstones.MemTombstones
 
+	iso *isolation
+
 	cardinalityMutex      sync.Mutex
 	cardinalityCache      *index.PostingsStats // Posting stats cache which will expire after 30sec.
 	lastPostingsStatsCall time.Duration        // Last posting stats call (PostingsCardinalityStats()) time for caching.
@@ -105,8 +107,6 @@ type headMetrics struct {
 	chunksCreated           prometheus.Counter
 	chunksRemoved           prometheus.Counter
 	gcDuration              prometheus.Summary
-	minTime                 prometheus.GaugeFunc
-	maxTime                 prometheus.GaugeFunc
 	samplesAppended         prometheus.Counter
 	walTruncateDuration     prometheus.Summary
 	walCorruptionsTotal     prometheus.Counter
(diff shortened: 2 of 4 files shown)


beorn7:
Fixes #1893.

tsdb/isolation.go
```diff
@@ -0,0 +1,220 @@
...
+	return isoState
+}
+
+// newWriteID increments the transaction counter and returns a new transaction ID.
+func (i *isolation) newWriteID() uint64 {
+	i.writeMtx.Lock()
+	i.lastWriteID++
+	writeID := i.lastWriteID
+	i.writesOpen[writeID] = struct{}{}
+	i.writeMtx.Unlock()
```
brian-brazil: We can use defer now without a performance penalty.

tsdb/db_test.go
```diff
@@ -2470,6 +2474,133 @@ func TestDBReadOnly_FlushWAL(t *testing.T) {
...
+	defer querier.Close()
+
+	ss, err = querier.Select(labels.MustNewMatcher(labels.MatchEqual, "foo", "bar"))
+	testutil.Ok(t, err)
+
+	seriesSet = readSeriesSet(t, ss)
+	testutil.Equals(t, seriesSet, map[string][]sample{`{foo="bar"}`: []sample{{t: 0, v: 0}}})
+}
+
+func readSeriesSet(t *testing.T, ss SeriesSet) map[string][]sample {
```
brian-brazil: We have expandSeriesSet these days.

tsdb/head.go
```diff
@@ -1558,6 +1603,7 @@ func (s *stripeSeries) gc(mint int64) (map[uint64]struct{}, int) {
 			for _, series := range all {
 				series.Lock()
 				rmChunks += series.truncateChunksBefore(mint)
+				series.cleanupExtraWriteIds()
```
brian-brazil: This doesn't feel right. Shouldn't we block GC of a series if it'd break isolation?

beorn7:
Note: Recently merged #6777 didn't trigger any formal merge conflicts, but it still breaks the tests in here. I'll rebase at my next convenience. This should not touch anything else but tests, as far as I can see, so please go ahead with the review.

tsdb/isolation.go
```diff
@@ -0,0 +1,220 @@
...
+		if pos == len(txr.txIDs) {
+			pos = 0
+		}
+	}
+
+	txr.txIDFirst %= len(txr.txIDs)
+}
+
+// cutoffN will only keep the latest N transactions in the ring.
+// Will also downsize the ring if possible.
```
brian-brazil: Are we sure we want to downsize? My thinking previously was that if it got that big before, it'll get that big again.

tsdb/head.go
```diff
@@ -1832,23 +1897,79 @@ func (s *memSeries) iterator(id int, it chunkenc.Iterator) chunkenc.Iterator {
...
+		// Removing the extra transactionIDs that are relevant for samples that
+		// come after this chunk, from the total transactionIDs.
+		writeIDsToConsider := s.txs.txIDCount - (totalSamples - (previousSamples + numSamples))
+
+		// Iterate over the writeIDs, find the first one that the isolation state says not
+		// to return.
+		it := s.txs.iterator()
+		for index := 0; index < writeIDsToConsider; index++ {
+			writeID := it.At()
+			if _, ok := isoState.incompleteWrites[writeID]; ok || writeID > isoState.maxWriteID {
```
brian-brazil: It may be faster to do this comparison the other way around.

tsdb/isolation.go
beorn7: Indeed. Done.

tsdb/db_test.go
beorn7: But `expandSeriesSet` only returns the labels, not the values.

tsdb/head.go
beorn7: @gouthamve did you have any thoughts about this?

tsdb/isolation.go
beorn7: Again, perhaps @gouthamve has context.

beorn7: Rough thoughts from my side: If it is changing at all, that probably means there is a varying ingestion load. In steady state, it should not change that much. Which means the downsizing will help to downsize after an irregular ingestion peak, which is good. It will also downsize if ingestion stalls. Which is not desirable, but not really that harmful, either. Being able to shrink again after a spike seems to be worth the price.
(In practice, it's probably pretty irrelevant anyway.)

tsdb/head.go
beorn7: Indeed. Refactoring…

roidelapluie:
Just dropping a note that I would like, once this is ready for merge, to prombench this against master.

brian-brazil:
We should prombench as soon as it compiles. I wonder how my memory estimates from 3 years ago hold up.

roidelapluie:
Is this documentations worthy ? in tsdb/docs ?

tsdb/db_test.go
beorn7: I could extend `expandSeriesSet` to also return values. WDYT?

roidelapluie:
> /benchmark master

/prombench master

tsdb/db_test.go
brian-brazil: If that's sane.

beorn7:
Something locks up, and then the instances OOM.
Debugging now.
@gouthamve your eyes on this will be much appreciated.

roidelapluie:
This pull request has been worked on by multiple people in the last years. That alone shows how this feature is needed and also the complexity of this issue.

I want to drop a note here that as this is a important and risky change, as release shepherd, this pull request is under my watch and if we do not settle on it (merge it) before the 5th of March, I will put my veto on it for the 2.17 release.

That said, we are still 2 weeks away from this date and the prometheus and tsdb maintainers can merge it in the time between without my intervention.

krasi-georgiev:
@beorn7 the prombench has loki logs, so maybe that will help with the debuging
I had a quick look and can't find the logs though, maybe @geekodour can point us to where the logs are for this crashing prometheus

geekodour:
@krasi-georgiev [these logs?](http://prombench.prometheus.io/grafana/explore?orgId=1&left=%5B%221582096421027%22,%221582130750242%22,%22loki-meta%22,%7B%22expr%22:%22%7Bprometheus%3D%5C%22test-pr-6841%5C%22%7D%22%7D,%7B%22mode%22:%22Logs%22%7D,%7B%22ui%22:%5Btrue,true,true,%22none%22%5D%7D%5D)

beorn7:
I had a look at the logs, that's where my conclusions are coming from. My next goal is to reproduce the crash locally so that I don't have to run prombench for eight hours. So far no luck.
It would be great if @gouthamve could chime in as he has probably the most intimate understanding how this code works.

beorn7:
@roidelapluie yes sure. It's your call when the time has come. Currently, we don't even know if this will work at all.

beorn7:
Looking at [this dashboard](http://prombench.prometheus.io/grafana/explore?orgId=1&left=%5B%221582065862685%22,%221582105648252%22,%22prometheus-meta%22,%7B%22expr%22:%22prometheus_tsdb_head_chunks%7Bjob%3D%5C%22prometheus%5C%22,namespace%3D%5C%22prombench-6841%5C%22,prometheus%3D~%5C%22(test%2B).*%5C%22%7D%22,%22format%22:%22time_series%22,%22intervalFactor%22:2,%22legendFormat%22:null,%22metric%22:%22prometheus_local_storage_memory_chunks%22,%22step%22:null,%22context%22:%22explore%22%7D,%7B%22mode%22:%22Metrics%22%7D,%7B%22ui%22:%5Btrue,true,true,%22none%22%5D%7D%5D&right=%5B%221582065862685%22,%221582105648252%22,%22prometheus-meta%22,%7B%22expr%22:%22prometheus_tsdb_isolation_high_watermark%7Bjob%3D%5C%22prometheus%5C%22,namespace%3D%5C%22prombench-6841%5C%22,prometheus%3D~%5C%22(test%2B).*%5C%22%7D%22,%22format%22:%22time_series%22,%22intervalFactor%22:2,%22legendFormat%22:null,%22metric%22:%22prometheus_local_storage_memory_chunks%22,%22step%22:null,%22context%22:%22explore%22%7D,%7B%22expr%22:%22prometheus_tsdb_isolation_low_watermark%7Bjob%3D%5C%22prometheus%5C%22,namespace%3D%5C%22prombench-6841%5C%22,prometheus%3D~%5C%22(test%2B).*%5C%22%7D%22,%22context%22:%22explore%22%7D,%7B%22mode%22:%22Metrics%22%7D,%7B%22ui%22:%5Btrue,true,true,%22none%22%5D%7D%5D), the hypothesis is the following:

For some reason, the low watermark got stuck around 02:45 UTC. First crash happened a while after the head truncation at 05:00, which apparently ran into trouble (head chunks didn't drop). Things didn't look to bad after that, but then the server went into a tight crash loop.

I'll focus my investigation on possible reasons why the low watermark didn't get updated. (Wild guess: A scrape got canceled halfway through.)

Still, hints from people who know this better than me highly appreciated.

beorn7:
I removed the downsizing code. Prombench has run for ~20 hours now without issues. RAM usage is increased a bit, but not more than before, when the downsizing code was included.

Maybe the downsizing code has a bug, which is plausible, but not proven yet.
Or we were just lucky in the current Prombench run.

Today I'm busy with other things, so I'll let Prombench run for a few days.

tsdb/head.go
```diff
@@ -1833,23 +1898,80 @@ func (s *memSeries) iterator(id int, it chunkenc.Iterator) chunkenc.Iterator {
...
+
 	if id-s.firstChunkID < len(s.chunks)-1 {
-		return c.chunk.Iterator(it)
+		// TODO(beorn7): Reviewers, please double-check this section extra carefully.
+		if msIter, ok := it.(*memSafeIterator); ok {
+			msIter.Iterator = c.chunk.Iterator(msIter.Iterator)
+			msIter.i = -1
+			msIter.total = numSamples
+			msIter.stopAfter = stopAfter
+			msIter.bufferedSamples = 0
```
bwplotka: No buf needed?

bwplotka: I assume, yes.

tsdb/head_test.go
```diff
@@ -1322,3 +1332,128 @@ func TestAddDuplicateLabelName(t *testing.T) {
...
+	testutil.Ok(t, app.Commit())
+	testutil.Equals(t, 999, lastValue(998))
+	testutil.Equals(t, 999, lastValue(999))
+	testutil.Equals(t, 1000, lastValue(1000))
+	testutil.Equals(t, 1001, lastValue(1001))
+	testutil.Equals(t, 1002, lastValue(1002))
+	testutil.Equals(t, 1002, lastValue(1003))
+}
+
+// func TestHead_Truncate_WriteIDs(t *testing.T) {
```
bwplotka: ?

brian-brazil:
> Do we know how much of overhead this simple solution would give? Would it really be too slow? Do we have data?

https://promcon.io/2017-munich/talks/staleness-in-prometheus-2-0/

bwplotka:
Thanks @brian-brazil ! Some homework for me to do :heart:

beorn7:
Context for @bwplotka:
- First of all, thanks for the review, will go through it and address what I can.
- Certainly the commented code is not here to stay. It's just an experiment to see if (a) that code contains a bug and (b) it is needed at all. I would like to get some input from the master minds behind this to see if the removal makes sense.
- Speaking of which: the master minds behind this are @brian-brazil (who came up with the idea in the first place) and then @gouthamve (who turned this into a university project). I simply resurrected an ancient stale PR in the archived tsdb repo and ported it to the current world order.
- To find out about the overhead, I'm running prombench.

beorn7:
Preliminary results:

- Just to be sure this is all worth it, I have demonstrated the practical relevance of ingestion atomicity on a regular production Prometheus server.
- I have also created a demo setup where the problem can be shown immediately.
- The actual good news: If I compile Prometheus with this PR, my demo setup shows that the isolation works in practice!
- Prombench has now been running for almost 4d without any problem, suggesting that the bug the first Prombench run has detected is really in the downsizing code I have now removed (see also @brian-brazil 's suspicion in his very early comment above).
- 1% increase in CPU usage.
- Avg query latencies slightly increased: 4% for query, 7% for query_range.
- Avg. go_memstats_alloc_bytes are 9% increased.
- Even with the clean-up code in place, there was some noticeable memory increase (for as long as the server didn't hit the bug).

@brian-brazil & @gouthamve are those increased resource usages within your expectations? Close to 10% mem increase is not nice, but neither prohibitive. Would be cool, of course, if we could improve that. As far as I can tell, the cleanup didn't really help, so I'd be inclined to simply remove it and get rid of the (as yet not well understood) bug.

beorn7:
@brian-brazil already answered the question back in 2017: In the state of this change back then, he saw +5% CPU and +11% RAM and said that was what he expected. So this is even better than what we had back then.

brian-brazil:
>  (see also @brian-brazil 's suspicion in his very early comment above).

To be fair my question was more if it made sense performance wise, nothing about correctness.

> are those increased resource usages within your expectations?

CPU usage is actually a bit lower than my numbers from a few years ago, memory is as expected. If anything memory should be worse now in % terms as so much has been optimised since then.

beorn7:
More nuggets from @brian-brazil 's slides:

"Will need to be able to work across Queriers so all selectors in a PromQL
expression have the same view of the data. Thus isolation needs to be exposed in
the tsdb API." → I think this has been accomplished in the current PR without being exposed in the TSDB API as everything goes through the one head now. (I think that was @gouthamve 's doing in his PR, from which I've created this PR.)

"Currently scrape data and up & friends are written via separate Appenders. They
should use a single Appender so data is properly isolated." → Brian called that "tricky" in his talk, and AFAICS we still use a separate appender for the reporting. So that's still a TODO, I presume.

tsdb/head.go
beorn7: I will remove the cleanup of extra write IDs as it doesn't seem to be necessary and also seems to be buggy.

tsdb/isolation.go
beorn7: Ass said above, will remove that as it doesn't seem to make a dent and appears to be buggy.

tsdb/db_test.go
```diff
@@ -1276,19 +1280,29 @@ func TestNotMatcherSelectsLabelsUnsetSeries(t *testing.T) {
...
+		for _, ll := range c.series {
+			_, ok := ess[ll.String()]
+			testutil.Assert(t, ok, "expected series %s not found in result", ll.String())
+		}
 	}
 }
 
-func expandSeriesSet(ss storage.SeriesSet) ([]labels.Labels, error) {
-	result := []labels.Labels{}
+func expandSeriesSet(ss storage.SeriesSet) (map[string][]sample, error) {
```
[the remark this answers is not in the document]
beorn7: That's true, but it would require a separate data structure. I'll change it to return the raw labels (not stringified) separately.

roidelapluie:
On 24 Feb 11:59, Björn Rabenstein wrote:
> "Currently scrape data and up & friends are written via separate Appenders. They
> should use a single Appender so data is properly isolated." → Brian called that "tricky" in his talk, and AFAICS we still use a separate appender for the reporting. So that's still a TODO, I presume.
> 

I guess one of the reasons it is tricky is that if Commit() or Add() fails for
the scraped metrics, then up() should still be Added and Commited (but with a 0
value)

tsdb/db_test.go
```diff
@@ -2477,6 +2491,111 @@ func TestDBReadOnly_FlushWAL(t *testing.T) {
...
+			}
+		}
+	}()
+
+	<-firstInsert
+
+	// This is a race condition, so do a few tests to tickle it.
+	// Usually most will fail.
+	inconsistencies := 0
+	for i := 0; i < 10; i++ {
```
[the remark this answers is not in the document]
beorn7: Yes…

tsdb/head.go
```diff
@@ -81,18 +81,20 @@ type Head struct {
 
 	symMtx  sync.RWMutex
 	symbols map[string]struct{}
-	values  map[string]stringset // label names to possible values
+	values  map[string]stringset // Label names to possible values.
```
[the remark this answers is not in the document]
beorn7: Couldn't resist…

tsdb/head.go
```diff
@@ -209,6 +213,20 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 		Name: "prometheus_tsdb_checkpoint_creations_total",
 		Help: "Total number of checkpoint creations attempted.",
 	})
+	m.lowWatermark = prometheus.NewGaugeFunc(prometheus.GaugeOpts{
+		Name: "prometheus_tsdb_isolation_low_watermark",
+		Help: "The lowest write ID that is still referenced.",
```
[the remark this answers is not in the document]
beorn7: Just scrape. I'll update the help string.

tsdb/head.go
```diff
@@ -1559,6 +1604,7 @@ func (s *stripeSeries) gc(mint int64) (map[uint64]struct{}, int) {
 			for _, series := range all {
 				series.Lock()
 				rmChunks += series.truncateChunksBefore(mint)
+				// series.cleanupExtraWriteIds()
```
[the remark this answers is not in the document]
beorn7: See separate comment. This will be removed.

tsdb/head.go
```diff
@@ -1773,7 +1822,7 @@ func (s *memSeries) truncateChunksBefore(mint int64) (removed int) {
 }
 
 // append adds the sample (t, v) to the series.
```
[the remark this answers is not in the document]
beorn7: Amended.

tsdb/head.go
```diff
@@ -840,24 +863,33 @@ func (a *initAppender) Rollback() error {
 func (h *Head) Appender() storage.Appender {
 	h.metrics.activeAppenders.Inc()
 
+	writeID := h.iso.newWriteID()
```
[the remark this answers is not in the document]
beorn7: While the writeID is created during appender creation, I think it would be confusing to call it appenderID. This is propagated everywhere, including chunk iterators. When reading from the TSDB, you probably don't care if writes are done by appenders or batchers or scrapers.

writeID is also the term used in Brian's talk. I think the terminology makes sense.

tsdb/head.go
```diff
@@ -1810,9 +1859,25 @@ func (s *memSeries) append(t int64, v float64) (success, chunkCreated bool) {
...
 	return true, chunkCreated
 }
 
+// cleanupWriteIDsBelow cleans up older writeIds. Has to be called after acquiring
+// lock.
+func (s *memSeries) cleanupWriteIDsBelow(bound uint64) {
+	s.txs.cleanupWriteIDsBelow(bound)
+}
+
+// func (s *memSeries) cleanupExtraWriteIds() {
```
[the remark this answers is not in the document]
beorn7: See separate comment. This will be removed.

tsdb/head_test.go
beorn7: See separate comment. This will be removed.

tsdb/head.go
beorn7: No buffered samples at that moment, if I understand that correctly.

beorn7:
I have rebased on top of master as there were merge conflicts.

beorn7:
I think the issue with the same appender for the actual samples and the synthetic reporting time series should be addressed separately from this PR. (Even without it, the isolation is already of tremendous value.)

Let's focus on what is left to be done here.

roidelapluie:
On 24 Feb 12:43, Björn Rabenstein wrote:
> I think the issue with the same appender for the actual samples and the synthetic reporting time series should be addressed separately from this PR. (Even without it, the isolation is already of tremendous value.)
> 
> Let's focus on what is left to be done here.

Yes, I agree with this, I was just documenting.

roidelapluie:
Should we have a cli flag to disable isolation? That would be an advanced user flag that can act as a workaround in case we have blocking bugs in this PR or we face other issues with isolation.

brian-brazil:
No, this is something that should always be on as you get incorrect results otherwise.

roidelapluie:
@brian-brazil I have a rule group that takes 2 minutes to run, from the first rule to the last one. I should check if all my queries will be 2 minutes back in time or if we only take the appender when all the rules are executed.

tsdb/head.go
brian-brazil: Yeah, we only need to do this for the last chunk as it's the only one that has buffered samples.

tsdb/head.go
```diff
@@ -1866,23 +1923,80 @@ func (s *memSeries) iterator(id int, it chunkenc.Iterator) chunkenc.Iterator {
...
 	}
 	// Serve the last 4 samples for the last chunk from the sample buffer
 	// as their compressed bytes may be mutated by added samples.
 	if msIter, ok := it.(*memSafeIterator); ok {
 		msIter.Iterator = c.chunk.Iterator(msIter.Iterator)
 		msIter.i = -1
-		msIter.total = c.chunk.NumSamples()
+		msIter.total = numSamples
+		msIter.stopAfter = stopAfter
+		msIter.bufferedSamples = 4
```
brian-brazil: What if there have been fewer than 4 samples appended so far? I think the stopAfter covers it, but we should add a test if there's not already one.

tsdb/head.go
beorn7: Thanks everyone.

tsdb/head.go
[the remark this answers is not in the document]
beorn7: As per discussion on IRC, how about `appendID`? It would convey that it is per appender, while it would still somewhat keep the symmetry between _read_ and _append_ (formerly _write_).

roidelapluie:
I did not get to the bottom of this PR, but it looks good.

Does commit() guarantee that upon return the transaction is visible to all readers? or could another parallel transaction prevent that?

See pic:

iso


My question is: is that possible that it breaks the rules from a sale rule group from seeing each other because of parallel transaction 'blocking' the watermark?

tsdb/db_test.go
```diff
@@ -2477,6 +2490,111 @@ func TestDBReadOnly_FlushWAL(t *testing.T) {
...
+	testutil.Ok(t, err)
+	defer db.Close()
+
+	querier, err := db.Querier(context.Background(), 0, 1000000)
+	testutil.Ok(t, err)
+	defer querier.Close()
+
+	app := db.Appender()
+	_, err = app.Add(labels.FromStrings("foo", "bar"), 0, 0)
+	testutil.Ok(t, err)
```
roidelapluie: Maybe we want a test with the querier created between the Add and the Commit, WDYT?

brian-brazil:
> My question is: is that possible that it breaks the rules from a same rule group from seeing each other because of parallel transaction 'blocking' the watermark?

If you've two different sources writing to the same time series you're already going to have big problems.

roidelapluie:
> > My question is: is that possible that it breaks the rules from a same rule group from seeing each other because of parallel transaction 'blocking' the watermark?
> 
> If you've two different sources writing to the same time series you're already going to have big problems.

I do not say that it is the same time series. In my example, 'long' might just be the commit phase of a very long scrape. In that case, R2 should not see the scrape. But if R1 has started after the beginning of 'long', will R2 see R1?

brian-brazil:
> I do not say that it is the same time series.

That's the only case where it'd make sense. This all works on a per time series basis.

roidelapluie:
> > I do not say that it is the same time series.
> 
> That's the only case where it'd make sense. This all works on a per time series basis.

Thanks, I see it now.

beorn7:
Thanks everyone for the review and the discussion. I'll try to address the suggestions from both of you for tests before merging.

roidelapluie:
Maybe you also want to add co-authored-by with the original authors, too, in the commit messages.

roidelapluie:
I notice that the Rollback function is not touched. Is there anything to do about the writeId here? like series.cleanupWriteIDsBelow(a.cleanupWriteIDsBelow)  or a.head.iso.closeWrite(a.writeID) ?

tsdb/head.go
```diff
@@ -209,6 +213,20 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 		Name: "prometheus_tsdb_checkpoint_creations_total",
 		Help: "Total number of checkpoint creations attempted.",
 	})
+	m.lowWatermark = prometheus.NewGaugeFunc(prometheus.GaugeOpts{
```
bwplotka: Silly question, but do we need to keep those variables? We could just `MustRegister(prometheus.NewGaugeFunc...., ...)`, right?

tsdb/head.go
```diff
@@ -314,7 +335,7 @@ func (h *Head) processWALSamples(
 				}
 				refSeries[s.Ref] = ms
 			}
-			_, chunkCreated := ms.append(s.T, s.V)
+			_, chunkCreated := ms.append(s.T, s.V, 0)
```
bwplotka: if _, chunkCreated := ...; chunkCreated {` pattern availabe (:

tsdb/head.go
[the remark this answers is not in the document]
bwplotka: :+1:

tsdb/head.go
bwplotka: :+1:

tsdb/isolation.go
```diff
@@ -0,0 +1,194 @@
...
+import (
+	"sync"
+)
+
+// IsolationState holds the isolation information.
+type IsolationState struct {
+	// We will ignore all writes above the max, or that are incomplete.
+	maxWriteID       uint64
+	incompleteWrites map[uint64]struct{}
+	lowWaterMark     uint64 // Lowest of incompleteWrites/maxWriteId.
```
bwplotka: nit: inconsistency: sometimes we do `lowWaterMark` sometimes `lowWatermark`. Same in comment here: `maxWriteId/maxWriteID`

tsdb/head.go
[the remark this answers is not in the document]
beorn7: Done.

tsdb/head.go
beorn7: The number was always 4 before this change. I'll still make sure this is covered in tests.

tsdb/head.go
beorn7: That's true. The functional metrics never need to be referenced. I'll register them directly and remove them from the struct.

beorn7: Unless we need them in tests (which is the case for `m.series`).

tsdb/isolation.go
beorn7: Fixed.

tsdb/db_test.go
beorn7: I modified the test, with an interesting result. Will comment below.

tsdb/head.go
beorn7: I have refactored the iterators. Will point you to it below for a re-review.

beorn7:
OK, I have addressed all comments, but now we are left with two tasks:

1. I have refactored (and hopefully improved) the iterators in head.go. Please check out https://github.com/prometheus/prometheus/pull/6841/commits/b8ab4062dfcb6576475cd3433e9717dc03ea97cb if it makes sense. (Tests are passing, but I'll run prombench on top of that to make sure it still works and performance isn't worse.)
2. I modified `TestDBQueryDoesntSeeAppendsAfterCreation` as @roidelapluie suggested, with the interesting result that a querier created after `Add` but before `Commit` sees the series, but without samples. This _might_ be OK, and perhaps even expected, but I'm not sure. Could you all have a look at the test in its current form and let me know what you think?

Thanks once more!

roidelapluie:
>     2. I modified `TestDBQueryDoesntSeeAppendsAfterCreation` as @roidelapluie suggested, with the interesting result that a querier created after `Add` but before `Commit` sees the series, but without samples. This _might_ be OK, and perhaps even expected, but I'm not sure. Could you all have a look at the test in its current form and let me know what you think?

This is expected.

IRC log:
> 17:02:33 < roidelapluie> this PR is only isolating samples not time series right
> 17:02:57 < roidelapluie> e.g. as soon as we run Add we can get the time series, even if no samples have been Committed
> 17:04:04 <@bbrazil> yeah, that's always been like that - and always will be

beorn7:
Still interesting that the previously created Querier doesn't even see the series.

beorn7:
OK, will now run prombench.

Once I have green light on the iterator optimization, I'll squash this appropriately and add the recommended links to talks and slides and to co-authors.

beorn7:
I should also rebase onto master before running prombench.

tsdb/isolation.go
```diff
@@ -0,0 +1,193 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tsdb
+
+import (
+	"sync"
+)
+
+// IsolationState holds the isolation information.
```
roidelapluie: Is there a point in making this public? That should probably be internal.

beorn7: I was asking myself the same question, but then thought that @brian-brazil and/or @gouthamve probably had a reason. And then I stopped thinking about it… (o: But you are right, I guess it should be unexported.

roidelapluie: This does not seem to be usable with public apis in any ways...

roidelapluie:
There is probably still a tiny memory leak as rollback does not call a.head.iso.closeAppend(a.appendID).

roidelapluie:
> There is probably still a tiny memory leak as rollback does not call a.head.iso.closeAppend(a.appendID).
> 
> EDIT: and cleanupAppendIDsBelow ?

I confirm this. This can be tested with a target that exposes 'a{a="a",a="a"} 1' ; in this case we roll back and the incompleteAppends is growing on each (unsuccessful) scrape.

roidelapluie:
Note that a rollback is what could have caused the low watermark not being updated in the previous prombench; so the downsize code is probably not guilty.

beorn7:
Thanks for the catch, @roidelapluie . I'll investigate ASAP.

tsdb/isolation.go
beorn7: Unexported IsolationState.

beorn7:
Soooo, I think I fixed the rollback. While writing the test for it, I realized another problem, see commit description.

Please, dear reviewers, have another look if that's sane.

roidelapluie:
Don't we need (*txRing) cleanupAppendIDsBelow when Rolling Back ?

roidelapluie:
I have another question: Should we have a test that tests the isolation, but with AddFast ?

beorn7:
Benchmark looks OK, but we have to benchmark the new changes anyway. So canceling the current run.

beorn7:
> Don't we need (*txRing) cleanupAppendIDsBelow when Rolling Back ?

We do. I have added it, including amending a test so that it now exposes the bug.

tsdb/head_test.go
```diff
@@ -1508,6 +1508,19 @@ func TestMemSeriesIsolation(t *testing.T) {
...
+	app = hb.appender(uint64(i), 1001)
+	_, err = app.Add(labels.FromStrings("foo", "bar"), int64(i), float64(i))
+	testutil.Ok(t, err)
+	testutil.Ok(t, app.Rollback())
+	testutil.Equals(t, 1000, lastValue(999))
+	testutil.Equals(t, 1000, lastValue(1000))
+	testutil.Equals(t, 1001, lastValue(1001))
+	testutil.Equals(t, 1002, lastValue(1002))
+	testutil.Equals(t, 1002, lastValue(1003))
+
```
roidelapluie: extra wl not needed

beorn7:
Prombench results:
- 3% increase in CPU usage.
- Avg query latencies increased 4% for query, 19% for query_range.
- Avg. go_memstats_alloc_bytes are 19% increased.

This is now much higher than before, but I believe due to the buggy "cleanup everything while no reads are in progress" we had before, we didn't really do the full story before.

beorn7:
I think those values are still within expectations, based on what @brian-brazil said [above](https://github.com/prometheus/prometheus/pull/6841#issuecomment-590511101).

This will still, I guess, be noticed painfully by users with high-load tight-resource setups.

beorn7:
Rebased and squashed.

Last call for objections, otherwise I'll merge in about an hour or so…

roidelapluie:
Hooray :tada::tada::tada::tada:

roidelapluie:
@beorn7 @brian-brazil can we agree on a sentence to put at the beginning of the release notes about this? Especially something about the memory increase.

beorn7:
I'm also running this now on a moderately loaded production server for comparison (500k series, 30k samples/s, very low query volume). The increase in RAM consumption is even more pronounced here: The peak value of `go_memstats_heap_alloc_bytes` jumps from 4.4GB to 7.2GB. `container_memory_working_set_bytes` goes from 8.2GB to 11.3GB.

I'll do a bit of heap analysis at my next convenience to see if there are low hanging fruit.

beorn7:
This should definitely come with a big warning in the release notes. It will anyway be a hard sell, given that only very few users will have ever noticed the problems with isolation. I'd also keep reverting this on the agenda. We need to give it some thought…

beorn7:
Peak `container_memory_working_set_bytes` (just before head truncation) is 11.7GB vs. 8.6GB, i.e. 36% increase.

roidelapluie:
That hurts

beorn7:
I think I have found the bug: When we replay the WAL, we dutifully append every sample with appendID 0. Those are all cleaned up after the next commit, but the ring buffer has then already grown to accommodate all samples in the WAL, and it will never shrink again.

The solution should be easy: appendID==0 should never be recorded. PR in preparation…

roidelapluie:
> I think I have found the bug: When we replay the WAL, we dutifully append every sample with appendID 0. Those are all cleaned up after the next commit, but the ring buffer has then already grown to accommodate all samples in the WAL, and it will never shrink again.
> 
> The solution should be easy: appendID==0 should never be recorded. PR in preparation…

I still think it might be worthy to have the downsizing code again

beorn7:
> I still think it might be worthy to have the downsizing code again

Let's see how it turns out in practice. Downsizing could easily let to oscillating behavior, which would be worse than now downsizing (spikier memory usage, more allocations, slower appends).

roidelapluie:
> > I still think it might be worthy to have the downsizing code again
> 
> Let's see how it turns out in practice. Downsizing could easily let to oscillating behavior, which would be worse than no downsizing (spikier memory usage, more allocations, slower appends).

Would it make sense to downsize at certain moments, e.g. head gc ?

bobrik:
I think this could fix #4580 as well.

PR #6845, opened 2020-02-19
Fix golint failures in pkg/labels/labels.go

sshukun:


sshukun:
Close this PR as the if is left by design.

PR #6852, opened 2020-02-20
Log scrape append failures as debug rather than warn.

brian-brazil:
This is most likely due to an endpoint not producing valid
metrics output, which we should treat the same as a failed
scrape, and thus not spam the application logs with it.


roidelapluie:
Will there still be errors in the logs by default in case of disk full?

brian-brazil:
You'll still get the "appending scrape report failed" log in that case.

PR #6831, opened 2020-02-17
RELEASE.md: add instructions for React dep updates

juliusv:
Signed-off-by: Julius Volz <julius.volz@gmail.com>

PR #6855, opened 2020-02-21
Update React vendoring

SuperQ:
`yarn upgrade` all deps.
* Fixe security warning for `serialize-javascript`.

### DIFF
--- a/web/ui/react-app/package.json
+++ b/web/ui/react-app/package.json
@@ -34,7 +34,7 @@
     "react-copy-to-clipboard": "^5.0.1",
     "react-dom": "^16.7.0",
     "react-resize-detector": "^4.2.1",
-    "react-scripts": "^3.2.0",
+    "react-scripts": "^3.4.0",
     "react-test-renderer": "^16.9.0",
     "reactstrap": "^8.0.1",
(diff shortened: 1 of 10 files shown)


juliusv:
This causes a bunch of failures in the React tests, so it seems it would require a bit of a deeper look into things.

Alternatively you could try just updating the one package (I think `yarn upgrade serialize-javascript`?).

SuperQ:
Yea, I tried just the minimal upgrade first, but it is's a burred deep dependency, with a major version change. I'll see if I can walk up the tree to see if I can make yarn upgrade just the bare minimum.

roidelapluie:
Does https://github.com/prometheus/prometheus/pull/6831 help?

SuperQ:
Ok, I think I figured out the minimum update chain to fix the security warning.

SuperQ:
Looks like eslint got stricter, will see if I can fix things.

SuperQ:
Looks like there's a couple of additional test failures.

SuperQ:
I fixed up one test fixture.

juliusv:
:+1: Thanks!

Issue #6467, opened 2019-12-16
`promtool check rules` should expand templates before inspecting rules


spads-spads:
## Proposal
**Use case. Why is this important?**

Since alertmanager supports golang templating, it should not parse the YAML until it has at least tried to render the templates into the target rules file.  Currently any inline template definition or use that involves newlines or indentation changes will generate an error.

## Bug Report
**What did you do?**

I created a rules file with a `{{define "macro-name"}}` section, per the docs.

**What did you expect to see?**

Either a warning that the template could not be expanded without more context, or a successfully linted rules file complete with rendered resulting checks.

**What did you see instead? Under which circumstances?**

```
  FAILED:
yaml: line 10: could not find expected ':'
```

**Environment**

* System information:

```
Linux 4.15.0-72-generic x86_64
```

* Prometheus version:

```
prometheus, version 2.1.0+ds (branch: debian/sid, revision: 2.1.0+ds-1)
  build user:       pkg-go-maintainers@lists.alioth.debian.org
  build date:       20180121-21:30:42
  go version:       go1.9.2
```

* Alertmanager rules file:
```
groups:
- name: test
  rules:
  - alert: Test Alert
    expr:  time() > 0
    annotations:
      description: This is a test alert that should always fire
    labels:
      severity: warning
{{define "test-macro" (args 1)}}
    expr:  time() > {{arg0}}
{{end}}
```

brian-brazil:
I thought we'd added this to the general reload a good while back, if not we should do so and ensure that promtool is also running that code.

spads-spads:
Do you know where the code to do this lives?  I'd be happy to build a custom version of promtool for my own linting purposes in the short term.

brian-brazil:
Not offhand, but see where `template` is referenced.

To be clear we can only check that a template parses, not that it'll subsequently evaluate.

roidelapluie:
Your exemple, a *prometheus* rule file, is invalid. The error message is correct, this file is not yaml valid.

We support go template only in two locations: label value and annotation value ( https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/ look for tmpl_string  )

My experience tells me that those templates are correctly checked during validation.


Here is what you could do:

```
groups:
- name: test
  rules:
  - alert: Test Alert
    expr:  time() > 0
    annotations:
      description: '{{"go template"}}'
    labels:
      severity: '{{"go template"}}'
```

[...]

PR #6856, opened 2020-02-21
Adds support for expanding templates before rules check

Harkishen-Singh:
Fixes: #6467 

Explanatory gif:

expd

Output before the patch:
```
Checking test.yml
  FAILED:
../../test.yml: yaml: line 11: could not find expected ':'
```

**test.yml**
```
groups:
    - name: test
      rules:
      - alert: Test Alert
        expr:  up == 0
        annotations:
          description: This is a test alert that should always fire
        labels:
          severity: warning
        {{define "test-macro"}}
            expr:  time() > {{.arg0}}
        {{end}}
```

### DIFF
--- a/pkg/rulefmt/rulefmt.go
+++ b/pkg/rulefmt/rulefmt.go
@@ -15,8 +15,8 @@ package rulefmt
 
 import (
 	"context"
-	"io/ioutil"
 	"strings"
+	templ "text/template"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


juliusv:
@Harkishen-Singh I think there was some misunderstanding unfortunately - this looks like it's adding templating functionality to the top-level file, before YAML is even parsed. The creator of the linked issue https://github.com/prometheus/prometheus/issues/6467 apparently assumed that this was functionality that was supposed to work and that we want to have, but so far we've decided pretty explicitly against having general templating functionality in the top-level of the YAML. And for the specific fields where templating is supported, we already check the template parsing when the file is loaded. So I believe there is nothing to actually do there, as indicated by https://github.com/prometheus/prometheus/issues/6467#issuecomment-589329945.

brian-brazil:
Ah indeed, this is not something that should work.

Harkishen-Singh:
@juliusv @brian-brazil I get your point. Thank you for explaining the issue. Can we close the issue if it points to something we do not need?

roidelapluie:
I have closed it. Thanks however to look into it. Sorry you did not see my comment in time.

PR #6860, opened 2020-02-23
API_V1: Extract time param parsing logic

boyskila:
Moving all the changes related to time param parsing from this [PR](https://github.com/prometheus/prometheus/pull/6837) into its own since different parts of the API, not related to what the PR should do are affected.

@juliusv 
@krasi-georgiev Please review

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -317,7 +317,7 @@ func (api *API) options(r *http.Request) apiFuncResult {
 }
 
 func (api *API) query(r *http.Request) apiFuncResult {
-	_, ts, err := parseTimeParam(r, "time", api.now())
+	ts, err := parseTimeParam(r, "time", api.now())
 	if err != nil {
 		return apiFuncResult{nil, &apiError{errorBadData, err}, nil, nil}
 	}
@@ -364,12 +364,14 @@ func (api *API) query(r *http.Request) apiFuncResult {
 }
 
 func (api *API) queryRange(r *http.Request) apiFuncResult {
-	_, start, err := parseTimeParam(r, "start", minTime)
+	start, err := parseTime(r.FormValue("start"))
 	if err != nil {
+		err = errors.Wrapf(err, "invalid parameter 'start'")
 		return apiFuncResult{nil, &apiError{errorBadData, err}, nil, nil}
 	}
-	_, end, err := parseTimeParam(r, "end", maxTime)
+	end, err := parseTime(r.FormValue("end"))
 	if err != nil {
+		err = errors.Wrapf(err, "invalid parameter 'end'")
 		return apiFuncResult{nil, &apiError{errorBadData, err}, nil, nil}
 	}
 	if end.Before(start) {
@@ -507,11 +509,11 @@ func (api *API) series(r *http.Request) apiFuncResult {
 		return apiFuncResult{nil, &apiError{errorBadData, errors.New("no match[] parameter provided")}, nil, nil}
 	}
 
-	start, _, err := parseTimeParam(r, "start", minTime)
+	start, err := parseTimeParam(r, "start", minTime)
 	if err != nil {
 		return apiFuncResult{nil, &apiError{errorBadData, err}, nil, nil}
 	}
-	end, _, err := parseTimeParam(r, "end", maxTime)
+	end, err := parseTimeParam(r, "end", maxTime)
 	if err != nil {
 		return apiFuncResult{nil, &apiError{errorBadData, err}, nil, nil}
 	}
@@ -525,7 +527,7 @@ func (api *API) series(r *http.Request) apiFuncResult {
 		matcherSets = append(matcherSets, matchers)
 	}
 
-	q, err := api.Queryable.Querier(r.Context(), start, end)
+	q, err := api.Queryable.Querier(r.Context(), timestamp.FromTime(start), timestamp.FromTime(end))
 	if err != nil {
 		return apiFuncResult{nil, &apiError{errorExec, err}, nil, nil}
 	}
@@ -1302,11 +1304,11 @@ func (api *API) deleteSeries(r *http.Request) apiFuncResult {
 		return apiFuncResult{nil, &apiError{errorBadData, errors.New("no match[] parameter provided")}, nil, nil}
 	}
 
-	start, _, err := parseTimeParam(r, "start", minTime)
+	start, err := parseTimeParam(r, "start", minTime)
 	if err != nil {
 		return apiFuncResult{nil, &apiError{errorBadData, err}, nil, nil}
 	}
-	end, _, err := parseTimeParam(r, "end", maxTime)
+	end, err := parseTimeParam(r, "end", maxTime)
 	if err != nil {
 		return apiFuncResult{nil, &apiError{errorBadData, err}, nil, nil}
 	}
@@ -1317,7 +1319,7 @@ func (api *API) deleteSeries(r *http.Request) apiFuncResult {
 			return apiFuncResult{nil, &apiError{errorBadData, err}, nil, nil}
 		}
 
-		if err := db.Delete(start, end, matchers...); err != nil {
+		if err := db.Delete(timestamp.FromTime(start), timestamp.FromTime(end), matchers...); err != nil {
 			return apiFuncResult{nil, &apiError{errorInternal, err}, nil, nil}
 		}
 	}
@@ -1443,16 +1445,16 @@ func (api *API) respondError(w http.ResponseWriter, apiErr *apiError, data inter
 	}
 }
 
-func parseTimeParam(r *http.Request, paramName string, defaultValue time.Time) (int64, time.Time, error) {
+func parseTimeParam(r *http.Request, paramName string, defaultValue time.Time) (time.Time, error) {
 	val := r.FormValue(paramName)
-	if val != "" {
-		result, err := parseTime(val)
-		if err != nil {
-			return -1, time.Time{}, errors.Wrapf(err, "Invalid parameters '%s'", paramName)
-		}
-		return timestamp.FromTime(result), result, nil
+	if val == "" {
+		return defaultValue, nil
 	}
-	return timestamp.FromTime(defaultValue), defaultValue, nil
+	result, err := parseTime(val)
+	if err != nil {
+		return time.Time{}, errors.Wrapf(err, "Invalid parameters '%s'", paramName)
+	}
+	return result, nil
 }
 
 func parseTime(s string) (time.Time, error) {
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -1452,7 +1452,7 @@ func parseTimeParam(r *http.Request, paramName string, defaultValue time.Time) (
 	}
 	result, err := parseTime(val)
 	if err != nil {
-		return time.Time{}, errors.Wrapf(err, "Invalid parameters '%s'", paramName)
+		return time.Time{}, errors.Wrapf(err, "Invalid time value for '%s'", paramName)
 	}
 	return result, nil
 }
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -317,18 +317,10 @@ func (api *API) options(r *http.Request) apiFuncResult {
 }
 
 func (api *API) query(r *http.Request) apiFuncResult {
-	var ts time.Time
-	if t := r.FormValue("time"); t != "" {
-		var err error
-		ts, err = parseTime(t)
-		if err != nil {
-			err = errors.Wrapf(err, "invalid parameter 'time'")
-			return apiFuncResult{nil, &apiError{errorBadData, err}, nil, nil}
-		}
-	} else {
-		ts = api.now()
+	_, ts, err := parseTimeParam(r, "time", api.now())
+	if err != nil {
+		return apiFuncResult{nil, &apiError{errorBadData, err}, nil, nil}
 	}
-
 	ctx := r.Context()
 	if to := r.FormValue("timeout"); to != "" {
 		var cancel context.CancelFunc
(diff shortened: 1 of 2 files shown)


web/api/v1/api.go
```diff
@@ -1479,6 +1443,18 @@ func (api *API) respondError(w http.ResponseWriter, apiErr *apiError, data inter
...
+func parseTimeParam(r *http.Request, paramName string, defaultValue time.Time) (int64, time.Time, error) {
+	val := r.FormValue(paramName)
+	if val != "" {
+		result, err := parseTime(val)
+		if err != nil {
+			return -1, time.Time{}, errors.Wrapf(err, "Invalid parameters '%s'", paramName)
+		}
+		return timestamp.FromTime(result), result, nil
+	}
+	return timestamp.FromTime(defaultValue), defaultValue, nil
```
roidelapluie: it looks like this would call timestamp.FromTime even when we don't need it.

web/api/v1/api.go
```diff
@@ -1479,6 +1443,18 @@ func (api *API) respondError(w http.ResponseWriter, apiErr *apiError, data inter
 	}
 }
 
+func parseTimeParam(r *http.Request, paramName string, defaultValue time.Time) (int64, time.Time, error) {
+	val := r.FormValue(paramName)
+	if val != "" {
```
roidelapluie: this can be better inlined by inverting the condition.

boyskila:
> Also tests fail.

Yep, due to ```Please ensure you are using go version go1.13.8 linux/amd64 for formatting code.```
I have to upgrade my go version. Locally tests itself are passing properly

web/api/v1/api.go
boyskila: Yes, thanks!

web/api/v1/api.go
boyskila: Yeah, I know....just I don't want to provide an additional flag for this. Any ideas of how this can be improved?
Btw this could be a good ```golang``` feature - lazy loading return values :)

roidelapluie: We only use timestamp.FromTime once, so you can still use it externally.

boyskila: Actually is used 4 times - in deleteSeries, in series and + 2 times more when is combined with this PR https://github.com/prometheus/prometheus/pull/6837

web/api/v1/api.go
```diff
@@ -372,14 +364,12 @@ func (api *API) query(r *http.Request) apiFuncResult {
 }
 
 func (api *API) queryRange(r *http.Request) apiFuncResult {
-	start, err := parseTime(r.FormValue("start"))
+	_, start, err := parseTimeParam(r, "start", minTime)
```
juliusv: I don't think we should be changing the behavior like this in a refactoring PR, without explicitly discussing / wanting it. Previously a missing start/end yielded an API error, now a default time is chosen. I'm not even sure if defaulting to minTime / maxTime is a good idea in general, because it will default to the most expensive query range (which could cause unintended overload for some people when they get API usage wrong).

juliusv: (for the `/api/v1/query_range` endpoint, that is)

web/api/v1/api.go
```diff
@@ -1479,6 +1443,18 @@ func (api *API) respondError(w http.ResponseWriter, apiErr *apiError, data inter
 	}
 }
 
+func parseTimeParam(r *http.Request, paramName string, defaultValue time.Time) (int64, time.Time, error) {
```
juliusv: I think I'd prefer keeping the return signature more "normalized" (less redundant) and simpler by just returning `time.Time` and letting the user do the `timestamp.FromTime()` whenever it is really needed.

krasi-georgiev: I commented for this in the original PR, but I don't mind either way.

web/api/v1/api.go
boyskila: I thought is a good idea, but yeah, it doesn't make sense to query min-max range (except user wants it). I'll revert this back. Thanks!

web/api/v1/api.go
boyskila: Yep, this matches with @roidelapluie comment https://github.com/prometheus/prometheus/pull/6860#discussion_r383118014

boyskila:
All tests now pass and requested changes are addressed. @krasi-georgiev  @juliusv @roidelapluie

boyskila:
Collecting more likes @juliusv  @roidelapluie :)

web/api/v1/api_test.go
```diff
@@ -41,6 +41,7 @@ import (
 	"github.com/prometheus/common/promlog"
 	"github.com/prometheus/common/route"
 
+	errPkg "github.com/pkg/errors"
```
roidelapluie: we import this 89 times in our code base and never need to aliase it. why is it needed here?

web/api/v1/api_test.go
```diff
@@ -2170,6 +2171,75 @@ func TestRespondError(t *testing.T) {
 		t.Fatalf("Expected response \n%v\n but got \n%v\n", res, exp)
 	}
 }
+func TestParseTimeParam(t *testing.T) {
```
roidelapluie: missing blank line

web/api/v1/api_test.go
```diff
@@ -2170,6 +2171,75 @@ func TestRespondError(t *testing.T) {
...
+	type resultType struct {
+		asTime  time.Time
+		asError func() error
+	}
+	ts, _ := parseTime("1582468023986")
```
roidelapluie: testOk ? not strictly needed but best practice.

web/api/v1/api_test.go
boyskila: Because of https://github.com/prometheus/prometheus/blob/2cc32c332d1ddf522ada0abe2691d908e91d5182/web/api/v1/api_test.go#L20

web/api/v1/api_test.go
boyskila: Not sure what do you mean but I guess is renaming ```ts``` to ```testOk```?

roidelapluie:
```suggestion
	ts, err := parseTime("1582468023986")
	testutil.Ok(t, err)
```

web/api/v1/api_test.go
roidelapluie: aren't they compatible? can't we just use pkg/errors?

boyskila: No without an alias.
```Running tool: /usr/bin/go test -timeout 30s github.com/prometheus/prometheus/web/api/v1 -run ^(TestEndpoints)$

# github.com/prometheus/prometheus/web/api/v1 [github.com/prometheus/prometheus/web/api/v1.test]
/home/boyskila/Desktop/prometheus/web/api/v1/api_test.go:44:2: errors redeclared as imported package name
	/home/boyskila/Desktop/prometheus/web/api/v1/previous declaration at ./api_test.go:20:2
/home/boyskila/Desktop/prometheus/web/api/v1/api_test.go:2213:13: undefined: errPkg
FAIL	github.com/prometheus/prometheus/web/api/v1 [build failed]
FAIL
Error: Tests failed.
```

roidelapluie:
```diff
diff --git a/web/api/v1/api_test.go b/web/api/v1/api_test.go
index 50b0191b5..0d141e817 100644
--- a/web/api/v1/api_test.go
+++ b/web/api/v1/api_test.go
@@ -17,7 +17,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -41,7 +40,7 @@ import (
 	"github.com/prometheus/common/promlog"
 	"github.com/prometheus/common/route"
 
-	errPkg "github.com/pkg/errors"
+	"github.com/pkg/errors"
 	"github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/pkg/gate"
 	"github.com/prometheus/prometheus/pkg/labels"
@@ -2209,7 +2208,7 @@ func TestParseTimeParam(t *testing.T) {
 				asTime: time.Time{},
 				asError: func() error {
 					_, err := parseTime("baz")
-					return errPkg.Wrapf(err, "Invalid time value for '%s'", "foo")
+					return errors.Wrapf(err, "Invalid time value for '%s'", "foo")
 				},
 			},
 		},
```

That works for me

juliusv: I think you can just replace the existing `errors` import with the new one, since we only use `errors.New()` from `errors`, and `github.com/pkg/errors` also has a `.New()` function.

juliusv:
:+1: besides the last nits

web/api/v1/api_test.go
boyskila: yes, @roidelapluie you are right. it's working. I wasn't completely sure since both packages differ but if you think is fine I'll replace it.

roidelapluie: > yes, @roidelapluie you are right. it's working. I wasn't completely sure since both packages differ but if you think is fine I'll replace it.

I think we do this in lots of places already, so I would reuse that pattern.

web/api/v1/api_test.go
```diff
@@ -2171,6 +2171,79 @@ func TestRespondError(t *testing.T) {
...
+				result.asError().Error(), err.Error(),
+			)
+		} else {
+			if !asTime.Equal(result.asTime) {
+				t.Errorf(`
+					Time as return value: %s not parsed correctly. Expected %s. Actual %s`,
+					test.paramValue, result.asTime, asTime,
+				)
+			}
+		}
```
roidelapluie:
```suggestion
		if err != nil {
		        testutil.ErrorEqual(t, result.asError().Error(), err.Error())
		} else {
		        testutil.Assert(t, asTime.Equal(result.asTime), 'time as return value: %s not parsed correctly. Expected %s. Actual %s', test.paramValue, result.asTime, asTime)
			}
		}
```

web/api/v1/api_test.go
```diff
@@ -2171,6 +2171,79 @@ func TestRespondError(t *testing.T) {
...
+					return errors.Wrapf(err, "Invalid time value for '%s'", "foo")
+				},
+			},
+		},
+	}
+	for _, test := range tests {
+		req, err := http.NewRequest("GET", "localhost:42/foo?"+test.paramName+"="+test.paramValue, nil)
+		if err != nil {
+			panic(err)
+		}
```
roidelapluie:
```suggestion
		testutil.Ok(t, err)
```

boyskila: Aren't those changes out of the scope of the PR?

web/api/v1/api_test.go
boyskila: Interesting, thanks :)

boyskila: Those ```testutil``` helpers are pretty useful.

web/api/v1/api_test.go
```diff
@@ -2171,6 +2171,69 @@ func TestRespondError(t *testing.T) {
...
+				asError: func() error {
+					_, err := parseTime("baz")
+					return errors.Wrapf(err, "Invalid time value for '%s'", "foo")
+				},
+			},
+		},
+	}
+	for _, test := range tests {
+		req, err := http.NewRequest("GET", "localhost:42/foo?"+test.paramName+"="+test.paramValue, nil)
+
```
roidelapluie: extra white line

web/api/v1/api_test.go
```diff
@@ -2171,6 +2171,69 @@ func TestRespondError(t *testing.T) {
...
+		result := test.result
+		asTime, err := parseTimeParam(req, test.paramName, test.defaultValue)
+
+		if err != nil {
+			testutil.ErrorEqual(t, result.asError(), err)
+		} else {
+			testutil.Assert(t, asTime.Equal(result.asTime), "time as return value: %s not parsed correctly. Expected %s. Actual %s", test.paramValue, result.asTime, asTime)
+		}
+
```
roidelapluie: unneeded blank line :)

web/api/v1/api_test.go
```diff
@@ -2171,6 +2171,69 @@ func TestRespondError(t *testing.T) {
...
+			defaultValue: maxTime,
+			result: resultType{
+				asTime: time.Time{},
+				asError: func() error {
+					_, err := parseTime("baz")
+					return errors.Wrapf(err, "Invalid time value for '%s'", "foo")
+				},
+			},
+		},
+	}
```
roidelapluie: missing blank line :)

web/api/v1/api_test.go
```diff
@@ -2171,6 +2171,69 @@ func TestRespondError(t *testing.T) {
 	}
 }
 
+func TestParseTimeParam(t *testing.T) {
+	type resultType struct {
+		asTime  time.Time
+		asError func() error
+	}
+	ts, err := parseTime("1582468023986")
+
```
roidelapluie: extra blank line

web/api/v1/api_test.go
```diff
@@ -2171,6 +2171,69 @@ func TestRespondError(t *testing.T) {
 	}
 }
 
+func TestParseTimeParam(t *testing.T) {
+	type resultType struct {
+		asTime  time.Time
+		asError func() error
+	}
```
roidelapluie: missing blank line :)

web/api/v1/api_test.go
boyskila: I ran the ```go fmt``` command but looks like it's not working for blank lines

web/api/v1/api_test.go
boyskila: Haha...you are killing me :D

web/api/v1/api.go
roidelapluie: It looks like it is not reverted?

roidelapluie:
Thanks @boyskila

PR #6861, opened 2020-02-24
Redundant parentheses

liguozhong:
Redundant parentheses


liguozhong:
image

brian-brazil:
The code is clearer as-is, I think these should stay.

PR #6867, opened 2020-02-25
Use a cryptographically secure random generator for ULID

pracucci:
Currently, the ULID entropy generator is quite poor because its seed is initialized each time with  the current timestamp (but also the first part of the ID is based on the current timestamp), so it doesn't add a real entropy to the second part block ID.

In this PR I'm proposing to switch to a cryptographically secure random generator.

/cc @codesome @pstibrany @bwplotka

### DIFF
--- a/tsdb/compact.go
+++ b/tsdb/compact.go
@@ -15,10 +15,10 @@ package tsdb
 
 import (
 	"context"
+	"crypto/rand"
 	"fmt"
 	"io"
 	"math"
-	"math/rand"
 	"os"
 	"path/filepath"
 	"sort"
@@ -413,8 +413,7 @@ func (c *LeveledCompactor) Compact(dest string, dirs []string, open []*Block) (u
 		uids = append(uids, meta.ULID.String())
 	}
 
-	entropy := rand.New(rand.NewSource(time.Now().UnixNano()))
-	uid = ulid.MustNew(ulid.Now(), entropy)
+	uid = ulid.MustNew(ulid.Now(), rand.Reader)
 
 	meta := compactBlockMetas(uid, metas...)
 	err = c.write(dest, meta, blocks...)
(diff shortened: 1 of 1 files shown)


roidelapluie:
Why would we want that? It does not seem there is an actual issue with this. We don't use that for security.

roidelapluie:
I get that this is an improvement in the code however, so that would be refactoring, but not really security.

pracucci:
Improving the entropy would reduce the chances of ULID collisions (they're
already very low but with a better entropy we would get an higher
confidence).

pracucci:
Sorry for not being clear with the initial PR description. This PR aims to
improve the entropy to avoid ULID collisions in environments were a large
number of TSDB gets compacted concurrently and there's the requirement to
have unique block IDs across all TSDB (ie. The new Cortex blocks storage).

tsdb/compact.go
```diff
@@ -468,8 +467,7 @@ func (c *LeveledCompactor) Compact(dest string, dirs []string, open []*Block) (u
 func (c *LeveledCompactor) Write(dest string, b BlockReader, mint, maxt int64, parent *BlockMeta) (ulid.ULID, error) {
 	start := time.Now()
 
-	entropy := rand.New(rand.NewSource(time.Now().UnixNano()))
-	uid := ulid.MustNew(ulid.Now(), entropy)
+	uid := ulid.MustNew(ulid.Now(), rand.Reader)
```
bwplotka: Can we pass entropy to compactor instead?

bwplotka:
Mentioned discussion: https://github.com/cortexproject/cortex/issues/2069

roidelapluie:
Prometheus will never create two blocks at the same time.

Le mar. 25 févr. 2020 à 12:41, Marco Pracucci <notifications@github.com> a
écrit :

bwplotka:
Sure, but it's inside TSDB and many other projects use the Compactor package.

roidelapluie:
I do not think it is wise to use compactor in different go routines, even
for dependencies. Afaik Prometheus does not do this.

brian-brazil:
I don't think we should be that strict here, this is a reasonable change.

roidelapluie:
Yes it's a nice refactoring. I assume that we already seed crypto/Rand.


Le mar. 25 févr. 2020 à 13:31, Brian Brazil <notifications@github.com> a
écrit :

pstibrany:
> Yes it's a nice refactoring. I assume that we already seed crypto/Rand.

Go runtime makes sure it's secure (uses OS-provided ways).

roidelapluie:
Thanks!

bwplotka:
Well... @roidelapluie you merged PR that had still unaddressed comment. ):

bwplotka:
Maybe it's my bad: I sometimes mark PR a green `approved` but still with `One comment and LGTM` - maybe I should mark it green only when fully approved?

roidelapluie:
Oh sorry. Indeed I misunderstood your approval.

PR #6869, opened 2020-02-25
docs: minor corrections to the docker documentation

larsks:
The documentation referenced "data volume containers", which were
superseded by named volume support in Docker several years ago.

There were to bind-mounting examples in the docs that are effectively
doing the same thing, but the description of the second was somewhat
erroneous.


brian-brazil:
Can you add the DCO?

brian-brazil:
Thanks!

PR #6870, opened 2020-02-25
Introduce a config option to customise TSDB entropy generator

pracucci:
The PR #6867 has been merged before I had the chance to address @bwplotka [comment](https://github.com/prometheus/prometheus/pull/6867/files#r383830627):

> Can we pass entropy to compactor instead?

In this PR:
- Added `Entropy` to DB options (nil value == default)
- Rollback default entropy generator to the one we had before PR #6867 was merged

Questions:
- `DBReadOnly` has no options support. Should we always use the default (as in this PR), pick the entropy in input or add options support to `DBReadOnly` too?

### DIFF
--- a/tsdb/block_test.go
+++ b/tsdb/block_test.go
@@ -251,7 +251,7 @@ func TestBlockSize(t *testing.T) {
 		testutil.Ok(t, err)
 		testutil.Equals(t, expAfterDelete, actAfterDelete, "after a delete reported block size doesn't match actual disk size")
 
-		c, err := NewLeveledCompactor(context.Background(), nil, log.NewNopLogger(), []int64{0}, nil)
+		c, err := NewLeveledCompactor(context.Background(), nil, log.NewNopLogger(), []int64{0}, nil, nil)
 		testutil.Ok(t, err)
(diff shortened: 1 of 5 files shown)


brian-brazil:
This doesn't seem like something that should be part of a public API. Can we not hardcode to a good one internally?

pstibrany:
> Is hardcoded global one `rand.Reader` really a good one? 

It's crypto/rand.Reader. Yes, that's a good one.

pstibrany:
> They could just use good hardcoded value as well. (:

They should have indeed. There is very little extra value in making this optional.

bwplotka:
Why? It's useful for testing purposes so we can have deterministic ulid! (: I glad they did!

bwplotka:
But yea, I don't have very strong feelings about this being an option for compactor agree here.

Thanks for your work @pracucci anyway. I think we can postpone it until we will have problems because of this global variable used.

pstibrany:
> Why? It's useful for testing purposes so we can have deterministic ulid! (: I am glad they did.

That’s a valid use case indeed.

brian-brazil:
If you want to do that it'd seem better to pass in the name of the output directory, rather than indirecting through randomness.

pracucci:
_Closing this PR based on the discussion above_

PR #6872, opened 2020-02-25
PromQL langserver api

slrtbtfs:
This adds the PromQL langserver HTTP API to to a new `/langserver` endpoint of Prometheus. For an extensive discussion, see #6160 .

Things that need to be done before this can be merged (incomplete list):

- [x] There should be a tagged langserver release including the new API, that Prometheus can depend upon.
- [x] The logging output of the langserver should use the same logging solution as Prometheus, not plain stderr.
-  [ ] There needs to be more agreement, whether we want this and if yes, where the new API endpoint should be located.


web/web.go
```diff
@@ -402,6 +404,15 @@ func New(logger log.Logger, o *Options) *Handler {
 		router.Get("/user/*filepath", route.FileServe(o.UserAssetsPath))
 	}
 
+	langserver, err := langserver.CreateHandler(context.Background(), fmt.Sprint("http://", o.ListenAddress))
```
slrtbtfs: Whats the right way to deal with an error raised here?

There are two cases which could cause an error here:

* A invalid URL argument.
* A bug in the language server.


(I'm aware the `fmt.Println` statement below isn't proper error handling)

slrtbtfs:
cc @juliusv @boyskila @squat

slrtbtfs:
cc @juliusv @boyskila @squat

slrtbtfs:
cc @juliusv @boyskila @squat

slrtbtfs:
cc @juliusv @boyskila @squat

slrtbtfs:
Sorry for the duplicate comments (deleted now).

GitHub reported the comment could not be sent. Later all retries were added to the PR as separate comments.

web/web.go
brian-brazil: In either case, Prometheus should fail to start. It sounds like you're not handling external url correctly here though.

juliusv: Yeah, I guess we'd always want to connect to localhost instead of the external URL's hostname though (which might not even be reachable from the Prometheus machine), but we'd still have to append the route prefix.

Even better would be if we could avoid the network completely for these internal calls. Like if the langserver handler could take a Prometheus API object or HTTP handler, instead of a Prometheus URL to connect to. That's just a very rough idea though, no idea how doable that is :)

roidelapluie: I think we should indeed avoid http requests here.

juliusv:
> There needs to be more agreement, whether we want this and if yes, where the new API endpoint should be located.

Probably *somewhere* under `/api` (since it's for machines, though we have `/federate` at the root as well, but that's similar enough to `/metrics`), but maybe under a completely separate hierarchy like `/api/langserver`. IMO the most important thing is that we mark it clearly experimental at the beginning so we can experiment with it and still decide on another path later.

vendor/modules.txt
```diff
@@ -196,6 +196,8 @@ github.com/gophercloud/gophercloud/openstack/identity/v2/tokens
 github.com/gophercloud/gophercloud/openstack/identity/v3/tokens
 github.com/gophercloud/gophercloud/openstack/utils
 github.com/gophercloud/gophercloud/pagination
+# github.com/gorilla/websocket v1.4.1
+github.com/gorilla/websocket
```
brian-brazil: I presume this is an indirect dependency that we're not actually using?

web/web.go
brian-brazil: Yeah, this shouldn't be hitting the network stack at all. We've had enough problems from that in the past.

juliusv:
> If this is a pure internal API that's never going to be used by others, then where it is is fine and we should explicitly document it as such in the stability guarantees.

In my understanding this is meant to be used by Grafana and other UIs that want to integrate better autocompletion and other language server features (while not requiring the normal stateful language server protocol).

brian-brazil:
Then we need to fully specify it under v1.

vendor/modules.txt
slrtbtfs: Yes.

I'm planning to do some refactoring in the language server, that will stop pulling in this indirect dependency.

slrtbtfs: Fixed in https://github.com/prometheus-community/promql-langserver/pull/116 .

web/web.go
slrtbtfs: https://github.com/prometheus-community/promql-langserver/issues/117

slrtbtfs:
Since there is still some work required on the language server side, I've created the [REST API](https://github.com/prometheus-community/promql-langserver/labels/REST%20API) label in the language server issues.

Help with those issues is appreciated, especially since today is my last day of my Red Hat internship, which means that I'm no longer working full time on Prometheus and language server related stuff.

roidelapluie:
> today is my last day of my Red Hat internship, which means that I'm no longer working full time on Prometheus and language server related stuff.

Thanks for all your work!

web/web.go
slrtbtfs: The language server now has been updated in a way that allows not using the network stack (Thanks to @Nexucis !). This PR will be eventually updated accordingly.

slrtbtfs:
I'm somewhat confused about the logging part. The `web` package currently seems to use `go-kit/logger` while according to @SuperQ on IRC there seems to be an effort to switch to `go-common/promlog`. So which kind object should the langserverHandler use?

web/web.go
Nexucis: You are welcome :)

Just to complete a bit what @slrtbfs said, I created an interface that handle the request to Prometheus:
https://github.com/prometheus-community/promql-langserver/blob/11052ee3baf9ed3f5ddfc1d604ba819bbf21377b/prometheus/client.go#L88-L104

The method createHandler is taking in parameter this interface. 
What is remain now is to create a client that implements this interface and that won't use the network.

But the main question de have today with @slrbtfs is where to put this implementation. ( Talk is here https://github.com/prometheus-community/promql-langserver/issues/134) For me it makes sense to do it on Prometheus side and to not have it in the langserver side.

Do you think the same ?

brian-brazil:
Those are the same thing, pass around go-kit loggers.

web/web.go
juliusv: @Nexucis Others who are deeper into this may speak up and disagree, but from my current understanding I would also think that a Prometheus-server-specific (non-networked) implementation of that metadata client would fit best into the Prometheus repo itself.

slrtbtfs:
> > There needs to be more agreement, whether we want this and if yes, where the new API endpoint should be located.
> 
> Probably _somewhere_ under `/api` (since it's for machines, though we have `/federate` at the root as well, but that's similar enough to `/metrics`), but maybe under a completely separate hierarchy like `/api/langserver`. IMO the most important thing is that we mark it clearly experimental at the beginning so we can experiment with it and still decide on another path later.

I'd prefer the API to be marked experimental (with full specs) until we've gathered some experience with it, too. It can still be stabilized later.

So maybe `/api/langserver` would be a good place.

slrtbtfs:
Also @Nexucis is going to continue working on this PR, since I currently don't have the necessary time.

Thanks for helping with this!

Nexucis:
Hello everyone,

I restarted a bit the PR and as you can see I introduced a new object `webkit`. I use it to put in common the code used by the `api/v1` and the code for the light prometheus client needed by the langserver.

I'm feeling this webkit could be improved a bit more in order to rationalized the code that is in the api.
The business logic would be in this object `webkit` and the logic of checking all HTTP parameter remains to the API.
Like that each logic are splitted and dedicated to a single different object.

I hope you will like it :). 

Also I finally put the langServer in the `api/v1` after coding the light prometheus client + the webkit, I think it makes more sense to have it as well in the same package since langServer is using all object from the `api/v1`

slrtbtfs:
The name webkit is a bit confusing since there is already a well known HTML engine with that name.

slrtbtfs:
@Nexucis There should be documentation about the new API in `docs/querying/api.md`.

That documentation can to some extend be taken from https://pkg.go.dev/github.com/prometheus-community/promql-langserver@v0.5.1/rest?tab=doc .

Nexucis:
I'm not really againts to change the name `webkit` if it's really confusing. I named it like that because mainly I like the name ^^ and also I didn't know this HTML engine. 
You can propose something to change it's name if you have something already in mind it's not a problem

For the documentation I will do my best to have something nice

web/api/v1/langserver.go
```diff
@@ -0,0 +1,134 @@
...
+		Help: metadata[metric][0].Help,
+		Unit: metadata[metric][0].Unit,
+	}, nil
+}
+
+// AllMetadata returns metadata about metrics currently scraped for all existing metrics.
+func (c *prometheusLangServerClient) AllMetadata(ctx context.Context) (map[string][]v1.Metadata, error) {
+	metadata := c.webkit.metadata(ctx, "", -1)
+	result := make(map[string][]v1.Metadata)
+	for metric, infos := range metadata {
```
slrtbtfs, line 69: Can we avoid the copying here?

Nexucis: I don't think that's possible. The webkit is returning the same map but with the private struct `metadata` which is different from the one coming from the public package `v1` from the go client.

The only way to avoid it so far is that the prometheus API is using the public struct `v1.Metadata` instead of the private one.

Nexucis: which could make sense since that would mean the client lib and the API would share the same data model. But this something I cannot decide by my own.

juliusv:
Looking at it! Just a bit swamped, but in the next days :)

brian-brazil:
Why does this seem to be duplicating many of the existing v1 APIs?

slrtbtfs:
> Why does this seem to be duplicating many of the existing v1 APIs?

The language server basically wants the same information as the v1 API provides, but in a non-networked way.

But instead of duplicating everything this PR creates an abstraction named `webkit` which is then used both by the v1 API and by the implementation of the interface provided by the language server.

Also see https://github.com/prometheus/prometheus/pull/6872#issuecomment-636517796 .

brian-brazil:
I'm not following. If it wants to talk to the TSDB without using the network it can do do directly without adding HTTP endpoints.

slrtbtfs:
We're not adding endpoints (except those provided by the language server).

The language server instance providing the langserver API talks to an implementation of this interface:

https://github.com/prometheus-community/promql-langserver/blob/3992a10afc80d38a7949b9d719e2d30db4ff222c/prometheus/client.go#L87-L104

In a standalone langserver, the implementation talks to the prometheus v1 API.

In case of the langserver providing the langserver API in Prometheus, the implementation of that interface resides in `web/api/v1/langserver.go`.

Since the data that is retrieved from TSDB here is pretty much the same as in the v1 API implementation, some the TSDB access Code of the v1 API has been moved into an object called `webkit` which is implemented in `web/api/v1/webkit.go`.

Both the v1 API and the interface implementation used by the langserver then talk to a `webkit` instance instead of talking to TSDB directly.

brian-brazil:
Surely we can do this without having to add indirection to how the API works?

brian-brazil:
Could we not do something along the lines of using the existing http router?

slrtbtfs:
For the TSDB access or for something else?

slrtbtfs:
Another way to handle this would be to call the respective methods of the v1 api object in the interface that delivers data for the langserver.


Like this one:

https://github.com/prometheus/prometheus/blob/4284dd1f1b8ec701673d65e4d12db1df59851061/web/api/v1/api.go#L484-L505

That would mean assembling a http request including the right parameters, passing it to the method, and forwarding the result to the language server.

Nexucis:
On my point of view it was a good enhancement to split the business logic and the checking of the api parameter.
And at least that's a quite simple approach to let the langserver used prometheus without any network thinguy since the client interface has been mostly désigned based on what the API was returning.

I thought the initial idea was to avoid to build an HTTP request. And introducing the `webkit` doesn't increase the complexity of the code. I think that's quite the opposite actually. But it's just my opinion

brian-brazil:
> And introducing the webkit doesn't increase the complexity of the code. 

I disagree there, it's extra complexity for the general API code.

brian-brazil:
> That would mean assembling a http request including the right parameters, passing it to the method, and forwarding the result to the language server.

To be clear, this seems fine to me.

Nexucis:
For me doing that is comparable to using a non typed language such as Javascript.
Basically what you are ok for is more or less the same thing than saying:

"I'll compress all my parameter in a single string and then I will pass this string to my method hoping that this method will parse correctly the parameters I passed." 

Instead of using some typed parameter which can be checked at compilation level.

Furthermore, what I proposed here is not really something new. That's usually the way to design an HTTP API : 
* **First layer** is the HTTP endpoint which is in charge of extracting and checking that the HTTP parameters are statically correct (correct type, correct range of value ...etc.). Once all parameters are ok, it passed them to the 2nd layer which won't have to verify anymore that the type of the parameter are not ok.
* **Second layer** is the business layer which is in charge of checking that the parameters are ok in dynamic way (like the resources wanted already existed in the database or not). It's also contains all the business logic and once everything is ok it asked to the third layer to perform some actions.
* **Third layer** can be the database or a proxy to the database

And finally, you can see the `webkit` as a way to enter in prometheus without considering the HTTP protocol. It could be used by other protocol for example.

brian-brazil:
Adding a single new HTTP API should not require moving the code around for existing HTTP APIs, this change is proposing things beyond what it needs to actually work. If you want to propose a rearchitecture of how our APIs work that should be done as a separate proposal, not bundled in with a different feature.

juliusv:
> Adding a single new HTTP API should not require moving the code around for existing HTTP API

I think it makes sense to factor out common model access code. I am with @Nexucis on the argument that it would be worse to funnel everything through a layer of string serialization+deserialization that is not needed and introduces lack of type safety.

web/api/v1/langserver.go
```diff
@@ -0,0 +1,134 @@
...
+				},
+			},
+		}, time.Now().Add(-100*time.Hour), time.Now())
+	if err != nil {
+		return nil, err
+	}
+	var result []string
+	for _, ln := range labelNames {
+		for l := range ln {
+			result = append(result, string(l))
```
juliusv, line 105: Isn't this missing deduping of label names when the same label name occurs in multiple series?

Also, the `for l := range ln` makes `l` be the integer index of the current label name, not the actual string label name, so I think it returns a set of bogus strings: https://play.golang.org/p/mWGGPV4XASc. I think you want `for _, l := range ln` and then append `l.Value` instead of `string(l)`: https://play.golang.org/p/v4u7tVsRteo

web/api/v1/langserver.go
```diff
@@ -0,0 +1,134 @@
...
+	return result, nil
+}
+
+// LabelValues performs a query for the values of the given label.
+func (c *prometheusLangServerClient) LabelValues(ctx context.Context, label string) ([]model.LabelValue, error) {
+	values, _, err := c.webkit.labelValues(ctx, label, minTime, maxTime)
+	if err != nil {
+		return nil, err
+	}
+	var labelValues []model.LabelValue
```
juliusv, line 117: When the length of the required slice is known beforehand, allocate the required length right away:

```suggestion
	labelValues := make([]model.LabelValue, 0, len(values))
```

web/api/v1/api.go
```diff
@@ -521,28 +520,13 @@ func (api *API) labelValues(r *http.Request) (result apiFuncResult) {
...
-	// in the case of a panic).
-	defer func() {
-		if result.finalizer == nil {
-			q.Close()
-		}
-	}()
-	closer := func() {
-		q.Close()
-	}
+	vals, warnings, err := api.webkit.labelValues(r.Context(), name, start, end)
```
juliusv, line 523: In all these cases, I haven't double-checked all the logic around this finalizer stuff yet, especially around where a panic can happen (what is outlined in the removed comments). There seems to be a reason why we don't usually close the querier directly in the handler functions, but pass it up to the caller and let it handle it at the end? Something to double-check at the end...

web/api/v1/langserver.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	return v1.Metadata{
+		Type: v1.MetricType(metadata[metric][0].Type),
+		Help: metadata[metric][0].Help,
+		Unit: metadata[metric][0].Unit,
+	}, nil
+}
+
+// AllMetadata returns metadata about metrics currently scraped for all existing metrics.
+func (c *prometheusLangServerClient) AllMetadata(ctx context.Context) (map[string][]v1.Metadata, error) {
```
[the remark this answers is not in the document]
Nexucis: The prometheusLangServerClient is implementing the interface coming from here: https://github.com/prometheus-community/promql-langserver/blob/3992a10afc80d38a7949b9d719e2d30db4ff222c/prometheus/client.go#L88

So that's its exported :)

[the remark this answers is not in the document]
juliusv: I see, thanks! Then ignore the export comments here :)

Might still be good to qualify the `Metadata()` / `AllMetadata()` methods in the original interface at some point to clarify that they are about metrics metadata and not some other metadata (like series names).

slrtbtfs:
Another issue with this PR that needs to be considered is the circular module dependency introduced here:

`prometheus/web/api/v1  -----> promql-langserver/rest -----> promql-langserver/langserver -----> prometheus/promql/parser`.

As long as there are no sufficiently breaking changes to the `prometheus/promql/parser` package this surprisingly works fine, but if someone wants to introduce such changes, it gets messy.

web/api/v1/langserver.go
Nexucis: > Isn't this missing deduping of label names when the same label name occurs in multiple series?

Yep  it needs that. Thanks !

> Also, the for l := range ln makes l be the integer index of the current label name,

oula yes, big mistake here. Thanks for the highlight!

web/api/v1/langserver.go
```diff
@@ -0,0 +1,134 @@
...
+				Type: v1.MetricType(info.Type),
+				Help: info.Help,
+				Unit: info.Unit,
+			}
+		}
+	}
+	return result, nil
+}
+
+// LabelNames returns all the unique label names present in the block in sorted order.
```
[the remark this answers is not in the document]
Nexucis: maybe instead of using min and max for this case, we could use the hardcoded time `-100*time.Hour`

Or we just put the timerange as a parameter in the method. But that would just move the hardcoded value to the caller.

web/api/v1/webkit.go
```diff
@@ -0,0 +1,135 @@
...
+	"context"
+	"time"
+
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/pkg/timestamp"
+	"github.com/prometheus/prometheus/storage"
+)
+
+// Webkit is used by the apiv1 and the langServer to retrieve metrics, labels ...etc.
+type webkit struct {
```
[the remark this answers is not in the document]
Nexucis: I think it's depending if you want after this PR another one to refactor the complete API.
If the `webkit` is going to hold the business logic for all endpoint, then maybe another name is needed.

Otherwise, `metadataProvider` is fine for me.

web/api/v1/langserver.go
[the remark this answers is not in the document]
Nexucis: Yeah that would be more appropriate. I guess that's fine to rename it even if it will introduce a breaking change in the langserver. Are you ok with that @slrtbtfs ?

web/api/v1/api.go
Nexucis: For this part, to be honest I didn't understand why it was like that. Because in all cases the finalizer is the same.
And the `defer` function is called even during a panic. Or to do so, it should be wrapped in a anonymous function like it was the case before. Maybe that's the only thing that should be modified here.

But maybe it required some golang knowledge I don't have.

web/api/v1/langserver.go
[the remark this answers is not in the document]
slrtbtfs: That part of the language server explicitly has no stability guarantees right now, so it's fine with me.

web/api/v1/webkit.go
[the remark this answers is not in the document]
juliusv: Yeah, I don't think we should add more stuff to it, so IMO `metadataProvider` should be fine. Though actually I don't totally like the `Provider` part that much either, as usually I associate `Provider` with switching out the complete source of some data or functionality, rather than just something that provides an intermediate transformation step. But oh well, it's ok for now :)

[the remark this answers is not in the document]
Nexucis: or maybe `metadataService` ?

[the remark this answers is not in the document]
juliusv: For me that sounds like the same category as `Provider`, but fine with me :)

web/api/v1/api.go
brian-brazil: This isn't a Go thing, it's a Prometheus/TSDB thing. Any strings returned are only valid until the finaliser is called, as they could be yoloStrings backed by mmap.

web/api/v1/langserver.go
```diff
@@ -0,0 +1,134 @@
...
+	if len(metricName) == 0 {
+		names, _, err := c.webkit.labelNames(ctx, minTime, maxTime)
+		return names, err
+	}
+	labelNames, _, err := c.webkit.series(ctx,
+		[][]*labels.Matcher{
+			{
+				&labels.Matcher{
+					Type:  labels.MatchEqual,
+					Name:  "__name__",
```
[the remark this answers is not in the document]
brian-brazil: Please avoid any additional uses of model, especially for accessing constants. Limit it to time parsing stuff.

web/api/v1/langserver.go
```diff
@@ -0,0 +1,134 @@
...
+	return labelValues, nil
+}
+
+// ChangeDataSource is not used in this context because we are inside prometheus.
+// So datasource/url doesn't mean anything and cannot be changed
+func (c *prometheusLangServerClient) ChangeDataSource(_ string) error {
+	return nil
+}
+
+// GetURL is returning the url used to contact the prometheus server
```
brian-brazil, line 130: Why is this needed if all communication is internal?

web/api/v1/webkit.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	return res
+}
+
+func (w *webkit) labelNames(ctx context.Context, start time.Time, end time.Time) ([]string, storage.Warnings, error) {
+	q, err := w.queryable.Querier(ctx, timestamp.FromTime(start), timestamp.FromTime(end))
+	if err != nil {
+		return nil, nil, err
+	}
+	defer q.Close()
```
brian-brazil, line 92: All these queriers are being closed way too early, the caller needs to do it once it's finished with the results.

web/api/v1/langserver.go
slrtbtfs: It's part of the interface provided by the language server for a generic data source.

For a normal language server it's used to provide links to view a query in the Prometheus Web UI.

If the language server is used to provide its REST API `GetURL` is never called.

slrtbtfs: It's part of the interface provided by the language server for a generic data source.

For a normal language server it's used to provide links to view a query in the Prometheus Web UI.

If the language server is used to provide its REST API `GetURL` is never called.

brian-brazil: Could we provide a dummy value then rather than plumbing through something that isn't used?

slrtbtfs: Yes.

slrtbtfs: What might be a cleaner solution is to change the interface function in the language server to allow returning errors and return an `not implemented` error here.

Nexucis: we could just return an empty string. It will be simpler.

web/api/v1/api.go
Nexucis: So if I understood correctly, if it closed too earlier, the string is going to be deleted ?

Sorry in advance for the dummy question, but I thought the string type cannot be changed like that unless you are using a pointer. And since it doesn't return a pointer, then how the result can be changed once the finaliser is called ?

web/api/v1/langserver.go
slrtbtfs: That'd be ok, too.

If we end up making changes to this interface because of other reasons, we can still add the error return value then.

web/api/v1/api.go
slrtbtfs: https://github.com/prometheus/prometheus/blob/0cff72d756d3b0eb1690dce66b9ca62b32f1deb8/tsdb/index/index.go#L1721-L1723

web/api/v1/langserver.go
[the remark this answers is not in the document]
Nexucis: but that's quite clean to use a constant. I hardcoded the string here because I didn't know about this constant.
It limited the number of error if the value is changing. But if it's really what you want, then I will keep the harcoded value.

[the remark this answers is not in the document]
slrtbtfs: @brian-brazil What's the problem with using `model`?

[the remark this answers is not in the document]
brian-brazil: It's largely a historical thing where various things were attempted to be over-formalised into separate types, code that should have been in one repo ended up split across two, and generally getting in the way. There's a few things where there is actually common code (e.g. duration parsing), but the rest should go away and live in the actual repo that cares about it.

Another example in this PR is using a special type for labelnames/values, where the current practice is to use a simple string.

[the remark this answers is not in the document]
juliusv: I don't feel strongly about the whole `model` repo or not, but still having this as a constant then somewhere in the Prometheus repo would make sense at least. But in the interest of not opening up that larger discussion (which would involve replacing the existing `model.MetricNameLabel` everywhere, I'm fine having it as a string literal here for now.

web/api/v1/langserver.go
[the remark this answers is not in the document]
juliusv: Yeah, so I guess the language server interface doesn't provide a way to pass time parameters for any metadata lookups yet?

In an ideal world this would be a setting that the frontend could dictate, depending on what kind of lookup it is doing (e.g. depending on graph range or something like that). In a slightly less ideal world, it would be a Prometheus-wide flag. In the short-term world, it could just be hardcoded here for now to the most recent couple of hours.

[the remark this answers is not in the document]
Nexucis: Yes for the moment the interface doesn't allow that. But I'm currently changing it with your last comments. So I can change it as well. It will give much more flexibility anyway and @slrtbtfs wanted to do it if I remember well. I just forgot a bit to change it.

web/api/v1/api.go
Nexucis: ok thanks for the answer. I think I understand better why you would like to avoid the introduction of the `webkit` struct @brian-brazil.

I will try to propose something easy to handle this closure issue.

Nexucis: but then if there is a problem here, then we will have the same problem for the langServer endpoints. So the endpoints won't return anything.

Nexucis: or to avoid this issue, I could copy the result of the query, so it will be immutable and then the closure can stay in the `defer` statement and in the method associated to the struct `webkit`

brian-brazil: That's one approach, though it'll create more garbage.

Nexucis: yeah you are right, that's not a good approach. I didn't have the mind clear lol.
I need to think more about it.

web/api/v1/langserver.go
```diff
@@ -0,0 +1,134 @@
...
+
+func newPrometheusLangServerClient(externalURL string, webkit *webkit) prometheus.Client {
+	return &prometheusLangServerClient{
+		externalURL: externalURL,
+		webkit:      webkit,
+	}
+}
+
+// Metadata returns the first occurrence of metadata about metrics currently scraped by the metric name.
+func (c *prometheusLangServerClient) Metadata(ctx context.Context, metric string) (v1.Metadata, error) {
```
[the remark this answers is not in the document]
Nexucis: done

web/web.go
```diff
@@ -301,6 +301,12 @@ func New(logger log.Logger, o *Options) *Handler {
 	factoryAr := func(_ context.Context) api_v1.AlertmanagerRetriever { return h.notifier }
 	FactoryRr := func(_ context.Context) api_v1.RulesRetriever { return h.ruleManager }
 
+	langServerHandler, err := api_v1.NewLangServer(o.ListenAddress, h.storage, factoryTr, logger)
```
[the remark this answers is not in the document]
Nexucis: the langServer is now completely built in the `api/v1` and is not anymore accessible outside of the package

web/api/v1/webkit.go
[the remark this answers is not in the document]
Nexucis: so final conclusion here:
* The prometheus client on langserver side has been renamed to `MetadataService`
* webkit is renamed in favor of `MetatadaProvider`

web/api/v1/langserver.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/prometheus-community/promql-langserver/rest"
+	"github.com/prometheus/client_golang/api/prometheus/v1"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/storage"
+)
+
+func NewLangServer(externalURL string, q storage.Queryable,
+	tr func(context.Context) TargetRetriever, logger log.Logger) (http.Handler, error) {
+	// create the custom prometheus client dedicated to the langserver
```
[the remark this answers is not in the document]
Nexucis: normally it's done

web/api/v1/webkit.go
```diff
@@ -0,0 +1,135 @@
...
+		queryable:       queryable,
+		targetRetriever: targetRetriever,
+	}
+}
+
+func (w *webkit) metadata(ctx context.Context, metric string, limit int) map[string][]metadata {
+	metrics := map[string]map[metadata]struct{}{}
+	for _, tt := range w.targetRetriever(ctx).TargetsActive() {
+		for _, t := range tt {
+
```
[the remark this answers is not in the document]
Nexucis: done

web/api/v1/langserver.go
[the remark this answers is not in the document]
Nexucis: all renamed

web/api/v1/langserver.go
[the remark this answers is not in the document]
Nexucis: the literal `__name__` will stay like that until there is a consensus regarding the usage of the package `model`

web/api/v1/langserver.go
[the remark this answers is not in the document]
Nexucis: I added an attribute called `lookbackInterval` that will be used to change the interval duration.

It can be changed at the runtime if needed

web/api/v1/langserver.go
```diff
@@ -0,0 +1,134 @@
...
+	labelNames, _, err := c.webkit.series(ctx,
+		[][]*labels.Matcher{
+			{
+				&labels.Matcher{
+					Type:  labels.MatchEqual,
+					Name:  "__name__",
+					Value: metricName,
+				},
+			},
+		}, time.Now().Add(-100*time.Hour), time.Now())
```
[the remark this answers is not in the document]
Nexucis: that was a mistake. It's fixed now.

web/api/v1/langserver.go
Nexucis: fixed

web/api/v1/langserver.go
Nexucis: fixed

web/api/v1/api.go
Nexucis: So here I choose to return the query created. Like that the API can close it when needed.

On the langserver side, I had to copy the result before closing the query

I hope you are ok with this approach

web/api/v1/langserver.go
Nexucis: usage of the externalURL has been removed. The method returns an empty string

web/api/v1/webkit.go
Nexucis: normally it should be better now. Tell me if it's still not the case

web/web.go
```diff
@@ -301,6 +301,12 @@ func New(logger log.Logger, o *Options) *Handler {
 	factoryAr := func(_ context.Context) api_v1.AlertmanagerRetriever { return h.notifier }
 	FactoryRr := func(_ context.Context) api_v1.RulesRetriever { return h.ruleManager }
```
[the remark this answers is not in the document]
Nexucis: fixed

Nexucis:
So normally I addressed all changes requested. I just hope I didn't forget anything. 

The build is failing for the moment because  after the upgrade of the golang_client, the promTool is not up to date anymore.
The build should be fixed through the PR #7504

slrtbtfs:
So, the main thing still missing is documenting the new API, right?

Nexucis:
normally yes. But maybe after a 2nd review there will be other things to change.

Nexucis:
here we go, I added the documentation for the different endpoints concerned.

So finally, now everything has been fixed and so I'm ready for a second review if you are ok to do it again @juliusv @brian-brazil. 
Thanks by advance for your time.

docs/querying/api.md
```diff
@@ -1022,3 +1022,205 @@ $ curl -XPOST http://localhost:9090/api/v1/admin/tsdb/clean_tombstones
 ```
 
 *New in v2.1 and supports PUT from v2.9*
+
+## LangServer
+
+The following endpoints provides the different feature handled by LSP (Language Server Protocol) in a stateless way.
```
slrtbtfs, line 1028:
```suggestion
This API is designed to help Web UIs provide IDE like features when writing PromQL queries, without introducing the complexity of the stateful [Language Server Protocol (LSP)](https://microsoft.github.io/language-server-protocol/).

Internally it uses the [PromQL Language Server](https://github.com/prometheus-community/promql-langserver).

The returned values follow JSON schemes specified by the LSP.
To deal with these types on the client side it might be helpful to use the [`vscode-languageserver-types`](https://www.npmjs.com/package/vscode-languageserver-types)  NPM package which provides the respective types.
```

docs/querying/api.md
```diff
@@ -1022,3 +1022,205 @@ $ curl -XPOST http://localhost:9090/api/v1/admin/tsdb/clean_tombstones
...
+
+### Diagnostics
+This endpoint returns the different errors contained in the PromQL expression provided in the body
```
slrtbtfs, line 1031:
```suggestion
This endpoint returns the parsing errors caused by the PromQL expression provided in the body, if any.

The result is a list of  [`Diagnostic`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#diagnostic)s. An empty list means that the Query provided is valid PromQL.
```

docs/querying/api.md
```diff
@@ -1022,3 +1022,205 @@ $ curl -XPOST http://localhost:9090/api/v1/admin/tsdb/clean_tombstones
...
+      },
+      "severity": 1,
+      "source": "promql-lsp",
+      "message": "unknown function with name \"some_metric\""
+    }
+  ]
+```
+
+### Completion
+This endpoint returns a list of the possible value for the given word contains in the promQL expression provided
```
slrtbtfs, line 1070:
```suggestion
This endpoint provides completions based on the provided PromQL expression and cursor position as well as the data stored in the given Prometheus instance.

The result is a list of  [`CompletionItem`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_completion)s.

It is recommended to limit the number of results.
```

docs/querying/api.md
```diff
@@ -1022,3 +1022,205 @@ $ curl -XPOST http://localhost:9090/api/v1/admin/tsdb/clean_tombstones
...
+          }
+        },
+        "newText": "go_gc_duration_seconds_count"
+      }
+    }
+  ]
+```
+
+### Hover
+This endpoint returns a documentation according to the given word contains in the promQL expression provided
```
slrtbtfs, line 1132:
```suggestion
This endpoint provides documentation based on the provided PromQL expression and cursor position as well as the data stored in the given Prometheus instance. It is typically used when the mouse hovers over the query.

The result is a  [`Hover`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_hover).
```

docs/querying/api.md
```diff
@@ -1022,3 +1022,205 @@ $ curl -XPOST http://localhost:9090/api/v1/admin/tsdb/clean_tombstones
...
+    "end": {
+      "line": 0,
+      "character": 26
+    }
+  }
+}
+```
+
+### Signature Help
+This endpoint returns the signature associated to the method selected in the promQL expression provided
```
slrtbtfs, line 1192:
```suggestion
This endpoint provides a function signature if the provided cursor position is within a PromQL function call. Otherwise a falsy value is returned.

The result is a  [`SignatureHelp`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_signatureHelp).
```

docs/querying/api.md
```diff
@@ -1022,3 +1022,205 @@ $ curl -XPOST http://localhost:9090/api/v1/admin/tsdb/clean_tombstones
 ```
 
 *New in v2.1 and supports PUT from v2.9*
```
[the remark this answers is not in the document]
Nexucis: yeah that's normal, this line belongs to the tsdb endpoints ;). So normally it should be ok.
Perhaps I should add a line like this one. But no idea in which version it will be added.

docs/querying/api.md
Nexucis: done

docs/querying/api.md
Nexucis: done

docs/querying/api.md
Nexucis: done

docs/querying/api.md
Nexucis: done

docs/querying/api.md
Nexucis: done

docs/querying/api.md
```diff
@@ -1022,3 +1022,205 @@ $ curl -XPOST http://localhost:9090/api/v1/admin/tsdb/clean_tombstones
...
+
+## LangServer
```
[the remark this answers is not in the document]
Nexucis: done

.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea/
```
slrtbtfs, line 1: This should be added to you global gitignore instead of this repo.

gouthamve:
@juliusv @slrtbtfs @Nexucis Hi folks, we were looking at this pr in the bug scrub and were wondering what the status of it is. Would be great to have the langserver support in Prometheus!

slrtbtfs:
As far as I know, the PR should be mostly complete, except rebasing and doing a final review.


To clear up some possible confusion: For a while @Nexucis has worked on completing this, so despite having opened the PR I've acted more as a reviewer than as the author of the PR.

juliusv:
So I've been talking with @Nexucis about this, and we're both more and more unsure whether it's worth adding this to Prometheus itself at the moment. The reason is that in our explorations around PromQL editing features in https://github.com/prometheus-community/codemirror-promql (the non-LSP modes), it was a clear benefit for autocompletion features and linting to work either completely or partially client-side, for both snappiness as well as complete offline support (no server present).

And we now have a client-side implementation of most of those features. The offline linter in https://github.com/prometheus-community/codemirror-promql is not 100% on par yet with the LSP one, but close. The offline autocomplete works really well by now, and since it can cache e.g. metric names and label names/values locally and make decisions around them, it's super fast, and data-independent autocompletion continues to work without a Prometheus server present. And for syntax highlighting, you want it to be client-side anyway, so we had to build a lezer grammar for that (https://github.com/promlabs/lezer-promql), and that helped build the other features as well.

So I still think LSP is great, but it's a much better fit for local editors than web apps. The REST wrapper around LSP helps, but it still doesn't take away the inherent drawback of having to do client-server roundtrips for everything, which https://github.com/prometheus-community/codemirror-promql avoids.

slrtbtfs:
I mostly agree with what @juliusv wrote above, so I'd propose to close this PR if everyone is fine with that.

Additionally codemirror-promql is currently being worked on way more actively than the language server.

Still, it was really nice to work on this and possibly useful for exploration. Thanks to anyone who worked on this and provided reviews!

juliusv:
@slrtbtfs Yes, thanks for the exploration and sorry for taking so long with this! Maybe it will make sense at a later point, but for now let's close it then. I still think LSP for desktop editors is a good thing to have, but that's separate.

PR #6874, opened 2020-02-26
[fix] ui: th to td

liguozhong:
[fix] ui: th to td


liguozhong:
@juliusv  ui PR

brian-brazil:
Thanks!

PR #6886, opened 2020-02-27
[test] Add tests for "xor.go" to verify the accuracy of "delta of delta" compact algorithm and "XOR 'compact algorithm"

liguozhong:
[test] Add tests for "xor.go" to verify the accuracy of "delta of delta" compact algorithm and "XOR 'compact algorithm"

### DIFF
--- a/tsdb/chunkenc/chunk_test.go
+++ b/tsdb/chunkenc/chunk_test.go
@@ -217,3 +217,46 @@ func benchmarkAppender(b *testing.B, newChunk func() Chunk) {
 
 	fmt.Println("num", b.N, "created chunks", len(chunks))
 }
+
+func TestXor(t *testing.T) {
+	timestamp := int64(1234123324)
+	v := 83665193758603.123
+	num := 10
+	var samples []pair
+	for i := 0; i < num; i++ {
+		timestamp += int64(1000)
+		v += float64(100)
+		samples = append(samples, pair{t: timestamp, v: v})
+	}
+
+	chunk := NewXORChunk()
+	a, err := chunk.Appender()
+	if err != nil {
+		t.Fatalf("get appender: %s", err)
+	}
+	for _, p := range samples {
+		a.Append(p.t, p.v)
+	}
+
+	res := make([]pair, 0)
+	var it Iterator
+	it = chunk.Iterator(it)
+	for it.Next() {
+		t, v := it.At()
+		res = append(res, pair{t: t, v: v})
+	}
+	if it.Err() != io.EOF {
+		testutil.Ok(t, it.Err())
+	}
+
+	for i := 0; i < len(samples); i++ {
+		s := samples[i]
+		r := res[i]
+		if s.t != r.t {
+			t.Fatal("test compact timestamp fail ,delta of delta", s, r)
+		}
+		if s.v != r.v {
+			t.Fatal("test compact value fail ,xor", s, r)
+		}
+	}
+}
--- a/tsdb/chunkenc/chunk_test.go
+++ b/tsdb/chunkenc/chunk_test.go
@@ -247,6 +247,7 @@ func TestXor(t *testing.T) {
 		testutil.Ok(t, it.Err())
 	}
 
+	testutil.Equals(t, len(samples), len(res))
 	for i := 0; i < len(samples); i++ {
 		s := samples[i]
 		r := res[i]
--- a/tsdb/chunkenc/chunk_test.go
+++ b/tsdb/chunkenc/chunk_test.go
@@ -231,14 +231,12 @@ func TestXor(t *testing.T) {
 
 	chunk := NewXORChunk()
 	a, err := chunk.Appender()
-	if err != nil {
-		t.Fatalf("get appender: %s", err)
-	}
+	testutil.Ok(t, err)
 	for _, p := range samples {
 		a.Append(p.t, p.v)
 	}
 
-	res := make([]pair, 0)
+	res := make([]pair, num)
 	var it Iterator
 	it = chunk.Iterator(it)
 	for it.Next() {
@@ -252,11 +250,7 @@ func TestXor(t *testing.T) {
 	for i := 0; i < len(samples); i++ {
 		s := samples[i]
 		r := res[i]
-		if s.t != r.t {
-			t.Fatal("test compact timestamp fail ,delta of delta", s, r)
-		}
-		if s.v != r.v {
-			t.Fatal("test compact value fail ,xor", s, r)
-		}
+		testutil.Equals(t, s.t, r.t, "test compact timestamp fail ,delta of delta", s, r)
+		testutil.Equals(t, s.v, r.v, "test compact value fail ,xor", s, r)
 	}
 }
--- a/tsdb/chunkenc/chunk_test.go
+++ b/tsdb/chunkenc/chunk_test.go
@@ -224,8 +224,8 @@ func TestXor(t *testing.T) {
 	num := 10
 	var samples []pair
 	for i := 0; i < num; i++ {
-		timestamp += int64(1000)
-		v += float64(100)
+		timestamp += 1000
+		v += 100
 		samples = append(samples, pair{t: timestamp, v: v})
 	}
 
@@ -236,7 +236,7 @@ func TestXor(t *testing.T) {
 		a.Append(p.t, p.v)
 	}
 
-	res := make([]pair, num)
+	res := make([]pair, 0)
 	var it Iterator
 	it = chunk.Iterator(it)
 	for it.Next() {
--- a/tsdb/chunkenc/chunk_test.go
+++ b/tsdb/chunkenc/chunk_test.go
@@ -236,7 +236,7 @@ func TestXor(t *testing.T) {
 		a.Append(p.t, p.v)
 	}
 
-	res := make([]pair, 0)
+	res := make([]pair, 0, num)
 	var it Iterator
 	it = chunk.Iterator(it)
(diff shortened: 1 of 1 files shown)


tsdb/chunkenc/chunk_test.go
```diff
@@ -217,3 +217,46 @@ func benchmarkAppender(b *testing.B, newChunk func() Chunk) {
...
+		res = append(res, pair{t: t, v: v})
+	}
+	if it.Err() != io.EOF {
+		testutil.Ok(t, it.Err())
+	}
+
+	for i := 0; i < len(samples); i++ {
+		s := samples[i]
+		r := res[i]
+		if s.t != r.t {
```
roidelapluie: can you use testutil.Equals?

tsdb/chunkenc/chunk_test.go
```diff
@@ -217,3 +217,46 @@ func benchmarkAppender(b *testing.B, newChunk func() Chunk) {
...
+			t.Fatal("test compact timestamp fail ,delta of delta", s, r)
+		}
+		if s.v != r.v {
+			t.Fatal("test compact value fail ,xor", s, r)
```
roidelapluie: can you use testutil.Equals?

tsdb/chunkenc/chunk_test.go
```diff
@@ -217,3 +217,46 @@ func benchmarkAppender(b *testing.B, newChunk func() Chunk) {
...
 	fmt.Println("num", b.N, "created chunks", len(chunks))
 }
+
+func TestXor(t *testing.T) {
+	timestamp := int64(1234123324)
+	v := 83665193758603.123
+	num := 10
+	var samples []pair
+	for i := 0; i < num; i++ {
+		timestamp += int64(1000)
```
roidelapluie: is this explicit cast needed?

tsdb/chunkenc/chunk_test.go
```diff
@@ -217,3 +217,46 @@ func benchmarkAppender(b *testing.B, newChunk func() Chunk) {
...
+		v += float64(100)
```
roidelapluie: is this explicit cast needed?

tsdb/chunkenc/chunk_test.go
```diff
@@ -217,3 +217,46 @@ func benchmarkAppender(b *testing.B, newChunk func() Chunk) {
...
+		samples = append(samples, pair{t: timestamp, v: v})
+	}
+
+	chunk := NewXORChunk()
+	a, err := chunk.Appender()
+	if err != nil {
```
roidelapluie: can you use testutil.Ok ?

tsdb/chunkenc/chunk_test.go
```diff
@@ -217,3 +217,46 @@ func benchmarkAppender(b *testing.B, newChunk func() Chunk) {
...
+		t.Fatalf("get appender: %s", err)
+	}
+	for _, p := range samples {
+		a.Append(p.t, p.v)
+	}
+
+	res := make([]pair, 0)
```
roidelapluie:
```suggestion
	res := make([]pair, num)
```

tsdb/chunkenc/chunk_test.go
```diff
@@ -217,3 +217,46 @@ func benchmarkAppender(b *testing.B, newChunk func() Chunk) {
...
+	it = chunk.Iterator(it)
+	for it.Next() {
+		t, v := it.At()
+		res = append(res, pair{t: t, v: v})
+	}
+	if it.Err() != io.EOF {
+		testutil.Ok(t, it.Err())
+	}
+
+	for i := 0; i < len(samples); i++ {
```
roidelapluie:
```suggestion
	for i, sample := range samples {
```

tsdb/chunkenc/chunk_test.go
liguozhong: done,thanks for your review.

tsdb/chunkenc/chunk_test.go
liguozhong: done,thanks for your review.

tsdb/chunkenc/chunk_test.go
```diff
@@ -217,3 +217,40 @@ func benchmarkAppender(b *testing.B, newChunk func() Chunk) {
...
+	res := make([]pair, num)
+	var it Iterator
+	it = chunk.Iterator(it)
+	for it.Next() {
+		t, v := it.At()
+		res = append(res, pair{t: t, v: v})
+	}
+	if it.Err() != io.EOF {
+		testutil.Ok(t, it.Err())
+	}
```
roidelapluie:
```suggestion
	testutil.ErrorEqual(t, io.EOF, it.Err())
```

tsdb/chunkenc/chunk_test.go
```diff
@@ -217,3 +217,40 @@ func benchmarkAppender(b *testing.B, newChunk func() Chunk) {
...
+
+	for i := 0; i < len(samples); i++ {
+		s := samples[i]
+		r := res[i]
+		testutil.Equals(t, s.t, r.t, "test compact timestamp fail ,delta of delta", s, r)
+		testutil.Equals(t, s.v, r.v, "test compact value fail ,xor", s, r)
```
roidelapluie:
```suggestion
		testutil.Equals(t, s.t, r.t, "test compact timestamp fail (delta of delta)")
		testutil.Equals(t, s.v, r.v, "test compact value fail (xor)")
```

testutil.Equal will show the right values,

Would testutil.Equals(t, s, r) work, too?

tsdb/chunkenc/chunk_test.go
liguozhong: done,thanks for your review.

tsdb/chunkenc/chunk_test.go
liguozhong: done,thanks for your review.

tsdb/chunkenc/chunk_test.go
liguozhong: done,thanks for your review.

tsdb/chunkenc/chunk_test.go
liguozhong: r := res[i]

There is also a "res" array that needs to be looped. If there is no "I" index, it will be strange to loop the "res" array

tsdb/chunkenc/chunk_test.go
liguozhong: image

tsdb/chunkenc/chunk_test.go
liguozhong: done,thanks for your review.

tsdb/chunkenc/chunk_test.go
roidelapluie: Then can you add `testutil.Equals(t, len(res), len(samples))` ?

tsdb/chunkenc/chunk_test.go
roidelapluie: Then that would be testutil.Ok() .. don't we always get nil ?

roidelapluie: In wchich case do we get EOF? is that `if` needed ?

liguozhong: The code can't be changed like this. It can only be written in the original way

roidelapluie: So it means that sometimes it is EOF and other times nil ?

tsdb/chunkenc/chunk_test.go
liguozhong: If I write this way, I need to introduce an additional 'i' index into the code, which will make it more troublesome. I
n addition, "num = 5" memory consumption can be ignored, 5 is enough to verify the accuracy of the algorithm

roidelapluie:
```suggestion
	res := make([]pair, num, 0)
```
then :)

liguozhong:
All the code modification suggestions have been completed. Does this PR need to do anything else.
@roidelapluie

tsdb/chunkenc/chunk_test.go
roidelapluie: Anyway that does not make a difference, you can ignore this

tsdb/chunkenc/chunk_test.go
liguozhong: image
The "benchmark iterator" function also deals with this err problem

roidelapluie: funny, thanks

tsdb/chunkenc/chunk_test.go
liguozhong: image
done,Thanks

tsdb/chunkenc/chunk_test.go
liguozhong: image
done,thanks

roidelapluie:
Thanks @codesome 

FTR I have run coverage and it looks like indeed TestChunk covers it.

https://roidelapluie.be/uploads/coverage1.html
https://roidelapluie.be/uploads/coverage2.html

liguozhong:
> We already have `TestChunk`, I don't think we need this

thanks，It's my carelessness

liguozhong:
> Thanks @codesome
> 
> FTR I have run coverage and it looks like indeed TestChunk covers it.
> 
> https://roidelapluie.be/uploads/coverage1.html
> https://roidelapluie.be/uploads/coverage2.html

thanks

Issue #6878, opened 2020-02-26
promql: Add support for .Select calls concurrency. 


bwplotka:
Hi (again) Prometheus family :wave: 

## Proposal

We noticed with @brancz during our investigations on query latency that `Select` calls done by PromQL are done sequentially: 

![image](https://user-images.githubusercontent.com/6950331/75368505-c9705e00-58b9-11ea-9313-9a8c126aa819.png)

It appears that those are subqueries required for the main (quite complex) instant query. Looks like they are done sequentially, where in our opinion they could be done concurrently. 
We would like to propose:

* Running independent subqueries in different goroutines.
* Controlled by some `subqueryMaxConcurrency` flag.

Would that make sense or lack of concurrency here is intended?

**Use case. Why is this important?**
* We are just waiting on one subquery to finish to start another independent subquery. We just waste time, CPU / Resources. We could do those asynchronously.  

cc @roidelapluie @slrtbtfs @brian-brazil @beorn7

brian-brazil:
> Would that make sense or lack of concurrency here is intended?

It's more that noone brought it up yet. This is also nothing to do with subqueries, you're talking about selectors.

The one thing to watch for is needlessly spawning goroutines, we never need this for local storage for example. A traffic spike to a remote system could also cause it issues.

bwplotka:
Yes, that's why something max concurrency flag maybe make sense?
 
And even with this, the remote system should be it's own concurrency limit logics I guess.

Anyway.. looks like it's OK to start some PR for it?

roidelapluie:
> > Would that make sense or lack of concurrency here is intended?
> 
> The one thing to watch for is needlessly spawning goroutines, we never need this for local storage for example. A traffic spike to a remote system could also cause it issues.

Even when we run `foo and bar` ?

brian-brazil:
A hardcoded limit is probably fine, just to avoid a potential DOS on FDs or local ports.

krasi-georgiev:
but how can run the main query and the subquery in paralel if the main query depends on the subquery?

brian-brazil:
This issue is about selectors, not subqueries.

krasi-georgiev:
hm , @bwplotka can you give more details on what exactly would run in paraller?

bwplotka:
It's not only subqueries. All `Selects`: https://github.com/prometheus/prometheus/blob/master/storage/interface.go#L63

[...]

PR #6887, opened 2020-02-27
Parallelize series population

slrtbtfs:
Attempt at #6878

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -655,9 +655,14 @@ func (ng *Engine) populateSeries(ctx context.Context, q storage.Queryable, s *pa
 	// the variable.
 	var evalRange time.Duration
 
+	// Keeps track of running Select() processes
+	var wg sync.WaitGroup
+
+	// Error that might occur during series population subprocesses
+	var selectErr error
+	var selectErrMu sync.Mutex
+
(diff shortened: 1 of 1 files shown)


slrtbtfs:
TODO: We probably want the concurrency limit global or per engine, not per query.

slrtbtfs:
I've just run the engine benchmarks and it looks like this PR doesn't actually improve performance.

Rather, especially the queries that run in parallel seem to be much slower.

It might be that this is only the case with the teststorage used by the engine benchmark, though.

krasi-georgiev:
maybe a trace would explain what is happening?

brian-brazil:
It'll run slower against local storage, to do this efficiently you'd need to redesign the interface a bit.

bwplotka:
Do we want to redesign?

brancz:
Yeah it seems like this needs a little more thought.

slrtbtfs:
If anyone wants to take over, feel free.

roidelapluie:
Just to clarify: I would prefer if we defer this to next releases at this point. Let's keep that out of 2.17.0 scope.

bwplotka:
Not one works on this actively, so closing to not confuse contributors. Still, it will be linked on the issue so someone can grab this as a start (: Thanks @slrtbtfs !

PR #6888, opened 2020-02-27
Let promql & tsdb have their own go mod versioning

sylr:
This is an attempt at regularizing the go mod versioning of some of the 
packages used outside of prometheus itself (by cortex, thanos ... etc).

I choose to do so for **promql & tsdb** because I believe they are cornerstones of the prometheus ecosystem. If you believe others should also get the same treatment please mention it.

The purpose of this is to have a development release lifecycle for thoses
packages dissociated from the user oriented release lifecycle of
prometheus itself.

To ease the burden of managing internal dependencies inside
prometheus we bypass the go mod versioning of the newly versioned
packaged by using the replace directive in the root go.mod file, i.e.:

```
replace (
        github.com/prometheus/prometheus => ./
        github.com/prometheus/prometheus/promql => ./promql
        github.com/prometheus/prometheus/tsdb => ./tsdb
)
```

Since current major version of prometheus is v2 the go mod versioning of
promql and tsdb starts with v2 as well.

This means that external projects which would want to use those packages
should import:

- `github.com/prometheus/prometheus/promql/v2`
- `github.com/prometheus/prometheus/tsdb/v2`

Note that as described previously, with the replace directive set at the root of
prometheus, we should not use /v2 suffixes internally.

In order to release new versions of promql and tsdb new git tags would
need to be created following the pattern, `<pkg>/<semver>`, e.g:

- `promql/v2.0.0`
- `tsdb/v2.0.0`

By respecting this go mod convention we ensure that go mod will
correctly resolves the package tags and not the ones of prometheus.

```shell
# given that tag "tsdb/v2.0.0" exists
# this resolves tag "tsdb/v2.0.0" within repo github.com/prometheus/prometheus
go get github.com/prometheus/prometheus/tsdb/v2@v2.0.0

# given that tag "tsdb/v2.13.0" does NOT exist
# given that tag "v2.13.0" exists
# this resolves tag "v2.13.0" within repo github.com/prometheus/prometheus
# but since tsdb/go.mod did not exists at that time it will fail
go get github.com/prometheus/prometheus/tsdb/v2@v2.13.0
```


sylr:
The CI is going to fail as it needs some promu tweaks.

roidelapluie:
I do not think it is worth it. There are even more packages that are used, we recently learned that even web/ was imported by cortex, so that would not solve the problem.

roidelapluie:
Also would that mean that we need to duplicate the vendor directories? And that it could be different from when used by Prometheus and when importing tsdb? That does not seem promising.

sylr:
> I do not think it is worth it.

Of course it's worth it. All people depending on those packages will benefit from a proper versioning.

> Also would that mean that we need to duplicate the vendor directories

I removed vendoring altogether, it does not make sense to keep vendoring since go 1.13 and GOPROXY being turned on by default.

roidelapluie:
> > I do not think it is worth it.
> 
> Of course it's worth it. All people depending on those packages will benefit from a proper versioning.

We have seen people depending on pkg/, other from web...

> > Also would that mean that we need to duplicate the vendor directories
> 
> I removed vendoring altogether, it does not make sense to keep vendoring since go 1.13 and GOPROXY being turned on by default.

But go1.14 makes the right move to default to -mod=vendor when there is a mod directory

sylr:
> But go1.14 makes the right move to default to -mod=vendor when there is a mod directory

https://golang.org/doc/go1.14#go-command

I removed the top vendor/ directory so it will not default to -mod=vendor

roidelapluie:
Anyway, there was no consensus on the mailing list about this, so please continue the discussion there.

Thanks for your work anyway.

Issue #6668, opened 2020-01-20
PromQL: add "involved samples" for query stats


Betula-L:
## Proposal

Prometheus will record `timings stats` for every query, such as `evalTotalTime `, `resultSortTime` now. It need more stats to evaluate query performance.

**What is we need?**

Expose stats "involved sample" as a query stats, i.e. `currentSamples` https://github.com/prometheus/prometheus/blob/0eb34299dad0ab997bc9d9363a06d6ea6eb45dad/promql/engine.go#L785 

**Why is it important?**

The `timings stats` is an unstable metric, which may be affected by hardware, query queue, etc., especially it's a random number. Sometimes we wanna evaluate whether query was bad or good, using `timings stats` is not a good idea.

However, `currentSamples` is stable and reliable which precisely records how many samples a query travelled. It also reflect how much computation resource consumed, such as memory or cpu.

brian-brazil:
Current samples is a guage, not a counter, so is updated throughout the query. What you'd want is peak samples.

Betula-L:
https://groups.google.com/forum/#!topic/prometheus-users/kZ2DvZnYHUA

Harkishen-Singh:
I am taking this up. Thank you.

roidelapluie:
I think we want: the number of series, the number of samples and the peak number of samples. Thanks.

roidelapluie:
Expectations:

```
$ curl 127.0.0.1:9090/api/v1/query?query=up\&stats=true |python -m json.tool
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   354  100   354    0     0   345k      0 --:--:-- --:--:-- --:--:--  345k
{
    "data": {
        "result": [
            {
                "metric": {
                    "__name__": "up",
                    "instance": "127.0.1.1:9090",
                    "job": "prometheus"
                },
                "value": [
                    1582924276.93,
                    "1"
                ]
            }
        ],
        "resultType": "vector",
        "stats": {
            "samples": {
                "peakSamples": 9000,
                "totalSamples": 12000,
                "totalEvaluators": 2,
              },
            "timings": {
                "evalTotalTime": 5.4748e-05,
                "execQueueTime": 6.421e-06,
                "execTotalTime": 6.4869e-05,
                "innerEvalTime": 1.4978e-05,
                "queryPreparationTime": 3.7253e-05,
                "resultSortTime": 0
            }
[...]

PR #6890, opened 2020-02-27
Adds support for involved samples for query stats

Harkishen-Singh:
Fixes: #6668


roidelapluie:
Hello, thank you!

This is not right. We want to update the stats, not having an extra metric.

Plus, we want two stats:

Total touched samples and maximum touched samples at a certain point.

I have imagined adding in stats:

(pseudocode)
```
stats.Observe(samples){
 if samples > max {
  max = samples
 }
 total += samples
}
```

Harkishen-Singh:
@roidelapluie I think the stats are being sent as metrics. That's why when we have an extra parameter `stats=true`, we send timings which are stats as a response. Please correct me if I am wrong.

juliusv:
The stats are a separate bag of metadata kept in https://github.com/prometheus/prometheus/blob/45fbed94d6ee17840254e78cfc421ab1db78f734/promql/engine.go#L145, not normal Prometheus metrics.

Currently those stats living in https://github.com/prometheus/prometheus/blob/master/util/stats/query_stats.go only support timers. The whole type would have to be extended / refactored / renamed to support simple non-duration stats fields as well.

Harkishen-Singh:
The output from the current implementation is

```
[hsingh@localhost ~]$ curl 127.0.0.1:9090/api/v1/query?query=up\&stats=true |python -m json.tool
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   708  100   708    0     0   345k      0 --:--:-- --:--:-- --:--:--  345k
{
    "status": "success",
    "data": {
        "resultType": "vector",
        "result": [
            {
                "metric": {
                    "__name__": "up",
                    "instance": "localhost:9191",
                    "job": "prometheus",
                    "my": "label",
                    "your": "label"
                },
                "value": [
                    1583082667.264,
                    "0"
                ]
            },
            {
                "metric": {
                    "__name__": "up",
                    "instance": "localhost:9100",
                    "job": "node_exporter"
                },
                "value": [
                    1583082667.264,
                    "0"
                ]
            },
            {
                "metric": {
                    "__name__": "up",
                    "instance": "localhost:9090",
                    "job": "prometheus",
                    "my": "label",
                    "your": "label"
                },
                "value": [
                    1583082667.264,
                    "1"
                ]
            }
        ],
        "stats": {
            "timings": {
                "evalTotalTime": 0.000169596,
                "resultSortTime": 0,
                "queryPreparationTime": 0.000111712,
                "innerEvalTime": 4.7639e-05,
                "execQueueTime": 1.4982e-05,
                "execTotalTime": 0.000195131,
                "involvedSamples": 0
            },
            "samples": {
                "totalSamples": 3,
                "peakSamples": 3,
                "evaluatorCount": 1
            }
        }
    }
}

```

promql/engine.go
```diff
@@ -601,6 +618,12 @@ func (ng *Engine) execEvalStmt(ctx context.Context, query *query, s *EvalStmt) (
 	}
 	evalSpanTimer.Finish()
 
+	if currentSamples = evaluator.currentSamples; currentSamples > query.sampleStats.PeakSamples {
```
roidelapluie: I would deal with that in the stats package; here I would just do query.sampleStats,ObserveSamples(evaluator.currentSamples)

and in querystat that would set the peak, total, and increase by 1 evaluator.

Harkishen-Singh: Sure. Will do this change.

Harkishen-Singh: Done. Please have a look @roidelapluie.

util/stats/query_stats.go
```diff
@@ -82,17 +82,28 @@ type queryTimings struct {
 	InnerEvalTime        float64 `json:"innerEvalTime"`
 	ExecQueueTime        float64 `json:"execQueueTime"`
 	ExecTotalTime        float64 `json:"execTotalTime"`
+	InvolvedSamples      int     `json:"involvedSamples"`
```
roidelapluie: not needed anymore

util/stats/query_stats.go
```diff
@@ -111,7 +122,15 @@ func NewQueryStats(tg *QueryTimers) *QueryStats {
...
-	qs := QueryStats{Timings: qt}
+	if ts != nil {
+		samples = querySamples{
+			EvaluatorCount: ts.EvaluatorCount,
+			PeakSamples:    ts.PeakSamples,
+			TotalSamples:   ts.TotalSamples,
+		}
+	}
+
+	qs := QueryStats{Timings: qt, Samples: samples}
```
roidelapluie:
```suggestion
	qs := QueryStats{Timings: qt, Samples: ts}
```

alwo why ts  as var name?

Harkishen-Singh: yeah it shouldn't be tha tway. I have now changed it to `sp` for `samples`

util/stats/query_stats.go
Harkishen-Singh: sure. removing it.

roidelapluie:
It does not work as I would expect:

up: 1/1/1
```
{
    "httpRequest": {
        "clientIP": "127.0.0.1",
        "method": "GET",
        "path": "/api/v1/query"
    },
    "params": {
        "end": "2020-03-01T20:43:09.723Z",
        "query": "up",
        "start": "2020-03-01T20:43:09.723Z",
        "step": 0
    },
    "stats": {
        "samples": {
            "evaluatorCount": 1,
            "peakSamples": 1,
            "totalSamples": 1
        },
        "timings": {
            "evalTotalTime": 2.4483e-05,
            "execQueueTime": 6.4855e-05,
            "execTotalTime": 9.6809e-05,
            "innerEvalTime": 1.0392e-05,
            "queryPreparationTime": 1.1118e-05,
            "resultSortTime": 0
        }
    },
    "ts": "2020-03-01T20:43:09.723Z"
}
```

`max_over_time(up[15m])`: 1/20/20
```
{
    "httpRequest": {
        "clientIP": "127.0.0.1",
        "method": "GET",
        "path": "/api/v1/query"
    },
    "params": {
        "end": "2020-03-01T20:43:43.904Z",
        "query": "max_over_time(up[15m])",
        "start": "2020-03-01T20:43:43.904Z",
        "step": 0
    },
    "stats": {
        "samples": {
            "evaluatorCount": 1,
            "peakSamples": 20,
            "totalSamples": 20
        },
        "timings": {
            "evalTotalTime": 2.9762e-05,
            "execQueueTime": 5.921e-06,
            "execTotalTime": 4.1347e-05,
            "innerEvalTime": 1.6499e-05,
            "queryPreparationTime": 1.0822e-05,
            "resultSortTime": 0
        }
    },
    "ts": "2020-03-01T20:43:43.905Z"
}
```

I would expect `up and max_over_time(up[15m])` to give 2/21/21 but it gives 1/1/1?
Maybe evaluatorCount is not what we want.

```
{
    "httpRequest": {
        "clientIP": "127.0.0.1",
        "method": "GET",
        "path": "/api/v1/query"
    },
    "params": {
        "end": "2020-03-01T20:43:37.265Z",
        "query": "up and max_over_time(up[15m])",
        "start": "2020-03-01T20:43:37.265Z",
        "step": 0
    },
    "stats": {
        "samples": {
            "evaluatorCount": 1,
            "peakSamples": 1,
            "totalSamples": 1
        },
        "timings": {
            "evalTotalTime": 3.6552e-05,
            "execQueueTime": 3.0731e-05,
            "execTotalTime": 7.1705e-05,
            "innerEvalTime": 2.2886e-05,
            "queryPreparationTime": 1.1477e-05,
            "resultSortTime": 0
        }
    },
    "ts": "2020-03-01T20:43:37.265Z"
}
```

brian-brazil:
> Maybe evaluatorCount is not what we want.

Yeah, that's a property of the query not the execution of it.

Harkishen-Singh:
@brian-brazil done. Please have a look.

util/stats/query_stats.go
```diff
@@ -148,10 +164,34 @@ type QueryTimers struct {
...
+	PeakSamples    int
+	TotalSamples   int
+}
+
+type Stats struct {
+	TimerStats  QueryTimers
+	SampleStats QuerySamples
+}
+
+// ObserveSamples observes the samples through which the query
```
brian-brazil: grammar is off here, queries don't pass through samples.

Harkishen-Singh:
@roidelapluie @brian-brazil can we have a second round of review on this feature now? I have made some changes in order to record the samples count during the evaluation as well.

@roidelapluie 
I did a query as said in the comment:
> up and max_over_time(up[15m])

and got the following result. Hope it serves the purpose.

```
[hsingh@localhost ~]$ curl 'localhost:9090/api/v1/query?query=up%20and%20max_over_time(up%5B15m%5D)&start=1587213158.27&end=1587216758.27&step=14&_=1587207960280&stats=true' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   561  100   561    0     0   547k      0 --:--:-- --:--:-- --:--:--  547k
{
  "status": "success",
  "data": {
    "resultType": "vector",
    "result": [
      {
        "metric": {
          "__name__": "up",
          "instance": "localhost:9090",
          "job": "prometheus",
          "my": "label",
          "your": "label"
        },
        "value": [
          1587217531.925,
          "1"
        ]
      },
      {
        "metric": {
          "__name__": "up",
          "instance": "localhost:9191",
          "job": "prometheus",
          "my": "label",
          "your": "label"
        },
        "value": [
          1587217531.925,
          "0"
        ]
      }
    ],
    "stats": {
      "timings": {
        "evalTotalTime": 7.3731e-05,
        "resultSortTime": 0,
        "queryPreparationTime": 1.7871e-05,
        "innerEvalTime": 5.1752e-05,
        "execQueueTime": 4.672e-05,
        "execTotalTime": 0.000129579
      },
      "samples": {
        "totalSamples": 72,
        "peakSamples": 4
      }
    }
  }
}

```

.gitignore
```diff
@@ -3,6 +3,7 @@
 /*.yaml
 /*.yml
 *.exe
+.idea
```
brian-brazil, line 6: This would better line in your personal gitrc.

promql/engine.go
```diff
@@ -129,6 +129,8 @@ type Query interface {
 	Statement() parser.Statement
 	// Stats returns statistics about the lifetime of the query.
 	Stats() *stats.QueryTimers
+	// SampleStats returns sample statistics about the lifetime of the query.
+	SampleStats() *stats.QuerySamples
```
brian-brazil, line 133: We should probably extend the above function rather than adding a new one.

promql/engine.go
```diff
@@ -925,6 +939,7 @@ func (ev *evaluator) rangeEval(f func([]parser.Value, *EvalNodeHelper) Vector, e
 							// past points at the next step.
 							matrixes[i][si].Points = series.Points[1:]
 							ev.currentSamples++
+							ev.updateSamples(ev.currentSamples - tempNumSamples)
```
brian-brazil, line 942: For calculating the peak shouldn't we always be passing in currentSamples, as that's what's in memory?

Harkishen-Singh: Well, the currentSamples count is already being maintained inside the `updateSamples` func, and hence passing currentSamples would be like `currentSamples * 2 + tempNumSamples`. Hence, did this way.

brian-brazil: This seems like you're more than double counting then, as you descend down the tree you'll be counting the already-computed LHS of parents as being processed again.

Harkishen-Singh: Well not sure on this, but each time we increase the currentSamples, it refers to the samples being increased in down the tree. that means, we need to save them in totalSamples as well and hence totalSamples are increased by the amount of addition to the currentSamples only.

Also, we need to capture the total samples involved in the evaluation of expr like binary which will involve combination of samples from more than one `currentSamples` tree. 

So in short, adding the total count of samples by the amount of increment done to the currentSamples should do the thing.

Harkishen-Singh: @brian-brazil any views?

brian-brazil: I'm not sure, and even if that were true the code seems fragile.

Harkishen-Singh: Sure, will try to have a second go through the code and try to dig more in a reliable way if possible.

promql/engine.go
Harkishen-Singh: Done. Extended the `Stats()`

Betula-L:
Just for information. May i ask that, is there any process now about this PR @Harkishen-Singh ?

Harkishen-Singh:
Yeah, the working is okay. Need to confirm and then an indepth review from brian. Will do in some days as stuck in another feature.

promql/engine.go
Harkishen-Singh: I think, that should be `ev.updateSamples(1)`. What do you say? @brian-brazil

Harkishen-Singh: Well, I went through this and realized that we need two inputs to the `updateSamples()`, one for the total samples, and the other for having instantaneous peak samples in order to capture the peak samples. I don't think that now, samples are being caught more than once, however still give a look at the implementation.

Harkishen-Singh:
@brian-brazil up for another review.

promql/engine.go
```diff
@@ -1332,6 +1348,7 @@ func (ev *evaluator) eval(expr parser.Expr) parser.Value {
 
 		res := newEv.eval(e.Expr)
 		ev.currentSamples = newEv.currentSamples
+		ev.updateSamples(ev.currentSamples, newEv.currentSamples)
```
brian-brazil, line 1351: Might these have already been counted inside the subquery?

Also, not related to this PR but are we losing any other samples that were already in memory prior to the subquery?

util/stats/query_stats.go
```diff
@@ -144,14 +162,45 @@ func (s *SpanTimer) Finish() {
...
+}
+
+type Stats struct {
+	TimerStats  QueryTimers
+	SampleStats QuerySamples
+}
+
+// UpdateStats updates the stats according to the samples
+// passed as params to the callback.
+func (qs *QuerySamples) UpdateStats() func(int, int) {
```
brian-brazil, line 187: The first argument is always current samples, could we not simplify a bit?

gouthamve:
Hi @Harkishen-Singh, we've looked at this PR in the bug scrub and were wondering if you have the bandwidth to pick this up again. I personally would love these stats and can help you pick this up if required, let me know!

Harkishen-Singh:
@gouthamve I am comfortable continuing on this PR. I got involved in #7779 from 2nd Jun. Now that it is mostly done, I can continue working on remaining open PRs. I will go through the discussion in this PR and fix as required soon.

Harkishen-Singh:
If we are concerned with the number of samples the query has to go through, why not have `NumSamples()` that returns the count of samples from the SeriesSet?

Will that be a good idea?

promql/engine.go
Harkishen-Singh: > Might these have already been counted inside the subquery?

Very true.

> Also, not related to this PR but are we losing any other samples that were already in memory prior to the subquery?

I don't think so. Since a subquery is just creating a new evaluator that handles evaluations and the `samplesStats` variable is a pointer so, if there were to be sample prior to the subquery, they would be already caught in that case.

Harkishen-Singh:
@brian-brazil up for review!

util/stats/query_stats.go
```diff
@@ -144,14 +162,45 @@ func (s *SpanTimer) Finish() {
...
+}
+
+type Stats struct {
+	TimerStats  QueryTimers
+	SampleStats QuerySamples
+}
+
+// Increment increments the samples count.
+func (qs *QuerySamples) Increment(samples int) {
+	if samples > qs.PeakSamples {
```
brian-brazil, line 190: This isn't right, as this is only how many samples we just processed - not how many are currently held in memory.

Harkishen-Singh: That is what the total samples is for: to catch and store the total count of samples while querying.

Could you explain in detail what you meant or expected?

brian-brazil: The PeakSamples should be the highest amount of samples we had in memory at once.

Harkishen-Singh: Got it. However, I wanted to know that does this issue aims to count the number of samples the query is considering while evaluation (which is similar to the number of points) or does this aims for the number of samples it had to scan (including the ones that were skipped).

brian-brazil: I'd do it as every input point processed. At the PromQL level, we don't know anything about skipped samples.

Harkishen-Singh: > I'd do it as every input point processed

Assuming this as when every point is being appended to the vector.

> PeakSamples should be the highest amount of samples we had in memory at once

I hope this means that number of points as whole added to a vector while evaluating a node.

brian-brazil: > Assuming this as when every point is being appended to the vector.

Before it's passed in I'd say. A rate 1d is more expensive to calculate than a 1h.

> I hope this means the total number of samples fetched from tsdb into memory. Is that correct?

That's part of it, it's also all the outputs that are currently in memory.

Harkishen-Singh: > That's part of it, it's also all the outputs that are currently in memory.

That means the vales of PeakSamples is monotonicaly increasing?

brian-brazil: It's a peak value, so it can't go down within a query.

Harkishen-Singh: Okay got it. Can you differentiate the between of `totalSamples` vs `peakSamples` as per what you think, so that it clears all my doubts. I will makes the changes right away.

brian-brazil: They're very different things. One is data processed, one is peak memory.

Harkishen-Singh: If that's the case, for **peak memory**, why not have a function in SeriesIterator as `SamplesCount()` that returns the number of samples across all the series?

brian-brazil: That's not peak memory, that's input series.

Harkishen-Singh: > input samples

Aren't the input samples the ones that contribute to memory usage?

brian-brazil: Yes, but they're generally a pretty small part of it as how we evaluate range vector functions is optimised so we only need to hold the range for one series in memory at a time. It's the intermediate results that are more the issue.

Harkishen-Singh:
@brian-brazil , since samples are fetched only in vector selectors, should peakSamples be calculated only as the sum of all the samples across all the vectorselectors? Otherwise, it will be a repeat calculation and the result might be wrong.

Harkishen-Singh:
@brian-brazil I have updated the logic as per my current understanding, keeping in mind to avoid repeated samples in peak sample calculations as well.

Harkishen-Singh:
The current implementation is on the following:
Total samples: They represent the number of samples involved in the calculation of the root AST node (only). 

Peak samples: Samples that are brought to memory from tsdb prior to performing any calculation. Since, fetching is only at vector selectors, it is the sum of samples of all vector selector nodes in any expression.

brian-brazil:
> They represent the number of samples involved in the calculation of the root AST node (only). 

That's not really of much use, the root node is generally the cheapest node to calculate. The useful number is how many samples all nodes touch.

> Peak samples: Samples that are brought to memory from tsdb prior to performing any calculation. 

That's not peak samples in memory, that's samples loaded from the tsdb. Peak samples is basically the highest value currentSamples hits.

Harkishen-Singh:
@brian-brazil updated as per what you mentioned as requirements.

brian-brazil:
Peak looks better, but you've missed a few other places where it needs to be checked.

The total is a mix of input and output, it should only be the inputs.

Harkishen-Singh:
@brian-brazil addressed!

Harkishen-Singh:
Does that look good? @brian-brazil

promql/engine.go
```diff
@@ -943,6 +954,7 @@ func (ev *evaluator) rangeEval(f func([]parser.Value, *EvalNodeHelper) Vector, e
 		enh.out = result[:0] // Reuse result vector.
 
 		ev.currentSamples += len(result)
+		ev.samplesStats.IncrementPeak(len(result))
```
brian-brazil, line 957: You're assuming that we're at the peak right now, which is not a good assumption. You need to check if currentSamples is actually the peak now, and if so update.

You should only ever be setting the peak, not incrementing it.

promql/engine.go
```diff
@@ -1332,6 +1350,9 @@ func (ev *evaluator) eval(expr parser.Expr) parser.Value {
 
 		res := newEv.eval(e.Expr)
 		ev.currentSamples = newEv.currentSamples
+		// Only update the peak samples if result evaluated by the new evaluator is
```
brian-brazil, line 1353: This comment doesn't seem necessary, as that's handled inside the function.

promql/engine.go
Harkishen-Singh: so, does that mean that peak is just the max number samples added to the result at an instant among many instances (and just catch those numbers and not increment as incrementing will be mostly same as total samples)?

promql/engine.go
Harkishen-Singh: Sure, will remove it.

promql/engine.go
brian-brazil: No, it's the peak value of currentSamples. Consider for example if a big aggregation set the peak to 1M, and we've now aggregated down to a single sample. No further processing will increase the samples in memory beyond the peak.

Harkishen-Singh: Updated. How does it look now?

brian-brazil: That's not it, the only number that should ever be considered for updating the peak is currentSamples as it's the peak value of currentSamples that we're trying to capture.

Harkishen-Singh: That's what is being considered, or the increase of it. Any other places you see please comment.

brian-brazil: You're not passing in currentSamples here, you're passing in something else. The biggest increase to currentSamples isn't interesting, it's the highest value it has that is interesting.

Harkishen-Singh:
@brian-brazil I have updated the peak to consider only the maximum value of currentSamples. I hope that is the requirement.

util/stats/query_stats.go
```diff
@@ -144,14 +162,52 @@ func (s *SpanTimer) Finish() {
 	}
 }
 
+type Statistics struct {
+	Timers  *QueryTimers
```
brian-brazil, line 166: These don't need to be references, which will save a bit of CPU for GC.

util/stats/query_stats.go
```diff
@@ -144,14 +162,52 @@ func (s *SpanTimer) Finish() {
...
 	*TimerGroup
 }
 
+type QuerySamples struct {
+	// PeakSamples represent the highest count of samples considered
+	// while evaluating a query. It corresponds to the peak value of
+	// currentSamples.
+	PeakSamples int
+	// TotalSamples represents the total number of samples scanned
+	// while evaluation a query.
```
brian-brazil, line 180: evaluating.

promql/engine.go
```diff
@@ -1498,17 +1521,20 @@ func (ev *evaluator) matrixIterSlice(it *storage.BufferedSeriesIterator, mint, m
 			}
 			out = append(out, Point{T: t, V: v})
 			ev.currentSamples++
+			ev.samplesStats.UpdatePeak(ev.currentSamples)
```
brian-brazil, line 1524: Can we do this outside the for? That'll save the conditional branch in the hot path.

This can be done in other places too.

promql/engine.go
```diff
@@ -1271,6 +1287,7 @@ func (ev *evaluator) eval(expr parser.Expr) parser.Value {
 		checkForSeriesSetExpansion(ev.ctx, e)
 		mat := make(Matrix, 0, len(e.Series))
 		it := storage.NewBuffer(durationMilliseconds(ev.lookbackDelta))
+		samplesCount := 0
```
brian-brazil, line 1290: This isn't used.

promql/engine.go
```diff
@@ -1488,6 +1510,7 @@ func (ev *evaluator) matrixIterSlice(it *storage.BufferedSeriesIterator, mint, m
 	buf := it.Buffer()
 	for buf.Next() {
 		t, v := buf.At()
+		ev.samplesStats.Increment(1)
```
brian-brazil, line 1513: This is going to under count, as a given sample might be processed in multiple steps.

You'll probably need to track this over around line 1150 instead.

util/stats/query_stats.go
Harkishen-Singh: Oh yes. I guess, this was done for some other purpose where structs were long, but over the due course of time, that was changed leaving this.

promql/engine.go
Harkishen-Singh: Yes, that will be better.

Harkishen-Singh: Also, do note that in some parts, the UpdatePeak is present inside the loop so that if any errors occur, then the information about peak samples is there as well.

Harkishen-Singh:
Done @brian-brazil .

promql/engine.go
Harkishen-Singh: Done.

promql/engine.go
```diff
@@ -919,12 +929,14 @@ func (ev *evaluator) rangeEval(f func([]parser.Value, *EvalNodeHelper) Vector, e
...
 				for _, point := range series.Points {
 					if point.T == ts {
+						ev.samplesStats.Increment(1)
 						if ev.currentSamples < ev.maxSamples {
 							vectors[i] = append(vectors[i], Sample{Metric: series.Metric, Point: point})
 							// Move input vectors forward so we don't have to re-scan the same
 							// past points at the next step.
 							matrixes[i][si].Points = series.Points[1:]
 							ev.currentSamples++
+							ev.samplesStats.UpdatePeak(ev.currentSamples)
```
brian-brazil, line 939: This is an inner loop, you can move this out. Also the stats increment.

promql/engine.go
```diff
@@ -1281,6 +1298,7 @@ func (ev *evaluator) eval(expr parser.Expr) parser.Value {
 			for ts := ev.startTimestamp; ts <= ev.endTimestamp; ts += ev.interval {
 				_, v, ok := ev.vectorSelectorSingle(it, e, ts)
 				if ok {
+					ev.samplesStats.Increment(1)
```
brian-brazil, line 1301: I think we're double counting here, as we'll also count these when they're an input to rangeEval.

This means that a query that is only a selector will have a total of 0, as no math was done.

promql/engine.go
Harkishen-Singh: Sure, assuming you meant outer by one-loop level.

promql/engine.go
Harkishen-Singh: Yes, we are double counting. That's what I said way before in comments.

Harkishen-Singh: That's why I said that total samples should be only the count in vector_selector and not in any other places since vector selectors are the only area that fetches the samples/series.

promql/engine.go
brian-brazil: The increment I think you can only really move out one, the other can be moved out two - but as long as we're avoiding the inner loop we should be okay. We should check the benchmarks don't show much difference when this is all done.

promql/engine.go
brian-brazil: That'd under count though, it's the number of samples that are passed into rate that determine performance more than the number of samples fetched. A rate(1h) with a 1m step is more expensive than a rate 1h with a 1h step over the same data.

Harkishen-Singh: > This means that a query that is only a selector will have a total of 0, as no math was done.

Only when I remove the current line.

Harkishen-Singh: Any suggestions here? @brian-brazil

Harkishen-Singh: >  A rate(1h) with a 1m step is more expensive than a rate 1h with a 1h step over the same data.

In that case, the current implementation does not seem wrong as those samples are being touched that are responsible for making queries expensive. We should consider again even though they have been considered during VS since they are being calculated again in the loops which increase the expense. Otherwise, if all samples other than the ones considered at VS are ignored, meaning, we should only consider the ones at VS which is pointless.

brian-brazil: That's still double counting, we shouldn't count them here - only when we do math on them subsequently.

Harkishen-Singh: Is there any other way in this then?

Harkishen-Singh: Ahh, I see you mean that not to count here as 0 is acceptable to us.

Harkishen-Singh: So, is 0 samples, in this case, is acceptable to us?

brian-brazil: Yeah.

We might add on the output samples at the very end just so it doesn't look too weird to users.

Harkishen-Singh: > We might add on the output samples at the very end just so it doesn't look too weird to users.

Assuming that will be in this PR, should I keep it as a different property as `outputSamples` alongside with `totalSamples` and `peakSamples`?

brian-brazil: I meant add it into totalSamples.

There's no need for an outputSamples, the user can already calculate that from the output.

Harkishen-Singh:
@brian-brazil done!

promql/engine.go
```diff
@@ -782,11 +791,14 @@ func (ev *evaluator) recover(ws *storage.Warnings, errp *error) {
...
 
-func (ev *evaluator) Eval(expr parser.Expr) (v parser.Value, ws storage.Warnings, err error) {
+func (ev *evaluator) Eval(expr parser.Expr) (val parser.Value, ws storage.Warnings, err error) {
 	defer ev.recover(&ws, &err)
 
-	v, ws = ev.eval(expr)
-	return v, ws, nil
+	val, ws = ev.eval(expr)
+	if mat, ok := val.(Matrix); ok {
+		ev.samplesStats.Increment(mat.TotalSamples())
```
brian-brazil, line 799: This deserves a comment, that's we're doing a single count of the overall output.

brian-brazil (commented):
Are there any docs to be updated? We should look at adding this to the UI, though not necessarily in this PR.

This should have some unittests. Expanding the existing sample limit tests should do the trick.

promql/engine.go
```diff
@@ -1503,6 +1526,7 @@ func (ev *evaluator) matrixIterSlice(it *storage.BufferedSeriesIterator, mint, m
 	// The seeked sample might also be in the range.
 	if ok {
 		t, v := it.Values()
+		ev.samplesStats.Increment(1)
```
brian-brazil, line 1529: This is still double-counting, and should be removed..

Harkishen-Singh:
> Are there any docs to be updated? We should look at adding this to the UI, though not necessarily in this PR.

Not sure but yes, I do think we are not mentioning anything about `stats` in querying in our docs.

Harkishen-Singh:
@brian-brazil done.

promql/engine.go
brian-brazil: This comment reads oddly. Something like Include output samples in overall samples touched would be clearer.

promql/engine_test.go
```diff
@@ -634,8 +648,10 @@ load 10s
 			Start: time.Unix(10, 0),
 		},
 		{
-			Query:      "metric[20s:5s]",
-			MaxSamples: 3,
+			Query:        "metric[20s:5s]",
+			MaxSamples:   3,
+			TotalSamples: 3,
```
brian-brazil, line 653: This feels low for a subquery, what does a more complex subquery produce?

To test this properly we should also add in some aggregators, binary operators, and functions on instant vectors.

Harkishen-Singh: Ran instant query (after a min of prometheus start) for the following:

`go_goroutines[5m:10s]` => 13 (verified via console)

`go_goroutines[5m:10s]` => 201 (verified in console)

Harkishen-Singh: @brian-brazil does it look satisfactory now?

promql/engine_test.go
```diff
@@ -647,6 +647,23 @@ load 10s
 			},
 			Start: time.Unix(10, 0),
 		},
+		{
+			Query:        "irate(metric[100s])",
```
brian-brazil, line 651: This duplicates the above test, we don't need a 2nd range vector function.

promql/engine_test.go
brian-brazil: There still needs to be a complex subquery.

promql/engine_test.go
```diff
@@ -739,8 +786,38 @@ load 10s
...
+				nil,
+				Matrix{Series{Metric: labels.Labels{}, Points: []Point{{T: 0, V: 1}, {T: 5000, V: 1}, {T: 10000, V: 2}}}},
+				nil,
+			},
+			Start:    time.Unix(0, 0),
+			End:      time.Unix(10, 0),
+			Interval: 5 * time.Second,
+		},
+		{
+			Query:        `count(metric)`,
```
brian-brazil, line 803: This duplicates the above test

promql/engine_test.go
```diff
@@ -739,8 +786,38 @@ load 10s
 			Interval: 5 * time.Second,
 		},
 		{
-			Query:      "rate(bigmetric[1s])",
-			MaxSamples: 1,
+			Query:        `sum(metric)`,
+			MaxSamples:   8,
+			TotalSamples: 9,
```
brian-brazil, line 791: This seems too high. It reads in 3 samples, and outputs 3 - so I'd expect 6.

promql/engine_test.go
```diff
@@ -750,6 +827,34 @@ load 10s
 			End:      time.Unix(10, 0),
 			Interval: 5 * time.Second,
 		},
+		{
+			Query:        "irate(bigmetric[1s])",
```
brian-brazil, line 831: This duplicates the above test

promql/engine_test.go
```diff
@@ -739,8 +786,38 @@ load 10s
...
+				nil,
+				Matrix{Series{Metric: labels.Labels{}, Points: []Point{{T: 0, V: 1}, {T: 5000, V: 1}, {T: 10000, V: 1}}}},
+				nil,
+			},
+			Start:    time.Unix(0, 0),
+			End:      time.Unix(10, 0),
+			Interval: 5 * time.Second,
+		},
+		{
+			Query:        "rate(bigmetric[1s])",
```
brian-brazil, line 817: A more realistic range so that this outputs something would be better

promql/engine_test.go
```diff
@@ -750,6 +827,34 @@ load 10s
...
+				nil,
+				Matrix{},
+				nil,
+			},
+			Start:    time.Unix(0, 0),
+			End:      time.Unix(10, 0),
+			Interval: 5 * time.Second,
+		},
+		{
+			Query:        `metric + metric`,
```
brian-brazil, line 845: Can we get a test for metric + 1 too?

promql/engine_test.go
```diff
@@ -750,6 +827,34 @@ load 10s
...
+			MaxSamples:   11,
+			TotalSamples: 15,
```
brian-brazil, line 847: I'd expect this to be 9

promql/engine_test.go
Harkishen-Singh: @brian-brazil any example for what you expect to be a complex subquery?

brian-brazil: Anything that's not just a selector, so a function or operator.

promql/engine_test.go
Harkishen-Singh: Bringing the Increment out from the inner loop (previous comments) in rangeEval, where we are checking the `ts` comparison is the culprit. We need to make sure to consider only those that match the timestamp else we are considering all the series. Thats the reason why we are getting 9 and not 6. Fixing this back to the inner loop.

promql/engine_test.go
Harkishen-Singh: Same culprit here too. Let's have the increment inside the `if point.T == ts` block in rangeEval as it was earlier. That gives the exact calculation.

Harkishen-Singh:
@brian-brazil done! Does the tests look satisfactory now?

promql/engine_test.go
brian-brazil: No, the issue was that we were using the wrong variable. vectors[i] is what we should have been taking the length of.

promql/engine_test.go
```diff
@@ -634,21 +648,25 @@ load 10s
 			Start: time.Unix(10, 0),
 		},
 		{
-			Query:      "metric[20s:5s]",
-			MaxSamples: 3,
+			Query:        "sum_over_time(metric[10s:1s])",
```
brian-brazil, line 651: This is still a simple subquery, we need to also have some complexity inside the subquery itself.

Harkishen-Singh: Is `sum(count_over_time(sum_over_time(metric[10s:1s])[10s:]))[10s:2s]` satisfactory?

brian-brazil: Yeah, we should probably test multiple levels of subqueries. Let's test the intermediates too, otherwise it'd be a bit much to mentally check all together.

Harkishen-Singh: Sorry, but could you tell more on this? `multiple levels of subqueries` and `intermediates`.

Do they mean break this query test into multiple small tests as well and join them again as tests so that we can easily understand the result?

brian-brazil: That is a 3-level subquery, so it has multiple levels. Each node on the AST is an intermediates.

Yes, one per level should do it.

Harkishen-Singh: Done! @brian-brazil

promql/engine_test.go
```diff
@@ -637,27 +637,71 @@ load 10s
...
+					Points: []Point{{T: 0, V: 1}, {T: 1000, V: 1}, {T: 2000, V: 1}, {T: 3000, V: 1}, {T: 4000, V: 1}, {T: 5000, V: 1}, {T: 6000, V: 1}, {T: 7000, V: 1}, {T: 8000, V: 1}, {T: 9000, V: 1}, {T: 10000, V: 2}},
+				}},
+				nil,
+			},
+			Start: time.Unix(10, 0),
+		},
+		{
+			Query:        "sum_over_time(metric[10s:1s])[10s:]",
+			MaxSamples:   25,
+			TotalSamples: 2,
```
brian-brazil, line 667: This is wrong, it can't be less than the TotalSamples for metric[10s:1s]

brian-brazil (commented):
Can you do a quick PromQL benchmark too, just in case there's a notable performance regression for some odd reason?

promql/engine_test.go
Harkishen-Singh: I think thats because in case of vector selectors, we only calculate during the output. But, here just after evaluating the VS, we go to subquery and then prom function, and hence the samples are selectively chosen based on timestamps, thus reducing the total samples. But, since peak is carried over, so it remains good.

Harkishen-Singh: So, should we consider adding the samples from `len(res.(Matrix))` (pseudo code) to `totalSamples` after evaluating a subquery?

brian-brazil: Yes, and ensure than any TotalSamples from that query is also included.

Harkishen-Singh:
@brian-brazil benchmarks upto 500 cases.

[prom_engine_samples_stats_bench.txt](https://github.com/prometheus/prometheus/files/5354206/prom_engine_samples_stats_bench.txt)

brian-brazil:
You'll want to do a benchcmp against the current head.

Harkishen-Singh:
@brian-brazil benchcmp results.

[cmp.txt](https://github.com/prometheus/prometheus/files/5354460/cmp.txt)

Harkishen-Singh:
Benchcmp avoiding the hotpath (avoiding innermost for in rangeEval by introducing a temp var).
[cmp_avoid_hot_path.txt](https://github.com/prometheus/prometheus/files/5354596/cmp_avoid_hot_path.txt)

Harkishen-Singh:
@brian-brazil which one is better?

Harkishen-Singh:
in hotpath vs without the hotpath in rangeEval
[cmp_two_diff_paths.txt](https://github.com/prometheus/prometheus/files/5354601/cmp_two_diff_paths.txt)

brian-brazil:
Those benchmarks are all showing severe performance changes, something isn't right here. If your machine has CPU throttling, make sure it's disabled when running the benchmarks.

Harkishen-Singh:
@brian-brazil I do think CPU throttling is enabled as I have dynamic scaling or frequency step multiplier active.

Harkishen-Singh:
@brian-brazil benchcmp after doing `cpupower frequency-set -g performance` along with pulling changes from master. Though, the frequency still varied between 3.0GHz to 3.8Ghz.

[cmp_performance_pull.txt](https://github.com/prometheus/prometheus/files/5355275/cmp_performance_pull.txt)

brian-brazil:
There's still some big swings there such as some queries being twice as expensive, which would not be acceptable for the sake of adding a few stats for debugging. Maybe try a different machine?

Harkishen-Singh:
>  Maybe try a different machine?

Is there any other option? That is not possible for me.

brian-brazil:
You'll need to get stable benchmark results on your machine then.

Harkishen-Singh:
@brian-brazil I tried to stabilize the CPU throttling and it worked. The results look much better than earlier. Are they satisfactory now?
[cmp_stable_cpu_throttling.txt](https://github.com/prometheus/prometheus/files/5356072/cmp_stable_cpu_throttling.txt)

brian-brazil:
Hmm, the increase around the histogram_quantile is a bit consistent to be noise.

Harkishen-Singh:
@brian-brazil yup! Working around that to make it more efficient.

brian-brazil:
It's probably just an artifact of the benchmark, but needs to be looked into anyway.

Harkishen-Singh:
@brian-brazil when I avoid the hotpath in rangeEval, I get better results now. The histogram_quantile noise seems to be solved now. Any comments? 
[cmp_stable_cpu_throttling_6.txt](https://github.com/prometheus/prometheus/files/5357086/cmp_stable_cpu_throttling_6.txt)

brian-brazil:
More noise than I'd like, but the absent_over_time is the only thing that stands out.

codesome:
Now that we have feature flags, since these statistics are slowing down queries, we can have it behind a feature flag for interested folks to turn this on. I am not sure what is our stance on this, do we move ahead with this behind the flag?

roidelapluie:
I wonder why that slows down queries - could it be that the benchmarks have some kind of issues?

Harkishen-Singh:
I guess throttling can be a case. I do remember disabling throttling while running benchmarks, but will try again.

codesome:
Query benchmarks do take a lot of time, can you try running only the subset of them that are too slow?

Harkishen-Singh:
Closing this as #10369 implements this.

PR #6891, opened 2020-02-27
tsdb: don't allow ingesting empty labelsets

roidelapluie:
See #5617

When we ingest an empty labelset in the head, further blocks can not be
compacted, with the error:

```
level=error ts=2020-02-27T21:26:58.379Z caller=db.go:659 component=tsdb
msg="compaction failed" err="persist head block: write compaction:
add series: out-of-order series added with label set \"{}\" / prev:
\"{}\""
```

We should therefore reject those invalid empty labelsets upfront.

This can be reproduced with the following:

```
cat << END > prometheus.yml
scrape_configs:
  - job_name: 'prometheus'
    scrape_interval: 1s
    basic_auth:
      username: test
      password: test
    metric_relabel_configs:
    - regex: ".*"
      action: labeldrop

    static_configs:
    - targets:
      - 127.0.1.1:9090
END
./prometheus --storage.tsdb.min-block-duration=1m
```
And wait a few minutes.

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -1154,6 +1154,11 @@ loop:
 				continue
 			}
 
+			if !lset.Has(labels.MetricName) {
+				app.Rollback()
+				return total, added, seriesAdded, errors.New("metric names are mandatory")
+			}
+
 			var ref uint64
 			ref, err = app.Add(lset, t, v)
 			switch err {
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -47,6 +47,8 @@ import (
 	"github.com/prometheus/prometheus/storage"
 )
 
+var errNameLabelMandatory = fmt.Errorf("missing metric name (%s label)", labels.MetricName)
+
 var (
 	targetIntervalLength = prometheus.NewSummaryVec(
 		prometheus.SummaryOpts{
@@ -1155,8 +1157,8 @@ loop:
 			}
 
 			if !lset.Has(labels.MetricName) {
-				app.Rollback()
-				return total, added, seriesAdded, errors.New("metric names are mandatory")
+				err = errNameLabelMandatory
+				break loop
 			}
 
 			var ref uint64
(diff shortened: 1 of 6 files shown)


brian-brazil:
Shouldn't we also check that there's at least a metric name when scraping?

roidelapluie:
This is an issue with metrics_relabeling. I don't know if we support empty
names or not.

Le ven. 28 févr. 2020 à 10:00, Brian Brazil <notifications@github.com> a
écrit :

brian-brazil:
All metrics have to have a metric name.

roidelapluie:
I will rewrite the PR. If we check for name we don't have to check for
emptiness. Note: labels are sorted, so __name will come in the first
labels, right? I can loop from the beginning.

Le ven. 28 févr. 2020 à 10:14, Brian Brazil <notifications@github.com> a
écrit :

brian-brazil:
The TSDB doesn't care about metric names though, it's agnostic to that. So we need checks in two places.

roidelapluie:
something like this --  test to be added in scrape package

scrape/scrape.go
```diff
@@ -1154,6 +1154,11 @@ loop:
 				continue
 			}
 
+			if !lset.Has(labels.MetricName) {
+				app.Rollback()
```
brian-brazil: This doesn't seem right, we should be treating this the same as a parse error.

roidelapluie:
But we need relabel to happen first

Le ven. 28 févr. 2020 à 11:31, Brian Brazil <notifications@github.com> a
écrit :

brian-brazil:
Yes, but the rest of the logic should be the same as a parse error.

scrape/scrape.go
```diff
@@ -1154,6 +1156,11 @@ loop:
 				continue
 			}
 
+			if !lset.Has(labels.MetricName) {
+				err = errNameLabelMandatory
+				break loop
```
roidelapluie: @brian-brazil this should be better

PR #6892, opened 2020-02-28
Fix new targets page not loading

### LINKED ISSUE #6732: React UI: Targets page doesn't work if there job name with special characters

csmarchbanks:
This fixes an issue where the /new/targets page will not load when there
are jobs with invalid CSS characters in them, such as the
namespace/service/0 form used by the Prometheus Operator.

Example console error for a job named `monitoring/alertmanager/0` such
as commonly added by the Prometheus operator.
```
DOMException: Failed to execute 'querySelectorAll' on 'Document':
'series-labels-monitoring\/alertmanager\/0-0' is not a valid selector.
```

Hopefully Fixes #6732 for real this time.

### DIFF
--- a/web/ui/react-app/src/pages/targets/TargetLabels.test.tsx
+++ b/web/ui/react-app/src/pages/targets/TargetLabels.test.tsx
@@ -25,7 +25,7 @@ describe('targetLabels', () => {
   it('renders a div of series labels', () => {
     const div = targetLabels.find('div').filterWhere(elem => elem.hasClass('series-labels-container'));
     expect(div).toHaveLength(1);
-    expect(div.prop('id')).toEqual('series-labels-cortex\\/node-exporter_group\\/0-1');
+    expect(div.prop('id')).toEqual('series-labels-cortex/node-exporter_group/0-1');
   });
 
   it('wraps each label in a label badge', () => {
--- a/web/ui/react-app/src/pages/targets/TargetLabels.tsx
+++ b/web/ui/react-app/src/pages/targets/TargetLabels.tsx
@@ -20,7 +20,7 @@ const TargetLabels: FC<TargetLabelsProps> = ({ discoveredLabels, labels, idx, sc
   const [tooltipOpen, setTooltipOpen] = useState(false);
 
   const toggle = (): void => setTooltipOpen(!tooltipOpen);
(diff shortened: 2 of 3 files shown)


juliusv:
@csmarchbanks Thanks, whoops!

The reason why I prefer full escaping over just replacing invalid characters with valid ones is that it leaves no possibility for a user to construct conflicting job names (after replacement - however unlikely).

It seems that all that is needed to fix the tooltip error is to escape the CSS id once more in the context of the `target` attribute? That is, I changed `target={id}` to `target={CSS.escape(id)}` and the page started working properly again. Or is there an issue with that approach that I'm overlooking?

csmarchbanks:
Hmm, yeah, I like that approach more. In that case should we un-escape `id` and only escape when passed to `target`?

juliusv:
@csmarchbanks Ah yeah, that's even better. I wasn't sure whether it was already a problem for the actual `id` attribute on each job panel to include these characters, but just escaping the selector seems to be enough.

csmarchbanks:
Uhhh... I did not click a single button and somehow this closed when I pushed a change? Github still having issues?

csmarchbanks:
Anyway, seems to work with `/` in the id. I thought I saw one issue where I got a similar error for `pool-{scrapePool}` but that was in the middle of me responding to an incident and I haven't seen it since. If I do find that issue again I will fix it :)

juliusv:
:+1: Thanks!

PR #6896, opened 2020-02-28
React UI: Fixed data table for matrix always showing 1 as value

skirsten:
The new React UI always displayed `1` as value if the DataTable with type matrix was displayed.
Example query: `prometheus_http_requests_total[5m]` and set the display mode to Table.

This fixes that.

@juliusv


juliusv:
Whooops! What a difference one letter can make... Thanks for catching this. :+1:

PR #6676, opened 2020-01-21
Make head Postings only return series in time range

roidelapluie:
Reworks and closes #6458

PR #6899, opened 2020-02-29
Do not attempt isolation for appendID == 0

beorn7:
Easy fix.

Now the memory usage on my production server is 7.7GB vs 6.5GB (`container_memory_working_set_bytes`), 18% increase. That's more what we expected. (Heap size is currently even smaller, but that might change after longer runtime.)

### DIFF
--- a/tsdb/head_test.go
+++ b/tsdb/head_test.go
@@ -1558,7 +1558,7 @@ func TestIsolationLowWatermarkMonotonous(t *testing.T) {
 	_, err = app1.Add(labels.FromStrings("foo", "bar"), 0, 0)
 	testutil.Ok(t, err)
 	testutil.Ok(t, app1.Commit())
-	testutil.Equals(t, uint64(1), hb.iso.lowWatermark())
+	testutil.Equals(t, uint64(1), hb.iso.lowWatermark(), "Low watermark should by 1 after 1st append.")
 
 	app1 = hb.Appender()
 	_, err = app1.Add(labels.FromStrings("foo", "bar"), 1, 1)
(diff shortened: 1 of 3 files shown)


beorn7:
https://github.com/prometheus/prometheus/pull/6841#issuecomment-592805023 for context.

brian-brazil:
:+1: 

Should we have a unittest for this?

beorn7:
Great joy: After 12 hours of running this, the server _with_ isolation takes _less_ memory, for both `container_memory_working_set_bytes` and `go_memstats_heap_alloc_bytes`. (However, I think the original prombench results from #6841 are valid because that instance never had to replay the WAL.)

I'll follow up with more results after the server has run for the weekend. I will also add unit tests. They are trivial, but that also means they are easy to add.

beorn7:
CPU usage is about 10% increased (1.1 vs. 1.0, so not much in absolute terms anyway). The server is rarely queried but has about 700 rules.

beorn7:
Tests added.

beorn7:
Resource usage after running the described relatively light-weight but somewhat rule-heavy production server for a day:
- Memory usage is essentially unaffected (I even see slightly lower memory usage right now, but that's just noise.)
- CPU usage +10% (but low in absolute terms: 1.1 cores instead of 1.0)
- Average rule evaluation time up from 11ms to 26ms. 99th percentile up from 140ms to 220ms.

I guess rule evaluation is more affected than the usual heavy range query for a dashboard because it almost exclusively deals with samples where the appendIDs have to be checked.

tsdb/head_test.go
```diff
@@ -1580,3 +1580,17 @@ func TestIsolationLowWatermarkMonotonous(t *testing.T) {
 	is.Close()
 	testutil.Equals(t, uint64(3), hb.iso.lowWatermark(), "After read has finished (iso state closed), low watermark should jump to three.")
 }
+
+func TestIsolationAppendIDZeroIsNoOp(t *testing.T) {
```
roidelapluie:
```suggestion
func TestIsolationAppendIDZeroIsNoop(t *testing.T) {
```

Like in /storage/remote/read_test.go we usually use `Noop`.

roidelapluie:
thanks.
don't forget that rules also write to tsdb, which also now has a small overhead and locks.

tsdb/head_test.go
beorn7: Done

beorn7:
> don't forget that rules also write to tsdb, which also now has a small overhead and locks.

Right, so it might not be so much query related.

roidelapluie:
> Great joy: After 12 hours of running this, the server _with_ isolation takes _less_ memory, for both `container_memory_working_set_bytes` and `go_memstats_heap_alloc_bytes`. (However, I think the original prombench results from #6841 are valid because that instance never had to replay the WAL.)
> 
> I'll follow up with more results after the server has run for the weekend. I will also add unit tests. They are trivial, but that also means they are easy to add.

Can you please tell us the version of master without isolation? Is it master minus isolation or 2.16.0? Not sure if that explains but we made some memory savings around Postings since 2.16.0.

beorn7:
That might explain it… In our production environment, I'm running v2.16.0 vs. this PR, while prombench was always tip of master vs. this PR.

roidelapluie:
Memory saving PR is #6676

PR #6901, opened 2020-02-29
[log] ServiceDiscovery: Add debug log when target has duplicate source

liguozhong:
Add debug log when target has duplicate source.

---- why?
We implement  ECS service discovery through the way of custom service discovery. Yesterday, our customers reported that there were 175 ECS, but we only found 14 ECSS.

In order to help customers solve the problem, we added logs for troubleshooting
----debug log code.
image
(EcsDiscoverer result log)
image
(Ecs Adapter result log)
----log
2020-02-28 03:20:52   - find ecs size : 175
2020-02-28 03:20:55   -  a.groups.len : 14
2020-02-28 03:20:55   -  a.arr.len : 14

----Why more than 131 ECCS are missing？
By debugging the Prometheus source code, we find that target will be discarded when the source is the same
image

---review our custom service discovery
image

Finally, by consulting our customers to confirm whether there is an ECS with the same InstanceName. 
Customer feedback: the name of ECS generated by k8s is the same。
----fix
Finally, we solve this problem by changing the value of source to instanceid
image

----Review of problems
We can't rule out that there may be duplicate instanceid in the future, or other Prometheus users may also encounter this problem, 
so it is recommended that when servicediscovery discards target, there is a log to remind our operation and maintenance personnel

### DIFF
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -298,6 +298,10 @@ func (m *Manager) updateGroup(poolKey poolKey, tgs []*targetgroup.Group) {
 			if _, ok := m.targets[poolKey]; !ok {
 				m.targets[poolKey] = make(map[string]*targetgroup.Group)
 			}
+			oldTg, ok := m.targets[poolKey][tg.Source]
+			if ok {
+				level.Debug(m.logger).Log("msg", "updateGroup Duplicate source", "source", tg.Source, "targets", tg.Targets, "oldTarget.targets", oldTg.Targets)
+			}
 			m.targets[poolKey][tg.Source] = tg
 		}
 	}


brian-brazil:
Can you explain more about how exactly this bug manifested? We don't have ECS discovery.

liguozhong:
ECS service discovery is implemented according to the customer service discovery document, not in the Prometheus source code. It's a bug in our own code.

Due to the bugs in the ECS service discovery function implemented by our team, the number of ECS is not correct. But when we look at the Prometheus log, there is no log related to root cause. So we suggest that when the source is repeated, there should be one more log

https://prometheus.io/blog/2018/07/05/implementing-custom-sd/
https://ecs.console.aliyun.com/
image

roidelapluie:
Are you using file_sd service discovery?

liguozhong:
> Are you using file_sd service discovery?

Yes, and yesterday's bug was our team's problem, not Prometheus's

roidelapluie:
What would you think about the lightweight https://github.com/prometheus/prometheus/pull/6902 ?

liguozhong:
> What would you think about the lightweight #6902 ?

This code seems to be helpful. Because this source repeatedly discards this line of target code, our customers almost complain about the service quality of our team.

roidelapluie:
> > What would you think about the lightweight #6902 ?
> 
> This code seems to be helpful. Because this source repeatedly discards target code, our customers almost complain about the service quality of our team.

I fail to see how that can happen to you as the source in file_sd is the source file + the position in the file. Can you provide an exemple of file that generates discards?

liguozhong:
> > > What would you think about the lightweight #6902 ?
> > 
> > 
> > This code seems to be helpful. Because this source repeatedly discards target code, our customers almost complain about the service quality of our team.
> 
> I fail to see how that can happen to you as the source in file_sd is the source file + the position in the file. Can you provide an exemple of file that generates discards?

ecs_discovery->service discovery manager->adapter->flie.  ->file discovery
175                   ->14.                                         ->14.       ->14   ->14

liguozhong:
> > > What would you think about the lightweight #6902 ?
> > 
> > 
> > This code seems to be helpful. Because this source repeatedly discards target code, our customers almost complain about the service quality of our team.
> 
> I fail to see how that can happen to you as the source in file_sd is the source file + the position in the file. Can you provide an exemple of file that generates discards?

Target is discarded in the manager.go's code. It's not a file_SD problem. It was discarded before the file was generated

roidelapluie:
Your code does not generate a message with the following POC:

prometheus.yml:

```
# my global config
global:

# Alertmanager configuration
alerting:
  alertmanagers:
  - static_configs:
    - targets:
      # - alertmanager:9093

rule_files:
   - "first_rules.yml"
  # - "second_rules.yml"

scrape_configs:
  - job_name: 'prometheus'

    # scheme defaults to 'http'.

    static_configs:
    - targets: ['localhost:9090']
    file_sd_configs:
      - files: ['a.json', 'b.json']
```

a.json AND b.json (same content)

```
[
   {
      "targets": [
         "127.0.0.1",
         "127.0.0.1"
      ]
   },
   {
      "targets": [
         "127.0.0.1",
         "127.0.0.1"
      ]
   }
]
```

In this case, prometheus will "silently drop" 7 / 8  targets because they are the same, but it does not happend in the code you have patched.

roidelapluie:
In my example, tg.Source will be: b.json:0, b.json:1, a.json:0, a.json:1

liguozhong:
image
(1 :174)
image
(2 :174->14)
image
(3 :14)
image
(4 :14)

liguozhong:
> In my example, tg.Source will be: b.json:0, b.json:1, a.json:0, a.json:1

It's not a file_SD bug. Before the file is generated, the target is lost. After the service discovery manager.go. The target of the same source is discarded

roidelapluie:
I do believe it is dropped https://github.com/prometheus/prometheus/blob/45fbed94d6ee17840254e78cfc421ab1db78f734/scrape/scrape.go#L405

roidelapluie:
oh wait -- do you mean that your file_sd implementation itself imports "github.com/prometheus/prometheus/discovery" ??

liguozhong:
> oh wait -- do you mean that your file_sd implementation itself imports "github.com/prometheus/prometheus/discovery" ??

No, file_SD is provided by Prometheus. Only the run() function is implemented

func  Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
}

liguozhong:
> oh wait -- do you mean that your file_sd implementation itself imports "github.com/prometheus/prometheus/discovery" ??

https://github.com/prometheus/prometheus/blob/master/documentation/examples/custom-sd/adapter/adapter.go

roidelapluie:
with file_sd, tg.source is unique per filename/target. So your patch is probably incorrect and unlikely to display anything. Or I miss something. Maybe we need more eyes here because I don't see how that can happen.

roidelapluie:
Thank you. Now it starts to make sense. So to my request:

> oh wait -- do you mean that your file_sd implementation itself imports "github.com/prometheus/prometheus/discovery" ??

the answer is yes.

I think the bug is more likely in https://github.com/prometheus/prometheus/blob/master/documentation/examples/custom-sd/adapter/adapter.go

liguozhong:
> with file_sd, tg.source is unique per filename/target. So your patch is probably incorrect and unlikely to display anything. Or I miss something. Maybe we need more eyes here because I don't see how that can happen.

thanks .
When you try to implement a custom service discovery, you will encounter the selection of the source field of the target. When you use a possible duplicate field as the source, you will encounter the problem that the target I mentioned is lost.

liguozhong:
> Thank you. Now it starts to make sense. So to my request:
> 
> > oh wait -- do you mean that your file_sd implementation itself imports "github.com/prometheus/prometheus/discovery" ??
> 
> the answer is yes.
> 
> I think the bug is more likely in https://github.com/prometheus/prometheus/blob/master/documentation/examples/custom-sd/adapter/adapter.go

yes. However, when a bug occurs, the core module of Prometheus does not print logs to give clues to solve the bug

roidelapluie:
We more likely want to adapt  https://github.com/prometheus/prometheus/blob/master/documentation/examples/custom-sd/adapter/adapter.go then.

liguozhong:
> Thank you. Now it starts to make sense. So to my request:
> 
> > oh wait -- do you mean that your file_sd implementation itself imports "github.com/prometheus/prometheus/discovery" ??
> 
> the answer is yes.
> 
> I think the bug is more likely in https://github.com/prometheus/prometheus/blob/master/documentation/examples/custom-sd/adapter/adapter.go

For the coder of custom service discovery, it's not easy to set a non repetitive source for the target.

liguozhong:
> We more likely want to adapt https://github.com/prometheus/prometheus/blob/master/documentation/examples/custom-sd/adapter/adapter.go then.

custom Service discovery needs to go through the process of discovery / manager.go. 
The target of the same source will be removed here, as long as the target has the same source.
 So we can't solve this problem by modify adapter.go

roidelapluie:
My opinion is that as this can not happen in prometheus itself, it is a pure 3rd party file sd issue ; which should be treated as such; and I don't think we need to take action in main prometheus code.

roidelapluie:
Thanks for the report and taking the time to explain all this.

liguozhong:
> Thanks for the report and taking the time to explain all this.

Thanks for your review

roidelapluie:
You should also know that our example will probably generate a HUGE binary because we import all the SD in prometheus/discovery. Maybe you can add this fix to a fork of you and remove all the imports ?

liguozhong:
> You should also know that our example will probably generate a HUGE binary because we import all the SD in prometheus/discovery. Maybe you can add this fix to a fork of you and remove all the imports ?

Our boss requires our R & D personnel to be consistent with the open source version of Prometheus, so I can't fock an independent version.

In addition, we have developed two service discovery through this custom approach. Strong dependence on this module.

An istio cluster for discovering more than 10000 host in private networks and an ECS cluster

liguozhong:
> Can you explain more about how exactly this bug manifested? We don't have ECS discovery.

Does this log make sense? Do I need to close this PR？

roidelapluie:
All you need to be prometheus/file_sd "compliant" is to produce files. We do not require you to use our example as is or even to use golang.

I would suggest that you close this PR as it does not look like this can happen in Prometheus itself and it could be tested easily in the file_sd implementation logic itself.

Thanks for your work and all your explanations.

liguozhong:
> All you need to be prometheus/file_sd "compliant" is to produce files. We do not require you to use our example as is or even to use golang.
> 
> I would suggest that you close this PR as it does not look like this can happen in Prometheus itself and it could be tested easily in the file_sd implementation logic itself.
> 
> Thanks for your work and all your explanations.

ok

PR #6905, opened 2020-03-01
Merge two duplicate constants of labels.go

liguozhong:
Merge two duplicate constants of labels.go


liguozhong:
Both labels.go have the same constant


image

image

brian-brazil:
I'm trying to get rid of model for simple constants like this, please don't spread it further.

liguozhong:
> I'm trying to get rid of model for simple constants like this, please don't spread it further.

ok

PR #6907, opened 2020-03-01
Fix race condition in Rule manager.Update() function

liguozhong:
Fix race condition in Rule manager.Update() function
image

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -921,7 +921,7 @@ func (m *Manager) Update(interval time.Duration, files []string, externalLabels 
 		}
 
 		wg.Add(1)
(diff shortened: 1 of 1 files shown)


roidelapluie:
Thanks!

roidelapluie:
That would mean that we could pass newg to the inner goroutine as well

liguozhong:
> Thanks!

thanks

roidelapluie:
Now that I look deeper I don't see a race here as the olg id scoped to the inner loop.

roidelapluie:
> That would mean that we could pass newg to the inner goroutine as well

That comment is invalid. We don't need to.

PR #6908, opened 2020-03-01
Fix race condition in notifier/Manager.sendAll() function

liguozhong:
Fix race condition in notifier/Manager.sendAll() function.
image

### DIFF
--- a/notifier/notifier.go
+++ b/notifier/notifier.go
@@ -518,7 +518,7 @@ func (n *Manager) sendAll(alerts ...*Alert) bool {
 
 			ctx, cancel := context.WithTimeout(n.ctx, time.Duration(ams.cfg.Timeout))
 
-			go func(ctx context.Context, client *http.Client, url string) {
+			go func(ctx context.Context, cancel context.CancelFunc, client *http.Client, url string) {
 				defer cancel()
 				if err := n.sendOne(ctx, client, url, payload); err != nil {
 					level.Error(n.logger).Log("alertmanager", url, "count", len(alerts), "msg", "Error sending alert", "err", err)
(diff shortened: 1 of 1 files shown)


roidelapluie:
If I am right the cancel should be in the goroutine, too

liguozhong:
> If I am right the cancel should be in the goroutine, too

done,thanks.

image

roidelapluie:
I think we can create the with timeout in the goroutine so we don't need to pass cancel() ?

liguozhong:
> I think we can create the with timeout in the goroutine so we don't need to pass cancel() ?

done,I agree with that
image

notifier/notifier.go
```diff
@@ -530,7 +530,7 @@ func (n *Manager) sendAll(alerts ...*Alert) bool {
 				n.metrics.sent.WithLabelValues(url).Add(float64(len(alerts)))
 
 				wg.Done()
-			}(ams.client, am.url().String())
+			}(ctx, cancel, ams.client, am.url().String())
```
roidelapluie: I do not think there is a race with ctx and cancel. Why do you think so? They are scoped to the inner loop.

roidelapluie:
So this PR should only move cancel() in the go routine.

liguozhong:
> So this PR should only move cancel() in the go routine.

yes,done.

liguozhong:
> So this PR should only move cancel() in the go routine.
> 
> yes,done.

image

notifier/notifier.go
codesome: ^ same. Even in https://github.com/prometheus/prometheus/pull/6907

PR #6911, opened 2020-03-01
scrape/target_test.go: remove deprecated function BuildNameToCertificate()

mjtrangoni:
This fix a staticcheck issue related to go1.14 function deprecation.
See the change here: https://github.com/golang/go/commit/eb93c684d40de4924fc0664d7d9e98a84d5a100b

See the error message,
```
$ go version
go version go1.14 linux/amd64
$ make lint
>> running golangci-lint
GO111MODULE=on go list -e -compiled -test=true -export=false -deps=true -find=false -tags= -- ./... > /dev/null
GO111MODULE=on /home/mt/go/packages/bin/golangci-lint run  ./...
scrape/target_test.go:260:2: SA1019: tlsConfig.BuildNameToCertificate is deprecated: NameToCertificate only allows associating a single certificate with a given name. Leave that field nil to let the library select the first compatible chain from Certificates.  (staticcheck)
	tlsConfig.BuildNameToCertificate()
	^
scrape/target_test.go:357:2: SA1019: tlsConfig.BuildNameToCertificate is deprecated: NameToCertificate only allows associating a single certificate with a given name. Leave that field nil to let the library select the first compatible chain from Certificates.  (staticcheck)
	tlsConfig.BuildNameToCertificate()
	^
make: *** [Makefile.common:181: common-lint] Error 1
```
Signed-off-by: Mario Trangoni <mjtrangoni@gmail.com>

### DIFF
--- a/scrape/target_test.go
+++ b/scrape/target_test.go
@@ -257,7 +257,6 @@ func TestNewHTTPClientCert(t *testing.T) {
 	tlsConfig := newTLSConfig("server", t)
 	tlsConfig.ClientAuth = tls.RequireAndVerifyClientCert
 	tlsConfig.ClientCAs = tlsConfig.RootCAs
-	tlsConfig.BuildNameToCertificate()
 	server.TLS = tlsConfig
 	server.StartTLS()
 	defer server.Close()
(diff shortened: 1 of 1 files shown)


roidelapluie:
We are not yet in go1.14 because of an upstream bug

mjtrangoni:
@roidelapluie  Thanks for the info, which is that bug?

roidelapluie:
I suggest you follow https://github.com/prometheus/golang-builder/pull/85

simonpasquier:
:+1: It doesn't hurt to merge this PR irrespective of whether the project uses Go 1.13 or 1.14.

roidelapluie:
Thanks!

PR #6913, opened 2020-03-01
[config] scrape：Add "scrape.reload.time" to the startup parameter

liguozhong:
[config] scrape：Add "scrape.reload.time" to the startup parameter

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -120,7 +120,8 @@ func main() {
 		prometheusURL   string
 		corsRegexString string
 
-		promlogConfig promlog.Config
+		promlogConfig    promlog.Config
+		scrapeReloadTime model.Duration
 	}{
 		notifier: notifier.Options{
 			Registerer: prometheus.DefaultRegisterer,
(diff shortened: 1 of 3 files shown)


roidelapluie:
This is totally internal and should not be changed or changeable. What is your use case?

liguozhong:
> This is totally internal and should not be changed or changeable. What is your use case?

       We collect more than 10000+ machines, and the acquisition target will constantly change. We attach great importance to the stability of the scraper.
       We do not want the scraper to be reloaded all the time, which will lead to the blank in grafana, or make the alarm false alarm, so we hope that the frequency of the scraper will be lower.
        A 10 minute update would make sense for us. The monitoring system developed by our company is also updated every 10 minutes
       However, changing the default reload frequency of Prometheus obviously does not meet the scenario of other customers, so we hope reload frequency is an option

brian-brazil:
Your use case is I believe using file_sd, if you only have updates every 10m then don't update the file more often than that.

This 5s is a sanity throttling, in case some discovery started spewing large amounts of target updates. It shouldn't be configurable.

liguozhong:
> Your use case is I believe using file_sd, if you only have updates every 10m then don't update the file more often than that.
> 
> This 5s is a sanity throttling, in case some discovery started spewing large amounts of target updates. It shouldn't be configurable.

OK, I got it. I'll  close this PR first.
 This is a difference I found when comparing the internal monitoring system with Prometheus.
 The main interior is 10 minutes, Prometheus is 5 seconds.

roidelapluie:
> > Your use case is I believe using file_sd, if you only have updates every 10m then don't update the file more often than that.
> > This 5s is a sanity throttling, in case some discovery started spewing large amounts of target updates. It shouldn't be configurable.
> 
> OK, I got it. I'll close this PR first.
> This is a difference I found when comparing the internal monitoring system with Prometheus.
> The main interior is 10 minutes, Prometheus is 5 seconds.

It is not like that. Various SD's have configurable sync intervals. The 5s is not for this use case.

PR #6914, opened 2020-03-01
Expose EC2 instance lifecycle as label

alexgaganov-fiverr:
Signed-off-by: Alex Gaganov <alex.gaganov@fiverr.com>

This change exposes the `InstanceLifecycle` attribute during EC2 service discovery and makes it available as a `__meta_ec2_instance_lifecycle` label. The attribute's value can be either `spot` or `scheduled`. In the case of "regular" instance, the attribute will be `nil`.
The new label will come in handy when there's a need to distinguish between `spot` and `on-demand` instances.

There was an attempt to address this in #6017, however, the PR has some pending comments and got stale due to inactivity :-(  

@brian-brazil your input would be greatly appreciated.

### DIFF
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -428,6 +428,7 @@ The following meta labels are available on targets during [relabeling](#relabel_
 
 * `__meta_ec2_availability_zone`: the availability zone in which the instance is running
 * `__meta_ec2_instance_id`: the EC2 instance ID
+* `__meta_ec2_instance_lifecycle`: the lifecycle of the EC2 instance, set only for 'spot' or 'scheduled' instances, absent otherwise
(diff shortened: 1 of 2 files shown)


docs/configuration/configuration.md
```diff
@@ -430,6 +430,7 @@ The following meta labels are available on targets during [relabeling](#relabel_
 * `__meta_ec2_instance_id`: the EC2 instance ID
 * `__meta_ec2_instance_state`: the state of the EC2 instance
 * `__meta_ec2_instance_type`: the type of the EC2 instance
+* `__meta_ec2_instance_lifecycle`: the lifecycle of the EC2 instance, set only for 'spot' or 'scheduled' instances, absent otherwise
```
brian-brazil: Keep it sorted

brian-brazil:
Thanks!

PR #6915, opened 2020-03-01
Revert "Fix race condition in Rule manager.Update() function"

roidelapluie:
This reverts commit 8b11d2cfb6f9a7ae37bd402ca5176da034c572f8.

oldg is scoped to the inner loop so there is no race here.

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -921,7 +921,7 @@ func (m *Manager) Update(interval time.Duration, files []string, externalLabels 
 		}
 
 		wg.Add(1)
-		go func(newg *Group, oldg *Group) {
+		go func(newg *Group) {
 			if ok {
 				oldg.stop()
 				newg.CopyState(oldg)
@@ -934,7 +934,7 @@ func (m *Manager) Update(interval time.Duration, files []string, externalLabels 
 				newg.run(m.opts.Context)
 			}()
 			wg.Done()
-		}(newg, oldg)
+		}(newg)
 	}
 
 	// Stop remaining old groups.


roidelapluie:
cc @liguozhong

liguozhong:
> cc @liguozhong

thanks，Here's my opinion. I don't know if it's right. Can you help me review it again？

This for loop causes "oldg" to be reassigned next time. In this way, before goroutine starts, the old "oldg" will be replaced next time. As a result, the "oldg" that should have executed stop is not executed. In addition, it is found that the "OK" variable also needs to be transferred to the parameter

image

liguozhong:
> cc @liguozhong

I'm very sorry. My time zone here is UTC + 8，What's your time zone? I can adjust the time next time I raise PR

roidelapluie:
> > cc @liguozhong
> 
> I'm very sorry. My time zone here is UTC + 8，What's your time zone? I can adjust the time next time I raise PR

We work asynchronously. I see no issue with responding at your best convenience. Please don't change anything on your side.

roidelapluie:
> > cc @liguozhong
> 
> thanks，Here's my opinion. I don't know if it's right. Can you help me review it again？
> 
> This for loop causes "oldg" to be reassigned next time. In this way, before goroutine starts, the old "oldg" will be replaced next time. As a result, the "oldg" that should have executed stop is not executed. In addition, it is found that the "OK" variable also needs to be transferred to the parameter
> 
> image

The variable is not changed, it is at every call a new one: `:=`. So this is safe.

liguozhong:
> 
> The variable is not changed, it is at every call a new one: `:=`. So this is safe.

👍，I tested it locally, and ": =" is new and independent in every call

image

roidelapluie:
Thanks for double checking

Issue #6909, opened 2020-03-01
Isolation followup


roidelapluie:
Some things I would like to clarify with isolation:

- [x] Verify across code that we always rollback or commit
- [x] Verify that appenders and queriers are as short-lived as possible
- [x] Verify that everywhere we create a head (e.g. in snapshots) it does not blow up memory

roidelapluie:
cc @beorn7

beorn7:
> Verify across code that we always rollback or commit

Note that there is `prometheus_tsdb_head_active_appenders`, which gets incremented on appender creation and decremented on rollback/commit. If we miss the rollback or commit, it shows up in this metrics not going back to zero (and with the new isolation code in place, it will show up as a stuck low watermark). I have seen that very occasionally in production. Those might have been Prometheus servers that were in some kind of distress anyway. But it's definitely worth a look how this can possibly happen.

> Verify that appenders and queriers are as short-lived as possible

That was desirable even before. @brian-brazil @codesome @gouthamve @krasi-georgiev as the TSDB experts, perhaps you have context for that?

> Verify that everywhere we create a head (e.g. in snapshots) it does not blow up memory

I cannot see how, but then again the experts for TSDB certainly know more than I do. Paging @brian-brazil @codesome @gouthamve @krasi-georgiev again.

roidelapluie:
> > Verify that everywhere we create a head (e.g. in snapshots) it does not blow up memory
> 
> I cannot see how, but then again the experts for TSDB certainly know more than I do. Paging @brian-brazil @codesome @gouthamve @krasi-georgiev again.

I guess it depends what we do with the head. subtle issues like #6899 could happen.

roidelapluie:
> * [x]  Verify across code that we always rollback or commit
 I have checked this

beorn7:
> I guess it depends what we do with the head. subtle issues like #6899 could happen.

That wasn't exactly subtle. And it's fixed now.

roidelapluie:
> > I guess it depends what we do with the head. subtle issues like #6899 could happen.
> 
> That wasn't exactly subtle. And it's fixed now.

I would be happy if you just take a snapshot with head of your test prometheus and confirm it's allright :)

brian-brazil:
> as the TSDB experts, perhaps you have context for that?

Appenders I'm not worried about. For queriers it's the 2m query timeout that ultimately matters, and tweaking querier lifetimes is like at most a microoptimisation.

[...]

PR #6918, opened 2020-03-02
Appending to TSDB: Move crucial actions to defer

beorn7:
With defer having less of a performance penalty, there is no reason
not to do those crucial operations via defer.

Context: With isolation in place, if we forget to Commit/Rollback, the
low watermark will get stuck forever.

The current code should not have any bugs, but moving to defer helps
to avoid future bugs.

This is also moving the `closeAppend` in the `Commit` implementation
itself to defer. If logging to the WAL fails, we would have missed the
`closeAppend`.

Related to #6909 .

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -1069,6 +1069,19 @@ func (sl *scrapeLoop) append(b []byte, contentType string, ts time.Time) (total,
 	)
 	var sampleLimitErr error
 
+	defer func() {
+		if err != nil {
+			app.Rollback()
+			return
+		}
+		if err = app.Commit(); err != nil {
+			return
+		}
+		// Only perform cache cleaning if the scrape was not empty.
+		// An empty scrape (usually) is used to indicate a failed scrape.
+		sl.cache.iterDone(len(b) > 0)
+	}()
+
 loop:
 	for {
 		var et textparse.Entry
@@ -1229,19 +1242,7 @@ loop:
 			return err == nil
 		})
 	}
-	if err != nil {
-		app.Rollback()
-		return total, added, seriesAdded, err
-	}
-	if err := app.Commit(); err != nil {
-		return total, added, seriesAdded, err
-	}
-
-	// Only perform cache cleaning if the scrape was not empty.
-	// An empty scrape (usually) is used to indicate a failed scrape.
-	sl.cache.iterDone(len(b) > 0)
-
-	return total, added, seriesAdded, nil
+	return
 }
 
 func yoloString(b []byte) string {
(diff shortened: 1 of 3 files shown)


roidelapluie:
I put my release shepherd veto on this pull request until we use go1.14, which we are not at this moment.

beorn7:
Yes, sure.

For context: https://github.com/prometheus/golang-builder/pull/85 describes what is holding us back.

roidelapluie:
> Yes, sure.
> 
> For context: [prometheus/golang-builder#85](https://github.com/prometheus/golang-builder/pull/85) describes what is holding us back.

Yes, sorry, github does not send updated comments by email.

scrape/scrape.go
```diff
@@ -1258,67 +1259,71 @@ const (
...
+			return
+		}
+		err = app.Commit()
+	}()
 
-	if err := sl.addReportSample(app, scrapeHealthMetricName, ts, health); err != nil {
-		app.Rollback()
-		return err
+	if err = sl.addReportSample(app, scrapeHealthMetricName, ts, health); err != nil {
+		return
```
brian-brazil: It's a little weird to see returning without the err, could we make this explicit?

tsdb/head.go
```diff
@@ -1068,13 +1069,14 @@ func (a *headAppender) Commit() error {
...
 	a.head.metrics.samplesAppended.Add(float64(total))
 	a.head.updateMinMaxTime(a.mint, a.maxt)
-	a.head.iso.closeAppend(a.appendID)
 
 	return nil
 }
 
 func (a *headAppender) Rollback() error {
-	a.head.metrics.activeAppenders.Dec()
+	defer a.head.metrics.activeAppenders.Dec()
```
brian-brazil: Why defer rather than immediately?

scrape/scrape.go
```diff
@@ -1258,67 +1259,71 @@ const (
...
 
 	var health float64
-	if err == nil {
+	if scrapeErr == nil {
 		health = 1
 	}
 	app := sl.appender()
+	defer func() {
+		if err != nil {
+			app.Rollback()
```
roidelapluie: while we are at it should we use multierr ?

scrape/scrape.go
beorn7: While I avoid naked returns in general, in this particular case it is a legitimate pattern (inspecting and changing the returned error in `defer`). And In that case, I'd like to embrace the naked return completely and mark it clearly by, well, a naked return.

beorn7: Having said that, if we have a style convention to still put the names in the return statement, I'm happy to do so, but I'm not aware of any.

tsdb/head.go
beorn7: To be consistent with what we do in `Commit`. I also think it is clearer to do it in the end in this way. Makes debugging easier if for example one of the locks cannot get acquired.

scrape/scrape.go
beorn7: I'd like to keep unrelated changes out of this PR.

beorn7: Or perhaps not completely unrelated, depending on how you see it.

But still, let's do one (non-trivial) thing per PR.

beorn7:
I have rebased this on top of the release-2.17 branch and changed the base.

Ironically, this doesn't really help with the panic in #6977 but I think, now that we have seen panics getting recovered because they are in the HTTP serving path, we should be conservative and merge this into release-2.17.

I'll keep checking the querying path for similar requirements to move to defer.

roidelapluie:
Can we reach consensus regarding the explicit returns discussion? @brian-brazil @beorn7

roidelapluie:
> Ironically, this doesn't really help with the panic in #6977

Can we do something for that kind of panics?

beorn7:
> Can we do something for that kind of panics?

Done in #6983 . (o:

brian-brazil:
> Can we reach consensus regarding the explicit returns discussion?

I'm not that pushed on it.

roidelapluie:
Thanks!

PR #6923, opened 2020-03-03
Fix computeChunkEndTime

codesome:
```go
func computeChunkEndTime(start, cur, max int64) int64 {
	a := (max - start) / ((cur - start + 1) * 4)
	if a == 0 {
		return max
	}
	return start + (max-start)/a
}
```
Currently, as `a` is an int, we return `max` when `a` it upto `1.999...` (when considered to be float, because `(max-start)/a = (max-start)` as `int(1.999...) = 1`). This should not be the case. Also `start + (max-start)/a` looks tricky because of division by `a`. So I have simplified it too.

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -1907,11 +1907,11 @@ func (s *memSeries) cleanupAppendIDsBelow(bound uint64) {
 // chunk, its current timestamp and the upper bound up to which we insert data.
 // It assumes that the time range is 1/4 full.
 func computeChunkEndTime(start, cur, max int64) int64 {
-	a := (max - start) / ((cur - start + 1) * 4)
-	if a == 0 {
+	expTimeRange := ((cur - start + 1) * 4) // For a full chunk.
+	if (max - start) < expTimeRange {
 		return max
 	}
-	return start + (max-start)/a
+	return start + expTimeRange
 }
 
 func (s *memSeries) iterator(id int, isoState *isolationState, it chunkenc.Iterator) chunkenc.Iterator {


roidelapluie:
Are there there implications for end users? Which versions have the bug? all 2.0?

codesome:
This is not actually a user-facing bug, so no implications. And I _guess_ this was since 2.0. It is just that chunks might end up being larger than 120 samples currently (which is fine).

roidelapluie:
> Ok, what about quick test, showing regression? Otherwise LGTM (:
> 
> And thanks for simplifying!

Just fixing the tests seem enough :)

gouthamve:
@codesome poke :)

codesome:
`TestMemSeries_append` is behaving weirdly with this change (while `TestComputeChunkEndTime` can be fixed as expected from this change). I will spend some more time soon on this.

tsdb/head.go
```diff
@@ -1907,11 +1907,11 @@ func (s *memSeries) cleanupAppendIDsBelow(bound uint64) {
 // chunk, its current timestamp and the upper bound up to which we insert data.
 // It assumes that the time range is 1/4 full.
 func computeChunkEndTime(start, cur, max int64) int64 {
-	a := (max - start) / ((cur - start + 1) * 4)
-	if a == 0 {
+	expTimeRange := ((cur - start + 1) * 4) // For a full chunk.
```
bwplotka, line 1910: No need for extra parenthesis 
```suggestion
	expTimeRange := (cur - start + 1) * 4 // For a full chunk.
```

bwplotka (approved):
Makes sense to me, but let's investigate the tests

codesome:
Closing this since I don't think I will be able to investigate the tests and fix them. We can always come back if this is causing any problems.

PR #6928, opened 2020-03-04
[test] add Goroutine Leak Detector for fanout Querier.SelectSorted

liguozhong:
[test] add Goroutine Leak Detector for fanout Querier.SelectSorted


liguozhong:
image
image

liguozhong:
> Awesome!
> 
> However the idea was rather different. I think most of our existing tests can be additionally checked using leaktest (without changing their test names!). Let's go through all _existing tests_ that use components with goroutines and try to put `defer leaktest.CheckTimeout(t, 10*time.Second)()` as **a first statement in test** which will:
> 
> 1. Get snasphot of go routines now.
> 2. Allow test execution
> 3. ON defer get snapshot of all go routines and compare. If there is a diff for longer than 10s it fails the test...
> 
> What do you think? 🤗

done，Ci cannot pass because a go mod dependency has been added. I'm stuck with this problem. Can you help me add this dependency correctly


ci/circleci: test — Your tests failed on CircleCI.

roidelapluie:
Thank you! This has been implemented in #7566.

PR #6930, opened 2020-03-04
Separate scrape add error checking out into it's own function.

cstyan:
While working on the exemplars implementation I wanted a simpler way to know whether adding a sample had succeeded to be able to simplify the conditional check for adding an exemplar. Rather than making a change as part of that work, I'm opening this PR to refactor the error checking for the calls to `Appender`'s `add/addFast` functions.

I think the behaviour here remains the same, with the addition of a conditional check if the error was nil for the no cache entry/create new series path, but in that case `checkAddError` will be passed a nil pointer to a `cacheEntry` and so again the behaviour should not have changed.

I went with two bools and an error as the return values but we could also pass `sampleLimitError` in.

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -1072,7 +1072,6 @@ func (sl *scrapeLoop) append(b []byte, contentType string, ts time.Time) (total,
 		appErrs        = appendErrors{}
 		sampleLimitErr error
 		sampleAdded    bool
-		breakLoop      bool
 	)
 
 	defer func() {
@@ -1129,8 +1128,13 @@ loop:
 
 		if ok {
 			err = app.AddFast(ce.ref, t, v)
-			sampleAdded, breakLoop = sl.checkAddError(ce, met, tp, err, &sampleLimitErr, appErrs)
-		} else {
+			sampleAdded, err = sl.checkAddError(ce, met, tp, err, &sampleLimitErr, appErrs)
+			// In theory this should never happen.
+			if err == storage.ErrNotFound {
+				ok = false
+			}
+		}
+		if !ok {
 			var lset labels.Labels
 
 			mets := p.Metric(&lset)
@@ -1154,10 +1158,14 @@ loop:
 			var ref uint64
 			ref, err = app.Add(lset, t, v)
 			// call new check error function here
-			sampleAdded, breakLoop = sl.checkAddError(nil, met, tp, err, &sampleLimitErr, appErrs)
-			if breakLoop {
+			sampleAdded, err = sl.checkAddError(nil, met, tp, err, &sampleLimitErr, appErrs)
+			if err != nil {
+				if err != storage.ErrNotFound {
+					level.Debug(sl.l).Log("msg", "unexpected error", "series", string(met), "err", err)
+				}
 				break loop
 			}
+
 			if tp == nil {
 				// Bypass staleness logic if there is an explicit timestamp.
 				sl.cache.trackStaleness(hash, lset)
@@ -1166,10 +1174,6 @@ loop:
 			if sampleAdded && sampleLimitErr == nil {
 				seriesAdded++
 			}
-		}
-
-		if breakLoop {
-			break loop
 		}
 
 		// Match the previous behaviour, increment added even if there's a sampleLimitErr.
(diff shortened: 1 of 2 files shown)


scrape/scrape.go
```diff
@@ -1107,40 +1113,14 @@ loop:
...
-				sampleLimitErr = err
-				added++
-				continue
-			default:
-				break loop
-			}
-		}
-		if !ok {
+			err = app.AddFast(ce.ref, t, v)
+			// call new check error function here
```
brian-brazil: style

scrape/scrape.go
```diff
@@ -1163,42 +1143,34 @@ loop:
...
-				targetScrapeSampleOutOfBounds.Inc()
-				continue
-			case errSampleLimit:
-				sampleLimitErr = err
-				added++
-				continue
-			default:
-				level.Debug(sl.l).Log("msg", "unexpected error", "series", string(met), "err", err)
+			// call new check error function here
+			sampleAdded, breakLoop, tempSampleLimitErr = sl.checkAddError(nil, met, tp, err, appErrs)
```
brian-brazil: It's not easy to track what all of these parameters mean

scrape/scrape.go
```diff
@@ -1163,42 +1143,34 @@ loop:
...
+			if sampleAdded && sampleLimitErr == nil {
+				seriesAdded++
+			}
+		}
+
+		if breakLoop {
+			break loop
+		}
+
+		if tempSampleLimitErr != nil {
```
brian-brazil: Can we not write to sampleLimitErr directly?

scrape/scrape.go
cstyan: The parameters, or return values? Either way, I'd prefer to clarify via a comment on the `checkAddError` function, but if you have another suggestion let me know.

scrape/scrape.go
roidelapluie: not sure we need that comment at all.

roidelapluie:
Hello again,

have you tested it? I would say that  appErrs should be a pointer otherwhise counters the zero value will be copied and that would not work. maybe we miss some unit tests? or is this correct?

cstyan:
We're probably missing unit tests.

> In addition to that, the behaviour of this PR does not seem correct, as apparently it does not run Add upon storage.NotFoundError. Hence why we use if !ok instead of else previously.

I misread this block/error as the storage not existing yet, not "the series wasn't found in cache/head". So yes, I broke that behaviour but there was not a test for it. If it's something we want to be able to do but never expect to happen maybe we should have a log/metric for it?

I'll go over this PR again in an hour or so, in the middle of something.

cstyan:
I've added a test for when we find a cache entry but then get `ErrNotFound` when calling `AddFast`, which in theory should never happen but the previous code handled via the `!ok` conditional, and updated my changes to pass that test.

So to respond to your previous comments:
> In addition to that, the behaviour of this PR does not seem correct, as apparently it does not run Add upon storage.NotFounfError. Hence why we use if !ok instead of else previously.

Unless I'm misunderstanding how the scrape cache works in conjunction with tsdb Head, in theory this should never happen. We just checked for a cache entry for the metric. But as I mentioned above I've updated my code to handle this case. We should perhaps add a metric/log line specifically for this case.

> The second thing that we handle differenly the case where we have storage.NotFoundError upon Add. This should never happen in theory but the old behaviour and this PR will handle that differently now.

Right, in theory it should never happen. If it does, it means `Add` has called `AddFast` after calling `head.getOrCreate` (which would have returned it a series with a ref ID), and `AddFast` has then got `ErrNotFound` for it's call to `head.getByID`.

As far as I can tell my changes are handling things the same was as before for this case now.

cstyan:
@roidelapluie @brian-brazil PTAL

scrape/scrape_test.go
```diff
@@ -1011,6 +1011,48 @@ func TestScrapeLoopAppend(t *testing.T) {
 	}
 }
 
+func TestScrapeLoopAppendCacheEntryButErrNotFound(t *testing.T) {
+	// collectResultAppender's AddFast always returns ErrNotFound if we don't give it a next
```
brian-brazil: Full stop

scrape/scrape.go
```diff
@@ -1176,42 +1157,29 @@ loop:
...
-				level.Debug(sl.l).Log("msg", "Out of bounds metric", "series", string(met))
-				targetScrapeSampleOutOfBounds.Inc()
-				continue
-			case errSampleLimit:
-				sampleLimitErr = err
-				added++
-				continue
-			default:
-				level.Debug(sl.l).Log("msg", "unexpected error", "series", string(met), "err", err)
+			// call new check error function here
```
brian-brazil: This comment doesn't seem to add anything

scrape/scrape.go
```diff
@@ -1176,42 +1157,29 @@ loop:
...
 				sl.cache.trackStaleness(hash, lset)
 			}
 			sl.cache.addRef(mets, ref, lset, hash)
-			seriesAdded++
+			if sampleAdded && sampleLimitErr == nil {
+				seriesAdded++
+			}
+		}
+
+		// Match the previous behaviour, increment added even if there's a sampleLimitErr.
```
brian-brazil: It's more that this is the number of samples post relabelling. Referring to code history like this is a bit confusing.

cstyan: Modified to more closely match the comment that was present before, specifically mentioning that this is used to track the number of samples scraped.

scrape/scrape.go
```diff
@@ -1058,16 +1058,21 @@ func (sl *scrapeLoop) getCache() *scrapeCache {
...
 	var (
 		app            = sl.appender()
 		p              = textparse.New(b, contentType)
 		defTime        = timestamp.FromTime(ts)
-		numOutOfOrder  = 0
-		numDuplicates  = 0
-		numOutOfBounds = 0
+		appErrs        = appendErrors{}
+		sampleLimitErr error
+		sampleAdded    bool
```
brian-brazil: This should be inside the for for clarity/safety

brian-brazil:
:+1:

cstyan:
I'll merge today/tomorrow

PR #6933, opened 2020-03-05
PromQL: Allow more keywords as metric names

slrtbtfs:
Also, now the behaviour is more clearly documented.


brian-brazil:
Thanks!

PR #6935, opened 2020-03-05
PromQL: Fix broken regression tests

slrtbtfs:
This PR removes the regression tests for the issue fixed in #6931 .

The reason for that is that all known queries that triggered the regression have become more or less valid syntax in #6933 (they might still fail type checking).

Signed-off-by: Tobias Guggenmos <tobias.guggenmos@uni-ulm.de>

### DIFF
--- a/promql/parser/parse_test.go
+++ b/promql/parser/parse_test.go
@@ -2164,18 +2164,6 @@ var testExpr = []struct {
 		input:  "a>b()",
 		fail:   true,
 		errMsg: `unknown function`,
-	}, {
-		input:  "rate(avg)",
-		fail:   true,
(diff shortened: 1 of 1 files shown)


brian-brazil:
Shouldn't we check that they parse as expected instead?

slrtbtfs:
Yes.On 5 Mar 2020 17:10, Brian Brazil <notifications@github.com> wrote:Shouldn't we check that they parse as expected instead?

slrtbtfs:
@brian-brazil 

> Shouldn't we check that they parse as expected instead?

@brian-brazil The test expressions are kept now.

brian-brazil:
Thanks!

PR #6937, opened 2020-03-06
[test] Add tests for promql/parser.Tree()

liguozhong:
[test] Add tests for promql/parser.Tree()

### DIFF
--- a/promql/parser/printer_test.go
+++ b/promql/parser/printer_test.go
@@ -103,3 +103,149 @@ func TestExprString(t *testing.T) {
 		testutil.Equals(t, expr.String(), exp)
 	}
 }
+
+func TestTree(t *testing.T) {
+	// A list of valid expressions that are expected to be
+	// returned as out when calling String(). If out is empty the output
+	// is expected to equal the input.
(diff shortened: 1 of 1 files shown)


roidelapluie:
Maybe we can update existing tests instead of writing a new ones? Also it is not clear if this year is needed, is this function used by Prometheus?

liguozhong:
> Maybe we can update existing tests instead of writing a new ones? Also it is not clear if this year is needed, is this function used by Prometheus?

This function has not been used inside Prometheus, presumably for other systems.
 When I saw this function, I found that this function was very complex, but there was no test function, so I added one test.

slrtbtfs:
The only use case for that tree function I'm aware of is debugging the parser.

Having a test for it doesn't hurt, though.

gouthamve:
Hrm, do we really need unit-tests for this though? There are regression tests and we are not sure if tests are warranted for this code. We've decided to close this as part of our bug scrub, sorry! Thanks for submitting the PR though, would always appreciate more tests for a lot of the components, but sadly not this one :(

PR #6938, opened 2020-03-07
tsdb: fix typo

johncming:
Fix typo here.

The cap of `tmpFiles` is the number of `dirs` not length of string `dir`.

@codesome


brian-brazil:
Thanks!

PR #6941, opened 2020-03-07
remove Unused parameter 'reg' in wal.Open function

liguozhong:
remove Unused parameter 'reg' in wal.Open function


roidelapluie:
Did you check if that was there to respect an interface or so? Maybe that's useful fir corthanos?

PR #6944, opened 2020-03-09
tsdb/chunks: remove un-used code.

johncming:
This code below is not used. If not used in the future, it is better to delete it.

https://github.com/prometheus/prometheus/blob/0a27df92f091a7d7eb6087892f307c72e70b2f0d/tsdb/chunks/chunks.go#L460-L462

It's not a bug.

### DIFF
--- a/tsdb/chunks/chunks.go
+++ b/tsdb/chunks/chunks.go
@@ -457,10 +457,6 @@ func (b realByteSlice) Range(start, end int) []byte {
 	return b[start:end]
 }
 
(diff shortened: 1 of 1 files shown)


roidelapluie:
We should check with Thanos and Cortex cc @pracucci @bwplotka

pracucci:
> Cortex

Thanks @roidelapluie for the heads up! I haven't found any reference in Cortex.

gouthamve (approved):

gouthamve:
Thanks!

PR #6945, opened 2020-03-09
[performance optimization] tsdb: Merge two loops in ‘delete’ function

liguozhong:
[performance optimization] tsdb: Merge two loops in ‘delete’ function

### DIFF
--- a/tsdb/index/postings.go
+++ b/tsdb/index/postings.go
@@ -214,22 +214,18 @@ func (p *MemPostings) Delete(deleted map[uint64]struct{}) {
 			p.mtx.Lock()
 
 			found := false
-			for _, id := range p.m[n][l] {
-				if _, ok := deleted[id]; ok {
-					found = true
-					break
-				}
-			}
-			if !found {
-				p.mtx.Unlock()
-				continue
-			}
 			repl := make([]uint64, 0, len(p.m[n][l]))
 
 			for _, id := range p.m[n][l] {
 				if _, ok := deleted[id]; !ok {
 					repl = append(repl, id)
+				} else {
+					found = true
 				}
+			}
+			if !found {
+				p.mtx.Unlock()
+				continue
 			}
 			if len(repl) > 0 {
 				p.m[n][l] = repl


csmarchbanks:
Do you have benchmarks showing a performance improvement? This change will always alloce a slice even in the case where nothing is to be deleted which could cause worse performance.

liguozhong:
> Do you have benchmarks showing a performance improvement? This change will always alloce a slice even in the case where nothing is to be deleted which could cause worse performance.

yes. I run the benchmark and confirmed that it can double the performance.
image
(this pr)
image
(old delete )

roidelapluie:
@csmarchbanks Should we run prombench on this? I don't know if we have other benchmarks that reflects real use cases for this.

csmarchbanks:
It will probably be hard to tell in a prombench if this makes any difference as I think it would only happen during compaction where a lot of cpu intensive things are happening.

To merge this I would like to see benchmarks added to this MR that can be verified, when I ran the benchmark that seems to have been deleted from this thread I actually found the new code to be slower.
```
benchmark                         old ns/op     new ns/op     delta
BenchmarkMemPostings_Delete-8     94931304      105818435     +11.47%
```

liguozhong:
> It will probably be hard to tell in a prombench if this makes any difference as I think it would only happen during compaction where a lot of cpu intensive things are happening.
> 
> To merge this I would like to see benchmarks added to this MR that can be verified, when I ran the benchmark that seems to have been deleted from this thread I actually found the new code to be slower.
> 
> ```
> benchmark                         old ns/op     new ns/op     delta
> BenchmarkMemPostings_Delete-8     94931304      105818435     +11.47%
> ```
OK, I'll try

gouthamve:
Hi @liguozhong, thanks for submitting the PR, seems useful, but with perf its always tricky. Did you have new benchmarks for this?

liguozhong:
> Hi @liguozhong, thanks for submitting the PR, seems useful, but with perf its always tricky. Did you have new benchmarks for this?

My benchmark shows that this PR makes the performance worse, this is not a good idea

brian-brazil:
Thanks for the update, I guess we should close this off then.

PR #6947, opened 2020-03-09
Add new "globalUrl" field to /api/v1/targets docs

juliusv:
See https://github.com/prometheus/prometheus/pull/6794


juliusv:
@roidelapluie I didn't elaborate on the meaning of the field, since we don't do that for most other existing fields either (yet)...

brian-brazil:
I'm not sure that this UI implementation detail should leak into our public APIs.

juliusv:
@brian-brazil See https://github.com/prometheus/prometheus/pull/6794#issuecomment-583842388, I allowed it since I thought it could be useful outside of UI purposes as well.

brian-brazil:
Ah, I'd missed that question to me. I've personally wondered if we should do this munging at all.

juliusv:
I think it's useful enough to do.

brian-brazil:
It's a bit messy, as we're trying to guess how name resolution looks like to unknown 3rd parties. I've always kinda wondered if it was worth it. Personally I'd only every scrape "localhost", as against the machine name, when throwing together a quick test Prometheus as otherwise you don't know which localhost is which as the setup gets more complicated.

roidelapluie:
What is the way forward?

juliusv:
Not sure. I buy the argument that localhost links aren't a good idea for production setups. In playground setups though this setting has been quite handy :)

brian-brazil:
In a playground wouldn't localhost be the machine your browser is on :)

juliusv:
Concretely I was thinking about workshops where I had participants use a VM and point the Prometheus on there at all kinds of local stuff (Prometheus, PGW, Node Exporter, Alertmanager, etc.). You can also use absolute IPs for that, it's just more work in terms of participants needing to customize their example config.

juliusv:
I also imagine some (many?) people will try out https://prometheus.io/docs/introduction/first_steps/ not on their own machine, but some VM, in which case the localhost URL we provide in the example config in there will also not be clickable.

juliusv:
Though TBF in that case we also don't provide an external URL, so we don't solve that problem in that case either.

roidelapluie:
The fact that it is documented or not does not change much in term of API compatibility. We can't remove it without changing the API. Since we still expose this today, I will merge this.

PR #6951, opened 2020-03-09
tsdb: Do a full rollback upon commit error

beorn7:
I think the previous behavior is problematic as it will leave
`memSeries` around that still have `pendingCommit` set to `true`.

The only case where this can happen in this code path is a failure to
write to the WAL, in which case we are probably in trouble anyway. I
believe, however, we should still try to do the right thing and do the
full rollback. This will implicitly try to write to the WAL again, but
this time without samples, which may even succeed. (But we propagate
the previous error in any case.)

This is another thing I ran into while staring at code for the isolation implementation.

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -1042,14 +1042,16 @@ func (a *headAppender) log() error {
 }
 
 func (a *headAppender) Commit() error {
+	if err := a.log(); err != nil {
+		//nolint: errcheck
+		a.Rollback() // Most likely the same error will happen again.
+		return errors.Wrap(err, "write to WAL")
+	}
+
 	defer a.head.metrics.activeAppenders.Dec()
 	defer a.head.putAppendBuffer(a.samples)
 	defer a.head.putSeriesBuffer(a.sampleSeries)
 	defer a.head.iso.closeAppend(a.appendID)
-
-	if err := a.log(); err != nil {
-		return errors.Wrap(err, "write to WAL")
-	}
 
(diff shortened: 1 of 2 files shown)


roidelapluie:
I don't think you meant to put that before the defer, did you?

roidelapluie:
> I don't think you meant to put that before the defer, did you?

oh Rollback will have the same defer.

tsdb/head.go
```diff
@@ -1042,15 +1042,17 @@ func (a *headAppender) log() error {
 }
 
 func (a *headAppender) Commit() error {
+	if err := a.log(); err != nil {
+		//nolint: errcheck
+		a.Rollback() // Most likely the same error will happen again.
```
codesome: `a.head.putSeriesBuffer(a.sampleSeries)` seems to be missing in `Rollback` (which is in `Commit`). Not sure if that intentional, but I guess we could add it in there too.

roidelapluie:
What are the risks of running a.log twice?

codesome:
> What are the risks of running a.log twice?

No risks. Duplicate series records are handles in the WAL replay.

beorn7:
@codesome I have addressed your comments. Perhaps you could have another look before I merge this.

beorn7:
@codesome is asleep now. @roidelapluie since you approved and @codesome's comments are clearly addressed now, I'll merge (to not miss tomorrow's rc.0 train).

PR #6953, opened 2020-03-10
[OpenStack SD] Add HypervisorID meta labels and OpenStack Api version…

pzqu:
… SDConfig

Add extra meta labels which will be useful in the case
Prometheus discovery hypervisor from all projects.
The id of the hypervisor as id Available until version 2.52,
The id of the hypervisor as a UUID New in version 2.53
Distinguish between different OpenStack Api versions.
maintain the backward compatibility.

### DIFF
--- a/discovery/openstack/hypervisor.go
+++ b/discovery/openstack/hypervisor.go
@@ -30,6 +30,7 @@ import (
 )
 
 const (
+	openstackLabelHypervisorID       = openstackLabelPrefix + "hypervisor_id"
 	openstackLabelHypervisorHostIP   = openstackLabelPrefix + "hypervisor_host_ip"
 	openstackLabelHypervisorHostName = openstackLabelPrefix + "hypervisor_hostname"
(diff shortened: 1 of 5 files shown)


brian-brazil:
Adding a requirement that all users now specify the exact version they're talking to seems a bit much for the sake of a single metadata label. Could we do something more automatic?

pzqu:
"api_version" is an optional condition. It is not necessary.
When there is no "api_version" in the configuration file, the original logic is performed.

code in there:
https://github.com/prometheus/prometheus/blob/261cbab8e9fa470b72693c984cc25e9be6e33e83/vendor/github.com/gophercloud/gophercloud/service_client.go#L65

OpenStack api uses micro version, the output will be affected by micro version (compatible with older versions)
reference:
https://docs.openstack.org/api-guide/compute/microversions.html

So this feature provides a configurable API version

brian-brazil:
So that option could cause all sorts of unexpected behaviour down the line. This sort of detail should not be exposed to the user.

pzqu:
thanks

PR #6958, opened 2020-03-10
PromQL: Fix lexer error handling

slrtbtfs:
This fixes bugs in the handling of lexer errors that are only noticeable for users of the language server and caused https://github.com/prometheus-community/promql-langserver/issues/104 .

Signed-off-by: Tobias Guggenmos <tobias.guggenmos@uni-ulm.de>

### DIFF
--- a/promql/parser/parse.go
+++ b/promql/parser/parse.go
@@ -242,6 +242,11 @@ func (p *parser) addParseErr(positionRange PositionRange, err error) {
 func (p *parser) unexpected(context string, expected string) {
 	var errMsg strings.Builder
 
+	// Do not report lexer errors twice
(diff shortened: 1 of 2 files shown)


roidelapluie:
Is that something that can be tested?

slrtbtfs:
Yes. I've manually tested using the language server.

The existing parser unit tests don't catch the bugs fixed here because they only look at the first error message raised and ignore the end positions for error ranges.

Properly checking these things in the parser unit tests would need either adding a new test or extensively rewriting the existing ones.

There are some basic sanity checks that take less effort which I will add to this PR.

slrtbtfs:
I've added some sanity checks for position ranges.

roidelapluie:
Thanks!

Issue #6952, opened 2020-03-10
prometheus_remote_storage_pending_samples are drastically increasing over time.


Seitanas:
**What did you do?**
Using 3 standalone victoria-metrics as LTS storage for prometheus.

**What did you expect to see?**
`prometheus_remote_storage_pending_samples` should not increase over time.

**What did you see instead? Under which circumstances?**
Increase on `prometheus_remote_storage_pending_samples` metric.
It does not look that there is a metric lag on victoria-metrics site though.

**Environment**
Centos 7. Prometheus 2.16.0 (same issue on older versions). Victoria-metrics 1.32.6
About 1.5k scrape targets.

* System information:

	`Linux 3.10.0-957.21.3.el7.x86_64 x86_64`

* Prometheus version:
```
prometheus, version 2.16.0 (branch: HEAD, revision: b90be6f32a33c03163d700e1452b54454ddce0ec)
  build user:       root@7ea0ae865f12
  build date:       20200213-23:50:02
  go version:       go1.13.8
```

* Prometheus configuration file:
```
---
global:
  scrape_interval: 30s
  scrape_timeout: 5s
rule_files:
- "/etc/prometheus/rules/*"
remote_write:
- url: http://victoria-metrics1:8428/api/v1/write
  queue_config:
    max_shards: 50
    min_shards: 2
    capacity: 1500000
    max_samples_per_send: 150000
    batch_send_deadline: 2s
- url: http://victoria-metrics2:8428/api/v1/write
  queue_config:
    max_shards: 50
    min_shards: 2
    capacity: 1500000
    max_samples_per_send: 150000
    batch_send_deadline: 2s
- url: http://victoria-metrics3:8428/api/v1/write
  queue_config:
    max_shards: 50
    min_shards: 2
    capacity: 1500000
    max_samples_per_send: 150000
    batch_send_deadline: 2s
```

Prometheus seems to work fine for some time, but then it will start drastically increse `prometheus_remote_storage_pending_samples` metrics. It does not seem, that there's any lag in metrics, which are written to remote storage.
It also appears, that `prometheus_remote_storage_pending_samples` begins to increase right after `WAL checkpoint`.  You can see that during checkpoint, prometheus begins increasing shard count and  also `pending_samples` begins to grow:

`prometheus_remote_storage_shards`:
[...]

PR #6960, opened 2020-03-10
Fix bug that allows for sample enqueuing to shards that are being shutdown

cstyan:
We need to check ctx.Done in it's own select call since `select` ordering is non-deterministic. Without doing this, it's possible that even though `hardShutdown` has been initiated, the previous select block could have still chosen to allow new samples to be enqueued. The test I've added will fail consistently on master.

I think this is at least part of the cause of #6952 

cc @roidelapluie @Seitanas

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -791,7 +791,10 @@ func (s *shards) runShard(ctx context.Context, shardID int, queue chan sample) {
 		select {
 		case <-ctx.Done():
 			return
+		default:
+		}
 
+		select {
 		case sample, ok := <-queue:
 			if !ok {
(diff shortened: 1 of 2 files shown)


roidelapluie:
@cstyan Is there an impact for end users beside the metric potentially fixed?

cstyan:
Yeah @roidelapluie don't merge this yet. I was just able to reproduce the issue outside of a test so I want to look into this more tomorrow. If the release is set for tomorrow then just go ahead with it.

Seitanas:
Hello, any update on this?

cstyan:
closing this for now until we have time to confirm the entire cause for #6952

PR #6961, opened 2020-03-11
tsdb: update pool initialization and size.

johncming:
1. Use the recommended method to initialize pool
2. Increase `bytesPool` size. Each series or sample has 2 bytes, which is small. It will lead to more allocation. Because of a series(one label) or a sample use about 16bytes after encoded in memory(just a rough estimate). After the program has been running for some time, the `cap` of the byte array in the pool will be inconsistent. Some byte arrays are bigger than `1024`.

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -293,6 +293,21 @@ func NewHead(r prometheus.Registerer, l log.Logger, wal *wal.WAL, chunkRange int
 		tombstones: tombstones.NewMemTombstones(),
 		iso:        newIsolation(),
 		deleted:    map[uint64]int{},
+		appendPool: sync.Pool{
+			New: func() interface{} {
+				return make([]record.RefSample, 0, 512)
+			},
+		},
+		seriesPool: sync.Pool{
+			New: func() interface{} {
+				return make([]*memSeries, 0, 512)
+			},
+		},
+		bytesPool: sync.Pool{
+			New: func() interface{} {
+				return make([]byte, 0, 8192)
+			},
+		},
 	}
 	h.metrics = newHeadMetrics(h, r)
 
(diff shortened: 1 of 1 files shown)


tsdb/head.go
```diff
@@ -293,6 +293,21 @@ func NewHead(r prometheus.Registerer, l log.Logger, wal *wal.WAL, chunkRange int
...
+			},
+		},
+		seriesPool: sync.Pool{
+			New: func() interface{} {
+				return make([]*memSeries, 0, 512)
+			},
+		},
+		bytesPool: sync.Pool{
+			New: func() interface{} {
+				return make([]byte, 0, 8192)
```
codesome: Still not convinced with your explanation for increasing this size, do you have any benchmark results? This pool seems to be used only during logging into the WAL, so that data doesn't increase with time. And it should be fine to have a slightly smaller size here, if it grows at some point, it will be re-used, hence there won't be frequent allocations.

johncming: 1. `buf`'s cap is no longer the same
After taking out the `buf` from the pool, if the `buf`'s cap changes, put this `buf`back in the pool, and then take it out, the `buf`'s cap is equal to last changed value. Over time, the `buf` in the pool will no longer be the same size. Small caps are more susceptible.

johncming: 2. benchmark
Maybe there is no optimal size. Because there are more situations.

johncming: Is it appropriate to introduce metric here to record(considering cost)？I have no idea about it.

codesome: Without any real numbers on if it is actually benefiting much and not affecting the memory of smaller setups, I would not like to change the size here. The growth in the slice cap would be only a few times before it becomes stable and that is fine.

johncming: good

codesome:
Wait why did the PR close? :) Other than changing the size, everything else was good and to be merged. Can you reopen with `bytesPool` set to the old size?

PR #6965, opened 2020-03-11
Use `struct{}` as underlying type for context keys

beorn7:
This is an alternative to #6963.

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -152,9 +152,7 @@ type query struct {
 	ng *Engine
 }
 
-type queryCtx int
-
-var queryOrigin queryCtx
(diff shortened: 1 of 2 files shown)


beorn7:
Will remove the Dockerfile changes, sorry.

promql/engine.go
```diff
@@ -152,9 +152,7 @@ type query struct {
 	ng *Engine
 }
 
-type queryCtx int
-
-var queryOrigin queryCtx
+type queryOrigin struct{}
```
roidelapluie:
```suggestion
var queryOrigin = struct{}{}
```

roidelapluie:
Thanks!

PR #6966, opened 2020-03-11
Release 2.17.0-rc.0

roidelapluie:


CHANGELOG.md
```diff
@@ -1,3 +1,25 @@
+## 2.17.0-rc.0 / 2020-03-11
+
+This release implements isolation in TSDB. API queries and recording rules are guaranteed to only see full scrapes and full recording rules. This causes a slight increase in memory usage.
+
+* [ENHANCEMENT] PromQL: Allow more keywords as metric names #6933
+* [ENHANCEMENT] React UI: Adds normalization of localhost urls in targets page react #6794
```
roidelapluie:
```suggestion
* [ENHANCEMENT] React UI: Adds normalization of localhost urls in targets page #6794
```

simonpasquier:
LGTM

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
+## 2.17.0-rc.0 / 2020-03-11
+
+This release implements isolation in TSDB. API queries and recording rules are guaranteed to only see full scrapes and full recording rules. This causes a slight increase in memory usage.
```
beorn7: "…see full scrapes and full recording rules." → "…see the full result of a scrape or rule evaluation."

"This causes a slight increase in memory usage." → It's not only memory usage. Perhaps something like: "This comes with a certain overhead in resource usage. Depending on the situation, there might be a noticeable increase in memory usage, CPU usage, or query latency."

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
...
+
+* [ENHANCEMENT] PromQL: Allow more keywords as metric names #6933
+* [ENHANCEMENT] React UI: Add normalization of localhost urls in targets page #6794
```
beorn7: urls → URLs

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
...
+* [ENHANCEMENT] Rules: Make deleted rule series as stale after a reload #6745
```
beorn7: Make → Mark

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
...
+* [ENHANCEMENT] Scrape: Log scrape append failures as debug rather than warn #6852
+* [ENHANCEMENT] TSDB: Support isolation #6841
+* [ENHANCEMENT] TSDB: Log WAL replay duration #6922
+* [ENHANCEMENT] TSDB: Improve query performance for queries that partially hit the head #6676
+* [ENHANCEMENT] Consul SD: Add health label to metrics #5313
+* [ENHANCEMENT] EC2 SD: Expose EC2 instance lifecycle as label #6914
+* [ENHANCEMENT] Kubernetes SD: add service type metadata to k8s service role #6496 #6684
```
beorn7: add → Add
k8s → K8s

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
...
+* [ENHANCEMENT] Kubernetes SD: expose label_selector and field_selector #6807
```
beorn7: expose → Expose

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
...
+* [ENHANCEMENT] Openstack SD: Add openstack_hypervisor_id meta label #6962
+* [BUGFIX] PromQL: Do not escape html-like chars in query log #6834 #6795
```
beorn7: html → HTML

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
...
+* [BUGFIX] React UI: Fix data table matrix values #6896
+* [BUGFIX] React UI: Fix new targets page not loading when using non-ascii characters #6892
```
beorn7: ascii → ASCII

CHANGELOG.md
roidelapluie: Indeed. Updated.

CHANGELOG.md
roidelapluie: Thanks

CHANGELOG.md
```diff
@@ -1,3 +1,27 @@
+## 2.17.0-rc.0 / 2020-03-11
+
+This release implements isolation in TSDB. API queries and recording rules are
+guaranteed to only see full scrapes and full recording rules. This comes with a
+certain overhead in resource usage. Depending on the situation, there might be a
```
bwplotka: It sounds like "Inevitable you will see certain overhead". It really IMO sounds more negative then it is... I think we can tell this softer.

CHANGELOG.md
```diff
@@ -1,3 +1,27 @@
...
+noticeable increase in memory usage, CPU usage, or query latency.
```
bwplotka:
```suggestion
some increase in memory usage, CPU usage, or query latency.
```

roidelapluie: @beorn7 WDYT?

beorn7: I'm a fan of "underpromise, overdeliver", and I found my wording already quite cautious (with "might" and such), but I don't feel strongly.

But if you do want to change "noticeable" to "some", then you have to delete the "a" in the previous line.

CHANGELOG.md
```diff
@@ -1,3 +1,27 @@
...
+
+* [ENHANCEMENT] PromQL: Allow more keywords as metric names #6933
+* [ENHANCEMENT] React UI: Add normalization of localhost URLs in targets page #6794
+* [ENHANCEMENT] Rules: Mark deleted rule series as stale after a reload #6745
+* [ENHANCEMENT] Scrape: Log scrape append failures as debug rather than warn #6852
+* [ENHANCEMENT] TSDB: Support isolation #6841
```
brian-brazil: This is a FEATURE, it's pretty big

CHANGELOG.md
```diff
@@ -1,3 +1,27 @@
...
+* [ENHANCEMENT] TSDB: Log WAL replay duration #6922
+* [ENHANCEMENT] TSDB: Improve query performance for queries that partially hit the head #6676
+* [ENHANCEMENT] Consul SD: Add health label to metrics #5313
```
brian-brazil: To metadata?

CHANGELOG.md
```diff
@@ -1,3 +1,27 @@
...
+guaranteed to only see full scrapes and full recording rules. This comes with a
+certain overhead in resource usage. Depending on the situation, there might be a
+noticeable increase in memory usage, CPU usage, or query latency.
+
+* [ENHANCEMENT] PromQL: Allow more keywords as metric names #6933
+* [ENHANCEMENT] React UI: Add normalization of localhost URLs in targets page #6794
+* [ENHANCEMENT] Rules: Mark deleted rule series as stale after a reload #6745
+* [ENHANCEMENT] Scrape: Log scrape append failures as debug rather than warn #6852
+* [ENHANCEMENT] TSDB: Support isolation #6841
+* [ENHANCEMENT] TSDB: Log WAL replay duration #6922
```
brian-brazil: This seems a bit small to mention

roidelapluie:
Alright, I merge & publish the RC tomorrow morning.

roidelapluie:
@beorn7 Instantiating a queries now takes a very long time:
prom

Maybe that's worth taking a look at the code.

beorn7:
That's a lot of prepare time. Perhaps most of the query overhead of isolation is happening here? It's not a lot relative to total query time.

The benchmark seems to trigger quite serious peaks in watermark differences. That's not what I saw in production servers under normal query load. I guess that makes the isolation state snapshotting quite expensive…

Anyday, definitely worth some investigation.

beorn7:
I did some profiling.

The additional time spent in prepare seems to be burned mostly in https://github.com/prometheus/prometheus/blob/bf0055536d8463a27725ce9a35ccba55e97d6e5b/tsdb/head.go#L1404 . The code in there has indeed changed a lot between 2.16.0 and now. [This whole section](https://github.com/prometheus/prometheus/blob/bf0055536d8463a27725ce9a35ccba55e97d6e5b/tsdb/head.go#L1408-L1433) is new, replacing a single line we had previously.

This is coming from https://github.com/prometheus/prometheus/pull/6676 , which is actually an optimization.

From my (limited) understanding, I'd say that what we see here is not additional time spent, but just time moved from one phase of the query evaluation to another phase of the query evaluation, i.e. the selection of series within the time range is now done during the prepare phase.

In any case, I think this is likely not related to isolation.

roidelapluie:
/prombench cancel

We will have another long run with the rc.1.

PR #6971, opened 2020-03-12
storage: Removed SelectSorted method while maintaining logic; Simplified interface.

bwplotka:
I found during work on https://github.com/prometheus/prometheus/pull/5882 that
we do so many repetitions because of this, for no good reason. I think
I found a good balance between convenience and readability with just one method.
Smaller the interface = better.

Also, I don't know what TestSelectSorted was testing, but now it's testing sorting.

cc @beorn7 @brian-brazil 

### DIFF
--- a/storage/fanout/fanout_test.go
+++ b/storage/fanout/fanout_test.go
@@ -79,7 +79,7 @@ func TestSelectSorted(t *testing.T) {
 	matcher, err := labels.NewMatcher(labels.MatchEqual, model.MetricNameLabel, "a")
 	testutil.Ok(t, err)
 
-	seriesSet, _, err := querier.Select(false, nil, matcher)
+	seriesSet, _, err := querier.Select(true, nil, matcher)
 	testutil.Ok(t, err)
 
 	result := make(map[int64]float64)
--- a/storage/fanout/fanout_test.go
+++ b/storage/fanout/fanout_test.go
@@ -79,7 +79,7 @@ func TestSelectSorted(t *testing.T) {
 	matcher, err := labels.NewMatcher(labels.MatchEqual, model.MetricNameLabel, "a")
 	testutil.Ok(t, err)
 
-	seriesSet, _, err := querier.SelectSorted(nil, matcher)
+	seriesSet, _, err := querier.Select(false, nil, matcher)
 	testutil.Ok(t, err)
 
 	result := make(map[int64]float64)
--- a/storage/remote/codec.go
+++ b/storage/remote/codec.go
@@ -79,22 +79,22 @@ func EncodeReadResponse(resp *prompb.ReadResponse, w http.ResponseWriter) error 
 }
 
 // ToQuery builds a Query proto.
-func ToQuery(from, to int64, matchers []*labels.Matcher, p *storage.SelectParams) (*prompb.Query, error) {
+func ToQuery(from, to int64, matchers []*labels.Matcher, hints *storage.SelectHints) (*prompb.Query, error) {
 	ms, err := toLabelMatchers(matchers)
 	if err != nil {
 		return nil, err
 	}
 
 	var rp *prompb.ReadHints
-	if p != nil {
+	if hints != nil {
 		rp = &prompb.ReadHints{
-			StepMs:   p.Step,
-			Func:     p.Func,
-			StartMs:  p.Start,
-			EndMs:    p.End,
-			Grouping: p.Grouping,
-			By:       p.By,
-			RangeMs:  p.Range,
+			StartMs:  hints.Start,
+			EndMs:    hints.End,
+			StepMs:   hints.Step,
+			Func:     hints.Func,
+			Grouping: hints.Grouping,
+			By:       hints.By,
+			RangeMs:  hints.Range,
 		}
 	}
 
@@ -145,7 +145,7 @@ func ToQueryResult(ss storage.SeriesSet, sampleLimit int) (*prompb.QueryResult, 
 }
 
 // FromQueryResult unpacks and sorts a QueryResult proto.
-func FromQueryResult(res *prompb.QueryResult) storage.SeriesSet {
+func FromQueryResult(sortSeries bool, res *prompb.QueryResult) storage.SeriesSet {
 	series := make([]storage.Series, 0, len(res.Timeseries))
 	for _, ts := range res.Timeseries {
 		labels := labelProtosToLabels(ts.Labels)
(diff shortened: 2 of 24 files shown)


brian-brazil:
I don't think it's safe to use SelectParams for this. SelectParams are for optimisations that you can ignore, sorting is about correctness.

bwplotka:
I would vote for changing that then. Any blocker for that? It's well-known API, and I am 100% sure we don't ignore SelectParams anywhere.

brian-brazil:
That'd be a breaking change to the API, for example no version of Prometheus uses these.

SelectParams are intended to allow remote storage implementations to experiment with optimisations. A given PromQL query should return the exact same ultimate result if the SelectParams were missing.

bwplotka:
> That'd be a breaking change to the API, for example, no version of Prometheus uses these.

That's is not breaking any API that we version. `Select` is our public Go API which we agreed that we don't version until proper module/submodule solution.

> SelectParams are intended to allow remote storage implementations to experiment with optimizations. A given PromQL query should return the exact same ultimate result if the SelectParams were missing.

Agree :+1:  but I did not change any behavior. Remote read client behaves exactly the same. Only if SelectParams is not empty we pass select params as hints :+1:

roidelapluie:
I fail really see why we need a new struct with only 2 fields. I will dig into this later.

bwplotka:
@roidelapluie  As there is no default value for our timestamp fields  `Start` and `End` (0 is valid time) (: So have to use the pointer  (:

brian-brazil:
> That's is not breaking any API that we version. Select is our public Go API which we agreed that we don't version until proper module/submodule solution.

This is the same interface we use for remote read, and while that's technically experimental we shouldn't be doing anything to break it. This internal detail of the tsdb also shouldn't leak to remote read.

bwplotka:
> This is the same interface we use for remote read

What do you mean? That's not a remote read interface. The proto one is, right? (:

brian-brazil:
The two are kept in sync, and we've had problems in the past where they weren't.

bwplotka:
They not in sync anyway:

* Number of methods: One `api/v1/read` vs  2 (!) Select methods.
* Naming of methods (!) Select vs Read
* hints vs params
* Warnings vs no warnings
* Non streaming version available

In fact, sorting is very important, and having this part inside params is much more verbose than comment in proto that one type (non streamed) is not sorted and second is sorted.

Is there any alternative to this? We could add another `seriesSorted bool` parameter to `Select` interface, but given we have params I would go for this. :hugs: (: It's extremely hard to know that those two (/api/v1/read vs select) has to be kept in sync. This is nice, but even me - I was not aware there they meant to be kept in sync 1:1 (even though they are far from being in sync).

brian-brazil:
I meant SelectParams and ReadHints - these should and do match.

bwplotka:
Ok, what do you think @brian-brazil  to change all to `Select(sortSeries bool, SelectHints *hints, ... matchers)` then? :thinking: That would make sense and pass your sync rule (:

brian-brazil:
I'm not sure that's much of an improvement, there's a school of thought that boolean params like that should be two functions instead.

bwplotka:
> I'm not sure that's much of an improvement, there's a school of thought that boolean params like that should be two functions instead.

Maybe but it get's tricky. With chunk Select (select that gives chunk iterators) we have 4 functions. That's a no-go, so I would vote for this. WDYT?

brian-brazil:
If we're getting to that stage then a bool argument probably makes sense.

bwplotka:
Awesome!

Actually I think would could change it to `Select(SelectParams params, SelectHints *hints, ... matchers)` if that's ok, so it's more extensible? :thinking: WDYT?

brian-brazil:
It's an internal interface so we can extend it as needed in future.

bwplotka:
All applied @brian-brazil (: 

PTAL! :heart:

storage/fanout/fanout_test.go
```diff
@@ -79,7 +79,7 @@ func TestSelectSorted(t *testing.T) {
 	matcher, err := labels.NewMatcher(labels.MatchEqual, model.MetricNameLabel, "a")
 	testutil.Ok(t, err)
 
-	seriesSet, _, err := querier.SelectSorted(nil, matcher)
+	seriesSet, _, err := querier.Select(false, nil, matcher)
```
brian-brazil: Shouldn't this be true so the test is deterministic?

storage/remote/codec.go
```diff
@@ -158,7 +158,10 @@ func FromQueryResult(res *prompb.QueryResult) storage.SeriesSet {
 			samples: ts.Samples,
 		})
 	}
-	sort.Sort(byLabel(series))
+
+	if sortSeries {
```
brian-brazil: Is there ever a case where we don't want this sorted?

tsdb/querier.go
```diff
@@ -85,23 +85,21 @@ func (q *querier) lvals(qs []storage.Querier, n string) ([]string, storage.Warni
...
+		return q.blocks[0].Select(sortSeries, hints, ms...)
+	}
+
 	ss := make([]storage.SeriesSet, len(q.blocks))
 	var ws storage.Warnings
 	for i, b := range q.blocks {
-		s, w, err := b.SelectSorted(p, ms...)
+		// Sorting Head series is slow, and unneeded when only the
+		// Head is being queried. Sorting blocks is a noop.
+		// Still we have to sort if blocks > 1 as Merged Series requires.
```
brian-brazil: as MergedSeriesSet requires it.

storage/fanout/fanout_test.go
bwplotka: Ah, nice ! missed this

storage/remote/codec.go
bwplotka: We query JUST one remote read storage? (based on required matchers? :thinking:

brian-brazil: I think we'd still try to query local storage then.

brian-brazil:
:+1:

tsdb/querier.go
```diff
@@ -85,23 +85,21 @@ func (q *querier) lvals(qs []storage.Querier, n string) ([]string, storage.Warni
...
+	if len(q.blocks) == 1 {
+		return q.blocks[0].Select(sortSeries, hints, ms...)
+	}
+
 	ss := make([]storage.SeriesSet, len(q.blocks))
 	var ws storage.Warnings
 	for i, b := range q.blocks {
-		s, w, err := b.SelectSorted(p, ms...)
+		// Sorting Head series is slow, and unneeded when only the
+		// Head is being queried. Sorting blocks is a noop.
```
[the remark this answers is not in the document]
bwplotka: I copied this comment from somewhere, will revisit

bwplotka:
Thanks all for a quick review! :+1:

vsliouniaev:
Any chance of this making it into 2.17? Am hoping for the above bugfix

roidelapluie:
can we pinpoint why this fixes the bug? which part of the code?

PR #6974, opened 2020-03-13
changes in results vector components

seepls:
I have made changes in results which is a promql vector, thought should have all possible types. Kindly Review.


brian-brazil:
This is causing the tests to fail, what are you trying to do?

seepls:
I got confused with the conditions. Sorry.

brian-brazil:
No problem.

PR #6975, opened 2020-03-13
Fix panic when a remote read store errors

csmarchbanks:
In v2.17.0-rc.0 if you configure a remote read endpoint that fails for some reason the entire query panics due to a `nil` series set being added to the result.

This only affects the release candidate, not previous Prometheus versions.

@roidelapluie I opened this against branch release-2.17 to keep the normal flow described in RELEASE.md, let me know if you would like me to do something different.

### DIFF
--- a/storage/fanout.go
+++ b/storage/fanout.go
@@ -262,6 +262,7 @@ func (q *mergeQuerier) SelectSorted(params *SelectParams, matchers ...*labels.Ma
 			} else {
 				priErr = qryResult.selectError
 			}
+			continue
 		}
 		seriesSets = append(seriesSets, qryResult.set)
 	}
--- a/storage/fanout/fanout_test.go
+++ b/storage/fanout/fanout_test.go
@@ -15,6 +15,7 @@ package storage
 
 import (
 	"context"
+	"errors"
 
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
@@ -26,7 +27,6 @@ import (
 )
 
 func TestSelectSorted(t *testing.T) {
-
 	inputLabel := labels.FromStrings(model.MetricNameLabel, "a")
 	outputLabel := labels.FromStrings(model.MetricNameLabel, "a")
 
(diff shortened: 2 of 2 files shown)


storage/fanout.go
```diff
@@ -262,6 +262,7 @@ func (q *mergeQuerier) SelectSorted(params *SelectParams, matchers ...*labels.Ma
 			} else {
 				priErr = qryResult.selectError
 			}
+			continue
 		}
 		seriesSets = append(seriesSets, qryResult.set)
 	}
```
roidelapluie: What would you think on inlining

```
		if qryResult.selectError != nil {
			q.failedQueriers[qryResult.qr] = struct{}{}
			if qryResult.qr != q.primaryQuerier {
				warnings = append(warnings, qryResult.selectError)
			} else {
				priErr = qryResult.selectError
			}
			continue
		}
		seriesSets = append(seriesSets, qryResult.set)
```

to

```
		if qryResult.selectError == nil {
			seriesSets = append(seriesSets, qryResult.set)
			continue
                }
		q.failedQueriers[qryResult.qr] = struct{}{}
		if qryResult.qr != q.primaryQuerier {
			warnings = append(warnings, qryResult.selectError)
		} else {
			priErr = qryResult.selectError
		}
```

csmarchbanks: Go style suggests handling the error case first, and leaving the normal case un-indented. I don’t feel super strongly but would prefer to stay with that idiom.

roidelapluie: ok!

bwplotka: I like this suggestion.

bwplotka:
I still think what @roidelapluie suggest is better. Go does not suggest err case first. Go suggest `smaller` and `less interesting` case first as it's easier to read. Usually error is, but not in this case.

storage/fanout.go
csmarchbanks: The [Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow) specifies trying to keep the normal path at minimal indentation, and handling errors first. If you still think I should change it I am happy to though! 

In this case I can see either way, I kind of like that we handle warnings and errors together, but the error handling is complicated enough that I understand trying to un-indent it.

csmarchbanks: I.e. the suggested flow ends up with:
```
handle warnings
handle success case
handle error case (possibly by adding a warning)
```
which is a bit strange.

roidelapluie: You are correct. I did not take context into account.

roidelapluie: @bwplotka Do you agree with merging the PR as it? What @csmarchbanks makes sense, too.

roidelapluie:
For the rc.1 let's take the PR as it as the latest comments of @csmarchbanks make sense to me. We can still refactor in master if someone really objects.

roidelapluie:
Thanks!

PR #6980, opened 2020-03-13
Make sure deleted intervals are excluded from Seek

gouthamve:
Before in the deletedIterator, we were returning deleted data when `Seek(); At()` were called. This fixes that.

The test `TestDeletedIterator_WithSeek` was just a copy of `TestDeletedIterator/Simple`. And the `TestDeletedIterator/Seek` cases were wrong (basically the bug I fixed was being ignored).

### DIFF
--- a/tsdb/querier.go
+++ b/tsdb/querier.go
@@ -1183,7 +1183,28 @@ func (it *deletedIterator) Seek(t int64) bool {
 	if it.it.Err() != nil {
 		return false
 	}
-	return it.it.Seek(t)
+	if ok := it.it.Seek(t); !ok {
+		return false
+	}
+
+	// Now double check if the entry falls into a deleted interval.
+	ts, _ := it.At()
+	for _, itv := range it.intervals {
+		if ts < itv.Mint {
+			return true
+		}
+
+		if ts > itv.Maxt {
+			it.intervals = it.intervals[1:]
+			continue
+		}
+
+		// We're in the middle of an interval, we can now call Next().
+		return it.Next()
+	}
+
+	// The timestamp is greater than all the deleted intervals.
+	return true
 }
 
 func (it *deletedIterator) Next() bool {
--- a/tsdb/querier_test.go
+++ b/tsdb/querier_test.go
@@ -234,7 +234,7 @@ func createIdxChkReaders(t *testing.T, tc []seriesSamples) (IndexReader, ChunkRe
 				app.Append(smpl.t, smpl.v)
 			}
 			chkReader[chunkRef] = chunk
-			chunkRef += 1
+			chunkRef++
 		}
 
 		ls := labels.FromMap(s.lset)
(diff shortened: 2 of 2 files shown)


tsdb/querier.go
```diff
@@ -1183,7 +1183,28 @@ func (it *deletedIterator) Seek(t int64) bool {
...
+
+	// Now double check if the entry falls into a deleted interval.
+	ts, _ := it.At()
+	for _, itv := range it.intervals {
+		if ts < itv.Mint {
+			return true
+		}
+
+		if ts > itv.Maxt {
+			it.intervals = it.intervals[1:]
```
krasi-georgiev: shouldn't this use the loop index?
```suggestion
			it.intervals = it.intervals[i:]
```

tsdb/querier.go
```diff
@@ -1183,7 +1183,28 @@ func (it *deletedIterator) Seek(t int64) bool {
...
+			continue
+		}
+
+		// We're in the middle of an interval, we can now call Next().
+		return it.Next()
```
krasi-georgiev: why is this call here?

tsdb/querier_test.go
```diff
@@ -1338,55 +1301,36 @@ func TestDeletedIterator_WithSeek(t *testing.T) {
...
-		{r: tombstones.Intervals{{Mint: 1, Maxt: 10}, {Mint: 12, Maxt: 20}, {Mint: 21, Maxt: 23}, {Mint: 25, Maxt: 30}}},
-		{r: tombstones.Intervals{{Mint: 1, Maxt: 10}, {Mint: 12, Maxt: 20}, {Mint: 20, Maxt: 30}}},
-		{r: tombstones.Intervals{{Mint: 1, Maxt: 10}, {Mint: 12, Maxt: 23}, {Mint: 25, Maxt: 30}}},
-		{r: tombstones.Intervals{{Mint: 1, Maxt: 23}, {Mint: 12, Maxt: 20}, {Mint: 25, Maxt: 30}}},
-		{r: tombstones.Intervals{{Mint: 1, Maxt: 23}, {Mint: 12, Maxt: 20}, {Mint: 25, Maxt: 3000}}},
-		{r: tombstones.Intervals{{Mint: 0, Maxt: 2000}}},
-		{r: tombstones.Intervals{{Mint: 500, Maxt: 2000}}},
-		{r: tombstones.Intervals{{Mint: 0, Maxt: 200}}},
-		{r: tombstones.Intervals{{Mint: 1000, Maxt: 20000}}},
+		{r: tombstones.Intervals{{Mint: 1, Maxt: 20}}, seek: 1, ok: true, seekedTs: 21},
```
krasi-georgiev: hm , so you seek ts 1 which exists and the result is ts 21? I don't get it.

tsdb/querier.go
gouthamve: That'll mess things up. See: https://play.golang.org/p/9NXnEvXqp2K

Here we're removing one element at a time.

tsdb/querier.go
gouthamve: We're in the middle of a deleted interval, and we need to get to the next valid timestamp. Calling `Next()` is the best thing I saw.

tsdb/querier_test.go
gouthamve: Because `[1, 20]` were deleted. Tombstones are inclusive on both ends.

tsdb/querier.go
codesome: Another confirmation, we are removing the elements from `it.intervals` here because we never need the old intervals as iteration is never reset?

gouthamve: Yes, you can never seek back.

PR #6981, opened 2020-03-13
Fix panic in memSeries.minTime

### LINKED ISSUE #6977: Isolation: Low Watermark stuck
During the 2.17.0-rc.0 Prombench, the low watermark got stuck, and there was a OOM situation:

oom
oom2
http://prombench.prometheus.io/grafana/d/7gmLoNDmz/prombench?orgId=1&var-RuleGroup=All&var-pr-number=6966&from=now-2d&to=now

beorn7:
Partially fixes #6977.

@codesome perhaps you have an idea why nill chunks could happen and what to do about it.

@roidelapluie in any case, I think this is a conservative approach to at least prevent the panic.

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -1762,7 +1762,9 @@ func newMemSeries(lset labels.Labels, id uint64, chunkRange int64) *memSeries {
 }
 
 func (s *memSeries) minTime() int64 {
-	if len(s.chunks) == 0 {
+	// TODO(beorn7): s.chunks[0] == nil has been observed in the wild but it
+	// is not clear if its occurrence is regular or a bug.
+	if len(s.chunks) == 0 || s.chunks[0] == nil {
 		return math.MinInt64
 	}
 	return s.chunks[0].minTime


roidelapluie:
I would like some input from @codesome . It would be good to know if there are reasonable tests that can be added.

roidelapluie:
Is that a 2.17 regression?

beorn7:
> Is that a 2.17 regression?

I don't know. I guess the same could have happened before, but it wasn't a big deal because there was no low watermark to get stuck. It would have just leaked a few resources, at worst. And as we have seen with the test runs so far, this seems to happen very rarely.

roidelapluie:
[ Without a test we can't assume that it won't panic elsewhere. ]

brian-brazil:
I'm pretty sure the nil is a bug.

beorn7:
I'd say we shouldn't mask a bug.
With #6983 in, at least the panic should not crate a stuck watermark.

But the bug is kind of serious, then. If it weren't in the HTTP serving call stack, it would crash the whole binary.

roidelapluie:
That bug is highlighted by recent optimizations made by Brian and me.

tsdb/head.go
```diff
@@ -1762,7 +1762,9 @@ func newMemSeries(lset labels.Labels, id uint64, chunkRange int64) *memSeries {
 }
 
 func (s *memSeries) minTime() int64 {
-	if len(s.chunks) == 0 {
+	// TODO(beorn7): s.chunks[0] == nil has been observed in the wild but it
+	// is not clear if its occurrence is regular or a bug.
+	if len(s.chunks) == 0 || s.chunks[0] == nil {
```
codesome: This can never be `nil`. There is only 1 place where chunks are added to the slice https://github.com/prometheus/prometheus/blob/4d6f9b75e230465443391ca76a4182b130eb99e9/tsdb/head.go#L1780-L1785

As you said, it won't be good to mask this bug. I will take a deeper look at this (and also isolation work in general) on Monday.

roidelapluie: We also shrink that. I am more thinking about the fact that we might need some locking about s.chunks

https://github.com/prometheus/prometheus/blob/4d6f9b75e230465443391ca76a4182b130eb99e9/tsdb/head.go#L1840-L1848

roidelapluie: See full issue #6977

codesome: Yup, but shrinking also doesn't seem to introduce `nil` chunks. And about locking, we have locks around the series itself.

roidelapluie:
Superseeded by #6985; closing this one to limit the places of discussion.

PR #6983, opened 2020-03-13
TSDB: Extract parts out of populateSeries

beorn7:
This addresses fabxc's TODO.

More importantly, it now properly defers the
querier.Close(). Previously, if a panic happened after creation of the
querier within the populateSeries function, querier.Close() was never called.

The latter was responsible for #6977.


brian-brazil:
:+1:

Issue #6979, opened 2020-03-13
WAL can't correctly handle more than 1 million segments


brian-brazil:
The segments are created in files with names following %08d, and checkpoints %06d. We take a  list of checkpoints/files sorted as strings, but that's not going to work once there's more than 1M segments. We should do our own sorting based on the number, and also use consistent formatting between checkpoints and segments.

We're also using ioutil.ReadDir in one place and our own fileutil.Readdir on top of Readdirnames in another, we should unify this on ioutil.Readdir rather than inventing our own thing.

codesome:
Closed by https://github.com/prometheus/prometheus/pull/6987 and created https://github.com/prometheus/prometheus/issues/7029 for `fileutil.Readdir`

PR #6987, opened 2020-03-15
tsdb: sort checkpoints by segment number

zhulongcheng:
Updates #6979

### DIFF
--- a/tsdb/wal/checkpoint.go
+++ b/tsdb/wal/checkpoint.go
@@ -21,6 +21,7 @@ import (
 	"math"
 	"os"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


brian-brazil:
This only covers checkpoints, segments need fixing too.

zhulongcheng:
> This only covers checkpoints, segments need fixing too.

WAL segments have fixed by `listSegments `:
https://github.com/prometheus/prometheus/blob/4d6f9b75e230465443391ca76a4182b130eb99e9/tsdb/wal/wal.go#L761

brian-brazil:
:+1: 

So it is.

zhulongcheng:
thanks @cstyan 

need `FileInfo`  to know if it is a dir here.

do you think it is better to refactor `fileutil.Readdir` in a new single PR?

Issue #6988, opened 2020-03-15
"promtool update" is missing


RalfJung:
**What did you do?** We did a dist-upgrade of our Debian machine, which involves migrating to Prometheus 2.0. We are getting obscure error messages that, I think, are about our old rules not being in the new format. According to [the latest migration docs](https://prometheus.io/docs/prometheus/latest/migration/), we can run `promtool update` to convert our rules.

**What did you expect to see?** `promtool update` should do what it says it does in the latest documentation.

**What did you see instead? Under which circumstances?**

```
$ sudo promtool update rules summed_traffic.rules 
promtool: error: expected command but got "update", try --help
```

**Environment**

* System information: Linux 4.19.0-8-amd64 x86_64

* Prometheus version:

```
prometheus, version 2.7.1+ds (branch: debian/sid, revision: 2.7.1+ds-3+b11)
  build user:       pkg-go-maintainers@lists.alioth.debian.org
  build date:       20190608-09:54:04
  go version:       go1.11.6
```

hoffie:
Thanks for the report.

`promtool update` has been removed as part of https://github.com/prometheus/prometheus/pull/3839. Prometheus 2.5.0 seems to be the last release, which still contained the subcommand -- the 2.6.0 release notes [mention the removal](https://github.com/prometheus/prometheus/releases/tag/v2.6.0).
I agree that the docs are confusing. Apparently it still makes sense to keep a notice regarding rule conversion. I'll open a PR to extend the docs a bit.

In your case, I suggest getting Prometheus 2.5.0 and using promtool from there to convert your rules. I don't think there were any incompatible changes afterwards, so this should work flawlessly. Also, you don't need to install anything.

If Debian did not provide an intermediate version between 1.x and 2.7, then you might want to take this issue up with the maintainers of the Debian package to provide extended update instructions.

roidelapluie:
It makes more sense to ask questions like this on the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users) rather than in a GitHub issue. On the mailing list, more people are available to potentially respond to your question, and the whole community can benefit from the answers provided.

RalfJung:
> If Debian did not provide an intermediate version between 1.x and 2.7

Nope, [oldstable is on 1.5 and stable is on 2.7](https://tracker.debian.org/pkg/prometheus).

[...]

PR #6989, opened 2020-03-15
docs: add promtool version in rule migration cmd

hoffie:
`promtool update rules` is no longer available after Prometheus 2.5.
Document this to avoid confusion such as in #6988.


roidelapluie:
Thanks! Can we be shorter and more pragmatic?

hoffie:
Tried to condense it into a single sentence. If you think it should be shortened even further, we would need to drop details, I think.

roidelapluie:
Thanks!

PR #6990, opened 2020-03-16
Added Chunks*Iterator/Querier/Selector interfaces and implementations + More reuse!

bwplotka:
Finally!! :muscle: :heart: 

Sorry for the huge PR, but it's really hard to avoid. Even added some TODOs for further simplifications in the next PRs to avoid having it even bigger.

TODO: Fix/add more unit tests (: 

* Added Chunk versions of all iterating methods. It all starts in Querier/ChunkQuerier. The plan is that
Storage will implement both chunked and samples.
* Added Seek to chunks.Iterator interface for iterating over chunks.
* Mock, NewTestSeries, SampleSeriesIterator, and ChunkSeriesIterator are now available from storage package and reuses instead of
being recreated in many places. NewConcreteSeries was created to replace concreteSeries.
* NewMergeChunkQuerier was added; Both this and NewMergeQuerier are now using generigMergeQuerier to share the code. A generic code was added.
* Both Compactor and block Querier use *exactly* the same iterators. (blockChunkSeriesSet).

### DIFF
--- a/storage/fanout.go
+++ b/storage/fanout.go
@@ -16,6 +16,7 @@ package storage
 import (
 	"container/heap"
 	"context"
+	"math"
 	"sort"
 	"strings"
 
@@ -25,6 +26,8 @@ import (
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/tsdb/chunkenc"
+	"github.com/prometheus/prometheus/tsdb/chunks"
+	tsdb_errors "github.com/prometheus/prometheus/tsdb/errors"
 )
 
 type fanout struct {
@@ -47,7 +50,7 @@ func NewFanout(logger log.Logger, primary Storage, secondaries ...Storage) Stora
 // StartTime implements the Storage interface.
 func (f *fanout) StartTime() (int64, error) {
 	// StartTime of a fanout should be the earliest StartTime of all its storages,
-	// both primary and secondaries.
+	// both primaryQuerier and secondaries.
 	firstTime, err := f.primary.StartTime()
 	if err != nil {
 		return int64(model.Latest), err
@@ -68,24 +71,24 @@ func (f *fanout) StartTime() (int64, error) {
 func (f *fanout) Querier(ctx context.Context, mint, maxt int64) (Querier, error) {
 	queriers := make([]Querier, 0, 1+len(f.secondaries))
 
-	// Add primary querier
+	// Add primaryQuerier querier.
 	primaryQuerier, err := f.primary.Querier(ctx, mint, maxt)
 	if err != nil {
 		return nil, err
 	}
 	queriers = append(queriers, primaryQuerier)
 
-	// Add secondary queriers
+	// Add secondary queriers.
 	for _, storage := range f.secondaries {
 		querier, err := storage.Querier(ctx, mint, maxt)
 		if err != nil {
-			NewMergeQuerier(primaryQuerier, queriers).Close()
+			NewMergeQuerier(primaryQuerier, queriers, VerticalSeriesMergeFunc(ChainedSeriesMerge)).Close()
 			return nil, err
 		}
 		queriers = append(queriers, querier)
 	}
 
-	return NewMergeQuerier(primaryQuerier, queriers), nil
+	return NewMergeQuerier(primaryQuerier, queriers, VerticalSeriesMergeFunc(ChainedSeriesMerge)), nil
 }
 
 func (f *fanout) Appender() Appender {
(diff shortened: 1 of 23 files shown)


roidelapluie:
Would be nice to wait for stable to go out before merging this so we can easily backport tsdb fixes into master.

storage/interface.go
```diff
@@ -51,19 +51,41 @@ type Storage interface {
...
 type Queryable interface {
 	// Querier returns a new Querier on the storage.
 	Querier(ctx context.Context, mint, maxt int64) (Querier, error)
 }
 
-// Querier provides querying access over time series data of a fixed
-// time range.
+// A ChunkQueryable handles queries against a storage.
+// Use it when you need to have access to samples in encoded format. This is usually more efficient when asking
+// for data older than 2h. e.g remote read.
```
brian-brazil: It's not about age, the Head block is chunks bar the last few samples.

tsdb/chunks/chunks.go
```diff
@@ -242,84 +256,85 @@ func (w *Writer) write(b []byte) error {
...
-	if ait.Err() != nil {
-		return nil, ait.Err()
-	}
-	if bit.Err() != nil {
-		return nil, bit.Err()
-	}
-	return newChunk, nil
-}
+//
+//// MergeOverlappingChunks removes the samples whose timestamp is overlapping.
```
brian-brazil: Remove rather than commenting out.

tsdb/tsdbutil/chunks.go
```diff
@@ -18,23 +18,37 @@ import (
 	"github.com/prometheus/prometheus/tsdb/chunks"
 )
 
+type Samples interface {
```
brian-brazil: Why are you adding this? Using a slice directly would be simpler.

storage/remote/codec.go
```diff
@@ -188,8 +192,83 @@ func NegotiateResponseType(accepted []prompb.ReadRequest_ResponseType) (prompb.R
...
+		}
+	}
+	if err := ss.Err(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// TODO(bwplotka): Delete. For benchmarking purposes only.
```
brian-brazil: Leaving this here makes it really hard to figure out the diff

tsdb/compact.go
```diff
@@ -730,102 +726,43 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
-			} else
-			// Sanity check for disk blocks.
-			// chk.MaxTime == meta.MaxTime shouldn't happen as well, but will brake many users so not checking for that.
-			if chk.MinTime < meta.MinTime || chk.MaxTime > meta.MaxTime {
-				return errors.Errorf("found chunk with minTime: %d maxTime: %d outside of compacted minTime: %d maxTime: %d",
-					chk.MinTime, chk.MaxTime, meta.MinTime, meta.MaxTime)
-			}
-
-			if len(dranges) > 0 {
-				// Re-encode the chunk to not have deleted values.
```
brian-brazil: Where are deleted ranges being handled now?

storage/series.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/prometheus/prometheus/tsdb/chunks"
+	"github.com/prometheus/prometheus/tsdb/tsdbutil"
+)
+
+type ConcreteSeries struct {
+	labels           labels.Labels
+	SampleIteratorFn func() chunkenc.Iterator
+}
+
+func NewTestSeries(lset labels.Labels, samples ...[]tsdbutil.Sample) *ConcreteSeries {
```
brian-brazil: This sounds like it belongs in a test file

storage/fanout.go
```diff
@@ -590,22 +676,164 @@ func (c *mergeIterator) Err() error {
...
+	old := *h
+	n := len(old)
+	x := old[n-1]
+	*h = old[0 : n-1]
+	return x
+}
+
+// VerticalChunkMergeFunc represents a function that merges multiple time overlapping chunks.
+// Passed chunks:
+// * have to be sorted by MinTime and should be
```
brian-brazil: and should be....?

storage/series.go
bwplotka: Yes, but TBH we would like to have just one  implementation of this and import from other tests, instead of replicating it everywhere. Any thoughts on how to resolve this?

tsdb/compact.go
bwplotka: Everything is handled by `populateWithTombstonesChunkIterator`

storage/remote/codec.go
bwplotka: True, will delete

tsdb/tsdbutil/chunks.go
bwplotka: Hm hm, I copied it over

bwplotka:
Yea @roidelapluie agree with you.. it's also far to be merged. (: 

@brian-brazil I will explain, but I kind of consider splitting PR even more... maybe adding interfaces first, then implementations?

storage/series.go
brian-brazil: I'm not seeing the problem here

krasi-georgiev: I guess it is the cycling dependancy? Try to put it in a package called `package storage_test` and than import in the actual tests

bwplotka: You cannot import anything from other packages's `_test.go` file

bwplotka:
Ok, sorry for the noise. Will split this into smaller pieces - it's too huge even for me to address comments. It will take ages (:

PR #6993, opened 2020-03-16
Fix tsdb to use errors from the storage package

### LINKED ISSUE #6992: 2.17.0-rc.1: err="unknown series:  not found"
**What did you do?**

Run Prombench

2.17 can scrape the same targets as 2.16

2.17 sometimes fails with:
level=debug ts=2020-03-16T19:18:07.614Z caller=scrape.go:968 component="scrape manager" 
                         scrape_pool=fake-webservers-12 target=http://10.16.50.105:8081/metrics msg="append failed" err="unknown series: 
                         not found"

**Environment**

* System information:
http://prombench.prometheus.io/6991/prometheus-pr/graph?g0.range_input=2h&g0.expr=count_values(%22up%22%2Cup)&g0.tab=0
http://prombench.prometheus.io/grafana/d/7gmLoNDmz/prombench?orgId=1&var-RuleGroup=All&var-pr-number=6991&from=1584377715743&to=1584387589755
samp

roidelapluie:
Make TSDB use storage errors

    This fixes #6992, which was introduced by #6777. There was an
    intermediate component which translated TSDB errors into storage errors,
    but that component was deleted and this bug went unnoticed, until we
    were watching at the Prombench results. Without this, scrape will fail
    instead of dropping samples or using "Add" when the series have been
    garbage collected.

Fixes #6992

https://play.golang.org/p/JGcqpSuJ5wG

### DIFF
--- a/tsdb/querier.go
+++ b/tsdb/querier.go
@@ -739,7 +739,7 @@ func (s *baseChunkSeries) Next() bool {
 		ref := s.p.At()
 		if err := s.index.Series(ref, &lset, &chkMetas); err != nil {
 			// Postings may be stale. Skip if no underlying series exists.
-			if errors.Cause(err) == ErrNotFound {
+			if errors.Cause(err) == storage.ErrNotFound {
 				continue
 			}
 			s.err = err
@@ -819,7 +819,7 @@ func (s *populatedChunkSeries) Next() bool {
 			c.Chunk, s.err = s.chunks.Chunk(c.Ref)
 			if s.err != nil {
 				// This means that the chunk has be garbage collected. Remove it from the list.
-				if s.err == ErrNotFound {
+				if s.err == storage.ErrNotFound {
 					s.err = nil
 					// Delete in-place.
 					s.chks = append(chks[:j], chks[j+1:]...)
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -599,7 +599,7 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
 			}()
 			for _, s := range vector {
 				if _, err := app.Add(s.Metric, s.T, s.V); err != nil {
-					switch err {
+					switch errors.Cause(err) {
 					case storage.ErrOutOfOrderSample:
 						numOutOfOrder++
 						level.Debug(g.logger).Log("msg", "Rule evaluation result discarded", "err", err, "sample", s)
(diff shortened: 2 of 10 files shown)


roidelapluie:
cc @beorn7

roidelapluie:
This was fun. Who wants to dig to see which releases of Prometheus are impacted? also with the other errors (out of bounds etc)?

beorn7:
:confused: How could that have worked ever before?

beorn7:
More `tsdb.ErrorNotFound` used, says the compiler…

roidelapluie:
cc @bwplotka

roidelapluie:
Oh are they the same time? Just fixing the import might be enough

roidelapluie:
Yes. In storage we have go "errors" and in tsdb "github.com/pkg/errors"

brancz:
I'm not sure the title of this PR accurately reflects the content, I'm only saying this because we often use the PR title as the changelog item, although in this case I guess we're just going to skip it as this was introduced and fixed in 1.17?

roidelapluie:
It won't make it in the changelog. But still I will fix it. What should be
the fix? Using errors grin storage package and/or fixing storage to use
github/pkg/errors ?

brancz:
“Fix storage to use /pkg/errors appropriately” is what I think reflects it best.

roidelapluie:
mmh actually this does not seem to be the issue. storage always used errors.

tsdb/querier.go
```diff
@@ -819,7 +819,7 @@ func (s *populatedChunkSeries) Next() bool {
 			c.Chunk, s.err = s.chunks.Chunk(c.Ref)
 			if s.err != nil {
 				// This means that the chunk has be garbage collected. Remove it from the list.
-				if s.err == ErrNotFound {
+				if s.err == storage.ErrNotFound {
```
bwplotka: I think all of this should be `errors.Cause`, so we are safe on potential wraps someday.

roidelapluie:
But why did it stop working?

roidelapluie:
Also errors.Cause on scrape would need some refactoring as we are in a `switch`

beorn7:
Note that each call to `errors.New` creates a new error that is not equal to another error created with `errors.New`, even if it uses the same string.

https://play.golang.org/p/sUQ_apcRvl8

roidelapluie:
Why the AddFast -> not found -> Add  logic work before? That is what I don't get. I am not willing to merge until I understand, and I will dig into this tonight.

roidelapluie:
https://github.com/prometheus/prometheus/blob/release-2.16/storage/tsdb/tsdb.go#L306-L320

The removal of this cause is causing this issue

tsdb/querier.go
roidelapluie: :+1:

roidelapluie:
Are we ok with the small CPU increase? otherwise we can remove the error Wrap and errors.Cause and start another benchmark:

err

brancz:
It doesn't look significant enough to me to even be certain that it's a real increase. Even if then it seems small enough that I'd be fine with it.

bboreham:
Somewhat reminiscent of #4628

PR #6995, opened 2020-03-17
Optimize away label=~".*" matchers.

pstibrany:
These are useless, and just add extra work. By removing them from AST (instead of doing this optimization at TSDB layer), this also automatically applies to other engines that don't use TSDB backend.

### DIFF
--- a/promql/parser/parse.go
+++ b/promql/parser/parse.go
@@ -608,6 +608,16 @@ func (p *parser) checkAST(node Node) (typ ValueType) {
 			}
 		}
 
+		// remove label matchers doing regex match on ".*" -- that matches everything, incl. nothing, so it's a no-op
+		for ix := 0; ix < len(n.LabelMatchers); ix++ {
+			m := n.LabelMatchers[ix]
+			if m.Type == labels.MatchRegexp && m.Value == ".*" {
+				// There must be some other matcher matching non-empty string as well,
+				// so we cannot end up in a situation that this was the last matcher and it was just removed.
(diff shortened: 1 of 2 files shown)


roidelapluie:
I don't think this optimization belongs to the parser. It will also have side effects on the language server.

brian-brazil:
Yeah, this belongs down in TSDB. Technically it's a breaking change for remote read otherwise.

bwplotka:
Why it is braking change?

bwplotka:
Anyway, it would be so nice to have this! Happy for this to be on TSDB - this might mean separate PR to Thanos as well (: 


Thanks @pstibrany for focusing on this!

pstibrany:
I'll try to send this optimization in TSDB layer, as soon as I get tests passing ;-)

brian-brazil:
> Why it is braking change?

A remote read endpoint may depend on this, not all of them are storage.

bwplotka:
> > Why it is braking change?

> A remote read endpoint may depend on this, not all of them are storage.

Good point, totally true - missed that. Thanks!

codesome:
Would the query engine be a fine place to have such optimizations?

roidelapluie:
> Would the query engine be a fine place to have such optimizations?

That would be something for a query planner

brian-brazil:
The problem is that remote storage is under no obligation to return samples that match the selector passed to them, so that makes any semantic query optimisation very challenging.

PR #6996, opened 2020-03-17
Optimize away label=~".*" matchers in TSDB layer.

pstibrany:
This is alternative to #6995, except it removes these useless matchers in TSDB layer.

Benchmarks in progress, will update the post when finished.

Update: well, first benchmarks show difference where you'd expect -- when using multiple matchers, one of them being `~= ".*"`. I need to let them run longer, and they block my computer, so I'll do that later.

### DIFF
--- a/tsdb/querier.go
+++ b/tsdb/querier.go
@@ -314,6 +314,13 @@ func PostingsForMatchers(ix IndexReader, ms ...*labels.Matcher) (index.Postings,
 	}
 
 	for _, m := range ms {
+		if m.Type == labels.MatchRegexp && (m.Value == ".*" || m.Value == "^.*$") && len(ms) > 1 {
+			// Ignore this matcher completely. This matches anything, so it's a no-op.
+			// It's safe to ignore, because there must be some matcher matching non-empty string.
+			// Some tests only use single label=~".*" matcher, and for those, we include the length condition.
+			continue
+		}
+
 		if labelMustBeSet[m.Name] {
 			// If this matcher must be non-empty, we can be smarter.
 			matchesEmpty := m.Matches("")
--- a/tsdb/querier.go
+++ b/tsdb/querier.go
@@ -315,10 +315,15 @@ func PostingsForMatchers(ix IndexReader, ms ...*labels.Matcher) (index.Postings,
 
 	for _, m := range ms {
 		if m.Type == labels.MatchRegexp && (m.Value == ".*" || m.Value == "^.*$") && len(ms) > 1 {
-			// Ignore this matcher completely. This matches anything, so it's a no-op.
+			// Ignore this matcher completely. This matches any value, including no value, so it's a no-op.
 			// It's safe to ignore, because there must be some matcher matching non-empty string.
 			// Some tests only use single label=~".*" matcher, and for those, we include the length condition.
 			continue
+		}
+
+		// NOT matching any value = match nothing. We can shortcut this easily.
+		if m.Type == labels.MatchNotRegexp && (m.Value == ".*" || m.Value == "^.*$") {
+			return index.EmptyPostings(), nil
 		}
 
 		if labelMustBeSet[m.Name] {
--- a/tsdb/querier.go
+++ b/tsdb/querier.go
@@ -313,11 +313,12 @@ func PostingsForMatchers(ix IndexReader, ms ...*labels.Matcher) (index.Postings,
 		}
 	}
 
+	all := false
+
 	for _, m := range ms {
-		if m.Type == labels.MatchRegexp && (m.Value == ".*" || m.Value == "^.*$") && len(ms) > 1 {
-			// Ignore this matcher completely. This matches any value, including no value, so it's a no-op.
-			// It's safe to ignore, because there must be some matcher matching non-empty string.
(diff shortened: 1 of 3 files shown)


roidelapluie:
Can we also tackle the case `!~".*"` ?

pstibrany:
> Can we also tackle the case `!~".*"` ?

By returning empty set, and thus postings? Yes, I think we can.

pstibrany:
I did not add a new test for this, since there is already a test that triggered this branch here: https://github.com/prometheus/prometheus/blob/fe802f29c99d76be27bf1b1e226ebb447fab4627/tsdb/querier_test.go#L2076-L2075

tsdb/querier.go
```diff
@@ -314,6 +314,18 @@ func PostingsForMatchers(ix IndexReader, ms ...*labels.Matcher) (index.Postings,
 	}
 
 	for _, m := range ms {
+		if m.Type == labels.MatchRegexp && (m.Value == ".*" || m.Value == "^.*$") && len(ms) > 1 {
```
brian-brazil: Does this work given how we wrap regexes?

tsdb/querier.go
```diff
@@ -314,6 +314,18 @@ func PostingsForMatchers(ix IndexReader, ms ...*labels.Matcher) (index.Postings,
...
+			// Ignore this matcher completely. This matches any value, including no value, so it's a no-op.
+			// It's safe to ignore, because there must be some matcher matching non-empty string.
+			// Some tests only use single label=~".*" matcher, and for those, we include the length condition.
```
brian-brazil: It's not just about tests, it's about correctness. Can we fast path those?

tsdb/querier.go
pstibrany: I believe so. This is value from the promql. We only wrap regexes in `m.re` field.

tsdb/querier.go
pstibrany: You're right. I've modified the code to use all postings if we find this kind of matcher. We only return all postings, if there are no other positive matchers.

tsdb/querier.go
```diff
@@ -365,8 +378,14 @@ func PostingsForMatchers(ix IndexReader, ms ...*labels.Matcher) (index.Postings,
 		}
 	}
 
-	// If there's nothing to subtract from, add in everything and remove the notIts later.
-	if len(its) == 0 && len(notIts) != 0 {
+	// When doing substraction, add in everything and remove the notIts later.
```
brian-brazil: subtraction

brian-brazil:
We've had regressions before for things like this, so we really need a test.

pstibrany:
I've added more test cases covering the shortcuts, and verified that they work with old version of `PostingsForMatchers`, as well as with new shortcuts.



> We've had regressions before for things like this, so we really need a test.

I'm not quite sure how to approach this test. Any suggestions?

brian-brazil:
See #6540

pstibrany:
> See #6540

Do you suggest to move these shortcuts into a separate function to make it more testable? I'd prefer to avoid doing that, as it would the complicate the code. Given that, I'm out of ideas how to test it in a way that you're suggesting.

brian-brazil:
That's how we're doing existing tests of this sort of thing, so I hoped it'd provide some inspiration. I mainly care that it is tested, so any reasonable approach.

pstibrany:
> That's how we're doing existing tests of this sort of thing, so I hoped it'd provide some inspiration. I mainly care that it is tested, so any reasonable approach.

I think I finally have a good way to test these shortcuts -- I simply check if returned posting is empty or "all", or expected label postings match. PTAL.

pstibrany:
Tests are failing in `react-app-test` step, seems unrelated to my changes. (Perhaps I shouldn't have rebased on top of master)

pstibrany:
Is there more I should do here, or does the test look good now?

brian-brazil:
Thinking on this there, this optimisation isn't correct at all as `.` doesn't match newline by default in Go.

pstibrany:
> Thinking on this there, this optimisation isn't correct at all as `.` doesn't match newline by default in Go.

That's a valid point. Since it is not very typical to have newlines in label values, and most people mean to match everything when they write `.*`, would it make sense to make a breaking change and let `.` match newline too? (By using `"^(?s:" + v + ")$"` for anchoring)

But I assume that such discussion has already happened in Prometheus before, and the answer is No.

beorn7:
It's even remotely plausible that somebody uses `foo=~".*"` to filter out labels with newlines in it. :cry:

brian-brazil:
That would be a breaking change to PromQL, and thus not something that's doable in the current major version of Prometheus. We'd also have to change how regexes across the other repos, and for the SNMP exporter at least newlines in strings have come up.

pstibrany:
You're right. 😞

pstibrany:
Thanks for your review and feedback!

roidelapluie:
> Thinking on this there, this optimisation isn't correct at all as `.` doesn't match newline by default in Go.

Should we have unit tests around this?

roidelapluie:
Note: We have already got requests from users which use new lines in labels, so that is a thing.

Issue #6997, opened 2020-03-17
Fatal error after reload when configuration got changed (increasing sample_limit)


hawran:
**What did you do?**
A configuration has been changed:
`sample_limit` from `100000` to `130000`
then the new configuration has been deployed to kubernetes.

**What did you expect to see?**
A configuration reload as usual.

**What did you see instead? Under which circumstances?**
Prometheus got restarted with a bunch of errors within logs.

**Environment**

* System information:
```
 Kernel Version:                  5.3.0-24-generic
 OS Image:                        Ubuntu 18.04.1 LTS
 Operating System:                linux
 Architecture:                    amd64
 Kubelet Version:                 v1.14.8
```

* Prometheus version:
2.16.0

* Logs:
```
fatal error: concurrent map read and map write

goroutine 7926 [running]:
runtime.throw(0x25d4f3b, 0x21)
	/nix/store/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee-go-1.12.16/share/go/src/runtime/panic.go:617 +0x72 fp=0xc0248e16c0 sp=0xc0248e1690 pc=0x42f552
runtime.mapaccess2_fast64(0x217f780, 0xc21aaea3c0, 0x329a094c9e427070, 0x45dea01, 0xc256bc5d00)
	/nix/store/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee-go-1.12.16/share/go/src/runtime/map_fast64.go:61 +0x1c2 fp=0xc0248e16e8 sp=0xc0248e16c0 pc=0x412ed2
github.com/prometheus/prometheus/scrape.(*scrapeCache).forEachStale(0xc21aaf0100, 0xc0248e1a10)
	/build/go/src/github.com/prometheus/prometheus/scrape/scrape.go:766 +0xdf fp=0xc0248e1798 sp=0xc0248e16e8 pc=0x1d7fbff
github.com/prometheus/prometheus/scrape.(*scrapeLoop).append(0xc0f1760240, 0x45dd750, 0x0, 0x0, 0x0, 0x0, 0xbf94458d1f2e81b8, 0x6181c1af2c1ca, 0x45bf9e0, 0x0, ...)
	/build/go/src/github.com/prometheus/prometheus/scrape/scrape.go:1219 +0xce5 fp=0xc0248e1a48 sp=0xc0248e1798 pc=0x1d83825
[...]

PR #6998, opened 2020-03-17
Fix bug with WAL watcher and Live Reader metrics usage.

cstyan:
I noticed then when looking into the pending samples issue again.

Calling NewXMetrics when creating a Watcher or LiveReader results in a
registration error, which we're ignoring, and as a result other than the
first Watcher/Reader created, we had no metrics for either. So we would
only have metrics like Watcher Records Read for the first remote write
config in a users config file.

cc @roidelapluie it would be good to get this in the 2.17 release if possible

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -271,7 +271,7 @@ type QueueManager struct {
 }
 
 // NewQueueManager builds a new QueueManager.
-func NewQueueManager(reg prometheus.Registerer, metrics *queueManagerMetrics, watcherMetrics *wal.WatcherMetrics, readerMetrics *wal.LiveReaderMetrics, logger log.Logger, walDir string, samplesIn *ewmaRate, cfg config.QueueConfig, externalLabels labels.Labels, relabelConfigs []*relabel.Config, client StorageClient, flushDeadline time.Duration) *QueueManager {
+func NewQueueManager(metrics *queueManagerMetrics, watcherMetrics *wal.WatcherMetrics, readerMetrics *wal.LiveReaderMetrics, logger log.Logger, walDir string, samplesIn *ewmaRate, cfg config.QueueConfig, externalLabels labels.Labels, relabelConfigs []*relabel.Config, client StorageClient, flushDeadline time.Duration) *QueueManager {
 	if logger == nil {
 		logger = log.NewNopLogger()
 	}
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -271,7 +271,7 @@ type QueueManager struct {
 }
 
 // NewQueueManager builds a new QueueManager.
-func NewQueueManager(reg prometheus.Registerer, metrics *queueManagerMetrics, logger log.Logger, walDir string, samplesIn *ewmaRate, cfg config.QueueConfig, externalLabels labels.Labels, relabelConfigs []*relabel.Config, client StorageClient, flushDeadline time.Duration) *QueueManager {
+func NewQueueManager(reg prometheus.Registerer, metrics *queueManagerMetrics, watcherMetrics *wal.WatcherMetrics, readerMetrics *wal.LiveReaderMetrics, logger log.Logger, walDir string, samplesIn *ewmaRate, cfg config.QueueConfig, externalLabels labels.Labels, relabelConfigs []*relabel.Config, client StorageClient, flushDeadline time.Duration) *QueueManager {
 	if logger == nil {
 		logger = log.NewNopLogger()
 	}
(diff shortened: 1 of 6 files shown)


tsdb/wal/watcher.go
```diff
@@ -125,25 +125,25 @@ func NewWatcherMetrics(reg prometheus.Registerer) *WatcherMetrics {
 	}
 
 	if reg != nil {
-		_ = reg.Register(m.recordsRead)
-		_ = reg.Register(m.recordDecodeFails)
-		_ = reg.Register(m.samplesSentPreTailing)
-		_ = reg.Register(m.currentSegment)
+		fmt.Println(reg.Register(m.recordsRead))
```
csmarchbanks: Debug Printlns?

roidelapluie:
Let's get this in 2.17.

roidelapluie:
Can you please provide a one-liner for changelog? e.g. the name of the metric(s) fixed.

cstyan:
@roidelapluie something like
> Fixed bug with registration of WAL watcher metrics that would result in metrics for any remote write queues other than the first to not be present in /metrics

Basically, we were trying to register metrics that already existed, swallowing that error, and then proceeding with nil metrics for anything except the first remote write queue in the config file.

cstyan:
Mirroring what I said in IRC, the bug would have affected at least all these metrics: https://github.com/prometheus/prometheus/blob/master/tsdb/wal/watcher.go#L87-L135

For all but the first remote write in the configuration file.

roidelapluie:
Closing to go to 2.17 directly, I can't change branches.

roidelapluie:
There are comments here: https://github.com/prometheus/prometheus/pull/7002

Can you address them and rebase the PR to release-2.17? I hope we can fix #6997 in a 2.17.1, together with this bug.

cstyan:
To reply to the comments from the other pr:

https://github.com/prometheus/prometheus/pull/7002#pullrequestreview-376882433 @beorn7 definitely we shouldn't ignore the errors, I'll use `MustRegister`. But the root of the bug here was calling the metrics creation function multiple times, when we should have been calling it once and passing a pointer to the metrics to each struct we created.

https://github.com/prometheus/prometheus/pull/7002#issuecomment-600643780 @bwplotka I thought we were avoiding using `promauto` in `prometheus/prometheus`?

beorn7:
> I thought we were avoiding using promauto in prometheus/prometheus?

`promauto` has learned a new trick. It can now deal with custom registries, too. @bwplotka is a great fan of it. I'm not against introducing it in Prometheus, but perhaps not in the context of a bug fix but as a separate step?

cstyan:
`promauto` is currently only used for two metrics, both in the remote storage package. I was planning to move those away from using `promauto` soon.

beorn7:
I guess that usage is still the "old" way of using `promauto` only with the global registry.

Wherever you create metrics just to `MustRegister` them directly after, you could as well use the new `promauto.With(registry).New...` pattern.

beorn7:
But as said, I don't think we should do any of those changes here in this PR.

storage/remote/queue_manager.go
```diff
@@ -271,7 +271,7 @@ type QueueManager struct {
 }
 
 // NewQueueManager builds a new QueueManager.
-func NewQueueManager(reg prometheus.Registerer, metrics *queueManagerMetrics, logger log.Logger, walDir string, samplesIn *ewmaRate, cfg config.QueueConfig, externalLabels labels.Labels, relabelConfigs []*relabel.Config, client StorageClient, flushDeadline time.Duration) *QueueManager {
+func NewQueueManager(reg prometheus.Registerer, metrics *queueManagerMetrics, watcherMetrics *wal.WatcherMetrics, readerMetrics *wal.LiveReaderMetrics, logger log.Logger, walDir string, samplesIn *ewmaRate, cfg config.QueueConfig, externalLabels labels.Labels, relabelConfigs []*relabel.Config, client StorageClient, flushDeadline time.Duration) *QueueManager {
```
beorn7: The `reg` argument isn't used anymore and can be removed here.

In different news, with the large number of arguments, I'd put each on its own line or do some meaningful grouping per line, but not all on one line.

storage/remote/write.go
```diff
@@ -65,13 +68,15 @@ func NewWriteStorage(logger log.Logger, reg prometheus.Registerer, walDir string
 		logger = log.NewNopLogger()
 	}
 	rws := &WriteStorage{
-		queues:        make(map[string]*QueueManager),
-		reg:           reg,
```
beorn7: Once `NewQueueManager` doesn't need the `reg` arg anymore, this can go from the struct, too.

cstyan:
Thanks for the reviews.

I think we're :+1: here now @roidelapluie, ready for another rc or 2.17.1 when you are. Let me know if you need any help.

PR #7006, opened 2020-03-18
release 2.17.0-rc.3

roidelapluie:
We need a 2.17.0-rc.3 because the docker images of 2.17.0-rc.2 were not published correctly. And relaunching the job creates new tarballs that them have different checksums so it is better to simply release a new RC.


simonpasquier:
> And relaunching the job creates new tarballs that them have different checksums so it is better to simply release a new RC.

You should have been able to delete the tarballs from the release page. Or am I missing something? Anyway that's not a big deal, I agree.

roidelapluie:
> > And relaunching the job creates new tarballs that them have different checksums so it is better to simply release a new RC.
> 
> You should have been able to delete the tarballs from the release page and relaunch the CI workflow. Or am I missing something? Anyway that's not a big deal, I agree.

I sent the announcement and published the github release already. I did not want two releases with different sha1 but same version to be published.

PR #7009, opened 2020-03-19
Merge release-2.17 into master

beorn7:
So that master can also benefit from all the sweet bug fixes.


beorn7:
The DCO check failure is spurious. Not sure how we can prevent that kind of failure mode.

roidelapluie:
> The DCO check failure is spurious. Not sure how we can prevent that kind of failure mode.

I have overwritten it.

beorn7:
Hmmm, that broke the tests (but not the build). I think our CI only tests in the branch the PR wants to pull from, but doesn't run the tests on the result of the merge. Not sure if that's intended. Perhaps @simonpasquier knows more? (I have a vague memory that our Travis setup used to test the result of the merge, but CircleCI tests in the branch to merge from…)

simonpasquier:
@beorn7 I need to investigate further (COVID-19 hinders my availability so I can't really commit to when). But from what I understand, Circle CI runs on the PR without merging against the base branch but it will run again after the merge hence the failure you've noticed.

beorn7:
Looks like it, and after the train of thought documented on #7010, I now think that's exactly the right thing to do.

PR #7013, opened 2020-03-20
tsdb: add log when compress ignored.

johncming:
Add log output when the record compression ignored, which's unusual to get larger data after compression.  Look at the proportions in the actual environment.

### DIFF
--- a/tsdb/wal/wal.go
+++ b/tsdb/wal/wal.go
@@ -622,6 +622,8 @@ func (w *WAL) log(rec []byte, final bool) error {
 		if len(w.snappyBuf) < len(rec) {
 			rec = w.snappyBuf
 			compressed = true
+		} else {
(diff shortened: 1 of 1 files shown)


cstyan:
Not sure how often this would actually come up but it could be a bit spammy. Maybe look at moving to debug and/or using a metric instead?

johncming:
> Not sure how often this would actually come up but it could be a bit spammy. Maybe look at moving to debug and/or using a metric instead?

How often it happened is a question.  It's appropriate to use a metric to count it.

Is this ignored logic reasonable? Compression is set, but may not actually be compressed. it's inconsistent with the help.

brian-brazil:
> Is this ignored logic reasonable?

Yes, it's a pretty normal check for compression to not do it if it'd cause a blowup.

johncming:
Close this pr and add a new pr #7024 which uses metrics instead.

PR #7015, opened 2020-03-20
[ENHANCEMENT] remote storage:Add default api implementation of remote write

liguozhong:
[ENHANCEMENT] remote storage:Add default api implementation of remote write

image

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -74,7 +74,6 @@ const (
 type errorType string
 
 const (
-	errorNone        errorType = ""
 	errorTimeout     errorType = "timeout"
 	errorCanceled    errorType = "canceled"
 	errorExec        errorType = "execution"
@@ -226,7 +225,7 @@ func NewAPI(
 ) *API {
 	return &API{
 		QueryEngine:           qe,
-		refs:                  make(map[string]uint64, 0),
+		refs:                  make(map[string]uint64),
 		Queryable:             q,
 		Appendable:            q,
 		targetRetriever:       tr,
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -16,6 +16,7 @@ package v1
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"math"
 	"math/rand"
 	"net"
@@ -32,6 +33,8 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+	"github.com/gogo/protobuf/proto"
+	"github.com/golang/snappy"
 	jsoniter "github.com/json-iterator/go"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
@@ -171,6 +174,8 @@ type TSDBAdmin interface {
 // them using the provided storage and query engine.
 type API struct {
 	Queryable   storage.Queryable
+	Appendable  storage.Appendable
+	refs        map[string]uint64
 	QueryEngine *promql.Engine
 
 	targetRetriever       targetRetriever
@@ -201,7 +206,7 @@ func init() {
 // NewAPI returns an initialized API type.
 func NewAPI(
 	qe *promql.Engine,
-	q storage.Queryable,
+	q storage.Storage,
 	tr targetRetriever,
 	ar alertmanagerRetriever,
 	configFunc func() config.Config,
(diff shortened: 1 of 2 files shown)


liguozhong:
@csmarchbanks @cstyan 
remote storage module pr

liguozhong:
# prometheus.yaml 1 
global:
scrape_configs:
  - job_name: 'prometheus'
    static_configs:
    - targets: ['localhost:9090']

liguozhong:
# prometheus.yaml 2
global:
scrape_configs:
  - job_name: 'kafka'
    static_configs:
    - targets: ['121.40.124.46:9997']
  - job_name: 'zookeeper'
    static_configs:
    - targets: ['121.40.124.46:8997']
remote_write:
  - url: http://localhost:9090/api/v1/write
    name: prometheus-hangzhou

brian-brazil:
This PR appears to be adding push to Prometheus, which is an explicit non-goal of the project.

roidelapluie:
Is that really something that can be solved in 126 lines of code? What is the catch?

cstyan:
@liguozhong Yeah this isn't really something we plan to support. You might find Thanos appealing as it's deployment on top of Prometheus to add a global query layer is relatively minimal. Otherwise remote write systems such as those suggest by Brian or others listed [here](https://prometheus.io/docs/operating/integrations/#remote-endpoints-and-storage) are your best options.

liguozhong:
> This PR appears to be adding push to Prometheus, which is an explicit non-goal of the project.
> 
> I'd suggest looking at options like Cortex or M3.

thanks.
We use it in an IOT project. We write it to a remote storage of influxdb through remote write. Because the memory and CPU disk configuration of edge host are relatively low, and the network is unidirectional, but in the case of regular condition label filtering such as metrics {namespace ~ = ". *"}, influxdb The query performance of is very slow,And we are worried that in the future, if the influxdb fails, it will be difficult for us to recover from the failure . so we want to switch to Prometheus as a PlanB. 

4 cpu core ,8GB mem: [edge host 1] [edge host 2] [edge host 3] （prometheus server）
⬇️（by remote storage write）
24 core, 128GB mem: [iot master host 1]  （influxdb）
⬇️（by remote storage read）
[prometheus server] 
⬇️(by promql  http api)
[grafana]

We only have one large memory host that can be used to store time series data. We don't know whether cortex or m3 can be deployed on one host. In addition, our team is only familiar with Prometheus thanos at present.

yeah , Prometheus is a pull model. I will close this PR first. We will maintain it in our branch for a while.

liguozhong:
> Yeah

thanks. 
We have used thanos in other projects, Thanos is a great project.
but thanos is not suitable for IOT projects. Thanos needs 128GB memory host, and the network is bidirectional. 
Edge host has only 8 GB of memory, and the network can only send out and prohibit inward network communication

liguozhong:
> Is that really something that can be solved in 126 lines of code? What is the catch?



> Is that really something that can be solved in 126 lines of code? What is the catch?

Yes, the 126 line code works. We are going to use this PR in an IOT project, but this PR is still in the stage of performance test and stress test.

refs map[string]uint64

We are worried that this "refs cache" will expand infinitely and eventually lead to failure, but there is no good way to solve it at present. We hope that it can be put in the community and other experts can help to complete the optimization of "refs cache"

> Is that really something that can be solved in 126 lines of code? What is the catch?


Yes, the 126 line code works. We are going to use this PR in an IOT project, but this PR is still in the stage of performance test and stress test.

refs map[string]uint64

We are worried that this "refs cache" will expand infinitely and eventually lead to failure, but there is no good way to solve it at present. We hope that it can be put in the community and other experts can help to complete the optimization of "refs cache"

roidelapluie:
Hello @liguozhong , we are willing to support this! Can you please help us? We are currently working on a design document:

#8307  https://docs.google.com/document/d/1PpTDtVaKAnhMRxBJ0k7_rCQ3Z3uJVcALgdPo0QpObSk

Thanks!!!!

liguozhong:
Okay, I will complete this PR task

> Hello @liguozhong , we are willing to support this! Can you please help us? We are currently working on a design document:
> 
> #8307 https://docs.google.com/document/d/1PpTDtVaKAnhMRxBJ0k7_rCQ3Z3uJVcALgdPo0QpObSk
> 
> Thanks!!!!

Okay, I will complete this PR task

Issue #6967, opened 2020-03-11
Prometheues v2.16.0: Duplicated metrics from remote storage due to external labels


clausa20:
Hello,
I have 3 Prometheus instances with a load balancer in the front. Each of these Prometheus has set external labels in their settings, for example:
```
external_labels:
    external_label_1: external_label_value
    external_label_2: external_label_value_0  (changes to 1 and 2 corresponding)
```
Also I have set a remote storage using CrateDB.

I've tried to update Prometheus version from v2.14.0 to v2.16.0 but....

When I perform a query of a metric saved in remote storage in version 2.14.0, I have a single result, for example:
`metric1{label1="value1",label2="value2"}`

but when I run the same query with version 2.16.0 I have 4 results, 3 of these results specifying the external label and 1 result without the external labels (the same as version 2.14.0):
```
metric1{label1="value1",label2="value2",external_label_1="external_label_value",external_label_2="external_label_value_0"}
metric1{label1="value1",label2="value2",external_label_1="external_label_value",external_label_2="external_label_value_1"}
metric1{label1="value1",label2="value2",external_label_1="external_label_value",external_label_2="external_label_value_2"}
metric1{label1="value1",label2="value2"}
```

For both version I see that in CrateDB metrics are saved with external labels, but for version 2.14.0 Prometheus is able to de-duplicate.

I would expect that version 2.16.0 has a single result as version 2.14.0.
Was something changed in version 0.16.0 that it cannot deduplicate metrics from external labels?

Thanks.

csmarchbanks:
You mean v2.14.0 and v2.16.0 not 0.X.0 correct?

clausa20:
@csmarchbanks yes correct. my bad, I meant v2.14.0 and v2.16.0 . I will try to update the description to correct it. Also I want to add that I tested the version v2.15.2 and it seems to have the same behaviour as v2.14.0, so I only see the duplicated metrics in version v2.16.0. Thanks.

csmarchbanks:
Hello @clausa20, I finally dug into this and I could reproduce it before https://github.com/prometheus/prometheus/pull/6971 was merged, but not afterward. Could you could verify that this has been fixed on the latest master build of Prometheus?

clausa20:
@csmarchbanks  yes I can test it.
In this release is the fix? v2.17.0-rc.1 
Or where is the build you mention? thanks.

roidelapluie:
[...]

PR #7018, opened 2020-03-20
Backport #6971 to release-2.17

roidelapluie:
storage: Removed SelectSorted method; Simplified interface; Added requirement for remote read to sort response.

This is technically BREAKING CHANGE, but it was like this from the beginning: I just notice that we rely in
Prometheus on remote read being sorted. This is because we use selected data from remote reads in MergeSeriesSet
which rely on sorting.

I found during work on https://github.com/prometheus/prometheus/pull/5882 that
we do so many repetitions because of this, for not good reason. I think
I found a good balance between convenience and readability with just one method.
Smaller the interface = better.

Also I don't know what TestSelectSorted was testing, but now it's testing sorting.

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -655,7 +655,7 @@ func (ng *Engine) populateSeries(ctx context.Context, querier storage.Querier, s
 	parser.Inspect(s.Expr, func(node parser.Node, path []parser.Node) error {
 		var set storage.SeriesSet
 		var wrn storage.Warnings
-		params := &storage.SelectParams{
+		hints := &storage.SelectHints{
 			Start: timestamp.FromTime(s.Start),
 			End:   timestamp.FromTime(s.End),
 			Step:  durationToInt64Millis(s.Interval),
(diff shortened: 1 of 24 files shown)


roidelapluie:
The goal is to fix #6967

beorn7:
Ooof, that's quite a change… @roidelapluie I guess you have already thought carefully about the tradeoff here to pull thin into the 2.17 release.

@bwplotka you should do the actual code review. I have nothing to object here, I just don't feel qualify to vet the risk.

roidelapluie:
> Ooof, that's quite a change… @roidelapluie I guess you have already thought carefully about the tradeoff here to pull thin into the 2.17 release.
> 
> @bwplotka you should do the actual code review. I have nothing to object here, I just don't feel qualify to vet the risk.

So, in short I'd better have this in 2.17.0 than changing the behavior in between 2.17.0 and a 2.17.1. It looks like this is a pretty important regression for remote read users.

roidelapluie:
So, I think that this regression in 2.16.0 is really big and that's why I want it.

brian-brazil:
#6971 is a refactoring, it shouldn't be changing any logic.

roidelapluie:
> #6971 is a refactoring, it shouldn't be changing any logic.

I changes sorting on remote read, as per the commit comment.

brian-brazil:
The commit description is incorrect, it doesn't change that.

brian-brazil:
What exactly is the bug that this is supposed to be fixing?

roidelapluie:
> What exactly is the bug that this is supposed to be fixing?

#6967

roidelapluie:
> What exactly is the bug that this is supposed to be fixing?

See https://github.com/prometheus/prometheus/issues/6967#issuecomment-600197799 for the details

bwplotka:
I don't thing we changed anything. Sorry for misleading commit comment. We only assured that on API level by commenting proto method. ):

Sorry for confusion and thanks Julien for the work.

bwplotka:
This is when we iterate on PR description but never change commit description ):

bwplotka:
Hm, ok so actually I fixed something by this? Still cannot see the exact place that is fixing the issue. Can you point to exact code line?

roidelapluie:
https://github.com/prometheus/prometheus/blob/release-2.16/storage/remote/read.go#L127-L134
look at that Select method: when SelectSorted was added, SelectSorted was not added to that file
so when you call Select Sorted on remote read, we do not remove the external labels

PR #7020, opened 2020-03-21
fix bug  missing an error.

johncming:
May lose error: `err` is nil and `_err` is not nil. This lead to return `[]error{nil}`

Although this may not happen, it is better not to base on `Unmarshal` implementation.


brian-brazil:
Thanks!

PR #7023, opened 2020-03-21
fix typo.

johncming:
fix two typos, one of which is a bug.


brian-brazil:
Do we have tests that should have caught this?

johncming:
> Do we have tests that should have caught this?

There are no tests for this.  Add more tests for package`rulefmt` later is better.

This bug `0` is obvious. If I have more time and I will add more tests.

pkg/rulefmt/rulefmt.go
```diff
@@ -98,7 +98,7 @@ func (g *RuleGroups) Validate(node ruleGroups) (errs []error) {
 		}
 	}
 
-	return errs
+	return
```
brian-brazil: There's no defer, so we shouldn't need a named return at all

bwplotka: Also we should never have `return` when there are arguments to return, this very prone to errors

Please see https://www.bwplotka.dev/2020/how-thanos-would-program-in-go/#use-named-return-parameters-carefully for rationales.

roidelapluie:
@johncming Can you please address the comment and add a unit test that highlights the bug?

johncming:
> @johncming Can you please address the comment and add a unit test that highlights the bug?

OK. 

Some changes: distinguishing between two situations of
error message for both setting and non-setting of 'alert' and 'record' at the same time.

pkg/rulefmt/rulefmt.go
johncming: ok

brian-brazil: @bwplotka Are you happy with this now?

roidelapluie:
Closing in favour of #12903. Thanks!

PR #7024, opened 2020-03-21
add metrics for  wal log compressed.

johncming:
add two metrics for wal log compressed which can calculate the ratio of compressed data.

### DIFF
--- a/tsdb/wal/wal.go
+++ b/tsdb/wal/wal.go
@@ -181,13 +181,15 @@ type WAL struct {
 }
 
 type walMetrics struct {
-	fsyncDuration   prometheus.Summary
-	pageFlushes     prometheus.Counter
-	pageCompletions prometheus.Counter
-	truncateFail    prometheus.Counter
-	truncateTotal   prometheus.Counter
-	currentSegment  prometheus.Gauge
-	writesFailed    prometheus.Counter
+	fsyncDuration    prometheus.Summary
+	pageFlushes      prometheus.Counter
+	pageCompletions  prometheus.Counter
+	truncateFail     prometheus.Counter
+	truncateTotal    prometheus.Counter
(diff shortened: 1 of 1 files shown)


tsdb/wal/wal.go
```diff
@@ -593,6 +605,7 @@ func (w *WAL) Log(recs ...[]byte) error {
 // - the record is bigger than the page size
 // - the current page is full.
 func (w *WAL) log(rec []byte, final bool) error {
+	defer w.metrics.writesTotal.Inc()
```
codesome: We are measuring `writesFailed` in `Log`, so it will lead to confusion if anyone is trying to relate `writesFailed` with `writesTotal`. (Additionally, we should not have defer `defer` else it is possible to have `writesCompressed` > `writesTotal` which doesn't make sense :) )
```suggestion
	w.metrics.writesTotal.Inc()
```

roidelapluie:
@johncming  can you please address the comments on this pull request? It looks like those metrics would be useful.

johncming:
> @johncming can you please address the comments on this pull request? It looks like those metrics would be useful.

The existing part of this pr isn't very useful. As @csmarchbanks  mentioned, byte level statistics may be more useful. For example, it's an understanding I have, using counters, count the total number of record bytes before compression and the total number of record bytes after compression, but one thing to consider is whether such counters would degrade the log performance of wal. Also, I found no metric for the total size of the wal (or I may be missing something)

Thanks.

tsdb/wal/wal.go
johncming: Yes, after review, the defer doesn't fit.

roidelapluie:

We have decided to close this pull request because of the lack of updates. @csmarchbanks @cstyan if this is still needed, can you please open a help-wanted issue so we still track this?

PR #7026, opened 2020-03-21
Refactored Prometheus flags to bwplotka/flagarize library.

bwplotka:
Hi :wave: 

I created some library (https://github.com/bwplotka/flagarize) for Thanos purpose, but since it's native to kingpin, I thought I can improve Prometheus flags as well. By accident, I found one bug where we define a flag that is no longer used.
The idea behind this is that you can define all flags in struct tags and since already use Options/Config structs this is now much cleaner. Let me know if you are happy to have this. (: 

I compared --help AFTER vs BEFORE and this is the only diff (except manual reordering - I don't know why but kingpin does not sort flags):

image

image


brian-brazil:
We've many repos that use Go flags, are you proposing to switch all of them?

bwplotka:
For consistency yes, it would be nice -  if we agree on such an approach. 

However, there is nothing blocking for other repos to use normal flags, if there are only a few. It's still kingpin after all. 

I am writing blog post about `flagarize` now, will explain some rationales there.

brian-brazil:
> However, there is nothing blocking for other repos to use normal flags, if there are only a few.

We should be consistent across all repos. It took us literally years to complete to move to kingpin, I don't want to go back to an inconsistent state with no defined end date.

bwplotka:
It makes sense to me. Happy to do the work if we decide to follow this approach.

roidelapluie:
Before merging this it means we need to code review your lib. That can take some time.

bwplotka:
TBH there is also https://github.com/alecthomas/kong of the same author as our current library. But it's a bit too complex for our needs IMO. Will explain in blog post.

bwplotka:
Yes, of course, no rush (: 

Will also post here our attempt with Thanos.

cmd/prometheus/main.go
```diff
@@ -94,233 +92,118 @@ func main() {
...
-
-	a.Flag("rules.alert.resend-delay", "Minimum amount of time to wait before resending an alert to Alertmanager.").
-		Default("1m").SetValue(&cfg.resendDelay)
-
-	a.Flag("alertmanager.notification-queue-capacity", "The capacity of the queue for pending Alertmanager notifications.").
-		Default("10000").IntVar(&cfg.notifier.QueueCapacity)
-
-	a.Flag("alertmanager.timeout", "Timeout for sending alerts to Alertmanager.").
-		Default("10s").SetValue(&cfg.notifierTimeout)
+	// TODO(bwplotka): This flag is not used... by accident? Should we remove?
```
bwplotka: Any ideas what happened here? We have this but it was never used... :scream:

roidelapluie: or we should use it :)

bwplotka: hah yea

bwplotka:
Blog post: https://www.bwplotka.dev/2020/flagarize/

roidelapluie:
Thanks for this proposal.

I would generally think that using more popular repos like 'kong' should be preferred over rewriting them. You blog post did not really convince me that kong would be a problem for prometheus.

In addition to that, I am opposed to the fact that we deal with flags in other places than `cmd/`. It seems that now we deal with flags in, e.g., notifier.go., web.go, etc. We should deal with flags outside of the cmd package. I'd hope all non-cmd packages would not depend on that.

bwplotka:
> I would generally think that using more popular repos like 'kong' should be preferred over rewriting them. You blog post did not really convince me that kong would be a problem for prometheus.

Why? What exactly did not convince you? And I don't know if "more popular" is a good measure. I can ask around friends to give me more stars on repo than 150 `kong` has. Would this make my library more popular? In fact I did not know about `kong` until I started my blog post ;p 

I don't think we should be convinced by popularity. If that would be the case we would all use still Graphite and Nagios, not Prometheus, right @roidelapluie? (: 

> In addition to that, I am opposed to the fact that we deal with flags in other places than cmd/. It seems that now we deal with flags in, e.g., notifier.go., web.go, etc. We should not deal with flags outside of the cmd package. I'd hope all non-cmd packages would not depend on that.

Yes. In fact this discussion with @peterbourgon https://twitter.com/bwplotka/status/1241759656864817153 exactly mentioned that. We should have special from-flag configuration struct in `cmd` and then convert it to package ones. We don't do this right now either.

roidelapluie:
> > I would generally think that using more popular repos like 'kong' should be preferred over rewriting them. You blog post did not really convince me that kong would be a problem for prometheus.
> 
> Why? What exactly did not convince you? And I don't know if "more popular" is a good measure. I can ask around friends to give me more stars on repo than 150 `kong` has. Would this make my library more popular? In fact I did not know about `kong` until I started my blog post ;p

If not about popularity and collaboration, why don't we keep kingpin then? What problems are we trying to solve? I was thinking we move because kingpin is no longer really maintained.

bwplotka:
I propose to move from the reasons defined in my blog post, not because of `kingpin` being in contribution mode (it's maintained).

```
1. It’s not clear what fields from the Go struct are actually set from flags, which are manually set. Especially for nested configuration structs. It’s worth noting that overall, as mentioned by Peter Bourgon, mixing user intent with derived config is a bad design smell. Such a mistake is quite hard to detect with flag definition like above.
2. It’s easy to miss that some configuration variable is not used anymore, but some flag is still defined for it. This is because Go will not detect it being used, as it’s passed in, for example, `DurationVar(&cfg.var)`. In fact, during the move of Prometheus flags to my library, I found that one config field was exactly like this.
3. We just defined a field in configuration struct, why we need to create additional 2 complex lines for each of it?
4. What if we would like to parse our configuration from JSON or YAML as well? It would require huge refactoring.
5. What about complex flags like Thanos’s path or content? Registering of those is quite tedious.
```

krasi-georgiev:
> We should have special from-flag configuration struct in cmd and then convert it to package ones. We don't do this right now either.

How would that look like? 

I agree that it would be e bit harder to know what are all cli args if these are spread in different files.

bwplotka:
Looks like the benefit of this pattern is not big enough to justify such change. (: 

The same conclusion was on Thanos PR: https://github.com/thanos-io/thanos/pull/2267#discussion_r410712727

Thanks all for reviews (:

PR #7034, opened 2020-03-22
Release 2.17.0

roidelapluie:
I plan to tag & release 2.17.0 tomorrow evening.


bwplotka:
(working day)

brian-brazil:
Unless there were changes that were expected to have a notable impact on performance, that's fine.

bwplotka:
Hm, it's not only performance, it's also opportunity for users to take RC and test out. With just Sunday, there was literally no point in releasing RC.4 if we do proper release today. I am quite sure **no one** deployed RC.4 yet. So we could do straight release on Sunday and it would have an equal results. (:

roidelapluie:
> Hm, it's not only performance, it's also opportunity for users to take RC and test out. With just Sunday, there was literally no point in releasing RC.4 if we do proper release today. I am quite sure **no one** deployed RC.4 yet. So we could do straight release on Sunday and it would have an equal results. (:

Given the bugs we already had in RC, and that were only reported by prombench, not by users, I guess no one deploys any RC.

Prombench will have run for 3 days, as I launched it on Friday evening, which is the RELEASE.md requirement.

Yesterday I also ran a bunch of extra tests with the prometheus-prombench of #7018, which is the same as the RC (did not use the RC one not to falsify prombench tests) and it looks all fine.

Also, this release has been in the making for 10 days now, I am confident that this is a good release.

beorn7:
I'll deploy rc.4 right now on a production server at Grafana Labs. It's probably good to give it 24 hours of time to prove itself.

beorn7:
By which I meant to say: Let's to the release Tuesday afternoon.

brancz:
I think we did end up on this conclusion, but

> Also, this release has been in the making for 10 days now, I am confident that this is a good release.

Is not a reason to shorten the agreed upon process in my opinion. The whole point is to give people enough time to try out release candidates, and we should not be inconsistent with that, no matter how many release candidates we end up cutting for a minor release.

roidelapluie:
> I think we did end up on this conclusion, but
> 
> > Also, this release has been in the making for 10 days now, I am confident that this is a good release.
> 
> Is not a reason to shorten the agreed upon process in my opinion. The whole point is to give people enough time to try out release candidates, and we should not be inconsistent with that, no matter how many release candidates we end up cutting for a minor release.

(No process is specifying that we need to wait 3 working days before cutting releases, if so, we probably should update RELEASE.md)

roidelapluie:
> I'll deploy rc.4 right now on a production server at Grafana Labs. It's probably good to give it 24 hours of time to prove itself.

Is it the first rc you deploy?

beorn7:
The previous one was rc.0. (It had no issues, either.)

PR #7035, opened 2020-03-22
remove deprecated methods from the MetricProvider interface

tariq1890:
Since updating to the latest k8s deps, the deprecate methods of MetricProvider have been removed. They are now just cruft in the prometheus codebase

bonus change: some goimports improvements in the k8s sd module :)

### DIFF
--- a/discovery/kubernetes/client_metrics.go
+++ b/discovery/kubernetes/client_metrics.go
@@ -18,7 +18,6 @@ import (
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
-
 	"k8s.io/client-go/tools/metrics"
 	"k8s.io/client-go/util/workqueue"
 )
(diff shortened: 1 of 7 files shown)


tariq1890:
The failure doesn't seem to be related to the code

brancz:
Restarted the job. Let’s see how it goes :)

tariq1890:
@juliusv Any idea why these changes cause the react builds to fail?

juliusv:
@tariq1890 I have no idea, locally everything passes for me with your PR branch. I'll restart the test just one more time to see if it continues persisting...

csmarchbanks:
These look like the  errors I fixed in https://github.com/prometheus/prometheus/pull/7077. The errors were reproducible locally if I ran with `CI=true`, would you try rebasing your branch and see if they go away?

tariq1890:
That worked! Thanks @csmarchbanks :)

Issue #4125, opened 2018-04-30
Add linters to golangci lint.


SuperQ:
## Proposal

Enable additional linters in `golangci-lint`
```yaml
linters:
  enable:
  - golint
```

brian-brazil:
Do we know what level of false positives we can expect?

SuperQ:
Good question, Maybe @mjtrangoni has some more experience with that.

mjtrangoni:
@brian-brazil you are right, there are going to be **many** false positives. 
IMHO, the best and simplest way I found was using [gometalinter](https://github.com/alecthomas/gometalinter) plus a JSON configuration file, like [this](https://github.com/mjtrangoni/flexlm_exporter/blob/master/.gometalinter.json). This configuration file should be also project-specific.
Another interesting thing I found was that gometalinter introduces some specific flags for some linters. I mean, It is not only running the linter itself.
OTOH, I think this will be adding some great value to the projects.

juliusv:
Yeah, careful with linting, at least https://github.com/golang/lint's README.md says "In short, this tool is not, and will never be, trustworthy enough for its suggestions to be enforced automatically, for example as part of a build process.", and we've known that for a while :)

We can still have targets for it of course, just not have them be part of CI checks.

PR #7037, opened 2020-03-23
Enable golint in CI

SuperQ:
Enable the golint module in golangci-lint.

Fixes: https://github.com/prometheus/prometheus/issues/4125

### DIFF
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -301,16 +301,14 @@ func Compare(a, b Labels) int {
 		if a[i].Name != b[i].Name {
 			if a[i].Name < b[i].Name {
 				return -1
-			} else {
-				return 1
 			}
+			return 1
 		}
 		if a[i].Value != b[i].Value {
 			if a[i].Value < b[i].Value {
 				return -1
-			} else {
-				return 1
 			}
+			return 1
 		}
 	}
 	// If all labels so far were in common, the set with fewer labels comes first.
--- a/pkg/textparse/openmetricsparse.go
+++ b/pkg/textparse/openmetricsparse.go
@@ -233,7 +233,7 @@ func (p *OpenMetricsParser) Next() (Entry, error) {
 	p.hasExemplarTs = false
 
 	switch t := p.nextToken(); t {
-	case tEofWord:
+	case tEOFWord:
 		if t := p.nextToken(); t != tEOF {
 			return EntryInvalid, errors.New("unexpected data after # EOF")
 		}
--- a/promql/parser/printer.go
+++ b/promql/parser/printer.go
@@ -113,7 +113,7 @@ func (node *Call) String() string {
 
 func (node *MatrixSelector) String() string {
 	// Copy the Vector selector before changing the offset
-	var vecSelector VectorSelector = *node.VectorSelector.(*VectorSelector)
+	vecSelector := *node.VectorSelector.(*VectorSelector)
 	offset := ""
 	if vecSelector.Offset != time.Duration(0) {
 		offset = fmt.Sprintf(" offset %s", model.Duration(vecSelector.Offset))
--- a/tsdb/wal/wal.go
+++ b/tsdb/wal/wal.go
@@ -837,6 +837,7 @@ type segmentBufReader struct {
 	off  int // Offset of read data into current segment.
 }
 
+// nolint:golint // TODO: Consider exporting segmentBufReader
 func NewSegmentBufReader(segs ...*Segment) *segmentBufReader {
 	return &segmentBufReader{
 		buf:  bufio.NewReaderSize(segs[0], 16*pageSize),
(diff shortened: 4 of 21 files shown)


SuperQ:
There are two additional lint items I haven't fixed yet.

```
tsdb/db.go:42:2: a blank import should be only in a main or test package, or have a comment justifying it (golint)
	_ "github.com/prometheus/prometheus/tsdb/goversion"
	^
tsdb/wal/wal.go:840:44: exported func NewSegmentBufReader returns unexported type *wal.segmentBufReader, which can be annoying to use (golint)
func NewSegmentBufReader(segs ...*Segment) *segmentBufReader {
                                           ^
```

SuperQ:
For the tsdb/goversion issue, I think a comment explaining that it's to load the package into main is justified.

brian-brazil:
> For the tsdb/goversion issue, I think a comment explaining that it's to load the package into main is justified.

I don't see the need for a comment at all, this sounds more like a personal preference than a lint issue.

mdlayher:
"Load the package into main" is implied by the import statement; what sort of initialization is happening with the underscore import? Leave a quick comment explaining that and move on. It's less hassle to not fight the linter and move on.

beorn7:
Can't you just put a golint pragma there to disable the warning? We can later think about more invasive/controversial changes. It's still better to have two more "TODOs" in the code than missing all the other lint issues.

SuperQ:
@beorn7 Yes, you can `nolint:modulename`. I wanted to fix as many things as many reasonable changes as possible before resorting to ignoring things.

SuperQ:
All lint issues resolved.

pkg/textparse/openmetricsparse.go
```diff
@@ -233,7 +233,7 @@ func (p *OpenMetricsParser) Next() (Entry, error) {
 	p.hasExemplarTs = false
 
 	switch t := p.nextToken(); t {
-	case tEofWord:
+	case tEOFWord:
```
beorn7: Shouldn't this be `EOfWord`?

tsdb/wal/wal.go
```diff
@@ -837,6 +837,7 @@ type segmentBufReader struct {
 	off  int // Offset of read data into current segment.
 }
 
+// nolint:golint
```
beorn7: Perhaps add something like `TODO: Consider exporting segmentBufReader`?

SuperQ: Sure.

pkg/textparse/openmetricsparse.go
mdlayher: Go precedent is `EOF`, as in `io.EOF`.

promql/parser/printer.go
```diff
@@ -113,7 +113,7 @@ func (node *Call) String() string {
 
 func (node *MatrixSelector) String() string {
 	// Copy the Vector selector before changing the offset
-	var vecSelector VectorSelector = *node.VectorSelector.(*VectorSelector)
+	var vecSelector = *node.VectorSelector.(*VectorSelector)
```
mdlayher: Nit: should just be `:=`.

pkg/textparse/openmetricsparse.go
beorn7: `EOF` is an acronym that stands for "End Of File".

In `EOfWord` only the `E` is an abbreviation, it stands for "End Of Word".

`EOFWord` would suggest "End Of File Word" or "End Of Foo Word" or something.

beorn7: And now I have to correct myself, it actually does stand for "End Of File Word" or "EOF Word". So please forget my whole line of argument here. All good.

mdlayher: Ack, it didn't even occur to me that this might not mean "end-of-file". Apologies for jumping in!

beorn7: But it does. You were right, and I was wrong. Thanks for jumping in! :o)

promql/parser/printer.go
SuperQ: Ohh, good one.

pkg/labels/labels.go
```diff
@@ -301,16 +301,14 @@ func Compare(a, b Labels) int {
 		if a[i].Name != b[i].Name {
 			if a[i].Name < b[i].Name {
 				return -1
-			} else {
-				return 1
 			}
+			return 1
```
roidelapluie: Should we be consistent with https://github.com/prometheus/prometheus/pull/6845

beorn7: I don't understand #6845. How can that be by design?

roidelapluie: This was left as it until now as it is easier to reason about. (that was @brian-brazil 's call at the time)

roidelapluie: But yeah, if we need to chose between this and addinfg a comment to disable the check, let's pick this solution.

PR #7038, opened 2020-03-23
share string interner between scraper and remote_write

rfratto:
This PR exposes the string interner globally with a shared pool and modifies both the scraper and remote_write to use that shared pool. This change has been in use by grafana/agent, and seems to show minor improvements in memory usage when remote_write relabeling rules are not used. 

Note that the `string_interner_zero_reference_releases_total` series now has a different namespace and subsystem from what it was initially set to in the RW code.

Signed-off-by: Robert Fratto <robert.fratto@grafana.com>


rfratto:
I'll take a look at the failing tests soon.

brian-brazil:
Memory usage is more complex in Prometheus, what's in the scrape also end up held in the tsdb so we'd need to verify this is all sane there.

rfratto:
> Memory usage is more complex in Prometheus, what's in the scrape also end up held in the tsdb so we'd need to verify this is all sane there.

Could you explain in more detail?

brian-brazil:
Last I looked into this, we keep basically the entire scrape in memory via tsdb references so adding this could actually make memory usage worse.

Have you actually tested this with Prometheus?

rfratto:
> Have you actually tested this with Prometheus?

I haven't. My logic was that the write path was similar enough, but I didn't account for the TSDB. I'll do some digging and testing and report back.

roidelapluie:
@rfratto Any news on your tests? Does this help? Thanks!

rfratto:
@roidelapluie Thanks for reminding me! This fell off my radar for a bit; I'll test this with Prometheus this week and report back in a few days.

rfratto:
I've been running two Prometheus servers scraping the same jobs for a bit (one without this change, one with the change) and it seems like the memory difference is within a margin of error. Specifically, I'm seeing a 2-4% decrease in memory usage, but I don't particularly trust it. 

Can I get a prombench run for more detailed results? 

FYI: I'm going on holiday for two weeks, but I'll keep looking into this once I'm back. I don't want to forget and accidentally let this sit for another six months again 🙂

roidelapluie:
Did you rebase on master?

rfratto:
Yeah, this is rebased on top of ed2a483beb23c43ae1ff4d3ec125f28447ad0d22.

rfratto:
It _seems_ as though memory usage is slightly improved but allocs/s is slightly worse. Am I reading that right?

brian-brazil:
It's hard to tell if there's any change in memory - maybe a few % but it's not consistent. Similarly with CPU.

rfratto:
Yeah, I agree that it seems to be basically a no-op here. It'd probably be more of an improvement if this was also utilized in the TSDB (if it could be utilized, anyways), but that's something I unfortunately lack the bandwidth for at the moment. 

What should we do with this PR?

brian-brazil:
We should probably drop it, as any benefits are minimal and the current scrape/tsdb memory management is quite intricate already. Getting this right on top of it would make that even harder to maintain/improve (this PR is not correct from a quick peek for example).

rfratto:
Thanks. I'll close this for now. 

I'll investigate what effects this has on the Agent if I don't have the shared interner. At the time I wrote this, I had measured that most of the memory utilization in the Agent was coming from labels, so this had helped quite a bit. Long term, I don't want grafana/agent to just use a mirror with "unofficial" changes forever, so I'm more than happy to investigate if my initial findings were wrong. 

I'll come back with an issue for more discussion if it turns out that the Agent really needs the change.

brian-brazil:
> I'm more than happy to investigate if my initial findings were wrong.

I'd guess not, however what Prometheus is doing is more complex than just writing to the WAL so that particular code path has already been optimised in terms of memory allocations.

roidelapluie:
@rfratto it seems that the grafana agent has this feature, do you have any feedback on it?

rfratto:
> @rfratto it seems that the grafana agent has this feature, do you have any feedback on it?
I haven't done any before/after benchmarks against the latest Prometheus release, but at the time I introduced the feature to the Agent (back in March), I believe it significantly helped with memory usage. There were 3 copies of all labels, and sharing the interner reduced that to one. 

I looked through slack history to see if I ever documented the specific change in memory usage, but unfortunately it got bundled into a series of other changes so I don't have anything accurate. I pulled up an old profile and label strings were responsible for about 17% of total memory usage, so removing 2 extra copies would've brought that down to ~5%, and made a noticeable difference. 

I'm sorry I don't have anything more reliable than the pencil math at the moment.

roidelapluie:
@codesome do you think we could have that improvement as an experimental feature during one release, and then make that mainstream in the next release? would that make sense?

PR #7045, opened 2020-03-25
move pathprefix to a separated hook to avoid passing it by hand

bilxio:
Signed-off-by: bilxio <haibxz@gmail.com>

### DIFF
--- a/web/ui/react-app/src/App.test.tsx
+++ b/web/ui/react-app/src/App.test.tsx
@@ -7,7 +7,7 @@ import { Router } from '@reach/router';
 import { Alerts, Config, Flags, Rules, ServiceDiscovery, Status, Targets, TSDBStatus, PanelList } from './pages';
 
 describe('App', () => {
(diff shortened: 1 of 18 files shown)


roidelapluie:
@juliusv Can you please triage this pull request? Thanks!

juliusv:
@bilxio Hi, thanks, and sorry for missing this!

If we want to treat the path prefix more as global state vs. passing it down explicitly, then the React way of doing this would be to use [Context](https://reactjs.org/docs/context.html), along with the [`useContext()`](https://reactjs.org/docs/hooks-reference.html#usecontext). Then we would put the consoles link (https://github.com/prometheus/prometheus/blob/5ea6bdca1f4731fb095d37714c268159213f2b10/web/ui/react-app/src/index.tsx#L23-L29) in there as well probably, and wrap the whole app in a config context.

roidelapluie:

Since the original author did not reply further, we have decided to close this pull request.

PR #7049, opened 2020-03-25
Sample limit relabeling

nabokihms:
Add the ability to relabel sample_limit setting. It is convenient to use with service discovery. 

1. Applications now can tell you "If I give you more then 1k samples, I'm broken!"
2. You can define sample limits individually for namespaces (or environments) and reduce boilerplate code.

Signed-off-by: m.nabokikh <maksim.nabokikh@flant.com>

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -221,6 +221,10 @@ func newScrapePool(cfg *config.ScrapeConfig, app storage.Appendable, jitterSeed 
 		}
 		opts.target.SetMetadataStore(cache)
 
+		limit := opts.target.SampleLimit()
+		if limit == 0 {
+			limit = opts.limit
+		}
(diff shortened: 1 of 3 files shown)


brian-brazil:
Thank you for your PR. This is meant to be a broadly applied emergency circuit breaker to protect Prometheus, not a finely grained quota/protection mechanism, so I don't think this makes sense to add.

roidelapluie:
I second @brian-brazil 's view.

roidelapluie:
Duplicate of #7340, which was implemented in 2.21.

roidelapluie:
We looked at it during the bug scrub and decided to close it as per above comments. Thanks!

nabokihms:
Anyway, thanks for paying your attention! I appreciate it.

PR #7050, opened 2020-03-25
Compare querier storage to primary storage via reflect.DeepEqual.

cstyan:
This is meant to fix a test failure introduced in #7005. I believe the issue is related to wrapping of the base primary storage struct in other types with the new Querier types, but I could be wrong. Essentially, the comparison of `qryResult.qr != q.primaryQuerier` no longer works.


bwplotka:
Merging to unblock other PRs.

Issue #7048, opened 2020-03-25
Query/Rule Time Regression After Upgrade To 2.17.0


hawran:
**What did you do?**
Upgraded from 2.16.0 -> 2.17.0.

**What did you expect to see?**
Not such a big increase of rule's evaluation time.

**What did you see instead? Under which circumstances?**
As "_a picture is worth a thousand words_" see the screenshots attached:

**prometheus_rule_group_last_duration_seconds:**
![200325-170434 prometheus_rule_group_last_duration_seconds](https://user-images.githubusercontent.com/25924718/77559167-1dbc2d00-6ebc-11ea-92da-f47334555870.png)

**engine_query_duration_seconds (the second graph uses the log10 scale for the Y-axis):**
![200325-170802 engine_query_duration_seconds](https://user-images.githubusercontent.com/25924718/77559239-3a586500-6ebc-11ea-9208-a8514a60fc61.png)

Note the increase of the `prepare_time` slice...

During the time from about 10 am to about 15 pm we were using the version 2.17.0.

**Environment**

* System information:
```
 Kernel Version:                  5.3.0-24-generic
 OS Image:                        Ubuntu 18.04.2 LTS
 Operating System:                linux
```

* Prometheus version:
	2.17.0

bwplotka:
Looks really similar to https://github.com/prometheus/prometheus/issues/6968, but indeed it's for even newer version bumb.

SuperQ:
We also just had to rollback to 2.16.0 due to memory and CPU overload with isolation.

https://gitlab.com/gitlab-com/gl-infra/production/-/issues/1838

We have a very large number or recording rules, some with wide (6 hour) range vectors.

roidelapluie:
I've open #7051

hawran:
> .. are you sure you had 2.16 before? (:

Well, I'm sure it was 2.16.

hawran:
Just to give you more concrete examples (rule groups snippets):

```
expr: sum
  by(namespace, server) (rate(http_client_requests_total{app="watcher",container_name="watcher",status_code!="200"}[5m]))
  / sum by(namespace, server) (rate(http_client_requests_total{app="watcher",container_name="watcher"}[5m]))
  > 0.4
for: 5m
```
![200326-093053 watcher](https://user-images.githubusercontent.com/25924718/77631663-2d378680-6f4d-11ea-94ef-c57da87bef5d.png)


```
expr: sum
[...]

Issue #6968, opened 2020-03-11
Query/Rule Time Regression


ipstatic:
**What did you do?**
Upgraded from Prometheus 2.8.1 to 2.16.0

**What did you expect to see?**
Better or the same query/recording rule timing.

**What did you see instead? Under which circumstances?**
Instantly saw recording rule times increase by 3-5x. Other queries taking 60-100ms longer to return results.

**Environment**

* System information:

	Linux 4.14.94+ x86_64

* Prometheus version:

	2.16.0

I have attached some images to show the differences.

This image shows a few queries on a 2.8 instance.
<img width="1535" alt="2 8" src="https://user-images.githubusercontent.com/332617/76447698-7c9f8380-639f-11ea-865d-94785b8af167.png">

These are the same queries on the same instance after upgrading to 2.16.0 (notice the load time is 60ms more). We noticed this due to the `prometheus.html` console taking ~60 seconds to load due to the increased query time (it shows a listing of all Prometheus instances plus their memory, TSDB and ingestion stats).
<img width="1538" alt="2 16" src="https://user-images.githubusercontent.com/332617/76447701-7d381a00-639f-11ea-92b2-a8917257ee27.png">

And here is the same instance showing how a recording rule group went from nearly instant to 5 seconds.
<img width="1329" alt="recording-rules" src="https://user-images.githubusercontent.com/332617/76447702-7d381a00-639f-11ea-92ac-8f032e8644cd.png">

brian-brazil:
Can you share the full queries and how many series are touched? Did anything else change? Is the machine overloaded?

Performance can and does vary across versions.

ipstatic:
It is basically any query. We just attempted to upgrade half our fleet to 2.16 and had to roll back due to increased rule evaluation / query times. We are seeing this across the board for everything. These instances are not overloaded, we give Prometheus a lot of memory headroom (usually Prometheus is only using 40-60% of memory to allow for large queries).

That recording rule group that jumped in the screenshot above is actually our alert rules and has very simple queries like `up{job='job'} == down`. If I can be more helpful please let me know. I know this is kind of general but we are seeing this on every instance we attempt to upgrade.

For example, this one instance is our largest instance and has some very high cardinality metrics. The total recording rule group time was ~200 seconds. After the upgrade it went up to 3,000+ seconds. It is represented by the green line below.
[...]

Issue #6672, opened 2020-01-21
Flaky test on Windows


roidelapluie:
--- FAIL: TestQueryTimeout (0.05s)
    engine_test.go:107: expected timeout error but got none: expected error, got none

That test sometimes fail on windows (master branch).

To be investigated, but I think it is related to Window's being not precise enough with time.Time.

roidelapluie:
Another 
--- FAIL: TestManagerTargetsUpdates (0.03s)
    manager_test.go:356: Scrape manager's channel remained blocked after the set threshold.

roidelapluie:
--- FAIL: TestShutdown (1.21s)
    queue_manager_test.go:168: Took too long to shutdown: 1.1033911s > 1s

roidelapluie:
=== RUN   TestQueryConcurrency
--- FAIL: TestQueryConcurrency (0.05s)
    engine_test.go:69: Query within concurrency threshold not being executed

roidelapluie:
=== RUN   TestMetricsUpdate
*** Test killed: ran too long (10m0s).
FAIL	github.com/prometheus/prometheus/rules	600.087s

PR #7051, opened 2020-03-25
Revert head posting optimization

roidelapluie:
See #6968 #7048

### DIFF
--- a/tsdb/block.go
+++ b/tsdb/block.go
@@ -111,9 +111,8 @@ type ChunkReader interface {
 
 // BlockReader provides reading access to a data block.
 type BlockReader interface {
-	// Index returns an IndexReader over the block's data within the specified
-	// timeframe.
-	Index(mint, maxt int64) (IndexReader, error)
+	// Index returns an IndexReader over the block's data.
+	Index() (IndexReader, error)
 
 	// Chunks returns a ChunkReader over the block's data.
 	Chunks() (ChunkReader, error)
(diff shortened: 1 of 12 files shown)


bwplotka:
I guess it was not limiting the stuff we touch index in head on query time?

roidelapluie:
Probably because we scan the full head before filtering on labels. That is a huge loop. Is that correct?

bwplotka:
What's wrong for windows? ):

roidelapluie:
> What's wrong for windows? ):

https://github.com/prometheus/prometheus/issues/6672

bwplotka:
Happy for this to be merged @roidelapluie

roidelapluie:
I think we can wait a few hours to see if @SuperQ can install that on its env?

bwplotka:
And what if not fully solved? There might be more sources of regression not
only this.

But yea, let's see first (:

SuperQ:
I've deployed this patch to two nodes. It seems to be in the order of expected additional CPU time required for isolation. (5-10%) 

I'm going to give it another 30min so I can verify that it stays that way.

roidelapluie:
> I've deployed this patch to two nodes. It seems to be in the order of expected additional CPU time required for isolation. (5-10%)
> 
> I'm going to give it another 30min so I can verify that it stays that way.

What about recording rule time?

SuperQ:
Recording rule time is also "normal"

image

SuperQ:
Just to followup, CPU time:

image

bwplotka:
Thanks for quick check @SuperQ ! :tada:

PR #7058, opened 2020-03-26
change labelset comparison in promql engine to avoid false positive during detection of duplicates

### LINKED ISSUE #5724: hash collision in engine leads to incorrect "multiple matches for labels" error

cstyan:
This PR changes from using hashes to compare labelsets to see if labelsets on each side of a promql query are dupicated to just using strings of `labels.Labels`.

### DIFF
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -18,7 +18,6 @@ import (
 	"encoding/json"
 	"sort"
 	"strconv"
-	"unsafe"
 
 	"github.com/cespare/xxhash"
 )
@@ -63,36 +62,10 @@ func (ls Labels) String() string {
 	return b.String()
 }
 
-// NoRenderString returns ls as a string.
-// It uses an byte invalid character as a separator and so
-// should not be used for printing.
-func (ls Labels) NoRenderString(b *bytes.Buffer) string {
-	b.Reset()
-	b.WriteByte('{')
-	for i, l := range ls {
-		if i > 0 {
-			b.WriteByte(',')
-			b.WriteByte(' ')
-		}
-		b.WriteString(l.Name)
-		b.WriteByte('=')
-		b.WriteByte(sep)
-		b.WriteString(l.Value)
-		b.WriteByte(sep)
-	}
-	b.WriteByte('}')
-	return b.String()
-}
-
-func yoloString(b []byte) string {
-	return *((*string)(unsafe.Pointer(&b)))
-}
-
-// YoloString returns ls as an unsafe pointer to b's bytes casted to a string.
-// This is done to save on allocations that would happen during b.String() calls.
+// Bytes returns ls as a byte slice.
 // It uses an byte invalid character as a separator and so should not be used for printing.
-func (ls Labels) YoloString(b *bytes.Buffer) string {
-	b.Reset()
+func (ls Labels) Bytes(buf []byte) []byte {
+	b := bytes.NewBuffer(buf[:0])
 	b.WriteByte('{')
 	for i, l := range ls {
 		if i > 0 {
@@ -106,7 +79,7 @@ func (ls Labels) YoloString(b *bytes.Buffer) string {
 		b.WriteByte(sep)
 	}
 	b.WriteByte('}')
-	return yoloString(b.Bytes())
+	return b.Bytes()
 }
 
 // MarshalJSON implements json.Marshaler.
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -173,6 +173,7 @@ func (ls Labels) HashWithoutLabels(b []byte, names ...string) (uint64, []byte) {
 }
 
 // WithLabels returns a new labels.Labels from ls that only contains labels matching names.
+// 'names' have to be sorted in ascending order.
 func (ls Labels) WithLabels(names ...string) Labels {
 	var ret Labels
 	i, j := 0, 0
@@ -190,6 +191,8 @@ func (ls Labels) WithLabels(names ...string) Labels {
 	return ret
 }
 
+// WithLabels returns a new labels.Labels from ls that contains labels not matching names.
+// 'names' have to be sorted in ascending order.
 func (ls Labels) WithoutLabels(names ...string) Labels {
 	var ret Labels
 	j := 0
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -17,7 +17,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"sort"
-	"strconv"
+	"sync"
 
 	"github.com/cespare/xxhash"
 )
@@ -31,6 +31,13 @@ const (
 	BucketLabel  = "le"
 	InstanceName = "instance"
 )
+
+var bufPool = sync.Pool{
+	New: func() interface{} {
+		// 96 seems to be the best for the benchmarks but that might not mean anything for real scenarios
+		return bytes.NewBuffer(make([]byte, 0, 96))
+	},
+}
 
 // Label is a key/value pair of strings.
 type Label struct {
@@ -46,7 +53,8 @@ func (ls Labels) Swap(i, j int)      { ls[i], ls[j] = ls[j], ls[i] }
 func (ls Labels) Less(i, j int) bool { return ls[i].Name < ls[j].Name }
 
 func (ls Labels) String() string {
-	var b bytes.Buffer
+	var ret string
+	b := bufPool.Get().(*bytes.Buffer)
 
 	b.WriteByte('{')
 	for i, l := range ls {
@@ -56,11 +64,16 @@ func (ls Labels) String() string {
 		}
 		b.WriteString(l.Name)
 		b.WriteByte('=')
-		b.WriteString(strconv.Quote(l.Value))
+		b.WriteByte('"')
+		b.WriteString(l.Value)
+		b.WriteByte('"')
+
 	}
 	b.WriteByte('}')
-
-	return b.String()
+	ret = b.String()
+	b.Reset()
+	bufPool.Put(b)
+	return ret
 }
 
 // MarshalJSON implements json.Marshaler.
@@ -175,7 +188,8 @@ func (ls Labels) HashWithoutLabels(b []byte, names ...string) (uint64, []byte) {
 // WithLabels returns a new labels.Labels from ls that only contains labels matching names.
 // 'names' have to be sorted in ascending order.
 func (ls Labels) WithLabels(names ...string) Labels {
-	var ret Labels
+	ret := make([]Label, 0, len(ls))
+
 	i, j := 0, 0
 	for i < len(ls) && j < len(names) {
 		if names[j] < ls[i].Name {
@@ -194,7 +208,8 @@ func (ls Labels) WithLabels(names ...string) Labels {
 // WithLabels returns a new labels.Labels from ls that contains labels not matching names.
 // 'names' have to be sorted in ascending order.
 func (ls Labels) WithoutLabels(names ...string) Labels {
-	var ret Labels
+	ret := make([]Label, 0, len(ls))
+
 	j := 0
 	for i := range ls {
 		for j < len(names) && names[j] < ls[i].Name {
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -18,7 +18,6 @@ import (
 	"encoding/json"
 	"sort"
 	"strconv"
-	"sync"
 	"unsafe"
 
 	"github.com/cespare/xxhash"
@@ -34,20 +33,6 @@ const (
 	InstanceName = "instance"
 )
 
-var bufPool = sync.Pool{
-	New: func() interface{} {
-		// 96 seems to be the best for the benchmarks but that might not mean anything for real scenarios
-		return bytes.NewBuffer(make([]byte, 0, 96))
-	},
-}
-
-var yoloPool = sync.Pool{
-	New: func() interface{} {
-		// 96 seems to be the best for the benchmarks but that might not mean anything for real scenarios
-		return bytes.NewBuffer(make([]byte, 0, 96))
-	},
-}
-
 // Label is a key/value pair of strings.
 type Label struct {
 	Name, Value string
@@ -62,8 +47,7 @@ func (ls Labels) Swap(i, j int)      { ls[i], ls[j] = ls[j], ls[i] }
 func (ls Labels) Less(i, j int) bool { return ls[i].Name < ls[j].Name }
 
 func (ls Labels) String() string {
-	var ret string
-	b := bufPool.Get().(*bytes.Buffer)
+	var b bytes.Buffer
 
 	b.WriteByte('{')
 	for i, l := range ls {
@@ -76,16 +60,14 @@ func (ls Labels) String() string {
 		b.WriteString(strconv.Quote(l.Value))
 	}
 	b.WriteByte('}')
-	ret = b.String()
-	b.Reset()
-	bufPool.Put(b)
-	return ret
+	return b.String()
 }
 
-func (ls Labels) NoRenderString() string {
-	var ret string
-	b := bufPool.Get().(*bytes.Buffer)
-
+// NoRenderString returns ls as a string.
+// It uses an byte invalid character as a separator and so
+// should not be used for printing.
+func (ls Labels) NoRenderString(b *bytes.Buffer) string {
+	b.Reset()
 	b.WriteByte('{')
 	for i, l := range ls {
 		if i > 0 {
@@ -94,25 +76,23 @@ func (ls Labels) NoRenderString() string {
 		}
 		b.WriteString(l.Name)
 		b.WriteByte('=')
-		b.WriteByte('"')
+		b.WriteByte(sep)
 		b.WriteString(l.Value)
-		b.WriteByte('"')
+		b.WriteByte(sep)
 	}
 	b.WriteByte('}')
-	ret = b.String()
-	b.Reset()
-	bufPool.Put(b)
-	return ret
+	return b.String()
 }
 
 func yoloString(b []byte) string {
 	return *((*string)(unsafe.Pointer(&b)))
 }
 
-func (ls Labels) YoloString() string {
-	b := yoloPool.Get().(*bytes.Buffer)
+// YoloString returns ls as an unsafe pointer to b's bytes casted to a string.
+// This is done to save on allocations that would happen during b.String() calls.
+// It uses an byte invalid character as a separator and so should not be used for printing.
+func (ls Labels) YoloString(b *bytes.Buffer) string {
 	b.Reset()
-
 	b.WriteByte('{')
 	for i, l := range ls {
 		if i > 0 {
@@ -121,12 +101,11 @@ func (ls Labels) YoloString() string {
 		}
 		b.WriteString(l.Name)
 		b.WriteByte('=')
-		b.WriteByte('"')
+		b.WriteByte(sep)
 		b.WriteString(l.Value)
-		b.WriteByte('"')
+		b.WriteByte(sep)
 	}
 	b.WriteByte('}')
-	yoloPool.Put(b)
 	return yoloString(b.Bytes())
 }
 
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -172,6 +172,39 @@ func (ls Labels) HashWithoutLabels(b []byte, names ...string) (uint64, []byte) {
 	return xxhash.Sum64(b), b
 }
 
+// WithLabels returns a new labels.Labels from ls that only contains labels matching names.
+func (ls Labels) WithLabels(names ...string) Labels {
+	var ret Labels
+	i, j := 0, 0
+	for i < len(ls) && j < len(names) {
+		if names[j] < ls[i].Name {
+			j++
+		} else if ls[i].Name < names[j] {
+			i++
+		} else {
+			ret = append(ret, ls[i])
+			i++
+			j++
+		}
+	}
+	return ret
+}
+
+func (ls Labels) WithoutLabels(names ...string) Labels {
+	var ret Labels
+	j := 0
+	for i := range ls {
+		for j < len(names) && names[j] < ls[i].Name {
+			j++
+		}
+		if ls[i].Name == MetricName || (j < len(names) && ls[i].Name == names[j]) {
+			continue
+		}
+		ret = append(ret, ls[i])
+	}
+	return ret
+}
+
 // Copy returns a copy of the labels.
 func (ls Labels) Copy() Labels {
 	res := make(Labels, len(ls))
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -22,14 +22,15 @@ import (
 	"github.com/cespare/xxhash"
 )
 
-const sep = '\xff'
-
 // Well-known label names used by Prometheus components.
 const (
 	MetricName   = "__name__"
 	AlertName    = "alertname"
 	BucketLabel  = "le"
 	InstanceName = "instance"
+
+	sep      = '\xff'
+	labelSep = '\xfe'
 )
 
 // Label is a key/value pair of strings.
@@ -66,19 +67,16 @@ func (ls Labels) String() string {
 // It uses an byte invalid character as a separator and so should not be used for printing.
 func (ls Labels) Bytes(buf []byte) []byte {
 	b := bytes.NewBuffer(buf[:0])
-	b.WriteByte('{')
+	b.WriteByte(labelSep)
 	for i, l := range ls {
 		if i > 0 {
-			b.WriteByte(',')
-			b.WriteByte(' ')
+			b.WriteByte(sep)
 		}
 		b.WriteString(l.Name)
-		b.WriteByte('=')
 		b.WriteByte(sep)
 		b.WriteString(l.Value)
-		b.WriteByte(sep)
 	}
-	b.WriteByte('}')
+	b.WriteByte(labelSep)
 	return b.Bytes()
 }
 
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -17,7 +17,9 @@ import (
 	"bytes"
 	"encoding/json"
 	"sort"
+	"strconv"
 	"sync"
+	"unsafe"
 
 	"github.com/cespare/xxhash"
 )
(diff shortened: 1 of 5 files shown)


roidelapluie:
Should we also update Hash() to use this?

brian-brazil:
We should look at eliminating the hashing, rather than changing the function.

cstyan:
> We should look at eliminating the hashing, rather than changing the function.

Can you elaborate as to how that would work? Just compare labels.Labels and ensure there's at least one unique label on one side?

Also, are you opposed to changing the hashing function where I have in order to fix the issue in the short term, assuming we move to go1.14?

brian-brazil:
Changing the hash function only changes what the collisions are.

cstyan:
Given that this is what Go uses internally for maps I don't see the issue with merging this, given that it should be less prone to collisions that what we have now, and implementing something better that doesn't use hashes in the future.

brian-brazil:
But Go has collision handling, so making presumptions about its collision properties merely because Go uses it seems a bit of a leap.

roidelapluie:
> But Go has collision handling, so making presumptions about its collision properties merely because Go uses it seems a bit of a leap.

 The hash functions are collision-resistant but not cryptographically secure. (See crypto/sha256 and crypto/sha512 for cryptographic use.) 
https://golang.org/pkg/hash/maphash/

cstyan:
It's been 9 months since the original issue was filed and no one's come forward with an implementation other than replacing the hash function. Should we not just move forward with one of those and file an issue for the other implementation?

roidelapluie:
I'd be in favor of using maphash as it is documented as collision resistant (which is the issue here). I'd still like to see it applied everywhere then.

brian-brazil:
Just because noone has yet created the correct fix doesn't mean that we should implement an incorrect fix that will itself cause unknown problems. Collisions are inevitable.

pstibrany:
> Just because noone has yet created the correct fix doesn't mean that we should implement an incorrect fix that will itself cause unknown problems. Collisions are inevitable.

Using a collision resistant hash function means that you would need to create *large* number of hash values to get a guaranteed collision (think 7x10^9 for 50% collision chance for 64-bit hash space). By that time, you will run out of memory first. Probability of getting a collision with random values is much lower than with "traditional" hash function. That still sounds like an improvement to me.

brian-brazil:
The rule of thumb is the square root for a 50% collision chance. Given the number of users we have, we're going to have collisions.

pstibrany:
> The rule of thumb is the square root for a 50% collision chance. Given the number of users we have, we're going to have collisions.

Yeah, but square root of 64-bit hash space is still 5.1x10^9 values before you get there. (I misremembered, and wrote 7 above).

beorn7:
With no PR in sight to implement collision detection, I'm all for a simple change that makes it less likely to run into a hash collision.

However, is the new `maphash` in Go 1.14 really more collision-resistant than the `xxhash` we are using now? [xxhash](https://cyan4973.github.io/xxHash/) claims to score perfectly in the [SMHasher](https://github.com/aappleby/smhasher/wiki/SMHasher) test. I don't know if that means it is close to the above mentioned square-root probability. And of course, we cannot really say how close the `maphash` hash function is (I suspect the actual function used might depend on the hardware).

Has anybody any evidence how the two compare?

cstyan:
@brian-brazil so it sounds like the "correct solution" is to replace the maps in the engine that are `uint64 -> labels.Labels` with maps of `labels.String() -> labels.Labels` or something similar. Without hashing the label set ourselves we could just rely on the runtimes collision handling for maps.

brian-brazil:
Yeah, something like that would be the way to do it. The main question is what the CPU hit will be.

cstyan:
Okay I think I have something working via maps, I'll clean it up and push in a bit.

@brian-brazil AFACIT for `TestEvaluations` the test 
`eval instant at 1m absent_over_time(rate(nonexistant[5m])[5m:])` from `testdata/functions.test` fails even on master. Is that something we should be looking into?

roidelapluie:
```
~/go/src/github.com/prometheus/prometheus/promql $ git rev-parse HEAD
0c4ec8d9dd1c64363dce59bed45a759f7f2149df
~/go/src/github.com/prometheus/prometheus/promql $ go test
PASS
ok      github.com/prometheus/prometheus/promql 0.361s
```

it passes for me

roidelapluie:
@cstyan this test is not in git :) probably some test on your machine :)

cstyan:
It only seems to fail if you specifically run `TestEvaluations`

```
prometheus/promql branch  master [$?] go v1.14.1 ➜ git rev-parse HEAD
0c4ec8d9dd1c64363dce59bed45a759f7f2149df

prometheus/promql branch  master [$?] go v1.14.1 ➜ go test -run TestEvaluations -v
=== RUN   TestEvaluations
    TestEvaluations: promql_test.go:31: unexpected error: error evaluating query "absent_over_time(rate(nonexistant[5m])[5m:])" (line 590): unexpected error: runtime error: integer divide by zero
--- FAIL: TestEvaluations (0.14s)
FAIL
exit status 1
FAIL	github.com/prometheus/prometheus/promql	0.143s

prometheus/promql branch  master [$?] go v1.14.1 ➜ go test ./...
ok  	github.com/prometheus/prometheus/promql	(cached)
ok  	github.com/prometheus/prometheus/promql/parser	(cached)```

File: pkg/labels/labels.go [hunk 1]
```diff
@@ -172,6 +172,39 @@ func (ls Labels) HashWithoutLabels(b []byte, names ...string) (uint64, []byte) {
 	return xxhash.Sum64(b), b
 }
 
+// WithLabels returns a new labels.Labels from ls that only contains labels matching names.
```
brian-brazil, line 175: This should mention that the names have to be sorted.

pkg/labels/labels.go
```diff
@@ -172,6 +172,39 @@ func (ls Labels) HashWithoutLabels(b []byte, names ...string) (uint64, []byte) {
...
+		} else {
+			ret = append(ret, ls[i])
+			i++
+			j++
+		}
+	}
+	return ret
+}
+
+func (ls Labels) WithoutLabels(names ...string) Labels {
```
brian-brazil, line 193: This should have a comment.

promql/engine.go
```diff
@@ -1749,7 +1755,7 @@ func resultMetric(lhs, rhs labels.Labels, op parser.ItemType, matching *parser.V
 	}
 
 	ret := lb.Labels()
-	enh.resultMetric[h] = ret
+	enh.resultMetric[lhs.String()+rhs.String()] = ret
```
brian-brazil, line 1758: Does it make a difference if we calculate the key only once?

promql/engine.go
```diff
@@ -1702,19 +1701,26 @@ func signatureFunc(on bool, names ...string) func(labels.Labels) uint64 {
 	}
 }
 
+func signatureFuncString(on bool, names ...string) func(labels.Labels) string {
```
brian-brazil, line 1704: Why not replace signatureFunc and fix all of this while you're at it?

cstyan:
https://gist.github.com/cstyan/d80b13d2c5e62447abbdcf3c6bb6417c @brian-brazil the alloc/bytes benchmarks don't look great. I'll let you comment on whether the changes are acceptable based on the benchmarks, or if there's something I'm doing in an inefficient way. I've made the changes to signatureFunc and various maps based on my understanding of our discussion in earlier comments.

brian-brazil:
I'm more concerned that it's causing those queries to take twice as long. Are there any easy wins for optimsation?

promql/engine.go
brian-brazil: We don't need the "String" now.

cstyan:
Made some optimizations in the latest commit, mostly around using a `sync.Pool` in for the key in `resultMetric` and for the buffers in `labels.go` `String`. 

Here are the benchmark comparisons of master vs the latest commit and master vs the previous commit: https://gist.github.com/cstyan/20b58011ca95694fd56ec89d49000729

**Highlights:**

Worst ns/op in previous commit 
```
benchmark                                                                                                             old ns/op      new ns/op      delta
BenchmarkRangeQuery/expr=a_hundred_-_b_hundred,steps=10000-8                                                          868686865      2352684729     +170.83%
```

And worst ns/op in latest commit
```
benchmark                                                                                                             old ns/op      new ns/op      delta
BenchmarkRangeQuery/expr=a_ten_-_b_ten,steps=10000-8                                                                  80429636       130171151      +61.84%
```

Worst allocs in previous commit
```
benchmark                                                                                                             old allocs     new allocs     delta
BenchmarkRangeQuery/expr=a_hundred_-_b_hundred,steps=10000-8                                                          140992         21150098       +14900.92%
```

Worst allocs in latest commit
```
benchmark                                                                                                             old allocs     new allocs     delta
BenchmarkRangeQuery/expr=a_hundred_-_b_hundred,steps=10000-8                                                          140992         5176815        +3571.71%
```

And finally worst bytes/op in previous commit
```
benchmark                                                                                                             old bytes     new bytes     delta
BenchmarkRangeQuery/expr=a_hundred_-_b_hundred,steps=10000-8                                                          22621664      598825712     +2547.13%
```

And worst bytes/op in latest commit
```
benchmark                                                                                                             old bytes     new bytes     delta
BenchmarkRangeQuery/expr=histogram_quantile(0.9,_rate(h_hundred[5m])),steps=1000-8                                    7728456       86935588      +1024.88%
```
This is a different expression though, that expression in the previous commit
```
BenchmarkRangeQuery/expr=histogram_quantile(0.9,_rate(h_hundred[5m])),steps=1000-8                                    7728456       138916932     +1697.47%
```

Thanks to @cyriltovena and @peterbourgon for their help with benchmarks and potential optimizations

pkg/labels/labels_test.go
```diff
@@ -485,3 +486,39 @@ func TestLabels_Has(t *testing.T) {
 		testutil.Equals(t, test.expected, got, "unexpected comparison result for test case %d", i)
 	}
 }
+
+func BenchmarkLabelsString(b *testing.B) {
```
cstyan, line 490: this benchmark doesn't need to stay but it was useful for optimizing the `String` function so far

pkg/labels/labels.go
```diff
@@ -56,11 +64,16 @@ func (ls Labels) String() string {
 		}
 		b.WriteString(l.Name)
 		b.WriteByte('=')
-		b.WriteString(strconv.Quote(l.Value))
```
cstyan, line 59: getting rid of `strconv` call here was a big perf boost

brian-brazil: That strconv is required for correctness, as this is used to render the labels for humans.

brian-brazil:
Instead of introducing a Pool, I'd suggest having a variant of String which you can pass a buffer to.

pkg/labels/labels.go
cstyan: is this because of what characters people can use in label values? The strconv call is very expensive in terms of allocs and cpu time. We could have a new function we don't use when rendering/logging for the strings we store in the map within promql engine?

beorn7: It's even necessary for correctness in the query engine as otherwise certain bizarre label values would be indistinguishable from regular metrics. Compare `foo="bar",dings="bums"` to `foo="bar\",dings=\"bums"`.

My suggestion is to not use the normal `String()` method for the hash replacement, but a special one that doesn't quote but uses a byte invalid in UTF-8 as a separator, similar to https://github.com/prometheus/pushgateway/blob/master/storage/diskmetricstore.go#L458-L479 .

beorn7: (Note: I'm only referring to the approach with the separator byte to avoid expensive quoting. The linked code is _not_ meant as an example for efficient string allocation handling. That's a different story.)

cstyan:
Here's the latest benchmark: https://gist.github.com/cstyan/47f9641e83f1c8c9d62f340db72f5099

pkg/labels/labels.go
```diff
@@ -59,8 +76,58 @@ func (ls Labels) String() string {
...
+		}
+		b.WriteString(l.Name)
+		b.WriteByte('=')
+		b.WriteByte('"')
+		b.WriteString(l.Value)
+		b.WriteByte('"')
+	}
+	b.WriteByte('}')
+	yoloPool.Put(b)
+	return yoloString(b.Bytes())
```
brian-brazil, line 130: This isn't a Yolostring, there's nothing backing it that's hanging around. This is just unsafe as b has been put back in the pool.

promql/engine.go
```diff
@@ -852,20 +856,23 @@ func (enh *EvalNodeHelper) dropMetricName(l labels.Labels) labels.Labels {
...
+	f := signatureFuncString(on, names...)
+	return func(l labels.Labels) string {
+		ls := l.YoloString()
+		ret, ok := enh.sigf[ls]
 		if ok {
 			return ret
 		}
 		ret = f(l)
-		enh.sigf[h] = ret
+		enh.sigf[l.NoRenderString()] = ret
```
brian-brazil, line 875: Why are you calculating the same thing twice?

cstyan:
@brian-brazil @beorn7 thanks for the review and suggestions. I think I've addressed all the comments, the `labels.Builder` and `bytes.Buffer` used are now both part of the `EvalNodeHelper` struct, and Labels strings functions used in the engine for the replacement of hashing are using hex 255 as the separator.

Here's the benchmarks for master vs the latest commit: https://gist.github.com/cstyan/a9d35abab849bd0ef4ac8f0dfe8baec0

They look off too me, so I will run them again in the morning to confirm.

pkg/labels/labels.go
```diff
@@ -59,10 +60,55 @@ func (ls Labels) String() string {
 		b.WriteString(strconv.Quote(l.Value))
 	}
 	b.WriteByte('}')
+	return b.String()
+}
 
+// NoRenderString returns ls as a string.
+// It uses an byte invalid character as a separator and so
+// should not be used for printing.
+func (ls Labels) NoRenderString(b *bytes.Buffer) string {
```
brian-brazil, line 69: You don't need two functions, return the buffer and leave the memory management to the caller.

promql/functions.go
```diff
@@ -598,10 +599,11 @@ func funcPredictLinear(vals []parser.Value, args parser.Expressions, enh *EvalNo
 func funcHistogramQuantile(vals []parser.Value, args parser.Expressions, enh *EvalNodeHelper) Vector {
 	q := vals[0].(Vector)[0].V
 	inVec := vals[1].(Vector)
-	sigf := enh.signatureFunc(false, excludedLabels...)
+	b := bytes.Buffer{}
```
brian-brazil, line 602: Why not use the one in enh?

promql/engine.go
```diff
@@ -852,20 +857,23 @@ func (enh *EvalNodeHelper) dropMetricName(l labels.Labels) labels.Labels {
...
+	f := signatureFunc(on, &enh.lblBuf, names...)
+	return func(l labels.Labels) string {
+		ls := l.YoloString(&enh.lblBuf)
+		ret, ok := enh.sigf[ls]
 		if ok {
 			return ret
 		}
 		ret = f(l)
-		enh.sigf[h] = ret
+		enh.sigf[l.NoRenderString(&enh.lblBuf)] = ret
```
brian-brazil, line 876: You're still calculating this twice, you already have this from `ls` it only needs to be converted to a string.

brian-brazil:
Performance wise it looks like binops are 20-25% slower, but otherwise we're in the noise (once my above comment on histogram_quantile is handled) so this seems okay.

promql/engine.go
cstyan: I haven't been able to get reuse of `ls` here to work properly without explicitly copying the byte buffer to a new variable: 
```		
var cp string
copy([]byte(cp), ls)
enh.sigf[cp] = ret
```

promql/functions.go
cstyan: thanks, I missed that enh was being passed in

promql/engine.go
brian-brazil: Hmm, maybe pass around a `[]byte` and only create a Buffer when we need it?

cstyan:
@brian-brazil everything I tried re: one function in labels.go that returns just the buffer or a byte slice instead of one (or two) that return a string/yolo string was significantly less performant than what's here right now. Like 60-65% slower instead of 25-30% slower.

If you feel strongly about either of those changes I can dig some more.

brian-brazil:
I played with it and didn't have any performance issues. I also spotted that you're using the buffer for two different things at the same time in resultMetrics, which needs to be fixed.

cstyan:
Thanks for your help again @brian-brazil. The only real change I made to what you pushed to your branch is not to use `yoloString`, as it seems for map looks go has a compiler optimization that should make that same pointer for us: https://golang.org/issue/3512

Here are the latest benchmarks: https://gist.github.com/cstyan/1a00155d189c1a2356d4638eedcff009

When running and comparing a small subset of the benchmarks, not making the extra `yoloString` call seemed to save us ~1-2% ns/op and a small amount of allocs.

pkg/labels/labels.go
```diff
@@ -59,10 +59,29 @@ func (ls Labels) String() string {
...
 	b.WriteByte('}')
-
 	return b.String()
 }
 
+// Bytes returns ls as a byte slice.
+// It uses an byte invalid character as a separator and so should not be used for printing.
+func (ls Labels) Bytes(buf []byte) []byte {
+	b := bytes.NewBuffer(buf[:0])
+	b.WriteByte('{')
```
roidelapluie, line 69: This

pkg/labels/labels.go
```diff
@@ -59,10 +59,29 @@ func (ls Labels) String() string {
...
+	for i, l := range ls {
+		if i > 0 {
+			b.WriteByte(',')
+			b.WriteByte(' ')
```
roidelapluie, line 73: This

pkg/labels/labels.go
```diff
@@ -59,10 +59,29 @@ func (ls Labels) String() string {
...
+		}
+		b.WriteString(l.Name)
+		b.WriteByte('=')
+		b.WriteByte(sep)
+		b.WriteString(l.Value)
+		b.WriteByte(sep)
+	}
+	b.WriteByte('}')
```
roidelapluie, line 81:
```suggestion
	b.WriteByte(sep)
```

pkg/labels/labels.go
```diff
@@ -59,10 +59,29 @@ func (ls Labels) String() string {
...
+func (ls Labels) Bytes(buf []byte) []byte {
+	b := bytes.NewBuffer(buf[:0])
+	b.WriteByte('{')
+	for i, l := range ls {
+		if i > 0 {
+			b.WriteByte(',')
+			b.WriteByte(' ')
+		}
+		b.WriteString(l.Name)
+		b.WriteByte('=')
```
roidelapluie, line 76: This

pkg/labels/labels.go
roidelapluie: This is for when we chain 2 labels together.

roidelapluie: Mmh maybe not use SEP as if we do not have existing label it would be confusting

pkg/labels/labels.go
```diff
@@ -59,10 +59,29 @@ func (ls Labels) String() string {
 		b.WriteString(strconv.Quote(l.Value))
 	}
 	b.WriteByte('}')
-
 	return b.String()
 }
 
+// Bytes returns ls as a byte slice.
+// It uses an byte invalid character as a separator and so should not be used for printing.
+func (ls Labels) Bytes(buf []byte) []byte {
```
beorn7, line 67: Relating to @roidelapluie's comments:

This function doesn't need to insert `{`, `}`, or `=` at all.

For empty labels, it doesn't need to return anything.

For the other cases, it returns, for each label, `sep + l.Name + sep + l.Value`. That should be unique for each possible unique label set.

roidelapluie: } is needed when we concatenant labels together, e.g. to see the difference between:

{a="b",b="c"}{c="d"}
{a="b"}{b="c",c="d"}

> a$b$b$c$c$d$

We need an extra separator:

> a$b$$b$c$c$d$
> a$b$b$c$$c$d$

beorn7: Even if we did, would it be a difference?

cstyan: > weelll I don't think we ever do this so this is not needed

We do exactly this in `resultMetric`. Keeping `{ ... }` to differentiate between lhs and rhs when concatenating in `resultMetric` but removing `, ` and `=` so that the bytes are just `{nameSEPvalueSEPnameSEPvalue}` for the main benchmark I usually run buys us a little bit of ns/op improvement but seems to both reduce and increase bytes/allocs.

beorn7: Then we have to do something else. `}` and `{` could be part of a label value. How about three separator bytes to separate lhs and rhs?

roidelapluie: Or another separator

brian-brazil: Keep in mind that label names cannot be empty.

cstyan: > Then we have to do something else. } and { could be part of a label value. How about three separator bytes to separate lhs and rhs?

> Or another separator

A different invalid character separator seems like the best option.

> Keep in mind that label names cannot be empty.

@brian-brazil are you point out that at this point label names are guaranteed to not be empty, or saying we need to check that names are not empty here? Just had a quick look but I don't see anywhere else we're checking for empty names. `labels.WithoutEmpty` only looks for empty values.

brian-brazil: My point is that could be a useful property for ensuring uniqueness when the strings are concatenated.

beorn7: Given that there are plenty of bytes that can never occur in a valid UTF-8 stream, I agree with @cstyan that just using another invalid byte is the easiest.

pkg/labels/labels.go
```diff
@@ -66,19 +67,16 @@ func (ls Labels) String() string {
...
+			b.WriteByte(sep)
 		}
 		b.WriteString(l.Name)
-		b.WriteByte('=')
 		b.WriteByte(sep)
 		b.WriteString(l.Value)
-		b.WriteByte(sep)
 	}
-	b.WriteByte('}')
+	b.WriteByte(labelSep)
```
beorn7, line 79: I think we don't need this trailing byte. Just have the leading byte should be enough as that marks the beginning of a new label set in case of concatenation.

cstyan:
@brian-brazil PTAL

promql/engine.go
```diff
@@ -1692,42 +1695,44 @@ func (ev *evaluator) VectorBinop(op parser.ItemType, lhs, rhs Vector, matching *
...
-	if ret, ok := enh.resultMetric[h]; ok {
-		return ret
+		enh.resultMetric = make(map[string]labels.Labels, len(enh.out))
 	}
+	if enh.lb == nil {
+		enh.lb = labels.NewBuilder(lhs)
+	} else {
+		enh.lb.Reset(lhs)
+	}
+	// enh.lblBuf.Reset()
```
brian-brazil, line 1721: Remove

PR #7059, opened 2020-03-26
storage: Added support for chunk iterators: Remote read client, readyStorage, fanout.

bwplotka:
## Major Changes

* Added ChunkQueryable implementation to fanout and readyStorage.
* Added more comments.
* Changed NewVerticalChunkSeriesMerger to CompactingChunkSeriesMerger, removed tiny interface by reusing VerticalSeriesMergeFunc for overlapping algorithm for
both chunks and series, for both querying and compacting (!) + made sure duplicates are merged.
* Added ErrChunkSeriesSet
* Added Samples interface for seamless []promb.Sample to []tsdbutil.Sample conversion.
* Deprecating non chunks serieset based StreamChunkedReadResponses, added chunk one.
* Improved tests.
* Split remote client into Write (old storage) and read.
* Queryable client is now SampleAndChunkQueryable. Since we cannot use nice QueryableFunc I moved
all config based options to sampleAndChunkQueryableClient to aboid boilerplate.

In next commit: Changes for TSDB.

### DIFF
--- a/storage/fanout.go
+++ b/storage/fanout.go
@@ -14,6 +14,7 @@
 package storage
 
 import (
+	"bytes"
 	"container/heap"
 	"context"
 	"sort"
@@ -24,7 +25,6 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
-
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/tsdb/chunkenc"
 	"github.com/prometheus/prometheus/tsdb/chunks"
@@ -94,10 +94,32 @@ func (f *fanout) Querier(ctx context.Context, mint, maxt int64) (Querier, error)
 			}
 			return nil, errs.Err()
 		}
-
 		secondaries = append(secondaries, querier)
 	}
 	return NewMergeQuerier(primary, secondaries, ChainedSeriesMerge), nil
+}
+
+func (f *fanout) ChunkQuerier(ctx context.Context, mint, maxt int64) (ChunkQuerier, error) {
+	primary, err := f.primary.ChunkQuerier(ctx, mint, maxt)
+	if err != nil {
+		return nil, err
+	}
+
+	secondaries := make([]ChunkQuerier, 0, len(f.secondaries))
+	for _, storage := range f.secondaries {
+		querier, err := storage.ChunkQuerier(ctx, mint, maxt)
+		if err != nil {
+			// Close already open Queriers, append potential errors to returned error.
+			errs := tsdb_errors.MultiError{err}
+			errs.Add(primary.Close())
+			for _, q := range secondaries {
+				errs.Add(q.Close())
+			}
+			return nil, errs.Err()
+		}
+		secondaries = append(secondaries, querier)
+	}
+	return NewMergeChunkQuerier(primary, secondaries, NewCompactingChunkSeriesMerger(ChainedSeriesMerge)), nil
 }
 
 func (f *fanout) Appender() Appender {
@@ -220,7 +242,7 @@ func NewMergeQuerier(primary Querier, secondaries []Querier, mergeFn VerticalSer
 //
 // In case of overlaps between the data given by primary + secondaries Selects, merge function will be used.
 // TODO(bwplotka): Currently merge will compact overlapping chunks with bigger chunk, without limit. Split it: https://github.com/prometheus/tsdb/issues/670
-func NewMergeChunkQuerier(primary ChunkQuerier, secondaries []ChunkQuerier, mergeFn VerticalChunkSeriesMergerFunc) ChunkQuerier {
+func NewMergeChunkQuerier(primary ChunkQuerier, secondaries []ChunkQuerier, mergeFn VerticalChunkSeriesMergeFunc) ChunkQuerier {
 	queriers := make([]genericQuerier, 0, len(secondaries)+1)
 	if primary != nil {
 		queriers = append(queriers, newGenericQuerierFromChunk(primary))
@@ -232,7 +254,7 @@ func NewMergeChunkQuerier(primary ChunkQuerier, secondaries []ChunkQuerier, merg
 	}
 
 	return &chunkQuerierAdapter{&mergeGenericQuerier{
-		mergeFn:  (&chunkSeriesMergerAdapter{VerticalChunkSeriesMergerFunc: mergeFn}).Merge,
+		mergeFn:  (&chunkSeriesMergerAdapter{VerticalChunkSeriesMergeFunc: mergeFn}).Merge,
 		queriers: queriers,
 	}}
 }
@@ -270,7 +292,7 @@ func (q *mergeGenericQuerier) Select(sortSeries bool, hints *SelectHints, matche
 	return &lazySeriesSet{create: create(seriesSets, q.mergeFn)}
 }
 
-func create(seriesSets []genericSeriesSet, mergeFn genericSeriesMergeFunc) func() (genericSeriesSet, bool) {
+func create(seriesSets []genericSeriesSet, mergeFunc genericSeriesMergeFunc) func() (genericSeriesSet, bool) {
 	// Returned function gets called with the first call to Next().
 	return func() (genericSeriesSet, bool) {
 		if len(seriesSets) == 1 {
@@ -292,9 +314,9 @@ func create(seriesSets []genericSeriesSet, mergeFn genericSeriesMergeFunc) func(
 			}
 		}
 		set := &genericMergeSeriesSet{
-			mergeFn: mergeFn,
-			sets:    seriesSets,
-			heap:    h,
+			mergeFunc: mergeFunc,
+			sets:      seriesSets,
+			heap:      h,
 		}
 		return set, set.Next()
 	}
@@ -361,8 +383,10 @@ func mergeTwoStringSlices(a, b []string) []string {
 
 // LabelNames returns all the unique label names present in the block in sorted order.
 func (q *mergeGenericQuerier) LabelNames() ([]string, Warnings, error) {
-	labelNamesMap := make(map[string]struct{})
-	var warnings Warnings
+	var (
+		labelNamesMap = make(map[string]struct{})
+		warnings      Warnings
+	)
 	for _, querier := range q.queriers {
 		names, wrn, err := querier.LabelNames()
 		if wrn != nil {
@@ -403,34 +427,36 @@ func (q *mergeGenericQuerier) Close() error {
 // It has to handle time-overlapped series as well.
 type VerticalSeriesMergeFunc func(...Series) Series
 
-// VerticalChunkSeriesMergerFunc returns merged chunk series implementation that merges series with same labels together.
-// It has to handle time-overlapped chunk series as well.
-type VerticalChunkSeriesMergerFunc func(...ChunkSeries) ChunkSeries
-
-// NewMergeSeriesSet returns a new SeriesSet that merges results of chkQuerierSeries SeriesSets.
-func NewMergeSeriesSet(sets []SeriesSet, merger VerticalSeriesMergeFunc) SeriesSet {
+// NewMergeSeriesSet returns a new SeriesSet that merges many SeriesSets together.
+func NewMergeSeriesSet(sets []SeriesSet, mergeFunc VerticalSeriesMergeFunc) SeriesSet {
 	genericSets := make([]genericSeriesSet, 0, len(sets))
 	for _, s := range sets {
 		genericSets = append(genericSets, &genericSeriesSetAdapter{s})
 
 	}
-	return &seriesSetAdapter{newGenericMergeSeriesSet(genericSets, (&seriesMergerAdapter{VerticalSeriesMergeFunc: merger}).Merge)}
+	return &seriesSetAdapter{newGenericMergeSeriesSet(genericSets, (&seriesMergerAdapter{VerticalSeriesMergeFunc: mergeFunc}).Merge)}
 }
 
-// NewMergeChunkSeriesSet returns a new ChunkSeriesSet that merges results of chkQuerierSeries ChunkSeriesSets.
-func NewMergeChunkSeriesSet(sets []ChunkSeriesSet, merger VerticalChunkSeriesMergerFunc) ChunkSeriesSet {
+// VerticalChunkSeriesMergeFunc returns merged chunk series implementation that merges potentially time-overlapping
+// chunk series with the same labels into single ChunkSeries.
+//
+// NOTE: It's up to implementation how series are vertically merged (if chunks are sorted, re-encoded etc).
+type VerticalChunkSeriesMergeFunc func(...ChunkSeries) ChunkSeries
+
+// NewMergeChunkSeriesSet returns a new ChunkSeriesSet that merges many SeriesSet together.
+func NewMergeChunkSeriesSet(sets []ChunkSeriesSet, mergeFunc VerticalChunkSeriesMergeFunc) ChunkSeriesSet {
 	genericSets := make([]genericSeriesSet, 0, len(sets))
 	for _, s := range sets {
 		genericSets = append(genericSets, &genericChunkSeriesSetAdapter{s})
 
 	}
-	return &chunkSeriesSetAdapter{newGenericMergeSeriesSet(genericSets, (&chunkSeriesMergerAdapter{VerticalChunkSeriesMergerFunc: merger}).Merge)}
+	return &chunkSeriesSetAdapter{newGenericMergeSeriesSet(genericSets, (&chunkSeriesMergerAdapter{VerticalChunkSeriesMergeFunc: mergeFunc}).Merge)}
 }
 
 // genericMergeSeriesSet implements genericSeriesSet.
 type genericMergeSeriesSet struct {
 	currentLabels labels.Labels
-	mergeFn       genericSeriesMergeFunc
+	mergeFunc     genericSeriesMergeFunc
 
 	heap        genericSeriesSetHeap
 	sets        []genericSeriesSet
@@ -441,8 +467,8 @@ type genericMergeSeriesSet struct {
 // series returned by the series sets when iterating.
 // Each series set must return its series in labels order, otherwise
 // merged series set will be incorrect.
-// Overlapping cases are merged using provided mergeFn.
-func newGenericMergeSeriesSet(sets []genericSeriesSet, mergeFn genericSeriesMergeFunc) genericSeriesSet {
+// Overlapped situations are merged using provided mergeFunc.
+func newGenericMergeSeriesSet(sets []genericSeriesSet, mergeFunc genericSeriesMergeFunc) genericSeriesSet {
 	if len(sets) == 1 {
 		return sets[0]
 	}
@@ -459,9 +485,9 @@ func newGenericMergeSeriesSet(sets []genericSeriesSet, mergeFn genericSeriesMerg
 		}
 	}
 	return &genericMergeSeriesSet{
-		mergeFn: mergeFn,
-		sets:    sets,
-		heap:    h,
+		mergeFunc: mergeFunc,
+		sets:      sets,
+		heap:      h,
 	}
 }
 
@@ -507,7 +533,7 @@ func (c *genericMergeSeriesSet) At() Labels {
 	for _, seriesSet := range c.currentSets {
 		series = append(series, seriesSet.At())
 	}
-	return c.mergeFn(series...)
+	return c.mergeFunc(series...)
 }
 
 func (c *genericMergeSeriesSet) Err() error {
@@ -549,10 +575,16 @@ func (h *genericSeriesSetHeap) Pop() interface{} {
 	return x
 }
 
-// ChainedSeriesMerge returns single series from many same series by chaining samples together.
-// In case of the timestamp overlap, the first overlapped sample is kept and the rest samples with the same timestamps
-// are dropped. We expect the same labels for each given series.
-// TODO(bwplotka): This has the same logic as tsdb.verticalChainedSeries. Remove this in favor of ChainedSeriesMerge in next PRs.
+// ChainedSeriesMerge returns single series from many same, potentially overlapping series by chaining samples together.
+// If one or more samples overlap, one sample from random overlapped ones is kept and all others with the same
+// timestamp are dropped.
+//
+// This works the best with replicated series, where data from two series are exactly the same. This does not work well
+// with "almost" the same data, e.g. from 2 Prometheus HA replicas. This is fine, since from the Prometheus perspective
+// this never happens.
+//
+// NOTE: Use this only when you see potentially overlapping series, as this introduces small overhead to handle overlaps
+// between series.
 func ChainedSeriesMerge(s ...Series) Series {
 	if len(s) == 0 {
 		return nil
@@ -580,8 +612,9 @@ func (m *chainSeries) Iterator() chunkenc.Iterator {
 	return newChainSampleIterator(iterators)
 }
 
-// chainSampleIterator is responsible to iterate over samples from different iterators of the same time series.
-// If one or more samples overlap, the first one is kept and all others with the same timestamp are dropped.
+// chainSampleIterator is responsible to iterate over samples from different iterators of the same time series in timestamps
+// order. If one or more samples overlap, one sample from random overlapped ones is kept and all others with the same
+// timestamp are dropped.
 type chainSampleIterator struct {
 	iterators []chunkenc.Iterator
 	h         samplesIteratorHeap
@@ -645,12 +678,13 @@ func (c *chainSampleIterator) Next() bool {
 }
 
 func (c *chainSampleIterator) Err() error {
+	var errs tsdb_errors.MultiError
 	for _, iter := range c.iterators {
 		if err := iter.Err(); err != nil {
-			return err
+			errs.Add(err)
 		}
 	}
-	return nil
+	return errs.Err()
 }
 
 type samplesIteratorHeap []chunkenc.Iterator
@@ -676,77 +710,76 @@ func (h *samplesIteratorHeap) Pop() interface{} {
 	return x
 }
 
-// VerticalChunkMergeFunc represents a function that merges multiple time overlapping chunks.
-// Passed chunks:
-// * have to be sorted by MinTime.
-// * have to be part of exactly the same timeseries.
-// * have to be populated.
-type VerticalChunksMergeFunc func(chks ...chunks.Meta) chunks.Iterator
-
-type verticalChunkSeriesMerger struct {
-	verticalChunksMerger VerticalChunksMergeFunc
+type compactChunkSeriesMerger struct {
+	mergeFunc VerticalSeriesMergeFunc
 
 	labels labels.Labels
 	series []ChunkSeries
 }
 
-// NewVerticalChunkSeriesMerger returns VerticalChunkSeriesMerger that merges the same chunk series into one or more chunks.
-// In case of the chunk overlap, given VerticalChunkMergeFunc will be used.
+// NewCompactingChunkSeriesMerger returns VerticalChunkSeriesMergeFunc that merges the same chunk series into single chunk series.
+// In case of the chunk overlaps, it compacts those into one or more time-ordered non-overlapping chunks with merged data.
+// Samples from overlapped chunks are merged using series vertical merge func.
 // It expects the same labels for each given series.
-func NewVerticalChunkSeriesMerger(chunkMerger VerticalChunksMergeFunc) VerticalChunkSeriesMergerFunc {
+//
+// NOTE: Use this only when you see potentially overlapping series, as this introduces small overhead to handle overlaps
+// between series.
+func NewCompactingChunkSeriesMerger(mergeFunc VerticalSeriesMergeFunc) VerticalChunkSeriesMergeFunc {
 	return func(s ...ChunkSeries) ChunkSeries {
 		if len(s) == 0 {
 			return nil
 		}
-		return &verticalChunkSeriesMerger{
-			verticalChunksMerger: chunkMerger,
-			labels:               s[0].Labels(),
-			series:               s,
+		return &compactChunkSeriesMerger{
+			mergeFunc: mergeFunc,
+			labels:    s[0].Labels(),
+			series:    s,
 		}
 	}
 }
 
-func (s *verticalChunkSeriesMerger) Labels() labels.Labels {
+func (s *compactChunkSeriesMerger) Labels() labels.Labels {
 	return s.labels
 }
 
-func (s *verticalChunkSeriesMerger) Iterator() chunks.Iterator {
+func (s *compactChunkSeriesMerger) Iterator() chunks.Iterator {
 	iterators := make([]chunks.Iterator, 0, len(s.series))
 	for _, series := range s.series {
 		iterators = append(iterators, series.Iterator())
 	}
-	return &chainChunkIterator{
-		overlappedChunksMerger: s.verticalChunksMerger,
-		iterators:              iterators,
-		h:                      nil,
+	return &compactChunkIterator{
+		mergeFunc: s.mergeFunc,
+		labels:    s.labels,
+		iterators: iterators,
 	}
 }
 
-// chainChunkIterator is responsible to chain chunks from different iterators of same time series.
-// If they are time overlapping overlappedChunksMerger will be used.
-type chainChunkIterator struct {
-	overlappedChunksMerger VerticalChunksMergeFunc
+// compactChunkIterator is responsible to compact chunks from different iterators of the same time series into single chainSeries.
+// If time-overlapping chunks are found, they are encoded and passed to series merge and encoded again into one bigger chunk.
+// TODO(bwplotka): Currently merge will compact overlapping chunks with bigger chunk, without limit. Split it: https://github.com/prometheus/tsdb/issues/670
+type compactChunkIterator struct {
+	mergeFunc VerticalSeriesMergeFunc
 
+	labels    labels.Labels
 	iterators []chunks.Iterator
-	h         chunkIteratorHeap
+
+	h chunkIteratorHeap
 }
 
-func (c *chainChunkIterator) At() chunks.Meta {
+func (c *compactChunkIterator) At() chunks.Meta {
 	if len(c.h) == 0 {
-		panic("chainChunkIterator.At() called after .Next() returned false.")
+		panic("compactChunkIterator.At() called after .Next() returned false.")
 	}
 
 	return c.h[0].At()
 }
 
-func (c *chainChunkIterator) Next() bool {
+func (c *compactChunkIterator) Next() bool {
 	if c.h == nil {
 		for _, iter := range c.iterators {
 			if iter.Next() {
 				heap.Push(&c.h, iter)
 			}
 		}
-
 		return len(c.h) > 0
 	}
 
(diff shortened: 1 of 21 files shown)


tsdb/db.go
```diff
@@ -1311,17 +1310,45 @@ func (db *DB) Querier(_ context.Context, mint, maxt int64) (storage.Querier, err
...
-		return &verticalQuerier{
-			querier: querier{
-				blocks: blockQueriers,
-			},
-		}, nil
+	return newQuerier(blockQueriers), nil
+}
+
+// Querier returns a new querier over the data partition for the given time range.
+// A goroutine must not handle more than one open Querier.
```
brian-brazil: I don't think this is what you meant to say

storage/fanout.go
```diff
@@ -696,32 +720,29 @@ func (h *samplesIteratorHeap) Pop() interface{} {
...
-// In case of the chunk overlap, given VerticalChunkMergeFunc will be used.
+// NewCompactingChunkSeriesMerger returns VerticalChunkSeriesMergeFunc that merges the same chunk series into one chunk series.
+// In case of the chunk overlaps, it compacts overlapping chunks into one or more time-ordered chunks with merged data.
+// Samples from overlapped chunks are merged using *series* (not chunk) vertical merge func.
 // It expects the same labels for each given series.
-func NewVerticalChunkSeriesMerger(chunkMerger VerticalChunksMergeFunc) VerticalChunkSeriesMergerFunc {
+//
+// TODO(bwplotka): Currently we have only ChainingSeriesMerge as overlapSeriesMergeFunc. This works the best with
+// replicated series, where data from two series are exactly the same. This does not work well with "almost" the same
+// data e.g from 2 Prometheus HA replicas. Propose penalty based algorithm from Thanos as alternative.
```
brian-brazil: From a Prometheus standpoint, non-identical data is not a sane setup for remote read. A merge must be a merge, without losing samples.

bwplotka: Yes agree. I just propose to add some alternative that will work for all cases. Will add your mentioned use case to comment.

brian-brazil: We shouldn't be documenting Thanos design decisions that don't apply to Prometheus inside Prometheus.

bwplotka: Generally agree, but what if someone want to merge two blocks from different Prometheus replicas (fair use case) without even Thanos?

bwplotka: I can remove Thanos mention but will mention about some potentiall improvement we can make for such use case, WDYT?

brian-brazil: Then they're going to have unsuable data, that it may be inefficient doesn't really matter as the use case isn't sane.

storage/fanout.go
```diff
@@ -91,7 +91,33 @@ func (f *fanout) Querier(ctx context.Context, mint, maxt int64) (Querier, error)
 		queriers = append(queriers, querier)
 	}
 
-	return NewMergeQuerier(primaryQuerier, queriers, ChainedSeriesMerge), nil
+	return NewMergeQuerier(primaryQuerier, queriers, ChainingSeriesMerge), nil
+}
+
+func (f *fanout) ChunkQuerier(ctx context.Context, mint, maxt int64) (ChunkQuerier, error) {
```
bwplotka: This deserves a separate file, but will move in next PR to have some diff here.

storage/fanout.go
bwplotka: Fair. Added comment to `ChainingSeriesMerge` instead. (:

tsdb/db.go
bwplotka: Copied comment, sorry. FIxing.

bwplotka: If you mean

 > // A goroutine must not handle more than one open Querier.

That was 1:1 copied from db.go ;p  Since I don't get what it means, I will remove it.

bwplotka:
Ok as proposed by @codesome offline I split this into one piece here to fix and prepare storage and remote packages. This is follow up to implement ChunkQuerier in tsdb and unify series set merges: https://github.com/prometheus/prometheus/pull/7069.

Will fix tests for both PRs  tomorrow

codesome:
Thanks for splitting!

bwplotka:
All tests should be fixed and this PR should be ready for careful review and merge :hugs: 

cc @brian-brazil @codesome @brian-brazil

storage/remote/read_test.go
```diff
@@ -231,182 +211,310 @@ func TestSeriesSetFilter(t *testing.T) {
...
+				labels.FromStrings("a", "b2", "region", "europe"),
 			},
+		},
+		{
+			name: "external labels specified, explicitly requested not europe",
+			mint: 1, maxt: 2,
 			matchers: []*labels.Matcher{
-				labels.MustNewMatcher(labels.MatchRegexp, "special", "label"),
+				labels.MustNewMatcher(labels.MatchNotEqual, "a", "something"),
+				// TODO(bwplotka): Is this what we want? I prometheus is region=europe and user requests region=us we
```
bwplotka: A bit surprised with this. Is this intended (this PR does not change any semantic, just I added a test case for this), @brian-brazil ? :thinking:

bwplotka:
Also cc @pstibrany and @pracucci

storage/remote/client.go
```diff
@@ -38,30 +39,71 @@ const maxErrMsgLen = 256
...
+			Name:      "remote_read_queries",
+			Help:      "The number of in-flight remote read queries.",
+			ConstLabels: prometheus.Labels{
+				remoteName: name,
+				endpoint:   conf.URL.String(),
+			},
+		},
+	)
+	if reg != nil {
+		reg.Unregister(readQueries)
```
brian-brazil: This doesn't seem right, you should never have to unreigster outside of tests.

storage/fanout.go
```diff
@@ -190,66 +215,80 @@ func (f *fanoutAppender) Rollback() (err error) {
...
+func NewMergeQuerier(queriers []Querier, primaryQuerier int, mergeFunc VerticalSeriesMergeFunc) Querier {
 	filtered := make([]genericQuerier, 0, len(queriers))
-	for _, querier := range queriers {
+	for i, querier := range queriers {
 		if _, ok := querier.(noopQuerier); !ok && querier != nil {
 			filtered = append(filtered, newGenericQuerierFrom(querier))
+			continue
+		}
+		if i == primaryQuerier {
+			// Primary was filtered, no primary anymore.
```
brian-brazil: How can this happen?

bwplotka: Well any consumer of this function can do this anytime, but just specifying `MergeQuerier([]Querier{NoopQuerier()}, 0, ...)` no?

storage/remote/client.go
bwplotka: Hm, what're rationales behind this?

Because otherwise I would need to move this gauge to storage (which is fair I guess)

storage/fanout.go
brian-brazil: Do we have any of those in Prometheus, or are you keeping this general for Thanos?

storage/remote/client.go
brian-brazil: A given metric should exist for the lifetime of a process, if you're unregistering then that implies something is off with the metric lifecycle.

storage/fanout.go
bwplotka: We don't have at the current moment, but what if write code that does it? We don't use this at all for Thanos.

storage/remote/client.go
bwplotka: oke will fix

storage/fanout.go
brian-brazil: We can worry about it then, as-is we always have a primary querier and it's always present.

bwplotka:
Ok, fixed @brian-brazil  (:

storage/fanout.go
brian-brazil: The primaryQuerier argument to mergeGenericQuerier seemed unneeded now, and we're not handling it correctly now anyway.

bwplotka:
Friendly ping @brian-brazil @krasi-georgiev (:

storage/fanout.go
brian-brazil: This still doesn't look correct.

Sudhar287:
Hello bwplotka!
You had [asked](https://github.com/prometheus/prometheus/pull/6760#pullrequestreview-383385739) me to work on the `Iterator` function in `storage/series.go`. 
I think a lot of my code from that PR can simply be moved over to this one. 
I have a question though. Are you planning to commit the basic structure of the test code (in `storage/series_test.go`) you would like me to follow for the `Iterator` function? I'm asking this because I think it's better to have the structure that you expect rather than going back and forth in a couple of PRs. :) Any examples from other Prometheus test files that I should follow?

storage/fanout.go
bwplotka: looking

bwplotka: Hm, what do you mean?

PrimaryQuerier argument actually tells us if it should be error or warning e.g:

```
	           if q.primaryQuerier == 0 {
				return nil, w, err
			}
			// Nil will be converted to noop in adapter.
			return nil, append(w, err), nil
```

and

```
                if r.selectError != nil {
			q.failedQueriers[q.queriers[r.i]] = struct{}{}
			if r.i != q.primaryQuerier {
				warnings = append(warnings, r.selectError)
				continue
			}
			priErr = r.selectError
			continue
		}
```
So it's used, and to me it looks correctly.

bwplotka: @brian-brazil

bwplotka: What's broken it's that the first if `len(queries) == 1` is not needed.

bwplotka: Removed.

bwplotka:
@Sudhar287 I `proposed` rather. (: 

But I think I have a better idea. Looks regarding this code it's a bit risky as it's still in review and can change. If you want to help I would recommend more important feature: https://github.com/prometheus/prometheus/issues/5926  WDYT? (: 

Friendly ping for review @brian-brazil @codesome @krasi-georgiev, this PR sits here for a week already.

storage/fanout.go
brian-brazil: This still doesn't look right (warnigns vs errors), and even if it is it's going to be fragile. I think passing the primary querier as an index rather than the actual object is the fundamental issue, so would suggest going back to having a primary querier and a list of secondaries.

bwplotka: Hm tend to agree with this, but the previous logic was extremely surprising as we expected caller to fill all queries + primary additionally.

I think I would be fine with primary and secondaries as different args, I think that would  work for both of all (:

storage/fanout.go
```diff
@@ -190,48 +215,54 @@ func (f *fanoutAppender) Rollback() (err error) {
...
-// NewMergeQuerier will return NoopQuerier if no queriers are passed to it
-// and will filter NoopQueriers from its arguments, in order to reduce overhead
-// when only one querier is passed.
-// The difference between primary and secondary is as follows: f the primaryQuerier returns an error, query fails.
-// For secondaries it just return warnings.
-func NewMergeQuerier(primaryQuerier Querier, queriers []Querier, mergeFunc VerticalSeriesMergeFunc) Querier {
+// NewMergeQuerier returns a new Querier that merges results of given slice of queriers.
+// NewMergeQuerier will return NoopQuerier if no queriers or noopQueriers are passed.
+// Function allows to optionally mark certain querier in the slice to be primary one. The difference between primary and
+// rest is as follows: if the primary querier returns an error, query fails. For secondaries it just return warnings.
```
brancz, line 227: "query fails" -> "the query fails"

"it just return" -> "it just returns"

storage/fanout.go
```diff
@@ -190,48 +215,54 @@ func (f *fanoutAppender) Rollback() (err error) {
...
+// Argument 'primaryQuerier' set to -1 means no primary querier.
+
+// In case of overlaps in the returned data by all queriers, mergeFunc will be used.
+func NewMergeQuerier(queriers []Querier, primaryQuerier int, mergeFunc VerticalSeriesMergeFunc) Querier {
```
brancz, line 231: this seems like quite an odd API, why not just pass the primary querier separately from the secondary ones?

bwplotka: We discussed it here: https://github.com/prometheus/prometheus/pull/7059#discussion_r400914446

I think at the end that's the plan, will update in ~2h/tomorrow (:

storage/remote/client.go
```diff
@@ -38,30 +39,54 @@ const maxErrMsgLen = 256
...
+// ClientConfig configures a client.
 type ClientConfig struct {
 	URL              *config_util.URL
 	Timeout          model.Duration
 	HTTPClientConfig config_util.HTTPClientConfig
 }
 
-// NewClient creates a new Client.
-func NewClient(remoteName string, conf *ClientConfig) (*Client, error) {
+// TODO(bwplotka): Add streamed chunked remote read (add issue link).
```
brancz, line 59: did you already open this issue?

bwplotka: yes, will update link: https://github.com/prometheus/prometheus/issues/5926 forgot

storage/remote/read.go
```diff
@@ -219,6 +199,40 @@ func (q externalLabelsQuerier) addExternalLabels(ms []*labels.Matcher) ([]*label
 	return ms, el
 }
 
+// LabelValues implements storage.Querier and is a noop.
```
brancz, line 202: I can't find it right now but I feel like I've seen an issue or WIP PR about this.

storage/series.go
```diff
@@ -128,11 +138,83 @@ func (c *chunkSetToSeriesSet) Err() error {
...
+func (c *seriesSetToChunkSet) Err() error {
+	return c.SeriesSet.Err()
+}
+
+type seriesToChunkEncoder struct {
+	Series
+}
+
+// TODO(bwplotka): Currently encoder will just naively build one chunk, without limit. Split it: https://github.com/prometheus/tsdb/issues/670
+// DO IT NOW.
```
brancz, line 183: :thinking: ?

brancz:
Largely looks reasonable. Have we run prombench on this?

bwplotka:
>  Have we run prombench on this?

Nope, let's run on third PR to include whole feature maybe?

storage/series.go
bwplotka: XD talking to myself :man_facepalming: 

will remove.

storage/remote/read.go
bwplotka: ack will search and link

brancz:
That works for me.

storage/fanout.go
bwplotka: Wonder if primary should be slice for consistency... Maybe for now as we don't need that, let's stick to one.

brian-brazil: Prometheus only has one local tsdb.

bwplotka:
All should be addressed here, please also look on the next one: https://github.com/prometheus/prometheus/pull/7069

storage/remote/client.go
```diff
@@ -38,30 +39,54 @@ const maxErrMsgLen = 256
...
 type ClientConfig struct {
 	URL              *config_util.URL
 	Timeout          model.Duration
 	HTTPClientConfig config_util.HTTPClientConfig
 }
 
-// NewClient creates a new Client.
-func NewClient(remoteName string, conf *ClientConfig) (*Client, error) {
+// TODO(bwplotka): Add streamed chunked remote read method as well (https://github.com/prometheus/prometheus/issues/5926).
+type ReadClient interface {
```
brian-brazil, line 60: Public interfaces should have comments.

storage/remote/client.go
```diff
@@ -38,30 +39,54 @@ const maxErrMsgLen = 256
...
+	Read(ctx context.Context, query *prompb.Query) (*prompb.QueryResult, error)
+}
+
+// NewReadClient creates a new client for remote read.
+// Caller has to ensure there are no duplicated clients.
+func NewReadClient(readQueries prometheus.Gauge, name string, conf *ClientConfig) (ReadClient, error) {
```
brian-brazil, line 66: Should we pass in a registry instead?

storage/fanout.go
```diff
@@ -69,29 +69,47 @@ func (f *fanout) StartTime() (int64, error) {
...
 
-	// Add secondary queriers.
+	secondaries := make([]Querier, 0, len(f.secondaries))
 	for _, storage := range f.secondaries {
 		querier, err := storage.Querier(ctx, mint, maxt)
 		if err != nil {
-			for _, q := range queriers {
+			for _, q := range secondaries {
+				// TODO(bwplotka): Log error.
+				_ = q.Close()
```
brian-brazil, line 83: Don't you need to close the primary too?

storage/fanout.go
```diff
@@ -69,29 +69,47 @@ func (f *fanout) StartTime() (int64, error) {
...
+		return nil, err
+	}
+
+	secondaries := make([]ChunkQuerier, 0, len(f.secondaries))
+	for _, storage := range f.secondaries {
+		querier, err := storage.ChunkQuerier(ctx, mint, maxt)
+		if err != nil {
+			for _, q := range secondaries {
 				// TODO(bwplotka): Log error.
 				_ = q.Close()
```
brian-brazil, line 105: Ditto

storage/fanout.go
```diff
@@ -188,142 +206,146 @@ func (f *fanoutAppender) Rollback() (err error) {
...
 	}}
 }
 
 // Select returns a set of series that matches the given label matchers.
-func (q *mergeGenericQuerier) Select(sortSeries bool, hints *SelectHints, matchers ...*labels.Matcher) (genericSeriesSet, Warnings, error) {
-	if len(q.queriers) == 1 {
-		return q.queriers[0].Select(sortSeries, hints, matchers...)
-	}
-
+// NOTE: mergeGenericQuerier selects always return series sorted no matter if sorting was not required from caller.
```
brian-brazil, line 272: I can see people depending on this when they shouldn't as it's a function comment, I'd put this inside the function as an implementation note.

bwplotka: ack

storage/remote/client.go
bwplotka: lemme see

bwplotka:
React issue on CI

storage/remote/client.go
bwplotka: Ok, so no, in this case metric is what we want because we can create and delete clients for the whole life of the process and as you commented that we cannot use `Unregister` it's more explicit and nicer to register those on Storage level which is singleton.

storage/fanout.go
bwplotka: Nice catch!

storage/fanout.go
```diff
@@ -323,6 +324,8 @@ func (q *mergeGenericQuerier) Select(_ bool, hints *SelectHints, matchers ...*la
 	if priErr != nil {
 		return nil, warnings, priErr
 	}
+
+	// NOTE: newGenericMergeSeriesSet always return series sorted.
```
brian-brazil, line 328: No need for NOTE, it's not that important a detail.

storage/remote/client.go
```diff
@@ -56,6 +56,7 @@ type ClientConfig struct {
 	HTTPClientConfig config_util.HTTPClientConfig
 }
 
+// ReadClient allows to use SAMPLES method read against remote server.
```
brian-brazil, line 59: uses the SAMPLES

It doesn't permit it, it does it.

storage/remote/read_test.go
```diff
@@ -350,7 +350,7 @@ func TestSampleAndChunkQueryableClient(t *testing.T) {
 			mint: 1, maxt: 2,
 			matchers: []*labels.Matcher{
 				labels.MustNewMatcher(labels.MatchNotEqual, "a", "something"),
-				// TODO(bwplotka): Is this what we want? I prometheus is region=europe and user requests region=us we
+				// TODO(bwplotka): Is this what we want? If prometheus is in region=europe and user requests region=us we
```
brian-brazil, line 353: Yes, this is the desired behaviour. US&Europe isn't very sane as they're at the same level, but think about a hierarchical setup.

storage/remote/client.go
brian-brazil: It still feels weird to be passing metrics around. What happens if we add more metrics such as latency around?

storage/remote/read_test.go
bwplotka: ok

storage/remote/client.go
bwplotka: Why? We use that pattern in many many places in different projects to allow sharing the same `GaugeVec`. If we will have more than one we can create `readClientStruct` like we do in many places in Prometheus as well (:

storage/remote/client.go
bwplotka: ;po

bwplotka:
@brian-brazil addressed (also rebased - to hope React unit test was fixed)

storage/remote/client.go
brian-brazil: A given metric should only usually exist in a single function and file, if you're passing it around that indicates that the factoring probably isn't correct.

bwplotka: This is not usual then. (: It's part of bigger system. :thinking: Let's talk offline, I really feel strong opinion that this is better than global metric. Cannot really see alternative here.

@beorn7 any feedback on this?

brian-brazil: Yes, and its lack of usualness is my concern. Instantiators of a class shouldn't care about what metrics it defines.

bwplotka: That is fair statement. OK, I think I know the solution, let's just not export this. It is not used anywhere, just in `Storage` type. We can also have no constructor as it's constructed just once. I believe this will not violate `Instantiators of a class shouldn't care about what metrics it defines.` as it will internal type, tied to storage, in the same some `seriesSet` might have tied `seriesSetIterator` in it. WDYT?

bwplotka: Changed, so it is not exported anymore, thus tied to stoage.

brian-brazil: Still seems a bit weird, visibility doesn't really change things.

beorn7: I think you can pass around a registry and then either register a `Gauge` with the `remoteName` and `endpoint` labels as const labels (the classical approach) or use [WrapRegistererWith](https://pkg.go.dev/github.com/prometheus/client_golang@v1.5.0/prometheus?tab=doc#WrapRegistererWith) for the same effect.

bwplotka: Well that's the problem. None of those solutions works because we can create many clients of same labels for lifetime of the process (configure remote read for server example.com, then remove then configure again).

bwplotka: That would help indeed. But my question is.. why complicating things? We use the same pattern as I suggest in a dozen of places in Thanos and it works really well. What's wrong with it? (: 

Anyway, don't mind to change to AreadyRegisteredError just curious.

brian-brazil: If a you get a AreadyRegisteredError I'd consider that a programming error and if you get it on purpose, poor architecture as you're trying to re-use metrics rather than define the right one once.

bwplotka: Done. Also talked offline https://matrix.to/#/!WaUKIfoqfiyWQhenET:matrix.org/$15869391701329748dlPKk:matrix.org?via=matrix.org

Looks like it looks not bad,  happy with that. Thanks for discussion @brian-brazil and @beorn7. PTAL (:

bwplotka:
Done @brian-brazil (:

storage/remote/client.go
brian-brazil: As indicated above, I consider any use of AreadyRegisteredError to be incorrect.

bwplotka: I have to reuse one... that's the whole point.

bwplotka: Ok at this point ... we are blocked. So far @brian-brazil  you did not propose ANY working idea and all my 4 attempts (Unregister, passing gauge, passing gauge on a private struct, AlreadyExistsError different) you rejected. So... what we do know? Because I am 100% sure the version with global is wrong. 

What would you suggest then?

brian-brazil: They way I'd approach this is to have a statically defined metric, pass down just the label value down from the instantiator, and have the instance access the child it needed via that label value.

brian-brazil: Search for Publisher on https://www.robustperception.io/label-lookups-and-the-child for the basic pattern I'd use.

bwplotka: Hah that's interesting as the child pattern is exactly what I did with `newReadClient(readQueries.WithLabels(..., ...), name, ...)`, just the parent metric was not global/static (package level), but from parent struct. (:

bwplotka: Otherwise all of it is package level so not really good solution.

bwplotka: Anyway, looks like there is no agreement here, so, lets do this, let's choose whatever now from the 5 attempts we had and I will start separate issue with some options how we can fix many package level metric definitions (and If we want to fix it) in one go (:  WDYT? I can even revert to package level even though I think it's a bad solution.

beorn7: I find the (package-)global variable the most problematic of all suggested ways. It's not only a style question, it just hard-breaks if you import the package multiple times with different import paths (as it is common in Go anyway, but now also a core concept of Go modules).

I see drawbacks of both the AlreadyExistsError pattern and the unregister approach, but both are not too bad IMHO.

bwplotka: Thanks. I am opening a separate issue about it will summarize our findings.

Let's focus on iterating over chunks here :hugs:

bwplotka:
I think this is slowly ready to be merged, but happy to fix https://github.com/prometheus/prometheus/pull/7069 (fixing last nit) chained PR and run prombench on both. :hugs:

storage/remote/client.go
brian-brazil: > as the child pattern is exactly what I did with

No, that's quite different as you're passing the metric rather than the label value. Metric objects should not pass API boundaries.

> I see drawbacks of both the AlreadyExistsError pattern and the unregister approach

I think that any approach that relies on known anti-patterns is not a good approach, and we should not be encouraging such things by putting them inside Prometheus itself. Users looking to use these are usually either instrumenting at the wrong level, or should be writing a custom collector instead.

> I can even revert to package level even though I think it's a bad solution.

It's what the code already did, and it's a fine pattern that won't cause us any problems.

beorn7: > It's what the code already did, and it's a fine pattern that won't cause us any problems.

It's an anti-pattern, and it's fundamentally broken in any non-trivial code.

storage/fanout.go
```diff
@@ -188,142 +208,146 @@ func (f *fanoutAppender) Rollback() (err error) {
...
+			resultChan <- &queryResult{q: q, set: set, wrn: wrn, err: err}
+			wg.Done()
+		}(q)
+	}
+	go func() {
+		wg.Wait()
+		close(resultChan)
+	}()
+
+	for r := range resultChan {
```
codesome, line 305: Few questions here:
1. Don't we need this to be in time-order as we use this with chained merge too?
2. (unrelated to this line) Do we attempt vertical merge for the queries which do not have overlapping data too?

storage/fanout.go
```diff
@@ -569,11 +593,14 @@ func (h *genericSeriesSetHeap) Pop() interface{} {
 	return x
 }
 
-// ChainedSeriesMerge returns single series from many same series by chaining samples together.
+// ChainingSeriesMerge returns single series from many same series by chaining samples together.
 // In case of the timestamp overlap, the first overlapped sample is kept and the rest samples with the same timestamps
```
codesome, line 597: > the first overlapped sample is kept

I just noticed this. In that case, we need to be consistent (maybe a future PR for someone). When vertical queries and compaction was introduced, the design was to keep the last appearing sample.

storage/fanout.go
```diff
@@ -569,11 +593,14 @@ func (h *genericSeriesSetHeap) Pop() interface{} {
...
 // are dropped. We expect the same labels for each given series.
-// TODO(bwplotka): This has the same logic as tsdb.verticalChainedSeries. Remove this in favor of ChainedSeriesMerge in next PRs.
-func ChainedSeriesMerge(s ...Series) Series {
+//
+// This works the best with replicated series, where data from two series are exactly the same. This does not work well
+// with "almost" the same data, e.g. from 2 Prometheus HA replicas. This is fine, since from the Prometheus perspective
+// this never happens.
+func ChainingSeriesMerge(s ...Series) Series {
```
codesome, line 603: While the name was changed back in #7005, I see that this also merges the overlapping data. Chaining sounds more like non-overlapping and attaching results directly.

codesome:
(My chrome was choking with the large diff open :) )

storage/fanout.go
bwplotka: Will check thanks!

storage/fanout.go
```diff
@@ -188,142 +208,146 @@ func (f *fanoutAppender) Rollback() (err error) {
...
 			filtered = append(filtered, newGenericQuerierFrom(querier))
+			continue
 		}
 	}
 
 	if len(filtered) == 0 {
-		return primaryQuerier
-	}
-
-	if primaryQuerier == nil && len(filtered) == 1 {
```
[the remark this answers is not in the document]
bwplotka: See: https://github.com/prometheus/prometheus/pull/7059#discussion_r400914446

storage/fanout.go
```diff
@@ -69,29 +69,49 @@ func (f *fanout) StartTime() (int64, error) {
...
 
-	// Add secondary queriers.
+	secondaries := make([]Querier, 0, len(f.secondaries))
 	for _, storage := range f.secondaries {
 		querier, err := storage.Querier(ctx, mint, maxt)
 		if err != nil {
-			for _, q := range queriers {
+			_ = primary.Close()
+			for _, q := range secondaries {
+				// TODO(bwplotka): Log error.
```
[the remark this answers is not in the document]
bwplotka: Let's decide on next PR, no regression here.

storage/fanout.go
bwplotka: 1. We don't.  `ChainingSeriesMerge` handles this just fine. But good point to add note to `mergeFunc` that it has to handle underodered time wise series (: Thanks! 
2. No, see:

```
func (c *genericMergeSeriesSet) At() Labels {
	if len(c.currentSets) == 1 {
		return c.currentSets[0].At()
	}
	series := make([]Labels, 0, len(c.currentSets))
	for _, seriesSet := range c.currentSets {
		series = append(series, seriesSet.At())
	}
	return c.mergeFunc(series...)
}

```

storage/fanout.go
bwplotka: I mean... what is last and first? It's really random and always was... which one is first which last ;p  Maybe not from algo perspective but from the caller perspective - we randomly put results from different remote APIs (in random order)

bwplotka: Changed docs to have "random one"

storage/fanout.go
bwplotka: Cool, so what would you suggest? Because to me this algo is exactly doing that. It does not merge anything.. It's just appends ALL together not caring about any scrape interval at all....

bwplotka:
Addressed all @codesome :heart: 

Rebased and fixed all. Continuing tomorrow on finishing part 3 finally to make sure we can run prombench on all: https://github.com/prometheus/prometheus/pull/7069

brian-brazil:
:+1:

storage/fanout.go
codesome: If we sort the blocks based on their minTime, then the block with lower mint is first and higher is the last - that was what was used in compaction and vertical query. Because we use heap, I don't think that is easy to take care of it.

storage/fanout.go
codesome: https://github.com/prometheus/prometheus/blob/a2000f39a70983ec1f25fa0c6bc43e68a6384d1f/storage/fanout.go#L681-L685

This tells me that we are actually vertically merging it here by dropping the samples with same timestamp.

tsdb/db.go
```diff
@@ -39,6 +39,7 @@ import (
 	"github.com/prometheus/prometheus/tsdb/chunkenc"
 	tsdb_errors "github.com/prometheus/prometheus/tsdb/errors"
 	"github.com/prometheus/prometheus/tsdb/fileutil"
+
```
codesome, line 42:
```suggestion
```

bwplotka: `goimports does this

bwplotka:
@codesome I renamed `ChainingSeriesMerge` to `OverlappedSeriesMerge` I think you point was good, we should make it clear

storage/fanout.go
```diff
@@ -188,142 +208,145 @@ func (f *fanoutAppender) Rollback() (err error) {
...
+	}
+	resultChan := make(chan *queryResult)
+
+	for _, q := range queries {
+		wg.Add(1)
+		go func(q genericQuerier) {
+			// We need to sort for newGenericMergeSeriesSet to work.
+			set, wrn, err := q.Select(true, hints, matchers...)
+			resultChan <- &queryResult{q: q, set: set, wrn: wrn, err: err}
+			wg.Done()
```
kakkoyun, line 296: In case the `q.Select` implementation panics, maybe it should be a deferred call. `defer wg.Done()`

storage/fanout.go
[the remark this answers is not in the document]
bwplotka: This is actually good point, I am adding about improvements to https://github.com/prometheus/prometheus/pull/7251 let's get this quick path back.

storage/fanout.go
bwplotka: if it panics then all panics anyway, but defer sounds good

bwplotka:
:wave: Rebased and updated all - now will do part 2 in next PR so we can run prombench :hugs: 

Also fixed small nits introduced by https://github.com/prometheus/prometheus/pull/7334

storage/remote/client.go
```diff
@@ -42,37 +42,86 @@ const maxErrMsgLen = 256
...
+			Buckets:   append(prometheus.DefBuckets, 25, 60),
+		},
+		[]string{remoteName, endpoint},
+	)
 )
 
-// Client allows reading and writing from/to a remote HTTP endpoint.
-type Client struct {
+func init() {
+	prometheus.MustRegister(remoteReadQueriesTotal, remoteReadQueries, remoteReadQueryDuration)
```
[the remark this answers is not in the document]
bwplotka: I just move the code don't laugh! ;p

storage/remote/client.go
```diff
@@ -42,37 +42,86 @@ const maxErrMsgLen = 256
...
+			Name:      "read_queries_total",
+			Help:      "The total number of remote read queries.",
+		},
+		[]string{remoteName, endpoint, "code"},
+	)
+	remoteReadQueries = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "remote_read_queries",
```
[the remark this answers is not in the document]
bwplotka: With what? I think I would try to not change things from outside, that would mean some compatibility issue potentially (:

web/api/v1/api.go
```diff
@@ -1215,78 +1216,142 @@ func (api *API) remoteRead(w http.ResponseWriter, r *http.Request) {
...
-					api.remoteReadMaxBytesInFrame,
-				)
-			})
-			for _, w := range ws {
-				level.Warn(api.logger).Log("msg", "warnings on remote read query", "err", w.Error())
+func (api *API) remoteReadSamples(ctx context.Context, w http.ResponseWriter, req *prompb.ReadRequest, externalLabels map[string]string, sortedExternalLabels []prompb.Label) {
+	w.Header().Set("Content-Type", "application/x-protobuf")
+	w.Header().Set("Content-Encoding", "snappy")
+
+	// On empty or unknown types in req.AcceptedResponseTypes we default to non streamed, raw samples response.
```
brian-brazil, line 1229: This comment would make more sense up where this function is called.

bwplotka: Will fix in next PR :hugs:

bwplotka: This one: https://github.com/prometheus/prometheus/pull/7069

PR #7065, opened 2020-03-28
Added feature to add URL Scheme to promtool URL

AdaephonBen:
This is related to #6397 and adds the URL Scheme "http" to the promtool argument URL string, which should close the issue.


cmd/promtool/main.go
```diff
@@ -191,6 +191,13 @@ func checkFileExists(fn string) error {
 	return err
 }
 
+func addScheme(urlString string) (string, error) {
+	// Add HTTP Scheme to urlString
+	urlObject, err := url.Parse(urlString)
+	urlObject.Scheme = "http"
```
brian-brazil: This would lose a https scheme. It's generally easiest to check if the string already has http/https

AdaephonBen: Thanks for the feedback. I'll modify the function to add the scheme only if one doesn't already exist.

AdaephonBen: This has been done in 5300e2a

cmd/promtool/main.go
```diff
@@ -191,6 +191,16 @@ func checkFileExists(fn string) error {
 	return err
 }
 
+func addScheme(urlString string) (string, error) {
```
roidelapluie: Can we have tests for this function?

AdaephonBen: Sure. While adding one, I was looking at [main_test.go](https://raw.githubusercontent.com/prometheus/prometheus/master/cmd/promtool/main_test.go), found that it's setting up a mock server and testing query range with this server. The only problem is that for our use case, the mock server returns a URL that always has a schema(and hence is not a valid test). Should I manually edit out the schema from the URL argument of type server in [httptest](https://golang.org/pkg/net/http/httptest/)? If not, how should I write this test?

roidelapluie: just a simple test of addScheme helper function in its own; not starting a prometheus function

AdaephonBen: I have added a test in [bd40935](https://github.com/prometheus/prometheus/commit/bd40935a69d7ecad3e7139bb884318a5bb0885ba)

cmd/promtool/main.go
```diff
@@ -191,6 +191,16 @@ func checkFileExists(fn string) error {
...
+	// Add HTTP Scheme to urlString
+	urlObject, err := url.Parse(urlString)
```
roidelapluie, line 196: we should check for err

cmd/promtool/main_test.go
```diff
@@ -64,6 +64,18 @@ func TestQueryRange(t *testing.T) {
 	}
 }
 
+func TestAddScheme(t *testing.T) {
+	url := "google.com"
+	urlWithScheme, err := addScheme(url)
```
roidelapluie, line 69: we have helpers:

testutil.Ok(t,err)

roidelapluie:
Can you rebase this?

roidelapluie:
We have looked at it in our bug scrub and decided to close this due to inactivity. Feel free to reopen when you are available to work on it again.

AdaephonBen:
@roidelapluie Sorry for my lack of a response. I'd like to pick this issue up again, and finish it this time. I've resolved both issues you brought up. I'm assuming a rebase is necessary to merge this? I'm not sure I have the necessary permissions to reopen this PR so can you do it? Or should I open a new one?

roidelapluie:
I have re-open the PR

AdaephonBen:
@roidelapluie I've fixed the errors and merged the latest upstream `master` branch. However, although `make test` runs fine locally, `circle: test` fails because of a `common-unused` error in the Makefile which seems to refer to the NothingSelector in `vendor/k8s.io/apimachinery/pkg/labels/selector.go`. Since I did not make any changes to this file, I'm rather confused about why it's failing especially considering that `make test` reports no errors. Could you please advise me on how to resolve this issue?

PR #7066, opened 2020-03-29
Use .UTC everywhere we use time.Unix

brian-brazil:
time.Unix attaches the local timezone, which can then
leak out (e.g. in the alert json). While this is harmless,
we should be consistent.


hoffie:
Although this change seems consistent and matches the docs, it was somehow suprising for us, especially since we did not spot a hint in the Changelog.
We have an "alert firing since:" line in our notification mails and this PR seems to have changed that timestamp from local time to UTC.

Luckily, this can easily be worked around in alertmanager templates by using `.StartsAt.Local.Format ...` etc.

PR #7070, opened 2020-03-30
fix typo to more precise.

johncming:
change 
```go
discoverCancel: []context.CancelFunc{},
````

to 

```go
discoverCancels: []context.CancelFunc{},
````


roidelapluie:
We have looked at it in our bug scrub but it is not a typo.

Thanks.

PR #7075, opened 2020-03-31
[Docs] consul_sd_config: Add default value for `allow_stale`

fhemberger:
Ref: https://github.com/prometheus/prometheus/blob/master/discovery/consul/consul.go#L97


brian-brazil:
Thanks!

PR #7076, opened 2020-04-01
Query remote targets for label values

mtknapp:
Signed-off-by: Mark Knapp <knappm@optonline.net>

This is hopefully to address https://github.com/prometheus/prometheus/issues/3351 to allow for getting labelvalues from remote read targets.

Seems there's been several attempts at this before, but it seems they were abandoned/unfinished/not ideal. This is my first real-world attempt at using protobuf, largely copied/pasted from other examples in remote read, especially `func (c *Client) LabelValues`. Refactoring to cut down on repetition may be desired.

The question of what url endpoint to hit was raised in another PR, there they chose to define a new url in addition to the remote read url. To me this seemed redundant, and I feel it's implied that label values should be scraped from the same target as the remote reads, so I chose to hit /api/v1/read/**labelvalues**

Open to any suggestions, just kinda threw this together :)


mtknapp:
The failing tests seems to be happening on other PR's too and seems unrelated to my changes.

brian-brazil:
> so I chose to hit /api/v1/read/labelvalues

Are you sure about that?

mtknapp:
> > so I chose to hit /api/v1/read/labelvalues
> 
> Are you sure about that?

I'm not sure I follow what you're asking? Am I sure I'm hitting `/api/v1/read/labelvalues`? Or am I sure sure that's a good idea? 

If the former then I imagine you're hinting at more explicit wording. If a user is setup to remote read from another prometheus using the default pathing, meaning no trickery with reverse proxy's or anything, then yes it'll be hitting `/api/v1/read/labelvalues`. But more specifically it'll hit `client.url.String()+"/labelvalues"`, with the url being whatever the user defined in the config.

brian-brazil:
> Am I sure I'm hitting /api/v1/read/labelvalues?

Yes, because that's not what's in the PR (the PR does appear to be handling this correctly, your description is incorrect).

mtknapp:
I must be missing something here then, how is that not what's happening? I can see the requests metrics in prometheus

prometheus_http_requests_total{code="200",handler="/api/v1/read/labelvalues",instance="localhost:9090",job="prometheus"} | 7
-- | --

https://github.com/prometheus/prometheus/blob/fa1e6f745b46bf173d001c7d2ab105f76d79fc26/web/api/v1/api.go#L294-L296
https://github.com/prometheus/prometheus/blob/fa1e6f745b46bf173d001c7d2ab105f76d79fc26/storage/remote/client.go#L202-L204

brian-brazil:
That's a bit of a bug then, none of the remote read stuff should be under v1 as it's not a v1 API. v1 APIs are JSON.

mtknapp:
Ah ok, is that something I should look into changing with this? Or should that be a separate task?

brian-brazil:
We can't really change it now as that'd be breaking, but we should probably not make it worse.

mtknapp:
Right, that makes sense. Should I instead pursue using grpc like what's used for api/v2 admin functions?

mtknapp:
Took a crack at doing a grpc implementation, let me know if this seems like a better approach going forward with remote target operations. Presumably something similar would be done for `/api/v1/labels`

brian-brazil:
v2 is dead, and there's no plans to switch anything to grpc.

mtknapp:
Ok... could you advise on what should be done then?

brian-brazil:
We should probably get input from other developers on how to handle this.

mtknapp:
@gouthamve @tomwilkie @codesome @simonpasquier Any opinions on this? Looking for a little more feedback

bwplotka:
Ok, we had an offline discussion on IRC: https://matrix.to/#/!WaUKIfoqfiyWQhenET:matrix.org/$158641669586465FJBvM:matrix.org?via=matrix.org&via=half-shot.uk

Our question is, why we need to create another endpoint? Looks like such protobuf-based label values would expose exactly the same functionality as our JSON one. So why not define that remote target has to (actually it should **optionally** expose to not add braking change):
* api/v1/read
* api/v1/labels/{name}/values
* api/v1/labels/names

We can totally do that, document that and write client for those in Prometheus  `api/v1/labels/{name}/values`  `api/v1/labels/names` to support. Cannot see any blocker for that :hugs: 

Thoughts?

bwplotka:
Also let's focus on what would be the exact use cases here? (: I think there can be many would be nice to define them clearly.

mtknapp:
> Our question is, why we need to create another endpoint? Looks like such protobuf-based label values would expose exactly the same functionality as our JSON one. So why not define that remote target has to (actually it should **optionally** to not add braking change) expose:
> 
> * api/v1/read
> * api/v1/labels/{name}/values
> * api/v1/labels/names

Just to be clear, would the design be then to have a config with all three endpoints defined? Optionally excluding `label/{name}/values` and `/labels`? So something like:
```
remote_read:
- url: http:target:9090/api/v1/read
  labels_url: http:target:9090/api/v1/labels
  label_values_url: http:target:9090/api/v1/labels/...
```
With `labels_url` and `label_values_url` being optional? If so, it isn't really intuitive what the `label_values_url` should look like.

Alternatively, should the `/labels` and `/labels/{name}/values` endpoint be interpreted/manipulated from the given url and instead have optional flags to opt in to gathering label information from remote read targets?

The use case for me is essentially what brian pointed out in the that discussion of having sharded instances because we have a lot of targets, and remote reading from those shards. I know it's not a conventional setup but it seems to be working for us for now. We skirted around this issue in the meantime by reverse proxying those endpoints to a script that coalesces the values from the shards 😬Not the prettiest

brian-brazil:
I've never particularly been a fan of adding additional fields to the config in the various incarnations of this PR over time, I think manipulating to make it just work (barring the warnings you'll get for non-Prometheus backends that always fail it) with no additional configuration would be good. More intuitive configuration would be nice, but it's probably more something to consider for v3.

mtknapp:
Do you still want it to be an opt-in thing? If so, how would you propose allowing that option without making a configuration changes?

brian-brazil:
I'd design for the long term and not have it configurable, as any configuration field would ultimately be vestigial.

mtknapp:
@bwplotka Think you could take a look at this?

mtknapp:
Bump

brian-brazil:
@bwplotka Can you take a look at this?

roidelapluie:
It looks like this is lacking documentation as well.

storage/remote/client.go
```diff
@@ -187,3 +190,70 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
...
+
+	ctx, cancel := context.WithTimeout(ctx, c.timeout)
+	defer cancel()
+
+	httpResp, err := c.client.Do(httpReq.WithContext(ctx))
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "error sending request")
+	}
+	defer func() {
+		io.Copy(ioutil.Discard, httpResp.Body)
```
bwplotka, line 216: Pro tip, we can be explicit when we discard errors.

```suggestion
		_, _ io.Copy(ioutil.Discard, httpResp.Body)
```

storage/remote/client.go
```diff
@@ -187,3 +190,70 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
...
+		httpResp.Body.Close()
```
bwplotka, line 217:
```suggestion
		_ = httpResp.Body.Close()
```

storage/remote/client.go
```diff
@@ -187,3 +190,70 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
...
+	}()
+
+	body, err := ioutil.ReadAll(httpResp.Body)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, fmt.Sprintf("error reading response. HTTP status code: %s", httpResp.Status))
```
bwplotka, line 222: 1. Let's use `Wrapf` instead of nested Sprintf
2. No need to repeat mentioning that something failed. We are wrapping error, so no need for mentioning that something failed. What we want mention in wrap is what we were doing when err happened 🤗 

```suggestion
		return nil, nil, errors.Wrapf(err, "reading response. HTTP status code: %s", httpResp.Status)
```

storage/remote/client.go
```diff
@@ -187,3 +190,70 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
...
+
+	if httpResp.StatusCode/100 != 2 {
+		return nil, nil, errors.Errorf("remote server %s returned HTTP status %s: %s", url, httpResp.Status, strings.TrimSpace(string(body)))
+	}
+
+	var resp labelsResponse
+	err = json.Unmarshal(body, &resp)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "unable to unmarshal response body")
+	}
```
bwplotka, line 233:
```suggestion
	if err = json.Unmarshal(body, &resp); err != nil {
		return nil, nil, errors.Wrap(err, "unable to unmarshal response body")
	}
```

storage/remote/client.go
```diff
@@ -187,3 +190,70 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
...
 }
+
+type labelsResponse struct {
+	Data     []string `json:"data,omitempty"`
+	Error    string   `json:"error,omitempty"`
+	Warnings []string `json:"warnings,omitempty"`
+}
+
+// makeLabelsRequest makes the actual request for /label and /labels/{name}/values from the remote client
+func (c *Client) makeLabelsRequest(ctx context.Context, url string) ([]string, storage.Warnings, error) {
```
bwplotka, line 201: Wonder, can we use https://github.com/prometheus/client_golang/blob/master/api/prometheus/v1/api.go#L235 instead?

storage/remote/client.go
```diff
@@ -187,3 +190,70 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
...
+	for _, warning := range resp.Warnings {
+		warnings = append(warnings, errors.New(warning))
+	}
+
+	return resp.Data, warnings, nil
+}
+
+// LabelValues queries label values from a remote endpoint.
+func (c *Client) LabelValues(ctx context.Context, name string) ([]string, storage.Warnings, error) {
+	re := regexp.MustCompile(`read/?$`)
```
bwplotka, line 249: This looks like a code smell. Can we create client with correct URLs once to not rely on computing regexp dynamically every time? Also, can we use https://github.com/prometheus/client_golang/blob/master/api/prometheus/v1/api.go#L235

bwplotka (commented):
Looks generally good, but I wonder if we could use the client we maintain https://github.com/prometheus/client_golang/blob/master/api/prometheus/v1/api.go#L235

cc @beorn7 does it make sense?

Thanks!

beorn7:
@bwplotka about using https://github.com/prometheus/client_golang/blob/master/api/prometheus/v1/api.go#L235 : I think you know that API client much better than I do. (Remember, I “maintain” it mostly by accident.) Superficially, it totally makes sense to me to use it. But I don't have any deeper insights about pros and cons.

bwplotka:
Thanks, @beorn7, I would use it. For streamed operation with protobuf we need to do custom things, but here we could reuse solid API we already distribute. What do you think @mtknapp? (:

sgrzemski:
@bwplotka @beorn7 @mtknapp what is the status of this feature? Can I take over and finish this task? Having the labels and autocomplete in Prometheus set to remote read would be a feature nice to have.

roidelapluie:
Closing as duplicate of #11877

PR #7077, opened 2020-04-01
Fix react tests

csmarchbanks:
https://github.com/facebook/create-react-app/issues/8689 is causing our
tests to fail in the CI pipeline. As the comments suggest, downgrading
to react-scripts 3.4.0 fixes the problem.

In addition, fix a test warning due to a missing id field.

cc: @juliusv


juliusv:
Could you add a comment line pointing to this PR in package.json so we know why it's like that, and that we can try updating it again later but need to be careful?

csmarchbanks:
How should I add a comment to package.json? A brief internet search shows that I could add a comments section but that is pretty far removed from the dependency.

juliusv:
@csmarchbanks I forgot for a moment that plain old JSON of course doesn't allow for comments. Ok, then we'll just have to live with it like that, yeah. Thanks!

Issue #7080, opened 2020-04-02
prometheus_web_federation metrics not available on prometheus metric endpoint


juliantaylor:
The prometheus web_federation error metrics are not available on the metrics endpoint where the other prometheus metrics are available
The metrics are in the source code though:
https://github.com/prometheus/prometheus/blob/master/web/federate.go#L37

Maybe some registering is missing?

**What did you do?**
on a prometheus with federations timing out
`curl -s prometheus:9090/metrics | grep federation_fai`

The log of the prometheus prints the failures, the log is printed with the metric export in the source so the metric should be there:
```
level=error ts=2020-04-02T11:31:17.906Z caller=federate.go:184 component=web msg="federation failed" err="write tcp 100.70.202.8:9090->100.68.240.115:45096: write: broken pipe"
```

**What did you expect to see?**
I expected the metrics to appear:
prometheus_web_federation_errors_total
prometheus_web_federation_warnings_total

**What did you see instead? Under which circumstances?**
nothing

* Prometheus version:

	2.16.0

brian-brazil:
Looks like they were never registered, would you like to send a PR?

juliantaylor:
I'll give it a try

PR #7081, opened 2020-04-02
register federation failure metrics

juliantaylor:
Closes gh-7080

### DIFF
--- a/web/federate.go
+++ b/web/federate.go
@@ -44,6 +44,10 @@ var (
 	})
 )
 
+func registerFederationMetrics(r prometheus.Registerer) {
+	r.MustRegister(federationWarnings, federationErrors)
+}
+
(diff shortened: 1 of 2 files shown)


brian-brazil:
The tests are failing.

web/federate.go
```diff
@@ -44,6 +44,10 @@ var (
 	})
 )
 
+func registerFederationMetrics(r prometheus.Registerer) {
```
roidelapluie, line 47: Why do we need a function?

juliantaylor: so the metric registration is in the same file as the metrics themselves so it is less likely a register is forgotten when adding metrics

do you prefer it differently?

juliantaylor:
test failures are fixed

brian-brazil:
Thanks!

Issue #6950, opened 2020-03-09
Ingestion: Use the same appender for regular samples and reporting (`up` and friends)


beorn7:
This is a follow-up issue to #1893.

The regular samples are appended to TSDB with an appender that is committed or rolled-back, depending on success or failure. After that, a new appender is created to append the reporting samples (`up`, `scrape_...`, etc.). The problem with this approach is that the reporting samples will become visible to the query engine a bit after the regular samples are visible.

To fix this, only one appender has to be used for both samples in the case of success. If appending the regular samples already fails for some reason, that appender has to be rolled back and a new appender has to be used for the reporting samples (which still need to be appended because `up` is now 0). If all looks successful but then the final commit for the one appender errors out, things are most likely so broken that we don't need to retry appending with just the reporting samples.

roidelapluie:
Note to anyone who wants to implement this: the code will overlap with pending pull request #6918

PR #7085, opened 2020-04-03
Ingestion: Use the same appender for regular samples and reporting (`up` and friends) #6950

liguozhong:
Ingestion: Use the same appender for regular samples and reporting (`up` and friends) #6950

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -973,7 +973,7 @@ mainLoop:
 			level.Debug(sl.l).Log("msg", "append failed", "err", appErr)
 			// The append failed, probably due to a parse error or sample limit.
 			// Call sl.append again with an empty scrape to trigger stale markers.
-			if _, _, _, err := sl.append(scrapeAppender, []byte{}, "", start); err != nil {
+			if _, _, _, err := sl.append(sl.appender(), []byte{}, "", start); err != nil {
 				level.Warn(sl.l).Log("msg", "append failed", "err", err)
 			}
 		}
@@ -983,7 +983,9 @@ mainLoop:
 		if scrapeErr == nil {
 			scrapeErr = appErr
 		}
-
+		if scrapeErr != nil {
+			scrapeAppender = nil
+		}
 		if err := sl.report(scrapeAppender, start, time.Since(start), total, added, seriesAdded, scrapeErr); err != nil {
 			level.Warn(sl.l).Log("msg", "appending scrape report failed", "err", err)
 		}
@@ -1291,7 +1293,8 @@ func (sl *scrapeLoop) report(app storage.Appender, start time.Time, duration tim
 	var health float64
 	if scrapeErr == nil {
 		health = 1
-	} else {
+	}
+	if app == nil {
 		app = sl.appender()
 	}
 	defer func() {
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -967,12 +967,13 @@ mainLoop:
 
 		// A failed scrape is the same as an empty scrape,
 		// we still call sl.append to trigger stale markers.
-		total, added, seriesAdded, appErr := sl.append(b, contentType, start)
+		scrapeAppender := sl.appender()
+		total, added, seriesAdded, appErr := sl.append(scrapeAppender, b, contentType, start)
 		if appErr != nil {
 			level.Debug(sl.l).Log("msg", "append failed", "err", appErr)
 			// The append failed, probably due to a parse error or sample limit.
 			// Call sl.append again with an empty scrape to trigger stale markers.
-			if _, _, _, err := sl.append([]byte{}, "", start); err != nil {
+			if _, _, _, err := sl.append(scrapeAppender, []byte{}, "", start); err != nil {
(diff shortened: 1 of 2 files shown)


scrape/scrape.go
```diff
@@ -1275,16 +1283,17 @@ const (
...
+func (sl *scrapeLoop) report(app storage.Appender, start time.Time, duration time.Duration, scraped, appended, seriesAdded int, scrapeErr error) (err error) {
 	sl.scraper.Report(start, duration, scrapeErr)
 
 	ts := timestamp.FromTime(start)
 
 	var health float64
 	if scrapeErr == nil {
 		health = 1
+	} else {
+		app = sl.appender()
```
codesome, line 1295: Why is it in this `else`? Shouldn't we check for `app == nil`?

liguozhong: When the scrape fails, the app is still not nil.
As mentioned in issue, use a new appender when the scraper fails.

scrape/scrape.go
```diff
@@ -967,12 +967,13 @@ mainLoop:
...
 		// we still call sl.append to trigger stale markers.
-		total, added, seriesAdded, appErr := sl.append(b, contentType, start)
+		scrapeAppender := sl.appender()
+		total, added, seriesAdded, appErr := sl.append(scrapeAppender, b, contentType, start)
 		if appErr != nil {
 			level.Debug(sl.l).Log("msg", "append failed", "err", appErr)
 			// The append failed, probably due to a parse error or sample limit.
 			// Call sl.append again with an empty scrape to trigger stale markers.
-			if _, _, _, err := sl.append([]byte{}, "", start); err != nil {
+			if _, _, _, err := sl.append(scrapeAppender, []byte{}, "", start); err != nil {
```
brian-brazil, line 976: Don't we need a new appender here?

scrape/scrape.go
brian-brazil: Spreading the logic across functions is fragile, I think this should all be in `run`

scrape/scrape.go
liguozhong: thanks .done

scrape/scrape.go
liguozhong: > Spreading the logic across functions is fragile, I think this should all be in `run`
thanks .done.

image
image

scrape/scrape.go
```diff
@@ -967,12 +967,13 @@ mainLoop:
...
 		// we still call sl.append to trigger stale markers.
-		total, added, seriesAdded, appErr := sl.append(b, contentType, start)
+		scrapeAppender := sl.appender()
+		total, added, seriesAdded, appErr := sl.append(scrapeAppender, b, contentType, start)
 		if appErr != nil {
 			level.Debug(sl.l).Log("msg", "append failed", "err", appErr)
 			// The append failed, probably due to a parse error or sample limit.
 			// Call sl.append again with an empty scrape to trigger stale markers.
-			if _, _, _, err := sl.append([]byte{}, "", start); err != nil {
+			if _, _, _, err := sl.append(sl.appender(), []byte{}, "", start); err != nil {
```
roidelapluie, line 976: why not reusing  scrapeAppender ?

scrape/scrape.go
```diff
@@ -1045,7 +1048,7 @@ func (sl *scrapeLoop) endOfRunStaleness(last time.Time, ticker *time.Ticker, int
 	// Call sl.append again with an empty scrape to trigger stale markers.
 	// If the target has since been recreated and scraped, the
 	// stale markers will be out of order and ignored.
-	if _, _, _, err := sl.append([]byte{}, "", staleTime); err != nil {
+	if _, _, _, err := sl.append(nil, []byte{}, "", staleTime); err != nil {
 		level.Error(sl.l).Log("msg", "stale append failed", "err", err)
 	}
 	if err := sl.reportStale(staleTime); err != nil {
```
roidelapluie, line 1054: this is also using an appender

roidelapluie:
I think this PR is not ready t be merged. By reading the code, it looks like we could Commit or Rollback twice.

scrape/scrape.go
liguozhong: thanks for your review ,done

scrape/scrape.go
liguozhong: thanks for your review ,done

scrape/scrape.go
brian-brazil: We still need a new appender here, and the old one rolled back.

scrape/scrape.go
```diff
@@ -1284,7 +1294,9 @@ func (sl *scrapeLoop) report(start time.Time, duration time.Duration, scraped, a
 	if scrapeErr == nil {
 		health = 1
 	}
-	app := sl.appender()
+	if app == nil {
```
brian-brazil, line 1297: It should always be passed in.

liguozhong: thanks.done

roidelapluie:
Hello,

This PR has gone stale for 4 months and I actually forgot about it, so I ended up implementing it in a different way in #7561.

Sorry about that, and thanks for the work you've done.

PR #7086, opened 2020-04-04
discovery: ec2: expose IPv6 as label

### LINKED ISSUE #4611: Support for IPv6 in EC2 service discovery
## Proposal
EC2 instances might not be reachable via private IPv4 addresses (if scraping is done remotely), and might not have public IPv4 addresses.

It would be a very nice addition to expand service discovery to include IPv6 addresses, if possible.

Thank you for your consideration.

mfrw:
Signed-off-by: Muhammad Falak R Wani <falakreyaz@gmail.com>

Closes #4611

### DIFF
--- a/discovery/ec2/ec2.go
+++ b/discovery/ec2/ec2.go
@@ -40,21 +40,22 @@ const (
 	ec2Label                  = model.MetaLabelPrefix + "ec2_"
 	ec2LabelAZ                = ec2Label + "availability_zone"
 	ec2LabelArch              = ec2Label + "architecture"
+	ec2LabelIPv6              = ec2Label + "ipv6"
 	ec2LabelInstanceID        = ec2Label + "instance_id"
+	ec2LabelInstanceLifecycle = ec2Label + "instance_lifecycle"
 	ec2LabelInstanceState     = ec2Label + "instance_state"
 	ec2LabelInstanceType      = ec2Label + "instance_type"
-	ec2LabelInstanceLifecycle = ec2Label + "instance_lifecycle"
 	ec2LabelOwnerID           = ec2Label + "owner_id"
 	ec2LabelPlatform          = ec2Label + "platform"
-	ec2LabelPublicDNS         = ec2Label + "public_dns_name"
-	ec2LabelPublicIP          = ec2Label + "public_ip"
+	ec2LabelPrimarySubnetID   = ec2Label + "primary_subnet_id"
 	ec2LabelPrivateDNS        = ec2Label + "private_dns_name"
 	ec2LabelPrivateIP         = ec2Label + "private_ip"
-	ec2LabelPrimarySubnetID   = ec2Label + "primary_subnet_id"
+	ec2LabelPublicDNS         = ec2Label + "public_dns_name"
+	ec2LabelPublicIP          = ec2Label + "public_ip"
 	ec2LabelSubnetID          = ec2Label + "subnet_id"
 	ec2LabelTag               = ec2Label + "tag_"
 	ec2LabelVPCID             = ec2Label + "vpc_id"
-	subnetSeparator           = ","
+	ec2LabelSeparator         = ","
 )
 
 // DefaultSDConfig is the default EC2 SD configuration.
@@ -228,10 +229,21 @@ func (d *Discovery) refresh(ctx context.Context) ([]*targetgroup.Group, error) {
 					labels[ec2LabelVPCID] = model.LabelValue(*inst.VpcId)
 					labels[ec2LabelPrimarySubnetID] = model.LabelValue(*inst.SubnetId)
 
-					// Deduplicate VPC Subnet IDs maintaining the order of the network interfaces returned by EC2.
+					// Deduplicate IPv6s & VPC Subnet IDs maintaining the order of the network interfaces returned by EC2.
 					var subnets []string
+					var ipv6addrs []string
 					subnetsMap := make(map[string]struct{})
+					ipv6Map := make(map[string]struct{})
 					for _, eni := range inst.NetworkInterfaces {
+						if eni.Ipv6Addresses != nil {
+							for _, ipv6addr := range eni.Ipv6Addresses {
+								i := ipv6addr.String()
+								if _, ok := ipv6Map[i]; !ok {
+									ipv6Map[i] = struct{}{}
+									ipv6addrs = append(ipv6addrs, i)
+								}
+							}
+						}
 						if eni.SubnetId == nil {
 							continue
 						}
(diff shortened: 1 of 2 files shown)


discovery/ec2/ec2.go
```diff
@@ -40,21 +40,22 @@ const (
 	ec2Label                  = model.MetaLabelPrefix + "ec2_"
 	ec2LabelAZ                = ec2Label + "availability_zone"
 	ec2LabelArch              = ec2Label + "architecture"
+	ec2LabelIPv6              = ec2Label + "ipv6"
```
brian-brazil, line 43: This isn't very clear, ipv6 what?

mfrw: sure, will update it to ipv6_address

brian-brazil: We don't provide a list of addresses anywhere, and that's not very useful for scraping as we need a single address. How are you planning on using this? Are there public and private v6 addresses?

mfrw: EC2 instances might not be reachable via private IPv4 addresses (if scraping is done remotely), and might not have public IPv4 addresses.
In my opinion there are no private/public v6 addrs.
Also, i don't have a usecase for this, i wanted to contribute.. and had a look at: #4611

brian-brazil: We probably need to know a bit more about the use case to figure out how to best represent this, as the idea appears to be using it as the thing to scrape rather than just for niche filtering use cases.

mfrw: sure 👍

gebn:
I can provide some more background on my use case - I found this issue looking for a way to scrape over IPv6. There is indeed no concept of public and private v6 addresses in AWS. However, a given [ENI](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html) may have multiple addresses, and an instance may have multiple ENIs attached, so there can still be multiple IPv6 address, and the first one returned by the API may not be the correct one to use. The order is stable, so in practice the relabelling task is to "replace `__address__` with the nth IPv6 address of the instance". This would break if ENIs were modified (e.g. what was the 3rd IP in the list moves to 1st), however that's a fairly major change.

The current implementation in this PR has deduplication logic; I don't *think* it is possible for two ENIs with the same IP to be attached to a single instance simultaneously, so this shouldn't be required.

brian-brazil:
Is the ordering predictable? E.g. that the 2nd IPv6 always means the same thing across instances? Being stable per instance isn't sufficient.

gebn:
I asked AWS support about this. They've concluded the ordering of ENIs and IPs in the API output follows the attachment/assignment order, which may not follow the ordering of interfaces within the OS (it doesn't for one of my instances). Perhaps sorting the IPs would provide the necessary stability across ENIs and instances? Depending on attachment order feels unintuitive.

brian-brazil:
Sorting won't be predictable, as a given IPs will be in different places depending on what other networks an instance is part of.

What are you trying to do in your use case? My presumption would be that the logic would be something use an IP from this subnet, and if it's not that one use an IP from this other subnet.

gebn:
You're correct - for me, the instances in question terminate a VPN, with one interface in a public subnet, and one in a private subnet. I'd want to scrape from the private subnet using the latter. Sorting would work here, as the subnets are non-overlapping, so I could always take either the first or second IP. I agree it's not perfect, but at the point where if/else logic is truly needed, I think it would be entirely reasonable, and more maintainable, to make folks query AWS themselves and generate the `static_configs`. The overwhelming majority of v6 instances will have a single interface and single address.

brian-brazil:
Given they've v6 addresses, would providing them as an unordered list and then regexing to get the IP with the right subnet prefix work?

gebn:
Yep, that would work! All VPCs in AWS are a /56, and all subnets are a /64.

gebn:
I've implemented this [here](https://github.com/prometheus/prometheus/compare/master...gebn:sd-ec2-ipv6). If there are no addresses, the label is not added, otherwise:

Single address:

```
__meta_ec2_ipv6_addresses=",2a05:d01c:55:d320::5,"
```

Multiple addresses:

```
__meta_ec2_ipv6_addresses=",2a05:d01c:55:d32f::4,2a05:d01c:55:d320::4,"
```

It's broadly the same as this PR, but without the deduplication logic, and I've renamed `subnetSeparator` to `valueSeparator` rather than `ec2LabelSeparator` as the `ec2Label` prefix looks to be intended for only label names. There is one correctness vs. consistency point: one one hand, I think the plural *addresses* is clearer, however the existing `_subnet_id` label is actually a list. Pending resolution, happy for @mfrw to update this PR, or I can create a new one after a grace period.

mfrw:
@gebn let me do the changes that you suggested .. and if we all are happy lets move forward :)

discovery/ec2/ec2.go
```diff
@@ -228,22 +229,33 @@ func (d *Discovery) refresh(ctx context.Context) ([]*targetgroup.Group, error) {
...
 
-					// Deduplicate VPC Subnet IDs maintaining the order of the network interfaces returned by EC2.
 					var subnets []string
+					var ipv6addrs []string
 					subnetsMap := make(map[string]struct{})
 					for _, eni := range inst.NetworkInterfaces {
 						if eni.SubnetId == nil {
 							continue
 						}
+						// Deduplicate IPv6s & VPC Subnet IDs maintaining the order of the subnets returned by EC2.
```
brian-brazil, line 239: You're not deduplicating addresses

docs/configuration/configuration.md
```diff
@@ -442,6 +442,7 @@ The following meta labels are available on targets during [relabeling](#relabel_
 * `__meta_ec2_subnet_id`: comma separated list of subnets IDs in which the instance is running, if available
 * `__meta_ec2_tag_<tagkey>`: each tag value of the instance
 * `__meta_ec2_vpc_id`: the ID of the VPC in which the instance is running, if available
+* `__meta_ec2_ipv6_address`: comma seperated list of IPv6 addresses for the all the network interfaces of the instance, if available
```
brian-brazil, line 445: Keep it sorted

discovery/ec2/ec2.go
mfrw: Yes :( silly copy paste error

docs/configuration/configuration.md
mfrw: Sure

brian-brazil:
Thanks!

gebn:
@brian-brazil, can you correct `for the all the` in `configuration.md`?

brian-brazil:
Would you like to send a PR?

gebn:
Sure, wondered if it was overkill for such a minor change! Give me 10 mins.

PR #7089, opened 2020-04-04
promql: Add support for .Select calls concurrency. #6878

liguozhong:
promql: Add support for .Select calls concurrency. #6878

### DIFF
--- a/pkg/gate/gate_test.go
+++ b/pkg/gate/gate_test.go
@@ -38,7 +38,7 @@ func TestLabels_IsOverload(t *testing.T) {
 				return
 			}
 			if err := gate.Start(context.Background()); err != nil {
-
+				t.Fatal(err)
 			}
 			time.Sleep(2 * time.Second)
 			gate.Done()
--- a/pkg/gate/gate_test.go
+++ b/pkg/gate/gate_test.go
@@ -30,8 +30,8 @@ func TestLabels_IsOverload(t *testing.T) {
 	var wg sync.WaitGroup
 
 	for i := 0; i < callCount; i++ {
+		wg.Add(1)
 		go func() {
-			wg.Add(1)
 			defer wg.Done()
 			if gate.IsOverload() {
 				overloadChan <- 1
--- a/pkg/gate/gate_test.go
+++ b/pkg/gate/gate_test.go
@@ -1,4 +1,4 @@
-// Copyright 2016 The Prometheus Authors
+// Copyright 2020 The Prometheus Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
--- a/pkg/gate/gate_test.go
+++ b/pkg/gate/gate_test.go
@@ -29,7 +29,7 @@ func TestLabels_IsOverload(t *testing.T) {
 	overloadChan := make(chan int, callCount-maxGoRoutine)
 	var wg sync.WaitGroup
 
-	var mainErr error = nil
+	var mainErr error
 	for i := 0; i < callCount; i++ {
 		wg.Add(1)
 		go func() {
--- a/pkg/gate/gate_test.go
+++ b/pkg/gate/gate_test.go
@@ -29,6 +29,7 @@ func TestLabels_IsOverload(t *testing.T) {
 	overloadChan := make(chan int, callCount-maxGoRoutine)
 	var wg sync.WaitGroup
 
+	var mainErr error = nil
 	for i := 0; i < callCount; i++ {
 		wg.Add(1)
 		go func() {
@@ -38,11 +39,14 @@ func TestLabels_IsOverload(t *testing.T) {
 				return
 			}
 			if err := gate.Start(context.Background()); err != nil {
-				t.Fatal(err)
+				mainErr = err
 			}
 			time.Sleep(2 * time.Second)
 			gate.Done()
 		}()
+	}
+	if mainErr != nil {
+		t.Fatal(mainErr)
 	}
 
 	wg.Wait()
--- a/pkg/gate/gate_test.go
+++ b/pkg/gate/gate_test.go
@@ -1,0 +1,51 @@
+// Copyright 2016 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gate
+
+import (
+	"context"
+	"github.com/prometheus/prometheus/util/testutil"
+	"sync"
+	"testing"
+	"time"
+)
+
+func TestLabels_IsOverload(t *testing.T) {
+
+	maxGoRoutine := 3
+	gate := New(maxGoRoutine)
+	callCount := 10
+	overloadChan := make(chan int, callCount-maxGoRoutine)
+	var wg sync.WaitGroup
+
+	for i := 0; i < callCount; i++ {
+		go func() {
+			wg.Add(1)
+			defer wg.Done()
+			if gate.IsOverload() {
+				overloadChan <- 1
+				return
+			}
+			if err := gate.Start(context.Background()); err != nil {
+
+			}
+			time.Sleep(2 * time.Second)
+			gate.Done()
+		}()
+	}
+
+	wg.Wait()
+	testutil.Equals(t, len(overloadChan), callCount-maxGoRoutine)
+
+}
--- a/pkg/gate/gate_test.go
+++ b/pkg/gate/gate_test.go
@@ -21,7 +21,7 @@ import (
 	"time"
 )
 
-func TestLabels_IsOverload(t *testing.T) {
+func TestIsOverload(t *testing.T) {
 
 	maxGoRoutine := 3
 	gate := New(maxGoRoutine)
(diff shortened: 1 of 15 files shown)


brian-brazil:
This is still going to create goroutines when there's only the local tsdb to talk to. We want to talk remotely concurrently, but not locally.

roidelapluie:
> This is still going to create goroutines when there's only the local tsdb to talk to. We want to talk remotely concurrently, but not locally.

Like #6770?

liguozhong:
> This is still going to create goroutines when there's only the local tsdb to talk to. We want to talk remotely concurrently, but not locally.

done.

image

------- [9090 dev debug info]--------
image
image

------- [9010 dev debug info]--------
image
image

promql/engine.go
```diff
@@ -651,57 +654,132 @@ func (ng *Engine) populateSeries(ctx context.Context, querier storage.Querier, s
...
+		}
+		parser.Inspect(s.Expr, func(node parser.Node, path []parser.Node) error {
+			switch node.(type) {
+			case *parser.VectorSelector:
+				RemoteSelectCalls++
+			}
+			return nil
+		})
+	}
+	if RemoteSelectCalls < 2 {
```
slrtbtfs, line 670: Would it significantly affect performance to move that `if` inside the `switch` and  thus avoid code duplication?

liguozhong: [outside]
image
[inside]
image

promql/engine.go
```diff
@@ -651,57 +653,133 @@ func (ng *Engine) populateSeries(ctx context.Context, querier storage.Querier, s
 		warnings  storage.Warnings
 		err       error
 	)
+	RemoteSelectCalls := 0
+	if querier.Remotely() {
```
brian-brazil, line 657: PromQL shouldn't care about how the querier is implemented, it should always be the same logic.

liguozhong: thanks for your review.
I move "remotely" to the engine. Is that ok? Only three files need to be changed.
image
image
image

brian-brazil: That sort of check shouldn't appear anywhere in the promql directory, any special handling of local vs remote should be in fanout.

liguozhong: Thanks, I will think about how to implement this feature in "fanout.go"

krasi-georgiev:
> This is still going to create goroutines when there's only the local tsdb to talk to. We want to talk remotely concurrently, but not locally.

Why not locally as well?

brian-brazil:
> Why not locally as well?

It'd make queries more expensive overall, and would also mean that there was no longer any limit to how many cores PromQL could use - neither of which is desirable.

promql/engine.go
liguozhong: I have moved the remote related logic to the fanout directory.
done . please review again ,thanks.
image
debug info 9010 is remotely result, 9090 is locally result
image

promql/engine.go
```diff
@@ -689,19 +687,40 @@ func (ng *Engine) populateSeries(ctx context.Context, querier storage.Querier, s
...
+			sp := &storage.SelectParam{SortSeries: false, Hints: hints, Matchers: n.LabelMatchers}
+			selectParams = append(selectParams, sp)
+			selectCalls[sp] = n
 		case *parser.MatrixSelector:
 			evalRange = n.Range
 		}
 		return nil
 	})
+	switch q := querier.(type) {
+	case storage.BatchQuerier:
```
brian-brazil, line 699: You're still special casing here, the engine should always be doing the same thing no matter what storage it is being passed.

krasi-georgiev:
> It'd make queries more expensive overall, and would also mean that there was no longer any limit to how many cores PromQL could use - neither of which is desirable.

yes this it the tradeoff, but some use cases might have enough resources and would want that faster processing.

> and would also mean that there was no longer any limit to how many cores PromQL could use - neither of which is desirable.

we could limit the goroutines to the number of available cores, no?

roidelapluie:
> > It'd make queries more expensive overall, and would also mean that there was no longer any limit to how many cores PromQL could use - neither of which is desirable.
> 
> yes this it the tradeoff, but some use cases might have enough resources and would want that faster processing.
> 
> > and would also mean that there was no longer any limit to how many cores PromQL could use - neither of which is desirable.
> 
> we could limit the goroutines to the number of available cores, no?

not really as one core should be reserved for compaction

krasi-georgiev:
yes sure, my point was that we could still use concurrency but also limit the total goroutines.

bwplotka:
Ok, I am confused here. 

> This is still going to create goroutines when there's only the local tsdb to talk to. We want to talk remotely concurrently, but not locally.

What exactly this means? PromQL can have 100 Select to make and it will do it sequentially, no matter what storage is right now (: Let's say someone wants to scale read part of Prometheus (or in fact any user of `promql` package). So are we ok to add concurrency here or not?

promql/engine.go
liguozhong: I see. I'll fix it right away

brian-brazil:
> So are we ok to add concurrency here or not?

We want concurrency, but not for the tsdb so as to limit CPU usage to one core so that it's all predictable. This is in line with the original issue, which is about remote read queries.

roidelapluie:
> > So are we ok to add concurrency here or not?
> 
> We want concurrency, but not for the tsdb so as to limit CPU usage to one core so that it's all predictable. This is in line with the original issue, which is about remote read queries.

When we run 10 queries in parallel, there is no select happening in parallel?

slrtbtfs:
> we could limit the goroutines to the number of available cores, no?

The select goroutines will spend a lot of time waiting on I/O, so even if we have as many goroutines as cores we might not have full CPU load.

Also, schedulers exist, so even in case of full CPU load other tasks would still get some attention.

brian-brazil:
> The select goroutines will spend a lot of time waiting on I/O,

It's mostly CPU based, Select only touches the indexes which will tend to be in RAM.

> Also, schedulers exist, so even in case of full CPU load other tasks would still get some attention.

Allowing PromQL queries to use unbounded CPU is a bad idea, that'll affect ingestion and other things on the machine. Overload is overload, schedulers don't change that.

> When we run 10 queries in parallel, there is no select happening in parallel?

There is, limited to no more than 10 cores.

slrtbtfs:
Wouldn't it work to limit the total number of concurrent Select Calls in the storage implementation?

In that case local storage could always do only one select call at once and remote storage can have a higher limit.

roidelapluie:
> Wouldn't it work to limit the total number of concurrent Select Calls in the storage implementation?
> 
> In that case local storage could always do only one select call at once and remote storage can have a higher limit.

That would make little queries a lot longer if they have to wait for long queries to run

brian-brazil:
> In that case local storage could always do only one select call at once and remote storage can have a higher limit.

That's basically what I'm saying.

> That would make little queries a lot longer if they have to wait for long queries to run

That's already the case today, and it hasn't caused any issues I'm aware of.

promql/engine.go
liguozhong: done.

code /promql/
image
code /fanout/
image

roidelapluie:
What I noted on #6878 is that we would need a design document and I have yet to see it.

roidelapluie:
> What I noted on #6878 is that we would need a design document and I have yet to see it.

in https://github.com/prometheus/prometheus/pull/6887 *

brancz:
@roidelapluie where did you say that? I went over there and tried to search for it, but couldn't find it.

I personally don't think we need a design for this. I'm not saying we shouldn't discuss the strategy, but a full blown design doc seems exaggerated.

roidelapluie:
> @roidelapluie where did you say that? I went over there and tried to search for it, but couldn't find it.
> 
> I personally don't think we need a design for this. I'm not saying we shouldn't discuss the strategy, but a full blown design doc seems exaggerated.

I did not say that but the previous PR tended to show that we should not run queries in parallel against local storage

brian-brazil:
Yeah, this requires a bit of thought and upfront design but nothing that can't be done in the context of an issue/PR.

brancz:
@roidelapluie sorry misunderstood your comment then. Seems like we’re in agreement :)

liguozhong:
The idea of this PR is to first convert select() to batch query by adding a func of selects(). On the basis of batch query, add support for .Select calls concurrency. 
The screenshot is the flow chart of this PR.

image

pkg/gate/gate_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	maxGoRoutine := 3
+	gate := New(maxGoRoutine)
+	callCount := 10
+	overloadChan := make(chan int, callCount-maxGoRoutine)
+	var wg sync.WaitGroup
+
+	for i := 0; i < callCount; i++ {
+		go func() {
+			wg.Add(1)
```
slrtbtfs, line 34: The `wg.Add(1)` must happen before launching the goroutine, otherwise this is a race condition.

liguozhong: thanks，done

bwplotka:
Hey, @liguozhong are you available to continue this PR? I and @kakkoyun might have internally more time to move this forward this sprint if you need some help. I think we still need some approved designs for this. I might think we need to elaborate more than just couple of screenshots. I like Google Docs because we can comment easily on each stuff, but we could do this here or on https://github.com/prometheus/prometheus/issues/6878. Let's work together on some design that would do if you are ok with that. (:

liguozhong:
> Hey, @liguozhong are you available to continue this PR? I and @kakkoyun might have internally more time to move this forward this sprint if you need some help. I think we still need some approved designs for this. I might think we need to elaborate more than just couple of screenshots. I like Google Docs because we can comment easily on each stuff, but we could do this here or on #6878. Let's work together on some design that would do if you are ok with that. (:

thanks . I have no experience with Google Docs, and I agree to let other contributors complete this PR. Looking forward to seeing this feature in 2.18.0 release.
I will continue to pay attention to this issue, our production environment also encountered this problem.
My experience is just to write code, if there are any work items that I can accomplish, I am happy to complete.

kakkoyun:
@liguozhong Awesome, we will try to move this forward then with some design first. Make sure we have a good understanding of the issue and a consensus on implementation details.

liguozhong:
> @liguozhong Awesome, we will try to move this forward then with some design first. Make sure we have a good understanding of the issue and a consensus on implementation details.

Looking forward to completion soon

bwplotka:
Let's close this for now until we have agreed design in place. :hugs:

PR #7096, opened 2020-04-07
Use local tsdb for federation metrics

### LINKED ISSUE #4270: Remote storage should not be used by federation
Capturing IRC discussion:
```
10:52 < iksaif> hey, do we really want /federate to use remote read?
13:18 < bbrazil> I'm not sure what the current behaviour is, but it should not
13:26 < jrv> turns out this problem exists since 2.0.0... adding a remote_read endpoint there and then hitting /federate causes a nil pointer crash as well
13:27 < jrv> in 1.x I handed down some special parameters to a query that indicated that only local storage should be consulted (for rules and federation), but that didn't make it into 2.x
13:30 < jrv> in 1.x it was done like this: https://github.com/prometheus/prometheus/blob/release-1.8/cmd/prometheus/main.go#L129
13:48 < jrv> In 2.x, would such a (only use local storage) hint now go into the `SelectParams`?
13:53 < bbrazil> jrv: no, you should use a storage that doesn't include the remote ones
13:53 < bbrazil> so you'd just be using the local storage rather than a fanout storage that included everything
13:54 < jrv> bbrazil: ok, so we should hand in two storages into the web layer, a local-only one for federation and a combined one for normal queries
13:55 < jrv> but even for normal queries at some point we want the user to be able to set an argument that indicates whether to also hit remote storage or not... same for rules. Might be cleaner to abstract 
             that away into a query option and a common place (like fanin in 1.x) then decides which one of the underlying queriers to use?
13:56 < jrv> so we don't put that logic and hand out two storages into each of the places that have to make this decision
13:56 < bbrazil> jrv: something like that, but it doesn't belong as an option to the storage itself
13:56 < bbrazil> we've only 2 places that need this, no need to get too fancy
13:58 < jrv> bbrazil: yeah, not to the storage itself, but the fanin layer above it then
13:58 < jrv> like in 1.x
13:59 < bbrazil> I see us passing in two storages, and chosing which to use in the api/rules code
14:02 < jrv> not sure how that's better, but works too
14:02 < jrv> I think from a cleanliness perspective I prefer passing in the fanin one and making that aware of what's local and what's remote
14:09 < bbrazil> to me that's making it harder to dig down to find the relevant if statement
```

yeya24:

Fixes #4270

### DIFF
--- a/web/web.go
+++ b/web/web.go
@@ -180,6 +180,7 @@ type Handler struct {
 	context       context.Context
 	tsdb          func() *tsdb.DB
 	storage       storage.Storage
+	localStorage  storage.Storage
 	notifier      *notifier.Manager
 
 	apiV1 *api_v1.API
@@ -289,6 +290,10 @@ func New(logger log.Logger, o *Options) *Handler {
 		now: model.Now,
 
 		ready: 0,
+	}
+
+	if h.tsdb != nil {
+		h.localStorage = h.tsdb()
 	}
 
 	h.apiV1 = api_v1.NewAPI(h.queryEngine, h.storage, h.scrapeManager, h.notifier,
--- a/web/web.go
+++ b/web/web.go
@@ -285,15 +285,12 @@ func New(logger log.Logger, o *Options) *Handler {
 		lookbackDelta: o.LookbackDelta,
 		tsdb:          o.TSDB,
 		storage:       o.Storage,
+		localStorage:  o.TSDB(),
 		notifier:      o.Notifier,
 
 		now: model.Now,
 
 		ready: 0,
-	}
-
-	if h.tsdb != nil {
-		h.localStorage = h.tsdb()
 	}
 
 	h.apiV1 = api_v1.NewAPI(h.queryEngine, h.storage, h.scrapeManager, h.notifier,
--- a/web/federate.go
+++ b/web/federate.go
@@ -75,7 +75,7 @@ func (h *Handler) federation(w http.ResponseWriter, req *http.Request) {
 	)
 	w.Header().Set("Content-Type", string(format))
 
-	q, err := h.storage.Querier(req.Context(), mint, maxt)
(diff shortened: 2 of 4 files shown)


web/web.go
```diff
@@ -291,6 +292,10 @@ func New(logger log.Logger, o *Options) *Handler {
 		ready: 0,
 	}
 
+	if h.tsdb != nil {
```
brian-brazil, line 295: Why would tsdb ever be nil?

roidelapluie: probably when cortex imports web

brian-brazil: Seems unlikely as it'd be panicing a lot, plus that's not a good reason not to simplify the code.

yeya24: It happens a lot in the test code here https://github.com/prometheus/prometheus/blob/master/web/web_test.go. TSDB is not initialized here. Does it make sense to you or I should initialize TSDB in the test code?

brian-brazil: I'd update the test code. This logic is already fairly intricate without making more complex.

yeya24: OK. Let me update it.

yeya24: Updated. Please take a look. Thanks!

brian-brazil:
Thanks!

juliantaylor:
This is a pretty big breaking change. It breaks setups where a prometheus is used to provide a single aggregation point for other prometheus instances.
While this is not a recommended way to use prometheus and there are better options to solve this problem (e.g. promxy or thanos), it does still work.

As the code change seems pretty small, so it appears to not be a big maintenance overhead for you, is it maybe possible to make this behaviour configurable?
A commandline one for all config flag that may also default to only use local data would be fine.

brian-brazil:
I don't think we'd add a configurable purely to allow for a non-recommended architecture. If you need federation from multiple instances, federate from them directly.

PR #7098, opened 2020-04-07
Reduce how much old WAL we keep around.

brian-brazil:
Previously we were keeping up to around 6 hours of WAL around by
removing 1/3 every hours. This was excessive, so switch to removing 2/3
which will up to around 3 hours of WAL around.

This will roughly halve the size of the WAL and halve startup time for
those who are I/O bound. This may increase the checkpoint size for
those with certain churn patterns, but by much less than we're saving
from the segments.

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -681,9 +681,11 @@ func (h *Head) Truncate(mint int64) (err error) {
 	if last < 0 {
 		return nil // no segments yet.
 	}
-	// The lower third of segments should contain mostly obsolete samples.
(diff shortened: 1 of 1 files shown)


brian-brazil:
I presume because noone did the math to determine the steady state, I've read worked with this code many times but only noticed it when writing my next blog post.

angelincb:
Hi @brian-brazil 

I have a question. What if the remote write is down for more than 3 hours? Is there a way to retain WAL data for longer duration and not loose data?

PR #7100, opened 2020-04-07
switch to go 1.14

roidelapluie:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ executors:
   # should also be updated.
   golang:
     docker:
-    - image: circleci/golang:1.13-node
+    - image: circleci/golang:1.14-node
 
   fuzzit:
     docker:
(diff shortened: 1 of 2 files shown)


roidelapluie:
@simonpasquier any idea why we don't have yarn here?

simonpasquier:
It was an oversight when reviewing https://github.com/prometheus/golang-builder/pull/78. Because `yarn` wasn't added to the 1.12 image, it was missing when we switched to 1.14 from 1.12...

https://github.com/prometheus/golang-builder/pull/87 should fix it.

roidelapluie:
it is green. should we run prombench?

roidelapluie:
Prombench looked really good.

geekodour:
@roidelapluie prombench usually breaks when there is a go version bump, o.O I think prombench built prometheus with 1.13 when you ran it.

PR #7104, opened 2020-04-08
Update mixin dashboards and alerts for new remote write label names.

cstyan:
In recent release we changed the remote write metrics labesl from `queue` to `remote_name` and `url` but never updated (most of ) the related alerts/dashboards.


brancz:
lgtm

PR #7105, opened 2020-04-08
Add missing remote write config name to docs.

cstyan:
The config option to name a remote write config/endpoint was added in https://github.com/prometheus/prometheus/pull/6043 but I didn't update docs.

### DIFF
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -1294,6 +1294,11 @@ url: <string>
 write_relabel_configs:
   [ - <relabel_config> ... ]
 
+# Name of the remote write config, which if specified must be unique among remote write configs. 
+# The name will be used in metrics and logging in place of a generated value to help users distinguish between
+# remote write configs.
+[ name: <string> ]
+
 # Sets the `Authorization` header on every remote write request with the
 # configured username and password.
 # password and password_file are mutually exclusive.
@@ -1348,6 +1353,11 @@ with this feature.
 ```yaml
 # The URL of the endpoint to query from.
 url: <string>
+
+# Name of the remote read config, which if specified must be unique among remote read configs. 
+# The name will be used in metrics and logging in place of a generated value to help users distiguish between
+# remote read configs.
+[ name: <string> ]
 
 # An optional list of equality matchers which have to be
 # present in a selector to query the remote read endpoint.


docs/configuration/configuration.md
```diff
@@ -1294,6 +1294,9 @@ url: <string>
 write_relabel_configs:
   [ - <relabel_config> ... ]
 
+# Name of the remote write config, useful for users who have multiple remote writes configured.
```
brian-brazil, line 1297: This is a bit vague, what does it actually do?

cstyan: Just that, a name for the remote write queue. If you don't set one it defaults to a hash of that remote write config. It's useful for users who have multiple remote writes, especially if they have more than one remote write to the same endpoint but with different relabel configs for example. I can include that last bit in the comment here?

brian-brazil: That still doesn't explain what it actually does, only the general intention behind it. You need to mention where this value will actually appear.

cstyan: It's strictly for identification (by the user, not internally within remote write) of a given queue. So the name would show up in a field during structured logging, a label in remote write metrics, etc.

brian-brazil: If it has a semantic effect (and metric labels count) that needs to be mentioned here. Otherwise it's unclear what this actually does, it could just be a form of comment that never leaves the config file.

cstyan: updated the comment

docs/configuration/configuration.md
```diff
@@ -1294,6 +1294,11 @@ url: <string>
...
+# The name will be used in structured logging and as a label in remote write metrics to help users identify
+# a unique remote write queue.
```
roidelapluie, line 1299: do we check for uniqueness?

csmarchbanks: Yes, uniqueness is required as part of parsing the configuration.

docs/configuration/configuration.md
brian-brazil: This is still not quite right for a reference doc. I'd do something like:

Name of the remote write config, which if specified must be unique. This will be used in metrics etc. in place of a generated value to help users distinguish between remote write configs.


This is also not the syntax we use to indicate an optional field.

brian-brazil:
:+1:

docs/configuration/configuration.md
```diff
@@ -1294,6 +1294,11 @@ url: <string>
 write_relabel_configs:
   [ - <relabel_config> ... ]
 
+# Name of the remote write config, which if specified must be unique among remote write configs. 
+# The name will be used in metrics and logging in place of a generated value to help users distibguish between
```
csmarchbanks, line 1298: Distinguish has a typo in both descriptions.

cstyan: thanks :+1:

PR #7106, opened 2020-04-09
Fix warning: redundant type from array, slice, or map composite literal

Hellcatlk:
Signed-off-by: ZouYu <zouy.fnst@cn.fujitsu.com>


brian-brazil:
Thanks!

PR #7116, opened 2020-04-11
Add unit test for Get in pkg/labels/labels.go

hs0210:
This PR is about adding a unit test for Get in pkg/labels/labels.go.

### DIFF
--- a/pkg/labels/labels_test.go
+++ b/pkg/labels/labels_test.go
@@ -485,3 +485,16 @@ func TestLabels_Has(t *testing.T) {
 		testutil.Equals(t, test.expected, got, "unexpected comparison result for test case %d", i)
 	}
 }
+
+func TestLabels_Get(t *testing.T) {
+	testutil.Equals(t, "", Labels{{"aaa", "111"}, {"bbb", "222"}}.Get("foo"))
+	testutil.Equals(t, "111", Labels{{"aaa", "111"}, {"bbb", "222"}}.Get("aaa"))
+}
+
+func TestLabels_Copy(t *testing.T) {
+	testutil.Equals(t, Labels{{"aaa", "111"}, {"bbb", "222"}}, Labels{{"aaa", "111"}, {"bbb", "222"}}.Copy())
+}
+
+func TestLabels_Map(t *testing.T) {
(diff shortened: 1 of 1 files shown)


cstyan:
LGTM but given that this is the 2nd PR for tests in this package recently it seems likely there might be a few more that could be added. If you find any please add them to this PR @hs0210 :)

hs0210:
PTAL

hs0210:
@cstyan I have added two other tests. PTAL. If there are still problems, I will fix it.

pkg/labels/labels_test.go
```diff
@@ -485,3 +485,82 @@ func TestLabels_Has(t *testing.T) {
 		testutil.Equals(t, test.expected, got, "unexpected comparison result for test case %d", i)
 	}
 }
+
+func TestLabels_Get(t *testing.T) {
+	tests := []struct {
```
bwplotka, line 490: This table test for single line test might be an overkill..

The same test will have just 2 lines...

Can we add just two lines instead?

pkg/labels/labels_test.go
```diff
@@ -485,3 +485,82 @@ func TestLabels_Has(t *testing.T) {
...
+	}
+
+	for i, test := range tests {
+		got := labelsSet.Get(test.input)
+		testutil.Equals(t, test.expected, got, "unexpected comparison result for test case %d", i)
+	}
+}
+
+func TestLabels_Copy(t *testing.T) {
+	expected := Labels{
```
bwplotka, line 522: same here, but one line

pkg/labels/labels_test.go
```diff
@@ -485,3 +485,82 @@ func TestLabels_Has(t *testing.T) {
...
+	got := labelsSet.Copy()
+	testutil.Equals(t, expected, got)
+}
+
+func TestLabels_Map(t *testing.T) {
+	expected := make(map[string]string)
+	expected["aaa"] = "111"
+	expected["bbb"] = "222"
+
+	labelsSet := Labels{
```
bwplotka, line 553: No need for any of those variables... just one line is enough.

```
testutil.Equals(t, map[string]string{"aaa":"111","bbb":"222"}, Labels{
{
			Name:  "aaa",
			Value: "111",
		},
		{
			Name:  "bbb",
			Value: "222",
		},
	}.Map())
```

hs0210:
@bwplotka PTAL.

PR #7117, opened 2020-04-13
Docs: Edit query basics

oddlittlebird:
Minor edits to improve grammar and formatting. No content changes except for adding an introduction sentences to the Troubleshooting section, previously known as "Gotchas."

### DIFF
--- a/docs/querying/basics.md
+++ b/docs/querying/basics.md
@@ -199,11 +199,7 @@ PromQL supports line comments that start with `#`. Example:
 
         # This is a comment
 
-## Troubleshooting
-
-The following are common difficulties you might encounter while querying.
-
-### Staleness
+## Staleness
 
 When queries are run, timestamps at which to sample data are selected
 independently of the actual present time series data. This is mainly to support
@@ -213,8 +209,8 @@ Prometheus needs to assign a value at those timestamps for each relevant time
 series. It does so by simply taking the newest sample before this timestamp.
 
 If a target scrape or rule evaluation no longer returns a sample for a time
-series that was previously present, that time series will be marked as stale.
-If a target is removed, its previously returned time series will be marked as
+series that was previously present, then that time series is marked as stale.
+If a target is removed, its previously returned time series is marked as
 stale soon afterwards.
 
 If a query is evaluated at a sampling timestamp after a time series is marked
@@ -227,9 +223,9 @@ effectively means that time series "disappear" from graphs at times where their
 latest collected sample is older than five minutes or after they are marked stale.
 
 Staleness will not be marked for time series that have timestamps included in
-their scrapes. Only the 5 minute threshold will be applied in that case.
+their scrapes. Only the five-minute threshold applies in that case.
 
-### Avoiding slow queries and overloads
+## Avoiding slow queries and overloads
 
 If a query needs to operate on a very large amount of data, graphing it might
 time out or overload the server or browser. Thus, when constructing queries
@@ -243,7 +239,7 @@ rule](../configuration/recording_rules.md#recording-rules).
 This is especially relevant for Prometheus's query language, where a bare
 metric name selector like `api_http_requests_total` could expand to thousands
 of time series with different labels. Also keep in mind that expressions which
-aggregate over many time series will generate load on the server even if the
+aggregate over many time series generate load on the server even if the
 output is only a small number of time series. This is similar to how it would
 be slow to sum all values of a column in a relational database, even if the
 output value is only a single number.
--- a/docs/querying/basics.md
+++ b/docs/querying/basics.md
@@ -58,21 +58,20 @@ Scalar float values can be literally written as numbers of the form
 
     -2.43
 
-## Time series Selectors
+## Time series selectors
 
 ### Instant vector selectors
 
 Instant vector selectors allow the selection of a set of time series and a
-single sample value for each at a given timestamp (instant): in the simplest
+single sample value for each at a given timestamp (instant). In the simplest
 form, only a metric name is specified. This results in an instant vector
 containing elements for all time series that have this metric name.
 
-This example selects all time series that have the `http_requests_total` metric
-name:
+This example selects all time series that have the `http_requests_total` metric name:
 
     http_requests_total
 
-It is possible to filter these time series further by appending a comma separated list of label
+You can filter time series further by appending a comma separated list of label
 matchers in curly braces (`{}`).
 
 This example selects only those time series with the `http_requests_total`
@@ -81,8 +80,8 @@ metric name that also have the `job` label set to `prometheus` and their
 
     http_requests_total{job="prometheus",group="canary"}
 
-It is also possible to negatively match a label value, or to match label values
-against regular expressions. The following label matching operators exist:
+You can also negatively match a label value, or match label values
+against regular expressions with the following label-matching operators:
 
 * `=`: Select labels that are exactly equal to the provided string.
 * `!=`: Select labels that are not equal to the provided string.
@@ -90,7 +89,7 @@ against regular expressions. The following label matching operators exist:
 * `!~`: Select labels that do not regex-match the provided string.
 
 For example, this selects all `http_requests_total` time series for `staging`,
-`testing`, and `development` environments and HTTP methods other than `GET`.
+`testing`, and `development` environments and HTTP methods other than `GET`:
 
     http_requests_total{environment=~"staging|testing|development",method!="GET"}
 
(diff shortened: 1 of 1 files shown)


docs/querying/basics.md
```diff
@@ -4,17 +4,15 @@ nav_title: Basics
...
-Language) that lets the user select and aggregate time series data in real
+Language) that lets you select and aggregate time series data in real
 time. The result of an expression can either be shown as a graph, viewed as
 tabular data in Prometheus's expression browser, or consumed by external
 systems via the [HTTP API](api.md).
 
-## Examples
-
-This document is meant as a reference. For learning, it might be easier to
+> **Note:** This document is meant as a reference. For learning, it might be easier to
```
brian-brazil, line 15: No need to shout.

docs/querying/basics.md
```diff
@@ -202,7 +199,9 @@ PromQL supports line comments that start with `#`. Example:
 
         # This is a comment
 
-## Gotchas
+## Troubleshooting
```
brian-brazil, line 202: This section contains reference material about PromQL semantics, it does not make sense to be called troubleshooting.

docs/querying/basics.md
```diff
@@ -4,17 +4,15 @@ nav_title: Basics
 sort_rank: 1
 ---
 
-# Querying Prometheus
+# Querying basics
 
 Prometheus provides a functional query language called PromQL (Prometheus Query
-Language) that lets the user select and aggregate time series data in real
+Language) that lets you select and aggregate time series data in real
```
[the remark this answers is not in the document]
oddlittlebird: The problem with passive voice is that it is less clear than active voice and uses more words. Best practice in technical writing is to use active voice whenever possible.

docs/querying/basics.md
oddlittlebird: I changed it to Troubleshooting because it seems to be a section with information about how to solve common problems. "Gotchas" is vague and unclear, and certainly would not be commonly used in web searches. "Troubleshoot queries" would be.

Any countersuggestions to Troubleshooting?

docs/querying/basics.md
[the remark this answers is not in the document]
brian-brazil: Our docs are primarily passive voice rather than 2nd person, and if that's to be changed that would need to be discussed more broadly.

docs/querying/basics.md
brian-brazil: Maybe just remove the heading?

The staleness stuff is reference so belongs here, while the performance stuff is more guidey but it's probably the best place for it.

oddlittlebird: Done

docs/querying/basics.md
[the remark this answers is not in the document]
codesome: My $0.02: active voice here feels more engaging as if the docs are talking to "me", so maybe we could gradually shift all docs to active voice?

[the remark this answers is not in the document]
brian-brazil: One way or the other making piecemeal changes to our doc style is not the way to do that. As is stands this is not our style, so PRs should be looking to eliminate rather than propagate it.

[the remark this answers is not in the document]
oddlittlebird: Where is this stated? I have not found a style guide for documentation with any sort of grammar or formating rules for this project.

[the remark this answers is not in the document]
brian-brazil: I looked at what was common in this doc and the related ones, consistency is enough of a reason.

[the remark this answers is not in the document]
roidelapluie: I agree with @brian-brazil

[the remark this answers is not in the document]
oddlittlebird: Reverted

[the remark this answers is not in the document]
oddlittlebird: But...consistent? Really? Here are some examples of the second-person already existing in your docs:

https://prometheus.io/docs/practices/histograms/#count-and-sum-of-observations
  ` In principle, however, you can use summaries and histograms to observe negative values (e.g. temperatures in centigrade). In that case, the sum of observations can go down, so you cannot apply rate() to it anymore.`

https://prometheus.io/docs/practices/histograms/#quantiles
`You can use both summaries and histograms to calculate so-called φ-quantiles, where 0 ≤ φ ≤ 1. `

https://prometheus.io/docs/prometheus/latest/configuration/configuration/
`You may wish to check out the 3rd party Prometheus Operator, which automates the Prometheus setup on top of Kubernetes.`

I could go on, but I think you get the idea.

I'm not even going to touch all the many examples of active voice that are already in the Prometheus docs. Point is, you are not consistent. Why not allow people to improve the docs and eventually move the needle towards "better"?

roidelapluie:
On 14 Apr 13:50, Diana Payton wrote:
> I could go on, but I think you get the idea.
> 
> I'm not even going to touch all the many examples of active voice that are already in the Prometheus docs. Point is, you are not consistent. Why not allow people to improve the docs and eventually move the needle towards "better"?

When I read Brian's point, he just proposed that this should be discussed
more widely to get broader consensus before changing this.

docs/querying/basics.md
```diff
@@ -60,21 +58,20 @@ Scalar float values can be literally written as numbers of the form
...
 containing elements for all time series that have this metric name.
 
-This example selects all time series that have the `http_requests_total` metric
-name:
+This example selects all time series that have the `http_requests_total` metric name:
 
     http_requests_total
 
-It is possible to filter these time series further by appending a comma separated list of label
+You can filter time series further by appending a comma separated list of label
```
brian-brazil, line 74: Don't change the style of the text

docs/querying/basics.md
```diff
@@ -83,16 +80,16 @@ metric name that also have the `job` label set to `prometheus` and their
 
     http_requests_total{job="prometheus",group="canary"}
 
-It is also possible to negatively match a label value, or to match label values
-against regular expressions. The following label matching operators exist:
+You can also negatively match a label value, or match label values
```
brian-brazil, line 83: Ditto

docs/querying/basics.md
brian-brazil: This hasn't been reverted.

brian-brazil:
Different people have different opinions of what "improve" is, and PRs which piecemeal attempt to change style usually only end up increasing inconsistency which is worse from my standpoint - for example the vast majority of this page would still be in the passive voice.

RichiH:
FWIW, personally, I happen to prefer active voice; having one consistent scheme throughout the docs is preferable, though unifying either way is unlikely to happen in one huge PR, but can done in a one-file PR at a time.

I like the general direction of this PR, e.g. "will be evaluated at X" is fine, but should be present tense, i.e. ongoing, when it's not linked to a specific future time.

brian-brazil:
Keep in mind that these are reference docs describing the behaviour of Prometheus, not guides or tutorials showing how you could use these featuers.

RichiH:
We have had this discussion in the past and I disagree that we need a reference first and foremost, we need docs first and foremost. I tried extending them that way in 2016.

Would you agree with copying everything we have into `reference/` and then using what we have as a starting to point to rewrite `docs/`?

brian-brazil:
We need several types of docs, right now we mainly have reference docs - and I intend to keep them as useful reference docs.

If you wish to add guides, there's already a section on the website for those.

brian-brazil:
> Would you agree with copying everything we have into reference/ and then using what we have as a starting to point to rewrite docs/?

I don't agree with this. It'd be busy work, and would break all the URLs.

RichiH:
> I don't agree with this. It'd be busy work, and would break all the URLs.

No, it would simply mean that existing incoming links point to something useful for learning, instead of something useful for everyone who already knows how it works.

brian-brazil:
Guides are a completely different type of documentation than reference, the layout of all the pages etc. would have to change to be useful. Such documentation should be built up independently, and there's already a section on our website for that. Even if we were to do as you say we'd still be breaking users who rely on those URLs to contain their current content in its current style - just because some users wish the pages had different content doesn't change that they have the content they have now.

This is all off topic for this PR though, which is aiming to make only some minor tweaks.

RichiH:
I don't think any single one of us, myself included, have the authority to
decide on direction alone; I agree that we need wider discussion and will
move this to devel@.

Sent by mobile; please excuse my brevity.

docs/querying/basics.md
juliusv: I think this is worth calling out specifically, at least if we want to keep that strict separation.

juliusv:
A general comment: Some of the tone in this review was such that if I had been a new contributor and this was one of my first interactions with the project, I would likely never want to contribute to Prometheus again. I get that under a lot of work stress, review comments become curt and harsh-sounding without any bad intentions, but we should always strive to do better.

On the content-side: I agree that consistency is good, though we've never been terribly consistent to begin with. I'm fine with the small changes proposed here though.

docs/querying/basics.md
brian-brazil: That applies to this entire section of docs though, not just this one.

juliusv: I agree, but isn't that aspect separate from how bold we make the markup around it?

brian-brazil: I see your point. This is already an entire section to itself, adding bold and a note on top of that seems excessive - the current text is already emphasised quite a bit.

RichiH:
For cross-reference: https://groups.google.com/forum/#!topic/prometheus-developers/2eG63sR_W4c

docs/querying/basics.md
oddlittlebird: I removed the section titled Examples and moved the content to this note to reduce confusion. When a reader sees a section titled "Examples," they expect to have a section with examples in it. This was not what the section did. A note is clearer and more appropriate in this context.

docs/querying/basics.md
oddlittlebird: "Reference documentation is about describing a piece of software, so we should be talking about the software in the 3rd person rather than talking to the user in the 2nd person. "it is possible to" would be unclear for reference documentation, I don't want to know some things that might be possible - I want to know exactly how the software behaves." - Brian Brazil

Would you prefer "Prometheus allows you to"?

docs/querying/basics.md
brian-brazil: I can see that argument.

docs/querying/basics.md
brian-brazil: I'd personally go with something like "Time series can be further filtered by"

roidelapluie:

We have decided to close this pull request for now as there is more fundamental work to do with the documentation first, in the docs WG.

cc @hdost maybe we can still cherry-pick some improvements in the PR in a new PR?

PR #7125, opened 2020-04-15
API: Allow `TargetRetriever` to receive a Context

### LINKED ISSUE #7103: Proposal: Add a context to API dependencies interfaces
## Proposal

At the moment, API dependencies interfaces don't support any arguments e.g.

```golang
type targetRetriever interface {
	TargetsActive() map[string][]*scrape.Target
	TargetsDropped() map[string][]*scrape.Target
}

type alertmanagerRetriever interface {
	Alertmanagers() []*url.URL
	DroppedAlertmanagers() []*url.URL
}

type rulesRetriever interface {
	RuleGroups() []*rules.Group
	AlertingRules() []*rules.AlertingRule
}
```

Given the Prometheus API package can be exported by third-parties, it is useful to pass a context to the dependant functions as they get called directly on the API endpoints. This allows for third-parties to introduce cancellation or authentication concerns on the implementations. 

Particularly, we need it in Cortex as it handles authentication via the Context.

gotjosh:
Fixes #7103

I only changed `TargetRetriever` as that is the originally agreed scope on the issue, but the same could be applied to the other dependencies.

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -100,7 +100,8 @@ func (e *apiError) Error() string {
 	return fmt.Sprintf("%s: %s", e.typ, e.err)
 }
 
-type targetRetriever interface {
+// TargetRetriever provides the list of active/dropped targets to scrape or not.
+type TargetRetriever interface {
 	TargetsActive() map[string][]*scrape.Target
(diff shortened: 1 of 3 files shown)


brian-brazil:
Thanks!

Issue #7127, opened 2020-04-15
panic in parser


roidelapluie:
```
parser panic: interface conversion: interface {} is *parser.BinaryExpr, not *parser.VectorSelector
goroutine 4130297 [running]:
github.com/prometheus/prometheus/promql/parser.(*parser).recover(0xc010833000, 0xc050374e40)
	/app/promql/parser/parse.go:269 +0x122
panic(0x224d860, 0xc0d7652cc0)
	/usr/local/go/src/runtime/panic.go:679 +0x1b2
github.com/prometheus/prometheus/promql/parser.ParseMetricSelector(0xc26be78ab0, 0x2c, 0x0, 0x0, 0x0, 0x0, 0x0)
	/app/promql/parser/parse.go:156 +0x219
github.com/prometheus/prometheus/web/api/v1.(*API).series(0xc0002d9950, 0xc0d74e0b00, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/app/web/api/v1/api.go:518 +0x7d7
github.com/prometheus/prometheus/web/api/v1.(*API).Register.func1.1(0x2e06680, 0xc0d74edb60, 0xc0d74e0b00)
	/app/web/api/v1/api.go:251 +0x8e
net/http.HandlerFunc.ServeHTTP(0xc00010dd40, 0x2e06680, 0xc0d74edb60, 0xc0d74e0b00)
	/usr/local/go/src/net/http/server.go:2007 +0x44
github.com/prometheus/prometheus/util/httputil.CompressionHandler.ServeHTTP(0x2dced60, 0xc00010dd40, 0x7f4ed5aa1660, 0xc0d74a9f90, 0xc0d74e0b00)
	/app/util/httputil/compression.go:90 +0x7c
github.com/prometheus/prometheus/web.(*Handler).testReady.func1(0x7f4ed5aa1660, 0xc0d74a9f90, 0xc0d74e0b00)
	/app/web/web.go:489 +0x55
net/http.HandlerFunc.ServeHTTP(0xc00010dda0, 0x7f4ed5aa1660, 0xc0d74a9f90, 0xc0d74e0b00)
	/usr/local/go/src/net/http/server.go:2007 +0x44
github.com/prometheus/client_golang/prometheus/promhttp.InstrumentHandlerResponseSize.func1(0x7f4ed5aa1660, 0xc0d74a9f40, 0xc0d74e0b00)
[...]

PR #7129, opened 2020-04-15
Defer finalizer

roidelapluie:
If we panic during the api response, we could in theory not close the queriers.

Could be related to #7120 ?

cc @beorn7 

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -249,15 +249,15 @@ func (api *API) Register(r *route.Router) {
 		hf := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			httputil.SetCORS(w, api.CORSOrigin, r)
 			result := f(r)
+			if result.finalizer != nil {
+				defer result.finalizer()
+			}
 			if result.err != nil {
 				api.respondError(w, result.err, result.data)
 			} else if result.data != nil {
 				api.respond(w, result.data, result.warnings)
 			} else {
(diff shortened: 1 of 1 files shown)


brian-brazil:
:+1: 

I'm not sure how we could panic there, but it doesn't hurt to be robust.

roidelapluie:
Closing in favor of #7130

beorn7:
This is a good catch.

But #7127 would not be affected by this.

However, by staring at this code, I realized how panicking in certain scenarios can evade the `qry.Close` call. If [this `qry.Exec` call](https://github.com/prometheus/prometheus/blob/ae041f97cfc6f43494bed65ec4ea4e3a0cf2ac69/web/api/v1/api.go#L346) panics or [this one](https://github.com/prometheus/prometheus/blob/ae041f97cfc6f43494bed65ec4ea4e3a0cf2ac69/web/api/v1/api.go#L418) or [this q.LabelValues call](https://github.com/prometheus/prometheus/blob/ae041f97cfc6f43494bed65ec4ea4e3a0cf2ac69/web/api/v1/api.go#L483) or anything in this method between these calls and a regular non-panic return, then the caller will never get to call the finalizer. What we need is a `defer` that will only trigger if the function is left by panicking rather than a replar return. Not sure if there is a sane way to do that.

Alternatively, we have to rely on those calls to never panic… which doesn't sound robust.

beorn7:
I think I have a sane approach. Will send a PR in a minute.

PR #7130, opened 2020-04-15
api/v1: remove finalizer

roidelapluie:
Lots of places already defer querier.Close, this PR normalizes that by
applying it to all functions.

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -150,10 +150,9 @@ type response struct {
 }
 
 type apiFuncResult struct {
-	data      interface{}
-	err       *apiError
-	warnings  storage.Warnings
-	finalizer func()
+	data     interface{}
+	err      *apiError
+	warnings storage.Warnings
 }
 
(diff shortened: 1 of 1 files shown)


roidelapluie:
Could be related to #7120 ?

web/api/v1/api.go
```diff
@@ -338,14 +334,15 @@ func (api *API) query(r *http.Request) apiFuncResult {
 	qry, err := api.QueryEngine.NewInstantQuery(api.Queryable, r.FormValue("query"), ts)
 	if err != nil {
 		err = errors.Wrapf(err, "invalid parameter 'query'")
-		return apiFuncResult{nil, &apiError{errorBadData, err}, nil, nil}
+		return apiFuncResult{nil, &apiError{errorBadData, err}, nil}
 	}
+	defer qry.Close()
```
brian-brazil, line 339: This is incorrect, as we need to keep everything allocated until the API result has been sent out over HTTP - which is after this function and its defers have completed.

PR #7133, opened 2020-04-16
notifier: forget unlock before return

BurtonQin:
In function sendAll(), `ams.mtx.RLock()` is called at the start of a loop
https://github.com/prometheus/prometheus/blob/24af5049bbf665376d8da3d72404d670cdabfa9d/notifier/notifier.go#L471-L477
`ams.mtx.RUnlock()` is called at the end of the loop
https://github.com/prometheus/prometheus/blob/24af5049bbf665376d8da3d72404d670cdabfa9d/notifier/notifier.go#L536

However, `ams.mtx.RUnlock()` is not called before return on [486](https://github.com/prometheus/prometheus/blob/24af5049bbf665376d8da3d72404d670cdabfa9d/notifier/notifier.go#L486), [500](https://github.com/prometheus/prometheus/blob/24af5049bbf665376d8da3d72404d670cdabfa9d/notifier/notifier.go#L500), [512](https://github.com/prometheus/prometheus/blob/24af5049bbf665376d8da3d72404d670cdabfa9d/notifier/notifier.go#L512)

The fix adds `ams.mtx.RUnlock()` before the three return statements.


brian-brazil:
Can we use a defer here to this is less likely to come up in future?

BurtonQin:
The Lock and Unlock are in a loop.

brian-brazil:
We can wrap it in an anonymous function.

BurtonQin:
I think it is very hard to apply the anonymous function trick. See the following code:
```
// inside func sendAll
for _, ams := range amSets {
    ...
    let res := func() bool {  // Try to add the anonymous func
        ams.mtx.RLock()
        defer ams.mtx.RUnlock()
        switch ams.cfg.APIVersion {
           // return false on Error
        }
        for _, am := range ams.ams {  // The for loop is also included since mtx originally protects it 
            wg.Add(1)
            ctx, cancel := context.WithTimeout(n.ctx, time.Duration(ams.cfg.Timeout))
            defer cancel() // cancel() is called when the anonymous func returns.
            // But originally cancel() is called after wg.Wait() in sendAll() !
            go func(...) {
                // wg.Done()
            }(...)
         }
    }()
    if !res {
        return false
    }
}  // End of outer for
wg.Wait()
return numSuccess > 0  
```
We cannot simply wrap the code between Lock and Unlock into an anonymous function. Because defer cancel() in the inner for loop `for _, am := range ams.ams` is originally called in the sendAll(), but now it is called in the anonymous function! Before wg.Wait()!

brian-brazil:
Okay, I guess this will do for now then. Let me see if retrying the CI helps.

bwplotka:
Thanks :tada:

PR #7135, opened 2020-04-16
TSDB: Isolation: avoid creating appenderId's without appender

roidelapluie:
Prior to this commit we could have situations where we are creating an
appenderId but never creating an appender to go with it, therefore
blocking the low watermak.

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -803,8 +803,6 @@ func (h *RangeHead) Meta() BlockMeta {
 type initAppender struct {
 	app  storage.Appender
 	head *Head
-
-	appendID, cleanupAppendIDsBelow uint64
 }
 
 func (a *initAppender) Add(lset labels.Labels, t int64, v float64) (uint64, error) {
@@ -812,7 +810,10 @@ func (a *initAppender) Add(lset labels.Labels, t int64, v float64) (uint64, erro
 		return a.app.Add(lset, t, v)
 	}
 	a.head.initTime(t)
-	a.app = a.head.appender(a.appendID, a.cleanupAppendIDsBelow)
+	appendID := a.head.iso.newAppendID()
+	cleanupAppendIDsBelow := a.head.iso.lowWatermark()
+
+	a.app = a.head.appender(appendID, cleanupAppendIDsBelow)
 
 	return a.app.Add(lset, t, v)
 }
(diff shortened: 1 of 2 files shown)


roidelapluie:
Potential fix for #7120 

cc @beorn7 @codesome @brian-brazil

roidelapluie:
I have added a second commit where we only create appender when needed but which has an impact on the tests. Let me know if you prefer the approach of the first commit or of the second commit.

tsdb/head.go
```diff
@@ -867,8 +860,8 @@ func (h *Head) appender(appendID, cleanupAppendIDsBelow uint64) *headAppender {
 		maxt:                  math.MinInt64,
 		samples:               h.getAppendBuffer(),
 		sampleSeries:          h.getSeriesBuffer(),
-		appendID:              appendID,
-		cleanupAppendIDsBelow: cleanupAppendIDsBelow,
+		appendID:              h.iso.newAppendID(),
+		cleanupAppendIDsBelow: h.iso.lowWatermark(),
```
roidelapluie, line 864: Note: go will go execute that in correct order: https://play.golang.org/p/tjCrTI77UwZ

brian-brazil: Is that guaranteed by the spec?

tsdb/head_test.go
```diff
@@ -1368,14 +1368,16 @@ func TestMemSeriesIsolation(t *testing.T) {
...
+	i := 1
 	for ; i <= 1000; i++ {
 		var app storage.Appender
 		// To initialize bounds.
 		if hb.MinTime() == math.MaxInt64 {
-			app = &initAppender{head: hb, appendID: uint64(i), cleanupAppendIDsBelow: 0}
+			app = &initAppender{head: hb}
 		} else {
-			app = hb.appender(uint64(i), 0)
+			_app := hb.appender()
```
codesome, line 1378: Haven't come across such variable names often in Go. Not a blocker, but can we avoid `_` prefix? Maybe just `a`? :P

codesome:
> I have added a second commit where we only create appenderID when needed but which has an impact on the tests. Let me know if you prefer the approach of the first commit or of the second commit.

I prefer the current version; as it's closely tied, why not keep it together.

tsdb/head.go
roidelapluie: I don't find that info, and for maps and slices it is not guaranteed. https://www.facebook.com/ I have asked golang-nuts.

roidelapluie: In the meantime I have made the execution order explicit.

tsdb/head_test.go
roidelapluie: done

tsdb/head_test.go
```diff
@@ -1368,14 +1368,16 @@ func TestMemSeriesIsolation(t *testing.T) {
...
+	i := 1
 	for ; i <= 1000; i++ {
 		var app storage.Appender
 		// To initialize bounds.
 		if hb.MinTime() == math.MaxInt64 {
-			app = &initAppender{head: hb, appendID: uint64(i), cleanupAppendIDsBelow: 0}
+			app = &initAppender{head: hb}
 		} else {
-			app = hb.appender(uint64(i), 0)
+			a := hb.appender()
```
brian-brazil, line 1378: Why do you need this temporary variable?

roidelapluie: because init appender and appender both meet the storage.Appender interface, but I need to change manually the cleanupAppendIDsBelow which is not part of the interface.

beorn7:
Reviewing now… but already congrats for this catch!

PR #7136, opened 2020-04-17
Ensure queries are closed in API calls

beorn7:
As discussed in #7129.

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -317,7 +317,7 @@ func (api *API) options(r *http.Request) apiFuncResult {
 	return apiFuncResult{nil, nil, nil, nil}
 }
 
-func (api *API) query(r *http.Request) apiFuncResult {
+func (api *API) query(r *http.Request) (result apiFuncResult) {
 	ts, err := parseTimeParam(r, "time", api.now())
 	if err != nil {
 		return apiFuncResult{nil, &apiError{errorBadData, err}, nil, nil}
@@ -340,6 +340,14 @@ func (api *API) query(r *http.Request) apiFuncResult {
 		err = errors.Wrapf(err, "invalid parameter 'query'")
 		return apiFuncResult{nil, &apiError{errorBadData, err}, nil, nil}
 	}
+	// From now on, we must only return with a finalizer in the result (to
+	// be called by the caller) or call qry.Close ourselves (which is
+	// required in the case of a panic).
+	defer func() {
+		if result.finalizer == nil {
+			qry.Close()
+		}
+	}()
 
 	ctx = httputil.ContextFromRequest(ctx, r)
 
@@ -361,7 +369,7 @@ func (api *API) query(r *http.Request) apiFuncResult {
 	}, nil, res.Warnings, qry.Close}
 }
 
-func (api *API) queryRange(r *http.Request) apiFuncResult {
+func (api *API) queryRange(r *http.Request) (result apiFuncResult) {
 	start, err := parseTime(r.FormValue("start"))
 	if err != nil {
 		err = errors.Wrapf(err, "invalid parameter 'start'")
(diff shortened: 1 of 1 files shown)


beorn7:
One of the many fixes that might help #7120 .

brian-brazil:
I was thinking we might be able to do a defer that only Closed if the named return is missing.

beorn7:
That might work as well. I'll give it a try.

beorn7:
Done. Definitely much nicer.

roidelapluie:
Can you also fix the endpoints which use defer in place of finalizer?

beorn7:
But they don't need fixing, do they?

roidelapluie:
public irc logs
```
20:43:25 <@bbrazil> hmm, series() is using defer - that may cause memory corruption
20:43:41 <@bbrazil> we need to keep the strings referenced until the api call is complete
20:44:00 <@bbrazil> well, a memory exposure to be more precise
20:44:48 < roidelapluie> series, labels, label_values..
20:49:22 <@bbrazil> not sure how I missed those, because I know I checked when making those changes
20:49:47 <@bbrazil> maybe they are safe, but the tsdb internals will only get more mmap-y so may aswell make them future proof now
```

roidelapluie:
> But they don't need fixing, do they?

Yeah we can analyze that deeper later.

beorn7:
I see. I read that discussion, but didn't connect it to _this_ `series` call. I'll leave that to the TSDB experts. (o:

brian-brazil:
:+1: 

Yeah, `series` needs the same pattern.

beorn7:
Then let me add that quickly.

web/api/v1/api.go
```diff
@@ -549,15 +549,25 @@ func (api *API) series(r *http.Request) apiFuncResult {
...
+	}
 
 	var sets []storage.SeriesSet
 	var warnings storage.Warnings
 	for _, mset := range matcherSets {
 		s, wrn, err := q.Select(false, nil, mset...)
 		warnings = append(warnings, wrn...)
 		if err != nil {
-			return apiFuncResult{nil, &apiError{errorExec, err}, warnings, nil}
+			return apiFuncResult{nil, &apiError{errorExec, err}, warnings, closer}
```
roidelapluie, line 570:
```suggestion
			return apiFuncResult{nil, &apiError{errorExec, err}, warnings, q.Close}
```

web/api/v1/api.go
```diff
@@ -568,10 +578,10 @@ func (api *API) series(r *http.Request) apiFuncResult {
 		metrics = append(metrics, set.At().Labels())
 	}
 	if set.Err() != nil {
-		return apiFuncResult{nil, &apiError{errorExec, set.Err()}, warnings, nil}
+		return apiFuncResult{nil, &apiError{errorExec, set.Err()}, warnings, closer}
```
roidelapluie, line 581:
```suggestion
		return apiFuncResult{nil, &apiError{errorExec, set.Err()}, warnings, q.Close}
```

web/api/v1/api.go
```diff
@@ -568,10 +578,10 @@ func (api *API) series(r *http.Request) apiFuncResult {
...
 	}
 
-	return apiFuncResult{metrics, nil, warnings, nil}
+	return apiFuncResult{metrics, nil, warnings, closer}
```
roidelapluie, line 584:
```suggestion
	return apiFuncResult{metrics, nil, warnings, q.Close}
```

beorn7: Computer says no.

Or more precisely: “cannot use q.Close (type func() error) as type func() in field value”

roidelapluie: ooh ok.

roidelapluie:
LGTM

brian-brazil:
:+1: 

Thanks

PR #7138, opened 2020-04-17
Fix updating rule manager never finishing

csmarchbanks:
Rather than sending a value to the done channel on a group to indicate
whether or not to add stale markers to a closing rule group use an
explicit boolean. This allows more functions than just run() to read
from the done channel and fixes an issue where Eval() could consume the
channel during an update, causing run() to never return.

cc: @roidelapluie @brian-brazil

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -232,7 +232,8 @@ type Group struct {
 
 	shouldRestore bool
 
-	done        chan bool
+	markStale   bool
+	done        chan struct{}
 	terminated  chan struct{}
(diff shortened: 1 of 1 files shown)


roidelapluie:
g.done doesn't have to be a chan bool anymore

csmarchbanks:
> g.done doesn't have to be a chan bool anymore

Good catch, fixed!

brian-brazil:
:+1:

roidelapluie:
Thanks!

PR #7141, opened 2020-04-18
A bit more explict detail about comparison operators and alert templates

axlrosen:
@brian-brazil

From the discussion here: https://groups.google.com/forum/#!topic/prometheus-users/-aC42iqeTgo

The Prometheus comparison operators work differently from most comparison opeators
in the world. It makes total sense that they work this way (filtering instead of
resulting in a boolean value), but it can be confusing for newbies. This is why
I thought that $value in the template would always be true! So this is just a few
sentences to be more explicit about how they work.

### DIFF
--- a/docs/configuration/alerting_rules.md
+++ b/docs/configuration/alerting_rules.md
@@ -79,6 +79,10 @@ groups:
       description: "{{ $labels.instance }} has a median request latency above 1s (current value: {{ $value }}s)"
 ```
 
+Remember that [comparison operators](https://prometheus.io/docs/prometheus/latest/querying/operators/#comparison-binary-operators)
+are filters by default, so `$value` for the expression `api_http_request_latencies_second{quantile="0.5"} > 1` will be a number,
+not a boolean.
+
 ### Inspecting alerts during runtime
 
 To manually inspect which alerts are active (pending or firing), navigate to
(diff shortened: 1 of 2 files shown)


docs/configuration/alerting_rules.md
```diff
@@ -79,6 +79,10 @@ groups:
       description: "{{ $labels.instance }} has a median request latency above 1s (current value: {{ $value }}s)"
 ```
 
+Remember that [comparison operators](https://prometheus.io/docs/prometheus/latest/querying/operators/#comparison-binary-operators)
```
brian-brazil, line 82: This page is about alerting rules, not binary operators, so this doesn't belong here.

docs/querying/operators.md
```diff
@@ -52,8 +52,9 @@ The following binary comparison operators exist in Prometheus:
 * `<=` (less-or-equal)
 
 Comparison operators are defined between scalar/scalar, vector/scalar,
-and vector/vector value pairs. By default they filter. Their behavior can be
-modified by providing `bool` after the operator, which will return `0` or `1`
+and vector/vector value pairs. By default they filter: for example, the value of
```
brian-brazil, line 55: This is duplicating the information right below this, and confusing as it's only mention one of the three sets of semantics.

roidelapluie:
Based on @brian-brazil 's comments, we have decided to close this pull request.

Issue #7128, opened 2020-04-15
RuleGroups.Validate(...) needs unexported struct ruleGroups as input


sylr:
I'm trying to upgrade prometheus dep in thanos and I'm stuck with this:

```shell
$ make build
all modules verified
>> building binaries /Users/s.rabot/go/bin
 >   thanos
# github.com/thanos-io/thanos/cmd/thanos
cmd/thanos/check.go:88:29: not enough arguments in call to promRgs.Validate
	have ()
	want (rulefmt.ruleGroups)
```

https://github.com/prometheus/prometheus/blob/51c824543bca2f25baa6596827cf4fad8f18dc2e/pkg/rulefmt/rulefmt.go#L66

cc @bwplotka

bwplotka:
Why you put issue on Prometheus for this? It's normal we will have build errors to fix.  We plan to upgrade deps after chunk iterator.

sylr:
Because, unless I'm mistaken, I thought there was little that can be done in thanos if prometheus does not export `ruleGroups` or adds a `NewruleGroups() ruleGroups` func.

Am I wrong ?

aclowkey:
I think it should be exported for non-thanos related reasons.
For example if I want to build a tool that manages alerts, and not implement my own validation logic, I can't.

brian-brazil:
Use of our internals by 3rd parties is unsupported, but we will generally accept PRs to make things public.

bwplotka:
Let's move forward with this then... (: 

It's blocking https://github.com/thanos-io/thanos/pull/2535

bwplotka:
Are you still happy to continue this @sylr

sylr:
@bwplotka I wish I could but I am pretty busy at the moment.

roidelapluie:
Actually the way forward is to get back to yaml.v2 as there are missing features in yaml.v3

bwplotka:
Do you have any details / related discussion @roidelapluie ?

sylr:
https://github.com/prometheus/prometheus/pull/7142#discussion_r410899455

roidelapluie:
Actually things are moving now.. I will see that later on.

roidelapluie:
cmd/thanos/check.go > I do not see that file in thanos codebase?

bwplotka:
It just changed - it is now in cmd/thanos/tools.go

PR #7142, opened 2020-04-19
Shift to yaml v3 and fix duplicate errors in Promtool

### LINKED ISSUE #7042: `promtool check rules` no longer identifies invalid fields as of 2.16.0
**What did you do?**
We use `promtool check rules` with Prometheus 2.16.0 in our CI to validate our Prometheus alerts. We recently had a misconfiguration where an annotation had incorrect indentation, making it an illegal field of the rule rather than an annotation.

Promtool should through an error `field foo not found in type rulefmt.Rule` as it does in Prometheus 2.15.2

Promtool succeeded.

**Repro steps**

```sh
➜  cat << EOF | docker run --entrypoint promtool -i prom/prometheus:v2.15.2 check rules /dev/stdin
groups:
- name: group
  rules:
  - alert: alert
    expr: expr
    labels:
      key: value
    foo: bar
EOF
  FAILED:
/dev/stdin: yaml: unmarshal errors:
  line 8: field foo not found in type rulefmt.Rule
Checking /dev/stdin

➜  cat << EOF | docker run --entrypoint promtool -i prom/prometheus:v2.16.0 check rules /dev/stdin
groups:
- name: group
  rules:
  - alert: alert
    expr: expr
    labels:
      key: value
    foo: bar
EOF
Checking /dev/stdin
  SUCCESS: 1 rules found
```

Harkishen-Singh:
Fixes: #7042 
Fixes: #7128 

Improvements:
1. (Fixed by another PR as well) The reported issue of labels parsing error being skipped is fixed. Hence,
```
groups:
  - name: group
    rules:
      - alert: alert
        expr: expr
        labels:
          key : value
        foo: bar
```

will result to

```
[hsingh@localhost prometheus]$ ./promtool check rules rule.yml 
Checking rule.yml
  FAILED:
rule.yml: yaml: unmarshal errors:
  line 8: field foo not found in type rulefmt.Rule
```

2. Duplicate errors in the previous version related to yaml unmarshalling is fixed.
3. Shift to yaml v3 library across the application.


pkg/rulefmt/rulefmt.go
```diff
@@ -21,6 +21,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
+	yamlv2 "gopkg.in/yaml.v2"
```
brian-brazil, line 24: Why do we need two versions?

Harkishen-Singh: We need strict unmarshalling checks which is only in yaml.v2 lib. Yamlv3 was required for line and columns support. Hence, I don't think we have any other way than using the features from both lib. Later if v3 gets strict unmarshalling, then this could be confined to a single lib. Other ways might be a hack and not recommended.

What do you think?

brian-brazil: I wasn't aware that this had regressed, which would be a bug anywhere we use v3 then. We either need to get it added to v3, return to v2, or use the approach we used before v2 had this.

Harkishen-Singh: well, in that case, I will open an issue in yaml v3. Sticking to this will be good I think since if we switch over to v2, then features like column and line numbers would be lost.

brian-brazil: Do you know how this slipped past our unittests?

Harkishen-Singh: Our unittests does not cover the criteria where an unknown field is present. Will update that in this PR as the yaml v3 is fixed.

Link of the issue in go-yaml v3: https://github.com/go-yaml/yaml/issues/602

pkg/rulefmt/rulefmt.go
```diff
@@ -50,6 +51,7 @@ func (err *Error) Error() string {
 	} else if err.Err.node != nil {
 		return errors.Wrapf(err.Err.err, "%d:%d: group %q, rule %d, %q", err.Err.node.Line, err.Err.node.Column, err.Group, err.Rule, err.RuleName).Error()
 	}
+
```
roidelapluie, line 54: why this?

roidelapluie:
@brian-brazil What do you think about this?

brian-brazil:
I still think that we should not be importing two different YAML library versions, and should instead be fixing our usage of the existing library.

pkg/rulefmt/rulefmt.go
Harkishen-Singh: You mean the trailing line?

pkg/rulefmt/rulefmt.go
```diff
@@ -58,10 +60,50 @@ type RuleGroups struct {
...
+	Record      string            `yaml:"record,omitempty"`
+	Alert       string            `yaml:"alert,omitempty"`
+	Expr        string            `yaml:"expr"`
+	For         model.Duration    `yaml:"for,omitempty"`
+	Labels      map[string]string `yaml:"labels,omitempty"`
+	Annotations map[string]string `yaml:"annotations,omitempty"`
+}
+
 // Validate validates all rules in the rule groups.
 func (g *RuleGroups) Validate(node ruleGroups) (errs []error) {
```
gotjosh, line 108: Now that we're here, can we also fix this function? It is a public function but it is expecting a private type.

gotjosh: Which would fix https://github.com/prometheus/prometheus/issues/7128

pkg/rulefmt/rulefmt.go
```diff
@@ -58,10 +60,50 @@ type RuleGroups struct {
...
+	Labels      map[string]string `yaml:"labels,omitempty"`
+	Annotations map[string]string `yaml:"annotations,omitempty"`
+}
+
+// ruleGroups type of yaml.v3 features in the groups field.
 type ruleGroups struct {
 	Groups []yaml.Node `yaml:"groups"`
 }
 
+// validateRules type for validating the groups inside the
```
gotjosh, line 85: What's the rationale behind the public and private types here? I see that our public types are using `yaml.Node` which currently seems to have rendering problem per https://github.com/go-yaml/yaml/issues/575

gotjosh:
I agree with @brian-brazil seems like the addition of yaml.v3 has introduced a number of regressions.

pkg/rulefmt/rulefmt.go
Harkishen-Singh: Sure. I will update this to public. however, the merge might take a good amount of time due to lack of sufficient features in yaml v3.

Harkishen-Singh:
@brian-brazil I think we have some after this [comment](https://github.com/go-yaml/yaml/issues/602#issuecomment-623485602).

brian-brazil:
Let's update our usage then to fix the regression.

Harkishen-Singh:
@brian-brazil we can move ahead in this now. Thank you for the suggestions @roidelapluie.

pkg/rulefmt/rulefmt.go
Harkishen-Singh: Done. The function is public now.

Harkishen-Singh:
@roidelapluie implemented the changes from #7208. Please have a look.

pkg/rulefmt/rulefmt.go
```diff
@@ -14,15 +14,15 @@
 package rulefmt
 
 import (
+	"bytes"
 	"context"
+	"gopkg.in/yaml.v3"
```
roidelapluie, line 19: this is misplaced

pkg/rulefmt/rulefmt.go
```diff
@@ -50,20 +50,51 @@ func (err *Error) Error() string {
...
+	Record      yaml.Node         `yaml:"record,omitempty"`
+	Alert       yaml.Node         `yaml:"alert,omitempty"`
+	Expr        yaml.Node         `yaml:"expr"`
+	For         model.Duration    `yaml:"for,omitempty"`
+	Labels      map[string]string `yaml:"labels,omitempty"`
+	Annotations map[string]string `yaml:"annotations,omitempty"`
 }
 
+// ruleGroups type for yaml.v3 features in the groups field.
 type ruleGroups struct {
```
roidelapluie, line 81: move that below Rule

pkg/rulefmt/rulefmt.go
```diff
@@ -50,20 +50,51 @@ func (err *Error) Error() string {
...
+	Annotations map[string]string `yaml:"annotations,omitempty"`
+}
+
 // Validate validates all rules in the rule groups.
-func (g *RuleGroups) Validate(node ruleGroups) (errs []error) {
+func (g *RuleGroups) Validate() (errs []error) {
+	node := g.nodes
 	set := map[string]struct{}{}
 
 	for j, g := range g.Groups {
```
roidelapluie, line 100: g is shadowing g

pkg/rulefmt/rulefmt.go
```diff
@@ -50,20 +50,51 @@ func (err *Error) Error() string {
...
+	Expr        string            `yaml:"expr"`
+	For         model.Duration    `yaml:"for,omitempty"`
+	Labels      map[string]string `yaml:"labels,omitempty"`
+	Annotations map[string]string `yaml:"annotations,omitempty"`
+}
+
 // Validate validates all rules in the rule groups.
-func (g *RuleGroups) Validate(node ruleGroups) (errs []error) {
+func (g *RuleGroups) Validate() (errs []error) {
+	node := g.nodes
```
roidelapluie, line 97: that seems unnecessary

Harkishen-Singh:
@roidelapluie done!

rules/manager_test.go
```diff
@@ -754,7 +754,7 @@ func TestUpdate(t *testing.T) {
 	defer tmpFile.Close()
 
 	err = ruleManager.Update(10*time.Second, []string{tmpFile.Name()}, nil)
-	testutil.Ok(t, err)
+	testutil.NotOk(t, err)
```
Harkishen-Singh, line 757: `NotOk` since enabling the `KnownFiled(true)` will create an error for empty file due to `EOF`. Hence, here an error should occur. `ioutil.TempFile` will give an empty file and hence an error is expected.

Harkishen-Singh:
@brian-brazil @roidelapluie should we move ahead.

pkg/rulefmt/rulefmt.go
```diff
@@ -14,14 +14,15 @@
...
 import (
+	"bytes"
 	"context"
 	"io/ioutil"
 	"strings"
 	"time"
 
 	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
-	yaml "gopkg.in/yaml.v3"
```
brian-brazil, line 24: Leave this as-is, it makes it more obvious what it is being imported as.

pkg/rulefmt/rulefmt.go
```diff
@@ -50,37 +51,67 @@ func (err *Error) Error() string {
...
-func (g *RuleGroups) Validate(node ruleGroups) (errs []error) {
+func (g *RuleGroups) Validate() (errs []error) {
 	set := map[string]struct{}{}
 
-	for j, g := range g.Groups {
-		if g.Name == "" {
-			errs = append(errs, errors.Errorf("%d:%d: Groupname should not be empty", node.Groups[j].Line, node.Groups[j].Column))
+	for j, grp := range g.Groups {
+		if grp.Name == "" {
+			errs = append(errs, errors.Errorf("%d:%d: Groupname should not be empty", g.nodes.Groups[j].Line, g.nodes.Groups[j].Column))
```
brian-brazil, line 102: hmm, shouldn't the error be lower case? "group name" would also be clearer.

pkg/rulefmt/rulefmt.go
```diff
@@ -50,37 +51,67 @@ func (err *Error) Error() string {
...
+}
+
+// RuleGroup is a list of sequentially evaluated recording and alerting rules.
+type RuleGroup struct {
+	Name     string         `yaml:"name"`
+	Interval model.Duration `yaml:"interval,omitempty"`
+	Rules    []RuleNode     `yaml:"rules"`
 }
 
+// RuleNode adds yaml.v3 layer to support line and column outputs for invalid rules.
```
brian-brazil, line 71: We don't need to mention yamlv3 here, describe what RuleNode represents.

Why do we need both RuleNode and Rule?

Harkishen-Singh: `RuleNode` for yaml.v3 features but yes, it is not required to be public. Will change that to private since it is only used in the rulefmt.go.

pkg/rulefmt/rulefmt.go
Harkishen-Singh: sure

pkg/rulefmt/rulefmt.go
Harkishen-Singh: Sure.

pkg/rulefmt/rulefmt.go
Harkishen-Singh: Sure. I have removed `Rule` and converted the `RuleNode` to `Rule` having the yaml properties from the v3 library. Everything should work now with just the Rule only.

pkg/rulefmt/rulefmt.go
brian-brazil: This seems to have gotten reverted

roidelapluie:
Tests are failing

pkg/rulefmt/rulefmt.go
Harkishen-Singh: Fixed it in the recent commits.

Harkishen-Singh:
@roidelapluie fixed! No idea on the failing ones like windows and the fuzzing. Seems like system issues.

Harkishen-Singh:
@brian-brazil @roidelapluie cc

Harkishen-Singh:
@brian-brazil can we move ahead?

rules/manager_test.go
```diff
@@ -819,7 +829,10 @@ func formatRules(r *rulefmt.RuleGroups) ruleGroupsTest {
 }
 
 func reloadAndValidate(rgs *rulefmt.RuleGroups, t *testing.T, tmpFile *os.File, ruleManager *Manager, expected map[string]labels.Labels, ogs map[string]*Group) {
+	// grps := formatRules(rgs)
```
brian-brazil, line 832: Debug code

rules/recording.go
```diff
@@ -188,7 +188,11 @@ func (rule *RecordingRule) HTMLSnippet(pathPrefix string) template.HTML {
 		labels[l.Name] = template.HTMLEscapeString(l.Value)
 	}
 
-	r := rulefmt.Rule{
+	r := struct {
+		Record string            `yaml:"record"`
```
brian-brazil, line 192: Why can't we continue to use rulefmt.Rule here?

Harkishen-Singh: we want to avoid the yaml.Node here else the output will be a node structure. The node structure is like [this](https://godoc.org/gopkg.in/yaml.v3#Node) which is of course, of no use to us for the htmlsnippet. So for this case where output is string specific and there is no work of debugging details like line/column numbers, we choose to go with normal strings. If we don't do go this way, then the output is like 

```
record:
            kind: 0
            style: 0
            tag: ""
            value: <a href="/test/prefix/graph?g0.expr=testrule&g0.tab=1">testrule</a>
            anchor: ""
            alias: null
            content: []
            headcomment: ""
            linecomment: ""
            footcomment: ""
            line: 0
            column: 0
```

which is of no use here (only).

rules/manager_test.go
Harkishen-Singh: done!

Harkishen-Singh:
@brian-brazil cc

rules/manager_test.go
brian-brazil: And the lines after it.

rules/recording.go
brian-brazil: Hmm, so how will we do this once we switch fully to v3?

Harkishen-Singh: Absolutely not an issue. This format comes only when `yaml.Node` is needed. This was the requirement only for checking the rules and their validation with more control on the details. As far as I see, there seems to be no more use of `yaml.Node` (since non of our components require validation and detailed error in line/column or any other functionalities related to Node), that means a `string` type will work. **`string` type do not have this while unmarshalling**. Only the `yaml.Node` have these details when they are unmarshalled. 

I can submit another PR after this (since this one has a different aim) that converts and unifies the yaml v3 library across Prometheus main repo.

rules/manager_test.go
Harkishen-Singh: Ahh. Removed them too.

Harkishen-Singh:
@brian-brazil I think we can go ahead with the merge if everything seems fine to us.

rules/recording.go
brian-brazil: So after the followup PR, this anonymous type would be gone?

Harkishen-Singh: Since both anonymous structs are same, we can rather declare in rulefmt as non yaml node and import from there. Thats what the following will have for this thing.

brian-brazil: That still doesn't sound right, there shouldn't be anonymous types at all. Otherwise we'll have to duplicate the entire of the config structure just to display it on the UI.

Harkishen-Singh: How about putting these into a defined structs with string as some types, since node is not required in this case?

bwplotka: hm but we already have `rulefmt.Rule{` what's wrong with that? ( I might miss context).

brian-brazil: My point is that we shouldn't need to duplicate this type at all. We should be able to marshal one type, not have to duplicate types if we want to marshal anything.

bwplotka:
Essentially our worry is that we have sooo many Rules structs. For API, For internal use, For yaml parsing and for UI. Now for yaml parsing we have 2 which is crazy.

In comparison in Thanos we have just one: which is capable to parse to many different things (including protobuf itself), all done from proto: https://github.com/thanos-io/thanos/blob/714ac05388615da8a37b1cc50d916cd90509ac1f/pkg/rules/rulespb/rpc.proto#L67

This means we can do much better here in Prometheus.

Harkishen-Singh:
@bwplotka the point is that if we unmarshal any struct that contains a `yaml.Node`, we get additional details like `linecomment`, `headcomment`, etc, which looks messy and unnecessary for the purpose of showing in the view. Thats the reason we need to have at least 2 structs, First, for features for promtool, Second, for views.

rules/recording.go
Harkishen-Singh: I think, two types are required, else might look messy, keeping in mind, that we require some features from `v3` and also a marshalled output without `yaml.Node` details.

Harkishen-Singh:
I can convert the anonymous types to a defined one, but that would be along with `rulefmt.Rule`, like `rulefmt.RuleTempl` or something around that. If that does not sound good, please give some suggestions on how to have within a single struct. @brian-brazil @bwplotka

brian-brazil:
> Thats the reason we need to have at least 2 structs, First, for features for promtool, Second, for views.

If yaml v3 can't do such marshalling without duplicating all of our config structs, we either need to learn how to do it, get it added upstream, or abandon v3.

Harkishen-Singh:
But referencing nodes as pointers, that might be possible, but I think that will also involve few changes in other files as well.

roidelapluie:
> > Thats the reason we need to have at least 2 structs, First, for features for promtool, Second, for views.
> 
> If yaml v3 can't do such marshalling without duplicating all of our config structs, we either need to learn how to do it, get it added upstream, or abandon v3.

It can but in this case it is needed to get the exact line / position in the file.

brian-brazil:
> It can do without duplicate but in this case duplicate is needed to get the exact line / position in the file.

My point is that we'd ultimately want to be on v3 everywhere, and if we need to duplicate the entirety of prometheus.yml just to show it on the status page then I question if v3 is worth it. This seems like a usage/feature issue with v3 which we should resolve.

rules/manager_test.go
roidelapluie: Not sure about this. Would break users.

roidelapluie:
Couldn't this pull request be limited to:

```
diff --git a/pkg/rulefmt/rulefmt.go b/pkg/rulefmt/rulefmt.go
index 1d6716beb..804c05352 100644
--- a/pkg/rulefmt/rulefmt.go
+++ b/pkg/rulefmt/rulefmt.go
@@ -14,6 +14,7 @@
 package rulefmt
 
 import (
+       "bytes"
        "context"
        "io/ioutil"
        "strings"
@@ -267,7 +268,9 @@ func Parse(content []byte) (*RuleGroups, []error) {
                errs   []error
        )
 
-       err := yaml.Unmarshal(content, &groups)
+       decoder := yaml.NewDecoder(bytes.NewReader(content))
+       decoder.KnownFields(true)
+       err := decoder.Decode(&groups)
        if err != nil {
                errs = append(errs, err)
        }
```

Harkishen-Singh:
> Couldn't this pull request be limited to:

I think, that might fail the tests that involve HTML outputs. That's the reason why the anonymous structs are used.

Harkishen-Singh:
Sorry for the delay on this. I was involved in #7779. Now that it seems almost on track, I will focus on things required to get this done.

Harkishen-Singh:
@brian-brazil @bwplotka @gotjosh @roidelapluie  I have reduced and made this work with lesser structs. That seemed like the most important issue. With this, I think we can have another round of review.

pkg/rulefmt/rulefmt.go
```diff
@@ -91,8 +105,8 @@ func (g *RuleGroups) Validate(node ruleGroups) (errs []error) {
 					ruleName = r.Record
 				}
 				errs = append(errs, &Error{
-					Group:    g.Name,
-					Rule:     i + 1,
+					Group:    grp.Name.Value,
+					Rule:     i,
```
brian-brazil, line 109: This has regressed

rules/manager_test.go
```diff
@@ -789,18 +789,27 @@ type ruleGroupsTest struct {
...
 // ruleGroupTest forms a testing struct for running tests over rules.
 type ruleGroupTest struct {
-	Name     string         `yaml:"name"`
+	Name     yaml.Node      `yaml:"name"`
 	Interval model.Duration `yaml:"interval,omitempty"`
-	Rules    []rulefmt.Rule `yaml:"rules"`
+	Rules    []rule         `yaml:"rules"`
+}
+
+type rule struct {
```
brian-brazil, line 797: Why do you need these copies of the existing structs?

rules/recording.go
brian-brazil: This comment stands.

Harkishen-Singh: Yes, we can use this now.

rules/alerting.go
```diff
@@ -534,7 +538,13 @@ func (r *AlertingRule) HTMLSnippet(pathPrefix string) html_template.HTML {
...
 	}
 
-	ar := rulefmt.Rule{
+	ar := struct {
+		Alert       string
+		Expr        string
+		For         model.Duration    `yaml:"for,omitempty"`
+		Labels      map[string]string `yaml:"labels,omitempty"`
+		Annotations map[string]string `yaml:"annotations,omitempty"`
+	}{
```
Harkishen-Singh, line 547: Infact, over here too, `rulefmt.Rule{}` can be used as it was earlier.

rules/manager_test.go
Harkishen-Singh: This is now back to `testutil.Ok`. :grinning:

rules/alerting.go
Harkishen-Singh: Addressed since now this is using `rulefmt.Rule`.

rules/manager_test.go
Harkishen-Singh: Addressed this.

rules/recording.go
Harkishen-Singh: I think this can be resolved now.

rules/manager_test.go
Harkishen-Singh: Resolving this since its now addressed.

Harkishen-Singh:
@brian-brazil addressed the comments!

rules/manager_test.go
```diff
@@ -784,42 +784,35 @@ func TestUpdate(t *testing.T) {
...
-
-type rule struct {
-	Record      string            `yaml:"record,omitempty"`
-	Alert       string            `yaml:"alert,omitempty"`
-	Expr        string            `yaml:"expr"`
-	For         model.Duration    `yaml:"for,omitempty"`
-	Labels      map[string]string `yaml:"labels,omitempty"`
-	Annotations map[string]string `yaml:"annotations,omitempty"`
-}
+//type ruleGroupTest struct {
```
brian-brazil, line 791: This is left over

Harkishen-Singh: Ahh, will remove in the following commit soon. BTW, can you say why is the `test` failing?

brian-brazil: No idea offhand.

Harkishen-Singh: Got it, it was at `t.Fail()`. Now, fixed it.

brian-brazil:
Tests still aren't happy, but only breaking on Windows for some odd reason for something that isn't platform specific.

brian-brazil:
There should be nothing platform specific in any of this code, so any issues like that need to be investigated rather than removing key tests.

Harkishen-Singh:
> There should be nothing platform specific in any of this code, so any issues like that need to be investigated rather than removing key tests.

It is platform dependent. the fail is at `testdata/duplicate_grp.bad.yaml` where windows expects it as `testdata\duplicate_grp.bad.yaml` and linux as `testdata/duplicate_grp.bad.yaml`. Any suggestions in this case? 

@brian-brazil

Harkishen-Singh:
I guess, simply removing the path will work since we are comparing `strings.Contains(errs[0], errMsg)`

Harkishen-Singh:
@brian-brazil done!

pkg/rulefmt/rulefmt_test.go
```diff
@@ -84,6 +89,8 @@ func TestParseFileFailure(t *testing.T) {
 			continue
 		}
 		if !strings.Contains(errs[0].Error(), c.errMsg) {
+			fmt.Println("at zero is :", errs[0].Error())
```
brian-brazil, line 92: Debug code.

brian-brazil (commented):
Can you update the commit description to what this PR is doing now?

Harkishen-Singh:
> Can you update the commit description to what this PR is doing now?

I guess, only the duplicate errors thing. But, since this has been quite sometime around, I am updating this PR to shift from yaml.v2 to yaml.v3 completely in reliable terms.

Harkishen-Singh:
> Can you update the commit description to what this PR is doing now?

Done!

Harkishen-Singh:
Can someone please help me fix the failing test due to `go mod`?

Harkishen-Singh:
@brian-brazil updated!

roidelapluie:
@abursavich You have played with yamlv3, would you want to help us reviewing this? :) Thanks :)

config/testdata/conf.good.yml
```diff
@@ -73,7 +73,7 @@ scrape_configs:
     target_label:  cde
   - replacement:   static
     target_label:  abc
-  - regex:
+  - regex: ""
```
brian-brazil, line 76: Why did this change? It's important that the previous version also is parsed, as we've had panics from things like that in the past.

Harkishen-Singh: It seems like the default value of regex when nothing is mentioned is `(.*)`. That's why I had to apply an empty string.

brian-brazil: That sounds like a regression then, I'd expect this to be parsed as the empty string as that's what's in the YAML.

Harkishen-Singh: It is dependent on regex lib most probably, since I have seen yaml v3 parsing empty as someone does expect.

I see one way as a workaround. We can make `Regex` a string and a new private member `reg` that contains the actual reg expr which will be used. This gives us the conditional control on regex and an help solve the regression.

Harkishen-Singh: Well, the thing is that if we leave `regex` field empty, then after parsing, it gets a value of `null`. But, we use `DefaultConfig` where it is initialized to `(.*)` which leads to failing.

brian-brazil: That's a regression that needs fixing then.

Don't silently change unittest behaviour in PRs like this, this would have been a breaking change.

Harkishen-Singh: Should I just keep yaml package of the particular func to v2? Since, rest all places the shift to v3 seems a success.

brian-brazil: We should be switching everything to v3, having bits here and there will only cause confusion long term.

Harkishen-Singh: Then I will be happy if I can get some suggestions to solve the issue related to this particular case only(i.e., `Load()` in `config.go`). Maybe, `help wanted` tag if you don't have time for this.

abursavich:
> @abursavich You have played with yamlv3, would you want to help us reviewing this? :) Thanks :)

Yeah, sure. I'll take a look at this today, but as recently as few weeks ago I would've said that v3 isn't ready yet.

abursavich:
When I was working on the config stuff, I upgraded everything to v3 thinking it might make things better, but it only created more problems. Sorry to show up out of nowhere and be a party pooper, but nothing has changed in the meantime. There are too many serious issues with v3 and there's a reason why a v3.0.0 release hasn't been tagged yet.

roidelapluie:
> Sorry to show up out of nowhere and be a party pooper, but nothing has changed in the meantime

Well I did ask for your opinion, so you do not pop up from nowhere. Thanks for your message, and indeed:

- v2 is still default branch of the upstream repository
- v3 has a list of issues in the bug tracker

Note that we had a bug in the rules (which use v3) that was there in 2.16 to 2.20 and then fixed in 2.21, because we did not introduce strict marshalling. I'd be very careful about using more v3.

Harkishen-Singh:
Ahh, then I think it will be good to keep this PR open and wait for some stable release of YAML v3. Since there's only one place left, it will be soon to implement the changes as soon as a stable release of YAML v3 is out.

config/testdata/conf.good.yml
dgl: I opened https://github.com/go-yaml/yaml/pull/689 on the go-yaml side for this, I updated a branch of this PR locally and aside from this issue I don't see any major blockers now.

Issue #7124, opened 2020-04-14
Log flooding with "skipping resharding" warning (remote_write with write_relabel_configs)


Brightside56:
**What did you do?**
Installed prometheus which scrapes some targets and writes metrics with label `longterm=true` (with the help of `write_relabel_configs`) to remote (InfluxDB v1.7.10). Prometheus is up and running for 5 minutes.

**What did you expect to see?**
Samples of metrics with label `longterm=true` in remote storage and crystal clean prometheus log without errors/warnings.

**What did you see instead? Under which circumstances?**
Remote write is working and samples I need are writing into remote storage, however prometheus is constantly flooding to log with `Skipping resharding, last successful send was beyond threshold`. I believe it's not normal for production use. When I disable `write_relabel_configs` and prometheus starts to write all metrics to remote - I don't see warnings anymore, however writing all metrics to remote is not my case)

**Environment**

* System information:

	Prometheus is run in Docker container prom/prometeus:v2.17.1 (on Kubernetes 1.11) with Linux 4.19.88 x86_64 

* Prometheus version:

```
prometheus, version 2.17.1 (branch: HEAD, revision: ae041f97cfc6f43494bed65ec4ea4e3a0cf2ac69)
  build user:       root@806b02dfe114
  build date:       20200326-16:18:19
  go version:       go1.13.9
```

I've also tried prometheus docker image with version v2.15.1 which I have on another cluster. Result is same.

* Prometheus configuration file:
```
global:
  scrape_interval: 15s
  scrape_timeout: 15s
  evaluation_interval: 15s
scrape_configs:
- job_name: prometheus
  static_configs:
    - targets: ['localhost:9090']
- job_name: 'backup'
  static_configs:
    - targets: ['xx.xx.xx.xx:9100']
  metric_relabel_configs:
  - source_labels: [__name__]
    regex: job_name
    target_label: longterm
    replacement: "true"
    action: replace
remote_write:
- url: "http://influxdb:8086/api/v1/prom/write?db=prometheus"
  write_relabel_configs:
  - source_labels: [longterm]
    regex: "true"
    action: keep
```
Also I've tried to use VictoriaMetrics v1.34.7 as remote storage. Result is same. I believe it's Prometheus issue and it's not remote storage related. When I disable `write_relabel_configs` there are no warnings anymore, but I don't want to write all metrics from Prometheus to remote, only part of them.

* Logs (with debug enabled):
```
[...]

PR #7143, opened 2020-04-19
Adding a shouldReshard function to modularize logic for the QueueManager deciding if it should shard or not

thaniri:
This is a fix for https://github.com/prometheus/prometheus/issues/7124

The function calculateDesiredShards() should only do what it says. Calculate the number of desired shards.

From the description of the function:

```
// calculateDesiredShards returns the number of desired shards, which will be
// the current QueueManager.numShards if resharding should not occur for reasons
// outlined in this functions implementation. It is up to the caller to reshard, or not,
// based on the return value.
```

- Moved around some logic on deciding whether or not the QueueManager should reshard from updateShardsLoop() and calculateDesiredShards() into a new function called shouldReshard(). Mostly it's just copy pasting if statements from the first two functions into the new function.
- Renamed TestCalculateDesiredsShards to TestShouldReshard to more accurately reflect the purpose of that test.

With these changes we move the logic to decide whether or not to reshard to a higher level in the code.

### DIFF
--- a/storage/remote/queue_manager_test.go
+++ b/storage/remote/queue_manager_test.go
@@ -292,16 +292,14 @@ func TestReleaseNoninternedString(t *testing.T) {
 	testutil.Assert(t, metric == 0, "expected there to be no calls to release for strings that were not already interned: %d", int(metric))
 }
 
-func TestCalculateDesiredsShards(t *testing.T) {
+func TestShouldReshard(t *testing.T) {
 	type testcase struct {
 		startingShards        int
-		samplesIn, samplesOut int64
 		reshard               bool
+		samplesIn, samplesOut int64
 	}
 	cases := []testcase{
 		{
-			// Test that ensures that if we haven't successfully sent a
-			// sample recently the queue will not reshard.
 			startingShards: 10,
 			reshard:        false,
 			samplesIn:      1000,
@@ -329,8 +327,9 @@ func TestCalculateDesiredsShards(t *testing.T) {
 		}
 		m.Start()
 		desiredShards := m.calculateDesiredShards()
+		shouldReshard := m.shouldReshard(desiredShards)
 		m.Stop()
-		if !c.reshard {
+		if !(c.reshard == shouldReshard) {
 			testutil.Assert(t, desiredShards == m.numShards, "expected calculateDesiredShards to not want to reshard, wants to change from %d to %d shards", m.numShards, desiredShards)
 		} else {
 			testutil.Assert(t, desiredShards != m.numShards, "expected calculateDesiredShards to want to reshard, wants to change from %d to %d shards", m.numShards, desiredShards)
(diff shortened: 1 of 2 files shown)


storage/remote/queue_manager_test.go
```diff
@@ -329,8 +327,9 @@ func TestCalculateDesiredsShards(t *testing.T) {
 		}
 		m.Start()
 		desiredShards := m.calculateDesiredShards()
+		shouldReshard := m.shouldReshard(desiredShards)
 		m.Stop()
-		if !c.reshard {
+		if !(c.reshard == shouldReshard) {
```
csmarchbanks, line 332: I don't think we need to test the specific cases down below anymore as this test is targed at `shouldReshard`, one single `testutil.AssertEqual(t, c.reshard, shouldReshard)` should cover it.

storage/remote/queue_manager_test.go
```diff
@@ -321,20 +322,16 @@ func TestCalculateDesiredsShards(t *testing.T) {
...
+		shouldReshard := m.shouldReshard(desiredShards)
+
 		m.Stop()
-		if !c.reshard {
-			testutil.Assert(t, desiredShards == m.numShards, "expected calculateDesiredShards to not want to reshard, wants to change from %d to %d shards", m.numShards, desiredShards)
-		} else {
-			testutil.Assert(t, desiredShards != m.numShards, "expected calculateDesiredShards to want to reshard, wants to change from %d to %d shards", m.numShards, desiredShards)
-		}
+
+		testutil.Assert(t, shouldReshard == c.expectedToReshard, "shouldReshard set to %t, expected %t", shouldReshard, c.expectedToReshard)
```
csmarchbanks, line 334: `testutil.Equals(t, c.expectedToReshard, shouldReshard)` rather than `Assert` will handle the print formatting for you in a consistent manner with other test functions.

thaniri:
Yes csmarchbanks, that is a perfectly reasonable request.

I have added a new commit to this PR.

Do you want me to squash my commits or will they be squashed when merged?

storage/remote/queue_manager_test.go
thaniri: This is now a single assertion.

csmarchbanks: Yep, thanks, the logic is much easier to follow than before!

Since you are now comparing two values usually we use `testutil.Equals(t, expected, actual)` instead of `testutil.Assert(t, expected == actual, "my message")` as the former provides some nice out of the box formatting.

csmarchbanks:
The commits will be squashed when they are merged so no worries about squashing now. Thanks again!

storage/remote/queue_manager_test.go
thaniri: Done, thanks for the suggestion!

Sorry for the weird reply to the initial comment, you had made a different comment to simplify the two test assertions in the old file, which I changed and then did a rebase with force push.

It looks like Github didn't display the old comment any more (due to the force push) and you made this comment on exactly the same line.

I went on autopilot mode and just replied to this one as though it was the old one.

I'll be more careful in the future.

Issue #6972, opened 2020-03-12
Variance of letter case on logs


gotjosh:
Hello Prometheans,

Unsure whenever this is best suited as a bug report, but I've observed different letter casing in log lines and was wondering whenever this is indented.

I _think_ there might be a pattern but would like to confirm. It seems that for components starting and stopping (e.g. [scrape manager](https://github.com/prometheus/prometheus/blob/master/cmd/prometheus/main.go#L571), [remote write](https://github.com/prometheus/prometheus/blob/master/storage/remote/queue_manager.go#L398), [tsdb](https://github.com/prometheus/prometheus/blob/master/cmd/prometheus/main.go#L667), [Prometheus itself](https://github.com/prometheus/prometheus/blob/master/cmd/prometheus/main.go#L333)) log lines are capitalized. 

For everything else that appears to be operational (e.g. [scrape manager](https://github.com/prometheus/prometheus/blob/master/scrape/manager.go)), it seems to be lower case. 

However, sometimes even this pattern is broken (e.g. [remote write](https://github.com/prometheus/prometheus/blob/master/storage/remote/queue_manager.go#L538)).

Do we have any consensus on when to use which? I couldn't find any documentation that indicates that, and I'd be happy to fix/create anything that might need it.

brian-brazil:
Log messages should be capitalised, however for go errors should not be as they'll likely be prefixed. That can sometimes result in an uncapitalised log message. If there's cases that are easy to fix, we should fix them.

thaniri:
I can work on this.

brian-brazil:
Create a fork on github and push it to your fork.

gotjosh:
ping me if you need any help @thaniri, I think a first step you can do (mostly for the benefit on getting familiar with the different Prometheus areas) is to identify which are the different areas we have to look for the log lines (e.g. storage, tsdb, scrape, service discovery). 

Then, make a checklist as part of this issue so that we can keep track of which areas we have scanned already. Then, either you (or others) can each take up on the different areas and submit the PRs for them.

I think this makes it easier for everything, you can keep track of what you're doing and the reviewers only need to look at a particular area at a time.

thaniri:
It is my personal opinion that creating a different PR for each package within the repository will result in such a trivial change taking too much time as too many people will need to get involved in this change. Really this PR could have probably been done with a sed command and some review :)

Here is the PR ready for review. I am hoping that I have followed the contributing guidelines correctly.

[...]

PR #7148, opened 2020-04-20
Add Jaeger

vsliouniaev:
Hoping I understand https://github.com/prometheus/prometheus/issues/4509 correctly to expose the traces to Jaeger.

Looking for guidance on best way to proceed here around setting up configuration sections.

---

Also propagate the trace when performing remote read.


config/config.go
```diff
@@ -138,6 +153,9 @@ type Config struct {
 	RemoteWriteConfigs []*RemoteWriteConfig `yaml:"remote_write,omitempty"`
 	RemoteReadConfigs  []*RemoteReadConfig  `yaml:"remote_read,omitempty"`
 
+	// experimental tracing configuration section
```
brian-brazil, line 156: We shouldn't be adding anything to the config file. Adding Jaeger is temporary, and not a part of the semantics of the Prometheus.

roidelapluie: So you want it completely controlled by env variables?

brian-brazil: I never said that, how does it work out of the box?

vsliouniaev: It can be done completely with env variables out of the box. If that's the better approach I'm happy to take the config bit out

brian-brazil: That sounds okay, as long as we're never touching them ourselves.

roidelapluie:
There is no documentation in this PR.

brian-brazil:
I believe that this is mainly for the use of Prometheus developers ourselves, and will not be a long term feature, so having no docs is fine.

vsliouniaev:
The reason I'm interested in adding this is to trace queries from grafana through prometheus and into fanout storage. In the configuration where this is useful for us, prometheus is just one portion of the call graph and it's currently missing tracing.

brian-brazil:
https://github.com/prometheus/prometheus/issues/4509#issuecomment-578412331 is the current concensus on this. Adding Jaeger is only a stopgap until the dust settles around OpenTelemetry.

config/config.go
vsliouniaev: Now configuration is sourced entirely from environment variables and everything is off by default.

cmd/prometheus/main.go
```diff
@@ -498,6 +504,13 @@ func main() {
 		})
 	}
 
+	closer, err := initTracing(logger)
+	if err != nil {
+		level.Error(logger).Log("err", err)
```
brian-brazil, line 509: This should have a msg too

cmd/prometheus/main.go
```diff
@@ -1005,3 +1018,52 @@ func (opts tsdbOptions) ToTSDBOptions() tsdb.Options {
 		MaxBlockDuration:       int64(time.Duration(opts.MaxBlockDuration) / time.Millisecond),
 	}
 }
+
+func initTracing(logger log.Logger) (io.Closer, error) {
+	// Set tracing configuration defaults
```
brian-brazil, line 1023: full stop

cmd/prometheus/main.go
```diff
@@ -1005,3 +1018,52 @@ func (opts tsdbOptions) ToTSDBOptions() tsdb.Options {
...
+		Sampler: &jaegercfg.SamplerConfig{
+			Type:  jaeger.SamplerTypeConst,
+			Param: 1,
+		},
+		Reporter: &jaegercfg.ReporterConfig{
+			LogSpans: false,
+		},
+	}
+
+	// Override defaults from environment variables. Available options can be seen here:
```
brian-brazil, line 1036: I'm not sure about this mix of hardcoded and environment. Are there not sensible defaults?

roidelapluie: The comment is wrong

roidelapluie: You need at least to be able to enable it, set your jaeger server, your extra tags, etc

vsliouniaev: - The `ServiceName` is required.
- Tracing is enabled by default, I think that should definitely be flipped around.
- Without a sampler config, nothing is captured, so I think it's a sensible default for a feature targeted at developers.
- `Reporter` can be left out.

With this
```go
jaegercfg.Configuration{
  ServiceName: "prometheus",
  Disabled:    true,
}
```
the env vars to set are:
```bash
JAEGER_DISABLED=false 
JAEGER_SAMPLER_TYPE=const 
JAEGER_SAMPLER_PARAM=1 
JAEGER_ENDPOINT=http://jaeger.local
```

cmd/prometheus/main.go
vsliouniaev: added

cmd/prometheus/main.go
vsliouniaev: added

cmd/prometheus/main.go
brian-brazil: That seems reasonably minimal.

vsliouniaev: Updated to use minimum config in code

cmd/prometheus/main.go
```diff
@@ -17,6 +17,10 @@ package main
 import (
 	"context"
 	"fmt"
+	"github.com/opentracing/opentracing-go"
+	"github.com/uber/jaeger-lib/metrics"
```
roidelapluie, line 21: this not our usual order

cmd/prometheus/main.go
```diff
@@ -498,6 +503,13 @@ func main() {
 		})
 	}
 
+	closer, err := initTracing(logger)
+	if err != nil {
+		level.Error(logger).Log("msg", "Unable to init tracing", "err", err)
+		os.Exit(1)
```
roidelapluie, line 509: exit 2 is what we do for such kind of errors

storage/remote/client.go
```diff
@@ -18,6 +18,8 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
```
roidelapluie, line 22: this is not our usual order

roidelapluie:
Would it be possible to add the span id to the query log?

around here:
https://github.com/prometheus/prometheus/blob/4b5e7d4984a135abd3729b9a2a5904336d35bc21/promql/engine.go#L441

storage/remote/client.go
vsliouniaev: Should be correct now

cmd/prometheus/main.go
vsliouniaev: Should be correct now

cmd/prometheus/main.go
vsliouniaev: Updated

cmd/prometheus/main.go
```diff
@@ -17,6 +17,8 @@ package main
 import (
 	"context"
 	"fmt"
+
```
roidelapluie, line 20: extra space

cmd/prometheus/main.go
```diff
@@ -1005,3 +1017,45 @@ func (opts tsdbOptions) ToTSDBOptions() tsdb.Options {
...
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to get tracing config from environment")
+	}
+
+	jLogger := jaegerLogger{logger: log.With(logger, "component", "tracing")}
+	jMetricsFactory := metrics.NullFactory
+
+	tracer, closer, err := cfg.NewTracer(
+		jaegercfg.Logger(jLogger),
+		jaegercfg.Metrics(jMetricsFactory),
```
roidelapluie, line 1039: can't we have prometheus metrics?

cmd/prometheus/main.go
```diff
@@ -1005,3 +1016,45 @@ func (opts tsdbOptions) ToTSDBOptions() tsdb.Options {
...
 }
+
+func initTracing(logger log.Logger) (io.Closer, error) {
+	// Set tracing configuration defaults.
+	cfg := &jaegercfg.Configuration{
+		ServiceName: "prometheus",
+		Disabled:    true,
+	}
+
+	// Override defaults from environment variables. Available options can be seen here:
```
brian-brazil, line 1027: Override isn't the right word, I'd remove that first sentance.

promql/engine.go
```diff
@@ -444,6 +444,7 @@ func (ng *Engine) exec(ctx context.Context, q *query) (v parser.Value, w storage
 					f = append(f, k, v)
 				}
 			}
+			f = append(f, "span", opentracing.SpanFromContext(ctx))
```
roidelapluie, line 447: I think we only want the span id

roidelapluie: and only if tracing is enabled (e.g. there is a span)

brian-brazil:
:+1:

promql/engine.go
vsliouniaev: Updated. I found some discussion around the opentracing api itself not facilitate getting this info, but jaeger does.

cmd/prometheus/main.go
vsliouniaev: Removed

cmd/prometheus/main.go
vsliouniaev: Added

cmd/prometheus/main.go
vsliouniaev: Space removed

brian-brazil:
I think your vendoring might be off.

roidelapluie:
I will test this with an actual jaeger setup tonight, then we can merge. Thanks.

promql/engine.go
roidelapluie: Yes, we needed to wait for jaeger to get the information.

roidelapluie:
It works but I am not sure that every HTTP request should have its own operation name (especially 404 and static files):

GET
GET2

roidelapluie:
I would like to know your opinion on this @vsliouniaev , if that makes sense or not and if we can improve that in a further PR.

cmd/prometheus/main.go
```diff
@@ -1005,3 +1016,44 @@ func (opts tsdbOptions) ToTSDBOptions() tsdb.Options {
...
+
+func initTracing(logger log.Logger) (io.Closer, error) {
+	// Set tracing configuration defaults.
+	cfg := &jcfg.Configuration{
+		ServiceName: "prometheus",
+		Disabled:    true,
+	}
+
+	// Available options can be seen here:
+	// https://github.com/jaegertracing/jaeger-client-go/blob/master/config/config_env.go
```
roidelapluie, line 1028:
```suggestion
	// https://github.com/jaegertracing/jaeger-client-go#environment-variables
```

cmd/prometheus/main.go
```diff
@@ -1005,3 +1016,44 @@ func (opts tsdbOptions) ToTSDBOptions() tsdb.Options {
...
+	cfg, err := cfg.FromEnv()
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to get tracing config from environment")
```
roidelapluie, line 1031: https://github.com/prometheus/prometheus/issues/6972#issuecomment-598137385 -> please capitalize

cmd/prometheus/main.go
```diff
@@ -1005,3 +1016,44 @@ func (opts tsdbOptions) ToTSDBOptions() tsdb.Options {
...
+	}
+
+	jLogger := jaegerLogger{logger: log.With(logger, "component", "tracing")}
+
+	tracer, closer, err := cfg.NewTracer(
+		jcfg.Logger(jLogger),
+		jcfg.Metrics(jprom.New()),
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to init tracing")
```
roidelapluie, line 1041: https://github.com/prometheus/prometheus/issues/6972#issuecomment-598137385 -> please capitalize

roidelapluie: oops actually no

cmd/prometheus/main.go
roidelapluie: no, this an err not a message. sorry

roidelapluie:
Just une URL to fix in comments and we are good to merge :)

vsliouniaev:
> It works but I am not sure that every HTTP request should have its own operation name (especially 404 and static files):

I agree, but I've only been looking at Jaeger for about a week, so not certain of best practices at all. I can take a look at how to improve the current behaviour

roidelapluie:
That code has not changed since 4d9b917d11 (2017)

vsliouniaev:
> I am not sure that every HTTP request should have its own operation name

The http method is already included as a tag, so that can be dropped.

There are certain routes that I think would be useful to drop traces for entirely because they don't add much value, for example
```
/new
/static
```

an argument could be made as well for 
```
/-/health
/-/ready
```

The `/api` route gets somewhat complicated because of parts of the path are parameters, but the default route set-up doesn't seem to have a neat way of extracting that information that I have found. There are other routers available that are able to automatically handle this, but I don't think that much of a change is really warranted in this case.

The minimum I think that we should do is
- don't include the path prefix if it's specified
- don't include the http method from the operation
- remove the full path for static files to `/static`

Ideally I would like to create generic operations for calls that look like this `/api/v1/labels/job/....` to something that looks like `/api/v1/labels/:name/`, but I don't want to have an explicit list of mappings between the route setup and their operation representations

storage/remote/client.go
```diff
@@ -149,6 +151,17 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
 	httpReq.Header.Set("User-Agent", userAgent)
 	httpReq.Header.Set("X-Prometheus-Remote-Read-Version", "0.1.0")
 
+	if span := opentracing.SpanFromContext(ctx); span != nil {
```
yurishkuro, line 154: nit: the recommended way to trace remote calls is to create a new child span here, and capture in its tag various useful info like the query.

PR #7157, opened 2020-04-23
tsdb: capture total number of label pairs in head

### LINKED ISSUE #7147: Expose total number of label pairs in head
## Proposal

For scale analysis it would be great if we either exposed a metric that reflects the total number of labels (name + value combinations) in the head block. Potentially this could also be an entry in the head block cardinality section on the status page. I don't care how I just want to get the numbers out somehow and can't today.

Side note, yes top 10 cardinality labels are useful, but may be hiding a long tail of labels that still affect the head's index, but we can't be sure until we are able to validate with numbers.

Overall context: I'm researching some other indexing strategies and it's crucial to have some real world numbers for that.

@brian-brazil @roidelapluie @bwplotka @csmarchbanks @simonpasquier

khyatisoneji:
Resolves #7147

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -101,6 +101,7 @@ type headMetrics struct {
 	checkpointDeleteTotal   prometheus.Counter
 	checkpointCreationFail  prometheus.Counter
 	checkpointCreationTotal prometheus.Counter
+	symbols                 prometheus.Histogram
 }
 
 func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
@@ -179,6 +180,10 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 			Name: "prometheus_tsdb_checkpoint_creations_total",
 			Help: "Total number of checkpoint creations attempted.",
 		}),
+		symbols: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Name: "prometheus_tsdb_head_symbols_total",
+			Help: "Total number of symbols in the head",
+		}),
 	}
 
 	if r != nil {
@@ -201,6 +206,7 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 			m.checkpointDeleteTotal,
 			m.checkpointCreationFail,
 			m.checkpointCreationTotal,
+			m.symbols,
 			// Metrics bound to functions and not needed in tests
 			// can be created and registered on the spot.
 			prometheus.NewGaugeFunc(prometheus.GaugeOpts{
@@ -1177,6 +1183,7 @@ func (h *Head) gc() {
 
 	h.symbols = symbols
 	h.values = values
+	h.metrics.symbols.Observe(float64(len(h.symbols)))
 
 	h.symMtx.Unlock()
 }
@@ -1494,6 +1501,7 @@ func (h *Head) getOrCreateWithID(id, hash uint64, lset labels.Labels) (*memSerie
 		h.symbols[l.Name] = struct{}{}
 		h.symbols[l.Value] = struct{}{}
 	}
+	h.metrics.symbols.Observe(float64(len(h.symbols)))
 
 	return s, true
 }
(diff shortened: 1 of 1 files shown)


khyatisoneji:
Thank you very much for your inputs @codesome 
I have a doubt regarding the metric to be used for histogram, since the number of values would vary based on the label. Should we only use one metric for measuring the number of values in all the labels? Or we will need to have different metrics for each label?

khyatisoneji:
One more question is around what should we name the Gauge metric for `h.values`? Earlier I had used `prometheus_tsdb_head_symbols_total`, but that shouldn't be used. What would be a more appropriate metric name? Sorry for the trouble

brian-brazil:
I'd go for labelpairs, symbols are something different.

khyatisoneji:
@brian-brazil can you please also check my other doubt https://github.com/prometheus/prometheus/pull/7157#issuecomment-618323899

brian-brazil:
This should be a gauge with no labels.

khyatisoneji:
Can you please check if current changes look fine @brian-brazil @codesome . Thanks for your help

tsdb/head.go
```diff
@@ -1167,6 +1179,7 @@ func (h *Head) gc() {
 			values[t.Name] = ss
 		}
 		ss.set(t.Value)
+		h.metrics.labelpairValues.Set(float64(ss.Size()))
```
brian-brazil, line 1182: This isn't right, this is the number of label values for the last label touched.

tsdb/head.go
```diff
@@ -1177,6 +1190,7 @@ func (h *Head) gc() {
 
 	h.symbols = symbols
 	h.values = values
+	h.metrics.labelpairs.Set(float64(len(h.values)))
```
brian-brazil, line 1193: And this would be the number of label names.

You'd need to iterate over h.values to count up the number of unique label pairs.

khyatisoneji:
@brian-brazil Can you please inform what would be the values if `h.values` is the following:
```
"a": {"1": struct{}{}, "2": struct{}{}},
"b": {"1": struct{}{}},
"c":  {"3": struct{}{}},
```

Questions: 
1. `prometheus_tsdb_head_labelpairs_total` (total number of label pairs) would be 3. Am I correct? Can I use `reflect.DeepEqual` to check the unique count? I believe that is the only way to compare `map[string]struct{}`

2. What would be the value for `value and name length`? 
Thank you very much for your inputs.

brian-brazil:
There are 4 labelpairs in that example. Value length would also be 4 here, as all the values are only one char.

khyatisoneji:
I didn't understand what you meant by one char, can you give me one example where the number of labelpairs and value length is different. I was looking at the tests to get more understanding, but there is only one example of `h.values` which is the one that I posted above. Thank you for your continued help.

tsdb/head.go
```diff
@@ -179,6 +181,14 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 			Name: "prometheus_tsdb_checkpoint_creations_total",
 			Help: "Total number of checkpoint creations attempted.",
 		}),
+		labelPairs: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "prometheus_tsdb_head_label_pairs_total",
+			Help: "Total number of label pairs in the head",
+		}),
+		labelValues: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "prometheus_tsdb_head_label_values_total",
+			Help: "Total number of label values in the head",
```
brian-brazil, line 190: This is the size of the label values, not the number of label values (which would be the same as the number of label pairs).

tsdb/head.go
```diff
@@ -1177,7 +1202,7 @@ func (h *Head) gc() {
 
 	h.symbols = symbols
 	h.values = values
-
+	h.metrics.collectLabelMetrics(h.values)
```
brian-brazil, line 1205: Iterating over ever every label name each time a series is created seems like it may cause performance problems.

roidelapluie:
Hey @khyatisoneji , can you please address the comments in this PR? Thanks!

khyatisoneji:
Hello @roidelapluie I saw there was another PR: https://github.com/prometheus/prometheus/pull/7381 addressing the issue, thus I haven't completed the changes

roidelapluie:
Thanks for you reply, I will therefore close this issue.

PR #7158, opened 2020-04-23
Updated all deps.

bwplotka:
Pinned github.com/googleapis/gnostic as they introduced the braking change.

(step 0 for release: of https://github.com/prometheus/prometheus/blob/master/RELEASE.md#0-updating-dependencies)


brian-brazil:
Some tests need updating too.

juliusv:
@bwplotka LOL. Ok, I'm starting to lean more and more in favor of getting rid of vendoring :laughing: 

Looks good besides the testing breakage.

brian-brazil:
I'm not seeing any issues here arising from vendoring, just the usual stuff when dependencies are updated and contain breaking changes.

bwplotka:
IMO **Vendoring is really bad and should be avoided for all cost**

But at least I will bump my code line changes stats! :tada: 

Let's discuss further this at some point, what is the benefit of vendoring and how we can enable those benefits without committing code we don't own to our repo. Email thread was already started: https://groups.google.com/d/msgid/prometheus-developers/CAMssQwYMRzrpwWAZ4xGAzLp7jDdn1rHC%2B8Tb9vGuz6%2Bcrj4U%2BQ%40mail.gmail.com

Let's add this to DevSummit discussions as well

juliusv:
@brian-brazil Yeah, no concrete new problem, just the general unwieldiness of it.

bwplotka:
uuuups

> vendor\k8s.io\apimachinery\pkg\util\errors\errors.go:99:10: undefined: errors.Is

Hm

brancz:
Not having vendering makes offline build a pita. We (as in Red Hat) for one would be feeling this pretty harshly. I don’t see it as much of a burden either except github stats 🤷‍♂️

juliusv:
> Not having vendering makes offline build a pita.

True, though we can't build Prometheus offline anymore anyway because `node_modules` is also not checked in.

bwplotka:
You can totally cache those things though.

bwplotka:
>  I don’t see it as much of a burden either except github stats 

It's far more serious than this @brancz . For example, I am now fixing various tests in Prometheus because e.g we have some flaky tests https://circleci.com/gh/prometheus/prometheus/30352?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link plus some tests were affected because of K8s discovery client change.

Because you cannot even load the review page because it's too huge there is no way you can review any of this.

We are literally changing code without review now.

roidelapluie:
I don't think we can update kube dependency before circleci has updated go to >= 1.13 on windows

bwplotka:
After updating (!) orbs, we have lover Go version :facepalm: 

Makefile.common:54: This repository requires Go >= 1.11 because of Go modules
Makefile.common:55: Some recipes may not work as expected as the current Go runtime is 'go1.9.1'

bwplotka:
At this point I have zero idea how those orbs works... @simonpasquier any help here?

bwplotka:
I just changed orb windows version, used only by windows. Somehow all are red now because of old Go.. This makes zero sense ):

bwplotka:
Asked @simonpasquier for help, if we can find the reason now, I will just skip updating deps and will move with release without this PR

bwplotka:
Looks like it was some magic cache on Circle CI side.

Anyway, giving up with this for now. Trying without k8s.io/client: https://github.com/prometheus/prometheus/pull/7162

PR #7159, opened 2020-04-23
Callbacks for lifecycle of series in TSDB

codesome:
This PR implements callbacks in TSDB for series lifecycle based on proposal in https://groups.google.com/forum/#!topic/prometheus-developers/08F1z6RdkH4

/cc @gouthamve @pracucci @pstibrany @bwplotka @brian-brazil

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -65,7 +65,8 @@ type Head struct {
 	memChunkPool sync.Pool
 
 	// All series addressable by their ID or hash.
-	series *stripeSeries
+	series         *stripeSeries
+	seriesCallback SeriesLifecycleCallback
 
 	symMtx  sync.RWMutex
 	symbols map[string]struct{}
@@ -284,12 +285,15 @@ func (h *Head) PostingsCardinalityStats(statsByLabelName string) *index.Postings
 // stripeSize sets the number of entries in the hash map, it must be a power of 2.
 // A larger stripeSize will allocate more memory up-front, but will increase performance when handling a large number of series.
 // A smaller stripeSize reduces the memory allocated, but can decrease performance with large number of series.
-func NewHead(r prometheus.Registerer, l log.Logger, wal *wal.WAL, chunkRange int64, chkDirRoot string, pool chunkenc.Pool, stripeSize int) (*Head, error) {
+func NewHead(r prometheus.Registerer, l log.Logger, wal *wal.WAL, chunkRange int64, chkDirRoot string, pool chunkenc.Pool, stripeSize int, seriesCallback SeriesLifecycleCallback) (*Head, error) {
 	if l == nil {
 		l = log.NewNopLogger()
 	}
 	if chunkRange < 1 {
 		return nil, errors.Errorf("invalid chunk range %d", chunkRange)
+	}
+	if seriesCallback == nil {
+		seriesCallback = &noopSeriesLifecycleCallback{}
 	}
 	h := &Head{
 		wal:        wal,
@@ -297,7 +301,7 @@ func NewHead(r prometheus.Registerer, l log.Logger, wal *wal.WAL, chunkRange int
 		chunkRange: chunkRange,
 		minTime:    math.MaxInt64,
 		maxTime:    math.MinInt64,
-		series:     newStripeSeries(stripeSize),
+		series:     newStripeSeries(stripeSize, seriesCallback),
 		values:     map[string]stringset{},
 		symbols:    map[string]struct{}{},
 		postings:   index.NewUnorderedMemPostings(),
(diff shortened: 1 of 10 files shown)


tsdb/head.go
```diff
@@ -466,11 +471,16 @@ func (h *Head) loadWAL(r *wal.Reader, multiRef map[uint64]uint64) (err error) {
...
 
+Outer:
 	for d := range decoded {
 		switch v := d.(type) {
 		case []record.RefSeries:
 			for _, s := range v {
-				series, created := h.getOrCreateWithID(s.Ref, s.Labels.Hash(), s.Labels)
+				series, created, err := h.getOrCreateWithID(s.Ref, s.Labels.Hash(), s.Labels)
+				if err != nil {
+					errCh <- err
```
codesome, line 481: @brian-brazil the channel size was set to 2 because we could add error to channel even here in addition to the decoding channel. Probably this is not a good design to do so and a variable should be enough. Also draining the `decoded` channel upon this error is missing, will add that.

brian-brazil: Without looking at the code, I'd expect there to be 2 channels in that case.

codesome: And with a quick look, no need for channels. Just 2 variables are enough :)

codesome:
This PR is now ready for review

tsdb/head.go
```diff
@@ -258,26 +259,30 @@ func (h *Head) PostingsCardinalityStats(statsByLabelName string) *index.Postings
 // stripeSize sets the number of entries in the hash map, it must be a power of 2.
 // A larger stripeSize will allocate more memory up-front, but will increase performance when handling a large number of series.
 // A smaller stripeSize reduces the memory allocated, but can decrease performance with large number of series.
-func NewHead(r prometheus.Registerer, l log.Logger, wal *wal.WAL, chunkRange int64, stripeSize int) (*Head, error) {
+func NewHead(r prometheus.Registerer, l log.Logger, wal *wal.WAL, chunkRange int64, stripeSize int, seriesLifecycleCallback SeriesLifecycleCallback) (*Head, error) {
```
bwplotka, line 262:
```suggestion
func NewHead(r prometheus.Registerer, l log.Logger, wal *wal.WAL, chunkRange int64, stripeSize int, callback SeriesLifecycleCallback) (*Head, error) {
```

bwplotka: let's maybe keep it simpler we don't have many callbacks (:

tsdb/head.go
```diff
@@ -258,26 +259,30 @@ func (h *Head) PostingsCardinalityStats(statsByLabelName string) *index.Postings
...
+		minTime:                 math.MaxInt64,
+		maxTime:                 math.MinInt64,
+		series:                  newStripeSeries(stripeSize, seriesLifecycleCallback),
+		values:                  map[string]stringset{},
+		symbols:                 map[string]struct{}{},
+		postings:                index.NewUnorderedMemPostings(),
+		tombstones:              tombstones.NewMemTombstones(),
+		iso:                     newIsolation(),
+		deleted:                 map[uint64]int{},
+		seriesLifecycleCallback: seriesLifecycleCallback,
```
bwplotka, line 285:
```suggestion
		callback: callback,
```

tsdb/head.go
```diff
@@ -2035,3 +2068,21 @@ func (ss stringset) slice() []string {
 	sort.Strings(slice)
 	return slice
 }
+
+// SeriesLifecycleCallback specifies a list of callbacks that will be called during a lifecycle of a series.
+// It is always a no-op in Prometheus and mainly meant for external users who import TSDB.
+type SeriesLifecycleCallback interface {
+	// PreCreationCallback is called before creating a series to indicate if the series can be created.
+	// A non nil error means the series should not be created.
+	PreCreationCallback(labels.Labels) error
```
bwplotka, line 2077: Maybe just without callback? we are already in callback interface. 
```suggestion
	PreCreation(labels.Labels) error
```
And below

tsdb/head.go
```diff
@@ -1648,25 +1673,33 @@ func (s *stripeSeries) getByHash(hash uint64, lset labels.Labels) *memSeries {
 	return series
 }
 
-func (s *stripeSeries) getOrSet(hash uint64, series *memSeries) (*memSeries, bool) {
-	i := hash & uint64(s.size-1)
+func (s *stripeSeries) getOrSet(hash uint64, series *memSeries) (*memSeries, bool, error) {
+	createSeriesErr := s.seriesLifecycleCallback.PreCreationCallback(series.lset)
```
bwplotka, line 1677: This is unexpected to me. I would expect preCreation callback to invoked before creation. That moment is after 

```
if prev := s.hashes[i].get(hash, series.lset); prev != nil {
		s.locks[i].Unlock()
		return prev, false, nil
	}
```
I think... no?

tsdb/head.go
codesome: But keeping it plain `callback` is confusing too, how about `seriesCallback`?

tsdb/head.go
codesome: We don't want to have the callback inside the lock. Hence `PreCreation` is just a hint to TSDB about what to do if it had to create a series and it is a best-effort as we want to avoid callbacks inside lock.

bwplotka: Interesting.... I think this make sense, but deserve a comment.. WDYT? (:

codesome: Definitely

PR #7160, opened 2020-04-23
Unblock goroutines in case of WAL corruptions

codesome:
Currently it looks like if there is a repairable WAL corruption the goroutines from WAL replay are always stuck.


brian-brazil:
Do we still need the defer doing this up near the start of the function?

codesome:
:facepalm: I did not notice that, my bad. We don't need this PR.

PR #7165, opened 2020-04-24
Cut release 2.18.0-rc.0

bwplotka:


CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+## 2.18.0-rc.0 / 2020-04-24
+
+* [CHANGED] Use .UTC timezeone everywhere we use time.Unix. #7066
+* [CHANGED] Federation: It does not use remote storages anymore as designed initially. #7077
+* [CHANGED] Rules: `rule_evaluations_total` and `rule_evaluation_failures_total` have `rule_group` label now. #7094
+* [CHANGED] Moved to Go 1.14.2. #7100
```
roidelapluie, line 6: That is not relevant for end users

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+## 2.18.0-rc.0 / 2020-04-24
+
+* [CHANGED] Use .UTC timezeone everywhere we use time.Unix. #7066
+* [CHANGED] Federation: It does not use remote storages anymore as designed initially. #7077
+* [CHANGED] Rules: `rule_evaluations_total` and `rule_evaluation_failures_total` have `rule_group` label now. #7094
```
roidelapluie, line 5:
```suggestion
* [ENHANCEMENT] Rules: `rule_evaluations_total` and `rule_evaluation_failures_total` have `rule_group` label now. #7094
```

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+## 2.18.0-rc.0 / 2020-04-24
+
+* [CHANGED] Use .UTC timezeone everywhere we use time.Unix. #7066
+* [CHANGED] Federation: It does not use remote storages anymore as designed initially. #7077
```
roidelapluie, line 4:
```suggestion
* [CHANGE] Federation: It does not use remote storages anymore as designed initially. #7077
```

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+## 2.18.0-rc.0 / 2020-04-24
+
+* [CHANGED] Use .UTC timezeone everywhere we use time.Unix. #7066
```
roidelapluie, line 3: Is that relevant for end users?

```suggestion
* [BUGFIX] Use UTC timezone everywhere we use time.Unix. #7066
```

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
...
+* [CHANGED] Federation: It does not use remote storages anymore as designed initially. #7077
+* [CHANGED] Rules: `rule_evaluations_total` and `rule_evaluation_failures_total` have `rule_group` label now. #7094
+* [CHANGED] Moved to Go 1.14.2. #7100
+* [FEATURE] Tracing(!): Added experimental Jaeger support #7148
+* [ENHANCEMENT] TSDB: Significantly reduce how much old WAL we keep around to 3h (previously 6h). #7098
+* [BUGFIX] Federation: Fixed not registered `prometheus_web_federation_errors_total` and `prometheus_web_federation_warnings_total` metrics. #7081
+* [BUGFIX] Discovery: Add `archiecture` meta label for EC2. #7000
```
roidelapluie, line 10:
```suggestion
* [BUGFIX] Discovery: Add `architecture` meta label for EC2. #7000
```

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+## 2.18.0-rc.0 / 2020-04-24
+
+* [CHANGED] Use .UTC timezeone everywhere we use time.Unix. #7066
+* [CHANGED] Federation: It does not use remote storages anymore as designed initially. #7077
+* [CHANGED] Rules: `rule_evaluations_total` and `rule_evaluation_failures_total` have `rule_group` label now. #7094
+* [CHANGED] Moved to Go 1.14.2. #7100
+* [FEATURE] Tracing(!): Added experimental Jaeger support #7148
+* [ENHANCEMENT] TSDB: Significantly reduce how much old WAL we keep around to 3h (previously 6h). #7098
+* [BUGFIX] Federation: Fixed not registered `prometheus_web_federation_errors_total` and `prometheus_web_federation_warnings_total` metrics. #7081
```
roidelapluie, line 9: Was backported to 2.17.2 already

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+## 2.18.0-rc.0 / 2020-04-24
+
+* [CHANGED] Use .UTC timezeone everywhere we use time.Unix. #7066
+* [CHANGED] Federation: It does not use remote storages anymore as designed initially. #7077
+* [CHANGED] Rules: `rule_evaluations_total` and `rule_evaluation_failures_total` have `rule_group` label now. #7094
+* [CHANGED] Moved to Go 1.14.2. #7100
+* [FEATURE] Tracing(!): Added experimental Jaeger support #7148
```
roidelapluie, line 7: This should not be documented as it will not stay for long and is not documented, also it is not relevant for users

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
...
+* [ENHANCEMENT] TSDB: Significantly reduce how much old WAL we keep around to 3h (previously 6h). #7098
```
roidelapluie, line 8: is it needed to specify the hours?

roidelapluie:
Thanks! I have added some comments

CHANGELOG.md
bwplotka: :+1:

CHANGELOG.md
bwplotka: I think actually it  might be the most relevant part of this release (: Except maybe Federation (potentially breaking) change

CHANGELOG.md
bwplotka: No harm, it's useful to get idea

roidelapluie: is the WAL size depending on parameters that can be changed? Can we instead say divided by two? WDYT?

CHANGELOG.md
roidelapluie: It is something mainly aimed at developers, not users, and that we explicitely did not document because we do not want to support it in short or long term. Adding it here would imply that we need documentation otherwise we will get many questions.

WDYT?

Also, cc @brian-brazil on this one.

bwplotka: I see your point, we can mention that documentation will be added at some point once production-ready.

There is no point in adding experimental feature without any mention (: Who will experiment then?

cc @RichiH @juliusv @brancz

CHANGELOG.md
bwplotka: Fair point, Happy with that

CHANGELOG.md
roidelapluie: I have said what I had to say, let's get more input. But it is not experimental, it is stop-gap until we have opentelemetry.

CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+## 2.18.0-rc.0 / 2020-04-24
+
+* [CHANGE] Federation: It does not use remote storages anymore as designed initially. #7077
```
roidelapluie, line 3: Can we use a positive sentence here?

```suggestion
* [CHANGE] Federation: Only use local TSDB for federation (ignore remote read). #7077
```

brian-brazil: "Federation now only uses local TSDB" would be clearer.

CHANGELOG.md
brian-brazil: It should be listed, but given the odd situation I'd downplay it and put it as an enhancement.

Experimental does not mean not production ready, it means we're allowed break it - or in this case completely remove it in future.

CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
...
+* [FEATURE] Tracing(!): Added experimental Jaeger support #7148
+* [ENHANCEMENT] Rules: `rule_evaluations_total` and `rule_evaluation_failures_total` have `rule_group` label now. #7094
+* [ENHANCEMENT] TSDB: Significantly reduce how much old WAL we keep around. #7098
+* [BUGFIX] Discovery: Add `architecture` meta label for EC2. #7000
+* [BUGFIX] React UI: Fixed multiselect legend on OSX. #6880
+* [BUGFIX] Remote Write: Fixed blocked resharding edge case. #7122
+* [BUGFIX] Remote Write: Fixed the edge case of not updating remote write on relabel configs change. #7073
+* [BUGFIX] Use .UTC timezeone everywhere we use time.Unix. #7066
```
brian-brazil, line 11: Technically this isn't even a bugfix, as we weren't doing anything wrong per-se. It's more of a cleanup.

CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+## 2.18.0-rc.0 / 2020-04-24
+
+* [CHANGE] Federation: It does not use remote storages anymore as designed initially. #7077
+* [FEATURE] Tracing(!): Added experimental Jaeger support #7148
+* [ENHANCEMENT] Rules: `rule_evaluations_total` and `rule_evaluation_failures_total` have `rule_group` label now. #7094
+* [ENHANCEMENT] TSDB: Significantly reduce how much old WAL we keep around. #7098
+* [BUGFIX] Discovery: Add `architecture` meta label for EC2. #7000
```
brian-brazil, line 7: Enhancement

CHANGELOG.md
brian-brazil: This is a CHANGE

CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+## 2.18.0-rc.0 / 2020-04-24
+
+* [CHANGE] Federation: Only use local TSDB for federation (ignore remote read). #7077
+* [FEATURE] Tracing(!): Added experimental Jaeger support #7148
+* [ENHANCEMENT] Rules: `rule_evaluations_total` and `rule_evaluation_failures_total` have `rule_group` label now. #7094
```
brian-brazil, line 5: This is a CHANGE, as it'll break any users who were depending on this not having any instrumentation labels previously.

bwplotka: cc @roidelapluie I had `CHANGE` previously (:

CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
...
+* [ENHANCEMENT] TSDB: Significantly reduce how much old WAL we keep around. #7098
+* [BUGFIX] Discovery: Add `architecture` meta label for EC2. #7000
+* [BUGFIX] React UI: Fixed multiselect legend on OSX. #6880
+* [BUGFIX] Remote Write: Fixed blocked resharding edge case. #7122
+* [BUGFIX] Remote Write: Fixed the edge case of not updating remote write on relabel configs change. #7073
```
csmarchbanks, line 11: I don't think saying this is an edge case adds anything. `Fixed remote write not updating on relabel configs change` is what we did.

CHANGELOG.md
csmarchbanks: I think the feature of `Tracing` is here to stay, and should be in the release notes. The fact that it is Jaeger might be short term, but the experimental marker means we can change the implementation at any time. What is the `(!)` meant to convey?

CHANGELOG.md
bwplotka: :+1:

CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
+## 2.18.0-rc.0 / 2020-04-24
+
+* [CHANGE] Federation: Only use local TSDB for federation (ignore remote read). #7077
+* [CHANGE] Rules: `rule_evaluations_total` and `rule_evaluation_failures_total` have `rule_group` label now. #7094
+* [FEATURE] Tracing(!): Added experimental Jaeger support #7148
+* [ENHANCEMENT] TSDB: Significantly reduce how much old WAL we keep around. #7098
+* [ENHANCEMENT] Use .UTC timezeone everywhere we use time.Unix. #7066
```
brian-brazil, line 7: This shouldn't be listed at all, it's a very minor cleanup.

CHANGELOG.md
brian-brazil: Yeah, (!) is not something the changelog should contain.

bwplotka:
Are we good to go? (: @brian-brazil @csmarchbanks @roidelapluie ?

CHANGELOG.md
brian-brazil: The `(!)` is still here.

beorn7: I'd also remove the `(!)`.

> Experimental does not mean not production ready, it means we're allowed break it - or in this case completely remove it in future.

Just FTR: I don't know where @brian-brazil is taking this from. I have certainly marked many features experimental not _only_ to mark it as something that might change later without a major version bump but _also_ to not set any expectation it's production ready. We don't need to discuss the details here. I have added an item to the dev summit agenda already. I mostly want to file my objection here to avoid @brian-brazil to once more use an imaginary past consensus in future arguments.

CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+## 2.18.0-rc.0 / 2020-04-24
+
+* [CHANGE] Federation: Only use local TSDB for federation (ignore remote read). #7077
+* [CHANGE] Rules: `rule_evaluations_total` and `rule_evaluation_failures_total` have `rule_group` label now. #7094
```
beorn7, line 4: I knew changelog lines are very condensed, but inserting an `a` after the `have` is just one letter and reads much better.

CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
...
+* [FEATURE] Tracing(!): Added experimental Jaeger support #7148
+* [ENHANCEMENT] TSDB: Significantly reduce how much old WAL we keep around. #7098
```
beorn7, line 6: The "we" reads weirdly. How about: "TSDB: Significantly reduce WAL size kept around after truncation."

CHANGELOG.md
brian-brazil: I'm working off https://prometheus.io/docs/prometheus/latest/stability/
Our language here has historically not been the colloquial usage, so I want to avoid confusion.

beorn7: That document doesn't state anything about production readiness.

> Our language here has historically not been the colloquial usage, so I want to avoid confusion.

Once more, you are using "we/us/our" when you mean "I/me/my". Or if you really want to speak for all of us, you should at least mark that you are describing your personal impression of what we are doing, which historically has deviated repeatedly from the impression of the rest of us.

bwplotka:
Thanks @beorn7 :hugs: Addressed.

CHANGELOG.md
brian-brazil: > That document doesn't state anything about production readiness.

Exactly, and I don't believe we have any docs that do.

> Once more, you are using "we/us/our" when you mean "I/me/my".

I mean "we" here. We as a group have been sloppy in our usage of the word "experimental", with the prime example being remote write which has been both experimental and production ready for some years. This was never an explicit consensus or decision that I know of, rather a thing that naturally happened over time. It'd be good to discuss how to clean this up at the next dev summit, as this situation has caused some user confusion over the years.

Anyway, in this particular case the words in the changelog entry seem fine.

brian-brazil:
:+1:

CHANGELOG.md
roidelapluie: ack

CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
...
+
+* [CHANGE] Federation: Only use local TSDB for federation (ignore remote read). #7077
+* [CHANGE] Rules: `rule_evaluations_total` and `rule_evaluation_failures_total` have a `rule_group` label now. #7094
+* [FEATURE] Tracing: Added experimental Jaeger support #7148
+* [ENHANCEMENT] TSDB: Significantly reduce WAL size kept around after a block cut. #7098
+* [ENHANCEMENT] Discovery: Add `architecture` meta label for EC2. #7000
+* [BUGFIX] React UI: Fixed multiselect legend on OSX. #6880
+* [BUGFIX] Remote Write: Fixed blocked resharding edge case. #7122
+* [BUGFIX] Remote Write: Fixed remote write not updating on relabel configs change. #7073
+
```
roidelapluie, line 11: one new line only

roidelapluie: (can be fixed when we cut the next release)

CHANGELOG.md
beorn7: > > That document doesn't state anything about production readiness.
>
> Exactly, and I don't believe we have any docs that do.

I am not a native speaker, but I'm pretty sure that the word "experimental" generally does not mean "as production ready as a stable feature". Should our intention be that the only implication of "experimental" is that those feature do not fall under semantic versioning, then we should better call that out. As we don't call that out, and the document you quoted doesn't even look like a definition of "experimental", much less an exhaustive one, I don't get your line of argument, which with you have now derailed not only this PR but also the node_exporter TLS discussion.

> I mean "we" here.

It's hopeless… but I have realized that early 2017 already. What am I even doing here?
I'll stop here and now.

bwplotka: Please, let's not stop @beorn7 @brian-brazil ! 

We might have different opinions, but by **doing nothing** and just hiding we won't find a viable middle ground where everyone is happy! And I am pretty sure we can all find that spot :hugs: 

@beorn7, please give you feedback as often as possible! Your opinion is extremely valuable for the project, community (and definitely for me!), and the same is for @brian-brazil 's opinion. We all can learn only by communication ([as Go, humans do not share memory](https://blog.golang.org/codelab-share))!

I think it's totally fair to take discussion about `experimental` meaning to DevSummit. I will make sure the topic is there. No matter what was said before or now, it's not clear for some of us, so let's agree on this at next meeting.

My 2c: @brian-brazil you are amazing, but I will be honest here: It's a bit rude to say `we/us` for the topic where clearly at least two people (me and @beorn7) among those `we` are not agreeing with the opinion. (: PLUS people ask you many times to not say `we`...

**TBH, no matter if by saying `we` @brian-brazil you mean 90% or 1%. population If your friends kindly ask you to not use one word that in some way hurts them (even though you don't mean to), let's maybe not use that. It is really the same as the word `he` or `black` for gender/race contexts.** Can we kindly ask you?  :hugs:

And let's discuss further on our meetings! :pray:

bwplotka:
So far so good in terms of performance. Bit better allocs wise.

CHANGELOG.md
RichiH: Not something I plan to argue about, but with my user hat on, that has been relevant several times. Much of my initial need for help in 2015 was around Go and the version needed to compile, not around Prometheus proper.

roidelapluie: Users don't compile prometheus (?)

bwplotka:
Looks all is good or even a bit better (especially cut seems to be quicker/earlier and we evict memory quicker (probably due to WAL size?)

Over 2d:

image
image
image

Will release 2.18 tomorrow or on Thur to give at least 3 work days for users to give feedback.

vipinvkmenon:
@roidelapluie @bwplotka ..
As the jaeger support is experimental I suppose there wouldn't be a documentation.,
However any pointers on evaluating it / trying it out and some information about the feature goal?

roidelapluie:
The Jaeger support is configurable with env variables:

e.g.

JAEGER_DISABLED=false JAEGER_SAMPLER_TYPE=const JAEGER_SAMPLER_PARAM=1 ./prometheus

In mid term it will be replaced by open tracing.

PR #7169, opened 2020-04-26
Fix race in tsdb

gaurav1086:
When a go loop launches goroutines, it is a common mistake to assume
that the routines can reference the loop value from their specific loop
iteration. Actually, all goroutines share the same reference to the loop
veriable, and they typically all see the last value. To correct this, the loop variable must be captured as a parameter to the anonymous func.

### DIFF
--- a/discovery/kubernetes/client_metrics.go
+++ b/discovery/kubernetes/client_metrics.go
@@ -18,7 +18,6 @@ import (
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
-
 	"k8s.io/client-go/tools/metrics"
 	"k8s.io/client-go/util/workqueue"
 )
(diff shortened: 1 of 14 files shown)


tsdb/index/postings.go
```diff
@@ -168,12 +168,12 @@ func (p *MemPostings) EnsureOrder() {
 	wg.Add(n)
 
 	for i := 0; i < n; i++ {
-		go func() {
+		go func(i interface{}) {
```
bwplotka, line 171:
```suggestion
		go func(i int) {
```

tsdb/index/postings.go
```diff
@@ -168,12 +168,12 @@ func (p *MemPostings) EnsureOrder() {
...
 			for l := range workc {
 				sort.Slice(l, func(i, j int) bool { return l[i] < l[j] })
```
roidelapluie, line 173:
```suggestion
				sort.Slice(l, func(a, b int) bool { return l[a] < l[b] })
```

tsdb/index/postings.go
roidelapluie, line 171:
```suggestion
		go func() {
```
That will need to be reverted

tsdb/index/postings.go
bwplotka: or this

discovery/kubernetes/client_metrics.go
```diff
@@ -160,24 +159,3 @@ func (clientGoWorkqueueMetricsProvider) NewRetriesMetric(name string) workqueue.
 	// Retries are not used so the metric is omitted.
 	return noopMetric{}
 }
-func (clientGoWorkqueueMetricsProvider) NewDeprecatedDepthMetric(name string) workqueue.GaugeMetric {
```
bwplotka, line 163: unrelated change?

tsdb/head.go
```diff
@@ -681,9 +681,11 @@ func (h *Head) Truncate(mint int64) (err error) {
...
 		return nil // no segments yet.
 	}
-	// The lower third of segments should contain mostly obsolete samples.
-	// If we have less than three segments, it's not worth checkpointing yet.
-	last = first + (last-first)/3
+	// The lower two thirds of segments should contain mostly obsolete samples.
+	// If we have less than two segments, it's not worth checkpointing yet.
+	// With the default 2h blocks, this will keeping up to around 3h worth
+	// of WAL segments.
+	last = first + (last-first)*2/3
```
codesome, line 688: Could you rebase your PR? This change is already in master.

gaurav1086:
Will create a fresh PR. Closing this one.

PR #7175, opened 2020-04-27
[tsdb] Goroutine fix ambiguous variable

gaurav1086:
Fix ambiguous variable in the goroutine.


gaurav1086:
ref #7169 
@bwplotka @roidelapluie @codesome

Issue #6359, opened 2019-11-22
Starting prometheus with a very old block causes many empty wal segments.


krasi-georgiev:
By default each compaction processes 2h of data from the head and than reloads the db. At the reload it sets the head mint progressively advancing 2h on every compaction run. On every compaction run a new segment is created which is empty. 
So starting Prometheus with a block that is 10d old will cause around 300 compaction with a new segment for each run and a checkpoint when it accumulates 4 segments.

Need to revisit how is the head min time set when running the compaction and reloading the db.

I think the  solution should be to set the head min time to the first sample in the wal when opening the db of if the wal doesn't exist should set the head mint to the first appended sample and the right place for this should be when reloading the DB.

cstyan:
@krasi-georgiev I added the logic to always create a new segment, so I'd like to have a go at fixing this. Do you have any logs from when you've seen this happen? The segment creation should only happen if there was an error in `db.head.Init`, in which case we call `wlog.Repair` which has the logic to always create a new segment.

krasi-georgiev:
like I said just create a block that is a year or few months old and you will replicate it.

codesome:
> I think the solution should be to set the head min time to the first sample in the wal when opening the db of if the wal doesn't exist should set the head mint to the first appended sample and the right place for this should be when reloading the DB.

This sounds like the right thing to do as after recovery and before any append, the `Head.MinTime()` should return the min time of the head. `initAppender` already takes care of setting it to the timestamp of the first sample if nothing is recovered from WAL.

Transwert:
Hey guys,  is anybody working on this issue,  and if they are, can I join as well, 
And also can some put pointers as well for correct direction, ( along a little-bit briefing about issue as well ).

codesome:
@Transwert go ahead with the fix.
Krasi mentioned 1 possible solution above which can be attempted. Reiterating it:
* Set the head min time to the first sample in the wal when opening the db
* If the wal doesn't exist should set the head mint to the first appended sample and the right place for this should be when reloading the DB.
    * For this point, if empty WAL files get created even before the first sample append, we could set it to the current time and then set it to first sample timestamp when it comes.

Pointers: `loadWAL` and sample appending logic in `tsdb/head.go`. Let me know if you need any help.

codesome:
More information here:
It's not just an old block, it also happens if the head has very old data (a big gap between Prometheus coming down and starting again).

[...]

PR #7178, opened 2020-04-27
WIP: Fixes empty segments creation in tsdb when presence of a very old block

Harkishen-Singh:

Fixes: #6359

Fixes empty segments creation when the most recent block is old. If no segments found, the `minTime` is shifted to the maxTime of the last block.

**Note:** Unit-test cases are not updated yet. Expect CI tests to fail.

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -39,6 +39,7 @@ import (
 	"github.com/prometheus/prometheus/tsdb/chunkenc"
 	tsdb_errors "github.com/prometheus/prometheus/tsdb/errors"
 	"github.com/prometheus/prometheus/tsdb/fileutil"
+	"github.com/prometheus/prometheus/tsdb/record"
 	// Load the package into main to make sure minium Go version is met.
 	_ "github.com/prometheus/prometheus/tsdb/goversion"
 	"github.com/prometheus/prometheus/tsdb/wal"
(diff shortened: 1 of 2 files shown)


Harkishen-Singh:
cc @brian-brazil @codesome @krasi-georgiev

tsdb/db.go
```diff
@@ -606,7 +607,29 @@ func open(dir string, l log.Logger, r prometheus.Registerer, opts *Options, rngs
...
+		dec := record.Decoder{}
+		for reader.Next() {
+			currRecord := reader.Record()
+			if record.Samples == dec.Type(currRecord) {
+				var samples []record.RefSample
+				samples, err := dec.Samples(currRecord, samples)
+				if err != nil {
+					return nil, errors.Wrap(err, "samples append")
+				}
+				minValidTime = samples[0].T
```
brian-brazil, line 626: The first sample is not required to be the one with the lowest timestamp.

Harkishen-Singh: I think I did this since it was written in the issue as a desirable approach. Well, then should it be the minimum timestamp?

brian-brazil: Yes, we need to consider all the samples.

Harkishen-Singh:
@brian-brazil done!

brian-brazil:
This would have us read the WAL twice at startup, which isn't desirable. The head rather than the db should also be dealing with WAL-level stuff.

Harkishen-Singh:
@brian-brazil I hope the recent changes pushed is what we want.

@codesome, for the case involving no wal but a very old block, should we set the `h.minValidTime` to current timestamp?

brian-brazil:
That looks better alright.

> for the case involving no wal but a very old block, should we set the h.minValidTime to current timestamp?

Could we fallback to whatever logic we use when there's no data at all?

krasi-georgiev:
> @codesome, for the case involving no wal but a very old block, should we set the h.minValidTime to current timestamp?

I think to the maxt of the last block on disk.

Harkishen-Singh:
How do we decided when to call a block very old. Like beyond a particular time range? 

Basically what I get through this discussion is that we simply consider the first sample. But, if wal is empty, take the maxt of last block. However, the later part will again cause lots of empty segments.

@krasi-georgiev @brian-brazil

krasi-georgiev:
yeah I am not sure, I guess need to dig in the code and make a list of all cases.

NO WAL NO blocks.
NO WAL with blocks
With WAL and NO blocks
With WAL and with blocks

brian-brazil:
We'd need to handle this at the compaction stage rather than the WAL reading stage to do that.

It also occurs to me that setting min valid time like this is incorrect, as it prevents ingesting data that is older than the oldest sample in the WAL. Not that that should be common, but it's not something we should break.

krasi-georgiev:
@brian-brazil you probably have the most context here so can you sum up how should the head min/maxt be set then? 

Maybe a good start would be to remind us what was the accepted time range in the past that samples are accepted. For example if current timestamp is 10000 how much in the past would tsdb accept samples?

brian-brazil:
1-3 hours, depending on where you are in the compaction cycle.

I suspect this is something we need to handle at the compaction stage.

Harkishen-Singh:
@codesome as far as I remember, there was some other PR which covered this PR's purpose as well. Right?

codesome:
That should be https://github.com/prometheus/prometheus/pull/7579

roidelapluie:
We have looked at it in the bug scrub and decided to close this as superseded by  #7579

Issue #5735, opened 2019-07-03
remote_write doesn't log error if using proxy_url


AaronM04:
## Proposal
**Use case. Why is this important?**

It's important to have log output when something goes wrong. For example, if metrics are not sent to persistent storage, we need to have a log event.

## Bug Report

**What did you do?**

I created a `remote_write` section in my config with an HTTPS URL, tested it, then added a `proxy_url` section containing `'http://squid.proxy.that.blocks.requests:3128/'`. The proxy is configured to block access to the `url` we are using.

**What did you expect to see?**

I expected to see at least one error log line, saying the remote write failed. 

**What did you see instead? Under which circumstances?**

Metrics did not show up, and no log lines were logged. Circumstances: proxy is blocking requests, as noted above.

**Environment**

* System information:

	Linux 2.6.32-696.el6.x86_64 x86_64

* Prometheus version:

```
prometheus, version 2.10.0 (branch: HEAD, revision: d20e84d0fb64aff2f62a977adc8cfb656da4e286)
  build user:       root@a49185acd9b0
  build date:       20190525-12:28:13
  go version:       go1.12.5
```

* Alertmanager version:

	N/A

* Prometheus configuration file:
```
global:
  scrape_interval:     15s
  evaluation_interval: 15s
  scrape_timeout:      2s

# Alertmanager configuration
alerting:
  alertmanagers:
  - static_configs:
    - targets:
      # - alertmanager:9093

# Load rules once and periodically evaluate them according to the global 'evaluation_interval'.
rule_files:
  # - "first_rules.yml"
  # - "second_rules.yml"

# A scrape configuration containing exactly one endpoint to scrape:
# Here it's Prometheus itself.
scrape_configs:
  # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
  - job_name: 'prometheus'

    # metrics_path defaults to '/metrics'
    # scheme defaults to 'http'.

    static_configs:
    - targets: ['localhost:9090']
    
  - job_name: 'some-job-name'
    static_configs:
    - targets:
        - 'localhost:10002'
        - 'localhost:10000'
        - 'localhost:10001'
      labels:
        environment: 'someenvname'
        host: 'internal-host-fqdn'

remote_write:
  - url: 'https://some.cortex.instance/api/prom/push'

    basic_auth:
      username: 'some_user'
      password: 'some_password'

    proxy_url: 'http://squid.proxy.that.blocks.requests:3128/'
```

[...]

PR #7184, opened 2020-04-29
[remote-storage] use warn log level when send samples to remote failed

qingwave:
**What type of PR is this?**
Add a detail error for [sendSamplesWithBackoff](https://github.com/prometheus/prometheus/blob/05038b48bdf08dd72731bfaf74c9f5e5eaff5f39/storage/remote/queue_manager.go#L875)

**What this PR does / why we need it:**
This useful for user to find remote write error directly, rather than a `context deadline exceeded`. 
e.g. `no route to host` occured, we can easily find it.

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -15,6 +15,7 @@ package remote
 
 import (
 	"context"
+	"fmt"
 	"math"
 	"strconv"
 	"sync"
@@ -873,6 +874,7 @@ func (s *shards) sendSamples(ctx context.Context, samples []prompb.TimeSeries, b
 
 // sendSamples to the remote storage with backoff for recoverable errors.
 func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.TimeSeries, buf *[]byte) error {
+	var err error
 	backoff := s.qm.cfg.MinBackoff
 	req, highest, err := buildWriteRequest(samples, *buf)
 	*buf = req
@@ -885,11 +887,11 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
 	for {
 		select {
 		case <-ctx.Done():
-			return ctx.Err()
+			return fmt.Errorf("%v, with send batch error %v", ctx.Err(), err)
 		default:
 		}
 		begin := time.Now()
-		err := s.qm.client().Store(ctx, req)
+		err = s.qm.client().Store(ctx, req)
 
 		s.qm.sentBatchDuration.Observe(time.Since(begin).Seconds())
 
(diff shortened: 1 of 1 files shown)


qingwave:
> Thank you for the PR!
> 
> A couple things here, I don't like that we could potentially be printing out nil errors here, and would prefer the error to be reported more quickly as it would currently be necessary to shutdown or change a queue to get this error.
> 
> What would you and @cstyan think of increasing the level of the "Failed to send batch, retrying" log message? We use a dedupe logger so it shouldn't be too noisy.
> 
> This could fix #5735 depending on the implementation

If dedupe logger is ok, increasing the `send batch error` log level maybe reasonable.

cstyan:
I think bumping the log level while we're still using the deduper is fine. Especially given that we also have the metric for retried samples.

unfor19:
Was this merged to `prom/prometheus:v2.24.1` ? I can see the proper error when using `debug` level but not `info` level (which is "higher" than `warn`)

Am I missing something?

roidelapluie:
Can you paste the log line ?

roidelapluie:
Also note that the error will not be repeated if it is the same (IIRC)

unfor19:
@roidelapluie
1. Changed `log.level` to `debug`, `info` and `warn`
1. Disallowed outbound connection (AWS Security Group > removed `outbound to 0.0.0.0/0` rule)
1. Restarted Prometheus - `kill -HUP 1`

`--log.level=debug` (works as expected)
```
ts=2021-01-27T11:21:21.540Z caller=dedupe.go:112 component=remote
level=debug remote_name=lalala url="https://metric-api.newrelic.com/prometheus/v1/write?X-License-Key=TOKEN&prometheus_server=prom" msg="failed to send batch, retrying" err="Post \"https://metric-api.newrelic.com/prometheus/v1/write?X-License-Key=TOKEN&prometheus_server=prom\": context deadline exceeded"
```

`--log.level=info` - no error appears, even though there's outbound connection is not allowed
```
ts=2021-01-27T11:51:24.781Z caller=dedupe.go:112 component=remote
level=info remote_name=lalala url="https://metric-api.newrelic.com/prometheus/v1/write?X-License-Key=TOKEN&prometheus_server=prom" msg="Done replaying WAL" duration=14.391332476s
```

`--log.level=warn` - nothing to show, no logs at all

csmarchbanks:
This was a part of v2.19, but it looks like in v2.23 a change was merged that accidentally changed it back to debug. Here is a PR to change this line back to warn: https://github.com/prometheus/prometheus/pull/8412.

So v2.19.x -> v2.22.x should have the debug log if you need it. Sorry about that!

unfor19:
@csmarchbanks - Thank you very much for the clarification, I'll use v2.22.x, no biggy. Originally I used 2.18.1 so 2.22.x is still an upgrade ... 😄 

Thanks!

PR #7187, opened 2020-04-29
Jaeger: Create child span for remote read

roidelapluie:
This creates a child span for remote read calls instead of modifying the parent span.

The previous version would change the values of the parent span:
remread

Which is problematic because parent span is about the engine itself and should not get the http properties, + would be erased if we have multiple remote reads.

The new behaviour is to have a clean remote read span and a clear http post span:

remread

### DIFF
--- a/storage/remote/client.go
+++ b/storage/remote/client.go
@@ -163,7 +163,12 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
 
 	if parentSpan := opentracing.SpanFromContext(ctx); parentSpan != nil {
 		var ht *nethttp.Tracer
-		httpReq, ht = nethttp.TraceRequest(parentSpan.Tracer(), httpReq)
+		httpReq, ht = nethttp.TraceRequest(
+			parentSpan.Tracer(),
+			httpReq,
+			nethttp.OperationName("Remote Read"),
+			nethttp.ClientTrace(false),
+		)
 		defer ht.Finish()
 	}
 
--- a/storage/remote/client.go
+++ b/storage/remote/client.go
@@ -151,7 +151,12 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
 	httpReq.Header.Set("User-Agent", userAgent)
 	httpReq.Header.Set("X-Prometheus-Remote-Read-Version", "0.1.0")
 
-	if span := opentracing.SpanFromContext(ctx); span != nil {
+	if parentSpan := opentracing.SpanFromContext(ctx); parentSpan != nil {
+		span := opentracing.StartSpan(
+			"Remote Read",
+			opentracing.ChildOf(parentSpan.Context()),
+		)
+		defer span.Finish()
 		ext.SpanKindRPCClient.Set(span)
 		ext.HTTPUrl.Set(span, httpReq.URL.String())
 		ext.HTTPMethod.Set(span, httpReq.Method)
--- a/storage/remote/client.go
+++ b/storage/remote/client.go
@@ -27,12 +27,12 @@ import (
 	"github.com/gogo/protobuf/proto"
 	"github.com/golang/snappy"
 	"github.com/opentracing/opentracing-go"
-	"github.com/opentracing/opentracing-go/ext"
 	"github.com/pkg/errors"
 	config_util "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/common/version"
 
+	"github.com/opentracing-contrib/go-stdlib/nethttp"
 	"github.com/prometheus/prometheus/prompb"
 )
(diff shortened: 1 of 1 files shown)


roidelapluie:
cc @yurishkuro

storage/remote/client.go
```diff
@@ -151,7 +151,17 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
 	httpReq.Header.Set("User-Agent", userAgent)
 	httpReq.Header.Set("X-Prometheus-Remote-Read-Version", "0.1.0")
 
-	if span := opentracing.SpanFromContext(ctx); span != nil {
+	var span opentracing.Span
+	if parentSpan := opentracing.SpanFromContext(ctx); parentSpan != nil {
```
yurishkuro, line 155: have you considered simply using a middleware like https://github.com/opentracing-contrib/go-stdlib ?

storage/remote/client.go
```diff
@@ -151,7 +151,17 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
...
+		span = opentracing.StartSpan(
+			"Remote Read",
+			opentracing.ChildOf(parentSpan.Context()),
+		)
+		defer func() {
+			if span != nil {
```
yurishkuro, line 161: it's never nil in this position

storage/remote/client.go
```diff
@@ -169,6 +179,10 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
 	if err != nil {
 		return nil, errors.Wrap(err, "error sending request")
 	}
+	if span != nil {
+		span.Finish()
```
yurishkuro, line 183: already done in L162

storage/remote/client.go
```diff
@@ -151,7 +151,17 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
 	httpReq.Header.Set("User-Agent", userAgent)
 	httpReq.Header.Set("X-Prometheus-Remote-Read-Version", "0.1.0")
 
-	if span := opentracing.SpanFromContext(ctx); span != nil {
+	var span opentracing.Span
```
yurishkuro, line 154: can be moved inside the following if()

storage/remote/client.go
roidelapluie: It is nil if the defer is called after we have finished the span already

storage/remote/client.go
roidelapluie: This is to close the child span just after the http request

storage/remote/client.go
roidelapluie: Not if I want to close it just after the http request is done

storage/remote/client.go
roidelapluie: thank you for your review, I will look into that.

roidelapluie:
@yurishkuro Thanks for your review. I have changed to use the middleware. WDYT?
remread

storage/remote/client.go
```diff
@@ -151,21 +156,18 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
...
-		)
-	}
-
 	ctx, cancel := context.WithTimeout(ctx, c.timeout)
 	defer cancel()
 
-	httpResp, err := c.client.Do(httpReq.WithContext(ctx))
+	httpReq = httpReq.WithContext(ctx)
+
+	if parentSpan := opentracing.SpanFromContext(ctx); parentSpan != nil {
```
yurishkuro, line 164: Looks much cleaner. You may want to consider using ClientOptions, specifically `OperationName()` to get a more business-specific span name than `HTTP POST`, and `ClientTrace(false)` to avoid the overhead of fine-grained micrologs added to the span from httptrace.

roidelapluie: I have added the two options, indeed this is nice.

roidelapluie:
remread

roidelapluie:
cc @tomwilkie @csmarchbanks @cstyan

cstyan:
Looks like this span was already being created, just in a slightly different way? Can you clarify what this will change in the first comment at the top of the PR.

roidelapluie:
> Looks like this span was already being created, just in a slightly different way? Can you clarify what this will change in the first comment at the top of the PR.

This creates a child span for remote read calls instead of modifying the parent span.

The previous version would change the values of the parent span:
remread

Which is problematic because parent span is about the engine itself and should not get the http properties, + would be erased if we have multiple remote reads.

The new behaviour is to have a clean remote read span and a clear http post span:

remread

roidelapluie:
@cstyan is that ok for you?

cstyan:
@roidelapluie Sorry! Yes, thanks for the clarification.

PR #7188, opened 2020-04-30
Remove duplicate metrics in QueueManager

csmarchbanks:
Right now any new metrics added for remote write need to be added to
both the QueueManager struct, and the queueManagerMetrics struct.
Instead, use the queueManagerMetrics struct directly from QueueManager.

The newQueueManagerMetrics constructor will now create the metrics for a
specific queue with name and endpoint pre-populated, and a new copy of
the struct will be created specifically for each queue.

This also fixes a bug where prometheus_remote_storage_sent_bytes_total
is not being unregistered after a queue is changed.

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -46,152 +46,134 @@ const (
 )
 
 type queueManagerMetrics struct {
-	succeededSamplesTotal     *prometheus.CounterVec
-	failedSamplesTotal        *prometheus.CounterVec
-	retriedSamplesTotal       *prometheus.CounterVec
-	droppedSamplesTotal       *prometheus.CounterVec
-	enqueueRetriesTotal       *prometheus.CounterVec
-	sentBatchDuration         *prometheus.HistogramVec
-	queueHighestSentTimestamp *prometheus.GaugeVec
-	queuePendingSamples       *prometheus.GaugeVec
-	shardCapacity             *prometheus.GaugeVec
-	numShards                 *prometheus.GaugeVec
-	maxNumShards              *prometheus.GaugeVec
-	minNumShards              *prometheus.GaugeVec
-	desiredNumShards          *prometheus.GaugeVec
-	bytesSent                 *prometheus.CounterVec
+	reg prometheus.Registerer
+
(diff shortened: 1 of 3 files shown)


storage/remote/queue_manager.go
```diff
@@ -46,152 +46,134 @@ const (
 )
 
 type queueManagerMetrics struct {
```
cstyan, line 48: can we leave a comment saying any metrics added need to be both `newQueueManagerMetrics` and `unregister`

storage/remote/queue_manager.go
```diff
@@ -214,6 +196,25 @@ func newQueueManagerMetrics(r prometheus.Registerer) *queueManagerMetrics {
 	return m
 }
 
+func (m *queueManagerMetrics) unregister() {
```
cstyan, line 199: would be nice if we could add slices of pointers (interface types? ) to the struct and just range loop through those for unregister?

storage/remote/queue_manager.go
csmarchbanks: I added a test to enforce that any metric registered is also unregistered, WDYT?

storage/remote/queue_manager.go
csmarchbanks: Hmmm... I wouldn't want consumers to need to use a map and casting, so would that be in addition to the metrics being defined on the struct?

csmarchbanks:
Yeah, maybe duplicate wasn't the right word, I just didn't like having to remember to go to many different places anytime a metric was modified.

cstyan:
@csmarchbanks LGTM except you have a merge conflict now. Just to clarify, the test ensures any registered metric from Queue metrics is unregistered, but we can't test that all the metrics were registered in the first place?

csmarchbanks:
I have fixed the merge conflict.

@cstyan, you are correct in understanding the test. In a future PR we could consider a better way of making sure metrics are registered (or possibly tested). Or perhaps move to the promauto with registry pattern?

cstyan:
:+1: sounds good to me, we can make more improvements soon, no reason to not merge this now imo

PR #7189, opened 2020-04-30
Support environment variable for global.external_labels

ryli17:
Prometheus team declined to enable environment variable support in config file due to security concerns. However, it is not convenient when deploy Prometheus HA cluster. For example, deploying 3 replicas of prometheus and remote_write to cortex. Cortex will perform de-duplication based on external labels.

This change will enable environment variable expanding only for global.external_labels, so that the security risk is minimized, and deployment convenience is satisfied.

How to use:
In prometheus configuration file, you can have the following:

```
global:
  external_labels:
    __replica__: $POD_NAME

```
$POD_NAME will be replaced by environment variable value.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/url"
+	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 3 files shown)


juliusv:
Here is the full discussion context for that vote: https://groups.google.com/forum/#!msg/prometheus-developers/tSCa4ukhtUw/SYdSwFfqCAAJ

(One immediate issue with just doing `os.ExpandEnv()` is that it breaks backwards-compatibility and even if enabled optionally, brings up escaping questions around what we would have to do when someone did want to have the character string `$foo` literally in one of their config fields.)

roidelapluie:
As per the bug scrub and the mentioned vote, we have decided to close this pull request. Thanks!

PR #7193, opened 2020-05-01
Simplify the Getting Started documentation.

### LINKED ISSUE #6956: Getting started should not require a Golang installation
@oddlittlebird gave some really good feedback that we shouldn't require users to install the Go toolchain in the Getting Started guide here: https://prometheus.io/docs/prometheus/latest/getting_started/#starting-up-some-sample-targets

Can we use a binary or something else there? Also Diana could you also add more feedback here around the Getting Started guide?

hdost:
- Reduce the level of entry to start gathering metrics with prometheus
  by suggesting to just download pre-built exporters instead of requiring
  the reader to download an entire Golang build chain and checkout a project.
- Include examples for both *nix and Windows.

Fix #6956

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -118,7 +118,7 @@ For more about the expression language, see the
 To graph expressions, navigate to http://localhost:9090/graph and use the "Graph"
 tab.
 
-For example, enter the following expression to graph the per-second rate of chunks 
+For example, enter the following expression to graph the per-second rate of chunks
 being created in the self-scraped Prometheus:
 
 ```
@@ -132,36 +132,26 @@ Experiment with the graph range parameters and other settings.
 Let us make this more interesting and start some example targets for Prometheus
 to scrape.
 
-The Go client library includes an example which exports fictional RPC latencies
-for three services with different latency distributions.
-
-Ensure you have the [Go compiler installed](https://golang.org/doc/install) and
-have a [working Go build environment](https://golang.org/doc/code.html) (with
-correct `GOPATH`) set up.
-
-Download the Go client library for Prometheus and run three of these example
-processes:
+The Node Exporter is used as an example target, for more information on using it
+[see these instructions.](https://prometheus.io/docs/guides/node-exporter/)
 
 ```bash
-# Fetch the client library code and compile example.
-git clone https://github.com/prometheus/client_golang.git
-cd client_golang/examples/random
-go get -d
-go build
+tar -xzvf node_exporter-*.*.tar.gz
+cd node_exporter-*.*
 
 # Start 3 example targets in separate terminals:
-./random -listen-address=:8080
-./random -listen-address=:8081
-./random -listen-address=:8082
+./node_exporter --web.listen-address 127.0.0.1:8080
+./node_exporter --web.listen-address 127.0.0.1:8081
+./node_exporter --web.listen-address 127.0.0.1:8082
 ```
 
 You should now have example targets listening on http://localhost:8080/metrics,
 http://localhost:8081/metrics, and http://localhost:8082/metrics.
 
-## Configuring Prometheus to monitor the sample targets
+## Configure Prometheus to monitor the sample targets
 
 Now we will configure Prometheus to scrape these new targets. Let's group all
-three endpoints into one job called `example-random`. However, imagine that the
+three endpoints into one job called `node`. However, imagine that the
 first two endpoints are production targets, while the third one represents a
 canary instance. To model this in Prometheus, we can add several groups of
 endpoints to a single job, adding extra labels to each group of targets. In
@@ -173,7 +163,7 @@ section in your `prometheus.yml` and restart your Prometheus instance:
 
 ```yaml
 scrape_configs:
-  - job_name:       'example-random'
+  - job_name:       'node'
 
     # Override the global default and scrape targets from this job every 5 seconds.
     scrape_interval: 5s
@@ -189,8 +179,7 @@ scrape_configs:
 ```
 
 Go to the expression browser and verify that Prometheus now has information
-about time series that these example endpoints expose, such as the
-`rpc_durations_seconds` metric.
+about time series that these example endpoints expose, such as `node_cpu_seconds_total`.
 
 ## Configure rules for aggregating scraped data into new time series
 
@@ -198,27 +187,26 @@ Though not a problem in our example, queries that aggregate over thousands of
 time series can get slow when computed ad-hoc. To make this more efficient,
 Prometheus allows you to prerecord expressions into completely new persisted
 time series via configured recording rules. Let's say we are interested in
-recording the per-second rate of example RPCs
-(`rpc_durations_seconds_count`) averaged over all instances (but
-preserving the `job` and `service` dimensions) as measured over a window of 5
-minutes. We could write this as:
+recording the per-second rate of cpu time (`node_cpu_seconds_total`) averaged
+over all cpus per instance (but preserving the `job`, `instance` and `mode`
+dimensions) as measured over a window of 5 minutes. We could write this as:
 
 ```
-avg(rate(rpc_durations_seconds_count[5m])) by (job, service)
+avg by (job, instance, mode) (rate(node_cpu_seconds_total[5m]))
 ```
 
 Try graphing this expression.
 
 To record the time series resulting from this expression into a new metric
-called `job_service:rpc_durations_seconds_count:avg_rate5m`, create a file
+called `job_instance_mode:node_cpu_seconds:avg_rate5m`, create a file
 with the following recording rule and save it as `prometheus.rules.yml`:
 
 ```
 groups:
-- name: example
+- name: cpu-node
   rules:
-  - record: job_service:rpc_durations_seconds_count:avg_rate5m
-    expr: avg(rate(rpc_durations_seconds_count[5m])) by (job, service)
+  - record: job_instance_mode:node_cpu_seconds:avg_rate5m
+    expr: avg by (job, instance, mode) (rate(node_cpu_seconds_total[5m]))
 ```
 
 To make Prometheus pick up this new rule, add a `rule_files` statement in your `prometheus.yml`. The config should now
@@ -245,7 +233,7 @@ scrape_configs:
     static_configs:
       - targets: ['localhost:9090']
 
-  - job_name:       'example-random'
+  - job_name:       'node'
 
     # Override the global default and scrape targets from this job every 5 seconds.
     scrape_interval: 5s
@@ -261,5 +249,5 @@ scrape_configs:
 ```
 
 Restart Prometheus with the new configuration and verify that a new time series
-with the metric name `job_service:rpc_durations_seconds_count:avg_rate5m`
+with the metric name `job_instance_mode:node_cpu_seconds:avg_rate5m`
 is now available by querying it through the expression browser or graphing it.


docs/getting_started.md
```diff
@@ -132,36 +132,53 @@ Experiment with the graph range parameters and other settings.
...
-./random -listen-address=:8081
-./random -listen-address=:8082
+./node_exporter --web.listen-address "127.0.0.1:8080"
+./node_exporter --web.listen-address "127.0.0.1:8081"
+./node_exporter --web.listen-address "127.0.0.1:8082"
 ```
 
-You should now have example targets listening on http://localhost:8080/metrics,
-http://localhost:8081/metrics, and http://localhost:8082/metrics.
+### Setup WMI exporter on Windows
```
brian-brazil, line 162: Let's stick with just the node exporter for a 1st steps, otherwise we'll end up needing to duplicate things. This is a guide to using Prometheus, not general monitoring for all common OSes.

docs/getting_started.md
```diff
@@ -132,36 +132,53 @@ Experiment with the graph range parameters and other settings.
...
+./wmi_exporter-<version>-<arch>.exe --telemetry.addr ":8080"
+./wmi_exporter-<version>-<arch>.exe --telemetry.addr ":8081"
+./wmi_exporter-<version>-<arch>.exe --telemetry.addr ":8082"
+```
 
 ## Configuring Prometheus to monitor the sample targets
 
 Now we will configure Prometheus to scrape these new targets. Let's group all
-three endpoints into one job called `example-random`. However, imagine that the
+three endpoints into one job called `example-exporter`. However, imagine that the
```
brian-brazil, line 181: As we're using the node exporter, the job should now be node.

docs/getting_started.md
```diff
@@ -190,35 +207,38 @@ scrape_configs:
...
 
 ```
 groups:
-- name: example
+- name: cpu-example
   rules:
-  - record: job_service:rpc_durations_seconds_count:avg_rate5m
-    expr: avg(rate(rpc_durations_seconds_count[5m])) by (job, service)
+  - record: job_instance:node_cpu_usage_ratio:avg_rate5m
+    expr: avg(irate(node_cpu_seconds_total[5m])) by (job, instance, mode)
```
brian-brazil, line 241: This should stay a rate, and mode should not be added.

The convention is also to put the by after the operator.

docs/getting_started.md
hdost: So I understand this is a guide for "getting started" in prometheus. That means the audience very well might not be using Linux. Especially if they are just getting into this. Originally I wasn't going to add the windows specific section, but if it the idea is to make it a quick start guide then it should be for all.

That being said, possibly a compromise. Instead of node exporter, maybe something that is compatible with all major OSs, so that the instructions can be singular, but still provide the beginners with something to they don't have to push too far to figure out.

brian-brazil: The vast majority of our users will be not be solely on Windows, we just need enough to get them going.

docs/getting_started.md
```diff
@@ -132,36 +132,53 @@ Experiment with the graph range parameters and other settings.
...
+The next steps will depend what operating system you're running. You only need to
+follow one of them, and after that you can move on to the next section. By the end
+of the next sub-sections you should have example targets listening on
+http://localhost:8080/metrics, http://localhost:8081/metrics, and
+http://localhost:8082/metrics.
 
-Ensure you have the [Go compiler installed](https://golang.org/doc/install) and
-have a [working Go build environment](https://golang.org/doc/code.html) (with
-correct `GOPATH`) set up.
+### Setup node_exporter on Linux/Mac/Unix
```
oddlittlebird, line 141:
```suggestion
### Set up node_exporter on Linux/Mac/Unix
```

oddlittlebird: setup = noun
set up = verb

docs/getting_started.md
oddlittlebird, line 162:
```suggestion
### Set up WMI exporter on Windows
```

docs/getting_started.md
```diff
@@ -132,36 +132,53 @@ Experiment with the graph range parameters and other settings.
...
+# Go to the download location
+cd ~\Downloads
+
+# <version> and <arch> will depend on what you download
+./wmi_exporter-<version>-<arch>.exe --telemetry.addr ":8080"
+./wmi_exporter-<version>-<arch>.exe --telemetry.addr ":8081"
+./wmi_exporter-<version>-<arch>.exe --telemetry.addr ":8082"
+```
 
 ## Configuring Prometheus to monitor the sample targets
```
oddlittlebird, line 178:
```suggestion
## Configure Prometheus to monitor the sample targets
```

oddlittlebird: Changed to maintain consistency with other headings in this topic.

docs/getting_started.md
hdost: I can change the convention and the function, however adding the `mode` provides the most useful information. Why would we want to provide essentially a straight line for people to look at? If you include mode you can see useful information and provide people with something.

docs/getting_started.md
hdost: Ended up needing to remove section header :(

docs/getting_started.md
hdost: Ended up needing to remove section header :(

docs/getting_started.md
hdost: As a follow up on convention, does that mean that examples like https://prometheus.io/docs/prometheus/latest/querying/functions/#histogram_quantile should also be updated to follow the `<func> by (<labels>) (<metric>)` convention?

docs/getting_started.md
hdost: I have included a link to WMI and a small reference to the a supplemental metric.

hdost: Check the new commit, I think this would be a good compromise.

docs/getting_started.md
```diff
@@ -132,36 +132,35 @@ Experiment with the graph range parameters and other settings.
...
 
 You should now have example targets listening on http://localhost:8080/metrics,
 http://localhost:8081/metrics, and http://localhost:8082/metrics.
 
-## Configuring Prometheus to monitor the sample targets
+## Configure Prometheus to monitor the sample targets
 
 Now we will configure Prometheus to scrape these new targets. Let's group all
-three endpoints into one job called `example-random`. However, imagine that the
+three endpoints into one job called `example-node`. However, imagine that the
```
brian-brazil, line 163: Just node

docs/getting_started.md
brian-brazil: If we're using this metric then mode is useful, but we're not really aggregating anything then as you're included the instance label.

docs/getting_started.md
```diff
@@ -190,35 +189,38 @@ scrape_configs:
...
 ```
-avg(rate(rpc_durations_seconds_count[5m])) by (job, service)
+avg by (job, instance,mode) (rate(node_cpu_seconds_total[5m]))
 ```
 
 Try graphing this expression.
 
 To record the time series resulting from this expression into a new metric
-called `job_service:rpc_durations_seconds_count:avg_rate5m`, create a file
+called `job_instance:node_cpu_usage_ratio:avg_rate5m`, create a file
```
brian-brazil, line 215: This isn't a ratio, there's no division going on.

docs/getting_started.md
```diff
@@ -132,36 +132,35 @@ Experiment with the graph range parameters and other settings.
...
+
+# <version> and <target> will depend on what you downloaded.
+tar -xzvf node_exporter-<version>.<target>.tar.gz
+cd node_exporter-<version>.<target>
 
 # Start 3 example targets in separate terminals:
-./random -listen-address=:8080
-./random -listen-address=:8081
-./random -listen-address=:8082
+./node_exporter --web.listen-address "127.0.0.1:8080"
```
brian-brazil, line 152: Why the `"` ?

docs/getting_started.md
```diff
@@ -190,35 +189,38 @@ scrape_configs:
...
 with the following recording rule and save it as `prometheus.rules.yml`:
 
 ```
 groups:
-- name: example
+- name: cpu-example
   rules:
-  - record: job_service:rpc_durations_seconds_count:avg_rate5m
-    expr: avg(rate(rpc_durations_seconds_count[5m])) by (job, service)
+  - record: job_instance:node_cpu_usage_ratio:avg_rate5m
```
roidelapluie, line 222: @brian-brazil we remove cpu id

brian-brazil: Ah yes, you're right.

docs/getting_started.md
hdost: Valid, then potentially removing the instance? I know for this Quickstart it won't vary but seeing the average utilization across machines. So maybe:
```
- record: job_cluster:node_cpu_usage_ratio:avg_rate5m
  expr: avg by (job, mode) (rate(node_cpu_sections_total[5m]))
```

hdost: I see this has continued in a different thread. Resolving.

docs/getting_started.md
hdost: Keeping mode then :+1:

docs/getting_started.md
```diff
@@ -132,36 +132,35 @@ Experiment with the graph range parameters and other settings.
...
+** If you don't know it's likely `amd64` for 64-bit and either `linux` or `darwin` (macOS).
 
 ```bash
-# Fetch the client library code and compile example.
-git clone https://github.com/prometheus/client_golang.git
-cd client_golang/examples/random
-go get -d
-go build
+# Go to the location of the download.
+cd ~/Downloads
```
brian-brazil, line 145: You're making a presumption that a browser was used and that this is the download location, it's easier to not mention this and presume that the user knows where their files are.

docs/getting_started.md
```diff
@@ -190,35 +181,38 @@ scrape_configs:
 
 Go to the expression browser and verify that Prometheus now has information
 about time series that these example endpoints expose, such as the
-`rpc_durations_seconds` metric.
+`node_cpu_seconds_total` (node_exporter) or `wmi_cpu_time_total`
```
brian-brazil, line 184: Old WMI mention here

docs/getting_started.md
```diff
@@ -190,35 +181,38 @@ scrape_configs:
...
+(wmi_exporter) metric.
 
 ## Configure rules for aggregating scraped data into new time series
 
+If you are using windows you should be able to substitute in
```
brian-brazil, line 189: And here

docs/getting_started.md
```diff
@@ -190,35 +181,38 @@ scrape_configs:
...
 ```
-avg(rate(rpc_durations_seconds_count[5m])) by (job, service)
+avg by (job, instance, mode) (rate(node_cpu_seconds_total[5m]))
 ```
 
 Try graphing this expression.
 
 To record the time series resulting from this expression into a new metric
-called `job_service:rpc_durations_seconds_count:avg_rate5m`, create a file
+called `job_instance:node_cpu_usage:avg_rate5m`, create a file
```
brian-brazil, line 207: Just node_cpu, usage isn't in the metric name. mode is missing from the prefix.

docs/getting_started.md
```diff
@@ -132,36 +132,27 @@ Experiment with the graph range parameters and other settings.
...
-The Go client library includes an example which exports fictional RPC latencies
-for three services with different latency distributions.
-
-Ensure you have the [Go compiler installed](https://golang.org/doc/install) and
-have a [working Go build environment](https://golang.org/doc/code.html) (with
-correct `GOPATH`) set up.
-
-Download the Go client library for Prometheus and run three of these example
-processes:
+The next steps are for `node_exporter` to install follow
```
brian-brazil, line 135: This doesn't parse. I'd go for something like:

The node exporter is used as the example target, for more on using it see...

docs/getting_started.md
```diff
@@ -132,36 +132,27 @@ Experiment with the graph range parameters and other settings.
...
+[these steps.](https://prometheus.io/docs/guides/node-exporter/)
 
 ```bash
-# Fetch the client library code and compile example.
-git clone https://github.com/prometheus/client_golang.git
-cd client_golang/examples/random
-go get -d
-go build
+# <version> and <target> will depend on what you downloaded.
+tar -xzvf node_exporter-<version>.<target>.tar.gz
```
brian-brazil, line 140: You can do * instead of the pseudo-variables, as long as only one version is downloaded it'll do the right thing and is easier for users.

hdost:
I think the latest commit should be the last needed :crossed_fingers:

docs/getting_started.md
```diff
@@ -190,35 +180,34 @@ scrape_configs:
 
 Go to the expression browser and verify that Prometheus now has information
 about time series that these example endpoints expose, such as the
-`rpc_durations_seconds` metric.
+`node_cpu_seconds_total` (node_exporter).
```
brian-brazil, line 183: (node_exporter) is redundant here.

docs/getting_started.md
```diff
@@ -190,35 +180,34 @@ scrape_configs:
...
 ```
-avg(rate(rpc_durations_seconds_count[5m])) by (job, service)
+avg by (job, instance, mode) (rate(node_cpu_seconds_total[5m]))
 ```
 
 Try graphing this expression.
 
 To record the time series resulting from this expression into a new metric
-called `job_service:rpc_durations_seconds_count:avg_rate5m`, create a file
+called `job_instance_mode:node_cpu:avg_rate5m`, create a file
```
brian-brazil, line 202: node_cpu_seconds

We'll get there eventually :)

docs/getting_started.md
```diff
@@ -132,36 +132,26 @@ Experiment with the graph range parameters and other settings.
...
-The Go client library includes an example which exports fictional RPC latencies
-for three services with different latency distributions.
-
-Ensure you have the [Go compiler installed](https://golang.org/doc/install) and
-have a [working Go build environment](https://golang.org/doc/code.html) (with
-correct `GOPATH`) set up.
-
-Download the Go client library for Prometheus and run three of these example
-processes:
+The node_exporter is used as an example target, for more information on using it
```
brian-brazil, line 135: Node Exporter

We're referring to the exporter itself, not the repo.

docs/getting_started.md
```diff
@@ -190,35 +180,34 @@ scrape_configs:
 
 Go to the expression browser and verify that Prometheus now has information
 about time series that these example endpoints expose, such as the
```
brian-brazil, line 182: s/the /

Removing "metric" means the the is no longer required.

brian-brazil:
Thanks!

PR #7194, opened 2020-05-02
cleanup: remove unnacessary nil check before range

beautytiger:
Map or slice nil check before range is not necessary, so let's remove the if..nil check.


bwplotka:
Thanks!

PR #7199, opened 2020-05-04
Add funcbench workflow

nevill:
This is GitHub workflow config for funcbench.
Please review it @krasi-georgiev @geekodour .

### DIFF
--- a/.github/workflows/funcbench.yml
+++ b/.github/workflows/funcbench.yml
@@ -1,0 +1,57 @@
+on: repository_dispatch
+name: Funcbench Workflow
+jobs:
+  run_funcbench:
+    name: Running funcbench
+    if: github.event.action == 'funcbench_start'
+    runs-on: ubuntu-latest
+    env:
+      AUTH_FILE: ${{ secrets.TEST_INFRA_GKE_AUTH }}
+      BRANCH: ${{ github.event.client_payload.BRANCH }}
+      BENCH_FUNC_REGEX: ${{ github.event.client_payload.BENCH_FUNC_REGEX }}
+      CLUSTER_NAME: test-infra
+      GITHUB_TOKEN: ${{ secrets.PROMBOT_TOKEN }}
+      GITHUB_ORG: prometheus
+      GITHUB_REPO: prometheus
+      GITHUB_STATUS_TARGET_URL: https://github.com/${{github.repository}}/actions/runs/${{github.run_id}}
+      LAST_COMMIT_SHA: ${{ github.event.client_payload.LAST_COMMIT_SHA }}
+      PROJECT_ID: macro-mile-203600
+      PR_NUMBER: ${{ github.event.client_payload.PR_NUMBER }}
+      ZONE: europe-west3-a
+    steps:
+    - name: Update status to pending
+      run: >-
+        curl -i -X POST
+        -H "Authorization: Bearer $GITHUB_TOKEN"
+        -H "Content-Type: application/json"
+        --data '{"state":"pending","context":"funcbench-status","target_url":"'$GITHUB_STATUS_TARGET_URL'"}'
+        "https://api.github.com/repos/$GITHUB_REPOSITORY/statuses/$LAST_COMMIT_SHA"
+    - name: Prepare nodepool
+      uses: docker://prominfra/prombench:master
+      with:
+        args: make funcbench_nodepool_create
+    - name: Run funcbench
+      uses: docker://prominfra/prombench:master
+      with:
+        args: make funcbench_resource_apply
+    - name: Delete all resources
+      if: always()
+      uses: docker://prominfra/prombench:master
+      with:
+        args: make funcbench_resource_delete; make funcbench_nodepool_delete
+    - name: Update status to failure
+      if: failure()
+      run: >-
+        curl -i -X POST
+        -H "Authorization: Bearer $GITHUB_TOKEN"
+        -H "Content-Type: application/json"
+        --data '{"state":"failure","context":"funcbench-status","target_url":"'$GITHUB_STATUS_TARGET_URL'"}'
+        "https://api.github.com/repos/$GITHUB_REPOSITORY/statuses/$LAST_COMMIT_SHA"
+    - name: Update status to success
+      if: success()
+      run: >-
+        curl -i -X POST
+        -H "Authorization: Bearer $GITHUB_TOKEN"
+        -H "Content-Type: application/json"
+        --data '{"state":"success","context":"funcbench-status","target_url":"'$GITHUB_STATUS_TARGET_URL'"}'
+        "https://api.github.com/repos/$GITHUB_REPOSITORY/statuses/$LAST_COMMIT_SHA"


.github/workflows/funcbench.yml
```diff
@@ -0,0 +1,56 @@
...
+jobs:
+  run_funcbench:
+    name: Running funcbench
+    if: github.event.action == 'funcbench_start'
+    runs-on: ubuntu-latest
+    env:
+      AUTH_FILE: ${{ secrets.PROMBENCH_GKE_AUTH }}
+      BRANCH: ${{ github.event.client_payload.BRANCH }}
+      BENCH_FUNC_REGEX: ${{ github.event.client_payload.BENCH_FUNC_REGEX }}
+      CLUSTER_NAME: prombench
```
krasi-georgiev, line 12:
```suggestion
      CLUSTER_NAME: funcbench
```

I think best to keep those separate.

OTOH would this allow running more than a single funcbench on different PRs? IIRC in `funcbench_nodepool_create` it will create a new node pool for sufficed with the PR number.

.github/workflows/funcbench.yml
```diff
@@ -0,0 +1,56 @@
...
+      GITHUB_TOKEN: ${{ secrets.PERSONAL_TOKEN }}
+      GITHUB_ORG: prometheus
+      GITHUB_REPO: prometheus
+      LAST_COMMIT_SHA: ${{ github.event.client_payload.LAST_COMMIT_SHA }}
```
krasi-georgiev, line 16: can't this be fetched using `github.event.client_payload.PR_NUMBER` ? I mean once you know the PR number make a query to find out the latest commit she of this PR. This will remove the need for this variable.

.github/workflows/funcbench.yml
```diff
@@ -0,0 +1,56 @@
...
+        "https://api.github.com/repos/$GITHUB_REPOSITORY/statuses/$LAST_COMMIT_SHA"
+    - name: Prepare nodepool
+      uses: docker://prominfra/prombench:master
+      with:
+        args: make funcbench_nodepool_create
+    - name: Run funcbench
+      uses: docker://prominfra/prombench:master
+      with:
+        args: make funcbench_resource_apply
+    - name: Recycle all the resources
```
krasi-georgiev, line 36:
```suggestion
    - name: Delete all resources
```

.github/workflows/funcbench.yml
```diff
@@ -0,0 +1,56 @@
...
+    steps:
+    - name: Update status to pending
+      run: >-
+        curl -i -X POST
+        -H "Authorization: Bearer $GITHUB_TOKEN"
+        -H "Content-Type: application/json"
+        --data '{"state":"pending","context":"funcbench-status","target_url":"https://github.com/'$GITHUB_REPOSITORY'/actions"}'
+        "https://api.github.com/repos/$GITHUB_REPOSITORY/statuses/$LAST_COMMIT_SHA"
+    - name: Prepare nodepool
+      uses: docker://prominfra/prombench:master
```
krasi-georgiev, line 29: I think we should keep prombench and funcbench completely separate and [added a follow up issue](https://github.com/prometheus/test-infra/issues/356) for this. This refactoring  will take some time so let's not block this PR and will refactor as a follow up.

.github/workflows/funcbench.yml
nevill: `LAST_COMMIT_SHA` is used for setting the commit status to Pending, to prevent it for being merged, like DCO check.
If we don't need this check, we can get rid of it.

.github/workflows/funcbench.yml
nevill: Yes, it allows funcbench run for different PRs at the same time. We have different namespaces / nodepools for different PRs.
But if we launch `funcbench` for the same PR twice at the same time, it will have an issue. The one which finishes firstly will always interrupt the second one.

krasi-georgiev: I think this was somehow solved in prombench so worth checking with @geekodour

.github/workflows/funcbench.yml
krasi-georgiev: I think this is not to prevent a merge, but more for showing at the bottom of the page that there is a funcbnech running and if I am correct we need this.

.github/workflows/funcbench.yml
geekodour: currently, prombench gh action workflow will just keep showing status pending till the currently running test does not get over. It concludes that the running test is over if all the nodepools in `nodepools.yaml` are deleted.

One could cancel the workflow in the meantime, otherwise, the test will start once the previous test gets completely canceled.

geekodour: > `CLUSTER_NAME: funcbench`

This will mean we'll need to set up another cluster (i.e `funcbench` will have its own `cluster.yml`), what about renaming the current cluster to `test-infra`, and then it'll be intuitive to add different test-infra components to it. Otherwise we can also have funcbench spin up a cluster every time it's run as well but then again will have to take care of subsequent `/funcbench` command in the same PR slightly differently than what `prombench` does currently.

.github/workflows/funcbench.yml
geekodour: `LAST_COMMIT_SHA` is fetched from the pr number only but here it is done by comment-monitor  when it sees a comment as it is almost always required if we want to enable Github status towards the bottom of the PR

.github/workflows/funcbench.yml
krasi-georgiev: if using the same cluster simplifies the setup yes, lets rename it to test-infra.

.github/workflows/funcbench.yml
```diff
@@ -0,0 +1,56 @@
...
+  run_funcbench:
+    name: Running funcbench
+    if: github.event.action == 'funcbench_start'
+    runs-on: ubuntu-latest
+    env:
+      AUTH_FILE: ${{ secrets.PROMBENCH_GKE_AUTH }}
+      BRANCH: ${{ github.event.client_payload.BRANCH }}
+      BENCH_FUNC_REGEX: ${{ github.event.client_payload.BENCH_FUNC_REGEX }}
+      CLUSTER_NAME: test-infra
+      GITHUB_TOKEN: ${{ secrets.PERSONAL_TOKEN }}
```
geekodour, line 13:
```suggestion
      GITHUB_TOKEN: ${{ secrets.PROMBOT_TOKEN }}
```

.github/workflows/funcbench.yml
```diff
@@ -0,0 +1,56 @@
+on: repository_dispatch
+name: Funcbench Workflow
+jobs:
+  run_funcbench:
+    name: Running funcbench
+    if: github.event.action == 'funcbench_start'
+    runs-on: ubuntu-latest
+    env:
+      AUTH_FILE: ${{ secrets.PROMBENCH_GKE_AUTH }}
```
geekodour, line 9:
```suggestion
      AUTH_FILE: ${{ secrets.TEST_INFRA_GKE_AUTH }}
```

geekodour: if #7214  gets merged.

PR #7206, opened 2020-05-05
Trace Remote Write requests

cboggs:
Continuing the recent work on pulling in Jaeger and instrumenting remote queries, we figured we would also instrument remote writes.

This currently only spits out one root span for each call to `sendSamples`, and is mostly valuable as a starting point for upstream systems receiving the sample batches that may have their own tracing in place.

Kept it simple and didn't use the pattern in `utils/query_stats.go` since it seemed overkill for the simple case of remote writes. Also, we didn't instrument `remote.Client.Store` because remote writes can retry forever, and we end up with potentially infinite spans under a root-less trace (since the root span never finishes).

Screenshot from 2020-05-05 14-12-13

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -25,6 +25,8 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/gogo/protobuf/proto"
 	"github.com/golang/snappy"
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/config"
@@ -844,36 +846,66 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
 		return err
 	}
 
+	try := 0
+
 	for {
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
 		default:
 		}
-		begin := time.Now()
-		err := s.qm.client().Store(ctx, req)
 
-		s.qm.metrics.sentBatchDuration.Observe(time.Since(begin).Seconds())
-
+		err = s.attemptSendSamples(ctx, samples, buf, highest, try)
 		if err == nil {
-			s.qm.metrics.succeededSamplesTotal.Add(float64(len(samples)))
-			s.qm.metrics.bytesSent.Add(float64(len(req)))
-			s.qm.metrics.highestSentTimestamp.Set(float64(highest / 1000))
 			return nil
 		}
 
-		if _, ok := err.(recoverableError); !ok {
-			return err
-		}
-		s.qm.metrics.retriedSamplesTotal.Add(float64(len(samples)))
-		level.Warn(s.qm.logger).Log("msg", "Failed to send batch, retrying", "err", err)
-
+		// If we make it this far, we've encountered a recoverable error and will retry.
 		time.Sleep(time.Duration(backoff))
 		backoff = backoff * 2
 		if backoff > s.qm.cfg.MaxBackoff {
 			backoff = s.qm.cfg.MaxBackoff
 		}
+
+		try++
 	}
+}
+
+func (s *shards) attemptSendSamples(ctx context.Context, samples []prompb.TimeSeries, req *[]byte, highest int64, try int) error {
+	req_size := len(*req)
+	sample_count := len(samples)
+
+	span, ctx := opentracing.StartSpanFromContext(ctx, "Remote Send Batch")
+	defer span.Finish()
+
+	span.SetTag("samples", sample_count)
+	span.SetTag("requestSize", req_size)
+	span.SetTag("try", try)
+	span.SetTag("remote_name", s.qm.storeClient.Name())
+	span.SetTag("remote_url", s.qm.storeClient.Endpoint())
+
+	begin := time.Now()
+	err := s.qm.client().Store(ctx, *req)
+
+	s.qm.metrics.sentBatchDuration.Observe(time.Since(begin).Seconds())
+
+	if err == nil {
+		s.qm.metrics.succeededSamplesTotal.Add(float64(sample_count))
+		s.qm.metrics.bytesSent.Add(float64(req_size))
+		s.qm.metrics.highestSentTimestamp.Set(float64(highest / 1000))
+		return nil
+	}
+
+	if _, ok := err.(recoverableError); !ok {
+		span.LogKV("error", err)
+		ext.Error.Set(span, true)
+		return err
+	}
+	s.qm.metrics.retriedSamplesTotal.Add(float64(sample_count))
+	span.LogKV("error", err)
+	ext.Error.Set(span, true)
+	level.Warn(s.qm.logger).Log("msg", "Failed to send batch, retrying", "err", err)
+	return err
 }
 
 func buildWriteRequest(samples []prompb.TimeSeries, buf []byte) ([]byte, int64, error) {
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -888,7 +888,6 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
 		if err != nil {
 			// If the error is unrecoverable, we should not retry.
 			if _, ok := err.(recoverableError); !ok {
-				level.Error(s.qm.logger).Log("msg", "Failed to send batch, unrecoverable error, will not retry", "err", err)
 				return err
 			}
 
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -837,28 +837,28 @@ func (s *shards) sendSamples(ctx context.Context, samples []prompb.TimeSeries, b
 
 // sendSamples to the remote storage with backoff for recoverable errors.
 func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.TimeSeries, buf *[]byte) error {
-	backoff := s.qm.cfg.MinBackoff
 	req, highest, err := buildWriteRequest(samples, *buf)
-	reqSize := len(*buf)
-	sampleCount := len(samples)
-	*buf = req
 	if err != nil {
 		// Failing to build the write request is non-recoverable, since it will
 		// only error if marshaling the proto to bytes fails.
 		return err
 	}
 
+	backoff := s.qm.cfg.MinBackoff
+	reqSize := len(*buf)
+	sampleCount := len(samples)
+	*buf = req
 	try := 0
 
-	// This anonymous function is used to allow us to defer the finishing of our per-try span
-	// without causing a memory leak, and it has the nice effect of avoiding propagating most of
-	// the parameters for sendSamplesWithBackoff/3.
+	// An anonymous function allows us to defer the completion of our per-try spans
+	// without causing a memory leak, and it has the nice effect of not propagating any
+	// parameters for sendSamplesWithBackoff/3.
 	attemptStore := func() error {
 		span, ctx := opentracing.StartSpanFromContext(ctx, "Remote Send Batch")
 		defer span.Finish()
 
 		span.SetTag("samples", sampleCount)
-		span.SetTag("requestSize", reqSize)
+		span.SetTag("request_size", reqSize)
 		span.SetTag("try", try)
 		span.SetTag("remote_name", s.qm.storeClient.Name())
 		span.SetTag("remote_url", s.qm.storeClient.Endpoint())
@@ -870,9 +870,9 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
 		if err != nil {
 			span.LogKV("error", err)
 			ext.Error.Set(span, true)
-			level.Warn(s.qm.logger).Log("msg", "Failed to send batch, retrying", "err", err)
 			return err
 		}
+
 		return nil
 	}
 
@@ -888,13 +888,16 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
 		if err != nil {
 			// If the error is unrecoverable, we should not retry.
 			if _, ok := err.(recoverableError); !ok {
+				level.Error(s.qm.logger).Log("msg", "Failed to send batch, unrecoverable error, will not retry", "err", err)
 				return err
 			}
 
 			// If we make it this far, we've encountered a recoverable error and will retry.
 			s.qm.metrics.retriedSamplesTotal.Add(float64(sampleCount))
+			level.Warn(s.qm.logger).Log("msg", "Failed to send batch, retrying", "err", err)
 			time.Sleep(time.Duration(backoff))
 			backoff = backoff * 2
+
 			if backoff > s.qm.cfg.MaxBackoff {
 				backoff = s.qm.cfg.MaxBackoff
 			}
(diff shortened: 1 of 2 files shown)


roidelapluie:
I do not get why not the remote.Client.Store. Can you elaborate?

This means that this PR can create a span that never finishes?

cboggs:
Yes, this can potentially end up with a never-finishing span, such that it simply never gets emitted. That said, the (seemingly) more likely scenarios are:
1. unrecoverable errors (we do get a span for these)
1. context cancelation (Prom is stopped in the middle of a retry loop or something, we get spans here too)
1. Remote storage becomes available again, and the span eventually finishes once the request is complete (and we track retries in the resulting span)

Technically, a "healthy" Prometheus can end up in an infinite loop here (independent of this PR), and thus the span we start will never finish. That said, it seems unlikely to happen often, and there doesn't seem to be any data of value that is lost in that event. :-)

roidelapluie:
I think that if we are stuck in an infinite loop we DO want the remote write spans, so I would properly instrument  remote.Client.Store; that would also give us the address, the return code, etc

csmarchbanks:
Doing a bit of math, if we instrumented `client.Store()`, and the remote went down for say 15 minutes the following would happen:
* Retry the request every 100ms (default max backoff)
* Requests retried ` = 10 req/s * 15 minutes = 9000 requests`
* If we use nethttp that would be two spans per retry

Thus we would have a single trace with 18,000 spans underneath it if `client.Store()` is instrumented. That is not particularly helpful, and if the remote store was down for longer it could get  really bad. Personally, I think the retry counter is preferable.

Alternatively we could limit the number of retries where spans are created but that would be more invasive, and confusing for anyone not intimately familiar with that code.

csmarchbanks:
We will also have logs with that information for each retry including the error code so if debugging an issue the information is still available and not as spammy due to the use of a dedupe logger.

roidelapluie:
OK;

I would still like in this PR more tags related to the remote write, such as the remote write name.

roidelapluie:
Do we want to send the spanId to the outgoing call?

csmarchbanks:
:+1: to adding the remote write name as a tag and wiring up the proper headers to the outgoing call. Then we could add a comment with why we aren't creating sub-spans in `client.Store()` as well.

roidelapluie:
mmh what are the risk that we tell the other side to also do 18000 spans?

csmarchbanks:
That could definitely happen. Alternatively we could trace each retry separately and let sampling handle it? I.e. Keep the code as it is to have information at the batch level, but add an additional, non-connected, span for each retry. That's a pattern I haven't seen before, but what do both of you think?

cboggs:
I'm not wild about the idea of doing a batch-level trace discretely from a "re/try"-level trace.

If anything I'd propose we just do the per-try traces and let sampling handle the "uh oh" factor of many many retries. :-)

Those of us without sampling will have to figure something out, though.

csmarchbanks:
I'd be good with a per-try trace level. Perhaps start the span around 
```
err := s.qm.client().Store(ctx, req)
```
in `sendSamplesWithBackoff()` so that we can still get the tags about number of samples, and which retry it is?

cboggs:
I think that'd be just fine. :-)

storage/remote/queue_manager.go
```diff
@@ -927,6 +925,7 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
 		if backoff > s.qm.cfg.MaxBackoff {
 			backoff = s.qm.cfg.MaxBackoff
 		}
+		span.Finish()
```
csmarchbanks, line 928: What about the `return` cases?

storage/remote/queue_manager.go
```diff
@@ -887,14 +884,15 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
...
-
 	retries := 0
-	// Anonymous function is required so that retries value is not captured as zero.
-	defer func() { span.SetTag("retries", retries) }()
 
 	for {
+		span, ctx := opentracing.StartSpanFromContext(ctx, "Remote Send Batch")
+		span.SetTag("samples", len(samples))
+		span.SetTag("requestSize", len(req))
+		span.SetTag("retries", retries)
```
csmarchbanks, line 893: Rather than `retries` could this be something like `try`? If I see a `retries` tag in a trace I would expect all of those retries to be part of that one trace.

storage/remote/queue_manager.go
```diff
@@ -887,14 +884,15 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
...
+		span.SetTag("remoteStoreName", s.qm.storeClient.Name())
```
csmarchbanks, line 894: We use `remote_name` in the metrics, it would be nice to be consistent. Perhaps let's also do `snake_case` for all tags?

cstyan: we should maybe include remote url as well since name might not always be useful (user hasn't set a name in the config, so it defaults to the hash)?

cboggs:
Yea, I just pushed this on as a quick WIP commit to get it safely off my
machine. Had some work things come up that I needed to take care of
instead, will revisit this shortly!

storage/remote/queue_manager.go
cboggs: Agreed.

storage/remote/queue_manager.go
cboggs: Yup, incoming!

storage/remote/queue_manager.go
cboggs: Good call. :-)

cboggs:
Screenshots below don't have the update from `retries` to `try`, but the rest should be sane!
trace-remote-write-good
trace-remote-write-error

storage/remote/queue_manager.go
csmarchbanks: This works now, but I don't love that we have to remember to put the `span.Finish()` calls in so many places. Would it be worth breaking the actual try into a function, possibly even an anonymous function then we could use defer?

storage/remote/queue_manager.go
```diff
@@ -844,9 +846,21 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
 		return err
 	}
 
+	retries := 0
+
 	for {
+		span, ctx := opentracing.StartSpanFromContext(ctx, "Remote Send Batch")
```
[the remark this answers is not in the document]
cboggs: Fair point.

cboggs:
I like factoring out the try logic entirely, that avoids the hazard of `defer`ing inside a loop.

cboggs:
@csmarchbanks I think I managed to retain the performance characteristics of the `sendSamples*` bits (reuse of the `*[]byte` buffer), but let me know if I goofed.

storage/remote/queue_manager.go
```diff
@@ -844,36 +846,66 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
...
+	s.qm.metrics.sentBatchDuration.Observe(time.Since(begin).Seconds())
+
+	if err == nil {
+		s.qm.metrics.succeededSamplesTotal.Add(float64(sampleCount))
+		s.qm.metrics.bytesSent.Add(float64(reqSize))
+		s.qm.metrics.highestSentTimestamp.Set(float64(highest / 1000))
+		return nil
+	}
+
+	if _, ok := err.(recoverableError); !ok {
```
csmarchbanks, line 899: This check should be in the other loop still I think. We need to continue to loop for recoverable errors, but return for un-recoverable. I think right now we would also retry un-recoverable errors?

storage/remote/queue_manager.go
```diff
@@ -844,36 +846,66 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
...
+		span.LogKV("error", err)
```
csmarchbanks, line 900: You only need to set these two lines once right after the `err == nil` check. No matter what after that point  there was an error.

storage/remote/queue_manager.go
cboggs: Oops, good catch! Will fix.

storage/remote/queue_manager.go
cboggs: :+1:

storage/remote/queue_manager.go
cboggs: Better.
Screenshot from 2020-05-15 08-13-20
Screenshot from 2020-05-15 08-12-27

storage/remote/queue_manager.go
```diff
@@ -844,38 +846,70 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
...
 		case <-ctx.Done():
 			return ctx.Err()
 		default:
 		}
-		begin := time.Now()
-		err := s.qm.client().Store(ctx, req)
-
-		s.qm.metrics.sentBatchDuration.Observe(time.Since(begin).Seconds())
 
+		err = s.attemptSendSamples(ctx, samples, buf, highest, try)
```
csmarchbanks, line 858: Let's pass down `req` here rather than `buf`. `buf` is just for allocation optimizations and it is nice to keep its scope as limited as possible.

storage/remote/queue_manager.go
```diff
@@ -844,38 +846,70 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
...
+	s.qm.metrics.sentBatchDuration.Observe(time.Since(begin).Seconds())
+
+	if err == nil {
+		s.qm.metrics.succeededSamplesTotal.Add(float64(sampleCount))
+		s.qm.metrics.bytesSent.Add(float64(reqSize))
+		s.qm.metrics.highestSentTimestamp.Set(float64(highest / 1000))
+		return nil
+	}
+
+	s.qm.metrics.retriedSamplesTotal.Add(float64(sampleCount))
```
csmarchbanks, line 904: This metric should be incremented in the for loop still. This may or may not be a retry-able failure.

storage/remote/queue_manager.go
```diff
@@ -844,38 +846,70 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
...
+
+	span.LogKV("error", err)
+	ext.Error.Set(span, true)
+
+	level.Warn(s.qm.logger).Log("msg", "Failed to send batch, retrying", "err", err)
```
csmarchbanks, line 909: This log line should also be in the for loop as we don't know if it will be retried or not.

storage/remote/queue_manager.go
```diff
@@ -844,38 +846,70 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
...
+	span.SetTag("try", try)
+	span.SetTag("remote_name", s.qm.storeClient.Name())
+	span.SetTag("remote_url", s.qm.storeClient.Endpoint())
+
+	begin := time.Now()
+	err := s.qm.client().Store(ctx, *req)
+
+	s.qm.metrics.sentBatchDuration.Observe(time.Since(begin).Seconds())
+
+	if err == nil {
```
csmarchbanks, line 897: Now that this is in a different function, could we invert the logic to the more normal `if err != nil` flow?

storage/remote/queue_manager.go
```diff
@@ -844,38 +846,70 @@ func (s *shards) sendSamplesWithBackoff(ctx context.Context, samples []prompb.Ti
...
 		backoff = backoff * 2
 		if backoff > s.qm.cfg.MaxBackoff {
 			backoff = s.qm.cfg.MaxBackoff
 		}
+
+		try++
 	}
 }
 
+func (s *shards) attemptSendSamples(ctx context.Context, samples []prompb.TimeSeries, req *[]byte, highest int64, try int) error {
```
csmarchbanks, line 879: I don't love having to pass down `samples` or `highest`. Would doing this as an anonymous function fix that? Is that any better?

storage/remote/queue_manager.go
cboggs: Yea, copy/paste got away from me there. :-D

storage/remote/queue_manager.go
cboggs: Oh snap, correct.

storage/remote/queue_manager.go
cboggs: Definitely did not pay attention to the metric name, oops!

storage/remote/queue_manager.go
cboggs: :+1: Sounds good, I was 50/50 on that one.

storage/remote/queue_manager.go
cboggs: Hmm... Anonymous is fine for removing the args, since they'll all be in scope. Will need a clear comment to explain why on earth it's helpful, which is also fine. :-)

I think it'll make it clearer (in my head, at least) which parts need to be in the attempt and which parts need to be in the enclosing loop.

cboggs:
Ok, seems that the latest commit still works as intended, but yields sane metrics. Also verified that I wasn't going to hit some odd `defer` behavior by still having it inside a loop, despite it being inside an anonymous function. Was mostly concerned about a memory leak, but after an hour's worth of retries and such, everything was nice and stable.

storage/remote/queue_manager.go
```diff
@@ -837,42 +839,75 @@ func (s *shards) sendSamples(ctx context.Context, samples []prompb.TimeSeries, b
...
+		span.SetTag("remote_url", s.qm.storeClient.Endpoint())
+
+		begin := time.Now()
+		err := s.qm.client().Store(ctx, *buf)
+		s.qm.metrics.sentBatchDuration.Observe(time.Since(begin).Seconds())
+
+		if err != nil {
+			span.LogKV("error", err)
+			ext.Error.Set(span, true)
+			level.Warn(s.qm.logger).Log("msg", "Failed to send batch, retrying", "err", err)
```
csmarchbanks, line 873: This log line is still in the incorrect place. It should be with the `retriedSamplesTotal.Add()` metric call.

storage/remote/queue_manager.go
```diff
@@ -837,42 +839,75 @@ func (s *shards) sendSamples(ctx context.Context, samples []prompb.TimeSeries, b
...
+
+	// This anonymous function is used to allow us to defer the finishing of our per-try span
+	// without causing a memory leak, and it has the nice effect of avoiding propagating most of
+	// the parameters for sendSamplesWithBackoff/3.
+	attemptStore := func() error {
+		span, ctx := opentracing.StartSpanFromContext(ctx, "Remote Send Batch")
+		defer span.Finish()
+
+		span.SetTag("samples", sampleCount)
+		span.SetTag("requestSize", reqSize)
```
csmarchbanks, line 861: `snake_case` here too please for consistency.

storage/remote/queue_manager.go
```diff
@@ -835,44 +837,80 @@ func (s *shards) sendSamples(ctx context.Context, samples []prompb.TimeSeries, b
...
-		if err == nil {
-			s.qm.metrics.succeededSamplesTotal.Add(float64(len(samples)))
-			s.qm.metrics.bytesSent.Add(float64(len(req)))
-			s.qm.metrics.highestSentTimestamp.Set(float64(highest / 1000))
-			return nil
-		}
+		if err != nil {
+			// If the error is unrecoverable, we should not retry.
+			if _, ok := err.(recoverableError); !ok {
+				level.Error(s.qm.logger).Log("msg", "Failed to send batch, unrecoverable error, will not retry", "err", err)
```
csmarchbanks, line 891: This log line covers the same information as what is on line 827 on your branch:
```
level.Error(s.qm.logger).Log("msg", "non-recoverable error", "count", len(samples), "err", err)
```
I would say remove this one and only log in the other place, as we also need to log errors when building requests and such.

csmarchbanks:
CI failure looks to be from the quay.io outage, but running once more to make sure. If all looks well I will merge this tomorrow.

Issue #5317, opened 2019-03-08
util/promlint: consider making it easier for 3rd parties to import


mdlayher:
## Proposal

I originally added `promlint` for use in `promtool check metrics`. It is very useful for that purpose, however, it would also be nice to allow the package to be easily vendored and used in other applications for unit tests and such.

Because this main Prometheus repository is fairly large and also contains a great deal of unrelated code, I propose moving `util/promlint` to `github.com/prometheus/common/promlint` or similar. This way, library code lives with a library instead of an entire application, and is much easier to vendor.

At that point, we can vendor in the updated copy of `common` and switch the import statements used within `promtool`.

mdlayher:
Talked with @SuperQ on IRC and it appears I was wrong about needing to vendor the entire repo. `dep` at least appears to pick out the package by itself if I `dep ensure -add github.com/prometheus/prometheus/util/promlint`.

I do still think it'd be reasonable to move this to common, though.

brian-brazil:
It's only used in one of our repos, so I see no need to move it to common. Common is for stuff shared across our repos.

RainbowMango:
If `github.com/prometheus/common
` isn't the proper place, then how about `github.com/prometheus/client_golang`? 

Maybe we can move it to `client_golang/prometheus/testutil`.

Involve @brancz here.

mcristina422:
@brian-brazil @mdlayher Thoughts on moving promlint into http://github.com/prometheus/client_golang like @RainbowMango suggested? 

Or maybe even http://github.com/prometheus/promlint ? Moving it out of `github.com/prometheus/prometheus` would be extremely helpful

brian-brazil:
As indicated in https://github.com/kubernetes/kubernetes/pull/86477 you'd probably be best using it is a library. As-is the code appropriately located from a Prometheus standpoint, as it is only used in promtool.

RichiH:
We are currently doing a bug scrub.

@beorn is considering to move this to client_golang. @brancz would you be interested in doing that?

beorn7:
To be precise: `client_golang/prometheus/testutil` could make sense as a location.

beorn7:
In particular, client_golang would not be a good place for a stand-alone linter binary, but `promtool` can stay where it is and just import `client_golang/prometheus/testutil` to get the code.

RainbowMango:
I'm interested in this, can I join or take this?
@RichiH @beorn7 @brancz

brancz:
[...]

PR #7209, opened 2020-05-06
switch promlint to client_golang

RainbowMango:
Fixes https://github.com/prometheus/prometheus/issues/5317

This PR switched the implementation of `promlint` with [client_golang promlint](https://github.com/prometheus/client_golang/tree/master/prometheus/testutil/promlint).

~~In consideration of the compatibility, we can't change any exported types and functions.~~

**Does this PR introduce a user-facing change?**:
```release-note
The package `github.com/prometheus/prometheus/util/promlint` has been deprecated, please switch to github.com/prometheus/client_golang/prometheus/testutil/promlint for a replacement.
```


util/promlint/promlint.go
```diff
@@ -12,17 +12,14 @@
...
 import (
-	"fmt"
 	"io"
-	"regexp"
-	"sort"
-	"strings"
 
-	dto "github.com/prometheus/client_model/go"
-	"github.com/prometheus/common/expfmt"
+	"github.com/prometheus/client_golang/prometheus/testutil/promlint"
```
brian-brazil, line 22: Can we not delete this entire file?

roidelapluie: My assuption was that we would replace  github.com/prometheus/prometheus/testutil/promlint by github.com/prometheus/client_golang/prometheus/testutil/promlint in our codebase indeed.

RainbowMango: > Can we not delete this entire file?

Do you mean remove the package directly? My concern is it will violate the compatibility practice.

brian-brazil: Yes, remove the entire package. There's no compatibility guarantees in this repository at the code level.

RainbowMango: To be honest, it's a shock to me.
But, I'm ok, if there are no compatibility guarantees.
cc @beorn7 @brancz

roidelapluie: I agree with brian, that is why we moved to client_golang, to provide those guarantees

beorn7: @RainbowMango so that's the problem with what semantic versioning refers to. The release versions for prometheus/prometheus refer to the behavior of the Prometheus binary (and promtool), not to code-level interfaces. That's different in prometheus/client_golang, which is a true library. Of course, many people use packages from prometheus/prometheus as libraries, and Go modules just assume the version tags in prometheus/prometheus refer to code level compatibility, which is simply not the case.

We discussed the problem a lot. There are several possible ways out of it. We haven't really settled on any of them yet, we haven't even decided if that's a problem that should be tackled at all, but https://github.com/modularise/modularise is a quite elaborate approach which also offers extracted versions of some prometheus/prometheus packages typically used externally:
- https://github.com/modularise/prometheus-tsdb
- https://github.com/modularise/prometheus-promql
- https://github.com/modularise/prometheus-discovery

RainbowMango: Copy that! Thanks for your information.

brian-brazil:
:+1:

RainbowMango:
Hi, I pushed a new commit yesterday, addressed all comments I think. So please take a look.

brian-brazil:
Thanks!


I see no disagreement, so merging.

PR #7210, opened 2020-05-06
Alibaba/ali ecs sd fix 20200506

KeyOfSpectator:
fix queryFromListTagResources bug when ECS with labels.


brian-brazil:
Can you explain what you're trying to do with this PR?

Note that it's not practical to have an ECS SD last we looked, the API quota limits are too low.

roidelapluie:
We have looked at this in the bug scrub and decided to close it based on the above comment. Thanks!

Issue #7107, opened 2020-04-09
Service Discovery: Reduce number of packages that are dependent on our SD implementations (e.g K8s, Azure) to just main.


bwplotka:
## Proposal

Prometheus has official support for numerous discovery mechanisms by default.

Currently, those are in the form of separate packages: https://github.com/prometheus/prometheus/tree/0730d6eb74a2a6e268260a800f5de90958d0253a/discovery
Each of those implementations is pretty specific and depends on large (very often larger than Prometheus itself) dependant Go modules (e.g Azure, Google, AWS).

**That causes a few issues:**

* Big Prometheus binary size (not a big problem, but causes some issues for rate users who run Prom within IoT, e.g. in their fridge)
* It is blocking the reusability of many packages. Currently, many packages depend on `config` -> depends on discovery stuff. This means that many packages are affected by huge dependency baggage. This blocks external projects e.g Thanos or Cortex to reuse such packages does not want to have extra dependencies that cause de hell (e.g mismatching versions). 

All of the above also results in a blocker to extend official ones with custom SD plugins (e.g for docker).

Unfortunately on top of this there is overall go module problems described many times.

**The overall first step, we should think about now, is how we can reduce the number of packages that depend on SD implementation to just one (main).**

### Prometheus Discovery package dependency Inspection

`goda graph -short github.com/prometheus/prometheus/cmd/prometheus | dot -Tsvg -o graph.svg`

NOTE: It's too big to render readable PNG Only SVG is usable. But just look on this! 

![ezgif com-gif-maker (2)](https://user-images.githubusercontent.com/6950331/78891355-554fdb00-7a5f-11ea-912c-48ba7cedabc5.png)

Most of it is Kubernetes, so after we ignore k8s to see anything and focusing on discovery:

`goda graph -short github.com/prometheus/prometheus/cmd/prometheus - github.com/prometheus/prometheus/discovery/kubernetes | dot -Tsvg -o smgraph.svg`

![ezgif com-crop (4)](https://user-images.githubusercontent.com/6950331/78892753-d90ac700-7a61-11ea-829a-14be5c25fbf0.png)

As we see there are 3 links we might want to change as part of this work:

* [ ] `main` -> `discovery` -> `discovery/config` -> SD implementations
* [ ] `main` -> `discovery/config` -> SD implementations
* [ ] `main` -> `config` ->  `discovery/config` -> SD implementations

[...]

PR #7226, opened 2020-05-07
WIP - Make consul optional

roidelapluie:
Initial work on #7107


discovery/consul/consul.go
```diff
@@ -68,6 +69,10 @@ const (
 	namespace = "prometheus"
 )
 
+func init() {
+	discovery.Register("consul_sd_configs", SDConfig)
```
bwplotka, line 73: Why we need this? This feels wrong.

PR #7229, opened 2020-05-08
Snapshot in-memory chunks on shutdown for faster restarts

codesome:
This PR is based on this design doc: https://docs.google.com/document/d/1c90BmhErybHH94hoGNsgJC9pyaUTU3dgHoyD3dm6Kw4/edit?usp=sharing (the doc is a little outdated after some discussions here)

Snapshot happens during shutdown and not at regular intervals to avoid write amplification.

The snapshot name is of the format `chunk_snapshot.<last WAL segment>.<offset in the segment>`. The offset in the WAL segment gives us more precision on where to continue the snapshot and avoids unnecessary replay of entire WAL segment.

**Format of the snapshot**
Starts with series records. Each series gets 1 record.

```
┌────────────────────────────┬────────────────────────────┐
│      Record Type <byte>    │    Series Ref <uint64>     │
├────────────────────────────┴────────────────────────────┤
│                Number of  Labels <uvarint>              │
├────────────────────────────┬────────────────────────────┤
│    len(name_1) <uvarint>   │      name_1 <bytes>        │
├────────────────────────────┼────────────────────────────┤
│    len(val_1) <uvarint>    │      val_1 <bytes>         │
├────────────────────────────┴────────────────────────────┤
│                          . . .                          │
├────────────────────────────┬────────────────────────────┤
│    len(name_N) <uvarint>   │      name_N <bytes>        │
├────────────────────────────┼────────────────────────────┤
│    len(val_N) <uvarint>    │      val_N <bytes>         │
├────────────────────────────┴────────────────────────────┤
│                   Chunk Range <int64>                   │
├─────────────────────────────────────────────────────────┤
│                  Chunk Exists <uvarint>                 │
│  # 1 if head chunk exists, 0 otherwise to detect a nil  |
|  # chunk. Below fields exists only when it's 1 here.    |
├────────────────────────────┬────────────────────────────┤
│      Chunk Mint <int64>    │     Chunk Maxt <int64>     │
├────────────────────────────┴────────────────────────────┤
│                  Chunk Encoding <byte>                  │
├────────────────────────────┬────────────────────────────┤
│     len(Chunk) <uvarint>   │     Chunk <bytes>          │
├────────────────────────────┼────────────────────────────┤
|   sampleBuf[0].t <int64>   |  sampleBuf[0].v <float64>  | 
├────────────────────────────┼────────────────────────────┤
|   sampleBuf[1].t <int64>   |  sampleBuf[1].v <float64>  | 
├────────────────────────────┼────────────────────────────┤
|   sampleBuf[2].t <int64>   |  sampleBuf[2].v <float64>  | 
├────────────────────────────┼────────────────────────────┤
|   sampleBuf[3].t <int64>   |  sampleBuf[3].v <float64>  | 
└────────────────────────────┴────────────────────────────┘
```

Ends with a single record for tombstones. The encoded tombstones uses the same encoding as tombstone file in blocks.

```
┌─────────────────────────────────────────────────────────────────┐
│                        Record Type <byte>                       │
├───────────────────────────────────┬─────────────────────────────┤
│ len(Encoded Tombstones) <uvarint> │ Encoded Tombstones <bytes>  │
└───────────────────────────────────┴─────────────────────────────┘
```

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -151,6 +151,9 @@ type Options struct {
 	// Enables the in memory exemplar storage,.
 	EnableExemplarStorage bool
 
+	// Enables the snapshot of in-memory chunks on shutdown. This makes restarts faster.
+	EnableMemorySnapshotOnShutdown bool
+
 	// MaxExemplars sets the size, in # of exemplars stored, of the single circular buffer used to store exemplars in memory.
 	// See tsdb/exemplar.go, specifically the CircularExemplarStorage struct and it's constructor NewCircularExemplarStorage.
 	MaxExemplars int64
@@ -722,6 +725,7 @@ func open(dir string, l log.Logger, r prometheus.Registerer, opts *Options, rngs
 	headOpts.SeriesCallback = opts.SeriesLifecycleCallback
 	headOpts.EnableExemplarStorage = opts.EnableExemplarStorage
 	headOpts.MaxExemplars.Store(opts.MaxExemplars)
+	headOpts.EnableMemorySnapshotOnShutdown = opts.EnableMemorySnapshotOnShutdown
 	db.head, err = NewHead(r, l, wlog, headOpts, stats.Head)
 	if err != nil {
 		return nil, err
--- a/tsdb/encoding/encoding.go
+++ b/tsdb/encoding/encoding.go
@@ -17,6 +17,7 @@ import (
 	"encoding/binary"
 	"hash"
 	"hash/crc32"
+	"math"
 	"unsafe"
 
 	"github.com/dennwc/varint"
@@ -40,6 +41,7 @@ func (e *Encbuf) Len() int    { return len(e.B) }
 
 func (e *Encbuf) PutString(s string) { e.B = append(e.B, s...) }
 func (e *Encbuf) PutByte(c byte)     { e.B = append(e.B, c) }
+func (e *Encbuf) PutBytes(b []byte)  { e.B = append(e.B, b...) }
 
 func (e *Encbuf) PutBE32int(x int)      { e.PutBE32(uint32(x)) }
 func (e *Encbuf) PutUvarint32(x uint32) { e.PutUvarint64(uint64(x)) }
@@ -54,6 +56,10 @@ func (e *Encbuf) PutBE32(x uint32) {
 func (e *Encbuf) PutBE64(x uint64) {
 	binary.BigEndian.PutUint64(e.C[:], x)
 	e.B = append(e.B, e.C[:8]...)
+}
+
+func (e *Encbuf) PutBEFloat64(x float64) {
+	e.PutBE64(math.Float64bits(x))
 }
 
 func (e *Encbuf) PutUvarint64(x uint64) {
@@ -71,6 +77,12 @@ func (e *Encbuf) PutUvarintStr(s string) {
 	b := *(*[]byte)(unsafe.Pointer(&s))
 	e.PutUvarint(len(b))
 	e.PutString(s)
+}
+
+// PutUvarintBytes writes a a variable length byte buffer.
+func (e *Encbuf) PutUvarintBytes(b []byte) {
+	e.PutUvarint(len(b))
+	e.PutBytes(b)
 }
 
 // PutHash appends a hash over the buffers current contents to the buffer.
@@ -247,6 +259,10 @@ func (d *Decbuf) Be64() uint64 {
 	x := binary.BigEndian.Uint64(d.B)
 	d.B = d.B[8:]
 	return x
+}
+
+func (d *Decbuf) Be64Float64() float64 {
+	return math.Float64frombits(d.Be64())
 }
 
 func (d *Decbuf) Be32() uint32 {
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -97,6 +97,8 @@ type Head struct {
 	// chunkDiskMapper is used to write and read Head chunks to/from disk.
 	chunkDiskMapper *chunks.ChunkDiskMapper
 
+	chunkSnapshotMtx sync.Mutex
+
 	closedMtx sync.Mutex
 	closed    bool
 
@@ -122,9 +124,10 @@ type HeadOptions struct {
 	// StripeSize sets the number of entries in the hash map, it must be a power of 2.
 	// A larger StripeSize will allocate more memory up-front, but will increase performance when handling a large number of series.
 	// A smaller StripeSize reduces the memory allocated, but can decrease performance with large number of series.
-	StripeSize            int
-	SeriesCallback        SeriesLifecycleCallback
-	EnableExemplarStorage bool
+	StripeSize                     int
+	SeriesCallback                 SeriesLifecycleCallback
+	EnableExemplarStorage          bool
+	EnableMemorySnapshotOnShutdown bool
 
 	// Runtime reloadable options.
 	MaxExemplars atomic.Int64
@@ -439,11 +442,25 @@ func (h *Head) Init(minValidTime int64) error {
 	h.minValidTime.Store(minValidTime)
 	defer h.postings.EnsureOrder()
 	defer h.gc() // After loading the wal remove the obsolete data from the head.
+	defer func() {
+		// Loading of m-mapped chunks and snapshot can make the mint of the Head
+		// to go below minValidTime.
+		if h.MinTime() < h.minValidTime.Load() {
+			h.minTime.Store(h.minValidTime.Load())
+		}
+	}()
 
 	level.Info(h.logger).Log("msg", "Replaying on-disk memory mappable chunks if any")
 	start := time.Now()
 
-	mmappedChunks, err := h.loadMmappedChunks()
+	snapIdx, snapOffset, refSeries, err := h.loadChunkSnapshot()
+	if err != nil {
+		return err
+	}
+	level.Info(h.logger).Log("msg", "Chunk snapshot loading time", "duration", time.Since(start).String())
+
+	mmapChunkReplayStart := time.Now()
+	mmappedChunks, err := h.loadMmappedChunks(refSeries)
 	if err != nil {
 		level.Error(h.logger).Log("msg", "Loading on-disk chunks failed", "err", err)
 		if _, ok := errors.Cause(err).(*chunks.CorruptionErr); ok {
@@ -451,10 +468,10 @@ func (h *Head) Init(minValidTime int64) error {
 		}
 		// If this fails, data will be recovered from WAL.
 		// Hence we wont lose any data (given WAL is not corrupt).
-		mmappedChunks = h.removeCorruptedMmappedChunks(err)
+		mmappedChunks = h.removeCorruptedMmappedChunks(err, refSeries)
 	}
 
-	level.Info(h.logger).Log("msg", "On-disk memory mappable chunks replay completed", "duration", time.Since(start).String())
+	level.Info(h.logger).Log("msg", "On-disk memory mappable chunks replay completed", "duration", time.Since(mmapChunkReplayStart).String())
 	if h.wal == nil {
 		level.Info(h.logger).Log("msg", "WAL not found")
 		return nil
@@ -502,6 +519,9 @@ func (h *Head) Init(minValidTime int64) error {
 
 	walReplayStart := time.Now()
 
+	if snapIdx > startFrom {
+		startFrom = snapIdx
+	}
 	// Backfill segments from the most recent checkpoint onwards.
 	for i := startFrom; i <= endAt; i++ {
 		s, err := wal.OpenReadSegment(wal.SegmentName(h.wal.Dir(), i))
@@ -509,7 +529,14 @@ func (h *Head) Init(minValidTime int64) error {
 			return errors.Wrap(err, fmt.Sprintf("open WAL segment: %d", i))
 		}
 
-		sr := wal.NewSegmentBufReader(s)
+		offset := 0
+		if i == snapIdx {
+			offset = snapOffset
+		}
+		sr, err := wal.NewSegmentBufReaderWithOffset(offset, s)
+		if err != nil {
+			return errors.Wrapf(err, "segment reader (offset=%d)", offset)
+		}
 		err = h.loadWAL(wal.NewReader(sr), multiRef, mmappedChunks)
 		if err := sr.Close(); err != nil {
 			level.Warn(h.logger).Log("msg", "Error while closing the wal segments reader", "err", err)
@@ -533,29 +560,49 @@ func (h *Head) Init(minValidTime int64) error {
 	return nil
 }
 
-func (h *Head) loadMmappedChunks() (map[uint64][]*mmappedChunk, error) {
+func (h *Head) loadMmappedChunks(refSeries map[uint64]*memSeries) (map[uint64][]*mmappedChunk, error) {
 	mmappedChunks := map[uint64][]*mmappedChunk{}
 	if err := h.chunkDiskMapper.IterateAllChunks(func(seriesRef, chunkRef uint64, mint, maxt int64, numSamples uint16) error {
 		if maxt < h.minValidTime.Load() {
 			return nil
 		}
-
-		slice := mmappedChunks[seriesRef]
-		if len(slice) > 0 {
-			if slice[len(slice)-1].maxTime >= mint {
-				return &chunks.CorruptionErr{
-					Err: errors.Errorf("out of sequence m-mapped chunk for series ref %d", seriesRef),
-				}
+		ms, ok := refSeries[seriesRef]
+		if !ok {
+			slice := mmappedChunks[seriesRef]
+			if len(slice) > 0 && slice[len(slice)-1].maxTime >= mint {
+				return errors.Errorf("out of sequence m-mapped chunk for series ref %d", seriesRef)
 			}
+
+			slice = append(slice, &mmappedChunk{
+				ref:        chunkRef,
+				minTime:    mint,
+				maxTime:    maxt,
+				numSamples: numSamples,
+			})
+			mmappedChunks[seriesRef] = slice
+			return nil
 		}
 
-		slice = append(slice, &mmappedChunk{
+		if len(ms.mmappedChunks) > 0 && ms.mmappedChunks[len(ms.mmappedChunks)-1].maxTime >= mint {
+			return errors.Errorf("out of sequence m-mapped chunk for series ref %d", seriesRef)
+		}
+
+		h.metrics.chunks.Inc()
+		h.metrics.chunksCreated.Inc()
+		ms.mmappedChunks = append(ms.mmappedChunks, &mmappedChunk{
 			ref:        chunkRef,
 			minTime:    mint,
 			maxTime:    maxt,
 			numSamples: numSamples,
 		})
-		mmappedChunks[seriesRef] = slice
+		h.updateMinMaxTime(mint, maxt)
+		if ms.headChunk != nil && maxt >= ms.headChunk.minTime {
+			// The head chunk was completed and was m-mapped after taking the snapshot.
+			// Hence remove this chunk.
+			ms.nextAt = 0
+			ms.headChunk = nil
+			ms.app = nil
+		}
 		return nil
 	}); err != nil {
 		return nil, errors.Wrap(err, "iterate on on-disk chunks")
@@ -565,7 +612,7 @@ func (h *Head) loadMmappedChunks() (map[uint64][]*mmappedChunk, error) {
 
 // removeCorruptedMmappedChunks attempts to delete the corrupted mmapped chunks and if it fails, it clears all the previously
 // loaded mmapped chunks.
-func (h *Head) removeCorruptedMmappedChunks(err error) map[uint64][]*mmappedChunk {
+func (h *Head) removeCorruptedMmappedChunks(err error, refSeries map[uint64]*memSeries) map[uint64][]*mmappedChunk {
 	level.Info(h.logger).Log("msg", "Deleting mmapped chunk files")
 
 	if err := h.chunkDiskMapper.DeleteCorrupted(err); err != nil {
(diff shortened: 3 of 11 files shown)


codesome:
There were some bugs in the replay part in terms of handing the file, will benchmark again after fixing them. But from what I could gather with this:

Snapshotting chunks ~1.5M series takes ~3.6s, lets take 4s for any anomalies. So that is like ~30s for 10M series.

After running for 2hrs, replay of master took 2m34s. Excluding the mmap replay of 14s, that naively translates to ~11.6s for 10m worth of WAL data.

Replaying the chunk snapshot took around 11s with snapshot interval being 10m. As there was some bug in code, taking 10m of WAL replay from above, replaying entire data with snapshots would be 11+14+11.6=36.6s roughly in the worst case. That is 4-5x faster replay, and expect it to be >5x in general case.

I forgot to check the size of snapshots.

Before I dive into fixing the issues, how do these numbers look for a trade-off? @brian-brazil

codesome:
Additionally, replay time won't increase linearly (or linear with very small factor) with partial chunks like WAL replay would. This was tested only for 2h of WAL data, but in extended use, I guess it goes upto 3h or more. So in that case, the difference would be >5x.

codesome:
I compared only the replay times here whereas the turnaround time is the one that matters. With <4s for taking a snapshot (with most hovering around 3s), I see not much change with the speedup even for turnaround time :)

brian-brazil:
To summaries:
90s for 30M series, or an additional ~15% duty cycle on I/O for 10m snapshots. That's a fair bit.

The benchmark was for ~250k/s. For 1M/s, we'd be looking at 2h of WAL replay taking ~616s with the current mmap code.

Replaying just the chunk snapshot for 30M series would be ~220s plus ~46s for the WAL replay at 1M/s so ~266s in all. So the saving here is only about 2x for very large Prometheus servers, and we're still over a minute.

I'm not sure this complexity is worth it given all the additional disk writes it'd cause. I'd prefer to keep things simple to allow for some future solution that'd get us under a minute for large servers.

codesome:
> get us under a minute for large servers.

Might be possible for only replays, but for turn around time that sounds very ambitious unless we can write out the data to disk faster or m-map partial chunks (unlikely) :) I will think of more ways

codesome:
From the way I see it, for the sake of some arbitrary target we should not block some optimization which is possible right now. The tradeoff seems to be good.

Additionally, the code and the logic are well separated from the rest of the Head block logic and I don't see this being a complex addition. It is also safe to downgrade without any issues as it will fallback to the old replay which makes it easy to upgrade to any new techniques which hits the target in the future. This could also live behind a flag for now with disabled by default.

brancz:
I tend to agree with @codesome

brian-brazil:
I can see that argument, but the duty cycle and additional disk writes still concern me. Do you have a rough idea of what the overall impact of the writes is on top of the WAL/mmap/compactions we already do? Write amplification is something we should keep an eye on, and I think this could double it.

brancz:
I agree we need to measure and decide based on what's measured, but I don't think anyone suggested anything contrary to that.

codesome:
Here are some calculations (for a single series)

Considering a time window of 2h, and 15s scrape interval, the calculations are based on how many chunks will be written to disk for the chunks created in these 2h in it's entire lifecycle.

**Current writes**

Chunks written in 2h to disk = 4 (from mmapping) + 4 (to write to blocks) = 8

I am not totally sure about the compression ratio between WAL and XORChunk, but considering that a chunk in WAL would be twice the size than XORChunk. So here it is 8 + 8 = 16. I will add +2 for the checkpoints (or should I not?). So now it's 12+2 = 14.

If we consider that a block goes through at least 2 compactions in it's lifecycle, then writing those 4 chunks twice = 16 + (4+4) = 24

**Writes from partial chunks**

Considering 15m snapshot interval, we can say it writes 0.5 chunks at 15m then 1 chunks at 30m. As the cycle continues, it would be 4*(0.5+1) = 6 chunks for 2 hrs.

**Additional writes**

6 additional chunks compared to 24 chunks currently, that is 25% increase in writes.

I have not considered block index writes and >2 compactions which will happen, so increase in writes would be less than <20% I believe.

Did I miss something?

brian-brazil:
> considering that a chunk in WAL would be twice the size than XORChunk

It's going to be about 3x with compression, presuming a conservative 2 bytes/samples in a chunk.

> Considering 15m snapshot interval

Why the switch from the 10m used previously?

My math is in terms of a single sample which is conservatively ~2 bytes. In terms of full blocks we'll currently write it out 9 times through compaction, plus another 6 bytes for a compressed WAL which is the same as 3 writes. So our current write amplification is around 12.

Mmaping I believe works out as no more than one additional write so that'd bring it to 13.

Snapshots of incomplete chunks every 10m could write a sample out 12 times if the chunk is never completed, which is plausible with a 60s interval. Halve that for the average case as later samples in the 2h get written fewer times, so 6. Overall impact is 13->19 so roughly adding ~46% more write amplification.

Is shaving off half of restart time worth reducing SSD lifetime by a third?


Doing the math for a Samsung 860 EVO 500GB with 200k/s (i.e. 15d retention just fits) which quotes 300TB of writes that takes the official supported lifetime from ~22 months to ~15 months. Now I'd presume it'll last a bit longer than that, but it's not as if it's 22 years versus 15 years where it'd be clear that it wouldn't be a problem.

codesome:
Just a thought: given the write amplification with regular snapshotting, one of the ways to use this snapshotting can be to snapshot only when Prometheus is exiting. This way, a graceful restart will be fast, but not crashes. 

With graceful restarts being very few in number, I expect not much impact on write amplification.

brian-brazil:
That would reduce the disk writes, but then we'd be slowing down shutdown and the code would be exercised less so more likely a bug will slip in. I'm a bit wary of that, and keep in mind that the shutdown time would now needed to be added in to the restart time.

codesome:
> but then we'd be slowing down shutdown

> and keep in mind that the shutdown time would now needed to be added in to the restart time.

But the overall turnaround time is still faster. Maybe even faster than previous calculation because with snapshotting at the end, the WAL to be replayed is 0, and snapshotting takes less time than replaying that extra WAL in the old calculation.

Some recalculation with snapshot at the end with the data gathered in this comment https://github.com/prometheus/prometheus/pull/7229#issuecomment-631465164

For the turn around time we only have writing the snapshot + the replay of mmap and snapshot (no WAL)

For 1.5M series for 2h of data

* Master = 2m34s for replay (hence turnaround time)
* With Snapshotting = 4s (writing snapshot) + 14s (mmap replay) + 11s (snapshot replay) + 0s (WAL replay) = 29s

So that is ~80% reduction; better than before because the earlier calculation was based on the worst case of having to replay some WAL.

As we will have more than 2h of WAL, the reduction is going to be more.

> the code would be exercised less so more likely a bug will slip in

The bug can be fixed :) It is similar to WAL replay - it is also exercised only when Prometheus starts so more likely a bug will slip, but we still have it.

roidelapluie:
> > the code would be exercised less so more likely a bug will slip in
> 
> The bug can be fixed :) It is similar to WAL replay - it is also exercised only when Prometheus starts so more likely a bug will slip, but we still have it.

Prometheus does not always shut down nicely and does not always shutdown attended. So I am on brian's side that there is an important risk here.

codesome:
If the snapshot is corrupt, it can be skipped and fallen back to using the WAL.

brian-brazil:
I didn't say corruption, I said bug. A bug could happen at any stage of the reading or writing of the snapshot, and code that is run less often is code more likely to have undiscovered bugs.

codesome:
Right. But we could also have extensive tests around it. I am on the side of having this feature in, maybe behind a flag, and disabled by default till we are confident about it instead of blocking it altogether.

brian-brazil:
Hiding it behind a flag is worse in my mind. We get all the complexity costs, and the code is exercised even less often.

codesome:
I would not worry about complexity as the code for this is well isolated and it is easy to turn it on or off. But about the code being used, right, it would be used less, but I expect Cortex/Thanos which import TSDB to enable it and it getting tested extensively.

Do you have any other suggestions if we were to not having it behind the flag? Is the code being executed less the only concern (which would make subtle bug invisible)?

brian-brazil:
> I would not worry about complexity as the code for this is well isolated 

From a look at the PR this isn't the case, it's intertwined with the WAL reading code.

> Do you have any other suggestions if we were to not having it behind the flag? Is the code being executed less the only concern (which would make subtle bug invisible)?

My point is more that putting it behind a flag makes it worse from my standpoint. If the main answer to the disadvantages of feature is that it's okay because it's behind a rarely used flag, it'd be better to not have the feature at all.

A feature like this should be always enabled, and justified on that basis.

codesome:
After some offline discussions, here are some points on some aspects

**Benefits**

I ran a quick test of this PR as a part of Cortex to again verify the numbers. I was able to conclude a 50-80+% reduction in turnover time. Ideally, there would be no WAL to replay as snapshot happens at the end, but due to Cortex usage of TSDB or bug in this PoC, The 50% number comes up with some WAL to replay. Else it would be 75%+ reduction in most cases as calculated before.

**Complexity**

The majority of the code is well isolated. The writing of snapshot and the loading of snapshot is an isolated code in its own method. And writing/loading snapshot is as simple as calling those methods are not.

The main complexity comes from the special handling of m-mapped chunks during the creation of a series from the WAL. When you create a series from WAL (after having replayed the snapshot and m-mapped chunks already), you have to check if the m-mapped chunks for the new series overlaps with the chunk from snapshot and hence discard (in case its a duplicate series record). So this is the only additional complexity.

**Risks**

* We track the exact byte in the WAL file from where to start replaying. So, any races that might exist between writing a record and tracking that byte should be eliminated.
* With the view of eliminating races, it is best if it can be assured that no samples are appended once we call the Close() method on Head.

codesome:
**Risks**

* Should properly stitch back the chunk from snapshot with other parameters inside the code properly set. Else it can cause problems in future ingestion of samples.

bwplotka:
Hey, small update from our side: Current version of TSDB vendored by Thanos  (or Thanos mods on top of it or mix) is struggling a little bit for Receiver + millions of series. Struggling version: https://github.com/prometheus/prometheus/commit/3268eac2ddda4300f962213feaf6b763a0ecd295

We are experimenting, bisecting etc, but I might be slower in reviews until we get back on track with TSDB, will keep you posted.

krasi-georgiev:
discussed the design with Ganesh on slack and can't think of any blockers. 

I agree that the code should be as isolated as possible and should have a test to ensure failover to WAL on startup if the snapshot is corrupted or unrecognized format.

tsdb/chunkenc/xor.go
```diff
@@ -64,6 +64,11 @@ func NewXORChunk() *XORChunk {
 	return &XORChunk{b: bstream{stream: b, count: 0}}
 }
 
+// NewXORChunk returns a new chunk with XOR encoding of the given size.
+func NewXORChunkFromBytes(b []byte) *XORChunk {
```
bwplotka, line 68: Do we need this? To me it's unnecessary. You have a super quick `chunk.FromData` function. I would drop it and avoid duplicates in API :hugs:

tsdb/encoding/encoding.go
```diff
@@ -184,23 +191,27 @@ func (d *Decbuf) Skip(l int) {
 }
 
 func (d *Decbuf) UvarintStr() string {
-	return string(d.UvarintBytes())
+	return string(d.UvarintBytes(nil))
 }
 
 // The return value becomes invalid if the byte slice goes away.
```
bwplotka, line 197: > // The return value becomes invalid if the byte slice goes away.

I think this is not valid, it's obsolete comment, and it was obsolete even before this change, as we always copy, no? :thinking:

tsdb/head.go
```diff
@@ -524,23 +531,49 @@ Outer:
 					seriesCreationErr = err
```
bwplotka, line 531: Why not `series, created, seriesCreationErr := ` above? :thinking:

tsdb/head.go
```diff
@@ -524,23 +531,49 @@ Outer:
...
+					// Hence replace the old m-mapped chunks with new chunks.
+
+					if overlapsClosedInterval(
+						series.mmappedChunks[0].minTime,
+						series.mmappedChunks[len(series.mmappedChunks)-1].maxTime,
+						mmc[0].minTime,
+						mmc[len(mmc)-1].maxTime,
+					) {
+						// TOOD(codesome) return this instead.
+						panic("overlapping mmapped chunks")
```
bwplotka, line 560: this? 
I guess todo for this PR

tsdb/head.go
```diff
@@ -524,23 +531,49 @@ Outer:
...
 						h.updateMinMaxTime(series.minTime(), series.maxTime())
 					}
-				} else {
-					// TODO(codesome) Discard old samples and mmapped chunks and use mmap chunks for the new series ID.
+				} else if len(mmc) > 0 && len(series.mmappedChunks) > 0 &&
+					series.mmappedChunks[len(series.mmappedChunks)-1].maxTime > mmc[len(mmc)-1].maxTime {
+					// These new m-mapped chunks in the future for the same series.
+					// Hence replace the old m-mapped chunks with new chunks.
+
+					if overlapsClosedInterval(
```
bwplotka, line 553: Just curious, why would such overlap happen? :thinking:

bwplotka: Essentially why would you have multiple `series.mmappedChunks` for the same series.ref?

tsdb/head.go
```diff
@@ -524,23 +531,49 @@ Outer:
...
+					}
+
+					h.metrics.chunksCreated.Add(float64(len(mmc)))
+					h.metrics.chunksRemoved.Add(float64(len(series.mmappedChunks)))
+					h.metrics.chunks.Add(float64(len(mmc) - len(series.mmappedChunks)))
+
+					series.mmappedChunks = mmc
+
+					if series.headChunk != nil && mmc[len(mmc)-1].maxTime >= series.headChunk.minTime {
+						// The head chunk was completed and was m-mapped after taking the snapshot.
```
bwplotka, line 570: again, why this would happen? Why something could be mapped after taking (chunk) snapshot? This can be done only on shutdown

tsdb/head.go
```diff
@@ -645,23 +678,33 @@ func (h *Head) Init(minValidTime int64) error {
 	level.Info(h.logger).Log("msg", "Replaying WAL and on-disk memory mappable chunks if any, this may take a while")
 	start := time.Now()
 
-	mmappedChunks, err := h.loadMmappedChunks()
+	s := time.Now()
+	snapIdx, snapOffset, refSeries, err := h.loadChunkSnapshot()
+	if err != nil {
+		return nil
```
bwplotka, line 684: :scream:

tsdb/head.go
```diff
@@ -645,23 +678,33 @@ func (h *Head) Init(minValidTime int64) error {
...
+	}
+	level.Info(h.logger).Log("msg", "chunk snapshot loading time", "duration", time.Since(s).String())
```
bwplotka, line 686: We do log lines with upper case

codesome:
Regarding the code complexity in the WAL replay, I realised this now: The only thing where it touches the WAL replay in _this PR_ is when we create a new series to handle the m-mapped chunks. _This part is actually independent of chunks snapshot and an enhancement to how we handle m-mapped chunks, and I can open a separate PR for that._

tsdb/head.go
codesome: Would be possible if we have the min valid time grace period :)

tsdb/head.go
codesome: Yes, TODO is for this PR. Still WIP.

tsdb/head.go
codesome: Currently, the code considers snapshot at intervals, hence this was possible. Once we can confirm there is no race during shutdown where it can happen, we can remove this check. (Race of a sample coming in after snapshot which causes the chunk to m-map).

bwplotka: :+1:  Let's commen that

codesome:
I have opened https://github.com/prometheus/prometheus/pull/7438 for the WAL replay change in this PR (after that, the snapshotting and replaying it is well isolated and best effort).

codesome:
So, along with chunks, we also need tombstones to go in this snapshot. There are 2 ways to do it with tradeoffs in both:

1. Store the tombstones with the series itself. Much simpler design as its a single record type. Trade-off is: we will be calling `tombstones.Get(ref)` for all series though we won't have tombstones most of the time. That call has locking and unlocking. But I think locking and unlocking a million times would be very quick.
2. Store all tombstones in a single place either in the beginning or the end of snapshot. This way, its a single locking and only going through actual tombstones if they exist. Much faster to write tombstones. But do we actually need this optimization?

brian-brazil:
> That call has locking and unlocking. But I think locking and unlocking a million times would be very quick.

It's around 50ns for an uncontended mutex, so 50ms. And that's not too bad for larger numbers of series either. If it's more than once per series that could get slow - but on the other hand tombstones shouldn't be common so I'd suggest keeping it simple.

codesome:
The following code took <200ms to run (simulates getting tombstones for 10M series). The first design it is! Storing with the series itself.

```go
package main

import (
	"fmt"
	"sync"
	"time"
)

var someMap = make(map[uint64]interface{})
var someLock sync.RWMutex

func returnNilAndNil(ref uint64) (interface{}, error) {
	someLock.RLock()
	defer someLock.RUnlock()
	return someMap[ref], nil
}

func main() {
	start := time.Now()
	for i := uint64(0); i < 10000000; i++ {
		returnNilAndNil(i)
	}
	elapsed := time.Since(start)
	fmt.Println(elapsed)
}
```

codesome:
Tests are passing now and I have added a unit test for this. Additionally, now we have tombstones in the snapshot and also the sampleBuf to avoid iterating through the chunk on replay to restore the sampleBuf.

The updated format is in the PR description. This PR is now ready for review! :shipit:

tsdb/head.go
```diff
@@ -524,28 +532,59 @@ Outer:
 					seriesCreationErr = err
 					break Outer
 				}
+				if !created {
```
codesome, line 535: For all these changes in the `loadWAL`, please review https://github.com/prometheus/prometheus/pull/7438 (same changes in different PR as it's not totally related to snapshotting)

tsdb/encoding/encoding.go
codesome: Now that we are re-using the slice, it _makes sense_ because if slice does not exist, then there is no guarantee about the data :)

tsdb/head.go
codesome: To be explicit that the scope of `seriesCreationErr` is outside the loop, else its not straightforward to know if we use `:=`

tsdb/head.go
codesome: You are right, I should not be using `series.ref` here to get mmap chunks, rather use `s.Ref` from the record. I will fix that in https://github.com/prometheus/prometheus/pull/7438. (So in that case overlapping could be possible for various edge cases)

codesome: Nice catch about `series.ref`!

tsdb/head.go
codesome: Another reason why this could happen: Prometheus shutdown ungracefully and restart led it to pick up the old snapshot, hence we have a old chunk which needs to be cleaned up here. Will add that in the comments.

codesome: It will go in https://github.com/prometheus/prometheus/pull/7438

tsdb/head.go
codesome: Will fix this in https://github.com/prometheus/prometheus/pull/7438

tsdb/tombstones/tombstones.go
```diff
@@ -98,33 +97,27 @@ func WriteFile(logger log.Logger, dir string, tr Reader) (int64, error) {
...
 		return 0, err
 	}
 	size += n
 
-	mw := io.MultiWriter(f, hash)
-
-	if err := tr.Iter(func(ref uint64, ivs Intervals) error {
-		for _, iv := range ivs {
-			buf.Reset()
+	bytes, err := Encode(tr)
```
codesome, line 106: For these changes in the tomstones, I have another PR for that https://github.com/prometheus/prometheus/pull/7967

tsdb/db.go
```diff
@@ -1301,20 +1320,21 @@ func (db *DB) Close() error {
...
-
 	merr.Add(g.Wait())
-
 	if db.lockf != nil {
 		merr.Add(db.lockf.Release())
 	}
-	merr.Add(db.head.Close())
+	if withHeadSnapshot {
+		merr.Add(db.head.Close())
+	} else {
```
bwplotka, line 1335: never use else! (:

tsdb/head.go
```diff
@@ -2322,7 +2579,333 @@ type mmappedChunk struct {
...
+	chunkBytesCopy := make([]byte, len(chunkBytes))
+	copy(chunkBytesCopy, chunkBytes)
+
+	chk, err := chunkenc.FromData(enc, chunkBytesCopy)
+	if err != nil {
+		return csr, errors.Wrap(err, "chunk from data")
+	}
+	csr.mc.chunk = chk
+
+	for i := range csr.sampleBuf {
```
bwplotka, line 2667: Would be interesting why we need exactly 4 sample buf? (:

bwplotka (commented):
Generally looking good but I need to review properly.

Let's also add some `.md` file with this format to our docs next to index format

Thanks :+1:

roidelapluie:
Should we resume work here? Additionally, since the extra disk space needed might surprise users, it might require a feature flag.

codesome:
_We should resume_, but I don't think I can get to it for many weeks to come. And yes, I was thinking of feature flag for this too :)

bhiravabhatla:
Waiting for this PR to be merged. This is a very interesting/important feature we would like to have. Any planned ETA for this?

codesome:
I have rebased the PR and added a feature flag for this now. It is ready for review :)

bwplotka (approved):
Looks solid. It's a lot of complexity, but given feature flags, we can experiment with this! LGTM, thanks!

codesome:
WOW, I did not expect a direct merge XD Thanks :)
Do you have any possible suggestions for the record formats @bwplotka?

PR #7231, opened 2020-05-09
Correction of misleading help text(#5142)

### LINKED ISSUE #5142: Misleading help text for storage.tsdb.retention.size
I installed 2.7.0 on a test server and ran into one point of confusion:

```
prometheus, version 2.7.0 (branch: HEAD, revision: 410ee9e04acb8f59f400858752ca82b4ef88035e)
  build user:       root@3bc81b516055
  build date:       20190128-10:09:51
  go version:       go1.11.5
```

The help text for `--storage.tsdb.retention.size` says this:

```
[EXPERIMENTAL] Maximum number of bytes that can be stored for blocks. Units supported: KB, MB, GB, TB, PB. This flag is experimental and can be changed in future releases.
```

I interpreted that to mean that it followed the common convention of accepting a number as bytes while allowing units as a shortcut for readability. What it actually means is that you must specify one of the units or get an error like this:

```
Error parsing commandline arguments: units: unknown unit  in 96589578240b
```

It would be nice if it either defaulted to bytes or the wording was something like “Limit the maximum size of block storage. Specify the unit suffix using one of: KB, MB, GB, TB, PB.” to make it obvious that it's not optional.

ArthurSens:
It's my first PR to prometheus, so if I did anything wrong please tell me and I will be glad to fix any problem

Fixes #5142

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -204,7 +204,7 @@ func main() {
 	a.Flag("storage.tsdb.retention.time", "How long to retain samples in storage. When this flag is set it overrides \"storage.tsdb.retention\". If neither this flag nor \"storage.tsdb.retention\" nor \"storage.tsdb.retention.size\" is set, the retention time defaults to "+defaultRetentionString+". Units Supported: y, w, d, h, m, s, ms.").
 		SetValue(&newFlagRetentionDuration)
 
-	a.Flag("storage.tsdb.retention.size", "[EXPERIMENTAL] Maximum number of bytes that can be stored for blocks. Be aware that a unit is required, supported units: B, KB, MB, GB, TB, PB. Ex: \"512MB\". This flag is experimental and can be changed in future releases.").
+	a.Flag("storage.tsdb.retention.size", "[EXPERIMENTAL] Maximum number of bytes that can be stored for blocks. A unit is required, supported units: B, KB, MB, GB, TB, PB, EB. Ex: \"512MB\". This flag is experimental and can be changed in future releases.").
 		BytesVar(&cfg.tsdb.MaxBytes)
 
 	a.Flag("storage.tsdb.no-lockfile", "Do not create lockfile in data directory.").
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -204,7 +204,7 @@ func main() {
 	a.Flag("storage.tsdb.retention.time", "How long to retain samples in storage. When this flag is set it overrides \"storage.tsdb.retention\". If neither this flag nor \"storage.tsdb.retention\" nor \"storage.tsdb.retention.size\" is set, the retention time defaults to "+defaultRetentionString+". Units Supported: y, w, d, h, m, s, ms.").
 		SetValue(&newFlagRetentionDuration)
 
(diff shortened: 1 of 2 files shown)


cmd/prometheus/main.go
```diff
@@ -204,7 +204,7 @@ func main() {
 	a.Flag("storage.tsdb.retention.time", "How long to retain samples in storage. When this flag is set it overrides \"storage.tsdb.retention\". If neither this flag nor \"storage.tsdb.retention\" nor \"storage.tsdb.retention.size\" is set, the retention time defaults to "+defaultRetentionString+". Units Supported: y, w, d, h, m, s, ms.").
 		SetValue(&newFlagRetentionDuration)
 
-	a.Flag("storage.tsdb.retention.size", "[EXPERIMENTAL] Maximum number of bytes that can be stored for blocks. Units supported: KB, MB, GB, TB, PB. This flag is experimental and can be changed in future releases.").
+	a.Flag("storage.tsdb.retention.size", "[EXPERIMENTAL] Maximum number of bytes that can be stored for blocks. Make sure to specify the unit suffix. Ex: \"512MB\". Units supported: KB, MB, GB, TB, PB. This flag is experimental and can be changed in future releases.").
```
brian-brazil, line 207: B is also supported.

The new text doesn't feel right. Maybe something like A unit is required, supported units:

ArthurSens: Be aware that a unit is required, supported units: B, KB, MB, GB, TB, PB. Ex: "512MB"

Is that ok?

cmd/prometheus/main.go
```diff
@@ -204,7 +204,7 @@ func main() {
 	a.Flag("storage.tsdb.retention.time", "How long to retain samples in storage. When this flag is set it overrides \"storage.tsdb.retention\". If neither this flag nor \"storage.tsdb.retention\" nor \"storage.tsdb.retention.size\" is set, the retention time defaults to "+defaultRetentionString+". Units Supported: y, w, d, h, m, s, ms.").
 		SetValue(&newFlagRetentionDuration)
 
-	a.Flag("storage.tsdb.retention.size", "[EXPERIMENTAL] Maximum number of bytes that can be stored for blocks. Make sure to specify the unit suffix. Ex: \"512MB\". Units supported: KB, MB, GB, TB, PB. This flag is experimental and can be changed in future releases.").
+	a.Flag("storage.tsdb.retention.size", "[EXPERIMENTAL] Maximum number of bytes that can be stored for blocks. Be aware that a unit is required, supported units: B, KB, MB, GB, TB, PB. Ex: \"512MB\". This flag is experimental and can be changed in future releases.").
```
brian-brazil, line 207: No need for "Be aware that", it doesn't add anything.

roidelapluie:
Should we add EB too?

brian-brazil:
If it's supported but missing, we should add it.

ArthurSens:
And is EB supported? I am reading the code but can't seem to find anything on that

If someone could show me where this flag is used on tsdb just so I can better understand the code... I would really appreciate it :sweat_smile:

brian-brazil:
This handling all comes from upstream, which does have it: https://github.com/alecthomas/units/blob/master/bytes.go

brian-brazil:
Thanks!

ArthurSens:
thank you for your help in this pr

Issue #7241, opened 2020-05-11
react UI: clicking the same tab button you are on resets most recently queried data


GiedriusS:
**What did you do?**

Execute a simple query such as `process_virtual_memory_max_bytes` via the new react UI. Click on the same tab button that you are on at the moment.

**What did you expect to see?**

I expected the UI to remain the same and for the results to still be there.

**What did you see instead? Under which circumstances?**

I saw that the data had been wiped and the UI shows: "No data queried yet". Not reproducible with the old UI.

**Environment**

* System information:

	Linux 5.4.0-29-lowlatency x86_64

* Prometheus version:

	prometheus, version 2.18.1 (branch: master, revision: 5fa12ae0ef0ae23e146f02b9cbca209ab3af6188)
	build user:       gstatkevicius@gstatkevicius-desktop
	build date:       20200511-15:53:46
	go version:       go1.14.2

* Alertmanager version:

N/A

* Prometheus configuration file:

```
./documentation/examples/prometheus.yml
```

* Alertmanager configuration file:

N/A

* Logs:

N/A

juliusv:
Thanks for reporting this, I'm taking a look!

juliusv:
Fix in https://github.com/prometheus/prometheus/pull/7243

PR #7243, opened 2020-05-11
React UI: Don't null out data when clicking on current tab

juliusv:
Fixes https://github.com/prometheus/prometheus/issues/7241

### DIFF
--- a/web/ui/react-app/src/pages/graph/Panel.test.tsx
+++ b/web/ui/react-app/src/pages/graph/Panel.test.tsx
@@ -65,9 +65,13 @@ describe('Panel', () => {
       const className = tc.active ? 'active' : '';
       expect(link.prop('className')).toEqual(className);
       link.simulate('click');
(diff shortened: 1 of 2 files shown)


juliusv:
@boyskila :)

boyskila:
This looks like a proper fix, so 👍
However, If we have time we can consider refactor this component with hooks, where ```handleChangeType``` ```func``` can be wrapped in ```useCallback``` hook. Then we can specify both types in the dependency array which will guarantee that function will be executed only when the type got changed. WDYT?

juliusv:
I started with it, but it would at least also require untangling the `options` interface and the handling of it in the parent, so that calling `onUpdateOptions({...options, ...})` wouldn't make the fetch hook depend on the entire `options`, and while that's not hard, there's a good chance of introducing new bugs, so I'd prefer to just do the bug fix separately first.

PR #7244, opened 2020-05-11
Update docker example in migration guide

### LINKED ISSUE #4054: [docs] Docker run example with privileged port
**What did you do?**

I read the migration guide :)

Hints about the migration from Prometheus 1.x to 2.x.

Not _instead_, but I stumbled about the notes about non-root users in combination with privileged ports at the end of the migration guide: https://github.com/prometheus/prometheus/blob/master/docs/migration.md#prometheus-non-root-user

The guide gives an example how to use Docker to make Prometheus available on port 80:

    docker run -u root -p 80:80 prom/prometheus:v2.0.0-rc.2  --web.listen-address :80

An example like this works, too:

    docker run -p 80:9090 -v `pwd`/prometheus.yml:/prometheus.yml prom/prometheus:v2.2.1 --web.external-url=http://prometheus:80 --config.file=/prometheus.yml

Please note that I also added the necessary config file via volume bind.

The example wouldn't drive users to let Prometheus run as root and it also leverages a container's port mapping possibilities - which makes the outer and inner perspective of a container more explicit.

My question is: are there any reasons for the way the current example has been documented? If not, I can submit a PR to improve the Docker example.

ArthurSens:
Fixes #4054

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -189,7 +189,7 @@ for more details.
 If you're using Docker, then the following snippet would be used:
 
 ```
-docker run -u root -p 80:80 prom/prometheus:v2.0.0-rc.2  --web.listen-address :80
+docker run -p 9090:9090 prom/prometheus:latest
 ```
 
 ### Prometheus lifecycle


ArthurSens:
I didn't know if the desired example was to run on port 80 or 9090...

Also, I've changed the prometheus version there because it was an old release... but I can change it back if requested

roidelapluie:
LGTM, any comment @simonpasquier @SuperQ

PR #7250, opened 2020-05-12
Add discovery support for triton global zones

### LINKED ISSUE #3871: Can't add GZ zone discovery end points
Hello @richardkiene 

Since the endpoint is consist of Endpoint, Port, Version.

Could you please add a Zone as a config as well?

`var endpoint = fmt.Sprintf("https://%s:%d/v%d/discover", d.sdConfig.Endpoint, d.sdConfig.Port, d.sdConfig.Version)`

becomes

`var endpoint = fmt.Sprintf("https://%s:%d/v%d/%s/discover", d.sdConfig.Endpoint, d.sdConfig.Port, d.sdConfig.Version,d.sdConfig.Zone)`

where d.sdConfig.Zone can be 'gz'
eg:
"https://cmon.cns.x.com:9163/v1/gz/discover"

Related code:
https://github.com/prometheus/prometheus/blob/675ce533c97bf3b0d952ac346339b4cad17d5588/discovery/triton/triton.go#L193

jzinkweg:
Added optional configuration item server_type, defaults to 'container'.
Setting server-type to 'gz' will discover global zones instead.

Human-friendly hostname discovery depends on cmon 1.7.0:
https://github.com/joyent/triton-cmon/commit/c1a2aeca36cc3c4e8b081a35a01a0b770d30e4ef

Adjust testcases to use discovery config per case as two different types are now supported.

Fixes #3871.

### DIFF
--- a/discovery/triton/triton.go
+++ b/discovery/triton/triton.go
@@ -46,6 +46,7 @@ const (
 
 // DefaultSDConfig is the default Triton SD configuration.
 var DefaultSDConfig = SDConfig{
+	Role:            "container",
 	Port:            9163,
 	RefreshInterval: model.Duration(60 * time.Second),
 	Version:         1,
@@ -54,6 +55,7 @@ var DefaultSDConfig = SDConfig{
 // SDConfig is the configuration for Triton based service discovery.
 type SDConfig struct {
 	Account         string                `yaml:"account"`
+	Role            string                `yaml:"role"`
 	DNSSuffix       string                `yaml:"dns_suffix"`
 	Endpoint        string                `yaml:"endpoint"`
 	Groups          []string              `yaml:"groups,omitempty"`
@@ -71,6 +73,9 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err != nil {
 		return err
 	}
+	if c.Role != "container" && c.Role != "cn" {
+		return errors.New("triton SD configuration requires role to be 'container' or 'cn'")
+	}
 	if c.Account == "" {
 		return errors.New("triton SD configuration requires an account")
 	}
@@ -87,7 +92,7 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 }
 
 // DiscoveryResponse models a JSON response from the Triton discovery.
-type discoveryResponse struct {
+type DiscoveryResponse struct {
 	Containers []struct {
 		Groups      []string `json:"groups"`
 		ServerUUID  string   `json:"server_uuid"`
@@ -96,6 +101,14 @@ type discoveryResponse struct {
 		VMImageUUID string   `json:"vm_image_uuid"`
 		VMUUID      string   `json:"vm_uuid"`
 	} `json:"containers"`
+}
+
+// ComputeNodeDiscoveryResponse models a JSON response from the Triton discovery /gz/ endpoint.
+type ComputeNodeDiscoveryResponse struct {
+	ComputeNodes []struct {
+		ServerUUID     string `json:"server_uuid"`
+		ServerHostname string `json:"server_hostname"`
+	} `json:"cns"`
 }
 
 // Discovery periodically performs Triton-SD requests. It implements
@@ -137,15 +150,32 @@ func New(logger log.Logger, conf *SDConfig) (*Discovery, error) {
 	return d, nil
 }
 
+// triton-cmon has two discovery endpoints:
+// https://github.com/joyent/triton-cmon/blob/master/lib/endpoints/discover.js
+//
+// The default endpoint exposes "containers", otherwise called "virtual machines" in triton,
+// which are (branded) zones running on the triton platform.
+//
+// The /gz/ endpoint exposes "compute nodes", also known as "servers" or "global zones",
+// on which the "containers" are running.
+//
+// As triton is not internally consistent in using these names,
+// the terms as used in triton-cmon are used here.
+
 func (d *Discovery) refresh(ctx context.Context) ([]*targetgroup.Group, error) {
-	var endpoint = fmt.Sprintf("https://%s:%d/v%d/discover", d.sdConfig.Endpoint, d.sdConfig.Port, d.sdConfig.Version)
+	var endpointFormat string
+	switch d.sdConfig.Role {
+	case "container":
+		endpointFormat = "https://%s:%d/v%d/discover"
+	case "cn":
+		endpointFormat = "https://%s:%d/v%d/gz/discover"
+	default:
+		return nil, errors.New(fmt.Sprintf("unknown role '%s' in configuration", d.sdConfig.Role))
+	}
+	var endpoint = fmt.Sprintf(endpointFormat, d.sdConfig.Endpoint, d.sdConfig.Port, d.sdConfig.Version)
 	if len(d.sdConfig.Groups) > 0 {
 		groups := url.QueryEscape(strings.Join(d.sdConfig.Groups, ","))
 		endpoint = fmt.Sprintf("%s?groups=%s", endpoint, groups)
-	}
-
-	tg := &targetgroup.Group{
-		Source: endpoint,
 	}
 
 	req, err := http.NewRequest("GET", endpoint, nil)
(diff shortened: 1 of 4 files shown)


brian-brazil:
This seems to have a very different return type, is this really the same service discovery?

Docs will need updating.

jzinkweg:
For some reason Joyent decided to return a different JSON format for global zones (i.e.: head/compute node). There is less metadata available for these nodes than the VM's running on them, but it's still an odd implementation choice.
The discovery is and actual metric scraping is done via the same triton-cmon infrastructure, so to me it still makes sense to have this as part of the same discovery solution.

I'll update the docs in a moment.

(we're running triton on-prem, no affiliation with Joyent)

jzinkweg:
When updating the documentation (pushed now) I realized using 'container' instead of 'vm' makes more sense given the existing names in use, so I changed the configuration parameter to match the existing convention.

brian-brazil:
Containers and VMs are quite different things, what does Triton call them?

jzinkweg:
Triton refers to them as virtual machines. They can be smartos (lx) zones, docker containers or actual VM's (kvm or bhyve).
I tried to keep the change as small as possible but as this is all in docs or new config, there is no impact from changing "containers" to "virtual machines" instead.

brian-brazil:
Let's keep with upstream terminology.

jzinkweg:
Reverted back to 'vm' setting and updated documentation to use virtual machine instead of container.

discovery/triton/triton.go
```diff
@@ -71,6 +73,9 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err != nil {
 		return err
 	}
+	if c.ServerType != "vm" && c.ServerType != "gz" {
+		return errors.New("triton SD configuration requires server_type to be 'container' or 'gz'")
```
brian-brazil, line 77: container?

docs/configuration/configuration.md
```diff
@@ -1023,34 +1023,42 @@ discovery endpoints.
 The following meta labels are available on targets during relabeling:
 
 * `__meta_triton_groups`: the list of groups belonging to the target joined by a comma separator
```
brian-brazil, line 1025: We generally split into two lists when there's a parameter like this

docs/configuration/configuration.md
```diff
@@ -1023,34 +1023,42 @@ discovery endpoints.
...
+# A list of groups for which targets are retrieved, only supported for virtual machines.
+# If omitted all vm's available to the requesting account are scraped.
 groups:
   [ - <string> ... ]
 
 # The port to use for discovery and metric scraping.
 [ port: <int> | default = 9163 ]
 
-# The interval which should be used for refreshing target containers.
+# The interval which should be used for refreshing target virtual machines / global zones.
```
brian-brazil, line 1061: refreshing targets.  is enough

docs/configuration/configuration.md
```diff
@@ -1023,34 +1023,42 @@ discovery endpoints.
...
 
-# The account to use for discovering new target containers.
+# The account to use for discovering new targets.
 account: <string>
 
-# The DNS suffix which should be applied to target containers.
+# The type of servers to discover, can be set to:
+# * "vm" to discover virtual machines (zones, bhyve vms) running on triton
+# * "gz" to discover global zones (head/compute nodes) making up the triton infrastructure
+[ server_type: <string> | default = "vm" ]
```
brian-brazil, line 1044: Is server type really the term Triton uses for this?

jzinkweg: No, I just can't think of a better descriptor, perhaps something like discovery_type would be more fitting.

brian-brazil: role is what we use elsewhere, but I'm not quite sure that fits here.

jzinkweg: It seems the `role` as used in openstack is very similar and at least it's consistent within the prometheus config blocks. This would be a better name than server_type.

As to the values, I'm unhappy with the way triton is mixing their names, in the admin console it uses "server" and "virtual machine", but in the triton-cmon implementation, the roles are called 'container' and 'cn' (compute node) for the two discovery endpoints:
https://github.com/joyent/triton-cmon/blob/master/lib/endpoints/discover.js#L145

Neither solution matches reality as the vm/container instances are smartos zones but this name not seen in either location :-/
Although "container" is an overloaded term, I think it would be best to stick with the names used in triton-cmon as that is what the administrator will be looking at while configuring prometheus.

brian-brazil: Let's keep it consistent then, and put in a comment in the code as to which way we're going here. We can also mention the alternate names in the docs.

jzinkweg: `server_type` is changed to `role` and the docs are now a mirror of the openstack setup whith two lists and I've include all three variants of the names used.

I've added a comment block to the top of the refresh method:
https://github.com/prometheus/prometheus/pull/7250/files#diff-8f9f3d2f206c0f47ca80fb105f06f34bR153

discovery/triton/triton.go
jzinkweg: It's all container now :)

discovery/triton/triton.go
```diff
@@ -54,6 +55,7 @@ var DefaultSDConfig = SDConfig{
 // SDConfig is the configuration for Triton based service discovery.
 type SDConfig struct {
 	Account         string                `yaml:"account"`
+	Role            string                `yaml:"role,omitempty"`
```
brian-brazil, line 58: Given the default is non-empty, we probably don't want the omittempty here as if it ever applied it'd likely just cause confusion.

docs/configuration/configuration.md
```diff
@@ -1020,37 +1020,58 @@ Serverset data must be in the JSON format, the Thrift format is not currently su
 scrape targets from [Container Monitor](https://github.com/joyent/rfd/blob/master/rfd/0027/README.md)
 discovery endpoints.
 
-The following meta labels are available on targets during relabeling:
+One of the following `<triton_role>` types can be configured to discover targets:
+
+#### `container`
+
+The `container` role discovers one target per "virtual machine" owned by the `account`. These are smartos zones or lx/kvm/bhyve branded zones.
```
brian-brazil, line 1027: Isn't it SmartOS?

docs/configuration/configuration.md
```diff
@@ -1020,37 +1020,58 @@ Serverset data must be in the JSON format, the Thrift format is not currently su
...
 * `__meta_triton_machine_brand`: the brand of the target container
 * `__meta_triton_machine_id`: the UUID of the target container
-* `__meta_triton_machine_image`: the target containers image type
-* `__meta_triton_server_id`: the server UUID for the target container
+* `__meta_triton_machine_image`: the target container's image type
+* `__meta_triton_server_id`: the server UUID the target container is running on
+
+#### `cn`
+
+The `cn` role discovers one target for per compute node (server/global zone) making up the triton infrastructure. The `account` must be a triton operator and is currently required to own at least one `container`.
```
brian-brazil, line 1040: (also known as ....) may be a bit clearer

docs/configuration/configuration.md
```diff
@@ -1020,37 +1020,58 @@ Serverset data must be in the JSON format, the Thrift format is not currently su
...
+
+#### `container`
+
+The `container` role discovers one target per "virtual machine" owned by the `account`. These are smartos zones or lx/kvm/bhyve branded zones.
+
+The following meta labels are available on targets during [relabeling](#relabel_config):
 
 * `__meta_triton_groups`: the list of groups belonging to the target joined by a comma separator
 * `__meta_triton_machine_alias`: the alias of the target container
 * `__meta_triton_machine_brand`: the brand of the target container
```
brian-brazil, line 1033: Hmm, could this have a value of cn in the container role?

Generally if the cn role doesn't actually have a brand, we shouldn't force it.

jzinkweg: No, currently brand would be one of "joyent" (possibly with -suffix), "lx", "kvm" or "bhyve".
The intent was to expose that this is a compute node, if there is a better way to do this that would be fine too.

discovery/triton/triton.go
jzinkweg: I changed this when falling back to the default config seemed to be broken, but I can't reproduce this now. Thinking this over again I don't think it actually makes a difference for loading yaml files as the field will be set to null anyway.
I'll remove the flag as it's pointless.

docs/configuration/configuration.md
jzinkweg: Yes, KVM also needs to be upper-case to stick with branding.

docs/configuration/configuration.md
brian-brazil: That sounds like we're making up data then, if there's no brand coming from the cn mode then let's no add it.

jzinkweg: I've dropped the fixed-value brand from compute nodes.

docs/configuration/configuration.md
jzinkweg: Done!

docs/configuration/configuration.md
jzinkweg: Changed triton -> Triton where used in text, same for SmartOS and KVM.

brian-brazil:
Thanks!

PR #7198, opened 2020-05-04
promql/engine: Introduce SelectManager to control execution of .Select requests

kakkoyun:
Signed-off-by: Kemal Akkoyun <kakkoyun@gmail.com>

This PR attempts to fix https://github.com/prometheus/prometheus/issues/6878

Currently, Prometheus executes PromQL “.Select” queries sequentially. Particularly for the remote storages, this implementation wastes CPU cycles while waiting for independent queries.

Requirements that gathered from the comments of https://github.com/prometheus/prometheus/issues/6878 and https://github.com/prometheus/prometheus/pull/6887.

#### Requirements
* Should NOT use all the available CPU.
  * Limit concurrency
* PromQL should not care about how the querier is implemented, it should always be the same logic.
* Should avoid introducing changes to LOCAL storage.
* Should avoid spawning goroutines if there is just a single selector.

This PR fulfils the requirements listed above by introducing a small component to regulate select requests against querier.

It still misses tests and also I'm gonna add benchmarks.

P.S: Regarding https://github.com/prometheus/prometheus/pull/7089#issuecomment-617132242 after checking out the comments, I decided a design doc might not be needed.

<!--
    Don't forget!
    
    - If the PR adds or changes a behaviour or fixes a bug of an exported API it would need a unit/e2e test.
    
    - Where possible use only exported APIs for tests to simplify the review and make it as close as possible to actual library usage.
    
    - No tests are needed for internal implementation changes.
    
    - Performance improvements would need a benchmark test to prove it.
    
    - All exposed objects should have a comment.
    
    - All comments should start with a capital letter and end with a full stop.
 -->

### Benchmarks

#### Old
<details>Master
<p>

```
goos: darwin
goarch: amd64
pkg: github.com/prometheus/prometheus/promql
BenchmarkRangeQuery/expr=a_one,steps=1-12         	   75193	     17112 ns/op	    5235 B/op	     101 allocs/op
BenchmarkRangeQuery/expr=a_one,steps=10-12        	   74503	     17711 ns/op	    5235 B/op	     101 allocs/op
BenchmarkRangeQuery/expr=a_one,steps=100-12       	   50686	     23400 ns/op	    5394 B/op	     103 allocs/op
BenchmarkRangeQuery/expr=a_one,steps=1000-12      	   12404	     80742 ns/op	    7238 B/op	     114 allocs/op
BenchmarkRangeQuery/expr=a_ten,steps=1-12         	   13332	     89027 ns/op	   12116 B/op	     195 allocs/op
[...]

PR #7251, opened 2020-05-12
*: Consistent Error/Warning handling for SeriesSet iterator: Allowing Async Select

kakkoyun:
Signed-off-by: Kemal Akkoyun <kakkoyun@gmail.com>

As discussed in https://github.com/prometheus/prometheus/pull/7198, this PR attempts to fix https://github.com/prometheus/prometheus/issues/6878, using deferred SeriesSet iterator.

Currently, Prometheus executes PromQL “.Select” queries sequentially. Particularly for the remote storages, this implementation wastes CPU cycles while waiting for independent queries.

Requirements that gathered from the comments of https://github.com/prometheus/prometheus/issues/6878 and https://github.com/prometheus/prometheus/pull/6887.

#### Requirements
* Should NOT use all the available CPU.
  * Limit concurrency
* PromQL should not care about how the querier is implemented, it should always be the same logic.
* Should avoid introducing changes to LOCAL storage.
* Should avoid spawning goroutines if there is just a single selector.

#### Changes
This PR fulfils the requirements listed above by introducing changes to `storage.SeriesSet`, `storage.Querier`, `promql/engine` and `storage/remote`.
* Add `Warnings()` to `storage.SeriesSet` https://github.com/prometheus/prometheus/commit/18dddfb771f0a8711e9158dc624bd8d18beaf544
* Only return `storage.SeriesSet` from `storage.Querier.Select()` https://github.com/prometheus/prometheus/commit/354a74630056f5b58251376eb96888cc22f1d1c5
* Modify `promql/engine` to propogate errors at evaluation stage. https://github.com/prometheus/prometheus/commit/72203b0f7b29bb83fa5716ae026cd576428039da
* And adds async implementation of SeriesSet to remote reads https://github.com/prometheus/prometheus/commit/1c28c9fb31ee4dee5c4a54f204436c9d2c1e815e

This PR could be separated into smaller PRs if needed.

#### TODO
* [ ] Benchmarks


promql/engine.go
```diff
@@ -997,29 +1005,32 @@ func (ev *evaluator) rangeEval(f func([]parser.Value, *EvalNodeHelper) Vector, e
...
-	vs := &parser.VectorSelector{
-		Offset: subq.Offset,
-		Series: make([]storage.Series, 0, len(val)),
-	}
-	ms := &parser.MatrixSelector{
-		Range:          subq.Range,
-		VectorSelector: vs,
-	}
-	for _, s := range val {
+func (ev *evaluator) evalSubquery(subq *parser.SubqueryExpr) (*parser.MatrixSelector, storage.Warnings) {
```
brian-brazil, line 1013: In a big PR like this, it's best not to make incidental style changes like in this function. It makes the diff bigger than it needs to be.

promql/value.go
```diff
@@ -186,12 +186,16 @@ func (m Matrix) ContainsSameLabelset() bool {
 	return false
 }
 
+// Warnings holds evaluation warnings. Currently, it is just a type alias to
```
brian-brazil, line 189: This is not what type aliasing should be used for, using storage.Warnings is fine as we use storage in other parts of the API too.

promql/engine.go
```diff
@@ -648,19 +644,13 @@ func (ng *Engine) findMinTime(s *parser.EvalStmt) time.Time {
...
-func (ng *Engine) populateSeries(ctx context.Context, querier storage.Querier, s *parser.EvalStmt) (storage.Warnings, error) {
-	var (
-		// Whenever a MatrixSelector is evaluated, evalRange is set to the corresponding range.
-		// The evaluation of the VectorSelector inside then evaluates the given range and unsets
-		// the variable.
-		evalRange time.Duration
-		warnings  storage.Warnings
-		err       error
-	)
+func (ng *Engine) populateSeries(ctx context.Context, querier storage.Querier, s *parser.EvalStmt) {
```
bwplotka, line 647: hm? :thinking: 

```suggestion
func (ng *Engine) asyncPopulateSeries(ctx context.Context, querier storage.Querier, s *parser.EvalStmt) {
```

promql/engine.go
```diff
@@ -742,34 +724,48 @@ func extractGroupsFromPath(p []parser.Node) (bool, []string) {
...
 		default:
 		}
 		res = append(res, it.At())
+		ws = append(ws, it.Warnings()...)
 	}
-	return res, it.Err()
+	return res, ws, it.Err()
 }
 
+type errorWithWarnings struct {
```
bwplotka, line 762: What is this? I don't  think we need to preserve warnings on error? cc @brian-brazil what would be your opinion? :thinking:

promql/engine.go
```diff
@@ -880,17 +883,20 @@ func (enh *EvalNodeHelper) signatureFunc(on bool, names ...string) func(labels.L
 // the given function with the values computed for each expression at that
 // step.  The return value is the combination into time series of all the
 // function call results.
-func (ev *evaluator) rangeEval(f func([]parser.Value, *EvalNodeHelper) Vector, exprs ...parser.Expr) Matrix {
+func (ev *evaluator) rangeEval(f func([]parser.Value, *EvalNodeHelper) (Vector, storage.Warnings), exprs ...parser.Expr) (Matrix, storage.Warnings) {
```
bwplotka, line 886: I would expect error as well here :thi

bwplotka:
But overall let's try to minimize changes (:

promql/engine.go
brian-brazil: We don't have to, but it wouldn't hurt.

bwplotka: Cool, maybe it would be good candidate to do in following PR

promql/engine.go
kakkoyun: It doesn't necessarily need to be async. It depends on implementation. It puts an unexpanded SeriesSet to the statement node. I can still come up with a better name.

promql/engine.go
kakkoyun: Error handling is a bit different in engine code. It panics when it encounters error and recovers in a top-level `Eval` method, hence we don't return errors in here and similar helper functions.

promql/engine.go
bwplotka: Cool :+1: It's fine. Not a big deal, populate makes sense, just it does not mean things are populated when function finishes.

kakkoyun:
@bwplotka @brian-brazil I have removed remote read changes will send another PR for those.
Should I split this one as well? I can split it into two: One for SeriesSet changes and one for the Querier and Engine changes. What do you think?

bwplotka:
I think it's fine together. Buildable and complete

liguozhong:
We really need this feature
image

kakkoyun:
@liguozhong There will a subsequent PR to actually implement async remote read. Those pieces removed, not to make this PR any larger.

kakkoyun:
@bwplotka @krasi-georgiev @brian-brazil Would you take another look at this?

promql/engine.go
```diff
@@ -742,34 +724,48 @@ func extractGroupsFromPath(p []parser.Node) (bool, []string) {
 	return false, nil
 }
 
-func checkForSeriesSetExpansion(ctx context.Context, expr parser.Expr) {
+func checkAndExpandSeriesSet(ctx context.Context, expr parser.Expr) storage.Warnings {
 	switch e := expr.(type) {
 	case *parser.MatrixSelector:
-		checkForSeriesSetExpansion(ctx, e.VectorSelector)
+		if ws := checkAndExpandSeriesSet(ctx, e.VectorSelector); len(ws) > 0 {
```
bwplotka, line 730: It's not an error, so we can just return ws and without check return below.

If we do like this it looks like warnings should be returned in certain case which is not true and might mislead here (:

promql/engine.go
```diff
@@ -742,34 +724,48 @@ func extractGroupsFromPath(p []parser.Node) (bool, []string) {
...
 		if e.Series == nil {
-			series, err := expandSeriesSet(ctx, e.UnexpandedSeriesSet)
+			series, ws, err := expandSeriesSet(ctx, e.UnexpandedSeriesSet)
 			if err != nil {
-				panic(err)
+				panic(errorWithWarnings{err: err, warnings: ws})
 			} else {
 				e.Series = series
 			}
+			if len(ws) > 0 {
```
bwplotka, line 741: No need for this, let's put to function level `ws` and return below

promql/engine.go
```diff
@@ -742,34 +724,48 @@ func extractGroupsFromPath(p []parser.Node) (bool, []string) {
...
+func expandSeriesSet(ctx context.Context, it storage.SeriesSet) (res []storage.Series, ws storage.Warnings, err error) {
 	for it.Next() {
 		select {
 		case <-ctx.Done():
-			return nil, ctx.Err()
+			return nil, nil, ctx.Err()
 		default:
 		}
 		res = append(res, it.At())
+		ws = append(ws, it.Warnings()...)
```
bwplotka, line 757: Curious, why `.Warnings` is not per SeriesSet? Like `Err`? I think that would enough and avoid some overhead if there are 1 millions series... WDYT @brian-brazil ? 

This would be similar to the existing flow then.

kakkoyun: It's per series set. This a mistake. I'll fix it.

bwplotka: Got offline  info, it's bug only, so all good after we fix this

web/api/v1/api.go
```diff
@@ -1207,10 +1204,7 @@ func (api *API) remoteRead(w http.ResponseWriter, r *http.Request) {
 		for i, query := range req.Queries {
 			err := api.remoteReadQuery(ctx, query, externalLabels, func(querier storage.Querier, hints *storage.SelectHints, filteredMatchers []*labels.Matcher) error {
 				// The streaming API has to provide the series sorted.
-				set, _, err := querier.Select(true, hints, filteredMatchers...)
-				if err != nil {
-					return err
-				}
+				set := querier.Select(true, hints, filteredMatchers...)
 
 				return remote.StreamChunkedReadResponses(
```
brian-brazil, line 1209: This doesn't seem to be handling warnings.

web/api/v1/api.go
```diff
@@ -1239,10 +1233,7 @@ func (api *API) remoteRead(w http.ResponseWriter, r *http.Request) {
 		}
 		for i, query := range req.Queries {
 			err := api.remoteReadQuery(ctx, query, externalLabels, func(querier storage.Querier, hints *storage.SelectHints, filteredMatchers []*labels.Matcher) error {
-				set, _, err := querier.Select(false, hints, filteredMatchers...)
-				if err != nil {
-					return err
-				}
+				set := querier.Select(false, hints, filteredMatchers...)
```
brian-brazil, line 1236: Ditto

storage/fanout.go
```diff
@@ -471,26 +453,45 @@ func newGenericMergeSeriesSet(sets []genericSeriesSet, querier *mergeGenericQuer
...
 		mergeFunc: mergeFunc,
-		heap:      h,
 		sets:      sets,
 		querier:   querier,
 	}
 }
 
 func (c *genericMergeSeriesSet) Next() bool {
+	if c.heap == nil {
+		// Sets need to be pre-advanced, so we can introspect the label of the
```
brian-brazil, line 465: This is only advancing this select, we need to advance all the selects done against the querier to spot if any of them failed so that we can filter out responses from remotes that fail some of their requests.

liguozhong:
> @liguozhong There will a subsequent PR to actually implement async remote read. Those pieces removed, not to make this PR any larger.

Thanks, looking forward to seeing this feature in the 2.19.0 release.

web/api/v1/api.go
kakkoyun: Previously, warnings were being ignored. How should we proceed with it in here? Should we log warnings? Should we include the warnings in the response?

storage/fanout.go
kakkoyun: Could you please elaborate this one a bit more?

web/api/v1/api.go
brian-brazil: They shouldn't be even happening currently and remote read doesn't support them, so logging is probably fine.

promql/engine.go
bwplotka: Yea agree, especially that  the logic is different: Series: make([]storage.Series, 0, len(mat)),

promql/engine.go
```diff
@@ -1203,20 +1215,23 @@ func (ev *evaluator) eval(expr parser.Expr) parser.Value {
...
 
-		return mat
+		return mat, warnings
 
 	case *parser.ParenExpr:
 		return ev.eval(e.Expr)
 
 	case *parser.UnaryExpr:
-		mat := ev.eval(e.Expr).(Matrix)
+		var (
```
bwplotka, line 1231: Sometimes we do `var(` sometimes direct `:=` , let's pick one (:

storage/fanout.go
```diff
@@ -257,53 +257,32 @@ func NewMergeChunkQuerier(primaryQuerier ChunkQuerier, queriers []ChunkQuerier,
...
+			queryResults <- &queryResult{qr: qr, set: set}
 		}(querier)
 	}
 	for i := 0; i < len(q.queriers); i++ {
-		qryResult := <-queryResultChan
-		q.setQuerierMap[qryResult.set] = qryResult.qr
-		if qryResult.wrn != nil {
-			warnings = append(warnings, qryResult.wrn...)
-		}
-		if qryResult.selectError != nil {
```
bwplotka, line 291: We might want to ensure correct logic of this in the remote read Select implementation (:

storage/fanout.go
bwplotka: I think what Brian is saying you exactly did below @kakkoyun , so we make sure `Next` is done so we ensure no error is there.

However TBH we  can simplify this whole thing by one simple trick: We can make sure remote read Select will behave like this (if select fails we only return warning not error). This will make the flow super easy (: We can name it even `nonErrReadQuerier` or something.

storage/fanout.go
```diff
@@ -502,6 +506,9 @@ func (c *genericMergeSeriesSet) Next() bool {
 			if set.Next() {
 				heap.Push(&c.heap, set)
 			}
+			if err := set.Err(); err != nil && c.err == nil {
```
bwplotka, line 509: I think we can skip those, and just do return false in case of error and then allow for loop to loop through `set.Err` as we had before, just easier ..

bwplotka: No a blocker here, up to you

storage/interface.go
```diff
@@ -154,6 +154,8 @@ type SeriesSet interface {
 	Next() bool
 	At() Series
 	Err() error
+	// A collection of warnings for the whole set.
```
bwplotka, line 157: Let's to the same for Err. 

Also let's mention that it also is there for both error and non error sets

storage/remote/read.go
```diff
@@ -116,13 +116,10 @@ type externalLabelsQuerier struct {
...
 // removed from the returned series sets.
-func (q externalLabelsQuerier) Select(sortSeries bool, hints *storage.SelectHints, matchers ...*labels.Matcher) (storage.SeriesSet, storage.Warnings, error) {
+func (q externalLabelsQuerier) Select(sortSeries bool, hints *storage.SelectHints, matchers ...*labels.Matcher) storage.SeriesSet {
 	m, added := q.addExternalLabels(matchers)
-	s, warnings, err := q.Querier.Select(sortSeries, hints, m...)
-	if err != nil {
-		return nil, warnings, err
-	}
-	return newSeriesSetFilter(s, added), warnings, nil
+	s := q.Querier.Select(sortSeries, hints, m...)
```
bwplotka, line 121: Let's inline I think (:

tsdb/querier.go
```diff
@@ -125,31 +119,22 @@ type verticalQuerier struct {
...
-	if err != nil {
-		return nil, ws, err
-	}
-	b, w, err := q.sel(sortSeries, hints, qs[l:], ms)
-	ws = append(ws, w...)
-	if err != nil {
-		return nil, ws, err
-	}
-	return newMergedVerticalSeriesSet(a, b), ws, nil
+	a := q.sel(sortSeries, hints, qs[:l], ms)
```
bwplotka, line 135: inline please if we can as per https://thanos.io/contributing/coding-style-guide.md/#avoid-defining-variables-used-only-once :hugs:

tsdb/querier.go
```diff
@@ -509,9 +499,11 @@ func (s *mergedSeriesSet) nextAll() {
 	for _, ss := range s.all {
 		if ss.Next() {
 			s.buf = append(s.buf, ss)
+			s.ws = append(s.ws, ss.Warnings()...)
```
bwplotka, line 502: :+1:

tsdb/querier.go
```diff
@@ -509,9 +499,11 @@ func (s *mergedSeriesSet) nextAll() {
...
 		} else if ss.Err() != nil {
```
bwplotka, line 503: I would vote for no else but `continue` instead (: https://thanos.io/contributing/coding-style-guide.md/#control-structure-prefer-early-returns-and-avoid-else

web/api/v1/api.go
bwplotka: Yea, we talked about this in chunk iterator "in progress" PRs, I think just TODO is fine for now

bwplotka: I think back then we said it will be too spammy  :thinking:

storage/fanout.go
brian-brazil: You're still only advancing the sets of this Select, not the sets of all Selects.

tsdb/querier.go
```diff
@@ -132,9 +132,10 @@ func (q *verticalQuerier) sel(sortSeries bool, hints *storage.SelectHints, qs []
 	}
 	l := len(qs) / 2
 
-	a := q.sel(sortSeries, hints, qs[:l], ms)
-	b := q.sel(sortSeries, hints, qs[l:], ms)
-	return newMergedVerticalSeriesSet(a, b)
+	return newMergedVerticalSeriesSet(
```
bwplotka, line 135: :+1:

storage/fanout.go
```diff
@@ -309,11 +318,27 @@ func (q *mergeGenericQuerier) LabelValues(name string) ([]string, Warnings, erro
 	return mergeStringSlices(results), warnings, nil
 }
 
+func (q *mergeGenericQuerier) MarkAsFailedSet(set genericSeriesSet) {
+	q.failedQueriers[q.setQuerierMap[set]] = struct{}{}
+}
+
+func (q *mergeGenericQuerier) DoesBelongToPrimary(set genericSeriesSet) bool {
```
bwplotka, line 325: does those have to be exported? :thinking:

bwplotka: and if yes, let's comment

storage/fanout.go
```diff
@@ -282,7 +284,14 @@ func (q *mergeGenericQuerier) Select(sortSeries bool, hints *SelectHints, matche
...
 		seriesSets = append(seriesSets, res.set)
 	}
-	return newGenericMergeSeriesSet(seriesSets, q, q.mergeFunc)
+
+	set := &genericMergeSeriesSet{
+		querier:   q,
+		mergeFunc: q.mergeFunc,
+		sets:      seriesSets,
+	}
+	q.mergeSeriesSets = append(q.mergeSeriesSets, set)
```
bwplotka, line 293: Can you explain here what is happening here? I don't follow this code.. why have to remember the set during `Select` in `mergeSeriesSets`. This sounds wrong as Select can be done multiple times.

storage/fanout.go
brian-brazil: Same question from me, these all appear to be internal stuff

storage/fanout.go
```diff
@@ -429,7 +433,12 @@ type genericMergeSeriesSet struct {
 	sets []genericSeriesSet
 
 	currentSets []genericSeriesSet
-	querier     *mergeGenericQuerier
+
+	// 'querier' is optional and can be nil. Pass Querier if you want to retry query in case of failing series set.
```
brian-brazil, line 437: We never retry, why is this here?

storage/fanout.go
```diff
@@ -457,40 +466,67 @@ func NewMergeChunkSeriesSet(sets []ChunkSeriesSet, merger VerticalChunkSeriesMer
...
-	return &genericMergeSeriesSet{
-		mergeFunc: mergeFunc,
-		heap:      h,
-		sets:      sets,
-		querier:   querier,
-	}
 }
 
 func (c *genericMergeSeriesSet) Next() bool {
+	if c.querier != nil { // since querier is optional.
```
brian-brazil, line 521: Since

storage/fanout.go
```diff
@@ -330,11 +318,27 @@ func (q *mergeGenericQuerier) LabelValues(name string) ([]string, Warnings, erro
...
+func (q *mergeGenericQuerier) DoesBelongToPrimary(set genericSeriesSet) bool {
+	return reflect.DeepEqual(q.setQuerierMap[set], q.primaryQuerier)
+}
+
 func (q *mergeGenericQuerier) IsFailedSet(set genericSeriesSet) bool {
 	_, isFailedQuerier := q.failedQueriers[q.setQuerierMap[set]]
 	return isFailedQuerier
 }
 
+func (q *mergeGenericQuerier) AdvanceAllSets() {
```
brian-brazil, line 334: This is only used once, you can inline this.

storage/fanout.go
```diff
@@ -330,11 +318,27 @@ func (q *mergeGenericQuerier) LabelValues(name string) ([]string, Warnings, erro
 	return mergeStringSlices(results), warnings, nil
 }
 
+func (q *mergeGenericQuerier) MarkAsFailedSet(set genericSeriesSet) {
```
brian-brazil, line 321: This is also only used once

storage/fanout.go
kakkoyun: It doesn't have to be. I've just followed the same fashion with `IsFailedSet`. It's also not used outside `fanout`. Maybe be better to un-export `IsFailedSet`?

storage/fanout.go
kakkoyun: This is a moved comment from a factor function to the struct. I believe it's an outdated one, I'll remove.

storage/fanout.go
kakkoyun: I've removed and inlined `IsFailedSet` as well.

storage/fanout.go
kakkoyun: As @brian-brazil pointed out, we need to make sure all sets from all selects are succeeded https://github.com/prometheus/prometheus/pull/7251#discussion_r426546061 before we merge sets, otherwise we ignore to sets that come from that querier. Thus I added this to remember sets.

Also I've added a comment for this.

kakkoyun:
> Sorry, but I don't follow last commit ): It might need some explanation or redesign. What do you want to achieve? What was wrong without this commit?

I have added the last two commits to address the issue that has raised by Brian. https://github.com/prometheus/prometheus/pull/7251#discussion_r426546061

I'm happy to come up with a better design if it doesn't match the requirements. And also I've already addressed the issue with exported methods by removing and inlining them.

storage/fanout.go
```diff
@@ -539,14 +567,28 @@ func (c *genericMergeSeriesSet) At() Labels {
...
 		if err := set.Err(); err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
+func (c *genericMergeSeriesSet) Warnings() Warnings {
+	for _, set := range c.sets {
+		if ws := set.Warnings(); len(ws) > 0 {
```
brian-brazil, line 585: Hmm, do we need the if here? I'd expect append to deal with the empty variadic.

promql/engine.go
```diff
@@ -742,34 +724,43 @@ func extractGroupsFromPath(p []parser.Node) (bool, []string) {
...
+			return nil, nil, ctx.Err()
 		default:
 		}
 		res = append(res, it.At())
 	}
-	return res, it.Err()
+	return res, it.Warnings(), it.Err()
 }
 
+type errorWithWarnings struct {
```
bwplotka, line 757:
```suggestion
type errWithWarnings struct {
```

Tiny nit

rules/manager.go
```diff
@@ -727,6 +722,13 @@ func (g *Group) RestoreForState(ts time.Time) {
 				}
 			}
 
+			if err := sset.Err(); err != nil {
+				// Querier Warnings are ignored. We do not care unless we have an error.
+				level.Error(g.logger).Log("msg", "Failed to restore 'for' state",
```
bwplotka, line 727: Incorrent formatting ):

storage/fanout.go
bwplotka: Is this addressed?

storage/fanout.go
```diff
@@ -457,40 +446,76 @@ func NewMergeChunkSeriesSet(sets []ChunkSeriesSet, merger VerticalChunkSeriesMer
...
-			continue
-		}
-		if set.Next() {
-			heap.Push(&h, set)
+	mset.advance()
+	return mset
+}
+
+func (c *genericMergeSeriesSet) advance() {
+	if !c.advanced {
```
bwplotka, line 473: This has unnecessary indent.. (: Can you just return if `c.advanced?`

bwplotka: This is a code smell as well. Why would you `advance` if it was advanced before? It's better if caller will make sure this will not happen, otherwise we can end up in unnecessary  invoking `advance` thousand times `just in case`. This creates confusion and is extremely hard to follow. I would fix this TBH before merging.

storage/fanout.go
```diff
@@ -457,40 +446,76 @@ func NewMergeChunkSeriesSet(sets []ChunkSeriesSet, merger VerticalChunkSeriesMer
...
+		for _, set := range c.sets {
+			if set == nil {
+				continue
+			}
+			if set.Next() {
+				heap.Push(&c.heap, set)
+			}
+
+			if err := set.Err(); err != nil {
+				if c.querier != nil { // Since querier is optional.
```
bwplotka, line 485: Wait, so if querier is nil, then we ignore error? That sounds bad.

storage/fanout.go
kakkoyun: Errors are returned by remote read are now wrapped with `ErrSeriesSet`.

```go
// Select implements storage.Querier and uses the given matchers to read series sets from the Client.
func (q *querier) Select(sortSeries bool, hints *storage.SelectHints, matchers ...*labels.Matcher) storage.SeriesSet {
	query, err := ToQuery(q.mint, q.maxt, matchers, hints)
	if err != nil {
		return storage.ErrSeriesSet(err)
	}

	remoteReadGauge := remoteReadQueries.WithLabelValues(q.client.remoteName, q.client.url.String())
	remoteReadGauge.Inc()
	defer remoteReadGauge.Dec()

	res, err := q.client.Read(q.ctx, query)
	if err != nil {
		return storage.ErrSeriesSet(fmt.Errorf("remote_read: %v", err))
	}

	return FromQueryResult(sortSeries, res)
}
```

Is this what you meant?

kakkoyun:
@bwplotka That'd awesome, let's review it closer together.

storage/fanout.go
```diff
@@ -512,8 +537,11 @@ func (c *genericMergeSeriesSet) Next() bool {
 		c.currentLabels = c.heap[0].At().Labels()
 		for len(c.heap) > 0 && labels.Equal(c.currentLabels, c.heap[0].At().Labels()) {
 			set := heap.Pop(&c.heap).(genericSeriesSet)
-			if c.querier != nil && c.querier.IsFailedSet(set) {
```
kakkoyun, line 515: @bwplotka To address your comment in here https://github.com/prometheus/prometheus/pull/7251#discussion_r430309497, querier was optional previously.

storage/fanout.go
```diff
@@ -457,40 +446,76 @@ func NewMergeChunkSeriesSet(sets []ChunkSeriesSet, merger VerticalChunkSeriesMer
...
-	return &genericMergeSeriesSet{
-		mergeFunc: mergeFunc,
-		heap:      h,
-		sets:      sets,
-		querier:   querier,
-	}
 }
 
 func (c *genericMergeSeriesSet) Next() bool {
+	if c.querier != nil { // Since querier is optional.
```
bwplotka, line 503: Why we check `c.querier` twice?

bwplotka:
As promised: https://github.com/kakkoyun/prometheus/pull/3

kakkoyun:
Just adding the last comment that I had in https://github.com/kakkoyun/prometheus/pull/3 in case we forget.

> Looks great much easier on the eyes :)
> I had a quick look at it and I want to mediate on it a bit more.
> 
> Besides the nits, my major concern is; this removes the ability to exec async request against queriers.
> Both mergeSeriesSet and secondarySeriesSet constructed with newGenericMergeSeriesSet() and it advances all the set at least once. So fanned out .Select calls are must be completed before leaving the overarching .Select call. Thus I had put the deferred set wrapper in place.
> 
> What do you think?

bwplotka:
Ok I think we got consensus offline with @kakkoyun 

The idea for implementation of async/lazy series set is to schedule request in `Select` and do blocking receive on first `Next` With current code we can do exactly that :hugs: 

This should work well. Still, I am adding some last minutes tests for error handling check, but overall it should be good for review @brian-brazil :hugs:

bwplotka:
> Where has the logic gone that if a Select for a secondary fails, then all other Selects for that secondary must not return anything?

Ah that's why this was needed .... =D Now all clear will add back with proper comment and tests against this requirement :+1: 

Thanks for review @brian-brazil 

(I am on it as @kakkoyun is OOO today)

bwplotka:
Tried to summarise our confusion part here:

## Current Flow
Prometheus PromQL v2 19 Flow

## Partial Result Issue
Prometheus PromQL_ Partial Response issue

storage/remote/storage.go
```diff
@@ -145,6 +145,7 @@ func (s *Storage) Querier(ctx context.Context, mint, maxt int64) (storage.Querie
 		}
 		queriers = append(queriers, q)
 	}
+	// All of those queriers are considered best effort, so we pass them as secondary queriers.
```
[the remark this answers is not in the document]
bwplotka: Why? This is true, all of them are secondary (like it was before)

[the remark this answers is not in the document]
brian-brazil: This comment has nothing to do with the code that's here.

bwplotka:
All should be addressed. I only need to add a few more test use cases, will add tomorrow. Otherwise PTAL @kakkoyun @brian-brazil  (:

storage/fanout.go
```diff
@@ -188,153 +187,108 @@ func (f *fanoutAppender) Rollback() (err error) {
...
+	go func() {
+		wg.Wait()
+		close(seriesSetChan)
+	}()
+
+	for r := range seriesSetChan {
+		seriesSets = append(seriesSets, r)
 	}
-	return newGenericMergeSeriesSet(seriesSets, q, q.mergeFunc), warnings, nil
+	return newGenericMergeSeriesSet(seriesSets, q.mergeFn)
```
kakkoyun, line 269: @bwplotka Even though `Select` from `SecondQuerier` returns a lazy set, in `newGenericMergeSeriesSet` we call `Next()` to advance sets so we unwrap lazy set before we leave this function. I don't think this is what we want, is it?

storage/secondary.go
```diff
@@ -13,25 +13,39 @@
...
+
+// secondaryQuerier is a wrapper that allows a querier to be treated in a best effort manner.
+// This means that an error on any method returned by Querier except Close will be returned as a warning,
+// and the result will be empty.
+//
+// Additionally, Querier ensures that if ANY SeriesSet returned by this querier's Select failed on an initial Next,
+// All other SeriesSet will be return no response as well. This ensures consistent partial response strategy, where you
+// have either full results or none from each secondary Querier.
+// NOTE: This works well only for implementations that only fail during first Next() (e.g fetch from network). If implementation fails
+// during further iterations, set will panic. If Select is invoked after first Next of any returned SeriesSet, querier will panic.
```
brian-brazil, line 31: Select will panic?

storage/secondary.go
```diff
@@ -50,18 +64,88 @@ func (s *secondaryQuerier) LabelNames() ([]string, Warnings, error) {
...
+func (s *secondaryQuerier) createFn(asyncSet genericSeriesSet) func() (genericSeriesSet, bool) {
+	s.asyncSets = append(s.asyncSets, asyncSet)
+	curr := len(s.asyncSets) - 1
+	return func() (genericSeriesSet, bool) {
+		s.once.Do(func() {
+			// At first create invocation we iterate over all sets and ensure its Next() returns some value without
+			// errors. This is to ensure we support consistent partial failures.
+			var errs tsdb_errors.MultiError
+			for i, set := range s.asyncSets {
+				if set.Next() {
```
brian-brazil, line 76: Hmm, might this lose us the first series?

bwplotka: Yea, not as we lazySeriesSet cares for that

storage/secondary.go
bwplotka: Yes, what else to do if Select will be run after `once` will already complete?

brian-brazil: My point was s/set/Select/

storage/fanout_test.go
```diff
@@ -596,3 +639,261 @@ func BenchmarkMergeSeriesSet(b *testing.B) {
...
+				&secondaryQuerier{genericQuerier: &mockGenericQuerier{resp: []string{"c"}, warnings: nil, err: nil}},
+			},
+			expectedSelectsSeries: []labels.Labels{
+				labels.FromStrings("test", "b"),
+				labels.FromStrings("test", "c"),
+			},
+			expectedLabels: []string{"b", "c"},
+		},
+		{
+			// TODO(kakkoyun): ?
```
bwplotka, line 796: looks like some left over comment? The case looks good now.

rules/manager.go
```diff
@@ -727,6 +722,17 @@ func (g *Group) RestoreForState(ts time.Time) {
 				}
 			}
 
+			if err := sset.Err(); err != nil {
```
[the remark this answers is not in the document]
kakkoyun: Yes, apparently we haven't.

storage/fanout.go
```diff
@@ -37,8 +38,15 @@ type fanout struct {
 	secondaries []Storage
 }
 
-// NewFanout returns a new fan-out Storage, which proxies reads and writes
+// NewFanout returns a new fanout Storage, which proxies reads and writes
 // through to multiple underlying storages.
+//
+// The difference between primary and secondary Storage is only for read (Querier) path and it goes as follows:
+// * If the primary querier returns an error, the any of the Querier operation will fail.
```
brian-brazil, line 45: then any of the Querier's operations

storage/fanout.go
```diff
@@ -37,8 +38,15 @@ type fanout struct {
...
+// * If the any secondary querier returns an error the result is discarded, fanout returns only warnings but the Querier
```
brian-brazil, line 46: If any secondary querier returns an error the result from that queries is discarded. The overall operation will succeed, and the error from the secondary querier will be returned as a warning.

storage/secondary.go
```diff
@@ -0,0 +1,107 @@
...
+					errs.Add(err)
+					ws = append([]error{err}, ws...)
+				}
+				s.asyncSets[i] = warningsOnlySeriesSet(ws)
+			}
+
+			if errs.Err() != nil {
+				// One failed, ensure all sets returns nothing. (All or nothing logic).
+				for i, set := range s.asyncSets {
+					s.asyncSets[i] = warningsOnlySeriesSet(set.Warnings())
```
brian-brazil, line 91: Hmm, would this mean the warnings will be duplicated many times to the user? Might we want to do a noop series set for the others instead?

bwplotka: This is very valid point, let's do that and ensure on our new tests. I remember I wanted to do it , but forgot.

PR #7252, opened 2020-05-12
Update alerting_rules.md

justinwalz:
Docs only change. Fix `ALERTS` example to use regex.

### DIFF
--- a/docs/configuration/alerting_rules.md
+++ b/docs/configuration/alerting_rules.md
@@ -86,7 +86,7 @@ the "Alerts" tab of your Prometheus instance. This will show you the exact
 label sets for which each defined alert is currently active.
 
 For pending and firing alerts, Prometheus also stores synthetic time series of
-the form `ALERTS{alertname="<alert name>", alertstate="pending|firing", <additional alert labels>}`.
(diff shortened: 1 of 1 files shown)


roidelapluie:
This is not a promql expression, therefore I don't think this should be fixed.

justinwalz:
Interesting. Why's that? 

Someone could be reading the docs, see this expression, copy and paste it into their expression executor, run it, and not seeing anything.

Copy:
```
ALERTS{alertname="<alert name>", alertstate="pending|firing", <additional alert labels>}
```
Paste and Adjust to:
```
ALERTS{alertstate="pending|firing"}
```

No results, head-scratch. Or simply believe they don't have any firing or pending alerts.

roidelapluie:
Mmh indeed it does not seem harmful.

roidelapluie:
We have discussed this in the bug scrub and decided that in both case this is confusing, we should improve that part of the documentation to make it more meta-syntactic.

Something like ` alertstate="<pending or firing>" `.

Are you willing to update this pull request in that direction? Thanks!

justinwalz:
Sure thing, that works for me. Updated.

brian-brazil:
Thanks!

PR #7256, opened 2020-05-13
API: Allow AlertmanagerRetriever and RulesRetriever to receive a Context

gotjosh:
This is a follow up from #7125 and expands on the discussion #7103.

Initially, #7125 only took into account `TargetRetriever` but the same can be applied to other dependencies. Given this pattern is quite unobtrusive and only belongs in the API, I hope this is better suited than #4999

cc: @brian-brazil 

### DIFF
--- a/web/api/v1/api_test.go
+++ b/web/api/v1/api_test.go
@@ -191,6 +191,12 @@ func (t testAlertmanagerRetriever) DroppedAlertmanagers() []*url.URL {
 	}
 }
 
+func (t testAlertmanagerRetriever) toFactory() func(context.Context) AlertmanagerRetriever {
+	return func(context.Context) AlertmanagerRetriever {
+		return t
+	}
+}
+
 type rulesRetrieverMock struct {
 	testing *testing.T
 }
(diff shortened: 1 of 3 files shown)


web/web.go
```diff
@@ -300,7 +300,13 @@ func New(logger log.Logger, o *Options) *Handler {
 	factoryTr := func(_ context.Context) api_v1.TargetRetriever {
 		return h.scrapeManager
 	}
-	h.apiV1 = api_v1.NewAPI(h.queryEngine, h.storage, factoryTr, h.notifier,
+	factoryAr := func(_ context.Context) api_v1.AlertmanagerRetriever {
+		return h.notifier
+	}
+	FactoryRr := func(_ context.Context) api_v1.RulesRetriever {
```
bwplotka, line 306: All of this should be just inlined, especially if you can have just one line (:  WDYT? No need for this explicitness.

web/api/v1/api_test.go
```diff
@@ -191,6 +191,12 @@ func (t testAlertmanagerRetriever) DroppedAlertmanagers() []*url.URL {
 	}
 }
 
+func (t testAlertmanagerRetriever) toFactory() func(context.Context) AlertmanagerRetriever {
```
bwplotka, line 194: I would vote to just inline, it's just used once....

bwplotka: less LOC

web/web.go
bwplotka: + less LOC if you do so

gotjosh: Happy to, done.

web/api/v1/api_test.go
gotjosh: `gofmt` only allowed inlining of the inner function on this one, did that.

bwplotka:
Thanks!

PR #7259, opened 2020-05-14
Makefile.common: add 'update-go-deps' target

simonpasquier:
Also updated the RELEASE.md document to adjust the instructions about
dependencies management.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -150,6 +150,18 @@ else
 	$(GO) get $(GOOPTS) -t ./...
 endif
 
+.PHONY: update-deps
+update-deps:
+	@echo ">> updating Go dependencies"
+	@jq --version 1>/dev/null 2>&1 || { echo "Make sure that 'jq' is in your PATH"; exit 1; }
+	@for m in $$(go list -m -json all | jq -rc 'select(.Indirect != true and .Main != true) | .Path'); do \
+		go get $$m; \
+	done
+	GO111MODULE=$(GO111MODULE) $(GO) mod tidy
+ifneq (,$(wildcard vendor))
+	GO111MODULE=$(GO111MODULE) $(GO) mod vendor
+endif
+
 .PHONY: common-test-short
 common-test-short: $(GOTEST_DIR)
 	@echo ">> running short tests"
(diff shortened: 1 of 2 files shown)


roidelapluie:
should we call it update go-deps ?

simonpasquier:
@roidelapluie sounds good, PR adjusted

roidelapluie:
That does not remove existing indirects

Makefile.common
```diff
@@ -150,6 +150,18 @@ else
 	$(GO) get $(GOOPTS) -t ./...
 endif
 
+.PHONY: update-go-deps
+update-go-deps:
+	@echo ">> updating Go dependencies"
+	@jq --version 1>/dev/null 2>&1 || { echo "Make sure that 'jq' is in your PATH"; exit 1; }
+	@for m in $$(go list -m -json all | jq -rc 'select(.Indirect != true and .Main != true) | .Path'); do \
```
roidelapluie, line 157:
```suggestion
	@for m in $$(go list -m -f '{{ if and (not .Indirect) (not .Main)}}{{.Path}}{{end}}' all); do \
```

roidelapluie: No need for jq

roidelapluie: -mod=readonly is also needed

simonpasquier:
> That does not remove existing indirects

I don't think it should in the scope of this change.

Makefile.common
simonpasquier: :+1:

Makefile.common
```diff
@@ -150,6 +150,17 @@ else
 	$(GO) get $(GOOPTS) -t ./...
 endif
 
+.PHONY: update-go-deps
+update-go-deps:
+	@echo ">> updating Go dependencies"
+	@for m in $$(go list -mod=readonly -m -f '{{ if and (not .Indirect) (not .Main)}}{{.Path}}{{end}}' all); do \
```
roidelapluie, line 156:
```suggestion
	@for m in $$($(GO) list -mod=readonly -m -f '{{ if and (not .Indirect) (not .Main)}}{{.Path}}{{end}}' all); do \
```

roidelapluie:
:+1:

Issue #2454, opened 2017-02-28
Added alert labels cannot be used in annotation


JorritSalverda:
**What did you do?**

When I add a label in an alert via LABELS I would like to be able to use it immediately in the annotation, however that doesn't seem to be possible. The label does show up in the ALERTS timeline series and can be used in the alertmanager.

```
ALERT HaproxyErrorRateAboveThreshold
  IF sum(rate(haproxy_backend_http_responses_total{code="5xx"}[1m])) by (app,team,kubernetes_namespace) / sum(rate(haproxy_backend_http_responses_total[1m])) by (app,team,kubernetes_namespace) > 0.02
  FOR 5m
  LABELS { severity = "page", kubernetes_cluster = "cluster-name", gcloud_project = "project-name" }
  ANNOTATIONS {
    summary = "Error rate of {{ $labels.app }} is above threshold",
    description = "Error rate of {{ $labels.app }}.{{ $labels.kubernetes_namespace }}.{{ $labels.kubernetes_cluster }} is above threshold of 2% for more than 5 minutes.",
  }
```

**What did you expect to see?**

I would expect `{{ $labels.kubernetes_cluster }}` in the annotation description to be replaced with the value I just set in the LABELS section.

**What did you see instead? Under which circumstances?**

It gets replaced with an empty string.

**Environment**

* System information:

`Linux 4.4.21+ x86_64`

* Prometheus version: 

```
prometheus, version 1.5.2 (branch: master, revision: bd1182d29f462c39544f94cc822830e1c64cf55b)
  build user:       root@1a01c5f68840
  build date:       20170210-16:23:28
  go version:       go1.7.5

```
* Alertmanager version:

```
alertmanager, version 0.5.1 (branch: master, revision: 0ea1cac51e6a620ec09d053f0484b97932b5c902)
  build user:       root@fb407787b8bf
  build date:       20161125-08:14:40
  go version:       go1.7.3

```
* Prometheus configuration file:

prometheus.yml
```
# By default, scrape targets every 15 seconds.
global:
  scrape_interval:     15s

# A scrape configuration for running Prometheus on a Kubernetes cluster.
# This uses separate scrape configs for cluster components (i.e. API server, node)
# and services to allow each to use different authentication configs.
#
# Kubernetes labels will be added as Prometheus labels on metrics via the
# `labelmap` relabeling action.

# Scrape config for API servers.
#
# Kubernetes exposes API servers as endpoints to the default/kubernetes
[...]

PR #7262, opened 2020-05-15
get labels from config for interpolation in annotations

Gnoale:
When an alert fire, you may or may not get labels, depending on the expression used return a metrics or just a value. 

The labels configured inside the alerts definition are effectively passed to the alert manager and can be used to route the alert.

But in the alert template rendering,  the labels comes from the metrics only, which results in poor alerts Summary, Dashboard etc.  

Example :

```
  - alert: TestLabels
    expr: |
      sum(requests_total) > 1
    for: 1m
    labels:
      team: example
      job: example-runner
      instance: dev
      severity: critical
    annotations:
      summary: "job {{ $labels.job }} total requests = {{ $value }}"
      dashboard: "https://example-dashboard/example?var-env={{ $labels.instance }}&var-team={{ $labels.team }}&var-service={{ $labels.job }}"
```

Without the patch, the alert payload is :
```
    {
      "labels": {
        "alertname": "TestLabels",
        "instance": "dev",
        "job": "example-runner",
        "severity": "critical",
        "team": "example"
      },
      "annotations": {
        "dashboard": "https://example-dashboard/example?var-env=&var-team=&var-service=",
        "summary": "job total requests = 1614"
      },
```

With the patch :
```
    {
      "labels": {
        "alertname": "TestLabels",
        "instance": "dev",
        "job": "example-runner",
        "severity": "critical",
        "team": "example"
      },
      "annotations": {
        "dashboard": "https://example-dashboard/example?var-env=dev&var-team=example&var-service=example-runner",
        "summary": "job example-runner total requests = 1614"
      },
```

I ran the tests successfully

### DIFF
--- a/rules/alerting.go
+++ b/rules/alerting.go
@@ -319,7 +319,12 @@ func (r *AlertingRule) Eval(ctx context.Context, ts time.Time, query QueryFunc, 
 		for _, lbl := range smpl.Metric {
 			l[lbl.Name] = lbl.Value
 		}
-
+		// get labels from config if the evaluation did not return any label
+		if len(l) == 0 && len(r.labels) > 1 {
+			for _, lbl := range r.labels {
+				l[lbl.Name] = lbl.Value
+			}
+		}
 		tmplData := template.AlertTemplateData(l, r.externalLabels, smpl.V)
 		// Inject some convenience variables that are easier to remember for users
 		// who are not used to Go's templating system.


brian-brazil:
Thanks for your PR, this has been discussed before such as in #2454 and it is not something we plan on supporting. In this case it looks like you are inappropriately aggregating away labels in your expression, which is easy to resolve.

Gnoale:
Ok thanks, I didn't know about the the aggregation ... (`sum(requests_total) by (job,instance)`)  :+1:

PR #7264, opened 2020-05-17
[metrics] api module: add API response Code counter

liguozhong:
Signed-off-by: fuling <fuling.lgz@alibaba-inc.com>

We tried to use the "RED" method to implement SLO monitoring on Prometheus, and found that the API module had no "Error" indicator
image

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -88,6 +88,13 @@ var (
 		Name:      "remote_read_queries",
 		Help:      "The current number of remote read queries being executed or waiting.",
 	})
+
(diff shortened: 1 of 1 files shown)


brian-brazil:
We should use what client_golang provides for us here, as we do with the existing http metrics.

liguozhong:
> We should use what client_golang provides for us here, as we do with the existing http metrics.

 there is an error indicator. thanks

requestCounter: prometheus.NewCounterVec(
			prometheus.CounterOpts{
				Name: "prometheus_http_requests_total",
				Help: "Counter of HTTP requests.",
			},
			[]string{"handler", "code"},
		),

PR #7271, opened 2020-05-19
queue_manager: Use buffered channel to avoid blocking

gaurav1086:
Signed-off-by: Gaurav Singh <gaurav1086@gmail.com>

The send operation to the channel done will certainly be executed, but the receive operation may not, if the goroutine times out. Hence the goroutine might get stuck and cause a leak. A simple fix is to use a buffered channel so that the goroutine can write to it without having to care if there's a reader or not. This PR converts the channel to a buffered channel.

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -676,7 +676,7 @@ func (t *QueueManager) reshardLoop() {
 func (t *QueueManager) newShards() *shards {
 	s := &shards{
 		qm:   t,
-		done: make(chan struct{}),
+		done: make(chan struct{}, 1),
 	}
 	return s
 }
@@ -722,7 +722,7 @@ func (s *shards) start(n int) {
 	hardShutdownCtx, s.hardShutdown = context.WithCancel(context.Background())
 	s.softShutdown = make(chan struct{})
 	s.running = int32(n)
-	s.done = make(chan struct{})
+	s.done = make(chan struct{}, 1)
 	for i := 0; i < n; i++ {
 		go s.runShard(hardShutdownCtx, i, newQueues[i])
 	}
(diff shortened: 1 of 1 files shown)


gaurav1086:
@csmarchbanks thanks for the comment. On line 750, we have 

select {
	case <-s.done:
		return
	case <-time.After(s.qm.flushDeadline):
		level.Error(s.qm.logger).Log("msg", "Failed to flush all samples on shutdown")
	}
where if the timeout happens, then any goroutine writing s.done will keep hanging unless its a buffered channel, however after timeout, on line 759, we have an explicit

<-s.done

which will block until the corresponding goroutine writes to it. So its not quite a leak as I proposed, but just a blocking call after the timeout. I will change the heading as per this. Thoughts ?

csmarchbanks:
I still do not think this is necessary. There are no goroutines writing to the `done` channel so nothing will be blocked. The only interaction with the `done` channel is `runShard()` closing the channel when finished. Closing a channel does not block if there is no receiver ready ready to get a message.

storage/remote/queue_manager.go
```diff
@@ -756,7 +756,6 @@ func (s *shards) stop() {
 
 	// Force an unclean shutdown.
 	s.hardShutdown()
-	<-s.done
```
csmarchbanks, line 759: Waiting for `s.done` to be closed is necessary. The `stop()` function needs to wait for all shards to shutdown before returning. Otherwise, after resharding it will be possible for newer samples to be sent out before older samples in the same series are flushed. Removing this line breaks the strict ordering guarantee of remote write.

gaurav1086: @csmarchbanks thanks for the pointers. I removed this line since I made this a buffered channel above. I am not experiencing any issue, just came across this as an improvement. I agree that its not a potential bug per se. The go channels, I believe is FIFO in nature so it will guarantee the ordering regardless. However, I agree that since its working as designed, we don't need this PR. Thanks again. Closing this for now.

PR #7272, opened 2020-05-19
tsdb: fix goroutine leak on channel stopC

gaurav1086:
Signed-off-by: Gaurav Singh <gaurav1086@gmail.com>

In the function: func (db *DB) run() {} , select listens on 2 channels with a timeout. The channel: db.compactC is a buffered channel however the other channel db.stopC is not buffered. In case some goroutine is trying to write to it but is timed out, it can cause leak. The PR converts the db.stopC to a buffered channel just like the db.compactC .


gaurav1086:
@csmarchbanks , ah yes, you are right. Thanks for pointing out. Closing this.

PR #7277, opened 2020-05-20
More explicit chunks and  head error handling.

krasi-georgiev:
Signed-off-by: Krasi Georgiev <8903888+krasi-georgiev@users.noreply.github.com>

before this trying to read from a closed head would return
```
unexpected error: write compaction: iterate compaction set: chunk 16777216 not found: not found
```
and with this it will return something like:
```
 unexpected error: write compaction: open chunk reader for block {ULID:2ZBXFNYQTZD1JP2S2ZBXFNYQTZ MinTime:0 MaxTime:49999 Stats:{NumSamples:0 NumSeries:1 NumChunks:0 NumTombstones:0} Compaction:{Level:0 Sources:[] Deletable:false Parents:[] Failed:false} Version:0}: can't read from a closed head
```

Quite easy to mess up in tests and close the head before reading form it in a defer statement. Without this change it is diff to find out the cause from the not found error.

here is a test that triggered this error
https://github.com/thanos-io/thanos/pull/2633/files#diff-37004299aa94ffd8b4ea675ceccb28aaR1118-R1120

### DIFF
--- a/tsdb/chunks/head_chunks.go
+++ b/tsdb/chunks/head_chunks.go
@@ -432,7 +432,6 @@ func (cdm *ChunkDiskMapper) flushBuffer() error {
 }
 
 // Chunk returns a chunk from a given reference.
-// It panics for data corruption errors as these are non recoverable.
 func (cdm *ChunkDiskMapper) Chunk(ref uint64) (chunkenc.Chunk, error) {
 	cdm.readPathMtx.RLock()
 	// We hold this read lock for the entire duration because if the Close()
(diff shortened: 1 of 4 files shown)


tsdb/head.go
```diff
@@ -1420,6 +1425,7 @@ func (h *Head) Close() error {
 	if h.wal != nil {
 		merr.Add(h.wal.Close())
 	}
+	h.closed = true
```
codesome, line 1428: It would be safe to mark this in the beginning so that there is no race in accessing the ones that are already closed (h.chunkDiskMapper and h.wal).

tsdb/head.go
```diff
@@ -1360,10 +1362,13 @@ func (h *Head) indexRange(mint, maxt int64) *headIndexReader {
 
 // Chunks returns a ChunkReader against the block.
 func (h *Head) Chunks() (ChunkReader, error) {
-	return h.chunksRange(math.MinInt64, math.MaxInt64, h.iso.State()), nil
+	return h.chunksRange(math.MinInt64, math.MaxInt64, h.iso.State())
 }
 
-func (h *Head) chunksRange(mint, maxt int64, is *isolationState) *headChunkReader {
+func (h *Head) chunksRange(mint, maxt int64, is *isolationState) (*headChunkReader, error) {
+	if h.closed {
```
codesome, line 1369: How is this exactly helping here? Because IIUC, there can be a race between calling this and `Close()`, we could get past this check while it is being closed simultaneously.

tsdb/head.go
```diff
@@ -2012,9 +2018,6 @@ func (s *memSeries) chunk(id int, chunkDiskMapper *chunks.ChunkDiskMapper) (chun
 	}
 	chk, err := chunkDiskMapper.Chunk(s.mmappedChunks[ix].ref)
 	if err != nil {
-		if err == chunks.ErrChunkDiskMapperClosed {
-			return nil, false
-		}
```
codesome, line 2017: Any reasons for removing this? It would create an unnecessary panic when we know we can get away with it safely (maybe a log with this?) :)

tsdb/head.go
codesome: I see the issue description where it is helping :). Is there any possibility of race as I mentioned? Or is it expected to not be accessed when we are closing?

krasi-georgiev: I don't think there is a race but since this is not a hot path it is just safer to add a mutex so I updated the PR.

tsdb/chunks/head_chunks.go
```diff
@@ -432,7 +432,7 @@ func (cdm *ChunkDiskMapper) flushBuffer() error {
 }
 
 // Chunk returns a chunk from a given reference.
-// Note: The returned chunk will turn invalid after closing ChunkDiskMapper.
+// It panics for data corruption errors as these are non recoverable.
```
codesome, line 435: No longer valid :)

tsdb/head.go
```diff
@@ -1998,31 +2012,33 @@ func (s *memSeries) appendable(t int64, v float64) error {
...
+		if s.headChunk == nil {
+			return nil, false, errors.New("invalid head chunk")
+		}
+		return s.headChunk, false, nil
 	}
 	chk, err := chunkDiskMapper.Chunk(s.mmappedChunks[ix].ref)
 	if err != nil {
-		if err == chunks.ErrChunkDiskMapperClosed {
-			return nil, false
+		if _, ok := err.(*wal.CorruptionErr); ok {
```
codesome, line 2032:
```suggestion
		if _, ok := err.(*chunks.CorruptionErr); ok {
```

And can we move the comment about this panic from head_chunks.go to here?

PR #7285, opened 2020-05-22
Better evaluate impact in memory of range selection

roidelapluie:

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -952,7 +952,7 @@ func (ev *evaluator) rangeEval(f func([]parser.Value, *EvalNodeHelper) Vector, e
 		}
 		enh.out = result[:0] // Reuse result vector.
 
-		ev.currentSamples += len(result)
+		ev.currentSamples = tempNumSamples + len(result)
 		// When we reset currentSamples to tempNumSamples during the next iteration of the loop it also
 		// needs to include the samples from the result here, as they're still in memory.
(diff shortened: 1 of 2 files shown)


promql/engine.go
```diff
@@ -1158,10 +1158,11 @@ func (ev *evaluator) eval(expr parser.Expr) parser.Value {
 				// Only buffer stepRange milliseconds from the second step on.
 				it.ReduceDelta(stepRange)
 			}
+			points = points[:0]
```
brian-brazil, line 1161: It's a bit weird to do this after rather than before, it makes the logic less obvious.

roidelapluie:
Yes I am working on test now and we are somehow still overcounting.

brian-brazil:
This PR only corrects the overcount after an entire series is processed, not on each step.

promql/engine.go
```diff
@@ -1144,6 +1145,7 @@ func (ev *evaluator) eval(expr parser.Expr) parser.Value {
 				mint := maxt - selRange
 				// Evaluate the matrix selector for this series for this step.
 				points = ev.matrixIterSlice(it, mint, maxt, points)
+				ev.currentSamples = tmpEvalPoints
```
brian-brazil, line 1148: Not quite, the points are still in memory.

We probably want to rethink how matrixIterSlice tracking works - maybe make it external to matrixIterSlice?

Issue #6865, opened 2020-02-24
HTTP API: Time range selectors for `/api/v1/labels` and `/api/v1/label/<label_name>/values`


bwplotka:
Hi Prometheus Family :wave: 

## What

I would like to propose extending existing HTTP v1 APIs for fetching labels and label values to accept time range selection:

```
start=<rfc3339 | unix_timestamp>: Start timestamp.
end=<rfc3339 | unix_timestamp>: End timestamp.
```

## Why

Nowadays in Prometheus (TSDB) we don't cache all values and label names anymore like we used to do. With mmap + 1/32 symbols and 1/32 posting offset block loading we have now from 2.15, each call for labels or label values will allocate extra memory (+CPU cycles) every time those APIs are invoked. This is fine for small retention, but with years of data, this might be a major bottleneck.

The main problem is that we usually don't need label names or values for EVERYTHING. The typical use case is in dashboards, where we want to know what are metric (e.g `__name__`) values available for certain e.g 1h time range we want to query for! So fetching all names and values for whole available retention is highly unnecessary. Its also both inefficient and incorrect (**this label name might not exist in the certain 1h time range we query for**). Because of that API limitation, even though e.g. Grafana, might know what time range we are interested in, it can't pass this info to Prometheus.

It is even more problematic for LTS solutions like Thanos and Cortex. In fact, Cortex already ignores correctness and returns just ~2h of labels name / values. Thanos is now correct, but with millions of series stored for years, it's super silly to fetch all label names and values for every request. Yes, we can cache things, but it would be awesome if Grafana, Prometheus UI or any client could just select the time range it is interested in.

## How

The proposed solution allows time slicing, but to be fully correct on the Prometheus side (and same will be for Thanos), we would need to perform literally `select` operation to know which labels are within desired time range. We can optimize for the case where time range fully overlaps with block though:

### /api/v1/labels?start=T1&end=T2

* Narrow down to block-level only if we time range fully overlaps with a block time range.
* Perform series call to fetch label names within block otherwise.

### /api/v1/label/<label_name>/values?start=T1&end=T2

* Narrow down to block-level only if we time range fully overlaps with a block time range.
* Perform series call to fetch label names within block otherwise.

Reference: https://github.com/thanos-io/thanos/issues/1811

Thoughts? (: 

@pracucci @brancz @brian-brazil @davkal @tomwilkie ?

bwplotka:
[...]

PR #7288, opened 2020-05-24
Added time range parameters to labelNames API

JessicaGreben:
This PR is for issue: https://github.com/prometheus/prometheus/issues/6865

This PR adds the ability to pass a `start` and `end` time range parameter to the `api/v1/label` and `/api/v1/label/<label_name>/values` route.  

- for simplicity, this change does not add any matchers functionality
- this change does not use the `series` api ([ref](https://github.com/prometheus/prometheus/issues/6865#issuecomment-590714724))
- this "keeps it simple and allow for false positives when the time doesn't exactly align with a block" ([ref](https://github.com/prometheus/prometheus/issues/6865#issuecomment-590348567))
- the start/end parameters are optional and API backwards compatibility is maintained

Other implementation details:
- It turns out the head index reader was ignoring the timestamps passed from the Querier mint(start)/maxt(end) ([ref](https://github.com/prometheus/prometheus/blob/4658ce60d1f21529a7ffe35ab0a30baeefbe1a04/tsdb/head.go#L1554-L1567)). This PR adds logic so that the head index reader evaluates these times now.  If the start is after the last item in the head or if the end is less than the first item in the head, no labels will be returned. This means we are saying the official "end" of the head block - is the maximum-seen sample timestamp for that in the head.

Open questions:
- Should we change `maxt >= db.head.MinTime()` to ` if maxt >= db.head.MinTime() && mint <= db.head.MaxTime()` in `tsdb/db.go/db.Querier` ([ref](https://github.com/prometheus/prometheus/blob/4658ce60d1f21529a7ffe35ab0a30baeefbe1a04/tsdb/db.go#L1305-L1311))?  Generally there is a hierarchical filtering structure, with the mint/maxt being passed all the way down... in db.go if we can already detect the entire head doesn't overlap, I don't think we need to append the head there. And then the mint/maxt within the actual blocks/head should could mainly be used for more finer-grained checks (where its possible - we decided against that for LabelNames() for now and only evaluate at the full block)
- Since the head index reader was ignoring the timestamps passed from the Querier mint(start)/maxt(end) ([ref](https://github.com/prometheus/prometheus/blob/4658ce60d1f21529a7ffe35ab0a30baeefbe1a04/tsdb/head.go#L1554-L1567)),  it would just be good to apply the same check systematically in other places where it is applicable (other metadata methods, etc.). Should we make that change in this PR or save those fixes for another PR?

### DIFF
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -268,6 +268,12 @@ GET /api/v1/labels
 POST /api/v1/labels
 ```
 
+URL query parameters:
+
+- `start=<rfc3339 | unix_timestamp>`: Start timestamp. Optional.
+- `end=<rfc3339 | unix_timestamp>`: End timestamp. Optional.
+
+
 The `data` section of the JSON response is a list of string label names.
 
 Here is an example.
@@ -309,6 +315,12 @@ The following endpoint returns a list of label values for a provided label name:
 ```
 GET /api/v1/label/<label_name>/values
 ```
+
+URL query parameters:
+
+- `start=<rfc3339 | unix_timestamp>`: Start timestamp. Optional.
+- `end=<rfc3339 | unix_timestamp>`: End timestamp. Optional.
+
 
 The `data` section of the JSON response is a list of string label values.
 
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -1555,6 +1555,11 @@ func (h *headIndexReader) LabelValues(name string) ([]string, error) {
 func (h *headIndexReader) LabelNames() ([]string, error) {
 	h.head.symMtx.RLock()
 	defer h.head.symMtx.RUnlock()
+
+	if h.maxt < h.head.MinTime() || h.mint > h.head.MaxTime() {
+		return []string{}, nil
+	}
+
 	labelNames := make([]string, 0, len(h.head.values))
 	for name := range h.head.values {
 		if name == "" {
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -1539,7 +1539,8 @@ func (h *headIndexReader) Symbols() index.StringIter {
 	return index.NewStringListIter(res)
 }
 
-// LabelValues returns the possible label values
+// LabelValues returns label values present in the head for the
+// specific label name that are within the time range mint to maxt.
 func (h *headIndexReader) LabelValues(name string) ([]string, error) {
 	h.head.symMtx.RLock()
 	if h.maxt < h.head.MinTime() || h.mint > h.head.MaxTime() {
@@ -1555,7 +1556,8 @@ func (h *headIndexReader) LabelValues(name string) ([]string, error) {
 	return sl, nil
 }
 
-// LabelNames returns all the unique label names present in the head.
+// LabelNames returns all the unique label names present in the head
+// that are within the time range mint to maxt.
 func (h *headIndexReader) LabelNames() ([]string, error) {
 	h.head.symMtx.RLock()
 	defer h.head.symMtx.RUnlock()
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -1543,7 +1543,9 @@ func (h *headIndexReader) Symbols() index.StringIter {
 // specific label name that are within the time range mint to maxt.
 func (h *headIndexReader) LabelValues(name string) ([]string, error) {
 	h.head.symMtx.RLock()
+
 	if h.maxt < h.head.MinTime() || h.mint > h.head.MaxTime() {
+		h.head.symMtx.RUnlock()
 		return []string{}, nil
 	}
 
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -1542,6 +1542,10 @@ func (h *headIndexReader) Symbols() index.StringIter {
 // LabelValues returns the possible label values
 func (h *headIndexReader) LabelValues(name string) ([]string, error) {
 	h.head.symMtx.RLock()
+	if h.maxt < h.head.MinTime() || h.mint > h.head.MaxTime() {
+		return []string{}, nil
+	}
+
 	sl := make([]string, 0, len(h.head.values[name]))
 	for s := range h.head.values[name] {
 		sl = append(sl, s)
--- a/tsdb/head_test.go
+++ b/tsdb/head_test.go
@@ -1813,9 +1813,9 @@ func newTestHead(t testing.TB, chunkRange int64, compressWAL bool) (*Head, *wal.
 
 func TestHeadLabelNamesValuesWithMinMaxRange(t *testing.T) {
 	head, _, closer := newTestHead(t, 1000, false)
+	defer closer()
 	defer func() {
-		closer()
-		testutil.Ok(t, head.Close())
+		testutil.Ok(t, hb.Close())
 	}()
 
 	const (
--- a/tsdb/head_test.go
+++ b/tsdb/head_test.go
@@ -1813,8 +1813,8 @@ func newTestHead(t testing.TB, chunkRange int64, compressWAL bool) (*Head, *wal.
 
 func TestHeadLabelNamesValuesWithMinMaxRange(t *testing.T) {
 	head, _, closer := newTestHead(t, 1000, false)
-	defer closer()
 	defer func() {
+		closer()
 		testutil.Ok(t, head.Close())
 	}()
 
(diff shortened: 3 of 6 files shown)


web/api/v1/api.go
```diff
@@ -482,7 +482,18 @@ func returnAPIError(err error) *apiError {
 }
 
 func (api *API) labelNames(r *http.Request) apiFuncResult {
-	q, err := api.Queryable.Querier(r.Context(), math.MinInt64, math.MaxInt64)
+	start, err := parseTimeParam(r, "start", minTime)
+	if err != nil {
+		err = errors.Wrapf(err, "invalid parameter 'start'")
```
bwplotka, line 487: No need to use WrapF, as there is no formatting needed.

web/api/v1/api.go
```diff
@@ -482,7 +482,18 @@ func returnAPIError(err error) *apiError {
 }
 
 func (api *API) labelNames(r *http.Request) apiFuncResult {
-	q, err := api.Queryable.Querier(r.Context(), math.MinInt64, math.MaxInt64)
+	start, err := parseTimeParam(r, "start", minTime)
```
bwplotka, line 485: I did not look deeper but we need to make sure to not break compatibility in the same major version of Prometheus. We can only add things, we cannot suddenly fail all requests without min and max time for all tens of thousands users (((: 

Let's make sure we assume no start and end params works as previously. A test case for this would be amazing as well. 🤗

bwplotka:
Ah be carefoul on one thing @JessicaGreben The querier will never give you exact labels name and values for given time up to millisecond precision. The problem is that we don't index that. We only index series for time precision within block. The means we either have to make those endpoints more expensive or return labels in just per block precision. (: I this we agreed on latter initially with @brian-brazil 
 
This is why you might get confusing test failures (:

web/api/v1/api.go
JessicaGreben: Start and end are optional parameters so that compatibility should be maintained. If a start and end are not provided, the code defaults to how the code was operating previously. 

I believe there are a few tests that ensure this:
- https://github.com/prometheus/prometheus/pull/7288/files#diff-c32caf482fb76fa695ceffe3455c6d9aR1479-R1480?
- https://github.com/prometheus/prometheus/blob/957a57db5bb478d5723fe93d2678d674bce87e20/web/api/v1/api_test.go#L404

Let me know if you think I should add any other tests or if this is sufficient.

juliusv:
> Should we change `maxt >= db.head.MinTime()` to ` if maxt >= db.head.MinTime() && mint <= db.head.MaxTime()` in `tsdb/db.go/db.Querier` ([ref](https://github.com/prometheus/prometheus/blob/4658ce60d1f21529a7ffe35ab0a30baeefbe1a04/tsdb/db.go#L1305-L1311))?  Generally there is a hierarchical filtering structure, with the mint/maxt being passed all the way down... in db.go if we can already detect the entire head doesn't overlap, I don't think we need to append the head there. And then the mint/maxt within the actual blocks/head should could mainly be used for more finer-grained checks (where its possible - we decided against that for LabelNames() for now and only evaluate at the full block)

Yes, I think this condition should be added to your reference in db.go (skip the head if it's not overlapping). If we do that, I'm not sure we still even need the check in head.go, if it's not finer-grained than the one that we do on the higher level. OTOH, it can't hurt, and maybe the head index reader is also used from other places, so maybe it makes sense to keep the check there too. Someone who is deeper into the TSDB code and its implicit assumptions / invariants should comment on that.

brian-brazil:
The docs should also be updated in this PR.

JessicaGreben:
@brian-brazil I added some docs, would you mind letting me know if its sufficient or if I missed something?

tsdb/querier.go
```diff
@@ -64,7 +64,7 @@ func (q *querier) LabelNames() ([]string, storage.Warnings, error) {
 
 func (q *querier) lvals(qs []storage.Querier, n string) ([]string, storage.Warnings, error) {
 	if len(qs) == 0 {
-		return nil, nil, nil
+		return []string{}, nil, nil
```
JessicaGreben, line 67: I'm not sure if the changes in this method `lvals` break backwards compatibility.  

Before this change if this code block was executed, the api would return `[]string(nil)`, but that was inconsistent with what most nil cases would return as `[]string{}`.  Is this an ok change to make? Or should I keep it as `[]string(nil)` for compatibility?

web/api/v1/api_test.go
```diff
@@ -1474,11 +1474,216 @@ func testEndpoints(t *testing.T, api *API, tr *testTargetRetriever, testLabelAPI
 				},
 				errType: errorBadData,
 			},
+			// Start and end before LabelValues starts.
```
JessicaGreben, line 1477: @brian-brazil  @juliusv 
In all these tests added here, we are only testing series that exist for the entire range via this 
```
load 1m
		test_metric1{foo="bar"} 0+100x100
		test_metric1{foo="boo"} 1+0x100
		test_metric2{foo="boo"} 1+0x100
```

Is it important to test that a record only exists for a portion of the time range? If so, should I add support for that in this PR or do it in another?

tsdb/head.go
```diff
@@ -1542,6 +1542,10 @@ func (h *headIndexReader) Symbols() index.StringIter {
 // LabelValues returns the possible label values
 func (h *headIndexReader) LabelValues(name string) ([]string, error) {
 	h.head.symMtx.RLock()
+	if h.maxt < h.head.MinTime() || h.mint > h.head.MaxTime() {
```
cstyan, line 1545: We should probably add a test for this behaviour here and for `LabelNames`. I'd put it in `head_test.go` now.

web/api/v1/api.go
bwplotka: No it's perfect, looks like this is done then (:

tsdb/head.go
JessicaGreben: added the test in this commit: https://github.com/prometheus/prometheus/pull/7288/commits/9c57a70e1adddbee5760fa82beface259be8122a, let me know what you think.  

Also @codesome, @krasi-georgiev would you mind reviewing this tsdb change?

krasi-georgiev: yes this change looks good.

tsdb/head_test.go
```diff
@@ -1810,3 +1810,105 @@ func newTestHead(t testing.TB, chunkRange int64, compressWAL bool) (*Head, *wal.
...
+			testutil.Ok(t, err)
+			testutil.Equals(t, expectedLabelValues[i], actualLabelValue)
+		}
+	}
+	// maxt is less than head min and mint is greater than head max
+	// expect no labels returned since the mint and maxt are outside
+	// of head timestamps
+	{
+		mint := head.MaxTime() + 10
+		maxt := head.MinTime() + 10
```
nevill, line 1887: based on the comments, should it be `- 10` here?

JessicaGreben: oops, good catch @nevill. thank you. let me fix.

tsdb/head_test.go
```diff
@@ -1810,3 +1810,105 @@ func newTestHead(t testing.TB, chunkRange int64, compressWAL bool) (*Head, *wal.
 		testutil.Ok(t, os.RemoveAll(dir))
 	}
 }
+
+func TestHeadLabelNamesValuesWithMinMaxRange(t *testing.T) {
+	head, _, closer := newTestHead(t, 1000, false)
+	defer closer()
```
bwplotka, line 1816: You can invoke closer below just to have it concise (:

tsdb/head_test.go
```diff
@@ -1810,3 +1810,105 @@ func newTestHead(t testing.TB, chunkRange int64, compressWAL bool) (*Head, *wal.
...
+		expectedLabelValues = []string{"d", "e", "f"}
+	)
+
+	app := head.Appender()
+	for i, name := range expectedLabelNames {
+		_, err := app.Add(labels.Labels{{Name: name, Value: expectedLabelValues[i]}}, seriesTimestamps[i], 0)
+		testutil.Ok(t, err)
+	}
+	err := app.Commit()
+	testutil.Ok(t, err)
```
bwplotka, line 1841: We can inline the above to: 
```suggestion
	testutil.Ok(t, app.Commit())
```

bwplotka: (Goal: Bit less code to read without complexity, so less cognitive load)

tsdb/head_test.go
```diff
@@ -1810,3 +1810,105 @@ func newTestHead(t testing.TB, chunkRange int64, compressWAL bool) (*Head, *wal.
...
+
+	testutil.Assert(t,
+		head.MinTime() == firstSeriesTimestamp,
```
bwplotka, line 1844: For this we always use the `Equals`. The advantage is that it prints what is expected and what provided

tsdb/head_test.go
```diff
@@ -1810,3 +1810,105 @@ func newTestHead(t testing.TB, chunkRange int64, compressWAL bool) (*Head, *wal.
...
+		"the min timestamp of the head should be equal to the timestamp of first series added",
+	)
+	testutil.Assert(t,
```
bwplotka, line 1847: ditto

tsdb/head_test.go
```diff
@@ -1810,3 +1810,105 @@ func newTestHead(t testing.TB, chunkRange int64, compressWAL bool) (*Head, *wal.
...
+		head.MaxTime() == lastSeriesTimestamp,
+		"the max timestamp of the head should be equal to the timestamp of last series added",
+	)
+
+	// maxt is less than head min and mint is less than head max
```
bwplotka, line 1852: I would be nitty picky here.. I think this deserves table test (: Essentially there are many repetitions here and table test would fit here perfectly. What do you think? (: 

I always have this template for that:
```
for _, tcase := range []struct{
  name string 
  // other things
}{
    { 
      // cases
    },
}{
  t.Run(tcase.name, func(t *testing.T) {
            // test
  })
}
```

web/api/v1/api.go
```diff
@@ -482,7 +482,18 @@ func returnAPIError(err error) *apiError {
...
+	if err != nil {
+		err = errors.Wrap(err, "invalid parameter 'start'")
+		return apiFuncResult{nil, &apiError{errorBadData, err}, nil, nil}
+	}
+	end, err := parseTimeParam(r, "end", maxTime)
+	if err != nil {
+		err = errors.Wrap(err, "invalid parameter 'end'")
```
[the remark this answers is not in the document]
bwplotka: This relates to other changes in this file as well (:

tsdb/head.go
cstyan: Small nit; should be consistent about whitespace before/after the if statement. Here you don't have a newline before but do after, and in `LabelNames` you have newline before and after. I would go with newline before and after personally.

tsdb/head.go
```diff
@@ -1539,9 +1539,14 @@ func (h *headIndexReader) Symbols() index.StringIter {
 	return index.NewStringListIter(res)
 }
 
-// LabelValues returns the possible label values
+// LabelValues returns label values present in the head for the
+// specific label name that are within the time range mint to maxt.
 func (h *headIndexReader) LabelValues(name string) ([]string, error) {
 	h.head.symMtx.RLock()
+	if h.maxt < h.head.MinTime() || h.mint > h.head.MaxTime() {
+		return []string{}, nil
```
codesome, line 1547: `h.head.symMtx.RUnlock()` before returning :) Not in a `defer` above because we don't want `sort.Strings(sl)` inside the lock.

Optional and unrelated to this PR: I see that `LabelNames()` has the unlock in the defer, can you change it to be similar to this method too?

tsdb/head_test.go
codesome: +1, and the test cases for empty returns are a bit weird. We could assume mint <= maxt and just test these 4 cases

```
1
      |------|
|---|

2
      |------|
    |---|

3
      |------|
           |---|

4
      |------|
               |---|
```

tsdb/head_test.go
```diff
@@ -1810,3 +1810,105 @@ func newTestHead(t testing.TB, chunkRange int64, compressWAL bool) (*Head, *wal.
...
+		mint := head.MaxTime() - 10
+		maxt := head.MinTime() - 10
+		headIdxReader := head.indexRange(mint, maxt)
+		actualLabelNames, err := headIdxReader.LabelNames()
+		testutil.Ok(t, err)
+		testutil.Equals(t, []string{}, actualLabelNames)
+		for i, name := range actualLabelNames {
+			actualLabelValue, err := headIdxReader.LabelValues(name)
+			testutil.Ok(t, err)
+			testutil.Equals(t, expectedLabelValues[i], actualLabelValue)
```
codesome, line 1864: Shouldn't `actualLabelValue` match `[]string{}`? Am I missing something?

tsdb/head_test.go
JessicaGreben: @codesome I picked those test cases since they execute all 4 cases of the `if` block i added here https://github.com/prometheus/prometheus/pull/7288/files#diff-6fb13507bfadf9819fea5bda61f599e6R1565

tsdb/head_test.go
JessicaGreben: because `actualLabelNames` returns []string{}, this range block is not executed

tsdb/head_test.go
bwplotka: Sorry, I should more explicit. I was suggesting:

```
defer func() {
		closer()
		testutil.Ok(t, head.Close())
	}()
```

tsdb/head_test.go
```diff
@@ -1810,3 +1810,60 @@ func newTestHead(t testing.TB, chunkRange int64, compressWAL bool) (*Head, *wal.
...
+	app := head.Appender()
+	for i, name := range expectedLabelNames {
+		_, err := app.Add(labels.Labels{{Name: name, Value: expectedLabelValues[i]}}, seriesTimestamps[i], 0)
+		testutil.Ok(t, err)
+	}
+	testutil.Ok(t, app.Commit())
+	testutil.Equals(t, head.MinTime(), firstSeriesTimestamp)
+	testutil.Equals(t, head.MaxTime(), lastSeriesTimestamp)
+
+	var testCases = []struct {
```
bwplotka, line 1844: Amazing (: Much cleaner (: 

You make it even more compact by using anonymous struct:

```
for _, tt := range  []struct {
....
```
Not a blocker :hugs:

JessicaGreben: i like it separate better, easier to read imo

tsdb/head_test.go
```diff
@@ -1810,3 +1810,60 @@ func newTestHead(t testing.TB, chunkRange int64, compressWAL bool) (*Head, *wal.
...
+		t.Run(tt.name, func(t *testing.T) {
+			headIdxReader := head.indexRange(tt.mint, tt.maxt)
+			actualLabelNames, err := headIdxReader.LabelNames()
+			testutil.Ok(t, err)
+			testutil.Equals(t, tt.expected, actualLabelNames)
+			for i, name := range actualLabelNames {
+				actualLabelValue, err := headIdxReader.LabelValues(name)
+				testutil.Ok(t, err)
+				testutil.Equals(t, []string{expectedLabelValues[i]}, actualLabelValue)
+			}
```
codesome, line 1866: This does not actually test `LabelValues()` when the time ranges are outside, we should be using `range expectedLabelNames` instead of `range actualLabelNames` and expect an empty slice when time range is outside.

bwplotka:
Can we have the same on label values? =D

docs/querying/api.md
```diff
@@ -268,6 +268,12 @@ GET /api/v1/labels
 POST /api/v1/labels
 ```
 
+URL query parameters:
+
+- `start=<rfc3339 | unix_timestamp>`: Start timestamp. Optional.
```
bwplotka, line 273: I forgot about would be nice to mention that labels can be OUTSIDE of given ranges (:

bwplotka: As this is just a hint

PR #7294, opened 2020-05-26
Pass reference to checkAddError so appendErrors is updated.

brian-brazil:
This was preventing the warnings from being logged.


brian-brazil:
I considered that, but we don't generally test log messages.

bwplotka:
It's rather the append Error stats being propagated to caller

brian-brazil:
That makes sense, done.

Issue #7284, opened 2020-05-22
PromQL tests requires > 16GB of storage


roidelapluie:
**What did you do?**

When I run prometheus tests, it fills my /tmp disk: /tmp/test_storage632520516/ is 16Gb

~/go/src/github.com/prometheus/prometheus/promql $ go test                                                                     
--- FAIL: TestSubquerySelector (1.25s)                               
    test.go:536: closing test storage: sync /tmp/test_storage396011703/chunks_head/000029: file already closed

```
$ df -hT /tmp
Filesystem     Type   Size  Used Avail Use% Mounted on
tmpfs          tmpfs   16G   16G  116M 100% /tmp

$ uname -a
Linux oxygen 5.6.8-roidelapluie #1 SMP Thu Apr 30 23:23:52 CEST 2020 x86_64 GNU/Linux
```

roidelapluie:
this is because I am using Linux tmpfs as /tmp, and therefore tests fail.

roidelapluie:
So the head chunk PR will probably make prometheus fail for tmpfs users.

roidelapluie:
```
~/go/src/github.com/prometheus/prometheus/promql $ go test
--- FAIL: TestSubquerySelector (1.25s)
test.go:536: closing test storage: sync /tmp/test_storage396011703/chunks_head/000029: file already closed
```

What is NOT seen here is a panic that is somehow not showing up in here:

https://github.com/prometheus/prometheus/blob/master/tsdb/head.go#L1980

In my case:

no space left on device


So my 16GB /tmp is not enough to run promql engine tests anymore..

Issue #7305, opened 2020-05-28
M-mapping of Head chunks taking lots of disk space on first start


codesome:
**What did you do?**

Ran the TSDB as part of Cortex which includes the [mmap of head chunks](https://github.com/prometheus/prometheus/pull/6679). 

**What did you expect to see?**

After/during the replay of WAL, TSDB creating the m-map chunks from the WAL where the chunks' size would be ideally less than the WAL size.

**What did you see instead? Under which circumstances?**

During replay, m-mapped chunks take a lot of space and crash after spitting "disk out of space" error. The m-mapped chunks had crossed 10x the size of WAL when it crashed. 

* Prometheus version:

Prometheus at https://github.com/prometheus/prometheus/commit/f4dd45609a05

codesome:
I am investigating this right now, but any inputs would help too

brian-brazil:
We've also seen a big increase in how much disk space the unittests take, as in over 10G.

codesome:
Oh yes, nice pointer, now I know where to dig from.

codesome:
I can see the reason why unit tests would occupy so much disk: The mmap files are allocated 512MB when they are created. So, multiple tests running in parallel can quickly take up multiple of 0.5GB files. That can potentially explain it and for tests we can try and reduce it. Still investigating the issue from the description if replay cuts a lot of files unnecessarily.

roidelapluie:
#7284

manutenfruits:
Does anyone know what version introduced this bug? Some people are affected and it'll be helpful to know what version to downgrade to until it's fixed.

roidelapluie:
On 28 May 06:23, Manuel Díez wrote:
> Does anyone know what version introduced this bug? Some people are affected and it'll be helpful to know what version to downgrade to until it's fixed.
> 

It has not been released. It is only on master.

> -- 
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/prometheus/prometheus/issues/7305#issuecomment-635347226

-- 
 (o-    Julien Pivotto
 //\    Open-Source Consultant
 V_/_   Inuits - https://www.inuits.eu

codesome:
Closed by https://github.com/prometheus/prometheus/pull/7306

PR #7306, opened 2020-05-28
Preallocate m-map file only for Windows

codesome:
Closes https://github.com/prometheus/prometheus/issues/7284 and https://github.com/prometheus/prometheus/issues/7305

This PR makes pre-allocations of m-map file only necessary for Windows, as it has issues m-mapping an empty file with arbitrary m-map size.

For #7284: Parallel allocations of lots of 512MB files caused the disk to fill up. Tests are now able to run in my local machine which used to run out of space. Can you test this @roidelapluie?

For #7305: it's the same reason, but additionally, if the WAL contains a lot of not-in-sync samples, it causes a lot of unnecessary new m-map files because of hitting the 40m time cap for a file (consider 4 chunks with their maxt appearing in order 5m 50m 5m 50m, that would cause 2 files for just 2 chunks) and hence fill up the disk.

Sorry Windows users.

### DIFF
--- a/tsdb/chunks/head_chunks_other.go
+++ b/tsdb/chunks/head_chunks_other.go
@@ -1,0 +1,22 @@
+// Copyright 2020 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build !windows
+
+package chunks
+
+var (
+	// HeadChunkFilePreallocationSize is the size to which the m-map file should be preallocated when a new file is cut.
+	// Windows needs pre-allocations while the other OS does not.
+	HeadChunkFilePreallocationSize int64
+)
--- a/tsdb/chunks/head_chunks.go
+++ b/tsdb/chunks/head_chunks.go
@@ -188,7 +188,7 @@ func (cdm *ChunkDiskMapper) openMMapFiles() (returnErr error) {
 	for seq, fn := range files {
 		f, err := fileutil.OpenMmapFile(fn)
 		if err != nil {
-			return errors.Wrap(err, "mmap files")
+			return errors.Wrapf(err, "mmap files, file: %s", fn)
 		}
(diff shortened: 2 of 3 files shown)


tsdb/chunks/head_chunks_other.go
```diff
@@ -0,0 +1,22 @@
+// Copyright 2020 The Prometheus Authors
```
krasi-georgiev, line 1: do you even need the `_other` suffix? Should it default to all OSes without this suffix?

codesome: without this suffix it means the original head_chunks.go file. In that case, do we set it to 0 as default and an `init()` method in windows file?

codesome: > Should it default to all OSes without this suffix?

With `// +build !windows` in that file, it will be all non windows OS

roidelapluie:
```
$ cat ./vendor/golang.org/x/net/ipv4/control_unix.go                                 
// Copyright 2012 The Go Authors. All rights reserved.               
// Use of this source code is governed by a BSD-style                
// license that can be found in the LICENSE file.                    
                                                                                                                                           
// +build aix darwin dragonfly freebsd linux netbsd openbsd solaris      
```
Should we use _unix.go?

codesome: I think it's just easier to say `!windows` instead of listing everything.

codesome: I followed the naming from here https://github.com/prometheus/prometheus/tree/master/tsdb/fileutil
where we have some OS mentioned and `_other` for all others

roidelapluie: We also have a case with _unix.go in that directory

codesome: As the (known) issue is only with windows, I would prefer to explicitly have windows and !windows. Is there any gain mentioning all explicitly?

codesome: Another way could be get rid of `_others`, have the variable in main file, and have an `init()` function in `_windows`.

roidelapluie: No gain, no. I was just pointing out how upstream go does it but at the end it is your call.

krasi-georgiev:
how about adding some test?

bwplotka:
Awesome! Thanks for quick reaction.

Is this candidate for 2.18.2? Also if yes, then we might want to revert/commit to `release-2.18` branch then.

codesome:
M-mapping is not there in 2.18.x, it's just in master

bwplotka:
Good point I checked wrongly, thanks!

PR #7308, opened 2020-05-28
tsdb: improve logs when encountering corruption

simonpasquier:
I've spotted a few places where logs and errors could be improved while intentionally breaking my local tsdb...

cc @codesome, @krasi-georgiev

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -441,10 +441,14 @@ func (db *DBReadOnly) Blocks() ([]BlockReader, error) {
 	if len(corrupted) > 0 {
 		for _, b := range loadable {
 			if err := b.Close(); err != nil {
-				level.Warn(db.logger).Log("msg", "Closing a block", err)
+				level.Warn(db.logger).Log("msg", "Closing block failed", "err", err, "block", b)
 			}
 		}
-		return nil, errors.Errorf("unexpected corrupted block:%v", corrupted)
+		var merr tsdb_errors.MultiError
+		for ulid, err := range corrupted {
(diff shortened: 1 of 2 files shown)


tsdb/db.go
```diff
@@ -880,7 +884,11 @@ func (db *DB) reload() (err error) {
 				block.Close()
 			}
 		}
-		return fmt.Errorf("unexpected corrupted block:%v", corrupted)
+		var merr tsdb_errors.MultiError
+		for ulid, err := range corrupted {
+			merr.Add(errors.Wrap(err, ulid.String()))
+		}
+		return merr.Err()
```
bwplotka, line 891: I think we lost wrapping here.. should we had errors.Wrap here as well?

tsdb/db.go
```diff
@@ -441,10 +441,14 @@ func (db *DBReadOnly) Blocks() ([]BlockReader, error) {
...
-				level.Warn(db.logger).Log("msg", "Closing a block", err)
+				level.Warn(db.logger).Log("msg", "Closing block failed", "err", err, "block", b)
 			}
 		}
-		return nil, errors.Errorf("unexpected corrupted block:%v", corrupted)
+		var merr tsdb_errors.MultiError
+		for ulid, err := range corrupted {
+			merr.Add(errors.Wrap(err, ulid.String()))
+		}
+		return nil, merr.Err()
```
bwplotka, line 451: ditto

simonpasquier: done

tsdb/db.go
simonpasquier: done

PR #7313, opened 2020-05-29
Added fuzzer for promql package

AdamKorcz:
This PR adds a fuzzer for the Exec function in the promql package which according to my analysis wasn't covered by the existing fuzzers.

The ultimate target is `_ = vectorQuery.Exec(ctx)` on line 154.
`vectorQuery` holds the random data by means of a `NewInstantQuery` with the random data.

Everything else added is just necessary to make that happen. The comments should explain that.

### DIFF
--- a/promql/fuzz.go
+++ b/promql/fuzz.go
@@ -17,10 +17,15 @@
 package promql
 
 import (
+	"context"
+	"errors"
 	"io"
+	"time"
 
+	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/pkg/textparse"
 	"github.com/prometheus/prometheus/promql/parser"
+	"github.com/prometheus/prometheus/storage"
 )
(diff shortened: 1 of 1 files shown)


brian-brazil:
I don't think this is actually fuzzing anything new. This code will first parse the expression, which we already fuzz. The few cases that don't fail due to parse failure will fail with a storage error, before any real PromQL execution occurs.

AdamKorcz:
I found that this fuzzer added coverage to engine.go by going into `execEvalStmt()` on line 493 and from there into `Eval()` on line 542.
This is based on analysis on the grounds of a corpus built up from 4-8 hours of running the existing fuzzers.

brian-brazil:
That code doesn't do anything with the query string though, that's mainly the time stuff.

AdamKorcz:
Fair enough. I am closing then.

PR #7316, opened 2020-05-29
Fix consul_sd_config to follow types convention

roidelapluie:

### DIFF
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -359,9 +359,9 @@ services:
 tags:
   [ - <string> ]
 
-# Node metadata used to filter nodes for a given service.
+# Node metadata key/value pairs to filter nodes for a given service.
 [ node_meta:
-  [ <name>: <value> ... ] ]
+  [ <string>: <string> ... ] ]
 
 # The string by which Consul tags are joined into the tag label.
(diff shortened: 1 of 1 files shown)


docs/configuration/configuration.md
```diff
@@ -359,9 +359,9 @@ services:
 tags:
   [ - <string> ]
 
-# Node metadata used to filter nodes for a given service.
+# Node metadata pairs used to filter nodes for a given service.
```
brian-brazil, line 362: "names and values" maybe?

roidelapluie: Here is what consul says:

```
Metadata to filter nodes with the given key=value pairs. This flag may be specified multiple times to filter on multiple sources of metadata.
```

Thet do not speak about names but keys/values.

roidelapluie: Should I just put

```
Metadata to filter nodes with the given key=value pairs for a given service
```

as description?

brian-brazil: Okay for that bit, but the grammar feels off. How about:

Node metadata key/value pairs to filter nodes for a given service

roidelapluie: fixed

brian-brazil:
:+1:

PR #7318, opened 2020-05-31
return NaN when buckets have 0 observations

### LINKED ISSUE #6897: Histogram quantile impl has erratic behavior when distribution contains no samples
**What did you do?**
1. Run histogram quantile on any distribution where the the total count of observations across all buckets is 0.
2. This came up for me in a use case where I am doing quantile on a bucketed throughput metric and throughput can die off for an interval. In my particular use case I'm looking at signed estimate error for an estimation system using a histogram where the le tags have a large negative to positive range (for underestimating -> overestimating).
3. Results in specifying the lowest negative bucket as the value of all quantile inputs for the interval with no observations.

Open question - I think I would expect NaN if there are no samples to interpolate against in any buckets. I could be convinced 0 but without domain knowledge of the histogram this seems off. 

I saw in this case -8192 - my largest negative LE (I believe caused by https://github.com/prometheus/prometheus/blob/41151ca8dc069448515f48893b8631b9a3ad8df8/promql/quantile.go#L91)

The rank variable is 0 since there are no samples - then the go binary search finds the min bucket.

**Environment**

* System information:

	insert output of `uname -srm` here

* Prometheus version:
Don't have access to box - but from UI:

Version: 2.7.1
Revision: 62e591f928ddf6b3468308b7ac1de1c63aa7fcf3
Branch: HEAD
BuildDate: 20190131-11:16:59
GoVersion: go1.11.5

JBerny:
Hi @brian-brazil, I tried to fix #6897 by returning NaN in case the buckets have 0 observations.
Could you please review it?

### DIFF
--- a/docs/querying/functions.md
+++ b/docs/querying/functions.md
@@ -182,8 +182,8 @@ is assumed to be 0 if the upper bound of that bucket is greater than
 bucket. Otherwise, the upper bound of the lowest bucket is returned
 for quantiles located in the lowest bucket.
 
-If `b` contains fewer than two buckets, `NaN` is returned. For φ < 0, `-Inf` is
-returned. For φ > 1, `+Inf` is returned.
+If `b` has 0 observations, `NaN` is returned. If `b` contains fewer than two buckets,
+`NaN` is returned. For φ < 0, `-Inf` is returned. For φ > 1, `+Inf` is returned.
 
 ## `holt_winters()`
 
--- a/promql/quantile.go
+++ b/promql/quantile.go
@@ -61,6 +61,8 @@ type metricWithBuckets struct {
 // happening during evaluations of AST functions, we should report those
 // explicitly):
 //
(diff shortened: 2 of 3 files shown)


brian-brazil:
That looks right. The function comment and docs will also need updating.

JBerny:
I updated the docs and function comments

docs/querying/functions.md
```diff
@@ -185,6 +185,8 @@ for quantiles located in the lowest bucket.
 If `b` contains fewer than two buckets, `NaN` is returned. For φ < 0, `-Inf` is
 returned. For φ > 1, `+Inf` is returned.
 
+If `b` has 0 observations, `NaN` is returned.
```
brian-brazil, line 188: You can put this in with the previous paragraph with the other special case that returns NaN

JBerny: Right. I put it in the same order as the function docs

brian-brazil:
Thanks!

PR #7323, opened 2020-06-01
increase the remote write bucket range

BertHartm:
Increase the range of remote write buckets to capture times above 10s for laggy scenarios
Buckets had been: {.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}
Buckets are now: {0.03125, 0.0625, 0.125, 0.25, 0.5, 1, 2, 4, 8, 16, 32, 64, 128, 256, 512}

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -115,7 +115,7 @@ func newQueueManagerMetrics(r prometheus.Registerer, rn, e string) *queueManager
 		Subsystem:   subsystem,
 		Name:        "sent_batch_duration_seconds",
 		Help:        "Duration of sample batch send calls to the remote storage.",
-		Buckets:     append(prometheus.DefBuckets, 20, 45, 60, 150, 300),
+		Buckets:     append(prometheus.DefBuckets, 25, 60, 120, 300),
 		ConstLabels: constLabels,
 	})
 	m.highestSentTimestamp = &maxGauge{
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -115,7 +115,7 @@ func newQueueManagerMetrics(r prometheus.Registerer, rn, e string) *queueManager
 		Subsystem:   subsystem,
 		Name:        "sent_batch_duration_seconds",
 		Help:        "Duration of sample batch send calls to the remote storage.",
-		Buckets:     prometheus.DefBuckets,
+		Buckets:     prometheus.ExponentialBuckets(.03125, 2, 15),
 		ConstLabels: constLabels,
 	})
 	m.highestSentTimestamp = &maxGauge{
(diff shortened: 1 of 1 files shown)


storage/remote/queue_manager.go
```diff
@@ -115,7 +115,7 @@ func newQueueManagerMetrics(r prometheus.Registerer, rn, e string) *queueManager
 		Subsystem:   subsystem,
 		Name:        "sent_batch_duration_seconds",
 		Help:        "Duration of sample batch send calls to the remote storage.",
-		Buckets:     prometheus.DefBuckets,
+		Buckets:     prometheus.ExponentialBuckets(.03125, 2, 15),
```
csmarchbanks, line 118: Rather than redefining several buckets that people might be using for SLIs, what would you think of `append(prometheus.DefBuckets, 30, etc...)`? It is also a bit easier to know what the range is at a glance that way.

BertHartm:
for reference, the default remote timeout is 30s, and I've found it useful to make that much higher in cases where I expect bad network or slow receivers.

I tried to pick an exponential series that kept a few of the original buckets, and also came near useful numbers (1s, 1m, etc).

csmarchbanks:
I think I was writing my comment while you commented :)

I am curious what kinds of values you update the remote write timeout to? In a quick github search I only found one instance of increasing the default, and that was to 60s.

storage/remote/queue_manager.go
BertHartm: My original thought was to keep config simpler with using one fo the generated bucket sizes, but that wasn't terribly straightforward, so this seems reasonable.

I'm wondering if there should be a bucket at 20 or 25 instead of 30, since I don't know if a timeout would register for the 30s bucket.

BertHartm:
> I think I was writing my comment while you commented :)
> 
> I am curious what kinds of values you update the remote write timeout to? As I said, in a quick github search I only found one instance of increasing the default, and that was to 60s.

I pushed it to 4 minutes (fairly arbitrarily). This both increased the number of successful calls, and also reduced retries on our degraded networks (which made the network marginally happier).

To cover those use cases, what about adding `20, 45, 60, 150, 300` as new buckets?

cstyan:
> To cover those use cases, what about adding 20, 45, 60, 150, 300 as new buckets?

Adding these new buckets instead of modifying existing buckets seems like the best way to proceed IMO.

BertHartm:
@cstyan I pushed an update with those buckets added.

storage/remote/queue_manager.go
csmarchbanks: How about `25, 60, 120, 300` That keeps with our other ratios of 2-2.5x increase in size between buckets. It's a bit weird to add 25s, then only 15s when building a histogram.

BertHartm: @csmarchbanks makes sense, updated as proposed

BertHartm:
> My only general concern is that timeouts over one minute aren't really supported right now since reshard and shutdown calculations will force kill all shards after one minute. I think I am ok merging this, but perhaps we should extend the shard shutdown timeout to `max(1 minute, send timeout)`?

That feels like a bug. I'd be happy to take a look at fixing (as a person with the usecase it would help). Any objections to having it as a second PR?

PR #7324, opened 2020-06-02
Update RELEASE doc

SuperQ:
* Simplify tag snipit slightly.
* Add a git alias helper.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -103,10 +103,19 @@ Entries in the `CHANGELOG.md` are meant to be in this order:
 Tag the new release via the following commands:
 
 ```bash
-$ tag=$(< VERSION)
-$ git tag -s "v${tag}" -m "v${tag}"
-$ git push origin "v${tag}"
+$ tag="v$(< VERSION)"
+$ git tag -s "${tag}" -m "${tag}"
+$ git push origin "${tag}"
 ```
+
+Optionally, you can use this handy `.gitconfig` alias.
+
+```ini
+[alias]
+  tag-release = "!f() { tag=v${1:-$(cat VERSION)} ; git tag -s ${tag} -m ${tag} && git push origin ${tag}; }; f"
+```
+
+Then release with `git tag-release`.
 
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -103,11 +103,20 @@ Entries in the `CHANGELOG.md` are meant to be in this order:
...
 ```bash
-$ tag=$(< VERSION)
-$ git tag -s "v${tag}" -m "v${tag}"
-$ git push origin "v${tag}"
+$ tag="v$(< VERSION))"
+$ git tag -s "${tag}" -m "${tag}"
+$ git push origin "${tag}"
 ```
 
+Optionally, you can use this handy `.gitconfig` alias.
```
brian-brazil, line 111: I'm not sure about asking people to adjust their personal config files like this, if they want to this they already can.

RichiH: It's marked as optional and is not required for the workflow, which seems fine to me.

roidelapluie: We should make the alias read VERSION then

RELEASE.md
```diff
@@ -103,11 +103,20 @@ Entries in the `CHANGELOG.md` are meant to be in this order:
 Tag the new release via the following commands:
 
 ```bash
-$ tag=$(< VERSION)
-$ git tag -s "v${tag}" -m "v${tag}"
-$ git push origin "v${tag}"
+$ tag="v$(< VERSION))"
```
simonpasquier, line 106: extra trailing `)`

RELEASE.md
SuperQ: I wasn't sure how much shell functionality I could assume in git aliases. Doing some tests, it looks somewhat limited.

This works:
```
tag-release = "!f() { v=${1:-$(cat VERSION)} ; echo git tag -s v$v -m v$v && echo git push origin v$v; }; f"
```

But this fails:
```
tag-release = "!f() { v=${1:-$(< VERSION)} ; echo git tag -s v$v -m v$v && echo git push origin v$v; }; f"
```

RELEASE.md
SuperQ: Whups.

brian-brazil:
:+1:

PR #7331, opened 2020-06-03
Update dependencies

codesome:
Part of 2.19 release. `make update-go-deps` tried to upgrade `k8s.io/client-go` to some version which was causing compilation issues and the current version [ (v0.18.3) seems to be the latest](https://github.com/kubernetes/client-go/releases). So I have skipped it.


brian-brazil:
:+1:

Issue #7242, opened 2020-05-11
Add metrics around remote_read queries


amitabh91:
I have integrated M3DB as remote storage for Prometheus as a long term storage solution. Where M3coordinator runs as a side along with Prometheus which pushes the data to the M3DB cluster. I am able to write the metrics data into M3DB successfully. 

On executing the query prometheus_remote_storage_sent_bytes_total from Prom UI we see some incremental values, unlike prometheus_remote_storage_remote_read_queries. 

I am attaching the screenshot below for the quick reference 

**Write Data Logs**
![image](https://user-images.githubusercontent.com/7939820/81582108-41382a00-93cd-11ea-8a5d-eb67582e3894.png)

**Read Data Logs**
![image](https://user-images.githubusercontent.com/7939820/81582208-6167e900-93cd-11ea-8e68-d62b13956f32.png)

**Environment**
* System information:

   Linux 3.10.0-957.el7.x86_64 x86_64

* Prometheus version:

  prometheus, version 2.15.2 
  build user:       root@688433cf4ff7
  build date:       20200106-14:50:51
  go version:       go1.13.5

* Prometheus configuration file:
```
remote_write:
- url: http://localhost:7201/api/v1/prom/remote/write
  remote_timeout: 30s
  queue_config:
    capacity: 500
    max_shards: 1000
    min_shards: 1
    max_samples_per_send: 100
    batch_send_deadline: 5s
    min_backoff: 30ms
    max_backoff: 100ms
remote_read:
- url: http://localhost:7201/api/v1/prom/remote/read
  remote_timeout: 1m
  read_recent: true
```
Whenever we have mismatch in the  namespace value of  m3coordinator and M3DB cluster we see the below error while reading.

**m3coordinaor config**
```
clusters:
  - namespaces:
      - namespace: default
        retention: 48h
        type: unaggregated
    client:
      config:
        service:
          env: default_env
          zone: embedded
          service: m3db
          cacheDir: /var/lib/m3kv
          etcdClusters:
            - zone: embedded
              endpoints:
                - http://etcd.infra.svc.cluster.local:2379
```
**Error** 
![image](https://user-images.githubusercontent.com/7939820/81577182-c409b680-93c6-11ea-89e2-05cbe9bbec2f.png)

And with the correct namespace we don't see the error but prometheus_remote_storage_remote_read_queries still remains empty. So I believe this is a bug that needs a fix.

[...]

PR #7334, opened 2020-06-04
Added the remote read histogram

njingco:
Added the  remote read queries Histogram metrics:
- `prometheus_remote_storage_remote_read_request_duration_seconds_bucket` 
- `prometheus_remote_storage_remote_read_request_duration_seconds_count`
- `prometheus_remote_storage_remote_read_request_duration_seconds_sum`

Additional feature to PR #7328  and Issue #7242. 

Open Issue:
-  Adding status code label to the `prometheus_remote_storage_remote_read_queries_total` counter.


storage/remote/read.go
```diff
@@ -42,16 +43,27 @@ var remoteReadQueriesTotal = prometheus.NewCounterVec(
 	[]string{remoteName, endpoint},
 )
 
+var remoteReadQueriesHistogram = prometheus.NewHistogramVec(
+	prometheus.HistogramOpts{
+		Namespace: namespace,
+		Subsystem: subsystem,
+		Name:      "remote_read_request_duration_seconds",
+		Help:      "Histogram of the latency for remote read requests.",
+		Buckets:   prometheus.DefBuckets,
```
csmarchbanks, line 52: Since the default remote read timeout is 60s I think it would be worth appending 25s and 60s as additional buckets on top of the default.

cstyan:
So there's a few things happening here.

There's already precedence for a counter that has a status code label, and a histogram that doesn't: https://github.com/prometheus/prometheus/blob/master/web/web.go#L114-L129

And I think because we're using GRPC for remote write/read we can't easily get the HTTP status code without either importing `httpgrpc` from `weaveworks/common` or duplicating coding from there. We should use the same pattern from remote write however, and include a metric for failed requests: https://github.com/prometheus/prometheus/blob/master/storage/remote/queue_manager.go#L78-L91

So we could have `queries_total` and `queries_failed_total`.

cc @gouthamve

csmarchbanks:
Changing the counter to only count failed, and have the histogram count everything seems like it would work well. We still wouldn't need `queries_total` in addition to the histogram.

cstyan:
> Changing the counter to only count failed, and have the histogram count everything seems like it would work well. We still wouldn't need queries_total in addition to the histogram.

I see what you're saying, good idea.

brian-brazil:
> prometheus_remote_storage_remote_read_request_duration_seconds_bucket

This is kinda long, remote doesn't need to be said twice.

> I think because we're using GRPC for remote write/read

We don't use GRPC for either, it's all HTTP.

cstyan:
> We don't use GRPC for either, it's all HTTP.

As far as `Select` is concerned it might as well be, the clients `Read` function returns the `prompb` response which does not contain the HTTP response and status code.

storage/remote/read.go
cstyan: @njingco you can look at this PR as an example of how to add additional buckets: https://github.com/prometheus/prometheus/pull/7323

njingco: yeah for sure!

storage/remote/client.go
```diff
@@ -40,6 +42,26 @@ const maxErrMsgLen = 256
...
+var remoteReadQueriesRequestTotal = prometheus.NewCounterVec(
+	prometheus.CounterOpts{
+		Namespace: namespace,
+		Subsystem: subsystem,
+		Name:      "read_queries_request_total",
+		Help:      "Counter of remote read queries by HTTP status code.",
+	},
+	[]string{remoteName, endpoint, "code"},
+)
+
```
cstyan, line 64: This should be one metric, I'd keep the name/help text from the first.

storage/remote/client.go
```diff
@@ -55,6 +77,11 @@ type ClientConfig struct {
 	HTTPClientConfig config_util.HTTPClientConfig
 }
 
+// Initalize Counter
```
cstyan, line 80: we don't need a comment here until some point in the future when we're maybe doing something more complex in `init` :)

storage/remote/client.go
```diff
@@ -185,6 +212,14 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
 	}
 
 	httpResp, err := c.client.Do(httpReq)
+
+	remoteReadTotalCounter := remoteReadQueriesTotal.WithLabelValues(c.remoteName, c.url.String())
+	remoteReadTotalCounter.Inc()
+
+	fmt.Println("Status Code: ", httpResp.StatusCode)
```
cstyan, line 219: debug println :)

storage/remote/client.go
```diff
@@ -195,10 +230,15 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
 
 	compressed, err = ioutil.ReadAll(httpResp.Body)
 	if err != nil {
+		remoteReadRequestTotalCounter = remoteReadQueriesRequestTotal.WithLabelValues(c.remoteName, c.url.String(), "504")
```
cstyan, line 233: Above you've already called `remoteReadRequestTotalCounter := remoteReadQueriesRequestTotal.WithLabelValues(c.remoteName, c.url.String(), strconv.Itoa(httpResp.StatusCode))`, which is correct. We've got an HTTP response which should have an error code. The only thing to change up there is that we need to check `if err != nil` before we access `httpResp` to get the status code. The incrementing of the counter can also be done above.

Here (and below) we shouldn't have to make add any instrumentation code, errors with reading the body, decoding, etc. are separate from the metric that just counts the # of remote read queries executed.

njingco:
> Generally this seems to be coming together well, however we are missing out on a metric accounting for failures such as connection refused or timeouts in the current approach. Is that something this PR intends to address?

We can do it in another PR and open an issue for this.

PR #7335, opened 2020-06-04
Pending Samples metric includes samples in channel

### LINKED ISSUE #6952: prometheus_remote_storage_pending_samples are drastically increasing over time.

csmarchbanks:
The pending samples metric should also include samples waiting in the
channels to be sent to provide a more accurate measure. In addition,
make sure that the pending samples is reset to 0 anytime a queue is
started as we remake all of the shards at that time.

This may fix #6952 as most increases of the metric happened around
resharding and now the metric will be reset to zero then.

### DIFF
--- a/storage/remote/queue_manager.go
+++ b/storage/remote/queue_manager.go
@@ -360,7 +360,6 @@ outer:
 func (t *QueueManager) Start() {
 	// Initialise some metrics.
 	t.metrics.shardCapacity.Set(float64(t.cfg.Capacity))
-	t.metrics.pendingSamples.Set(0)
 	t.metrics.maxNumShards.Set(float64(t.cfg.MaxShards))
 	t.metrics.minNumShards.Set(float64(t.cfg.MinShards))
 	t.metrics.desiredNumShards.Set(float64(t.cfg.MinShards))
@@ -675,6 +674,9 @@ func (s *shards) start(n int) {
 	s.mtx.Lock()
 	defer s.mtx.Unlock()
 
+	s.qm.metrics.pendingSamples.Set(0)
+	s.qm.metrics.numShards.Set(float64(n))
+
 	newQueues := make([]chan sample, n)
 	for i := 0; i < n; i++ {
 		newQueues[i] = make(chan sample, s.qm.cfg.Capacity)
@@ -690,7 +692,6 @@ func (s *shards) start(n int) {
 	for i := 0; i < n; i++ {
 		go s.runShard(hardShutdownCtx, i, newQueues[i])
 	}
-	s.qm.metrics.numShards.Set(float64(n))
 }
 
 // stop the shards; subsequent call to enqueue will return false.
@@ -740,6 +741,7 @@ func (s *shards) enqueue(ref uint64, sample sample) bool {
 	case <-s.softShutdown:
 		return false
 	case s.queues[shard] <- sample:
+		s.qm.metrics.pendingSamples.Inc()
 		return true
 	}
 }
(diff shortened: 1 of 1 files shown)


storage/remote/queue_manager.go
```diff
@@ -777,6 +779,7 @@ func (s *shards) runShard(ctx context.Context, shardID int, queue chan sample) {
 	for {
 		select {
 		case <-ctx.Done():
+			s.qm.metrics.pendingSamples.Sub(float64(nPending + len(queue)))
```
cstyan, line 782: My only concern is that this + `Set(0)` in start would just hide the actual issue. Maybe we should have another metric that's "samples dropped because their shard was shutdown but was unable to flush it's samples within the timeout"?

csmarchbanks: Good call, I have now added another line after this to add the dropped samples to the failed samples metric. Does that seem reasonable to you?

cstyan: I think that should be okay, since we also increment dropped samples for samples that we don't see a labels cache entry for (no entry for ref ID). But we should possibly also log this?

csmarchbanks: We already have the log for `Failed to flush all samples on shutdown`, perhaps move that log line here? Or are you thinking a separate log line?

cstyan: Hmm. I think that log line is enough but it would be nice to add the # of samples we failed to flush to that log line. If that's a non-trivial change then we can just merge this as is and file that as a nice to have in the remote write meta issue.

csmarchbanks: :+1: I added a commit with one way of adding the # of samples to the log line. I thought it would be the least invasive, but happy to try something else too.

csmarchbanks: @cstyan bump for another review on the most recent changes when you get a chance!

PR #7342, opened 2020-06-04
Only relabelling should apply for scrape_samples_scraped_post_relabel…

brian-brazil:
…ling.

More consistent variable names.

Fixes #7298

PR #7345, opened 2020-06-04
Add regression test for scrape_samples_scraped_post_relabelling metric

roidelapluie:

### DIFF
--- a/scrape/scrape_test.go
+++ b/scrape/scrape_test.go
@@ -1928,3 +1928,32 @@ func TestCheckAddError(t *testing.T) {
 	sl.checkAddError(nil, nil, nil, storage.ErrOutOfOrderSample, nil, &appErrs)
 	testutil.Equals(t, 1, appErrs.numOutOfOrder)
 }
+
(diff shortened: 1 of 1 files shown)


brian-brazil:
TestScrapeLoopAppendGracefullyIfAmendOrOutOfOrderOrOutOfBounds seems to already cover this.

roidelapluie:
It does not test that the removed/relabeled items are not counted

brian-brazil:
TestScrapeLoopAppendSampleLimit does though.

roidelapluie:
I will try to think why that test did not catch the regression then

roidelapluie:
https://github.com/prometheus/prometheus/pull/7342/files did not touch TestScrapeLoopAppendSampleLimit

roidelapluie:
https://github.com/prometheus/prometheus/pull/7342/files did not touch TestScrapeLoopAppendSampleLimit because there was an exception on ErrAppendSampleLimit and that test always produces that.

brian-brazil:
That exception was removed though, so the existing tests should now cover things sufficiently.

PR #7351, opened 2020-06-05
promql: cleanup: use errors.As

roidelapluie:
This was TODO because circleci was not in go1.13 yet.


brian-brazil:
:+1:

PR #7354, opened 2020-06-06
added streaming remote read client side functionality

Sudhar287:
For #5926 
Thank you to @YaoZengzeng for his initial design. I've [implemented](https://github.com/prometheus/prometheus/issues/5926#issuecomment-526145047) his comments from here.
FYI @bwplotka

### DIFF
--- a/storage/remote/client.go
+++ b/storage/remote/client.go
@@ -36,6 +36,7 @@ import (
 
 	"github.com/opentracing-contrib/go-stdlib/nethttp"
 	"github.com/prometheus/prometheus/prompb"
+	"github.com/prometheus/prometheus/storage"
 )
 
 const maxErrMsgLen = 512
@@ -99,7 +100,7 @@ type ClientConfig struct {
 // ReadClient uses the SAMPLES method of remote read to read series samples from remote server.
 // TODO(bwplotka): Add streamed chunked remote read method as well (https://github.com/prometheus/prometheus/issues/5926).
 type ReadClient interface {
-	Read(ctx context.Context, query *prompb.Query) (*prompb.QueryResult, error)
+	StartRead(ctx context.Context, query *prompb.Query, sortSeries bool) (storage.SeriesSet, error)
 }
 
 // NewReadClient creates a new client for remote read.
@@ -214,7 +215,7 @@ func (c Client) Endpoint() string {
 }
 
 // Read reads from a remote endpoint.
-func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryResult, error) {
+func (c *Client) StartRead(ctx context.Context, query *prompb.Query, sortSeries bool) (storage.SeriesSet, error) {
 	c.readQueries.Inc()
 	defer c.readQueries.Dec()
 
@@ -223,6 +224,10 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
 		// as the protobuf interface allows for it.
 		Queries: []*prompb.Query{
 			query,
+		},
+		AcceptedResponseTypes: []prompb.ReadRequest_ResponseType{
+			prompb.ReadRequest_STREAMED_XOR_CHUNKS,
+			prompb.ReadRequest_SAMPLES,
 		},
 	}
 	data, err := proto.Marshal(req)
@@ -262,6 +267,28 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
 	if err != nil {
 		return nil, errors.Wrap(err, "error sending request")
 	}
+
+	contentType := httpResp.Header.Get("Content-Type")
+	if strings.HasPrefix(contentType, "application/x-protobuf") {
+		return c.handleSampledResponse(httpResp, start, req, sortSeries)
+	}
+
+	if httpResp.StatusCode/100 != 2 {
+		httpResp.Body.Close()
+		return nil, errors.Errorf("remote server %s returned HTTP status %s: %s", c.url.String(), httpResp.Status, strings.TrimSpace(string(compressed)))
+	}
+
+	if !strings.HasPrefix(contentType, "application/x-streamed-protobuf; proto=prometheus.ChunkedReadResponse") {
+		httpResp.Body.Close()
+		return nil, errors.Errorf("not supported remote read content type: %s", contentType)
+	}
+
+	// Handling streamed response
+	stream := NewChunkedReader(httpResp.Body, DefaultChunkedReadLimit, nil)
+	return &chunkedResponseSeriesSet{stream: stream, httpResp: httpResp}, nil
+}
+
+func (c *Client) handleSampledResponse(httpResp *http.Response, start time.Time, req *prompb.ReadRequest, sortSeries bool) (storage.SeriesSet, error) {
 	defer func() {
 		io.Copy(ioutil.Discard, httpResp.Body)
 		httpResp.Body.Close()
@@ -269,15 +296,10 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
 	c.readQueriesDuration.Observe(time.Since(start).Seconds())
 	c.readQueriesTotal.WithLabelValues(strconv.Itoa(httpResp.StatusCode)).Inc()
 
-	compressed, err = ioutil.ReadAll(httpResp.Body)
+	compressed, err := ioutil.ReadAll(httpResp.Body)
 	if err != nil {
 		return nil, errors.Wrap(err, fmt.Sprintf("error reading response. HTTP status code: %s", httpResp.Status))
 	}
-
-	if httpResp.StatusCode/100 != 2 {
-		return nil, errors.Errorf("remote server %s returned HTTP status %s: %s", c.url.String(), httpResp.Status, strings.TrimSpace(string(compressed)))
-	}
-
 	uncompressed, err := snappy.Decode(nil, compressed)
 	if err != nil {
 		return nil, errors.Wrap(err, "error reading response")
@@ -292,6 +314,5 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
 	if len(resp.Results) != len(req.Queries) {
 		return nil, errors.Errorf("responses: want %d, got %d", len(req.Queries), len(resp.Results))
 	}
-
-	return resp.Results[0], nil
+	return FromQueryResult(sortSeries, resp.Results[0]), nil
 }
--- a/storage/remote/codec.go
+++ b/storage/remote/codec.go
@@ -160,6 +160,125 @@ func FromQueryResult(sortSeries bool, res *prompb.QueryResult) storage.SeriesSet
 	}
 }
 
+// chunkedResponseSeriesSet implements storage.SeriesSet.
+type chunkedResponseSeriesSet struct {
+	stream        *ChunkedReader
+	currentChunks []prompb.Chunk
+	currentLabels []prompb.Label
+	httpResp      *http.Response
+	err           error
+}
+
+func (c *chunkedResponseSeriesSet) Next() bool {
+	c.currentLabels = nil
+	c.currentChunks = nil
+	res := &prompb.ChunkedReadResponse{}
+
+	if err := c.stream.NextProto(res); err != nil {
+		if err != io.EOF {
+			c.err = err
+		}
+		c.httpResp.Body.Close()
+		return false
+	}
+
+	c.currentLabels = res.ChunkedSeries[0].Labels
+	c.currentChunks = append(c.currentChunks, res.ChunkedSeries[0].Chunks...)
+	return true
+}
+
+func (c *chunkedResponseSeriesSet) At() storage.Series {
+	return &chunkSeries{
+		labels: prompb.Labels{Labels: c.currentLabels},
+		chunks: c.currentChunks,
+	}
+}
+
+func (c *chunkedResponseSeriesSet) Warnings() storage.Warnings { return nil }
+
+func (c *chunkedResponseSeriesSet) Err() error {
+	return c.err
+}
+
+// chunkSeries implements storage.Series
+type chunkSeries struct {
+	labels     prompb.Labels
+	chunks     []prompb.Chunk
+	mint, maxt int64
+}
+
+func (c *chunkSeries) Labels() labels.Labels {
+	return labelProtosToLabels(c.labels.Labels)
+}
+
+func (c *chunkSeries) Iterator() chunkenc.Iterator {
+	streamingChunkSeriesIterator := &chunkSeriesIterator{chunks: c.chunks, i: 0, mint: c.mint, maxt: c.maxt}
+	streamingChunkSeriesIterator.resetCurIterator()
+	return streamingChunkSeriesIterator
+}
+
+type chunkSeriesIterator struct {
+	// Index poiting to a specific chunk in []prompb.Chunk.
+	i int
+	// Current chunks iterator.
+	cur        chunkenc.Iterator
+	maxt, mint int64
+	chunks     []prompb.Chunk
+	err        error
+}
+
+func (it *chunkSeriesIterator) resetCurIterator() {
+	currectChunk := it.chunks[it.i]
+	it.mint = currectChunk.MinTimeMs
+	it.maxt = currectChunk.MaxTimeMs
+	chk, err := chunkenc.FromData(chunkenc.Encoding(currectChunk.Type), currectChunk.Data)
+	it.err = err
+	it.cur = chk.Iterator(nil)
+}
+
+func (it *chunkSeriesIterator) At() (int64, float64) {
+	return it.cur.At()
+}
+
+func (it *chunkSeriesIterator) Next() bool {
+	if it.cur.Next() {
+		return true
+	}
+	if it.i == len(it.chunks)-1 {
+		return false
+	}
+	it.i++
+	it.resetCurIterator()
+	return it.Next()
+}
+
+// Seek advances the iterator to the element at time t or greater.
+func (it *chunkSeriesIterator) Seek(t int64) bool {
+	if t > it.maxt {
+		return false
+	}
+	if t < it.mint {
+		t = it.mint
+	}
+	for ; it.chunks[it.i].MaxTimeMs < t; it.i++ {
+		if it.i == len(it.chunks)-1 {
+			return false
+		}
+	}
+	it.resetCurIterator()
+	for it.cur.Next() {
+		t0, _ := it.cur.At()
+		if t0 >= t {
+			return true
+		}
+	}
+	return false
+}
+
+func (it *chunkSeriesIterator) Err() error {
+	return it.err
+}
+
 // NegotiateResponseType returns first accepted response type that this server supports.
 // On the empty accepted list we assume that the SAMPLES response type was requested. This is to maintain backward compatibility.
 func NegotiateResponseType(accepted []prompb.ReadRequest_ResponseType) (prompb.ReadRequest_ResponseType, error) {
--- a/storage/remote/codec_test.go
+++ b/storage/remote/codec_test.go
@@ -14,12 +14,17 @@
 package remote
 
 import (
+	"bytes"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 	"testing"
 
+	"github.com/gogo/protobuf/proto"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/prompb"
 	"github.com/prometheus/prometheus/storage"
+	"github.com/prometheus/prometheus/tsdb/tsdbutil"
 	"github.com/prometheus/prometheus/util/testutil"
 )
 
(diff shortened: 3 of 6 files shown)


storage/remote/codec.go
```diff
@@ -167,6 +167,122 @@ func FromQueryResult(sortSeries bool, res *prompb.QueryResult) storage.SeriesSet
...
+	currentChunks []prompb.Chunk
+	currentLabels []prompb.Label
+	closer        io.Closer
+	responseError error
+}
+
+func (c *chunkedResponseSeriesSet) Next() bool {
+	c.currentLabels = nil
+	c.currentChunks = nil
+	res := &prompb.ChunkedSeries{}
```
wwformat, line 183: i think this should be use prompb.ChunkedReadResponse type for unmarshal, like [thanos](https://github.com/thanos-io/thanos/blob/c733564d44745af1a023bfa5d51d6d205404dc82/pkg/store/prometheus.go#L285)

Sudhar287: Hmm.
WDYT @bwplotka ?

bwplotka: It's kind of easy to say. Just check what is Written by remote read handler (: 

And it's here: https://github.com/prometheus/prometheus/blob/b788986717e1597452ca25e5219510bb787165c7/storage/remote/codec.go#L359

so hope that answers your question (:

storage/remote/codec.go
```diff
@@ -167,6 +167,122 @@ func FromQueryResult(sortSeries bool, res *prompb.QueryResult) storage.SeriesSet
 	}
 }
 
+// design approach taken from https://prometheus.io/blog/#solution
```
bwplotka, line 170: Not a full sentence

Sudhar287:
Sorry I didn't try @wwformat suggestion before reaching you on slack because I thought using `&prompb.ChunkedSeries` is right as no objections about it were raised on my previous PR. But now I think I should have done it. Sorry!

Sudhar287:
Actively working on the fix and some issues with `TestEndpoints`. Will try to get the revision within a week. :-)

storage/remote/codec_test.go
```diff
@@ -229,3 +232,147 @@ func TestMergeLabels(t *testing.T) {
 		testutil.Equals(t, tc.expected, MergeLabels(tc.primary, tc.secondary))
 	}
 }
+
+func TestChunkSeriesSet(t *testing.T) {
+	b := &bytes.Buffer{}
+	f := &mockedFlusher{}
+	w := NewChunkedWriter(b, f)
+
+	chk1 := tsdbutil.ChunkFromSamples([]tsdbutil.Sample{tsdbutil.NewSample(1, 10), tsdbutil.NewSample(2, 20)})
```
Sudhar287, line 241: NewSample is being used here.

tsdb/tsdbutil/buffer.go
```diff
@@ -111,6 +111,10 @@ func (s sample) V() float64 {
 	return s.v
 }
 
+func NewSample(time int64, value float64) Sample {
```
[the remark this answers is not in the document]
Sudhar287: Yes, this is how a new samples are created to make a chunk. Its used here: https://github.com/prometheus/prometheus/pull/7354/files#diff-22f528c8a1ef9be43f923778809217aaR241

storage/remote/client.go
```diff
@@ -257,22 +262,36 @@ func (c *client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
...
 	}
+
+	contentType := httpResp.Header.Get("Content-Type")
+	if strings.HasPrefix(contentType, "application/x-protobuf") {
+		return c.handleSampledResponse(httpResp, start, req, sortSeries)
+	}
+
+	if httpResp.StatusCode/100 != 2 {
+		io.Copy(ioutil.Discard, httpResp.Body)
+		httpResp.Body.Close()
```
Sudhar287, line 273: I hope this is what you meant.

storage/remote/codec_test.go
bwplotka: But this is wrong as I suggested in the comment or I hope to make. The idea is to rather implement separate `sample{` type here not on tsdbutil.

Sudhar287: hmm okay. But why is this necessary? :thinking:

bwplotka: Hey, sorry I should explain.

We do that for purpose to have clear init statement. Instead of `tsdbutil.NewSample(1, 10)` we can do `sample{ }` which is just much less code to read when we do that a lot. Since we do this everywhere, I feel like we should be make it consistent with rest of the code. IF we want to change it, there should be another PR changing thousands of places like this where we define `type sample struct` etc. BTW that's why tsdbutil.Sample is just interface.

storage/remote/client.go
bwplotka: Why not having:

```
	defer func() {
		io.Copy(ioutil.Discard, httpResp.Body)
		httpResp.Body.Close()
}()
```
here?

storage/remote/codec_test.go
bwplotka: still not addressed (:

storage/remote/client.go
Sudhar287: Really sorry Bartek. Missed the notification.

storage/remote/client.go
```diff
@@ -257,22 +262,38 @@ func (c *client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
...
 		return nil, errors.Wrap(err, "error sending request")
 	}
+
+	contentType := httpResp.Header.Get("Content-Type")
+	if strings.HasPrefix(contentType, "application/x-protobuf") {
+		return c.handleSampledResponse(httpResp, start, req, sortSeries)
+	}
+
+	if httpResp.StatusCode/100 != 2 {
+		defer func() {
```
bwplotka, line 272: this does not help.

Please look on this function and think about what are the cases when we need to close the body and exhaust? I think you are missing one.

I think because we run functions it's totally ok to defer this in all cases JUST after we get response. and let's make sure nothing closes twice inside functions WDYT?

Sudhar287: Thanks for taking a look at my code Bartek.
I think you are referring to closing the `httpRest.Body` over here:
```
func (c *client) handleStreamedResponse(httpResp *http.Response) (storage.SeriesSet, error) {
	stream := NewChunkedReader(httpResp.Body, DefaultChunkedReadLimit, nil)
	return &chunkedResponseSeriesSet{stream: stream}, nil
}
``` 
So I have to `close()` the `bufio.Reader` in `ChunkedReader`?

Sudhar287: PLTA! :-)

storage/remote/client.go
bwplotka: Still not addressed (:

storage/remote/codec.go
```diff
@@ -160,6 +160,123 @@ func FromQueryResult(sortSeries bool, res *prompb.QueryResult) storage.SeriesSet
...
+func (it *chunkSeriesIterator) resetCurIterator() error {
+	currectChunk := it.chunks[it.i]
+	it.mint = currectChunk.MinTimeMs
+	it.maxt = currectChunk.MaxTimeMs
+	chk, err := chunkenc.FromData(chunkenc.Encoding(currectChunk.Type), currectChunk.Data)
+	if err != nil {
+		return err
+	}
+	it.cur = chk.Iterator(nil)
+	return nil
```
bwplotka, line 234: Why we return error if we never use it?

storage/remote/codec.go
```diff
@@ -160,6 +160,123 @@ func FromQueryResult(sortSeries bool, res *prompb.QueryResult) storage.SeriesSet
...
+	currentLabels []prompb.Label
+	httpResp      *http.Response
+}
+
+func (c *chunkedResponseSeriesSet) Next() bool {
+	c.currentLabels = nil
+	c.currentChunks = nil
+	res := &prompb.ChunkedReadResponse{}
+	if err := c.stream.NextProto(res); err != nil {
+		if c.httpResp != nil {
```
bwplotka, line 176: I think we should handle io.EOF differently then casual error, no? (:

bwplotka (approved):
I see two things missing. Otherwise LGTM!

storage/remote/codec.go
Sudhar287: I am sorry. I dont understand what you mean. Can you please elaborate Bartek?

bwplotka: Sure, This special error sentinel means EOF - End of File

There is good explanation here: https://golang.org/pkg/io/#pkg-variables

storage/remote/client.go
bwplotka: still not addressed, in case below with unsupported contentType the response body will be not closed/exhausted leaving the connection alive without good reason.

storage/remote/client.go
```diff
@@ -287,6 +306,10 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
 	if len(resp.Results) != len(req.Queries) {
 		return nil, errors.Errorf("responses: want %d, got %d", len(req.Queries), len(resp.Results))
 	}
+	return FromQueryResult(sortSeries, resp.Results[0]), nil
+}
 
-	return resp.Results[0], nil
+func (c *Client) handleStreamedResponse(httpResp *http.Response) (storage.SeriesSet, error) {
+	stream := NewChunkedReader(httpResp.Body, DefaultChunkedReadLimit, nil)
```
bwplotka, line 313: This method might be too shallow I would just inline this.

storage/remote/codec.go
bwplotka: Why this could be nil?  What if error is non-nil and not io EOF? It would be nice to return it in SeriesSet.Err right? (:

bwplotka (changes requested):
:wave: 

Let's get back to this if you want (: Added a few comments, otherwise good :+1: Also some old comments are still not addressed (again). Just letting you know that it's natural that if maintainers see that their comments are not addressed (nothing done, nothing responded, nothing adjusted, just ignored) it might mean that ther time is being wasted. If you wish to contribute and have PR merged fasted, you need to keep attention to details like this. (:

storage/remote/codec.go
```diff
@@ -167,6 +167,118 @@ func FromQueryResult(sortSeries bool, res *prompb.QueryResult) storage.SeriesSet
...
+	cur        chunkenc.Iterator
+	maxt, mint int64
+	chunks     []prompb.Chunk
+}
+
+func (it *chunkSeriesIterator) resetCurIterator() {
+	currectChunk := it.chunks[it.i]
+	it.mint = currectChunk.MinTimeMs
+	it.maxt = currectChunk.MaxTimeMs
+	chk, _ := chunkenc.FromData(chunkenc.Encoding(currectChunk.Type), currectChunk.Data)
```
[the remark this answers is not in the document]
bwplotka: Not addressed still. Why you add +1 emoji if you never actually address this? (:

bwplotka (commented):
Also it's ok to say that you are busy and cannot complete this PR. Let's just close it and allow others to finish this if they have time :+1:  (:

storage/remote/codec.go
Sudhar287: This was done in order to address your comments on: about closing the connection.
```
if httpResp.StatusCode/100 != 2 {
		io.Copy(ioutil.Discard, httpResp.Body)
		httpResp.Body.Close()
```

storage/remote/codec.go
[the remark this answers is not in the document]
Sudhar287: My mistake sorry. I thought I had address this earlier itself. Sorry overlooking this.

storage/remote/client.go
Sudhar287: Sorry I got confused with this and closing the HTTP connection for the streamed use-case. 
I solved the issue of closing the `streaming` HTTP connection and considered this ^ issue solved too.
```
res := &prompb.ChunkedReadResponse{}
	if err := c.stream.NextProto(res); err == io.EOF {
		if c.httpResp != nil {
			c.httpResp.Body.Close()
		}
		return false
```

Sudhar287: To solve the problem with the unsupported contentType, is the additional line here enough to solve the issue?
```
	if !strings.HasPrefix(contentType, "application/x-streamed-protobuf; proto=prometheus.ChunkedReadResponse") {
-->		httpResp.Body.Close()
		return nil, errors.Errorf("not supported remote read content type: %s", contentType)
	}
```

storage/remote/codec.go
Sudhar287: `c.httpResp` is null in the test case (`storage/remote/codec_test.go`) where the object is initialized as `ss := &chunkedResponseSeriesSet{stream: r}`

Sudhar287: [This](https://github.com/prometheus/prometheus/blob/master/storage/remote/chunked.go#L108) about NextProro() states that the error type can either be `io.EOF` or `io.ErrUnexpectedEOF`  Relating to your comment on `SeriesSet.Err`, do you mean something like this?
```
err := c.stream.NextProto(res)
if err!= nil {
     c.nextChunkError = err
}
```
and then
```
func (c *chunkedResponseSeriesSet) Err() error {
	return c.nextChunkError
}
```

The above pattern breaks some test cases in `TestEndpoints -v ./web/api/v1/` I'm investigating it....

Sudhar287 (commented):
Hey Bartek, I have some small queries. Can you please check them out when you're available? :)

storage/remote/codec.go
bwplotka: Hey the main problem with current code is that when you have error from ` c.stream.NextProto(res)` it's never handled. That's a problem. It always return true.

I would definitely except something like you propose in https://github.com/prometheus/prometheus/pull/7354/files#r487503100 just `c.err` would be enough.

What you need though is: 

```
if err := c.stream.NextProto(res); err!= nil {
     if err != io.EOF{
         c.err = err
     }
     return false
}
```

bwplotka: > This link about NextProto() states that the error type can either be io.EOF or io.ErrUnexpectedEOF 

This does not matter what errors the current implementation can return. In next commit someone can return some other error (:

bwplotka (commented):
Nice, responded for missed comment, sorry for missing it.

storage/remote/client.go
Sudhar287: Hmm, I considered the above option but won't it be incorrect? 
for eg: We're going to return `return &chunkedResponseSeriesSet{stream: stream, httpResp: httpResp}, nil` in `StartRead()` and thus if we have 
```	
defer func() {
		io.Copy(ioutil.Discard, httpResp.Body)
		httpResp.Body.Close()
        }()
```
before the `return &chunkedResponseSeriesSet{stream: stream, httpResp: httpResp}, nil` we will end up closing the `http` connection that is needed for `stream.NextProto(res)`. 
From my understanding, the most straightforward option seems to be to close the `HttpRest` individually/separately depending on the use case.

storage/remote/codec.go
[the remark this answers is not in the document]
Sudhar287: Done.
The return err is used in
```
func (it *chunkSeriesIterator) Err() error {
	return it.err
}
```

storage/remote/client.go
Sudhar287: Done.

storage/remote/codec.go
Sudhar287: Done. Please check

storage/remote/codec.go
Sudhar287: Oops, I misunderstood this comment and removed the return statement altogether which caused some more confusion on 'ignoring comments/review' below. Sorry about that.
It's been fixed!

storage/remote/codec_test.go
Sudhar287: fyi, it has been implemented in one of the previous commits. 
PS: A good practice on my end would be to start commenting 'done' as and when I solve the issues. Haha

tsdb/tsdbutil/buffer.go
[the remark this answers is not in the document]
Sudhar287: You are right, we dont need this.
Have fixed this by creating a 
```
type TestSample struct {
	t int64
	v float64
}
```
in `codec_test.go`

storage/remote/client.go
```diff
@@ -144,13 +144,17 @@ func (c Client) Endpoint() string {
 }
 
 // Read reads from a remote endpoint.
-func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryResult, error) {
+func (c *Client) Read(ctx context.Context, query *prompb.Query) (*http.Response, error) {
```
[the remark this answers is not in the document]
Sudhar287: Done in one of the previous commits.

storage/remote/codec_test.go
```diff
@@ -210,3 +212,58 @@ func TestNegotiateResponseType(t *testing.T) {
...
+		_, err := w.Write(msg)
+		testutil.Ok(t, err)
+	}
+	r := NewChunkedReader(b, 5e+7, nil)
+	ss := &chunkedResponseSeriesSet{stream: r}
+	ss.Next()
+	// I want to write test for the iterator; something similar to what is there below.
+	// But it dosent work becuase the types are not the same. Its also difficult for me to understand
+	// wahts happening becuase of all the protobufs, marshalling and encoding.
+	// How should I proceed with writing tests for the iterator while its remote reading?
```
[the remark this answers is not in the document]
Sudhar287: Done in one of the previous commits.

storage/remote/codec.go
Sudhar287: Yes. Thanks for the help/guidance Bartek.

storage/remote/client.go
bwplotka: well there is better way. You are right that we cannot do it for the case when we return iterator. So in this case I would do this.:

```
defer func() {
            if err != nil {
		io.Copy(ioutil.Discard, httpResp.Body)
		httpResp.Body.Close()
              }
        }()
```
Then make sure the function returns `(<something>, err error)` and it will only close on error (:

storage/remote/client.go
bwplotka: Suggested something here: https://github.com/prometheus/prometheus/pull/7354#discussion_r494860692

storage/remote/client.go
```diff
@@ -257,22 +262,39 @@ func (c *Client) Read(ctx context.Context, query *prompb.Query) (*prompb.QueryRe
...
+		httpResp.Body.Close()
+		return nil, errors.Errorf("remote server %s returned HTTP status %s: %s", c.url.String(), httpResp.Status, strings.TrimSpace(string(compressed)))
+	}
+
+	if !strings.HasPrefix(contentType, "application/x-streamed-protobuf; proto=prometheus.ChunkedReadResponse") {
+		httpResp.Body.Close()
+		return nil, errors.Errorf("not supported remote read content type: %s", contentType)
+	}
+
+	// Handling streamed response
```
bwplotka, line 281:
```suggestion
	// Handling streamed response.
```

storage/remote/codec_test.go
```diff
@@ -229,3 +234,160 @@ func TestMergeLabels(t *testing.T) {
 		testutil.Equals(t, tc.expected, MergeLabels(tc.primary, tc.secondary))
 	}
 }
+
+type TestSample struct {
```
bwplotka, line 238:
```suggestion
type sample struct {
```

bwplotka (commented):
Nice, just few comments and LGTM! (: Thanks for patience :+1:

bwplotka:
Can you also resolve comments which are addressed to keep this PR discussion clear?

bwplotka:
Tests are definitely failing because of this PR. I can see the endpoint tests are failing with `context.Cancelled`. Maybe we close connection sooner than we should?

storage/remote/client.go
Sudhar287: This makes sense to me. (I have a small clarifying question though)
But, I can see that won't be returning an error when we see an `io.EOF` in the `chunkedResponseSeriesSet` `Next()`. The connection needs to be closed in this case too, right?
Should the `Next()` be replaced from 
```
if err := c.stream.NextProto(res); err != nil {
		if err != io.EOF {
			c.err = err
		}
		return false
	}
```
Replaced to 
```
if err := c.stream.NextProto(res); err != nil {
		c.err = err
		return false
	}
```
This above approach produces some errors like:
`api_test.go:1784: Unexpected error: execution: EOF`
Or is this the best version:
```
if err := c.stream.NextProto(res); err != nil {
                httpResp.Body.Close()		
                if err != io.EOF {
			c.err = err
		}
		return false
	}
```

storage/remote/client.go
Sudhar287: Done

storage/remote/codec_test.go
Sudhar287: This is not possible as it collides with this https://github.com/prometheus/prometheus/blob/90fc6be70f5c11aabcc25a86b5e0bd22a15ae832/storage/remote/queue_manager.go#L650
Anyway, I have replaced it to `testSample`

tomwilkie:
This has been idle for a couple of months now, and @wwformat wants to continue work on it in #8351 - I suggest we close this and continue there?

roidelapluie:

This is now a duplicate of #11379.

Issue #7326, opened 2020-06-02
PromQL: Improve space complexity by streaming more; avoid series pre-lookup.


bwplotka:
Hi :wave: 

While changing [SeriesSet interfaces](https://github.com/prometheus/prometheus/pull/7251/), we found with @brian-brazil that our usage of SerieSet in PromQL might cause extra unnecessary allocations for certain implementations and cases. Let me describe the problem in two diagrams:

## Current PromQL logic for Range Query with Metric Selection:

![Prometheus PromQL v2 19 Flow](https://user-images.githubusercontent.com/6950331/83543350-2c464680-a4f4-11ea-9d90-8d6efc457f81.png)
_[Google Draw Link](https://docs.google.com/drawings/d/1NfV8Gp53afJY3CfGwn3L1_-dFImA6f2QvsUMDg4x0Bs/edit)_

All function names should match the current master: https://github.com/prometheus/prometheus/tree/18d9ebf0ffc26b8bd0e136f552c8e9886d29ade4

## Issues:

![Prometheus PromQL v2 19 Flow_ Unnecessary PreLookup of Series (1)](https://user-images.githubusercontent.com/6950331/83545333-f787be80-a4f6-11ea-8d4e-cc4aa99038e0.png)
_[Google Draw Link](https://docs.google.com/drawings/d/1Ice50aJRN7xgRIQKe-T3OM4uDLjw3dGeshVjzHT22X8/edit)_

**Problem**: We are accumulating all series first, to gather all iterators.

With the current interfaces (StoreSet, Remote Read, Thanos StoreAPI) “Series by Series” means that we stream Series labels together with labels. 

**Current Flow Issues:**

1. To finish this `eval` PromQL stage, we have to buffer & receive all data. Especially for remote data & Thanos, this means literally all. (: **Realistic space complexity is now O(2 * series * chunks), for the biggest result, so ~ O(2 series * chunks) B**
2. Current space complexity is also not great further with multi arg functions, as we build a hash map. For large cardinality queries, we end up hashing millions of Series potentially.

## Solution: Stream more!

The idea would be to follow really exactly the same algorithm as everywhere else (vertical dedup, merge series etc): Implement Heap / MergeSort that will iterate over "sorted" series from two SeriesSets. 

[...]

PR #7361, opened 2020-06-08
Fixed incorrect query results caused by buffer reuse in merge adapter

### LINKED ISSUE #7359: Merger adapter causes invalid query results due to slice reuse
**What did you do?**

I'm investigating an issue in Cortex blocks storage (running Prometheus `master`) and I ended up realising that the issue is caused by `seriesMergerAdapter`'s buffer.

The problem is that `seriesMergerAdapter` reuse the same slice to cast the `Series` and then this slice is passed to `ChainedSeriesMerge` which retains the slice, so subsequent calls to `seriesMergerAdapter.Merge()` will overwrite the (unintentionally) shared slice:
https://github.com/prometheus/prometheus/blob/e5ffa8c9a08a5ee4185271c8c26051ddc1388b7a/storage/generic.go#L114-L120

_I'm still checking the impact on Prometheus, but I preferred to open the issue earlier to raise a flag._

pracucci:
Fixes #7359.

### DIFF
--- a/storage/fanout_test.go
+++ b/storage/fanout_test.go
@@ -195,10 +195,18 @@ func TestMergeQuerierWithChainMerger(t *testing.T) {
 			}
 			qs = append(qs, tc.extraQueriers...)
 
-			merged, _, _ := NewMergeQuerier(qs[0], qs, ChainedSeriesMerge).Select(false, nil)
-			for merged.Next() {
+			mergedQuerier, _, _ := NewMergeQuerier(qs[0], qs, ChainedSeriesMerge).Select(false, nil)
+
+			// Get all merged series upfront to make sure there are no incorrectly retained shared
+			// buffers causing bugs.
+			var mergedSeries []Series
+			for mergedQuerier.Next() {
+				mergedSeries = append(mergedSeries, mergedQuerier.At())
+			}
+			testutil.Ok(t, mergedQuerier.Err())
+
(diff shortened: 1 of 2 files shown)


storage/fanout_test.go
```diff
@@ -195,10 +195,18 @@ func TestMergeQuerierWithChainMerger(t *testing.T) {
...
 
-			merged, _, _ := NewMergeQuerier(qs[0], qs, ChainedSeriesMerge).Select(false, nil)
-			for merged.Next() {
+			mergedQuerier, _, _ := NewMergeQuerier(qs[0], qs, ChainedSeriesMerge).Select(false, nil)
+
+			// Get all merged series upfront to make sure there are no incorrectly retained shared
+			// buffers causing bugs.
+			var mergedSeries []Series
+			for mergedQuerier.Next() {
+				mergedSeries = append(mergedSeries, mergedQuerier.At())
```
bwplotka, line 204: Eh, again we are hit but "broken" usage of PromQL. Hopefully, we can fix it here: https://github.com/prometheus/prometheus/issues/7326 

This is good hot fix for now :muscle:

bwplotka:
Once merged I am happy to cherry-pick for v.2.

bwplotka:
https://github.com/prometheus/prometheus/pull/7364

PR #7364, opened 2020-06-09
Cut 2.18.2 release with query bug fix.

bwplotka:
Hugo kudos to @pracucci for quickly fixing the problem introduced in 2.18. Thanks!


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.18.2 / 2020-06-09
+
+* [BUGFIX] TSDB: Fixed incorrect fanout series merge causing incorrect query results in certain cases. #7361
```
roidelapluie, line 3: That is obscure for end users.

Suggestion:

TSDB: Fix corruption in query results when using remote read

bwplotka: This is not even true, it has nothing to do to remote read.

bwplotka: In fact remote read was not affected

roidelapluie: In practice it is with remote read that we merge series in the Prometheus server. Is there another case?

bwplotka: Happy to change, what is not obscure but also true?

roidelapluie: I used the words "when using remote read" not "with remote read". The two bug reports from users mentioned remote read issues.

bwplotka: sorry I misinterpret you, you are right actually from Prometheus user standpoint that's the only time it can be triggered :facepalm:

bwplotka: >  when using remote read

Means for me when you call `/api/v1/read` or Prometheus that's why I was confused, will rephrase

bwplotka:
You were right, I rephrased this to:

> * [BUGFIX] TSDB: Fix incorrect query results when using Prometheus with remote reads configured #7361

cc @codesome

brian-brazil:
:+1:

PR #7366, opened 2020-06-09
Cut v2.19.0

codesome:


codesome:
In the benchmarks, compaction time has increased and CPU usage is slightly higher. The compaction time increase can be explained by needing the load the mapped chunks to compact them. I suspect CPU usage increase is also due to regular access of m-mapped chunks. 

Here is the link to the dashboard: http://prombench.prometheus.io/grafana/d/7gmLoNDmz/prombench?orgId=1&var-pr-number=7341&from=1591291893377&to=1591543292355

brian-brazil:
:+1: 

The CPU difference is tiny.

PR #7369, opened 2020-06-09
Pulled 2.18.2 to master

bwplotka:
This also backports @pracucci fix ( https://github.com/prometheus/prometheus/pull/7361) to master.

Sorry with mess with revert due to squashing on first try. This PR has to be non-squashed as well.


brian-brazil:
:+1:

PR #7375, opened 2020-06-10
Add some unit tests

hs0210:
This PR is about adding a unit test for for pkg/labels/labels.go and a unit test for storage/buffer.go.

### DIFF
--- a/storage/buffer_test.go
+++ b/storage/buffer_test.go
@@ -161,6 +161,43 @@ func TestBufferedSeriesIteratorNoBadAt(t *testing.T) {
 	it.Next()
 }
 
+func TestBufferReduceDelta(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pkg/labels/labels_test.go
```diff
@@ -612,3 +612,7 @@ func TestBuilder_Labels(t *testing.T) {
 		}).Labels(),
 	)
 }
+
+func TestLabels_Bytes(t *testing.T) {
```
brian-brazil, line 616: This is over-testing, we don't actually care about the exact bytes this returns. PromQL tests already cover this.

hs0210: OK. I have removed this unit test.

hs0210:
PTAL.

PR #7380, opened 2020-06-10
storage/remote/read: Implement concurrent select for remote read

kakkoyun:
Signed-off-by: Kemal Akkoyun <kakkoyun@gmail.com>

This implements concurrent selects for remote read utilizing async selects, ref: https://github.com/prometheus/prometheus/pull/7251

Completes https://github.com/prometheus/prometheus/issues/6878

### DIFF
--- a/storage/remote/read.go
+++ b/storage/remote/read.go
@@ -141,37 +141,54 @@ type querier struct {
 // If requiredMatchers are given, select returns a NoopSeriesSet if the given matchers don't match the label set of the
 // requiredMatchers. Otherwise it'll just call remote endpoint.
 func (q *querier) Select(sortSeries bool, hints *storage.SelectHints, matchers ...*labels.Matcher) storage.SeriesSet {
-	if len(q.requiredMatchers) > 0 {
-		// Copy to not modify slice configured by user.
-		requiredMatchers := append([]*labels.Matcher{}, q.requiredMatchers...)
-		for _, m := range matchers {
-			for i, r := range requiredMatchers {
-				if m.Type == labels.MatchEqual && m.Name == r.Name && m.Value == r.Value {
-					// Requirement matched.
-					requiredMatchers = append(requiredMatchers[:i], requiredMatchers[i+1:]...)
+	promise := make(chan storage.SeriesSet, 1)
+	go func() {
+		defer close(promise)
+
+		if len(q.requiredMatchers) > 0 {
+			// Copy to not modify slice configured by user.
+			requiredMatchers := append([]*labels.Matcher{}, q.requiredMatchers...)
+			for _, m := range matchers {
+				for i, r := range requiredMatchers {
+					if m.Type == labels.MatchEqual && m.Name == r.Name && m.Value == r.Value {
+						// Requirement matched.
+						requiredMatchers = append(requiredMatchers[:i], requiredMatchers[i+1:]...)
+						break
+					}
+				}
+				if len(requiredMatchers) == 0 {
 					break
 				}
 			}
-			if len(requiredMatchers) == 0 {
-				break
+			if len(requiredMatchers) > 0 {
+				promise <- storage.NoopSeriesSet()
+				return
 			}
 		}
-		if len(requiredMatchers) > 0 {
-			return storage.NoopSeriesSet()
+
+		m, added := q.addExternalLabels(matchers)
+		query, err := ToQuery(q.mint, q.maxt, m, hints)
+		if err != nil {
+			promise <- storage.ErrSeriesSet(errors.Wrap(err, "to_query"))
+			return
 		}
-	}
 
-	m, added := q.addExternalLabels(matchers)
-	query, err := ToQuery(q.mint, q.maxt, m, hints)
-	if err != nil {
-		return storage.ErrSeriesSet(errors.Wrap(err, "toQuery"))
-	}
+		res, err := q.client.Read(q.ctx, query)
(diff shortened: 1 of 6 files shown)


storage/remote/read.go
```diff
@@ -47,6 +50,8 @@ func init() {
 	prometheus.MustRegister(remoteReadQueriesTotal)
 }
 
+const defaultMaxConcurrentSelect = 4
```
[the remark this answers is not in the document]
kakkoyun: I thought we don't want to spawn unbounded goroutines for querying. as far as I can recall from the discussion we had in https://github.com/prometheus/prometheus/pull/7251.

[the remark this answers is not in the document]
brian-brazil: Using more than one core inside Prometheus would be bad, but Select from remote servers is not the same thing and it's very cheap to send out the requests.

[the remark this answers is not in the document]
kakkoyun: Got it, I'm gonna remove the limit.

storage/remote/read.go
```diff
@@ -278,3 +294,40 @@ func (sf seriesFilter) Labels() labels.Labels {
 	}
 	return labels
 }
+
+type lazySeriesSet struct {
```
brian-brazil, line 298: Don't we already have one of these?

kakkoyun: What we have implements `genericSeriesSet`, this one implements `storage.SeriesSet`.

brian-brazil: Why do we need two mildly different variants? I thought the whole purpose of all the generic/adapter complexity was to avoid duplication.

kakkoyun: I don't know the exact design decisions behind generic/adapter approach. 
But sure, same the behaviour could be achieved by exposing generic and adapter types and use it in here.
```go
&storage.SeriesSetAdapter{
		&storage.LazySeriesSet{create: func() (storage.GenericSeriesSet, bool) {
			set, ok := <-promise
			if !ok {
				return storage.SeriesSetAdapter{storage.ErrSeriesSet(errors.New("channel closed before a value received"))}, false
			}
			s := storage.SeriesSetAdapter{set}
			return s, s.Next()
		}}
```
However I don't think this is worth exposing all those types.

brian-brazil: @bwplotka How is this meant to work?

kakkoyun: Another option to keep a single `LazySeriesSet` and use `genericSeriesSetAdapter` to wrap it and use it in fanout.

roidelapluie:
How does this compare with #6770 ?

kakkoyun:
> How does this compare with #6770 ?

@roidelapluie In #6770, we send select requests concurrently to different storages (local tsdb, remote storage). However we wait for their responses and merge them so the concurrency is limited. With https://github.com/prometheus/prometheus/pull/7251, we have introduced an ability to defer selects until the first call of the iterators. This PR builds upon #7251, so that we can send concurrent request against remote storage.

I'll reflect the latest interface changes and update this PR when I've some spare time.

krasi-georgiev (commented):
I also think we should try and reuse the existing structs so let's wait for Bartek and see what was his idea here.

storage/remote/read.go
krasi-georgiev: ok I had another look and don't think we can avoid a bit of duplication, because of the `init` function signature, but don't need the Err and Warning methods to can be like:

```
func LazySeriesSet(init func() (SeriesSet, bool)) SeriesSet {
	return &lazySeriesSet{init: init}
}

// lazySeriesSet is a wrapped series set that is initialised on first call to Next().
type lazySeriesSet struct {
	lazyGenericSeriesSet
	init func() (s SeriesSet, ok bool)

	set SeriesSet
}

func (c *lazySeriesSet) Next() bool {
	if c.set != nil {
		return c.set.Next()
	}

	var ok bool
	c.set, ok = c.init()
	return ok
}

func (c *lazySeriesSet) At() Series {
	if c.set != nil {
		return c.set.At()
	}
	return nil
}
```

storage/remote/read.go
```diff
@@ -140,37 +142,60 @@ type querier struct {
...
+
+		promise <- newSeriesSetFilter(FromQueryResult(sortSeries, res), added)
+	}()
+
+	return storage.LazySeriesSet(func() (storage.SeriesSet, bool) {
+		set, ok := <-promise
+		if !ok {
+			return storage.ErrSeriesSet(errors.New("channel closed before a value received")), false
+		}
+		return set, set.Next()
```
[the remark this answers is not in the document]
kakkoyun: This is kind of the expectation, this call is deferred. This is called when iterator received the first `Next` called.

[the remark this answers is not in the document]
krasi-georgiev: ooh yeah I forgot that this gets executed on the first Next.

bwplotka:
Need to get back to this at some point (: On my TODO list.

kakkoyun:
> Need to get back to this at some point (: On my TODO list.

It must be a long To-Do list :stuck_out_tongue_winking_eye: Shall we revive this or close it?

roidelapluie:
last call from the bug scrub to @bwplotka , what should we do with this PR? Thanks!

bwplotka:
Thanks for pinging, looks like it got out of our attention for some time and might be less relevant, let's close it. We can always reopen with @kakkoyun

PR #7381, opened 2020-06-10
Fix #7147

### LINKED ISSUE #7147: Expose total number of label pairs in head

nevill:
Add a new metric prometheus_tsdb_head_label_pairs, to count the unique number lable pairs in the head block. This will close #7147.

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -116,6 +116,7 @@ type headMetrics struct {
 	checkpointCreationFail   prometheus.Counter
 	checkpointCreationTotal  prometheus.Counter
 	mmapChunkCorruptionTotal prometheus.Counter
+	labelPairs               prometheus.Gauge
 }
 
 func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
@@ -206,6 +207,10 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 			Name: "prometheus_tsdb_mmap_chunk_corruptions_total",
 			Help: "Total number of memory-mapped chunk corruptions.",
 		}),
+		labelPairs: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "prometheus_tsdb_head_label_pairs",
+			Help: "Total number of unique label pairs in the head block.",
+		}),
 	}
 
 	if r != nil {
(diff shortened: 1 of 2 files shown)


nevill:
@brian-brazil can you review it?

brian-brazil:
@codesome is the tsdb maintainer.

nevill:
Sorry @brian-brazil :)
Hi @codesome can you take a review?

tsdb/head.go
```diff
@@ -1327,7 +1334,13 @@ func (h *Head) gc() {
...
 			values[t.Name] = ss
 		}
-		ss.set(t.Value)
+
+		if _, ok := ss[t.Value]; !ok {
+			if t.Name != "" {
+				numLabelPairs++
+			}
+			ss.set(t.Value)
+		}
```
Harkishen-Singh, line 1343: Not sure, but I think `getOrCreateWithID()` should be sufficient for this.

codesome (commented):
Apologies for the super delayed review. Would you be able to make some changes?

Currently the implementation has some race. It would be much simpler if we keep the track of number of label pairs in `MemPostings` struct hence update it in `postings.Delete()` and `postings.Add()` call instead of tracking it in Head.

tsdb/head.go
codesome: Only `getOrCreateWithID()` is not enough since it only accounts for addition of new label pairs and not deletion.

nevill:
@codesome OK, glad to have some feedback, but I really need to refresh my mind to follow some recent changes of the project first.

ankitpatel96:
bump - this would be a reallly really nice feature for us!

PR #7393, opened 2020-06-14
Fix off-by-one error in funcHistogramQuantile / ensureMonotonic

linasm:
Signed-off-by: Linas Medziunas <linas.medziunas@gmail.com>

Fixes an off-by-one error in `ensureMonotonic` (that is used in histogram quantiles calculation):
- original implementation was unintentionally iterating from `0` to `len(buckets) - 2`, thus not fixing the last bucket value
- fixed implementation iterates from `1` to `len(buckets) - 1` as intended

### DIFF
--- a/promql/quantile.go
+++ b/promql/quantile.go
@@ -168,7 +168,7 @@ func coalesceBuckets(buckets buckets) buckets {
 
 func ensureMonotonic(buckets buckets) {
 	max := buckets[0].count
-	for i := range buckets[1:] {
+	for i := 1; i < len(buckets); i++ {
 		switch {
 		case buckets[i].count > max:
 			max = buckets[i].count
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thanks, could we have tests around each of the non-monotonic buckets?

roidelapluie:
Working on tsdb test failure at https://github.com/prometheus/prometheus/pull/7394

roidelapluie:
When you add the tests Brian asked, would you please rebase on top of master? Thanks!

linasm:
@brian-brazil it is not possible to hit the non-monotonic buckets with binary search directly (because their value is made equal to the value of the preceding bucket). So I have altered the test data to alternate the buckets (increase -> decrease -> increase -> ...) and added test queries to hit every increase -> decrease pair.
Let me know if this is similar to what you were asking for (I'm not 100% sure).

@roidelapluie thanks for fixing the master so quickly.

brian-brazil:
Thanks!

Issue #7396, opened 2020-06-15
Configurable minValidTime in the head block


codesome:
## Introduction 

Head block has a concept of `minValidTime` wherein it won't allow samples before 1h to be ingested (Called out of bound samples)

## Proposal

Here I propose making `minValidTime` configurable which is solely for external users of TSDB and does not affect the functioning of Prometheus.

Sometimes we want to ingest some old data because of long outages lasting >1h. In such cases, the `minValidTime` is a blocker. Making it configurable will allow ingesting old data.

Out of order samples are still discarded.

## Changes required

**To configure `minValidTime`**

There are 2 possible ways we could make it configurable.

1. Introduce a method `SetMinValidTime()` which will set it to the value passed. But this one is easy to make mistakes.

2. Pass a config, let's say `MinValidTimeGracePeriod`, which will accept samples up to `MinValidTimeGracePeriod` back in time from the current `minValidTime` that we calculate.

**To the compaction**

Allowing old samples can interfere with head compaction. So the compaction check for Head can be changed 
from `head.maxt - head.mint >= chunkRange * 3/2`
to `if len(blocks) > 0 { head.maxt - lastBlock.maxt >= chunkRange * 3/2 } else { old condition }`

Vertical compaction and vertical queries needs to be enabled to take care of overlapping data.

## Is it safe

Memory: It is proportional to the number of series. With m-mapping in place, taking in lots of old samples is not a problem.

Data: The vertical query and vertical compaction take care of everything. The user must be aware of what they are getting into.

bwplotka:
cc @JessicaGreben

brian-brazil:
Uhm, don't we already have this setting via the min block size?

bwplotka:
Vertical compaction would handle those.

codesome:
> Uhm, don't we already have this setting via the min block size?

The idea is to keep the block size same as before and allow old samples in case needed in certain scenarios.

roidelapluie:
That sounds amazing for #535 too

bwplotka:
Yes, this is an amazing step forward in to backfillable Remote Write :+1:

codesome:
This might not help much for #535 as we need to align the data with existing blocks to avoid large blocks. But backfillable Remote Write, definitely.

While using configurable minValidTime, the compaction planner needs to be a little different else it is possible to create very large blocks with carefully planned samples. Hence I don't see this being exposed to users of Prometheus.

roidelapluie:
[...]

PR #7398, opened 2020-06-15
Configurable minValidTime for TSDB

codesome:
Closes https://github.com/prometheus/prometheus/issues/7396

Stuff that this PR adds/changes:
1. `HeadOptions` for the head block to cut down on arguments for `NewHead()`
2. `SetMinValidTime` method for Head.
3. `MinValidTimeGracePeriod` as described in the linked issue. 
4. Updates the head compaction logic.
5. Adds `prometheus_tsdb_head_compactions_triggered_total` to know exactly how many times the head compaction was triggered.

TODO:
- [x] Unit tests
- [ ] Test it in Cortex to find any issues

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -725,7 +725,7 @@ func (a dbAppender) Commit() error {
 
 	// We could just run this check every few minutes practically. But for benchmarks
 	// and high frequency use cases this is the safer way.
-	if a.db.head.compactable() {
+	if a.db.isHeadCompactable() {
 		select {
 		case a.db.compactc <- struct{}{}:
 		default:
@@ -757,7 +757,7 @@ func (db *DB) Compact() (err error) {
 			return nil
 		default:
 		}
-		if !db.head.compactable() {
+		if !db.isHeadCompactable() {
 			break
 		}
 		mint := db.head.MinTime()
@@ -777,6 +777,21 @@ func (db *DB) Compact() (err error) {
 	}
 
 	return db.compactBlocks()
+}
+
+// isHeadCompactable returns whether the head has a compactable range.
+// The head has a compactable range when the head time range is 1.5 times the chunk range.
+// The 0.5 acts as a buffer of the appendable window.
+// This takes into account the grace period for min valid time and checks the range
+// from the maxt of the last block if any block exist.
+func (db *DB) isHeadCompactable() bool {
+	compactable := db.head.MaxTime()-db.head.MinTime() > db.head.opts.ChunkRange/2*3
+	if blocks := db.Blocks(); len(blocks) > 0 {
+		// We do this extra check with blocks to not compact because of the grace period
+		// allowed for the min valid time.
+		compactable = compactable && db.head.MaxTime()-blocks[len(blocks)-1].MaxTime() > db.head.opts.ChunkRange/2*3
+	}
+	return compactable
 }
 
 // CompactHead compacts the given the RangeHead.
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -118,6 +118,10 @@ type Options struct {
 	// SeriesLifecycleCallback specifies a list of callbacks that will be called during a lifecycle of a series.
 	// It is always a no-op in Prometheus and mainly meant for external users who import TSDB.
 	SeriesLifecycleCallback SeriesLifecycleCallback
+
+	// MinValidTimeGracePeriod is the additional grace period w.r.t the current min valid time
+	// for samples to be appended before they are considered out of bound.
+	MinValidTimeGracePeriod int64
 }
 
 // DB handles reads and writes of time series falling into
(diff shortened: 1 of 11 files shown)


tsdb/db.go
```diff
@@ -118,6 +118,10 @@ type Options struct {
 	// SeriesLifecycleCallback specifies a list of callbacks that will be called during a lifecycle of a series.
 	// It is always a no-op in Prometheus and mainly meant for external users who import TSDB.
 	SeriesLifecycleCallback SeriesLifecycleCallback
+
+	// MinValidTimeGracePeriod is the additional grace period w.r.t the current min valid time
```
brian-brazil, line 122: Why not have it as a hard value? Having the value consistently set to 1h for Prometheus would probably avoid some confusion.

codesome: If with a hard value you mean a constant, there might not be a single value that satisfies all use cases for external users, so I think configurable is better. And I was thinking of this being 0 for Prometheus for now as we have a flag to change the block size (though hidden), and setting this to a fixed size means the data is no longer assured to be non-overlapping.

brian-brazil: I mean that rather as relative to whatever the current min block time, have it relative to the max block time.

brian-brazil: I'm also strongly opposed to this being configurable for Prometheus - we already have enough problems with users messing with the min block size and trying to use Prometheus for push without forcing Cortex design decisions onto Prometheus.

codesome: Do you mean to say take in a multiplier and use min/max block time on that to get the grace period?

Having it >0 for Prometheus means that we need to enable vertical compaction always, so let it be 0 for Prometheus. This option is only exposed in the TSDB package but nowhere configurable from the Prometheus binary.

brian-brazil: I'm suggesting that rather than the head taking in 1-3h old data in Prometheus depending on where it is in the compaction cycle, that it'd always be a consistent 1h.

codesome: I miss spoke in the issue description (fixed now). It is actually 1h old consistently for Prometheus. So with this grace period, it will be `1h + grace period` consistently.

The relevant piece of code which maintains that 1h: https://github.com/prometheus/prometheus/blob/7b24bb311631f4eaa5d794b849770f436fc75b5a/tsdb/head.go#L1015

brian-brazil: Ah, I didn't realise that.

In that case does that do the right thing in terms of blocks/WAL given that this minValidTime isn't tied to them?

codesome: What do you mean by the right thing here? If you mean `1h + grace period` old for samples, then yes, it will be that. And during WAL replay, as we use the time passed in the `Head.Init` as the min valid time, we just allow upto that `mint - grace period` for samples.

brian-brazil: If the source data is coming from a Prometheus's WAL then that will be aligned with the 2h of the compaction cycle. This PR as proposed however is aligned with the current time. So there's a mismatch there.

codesome: Nice catch. IIUC we want the out of bounds to be always 1h consistent. So I will fix it during replay to consider that.

codesome: (after the open discussions are concluded in https://github.com/prometheus/prometheus/issues/7396)

brian-brazil: Hmm, that doesn't sound right if your goal is to handle delayed data from remote write as that's not 1h delayed - it goes back to the start of the head which is on the 2h cycle.

codesome: I might be confusing things here. Here is what I understand right now:

1. Configurable grace period and not a fixed grace period. So, samples can overlap blocks too.
2. During an append, currently, it's fixed to max 1h old sample w.r.t. current maxt. So, this makes it `1h + grace period`
3. During WAL replay, we take min valid time as the maxt of the last block throughout the replay. So, while applying the grace period, we should take the min valid time w.r.t. the maxt of head block so that we don't end up with samples which are already in the block.

brian-brazil: That all sounds right, but that's not what I'm talking about.

Consider you have a normal Prometheus using remote write, which has remote write delayed for some reason and then restarts. It will start re-sending samples from the WAL based on where the head is.

Now that's going to Cortex with all this code, will that Cortex be able to ingest all the samples from that Prometheus's remote write and if so how much grace does it need to do so?

codesome: > Now that's going to Cortex with all this code, will that Cortex be able to ingest all the samples from that Prometheus's remote write and if so how much grace does it need to do so?

Cortex will have a limit too on how old samples can it take and can't get away from it. The grace period will be a decision that needs to be taken on how long do you want to allow. This configuration would just give some extra time to fix things and wont completely eliminate losing data.

codesome:
https://github.com/prometheus/prometheus/issues/7396#issuecomment-649451086

PR #7407, opened 2020-06-16
Add DigitalOcean service discovery

### LINKED ISSUE #4166: digitalocean service discovery
## Proposal

I want to use prometheus on DigitalOcean. My previous experience was on aws, using ec2_sd and all that.

I noticed, though, that there isn't a digitalocean_sd or something like that for digitalocean.

I could in theory use SD from files, etcd or whatever... but I think it would be useful for more people.

Is that from the interest of you guys? If it is, I can work on a PR.

Thanks!

roidelapluie:
Fixes #4166


roidelapluie:
See #4166

discovery/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,243 @@
...
+type SDConfig struct {
+	AccessToken     config_util.Secret `yaml:"access_token,omitempty"`
+	AccessTokenFile string             `yaml:"access_token_file,omitempty"`
+
+	// A tag that can be used to filter the droplets.
+	Tag string `yaml:"tag,omitempty"`
+
+	RefreshInterval model.Duration `yaml:"refresh_interval,omitempty"`
+	Port            int            `yaml:"port"`
+	Separator       string         `yaml:"separator,omitempty"`
```
brian-brazil, line 66: Why is this configurable?

discovery/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,243 @@
...
+	Separator:       ",",
+	RefreshInterval: model.Duration(60 * time.Second),
+}
+
+// SDConfig is the configuration for DO based service discovery.
+type SDConfig struct {
+	AccessToken     config_util.Secret `yaml:"access_token,omitempty"`
+	AccessTokenFile string             `yaml:"access_token_file,omitempty"`
+
+	// A tag that can be used to filter the droplets.
```
brian-brazil, line 61: Is this needed for performance?

discovery/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,243 @@
...
+		labels[model.AddressLabel] = model.LabelValue(addr)
+
+		if droplet.Features != nil && len(droplet.Features) > 0 {
+			// We surround the separated list with the separator as well. This way regular expressions
+			// in relabeling rules don't have to consider features positions.
+			features := d.separator + strings.Join(droplet.Features, d.separator) + d.separator
+			labels[doLabelFeatures] = model.LabelValue(features)
+		}
+
+		if droplet.Tags != nil && len(droplet.Tags) > 0 {
```
brian-brazil, line 200: Shouldn't the len be sufficient?

discovery/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,243 @@
...
+		time.Duration(conf.RefreshInterval),
+		d.refresh,
+	)
+	return d, nil
+}
+
+func (d *Discovery) refresh(ctx context.Context) ([]*targetgroup.Group, error) {
+	source := "DO"
+	if d.tag != "" {
+		source += "_" + d.tag
```
brian-brazil, line 142: Why do you do this?

discovery/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,243 @@
...
+			doLabelSize:        model.LabelValue(droplet.SizeSlug),
+			doLabelStatus:      model.LabelValue(droplet.Status),
+		}
+
+		var addr string
+		if privateIPv4 != "" {
+			addr = privateIPv4
+		} else if publicIPv4 != "" {
+			addr = publicIPv4
+		} else {
```
brian-brazil, line 187: Why not v6?

docs/configuration/configuration.md
```diff
@@ -386,6 +390,50 @@ users with thousands of services it can be more efficient to use the Consul API
...
+
+# The time after which the droplets are refreshed.
+# On large setup it might be a good idea to increase this value wrt DigitalOcean
+# rate limits.
+[ refresh_interval: <duration> | default = 60s ]
+```
+
+Note that the IP address used to scrape the targets is the private IPv4 or the
+public IPv4, in that order.
+If you want to use the IPv6, you can use the [relabel](#relabel_config)
```
brian-brazil, line 434: No need to document that relabelling exists.

discovery/digitalocean/digitalocean.go
roidelapluie: Indeed, tags can't have ,.

discovery/digitalocean/digitalocean.go
roidelapluie: No, and the service discovery is doing one API call per 20 droplets ; the digital ocean request limit is at 5000 / hours so we could in theory go up to 1660 droplets.

I think for simplicity we could remove this.

discovery/digitalocean/digitalocean.go
roidelapluie: Just as it made the tag made the source unique. Will remove as we will remove tag anyway.

discovery/digitalocean/digitalocean.go
roidelapluie: By default DigitalOcean provides IPv4 and it is not something you can disable. I prefer consistency.

discovery/digitalocean/digitalocean.go
roidelapluie: done

docs/configuration/configuration.md
roidelapluie: ack

roidelapluie:
Added tests.

discovery/digitalocean/digitalocean_test.go
```diff
@@ -0,0 +1,99 @@
...
+	testutil.Assert(t, tg.Targets != nil, "tg.targets should not be nil")
+	testutil.Equals(t, 2, len(tg.Targets))
+
+	for i, lbls := range []model.LabelSet{
+		{
+			"__address__":            model.LabelValue("104.236.32.182:80"),
+			"__meta_do_id":           model.LabelValue("3164444"),
+			"__meta_do_name":         model.LabelValue("example.com"),
+			"__meta_do_image":        model.LabelValue("ubuntu-16-04-x64"),
+			"__meta_do_private_ipv4": model.LabelValue(""),
```
roidelapluie, line 72: @brian-brazil is this OK or should I not add the empty label?

discovery/digitalocean/digitalocean.go
SuperQ: Yes, this should probably check for v6 first, and use v4 as a fallback.

roidelapluie: done

discovery/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,224 @@
...
+	config_util "github.com/prometheus/common/config"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/common/version"
+
+	"github.com/prometheus/prometheus/discovery/refresh"
+	"github.com/prometheus/prometheus/discovery/targetgroup"
+)
+
+const (
+	doLabel            = model.MetaLabelPrefix + "do_"
```
brian-brazil, line 35: We spell out kubernetes, so we should do the same here.

discovery/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,224 @@
...
+
+// DefaultSDConfig is the default DigitalOcean SD configuration.
+var DefaultSDConfig = SDConfig{
+	Port:            80,
+	RefreshInterval: model.Duration(60 * time.Second),
+}
+
+// SDConfig is the configuration for DO based service discovery.
+type SDConfig struct {
+	BearerToken     config_util.Secret `yaml:"bearer_token,omitempty"`
```
brian-brazil, line 58: Can we not use standard http config here, rather than partially reinventing it and losing things like tls/proxy_url?

discovery/digitalocean/digitalocean.go
brian-brazil: Hmm, is it safe to presume that if v6 is on a droplet that a Prometheus that could get to it via v4 could also get to it via v6?

discovery/digitalocean/digitalocean_test.go
brian-brazil: It should be harmless, if not that's a bug in the general SD code.

discovery/digitalocean/digitalocean.go
brian-brazil: This is still here.

discovery/digitalocean/digitalocean.go
roidelapluie: done

discovery/digitalocean/digitalocean.go
roidelapluie: done

discovery/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,195 @@
...
+
+		if publicIPv4 == "" {
+			return nil, fmt.Errorf("no suitable address for droplet %d: %w", droplet.ID, err)
+		}
+		addr := fmt.Sprintf("%s:%d", publicIPv4, d.port)
+		labels[model.AddressLabel] = model.LabelValue(addr)
+
+		if len(droplet.Features) > 0 {
+			// We surround the separated list with the separator as well. This way regular expressions
+			// in relabeling rules don't have to consider features positions.
```
brian-brazil, line 161: feature

docs/configuration/configuration.md
```diff
@@ -386,6 +390,60 @@ users with thousands of services it can be more efficient to use the Consul API
...
+# TLS configuration.
+tls_config:
+  [ <tls_config> ]
+
+
+# The port to scrape metrics from.
+[ port: <int> | default = 80 ]
+
+# The time after which the droplets are refreshed.
+# On large setup it might be a good idea to increase this value wrt DigitalOcean
```
brian-brazil, line 442: There's no need to explain how you might use it.

docs/configuration/configuration.md
```diff
@@ -386,6 +390,60 @@ users with thousands of services it can be more efficient to use the Consul API
...
+[ bearer_token_file: <filename> ]
+
+# Optional proxy URL.
+[ proxy_url: <string> ]
+
+# TLS configuration.
+tls_config:
+  [ <tls_config> ]
+
+
```
brian-brazil, line 437: extra blank line

SuperQ:
I still think that we should default to v6 if it's configured. Turning on v6 is still optional in DO, which means users that intentionally turn it on expect to be using it.

brian-brazil:
In SD the only time we don't pick a single value for the address is when that's simply impossible, namely when there's no source field that's always present. Otherwise we're introducing business logic, and that's what relabelling is for.

Public v4 is the only thing we know will be present, thus it is the right value to use. 

In addition that v6 is enabled on the scrape target does not mean that it will work for Prometheus, such as if it's running inside docker or the machine Prometheus is running on didn't have v6 enabled.

brian-brazil:
:+1:

roidelapluie:
ipv6 not only needs to be enabled on the DO level but also requires manual changes on VM ; so it is also less likely to work

SuperQ:
@roidelapluie Manual changes are only required if v6 is enabled after droplet creation. If you enable v6 at creation time, everything just works.

Basically, turning on v6 in DO is something that doesn't happen by accident.

Configuration problems outside of this are out of scope. We don't second guess other discovery methods that use DNS and get a AAAA result. If there is v6 on, we use it by default. This is how it's supposed to be.

roidelapluie:
Why prefer v4:
- all droplets have one by default
- it serves as a unique identifier
- it will work for most users

Why prefer private v4:
- to not use public address to route promethes traffic - which is something which would NOT change the way exporters work so that is not really relevant

Why prefer v6:
- best practice
- people who enable it probably know what they do and want to use it

issues with v6:
- Can be added when a droplet is created, breaking its unique id and all its metrics
- Prometheus in DO would have a v4 but maybe not a v6
- It would add small logic in the SD that might surprise/break users

middle ground:

- default to v4 for this provider because that is how the cloud provider is designed (always a v4 is easy to remember and explain)
- not making this a strict rule for all service discoveries
- document how to use v6 or private addresses (like we do for marathon: provide an exemple prometheus.yml)

brian-brazil:
> people who enable it probably know what they do and want to use it

I think that's a bit of an assumption. I have v6 enabled on some of my droplets but not others, and considering I don't have any DNS pointing at the v6 address I'd not expect anything to try to use it.

SuperQ:
@roidelapluie Thanks for documenting the reasoning. I think this is a good enough compromise.

documentation/examples/prometheus-digitalocean.yml
```diff
@@ -0,0 +1,27 @@
...
+  # Make Prometheus scrape itself for metrics.
+  - job_name: 'prometheus'
+    static_configs:
+    - targets: ['localhost:9090']
+
+  # Discover Node Exporter instances to scrape.
+  - job_name: 'node'
+
+    digitalocean_sd_configs:
+      - bearer_token: 8034e306bdb44ad882096043353fb303ba58aeff384b43c1a36b8926cb00bf8a
```
brian-brazil, line 15: This looks suspiciously like an actual value. I'd make it more obvious that this needs to be substituted in.

documentation/examples/prometheus-digitalocean.yml
```diff
@@ -0,0 +1,27 @@
...
+        port: 9100
+    relabel_configs:
+      # Only scrape targets that have a tag 'monitoring'.
+      - source_labels: [__meta_digitalocean_tags]
+        regex: '.*,monitoring,.*'
+        action: keep
+
+      # Use the public IPv6 address to scrape the target.
+      - source_labels: [__address__, __meta_digitalocean_public_ipv6]
```
brian-brazil, line 24: Given that the port number comes from the config above, it'd seem simpler to hardcode here rather than try to extract it from the address.

PR #7390, opened 2020-06-12
Optimized bstream reader used by XORChunk iterator

pracucci:
I spent the last couple of days investigating query performances and I've noticed that, for several queries we see in production, a significant amount of time is spent iterating series and sample. In this analysis, I've noticed most of the time spent iterating samples is in the `XORChunk` iterator and further benchmarks showed that the `bstream` reader introduces a significant bottleneck.

I've tried different approaches to optimise the `bstream` reader and in this PR I'm proposing the optimisation I've found having the least downsides. The optimisation is split into two parts:
1. Work with a `uint64` buffer in order to optimise bit-wise operations (assuming most production architectures are 64 bit).
2. Introduce `Fast()` version of read bit/bits which get inlined by the compiler. These functions must be leafs and super small, reason why may return an error in case the current buffer  doesn't contain enough bits to satify the request. In case a `Fast()` function returns error, it must be retried using the normal (but slower) one.

The following show the result of a couple of benchmarks. I've measured similar query reduction times running benchmarks in Cortex too.

## BenchmarkXORIterator

```
benchmark                  old ns/op     new ns/op     delta
BenchmarkXORIterator-4     33.0          22.7          -31.21%

benchmark                  old allocs     new allocs     delta
BenchmarkXORIterator-4     0              0              +0.00%

benchmark                  old bytes     new bytes     delta
BenchmarkXORIterator-4     1             1             +0.00%
```

## BenchmarkRangeQuery

```
benchmark                                                                                                             old ns/op      new ns/op      delta
BenchmarkRangeQuery/expr=a_one,steps=1-4                                                                              15400          14621          -5.06%
BenchmarkRangeQuery/expr=a_one,steps=10-4                                                                             15519          14847          -4.33%
BenchmarkRangeQuery/expr=a_one,steps=100-4                                                                            23341          22079          -5.41%
BenchmarkRangeQuery/expr=a_one,steps=1000-4                                                                           86703          76500          -11.77%
BenchmarkRangeQuery/expr=a_ten,steps=1-4                                                                              74669          65145          -12.75%
BenchmarkRangeQuery/expr=a_ten,steps=10-4                                                                             74867          66844          -10.72%
BenchmarkRangeQuery/expr=a_ten,steps=100-4                                                                            155213         135659         -12.60%
BenchmarkRangeQuery/expr=a_ten,steps=1000-4                                                                           806163         677584         -15.95%
[...]

PR #7413, opened 2020-06-18
Cleanup bstream based on post-merge feedback received on 7390

pracucci:
@codesome left some valuable feedback to #7390 after getting it merged. In this PR I'm addressing the feedback.

We also agreed offline to not address [this comment](https://github.com/prometheus/prometheus/pull/7390#discussion_r440571574) because it's hard to know whether the `Fast()` version of a function is expected to return an error or not for every single call.

### DIFF
--- a/tsdb/chunkenc/bstream.go
+++ b/tsdb/chunkenc/bstream.go
@@ -41,7 +41,10 @@
 
 package chunkenc
 
-import "io"
+import (
+	"encoding/binary"
+	"io"
+)
 
 // bstream is a stream of bits.
 type bstream struct {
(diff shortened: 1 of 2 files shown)


roidelapluie:
Where is the copyright going then?

roidelapluie:
"Redistributions of source code must retain the above copyright notice"

pracucci:
> Where is the copyright going then?

It was erroneously copied from `bstream.go` but it doesn't apply to the test file. The test file should just have the standard Prometheus license.

roidelapluie:
For bstream.go, to respect  `Redistributions in binary form must reproduce the above copyright notice`, it should be added in the NOTICE file.

codesome:
The code in bstream_test.go does not copy any code from there, so I don't think it's necessary

roidelapluie:
for _test, ok but bstream.go sounds like something that should be added to NOTICE for the binary redistribution

codesome:
Let's fix that in another PR

Issue #7412, opened 2020-06-17
opening storage failed: found unsequential head chunk files 23 and 25


roidelapluie:
**What did you do?**

Exit prometheus cleanly

```
level=warn ts=2020-06-17T18:14:44.525Z caller=main.go:524 msg="Received SIGTERM, exiting gracefully..."
level=info ts=2020-06-17T18:14:44.525Z caller=main.go:547 msg="Stopping scrape discovery manager..."
level=info ts=2020-06-17T18:14:44.525Z caller=main.go:561 msg="Stopping notify discovery manager..."
level=info ts=2020-06-17T18:14:44.525Z caller=main.go:583 msg="Stopping scrape manager..."                          
level=info ts=2020-06-17T18:14:44.525Z caller=main.go:543 msg="Scrape discovery manager stopped"                
level=info ts=2020-06-17T18:14:44.525Z caller=main.go:557 msg="Notify discovery manager stopped"                       
level=info ts=2020-06-17T18:14:44.525Z caller=manager.go:888 component="rule manager" msg="Stopping rule manager..."
level=info ts=2020-06-17T18:14:44.525Z caller=manager.go:898 component="rule manager" msg="Rule manager stopped"       
level=info ts=2020-06-17T18:14:44.525Z caller=main.go:577 msg="Scrape manager stopped"                                                
level=info ts=2020-06-17T18:14:44.531Z caller=notifier.go:601 component=notifier msg="Stopping notification manager..."               
level=info ts=2020-06-17T18:14:44.531Z caller=main.go:749 msg="Notifier manager stopped"
level=info ts=2020-06-17T18:14:44.531Z caller=main.go:761 msg="See you next time!"
```

**What did you expect to see?**

I can restart it

**What did you see instead? Under which circumstances?**

```
level=info ts=2020-06-17T18:14:51.603Z caller=main.go:302 msg="No time or size retention was set so using the default time retention" durat
ion=15d
level=info ts=2020-06-17T18:14:51.603Z caller=main.go:337 msg="Starting Prometheus" version="(version=2.19.0, branch=master, revision=86dc8
9a39aefa5f8f64af25c22f2055137ba3423)"
level=info ts=2020-06-17T18:14:51.603Z caller=main.go:338 build_context="(go=go1.15beta1, user=roidelapluie@oxygen, date=20200617-18:00:10)
"
[...]

PR #7414, opened 2020-06-18
Fix unsequential m-map files

codesome:
Fixes https://github.com/prometheus/prometheus/issues/7412

Note that this is against the release branch as it requires a v2.19.1

### DIFF
--- a/tsdb/chunks/head_chunks.go
+++ b/tsdb/chunks/head_chunks.go
@@ -670,8 +670,8 @@ func (cdm *ChunkDiskMapper) Truncate(mint int64) error {
 
 	var removedFiles []int
 	for _, seq := range chkFileIndices {
-		if seq == cdm.curFileSequence {
-			continue
+		if seq == cdm.curFileSequence || cdm.mmappedChunkFiles[seq].maxt >= mint {
+			break
 		}
 		if cdm.mmappedChunkFiles[seq].maxt < mint {
 			removedFiles = append(removedFiles, seq)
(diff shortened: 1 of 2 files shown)


codesome:
In this PR:

1. Breaks the truncation after the first file which wont be deleted.
2. Does not create a new file on truncation if the current file is empty.

tsdb/chunks/head_chunks_test.go
```diff
@@ -245,14 +245,91 @@ func TestHeadReadWriter_Truncate(t *testing.T) {
...
+func TestHeadReadWriter_Truncate_NoUnsequentialFiles(t *testing.T) {
+	hrw, close := testHeadReadWriter(t)
+	defer func() {
+		testutil.Ok(t, hrw.Close())
+		close()
+	}()
+
+	testutil.Assert(t, !hrw.fileMaxtSet, "")
+	testutil.Ok(t, hrw.IterateAllChunks(func(_, _ uint64, _, _ int64, _ uint16) error { return nil }))
+	testutil.Assert(t, hrw.fileMaxtSet, "")
```
krasi-georgiev, line 270: would it make sence to put all this bootstrap inside `testHeadReadWriter`

tsdb/chunks/head_chunks.go
```diff
@@ -680,7 +680,10 @@ func (cdm *ChunkDiskMapper) Truncate(mint int64) error {
 	cdm.readPathMtx.RUnlock()
 
 	var merr tsdb_errors.MultiError
-	merr.Add(cdm.CutNewFile())
+	if cdm.curFileNumBytes > HeadChunkFileHeaderSize {
+		// Cut a new file only if the current file has some chunks.
```
krasi-georgiev, line 684: move the comment before the `if`?

PR #7420, opened 2020-06-18
Docker Swarm service discovery

### LINKED ISSUE #1766: Docker engine swarm api service discovery

roidelapluie:
This is a docker swarm implementation without business logic.

Fixes #1766 


discovery/dockerswarm/dockerswarm.go
```diff
@@ -0,0 +1,160 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/docker/docker/api/types"
```
brian-brazil, line 24: These have changed repo name, it'd be best to follow.

docs/configuration/configuration.md
```diff
@@ -444,6 +448,55 @@ tls_config:
...
+* `__meta_dockerswarm_endpoint_port_name`: the name of the endpoint port, if available
+* `__meta_dockerswarm_endpoint_port_publish_mode`: the publish mode of the endpoint port
+* `__meta_dockerswarm_service_id`: the id of the service
+* `__meta_dockerswarm_service_label_`: each label of the service
+* `__meta_dockerswarm_service_name`: the name of the service
+* `__meta_dockerswarm_updating_status`: the status of the service, if available
+* `__meta_dockerswarm_virtual_ip_address`: the address of the virtual ip
+* `__meta_dockerswarm_virtual_ip_network_id`: the network id of the virtual ip
+
+```yaml
```
brian-brazil, line 468: Host is missing

docs/configuration/configuration.md
```diff
@@ -444,6 +448,55 @@ tls_config:
...
+# Authentication information used to authenticate to the Docker daemon.
```
brian-brazil, line 469: None of the http stuff is likely relevant here, I'd put it down the end.

discovery/dockerswarm/dockerswarm.go
```diff
@@ -0,0 +1,160 @@
...
+
+	opts := []client.Opt{
+		client.WithHost(conf.Host),
+		client.WithHTTPClient(&http.Client{
+			Transport: rt,
+			Timeout:   5 * time.Duration(conf.RefreshInterval),
+		}),
+	}
+	if conf.Insecure {
+		opts = append(opts, client.WithScheme("http"))
```
brian-brazil, line 98: This is a bit of a confusing name for what it does, it's a bit close to insecure_skip_verify.

Also, host appears to be a URL so why do you need this? Also the default scheme is "unix" or "npipe" so changing it to "http" seems a bit off.

discovery/dockerswarm/dockerswarm.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// SDConfig is the configuration for Docker Swarm based service discovery.
+type SDConfig struct {
+	HTTPClientConfig config_util.HTTPClientConfig `yaml:",inline"`
+
+	Host     string `yaml:"host"`
+	Insecure bool   `yaml:"insecure"`
+
+	RefreshInterval model.Duration `yaml:"refresh_interval,omitempty"`
```
brian-brazil, line 60: Why would this be empty?

discovery/dockerswarm/dockerswarm.go
```diff
@@ -0,0 +1,160 @@
...
+	rt, err := config_util.NewRoundTripperFromConfig(conf.HTTPClientConfig, "dockerswarm_sd", false)
+	if err != nil {
+		return nil, err
+	}
+
+	opts := []client.Opt{
+		client.WithHost(conf.Host),
+		client.WithHTTPClient(&http.Client{
+			Transport: rt,
+			Timeout:   5 * time.Duration(conf.RefreshInterval),
```
brian-brazil, line 94: That seems a bit long

discovery/dockerswarm/dockerswarm.go
```diff
@@ -0,0 +1,160 @@
...
+			if e.Protocol != swarm.PortConfigProtocolTCP {
+				continue
+			}
+			for _, p := range s.Endpoint.VirtualIPs {
+				labels := model.LabelSet{
+					swarmLabelServiceID:               model.LabelValue(s.ID),
+					swarmLabelServiceName:             model.LabelValue(s.Spec.Name),
+					swarmLabelEndpointPortName:        model.LabelValue(e.Name),
+					swarmLabelEndpointPortPublishMode: model.LabelValue(e.PublishMode),
+					swarmLabelVirtualIPAddr:           model.LabelValue(p.Addr),
```
brian-brazil, line 136: This name is a little misleading, as it's a bit more than an address.

discovery/dockerswarm/dockerswarm.go
roidelapluie: I am following their documentation https://docs.docker.com/engine/api/sdk/

roidelapluie:
> Are there up to date godocs for the library somewhere? The usual ones are years out of date.

No, I run godoc -http=:8080 locally ... :cry:

discovery/dockerswarm/dockerswarm.go
roidelapluie: Swarm does not support unix/npipe.

If we put http in the url, it is still changed to https, unless we force the protocol.

discovery/dockerswarm/dockerswarm.go
roidelapluie: Unless users start lowering the refresh interval to 10s. WDYT? We do this for other SD's.

docs/configuration/configuration.md
roidelapluie: ack

docs/configuration/configuration.md
roidelapluie: That information is not exposed by the API

discovery/dockerswarm/dockerswarm.go
roidelapluie: Switched to scheme.

discovery/dockerswarm/dockerswarm.go
brian-brazil: Having a timeout longer than the refresh doesn't seem right, what other SDs do this?

discovery/dockerswarm/dockerswarm.go
brian-brazil: That's not what the source code seems to say, my reading is that Host is sufficient for everything and I'm not seeing anything about changing the scheme: https://github.com/moby/moby/blob/master/client/options.go#L81-L86

docs/configuration/configuration.md
roidelapluie: There migth be a solution to this

brian-brazil: I mean you haven't documented a config field.

discovery/dockerswarm/dockerswarm.go
roidelapluie: openstack

roidelapluie:
Note: we will want the tasks too, to get relevant metadata.

Note: enable version negotiation.

docs/configuration/configuration.md
roidelapluie: oops ;)

discovery/dockerswarm/dockerswarm.go
```diff
@@ -0,0 +1,242 @@
...
+	swarmLabelNodeRole                 = swarmLabelNodePrefix + "role"
+	swarmLabelNodeStatus               = swarmLabelNodePrefix + "status"
+)
+
+// DefaultSDConfig is the default Docker Swarm SD configuration.
+var DefaultSDConfig = SDConfig{
+	RefreshInterval: model.Duration(60 * time.Second),
+	Kind:            "services",
+	Port:            80,
+	Scheme:          "https",
```
roidelapluie, line 73: @brian-brazil Here is the code that forces HTTPS as scheme in the client:

https://github.com/moby/moby/blob/a23ca165c919b62da681cf062429ffe36d249a1b/client/client.go#L151

hence scheme is required here

roidelapluie: I have removed the need to expose those details to end users.

discovery/dockerswarm/dockerswarm.go
roidelapluie: solved

roidelapluie:
I have reviewed the docs and this is ready for a deeper review @brian-brazil @SuperQ

For your convenience & easy review, I have pushed relevant go docs here  https://roidelapluie.github.io/godoc-docker/pkg/github.com/docker/docker/client/index.html

discovery/dockerswarm/dockerswarm.go
roidelapluie: Also all the imports use docker/docker internally.

roidelapluie:
I have been looking at what the existing service discoveries are doing ; next to the business logic, it is clear that the labels we get with one API call do not cut it. Hence I have added calls to get information about the networks & nodes, and added the useful labels to the services and tasks.

Note: This is not unusual as we do the same kind of things with other service discoveries, like EC2, where multiple calls are needed to get target info.

brian-brazil:
> This is not unusual as we do the same kind of things with other service discoveries

In general that's fine, as long as there's not quotas or general slowness in the way.

roidelapluie:
We still miss the tasks that publish host ports directly

discovery/dockerswarm/dockerswarm.go
```diff
@@ -0,0 +1,151 @@
...
+)
+
+const (
+	swarmLabel = model.MetaLabelPrefix + "dockerswarm_"
+)
+
+// DefaultSDConfig is the default Docker Swarm SD configuration.
+var DefaultSDConfig = SDConfig{
+	RefreshInterval: model.Duration(60 * time.Second),
+	Host:            "http://127.0.0.1:2375/",
```
brian-brazil, line 40: Hmm, does it make sense to default this to local? The only other place we do that is Consul, but that's because it's deployed everywhere and you're only meant to talk to the local one.

discovery/dockerswarm/dockerswarm.go
```diff
@@ -0,0 +1,151 @@
...
+	Kind:            "services",
```
brian-brazil, line 41: Kind is not defaulted for other SDs, so we should require explicit config here too.

discovery/dockerswarm/dockerswarm.go
```diff
@@ -0,0 +1,151 @@
...
+	Values []string `yaml:"values"`
+}
+
+// SDConfig is the configuration for Docker Swarm based service discovery.
+type SDConfig struct {
+	HTTPClientConfig config_util.HTTPClientConfig `yaml:",inline"`
+
+	Host    string `yaml:"host"`
+	url     *url.URL
+	Filters []Filter `yaml:"filters"`
```
brian-brazil, line 56: Same question as before, are these actually needed for performance? Filters like these should only exist when users are running into real-world performance issues, not as an alternative to relabelling. Especially when there's multiple kinds that interact, as that get's complicated as you can see in k8.

discovery/dockerswarm/dockerswarm.go
```diff
@@ -0,0 +1,151 @@
...
+	err := unmarshal((*plain)(c))
+	if err != nil {
+		return err
+	}
+	url, err := url.Parse(c.Host)
+	if err != nil {
+		return err
+	}
+	c.url = url
+	if _, ok := (map[string]struct{}{"services": {}, "nodes": {}, "tasks": {}})[c.Kind]; !ok {
```
brian-brazil, line 76: This is a bit weird, a switch would be clearer.

discovery/dockerswarm/dockerswarm.go
```diff
@@ -0,0 +1,151 @@
...
+		port:    conf.Port,
+		filters: conf.Filters,
+	}
+
+	rt, err := config_util.NewRoundTripperFromConfig(conf.HTTPClientConfig, "dockerswarm_sd", false)
+	if err != nil {
+		return nil, err
+	}
+
+	// This is used in tests. In normal situation, it is set when Unmarshaling.
```
brian-brazil, line 103: situations

discovery/dockerswarm/network.go
```diff
@@ -0,0 +1,52 @@
...
+	swarmLabelNetworkIngress     = swarmLabelNetworkPrefix + "ingress"
+	swarmLabelNetworkLabelPrefix = swarmLabelNetworkPrefix + "label_"
+)
+
+func (d *Discovery) getNetworkLabels(ctx context.Context, networkID string) (model.LabelSet, error) {
+	network, err := d.client.NetworkInspect(ctx, networkID, types.NetworkInspectOptions{})
+	if err != nil {
+		return nil, err
+	}
+	labels := model.LabelSet{
```
brian-brazil, line 40: Can we not use a plain map of strings? Use of model should be avoided for basic types, using it like this just makes things more complicated.

discovery/dockerswarm/nodes.go
```diff
@@ -0,0 +1,108 @@
...
+			swarmLabelNodeAvailability:         model.LabelValue(n.Spec.Availability),
+			swarmLabelNodeHostname:             model.LabelValue(n.Description.Hostname),
+			swarmLabelNodePlatformArchitecture: model.LabelValue(n.Description.Platform.Architecture),
+			swarmLabelNodePlatformOS:           model.LabelValue(n.Description.Platform.OS),
+			swarmLabelNodeEngineVersion:        model.LabelValue(n.Description.Engine.EngineVersion),
+			swarmLabelNodeStatus:               model.LabelValue(n.Status.State),
+			swarmLabelNodeAddress:              model.LabelValue(n.Status.Addr),
+		}
+		if n.ManagerStatus != nil {
+			labels[model.LabelName(swarmLabelNodeManagerLeader)] = model.LabelValue(fmt.Sprintf("%v", n.ManagerStatus.Leader))
```
brian-brazil, line 69: Prefer being explicit over using %v

discovery/dockerswarm/nodes.go
```diff
@@ -0,0 +1,108 @@
...
+		labels[model.AddressLabel] = model.LabelValue(addr)
+
+		tg.Targets = append(tg.Targets, labels)
+
+	}
+	return []*targetgroup.Group{tg}, nil
+}
+
+func (d *Discovery) getNodeLabels(ctx context.Context, nodeID string) (model.LabelSet, error) {
+	n, _, err := d.client.NodeInspectWithRaw(ctx, nodeID)
```
brian-brazil, line 89: This is a lot of RPCs, can we not get this from NodeList?

Similarly with the others.

discovery/dockerswarm/services.go
```diff
@@ -0,0 +1,136 @@
...
+		ln := strutil.SanitizeLabelName(k)
+		labels[model.LabelName(swarmLabelServiceLabelPrefix+ln)] = model.LabelValue(v)
+	}
+
+	return labels, s.Endpoint.Ports, nil
+}
+
+func getServiceValueMode(s swarm.Service) model.LabelValue {
+	if s.Spec.Mode.Global != nil {
+		return swarmLabelValueModeGlobal
```
brian-brazil, line 130: I don't like indirection like this so I have to search the whole codebase to see what the value actually is, use a constant here.

discovery/dockerswarm/services.go
```diff
@@ -0,0 +1,136 @@
...
+		for _, e := range s.Endpoint.Ports {
+			if e.Protocol != swarm.PortConfigProtocolTCP {
+				continue
+			}
+			for _, p := range s.Endpoint.VirtualIPs {
+				labels := model.LabelSet{
+					swarmLabelServiceEndpointPortName:        model.LabelValue(e.Name),
+					swarmLabelServiceEndpointPortPublishMode: model.LabelValue(e.PublishMode),
+					swarmLabelServiceID:                      model.LabelValue(s.ID),
+					swarmLabelServiceMode:                    getServiceValueMode(s),
```
brian-brazil, line 69: Generally we'd inline this, rather than hiding it among the simple string assignments.

discovery/dockerswarm/tasks.go
```diff
@@ -0,0 +1,164 @@
...
+
+			labels := model.LabelSet{
+				model.LabelName(swarmLabelTaskPortMode): model.LabelValue(p.PublishMode),
+			}
+
+			for k, v := range commonLabels {
+				labels[k] = v
+			}
+
+			for k, v := range serviceLabels[s.ServiceID] {
```
brian-brazil, line 111: Didn't you already add these back in line 90?

docs/configuration/configuration.md
```diff
@@ -444,6 +448,148 @@ tls_config:
...
+This is the default `<swarm_kind>`.
+
+Available meta labels:
+
+* `__meta_dockerswarm_service_id`: the id of the service
+* `__meta_dockerswarm_service_name`: the name of the service
+* `__meta_dockerswarm_service_mode`: the mode of the service
+* `__meta_dockerswarm_service_endpoint_port_name`: the name of the endpoint port, if available
+* `__meta_dockerswarm_service_endpoint_port_publish_mode`: the publish mode of the endpoint port
+* `__meta_dockerswarm_service_label_`: each label of the service
```
brian-brazil, line 470: `label_<foo>`

discovery/dockerswarm/tasks.go
```diff
@@ -0,0 +1,164 @@
...
+					}
+
+					ip, _, err := net.ParseCIDR(address)
+					if err != nil {
+						return nil, fmt.Errorf("error while parsing address %s: %w", address, err)
+					}
+					addr := net.JoinHostPort(ip.String(), strconv.FormatUint(uint64(p.PublishedPort), 10))
+					labels[model.AddressLabel] = model.LabelValue(addr)
+
+					tg.Targets = append(tg.Targets, labels)
```
brian-brazil, line 158: Why are we adding two different types of target?

discovery/dockerswarm/dockerswarm.go
roidelapluie: indeed let's make it mandatory.

discovery/dockerswarm/dockerswarm.go
roidelapluie: ack

discovery/dockerswarm/dockerswarm.go
roidelapluie: Yes. The Swarm API's are not paginated and you can go to very big swarm clusters.

Also, filter is common for all the kinds.

discovery/dockerswarm/dockerswarm.go
roidelapluie: done

discovery/dockerswarm/services.go
roidelapluie: inlined.

discovery/dockerswarm/tasks.go
roidelapluie: Because there are two ways exposing tasks

https://docs.docker.com/engine/swarm/services/#publish-ports

- Publish a service’s ports using the routing mesh
- Publish a service’s ports directly on the swarm node

discovery/dockerswarm/nodes.go
roidelapluie: done

docs/configuration/configuration.md
roidelapluie: Addressed

roidelapluie:
I have addressed all comments :flying_saucer:

discovery/dockerswarm/dockerswarm.go
brian-brazil: If that's the case, then we probably need to approach this like k8 and have a per-type filter.

I'd still prefer we wait for users to bring this up, we only got the k8 one due to an odd non-Prometheus use case from Loki - it was never really a problem for general users.

discovery/dockerswarm/nodes.go
```diff
@@ -0,0 +1,111 @@
...
+		Source: "DockerSwarm",
+	}
+
+	nodes, err := d.client.NodeList(ctx, types.NodeListOptions{Filters: makeArgs(d.filters)})
+	if err != nil {
+		return nil, fmt.Errorf("error while listing swarm nodes: %w", err)
+	}
+
+	for _, n := range nodes {
+		labels := model.LabelSet{
```
brian-brazil, line 57: You're still using model here and a few other places.

discovery/dockerswarm/tasks.go
brian-brazil: Should they be different roles?

docs/configuration/configuration.md
```diff
@@ -444,6 +448,148 @@ tls_config:
...
+
+Available meta labels:
+
+* `__meta_dockerswarm_task_id`: the id of the task
+* `__meta_dockerswarm_task_container_id`: the container id of the task
+* `__meta_dockerswarm_task_desired_state`: the desired state of the task
+* `__meta_dockerswarm_task_label_<labelname>`: each label of the task
+* `__meta_dockerswarm_task_slot`: the slot of the task
+* `__meta_dockerswarm_task_state`: the state of the task
+* `__meta_dockerswarm_task_port_publish_mode`: the publish node of the task port
```
brian-brazil, line 493: mode

docs/configuration/configuration.md
```diff
@@ -444,6 +448,148 @@ tls_config:
...
+
+# Optional filters to limit the discovery process to a subset of available resources. 
+
+# Note: When making decision about using filters make sure that this is the best
+# approach - it will prevent Prometheus from reusing single API calls for all
+# scrape configs. This might result in a bigger load on the Docker API, because
+# per each filters combination there will be additional API calls. On the other hand,
+# if you just want to monitor small subset of resources in large cluster it's
+# recommended to use filters.
+# Decision, if filters should be used or not depends on the particular situation.
```
brian-brazil, line 562: This line doesn't add anything.

discovery/dockerswarm/tasks.go
roidelapluie: No because from a users and commands perspective they are really similar.

docs/configuration/configuration.md
```diff
@@ -444,6 +448,148 @@ tls_config:
...
+
+# Optional proxy URL.
+[ proxy_url: <string> ]
+
+# TLS configuration.
+tls_config:
+  [ <tls_config> ]
+
+# Kind of the targets to retrieve.
+kind: <swarm_kind>
```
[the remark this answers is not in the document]
roidelapluie: We also have
<domain_name>
<query_type>
to fix later

discovery/dockerswarm/nodes.go
roidelapluie: This is the list of targets, not an intermediate step, I have to.

roidelapluie: I have found some placed where that was not needed

discovery/dockerswarm/tasks.go
brian-brazil: We should probably clarify a little in the docs which labels are present for each then.

roidelapluie: It is here already:

The __meta_dockerswarm_network_* meta labels are not populated for ports which are published with mode=host.

docs/configuration/configuration.md
```diff
@@ -444,6 +448,132 @@ tls_config:
...
+* `__meta_dockerswarm_node_manager_reachability`: the reachability of the manager component of the node
+* `__meta_dockerswarm_node_platform_architecture`: the architecture of the node
+* `__meta_dockerswarm_node_platform_os`: the operating system of the node
+* `__meta_dockerswarm_node_role`: the role of the node
+* `__meta_dockerswarm_node_status`: the status of the node
+
+See below for the configuration options for Docker Swarm discovery:
+
+```yaml
+# Address of the Docker daemon. Docker Swarm supports http and https protocol.
```
brian-brazil, line 540: protocols

Do we even need to mention what protocols are supported? These docs are for the SD, not swarm in general.

roidelapluie: indeed. We can add it later if we have lots of user questions trying the unix socket.

SuperQ:
Awesome work @roidelapluie.

darkl0rd:
Just tested this with the current RC:

I would suggest that the "host" parameter also supports unix-sockets. Exposing the docker socket over TCP is not something you would usually do in a production environment. Granting other containers access to the docker management port is commonly done by mounting /var/run/docker.sock read-only inside the container.

roidelapluie:
Thank you. We do not limit that; it is a limitation of docker swarm AFAIK.

darkl0rd:
Docker (Swarm) allows you to access the docker.sock; I was under the impression that the 'host' parameter (of prometheus) only accepts a address. Are you saying that one can also use file:// as a value for host?

2375 is by default not exposed, nor is 2376. Since 19.03 this is also considered malpractice, as docker can be managed through SSH (DOCKER_HOST=ssh://manager-node.domain) - which means you no longer need to expose the socket insecurely (2375) or setup a separate PKI (2376) but rather securely manage your Swarm using your existing PKI (SSH). Effectively the Docker client establishes an SSH session to the ${DOCKER_HOST} and communicates with the Unix socket (/var/run/docker.sock) of that host.

roidelapluie:
> Docker (Swarm) allows you to access the docker.sock; I was under the impression that the 'host' parameter (of prometheus) only accepts a address. Are you saying that one can also use file:// as a value for host?
> 
> 2375 is by default not exposed, nor is 2376. Since 19.03 this is also considered malpractice, as docker can be managed through SSH (DOCKER_HOST=ssh://manager-node.domain) - which means you no longer need to expose the socket insecurely (2375) or setup a separate PKI (2376) but rather securely manage your Swarm using your existing PKI (SSH). Effectively the Docker client establishes an SSH session to the ${DOCKER_HOST} and communicates with the Unix socket (/var/run/docker.sock) of that host.

Can you comment on the issue #7603? thanks

darkl0rd:
I just noticed, that only services seem to be discovered which have a port published? Is there a way around this? Many services do not expose their ports externally as they are interconnected by Docker overlay networks.

roidelapluie:
> I just noticed, that only services seem to be discovered which have a port published? Is there a way around this? Many services do not expose their ports externally as they are interconnected by Docker overlay networks.

Prometheus 2.21.0 (out end of August) will expose those services as well (using the port: specified in the config). You can test it with the docker images `prom/prometheus:master`. Note that that version of prometheus is for testing only.

darkl0rd:
My finding is the opposite, services which do **not** have a port exposed (those without a port: in the config) are not being discovered.

umang5195:
@roidelapluie I tired using Docker Swarm Service Discovery in prometheus, but facing problems using it. I followed the [docker swarm support documentation](https://prometheus.io/docs/guides/dockerswarm/). Created a daemon.json file and mounted /var/run/docker.sock in prometheus container. Container is giving permission denied error as prometheus is running as `nobody` and doesn't have access to mounted /var/run/docker.sock. Below is my prometheus.yml.
Prometheus Version : v2.20.1

```
 prometheus:
    image: prom/prometheus
    networks:
      - monitor
    ports:
      - "9090:9090"
    command:
      - '--config.file=/etc/prometheus/prometheus.yml'
      - '--storage.tsdb.path=/prometheus'
      - '--storage.tsdb.retention=${PROMETHEUS_RETENTION:-24h}'
    volumes:
      - prometheus:/prometheus
      - /home/efs/devops/dsm:/etc/prometheus:ro
      - /var/run/docker.sock:/var/run/docker.sock:ro
    deploy:
      mode: replicated
      replicas: 1
      resources:
        limits:
          memory: 1024M
        reservations:
          memory: 128M
```

Prometheus.yml

```
scrape_configs:
  - job_name: 'docker'
    dockerswarm_sd_configs:
    - host: unix:///var/run/docker.sock
      role: nodes
```

Error:
`monitor_prometheus.1.uwda893b1nws@monitoring.dev.ai | level=error ts=2020-08-06T07:21:19.106Z caller=refresh.go:98 component="discovery manager scrape" discovery=dockerswarm msg="Unable to refresh target groups" err="error while listing swarm nodes: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get \"http://%2Fvar%2Frun%2Fdocker.sock/v1.24/nodes\": dial unix /var/run/docker.sock: connect: permission denied"`

hamiltont:
@umang5195 prometheus runs as the linux user `nobody` inside the container. Permissions on the docker socket are effectively restricted to `root`. Either re-configure prometheus to talk to docker over HTTP, or add `user: root` to your docker-compose file

roidelapluie:
I think that those kind of questions could be answered by the Swarm community as well. Running prometheus as root is not a good idea.

hamiltont:
> Running prometheus as root is not a good idea.

Agree. However, the problem is much broader than this, and that makes running as root arguably one of the better short-term ways to remove the error message. Access to the docker daemon is all-or-nothing, and anyone with access has effective root on the host machine. So regardless of how we do it (unix socket or TCP), if prometheus can access the docker daemon is effectively running with root access on the host system. [In fact, *not* doing this is OWASPs #1 rule for securing docker](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-1-do-not-expose-the-docker-daemon-socket-even-to-the-containers). 

Even if we wrote a different Dockerfile with a custom linux user (e.g. `promd`), any process running as `promd` user could use the docker daemon to run as effective root on the host. A simple example is `docker run -it -v /:/host-root ubuntu bash` - cd into the `/host-root` and you can relatively-freely browse the entire host filesystem. 

We *could* rewrite the main example to use TLS. This would shift, not solve, the problem. We would no longer need to run prometheus as `root`, because there would be no problems with reading the socket. However, and process that could read the TLS certificates could still use the docker socket. So we're still effectively running our container with 'root-on-host' permissions. 

There are some settings that can help to mitigate the all-or-nothing problem by reducing the set of "all" from "everything under the sun" to "only run containers without these settings". For example, rootless docker daemon is an ongoing project, but has serious limitations (no cgroup support). 

As far as I know, the only real solution is for the docker daemon itself to have access controls built into the system. This is exactly what the [Docker Access Authorization Plugin](https://docs.docker.com/engine/extend/plugins_authorization/#basic-architecture) is all about - allowing you to grant access to the daemon without that access being all-or-nothing. 

Pragmatically (e.g. what can prom authors do about this), the only thing we can do is warn users if they are using swarm service discovery that access to the docker socket is dangerous, and perhaps link them to a good summary page on docker's website. @umang5195 is sadly right that the example in the documentation will be broken out of the box unless you someone give read access to the docker socket. Adding TCP access or a custom-built Dockerfile with a dedicated Linux user to that example would drown out the actual example with pages of configuration guidance. My suggestion would be to have some type of warning box right in the example with red coloring. For example: ("Warning: To run this example, we have to grant prometheus access to the docker socket by running as root. This should not be done in production. See here for information on accessing the docker socket"). Sadly I don't think we can say "on accessing the docker socket safely", because there simply is not an option to allow for that right now

darkl0rd:
A solution, well workaround is, to expose the docker socket through a sidecar (eg caddy) which mounts the docker.sock read only and has a private network with the Prometheus container. It doesn’t solve all the security concerns, and addresses part of them through, if you will, obscurity. But it’s the best thing I have come up with so far.

PR #7423, opened 2020-06-19
Use 10m timeouts for watches

pierresouchay:
use ?wait=10m will give results as fast as usual when data is changing
but will perform far fewer requests when services do not change.

On large infrastructure, this will reduce quite a lot the number of
queries/s on Consul servers while having the same performance for freshness
of results.

### DIFF
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -36,7 +36,7 @@ import (
 )
 
 const (
-	watchTimeout  = 30 * time.Second
+	watchTimeout  = 10 * time.Minute
 	retryInterval = 15 * time.Second
 
 	// addressLabel is the name for the label containing a target's address.
@@ -95,7 +95,7 @@ var (
 		Scheme:          "http",
 		Server:          "localhost:8500",
 		AllowStale:      true,
-		RefreshInterval: model.Duration(watchTimeout),
+		RefreshInterval: model.Duration(30 * time.Second),
 	}
 )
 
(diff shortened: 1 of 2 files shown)


brian-brazil:
@iksaif Does this sound sane to you?

roidelapluie:
Thank you Pierre,

did you check what happens on reload? Do we cleany cancel the queries? If consul restart, do we notice it before the 10 minutes?

pierresouchay:
Hello Julien,

> did you check what happens on reload? Do we cleanly cancel the queries?

In theory yes, context is canceled and request too, I am gonna test the change extensively in real life

> If consul restart, do we notice it before the 10 minutes?

Data will be returned immediately in case of a restart of the Consul agent, so all notifications will be as fast it is as fast as it used to.

The difference is: when the index is not changed for 10min, we just do a single call to Consul, while we used to perform 20 of them. On large stable clusters, it might thus decrease by 20 times the number of RPCs.

?wait=10m simply returns at minimum every 10m instead of 30s, but all updates are as fast as it used to be.

Of course, I also had to change RefreshInterval because RefreshInterval and Watch Timeout are semantically different, but for the end-user, the behavior should be exactly the same.

I did work a bit with @iksaif on the previous implementation (we used to be colleagues), especially around optimizations we put on Consul at that time, and we chose 30s at that time to be conservative, but few years passed and we gained experience, and I can tell you using 10m is gonna be safe and better for Consul cluster.

roidelapluie:
My question was more: if the connection with consul is broken, would it take 10 minutes to notice it?

pierresouchay:
No, detection would be fast and return http 500

iksaif:
Looks good to me

Le ven. 19 juin 2020 à 18:03, Pierre Souchay <notifications@github.com> a
écrit :

brian-brazil:
Thanks!

PR #7426, opened 2020-06-19
REACT UI: CollapsibleAlertPanel - value field more friendly human readable (scientific notation -> number)

nunokisc:
Clean solution to friendly human readable value in CollapsibleAlertPanel value field.  (scientific notation -> number)

### DIFF
--- a/web/ui/react-app/src/pages/alerts/AlertContents.test.tsx
+++ b/web/ui/react-app/src/pages/alerts/AlertContents.test.tsx
@@ -30,7 +30,7 @@ describe('AlertsContent', () => {
     });
   });
 
-  it('toggles the "annotations" checkbox from false to true when clicked and back to true when clicked again', () => {
+  it('toggles the "annotations" checkbox from false to true when clicked and back to false when clicked again', () => {
     wrapper.find('#show-annotations-toggler').invoke('onClick')();
     expect(wrapper.find('#show-annotations-toggler').prop('checked')).toBe(true);
     wrapper.find('#show-annotations-toggler').invoke('onClick')();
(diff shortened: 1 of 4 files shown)


roidelapluie:
mmh I think we should have that as a function on alerts and have tests on it, WDYT? See what it gives on small and large values.

nunokisc:
The default javascript uses float64, the field value is also a float64, the numbers function will never be a problem. When the numbers function is unable to convert a small number such as 0.0000007 (javascript only convert 6 zeros before a number) it will maintain scientific notation (7e-7). The same will happen with large numbers, the javacript converts them up to 7 zeros to the right, more than that it will be in scientific notation.
Correct me if I was wrong please.
But if the functions is the better approach its ok to me.

juliusv:
It seems like `Number()` returns `NaN` for `Inf`, `+Inf`, and `-Inf`, so we would have to handle that case separately. I'd probably just check whether `isNaN(result)` and if so, show the raw string value instead of the conversion.

juliusv:
And yeah, it would be nice to have that as a separate function and have tests for those cases, to make sure it's doing the right thing.

nunokisc:
You are absolutely right I did not consider the possibility of reaching values ​​like Inf, + Inf, and -Inf.

nunokisc:
Commited function convertSCToNumber taking in consideration Inf +Inf -Inf.
Commited tests to function convertSCToNumber.

nunokisc:
The function name has been normalized to "convertScientificNotationToNumber" which makes more sense

juliusv:
@nunokisc Thanks, looks mostly good, just `convertScientificNotationToNumber()` isn't quite descriptive of what the function is doing, since it parses any other floating point notations that are not scientific as well. How about just `parsePrometheusFloat()`, to differentiate it from `parseFloat()`, which doesn't support special float values?

nunokisc:
@juliusv You are right its the almost the same like parseFloat the difference is, in parseFloat there is Infinity and not Inf so thats is the perfect name. I'm gona commit the name change right now.

juliusv:
:+1: Great, thanks :)

PR #7430, opened 2020-06-21
fix tsdb analyze limit not work expected

ifooth:
## fix tsdb analyze limit not work expected

Signed-off-by: joelei thezero12@hotmail.com

today when i first use tsdb cmd analyze a block, limit flag not work expect; when i dig the code,  it's a little bug, i think; so i initiate a pr to fix it，thanks all guy

before fix:
image

after fix:
image


brian-brazil:
Can you add the DCO please?

ifooth:
> Can you add the DCO please?

Done, I have forced push for the merge commit

brian-brazil:
Thanks!

PR #7433, opened 2020-06-22
Update repo sync policies

SuperQ:
* Don't try and sync non-apache license files.
* Force create CODE_OF_CONDUCT.md.
* Switch to using an array of files to update.

### DIFF
--- a/scripts/sync_repo_files.sh
+++ b/scripts/sync_repo_files.sh
@@ -53,17 +53,30 @@ post_pull_request() {
     "https://api.github.com/repos/${1}/pulls"
 }
 
+check_license() {
+  # Check to see if the input is an Apache license of some kind
+  echo "$1" | grep --quiet --no-messages --ignore-case 'Apache License'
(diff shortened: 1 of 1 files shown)


scripts/sync_repo_files.sh
```diff
@@ -53,28 +53,41 @@ post_pull_request() {
     "https://api.github.com/repos/${1}/pulls"
 }
 
+check_license() {
+  # Check to see if the input is an Apache license of some kind
+  echo "$1" | grep --quiet --no-messages --ignore-case 'Apache License'
```
roidelapluie, line 58:
```suggestion
  grep --quiet --no-messages --ignore-case 'Apache License' "$1"
```

SuperQ: The input is a string with the contents of the target file, not a filename.

roidelapluie: oh OK

PR #7435, opened 2020-06-22
Fixed returned API status code on error

pracucci:
The PR https://github.com/prometheus/prometheus/pull/7251 has introduced the wrapping of `Select()` error (ie. [see this](https://github.com/prometheus/prometheus/pull/7251/files#diff-83539908220a826e39b7ba3a4f27d5d6R1115-R1119)). This change causes `returnAPIError()`   to return `errorExec`, which turns into having the API response status code 422 in case of a `promql.ErrStorage` (this error is used by Cortex in several circumstances).

This PR fixes it.

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -469,13 +469,16 @@ func returnAPIError(err error) *apiError {
 		return nil
 	}
 
-	switch err.(type) {
-	case promql.ErrQueryCanceled:
-		return &apiError{errorCanceled, err}
-	case promql.ErrQueryTimeout:
-		return &apiError{errorTimeout, err}
-	case promql.ErrStorage:
-		return &apiError{errorInternal, err}
+	// The error may have been wrapped, so we should also check the parent ones.
+	for cause := err; cause != nil; cause = errors.Unwrap(cause) {
+		switch cause.(type) {
+		case promql.ErrQueryCanceled:
+			return &apiError{errorCanceled, cause}
+		case promql.ErrQueryTimeout:
+			return &apiError{errorTimeout, cause}
+		case promql.ErrStorage:
+			return &apiError{errorInternal, cause}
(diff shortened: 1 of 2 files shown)


web/api/v1/api.go
```diff
@@ -469,13 +469,16 @@ func returnAPIError(err error) *apiError {
...
 	}
 
-	switch err.(type) {
-	case promql.ErrQueryCanceled:
-		return &apiError{errorCanceled, err}
-	case promql.ErrQueryTimeout:
-		return &apiError{errorTimeout, err}
-	case promql.ErrStorage:
-		return &apiError{errorInternal, err}
+	// The error may have been wrapped, so we should also check the parent ones.
```
codesome, line 472: Unwrap feels like we are looking at child errors (just my opinion, not a blocker)

web/api/v1/api.go
```diff
@@ -469,13 +469,16 @@ func returnAPIError(err error) *apiError {
...
+	for cause := err; cause != nil; cause = errors.Unwrap(cause) {
```
bwplotka, line 473: Hm, do we need a loop?

Why not `errors.Cause`? This is the usual pattern and I don't we need unwrapped cause as second parameter :thinking: It's the code that matters.

The benefit of errors.Cause is to avoid this loop, which looks quite odd to me (first time I see something like this)

web/api/v1/api.go
pracucci: I've used the "parent" nomenclature because it's something "happening before". In this sense, parents come first then children.

codesome: Sounds fine, no issues

web/api/v1/api.go
pracucci: Oh yes sure, done.

PR #7437, opened 2020-06-23
Simplify TestHeadReadWriter_Truncate

codesome:
The `verifyFiles` method was very confusing even though I wrote it myself. I have modified it to be like in `TestHeadReadWriter_Truncate_NoUnsequentialFiles` which was added recently which is much clear to read.

For reviewers: It is easier to read and understand the entire test from top to bottom instead of reading the diff in this PR.

PS: I fixed the test names to match with the updated names for the structs.


beorn7:
I'll review this ASAP. (Wanted to do it today, but on-call etc. blabla.0

tsdb/chunks/head_chunks_test.go
```diff
@@ -188,51 +185,41 @@ func TestHeadReadWriter_Truncate(t *testing.T) {
...
 
-		for i := 1; i <= totalFiles; i++ {
+		for _, i := range remainingFiles {
 			_, ok := hrw.mmappedChunkFiles[i]
-			if i < startIndex {
-				testutil.Equals(t, false, ok)
-			} else {
-				testutil.Equals(t, true, ok)
-			}
+			testutil.Equals(t, true, ok)
```
beorn7, line 205: Isn't `testutil.Assert` better for a bool?

Also, I'd add helpful information here, e.g.:
```Go
testutil.Assert(t, ok, "remaining file %d not in hrw.mmappedChunkFiles", i)
```

tsdb/chunks/head_chunks_test.go
```diff
@@ -246,29 +233,35 @@ func TestHeadReadWriter_Truncate(t *testing.T) {
...
+	testutil.Ok(t, hrw.Truncate(sixthFileMint+1))
+	verifyFiles([]int{6, 7, 8, 9, 10})
 	addChunk()
-	totalFiles++
 
 	// Truncating till current time should not delete the current active file.
-	testutil.Ok(t, hrw.Truncate(int64(timeRange+fileTimeStep)))
-	verifyFiles(2, totalFiles) // One file is the active file and one was newly created.
+	testutil.Ok(t, hrw.Truncate(int64(timeRange+(2*fileTimeStep))))
+	verifyFiles([]int{10, 11}) // One file is the previously active file and one was newly created.
```
beorn7, line 252: Nit: If you use an adverb here ("previously"), you should use one ("currently") in line 250, too.

tsdb/chunks/head_chunks_test.go
```diff
@@ -156,24 +156,21 @@ func TestHeadReadWriter_WriteChunk_Chunk_IterateChunks(t *testing.T) {
...
 		close()
 	}()
 
 	timeRange := 0
 	fileTimeStep := 100
-	totalFiles := 7
-	startIndexAfter1stTruncation, startIndexAfter2ndTruncation := 3, 6
-	filesDeletedAfter1stTruncation, filesDeletedAfter2ndTruncation := 2, 5
-	var timeToTruncate, timeToTruncateAfterRestart int64
+	var thirdFileMint, sixthFileMint int64
```
beorn7, line 173: Throughout the file, I'd capitalize the "t" here to make clear it's a timestamp, e.g. `thirdFileMinT`, but also `minT` instead of `mint`. I was really confused at first. Is "mint" something the creates coins of some kind? https://en.wikipedia.org/wiki/Mint_(facility) Or is it refering to https://en.wikipedia.org/wiki/Mentha ?

tsdb/chunks/head_chunks_test.go
```diff
@@ -188,51 +185,41 @@ func TestHeadReadWriter_Truncate(t *testing.T) {
 		return mint
 	}
 
-	cutFile := func(i int) {
+	cutFile := func(i int) int64 {
```
beorn7, line 188: `cutFile` is now so short that I think it's less mental overhead if you inline it.

beorn7: In any case, the `i` parameter is not used anymore.

tsdb/chunks/head_chunks_test.go
```diff
@@ -188,51 +185,41 @@ func TestHeadReadWriter_Truncate(t *testing.T) {
...
-		}
-	}
-
-	// Cut segments.
-	for i := 1; i <= totalFiles; i++ {
-		cutFile(i)
+		return int64(addChunk())
 	}
 
 	// Verifying the files.
```
beorn7, line 193: This comment is redundant with the name of the function in the next line. I'd just leave it out.

tsdb/chunks/head_chunks_test.go
```diff
@@ -246,29 +233,35 @@ func TestHeadReadWriter_Truncate(t *testing.T) {
...
 // that truncation leaves no unsequential files on disk, mainly under the following case
 // * There is an empty file in between the sequence while the truncation
 //   deletes files only upto a sequence before that (i.e. stops deleting
 //   after it has found a file that is not deletable).
 // This tests https://github.com/prometheus/prometheus/issues/7412 where
 // the truncation used to check all the files for deletion and end up
 // deleting empty files in between and breaking the sequence.
-func TestHeadReadWriter_Truncate_NoUnsequentialFiles(t *testing.T) {
-	hrw, close := testHeadReadWriter(t)
+func TestChunkDiskMapper_Truncate_NoUnsequentialFiles(t *testing.T) {
```
beorn7, line 263: `NoUsequentialFiles` is a double negation, which is difficult to understand. How about `PreservesFileSequence`?

I would also like to suggest a more straightforward doc comment:

TestChunkDiskMapper_Truncate_PreservesFileSequence tests that truncation doesn't poke holes into the file sequence, even if there are empty files in between non-empty files. This test exposes https://github.com/prometheus/prometheus/issues/7412 where the truncation simply deleted all empty files instead of stopping once it encountered a non-empty file.

tsdb/chunks/head_chunks_test.go
```diff
@@ -337,7 +330,7 @@ func TestHeadReadWriter_Truncate_NoUnsequentialFiles(t *testing.T) {
 	verifyFiles([]int{3, 4, 5, 6, 7})
 }
 
-func testHeadReadWriter(t *testing.T) (hrw *ChunkDiskMapper, close func()) {
+func testChunkDiskMapper(t *testing.T) (hrw *ChunkDiskMapper, close func()) {
```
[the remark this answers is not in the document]
codesome: I will change the name and add the doc. Some tests use `hrw.Close()` to restart the chunk disk mapper and use in various ways, so we need it separated from `close` which mainly does the cleanup of the directories.

tsdb/chunks/head_chunks_test.go
codesome: Right!

tsdb/chunks/head_chunks_test.go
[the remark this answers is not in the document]
beorn7: But `hrw.Close()` will still continue to exist, so you can call it separately. But whenever you call  the returned `close` function, you also call `hrw.Close()`, so I'd include it in`close`.

roidelapluie:
Hello @codesome, are you still interested in this? thanks!

codesome:
Yes

tsdb/chunks/head_chunks_test.go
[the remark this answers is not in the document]
codesome: With `t.Cleanup` now, no need to return the close function now.

tsdb/chunks/head_chunks_test.go
codesome: I will change for `thirdFileMint` and `sixthFileMint`, but `mint` and `maxt` is used in way too many places :P I will do it for this file though

codesome: Or leave those `mint` to `minT` changes to a different PR

codesome:
I have addressed the review and fixed the conflicts. Should be good for another review.

PS: "For reviewers: It is easier to read and understand the entire TestHeadReadWriter_Truncate test from top to bottom instead of reading the diff in this PR." still holds :)

bwplotka (approved):
LGTM! :+1:

PR #7438, opened 2020-06-23
Enhanced WAL replay for duplicate series record

codesome:
In this PR I fix a TODO I had put before `// TODO(codesome) Discard old samples and mmapped chunks and use mmap chunks for the new series ID.`

Assertion:
Basically, if there is a duplicate series record in the WAL, it means that the compaction happened in between and the series including all it's samples were compacted before this duplicate record. Hence, we can discard all that we have replayed till now for that series when we come across a duplicate record.

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -1466,7 +1466,11 @@ type memChunk struct {
 
 // OverlapsClosedInterval returns true if the chunk overlaps [mint, maxt].
 func (mc *memChunk) OverlapsClosedInterval(mint, maxt int64) bool {
-	return mc.minTime <= maxt && mint <= mc.maxTime
+	return overlapsClosedInterval(mc.minTime, mc.maxTime, mint, maxt)
+}
+
+func overlapsClosedInterval(mint1, maxt1, mint2, maxt2 int64) bool {
+	return mint1 <= maxt2 && mint2 <= maxt1
 }
 
 type mmappedChunk struct {
@@ -1477,7 +1481,7 @@ type mmappedChunk struct {
 
 // Returns true if the chunk overlaps [mint, maxt].
 func (mc *mmappedChunk) OverlapsClosedInterval(mint, maxt int64) bool {
-	return mc.minTime <= maxt && mint <= mc.maxTime
+	return overlapsClosedInterval(mc.minTime, mc.maxTime, mint, maxt)
 }
 
 type noopSeriesLifecycleCallback struct{}
--- a/tsdb/head_wal.go
+++ b/tsdb/head_wal.go
@@ -18,6 +18,7 @@ import (
 	"math"
 	"runtime"
 	"sync"
+	"time"
 
 	"github.com/go-kit/log/level"
 	"github.com/pkg/errors"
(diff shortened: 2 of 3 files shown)


codesome:
I have simplified it further now compared to what it was in https://github.com/prometheus/prometheus/pull/7229 on a simple assertion mentioned in the PR description. Earlier I was comparing the m-mapped chunks of new series id with the old series id, which made it complex. Now we just keep m-mapped chunks for the series id in the latest record.

roidelapluie:
This now has conflicts

codesome:
I think I will be taking care of this in https://github.com/prometheus/prometheus/pull/7229, but will keep this open till then in case I don't do it there. So, no need to review this PR right now. (I will be getting to snapshotting chunks in a few weeks, caught up with some other stuff)

codesome:
I take back my above comment. Let's do it in this PR itself to keep the review of snapshot PR simpler. I have rebased this and is ready for another review.

roidelapluie:
Hello @bwplotka Are you interested in reviewing this? Thanks!

tsdb/head.go
```diff
@@ -525,27 +525,57 @@ Outer:
...
-					if len(series.mmappedChunks) > 0 {
-						h.updateMinMaxTime(series.minTime(), series.maxTime())
+				if len(series.mmappedChunks) > 0 {
+					if overlapsClosedInterval(
+						series.mmappedChunks[0].minTime,
+						series.mmappedChunks[len(series.mmappedChunks)-1].maxTime,
+						mmc[0].minTime,
+						mmc[len(mmc)-1].maxTime,
+					) {
+						// The m-map chunks for the new series ref overlaps with old m-map chunks.
```
bwplotka, line 553: This should never happen, right? Worth commenting this?

tsdb/head.go
```diff
@@ -525,27 +525,57 @@ Outer:
...
-					series.mmappedChunks = mmappedChunks[series.ref]
+				if h.lastSeriesID.Load() < s.Ref {
+					h.lastSeriesID.Store(s.Ref)
+				}
 
-					h.metrics.chunks.Add(float64(len(series.mmappedChunks)))
-					h.metrics.chunksCreated.Add(float64(len(series.mmappedChunks)))
+				if !created {
+					// There's already a different ref for this series.
+					multiRef[s.Ref] = series.ref
```
bwplotka, line 534: don't we want to `continue` here? :thinking:

bwplotka (commented):
Some comments.

tsdb/head.go
codesome: If there is a different ref already, that means this was a duplicate record (hence from the future). So we need to check for m-mapped chunks from later point of time (that's exactly the enhancement here :) to replace old chunks with new ones).

tsdb/head.go
codesome: In an ideal world this would never happen. But if someone is tinkering around the data dir, this is possible. I will comment that this should not happen unless you are experimenting with something.

bwplotka, line 553: Let's comment that this should never happen

bwplotka (commented):
Some pair review comments (:

codesome:
With some more ordering assumptions that we can make, the code is _much_ simpler now and easy to understand. I have push the changes.

codesome:
But it has some races :( I will fix them

codesome:
Race fixed. Ready for review again.

codesome:
I plan to continue the work on snapshots on shutdown and this PR is a prerequisite for that. I have rebased this PR to remove the conflicts. It is ready for another review now.

tsdb/head.go
```diff
@@ -680,38 +680,73 @@ func (h *Head) loadWAL(r *wal.Reader, multiRef map[uint64]uint64, mmappedChunks
...
-				} else {
-					// TODO(codesome) Discard old samples and mmapped chunks and use mmap chunks for the new series ID.
+				multiRef[walSeries.Ref] = mSeries.ref
 
-					// There's already a different ref for this series.
-					multiRef[s.Ref] = series.ref
+				// It is possible that some old sample is being processed in processWALSamples that could
+				// could cause race below. So we wait for the goroutine to empty input the buffer.
+				idx := mSeries.ref % uint64(n)
+				inputs[idx] <- []record.RefSample{} // To ensure that all the old samples are processed.
```
bwplotka, line 718: This comment seems to be a duplicate of one 2 lines above.

tsdb/head.go
```diff
@@ -680,38 +680,73 @@ func (h *Head) loadWAL(r *wal.Reader, multiRef map[uint64]uint64, mmappedChunks
...
-					if len(series.mmappedChunks) > 0 {
-						h.updateMinMaxTime(series.minTime(), series.maxTime())
-					}
-				} else {
-					// TODO(codesome) Discard old samples and mmapped chunks and use mmap chunks for the new series ID.
+				multiRef[walSeries.Ref] = mSeries.ref
 
-					// There's already a different ref for this series.
-					multiRef[s.Ref] = series.ref
+				// It is possible that some old sample is being processed in processWALSamples that could
```
bwplotka, line 715:
```suggestion
				// It is possible that some old sample is being processed in processWALSamples that
```

bwplotka: Can we move this comment to exactly before `for` loop?

tsdb/head.go
```diff
@@ -680,38 +680,73 @@ func (h *Head) loadWAL(r *wal.Reader, multiRef map[uint64]uint64, mmappedChunks
...
+				// could cause race below. So we wait for the goroutine to empty input the buffer.
+				idx := mSeries.ref % uint64(n)
+				inputs[idx] <- []record.RefSample{} // To ensure that all the old samples are processed.
+				for len(inputs[idx]) != 0 {
```
bwplotka, line 719: This is first time I see this pattern of checking len on channel, nice. Isn't that racy, when writer is not fast enough to give elements to input[idx] chan and buffer is read faster than writer writes?

bwplotka (approved):
It looks good, just minor nits and one question. Otherwise LGTM!

tsdb/head.go
codesome: The writer is giving elements to `input[idx]` in the same loop, and it is serial to this line. So when we are here, there is nothing giving elements to any channels in `input` slice. So there won't be any read-write race, there is only reading.

tsdb/head.go
codesome: I will move this just before `inputs[idx] <- []record.RefSample{}` (which is just above `for` loop), since that is an important part of making sure all old samples are processed.

codesome:
I will wait for https://github.com/prometheus/prometheus/pull/9147 to be merged before this.

tsdb/head_wal.go
```diff
@@ -181,38 +182,74 @@ func (h *Head) loadWAL(r *wal.Reader, multiRef map[uint64]uint64, mmappedChunks
...
+				multiRef[walSeries.Ref] = mSeries.ref
 
-					// There's already a different ref for this series.
-					multiRef[s.Ref] = series.ref
+				idx := mSeries.ref % uint64(n)
+				// It is possible that some old sample is being processed in processWALSamples that
+				// could cause race below. So we wait for the goroutine to empty input the buffer and finish
+				// processing all old samples after emptying the buffer.
+				inputs[idx] <- []record.RefSample{}
+				for len(inputs[idx]) != 0 {
```
bboreham, line 222: Both of these lines can deadlock; see #9169 for one case.

PR #7444, opened 2020-06-23
Add Bartek as remote read maintainer.

cstyan:
A few times now issues or PRs for remote read have been opened and Chris and I have been pinged as the `storage/remote` maintainers. As I've never extensively used or worked on remote read I feel I don't have proper context to give those issues/PRs the attention they need.

I'm proposing we add Bartek as a maintainer, splitting remote read and remote write maintainership.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -7,7 +7,7 @@
 * `documentation`
   * `prometheus-mixin`: @beorn7
 * `storage`
-  * `remote`: @csmarchbanks, @cstyan
+  * `remote`: @csmarchbanks, @cstyan, @bwplotka
 * `tsdb`: @codesome, @krasi-georgiev
 * `web`
   * `ui`: @juliusv


MAINTAINERS.md
```diff
@@ -7,7 +7,9 @@
 * `documentation`
   * `prometheus-mixin`: @beorn7
 * `storage`
-  * `remote`: @csmarchbanks, @cstyan
+  * `remote`
```
brian-brazil, line 10: This is organised by directory, so we shouldn't be going more granularly than that.

cstyan: Is there a reason it needs to be that way? There's nothing stopping me from splitting remote read and write into their own subdirectories within `storage/remote`. 

Listing people as maintainers of either remote read or remote write (or both) is more useful to those outside the project than just listing all three of us as maintainers of `storage/remote` when I personally am not going to be of any use to someone looking for a review on a remote read PR.

beorn7:
I don't think the maintainer structure _has_ to be organized by directory, but it's nice if it works that way. Let's do it as @bwplotka says, and if you feel over time that's not working well, we can reconsider and think about a not-quite-as-strictly-directory-wise organization.

MAINTAINERS.md
brian-brazil: That's the way I laid it out as it made sense to me. Any change to one likely affects the other, and we can route amongst ourselves anyway.

csmarchbanks:
I agree that adding @bwplotka as a remote maintainer is a reasonable iteration. If the routing continues to not work well we can consider additional measures.

I am also wondering if by having different maintainers for read and write, perhaps they *should* be in different packages.

bwplotka:
Do you mind @cstyan to put me for full `remote` then? (:

cstyan:
Updated.

Given that this is not the first time this change has come up, and listing maintainers by directory vs by subsystem (especially when there's nothing stopping us from refactoring `storage/remote` into `storage/remote/read` and `storage/remote/write`) is a single line change in a markdown file, I'm honestly surprised we want to wait until a similar situation arises again before changing this.

brian-brazil:
:+1: 

It's a bit more subtle, as part of this is about where a new contributor would send a PR and they may not know where all of our code is located. Even as it stands, remote storage is dotted around several places. Going with directories doesn't seem like a major burden in practice, and if our code is coupled in ways that are suboptimal then I'd suggest we first consider that.

PR #7445, opened 2020-06-24
Add two unit tests

hs0210:
This PR is about adding a unit test for storage/buffer.go and a unit test for tsdb/index/index.go

### DIFF
--- a/storage/buffer_test.go
+++ b/storage/buffer_test.go
@@ -161,6 +161,43 @@ func TestBufferedSeriesIteratorNoBadAt(t *testing.T) {
 	it.Next()
 }
 
+func TestBufferReduceDelta(t *testing.T) {
+	for _, tc := range []struct {
+		input    int64
+		r        *sampleRing
+		expexted bool
+	}{
+		{
+			input: 2,
+			r: &sampleRing{
+				delta: 1,
+			},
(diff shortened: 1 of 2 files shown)


tsdb/index/index_test.go
```diff
@@ -553,3 +553,14 @@ func TestSymbols(t *testing.T) {
 	}
 	testutil.Ok(t, iter.Err())
 }
+
+func TestLabelNames(t *testing.T) {
```
codesome, line 557: We are already testing label names in `TestDB_LabelNames`. This is not required.

storage/buffer_test.go
```diff
@@ -161,6 +161,43 @@ func TestBufferedSeriesIteratorNoBadAt(t *testing.T) {
 	it.Next()
 }
 
+func TestBufferReduceDelta(t *testing.T) {
```
codesome, line 164: We test the reduceDelta indirectly via `TestBufferedSeriesIterator`. IMO, testing all the individual internal components is not a good idea when those already get tested via other exported methods unless the internal component is huge to deserve a test for itself.

hs0210: Thanks for your review, I'll close it.

PR #7447, opened 2020-06-24
web/ui: bump jQuery to 3.5.1 for the legacy UI

### LINKED ISSUE #7446: Security issue on JQuery with version < 3.5
Hi all,

I use Prometheus in a monitoring project for my Company. The security team opened a vulnerability issue because on 9090 port they see a JQuery version below 3.5. They reported the following security report:
https://www.tenable.com/plugins/nessus/136929
that suggests upgrading to 3.5

If you follow the link inside you arrive here:
https://blog.jquery.com/2020/04/10/jquery-3-5-0-released/

If I well understand the problem, it is related to the use of a special HTML tag. It's clear that to fix the security issue you need to upgrade to 3.5 or just report that you don't use that tag.

Now looking at the commit history there is this commit, where it seems the fix is present.
https://github.com/prometheus/prometheus/commit/c7d9516695c51cdb2185166e34c8190feca0fdf8

the commit was included 26 days ago. I tried the Prometheus 2.19.1 (built 6 days ago) and I verified it includes JQuery 3.3.1. 

To verify this I did:
1. install Prometheus 2.19.1
2. start the server
3. Access to localhost:9090 with Chrome
4. Right-click->Inspect and click on Console tab
5. Enter this command: console.log(jQuery.fn.jquery);

it reports 3.3.1. My question is: why I do not have evidence that Prometheus 2.19.1 uses JQuery 3.5? If JQuery 3.5 is not included in 2.19.1, can you tell me whether the security exposure affects Prometheus or not? If so, when do you think it will be solved?
Thank you in advance for help.

simonpasquier:
jQuery prior to 3.4.0 is affected by an Object.prototype pollution
vulnerability (CVE-2019-11358). Even though our code doesn't seem to be
vulnerable to the issue, lets upgrade to the latest jQuery release so we
don't have to bother.

cc @juliusv I've tested quickly the UI and couldn't find any problem. I've checked the jQuery release notes for 3.4.0 and 3.5.0: I couldn't spot any compatibility issue that might affect us. 

Closes #7446


juliusv:
:+1: Thanks @simonpasquier!

PR #7456, opened 2020-06-25
Web: Scroll hash-fragment links with navbar height

### LINKED ISSUE #7434: prometheus web UI: Skip nav header when following #hash-fragment-links
## Proposal

If I click a hash-fragment link in the prometheus targets web UI, currently the header disappears behind the header.

http://localhost:9090/targets#job-domain_exporter

It would be nice to have the nav header not occlude the link. This would make it easier to link to parts of the page and see the context.

Major browsers seem to have a way to solve this with CSS:

```css
html {
    scroll-padding-top: 56px;
}
```

Browser support is pretty good: https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-padding-top.

I applied this the other day to https://graphviz.org/ which was having the same problem, and it's working well.

mhansen:
Previously, hash-fragment links like this:
http://mark-t510:9090/targets#job-alertmanager

Would scroll to have the header at the top, obscured by the nav bar.

Tested in both old and new UIs.

Fixes #7434

### DIFF
--- a/web/ui/react-app/src/App.css
+++ b/web/ui/react-app/src/App.css
@@ -1,3 +1,9 @@
+html {
+  /* https://github.com/prometheus/prometheus/issues/7434 */
+  /* Scroll to hash-fragment-links counting the fixed navbar 40px tall with 16px padding */
+  scroll-padding-top: 56px;
+}
+
 .panel {
   margin-bottom: 20px;
 }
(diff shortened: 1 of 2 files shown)


juliusv:
Thanks! The old `paddingTop: 70` was still intentional, as we want to have some spacing under the navbar as well, to not have the content stick so close to it.

mhansen:
Thanks for the review. I've reverted the 70px --> 56px change. It'd be good to clear up that comment at some stage in the future (I think it's out of date if there new rationale is to add extra padding) but don't want to block this change on that.

juliusv:
:+1: Thanks!

Btw. this made me discover another bug: https://github.com/prometheus/prometheus/issues/7473

mhansen:
Neat. Thank you!

PR #7461, opened 2020-06-25
Remove duplicate test in labels_test.go

csmarchbanks:
The logic in this test is covered by the test with the same name at https://github.com/prometheus/prometheus/blob/master/pkg/labels/labels_test.go#L371. Having two tests with the same name is causing builds to fail.


brian-brazil:
:+1:

PR #7463, opened 2020-06-25
Support time range in promtool query labels

yeya24:

This pr adds `start` and `end` flags in `promtool query labels` command to support time range query.


cmd/promtool/main.go
```diff
@@ -549,10 +551,35 @@ func QueryLabels(url *url.URL, name string, p printer) int {
 		return 1
 	}
 
+	var (
+		minTime = time.Now().Add(-9999 * time.Hour)
+		maxTime = time.Now().Add(9999 * time.Hour)
+	)
+
+	var stime, etime time.Time
```
bwplotka, line 559: you can merge those `vars` (:

cmd/promtool/main.go
```diff
@@ -549,10 +551,35 @@ func QueryLabels(url *url.URL, name string, p printer) int {
...
+
+	if start == "" {
+		stime = minTime
```
bwplotka, line 562: what about 
```
stime := minTime
if start != "" {
// change stime
}
```
and same below?

go.mod
```diff
@@ -38,7 +38,7 @@ require (
 	github.com/hashicorp/serf v0.9.0 // indirect
 	github.com/influxdata/influxdb v1.8.0
 	github.com/jpillora/backoff v1.0.0 // indirect
-	github.com/json-iterator/go v1.1.9
+	github.com/json-iterator/go v1.1.10
```
bwplotka, line 41: ideally we upgrade those in separate PR. Or at least you can mention in commit `update client_golang` (:

yeya24: Changelog added.

cmd/promtool/main.go
yeya24: Thanks, I like this way. Much cleaner than before.

go.mod
beorn7: Hmm, I don't think we should mention it in the CHANGELOG. I think @bwplotka was referring to the commit description. Furthermore, this is about json-iterator, not client_golang. But I guess it was _triggered_ by the client_golang update (which is normal with the way Go modules work).

So I'd say just add "Updated prometheus/client_golang and json-iterator/go" to the commit description (not the changelog).

bwplotka: Yes, I meant commit message (:

yeya24: Sorry I misunderstood :(

PR updated, PTAL.

PR #7069, opened 2020-03-29
tsdb: Added ChunkQueryable implementations to db; unified MergeSeriesSets and vertical to single struct.

bwplotka:
This PR adds support for all block Querier to operate on chunk and allow chunk iterating if needed. To do so, we reused all structs done before for merging queries (needed by fanout). This removes lots of code that was duplicated for same work: Querying, fan-out (e.g to multiple block) and merging results.

## Changes

* Added single SeriesEntry / ChunkEntry for all series implementations.
* Unified all vertical, and non vertical for compact and querying to single
merge series / chunk sets by reusing VerticalSeriesMergeFunc for overlapping algorithm (same logic as before)
* Added block (Base/Chunk/)Querier for block querying. We then use populateAndTomb(Base/Chunk/) to iterate over chunks or samples.
* Refactored endpoint tests and querier tests to include subtests.

Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>

PR #7471, opened 2020-06-26
tombstones: Fixed Add method in order to support trimming time series; Simplified the algo.

bwplotka:
We need that in order to simplify min, max trimming of chunks for Chunk Iterator. 

Blocks: https://github.com/prometheus/prometheus/pull/7069

cc @codesome @gouthamve @brian-brazil

### DIFF
--- a/tsdb/tombstones/tombstones.go
+++ b/tsdb/tombstones/tombstones.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sort"
 	"sync"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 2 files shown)


tsdb/tombstones/tombstones.go
```diff
@@ -286,47 +287,32 @@ type Intervals []Interval
...
+	// Find min and max indexes of intervals that overlap with the new interval.
+	// Intervals are closed [t1, t2] and t is discreet, so if neighbour intervals are 1 step difference
+	// to the new one, we can merge those together.
+	mini := sort.Search(len(in), func(i int) bool { return in[i].Maxt >= n.Mint-1 })
+	if mini == len(in) {
+		return append(in, n)
+	}
 
-			return newRange
+	maxi := sort.Search(len(in)-mini, func(i int) bool { return in[mini+i].Mint > n.Maxt+1 })
```
codesome, line 302: What am I missing here :thinking: I would have expected `>=` here

tsdb/tombstones/tombstones.go
```diff
@@ -286,47 +287,32 @@ type Intervals []Interval
...
+	if maxi == 0 {
+		if mini == 0 {
+			return append(Intervals{n}, in...)
 		}
+		return append(in[:mini], append(Intervals{n}, in[mini:]...)...)
```
codesome, line 307: Shouldn't mini be merged with n?

tsdb/tombstones/tombstones.go
bwplotka: We don't want that, since if the ranges are adjacent we merge them. This search ensures we have at least 1-size gap (:

tsdb/tombstones/tombstones.go
bwplotka: What do you mean? it is merged, just we need prepend.

tsdb/tombstones/tombstones_test.go
```diff
@@ -80,8 +81,13 @@ func TestAddingNewIntervals(t *testing.T) {
 		},
 		{
 			exist: Intervals{{1, 10}, {12, 20}, {25, 30}},
-			new:   Interval{21, 23},
-			exp:   Intervals{{1, 10}, {12, 23}, {25, 30}},
+			new:   Interval{21, 25},
+			exp:   Intervals{{1, 10}, {12, 30}},
```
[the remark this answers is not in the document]
bwplotka: That's the critical point. Below as you see 22,23 means we cannot merge those ranges, whereas above we can.

This was the logic before this PR as well, just never shown in tests (:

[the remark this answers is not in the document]
codesome: I mean we have the exact same test case below, see line 98 :)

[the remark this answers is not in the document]
bwplotka: Removed, there was exact duplicate indeed

tsdb/tombstones/tombstones.go
brian-brazil, line 302: Why are you binary searching here? Shouldn't it be faster to talk from mini?

bwplotka: That is what I am doing with `len(in)-mini,` (:

bwplotka:
> What do you mean by edge trimming? I'm not seeing any graphs that'd have edge which could be pruned. I think you mean coalescing.

I literally mean the last 4 test cases. So if someone request time range 20, 30 and no deletions, in separate PR we transform this into literally [-inf, 19], [30,inf+] deletion. This works better for readability as well as efficiency.

tsdb/tombstones/tombstones.go
brian-brazil: That's a binary search, not a linear walk.

brian-brazil:
> I literally mean the last 4 test cases.

Could you use the term "coalescing" rather than inventing a term which confusingly appears to be related to graph theory? I see no vertexes for which there could be edges, nor do I see any trimming.

tsdb/tombstones/tombstones.go
bwplotka: Yes, however binary search for already limited range: [mini, len(in)] so works better / not worse than linear idea. Am I missing anything here? (:

brian-brazil: That depends on how many tombstones you're expecting to exist.

bwplotka:
Ah, good point, yes (: 

Edge to me is also `End or Start` (: Maybe just "trimming" is enough then? `coalescing` is not part of this change, it was there before.

brian-brazil:
Trimming means making something shorter, this change makes intervals longer. "coalescing" or "merging" are the terms you are looking for.

bwplotka:
Ah intervals yes, but the series shorter ((: Fair point.

tsdb/tombstones/tombstones.go
bwplotka: True, but `Search` is very clean, so looks like it could be fine here, would you be happy to leave it? (:

brian-brazil: I'm okay with it.

PR #7472, opened 2020-06-26
rules: Use narrower interface for rule manager loading of for state

brancz:
To load ALERT_FOR_STATE only `storage.Queryable` interface is required,
so this patch uses this narrower interface for to perform this.

@brian-brazil @bwplotka @codesome @gouthamve

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -379,7 +379,7 @@ func main() {
 
 		ruleManager = rules.NewManager(&rules.ManagerOptions{
 			Appendable:      fanoutStorage,
-			TSDB:            localStorage,
+			Queryable:       localStorage,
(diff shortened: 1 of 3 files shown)


brancz:
If people want I'm also happy to rename the `TSDB` struct field, as `Queryable` arguably might make more sense now. Let me know.

brian-brazil:
:+1: 

Renaming makes sense to me.

roidelapluie:
Is it the purpose to make it importable? Because currently we are not resilient to nil interfaces.

brancz:
> Is it the purpose to make it importable? Because currently we are not resilient to nil interfaces.

It's so I can play around with alternative implementations and not have to implement the whole interface, and Prometheus happens to also only need this narrower interface, which is generally good practice.

PR #7476, opened 2020-06-29
Add unit test case for func State in alerting.go

fenggw-fnst:
Signed-off-by: Guangwen Feng <fenggw-fnst@cn.fujitsu.com>


brian-brazil:
Thanks!

PR #7480, opened 2020-06-29
Add group() aggregator

### LINKED ISSUE #7400: `any()` aggregation operator
## `any()` aggregation operator

It would be useful if we had an aggregation operator that explicitly states "i don't care which value you pick, just pick one". This could have a marginal performance advantage over choosing an arbitrary aggregation, but the real benefit is semantic. when you put, for example, `min()` in a query, the reader assumes that this particular aggregation was chosen for a reason, but in some cases it doesn't matter, like when multiple identical or non-overlapping series are being aggregated. picking an arbitrary aggregation adds meaningless specificity and can make it harder to understand complex queries, while something like `any()` would remove any false meaning from the expression.

roidelapluie:
Fix #7400 
cc @alice-sawatzky @beorn7 

### DIFF
--- a/docs/querying/operators.md
+++ b/docs/querying/operators.md
@@ -187,7 +187,7 @@ vector of fewer elements with aggregated values:
 * `min` (select minimum over dimensions)
 * `max` (select maximum over dimensions)
 * `avg` (calculate the average over dimensions)
-* `group` (returns 1 over dimensions)
+* `group` (all values in the resulting vector are 1)
 * `stddev` (calculate population standard deviation over dimensions)
 * `stdvar` (calculate population standard variance over dimensions)
 * `count` (count number of elements in the vector)
--- a/docs/querying/operators.md
+++ b/docs/querying/operators.md
@@ -187,6 +187,7 @@ vector of fewer elements with aggregated values:
 * `min` (select minimum over dimensions)
 * `max` (select maximum over dimensions)
 * `avg` (calculate the average over dimensions)
+* `group` (returns 1 over dimensions)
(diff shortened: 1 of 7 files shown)


promql/engine.go
```diff
@@ -2004,6 +2007,8 @@ func (ev *evaluator) aggregation(op parser.ItemType, grouping []string, without
 			group.groupCount++
 			group.mean += (s.V - group.mean) / float64(group.groupCount)
 
+		case parser.GROUP:
```
brian-brazil, line 2010: I'd add a comment to make it clear this purposefully does nothing.

beorn7:
I like it. Thank you very much.

Could you also add documentation?

roidelapluie:
> I like it. Thank you very much.
> 
> Could you also update the documentation?

It is updated, what else would you add ?

docs/querying/operators.md
```diff
@@ -187,6 +187,7 @@ vector of fewer elements with aggregated values:
 * `min` (select minimum over dimensions)
 * `max` (select maximum over dimensions)
 * `avg` (calculate the average over dimensions)
+* `group` (returns 1 over dimensions)
```
roidelapluie, line 190: @beorn7

beorn7: Sorry, missed that. Perhaps it's a bit dense? Even I needed to think a bit about what "returns 1 over dimensions" means.

Perhaps "(all values in the resulting vector are 1)"?

roidelapluie: Adapted

PR #7490, opened 2020-06-30
PromQL: Add a new random() function.

roidelapluie:
This function is primilarily intended for demos, simulation, exemples.

Randomness has other uses. Notably, I have used in real
situation a very complex variation of this with time() to hide the
source of some metrics by slightly altering their actual values.

Note that Prometheus has all the necessary methods to get variants of
this, e.g. ceil(vector(10*random()))

### DIFF
--- a/docs/querying/functions.md
+++ b/docs/querying/functions.md
@@ -325,6 +325,11 @@ regression](https://en.wikipedia.org/wiki/Simple_linear_regression).
 
 `predict_linear` should only be used with gauges.
 
+## `random()`
+
+`random()` returns a pseudo-random scalar in [0.0,1.0). This
+function returns different values in consecutive identical queries.
+
 ## `rate()`
 
 `rate(v range-vector)` calculates the per-second average rate of increase of the
--- a/promql/functions.go
+++ b/promql/functions.go
@@ -15,6 +15,7 @@ package promql
 
 import (
 	"math"
+	"math/rand"
 	"regexp"
 	"sort"
 	"strconv"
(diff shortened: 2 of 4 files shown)


brian-brazil:
I don't think we should be adding non-determinism to PromQL, if I run the same query on the same data I should get the same result. This also seems quite a bit away from time series monitoring.

> Notably, I have used in real situation a very complex variation of this with time() to hide the source of some metrics by slightly altering their actual values

Data which requires anonymisation probably shouldn't be in Prometheus in the first place from a privacy standpoint.

roidelapluie:
> I don't think we should be adding non-determinism to PromQL, if I run the same query on the same data I should get the same result. This also seems quite a bit away from time series monitoring.

Would it be acceptable to seed rand() with the ts to make it alsmost deterministic?

> Data which requires anonymisation probably shouldn't be in Prometheus in the first place from a privacy standpoint.

This was not for privacy reason but more "commercial". Being able to demo something with close-to-reality data without identification of the source of the metrics.

brian-brazil:
> Would it be acceptable to seed rand() with the ts to make it alsmost deterministic?

If that's good enough then you can already do something like (picking the first parameters off https://en.wikipedia.org/wiki/Linear_congruential_generator) `(time()*1664525 + 1013904223) % 2^32`

roidelapluie:
yeah we ended up with avg_over_time(vector((time() % 1000 - (time()-500) % 500) )[10m:]) / 500 which was better than nothing

roidelapluie:
1

Issue #7443, opened 2020-06-23
Increment WAL corruption metric if corruption is see during checkpointing


codesome:
Doing this (see title) will help users notice the corruption early on

Harkishen-Singh:
Will take this up!

PR #7491, opened 2020-06-30
Increments wal corruption metric on error during checkpointing

Harkishen-Singh:

Fixes: #7443

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -837,6 +837,7 @@ func (h *Head) Truncate(mint int64) (err error) {
 	h.metrics.checkpointCreationTotal.Inc()
 	if _, err = wal.Checkpoint(h.wal, first, last, keep, mint); err != nil {
 		h.metrics.checkpointCreationFail.Inc()
+		h.metrics.walCorruptionsTotal.Inc()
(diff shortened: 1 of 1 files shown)


roidelapluie:
should we have a labe on those metrics? We have WAL, MMAP-head, checkpoints..

brian-brazil:
That's probably over-complicating things. A WAL corruption is a WAL corruption.

tsdb/head.go
```diff
@@ -837,6 +837,7 @@ func (h *Head) Truncate(mint int64) (err error) {
 	h.metrics.checkpointCreationTotal.Inc()
 	if _, err = wal.Checkpoint(h.wal, first, last, keep, mint); err != nil {
 		h.metrics.checkpointCreationFail.Inc()
+		h.metrics.walCorruptionsTotal.Inc()
```
codesome, line 840: Checkpointing could fail for various reasons, we should increment only when it's a WAL corruption. There is `wal.CorruptionErr` which you could check for (don't forget `errors.Cause`).

Harkishen-Singh:
@codesome done! I hope that's what we want.

PR #7493, opened 2020-07-01
Add unit test for util/testutil/testing.go

Hellcatlk:
Signed-off-by: zouyu <zouy.fnst@cn.fujitsu.com>


brian-brazil:
I don't think we need tests for our unittests, if they break we'll know from all the failing tests.

Hellcatlk:
Thanks for your comment, I will close this PR.

PR #7494, opened 2020-07-01
OpenStack SD: Add availability config option, to choose endpoint type

stefreak:
In some environments Prometheus must query OpenStack via an alternative
endpoint type (gophercloud calls this `availability`.

Took over from Dennis. This PR replaces https://github.com/prometheus/prometheus/pull/4607

### DIFF
--- a/discovery/openstack/openstack.go
+++ b/discovery/openstack/openstack.go
@@ -15,6 +15,7 @@ package openstack
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"time"
 
@@ -34,6 +35,7 @@ import (
 var DefaultSDConfig = SDConfig{
 	Port:            80,
 	RefreshInterval: model.Duration(60 * time.Second),
+	Availability:    string(gophercloud.AvailabilityPublic),
 }
 
 // SDConfig is the configuration for OpenStack based service discovery.
@@ -55,6 +57,7 @@ type SDConfig struct {
 	Port                        int                   `yaml:"port"`
 	AllTenants                  bool                  `yaml:"all_tenants,omitempty"`
 	TLSConfig                   config_util.TLSConfig `yaml:"tls_config,omitempty"`
+	Availability                string                `yaml:"availability,omitempty"`
 }
 
 // Role is the role of the target in OpenStack.
@@ -91,12 +94,19 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err != nil {
 		return err
 	}
+
+	availability := gophercloud.Availability(c.Availability)
+	if availability != gophercloud.AvailabilityPublic && availability != gophercloud.AvailabilityInternal && availability != gophercloud.AvailabilityAdmin {
+		return fmt.Errorf("unknown availability %s, must be one of %s, %s or %s", c.Availability, gophercloud.AvailabilityAdmin, gophercloud.AvailabilityInternal, gophercloud.AvailabilityPublic)
+	}
+
 	if c.Role == "" {
 		return errors.New("role missing (one of: instance, hypervisor)")
 	}
 	if c.Region == "" {
 		return errors.New("openstack SD configuration requires a region")
 	}
+
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


docs/configuration/configuration.md
```diff
@@ -773,6 +773,9 @@ region: <string>
 # instead be specified in the relabeling rule.
 [ port: <int> | default = 80 ]
 
+# The endpoint type to connect to. Must be one of public, admin or internal.
+[ availability: <string> | default = public ]
```
brian-brazil, line 777: "public"

discovery/openstack/openstack.go
```diff
@@ -91,12 +94,19 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err != nil {
 		return err
 	}
+
+	availability := gophercloud.Availability(c.Availability)
+	if availability != gophercloud.AvailabilityPublic && availability != gophercloud.AvailabilityInternal && availability != gophercloud.AvailabilityAdmin {
```
brian-brazil, line 99: Do we have to hardcode this here, or is there a function we can use? Is it likely that this will change?

stefreak: There is no function in the gophercloud package at least. this is very unlikely to change – it denotes wether the api endpoint is reachable for the public or only within the cloud overlay of the openstack deployment.

docs/configuration/configuration.md
stefreak:
```suggestion
[ availability: <string> | default = "public" ]
```

discovery/openstack/openstack.go
stefreak: Alternatively, we can skip validation here, and allow any string. In my opinion it is nice though to protect against typos in the configuration validation already, instead of having trouble to investigate follow up errors (no endpoint found). I don't see a use case for more values than the defined constants in gophercloud

docs/configuration/configuration.md
```diff
@@ -773,6 +773,9 @@ region: <string>
 # instead be specified in the relabeling rule.
 [ port: <int> | default = 80 ]
 
+# The endpoint type to connect to. Must be one of public, admin or internal.
```
brian-brazil, line 776: If this is called availability, why do the docs say it's an endpoint type?

discovery/openstack/openstack.go
brian-brazil: Okay then. This would be cleaner with a select and using plain strings.

stefreak: I tried to incorporate your suggestions. is it fine now?

docs/configuration/configuration.md
stefreak: I tried to incorporate your suggestions. is it fine now?

discovery/openstack/openstack.go
```diff
@@ -91,12 +94,20 @@ func (c *SDConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err != nil {
 		return err
 	}
+
+	switch c.Availability {
+	case string(gophercloud.AvailabilityPublic), string(gophercloud.AvailabilityInternal), string(gophercloud.AvailabilityAdmin):
+	default:
+		return fmt.Errorf("unknown availability %s, must be one of %s, %s or %s", c.Availability, gophercloud.AvailabilityAdmin, gophercloud.AvailabilityInternal, gophercloud.AvailabilityPublic)
```
stefreak, line 101: @brian-brazil or did you mean you'd like this more:
```suggestion
	case "public", "internal", "admin":
	default:
		return fmt.Errorf("unknown availability %s, must be one of admin, internal or public", c.Availability)
```

brian-brazil: Yes, like that. All the indirection makes it difficult to read - I don't want to have to check an entire other repository to see what the strings are.

discovery/openstack/openstack.go
```diff
@@ -34,6 +35,7 @@ import (
 var DefaultSDConfig = SDConfig{
 	Port:            80,
 	RefreshInterval: model.Duration(60 * time.Second),
+	Availability:    string(gophercloud.AvailabilityPublic),
```
brian-brazil, line 38: This one too.

stefreak:
```suggestion
	Availability:    "public",
```

discovery/openstack/openstack.go
stefreak: Much quicker to read now, thank you for the suggestion.

stefreak:
@brian-brazil I believe the build job got stuck. Is there a way for me to cancel and restart it, or can only maintainers do that?

brian-brazil:
Thanks!

stefreak:
Thank you :)

PR #7495, opened 2020-07-01
Change rule index numbering from 0 to n+1 on errors found in rule files

mikejoh:
This PR changes the rule index number from starting with `1` instead of  `0` on reported errors with `promtool`.

Given this rule file (`service-alerts.yml`) with an error on rule `2`:
```
groups:
  - name: serivce.rules
    rules:
      - alert: ServiceDown
        expr: service_up == 0
        for: 5m
        labels:
          severity: critical
        annotations:
          summary: "Service down (instance {{ $labels.instance }})"
          description: "Service down\n  VALUE = {{ $value }}\n  LABELS: {{ $labels }}"

      - alert: ServiceAlertExample
        expr: example(wrong_thing_here)
        for: 5m
        labels:
          severity: critical
        annotations:
          summary: "A summary"
          description: "A description"
``` 

Running `promtool check rules service-alerts.yml` outputs the following:
```
Checking service-alerts.yml
  FAILED:
service-alerts.yml: 14:15: group "serivce.rules", rule 1, "ServiceAlertExample": could not parse expression: 1:1: parse error: unknown function with name "example"
```
and with this minor fix we'll instead get the following output:
```
Checking service-alerts.yml
  FAILED:
service-alerts.yml: 14:15: group "serivce.rules", rule 2, "ServiceAlertExample": could not parse expression: 1:1: parse error: unknown function with name "example"
```

### DIFF
--- a/pkg/rulefmt/rulefmt.go
+++ b/pkg/rulefmt/rulefmt.go
@@ -90,7 +90,7 @@ func (g *RuleGroups) Validate(node ruleGroups) (errs []error) {
 				}
 				errs = append(errs, &Error{
 					Group:    g.Name,
-					Rule:     i,
+					Rule:     i+1,
 					RuleName: ruleName.Value,
 					Err:      node,
 				})
--- a/pkg/rulefmt/rulefmt.go
+++ b/pkg/rulefmt/rulefmt.go
@@ -90,7 +90,7 @@ func (g *RuleGroups) Validate(node ruleGroups) (errs []error) {
 				}
 				errs = append(errs, &Error{
 					Group:    g.Name,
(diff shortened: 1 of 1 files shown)


brian-brazil:
Hmm, is there any where else we expose indexing to users that would also need updating?

mikejoh:
@brian-brazil Oh, right, good question. I should've digged a bit more.. i can definitely do that.

mikejoh:
@brian-brazil I have checked a bit more, didn't find any more places where we expose this in the same way as here, atleast not in `promtool` and related packges.

brian-brazil:
Thanks!

Issue #5547, opened 2019-05-08
"api/series" uses too much memory


vitkovskii:
## Bug Report
**What did you do?**
Send series query like: `/api/v1/series?match[]={le!=""}`

**What did you expect to see?**
Well, this is a very heavy query, so I expect Prometheus to hit timeout or other limits.

**What did you see instead? Under which circumstances?**
The query is executing for ~30 mins until Prometheus is OOM killed.
Normally Prometheus use ~25Gb of memory. This query forces to use more than 90Gb. 

Here is memory usage graph:
![image](https://user-images.githubusercontent.com/2269880/57389994-0866e300-71c4-11e9-976f-a748772eb1ec.png)

Disk read:
![image](https://user-images.githubusercontent.com/2269880/57390208-7c08f000-71c4-11e9-95f3-69b33805e778.png)

**Environment**
We deploy Prometheus to Kubernetes using Prometheus Operator

@brian-brazil

brian-brazil:
What version of Prometheus is this?

d-ulyanov:
`2.9.1` build with `go1.12.4`

brian-brazil:
How many series does that match, and how many do you have in total? Have you tried 2.9.2?

vitkovskii:
@brian-brazil We don't know how many metrics match the source query (because it can't finish). But if pass start and end(100 seconds range): `/api/v1/series?match[]={le!=""}&start=1557303607&end=1557303707`. The query will:
1) Run for ~20 sec for the first byte;
2) Additionally, run for ~2 mins 20 sec to output `497Mb` of data;
3) Return the result with `1357676` metrics.

That's for only 100 seconds range. If start/end is omitted query will run for all range (2 weeks). I assume that the source query won't return much more metrics, but it scans many metrics(`le!=""`) for a long time range. And where is no memory/time limits for the query.

brian-brazil:
Sounds like what I fixed in https://github.com/prometheus/tsdb/pull/595, which will be in 2.10

d-ulyanov:
Hey, @brian-brazil
It seems that `/series` endpoint implementation has no contexts propagation, maybe better to add ctx there and add timeouts? What do you think?

brian-brazil:
Given how the TSDB is currently implemented, it wouldn't make a difference.

zemek:
we're also experiencing this issue in 2.15.2

mateuszlitwin:
[...]

PR #7496, opened 2020-07-01
6178 matchers with label values

eklockare:
This PR is for issue #6178

This adds the ability to pass the ´match[]´ parameter to ´/api/v1/label/<label_name>/values´ endpoint.

Example:
> curl -s 'http://localhost:9090/api/v1/label/code/values?match[]=promhttp_metric_handler_requests_total%7Bcode%3D%22200%22%7D'

( url decoded: promhttp_metric_handler_requests_total{code="200"}  )

```
{
  "status": "success",
  "data": [
    "200"
  ]
}

```

Questions:
* Should `labelValuesByMatchers` be a method of the queriers to be consistent with `LabelValues`? I tried that out but ran in to issues, but could have another go at it.
* Is a seperate unittest for `labelValuesByMatchers` a good idea? From what I see tests are not run on specific functions of `api.go`. I did feel like having more labels in my tests but also didn't want to mess with all the other tests cases in `api_test.go`
* Build fails, but seems to be unrelated?

```
Unable to find image 'quay.io/prometheus/golang-builder:1.14-base' locally

Error response from daemon: Get https://quay.io/v2/prometheus/golang-builder/manifests/1.14-base: Get https://quay.io/v2/auth?scope=repository%3Aprometheus%2Fgolang-builder%3Apull&service=quay.io: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)

!! exit status 1

```

### DIFF
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -321,7 +321,7 @@ URL query parameters:
 - `start=<rfc3339 | unix_timestamp>`: Start timestamp. Optional.
 - `end=<rfc3339 | unix_timestamp>`: End timestamp. Optional.
 - `match[]=<series_selector>`: Repeated series selector argument that selects the
-  series from which to read the label's values 
+  series from which to read the label values 
 
 
 The `data` section of the JSON response is a list of string label values.
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -320,6 +320,8 @@ URL query parameters:
 
 - `start=<rfc3339 | unix_timestamp>`: Start timestamp. Optional.
 - `end=<rfc3339 | unix_timestamp>`: End timestamp. Optional.
+- `match[]=<series_selector>`: Repeated series selector argument that selects the
+  series from which to read the label's values 
 
 
 The `data` section of the JSON response is a list of string label values.
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -521,6 +521,14 @@ func (api *API) labelValues(r *http.Request) (result apiFuncResult) {
 	if err != nil {
 		return apiFuncResult{nil, &apiError{errorBadData, errors.Wrap(err, "invalid parameter 'end'")}, nil, nil}
 	}
+	var matcherSets [][]*labels.Matcher
+	for _, s := range r.Form["match[]"] {
+		matchers, err := parser.ParseMetricSelector(s)
+		if err != nil {
+			return apiFuncResult{nil, &apiError{errorBadData, err}, nil, nil}
+		}
+		matcherSets = append(matcherSets, matchers)
+	}
 
 	q, err := api.Queryable.Querier(r.Context(), timestamp.FromTime(start), timestamp.FromTime(end))
 	if err != nil {
@@ -538,12 +546,54 @@ func (api *API) labelValues(r *http.Request) (result apiFuncResult) {
 		q.Close()
 	}
 
-	vals, warnings, err := q.LabelValues(name)
+	var vals []string
+	var warnings storage.Warnings
+	if len(r.Form["match[]"]) > 0 {
+		// get all series which match
+		var sets []storage.SeriesSet
+		for _, mset := range matcherSets {
+			s := q.Select(false, nil, mset...)
+			sets = append(sets, s)
+		}
+		vals, warnings, err = labelValuesByMatchers(sets, name)
+	} else {
+		vals, warnings, err = q.LabelValues(name)
+	}
+
 	if err != nil {
 		return apiFuncResult{nil, &apiError{errorExec, err}, warnings, closer}
 	}
 
 	return apiFuncResult{vals, nil, warnings, closer}
+}
+
+// LabelValuesByMatchers uses matchers to filter out the series
+// from which we get the label values
+func labelValuesByMatchers(sets []storage.SeriesSet, name string) ([]string, storage.Warnings, error) {
+	// get all the labels in the selected series
+	set := storage.NewMergeSeriesSet(sets, storage.ChainedSeriesMerge)
+	labelValuesMap := make(map[string]string)
+	for set.Next() {
+		series := set.At()
+		labelValue := series.Labels().Get(name)
+		// to avoid duplicates, any better way?
+		_, ok := labelValuesMap[labelValue]
+		if !ok && labelValue != ""  {
+			labelValuesMap[labelValue] = labelValue
+		}
+	}
+
+	warnings := set.Warnings()
+	if set.Err() != nil {
+		return nil, warnings, set.Err()
+	}
+	// convert map to array
+	labelValues := []string{}
+	for key := range labelValuesMap {
+		labelValues = append(labelValues, key)
+	}
+	sort.Strings(labelValues)
+	return labelValues, warnings, nil
 }
 
 var (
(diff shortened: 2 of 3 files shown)


web/api/v1/api.go
```diff
@@ -538,14 +546,54 @@ func (api *API) labelValues(r *http.Request) (result apiFuncResult) {
...
 		return apiFuncResult{nil, &apiError{errorExec, err}, warnings, closer}
 	}
 
 	return apiFuncResult{vals, nil, warnings, closer}
 }
 
+// LabelValuesByMatchers uses matchers to filter out matching series, then label values are extracted.
+func labelValuesByMatchers(sets []storage.SeriesSet, name string) ([]string, storage.Warnings, error) {
+	set := storage.NewMergeSeriesSet(sets, storage.ChainedSeriesMerge)
+	labelValuesMap := make(map[string]string)
```
brian-brazil, line 573: A `map[string]struct{}` is the usual way to do a set.

web/api/v1/api.go
```diff
@@ -538,14 +546,54 @@ func (api *API) labelValues(r *http.Request) (result apiFuncResult) {
...
+	for set.Next() {
+		series := set.At()
+		labelValue := series.Labels().Get(name)
+		// This is to avoid duplicates.
```
brian-brazil, line 577: You can do a simple assignment and not worry about it.

web/api/v1/api.go
eklockare: Aha, didn't know, changed to `map[string]struct{}` :+1:

web/api/v1/api.go
eklockare: True, fixed

docs/querying/api.md
```diff
@@ -320,6 +320,8 @@ URL query parameters:
 
 - `start=<rfc3339 | unix_timestamp>`: Start timestamp. Optional.
 - `end=<rfc3339 | unix_timestamp>`: End timestamp. Optional.
+- `match[]=<series_selector>`: Repeated series selector argument that selects the
+  series from which to read the label's values
```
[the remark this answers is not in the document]
eklockare: fixed

eklockare:
@brian-brazil I'm not familiar with the inner workings of the data loading yet. Does your hint solution make it so that less (or no) data is loaded (i.e. no chunks)? So that we get a larger proportion of  metadata over actual data

web/api/v1/api.go
```diff
@@ -570,14 +570,12 @@ func (api *API) labelValues(r *http.Request) (result apiFuncResult) {
...
-	labelValuesMap := make(map[string]string)
+	labelValuesSet := make(map[string]struct{})
 	for set.Next() {
 		series := set.At()
 		labelValue := series.Labels().Get(name)
-		// This is to avoid duplicates.
-		_, ok := labelValuesMap[labelValue]
-		if !ok && labelValue != ""  {
-			labelValuesMap[labelValue] = labelValue
+		if labelValue != ""  {
```
brian-brazil, line 577: The label value can't be empty, if it is that's a bug in the tsdb.

brian-brazil:
> Does your hint solution make it so that less (or no) data is loaded (i.e. no chunks)?

It makes it so the chunks don't end up trashing the page cache. Basically due to this the feature in this PR could cause severe performance problems, so I think we need #5547 fixed first to be safe.

web/api/v1/api.go
eklockare: fixed

bwplotka (approved):
I would say, let's go for it.  It's worth to mention it's not as fast and non-matcher function, but it will do the work just fine.

This PR looks fine to me LGTM! :heart: 

I guess we want the same on label_name as well (:

bwplotka:
Plus we need to fix tests (:

bwplotka:
are you available @eklockare ? sorry for long delay! (:

web/api/v1/api.go
```diff
@@ -538,14 +546,50 @@ func (api *API) labelValues(r *http.Request) (result apiFuncResult) {
...
 	}
 
-	vals, warnings, err := q.LabelValues(name)
+	var vals []string
+	var warnings storage.Warnings
+	if len(r.Form["match[]"]) > 0 {
+		// Get all series which match matchers.
+		var sets []storage.SeriesSet
+		for _, mset := range matcherSets {
+			s := q.Select(false, nil, mset...)
```
brian-brazil, line 555: You need to pass in a hint here with the series function, otherwise this could trash the page cache.

web/api/v1/api.go
```diff
@@ -538,14 +546,50 @@ func (api *API) labelValues(r *http.Request) (result apiFuncResult) {
...
+		labelValue := series.Labels().Get(name)
+		labelValuesSet[labelValue] = struct{}{}
+	}
+
+	warnings := set.Warnings()
+	if set.Err() != nil {
+		return nil, warnings, set.Err()
+	}
+	// Convert the map to an array.
+	labelValues := []string{}
```
yeya24, line 585: Maybe we can use `labelValues := make([]string, 0, len(labelValueSet))` here.

brancz:
Hey @eklockare, very nice contribution! This seems very close to completion. Do you think you'll have time to resolve the remaining comments? Otherwise I'm happy to continue your work! :)

web/api/v1/api.go
```diff
@@ -538,14 +546,50 @@ func (api *API) labelValues(r *http.Request) (result apiFuncResult) {
...
+	var vals []string
+	var warnings storage.Warnings
+	if len(r.Form["match[]"]) > 0 {
+		// Get all series which match matchers.
+		var sets []storage.SeriesSet
+		for _, mset := range matcherSets {
+			s := q.Select(false, nil, mset...)
+			sets = append(sets, s)
+		}
+		vals, warnings, err = labelValuesByMatchers(sets, name)
```
bwplotka, line 558: Can we add TODO to extend LabelValues and LabelNames with matches?

brancz: I would also prefer if this was done on TSDB level. Last time we discussed this we decided against it, but I'm happy to have that discussion again (I realize that TSDB doesn't work that way today).

brancz:
> I guess we want the same on label_name as well (:

Side note, label-names are going to work slightly differently, as we need extra logic to ignore label-names that we have an exact matcher for.

brian-brazil:
> Side note, label-names are going to work slightly differently, as we need extra logic to ignore label-names that we have an exact matcher for.

Why's that?

codesome (commented):
I was discussing this offline with @bwplotka and we think this should be part of TSDB to make it more efficient. Maybe introducing hints for `LabelValues` similar to Select() which remote storages can also make use of.

brancz:
@brian-brazil it doesn't make sense to return a label name that we pass an exact matcher for, isn't that the whole point of passing matchers?

brian-brazil:
If I have a set of series per a selector, why wouldn't I return all of the label names that have non-empty values among those series?  
What you propose doesn't make sense to me semantically. That it appeared as a matcher doesn't change that it is a relevant label name.

brancz:
Yeah, thought about it again, and I agree.

bwplotka:
Looks like this PR is not active anymore, I heard @yeya24 want to continue this to get this merged 🤗 

Hope @eklockare, you are ok with this, we will use your commits as a start 👍🏽

brian-brazil:
Superseded by #8301

PR #7498, opened 2020-07-01
digitalocean: use a safer pagination method

roidelapluie:
this method is documented here: https://github.com/digitalocean/godo


brian-brazil:
:+1:

Issue #7454, opened 2020-06-25
Panic from /metrics during startup


carlpett:
**What did you do?**
Restart Prometheus

**What did you expect to see?**
Clean startup

**What did you see instead? Under which circumstances?**
Panic and stack trace during a call to /metrics before startup finished (that the upstream Prometheus server called /metrics despite Prometheus not having signalled ready appears to be a bug in our service discovery in that setup)
`{"caller":"web.go:99","client":"10.86.14.4:43362","component":"web","err":"runtime error: invalid memory address or nil pointer dereference","level":"error","msg":"panic while serving request","stack":"<see decoded stack below>","url":"/metrics"}`

Full logs supplied below, here's the stack:
```
goroutine 166 [running]:
github.com/prometheus/prometheus/web.withStackTracer.func1.1(0x2f3b7e0, 0xc0004ccc60, 0xc0000f6600)
	/app/web/web.go:98 +0xb5
panic(0x2334740, 0x42d4ba0)
	/usr/local/go/src/runtime/panic.go:969 +0x166
github.com/prometheus/prometheus/tsdb.(*Head).MinTime(...)
	/app/tsdb/head.go:1407
github.com/prometheus/prometheus/tsdb.newDBMetrics.func3(0x0)
	/app/tsdb/db.go:228 +0x80
github.com/prometheus/client_golang/prometheus.(*valueFunc).Write(0xc0001fab40, 0xc0001c0620, 0xc000113040, 0x3)
	/app/vendor/github.com/prometheus/client_golang/prometheus/value.go:76 +0x2f
github.com/prometheus/client_golang/prometheus.processMetric(0x2f53020, 0xc0001fab40, 0xc000112ed0, 0xc000112f00, 0x0, 0x0, 0x0)
	/app/vendor/github.com/prometheus/client_golang/prometheus/registry.go:595 +0x9b
github.com/prometheus/client_golang/prometheus.(*Registry).Gather(0xc0001385f0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/app/vendor/github.com/prometheus/client_golang/prometheus/registry.go:489 +0x87c
github.com/prometheus/client_golang/prometheus/promhttp.HandlerFor.func1(0x7f98a29be170, 0xc0000a1180, 0xc0000f6900)
	/app/vendor/github.com/prometheus/client_golang/prometheus/promhttp/http.go:126 +0x93
[...]

PR #7501, opened 2020-07-02
Fix panic from db metrics

codesome:
Fixes https://github.com/prometheus/prometheus/issues/7454

This issue exists in whole lot of releases (exists in 2.15.0, didnt check before that). So I am not planning to release 2.19.3 for this and leave it for 2.20.0.

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -569,13 +569,6 @@ func open(dir string, l log.Logger, r prometheus.Registerer, opts *Options, rngs
 		autoCompact: true,
 		chunkPool:   chunkenc.NewPool(),
 	}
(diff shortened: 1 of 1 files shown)


tsdb/db.go
```diff
@@ -617,6 +610,14 @@ func open(dir string, l log.Logger, r prometheus.Registerer, opts *Options, rngs
 		return nil, err
 	}
 
+	// Register metrics after assigning the head block else it causes panics is the gauge function.
```
brian-brazil, line 613: Grammar is off here, I'd end the sentence at block.

codesome: It should be `in the gauge function`. I will cut the sentence short.

PR #7504, opened 2020-07-02
Upgrade Golang Client + fix usage of LabelValues in promTool

Nexucis:
It should fix the build of the PR #6872


brian-brazil:
Hmm, is #7463 already doing this?

Nexucis:
ah yes indeed. And it's more or less exactly the same changes.
I will close this one then. No need to be duplicated.

Nexucis:
sorry for the duplication and thanks for the time you took @brian-brazil. I should have look at it before opening this PR ....

PR #7505, opened 2020-07-02
Promtool range query should exit when fail to parse time

yeya24:

When it fails to parse start time or end time is before start time, I think it should return 1 instead of continuing.


yeya24:
Is it good to merge? :eyes:

brian-brazil:
Thanks!

PR #7507, opened 2020-07-02
Openstack: Reduce timeouts

roidelapluie:
Set saner values for openstack timeouts


brian-brazil:
:+1:

Issue #7478, opened 2020-06-29
tsdb, index reader: Rename SortedLabelValue back to LabelValues and add UnsortedLabelValues.


bwplotka:
Hi, 

We just spent 30m figuring out the differences and the fact that `index.Reader.LabelValues` is unsorted and `querier.LabelValues`  is always sorted, creates real confusion. 

What do you think for quick readability improvement and changing things back and adding `UnsortedLabelValues` or `LabelValues(sorted)` as `Select` has? (: 

cc @gouthamve @brian-brazil @brancz

bwplotka:
Related to https://github.com/prometheus/prometheus/pull/7448

codesome:
Sounds good. I would prefer `UnsortedLabelValues` instead of passing a bool to avoid a lot of code changes (as `UnsortedLabelValues` would not be used very often)

eklockare:
Made an attempt, I hope I interpreted correctly

ngetahun:
Is this issue up for grabs? I'd like to submit a PR.

bwplotka:
Looks like this was blocked on our review, reviewed now. So depends if @eklockare has time (:

ngetahun:
I have a WIP [PR](https://github.com/prometheus/prometheus/pull/8098)

bwplotka:
@ngetahun Can you base your changes on https://github.com/prometheus/prometheus/pull/7511 PR, because it super close to be done, just small nits. WDYT?

yashrsharma44:
Can I work on this issue, if it's not resolved?

palash25:
hi is this still up for grabs? i see 3 different open PRs for this

codesome:
Pinged on #8624 to bring it back alive. I would wait for a week for any response.

roidelapluie:
We have looked at this pull request during our bug scrub.

By now, we aren't even sure if it is worth changing the names. People are really used to the semantics now, so better leave it as is.

Thank you for your contribution.

PR #7511, opened 2020-07-03
Changed labelvalues method names for readers

eklockare:
Implements #7478 

SortedLabelValues changed to LabelValues
LabelValues changed to UnsortedLabelValues

for IndexReader interface implemented by:
headIndexReader, Reader and blockIndexReader

### DIFF
--- a/tsdb/block.go
+++ b/tsdb/block.go
@@ -61,11 +61,11 @@ type IndexReader interface {
 	// beyond the lifetime of the index reader.
 	Symbols() index.StringIter
 
-	// SortedLabelValues returns sorted possible label values.
-	SortedLabelValues(name string) ([]string, error)
-
-	// LabelValues returns possible label values which may not be sorted.
+	// LabelValues returns sorted possible label values.
 	LabelValues(name string) ([]string, error)
+
+	// UnsortedLabelValues returns possible label values which may not be sorted.
+	UnsortedLabelValues(name string) ([]string, error)
 
 	// Postings returns the postings list iterator for the label pairs.
 	// The Postings here contain the offsets to the series inside the index.
@@ -421,13 +421,13 @@ func (r blockIndexReader) Symbols() index.StringIter {
 	return r.ir.Symbols()
 }
 
-func (r blockIndexReader) SortedLabelValues(name string) ([]string, error) {
-	st, err := r.ir.SortedLabelValues(name)
+func (r blockIndexReader) LabelValues(name string) ([]string, error) {
+	st, err := r.ir.LabelValues(name)
 	return st, errors.Wrapf(err, "block: %s", r.b.Meta().ULID)
 }
 
-func (r blockIndexReader) LabelValues(name string) ([]string, error) {
-	st, err := r.ir.LabelValues(name)
+func (r blockIndexReader) UnsortedLabelValues(name string) ([]string, error) {
+	st, err := r.ir.UnsortedLabelValues(name)
 	return st, errors.Wrapf(err, "block: %s", r.b.Meta().ULID)
 }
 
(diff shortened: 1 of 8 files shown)


tsdb/querier_test.go
```diff
@@ -1400,13 +1400,13 @@ func (m mockIndex) Close() error {
 	return nil
 }
 
-func (m mockIndex) SortedLabelValues(name string) ([]string, error) {
+func (m mockIndex) LabelValues(name string) ([]string, error) {
 	values, _ := m.LabelValues(name)
```
codesome, line 1404: You missed this. Can you verify if you have missed anywhere else?

eklockare: My bad, fixed. Strange that the test passed, should have been a stackoverflow you'd think.
Will verify

eklockare: I have verified, they all seemed changed now, assuming that:
All outside calls that were to the previous `LabelValues` to still be to `LabelValues` ?
The return value will be sorted now, which I think is desired, correct me if I'm wrong

eklockare:
@bwplotka Sure, will look into doing the bool flag pattern :+1:

eklockare:
@bwplotka I added the boolean flag ` LabelValues(name string, sortedValues bool) ([]string, error)` to the touched files. But it is inconsistent with other parts of the code base. Should I change it everywhere?

roidelapluie:
Hello @bwplotka, can you review this PR? Thanks!

tsdb/index/index.go
```diff
@@ -1418,21 +1418,21 @@ func (r *Reader) SymbolTableSize() uint64 {
...
 	}
 	return values, err
 }
 
-// LabelValues returns value tuples that exist for the given label name.
+// UnsortedLabelValues returns value tuples that exist for the given label name.
 // It is not safe to use the return value beyond the lifetime of the byte slice
 // passed into the Reader.
-func (r *Reader) LabelValues(name string) ([]string, error) {
+func (r *Reader) UnsortedLabelValues(name string) ([]string, error) {
```
bwplotka, line 1435:
```suggestion
func (r *Reader) unsortedLabelValues(name string) ([]string, error) {
```

No need to make it public (:

tsdb/querier_test.go
```diff
@@ -1144,13 +1144,15 @@ func (m mockIndex) Close() error {
...
+func (m mockIndex) LabelValues(name string, sortedValues bool) ([]string, error) {
+	values, _ := m.UnsortedLabelValues(name)
+	if sortedValues {
+		sort.Strings(values)
+	}
 	return values, nil
 }
 
-func (m mockIndex) LabelValues(name string) ([]string, error) {
+func (m mockIndex) UnsortedLabelValues(name string) ([]string, error) {
```
bwplotka, line 1155: same here

tsdb/block.go
```diff
@@ -61,11 +61,8 @@ type IndexReader interface {
 	// beyond the lifetime of the index reader.
 	Symbols() index.StringIter
 
-	// SortedLabelValues returns sorted possible label values.
-	SortedLabelValues(name string) ([]string, error)
-
-	// LabelValues returns possible label values which may not be sorted.
-	LabelValues(name string) ([]string, error)
+	// LabelValues returns sorted possible label values.
+	LabelValues(name string, sortedValues bool) ([]string, error)
```
bwplotka, line 65: To be strict, can we switch parameter to be first? It will be then consistent with :hugs:  https://github.com/prometheus/prometheus/blob/24b897164f5bc5178d399f34843edd72967fb667/tsdb/querier.go#L122

bwplotka (approved):
Thanks so much for this work and sorry for huge delay in review. I missed this. LGTM, with quick nits :hugs:

roidelapluie:

By now, we aren't even sure if it is worth changing the names. People are really used to the semantics now, so better leave it as is.

PR #7514, opened 2020-07-04
openstack tests: use new test.Cleanup function

roidelapluie:
Since we dependend on go1.14 now, we can use T.Cleanup
https://golang.org/pkg/testing/#T.Cleanup

This provides a nicer approach to shut down the test server.

### DIFF
--- a/discovery/openstack/hypervisor_test.go
+++ b/discovery/openstack/hypervisor_test.go
@@ -26,10 +26,6 @@ type OpenstackSDHypervisorTestSuite struct {
 	Mock *SDMock
 }
 
-func (s *OpenstackSDHypervisorTestSuite) TearDownSuite() {
-	s.Mock.ShutdownServer()
-}
-
 func (s *OpenstackSDHypervisorTestSuite) SetupTest(t *testing.T) {
 	s.Mock = NewSDMock(t)
 	s.Mock.Setup()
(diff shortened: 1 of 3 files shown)


roidelapluie:
Note:

1/ there are other places where we would benefit from this. But in other packages/use cases, so it does not make sense to batch them in one PR
2/ this approach is already used in the docker swarm and digitalocean discovery

discovery/openstack/hypervisor_test.go
```diff
@@ -89,8 +85,6 @@ func TestOpenstackSDHypervisorRefresh(t *testing.T) {
 	} {
 		testutil.Equals(t, model.LabelValue(v), tg.Targets[1][model.LabelName(l)])
 	}
-
-	mock.TearDownSuite()
```
brian-brazil, line 93: This doesn't seem to be doing quite the same thing, as this is for the mock created in this function as against s.Mock.

roidelapluie: mock = s

roidelapluie: mock is  &OpenstackSDHypervisorTestSuite{}

func (s *OpenstackSDInstanceTestSuite) TearDownSuite()  is what I remove. That calls s.Mock.ShutdownServer().

roidelapluie: OpenstackSDInstanceTestSuite and OpenstackSDInstanceTestSuite both call mock.Setup

roidelapluie: The mock behind both is the same. I think this PR is correct.

brian-brazil:
:+1:

PR #7523, opened 2020-07-06
Fix the type of the alert name

### LINKED ISSUE #7083: Should alerting rules names really be valid metric names ?
Hi, on the documentation page about [Recording rules](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/#recording-rules), it's mentioned that :

>  The names of recording and alerting rules must be valid metric names

... meaning that it should match the regex `[a-zA-Z_:][a-zA-Z0-9_:]*`.

However, in my own configuration, I've used spaces in my alert names like `Daily Peak` and it's working just fine. Indeed, alert names end up as a label **value** (allowing any Unicode Character) as in  `ALERTS{alertname="Daily peak"}` and not as a metric or label **name** (constrained to an ascii regex).

So I'm wondering if alerting rules names should really be valid metric **names** ?
If not I'll make the pull request to add clarification on this subject.

Thx for this piece of software :wink:

roidelapluie:
The alert name should be a valid label value, not a metric name.

Fix #7083

### DIFF
--- a/docs/configuration/recording_rules.md
+++ b/docs/configuration/recording_rules.md
@@ -98,7 +98,7 @@ labels:
 The syntax for alerting rules is:
 
 ```
-# The name of the alert. Must be a valid metric name.
+# The name of the alert. Must be a valid label value.
 alert: <string>
 
 # The PromQL expression to evaluate. Every evaluation cycle this is


brian-brazil:
:+1:

PR #7524, opened 2020-07-06
Downgrade docker dependency to the stable branch

roidelapluie:
This PR downgrades the docker dependency to the stable 19.03 branch.

It means that we will negociate 1.40 by default and not the uncomplete
1.41 API.

It was not possible to use the stable branch before because a patch
was pending to build on netbsd and dragonfly:

https://github.com/moby/moby/commit/a40b877fbb9ec53925355cf9812fbb14fff925b9


brian-brazil:
:+1:

Issue #7373, opened 2020-06-09
Unknown symbol error when compacting Head block


codesome:
**What did you do?**

Ran TSDB inside Cortex

**What did you expect to see?**

No errors when compacting the Head block.

**What did you see instead? Under which circumstances?**

Compaction from Head failing with the following error
```persist head block: 2 errors: write compaction: add series: symbol entry for \"2020-06-08T07:02:36.407Z\" does not exist, unknown symbol \"2020-06-08T07:02:36.407Z\"; series not 16-byte aligned at 147394303```

And the same error repeats.

This error is not related to any disk corruption as all the data is from the memory. 

The main components involved in this compaction operations are the Head index reader and the Head block itself. It needs some investigations on how is it possible - whether some bug in the compaction code (or) some race somewhere when ingesting data which causes the symbols to be lost (or) something else.

Though this is seen inside Cortex when using TSDB, it might be very much possible while running Prometheus too.

---
Prometheus version: `v2.16.0`

**Comment by levydori** [comment]
Hey,
we are seeing the same issue on Prometheus (v2.15.2). 

`level=error ts=2020-06-17T19:04:17.039Z caller=db.go:609 component=tsdb msg="compaction failed" err="persist head block: 2 errors: write compaction: add series: symbol entry for \"bbbe9aa8-254f-44f6-b5a0-09c0ea64fd9f\" does not exist, unknown symbol \"bbbe9aa8-254f-44f6-b5a0-09c0ea64fd9f\"; series not 16-byte aligned at 22979295"` 

It's causing compaction to fail, and disk usage to slowly increase (until it reaches 100%).

**Comment by brian-brazil** [comment]
I also had a private report of this in Prometheus, not sure which version.

**Comment by bwplotka** [comment]
Hey @codesome  we just got it and we are on exact commit that was suppose to fix it ): 

```
level=error ts=2020-07-11T13:30:09.772232559Z caller=db.go:685 component=receive tenant=XXX msg="compaction failed" err="persist head block: 2 errors: write compaction: add series: symbol entry for \"olm-operator-666b97bf6d-f6mrj\" does not exist, unknown symbol \"olm-operator-666b97bf6d-f6mrj\"; series not 16-byte aligned at 112720367"
[...]

PR #7526, opened 2020-07-07
Fix unknown symbol error during head compaction

codesome:
Fixes https://github.com/prometheus/prometheus/issues/7373

At this particular point here in compaction

https://github.com/prometheus/prometheus/blob/9875afc491983cc7462fef336ab1c6b67da45020/tsdb/compact.go#L702-L710

We are getting the postings first and then the symbols. If you see the fix, previously we were adding postings first and symbols later separately in the head, so there is a race here where we capture the new series in compaction and get the old symbols. In this fix, we just add the postings in the head only after capturing the symbols.

I have added the test to re-produce this and the unknown symbol error disappeared after this fix.

Interestingly, this test reproduces another bug which we have seen before - `unexpected error: persist head block: write compaction: add series: out-of-order series added with label set "{n414=\"v414\"}"` - which this PR does not fix and it leads to a flaky test. So, either I will fix it in the same PR if I find it, or remove the test and add it back when I have a fix for that.

### DIFF
--- a/tsdb/head_test.go
+++ b/tsdb/head_test.go
@@ -1876,65 +1876,65 @@ func TestHeadLabelNamesValuesWithMinMaxRange(t *testing.T) {
 	}
 }
 
-func TestHeadCompactionRace(t *testing.T) {
-	for i := 0; i < 10; i++ {
-		t.Run(fmt.Sprintf("run %d", i), func(t *testing.T) {
-			tsdbCfg := &Options{
-				RetentionDuration: 100000000,
-				NoLockfile:        true,
-				MinBlockDuration:  1000000,
-				MaxBlockDuration:  1000000,
-			}
-
-			db, closer := openTestDB(t, tsdbCfg, []int64{1000000})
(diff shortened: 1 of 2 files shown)


codesome:
I have commented out the test so that the fix can go ahead. The test is still relevant to debug the other race mentioned in the issue description, so I plan to keep the commented test and remove/fix it when we have a fix for that error.

tsdb/head_test.go
```diff
@@ -1875,3 +1875,66 @@ func TestHeadLabelNamesValuesWithMinMaxRange(t *testing.T) {
 		})
 	}
 }
+
+//func TestHeadCompactionRace(t *testing.T) {
```
bwplotka, line 1879: If you want to skip, just do `t.Skip`. Otherwise let's remove (:

codesome: I will add `t.Skip` (didnt know about that, thanks!), dont want to remove as I will need it later

PR #7529, opened 2020-07-08
Clarify error messaging when a rule group name is empty

cstyan:
As far as I can tell, without looking at the logs there's no way to see failures of loading rules files. It would be nice to have a metric for this so it can be alerted on. I noticed that it seems none of our `Manager` types have their own metrics, if that's by design let me know.

I also modified the wording of an error message, 'should' was being used in a case where it seems the word used should be 'must'. 

cc @gouthamve and @beorn7 since you've both modified this area most recently

### DIFF
--- a/pkg/rulefmt/rulefmt.go
+++ b/pkg/rulefmt/rulefmt.go
@@ -68,7 +68,7 @@ func (g *RuleGroups) Validate(node ruleGroups) (errs []error) {
 
 	for j, g := range g.Groups {
 		if g.Name == "" {
-			errs = append(errs, errors.Errorf("%d:%d: Groupname should not be empty", node.Groups[j].Line, node.Groups[j].Column))
+			errs = append(errs, errors.Errorf("%d:%d: Groupname must not be empty", node.Groups[j].Line, node.Groups[j].Column))
 		}
 
 		if _, ok := set[g.Name]; ok {


brian-brazil:
Hmm, how does this work in terms of the config reload generally? I'd hope that this would cause the whole config reload to fail, which we already have metrics for and you'll need to look at logs anyway to debug.

beorn7:
I'm fine with the "should" → "must" change.

I also would have expected that the whole config reload fails and increments the respective metrics. Spontaneously I'd say if the config reload succeeds with invalid rules files, that's a bug.

I cannot investigate in more detail because I have too many things on my plate right now.

cstyan:
> Hmm, how does this work in terms of the config reload generally? I'd hope that this would cause the whole config reload to fail, which we already have metrics for and you'll need to look at logs anyway to debug.

Where is the metric for the config reload failure? Are you referring to the gauges in `cmd/prometheus/main.go` for tracking the status of the last reload and the last successful reloads timestamp?

The motivation behind adding the metric to the manager is that ecosystem projects that make use of the rules (or other packages) Manager type would need to add instrumentation around their calls to the Manager. The way we're using the Manager that change is relatively trivial, but I thought I'd see what everyone though about adding the metric upstream as well.

> I also would have expected that the whole config reload fails and increments the respective metrics. Spontaneously I'd say if the config reload succeeds with invalid rules files, that's a bug.

I think a config reload only requires that the contents of the config file itself are valid, it doesn't seem to load and parse rules files but just checks that the paths to them are valid.

brian-brazil:
> Are you referring to the gauges in cmd/prometheus/main.go for tracking the status of the last reload and the last successful reloads timestamp?

Yes.

> I think a config reload only requires that the contents of the config file itself are valid, it doesn't seem to load and parse rules files but just checks that the paths to them are valid.

That sounds like a bug then.

> The way we're using the Manager that change is relatively trivial, but I thought I'd see what everyone though about adding the metric upstream as well.

Given that in Prometheus once that bug is fixed we already have a metric covering it, I'd say let's not add another metric that might confuse users.

cstyan:
Sorry, my statement wasn't correct. I was only looking at the parsing of the config file. The reloading of the config in main.go results in attempting to load the config into each manager type. Which includes the Rules Manager.

It's a bit roundabout and hard to follow the first time you read it, but loading/reloading config would result in loading rules files and an error/metric increment if the rules files were erroneous.

> Given that in Prometheus once that bug is fixed we already have a metric covering it, I'd say let's not add another metric that might confuse users.

I'm fine with this.

brian-brazil:
Okay, sounds like this PR is going to just be should -> must then.

cstyan:
Update PR to only include the error message change.

PR #7540, opened 2020-07-09
add os.RemoveAll err verification

zhouhao3:

### DIFF
--- a/storage/remote/queue_manager_test.go
+++ b/storage/remote/queue_manager_test.go
@@ -60,7 +60,9 @@ func TestSampleDelivery(t *testing.T) {
 
 	dir, err := ioutil.TempDir("", "TestSampleDeliver")
 	testutil.Ok(t, err)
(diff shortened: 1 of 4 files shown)


q384566678:
> Do you think you can fix them?

@codesome  I don't know why this problem occurs under the windows system. At present, I have no solution. Can you give me some suggestions? Or merge this PR first, and then someone creates a new PR to solve this problem.

beorn7:
The problem occurs in the ci/circleci test for MS Windows. That should give a hint (and at the same time is the reason why we cannot merge this as is: It would break the CI tests).

codesome:
@q384566678 basically we are not closing files somewhere and in Windows you cannot delete a file/directory if it is open somewhere. So you have to check where all are we not closing files in the test and close them before the deletion phase.

zhouhao3:
@codesome Done, PTAL.

PR #7541, opened 2020-07-09
Refactor dockerswarm refresh for testing

roidelapluie:
We were missing testing on the behaviour of the configuration
unmarshalling.

This PR adds a refresh command that can be used to test that we
use the correct refresh function.


brian-brazil:
:+1:

PR #7542, opened 2020-07-09
Add Docker Swarm configuration example

roidelapluie:

### DIFF
--- a/documentation/examples/prometheus-dockerswarm.yml
+++ b/documentation/examples/prometheus-dockerswarm.yml
@@ -1,12 +1,15 @@
 # A example scrape configuration for running Prometheus with
 # Docker Swarm.
 #
-# This exemple works with Prometheus node_exporter deployed with:
-# docker service create --name node -l prometheus-job=node \
-#     --mode=global --publish published=9100,target=9100,mode=host \
-#     --mount type=bind,src=/sys,dst=/sys \
-#     --mount type=bind,src=/proc,dst=/proc \
-#     prom/node-exporter
+# This example works with cadvisor deployed with:
+# docker service create --name cadvisor -l prometheus-job=cadvisor
+#     --mode=global --publish published=8080,target=8080,mode=host
+#     --mount type=bind,src=/var/run/docker.sock,dst=/var/run/docker.sock,ro
+#     --mount type=bind,src=/,dst=/rootfs,ro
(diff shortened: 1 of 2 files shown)


documentation/examples/prometheus-dockerswarm.yml
```diff
@@ -0,0 +1,36 @@
+# A example scrape configuration for running Prometheus with
+# Docker Swarm.
+#
+# This exemple works with Prometheus node_exporter deployed with:
```
brian-brazil, line 4: example

documentation/examples/prometheus-dockerswarm.yml
```diff
@@ -0,0 +1,36 @@
...
+# docker service create --name node -l prometheus-job=node \
```
brian-brazil, line 5: Hmm, do we want to be encouraging people to run the node exporter in Docker? @SuperQ

documentation/examples/prometheus-dockerswarm.yml
```diff
@@ -0,0 +1,36 @@
...
+    relabel_configs:
+      # Only keep containers that should be running.
+      - source_labels: [__meta_dockerswarm_task_desired_state]
+        regex: running
+        action: keep
+      # Only keep containers that have a `prometheus-job` label.
+      - source_labels: [__meta_dockerswarm_service_label_prometheus_job]
+        regex: .+
+        action: keep
+      # Use the tasks labels that are prefixed by `prometheus-`.
```
brian-brazil, line 32: task

documentation/examples/prometheus-dockerswarm.yml
SuperQ: Similar to Kubernetes, for swarm users, this is likely to be common. Having correct documentation examples here is probably better than many of the incorrect docker-compose.yml configs I've seen from users having trouble with it.

SuperQ: I did a bit more digging, and it looks like the big problem is that Swarm mode doesn't support `--pid=host`, so I'm pretty sure this won't work correctly. :frowning:

roidelapluie: updated for cadvisor

brian-brazil:
:+1:

SuperQ:
Nice!

PR #7544, opened 2020-07-09
Display dates as well as timestamps in status page

sylr:
image

### DIFF
--- a/web/ui/templates/status.html
+++ b/web/ui/templates/status.html
@@ -99,14 +99,14 @@
         <tr>
            <th>Number Of Series </th>
            <th>Number of Chunks</th>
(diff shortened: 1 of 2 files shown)


roidelapluie:
Thanks! cc @codesome @juliusv

juliusv:
@sylr Thanks! Would you be able to do the same change to the new UI instead of the old one (or both)? Since the old one is going to disappear soon...

juliusv:
Hmm, actually it seems the head stats are even missing in the new UI on https://demo.promlabs.com/new/tsdb-status :thinking: 

So maybe ignore that for now, we need to look at that separately.

juliusv:
:+1:

PR #7545, opened 2020-07-09
Document that Clean Tombstones may need extra disk space

GabrielCastro:
For a process that is designed to reduce disk usage this I feel this is an important bit of information

### DIFF
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -1006,7 +1006,7 @@ $ curl -X POST \
 *New in v2.1 and supports PUT from v2.9*
 
 ### Clean Tombstones
-CleanTombstones removes the deleted data from disk and cleans up the existing tombstones. This can be used after deleting series to free up space. This process involves rerwiting existing blocks that have tombstones which requires free disk space (upto double the used space if all blocks have tombstones).
+CleanTombstones removes the deleted data from disk and cleans up the existing tombstones. This can be used after deleting series to free up space. This process involves rewriting existing blocks that have tombstones (aka normal compaction). Note that like normal compaction, this requires some extra disk space (one more TSDB block maximum).
 
 If successful, a `204` is returned.
 
--- a/docs/querying/api.md
+++ b/docs/querying/api.md
@@ -1006,7 +1006,7 @@ $ curl -X POST \
 *New in v2.1 and supports PUT from v2.9*
 
 ### Clean Tombstones
-CleanTombstones removes the deleted data from disk and cleans up the existing tombstones. This can be used after deleting series to free up space.
+CleanTombstones removes the deleted data from disk and cleans up the existing tombstones. This can be used after deleting series to free up space. This process involves rerwiting existing blocks that have tombstones which requires free disk space (upto double the used space if all blocks have tombstones).
 
 If successful, a `204` is returned.
 


brian-brazil:
Hmm, this shouldn't be more space than you already require for normal compaction so such a note may confuse more than help.

roidelapluie:
> Hmm, this shouldn't be more space than you already require for normal compaction so such a note may confuse more than help.

Do we detele all the old blocks at the end? Then its a bug. If it is more 'rolling' that should be fine.

brian-brazil:
We do one compaction at a time, followed by a reload which will do the deletion - so it's rolling.

docs/querying/api.md
```diff
@@ -1006,7 +1006,7 @@ $ curl -X POST \
 *New in v2.1 and supports PUT from v2.9*
 
 ### Clean Tombstones
-CleanTombstones removes the deleted data from disk and cleans up the existing tombstones. This can be used after deleting series to free up space.
+CleanTombstones removes the deleted data from disk and cleans up the existing tombstones. This can be used after deleting series to free up space. This process involves rerwiting existing blocks that have tombstones which requires free disk space (upto double the used space if all blocks have tombstones).
```
bwplotka, line 1009:
```suggestion
CleanTombstones removes the deleted data from disk and cleans up the existing tombstones. This can be used after deleting series to free up space. This process involves rewriting existing blocks that have tombstones (aka normal compaction). Note that like normal compaction, this requires some extra disk space (one more TSDB block maximum).
```

brian-brazil: I don't think this belongs here at all, it's an odd place to put general TSDB workings. We already have the Storage page for operational considerations.

GabrielCastro: @brian-brazil My experience purely as an operator, not a contributor and not someone with an in depth knowledge of TSDB was.

1. Performance was sluggish & we were low on disk space
1a. Turns out a bug in *our* app was had been generating a metric with very high cardinality in its labels
2. How do I delete these metrics; I get to https://prometheus.io/docs/prometheus/latest/querying/api/#delete-series
3. That then leads me to this section about CleanTombstones
4. We run out of disk space

While sure the details of TSDB are should be doc, It would have helped my experience and probably that of others if there's at least some warning here that more disk space will be used, maybe even a link to the correct section of the TSDB page.

brian-brazil: Cleaning tombstones doesn't use any extra disk compared to normal operations, you'd have ended up using that much disk space anyway on the next big compaction. So while unfortunate, that you ran out of resources wasn't due to using CleanTombstones but rather the cardinality issue.

GabrielCastro:
@roidelapluie
I can't seem to find where this rolling behaviour is:

The CleanTombstones function at the db level cleans all it's blocks [here](https://github.com/prometheus/prometheus/blob/c448ada63d83002e9c1d2c9f84e09f55a61f0ff7/tsdb/db.go#L1392), then only reloads once [at the end](https://github.com/prometheus/prometheus/blob/c448ada63d83002e9c1d2c9f84e09f55a61f0ff7/tsdb/db.go#L1400).

Unless there's something else I'm missing clean tombstones seems to not be rolling, a quick look found no places where CleanTombstones would be called from a loop outside.

This is inline with the disk usage pattern we saw during the operation.
image

brian-brazil:
Ah, I was looking a the general compaction code which is how tombstones are usually handled. That sounds like something we should improve then.

brian-brazil:
Silly me, it's an issue we need filed for this - not tagging this issue up. Unless you'd like to have a go at fixing this?

GabrielCastro:
I’ll make an attempt at fixing it 😃

brian-brazil:
I've filed #7957 for improving this, so I'm going to close this of.

PR #7549, opened 2020-07-10
Impove `promtool test rules` output diff on test failure

topliceanu:
### Why?

The `promtool test rules` utility produces error outputs that are sometimes hard to parse. This is because the expected and actual data structures being compared are printed using a `%#v` dump. It is hard to read the difference between the two structures when they are large - eg. a lot of labels and/or a lot of annotations.

### How?

I have imported `github.com/google/go-cmp@v0.5.0/cmp` and I'm using it to calculate and pretty-print a diff between the two structures under test on failure.

I don't feel strongly about using this library. Consider this as more of a PoC, I'm happy to proceed with this however the community feels its best: using a different library, adding a diff implementation to the source, etc.

An alternative is to defer the presentation to other tools by returning some kind of standardized output. 

Thanks for your feedback!

### DIFF
--- a/cmd/promtool/unittest.go
+++ b/cmd/promtool/unittest.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/go-kit/kit/log"
+	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 	yaml "gopkg.in/yaml.v2"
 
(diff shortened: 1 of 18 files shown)


topliceanu:
CC @simonpasquier

vanugrah:
+1.  Could we also get rid of the `/` delimiter around quotations as well for improved readability.

roidelapluie:
Hello @simonpasquier, can you have a look at this? Thanks! Also, it appears that it is not our usual way to compare structs.

dgl:
I looked at this a bit and one issue is using go-cmp prints out internal implementation details as it's really designed for unit tests, e.g.:

```
expr: "count(ALERTS) by (alertname, alertstate)", time: 4m,
        exp:"{alertname=\"InstanceDown\", alertstate=\"firing\"} 1E+00"
        got:"nil"
        diff (-exp, +got):
  []main.parsedSample(
-       {{Labels: s`{alertname="InstanceDown", alertstate="firing"}`, Value: 1}},
+       nil,
```

It's a bit confusing for the user to see the diff in the form of the internal structures that promtool uses, rather than PromQL syntax. I'm working on an alternative, first removing the use of `%#v` to make the output just clearer: #8064 and then I'm going to investigate some kind of diff.

codesome:
Closing in favor of https://github.com/prometheus/prometheus/pull/8064, thanks for the PR @topliceanu!

Issue #7548, opened 2020-07-10
memSeries: Iterator panic.


bwplotka:
Something we seen on our production:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x8f941a]

goroutine 473896 [running]:
github.com/prometheus/prometheus/tsdb.(*memSeries).iterator(0xc0d7afeb60, 0x0, 0xc1b6b7cfc0, 0xc000bd0180, 0x0, 0x0, 0x0, 0x0)
	/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20200629082805-315564210816/tsdb/head.go:2202 +0x84a
github.com/prometheus/prometheus/tsdb.(*safeChunk).Iterator(0xc1b76208d0, 0x0, 0x0, 0x41d7bcbe6a800001, 0xc1b761bd40)
	/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20200629082805-315564210816/tsdb/head.go:1513 +0x76
github.com/prometheus/prometheus/tsdb.(*chunkSeriesIterator).resetCurIterator(0xc1b761bd40)
	/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20200629082805-315564210816/tsdb/querier.go:1090 +0x6b
github.com/prometheus/prometheus/tsdb.newChunkSeriesIterator(...)
	/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20200629082805-315564210816/tsdb/querier.go:1083
github.com/prometheus/prometheus/tsdb.(*chunkSeries).Iterator(0xc1b761ba40, 0x0, 0x0)
	/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20200629082805-315564210816/tsdb/querier.go:891 +0xf6
github.com/prometheus/prometheus/tsdb.(*chainedSeriesIterator).Next(0xc1b76209c0, 0x1730c9725b8)
	/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20200629082805-315564210816/tsdb/querier.go:952 +0xaa
github.com/thanos-io/thanos/pkg/store.(*TSDBStore).encodeChunks(0xc0467bc040, 0x266ade0, 0xc1b76209c0, 0xffff, 0xc000af1880, 0x3, 0x4, 0xc1b7441c20, 0xf)
[...]

PR #7550, opened 2020-07-10
no panic the head memseries has chunks in it

krasi-georgiev:
Fixes https://github.com/prometheus/prometheus/issues/7548

I am now trying to figure out under what condition is this happening.

One theory is when the WAL is corrupted and loads series without any samples, but still unable to write a test that confirms this.

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -2208,10 +2208,8 @@ func (s *memSeries) iterator(id int, isoState *isolationState, chunkDiskMapper *
 				previousSamples += int(d.numSamples)
 			}
 		}
-		// mmappedChunks does not contain the last chunk. Hence check it separately.
-		if len(s.mmappedChunks) < ix {
-			previousSamples += s.headChunk.chunk.NumSamples()
-		} else {
+
+		if s.headChunk != nil {
 			totalSamples += s.headChunk.chunk.NumSamples()
 		}
 


krasi-georgiev:
yes that is why I said I am working on a test to reproduce it :) 
I have been playing with this for a while  and it indeed is connected with mapped head chunks.

codesome:
> There was something wrong in the original mmap changes itself

Because `len(s.mmappedChunks) < ix` is never possible, it always fell into the `else` block and `totalSamples` was incremented always, so the correctness is maintained, phew.

krasi-georgiev:
if you were able to reproduce it with a test please open a PR and I will review.

codesome:
I have not written any test, it is just a theory that is possible. I might not be able to write one at the moment.

krasi-georgiev:
ok in that case I will continue working on this.

beorn7:
@krasi-georgiev do you think you'll get this done soon enough so that we can include it in the 2.20 release?

krasi-georgiev:
> Took another look, and there is a case where it is possible for headChunk to be nil. Which is: After WAL replay, everything is in mmap chunks and no other sample in memory for that series. In that case the headChunk won't be created yet.


ok I figured it out, this can happen only when the wal is corrupted and the repair deletes some wal files. In normal operation, the wal always includes at least one extra sample which is not mapped thus creating a `headChunk`, but when the wal file that includes this extra sample is deleted then the `headChunk` is not created. See the test

tsdb/db_test.go
```diff
@@ -155,6 +155,65 @@ func TestDataAvailableOnlyAfterCommit(t *testing.T) {
...
+	// Append untill we have at least one mmaped chunk.
+	var expSamples []tsdbutil.Sample
+	var maxt int64
+	{
+		app := db.Appender()
+		for {
+			_, err := app.Add(labels.FromStrings("foo", "bar"), maxt, 0)
+			testutil.Ok(t, err)
+			err = app.Commit()
+			testutil.Ok(t, err)
```
codesome, line 175:
```suggestion
			testutil.Ok(t, app.Commit())
```

beorn7:
Sounds awesome to me. @roidelapluie @codesome do you agree?

If we have your approval for this change, I'd like to include it in the 2.20.0-rc.1.

tsdb/db_test.go
```diff
@@ -155,6 +155,65 @@ func TestDataAvailableOnlyAfterCommit(t *testing.T) {
...
+	db, closeFn := openTestDB(t, &Options{WALSegmentSize: 32 * 1024}, nil)
+	defer func() {
+		closeFn()
+	}()
+
+	// Append untill we have at least one mmaped chunk.
+	var expSamples []tsdbutil.Sample
+	var maxt int64
+	{
+		app := db.Appender()
```
codesome, line 170: Should we not have this inside the `for` loop? We have a change incoming which will make Commit/Rollback on the same appender invalid.

tsdb/db_test.go
```diff
@@ -155,6 +155,65 @@ func TestDataAvailableOnlyAfterCommit(t *testing.T) {
 	testutil.Equals(t, map[string][]tsdbutil.Sample{`{foo="bar"}`: {sample{t: 0, v: 0}}}, seriesSet)
 }
 
+// TestNoPanicAfterWALCorrutpion ensures that querying the db after a WAL corruption doesn't cause a panic.
+// https://github.com/prometheus/prometheus/issues/7548
+func TestNoPanicAfterWALCorrutpion(t *testing.T) {
+	db, closeFn := openTestDB(t, &Options{WALSegmentSize: 32 * 1024}, nil)
+	defer func() {
+		closeFn()
+	}()
```
codesome, line 164: Not a blocker, but we can start using the `t.Cleanup` method now. So this entire thing would turn out to be just `t.Cleanup(closeFn)` (I guess)

roidelapluie:
:+1:

tsdb/db_test.go
```diff
@@ -155,6 +155,65 @@ func TestDataAvailableOnlyAfterCommit(t *testing.T) {
...
+			testutil.Ok(t, err)
+			if len(files) > 0 {
+				break
+			}
+			maxt++
+		}
+		testutil.Ok(t, db.Close())
+	}
+
+	// Corrupt the second wal to trigger a repair.
```
codesome, line 187: How are we sure that there is a 2nd WAL file?

beorn7:
@krasi-georgiev could you rebase your changes on top of the release-2.20 branch?

tsdb/db_test.go
```diff
@@ -155,6 +155,65 @@ func TestDataAvailableOnlyAfterCommit(t *testing.T) {
...
+		testutil.Ok(t, err)
+		defer func() {
+			testutil.Ok(t, db.Close())
+		}()
+		testutil.Equals(t, 1.0, prom_testutil.ToFloat64(db.head.metrics.walCorruptionsTotal), "wal corruption count mismatch")
+
+		querier, err := db.Querier(context.TODO(), 0, maxt)
+		testutil.Ok(t, err)
+		seriesSet := query(t, querier, labels.MustNewMatcher(labels.MatchEqual, "", ""))
+		// The last sample should be missing as it was in the deleted WAL segment.
```
codesome, line 212: Again, how are we sure that only the last samples was in the 2nd file?

roidelapluie:
The rebate is not needed it seems

tsdb/db_test.go
krasi-georgiev: Because otherwise the `walFiles[1].Name()` will panic.

tsdb/db_test.go
krasi-georgiev: because the loop is stopped right after cutting a mapped chunk. This means that all but the last sample is loaded from the wal. Open to suggestions if you think there is a better way. To be honest I think there are plenty of other tests to ensure data correctness. The purpose of this test is a bit different so that is not even that important here.

tsdb/db_test.go
krasi-georgiev: yep good suggestion, updated

tsdb/db_test.go
krasi-georgiev: yep, thanks

tsdb/db_test.go
krasi-georgiev: hm ,that is strange, with a single appended it creates only a single wal file, investigating

krasi-georgiev:
Need a bit more time on the test as I suspect that there might be another case when this is triggered through the `CompactHead` method.

codesome:
Can we merge the fix and add test in follow up PR (to unblock 2.20 rc1, as the fix is harmless)

krasi-georgiev:
yes good idea I removed the test and will add in a separate PR feel free to merge as is.

beorn7:
> … feel free to merge as is.

For that, could you please rebase your changes on top of the release-2.20 branch?

krasi-georgiev:
done

beorn7:
Thank you everyone. Will merge once at least the tests are green (won't wait for build because that takes hours.)

tsdb/db_test.go
codesome: > This means that all but the last sample is loaded from the wal

Right! I missed that part. In that case we could corrupt saying "the last file" instead of a constant index (looks more generic). Will review your another PR.

PR #7551, opened 2020-07-10
Update dependencies

beorn7:
Note that by just running `make update-go-deps`, the K8s Go client was
set to `k8s.io/client-go v11.0.0+incompatible`. However, that doesn't
play well with `k8s.io/apimachinery v0.18.5`. I the manually changed
the Go client line to `k8s.io/client-go v0.18.5`, which made
everything work. I guess Go Modules got confused by the ginormous
v11.0.0 version tag. Or it is a problem that pulling k8s.io/client-go
with git results in a rather old repo without the v0.18.5
tag. github.com/kubernetes/client-go has all the right tags. I
actually don't understand how Go Modules still correctly figures out
the source from the `k8s.io/client-go v0.18.5` line.

If one of the reviewers could enlighten me, I'd much appreciate it.


beorn7:
And yeah, this is in preparation for the upcoming v2.20 release.

roidelapluie:
v11.0.0 > v0.18.5 that is why it was chosen. It seems a buggy tag.

brian-brazil:
:+1:

beorn7:
Thanks everyone.

PR #7553, opened 2020-07-11
Change Scrape Loop mtx to Mutex

roidelapluie:
It was still RWLock but we never use the read lock..


brian-brazil:
:+1:

PR #7556, opened 2020-07-11
Revert "Fix unknown symbol error during head compaction (#7526)"

bwplotka:
This reverts commit 30505a202a4c33ceeb10bfb3ba01e371a2a10906.

See rationales: https://github.com/prometheus/prometheus/issues/7373

We need to think about better fix, this makes things much worse. Nice unit test however, we can use it later one in next attempt


pracucci:
@codesome Would make sense to open an issue for the race you've noticed when adding the test in the PR #7526 (if not existing already).

codesome:
Before talking about the issues, some context for everyone around the initial fix and why was it wrong: The race that I found was actually synthetic and not reproducible in the real world. Reason: in my test, I try to add a sample for a new series and compact it at the same time. But in the real world, the compaction window won't cover the time range of the new series, hence with 0 chunks to compact from that series, that new series will be ignored, hence no issues here. 

What @bwplotka found in https://github.com/prometheus/prometheus/pull/7560 seems to be right as the error doesn't go away once it occurs and it is a lasting corruption until its restarted (but in my explanation before, it should go away in another compaction try, which it does not).

Lastly, the race that I had noticed was the out-of-order series error, and it might not be a race between append and compaction for the same reasons. Also, [there is an open issue for that](https://github.com/prometheus/prometheus/issues/5617).

PR #7559, opened 2020-07-11
Do not merge: Test showing race of compact vs append (with our reverted "fix")

bwplotka:
This is showcasing why our initial fix was not working: https://github.com/prometheus/prometheus/issues/7373 

Reproducing: https://github.com/prometheus/prometheus/issues/7373

cc @codesome 

Had to have this test without those hacks.. wonder if we can modify code to allow mocking for this :thinking: 


roidelapluie:
Can we close this?

PR #7560, opened 2020-07-11
Fixed race between compact (gc, populate) and head append causing unknown symbol error.

bwplotka:
Fixes https://github.com/prometheus/prometheus/issues/7373

Hopefully improved version of https://github.com/prometheus/prometheus/pull/7556
Why https://github.com/prometheus/prometheus/pull/7556 was not working is explained here: https://github.com/prometheus/prometheus/pull/7559

@codesome feel free to use this `branch` to test yourself things. It would be nice to add some test, but it's kind of impossible unless we rewrite things allowing better race mocks.

### DIFF
--- a/tsdb/compact.go
+++ b/tsdb/compact.go
@@ -656,7 +656,9 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta, 
 	defer func() {
 		var merr tsdb_errors.MultiError
 		merr.Add(err)
-		merr.Add(closeAll(closers))
+		if cerr := closeAll(closers); cerr != nil {
+			merr.Add(errors.Wrap(err, "close"))
+		}
 		err = merr.Err()
 		c.metrics.populatingBlocks.Set(0)
 	}()
@@ -708,7 +710,6 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta, 
 
 		s := newCompactionSeriesSet(indexr, chunkr, tombsr, all)
 		syms := indexr.Symbols()
-
 		if i == 0 {
 			set = s
 			symbols = syms
--- a/tsdb/compact.go
+++ b/tsdb/compact.go
@@ -657,7 +657,7 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta, 
 		var merr tsdb_errors.MultiError
 		merr.Add(err)
 		if cerr := closeAll(closers); cerr != nil {
-			merr.Add(errors.Wrap(err, "close"))
+			merr.Add(errors.Wrap(cerr, "close"))
 		}
 		err = merr.Err()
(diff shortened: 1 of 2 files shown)


brancz:
Changes make sense to me, but could use another eye as I haven't worked with this part of the code in a long time.

tsdb/head.go
```diff
@@ -1321,6 +1322,8 @@ func (h *Head) gc() {
 	symbols := make(map[string]struct{}, len(h.symbols))
 	values := make(map[string]stringset, len(h.values))
 
+	h.symMtx.Lock()
```
pracucci, line 1325: Is it correct reading `len(h.symbols)` and `len(h.values)` (above) outside of the lock?

pracucci, line 1325: Another question: in `getOrCreateWithID()` we do wrap `h.postings.Add()` with the lock, but in this function we do call `h.postings.Delete()` outside of the lock. Could this introduce another race?

pracucci: Answering to myself: no, it shouldn't, because what we need to protect is a de-sync between postings and symbols. The mere call to `h.postings.Delete()` doesn't touch symbols. It's in the `h.postings.Iter()` that we do build the map of updated symbols, so that's what we have to protect.

pracucci: I double checked and I believe we should take the lock before `make()` symbols and values above.

beorn7:
Thanks for all your efforts here.

FYI: It would be great to get this in for the upcoming 2.20 RC.

tsdb/head.go
bwplotka: agree

bwplotka: yup :+1:

bwplotka:
Addressed comments, let's take final look (:

tsdb/compact.go
```diff
@@ -656,7 +656,9 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
 	defer func() {
 		var merr tsdb_errors.MultiError
 		merr.Add(err)
-		merr.Add(closeAll(closers))
+		if cerr := closeAll(closers); cerr != nil {
+			merr.Add(errors.Wrap(err, "close"))
```
codesome, line 660: Oops, sorry for not noticing before
```suggestion
			merr.Add(errors.Wrap(eerr, "close"))
```

pracucci: Reason why `merr.Add(closeAll(closers))` is a less error prone approach. But let's not start a flame 🔥  😄

bwplotka: xD

bwplotka: BTW there should be vet for this

bwplotka: Also I cared about wrapping mostly here

bwplotka:
Merged cc @beorn7

beorn7:
\o/ Thanks everyone!

PR #7562, opened 2020-07-11
Reuse the same appender for report and scrape

### LINKED ISSUE #6950: Ingestion: Use the same appender for regular samples and reporting (`up` and friends)

roidelapluie:
Fix  #6950

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -951,8 +951,8 @@ mainLoop:
 	}
 }
 
-// scrapeAndReport is responsible for running a scape, and inserting with its
-// report in the storage, with a single appender.
+// scrapeAndReport performs a scrape and then appends the result to the storage
+// together with reporting metrics, all using the same appender.
 func (sl *scrapeLoop) scrapeAndReport(interval, timeout time.Duration, last time.Time, errc chan<- error) time.Time {
 	var (
 		start             = time.Now()
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -951,64 +951,8 @@ mainLoop:
 	}
 }
 
-func (sl *scrapeLoop) endOfRunStaleness(last time.Time, ticker *time.Ticker, interval time.Duration) {
-	// Scraping has stopped. We want to write stale markers but
-	// the target may be recreated, so we wait just over 2 scrape intervals
-	// before creating them.
-	// If the context is canceled, we presume the server is shutting down
-	// and will restart where is was. We do not attempt to write stale markers
-	// in this case.
-
-	if last.IsZero() {
-		// There never was a scrape, so there will be no stale markers.
-		return
-	}
-
-	// Wait for when the next scrape would have been, record its timestamp.
-	var staleTime time.Time
-	select {
-	case <-sl.parentCtx.Done():
-		return
-	case <-ticker.C:
-		staleTime = time.Now()
-	}
-
-	// Wait for when the next scrape would have been, if the target was recreated
-	// samples should have been ingested by now.
-	select {
-	case <-sl.parentCtx.Done():
-		return
-	case <-ticker.C:
-	}
-
-	// Wait for an extra 10% of the interval, just to be safe.
-	select {
-	case <-sl.parentCtx.Done():
-		return
-	case <-time.After(interval / 10):
-	}
-
-	// Call sl.append again with an empty scrape to trigger stale markers.
-	// If the target has since been recreated and scraped, the
-	// stale markers will be out of order and ignored.
-	var err error
-	app := sl.appender()
-	defer func() {
-		if err == nil {
-			err = app.Commit()
-		}
-		if err != nil {
-			app.Rollback()
-		}
-	}()
-	if _, _, _, err = sl.append(app, []byte{}, "", staleTime); err != nil {
-		level.Error(sl.l).Log("msg", "stale append failed", "err", err)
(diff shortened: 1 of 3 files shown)


scrape/scrape.go
```diff
@@ -1005,6 +951,83 @@ mainLoop:
 	}
 }
 
+// scrapeAndReport is responsible for running a scape, and inserting with its
```
beorn7, line 954: scape → scrape

scrape/scrape.go
```diff
@@ -1005,6 +951,83 @@ mainLoop:
...
+// report in the storage, with a single appender.
```
beorn7, line 955: Suggested rewording: "scrapeAndReport performs a scrape and then appends the result to the storage together with reporting metrics, all using the same appender."

roidelapluie:
Thanks for the review, I have addressed the comments.

roidelapluie:
@beorn7 I added another commit, which makes the end of run staleness consistent.

Also now it takes into consideration that Commit() will do a Rollback() if things go mad.

scrape/scrape_test.go
```diff
@@ -1928,3 +1970,66 @@ func TestCheckAddError(t *testing.T) {
 	sl.checkAddError(nil, nil, nil, storage.ErrOutOfOrderSample, nil, &appErrs)
 	testutil.Equals(t, 1, appErrs.numOutOfOrder)
 }
+
+func TestScrapeReportIsolation(t *testing.T) {
```
brian-brazil, line 1974: I'm not seeing how this is testing isolation, as there's nothing to ensure the query runs between the scrape ingestion and the report ingestion.

scrape/scrape.go
```diff
@@ -932,77 +932,99 @@ mainLoop:
...
-				sl.lastScrapeSize = len(b)
-			}
-		} else {
-			level.Debug(sl.l).Log("msg", "Scrape failed", "err", scrapeErr.Error())
-			if errc != nil {
-				errc <- scrapeErr
-			}
-		}
+// scrapeAndReport performs a scrape and then appends the result to the storage
+// together with reporting metrics, all using the same appender.
```
brian-brazil, line 955: "the same" is a little misleading as a 2nd appender will be created if there's a scrape error.

scrape/scrape.go
```diff
@@ -932,77 +932,99 @@ mainLoop:
...
 		}
-		last = start
+	}
 
-		select {
-		case <-sl.parentCtx.Done():
-			close(sl.stopped)
+	app := sl.appender()
+	var err error
+	defer func() {
```
brian-brazil, line 992: There's no early returns, so I don't see a need to use a defer here compared to putting this code at the end of the function - which would be clearer.

roidelapluie: We guard ourselves against potential panic()

scrape/scrape_test.go
roidelapluie: Queries run continuously. We should always have a multiple of 9 metrics.

That test is failing almost instantly without the code.

beorn7: Yes, I was assuming it would happen as @roidelapluie has confirmed now.

scrape/scrape.go
beorn7: But in case of a scrape error, the 1st appender is rolled back, so it will not append anything to the storage, so only one and the same appender will ever append any samples to the storage.

But if you have a better suggestion for the wording, I'm sure @roidelapluie will appreciate it.

roidelapluie: // scrapeAndReport performs a scrape and then appends the result to the storage
// together with reporting metrics, by using as few appenders as possible.
// In the happy scenario, a single appender is used.

beorn7:
Is this good to go @brian-brazil ? Would be great to have this in the release.

beorn7:
Or in case @brian-brazil is busy, I guess @codesome or @krasi-georgiev are qualified here, too, as it is close to TSDB.

scrape/scrape_test.go
brian-brazil: The test name is misleading, as this doesn't test anything about isolation per-se. What it's testing is that it's all done in one append.

beorn7: @roidelapluie that name should be easy to fix.
@brian-brazil is that the only remaining concern in this PR?

brian-brazil: Yes, the rest I can live with.

roidelapluie: Renamed. I originally named it like this because if the underlying appender does not support isolation, the test failed as well.

brian-brazil:
:+1:

beorn7:
:smile: Thanks everyone.

PR #7563, opened 2020-07-12
Make discovery independent of the azure dependencies

roidelapluie:
For #7107

@bwplotka what do you think of this approach?

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -52,6 +52,8 @@ import (
 	promlogflag "github.com/prometheus/common/promlog/flag"
 	"github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/discovery"
+
+	_ "github.com/prometheus/prometheus/discovery/azure"
 	sd_config "github.com/prometheus/prometheus/discovery/config"
 	"github.com/prometheus/prometheus/notifier"
 	"github.com/prometheus/prometheus/pkg/labels"
--- a/discovery/manager.go
+++ b/discovery/manager.go
@@ -27,7 +27,6 @@ import (
 	sd_config "github.com/prometheus/prometheus/discovery/config"
 	"github.com/prometheus/prometheus/discovery/targetgroup"
 
-	"github.com/prometheus/prometheus/discovery/azure"
 	"github.com/prometheus/prometheus/discovery/consul"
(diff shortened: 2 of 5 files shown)


cmd/prometheus/main.go
```diff
@@ -52,6 +52,8 @@ import (
 	promlogflag "github.com/prometheus/common/promlog/flag"
 	"github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/discovery"
+
+	_ "github.com/prometheus/prometheus/discovery/azure"
```
bwplotka, line 56: ?

discovery/manager.go
```diff
@@ -418,7 +443,11 @@ func (m *Manager) registerProviders(cfg sd_config.ServiceDiscoveryConfig, setNam
 	}
 	for _, c := range cfg.AzureSDConfigs {
 		add(c, func() (Discoverer, error) {
-			return azure.NewDiscovery(c, log.With(m.logger, "discovery", "azure")), nil
+			d, err := GetDiscoverer("azure")
```
bwplotka, line 446: Can we adjust `GetDiscoverer` in a way that would not require this anonymous function? :thinking:

discovery/manager.go
```diff
@@ -78,12 +77,38 @@ var (
...
+		return fmt.Errorf("provider %s is already registered", name)
+	}
+	providers[name] = d
+	return nil
+}
+
+// GetDiscoverer return a discoverer that was registered with
+// RegisterDiscoverer.
+func GetDiscoverer(name string) (func(interface{}, log.Logger) Discoverer, error) {
+	registerMtx.Lock()
```
bwplotka, line 103: I don't think init is ever done concurrently so we don't need any mutex here and above

roidelapluie: Do you think that all the init() functions are called in one go routing?

bwplotka: yea I think so

bwplotka:
So BTW it's all about what is the requirements / thing we want achieve.

I don't think we can avoiding importing all of those deps from `main`. However we want to remove spread from packages that can be importable. To achieve this I don't think we need dynamic import registering. I think there is a type safe method to do this :hugs:

brian-brazil:
Anonymous imports like this are the standard Go way to do this, and we use it in other places.

PR #7566, opened 2020-07-12
tsdb: test for goroutine leaks

### LINKED ISSUE #6925: Add `fortytw2/leaktest` to all unit tests that involves go routines.
Hi, :wave: 

We found this super useful in the [Thanos](https://thanos.io) project, it allows us to find memory leaks in the form of not closed go routines pretty fast. Thanks for @gotjosh who reminds me to put a request for that. :hugs:  

**Help wanted!** 

NOTE1: It does not work with `t.Parallel` until this: https://github.com/fortytw2/leaktest/issues/4
NOTE2: At some point, Uber created its own lib which does the same (without even trying to contribute to the existing https://github.com/fortytw2/leaktest) here: https://github.com/uber-go/goleak Since https://github.com/fortytw2/leaktest works well, is mature and have friendly maintainers, I would say let's stick to that. (:

roidelapluie:
Fix #6925
Supersedes #6928  which seems somehow abandoned.


roidelapluie:
@bwplotka I ended up using uber's goleak as it is less intrusive for the code. It show a bunch of places in the tests where we did not close everything properly.

go.mod
```diff
@@ -71,6 +71,7 @@ require (
 	github.com/uber/jaeger-lib v2.2.0+incompatible
 	go.mongodb.org/mongo-driver v1.3.2 // indirect
 	go.uber.org/atomic v1.6.0 // indirect
+	go.uber.org/goleak v1.0.0
```
bwplotka, line 74: I would vote for using https://github.com/fortytw2/leaktest

1. It has less toxic community last time I checked
2. We are working on improvements that would allow running it with parallel etc.

WDYT?

bwplotka: I can agree goleak is less intrusive though. leaktest does not have it :/

Still it does not show which test leaked, just overall leak, right?

roidelapluie: Given that goleak runs at the end of all tests, it works with tests in parallel too.

I find it great that it just works at the package level, WDYT?

For the community side, can you be more explicit?

roidelapluie:
Did I just trigger a race condition in GitHub? #7566 #7565

go.mod
roidelapluie: > I can agree goleak is less intrusive though. leaktest does not have it :/
> 
> Still it does not show which test leaked, just overall leak, right?

Yes but the README explains how to find out:

https://github.com/uber-go/goleak#determine-source-of-package-leaks

roidelapluie:
Let's wait for #7551 first

roidelapluie:
rebased

tsdb/wal_test.go
```diff
@@ -92,6 +92,7 @@ func TestSegmentWAL_Truncate(t *testing.T) {
 	}()
 
 	w, err := OpenSegmentWAL(dir, nil, 0, nil)
+	defer func(wal *SegmentWAL) { testutil.Ok(t, wal.Close()) }(w)
```
codesome, line 95: This should come after checking the error

roidelapluie:
I guess we are good to go now

PR #7569, opened 2020-07-13
Arbitrary rule & group loading

owen-d:
This PR extends the rules package in two ways while leaving its current functionality unaltered.

First, It allows loading rules from arbitrary sources. This is accomplished by adding a new field to the ManagerOptions which is responsible for loading and parsing rules. While Prometheus retains fidelity to its local file approach, other projects which use the rules package can now optionally load rules from different sources (such as across network boundaries) by using this interface.

The second change narrows the parsed query of both the `AlertingRule` & `RecordingRule` types from `parser.Expr` to `fmt.Stringer`, bringing it in line with the `QueryFunc` type used to evaluate these rules which already accepts a string argument. This change also enables other non-default use cases in downstream projects such as parsing [LogQL](https://github.com/grafana/loki/blob/master/docs/logql.md) queries in [Loki](https://github.com/grafana/loki).

Regarding the implementation here, I chose to introduce `Parsed{Rule,Group}` intermediate representations in order to avoid polluting the new `GroupLoader` interface with the need to be aware of manager internals such as the `done` channel, the `ManagerOptions` field, etc. The manager's  pre-existing `LoadGroups` function is then responsible for mapping these IRs into the rule types. Please let me know if there is another preferred approach.

I believe these are relatively unobtrusive changes to the codebase which leave the default implementations intact while enabling downstream projects to reap the benefits of the rules package. Allowing Cortex to horizontally scale rule evaluation via a centralized object store without the need to regularly synchronize local files is another example benefit in addition to what we hope to achieve in Loki.


rules/alerting.go
```diff
@@ -113,7 +113,7 @@ type AlertingRule struct {
 	// The name of the alert.
 	name string
 	// The vector expression from which to generate alerts.
-	vector parser.Expr
+	vector fmt.Stringer
```
brian-brazil, line 116: This seems like an odd choice of type, why not a string? Why does this need to change at all? 

We've also had numerous bugs in the past from working with promql as rendered string rather than a parsed expression in recording rules, so we should not go back to that.

rules/loading.go
```diff
@@ -0,0 +1,94 @@
+// Copyright 2013 The Prometheus Authors
```
brian-brazil, line 1: 2020

rules/loading.go
```diff
@@ -0,0 +1,94 @@
...
+
+// Grouploader is an interface for loading and parsing rules from arbitrary sources.
+type GroupLoader interface {
+	// Load resolves a list of parsed rules from a set of identifiers.
+	Load(identifiers ...string) ([]*ParsedGroup, []error)
+}
+
+// ParsedGroup binds a set of ParsedRules with an evaluation interval.
+// 0 indicates that the global default interval should be used.
+type ParsedGroup struct {
```
brian-brazil, line 33: We already have this information in the RuleGroup&Rule structs - there's no need to add an additional basically identical set of structs

rules/manager.go
```diff
@@ -977,55 +980,47 @@ func (m *Manager) LoadGroups(
 
 	shouldRestore := !m.restored
 
-	for _, fn := range filenames {
```
brian-brazil, line 980: What I'd suggest doing is just factoring out these 5 lines of code, rather than changing the overall structure of the code.

rules/alerting.go
owen-d: This could definitely be `string`! I chose `Stringer` as it was the smallest code change that illustrated what I wanted. 

As for why it needs to change, it allows downstream consumers to use different types here that can be coercible to a string. I'm particularly interested in parsing LogQL queries in Loki in order to alert based upon logs.

owen-d:
Thanks for the initial review Brian. I'll update this with the changes you suggested.

owen-d:
I've rebased and made the change from `fmt.Stringer` -> `string` as requested which required many small diffs which are all variants of 
```
-	vector parser.Expr
+	vector string
```

Does this look good?

rules/manager.go
gouthamve: Agree with this, I think we can get away with a smaller diff if we just factor out `rgs, errs := rulefmt.ParseFile(fn)`

We could make an interface for the function `LoadGroups(id string) rgs, errors` and that would simplify things. And then in the `LoadGroups` function itself do the validation of
```
expr, err := parser.ParseExpr(r.Expr.Value)
if err != nil {
  return nil, []error{errors.Wrap(err, fn)}
}
```

Would that work @owen-d?

rules/loading.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"github.com/go-kit/kit/log"
+	"github.com/pkg/errors"
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/pkg/rulefmt"
+	"github.com/prometheus/prometheus/promql/parser"
+)
+
+// Grouploader is an interface for loading and parsing rules from arbitrary sources.
+type GroupLoader interface {
```
cstyan, line 27: style nit; I'm not sure if we're consistent about this in `prometheus/prometheus` but I'd tend to define interfaces where they're used (in manager.go in this case)

rules/manager.go
cstyan: This makes sense. We're probably going to want to keep everything in the `for _, rg := range rgs.Groups` loop consistent across group loader types. If a case comes up where we need to change that we can do so, but for now the loader can just return rule groups/errors

owen-d:
Thanks for the review all -- I think this newest commit is minimally intrusive & allows extending both the rule loading and query parsing functionality.

rules/alerting.go
brian-brazil: This still looks incorrect. Why does this need to change at all?

rules/manager.go
```diff
@@ -969,6 +975,22 @@ func (m *Manager) Update(interval time.Duration, files []string, externalLabels
...
+
+// FileLoader is the default GroupLoader implementation. It defers to rulefmt.ParseFile
+// and parser.ParseExpr
+type FileLoader struct{}
+
+func (FileLoader) Load(identifier string) (*rulefmt.RuleGroups, []error) {
+	return rulefmt.ParseFile(identifier)
+}
+
+func (FileLoader) Parse(query string) (fmt.Stringer, error) { return parser.ParseExpr(query) }
```
brian-brazil, line 992: I would expect a function called "Parse" to parse - not to return a string.

owen-d: the Parse function went from returning a `parser.Expr` to a `fmt.Stringer`. It still parses the query -- I've just narrowed the interface. I did this for two reasons:
1) Because the parsed expression is only used for it's `String()` method. Narrowing this interface is easy and better reflects how the return value is used.
2) It allows consumers of this interface to substitute other underlying types asides `parser.Expr` that implement `fmt.Stringer`.

brian-brazil: Why are we converting a string to a string?

owen-d: We're not; rather we're converting a string to a type which implements `fmt.Stringer`, which is still a valuable parsing step to ensure the query is valid. Previously this converted a `string` to a `parser.Expr`, but _only the `String` method was used_. This is similar to reducing `io.ReadWriter` to `io.Reader` when only the `Read` method is used. Does this make sense? We're still parsing the string, but now the parsing step is substitutable and doesn't bind the output type to `parser.Expr`.

owen-d: For some additional background (pulled from the initial PR description),
> This change also enables other non-default use cases in downstream projects such as parsing LogQL queries in Loki.

brian-brazil: Poking at the code, that we only use String is a long standing wart that I thought we had already fixed by now. This type should stay as-is, and we should not be doing an unnecessary round-trip through the parser here.

owen-d: Can you elaborate on 
> we should not be doing an unnecessary round-trip through the parser here.


As far as I can tell, this isn't an unnecessary round trip as there is no other validation of the query. That being said, I'll set it back to `parser.Expr`.

roidelapluie:
On 22 Jul 06:59, Owen Diehl wrote:
> @owen-d commented on this pull request.
> 
> 
> 
> > @@ -969,6 +975,22 @@ func (m *Manager) Update(interval time.Duration, files []string, externalLabels
>  	return nil
>  }
>  
> +// GroupLoader is responsible for loading rule groups from arbitrary sources and parsing them.
> +type GroupLoader interface {
> +	Load(identifier string) (*rulefmt.RuleGroups, []error)
> +	Parse(query string) (fmt.Stringer, error)
> +}
> +
> +// FileLoader is the default GroupLoader implementation. It defers to rulefmt.ParseFile
> +// and parser.ParseExpr
> +type FileLoader struct{}
> +
> +func (FileLoader) Load(identifier string) (*rulefmt.RuleGroups, []error) {
> +	return rulefmt.ParseFile(identifier)
> +}
> +
> +func (FileLoader) Parse(query string) (fmt.Stringer, error) { return parser.ParseExpr(query) }


I would advice loki to use the replace directive of go.mod:

replace github.com/prometheus/prometheus/promql/parser =>
github.com/loki/loki/parser

> 
> Can you elaborate on 
> > we should not be doing an unnecessary round-trip through the parser here.
> 
> 
> As far as I can tell, this isn't an unnecessary round trip as there is no other validation of the query. That being said, I'll set it back to `parser.Expr`.
> 
> -- 
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/prometheus/prometheus/pull/7569#discussion_r458813253

-- 
 (o-    Julien Pivotto
 //\    Open-Source Consultant
 V_/_   Inuits - https://www.inuits.eu

roidelapluie:
parser.Expr is an interface so I do not see that as an issue for downstream, e.g. loki?

rules/manager.go
brian-brazil: We parse the query and then render it again, and then parse the rendered value when we run the rule. If we want to do just validation, it shouldn't involve a return value or that rendering step.

brian-brazil:
Thanks!

owen-d:
@roidelapluie 
> parser.Expr is an interface so I do not see that as an issue for downstream, e.g. loki?

Yes, but it's not actually implementable without some embedding because you the `Expr` interface depends on an unexported signature:
https://github.com/prometheus/prometheus/blob/master/promql/parser/ast.go#L78-L79

For instance, this will not compile:

```
type Tst struct{}

func (Tst) String() string                      { return "" }
func (Tst) PositionRange() parser.PositionRange { return parser.PositionRange{} }
func (Tst) Type() parser.ValueType              { return parser.ValueTypeMatrix }
func (Tst) expr()                               {}

func compiles(x parser.Expr) parser.Expr { return x }

var Ok = compiles(Tst{})
```

->

```
cannot use Tst literal (type Tst) as type parser.Expr in argument to compiles:
	Tst does not implement parser.Expr (missing parser.expr method)
		have expr()
		want parser.expr()
```

We'll have to do something like

```
type EmbeddedExpr struct {
	parser.StringLiteral
	ActualStuffWeCareAbout
}
```
and then override the `String` method.

brian-brazil:
In general we're okay with making things public to all for extension, though that particular one might require additional thought.

roidelapluie:
Yeah the I think that just making Expr public is what you are looking for.

owen-d:
> I think that just making Expr public is what you are looking for.

Are you interested in accepting a PR for that @roidelapluie @brian-brazil?

roidelapluie:
I am. I think that Expr() is unique enough to guarantee that nothing unexpected implements the interface.

Issue #7469, opened 2020-06-26
Getting mmap: invalid argument error for 2.19.0 version


rajanikantpatel:
getting following error with prometheus 2.19.0 version.

level=error ts=2020-06-26T09:51:32.496Z caller=main.go:758 err="opening storage failed: mmap files, file: /prometheus/data/chunks_head/000108: mmap: invalid argument

following is the arguments 
- >-
              --web.external-url=https://prometheus-dev-kc1-iad.saase2e.pitneycloud.com
            - '--config.file=/etc/prometheus/prometheus.yml'
            - '--storage.tsdb.path=/prometheus/data'
            - '--storage.tsdb.retention.time=30d'
            - '--web.enable-lifecycle'
            - '--storage.tsdb.no-lockfile'
            - '--log.level=warn'

brian-brazil:
What operating system and filesystem are you using? What does strace say is happening?

> --storage.tsdb.no-lockfile

This is a dangerous flag that you should not use.

DandyDeveloper:
@brian-brazil @rajanikantpatel I experienced a similar problem today in 2.19.1.

Issue was very similar to this: https://github.com/prometheus/prometheus/issues/7412

I think the bugfix isn't working as we had an empty chunk file that basically needed to be removed and my understanding is the fix in the above issue removes those loose chunk files.

After removing the file in question that the error is complaining about, things started up as normal.

roidelapluie:
> @brian-brazil @rajanikantpatel I experienced a similar problem today in 2.19.1.
> 
> Issue was very similar to this: #7412
> 
> I think the bugfix isn't working as we had an empty chunk file that basically needed to be removed and my understanding is the fix in the above issue removes those loose chunk files.
> 
> After removing the file in question that the error is complaining about, things started up as normal.

Thanks, sounds like a different issue. Can you please open a new bug please?

codesome:
> we had an empty chunk file

0 bytes? There can be 0 chunks, but 0 bytes in a file seems unlikely as we write headers as soon as we create a file.

But checking the `mmap: invalid argument` error from this issue description, that (also) comes when we try to m-map a completely empty file, so I think _somewhere_ we are ending up creating empty files (even without headers). I will check that.

codesome:
@rajanikantpatel did the Prometheus shutdown gracefully when you restarted?

DandyDeveloper:
[...]

PR #7573, opened 2020-07-14
Avoid empty mmap files by using .tmp files to write headers

codesome:
Fixes https://github.com/prometheus/prometheus/issues/7469

### DIFF
--- a/tsdb/chunks/chunks.go
+++ b/tsdb/chunks/chunks.go
@@ -217,7 +217,7 @@ func (w *Writer) cut() error {
 	return nil
 }
 
-func cutSegmentFile(dirFile *os.File, magicNumber uint32, chunksFormat byte, allocSize int64) (headerSize int, newFile *os.File, seq int, err error) {
+func cutSegmentFile(dirFile *os.File, magicNumber uint32, chunksFormat byte, allocSize int64) (headerSize int, newFile *os.File, seq int, returnErr error) {
 	p, seq, err := nextSequenceFile(dirFile.Name())
 	if err != nil {
 		return 0, nil, 0, errors.Wrap(err, "next sequence file")
@@ -227,6 +227,18 @@ func cutSegmentFile(dirFile *os.File, magicNumber uint32, chunksFormat byte, all
 	if err != nil {
 		return 0, nil, 0, errors.Wrap(err, "open temp file")
 	}
+	defer func() {
+		if returnErr != nil {
+			var merr tsdb_errors.MultiError
+			merr.Add(returnErr)
+			if f != nil {
+				merr.Add(f.Close())
+			}
+			// Calling RemoveAll on a non-existent file does not return error.
+			merr.Add(os.RemoveAll(ptmp))
+			returnErr = merr.Err()
+		}
+	}()
 	if allocSize > 0 {
 		if err = fileutil.Preallocate(f, allocSize, true); err != nil {
 			return 0, nil, 0, errors.Wrap(err, "preallocate")
@@ -248,6 +260,7 @@ func cutSegmentFile(dirFile *os.File, magicNumber uint32, chunksFormat byte, all
 	if err := f.Close(); err != nil {
 		return 0, nil, 0, errors.Wrap(err, "close temp file")
 	}
+	f = nil
 
 	if err := fileutil.Rename(ptmp, p); err != nil {
 		return 0, nil, 0, errors.Wrap(err, "replace file")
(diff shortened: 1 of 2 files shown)


beorn7:
It would be great to get thin into the 2.20 release. I've requested reviews from a few people that might be qualified. I'll also give it a try myself once I have reached that point in my review queue.

roidelapluie:
> I don't get what this gives us. The process of writing headers is probably quite fast, so unlikely that this is causing empty files no? so I don't think we solve anything here  In the same time process might be finished etc at line 256 ?

Prometheus is so widespread now that even things that happen once in a million time happen every day.

roidelapluie:
If write fails, you would have an empty dir.

codesome:
> I don't get what this gives us. The process of writing headers is probably quite fast, so unlikely that this is causing empty files no? so I don't think we solve anything here thinking In the same time process might be finished etc at line 256 ?

If you look at the linked issue, it has been experienced by >1 person not long after the release, so it's not that rare. And disk writes/syncs are OS dependant and might not be flushed immediately?

Btw, if it happens (empty files), Prometheus will fail to start. So, it fixes that.

tsdb/chunks/chunks.go
```diff
@@ -242,7 +243,23 @@ func cutSegmentFile(dirFile *os.File, magicNumber uint32, chunksFormat byte, all
...
+
+	if err := fileutil.Rename(ptmp, p); err != nil {
+		return 0, nil, 0, errors.Wrap(err, "replace file")
+	}
+
+	f, err = os.OpenFile(p, os.O_WRONLY, 0666)
+	if err != nil {
+		return 0, nil, 0, errors.Wrap(err, "open final file")
+	}
+	// Skip the header for the further writes.
```
beorn7, line 260: Nit: Delete "the".

tsdb/chunks/chunks.go
```diff
@@ -220,19 +220,20 @@ func (w *Writer) cut() error {
...
 	if err != nil {
-		return 0, nil, 0, err
+		return 0, nil, 0, errors.Wrap(err, "next sequence file")
 	}
-	f, err := os.OpenFile(p, os.O_WRONLY|os.O_CREATE, 0666)
+	ptmp := p + ".tmp"
+	f, err := os.OpenFile(ptmp, os.O_WRONLY|os.O_CREATE, 0666)
 	if err != nil {
-		return 0, nil, 0, err
+		return 0, nil, 0, errors.Wrap(err, "open temp file")
```
beorn7, line 228: I think, after this block (i.e. after line 229), we should have a `defer` for closing and deleting the tmp file in case any of the errors below trigger, to not leak file descriptors and leave tmp files behind. (In case the closing or renaming has already happened, the close or delete call will return an error, but we don't mind then. You could also guard the deferred closing and deletion with a boolean to avoid spurious calls.)

codesome: Good point, I have updated the PR

tsdb/chunks/chunks.go
```diff
@@ -217,22 +217,32 @@ func (w *Writer) cut() error {
...
 	if err != nil {
-		return 0, nil, 0, err
-	}
+		return 0, nil, 0, errors.Wrap(err, "open temp file")
+	}
+	defer func() {
+		if returnErr != nil {
+			var merr tsdb_errors.MultiError
+			merr.Add(returnErr)
+			// Calling RemoveAll on a non-existent file does not return error.
```
beorn7, line 234: I think we still need to call `f.Close()` to not leak file descriptors.

tsdb/chunks/chunks.go
```diff
@@ -217,22 +217,33 @@ func (w *Writer) cut() error {
...
 	if err != nil {
-		return 0, nil, 0, err
-	}
+		return 0, nil, 0, errors.Wrap(err, "open temp file")
+	}
+	defer func() {
+		if returnErr != nil {
+			var merr tsdb_errors.MultiError
+			merr.Add(returnErr)
+			merr.Add(f.Close())
```
beorn7, line 234: Sorry to beat this seemingly dead horse. But `f` can be `nil` if opening the final file in L266 fails. You are reusing `f` there, which is probably OK as you actually want to close the new `f` in case there is an error in L271.

I think you have to add a check here for `f != nil`. And then you could set `f = nil` in L261 so that you don't try to close the closed file in case the Rename in 262 fails.

codesome: Done! Thanks

beorn7:
Looks like this has now been comprehensively reviewed. (o:
Merging now to get this into the RC.

Issue #7574, opened 2020-07-14
Checkpoint WAL in Head only once in a DB.Compact() call


codesome:
It is possible that Head compaction can fail continuously for various reasons (disk issues, disk capacity, unsolved corruptions in the head, etc), in such cases a lot of WAL would be piled up until the issue is fixed.

When Prometheus starts finally, let's say we have 24h of data in Head not compacted, this will lead to 11 Head compactions (22h of data) in total one after another (in a single `DB.Compact()` call).

Currently, for each compaction, we truncate the Head, which also results into creating a checkpoint. And as its 24h now, the checkpoint is huge (2/3 of 24h data). With 11 compactions like this, it is a lot of re-write of the same data again and again into the checkpoint and the entire process becomes _very_ slow.

So I am proposing to truncate the Head only once in a `DB.Compact()` method so that in the example above, it would lead to a single checkpoint creation.

One of the issues with this is Prometheus coming down in between the compaction process and hence the truncation not happening. So with this proposal, additionally, we could check for possible truncation on startup. (Update: Created https://github.com/prometheus/prometheus/issues/7575 for it)

bwplotka:
cc @kakkoyun as we investigate some issue near this place.

kakkoyun:
> One of the issues with this is Prometheus coming down in between the compaction process and hence the truncation not happening. 

@codesome Seems like we've been hit by this last part in Thanos Receive. We hit this case when a shutdown process terminated prematurely, where we compact the head block. As a result, a block created but the head is not truncated. We ended up creating overlapping blocks. It'd be really nice to have an additional mechanism to prevent it.

codesome:
Opened https://github.com/prometheus/prometheus/issues/7575 dedicated to truncating WAL on startup

Harkishen-Singh:
Will try to take up this.

codesome:
There is already https://github.com/prometheus/prometheus/pull/7579, maybe you could review it?

PR #7579, opened 2020-07-15
Create a checkpoint only at the end of Compact call

codesome:
Fixes https://github.com/prometheus/prometheus/issues/7574 https://github.com/prometheus/prometheus/issues/6359


codesome:
This PR moves the checkpointing to the end of all head compaction in the loop instead of each head compaction.

(If you saw this PR before this comments, it was truncating the head at the end, but that has problems as we need to advance the mint of the head. So, now, we are only moving the checkpointing to the end)

beorn7:
This looks good to me as only being semi-qualified in TSDB matters.

@bwplotka @krasi-georgiev would be great to get this reviewed today so that I can include it into the 2.20 RC today.

codesome:
@beorn7 if this PR gets delayed, feel free to go ahead with 2.20 RC as this is merely an enhancement which can go later too.

pstibrany:
Previously `prometheus_tsdb_wal_truncate_duration_seconds` covered checkpoint time as well. After this change, it will not. I think we need new metric for checkpoint times.

codesome:
@pstibrany It still does. What I have done is just split `Truncate()` into `Truncate()` and `performCheckpoint()`, and the `prometheus_tsdb_wal_truncate_duration_seconds` calculation covers the entire `performCheckpoint()` (WAL truncation and checkpoint creation)

pstibrany:
> @pstibrany It still does. What I have done is just split `Truncate()` into `Truncate()` and `performCheckpoint()`, and the `prometheus_tsdb_wal_truncate_duration_seconds` calculation covers the entire `performCheckpoint()` (WAL truncation and checkpoint creation)

Ah, right. I was thinking about previous version of the code. Sorry and thanks for correcting me.

tsdb/db.go
```diff
@@ -740,6 +740,16 @@ func (db *DB) Compact() (err error) {
 	}()
 	// Check whether we have pending head blocks that are ready to be persisted.
 	// They have the highest priority.
+	checkpointTime := int64(math.MinInt64)
+	checkpointAttempted := false
+	defer func() {
+		if err != nil && !checkpointAttempted {
```
pracucci, line 746: If `db.stopc` is closed, no checkpoint is performed because of the check `err != nil`. Is it intended? Previous code would have run it.

tsdb/db.go
```diff
@@ -760,52 +770,82 @@ func (db *DB) Compact() (err error) {
...
+		t, err := db.compactHead(head)
+		if err != nil {
 			return err
 		}
+		if t > checkpointTime {
+			checkpointTime = t
+		}
+	}
+
+	checkpointAttempted = true
```
pracucci, line 782: If head was not compactable at first `for` iteration, is it correct we try to perform checkpoint?

tsdb/db.go
```diff
@@ -760,52 +770,82 @@ func (db *DB) Compact() (err error) {
...
 
+func (db *DB) performHeadCheckpoint(t int64) error {
+	if blocks := db.Blocks(); len(blocks) > 0 {
+		blockMaxt := blocks[len(blocks)-1].Meta().MaxTime
+		if blockMaxt > t {
+			t = blockMaxt
+		}
+	}
+
+	if t == math.MinInt64 {
```
pracucci, line 798: Uhm, zero value of `t` is `0`, not `math.MinInt64`. I think you have to initialise `t = math.MinInt64` at the beginning of the function, no?

roidelapluie: t is a parameter

tsdb/db.go
codesome: Good point! We need to truncate.

tsdb/db.go
codesome: Right, we should not :) thanks for pointing out

tsdb/db.go
codesome: `t` is an argument (hence should not initialize in the method), and we are using `math.MinInt64` as zero value (which comes from `compactHead` and `Compact`)

tsdb/db.go
codesome: s/truncate/checkpoint

tsdb/db.go
pracucci: 🤦

codesome: As Golang's `time.Unix(x,y)` method accepts 0 and even negative numbers, I thought of keeping the zero value as minint64.

codesome:
I have refactored a bit to keep the original behaviour of `Head.Truncate()` method (as it creates a lot of changes and confusion in unit tests), and instead introduced `Head.TruncateInMemory()` and `Head.TruncateWAL()` which is used during compaction here.

roidelapluie:
Hello @bwplotka, can you please review the last changes here? Thanks!

pstibrany:
> Have you considered adding few tests (only WAL no blocks no gap in the WAL, with gap in the WAL, blocks and WAL) and measuring that the expected number of checkpoints occurred?

Quoting this to make sure it's not forgotten ^ (it would increase the confidence in the change).

codesome:
Finally found the bug causing the tests to fail after few hours of investigation, and the fix is a 1 liner :) https://github.com/prometheus/prometheus/pull/7579/commits/7c3dcba4760f474b433b487d80924a92b3298fd7

Adding more tests now.

codesome:
PR is ready for another review. Should be much simpler to parse now, with added tests.

tsdb/head.go
```diff
@@ -781,8 +781,21 @@ func (h *Head) removeCorruptedMmappedChunks(err error) map[uint64][]*mmappedChun
 	return mmappedChunks
 }
 
-// Truncate removes old data before mint from the head.
+// Truncate removes old data before mint from the head and WAL.
 func (h *Head) Truncate(mint int64) (err error) {
+	initialize := h.MinTime() == math.MaxInt64
+	dontTruncateWAL := initialize || (!initialize && h.MinTime() >= mint)
```
pracucci, line 782:
```suggestion
	dontTruncateWAL := initialize || h.MinTime() >= mint
```

tsdb/db.go
```diff
@@ -839,21 +884,22 @@ func (db *DB) compactHead(head *RangeHead) (err error) {
...
 	if (uid == ulid.ULID{}) {
-		// Compaction resulted in an empty block.
-		// Head truncating during db.reload() depends on the persisted blocks and
 		// in this case no new block will be persisted so manually truncate the head.
-		if err = db.head.Truncate(maxt); err != nil {
-			return errors.Wrap(err, "head truncate failed (in compact)")
+		// Head truncating during db.reload() depends on the persisted blocks and
+		// Compaction resulted in an empty block.
+		if err = db.head.TruncateInMemory(maxt); err != nil {
+			return walTruncationTime, errors.Wrap(err, "head truncate failed (in compact)")
```
pracucci, line 897: This will return `walTruncationTime = maxt` but, unless I'm missing something, the returned `walTruncationTime` is never used if there's an error. I'm wondering if the could would be easier to follow if you always `return math.MinInt64` and at the end of this function you `return maxt, nil`

tsdb/db.go
```diff
@@ -805,31 +816,65 @@ func (db *DB) Compact() (err error) {
...
+		if err != nil {
 			return err
 		}
+		if t > walTruncationTime {
+			walTruncationTime = t
+		}
+		headCompacted = true
+	}
+
+	if headCompacted {
```
pracucci, line 829: What do you think to add a code comment to explain why we truncate the WAL only at the end?

pracucci (commented):
Thanks @codesome for working on this improvement! To my limited knowledge of TSDB, changes make sense. The test also cover the 2 cases we've seen, so 👍 .

tsdb/db.go
```diff
@@ -805,31 +816,65 @@ func (db *DB) Compact() (err error) {
...
+		walTruncationAttempted = true
+		if err := db.truncateWAL(walTruncationTime); err != nil {
+			return errors.Wrap(err, "WAL truncation in Compact")
```
bwplotka, line 832: Let's add different wrap here vs in defer above (:

tsdb/db.go
```diff
@@ -783,6 +783,17 @@ func (db *DB) Compact() (err error) {
 			db.metrics.compactionsFailed.Inc()
 		}
 	}()
+
+	walTruncationTime := int64(math.MinInt64)
+	walTruncationAttempted, headCompacted := false, false
+	defer func() {
+		if headCompacted && !walTruncationAttempted {
```
bwplotka, line 790: Not a fan of this. This logic with truncating something in memory then here, then also when head is compacted is kind of hard to follow. Let me look if I can suggest something better quickly.

bwplotka (commented):
Sorry for lag. Generally makes sense, but we might need slight refactor - the current proposal has many little flags and tweaks and might hard to follow. Ideally head totally mask WAL - it's optional part of head, so why DB has to manage it?

bwplotka:
Created simplified version: https://github.com/prometheus/prometheus/pull/8067 on top of your PR. 

Diff vs this PR: https://github.com/codesome/prometheus/compare/truncate-once...prometheus:checkpoint-at-the-end

codesome:
Closing this in favor of #8067, some minor thing to fix in that and it's LGTM for me :P

PR #7581, opened 2020-07-15
Log when starting to create a checkpoint

codesome:


tsdb/wal/checkpoint.go
```diff
@@ -87,10 +89,12 @@ const checkpointPrefix = "checkpoint."
 // segmented format as the original WAL itself.
 // This makes it easy to read it through the WAL package and concatenate
 // it with the original WAL.
-func Checkpoint(w *WAL, from, to int, keep func(id uint64) bool, mint int64) (*CheckpointStats, error) {
+func Checkpoint(logger log.Logger, w *WAL, from, to int, keep func(id uint64) bool, mint int64) (*CheckpointStats, error) {
 	stats := &CheckpointStats{}
 	var sgmReader io.ReadCloser
 
+	level.Info(logger).Log("msg", "creating checkpoint", "from_segment", from, "to_segment", to, "mint", mint)
```
roidelapluie, line 96:
```suggestion
	level.Info(logger).Log("msg", "Creating checkpoint", "from_segment", from, "to_segment", to, "mint", mint)
```

Is it info-worthy?

codesome: Yes. Because, users won't have debug mode enabled in general and in case of some issues in compaction/checkpointing/WAL truncation etc, it is helpful to know upto what point checkpoint was attempted (especially useful when checkpoint is ongoing or it crashed in between).

pracucci:
Thanks! 🙏

PR #7589, opened 2020-07-16
Replay m-map chunks irrespective of WAL

codesome:
We should always replay m-map chunks as that is something unavoidable, unlike WAL.

/cc @beorn7, would be nice to get it in 2.20, its a tiny change

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -638,10 +638,6 @@ func (h *Head) Init(minValidTime int64) error {
 	defer h.postings.EnsureOrder()
 	defer h.gc() // After loading the wal remove the obsolete data from the head.
 
-	if h.wal == nil {
-		return nil
-	}
-
 	level.Info(h.logger).Log("msg", "Replaying WAL and on-disk memory mappable chunks if any, this may take a while")
 	start := time.Now()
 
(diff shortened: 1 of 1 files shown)


beorn7:
Will merge on green.

beorn7:
Note: The problem solved here can only be exposed by using TSDB as a library. In the way it is used within Prometheus, `h.wal` can ever be `nil`.

roidelapluie:
Yes, BUT the extra log line is really great.

roidelapluie:
Should L710 be adapted to say: `WAL and memory chunks replay....` ? or should we reset start after playing the mmap chunks?

codesome:
I added the extra log line as that marks the end of the replay, so thought of keeping a single log at the end.
We could log time taken by mmap, checkpoint and WAL replay in the last log line.

codesome:
I modified the last log line to be `WAL and on-disk memory mappable chunks replay completed`

roidelapluie:
What I do not like is that I have a lot of time before the WAL replay starts - or at least that it what it looks like. Would love a log line after the in memory replay stops.

roidelapluie:
Especially since WAL replay is quite verbose, while mmap chunk replay is completely silent.

codesome:
I have added more logs. Also showing the time taken to replay each component (mmap chunks, checkpoint, and the WAL), which is useful.

PR #7592, opened 2020-07-16
rules: rollback recording rules in case of tsdb error

roidelapluie:
Something that I have seen while reading the code.

I will think about a way to test this.

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -589,8 +589,18 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
 			)
 
 			app := g.opts.Appendable.Appender()
+			var appErr error
 			seriesReturned := make(map[string]labels.Labels, len(g.seriesInPreviousEval[i]))
 			defer func() {
+				if appErr != nil {
+					app.Rollback()
+					app = g.opts.Appendable.Appender()
+					if err := g.addStalenessMarkers(app, ts, nil, i); err != nil {
+						// Make sure that if commit succeeds, next run won't re-create
+						// the staleness markers.
+						seriesReturned = nil
+					}
+				}
 				if err := app.Commit(); err != nil {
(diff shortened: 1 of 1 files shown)


rules/manager.go
```diff
@@ -589,8 +589,18 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
 			)
 
 			app := g.opts.Appendable.Appender()
+			var appErr error
 			seriesReturned := make(map[string]labels.Labels, len(g.seriesInPreviousEval[i]))
 			defer func() {
+				if appErr != nil {
+					app.Rollback()
```
codesome, line 596: We are also calling Commit() below. We should do only 1 of commit or rollback (especially after https://github.com/prometheus/prometheus/pull/7593). And I feel we should be committing the samples which did not give any error instead of rolling back everything.

roidelapluie: You are right, we should not commit.

However, I think we should roll back everything, just like we do for a failed scrape.

roidelapluie: Sorry: the commit was right: If we roll back, we commit staleness markers. We take a new appender here.

This is similar to what we do in #7593.

rules/manager.go
```diff
@@ -589,8 +589,18 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
...
+					app = g.opts.Appendable.Appender()
```
roidelapluie, line 597: @codesome hence we take a new appender

codesome: Ah right, didn't notice this reassignment

bwplotka:
Hello from the Bug Scrub!

Sounds like it's open for a while and we didn't have time to complete this (undraft). Let's close for now any reopen later if needed. Thanks!

PR #7593, opened 2020-07-16
TSDB: Error when we commit/rollback twice

roidelapluie:
Gathering feedback from TSDB people about this one. It looks like you can Commit or Rollback multiple times the same appender, which at least breaks isolation.

### DIFF
--- a/tsdb/db_test.go
+++ b/tsdb/db_test.go
@@ -521,7 +521,6 @@ func TestDB_Snapshot(t *testing.T) {
 		testutil.Ok(t, err)
 	}
 	testutil.Ok(t, app.Commit())
-	testutil.Ok(t, app.Rollback())
 
 	// create snapshot
 	snap, err := ioutil.TempDir("", "snap")
@@ -571,7 +570,6 @@ func TestDB_Snapshot_ChunksOutsideOfCompactedRange(t *testing.T) {
 		testutil.Ok(t, err)
 	}
 	testutil.Ok(t, app.Commit())
-	testutil.Ok(t, app.Rollback())
 
 	snap, err := ioutil.TempDir("", "snap")
 	testutil.Ok(t, err)
@@ -939,10 +937,14 @@ func TestWALSegmentSizeOptions(t *testing.T) {
 			opts.WALSegmentSize = segmentSize
 			db := openTestDB(t, opts, nil)
 
-			app := db.Appender()
 			for i := int64(0); i < 155; i++ {
-				_, err := app.Add(labels.Labels{labels.Label{Name: "wal", Value: "size"}}, i, rand.Float64())
+				app := db.Appender()
+				ref, err := app.Add(labels.Labels{labels.Label{Name: "wal" + fmt.Sprintf("%d", i), Value: "size"}}, i, rand.Float64())
 				testutil.Ok(t, err)
+				for j := int64(1); j <= 78; j++ {
+					err := app.AddFast(ref, i+j, rand.Float64())
+					testutil.Ok(t, err)
+				}
 				testutil.Ok(t, app.Commit())
 			}
 
@@ -2325,6 +2327,7 @@ func TestBlockRanges(t *testing.T) {
 
 	// Test that wal records are skipped when an existing block covers the same time ranges
 	// and compaction doesn't create an overlapping block.
+	app = db.Appender()
 	db.DisableCompactions()
 	_, err = app.Add(lbl, secondBlockMaxt+1, rand.Float64())
 	testutil.Ok(t, err)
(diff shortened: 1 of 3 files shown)


tsdb/head.go
```diff
@@ -1231,7 +1243,15 @@ func (a *headAppender) Commit() error {
...
 }
 
-func (a *headAppender) Rollback() error {
+func (a *headAppender) Rollback() (err error) {
+	if a.closed {
+		return ErrAppenderClosed
+	}
+	defer func() {
+		if err == nil {
+			a.closed = true
```
beorn7, line 1252: I think we should set this even if `err != nil`. We also do all the other deferred stuff below without condition on the error. Calling `Rollback()` twice would probably wreak havoc in any case.

beorn7:
Now the interesting parts is to study the test failures and see if that works at all.

roidelapluie:
Part of them (scrape package) were addressed in #7594

roidelapluie:
I have fixed some test but:

```
=== RUN   TestWALSegmentSizeOptions/WALSegmentSize_65536_test                                                                              
    TestWALSegmentSizeOptions: db_test.go:865: current WALSegmentSize should result in more than a single WAL file.                        
    TestWALSegmentSizeOptions/WALSegmentSize_65536_test: testing.go:916: test executed panic(nil) or runtime.Goexit: subtest may have calle
d FailNow on a parent test                                                                                                                 
--- FAIL: TestWALSegmentSizeOptions (0.01s)                                                                                                
    --- PASS: TestWALSegmentSizeOptions/WALSegmentSize_0_test (0.00s)                                                                      
    --- FAIL: TestWALSegmentSizeOptions/WALSegmentSize_65536_test (0.00s)
```

roidelapluie:
Commit calls a.log() which makes the WAL grow will all the series.

I guess that the first call of the function did log() 1 series, the second 2, etc.. so 23871 series?

roidelapluie:
Additional note:

The documentation is already saying that
- we should not reuse Commited and Rolled back appenders.
- Appender operations are not go-routing safe (therefore we do not need to protect `closed()` )

https://github.com/prometheus/prometheus/blob/b30654211cf99d244775c25f353d4becdb1aec99/storage/interface.go#L130-L157

roidelapluie:
This is now ready for review.

Also cc other tsdb users like @pracucci @bwplotka

tsdb/db_test.go
```diff
@@ -517,7 +516,6 @@ func TestDB_Snapshot_ChunksOutsideOfCompactedRange(t *testing.T) {
 		testutil.Ok(t, err)
 	}
 	testutil.Ok(t, app.Commit())
-	testutil.Ok(t, app.Rollback())
```
bwplotka, line 520: Can you leave those with `NotOk` then? (:

tsdb/db_test.go
```diff
@@ -889,10 +887,14 @@ func TestWALSegmentSizeOptions(t *testing.T) {
 			db, closeFn := openTestDB(t, opts, nil)
 			defer closeFn()
 
-			app := db.Appender()
 			for i := int64(0); i < 155; i++ {
-				_, err := app.Add(labels.Labels{labels.Label{Name: "wal", Value: "size"}}, i, rand.Float64())
+				app := db.Appender()
+				ref, err := app.Add(labels.Labels{labels.Label{Name: "wal" + fmt.Sprintf("%d", i), Value: "size"}}, i, rand.Float64())
```
bwplotka, line 892: Is this related to this PR? :thinking:

roidelapluie: Yes, as explained in or comments, when we reused a committed appender we accidentally logged all already committed samples again and again. With the new PR we need to append more samples to log() enough and get to the same wal size as before.

tsdb/db_test.go
roidelapluie: That seems unrelevant here but I can add a test for this.

roidelapluie:
I have added a small test

roidelapluie:
I have rebased on top of master.

tsdb/head_test.go
```diff
@@ -1852,3 +1852,39 @@ func TestHeadLabelNamesValuesWithMinMaxRange(t *testing.T) {
...
+	testutil.NotOk(t, app.Commit())
+	testutil.NotOk(t, app.Rollback())
+
+	app = head.Appender()
+	_, err = app.Add(labels.Labels{{Name: "test", Value: "test"}}, 1, 0)
+	testutil.Ok(t, err)
+	testutil.Ok(t, app.Rollback())
+	testutil.NotOk(t, app.Rollback())
+	testutil.NotOk(t, app.Commit())
+	app = head.Appender()
```
codesome, line 1875:
```
tsdb/head_test.go:1875:2: SA4006: this value of `app` is never used (staticcheck)
 
	app = head.Appender()
```

PR #7594, opened 2020-07-16
scrape tests: Make appenders more realistic

roidelapluie:
With this, the storage tests inside the scrape package are more
realistic.

Discovered with #7593, but fixed independently as #7593 will probably
take some time.


brian-brazil:
:+1:

PR #7596, opened 2020-07-17
Add unit test case for matcher.go

fenggw-fnst:
Signed-off-by: Guangwen Feng <fenggw-fnst@cn.fujitsu.com>

### DIFF
--- a/pkg/labels/matcher_test.go
+++ b/pkg/labels/matcher_test.go
@@ -117,3 +117,32 @@ func TestInverse(t *testing.T) {
 		testutil.Equals(t, test.expected.Type, result.Type)
 	}
 }
+
+func TestMatchType_String(t *testing.T) {
+	tests := []struct {
+		matchType MatchType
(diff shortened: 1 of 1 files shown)


brian-brazil:
Do we not already have tests that'd cover this that do general PromQL rendering?

fenggw-fnst:
Thanks for the review!

Yes, but I am afraid that they don't cover L34 ~ L36.
https://github.com/prometheus/prometheus/blob/22aa21e5089d0bbba31756a830386da52158c548/pkg/labels/matcher.go#L34
If I intentionally mess up the string value in L34 ~ L36, no test case will report failure. (Messing up L33 will cause test failure.)

brian-brazil:
Let's add a test for that at a higher level then, as that'll catch more issues.

fenggw-fnst:
OK, understood.
I'll close this one and add a test as you said. Thanks!

PR #7599, opened 2020-07-17
rules: fix bug for unknown alert state.

johncming:
It's a bug when the state is unknown state(fall into a loop).


brian-brazil:
Thanks!

PR #7604, opened 2020-07-17
Docker swarm: enable unix socket

### LINKED ISSUE #7603: Support docker swarm over Unix socket
Just tested this with the current RC:

I would suggest that the "host" parameter also supports unix-sockets. Exposing the docker socket over TCP is not something you would usually do in a production environment. Granting other containers access to the docker management port is commonly done by mounting /var/run/docker.sock read-only inside the container.

_Originally posted by @darkl0rd in https://github.com/prometheus/prometheus/pull/7420#issuecomment-660109510_

roidelapluie:
Fixes #7603

### DIFF
--- a/discovery/dockerswarm/dockerswarm.go
+++ b/discovery/dockerswarm/dockerswarm.go
@@ -108,8 +108,8 @@ func NewDiscovery(conf *SDConfig, logger log.Logger) (*Discovery, error) {
 	}
 
 	// There are other protocols that HTTP supported by the Docker daemon, like
-	// unix, which do not require an http client. Passing an http client to the
-	// docker client makes those non-http requests fail.
+	// unix, which are not supported by the HTTP client. Passing HTTP client
+	// options to the Docker client makes those non-HTTP requests fail.
 	if conf.url.Scheme == "http" || conf.url.Scheme == "https" {
 		rt, err := config_util.NewRoundTripperFromConfig(conf.HTTPClientConfig, "dockerswarm_sd", false)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


discovery/dockerswarm/dockerswarm.go
```diff
@@ -106,15 +102,29 @@ func NewDiscovery(conf *SDConfig, logger log.Logger) (*Discovery, error) {
...
-		client.WithHTTPClient(&http.Client{
-			Transport: rt,
-			Timeout:   time.Duration(conf.RefreshInterval),
-		}),
-		client.WithScheme(conf.url.Scheme),
 		client.WithAPIVersionNegotiation(),
-	)
+	}
+
+	// There are other protocols that HTTP supported by the Docker daemon, like
```
brian-brazil, line 110: that->than

brian-brazil:
This is not a bugfix, so I don't think it should be going into the release. @beorn7?

roidelapluie:
Wheter it is a bug or not can be discussed. However it was totally unintentional to prevent this, and I would rather enable it in the release if possible. But you are right that it is @beorn7 's call.

beorn7:
As we are merely in the RC phase, and this is a fairly limited and un-invasive change, I'd also say it should still go in.

If need be, we could classify it as a bugfix because the existing code "doesn't work as intended". However, I don't think we need to have that discussion at this stage.

roidelapluie:
@brian-brazil do you :+1: this one? I would love to see this in the release as it would probably enable more people to play with the SD right away.

brian-brazil:
:+1: 

> If need be, we could classify it as a bugfix because the existing code "doesn't work as intended".

I'd prefer we didn't pervert the definition of bugfix like that, and in this case it's your call anyway as the release shepard as to whether to include this in the .0.

Issue #7577, opened 2020-07-14
React UI: Alert Rule's "for" field not shown


s3lph:
**What did you do?**

Switch to the new React UI, then navigate to the "Alerts" tab or "Status/Rules" tab.

**What did you expect to see?**

I expected the alert rules to show all important fields, including the field `for`, e.g.:
```yaml
alert: SomethingWrong
expr: up < 1
for: 5m
labels:
  severity: critical
```

**What did you see instead? Under which circumstances?**

The field `for` is not shown, neither in the "Alerts" tab nor in the "Status/Rules" tab, e.g.
```yaml
alert: SomethingWrong
expr: up < 1
labels:
  severity: critical
```
The field is shown when using the classic UI.

**Environment**

* System information:

```
Linux 5.4.46-1-lts x86_64
```

* Prometheus version:

```
prometheus, version 2.19.2 (branch: tarball, revision: 2.19.2)
  build user:       someone@builder
  build date:       20200627-14:50:14
  go version:       go1.14.4
```

roidelapluie:
#7605 and #7606 are being worked on.

roidelapluie:
Fixed in #7605 and #7606

PR #7606, opened 2020-07-17
React UI: Multiple improvements on /rules page

roidelapluie:
- Add alerts duration
- Add rules labels
- Make rules look like alerts
- Hide annotations and labels if empty
- Make the alert name link point to the ALERTS query
 
### DIFF
--- a/web/ui/react-app/src/pages/rules/RulesContent.tsx
+++ b/web/ui/react-app/src/pages/rules/RulesContent.tsx
@@ -88,29 +88,29 @@ export const RulesContent: FC<RouteComponentProps & RulesContentProps> = ({ resp
                           <GraphExpressionLink title="alert" expr={r.name} />
                           <GraphExpressionLink title="expr" expr={r.query} />
                           {r.duration > 0 && (
-                          <div>
-                            <strong>for:</strong> {formatRange(r.duration)}
-                          </div>
+                            <div>
+                              <strong>for:</strong> {formatRange(r.duration)}
+                            </div>
                           )}
                           {Object.entries(r.labels).length > 0 && (
-                          <div>
-                            <strong>labels:</strong>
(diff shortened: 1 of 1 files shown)


roidelapluie:
See #7577

roidelapluie:
the react deployment does not work. I willtry to see why.

web/ui/react-app/src/pages/rules/RulesContent.tsx
```diff
@@ -95,7 +95,7 @@ export const RulesContent: FC<RouteComponentProps & RulesContentProps> = ({ resp
                             <strong>for:</strong> {formatRange(r.duration)}
                           </div>
                         )}
-                        {Object.entries(r.labels).length > 0 && (
+                        {r.labels && Object.entries(r.labels).length > 0 && (
```
juliusv, line 98: Same comment about `Object.entries` -> `Object.keys` here as in the other PR.

web/ui/react-app/src/pages/rules/RulesContent.tsx
```diff
@@ -95,7 +95,7 @@ export const RulesContent: FC<RouteComponentProps & RulesContentProps> = ({ resp
...
                           <div>
                             <strong>labels:</strong>
                             {Object.entries(r.labels).map(([key, value]) => (
```
juliusv, line 101: This is a pre-existing condition, but in case you sort out the quoting/escaping in the other PR, it would be good to do the same here with the label values.

juliusv:
:+1:

web/ui/react-app/src/pages/rules/RulesContent.tsx
roidelapluie: I will do both at the same time

Issue #7384, opened 2020-06-11
Log scraping errors at the info level with rate limiting


rfratto:
## Proposal 
Users that are running multiple Prometheus servers where more than one target is failing to scrape may find it cumbersome to navigate to each Prometheus UI to retrieve the last logged scrape error for each failing target. A centralized logging solution makes it easier to collect and act on the logs from each running server, but as the scraping errors are hidden behind the debug level, the debug level log verbosity might produce too high of a volume for some self-hosted logging platforms.

To allow users to collect logs on scraping errors without needing to accept a massive increase of log volume, I suggest letting scrape errors be logged at the info level. To prevent the volume of scrape errors logged massively increasing during a network outage, logging this message should be rate limited, where the rate limit for this message is shared between all targets. 

/cc @RichiH

brian-brazil:
This has been discussed before, scrape failures are not interesting at a Prometheus application level as scrape failures are normal. You can access these as debug logs, and from the UI.

Dupe of #2820.

rfratto:
The UI isn't particularly useful if you have a target that fails infrequently, as the UI only shows the error (if any) from the most recent scrape. 

I recognize scrape failures are normal but there's no other way for users to discover why Prometheus is unable to reach a target, and there should be some other mechanism in place to retrieve historical failures that doesn't require debug-level logging for the entire process.

brian-brazil:
Do you have a proposal along those lines for intermittent failures that doesn't involve logs?

beorn7:
Precisely because scrape failures are normal, we need a way to routinely track the reasons that does not require to set the whole server into debug-logging mode (because that's something you don't want to do routinely). Inspecting the UI at the right time doesn't really scale.

beorn7:
Along similar line is the problem of out-of-order and duplicate timestamps. It currently involves restarting the server with debug-level logging and wait for the incident to happen again. I'd also like to see _some_ way of logging those routinely in a way that can be tracked by a log processing system. Perhaps we can find one solution for both problems.

Info level logging with some way of dedup'ing / rate limiting might be an easy way out.

brian-brazil:
I don't see rate limited logs as a solution here, as in my mind application logs are somewhere where more than a log line per minute is spammy. To be useful here you'd need a fair bit more than that, which would detract from their role as application logs.

[...]

PR #7611, opened 2020-07-18
Allow changing log level at runtime

roidelapluie:
Helps (but does not solve) #7384 

Needs:
- Validation of the approach
- Tests
- PR to commons

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -265,6 +265,18 @@ func main() {
 	}
 
 	logger := promlog.New(&cfg.promlogConfig)
+	originalLevel := cfg.promlogConfig.Level
+
+	// Throw error for invalid config before starting other components.
+	var initCfg *config.Config
+	if initCfg, err = config.LoadFile(cfg.configFile); err != nil {
+		level.Error(logger).Log("msg", fmt.Sprintf("Error loading config (--config.file=%s)", cfg.configFile), "err", err)
+		os.Exit(2)
+	}
+
+	if initCfg.GlobalConfig.LogLevel != nil {
+		logger.SetLevel(initCfg.GlobalConfig.LogLevel)
+	}
 
 	cfg.web.ExternalURL, err = computeExternalURL(cfg.prometheusURL, cfg.web.ListenAddress)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


roidelapluie:
cc @beorn7 @SuperQ @brian-brazil

roidelapluie:
cc @rfratto to see if that would be helpful for his use case.

brian-brazil:
I'm against anything that involves having a setting coming from more than one place, as that's liable to cause confusion. I'm also wary of ending up with 10s of global options that very few users are likely to care about, and particular if they have no semantic effect such as this one.

What was agreed in #7384 was a way to view the last scrape error in the UI, and this PR does not help with the original claimed problem of #7384 of a logging system that can't handle the load of debug logs. If the level of debug logs are somehow a problem, they're a problem no matter how they're enabled.

roidelapluie:
On the other hand, @SuperQ mentionened in the community meeting that he'd like more settings to be tunable in the configuration. Additionally, this setting would be useful in other situation, like debugging recording rules.

roidelapluie:
I also note that the original reporter said that UI-only options are not enough to help them: https://github.com/prometheus/prometheus/issues/7384#issuecomment-648364335

brian-brazil:
> Additionally, this setting would be useful in other situation, like debugging recording rules.

How would that be relevant? There's only 3 debug logs in the recording rules, only 2 of those are at runtime and could be handled by a similar approach as for scrape failures.

brian-brazil:
> I also note that the original reporter said that UI-only options are not enough to help them: #7384 (comment)

That's what they said, but their claim seems to be that all the debug logs are so voluminous that enabling debug logging isn't viable for their logging solution can't handle it. I'm dubious of this claim, but even if it's true changing how debug logs are enabled doesn't help them as they still can't enable debug logs.

This PR does not help with #7384 as stated.

rfratto:
> That's what they said, but their claim seems to be that all the debug logs are so voluminous that enabling debug logging isn't viable for their logging solution can't handle it. I'm dubious of this claim, but even if it's true changing how debug logs are enabled doesn't help them as they still can't enable debug logs.
>
> This PR does not help with #7384 as stated.

To be fully transparent, the request is originating from a user whose logging system struggles to handle a high volume and they have a significant number of scrape targets. Debug level logging at their scrape load to the centralized logging system would cause a lot of issues for them. I didn't challenge the claim, but I find it to be reasonable (but this also isn't the forum to discuss that).

As Brian mentioned, while I think this is nice, this doesn't alleviate the problems from #7384; between dynamically changing log level and including the last received error in the UI, I think the latter is preferable.

roidelapluie:
thanks!

brian-brazil:
> I didn't challenge the claim, but I find it to be reasonable (but this also isn't the forum to discuss that).

Doing some quick estimation for a worst case, say 15K targets, 15s interval, 100B per log entry, and all scrapes are failing. That's 100kB/s, which is not a high volume. Even if that's off by an order of magnitude, that's still well within what a logging system should be able to handle - particularly as the number of Prometheus servers will tend to be only a small fraction of what is producing logs in such  system. I don't think we should be making design decisions based on the limits of suboptimally performing event-logging systems.

> including the last received error in the UI, I think the latter is preferable.

Indeed. That'd also presumably make it available via the HTTP APIs, and thus someone could build the centralised dashboard you had also mentioned on top of that.

PR #7613, opened 2020-07-18
Add {min,max}_over_time_last_timestamp

### LINKED ISSUE #5003: Actual timestamp() for max_over_time (and other agg functions)
## Proposal
Knowing when a min or max happened in defined window of time is useful. Imagine monitoring a gauge like metric over a week, you want to know the max value in that period (easy) and the time it happened (currently impossible using PromQL)

Something like:
```
max_over_time(shop_online_users{dimension="frontpage"}[1w])
```
Would give the value, but we don't know when the `max` point happened in `1w`.

**Suggestion**
If only the `timestamp()` (or a new function) would return the actual timestamp of evaluated point in time, the problem would be solved:
```
timestamp(max_over_time(shop_concurrent_users{dimension="frontpage"}[1w]))
```
It returns `now` or actually same value as `time()` at the moment.

I can see the `timestamp()` can return the actual timestamp of a metric (only last value) when used without aggregate functions, as discussed in #1557

roidelapluie:
Fixes #5003 

It does not really make sense of other aggr over time, which actually alter data.

### DIFF
--- a/promql/functions.go
+++ b/promql/functions.go
@@ -384,7 +384,6 @@ func funcCountOverTime(vals []parser.Value, args parser.Expressions, enh *EvalNo
 	})
 }
 
-// === floor(Vector parser.ValueTypeVector) Vector ===
 // === max_over_time(Matrix parser.ValueTypeMatrix) Vector ===
 func funcMaxOverTime(vals []parser.Value, args parser.Expressions, enh *EvalNodeHelper) Vector {
 	return aggrOverTime(vals, enh, func(values []Point) float64 {
@@ -398,6 +397,21 @@ func funcMaxOverTime(vals []parser.Value, args parser.Expressions, enh *EvalNode
 	})
 }
 
+// === max_over_time_last_timestamp(Matrix parser.ValueTypeMatrix) Vector ===
+func funcMaxOverTimeLastTimestamp(vals []parser.Value, args parser.Expressions, enh *EvalNodeHelper) Vector {
+	return aggrOverTime(vals, enh, func(values []Point) float64 {
+		max := values[0].V
+		maxTime := values[0].T
+		for _, v := range values {
+			if v.V >= max || math.IsNaN(max) {
+				max = v.V
+				maxTime = v.T
+			}
+		}
+		return float64(maxTime) / 1000
+	})
+}
+
 // === min_over_time(Matrix parser.ValueTypeMatrix) Vector ===
 func funcMinOverTime(vals []parser.Value, args parser.Expressions, enh *EvalNodeHelper) Vector {
 	return aggrOverTime(vals, enh, func(values []Point) float64 {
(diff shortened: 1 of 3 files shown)


brian-brazil:
I'm still not quite convinced by the use case for this. Why would this be a useful thing to do math on?

roidelapluie:
I have this on a dashboard (it used to work with more or less complex recording rules but was lost on a prometheus restart):

maxc

brian-brazil:
Is useful for operational reasons, or more a vanity metric? Do you need it often enough that you can't eyeball it on a graph, which will give you the trend/pattern too which is likely key to understanding the relevance of that raw number?

roidelapluie:
This is more to know when the last press conference happened so we know the timeframe we need to open the graphs at.

brian-brazil:
That sounds more like a matter for a calendar than a metrics-based monitoring system.

roidelapluie:
Based on that, I propose that on the next bug scrub we discuss this and close #5003.

roidelapluie:
I think we could think this again since we will have @ queries, I think that there could be some nice uses cases here.

brian-brazil:
How would that interact? Keep in mind that the time in an @ query cannot come from a query, as that isn't compatible with how isolation works - we need to know the time range of a query before touching any tsdb data.

roidelapluie:
http_request_rate and abs(timestamp(http_request_rate) - max_timestamp_over_time(http_request_rate[1h]) @ end < 100

Would show mtrics around the peaks of every timeserie

ctml91:
Much needed. Either an optional flag to use the metric timestamp as opposed to evaluated timestamp, or a separate agg function to get the time. Right now I have to parse the data all myself, which defeats the purpose of using min/max_over_time as I can get both in one shot if I'm going to need to search the data anyways.

roidelapluie:
> Much needed. Either an optional flag to use the metric timestamp as opposed to evaluated timestamp, or a separate agg function to get the time. Right now I have to parse the data all myself, which defeats the purpose of using min/max_over_time as I can get both in one shot if I'm going to need to search the data anyways.

What is your usecase?

ctml91:
Identifying peak periods of cpu/memory utilization (using prometheus on kubernetes). Graphs provide an approximate view, but aren't as precise as having the raw data. This is especially useful if building tabled utilization reports across a large number of applications. _Example_: peak CPU utilization of App 1 was 84%, what time was that? to find out, you need to process that externally, but at the same time defeats the purpose of using the function as you'll find both processing it yourself. 

This gives me an idea for another aggregate function, topk_over_time and bottomk_over_time. I think by default this would retain the timestamp of each value.

brian-brazil:
The output of any function needs to be the evaluation time, things would break pretty badly otherwise - plus those are already the guaranteed semantics.

roidelapluie:
Closing this because the implementation is not correct

Issue #11, opened 2013-01-04
Persist Retroactive Rule Reevaluations


juliusv:
We can already reevaluate queries on old data, but we should be able to persist that for a certain window from `[Oldest, Now)`.

brian-brazil:
My thoughts on this are to permit backfilling via the API, and then we can have a separate tool that does this.

Bplotka:
Hi, I would like to start a new discussion on this. 
Basically, there are some clear benefits of having some back-in-time rule evaluation option for Prometheus.

Currently, there are certain UX issues while using **rules**. Let's have some basic example from a user perspective:
- I created some cool, new metric.
- I created some fancy dashboard.
- Everybody loves my dashboard and it's quite popular, so I want to create a rule for the query used there to take out some load from Prometheus.
- I created a rule.

Problems:
1) **How to test my rule**? I need to wait some time to have it filled to check if I got it right for a longer period (that could include interesting cases). That requires lot's of context switches in my work and frustration (multiple frustrations by 100x when one is a newbie in PromQL)
2) **I cannot immediately put my rule on the dashboard because it's not yet filled**

I totally agree that this is not critical and quite hard to achieve when we are talking about persisting new time series back in time when that time range is already packed in blocks and maybe even compacted or backed up/downsampled in case of [thanos](https://github.com/improbable-eng/thanos).

But what about some simple solution (A), for example:
- User created new rule
- If flag XXX is set (or certain RPC is invoked on API), reevaluate the rule against all samples you have in memory (WAL)

The only unknown here is `remote write`, but when we make this feature optional - you can control if you want to have back-in-time rule evaluation or not.

Or maybe another option (B):
- User created new rule
- Backfill the rule in ephemeral space only (mem only) up to `flag-rule-reeval-time` or API RPC argument to make testing and using this rule on dashboards easier. "Past" samples will be ephemeral. Only new samples will be written normally.

Any thoughts?

CC @juliusv @brian-brazil @fabxc  @devnev

brian-brazil:
The requirement here is basically for bulk import, as I don't think this functionality belongs in Prometheus itself (though may be tied together via promtool).

Bplotka:
> bulk import
Sorry, what do you mean by that?

brian-brazil:
#535

amadeuszkryze:
Hi, I'd like to ask if any progress has been made regarding generation of metrics in the past based on recording rules, I've described use case in #6440.

[...]

PR #7616, opened 2020-07-20
WIP - init prototype for issue 11

JessicaGreben:
This is a very rough WIP for issue #11, mainly just a brainstorming first draft.

Next steps are to evaluate this PR #7586 and see how it fits in with this PR.

### DIFF
--- a/cmd/promtool/main.go
+++ b/cmd/promtool/main.go
@@ -40,6 +40,7 @@ import (
 	kingpin "gopkg.in/alecthomas/kingpin.v2"
 
 	"github.com/prometheus/prometheus/config"
+	"github.com/prometheus/prometheus/importers"
 	"github.com/prometheus/prometheus/pkg/rulefmt"
 )
 
@@ -103,6 +104,20 @@ func main() {
 		"The unit test file.",
 	).Required().ExistingFiles()
 
+	backfillCmd := app.Command("backfill", "Backfill Prometheus data.")
+	backfillRuleCmd := backfillCmd.Command("rules", "Backfill Prometheus data for new rules.")
+	backfillRuleURL := backfillRuleCmd.Flag("url", "Prometheus API url.").Required().String()
+	backfillRuleEvalInterval := backfillRuleCmd.Flag("evaluation_interval", "How frequently to evaluate rules when backfilling.").
+		Default("1m").Duration()
+	backfillRuleStart := backfillRuleCmd.Flag("start", "If a start time is provided, the new rule backfilling will begin at this time. The default evaluates from the first know data. Start time should be RFC3339 or Unix timestamp.").
+		Default("").Duration()
+	backfillRuleEnd := backfillRuleCmd.Flag("end", "If an end time is provided, the new rule backfilling will end at this time. The default will backfill to the current time. End time should be RFC3339 or Unix timestamp.").
+		Default("").Duration()
+	backfillRuleFiles := backfillRuleCmd.Arg(
+		"rule-files",
+		"The file containing the new rule that needs to be backfilled.",
+	).Required().ExistingFiles()
+
 	parsedCmd := kingpin.MustParse(app.Parse(os.Args[1:]))
 
 	var p printer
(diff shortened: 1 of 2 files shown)


cmd/promtool/main.go
```diff
@@ -103,6 +104,20 @@ func main() {
...
 	).Required().ExistingFiles()
 
+	backfillCmd := app.Command("backfill", "Backfill Prometheus data.")
+	backfillRuleCmd := backfillCmd.Command("rules", "Backfill Prometheus data for new rules.")
+	backfillRuleURL := backfillRuleCmd.Flag("url", "Prometheus API url.").Required().String()
+	backfillRuleEvalInterval := backfillRuleCmd.Flag("evaluation_interval", "How frequently to evaluate rules when backfilling.").
+		Default("1m").Duration()
+	backfillRuleStart := backfillRuleCmd.Flag("start", "If a start time is provided, the new rule backfilling will begin at this time. The default evaluates from the first know data. Start time should be RFC3339 or Unix timestamp.").
+		Default("").Duration()
+	backfillRuleEnd := backfillRuleCmd.Flag("end", "If an end time is provided, the new rule backfilling will end at this time. The default will backfill to the current time. End time should be RFC3339 or Unix timestamp.").
```
brian-brazil, line 114: Current time is dangerous, we need to go back 1-3 hours ago to before the head.

cmd/promtool/main.go
```diff
@@ -103,6 +104,20 @@ func main() {
 		"The unit test file.",
 	).Required().ExistingFiles()
 
+	backfillCmd := app.Command("backfill", "Backfill Prometheus data.")
+	backfillRuleCmd := backfillCmd.Command("rules", "Backfill Prometheus data for new rules.")
+	backfillRuleURL := backfillRuleCmd.Flag("url", "Prometheus API url.").Required().String()
+	backfillRuleEvalInterval := backfillRuleCmd.Flag("evaluation_interval", "How frequently to evaluate rules when backfilling.").
+		Default("1m").Duration()
+	backfillRuleStart := backfillRuleCmd.Flag("start", "If a start time is provided, the new rule backfilling will begin at this time. The default evaluates from the first know data. Start time should be RFC3339 or Unix timestamp.").
```
brian-brazil, line 112: You can't know when the first known data is from query_range alone without searching, in practice this will have to be provided by the user.

know->known

cmd/promtool/main.go
```diff
@@ -696,3 +714,33 @@ func (j *jsonPrinter) printLabelValues(v model.LabelValues) {
...
+	cfg := importers.RuleConfig{
+		Start:        start.String(),
+		End:          end.String(),
+		EvalInterval: evalInterval,
+		URL:          url,
+	}
+	ruleImporter := importers.NewRuleImporter(cfg)
+	err := ruleImporter.Init()
+	if err != nil {
+		return 1
```
brian-brazil, line 730: Don't fail without printing an error to the user

importers/rules.go
```diff
@@ -0,0 +1,239 @@
...
+	if err != nil {
+		return err
+	}
+	url, err := url.Parse(importer.config.URL)
+	if err != nil {
+		return err
+	}
+
+	appender := importer.headBlock.Appender()
+	// todo: do 2 hr blocks here?
```
brian-brazil, line 199: Yes, you want to do this in no more than 2h chunks. Doing every step individually would be much slower than it needs to be.

importers/rules.go
```diff
@@ -0,0 +1,239 @@
...
+			Metric: labels,
+			Point:  promql.Point{T: int64(value.Timestamp), V: float64(value.Value)},
+		})
+	}
+	return result
+}
+
+// ImportRule imports the historical data for a single rule
+func (importer *RuleImporter) ImportRule(ctx context.Context, rule rules.Rule) error {
+	ts, err := parseTime(importer.config.Start)
```
brian-brazil, line 185: You're using the user-supplied timestamp for alignment, that'll cause problems as it is unlikely to be idempotent in practice. Either work from 0, or what Prometheus itself would use for the rule group.

importers/rules.go
bwplotka: This is where `blocks.Writer` from https://github.com/prometheus/prometheus/pull/7586 might be helpful (:

importers/rules.go
bwplotka: Why from 0 if user-specified e.g last week as the start date?

bwplotka: Ok, I got it now. Multiply steps  from 0 (:

Issue #7483, opened 2020-06-30
Compaction running out of memory


clambin:
**What did you do?**

Since upgrading to 2.19.2, I frequently see Prometheus running out of memory when compacting:

```
level=error ts=2020-06-29T21:00:03.803Z caller=db.go:675 component=tsdb msg="compaction failed" err="reload blocks: head truncate failed: truncate chunks.HeadReadWriter: mmap: cannot allocate memory"

panic: write /prometheus/chunks_head/000292: file already closed
```

However, when I check system resources, there is plenty of memory available.  E.g. in the case above, RSS was only 256MB (and has been higher than that without issue).

I suspect this has something to do with #6679 , which talks about this specific panic.

**What did you expect to see?**

Compaction should complete successfully.

**What did you see instead? Under which circumstances?**

Been tracking this issue for a couple of days now, but no discernible pattern:  compactions with more memory and more head chunks have been successful.

**Environment**

* System information:

	Linux 4.19.118-v7l+ armv7l
        running in docker (swarm) Docker version 19.03.12, build 48a6621 on raspbian buster.

* Prometheus version:

	prometheus, version 2.19.2 (branch: HEAD, revision: c448ada63d83002e9c1d2c9f84e09f55a61f0ff7)
	build user:       root@dd72efe1549d
	build date:       20200626-10:09:28
	go version:       go1.14.4

* Alertmanager version:

	N/A

* Prometheus configuration file:
```
services:
  prometheus:
    image: prom/prometheus:v2.19.2
    networks:
      - net
    volumes:
      - /mnt/media/prometheus_data:/prometheus
      - ./prometheus:/etc/prometheus
      - /etc/timezone:/etc/timezone
    ports:
      - 9090:9090
    command:
      - '--config.file=/etc/prometheus/prometheus.yml'
      - '--storage.tsdb.path=/prometheus'
      - '--storage.tsdb.retention.time=90d'
      - '--web.console.libraries=/etc/prometheus/console_libraries'
      - '--web.console.templates=/etc/prometheus/consoles'
      - '--web.enable-lifecycle'
```

(retention was 90d. I've since reduced it to 45d but problem remains the same)

* Alertmanager configuration file:

N/A

* Logs:

```
[...]

PR #7617, opened 2020-07-20
tsdb: use roaring bitmap as MemPostings implement

Sasasu:
**Warning: this patch has very very bad performance. blame RoaringBitmap's interface based API**

prometheus's tsdb uses u64 array as posting list implementation. 

Those arrays will use huge memory when loading chunks. the bench:

```
BenchmarkRoaringMemoryUsage/list-8     38 31332405 ns/op     240222274 B/op    22 allocs/op   = 229.093813 MiB/op
BenchmarkRoaringMemoryUsage/roaring-8  7  155508017 ns/op    7024749   B/op    3296 allocs/op = 6.6993227  MiB/op
```

use roaring bitmap as a postlist implement will save a lot of memory.

https://github.com/RoaringBitmap/roaring/issues/228 

will fix https://github.com/prometheus/prometheus/issues/7483 and https://github.com/prometheus/prometheus/issues/5547 and a lot

full bench
<details>

```
goos: linux
goarch: amd64
pkg: github.com/prometheus/prometheus/tsdb/index
BenchmarkIntersect
BenchmarkIntersect/LongPostings1
BenchmarkIntersect/LongPostings1-8         	 9544412	       119 ns/op	      96 B/op	       2 allocs/op
BenchmarkIntersect/LongPostings2
BenchmarkIntersect/LongPostings2-8         	10448314	       108 ns/op	      96 B/op	       2 allocs/op
BenchmarkIntersect/ManyPostings
BenchmarkIntersect/ManyPostings-8          	    1772	    624165 ns/op	      32 B/op	       1 allocs/op
BenchmarkRoaring
BenchmarkRoaring/LongPostings1_roaring
BenchmarkRoaring/LongPostings1_roaring-8   	   19692	     60626 ns/op	    2480 B/op	     113 allocs/op
BenchmarkRoaring/LongPostings2_roaring
BenchmarkRoaring/LongPostings2_roaring-8   	       3	 345627385 ns/op	   11200 B/op	     347 allocs/op
BenchmarkRoaring/ManyPostings_roaring
BenchmarkRoaring/ManyPostings_roaring-8    	       1	1335282102 ns/op	16000032 B/op	  400001 allocs/op
BenchmarkPostings_Stats
BenchmarkPostings_Stats-8                  	      27	  38904193 ns/op
BenchmarkRoaringMemoryUsage
BenchmarkRoaringMemoryUsage/list
BenchmarkRoaringMemoryUsage/list-8         	      28	  43142683 ns/op	240222252 B/op	      21 allocs/op
BenchmarkRoaringMemoryUsage/roaring
BenchmarkRoaringMemoryUsage/roaring-8      	       6	 195825095 ns/op	 7024736 B/op	    3296 allocs/op
BenchmarkPostingStatsMaxHep
BenchmarkPostingStatsMaxHep-8              	      10	 118487476 ns/op
```

</details>


roidelapluie:
Thank you for this. Tests are currently failing.

Do we also have existing benchmarks that would be improved by this?

roidelapluie:
Sorry, I am trying to launch a benchmark.

codesome:
Hold on, we need to discuss more about this before we go ahead. There could be better ones. There is also this work https://github.com/prometheus/prometheus/pull/5884/ which uses 64 bit postings on disk and uses prefix compression on postings as roaring bitmaps didn't work out that well in terms of performance. And for compressions like this it would be better to implement it inside TSDB instead of importing a package for it.

codesome:
Thanks for the PR btw! If any of these compressions work on par with current performance, it would be nice enhancement.

Sasasu:
There is a mistake in the comment.

`Postings.Seek()` return `true` when the posting still has value.

https://github.com/prometheus/prometheus/blob/fe8d412ce9b7313ea5c31686d25a648aba11fc21/tsdb/block_test.go#L359-L379

the `genSeries` generate test data in memory, I think half of the memory consumed in generate test data.

roaring bitmap only uses a little memory and 150% slower than array when iterating.

codesome:
@Sasasu would you have time to implement roaring bitmaps and prefix compression and benchmark both of them out? I am afraid the PR at the current state is not feasible.

Sasasu:
only roaring bitmap in memory, not interest in prefix compression on disk.

I prefer to split to block first then use delta encoding + varint encoding + and some other encoding method as the on-disk structure.

codesome:
I was talking about trying out roaring bitmaps and prefix compression for in-memory postings. Regarding on-disk, that discussion could continue in https://github.com/prometheus/prometheus/pull/5884 as there are other things to consider with the disk involved.

roidelapluie:
Hello @codesome , do you think we should close this in favor of #5884? Thanks!

Issue #7618, opened 2020-07-20
PromQL: unexpected currentSamples count for BinaryExpr


Betula-L:
**What did you do?**

I have an expression `myserver_response_duration_ms * 1000` which `count(myserver_response_duration_ms)` is not large than maxSample/2 , however prometheus return it is out of maxSample. 

I found the reason is that `tempNumSamples` should not add previous `currentSample` since it is outdated in memory.
https://github.com/prometheus/prometheus/blob/490f9c664e48fbd1509c42ae85fa566901b11a99/promql/engine.go#L917
**What did you expect to see?**

The currentSample of metric `go_gc_duration_seconds` with 5 sample is `5`
![image](https://user-images.githubusercontent.com/6059935/87936729-ed7d4780-cac5-11ea-94a5-e616676c1b2a.png)

Then the currentSample of metric `go_gc_duration_seconds * 2` with 5 sample is `11`.
```
originSample(5) + `2`(1) + `result(5)` = 11
```

**What did you see instead? **

Current  currentSample of metric `go_gc_duration_seconds * 2` with 5 sample is `17`.
```
out-of-dated evaluator(6) + originSample(5) + `2`(1) + `result(5)`
```

**What did you see instead?**

**Environment**

* Prometheus version:

prometheus, version 2.19.0 (branch: master, revision: f42ed03dc5be7290a913fe427cdf96f6d1bee792)
  build user:       root@nvm-ops-moni-dev-4
  build date:       20200720-12:24:13
  go version:       go1.14.3

* Logs:

This is my debug logs
```
level=debug ts=2020-07-20T12:01:02.515Z caller=engine.go:1279 component="query engine" msg="before binary expr current samples" value=0
level=debug ts=2020-07-20T12:01:02.515Z caller=engine.go:896 component="query engine" msg="before eval expr current samples start" value=0
level=debug ts=2020-07-20T12:01:02.515Z caller=engine.go:1330 component="query engine" msg="before vector seletor current samples" value=0
level=debug ts=2020-07-20T12:01:02.515Z caller=engine.go:1358 component="query engine" msg="after vector seletor current samples" value=5
level=debug ts=2020-07-20T12:01:02.515Z caller=engine.go:896 component="query engine" msg="before eval expr current samples start" value=5
level=debug ts=2020-07-20T12:01:02.516Z caller=engine.go:929 component="query engine" msg="before eval expr current samples loop" value=5
[...]

PR #7619, opened 2020-07-20
fix current sample counter

Betula-L:
Fix https://github.com/prometheus/prometheus/issues/7618

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -914,7 +914,7 @@ func (ev *evaluator) rangeEval(f func([]parser.Value, *EvalNodeHelper) (Vector, 
 	}
 	enh := &EvalNodeHelper{out: make(Vector, 0, biggestLen)}
 	seriess := make(map[uint64]Series, biggestLen) // Output series by series hash.
-	tempNumSamples := ev.currentSamples
+	tempNumSamples := 0
 	for ts := ev.startTimestamp; ts <= ev.endTimestamp; ts += ev.interval {
 		if err := contextDone(ev.ctx, "expression evaluation"); err != nil {
 			ev.error(err)


promql/engine.go
```diff
@@ -914,7 +914,7 @@ func (ev *evaluator) rangeEval(f func([]parser.Value, *EvalNodeHelper) (Vector,
 	}
 	enh := &EvalNodeHelper{out: make(Vector, 0, biggestLen)}
 	seriess := make(map[uint64]Series, biggestLen) // Output series by series hash.
-	tempNumSamples := ev.currentSamples
+	tempNumSamples := 0
```
brian-brazil, line 917: This doesn't look right, this is ignoring the existence of anything that is memory when this function is called.

Betula-L: Make sense. It's not right.

I went through code again and find that there are three step for func `rangeEval`

```
- Step 1: func evalExpr (e []Expr) => (m []Matrix)
- Step 2: func matrixColumn (m []Matrix) => (v []Vector)  // temporary step
- Step 3: func eval(v []Vector, nh NodeHelp) => (result []Vector)
```

Since `Step 2` is counted, currentSamples is weird but exact.

Betula-L: Maybe more efficient implements for `rangeEval` here, however, it's a minor optimization.

brian-brazil: What are you proposing?

Betula-L: Intuitively, the involved samples a n*1 metric vector multiply a constant should be `n+1+n`, while prometheus return `n+1n+1+n` due to implementation.

That is not easy to understand for fresher of prometheus.

brian-brazil: I'm not following, what does that have to do with this code?

Betula-L: I mean involved samples is not correct intuitively.

But this PR is not a proper fix. I will close it.

PR #7621, opened 2020-07-20
Downgrade grpc to v1.29.1

codesome:
grpc was upgraded to v1.30.0 during 2.20 release which has a breaking change of removal of `naming` package (which Cortex uses and also other dependency (etcd) which Cortex imports).

While we can use `replace` directive to use v1.29.1 under the hood, the indirect dependency `google.golang.org/grpc/examples v0.0.0-20200709232328-d8193ee9cc3e // indirect` that v1.30.0 brings to Prometheus does not allow using the replace directive.

@beorn7 and I tried many different ways to get around this, but nothing works. So this PR downgrades grpc version to back where it was without any effect to Prometheus but helps Corted import Prometheus without any problem.


roidelapluie:
Is there an issue upstream about this?

beorn7:
An additional complication is that grpc went from mono-module repo to multi-module repo with 1.30, which knocks out the option to use `replace` in go.mod.

Effectively, they performed two breaking changes without bumping up the major number: Removed a package, and changed from mono to multi module. Thangs @google for sticking to your own rules…

beorn7:
> Is there an issue upstream about this?

I'm afraid, grpc won't bump to v2 because of that.

This would be OK if everyone adjusted to cope with grpc 1.30. https://github.com/etcd-io/etcd/issues/12153 for the corresponding effort in etcd.

beorn7:
Nothing in Prometheus needed the grpc 1.30 upgrade. It was just a routine upgrade. So I think we are fine for now to do it that way. But that's the 2nd wrinkle in dependency management (the K8s go client with it's broken version tags is the first).

codesome:
Thanks!

brian-brazil:
One option if this becomes a problem would be to drop the v2 HTTP API now, rather than the current plan which is to wait for Prometheus 3.x. That's the only thing which uses GRPC afaik, and would not classify as a breaking change.

bwplotka:
I think it still will be. The only effort to make it to not update will be to put `explicit` `replica` with comment and why. And yes, we should do it (:

roidelapluie:
@bwplotka see #7624

PR #7622, opened 2020-07-20
Cut 2.20.0-rc.1

beorn7:
Note that all new lines in the CHANGELOG except the last will
disappear in the final release because they only deal with issues
introduced in 2.20.0-rc.0.

This commit also adds an explanation for the WAL compression becoming
default.

Furthermore, this commit reverts the upgrade to grpc 1.30.0. It was
just a routine upgrade, but because grpc 1.30.0 contains two very
Go-module unfriendly breaking changes that in combination wreak havoc
for anyone importing from the prometheus/prometheus repo while still
depending on 1.29.1, we decided to revert as Prometheus itself doesn't
need the upgrade right now.


beorn7:
Wait, I still need to add the PR numbers. Just a sec…

beorn7:
Since none of the changes is expected to have any impact on the performance, I'm not running another prombench on this.

PR #7628, opened 2020-07-21
promql: Removed global and add ability to have better interval for subqueries if not specified

bwplotka:
## Changes

* Refactored tests for better hints testing
* Added various TODO in places to enhance.
* Moved DefaultEvalInterval global to opts with func(rangeMillis int64) int64 function instead

Motivation: At Thanos we would love to have better control over the subqueries step/interval.
This is important to choose proper resolution. I think having proper step also does not harm for
Prometheus and remote read users. Especially on stateless querier we do not know evaluation interval
and in fact putting global can be wrong to assume for Prometheus even.

I think ideally we could try to have at least 3 samples within the range, the same
way Prometheus UI and Grafana assumes.

Anyway this interfaces allows to decide on promQL user basis.

Open question: Is taking parent interval a smart move?

Motivation for removing global: I spent 1h fighting with:

```
=== RUN   TestEvaluations
    TestEvaluations: promql_test.go:31: unexpected error: error evaluating query "absent_over_time(rate(nonexistant[5m])[5m:])" (line 687): unexpected error: runtime error: integer divide by zero
--- FAIL: TestEvaluations (0.32s)
FAIL
```

At the end I found that this fails on most of the versions including current master if you run this test alone. If run together with many
other tests it passes. This is due to `SetDefaultEvaluationInterval(1 * time.Minute)`
in test that is ran before TestEvaluations. Thanks to globals (:

Let's fix it by dropping this global.

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -335,8 +335,8 @@ func main() {
 		}
 	}
 
-	noStepSuqueryInterval := &safePromQLNoStepSubqueryInterval{}
-	noStepSuqueryInterval.Set(config.DefaultGlobalConfig.EvaluationInterval)
+	noStepSubqueryInterval := &safePromQLNoStepSubqueryInterval{}
+	noStepSubqueryInterval.Set(config.DefaultGlobalConfig.EvaluationInterval)
 
 	// Above level 6, the k8s client would log bearer tokens in clear-text.
 	klog.ClampLevel(6)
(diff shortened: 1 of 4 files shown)


cmd/prometheus/main.go
```diff
@@ -334,7 +335,8 @@ func main() {
 		}
 	}
 
-	promql.SetDefaultEvaluationInterval(time.Duration(config.DefaultGlobalConfig.EvaluationInterval))
+	noStepSuqueryInterval := &safePromQLNoStepSubqueryInterval{}
```
cstyan, line 338: You missed the b in this and the next variable name

promql/engine.go
```diff
@@ -657,13 +646,14 @@ func (ng *Engine) populateSeries(querier storage.Querier, s *parser.EvalStmt) {
...
 			}
 
 			// We need to make sure we select the timerange selected by the subquery.
-			// TODO(gouthamve): cumulativeSubqueryOffset gives the sum of range and the offset
-			// we can optimise it by separating out the range and offsets, and subtracting the offsets
-			// from end also.
+			// The cumulativeSubqueryOffset function gives the sum of range and the offset.
+			// TODO(gouthamve): Consider optimising it by separating out the range and offsets, and subtracting the offsets
+			// from end also. See: https://github.com/prometheus/prometheus/issues/7629.
+			// TODO(bwplotka): Add support for better hints when subquerying. See: https://github.com/prometheus/prometheus/issues/7630.
```
brian-brazil, line 656: This appears to be an unrelated change.

bwplotka: Yea, can separate those, thanks

PR #7633, opened 2020-07-21
DigitalOcean, Docker Swarm: properly load files

roidelapluie:
- Make sure that token files etc follow prometheus standard rules.


roidelapluie:
The private field conf.url that was added on the DockerSwarm service discovery does prevent coalesce of service discoveries, preventing the same SD to be reused multiple times.

brian-brazil:
:+1:

Issue #7576, opened 2020-07-14
Generalise an interface for retention in TSDB


codesome:
The idea is to make the retention decisions via an interface that also allows looking at each block. With this, the time based and size based retention logic would be a single interface for Prometheus and allow external users to inject different interfaces as retention filters.

This is particularly helpful for Cortex and Thanos who don't want some blocks to be deleted based on some conditions.

/cc @pracucci @pstibrany @bwplotka @brancz

bwplotka:
Either generalization or API for deleting blocks I guess.

pracucci:
> API for deleting blocks I guess

Interesting approach, but how would that work? If a block is not "ready to be deleted" (because not shipped to the long-term storage yet) how should it work? The API fails and TSDB continuously retry? I would love to hear more thoughts on this 🙏

codesome:
Even with an API, you would need some control over the retention policy to _not delete_ blocks.

brancz:
I tend to agree that an pluggable interface would be best here. Refactoring shouldn't be too invasive to make this possible I think.

PR #7638, opened 2020-07-22
Optional retention filter for DB

codesome:
Closes https://github.com/prometheus/prometheus/issues/7576

It is a very small refactoring which will help Cortex and Thanos.

/cc @bwplotka @brancz @pracucci

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -119,11 +119,13 @@ type Options struct {
 	// It is always a no-op in Prometheus and mainly meant for external users who import TSDB.
 	SeriesLifecycleCallback SeriesLifecycleCallback
 
-	// RetentionFilter is a function which returns the blocks which can be deleted.
+	// BlocksToDelete is a function which returns the blocks which can be deleted.
 	// It is always the default time and size based retention in Prometheus and
 	// mainly meant for external users who import TSDB.
-	RetentionFilter func(blocks []*Block) map[ulid.ULID]*Block
+	BlocksToDelete BlocksToDeleteFunc
 }
+
+type BlocksToDeleteFunc func(blocks []*Block) map[ulid.ULID]struct{}
 
 // DB handles reads and writes of time series falling into
 // a hashed partition of a seriedb.
@@ -131,12 +133,12 @@ type DB struct {
 	dir   string
 	lockf fileutil.Releaser
 
-	logger          log.Logger
-	metrics         *dbMetrics
-	opts            *Options
-	chunkPool       chunkenc.Pool
-	compactor       Compactor
-	retentionFilter func(blocks []*Block) map[ulid.ULID]*Block
+	logger         log.Logger
+	metrics        *dbMetrics
+	opts           *Options
+	chunkPool      chunkenc.Pool
+	compactor      Compactor
+	blocksToDelete BlocksToDeleteFunc
 
 	// Mutex for that must be held when modifying the general block layout.
 	mtx    sync.RWMutex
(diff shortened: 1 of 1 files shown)


tsdb/db.go
```diff
@@ -560,14 +566,18 @@ func open(dir string, l log.Logger, r prometheus.Registerer, opts *Options, rngs
...
-		opts:        opts,
-		compactc:    make(chan struct{}, 1),
-		donec:       make(chan struct{}),
-		stopc:       make(chan struct{}),
-		autoCompact: true,
-		chunkPool:   chunkenc.NewPool(),
+		dir:             dir,
+		logger:          l,
+		opts:            opts,
+		compactc:        make(chan struct{}, 1),
```
bwplotka, line 572: actually it's the time in my life I see `c` suffix fo channel :thinking:

tsdb/db.go
```diff
@@ -118,6 +118,11 @@ type Options struct {
 	// SeriesLifecycleCallback specifies a list of callbacks that will be called during a lifecycle of a series.
 	// It is always a no-op in Prometheus and mainly meant for external users who import TSDB.
 	SeriesLifecycleCallback SeriesLifecycleCallback
+
+	// RetentionFilter is a function which returns the blocks which can be deleted.
+	// It is always the default time and size based retention in Prometheus and
+	// mainly meant for external users who import TSDB.
+	RetentionFilter func(blocks []*Block) map[ulid.ULID]*Block
```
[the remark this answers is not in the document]
codesome: In that case, it could just return a slice of ULID :)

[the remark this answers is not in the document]
codesome: Or maybe map is better to avoid needed addition deduplication calculation

roidelapluie:
Could we make our filters,e.g. beyondTimeRetention , public? does that make sense for thanos/cortex?

codesome:
> Could we make our filters,e.g. beyondTimeRetention , public? does that make sense for downstreams?

I had made only the combination public. But it would make sense to make both public. I will do it in this PR.

codesome:
I have updated the PR with
1. Name change
2. `*Block -> struct{}`
3. Additional `type BlocksToDeleteFunc func(blocks []*Block) map[ulid.ULID]struct{}` for a single place to tinker the function signature.

tsdb/db.go
```diff
@@ -119,24 +119,26 @@ type Options struct {
 	// It is always a no-op in Prometheus and mainly meant for external users who import TSDB.
 	SeriesLifecycleCallback SeriesLifecycleCallback
 
-	// RetentionFilter is a function which returns the blocks which can be deleted.
+	// BlockToDelete is a function which returns the blocks which can be deleted.
 	// It is always the default time and size based retention in Prometheus and
 	// mainly meant for external users who import TSDB.
-	RetentionFilter func(blocks []*Block) map[ulid.ULID]*Block
+	BlockToDelete func(blocks []*Block) map[ulid.ULID]struct{}
```
pracucci, line 125: Then why not `BlocksToDelete()` given it handle 1+ blocks? :)

tsdb/db.go
```diff
@@ -118,19 +118,27 @@ type Options struct {
 	// SeriesLifecycleCallback specifies a list of callbacks that will be called during a lifecycle of a series.
 	// It is always a no-op in Prometheus and mainly meant for external users who import TSDB.
 	SeriesLifecycleCallback SeriesLifecycleCallback
+
+	// BlocksToDelete is a function which returns the blocks which can be deleted.
+	// It is always the default time and size based retention in Prometheus and
+	// mainly meant for external users who import TSDB.
+	BlocksToDelete func(blocks []*Block) map[ulid.ULID]struct{}
```
pracucci, line 125:
```suggestion
	BlocksToDelete BlocksToDeleteFunc
```

codesome: :facepalm: Thanks!

PR #7639, opened 2020-07-22
parser.{Expr,Statement} publicly implementable

owen-d:
This PR makes both `parser.Expr` and `parser.Statement` interfaces publicly implementable without roundabout embedding. I've made the new exported methods use the `PromQL` prefix for two reasons:
1) to prevent method/field collisions (for instance, the `ParenExpr` struct has a field `Expr`)
2) to make accidental implementation more difficult now that these methods are exported.

This is a followup of https://github.com/prometheus/prometheus/pull/7569

### DIFF
--- a/promql/parser/ast.go
+++ b/promql/parser/ast.go
@@ -26,7 +26,7 @@ import (
 // Node is a generic interface for all nodes in an AST.
 //
 // Whenever numerous nodes are listed such as in a switch-case statement
-// or a chain of function definitions (e.g. String(), expr(), etc.) convention is
+// or a chain of function definitions (e.g. String(), PromQLExpr(), etc.) convention is
 // to list them as follows:
 //
 // 	- Statements
@@ -49,9 +49,9 @@ type Node interface {
 type Statement interface {
 	Node
 
-	// stmt ensures that no other type accidentally implements the interface
+	// Stmt ensures that no other type accidentally implements the interface
 	// nolint:unused
-	stmt()
+	PromQLStmt()
 }
 
 // EvalStmt holds an expression and information on the range it should
@@ -66,7 +66,7 @@ type EvalStmt struct {
 	Interval time.Duration
 }
 
-func (*EvalStmt) stmt() {}
+func (*EvalStmt) PromQLStmt() {}
 
 // Expr is a generic interface for all expression types.
 type Expr interface {
@@ -75,8 +75,8 @@ type Expr interface {
 	// Type returns the type the expression evaluates to. It does not perform
 	// in-depth checks as this is done at parsing-time.
 	Type() ValueType
-	// expr ensures that no other types accidentally implement the interface.
-	expr()
+	// Expr ensures that no other types accidentally implement the interface.
+	PromQLExpr()
 }
 
 // Expressions is a list of expression nodes that implements Node.
@@ -180,7 +180,7 @@ type VectorSelector struct {
 type TestStmt func(context.Context) error
 
 func (TestStmt) String() string { return "test statement" }
-func (TestStmt) stmt()          {}
+func (TestStmt) PromQLStmt()    {}
 
 func (TestStmt) PositionRange() PositionRange {
 	return PositionRange{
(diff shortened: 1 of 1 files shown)


owen-d:
@roidelapluie Here you go, let me know if you'd prefer that I handle the method/field collisions differently.

promql/parser/ast.go
```diff
@@ -49,9 +49,9 @@ type Node interface {
 type Statement interface {
 	Node
 
-	// stmt ensures that no other type accidentally implements the interface
+	// Stmt ensures that no other type accidentally implements the interface
```
brian-brazil, line 52: This and the one below need updating.

owen-d: Ah thanks. Will do.

slrtbtfs:
The whole point why these private interface functions exist, seems to be to make them not implementable from the outside. This is somewhat unidiomatic.

Instead of making these interface functions public, we could just remove them completely and let `Node` be the only interface used for AST Nodes.

brian-brazil:
> The whole point why these private interface functions exist, seems to be to make them not implementable from the outside. This is somewhat unidiomatic.

It's to prevent accidents. We don't support external implementation - but we'll usually be willing to make things public for the sake of it.

roidelapluie:
I find the prefix strange as you want to implement it for something else than PromQL.

brian-brazil:
I think a prefix is fine, it prevents confusion that these are methods that are actually useful in and of themselves - plus less likely to clash which is the point of these.

roidelapluie:
> I think a prefix is fine, it prevents confusion that these are methods that are actually useful in and of themselves - plus less likely to clash which is the point of these.

Yes, just the content of the prefix is strange, as it is exposed purposely for something else than PromQL

slrtbtfs:
> It's to prevent accidents. We don't support external implementation - but we'll usually be willing to make things public for the sake of it.


What kinds of accidents can this really prevent?

brian-brazil:
> What kinds of accidents can this really prevent?

A type which happens to satisfy Node being incorrectly treated as Node. It's a risk of the duck typing that Go does.

roidelapluie:
> > I think a prefix is fine, it prevents confusion that these are methods that are actually useful in and of themselves - plus less likely to clash which is the point of these.
> 
> Yes, just the content of the prefix is strange, as it is exposed purposely for something else than PromQL

mmh this is still in the PromQL tree to the prefix is fine.

slrtbtfs:
> A type which happens to satisfy Node being incorrectly treated as Node. It's a risk of the duck typing that Go does.

Sure, but do we really need `Statement` and `Expression` on top of that?

owen-d:
Seeing a data race issue in the untouched discovery package during the `build` CI step. Can I disregard this?

promql/parser/ast.go
```diff
@@ -49,9 +49,9 @@ type Node interface {
 type Statement interface {
 	Node
 
-	// stmt ensures that no other type accidentally implements the interface
+	// PromQLStmt ensures that no other type accidentally implements the interface
```
cstyan, line 52: this comment seems inaccurate/unnecessary now

promql/parser/ast.go
```diff
@@ -75,8 +75,8 @@ type Expr interface {
 	// Type returns the type the expression evaluates to. It does not perform
 	// in-depth checks as this is done at parsing-time.
 	Type() ValueType
-	// expr ensures that no other types accidentally implement the interface.
-	expr()
+	// PromQLExpr ensures that no other types accidentally implement the interface.
```
cstyan, line 78: same here

brian-brazil:
> I imagine our stance is still going to be "you shouldn't implement these interfaces unless you know what you're doing", should we leave a comment to that effect?

It's more that if you use this code at all not in the form of the Prometheus binary then it's entirely at your own risk. These interfaces are not special in this regard, so comments like that would only be clutter as they'd apply to everything.

promql/parser/ast.go
brian-brazil: The comment is needed to explain why this dead code is here.

cstyan:
> These interfaces are not special in this regard, so comments like that would only be clutter as they'd apply to everything.

Fair, I couldn't remember whether we'd introduced comments to this affect in other areas.

owen-d:
Perfect. Let me know if any additional action is needed from me.

promql/parser/ast.go
cstyan: @brian-brazil my point was that the wording, specifically the use of `accidentally`, is confusing and inaccurate in my opinion now that this function is exported.

brian-brazil: I'm not seeing that, the accident would be having a type that didn't mean to implement Statement inadvertently implementing Statement.

brian-brazil:
Are there any objections to me merging this as-is? If not I plan to merge this tomorrow.

roidelapluie:
OK for me.

cstyan:
:+1: no objections

brian-brazil:
Thanks!

PR #7641, opened 2020-07-22
Cut final v2.20.0

beorn7:
There is one minor change compared to the rc.1, but @roidelapluie expressed confidence that we don't need another RC for that one.


brian-brazil:
:+1: Once tests pass. I've kicked them off again.

PR #7645, opened 2020-07-22
Add files via upload

aaronnewcomb:
Please use this file to update the Sysdig logo on prometheus.io


roidelapluie:
Thank you for this. The prometheus.io website source is at https://github.com/prometheus/docs . We are also only likely to accept this if you are affiliated with sysdig.

PR #7651, opened 2020-07-23
Improve scalar literal format documentation

juliusv:

### DIFF
--- a/docs/querying/basics.md
+++ b/docs/querying/basics.md
@@ -59,7 +59,7 @@ Scalar float values can be written as literal integer or floating-point numbers 
 
     [-+]?(
           [0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?
-        | [-+]0[xX][0-9a-fA-F]+
+        | 0[xX][0-9a-fA-F]+
         | [nN][aA][nN]
         | [iI][nN][fF]
     )
--- a/docs/querying/basics.md
+++ b/docs/querying/basics.md
@@ -55,19 +55,23 @@ Example:
 
 ### Float literals
 
-Scalar float values can be literally written as integer or floating-point numbers, for example:
+Scalar float values can be written as literal integer or floating-point numbers in the format (whitespace only included for better readability):
+
+    [-+]?(
+          [0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?
+        | [-+]0[xX][0-9a-fA-F]+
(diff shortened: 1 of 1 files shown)


roidelapluie:
Thanks for pointing out this bug :)

#5984 should be applied to PromQL too :)

roidelapluie:
Mmmh it seems like we are not affected by #5984 in promql. However, I suggest we note that we do not support these.

docs/querying/basics.md
```diff
@@ -55,10 +55,19 @@ Example:
...
 ### Float literals
 
-Scalar float values can be literally written as numbers of the form
-`[-](digits)[.(digits)]`.
+Scalar float values can be literally written as integer or floating-point numbers, for example:
 
+    23
     -2.43
+    3.4e-9
+    0x8f
```
brian-brazil, line 63: Is this something we actually support? I think we should stick to basic floats and ints.

docs/querying/basics.md
```diff
@@ -55,10 +55,19 @@ Example:
...
+    -Inf
+
+Prometheus first attempts to parse scalar literals using Go's [`strconv.ParseInt()`](https://golang.org/pkg/strconv/#ParseInt)
```
brian-brazil, line 66: I don't think we should be mentioning implementation details like this, also it's the PromQL grammar that defines this - not the Go implementation.

docs/querying/basics.md
roidelapluie: We do not support that and #5984 made sure we do not scrape that either.

roidelapluie: Oh wait,yes we support that

brian-brazil: Indeed, this is not something we support. https://github.com/prometheus/prometheus/blob/master/promql/parser/lex.go#L717

roidelapluie: In practice: https://prometheus.demo.do.prometheus.io/graph?g0.range_input=1h&g0.expr=0x1&g0.tab=1

brian-brazil: So it is

juliusv:
It's funny how nobody was even sure what kind of number formats Prometheus accepts, good that we are documenting this :)

I removed the references to Go now and included a regex-based approach instead.

brian-brazil:
:+1:

PR #7652, opened 2020-07-23
Add more Go leak tests

roidelapluie:
cc @bwplotka

### DIFF
--- a/util/testutil/testing.go
+++ b/util/testutil/testing.go
@@ -25,9 +25,11 @@ package testutil
 import (
 	"fmt"
 	"reflect"
+	"testing"
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/pmezard/go-difflib/difflib"
+	"go.uber.org/goleak"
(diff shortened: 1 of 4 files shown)


util/testutil/testing.go
```diff
@@ -154,3 +156,14 @@ func formatMessage(msgAndArgs []interface{}) string {
 	}
 	return ""
 }
+
+// TolerantVerifyLeak verifies go leaks but excludes the go routines that are
+// launched as side effects of some of our dependencies.
+func TolerantVerifyLeak(m *testing.M) {
```
bwplotka, line 162: mhmhm, let's remove this function and add those explicitly so we are aware. This should go at some point as issue to such projects :hugs:

roidelapluie: Well they will be the same in a *lot* of prometheus packages.

bwplotka: mhmhmh, fair

PR #7653, opened 2020-07-24
Update tsdb.go

chicknsoup:
Added DURATION column to `tsdb list` command
The timestamps are just fine but I think showing the exact block duration will make it more human readable.

### DIFF
--- a/cmd/promtool/tsdb.go
+++ b/cmd/promtool/tsdb.go
@@ -363,15 +363,16 @@ func printBlocks(blocks []tsdb.BlockReader, humanReadable bool) {
 	tw := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
 	defer tw.Flush()
 
-	fmt.Fprintln(tw, "BLOCK ULID\tMIN TIME\tMAX TIME\tNUM SAMPLES\tNUM CHUNKS\tNUM SERIES")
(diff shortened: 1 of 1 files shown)


cmd/promtool/tsdb.go
```diff
@@ -363,15 +363,16 @@ func printBlocks(blocks []tsdb.BlockReader, humanReadable bool) {
...
 	for _, b := range blocks {
 		meta := b.Meta()
 
 		fmt.Fprintf(tw,
-			"%v\t%v\t%v\t%v\t%v\t%v\n",
+			"%v\t%v\t%v\t%v\t%v\t%v\t%v\n",
 			meta.ULID,
 			getFormatedTime(meta.MinTime, humanReadable),
 			getFormatedTime(meta.MaxTime, humanReadable),
+			(meta.MaxTime-meta.MinTime)/(3600*1000),
```
codesome, line 375: The header does not mention if it is hours/minutes/etc and we could lose precision if the block does not align with an hour. How about this:
```suggestion
			time.Duration(meta.MaxTime-meta.MinTime)*time.Millisecond,
```

chicknsoup: Yep, this is even better.

roidelapluie:
Isn't TSDB human-time-agnostic?

Issue #7647, opened 2020-07-23
Use uber-go/atomic in place of sync/atomic


codesome:
1. Avoids bugs like https://github.com/prometheus/prometheus/pull/7646 where the atomic variable is not accessed atomically everywhere.
2. Eliminates the need of keeping atomic variables at the top of the struct

Thanks @pstibrany for pointing it out!

johncming:
The case used `sync/atomic` in `tsdb` pkg.

tsdb/chunks/head_chunks.go
```
# fields in struct
ChunkDiskMapper.curFileNumBytes
```

tsdb/head.go
```
# fields in struct
Head.minTime
Head.maxTime
Head.minValidTime
Head.numSeries
Head.lastSeriesID

# variable in function
unknownRefs in Head.loadWAL method
```

bwplotka:
Nice! 

https://github.com/uber-go/atomic

jvrplmlmn:
I would like to take on this issue. It seems something approachable to get familiar with the codebase and start contributing to the project 😃

codesome:
@jvrplmlmn Sure! Go ahead

jvrplmlmn:
First PR that addresses the usage of `sync/atomic` in the `tsdb` package: https://github.com/prometheus/prometheus/pull/7659 

If the goal is to replace it in the whole codebase let me know and I will send following PRs to that one.

codesome:
These are the remaining places after the first PR
```
➜  prometheus git:(master) grep -r --exclude-dir=vendor "sync/atomic"
notifier/notifier.go:	"sync/atomic"
notifier/notifier_test.go:	"sync/atomic"
storage/remote/intern_test.go:	"sync/atomic"
storage/remote/intern.go:	"sync/atomic"
storage/remote/intern.go:	// See https://golang.org/pkg/sync/atomic/#pkg-note-BUG for more info.
storage/remote/queue_manager_test.go:	"sync/atomic"
storage/remote/ewma.go:	"sync/atomic"
storage/remote/ewma.go:	// See https://golang.org/pkg/sync/atomic/#pkg-note-BUG for more info.
storage/remote/queue_manager.go:	"sync/atomic"
storage/remote/queue_manager.go:	// https://golang.org/pkg/sync/atomic/#pkg-note-BUG
web/web.go:	"sync/atomic"
cmd/prometheus/main.go:	"sync/atomic"
tsdb/head_bench_test.go:	"sync/atomic"
```

[...]

PR #7659, opened 2020-07-24
tsdb: Replace sync/atomic with uber-go/atomic

jvrplmlmn:
Related with https://github.com/prometheus/prometheus/issues/7647, particularly for the `tsdb` package.

As this was my first contribution, I've preferred to keep the scope small and send additional PRs to keep iterating on https://github.com/prometheus/prometheus/issues/7647.

@codesome you are listed as a [maintainer](https://github.com/prometheus/prometheus/blob/master/MAINTAINERS.md) of this package, care to take a look? 👀

### DIFF
--- a/tsdb/chunks/head_chunks.go
+++ b/tsdb/chunks/head_chunks.go
@@ -25,12 +25,12 @@ import (
 	"sort"
 	"strconv"
 	"sync"
-	"sync/atomic"
 
 	"github.com/pkg/errors"
 	"github.com/prometheus/prometheus/tsdb/chunkenc"
 	tsdb_errors "github.com/prometheus/prometheus/tsdb/errors"
 	"github.com/prometheus/prometheus/tsdb/fileutil"
+	"go.uber.org/atomic"
 )
 
 // Head chunk file header fields constants.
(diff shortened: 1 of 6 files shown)


tsdb/db.go
```diff
@@ -762,7 +762,7 @@ func (db *DB) Compact() (err error) {
 			break
 		}
 		mint := db.head.MinTime()
-		maxt := rangeForTimestamp(mint, db.head.chunkRange)
+		maxt := rangeForTimestamp(mint, db.head.chunkRange.Load())
```
johncming, line 765: `chunkRange` only wrote on the init stage. Is it need to refactor?

bwplotka: I think it's behind the lock, that's why it was not atomically changed/loaded so far. I think Load is fine for this though

tsdb/head.go
```diff
@@ -53,11 +53,11 @@ var (
 type Head struct {
 	// Keep all 64bit atomically accessed variables at the top of this struct.
```
bwplotka, line 54: We can remove all of those comments (:

tsdb/chunks/head_chunks.go
```diff
@@ -80,7 +80,7 @@ func (e *CorruptionErr) Error() string {
 type ChunkDiskMapper struct {
 	// Keep all 64bit atomically accessed variables at the top of this struct.
 	// See https://golang.org/pkg/sync/atomic/#pkg-note-BUG for more info.
```
codesome, line 82: We can remove this comment everywhere `go.uber.org/atomic` is used :)

codesome: Ah, repeated comment :)

jvrplmlmn:
@bwplotka @codesome comments addressed!

codesome:
Thanks!

codesome:
Realized after merging that we missed https://github.com/prometheus/prometheus/blob/master/tsdb/head_bench_test.go from tsdb.

PR #7660, opened 2020-07-24
Add context to storage.Appendable interface

annanay25:
Hi Prometheus team,

We have a use case to allow extending the Ruler functionality in Cortex, and use different `storage.Appender`s depending on specific `Group` config options.

The proposal is to enable passing group specific information to `storage.Appendable`,  and then a different `storage.Appender` can be created depending on the config options passed.

This PR achieves this by passing a `context.Context` parameter to the `Appender` function in the `storage.Appendable` interface. This is also a good practice in general, because it allows for context propagation along the write path.

Signed-off-by: Annanay <annanayagarwal@gmail.com>


bwplotka:
Can you maybe show the code that might need that in Cortex? Some example. There might be more ways to accomplish this like your struct having many Appenders. Hard to say without examples. It would be nice to NOT extend interface if absolutely required. :hugs:

roidelapluie:
> Can you maybe show the code that might need that in Cortex? Some example. There might be more ways to accomplish this like your struct having many Appenders. Hard to say without examples. It would be nice to NOT extend interface if absolutely required.

Couldn't we benefit from this in prometheus too ? Having timeouts on appenders?

brian-brazil:
> Couldn't we benefit from this in prometheus too ? Having timeouts on appenders?

If appends aren't instant, the tsdb is already likely completely broken.

brancz:
I realize it's a bit of a cross cutting concern, but for tracing purposes I would very much welcome propagating a context through the storage layer. As it stands all I tend to see is one huge span in traces, it would be super nice to see more detail.

annanay25:
Thanks for the comments everyone :)

@bwplotka, let me do my best to give some more context 😉 

1. In Cortex, this is the struct that implements `Appendable` -
https://github.com/cortexproject/cortex/blob/7c069848b43325b468ad7bb618b73dba2f07ebfa/pkg/ruler/compat.go#L55-L59
2. This gets passed to the `Manager` as options - https://github.com/cortexproject/cortex/blob/7c069848b43325b468ad7bb618b73dba2f07ebfa/pkg/ruler/ruler.go#L539
3. Each group gets the same set of `ManagerOptions` here - https://github.com/prometheus/prometheus/blob/6f296594a8522600eff18c39b438ae40ebdb7477/rules/manager.go#L1051

So even if I had an `appender{}` that has multiple `Appenders`, it does not know _which one to pick_ for a particular group.

With the change that this PR brings, our `appender{}` could look like this -

```
type appender struct {
	appenders map[groupName]storage.Appender
	userID string
}

// Appender returns a storage.Appender
func (t *appender) Appender(ctx context.Context) storage.Appender {
	return &appendable{
		pusher: t.appenders[ctx.Value(groupName)],
		userID: t.userID,
	}
}
```

Another option that could work is if we could create `GroupOptions` on our own. To this end, we considered this - https://github.com/prometheus/prometheus/compare/master...annanay25:grouploader-refactor - but decided to drop it because it was not a good API to expose from the Manager.

brancz:
I guess what I was trying to say is, even if it was just for tracing, I'd be happy with this.

bwplotka:
Yea, I think I would be happy with this for tracing purposes, but time limiting and special meta passing might be not needed. 

Just curious why would you want different appenders per group? I think this is to make it concurrent, but then Isn't it something that Ruler manager itself could be doing in Prometheus?

roidelapluie:
It seems that cortex is reusing the same appender after Commit*(, which should not happen if you follow the Appendable interface doc.

annanay25:
> but then Isn't it something that Ruler manager itself could be doing in Prometheus?

Not really, Prometheus Manager has one set of options it uses for all RuleGroups. So we can't instantiate different appenders for different RuleGroups as of now.

roidelapluie:
Oh now I see it. But maybe that would be better managed with multiple rule managers.

scrape/scrape_test.go
```diff
@@ -374,8 +375,9 @@ func TestScrapePoolRaces(t *testing.T) {
 
 func TestScrapeLoopStopBeforeRun(t *testing.T) {
 	scraper := &testScraper{}
+	ctx := context.Background()
 
-	sl := newScrapeLoop(context.Background(),
+	sl := newScrapeLoop(ctx,
```
codesome, line 380: We don't need a new variable here. Same for all the other places, we can use `context.Background()` directly.

scrape/scrape_test.go
```diff
@@ -300,6 +300,7 @@ func TestScrapePoolReload(t *testing.T) {
 func TestScrapePoolAppender(t *testing.T) {
 	cfg := &config.ScrapeConfig{}
 	app := &nopAppendable{}
+	ctx := context.Background()
```
codesome, line 303: We can just use `context.Background()` in those 2 places, I don't think a new variable is required.

storage/remote/storage.go
```diff
@@ -169,7 +169,7 @@ func (s *Storage) ChunkQuerier(ctx context.Context, mint, maxt int64) (storage.C
 }
 
 // Appender implements storage.Storage.
-func (s *Storage) Appender() storage.Appender {
+func (s *Storage) Appender(_ context.Context) storage.Appender {
 	return s.rws.Appender()
```
codesome, line 173: Any reason for not passing it down here?

tsdb/compact_test.go
```diff
@@ -911,7 +911,8 @@ func TestDisableAutoCompactions(t *testing.T) {
 	// Trigger a compaction to check that it was skipped and
 	// no new blocks were created when compaction is disabled.
 	db.DisableCompactions()
-	app := db.Appender()
+	ctx := context.Background()
+	app := db.Appender(ctx)
```
codesome, line 915: Same, don't need new variables here.

tsdb/db.go
```diff
@@ -710,7 +710,7 @@ func (db *DB) run() {
 }
 
 // Appender opens a new appender against the database.
-func (db *DB) Appender() storage.Appender {
+func (db *DB) Appender(ctx context.Context) storage.Appender {
 	return dbAppender{db: db, Appender: db.head.Appender()}
```
codesome, line 714: And what about passing it down to `db.head.Appender()`

storage/remote/storage.go
annanay25: None. Fixed this.

storage/interface.go
```diff
@@ -34,7 +34,7 @@ var (
 // Appendable allows creating appenders.
 type Appendable interface {
 	// Appender returns a new appender for the storage.
-	Appender() Appender
+	Appender(ctx context.Context) Appender
```
gouthamve, line 37: Can we write a comment here that we might use the context to pass in more information or for tracing but not for timeouts?

roidelapluie: I think that this should be the choice of the implementation. Maybe we can add that the appender might decide not to use the context or check for context errors.

codesome: >  appender might decide not to use the context or check for context errors.

This sounds better and more general.

annanay25: Added comment.

PR #7661, opened 2020-07-24
promtool: Ensure alert rules are marked as restored in unit tests

dgl:
This makes sure the ALERTS timeseries is created when unit testing
alerting rules.

Signed-off-by: David Leadbeater <dgl@dgl.cx>

### DIFF
--- a/cmd/promtool/testdata/unittest.yml
+++ b/cmd/promtool/testdata/unittest.yml
@@ -8,6 +8,14 @@ tests:
     input_series:
       - series: 'up{job="prometheus", instance="localhost:9090"}'
         values: "0+0x1440"
+
+    promql_expr_test:
+      - expr: count(ALERTS) by (alertname, alertstate)
+        eval_time: 4m
+        exp_samples:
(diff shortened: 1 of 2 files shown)


dgl:
cc @simonpasquier -- as a maintainer of promtool

roidelapluie:
Not sure if that is the right fix ; shouldn't that happen out of the box? What are we doing wrong now?

dgl:
Essentially it's done by the rule code when [needed](https://github.com/prometheus/prometheus/blob/master/rules/manager.go#L361).

This code reimplements that run function, as it needs control of time of evaluation so it makes sense it should also do this. I think it's also the easiest behaviour to reason about for people writing tests, the ALERTS timeseries won't be delayed.

It's probably possible for this code to directly call [RestoreForState](https://github.com/prometheus/prometheus/blob/master/rules/manager.go#L668) -- but essentially the code in that would reduce to doing what I wrote, while potentially being more fragile.

roidelapluie:
LGTM.

cc @simonpasquier

roidelapluie:
Thanks!

PR #7667, opened 2020-07-24
Hints: Separating out the range and offsets of PromQL subqueries

### LINKED ISSUE #7629: promql: Improve what is Selected when subquery is used.
As suggested by @gouthamve we can improve a lot on what we select form storage for subqueries.

See TODO: https://github.com/prometheus/prometheus/blob/ce1db029095e3246591fd157b3ccddce01313019/promql/engine.go#L654

roidelapluie:
Fix #7629

### DIFF
--- a/promql/engine.go
+++ b/promql/engine.go
@@ -576,35 +576,39 @@ func (ng *Engine) execEvalStmt(ctx context.Context, query *query, s *parser.Eval
 	return mat, warnings, nil
 }
 
-// cumulativeSubqueryOffset returns the sum of range and offset of all subqueries in the path.
-func (ng *Engine) cumulativeSubqueryOffset(path []parser.Node) time.Duration {
-	var subqOffset time.Duration
+// subqueryOffsetRange returns the sum of offsets and ranges of all subqueries in the path.
+func (ng *Engine) subqueryOffsetRange(path []parser.Node) (time.Duration, time.Duration) {
+	var (
+		subqOffset time.Duration
+		subqRange  time.Duration
+	)
 	for _, node := range path {
 		switch n := node.(type) {
 		case *parser.SubqueryExpr:
-			subqOffset += n.Range + n.Offset
+			subqOffset += n.Offset
+			subqRange += n.Range
 		}
 	}
-	return subqOffset
+	return subqOffset, subqRange
 }
 
 func (ng *Engine) findMinTime(s *parser.EvalStmt) time.Time {
 	var maxOffset time.Duration
 	parser.Inspect(s.Expr, func(node parser.Node, path []parser.Node) error {
-		subqOffset := ng.cumulativeSubqueryOffset(path)
+		subqOffset, subqRange := ng.subqueryOffsetRange(path)
 		switch n := node.(type) {
 		case *parser.VectorSelector:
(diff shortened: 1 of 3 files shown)


promql/engine.go
```diff
@@ -597,35 +597,39 @@ func (ng *Engine) execEvalStmt(ctx context.Context, query *query, s *parser.Eval
...
 	parser.Inspect(s.Expr, func(node parser.Node, path []parser.Node) error {
-		subqOffset := ng.cumulativeSubqueryOffset(path)
+		subqOffset, subqRange := ng.subqueryOffsetRange(path)
 		switch n := node.(type) {
 		case *parser.VectorSelector:
-			if maxOffset < ng.lookbackDelta+subqOffset {
-				maxOffset = ng.lookbackDelta + subqOffset
+			if maxOffset < ng.lookbackDelta+subqOffset+subqRange {
+				maxOffset = ng.lookbackDelta + subqOffset + subqRange
 			}
```
roidelapluie, line 624: @brian-brazil GIven than n.offset is > 0 always
those 3 lines above seem redundant with below?

promql/engine.go
```diff
@@ -597,35 +597,39 @@ func (ng *Engine) execEvalStmt(ctx context.Context, query *query, s *parser.Eval
...
-				maxOffset = n.Offset + ng.lookbackDelta + subqOffset
+			if n.Offset+ng.lookbackDelta+subqOffset+subqRange > maxOffset {
+				maxOffset = n.Offset + ng.lookbackDelta + subqOffset + subqRange
 			}
 		case *parser.MatrixSelector:
-			if maxOffset < n.Range+subqOffset {
-				maxOffset = n.Range + subqOffset
+			if maxOffset < n.Range+subqOffset+subqRange {
+				maxOffset = n.Range + subqOffset + subqRange
 			}
```
roidelapluie, line 631: Same here

roidelapluie: Unless we are preparing for negatif offsets...

promql/engine.go
bwplotka: +1, but happy with negative offset to be well handed here as well

bwplotka:
I don't think https://github.com/prometheus/prometheus/issues/7629 is fixed by this, though. I think there is lots to improve in terms of `step` (interval). I think the lowest-level subquery interval should be passed as step.

Also it would be nice to add test cases for multiple nested subqueries within one query? :thinking:

roidelapluie:
```
		query: "count_over_time((foo offset 10s)[2m:1s] offset 10s)", start: 300000, end: 500000,
￼		expected: []*storage.SelectHints{
￼			{Start: 155000, End: 480000, Func: "count_over_time", Step: **1000**},
￼		},
```
Sounds like that is the case

promql/engine.go
brian-brazil: There can be multiple different offsets in one expression, which I believe this is handling. We may need a few more tests here.

roidelapluie: I have added a test. It handles foo offset 1m or foo offset 10m.

roidelapluie:
@bwplotka I think it is all addressed. WDYT?

PR #7675, opened 2020-07-27
Add rule importer to backfill

JessicaGreben:
This PR implements a solution for issue #11 to add the ability to retroactively evaluate new rules so that there is historical data for new rules. For details see design doc [here.](https://docs.google.com/document/d/16s_-RxYwQYcb4G4mvpmjPolEbD24o54a1LPqJ538Vhc/edit?usp=sharing)

This PR includes the following code changes:
- creates a rule importer. The rule importer parses recording rule files and creates tsdb blocks of recording rules from historical prometheus data.
- creates a promtool subcommand  `promtool tsdb create-blocks-from rules` to execute backfilling data for a new rule
- the output of the subcommand is a directory that contains blocks with the series created from evaluating the recording rules on old data

Things not in the PR that must be added:
- add documentation in the "storage" page ([ref](https://github.com/prometheus/prometheus/blob/main/docs/storage.md#backfilling-from-openmetrics-format)) 
- add stale markers

Things not in this PR, but might be nice to add at a later date:
- long running service to backfill any amount of data
- backfill alerting rules
- ability to move the output blocks to a prometheus data dir and perform validation before doing so 

Example usage:

```
$ promtool tsdb create-blocks-from rules --start=1606382352 --end=1606389552 \
    --output-dir <path/to/dir/to/write/data> \
    --url <prometheus api addr and port> \
    rules1.yml,rules2.yml
```


cmd/promtool/main.go
```diff
@@ -128,6 +130,20 @@ func main() {
...
 	dumpMaxTime := tsdbDumpCmd.Flag("max-time", "Maximum timestamp to dump.").Default(strconv.FormatInt(math.MaxInt64, 10)).Int64()
 
+	backfillCmd := app.Command("backfill", "Backfill Prometheus data.")
+	backfillRuleCmd := backfillCmd.Command("rules", "Backfill Prometheus data for new rules.")
+	backfillRuleURL := backfillRuleCmd.Flag("url", "Prometheus API url.").Required().String()
+	backfillRuleEvalInterval := backfillRuleCmd.Flag("evaluation_interval", "How frequently to evaluate rules when backfilling.").
+		Default("-3h").Duration()
+	backfillRuleStart := backfillRuleCmd.Flag("start", "The time to start backfilling the new rule from. It is required. Start time should be RFC3339 or Unix timestamp.").
+		Required().Duration()
+	backfillRuleEnd := backfillRuleCmd.Flag("end", "If an end time is provided, the new rule backfilling will end at this time. The default will backfill to the 3 hrs ago. End time should be RFC3339 or Unix timestamp.").
```
brian-brazil, line 140: s/the//

cmd/promtool/main.go
```diff
@@ -128,6 +130,20 @@ func main() {
...
+		Default("").Duration()
+	backfillRuleFiles := backfillRuleCmd.Arg(
+		"rule-files",
+		"The file containing the new rule that needs to be backfilled.",
```
brian-brazil, line 144: file or files?

cmd/promtool/main.go
```diff
@@ -747,3 +766,38 @@ func (j *jsonPrinter) printLabelValues(v model.LabelValues) {
 	//nolint:errcheck
 	json.NewEncoder(os.Stdout).Encode(v)
 }
+
+// BackfillRule backfills rules from the files provided
```
brian-brazil, line 770: full stop

cmd/promtool/main.go
```diff
@@ -747,3 +766,38 @@ func (j *jsonPrinter) printLabelValues(v model.LabelValues) {
...
+func BackfillRule(url string, start, end, evalInterval time.Duration, files ...string) int {
+	ctx := context.Background()
+	cfg := importers.RuleConfig{
+		Start:        start.String(),
```
brian-brazil, line 774: Why are you converting a string to a string?

I'd expect these args to be validated and parsed here.

importers/rules.go
```diff
@@ -0,0 +1,256 @@
...
+	return &RuleImporter{
+		config: config,
+	}
+}
+
+// Init initializes the rule importer which creates a new block writer
+// and creates an Prometheus API client
+func (importer *RuleImporter) Init() error {
+	// create new block writer
+	newBlockDir, err := ioutil.TempDir("", "rule_blocks")
```
brian-brazil, line 72: /tmp may not be big enough, I'd suggest working in the output directory.

When do you clean this up?

importers/rules.go
```diff
@@ -0,0 +1,256 @@
...
+// Close cleans up any open resources
+func (importer *RuleImporter) Close() error {
+	return importer.writer.Close()
+}
+
+// Parse parses the groups and rules from a list of rules files
+func (importer *RuleImporter) Parse(ctx context.Context, files []string) (errs []error) {
+	groups := make(map[string]*rules.Group)
+
+	for _, file := range files {
```
brian-brazil, line 99: We seem to be duplicating code with the manager here

tsdb/importer/blocks/writer.go
```diff
@@ -0,0 +1,139 @@
...
+// Flush implements Writer interface. This is where actual block writing
+// happens. After flush completes, no write can be done.
+func (w *TSDBWriter) Flush() ([]ulid.ULID, error) {
+	seriesCount := w.head.NumSeries()
+	if w.head.NumSeries() == 0 {
+		return nil, errors.New("no series appended; aborting.")
+	}
+
+	mint := w.head.MinTime()
+	maxt := w.head.MaxTime() + 1
```
brian-brazil, line 114: This +1 needs a comment.

tsdb/importer/blocks/writer.go
```diff
@@ -0,0 +1,139 @@
...
+
+// NewTSDBWriter create new block writer.
+//
+// The returned writer accumulates all series in memory until `Flush` is called.
+//
+// Note that the writer will not check if the target directory exists or
+// contains anything at all. It is the caller's responsibility to
+// ensure that the resulting blocks do not overlap etc.
+// Writer ensures the block flush is atomic (via rename).
+func NewTSDBWriter(logger log.Logger, dir string) (*TSDBWriter, error) {
```
brian-brazil, line 69: Why do we need this? We've already similar code in the unittests for the tsdb, and it overall seems relatively trivial so could be handled inline where needed.

tsdb/importer/blocks/multi.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+type errAppender struct{ err error }
+
+func (a errAppender) Add(l labels.Labels, t int64, v float64) (uint64, error) { return 0, a.err }
+func (a errAppender) AddFast(ref uint64, t int64, v float64) error            { return a.err }
+func (a errAppender) Commit() error                                           { return a.err }
+func (a errAppender) Rollback() error                                         { return a.err }
+
+func rangeForTimestamp(t int64, width int64) (maxt int64) {
```
brian-brazil, line 33: This doesn't seem like code we should be duplicating.

importers/rules.go
```diff
@@ -0,0 +1,256 @@
+// Copyright 2020 The Prometheus Authors
```
brian-brazil, line 1: I think this promtool-specific code should go beside promtool, not clutter up the top-level directory.

importers/rules.go
```diff
@@ -0,0 +1,256 @@
...
+	if err != nil {
+		return err
+	}
+	url, err := url.Parse(importer.config.URL)
+	if err != nil {
+		return err
+	}
+
+	appender := importer.writer.Appender()
+	for ts.Before(end) {
```
brian-brazil, line 219: You're still using the user-specified time for alignment, which won't work out.

importers/rules.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	err := importer.CreateBlocks()
+	if err != nil {
+		errs = append(errs, err)
+	}
+	return errs
+}
+
+func (importer *RuleImporter) queryFn(ctx context.Context, q string, t time.Time) (promql.Vector, error) {
+	val, warnings, err := importer.apiClient.Query(ctx, q, t)
```
brian-brazil, line 158: Why is this using query rather than query range?

importers/rules.go
```diff
@@ -0,0 +1,256 @@
...
+	if err != nil {
+		return promql.Vector{}, err
+	}
+	if warnings != nil {
+		fmt.Fprint(os.Stderr, "warning api.Query:", warnings)
```
brian-brazil, line 163: Libraries shouldn't print.

importers/rules.go
```diff
@@ -0,0 +1,256 @@
...
+		return err
+	}
+	url, err := url.Parse(importer.config.URL)
+	if err != nil {
+		return err
+	}
+
+	appender := importer.writer.Appender()
+	for ts.Before(end) {
+		vector, err := rule.Eval(ctx, ts, importer.queryFn, url)
```
brian-brazil, line 220: You shouldn't be evaluating the rule, leave that to the API. Once the API has given us the data, ingest it.

tsdb/importer/blocks/writer.go
bwplotka: It's copied from CSV work which we have.

importers/rules.go
bwplotka: Yes, we just talked about this.

The idea would be starting from zero if no data is yet in Prom
But actually we could use instead of 0, a group hash as in Prometheus. The tricky part is a filename, as it obviously will be different.

```
// evalTimestamp returns the immediately preceding consistently slotted evaluation time.
func (g *Group) evalTimestamp() time.Time {
	var (
		offset = int64(g.hash() % uint64(g.interval))
		now    = time.Now().UnixNano()
		adjNow = now - offset
		base   = adjNow - (adjNow % int64(g.interval))
	)

	return time.Unix(0, base+offset).UTC()
}
```
If the data exists we can fetch samples from Prometheus and aligned (:

importers/rules.go
bwplotka: Query range does not stream the response.

brian-brazil: We don't need streamed responses, at most query range will return 50M samples. We want to leave the query evaluation up to the other end, not do it inside promtool.

bwplotka: > We don't need streamed responses, at most query range will return 50M samples.

What do you mean? This is worse than lack of streaming as this means we cannot even do anything beyond 50M samples. Imagine someone backfilling 1y for single series. Although such a thing we might be happy to just reject.

Otherwise query range means less requests to the server indeed. We can definitely think about using the query range then, however, there is a nontrivial issue with `step` adjusting for query range to match the actual eval interval we want. :thinking: 

As per if I query `[0 - 100]` with step 2 will it start at 0 always?

tsdb/importer/blocks/multi.go
bwplotka: Again, this is copied from https://github.com/prometheus/prometheus/pull/7586 but let's talk about things to improve in those code as well anyway.

importers/rules.go
brian-brazil: I'm not proposing to pull in raw series and do PromQL evaluation in promtool - that'd be very inefficient. Rather get the Prometheus on the other end to do all the heavy lifting.

bwplotka:
Let us know if it's ready to review! (:

JessicaGreben:
Sorry for the slow rely to comments (had some life stuff to take care of).  I'm back at it now and aiming to have this PR ready for review in a week. Will ping you all when its ready.

cmd/promtool/main.go
JessicaGreben: what do you mean by "full stop"?

roidelapluie: The comments must ends with a full stop (a dot): .

JessicaGreben: ah thanks for the explanation, will fix

importers/rules.go
JessicaGreben: what is "the output dir"? should I take in a path as an argument where the blocks will be written?

importers/rules.go
JessicaGreben: i'm trying to de-duplicate the code and generalize the manager code so I can use it here as well, but its not going well... getting complicated. If you have any suggestions how to do it, let me know. The only thing I can come up with is that the rules.Manager might need a refactor to separate these different responsibilities.

importers/rules.go
JessicaGreben: I'm will give queryRange API and see how that works out. Pushing a commit with that change shortly

importers/rules.go
brian-brazil: Yes, wherever you're writing the output.

importers/rules.go
JessicaGreben: i moved it to cmd/promtool/rules.go

importers/rules.go
JessicaGreben: I exported the evalTimestamp on the Group and used that in my code for the alignment. Let me know if you see any glaring problems with that. I still need to test it, which I'm going to do today.

importers/rules.go
JessicaGreben: Do you have any recommendation on how I should handle the warnings?

tsdb/importer/blocks/multi.go
JessicaGreben: I copied this code over from #7586. Not sure if I should try to merge it here with this PR or instead help get 7586 merged. Im happy to do either.

JessicaGreben:
@bwplotka and @brian-brazil I made some changes per your comments, not sure if I'm handling the alignment correctly and the queryRange API right. I would appreciate another review when you all have a free moment. Thanks so much for the help!

importers/rules.go
JessicaGreben: on second thought, I wonder if I can get away without needing a lot of this logic. It seems like all I really need to know in order to create new time series for recording rules is the query expression, group interval, and the info to create the alignment. Let me try to pair this down to only the data needed.

cmd/promtool/main.go
```diff
@@ -128,6 +129,21 @@ func main() {
...
+	backfillCmd := app.Command("backfill", "Backfill Prometheus data.")
+	backfillRuleCmd := backfillCmd.Command("rules", "Backfill Prometheus data for new rules.")
+	backfillRuleStart := backfillRuleCmd.Flag("start", "The time to start backfilling the new rule from. It is required. Start time should be RFC3339 or Unix timestamp.").
+		Required().String()
+	backfillRuleEnd := backfillRuleCmd.Flag("end", "If an end time is provided, all recording rules in the rule files provided will be backfilled to the end time. Default will backfill up to 3 hrs ago. End time should be RFC3339 or Unix timestamp.").
+		Default("-3h").String()
+	backfillOutputDir := backfillRuleCmd.Flag("output dir", "The filepath on the local filesystem to write the output to. Output will be blocks containing the data of the backfilled recording rules.").Default("backfilldata/").String()
+	backfillRuleURL := backfillRuleCmd.Flag("url", "Prometheus API url with the data where the rule will be backfilled from.").Default("localhost:9090").String()
+	backfillRuleEvalInterval := backfillRuleCmd.Flag("evaluation_interval", "How frequently to evaluate rules when backfilling.").
+		Default("15s").Duration()
```
brian-brazil, line 141: The default in Prometheus is 60s, let's not have two different defaults in different places.

cmd/promtool/main.go
```diff
@@ -128,6 +129,21 @@ func main() {
...
 
+	backfillCmd := app.Command("backfill", "Backfill Prometheus data.")
+	backfillRuleCmd := backfillCmd.Command("rules", "Backfill Prometheus data for new rules.")
+	backfillRuleStart := backfillRuleCmd.Flag("start", "The time to start backfilling the new rule from. It is required. Start time should be RFC3339 or Unix timestamp.").
+		Required().String()
+	backfillRuleEnd := backfillRuleCmd.Flag("end", "If an end time is provided, all recording rules in the rule files provided will be backfilled to the end time. Default will backfill up to 3 hrs ago. End time should be RFC3339 or Unix timestamp.").
+		Default("-3h").String()
+	backfillOutputDir := backfillRuleCmd.Flag("output dir", "The filepath on the local filesystem to write the output to. Output will be blocks containing the data of the backfilled recording rules.").Default("backfilldata/").String()
+	backfillRuleURL := backfillRuleCmd.Flag("url", "Prometheus API url with the data where the rule will be backfilled from.").Default("localhost:9090").String()
+	backfillRuleEvalInterval := backfillRuleCmd.Flag("evaluation_interval", "How frequently to evaluate rules when backfilling.").
```
brian-brazil, line 140: Is this only a default, or does it override what's in the rule file itself?

cmd/promtool/main.go
```diff
@@ -747,3 +766,45 @@ func (j *jsonPrinter) printLabelValues(v model.LabelValues) {
...
+		if err != nil {
+			fmt.Fprintln(os.Stderr, "rule importer parse error", err)
+			return 1
+		}
+	}
+
+	errs = ruleImporter.ImportAll(ctx)
+	for _, err := range errs {
+		if err != nil {
+			fmt.Fprintln(os.Stderr, "rule importer error", err)
```
brian-brazil, line 805: Why doesn't this return with a non-zero exit code?

rules/manager.go
```diff
@@ -471,19 +473,19 @@ func (g *Group) GetEvaluationTimestamp() time.Time {
...
 
-// evalTimestamp returns the immediately preceding consistently slotted evaluation time.
-func (g *Group) evalTimestamp() time.Time {
+// EvalTimestamp returns the immediately preceding consistently slotted evaluation time.
+func (g *Group) EvalTimestamp(startTime int64) time.Time {
 	var (
 		offset = int64(g.hash() % uint64(g.interval))
-		now    = time.Now().UnixNano()
-		adjNow = now - offset
+		start  = startTime
```
brian-brazil, line 487: You can get rid of this variable.

tsdb/importer/blocks/multi.go
brian-brazil: No matter the PR, this code shouldn't be duplicated. So ultimately it sounds like it needs to be fixed in both PRs.

tsdb/importer/blocks/writer.go
brian-brazil: This comment stands, this entire file seems like an unnecessary abstraction.

cmd/promtool/rules.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/api"
+	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/rules"
+	"github.com/prometheus/prometheus/tsdb/importer/blocks"
+)
+
+const blockSize = 2 // in hours
```
brian-brazil, line 32: We should be using milliseconds everywhere. This should also be coming from a flag.

cmd/promtool/rules.go
```diff
@@ -0,0 +1,208 @@
...
+		groupLoader: rules.FileLoader{},
+	}
+}
+
+// Init initializes the rule importer which creates a new block writer
+// and creates an Prometheus API client.
+func (importer *RuleImporter) Init() error {
+	importer.writer = blocks.NewMultiWriter(importer.logger,
+		importer.config.OutputDir,
+		importer.config.EvalInterval.Nanoseconds(),
```
brian-brazil, line 69: This doesn't seem right, why would a block writer care about the eval interval?

cmd/promtool/rules.go
```diff
@@ -0,0 +1,208 @@
...
+	return nil
+}
+
+// Close cleans up any open resources.
+func (importer *RuleImporter) Close() error {
+	return importer.writer.Close()
+}
+
+// LoadGroups reads groups from a list of rule files.
+func (importer *RuleImporter) LoadGroups(ctx context.Context, filenames []string) (errs []error) {
```
brian-brazil, line 89: This duplicate code is still here, let's not have logic like this in two places as it'll get out of sync.

cmd/promtool/rules.go
```diff
@@ -0,0 +1,208 @@
...
+	return errs
+}
+
+// ImportRule imports the historical data for a single rule.
+func (importer *RuleImporter) ImportRule(ctx context.Context, ruleExpr string, stimeWithAlignment time.Time, internval time.Duration) error {
+	ts := stimeWithAlignment
+	appender := importer.writer.Appender()
+
+	for ts.Before(importer.config.End) {
+		currentBlockEnd := ts.Add(blockSize * time.Hour)
```
brian-brazil, line 157: Couldn't this produce a timestamp that's beyond the end of the block?

cmd/promtool/rules.go
```diff
@@ -0,0 +1,208 @@
...
+
+// ImportAll evaluates all the groups and rules and creates new time series
+// and stores them in new blocks.
+func (importer *RuleImporter) ImportAll(ctx context.Context) []error {
+	var errs = []error{}
+	for _, group := range importer.groups {
+		stimeWithAlignment := group.EvalTimestamp(importer.config.Start.UnixNano())
+
+		for _, r := range group.Rules() {
+			err := importer.ImportRule(ctx, r.Query().String(), stimeWithAlignment, group.Interval())
```
brian-brazil, line 138: I think we should be going block by block, rather than rule by rule. It should be more efficient.

cmd/promtool/rules.go
```diff
@@ -0,0 +1,208 @@
...
+				for k, v := range sample.Metric {
+					currentLabels = append(currentLabels, labels.Label{
+						Name:  string(k),
+						Value: string(v),
+					})
+				}
+				for _, value := range sample.Values {
+					_, err := appender.Add(currentLabels, value.Timestamp.Unix(), float64(value.Value))
+					if err != nil {
+						// todo: handle other errors, i.e. ErrOutOfOrderSample and ErrDuplicateSampleForTimestamp
```
brian-brazil, line 192: I wouldn't worry about these, as that'd imply clashing rules.

cmd/promtool/rules.go
```diff
@@ -0,0 +1,208 @@
...
+		if err != nil {
+			return err
+		}
+		if warnings != nil {
+			fmt.Fprint(os.Stderr, "warning api.QueryRange:", warnings)
+		}
+
+		var matrix model.Matrix
+		switch val.Type() {
+		case model.ValMatrix:
```
brian-brazil, line 179: You don't need a switch when you only care about one type:

    matrix, ok := val.(model.ValMatrix)

cmd/promtool/rules.go
```diff
@@ -0,0 +1,208 @@
...
+						// todo: handle other errors, i.e. ErrOutOfOrderSample and ErrDuplicateSampleForTimestamp
+						return err
+					}
+				}
+			}
+		default:
+			return errors.New("rule result is wrong type")
+		}
+
+		ts = currentBlockEnd
```
brian-brazil, line 201: can't this result in a time being calculated twice if the alignment is 0?

brian-brazil (commented):
Overall approach seems okay, but some details need improvement.

rules/manager.go
JessicaGreben: why?

brian-brazil: You only use it once.

JessicaGreben: oh I thought you meant to remove the argument. But yeah let me remove this indirection, good catch

cmd/promtool/main.go
JessicaGreben: no it doesn't override the rule. The rule interval will take precedence over this flag. I will add those details.

cmd/promtool/main.go
JessicaGreben: I will change to 60s here. Thank you

cmd/promtool/main.go
JessicaGreben: will do.

cmd/promtool/rules.go
JessicaGreben: i will change to milliseconds. Can we just hardcode 2 hr blocks for now and make it a flag another day if needed?

cmd/promtool/rules.go
JessicaGreben: oops my mistake. I'm removing this multiwriter anyhow.

cmd/promtool/rules.go
JessicaGreben: You left this same comment before and I replied to it here 
https://github.com/prometheus/prometheus/pull/7675#discussion_r487461258

cmd/promtool/rules.go
JessicaGreben: good idea. Let me change that

cmd/promtool/rules.go
JessicaGreben: yeah, let me add a check that fixes that

cmd/promtool/rules.go
JessicaGreben: hmm good thought. let me think about that and try to fix.

brian-brazil:
Did you forget to push?

> Can we just hardcode 2 hr blocks for now and make it a flag another day if needed?

It should be a flag from the start I think, so you can pass in a smaller value if you've an expensive query.

JessicaGreben:
I will add a flag, I will push a change in a bit.

codesome:
I have not looked at the implementation yet, but wanted to say that as with the normal backfill, we need to align the block boundaries of new blocks being created with existing blocks to avoid formation of a single huge block (vertical compaction merges all overlapping block into one).

cmd/promtool/main.go
```diff
@@ -747,3 +766,45 @@ func (j *jsonPrinter) printLabelValues(v model.LabelValues) {
...
+	cfg := RuleImporterConfig{
+		Start:        stime,
+		End:          etime,
+		OutputDir:    outputDir,
+		EvalInterval: evalInterval,
+		URL:          url,
+	}
+	logger := log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
+	ruleImporter := NewRuleImporter(logger, cfg)
+	err = ruleImporter.Init()
```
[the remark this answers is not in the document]
bwplotka: same everywhere (:

cmd/promtool/main.go
bwplotka: See my comment above

> It would be bit easier to return error here, but not a blocker.

To avoid this mistake in future.

cmd/promtool/rules.go
bwplotka: Sure, but in miliseconds would be better indeed

bwplotka (approved):
Some initial review.

bwplotka:
Ups, misclick I did not plan to put green mark yet ;p But very close!

cmd/promtool/rules.go
```diff
@@ -0,0 +1,208 @@
...
+	groups      map[string]*rules.Group
+	groupLoader rules.GroupLoader
+
+	apiClient v1.API
+
+	writer *blocks.MultiWriter
+}
+
+// RuleImporterConfig is the config for the rule importer.
+type RuleImporterConfig struct {
```
bwplotka, line 48: No need to make all of this public. We cannot import `main` package anyway.

cmd/promtool/rules.go
bwplotka: Do we mean having single rule (series) per block vs multiple per block? Let's dive into this more before deciding, I am not sure if this is more efficient. Spamming one-series blocks vs having one block with X series for X rules seems not very efficial me, but both of this will "work" for now. Why single series per block would be more efficient?

cmd/promtool/rules.go
bwplotka: But it's harder to print what it is if not `model.ValMatrix`. Happy with switch as well I think (:

cmd/promtool/main.go
```diff
@@ -747,3 +766,45 @@ func (j *jsonPrinter) printLabelValues(v model.LabelValues) {
 	//nolint:errcheck
 	json.NewEncoder(os.Stdout).Encode(v)
 }
+
+// BackfillRule backfills recording rules from the files provided. The output are blocks of data
+// at the outputDir location.
+func BackfillRule(url, start, end, outputDir string, evalInterval time.Duration, files ...string) int {
```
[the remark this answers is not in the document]
JessicaGreben: good idea. will do

rules/manager.go
```diff
@@ -278,9 +280,13 @@ func NewGroup(o GroupOptions) *Group {
 		metrics = NewGroupMetrics(o.Opts.Registerer)
 	}
 
+<<<<<<< HEAD
+	key := GroupKey(o.File, o.Name)
+=======
 	key := groupKey(o.File, o.Name)
 	metrics.iterationsMissed.WithLabelValues(key)
 	metrics.iterationsScheduled.WithLabelValues(key)
+>>>>>>> master
```
JessicaGreben, line 289: oops, let me fix

cmd/promtool/rules.go
JessicaGreben: i will leave as switch for now.

beorn7:
I collected a few thoughts about the command structure [here](https://github.com/prometheus/prometheus/pull/8084#issuecomment-720660677), roughly continuing where we just ended the sync meeting.

importers/rules.go
JessicaGreben: I logged the warnings instead.

cmd/promtool/rules.go
```diff
@@ -0,0 +1,197 @@
...
+
+				currentLabels = append(currentLabels, ruleLabels...)
+
+				for name, value := range sample.Metric {
+					currentLabels = append(currentLabels, labels.Label{
+						Name:  string(name),
+						Value: string(value),
+					})
+				}
+				for _, value := range sample.Values {
```
brian-brazil, line 141: It just occurred to me that we need to insert stale markers here when there's a sample missing for a given evaluation time, but the previous evaluation time had a sample.

JessicaGreben: oh interesting. Let me look into adding that.

bwplotka:
This was close to being merged, are you around @JessicaGreben to merge this in? (:

JessicaGreben:
Its not quite ready. Needs a little more testing, but almost there. I have been busy with other things, but I can get back on it this week.

cmd/promtool/main.go
```diff
@@ -146,6 +148,18 @@ func main() {
 	// TODO(aSquare14): add flag to set default block duration
 	importFilePath := openMetricsImportCmd.Arg("input file", "OpenMetrics file to read samples from.").Required().String()
 	importDBPath := openMetricsImportCmd.Arg("output directory", "Output directory for generated blocks.").Default(defaultDBPath).String()
+	importRulesCmd := importCmd.Command("rules", "Create new blocks of data from Prometheus data for new rules from recording rule files.")
+	importRulesStart := importRulesCmd.Flag("start", "The time to start backfilling the new rule from. It is required. Start time should be RFC3339 or Unix timestamp.").
+		Required().String()
+	importRulesEnd := importRulesCmd.Flag("end", "If an end time is provided, all recording rules in the rule files provided will be backfilled to the end time. Default will backfill up to 3 hrs ago. End time should be RFC3339 or Unix timestamp.").String()
```
roidelapluie, line 154:
```suggestion
	importRulesEnd := importRulesCmd.Flag("end", "If an end time is provided, all recording rules in the rule files provided will be backfilled to the end time. Default will backfill up to 3 hours ago. End time should be RFC3339 or Unix timestamp.").String()
```

roidelapluie (commented):
We have looked at this pull request during our contributor hour. The pull request is in a good shape and would be ready to be merged after we aligned blocks capped at the 2hrs mark.

We have agreed that staleness markers would be handled later, in a future pull request.

cmd/promtool/rules.go
```diff
@@ -0,0 +1,217 @@
...
+	blockDuration := tsdb.DefaultBlockDuration
+	startOfBlock := blockDuration * (start.Unix() / blockDuration)
+	for t := startOfBlock; t <= end.Unix(); t = t + blockDuration {
+		endOfBlock := t + blockDuration
+
+		val, warnings, err := importer.apiClient.QueryRange(ctx,
+			ruleExpr,
+			v1.Range{
+				Start: time.Unix(t, 0),
+				End:   time.Unix(endOfBlock, 0),
```
roidelapluie, line 106: This should be `min ( end , endOfBlock)`

cmd/promtool/rules.go
```diff
@@ -0,0 +1,217 @@
...
+func (importer *ruleImporter) importRule(ctx context.Context, ruleExpr, ruleName string, ruleLabels labels.Labels, start, end time.Time, interval time.Duration) (err error) {
+	blockDuration := tsdb.DefaultBlockDuration
+	startOfBlock := blockDuration * (start.Unix() / blockDuration)
+	for t := startOfBlock; t <= end.Unix(); t = t + blockDuration {
+		endOfBlock := t + blockDuration
+
+		val, warnings, err := importer.apiClient.QueryRange(ctx,
+			ruleExpr,
+			v1.Range{
+				Start: time.Unix(t, 0),
```
roidelapluie, line 105: This should be `max ( start , startOfBlock)`

cmd/promtool/rules.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	return errs
+}
+
+// importRule queries a prometheus API to evaluate rules at times in the past.
+func (importer *ruleImporter) importRule(ctx context.Context, ruleExpr, ruleName string, ruleLabels labels.Labels, start, end time.Time, interval time.Duration) (err error) {
+	blockDuration := tsdb.DefaultBlockDuration
+	startOfBlock := blockDuration * (start.Unix() / blockDuration)
+	for t := startOfBlock; t <= end.Unix(); t = t + blockDuration {
+		endOfBlock := t + blockDuration
```
roidelapluie, line 100: We probably want to calculate startOfBlock / start, and instead of executing a rule from the start of a block, delay it so we even even space results.

cmd/promtool/main.go
```diff
@@ -146,6 +147,18 @@ func main() {
...
 	importDBPath := openMetricsImportCmd.Arg("output directory", "Output directory for generated blocks.").Default(defaultDBPath).String()
+	importRulesCmd := importCmd.Command("rules", "Create new blocks of data from Prometheus data for new rules from recording rule files.")
+	importRulesStart := importRulesCmd.Flag("start", "The time to start backfilling the new rule from. It is required. Start time should be RFC3339 or Unix timestamp.").
+		Required().String()
+	importRulesEnd := importRulesCmd.Flag("end", "If an end time is provided, all recording rules in the rule files provided will be backfilled to the end time. Default will backfill up to 3 hours ago. End time should be RFC3339 or Unix timestamp.").String()
+	importRulesOutputDir := importRulesCmd.Flag("output-dir", "The filepath on the local filesystem to write the output to. Output will be blocks containing the data of the backfilled recording rules. Don't use an active Prometheus data directory. If command is run many times with same start/end time, it will create duplicate series.").Default("backfilldata/").String()
```
roidelapluie, line 154:
```suggestion
	importRulesOutputDir := importRulesCmd.Flag("output-dir", "Output directory for generated blocks.").Default(defaultDBPath).String()
```

cmd/promtool/main.go
```diff
@@ -146,6 +147,18 @@ func main() {
 	// TODO(aSquare14): add flag to set default block duration
 	importFilePath := openMetricsImportCmd.Arg("input file", "OpenMetrics file to read samples from.").Required().String()
 	importDBPath := openMetricsImportCmd.Arg("output directory", "Output directory for generated blocks.").Default(defaultDBPath).String()
+	importRulesCmd := importCmd.Command("rules", "Create new blocks of data from Prometheus data for new rules from recording rule files.")
+	importRulesStart := importRulesCmd.Flag("start", "The time to start backfilling the new rule from. It is required. Start time should be RFC3339 or Unix timestamp.").
```
roidelapluie, line 151:
```suggestion
	importRulesStart := importRulesCmd.Flag("start", "The time to start backfilling the new rule from. Must be a RFC3339 formated date or Unix timestamp. Required.").
```

cmd/promtool/main.go
roidelapluie: Those should go to the "storage" page in the documentation. We can do it in a second PR.

cmd/promtool/main.go
```diff
@@ -146,6 +147,18 @@ func main() {
...
+		Required().String()
+	importRulesEnd := importRulesCmd.Flag("end", "If an end time is provided, all recording rules in the rule files provided will be backfilled to the end time. Default will backfill up to 3 hours ago. End time should be RFC3339 or Unix timestamp.").String()
```
roidelapluie, line 153:
```suggestion
	importRulesEnd := importRulesCmd.Flag("end", "If an end time is provided, all recording rules in the rule files provided will be backfilled to the end time. Default will backfill up to 3 hours ago. Must be a RFC3339 formated date or Unix timestamp.").String()
```

cmd/promtool/main.go
```diff
@@ -146,6 +147,18 @@ func main() {
...
+	importRulesOutputDir := importRulesCmd.Flag("output-dir", "The filepath on the local filesystem to write the output to. Output will be blocks containing the data of the backfilled recording rules. Don't use an active Prometheus data directory. If command is run many times with same start/end time, it will create duplicate series.").Default("backfilldata/").String()
+	importRulesURL := importRulesCmd.Flag("url", "The URL for the Prometheus API with the data where the rule will be backfilled from.").Default("http://localhost:9090").URL()
```
roidelapluie, line 155: nit: we could have this as the first argument, like we do for the promtool query.

cmd/promtool/main.go
```diff
@@ -146,6 +147,18 @@ func main() {
...
+	importRulesEvalInterval := importRulesCmd.Flag("eval-interval", "How frequently to evaluate rules when backfilling if a value is not set in the recording rule files.").
+		Default("60s").Duration()
+	importRulesFiles := importRulesCmd.Arg(
+		"rule-files",
+		"A list of one or more files containing recording rules to be backfilled. All recording rules listed in the files will be backfilled. Alerting rules are not evaluated.",
```
roidelapluie, line 160:
```suggestion
		"A list of one or more files containing recording rules to be backfilled. Please refer to the storage docs for more details.",
```

cmd/promtool/rules.go
```diff
@@ -0,0 +1,234 @@
...
+				Start: startWithAlignment,
+				End:   time.Unix(min(endOfBlock/int64(time.Second/time.Millisecond), end.Unix()), 0),
+				Step:  grp.Interval(),
+			},
+		)
+		if err != nil {
+			return errors.Wrap(err, "query range")
+		}
+		if warnings != nil {
+			level.Warn(importer.logger).Log("backfiller", fmt.Sprintf("warnings QueryRange api: %v", warnings))
```
roidelapluie, line 118:
```suggestion
			level.Warn(importer.logger).Log("msg", "Range query returned warnings.", "warnings", warnings))
```

cmd/promtool/rules.go
```diff
@@ -0,0 +1,234 @@
...
+		return errs
+	}
+	importer.groups = groups
+	return nil
+}
+
+// importAll evaluates all the recording rules and creates new time series and writes them to disk in blocks.
+func (importer *ruleImporter) importAll(ctx context.Context) (errs []error) {
+	for name, group := range importer.groups {
+		level.Info(importer.logger).Log("backfiller", fmt.Sprintf("processing group, name: %s", name))
```
roidelapluie, line 81:
```suggestion
		level.Info(importer.logger).Log("backfiller", "processing group", "name", name))
```

cmd/promtool/rules.go
```diff
@@ -0,0 +1,234 @@
...
+
+		stimeWithAlignment := group.EvalTimestamp(importer.config.start.UnixNano())
+		for i, r := range group.Rules() {
+			level.Info(importer.logger).Log("backfiller", fmt.Sprintf("processing rule %d, name: %s", i+1, r.Name()))
```
roidelapluie, line 85:
```suggestion
			level.Info(importer.logger).Log("backfiller", "processing rule", "id", i, "name", r.Name()))
```

roidelapluie (commented):
Thank you again for your work!

I don't think we should have all those details in the command line. Let's instead work on documentation in the storage part in another pull request.

I have added a couple of small comments, then I think we are ready to go.

cmd/promtool/main.go
JessicaGreben: I would prefer to make this change once we add the docs.

tsdb/blockwriter.go
```diff
@@ -90,16 +89,10 @@ func (w *BlockWriter) Appender(ctx context.Context) storage.Appender {
 // Flush implements the Writer interface. This is where actual block writing
 // happens. After flush completes, no writes can be done.
 func (w *BlockWriter) Flush(ctx context.Context) (ulid.ULID, error) {
-	seriesCount := w.head.NumSeries()
```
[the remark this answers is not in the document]
JessicaGreben: I was only trying to remove line 94-96 because if I backfill a recording rule and there are no series created for that rule, it errors on line 95. Plus when I flush when there are no series, it does not create a block so I'm not sure why the  `w.head.NumSeries() == 0 ` check was being made.

However I can keep line 93 and line 102 to log that info.

cmd/promtool/rules.go
```diff
@@ -0,0 +1,234 @@
...
+		// that are at most half a block size older than the most recent sample appended so far.
+		// However, in the way we use the block writer here, compaction doesn't happen, while we
+		// also need to append samples throughout the whole block range. To allow that, we
+		// pretend that the block is twice as large here, but only really add sample in the
+		// original interval later.
+		w, err := tsdb.NewBlockWriter(log.NewNopLogger(), importer.config.outputDir, 2*tsdb.DefaultBlockDuration)
+		if err != nil {
+			return errors.Wrap(err, "new block writer")
+		}
+		defer func() {
```
roidelapluie, line 130:
```suggestion
var closed bool
		defer func() {
		if !closed {
```

cmd/promtool/rules.go
```diff
@@ -0,0 +1,234 @@
...
+				for _, value := range sample.Values {
+					if err := app.add(ctx, currentLabels, timestamp.FromTime(value.Timestamp.Time()), float64(value.Value)); err != nil {
+						return errors.Wrap(err, "add")
+					}
+				}
+			}
+		default:
+			return errors.New(fmt.Sprintf("rule result is wrong type %s", val.Type().String()))
+		}
+
```
roidelapluie, line 163: err = tsdb_errors.NewMulti(err, w.Close()).Err())
closed = true

roidelapluie (commented):
Thank you.

Since the end time is dynamic, I would add a log line about the actual start and end time picked by promtool.

I also think we should try to close the blocks earlier than at the end. I have added a suggestion.

roidelapluie:
Thank you! Did you notice that the tests are failing?

JessicaGreben:
oops let me fix those failing tests

beorn7:
As per message from @roidelapluie, merging this now.

bboreham:
Wow, heroic year-long effort.  Well done!

marco-m-pix4d:
Hello @JessicaGreben, when I discovered about this feature in the Medium article you wrote, I _had_ to find this PR, to thank you. You are reducing my pain with Prometheus rules. Thanks again.

PR #7678, opened 2020-07-27
DNS SD: add srv record target and port meta labels

### LINKED ISSUE #3196: DNS discovery should not trim final dot

roidelapluie:
close #3196 

### DIFF
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -592,8 +592,8 @@ queries, but not the advanced DNS-SD approach specified in
 The following meta labels are available on targets during [relabeling](#relabel_config):
 
 * `__meta_dns_name`: the record name that produced the discovered target.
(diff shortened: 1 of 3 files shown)


docs/configuration/configuration.md
```diff
@@ -589,9 +589,11 @@ This service discovery method only supports basic DNS A, AAAA and SRV record
 queries, but not the advanced DNS-SD approach specified in
 [RFC6763](https://tools.ietf.org/html/rfc6763).
 
-During the [relabeling phase](#relabel_config), the meta label
-`__meta_dns_name` is available on each target and is set to the
-record name that produced the discovered target.
+The following meta labels are available on targets during [relabeling](#relabel_config):
+
+* `__meta_dns_name`: the record name that produced the discovered target.
+* `__meta_dns_srv_record_target`: the target field of the SRV record, if the query type is SRV
```
brian-brazil, line 595: We can probably drop everything after the ,

PR #7685, opened 2020-07-28
Fix tsdb go routine leak on close

### LINKED ISSUE #7672: CI: Go routine leak in TSDB test
```
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 15500 in state chan send, with github.com/prometheus/prometheus/tsdb.(*SegmentWAL).cut.func2 on top of the stack:
goroutine 15500 [chan send]:
github.com/prometheus/prometheus/tsdb.(*SegmentWAL).cut.func2(0xc0000be630)
	/home/circleci/project/tsdb/wal.go:599 +0xc0
created by github.com/prometheus/prometheus/tsdb.(*SegmentWAL).cut
	/home/circleci/project/tsdb/wal.go:598 +0x160
]
FAIL	github.com/prometheus/prometheus/tsdb	172.643s
```

This error happens sometimes in CI

mneverov:
Fixes #7672

Since select choose cases arbitrary it is possible that [stopc](https://github.com/prometheus/prometheus/blob/master/tsdb/wal.go#L690) will be picked up even if there is a value ready in `actorc` channel.
The fix was inspired by [wal](https://github.com/prometheus/prometheus/blob/master/tsdb/wal/wal.go#L319).

Signed-off-by: Max Neverov neverov.max@gmail.com

@codesome please take a look

### DIFF
--- a/tsdb/wal.go
+++ b/tsdb/wal.go
@@ -566,24 +566,22 @@ func (w *SegmentWAL) cut() error {
 		}
 		// Finish last segment asynchronously to not block the WAL moving along
 		// in the new segment.
-		go func() {
-			w.actorc <- func() error {
-				off, err := hf.Seek(0, io.SeekCurrent)
-				if err != nil {
-					return errors.Wrapf(err, "finish old segment %s", hf.Name())
-				}
-				if err := hf.Truncate(off); err != nil {
-					return errors.Wrapf(err, "finish old segment %s", hf.Name())
-				}
-				if err := hf.Sync(); err != nil {
-					return errors.Wrapf(err, "finish old segment %s", hf.Name())
-				}
-				if err := hf.Close(); err != nil {
-					return errors.Wrapf(err, "finish old segment %s", hf.Name())
-				}
-				return nil
+		w.actorc <- func() error {
+			off, err := hf.Seek(0, io.SeekCurrent)
+			if err != nil {
+				return errors.Wrapf(err, "finish old segment %s", hf.Name())
 			}
-		}()
+			if err := hf.Truncate(off); err != nil {
+				return errors.Wrapf(err, "finish old segment %s", hf.Name())
+			}
+			if err := hf.Sync(); err != nil {
+				return errors.Wrapf(err, "finish old segment %s", hf.Name())
+			}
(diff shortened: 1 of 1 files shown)


mneverov:
@krasi-georgiev @bwplotka could you please take a look on this one?

tsdb/wal.go
```diff
@@ -699,6 +687,12 @@ func (w *SegmentWAL) run(interval time.Duration) {
 			}
 		}
 	}
+	// Drain and process any remaining functions.
+	for f := range w.actorc {
```
bwplotka, line 691: We can defer this so we can return on `679` line.

roidelapluie:
goleak is on all the tsdb tests by default, that is what catched this.

tsdb/wal.go
mneverov: nice catch, thanks

mneverov:
right, goleak is enabled via [test main](https://github.com/prometheus/prometheus/blob/master/tsdb/wal/wal_test.go#L34).

roidelapluie:
cc @bwplotka can you please review this? thanks!

tsdb/wal.go
```diff
@@ -566,24 +566,22 @@ func (w *SegmentWAL) cut() error {
...
-					return errors.Wrapf(err, "finish old segment %s", hf.Name())
-				}
-				if err := hf.Sync(); err != nil {
-					return errors.Wrapf(err, "finish old segment %s", hf.Name())
-				}
-				if err := hf.Close(); err != nil {
-					return errors.Wrapf(err, "finish old segment %s", hf.Name())
-				}
-				return nil
+		w.actorc <- func() error {
```
bwplotka, line 569: Ok, I think this will probably solve leak on close, but also this code can block easily. Think about the case when `w.actorc` is full.
 

To know for sure we might want to run some benchmarks. I think it's not the best idea to kill performance potentially only to fix close leak (when we close just once per Prometheus life) ):

bwplotka (changes requested):
I think this needs still more thinking

codesome:
Superseded by https://github.com/prometheus/prometheus/pull/10056

Issue #7506, opened 2020-07-02
Prometheus losing data with 'unexpected non-zero byte in page term bytes`


jakubgs:
# Issue
My container started showing these errors:
```
msg="error tailing WAL" err="readCheckpoint: readSegment: unexpected non-zero byte in page term bytes"
```
And it appears I'm losing metrics. There's a growing gap in metrics visible.

Based on what I read in https://github.com/prometheus/prometheus/issues/7299 and my previous experiences my understanding is that I need to remove the corrupted WAL file and all the ones after it. But the logs don't tell me which one is corrupted. What am I supposed to do?

# Info
Using Docker image `prom/prometheus:v2.19.2` on Ubuntu 18.04 LTS.

# Config
```yaml
global:
  scrape_interval:     30s
  scrape_timeout:      20s
  evaluation_interval: 30s
  external_labels:
    source: my-metrics-master

scrape_configs:
  - job_name: 'prometheus'
    static_configs:
      - targets: ['localhost:9090']

  - job_name: 'federate'
    scrape_interval: 15s
    honor_labels: true
    metrics_path: '/federate'
    static_configs:
      - targets:
          - '10.1.1.1:1234'
    params:
      'match[]':
        - '{job=~".+"}'
```

# Logs
```
caller=dedupe.go:112 component=remote level=info remote_name=69e3e5 url= msg="Replaying WAL" queue=69e3e5
caller=dedupe.go:112 component=remote level=error remote_name=69e3e5 url= msg="error tailing WAL" err="readCheckpoint: readSegment: unexpected non-zero byte in page term bytes"
caller=dedupe.go:112 component=remote level=info remote_name=69e3e5 url= msg="Replaying WAL" queue=69e3e5
caller=dedupe.go:112 component=remote level=error remote_name=69e3e5 url= msg="error tailing WAL" err="readCheckpoint: readSegment: unexpected non-zero byte in page term bytes"
caller=dedupe.go:112 component=remote level=info remote_name=69e3e5 url= msg="Replaying WAL" queue=69e3e5
caller=dedupe.go:112 component=remote level=error remote_name=69e3e5 url= msg="error tailing WAL" err="readCheckpoint: readSegment: unexpected non-zero byte in page term bytes"
```

jakubgs:
Oh, there is something that shows up from time to time in the logs though:
```
[...]

PR #7687, opened 2020-07-29
Chore: Log segment number when segment read failed

gastonqiu:
To manually fix the WAL files, it is good to know where the corrupt
happened so we should log the segment number when the read failed.

Related Issue #7506

### DIFF
--- a/tsdb/wal/watcher.go
+++ b/tsdb/wal/watcher.go
@@ -391,7 +391,7 @@ func (w *Watcher) watch(segmentNum int, tail bool) error {
 
 			// Ignore errors reading to end of segment whilst replaying the WAL.
 			if !tail {
-				if err != nil && err != io.EOF {
+				if err != nil && errors.Cause(err) != io.EOF {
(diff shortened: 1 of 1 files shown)


roidelapluie:
Should'nt we return that with the function and a fmt.Errorf("segment %d: %w", err) ?

gastonqiu:
You are right. It's better. 
> Should'nt we return that with the function and a fmt.Errorf("segment %d: %w", err) ?

cstyan:
Julien is right, we should be adding this in via `errors.Wrap(...)`. The function calling `watcher.Run()` does nothing more than log the error and then call `Run` again, it doesn't care if the error is some assertable type.

gastonqiu:
Will change it need some time.

gastonqiu:
@cstyan @roidelapluie I change the code according to the review. Can you help me to review again? Hope I don't misunderstand you guys.

gastonqiu:
@cstyan Are there any other reviewers can help on review this PR or it is fine to merge now?

PR #7689, opened 2020-07-29
Export RecoverableError

joe-elliott:
This PR simply exports RecoverableError from the remote package.  This allows callers of the exported method `WriteClient.Store` method to distinguish between recoverable and not recoverable errors.


joe-elliott:
Thanks for the quick turnaround!

PR #7695, opened 2020-07-29
Move promtool config/rules checks to their own package

cstyan:
This PR moves promtool config/rules files to their own package so they can be used by other projects programatically. For example, it would be useful if the Cortex Ruler API (which allows users to upload rules files) could run the `CheckRules` functionality and reject rules files with errors in them. cc @gotjosh 

### DIFF
--- a/util/verifyconfig/verifyconfig.go
+++ b/util/verifyconfig/verifyconfig.go
@@ -1,0 +1,232 @@
+// Copyright 2020 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


util/verifyconfig/verifyconfig.go
```diff
@@ -0,0 +1,232 @@
...
+}
+
+// CheckConfig validates configuration files.
+func CheckConfig(files ...string) int {
+	failed := false
+
+	for _, f := range files {
+		ruleFiles, err := checkConfig(f)
+		if err != nil {
+			fmt.Fprintln(os.Stderr, "  FAILED:", err)
```
brian-brazil, line 98: Libraries should not print.

cstyan:
> This seems like it should be factored to actually be a library, and some of this should probably live in the respective packages rather than as separate duplicate config handling logic.

That makes sense to me, but it's not clear where you want this code to live from your comment. In the config package? In the rules package?

brian-brazil:
It's going to depend on the code. For something like checking that files exist, it doesn't feel right that we're duplicating the mention of all the fields into another package (and I'm pretty sure that has lead to bugs).

roidelapluie:
We are closing this as part of the bug scrub. Thanks!

PR #7655, opened 2020-07-24
tsdb/wal: fix bug of goroutine leak.

johncming:
Fix the bug of the goroutine leak, which is below.

The goroutine which may be a leak
```
func (w *WAL) run() {
Loop:
	for {
		select {
		case f := <-w.actorc:
			f()
		case donec := <-w.stopc:      **<= waiting for a signal to break loop** 
			close(w.actorc)
			defer close(donec)
			break Loop
		}
	}
	// Drain and process any remaining functions.
	for f := range w.actorc {
		f()
	}
```

The signal may not be received if it fails to flush the page of the current segment.
```
	if w.page.alloc > 0 {
		if err := w.flushPage(true); err != nil {
			return err
		}
	}

	donec := make(chan struct{})
	w.stopc <- donec
	<-donec
```

These low-level IO errors are not common, and if it happened, it would cause subsequent `fsync` of the current segment skipped. So it's better to `fsync` the current segment firstly and then return the error.

PR #7697, opened 2020-07-30
tsdb: Eliminate the goroutine leak calling tests.

johncming:
A tool imported into repo to find possible goroutine leak https://github.com/prometheus/prometheus/pull/7566.
The deprecated `wal` implementations have a situation that causes goroutine leak. It leads to test failure.

The leak bug result like
```
=== Failed
=== FAIL: tsdb  (0.00s)
PASS
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 15483 in state chan send, with github.com/prometheus/prometheus/tsdb.(*SegmentWAL).cut.func2 on top of the stack:
goroutine 15483 [chan send]:
github.com/prometheus/prometheus/tsdb.(*SegmentWAL).cut.func2(0xc0000be210)
	/home/circleci/project/tsdb/wal.go:599 +0xc0
created by github.com/prometheus/prometheus/tsdb.(*SegmentWAL).cut
	/home/circleci/project/tsdb/wal.go:598 +0x160
]
FAIL	github.com/prometheus/prometheus/tsdb	174.747s


DONE 1022 tests, 17 skipped, 1 failure in 379.805s
make: *** [Makefile.common:172: common-test] Error 1

Exited with code exit status 2
```

The PR has relation with another PR https://github.com/prometheus/prometheus/pull/7655#issuecomment-666016301

### DIFF
--- a/tsdb/wal.go
+++ b/tsdb/wal.go
@@ -564,24 +564,29 @@ func (w *SegmentWAL) cut() error {
 		if err := w.flush(); err != nil {
 			return err
 		}
+
+		actor := func() error {
+			off, err := hf.Seek(0, io.SeekCurrent)
+			if err != nil {
+				return errors.Wrapf(err, "finish old segment %s", hf.Name())
+			}
+			if err := hf.Truncate(off); err != nil {
+				return errors.Wrapf(err, "finish old segment %s", hf.Name())
+			}
+			if err := hf.Sync(); err != nil {
+				return errors.Wrapf(err, "finish old segment %s", hf.Name())
+			}
+			if err := hf.Close(); err != nil {
+				return errors.Wrapf(err, "finish old segment %s", hf.Name())
+			}
+			return nil
+		}
 		// Finish last segment asynchronously to not block the WAL moving along
 		// in the new segment.
 		go func() {
-			w.actorc <- func() error {
-				off, err := hf.Seek(0, io.SeekCurrent)
-				if err != nil {
-					return errors.Wrapf(err, "finish old segment %s", hf.Name())
-				}
-				if err := hf.Truncate(off); err != nil {
-					return errors.Wrapf(err, "finish old segment %s", hf.Name())
-				}
-				if err := hf.Sync(); err != nil {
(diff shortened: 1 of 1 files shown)


tsdb/wal.go
```diff
@@ -673,7 +673,14 @@ func (w *SegmentWAL) run(interval time.Duration) {
 		defer ticker.Stop()
 		tick = ticker.C
 	}
-	defer close(w.donec)
+	defer func() {
+		// Drain and process any remaining functions.
+		for f := range w.actorc {
+			//nolint: errcheck
+			f()
```
codesome, line 680: Let us log this error like below.

codesome (approved):
LGTM with 1 comment

tsdb/wal.go
johncming: OK. I will add it.

codesome (changes requested):
Sorry, I take my LGTM back. If it was possible to add a function to the actor after stopping (even though we are draining it already before exiting), then it is possible that function will be added to the actor even after `close(w.actorc)`, hence cause a panic. We should actually reject WAL Logging after closing, but this doesn't eliminate the mentioned race as a WAL Logging could be on-going.

I don't have a way to fix this on top of my mind.

johncming:
> Sorry, I take my LGTM back. If it was possible to add a function to the actor after stopping (even though we are draining it already before exiting), then it is possible that function will be added to the actor even after `close(w.actorc)`, hence cause a panic. We should actually reject WAL Logging after closing, but this doesn't eliminate the mentioned race as a WAL Logging could be on-going.
> 
> I don't have a way to fix this on top of my mind.

Sorry. This PR can't solve the leak.  And close the channel in receiver goroutine will lead to panic. 

Since it has been marked as deprecated, I try to eliminate leak with minimal changes.

roidelapluie:
cc @codesome can we close this pull request? Thanks!

bwplotka:
Is this still reproducible? Let's reopen and start new PR if this happens again :hugs:

PR #7700, opened 2020-07-30
Scrape: defer report

roidelapluie:
When I started wotking on target_limit, scrapeAndReport did not exist
yet. Then I simply rebased my work without thinking.

It appears that there is a lot that can be inline if I defer() the
report.


brian-brazil:
:+1:

PR #7701, opened 2020-07-30
Invert service discovery dependencies

abursavich:
WIP fix for #7107.

The goal is to make all SD implementations depend on `github.com/prometheus/prometheus/discovery` instead of `discovery` depending on all SD implementations. My intention was to do it all at once, but I can do it implementation by implementation if necessary.


roidelapluie:
> WIP fix for #7107.
> 
> The plan is to make all SD implementations depend on `github.com/prometheus/prometheus/discovery` instead of `discovery` depending on all of the implementations. My intention was to do it all at once, but I can probably do it implementation by implementation if necessary.

For each SD we will need to carefully review things, e.g. the path of the files. I'd rather work by moving 2-3 implementations in one pr.

The last discoveries we want to move are probably: consul, k8s, file.

abursavich:
There should be enough there now to see where it's going... I tried to separate things out into the smallest chunks per commit... The high-level outline of what I've done so far is something like:

- shuffle things around to prevent cyclic dependencies
- add the discoverer.Config interface
- add the Marshaling/Unmarshaling logic and use it with static_configs
- remove the dependency on openstack
- remove the dependency on azure

Note that the commits that remove a particular dependency are very small.

Future steps would be something like:

- remove the dependencies on all other discovery implementations
- remove ServiceDiscoveryConfig and replace its uses with []discoverer.Config
- shuffle things around again since cycles are gone (package discoverer gets merged into package discovery)

config/config.go
```diff
@@ -570,12 +561,9 @@ func (c *AlertmanagerConfig) UnmarshalYAML(unmarshal func(interface{}) error) er
 
 	// Check for users putting URLs in target groups.
 	if len(c.RelabelConfigs) == 0 {
-		for _, tg := range c.ServiceDiscoveryConfig.StaticConfigs {
-			for _, t := range tg.Targets {
-				if err := CheckTargetAddress(t[model.AddressLabel]); err != nil {
-					return err
-				}
-			}
+		// TODO: can this either be enforced always or never?
```
roidelapluie, line 564: Always.

discovery/azure/azure.go
```diff
@@ -64,6 +66,10 @@ var DefaultSDConfig = SDConfig{
 	AuthenticationMethod: authMethodOAuth,
 }
 
+func init() {
+	config.RegisterServiceDiscovery("azure_sd_configs", &SDConfig{})
```
roidelapluie, line 70: The suffix _sd_configs should be enforced.

Maybe the public RegisterServiceDiscovery could add it and call a non-exported registerServiceDiscovery:

RegisterServiceDiscovery(name){
registerServiceDiscovery ( name+"_sd_configs")
}

and we register the static configs with registerServiceDiscovery ("static_configs")

discovery/azure/azure_test.go
```diff
@@ -18,11 +18,11 @@ import (
 	"testing"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
-	"go.uber.org/goleak"
+	"github.com/prometheus/prometheus/util/testutil"
 )
 
 func TestMain(m *testing.M) {
-	goleak.VerifyTestMain(m)
+	testutil.TolerantVerifyLeak(m)
```
roidelapluie, line 25: why is that change needed now?

discovery/discoverer/config.go
```diff
@@ -11,12 +11,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package config
+package discoverer
+
+// NOTE: The contents of this file will end up in the config package.
```
roidelapluie, line 16: TODO:

discovery/manager_test.go
```diff
@@ -762,51 +760,110 @@ func TestTargetSetRecreatesTargetGroupsEveryRun(t *testing.T) {
...
 	discoveryManager.ApplyConfig(c)
 
 	<-discoveryManager.SyncCh()
-	verifyPresence(t, discoveryManager.targets, poolKey{setName: "prometheus", provider: "string/0"}, "{__address__=\"foo:9090\"}", true)
-	verifyPresence(t, discoveryManager.targets, poolKey{setName: "prometheus", provider: "string/0"}, "{__address__=\"bar:9090\"}", true)
+	verifyPresence(t, discoveryManager.targets, poolKey{setName: "prometheus", provider: "static/0"}, "{__address__=\"foo:9090\"}", true)
+	verifyPresence(t, discoveryManager.targets, poolKey{setName: "prometheus", provider: "static/0"}, "{__address__=\"bar:9090\"}", false)
+}
+
+type staticConfig []*targetgroup.Group
```
roidelapluie, line 816: Should we move the static sd to a new file?

roidelapluie:
This is really going to the right direction, :+1:

roidelapluie:
The error message on unknown field is now obscure:

level=error ts=2020-08-02T12:27:48.142Z caller=main.go:287 msg="Error loading config (--config.file=prometheus.yml)" err="parsing YAML file
 prometheus.yml: yaml: unmarshal errors:\n  line 15: field lazure_sd_configs not found in type struct { JobName string \"yaml:\\\"job_name\
\\"\"; HonorLabels bool \"yaml:\\\"honor_labels,omitempty\\\"\"; HonorTimestamps bool \"yaml:\\\"honor_timestamps\\\"\"; Params url.Values 
\"yaml:\\\"params,omitempty\\\"\"; ScrapeInterval model.Duration \"yaml:\\\"scrape_interval,omitempty\\\"\"; ScrapeTimeout model.Duration \
"yaml:\\\"scrape_timeout,omitempty\\\"\"; MetricsPath string \"yaml:\\\"metrics_path,omitempty\\\"\"; Scheme string \"yaml:\\\"scheme,omite
mpty\\\"\"; SampleLimit uint \"yaml:\\\"sample_limit,omitempty\\\"\"; TargetLimit uint \"yaml:\\\"target_limit,omitempty\\\"\"; ServiceDisc
overyConfig discoverer.ServiceDiscoveryConfig \"yaml:\\\",inline\\\"\"; HTTPClientConfig config.HTTPClientConfig \"yaml:\\\",inline\\\"\"; 
RelabelConfigs []*relabel.Config \"yaml:\\\"relabel_configs,omitempty\\\"\"; MetricRelabelConfigs []*relabel.Config \"yaml:\\\"metric_relab
el_configs,omitempty\\\"\"; AUTO_azure_sd_configs []*azure.SDConfig \"yaml:\\\"azure_sd_configs,omitempty\\\"\"; AUTO_openstack_sd_configs 
[]*openstack.SDConfig \"yaml:\\\"openstack_sd_configs,omitempty\\\"\"; AUTO_static_configs []*config.staticConfig \"yaml:\\\"static_configs
,omitempty\\\"\" }"

config/config.go
abursavich: Just to confirm... It's okay to push that check into the targetgroup.Group UnmarshalYAML function, so it's always applied regardless of if a RelabelConfig exists?

discovery/azure/azure.go
abursavich: SGTM

discovery/azure/azure_test.go
abursavich: I think it's only temporary... I'm not entirely sure because I didn't trace exactly which transitive dependencies are pulling in `opencensus` and `klog`, but something definitely is.

We're flipping the dependencies around, so instead of `config` and `discovery`/`discoverer` depending on `azure`, now `azure` depends on `config` and `discovery`/`discoverer`. Since the migration is incomplete, that means that `azure` is now pulling in all other non-migrated discovery implementations (including `kubernetes`, which is my guess on where we're picking up the background goroutines). Once the migration is complete `config` and `discovery` won't depend on any of the discovery implementations, so they'll be safe for each discovery implementation to depend on.

I can add a TODO to revert this when the other dependencies are removed.

roidelapluie: Oh yes I see. This now depends in k8s indirectly. Thanks!

config/config.go
roidelapluie: No, relabel must exist. But if no relabel exist, it can fail on any SD's.

abursavich: Okay. Then I need to leave it as it is. I'll remove the TODO.

roidelapluie: This is because this is **valid**
```
scrape_configs:
  - job_name: 'blackbox'
    metrics_path: /probe
    params:
      module: [http_2xx]  # Look for a HTTP 200 response.
    static_configs:
      - targets:
        - http://prometheus.io    # Target to probe with http.
        - https://prometheus.io   # Target to probe with https.
        - http://example.com:8080 # Target to probe with http on port 8080.
    relabel_configs:
      - source_labels: [__address__]
        target_label: __param_target
      - source_labels: [__param_target]
        target_label: instance
      - target_label: __address__
        replacement: 127.0.0.1:9115  # The blackbox exporter's real hostname:port.
```

But this is **invalid**
```
scrape_configs:
  - job_name: 'blackbox'
    metrics_path: /probe
    params:
      module: [http_2xx]  # Look for a HTTP 200 response.
    static_configs:
      - targets:
        - http://prometheus.io    # Target to probe with http.
        - https://prometheus.io   # Target to probe with https.
        - http://example.com:8080 # Target to probe with http on port 8080.
```

discovery/azure/azure.go
abursavich: I would like the signature to be `RegisterServiceDiscovery(config discoverer.Config)` and use the `config.Name()` to set the YAML key name (e.g. `config.Name() + "_sd_configs"`). The only issue with that is there are two discovery mechanisms with logger values that are different from their YAML key names:
- kubernetes
  - yaml: kubernetes_sd_configs
  - log: discovery=k8s
- serverset
  - yaml: serverset_sd_configs
  - log: discovery=zookeeper

Are you okay with changing the logger values to match the YAML keys?
- kubernetes
  - yaml: kubernetes_sd_configs
  - log: discovery=kubernetes
- serverset
  - yaml: serverset_sd_configs
  - log: discovery=serverset

roidelapluie: I am ok with this. cc @brancz and @brian-brazil to double validate that it's ok.

As per https://prometheus.io/docs/prometheus/latest/stability/ the format of the logs is not considered stable.

config/config.go
abursavich: Yeah, I get it... My reason for asking was because the check didn't go deep into the RelabelConfigs to see if they replace the `__address__` label. I was thinking it might not be that important if relabeling something else was sufficient to bypass the check. People creating invalid targets is probably a big enough issue that it's worth keeping.

discovery/discoverer/config.go
abursavich: Done

discovery/azure/azure.go
abursavich: I went ahead with this change for now. If it's going to be a problem, I can push the log labeling into the Config.NewDiscoverer methods and each implementation can choose its own label that may or may not be its name.

discovery/manager_test.go
abursavich: These tests should've been using `discoverer.StaticConfig`... They happened in the same commit, but I think I fixed these tests before creating the `discoverer.StaticConfig` type.

abursavich:
I've resolved everything so far except the obscure error message. I assume that's gonna require some really gross and fragile error string matching and manipulating, but I haven't focused on it yet.

I'm probably gonna hit a few more dependencies before going back to the error message.

abursavich:
This is very close to complete now... What I've done so far:

- shuffle things around to prevent cyclic dependencies
- add the discoverer.Config interface
- add the Marshaling/Unmarshaling logic and use it with static_configs
- remove each dependency one at a time
- remove ServiceDiscoveryConfig and replace its uses with []discoverer.Config
- shuffle things around again since cycles are gone, moving types in pkg discoverer back into pkg discovery

There are still a few TODOs:
- fix the obscure error message
- move a few config functions from pkg discoverer to common module
- remove pkg discoverer

discovery/azure/azure.go
brancz: perfectly happy with this as well

abursavich:
This is done now and ready for review. I don't plan to make any more changes unless they are requested.

As I said before, I tried to limit each commit to a small logical chunk but I realize that overall this is a very large change.

The sequence of commits should be mergable at any step. Let me know if you need me to split it up.

abursavich:
I do actually have another long sequence of cleanup commits ready, but they can wait until after this :)

discovery/common.go
```diff
@@ -0,0 +1,44 @@
...
+	c.KeyFile = JoinDir(dir, c.KeyFile)
+}
+
+// JoinDir returns path relative to dir. If path is empty or absolute, it is return unchanged.
+func JoinDir(dir, path string) string {
+	if path == "" || filepath.IsAbs(path) {
+		return path
+	}
+	return filepath.Join(dir, path)
+}
```
abursavich, line 44: This file shouldn't exist. It was moved to the config package, but I guess a rebase left it behind... I guess I have one more rebase to do :\

discovery/discovery.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// ConfigOptions provides options for a Config.
+type ConfigOptions struct {
+	// Directory may be used to resolve relative file paths
+	// in the config (e.g. TLS certificates).
+	Directory string
+}
+
+// A Config provides the configuration and constructor for a Discoverer.
```
roidelapluie, line 56: This seems like its doing more than Config, any name idea?

I would also add here that this is experimental and could/will change in the future; it is intended for prometheus internal use.

config/common.go
```diff
@@ -0,0 +1,57 @@
+// Copyright 2020 The Prometheus Authors
```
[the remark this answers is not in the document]
abursavich: I had written this code to live in live in the prometheus/common/config package, which is what it deals with and hence the file name, but that has been rejected so far: https://github.com/prometheus/common/pull/248

config/common.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"path/filepath"
+
+	"github.com/prometheus/common/config"
+)
+
+// JoinDir returns path relative to dir.
+// If path is empty or absolute, it is return unchanged.
+func JoinDir(dir, path string) string {
```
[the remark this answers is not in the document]
abursavich: I struggled with naming this... Technically, it doesn't ensure an absolute path because dir is allowed to be relative... and I didn't want to change the existing behavior. It joins dir to path, if path is join-able.

cmd/prometheus/main.go
```diff
@@ -67,6 +66,8 @@ import (
 	"github.com/prometheus/prometheus/tsdb"
 	"github.com/prometheus/prometheus/util/strutil"
 	"github.com/prometheus/prometheus/web"
+
+	_ "github.com/prometheus/prometheus/discovery/install"
```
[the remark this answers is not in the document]
abursavich: SGTM

cmd/promtool/main.go
```diff
@@ -263,24 +267,25 @@ func checkConfig(filename string) ([]string, error) {
...
-			}
-		}
-
-		for _, filesd := range scfg.ServiceDiscoveryConfig.FileSDConfigs {
-			for _, file := range filesd.Files {
-				files, err := filepath.Glob(file)
-				if err != nil {
+		for _, c := range scfg.ServiceDiscoveryConfigs {
+			switch c := c.(type) {
+			case *kubernetes.SDConfig:
```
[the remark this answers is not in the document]
abursavich: Yeah, I wasn't happy with this either, but again I didn't want to change the existing behavior with this PR.

I would like to hold this off for a followup, if that's alright. I think we could add a `Validate() error` method to the `discovery.Config` interface or it could be on its own and we could check here if the `discovery.Config` implements `interface { Validate() error }`?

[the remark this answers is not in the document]
abursavich: I went ahead and added `Validate() error` to the `discovery.Config` interface.

discovery/discovery.go
abursavich: I added a boilerplate experimental warning.

I'm actually OK with the name on this one. It seems fairly idiomatic for Go and reminds me of something like [net.ListenConfig](https://pkg.go.dev/net?tab=doc#ListenConfig). I'm sure in some other languages it would have a really long name like `ConfigurableServiceDiscoveryFactoryInterface` ;)

abursavich:
> Thanks a lot for this.

It was fun :)

> Could you add some docs in discovery/README.md?

Done.

> Are we testing the marshalling of the config?

We were a little bit (kinda just making sure that it didn't return an error), but I just added yaml roundtrip testing.

abursavich:
@bwplotka, this should make you happy:

`goda graph -short github.com/prometheus/prometheus/cmd/prometheus - github.com/prometheus/prometheus/discovery/install | dot -Tsvg -o graph-install.svg`

graph-install

discovery/README.md
```diff
@@ -216,15 +216,62 @@ If all the targets in a group go away, we need to send the target groups with em
 ```
 down the channel.
 
+### The Config interface
+
+Now that your service discovery mechanism is ready to discover targets, you must help
+Prometheus discover it. This is done my implementing the `discovery.Config` interface
```
roidelapluie, line 222:
```suggestion
Prometheus discover it. This is done by implementing the `discovery.Config` interface
```

discovery/README.md
```diff
@@ -216,15 +216,62 @@ If all the targets in a group go away, we need to send the target groups with em
...
+	Logger log.Logger
+}
+```
+
+The value returned by `Name()` should be short, descriptive, lowercase, and unique.
+It's used to tag the provided `Logger` and as the part of the YAML key for your SD
+mechanism's list of configs in `scrape_config` and `alertmanager_config`
+(e.g. `${NAME}_sd_configs`).
+
+`Validate()` is used by `promtool` to help discover configuration errors. It should
```
roidelapluie, line 258: should we call that outside of promtool as well then? or specify that it is expected that the SD calls it on unmarshalling? WDYT?

cmd/prometheus/main.go
```diff
@@ -67,6 +66,9 @@ import (
 	"github.com/prometheus/prometheus/tsdb"
 	"github.com/prometheus/prometheus/util/strutil"
 	"github.com/prometheus/prometheus/web"
+
+	// register service discovery implementations
```
brian-brazil, line 70: Comments should be full sentances.

config/common.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// SetHTTPClientConfigDirectory sets filepaths to be relative to dir.
+func SetHTTPClientConfigDirectory(c *config.HTTPClientConfig, dir string) {
+	if c == nil {
+		return
+	}
+	SetTLSConfigDirectory(&c.TLSConfig, dir)
+	if a := c.BasicAuth; a != nil {
+		a.PasswordFile = JoinDir(dir, a.PasswordFile)
```
brian-brazil, line 40: You're modifying the user-provided config here, which is to be avoided as that could affect rendering it back and otherwise be confusing.

config/common.go
```diff
@@ -0,0 +1,116 @@
...
+		return fmt.Errorf("bearer token cannot specify both value and file: %v", c.BearerTokenFile)
+	}
+	if err := emptyOrFileExists(c.BearerTokenFile); err != nil {
+		return fmt.Errorf("bearer token file: %v: %w", c.BearerTokenFile, err)
+	}
+	return ValidateTLSConfig(&c.TLSConfig)
+}
+
+// ValidateTLSConfig checks if the TLS config is valid.
+func ValidateTLSConfig(c *config.TLSConfig) error {
```
brian-brazil, line 81: Errors like this should be caught at unmarshall time, and should also not be the concern of this module.

@cstyan may also be interested in this.

config/common.go
```diff
@@ -0,0 +1,116 @@
...
+	if c == nil {
```
brian-brazil, line 82: This sounds like an invalid config to me.

config/common.go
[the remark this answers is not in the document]
brian-brazil: It doesn't matter where the code lives, it should never modify the struct that we unmarshall from the config file.

Also none of this munging of config files should be required in the first place, as paths are relative to the pwd.

config/config_windows_test.go
```diff
@@ -11,7 +11,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package config
+package config_test
```
brian-brazil, line 14: Why do you need this?

config/config_test.go
```diff
@@ -11,24 +11,28 @@
...
 	"io/ioutil"
 	"net/url"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
```
brian-brazil, line 27: Do we really need this dependency? It's not what we already use for comparing structs, and is only for the sake of a unittest for something we don't support.

discovery/README.md
```diff
@@ -146,85 +146,132 @@ both cases.
...
+type Config interface {
+	// Name returns the name of the discovery mechanism.
+	Name() string
+
+	// NewDiscoverer returns a Discoverer for the Config
+	// with the given DiscovererOptions.
+	NewDiscoverer(DiscovererOptions) (Discoverer, error)
+
+	// SetOptions applies the ConfigOptions to the Config.
+	SetOptions(ConfigOptions)
```
brian-brazil, line 236: Neither SetOptions not Validate should be here. We don't need that option, and any validation should be done as part of unmarshalling - not as a separate step.

discovery/README.md
brian-brazil: Yes, this should be part of unmarshalling.

discovery/install/install.go
```diff
@@ -0,0 +1,32 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package install has the side-effect of registering all builtin
+// service discovery config types.
+package install
+
+import (
+	_ "github.com/prometheus/prometheus/discovery/azure"        // register azure
```
brian-brazil, line 19: The comments aren't needed.

discovery/discovery.go
brian-brazil: That it is only for internal use is implicit for everything, no need to mention it everywhere.

This seems like a DiscoveryFactory to me.

config/discovery.go
```diff
@@ -0,0 +1,238 @@
...
+	configFieldPrefix      = "AUTO_"
+	staticConfigsKey       = "static_configs"
+	staticConfigsFieldName = configFieldPrefix + staticConfigsKey
+)
+
+var (
+	configNames      = make(map[string]discovery.Config)
+	configFieldNames = make(map[reflect.Type]string)
+	configFields     []reflect.StructField
+	configTypes      sync.Map // map[reflect.Type]reflect.Type
```
brian-brazil, line 40: We don't use sync.Maps anywhere else in Prometheus, why here?

config/discovery.go
```diff
@@ -0,0 +1,238 @@
...
+
+	emptyStructType = reflect.TypeOf(struct{}{})
+	configSliceType = reflect.TypeOf([]discovery.Config{})
+)
+
+// RegisterServiceDiscovery registers the given Config type along with the YAML key for the
+// list of its type in the Configs object.
+func RegisterServiceDiscovery(config discovery.Config) {
```
brian-brazil, line 48: These should be registering to discovery, not config.

discovery/README.md
abursavich: It might be able to share some code with unmarshalling validation (e.g. checking required or "oneof" fields), but this validation also checks for the existence of files whose paths may still be relative at unmarshalling time.

abursavich:
```go
func (c *Config) validate() error {
  // check for field validity (e.g. required, "oneof", "enums", etc.)
}

func (c *Config) Validate() error {
  if err := c.validate(); err != nil { return err }
  // check for file existence and possibly parse them for validity
}

func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
  type plain Config
  v := DefaultConfig
  if err := unmarshal(&v); err != nil { return err }
  if err := (*Config)(&v).validate(); err != nil { return err }
  *c = Config(*v)
  return nil
}
```

brian-brazil: > checks for the existence of files whose paths may still be relative at unmarshalling time.

How is that a problem? Stat works with relative paths.

config/common.go
abursavich: This is not new. I am preserving the existing behavior.

config/common.go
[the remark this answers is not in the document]
abursavich: This is not new. I am preserving the existing behavior.

config/config_test.go
abursavich: It's used to compare configs after roundtrip-ing them through yaml. This causes the secrets to be masked and the existing comparison mechanism doesn't seem to be able to take that into account.

config/config_windows_test.go
abursavich: It prevents a import cycle. The config tests import various discovery packages and those discover packages register with config.

config/discovery.go
abursavich: Because I needed a read-optimized concurrency-safe map and there's one in the standard library.

discovery/install/install.go
abursavich: They're there to make golint happy, otherwise: `a blank import should be only in a main or test package, or have a comment justifying it`

discovery/README.md
abursavich: SetOptions is there to preserve the existing behavior.

I added Validate during the review to support `promtool`, whose existing behavior was only validating kubernetes configs.

abursavich:
> I haven't looked too in depth, but the factoring doesn't seem right here. The special discovery stuff should live as much as is possible in the discovery module, and config only have just enough to use it - discovery should not depend on config.

Please read the existing discussion starting here: https://github.com/prometheus/prometheus/issues/7107#issuecomment-667573474

TL;DR: The yaml package doesn't support the combination of (a) using a yaml.Unmarshaler to decode into dynamically named fields, (b) using yaml inlining to embed that dynamic yaml.Unmarshaler into other types (e.g. ScrapeConfig and AlertmanagerConfig), and (c) using strict decoding mode. There are several options. This seems like the best available without a breaking change.

> A long term aspiration is that other software can re-use our discovery code, so things shouldn't be too tightly bound.

I don't think there's anything about builtin discovery implementations telling the config package how to (un)marshal them as YAML in its other configs that prevents people from using them. The problem is the config package pulls in every single builtin discovery implementation, which _was_ prohibitive and is what this _is_ fixing.

If it'll make you feel better, I can move the register calls to the install package so the others won't need to import config.

roidelapluie:
> If it'll make you feel better, I can move the register calls to the install package so the others won't need to import config.

Can those interfaces be part of discovery/ and config import discovery? What would block that?

abursavich:
> Can those interfaces be part of discovery and config import discovery? What would block that?

Yeah, `config` does import `discovery` and `discovery` contains all the discovery-related interfaces. The purpose of registering each implementation is only for YAML (un)marshaling of the ScrapeConfig and AlertmanagerConfig types in which they are inlined.

This registration could be moved elsewhere (e.g. `discovery`), but then it would have to export some magical yaml (un)marshaling functions. I can do that.

config/common.go
abursavich: This is not a feature that I am trying to champion right now. I just don't want to break users that are already relying on it.

roidelapluie:
FYI others are already importing us: https://github.com/grafana/loki/blob/b24a1430434770a5c4eddc19bfe8e1eddf498a15/pkg/promtail/scrapeconfig/scrapeconfig.go

discovery/README.md
abursavich: Relative to the config file.

abursavich:
> FYI others are already importing us: https://github.com/grafana/loki/blob/b24a1430434770a5c4eddc19bfe8e1eddf498a15/pkg/promtail/scrapeconfig/scrapeconfig.go

Thanks for the heads up. I'll move the registration to `discovery` and export the yaml functions needed for inlining.

config/common.go
brian-brazil: I think it was never needed, and it's making this PR more complicated and less clean than it could be.

roidelapluie: @brian-brazil it is needed. This makes the relative files relative to the *config* file, instead of the working directory. That would break users.

brian-brazil: Files are all relative to the working directory I thought, as we never fixed that?

roidelapluie: https://github.com/prometheus/prometheus/blob/989260f7eddd63ad7abf14d41f70fd93af31d4c8/config/config.go#L65

roidelapluie:
cc @slim-bean @cyriltovena and @rfratto FYI

discovery/discovery.go
abursavich: Comment removed.

config/config_test.go
abursavich: Roundtrip-ing the yaml is to test marshaling. There were essentially no tests for marshaling (just one to make sure secrets were masked). Since there's a lot of effort behind testing the unmarshaling a large "good" config, this lets the marshal test piggyback on that for free.

abursavich:
Fixed:
-  registration, reflection, and yaml stuff is moved to the `discovery` package
- `Validate() error` is optional and `promtool` will use it if it exists
- typos/nits

I'm gonna wait for more concrete feedback before changing anything else.

config/common.go
brian-brazil: Ah, we must have fixed this back in Prometheus 2 then.

We should find a cleaner way to resolve it, such as passing it in the constructor rather than by adding whole new methods to the interface.

config/discovery.go
brian-brazil: Performance isn't important here, plus sync.Map is only for certain very specific use cases. Use a normal map with a lock.

cmd/promtool/main.go
```diff
@@ -255,57 +257,40 @@ func checkConfig(filename string) ([]string, error) {
...
-		}
-
-		for _, filesd := range scfg.ServiceDiscoveryConfig.FileSDConfigs {
-			for _, file := range filesd.Files {
-				files, err := filepath.Glob(file)
-				if err != nil {
-					return nil, err
+		for _, c := range scfg.ServiceDiscoveryConfigs {
+			switch v := c.(type) {
+			case interface{ Validate() error }:
```
brian-brazil, line 266: This is unnecessary complexity, validation should be happening at unmarshall time and we should also be avoiding optional interfaces.

abursavich: > This is unnecessary complexity, validation should be happening at unmarshall time

This is checking for the existence of files and validity of their content. Unmarshaling YAML shouldn't have the side-effect of reading TLS certificates from disk. What if I'm just shuffling YAML configs around on a server that doesn't have the TLS certificates (e.g. `prometheus-operator`)? Now I can't unmarshal YAML? Implementing this would be extremely complex given that the path resolution is unknown to the unmarshalers, and even if we managed it, it would be in direct contradiction to your other requirement of delaying path resolution until Discoverer construction.

> we should also be avoiding optional interfaces

It's optional because you objected to it being required.

It exists in the first place because a generic method of validating configs in `promtool` was requested during this review. Previously, only a subset of files in kubernetes configs were checked. Now all files in all configs are. I think this is a useful improvement to `promtool`, but again it's not a feature that I'm trying to champion.

brian-brazil: > Now I can't unmarshal YAML?

Yes, that's the existing behaviour. All this code is Prometheus-internal, and never used by anyone else.

> It's optional because you objected to it being required.

I object to the way it is implemented. Making it optional doesn't fix that, it only makes it more complicated.

config/discovery.go
abursavich: Done.

cmd/promtool/main.go
abursavich: The existing behavior is _not_ to fail unmarshaling if the object contains a reference to a file that doesn't exist. The existing unmarshaling doesn't have the context necessary to resolve the file paths.

I reverted the previously requested changes to `promtool` and removed the `Validate() error` method. Now this whole thing is a behavioral no-op again and only addresses the issue it is originally intended to: dependencies.

If you want the successful conversion of a byte buffer into a go struct to be dependent on the existence and validity of arbitrary files on disk, I suggest you create a separate issue.

cstyan: `If you want the successful conversion of a byte buffer into a go struct to be dependent on the existence and validity of arbitrary files on disk, I suggest you create a separate issue.` @brian-brazil this is something I still plan to do something about, related to the other PR I opened. If this is a blocker for @abursavich's PR I can prioritize making that change sooner rather than later.

roidelapluie:
https://github.com/prometheus/prometheus/issues/5222

We have here a request for users NOT to check for files during unmarshal.

discovery/README.md
brian-brazil: Why can't this go in NewDiscoverer?

cmd/promtool/main.go
brian-brazil: > The existing behavior is not to fail unmarshaling if the object contains a reference to a file that doesn't exist. 

What matters is if the config works or not, what exact stage a broken config is detected at internally isn't important from the user standpoint. A failed config load would be better than a failed scrape of course.

> If this is a blocker for @abursavich's PR I can prioritize making that change sooner rather than later.

It's not per-se, but the current way we do things is a bit all over the place and both this PR's proposed changes and @cstyan's would aggravate the situation. I'm hoping to at least not make things worse in this PR, and best case have this in a much cleaner situation for @cstyan's work plus #5222 which @roidelapluie mentioned. Presuming we decide to implement #5222 - I'm aware of it but it is non-trivial and poses maintainability issues given the current code layout, issues which this PR could and possibly has to solve.

As it stands it looks like we're adding complexity and further layering violations to paper over a suboptimal design that has built up over time, rather than addressing the core design issues here.

cmd/promtool/main.go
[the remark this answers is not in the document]
brian-brazil: Yes, this needs to be solved generally.

cmd/promtool/main.go
abursavich: > What matters is if the config works or not, what exact stage a broken config is detected at internally isn't important from the user standpoint. A failed config load would be better than a failed scrape of course.

Here's a sketch of formalizing the Validater interface, implementing it on all of the config types recursively, and using it to fail fast: https://github.com/abursavich/prometheus/compare/config...abursavich:validate

discovery/README.md
abursavich: By the time NewDiscoverer is called, the context of which directory the config file was loaded from is long gone.

FWIW, the resolving of paths relative to the config file's path isn't just for service discovery configs. It also happens in:
- rule files
- scrape configs (http client config)
- alertmanager configs (http client config)
- remote read configs (http client config)
- remote write configs (http client config)

If you're still set on changing this behavior, it can be a separate issue. Since there are no API guarantees, leaving SetOptions shouldn't be a problem for now.

cmd/promtool/main.go
brian-brazil: That seems like an okay approach, though several of the checks there should be caught at unmarhsall time, and that code should be down in common so it's properly recursive.

discovery/README.md
brian-brazil: > By the time NewDiscoverer is called, the context of which directory the config file was loaded from is long gone.

Why not plumb it through?

> If you're still set on changing this behavior, it can be a separate issue. Since there are no API guarantees, leaving SetOptions shouldn't be a problem for now.

Not having API guarantees means changes can break APIs, it doesn't mean that code health can be ignored.

abursavich: If we delay it until `NewDiscoverer`, then the configs can't be validated before use.

I understand the desire to keep the original config pristine for rendering. However, plumbing the original config's directory through to every use of rule files, scrape configs, alertmanager configs, remote read configs, remote write configs, and service discovery configs seems much less maintainable/healthy than the status quo of the last 5 years. The current unmarshaling contains many examples of overriding values with defaults, copying global config values into scrape configs, etc. The config served by the web API just isn't the same as the one on disk. Arguably more importantly, it's how Prometheus is interpreting the config on disk, which includes how it resolves file paths.

abursavich: Here's a sketch of a DirectorySetter interface implemented recursively with reflection, deep cloning implemented recursively with reflection, all to save the original config for the web api: https://github.com/abursavich/prometheus/compare/config...abursavich:setdir

brian-brazil: > If we delay it until NewDiscoverer, then the configs can't be validated before use.

We could catch it during reload, but I have a strong suspicion that the reload code doesn't generally handle that correctly currently.

> seems much less maintainable/healthy than the status quo of the last 5 years.

I'm not sure there. We've had the same bugs again and again where a new field wasn't added to the various places it needs to be added for all this to be handled correctly. That sort of accidental oversight needs to be dealt with, so something that you can't forget to implement would be an improvement. So if a SetDirectory approach is used, it needs to be either a panic or compile failure not to implement and call it.

> The current unmarshaling contains many examples of overriding values with defaults, copying global config values into scrape configs, etc. The config served by the web API just isn't the same as the one on disk.

That is a good point.

> deep cloning implemented recursively with reflection

That reflection doesn't feel right at all, we should be using explicit interfaces everywhere we can (which is basically everywhere bar where the SDs dynamically hook in).

abursavich: re: `SetDirectory`

My original sketch for `Clone` and `SetDirectory` both used reflection because I didn't trust that people would keep the `Clone`/`SetDirectory` methods up to date with changes to their types (e.g. new fields).

The reflection version of `SetDirectory` would recurse through the values calling `SetDirectory` on any type with the method. If your config struct had an `HTTPClientConfig` field, it would find the `HTTPClientConfig` and call its `SetDirectory` method without your type needing to implement a proxy. However, if you did implement `SetDirectory` on your type it would trust that you knew what you were doing and not recurse deeper.

I ended up making `SetDirectory` explicit and keeping reflection for `Clone`, since I generally think explicit is better than reflection (magic) and the number of things people have to track and get right for `SetDirectory` is much smaller than for `Clone`.

It's worth noting that `SetDirectory` and `Validate` will generally have the same shape and should probably use the same calling convention if both are implemented. Things that are known at `Unmarshal` won't need `SetDirectory` or `Validate`.

I could switch back to reflection for `SetDirectory` or keep it as is and use reflection in tests to make sure `SetDirectory` has been called on all of the "leaves"?

Side Note: The current behavior doesn't update QueryLogFile in GlobalConfig.

---

re: `Clone`

I only needed to add `Clone` to keep a separate copy of the config for the web API. If we drop that extra requirement for now, since there are many other areas where it doesn't match what's on disk, we can get rid of `Clone`.

Side Note: It would be simple to use the raw file data for the web API, but that would expose secrets which makes it a no-go.

---

I'm having a hard time keeping track of what the criteria is for getting this PR though. I would be happy to contribute to some of these other issues, but I would like to avoid this becoming a catch all for everything we want to change about configs. I think it makes sense to come to an agreement on what to do about `SetDirectory` and to spin off config validation and original config preservation to separate issues.

cstyan: > I would like to avoid this becoming a catch all for everything we want to change about configs. I think it makes sense to come to an agreement on what to do about SetDirectory and to spin off config validation and original config preservation to separate issues.

This seems reasonable to me.

abursavich: Here's `SetDirectory` where it's explicitly called down the value chain: https://github.com/abursavich/prometheus/compare/config...abursavich:setdir-v2

There's already some decent testing with the existing "good" test config, but it only cover fields that are used (e.g. values that have an `HTTPClientConfig` must specify paths in the testdata for it to be useful here). I added reflection-based testing that will take the "good" config  and set any field that looks like a file path (e.g. its name has the suffix `File` or `Files`).

brian-brazil: > I ended up making SetDirectory explicit and keeping reflection for Clone, since I generally think explicit is better than reflection (magic) 

I agree, let's avoid reflection where we can.

> I could switch back to reflection for SetDirectory or keep it as is

As-is I'd say.

> use reflection in tests to make sure SetDirectory has been called on all of the "leaves"?

I'd prefer to find a way that doesn't need it. For example can the HTTPClientConfig fail in some way if SetDirectory is not called either at Validate or instantiation? A explicit failure will be caught and easily fixed (even if it manages to make it out into a release), whereas silently not setting it will not be easy to spot or debug.

> The current behavior doesn't update QueryLogFile in GlobalConfig.

That is a bug we should fix at some point then.

> I only needed to add Clone to keep a separate copy of the config for the web API.

We don't need that it looks like.

> I'm having a hard time keeping track of what the criteria is for getting this PR though. 

I can appreciate that. At this point we're still figuring out how to fully design this, so some back and forth discussion is inevitable as this is a non-trivial problem. This also involves paying off some technical debt that has built up over the years, as we can no longer cheat with layering violations for validation.

> to spin off config validation and original config preservation to separate issues.

We can't break existing functionality, and both config validation and showing the config in the UI are existing features.

>  I added reflection-based testing that will take the "good" config and set any field that looks like a file path

Keep in mind that Prometheus is not the only thing using HTTPClientConfig, and reflection-based testing is enough of a smell already without having to replicate it to other repositories.

abursavich: > > spin off config validation and original config preservation to separate issues
> 
> We can't break existing functionality, and both config validation and showing the config in the UI are existing features.

The current version of this PR preserves the existing functionality (or lack thereof) regarding validation (same `Unmarshal` checks) and preservation (file paths were and are modified), which is why I suggest spinning them off to separate issues.

> Keep in mind that Prometheus is not the only thing using HTTPClientConfig, and reflection-based testing is enough of a smell already without having to replicate it to other repositories.

> For example can the HTTPClientConfig fail in some way if SetDirectory is not called either at Validate or instantiation? A explicit failure will be caught and easily fixed (even if it manages to make it out into a release), whereas silently not setting it will not be easy to spot or debug.

This is tricky... There are `Load` functions and `LoadFile` functions and only `LoadFile` sets the directory. Depending on the different use cases, calling it might not be required? It might make sense to construct and use an HTTPClientConfig that isn't read directly from a file?

I just looked at the alertmanager config code and it has plenty of `HTTPClientConfig`s, but [it only updates `Templates`](https://github.com/prometheus/alertmanager/blob/8e861c646bf67599a1704fc843c6a94d519ce312/config/config.go#L189-L217). This is probably a bug, but what to do about it now?

A nice thing about the reflection testing is it confirms that calling SetDirectory on a root value is propagated to its inner and leaf values, but it doesn't force you to into calling SetDirectory on anything. The opt in nature of this testing gives other components (e.g. alertmanager) time to sort out their behavior without failing at runtime.

I factored the testing code out into `common/config/configtest` so it can be easily adopted by other repos.

brian-brazil: > The current version of this PR preserves the existing functionality (or lack thereof) 

Not going backwards is the main thing. We can no longer reach into the internals of HTTPClientConfig from Prometheus though.

> It might make sense to construct and use an HTTPClientConfig that isn't read directly from a file?

It doesn't, they all come from either user-provided config files or unittests. Even then you could pass in pwd or whatnot if they came from elsewhere. We could for example add the base directory as a argument to Load - though unittests tend to instantiate a HTTPClientConfig directly.

> This is probably a bug, but what to do about it now?

Yip, that's a bug. Not a problem for this PR, but will need resolving when this change is propagated out. Current problems like this need to be fixed systemically, even if not all in this PR.

> it doesn't force you to into calling SetDirectory on anything. The opt in nature of this testing gives other components (e.g. alertmanager) time to sort out their behavior without failing at runtime.

We control when the alertmanager picks up a new version of common, and I'd expect that all to be done in one PR. I'd really prefer not to have that sort of reflection lying around the codebase.

abursavich: Okay... So I think we're just down to how to test/enforce that the code properly propagates calls to `SetDirectory`.

I prefer the reflection-based tests because they cover all `DirectorySetters` (e.g. `marathon.SDConfig` which includes an `AuthTokenFile`) and are now easy to add to any root config you want to test. Since tests instantiate `HTTPClientConfig` and `TLSConfig` directly for comparison, it would be burdensome to make them flip some unexported `hasSetDir` bool when they already use the paths themselves to verify that `testdata/` has been properly prepended. Also, every instance where `HTTPClientConfig` or `TLSConfig` is empty/unused in tests would require the same bit flip. Practically, that seems like a lot of tedium and headache for little gain.

abursavich: Once I created `common/config/configtest`, this is what was required to add the test to `prometheus/config`, and it would be similar for others like `alertmanager/config`:

```go
func TestSetDirectory(t *testing.T) {
	files, err := filepath.Glob("testdata/*.good.yml")
	testutil.Ok(t, err)
	testutil.Assert(t, len(files) > 0, "no good configs found")

	loadFile := func(file string) (config_util.DirectorySetter, error) {
		return LoadFile(file)
	}
	for _, file := range files {
		configtest.TestSetDirectory(t, file, loadFile, assertConfigsEqual)
	}
}
```

brian-brazil: > it would be burdensome to make them flip some unexported hasSetDir bool when they already use the paths themselves to verify that testdata/ has been properly prepended

It'd be far, far from the first time we needed to bulk change tests for something like that. We could also work off the paths being relative rather than absolute rather than an explicit flag.

> this is what was required to add the test to prometheus/config,

It's still a whole pile of reflection backing it, and I'd rather not have that reflection.

roidelapluie:
Test are failing.

If you import the new common (v0.12.0) please disable http2 in the client config.

abursavich:
> Test are failing.

The go tests are passing, but DCO and CI are failing because I rebased in an unsigned commit labeled `[DO NOT MERGE]` that vendors the changes in https://github.com/prometheus/common/pull/250, which isn't merged yet.

discovery/README.md
brian-brazil: I'm still seeing that reflection.

abursavich: > I'm still seeing that reflection.

Heh, yep :smile:

I also prefer not to use reflection when it isn't necessary, but its use in these tests covers more than would be by other means. It makes sure `SetDirectory` is propagated to any inner/leaf values that implement it and its heuristics find values like `QueryLogFile` in `GlobalConfig` that are files but aren't updated. The former of which may be left out of hand-rolled tests/validation and the later of which would never be covered by hand-rolled tests/validation.

config/config_test.go
brian-brazil: This dependency is still here. If you want to test this, do so without adding a 2nd way to do struct comparisons to the codebase.

abursavich:
> This dependency is still here. If you want to test this, do so without adding a 2nd way to do struct comparisons to the codebase.

As previously stated, it's needed to test marshaling via roundtrip-ing which obfuscates Secrets, in which case `"<secret>"` should be a sufficient match for `"foobar"`. Otherwise we're gonna have to write something brittle to replace all secrets, rely on subtle reflection, or not be able to reuse the existing unmarshaling test cases to test marshaling.

Thankfully [github.com/google/go-cmp/cmp](https://pkg.go.dev/github.com/google/go-cmp@v0.5.1/cmp?tab=doc) is specifically designed for testing and conveniently allows you to override equality comparisons for any type you wish. Cool, drop in a `cmp.Option` to customize the comparison of secrets:

```go
cmp.Comparer(func(x, y config.Secret) bool {
	// Secrets are marshaled as "<secret>" and can't be compared to the original.
	const secret = "<secret>"
	return x == y || (x == secret && y != "") || (y == secret && x != "")
})
```

However, I first introduced `cmp` when I was prototyping the new config unmarshaling before writing the marshal roundtrip-ing tests. Previously each discovery.Config type had its own slice but now they are all gathered into a single slice in a particular order (*foreshadowing*).

At some point I ran the existing tests with the existing comparison method using reflect.DeepEqual, spew.ConfigState.Sdump, and difflib.GetUnifiedDiffString, which yielded this incomprehensible output:
```
--- FAIL: TestLoadConfig (0.01s)
    /.../prometheus/config/config_test.go:736: [31m
        
        exp: &config.Config{GlobalConfig:config.GlobalConfig{ScrapeInterval:15000000000, ScrapeTimeout:10000000000, EvaluationInterval:30000000000, QueryLogFile:"", ExternalLabels:labels.Labels{labels.Label{Name:"foo", Value:"bar"}, labels.Label{Name:"monitor", Value:"codelab"}}}, AlertingConfig:config.AlertingConfig{AlertRelabelConfigs:[]*relabel.Config(nil), AlertmanagerConfigs:config.AlertmanagerConfigs{(*config.AlertmanagerConfig)(0x31866c0)}}, RuleFiles:[]string{"testdata/first.rules", "testdata/my/*.rules"}, ScrapeConfigs:[]*config.ScrapeConfig{(*config.ScrapeConfig)(0x3187440), (*config.ScrapeConfig)(0x3187560), (*config.ScrapeConfig)(0x3187680), (*config.ScrapeConfig)(0x31877a0), (*config.ScrapeConfig)(0x31878c0), (*config.ScrapeConfig)(0x31879e0), (*config.ScrapeConfig)(0x3187b00), (*config.ScrapeConfig)(0x3187c20), (*config.ScrapeConfig)(0x3187d40), (*config.ScrapeConfig)(0x3187e60), (*config.ScrapeConfig)(0x3187f80), (*config.ScrapeConfig)(0x3186d80), (*config.ScrapeConfig)(0x3186ea0), (*config.ScrapeConfig)(0x3186fc0), (*config.ScrapeConfig)(0x31870e0), (*config.ScrapeConfig)(0x3187200), (*config.ScrapeConfig)(0x3187320)}, RemoteWriteConfigs:[]*config.RemoteWriteConfig{(*config.RemoteWriteConfig)(0x3186880), (*config.RemoteWriteConfig)(0x3186980)}, RemoteReadConfigs:[]*config.RemoteReadConfig{(*config.RemoteReadConfig)(0x3185060), (*config.RemoteReadConfig)(0x3185120)}}
        
        got: &config.Config{GlobalConfig:config.GlobalConfig{ScrapeInterval:15000000000, ScrapeTimeout:10000000000, EvaluationInterval:30000000000, QueryLogFile:"", ExternalLabels:labels.Labels{labels.Label{Name:"foo", Value:"bar"}, labels.Label{Name:"monitor", Value:"codelab"}}}, AlertingConfig:config.AlertingConfig{AlertRelabelConfigs:[]*relabel.Config(nil), AlertmanagerConfigs:config.AlertmanagerConfigs{(*config.AlertmanagerConfig)(0xc0004581c0)}}, RuleFiles:[]string{"testdata/first.rules", "testdata/my/*.rules"}, ScrapeConfigs:[]*config.ScrapeConfig{(*config.ScrapeConfig)(0xc000868000), (*config.ScrapeConfig)(0xc000868480), (*config.ScrapeConfig)(0xc0008685a0), (*config.ScrapeConfig)(0xc0008687e0), (*config.ScrapeConfig)(0xc000868900), (*config.ScrapeConfig)(0xc000868a20), (*config.ScrapeConfig)(0xc000868b40), (*config.ScrapeConfig)(0xc000868c60), (*config.ScrapeConfig)(0xc000868d80), (*config.ScrapeConfig)(0xc000868ea0), (*config.ScrapeConfig)(0xc000868fc0), (*config.ScrapeConfig)(0xc000869200), (*config.ScrapeConfig)(0xc000869440), (*config.ScrapeConfig)(0xc000869680), (*config.ScrapeConfig)(0xc0008697a0), (*config.ScrapeConfig)(0xc0008698c0), (*config.ScrapeConfig)(0xc0008699e0)}, RemoteWriteConfigs:[]*config.RemoteWriteConfig{(*config.RemoteWriteConfig)(0xc0003acc30), (*config.RemoteWriteConfig)(0xc0003acd20)}, RemoteReadConfigs:[]*config.RemoteReadConfig{(*config.RemoteReadConfig)(0xc000520420), (*config.RemoteReadConfig)(0xc0005204d0)}}
        
        Diff:
        [39m
```

WHAT?

After trying to make sense of that for way too long, I dropped in a `cmp.Diff` instead:
```
--- FAIL: TestLoadConfig (0.01s)
    /.../prometheus/config/config_test.go:702: unexpected diff:
        
          &config.Config{
          	GlobalConfig:   {ScrapeInterval: s"15s", ScrapeTimeout: s"10s", EvaluationInterval: s"30s", ExternalLabels: {{Name: "foo", Value: "bar"}, {Name: "monitor", Value: "codelab"}}},
          	AlertingConfig: {AlertmanagerConfigs: {&{ServiceDiscoveryConfigs: {discovery.StaticConfig{&{Targets: {s`{__address__="1.2.3.4:9093"}`, s`{__address__="1.2.3.5:9093"}`, s`{__address__="1.2.3.6:9093"}`}, Source: "0"}}}, Scheme: "https", Timeout: s"10s", APIVersion: "v1", ...}}},
          	RuleFiles:      {"testdata/first.rules", "testdata/my/*.rules"},
          	ScrapeConfigs: []*config.ScrapeConfig{
          		&{
          			... // 8 identical fields
          			SampleLimit: 0,
          			TargetLimit: 0,
          			ServiceDiscoveryConfigs: discovery.Configs{
        - 				discovery.StaticConfig{s"0"},
          				&file.SDConfig{Files: {"testdata/foo/*.slow.json", "testdata/foo/*.slow.yml", "testdata/single/file.yml"}, RefreshInterval: s"10m"},
          				&file.SDConfig{Files: {"testdata/bar/*.yaml"}, RefreshInterval: s"5m"},
        + 				discovery.StaticConfig{s"0"},
          			},
          			HTTPClientConfig:     {BearerTokenFile: "testdata/valid_token_file"},
          			RelabelConfigs:       {&{SourceLabels: {"job", "__meta_dns_name"}, Separator: ";", Regex: s"^(?:(.*)some-[regex])$", TargetLabel: "job", ...}, &{SourceLabels: {"abc"}, Separator: ";", Regex: s"^(?:(.*))$", TargetLabel: "cde", ...}, &{Separator: ";", Regex: s"^(?:(.*))$", TargetLabel: "abc", Replacement: "static", ...}, &{Separator: ";", Regex: s"^(?:)$", TargetLabel: "abc", Replacement: "static", ...}},
          			MetricRelabelConfigs: nil,
          		},
          		&{JobName: "service-x", HonorTimestamps: true, ScrapeInterval: s"50s", ScrapeTimeout: s"5s", ...},
          		&{JobName: "service-y", HonorTimestamps: true, ScrapeInterval: s"15s", ScrapeTimeout: s"10s", ...},
          		... // 14 identical elements
          	},
          	RemoteWriteConfigs: {&{URL: &{URL: &{Scheme: "http", Host: "remote1", Path: "/push"}}, RemoteTimeout: s"30s", WriteRelabelConfigs: {&{SourceLabels: {"__name__"}, Separator: ";", Regex: s"^(?:expensive.*)$", Replacement: "$1", ...}}, Name: "drop_expensive", ...}, &{URL: &{URL: &{Scheme: "http", Host: "remote2", Path: "/push"}}, RemoteTimeout: s"30s", Name: "rw_tls", HTTPClientConfig: {TLSConfig: {CertFile: "testdata/valid_cert_file", KeyFile: "testdata/valid_key_file"}}, ...}},
          	RemoteReadConfigs:  {&{URL: &{URL: &{Scheme: "http", Host: "remote1", Path: "/read"}}, RemoteTimeout: s"1m", ReadRecent: true, Name: "default", ...}, &{URL: &{URL: &{Scheme: "http", Host: "remote3", Path: "/read"}}, RemoteTimeout: s"1m", Name: "read_special", HTTPClientConfig: {TLSConfig: {CertFile: "testdata/valid_cert_file", KeyFile: "testdata/valid_key_file"}}, ...}},
          }
```

Ahhhh... The expected value had its manually constructed slice of configs ordered differently. Cool, fix the ordering in the test case and drop in another `cmp.Option` so nobody is ever bitten by this again:
```go
cmp.Transformer("discovery.Config.Sorter", func(cfgs discovery.Configs) discovery.Configs {
	// The service discovery configs are marshaled and unmarshaled
	// in a particular order, sorted by their names and with
	// static_configs always at the end.
	//
	// At the time of writing this comment, all of the test configs
	// are in this expected order, but adding this tranformer protects
	// future test writers from going mad over inexplicably failed
	// comparisons between unmarshaled yaml configs and hand-crafted
	// go values (with configs in the "wrong" order).
	clone := append(make(discovery.Configs, 0, len(cfgs)), cfgs...)
	sort.SliceStable(clone, func(i, k int) bool {
		return clone[i].Name() < clone[k].Name()
	})
	return clone
})
```

I believe you're interested in the health and maintainability of the code base, but I don't think removing or crippling these tests is the right decision. That said, I'll do it if that's still your honest assessment.

roidelapluie:
Sometimes I also switch in my local env our struct comparison thing
because it does not show differences private fields but still errors on
them.

brian-brazil:
Changing how we do comparisons in tests across all of Prometheus is out of scope I'd say, I'm not sure we've even finished the last round of unittest refactoring.

If you need to do a test that marshalling is working you don't need to go such exhaustive lengths. For example pick one or two cases that cover the new code, and check that the output string matches a provided value. We don't need to consider secrets here, as it's that marshalling is working through the reflection logic is what needs to be tested (plus we've other tests to catch secret issues).

abursavich:
OK. Usage of `cmp`  and `reflect` are removed from tests. I just need https://github.com/prometheus/common/pull/250 merged before I can finalize this.

discovery/registry.go
```diff
@@ -0,0 +1,261 @@
...
+	if len(targets) > 0 {
+		configs = append(configs, StaticConfig(targets))
+	}
+	return configs, nil
+}
+
+// MarshalYAMLWithInlineConfigs helps implement yaml.Marshal for structs
+// that have a Configs field that should be inlined.
+func MarshalYAMLWithInlineConfigs(in interface{}) (interface{}, error) {
+	// This function can be removed if https://github.com/go-yaml/yaml/issues/642 is fixed.
```
roidelapluie, line 201: Is that true for v2 too?

abursavich: Yep

abursavich: https://play.golang.com/p/DjhRLsGmbza

abursavich: Actually... Maybe not... doing some tests.

abursavich: It's complicated. The behavior is different between the two versions.

But before getting into that issue, both v2 and v3 only allow inlining structs (or maps, but inlined maps destroy strictness and are another can of worms), and we have a slice of configs that we would like to inline. This is a separate issue and it makes that comment invalid because https://github.com/go-yaml/yaml/issues/642 isn't the only blocker.

Back to the original issue, in v3 it seems that when a struct implementing `yaml.Unmarshaler` is inlined in a parent struct, the fields that may or may not be in the inlined struct are not taken into account when enforcing strictness on the parent struct. So fields that exist in the inlined struct or exist in its custom unmarshaling logic will cause "field not found" errors on the parent.

In v2, it seems to ignore any custom unmarshaling logic when a `yaml.Unmarshaler` is inlined. Unmarshal is never called on the inlined struct. Strictness is enforced based on the union of fields in the parent struct and inlined struct. Again, fields that only exist in custom unmarshaling logic (which is never called) will cause "field not found" errors on the parent.

abursavich: Here's a demo for v2: https://play.golang.com/p/yvcjhettRth

abursavich: I removed the comment.

roidelapluie (approved):

config/testdata/empty_static_config.bad.yml
```diff
@@ -1,4 +1,4 @@
 scrape_configs:
 - job_name: "test"
   static_configs:
-  -
+  - targets: []
```
brian-brazil, line 4: Why did you change this test?

discovery/targetgroup/targetgroup.go
```diff
@@ -82,12 +77,27 @@ func (tg *Group) UnmarshalJSON(b []byte) error {
...
-		})
-	}
-	tg.Labels = g.Labels
+	tg.Targets = targets(g.Targets)
+	tg.Labels = labels(g.Labels)
 	return nil
 }
+
+func targets(addrs []string) []model.LabelSet {
+	if len(addrs) == 0 {
```
brian-brazil, line 86: This seems like an unneeded optimisation.

discovery/targetgroup/targetgroup.go
```diff
@@ -82,12 +77,27 @@ func (tg *Group) UnmarshalJSON(b []byte) error {
...
+	sets := make([]model.LabelSet, len(addrs))
+	for i, addr := range addrs {
+		sets[i] = model.LabelSet{
+			model.AddressLabel: model.LabelValue(addr),
+		}
+	}
+	return sets
+}
+
+func labels(lbls model.LabelSet) model.LabelSet {
```
brian-brazil, line 98: Why do you have this? These should be the same thing semantically.

documentation/examples/prometheus.yml
```diff
@@ -9,7 +9,7 @@ alerting:
   alertmanagers:
   - static_configs:
     - targets:
-      # - alertmanager:9093
```
brian-brazil, line 12: Leave this as-is, a new user probably isn't running an alertmanager

documentation/examples/prometheus.yml
```diff
@@ -26,4 +26,5 @@ scrape_configs:
     # scheme defaults to 'http'.
 
     static_configs:
-    - targets: ['localhost:9090']
+    - targets:
```
brian-brazil, line 29: Why are you changing this? Generally, teaching users to put things with colons in quotes is a good idea.

brian-brazil (commented):
Generally looks fine, a few smaller points.

abursavich:
> Generally looks fine, a few smaller points.

Issues were all with https://github.com/abursavich/prometheus/commit/d4693a847b0c15988411849f6ae4ff1dec1e3307, which I've dropped.

The config unmarshaling checks "empty or null" values. The custom `targetgroup.Group` unmarshaling causes `{}` to be unmarshaled as `targetgroup.Group{Targets: model.LabelSet{}}` instead of `targetgroup.Group{}`, because of this line: https://github.com/prometheus/prometheus/blob/25f929b7721c113a827c5cec4ea4c2f502c95dd9/discovery/targetgroup/targetgroup.go#L48

These values are semantically equivalent, but the one with a zero-length non-nil `Targets` field is not the [zero value](https://godoc.org/reflect#Value.IsZero) (i.e. empty) for `targetgroup.Group`.

brian-brazil:
Great, thanks for working through this! This is something we've wanted for quite a while.

roidelapluie:
Thank you @abursavich !!!!

abursavich:
Huzzah! Cheers!

brancz:
I can only echo @brian-brazil's comment, this has been on our list for a very long time, thank you so much for doing this!

bboreham:
Cortex image went down 33% after applying this change - thanks!
```
quay.io/cortexproject/cortex    master-e3ca9e391       588cf43d6151        9 days ago          66.7MB
quay.io/cortexproject/cortex    master-2ce5ce6f5       2bea160a1678        10 days ago         101MB
```

bboreham:
This PR is linked from the 2.21.0 changelog as "Disable HTTP/2 because of concerns with the Go HTTP/2 client."; I cannot easily see where that change is included in this PR and certainly wouldn't guess it from the description.

Please update the description with information about non-obvious changes included in this PR.

abursavich:
This change upgraded github.com/prometheus/common from v0.11.1 to v0.13.0. It required changes that were in v0.13.0. HTTP/2 was disabled in v0.12.0.

PR #7708, opened 2020-07-31
Add additional tag/log to rules Manager Eval trace span.

cstyan:
Adds a tag to the span that says whether the evaluation succeeded, and a log line to the span with the error message in the result of failure.

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -576,6 +576,8 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
 				if _, ok := err.(promql.ErrQueryCanceled); !ok {
 					level.Warn(g.logger).Log("msg", "Evaluating rule failed", "rule", rule, "err", err)
 				}
+				sp.SetTag("error", true)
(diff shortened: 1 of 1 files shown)


roidelapluie:
should we simply set the "standard" error=true on failed evaluations?

cstyan:
Fine by me, the name of the tag isn't support important to me and if there's a standard I'm not aware of it. However I do need both the status (failed/succeeded) and the error message log.

roidelapluie:
> Fine by me, the name of the tag isn't support important to me and if there's a standard I'm not aware of it. However I do need both the status (failed/succeeded) and the error message log.

As the jaeger UI offers error=true in its exemply by default I have always assumed it was kinda standard

roidelapluie:
error=true is detected by jaeger ui as well and displayed with a red tag.

cstyan:
`error=true is detected by jaeger ui as well and displayed with a red tag.` yep, you're right, I only just learned about this the day I opened this PR, @roidelapluie PTAL

PR #7710, opened 2020-07-31
Remote Write Checkpointing

njingco:
Currently remote write sends data via Prometheus WAL, but the pointer for each remote write queue in the WAL does not persist across restarts, as per #6333 .  This PR will attempt to solve this issue by recording an identifier to indicate which segment file to re transmit on Prometheus restart. Re transmitting the whole segment file will potentially have zero data loss. Please refer to my proposal document for more information: [Proposal Doc](https://docs.google.com/document/d/1gw1C731ysTwmIT7s3sqW5gwzC4qZdGKrpAPsz_tS7t0/edit?usp=sharing)

This PR is currently in progress, but will include the base of the proposal. It will include recording the identifier during shut down, and at a set time, identifying the segment number for each endpoint, and error/corruption checks for the record json file.

It currently has the ability to record the identifier into a json. Sample of the record file is in `SegmentRecord.json`. It also does corruption checks of the json file at the start of the `queue_manager.go`. 

Whats left to do in this PR add the timed recording from start of `queue_manager`, and implementing the time customization in the configuration file. 

This PR will not have the implementation of setting the segment file for re transmission. This will be in a future PR. This PR will only manage the recording of the identifiers.


storage/remote/queue_manager.go
```diff
@@ -367,6 +385,31 @@ func (t *QueueManager) Start() {
...
+	if err != nil {
+		fmt.Println(err)
+	} else {
+		for _, record := range check.Checkpoints {
+			segmentNumber, err := strconv.Atoi(record.Segment)
+			if err != nil {
+				fmt.Println(err)
+			}
+
+			if segmentNumber > -1 && segmentNumber < 100000000 {
```
cstyan, line 406: we could define these values somewhere `invalidSegmentNumber` and `maxSegmentNumber`, though I'd check the WAL package files to see if anything similar already exists?

storage/remote/queue_manager.go
```diff
@@ -367,6 +385,31 @@ func (t *QueueManager) Start() {
...
+	var check Checkpoints
+
+	jsonFile, err := os.Open("SegmentRecord.json")
+	defer jsonFile.Close()
+
+	byteValue, _ := ioutil.ReadAll(jsonFile)
+	err = json.Unmarshal(byteValue, &check)
+
+	if err != nil {
+		fmt.Println(err)
```
cstyan, line 398: We should use the queue managers logger rather than `fmt.Println`, you can look at usage in this file like `level.Info` or `level.Error` depending on what you want to log.

storage/remote/queue_manager.go
```diff
@@ -398,6 +443,37 @@ func (t *QueueManager) Stop() {
...
+	Checkpoint.Recorded = time.Now()
+	Checkpoint.Checkpoints = append(Checkpoint.Checkpoints, t.getSegmentRecords())
+
+	// now Marshal it
+	data, err := json.MarshalIndent(Checkpoint, "", "")
+	if err != nil {
+		fmt.Println(err)
+	}
+
+	err = ioutil.WriteFile("SegmentRecord.json", data, 0644)
```
cstyan, line 470: Have a look at how we create and write the normal WAL checkpoints: https://github.com/prometheus/prometheus/blob/master/tsdb/wal/checkpoint.go#L84-L257 Note the use of a temporary dir/file. What would happen if we, for some reason, were reading the current remote write checkpoint file and `RecordSegment` was called.

tsdb/wal/watcher.go
```diff
@@ -252,10 +255,13 @@ func (w *Watcher) Run() error {
 		}
 
 		// For testing: stop when you hit a specific segment.
+
```
cstyan, line 258: unnecessary newline

SegmentRecord.json
```diff
@@ -0,0 +1,25 @@
+{
```
cstyan, line 1: I don't think you meant to commit this file?

storage/remote/queue_manager.go
```diff
@@ -398,6 +443,37 @@ func (t *QueueManager) Stop() {
 	t.metrics.unregister()
 }
 
+// Getting the data from a shard and structuring it as a SegmentRecord from a shard
+func (t *QueueManager) getSegmentRecords() SegmentRecord {
+	RecordSegment := t.shards.qm.watcher.SegmentFile
+	RecordEndpoint := t.shards.qm.url
```
csmarchbanks, line 449: The URL will not uniquely identify a QueueManager. Instead, you should use `t.storeClient.Name()`.

storage/remote/queue_manager.go
```diff
@@ -398,6 +443,37 @@ func (t *QueueManager) Stop() {
 	t.metrics.unregister()
 }
 
+// Getting the data from a shard and structuring it as a SegmentRecord from a shard
+func (t *QueueManager) getSegmentRecords() SegmentRecord {
+	RecordSegment := t.shards.qm.watcher.SegmentFile
```
csmarchbanks, line 448: `t` is already the queue manager, so this can be simplified to just `t.watcher.SegmentFile`.

Also, what would happen if you try to get the `SegmentFile` from the watcher at the same time the watcher was changing `SegmentFile`?

tsdb/wal/watcher.go
```diff
@@ -252,10 +255,13 @@ func (w *Watcher) Run() error {
...
 		if currentSegment == w.MaxSegment {
 			return nil
 		}
 
+		// Sets the current Segment files name
+		w.SegmentFile = SegmentName("", currentSegment)
```
csmarchbanks, line 264: Would it be possible for the watcher to have progressed to a new segment, without sending all samples from the previous segment? I think if you had a large enough buffer capacity in the queues that it could happen.

storage/remote/queue_manager.go
```diff
@@ -398,6 +443,37 @@ func (t *QueueManager) Stop() {
...
+	record := SegmentRecord{
+		Segment:  RecordSegment,
+		Endpoint: RecordEndpoint,
+	}
+
+	return record
+}
+
+// Writes the segment record in a JSON file
+func (t *QueueManager) RecordSegment() {
```
csmarchbanks, line 460: Does `RecordSegment()` need to be exposed?

storage/remote/queue_manager.go
```diff
@@ -367,6 +385,31 @@ func (t *QueueManager) Start() {
 	t.metrics.minNumShards.Set(float64(t.cfg.MinShards))
 	t.metrics.desiredNumShards.Set(float64(t.cfg.MinShards))
 
+	// Checking if the JSON file is corrupted
+	var check Checkpoints
```
csmarchbanks, line 389: I think it would be nice to break this into a function like `loadCheckpoint()` rather than having all the logic in `Start()`.

SegmentRecord.json
njingco: Is this not the right format for a JSON file?

storage/remote/queue_manager.go
njingco: image
These are the values get for the endpoint when trying this. Is this right?

storage/remote/queue_manager.go
njingco: Would this still be able to get each individual shards segment number?

storage/remote/queue_manager.go
njingco: What do you mean by exposed?

SegmentRecord.json
cstyan: The format looks fine, but we shouldn't upload data files Prometheus is generating to the repo. If you want/need a file for use in tests, we have specific directories for those.

storage/remote/queue_manager.go
cstyan: Yep. If you don't specify a name (https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write) for a queue in the config file the name is defaulted to the hash of the config for that queue.

storage/remote/queue_manager.go
cstyan: It's not clear what you're trying to do here actually. Shards, and also the queue manager, have no notion of the WAL or segment files.

storage/remote/queue_manager.go
cstyan: It's like public vs priavte from other languages. If the name of a function starts with a lower case, `recordSegment`, it's only available within the package it's defined. In this case, `storage/remote`. If the name of a function starts with an upper case, `RecordSegment`, it's available outside that package to anyone who imports that package.

storage/remote/queue_manager.go
```diff
@@ -45,6 +48,9 @@ const (
 
 	// Allow 30% too many shards before scaling down.
 	shardToleranceFraction = 0.3
+
+	invalidSegmentNumber = -1
```
cstyan, line 52: we should check the WAL package itself to see if these are set, and if not we should probably define them there

storage/remote/queue_manager.go
```diff
@@ -261,6 +267,26 @@ type QueueManager struct {
...
 
 	metrics *queueManagerMetrics
+	url     string
+}
+
+var (
+	// Checkpoint is the struture of the Segment Record
+	Checkpoint = Checkpoints{}
+	// Number of shards saved
+	savedShardsNum = 0
```
cstyan, line 277: this doesn't seem to be used anywhere

storage/remote/queue_manager.go
```diff
@@ -375,12 +404,43 @@ func (t *QueueManager) Start() {
 	go t.reshardLoop()
 }
 
+// loadCheck will check if the SegmentRecord.json
+// contents and segment is valid
+func (t *QueueManager) loadCheck() {
+	// Checking if the JSON file is corrupted
```
cstyan, line 410: All comments should be full sentences, so they should start with a capital and end with a period. One exception is comments for functions whose first letter is not a capital (such as the function comment above).

tsdb/wal/watcher.go
```diff
@@ -149,7 +151,8 @@ func NewWatcher(metrics *WatcherMetrics, readerMetrics *LiveReaderMetrics, logge
 		quit:          make(chan struct{}),
 		done:          make(chan struct{}),
 
-		MaxSegment: -1,
+		MaxSegment:  -1,
```
cstyan, line 154: the comment for `MaxSegment` suggests it is supposed to be used only in tests, we should confirm how it's being used right now and either a) change the comment or b) create a new variable to use instead of `MaxSegment`

SegmentRecord.json
njingco: Ah I see

storage/remote/queue_manager.go
njingco: Ohh is this the watcher that's monitoring the segment file that WAL is writing on, not the one its getting read from?

cstyan: No, this is the watcher that reads the segment file and passes the samples back to the queue manager. I think the name of the function and your comments about shards are what's confusing me.

As Chris said, we can just access `t.watcher.SegmentFile`. My point was that shards and the queue manager have no notion or concept of WAL segments. So this question `Would this still be able to get each individual shards segment number?`, the shards do not have an individual segment number. The watcher is reading a segment, passing samples back to queue manager, and queue manager is deciding which of the shards to assign a given sample to.

tsdb/wal/watcher.go
njingco: i think this was already there before.

cstyan: Yeah it was. I thought I saw other code using `MaxSegment` but maybe I was mistaken, it looks like you're using `maxSegmentNumber` now. This change (extra space) should be removed though, as it's just noise when someone is trying to review your PR :)

tsdb/wal/watcher.go
njingco: Yes it is possible

storage/remote/queue_manager.go
```diff
@@ -375,12 +398,46 @@ func (t *QueueManager) Start() {
...
+
+	for _, record := range check.Checkpoints {
+		segmentNumber, err := strconv.Atoi(record.Segment)
+
+		if err != nil {
+			level.Error(t.logger).Log("err", err)
+			return false
+		}
+
+		if segmentNumber <= wal.MinSegmentNumber || segmentNumber >= wal.MaxSegmentNumber {
```
cstyan, line 424: this check still feels strange to me given that we don't already set a min or max segment # already, I think `if segmentNumber >= 0` we should consider it valid, whether the file exists is another issue

storage/remote/write.go
```diff
@@ -81,6 +88,8 @@ func NewWriteStorage(logger log.Logger, reg prometheus.Registerer, walDir string
 
 func (rws *WriteStorage) run() {
 	ticker := time.NewTicker(shardUpdateDuration)
+	rws.startTimedRecording()
```
cstyan, line 91: this would need to be run in a go routine, and the function should take a context or other way it can be shutdown/cleaned nicely

storage/remote/write.go
```diff
@@ -177,13 +186,61 @@ func (rws *WriteStorage) Appender(_ context.Context) storage.Appender {
...
+
+// writeCheckpointRecordJSON writes the segment record in a JSON file
+func (rws *WriteStorage) writeCheckpointRecordJSON() {
+	filepath := "data/CheckpointRecord.json"
+
+	rws.checkpoint.TimeRecorded = time.Now()
+	rws.checkpoint.Checkpoints = rws.getEndpointRecordData()
+
+	// Marshal json
+	data, err := json.MarshalIndent(rws.checkpoint, "", "")
```
cstyan, line 205: should the last parameter not be `"\t"`?

tsdb/wal/watcher.go
```diff
@@ -466,10 +471,14 @@ func (w *Watcher) readSegment(r *LiveReader, segmentNum int, tail bool) error {
 		samples []record.RefSample
 		send    []record.RefSample
 	)
+
 	for r.Next() && !isClosed(w.quit) {
 		rec := r.Record()
 		w.recordsReadMetric.WithLabelValues(recordType(dec.Type(rec))).Inc()
 
+		// Sets the current Segment files it's reading from
+		w.CurrSegmentFile = SegmentName("", segmentNum)
```
cstyan, line 480: this should be set in the same location as the metric is being set

cstyan (commented):
made some comments, good progress

storage/remote/queue_manager.go
njingco: I think that's a typo i made. It should be `if segmentNumber < wal.MinSegmentNumber || segmentNumber > wal.MaxSegmentNumber` return false. Min number being 0 and max number being 99999999

storage/remote/write.go
njingco: oh it is. ill add it in

tsdb/wal/watcher.go
njingco: would this not go in the readSegment() since this is where the current segment is being read

storage/remote/queue_manager.go
```diff
@@ -261,6 +264,20 @@ type QueueManager struct {
 	samplesIn, samplesDropped, samplesOut, samplesOutDuration *ewmaRate
 
 	metrics *queueManagerMetrics
+
+	endpointRecord EndpointRecord
+}
+
+// EndpointRecord Structure holds the segment the endpoint url
+type EndpointRecord struct {
+	Segment  string
```
cstyan, line 273: the fields on this and checkpoint record don't need to be exported if you don't have a need to access those fields directly outside of the `storage/remote` package.

storage/remote/queue_manager.go
```diff
@@ -375,12 +398,46 @@ func (t *QueueManager) Start() {
...
+	for _, record := range check.Checkpoints {
+		segmentNumber, err := strconv.Atoi(record.Segment)
+
+		if err != nil {
+			level.Error(t.logger).Log("err", err)
+			return false
+		}
+
+		// Segmentnumber out of bounds
+		if segmentNumber < wal.MinSegmentNumber || segmentNumber > wal.MaxSegmentNumber {
```
cstyan, line 425: We should maybe have the watcher have a `ValidSegment` function that uses `wal.Segments` to check if `segmentNumber` exists in the current WAL directory. Thoughts @csmarchbanks ?

roidelapluie:
hello @csmarchbanks @cstyan what is the status of this pull request? is this still valid/needed?

cstyan:
We can close this implementation. We'd still like this feature but no one has worked on this implementation path in quite some time.

PR #7712, opened 2020-08-01
scrape: fix typo.

johncming:
```
var err, appErr, scrapeErr error

level.Debug(sl.l).Log("msg", "Scrape failed", "err", scrapeErr.Error())
level.Debug(sl.l).Log("msg", "Append failed", "err", appErr)
```

No need to call `Error()` of `error` when using go-kit log.


brian-brazil:
Thanks!

PR #7716, opened 2020-08-02
promql: use explicit type declare instead of string.

johncming:
The type of `ValueTypeNone` and so on are `ValueType` actually.

https://github.com/prometheus/prometheus/blob/ff0ea1c1acbf4f108b692f818477b76c0f642cf8/promql/parser/value.go#L23-L32

Use `ValueType` explicitly is better than string.


brian-brazil:
Thanks!

PR #7718, opened 2020-08-02
promql: fix typo.

johncming:
Swap the result and bool order.


brian-brazil:
Thanks, though this is not a typo.

PR #7719, opened 2020-08-02
tsdb: add default logger for writeMetaFile.

johncming:
Add a default logger for writeMetaFile.

### DIFF
--- a/tsdb/block.go
+++ b/tsdb/block.go
@@ -203,6 +203,10 @@ func readMetaFile(dir string) (*BlockMeta, int64, error) {
 func writeMetaFile(logger log.Logger, dir string, meta *BlockMeta) (int64, error) {
 	meta.Version = metaVersion1
 
+	if logger == nil {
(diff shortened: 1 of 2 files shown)


roidelapluie:
Not sure it is worth it as this is a unexported function.

roidelapluie:
Yeah I would limit that to public functions.

johncming:
`logger` as a parameter should handle nil situation, function or method is a black box even if it's private.

It's easy and appropriate for a method to avoid `nil logger` as `logger` can be a field of the class. But way to function looks like tricky.

Another way in other projects use the package level logger may be a better way to handle nil logger rather than simply relying on conventions.

johncming:
> Checking for `nil` logger in _all_ methods looks does not smell that good to me. As @roidelapluie said, this is better to be limited only to public methods.

Yes, I agree with it now after review some code likes this. I will close the pr.

PR #7720, opened 2020-08-02
tsdb: add check for mint and maxt.

johncming:
Add a check for `mint` and `maxt`. The case that `mint` is bigger one leads to no samples created.


codesome:
After checking it again, I would like to take back my approval

> The case that mint is bigger one leads to no samples created.

This is fine (while it can be achieved by passing 0,0) and also required in some cases where we don't want samples. With the proposed change, it would create 0 series too which I don't think will be expected.

roidelapluie:
We have looked at this during the bug scrub and decided to close this as per above comments. Thanks!

PR #7721, opened 2020-08-02
promql: delete redundant return value.

johncming:
The return value of the method `initTime` in `Head` has no reference. 

https://github.com/prometheus/prometheus/blob/e6d7cc5fa43716281bdf5d1cf40415f99a0e9623/tsdb/head.go#L896-L898

https://github.com/prometheus/prometheus/blob/e6d7cc5fa43716281bdf5d1cf40415f99a0e9623/tsdb/head.go#L904

The relative PR is https://github.com/prometheus/prometheus/commit/4f037da4620f417c842f53dbf579e2357c43bdd7. The condition is no used in this pr.


brian-brazil:
Thanks!

Issue #6803, opened 2020-02-10
Scrape size metric?


gsstark:
We currently have a metric for scrape_duration but nothing for the number of bytes in a scrape. It's hard to tell what caused a change in the scrape duration if you can't tell whether the volume of data is changing along with it. It would also be really helpful to be able to recognize bandwidth limitations or calculate network capacity requirements.

roidelapluie:
We don't have such a number because prometheus will save the content of the scrape in its tsdb. You can use `{job="myjob"} unless {job="myjob"} offset 1h` to see which new metrics are added.
The last release candidate of prometheus has related metrics about the scrape cache that will help with this too.

gsstark:
I don't see how to calculate the byte size of the scrape from the number of metrics. It depends heavily on the size and number of labels, the size of the HELP texts, etc. Not all the metrics in the scrape will even be added if there's a filter. The size of the tsdb is an important thing to manage but it's orthogonal to having network issues due to scrape sizes.

brian-brazil:
scrape_samples_scraped should be enough of a proxy for that. Anything like this would be an additional sample be ingested on every single scrape by every single user, and I don't think this meets that bar.

roidelapluie:
We have looked at this during our bug scrub and decided to close this, as adding that metric for every scrape it is pretty expensive. We also have metrics around metadata cache if that's the issue you try to address.

Thanks!

PR #7722, opened 2020-08-02
Add scrape_bytes_scraped metric.

tomwilkie:
I was surprised to find we don't export the number of bytes scraped.  Or I missed it :-)

Not sure about (a) metric naming and (b) just using sl.lastScrapeSize - feedback welcome.

Signed-off-by: Tom Wilkie <tom.wilkie@gmail.com

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -1036,7 +1036,7 @@ func (sl *scrapeLoop) scrapeAndReport(interval, timeout time.Duration, last time
 	defer sl.buffers.Put(b)
 	buf := bytes.NewBuffer(b)
 
-	var total, added, seriesAdded int
+	var total, added, seriesAdded, bytes int
 	var err, appErr, scrapeErr error
 
 	app := sl.appender(sl.ctx)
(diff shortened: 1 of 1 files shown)


roidelapluie:
sl.lastScrapeSize would not work as we do not reset it when the scrape is dropped because of target_limit. Also, when the scrape is empty, we do not set sl.lastScrapeSize to 0, but we keep the old value.

tomwilkie:
Yes, that was why I didn't use it.

roidelapluie:
Also if I remember correctly that request was denied in the past as the number of samples scrapes should give a good hint about the size of the scrape. I try to find the discussion.

roidelapluie:
https://github.com/prometheus/prometheus/issues/6803

brian-brazil:
What is this for, and what exact operational problem is it trying to solve?

This metric increases the cost of every single scrape that every single Prometheus does, so there has to be a very strong use case for adding additional metrics of this nature.

roidelapluie:
Ping @tomwilkie  :) do you have further input?

roidelapluie:
We have looked at this during our bug scrub and decided to close this, as adding that metric for every scrape it is pretty expensive. We also have metrics around metadata cache if that's the issue you try to address.

Thanks!

roidelapluie:
This would actually be useful to debug #8762 where we do not know if there are big label values or lots of metrics, but we just want to find the biggest scrape target.

PR #7724, opened 2020-08-02
Consul: Reduce WatchTimeout to 2m and set it as timeout for requests

### LINKED ISSUE #7626: Consul SD Config concurrent connection issues
hum, there is an additional problem: Consul team did add a limitation on the number of concurrent requests by default in 1.6.2.

So, actually, the problem is more complex:
 * on a Consul instance with very high concurrent connections allowed, waiting 10m is probably the best: requests are blocking and values updated as fast as possible, you probably have 2k opened connections all the time to the agent if you are watching all the services at the same time, updates are instant
 * On a consul instance with default limited connections (aka 200 in 1.6.3+), if you watch more than 200 services for instances, some connections will be silently cut by Consul agent in version < 1.8 (see https://github.com/hashicorp/consul/issues/7527), in version > 1.8, it would return HTTP 429 (that helps quite a lot to detect the issue instead of a raw connection closed)

In the latter case, even with 30s instead of 10m, you had an issue, you had no idea when you would get updates because you could be rate-limited and thus some updates would be never performed, but when lucky (assuming the number of watches is not too big compared to number of allowed connections), it kinda work.

Probably, to make it plug and play:
 * Use long blocking queries when enough connections are allowed
 * When not enough connections are allowed, perform quick calls (not blocking) every n seconds to get the data (but you loose instant notifications)

@roidelapluie the long polling works well in my case for instance, because I allowed more connections on my agents that the number of watches, but the change from 10 minutes to 30 seconds is probably problematic for other people.

A possibility would be to add a new parameter (set by default to 200, maybe even better that could be read from attribute `HTTPMaxConnsPerClient` in agent's address in `http://address_of_cousul_agent/v1/agent/self`).
* When we watch more than HTTPMaxConnsPerClient => switch to instant polling every n seconds
* If we watch less than that, stay on blocking queries

Any opinion on that?

_Originally posted by @pierresouchay in https://github.com/prometheus/consul_exporter/pull/176#issuecomment-661807898_

roidelapluie:
#7423 raised the watch to 10m but the timeout was still 35 seconds.

This PR downgrades the watch to 2m and the timeout to 2m15s. The reason for NOT keeping 10m is that the client.Timeout applies to ALL the consul calls, even the /agent/self and the initial non-blocking calls.

Because of that, 2m is a good balance.

closes #7626

@beorn7 Please consider this for a 2.20.1. This is harmless for users as the discovery works BUT generates a LOT of logs and SD failure counter is incremented all the time.

### DIFF
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -36,7 +36,7 @@ import (
 )
 
 const (
-	watchTimeout  = 10 * time.Minute
+	watchTimeout  = 2 * time.Minute
 	retryInterval = 15 * time.Second
 
 	// addressLabel is the name for the label containing a target's address.
@@ -184,7 +184,7 @@ func NewDiscovery(conf *SDConfig, logger log.Logger) (*Discovery, error) {
 	}
 	wrapper := &http.Client{
 		Transport: transport,
-		Timeout:   35 * time.Second,
+		Timeout:   time.Duration(watchTimeout) + 15*time.Second,
 	}
 
 	clientConf := &consul.Config{
--- a/discovery/consul/consul_test.go
+++ b/discovery/consul/consul_test.go
@@ -223,20 +223,20 @@ func newServer(t *testing.T) (*httptest.Server, *SDConfig) {
 		switch r.URL.String() {
 		case "/v1/agent/self":
 			response = AgentAnswer
-		case "/v1/health/service/test?node-meta=rack_name%3A2304&stale=&tag=tag1&wait=600000ms":
+		case "/v1/health/service/test?node-meta=rack_name%3A2304&stale=&tag=tag1&wait=120000ms":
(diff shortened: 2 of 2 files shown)


roidelapluie:
cc @pierresouchay @iksaif

roidelapluie:
This is still WIP as I still have the bug

roidelapluie:
/v1/health/service/prom50?index=15&stale=&wait=120000ms (2m7.046164091s) from=127
.0.0.1:49360

Somehow the watch is not returning after 120000ms but prom coses the connection after 2m7s

roidelapluie:
I do not see higher response time than 2m+7s so 15 extra seconds should be good.

roidelapluie:
Users have reported that 2.20.0 makes consul unstable, which might be because golang does not clean connections which times out properly.

PR #7726, opened 2020-08-02
CircleCI: Fix nodejs cache poisoning

roidelapluie:
This wrong cache makes tests fail

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,6 @@ jobs:
     - restore_cache:
         keys:
         - v1-npm-deps-{{ checksum "web/ui/react-app/yarn.lock" }}
-        - v1-npm-deps-
     - run:
         command: make
         environment:


.circleci/config.yml
```diff
@@ -28,7 +28,6 @@ jobs:
     - restore_cache:
         keys:
         - v1-npm-deps-{{ checksum "web/ui/react-app/yarn.lock" }}
-        - v1-npm-deps-
```
juliusv, line 31: Interesting, in https://github.com/prometheus/prometheus/pull/6181#discussion_r338099899 @simonpasquier explicitly confirmed this line should be there, but I guess this can cause problems when we actually depend on slightly different packages.

juliusv:
:+1: I *think*, would be good to get confirmation by @simonpasquier

.circleci/config.yml
roidelapluie: What I see is tests fail:

```
o cache is found for key: v1-npm-deps-4FzFT0c3EEonV6JVriNMN0rzzFiM6RAWpqgTIHHJews=
Found a cache from build 36397 at v1-npm-deps-
Size: 56 MiB
Cached paths:
  * /home/circleci/project/web/ui/react-app/node_modules

Downloading cache archive...
Validating cache...

Unarchiving cache...
```

But 36397 is my update deps PR.

I don't know why the cache was not found (https://app.circleci.com/pipelines/github/prometheus/prometheus/8367/workflows/873e94be-223e-4fae-9daf-9f20348d10f1/jobs/36573) but I think we would better start fresh anyway.

roidelapluie: Especially since yarn does not seem to see that the dependencies are wrong.

roidelapluie:
#7727 might be a better fix

PR #7731, opened 2020-08-03
Pass groupName as context value in call to Appender

annanay25:
As a follow up action from this comment https://github.com/prometheus/prometheus/pull/7660#issuecomment-664317496, I'd like to propose the following change to pass `groupName` as a context value, in the call to `Appender()` from `group.run()`. 

This will enable us to choose group spec Appenders in the Cortex Ruler.

Signed-off-by: Annanay <annanayagarwal@gmail.com>

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -52,7 +52,11 @@ const (
 const namespace = "prometheus"
 
 // Typed string for passing group information to storage.Appendable
-const RuleGroupKey = "ruleGroup"
+type RuleGroupInfo string
+
+const (
+	RuleGroupKey RuleGroupInfo = "ruleGroup"
+)
 
 // Metrics for rule evaluation.
 type Metrics struct {
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -588,7 +588,7 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
 				numDuplicates = 0
 			)
 
-			app := g.opts.Appendable.Appender(ctx)
+			app := g.opts.Appendable.Appender(context.WithValue(ctx, "groupName", g.name))
 			seriesReturned := make(map[string]labels.Labels, len(g.seriesInPreviousEval[i]))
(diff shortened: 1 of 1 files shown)


roidelapluie:
Group names are not unique.

roidelapluie:
We already pass the file and the group to the promql engine. Can we reuse that or refactor it?

annanay25:
@roidelapluie - Thanks for pointing that out. 

In Prometheus, code flow will not be affected by this extra config key (we can choose to ignore it), and for our use case, groups with the same name can use the same appender. So it remains useful.

roidelapluie:
I'd still add it at the same place, because it is set once for all (and not each time we get an appender), and include the file so that we do not end up with a PR adding the file in the future, WDYT?

annanay25:
I think you're right @roidelapluie. :)

> We already pass the file and the group to the promql engine via context. Can we reuse that or refactor it?

We can make use of this key from the context. I will close this PR once I confirm that works for us.

roidelapluie:
> I think you're right @roidelapluie. :)
> 
> > We already pass the file and the group to the promql engine via context. Can we reuse that or refactor it?
> 
> We can make use of this key from the context. I will close this PR once I confirm that works for us.

You will not be able to do so as they use private types of the promql package.

roidelapluie:
My initial point about what about we simply make the rule manager instanciable multiple times so we don't need that in contexts at all was not answered; cortex could create one rule manager per Appendable.

annanay25:
Thanks @bwplotka I think that makes sense. I can change this to a typed string, which is a golang context-key recommendation as well.

annanay25:
> You will not be able to do so as they use private types of the promql package.

The other option is to make the promql key public. That would reduce adding duplicate information to the group context.

roidelapluie:
Closed by #7748

annanay25:
Thanks @roidelapluie and @bwplotka :)

PR #7737, opened 2020-08-04
Add tsdb startup duration metric

### LINKED ISSUE #7703: Metric for Prometheus startup time
## Proposal
When running large instances of Prometheus, startup time becomes a concern, even with many improvements done in recent versions instances can take 10m or more to load WAL and become ready/start scraping. Yet there is no metric to record this. Having it as a metric can help to see the effect of changes, optimizations, or environment growth and act on it.
I tried to calculate it somehow out of container_creation_time and container_ready metrics in K8S but wasn't successful.

mneverov:
Fixes #7703 

I created a new metric `prometheus_tsdb_startup_duration_seconds`.
Other possible solution may be exposing the [total_replay_duration](https://github.com/prometheus/prometheus/blob/master/tsdb/head.go#L723) as metric, since the time is spent mostly on WAL load. 

Signed-off-by: Maxim Neverov neverov.max@gmail.com

@bwplotka  please take a look

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -112,6 +112,7 @@ type headMetrics struct {
 	outOfOrderSamples        prometheus.Counter
 	walTruncateDuration      prometheus.Summary
 	walCorruptionsTotal      prometheus.Counter
+	walTotalReplayDuration   prometheus.Gauge
 	headTruncateFail         prometheus.Counter
 	headTruncateTotal        prometheus.Counter
 	checkpointDeleteFail     prometheus.Counter
@@ -169,6 +170,10 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 			Name: "prometheus_tsdb_wal_corruptions_total",
 			Help: "Total number of WAL corruptions.",
 		}),
+		walTotalReplayDuration: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "prometheus_tsdb_wal_total_replay_duration_seconds",
+			Help: "Duration of WAL to replay.",
+		}),
 		samplesAppended: prometheus.NewCounter(prometheus.CounterOpts{
 			Name: "prometheus_tsdb_head_samples_appended_total",
 			Help: "Total number of appended samples.",
@@ -224,6 +229,7 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 			m.gcDuration,
 			m.walTruncateDuration,
 			m.walCorruptionsTotal,
+			m.walTotalReplayDuration,
 			m.samplesAppended,
 			m.outOfBoundSamples,
 			m.outOfOrderSamples,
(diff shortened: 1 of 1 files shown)


roidelapluie:
Should'nt we leverage the timings used in the log messages we already have?

mneverov:
> Should'nt we leverage the timings used in the log messages we already have?

Fixed

tsdb/head.go
```diff
@@ -169,6 +170,10 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 			Name: "prometheus_tsdb_wal_corruptions_total",
 			Help: "Total number of WAL corruptions.",
 		}),
+		walTotalReplayDuration: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "prometheus_tsdb_wal_total_replay_duration_seconds",
```
codesome, line 174: Because it includes not just WAL, but also m-mapped chunks and will have chunks snapshot in future.
```suggestion
			Name: "prometheus_tsdb_wal_total_replay_duration_seconds",
```

tsdb/head.go
```diff
@@ -112,6 +112,7 @@ type headMetrics struct {
 	outOfOrderSamples        prometheus.Counter
 	walTruncateDuration      prometheus.Summary
 	walCorruptionsTotal      prometheus.Counter
+	walTotalReplayDuration   prometheus.Gauge
```
codesome, line 115: Should this be a `Summary`? Though it's observed only once. @roidelapluie

roidelapluie: There is only one observation, quantiles are not useful with one observation.

tsdb/head.go
roidelapluie: Did you forget your suggestion?

codesome: :facepalm:, I meant to suggest `prometheus_tsdb_data_replay_duration_seconds`

tsdb/head.go
codesome: Then gauge sounds fine. I was not sure about the usage :)

tsdb/head.go
```diff
@@ -169,6 +170,10 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 			Name: "prometheus_tsdb_wal_corruptions_total",
 			Help: "Total number of WAL corruptions.",
 		}),
+		walTotalReplayDuration: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "prometheus_tsdb_wal_total_replay_duration_seconds",
+			Help: "Duration of WAL to replay.",
```
codesome, line 175:
```suggestion
			Help: "Time taken to replay the data on disk.",
```

mneverov: Fixed, thanks

codesome (approved):
LGTM!

PR #7742, opened 2020-08-04
Do not specify go version in CONTRIBUTING.md

roidelapluie:
1. We set go1.14 in the go.mod so bad go version would yell
2. The way it is specified, its like it is only needed for dependencies.
   if we want the go version back in that file we should add it
   properly.


juliusv:
:+1:

PR #7743, opened 2020-08-04
Use faillint to ensure that no restricted imports are being used

jvrplmlmn:
This PR introduces the usage of `faillint` to ensure that no restricted imports are being used, this should be the last bit for https://github.com/prometheus/prometheus/issues/7647

I had to add `faillint` as a dependency for the project even though is only used for the dev toolchain and for Prometheus itself.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -187,6 +187,7 @@ common-vet:
 .PHONY: common-restricted_imports
 common-restricted_imports:
 	@echo ">> verifying that we are not using restricted packages"
+	$(GO) get github.com/fatih/faillint
 	faillint -paths "sync/atomic=go.uber.org/atomic" ./...
 
 .PHONY: common-lint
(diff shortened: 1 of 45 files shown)


jvrplmlmn:
PTAL @roidelapluie 🙏

roidelapluie:
mmh that go get is annoying

could we try golint standard?

https://github.com/golangci/golangci-lint/blob/master/.golangci.yml

the depguard thing

roidelapluie:
mmmh my initial test:

```
linters:
  enable:
  - depguard

linters-settings:
  depguard:
    list-type: blacklist
    packages:
      - sync/atomic
    include-go-root: true
    packages-with-error-message:
      - sync/atomic: use go.uber.org/atomic
```

roidelapluie:
@jvrplmlmn we are very interested by this. Can you implement it with depguard?

jvrplmlmn:
@roidelapluie thanks for the suggestion, I have tackled it on a separate PR since we are taking a different approach: https://github.com/prometheus/prometheus/pull/7825

PR #7713, opened 2020-08-01
Human-friendly durations in PromQL

roidelapluie:
This is the PromQL / documentation part of https://github.com/prometheus/common/pull/246

PR #7745, opened 2020-08-05
promql/parser: add testcases for lex.

johncming:
Add some test cases for new human-friendly durations(#7713)  in lex level parse.

### DIFF
--- a/promql/parser/lex_test.go
+++ b/promql/parser/lex_test.go
@@ -180,6 +180,18 @@ var tests = []struct {
 			}, {
 				input:    "1y",
 				expected: []Item{{DURATION, 0, "1y"}},
+			}, {
+				input:    "1h6ms",
+				expected: []Item{{DURATION, 0, "1h6ms"}},
+			}, {
+				input:    "10m5s",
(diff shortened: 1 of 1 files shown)


promql/parser/lex_test.go
```diff
@@ -180,6 +180,18 @@ var tests = []struct {
...
 				input:    "1y",
 				expected: []Item{{DURATION, 0, "1y"}},
+			}, {
+				input:    "1h6ms",
+				expected: []Item{{DURATION, 0, "1h6ms"}},
+			}, {
+				input:    "10m5s",
+				expected: []Item{{DURATION, 0, "10m5s"}},
+			}, {
+				input:    "10m5h",
```
brian-brazil, line 190: That this is lexed is an implementation detail, as it'll be rejected later on. I'd not test it here.

johncming: ok

roidelapluie:
We have looked at this from the bug scrub and decided to close this as this is covered at different levels already. Thanks!

PR #7746, opened 2020-08-05
storage/remote: fix concurrent access bug for pool.

johncming:
There is a bug about concurrent access.

## Bug detail

The race path of `TestIntern_MultiRef_Concurrent`
```go
	go interner.release(testString)

	interner.intern(testString)
```

Steps to reappear:
```go
// step 1
func (p *pool) intern(s string) string {
	if s == "" {
		return ""
	}

	p.mtx.RLock()
	interned, ok := p.pool[s]
	p.mtx.RUnlock()

// step 2
func (p *pool) release(s string) {
	p.mtx.RLock()
	interned, ok := p.pool[s]
	p.mtx.RUnlock()

	if !ok {
		noReferenceReleases.Inc()
		return
	}

// step3 in `intern`
	if ok {
                 // step 5: not running  until `release` finished.  <-- step 5 is here.
		interned.refs.Inc()
		return interned.s
	}

// step 4 in `release`
if !ok {
		noReferenceReleases.Inc()
		return
	}

	refs := interned.refs.Dec()
	if refs > 0 {
		return
	}

	p.mtx.Lock()
	defer p.mtx.Unlock()
	if interned.refs.Load() != 0 {
		return
	}
	delete(p.pool, s)
```

Then `TestIntern_MultiRef_Concurrent` failed at 
https://github.com/prometheus/prometheus/blob/b58a61344334651f0000f9d467e76066a9ea998f/storage/remote/intern_test.go#L85

## Tests

It's hard to reproduce this.  When I look at the ci result of pr #7745, I find the error https://app.circleci.com/pipelines/github/prometheus/prometheus/8443/workflows/7cf94e94-4776-4762-aabf-4e4c3d2712f3/jobs/36890.

It's not easy to write a test case if not modify `pool` implementation. So I send this pr without updating the  test case. 
Some tool may be useful like 
https://github.com/etcd-io/gofail
https://github.com/pingcap/failpoint

Inject a fail point to trigger a delay operation.

### DIFF
--- a/storage/remote/intern.go
+++ b/storage/remote/intern.go
@@ -61,13 +61,6 @@ func (p *pool) intern(s string) string {
 		return ""
 	}
 
-	p.mtx.RLock()
-	interned, ok := p.pool[s]
-	p.mtx.RUnlock()
-	if ok {
-		interned.refs.Inc()
-		return interned.s
-	}
 	p.mtx.Lock()
 	defer p.mtx.Unlock()
 	if interned, ok := p.pool[s]; ok {


storage/remote/intern.go
```diff
@@ -61,13 +61,6 @@ func (p *pool) intern(s string) string {
 		return ""
 	}
 
-	p.mtx.RLock()
```
csmarchbanks, line 64: Rather than removing the optimization to allow concurrent calls to `intern`, all that needs to be done to fix this is to not call `p.mtx.RUnlock()` until after `interned.refs.Inc()`. Though both branches would then need the unlock line.

roidelapluie:
@johncming can you please address the comments? Thanks!

csmarchbanks:
@johncming :wave: are you interested in updating this PR with the above comments to get it merged?

roidelapluie:
cc @csmarchbanks is this pull request valid? should I rebase it and fix it ?

cstyan:
@roidelapluie it looks like a relevant fix can still be made even on the same line #'s reported here, the code in this file hasn't been changed since 2019/2020

beorn7:
:wave: from the bug scrub. With no update by the original contributor, we would like to ask @cstyan to do the fix in a separate PR (or by updating this one). @cstyan could you do this?

cstyan:
@beorn7 I will confirm this week via a test whether we still need a change here

bboreham:
Hello from the bug-scrub!

Since this PR was created, a few things have changed.
In regular Prometheus, [this code is no longer called](https://github.com/prometheus/prometheus/blob/ad4857de5210bb1756b46c9ae6e653c72cc7654d/model/labels/labels_stringlabels.go#L442-L443), since Labels use different storage methods.

It will be called from projects vendoring the Prometheus code and using different build tags, but in at least one place this is found to be too expensive; @cstyan has a PR to disable interning: #14750.

Lastly, although it is unintended, the extent of the bug is that a string may get copied one extra time, which does not seem a huge problem.

Closing due to lack of movement over years.

PR #7747, opened 2020-08-05
storage/remote: split httpclient name.

johncming:
Split the origin name `remote_storage ` to two parts.

## Modification 

The `remote_storage_read_client` for remote read http client and the `remote_storage_write_client ` for remote write http client.

## Benefit

Two split names are better tracking for metrics `net_contrack_dialer_conn_attempted_total`.
https://github.com/prometheus/prometheus/blob/6a633eece19a782cfeddac435b60c1a7abc2488a/vendor/github.com/mwitkow/go-conntrack/dialer_reporter.go#L25-L31

### DIFF
--- a/storage/remote/client.go
+++ b/storage/remote/client.go
@@ -104,7 +104,7 @@ type ReadClient interface {
 
 // newReadClient creates a new client for remote read.
 func newReadClient(name string, conf *ClientConfig) (ReadClient, error) {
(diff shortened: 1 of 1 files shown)


cstyan:
@csmarchbanks @bwplotka any opinions on this change?

brian-brazil:
:+1: 

The purpose of this name is to be able to track what our different uses of HTTP are, plus spot things like connection leaks.

PR #7749, opened 2020-08-05
Gitpod config

svenefftinge:
This PR adds configuration for Gitpod to this repo. 
It allows anyone to quickly spin up fresh prebuilt dev environment in a browser for any branches, issues and pull requests.
I have configured it to prebuild your branches using (i.e. the init task in .gitpod.yml)
```
make build
```
On start it launches prometheus with the example config 
```
./prometheus --config.file=documentation/examples/prometheus.yml
```
in one terminal and the web UI in a separate one:

```
cd web/ui/react-app
yarn start
```

I hope the config is a good start. Please propose any changes to align better with your daily routines and use cases.

To get the prebuilds, you also need to install the [GitHub app](https://github.com/apps/gitpod-io) on this project. Without the app the config works, but users will have to wait for the init task (`make build`). 

You can try the experience on my fork https://gitpod.io/#https://github.com/svenefftinge/prometheus
Or even review this PR using gitpod by prefixing the URL of this page with `gitpod.io#`.

Screenshot 2020-07-26 at 11 42 40

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,18 @@
+tasks:
+  - init: 
+      make build
+    command: |
+      gp sync-done build
+      ./prometheus --config.file=documentation/examples/prometheus.yml
+  - command: |
+      cd web/ui/react-app
+      gp sync-await build
+      unset BROWSER
+      export DANGEROUSLY_DISABLE_HOST_CHECK=true
+      yarn start
+    openMode: split-right
+ports:
+  - port: 3000
+    onOpen: open-preview
+  - port: 9090
+    onOpen: ignore
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 [![Go Report Card](https://goreportcard.com/badge/github.com/prometheus/prometheus)](https://goreportcard.com/report/github.com/prometheus/prometheus)
 [![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/486/badge)](https://bestpractices.coreinfrastructure.org/projects/486)
 [![fuzzit](https://app.fuzzit.dev/badge?org_id=prometheus&branch=master)](https://fuzzit.dev)
+[![Gitpod ready-to-code](https://img.shields.io/badge/Gitpod-ready--to--code-blue?logo=gitpod)](https://gitpod.io/#https://github.com/prometheus/prometheus)
 
 Visit [prometheus.io](https://prometheus.io) for the full documentation,
 examples and guides.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,6 +28,8 @@ Should you wish to work on an issue, please claim it first by commenting on the 
 
 Please check the [`low-hanging-fruit`](https://github.com/prometheus/prometheus/issues?q=is%3Aissue+is%3Aopen+label%3A%22low+hanging+fruit%22) label to find issues that are good for getting started. If you have questions about one of the issues, with or without the tag, please comment on them and one of the maintainers will clarify it. For a quicker response, contact us over [IRC](https://prometheus.io/community).
 
+You can [spin up a prebuilt dev environment](https://gitpod.io/#https://github.com/prometheus/prometheus) using Gitpod.io.
+
 For complete instructions on how to compile see: [Building From Source](https://github.com/prometheus/prometheus#building-from-source)
 
 For quickly compiling and testing your changes do:


.gitpod.yml
```diff
@@ -0,0 +1,20 @@
...
+      export DANGEROUSLY_DISABLE_HOST_CHECK=true
+      yarn start
+    openMode: split-right
+ports:
+  - port: 3000
+    onOpen: open-preview
+  - port: 9090
+    onOpen: ignore
+  - port: 9091
+    onOpen: ignore
```
juliusv, line 20: nit: please end all files with a line break

.gitpod.yml
```diff
@@ -0,0 +1,20 @@
...
+      unset BROWSER
+      export DANGEROUSLY_DISABLE_HOST_CHECK=true
+      yarn start
+    openMode: split-right
+ports:
+  - port: 3000
+    onOpen: open-preview
+  - port: 9090
+    onOpen: ignore
+  - port: 9091
```
juliusv, line 19: Port 9091 would be the Pushgateway component usually, but we are not starting it here. Is this entry needed?

juliusv:
Looks pretty awesome besides small comments, thanks!

README.md
```diff
@@ -1,5 +1,6 @@
 # Prometheus
 
+[![Gitpod ready-to-code](https://img.shields.io/badge/Gitpod-ready--to--code-blue?logo=gitpod)](https://gitpod.io/#https://github.com/prometheus/prometheus)
```
brian-brazil, line 3: This should go at the end of the list, it's not more important than our CI, downloads etc.

juliusv: Yeah, I was on the fence about that one. You could argue it's more useful for actually clicking on it more often, but yeah, maybe it should be at the end for now.

.gitpod.yml
svenefftinge: I'm listing it here so users don't get a notification when the port is served. Same for 9090.

juliusv: But we don't expect anything to listen on this port, right? Just not sure why we specifically include port 9091, a port which does not occur in Prometheus anywhere.

svenefftinge: I do not remember which process it was, but there was something that opens on port 9091. Searching your repo at least there are various mentions of that port. https://github.com/prometheus/prometheus/search?q=9091&unscoped_q=9091
But I will remove it for now, as you can always add it again later.

juliusv: Yeah, those are mostly tests or examples/docs, probably because the Prometheus Pushgateway uses port 9091, but it's a completely independent process. If we included all the ports of the other Prometheus components, we'd have to include a large list of ports :) So yeah, best to omit it here.

juliusv:
:+1: Looks good now, will merge after CI goes green

svenefftinge:
@juliusv you should enable the GitHub app, so that prebuilds work (i.e. users don't need to wait for the init tasks).
Ideally you do that before the merge, so a prebuild runs on the master branch after merge already.
https://github.com/apps/gitpod-io

Let me know if you have questions or any issues. Also feel free to ask anything at https://community.gitpod.io or file issues over at https://github.com/gitpod-io/gitpod

juliusv:
@svenefftinge I just wanted to do that, and initially GitHub just allowed me to grant read access to a single repository (prometheus/prometheus), and write access to its commit status checks. But then on the next page off gitpod.io to install the app, I got this:

gitpod_access

If I understand this correctly, Gitpod wants to be able to read and write all repository data, including writing code, for all my repos, including private ones ("This application will be able to read and write all public and private repository data.").

Is there a way to restrict it to the permissions I granted on the first screen (read prometheus/prometheus, be able to set CI status)? I don't think I want to give it any further write access at this point.

juliusv:
Oh actually it seems it won't be able to access private repos without me granting that specifically for each private repo, but the point about being able to write to the public repos still stands.

svenefftinge:
If you add the approriate configuration, the app can also update pull request comments with a button.
By default (i.e. the config from this PR) the app will only checkout the code and update a status on PRs.
I fear we don't have fine grained scopes for the app ATM as we have for users.

juliusv:
@svenefftinge Ah ok, I wasn't aware of that, thanks. The problem is, security-wise I'm not sure the Prometheus team is ok with giving Gitpod write access to our code (and apparently my other orgs as well). Do you think there's going to be a more fine-grained approach soon, or is that far out?

svenefftinge:
You already provide Gitpod with the ability to push commits if you are using it (through oauth). For the app the difference is only that any writing action would be on behalf of the Gitpod app (authorized by you) instead of your behalf (oauth). Also FWIW tokens are stored in an encrypted fashion in our DB. 
We plan to allow an integration for prebuilds through a regular webhook, but that is currently only support in GitLab. It will probably take a couple of month until that lands.

juliusv:
@svenefftinge So my hope (or initial naive assumption) was that for our purposes we could get away with read permissions and CI-status-only write permissions, and that any user starting to actually contribute PRs back via Gitpod would be individually able to give Gitpod write permissions to e.g. their fork. OTOH now when I think back to what you said about being able to review PRs via Gitpod (which would be optional for us), then of course that would at least need more than CI write permissions.

I just want to gather one or two more team opinions on whether the permissions above in my screenstho are ok to give out, /cc @roidelapluie @brian-brazil and others.

brian-brazil:
I don't think we should be giving repo write permissions out to anything other than our CI (which needs them for releases). It should be able to work from what's already publicly readable.

juliusv:
If I understand the above correctly (correct me if I'm wrong, I'm not very fluent in GitHub permissions), what I would be allowing is also not a direct permissions relationship between Gitpod and Prometheus, but between me as a GitHub user and Gitpod, to allow Gitpod to write into any of my repos, so to basically impersonate me. I assume this because the permissions warnings lists my orgs, and is not Prometheus-org-specific. And Prometheus just happens to be one of the orgs I have access to. That would mean that we would have to forbid any Prometheus maintainers from using Gitpod for anything (even outside of Prometheus) if we wanted to prevent write access for Gitpod? Also, at this stage I wouldn't really want to tie the CI access to me as a user, but to the org only.

brian-brazil:
> That would mean that we would have to forbid any Prometheus maintainers from using Gitpod for anything (even outside of Prometheus) if we wanted to prevent write access for Gitpod? 

That not something we have done in the past, though some -team members are very careful with such things individually.

svenefftinge:
FWIW you can use gitpod without the prebuild app, of course. The downside is that users will have to wait for the init task when opening an environment. So if you are happy with adding the config, you can just do and add the prebuild integration later.

juliusv:
Ok, I'll just merge it like it is without enabling the app for now, as adding write permissions will need more discussion with the team.

The point that @roidelapluie brought up on the mailing list about GitHub Codespaces is also one we need to consider in the future. On one hand it's convenient (and probably avoids write-access issues) if that is available through GitHub directly, on the other hand it always makes me a bit sad that big platforms monopolize everything.

Issue #7097, opened 2020-04-07
support scrape loop graceful reload 


Betula-L:
**How to reproduce it?**

- Edit sample_limit from `10000` to `100001`
or
- Edit scrape_timeout from `30s` to `1m`

Trigger reload, then 30 of 3000 instances report `down` because of `lastError="Get http://10.69.140.12:9100/metrics: context canceled"`

**What did you expect to see?**

`Reload` will cancel involved current `scrapeLoop`, and prometheus mark this `scrapeLoop` as down. 

We can not distinguish whether instances were really down or just misreported by reload now.

**Possible solution 1:**
set `up` value as 'NAN' if `context canceled`

**Possible solution 2:**
do not record `up` if `context canceled`

**Environment**

* Prometheus version:

prometheus, version 2.16.0 (branch: HEAD, revision: b90be6f32a33c03163d700e1452b54454ddce0ec)
  build user:       root@7ea0ae865f12
  build date:       20200213-23:50:02
  go version:       go1.13.8

* Prometheus configuration file:

```
  - job_name: node_exporter
    honor_labels: true
    honor_timestamps: true
    scrape_interval: 1m
    scrape_timeout: 33s
    metrics_path: /metrics
    scheme: http
    sample_limit: 100001
```

brian-brazil:
Reporting up as 0 is correct, as the scrape failed. I'm not sure that we should be cancelling contexts on reload though, that's for shutdown.

Betula-L:
We can not distinguish whether instances were really down or just misreported by reload now.

brian-brazil:
Your monitoring generally should be tolerant to a single failed scrape, as that can happen for all sorts of reasons.

Betula-L:
Yes, you are right. 

However, in some sensitive condition, especially critical service, there should no unexpected dirty data by reload.

I think it is not expected behavior, does it?

brian-brazil:
It needs some investigation, which is why I haven't closed the issue.

Betula-L:
Got it, thanks.

Betula-L:
Current scrape loop reload step:
1. create new loop
2. stop old loop immediately (cancel all current scrape)
3. run new loop
4. return reload success

However, "cancel immediately" leads to two side effects:
1. misreport "up" metric due to reload cancel
2. prometheus do not scrape for a while after reload because random scraping offset.

Thus, it's better that stop old scrape loops gracefully.

Betula-L:
[...]

PR #7752, opened 2020-08-05
Do not stop scrapes in progress during reload

roidelapluie:
See #7097

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -1022,6 +1022,8 @@ mainLoop:
 // scrapeAndReport performs a scrape and then appends the result to the storage
 // together with reporting metrics, by using as few appenders as possible.
 // In the happy scenario, a single appender is used.
+// This function uses sl.parentCtx instead of sl.ctx on purpose. A scrape should
+// only be cancelled on shutdown, not on reloads.
 func (sl *scrapeLoop) scrapeAndReport(interval, timeout time.Duration, last time.Time, errc chan<- error) time.Time {
 	start := time.Now()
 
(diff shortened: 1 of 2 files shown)


roidelapluie:
@brian-brazil It is going to be difficult to test this one but I have tested it locally. I can reproduce the issue and see that this PR fixes it.

prometheus.yml:

```
# my global config
global:

# Alertmanager configuration
alerting:
  alertmanagers:
  - static_configs:
    - targets:
      # - alertmanager:9093

rule_files:
  # - "first_rules.yml"
  # - "second_rules.yml"

scrape_configs:
  - job_name: 'prometheus'

    # scheme defaults to 'http'.
    sample_limit: 10000

    file_sd_configs:
      - files: [all.yml]
```

all.yml https://gist.github.com/roidelapluie/2b0a0e30b66443741cd5b26eb07f9e66

roidelapluie:
@brian-brazil I am torn on this one.

I'd like to consider hard-cutting the scrape but skipping the stale markers and the registration of the error in the UI.

This PR would slow down reloads a lot. I could not fine a release of prometheus which would act differently, so it is probably like this for a long time.

brian-brazil:
Creating a goroutine per scrape does not seem wise. I also think this is a bit simpler, shouldn't it be the parentCtx that we're passing into the scrape?

brian-brazil:
That looks about right, can you add a unittest? Triggering the cancel from inside the scrape should do it.

roidelapluie:
> That looks about right, can you add a unittest? Triggering the cancel from inside the scrape should do it.

I have adapted the current test.

The modified test fails on master with:

--- FAIL: TestScrapeLoopStop (0.03s)                                 
    scrape_test.go:582: Expected at least 3 scrapes with 6 samples each, got 17 samples

On master we indeed get 17 samples instead of 18 due to the failed scrape.

brian-brazil:
:+1:

PR #7758, opened 2020-08-06
Check grouping

krasi-georgiev:
I think this would be on the heap so not sure if the complicator would detect empty grouping and skip the looping over all metrics


brian-brazil:
This is a micro-optimisation that applies in a case where the impact is going to be very tiny. In general I don't worry about constant costs for a query (which this is) as cardinality is the problem - not fixed costs per node evaluation.

This also makes the code harder to grok, and given this saves only 16B of allocations I don't think it makes sense.

PR #7767, opened 2020-08-07
Use strict unmarshalling of rules

### LINKED ISSUE #7042: `promtool check rules` no longer identifies invalid fields as of 2.16.0

roidelapluie:
Fix #7042

The problem is not only in promtool. Prometheus successfully load broken files now, which happened to me today. It loaded a rule file with annotation and not annotations.

The previous PR #7142 seems to have a larger scope, let's move on with this one and see what's left in #7142.

### DIFF
--- a/pkg/rulefmt/rulefmt.go
+++ b/pkg/rulefmt/rulefmt.go
@@ -16,6 +16,7 @@ package rulefmt
 import (
 	"bytes"
 	"context"
+	"io"
 	"io/ioutil"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


brian-brazil:
This seems like a better approach, there's a test failure though.

pkg/rulefmt/rulefmt.go
```diff
@@ -267,8 +269,11 @@ func Parse(content []byte) (*RuleGroups, []error) {
 		errs   []error
 	)
 
-	err := yaml.Unmarshal(content, &groups)
-	if err != nil {
+	decoder := yaml.NewDecoder(bytes.NewReader(content))
+	decoder.KnownFields(true)
+	err := decoder.Decode(&groups)
+	// Ignore io.EOF which happends with empty input.
```
brian-brazil, line 275: happens

Issue #7640, opened 2020-07-22
Discovery: data race in CI


roidelapluie:
```
=== RUN   TestIdenticalConfigurationsAreCoalesced
==================
WARNING: DATA RACE
Write at 0x00c0004a4e40 by goroutine 19:
  runtime.mapassign_faststr()
      /usr/local/go/src/runtime/map_faststr.go:202 +0x0
  github.com/prometheus/prometheus/discovery.(*Manager).updateGroup()
      /home/circleci/project/discovery/manager.go:303 +0x26f
  github.com/prometheus/prometheus/discovery.(*Manager).updater()
      /home/circleci/project/discovery/manager.go:249 +0xfe

Previous read at 0x00c0004a4e40 by goroutine 15:
  runtime.mapiternext()
      /usr/local/go/src/runtime/map.go:846 +0x0
  github.com/prometheus/prometheus/discovery.verifyPresence()
      /home/circleci/project/discovery/manager_test.go:734 +0x2e2
  github.com/prometheus/prometheus/discovery.TestIdenticalConfigurationsAreCoalesced()
      /home/circleci/project/discovery/manager_test.go:907 +0xb1c
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1039 +0x1eb

Goroutine 19 (running) created at:
  github.com/prometheus/prometheus/discovery.(*Manager).startProvider()
      /home/circleci/project/discovery/manager.go:233 +0x603
  github.com/prometheus/prometheus/discovery.(*Manager).ApplyConfig()
      /home/circleci/project/discovery/manager.go:208 +0x8b6
  github.com/prometheus/prometheus/discovery.TestIdenticalConfigurationsAreCoalesced()
      /home/circleci/project/discovery/manager_test.go:904 +0xa60
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1039 +0x1eb

Goroutine 15 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:1090 +0x700
  testing.runTests.func1()
      /usr/local/go/src/testing/testing.go:1334 +0xa6
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1039 +0x1eb
  testing.runTests()
      /usr/local/go/src/testing/testing.go:1332 +0x527
  testing.(*M).Run()
      /usr/local/go/src/testing/testing.go:1249 +0x43f
  main.main()
      _testmain.go:56 +0x223
==================
==================
[...]

PR #7769, opened 2020-08-09
Use temporary directory in test to avoid data race

mneverov:
Fixes: #7640 

The race happens because test uses the private manager targets field to [verify pool key presence](https://github.com/prometheus/prometheus/blob/master/discovery/manager_test.go#L910). All operations with this field are guarded [with the mutex](https://github.com/prometheus/prometheus/blob/master/discovery/manager.go#L150) and this invariant is violated in the test: there is no mutexes used.
But, there should not be any concurrent operations with the `targets`.
Because prometheus watches [not files, but a directory](https://github.com/prometheus/prometheus/blob/master/discovery/file/file.go#L205), and the test creates a config file directly in [/tmp](https://github.com/prometheus/prometheus/blob/master/discovery/manager_test.go#L862) it is possible to get events for the other test directories created in `/tmp` in other tests when run with `-race`.
By creating a directory devoted for this test the events from other tests should go away.

The other solution might be is to [watch over config files](https://github.com/prometheus/prometheus/blob/master/discovery/file/file.go#L215), not directories.

@roidelapluie could you please take a look?

### DIFF
--- a/discovery/manager_test.go
+++ b/discovery/manager_test.go
@@ -859,11 +859,15 @@ func TestTargetSetRecreatesEmptyStaticConfigs(t *testing.T) {
 }
 
 func TestIdenticalConfigurationsAreCoalesced(t *testing.T) {
-	tmpFile, err := ioutil.TempFile("", "sd")
+	tmpDir, err := ioutil.TempDir("", "prometheus-cfg-coalesced")
+	if err != nil {
+		t.Fatalf("error creating temporary directory: %v", err)
+	}
+	defer os.RemoveAll(tmpDir)
+	tmpFile, err := ioutil.TempFile(tmpDir, "sd")
 	if err != nil {
 		t.Fatalf("error creating temporary file: %v", err)
 	}
-	defer os.Remove(tmpFile.Name())
 	if _, err := tmpFile.Write([]byte(`[{"targets": ["foo:9090"]}]`)); err != nil {
 		t.Fatalf("error writing temporary file: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


discovery/manager_test.go
```diff
@@ -859,11 +859,15 @@ func TestTargetSetRecreatesEmptyStaticConfigs(t *testing.T) {
 }
 
 func TestIdenticalConfigurationsAreCoalesced(t *testing.T) {
-	tmpFile, err := ioutil.TempFile("", "sd")
+	tmpDir, err := ioutil.TempDir("", "prometheus-cfg-coalesced")
+	if err != nil {
```
roidelapluie, line 863:
```suggestion
	testutil.Ok(t, err)
```

discovery/manager_test.go
```diff
@@ -859,11 +859,15 @@ func TestTargetSetRecreatesEmptyStaticConfigs(t *testing.T) {
 }
 
 func TestIdenticalConfigurationsAreCoalesced(t *testing.T) {
-	tmpFile, err := ioutil.TempFile("", "sd")
+	tmpDir, err := ioutil.TempDir("", "prometheus-cfg-coalesced")
+	testutil.Ok(t, err)
+		t.Fatalf("error creating temporary directory: %v", err)
```
roidelapluie, line 864: then this needs to go and the next line too

mneverov: thanks!
Replaced err check with the testutil in the test

roidelapluie:
Can you rebase this? thanks

mneverov:
@roidelapluie I think this is obsolete after #7701

roidelapluie:
We have looked at this during the bug scrub and decided to close this as per previous comment. Thanks!

PR #7771, opened 2020-08-10
Add metadata to WAL, expose to WAL reader and send with remote write

robskillington:
Mailing list discussion: 
https://groups.google.com/g/prometheus-developers/c/w-eotGenBGg/m/UZWdxzcBBwAJ

Design doc:
https://docs.google.com/document/d/1LY8Im8UyIBn8e3LJ2jB-MoajXkfAqW2eKzY735aYxqo/edit#heading=h.bik9uwphqy3g

Have tested this end-to-end with a [modified remote write backend](https://github.com/m3db/m3/compare/r/test-prometheus-metadata?expand=1):
> {"level":"info","ts":1597033222.2789593,"msg":"received series","labels":"{__name__=\"prometheus_rule_group_iterations_total\",instance=\"localhost:9090\",job=\"prometheus01\",role=\"remote\"}","type":"counter","unit":"","help":"The total number of scheduled rule group evaluations, whether executed or missed."}

Background:
There are already efforts such as PR #6815 and the [corresponding design document](https://docs.google.com/document/d/1LoCWPAIIbGSq59NG3ZYyvkeNb8Ymz28PUKbg_yhAzvE/edit#) that is experimenting with syncing the entire set of metadata kept by a Prometheus server for the current active targets in memory every 15 seconds (mentioned by the doc). It is great that we can unblock experiments like this going forward and foster development, but there are some that would like to see a slightly longer term solution that is by design not sync based and can be sure to deliver no loss of any metadata in the regular case of a remote write tails the Prometheus WAL.

TODO:
- [ ] Tests.


roidelapluie:
Is this backwards compatible? Or could that be enabled with a flag?

This is a risky one, no matter the effort we put in, and I'd like people to be able to downgrade in case of issues without deleting their WAL. WDYT?

prompb/types.proto
```diff
@@ -27,6 +27,9 @@ message Sample {
 message TimeSeries {
   repeated Label labels   = 1 [(gogoproto.nullable) = false];
   repeated Sample samples = 2 [(gogoproto.nullable) = false];
+  string type = 3;
```
brian-brazil, line 30: type should be an enum

scrape/scrape.go
```diff
@@ -1408,6 +1440,30 @@ const (
...
+		Type: textparse.MetricTypeGauge,
+		// TODO: add help based on advice.
+	}
+	scrapeDurationMetricMeta = storage.Metadata{
+		Type: textparse.MetricTypeGauge,
+		Unit: "seconds",
+		// TODO: add help based on advice.
+	}
+	scrapeSamplesMetricMeta = storage.Metadata{
+		Type: textparse.MetricTypeCounter,
```
brian-brazil, line 1454: These are all gauges

scrape/scrape.go
```diff
@@ -805,11 +812,17 @@ func (c *scrapeCache) get(met string) (*cacheEntry, bool) {
 	return e, true
 }
 
-func (c *scrapeCache) addRef(met string, ref uint64, lset labels.Labels, hash uint64) {
+func (c *scrapeCache) addRef(met string, ref uint64, lset labels.Labels, meta storage.Metadata, hash uint64) {
```
brian-brazil, line 815: Passing structs around is inefficient, use a pointer.

scrape/scrape.go
```diff
@@ -1272,7 +1295,7 @@ loop:
 		ce, ok := sl.cache.get(yoloString(met))
 
 		if ok {
-			err = app.AddFast(ce.ref, t, v)
+			err = app.AddFast(ce.ref, ce.meta, t, v)
```
brian-brazil, line 1298: What if this is from a previous scrape, and this scrape doesn't have metadata?

scrape/scrape.go
```diff
@@ -714,13 +715,19 @@ type scrapeCache struct {
 	// seriesCur and seriesPrev store the labels of series that were seen
 	// in the current and previous scrape.
 	// We hold two maps and swap them out to save allocations.
-	seriesCur  map[uint64]labels.Labels
-	seriesPrev map[uint64]labels.Labels
+	seriesCur  map[uint64]seriesEntry
```
brian-brazil, line 718: Putting this in here feels weird, why is staleness being mixed with metadata?

Why is the cache entry no enough?

tsdb/wal.go
```diff
@@ -1143,6 +1169,31 @@ func (r *walReader) decodeSeries(flag byte, b []byte, res *[]record.RefSeries) e
 	return nil
 }
 
+func (r *walReader) decodeMetadata(flag byte, b []byte, res *[]record.RefMetadata) error {
+	dec := encoding.Decbuf{B: b}
+
+	for len(dec.B) > 0 && dec.Err() == nil {
+		ref := dec.Uvarint64()
```
brian-brazil, line 1176: The docs need updating for new formats.

tsdb/wal.go
```diff
@@ -102,6 +104,7 @@ type WAL interface {
 type WALReader interface {
 	Read(
 		seriesf func([]record.RefSeries),
+		metadataf func([]record.RefMetadata),
```
brian-brazil, line 107: Does checkpointing need updating for the new data?

tsdb/head.go
```diff
@@ -1137,6 +1141,23 @@ func (a *headAppender) AddFast(ref uint64, t int64, v float64) error {
 	if s == nil {
 		return errors.Wrap(storage.ErrNotFound, "unknown series")
 	}
+
+	// Without string interning / using symbol table / using siginficantly
+	// more memory, determine whether the metadata has changed for this entry
+	// using a hash of the metadata.
+	a.digest.Reset()
```
brian-brazil, line 1148: This sounds like it's going to be quite expensive for what is the most important hot path in all of Prometheus. We've also had quite a few problems in the past with hashes and collisions. If we're going to compare, let's do it literally.

Or leave it up to the scrape, so using Add implies new metadata and AddFast is always fast.

roidelapluie:
> > I'd like people to be able to downgrade in case of issues without deleting their WAL. WDYT?
> 
> We don't promise that you can downgrade, only that you can upgrade.

I am well aware of that. but if it comes at low cost, I'd still like that. People sometimes downgrade for many different reasons. Telling them that the upgrade will not enable them to downgrade will slow the adoption of newer releases.

scrape/scrape.go
robskillington: So the main reason this was required is that we write a "stale" datapoint for a series using appender "Add(...)" which requires the full labels, metadata, etc to need on hand when we call that appender method from within "forEachStale(...)".

prompb/types.proto
robskillington: Sounds good, will update.

scrape/scrape.go
robskillington: Sounds good, will update.

scrape/scrape.go
robskillington: Sounds good, will do.

scrape/scrape.go
robskillington: I was assuming perhaps that it would keep the metadata it previously declared? Or should it "zero" the metadata instead?

tsdb/wal.go
robskillington: Sounds good, will update.

tsdb/head.go
robskillington: That's fair, are you open to changing the appender interface to the following:

```go
Add(l labels.Labels, m Metadata, t int64, v float64) (uint64, error)
AddFast(ref uint64, int64, v float64) error
AndFastAndUpdate(ref uint64, m Metadata, t int64, v float64) error
```

Then we can call `AndFastAndUpdate` if and only if the metadata has been updated (which can be done from the scrape loop by doing literal comparison).

robskillington: Caught up about this, seems like we can just revert to calling `Add(...)` if the metadata has changed between scrape iterations and remove the `AddFastAndUpdate(...)` concept.

scrape/scrape.go
robskillington: Do you think we should revert to zero-ing the metadata if a scrape is stale? Or calling "Add(...)" with previously observed metadata is correct here? (in the calls to ".Add(...)" from within ".forEachStatel(...)" callback)

tsdb/wal.go
robskillington: Ah yes, good catch. Updated (still pending tests though).

tsdb/wal.go
robskillington: Updated.

robskillington:
@roidelapluie unfortunately this is not possible, basically any new WAL record type added to the WAL will cause previous versions to return "Invalid" when they read that entry type. This is due to the decoder `Type()` method which already exists on current versions (which are out in the wild already):
https://github.com/prometheus/prometheus/blob/2b75c1b19995377f2ffcea7aff6818982f2f2bc6/tsdb/record/record.go#L66-L75

scrape/scrape.go
brian-brazil: The metadata is still correct if a scrape fails, only if there is new (or missing) metadata should Add need to get called again.

brian-brazil:
> This is due to the decoder Type() method which already exists on current versions (which are out in the wild already):

We could adjust this now to be more relaxed (catching 0 as an error is likely enough in practice), and give it a release.

tsdb/docs/format/wal.md
```diff
@@ -50,6 +50,27 @@ Series records encode the labels that identifies a series and its unique ID.
...
+│ type = 1 <1b>                              │
+├────────────────────────────────────────────┤
+│ ┌────────────────────────────────────────┐ │
+│ │ id <uvarint>                           │ │
+│ ├─────────────────────┬──────────────────┤ │
+│ │ len(type) <uvarint> │ type <bytes>     │ │
+│ ├─────────────────────┴──────────────────┤ │
+│ │ len(unit) <uvarint> │ unit <bytes>     │ │
+│ ├─────────────────────┴──────────────────┤ │
+│ │ len(help) <uvarint> │ help <bytes>     │ │
```
brian-brazil, line 68: Should we allow for additional metadata in future?

tsdb/docs/format/wal.md
```diff
@@ -50,6 +50,27 @@ Series records encode the labels that identifies a series and its unique ID.
...
+Metadata records encode the metadata associated with a series.
+
+```
+┌────────────────────────────────────────────┐
+│ type = 1 <1b>                              │
+├────────────────────────────────────────────┤
+│ ┌────────────────────────────────────────┐ │
+│ │ id <uvarint>                           │ │
+│ ├─────────────────────┬──────────────────┤ │
+│ │ len(type) <uvarint> │ type <bytes>     │ │
```
brian-brazil, line 64: type should be an int

tsdb/wal/checkpoint.go
```diff
@@ -181,6 +184,24 @@ func Checkpoint(logger log.Logger, w *WAL, from, to int, keep func(id uint64) bo
 			stats.TotalSeries += len(series)
 			stats.DroppedSeries += len(series) - len(repl)
 
+		case record.Metadata:
+			metadata, err = dec.Metadata(rec, metadata)
+			if err != nil {
+				return nil, errors.Wrap(err, "decode series")
```
brian-brazil, line 190: series?

robskillington: Ah good call, yes will change this to "decode metadata".

tsdb/docs/format/wal.md
robskillington: That sounds good, I can add a versioning scheme to record types if you like. It may add some complexity and doesn't exist already - but if that's the direction that would be ideal to move towards I can attempt implementing a relatively simple method to try to achieve this. 

Otherwise we'd have to change the record type altogether (i.e. MetadataV2) which means on a rollback of version, all new MetadataV2 records would be skipped (which would not be ideal, ideally you want to partially read the fields you know about, just skip the new fields you don't know about).

tsdb/docs/format/wal.md
robskillington: Sounds good, I'll update `storage.Metadata` to use a uint8 as well.

scrape/scrape.go
```diff
@@ -714,19 +715,29 @@ type scrapeCache struct {
...
+}
+
 // metaEntry holds meta information about a metric.
 type metaEntry struct {
-	lastIter uint64 // Last scrape iteration the entry was observed at.
-	typ      textparse.MetricType
-	help     string
-	unit     string
+	lastIter             uint64 // Last scrape iteration the entry was observed at.
+	lastIterObservedType uint64 // Last scrape iteration the type was observed at.
```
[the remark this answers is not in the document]
robskillington: Currently metadata is not actually zeroed when there is new metadata per scrape, each field (type/unit/help) are all individually are updated if and only if they are seen, all of which in their setters bump the `lastIter` field to current iteration. Since we want the effect of setting any of these to "zero" out the metadata when we next inform the TSDB of a metadata update, we need to track which we actually updated this scrape and treat the other fields as zeroed.

The complexity with actually zeroing out the other fields when any of type/unit/help is set the first time during a scrape is that the lock is only held for the duration of setting the single metadata field (type or unit or help) so zeroing the other fields at that time will mean that the metadata goes missing temporarily for the field you didn't update (i.e. you set type and clear unit/help since it's the first update for this "iter", once you release the lock now the old values of unit/help aren't there anymore even though the very next line of the scrape you parse could be restoring the unit/help values to what they were just before you cleared them if they aren't being zeroed this scrape).

So for consistency what I was doing here is keeping the current behavior of "don't zero the other metadata so that metadata constantly moves only forwards and is never reset" for the current metadata consumers, but adding enough information to the variables to know actually which fields were present when we did observe a change and using that information to properly make those fields zeroed when passing along metadata to the TSDB and only including the metadata fields that were set with the scrape iteration that the metadata changed.

tsdb/docs/format/wal.md
brian-brazil: Don't add strong versioning, that just means keeping multiple parsers around. I'd add the number of entries that follow.

scrape/scrape.go
robskillington: The use case for this is we don't have a "Ref" kept around. Also even if we have a "Ref" and can use "AddFast" what happens if "ErrNotFound" is returned and we have to call "Add(...)" from this code which already is calling just "Add(...)" directly?
https://github.com/prometheus/prometheus/blob/2b75c1b19995377f2ffcea7aff6818982f2f2bc6/scrape/scrape.go#L1345-L1355

`seriesCur` and `seriesPrev` back what is passed into that inner method:
https://github.com/prometheus/prometheus/blob/2b75c1b19995377f2ffcea7aff6818982f2f2bc6/scrape/scrape.go#L832-L840

storage/remote/queue_manager.go
```diff
@@ -468,6 +501,19 @@ func releaseLabels(ls labels.Labels) {
 	}
 }
 
+func internMetadata(meta storage.Metadata) storage.Metadata {
+	meta.Type = textparse.MetricType(interner.intern(string(meta.Type)))
```
cstyan, line 505: this doesn't seem correct? should it not be just `interner.intern(string(meta.Type))`?

tsdb/head.go
```diff
@@ -1118,12 +1120,38 @@ func (a *headAppender) Add(lset labels.Labels, t int64, v float64) (uint64, erro
 		return 0, err
 	}
 
+	meta := false
+	s.Lock()
+	if s.typ != m.Type {
```
cstyan, line 1125: this block will result in whatever the last seen metadata for a series (based on labelset hash) was being used as the metadata for an entire block, is that right?

brian-brazil:
> I don't see the inclusion of metadata in write requests. Just want to confirm that's still todo and I'm not missing something?

It's hidden in queue_manager.go, the last diff.

robskillington:
Here's the results from testing:
- node_exporter exporting 309 metrics each by turning on a lot of optional collectors, all have help set, very few have unit set
- running 8 on the host at 1s scrape interval, each with unique instance label
- steady state 134kb/sec without this change
- steady state 172kb/sec with this change
- roughly 28% increase

Screen Shot 2020-08-19 at 4 11 38 AM

robskillington:
Here's the increase with a default config, the difference is relatively less measurable - likely to do with the previous config sending larger batches since the send deadline was longer with a max samples per send of 5,000:
- steady state ~177kb/sec without this change
- steady state ~210kb/sec with this change
- roughly 20% increase

Screen Shot 2020-08-19 at 4 38 08 AM

Default config:
```
remote_write:
  - url: http://localhost:3030/remote/write
    remote_timeout: 30s
```

tsdb/docs/format/wal.md
robskillington: After catching up, went with writing the size of the encoded fields upfront so we can skip any fields unknown to the decoder once we've decoded all known fields to the decoder.

I've added a unit test to cover this case.

cmd/promtool/tsdb.go
```diff
@@ -206,7 +206,7 @@ func (b *writeBenchmark) ingestScrapesShard(lbls []labels.Labels, scrapeCount in
 			s.value += 1000
 
 			if s.ref == nil {
-				ref, err := app.Add(s.labels, ts, float64(s.value))
+				ref, err := app.Add(s.labels, storage.Metadata{}, ts, float64(s.value))
```
Harkishen-Singh, line 209: Why not just make it as a pointer and pass as `nil`?

cstyan:
I'm still on the fence about 20-30% network usage increase by default, for large users in public cloud that's going to be a big increase in a network egress bill that's probably already a decent chunk of money. If we can move towards periodic inclusion of all metadata (like Josh's PR) with an option to always include all metadata. we can get closer to merging something.

robskillington:
@cstyan there are some attributes that are required by downstream systems always, for instance TYPE and Stackdriver's API:
https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors
> Defines a metric type and its schema. Once a metric descriptor is created, deleting or altering it stops data collection and makes the metric type's existing data unusable.
> The following are specific rules for service defined Monitoring metric descriptors:
type, metricKind, valueType and description fields are all required. The unit field must be specified if the valueType is any of DOUBLE, INT64, DISTRIBUTION.
> Maximum of default 500 metric descriptors per service is allowed.
> Maximum of default 10 labels per metric descriptor is allowed.

Also some systems work differently and cannot consume the metadata periodically like Josh's PR. I would like the project to make sure it considers all consumers of the API when making decisions of this nature, rather than catering to one or two specific implementations.

Accompanying the metadata with the samples themselves definitely is more straight forward for a wider set of consumers, as such I don't see a reason to block such an implementation. Providing alternatives seem sane, but blocking it seems heavy handed and in favor of specific consumers.

tsdb/docs/format/wal.md
```diff
@@ -50,6 +50,29 @@ Series records encode the labels that identifies a series and its unique ID.
...
+Metadata records encode the metadata associated with a series.
+
+```
+┌────────────────────────────────────────────┐
+│ type = 1 <1b>                              │
+├────────────────────────────────────────────┤
+│ ┌────────────────────────────────────────┐ │
+│ │ id <uvarint>                           │ │
+│ ├────────────────────────────────────────┤ │
+│ │ size <uvarint>                         │ │
```
brian-brazil, line 64: We usually do the length as a fixed 4b first in our formats.

tsdb/record/record.go
```diff
@@ -184,6 +237,51 @@ func (e *Encoder) Series(series []RefSeries, b []byte) []byte {
...
+func (e *Encoder) Metadata(series []RefMetadata, b []byte) []byte {
+	buf := encoding.Encbuf{B: b}
+	buf.PutByte(byte(Metadata))
+
+	for _, s := range series {
+		buf.PutUvarint64(s.Ref)
+
+		sizeAfterSeriesRef := buf.Len()
+
+		// Reserve space to write the size of the metadata fields
```
brian-brazil, line 254: This seems a bit over-complicated, and not how we approach this elsewhere. Write the metadata to a temporary buffer, gets its size and then work from there.

cstyan:
@robskillington I think we're in agreement here that the metadata implementation shouldn't be blocking your use case, but for others I think the 20-30% network increase is a bit much for "always include all metadata" to be the default. That's why I suggested `periodic inclusion of all metadata with an option to always include all metadata`, where the latter would solve your use case.

brian-brazil:
I don't think we should have a default that only works for one particular remote storage, and more generally I'd aim for something that doesn't require any configuration at all.

tsdb/record/record.go
robskillington: Are you ok if I make the signature then take two buffers, each that are independent and reused? 

i.e before: 
```go
func (e *Encoder) Metadata(series []RefMetadata, b []byte) []byte {
```

after:
```go
func (e *Encoder) Metadata(series []RefMetadata, b1, b2 []byte) (b1, b2 []byte) {
```

It will be kind of strange however, since callers will only want to use the first result "b1" ("b2" will only be for intermediary buffering on subsequent calls to "Metadata(...)").

The other path I can go down is putting the buffers on the `Encoder` themselves. e.g.
```go
type Encoder struct {
  buffer []byte
  metadataBuffer []byte
}

func NewEncoder(buffer, metadataBuffer []byte) Encoder {
  return Encoder{buffer: buffer, metadataBuffer: metadataBuffer}
}

func (e *Encoder) Bytes() []byte {
  return e.buffer
}

func (e *Encoder) Series(series []RefSeries) {
  // existing implementation but write to e.buffer after clearing, then let caller get the bytes with e.Bytes()
}

func (e *Encoder) Metadata(metadata []RefMetadata) {
  // existing implementation but write to e.buffer after clearing, also use e.metadataBuffer as separate independent buffer which is necessary since we need to write the size of entire record to e.buffer before the contents which is only known after encoding to e.metadataBuffer, then let caller get the bytes with e.Bytes()
}
```

cstyan:
@brian-brazil regardless of what the default is, without a way for users to rein in network usage with the inclusion of metadata I can't accept this PR. A 20-30% increase for users who are charged for network egress without an option to reduce that in any way is not acceptable. 

That being said, I have no intention of blocking something that solves @robskillington's use case.

csmarchbanks:
I finally had some time to try to deploy this to a couple of dev clusters. I saw a pretty large variance in remote write bandwidth, one cluster only had a 5% increase, while another was more like Rob's findings of 20%.

I am also a bit hesitant to force all users into accepting an extra 20-30% increase in egress cost for little or no benefit in many systems. Configuration would be nice, at least for the HELP text part IMO.

Also, could there be a toggle around writing the new record type to the WAL, defaulted to off? That would allow users to downgrade if they haven't turned it on yet. I think the last major WAL change was compression which had a toggle around it.

brian-brazil:
> I am also a bit hesitant to force all users into accepting an extra 20-30% increase in egress cost for little or no benefit in many systems. 

Remote write was added as the simplest thing that works, not the most efficient possible solution. I'm wary of complicating things just because some remote storages don't support every possible feature - especially as this is something that the general intention is to make more common across the ecosystem.

> Also, could there be a toggle around writing the new record type to the WAL, defaulted to off? That would allow users to downgrade if they haven't turned it on yet. I think the last major WAL change was compression which had a toggle around it.

What's in the WAL is really unrelated. The WAL file size change is going to be in the noise.

cstyan:
@brian-brazil you're talking about things as if Cortex and M3 are the only remote storage options.

> I don't think we should have a default that only works for one particular remote storage, and more generally I'd aim for something that doesn't require any configuration at all.

To my knowledge Rob and M3 are the only remote storage implementation that have come forward and said Josh's PR would explicitly not work for them, so saying that periodic inclusion of metadata would only work for one particular remote storage is a bit of a reach.

> Remote write was added as the simplest thing that works, not the most efficient possible solution. I'm wary of complicating things just because some remote storages don't support every possible feature - especially as this is something that the general intention is to make more common across the ecosystem.

Complicating things is part of the natural evolution of software whose original design didn't solve valid future use cases. There's also nothing in this discussion that's been said that could mean 'some remote storages don't support every possible feature'. This isn't a question of whether Cortex or M3 _**can**_ ingest all metadata for samples within a remote write request, but whether users of remote storage systems will accept a 20-30% increase in network traffic that they have no way of opting out of.

Chris and I have both stated our reluctance to not allow users to reduce remote write metadata traffic. I've also stated that wouldn't accept a remote write metadata PR that doesn't allow users to do so. If we can't move forward with something that solves both problems I suggest we discuss further with the rest of -team.

brian-brazil:
> so saying that periodic inclusion of metadata would only work for one particular remote storage is a bit of a reach.

True, on the other hand neither has any other remote storage come forward and said that it'd work. We've have however had a previous use case where Josh's PR would not have worked.

The more general point is that remote write should be generic, and be possible to be used correctly by everyone out of the box - even if it's not 100% efficient.
Rob's arguments make clear that Josh's proposal is not generic and it's also not in line with where we seem to be going with metadata. We've at this stage two use cases which require semantics where the metadata for a given sample has to be available at the time we write it, not to receive a partial unlinked set of metadata a few minutes later.

I don't think it makes sense to have a limited approach implemented in parallel with a full approach, as that's just going to end up as vestigial complication and technical debt. Particularly when in due course Cortex will presumably also require the full approach. Thus in a situation where Rob's PR is on the table, I don't believe a periodic sending of amorphous metadata similar to Josh's PR should be a configurable option, let alone the default.

> whether users of remote storage systems will accept a 20-30% increase in network traffic that they have no way of opting out of.

Pulling up some old numbers, I figure that network costs for the remote write proportion (presuming the worst case that you're paying for both egress and ingress) are about at most a quarter of the overall ingestion&storage costs. Thus the financial impact to overall remote storage costs is going to be nearer 5%, likely a fair bit lower when querying and other overheads (e.g. spare capacity for growth) are considered. With that context, do you consider this acceptable?

Failing that, I think we should look at ways to make this PR a bit more judicious in when it sends metadata.

cstyan:
We already have ways to limit ingestion and storage costs with relabelling and retention periods, and reduce remote write network usage again with relabelling. There's also other remote write users such as https://github.com/grafana/agent, where the cost % from remote write network usage is much higher. We need to include some way to allow users to limit the impact of the inclusion of metadata on their network bill. 

I'm not hung up on how we implement that, so if you're issue is with having both Rob and Josh's approach available to users then we can just drop that part of the discussion and move on to something more productive.

brian-brazil:
> We already have ways to limit ingestion and storage costs with relabelling and retention periods, and reduce remote write network usage again with relabelling. 

All those things will continue to work.

> There's also other remote write users such as https://github.com/grafana/agent, where the cost % from remote write network usage is much higher.

Unless I'm missing something that doesn't make a difference, as it'd be the same remote write request on the wire. You ultimately have to pay for not just the network for the request itself, but also the CPU/RAM/disk for the entire system which is processing and storing it on the other end too.

> We need to include some way to allow users to limit the impact of the inclusion of metadata on their network bill.

My point is that you can't look at one specific part of the bill in isolation, but rather at the total cost of the whole remote storage system. In that light does doing thing the naive way look okay?
The most efficient system I could previously come up with had about a quarter of the costs being ingestion network, so for a real world system a tenth is probably not too bad an estimate. Is it worth complicating things for the sake of an overall 1-2% cost increase, in systems which probably organically grow by more than that every month?

> I'm not hung up on how we implement that, so if you're issue is with having both Rob and Josh's approach available to users then we can just drop that part of the discussion and move on to something more productive.

Yes, that'd be my main issue.

csmarchbanks:
I am going to push on this once more: I don't think it is a good pattern to intentionally release a version of Prometheus that cannot be rolled back without manual intervention and data loss. I understand that it is not required, but I think all of Beorn's comments starting with https://github.com/prometheus-junkyard/tsdb/pull/609#issuecomment-494775294 ring true here as well. Is it significantly harder than I think to provide a flag that controls writing metadata to the WAL?

Such a flag would also alleviate the concerns around increasing bandwidth for everyone until we can add some sort of user control like what Callum mentions. For example, I would love the ability to use relabeling (or something else) to modify or remove the help text like I can for abusive labels.

brian-brazil:
> Is it significantly harder than I think to provide a flag that controls writing metadata to the WAL?

It's not, however it doesn't feel right either to be controlling remote write behaviour by disabling/enabling features at the TSDB level.

We could put out a release that supports both, so you can go back one minor.

> For example, I would love the ability to use relabeling (or something else) to modify or remove the help text like I can for abusive labels.

Relabelling applies to labelsets, not metadata. If there's abusive help text you'd have to drop all the samples that have that metadata. In that sense we already have that control (which incidentally Josh's PR would not have - but bandwidth there is much smaller).

cstyan:
It's really unproductive to have this lengthy of a discussion about whether to include something (Josh's metadata RW implementation) we as -team had already agreed to merge as part of the last dev summit. I'm all for another solution that accommodates both Rob's use case and other users wishes to be able to limit metadata sent via remote write, but at the moment we have Josh's implementation that could likely be re-implemented on top of Robs changes with the addition of a config option that has the metadata sent either periodically or in addition to the samples that are being sent.

Chris and I have already stated our desire to see an implementation that addresses our concerns regarding the 20-30% network usage increase. If you still disagree that a config option of some kind should be included to do something useful here we don't have a path forward, and I'll add this PR to the agenda for the next -team dev summit.

brian-brazil:
> we as -team had already agreed to merge as part of the last dev summit

This is true, but to add context that was as a stop gap. Given that a full solution has come to us (and much sooner than I personally was expecting), I think we should be looking at getting that full solution in rather than adding something that could be removed again in a release or two.

> If you still disagree that a config option of some kind should be included to do something useful here we don't have a path forward,

I generally prefer to avoid configuration where we can do without to keep things simpler for users, and as I've indicated above there are approaches that don't require any configuration. For example what if the metadata was not sent if the current sample's timestamp is less than 5m since the last sample where we sent on the metadata.

cstyan:
> For example what if the metadata was not sent if the current sample's timestamp is less than 5m since the last sample where we sent on the metadata.

@robskillington can you confirm whether or not this implementation would solve your use case?

robskillington:
@cstyan that would solve the StackDriver use case and Thanos & M3 aggregation use cases too yes (if we retained TYPE with each sample, so basically opting to only send HELP and UNIT every 5m between sample timestamps).

cstyan:
I would still like to have the option to configure the send interval for each metadata field, with some value being 'never send this metadata'. A default that is "send type and unit every time but help only every 5 minutes" is fine with me.

With this we're getting what seems to be a sane default, the option for others to tweak metadata inclusion to their needs, and not implementing/maintaining two ways of gathering/including metadata in remote write.

brian-brazil:
I think we should avoid configuration where we don't strictly need it, as it often ends up with user confusion when setting don't have the implications they realise plus it's one more bit of cognitive overhead. For example I don't think that never should be an option, as that could cause undesirable issues on the remote storage when a user decides to "optimise" things, plus it goes against the general direction of encouraging more metadata use.

I think we need  something that is always correct and has known bounds, and configuration goes against that goal while also ultimately making things worse for users.

brian-brazil:
> What's next steps, is there some consensus to come to or should I go ahead and make those changes? 

I gather you're okay here. When would you expect it to be ready?

> I hope including TYPE on each sample is not contentious (since it should be just a single byte or so depending on protobuf encoding boxing).

That sounds okay to me, what's the bandwidth implication?

brian-brazil:
Do you have an idea as to when this PR will be ready for review?

Given that this is taking a a while, and that #6815 was considered an acceptable stopgap while waiting for a more complete solution, I propose that we aim to get something in for the 2.23 release and if this isn't in a largely mergable state in 2 weeks that we go back to reviewing #6815.

cmacnaug:
@robskillington, are you still planning on moving forward with this PR? Even with #6815 merged, having `type` information emitted in-band with the time series would be very useful for systems that operate on the remote write data as a stream. 

At New Relic, we have a use case in which we convert counters to delta values during ingest and #6815 doesn't really give us what we are looking for because the time series metadata can arrive after the time series data itself. 

With #6815 available, I wonder if paring this down to just including type information would be sufficient for others' remaining use cases? It seems like doing so would address the concerns around network utilization and the need for extra configuration.

robskillington:
@cmacnaug yes was hoping to get to this over the holidays, let's see how that goes. Also we might have someone else take it over if that fails. Ideally this lands in the next 4-8 weeks. @brian-brazil @csmarchbanks + others, does this sound sane?

cstyan:
Yes, gathering metadata from the WAL and including metadata alongside sample data (with some kind of filtering options) is still something we'd like to include.

tpaschalis:
@robskillington I've read through the discussions, design doc and present work on this, nice work bringing it all together! 

I'm thinking of taking a stab and using your work here as a base to making this work on a new PR. Would you be alright with this?

robskillington:
@tpaschalis I had not seen your comment, that would be a really wonderful thing - let me know if there's anything I can do to support

roidelapluie:

It seems like metadata is appended to the WAL now, but not yet to the remote write. Given the age of the PR and the amount of work to rewrite it and get to a consensus about remote write #10312  we have decided to close this PR.

PR #7772, opened 2020-08-10
tsdb: Delete blocks atomically; Remove tmp blocks on start; Added test.

bwplotka:
Reason: When panic / `SEGFault` happens during deletion TSDB, can't start gracefully ): 

## Changes:

* Rename dir when deleting
* Ignoring blocks with broken meta.json on start (we do that on reload)
* Compactor writes <ulid>.tmp-for-creation blocks instead of just .tmp
* Delete tmp-for-creation and tmp-for-deletion blocks during DB open.

cc @brian-brazil @brancz @metalmatze @codesome 

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -49,6 +49,14 @@ import (
 const (
 	// Default duration of a block in milliseconds.
 	DefaultBlockDuration = int64(2 * time.Hour / time.Millisecond)
+
+	// Block dir suffixes to make deletion and creation operations atomic.
+	// We decided to do suffixes instead of creating meta.json as last (or delete as first) one,
+	// because in error case you still can recover meta.json from the block content within local TSDB dir.
+	// TODO(bwplotka): TSDB can end up with various .tmp files (e.g meta.json.tmp, WAL or segment tmp file. Think
+	// about removing those too on start to save space. Currently only blocks tmp dirs are removed.
+	tmpForDeletionBlockDirSuffix = ".tmp-for-deletion"
+	tmpForCreationBlockDirSuffix = ".tmp-for-creation"
 )
 
 var (
@@ -566,11 +574,15 @@ func open(dir string, l log.Logger, r prometheus.Registerer, opts *Options, rngs
 
 	// Fixup bad format written by Prometheus 2.1.
 	if err := repairBadIndexVersion(l, dir); err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "repair bad index version")
 	}
 	// Migrate old WAL if one exists.
 	if err := MigrateWAL(l, filepath.Join(dir, "wal")); err != nil {
 		return nil, errors.Wrap(err, "migrate WAL")
+	}
+	// Remove garbage, tmp blocks.
+	if err := removeBestEffortTmpDirs(l, dir); err != nil {
+		return nil, errors.Wrap(err, "remove tmp dirs")
 	}
 
 	db = &DB{
@@ -658,6 +670,23 @@ func open(dir string, l log.Logger, r prometheus.Registerer, opts *Options, rngs
 	go db.run()
 
 	return db, nil
+}
+
+func removeBestEffortTmpDirs(l log.Logger, dir string) error {
+	files, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return err
+	}
+	for _, fi := range files {
+		if isTmpBlockDir(fi) {
+			if err := os.RemoveAll(filepath.Join(dir, fi.Name())); err != nil {
+				level.Error(l).Log("msg", "failed to delete tmp block dir", "dir", filepath.Join(dir, fi.Name()), "err", err)
+				continue
+			}
+			level.Info(l).Log("msg", "Found and deleted tmp block dir", "dir", filepath.Join(dir, fi.Name()))
+		}
+	}
+	return nil
 }
 
 // StartTime implements the Storage interface.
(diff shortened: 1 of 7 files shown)


tsdb/db.go
```diff
@@ -1016,7 +1017,7 @@ func DefaultBlocksToDelete(db *DB) BlocksToDeleteFunc {
 	}
 }
 
-// deletableBlocks returns all blocks past retention policy.
+// deletableBlocks returns all currently loaded blocks past retention policy or already compacted into new block.
```
kakkoyun, line 1020:
```suggestion
// deletableBlocks returns all currently loaded blocks past retention policy or already compacted into a new block.
```

tsdb/db.go
```diff
@@ -1105,7 +1106,13 @@ func (db *DB) deleteBlocks(blocks map[ulid.ULID]*Block) error {
 				level.Warn(db.logger).Log("msg", "Closing block failed", "err", err, "block", ulid)
 			}
 		}
-		if err := os.RemoveAll(filepath.Join(db.dir, ulid.String())); err != nil {
+
+		// Rename atomically to avoid partial block when process crashing during this moment.
+		tmpToDelete := filepath.Join(db.dir, fmt.Sprintf("%s.tmp-to-delete", ulid))
+		if err := os.Rename(filepath.Join(db.dir, ulid.String()), tmpToDelete); err != nil {
+			return errors.Wrapf(err, "atomic rename for delete obsolete block %s", ulid)
```
squat, line 1113:
```suggestion
			return errors.Wrapf(err, "atomic rename for deleting obsolete block %s", ulid)
```

tsdb/db.go
```diff
@@ -1105,7 +1106,13 @@ func (db *DB) deleteBlocks(blocks map[ulid.ULID]*Block) error {
 				level.Warn(db.logger).Log("msg", "Closing block failed", "err", err, "block", ulid)
 			}
 		}
-		if err := os.RemoveAll(filepath.Join(db.dir, ulid.String())); err != nil {
+
+		// Rename atomically to avoid partial block when process crashing during this moment.
```
squat, line 1110:
```suggestion
		// Rename atomically to avoid partial block when process is crashing during this moment.
```

bwplotka:
Should we aim for deleting those on start in this PR? @squat @kakkoyun ?

kakkoyun:
@bwplotka I'd say yes. Since it's low-hanging fruit.

roidelapluie:
We should do this in 2 passes:

1. rename all blocks to tmp-to-delete

2. delete the tmp-to-delete

WDYT?

roidelapluie:
The alternative to remove the meta file first seems great too.

tsdb/db.go
krasi-georgiev, line 1110: use `fileutil.Replace` ?

tsdb/db.go
```diff
@@ -549,6 +549,7 @@ func validateOpts(opts *Options, rngs []int64) (*Options, []int64) {
 	return opts, rngs
 }
 
+// TODO(bwplotka): Detect .tmp and .tmp-to-delete dirs and remove on start.
```
krasi-georgiev, line 552: Why a todo? The PR is small enough to add here. Then it will be within the same context why these tmp might folders exist anyway.

krasi-georgiev:
yes to deleting  the tmp folders on start up.

bwplotka:
> We should do this in 2 passes:
> 
> rename all blocks to tmp-to-delete
> 
> delete the tmp-to-delete
> 
> WDYT?

Why it matters @roidelapluie 

I will explore removing meta.json then and assuming such block ready to be deleted

roidelapluie:
> > We should do this in 2 passes:
> > rename all blocks to tmp-to-delete
> > delete the tmp-to-delete
> > WDYT?
> 
> Why it matters @roidelapluie
> 
> I will explore removing meta.json then and assuming such block ready to be deleted

You protect with a crash in the middle of the deletes, but a crash at that time would not guarantee that all the deleted blocks would be deleted, or unseen.

bwplotka:
Ok, addressed all comments and changed a bit. All changes are in updated [description](https://github.com/prometheus/prometheus/pull/7772#issue-465470101)

Can you re-PTAL? @pracucci @kakkoyun @roidelapluie @krasi-georgiev ? (:

roidelapluie:
Oh I know why I wanted this to first rename all then delete:

If it crashed while deleting, if that is during compaction, Prometheus still would not restart unless you have the vertical merge enabled. Is this function called after compaction too?

bwplotka:
> If it crashed while deleting, if that is during compaction, Prometheus still would not restart unless you have the vertical merge enabled. Is this function called after compaction too?

Ah good point. Need to look up all deletions really. Including done by Delete API if full block is to remove etc.

bwplotka:
Ok, the delete on reload function is used for compaction as well - so kind of same delete path is used. 

However the crash during that path is fine - as long as deletion is atomic. We handed those overlaps just fine thanks to reload functionality of deduping those if parent matches:

> // By creating blocks with their parents, we can pick up the deletion where it left off during a crash

https://github.com/prometheus/prometheus/blob/9c12b0146897bd8a4487e7155e0e4783e6fe036d/tsdb/db.go#L927

So we are safe on this case (:

tsdb/db.go
krasi-georgiev: can't use `fileutil.Replace` ?

PR #7773, opened 2020-08-10
Ensure that step is in milliseconds in console graphs.

### LINKED ISSUE #7711: PromConsole area renderer elevates a time series unexpectedly
**What did you do?**

I used PromConsole.Graph with `renderer: "area"`, and often see time series elevated more than they should be, as if they were stacked on other time series which are not actually under them.  I attach a screenshot.

See https://github.com/livebungalow/promconsole-rendering-example for a small working example (well, non-working, which is the point).  For the purposes of that example, I monkey-patch `PromConsole.Graph.prototype.buildQueryUrl` to inject data which provokes the issue -- I copied this data from an actual `query_range` response, formatted it, and reduced it to a manageable size.

Time series resting on the x-axis.

Time series elevated off the x-axis.  The linked example repo has sample data that provokes the problem.

**Environment**

The linked example uses the docker image `prom/prometheus:v2.19.3` with its standard configuration file.

brian-brazil:
Further precision is truncated by the Prometheus API, so the
steps don't end up quite aligning subsequently.

Fixes #7711

### DIFF
--- a/web/ui/static/js/prom_console.js
+++ b/web/ui/static/js/prom_console.js
@@ -3,7 +3,6 @@
  * as graphs.
  *
  */
-
 PromConsole = {};
 
 PromConsole.NumberFormatter = {};
@@ -474,7 +473,7 @@ PromConsole.Graph.prototype._render = function(data) {
 			var newSeries = [];
 			var pos = 0;
 			var start = self.params.endTime - self.params.duration;
-      var step = self.params.duration / this.graphTd.offsetWidth;
+      var step = Math.floor(self.params.duration / this.graphTd.offsetWidth * 1000) / 1000;
 			for (var t = start; t <= self.params.endTime; t += step) {
 				// Allow for floating point inaccuracy.
 				if (series[seriesLen].data.length > pos && series[seriesLen].data[pos].x < t + step / 100) {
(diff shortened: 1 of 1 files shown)


web/ui/static/js/prom_console.js
```diff
@@ -562,7 +561,7 @@ PromConsole.Graph.prototype.buildQueryUrl = function(expr) {
   var p = this.params;
   return PATH_PREFIX + "/api/v1/query_range?query=" +
     encodeURIComponent(expr) +
-    "&step=" + p.duration / this.graphTd.offsetWidth +
+    "&step=" + Math.floor(self.params.duration / this.graphTd.offsetWidth * 1000) / 1000; +
```
roidelapluie, line 564:
```suggestion
    "&step=" + Math.floor(p.duration / this.graphTd.offsetWidth * 1000) / 1000; +
```

since we already have this

brian-brazil: Done

web/ui/static/js/prom_console.js
```diff
@@ -562,7 +561,7 @@ PromConsole.Graph.prototype.buildQueryUrl = function(expr) {
   var p = this.params;
   return PATH_PREFIX + "/api/v1/query_range?query=" +
     encodeURIComponent(expr) +
-    "&step=" + p.duration / this.graphTd.offsetWidth +
+    "&step=" + Math.floor(p.duration / this.graphTd.offsetWidth * 1000) / 1000; +
```
roidelapluie, line 564:
```suggestion
    "&step=" + (Math.floor(p.duration / this.graphTd.offsetWidth * 1000) / 1000) +
```

I guess the `;` wont work here

brian-brazil: huh, how did I not catch that in testing? Ah, the test case from the user bypasses this.

PR #7775, opened 2020-08-11
rules: add evaluationTimestamp when copy state.

johncming:
## Bug

`LastEvaluation` in web API may be missing in the interval between `CopyState` and `newg.run(m.opts.Context)`

## Detail

two properties used in web API
https://github.com/prometheus/prometheus/blob/e6d7cc5fa43716281bdf5d1cf40415f99a0e9623/web/api/v1/api.go#L1066-L1067

https://github.com/prometheus/prometheus/blob/c7a17f649168ef47caa683cf9c6e243f27379a92/rules/manager.go#L501-L503

the property `evaluationTimestamp` was not copied before a new group running.
https://github.com/prometheus/prometheus/blob/c7a17f649168ef47caa683cf9c6e243f27379a92/rules/manager.go#L947-L954

### DIFF
--- a/rules/manager.go
+++ b/rules/manager.go
@@ -500,6 +500,7 @@ func nameAndLabels(rule Rule) string {
 // first is matched with the first, second with the second etc.
 func (g *Group) CopyState(from *Group) {
 	g.evaluationDuration = from.evaluationDuration
+	g.evaluationTimestamp = from.evaluationTimestamp
 
 	ruleMap := make(map[string][]int, len(from.rules))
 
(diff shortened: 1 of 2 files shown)


brian-brazil:
That makes sense, should there be a unittest?

johncming:
> That makes sense, should there be a unittest?

OK.

brian-brazil:
Hmm, this is failing on Windows for some odd reason. Can you try rebasing?

johncming:
> Hmm, this is failing on Windows for some odd reason. Can you try rebasing?

OK.

brian-brazil:
Thanks!

PR #7779, opened 2020-08-11
WIP: Adds support for prettifying PromQL expressions in rules files and via CLI

### LINKED ISSUE #21: Rule formatting tool.

Harkishen-Singh:
This pull-request adds support for prettifying PromQL expressions in rules file via Promtool, along with preserving user comments in the PromQL expressions.

`./promtool prettify rules <path_of_rules_file>`

The code is implemented in line with the [design doc](https://docs.google.com/document/d/1nOBjpuCk4CsrOSm2ZjfVz2EL6gmA_CFGSbHCdY0Royg/edit) and the [implementation doc](https://docs.google.com/document/d/1ym4CXceIWyCx5sric9Y-WjiLnJXe3W-tidEEOWpoXhk/edit) that was shared in the Prometheus-dev mailing list.

Fixes #21

- [ ] Prettifying PromQL expressions in rules files
- [ ] Prettifying a single PromQL expression via promtool CLI


promql/prettier/prettier.go
```diff
@@ -0,0 +1,577 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package prettier
+
+import (
+	"fmt"
+	"gopkg.in/yaml.v3"
```
brian-brazil, line 18: Nothing under promql should know about yaml or rules files.

promql/prettier/prettier.go
```diff
@@ -0,0 +1,577 @@
...
+	// in VectorSelector are splittable.
+	expectAggregationLabels bool
+	// labelsSplittable determines whether the current formatting region lies in labels
+	// matchers in VectorSelector. Since labels in VectorSelectors are only splittable,
+	// is property is applied only after encountering a Left_Brace. This is dependent across
+	// multiple padder states and hence needs to be carried along.
+	labelsSplittable bool
+}
+
+// New returns a new prettier over the given slice of files.
```
brian-brazil, line 71: Promql should not know about files, only expressions.

Harkishen-Singh: This is intended to be served via Promtool. Hence, the promtool is the calling function mostly, however we can have an API endpoint as well. This is, in fact, the reason why it is kept as a separate module.

brian-brazil: Then promtool is what should be handling any file handling, not PromQL.

codesome:
> This should be primarily working through printer.go, not adding a new codepath to render PromQL.

I would prefer this be an independent thing right now because the flow of this and current printer is a little different while rendering. Once this is upstream and has been used for some time and fixed any bugs that we might find out later, we can remove the printer.go and eventually have a single way to render. WDYT?

brian-brazil:
I don't think that makes sense. If it is felt the code isn't ready to be inside Prometheus then we should wait until it is in such a state. As long as the code doesn't change the meaning of any expressions, it should be fine to use.

codesome:
Then I propose removing printer.go

codesome:
> As long as the code doesn't change the meaning of any expressions

Yes, that should be the case, we don't skip any items and print them in order.

brian-brazil:
> Then I propose removing printer.go

That's basically what I'm saying, anything like this should replace the existing printer.

promql/prettier/prettier.go
Harkishen-Singh: Made prettier independent of files. Now, it's handled by promtool.

cmd/promtool/main.go
```diff
@@ -421,6 +431,40 @@ func CheckMetrics() int {
...
+				}
+				rgs.Groups[i].Rules[j].Expr.SetString(prettifiedExpr)
+			}
+		}
+		data, err := yaml.Marshal(rgs)
+		if err != nil {
+			panic(err)
+		}
+		if err = ioutil.WriteFile(filePath, data, 0666); err != nil {
+			fmt.Fprintln(os.Stderr, fmt.Sprintf("failed formatted %s file:", filePath), err.Error())
```
brian-brazil, line 458: Shouldn't this return a failure?

promql/prettier/node.go
```diff
@@ -0,0 +1,169 @@
+// Copyright 2015 The Prometheus Authors
```
brian-brazil, line 1: 2015?

promql/prettier/node.go
```diff
@@ -0,0 +1,169 @@
...
+	for i := 0; i < len(history)-1; i++ {
+		if !(is(history[i], binaryExpr) && is(history[i+1], binaryExpr)) {
+			temp = append(temp, history[i])
+		}
+	}
+	temp = append(temp, history[len(history)-1])
+	return temp
+}
+
+func is(node parser.Node, typ uint) bool {
```
brian-brazil, line 150: `is` is a bit short of a function name, it looks like it's a Go thingy.

promql/prettier/prettier.go
```diff
@@ -0,0 +1,499 @@
...
```
brian-brazil, line 59: This needs updating.

promql/prettier/prettier.go
```diff
@@ -0,0 +1,499 @@
...
+func New(expression string) *Prettier {
+	return &Prettier{
+		expression: expression,
+		pd:         &padder{indent: "  ", isNewLineApplied: true, columnLimit: 100},
```
brian-brazil, line 63: If we're hardcoding then these may be better as file-level consts. That'l help avoid deviation by different users over time.

Presume that everything here is going to end up public, as that is the policy.

promql/prettier/prettier.go
```diff
@@ -0,0 +1,499 @@
...
+			}
+			itr = advanceComments(items, itr)
+			if items[itr+3].Typ == parser.COMMA {
+				skipBraces = rightBraceIndex-5 == advanceComments(items, leftBraceIndex)
+			} else {
+				skipBraces = rightBraceIndex-4 == advanceComments(items, leftBraceIndex)
+			}
+			identifierItem := parser.Item{Typ: parser.IDENTIFIER, Val: metricName, Pos: 0}
+			// The code below re-orders the lex items in the items array and update the same. The re-ordering of
+			// lex items should be such that the key of any key-value pair if is `__name__` and its corresponding value
```
brian-brazil, line 209: This doesn't parse

promql/prettier/prettier_test.go
```diff
@@ -0,0 +1,966 @@
...
+		expr:     `quantile(0.9, sum(min(go_alloc_bytes) by (job)) without(job)) by (localhost)`,
+		expected: `quantile by(localhost) (0.9, sum without(job) (min by(job) (go_alloc_bytes)))`,
+	},
+	{
+		expr:     `quantile(0.9, sum(min(go_alloc_bytes) by (job)) ignoring(job)) by (localhost)`,
+		expected: `quantile by (localhost) (0.9, sum ignoring(job) (min by (job) (go_alloc_bytes)))`,
+	},
+	{
+		expr:     `quantile(0.9, sum(min(go_alloc_bytes) by (job)) ignoring(job)) by (localhost)`,
+		expected: `quantile by (localhost) (0.9, sum ignoring(job) (min by (job) (go_alloc_bytes)))`,
```
brian-brazil, line 70: localhost is not a good label name to use in examples, as it's a label value.

promql/prettier/prettier_test.go
```diff
@@ -0,0 +1,966 @@
...
+    rev="34d0f99",
+    env="prod",
+    job="cluster-manager",
+    host="localhost",
+  }`,
+	},
+	{
+		expr: `instance_cpu_time_ns{app="lion", proc="web", rev="34d0f99", env="prod", job="cluster-manager", host="localhost",}`,
+		expected: `  instance_cpu_time_ns{
+    app="lion",
```
brian-brazil, line 277: This is more vertical than it needs to be, wrap so there's more than one matcher per line.

promql/prettier/prettier_test.go
```diff
@@ -0,0 +1,966 @@
...
+	{
+		expr: `metric_one{_a_some_very_large_label="_a_very_large_value", some_very_large_label="a_very_large_value"} + 1`,
+		expected: `    metric_one{
+      _a_some_very_large_label="_a_very_large_value",
+      some_very_large_label="a_very_large_value",
+    } + 1`,
+	},
+	{
+		expr:     `metric_one[5m]`,
+		expected: `  metric_one[5m]`,
```
brian-brazil, line 453: There shouldn't be an indent here.

promql/prettier/prettier_test.go
```diff
@@ -0,0 +1,966 @@
...
+	{
+		expr: `histogram_quantile(0.9, rate(http_request_duration_seconds_bucket[10m]))`,
+		expected: `  histogram_quantile(
+    0.9,
+    rate(http_request_duration_seconds_bucket[10m])
+  )`,
+	},
+	{
+		expr: `label_join(up{job="api-server",src1="a",src2="b",src3="c"}, "foo", ",", "src1", "src2", "src3")`,
+		expected: `  label_join(
```
brian-brazil, line 550: This shouldn't be indented either.

promql/prettier/prettier_test.go
```diff
@@ -0,0 +1,966 @@
...
+		expr:     `quantile(0.9, sum(min(go_alloc_bytes) by (job)) without(job)) by (localhost)`,
+		expected: `quantile by (localhost) (0.9, sum without(job) (min by (job) (go_alloc_bytes)))`,
+	},
+	{
+		expr:     `quantile(0.9, sum(min(go_alloc_bytes) by (job)) without(job)) by (localhost)`,
+		expected: `quantile by(localhost) (0.9, sum without(job) (min by(job) (go_alloc_bytes)))`,
+	},
+	{
+		expr:     `quantile(0.9, sum(min(go_alloc_bytes) by (job)) ignoring(job)) by (localhost)`,
+		expected: `quantile by (localhost) (0.9, sum ignoring(job) (min by (job) (go_alloc_bytes)))`,
```
brian-brazil, line 66: It'd be too to see how this looks with the full prettier.

cmd/promtool/main.go
Harkishen-Singh: Currently, the reason why we are not stopping at that point is, we want to write those files that are perfect to do so and print errors for the ones that cause any error.

Should we stop the moment we encounter error while writing any file, without caring about the remaining ones?

promql/prettier/node.go
Harkishen-Singh: I will update it to 2020.

promql/prettier/prettier.go
Harkishen-Singh: I didn't get the point. Could you please explain?

promql/prettier/prettier_test.go
Harkishen-Singh: Sure. Replacing with `instance`.

promql/prettier/prettier_test.go
Harkishen-Singh: This is the prettified output. I will add the same as a unit-test.

`quantile by(instance) (0.9, sum without(job) (min by(job) (go_alloc_bytes)))`

promql/prettier/node.go
Harkishen-Singh: updated to `compareNodeType`. How does that sound?

cmd/promtool/main.go
brian-brazil: For I/O errors we should probably stop.

promql/prettier/node.go
brian-brazil: Compare implies you're comparing two nodes. isNodeType should do it.

promql/prettier/prettier_test.go
brian-brazil: That's very dense semantically, you could easily miss the nonsense aggregation here. I think there shouldn't be more than one aggregation per line.

promql/prettier/prettier.go
brian-brazil: This sentence makes no sense grammatically.

cmd/promtool/main.go
brian-brazil: Also this way of writing files is not atomic, so a failure could lead to data loss.

promql/prettier/node.go
Harkishen-Singh: Oh yes.

promql/prettier/prettier_test.go
Harkishen-Singh: Do you mean, for those vector_selectors that violate column limit, keep their label_matchers on a different line only? 
Example: for the current scenario,

```
instance_cpu_time_ns{
  app="lion", proc="web", rev="34d0f99", env="prod", job="cluster-manager", host="localhost",
}
```

brian-brazil: yeah, avoid having one matcher per line anyway

Harkishen-Singh:
@brian-brazil, I think it will be tough to replace `printer.go` with `prettier`. This is because prettier requires 2 inputs - Node and Lex items of that node. But, I don't think we can get lex items of a node, especially when the node is recursive in nature, eg: Aggregate expr.

codesome:
The current printer is `Note -> string`, but this prettier needs the lex items specific to that node, so the prettier is `string -> string` where it gets the node and the lex items inside.

Prettier can also be `(Node, lex items) -> string`, but it's hard to get the exact set of lex items of a Node when you go into the child node. 

@brian-brazil do you have any suggestions? (or need more info?)

brian-brazil:
We only need to support printing the top level node, as that's what's we use in the UI.

promql/prettier/prettier_test.go
Harkishen-Singh: Addressed. Now, if any aggregate expression has an aggregate expression as an immediate parent, it will be printed on a separate line.

promql/prettier/prettier_test.go
Harkishen-Singh: Addressed in recent commits. Please resolve those comments that are satisfactory.

promql/prettier/prettier_test.go
Harkishen-Singh: Addressed in recent commits.

promql/prettier/prettier_test.go
Harkishen-Singh: Addressed. Now, all label marchers will be in a single newline (like the mentioned example in the above comment), rather than putting them on new lines each time.

promql/prettier/prettier.go
Harkishen-Singh: Updated the explanation. Please give it a read and comment on its satisfactoriness.

promql/prettier/prettier.go
Harkishen-Singh: Addressed this as well. Please comment on its satisfactoriness.

promql/prettier/node.go
Harkishen-Singh: Addressed this to `isNodeType`. Hence, resolving.

cmd/promtool/main.go
Harkishen-Singh: Addressed this. Now, this will stop and `return 1` in case of an error.

promql/prettier/prettier_test.go
Harkishen-Singh: This is kept since while printing in block scalar format, we are adding 2 spaces to give a better look.

Harkishen-Singh: Also, this has a depth of 1, so `2 spaces` in the AST, keeping things simple.

cmd/promtool/main.go
brian-brazil: The writing of the files is still not atomic.

promql/prettier/prettier_test.go
brian-brazil: I'm not following at all, there's no need for indenting here. If a caller wants to indent more, that's up to them.

promql/prettier/prettier.go
```diff
@@ -0,0 +1,545 @@
...
+	// labelsSplittable determines whether the current formatting region lies in labels
+	// matchers in VectorSelector. Since labels in VectorSelectors are only splittable,
+	// is property is applied only after encountering a Left_Brace. This is dependent across
+	// multiple padder states and hence needs to be carried along.
+	labelsSplittable bool
+	isFirstLabel     bool
+}
+
+// New returns a new prettier.
+func New(expression string, columnLimit int, indentType uint) *Prettier {
```
brian-brazil, line 69: How many spaces to indent should not be configurable, and I'm not sure about the column limit either. Consistency is important for this sort of thing.

promql/prettier/prettier.go
```diff
@@ -0,0 +1,545 @@
...
+				continue
+			}
+			itr = advanceComments(items, itr)
+			if items[itr+3].Typ == parser.COMMA {
+				skipBraces = rightBraceIndex-5 == advanceComments(items, leftBraceIndex)
+			} else {
+				skipBraces = rightBraceIndex-4 == advanceComments(items, leftBraceIndex)
+			}
+			identifierItem := parser.Item{Typ: parser.IDENTIFIER, Val: metricName, Pos: 0}
+			// The code below re-orders the lex items in the items array. The re-ordering of
```
brian-brazil, line 224: Still doesn't parse - that comma shouldn't be there.

promql/prettier/prettier.go
roidelapluie: We should either not put a limit or column or fix it at 80 as it's pretty standard.

I agree that this should not be configurable.

brian-brazil: 80 is pretty narrow these days. I'd suggest hardcoding to something in the 100-120 range.

slrtbtfs: Note, that embedding a PromQL query in YAML will add about 10 char widths of indentation.

cmd/promtool/main.go
brian-brazil: Yeah, consider what happens if you run out of disk space half way through writing the file.

promql/prettier/prettier_test.go
Harkishen-Singh: Please resolve if it is satisfactory.

cmd/promtool/main.go
Harkishen-Singh: Writing rules file now done using temp file and then renaming it. I have, that addresses our atomic issue.

promql/parser/prettier.go
```diff
@@ -0,0 +1,579 @@
...
+		return p.expression, err
+	}
+	formattedExpr, err := p.prettify(p.lexItems(standardizeExprStr))
+	if err != nil {
+		return p.expression, errors.Wrap(err, "Prettify")
+	}
+	return formattedExpr, nil
+}
+
+func (p *Prettier) sortItems(items []Item) []Item {
```
codesome, line 86: Can you describe what do you mean by sorting here? If it is the re-arrangement of lex items, then we might as well call it `rearrangeItems` (with comments ofcourse).

codesome (commented):
First high level pass. I haven't looked at prettier and nodeInfo implementation yet.

promql/parser/parse.go
```diff
@@ -117,7 +117,7 @@ func ParseExpr(input string) (expr Expr, err error) {
 	if len(p.parseErrors) == 0 {
 		p.checkAST(expr)
 	}
-
+	setExprStr(&expr, input)
```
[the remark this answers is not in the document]
Harkishen-Singh: I had done that at first, but it leads to OOM (as said in chat). Sounds strange but that's the reason why its kept outside. But, when I comment out the last two tests (mentioned below) in parse_test.go, it works fine.

```
{
		input:  "test[5d] OFFSET 10s [10m:5s]",
		fail:   true,
		errMsg: `1:1: parse error: subquery is only allowed on instant vector, got matrix in "test[5d] OFFSET 10s [10m:5s]"`,
	},
	{
		input:  `(foo + bar{nm="val"})[5m:][10m:5s]`,
		fail:   true,
		errMsg: `1:1: parse error: subquery is only allowed on instant vector, got matrix in "(foo + bar{nm="val"})[5m:][10m:5s]" instead`,
	},
```

promql/parser/parse.go
```diff
@@ -178,6 +178,31 @@ func newParser(input string) *parser {
 	return p
 }
 
+func setExprStr(expr *Expr, input string) {
+	switch e := (*expr).(type) {
+	case *AggregateExpr:
```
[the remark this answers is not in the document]
Harkishen-Singh: I have tried that way at first. But it turns out to be an error, and hence had to reference them in individual cases.

promql/parser/printer.go
```diff
@@ -15,14 +15,14 @@ package parser
...
+	"github.com/pkg/errors"
+	"github.com/prometheus/common/promlog"
 	"strings"
-	"time"
-
-	"github.com/prometheus/common/model"
-	"github.com/prometheus/prometheus/pkg/labels"
 )
 
+var logger = promlog.New(&promlog.Config{})
```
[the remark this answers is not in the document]
Harkishen-Singh: Shouldn't we inform the user in case of some error during the prettifying process?

promql/parser/printer.go
```diff
@@ -61,115 +61,91 @@ func (es Expressions) String() (s string) {
...
 }
 
 func (node *Call) String() string {
-	return fmt.Sprintf("%s(%s)", node.Func.Name, node.Args)
+	prettyExpr, err := PromqlPrettier(node.ExprString()).Prettify()
+	if err != nil {
+		level.Error(logger).Log(errors.Wrap(err, "pretty-print"))
+		return node.ExprString()
+	}
+	return prettyExpr
```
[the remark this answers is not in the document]
Harkishen-Singh: I had thought of it but keeping this in a single function will lead to the function having a switch-case block like in `setExprStr()`, where each expression will have to be in a different case block. I hope that will be satisfactory.

promql/parser/parse.go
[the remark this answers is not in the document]
codesome: Oh I did not notice the `nil` check inside `setExprStr`. It should be fine and no panics.

> Sounds strange

Maybe something related to `checkAST` which needs to be done before setting? If that is the case, can you add a comment here mentioning that?

promql/parser/parse.go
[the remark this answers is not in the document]
codesome: I meant
```
switch ... {
case *AggregateExpr, *BinaryExpr, *Call,
         *SubqueryExpr ... :
    e.SetExprString(input)
}
```

promql/parser/printer.go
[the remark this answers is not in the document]
codesome: When called `String()`, I would expect any form of string representation and won't care if it's prettified, as long as the output is deterministic. Also, why would prettifying a valid expression error out?

promql/parser/printer.go
```diff
@@ -61,115 +56,72 @@ func (es Expressions) String() (s string) {
...
+		expression = n.ExprString()
+	case *StringLiteral:
+		expression = n.ExprString()
+	case *VectorSelector:
+		expression = n.ExprString()
+	case *ParenExpr:
+		expression = n.ExprString()
+	case *UnaryExpr:
+		expression = n.ExprString()
+	}
```
codesome, line 121: > We could have a single function which takes `node.ExprString()` as input

^ this

I mean have a common method like this:

```
func methodName(expr string) string {
	prettyExpr, err := PromqlPrettier(expr).Prettify()
	if err != nil {
		return expr
	}
	return prettyExpr
}
```

And call `return methodName(node.ExprString())` everywhere.

promql/parser/node.go
```diff
@@ -0,0 +1,194 @@
...
+	return temp
+}
+
+func isNodeType(node Node, typ uint) bool {
+	switch typ {
+	case aggregateExpr:
+		if _, ok := node.(*AggregateExpr); ok {
+			return true
+		}
+		return false
```
codesome, line 176: This line is redundant in all the `case`. The `return false` at the end is enough.

promql/parser/printer.go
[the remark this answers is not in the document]
Harkishen-Singh: There should not be any error since the only errors possible is from the parser and the AST which are already checked well before entering the `.String()`. So, yes there will not be any errors as returns from any underlying function.

promql/parser/parse_test.go
```diff
@@ -2642,14 +2757,16 @@ var testExpr = []struct {
 			Offset: 10 * time.Minute,
 			EndPos: 37,
 		},
-	}, {
+	},
+	{
```
codesome, line 2761: Spurious change

promql/parser/parse_test.go
```diff
@@ -2642,14 +2757,16 @@ var testExpr = []struct {
...
 		input:  "test[5d] OFFSET 10s [10m:5s]",
 		fail:   true,
-		errMsg: "1:1: parse error: subquery is only allowed on instant vector, got matrix in \"test[5d] offset 10s[10m:5s]\"",
-	}, {
+		errMsg: `1:1: parse error: subquery is only allowed on instant vector, got matrix in ""`,
+	},
+	{
```
codesome, line 2766: Spurious change

promql/parser/parse.go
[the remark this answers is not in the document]
Harkishen-Singh: > can you add a comment here mentioning that?

As comment in the code?

codesome (commented):
I will require a few more passes to review all changes. A small second pass for now.

promql/parser/parse.go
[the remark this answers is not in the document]
Harkishen-Singh: > I meant

Yeah, I had done what is suggested but the output is a compilation error (mentioned in the second comment).

Screenshot from 2020-08-28 19-47-45

promql/parser/prettier.go
```diff
@@ -0,0 +1,575 @@
...
+	if result[len(result)-1] == ' ' {
+		result = result[:len(result)-1]
+	}
+	return result
+}
+
+type itemsIterator struct {
+	itemsSlice []Item
+	index      int
+}
```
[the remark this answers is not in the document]
Harkishen-Singh: Nothing special use case as such. Just makes it easier to understand since an Iterator. Should I remove `itemsIterator` and just keep the `for` loop?

promql/parser/node.go
```diff
@@ -0,0 +1,194 @@
...
+	buf         int
+	baseIndent  int
+}
+
+func (n *nodeInfo) violatesColumnLimit() bool {
+	if n.currentNode == nil {
+		panic("current node not set")
+	}
+	var (
+		it    = itemsIterator{n.items, -1}
```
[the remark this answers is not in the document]
Harkishen-Singh: Yes, we can. In our earlier discussion, we planned to use an iterator to iterate through the items, so this is implemented.

promql/parser/prettier.go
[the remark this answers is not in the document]
codesome: Yes, just a `for` loop. A simple loop is easier to understand than an iterator as iterators can have magic underneath ;)

Harkishen-Singh:
@brian-brazil should we format `sum without() (metric_name)`  to `sum(metric_name)`?

brian-brazil:
> @brian-brazil should we format sum without() (metric_name) to sum(metric_name)?

Those mean different things. `sum by () (metric_name)` is the exact same as `sum(metric_name)` though.

Harkishen-Singh:
Can you list those grouping modifiers that when empty have the same use-case as not used, like in case of `BY`?

brian-brazil:
It's only `by`.

promql/parser/node.go
```diff
@@ -0,0 +1,194 @@
...
+// isLastItem returns true if the item passed is the last item of the current node.
+func (n *nodeInfo) isLastItem(item Item) bool {
+	return n.currentNode.PositionRange().End == item.PositionRange().End
+}
+
+func (n *nodeInfo) previousIndent() int {
+	return n.buf
+}
+
+// has verifies whether the current node contains a particular entity.
```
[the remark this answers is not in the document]
Harkishen-Singh: Done. Please have a look and comment on its satisfactoriness.

promql/parser/parse.go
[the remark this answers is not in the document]
Harkishen-Singh: As per discussions in chat, this is now resolved.

promql/parser/printer.go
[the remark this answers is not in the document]
Harkishen-Singh: Addressed this. Hence, resolving.

promql/parser/prettier.go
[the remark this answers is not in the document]
Harkishen-Singh: Addressed. Hence, resolving.

promql/parser/parse.go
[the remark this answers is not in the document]
codesome: Yes

promql/parser/prettier.go
```diff
@@ -0,0 +1,562 @@
...
+				)
+				// Get index of the closing paren.
+				for j := keywordIndex; j <= len(items); j++ {
+					if items[j].Typ == RIGHT_PAREN {
+						closingParenIndex = j
+						break
+					}
+				}
+				if closingParenIndex == itemNotFound {
+					panic("invalid paren index: closing paren not found")
```
codesome, line 145: Let's not panic anywhere here and rather return errors. The user of `Prettify` would not be expecting a panic :)

promql/parser/prettier.go
```diff
@@ -0,0 +1,562 @@
...
+				for itemp := 0; itemp < len(items); itemp++ {
+					if itemp <= aggregatorIndex || itemp > closingParenIndex {
+						tempItems = append(tempItems, items[itemp])
+					} else if !finishedKeywordWriting {
+						// Immediately after writing the aggregator, write the keyword and
+						// the following expression till the closing paren. This is done
+						// to be in-order with the expected result.
+						for j := keywordIndex; j <= closingParenIndex; j++ {
+							tempItems = append(tempItems, items[j])
+						}
```
codesome, line 157: You should be able to do this in a single line.

promql/parser/prettier.go
```diff
@@ -0,0 +1,562 @@
...
+				for j := keywordIndex; j <= len(items); j++ {
+					if items[j].Typ == RIGHT_PAREN {
+						closingParenIndex = j
+						break
+					}
+				}
+				if closingParenIndex == itemNotFound {
+					panic("invalid paren index: closing paren not found")
+				}
+				// Re-order lexical items. TODO: consider using slicing of lists.
```
codesome, line 147: Let's do this TODO in this PR itself and get rid of the loop. The entire loop can be replaced with 4 lines.

1. Insert everything before `aggregatorIndex`.
2. Insert from `keywordIndex` to `closingParenIndex`
3. Insert from `aggregatorIndex` to `keywordIndex`
4. Insert everything beyond `closingParenIndex`

With comment on each line about what part is it handling. Else the loop currently is difficult to read.

promql/parser/prettier.go
```diff
@@ -0,0 +1,562 @@
...
+					break
+				}
+			}
+			if leftParenIndex+1 == rightParenIndex || leftParenIndex == rightParenIndex {
+				for index := 0; index < len(items); index++ {
+					if index >= groupingKeywordIndex && index <= rightParenIndex {
+						continue
+					}
+					tempItems = append(tempItems, items[index])
+				}
```
codesome, line 264: Let's use the slice tricks here too.

codesome (commented):
Another round

web/api/v1/api_test.go
```diff
@@ -1354,7 +1354,7 @@ func testEndpoints(t *testing.T, api *API, tr *testTargetRetriever, testLabelAPI
 							alertingRule{
 								State:       "inactive",
 								Name:        "test_metric4",
-								Query:       "up == 1",
+								Query:       "up== 1",
```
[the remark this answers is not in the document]
codesome: But why would the test case change here?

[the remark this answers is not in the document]
Harkishen-Singh: `==` is considered as `EQL` item by the lexer. This is good. But, `=` is also considered as `EQL`. When used in labels, it should result in no spaces in labels like `<label_name>=<label_value>`. Now here comes the issue. This will mean that where ever you encounter EQL, don't leave spaces before and after an EQL. 

Now in case of `up == 1`, up is IDENTIFIER. `==` is `EQL`, so it is printed as if it is the case of a label_matchers. And, `1` is numeric literal, so it has space before itself.

I hope, this makes sense.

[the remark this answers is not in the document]
codesome: > don't leave spaces before and after an EQL.

Right, the `after` is missing here. Can we tie the spacing logic to the `EQL`/`ASSIGN` items and not the RHS so that it is same on both sides always? `up== 1` is not pretty :)

promql/parser/prettier.go
Harkishen-Singh: Done!

promql/parser/prettier.go
Harkishen-Singh: Done, hence resolving.

Harkishen-Singh:
@codesome implemented all the suggestions!

web/api/v1/api_test.go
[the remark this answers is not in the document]
Harkishen-Singh: Ahh, I don't think how reliable that will be. But yes, whenever we encounter EQL, we can check if the value is `=` or `==` and based on the value, it will work perfectly. But then, the person (I can take it if we recognize that as a bug) who solves that issue should make changes here when that issue is fixed.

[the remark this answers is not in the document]
codesome: We can fix that spacing with the bug fix later. What I meant to say is we avoid `And, 1 is numeric literal, so it has space before itself`, which sounds like we look at the number and add a space. Rather, see that it's EQL/ASSIGN and either add spaces on both sides or dont add at all.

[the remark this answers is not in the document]
Harkishen-Singh: These are fixed now. Hence, resolving.

cmd/promtool/main.go
```diff
@@ -18,6 +18,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"gopkg.in/yaml.v3"
```
codesome, line 21: Wrong spot

promql/parser/prettier.go
```diff
@@ -0,0 +1,557 @@
...
+				if j <= rightBraceIndex && j >= leftBraceIndex {
+					// Before printing the left_brace, we print the metric_name. After this, we check for metric_name{}
+					// situation. If __name__ is the only label_matcher, we skip printing '{' and '}'.
+					if j == leftBraceIndex {
+						tmp = append(tmp, identifierItem)
+						if !skipBraces {
+							tmp = append(tmp, items[j])
+						}
+						continue
+					} else if items[itr+3].Typ == COMMA && j == itr+3 || j >= itr && j < itr+3 {
```
codesome, line 223: 1. `else` is not require, you can just have the `if`.
2. Please add relevant `( )` brackets else its hard to read. Also, I think you want to check `j == itr+3` before `items[itr+3].Typ == COMMA` to avoid any panics (even if it won't cause panics, it is uncomfortable to read anyway :)).

promql/parser/prettier.go
```diff
@@ -0,0 +1,557 @@
...
+			// The leftBraceIndex and rightBraceIndex that form a range. Since the position of items beyond
+			// this range do not change, they are simply copied. However, the items within the range require
+			// re-ordering. So, they are added to the lex item slice after appending the identifier (metric_name)
+			// to the lex item slice.
+			//
+			// Also, we need to avoid `metric_name{}` case while re-ordering a `{__name__="metric_name"}`.
+			// Hence, if the indexes of left and right braces are next to each other, we skip
+			// appending them (braces) to the items slice.
+			for j := range items {
+				if j <= rightBraceIndex && j >= leftBraceIndex {
```
codesome, line 214: We can use few slice tricks here too. Append everything before the `rightBraceIndex` to `tmp`, then loop from right to left brace index, then another slice trick to append everything after `leftBraceIndex`.

promql/parser/prettier.go
```diff
@@ -0,0 +1,557 @@
...
+					break
+				}
+			}
+			if leftParenIndex+1 == rightParenIndex || leftParenIndex == rightParenIndex {
+				tempItems := make([]Item, len(items[:groupingKeywordIndex]))
+				// Copy items which are just before keyword.
+				copy(tempItems, items[:groupingKeywordIndex])
+				// Append items after the RIGHT_PAREN of grouping_modifier.
+				tempItems = append(tempItems, items[rightParenIndex+1:]...)
+				items = tempItems
```
codesome, line 256:
```suggestion
				items = append(items[:groupingKeywordIndex], items[rightParenIndex+1:]...)
```

promql/parser/prettier.go
```diff
@@ -0,0 +1,557 @@
...
+			for j := keywordIndex; j <= len(items); j++ {
+				if items[j].Typ == RIGHT_PAREN {
+					closingParenIndex = j
+					break
+				}
+			}
+			if closingParenIndex == itemNotFound {
+				return items, errors.New("invalid paren index: closing paren not found")
+			}
+			tempItems := make([]Item, len(items[:aggregatorIndex+1]))
```
codesome, line 146: Because we are appending more later, so it will cause re-allocation anyway. A little extra allocation here is fine. (And start with 0 len so that we can be consistent and use `append` everywhere)
```suggestion
			tempItems := make([]Item, 0, len(items))
```

promql/parser/prettier.go
```diff
@@ -0,0 +1,557 @@
...
+			copy(tempItems, items[:aggregatorIndex+1])
+			// Append items to itemItems from the keyword till the RIGHT_PAREN of the grouping_modifier.
+			tempItems = append(tempItems, items[keywordIndex:closingParenIndex+1]...)
+			// Append the remaining items of the node.
+			for itemp := aggregatorIndex + 1; itemp < len(items); itemp++ {
+				if itemp >= keywordIndex && itemp <= closingParenIndex {
+					continue
+				}
+				tempItems = append(tempItems, items[itemp])
+			}
```
codesome, line 158:
```suggestion
			tempItems = append(tempItems, items[aggregatorIndex+1:keywordIndex]...)
			tempItems = append(tempItems, items[closingParenIndex+1:]...)
```

promql/parser/prettier.go
```diff
@@ -0,0 +1,557 @@
...
+					break
+				}
+			}
+			if closingParenIndex == itemNotFound {
+				return items, errors.New("invalid paren index: closing paren not found")
+			}
+			tempItems := make([]Item, len(items[:aggregatorIndex+1]))
+			// Copy all the items in the items slice from starting till the aggregatorIndex
+			// by value.
+			copy(tempItems, items[:aggregatorIndex+1])
```
codesome, line 149: For consistency
```suggestion
			tempItems = append(tempItems, items[:aggregatorIndex+1]...)
```

promql/parser/prettier.go
```diff
@@ -0,0 +1,557 @@
...
+			}
+			if leftBraceIndex == itemNotFound || rightBraceIndex == itemNotFound {
+				continue
+			}
+			itr = skipComments(items, itr)
+			if items[itr+3].Typ == COMMA {
+				skipBraces = rightBraceIndex-5 == skipComments(items, leftBraceIndex)
+			} else {
+				skipBraces = rightBraceIndex-4 == skipComments(items, leftBraceIndex)
+			}
```
codesome, line 199: Does it skip braces in case of `{__name__="metric_name",}`?

codesome (commented):
An incremental review just on `rearrangeItems`. It is more intricate than I would like, but that's what you get when playing directly with lex items I guess.

promql/parser/prettier.go
Harkishen-Singh: This is causing errors in unit-tests.

promql/parser/prettier.go
Harkishen-Singh: Yes, it does. I have added a unit test for the exact purpose.

Harkishen-Singh:
@codesome done!

promql/parser/prettier.go
```diff
@@ -0,0 +1,543 @@
...
+		switch item.Typ {
+		case EQL, EQL_REGEX, NEQ, NEQ_REGEX, COLON, LEFT_BRACE, LEFT_BRACKET, LEFT_PAREN,
+			RIGHT_BRACE, RIGHT_BRACKET, RIGHT_PAREN, STRING, NUMBER, IDENTIFIER, METRIC_IDENTIFIER:
+			expression += item.Val
+		case COMMA:
+			expression += item.Val + " "
+		case COMMENT:
+			expression += item.Val + "\n"
+		default:
+			expression += " " + item.Val + " "
```
codesome, line 514: And you can remove the `if` condition above
```suggestion
		case COMMENT:
			expression += item.Val + "\n"
		default:
			expression += item.Val
```

promql/parser/prettier.go
```diff
@@ -0,0 +1,543 @@
...
+		}
+	}
+	return expression
+}
+
+// pad provides an instantaneous padding.
```
codesome, line 520:
```suggestion
// pad returns `iter` times the configured indent.
```

promql/parser/prettier.go
```diff
@@ -0,0 +1,543 @@
...
+	for i := 1; i <= iter; i++ {
+		pad += pd.indent
+	}
+	return pad
+}
+
+func (pd *padder) newLine() string {
+	pd.isNewLineApplied = true
+	return "\n"
+}
```
codesome, line 532: We could get rid of this method and use "\n" inlined for easy reading of code. Additionally, `isNewLineApplied` could be a local variable to `prettify` method.

promql/parser/prettier.go
```diff
@@ -0,0 +1,543 @@
...
+		result = ""
+		root   struct {
+			node      Node
+			hasScalar bool
+		}
+	)
+	for i, item := range items {
+		var (
+			nodeInfo       = &nodeInfo{head: p.Node, columnLimit: 100, item: item, items: items}
+			node           = nodeInfo.node()
```
codesome, line 260:
```suggestion
			nodeOfCurrentItem           = nodeInfo.node()
```

codesome, line 260:
```suggestion
			nodeOfCurrItem           = nodeInfo.node()
```

promql/parser/prettier.go
```diff
@@ -0,0 +1,543 @@
...
+			nodeSplittable = nodeInfo.violatesColumnLimit()
```
codesome, line 261: We can get rid of this variable and inline `nodeInfo.violatesColumnLimit()` wherever is it used (which is not much)

promql/parser/prettier.go
```diff
@@ -0,0 +1,543 @@
...
+			// Binary expression use-case.
+			hasImmediateScalar               bool
```
codesome, line 263: Else it's confusing what variables this comment refers to
```suggestion
			hasImmediateScalar               bool // For binary expression.
```

promql/parser/prettier.go
```diff
@@ -0,0 +1,543 @@
...
+			hasGrouping, aggregationGrouping bool
+			// Aggregate expression use-case.
+			isAggregation, isParen, hasMultiArgumentCalls bool
```
codesome, line 266: As discussed offline we could improve these namings and maybe also add comments the context of each variable.

promql/parser/prettier.go
```diff
@@ -0,0 +1,543 @@
...
+				result += " " + item.Val + " "
+			} else {
+				result += item.Val
+			}
+		case ADD, SUB, MUL, DIV, GTE, GTR, LOR, LAND,
+			LSS, LTE, LUNLESS, MOD, POW:
+			if hasImmediateScalar {
+				result += " " + item.Val + " "
+				hasImmediateScalar = false
+			} else {
```
codesome, line 386: You can add a `break` in the `if` and get rid of this `else`

codesome (commented):
Another round of review, getting close

promql/parser/prettier.go
Harkishen-Singh:
```
default:
			expression += item.Val
```

has to be 

```
default:
			expression += item.Val + " "
```
followed by trim of space, else things like `a + bool b` will become `a+boolb` which is unparsable.

promql/parser/prettier.go
Harkishen-Singh: Not sure but I feel the naming seems self-explanatory. Please correct me if I am wrong.

Harkishen-Singh:
@codesome done!

promql/parser/prettier.go
```diff
@@ -0,0 +1,543 @@
...
+}
+
+// isMetricNameAtomic returns true if the metric name is singular in nature.
+// This is used during sorting of lex items.
+func isMetricNameNotAtomic(metricName string) bool {
+	// We aim to convert __name__ to an ideal metric_name if the value of that labels is atomic.
+	// Since a non-atomic metric_name will contain alphabets other than a-z and A-Z including _,
+	// anything that violates this ceases the formatting of that particular label item.
+	// If this is not done then the output from the prettier might be an un-parsable expression.
+	if regexp.MustCompile("^[a-z_A-Z]+$").MatchString(metricName) && metricName != "bool" {
```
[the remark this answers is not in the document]
Harkishen-Singh: Seems like the mentioned regex is not happy with promql_tests. This might be a regression on our promql_tests. The earlier one passed easily, but if I replace with the one given in the link, it gives me an unparsable error. Strange.

codesome:
Update: I don't have much time in hand at the moment to review this, so if any maintainer wishes to review this, feel free!

Another point coming up from the contributor office hours is that I would like to see we be able to print child nodes too and not just parent nodes, need some discussion on how to achieve this.

bboreham:
Was this completed in #10544 ?  Can we close this PR?

beorn7:
I believe we still haven't included the formatting into `promtool`, but that should be done in a new PR at the state we are now. So closing this seems fine.

Issue #7762, opened 2020-08-06
Prometheus can ingest bad metrics name via recording rules


roidelapluie:
The following recording rule:

```
groups:
  - name: test
    evaluation_interval: 1s
    rules:
      - record: "a1"
        expr: vector(1)
        labels:
          __name__: -1
```
Creates a metric named -1. You can also create 1a, 1m, etc.

Tested on prometheus 2.20.

mneverov:
hi, i'd like to work on this

PR #7787, opened 2020-08-12
Fix bad metrics name

mneverov:
Add new fmt node rule: if a label has a name "__name__" then the value will be checked if it is a valid label name.
Fixes: #7762.

@roidelapluie could you please take a look?

### DIFF
--- a/pkg/rulefmt/rulefmt.go
+++ b/pkg/rulefmt/rulefmt.go
@@ -186,7 +186,7 @@ func (r *RuleNode) Validate() (nodes []WrappedError) {
 	}
 
 	for k, v := range r.Labels {
-		if !model.LabelName(k).IsValid() {
+		if !model.LabelName(k).IsValid() || k == model.MetricNameLabel {
 			nodes = append(nodes, WrappedError{
 				err: errors.Errorf("invalid label name: %s", k),
 			})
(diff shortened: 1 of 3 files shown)


brian-brazil:
We should reject such rules, the one in the record is the only thing that should apply.

roidelapluie:
> We should reject such rules, the one in the record is the only thing that should apply.

How many users would that break? That can be dynamic, e.g. a prod:status:{{$labels.env}} .

brian-brazil:
> How many users would that break?

It's not breaking users to reject a semantically invalid configuration.

> That can be dynamic, e.g. a prod:status:{{$labels.env}} .

That's also an anti-pattern, label values don't belong in the metric name.

mneverov:
what about annotaitons?

brian-brazil:
Recording rules can't have annotations.

mneverov:
Sorry, what i mean is alerting annotations. It is also possible to use `__name__` as an annotation name. Is it ok?

roidelapluie:
Yes

mneverov:
ok, thank you.
Since prefix `__` is reserved for internal use only, I will change the `rulefmt` to check if the label name isValid and the label does not have prefix `__`. 
Please, correct me if I misunderstood `We should reject such rules`

roidelapluie:
We only want to remove __name__

mneverov:
@roidelapluie thanks, got it.
PR is updated

roidelapluie:
:+1:

roidelapluie:
@brian-brazil do you want the documentation to reflect this "" new "" behavior?

brian-brazil:
I don't see a need to mention it, we already document where the metric name comes from.

roidelapluie:
Thanks @mneverov !

PR #7789, opened 2020-08-12
Use SAN in test certificate

roidelapluie:
go 1.15 deprecated the common name verification.


brian-brazil:
:+1:

PR #7790, opened 2020-08-12
Doc: Add a note about preserving name label in bin-op

### LINKED ISSUE #6929: BinaryExpr on __name__ fails
**What did you do?**
Ran the following query:

`(sum by(__name__) ({__name__=~"metric_.*"})) / (count by(__name__) ({__name__=~"metric_.*"}))`

A result

```
error evaluating query "(sum by(__name__) ({__name__=~\"metric_.*\"})) / (count by(__name__) ({__name__=~\"metric_.*\"}))" (line 13): found duplicate series for the match group {__name__="metric_10"} on the right hand-side of the operation: [{__name__="metric_10"}, {__name__="metric_1"}];many-to-many matching not allowed: matching labels must be unique on one side
```

**Environment**

* System information:

	n/a

* Prometheus version:

	2.10

* Alertmanager version:

	n/a

* Test file

I was able to reproduce this in a test case, below is the file. I expected both `sum by (a)` and `sum by (__name__)` to behave similarly (both work or both don't work).

```
load 5m
    metric_1{a="1"} 1
    metric_10{a="10"} 10
    metric_1000{a="1000"} 1000
    metric_10000{a="10000"} 10000

eval instant at 5m (sum by(a) ({__name__=~"metric_.*"})) / (count by(a) ({__name__=~"metric_.*"}))
    {a="1"} 1
    {a="10"} 10
    {a="1000"} 1000
    {a="10000"} 10000

eval instant at 5m (sum by(__name__) ({__name__=~"metric_.*"})) / (count by(__name__) ({__name__=~"metric_.*"}))
    metric_1{} 1
    metric_10{} 10
    metric_1000{} 1000
    metric_10000{} 10000
```

roidelapluie:
Fix #6929

### DIFF
--- a/docs/querying/operators.md
+++ b/docs/querying/operators.md
@@ -271,3 +271,6 @@ highest to lowest.
 Operators on the same precedence level are left-associative. For example,
 `2 * 3 % 2` is equivalent to `(2 * 3) % 2`. However `^` is right associative,
 so `2 ^ 3 ^ 2` is equivalent to `2 ^ (3 ^ 2)`.
+
+Only the operators who do not alter the values (level 4, 5 and 6 in the list
+above) preserve the metric names (`__name__` label) in the resulting vector.


docs/querying/operators.md
```diff
@@ -271,3 +271,6 @@ highest to lowest.
 Operators on the same precedence level are left-associative. For example,
 `2 * 3 % 2` is equivalent to `(2 * 3) % 2`. However `^` is right associative,
 so `2 ^ 3 ^ 2` is equivalent to `2 ^ (3 ^ 2)`.
+
+Only the operators who do not alter the values (level 4, 5 and 6 in the list
```
brian-brazil, line 275: We cover this topic above already, so it should get a mention in relation to arithmetic operators up in between an instant vector and a scalar above - rather than adding a new place.

This also has nothing to do with precedence.

roidelapluie:
closing in favour of #7819

PR #7792, opened 2020-08-13
tsdb: optimize code.

johncming:
optimize code for `walDir`


johncming:
> I would not call it optimize, but rather simplification but... LGTM (:

Thanks, simplification it's a more accurate word.

PR #7793, opened 2020-08-13
tsdb: use NewRangeHead instead.

johncming:
Use `NewRangeHead` instead of the old struct init way.

https://github.com/prometheus/prometheus/blob/40ace418d19a93ca07dfd9a6e40b02f0e38254c3/tsdb/head.go#L927-L933


bwplotka:
LGTM, we need this. Thanks! =D

Issue #7776, opened 2020-08-11
ChunkQuerier: Accessing returned chunk.Bytes() causes SegFault


bwplotka:
Looks like we are sharing some memory that gets deallocated, causing https://github.com/thanos-io/thanos/issues/3013

This looks like it's happening on querying during `compactHead` is finished processing. Probably mmaped chunks are the reason - those bytes has to be copied in the iterator.

pracucci:
Thanks for reporting it. Could this issue affect Prometheus as well?

bwplotka:
Yes, using remote-read.

Investigating compaction process WITH delete API coming in same moment (:

bwplotka:
Fix: https://github.com/prometheus/prometheus/pull/7795

bwplotka:
Looks like from all conversations: It works an intended. 

I described this behavior better and added tests against it here: https://github.com/prometheus/prometheus/pull/7795 but overall this ready is good to be closed (:

PR #7795, opened 2020-08-13
tsdb: Added tests for bytes access; added chunk.Clone method; Added more comments on behavior of Querier interfaces.

bwplotka:
Fixes: https://github.com/prometheus/prometheus/issues/7776

### DIFF
--- a/storage/generic.go
+++ b/storage/generic.go
@@ -21,8 +21,7 @@ import (
 )
 
 type genericQuerier interface {
-	LabelQuerier
-	Closer
+	LabelCloserQuerier
 	Select(bool, *SelectHints, ...*labels.Matcher) genericSeriesSet
 }
 
@@ -52,8 +51,7 @@ func (a *genericChunkSeriesSetAdapter) At() Labels {
 }
 
 type genericQuerierAdapter struct {
-	LabelQuerier
-	Closer
+	LabelCloserQuerier
 
 	// One-of. If both are set, Querier will be used.
 	q  Querier
@@ -68,11 +66,11 @@ func (q *genericQuerierAdapter) Select(sortSeries bool, hints *SelectHints, matc
 }
 
 func newGenericQuerierFrom(q Querier) genericQuerier {
-	return &genericQuerierAdapter{Closer: q, LabelQuerier: q, q: q}
+	return &genericQuerierAdapter{LabelCloserQuerier: q, q: q}
 }
 
 func newGenericQuerierFromChunk(cq ChunkQuerier) genericQuerier {
-	return &genericQuerierAdapter{Closer: cq, LabelQuerier: cq, cq: cq}
+	return &genericQuerierAdapter{LabelCloserQuerier: cq, cq: cq}
 }
 
 type querierAdapter struct {
--- a/storage/generic.go
+++ b/storage/generic.go
@@ -17,14 +17,12 @@
 package storage
 
 import (
-	"io"
-
 	"github.com/prometheus/prometheus/pkg/labels"
 )
 
 type genericQuerier interface {
 	LabelQuerier
-	io.Closer
+	Closer
 	Select(bool, *SelectHints, ...*labels.Matcher) genericSeriesSet
 }
 
@@ -55,7 +53,7 @@ func (a *genericChunkSeriesSetAdapter) At() Labels {
 
 type genericQuerierAdapter struct {
 	LabelQuerier
-	io.Closer
+	Closer
 
 	// One-of. If both are set, Querier will be used.
 	q  Querier
--- a/storage/generic.go
+++ b/storage/generic.go
@@ -17,11 +17,14 @@
 package storage
 
 import (
+	"io"
+
 	"github.com/prometheus/prometheus/pkg/labels"
 )
 
 type genericQuerier interface {
 	LabelQuerier
+	io.Closer
 	Select(bool, *SelectHints, ...*labels.Matcher) genericSeriesSet
 }
 
@@ -52,6 +55,7 @@ func (a *genericChunkSeriesSetAdapter) At() Labels {
 
 type genericQuerierAdapter struct {
 	LabelQuerier
+	io.Closer
 
 	// One-of. If both are set, Querier will be used.
 	q  Querier
@@ -66,11 +70,11 @@ func (q *genericQuerierAdapter) Select(sortSeries bool, hints *SelectHints, matc
 }
 
 func newGenericQuerierFrom(q Querier) genericQuerier {
-	return &genericQuerierAdapter{LabelQuerier: q, q: q}
+	return &genericQuerierAdapter{Closer: q, LabelQuerier: q, q: q}
 }
 
 func newGenericQuerierFromChunk(cq ChunkQuerier) genericQuerier {
-	return &genericQuerierAdapter{LabelQuerier: cq, cq: cq}
+	return &genericQuerierAdapter{Closer: cq, LabelQuerier: cq, cq: cq}
 }
 
 type querierAdapter struct {
--- a/storage/interface.go
+++ b/storage/interface.go
@@ -16,6 +16,7 @@ package storage
 import (
 	"context"
 	"errors"
+	"io"
 
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/tsdb/chunkenc"
@@ -68,9 +69,14 @@ type Queryable interface {
 type Querier interface {
 	LabelQuerier
 
+	// Closer releases the resources of the Querier.
+	io.Closer
+
 	// Select returns a set of series that matches the given label matchers.
 	// Caller can specify if it requires returned series to be sorted. Prefer not requiring sorting for better performance.
 	// It allows passing hints that can help in optimising select, but it's up to implementation how this is used if used at all.
+	// Returned iterators are invokable until Querier is closed. Returned bytes are read only.
+	// It's caller responsibility to copy data to avoid segmentation faults before closing queriers.
 	Select(sortSeries bool, hints *SelectHints, matchers ...*labels.Matcher) SeriesSet
 }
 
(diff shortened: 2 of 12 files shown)


bwplotka:
Not much difference in resources if not better (chunks are compacted better) so canceling prombench (: All seems to be good.

image

bwplotka:
Windows test fail was flakiness, so this is ready to be reviewed. (:

tsdb/querier.go
```diff
@@ -109,6 +109,7 @@ type blockQuerier struct {
 }
 
 // NewBlockQuerier returns a querier against the block reader and requested min and max time range.
+// It's caller responsibility to close ChunkQuerier to release resources (e.g close chunk reader)
```
roidelapluie, line 112:
```suggestion
// It is the caller's responsibility to close ChunkQuerier to release resources (e.g close chunk reader)
```

roidelapluie:
I might be wrong  but do we use that in prometheus? Because currently it does not seem like you close them.

roidelapluie:
Can you rebase, so we have green on windows?

bwplotka:
Done! :+1:

bwplotka:
> I might be wrong but do we use that in prometheus? Because currently it does not seem like you close them.

Missed this. Close what?

storage/interface.go
```diff
@@ -68,9 +69,13 @@ type Queryable interface {
 type Querier interface {
 	LabelQuerier
 
+	// Closer releases the resources of the Querier.
+	io.Closer
+
 	// Select returns a set of series that matches the given label matchers.
 	// Caller can specify if it requires returned series to be sorted. Prefer not requiring sorting for better performance.
 	// It allows passing hints that can help in optimising select, but it's up to implementation how this is used if used at all.
+	// Returned iterators are invokable only during lifetime of the Querier.
```
brian-brazil, line 78: It's not clear here that Close is what defines the lifecycle.

tsdb/querier.go
```diff
@@ -109,6 +109,7 @@ type blockQuerier struct {
 }
 
 // NewBlockQuerier returns a querier against the block reader and requested min and max time range.
+// It is the caller responsibility to close ChunkQuerier to release resources (e.g close chunk reader)
```
brian-brazil, line 112: i.e.

This doesn't return a ChunkQuerier

tsdb/querier.go
```diff
@@ -141,6 +142,7 @@ type blockChunkQuerier struct {
 }
 
 // NewBlockChunkQuerier returns a chunk querier against the block reader and requested min and max time range.
+// It's caller responsibility to close ChunkQuerier to release resources (e.g close chunk reader)
```
brian-brazil, line 145: I'd stop the sentence after resources, how this is implemented at lower levels doesn't belong in the function docs.

tsdb/querier.go
```diff
@@ -609,10 +611,16 @@ func (p *populateWithDelChunkSeriesIterator) Next() bool {
 
 	p.curr = p.currChkMeta
 	if p.currDelIter == nil {
-		return true
+		// Select ensures the return bytes data are accessible to user, so we need to copy the bytes that are mmapped at this point.
```
brian-brazil, line 614: This isn't how we handle this elsewhere, instead we make sure we don't deallocate until things are closed.

If a caller needs a copy it should be done by the caller, not in general code like this.

bwplotka: I thought about it and I decided that it's better to do it inside the API. The problem is that in many cases this data is copied (e.g from memSeries) already, so copying on caller side, is:

* In many cases duplicated alloc as memory is safe
* You dont' know what exactly to copy. Whole response? E.g how do i Know if strings behind labels are safe to use or should I copy them?
* Easy to miss, hard to test against.

What do you think? (: @brian-brazil

brian-brazil: I think we should be consistent, and leave it entirely up to the caller like we do in all other APIs.

Which are the callers having the issue here? Why can't they call Close at the right time?

bwplotka: Yea, we cannot because we reuse StoreAPI internally and we don't want to leak Close for this - we don't control the exact moment for serialization ):

bwplotka: or worse - copying bytes to TCP packet ;p

bwplotka: so at end we would need to copy bytes all perform deep copy of whole response going with this idea "caller is  responsible" which is extremely hard to  do

Plus, Why you say it's consistent? Select for samples iterator does not requires you to copy anything on caller side :thinking:

brian-brazil: > Yea, we cannot because we reuse StoreAPI internally and we don't want to leak Close for this - we don't control the exact moment for serialization ):

Where is this in Prometheus? We already have the plumbing for this in the API.

> Select for samples iterator does not requires you to copy anything on caller side

Yes it does, the strings for the label values can disappear after the Close.

bwplotka: > Where is this in Prometheus? We already have the plumbing for this in the API.

It's on Thanos: https://github.com/thanos-io/thanos/pull/3046/files#diff-be92e59b0e37723a0a0be5cebd88d264R456

> Yes it does, the strings for the label values can disappear after the Close.

Technically they are mmaped indeed... but somehow it's hard to repro segfault on those, I would be not surprised if we copy those strings accidentally somewhere. That's the point - it's so easy to miss and extra allocate.

OK, in this case I guess I need to remove copying here, and check how I can maintain open querier across internal StoreAPI call. Let' me adjust PR:

* new requirement / note in comments about this behaviour
* adjust tests for expecting segfault only when closed is invoked on those.

bwplotka: Adjusted all.

storage/generic.go
```diff
@@ -17,11 +17,14 @@
...
 
 import (
+	"io"
+
 	"github.com/prometheus/prometheus/pkg/labels"
 )
 
 type genericQuerier interface {
 	LabelQuerier
+	io.Closer
```
brian-brazil, line 27: We use the explicit method in all of our other interfaces, we don't use io.Closer anywhere.

tsdb/chunkenc/chunk.go
```diff
@@ -51,20 +51,28 @@ type Chunk interface {
...
+type Clonable interface {
 	// Compact is called whenever a chunk is expected to be complete (no more
 	// samples appended) and the underlying implementation can eventually
 	// optimize the chunk.
 	// There's no strong guarantee that no samples will be appended once
-	// Compact() is called. Implementing this function is optional.
+	// Compact() is called.
 	Compact()
+
+	// Clone return exact, compacted copy of chunk.
```
brian-brazil, line 74: Is this still needed?

tsdb/chunkenc/xor.go
```diff
@@ -79,14 +79,20 @@ func (c *XORChunk) NumSamples() int {
...
 	if l := len(c.b.stream); cap(c.b.stream) > l+chunkCompactCapacityThreshold {
-		buf := make([]byte, l)
-		copy(buf, c.b.stream)
-		c.b.stream = buf
+		c.b.stream = c.Clone().Bytes()
 	}
 }
 
+// Clone return exact, compacted copy of chunk.
+func (c *XORChunk) Clone() Chunk {
```
brian-brazil, line 90: If this is still needed, it should have a unittest.

storage/generic.go
bwplotka: Not in `LabelQuerier` anymore

tsdb/chunkenc/chunk.go
bwplotka: Yes, it would be nice to have on our side, plus separating those special ones like Compact, Clone sounds like a good idea.

tsdb/chunkenc/xor.go
bwplotka: ack let's do it!

storage/generic.go
bwplotka: fair point

bwplotka:
Addressed all (:

storage/generic.go
brian-brazil: This doesn't solve it, now you've defined a new interface which we use in some places but not others.

bwplotka:
ping (: @brian-brazil @krasi-georgiev

storage/generic.go
brian-brazil: This comment stands.

bwplotka: ack

bwplotka: fixed! Reverted and actually renamed interface to makes more (:

storage/interface.go
```diff
@@ -101,22 +95,19 @@ type ChunkQuerier interface {
...
-	// Close releases all resources required by queriers.
-	// This might be mean that other operations like head compaction or block release are
-	// blocked until this is closed.
-	Close() error
-}
-
-// LabelQuerier provides querying access over labels.
-type LabelQuerier interface {
+// LabelCloserQuerier provides querying access over labels.
+type LabelCloserQuerier interface {
```
brian-brazil, line 99: I don't think having Closer in here is useful, we don't call Blocks BlockCloser for example.

tsdb/db_test.go
```diff
@@ -2816,3 +2817,202 @@ func TestOpen_VariousBlockStates(t *testing.T) {
...
+			testutil.Equals(t, 4, len(chks))
+		}))
+	})
+
+	done := make(chan struct{})
+	t.Run("delete & close block and access results", func(t *testing.T) {
+		bl := db.Blocks()
+		go func() {
+			// This should block until all queries are closed.
+			testutil.Ok(t, db.deleteBlocks(map[ulid.ULID]*Block{
```
krasi-georgiev, line 2943: Why do we need this? What behaviour is it testing?

storage/interface.go
bwplotka: Yes, but LabelQuerier it's extremely confusing. It's so easier to forget you have to close those

brian-brazil: It's more confusing to arbitrarily have Close in some interface names and not others. Without is cleaner - and half the time it's not as if the type is going to show up in the code anyway for you to notice that Close needs to be called.

bwplotka:
Tests are improved, but oveall clone is not really needed at the end, so I will split this PR into smaller pieces and apply all comments :hugs:

PR #7796, opened 2020-08-13
README: Remove direct reference to debian packages

roidelapluie:
We respect and appreciate the work made by the packaging team downstream
in Debian, however it seems that the package is deviating in a way that
I'd like to remove the direct reference from our README:

- The debian package is deviating from upstream in multiple ways:
  - Some Service discoveries are missing (Azure, Kubernetes)
  - The lock file is disabled and the flag has been renamed
  - The libraries are not at the same release
- We have no control over the backports and the quick availability of
  updates
- There are other distributions with different packages and we do not
  include them either
- I expect Debian users to look in their distribution and just see the
  package there 'as usual'

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,6 @@ is the recommended way of installing Prometheus.
 See the [Installing](https://prometheus.io/docs/introduction/install/)
 chapter in the documentation for all the details.
 
-Debian packages [are available](https://packages.debian.org/sid/net/prometheus).
-
 ### Docker images
 
 Docker images are available on [Quay.io](https://quay.io/repository/prometheus/prometheus) or [Docker Hub](https://hub.docker.com/r/prom/prometheus/).


roidelapluie:
The debian package also does not offer the new UI.

brian-brazil:
:+1:

dswarbrick:
For the record (and speaking as one of the Debian prometheus package maintainers), we do not remove features simply for the fun of it. Prometheus has an enormous list of build dependencies, and it is not always feasible to support all features due to missing build-deps in Debian. In the past, packages such as the Azure SDK have also had backward incompatible changes, which has meant that Debian has had to refrain from updating that SDK out of concern for other applications that would break.

Debian policy dictates that all binaries must be buildable from source, so if a build-dep cannot be satisfied, and we don't have the manpower to package that build-dep, decisions have been made in the past to disable or patch out that feature. It's not ideal, but _usually_ it only affects a small number of users. This policy also extends to "transpiled" JavaScript such as TypeScript and Elm (used by the Alertmanager web UI). Elm has an equally long list of dependencies, and last time I checked, would require a newer version of Haskell (!) than what is currently in Debian.

roidelapluie:
I am aware of that. I am sincerely grateful to the debian team for offering this package in the distribution out of the box.

SuperQ:
While I appreciate the attempt to package things in Debian, I find the current deviation from the official builds to be problematic. It's not just bug fixes or small defaults changes, but serious things like the missing user interface.

Until these issues are fixed, I would probably call the problems "release critical", and suggest that Debian drop these packages.

beorn7:
In the context of this PR, I don't think we have to discuss what Debian should do with their Prometheus packages.

Even if the Debian situation were different, I think we should remove the reference, for reasons already mentioned, mainly that there are other distributions and that users of a distribution have their own ways of finding packages.

Perhaps we should mention that packages exist for various distributions, but that those might lack behind or miss features? This provides some more context for a naive user, who might assume any distribution package is identical to the latest official Prometheus release. (Less naive users don't need the hint anyway.)

roidelapluie:
> Perhaps we should mention that packages exist for various distributions, but that those might lack behind or miss features? This provides some more context for a naive user, who might assume any distribution package is identical to the latest official Prometheus release. (Less naive users don't need the hint anyway.)

Getting that wording right would take a couple of back-and-forth so maybe that is a topic to be discussed independently of this PR.

PR #7797, opened 2020-08-13
scrape: Add previous error to the API

roidelapluie:
First step towards #7384


scrape/target.go
```diff
@@ -264,6 +274,30 @@ func (t *Target) LastScrapeDuration() time.Duration {
 	return t.lastScrapeDuration
 }
 
+// PreviousError returns the error encountered during the last errorful scrape.
```
brian-brazil, line 277: The mix of last and previous here is confusing.

roidelapluie: Yes but the v1 HTTP API (used by dashboards and UIs) is considered stable, so I have to keep last in the JSON API.

Maybe I can change to `current` or `lastCompleted` in the go variables?

brian-brazil: The Go should follow the JSON, or that'd also be confusing.

roidelapluie:
WDYT about " Saved Error " ?

roidelapluie:
I have adapted the wordings

brian-brazil:
Most recent maybe?

roidelapluie:
target.MostRecentFailedScrapeDuration is a long var name but we can change that in the future - aka prometheus 3.0

docs/querying/api.md
```diff
@@ -428,9 +428,9 @@ $ curl http://localhost:9090/api/v1/targets
         "lastError": "",
         "lastScrape": "2017-01-17T15:07:44.723715405+01:00",
         "lastScrapeDuration": 0.050688943,
-        "savedError": "data does not end with # EOF",
-        "savedFailedScrape": "2017-01-17T15:02:44.723715405+01:00",
-        "savedFailedScrapeDuration": 0.010683421,
+        "mostRecentError": "data does not end with # EOF",
```
brian-brazil, line 431: mostRecentFailedScrapeError would be more consistent

PR #7799, opened 2020-08-14
some fixies for consul sd.

johncming:
1. Make the rpc duration more accurate.
2. Returns immediately  after the context has expired, rather than waiting for a for cycle


simonpasquier (approved):
LGTM

PR #7805, opened 2020-08-14
Fix detail swatch glitch

juliusv:


juliusv:
@roidelapluie

roidelapluie:
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.

Issue #7803, opened 2020-08-14
Text is cropped in react ui tooltips


roidelapluie:
**What did you do?**

Look the tooltips in the new UI

**What did you expect to see?**

Like in the old UI:

![tt2](https://user-images.githubusercontent.com/291750/90293962-9a5ca180-de85-11ea-9efd-18f4024c7c2e.png)

**What did you see instead? Under which circumstances?**

Text is cropped:

![tt1](https://user-images.githubusercontent.com/291750/90293972-a21c4600-de85-11ea-8bb5-998315fe2c07.png)

* Prometheus version:

	2.20.1

roidelapluie:
cc @juliusv

juliusv:
Interesting, it doesn't happen on the PromLabs one, and I've never seen it do that before: https://demo.promlabs.com/new/graph?g0.expr=up&g0.tab=0&g0.stacked=0&g0.range_input=1h

Wondering if it's a recent regression or something specific triggering that issue, checking...

juliusv:
According to `git bisect`, the commit that introduces this bug is this jquery version bump: https://github.com/prometheus/prometheus/commit/c7d9516695c51cdb2185166e34c8190feca0fdf8

PR #7807, opened 2020-08-15
Update various UI components

SuperQ:
* Correct jquery version in prom.lib.
* Update to bootstrap 4.5.2.
* Bump yarn versions.


juliusv:
Have you double checked locally that this does not introduce https://github.com/prometheus/prometheus/issues/7803 in the old web UI or prom.lib?

SuperQ:
@juliusv No, I'll double check.

SuperQ:
Looked at the issue, the change from 3.3.1 is for the console libraries, which from what I can tell are currently broken since we don't vendor the old version of jquery. This is just correcting the console libraries reference..

juliusv:
Ah yeah, in `web/ui/templates/_base.html` it's just the bootstrap update.

:+1:

roidelapluie:
Thanks for this!

PR #7808, opened 2020-08-16
discovery/refresh: make metric more accurate.

johncming:
Make metric `prometheus_sd_refresh_duration_seconds` more accurate. Because `d.refreshf` is the concrete implementation of the subclasses.

### DIFF
--- a/discovery/refresh/refresh.go
+++ b/discovery/refresh/refresh.go
@@ -113,8 +113,8 @@ func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
 
 func (d *Discovery) refresh(ctx context.Context) ([]*targetgroup.Group, error) {
 	now := time.Now()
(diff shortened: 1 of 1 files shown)


Howie66:
> Make metric `prometheus_sd_refresh_duration_seconds` more accurate. Because `d.refreshf` is the concrete implementation of the subclasses.

does it have any differences with the old version?why?

johncming:
> > Make metric `prometheus_sd_refresh_duration_seconds` more accurate. Because `d.refreshf` is the concrete implementation of the subclasses.
> 
> does it have any differences with the old version?why?

A slight change in performance. It's more about the precision of the metric's range.

roidelapluie:
I think that this measurement can be seen both ways, therefore I'd recommend not to take this, as it is really a tiny change.

roidelapluie:
Thank you anyway!

PR #7809, opened 2020-08-16
fix bug of discovery/refresh

johncming:
## bug 1

https://github.com/prometheus/prometheus/blob/8ec6f028541531755bb53a52bd8ea863a45530dd/discovery/refresh/refresh.go#L76-L84

The order in which the SELECT statement is executed determines the possibility of losing the `tgs` just obtained. A more explicit order would solve the bug.

## bug 2

https://github.com/prometheus/prometheus/blob/8ec6f028541531755bb53a52bd8ea863a45530dd/discovery/refresh/refresh.go#L76-L80

1. Comparing `ctx.Err()` and `err` at the same time may cause error loss. For example, no-nil `err` return from `refresh()` but `ctx` has not finished until comparing the two just equal.
2. `ctx.Err()` is passed through via `err` in the subclass implementation.

### DIFF
--- a/discovery/refresh/refresh.go
+++ b/discovery/refresh/refresh.go
@@ -74,10 +74,8 @@ func NewDiscovery(l log.Logger, mech string, interval time.Duration, refreshf fu
 func (d *Discovery) Run(ctx context.Context, ch chan<- []*targetgroup.Group) {
 	// Get an initial set right away.
 	tgs, err := d.refresh(ctx)
-	if err != nil {
-		if ctx.Err() != context.Canceled {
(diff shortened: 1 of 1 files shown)


roidelapluie:
1. It is ok to lose the tgs if the context is cancelled.

2. Yes we want to check the CTX error I'm case of the implementation wraps the ctx error before returning.

johncming:
> 1. It is ok to lose the tgs if the context is cancelled.
> 2. Yes we want to check the CTX error I'm case of the implementation wraps the ctx error before returning.

Thanks.

Yes, the loss of `tgs` is ok and ctx is only the cancel context which is ok.

I will close the WIP pr.

PR #7813, opened 2020-08-16
Issue template: Remove kind/bug label added by default

roidelapluie:
There are many false positive. It is better to tag the bugs manually
while adding priority and component.


brian-brazil:
:+1:

PR #7814, opened 2020-08-17
fix 'make docker' command's error

jimersylee:


roidelapluie:
Make docker is used by our CI system. You would need to run promu crossbuild -p linux/amd64 to get the binaries at the right place.

jimersylee:
> Thanks for your PR!
> 
> `make docker` is used by our CI system. You would need to run `promu crossbuild -p linux/amd64` to get the binaries at the right place.

When i did what you said,it works,thank you! I will close the PR

Issue #7571, opened 2020-07-13
Possible race condition in TSDB head chunks


roidelapluie:
**What did you do?**

Build Prometheus master d77b56e88e3d554a499e22d2073812b59191256c with -race

Ran it with a couple of exporters + --storage.tsdb.min-block-duration=10s

**10s is very aggressive and not sane. it is yet to see if the race can be produced in real situation**

**What did you expect to see?**

No errors

**What did you see instead? Under which circumstances?**

```
==================                                                                                                                [56/5719]
WARNING: DATA RACE                                                                                                                         
Read at 0x00c0002a0840 by goroutine 18:                                                                                                    
  github.com/prometheus/prometheus/tsdb/chunks.(*ChunkDiskMapper).Truncate()                                                               
      /home/roidelapluie/dev/prometheus/tsdb/chunks/head_chunks.go:684 +0x50e                                                              
  github.com/prometheus/prometheus/tsdb.(*Head).Truncate()                                                                                 
      /home/roidelapluie/dev/prometheus/tsdb/head.go:796 +0x62c                                                                            
  github.com/prometheus/prometheus/tsdb.(*DB).reload()                                                                                     
      /home/roidelapluie/dev/prometheus/tsdb/db.go:958 +0x171f                                                                             
  github.com/prometheus/prometheus/tsdb.(*DB).compactHead()                                                                                
      /home/roidelapluie/dev/prometheus/tsdb/db.go:792 +0x17c                                                                              
  github.com/prometheus/prometheus/tsdb.(*DB).Compact()                                                                                    
      /home/roidelapluie/dev/prometheus/tsdb/db.go:763 +0x302                                                                              
  github.com/prometheus/prometheus/tsdb.(*DB).run()                                                                                        
      /home/roidelapluie/dev/prometheus/tsdb/db.go:684 +0x427                                                                              
                                                                                                                                           
Previous write at 0x00c0002a0840 by goroutine 115:                                                                                         
  sync/atomic.AddInt64()                                                                                                                   
      /home/roidelapluie/godist/go/src/runtime/race_amd64.s:276 +0xb                                                                       
  github.com/prometheus/prometheus/tsdb/chunks.(*ChunkDiskMapper).write()                                                                  
      /home/roidelapluie/dev/prometheus/tsdb/chunks/head_chunks.go:397 +0x9e                                                               
  github.com/prometheus/prometheus/tsdb/chunks.(*ChunkDiskMapper).writeCRC32()       
[...]

PR #7815, opened 2020-08-17
Fix Possible Race Condition in TSDB

mneverov:
Fixes: #7571

### DIFF
--- a/tsdb/chunks/head_chunks.go
+++ b/tsdb/chunks/head_chunks.go
@@ -354,6 +354,7 @@ func (cdm *ChunkDiskMapper) cut() (returnErr error) {
 		return err
 	}
 
+	cdm.readPathMtx.Lock()
 	cdm.curFileSequence = seq
 	cdm.curFile = newFile
 	if cdm.chkWriter != nil {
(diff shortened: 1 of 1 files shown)


mneverov:
@codesome could you please take a look?

codesome (commented):
We missed 1 thing, we need to move https://github.com/prometheus/prometheus/blob/ca6f2bde9489d274719fcc17069f6a7aabb833b0/tsdb/chunks/head_chunks.go#L357-L363 inside the `cdm.readPathMtx.Lock()` right below. ~LGTM on the remaining part.~

codesome:
We have 2.21 rc going out today, lets try to get this in before that :)

cc @roidelapluie

tsdb/chunks/head_chunks.go
```diff
@@ -720,17 +720,20 @@ func (cdm *ChunkDiskMapper) DeleteCorrupted(originalErr error) error {
 
 	// Delete all the head chunk files following the corrupt head chunk file.
 	segs := []int{}
+	cdm.readPathMtx.Lock()
```
codesome, line 723: We should be holding the `RLock()` and `RUnlock()` here.

mneverov: Fixed

codesome (approved):
LGTM, thanks! Will merge after test is green.

